PR #1183, opened 2018-04-02
mapwidget-mobile: don't start map in London

### LINKED ISSUE #1182: Subsurface-mobile: don't start map in London
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [ ] New feature request
- [x] Discussion request

### Issue long description:

I think it would be better to not start the first transition in London, but to recognize "no previous location" and simply start "out" a couple of zoom levels from the first location for which the map button was pressed...

### Operating system:

Android (maps still don't work on iOS)

### Subsurface version:

latest beta: 4.7.7.145
### Steps to reproduce:

open dive log
tap on map button for a dive with gps info
### Current behavior:

show London, zoom out, move, zoom in
### Expected behavior:

start above the location of that dive, but zoomed out 4x or 8x, then zoom in
### Mentions:

@neolit123

neolit123:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

- on startup move the map center / zoom to [0,0] / max-zoom-out
this way if the user opens the map without a specific dive he/she would see the whole globe instead of London.
- when a dive from the dive list is selected for the first time or when a GPS location is selected for the first time, open the map centered on that location without animation. this avoids starting an animation from an arbitrary starting location such as [0,0].

i did not proceed with adding a small zoom-in (instead of the hard-center/pan) on first run because this would require changes that would affect the desktop version too.

as is, this change does not affect the desktop version.

### Changes made:

- added helper functions in `qmlmapwidgethelper.cpp/.h` and `MapPage.qml`
- added a `firstRun` variable in `MapPage.qml`

### Related issues:

Fixes #1182

### Release note:

the "bigger" release note about adding the map widget support to mobile should be enough here.

### Documentation change:

none

### Mentions:

@dirkhh

### DIFF
--- a/map-widget/qmlmapwidgethelper.cpp
+++ b/map-widget/qmlmapwidgethelper.cpp
@@ -23,6 +23,15 @@ MapWidgetHelper::MapWidgetHelper(QObject *parent) : QObject(parent)
 	        this, SLOT(selectedLocationChanged(MapLocation *)));
 }
 
+QGeoCoordinate MapWidgetHelper::getCoordinatesForUUID(QVariant dive_site_uuid)
+{
+	const uint32_t uuid = qvariant_cast<uint32_t>(dive_site_uuid);
+	struct dive_site *ds = get_dive_site_by_uuid(uuid);
+	if (!ds || !dive_site_has_gps_location(ds))
+		return QGeoCoordinate(0.0, 0.0);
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
This looks like a reasonable solution to the issue.
Not exactly what I envisioned, but definitely better than what we have and most likely "good enough".
Thanks

PR #1184, opened 2018-04-03
Add "Tags" column in Desktop app's dive list view

dje29:
Add DiveItem::displayTags helper method to return Tags as a QString
New Tags column is
 by default inserted before "Photos" column
 by default disabled

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This change simply adds the possibility to display Dive's tags in the dive list view on Desktop app. 

### Release note:

Not sure if this kind of small change did need a CHANGELOG update, I did add one just in case.

### Documentation change:

A new Tags column can be enabled in Desktop app's dive list view. The new column is by default inserted before Photos column. It is not enabled by default to keep current behavior.

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -36,6 +36,7 @@ static QVariant dive_table_alignment(int column)
 	case DiveTripModel::SUIT:
 	case DiveTripModel::CYLINDER:
 	case DiveTripModel::GAS:
+	case DiveTripModel::TAGS:
 	case DiveTripModel::PHOTOS:
 	case DiveTripModel::COUNTRY:
 	case DiveTripModel::LOCATION:
@@ -136,6 +137,9 @@ QVariant DiveItem::data(int column, int role) const
 		case MAXCNS:
 			retVal = dive->maxcns;
 			break;
+		case TAGS:
+			retVal = displayTags();
+			break;
 		case PHOTOS:
 			retVal = countPhotos(dive);
 			break;
@@ -185,6 +189,9 @@ QVariant DiveItem::data(int column, int role) const
 				retVal = QString("%1%").arg(dive->maxcns);
 			else
 				retVal = dive->maxcns;
+			break;
+		case TAGS:
+			retVal = displayTags();
 			break;
 		case PHOTOS:
 			break;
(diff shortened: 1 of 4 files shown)


qt-models/divetripmodel.cpp
```diff
@@ -273,8 +283,7 @@ QVariant DiveItem::data(int column, int role) const
 			break;
 		case LOCATION:
 			retVal = tr("Location");
-			break;
-		}
+			break;		}
```
neolit123: probably this should not have been changed?

qt-models/divetripmodel.cpp
```diff
@@ -409,6 +418,17 @@ QString DiveItem::displayWeightWithUnit() const
...
 }
 
+QString DiveItem::displayTags() const
+{
+	struct dive *dive = get_dive_by_uniq_id(diveId);
+	if (!dive->tag_list)
+		return QString();
+
+	char buf[1024];
+	taglist_get_tagstring(dive->tag_list, buf, 1024);
```
neolit123, line 429: i think that our `taglist_get_tagstring()` could have been implemented in a better way.
https://github.com/Subsurface-divelog/subsurface/blob/master/core/dive.c#L3023
if the strings don't fit the buffer size it does a `return i;` which isn't the correct way to approach this generic `fit-this-in-a-buffer` problem.

the return is also never used in our code base.
your code is fine given it follows the our existing usage of `taglist_get_tagstring()` but this function needs changes.

i can take a look at that when i get some free time.

neolit123 (commented):
i think the change in terms of code is good apart from a small white space issue. 

but i question the addition in terms of screen real-estate.

a dive can have dozens of tags added to it. having the tags listed in the dive list might be a bit *too much* detail to have at all times and what size does one need for the dive list column to view all of them?

@dje29 without compiling and looking at this i would give it a -1. can you give us a screenshot? thanks.

@dirkhh 
what do you think about this change?

dirkhh (approved):
I like this overall. It's off by default, I know a couple of people who use tags in a very structured manner and this allows "at a glance" review of the tags. I think it will be a minority of users who use this, but that's true for a number of our features.
I mark this as approved here - but please fix the whitespace error...

dje29:
Oups, sorry for that coding style mistake! '}' is now back to its rightful place (the next line).

@neolit123 Here is a screenshot of the extra column when tags do not fit the current column size, Qt cuts the string and adds '...' to show that some tags are not displayed.
tags

Regarding the long tag list (more than 1024) case, I'm still thinking through a proposal at the time of writing :) I currently have some free time (after a year not contributing) so I could look into that topic with some guidance regarding coding guidelines.

neolit123:
@dje29 thanks for the update.

PR #1185, opened 2018-04-03
Change Desktop app dive list headers to be all 'left aligned'

dje29:
It looked strange that dive list headers were aligned inconsistently.
Having numerical values right aligned is good, but headers should still
be aligned all the same way.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix inconsistency in dive list headers alignment. 

### Changes made:

1) Return Qt::AlignLeft + Qt::AlignVCenter for all headers' TextAlignmentRole

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -473,7 +473,7 @@ QVariant DiveTripModel::headerData(int section, Qt::Orientation orientation, int
 
 	switch (role) {
 	case Qt::TextAlignmentRole:
-		ret = dive_table_alignment(section);
+		ret = (QVariant)(Qt::AlignLeft | Qt::AlignVCenter);
 		break;
 	case Qt::FontRole:
 		ret = defaultModelFont();


qt-models/divetripmodel.cpp
```diff
@@ -452,7 +452,7 @@ QVariant DiveTripModel::headerData(int section, Qt::Orientation orientation, int
 
 	switch (role) {
 	case Qt::TextAlignmentRole:
-		ret = dive_table_alignment(section);
+		ret = Qt::AlignLeft + Qt::AlignVCenter;
```
bstoeger: Not wrong per se, but `ret = Qt::AlignLeft | Qt::AlignVCenter` would be more idiomatic.

dirkhh:
I don't see a reason to change what we have...
Besides an aesthetic preference, is there an actual reason to change this?

dje29:
@dirkhh Indeed this change is cosmetic matter I do find the updated version looking more consistent. And I must say most app I know having such kind of list/table view render their headers this way.
(One could argue as well that change results in one less function call but I don't believe this will make you vote in favor of the change :D :D :D)
headers before
headers after
@neolit123 I have now updated the code using | instead of +

neolit123:
@dirkhh do you approve of the image that @dje29 presented?

dirkhh:
No, I really don't see a reason to change the alignment that we have.
NAK from me.

PR #1187, opened 2018-04-04
Change taglist_get_tagstring to support 'unlimited' tag list size

dje29:
Previous taglist_get_tagstring signature/implementation did not allow
handling of cases where inputted buffer could not contain all tags.
New implementation allocates buffer based on pre-computed size allowing to
insert all tags in the returned string.

Added get_taglist_string in qthelper to handle conversion to QString
Added TestTagList with tests for taglist_get_tagstring

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a follow up on our discussion regarding this issue in #1184 when user would use so many tags that generated string would not fit in the 1024 buffer. I have (more or less) followed @neolit123 implementation proposal. 

### Changes made:

I've described all changes in the commit message.

Since I do not like to write code without tests, I've added some too.

### Release note:

Note sure if this change needs a mention in the CHANGELOG.md plus I would not know what topic to head it with, is there a place where a list of topics to be used here is maintained?

### Mentions:

@neolit123

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -12,6 +12,7 @@
 #include "divelist.h"
 #include "qthelper.h"
 #include "metadata.h"
+#include "membuffer.h"
 
 /* one could argue about the best place to have this variable -
  * it's used in the UI, but it seems to make the most sense to have it
@@ -3020,30 +3021,28 @@ void taglist_cleanup(struct tag_entry **tag_list)
 	}
 }
 
-int taglist_get_tagstring(struct tag_entry *tag_list, char *buffer, int len)
+char *taglist_get_tagstring(struct tag_entry *tag_list)
 {
-	int i = 0;
-	struct tag_entry *tmp;
-	tmp = tag_list;
-	memset(buffer, 0, len);
+	bool first_tag = true;
+	struct membuffer b = { 0 };
+	struct tag_entry *tmp = tag_list;
 	while (tmp != NULL) {
-		int newlength = strlen(tmp->tag->name);
-		if (i > 0)
-			newlength += 2;
-		if ((i + newlength) < len) {
-			if (i > 0) {
-				strcpy(buffer + i, ", ");
-				strcpy(buffer + i + 2, tmp->tag->name);
+		if (!empty_string(tmp->tag->name)) {
+			if (first_tag) {
+				put_format(&b, "%s", tmp->tag->name);
+				first_tag = false;
 			} else {
-				strcpy(buffer, tmp->tag->name);
+				put_format(&b, ", %s", tmp->tag->name);
 			}
-		} else {
-			return i;
 		}
-		i += newlength;
 		tmp = tmp->next;
 	}
-	return i;
+	/* Ensures we do return null terminated empty string for:
+	 *  - empty tag list
+	 *  - tag list with empty tag only
+	 */
+	mb_cstring(&b);
+	return detach_buffer(&b);
 }
 
 static inline void taglist_free_divetag(struct divetag *tag)
--- a/core/dive.h
+++ b/core/dive.h
@@ -264,10 +264,13 @@ struct tag_entry *taglist_added(struct tag_entry *original_list, struct tag_entr
 void dump_taglist(const char *intro, struct tag_entry *tl);
 
 /*
- * Writes all divetags in tag_list to buffer, limited by the buffer's (len)gth.
- * Returns the characters written
+ * Writes all divetags form tag_list into internally allocated buffer
+ * Function returns pointer to allocated buffer
+ * Buffer contains comma separated list of tags names or null terminated string
+ *
+ * NOTE! The returned buffer must be freed once used.
  */
-int taglist_get_tagstring(struct tag_entry *tag_list, char *buffer, int len);
+char *taglist_get_tagstring(struct tag_entry *tag_list);
 
 /* cleans up a list: removes empty tags and duplicates */
 void taglist_cleanup(struct tag_entry **tag_list);
--- a/tests/testtaglist.cpp
+++ b/tests/testtaglist.cpp
@@ -1,0 +1,82 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "testtaglist.h"
+#include "core/dive.h"
+
+void TestTagList::initTestCase()
+{
+	taglist_init_global();
+}
+
+void TestTagList::cleanupTestCase()
+{
+	taglist_free(g_tag_list);
+	g_tag_list = NULL;
+}
+
+void TestTagList::testGetTagstringNoTags()
+{
+	struct tag_entry *tag_list = NULL;
+	char *tagstring = taglist_get_tagstring(tag_list);
+	QVERIFY(tagstring != NULL);
+	QCOMPARE(*tagstring, '\0');
+}
+
+void TestTagList::testGetTagstringSingleTag()
+{
+	struct tag_entry *tag_list = NULL;
+	taglist_add_tag(&tag_list, "A new tag");
+	char *tagstring = taglist_get_tagstring(tag_list);
+	QVERIFY(tagstring != NULL);
+	QCOMPARE(QString::fromUtf8(tagstring), QString::fromUtf8("A new tag"));
+	free(tagstring);
+}
+
+void TestTagList::testGetTagstringMultipleTags()
+{
+	struct tag_entry *tag_list = NULL;
+	taglist_add_tag(&tag_list, "A new tag");
+	taglist_add_tag(&tag_list, "A new tag 1");
+	taglist_add_tag(&tag_list, "A new tag 2");
+	taglist_add_tag(&tag_list, "A new tag 3");
+	taglist_add_tag(&tag_list, "A new tag 4");
+	taglist_add_tag(&tag_list, "A new tag 5");
+	char *tagstring = taglist_get_tagstring(tag_list);
+	QVERIFY(tagstring != NULL);
+	QCOMPARE(QString::fromUtf8(tagstring),
+		 QString::fromUtf8(
+			 "A new tag, "
+			 "A new tag 1, "
+			 "A new tag 2, "
+			 "A new tag 3, "
+			 "A new tag 4, "
+			 "A new tag 5"));
+	free(tagstring);
+}
+
+void TestTagList::testGetTagstringWithAnEmptyTag()
+{
+	struct tag_entry *tag_list = NULL;
+	taglist_add_tag(&tag_list, "A new tag");
+	taglist_add_tag(&tag_list, "A new tag 1");
+	taglist_add_tag(&tag_list, "");
+	char *tagstring = taglist_get_tagstring(tag_list);
+	QVERIFY(tagstring != NULL);
+	QCOMPARE(QString::fromUtf8(tagstring),
+		 QString::fromUtf8(
+			 "A new tag, "
+			 "A new tag 1"));
+	free(tagstring);
+}
+
+void TestTagList::testGetTagstringEmptyTagOnly()
+{
+	struct tag_entry *tag_list = NULL;
+	taglist_add_tag(&tag_list, "");
+	char *tagstring = taglist_get_tagstring(tag_list);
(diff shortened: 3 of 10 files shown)


core/dive.c
```diff
@@ -3020,30 +3020,38 @@ void taglist_cleanup(struct tag_entry **tag_list)
 	}
 }
 
-int taglist_get_tagstring(struct tag_entry *tag_list, char *buffer, int len)
+char* taglist_get_tagstring(struct tag_entry *tag_list)
```
neolit123: `char *tag...` rather than `char* tag...`

core/dive.c
```diff
@@ -3020,30 +3020,38 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
+		if (buffer_size > 0) {
+			buffer_size += 2;
 		}
-		i += newlength;
+		buffer_size += strlen(tmp->tag->name);
 		tmp = tmp->next;
 	}
-	return i;
+	/* Allocate buffer and generate string with ", " separator(s) */
+	if (buffer_size > 0) {
```
bstoeger: Please don't make it complicated for the caller (having to handle null). If there are no tags, return an empty string. Also makes the code here more straight-forward.

core/dive.c
```diff
@@ -3020,30 +3020,38 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
+		buffer = malloc(buffer_size + 1);
```
neolit123: you can just return `buffer` (`NULL`) here:
```
if (!buffer_size)
	return buffer;
```

core/dive.c
```diff
@@ -3020,30 +3020,38 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
+	}
+	if (buffer) {
```
neolit123:
```
if (!buffer) {
	fprintf(stderr, "taglist_get_tagstring(): cannot allocate buffer of size: u%\n);
	return NULL;
}
```
this would extract the contents of the branch bellow.

core/dive.c
```diff
@@ -3020,30 +3020,38 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
+		last_pos = buffer;
+		tmp = tag_list;
+		while(tmp != NULL) {
```
neolit123: space after `while`

core/dive.h
```diff
@@ -264,10 +264,12 @@ struct tag_entry *taglist_added(struct tag_entry *original_list, struct tag_entr
...
 /*
- * Writes all divetags in tag_list to buffer, limited by the buffer's (len)gth.
- * Returns the characters written
+ * Writes all divetags form tag_list into internally allocated buffer
+ * Function returns pointer to allocated buffer or NULL
+ *
+ * NOTE! The returned buffer must be freed once used.
  */
-int taglist_get_tagstring(struct tag_entry *tag_list, char *buffer, int len);
+char* taglist_get_tagstring(struct tag_entry *tag_list);
```
neolit123: `char *`

tests/testtaglist.h
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef TESTTAGLIST_H
+#define TESTTAGLIST_H
+
+#include <QtTest>
+
+class TestTagList : public QObject{
```
neolit123: add space before `{`

tests/testtaglist.cpp
```diff
@@ -0,0 +1,55 @@
...
+	taglist_add_tag(&tag_list, "A new tag 5");
+	char *tagstring = taglist_get_tagstring(tag_list);
+	QVERIFY(tagstring != NULL);
+	QCOMPARE(QString::fromUtf8(tagstring),
+		 QString::fromUtf8("A new tag"
+				   ", A new tag 1"
+				   ", A new tag 2"
+				   ", A new tag 3"
+				   ", A new tag 4"
+				   ", A new tag 5"));
```
neolit123: there is no let's change this to:
```
QCOMPARE(QString::fromUtf8(tagstring),
         QString::fromUtf8("A new tag, "
                           "A new tag 1, "
                           "A new tag 2, "
                           "A new tag 3, "
                           "A new tag 4, "
                           "A new tag 5")
);
```
notice:
- `, ` are at the end of the strings
- only spaces for the intent as this makes the indent consistent across tab sizes

core/dive.c
neolit123: i think it's OK to return `NULL` as this is also covered in the tests.

neolit123 (changes requested):
thanks for this. i've made some small comments about styling and the logic. please, amend.

extra points for including a test!

neolit123:
> Note sure if this change needs a mention in the CHANGELOG.md plus I would not know what topic to head it with, is there a place where a list of topics to be used here is maintained?

there is no need for adding change log entries for internal changes like this one - that do not affect the user directly (e.g. code cleanup).

we don't really have a list of topics, but for this one something like `Dive-tags:` is a good one.

core/dive.c
```diff
@@ -3020,30 +3020,38 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
-int taglist_get_tagstring(struct tag_entry *tag_list, char *buffer, int len)
+char* taglist_get_tagstring(struct tag_entry *tag_list)
 {
-	int i = 0;
-	struct tag_entry *tmp;
-	tmp = tag_list;
-	memset(buffer, 0, len);
+	char *buffer = NULL;
+	char *last_pos = NULL;
+	size_t buffer_size = 0;
```
neolit123: let's use `unsigned int` here instead of `size_t`, this will save a cast when doing the `fprintf` later on.

tests/testtaglist.cpp
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "testtaglist.h"
```
neolit123, line 2: i would have moved the test into a separate commit, but i guess this is fine too. see if you'd prefer that?
the first commit message would need an amend in that case.

core/dive.c
bstoeger: I think it's a really bad interface. The next person does a very reasonable
```
char *s = taglist_get_tagstring(...);
snprintf(buf, sizeof(buf), "Tags: %s\n", s);
free(s);
```
And boom  - strange and undefined behavior.

This code makes it more complicated for the caller *and* the callee. What's the advantage?

bstoeger:
Wouldn't `membuffer` with `put_format()` lead to much simpler code with a negligible performance impact?

neolit123:
it is possible to do it with `membuffer` but i dislike the fact that `realloc` will be called here for each tag entry.

bstoeger:
> it is possible to do it with membuffer but i dislike the fact that realloc will be called here for each tag entry.

It won't:
```
                /* round it up to not reallocate all the time.. */
                needed = needed * 9 / 8 + 1024;
                n = realloc(b->buffer, needed);
```
In the normal case (< 1024 bytes) you'll get only a single allocation.

core/dive.c
neolit123: i guess we can do that. please return `""` instead of `NULL`.
but this will also affect the test.

core/dive.c
```diff
@@ -3020,30 +3020,38 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
-			if (i > 0) {
-				strcpy(buffer + i, ", ");
-				strcpy(buffer + i + 2, tmp->tag->name);
-			} else {
-				strcpy(buffer, tmp->tag->name);
-			}
-		} else {
-			return i;
+		if (buffer_size > 0) {
+			buffer_size += 2;
```
neolit123: this needs a small change. notice if a tag length is returned as `0` the size will still increment by `2`.
to avoid this store the old size and compare it instead of comparing to `0`.

core/dive.c
```diff
@@ -3020,30 +3020,38 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
+	if (buffer_size > 0) {
+		buffer = malloc(buffer_size + 1);
+	}
+	if (buffer) {
+		last_pos = buffer;
+		tmp = tag_list;
+		while(tmp != NULL) {
+			if (last_pos != buffer) {
+				strcpy(last_pos, ", ");
+				last_pos += 2;
```
neolit123: skip adding `, ` for empty tags?

neolit123:
> It won't:
> In the normal case (< 1024 bytes) you'll get only a single allocation.

 i guess that's true. i leave it to @dje29 to decide if he wants to use `membuffer` with `put_format`.
examples in `core/save-xml.c`.

core/dive.c
bstoeger: My idea was to remove both `if`-conditions - if there are not tags it would simply fall back to an empty string, no?

neolit123: let's see the next iteration that @dje29 comes up with.

dirkhh:
I wonder why the test failed - that needs to be investigated.
Great reviews, thanks @neolit123 and @bstoeger 
I agree that I'd love to see this use a membuffer. That would result in easier, cleaner code, IMHO.
This is always the balance of being welcoming and supportive to new contributors, and pushing for consistency and adherence to our standards... @dje29 would you be willing to change the implementation to use that instead?

dje29:
Here is the change with new membuffer based implementation, I did not know this utility, it is quite handy and indeed made the code much simpler. Happy to have learned about it.

- As proposed by @bstoeger now function returns null terminated empty string when no tags are passed, tests is updated accordingly
- Fixed the issue raised by @neolit123 regarding printing ", " when list contains empty tag (added test cases for that)
- I have fixed coding style mistakes and hopefully did not introduce new ones... 

Working on this contribution made me think that CONTRIBUTING.md could be extended adding
1.  Section with some coding guidelines (such has use membuffer instead of malloc, strcpy, ... when working with strings), but certainly there are others that currently resid only in your heads and can only be discovered "the hard way" :) It would simply need the habit to patch the section every time a review like the one we just did is made. I'll prepare a pull request with a proposal for such entry
2. Mention of the CHANGELOG.md use (with list of topix that can be used) since currently new contributors discover this only once they create a pull request.

core/dive.c
```diff
@@ -3020,30 +3021,30 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
 		}
-		i += newlength;
 		tmp = tmp->next;
 	}
-	return i;
+	/* Ensure we do return null terminated empty string for:
+	 *  - empty tag list
+	 *  - tag list with empty tag only
+	 */
+	if (b.len == 0) {
```
bstoeger: You can replace this `if`-block by
```
mb_cstring(&b);
```
I think the membuffer interface was  intended that way anyway.

core/dive.c
```diff
@@ -3020,30 +3021,30 @@ void taglist_cleanup(struct tag_entry **tag_list)
...
+	struct tag_entry *tmp = tag_list;
 	while (tmp != NULL) {
-		int newlength = strlen(tmp->tag->name);
-		if (i > 0)
-			newlength += 2;
-		if ((i + newlength) < len) {
-			if (i > 0) {
-				strcpy(buffer + i, ", ");
-				strcpy(buffer + i + 2, tmp->tag->name);
+		if (*tmp->tag->name != '\0') {
```
bstoeger: You could write this as
```
if (!empty_string(tmp->tag->name))
```
This handles `name` being the null pointer gracefully.

dirkhh:
@dje29 thanks for the rewrite.
I'd love a PR against CONTRIBUTING.md - I'm always happy to make things easier for new contributors joining our team.
As I mentioned on the mailing list, I'm planning to cut a release today, so we have a day or two to maybe consider the review by @bstoeger  :-)

dje29:
@dirkhh I preferred to fix the code as the topic was still fresh in my head, but it is indeed a good idea to stall the merge to after the current release will be done.

dirkhh:
Very nice - I like it. And the code is so much easier to read.
Thanks @dje29

PR #1151, opened 2018-03-05
[TESTING] Play videos in profile plot

bstoeger:
### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Note! This is only for testing, not an actual pull request.

The point of this PR is to let people test if my approach for playing videos in the profile widget works. A few comments:
1) The code here is quite verbose, since using a `QGraphicsVideoItem` did not work on my setup. Therefore, video frames are shown by hand.
2) Extracting stills as thumbnails is not yet implemented owing to a Qt bug discussed in #359 (this must be performed in the main thread or an external library must be used). This will need some refactoring of the thumbnail code.
3) To mark a file as a video, a null QImage is added to a new video-thumbnail-cache. This is for forward-compatibility: Once thumbnail-extraction is supported, actual images will be written to this cache.
4) At the moment, when a null-image is found in the video-thumbnail-cache, a dummy "video" icon is shown. A proper icon is needed.
5) Since at the moment there is no extraction of stills, videos are identified by file extensions. Since Qt provides no list of supported file formats, this is hard-coded. 

To test:
1) Go to dive picture tab
2) Upload an mp4 or mpeg.
3) Switch to different dive and back [so that the proper thumbnail is loaded].
4) Move video onto profile graph.
5) Hover mouse over video in profile graph.

Please let me know if this works for you.

PR #1189, opened 2018-04-05
Cleanup: delete stale DivePictureItems right away

bstoeger:
The function ProfileWidget2::plotPictures() replots all pictures
in the profile plot. Before doing so, it deletes the previously
displayed items.

This was done using deleteLater(), which did not free the memory
immediately, but only later in the event loop. Change this to properly
delete the item.

By deleting the item right away, it is not necessary to hide the item
beforehand. Remove the corresponding function call.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This cleanup frees memory earlier, which makes #1151 not behave as terribly as it does for a large number of pictures.

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1976,10 +1976,8 @@ void ProfileWidget2::keyEscAction()
 
 void ProfileWidget2::plotPictures()
 {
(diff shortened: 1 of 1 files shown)


bstoeger:
Superfluous after #1191.

PR #1191, opened 2018-04-06
Profile: On dataChanged() only update pictures that actually changed

bstoeger:
Only update those pictures of the DivePictureModel that actually changed.
This will be useful once pictures are loaded incrementally.

To do so, replace the pictures array by an array with stable ids. Before
this commit, not-shown pictures are left out of the pictures array, which
makes the mapping from DivePictureModel-ids to the picture array index
non-trivial.

Replace the QList<DivePictureItem *> by a std::vector<std::unique_ptr<DivePictureItem>>
to easy memory management. Sadly, owing to COW semantics, QVector is incompatible
with QScopedPointer.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Only update changed pictures in the profile plot. For discussion, see #1179.

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -778,7 +778,10 @@ void ProfileWidget2::plotDive(struct dive *d, bool force)
 		DivePlannerPointsModel *model = DivePlannerPointsModel::instance();
 		model->deleteTemporaryPlan();
 	}
-	plotPictures();
+	if (printMode)
+		clearPictures();
+	else
+		plotPictures();
 #endif
 
 	// OK, how long did this take us? Anything above the second is way too long,
@@ -1103,7 +1106,7 @@ void ProfileWidget2::setProfileState()
 
 	disconnectTemporaryConnections();
 #ifndef SUBSURFACE_MOBILE
-	connect(DivePictureModel::instance(), SIGNAL(dataChanged(QModelIndex, QModelIndex)), this, SLOT(plotPictures()));
+	connect(DivePictureModel::instance(), &DivePictureModel::dataChanged, this, &ProfileWidget2::updatePictures);
 	connect(DivePictureModel::instance(), SIGNAL(rowsInserted(const QModelIndex &, int, int)), this, SLOT(plotPictures()));
 	connect(DivePictureModel::instance(), SIGNAL(rowsRemoved(const QModelIndex &, int, int)), this, SLOT(plotPictures()));
 #endif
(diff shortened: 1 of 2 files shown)


atdotde (commented):
I admit I am uneasy with adding another source of complex data types (std::xxx in addition to Qt types). Obviously, it's my lacking of understanding of C++, but I guess (if I understand correctly what is going on) I would indirect rather via a QHash. In any case, I would like to hear @tcanabrava or @thiagomacieira  on this.

profile-widget/profilewidget2.cpp
```diff
@@ -1103,7 +1106,7 @@ void ProfileWidget2::setProfileState()
 
 	disconnectTemporaryConnections();
 #ifndef SUBSURFACE_MOBILE
-	connect(DivePictureModel::instance(), SIGNAL(dataChanged(QModelIndex, QModelIndex)), this, SLOT(plotPictures()));
+	connect(DivePictureModel::instance(), &DivePictureModel::dataChanged, this, &ProfileWidget2::updatePictures);
```
thiagomacieira: Unrelated code clean up?

profile-widget/profilewidget2.cpp
```diff
@@ -1974,26 +1977,46 @@ void ProfileWidget2::keyEscAction()
...
+		if (!offsetSeconds) {
+			pictures[picItemNr].reset();
 			continue;
-		DivePictureItem *item = new DivePictureItem();
+		}
+		if (!pictures[picItemNr]) {
+			pictures[picItemNr].reset(new DivePictureItem);
+			scene()->addItem(pictures[picItemNr].get());
+		}
+		DivePictureItem *item = pictures[picItemNr].get();
```
thiagomacieira: Suggestion: if you move this line before the `if`, you can save yourself from repeating `pictures[picItemNr]` three times. It's ok to `.get()` a null unique pointer -- you get nullptr.

thiagomacieira:
Code looks good, I don't see any problems in the change. I don't know the rest of the code to say whether there would be problems with the parts that aren't changed, but it's unlikely.

COW semantics made sense in the 90s when Qt started, before move semantics existed. There's also the fact that most developers aren't as good as you are, so providing it results in better code for an average developer. Those who know how to use `std::vector` properly like you can use it.

I'll also defend UTF-16. It's a lot easier to manipulate than UTF-8, but doesn't require as much space as UCS-4. Microsoft still recommends it and all the major APIs use it: ICU, Java, Cocoa and Win32.

dirkhh:
OK, this appears to have gotten the seal of approval, merging it now

PR #1194, opened 2018-04-07
Add section describing use of CHANGELOG.md in CONTRIBUTING.md

dje29:
Add description about file format
Add list of Areas that can be used

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

As discussed in one of my previous pull request #1187 here is a proposal for section describing use of CHANGELOG.md in CONTRIBUTING.md.
I have made recommendations (such as use of /) based on the usage I've seen being made in the history of the file. (Same for list of areas).
Doing so I realized that Desktop and Mobile is not really the same level of 'Area' compared to Bluetooth
Profile, Map-Widget... Not sure how much that would be useful for Release Note generation, but would it make sense to rather do something like two level Area description:
Target:Feature: Description
Targets being:
* Core (goes in all release notes)
* Desktop (goes in desktop app release notes only)
* Mobile (goes in all mobile app release notes) 
* Android (goes in android app release notes only)
* Ios (goes in Ios app release notes only)
Features being:
Planner, Import, ....

Here are some example how it'd look like:
Desktop:Planner: bla bla bla
Desktop/Mobile:Map-Widget: bla bla bla
Core:Import: bla bla bla
Android:Import: bla bla bla

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,27 @@ https://en.wikipedia.org/wiki/Imperative_mood
 
 Example with gitk
 
+Additionally when important changes to behaviors, fixes or new feature are introduced an entry must be added to  CHANGELOG.md file. Always add new entries at the very top of the file above other existing entries. Use this layout for new entries:
+
+Area: Details about the change [reference thread / issue]
+
+When multiple areas are affected list the areas separated with a /:
+
+Area1/Area2: Detail about the change [reference thread / issue]
+
+Here is a (non exhaustive) list of Areas that can be used:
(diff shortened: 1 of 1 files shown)


dirkhh:
I like the edit to the CONTRIBUTING.md
I don't want this to get too complex, I don't want to overthink this. We are not a project at the scale of Linux or k8s... so yes, the structure could be more orthogonal and more precise, but I think what we have now is sufficient.

PR #1206, opened 2018-04-18
[RFC] Dive pictures: Make picture code more scalable

bstoeger:
To potentially conserve memory, don't keep copies of scaled thumbnails.
Scale the thumbnails on demand.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Currently, all thumbnails are loaded into memory on startup and are saved to disk on exit. For big divelogs this can amount to a substantial memory-consumption. Moreover, at least three references of the thumbnails of the currently selected dive(s) are kept by the application:
1) In `thumbnailCache` with maximum size.
2) In `DivePictureModel`, according to current `zoomLevel` (for photo-tab).
3) In `DivePictureModel`, default size for the profile plot.

This commit attempts to make the dive picture code more scalable, by loading thumbnails on demand. Since the thumbnails are loaded from disk, displaying thumbnails now takes longer than previously. Therefore, this is now done in a background thread to avoid blocking the UI. Switching between dives with many pictures should now be "snappier".

Currently, there is only a single background thread, i.e. all thumbnailing code is serialized. This can be configured owing to the use of a `QThreadPool`.

As an added bonus of the new code, when adding an image from a remote URL, the image is shown after loading without having to switch between dives.

### Changes made:

1) Store thumbnails in individual files
2) Convert old thumbnail hash to individual files
3) Load thumbnails on demand
4) Do thumbnail processing in background thread
5) Update thumbnails after download

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -3843,18 +3843,6 @@ void picture_free(struct picture *picture)
 	free(picture);
 }
 
-// When handling pictures in different threads, we need to copy them so we don't
-// run into problems when the main thread frees the picture.
-
-struct picture *clone_picture(struct picture *src)
-{
-	struct picture *dst;
-
-	dst = alloc_picture();
-	copy_pl(src, dst);
-	return dst;
-}
-
 // Return true if picture was found and deleted
 bool dive_remove_picture(struct dive *d, const char *filename)
 {
--- a/core/dive.h
+++ b/core/dive.h
@@ -429,7 +429,6 @@ struct picture {
 	for (struct picture *picture = (_divestruct).picture_list; picture; picture = picture->next)
 
 extern struct picture *alloc_picture();
-extern struct picture *clone_picture(struct picture *src);
 extern bool dive_check_picture_time(struct dive *d, int shift_time, timestamp_t timestamp);
 extern void dive_create_picture(struct dive *d, const char *filename, int shift_time, bool match_all);
 extern void dive_add_picture(struct dive *d, struct picture *newpic);
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -140,7 +140,8 @@ QImage getHashedImage(const QString &file)
 	return res;
 }
 
-Thumbnailer::Thumbnailer()
+Thumbnailer::Thumbnailer() : failImage(QImage(":filter-close").scaled(maxThumbnailSize(), maxThumbnailSize(), Qt::KeepAspectRatio)), // TODO: Don't misuse filter close icon
+			     dummyImage(QImage(":photo-icon").scaled(maxThumbnailSize(), maxThumbnailSize(), Qt::KeepAspectRatio)) // TODO: Don't misuse photo-icon
 {
 	// Currently, we only process one image at a time. Stefan Fuchs reported problems when
 	// calculating multiple thumbnails at once and this hopefully helps.
@@ -200,16 +201,16 @@ static void addThumbnailToCache(const QImage &thumbnail, const QString &picture_
 	file.commit();
 }
 
-void Thumbnailer::processItem(QString filename, int size)
+void Thumbnailer::processItem(QString filename)
 {
 	QImage thumbnail = getThumbnailFromCache(filename);
 
 	if (thumbnail.isNull()) {
 		thumbnail = getHashedImage(filename);
 		if (thumbnail.isNull()) {
-			// TODO: Don't misuse filter close icon
-			thumbnail = QImage(":filter-close").scaled(size, size, Qt::KeepAspectRatio);
+			thumbnail = failImage;
 		} else {
+			int size = maxThumbnailSize();
 			thumbnail = thumbnail.scaled(size, size, Qt::KeepAspectRatio);
 			addThumbnailToCache(thumbnail, filename);
 		}
@@ -220,7 +221,7 @@ void Thumbnailer::processItem(QString filename, int size)
 	workingOn.remove(filename);
 }
 
-QImage Thumbnailer::fetchThumbnail(PictureEntry &entry, int size)
+QImage Thumbnailer::fetchThumbnail(PictureEntry &entry)
 {
 	QMutexLocker l(&lock);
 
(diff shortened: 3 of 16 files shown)


dirkhh:
This seems useful. It would be nice to get some actual data on memory saved vs. compute impact - but even without that I think I'm ok with this.
@atdotde - any comments?

bstoeger:
> It would be nice to get some actual data on memory saved vs. compute impact 

Unfortunately, both of these are really hard to measure.

I added two commits implementing a more radical approach at conserving memory, which definitely *should* be measurable.

Currently, we load *all* thumbnails into memory on startup. Obviously this doesn't scale to @sfuchs79 style logs with thousands(?) of images. I was thinking about writing one hash file per dive, but this would mean doing some fancy locking, caching, etc.pp. So it occurred to me that the simplest thing will be simply writing a file per thumbnail. Let the people who are actually good at this stuff (OS developers) take care of all the complicated stuff.

This will currently make switching between dives somewhat slower, because the thumbnails are fetched from disk. But in my tests with 200 images/dive, this was still very bearable. In the future, I plan to do the fetching from disk without blocking the UI (see also #1151).

Going with this route, in a next step we can limit the number of pictures shown in the photo-tab to a sane value (if e.g. the user selects all dives at once) to avoid spikes in memory-use.

I also added a commit which converts the old hash to single files. I think we don't need to localize the corresponding ProgressDialog, since it will only be shown once.

There's one point which should be discussed: In the current implementation, the thumbnail-files are composed of an integer and an QImage. The integer is currently set to 0. I plan to use this to differentiate plain images from videos (again, see #1151).

@sfuchs79: Would you perhaps have time to give this a try?

bstoeger:
Cherry-picked the appropriate commits from #1151. Memory consumption is reduced for me and interface is quite snappy. A perceived disadvantage might be that thumbnails are loaded incrementally.

Anyone willing to test?

atdotde:
I haven't looked at the code but I can confirm this is nice and snappy (for the few pictures I have in my dives) and no obvious flaws.

bstoeger:
Thanks. Are you fine with prepending an integer in the thumbnail files to indicate the media-type (picture/video)?

atdotde:
Sure. Can you do it in a way that is backwards compatible? Or at least use a different file name?

bstoeger:
The "new-style" thumbnails are saved in the folder ".suburface/thumbnails". Older versions will simply ignore the folder and recreate the thumbnails if needed. So switching between old and new versions should have no adverse effects, besides unnecessary thumbnail recalculations. But I have only tested old->new version, not the other direction.

dirkhh:
The iOS build fails with QThreadPool not found. Hmm. That needs looking into.

What are the effects for someone who has existing cache (so the common "upgrade" scenario)?

bstoeger:
> The iOS build fails with QThreadPool not found. Hmm. That needs looking into.

Mobile on desktop builds and runs for me - so not sure what I can do...?

> What are the effects for someone who has existing cache (so the common "upgrade" scenario)?

Commit dbf1cb993e3a0aad2eb20e7180bde083bd58c7a8 converts the big hash to individual files on startup. This should happen only once (except if the application does not close regularly, i.e. crash or termination by user).

bstoeger:
PS: This still needs clip art placeholders for "Image loading" and "Image loading failed". Currently, I misuse random icons which look horrible owing to scaling. How can we proceed in this respect?

dirkhh:
Getting better icons for something is always challenging. Google searches with license restrictions for scalable icons often help.
As for iOS failing - now that I had some sleep I'm thinking this is caused by the way I have minimized the size of the iOS Qt package, so I think you can ignore this - this is something that I need to fix.

bstoeger:
To make the maximum thumbnail-size scalable, we need SVG icons. The filter-close icon, which I misuse for load-failed, is an SVG. I added a few lines to render it to the appropriate size. For the moment I think it is clear what the icon means, but ultimately, someone will have to find a proper icon.

The photo-icon was a PNG, so I imported a camera-icon from the KDE-breeze icon set, which is licensed under LGPL. Is that OK? It feels weird to import other people's work without asking them.

From an aesthetic point of view, this all "fits like a fist to the eye" to use a German idiom, but this is simply not something that I can take care of.

dirkhh:
reusing icons from breeze is fine - we already do this for Subsurface-mobile
I'm not in love with simply copying them (instead of what we do in the mobile build), but it's not a licensing issue

bstoeger:
> instead of what we do in the mobile build

I see the `scripts/mobilecomponents.sh` thing. But do we really want to add such build-complexities to the desktop version for a single icon? If there was a bigger standardization effort to breeze icons, then sure. But for one (hopefully temporary) icon?

dirkhh:
> I see the scripts/mobilecomponents.sh thing. But do we really want to add such build-complexities to the desktop version for a single icon? If there was a bigger standardization effort to breeze icons, then sure. But for one (hopefully temporary) icon?

Clearly I am speaking in tongues these days.
No, we don't want to do that. All I am saying is that this is fine from a licensing point of view, that we already do it, but that I don't want the next person to turn around and copy all the mobile breeze icons into the sources as well.

sfuchs79:
@bstoeger 
Sorry for not giving a feedback for such a long time.
I was today finally able to test this.
In short: This is a really nice implementation. Very good performance and finally solves my issues with missing thumbnails. It also works nicely with dives with large number of pictures (~100). I didn't see any unwanted behavior after some quick tests. Thumbs up and thanks from my side!

sfuchs79:
So yes, I can confirm that this huge enhancement closes #1179

bstoeger:
> So yes, I can confirm that this huge enhancement closes #1179

That's nice to hear, because I wouldn't have thought that it would fix the problem. There's still room for improvement, but let's see if this causes any problems in the wild first.

dirkhh:
I've gone through a complete review. There are patterns in the code that I am not super familiar with, but I don't see anything WRONG per se. So I think the best bet is to merge it and move forward.

PR #1217, opened 2018-04-23
Updated INSTALL, based on first time building experience.

janiversen:
INSTALL is updated to reflect how to install/build subusurface
on Mac. It has also been updated with several generic parts.

3 parts are still outstanding (to come soon):
QtWebKit, need to document how to download/build
Mac, final touches
iOS, need to document the build process

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x ] Documentation change
- [ ] Language translation

I decided to file this pull-request, so you all had a chance of reviewing the general
sections of the INSTALL file.

I have been very careful to document the steps needed (on my Mac) to get subsurface
compiled, without depending on previous installed software. I still have a few finishing
touches outstanding for the Mac build (problem with libtool, QtWebKit), and then I will
move on to the iOS build.

I have tried to integrate the information I have been given in multiple mails, Thanks for 
using energy to explain things that are logical to the experienced developer.

My English is nowhere perfect, so please feel free to update the text as you see fit.

### Release note:

Updated INSTALL to reflect current build procedure

I did on purpose not add it to CHANGELOG.md, as I am unsure if you want to document
developer only change in the release notes.

Btw. some other projects generate CHANGELOG.md automatically by extracting all commit texts.

### Documentation change:

No user experience change

### Mentions:

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -76,16 +76,15 @@ To save time and disk space you can unselect Android and IOS packages
 - Qt Network Authorization
 - Qt Remote Objects
 - Qt WebGL Streaming
-- Qt Script
 - Qt 3D2
 - Qt Canvas 3D
 - Qt Extras
 This can be done later by use of MaintenanceTool.app in your Qt folder.
 
-Remark: most Qt installations does not make Qt available on the command
+Remark: most Qt installations do not make Qt available on the command
 line, which is needed. In order to make Qt available you need to 
-add a symbol link (do something like
-ln -s <Qt location>/<version>/<type>/bin/qmake /usr/local/bin/qmake
+add qmake to the path, do something like
+PATH=$PATH:<Qt location>/<version>/<type>/bin
 
 QtWebKit is needed, if you want to print, but no longer part of Qt5,
 so you need to download it and compile. In case you just want to test
@@ -102,9 +101,8 @@ libgit2 0.23 or newer.
 cmake build system
 ------------------
 
-As of Subsurface 4.5 we have switched our build system to cmake.
-qmake based builds are no longer supported. Remark even though
-no longer supported, qmake is still used by googlemaps and the iOS build.
+Our main build system is based on cmake. But qmake is needed
+for the googlemaps plugin and the iOS build.
 
 Download from https://cmake.org/download and follow the instructions
 to install it or alternatively follow the instruction specific to a
--- a/INSTALL
+++ b/INSTALL
@@ -5,54 +5,116 @@ Subsurface uses quite a few open source libraries and frameworks to do its
 job. The most important ones include libdivecomputer, Qt, libxml2, libxslt,
 libsqlite3, libzip, libgrantlee5 and libgit2.
 
-Below are instructions for building Subsurface under some popular
-Linux distributions, for building Subsurface using Homebrew on a Mac,
-and for cross-building Subsurface for Windows. The lack of a working
-package management system for Windows makes it really painful to build
-Subsurface natively under Windows, so we don't support that at all.
+Below are instructions for building Subsurface
+- on some popular Linux distributions,
+- MacOSX,
+- Windows (cross-building)
+- Android (cross-building)
+- iOS (cross-building)
 
-All of the prebuilt binaries that we provide (right now Windows, Mac,
-Ubuntu/Debian/LinuxMint, and openSUSE/Fedora) are built using our own
-custom "flavors" of libdivecomputer. You can get these from
 
-https://github.com/Subsurface-divelog/libdc.git
+Getting Subsurface source
+-------------------------
 
-If you have cloned subsurface it is easier,
-locate yourself in the subsurface repo, and do:
-git submoule init
-git submodule update
+You can get the sources to the latest development version from our git
+repository http://github.com/Subsurface-divelog/subsurface.git
 
-Those branches won't have a pretty history and will include ugly merges,
-but they should always allow a fast forward pull that tracks what we believe
-developers should build against.
+Remark, if you plan on making pull requests, the best way is to clone
+subsurface.git to your GitHub account, make a new branch (e.g. myPatches).
+Then clone that to your local computer. Doing this makes it very simple
+to generate pull requests.
 
-In contrast to that both repositories also have Subsurface-clean branches.
-These should allow distros to see which patches we have applied on top of
+
+Getting our flavor of libdivecomputer
+-------------------------------------
+
+All our prebuilt binaries (see https://subsurface-divelog.org/download)
+are built using our own custom "flavor" of libdivecomputer
+see https://github.com/Subsurface-divelog/libdc.git
+
+In order to get the modified sources, do
+- locate yourself in the subsurface repo on your local computer
+- run "git submoule init"
+- run "git submodule update"
+
+The branches won't have a pretty history and will include ugly merges,
+but they should always allow a fast forward pull that tracks what we
+believe developers should build against. All our patches are contained
+in the "Subsurface-branch" which are also HEAD.
+
+This should allow distros to see which patches we have applied on top of
 upstream. They will receive force pushes as we rebase to newer versions of
 upstream so they are not ideal for ongoing development (but they are of
 course easy to use for distributions as they always build "from scratch",
 anyway).
 
-The rationale for this is that we have no intention of forking either of
-these two projects. We simply are adding a few patches on top of their
-latest versions and want to do so in a manner that is both easy for our
+The rationale for this is that we have no intention of forking the
+project. We simply are adding a few patches on top of their latest
+version and want to do so in a manner that is both easy for our
 developers who try to keep them updated frequently, and anyone packaging
 Subsurface or trying to understand what we have done relative to their
 respective upstreams.
 
-At this point Qt5 is required, Qt5.4 or newer is recommended and on the
-Mac, in order to get native Bluetooth support, Qt5.5 is necessary.
 
-Similarly, in order for our cloud storage to be fully functional you need
+Getting Qt5
+-----------
+
+We use Qt5 in order to only maintain one UI across platforms. Qt5.8 is the
+oldest version supported.
+
+Download the Open Source version from https://www.qt.io/download and 
+follow the Qt instructions to install it or alternatively follow the
+instruction specific to a distribution (see build instructions).
+
+To save time and disk space you can unselect Android and IOS packages
+(Of course unless you want to build Android/iOS versions) as well as
+- Qt Data Visualisation
+- Qt Purchasing
+- Qt Virtual Keyboard",
+- Qt WebEngine
+- Qt Network Authorization
+- Qt Remote Objects
+- Qt WebGL Streaming
+- Qt Script
+- Qt 3D2
+- Qt Canvas 3D
+- Qt Extras
+This can be done later by use of MaintenanceTool.app in your Qt folder.
+
+Remark: most Qt installations does not make Qt available on the command
+line, which is needed. In order to make Qt available you need to 
+add a symbol link (do something like
+ln -s <Qt location>/<version>/<type>/bin/qmake /usr/local/bin/qmake
+
+QtWebKit is needed, but no longer part of Qt5, so you need to download it and compile.
+We want to replace QtWebKit, but at the moment the print process depends on functions
+only available in QtWebKit
+
+
+Other third party library dependencies
+--------------------------------------
+
+In order for our cloud storage to be fully functional you need
 libgit2 0.23 or newer.
 
-Finally, as of Subsurface 4.5 we have switched our build system to cmake.
-qmake based builds are no longer supported.
+
+cmake build system
+------------------
+
+As of Subsurface 4.5 we have switched our build system to cmake.
+qmake based builds are no longer supported. Remark even though
+no longer supported, qmake is still used by googlemaps and the iOS build.
+
+Download from https://cmake.org/download and follow the instructions
+to install it or alternatively follow the instruction specific to a
+distribution (see build instructions).
+
+
 
 Build options for Subsurface
 ----------------------------
 
-The following options are recognized when passed to cmake:
+The following options are recognised when passed to make:
 
  -DCMAKE_BUILD_TYPE=Release	create a release build
  -DCMAKE_BUILD_TYPE=Debug	create a debug build
@@ -62,7 +124,7 @@ verbose mode by calling
 
  make VERBOSE=1
 
-Many more variables are support, the easiest way to interact with them is
+Many more variables are supported, the easiest way to interact with them is
 to call
 
  ccmake .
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -5,54 +5,116 @@ Subsurface uses quite a few open source libraries and frameworks to do its
...
-git submoule init
-git submodule update
+You can get the sources to the latest development version from our git
+repository http://github.com/Subsurface-divelog/subsurface.git
 
-Those branches won't have a pretty history and will include ugly merges,
-but they should always allow a fast forward pull that tracks what we believe
-developers should build against.
+Remark, if you plan on making pull requests, the best way is to clone
+subsurface.git to your GitHub account, make a new branch (e.g. myPatches).
```
atdotde: I don't think this is the best way. At least I have subsurface-divelog/subsurface set as origin for my local repositories. Here I have a branch master which tracks origin/master. For coding I git checkout -b DESCRIPTIVE_TITLE . When done, I have my githup fork configured as a remote to which I push my branch as soon as it's ready for viewing (and finally create a PR). That way, I don't need to track origin/master remotely at GitHub.

In short, it's fine to clone the central repository, to create patches you'll need a fork, but that can be done later.

INSTALL
```diff
@@ -5,54 +5,116 @@ Subsurface uses quite a few open source libraries and frameworks to do its
...
+- Qt Canvas 3D
+- Qt Extras
+This can be done later by use of MaintenanceTool.app in your Qt folder.
+
+Remark: most Qt installations does not make Qt available on the command
+line, which is needed. In order to make Qt available you need to 
+add a symbol link (do something like
+ln -s <Qt location>/<version>/<type>/bin/qmake /usr/local/bin/qmake
+
+QtWebKit is needed, but no longer part of Qt5, so you need to download it and compile.
```
atdotde: I wouldn't say it that way. For my local builds, for example, I don't need printing. So I have that disabled (via ccmake) and instead build using WebEngine. This allows me to use off the shelf Qt at the price of not being able to print (I do my printing via the TeX export). But I guess, that's up to everybody's preferences.

INSTALL
```diff
@@ -5,54 +5,116 @@ Subsurface uses quite a few open source libraries and frameworks to do its
...
+Download from https://cmake.org/download and follow the instructions
+to install it or alternatively follow the instruction specific to a
+distribution (see build instructions).
+
+
 
 Build options for Subsurface
 ----------------------------
 
-The following options are recognized when passed to cmake:
```
atdotde, line 55: Are you sure about this?

INSTALL
```diff
@@ -163,64 +225,64 @@ You can run the following command:
...
-1)  Install Homebrew
 
-$ ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+0) You need to have XCode installed. The first time (and possibly after updating OSX)
 
-2) Install needed dependencies
+0.1) run "xcode-select --install"
 
-$ brew install asciidoc libzip cmake libusb pkg-config automake libtool hidapi curl libssh2
+1) You also need to install pkg-config
```
atdotde: I am convinced that our main recommendation should be to use home-brew. Because it works. When I have a new Mac, getting it to build subsurface from scratch, this is really the fastest and easiest way. 

You may have your reasons not to use it. But I would consider this a special case. It's good to somewhere document how to do it. But people should not be encouraged to do the manual install unless they know what they are doing. Plus, using home-brew, it's also much easier to stay up to date.

INSTALL
```diff
@@ -163,64 +225,64 @@ You can run the following command:
...
-bash subsurface/scripts/build.sh
+4.1) cd <repo>/..
+
+4.2) run "bash <repo>/scripts/build.sh -build-deps -desktop
 
 After the above is done, Subsurface.app will be available in the
 subsurface/build directory. You can run Subsurface with the command
 
-$ open subsurface/build/Subsurface.app
+4) open subsurface/build/Subsurface.app
```
atdotde: The disadvantage of running it like this is that you won't see diagnostic messages in the terminal.

To get those you directly run

subsurface/build/Subsurface.app/Contents/MacOS/Subsurface

(the TAB key is your friend here).

Maybe we should also mention that using Qt Creator (with our style settings from the coding style document) is the suggested way for development (thanks to commend completion, following symbols around and having a GUI for the debugger).

INSTALL
```diff
@@ -241,9 +303,14 @@ This is NOT RECOMMENDED. To the best of our knowledge there is one single
 person who regularly does this. The Subsurface team does not provide support
 for Windows binary build from sources natively under Windows...
 
+The lack of a working package management system for Windows makes it
+really painful to buildSubsurface natively under Windows,
+so we don't support that at all.
```
atdotde: I am not a Windows person, but my understanding was that now several people are successfully building using MinGW.

atdotde (changes requested):

INSTALL
janiversen: Ok, I will update the text, to make it clear you can do a normal “git clone”. For pull requests I think we say  early the same, let me be sure:
- you have cloned subsurface to you github account
- you have cloned subsurface to your local disk
- you create a local branch, work, and when finished push upstream to your github form
- you make a pull request, on github as usual

did I get that right?

INSTALL
janiversen: good point, will update

INSTALL
janiversen: no idea, it was in the existing version.

atdotde: but you changed cmake to make.

INSTALL
atdotde: My point was that I clone my local repository from github/subsurface-divelog (which thus becomes "origin") and not from GitHub/atdotde (which is just a remote like many other GitHub/username which I have to be able to check out branches of other developers). This allows me to easily track origin/master (on which my PR's should be based or rebased).

INSTALL
janiversen: I politely disagree, homebrew is known to really polute your mac and cause problems if you e.g. do not want to use the newest verdion of a library (I was told that is the reason why build.sh has a -builddeps option)

It is fair to write, that it can be done with homebrew, and giving the new developer the choise.

INSTALL
janiversen: Actually that is the old text, but I will update with your explanation.

INSTALL
janiversen: new version ready for review

INSTALL
```diff
@@ -17,13 +17,23 @@ Getting Subsurface source
...
-
+repository:
+    git clone http://github.com/Subsurface-divelog/subsurface.git
+
+You keep it updated by doing:
+    git checkout master
+    git pull -r
+    
+If you plan on making pull requests, you need furthermore to:
+   - clone subsurface.git to your GitHub account
```
dirkhh: the wording that GitHub uses is "fork the subsurface repository from your GitHub account" - we should match that.

INSTALL
```diff
@@ -17,13 +17,23 @@ Getting Subsurface source
...
+   - make a new local work branch, name it after the patch like e.g. myFirstPatch
```
dirkhh: That's actually optional. It's highly recommended, but not required. You could work in your master and create the pull request from there. So maybe line 27 above should read "Here are some best practices for those planning to contribute and create pull requests for Subsurface:"
And furthermore - why is this section in the INSTALL file? Isn't this part of CONTRIBUTING.md?
So I'd suggest removing 27-36 from INSTALL

INSTALL
```diff
@@ -86,9 +96,11 @@ line, which is needed. In order to make Qt available you need to
 add a symbol link (do something like
 ln -s <Qt location>/<version>/<type>/bin/qmake /usr/local/bin/qmake
 
-QtWebKit is needed, but no longer part of Qt5, so you need to download it and compile.
-We want to replace QtWebKit, but at the moment the print process depends on functions
-only available in QtWebKit
+QtWebKit is needed, if you want to print, but no longer part of Qt5,
+so you need to download it and compile. We want to replace QtWebKit,
```
dirkhh: Actually, this is no longer true. So let's remove the sentence starting with "We want to replace it..."

INSTALL
```diff
@@ -86,9 +96,11 @@ line, which is needed. In order to make Qt available you need to
...
+but at the moment the print process depends on functions only available
+in QtWebKit. In case you just want to test without print possibility
+omit this step.
```
dirkhh: What would make sense is to add a few lines about how to install QtWebKit from source. Possibly in its own commit.

INSTALL
```diff
@@ -228,16 +240,22 @@ sudo ldconfig ~/src/install-root/lib
 Building Subsurface under MacOSX
 --------------------------------
 
-Simple version:
-
-brew install libzip cmake libusb pkg-config automake libtool hidapi curl libssh2
-
+You have 2 options for the first time install, either using homebrew or manual,
```
dirkhh: either using homebrew or by manually building the dependencies from source

dirkhh (changes requested):
Good edits, but some comments - and especially the one regarding INSTALL vs CONTRIBUTING.md I'd like you to fix, please

dirkhh:
As for building ON Windows (i.e., not cross building): the only person who ever did that was @neolit123  and I think even he by now cross builds (not 100% sure). It's not something we will ever support as the process is indeed insanely painful, whereas the cross build from Linux is fairly straight forward (even though the first time is quite time consuming)

INSTALL
dirkhh: This is an interesting argument. For the longest time I would have agreed with Robert. But at this point I'll admit that I have spent so much time to make sure that the -build-deps option works flawlessly, that I'm ok with giving this equal billing.
BTW, the reason for that option is NOT that you might want an older version of a library (though that's ALSO a potential problem), the real problem FOR ME (as someone building distribution binaries) is that Homebrew always compiles everything for your current macOS version. The developers quite explicitly and intentionally reject the idea that you might want to build for an older version of macOS. Their vision is that every individual user "simply" installs everything using Homebrew. Since that is not what I want our users to have to do and since I always want to support users back to the oldest version that Qt supports (which is just about to bump from 10.10 to 10.11, sadly), I cannot use Homebrew for my "production builds". Which is why I spend all that time ensuring that we can build all of our dependencies in the first place. And I think I'll add QtWebKit as an option there as well.

INSTALL
janiversen: Yes I planned on doing that in a separate commit.

janiversen:
Just updated due to the comments from dirkhh.

I am right now fighting (no less) with hidapi, because ./bootstrap fails. First it seems configure.ac does not like "libtool", secondly $host is set to "" in configure, leading to all kinds of disasters.

if somebody with a Mac, could run ./bootstrap in hidapi redirect the output and mail it to me, I will be grateful.

dirkhh:
I don't think this will really help you unless you have working autotools
```bash bootstrap
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
autoreconf: configure.ac: tracing
autoreconf: running: glibtoolize --copy --force
glibtoolize: putting auxiliary files in '.'.
glibtoolize: copying file './ltmain.sh'
glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
glibtoolize: copying file 'm4/libtool.m4'
glibtoolize: copying file 'm4/ltoptions.m4'
glibtoolize: copying file 'm4/ltsugar.m4'
glibtoolize: copying file 'm4/ltversion.m4'
glibtoolize: copying file 'm4/lt~obsolete.m4'
autoreconf: running: /usr/local/Cellar/autoconf/2.69/bin/autoconf --force
autoreconf: running: /usr/local/Cellar/autoconf/2.69/bin/autoheader --force
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:25: installing './compile'
configure.ac:22: installing './missing'
hidtest/Makefile.am: installing './depcomp'
autoreconf: Leaving directory `.'
```

INSTALL
janiversen: Sorry I should have written newest SDK, which was what I meant.

Adding QtWebKit as an option is a real good thing. I am playing with adding an option "-prerequites" that installs the tools needed.

janiversen:
Thanks a lot, it actually did help me, "autoreconf: running: glibtoolize --copy --force" I do have Autotools installed (as pr. INSTALL), but glibtoolize is missing.

neolit123:
> Are you using Cygwin ? I use that in another very big project, Autotools etc. work like a charm for me in a Cygwin shell. Furthermore Cygwin comes with bash, so I can run linux build scripts without any changes.

that's true. there are even cygwin Qt packages, which are probably quite out of date though.

i have an `msys` setup that comes with `msys-git`, so i do have a `sh` shell. yet i want to build Subsurface out side of the `cygwin` sandbox, with a fairly recent Qt and in that case the native tools do have issues.

INSTALL
```diff
@@ -5,54 +5,117 @@ Subsurface uses quite a few open source libraries and frameworks to do its
...
+To save time and disk space you can unselect Android and IOS packages
+(Of course unless you want to build Android/iOS versions) as well as
+- Qt Data Visualisation
+- Qt Purchasing
+- Qt Virtual Keyboard",
+- Qt WebEngine
+- Qt Network Authorization
+- Qt Remote Objects
+- Qt WebGL Streaming
+- Qt Script
```
dirkhh: IIRC the googlemaps plugin requires Qt Script. Or was it Grantley? I just recently did that an noticed that I had to install it in order to complete a build even though Qt lists it as deprecated.
Did it work for you without Qt Script?

INSTALL
```diff
@@ -5,54 +5,117 @@ Subsurface uses quite a few open source libraries and frameworks to do its
...
+- Qt 3D2
+- Qt Canvas 3D
+- Qt Extras
+This can be done later by use of MaintenanceTool.app in your Qt folder.
+
+Remark: most Qt installations does not make Qt available on the command
```
dirkhh: most Qt installations *do* not make...

INSTALL
```diff
@@ -5,54 +5,117 @@ Subsurface uses quite a few open source libraries and frameworks to do its
...
+line, which is needed. In order to make Qt available you need to 
+add a symbol link (do something like
+ln -s <Qt location>/<version>/<type>/bin/qmake /usr/local/bin/qmake
```
dirkhh: Actually, I think it's preferred to do
```
PATH=$PATH:<Qt location>/<version>/<type>/bin
```
what do you think?

INSTALL
```diff
@@ -5,54 +5,117 @@ Subsurface uses quite a few open source libraries and frameworks to do its
...
 
-Finally, as of Subsurface 4.5 we have switched our build system to cmake.
-qmake based builds are no longer supported.
+
+cmake build system
+------------------
+
+As of Subsurface 4.5 we have switched our build system to cmake.
+qmake based builds are no longer supported. Remark even though
+no longer supported, qmake is still used by googlemaps and the iOS build.
```
dirkhh: Historically this made sense. I think I'd now would phrase this differently (for people like you, coming new to the project):

Our main build system is based on cmake. But qmake is needed for the googlemaps
plugin and the iOS build.

dirkhh (changes requested):
Sorry, I hadn't realized that you pushed additional commits instead of combining them into one. I'll do that as I apply the changes.
So this review contains a couple more small suggestions. My apologies for the two-time hassle.

INSTALL
janiversen: I can link but it will not run, so I am back to the tool chain right now. line 79 removed in next update.

Sorry for doing multiple commits, but that makes it easier for me, to track the review comments, feel free to squash them, that is only logical.

INSTALL
janiversen: Heh, I guess I am just a burned child. I try to avoid adding to PATH, since a while ago, where I had severe problems, because osx have or had a PATH limitation of 1k, but did not report any errors. I agree PATH is the better suggestion, updated in next commit.

INSTALL
janiversen: Done, and pushed a new commit. I will try to work with --amend in the future, different projects different styles :-)

INSTALL
dirkhh: I tell people to add PATH in the shell in which they are building, not in the bashrc file. That's the beauty of doing it this way. It affects you where you build, but it doesn't leave artifacts behind in random places in the file system.
Worse, the next time you install a newer Qt (which goes in a different directory), the link will still point to the old qmake and things will not work as you expect. :-)

INSTALL
dirkhh: It's a fine line. And I know that some projects have very different views on this.
I want people to do multiple commits when making multiple changes. And I want these commits to stay in our repo - so all these projects that by default always squish together every PR... they are IMHO doing it wrong.
But at the same time, as you update things based on review feedback, those changes should be merged together.
So in the case that you have a PR with four commits, sometimes we'll ask you to update a commit in the middle and then do a force push so we have a clean, useful git history.
Makes sense?

INSTALL
janiversen: That is the way to go, I actually got tired of the conflicts, so I have a series of "buildProfile_<project>" which I simply source in the shell where I work.

It is actually amazing how conflicting some open source projects are, and how unwilling the projects are to solve this kind of problems.

INSTALL
janiversen: It makes a lot of sense, I am very careful to only have 1 change/problem in 1 commit, but of course the pull request, might contain a couple of those. For this document change it makes sense to squash them all.

Actually I am preparing some changes to build.sh just like that:
1) add some "rm" to do cleanup
2) add my new build option -prereq, that will check for missing tools

dirkhh (approved):

dirkhh:
@atdotde - I think all of your requests have been addressed as well - can you approve this PR?

atdotde (approved):

PR #1219, opened 2018-04-24
added missing cleanup of libzip....gz

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

I marked "Build system change" even though it is just a post cleanup, so nobody should be affected.

When doing build.sh -build-deps the downloaded libzip* file is not deleted.

It would be nice to allow parallel build (e.g. Mac build debug/release as well as iOS build debug/release) that way a change can be tested fast on all 4 versions. The standard answer just clone
4 times or use "git corktree", means you do not share your edited code file.

### Changes made:

1) rm libzip...


dirkhh:
But the question is... do I want to delete that tar.gz file?
libzip is the only one that we don't get from a git repo - having the "original" tar file available to be able to tell if any changes were made seems actually useful. So I'm not convinced that I want this.

janiversen:
you do have the version number, but I will not argue. I just found somethung interesting. Making a clean build with build.sh downloads at least googlemaps that look like another problem

PR #1221, opened 2018-04-27
Turn off optimization while debugging

atdotde:
With the optimizer eliminating local variables the debugger
becomes useless.

This reverses 0f3bda4c44

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

I found that with optimisation, the debugger is pretty much unusable since it loses access to local variables. Since debug builds should allow debugging, turned it off again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,9 +105,9 @@ endif()
 set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
 set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
 
-# optimize -O2 even for debug builds
-set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O2")
-set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O2")
+# Turn off optimization when debugging, let the debugger get a chance at local variables
+set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O0")
+set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
 
 # pkgconfig for required libraries
 find_package(PkgConfig)


neolit123:
@dirkhh @atdotde @bstoeger 

i guess @bstoeger 's observation is that `-O2` generates more readable assembly.

on the other what @atdotde reports:
> the debugger is pretty much unusable since it loses access to local variables

i think out weights the previous problem and the commit should be indeed reverted.

instead of hard coding the optimization level we should remove all `-Oxxxx` flags from CMake and leave it to the user to specify them via the command line.

bstoeger:
Perhaps the more important point was that some warnings depend on data-analysis that is only performed with -O2. Another point is that it seems like a good idea to test code that is as close as possible to release builds.

> the debugger is pretty much unusable since it loses access to local variables

This is only partially true (test program, compiled with -O3):
```
(gdb) next
7               for (int i = 0; i < 100; ++i)
(gdb) next
8                       printf("= %d =\n", i);
(gdb) next
= 2 =
7               for (int i = 0; i < 100; ++i)
(gdb) print i
$1 = 3
```
But yes, with SIMD-vectorization, copy-elision, inlining, etc I can see how stepping through an optimized program might become a nightmare.

Even though debuggers were never very central to my work-flow, I really would like to have builds with -O2 *and* debug-symbols for the odd use-case. Strategic `printf`s have always been more how I operate.

I think this should be discussed on the mailing list.

bstoeger:
PS: This is the original mail leading to the referenced PR:
http://lists.subsurface-divelog.org/pipermail/subsurface/2017-December/031248.html

neolit123:
i remember local variable working from gdb with optimization flags in other projects.
@atdotde can you provide an example of where this breaks for you in Subsurface?

atdotde:
I didn't do this out of aesthetic considerations. I was actually trying to find a bug (and found it, #1222 is the fix for it). I was single stepping though code in profile.c dive.c and planner.c in particular tissues_at_end() in planner.c where t0 and t1 didn't have useful values in the debugger and also entry in plot_data() in profile.c was optimised away (taking with it all interesting state variables). As I said, the debugger was useless.

We could make this a cmake variable that can be set using ccmake according to one preferences for looking at assembly, looking for data analysis warnings or actual debugging.

neolit123:
@atdotde 
> We could make this a cmake variable 

i was able to reproduce the `t1` being optimized out in `tissue_at_end`. `t0` was not optimized out for me, but i guess this still makes a valid point that using `-O2` all the time is not desired.

have a look at:
https://github.com/Subsurface-divelog/subsurface/pull/1224

following your suggestion, this PR adds the option to control the GCC optimization flags for debug builds.
by default we use `-O2` so if you want to disable that use `-DGCC_OPTIMIZATION_FLAGS:STRING=-O0` or set it via the GUI.

dirkhh:
The problem with having the Travis build be different from what we do for production builds should be obvious... I'd much rather make sure that we keep those consistent.

PR #1222, opened 2018-04-27
When looking for gas change events only use those

atdotde:
This function looks for the last gas change before a
given time. We should initialize it with a gaschange
event as we might later use this event to read a
gasmix from it.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This function traverses the linked list of events to find the last gas change before a given time. To this end, it looks ahead for the next gas change event following this and keeps this event (or NULL if there is none). Once it is called with a time after this next event, it returns the gas mix from this gas change.

This patch makes sure that it is initialled with a gaschange event (or NULL), so we can at any time read a gas mix from it (otherwise, we could eventually read a wrong event and interpret it as a gas change). I am surprised this bug did not show up earlier.

### DIFF
--- a/core/dive.h
+++ b/core/dive.h
@@ -960,7 +960,7 @@ static inline struct gasmix *get_gasmix(struct dive *dive, struct divecomputer *
 	if (!gasmix) {
 		int cyl = explicit_first_cylinder(dive, dc);
 		gasmix = &dive->cylinder[cyl].gasmix;
-		ev = dc ? dc->events : NULL;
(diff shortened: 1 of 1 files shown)


neolit123:
perhaps the commit title was cut?:
`When looking for gas change events only use those `

i would title it something in the lines of:
`get_gasmix: start with a 'gaschange' event`

neolit123 (approved):
i misunderstood the title. looks good.

PR #1225, opened 2018-04-28
DO-NOT-MERGE: debug android CI

neolit123:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

In android-build-wrapper.sh:
- Always create the folder 'licenses'
- Export 'SDKMANAGER_OPTS'

Fixes a recent failure of the Android CI build
where 'licenses/android-sdk-license' was no found
after the 'yes | ...' call.

i don't really know what's going on here. found a solution online and tried the CI multiple times until it worked.

related thread:
https://stackoverflow.com/a/47150411

### Mentions:

@glance- @dirkhh

### DIFF
--- a/packaging/android/android-build-wrapper.sh
+++ b/packaging/android/android-build-wrapper.sh
@@ -77,6 +77,9 @@ if [ ! -d $ANDROID_SDK ] ; then
 	mkdir $ANDROID_SDK
 	pushd $ANDROID_SDK
 	unzip -q ../$SDK_TOOLS
+	mkdir -p licenses
+	# use this fix from https://stackoverflow.com/a/47150411 or sdkmanager throws Java exceptions
+	export SDKMANAGER_OPTS='-XX:+IgnoreUnrecognizedVMOptions --add-modules java.se.ee'
 	yes | tools/bin/sdkmanager --licenses > /dev/null 2>&1 || echo "d56f5187479451eabf01fb78af6dfcb131a6481e" > licenses/android-sdk-license
 	cat licenses/android-sdk-license
 	echo ""
--- a/packaging/android/android-build-wrapper.sh
+++ b/packaging/android/android-build-wrapper.sh
@@ -23,7 +23,7 @@ USE_X=$(case $- in *x*) echo "-x" ;; esac)
 
 QT_VERSION=5.10
 LATEST_QT=5.10.1
-NDK_VERSION=r14b
+NDK_VERSION=r16b
 SDK_VERSION=3859397  # if you change this version, you'll likely have to change the android-sdk-license key fallback below
 
 ANDROID_NDK=android-ndk-${NDK_VERSION}
--- a/packaging/android/android-build-wrapper.sh
+++ b/packaging/android/android-build-wrapper.sh
@@ -46,7 +46,7 @@ fi
 
 # make sure we have the required commands installed
 MISSING=
-for i in git cmake autoconf libtool java ant wget unzip; do
+for i in git cmake autoconf libtool java ant wget unzip makedepend; do
 	command -v $i >/dev/null ||
 		if [ $i = libtool ] ; then
 			MISSING="${MISSING}libtool-bin "
--- a/packaging/android/build.sh
+++ b/packaging/android/build.sh
@@ -40,6 +40,7 @@ SUBSURFACE_DESKTOP=OFF
 ARCH=arm
 # Which SDK buildtools revision is used?
 ANDROID_BUILDTOOLS_REVISION=25.0.3
+ANDROID_API=16
 
 while [ "$#" -gt 0 ] ; do
 	case "$1" in
@@ -142,7 +143,7 @@ else
 fi
 
 if [ ! -e ndk-"$ARCH" ] ; then
-	"$ANDROID_NDK_ROOT/build/tools/make_standalone_toolchain.py" --arch="$ARCH" --install-dir=ndk-"$ARCH" --api=16
+	"$ANDROID_NDK_ROOT/build/tools/make_standalone_toolchain.py" --arch="$ARCH" --install-dir=ndk-"$ARCH" --api="$ANDROID_API"
 fi
 export BUILDROOT=$PWD
 export PATH=${BUILDROOT}/ndk-$ARCH/bin:$PATH
@@ -302,6 +303,7 @@ if [ ! -e "$PKG_CONFIG_LIBDIR/libgit2.pc" ] ; then
 	mkdir -p libgit2-build-"$ARCH"
 	pushd libgit2-build-"$ARCH"
 	cmake -DCMAKE_SYSTEM_NAME=Android -DSHA1_TYPE=builtin \
+		-DCMAKE_C_FLAGS=-D__ANDROID_API__="$ANDROID_API" \
 		-DBUILD_CLAR=OFF -DBUILD_SHARED_LIBS=OFF \
 		-DCMAKE_INSTALL_PREFIX="$PREFIX" \
 		-DCURL=OFF \
@@ -355,7 +357,7 @@ fi
 if [ ! -e "$PKG_CONFIG_LIBDIR/libftdi1.pc" ] && [ "$PLATFORM" != "Darwin" ] ; then
 	mkdir -p libftdi1-build-"$ARCH"
 	pushd libftdi1-build-"$ARCH"
-	cmake ../libftdi1-${LIBFTDI_VERSION} -DCMAKE_C_COMPILER="$CC" -DCMAKE_INSTALL_PREFIX="$PREFIX" -DCMAKE_PREFIX_PATH="$PREFIX" -DSTATICLIBS=ON -DPYTHON_BINDINGS=OFF -DDOCUMENTATION=OFF -DFTDIPP=OFF -DBUILD_TESTS=OFF -DEXAMPLES=OFF -DFTDI_EEPROM=OFF
+	cmake ../libftdi1-${LIBFTDI_VERSION} -CMAKE_C_FLAGS=-D__ANDROID_API__="$ANDROID_API" -DCMAKE_C_COMPILER="$CC" -DCMAKE_INSTALL_PREFIX="$PREFIX" -DCMAKE_PREFIX_PATH="$PREFIX" -DSTATICLIBS=ON -DPYTHON_BINDINGS=OFF -DDOCUMENTATION=OFF -DFTDIPP=OFF -DBUILD_TESTS=OFF -DEXAMPLES=OFF -DFTDI_EEPROM=OFF
 	make
 	make install
 	popd
@@ -427,6 +429,8 @@ fi
 
 PKGCONF=$(which pkg-config)
 cmake $MOBILE_CMAKE \
+	-DCMAKE_C_FLAGS=-D__ANDROID_API__="$ANDROID_API" \
+	-DCMAKE_CXX_FLAGS=-D__ANDROID_API__="$ANDROID_API" \
 	-DPKG_CONFIG_EXECUTABLE="$PKGCONF" \
 	-DQT_ANDROID_SDK_ROOT="$ANDROID_SDK_ROOT" \
 	-DQT_ANDROID_NDK_ROOT="$ANDROID_NDK_ROOT" \
--- a/packaging/android/build.sh
+++ b/packaging/android/build.sh
@@ -255,7 +255,8 @@ fi
 if [ ! -e "$PKG_CONFIG_LIBDIR/libzip.pc" ] ; then
 	mkdir -p libzip-build-"$ARCH"
 	pushd libzip-build-"$ARCH"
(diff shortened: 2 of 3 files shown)


packaging/android/android-build-wrapper.sh
```diff
@@ -77,6 +77,9 @@ if [ ! -d $ANDROID_SDK ] ; then
 	mkdir $ANDROID_SDK
 	pushd $ANDROID_SDK
 	unzip -q ../$SDK_TOOLS
+	mkdir -p licenses
+	# use this fix from https://stackoverflow.com/a/47150411 or sdkmanager throws Java exceptions
+	export SDKMANAGER_OPTS='-XX:+IgnoreUnrecognizedVMOptions --add-modules java.se.ee'
 	yes | tools/bin/sdkmanager --licenses > /dev/null 2>&1 || echo "d56f5187479451eabf01fb78af6dfcb131a6481e" > licenses/android-sdk-license
```
dirkhh, line 83: the weird thing is that this last line after the '||' already tries to manually force a license. I assume that key is no longer valid or something?
If your solution works more reliably, even better.

dirkhh:
Build failed on Android - but with weird errors:
```
/workspace/libgit2-0.26.0/src/diff_print.c:739: error: undefined reference to 'stdout'
/workspace/libgit2-0.26.0/src/unix/map.c:52: error: undefined reference to 'mmap64'
make[2]: *** [libsubsurface-mobile.so] Error 1
make[1]: *** [CMakeFiles/subsurface-mobile.dir/all] Error 2
```
Did we change SDKs or something while I wasn't paying attention?

neolit123:
but then it fails with:
```
[ 81%] Linking CXX shared library libsubsurface-mobile.so
/workspace/libgit2-0.26.0/src/diff_print.c:739: error: undefined reference to 'stdout'
/workspace/libgit2-0.26.0/src/unix/map.c:52: error: undefined reference to 'mmap64'
```

sorry, didn't see the above comment earlier:
> Did we change SDKs or something while I wasn't paying attention?

that's a good question. i don't think we did and the breakage started happening yesterday,.

packaging/android/android-build-wrapper.sh
neolit123: i think the `export` made the call before `||` work.

neolit123:
possibly relevant comment:
https://github.com/android-ndk/ndk/issues/543#issuecomment-336477138

`so if you're getting undefined references to stdin/stdout/stderr, you're not building for < 23 even if you think you are.`

neolit123:
so. i've tried to fix the `stdout` issue but then this happens:
```
/workspace/libgit2-0.26.0/src/hash/hash_openssl.h:13:25: fatal error: openssl/sha.h: No such file or directory
```

even with:
```
		-DOPENSSL_INCLUDE_DIR="$PREFIX"/include/openssl \
```

tried to fix that with explicitly passing `-I"$PREFIX"/include/openssl'` but it doesn't work.

also, i'm seeing:
```
warning: implicit declaration of function 'futimens'
```
which will probably fail at link time for libgit.

no idea what what changed in the CI but it appears to not be on our side and is a PITA to deal with.

dirkhh:
Yes, debugging these builds on Travis is indeed a PITA.
If we go back to an older build that passed and restart that, does it still pass?
I'll try that.
Also, @glance- is typically our experts when I run out of ideas, maybe he can take a look?

dirkhh:
This build passed a couple of days ago, I just restarted it: https://travis-ci.org/Subsurface-divelog/subsurface/jobs/372257556
If that now fails the change wasn't ours...

dirkhh:
And yes, it does fail now :-(

BTW, we are using 14b SDK, latest is 16b. I wonder if Google changed something in the older SDK (that would be kinda rude without changing version numbers, but not unheard of)

I'm not sure if I'll have the time today, but I might just switch us to the latest SDK and try to get that to work...

neolit123:
> BTW, we are using 14b SDK, latest is 16b. I wonder if Google changed something in the older SDK (that would be kinda rude without changing version numbers, but not unheard of)

yes, pretty bad considering the potential other users too.

looking at some of those `ifdefs` that hide symbols like `stdout` in the NDK, for me makes the NDK people bonkers. i see hundreds of reports about the same thing (missing `stdout`) and i can't possibly imagine a reason to not define one of these standard file descriptors even if they are a NOP.

i can understand something like `futimens()` (POSIX) not being in the toolchain, but thins like `stdout` or `rand()`...wow.

> I'm not sure if I'll have the time today, but I might just switch us to the latest SDK and try to get that to work...

i guess i can try this for a couple of hours today.

dirkhh:
It used to be that "something" didn't build with clang - but, err, I don't recall what it was 😇 
Thanks for working on this one, @neolit123

neolit123:
i've tried clang yesterday with the old NDK and it didn't work.
trying it right now with the new one.

a last resort seems to be to wrap the GCC binaries and always pass a certain definition to them.

dirkhh:
build error for zimcmp, but we of course don't need that. I'm sure libzip has a way to just build the library...

neolit123:
oddly i don't see `zipcmp` in my local libzip build, but i'm using cmake instead of configure.

neolit123:
i don't see `zipcmp` as an optional target from cmake:
https://github.com/nih-at/libzip/blob/master/CMakeLists.txt#L238
https://github.com/nih-at/libzip/blob/master/src/CMakeLists.txt#L8

seems to be always built.
i guess if libzip is the only blocker for `clang` this can be patched in libzip directly. but i don't really know if it's the only blocker.

trying the `gcc` wrappers right now.

neolit123:
so seems like `zipcmp` breaks for gcc too.
bummer...i will try to avoid the build of these and switch to cmake in the process.

dirkhh:
Typically the make files have a target to simply build the library and not the tools. Or sometimes it's a configure option. Let me check... grmbl - this is NOT well done. But at least in the cmake version it would be straight forward to sed-out the `ADD_SUBDIRECTORY(src)` line (which causes us to build the applications)

neolit123:
they always build the tools, indeed.
`configure` seems to fail to catch the missing fts_* functions.
libzip has a define for that - `HAVE_FTS_H` but it's set to true.

could be the NDK having the header but not having the functions in there.

trying with cmake right now.

packaging/android/build.sh
```diff
@@ -269,7 +269,8 @@ fi
 if [ ! -e "$PKG_CONFIG_LIBDIR/libzip.pc" ] ; then
 	mkdir -p libzip-build-"$ARCH"
 	pushd libzip-build-"$ARCH"
-	../libzip-${LIBZIP_VERSION}/configure --host=${BUILDCHAIN} --prefix="$PREFIX" --enable-static --disable-shared
+	# ../libzip-${LIBZIP_VERSION}/configure --host=${BUILDCHAIN} --prefix="$PREFIX" --enable-static --disable-shared
+	cmake ..
```
dirkhh: if this still fails, simply use sed (or something) to comment out the `ADD_SUBDIRECTORY(src)` line :-)

neolit123:
for some very odd reason `openssl` is reporting:
```
*** Because of configuration changes, you MUST do the following before
*** building:
	make depend
+ make depend
making depend in crypto...
make[1]: Entering directory `/workspace/openssl-build-arm/crypto'
../util/domd: 33: ../util/domd: makedepend: not found
mv: cannot stat 'Makefile.new': No such file or directory
make[1]: *** [local_depend] Error 127
make[1]: Leaving directory `/workspace/openssl-build-arm/crypto'
make: *** [depend] Error 1
```

when building with clang and doesn't report the same error when building with gcc.
i'm confused.

will now try to install `makedepend` from the `xutils-dev` package via `sudo` but i think `sudo` is not enabled for this VM config. tried adding the `xutils-dev` in the .yml but i've added it to the linux config by mistake. i wonder if adding it to the android config will make a difference.

neolit123:
i've added this to the yaml for `android`:
```
sudo: required
...
      addons:
        apt:
          packages:
            - xutils-dev
```

> Can you do an ls -l /usr/bin/makedepend?

it's not present even after the yaml additions.

then i tried a:
```
if [ ! -e /usr/bin/makedepend ] ; then
	sudo apt-get install xutils-dev
fi
ls -l /usr/bin/makedepend
```

and it says:
```
sudo: command not found
```

as if `sudo: required` doesn't work.
ending this for now. will see if i can find some time tomorrow (today).

glance-:
```
      addons:                                                                   
        apt:                                                                    
          packages:                                                             
```
doesn't work there for the android builds. They are happening inside a container to get a sane build enviorment. You need to add it in scripts/android/before_install.sh instead.

Yes, its sort of odd, but it was the least 4kery to get it to work.

If the whole debacle isn't fixed by someone else I'll take a stab at it tonight.

neolit123:
got `makedepend` working.

but had this error:
`OpenSSL : unknown argument: '-mandroid'`

found this comment online:
`OpenSSL doesn't support compilation with clang (on Android) yet. You'll have to use GCC.`

well this sucks...but before i switch back to GCC i'm going to try to hack the makefiles to make them work for clang.

neolit123:
the latest error is:

```
arm-linux-androideabi-clang -I.. -I../.. -I../modes -I../asn1 -I../evp -I../../include  -fPIC -DOPENSSL_PIC -DOPENSSL_THREADS -D_REENTRANT -DDSO_DLFCN -DHAVE_DLFCN_H -march=armv7-a  -I/include -B/lib -O3 -fomit-frame-pointer -Wall -DOPENSSL_BN_ASM_MONT -DOPENSSL_BN_ASM_GF2m -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM -DAES_ASM -DBSAES_ASM -DGHASH_ASM -c --sysroot=/workspace/ndk-arm/sysroot  -c -o sha256-armv4.o sha256-armv4.S

sha256-armv4.S:2638:2: error: invalid instruction

 adrl r3,K256

```

i will remove some of the `DSHA256_ASM` to see if it helps.

neolit123:
alright, i'm going to have to put this to rest for today.
can probably go back to it this week. not sure when.

https://api.travis-ci.org/v3/job/373075755/log.txt

**the good:**
openssl now builds with clang

**the bad:**
libgit2 still has errors.

```

/workspace/libgit2-0.26.0/src/unix/posix.h: In function 'p_futimes':

/workspace/libgit2-0.26.0/src/unix/posix.h:82:2: warning: implicit declaration of function 'futimens' [-Wimplicit-function-declaration]

  return futimens(f, s);
```

the `posix.h` header needs `futimens` which seems to be missing for both GCC and Clang. this will surely fail and link time. i don't see means in libgit2 to not use this header or to tell it that the symbol is not available. maybe yet another inline hack is needed here. the NDK toolchains  have weird ways to define things like `stdout` so assuming that they are POSIX compliant is quite bold.

```
/workspace/libgit2-0.26.0/src/hash/hash_openssl.h: At top level:

/workspace/libgit2-0.26.0/src/hash/hash_openssl.h:13:25: fatal error: openssl/sha.h: No such file or directory

 #include <openssl/sha.h>
```

this one is weird given the previous action in the `.sh` is to `make install` openssl. but it appears that either pkg-config isn't working or something is messed up with the includes.

note: i was seeing these error messages with the old NDK and with GCC so it's not a Clang only issue.

it seems as if every other library has build issues. the `build.sh` is quite hacky and would be quite impressive if we get the APK to actually run.

dirkhh:
Given that everything was building fine with the OLD 14b until just a couple of days ago - I'm tempted to just upload that old 14b from my local copy. Not the right long term solution, but this is just a nightmare.

@glance- any insights from you would be highly appreciated.

dirkhh:
ok, looks like @glance- doesn't have the time to look into this right now (?)
I think I'll just say "SCREW THIS" and upload an older version of the 14b SDK that still should work - hopefully I'll get to this later today or tomorrow... gotta run right now

neolit123:
if the old archive works, great. :+1: 

debugging the CI in the case of such a bad toolchain is very time consuming and i wish there was a way to "pause" a Travis VM and SSH into it or something in those lines.

both the above issues seem solvable with extra defines and includes but i don't know what else will break next in the script. could give it another try next week.

dirkhh:
Took a step back and compared again what's different between a working build and a failing build.
And that made the ACTUAL problem obvious. No, Google did NOT change the old 14b repo without telling anyone. That would, frankly, be a pretty insane thing to do for them.
Instead our Travis build used Ubuntu:latest - and two weeks ago that switched to Bionic which breaks the build. Forcing this to be Xenial (16.04) made it all work again.
Closing this (sure, eventually we'll want to figure out how to build on Bionic; maybe).

neolit123:
thanks for finding the actual cause!

it's interesting how Bionic broke this that bad...
i will keep this branch if we intend a switch in the future.

PR #1224, opened 2018-04-28
cmake: add the GCC_OPTIMIZATION_FLAGS variable

neolit123:
### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Instead of always adding -O2 for CMAKE_C_FLAGS_DEBUG and
CMAKE_CXX_FLAGS_DEBUG allow the user to pass a custom value
via GCC_OPTIMIZATION_FLAGS.

Passing -DGCC_OPTIMIZATION_FLAGS:STRING=-O0 would disable
all optimizations.

### Changes made:

- the default value is still `-O2`.
- adding `-DGCC_OPTIMIZATION_FLAGS:STRING=-O0` from the command line or using the GUI would disable the optimizations.

### Related issues:

This PR should replace PR:
https://github.com/Subsurface-divelog/subsurface/pull/1221

### Mentions:

@atdotde @bstoeger @dirkhh

PR #1227, opened 2018-04-29
cmake: add the GCC_OPTIMIZATION_FLAGS variable

atdotde:
Instead of always adding -O2 for CMAKE_C_FLAGS_DEBUG and
CMAKE_CXX_FLAGS_DEBUG allow the user to pass a custom value
via GCC_OPTIMIZATION_FLAGS.

Passing -DGCC_OPTIMIZATION_FLAGS:STRING=-O0 would disable
all optimizations.

This can be set in ccmake

Suggested-by: Robert C. Helling <helling@atdotde.de>
Signed-off-by: Lubomir I. Ivanov <neolit123@gmail.com>

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Allow for local configuration using cmake. I would do this rather than #1224

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,8 @@ option(LIBGRANTLEE_FROM_PKGCONFIG "use pkg-config to retrieve grantlee" OFF)
 option(FORCE_LIBSSH "force linking with libssh to workaround libgit2 bug" ON)
 option(LIBGIT2_DYNAMIC "search for libgit2.so before libgit2.a" OFF)
 
+option(GCC_OPTIMIZATION_FLAGS "optimization flags in debug builds" -O2)
+
 #Options regarding disabling parts of subsurface.
 option(NO_DOCS "disable the docs" OFF)
 option(NO_PRINTING "disable the printing support" OFF)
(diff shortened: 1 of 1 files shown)


neolit123:
hm, here is what i see when using:
```
set (GCC_OPTIMIZATION_FLAGS "-O2" CACHE STRING "GCC optimization flags")
```

oflags

are there benefits when adding this as an option instead:
```
option(GCC_OPTIMIZATION_FLAGS "optimization flags in debug builds" -O2)
```

BTW, you could have pushed on top of my branch in #1224 instead of making a new PR.

neolit123:
`option` accepts `ON` and `OFF` so its a `BOOL`.
https://cmake.org/cmake/help/v3.0/command/option.html

....but testing `:STRING` from and command line or by using `-O2` also works for `option`, which of course makes no sense (same goes for the whole `cmake` build system and it's weird semi-type safety and case-insensitive non-sense).

if `:STRING` is not passed `cmake-gui` is not happy and treats the variable as a `BOOL`.
i don't have the curses based `ccmake` which is also a popular frontend, to test that with either `option` or `set`.

if your mod works better for you then we should go with it, the only problem is that it doesn't follow the documentation where `option` is supposedly `BOOL`.

atdotde:
Stupid me.

PR #1229, opened 2018-05-02
List Ubuntu 18.04 packages needed for build.

merinsky:
Signed-off-by: Martin Měřinský <mermar@centrum.cz>

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x ] Documentation change
- [ ] Language translation

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -157,27 +157,27 @@ sudo zypper install git gcc-c++ make autoconf automake libtool cmake libzip-deve
 
 On Debian Stretch this seems to work
 
-sudo apt-get install git g++ make autoconf automake libtool cmake pkg-config \
-	libxml2-dev libxslt1-dev libzip-dev libsqlite3-dev \
-	libusb-1.0-0-dev libssl-dev \
-	qt5-default qt5-qmake qtchooser qttools5-dev-tools libqt5svg5-dev \
-	libqt5webkit5-dev libqt5qml5 libqt5quick5 libqt5declarative5 \
-	qtscript5-dev libssh2-1-dev libcurl4-openssl-dev qttools5-dev \
-	qtconnectivity5-dev qtlocation5-dev qtpositioning5-dev \
-	libcrypto++-dev libssl-dev qml-module-qtpositioning qml-module-qtlocation \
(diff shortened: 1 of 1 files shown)


neolit123 (approved):
i guess it's OK to have the specific packages for a specific distro version. any objections by others?

mturkia:
I would say that this can be combined with Debian stretch. Thus less maintenance burden for the package list. As far as I know Stretch and 18.04 are built on same base. Thus we only would need to add the following packages to stretch and update header for that.

qtdeclarative5-dev
qtdeclarative5-private-dev

dirkhh:
Yeah, I like to keep this as generic as possible
And since it seems reasonable to assume that people building from source will be on a fairly recent version of their distro of choice, we should simply make sure that the instructions are available for just that.

neolit123:
@merinsky
as per the 2 comments above, could you modify your commit?

neolit123 (changes requested):
unify debian / ubuntu.

merinsky:
I compared packages lists, and they are different.
* Debian needs libqt5declarative5
* Ubuntu needs qtdeclarative5-dev qtdeclarative5-private-dev

So I sorted packages for easier future comparison and removed Ubuntu 16.10. Documentation for last stable (LTS) should be good enough.

dirkhh:
I don't think that's the right direction to go.
People who want the latest packages can use the AppImage.
Distribution packages should - where possible - use the distro supplied versions.
We used to have to bring our own libgit2 because we relied on a feature that was missing in the version available on (at that time) any of the distros. But this should long be fixed - every distro that is still supported appears to have at least 0.24 which is new enough.
Qt 5.9.5 is definitely new enough for the desktop version. 5.7.1 isn't great for BLE, but should work otherwise.
Bringing newer Qt to a distro is really tricky - unless you do it the AppImage way.

jbygdell:
It's a somewhat philosophical question, what are our stand on the software packaging strategies?
I'm all for the AppImage approach.

merinsky:
Requested changes have been made, squashed and force pushed.

Issue #1213, opened 2018-04-22
Not possible to edit dive information from dive list.


janiversen:
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When viewing "dive all" it is possible to edit most of the fields for an uploaded dive, which is nice.
When viewing "dive list" and trying to edit a dive an error dialog box occurs, this is inconsistent behaviour.

### Operating system:

macOSX 10.13.4

Mac

### Subsurface version:

4.7.8

Official release

### Steps to reproduce:

1) select "view"/"all"
2) Edit e.g. buddy, it works
<img width="1762" alt="screen shot 2018-04-22 at 10 55 50" src="https://user-images.githubusercontent.com/6185391/39093234-5d1bffea-461c-11e8-9b1f-924dca129ad4.png">

3) select "view"/"dive list",
4) choose an uploaded dive (NOT manually entered)
5) select "log"/"edit dive"
6) you get an error dialog box
<img width="1720" alt="screen shot 2018-04-22 at 10 56 51" src="https://user-images.githubusercontent.com/6185391/39093255-9566922a-461c-11e8-948a-880e9a55fdd1.png">

### Current behavior:

Error dialog box

### Expected behavior:

Change to "all" screen"

dirkhh:
I see your concern, but I will file this under "Works as Designed". The dive list is designed to be read only.

janiversen:
No problem, but then "edit dive" should not be selectable in the menu :-)

dirkhh:
True - we should gray that out in this mode. Reopening the issue...

bstoeger:
I tried to enable "Edit dive" for all dives: https://github.com/bstoeger/subsurface/tree/edit and removed a few lines of code-duplication while at it.
The thing is ... this puts the application in edit mode, but apart from that is pretty much a no-op.

Therefore another idea: Say "Edit dive profile" instead of "Edit dive" and enable only if a single, planned or manually entered dive is selected?

janiversen:
Sent from my iPad

> On 3 May 2018, at 16:58, bstoeger <notifications@github.com> wrote:
> 
> I tried to enable "Edit dive" for all dives: https://github.com/bstoeger/subsurface/tree/edit and removed a few lines of code-duplication while at it.
> The thing is ... this puts the application in edit mode, but apart from that is pretty much a no-op.
> 
[...]

PR #1236, opened 2018-05-03
[RFC] Change behavior of "Edit dive"

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

See discussion in #1213. The whole state-thing is horrible spaghetti code that needs a cleanup.

### Changes made:

1) Make the "Edit dive" menu entry functional for all dives (also with downloaded profiles).
2) Switch to a special Edit-mode, which shows info and profile.

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -1096,25 +1096,7 @@ void MainWindow::on_actionAddDive_triggered()
 
 void MainWindow::on_actionEditDive_triggered()
 {
-	if (information()->isEditing() || DivePlannerPointsModel::instance()->currentMode() != DivePlannerPointsModel::NOTHING) {
-		QMessageBox::warning(this, tr("Warning"), tr("Please, first finish the current edition before trying to do another."));
-		return;
-	}
-
-	const bool isTripEdit = dive_list()->selectedTrips().count() >= 1;
-	if (!current_dive || isTripEdit || (current_dive->dc.model && strcmp(current_dive->dc.model, "manually added dive"))) {
-		QMessageBox::warning(this, tr("Warning"), tr("Trying to edit a dive that's not a manually added dive."));
-		return;
-	}
-
-	DivePlannerPointsModel::instance()->clear();
-	disableShortcuts();
-	DivePlannerPointsModel::instance()->setPlanMode(DivePlannerPointsModel::ADD);
-	graphics()->setAddState();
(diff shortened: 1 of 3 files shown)


desktop-widgets/mainwindow.cpp
```diff
@@ -1096,25 +1096,7 @@ void MainWindow::on_actionAddDive_triggered()
...
-	if (information()->isEditing() || DivePlannerPointsModel::instance()->currentMode() != DivePlannerPointsModel::NOTHING) {
-		QMessageBox::warning(this, tr("Warning"), tr("Please, first finish the current edition before trying to do another."));
-		return;
-	}
-
-	const bool isTripEdit = dive_list()->selectedTrips().count() >= 1;
-	if (!current_dive || isTripEdit || (current_dive->dc.model && strcmp(current_dive->dc.model, "manually added dive"))) {
-		QMessageBox::warning(this, tr("Warning"), tr("Trying to edit a dive that's not a manually added dive."));
-		return;
-	}
```
dirkhh, line 1108: Can you explain why you remove these checks?
Starting dive edit when we already edit a dive cannot possibly be a valid thing to do, right?
Equally, the current code checks for multiple trips being selected (granted, the warning is wrong), but is that a valid state to be in when starting an edit?

dirkhh (commented):
this seems like a good idea (I haven't compiled/run it, just looking at the code), but I worry about the removal of all these checks...

desktop-widgets/mainwindow.cpp
bstoeger: 1) The whole `isEditing` check is done in `editCurrentDive()`.
2) The `!current_dive` check is done in `editCurrentDive()`.
3) The `manually added dive` check is removed on purpose.
4) The `isTripEdit` check is removed, because `editCurrentDive()` also works with multiple selected dives. I'm not sure, what it actually does though (it says "editing multiple dives").

So this will certainly need some testing against unintended consequences. I was hoping for the issue reporter. :)

neolit123:
scary area to touch in the code base.

janiversen:
Build on top of my outstanding PR. Build/tested on macosx and iOS.

Works as I wanted, nice job !

dirkhh:
Since I didn't see any other feedback that indicate an issue here, I'll merge it.

Issue #1211, opened 2018-04-21
Editing duration and save, dive listbox is not updated


janiversen:
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

on Mac, when updating the duration and saving, the listbox below is not updated.

This issue is a bit more complex, update the duration to e.g. 0:50, and save, this is accepted. But when reselecting the dive, the old value appears again. If updating to 0:50:00 it works. The field edit is broken, you can get random numbers by selecting a number and replacing it.

### Operating system:

macOS

Mac

### Subsurface version:

4.7.8

Official release, download from HP

### Steps to reproduce:

1) view a dive
2) update duration of dive to e.g. 0:50 WITHOUT seconds
3) save
4) check listbox of dives, column Duration is not updated
5) reselect the dive, and the old value appear again.

### Current behavior:

See above

### Expected behavior:

Edit should accept 0:50 and set seconds to 0.

atdotde:
Sorry, I don't get it. How exactly do you change the duration of a dive? Neither the field in the dive list nor the one in the statistics tab should be editable. In fact, the duration is computed from the profile. Or do you change the profile (for example via "edit dive in planner"?

janiversen:
When you add a dive manually you need this field (it is editable in the iOS version as well, but in minutes, and that works) because date/time signals start of the dive, and there are no date/time to end the dive. For sure if you upload a log/profile it will be calculated. See the attached screenshot.
<img width="1762" alt="screen shot 2018-04-22 at 08 46 39" src="https://user-images.githubusercontent.com/6185391/39092251-05cc8b54-460a-11e8-9ff8-17c951ae86ea.png">

dirkhh:
This used to work - I'm trying it here locally with current master and it seems to work just fine. Unless you start your number with a '5'. Try entering 40 and you'll see it work. Then try 50 and it fails. WTF? OK, something to look into.

janiversen:
Confirmed, I would never have guessed that 5 is a "dirty" number :-)

I am happy to help, but would beed a code pointer

dirkhh:
desktop-widgets/tab-widgets/maintab.cpp on_duration_textChanged() calls parseDurationToSeconds(text) and my guess is that that's what's broken.
This code was written by me and therefore is likely to be completely bogus.
[...]

PR #1237, opened 2018-05-05
Planner: calculate last manually entered sample on demand

bstoeger:
Commit 9a8bab21c9996020b16f1dd140647efe8e262a42 introduced calculation
of the time of the last manually entered sample. The problem is
that the calculation was done in create_dive_from_plan(), but
samples could also be created in fake_dc(), which resulted in an
invalid last manually entered sample. This in return led to the
issue reported in #1211.

To fix this problem, and remove the complexity of keeping track
of a cached last manually entered sample, calculate this datum on
demand. There seems to be no point in keeping track of this
redundant piece of information.

This improves the issue reported in #1211 significantly, but does
not fix it completely, because the conversion duration->samples->duration
is not a no-op.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Description is in the commit-log. Improves #1211 significantly.

### Mentions:

@willemferguson, @neolit123, @atdotde, @mturkia: Please check / test. I have no idea what the original change was intended to do and therefore cannot test if the behavior is still OK.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -823,6 +823,18 @@ void fixup_dc_duration(struct divecomputer *dc)
 	}
 }
 
+duration_t last_manually_entered_sample(struct divecomputer *dc)
+{
+	duration_t zero_duration = { 0 };
+	if (!dc)
+		return zero_duration;
+	for (int i = dc->samples - 1; i >= 0; i--) {
+		if (dc->sample[i].manually_entered)
+			return dc->sample[i].time;
+	}
+	return zero_duration;
+}
+
 /* Which cylinders had gas used? */
 #define SOME_GAS 5000
 static unsigned int get_cylinder_used(struct dive *dive)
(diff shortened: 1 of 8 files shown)


dirkhh:
For everything planner I always look to @atdotde :-)

atdotde:
I am on it.

atdotde:
No, this is not a fix. The problem is that sample.entered is not preserved when saving to disk: Plan a dive (with some deco), save it to disk, quit subsurface and load that dive. Re-edit dive in planner will show you all deco stops as manually entered. Also VPM-B relies on this information (though in a more subtle way).

NACK

bstoeger:
Ok - then please change fake_dc() such that it correctly fills out this value. Ideally on top of #1239.

As it stands, this thing completely broke manually entering of dives. A show-stopper.

bstoeger:
Never mind - pushed an update tp #1239. Please check there. Still don't know what this is all about.

PR #1238, opened 2018-05-05
Planner: Unify final ascent rates in plan() and fake_dc()

bstoeger:
When generating fake profiles for manually entered dives, fake_dc() and
plan() used different final ascent rates of 5 m/min and 4.5 m/min,
respectively. This led to dives that were 6 seconds longer than entered
by the user and to confusion. See #554.

Therefore, use the same ascent rate taken from the preferences field
flag.ascratelast6m in both cases.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This (partially?) fixes the long-standing bug #554.
### Related issues:

#554 #1211

### DIFF
--- a/core/device.c
+++ b/core/device.c
@@ -146,7 +146,7 @@ void fake_dc(struct divecomputer *dc)
 	if (avg_d == 0) {
 		/* we try for a sane slope, but bow to the insanity of
 		 * the user supplied data */
-		fill_samples_no_avg(fake, max_d, max_t, MAX(2.0 * max_d / max_t, 5000.0 / 60));
+		fill_samples_no_avg(fake, max_d, max_t, MAX(2.0 * max_d / max_t, (double)prefs.ascratelast6m));
 		if (fake[3].time.seconds == 0) { // just a 4 point profile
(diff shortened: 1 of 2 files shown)


atdotde (commented):
I totally agree that these should be unified. But 5m/min sounds like a strange value to me. Is there anybody diving like this? 6m/min is 1m/10s which is much easier to hold looking at the computer.

bstoeger:
The whole logic seems broken to me. Depending on the value we set here, manually entered dives will change their duration. That seems unhealthy. Is the last sample not saved to disk or is it ignored?

bstoeger:
What about using `prefs.ascratelast6m`?

atdotde:
Let's use the pref.

bstoeger:
Using preferences now - tested with a few dive durations, always got the same value out that I put in. Of course, already saved dives stay 6-seconds longer.

atdotde (approved):
LGTM

PR #1239, opened 2018-05-06
Planner: don't return static data in fake_dc()

bstoeger:
fake_dc() used to return a statically allocated dc with statically
allocated samples. This is of course a questionable practice in
the light of multi-threading / resource ownership. Once these
problems were recognized, the parameter "alloc" was added. If set
to true, the function would still return a statically allocated
dc, but heap-allocated samples, which could then be copied in
a different dc.

All in all an ownership nightmare and a recipie for disaster.
The returned static dc was only used as a pointer to the samples
anyway. There are four callers of fake_dc() and they all have access
to a dc-structure without samples. Therefore, change the semantics
of fake_dc() to fill out the passed in dc. If the caller does
not care about the samples, it can simply reset the sample number
to zero after work.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a code cleanup, which gives the `fake_dc()` functions sane semantics (no static buffers, please!): The samples of the passed-in dc structure are filled out. If the caller wants to throw them away later, it is free to do so by setting `dc->samples` to 0. I was able to hit 3 of the 4 occurrences of `fake_dc()`. The fourth is in the mobile code, which I can't compile at the moment:

```
[ 56%] Linking CXX executable subsurface-mobile
/usr/lib/gcc/x86_64-linux-gnu/7/../../../x86_64-linux-gnu/Scrt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
CMakeFiles/subsurface-mobile.dir/build.make:86: recipe for target 'subsurface-mobile' failed
make[2]: *** [subsurface-mobile] Error 1
CMakeFiles/Makefile2:288: recipe for target 'CMakeFiles/subsurface-mobile.dir/all' failed
make[1]: *** [CMakeFiles/subsurface-mobile.dir/all] Error 2
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
```

In a second commit, `last_manual_time` is set properly, to fix the problem described in #1211.
### Changes made:

1) Don't return static-allocation data in fake_dc()
2) Correctly fill out `last_manual_time`

### DIFF
--- a/core/device.c
+++ b/core/device.c
@@ -133,6 +133,9 @@ void fake_dc(struct divecomputer *dc)
 		return;
 	}
 
+	/* Set last manually entered time to the total dive length */
+	dc->last_manual_time = dc->duration;
+
 	/*
 	 * We want to fake the profile so that the average
 	 * depth ends up correct. However, in the absence of
--- a/core/device.c
+++ b/core/device.c
@@ -109,33 +109,29 @@ static void fill_samples_no_avg(struct sample *s, int max_d, int max_t, double s
 	}
 }
 
-struct divecomputer *fake_dc(struct divecomputer *dc, bool alloc)
+void fake_dc(struct divecomputer *dc)
 {
-	static struct sample fake_samples[6];
-	static struct divecomputer fakedc;
-	struct sample *fake = fake_samples;
+	alloc_samples(dc, 6);
+	struct sample *fake = dc->sample;
 	int i;
 
-	fakedc = (*dc);
-	if (alloc)
-		fake = malloc(sizeof(fake_samples));
-
-	fakedc.sample = fake;
-	fakedc.samples = 6;
+	dc->samples = 6;
 
 	/* The dive has no samples, so create a few fake ones */
 	int max_t = dc->duration.seconds;
 	int max_d = dc->maxdepth.mm;
 	int avg_d = dc->meandepth.mm;
 
-	memset(fake, 0, sizeof(fake_samples));
+	memset(fake, 0, 6 * sizeof(struct sample));
 	fake[5].time.seconds = max_t;
 	for (i = 0; i < 6; i++) {
-		fake_samples[i].bearing.degrees = -1;
-		fake_samples[i].ndl.seconds = -1;
+		fake[i].bearing.degrees = -1;
+		fake[i].ndl.seconds = -1;
 	}
-	if (!max_t || !max_d)
-		return &fakedc;
+	if (!max_t || !max_d) {
+		dc->samples = 0;
+		return;
+	}
 
 	/*
 	 * We want to fake the profile so that the average
@@ -149,10 +145,10 @@ struct divecomputer *fake_dc(struct divecomputer *dc, bool alloc)
 		 * the user supplied data */
 		fill_samples_no_avg(fake, max_d, max_t, MAX(2.0 * max_d / max_t, 5000.0 / 60));
 		if (fake[3].time.seconds == 0) { // just a 4 point profile
-			fakedc.samples = 4;
+			dc->samples = 4;
 			fake[3].time.seconds = max_t;
 		}
-		return &fakedc;
+		return;
 	}
 	if (avg_d < max_d / 10 || avg_d >= max_d) {
 		avg_d = (max_d + 10000) / 3;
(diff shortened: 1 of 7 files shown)


dirkhh:
Hmm - I am able to build Subsurface-mobile on Arch Linux... which OS are you trying to build on?
main() is of course in subsurface-mobile-main.cpp :-)

janiversen:
> El 6 may 2018, a las 19:25, Dirk Hohndel <notifications@github.com> escribió:
> 
> Hmm - I am able to build Subsurface-mobile on Arch Linux... which OS are you trying to build on?
> main() is of course in subsurface-mobile-main.cpp :-)
> 
> 
macOSX High Sierra.

Rgds
Jan I

dirkhh:
@janiversen The question was of course for @bstoeger - but since you spoke up... are you able to build and run Subsurface-mobile on macOS ? IIRC it builds but it doesn't run on Mac right now... or was that magically fixed when I wasn't looking? :-)

janiversen:
Sorry I was reading your comments, and did forget that you had changed PR :-)

I build with scripts/build.sh followed by packaging/ios/build.sh, I will test scripts/build.sh -mobile in 2 seconds.

bstoeger:
Ok - got it to build by clearing everything and running cmake again.

Mobile still works. I noticed a bug: for manually added dives the profile is not immediately shown. You have to switch back and forth between dives. But master is likewise affected, so not a regression.

neolit123:
> This is a code cleanup, which gives the fake_dc() functions sane semantics (no static buffers, please!): The samples of the passed-in dc structure are filled out. If the caller wants to throw them away later, it is free 
to do so by setting dc->samples to 0. I was able to hit 3 of the 4 occurrences of fake_dc().

i think that the non-static change is fine, but `alloc_samples()` is kind of misleading with that `if` on top. i would at least add a comment on how to use it as it's now exposed in `dive.h` as a very "global" helper.

also we need to take things away from `dive.h` and not add more, but that's a major refactor...

> The fourth is in the mobile code, which I can't compile at the moment:

this error should not happen. you can try cleaning the build directory completely before running `cmake` for the mobile build.

neolit123:
> Ok - got it to build by clearing everything and running cmake again.

^ missed this comment.

bstoeger:
> i think that the non-static change is fine, but alloc_samples() is kind of misleading

What about `make_room_for_samples()` in analogy to `membuffer.h`'s `make_room()`?

> also we need to take things away from dive.h

By moving the `*_sample()` functions to `device.c`, we could make that function local to `device.c`. But I'm not sure about the `dive.c`/`device.c` division...?

Or we could not export it in the header file, but add an `extern`-declaration in `device.c`. Though I'm not a fan of such redundant declarations.

neolit123:
> What about make_room_for_samples() in analogy to membuffer.h's make_room()?

i think the name is fine, but perhaps a comment on top of the function should indicate how one should use it given the `if` can turn the function into a no-op. it's more of a helper instead of API, yet we have a lot of these in dive.h.

> By moving the *_sample() functions to device.c, we could make that function local to device.c. But I'm not sure about the dive.c/device.c division...?
Or we could not export it in the header file, but add an extern-declaration in device.c. Though I'm not a fan of such redundant declarations.

so dive.c/.h are quite old and they ended up accumulating a collection of utils that need to be broken down into different, properly named .c/.h pairs. dive.h also serves as a practically global header that exposes backend symbols everywhere. it decreases the compilation times a lot i think as all sorts of C++ final objects include dive.h instead of only the symbols of interest.

so that's a major refactor...for example, i think that things like `struct divecomputer *dc` should be moved to `device.h` and also all the related DC related helpers should be in the device.c/.h pair. that alone will cause a lot of breakage.

for now adding the helpers in dive.h is fine, but i'm just pointing out that we probably need to open a ML thread about this eventually and discuss what should go where as some folks might have objections about naming and locations.

neolit123:
> so dive.c/.h are quite old and they ended up accumulating a collection of utils

actually dive.c itself is more dive-centric, while dive.h is used as a include in a lot of places due to the helpers it exposes.

atdotde:
ACK

PR #1241, opened 2018-05-06
Updated iOS build/debug environment, to match desktop environment

janiversen:
Updated INSTALL to point at packaging/ios/README
Updated README to "facts"
Deleted ios_build_instructions as they are covered in README

Update build.sh in a first version, to allow build of iOS versions

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ x] Build system change
- [ x] Documentation change
- [ ] Language translation

Updated documentation and build.sh around iOS build/debug session
(see commit text).

The toolchain at least works now as documented

Remark this might cause a problem with travis until travis is updated. Old build.sh
expected Qt to be double installed (one in iOS dir and ~/Qt for the desktop version, as
pr doc.)

This is a preparation for multiple PR, covering:
- Combine <repo>/.. downloads with iOS downloads
- Move 3rd party build directories to <repo>/../<lib>
- Move Subsurface build directories to <repo>
- Decide on scripts/build.sh and build.sh (both build iOS, but different)
- Move/Clean directories/files in git tree not versioned (e.g. info.plist)

My preference is to make PR for each small step, allowing me to work, while
others do review.


packaging/ios/build.sh
```diff
@@ -157,8 +157,7 @@ echo next building for $ARCH
 	fi
 
 	if [ ! -e libxml2-${LIBXML2_VERSION}.tar.gz ] ; then
-		# wget ftp://xmlsoft.org/libxml2/libxml2-${LIBXML2_VERSION}.tar.gz
-		wget http://api.github.com/repos/GNOME/libxml2/tarball/v${LIBXML2_VERSION} -O libxml2-${LIBXML2_VERSION}.tar.gz
+		curl -O ftp://xmlsoft.org/libxml2/libxml2-${LIBXML2_VERSION}.tar.gz
```
dirkhh: I actually switched away from xmlsoft for this because their website was down for weeks - api.github.com occasionally times out, but has MUCH higher availability
But fundamentally it might make more sense to just check it out from git

packaging/ios/build.sh
```diff
@@ -179,16 +178,16 @@ echo next building for $ARCH
...
 		make install
+
 		popd
 	fi
 
-	if [ ! -e libxslt-${LIBXSLT_VERSION}.tar.gz ] ; then
-		# wget ftp://xmlsoft.org/libxml2/libxslt-${LIBXSLT_VERSION}.tar.gz
-		wget http://api.github.com/repos/GNOME/libxslt/tarball/v${LIBXSLT_VERSION} -O libxslt-${LIBXSLT_VERSION}.tar.gz
+	if [ ! -e libxslt-${LIBXSLT_VERSION}.tar.xz ] ; then
+		curl -O https://git.gnome.org/browse/libxslt/snapshot/libxslt-${LIBXSLT_VERSION}.tar.xz
```
dirkhh: git.gnome might work - do they have libxml as well?
Or... check it out from git as well

packaging/ios/build.sh
```diff
@@ -45,9 +45,9 @@ fi
 # Which versions are we building against?
 SQLITE_VERSION=3090200
 LIBXML2_VERSION=2.9.2
-LIBXSLT_VERSION=1.1.28
-LIBZIP_VERSION=0.11.2
-LIBGIT2_VERSION=0.26.0
+LIBXSLT_VERSION=1.1.32
+LIBZIP_VERSION=1.2.0
+LIBGIT2_VERSION=v0.26.0
```
dirkhh: as we are updating things - did you check if 0.27 would work?
Otherwise we might at least look at 0.26.3 to get the bug fixes

dirkhh (commented):
I'm basically fine with this. Actually, I think I will merge it as is and we can look at my comments for further improvement.
This is definitely a step in the right direction.

dirkhh:
Except, of course, that the build for iOS fails because it can't find qmake :-)
That might need fixing...

janiversen:
I actually had IOS_QT="~/Qt" which follows the instructions in INSTALL, but might not be where it is installed on Travis.

I would prefer the simple version of IOS_QT

wdyt ?

dirkhh:
I'm not sure what you mean by that comment, @janiversen 
The way I usually do this (and that's what is implemented for Travis) is that we have a link "Qt" in the top level directory that points to the installed version of Qt
If what you are saying is "can we hard code this instead of relying on running qmake" then I'm ok with that

janiversen:
I will implement the Qt link so travis is happy, and document it, so people can change it if they want. 

Hang on for a commit update.

I did run Subsurface with the current commit (in simulator), but I still miss to try and build with <repo>/scripts/build.sh -mobile (have not looked into the differences).

janiversen:
Updated commit to use "git clone", makes sense, however I am planning to reuse the downloads in <src> in order not to have things double.

packaging/ios/build.sh works as planned, and the generated app runs in simulator
scripts/build.sh -mobile generates an Info.plist that cannot be loaded, will investigate that in a minute.

neolit123 (changes requested):
added comment about bumping versions.

packaging/ios/build.sh
```diff
@@ -44,10 +44,10 @@ fi
...
 # Which versions are we building against?
 SQLITE_VERSION=3090200
-LIBXML2_VERSION=2.9.2
-LIBXSLT_VERSION=1.1.28
-LIBZIP_VERSION=0.11.2
-LIBGIT2_VERSION=0.26.0
+#LIBXML2_VERSION=2.9.2
+LIBXSLT_VERSION=1.1.32
+LIBZIP_VERSION=1.2.0
+LIBGIT2_VERSION=v0.27.0
```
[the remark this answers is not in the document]
janiversen: I totally agree to that, but I had to do it in 2 PR, because this one is about a change in ios/build.sh and just got enhanced with the libgit2 change.

The other PR is only the libgit2 change in scripts/build.sh

So when both PR's are merged we use libgit2 overall.

Sorry for making 2 PR, but I felt it would be wrong to add scripts/build.sh in this PR.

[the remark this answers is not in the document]
neolit123: i think that the build script and documentation changes should be separate commits unless they intersect - e.g. bound versioning.

i guess my other point is that when library versions are bumped for one OS they should be bumped everywhere - i.e. Windows, Linux and Android too.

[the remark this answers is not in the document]
dirkhh: I agree with that in most cases.
Clearly I'm at fault here because I drove @janiversen in that direction with an earlier comment.

dirkhh (commented):
Fundamentally I think that @janiversen is doing the right thing here. I'm ok if we bump just the iOS versions of various libraries in this commit and then update the others in a later commit -- I'm equally happy to pull the bump of versions out of this commit.

So far it seems that there are still build problems, so I'm waiting for @janiversen to address that - but I approve of the principle of this commit

janiversen:
I have just removed to version bump from this PR, will amend the other PR with bump of all platforms.

hope that satisfies all outstanding concerns ?

dirkhh:
> I have just removed to version bump from this PR, will amend the other PR with bump of all platforms.
>
> hope that satisfies all outstanding concerns ?

@janiversen I think you are simply witnessing us figuring out what we want to do on the fly :-)
I appreciate your work on this and I'm sorry I sent you down a path causing extra discussion with an earlier comment.

You said in earlier comments that there was a build issue. Is that addressed now?

janiversen:
hmm I do not have build problems anymore.

I run subsurface/scripts/build.sh -build-deps -both without any errors, however there are an info.Plist error, when I try to install the mobile version on a simulator

I run subsurface/packaging/ios/build.sh (without deleting the previous build, as I am trying to combine things), it builds without problems, and installs nicely (apart from a mountain of clang warnings).

janiversen:
Yes that is adressed, it was a struct that was not initialised with all variables (went away this night).

And no problem, that you think before doing, as long as we don't have bike-shedding :-)

I am slowly getting deeper into the iOS stuff (working in parallel on 3 branches + the PR branches).

dirkhh:
> Yes that is adressed, it was a struct that was not initialised with all variables (went away this night).

In most cases lately that is a sign that your Subsurface and libdivecomputer don't match (i.e., you haven't updated the submodule, or you haven't rebuilt it / installed it)

> And no problem, that you think before doing, as long as we don't have bike-shedding :-)
>
> I am slowly getting deeper into the iOS stuff (working in parallel on 3 branches + the PR branches).

I am excited to hear that. If you build problems are gone and the Mac and iOS builds on Travis succeed, we should merge this... I'll try to do a new TestFlight build today as well so more people can test.

packaging/ios/build.sh
```diff
@@ -242,69 +236,15 @@ echo next building for $ARCH
 			-DCMAKE_PREFIX_PATH=${PREFIX} \
 			-DCURL=OFF \
 			-DUSE_SSH=OFF \
-			../libgit2-${LIBGIT2_VERSION}/
-		sed -i.bak 's/C_FLAGS = /C_FLAGS = -Wno-nullability-completeness -Wno-expansion-to-defined /' CMakeFiles/git2.dir/flags.make
+			../libgit2/
+#TEMP		sed -i.bak 's/C_FLAGS = /C_FLAGS = -Wno-nullability-completeness -Wno-expansion-to-defined /' CMakeFiles/git2.dir/flags.make
```
dirkhh: Not sure why you did that, but it breaks the Travis build because without removing those flags the build log exceeds 4MB

dirkhh (changes requested):

janiversen:
sed is back in. I removed it, because build.sh played a trick on me :-(

dirkhh (approved):

dirkhh:
The build failure of qt55 should be a fluke of a stalled download - I just restarted it.

janiversen:
@neolit123 I have removed the version bump from this PR, and combined it as you suggested. Hope that solves your concerns ?

packaging/ios/build.sh
```diff
@@ -44,10 +44,10 @@ fi
...
 # Which versions are we building against?
 SQLITE_VERSION=3090200
-LIBXML2_VERSION=2.9.2
-LIBXSLT_VERSION=1.1.28
-LIBZIP_VERSION=0.11.2
-LIBGIT2_VERSION=0.26.0
+#LIBXML2_VERSION=2.9.2
+LIBXSLT_VERSION=1.1.32
+LIBZIP_VERSION=1.2.0
+LIBGIT2_VERSION=v0.26.0
```
neolit123: hm, i still see the version changes here.

> +#LIBXML2_VERSION=2.9.2

not using a specific version and cloning the master would mean that we build the latest HEAD.
is my assumption correct?
see [1].

>+LIBXSLT_VERSION=1.1.32
>+LIBZIP_VERSION=1.2.0

these are technically version bumps. did you have troubles building with the older libraries?
LIBZIP 1.2.0 is a nice bump everywhere for the sake of using a 1.x release (they started using proper versioning, perhaps after i suggested it to them).
not sure about the need to bump LIBXSLT.

> +LIBGIT2_VERSION=v0.26.0

this isn't a version bump but it kind of stands out because of the extra `v`. the tag checkout is correct.
not a big issue.

packaging/ios/build.sh
```diff
@@ -156,56 +156,47 @@ echo next building for $ARCH
...
-	if [ ! -e libxslt-${LIBXSLT_VERSION} ] ; then
-		mkdir -p libxslt-${LIBXSLT_VERSION}
-		tar -zxf libxslt-${LIBXSLT_VERSION}.tar.gz --strip 1 -C libxslt-${LIBXSLT_VERSION}
-		# libxslt have too old config.sub
-		cp libxml2-${LIBXML2_VERSION}/config.sub libxslt-${LIBXSLT_VERSION}
-		pushd libxslt-${LIBXSLT_VERSION}
-		autoreconf --install
-		popd
+	if [ ! -d libxslt ] ; then
+		git clone https://github.com/GNOME/libxslt.git libxslt
```
neolit123, line 201: [1]

janiversen:
Reinserted all old versions, to avoid technical version bump. Also removed "v" from libzip version, this should be done to scripts/build.sh as well.

Hope this satisfies all review requests @neolit123

neolit123:
@janiversen 

hi, i will leave it to @dirkhh to give the final approval.
the script makes libxlst and libxml be built from their master instead of a certain tagged version like before.
so this isn't technically true:
> Reinserted all old versions, to avoid technical version bump

if he is OK with that, your commit is good to get merged. :)
thanks and sorry for the back and forth.

neolit123 (approved):

dirkhh:
This is my fault. And I know that we have gotten to the point where this has become just frustrating for @janiversen - my apologies.
Unfortunately @neolit123 is correct.
@janiversen here's my proposal: Instead of continuously asking you to "bring us another rock" (I hope this idiom translates well), would you be ok if I took your commit, refactored it into a couple of commits (keeping your authorship) and get this to a stage that would work for everyone?
Please look at #1252

janiversen:
Hust gave #1252 my ok. No need for apologies, but I can confirm being quite frustrated, mainly over discussing something that does not really bring us forward. If people have strong logistical opinions they should suggest an exact change, like you have dine, and not ask me to do it in vague terms.

janiversen:
substituted

dirkhh:
> Hust gave #1252 my ok. No need for apologies, but I can confirm being quite frustrated, mainly over discussing something that does not really bring us forward. If people have strong logistical opinions they should suggest an exact change, like you have dine, and not ask me to do it in vague terms.

And that's what I meant by "bring me another rock".
I don't think anyone means ill. And I felt that this was indeed my fault because the argument was all about something you did based on a request of mine - and that wasn't really related to what you were actually trying to do.
I'm glad we found a way to move forward so we can focus on stuff that makes a difference instead.
Thanks

PR #1243, opened 2018-05-07
Bumped libgit2 version to 0.27.0

janiversen:
(done in another PR for iOS)

Tested on macosx and iOS (simulator)

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ x] Build system change
- [ ] Documentation change
- [ ] Language translation

changed libgit2 from 0.26.0 to 0.27.0 and tested (based on comment from dirkh


neolit123:
hi, is there a reason to only change libgit's version in the main `build.sh` and not in the packaging scripts?
a more descriptive commit message is perhaps needed here - i.e. why is this done (we had troubles in the past with libgit2).

janiversen:
I agree, but there is a reason, please see my comment in #1241

dirkhh:
Looking through https://github.com/libgit2/libgit2/releases makes me think that there are several new options in 0.27 that we'd be interested in - obviously this will need quite a bit of testing (e.g., we need to make sure that this doesn't break our certificate check hack on Windows).
What concerns me more right now is that this breaks the Travis build on Mac - but I haven't stared at the error long enough to understand what happened there.
Having this spread across several commits isn't really bothering me that much.
But I agree that once we decide to switch, we should do so for all platforms / build environments, so that means `packaging/android/build.sh` and `scripts/windows/before_install.sh`

dirkhh (changes requested):
see my previous comment

janiversen:
I could not find any VERSION lines in packaging/windows, seems they are hidden somewhere else ?

I assume scripts/build.sh is general for all linux flavours.

Please note, I have NOT tested on linux, windows or android.

janiversen:
Seems a windows maintainer is needed :-)

Travis fails on ios, due to a sed command which is removed in my other PR.

dirkhh:
We have a Windows maintainer - @neolit123 
Except that he's also the only person to build natively on Windows :-)
So I tend to maintain the cross build and @sfuchs79 helps with that

dirkhh:
> Travis fails on ios, due to a sed command which is removed in my other PR.

That's the wrong fix. Find out why the file moved and edit the build script. Without that sed command Travis will always fail as we create way more than 4MB of build log

janiversen:
Readded sed, thanks for the hint.

janiversen:
Seems this PR was really premature, independent of my good intentions. No need to keep it around, until it is decided to bump version.

And looking at the build.sh's it would be a better patch, to have a dependencies.sh that does curl/clone etc in one place. This script could then be called from the different build.sh and would save a lot of duplicated but slightly different lines.

Abandoning for now.

dirkhh:
> And looking at the build.sh's it would be a better patch, to have a dependencies.sh that does curl/clone etc in one place. This script could then be called from the different build.sh and would save a lot of duplicated but slightly different lines.

The way we do dependency management across the different build scripts is really weird and inconsistent. Having something more structured there sounds wonderful. One challenge is that we need a different set of dependencies, depending on the desktop/mobile and target OS, but this is definitely something that is ugly and ill maintained today.

janiversen:
Such a script is beung tested in my env. but depends heavely on what happens with my current proposed build.sh (which is documented and acknowledged to be the first of a series of changes)

dirkhh:
That sounds great! Looking forward to seeing that.

PR #1244, opened 2018-05-07
Remove 76 AppIcon, not used after iOS 10.

janiversen:
The AppIcon size 76 and 76@2x caused Xcode to flag a problem
this size is not used in iOS versions higher than 10.

Removed to get rid of flagging.

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Depending on the Xcode profile, this can be a warning or an error (on my Mac it is an error)

### Changes made:

1) Updated Contents.json
2) Remove AppIcon-76*


dirkhh:
Hmm - but since we support iOS back to version 8, don't we need that size? i.e., will the app only work on 11 if we drop that?

janiversen:
Do we really support that old iOS, Apple is not making maintence releases anymore (only 10 & 11). Apart from that I read 60% runs 11.x 30% 10.x and the rest older versions, so considering our ios user base we talk about a couple of users at most, big effort for few people. I am aware that the situation fir Android is a lot worse, here it makes sense to support older versions.

If we support ios 8, then we should compile link with that sdk, something I personally do not do, I have only sdk11 installed. Qt does not seem to be a problem.

the app works with or without the png ( if not present it scales a smaller), but Xcode is bound to give problems with our current setup, since it is considered an error.

dirkhh:
Full disclosure: I barely know what I'm doing creating the iOS app, so I simply set things up in my Xcode to be compatible back to iOS 8 (and never had a way to test if that even works).
So far, frankly, it has never seen like it was any "effort" - but that may once again just be a sign of my ignorance.
The one thing I don't want to start is to just abandon older versions for no good reason. On macOS we constantly have the problem that new Qt versions abandon older OS releases and we get quite a few complaints about that.

janiversen:
I am pretty sure, the same happens with Qt for iOS, if you are not careful to install sdk8, it detects the version and uses the widget present. 

It is confusing, but in Xcode you have 2 settings, one is which iOS versions you support and the second is which sdk you compile/link against. If you say support for ios8 and onwards and link with sdk11, you can use iOS 11 functions without seeing a warning until you run it on a physical device (simulator will adjust).

Btw. do you build with scripts/build.sh or packaging/ios/build.sh that are quite different and generate quite different code (first one, includes a ton of frameworks). I just looked at the build.log, and could not find the iPhone SDK ?

I have no problem killing this patch, but it is very normal in the iOS world only to support the 2 newest main versions (10 and 11 for now) as it follows Apple support. The iOS world is far more controlled, than any other OS I know of.

dirkhh:
I think you missed my point, or I made it poorly. I have no idea what I'm doing here and I'm trying to understand the implications of what you propose.
It sounded to me like removing this icon made Subsurface-mobile iOS 11 only.
Oh well.

janiversen:
I did understand that you are not an iOS person (but there are others with strong opinions it seems). I just wanted to get rid of a annoying error message which I have now done by having a local "mySetup.sh".

Sorry but I am a bit disappointed today with how other look at my work, and considering if I am on the right path. Looking at my PRs the discussions (not from you, with the exception of zip 0.27.0) are bigger than the patches themselves. Many of the comments are very relevant, but in total a lot of it looks like bike-shedding (e.g. can a version number include a "v", which btw is OK in scripts/build.sh). Anyhow this is quite another topic.

janiversen:
Lets have a chat, when you are in a relaxed situation..on IRC or hangout. But no need to rush it, I will not disappear overnight.

PR #1245, opened 2018-05-07
mobile: show selected dive on details view

mfbernardes:
Call positionViewAtIndex in order to make the selected dive
visible.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As reported by Scott Ireland on the mailing list.

> Selecting a dive from the dive list does not accurately display the details for that dive.  What I'm getting (starting from a fresh app start) is something like this:
>- Select any dive from the dive list (e.g. 391 below).   Dive details from the most recent dive (e.g. 396, not the selected one) are displayed
>- Select any other dive from the list.  The most recent dive is still displayed
>- Swipe the dive details left to move to the previous dive.  The second most recent dive is displayed (e.g. 395).
>- Select another dive from the dive list.  On iPad, the dive details scroll partially (see screenshot) but still show the same dive as before.  On iPhone, the dive details appear to move to the previous dive (e.g. 394), which still does not match the selected dive.

### DIFF
--- a/mobile-widgets/qml/DiveDetails.qml
+++ b/mobile-widgets/qml/DiveDetails.qml
@@ -209,7 +209,7 @@ Kirigami.Page {
 
 	function showDiveIndex(index) {
 		currentIndex = index;
-		//diveDetailsListView.positionViewAtIndex(index, ListView.End);
+		diveDetailsListView.positionViewAtIndex(index, ListView.End);
 	}
 
 	function endEditMode() {


janiversen:
Tested version 2.0.1 (release) on iPad and on iPhone 6, in both cases, where it works correctly.

However newest master, shows the problem, and this patch fixes it. I recommend merging.

dirkhh:
So I explicitly switched to highlightFollowsCurrentItem in 783f9ee565adc1aaaa37cf1f2c50a9c591aca038 - presumably to fix a bug, specifically trying to fix the bug where seemingly randomly the details would very slowly scroll through the whole dive list.
I'm happy to merge this and see if that was a fluke - but this has me worried.

PR #1246, opened 2018-05-07
[build system] iOS: update bundle id to recommended variable

mfbernardes:
Since Xcode 7 the recommended bundle identifier is
the PRODUCT_BUNDLE_IDENTIFIER.

It will be set on project.pbxproj with the company ID you
set on Xcode plus the product ID.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Using the PRODUCT_BUNDLE_IDENTIFIER allows each developer to automatically have it's own identifier, without need to change Info.plist.

This makes (my) life easier. Just need to make sure it doesn't break Travis and @dirkhh release builds.

### DIFF
--- a/packaging/ios/Info.plist.in
+++ b/packaging/ios/Info.plist.in
@@ -11,7 +11,7 @@
 	<key>CFBundleIconName</key>
 	<string>AppIcon</string>
 	<key>CFBundleIdentifier</key>
-	<string>org.subsurface-divelog.subsurface-mobile</string>
+	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
 	<key>CFBundleName</key>
 	<string>Subsurface-mobile</string>
 	<key>CFBundlePackageType</key>


mfbernardes:
Murillo Bernardes

On Tue, May 8, 2018 at 8:21 AM, jan iversen <notifications@github.com>
wrote:

> The base for the patch is relevant, however it seems to cause problems
> when generating the Xcode project with Qt creator (packaging/ios/build.sh),
> seems as if Qt creator does not set variable.
>

Weird since the variable is define on the Xcode project created by qmake.

I take another look.


> How does it make your life easier ? when not using iCloud, the value seems
> little (unless I overlooked something). You get your own container (which
> cannot be deleted later), but your dives etc. are the same ?
>
Since I can't get a provisioning profile for
org.subsurface-divelog.subsurface-mobile I have to change it every time I
run packaging/ios/build.sh.

For some reason (never investigated the scenario) sometimes it even builds
correctly, but I can't load on a device due to a mismatch in provision and
info.plist.

I might just not know how to do it properly.

janiversen:
> Weird since the variable is define on the Xcode project created by qmake.

Sorry did not write it correctly, yes the Xcode project defines the variable, it just did not contain the correct value, but it was my fault. When running build.sh on master the Xcode project is filled with the correct value.

> Since I can't get a provisioning profile for
> org.subsurface-divelog.subsurface-mobile I have to change it every time I
> run packaging/ios/build.sh.

Hmmm that should not happen, unless we run it differently (I run it without parameters to generate all platforms). Your provision profile is only needed in the last phase of the Xcode run, which is a call to codesign. The Xcode run is done by Qt Creator (at least for me)

If you run Qt Creator and ask it to generate for the simulator it should never do codesigning.

I recommend that the PR gets merged.

dirkhh:
I'm confused by how this is supposed to work - or where I'm supposed to set that company identifier. With this patch applied I am now building com.yourcompanyname.subsurface-mobile - which is not really what I want :-)

janiversen:
You are not, it is set as default by QtCreator, when generating the Xcode project. Will try to test again on a native vm.

dirkhh:
My point was that I have to set it somewhere.
Once I figure out how to do that, I'll take the PR

Except that I have early meetings tomorrow... I'm already in bed, responding from my phone...
🙄
So this won't happen for a few hours.

mfbernardes:
I don’t remember from the top of my head. I’ll check in some hours when I have access to my Mac again.

dirkhh:
I spent a bit of time on this and I still don't understand at all how this is supposed to work.
Because Qt Creator does not set up the "com.yourcompany" part of the PRODUCT_BUNDLE_IDENTIFIER (trust me, I looked through the sources) - and I cannot seem to figure out how to get Xcode to use the correct one (which for me would be, indeed, org.subsurface-divelog). So how do YOU set it up, @mfbernardes ) to get the value that you need?

I'm happy to make changes that make everyone's lives easier - but not if they break my ability to actually build the app.

dirkhh:
Ah. Not Qt Creator. qmake is making that mess. And it has been broken a few times and is inconsistent between different Qt versions. Brilliant.

janiversen:
I stand corrected, I got mislead by my own system, which automatically generates a xcconfig file (which overwrites whatever settings Xcode has).

Unless you open the generated Xcode project, and set bundle_identifier it can only be set in the Info.plist file.

Since we are mostly concerned about making it easy to generate daily builds, releases etc. Changing Info.plist.in is not the way to go!

dirkhh:
OK, I'll close this PR then and instead hook something up that allows both this workflow and my workflow / the Travis workflow, err, work

janiversen:
I found a way to allow both workflows:
If you want to use your own Id, you need to define an environment variable, otherwise it is
set you org.....

This is a slight change (again) to build.sh, but with that everybody should be happy, I am testing the patch right now, before submitting it.

janiversen:
See #1256 that solves both concerns.

mfbernardes:
It's a love and hate with iOS dev. I don't really know and could not find what I did to get the company id set.

So, the way you did in the end looks good.

PR #1251, opened 2018-05-08
iOS, set lowest deployment target to 8.0

janiversen:
Define deployment target, instead of using user default.

iOS 8.0 is the oldest supported platform

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

iOS 8.0 is currently defined as supported, but Qt Creator did not know that, so it
used the user default.

### Changes made:

Added a variable to QtCreator that defines the min. deployment level.


dirkhh:
Thank you!

PR #1253, opened 2018-05-10
iOS: update build to use system sqlite3 and xml2, and bumped versions to match desktop

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [x ] Documentation change
- [ ] Language translation

Cleaned up iOS build, so it is easy to transfer to a new get_depends script.

Removed packages not needed (like libxml2) and use the apple supplied 
version instead (as scripts/build.sh) does.

Finally solved a problem, when libdivecomputer is cloned, but not
configured

On a personal note, hope we can solve this with less comments than
last time :-)

### DIFF
--- a/packaging/ios/Subsurface-mobile/Subsurface-mobile.pro
+++ b/packaging/ios/Subsurface-mobile/Subsurface-mobile.pro
@@ -107,12 +107,12 @@ RESOURCES += qml.qrc ../../../subsurface.qrc ../../../mobile-widgets/qml/mobile-
 
 LIBS += ../install-root/lib/libdivecomputer.a \
         ../install-root/lib/libgit2.a \
-        ../install-root/lib/libsqlite3.a \
         ../install-root/lib/libzip.a \
         ../install-root/lib/libxslt.a \
         ../install-root/lib/libxml2.a \
         ../googlemaps-build/libqtgeoservices_googlemaps.a \
-        -liconv
+        -liconv \
+	-lsqlite3
 
 INCLUDEPATH += ../install-root/include/ \
                ../install-root/lib/libzip/include \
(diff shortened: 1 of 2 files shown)


dirkhh:
Thanks for this.
Your commit title format doesn't follow our typical pattern (not a big deal, but I figured I'd point it out).
When I started on this, I couldn't get the bundled sqlite to work at all and the libxml2 was having problems with the libxslt. I assume that you tested this and this is no longer the case? In that case: AWESOME - I love saving build time and reducing dependencies.

janiversen:
I am using the version (currently in the simulator) without problems, please tell me if there are some special test I can run ?

Sorry for the wrong title format, I did not find a definition of how to do it, so I assumed "freestyle". Will have a look at some other commits. If you see it again, please notify me, and I will change it.

dirkhh:
Our test suite isn't usually run for the embedded builds, we don't have a formal set of regressions for iOS and Android :-(
If this works in the simulator and you can open cloud storage and save to cloud storage, that's about all the testing we can do.
Thanks for editing the commit titles - those look good.
The Travis fail looks like a timeout - I should change that to use git instead of the tar balls as well (just like you did for iOS). I restarted it, but my assumption is that this can be merged.

packaging/ios/Subsurface-mobile/Subsurface-mobile.pro
```diff
@@ -107,12 +107,12 @@ RESOURCES += qml.qrc ../../../subsurface.qrc ../../../mobile-widgets/qml/mobile-
...
 LIBS += ../install-root/lib/libdivecomputer.a \
         ../install-root/lib/libgit2.a \
-        ../install-root/lib/libsqlite3.a \
         ../install-root/lib/libzip.a \
         ../install-root/lib/libxslt.a \
         ../install-root/lib/libxml2.a \
         ../googlemaps-build/libqtgeoservices_googlemaps.a \
-        -liconv
+        -liconv \
+	-lsqlite3
```
dirkhh: ah, this looks like a whitespace issue?

dirkhh (approved):
tiny whitespace issue, but this can be fixed separately

mturkia:
It is tab vs space indentation issue. You are using tab for indent but this file uses spaces (against our CodingStyle). So one should use the same style that is used in the previous lines. BTW we do use space indent e.g. in XSLT files as well so the rule is somewhat language dependent.

PR #1256, opened 2018-05-10
Allow user defined PRODUCT_BUNDLE_IDENTIFIER

janiversen:
while keeping org..... for everybody else.

This allows a user to deploy using a wildcard provisioning profile,
while Travis and others have no change at all.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ x] Build system change
- [ ] Documentation change
- [ ] Language translation

Allows users to deploy on iOS devices using a wildcard deployment profile,
while securing that the default is still org....

See README on how to activate user defined bundle_product_identifier

### DIFF
--- a/packaging/ios/Info.plist.in
+++ b/packaging/ios/Info.plist.in
@@ -11,7 +11,7 @@
 	<key>CFBundleIconName</key>
 	<string>AppIcon</string>
 	<key>CFBundleIdentifier</key>
-	<string>org.subsurface-divelog.subsurface-mobile</string>
+	<string>@BUNDLE@</string>
 	<key>CFBundleName</key>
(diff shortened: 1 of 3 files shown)


janiversen:
That is of course your call, but your patch is not nearly as nice, since it means you have to open Xcode and change it. Also your patch is missing an update of README on how to use it.

dirkhh:
I agree with you on the lack of README update.
My patch was fundamentally based on Murillo's PR.
If you would rather see something based on an environment variable, I'll of course be happy to accept a fresh PR with that (as long as that doesn't break the workflows I care about, which Murillo's PR did).

PR #1257, opened 2018-05-10
Fixed erroneous comparison of cylinders and weight systems.

Schwaneberg:
Comparison of data structures with different type and size will always return true.

Signed-off-by: Oliver Schwaneberg <oliver.schwaneberg@gmail.com>

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

Removed a bug that caused the comparison of incompatible data structures.


dirkhh:
ouch, good catch!
For this one I'll let Travis finish (even though I don't see how this could fail)

neolit123 (approved):

PR #1262, opened 2018-05-11
added script to download/clone 3rd party libs.

janiversen:
Using 1 script instead of having the downloads in all
build.sh ensures consistency and lowers maintenance.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

get-dep-lib.sh supposed to be integrated in the different build.sh scripts, so that
we only have 1 file to maintain in regard to 3rd libraries.

Windows seems to have its own life, so currently not part of the script.

test can be done like

mkdir test; cd test
../subsurface/scripts/get-dep-lib android|desktop|ios .

If/when this PR is accepted, build.sh will be changed.

### DIFF
--- a/scripts/get-dep-lib.sh
+++ b/scripts/get-dep-lib.sh
@@ -1,0 +1,196 @@
+#!/bin/bash
+#
+
+# set version of 3rd party libraries
+CURRENT_LIBZIP="1.2.0"
+CURRENT_LIBGIT2="v0.26.0"
+CURRENT_HIDAPI="hidapi-0.7.0"
+CURRENT_LIBCURL="curl-7_54_1"
+CURRENT_LIBUSB="v1.0.21"
+CURRENT_OPENSSL="OpenSSL_1_1_0f"
+CURRENT_LIBSSH2="libssh2-1.8.0"
+CURRENT_XSLT="v1.1.29"
+CURRENT_SQLITE="3190200"
+CURRENT_LIBXML2="v2.9.4"
+CURRENT_LIBFTDI="1.3"
+
+
+
+# deal with all the command line arguments
+if [ $# -ne 2 ] ; then
+	echo "wrong number of parameters, format:"
+	echo "get-dep-lib <platform> <install dir>"
+	echo "where platform is one of scripts, ios or android"
+    echo "(the name of the directory where build.sh resides)"
+	exit -1
+fi
+
+PLATFORM=$1
+INSTDIR=$2
+if [ ! -d ${INSTDIR} ] ; then
+	echo "creating dir"
+	mkdir -p ${INSTDIR}
+fi
+
+case ${PLATFORM} in
+	scripts)
+		BUILD="libzip libgit2 googlemaps hidapi libcurl libusb openssl libssh2"
+		;;
+	ios)
+		BUILD="libzip libgit2 googlemaps libxslt"
+		;;
+	android)
+		BUILD="libzip libgit2 googlemaps libxslt sqlite libxml2 openssl libftdi libusb"
+		;;
+	*)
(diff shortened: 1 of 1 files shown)


janiversen:
Wonder, what would be the right prefix in the commit message for something general ?

dirkhh:
> Wonder, what would be the right prefix in the commit message for something general ?

We haven't been consistent in that. Sometimes it's things like "build system:" - but nothing is fine, too. Once we have established a pattern, it's nicer to stick with it (as it makes parsing the git log easier - changes that affect only one platform have such an established pattern, or changes that are just about Travis/CI).

scripts/get-dep-lib.sh
```diff
@@ -0,0 +1,195 @@
...
+
+if [[ $BUILD = *"googlemaps"* ]]; then
+	if [ ! -d googlemaps ] ; then
+		git clone https://github.com/Subsurface-divelog/googlemaps.git
+	fi
+    pushd googlemaps
+	git fetch origin
+    git checkout master
+    git pull --rebase
+	popd
```
dirkhh: space vs tab?

scripts/get-dep-lib.sh
```diff
@@ -0,0 +1,195 @@
...
+PLATFORM=$1
+INSTDIR=$2
+if [ ! -d ${INSTDIR} ] ; then
+	echo "creating dir"
+	mkdir -p ${INSTDIR}
+fi
+
+case ${PLATFORM} in
+	desktop)
+		BUILD="libzip libgit2 googlemaps hidapi libcurl libusb openssl libssh2"
```
dirkhh: I think that's a bit misleading - the Linux builds and Windows builds don't download these...
This PR doesn't include the intended USE of this script (which is fine), so I'm not sure how you are using this (maybe there's another PR for that already?), but "desktop" seems like a misleading choice.

dirkhh (approved):
I will refrain from starting new "bring me a rock" exercises and simply accept this. I have some vague ideas how this could evolve, but as it is it's good.

scripts/get-dep-lib.sh
janiversen: Ups. will correct

janiversen:
I am open for modifications, but lso think this script is somethung that will evolve as time goes on

dirkhh:
> I am open for modifications, but lso think this script is somethung that will evolve as time goes on

Based on recent experience I'd rather not do that in a PR. I'd appreciate a version with clean whitespace and then a PR that uses this script. If I have time later today (while doing my day job and figuring out what's broken with the EON Steel on Mac) I may tweak this a little :-)

janiversen:
I updated the commit message to clearly state how it is intended to be used.

Feel free to tweak it as you like, you know better than me what is "correct". To me the only point was to have a single place where we decide what to download.

Tweak it directly in my commit or do as you did last, close this PR and open a new one.

Disclaimer: I have some ideas of how different e.g. Mac/iOS could share the downloads, but that is future.

I will prepare a PR, with the changes to the 3 build.sh (each with its own commit) later this weekend. I am right now working on a idea I have for tables/phones).

Question: is FULLSCREEN_SUPPORT and !NO_USERMANUAL ever used, seems to me as being dead code ?

dirkhh:
FULLSCREEN_SUPPORT was a conditional compile added by @neolit123 in 3f7900bf4e382cc57133e307efeaa3856c90c505 as a fix / reminder to something that's broken.
The user manual viewer can be disabled - that's what this option is for.

dirkhh:
Manually merged for one remaining whitespace fix plus tiny edits in commit message

Issue #953, opened 2017-12-16
mobile: delete dive followed by add dive executes wrong action


janmulder:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Open a dive in details screen, and delete it. See that a UNDO notification is shown, and do not use this.
Back on the dive list, hit + for add dive. The just deleted dive is restored.

Further analysis shows that on hitting the +, in fact the undoDelete is called, and that is exactly what we see happening here: the just deleted dive is restored. Obviously, that undoDelete is called in only visible running in a debugger. Further, removing the showPassiveNotification() undo, makes the + working as it should (and the delete just deletes as it should).

So, my current idea is that the showPassiveNotification() from Kirigami messes with the application wide action button, and even when the user just waits for the notification to timeout, the proper state is not restored on that action button. Did not inspect Kirigami code at this point.

### Operating system:
What OS are you running, including OS version and the language that you are running in: mobile on desktop (Arch Linux), and android (7.1.2)
What device are you using? not relevant

### Subsurface version:
What version of Subsurface are you running?  master
Does the same happen on another Subsurface version? not analyzed, too much related issues in this area
Are you using official release, test build, or compiled yourself? self
Provide Git hash if your are building Subsurface yourself.  2890be92568d36

### Steps to reproduce:
1) Open a dive in details screen, and delete it.
2) See that a UNDO notification is shown, and do not use this
3) back on the dive list, hit + for add dive. The just deleted dive is restored.

### Current behavior:
After delete dive (on details screen), the + icon (add dive) acts as an undo delete.

### Expected behavior:
Obviously, delete dive and add dive are functionally unrelated, so undo shall undo and add dive shall add.

### Additional information:
related to #950, but technically very different

janmulder:
Fixed in #1265 (but found some weird enumeration like issue in the process, see #1265).

PR #1265, opened 2018-05-13
mobile: prevent Action buttons hijack after delete

mfbernardes:
This gets delete dive working properly.

Kirigami passive notification ends up hijacking area where the
"Add dive" or "Delete" or "Discard" buttons are shown. So after
deleting a dive the "Undo" button from the notification
keeps handling the touch events even when not visible.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

After deleting a dive a notification with Undo button is displayed. After 3s the notification disappear.

Any subsequent event intended to the right action button (either "Add dive" or "Delete" or "Discard", depending on the page) are actually handled by the invisible Undo button. This makes the first click to undelete the dive and all other clicks get "Trying to undo delete but can't find the deleted dive" into the log.

### DIFF
--- a/scripts/mobilecomponents.sh
+++ b/scripts/mobilecomponents.sh
@@ -80,6 +80,10 @@ sed -i -e "s/visible: root.action/visible: root.action \&\& \!Qt.inputMethod.vis
 sed -i -e "s/visible: root.leftAction/visible: root.leftAction \&\& \!Qt.inputMethod.visible/g" src/controls/private/ActionButton.qml
 sed -i -e "s/visible: root.rightAction/visible: root.rightAction \&\& \!Qt.inputMethod.visible/g" src/controls/private/ActionButton.qml
 
+# kirigami hack: passive notification hijacks area even after disabled.
+# https://bugs.kde.org/show_bug.cgi?id=394204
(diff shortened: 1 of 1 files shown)


dirkhh:
That sounds neat. One more of these things that I would never have found.
Since there's nothing on Travis that would even exercise this code, I'll just merge as is.

mfbernardes:
I haven't seen #953 before, but it fixes exactly what is described there.

I have tested that scenario, it works fine on iOS now.

janmulder:
Ok, a little more research. This time on desktop (instead of Android device). Hitting the + to add (after delete) takes at least 30 seconds for me to show the added (empty dive). But the page shown is correct. I'm pretty sure that we see (again) one of those weird enumeration problems (@dirkhh) we had before. As I test with a real logbook, with approx. 1000 dives in it, the enumeration over all the dives takes forever, and on my Android device it even crashes the device (highly likely to be an improperly handled out of resources thing).

So, I will close #953 as that seems properly fixed. @mfbernardes: do you test with a small logbook? In that case you might not see the things I'm seeing.

PR #1268, opened 2018-05-13
Build with depend

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Removed local download/clone from scripts/build.sh and packaging/ios/build.sh
and replaced with get-dep-lib

Android is not touched, because I currently do not have a android test environment
(soon to be changed).

Feel free to make changes to my commits.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -69,15 +69,6 @@ while [[ $# -gt 0 ]] ; do
 	shift
 done
 
-# most of these will only be needed with -build-deps on a Mac
-CURRENT_LIBZIP="1.2.0"
-CURRENT_HIDAPI="hidapi-0.7.0"
-CURRENT_LIBCURL="curl-7_54_1"
-CURRENT_LIBUSB="v1.0.21"
-CURRENT_OPENSSL="OpenSSL_1_1_0f"
-CURRENT_LIBSSH2="libssh2-1.8.0"
-CURRENT_LIBGIT2="v0.26.0"
-
 # Verify that the Xcode Command Line Tools are installed
 if [ $PLATFORM = Darwin ] ; then
 	if [ -d /Developer/SDKs ] ; then
@@ -202,23 +193,18 @@ if [[ $PLATFORM = Darwin || "$LIBGIT" < "24" ]] ; then
 	# feature). So we painfully need to build the dependencies ourselves.
 
 	if [ "$BUILD_DEPS" == "1" ] ; then
-		if [ ! -d libzip-${CURRENT_LIBZIP} ] ; then
-			curl -O https://libzip.org/download/libzip-${CURRENT_LIBZIP}.tar.gz
-			tar xzf libzip-${CURRENT_LIBZIP}.tar.gz
-		fi
-		cd libzip-${CURRENT_LIBZIP}
+		./subsurface/scripts/get-dep-lib.sh single . libzip
+		pushd libzip
 		mkdir -p build
 		cd build
 		../configure CFLAGS="$OLDER_MAC" --prefix=$INSTALL_ROOT
 		make -j4
 		make install
+		popd
 
-		cd $SRC
 
-		if [ ! -d hidapi ] ; then
-			git clone https://github.com/signal11/hidapi
-		fi
-		cd hidapi
+		./subsurface/scripts/get-dep-lib.sh single . hidapi
+		pushd hidapi
 		# there is no good tag, so just build master
 		bash ./bootstrap
 		mkdir -p build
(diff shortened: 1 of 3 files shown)


dirkhh:
It looks like something is off with libgit2 - Mac, Linux, and Linux/qt55 are failing because of it.
I haven't looked into "why" at all. Happy to help, but I'm assuming that you have started to look into it already?

dirkhh:
Yes, Travis isn't perfect, but it sure helps a lot.
I'll simply wait for you here - this isn't urgent.

janiversen:
Do not ask me how, but when I look at the raw log of the failing jobs, I cannot find the call to get-dep-lib (even though e.g. my pushd commands are there). Can you help me explain that ?

janiversen:
I think I got it, can it be travis runs without -build-deps ? 

That would explain why I do not see the get-dep-lib call.

dirkhh:
Travis is controlled by https://github.com/Subsurface-divelog/subsurface/blob/master/.travis.yml
which calls three scripts under `scripts/${SUBSURFACE_PLATFORM}/ ` depending on which of the test environments it is. Those scripts then trigger the setup, the actual build, and what to do after the build succeeds.
Right now, for Mac, we switched to not building our dependencies and instead using them from Homebrew (because it ended up that that was much faster). The call to `build.sh` is here https://github.com/Subsurface-divelog/subsurface/blob/master/scripts/mac/travisbuild.sh#L18

dirkhh:
Oh, and the reason this is faster is that we have a setup that allows us to run Homebrew to build everything and then cache that - and the download of the cache is much faster than the build.
Obviously, the build needs to work in either of these two scenarios.
Equally obviously (I hope), for your testing you could simply modify Line 18 referenced in the previous comment and have it build things as you want to see them built.

janiversen:
Finally ! will integrate in the android build once this is done.

dirkhh (commented):
Hmm - half way through the review the commits changed again... I'll submit this review now and start again with the latest versions, I guess...

scripts/build.sh
```diff
@@ -193,63 +193,70 @@ if [[ $PLATFORM = Darwin || "$LIBGIT" < "24" ]] ; then
 	# feature). So we painfully need to build the dependencies ourselves.
 
 	if [ "$BUILD_DEPS" == "1" ] ; then
```
dirkhh, line 195: so all of the changes in this commit are inside the `build-deps` condition. The commit message sounds like you are making changes for Travis, but of course this code isn't run on Travis.
I guess I'm confused...

dirkhh (changes requested):
I really like the direction you are taking this.
The central "getter" script is definitely an improvement.
But some of the changes in this patch series confuse me... I think this may need a little more work (at least to make the commit messages and actual changes correlate better)

janiversen:
I will implement your suggestions, and update the commit messages.

Right now I ran into a "naughty" problem. curl is accepted on all platforms except Android, do I really have to make a "if android use wget else use curl" ?

dirkhh:
We should be able to just install curl inside the container (note, Travis builds the Android apk inside a Ubuntu container, not inside the "outside VM" - so you need to install it as part of the `android-wrapper-script`, IIRC.

janiversen:
Cleaned up commits (the 2 commits earlier of scripts/build.sh were 2 different ideas) and texts.

Also implemented your comments.

Let us see, how much I have broken :-)

Will do the android part, once the general parts works.

dirkhh (approved):

dirkhh:
Android fails when building googlemaps plugin with 
```
/workspace/subsurface/packaging/android/build.sh: line 165: QMAKE: unbound variable
```

janiversen:
Already corrected, it was my semi-manual merge that took out too many lines (me blames my dm, because I am in parallel analysing my dives from earlier today)

janiversen:
This is not sufficient, how about I just quit the android integration. The rest should be ready to merge.

I can then try to setup a android build environment and test locally (but that will take some days)

wdyt?

dirkhh:
Works for me

dirkhh:
The nice thing about the way Android works on Travis is that it's really easy to recreate the exact same environment locally - simply install Ubuntu in a VM, install the packages listed and then run the scripts...

janiversen:
Yeah travis is GREEEN :-)

I am downloading ubuntu right now, but I suspect it will take a little while to get things working. I already have a android vm for testing. So the good thing (for the project) is that I should be able to e.g. try and bump the library versions myself.

PR #1271, opened 2018-05-14
Header refactor

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Baby step to shrink dive.h
I'm interested to hear what others think would be good logical ways to break things up.
I'm not convinced that this really is a compile time advantage, though.

### Changes made:

1) moved a couple of math macros and string inline functions into a separate header
2) dropped dive.h from the source files that didn't need it any more

### Release note:

Doesn't seem worse a mention in the Release note

### Mentions:

It would be nice if a couple of people take a look - if you think this is going the wrong way about it, happy to hear other ideas how to approach this.
@atdotde @mturkia @sfuchs79 @bstoeger @neolit123

### DIFF
--- a/core/cloudstorage.cpp
+++ b/core/cloudstorage.cpp
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "cloudstorage.h"
 #include "pref.h"
-#include "dive.h"
 #include "helpers.h"
 #include "core/subsurface-qt/SettingsObjectWrapper.h"
 #include <QApplication>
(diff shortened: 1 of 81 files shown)


neolit123 (approved):
the change looks good. :+1:

bstoeger:
Perhaps use the opportunity to un-inline `includes_string_caseinsensitive()`? Looks a bit voluminous for an inlined function.

PS: This functions seems to have only one caller - you might even move it to `core/git-access.c`

dirkhh:
Good idea @bstoeger - I'll do that

dirkhh:
Any more comments? Otherwise I'll just merge it

PR #1097, opened 2018-02-05
[RFC] Mobile: try to detangle the IO interface

bstoeger:
### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This is part of a long-term project attempting to establish a clear interface between the GUI and the IO backend. Sorry for the slow progress - I'm horribly busy at the moment.

One particularly intransparent aspect of the current implementation is the `prefs.git_local_only` flag, which is a mix of state, input and output parameter. A first step in cleaning this up will be making `prefs.git_local_only` effectively an input/output parameter. And in a second step also explicitly (see also 
#928).

Indeed, the mobile version already has a "property" `m_syncToCloud`, which makes `prefs.git_local_only` look kind of redundant. This PR attempts to make `m_syncToCloud` the authoritative flag and use `prefs.git_local_only` only to control the IO code.

I'm posting this as an RFC and it's obviously post-release material. It was tested lightly (sync to cloud with and without network), but I probably butchered it in a nasty way. It will be hard to test all the special cases.

BTW: As opposed to the desktop version, the mobile version does not seem to have something like an offline-mode. Is this correct?

PR #1272, opened 2018-05-14
Cloud-storage: consistently don't save empty file on desktop version

### LINKED ISSUE #1228: save (and sync cloud) and "Save to cloud", works different with empty dive list.
### Describe the issue:

- [ x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Start subsurface and make "cloud storage online".

Delete all dives.

Try "Save to cloud storage" and see you get an error message, cannot save empty logbook

Then "Save" which will automatically sync with the cloud, and an empty logbook is saved.

### Operating system:

MacOSX

Mac

No

### Subsurface version:

4.7.8

Official

### Steps to reproduce:

see above

### Current behavior:

inconsistent

### Expected behavior:

consistent, either allow "Save to cloud" to save an empty logbook, or give an error message from sync.

bstoeger:
Make the behavior consistent: Don't save an empty file to the cloud,
neither on selection of "Save to cloud" nor on "Save". The latter
was not the case. It was a bit hard to trigger: Open cloud, delete
all dives, save.

Fixes #1228

Reported-by: Jan Iversen <jani@apache.org>

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes an UI inconsistency reported in #1228. Originally I planned to move the "Divelog empty" check into the IO layer, but then it would have also hit general git repositories. And users of those should know what they are doing. I have a rather big patch set, which tries to abstract the file handling and which can bother about this kind of things. But this will have to wait (among other things eg for #1097).

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -609,13 +609,21 @@ void MainWindow::on_actionCloudstorageopen_triggered()
 	refreshDisplay();
 }
 
+// Return whether saving to cloud is OK. If it isn't, show an error return false.
+static bool saveToCloudOK()
+{
+	if (!dive_table.nr) {
+		report_error(qPrintable(QObject::tr("Don't save an empty log to the cloud")));
+		return false;
+	}
+	return true;
+}
+
 void MainWindow::on_actionCloudstoragesave_triggered()
 {
 	QString filename;
-	if (!dive_table.nr) {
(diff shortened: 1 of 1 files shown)


dirkhh:
Nice catch - yes, this makes the code more consistent.
I just kicked the Mac build again (stupid DNS resolve issue caused the failure).

PR #1279, opened 2018-05-15
Dive pictures: Don't enter infinity loop on invalid pictures

bstoeger:
The recently committed refactoring of the dive-picture code introduced
a severe bug:

If an image couldn't be loaded from disk owing to an invalid file, the
filename was interpreted as an url and loaded in the background. This
succeeded, because the file actually exists. After download, the file
would then still be invalid and the whole thing restarted, leading to
an infinity loop.

To fix this, do two things:

1) Don't even try to download local files.
2) If interpreting a downloaded file fails, don't try the downloading
business again.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a severe bug introduced in the recent dive-picture refactoring: An invalid image file lead to an infinite loop. Hopefully, this was the worst bomb that I introduced!
### Changes made:

1) Don't try to download local files.
2) If interpreting a downloaded file fails, don't try the downloading business again.

### DIFF
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -39,8 +39,11 @@ void ImageDownloader::load(QString filename, bool fromHash)
 {
 	QUrl url = fromHash ? cloudImageURL(qPrintable(filename)) : QUrl::fromUserInput(filename);
 
-	if (!url.isValid())
+	// If this is a file, we tried previously -> don't bother trying it again
+	if (url.scheme() == "file" || !url.isValid()) {
 		emit failed(filename);
+		return;
+	}
 
 	QNetworkRequest request(url);
 	request.setAttribute(QNetworkRequest::User, filename);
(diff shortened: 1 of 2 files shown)


core/imagedownloader.cpp
```diff
@@ -39,8 +39,11 @@ void ImageDownloader::load(QString filename, bool fromHash)
 {
 	QUrl url = fromHash ? cloudImageURL(qPrintable(filename)) : QUrl::fromUserInput(filename);
 
-	if (!url.isValid())
+	// If this is a file, we tried previously -> don't bother trying it again
+	if (url.scheme() == "file" || !url.isValid()) {
```
atdotde, line 43: It's fine like this, but you could have also written url.isLocalFile()

bstoeger: Damn - missed that function yesterday night. :( That is indeed much nicer. Care to submit a cleanup-PR?

PR #1281, opened 2018-05-16
Dive pictures: don't attempt reading the same image twice

bstoeger:
If loading of an image failed, we tried to see if we find a
canonical filename in the cache. There's no point in rereading
the picture if the canonical and the original filename are
the same.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a minor improvement of the picture code: Don't attempt to read an image a second time if the first time failed. This error condition is a bit hard to pull of: A picture was read once, but then corrupted. Nevertheless, this happened in my tests, so we might as well treat it gracefully.


atdotde (approved):
Looks good to me. Maybe at some point we should get rid of the qDebug's or translate them to actual output when running with -v

PR #1283, opened 2018-05-16
Planner: Enable CCR gas use calculations with bailout

willemferguson:
As the title says. Gas use during CCR dives is normally not
calculated in the planner. However, if the divemode of a segment
is NOT CCR, then the gas consumption should be calculated for all
non-CCR segments.

This change leaves the previous mechanism in place where CCR bailout is
initiated when a segments has a setpoint of zero. Eventually we will
probably standardise on only the first mechanism, but for now we leave
both mechanisms in place.

Signed-off-by: Willem Ferguson <willemferguson@zoology.up.ac.za>

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@atdotde

### DIFF
--- a/core/planner.c
+++ b/core/planner.c
@@ -260,13 +260,14 @@ static void create_dive_from_plan(struct diveplan *diveplan, struct dive *dive, 
 {
 	struct divedatapoint *dp;
 	struct divecomputer *dc;
-	struct sample *sample;
+	struct sample *sample, *s;
 	struct event *ev;
 	cylinder_t *cyl;
 	int oldpo2 = 0;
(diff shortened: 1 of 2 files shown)


willemferguson:
@atdotde , I did some changes to this PR in order to eliminate some of the previous problems with the code  that we discussed. I would appreciate it immensely if you would be prepared to look through this and see if it has been done in a sensible way?

willemferguson:
The purpose of this PR has been achieved by other PR's that have been merged. This PR is therefore now superfluous.

PR #1286, opened 2018-05-16
Dive pictures: Don't plot pictures twice when changing current dive

bstoeger:
In MainWindow::current_dive_changed() first plotDive() is called,
which replots all the pictures by calling plotPictures(). This
is pointess, because it plots the pictures of the previous dive.

Then, updateDiveInfo() is called, which resets the dive pictures
and automatically replots them. Thus, switching between dives
both with hundreds of pictures is way slower than necessary.

Switching the plotDive() and updateDiveInfo() calls doesn't work.
The reason is not 100% clear, but it doesn't make sense to plot
pictures of the new dive as long as the profile still shows the
old dive anyway.

As a quick-fix, add a flag to plotDive(), which tells the function
to clear the pictures list instead of redrawing it.
Ultimately, plotDive() should probably be split in two functions.
One for the callers who update the pictures themselves and one
for the others.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is an optimization of the profile plot code: On dive-change pictures of the *old* dive were plot first, which seems quite pointless.

The fix is quite disgusting in that it introduces an additional bool-parameter to `plotDives()`, which controls its behavior. But the whole `DivePictureModel` <-> `ProfileWidget2` interface is horribly messy to begin with. I don't dare to make in-depth changes to this.

I did some quick benchmarking for the `plotDive()`/`updateDiveInfo()` pair of calls. I switched continuously between two dives:

Before patch
```
dive changed to: 15
Elapsed: 135 msec
dive changed to: 14
Elapsed: 142 msec
dive changed to: 15
Elapsed: 133 msec
dive changed to: 14
Elapsed: 147 msec
dive changed to: 15
Elapsed: 132 msec
dive changed to: 14
Elapsed: 139 msec
dive changed to: 15
Elapsed: 134 msec
dive changed to: 14
Elapsed: 147 msec
dive changed to: 15
Elapsed: 137 msec
dive changed to: 14
Elapsed: 142 msec
```

After patch
```
dive changed to: 15
Elapsed: 63 msec
dive changed to: 14
Elapsed: 135 msec
dive changed to: 15
Elapsed: 64 msec
dive changed to: 14
Elapsed: 134 msec
dive changed to: 15
Elapsed: 65 msec
dive changed to: 14
Elapsed: 135 msec
dive changed to: 15
Elapsed: 64 msec
dive changed to: 14
Elapsed: 135 msec
dive changed to: 15
Elapsed: 63 msec
dive changed to: 14
Elapsed: 136 msec
dive changed to: 15
Elapsed: 64 msec
dive changed to: 14
Elapsed: 135 msec
```

Note that the dive with internal id 14 has a bunch of pictures on the profile plot, whereas id 15 has only a few. Therefore, there is a huge improvement when switching 14->15, but only a small one for 15->14.

Looks like the data is quite convincing. :)

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -531,7 +531,7 @@ void MainWindow::current_dive_changed(int divenr)
 	if (divenr >= 0) {
 		select_dive(divenr);
 	}
-	graphics()->plotDive();
+	graphics()->plotDive(nullptr, false, true);
 	information()->updateDiveInfo();
 	configureToolbar();
 	MapWidget::instance()->reload();
@@ -1095,7 +1095,7 @@ void MainWindow::on_actionAddDive_triggered()
 	graphics()->setAddState();
 	DivePlannerPointsModel::instance()->createSimpleDive();
 	configureToolbar();
-	graphics()->plotDive();
+	graphics()->plotDive(nullptr, false, true);
 	fixup_dc_duration(&displayed_dive.dc);
 	displayed_dive.duration = displayed_dive.dc.duration;
 
@@ -1314,7 +1314,7 @@ void MainWindow::on_actionPreviousDC_triggered()
 	unsigned nrdc = number_of_computers(current_dive);
 	dc_number = (dc_number + nrdc - 1) % nrdc;
 	configureToolbar();
-	graphics()->plotDive();
+	graphics()->plotDive(nullptr, false, true);
 	information()->updateDiveInfo();
 }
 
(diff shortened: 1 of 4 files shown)


neolit123:
lgtm, thanks.

> Looks like the data is quite convincing. :)

indeed.

neolit123 (approved):

PR #1287, opened 2018-05-16
Dive pictures: Fix crash on picture delete

bstoeger:
The recent simplification of the close button code introduced a crash:
Deletion of pictures caused an invalid memory access, because the
CloseButtonItem was deleted with the parent DivePicture item.
For some (not fully understood!) reason, a reference to this button
was stored in the depths of Qt.

Empirically, it was found out that removing the first line of the pair
       QGraphicsItem::mousePressEvent(event);
       emit clicked();
fixed the crash.

It seemed therefore prudent to remove the whole questionable signal/slot
mechanism and directly call the removePicture() function of the parent.
Thus, the intermediate DiveButtonItem class became unnecessary and was
removed, leading to a shallower class hierarchy.

Unfortunately, CloseButtonItem must still be derived from QObject owing
to the Q_PROPERTY machinery, which is in turn needed for animation.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a crash and is at the same time a code cleanup. The reason for the original crash is not fully understood. Qt's object model is mad.

### Changes made:

1) Don't use signal/slot system in CloseButtonItem.
2) Remove unnecessary DiveButtonItem class.

### DIFF
--- a/profile-widget/divepixmapitem.cpp
+++ b/profile-widget/divepixmapitem.cpp
@@ -16,31 +16,26 @@ DivePixmapItem::DivePixmapItem(QGraphicsItem *parent) : QGraphicsPixmapItem(pare
 {
 }
 
-DiveButtonItem::DiveButtonItem(QGraphicsItem *parent): DivePixmapItem(parent)
-{
-}
-
-void DiveButtonItem::mousePressEvent(QGraphicsSceneMouseEvent *event)
-{
-	QGraphicsItem::mousePressEvent(event);
-	emit clicked();
-}
-
-CloseButtonItem::CloseButtonItem(QGraphicsItem *parent): DiveButtonItem(parent)
+CloseButtonItem::CloseButtonItem(QGraphicsItem *parent): DivePixmapItem(parent)
 {
 	static QPixmap p = QPixmap(":list-remove-icon");
 	setPixmap(p);
 	setFlag(ItemIgnoresTransformations);
 }
 
+void CloseButtonItem::mousePressEvent(QGraphicsSceneMouseEvent *event)
+{
+	qgraphicsitem_cast<DivePictureItem*>(parentItem())->removePicture();
+}
+
 void CloseButtonItem::hide()
 {
-	DiveButtonItem::hide();
+	DivePixmapItem::hide();
 }
 
 void CloseButtonItem::show()
 {
-	DiveButtonItem::show();
+	DivePixmapItem::show();
 }
 
 DivePictureItem::DivePictureItem(QGraphicsItem *parent): DivePixmapItem(parent),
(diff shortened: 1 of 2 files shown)


neolit123 (approved):
```
The recent simplification of the close button code introduced a crash:
Deletion of pictures caused an invalid memory access, because the
CloseButtonItem was deleted with the parent DivePicture item.
For some (not fully understood!) reason, a reference to this button
was stored in the depths of Qt.

Empirically, it was found out that removing the first line of the pair
       QGraphicsItem::mousePressEvent(event);
       emit clicked();
fixed the crash.

It seemed therefore prudent to remove the whole questionable signal/slot
mechanism and directly call the removePicture() function of the parent.
Thus, the intermediate DiveButtonItem class became unnecessary and was
removed, leading to a shallower class hierarchy.
```

i think the fix is OK. i will let others comment if they have any concerns.

```
Unfortunately, CloseButtonItem must still be derived from QObject owing
to the Q_PROPERTY machinery, which is in turn needed for animation.
```

that's fine.

neolit123:
@bstoeger 
the linux build is complaining with:
```
[ 34%] Building C object core/CMakeFiles/subsurface_corelib.dir/import-csv.c.o
/home/travis/build/Subsurface-divelog/subsurface/profile-widget/divepixmapitem.cpp: In member function â€˜virtual void CloseButtonItem::mousePressEvent(QGraphicsSceneMouseEvent*)â€™:
/home/travis/build/Subsurface-divelog/subsurface/profile-widget/divepixmapitem.cpp:28:54: error: â€˜class DivePictureItemâ€™ has no member named â€˜removePictureâ€™
  qgraphicsitem_cast<DivePictureItem*>(parentItem())->removePicture();
                                                      ^
```
https://api.travis-ci.org/v3/job/379920526/log.txt

bstoeger:
This is utterly strange, because it means that `SUBSURFACE_MOBILE` is defined for the Linux build? Can you check that? I compile on Linux and there it works.

In any case, I removed the `#ifdef`, because if we have pictures, there seems to be no reason not to have a `removePicture()` function. Let's see if this compiles. If we don't want pictures on mobile, the whole class should not be compiled, isn't it?

neolit123:
> This is utterly strange, because it means that SUBSURFACE_MOBILE is defined for the Linux build? Can you check that? I compile on Linux and there it works.

that last log with the error on Linux didn't indicate usage of the SUBSURFACE_MOBILE macro.
this time the build passed, so your latest change is a fix.

i will leave it to someone else to review again and merge.

PR #1288, opened 2018-05-17
Fix dive modes names translation plus some code cleanup

sfuchs79:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [x] Language translation

Consistent names for dive modes:
Open circuit
CCR
pSCR
Freedive

Plus fix translations for dive mode names. This is changed to go on top of #1289 and correcting a few things there. Discussion can be found at #1289 

Plus add translation for event name "modechange".

Plus some code cleanup ("Swith").

### Additional information:

This is a little bit tricky because it could work as long as strings are not present in translation files but break s.th. after new strings are added. Please someone needs to carefully review this!

### Mentions:

@atdotde @willemferguson

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -36,7 +36,7 @@ const char *cylinderuse_text[] = {
 	QT_TRANSLATE_NOOP("gettextFromC", "OC-gas"), QT_TRANSLATE_NOOP("gettextFromC", "diluent"), QT_TRANSLATE_NOOP("gettextFromC", "oxygen"), QT_TRANSLATE_NOOP("gettextFromC", "not used")
 };
 
-// For user visible text
+// For user visible text but still not translated
 const char *divemode_text_ui[] = {
 	QT_TRANSLATE_NOOP("gettextFromC", "Open circuit"),
 	QT_TRANSLATE_NOOP("gettextFromC", "CCR"),
--- a/profile-widget/diveeventitem.cpp
+++ b/profile-widget/diveeventitem.cpp
@@ -194,7 +194,7 @@ void DiveEventItem::setupToolTipString(struct gasmix *lastgasmix)
 		}
 		*lastgasmix = *mix;
 	} else if (same_string(internalEvent->name, "modechange")) {
-		name += tr(": %1").arg(divemode_text_ui[internalEvent->value]);
+		name += QString(": %1").arg(divemode_text_ui[internalEvent->value]);
 	} else if (value) {
 		if (type == SAMPLE_EVENT_PO2 && same_string(internalEvent->name, "SP change")) {
 			name += QString(": %1bar").arg((double)value / 1000, 0, 'f', 1);
--- a/profile-widget/diveeventitem.cpp
+++ b/profile-widget/diveeventitem.cpp
@@ -194,7 +194,7 @@ void DiveEventItem::setupToolTipString(struct gasmix *lastgasmix)
 		}
 		*lastgasmix = *mix;
 	} else if (same_string(internalEvent->name, "modechange")) {
-		name += QString(": %1").arg(divemode_text_ui[internalEvent->value]);
+		name += QString(": %1").arg(tr(divemode_text_ui[internalEvent->value]));
 	} else if (value) {
 		if (type == SAMPLE_EVENT_PO2 && same_string(internalEvent->name, "SP change")) {
 			name += QString(": %1bar").arg((double)value / 1000, 0, 'f', 1);
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1623,7 +1623,8 @@ void ProfileWidget2::addDivemodeSwitch()
 	QPointF scenePos = mapToScene(mapFromGlobal(action->data().toPoint()));
 	for (i = 0; i < UNDEF_COMP_TYPE; i++)
 		if (QString(divemode_text_ui[i]) == action->text())
-			add_event(current_dc, lrint(timeAxis->valueAt(scenePos)), 8, 0, i, "modechange");
+			add_event(current_dc, lrint(timeAxis->valueAt(scenePos)), 8, 0, i,
+				QT_TRANSLATE_NOOP("gettextFromC", "modechange"));
 	invalidate_dive_cache(current_dive);
 	mark_divelist_changed(true);
 	replot();
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1436,21 +1436,21 @@ void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
 	QMenu *changeMode = m.addMenu(tr("Change divemode"));
 	if (divemode != OC) {
 		QAction *action = new QAction(&m);
-		action->setText(divemode_text_ui[OC]);
+		action->setText(tr(divemode_text_ui[OC]));
 		connect(action, SIGNAL(triggered(bool)), this, SLOT(addDivemodeSwitch()));
 		action->setData(event->globalPos());
 		changeMode->addAction(action);
 	}
 	if (divemode != CCR) {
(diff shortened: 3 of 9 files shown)


atdotde:
Hi,

> On 17. May 2018, at 22:02, willemferguson <notifications@github.com> wrote:
> 
> Stefan, two points. 1)Are you sure you wish to translate the terms used in the xml code of the dive log? Surely no user ever sees that. It is likely to bring about nonportability of a dive log and potentially all sorts of incompatibility with previous versions of the software.

You are right. We don’t translate stuff that gets written to xml.

> 2) Be very careful not to mess with the constant values that are defined in dive.h and dive.c. There the value OC has a very specific meaning because it is part of an enum.
> In general, my approach would be to ensure consistency and readability of items in the UI, but to leave the xml as is. Or am I mis-understanding?

The string for the OC value of the enum should be „Open circuit“.

Best
Robert

dirkhh:
How does this relate to #1289 ? :-)

atdotde:
Dirk,

> Am 18.05.2018 um 18:09 schrieb Dirk Hohndel <notifications@github.com>:
> 
> How does this relate to #1289 ? :-)
> 

This is supposed to be addressed in the fourth commit in that PR.  Which should of course be squashed with the first. 

Best
Robert

sfuchs79:
@willemferguson 
Clearly translating stuff going into the XML was not my idea but a bad mistake. Obviously I didn't look close enough at the code.

@atdotde @dirkhh 
Using Roberts PR #1289 instead of this one is ok for me. I guess it avoids some mistakes I did.
But also in #1289 I believe I still see some mistakes. And it doesn't cover everything I tried to cover here regarding translations.
I will test+review #1289 and comment there...

profile-widget/diveeventitem.cpp
```diff
@@ -194,7 +194,7 @@ void DiveEventItem::setupToolTipString(struct gasmix *lastgasmix)
 		}
 		*lastgasmix = *mix;
 	} else if (same_string(internalEvent->name, "modechange")) {
-		name += tr(": %1").arg(divemode_text[internalEvent->value]);
+		name += QString(": %1").arg(tr(divemode_text[internalEvent->value]));
```
atdotde: I think this does not work as the string to be translated has to be known at compile time.

profile-widget/profilewidget2.cpp
```diff
@@ -1610,8 +1610,9 @@ void ProfileWidget2::addDivemodeSwith()
 	QAction *action = qobject_cast<QAction *>(sender());
 	QPointF scenePos = mapToScene(mapFromGlobal(action->data().toPoint()));
 	for (i = 0; i < UNDEF_COMP_TYPE; i++)
-		if (QString(divemode_text[i]) == action->text())
-			add_event(current_dc, lrint(timeAxis->valueAt(scenePos)), 8, 0, i, "modechange");
+		if (QString(tr(divemode_text[i])) == action->text())
```
atdotde: as above.

profile-widget/profilewidget2.cpp
```diff
@@ -1610,8 +1610,9 @@ void ProfileWidget2::addDivemodeSwith()
...
+			add_event(current_dc, lrint(timeAxis->valueAt(scenePos)), 8, 0, i,
+				QT_TRANSLATE_NOOP("gettextFromC", "modechange"));
```
atdotde: Don't translate event names.

atdotde (commented):
Yesterday, I did not realise you actually had a PR, I thought this was just a bug report. Would you agree that I solved the things you mentioned in my PR?

profile-widget/profilewidget2.cpp
sfuchs79: I guess there is some misunderstanding regarding QT_TRANSLATE_NOOP. This doesn't really translate the string but only pushes it to the translation system. For the position where it is called it is a NO-OP.
It makes it possible to translate the string later-on for the UI only. For the event names this is done here:
https://github.com/sfuchs79/subsurface/blob/master/profile-widget/diveeventitem.cpp#L149

That is a strategy you could maybe also use in your PR #1289 and which would maybe help to avoid having to duplicate the const char *divemode_text...

sfuchs79: No, forget the last sentence I wrote. At the moment we want to change from "OC" to "Open circuit" for the UI we maybe really need two versions of the string list. Otherwise we may change the XML from "OC" to "Open cuircuit" as well and this would lead to the kind of incompatibility @willemferguson mentioned.

sfuchs79:
Updated.
Please review and test again!

To finally test this we would need to recreate translation source strings.

core/dive.c
```diff
@@ -36,7 +36,7 @@ const char *cylinderuse_text[] = {
 	QT_TRANSLATE_NOOP("gettextFromC", "OC-gas"), QT_TRANSLATE_NOOP("gettextFromC", "diluent"), QT_TRANSLATE_NOOP("gettextFromC", "oxygen"), QT_TRANSLATE_NOOP("gettextFromC", "not used")
 };
 
-// For user visible text
+// For user visible text but still not translated
```
atdotde, line 39: Sure, one can do it that way. But what's wrong with (and I thought we had settled on that) making the _ui() version translated (not only "OC" to "Open Circuit" but to "Offenes System") by replacing the NOOP by translate(), so the translation lookup has to be done only once?

Apart from that aesthetic comment I like the four commits in this PR.

profile-widget/profilewidget2.cpp
dirkhh: I still don't understand why and where we'd want "modechange" to be translated. Can you explain why you are marking it for translation?

dirkhh (commented):
Now I'm confused.
If we only translate the strings when we show them, then why have the two arrays in the first place?
I thought the point of the _ui array was for it to hold the translated strings...

atdotde:
@dirkhh there are two levels of translation: In the xml and git files, the mode is called "OC" and "PSCR" while we decided that should be pronounced "Open circuit" and "pSCR" when shown to the user. So, in any case, there is some level of translation going on. @sfuchs79 what's to keep that separate from the actual language translation. But as I said, I would prefer what you expected as well, to make _ui[] have the actual translation in it (so change the NOOP to translate()).

dirkhh:
OK, thinking about the implications some more, I'm ok with this solution and will take it. This way I can push new strings right now.

sfuchs79:
@atdotde @dirkhh 
In fact also my first idea was to have the strings in the ..._ui[] already translated. But this simply didn't work but throw a compile error. I think it is not possible to INIT the char * array with values coming from the translate function or macro.
Maybe there would have been a solution but I didn't see it.

profile-widget/profilewidget2.cpp
sfuchs79: @dirkhh 
My understanding is that we translate it like other event names when we display it on the UI in the info box in the profile. Relevant code for the translation is here:
https://github.com/sfuchs79/subsurface/blob/master/profile-widget/diveeventitem.cpp#L149

I tested this right now and this part works:
grafik


BUT the dive mode names (here "open circuit") are still nowhere translated and I don't know why...

dirkhh: My guess is because I haven't pulled the translations :)

dirkhh: OK, translations pulled, try now

sfuchs79: No, that is not the root cause. I downloaded from Transifex and merged the German language file manually for my test.

But I fixed it and unfortunately another PR will follow. I once again was lost with all those different tr() "domains". We already once had this discussion before. If I want to access a string provided in a C file from a CPP file I need to use
`gettextFromC::instance()->tr();` instead of a simple `tr();`

PR #1289, opened 2018-05-17
Various small fixes

atdotde:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Various smaller bug fixes

### Changes made:

1) Make sure we translate dive mode names and use the same strings in all places
2) Rename a variable to better reflect its meaning
3) Don't try to compute a total SAC when a cylinder was used where we don't have sane start and end pressures
4) Planner: Only store a set point for CCR segments

### Mentions:
 This is mainly a reaction to suggestions by @sfuchs79 and @willemferguson

### DIFF
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -447,7 +447,8 @@ PlannerSettingsWidget::PlannerSettingsWidget(QWidget *parent, Qt::WindowFlags f)
 	disableDecoElements((int) prefs.planner_deco_mode);
 
 	// should be the same order as in dive_comp_type!
-	rebreather_modes << tr("Open circuit") << tr("CCR") << tr("pSCR");
+	for (int i=0; i < FREEDIVE; i++)
+		rebreather_modes.append(QString(divemode_text_ui[i]));
 	ui.rebreathermode->insertItems(0, rebreather_modes);
 
 	modeMapper = new QSignalMapper(this);
--- a/profile-widget/diveeventitem.cpp
+++ b/profile-widget/diveeventitem.cpp
@@ -194,7 +194,7 @@ void DiveEventItem::setupToolTipString(struct gasmix *lastgasmix)
 		}
 		*lastgasmix = *mix;
 	} else if (same_string(internalEvent->name, "modechange")) {
-		name += tr(": %1").arg(divemode_text[internalEvent->value]);
+		name += tr(": %1").arg(divemode_text_ui[internalEvent->value]);
 	} else if (value) {
 		if (type == SAMPLE_EVENT_PO2 && same_string(internalEvent->name, "SP change")) {
 			name += QString(": %1bar").arg((double)value / 1000, 0, 'f', 1);
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1424,21 +1424,21 @@ void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
 	QMenu *changeMode = m.addMenu(tr("Change divemode"));
 	if (divemode != OC) {
 		QAction *action = new QAction(&m);
-		action->setText("OC");
+		action->setText(divemode_text_ui[OC]);
 		connect(action, SIGNAL(triggered(bool)), this, SLOT(addDivemodeSwith()));
 		action->setData(event->globalPos());
 		changeMode->addAction(action);
 	}
 	if (divemode != CCR) {
 		QAction *action = new QAction(&m);
-		action->setText("CCR");
+		action->setText(divemode_text_ui[CCR]);
 		connect(action, SIGNAL(triggered(bool)), this, SLOT(addDivemodeSwith()));
 		action->setData(event->globalPos());
 		changeMode->addAction(action);
 	}
 	if (divemode != PSCR) {
 		QAction *action = new QAction(&m);
-		action->setText("PSCR");
+		action->setText(divemode_text_ui[PSCR]);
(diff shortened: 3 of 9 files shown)


atdotde:
Oops, we need to make sure, no translated values get written to the files. This needs updating. But not now. Will do this tomorrow.

atdotde:
OK, this wasn't that difficult.

profile-widget/profilewidget2.cpp
```diff
@@ -1424,21 +1424,21 @@ void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
 	QMenu *changeMode = m.addMenu(tr("Change divemode"));
 	if (divemode != OC) {
 		QAction *action = new QAction(&m);
-		action->setText("OC");
+		action->setText(tr("Open circuit"));
```
dirkhh: why wouldn't we use divemode_text[OC] or something like this - otherwise we have the same string for translation multiple times...

dirkhh (commented):
This works as is - I just wonder if we should avoid the double translation (and of course this turns into a divemode_text_ui[] after the last commit)

sfuchs79:
What is missing here compared to #1288:
- Change from "OC" to "Open circuit" in statistics.c and maintab.cpp
https://github.com/Subsurface-divelog/subsurface/pull/1288/files#diff-b2ac680bbcb4867dcc85c95b7171a13bR143
https://github.com/Subsurface-divelog/subsurface/pull/1288/files#diff-53425902966c23ff66293ccac9394931R112
- Translation for string "modechange"
https://github.com/Subsurface-divelog/subsurface/pull/1288/files#diff-75dcd3eb96f043b3dcbce84e1a0e8088R1615

desktop-widgets/diveplanner.cpp
```diff
@@ -447,7 +447,8 @@ PlannerSettingsWidget::PlannerSettingsWidget(QWidget *parent, Qt::WindowFlags f)
 	disableDecoElements((int) prefs.planner_deco_mode);
 
 	// should be the same order as in dive_comp_type!
-	rebreather_modes << tr("Open circuit") << tr("CCR") << tr("pSCR");
+	for (int i=0; i < FREEDIVE; i++)
+		rebreather_modes.append(QString(divemode_text_ui[i]));
```
sfuchs79, line 451: This imho this removes translation instead of adding it. `divemode_text_ui[i]` holds untranslated strings.

profile-widget/diveeventitem.cpp
```diff
@@ -194,7 +194,7 @@ void DiveEventItem::setupToolTipString(struct gasmix *lastgasmix)
 		}
 		*lastgasmix = *mix;
 	} else if (same_string(internalEvent->name, "modechange")) {
-		name += tr(": %1").arg(divemode_text[internalEvent->value]);
+		name += tr(": %1").arg(divemode_text_ui[internalEvent->value]);
```
sfuchs79, line 197: Doing `tr(": %1")` makes no sense to me. Not this string has to be translated but again the `divemode_text_ui` in the argument.

profile-widget/profilewidget2.cpp
```diff
@@ -1610,7 +1610,7 @@ void ProfileWidget2::addDivemodeSwith()
 	QAction *action = qobject_cast<QAction *>(sender());
 	QPointF scenePos = mapToScene(mapFromGlobal(action->data().toPoint()));
 	for (i = 0; i < UNDEF_COMP_TYPE; i++)
-		if (QString(divemode_text[i]) == action->text())
+		if (QString(divemode_text_ui[i]) == action->text())
```
sfuchs79, line 1613: Careful here - isn't this now comparing a untranslated string with an translated string?

sfuchs79 (changes requested):

dirkhh (changes requested):
This looks all good except for the commit message of the first commit... would you like me to clean that up for you or can you just fix that and force push?

janiversen:
I will fix it!

janiversen:
Sorry I was in the wrong PR :-)

willemferguson:
I propose that we accept the changes for merging.

sfuchs79:
What about the open findings from my side? I don‘t think that was ready to be merged.

atdotde:
Sorry for not commenting earlier but I think these patches are correct, maybe except for the translation of the white space punctuation only string that was there before: the _ui string is the translated to be shown to the user and thus also to be compared to the action element.

sfuchs79:
But the QT_TRANSLATE_NOOP doesn‘t realy translate the string. Its a NOOP as the Name says.

dirkhh:
The NOOP variation marks a string for translation, but at that spot in the code it stays untranslated. You can use that for example to initialize an array with the untranslated values while at the same time marking those strings for translation. You can then call tr() on the members of the array right as   you are about to display them, and since the strings were marked for translation they will be in the lookup table and will displayed in their translated form.

sfuchs79:
@dirkhh 
Yes, that is exactly also my understanding.
But if I look at the code of this PR I can't see the "final" tr() call to actually do the translation. And also from @atdotde comments I think Robert wanted to have the strings translated already in the string list. But to achieve this the NOOP has to be changed to a real translate.

dirkhh:
So let's either revery this or maybe even easier, add another PR on top that fixes it.

sfuchs79:
Double checking it with @atdotde and then fix it with another PR would be the better way.

atdotde:
Of course you are right. That should be translate instead of QT_TRANSLATE_NOOP. Do you want to do do a patch or shall I do it?

PR #1291, opened 2018-05-18
mobile: add "Copy log to clipboard" button

janiversen:
on iOS it is practically impossible to copy the App log
to e.g. a mail! in iOS 11 the log file is stored within
the subsurface container and you first need to copy (actually
using the clipboard) out from there to the "normal" document
shared space, before it can be used.

At least iOS users (and I believe Android users) are not really
used to work with files, so the process is not easy to document
in an understandable way.

The alternative is to provide a button, which simply puts the
log on the general clipboard, allowing it to be pasted in a
multitud of applications.

Documentation will be updated in a seperate commit.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

See commit text. It is important that users can easily mail the app log in case of problems,
the current solution (at least on iOS) is not at all user friendly.

### Changes made:

1) Added button to mobile about window
2) Added function in MessageHandlermodel to return the log as string
3) Added function in qmlmanager to respond to button.

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -12,12 +12,14 @@
 #include <QElapsedTimer>
 #include <QTimer>
 #include <QDateTime>
+#include <QClipboard>
 
 #include <QBluetoothLocalDevice>
 
 #include "qt-models/divelistmodel.h"
 #include "qt-models/gpslistmodel.h"
 #include "qt-models/completionmodels.h"
+#include "qt-models/messagehandlermodel.h"
 #include "core/divelist.h"
 #include "core/device.h"
 #include "core/pref.h"
@@ -330,6 +332,16 @@ void QMLManager::cancelCredentialsPinSetup()
 	setStartPageText(tr("Starting..."));
 
 	setShowPin(false);
+}
+
+void QMLManager::copyAppLogToClipboard()
+{
+	/*
+	 * The user clicked the button, so copy the log file
+	 * to the clipboard for easy access
+	 */
+	QString copyString =  MessageHandlerModel::self()->logAsString();
+	QApplication::clipboard()->setText(copyString, QClipboard::Clipboard);
 }
 
 void QMLManager::finishSetup()
--- a/qt-models/messagehandlermodel.cpp
+++ b/qt-models/messagehandlermodel.cpp
@@ -51,6 +51,15 @@ void MessageHandlerModel::addLog(QtMsgType type, const QString& message)
 #endif
 }
 
+const QString MessageHandlerModel::logAsString()
+{
+	QString copyString;
+
+	// Loop through m_data and build big string to be put on the clipboard
+	for(const MessageData &data: m_data)
+		copyString += data.message + "\n";
+	return copyString;
+}
 QVariant MessageHandlerModel::data(const QModelIndex& idx, int role) const
 {
 	switch(role) {
--- a/qt-models/messagehandlermodel.h
+++ b/qt-models/messagehandlermodel.h
@@ -14,6 +14,7 @@ public:
 	QVariant data(const QModelIndex& idx, int role) const override;
 	QHash<int, QByteArray> roleNames() const override;
 	void addLog(QtMsgType type, const QString& message);
+	const QString logAsString();
 
 	/* call this to clear the debug data */
 	Q_INVOKABLE void reset();
--- a/mobile-widgets/qml/About.qml
+++ b/mobile-widgets/qml/About.qml
@@ -56,5 +56,14 @@ Kirigami.ScrollablePage {
 			anchors.horizontalCenter: parent.Center
 			horizontalAlignment: Text.AlignHCenter
 		}
+		SsrfButton {
+			id: copyAppLogToClipboard
+			Layout.alignment: Qt.AlignHCenter
+			text: qsTr("Copy app log to clipboard")
+			onClicked: {
+				manager.copyAppLogToClipboard()
+				rootItem.returnTopPage()
+				}
+		}
 	}
 }
--- a/mobile-widgets/qmlmanager.h
+++ b/mobile-widgets/qmlmanager.h
@@ -178,6 +178,7 @@ public slots:
 	void clearGpsData();
 	void clearCredentials();
 	void cancelCredentialsPinSetup();
+	void copyAppLogToClipboard();
 	void finishSetup();
 	void openLocalThenRemote(QString url);
 	void mergeLocalRepo();


mfbernardes:
The log is also available via Files app. From there you can share, copy to another app etc.

mobile-widgets/qmlmanager.cpp
```diff
@@ -332,6 +334,19 @@ void QMLManager::cancelCredentialsPinSetup()
...
+{
+	/*
+	 * The user clicked the button, so copy the log file
+	 * to the clipboard for easy access
+	 */
+	QString copyString;
+	QClipboard *p_Clipboard = QApplication::clipboard();
+
+	MessageHandlerModel::copyLogToString(copyString);
+    p_Clipboard->setText(QString(copyString), QClipboard::Clipboard);
```
neolit123: whitespace issue on this line.

janiversen:
I know it is available, but it is very indirectly, since you cannot access the Subsurface container directly, so you first copy it to a shared place and then the file can be used. It might be different in Android, I have only tried in iOS.

mobile-widgets/qmlmanager.cpp
```diff
@@ -332,6 +334,19 @@ void QMLManager::cancelCredentialsPinSetup()
...
+void QMLManager::copyAppLogToClipboard()
+{
+	/*
+	 * The user clicked the button, so copy the log file
+	 * to the clipboard for easy access
+	 */
+	QString copyString;
+	QClipboard *p_Clipboard = QApplication::clipboard();
+
+	MessageHandlerModel::copyLogToString(copyString);
```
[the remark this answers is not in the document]
janiversen: not really, it was just easier write :-) I will make an update shortly (incl. the variable naming).

janiversen:
Screenshot as requested. 
simulator screen shot - iphone 6s - 2018-05-18 at 15 45 00

neolit123:
thanks! so last time i checked the log was visible under the `Developer`. do think it would make more sense to put the button there?

janiversen:
I thought of that, but "developer" is only visible if selected in "settings", so I opted against it. 

Btw. as you write the log is nicely available in developer for developers, and first time I needed a log, I tried to copy/paste that window (as a natural action), because I needed some function names. But it seems that is a facility that yet has to be implemented.

qt-models/messagehandlermodel.cpp
```diff
@@ -51,6 +51,15 @@ void MessageHandlerModel::addLog(QtMsgType type, const QString& message)
 #endif
 }
 
+void MessageHandlerModel::copyLogToString(QString &copyString)
+{
+	QVector<MessageData>::iterator i = MessageHandlerModel::self()->m_data.begin();
```
bstoeger: Seems overly complicated:
```
for(const MessageData &data: m_data)
    copyString += data.message;
```

qt-models/messagehandlermodel.cpp
```diff
@@ -51,6 +51,15 @@ void MessageHandlerModel::addLog(QtMsgType type, const QString& message)
 #endif
 }
 
+void MessageHandlerModel::copyLogToString(QString &copyString)
```
bstoeger: Please return the `QString`, don't use an out-parameter.

mobile-widgets/qmlmanager.cpp
```diff
@@ -332,6 +334,18 @@ void QMLManager::cancelCredentialsPinSetup()
...
+void QMLManager::copyAppLogToClipboard()
+{
+	/*
+	 * The user clicked the button, so copy the log file
+	 * to the clipboard for easy access
+	 */
+	QString copyString;
+
+	MessageHandlerModel::self()->copyLogToString(copyString);
+	QApplication::clipboard()->setText(QString(copyString), QClipboard::Clipboard);
```
bstoeger: `copyString` *is* a `QString`. Why would you cast it again to `QString`?

qt-models/messagehandlermodel.h
```diff
@@ -14,6 +14,7 @@ class MessageHandlerModel : public QAbstractListModel {
 	QVariant data(const QModelIndex& idx, int role) const override;
 	QHash<int, QByteArray> roleNames() const override;
 	void addLog(QtMsgType type, const QString& message);
+	void copyLogToString(QString& copyText);
```
bstoeger: See above:
```
    QString logAsString() const;
```

neolit123:
@janiversen 
ok, understood. so in terms of the location, i still think that exposing the button only with a "developer" context in mind is the correct approach instead of it being always visible in `About`.

cc @glance- @dirkhh 

thanks.

neolit123:
@janiversen 
> I tried to copy/paste that window (as a natural action), because I needed some function names. But it seems that is a facility that yet has to be implemented.

i think we had to make it `readonly` or something in those lines to make it possible to copy the text. it's QML specific.

janiversen:
@neolit123 
I politely disagree on the location, we need ordinary users to mail us the log, when they report problems. If this was only for developers this function would not be needed at all.

neolit123:
i'm not much of a smart phone user in general, so we are going to wait on @dirkhh and @glance- 
 to see what they think about the button location.

dirkhh:
It's fun to be 9-10 hours behind all of you :-)
First: thanks @janiversen - that's a really cool idea
@mfbernardes is of course correct - the file is available in the Files app and it's fairly easy to find there - but then you can't open it and you need to do some really unnatural things to get to the data (IIRC you "share" the file to the mail app and then can send it to yourself and THEN you can read it). So the idea of the clipboard is brilliant.
As for the location of the button... I think making this easy for the user is a good plan - I'm not opposed to doing this in the About screen.
For the NEXT pull request we should consider how to also make the libdivecomputer.log file available... maybe if it is there and was created during the same "run" of Subsurface-mobile, simply add it to the clipboard? I'm open to other ideas as well.
So far I only read the discussion - I'll look at the code in a moment (I see there were some comments)

qt-models/messagehandlermodel.cpp
dirkhh: Yeah, idiomatically I'd prefer that as well

janiversen:
Might be better you wait an hour, I am right now testing the changes suggested.

It should be easy to add the libdivecomputer.log as well, but in a new PR.

qt-models/messagehandlermodel.cpp
dirkhh: For silly reasons (which I am sure a better developer with more Qt/QML understanding than I have) the log is currently in a model... so I think this IS the easiest way to do it.
But maybe someone can simply turn this into a string that is displayed?
IIRC the problem was that I couldn't get this to "line break" and scroll the way I wanted and so I turned it into a view of a Model instead...

dirkhh (changes requested):
Really cool.
I agree with @bstoeger that the out-parameter feels odd, but other than that, this looks great to me.
THANKS

dirkhh:
> Might be better you wait an hour, I am right now testing the changes suggested.It should be easy to add the libdivecomputer.log as well, but in a new PR.

What I meant was "I'll review the commit" - I wasn't planning to implement my own suggestion. And yes, not a request to add this to THIS PR - just something else that would be nice to have since you are already fixing things in that area :-)

qt-models/messagehandlermodel.cpp
bstoeger: I wasn't commenting on the model thing. My point was that the loop can be written in a much simpler way.

janiversen:
Hope to changed the code as requested in the different reviews. Some of the changes are difference in code styles, some I will get used to as time goes by, so thanks for telling me.

Does by chance anybody know how to install QtQuick (without homebrew!) on a Mac ?

qt-models/messagehandlermodel.cpp
janiversen: I happend to agree with @bstoeger and have changed it. I am just a slave of my past, where this kind of iterator hiding was forbidden.

dirkhh:
This looks great to me... @bstoeger any more concerns from your side?

dirkhh (approved):

qt-models/messagehandlermodel.cpp
bstoeger: > I am just a slave of my past, where this kind of iterator hiding was forbidden.

That's an exceptionally peculiar policy.

bstoeger (approved):

PR #1292, opened 2018-05-18
Remove implicit conversion warnings from units.h and dive.h

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When generating iOS, there are 2.400 warnings, mostly from implicit conversions.
Just sometimes an implicit conversion is wrong, but will surely be overlooked in the
mist of warnings.

This PR removes the implicit conversion warnings originating in header files (and therefore
multiplying the number of warnings).

Having a function that returns an int but uses a long function to generate the int, calls for
an explicit cast, to inform other developers that this is intentionally.

### Changes made:

1) update units.h
2) update dive.h

NO functional changes.

### DIFF
--- a/core/dive.h
+++ b/core/dive.h
@@ -151,7 +151,7 @@ static inline int interpolate(int a, int b, int part, int whole)
 	/* It is doubtful that we actually need floating point for this, but whatever */
 	if (whole) {
 		double x = (double)a * (whole - part) + (double)b * part;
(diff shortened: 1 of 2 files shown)


dirkhh:
Waiting for Travis, but I'm ok with this change

PR #1293, opened 2018-05-18
changing Android build to use get-dep-lib.sh

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ x] Build system change
- [ ] Documentation change
- [ ] Language translation

This is the promised continuation of using get-dep-lib.sh in our different build.sh

Currently it is a test to see if travis are similar to my own environment.

### DIFF
--- a/packaging/android/build.sh
+++ b/packaging/android/build.sh
@@ -110,14 +110,7 @@ else
 fi
 
 # Which versions are we building against?
-SQLITE_VERSION=3190200
-LIBXML2_VERSION=2.9.4
-LIBXSLT_VERSION=1.1.29
-LIBZIP_VERSION=1.1.3
-LIBGIT2_VERSION=0.26.0
-LIBUSB_VERSION=1.0.20
 OPENSSL_VERSION=1.0.2l
-LIBFTDI_VERSION=1.3
 
 if [ "$ARCH" = "arm" ] ; then
 	QT_ARCH=armv7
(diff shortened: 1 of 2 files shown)


dirkhh:
We have a few people actively working on Android.
The old OpenSSL library is simply a matter of "treat assessment". The mobile app only connects to ONE site, one that I control, and that it has a hard coded certificate verification for.
As an attack vector I fail to see how this could be weaponized. Unless you consider ME the thread vector. In which case you have, err, other problems (since I build the binaries that we ship) :-)
That said, I'm happy to see newer versions used - it just never got high enough on my priority list.

dirkhh (changes requested):
Excellent... but I'm a bit unhappy about the sequence of commits. The first one adds things that are removed / replaced in the third, and the second one calls a script with an argument that isn't added until the third one.
so... can we rebase this, squashing the first and third commit together into a new first one, followed by the current second commit?

janiversen:
You wrote what I was just thinking :-) I am apart from security, just sensitive to issues of different versions, that means android could have errors that are solved elsewhere.

Anyhow I hope, me overloading travis and our GitHub is something positive. I have free cycles until end of June (then I will do PADI DM while I work as helper).

janiversen:
I am perfectly OK with squashing commit 1 and 3, but to be honest that superseds my git knowledge. I have now tried 2 times, and end up with wrong commits. If you can do it, please do it.

dirkhh:
YAY, learning opportunity!!!
```
git rebase -i HEAD~3
```
an editor with the three commits will open. Move the third line into second position. Replace the `pick` at the beginning with `s`. Save.
It will now squash the first and third commit and open an editor to edit / adjust the commit message (filled with the two commit messages concatenated).
Save
Done
Review in `git log` (or, nicer, `gitk` - but you need to install that which is a bit of a pain without Homebrew)
`git push -f` into the GitHub branch which will update this PR

janiversen:
@dirkhh 
thanks for the lesson :-) 
"save" was not correct it was "squash", and it seems to do what you wanted.
I did not know you could move the lines, that was the solution.

Lets see if travis is still happy.

janiversen:
@dirkhh seems travis is happy so now it is up to you.

dirkhh:
I meant "save" as in save the edited file... I realize that "s" stands for "squash" :-)
Glad this worked for you. Git is an amazing tool. And the fun of using git to develop Subsurface on Linux is not lost on me (those are the three projects that Linus started...)

dirkhh (approved):

PR #1294, opened 2018-05-18
core: solved PictureEntry defined as class and struct.

janiversen:
PictureEntry was defined as class in imagedownloader.h and
as struct in divepicturemodel.h

A class has a vptr in front, so the difference is real at least
for the clang compiler.

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

class or struct is another example of where a warning is important.

I had some "Interesting" debugging effects due to a vptr expected or too much.
 
Disclaimer: I am not as such doing a lot with the picture code (for now).
### Changes made:

1) changed forward declaration from "class" to "struct"

### DIFF
--- a/core/imagedownloader.h
+++ b/core/imagedownloader.h
@@ -23,7 +23,7 @@ private:
 	void saveImage(QNetworkReply *reply);
 };
 
-class PictureEntry;
+struct PictureEntry;
 class Thumbnailer : public QObject {
 	Q_OBJECT
 public:


dirkhh:
I'm ok with this change, but if I look at the errata for page 154 here http://www.gotw.ca/publications/xc++-errata.htm than Herb Sutter appears to disagree with you about this not being allowed. Which means that the implementation detail in clang (or the way lldb deals with this) may not be conforming with the standard.
Either way, once Travis finishes I'll merge this as it's certainly the right thing to do...

janiversen:
I raise my hat, you are surely more well versed in quoting sources than I am, I just look at my debugger.
(try do look at disassembly).

PR #1295, opened 2018-05-18
[RFC] Profile: Fix leak in animation

bstoeger:
If animDelete() was called with prefs.animation_speed == 0, the
object would not be marked for deletion, as opposed to calling
with prefs.animation_speed != 0.

While touching this function, use the function-pointer version
of connect().

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A small leak-fix, though I'm not 100% sure about this one. I'm not happy about Qt's diffuse ownership model with owner-less objects that schedule themselves for deletion by the main-loop, etc.

### Changes made:

1) Schedule object for deletion, even if animation speed is set to 0 (=infinity).
2) Use the function-pointer version of `connect` for compile-time checking.

### DIFF
--- a/profile-widget/animationfunctions.cpp
+++ b/profile-widget/animationfunctions.cpp
@@ -33,12 +33,12 @@ namespace Animations {
 	{
 		if (prefs.animation_speed != 0) {
 			QPropertyAnimation *animation = new QPropertyAnimation(obj, "opacity");
-			obj->connect(animation, SIGNAL(finished()), SLOT(deleteLater()));
+			obj->connect(animation, &QPropertyAnimation::finished, &QObject::deleteLater);
 			animation->setStartValue(1);
 			animation->setEndValue(0);
 			animation->start(QAbstractAnimation::DeleteWhenStopped);
 		} else {
-			obj->setProperty("opacity", 0);
+			delete obj;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


neolit123:
re: RFC
the only comment i can make is that i don't like `deleteLater()`. i've gotten weird behavior out of it many times - maybe it was misused, maybe not.

if we can avoid it - great! if we can't and if you think this patch works, we should just merge it.

dirkhh (commented):
Always happy to see conversions to the new `connect()` syntax.
I'll admit that `deleteLater()` makes me queasy as well, but if it's the right thing to do, then let's do it...

bstoeger:
I used deleteLater() for reasons of symmetry. But of course, in the first if-branch it has to be used, because you can't connect to delete directly. I will try if a plain delete works as well.

bstoeger:
According to my analysis, a plain `delete` is fine: the caller doesn't keep references to the deleted items. I could hit the delete by setting animation speed to 0 (strange value for what is actually infinity). No double-free detected.

I added a commit with a few minor coding-style fixes. The double/int compares made me itch. :-P Not perfectly happy about the `steps`/`stepsInRange` pair of variables, but I hope that from the code it is quite clear what they are.

dirkhh:
Thanks - this looks good, including the cleanups!

PR #1296, opened 2018-05-18
added explecit conversions to avoid warnings and consusion

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

changed implicit conversions to explicit to avoid confusion.

### DIFF
--- a/core/configuredivecomputerthreads.cpp
+++ b/core/configuredivecomputerthreads.cpp
@@ -173,7 +173,7 @@ static dc_status_t read_suunto_vyper_settings(dc_device_t *device, DeviceDetails
 	if (rc != DC_STATUS_SUCCESS)
 		return rc;
 	// in ft * 128.0
-	int depth = feet_to_mm(data[0] << 8 ^ data[1]) / 128;
+	int depth = (int)feet_to_mm(data[0] << 8 ^ data[1]) / 128;
 	m_deviceDetails->maxDepth = depth;
 	EMIT_PROGRESS();
 
@@ -269,7 +269,7 @@ static dc_status_t read_suunto_vyper_settings(dc_device_t *device, DeviceDetails
 	rc = dc_device_read(device, SUUNTO_VYPER_ALARM_DEPTH, data, 2);
 	if (rc != DC_STATUS_SUCCESS)
 		return rc;
-	depth = feet_to_mm(data[0] << 8 ^ data[1]) / 128;
+	depth = (int)feet_to_mm(data[0] << 8 ^ data[1]) / 128;
 	m_deviceDetails->alarmDepth = depth;
 	EMIT_PROGRESS();
 
@@ -2072,7 +2072,7 @@ static dc_status_t write_ostc_settings(dc_device_t *device, DeviceDetails *m_dev
 	//sync date and time
 	if (m_deviceDetails->syncTime) {
 		QDateTime timeToSet = QDateTime::currentDateTime();
-		dc_datetime_t time = { 0 };
+		dc_datetime_t time = { 0,0,0,0,0,0,0 };
 		time.year = timeToSet.date().year();
 		time.month = timeToSet.date().month();
 		time.day = timeToSet.date().day();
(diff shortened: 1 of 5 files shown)


core/configuredivecomputerthreads.cpp
```diff
@@ -173,7 +173,7 @@ static dc_status_t read_suunto_vyper_settings(dc_device_t *device, DeviceDetails
 	if (rc != DC_STATUS_SUCCESS)
 		return rc;
 	// in ft * 128.0
-	int depth = feet_to_mm(data[0] << 8 ^ data[1]) / 128;
+	int depth = (int)feet_to_mm(data[0] << 8 ^ data[1]) / 128;
```
dirkhh, line 176: this is odd... I wonder if `feet_to_mm()` should return an int instead... actually, this might be one of those "Windows is still a 32bit binary" issues. But no, that shouldn't be a problem here... definitely worth looking into

core/configuredivecomputerthreads.cpp
```diff
@@ -2072,7 +2072,7 @@ static dc_status_t write_ostc_settings(dc_device_t *device, DeviceDetails *m_dev
 	//sync date and time
 	if (m_deviceDetails->syncTime) {
 		QDateTime timeToSet = QDateTime::currentDateTime();
-		dc_datetime_t time = { 0 };
+		dc_datetime_t time = { 0,0,0,0,0,0,0 };
```
dirkhh, line 2075: at some point we had a compiler who didn't like this, but Travis passes, so this might be history.

dirkhh (commented):
Seems useful, but this is something where it would be nice to get a couple more reviews...

torvalds:
I detest all of these pointless casts, and the addition of pointless members to an initializer.

Is the IOS compiler a C compiler or some kind of other garbage? 

Please get rid of whatever warning flag it is that causes these to be flagged on iOS instead.

core/configuredivecomputerthreads.cpp
neolit123: > core/units.h:172:static inline unsigned long feet_to_mm(double feet)

`long` and `int` would be the same for `-m32` or around 4.3 million meters of depth, so `unsigned int` is fine too.

neolit123: also, all callers of `feet_to_mm()` should use a `unsigned` receiving variable to avoid possible signed / unsigned warnings. AFAIK, we don't handle negative depth or somehow elevation above sea level.

atdotde:
If you asked me I would try to get rid of all "unsigned" and make everything signed: We don't need the extra bit for the values but it avoids strange behaviour like

int a = -1;
unsigned int b = 1;
if (a > b)
     printf("Oops\n");

which prints since int is converted to unsigned and and not the other way 'round.

Now, there will be somebody who doubts the existence of negative depths, but they exist as in 

unsigned int target = 0;
int current_depth = 5;
int velocity = 2;

while (target < current_depth) {
    // Ascent
    current_depth -= velocity;
}

is actually an infinite loop.

Thus, a while ago, I went through the source trying to eliminate all "unsigned" where possible. Please don't reintroduce them again.

core/configuredivecomputerthreads.cpp
bstoeger: I'm don't know the context of these functions, but relative changes might very well be negative. A unit-conversion function taking a `double` and returning an `unsigned long` does seem a bit surprising.

janiversen:
Seems I have a tendency to cause discussions with my PR's, but at least positive ones.

I highly agree with @torvalds "pointless casts", however not acting on the warnings also seems wrong. Many casts arise from a poorly designed function, and a PR like this one get people to think if the function returns the right type.

The iOS compiler is clang C++, and I get exactly the same warnings when compiling for my Mac, so the platform is not the issue here.

I can of course change the build system to suppress, but do we really want that ?

dirkhh:
You get the same warnings on Mac? That seems odd - and must have changed very recently because we used to have a "near clean" compile on Mac.
And if I look at Travis for Mac (here's a random build log https://api.travis-ci.org/v3/job/380862564/log.txt) I don't see the thousands of warnings that I get on iOS.
My understanding is that iOS forces on pretty much every insane warning flag there is - which is why the iOS build script actually uses `sed` to turn some of them off as otherwise Travis fails as the build log surpasses 4M.
So... 
(a) I would like to understand why you get all these warnings on Mac builds, yet Travis and I don't
(b) I like the idea of revisiting some of these functions and data types and once again asking ourselves if they are the best choice (and let's do that on the mailing list to get people like Linus and Thiago involved who tend to be our top people on questions like this)
(c) I would love to see a PR that tries to get rid of the noise in the iOS builds by disabling the sillier warning flags everywhere - I just hacked out the ones that were so bad that they made Travis fail

PR #1297, opened 2018-05-19
Don't repopulate DivePictureModel on deletion of pictures

bstoeger:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Deletion of pictures repopulated the whole DivePictureModel, which looked ugly in the UI (all pictures in the profile plot regenerated).

Make this smarter, by starting the implementing of proper model-semantics. This is a bit hairy, so testing would be appreciated (@sfuchs79?)

Drag/drop and picture addition will be harder, therefore postpone until this works properly.
### Changes made:

1) On reset, don't do a `removeRows()`/`addRows()` pair.
2) Change the interface of `removePicture()` to not take a `last` parameter, but a list of pictures. This part actually fixes a bug.
3) Implement `removePicture` with proper model-semantics.
### Mentions:

@sfuchs79

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1117,6 +1117,7 @@ void ProfileWidget2::setProfileState()
 	connect(DivePictureModel::instance(), &DivePictureModel::dataChanged, this, &ProfileWidget2::updatePictures);
 	connect(DivePictureModel::instance(), SIGNAL(rowsInserted(const QModelIndex &, int, int)), this, SLOT(plotPictures()));
 	connect(DivePictureModel::instance(), SIGNAL(rowsRemoved(const QModelIndex &, int, int)), this, SLOT(plotPictures()));
+	connect(DivePictureModel::instance(), &DivePictureModel::modelReset, this, &ProfileWidget2::plotPictures);
 #endif
 	/* show the same stuff that the profile shows. */
 
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1116,7 +1116,7 @@ void ProfileWidget2::setProfileState()
 #ifndef SUBSURFACE_MOBILE
 	connect(DivePictureModel::instance(), &DivePictureModel::dataChanged, this, &ProfileWidget2::updatePictures);
 	connect(DivePictureModel::instance(), SIGNAL(rowsInserted(const QModelIndex &, int, int)), this, SLOT(plotPictures()));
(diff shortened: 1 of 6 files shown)


dirkhh (commented):
This code is just slightly outside of my comfort zone. I don't see anything that I think is wrong - but I would have to spend a lot more time going through it in context to be convinced that it's right.
@sfuchs79 and @atdotde  - you know this stuff better than I do, could you look and approve if you like it?

atdotde:
Hmm, have to admit I am not really fluent with all this model business. So have to look into this a bit more for me to be able to say anything substantial. But for marginal things: The plural of “index” is “indices”. And could you explain to me why you removed the SIGAL/SLOT macros from the connect call?

bstoeger:
> The plural of “index” is “indices”.

I would have thought so as well, but
1) It turns out that "indexes" is likewise valid: https://www.merriam-webster.com/dictionary/index
2) Qt seems to use "indexes": http://doc.qt.io/qt-5/qitemselectionmodel.html
3) The original code is not by me - and I try to avoid gratuitous variable renaming.

> And could you explain to me why you removed the SIGAL/SLOT macros from the connect call?

The pointer-to-member-function versions give us at least *some* compile-time checking, so they are an improvement over the SIGNAL/SLOT business. But note that the crucial point in these changes is the different signal/slot connections, not the syntax. We now have a proper "reset" slot and the "removeRows" slot does no regenerate all profile pictures.

This is of course only a start towards proper semantics - later proper addRows and moveRows slots will have to be implemented. But I haven't looked into the ordering issues yet. I guess the pictures are ordered by timestamp..?

profile-widget/profilewidget2.cpp
```diff
@@ -2086,6 +2087,17 @@ void ProfileWidget2::plotPictures()
 		item->setPos(x, y);
 	}
 }
+
+void ProfileWidget2::removePictures(const QModelIndex &, int first, int last)
+{
+	DivePictureModel *m = DivePictureModel::instance();
+	first = std::max(0, first - m->rowDDStart);
+	last = std::min((int)pictures.size(), last + 1 - m->rowDDStart);	// Note the +1 owing to Qt's idiosyncratic range concept.
```
neolit123: probably best to move the comment on a separate line.

neolit123 (approved):

dirkhh:
I guess we should merge this? Or are you waiting for a change for that long comment, @neolit123 ?

neolit123:
yep, only the comment.

bstoeger:
I was a bit torn between removing and keeping the comment.

In the end I decided to keep it in - even if it doesn't help, it will also not hurt.

PR #1298, opened 2018-05-19
iOS: supress warnings

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

Bumped iOS version to 10.0, because the Qt libs are compiled with iOS 10.0
meaning if Qt uses a 10.0 specific function ssrf will crash.

Supressed warnings from clang C/C++

### Changes made:

1) update .pro file to 10.0
2) remove AppICon-76
3) Update .pro file with -w flags for C/C++


dirkhh:
OK, so my claim of "we supported 8.0" was bogus all along.
Apologies. Waiting for Travis, but this looks great to me.

dirkhh:
the completed iOS build is certainly much easier to read.
I could have sworn that I had removed the two offending warning command line flags in libdivecomputer - need to figure out why those are back.

PR #1299, opened 2018-05-20
Only turn off specific warnings

atdotde:
Turning off all warnings was too radical. Specifically turn
off only those we decide not to care about.

In fact, with this option, we see that there are a few others that should be taken care of.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/packaging/ios/Subsurface-mobile/Subsurface-mobile.pro
+++ b/packaging/ios/Subsurface-mobile/Subsurface-mobile.pro
@@ -217,8 +217,8 @@ include(deployment.pri)
 include(../../../mobile-widgets/qml/kirigami/kirigami.pri)
 
 # Supress some warnings
-QMAKE_CXXFLAGS += -w
-QMAKE_CFLAGS   += -w
+QMAKE_CXXFLAGS += -Wno-shorten-64-to-32 -Wno-missing-field-initializers
+QMAKE_CFLAGS   += -Wno-shorten-64-to-32 -Wno-missing-field-initializers
 
 ios {
 	QMAKE_ASSET_CATALOGS += ../storeIcon.xcassets


janiversen:
hmm @torvalds specifically stated he did not like the Many zeroes in initialisers, and that warning is now back on ?

atdotde:
Better?

janiversen:
just tested your patch and I still get:
/Users/jani/develop/src/subsurface/core/configuredivecomputerthreads.cpp:2075: warning: missing field 'month' initializer [-Wmissing-field-initializers]
                dc_datetime_t time = { 0 };
                                         ^
I tested -Wno-missing-field-initializers before, and it seems clang ignores that, that was the reason why I gave up and used -w instead (if a scalpel does not work, a broad sword sure will).

I am also seeing:
/Users/jani/develop/src/subsurface/core/qtserialbluetooth.cpp:44: warning: unused parameter 'context' [-Wunused-parameter]
static dc_status_t qt_serial_open(qt_serial_t **io, dc_context_t *context, const char* devaddr)
                                                                  ^
Which seems to be something that needs fixing.

And
/Users/jani/develop/src/subsurface/core/dive.c:169: warning: comparison of integers of different signs: 'int32_t' (aka 'int') and 'unsigned int' [-Wsign-compare]
        while (*p && (*p)->time.seconds <= time)
                     ~~~~~~~~~~~~~~~~~~ ^  ~~~~
Which to me is a real problem (can case all kinds of funny problems)

Hope that helps.

janiversen:
Added a PR #1303 for the unused parameter.

neolit123:
@atdotde @janiversen 
please revert these 2 commits too if we are now using `-Wno-shorten-64-to-32`:
https://github.com/Subsurface-divelog/subsurface/pull/1292

dirkhh:
I think `-w` is a big hammer. So let's find the "right sized" hammer to trim this down to warnings that are actually worth looking into.
The conflict is caused by the file moving, so I think the best bet would be to simply close this PR and start with a fresh one based on master (after all, this is a tiny change)

janiversen:
I can do that tomorrow and also revert the 2 other commits.

atdotde:
Superseeded by #1299

PR #1302, opened 2018-05-20
build-system: updated iOS with small amendments

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ x] Build system change
- [ ] Documentation change
- [ ] Language translation

Cleaned up build.sh and Subsurface-mobile.pro

### Changes made:

1) Solved problem that app was generated (qmake_all was just a FORCE)
2) removed duplicate deployment.pri
3) removed ios/Subsurface-mobile (was just confusing)


packaging/ios/build.sh
```diff
@@ -241,4 +241,4 @@ cd build-Subsurface-mobile-Qt_$(echo ${QT_VERSION} | tr . _)_for_iOS-${DEBUGRELE
 ${IOS_QT}/${QT_VERSION}/ios/bin/qmake ../Subsurface-mobile/Subsurface-mobile.pro \
 	-spec macx-ios-clang CONFIG+=$TARGET CONFIG+=$TARGET2 CONFIG+=$DRCONFIG
 
-make qmake_all
+make all
```
dirkhh: This may have changed from when I first did this. The rationale was that this would get things set up so that I could then use Xcode to upload the app. Otherwise everything is built twice...

dirkhh (commented):
Moving things around like that will break all of my private tooling here to create the iTunes packages, but I'm sure that's easy enough to fix.

dirkhh:
The linux build fail with the AppImage not creating the zsync file is something I need to investigate. The linux2 seemed just hung, so I restarted it. But the iOS one I'd like for you to take a look at (FYI, I'll be offline for the next three hours...)

janiversen:
Sorry for breaking your private toolset, but I try to make things simpler and more understandable :-)

My goal seems to be slightly different from yours, I wanted make to build the app dirs (like it did earlier) so that I can use them directly. But I am sure we can find a compromise ?

dirkhh:
Don't worry - I can easily work around this here.
My setup is designed around uploading to iTunes, yours is designed for you to work locally. Those are of course different.

janiversen:
It fails with
ld: warning: directory not found for option '-L/Users/hohndel/Qt5.10.1/5.10.1/ios/lib'
ld: warning: directory not found for option '-L/Users/hohndel/Qt5.10.1/5.10.1/ios/lib'
ld: library not found for -lqiosnsphotolibrarysupport_debug

Seems like it is a bit out of my reach ?

dirkhh:
oh weird. Sorry - I blindly assumed that this was something that resulted from your change.
It looks like we are now doing a debug build on Travis? For build time reasons, I had this set up to just do a release build there... maybe that's what causes it.
The Qt package that we download is aggressively stripped of things to be smaller -> faster download and overall run time. As a result no debug libraries, only release...

janiversen:
I am sorry if I have changed something that causes a debug build. It should only happen when doing
build.sh -debug

dirkhh:
So if I look through the build log at https://travis-ci.org/Subsurface-divelog/subsurface/jobs/381303885#L5985 I see that in line 5985 the Release build succeeds and then we start a debug build. And that's what then fails.
And that's clearly triggered by the `make all` as you can see when you scroll back to https://travis-ci.org/Subsurface-divelog/subsurface/jobs/381303885#L4760

janiversen:
hmm so I need to find another tag, let me just reset the make for now, that way travis (and hopefully you) and I have a simple step more

dirkhh:
The weird thing is that the Travis build previously did call make after build.sh completed. So I'm kinda surprised why that worked and yours fails... Look at the end of `scripts/ios/travisbuild.sh`

dirkhh:
You seem inactive on IRC - might be easier to discuss that there?

dirkhh:
merged manually - thanks

PR #1304, opened 2018-05-21
build-system: iOS only turn off specific warnings

janiversen:
Turning off all warnings (-w) was too radical.

Only turning off specific warnings.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

Do not use a sledge hammer when a scalpel is sufficient.
Warnings are reduced a lot.

This PR replaces #1299 which of course have precedence if @atdotde wants to make an update.

### DIFF
--- a/packaging/ios/Subsurface-mobile.pro
+++ b/packaging/ios/Subsurface-mobile.pro
@@ -217,8 +217,8 @@ include(./deployment.pri)
 include(../../mobile-widgets/qml/kirigami/kirigami.pri)
 
 # Supress some warnings
-QMAKE_CXXFLAGS += -w
-QMAKE_CFLAGS   += -w
+QMAKE_CXXFLAGS += -Wno-shorten-64-to-32 -Wno-missing-field-initializers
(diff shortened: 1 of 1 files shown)


atdotde:
Like it.

janiversen:
We have a number of sign-compare warnings, turnings those off seems dangerous to me ?

Like e.g.:
/Users/jani/develop/src/subsurface/core/uemis-downloader.c:1358: warning: comparison of integers of different signs: 'uint32_t' (aka 'unsigned int') and 'int' [-Wsign-compare]
        dive_to_read = mindiveid < first ? first - mindiveid : first;
                       ~~~~~~~~~ ^ ~~~~~

I would prefer using explicit cast here or even better correct the problem (without introducing more unsigned variables!!!).

wdyt ?

dirkhh:
Yes, I think looking carefully at maybe making a few variables signed might not be a bad initiative. Let's keep those warnings. The rest is fine with me.

PR #1305, opened 2018-05-21
profile-widget: Unused parm

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Added Q_UNUSED in 2 functions/files to silence warning about unused parameters.

It is done as 2 commits, since it is in 2 non-related files.

### DIFF
--- a/profile-widget/divepixmapitem.cpp
+++ b/profile-widget/divepixmapitem.cpp
@@ -23,7 +23,7 @@ CloseButtonItem::CloseButtonItem(QGraphicsItem *parent): DivePixmapItem(parent)
 	setFlag(ItemIgnoresTransformations);
 }
 
-void CloseButtonItem::mousePressEvent(QGraphicsSceneMouseEvent *event)
+void CloseButtonItem::mousePressEvent(QGraphicsSceneMouseEvent *)
 {
 	qgraphicsitem_cast<DivePictureItem*>(parentItem())->removePicture();
 }
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -532,6 +532,8 @@ void ProfileWidget2::plotDive(struct dive *d, bool force, bool doClearPictures)
 #ifndef SUBSURFACE_MOBILE
 	QTime measureDuration; // let's measure how long this takes us (maybe we'll turn of TTL calculation later
 	measureDuration.start();
+#else
+	Q_UNUSED(doClearPictures);
 #endif
 	if (currentState != ADD && currentState != PLAN) {
 		if (!d) {


profile-widget/profilewidget2.cpp
```diff
@@ -528,6 +528,7 @@ void ProfileWidget2::resetZoom()
 // Currently just one dive, but the plan is to enable All of the selected dives.
 void ProfileWidget2::plotDive(struct dive *d, bool force, bool doClearPictures)
 {
+	Q_UNUSED(doClearPictures);
```
bstoeger: Not correct for desktop -> this parameter is used.

profile-widget/divepixmapitem.cpp
```diff
@@ -25,6 +25,7 @@ CloseButtonItem::CloseButtonItem(QGraphicsItem *parent): DivePixmapItem(parent)
 
 void CloseButtonItem::mousePressEvent(QGraphicsSceneMouseEvent *event)
 {
+	Q_UNUSED(event);
```
bstoeger: In this case perhaps it would be less noisy to simply omit the parameter name as in:
```
void CloseButtonItem::mousePressEvent(QGraphicsSceneMouseEvent *)
```
@dirkhh: Is there a guideline concerning unused parameters? I see at least 3 versions in the codebase:
1) Q_UNUSED(param)
2) (void)param;
3) Omission of name

Since we already depend on 2) and 3), is there a point in using Qt's supposedly platform-independent version?

profile-widget/profilewidget2.cpp
janiversen: I know that, but I thought it was pure overkill to add a #ifdef IOS for just reducing noise. I can of course do that if so wished

profile-widget/divepixmapitem.cpp
neolit123: `(void)param` is the way of doing it outside of Qt - e.g. in C.
but also ... `#define Q_UNUSED(x) (void)x;`:
https://github.com/qt/qtbase/blob/5.11/src/corelib/global/qglobal.h#L117

both `Q_UNUSED` and omitting are fine here, omitting is better as it saves a line.

profile-widget/profilewidget2.cpp
neolit123: i'm fine with the `Q_UNUSED` being added without an `#ifdef` even if it's only the case on `SUBSURFACE_MOBILE`.

neolit123: you can also do the following near the prologue:
```
{
#ifndef SUBSURFACE_MOBILE
	QTime measureDuration; // let's measure how long this takes us (maybe we'll turn of TTL calculation later
	measureDuration.start();
#else
	Q_UNUSED(doClearPictures);
#endif
	static bool firstCall = true;
```

profile-widget/divepixmapitem.cpp
janiversen: @dirkhh is we decide for one standard e.g. 3. where possible else 2., I volunteer to do code cleanup, since it gives me a reason to visit other code parts.

profile-widget/profilewidget2.cpp
janiversen: Of course I must have been sleeping (it was before my morning coffee) all corrected.

neolit123 (approved):

dirkhh:
I like Q_UNUSED() because it's so explicit.
Redefining Q_UNUSED() if it is undef in one of our headers and using that consistently would help to use it everywhere...
Of course I just learned from you that dropping the variable name is a better solution (do I look like an idiot if I admit that I didn't know that worked?) - so let's go for that as the preferred solution.

In the meantime, this looks good.

janiversen:
I will have a look at 
1) where Q_UNUSED is used, but should be "remove parameter name"
2) where (void)parm is used, and change to Q_UNUSED or "remove parameter name"
a nice little task !

neolit123:
omission of function arguments is C++ only.

we shouldn't use `Q_UNUSED` in .c files, as those are not bound to Qt.
maybe define a separate macro `UNUSED` for them.
```
#define UNUSED(x) (void)x;
```

dirkhh:
@neolit123 
thanks - at least now I know why I didn't know about the omission of argument names... I'm a C programmer at heart :-)
as for Q_UNUSED vs UNUSED... meh. Either works for me. It's not like we are going to port Subsurface to a different UI library. What a preposterous idea...
(oops)

PR #1306, opened 2018-05-21
mobile: make libdivecomputer.log available on the clipboard together with subsurface.log

janiversen:

- [ ] Bug fix
- [x ] Functional change
- [ x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

libdivecomputer.log is always written, in order for users to simpler report problems.

When user presses "copy log(s) to clipboard" both logs are copied.

### Changes made:

1) removed active dive log in settings.
2) store libdivecomputer in documents on iOS (to make it accessible)
3) updated text in about:
captura de pantalla 2018-05-21 a las 11 17 03
4) in action read file, append to clipboard, with headings.

### DIFF
--- a/mobile-widgets/qml/DownloadFromDiveComputer.qml
+++ b/mobile-widgets/qml/DownloadFromDiveComputer.qml
@@ -34,7 +34,7 @@ Kirigami.Page {
 		deviceData.deviceId : 0
 		deviceData.diveId : 0
 		deviceData.saveDump : false
-		deviceData.saveLog : false
+		deviceData.saveLog : true
 
 		onFinished : {
 			importModel.repopulate()
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -144,15 +144,16 @@ QMLManager::QMLManager() : m_locationServiceEnabled(false),
 	timer.start();
 	connect(qobject_cast<QApplication *>(QApplication::instance()), &QApplication::applicationStateChanged, this, &QMLManager::applicationStateChanged);
 
-	QString libdcLogFileName = QStandardPaths::standardLocations(QStandardPaths::GenericDataLocation).first() + "/libdivecomputer.log";
-	logfile_name = copy_qstring(libdcLogFileName);
 #if defined(Q_OS_ANDROID) || defined(Q_OS_IOS)
 #if defined(Q_OS_ANDROID)
 	appLogFileName = QStandardPaths::standardLocations(QStandardPaths::GenericDataLocation).first() + "/subsurface.log";
+	QString libdcLogFileName = QStandardPaths::standardLocations(QStandardPaths::GenericDataLocation).first() + "/libdivecomputer.log";
 #elif defined(Q_OS_IOS)
 	// on iOS we should save the data to the DocumentsLocation so it becomes accessible to the user
 	appLogFileName = QStandardPaths::standardLocations(QStandardPaths::DocumentsLocation).first() + "/subsurface.log";
+	QString libdcLogFileName = QStandardPaths::standardLocations(QStandardPaths::DocumentsLocation).first() + "/libdivecomputer.log";
 #endif
+	logfile_name = copy_qstring(libdcLogFileName);
 	appLogFile.setFileName(appLogFileName);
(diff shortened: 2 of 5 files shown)


mobile-widgets/qml/DownloadFromDiveComputer.qml
```diff
@@ -34,7 +34,7 @@ Kirigami.Page {
 		deviceData.deviceId : 0
 		deviceData.diveId : 0
 		deviceData.saveDump : false
-		deviceData.saveLog : false
+		deviceData.saveLog : true
```
dirkhh, line 37: since this is no longer optional, why not simply remove that parameter?

mobile-widgets/qmlmanager.cpp
```diff
@@ -144,15 +144,16 @@ QMLManager::QMLManager() : m_locationServiceEnabled(false),
...
 #if defined(Q_OS_ANDROID) || defined(Q_OS_IOS)
 #if defined(Q_OS_ANDROID)
 	appLogFileName = QStandardPaths::standardLocations(QStandardPaths::GenericDataLocation).first() + "/subsurface.log";
+	QString libdcLogFileName = QStandardPaths::standardLocations(QStandardPaths::GenericDataLocation).first() + "/libdivecomputer.log";
 #elif defined(Q_OS_IOS)
 	// on iOS we should save the data to the DocumentsLocation so it becomes accessible to the user
 	appLogFileName = QStandardPaths::standardLocations(QStandardPaths::DocumentsLocation).first() + "/subsurface.log";
+	QString libdcLogFileName = QStandardPaths::standardLocations(QStandardPaths::DocumentsLocation).first() + "/libdivecomputer.log";
 #endif
+	logfile_name = copy_qstring(libdcLogFileName);
```
dirkhh, line 156: Since this is now only done for Android or iOS, Subsurface-mobile on desktop might end up behaving in uncool ways, right? There's no longer a filename set for it - and no way to set it.

dirkhh (changes requested):
One suggestion and one request:
(a) that parameter the user can no longer change should just go away?
(b) make sure this still works on Mac and Linux

mobile-widgets/qml/DownloadFromDiveComputer.qml
janiversen: A good idea, will update that.

mobile-widgets/qmlmanager.cpp
janiversen: Actually no, the filename is libdivecomputer.log, and there was no way before to change the name.

I have on purpose not removed the files, so both libdivecomputer.log and subsurface.log are still produced.

dirkhh: Ah, so it just gets overwritten for iOS/Android, but otherwise it still works. Cool - I was afraid that it would be undefined in Subsurface-mobile with your change when compiling on Linux/Mac

janiversen: As far as I can see the same happens for the desktop version, but I might be mistaken (have not tested it, just looked at where the open statement is).

I am doing a lot to ensure that none of my patches break on Linux/Mac, because theoretically mobile could grow into something useful on these platforms as well :-)

mobile-widgets/qml/DownloadFromDiveComputer.qml
janiversen: on second thought (and looking in code), bad idea, deviceData is shared with the desktop version, so I would need to introduce a number of #ifdef MOBILE. I think this solution is cleaner.

janiversen: And I have tested both desktop and mobile on Mac, which works.

dirkhh: OK, so my review was utter nonsense.
Glad I'm able to provide value here :-(

dirkhh (approved):

mobile-widgets/qml/DownloadFromDiveComputer.qml
janiversen: no no, you made me think and argue for my case, that is not at all a bad idea :-)

PR #1308, opened 2018-05-21
iOS: qt-ble.cpp remove unsupported event.

janiversen:
QEvent::DeferredDelete is not supported in iOS

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

QEvent::DeferredDelete is not supported in iOS so adding #ifdef 

### Changes made:

1) added #ifdef

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -38,7 +38,9 @@ static void waitFor(int ms) {
 
 	do {
 		QCoreApplication::processEvents(QEventLoop::AllEvents, ms);
+#if !defined(Q_OS_IOS)
 		QCoreApplication::sendPostedEvents(nullptr, QEvent::DeferredDelete);
+#endif
 		QThread::msleep(10);
 	} while (timer.elapsed() < ms);
 }


dirkhh:
So did this get us a warning at run time? How did this manifest?

janiversen:
When downloading from a DC (in my case the EON CORE) the log contains plenty of these warnings. The user does not see anything.
For me it is a "log cleanup" so that whenever the log shows a problem, we know it is a new problem.

janiversen:
Of course,

1. isn´t needed is not correct, Qt tells deferredDelete is not implemented in iOS (was in the app log earlier).
2. It is implemented at least on mac.

Normally the call processEvents should be enough, I am not sure why a deferredDelete in posted, that is not something I have seen before. I assumed someone had a special reason to post the deferredDelete.

I agree it does not look clean, but I think someone with linux and bluetooth needs to twst and secure there are no problems if the sendPost is removed.

bstoeger:
What I get from a quick search is that the construct
```
	QCoreApplication::sendPostedEvents(nullptr, QEvent::DeferredDelete);
```
is used to delete objects that have been marked for deletion with `deleteLater()`.

I see no `deleteLater()` in anything related to the BT code, therefore I believe that this line should be removed unconditionally.

janiversen:
Your search is correct deferredDelete is a signal to cleanup deleteLater calls. Remark they may have been deleted earlier.

I was not sure if e.g. the kirami code had some deleteLater.

PR #1310, opened 2018-05-21
mobile: get dive event on profile.

janiversen:
print mode was used to limit the functionality of the profile,
when used in ssrf-mobile.

The effect is that DC events are displayed, but not selectable

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When using your mobile to upload from DC, it is nice to see if there were critical events
in your dive.

This is the first step in making profile more mobile user friendly (zoom/move/tooltips is
on the drawing board).

### Changes made:

1) disabled print mode for profile in ssrf-mobile, no change for desktop

### DIFF
--- a/mobile-widgets/qmlprofile.cpp
+++ b/mobile-widgets/qmlprofile.cpp
@@ -15,7 +15,7 @@ QMLProfile::QMLProfile(QQuickItem *parent) :
 	setAntialiasing(true);
 	m_profileWidget = new ProfileWidget2(0);
 	m_profileWidget->setProfileState();
-	m_profileWidget->setPrintMode(true);
+	m_profileWidget->setPrintMode(false);
 	m_profileWidget->setFontPrintScale(0.8);
 	connect(QMLManager::instance(), &QMLManager::sendScreenChanged, this, &QMLProfile::screenChanged);
 	setDevicePixelRatio(QMLManager::instance()->lastDevicePixelRatio());


neolit123:
i think we had a good reason to do that related to fonts:
9df3835035f016231d02e95968c02bd5f100ee1a

cc @dirkhh

janiversen:
Tested on my Mac (ssrf-mobile), my iPad Pro and my little iPhone, no problems here...but do not know about android.

dirkhh:
So this was in 9df3835035f016231d02e95968c02bd5f100ee1a and the logic was that we are just rendering the profile - you get no interactivity anyway - and that without this on many displays the font sizes were out of whack.
What additional capability do you think this will give us? The profile shown is currently not a live profile, it's just a pixmap of one...

janiversen:
I was really just confused that doPrint was used in the widget to limit functionality, that looked very wrong.

I want to see the DC events, and then slowly add e.g. tooltips on the events.

If activating print is a problem then I will restore that and do a couple of if in the widget itself.

dirkhh:
So let's talk a bit about the architecture here.
What you are seeing on the screen in Subsurface-mobile is NOT the widget. You are seeing the widget being printed to a pixmap, and that static pixmap is then scaled and shown on the QML page.
I understand that there is brand new code in Qt (I don't recall if this has landed in 5.10 or is scheduled for 5.11 - @tcanabrava might know) that would allow us to actually show the widget there - which will then give you the option for all the other cool things you want to do. But right now, I think our architecture doesn't allow for it.

janiversen:
ok, but actually for now I have. what I wanted. You download your dc to your mobile and can imidiatly see if the dc have issued warnings



so lets see what the future brings.

but again if you worry abut the fonts I will change that back and do a couple of if instead.


Enviado desde mi iPad

dirkhh:
Let's do this and see what breaks :-)

dirkhh:
we can always revert it

janiversen:
mail and dile does not work, lets try this

imagen png

PR #1312, opened 2018-05-22
Last round of Q_UNUSED / UNUSED updates

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

cleanup how to silence not used parameters

added ssrf.h as a global product header in contrast to
dive.h that are the global dive header.

ssrf.h should slowly be extended with common macros, includes etc.
for now it contains UNUSED(x)

added a #pragma to silence missing initialisers, due to a clang bug.

with this are source should be consistent (apart from the few I have not catched).

### Changes made:

1) replaced (void)x with Q_UNUSED(x) or remove parameter name in cpp files
2) added ssrf.h
3) replaced (void)x with UNUSED(x) in C files

### DIFF
--- a/core/cochran.c
+++ b/core/cochran.c
@@ -4,6 +4,7 @@
 #pragma clang diagnostic ignored "-Wmissing-field-initializers"
 #endif
 
+#include "ssrf.h"
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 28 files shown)


janiversen:
I totally agree. Currently dive.h includes a small mountain of includes, how you you feel if I moved (most of) these to ssrf.h, maybe there a couple of other generic ones which could be moved too.

And then include ssrf.h in just about every c/cpp file ?

dirkhh:
I don't want to go from "everything includes dive.h" to "everything includes these four shorter .h files"
I'd love to see some more logical structure in our include files. We have the various helper includes. It would be nice to really sort them logically and to thin out the mountains of things we always include in every file, regardless of what we actually need.

My first run to remove the string functions from dive.h was made with the help of a little tooling. I wrote a script that removed dive.h file by file and ran a build. If it failed, it added just subsurface-string.h, if that failed it added just dive.h, if that still failed, it added both of them.

From a build time perspective it would be nice to be more careful with what's included when.

PR #1313, opened 2018-05-22
core: qt-ble.cpp remove postEvent for all platforms

janiversen:
the postEvent is only called when downloading from a dc
with bluetooth, so in most it does not have an effect
on the deleteLater() in the code.

there are no reason to do special cleanup while waiting
for bluetooth

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

as discussed in #1308 there are no need for a postEvent to delete "deleteLater", since 
this is only a bluetooth loop, and postEvent are not called elsewhere.

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -38,9 +38,6 @@ static void waitFor(int ms) {
 
 	do {
 		QCoreApplication::processEvents(QEventLoop::AllEvents, ms);
-#if !defined(Q_OS_IOS)
-		QCoreApplication::sendPostedEvents(nullptr, QEvent::DeferredDelete);
-#endif
 		QThread::msleep(10);
 	} while (timer.elapsed() < ms);
 }


bstoeger:
I think this is correct, but I haven't had time to test this.

bstoeger (approved):

dirkhh:
This code came from Alex @ablasche when he helped us correctly use Qt BLE support (which he is the maintainer of). So before we rip this out, I'd love to hear from him :-)

bstoeger:
FWIW, no difference what so ever for me on Linux.

I understand that the changes in 5faa167e9fdab0e0c9b1a24346440c3006659c66 will be reverted before release? Because it produces funky error messages.

ablasche:
Let's start with a disclaimer, I am not the expert on event loops ;) However there are two things to consider:

1.) The to-be-removed code is still valid implementation even for Qt 5.11. QTestlib uses sth similar:

http://code.qt.io/cgit/qt/qtbase.git/tree/src/testlib/qtestsystem.h

You may even convert QElapsedTimer to QDeadlineTimer which wasn't around when I added this code. The ifdef is not necessary because Qt internals take care of any potential "not supported" platforms. Hence ifdef merely bloats the code line in subsurface. 

2.) The question is to what extent deleteLater() plays a role while this event loop is spinning. If you have a very good understanding that no deleteLater() calls are happening you may get away removing the line altogether. You run the risk though that code added later on may introduce deleteLater() calls. Since the connection between this code and the new one is not obvious, you may get hidden problems.

If the motivation for this change is to merely cleanup, then I recommend to keep it on the grounds that Qt itself does it and subsurface wanting to play safe (if deleteLater() may be introduced). If the reason goes beyond a cleanup scope I would like to hear what actual issue is before making a recommendation.

janiversen:
Well without the ifdef our log gets filled with “not supported in iOS”. that was the reason I first removed it for iOS. We slready have deleteLater calls in the general code, without this postevent. 

You compare with qt test code, but forget one important difference, we have a global  (in  qt) eventhandler running as every app has. The test system is specically build without a global handler so the test program can control the sequence.

bstoeger:
I have checked all `deleteLater()`s. Unless I'm missing something, none of them is active during BLE busy-looping. I think we should use `deleteLater()` only under very specific circumstances. And even if we use it, we should not rely on the destructor being called at a specific time (that sounds like a recipe for disaster).

My understanding was that the `sendPostedEvents` thing does not work for iOS (or does it do some magic we're not aware of?). Therefore, we cannot rely on `deleteLater()` in this part of the code anyway. I'd much rather have all platforms behave identically than having obscure bugs only on iOS (or vice-versa). With all this in mind, I believe we should remove the call for now.

janiversen:
Just to secure there are no mistake. sendPostEvents works in general and are used frequently. It is the parameter that is unsupported.
I agree with @bstoeger having identical code is of great value both for maintenence and test.

PR #799, opened 2017-11-11
[Please test] Save Bluetooth devices to preferences

bstoeger:
### Describe the pull request:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Currently, the BT device selector starts with an empty list, which means that the user always has to click on scan. This PR tries to starts the scan of the BT device selection dialog on creation, power on or change of local device. Moreover, an appropriate device is automatically selected according to priority:
1) Previous device
2) BT name corresponding to selected dive computer
3) BT name recognized as dive computer
4) Any paired device

Thus, an accidental raising of the dialog can immediately be dismissed by pressing enter.

This is WIP, which currently has only been tested on Linux.

### Changes made:

1) Start scan if dialog is initialized, power is turned on or local device is changed.
2) Add a scan status field with the states "invalid" (powered off), "idle" (powered on, but not scanning) or "scanning"
3) If scanning, allow the user to abort the scan.
4) Automatically select detected devices according to priority.

### Documentation change:
the BT device selector list is auto-filled.

PR #1314, opened 2018-05-22
core: btdiscovery, remove rediscovery() in constructor

janiversen:
The btdiscovery object is created during startup, doing
a discovery at the point in time is prolonging the startup
time which is long enough (15sec on an iPhone).

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Doing a discovery when creating the btdiscovery class makes the startup phase
(at least on an iPhone) longer, and there are no need to do it, since we do not
know if the user will connect to a dc with bluetooth.

### Changes made:

1) removed call of rediscovery in btdiscovery constructor.

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -72,9 +72,6 @@ BTDiscovery::BTDiscovery(QObject*) : m_btValid(false),
 		return;
 	}
 	m_instance = this;
-#if defined(BT_SUPPORT)
-	BTDiscoveryReDiscover();
-#endif
 }
 
 void BTDiscovery::BTDiscoveryReDiscover()


dirkhh:
So the initial idea here was that this would ensure that the dropdown is populated with connections when you switch to the download screen. And even better, it also selects the correct dive computer for you :-)
How does this work with this call removed? Could this be done in the background?

janiversen:
It cuts 2 seconds start time on my iPhone, when bluetooth is not enabled (something Qt does not support checking for, but iOS does).

We are not at all ripping his code out, just not in the constructor, please. There is a PR #799 about saving selected bt to preferences, that is the way to go. Just an example I used my iPhone after diving today, to show the other divers why the should not buy software, but support us.

It seems a bit as if the PR #799, is about something else. I planned to make a simple file local to save the last selection (it cannot be saved in the normal prefs. since they might be used across devices).

dirkhh:
>It cuts 2 seconds start time on my iPhone, when bluetooth is not enabled (something Qt does not support checking for, but iOS does).

2 seconds isn't much - especially if in return the experience when switching to the download screen is significantly worse.

> We are not at all ripping his code out, just not in the constructor, please. 

Wait, now I'm confused - the comment about Alex was in a different PR - this is code that I think I have originally written :-)

> There is a PR #799 about saving selected bt to preferences, that is the way to go. 

I agree in principle, but IIRC on macOS and iOS this doesn't work because the UUIDs are not constant, I believe you actually have to scan before you are able to use them. We still can of course try and remember which dive computer was used last, just I don't think (on iOS) we can get away with not scanning.

> It seems a bit as if the PR #799, is about something else. I planned to make a simple file local to save the last selection (it cannot be saved in the normal prefs. since they might be used across devices).

You should save in preferences - they are device local. Just a couple of things are stored in git in order to share across devices (units, autogroup, a couple more)

janiversen:
I cannot talk about other os than iOS, but here the id stays the same until you unpair the dc, then it changes. This is actually I see used in other apps.

But please try to load master on an iphone, deactivate bluetooth and start ssrf.....you will see
1) launch screen
2) a dialog box asking you to activate bt (which you do not want to do, e.g. due to battery usage).
3) watch the black screen for several seconds

now do the same with bt active, and tell me if that is a good user experience.

janiversen:
using the official insight
https://github.com/Subsurface-divelog/subsurface/graphs/contributors

janiversen:
wrong pr

janiversen:
I suspect it is better not to push for this change, and instead hope the community agrees with #1326, and help me do something about it.

PR #1317, opened 2018-05-23
Fix includes for building against Qt-5.11.0

vesath:
This allows building against Qt-5.11.0.


neolit123:
thanks, the change looks good.
merging this as the CI is happy.

PR #1318, opened 2018-05-23
add get-dep-lib.sh to mobilecomponents.sh

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ x] Build system change
- [ ] Documentation change
- [ ] Language translation

Updated mobilecomponents.sh to use get-dep-libs.sh and while at 
removed the -nopull option

mobilecomponents copies kirigami for no special reason, however this
is left to another PR (for a rainy day).

mobilecomponents use a specific commit as base point, it would be more logical to
use the tag "v2.2.0", , however this
is left to another PR (for a rainy day).

### Changes made:

1) get-dep-lib add support for kirigami and breeze-icons
2) integrate get-dep-lib in mobilecomponents
3) update .gitignore with extra dirs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,9 +38,8 @@ mobile-widgets/qml/kirigami
 packaging/ios/translations/
 packaging/ios/googlemaps*
 packaging/ios/install-root*
-packaging/ios/libdivecomputer-*
-packaging/ios/libgit2*
-packaging/ios/libxslt*
-packaging/ios/libzip*
+packaging/ios/lib*
+packaging/ios/kirigami
+packaging/ios/breeze-icons
 packaging/ios/Info.plist
 packaging/ios/Qt
--- a/scripts/get-dep-lib.sh
+++ b/scripts/get-dep-lib.sh
@@ -13,6 +13,8 @@ CURRENT_XSLT="v1.1.29"
 CURRENT_SQLITE="3190200"
 CURRENT_LIBXML2="v2.9.4"
 CURRENT_LIBFTDI="1.3"
+CURRENT_KIRIGAMI="70c025ef6f6dc63c85180867f70f5e00ba5a8dba"
+CURRENT_BREEZE_ICONS=""
 
 # deal with all the command line arguments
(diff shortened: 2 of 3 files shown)


janiversen:
Now this is funny, the commit we use for kirigami is after v2.2.0 but before v5.37.0 so we are somewhere in the middle.

Newest release of kirigami is v5.46.0 maybe it is really time, to do a major update here.

mfbernardes:
Having Kirigami and breeze-icons listed in BUILD_COMMON makes them be downloaded twice.

build.sh calls get-up-libs and download them to packaging/ios, then mobilecomponents.sh gets out of subsurface source tree and downloads again (and finally uses them).

With the way things are now, just removing them from common fix this behavior, but feel free to do it differently if you prefer ;)
-BUILD_COMMON="libzip libgit2 googlemaps kirigami breeze-icons"
+BUILD_COMMON="libzip libgit2 googlemaps"

mfbernardes:
I was working on Kirigami update, but I could see no improvement AND it added some useless buttons on the application header. Since I could not figure out yet how to get rid of these buttons I didn't push anything yet.

FYI: v5.46.0 doesn't build (.pri missing files). Current master builds fine, seems to work fine, but the useless buttons are annoying.

janiversen:
ok, I will do that in a follow up patch. but just for the record, I merely replaced git clone statements already in mobilecomponents so I suspect this is an old problem.

mfbernardes:
It's new, because previously Kirigami and breeze-icons were downloaded only by mobilecomponents.sh.

Now it's downloaded by build.sh and mobilecomponents.sh.

That's why I suggested remove it from BUILD_COMMON variable.

janiversen:
sorry I see what you mean. It. is because in some build.sh I do not use “single”. However they should not be downloaded twice, because if the directory exist the download is skipped.

I will test it tomorrow and to the needed adjustment, thanks for pointing it out.

janiversen:
and do

janiversen:
see #1325

PR #1321, opened 2018-05-24
iOS: updated launchscreen so it is more centered.

janiversen:
The text and icon was so low, that the page started to scroll on
an iPhone (at least on a model 6s).

Anchored Text to top, and icon to text.

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

No text/icon changes, just a layout change. The text is fixed distance to the
top, and the icon has fixed distance to the text. This avoid having half of the icon
cut.

### Changes made:

1) Updated launch screen.xib file, with xcode


dirkhh:
I haven't tried this, but I trust that this is an improvement :-)

PR #1327, opened 2018-05-25
[RFC] Dive pictures: Derive thumbnail file from picture filename

bstoeger:
Since commit 6618c9ebfc6a7cebbef687fcb3aa74c70f504ff2, thumbnails
are saved in individual files. The filename was simply the picture-hash.
In a mailing-list discussion it turned out that in the future we might
not hash images or change the hash. Therefore, derive the thumbnail
filename from the image filename, using the SHA1 algorithm.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This changes the way the thumbnail filenames are generated: Instead of using the image hash, derive it from the image filename. The reason is that according to a mailing-list discussion, we might change or even remove the hash.

The thumbnail filename is generating by SHA1ing the (canonical, not local) picture filename. I wonder if a less heavy hash-function would suffice. After all, all we want to avoid is collisions, isn't it?

### DIFF
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -167,7 +167,6 @@ Thumbnailer *Thumbnailer::instance()
 
 static QImage getThumbnailFromCache(const QString &picture_filename)
 {
-	// First, check if we know a hash for this filename
 	QString filename = thumbnailFileName(picture_filename);
 	if (filename.isEmpty())
 		return QImage();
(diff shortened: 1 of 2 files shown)


neolit123 (approved):
lgtm, i wonder if this needs a release note?

PR #1328, opened 2018-05-25
Cleanup: remove three instances of deleteLater() in mainwindow.cpp and one in qmlprofile.c

bstoeger:
deleteLater() can be dangerous. Remove were not necessary.

Analysis:

1) `helpView` was a pointer which was initialized on demand. close() and
deleteLater() were called on closure of the main window. Firstly, there's
no point in calling deleteLater(), because no references to helpView
are used later on. Secondly, the deletion (and closing) can be done
automatically in the destructor, by passing `this` as parent object.

2) `survey`: pretty much the same situation. But here, `this` was already
passed as parent object.

3) `progressDialog` is a global (not thread safe!) pointer. The object
is deleted after use. There is no point in using deleteLater(), because
the callers are not active after hideProgressBar(), which is the
place were the deleteLater() call was found.

-------------

Cleanup: remove deleteLater() in QMLProfile

The actual profile object was destroyed with deleteLater() in the
destructor of QMLProfile. This is ominous, because the subobject
shouldn't survive the parent object.

Therefore, automatically destroy the profile by using a QScopedPointer.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Removes unnecessary `deleteLater()`s. Note that I test the progressDialog and survey delete, but not the manual. Apparently I compile without user manual.

There seems to be one more questionable `deleteLater()`: in the destructor of `QMLProfile`.
EDIT: I added a commit to remove that as well. Attention: Untested! Currently my desktop-on-mobile version refuses to run:
```
INFO: QQmlApplicationEngine failed to load component
INFO: qrc:///qml/main.qml:530 Type MapPage unavailable
qrc:///qml/MapPage.qml:17 Type MapWidget unavailable
qrc:///qml/MapWidget.qml:3 module "QtLocation" is not installed

can't create window object
```

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -1354,18 +1354,16 @@ void MainWindow::on_action_Check_for_Updates_triggered()
 void MainWindow::on_actionUserManual_triggered()
 {
 #ifndef NO_USERMANUAL
-	if (!helpView) {
-		helpView = new UserManual();
-	}
+	if (!helpView)
+		helpView = new UserManual(this);
 	helpView->show();
 #endif
 }
 
 void MainWindow::on_actionUserSurvey_triggered()
 {
-	if(!survey) {
+	if(!survey)
 		survey = new UserSurvey(this);
-	}
 	survey->show();
 }
 
@@ -1559,18 +1557,6 @@ void MainWindow::closeEvent(QCloseEvent *event)
 		on_actionQuit_triggered();
 		event->ignore();
 		return;
-	}
-
-#ifndef NO_USERMANUAL
-	if (helpView && helpView->isVisible()) {
-		helpView->close();
-		helpView->deleteLater();
-	}
-#endif
-
-	if (survey && survey->isVisible()) {
-		survey->close();
-		survey->deleteLater();
 	}
 
 	if (unsaved_changes() && (askSaveChanges() == false)) {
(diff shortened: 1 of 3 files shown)


neolit123 (approved):
tried the mobile version and couldn't get it to crash.
the changes lgtm.

bstoeger:
> mobile on desktop should work. I use it both on Mac and Linux.
Which OS are you on?

Upgrading to Kubuntu 18.04 messed up the Google maps support (c.f. mailing list). On desktop, the map-widget shows an error message; mobile-on-desktop refuses to start. Not sure if it's worth making mobile-on-desktop more robust.

dirkhh:
Subsurface-mobile on desktop is great for development. And typically those developers who want to work on it make sure that it works on their OS of choice. I don't see a lot of value in anything else there.

bstoeger:
Ugh. Now I installed the package `qml-module-qtlocation` and both subsurface and subsurface-mobile segfault.

So the only way to compile and run the desktop version that I have at the moment is
1) Install `qtlocation5-dev` (to make it compile)
2) Do not install `qml-module-qtlocation` (to make it run)

subsurface-mobile I can't run at all. So I'm a bit lost...

janiversen:
current master work in my mac (high sierra) and my iPad (iOS). Just to be sure I tested the mobile version.

I am instead figthing to be able to use the designer in QtCreator with kirigami (see #1329)

PR #1330, opened 2018-05-26
Cleanup: remove reply and request member variables in QMLManager

bstoeger:
The reply member variable was used to access the reply in the
handleSslErrors, handleError and retrieveUserid slots. This is a
very scary proposition in the light of multi-threading. Instead,
the reply can be accessed by using the QObject::sender() function.
Thus, we can remove the member variable.

The request member was just downright weird. This was only used
locally to describe a network request. Since QNetworkAccessManager::get()
copies the request, it can be destructed right away. Nevertheless,
the data was kept as a subobject. Remove member and make it function-local.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes two ownership issues in qmlmanager.cpp. Warning: Only compile tested!

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -130,7 +130,6 @@ void QMLManager::btRescan()
 
 QMLManager::QMLManager() : m_locationServiceEnabled(false),
 	m_verboseEnabled(false),
-	reply(0),
 	deletedDive(0),
 	deletedTrip(0),
 	m_updateSelectedDive(-1),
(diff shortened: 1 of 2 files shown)


mobile-widgets/qmlmanager.cpp
```diff
@@ -566,6 +565,7 @@ void QMLManager::provideAuth(QNetworkReply *reply, QAuthenticator *auth)
 
 void QMLManager::handleSslErrors(const QList<QSslError> &errors)
 {
+	auto *reply = qobject_cast<QNetworkReply *>(sender());
 	setStartPageText(RED_FONT + tr("Cannot open cloud storage: Error creating https connection") + END_FONT);
```
neolit123, line 569: normally i'm not a fan of `auto`, but in this case the cast indicates the type.

neolit123 (approved):
lgtm, good concurrency fix.

PR #1332, opened 2018-05-26
mobile: Fix missing icons

### LINKED ISSUE #1267: iOS: icons problems
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

1. Some icons are not being displayed even with the correct source path.
    Delete on list:
  img_4175
    GPS fixes list:   
 img_4191 2

1. The application header back button is misaligned:
With PR#1266 applied 
img_4172

I investigated and tried to fix but could not find a solution, so reporting to keep track.

### Operating system:

iOS

mfbernardes:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Icons on actions of GPS fixes list and dive list were missing.

Application header icon was not useful and unaligned, so removed.

Fix #1267

### DIFF
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -211,6 +211,8 @@ Kirigami.ScrollablePage {
 							verticalCenter: parent.verticalCenter
 						}
 						source: ":/icons/trash-empty"
+						width: Kirigami.Units.iconSizes.smallMedium
+						height: width
 					}
 					MouseArea {
 						anchors.fill: parent
--- a/mobile-widgets/qml/GpsList.qml
+++ b/mobile-widgets/qml/GpsList.qml
@@ -74,7 +74,7 @@ Kirigami.ScrollablePage {
 			actions: [
 				Kirigami.Action {
 					icon {
-						name: "trash-empty"
+						name: ":/icons/trash-empty"
 					}
 					onTriggered: {
 						print("delete this!")
@@ -83,7 +83,7 @@ Kirigami.ScrollablePage {
 				},
 				Kirigami.Action {
 					icon {
-						name: "gps"
+						name: ":/icons/gps"
 					}
 					onTriggered: {
 						showMap()


scripts/mobilecomponents.sh
```diff
@@ -62,14 +62,6 @@ sed -i -e "s/visible: root.rightAction/visible: root.rightAction \&\& \!Qt.input
...
 sed -i -e "s/width: backgroundRect/enabled: root.enabled;    width: backgroundRect/g" src/controls/templates/private/PassiveNotification.qml
 
-# another hack. Do not include the Kirigami resources (on static build). It causes
-# double defined symbols in our setting. I would like a nicer fix for this
-# issue, but failed to find one. For example, not adding the resource in
-# our build causes the qrc file not to be generated. Manual generation
-# of the resource file (using rcc) introduces the double symbols again. 
-# so it seems some Kirigami weirdness (but their staticcmake example compiles
-# correctly).
-sed -i -e "s/#include <qrc_kirigami.cpp>/\/\/#include <qrc_kirigami.cpp>/" src/kirigamiplugin.cpp
```
neolit123: i did this today and it was needed otherwise i was getting weird duplicate symbols when building.

mobile-widgets/qml/main.qml
```diff
@@ -19,6 +19,7 @@ Kirigami.ApplicationWindow {
 		preferredHeight: Math.round(Kirigami.Units.gridUnit * (Qt.platform.os == "ios" ? 2 : 1.5))
 		maximumHeight: Kirigami.Units.gridUnit * 2
 		background: Rectangle { color: subsurfaceTheme.primaryColor }
+		backButtonEnabled: false
```
neolit123: i think this button is actually used.
would be great if others can confirm.

neolit123 (commented):
added a couple of comments.
i mostly test the app on desktop, so requesting a review from actual users on mobile.

scripts/mobilecomponents.sh
neolit123: the travis build failed and it kind of confirms my observation:
https://api.travis-ci.org/v3/job/384204511/log.txt
```

/workspace/ndk-arm/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: CMakeFiles/subsurface-mobile.dir/qrc_kirigami.cpp.o: multiple definition of 'qInitResources_kirigami()'

/workspace/ndk-arm/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: CMakeFiles/subsurface-mobile.dir/mobile-widgets/qml/kirigami/src/kirigamiplugin.cpp.o: previous definition here

/workspace/ndk-arm/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: CMakeFiles/subsurface-mobile.dir/qrc_kirigami.cpp.o: multiple definition of 'qCleanupResources_kirigami()'

/workspace/ndk-arm/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: CMakeFiles/subsurface-mobile.dir/mobile-widgets/qml/kirigami/src/kirigamiplugin.cpp.o: previous definition here

```

mfbernardes: It was not a problem for iOS, but if it breaks android or mobile on desktop I'll leave it.

I just removed this commit.

mobile-widgets/qml/main.qml
mfbernardes: I checked the current user guide and I don't see it there.

It's unaligned on current Kirigami and it takes even more space on newer Kirigami versions.

The button is so small on an iPhone 7 Plus, I doubt it's usable on smaller screens anyway.

janiversen: We do not use it on iOS, because swipe is active, but did you check if it is connected to an action on android (I am not on my mac the next hours so cannot really check).

I agree it is useless on small devices, provided there are e.g. swipe.

mfbernardes: To be sure it's clear which button is disabled here. This will just make it look like before.

Current master:
back

From the user manual:
user manual

neolit123: i remember using this `<` button while testing on the desktop, but it looked fine.
perhaps we should fix the alignment and size of the button, as i do think that folks actually use it.

waiting on more comments:
@dirkhh @atdotde

janiversen:
@mfbernardes the 2 pictures you show, represent 2 different situations. The back arrow does not appear until you use the second window in the stack...e.g. you can go from divelist to divedetail without ever seeing the back arrow, but the choose download, and it appears.

@neolit123 having had time to test both on my Mac and my iPad, I really prefer if we keep it (and find the alignment problem). It is however correct that on a small device it is really only an indicator, that you can swipe back to the divelist.

mfbernardes:
@janiversen the two pictures represent the same situation, even if from different ways to get there. Both contain two pages on the stack (depending where you go the stack does not get cleared in others it does).

OK. I'll just remove this commit and be done with it. The other commit is the one that really matters.

janiversen:
Thanks, I recommend merging this patch.

@mfbernardes, did not test but if the alignment is still a problem after the kirigami update, it would be nice to have either a new issue or an update to the old issue, so we do not forget it.

PR #1335, opened 2018-05-27
mobile: Adjust About.qml to fit small screens.

janiversen:
Change to Page instead of ScrollPage to ensure that
the buttom is always visible.

Adjust size of image and text.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The About box was so large that on a iPhone 6s, you did not see the clipboard
bottom.

### Changes made:

1) Changed from ScrollablePage to Page
2) Sized icon to pagewidth / 2
3) made top text fixed with font adjustable

### DIFF
--- a/mobile-widgets/qml/About.qml
+++ b/mobile-widgets/qml/About.qml
@@ -21,16 +21,16 @@ Kirigami.ScrollablePage {
 			Layout.topMargin: Kirigami.Units.gridUnit
 			Layout.alignment: Qt.AlignHCenter
 			Layout.maximumWidth: pageWidth
-			wrapMode: TextEdit.WrapAtWordBoundaryOrAnywhere
+			wrapMode: TextEdit.NoWrap
+			fontSizeMode: Text.Fit
 		}
 		Image {
 			id: image
 			source: "qrc:/qml/subsurface-mobile-icon.png"
-			width: pageWidth / 2
-			height: width
-			fillMode: Image.Stretch
-			Layout.alignment: Qt.AlignCenter
-			horizontalAlignment: Image.AlignHCenter
+			fillMode: Image.PreserveAspectCrop
+			Layout.alignment: Qt.AlignHCenter + Qt.AlignVCenter
+			Layout.maximumWidth: pageWidth / 2
+			Layout.maximumHeight: Layout.maximumWidth
 		}
 
 		Kirigami.Heading {


mobile-widgets/qml/About.qml
```diff
@@ -21,16 +21,16 @@ Kirigami.ScrollablePage {
...
 			source: "qrc:/qml/subsurface-mobile-icon.png"
-			width: pageWidth / 2
-			height: width
-			fillMode: Image.Stretch
-			Layout.alignment: Qt.AlignCenter
-			horizontalAlignment: Image.AlignHCenter
+			fillMode: Image.PreserveAspectCrop
+			Layout.alignment: Qt.AlignHCenter
+			Layout.maximumWidth: pageWidth / 2
+			Layout.maximumHeight: Layout.maximumWidth
```
neolit123: would this work on both landscape and portrait mode?

http://doc.qt.io/qt-5/qml-qtquick-image.html#fillMode-prop
why not do `PreserveAspectFit`?

neolit123: extra points for including screenshots :+1:

neolit123: > ou got a point with landscape, I will look into that (I hope kirigami can allow another layout for landscape).

> If I understand it correctly (and saw how it looked when trying) the difference between
PreserveAspectFit and PreserveAspectCrop is that "Fit" will stretch if needed while crop
will reduce while needed.

i see,
something in these lines might work (untested):
```
property int pageWidth: aboutPage.width - aboutPage.leftPadding - aboutPage.rightPadding
property int pageHeight: aboutPage.height - aboutPage.topPadding - aboutPage.bottomPadding

...

fillMode: Image.PreserveAspectCrop
Layout.alignment: Qt.AlignHCenter
Layout.maximumWidth: pageWidth < pageHeight ? pageWidth / 2 : pageHeight / 2
Layout.maximumHeight: Layout.maximumWidth
```

EDIT: but then again i think `pageWidth >= pageHeight` might never happen, due to the kirigami thing of showing multiple pages in landscape instead of stretching a single page.

the contents of the second image are not centered horizontally, not sure if that was the case before and what is causing it.

janiversen:
The image is not centered vertically (and was not), I think it has something to do with the text paddings, but it does not disturb.

I had now solved landscape as well (@neolit123 thanks for pointing it out).

captura de pantalla 2018-05-27 a las 16 02 24
captura de pantalla 2018-05-27 a las 16 02 38
captura de pantalla 2018-05-27 a las 16 02 47

@neolit123 happy ?

neolit123 (approved):
> @neolit123 happy ?

yes, thank you.
comments by others?

PR #1337, opened 2018-05-28
[WIP] mobile: temporary measure to get startup times from different devices

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] Test

There is a timer object started in subsurface-mobile-main.cpp (main)
and with the help of a macro, different steps in
first main
then qmlmanager
is added to the clipboard (as a sort of log extension).

This allows us to publish a new TestFlight (and whatever it is called in android), so
we. can ask the testers, to mail us their logs after startup. That way we can look for a
consistent pattern (on my iPhone it is run_ui, that causes the biggest delay).

the following is from and iPhone 6s:
---------- startup timer ----------
STP 0 ms, main starting
STP 306 ms, main Qt started
STP 311 ms, main git loaded
STP 313 ms, main xml parsed
STP 313 ms, main taglist done
STP 353 ms, main init_ui done
STP 353 ms, main call run_ui (continue in qmlmanager)
STP 3372 ms, qmlmgr starting
STP 3376 ms, qmlmgr log started
STP 3378 ms, qmlmgr start page
STP 3381 ms, qmlmgr bt available
STP 3388 ms, qmlmgr gps started
STP 3388 ms, qmlmgr git update
STP 3389 ms, qmlmgr sync load cloud
---------- finish ----------

Once the analysis is done, the code can be removed again.

### DIFF
--- a/core/ssrf.h
+++ b/core/ssrf.h
@@ -19,7 +19,11 @@ extern "C" {
 #define UNUSED(x) (void)x
 #endif
 
+#ifdef ENABLE_STARTUP_TIMING
 #ifdef STARTUP_TIMER
+#define STP_SETUP() QTime stpDuration; \
+					QString stpText
+#define STP_RUN() stpDuration.start()
 // Block to help determine where time is "lost" in
 // the mobile version during startup
 #include <QTime>
@@ -30,6 +34,11 @@ extern QString stpText;
 							.append(" ms, ") \
 							.append(x) \
 							.append("\n")
+#endif // STARTUP_TIMER
+#else // Release version
+#define STP_SETUP() 
+#define STP_RUN()
+#define LOG_STP(x)
 #endif
 
 #endif // SSRF_H
--- a/core/ssrf.h
+++ b/core/ssrf.h
@@ -19,4 +19,17 @@ extern "C" {
 #define UNUSED(x) (void)x
 #endif
 
+#ifdef STARTUP_TIMER
+// Block to help determine where time is "lost" in
+// the mobile version during startup
+#include <QTime>
+extern QTime stpDuration;
+extern QString stpText;
+#define LOG_STP(x) stpText +=QString("STP ") \
+							.append(QString::number(stpDuration.elapsed())) \
+							.append(" ms, ") \
+							.append(x) \
+							.append("\n")
+#endif
+
 #endif // SSRF_H
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -362,6 +362,8 @@ void QMLManager::copyAppLogToClipboard()
 		QTextStream in(&f);
 		copyString += in.readAll();
 	}
+	copyString += "---------- startup timer ----------\n";
+	copyString += stpText;
 	copyString += "---------- finish ----------\n";
 
 	// and copy to clipboard
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -362,8 +362,10 @@ void QMLManager::copyAppLogToClipboard()
 		QTextStream in(&f);
 		copyString += in.readAll();
 	}
+#ifdef ENABLE_STARTUP_TIMING
 	copyString += "---------- startup timer ----------\n";
 	copyString += stpText;
+#endif
 	copyString += "---------- finish ----------\n";
 
 	// and copy to clipboard
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -32,6 +32,9 @@
 #include "qt-models/tankinfomodel.h"
 #include "core/downloadfromdcthread.h"
 
+#define STARTUP_TIMER
+#include "core/ssrf.h"
+
 QMLManager *QMLManager::m_instance = NULL;
 
 #define RED_FONT QLatin1Literal("<font color=\"red\">")
@@ -138,6 +141,7 @@ QMLManager::QMLManager() : m_locationServiceEnabled(false),
 	alreadySaving(false),
 	m_device_data(new DCDeviceData(this))
 {
+	LOG_STP("qmlmgr starting");
 	m_instance = this;
 	m_lastDevicePixelRatio = qApp->devicePixelRatio();
 	timer.start();
@@ -164,6 +168,7 @@ QMLManager::QMLManager() : m_locationServiceEnabled(false),
 				+ " at " + QDateTime::currentDateTime().toString());
 	}
 #endif
+	LOG_STP("qmlmgr log started");
 	set_error_cb(&showErrorFromC);
 	appendTextToLog("Starting " + getUserAgent());
 	appendTextToLog(QStringLiteral("built with libdivecomputer v%1").arg(dc_version(NULL)));
@@ -172,11 +177,13 @@ QMLManager::QMLManager() : m_locationServiceEnabled(false),
 	git_libgit2_version(&git_maj, &git_min, &git_rev);
 	appendTextToLog(QStringLiteral("built with libgit2 %1.%2.%3").arg(git_maj).arg(git_min).arg(git_rev));
 	setStartPageText(tr("Starting..."));
+	LOG_STP("qmlmgr start page");
 
 	// ensure that we start the BTDiscovery - this should be triggered by the export of the class
 	// to QML, but that doesn't seem to always work
 	BTDiscovery *btDiscovery = BTDiscovery::instance();
 	m_btEnabled = btDiscovery->btAvailable();
+	LOG_STP("qmlmgr bt available");
 	connect(&btDiscovery->localBtDevice, &QBluetoothLocalDevice::hostModeStateChanged,
 		this, &QMLManager::btHostModeChange);
 	setShowPin(false);
@@ -185,10 +192,13 @@ QMLManager::QMLManager() : m_locationServiceEnabled(false),
 	progress_callback = &progressCallback;
 	connect(locationProvider, SIGNAL(haveSourceChanged()), this, SLOT(hasLocationSourceChanged()));
 	setLocationServiceAvailable(locationProvider->hasLocationsSource());
+	LOG_STP("qmlmgr gps started");
 	set_git_update_cb(&gitProgressCB);
+	LOG_STP("qmlmgr git update");
 
 	// make sure we know if the current cloud repo has been successfully synced
 	syncLoadFromCloud();
+	LOG_STP("qmlmgr sync load cloud");
 }
 
 void QMLManager::applicationStateChanged(Qt::ApplicationState state)
--- a/subsurface-mobile-helper.cpp
+++ b/subsurface-mobile-helper.cpp
@@ -29,6 +29,9 @@
 
 #include "mobile-widgets/qml/kirigami/src/kirigamiplugin.h"
 
+#define STARTUP_TIMER
+#include "core/ssrf.h"
+
 QObject *qqWindowObject = NULL;
 
 void set_non_bt_addresses() {
@@ -51,6 +54,7 @@ void init_ui()
 
 void run_ui()
 {
+	LOG_STP("run_ui starting");
 	qmlRegisterType<QMLManager>("org.subsurfacedivelog.mobile", 1, 0, "QMLManager");
 	qmlRegisterType<QMLProfile>("org.subsurfacedivelog.mobile", 1, 0, "QMLProfile");
 
@@ -63,6 +67,7 @@ void run_ui()
 	qmlRegisterType<MapLocation>("org.subsurfacedivelog.mobile", 1, 0, "MapLocation");
 
 	QQmlApplicationEngine engine;
+	LOG_STP("run_ui qml engine started");
 	KirigamiPlugin::getInstance().registerTypes();
 #if defined(__APPLE__) && !defined(Q_OS_IOS)
 	// when running the QML UI on a Mac the deployment of the QML Components seems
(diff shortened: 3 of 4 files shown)


janiversen:
@bstoeger do you have other places we could/should measure time...run_ui can maybe be split ?

@dirkhh if you agree to the nature of this PR, could you please make a new testFlight and similar for Android asking the tester to mail us (or me, not to overload the ML), so we can get a better picture of where the startup problem are.

neolit123:
yes, it's please add some log calls in `run_ui()` before asking for results.
i won't be surprised if the primary cause for stalling that much in `run_ui()` are things that we cannot easily change.

dirkhh:
I like this. And I think it's reasonable to say that unless this gets merged and added to "beta builds" we won't really get a lot of data.
Let me do a quick code review to see if I have any real issues here and then wait for Travis (which of course just got restarted...)

janiversen:
Updated with run_ui()

@dirkhh please do the code review on run_ui (last commit) as well.

subsurface-mobile-helper.cpp
```diff
@@ -51,6 +54,7 @@ void init_ui()
 
 void run_ui()
 {
+    LOG_STP("run_ui starting");
```
dirkhh: white space... everywhere... you really should just update your editor settings :-)

janiversen: I did that a couple of days ago, give me half an hour.

dirkhh: no problem - I just find it easier to get things like this fixed before they go into the code :-)

neolit123: i don't think we should merge this in master as is.
isn't this PR only for testing purposes?

janiversen:
Done, and just to be fair it was only in the last file :-)

The problem is the macOS clipboard, it converts tabs to spaces, when doing a copy/paste, did not know that, and more importantly have not a clue how to avoid that. So I will be more careful with copy/paste.

neolit123:
@janiversen what results do you get after the `run_ui()` breakdown?

subsurface-mobile-helper.cpp
dirkhh: (a) I find it incredibly painful to build test binaries that are pushed to TestFlight / GooglePlay Beta that are from branches - and it confuses the versioning of those builds
(b) I think it's reasonable to have the code in place to do such timing, even if we disable it for releases

neolit123:
> The problem is the macOS clipboard, it converts tabs to spaces, when doing a copy/paste.

this is so developer friendly. maybe try a clipboard manager.

dirkhh:
At least with QtCreator you can tell it to convert to our style on saving... which should avoid most of these situations

janiversen:
@neolit123 new numbers with run_ui():
STP 1 ms, main starting
STP 332 ms, main Qt started
STP 339 ms, main git loaded
STP 345 ms, main xml parsed
STP 346 ms, main taglist done
STP 417 ms, main init_ui done
STP 417 ms, main call run_ui (continue in qmlmanager)
STP 417 ms, run_ui starting
STP 549 ms, run_ui qml engine started
STP 588 ms, run_ui diveListModel started
STP 592 ms, run_ui diveListModel sorted
STP 595 ms, run_ui set_non_bt_adresses
STP 2788 ms, qmlmgr starting
STP 2792 ms, qmlmgr log started
STP 2795 ms, qmlmgr start page
STP 2798 ms, qmlmgr bt available
STP 2807 ms, qmlmgr gps started
STP 2808 ms, qmlmgr git update
STP 2808 ms, qmlmgr sync load cloud
STP 5780 ms, run_ui qml loaded
STP 7619 ms, run_ui qmlmanager instance started
STP 7620 ms, run_ui show_computer_list
STP 7621 ms, run_ui running exec

a bit mixed since it happens it 2 threads.

@dirkhh maybe I should learn my vi to do that as well, must be possible.

@dirkhh I can add a "disabling" mechanism if you want not a big problem, by adding a -D<something> when we want to activate it. Do you want that in this PR or later ?

janiversen:
release friendly version. If -DENABLE_STARTUP_TIMING is not set, all macros for startup_timer will be ignored.

Please make sure to define ENABLE_STARTUP_TIMING (e.g. in a local version of ssrf.h) when making the beta versions.

neolit123:
> @neolit123 new numbers with run_ui():

i'm confused by the results - will try to run them locally.

janiversen:
please do, we need to see these number from different iOS as well as Android devices.

Remark my iPhone does not have BT enabled.

core/ssrf.h
```diff
@@ -19,4 +19,26 @@ extern "C" {
...
+#ifdef STARTUP_TIMER
+#define STP_SETUP() QTime stpDuration; \
+					QString stpText
+#define STP_RUN() stpDuration.start()
+// Block to help determine where time is "lost" in
+// the mobile version during startup
+#include <QTime>
+extern QTime stpDuration;
+extern QString stpText;
+#define LOG_STP(x) stpText +=QString("STP ") \
```
neolit123, line 32:
```
In file included from C:/bin/qt/5.9/mingw53_32/include/QtCore/qstring.h:49:0,
                 from C:/bin/qt/5.9/mingw53_32/include/QtCore/qdatetime.h:44,
                 from C:/bin/qt/5.9/mingw53_32/include/QtCore/QTime:1,
                 from C:\dev\subsurface\core\ssrf.h:35,
                 from C:\dev\subsurface\core\cochran.c:7:
```
`cochran.c` shouldn't include Qt code.

core/ssrf.h
```diff
@@ -19,4 +19,26 @@ extern "C" {
...
 
+#ifdef ENABLE_STARTUP_TIMING
+#ifdef STARTUP_TIMER
+#define STP_SETUP() QTime stpDuration; \
+					QString stpText
+#define STP_RUN() stpDuration.start()
+// Block to help determine where time is "lost" in
+// the mobile version during startup
+#include <QTime>
+extern QTime stpDuration;
```
neolit123, line 30: the way to test heavy code in Qt is `QElapsedTimer`.

core/ssrf.h
```diff
@@ -19,4 +19,26 @@ extern "C" {
...
+extern QString stpText;
```
neolit123, line 31: QString is reentrant but not thread-safe.

dirkhh:
@neolit123 points out something that I missed in my review - we should put the measurement code into its own header file that is only included in C++ files. And yes, QElapsedTimer sounds like the better choice.

janiversen:
Will update with the wanted changes tomorrow. Any suggestion how to easy make a thread safe String, that do not influence the timing a lot ?

mobile-widgets/qmlmanager.cpp
```diff
@@ -32,6 +32,9 @@
 #include "qt-models/tankinfomodel.h"
 #include "core/downloadfromdcthread.h"
 
+#define STARTUP_TIMER
```
neolit123, line 35: i think the logger code needs a better implementation - i.e. a better implementation of a common C/C++ logging pattern controlled by a pre-processor macro.
yet i don't have the time to propose a better solution ATM.

mobile-widgets/qmlmanager.cpp
```diff
@@ -352,6 +362,10 @@ void QMLManager::copyAppLogToClipboard()
 		QTextStream in(&f);
 		copyString += in.readAll();
 	}
+#ifdef ENABLE_STARTUP_TIMING
+	copyString += "---------- startup timer ----------\n";
+	copyString += stpText;
+#endif
 	copyString += "---------- finish ----------\n";
```
neolit123, line 369: ~we should just write the logging output among the rest of the logging output.~
...except that the logger is started from from QMLmanager.

janiversen: And how do I write logging output before the log file is opened (subsurface-mobile-main.cpp) ?

mobile-widgets/qmlmanager.cpp
janiversen: Please, I am not rewriting our logger or anything, just providing a simple test tool for a specific problem.

Might be better if I just close this, and do my own testing....it is a pity because I think we could benefit from having numbers from different devices, but I do not have the time to make a new elegant logger solution.

neolit123:
i have concerns that this code is not structured properly. so a NAK for now...
we need to think of a way to make it better.

janiversen:
point taken, please start thinking, I am done.

neolit123:
the real bottleneck here is:
`engine.load(QUrl(QStringLiteral("qrc:///qml/main.qml")));`

on the desktop it can take more than 10 seconds the first time this is cached in memory.
everything else is insignificant overhead, thus we don't need a performance timer outside of QML (for now).

in terms of performance logging inside QML this can be a pain to process correctly because QML is confusing and it executes everything out of order.

i can tell your right now that this is what's causing it:
```

	Settings {
		id: settingsWindow
		visible: false
	}

	About {
		id: aboutWindow
		visible: false
	}

	DiveDetails {
		id: detailsWindow
		visible: false
		anchors.fill: parent
	}

	Log {
		id: logWindow
		visible: false
	}

	GpsList {
		id: gpsWindow
		visible: false
	}

	DownloadFromDiveComputer {
		id: downloadFromDc
		visible: false
	}

	MapPage {
		id: mapPage
		visible: false
		anchors.fill: parent
	}
```

^ we instantiate all pages all at once and not on demand. this can be improved in the following way:
- implement a `on-demand-page-creator` that pushes pages on the stack at the correct position.
- implement a "loading..." overlay that is shown to the user while the `on-demand-page-creator` is doing it's thing.

that said, i have a smart phone from 2016 that starts subsurface much faster than 10seconds (desktop is slower) - around 5 seconds. which i'd say is **quite tolerable**.

the above solution is doable in theory, but one can only guess what surprises QML and Kirigami will present....so that's my thinking on the subject. if someone want's to take a stab at the QML optimizations, please go ahead and i will be happy to help with reviews (cough) and ideas if time permits.

dirkhh:
Ugh. Frustrating to spend time with the family and then see things go sideways like this.
@neolit123 @janiversen - you are both trying to do the right thing, but it sounds like you are talking at cross purposes sometimes.

This is NOT rewriting our logger, this is assembling a string that a user can easily copy after starting Subsurface-mobile and send to us so we get better information as to where we are spending time at app startup. And I expect that information to be potentially significantly different between "Subsurface-mobile on desktop", "Subsurface-mobile on Android" and "Subsurface-mobile on iOS".

So, the purpose is valid, adding it to master makes sense to me.
What we need is a simple, thread-safe way to add to a QString. And it turns out that's hard. But how about we start with something simple as explain in the answer here https://stackoverflow.com/questions/18934736/stdstring-append-thread-safe-naive-solution and accept the small error we'll introduce with the mutex?

janiversen:
Indeed frustrating, but then try to be at my end, being hit by a core developer, who hold the right to qualify my work.

I highly agree, that these numbers have a great value for us, and as you can see above, I was prepared to change the PR, with the requested changes (some of them simply not to step on toes, because I do not agree with them). But after that I got hit with a load of comments/demands like:
- I had to use the normal log (demand was changed, after I pointed out, that logging is started much later)
- there is an issued NAK to the solution (without even considering that I proposed to rewrite it)
- the real bottleneck is already found (which btw totally ignores the warning from Qt about JIT)

I am a burned kid with the experience that I do not stand a chance once a core developer have
marked my PR´s as bad (remark I do NOT speak about the normal and useful review comments, which
are highly welcome). In those cases it is simply easier to move on.

I am not giving up, I have 2 major mobile rewrites ongoing (divelist and profile), which I want for my iPad, but I have serious doubts on whether continuing with PR´s or make it in my clone, and offer the changes "as is" (meaning not participating in the discussions).

I am aware that I have entered this community with a very high level of commitment and work, which I suspect have surprised (and/or scared?) some core developers, but that is my style as I do not believe in sneaking.

I will wait a while and see how this plays out (meaning watch how core developers settle a disagreement) and see if there is reason for me to regain my motivation.

Sorry for writing directly, I meant no offence to anybody and apologise upfront if anybody feel my description is unjust.

bstoeger:
> What we need is a simple, thread-safe way to add to a QString. And it turns out that's hard.

Actually, this seems to be trivial to me?
```
static QString log;
static QMutex logMutex;
void addToLog(const QString &s)
{
    QMutexLocker l(&logMutex);
    log +=s;
}
QString getLog()
{
    QMutexLocker l(&logMutex);
    return log;
}
```

But I must admit that I share @neolit123's concerns. Currently, this seems to be the equivalent of debugging `printf`s. I see no particular urge to merge this into master. Especially since the major culprit now seems to be identified.

I think the proper approach would now be doing the detail work. E.g. slapping a timer around
```
engine.load(QUrl(QStringLiteral("qrc:///qml/main.qml")));
```
And try to optimize the QML. No?

dirkhh:
The inability to re-open appears to be caused by the branch having changed in a way that GitHub can't parse - so just open a new PR.
A few comments on the other items:

1) I think our "rules" are far less fixed than they are in the Apache community. @neolit123 is indeed a core contributor - and he typically gives very good reviews and pays great attention to detail. That doesn't mean that just because he doesn't like something it's dead. If Linus or I dislike things, they may take a little more elbow grease to move forward - but quite a few things have gone into Subsurface even though BOTH Linus and I were not excited (or outright against them).
I think this is the major difference to some other projects - for Subsurface your best bet is to keep making your case. Respond to requests, push back where you disagree, try to engage more of the core people. We don't have a hierarchical model or a voting model but much more a "get people on board" model. Less well defined, but based on the way Linux has been developed for 26 years...

2) Fundamentally I think this idea of providing startup timing is the right thing to do - and I said this multiple times.
So let's make this thread safe, let's make it low overhead, let's make it easy to disable (or disable it by default and make it easy to enable in a build).

3) one of the key reasons why I think this is very useful is because the three platforms (iOS, Android, desktop) do behave so oddly differently. Lack of JIT on iOS, interfacing with the Java loader on Android, etc.

4) I personally find the startup time disappointing. Whenever I show Subsurface-mobile to people there is this looong wait. So I personally love to see improvement.

5) finally... I don't think your activity level scares anyone. When @tcanabrava joined the Subsurface team and did 80+% of the work to port Subsurface to Qt, his ramp in activity was flat out insane :-)
IIRC he had something like 150 commits in his first 6 weeks :-)

PR #1341, opened 2018-05-29
[WIP] core: btdiscovery, remove rediscovery() in constructor

janiversen:
The btdiscovery object is created during startup, doing
a discovery at the point in time is prolonging the startup
time which is long enough (15sec on an iPhone).

- [ ] Bug fix
- [x ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I am reopening this PR, for several reasons:
1) if BT is not active, you get a request to activate BT
2) startup time is prolonged due to the scan
3) a paired DC, does not change id on iOS and Android (acc. doc.)
    so a startup scan is really not needed.

### Changes made:

1) removed btDiscovery from the constructor

### Mentions:

@dirkhh

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -72,9 +72,6 @@ BTDiscovery::BTDiscovery(QObject*) : m_btValid(false),
 		return;
 	}
 	m_instance = this;
-#if defined(BT_SUPPORT)
-	BTDiscoveryReDiscover();
-#endif
 }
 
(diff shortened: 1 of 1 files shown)


janiversen:
on my iphone (which do not use BT) I have to scan if I activate BT. On my ipad I forgot the patch to update prefs. 

I will add another commit later

PR #1347, opened 2018-06-01
user-manual: add note that some Grantlee variables need '|safe'

neolit123:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

Grantlee variables need the '|safe' extension, so that they can use
HTML tags. Without it, tags like 'br' would not be converted to line
breaks - e.g. when exporting planned dive notes.

Add a note about that in the user manual.

### Changes made:

- added note in manual

### Related issues:

we have forum/ML reports for this

### Additional information:

none

### Mentions:

@willemferguson @dirkhh

### DIFF
--- a/Documentation/user-manual.txt
+++ b/Documentation/user-manual.txt
@@ -5035,6 +5035,12 @@ Only a subset of the dive data is exported:
 |firstGas| (*string*) first used gas
 |=====================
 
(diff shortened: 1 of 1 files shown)


willemferguson:
LGTM

PR #1348, opened 2018-06-03
Profile: open pictures using local file path

bstoeger:
Make the behavior of the profile-pictures consistent with the pictures
in the photos-tab: Use the local file path to open the picture in the
system viewer instead of the canonical filename (which might point to
a non-existing location).

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a tiny bug: If the pictures are not located at their original place, double-clicking in the photo-tab works, but clicking on the profile image fails. Make both behave consistently.

### DIFF
--- a/profile-widget/divepixmapitem.cpp
+++ b/profile-widget/divepixmapitem.cpp
@@ -3,6 +3,7 @@
 #include "profile-widget/animationfunctions.h"
 #include "qt-models/divepicturemodel.h"
 #include "core/pref.h"
+#include "core/qthelper.h"
 #ifndef SUBSURFACE_MOBILE
 #include "desktop-widgets/preferences/preferencesdialog.h"
 #endif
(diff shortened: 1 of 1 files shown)


janiversen (commented):
I do like the folding, but I have a concern about prefixing our includes with qt (I tend to expect qt<something> to be a qt file) I would much more have preferred the reverse fold qthelper into helper.

atdotde:
This obviously fixes a bug. @janiversen I don't understand the relation of your comment to this patch. Please elaborate.

bstoeger:
> I do like the folding, but I have a concern about prefixing our includes with qt (I tend to expect qt to be a qt file) I would much more have preferred the reverse fold qthelper into helper.

This should have gone below #1350. Since the implementations are in `qthelper.cpp`, it makes sense to fold into the corresponding header. Renaming is trivial and can be done in a subsequent commit.

janiversen:
@atdotde yes it fixes a bug, and I commented in the wrong PR as @bstoeger noted.

PR #1350, opened 2018-06-03
Cleanup: make copyPath() function of static linkage / fold helpers.h into qthelper.h

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

1) A tiny cleanup: make a function local.
2) A bigger cleanup: fold `helpers.h` into `qthelper.h`

The distinction between both headers seemed arbitrary.


janiversen:
@bstoeger I totally agree it makes sense to fold, and since the cpp file is qthelper of course the header file should have the same name. Apart from commenting on the wrong PR, I obviously did not explain myself well, my preference is (for cpp and h file) not to name them qt<something>. But it is just a note of my preference nothing more.

bstoeger:
I understood - my point was that removing "qt" from the file names can be done in a later commit.

PR #1353, opened 2018-06-04
Dive picture: fix toggling of picture-visibility

bstoeger:
In a recent commit, the profile widget was changed to reuse
DivePictureItems. This made toggling of picture-visibility
disfunctional, because the visibility was set in the constructor.
Only on change of dive would the setting take effect.

Therefore, move the setting of the visibility up to the
ProfileWidget2::plotPictures() call. This is more consistent
anyway, since at this level pixmap and fileUrl are set.

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a functional change / fix of a regression, which I introduced recently. With this fix, toggling of the "view pictures" setting takes effect immediately.


dirkhh:
Thanks, looks good.

PR #1354, opened 2018-06-04
mobile: show last used connection in Download page

janiversen:

- [ ] Bug fix
- [x ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The download page is created before the BT list is available so it is too early to
set index of connection combobox. Delayed setting until page becomes visible.

Removed unused function (+ Qt moc) in core/downloadFromComputer.*

Remark: the startup timing test, shows the BtRediscovery is not delaying the startup,
so no real need to remove that.


dirkhh:
Nice cleanups / improvements.

PR #1358, opened 2018-06-05
core: remove "static void copyPath" unused

janiversen:
Remove copyPath since it is static and not called
to avoid clang warning

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

commit message should say it all.

the warning comes when building mobile for iOS, seems desktop suppressed this warning.

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -711,20 +711,6 @@ QString getPrintingTemplatePathBundle()
 	return path;
 }
 
-static void copyPath(QString src, QString dst)
-{
-	QDir dir(src);
-	if (!dir.exists())
-		return;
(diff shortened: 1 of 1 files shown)


janiversen:
@bstoeger did you leave the function on purpose (in which case this PR should be closed), or was it just an oversight ?

bstoeger:
Oversight - I didn't realize that the only caller was the function itself.

janiversen:
Thanks, just wanted to be sure...actually I made the same mistake, because the function is used but by the function itself :-)

neolit123:
the only user was `desktop-widgets/templatelayout.cpp` but it was changed to have a custom `copy` function.

PR #1359, opened 2018-06-06
iOS: move get-dep-lib outside ARCH loop

janiversen:
Fetch all 3rd party libs before entering ARCH loop

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

no need to run get-dep-lib for each architecture

### Changes made:

1) see commit message.

### DIFF
--- a/packaging/ios/build.sh
+++ b/packaging/ios/build.sh
@@ -73,6 +73,9 @@ popd
 
 # now build all the dependencies for the three relevant architectures (x86_64 is for the simulator)
 
+# get all 3rd part libraries
+../../scripts/get-dep-lib.sh ios .
+
 for ARCH in $ARCHS; do
 
 echo next building for $ARCH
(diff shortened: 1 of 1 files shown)


janiversen:
I start to think that Travis-linux do not like me :-(

failed in
10 - TestPicture (Failed)

and this PR is only about iOS.

bstoeger:
This is probably my recent changes to the picture code. Linux builds twice and it could be that the hash-file has funny side-effects. I will try to investigate tonight.

bstoeger:
It's not your commit - I can reproduce this with current master. Seems to be some multi-threading thing as it only happens occasionally.

janiversen:
Thanks for the infirmation. Is it worth to restart the build (I do not have karma to dot it) ?

bstoeger:
I understand the problem now - will see what I can do about it.

I don't think that it is necessary to restart the build as obviously all other platforms succeeded.

PR #1352, opened 2018-06-04
Bugfix: Load PSC dive logs

willemferguson:
The code changes to standardise the named of divemodes and to
separate internal divemode names and UI divemode names introduced
a bug that caused non-backward compatability with existing
dive logs. The reason for this is the definition of the
divemode_text strings in dive.c

This change reverses that definition and brings about correct
loading of PSCR dive logs as well as correct parsing of bailout
events involving PSCR.

Signed-off-by: Willem Ferguson <willemferguson@zoology.up.ac.za>

### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@atdotde @sfuchs79

PR #1360, opened 2018-06-06
Provide correct cylinder pressures for bailout dives

willemferguson:
Calculate the correct cylinder pressures for rebreather dives with
bailout. Currently the cylinder pressures for a dive are calculated
assuming a single dive mode for that dive. Bailout indroduces more
than one dive mode for a single dive, i.e. transitions from
CCR or PSCR to OC and back. Currently the start and end pressures
for each cylinder are used to interpolate cylinder pressures while that
cylinder is used. However, the different gas consumption rates for
OC, PSCR and CCR are not taken into account in this interpolation
and the cylinder pressure is indicated by an averaged interpolation
accross the rebreather and OC legs of the dive. Consequently the
increased drop in cylinder pressure during OC is not shown. This
PR allows differentiation between CCR/PSCR legs of the dive and
the OC bailout segments, showing realistic interpolation that
indicate the increased rate of gas use during OC, illustrated in the
image blow. The image on the left shows the current code with averaging
of pressures for the cylinder that starts towards the bottom left, including the
period from bailout to changing back to rebreather. The image on the
right shows the code for this PR, showing the increased rate of gas use
during the period of bailout.

cylpressure1

This PR is dependent on #1352 without which it will not function.

Signed-off-by: Willem Ferguson <willemferguson@zoology.up.ac.za>

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@glance-

### DIFF
--- a/core/gaspressures.c
+++ b/core/gaspressures.c
@@ -23,6 +23,7 @@
 #include "display.h"
 #include "profile.h"
 #include "gaspressures.h"
+#include "pref.h"
 
 static pr_track_t *pr_track_alloc(int start, int t_start)
 {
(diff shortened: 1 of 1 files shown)


atdotde:
Restarted the Linux build (the others were fine). This idea with the factor is clever. I thought getting this right would be much more complicated. Like it.

PR #1361, opened 2018-06-06
Dive pictures: remove cache_picture() call in dive_add_picture()

bstoeger:
When adding a picture to a dive, cache_picture() was called, which
calculated the hash of the picture in a background-thread.

This made tests occasionally fail, because the tests depended on
the filename-to-localfilename being overwritten in a call running
in a different thread. Depending on which thread finished first,
the test succeeded or failed.

The easiest way to circumvent this problem is to remove the cache_picture()
call. The hash will be calculated anyway with the thumbnails. And
the only function of the hash is the "find moved images" function. Which
is not an issue here, because the user just loaded the images from
disk.

Reported-by: Jan Iversen <jani@apache.org>

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I have a really hard time categorizing this one - but I think "cleanup" is the best fit. It also fixes random test-failures so you might say that it's a (non-user visible) bug fix. And it hashes pictures only once, which might be considered a functional change.

This should speed up addition of pictures a little bit, because every hash is only calculated once instead of twice. In my tests there was no notable difference, presumable because the image data is already in cache. I don't think the removed call should lead to any realistic failure scenario. But another pair of eyes would be most welcome.
 
### Changes made:

Remove `cache_picture()` function.

### Related issues:

Fixes test-failure reported in #1359 

### Release note:

None needed.
### Documentation change:

None needed.
### Mentions:

@atdotde

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -3838,7 +3838,6 @@ void dive_add_picture(struct dive *dive, struct picture *newpic)
 		pic_ptr = &(*pic_ptr)->next;
 	newpic->next = *pic_ptr;
 	*pic_ptr = newpic;
-	cache_picture(newpic);
 	return;
 }
 
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -1216,12 +1216,6 @@ void learnHash(const QString &originalName, const QString &localName, const QByt
 	hashOf[originalName] = hash;
 }
 
-static bool haveHash(const QString &filename)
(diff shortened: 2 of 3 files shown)


dirkhh:
Nice work eliminating a race condition.

PR #1362, opened 2018-06-07
[WIP] iOS: cmake playground

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

update iOS to use cmake and avoid the need for dummy.qml and qml.qrc

Remark this is work in progress, and the PR is mostly used to secure travis
still works


dirkhh:
You seem to have broken more than just iOS :-)
And the need for dummy.qml is MUCH broader than just Googlemaps. It's needed to get all of the qml modules and libraries installed as well.

dirkhh:
+1 on having a playground PR

dirkhh:
The .pro file is exclusively to build Subsurface-mobile on iOS. The only reason it was added was because @tcanabrava and I couldn't figure out how else to create a Qt iOS app.
Given your crossroads, my preference would be a cmake based solution that reuses the existing infrastructure and does away with the need to have a .pro file

janiversen:
Thanks for the answer.

I have a couple of CMake examples even though for smaller projects, so I will use that and see where it takes me. But agreed the pro file is the fastest solution even though a bit limiting (like the demand for dummy.qml).

The CMake solution should be "compatible" with the current solution without need for bigger file movements or real changes in build.sh

The only thing I am considering is to reuse the 3rd sources in src/. I often test build from scratch (and a number of platforms) and it is a PITA to see the cloning take place more than once.

PR #1363, opened 2018-06-08
profile: make it self contained

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

move the code for profile widget to profile widget, in order to make it 
self contained.
Make the needed changed in pro/cmake files

update qml to not call plot dive before the widget is visible. With this PR
plotDive is not longer called during startup, but for some (still) unknown reason
QMLProfile is instantiated 2+ times with every dive change, leading to side effects,
like that the dive id never matches.

All code relating to making (not using) profile should be in the same directory, to
make code searches easier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -285,7 +285,6 @@ endif()
 if(${SUBSURFACE_TARGET_EXECUTABLE} MATCHES "MobileExecutable")
 	set(MOBILE_SRC
 		mobile-widgets/qmlmanager.cpp
-		mobile-widgets/qmlprofile.cpp
 		mobile-widgets/qml/kirigami/src/kirigamiplugin.cpp
 		mobile-widgets/qml/kirigami/src/settings.cpp
 		mobile-widgets/qml/kirigami/src/enums.cpp
@@ -298,6 +297,7 @@ if(${SUBSURFACE_TARGET_EXECUTABLE} MATCHES "MobileExecutable")
 		mobile-widgets/qml/kirigami/src/libkirigami/platformtheme.cpp
 		subsurface-mobile-main.cpp
 		subsurface-mobile-helper.cpp
+		profile-widget/qmlprofile.cpp
 		map-widget/qmlmapwidgethelper.cpp
 	)
 	include_directories(${CMAKE_SOURCE_DIR}/mobile-widgets/qml/kirigami/src/libkirigami)
--- a/mobile-widgets/qml/DiveDetailsView.qml
+++ b/mobile-widgets/qml/DiveDetailsView.qml
@@ -116,31 +116,31 @@ Item {
 			Kirigami.Icon {
 				width: height
 				height: subsurfaceTheme.regularPointSize
-				anchors.verticalCenter: ratingText.verticalCenter 
+				anchors.verticalCenter: ratingText.verticalCenter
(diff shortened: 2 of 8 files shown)


dirkhh:
Travis doesn't like you, @janiversen 
In other news, we really should find a different site to use for libzip - that one times out a lot. Or just cache it in our own storage...

CMakeLists.txt
```diff
@@ -298,6 +297,7 @@ if(${SUBSURFACE_TARGET_EXECUTABLE} MATCHES "MobileExecutable")
 		mobile-widgets/qml/kirigami/src/libkirigami/platformtheme.cpp
 		subsurface-mobile-main.cpp
 		subsurface-mobile-helper.cpp
+		profile-widget/qmlprofile.cpp
```
dirkhh, line 300: Interesting question.
Are we keeping all the qml / mobile UI related code in one place, or all profile code.
Not saying you are wrong, just pointing out the previous logic

dirkhh (approved):
Nice cleanups.
The first commit mixes white space cleanup and code changes (but it's clear enough, so unless you want to practice your git rebase skills it's ok). And of course there is the underlying question of "how do we sort files to be here or there". But I think this makes sense.

janiversen:
@dirkhh I am getting quite sure it is something personal :-) but I agree totally that we should cache it in our own storage, after all it is just a tgz file. We could store it on subsurface-divelog.org and I can change get-dep-lib.sh

I have been having the same problem all afternoon

dirkhh:
My server is very slow - so I pay for Google storage for the bigger files, but maybe something as small as this really can be on my server. Let me try to do this real quick...

CMakeLists.txt
janiversen: I follow your idea, but I think you forgot about map-widget/qml :-) that is not my work.

I have no problem having all qml code (and the related C++ code) in one place, I do however think it is easier to have the widgets (currently map and profile) as self sustained.

janiversen:
who mentioned the word "sponsor" :-)

janiversen:
Another solution would be to clone it to GitHub on our account. That way I could have get-dep-lib do a git clone. Much nicer, and no extra load for your server.

janiversen:
@dirkhh I took you challenge, I do not like rewriting history, but git rebase is very powerful.

With respect to the solution of having all qml in one place, we could easily move map_widget qml to mobile-widgets, however that qml is used by the desktop, so it would be a bad mix.

I see ssrf consisting of 3 widgets (seen from the mobile perspective) that are or should be pretty complex:
- profile (shared a lot of code)
- map (shared nearly all code, incl. qml)
- dive list (not a separate widget at the moment.
My hope is to have those 3 convert to have shared code, with very small exceptions.
The rest are mostly "just" data screen.

That is why I suggest to isolate the 3 widgets in 3 directories together with all code relevant
for each widget.

Having argued for a separation I also see the argument for having all qml in once place
(with the controlling code). So I am not fixed on one solution, just have a high preference :-)

dirkhh:
Nicely done on the rebase. I hope you agree that this made the first two commits so much easier to understand.

janiversen:
Absolutely !!! but we have a saying in Denmark, sometimes you cannot see the forest for all the trees, so review reminders are nice.

PR #1365, opened 2018-06-09
Cleanup: make lastUsedDir() functions static and non-slot, respectively

bstoeger:
The lastUsedDir() functions of MainWindow and Smrtk2ssrfcWindow don't
use any member-objects and are only used in their respective translation
units. Therefore, remove them from the class and made of static linkage.

The lastUsedImageDir() function was declared as a slog, which makes
no sense. Make it a normal static function (though one might argue
why it is assiociated with the DiveListView class in the first place).

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Very minor code cleanup. The `Smrtk2ssrfcWindow` is not tested. I hope Travis compiles this?

### DIFF
--- a/desktop-widgets/divelistview.h
+++ b/desktop-widgets/divelistview.h
@@ -35,6 +35,7 @@ public:
 	void restoreSelection();
 	void contextMenuEvent(QContextMenuEvent *event);
 	QList<dive_trip_t *> selectedTrips();
+	static QString lastUsedImageDir();
 public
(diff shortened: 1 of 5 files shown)


dirkhh:
Travis does indeed compile the Smrtk tool for both Windows and Linux

PR #1366, opened 2018-06-09
core: remove qml ui properties

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

downloadFromComputer contained qml properties which should not be in core and 
also not be compiled as part of the desktop.

The properties are added to qmlmanager

With the PR core is clean from qml properties.

### Changes made:

the commits are made in a sequence in order not to make a "Big Bang" change, but
show the logical change flow.

### DIFF
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -277,16 +277,19 @@ QString DCDeviceData::devBluetoothName() const
 	return m_devBluetoothName;
 }
 
+#ifdef SUBSURFACE_MOBILE
 QString DCDeviceData::descriptor() const
 {
 	return "";
 }
+#endif // SUBSURFACE_MOBILE
 
 bool DCDeviceData::bluetoothMode() const
 {
 	return data.bluetooth_mode;
 }
 
+#ifdef SUBSURFACE_MOBILE
 bool DCDeviceData::forceDownload() const
 {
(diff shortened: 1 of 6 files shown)


dirkhh:
The reason for the old organization of the code was a desire to prevent qmlmanager from becoming too big and to keep the code local to where it was "logically connected". But I can see the logic that you are applying, so I'm ok with this.
Let's wait for Travis (but given the area where you made changes, I doubt it would catch breakage beyond simple compile issues). We don't test UI / functionality after all.

janiversen:
I have tested it on my iPad over days, and also on my Mac due to the lack of functionality in mobile :-)

janiversen:
@dirkh Qt55 with merge errors, I have no idea what is going on.

dirkhh:
What the heck is wrong with the qt55 build. It just hangs when un-shallowing the git tree. That makes no bleeping sense. Restarting *again*

janiversen:
updated my master, rebase in the branch and a push -f, because it seems that the Qt55 problem is merging this PR.

dirkhh:
I can read and parse your words, but they make no sense to me. BTW, I'm on IRC if you want to chat, @janiversen

PR #1367, opened 2018-06-09
build-system: download libzip from our own server

dirkhh:
The libzip.org server often times out and causes build failures.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

I repackaged the two versions that we currently use as .xz (since my server tends to be fairly slow for downloads - compared to GCS or S3 at least).


janiversen:
Tested ok on my server. I have updated my local copy until it gets merged :-)

PR #1376, opened 2018-06-12
ssrf: add makefile at ssrf level

janiversen:
add makefile at ssrf level to ease make calls
it can be called as:
make
make all
- will generate both desktop and mobile
make mobile
- will mobile
make
- will generate desktop

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

add subsurface/makefile to facilitate make calls
the makefile sets LIBRARY_PATH and call make in either mobile or desktop

this is a convenience function.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,0 +1,22 @@
+# SPDX-License-Identifier: GPL-2.0
+
+# Default target executed when no arguments are given to make.
+default_target: all
+
+.PHONY : default_target
+
+# Allow only one "make -f <foo>" at a time, but pass parallelism.
+.NOTPARALLEL:
+
+
+mobile:
+	if test ! -d build-mobile; then (echo "error: please run build.sh before make"; exit -1;); fi
+	cd build-mobile; LIBRARY_PATH=../install_root/lib make 
+	cd build-mobile; LIBRARY_PATH=../install_root/lib make install
+
+desktop:
+	if test ! -d build; then (echo "error: please run build.sh before make"; exit -1;); fi
+	cd build; LIBRARY_PATH=../install_root/lib make 
+	cd build; LIBRARY_PATH=../install_root/lib make install
+
+all: desktop mobile


makefile
```diff
@@ -0,0 +1,20 @@
...
+default_target: all
+
+.PHONY : default_target
+
+# Allow only one "make -f <foo>" at a time, but pass parallelism.
+.NOTPARALLEL:
+
+
+mobile:
+	cd build-mobile; LIBRARY_PATH=../install_root/lib make
```
dirkhh: this will fail until the first time build.sh was run and created that directory

makefile
```diff
@@ -0,0 +1,20 @@
...
+# Allow only one "make -f <foo>" at a time, but pass parallelism.
+.NOTPARALLEL:
+
+
+mobile:
+	cd build-mobile; LIBRARY_PATH=../install_root/lib make 
+	cd build-mobile; LIBRARY_PATH=../install_root/lib make install
+
+desktop:
+	cd build; LIBRARY_PATH=../install_root/lib make
```
dirkhh: ditto

dirkhh (commented):
Hmm. I don't know.
Not sure which type of user / developer this would be useful for. The ones who can't be bothered to read even the most basic documentation, I guess?

makefile
janiversen: It is just a convenience for others and me, who constantly compile the 2 versions.

But no big deal if you do not like, I can do the same with an alias.

makefile
janiversen: correct there should be a test for that.

makefile
dirkhh: It's not that I don't like it. It's that I wonder who'd use it.
But once it's implemented in a sane way so it actually works on a freshly checked out tree... I'm fine.
You don't necessarily have to call build.sh from here - just don't fail with a weird error if someone calls this without calling build.sh first

janiversen: Corrected

dirkhh:
I don't think I'll use it, but I can see how someone might like it this way.

PR #1378, opened 2018-06-13
mobile: "isolate" QML variables used in Settings from QMLManager

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add class QMLPrefs that contain the variables used on the Settings page, in order to reduce the size of QMLManager.

This is the first step in direction of making 1 class the covers the user settings for both desktop and mobile (mobile will still not allow all settings to be set). QMLPrefs and settings.qml will be amended with extra settings for the mobile.

Currently preferences/settings are distributed over 3 places, core/pref.h core/subsurface-qt and mobile-widgets. Q_Property in core/subsurface-qt is compiled into the desktop version, and more importantly the classes are never registred in the mobile version. a second step is to unify these places e.g. use pref.h as base (since it is also used in C code) have a core/settings where all settings are kept (registred as classes in mobile).

### DIFF
--- a/mobile-widgets/qmlprefs.cpp
+++ b/mobile-widgets/qmlprefs.cpp
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "qmlprefs.h"
+
+QMLPrefs *QMLPrefs::m_instance = NULL;
+
+
+QMLPrefs::QMLPrefs()
+{
+	if (!m_instance)
+		m_instance = this;
+}
+
+QMLPrefs::~QMLPrefs()
+{
+	m_instance = NULL;
+}
+
+QMLPrefs *QMLPrefs::instance()
+{
+	return m_instance;
+}
--- a/mobile-widgets/qmlprefs.cpp
+++ b/mobile-widgets/qmlprefs.cpp
@@ -1,11 +1,25 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "qmlprefs.h"
+#include "qmlmanager.h"
 
+#include "core/membuffer.h"
+#include "core/subsurface-qt/SettingsObjectWrapper.h"
+#include "core/gpslocation.h"
+
+
+/*** Global and constructors ***/
 QMLPrefs *QMLPrefs::m_instance = NULL;
 
-
-QMLPrefs::QMLPrefs()
+QMLPrefs::QMLPrefs() :
+	m_credentialStatus(CS_UNKNOWN),
+	m_developer(false),
+	m_distanceThreshold(1000),
+	m_oldStatus(CS_UNKNOWN),
+	m_showPin(false),
+	m_timeThreshold(60)
 {
+	// This strange construct is needed because QMLEngine calls new and that
+	// cannot be overwritten
 	if (!m_instance)
 		m_instance = this;
 }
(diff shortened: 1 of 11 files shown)


mobile-widgets/qmlprefs.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include "core/membuffer.h"
+#include "core/subsurface-qt/SettingsObjectWrapper.h"
+#include "core/gpslocation.h"
+
+
+/*** Global and constructors ***/
+QMLPrefs *QMLPrefs::m_instance = NULL;
+
+QMLPrefs::QMLPrefs() :
+	m_cloudPassword(""),
```
bstoeger: QString() default constructs to the empty string. This line and the next two are not needed.

mobile-widgets/qmlprefs.cpp
```diff
@@ -0,0 +1,197 @@
...
+	m_cloudUserName(""),
+	m_credentialStatus(CS_UNKNOWN),
+	m_developer(false),
+	m_distanceThreshold(1000),
+	m_oldStatus(CS_UNKNOWN),
+	m_showPin(false),
+	m_timeThreshold(60)
+{
+	if (!m_instance)
+		m_instance = this;
```
bstoeger: This is very strange. I know we use something like this for MainWindow and QMLManager, but these are explicitly constructed/destructed. I would prefer not having more of this strange pattern.

Does this even have to be a global object? If yes, why not use the classical construct-on-demand pattern?

mobile-widgets/qmlprefs.cpp
```diff
@@ -0,0 +1,197 @@
...
+	}
+}
+
+void QMLPrefs::setDeveloper(bool value)
+{
+	m_developer = value;
+	emit developerChanged();
+}
+
+int QMLPrefs::distanceThreshold() const
```
bstoeger, line 97: All these getter-functions look so horribly pointless. Are we sure that they are needed?

I vaguely remember that the Q_PROPERTY stuff does not need explicit getters.

mobile-widgets/qmlprefs.cpp
janiversen: Corrected (sorry this was just writing without my head somewhere else)

mobile-widgets/qmlprefs.cpp
janiversen: I added a comment in the code. I would love to avoid this strange construct, but QMLEngine makes a new QMLPrefs and it does need to be a global object, at least for now, otherwise the variables may have different values.

I have searched for a pattern, where the constructor is called (with new ...) but the existing this pointer is returned, in practice make a singleton for several new calls.

Suggestions are very welcome.

mobile-widgets/qmlprefs.cpp
janiversen: Yes Q_PROPERTY can be written with MEMBER, as you can see in several places, but I needed to add getter functions for those variables used by other objects (in this case QMLManager).

I preferred a few getter functions to make QMLManager friend, but have no problem changing to that.

Medium term, all these variables should go into prefs and thus be shared without getter.

janiversen:
Regarding the somewhat strange constructor, I have been considering overloading new/delete for the class and let it construct first time and otherwise return m_instance.

@bstoeger Do you see that as a good alternative (I was strongly in doubt) ?

bstoeger:
> I have been considering overloading new/delete 

Please don't.

janiversen:
I totally agree. Once this PR is merged I will move on and make the classes with Q_PROPERTY data less (refencing the global prefs structure), and then it does not matter if the are multiple objects.

The interface to QML is very sensitive e.g. move a statement in main.qml and the objects are created in a different order.

janiversen:
data less == without data

mobile-widgets/qmlprefs.cpp
dirkhh: we ended up with this pattern because otherwise we weren't able to access the object from C++. Somehow the QML code instantiated its own copy of the class and we were forever trying to debug why changes that we made in C++ weren't reflected in QML

dirkhh (commented):
It seems very odd that one commit copies the code into the new files, and a couple commits later that same code is deleted at its original location.
I think it would make a lot more sense from a readability perspective to copy the code (and remove from the original spot) in the same commit.

mobile-widgets/qmlprefs.cpp
janiversen: Exactly what I experienced, the only sane way to avoid an ugly constructor is to use a interface/implementation pattern, which we have 98% in using pref.h. I wanted first to showcase that a separation from QMLManager works (this PR), and then do the full implementation.

janiversen:
I did it that way, step-by-step, to secure I had a working system after each commit. But surely I can squash the commit if you think that is better, no problem.

janiversen:
Had to squash a little more, to make it work. This commit is a bit big for my taste, but as you wrote it is one logical operation (a move).

@dirkhh A question, I am used to testing after each commit that the code still builds and runs, however I am not sure it is seen as a "demand" in ssrf ?

janiversen:
@dirkhh Failure only in linux build: 11 - TestMerge (OTHER_FAULT)
It worked before I squashed the commits, so I hope it is caused by something else?

janiversen:
@dirkhh should I try a restart, by making a pseudo change and pushing?

dirkhh:
OK, commits look much nicer now. I think this is good to go.
@bstoeger are you ok with the explanations as well? Anything else you're concerned about?

bstoeger:
I don't have anything to object. But then I'm not really sure what this does, and the whole QML-machinery is extremely obscure to me. From the few things that I've seen, the interaction with C++ is very awkward. At the moment I can't even makes experiments, because the mobile-on-desktop app crashes owing to messed up (k)ubuntu packages.

janiversen:
@bstoeger I can only confirm your thoughts about the QML - C++ interface it is really something not very stable. 

I am working on a larger PR, that sits on top of this PR and #1379 to make the variable transfer less "magic" and a lot safer. The new PR not only gets rid of the strange constructor, but also makes sure whatever the QML code contains it will work (try e.g. in the current with 2 time QMLManager in main.qml, or just move it around).

PR #1379, opened 2018-06-13
core: make prefs-macros developer secure :-)

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

add {} to each macro in prefs-macros, in order to avoid unexpected effects of
e.g.
   if ()
      GET_TXT();

with the PR is works as expected

### Changes made:

1) add {}
2) correct code indent and backslash indent

### DIFF
--- a/core/prefs-macros.h
+++ b/core/prefs-macros.h
@@ -8,79 +8,101 @@
 #define SB(V, B) s.setValue(V, (int)(B->isChecked() ? 1 : 0))
 
 #define GET_UNIT(name, field, f, t)                                 \
+	{															\
 	v = s.value(QString(name));                                 \
 	if (v.isValid())                                            \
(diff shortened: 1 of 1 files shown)


dirkhh (changes requested):
So the first commit seems perfectly sane and obviously a good idea. The second commit, when looked at on GitHub, makes me wonder about "sanity".
Does your editor have a different tab width? This doesn't look too "cleanup" to me... :-)

janiversen:
I do not know what happened, let me remove the second commit, it is not important.

janiversen:
Done.

dirkhh (approved):

PR #1380, opened 2018-06-15
Cleanup: Don't clear WA_QuitOnClose attribute on dialogs

bstoeger:
According to Qt's documentation, the application exits if all windows
with the WA_QuitOnClose attribute are closed. This attribute was cleared
for three dialogs. This seems not necessary because:
1) The application can't be closed as long as the modal dialog is shown.
2) The flag only concerns primary windows, which these are not.

See: http://doc.qt.io/qt-5/qguiapplication.html#quitOnLastWindowClosed-prop

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes the clearing of the `WA_QuitOnClose` flag for three dialogs. To my understanding, this it not necessary. The flag is not relevant, because the dialogs have a parent and you can't close the application anyway while they are shown.

This was introduced in 880b8394d2b063c1bbd6597c2cb6c51f0a1d66c1 which has a strange commit message:
> To solve that we mark all custom windows/dialogs with the Qt::WA_QuitOnClose attribute on instance creation.

But actually the commit *clears* the flag.

> Most child windows should be closed with the main application window otherwise if left open and if making specific modifictions could potentially cause a SIGSEGV.

But - at least according to the docs - the flag in question does the *opposite*: It closes the application if the marked windows are closed, not vice-versa.

### Release note:

Not needed.

### Documentation change:

Not needed.

### Mentions:

@neolit123: Original commit is by you. Please check

### DIFF
--- a/desktop-widgets/divecomputermanagementdialog.cpp
+++ b/desktop-widgets/divecomputermanagementdialog.cpp
@@ -28,7 +28,6 @@ void DiveComputerManagementDialog::init()
 DiveComputerManagementDialog *DiveComputerManagementDialog::instance()
 {
 	static DiveComputerManagementDialog *self = new DiveComputerManagementDialog(MainWindow::instance());
-	self->setAttribute(Qt::WA_QuitOnClose, false);
 	return self;
 }
 
--- a/desktop-widgets/diveshareexportdialog.cpp
+++ b/desktop-widgets/diveshareexportdialog.cpp
@@ -32,7 +32,6 @@ void DiveShareExportDialog::UIDFromBrowser()
 DiveShareExportDialog *DiveShareExportDialog::instance()
 {
 	static DiveShareExportDialog *self = new DiveShareExportDialog(MainWindow::instance());
-	self->setAttribute(Qt::WA_QuitOnClose, false);
 	self->ui->txtResult->setHtml("");
 	self->ui->buttonBox->setStandardButtons(QDialogButtonBox::Cancel);
 	return self;
--- a/desktop-widgets/subsurfacewebservices.cpp
+++ b/desktop-widgets/subsurfacewebservices.cpp
@@ -691,7 +691,6 @@ out:
 DivelogsDeWebServices *DivelogsDeWebServices::instance()
 {
 	static DivelogsDeWebServices *self = new DivelogsDeWebServices(MainWindow::instance());
-	self->setAttribute(Qt::WA_QuitOnClose, false);
 	return self;
 }
 


dirkhh:
a restart fixed the Travis issue. @neolit123 this is your code - comments? It looks fine to me.

bstoeger:
> so, you open a child window, which for some reason is non-modal (and i think we had parallel problem with modality here) and you close the main window, but the child window remains and can SIGSEGV because MainWindow was destroyed and it wants to call MainWindow::instance()->doStuff() and because the attribute in question is true by default.

This explanation doesn't make sense to me for 3 reasons:
1) If I read the documentation correctly, the `WA_QuitOnClose` flag concerns only "primary" windows, i.e. without parent. But all three dialogs have `MainWindow` as a parent.
2) Even if the dialogs were primary: by clearing `WA_QuitOnClose`, the application would quit even if the dialog still exist. So clearing the flag would have the *opposite* effect.
3) Since `MainWindow` is the parent of the dialogs in question, on destruction of `MainWindow` also the dialogs are destructed. I can't see how the dialogs could call into `MainWindow` after its destruction.

> the "perfect" way to test your patch @bstoeger is to to open all the affected dialogs and test for the modality on all our desktop OS-es with the oldest Qt we support there. if this works i'm ok with the cleanup.

I chose to do something different: I set the modality to false with:
```
self->setModal(false);
```
in `DiveComputerManagementDialog::instance()`
Then opened the device editing dialog (which is now non-modal) and quit the application (from the menu and by closing the main window). No segfault in either case.

Just to be sure, I also added a destructor for `DiveComputerManagementDialog` with a `printf`. Indeed the dialog is destructed.

I can try to repeat this experiment with the two other dialogs if you want.

dirkhh:
I think part of the problem really is that this was with Qt 4.8 back then and semantics may have changed and bugs may have changed.
Looking at the documentation I think what @bstoeger is saying makes perfect sense. But I distinctly remember the crazy crashes we had when you closed the main window while a dialog was open and we did all these weird things to avoid those crashes...

neolit123:
your 1,2,3 points form a valid argument except that it didn't work as described. also, i think i have some objections but i don't want to dig into that.

we had problems with "some" dialogs 5 years ago with an old Qt, possibly only a single OS or maybe all of them as i haven't pointed out that in the commit message.
why this was the fix i chose, i don't remember, but it worked as far as i remember.

i see that your argument is that the fix wasn't a fix?

> I can try to repeat this experiment with the two other dialogs if you want.

instead of that why not test your patch on top of the oldest Qt version we support on Linux (5.5). if that works possibly all is good...but we still have to test the latest Qt we use for OSX and Windows just to be sure.

can you help with the above?

dirkhh:
Qt 5.5 is the oldest we support anywhere - and almost all of our binaries are built against 5.9 or later.
The 5.5 support is mainly important for "in distro" binaries for older Linux distros.

janiversen:
I use Qt 5.10 on my mac, and will give it a spin this weekend.

rgds
jan i

Enviado desde mi iPad

bstoeger:
> i see that your argument is that the fix wasn't a fix?

That's not my argument. I can't tell what, if anything, the commit from 2013 fixed. I doubt that it's worth finding out.

My argument is that in the current code base this doesn't do anything. Especially, the failure case you describe (dialog calling into destructed `MainWindow`) can never happen, because the dialogs in question are destroyed with `MainWindow`. If this really was the issue, then it has been fixed properly.

If the failure case was that the application doesn't close when there is an open dialog - according to my quick glance of Qt's source this shouldn't happen anymore.

> instead of that why not test your patch on top of the oldest Qt version we support on Linux (5.5)

I don't have a Qt 5.5 and compiling from source on my laptop is pretty much out of the question. Travis doesn't seem to generate the appropriate binaries. Moreover, I don't even know what I should test, as the described failure mode can not happen.

> I use Qt 5.10 on my mac, and will give it a spin this weekend.

Thanks, I can imagine this was some Mac issue, as Macs behave quite differently when it comes to window handling.

bstoeger:
Just to check, I downloaded `qtbase-opensource-src-5.5.0` and searched for all instances of `WA_QuitOnClose`. The attribute is tested in two functions:
1) `bool QWidgetPrivate::close_helper(CloseMode mode)`:
```
    bool quitOnClose = q->testAttribute(Qt::WA_QuitOnClose);
    ...
    quitOnClose = quitOnClose && (parentWidget.isNull() || !parentWidget->isVisible());
    ...
    if (!w->isVisible() || w->parentWidget() || !w->testAttribute(Qt::WA_QuitOnClose))
```
2) `bool QApplicationPrivate::shouldQuit()`:
```
    if (w->isVisible() && !w->parentWidget() && w->testAttribute(Qt::WA_QuitOnClose))
```

Note that if `Widget::parentWidget` is non-null, the attribute has no effect anywhere in the Qt 5.5 code base. Unless I'm missing something, setting/clearing this flag is a no-op for dialogs with a parent.

dirkhh:
this seems like the correct thing to do.

PR #1382, opened 2018-06-16
DM5 import: support for larger sample size

mturkia:
Suunto has updated SampleBlob to use 30 byte blobs. This adds support
for the increased size. Note that this only parses the same fields we
have parsed before. (Currently I have no idea what the increased size is
used for.)

Note also that I do not currently have data with the new format so I
only tested this still works with old data.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/import-suunto.c
+++ b/core/import-suunto.c
@@ -458,6 +458,11 @@ extern int dm5_dive(void *param, int columns, char **data, char **column)
 				tempformat = 1;
 				block_size = 26;
 				break;
+			case 5:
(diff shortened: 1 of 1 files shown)


dirkhh:
As you said on the mailing list - test data and a test to make sure this works would be great, but I see no reason not to pull this. Thank you

PR #1383, opened 2018-06-16
desktop: better handling of progress dialog width

dirkhh:
We shouldn't need to manually set it, but it appears we do. To avoid
constant resizing, let's only grow it - and let's set the size before we
update the text.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The existing code caused the dialog to shrink and expand which lead to odd visual artifacts. This looks much better.
### Changes made:

1) set the size before setting the label
2) only resize if we need more space (i.e., don't shrink the dialog)
### Mentions:

@atdotde this fixes something that you introduced a few months ago. Please check that things still work for you

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -77,10 +77,13 @@ extern "C" int updateProgress(const char *text)
 	if (verbose)
 		qDebug() << "git storage:" << text;
 	if (progressDialog) {
+		// apparently we don't always get enough space to show the full label
+		// so let's manually make enough space (but don't shrink the existing size)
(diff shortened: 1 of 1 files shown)


atdotde:
I agree this should work at least as good as what we have before. The problem is that this whole QFontMetrics.width() thing appears to be totally broken on Mac. And you see it most clearly by setting the font size to huge (I played with 20pt). With that size (which is clearly too big for our UI), text gets cut off in master as well as in your branch. So. let's merge your branch. A bit of googling show that others complain about this being a Mac issue but this does not mean it get's addressed by Qt people.

Issue #1331, opened 2018-05-26
mobile: black subsurface logo due to Kirigami bug


mfbernardes:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Since the last Kirigami update the subsurface logo is black, as below:

 ![img_4173](https://user-images.githubusercontent.com/30494/39968147-d4da7d4c-56c8-11e8-95a1-d75eff06bf7a.jpg)

This is due to Kirigami having changed the Icon implementation, and erroneously tinting icons that should not be tinted.

I'll report and track with Kirigami dev.

### Operating system:

iOS 11.3

### Mentions:

@notmart

mfbernardes:
I've split #1267 because I found the cause of different icons missing. This one is unrelated to the others.

Some tests reported on #1307

neolit123:
@janiversen @mfbernardes 
please, close this issue if it's a duplicate or if we have a fix in `master` already.
thanks.

janiversen:
@neolit123 I just tested with master on my iPhone, problem is still there, just as @mfbernardes describes it. The idea it is being tinted, sounds very logical from the debugging I have done (the image is loaded from disk/memory, but then changed on its way).

PR #1385, opened 2018-06-16
Roll our own banner

dirkhh:

- [x] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Kirigami has a bug that causes us not to show the Subsurface-mobile icon in the global drawer #1331
To work around this we are rolling our own banner image, icon and text.
And because we are doing that, we get more flexibility with what we are showing there and add the cloud credentials as second line
### Changes made:

1) stop setting title and related info for GlobalDrawer
2) create our own banner with background image, gradient overlay, icon and title
3) create second text line with login ID
### Mentions:

@mfbernardes this should fix the problem you reported


janiversen:
Works for me

mfbernardes:
I just built it and for some reason no header appears.

img_4301

dirkhh:
Hmm - I had that on and off during my attempts to make it all work. Is there anything useful in the AppLog?

mfbernardes:
Not really... 
Just some binding loops:
```
INFO: qrc:///qml/main.qml:110:15: QML 
Image: Binding loop detected for property "width"
INFO: "1.745: AppState changed to active with no save ongoing and no unsaved changes"
INFO: qrc:/org/kde/kirigami/GlobalDrawer.qml:221:13: QML ColumnLayout: Binding loop detected for property "height"
```

Building with Qt 5.11, if that makes any difference in this case.

dirkhh:
I think that shows the problem. It can't figure out the width for the topComponent.
If you set a fixed width, does it show up?

mfbernardes:
You're right.

Setting a fixed width make it show up, but with a top and left margin...

simulator screen shot - iphone x - 2018-06-16 at 14 14 38

dirkhh:
OK - that should help me get closer...

mfbernardes:
Just:

```
+++ b/mobile-widgets/qml/main.qml
@@ -109,7 +109,7 @@ Kirigami.ApplicationWindow {
                topContent: Image {
                        source: "qrc:/qml/icons/dive.jpg"
-                       width: parent.width
+                       width: 300
                        sourceSize.width: width
                        fillMode: Image.PreserveAspectFit
                        asynchronous: true
```

mfbernardes:
Since it's inside the topContent the margin/padding is already set.

The best I could get was with:

```
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -109,7 +109,8 @@ Kirigami.ApplicationWindow {
                topContent: Image {
                        source: "qrc:/qml/icons/dive.jpg"
-                       width: parent.width
+                       Layout.fillWidth: true
+                       height: textblock.height
                        sourceSize.width: width
                        fillMode: Image.PreserveAspectFit
                        asynchronous: true
```

dirkhh:
You are right - I need to hack Kirigami in order for this to work as intended.
Ugly, but seems to work great... let me see if I can package this in a way that isn't too scary. Fundamentally, since we fix a specific version of Kirigami already, this isn't quite as fragile as it sounds.

dirkhh:
I have partially fixed this, @mfbernardes please take a look what you think - maybe you can figure out how to get rid of the left margin :-)

dirkhh:
@bstoeger as our resident C++ expert... can you explain to me why the linux2 build fails on Travis? 
this is building mobile on desktop (Linux) and can't find a matching connect function... yet that same code compiles fine for Android and iOS
And of course this code happily compiles here on my Linux (admittedly much newer ArchLinux).

bstoeger:
> @bstoeger as our resident C++ expert...

Puh. Not really. :(

This is ominous. What happens if you replace
```
            connect(m_item, &QObject::destroyed, ctx, &QObject::deleteLater);
```
by
```
            connect(m_item.data(), &QObject::destroyed, ctx, &QObject::deleteLater);
```
?

I can't compile this PR, I immediately get:
```
-- GCC optimization flags: -O2
-- Building Subsurface-mobile requires BT support
-- system name Linux
-- Configuring done
CMake Error at CMakeLists.txt:313 (add_executable):
  Cannot find source file:

    mobile-widgets/qml/kirigami/src/libkirigami/tabletmodewatcher.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx


-- Generating done
-- Build files have been written to: /home/bs/src/subsurface/build-mobile
Makefile:1447: recipe for target 'cmake_check_build_system' failed
```

dirkhh:
There's an issue with our `get-dep-lib.sh` script - the PR to fix it isn't quite ready. So you'll have to manually update Kirigami to the right SHA:
```
cd kirigami
git pull
git checkout 8b803828ad9ea67559a64a93c927862d6b19c96e
cd subsurface
./scripts/mobilecomponents.sh
```
that should do the trick for now to get this PR to build

mfbernardes:
@dirkhh this actually worked fine, no left margin.

img_4302

There is still a binding loop. I could not figure why yet, but I'm yet to have my morning coffee so I'm might be too asleep for that.

INFO: qrc:/org/kde/kirigami/GlobalDrawer.qml:221:13: QML ColumnLayout: Binding loop detected for property "height"

dirkhh:
Great to see that this worked for you. It works for me perfectly on the Mac, there's an odd margin on Linux. And of course we still have the failing test... but I think once I understand why the test fails (hoping for @bstoeger ) this is ready to be pulled...

bstoeger:
It compiles for me as well, which compiler version is this? Seems like a compiler bug.

Have you tried the `connect(m_item, ...);` -> `connect(m_item.data(), ...);` change?

If that doesn't work, you could still try
```
connect(m_item.data(), &QQuickItem::destroyed, ctx, &QQmlContext::deleteLater);
```
to make it very clear to the compiler what you mean.

mfbernardes:
@dirkhh I'm not sure you have seen the new "back button" Kirigami added. It appears on iOS, but based on kirigami's current logic would not show up on Android:
img_4310

I hate it particularly on a small device, but some people seemed to like it.

janiversen:
And (at least to me) the 2 ">" is strange. Once we come past the release we should start a brainstorm, about what should be changed in the design.

dirkhh:
@mfbernardes this does look odd. where does the '>' come from? 
On happy news, this now passed all tests (needed a newer compiler than what Trusty has).
I'll merge this as I consider it a required improvement over not having a logo :-)

mfbernardes:
@dirkhh Testing 2.1.0 from TestFlight: the banner rarely shows up. Most of the times I open the app there is no banner.

PR #1381, opened 2018-06-15
Cleanup: Make DiveComputerModel more sensible

bstoeger:
### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

`DiveComputerModel` is a mess. This PR tries to make it at least somewhat sane by implementing proper `QtModel` semantics and removing the `QMultiMap`->`QMap` conversions.
Moreover, a naked pointer is replaced by a `QScopedPointer` to remove a `delete`. 

Note that since I possess only one DC, this is only very lightly tested.

### Changes made:

1) Implement `QtModel` semantics in `DiveComputerModel`
2) Remove the quasi-noop function `DiveComputerModel::update`
3) Replace a naked pointer by a `QScopedPointer`

### Related issues:

I didn't find an outright bug, but perhaps this indirectly addresses #1278
 
### Release note:

None needed
### Documentation change:

None needed

Issue #1278, opened 2018-05-15
linux: crash when deleting dive computers


neolit123:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

A user reported at the user forum:

> This is not a critical issue, but when clearing out some old backup dive computers from my device list, I get a segfault straight after confirming deletion. I'm using subsurface 4.7.8 on Mint Linux 18.3. Problem doesn't appear on my Win10 laptop.

### Operating system:

Linux Mint

### Subsurface version:

Subsurface from PPA 4.7.8.

### Current behavior:

crash when deleting DC.

### Expected behavior:

no crash.

### Additional information:

stack trace:
https://pastebin.com/712Gvaax

the crash seems to happen somwhere in:
https://github.com/Subsurface-divelog/subsurface/blob/master/qt-models/divecomputermodel.cpp

bstoeger:
Could they perhaps check current master? The `DiveComputerModel` was refactored in #1386.

neolit123:
cannot be reproduced on Linux with latest master.

PR #1386, opened 2018-06-16
[RFC][Alternative to #1381]Turn dive computer map into sorted vector

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a more radical alternative to #1381: Turn the map of dive computers into a sorted list. Thus conversions map<->list or awkward stepping through the map are avoided. As opposed to #1381, this does not use any template magic, so that might be perceived as an advantage. On the flip side, it uses binary search instead of associative arrays to access the dive computers, which might be unfamiliar to some.

I tested it only lightly, because I'm mostly interested in comments. 
### Changes made:

1) Turns the dive computer map into list
2) Implement a proper Qt-style model/view interface for the `DiveComputerModel`
3) Remove unnecessary constructors, use a `QScopedPointer` and other minor cleanups
### Related issues:

Might address #1278 
Alternative to #1381

### Release note:

None needed.
### Documentation change:

None needed

### DIFF
--- a/core/divecomputer.cpp
+++ b/core/divecomputer.cpp
@@ -19,6 +19,11 @@ bool DiveComputerNode::operator!=(const DiveComputerNode &a) const
 	return !(*this == a);
 }
 
+bool DiveComputerNode::operator<(const DiveComputerNode &a) const
+{
+	return std::tie(model, deviceId) < std::tie(a.model, a.deviceId);
+}
+
 bool DiveComputerNode::changesValues(const DiveComputerNode &b) const
 {
 	if (model != b.model || deviceId != b.deviceId) {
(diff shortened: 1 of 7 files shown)


janiversen:
Templates are even better than (my) #define macros to make code "specialist touch only". Templates are relevant in many cases, but can also be used to make the code less readable.

Does it really matter for performance if we use one or the other search, please remember our list of supported DC's are rather small and most people only have one DC. I would anytime go with the search that has the simplest code (binary search).

I have not tested to code, but only looked it over (while the system is busy building), and I like the code.

So all in all, my preference is clearly this one over #1381

bstoeger:
> Templates are even better than (my) #define macros to make code "specialist touch only". Templates are relevant in many cases, but can also be used to make the code less readable.

I'm not sure about that. I think the template in #1381 is more readable than the equivalent macro solution.

> Does it really matter for performance if we use one or the other search

No, performance is irrelevant in this case. It's about "clean" code. We have `DiveComputerList` which takes care of the dive computers and `DiveComputerModel` which provides access to the dive computers. The former uses a `QMultiMap`, the latter needs index-based access. The old code therefore converted the map to a list on every function invocation. I see three possibilities to make this more clear (and also improve performance as a side-effect):
1) Step through `QMultiMap` with `std::next` (#1381).
2) Universally use a sorted vector (this PR).
3) Keep the map-to-vector conversion, but do it only in the constructor and when committing changes.

> So all in all, my preference is clearly this one over #1381

I think me too. But I fear that some might be unfamiliar with the keep-a-sorted-vector idiom.

dirkhh:
This is simply a comment on me being fundamentally a C programmer who writes C++ if he has to. The new code is really hard for me to read and just understand what it does. I realize that is first and foremost MY problem. But I worry about that. Happy to hear what others think, but I'll have to think about it some more before (and not at 5am) to know if I'd feel comfortable editing this in the future.

neolit123:
ACK.

> The new code is really hard for me to read and just understand what it does

jumping on the hard-to-read C++ bandwagon was inevitable the moment we introduced C++ - it just has so many cool features. i'd take this binary search with `lower_bound + operator<` over the `decltype + template + auto` slightly insane syntax in the other PR (https://github.com/Subsurface-divelog/subsurface/pull/1381).

dirkhh:
Let me ask you this way, @neolit123 - do you think you understand / could fix bugs in this code?
That's my biggest worry.
I don't see anything wrong, I can read it and guess what it does, but... I guess I'm just being a chicken and should jump in, head first, into the murky water of unknown depth...

neolit123:
> Let me ask you this way, @neolit123 - do you think you understand / could fix bugs in this code?

@dirkhh 
yes, possibly.

i have a proposal too - why not make @bstoeger a collaborator?
he has contributed enough changes in a lot of different areas and seems to understand the C++ trade quite well.

edit: perhaps he is a diver too - i never asked?

dirkhh:
I guess that's a good way to think about it. @bstoeger certainly has been around for a while now and has consistently contributed at a very high level. So I should just ask him to commit to a little more :-)

dirkhh:
Why were you ice diving? The water here is 25°C and that really is the coldest one should do - unless one is wearing a dry suit... Too bad you can't make it to Bonaire. You'd like the water / viz there :-)

PR #1387, opened 2018-06-16
[wip] build-system: update git checkout

mfbernardes:
When any of the tags/commits are updated the script needs to
make sure the correct version is checked out, even if the
directory already exists.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When updating any of the tags/commits we use, the current script would not ensure they are updated unless you're starting from clean build.

### Mentions:

@janiversen

### DIFF
--- a/scripts/get-dep-lib.sh
+++ b/scripts/get-dep-lib.sh
@@ -38,10 +38,15 @@ git_checkout_library() {
 	fi
 	pushd "$name"
 
-	git fetch origin
-	if ! git checkout "$version" ; then
-		echo "Can't find the right tag in $name - giving up"
-		return -1
+	local current_sha=$(git rev-parse HEAD)
+	local target_sha=$(git rev-parse "$version")
+
+	if [ ! "$current_sha" = "$target_sha" ] ; then
+		git fetch origin
+		if ! git checkout "$version" ; then
+			echo "Can't find the right tag in $name - giving up"
+			return -1
+		fi
 	fi
 	popd
 }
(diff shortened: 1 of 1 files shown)


janiversen:
Checking out every time is of course less than cloning, but it is still lot to do for something that happens with months distance.

The patch is also not complete e.g. libgit and libftdi are missing and those would cause a download every time.

May I suggest an alternative patch, that solves your problem without adding cpu cycles.

1) when git clone / git checkout executes for a library, do something like "echo $CURRENT_LIBNAME > install_root/$LIBNAME"
2) compare install_root/$LIBNAME to $CURRENT_LIBNAME and if different do the clone etc

Wdyt ?

mfbernardes:
@janiversen agree with avoiding unnecessary checkout, but even if it's months apart when it happens it should be right.

I have a new proposal for the ones we download from git. 

Didn't do anything about the ones we download tarballs yet, but might be in the lines of what you mentioned.

janiversen:
I like the new function (I never came around to writing it) that makes that part a lot more consistent.

I also like your solution to check git based libraries.

But now people (like myself) get confused if e.g. libzip was to change :-) It looks easy to write a similar function because the download tar file stays in src/ so you can compare against that. If a user removes the tar file it is his problem that it is downloaded again.

dirkhh:
After reading this more carefully, I think this is worth merging now (assuming the Travis error clears up - it looked like one of those weird random Test failures that we still haven't solved). Yes, it doesn't handle the tar files, yet, but it's a huge improvement over what we have - and with Kirigami being updated in one of my PRs, I'd love for this to be in master by the time that PR gets merged :-)

PR #1388, opened 2018-06-16
Update to CHANGELOG.md

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

In preparation of release: changelog update for recent PRs by myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,15 @@
 - Shearwater import: include DC reported ceiling
+- Dive pictures: don't read invalid files twice
+- Dive pictures: don't plot pictures twice on switching dives
+- Dive pictures: don't replot all pictures on deletion
+- Dive pictures: remove stale code for cloud storage of pictures
+- Dive pictures: save thumbnails in individual files
+- Dive pictures: don't keep all thumbnails in memory
+- Dive pictures: automatically recalculate thumbnail if older than picture
+- Planner: fix output of ICD-notes
+- Desktop: fix crash on manual addition of dive to fresh logbook
+- Desktop: fix showing of the manual
+- Desktop: don't save empty file to cloud [#1228]
 - Desktop: allow editing from list-view [#1213]
 - Desktop: don't add six seconds to dive duration [#554]
 - Desktop: fix manual editing of dive duration [#1211]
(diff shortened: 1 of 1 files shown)


janiversen:
Just a question, are we all supposed to add snippets one by one to CHANGELOG.md or is someone pulling all the PR's that have been merged since v.4.7.8 ?

bstoeger:
I think the idea is that you provide a CHANGELOG.md update with every PR. Don't take my laziness as an example. :(

janiversen:
UPS, I think I learned too fast :-)

had a quick look at the merged PR´s, seems adding CHANGELOG.md is rather the exception. Suppose @dirkhh will remind us.

atdotde:
I guess you are supposed to add a line for everything noticible by the user (new feature, bug fixed, general improvement like speed up). But I would say one per feature or related group of commits, not one per commit. Changelog is for users to read to see why it’s worth updating to the new release.

janiversen:
Ok, the only user facing PR I have made is the log to clipboard feature, which might be worth noting. I will make a PR tomorrow.

dirkhh:
The problem is that no, I don't always remind people, nor do the other maintainers who merge PRs.
It's something we need to get better at as a team.

dirkhh:
The proposed changes are ok with me, but I'll admit that they are too detailed for me to use them in the ReleaseNotes - I always think of this as "what matters to a user". Many of the entries here are certainly meaningful to a developer - but those should use `git log` to figure out what's going on.
We can merge this (as it's better than what we have), or maybe we can uplevel this a little to make it more useful when creating a release in a couple of weeks...

bstoeger:
OK, I will try to condense this. But I'm not 100% sure about the appropriate level of detail. For example, certain bugs (e.g. "fix showing of the manual") might be relevant to some. But one could also classify it under "various minor bug fixes". So in such cases I'd tend towards more detail and leave the final editing to your discretion.

dirkhh:
Details are fine - it's always a judgment call. Typically things like memory leaks and implementation details are stuff that I would skip.

dirkhh:
I looked again and decided to just pull it - GitHub was confused, so I pulled it manually

PR #1390, opened 2018-06-17
build-system: add check for tar files in get-dep-lib

janiversen:
Complemented the git check with a similar tar check.
check if the correct download file is present if not download and expand it

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

@mfbernardes made a nice enhancement of get-dep-lib.sh, this PR completes it by
also handling tar files.

It is checked that the lib<foo><version> is present in src/ if not it is downloaded and expanded

### DIFF
--- a/scripts/get-dep-lib.sh
+++ b/scripts/get-dep-lib.sh
@@ -158,20 +158,26 @@ if [[ "$BUILD" = *"openssl"* ]]; then
 	git_checkout_library openssl $CURRENT_OPENSSL https://github.com/openssl/openssl.git
 fi
 
-if [[ "$BUILD" = *"libzip"* && ! -d libzip ]]; then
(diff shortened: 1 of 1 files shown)


mfbernardes:
I have a different approach for this at #1392.

dirkhh:
I like #1392 @mfbernardes did - except that something is wrong with ftdi - see the Android Travis failure...

janiversen:
No hard feeling here, if you prefer that. #1392 is nice because it makes a function, but my patch does without an extra file. Best thing would be if @mfbernardes would use the tar file test I do.

janiversen:
Duplicate to #1392, which uses a better approach (functions)

PR #1391, opened 2018-06-17
Use correct "tr" call for translating dive mode names

sfuchs79:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [x] Language translation

Correctly use gettextFromC::instance()->tr(); instead of a simple
tr(); to translate the dive mode names.
This goes on top of 0bc9edf855caf95e2d58ccaf704ceeda079e06f1
and finally makes the whole thing work.

### DIFF
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -5,6 +5,7 @@
 #include "core/planner.h"
 #include "core/qthelper.h"
 #include "core/subsurface-qt/SettingsObjectWrapper.h"
+#include "core/gettextfromc.h"
 
 #include "qt-models/cylindermodel.h"
 #include "qt-models/models.h"
@@ -448,7 +449,7 @@ PlannerSettingsWidget::PlannerSettingsWidget(QWidget *parent, Qt::WindowFlags f)
 
 	// should be the same order as in dive_comp_type!
 	for (int i=0; i < FREEDIVE; i++)
-		rebreather_modes.append(QString(tr(divemode_text_ui[i])));
+		rebreather_modes.append(gettextFromC::instance()->tr(divemode_text_ui[i]));
 	ui.rebreathermode->insertItems(0, rebreather_modes);
 
 	modeMapper = new QSignalMapper(this);
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -449,7 +449,7 @@ PlannerSettingsWidget::PlannerSettingsWidget(QWidget *parent, Qt::WindowFlags f)
 
 	// should be the same order as in dive_comp_type!
 	for (int i=0; i < FREEDIVE; i++)
-		rebreather_modes.append(gettextFromC::instance()->tr(divemode_text_ui[i]));
+		rebreather_modes.append(gettextFromC::tr(divemode_text_ui[i]));
 	ui.rebreathermode->insertItems(0, rebreather_modes);
 
 	modeMapper = new QSignalMapper(this);
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -112,7 +112,7 @@ MainTab::MainTab(QWidget *parent) : QTabWidget(parent),
 	// This needs to be the same order as enum dive_comp_type in dive.h!
 	QStringList types = QStringList();
 	for (int i = 0; i < NUM_DIVEMODE; i++)
-		types.append(gettextFromC::instance()->tr(divemode_text_ui[i]));
+		types.append(gettextFromC::tr(divemode_text_ui[i]));
 	ui.DiveType->insertItems(0, types);
 	connect(ui.DiveType, SIGNAL(currentIndexChanged(int)), this, SLOT(divetype_Changed(int)));
 
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -24,6 +24,7 @@
 #include "qt-models/filtermodels.h"
 #include "core/divesite.h"
 #include "core/subsurface-string.h"
+#include "core/gettextfromc.h"
 #include "desktop-widgets/locationinformation.h"
 
 #include "TabDiveExtraInfo.h"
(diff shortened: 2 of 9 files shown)


bstoeger:
@sfuchs79: Can you explain why this fixes the problem to someone who doesn't understand the translation machinery?

Because at a first glance, this change looks dubious. Note that `tr(...)` is static, so you shouldn't need an instance. It seems to me this should rather read as `gettextFromC::tr(...)`?

desktop-widgets/diveplanner.cpp
```diff
@@ -448,7 +449,7 @@ PlannerSettingsWidget::PlannerSettingsWidget(QWidget *parent, Qt::WindowFlags f)
 
 	// should be the same order as in dive_comp_type!
 	for (int i=0; i < FREEDIVE; i++)
-		rebreather_modes.append(QString(tr(divemode_text_ui[i])));
+		rebreather_modes.append(QString(gettextFromC::instance()->tr(divemode_text_ui[i])));
```
bstoeger: `tr(...)` returns a `QString`. The `QString(...)` is superfluous.

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -111,7 +112,7 @@ MainTab::MainTab(QWidget *parent) : QTabWidget(parent),
 	// This needs to be the same order as enum dive_comp_type in dive.h!
 	QStringList types = QStringList();
 	for (int i = 0; i < NUM_DIVEMODE; i++)
-		types.append(QString(tr(divemode_text_ui[i])));
+		types.append(QString(gettextFromC::instance()->tr(divemode_text_ui[i])));
```
bstoeger: Here too: No `QString` needed.

sfuchs79:
@bstoeger 
Good question and to be honest my answer is: No, can't exactly explain it.
Simply copied some other code here plus I remember that someone else (Dirk?) already once explained to me why there are these many different ways of calling "tr".
I now double checked and there are other occurrences of `gettextFromC::instance()->tr();` but no occurrence of `gettextFromC::tr()` in our code.
Maybe we need Dirk to explain this.

profile-widget/profilewidget2.cpp
```diff
@@ -1622,7 +1623,7 @@ void ProfileWidget2::addDivemodeSwitch()
 	QAction *action = qobject_cast<QAction *>(sender());
 	QPointF scenePos = mapToScene(mapFromGlobal(action->data().toPoint()));
 	for (i = 0; i < UNDEF_COMP_TYPE; i++)
-		if (QString(tr(divemode_text_ui[i])) == action->text())
+		if (QString(gettextFromC::instance()->tr(divemode_text_ui[i])) == action->text())
```
bstoeger: Again: no `QString` needed.

desktop-widgets/tab-widgets/maintab.cpp
janiversen: is QString is really needed, that is something that calls for a comment in the code, otherwise the next maintainer will remove it.

btw. why is needed ?

profile-widget/profilewidget2.cpp
janiversen: same comment as above.

qt-models/diveplannermodel.cpp
```diff
@@ -252,7 +253,7 @@ QVariant DivePlannerPointsModel::data(const QModelIndex &index, int role) const
 			else
 				return p.time / 60;
 		case DIVEMODE:
-			return QString(tr(divemode_text_ui[p.divemode]));
+			return QString(gettextFromC::instance()->tr(divemode_text_ui[p.divemode]));
```
bstoeger: Neither here.

qt-models/models.cpp
```diff
@@ -83,7 +84,7 @@ void DiveTypeSelectionModel::repopulate()
 {
 	QStringList modes = QStringList();
 	for (int i = 0; i < FREEDIVE; i++)
-		modes.append(QString(tr(divemode_text_ui[i])));
+		modes.append(QString(gettextFromC::instance()->tr(divemode_text_ui[i])));
```
bstoeger: Nor here...

bstoeger:
> I now double checked and there are other occurrences of gettextFromC::instance()->tr(); but no occurrence of gettextFromC::tr() in our code.

Could you try if the latter works? Because it seems to me that is more correct. If it does, then we should change the other instances.

> Maybe we need Dirk to explain this.

I think I kind of understand. This construct:
```
QT_TRANSLATE_NOOP("gettextFromC", "Open circuit")
```
associates the "gettextFromC" context, which will be used if you call `tr(...)` from the corresponding class.

sfuchs79:
Hmmm... did the test of changing to `gettextFromC::tr()` and yes, it compiles and yes it seems to work.
I'll put it into a seperate commit and push it and then the others can check this as well.

There are also some instances of `gettextFromC::instance()->trGettext()` which I did NOT touch yet...

bstoeger:
`gettextFromC::instance()->trGettext()` needs an instance, because it uses an translation-array. But actually this is only needed for calls from C. So I think it can also be changed to `gettextFromC::tr(...)`. But yes, let's maybe do all this in a different PR.

dirkhh:
@bstoeger you should now be able to restart builds with Travis issues causing failures.
Is this now considered the final version, or are you guys still working on this PR?

bstoeger:
> @bstoeger you should now be able to restart builds with Travis issues causing failures.

Cool - good to know.

> Is this now considered the final version, or are you guys still working on this PR?

lgtm - but I haven't tested it.

bstoeger (approved):

bstoeger:
@dirkhh: I think your recent changes confused Travis:
```
CMake Error at /usr/local/cmake-3.9.2/share/cmake-3.9/Modules/CMakeDetermineCCompiler.cmake:48 (message):
  Could not find compiler set in environment variable CC:
  gcc-5.
Call Stack (most recent call first):
  CMakeLists.txt:14 (PROJECT)
```

sfuchs79:
From my side this is ready and yes, I tested it with good results.

bstoeger:
Ok - after reading a bit into the translation thing, this PR looks "obviously correct".

PR #1392, opened 2018-06-17
build-system: use common code to download tarballs

mfbernardes:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/scripts/get-dep-lib.sh
+++ b/scripts/get-dep-lib.sh
@@ -51,6 +51,28 @@ git_checkout_library() {
 	popd
 }
 
+# Download and extract tarball dependencies
+#
+# Arguments:
+#	name - used as output directory
+#	base_url -
+#	filename - tarball file name
+#
+curl_download_library() {
+	[ $# -ne 3 ] && return -1
+
+	local name=$1
+	local base_url=$2
+	local filename=$3
+
+	if [ ! -f "$filename" ]; then
+		${CURL} "${base_url}${filename}"
+		rm -rf "$name"
+		mkdir "$name"
+		tar -C "$name" --strip-components=1 -xf "$filename"
+	fi
+}
+
 
 # deal with all the command line arguments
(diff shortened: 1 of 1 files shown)


mfbernardes:
Open for comments. I think this is generic enough, but I might have missed something.

mfbernardes:
@dirkhh Sure. Just pushed an update.

janiversen:
@dirkhh see the comment on #1390, checking for the existence of tar file is less error prone, than having another file. I have documented below what I mean, so I will close my PR.

@mfbernardes I do it like this:
+if [[ "$BUILD" = *"libzip"* ]]; then
 +	if [[ ! -e libzip-${CURRENT_LIBZIP}.tar.xz || ! -d libzip ]]; then
 +		${CURL} https://subsurface-divelog.org/downloads/libzip-${CURRENT_LIBZIP}.tar.xz
 +		tar xJf libzip-${CURRENT_LIBZIP}.tar.xz
 +		mv libzip-${CURRENT_LIBZIP} libzip
 +	fi

mfbernardes:
@janiversen This patch does check the tar file. If present it is not downloaded again.

I also chose to check if the final directory is from a known and expected version. If not, cleanup and unpack the known version.

Of course the first time it is run on an existing download it will trigger a new unpacking.

I have no love for this check. If just checking the presence of the tar file is prefered I can do just that, no problem.
@dirkhh comments?

janiversen:
Sorry, I saw that what I meant (but did not write clearly) was exactly what you are asking @dirkh to comment :-) This would be the optimal solution for me, but your patch is fine as it is!

dirkhh:
I think just testing for presence of the tar file is sufficient

mfbernardes:
Updated to check just the tar file and Travis finished ok.

PR #1395, opened 2018-06-17
libdc interface: remove debug messages

bstoeger:
Don't spill supported transports as error message to the user.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Remove debug messages which might be confusing to the user. See ML discussion.

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -1212,19 +1212,6 @@ char *transport_string[] = {
 	"BLE"
 };
 
-static char *transport_to_string(int t)
-{
-	static char buf[1024];
-	buf[0] = '\0';
-	for (int i = 0; i < 6; i++) {
-		if (t & (1<<i)) {
-			strncat(buf, transport_string[i], 1024);
-			strncat(buf, " ", 1024);
-		}
-	}
-	return buf;
-}
-
 /*
  * Get the transports supported by us (as opposed to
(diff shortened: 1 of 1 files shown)


janiversen:
I am wondering: doesn´t it make sense to keep it, but "silence" it with an #ifdef DEBUG ?

bstoeger:
That would be OK for me, but then the `strncat`, which the compiler correctly warns about (even if it can't produce a buffer overflow), should be fixed.

janiversen:
hmm I thought strncat was invented to secure against buffer overflow, but yes less warnings is something good (note I did not see the compiler warning)

bstoeger:
But only if you use it correctly (the last parameter is the maximum number of bytes appended to dest, *not* the maximum buffer size).

To avoid buffer overflows, we have `membuffer`. That would also avoid the static buffer. Not that dive.c would be anything close to reentrant. Sigh.

janiversen:
you got me there :-) I remembered it as buffer size (like most of the other n functions).

sure dive.c is not reentrant, and in general I would like to slowly move some of the c files into simple classes, because I strongly believe it will make ssrf more fault tolerant. I am however aware that this requires some foot work in convincing people.

PR #1397, opened 2018-06-17
mobile: revert e993d4f005e62c12765e2e45342ecf840476e3bf

janiversen:
The commit secured that plotDive was not called before actually being used.
However our (rather fragile) C++ qml interface did not work correctly (ony sometimes).

Revert the previous commit.

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

revert commit e993d4f005e62c12765e2e45342ecf840476e3bf
Due to a problem in our C++ qml interface this does not work as expected.

qmlprofilewidget.cpp and .h needs to be updated, but that will be after the release.
In general it seems our qml interface classes should be implemented using the
interface/implementation design pattern to avoid spurious construct/deconstruct.

currently using the onCompleted signal, which is (as the name says) too late, but works
since the page isn't visible. There are no onLoad signal which would have been logical,
this need instead to be transferred to the qmlprofilewidget constructor, and hence the need
for a interface/implementation pattern.

### Changes made:

1) revert e993d4f005e62c12765e2e45342ecf840476e3bf

### DIFF
--- a/mobile-widgets/qml/DiveDetailsView.qml
+++ b/mobile-widgets/qml/DiveDetailsView.qml
@@ -217,13 +217,6 @@ Item {
 				border.color: subsurfaceTheme.primaryColor
 				anchors.fill: parent
 			}
-
-			onVisibleChanged: {
-				if (visible) {
-					qmlProfile.diveId = model.dive.id;
-					qmlProfile.update();
-				}
-			}
 		}
 		Controls.Label {
 			id: noProfile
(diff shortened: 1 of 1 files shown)


janiversen:
@mfbernardes could you please confirm that is works now. Thanks in advance.

mfbernardes:
This works fine for me.

I understand the idea of the original change, but in my observations visibility/onLoad of a flickable can be a bit trickier than might seem at first.

janiversen:
yes, the definition of "onVisible" is something that can be interpreted. Sadly enough there are no "onLoad" event which would have been the best option, that has to be done in the C++ constructor.

PR #1398, opened 2018-06-17
mobile: move call to savePreferences back to manager.

janiversen:
calls to savePreferences was moved to prefs. in
b8eb348f546291c49e2d606649b3c85d23e43782, but the corresponding
C++ code was not merged.

Revert call to savePreferences to manager.

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Due to a mismatch of C++ and qml code in b8eb348f546291c49e2d606649b3c85d23e43782
savePreferences was wrongly seen as part of prefs.

Correct qml call prefs. -> manager.savePreferences

sorry for the inconvenience.

### DIFF
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -138,7 +138,7 @@ Kirigami.ScrollablePage {
 				onClicked: {
 					blueTheme()
 					prefs.theme = subsurfaceTheme.currentTheme
-					prefs.savePreferences()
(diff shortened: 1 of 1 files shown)


dirkhh:
This looks correct.
We should maybe slow down some of these more mechanical changes before the release, just to avoid having more of such inconsistencies slip through

PR #1400, opened 2018-06-17
[RFC] Localization: simplify gettextfromc functions

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Admittedly, this is a bit of a brutal cleanup. But at the end of the day, there seems to be no reason for having an instance of the `gettextFromC` class. C++ callers can simply use the `static` `tr()` function. C callers (or C++ callers wanting a persistent C-string) use the function `trGettext()`, which is the only user of the translation-cache. Thus, the translation cache can be made local to this function.

PS: This obviously is not (and never was) thread safe! I'm not positive that this is only called from GUI-thread context.

### Changes made:

1) Only use C++-translation functions from C++.
2) Remove the `gettextFromC` instance.

### Release note:

None needed.

### Documentation change:

None needed.

### DIFF
--- a/CodingStyle.md
+++ b/CodingStyle.md
@@ -135,7 +135,7 @@ other editors that implement this coding style, please add them here.
 * text strings
   The default language of subsurface is US English so please use US English
   spelling and terminology.
-  Where at all possible strings should be passed to the tr() function to enable
+  User-visible strings should be passed to the tr() function to enable
   translation into other languages.
 
  * like this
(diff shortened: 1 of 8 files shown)


bstoeger:
PPS: This code has (and always had) a subtle use-after-free problem for translations that evaluate to the empty string:
```
	if (result.isEmpty())
		result = gettextFromC::tr(text).toUtf8();
 ```

Suppose `gettextFromC::tr(text).toUtf8()` gives an empty array. A second call will overwrite and thus invalidate the data of the QByteArray.

Proof:
```
printf("%p %p\n", trGettext(""), trGettext(""));
```
gives:
```
0x559b411bba08 0x559b411ae388
```

Now, there is probably no translation evaluating to the empty string. And the worst thing that can happen is garbage output where there should be an empty string. Still not a nice situation.

**Comment by bstoeger** [comment]
Pushed a fix of both issues.

**Comment by dirkhh** [comment]
This looks really reasonable to me. I'm a little worried about unintended side effects, but yeah, let's do this (once Travis has given us a verdict)

**Comment by bstoeger** [comment]
I went a step further and replaced `translate("gettextFromC", ...)` by `gettextFromC::tr(...)` in `qthelper.cpp`. To my understanding the translation utility parses both, right?

> I'm a little worried about unintended side effects

This patch shouldn't change the needed translations. Is there a way to check this?

**Comment by janiversen** [comment]
Would it not be better to delay merging of this patch until after the release. We could also consider use release branches that cherry-pick from master.

**Comment by dirkhh** [comment]
I can check the impact on translations locally. I think @janiversen has a point, though... the again, no risk no fun :-)
I'm about to have breakfast and head out for a day of diving. I'll look into it this evening.

**Comment by bstoeger** [comment]
If it doesn't impact translations and survives some basic testing, then I think this is not really high risk. Then again, it's not a pressing issue either.

**Comment by bstoeger** [comment]
Added a few lines to `CodingStyle.md` detailing how to use translations from C and C++. Please check.

File: CodingStyle.md [hunk 1]
```diff
@@ -135,7 +135,7 @@ other editors that implement this coding style, please add them here.
 * text strings
   The default language of subsurface is US English so please use US English
   spelling and terminology.
-  Where at all possible strings should be passed to the tr() function to enable
+  User-visible strings should be be passed to the tr() function to enable
```
dirkhh: one 'be' wasn't enough? :-)

dirkhh (commented):
thanks for adding the documentation!!!

tcanabrava:
No break no fun, merged.

dirkhh:
Umm, @tcanabrava - I'm glad you are back. I wish you hadn't merged this.
In my preliminary testing this did cause some changes, so I had planned to wait until after the release this week...

dirkhh:
It doesn't seem to break anything, but it causes strings to not be recognized as their context changes. 
Reverting this might be the best solution. But I'll see if I can somehow fix this with a bit of scripting on my side and without causing extra work for the translators. That's why I wanted to merge this after the release...

dirkhh:
OK, I worked around the problems here - and then realized that @atdotde had added new strings in his PR. So we need additional translations anyway...

bstoeger:
> It doesn't seem to break anything, but it causes strings to not be recognized as their context changes.

Damn - that was unintended. For my understanding: What were these strings?

dirkhh:
I couldn't make sense of it - it was names of import types. UDDF, etc. I think Transifex managed to sort this all out for us by poking it to accept 100% matches.
With the this ended up not having any negative side effects I was able to pinpoint.
The new strings that I asked people to translate weren't caused by your PR :-)

Issue #1408, opened 2018-06-19
Subsurface-mobile beta crash


Ebi129:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
When a dive is manually added it is not possible to change depth or duration again. The app crashes when the save button is pressed.

### Operating system:
Android 7.0

### Subsurface version:
mobile 2.1.0(4.7.8.372) beta, official release 2.0.1(4.7.4.280) is o.k.

janiversen:
problem is in
void fake_dc(struct divecomputer *dc)

	memset(fake, 0, 6 * sizeof(struct sample));

Bad pointer

dirkhh:
should be fixed by #1411

PR #1411, opened 2018-06-19
core: bug in alloc_samples ?

janiversen:
When num > dc->alloc_samples we whould allocate space for num

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

alloc_samples does not use the parameter num, and dc->alloc_samples start with 0,
so change to realloc using num (if > dc->alloc_samples). I saw this problem while debugging #1408

Please this code is completely new land, so please check if my assumptions are correct.

The second commit for qmlmanager.cpp corrects the problem in #1408, but I am not sure it
is the correct way.

solves #1408

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -747,7 +747,7 @@ void copy_samples(struct divecomputer *s, struct divecomputer *d)
 void alloc_samples(struct divecomputer *dc, int num)
 {
 	if (num > dc->alloc_samples) {
-		dc->alloc_samples = (dc->alloc_samples * 3) / 2 + 10;
+		dc->alloc_samples = (num * 3) / 2 + 10;
 		dc->sample = realloc(dc->sample, dc->alloc_samples * sizeof(struct sample));
 		if (!dc->sample)
 			dc->samples = dc->alloc_samples = 0;
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -927,6 +927,7 @@ bool QMLManager::checkDuration(DiveObjectHelper *myDive, struct dive *d, QString
 			free(d->dc.sample);
 			d->dc.sample = 0;
 			d->dc.samples = 0;
+			d->dc.alloc_samples = 0;
 		} else {
 			appendTextToLog("Cannot change the duration on a dive that wasn't manually added");
 		}
@@ -949,6 +950,7 @@ bool QMLManager::checkDepth(DiveObjectHelper *myDive, dive *d, QString depth)
 				free(d->dc.sample);
 				d->dc.sample = 0;
 				d->dc.samples = 0;
+				d->dc.alloc_samples = 0;
 			}
 			return true;
 		}


janiversen:
not enough, but still an error in my opinion

bstoeger (commented):
lgtm - only a minor comment.

mobile-widgets/qmlmanager.cpp
```diff
@@ -927,6 +927,7 @@ bool QMLManager::checkDuration(DiveObjectHelper *myDive, struct dive *d, QString
 			free(d->dc.sample);
 			d->dc.sample = 0;
 			d->dc.samples = 0;
+			d->dc.alloc_samples = 0;
```
bstoeger, line 930: It is kind of sad that the UI-code directly manipulates internal data structures.

Therefore, in the short-run, I would suggest creating a `free_samples()` function, which does the above and is called here and in `planner.c`. But I think this is beyond the scope of this PR.

In the long run, I think we should reimplement `std::vector<>` in C [my personal choice would be simply using `std::vector<>` in the backend (while still providing a C-only interface). But no worries - I'm not going to try to convince anybody.]

janiversen: I agree that making  a free_samples function is the minimum we should do, but I wanted to keep this PR as small as possible (as you also write out of scope). If/when this PR is merged I am happy to make the function.

I agree with you on the vector considerations.

core/dive.c
```diff
@@ -747,7 +747,7 @@ void copy_samples(struct divecomputer *s, struct divecomputer *d)
 void alloc_samples(struct divecomputer *dc, int num)
 {
 	if (num > dc->alloc_samples) {
-		dc->alloc_samples = (dc->alloc_samples * 3) / 2 + 10;
+		dc->alloc_samples = (num * 3) / 2 + 10;
```
dirkhh, line 750: So this allocates 50% + 10 additional samples. Which of course works fine if your pattern everywhere in the code is to either simply allocate "one more" sample (so this gets called with 'nr+1'), or up to six samples in one block.
That is of course bleeping fragile and needs to be fixed - but given our call pattern, I wonder how you got into trouble here?

mobile-widgets/qmlmanager.cpp
dirkhh: C++ code compiles slower and runs usually an order of magnitude slower. Which is why we generally don't jump on that bandwagon for things that can just as easily done in C.

dirkhh (commented):
As I said in my comment on the other PR: please use
```
Fixes #1410
```
in the commit message which fixes that issue

core/dive.c
janiversen: just for the record, this did not bring trouble, but I saw it while debugging, and thought it looked wrong. I did not adjust the algorithm as such, the structure is not very big so overallocating is not a big issue.

PR #1413, opened 2018-06-20
Qt BLE latency improvements

torvalds:
This changes how we wait for incoming packets, to improve latency.

It seems to make an enormous difference to download reliability for me - I'm not entirely sure why, but I started doing this because my packet receive traces showed behavior that implied that we would apparently sometimes miss incoming packets simply because the incoming packets would be processed much too late.

It is *possible* that this makes a difference to other dive computers too (ie the occasional timeout  issue with the Shearwater Perdix AI, for example), but that may be wishful thinking.

This also flushes the old read queue in the (now unlikely) situation where we didn't process the reads in time, decided to time out, and then write a new command. We used to warn about it - and we still do - but now we also clean up the stale old receive queue.

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -155,6 +155,9 @@ dc_status_t BLEObject::write(const void *data, size_t size, size_t *actual)
 
 	if (!receivedPackets.isEmpty()) {
 		qDebug() << ".. write HIT with still incoming packets in queue";
+		do {
+			receivedPackets.takeFirst();
+		} while (!receivedPackets.isEmpty());
 	}
 
 	QList<QLowEnergyCharacteristic> list = preferredService()->characteristics();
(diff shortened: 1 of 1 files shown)


dirkhh:
I agree that this isn't the prettiest macro I've ever seen, but hey, if it works...
Let's be patient and wait for (ha!) Travis.

dirkhh:
Looking at the Travis failures, it looks like you are missing a `#include <QTime>` or something

Somehow I had assumed that you had compile tested this :-)

torvalds:
Oh damn.

I had other debugging-related patches for packet tracing in my tree that added the QTime header for independent reasons. But I had intentionally rebased them to be later because they were ugly as all hell. But I tested *with* the packet tracing, obviously.

torvalds:
Oh, I had on stupid  debug line mixed in with the qt-ble code that I hadn't separated out right.

PR #1414, opened 2018-06-20
Fixed Qt BLE latency

torvalds:
Another try at fixing the Qt BLE latency. This time with the debugging code properly removed.

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -29,17 +29,26 @@ static int debugCounter;
 #define MAXIMAL_HW_CREDIT	255
 #define MINIMAL_HW_CREDIT	32
 
-static void waitFor(int ms) {
-	Q_ASSERT(QCoreApplication::instance());
-	Q_ASSERT(QThread::currentThread());
+#define WAITFOR(expression, ms) do {					\
+	Q_ASSERT(QCoreApplication::instance());				\
+	Q_ASSERT(QThread::currentThread());				\
+									\
+	if (expression)							\
+		break;							\
+	QElapsedTimer timer;						\
+	timer.start();							\
+									\
+	do {								\
+		QCoreApplication::processEvents(QEventLoop::AllEvents, ms); \
+		if (expression)						\
+			break;						\
+		QThread::msleep(10);					\
+	} while (timer.elapsed() < (ms));				\
+} while (0)
 
-	QElapsedTimer timer;
-	timer.start();
-
-	do {
-		QCoreApplication::processEvents(QEventLoop::AllEvents, ms);
-		QThread::msleep(10);
-	} while (timer.elapsed() < ms);
+static void waitFor(int ms)
+{
+	WAITFOR(false, ms);
 }
 
 extern "C" {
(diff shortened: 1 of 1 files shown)


torvalds:
This changes how we wait for incoming packets, to improve latency.

It seems to make an enormous difference to download reliability for me - I'm not entirely sure why, but I started doing this because my packet receive traces showed behavior that implied that we would apparently sometimes miss incoming packets simply because the incoming packets would be processed much too late.

It is possible that this makes a difference to other dive computers too (ie the occasional timeout issue with the Shearwater Perdix AI, for example), but that may be wishful thinking.

This also flushes the old read queue in the (now unlikely) situation where we didn't process the reads in time, decided to time out, and then write a new command. We used to warn about it - and we still do - but now we also clean up the stale old receive queue.

core/qt-ble.cpp
```diff
@@ -146,6 +155,9 @@ dc_status_t BLEObject::write(const void *data, size_t size, size_t *actual)
 
 	if (!receivedPackets.isEmpty()) {
 		qDebug() << ".. write HIT with still incoming packets in queue";
+		do {
+			receivedPackets.takeFirst();
+		} while (!receivedPackets.isEmpty());
```
bstoeger, line 160: Am I missing something? Why not `receivedPackets.clear()`?

torvalds: Because I have never used QLists and we already used takeFirst?

bstoeger: Fair enough, could have been that I missed something subtle. Note that `QList` is *not* a linked list as the name would imply, but a "deque" based on a list of arrays of pointers. Qt's data structures are a mess.

torvalds: I basically do Qt (and C++ in general) programming by pattern matching. I  do try to look up the online docs, but QList had a ton of different functions, and the sanely named one (removeAll) did something entirely different, and I never noticed "clear()".

Oh well.

core/qt-ble.cpp
```diff
@@ -29,17 +29,26 @@ static int debugCounter;
...
-	Q_ASSERT(QThread::currentThread());
-
-	QElapsedTimer timer;
-	timer.start();
-
-	do {
-		QCoreApplication::processEvents(QEventLoop::AllEvents, ms);
-		QThread::msleep(10);
-	} while (timer.elapsed() < ms);
+#define WAITFOR(expression, ms) do {					\
```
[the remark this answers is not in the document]
torvalds: Yeah, a lambda expression is certainly _conceptually_ the right thing to do.

Maybe C++ compilers even do them right these days. Considering that C++ got the *simple* stuff completely wrong (no static inline helper functions for member functions? You have to declare them as private member functons? Christ) I wouldn't actually have expected a lambda function to be compiled right.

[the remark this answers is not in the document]
bstoeger: In my experience, lambdas are happily optimized away. In any case, the macro does its job.

dirkhh:
@bstoeger feel free to send a PR to make this cleaner / better. I wanted to quickly create a new Android beta, so I pulled this... and now I've been fighting with the Android store for 30 minutes... grmbl

PR #1418, opened 2018-06-20
Dive pictures: fix loading of remote images without local version

bstoeger:
Owing to the recent churn in imagedownloader.cpp, some of the
code was bogus.

Notably, in f60343eebbf6a31a4643dde9f4454f6ce84f61d3 the code
was changed such that always the local filename was used to access
the images. Yet, the old code remained, which after failure tried
again to access the local picture. This second access can obviously
be removed completely.

More seriously, after failing to load the local version, no
attempt was made to fetch the canonical filename. This could
produce the following sequence of events:
  - Import remote image
  - Delete thumbnail and local cache of image
  - Image loading would fail

Therefore, first try to load using local file-location. If
that fails, load using the canonical file-location. To do
so, split the file-access code in two functions. The code
should now be distinctly easier to follow.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I looked at the image-fetcher code and it is still bogus. Sigh. :( So here is another cleanup, hopefully this is on the way to become stable now. At least I think that the code is becoming distinctly easier to follow now.

This actually fixes a bug: Only the "local" file would be checked. Thus, if one imported an image from the web, then erased the cached image and the thumbnail, the image would not be reloaded from the web. I could reproduce this failure-mode and this commit indeed fixes the problem. Even though this is a somewhat obscure use-case, I guess this means that it should be applied before release.

Question: In the case of a non-remote filepath, do we also want to check the canonical filename if the local filename fails? Seems more robust, but might increase the number of disk accesses?

### Changes made:

1) Remove double checking of local filename.
2) Split image-fetch function in two (fetch by canonical filename / fetch by actual filename).
3) remove one-line `loadImage()` stub.

### Release note:

Not really needed. I believe the bug was introduced by myself during current release cycle.
### Documentation change:

None needed.

### DIFF
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -65,45 +65,57 @@ void ImageDownloader::saveImage(QNetworkReply *reply)
 	reply->deleteLater();
 }
 
-static void loadPicture(QUrl url, QString filename)
+// Fetch a picture from the given filename. If this is a non-remote filename, fetch it from disk.
+// Remote files are fetched from the net in a background thread. In such a case, the output-flag
+// "stillLoading" is set to true.
+// If the input-flag "tryDownload" is set to false, no download attempt is made. This is to
+// prevent infinite loops, where failed image downloads would be repeated ad infinitum.
+// Returns: fetched image, stillLoading flag
+static std::pair<QImage, bool> fetchImage(const QString &filename, const QString &originalFilename, bool tryDownload)
 {
-	// This has to be done in UI main thread, because QNetworkManager refuses
-	// to treat requests from other threads.
-	QMetaObject::invokeMethod(ImageDownloader::instance(), "load", Qt::AutoConnection, Q_ARG(QUrl, url), Q_ARG(QString, filename));
-}
-
-// Returns: thumbnail, still loading
-static std::pair<QImage,bool> getHashedImage(const QString &file_in, bool tryDownload)
-{
-	QString file = file_in.startsWith("file://", Qt::CaseInsensitive) ? file_in.mid(7) : file_in;
 	QImage thumb;
 	bool stillLoading = false;
-	QUrl url = QUrl::fromUserInput(localFilePath(file));
-	if (url.isLocalFile())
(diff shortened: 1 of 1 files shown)


core/imagedownloader.cpp
```diff
@@ -65,45 +65,57 @@ void ImageDownloader::saveImage(QNetworkReply *reply)
...
 }
 
-static void loadPicture(QUrl url, QString filename)
+// Fetch a picture from the given filename. If this is a non-remote filename, fetch it from disk.
+// Remote files are fetched from the net in a background thread. In such a case, the output-flag
+// "stillLoading" is set to true.
+// If the input-flag "tryDownload" is set to false, no download attempt is made. This is to
+// prevent infinite loops, where failed image downloads would be repeated ad infinitum.
+// Returns: fetched image, stillLoading flag
+static std::pair<QImage,bool> fetchImage(const QString &filename, const QString &originalFilename, bool tryDownload)
```
neolit123: a small nit would be to always add spaces between the template types:
`<QImage, bool>`

i'm ok with pair, tuple, tie and so on, and while these are new and cool (although, i just found a blog post that explains how these "smell"), i wonder how the rest of our developers feel about these.

bstoeger: > a small nit would be to always add spaces between the template types:

done.

> i'm ok with pair, tuple, tie and so on, and while these are new and cool (although, i just found a blog post that explains how these "smell"), i wonder how the rest of our developers feel about these.

I think that once it's more complex than a pair, a struct is probably more appropriate. The one occcasion where `tie` is really nice is for comparisons. Note that it's all in C++11, so not all that new. I can replace by a struct, but would prefer doing this with the video branch, when all this is more settled. Notably, in the video branch the `stillLoading` bool is replaced by a `mediatype_t` which can adopt quite a number of different states (picture, video, io-error, still-loading, unknown).

PR #1425, opened 2018-06-25
QML UI: correctly update the model

### LINKED ISSUE #1419: Subsurface-mobile beta: dive plot empty
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When a dive is manually added and saved the dive profile is not shown, only an empty graph is shown in the details view of the dive. By swiping two dives to the left and then back to the right the graph is now shown correctly. If values are changed the graph is not redrawn, only after swiping  two dives back and forth again.

### Operating system:

Android 7.0 german

### Subsurface version:

Subsurface-mobile beta 2.1.0(4.7.8.389)
### Current behavior:

grafik

dirkhh:
In order to trigger the redraw of an edited dive we need to make sure
the model realizes that it has been updated. So far the only way to make
sure this happens reliably appears to be to remove the item and
re-insert it. Seems weird, but with this the bug of not redrawing the
profile after an edit appears fixed.

Fixes #1419

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This has been bugging me for a while - trying the big hammer worked.
The trick is to make sure that the model realizes that the dive has changed. The only thing that seems to reliably work is to remove the dive and immediately re-add it.
### Related issues:

Fixes #1419
### Release note:

I have locally already moved the CHANGELOG to the ReleaseNotes (but haven't pushed that yet), so while this needs to go into the ReleaseNotes I didn't add that here as it will just create a merge conflict later
### Mentions:

@janiversen @mfbernardes @tcanabrava - I'd love a sanity check from people who have played with QML :-)

### DIFF
--- a/qt-models/divelistmodel.cpp
+++ b/qt-models/divelistmodel.cpp
@@ -94,8 +94,10 @@ void DiveListModel::removeDiveById(int id)
 void DiveListModel::updateDive(int i, dive *d)
 {
 	DiveObjectHelper *newDive = new DiveObjectHelper(d);
-	m_dives.replace(i, newDive);
-	emit dataChanged(createIndex(i, 0), createIndex(i, 0));
+	// we need to make sure that QML knows that this dive has changed -
+	// the only reliable way I've found is to remove and re-insert it
+	removeDive(i);
+	insertDive(i, newDive);
 }
 
 void DiveListModel::clear()


tcanabrava:
That's a Hammer, I don't know why the dataChanged failed, maybe because it's sending a dataChanged for the first column and the models in Qml are a bit weird. if the hammer approach works, I'm +1.

dirkhh:
It's the only way (and I tried quite a few) to make it draw a profile when you manually create a dive or update the profile when you edit such a manually added dive. Yes, it seems brutal, but... at least it works? And it fixes a very annoying user visible bug.

dirkhh:
I have now tested this
- on Linux / Desktop
- on Android (actual device)
- on iOS / simulator (didn't bring an iOS device on this trip)
I know I'm being ridiculously impatient, but I'll merge this - I simply want to create new test binaries for people to play with before calling it a night... and I'll spend most of tomorrow on a flight from China to the US...

janiversen:
tested ok on iOS 11.3

dirkhh:
Thanks @janiversen

PR #1426, opened 2018-06-25
QML UI: don't use anchors within Layouts

dirkhh:
Qt 5.11 adds useful warnings when code attempts to use anchors within
Layouts and even tells you how to fix things.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

It seems weird to add this so late in the cycle - but I finally had enough bandwidth to update my development machine to Qt5.11
This does seem to fix an actual bug that was just silently ignored before.

### Release note:

not user visible
### Mentions:

@janiversen @mfbernardes @tcanabrava - I'd love a sanity check from people who have played with QML :-)

### DIFF
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -97,9 +97,8 @@ Item {
 		}
 
 		RowLayout {
-			anchors.left: parent.left
-			anchors.right: parent.right
-			anchors.margins: Kirigami.Units.smallSpacing
+			Layout.fillWidth: true
+			Layout.margins: Kirigami.Units.smallSpacing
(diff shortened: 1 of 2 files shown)


tcanabrava:
Seems correct. +1.

mfbernardes:
lgtm

dirkhh:
Thanks - I'll merge this one as well. It did well in all of my testing (but I haven't been able to test on an iOS device). Of course all of our device builds are NOT against Qt 5.11.1, yet, so this may not make a difference (I'm unclear if earlier versions allowed those anchors or if they just silently ignored them)

PR #1432, opened 2018-06-26
Update pictures in profile plot on deletion / drag & drop

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This introduces two usability changes to the picture thumbnails in the profile plot. This needed some code-reshuffling. It introduces a helper function in `qthelper.h`, which is quite useful: Move a range in a vector or vector-like container. I plan to use this more often. If there is a better place, let me know.
 
### Changes made:

1) When deleting a picture, update the thumbnails immediately, not only on changing the dive. But don't recalculate everything, only adjust the coordinates of the remaining pictures.

2) When drag&dropping a picture, don't recalculate the whole thumbnail list. Simply rearrange the thumbnails accordingly. To enable this, the pictures are now sorted by timestamp and the list is kept in sorted order. This is a bit tricky, since we have a list in the C-backend, a list in `DivePictureModel` and a list in `ProfileWidget2`. This PR does not touch the C-backend at all.

### Release note:

Needed? Perhaps something like
- Improve deletion of images / adjusting timestamp of images.

?

### Documentation change:

Not needed.

### DIFF
--- a/core/qthelper.h
+++ b/core/qthelper.h
@@ -93,6 +93,42 @@ QString getUserAgent();
 #define TITLE_OR_TEXT(_t, _m) _t, _m
 #endif
 
+// This helper function moves a range in a vector to a different position.
+// The parameters are given according to the usual STL-semantics:
+//	v: a container with STL-like random access iterator via std::begin(...)
+//	rangeBegin: index of first element
+//	rangeEnd: index one *past* last element
+//	destination: index to element before which the range will be moved
+// The function returns the new index of the moved range.
+// Owing to std::begin() magic, this function works with STL-like containers:
+//	QVector<int> v{ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+//	moveInVector(v, 1, 4, 6);
+// as well as with C-style arrays:
+//	int array[10] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+//	moveInVector(array, 1, 4, 6);
+// Both calls will have the following effect:
+//	Before: 0 1 2 3 4 5 6 7 8 9
+//	After:  0 4 5 1 2 3 6 7 8 9
+// No sanitizing of the input arguments is performed.
+template <typename Vector>
+int moveInVector(Vector &v, int rangeBegin, int rangeEnd, int destination)
+{
+	auto it = std::begin(v);
+	if (destination > rangeEnd) {
+		// In C++11, std::rotate() returns an iterator to the element that was at
+		// the begining of the range before the rotation. We can't use that, since
+		// Ubuntu 14.04 ships with a buggy libstdc++, where std::rotate returns void.
+		// return std::rotate(it + rangeBegin, it + rangeEnd, it + destination) - it;
+		std::rotate(it + rangeBegin, it + rangeEnd, it + destination);
+		return rangeBegin + destination - rangeEnd;
+	} else if (destination < rangeBegin) {
+		std::rotate(it + destination, it + rangeBegin, it + rangeEnd);
+		return destination;
+	} else {
+		return rangeBegin;
+	}
+}
+
 #endif
 
 // 3) Functions visible to C and C++
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -2057,19 +2057,40 @@ void ProfileWidget2::updatePictures(const QModelIndex &from, const QModelIndex &
 	}
 }
 
+void ProfileWidget2::calculatePictureYPositions()
+{
+	double lastX = -1.0, lastY;
+	for (std::unique_ptr<DivePictureItem> &pic: pictures) {
+		if (!pic)
+			continue;
+		// let's put the picture at the correct time, but at a fixed "depth" on the profile
+		// not sure this is ideal, but it seems to look right.
+		double x = pic->x();
+		double y;
+		if (lastX <= 0.0)
+			y = 10;
+		else if (fabs(x - lastX) < 3 && lastY <= (10 + 14 * 3))
+			y = lastY + 3;
+		else
+			y = 10;
+		lastX = x;
+		lastY = y;
+		pic->setY(y);
+	}
+}
+
 void ProfileWidget2::plotPictures()
 {
 	DivePictureModel *m = DivePictureModel::instance();
 	pictures.resize(m->rowDDEnd - m->rowDDStart);
 
-	double x, y, lastX = -1.0, lastY = -1.0;
 	for (int i = m->rowDDStart; i < m->rowDDEnd; i++) {
 		int picItemNr = i - m->rowDDStart;
 		int offsetSeconds = m->index(i, 1).data(Qt::UserRole).value<int>();
-		// it's a correct picture, but doesn't have a timestamp: only show on the widget near the
-		// information area. A null pointer in the pictures array indicates that this picture is not
-		// shown.
-		if (!offsetSeconds) {
+		// it's a correct picture, but doesn't have a timestamp during the dive: only show in
+		// the DivePhotoTab, not on the profile. A null pointer in the pictures array indicates
+		// that this picture is not shown in the profile.
+		if (offsetSeconds <= 0) {
 			pictures[picItemNr].reset();
 			continue;
 		}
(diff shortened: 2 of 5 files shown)


bstoeger:
The `linux` Travis build fails with
```
/home/travis/build/Subsurface-divelog/subsurface/profile-widget/profilewidget2.cpp:2158:40:   required from here
/home/travis/build/Subsurface-divelog/subsurface/./core/qthelper.h:118:72: error: no match for ‘operator-’ (operand types are ‘void’ and ‘__gnu_cxx::__normal_iterator<std::unique_ptr<DivePictureItem>*, std::vector<std::unique_ptr<DivePictureItem> > >’)
   return std::rotate(it + rangeBegin, it + rangeEnd, it + destination) - it;
```

Note that `std::rotate(...)` is supposed to return `ForwardIt` since C++11: https://en.cppreference.com/w/cpp/algorithm/rotate

Since we use C++11 in numerous places (makes things *much* nicer compared to C++98), could we perhaps get a compliant build environment?

Certainly, in this case emulating C++98 would only mean a single additional line. Nevertheless, such an archaic build environment seems like a bad idea.

bstoeger:
PS: Here's the gcc defect report:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58357

neolit123:
@bstoeger 

re: release note
> Improve deletion of images / adjusting timestamp of images.

i would add two short release note items based on these two paragraphs. both are user facing.
1.
> When deleting a picture, update the thumbnails immediately, not only on changing the dive. But don't recalculate everything, only adjust the coordinates of the remaining pictures.
2.
> When drag&dropping a picture, don't recalculate the whole thumbnail list. Simply rearrange the thumbnails accordingly. To enable this, the pictures are now sorted by timestamp and the list is kept in sorted order. This is a bit tricky, since we have a list in the C-backend, a list in DivePictureModel and a list in ProfileWidget2. This PR does not touch the C-backend at all.

---- 
> Since we use C++11 in numerous places (makes things much nicer compared to C++98), could we perhaps get a compliant build environment?

what GCC are we using there? by compliant do you mean GCC that actually works? what version would that be? we could, but i'd rather remove the problematic C++11 feature for now.

> Certainly, in this case emulating C++98 would only mean a single additional line. Nevertheless, such an archaic build environment seems like a bad idea.

when it comes to C++ i think like an old man "i will use C++98 everywhere and i know it would work".
c++11 is 7 years old but it seems there are compiler that are still broken for it.

/cc @dirkhh

bstoeger:
> i would add two short release note items based on these two paragraphs. both are user facing.

Perhaps even three:
- Dive pictures: sort images by timestamp
- Dive pictures: immediately update picture positions in profile on deletion
- Dive pictures: don't recalculate all pictures on drag & drop

What do you think?

> what GCC are we using there? by compliant do you mean GCC that actually works? what version would that be? we could, but i'd rather remove the problematic C++11 feature for now.

Come on, that's not a "problematic C++11 feature", that's a blatant failure to implement the trivial library function `std::rotate`. Do we really want to work around all library bugs of the last _n_ years that might appear in some build environment? Note that only the Linux build fails.

Also note that less and less of our external dependencies will support old broken compilers. Example:

https://github.com/Subsurface-divelog/subsurface/commit/fca06fd865688012403dc2013689bdd9c784d342

I understand that we have to support old Qt, but old build environments with who-knows what kind of bugs? I imagine that code creation must be really terrible.

> c++11 is 7 years old but it seems there are compiler that are still broken for it.

I disagree. Sure, in every complex software you will find bugs concerning corner-cases. But in general, C++11 support is excellent in every close-to-recent major compiler.

dirkhh:
Arguing about C++ standards... right... that's what I want us to do
Ubuntu's actively supported Trusty from 2014 has a compiler and libraries that are broken wrt C++11. We can lament that, we can't really pretend it's not true.
That said, I've already switched to a newer compiler in some of our builds, so I can switch to the newer compiler in all of our Linux Travis builds. 
But that won't fix the problem for people who are on older Linux releases (they exist, they complain to me frequently when things don't build for them). Worse, I wonder if this will get us a potentially less portable AppImage.
If this really is ONE extra line to work around, I'd rather see us work around it, tbh. Mostly because I'm lazy and that is less work for me. But also because it seems silly to add a ton of work just because "C++11".

The release note entries are fine, just add them to CHANGELOG.md

Reminder - please do not merge this before the 4.8 release. I'm sure it's fine, but I'm also sure that we can add this to 4.8.1 which will (based on past experience) be necessary to fix a few bugs within a month.

neolit123:
@dirkhh 
> But that won't fix the problem for people who are on older Linux releases (they exist, they complain to me frequently when things don't build for them)

that's my concern too. someone on an ancient Ubuntu with an ancient compiler will get the `std::rotate` error and file an issue that i have to deal with in the GitHub tracker.

@bstoeger 
> Come on, that's not a "problematic C++11 feature", that's a blatant failure to implement the trivial library function std::rotate.

true, but our users that build from source don't care. :)
they say it's Subsurface fault and the way for us to deal with that is to not expose the potential for an error in the first place.

bstoeger:
> Ubuntu's actively supported Trusty from 2014 has a compiler and libraries that are broken wrt C++11.

"actively supported" should mean that they remove the worst bugs, doesn't it?

Note that the qt55 branch, which is the one that we have to test for old Ubuntu does *not* fail. Does this use a more modern build environment?

> I'm sure it's fine

This is quite complex code, I wouldn't be so sure of that.

dirkhh:
@bstoeger - do we really want a Subsurface pull request turn into a indictment of Ubuntu's support patterns?
As for which build fails: "qt55" is run in a Docker container on Ubuntu 16.04 which uses a newer compiler. The goal there is to test that we can build against Qt 5.5, nothing else. And using a container makes it faster.
"linux" on the other hand is built on Ubuntu 14.04 in order to then create an AppImage that we offer to our users. So THAT is the one that runs on the oldest environment. But with an updated Qt 5.9 in order to have a chance to have working BLE.

The reason that I said "I'm sure it's fine" was to avoid argument about whether or not this should land in 4.8.0

But what I'm really confused by is your approach to what Lubomir and I try to explain about the reality of how we create the AppImage. Are you telling me "my view of the true meaning of C++11 and my believe in the sanctity and integrity of my code is more important than reality"? Because in the end you don't have to deal with getting the AppImage to work, you don't have to deal with the support requests from the Debian stable users, you're not the one dealing with our build infrastructure, etc.

Sure, I can update the compiler - that's the reason why I split out linux2 because Subsurface-mobile already needs a newer compiler. We don't package that as an AppImage, so I compromised and only build the desktop version and the AppImages with the old compiler. But the ripple effects of that really will cost me a few days of work (at least). Are you really trying to tell me that's more important than adding an unnecessary line of code to your patch? Because if you are, then maybe I need to take a break from maintaining Subsurface for a while and re-evaluate what I am spending my time on.

bstoeger:
> As for which build fails: "qt55" is run in a Docker container on Ubuntu 16.04 which uses a newer compiler. The goal there is to test that we can build against Qt 5.5, nothing else. And using a container makes it faster.
"linux" on the other hand is built on Ubuntu 14.04 in order to then create an AppImage that we offer to our users. So THAT is the one that runs on the oldest environment. But with an updated Qt 5.9 in order to have a chance to have working BLE.

Thanks for the explanation.

> But what I'm really confused by is your approach to what Lubomir and I try to explain about the reality of how we create the AppImage. Are you telling me "my view of the true meaning of C++11 and my believe in the sanctity and integrity of my code is more important than reality"? Because in the end you don't have to deal with getting the AppImage to work, you don't have to deal with the support requests from the Debian stable users, you're not the one dealing with our build infrastructure, etc.

You're seriously over-interpreting things. :) I'm just trying to understand what's happening in the build system. Since this PR is not going to be merged in the next few days I didn't see any reason to act immediately.

dirkhh:
Written communication sucks at times. Sorry if I came across too intensely

bstoeger:
No worries - at university you grow a very thick skin. :) I guess my communication style is too brief.

To end this drama - updated PR + comment. CHANGELOG.md update after release, because it will break anyway.

PS: Still not certain that `qthelper.h` is the appropriate place for the helper. And also, if the new behavior on deletion is what we actually want.

bstoeger:
Let's close this for now and first do the "detach `ProfileWidget2` from `DivePictureModel`" thing to avoid unnecessary churn.

PR #1439, opened 2018-06-28
[RFC] Bluetooth: don't call deviceDiscoveryError() on scan finished

bstoeger:
In the remoteDeviceScanFinished slot, the old code called into
the deviceDiscoveryError() in case the device discovery agent
had the error flag set. This is not necessary, since the agent will
send an error signal in such a case.

For Qt's device discovery agent, the whole check-for-error is
unnecessary, as the documentation states:
"The signal is not going to be emitted if the device discovery
finishes with an error."

But for the homebrew WinBluetoothDeviceDiscoveryAgent, which
derives from QThread, both an error() *and* a finished()
signal will be sent. Therefore keep the test, but don't call
into the slot twice.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

While analyzing the Windows `WinBluetoothDeviceDiscoveryAgent` code, it occurred to me that on error we call into the error slot twice. Remove this. Note: this is not tested as I'm currently not in front of a Windows machine.
### Changes made:

1) Remove unnecessary call into error slot
2) Add comment why the error check is needed.
### Release note:

None needed.
### Documentation change:

None needed.

### DIFF
--- a/desktop-widgets/btdeviceselectiondialog.cpp
+++ b/desktop-widgets/btdeviceselectiondialog.cpp
@@ -194,11 +194,12 @@ void BtDeviceSelectionDialog::on_scan_clicked()
 
 void BtDeviceSelectionDialog::remoteDeviceScanFinished()
 {
-	if (remoteDeviceDiscoveryAgent->error() == QBluetoothDeviceDiscoveryAgent::NoError) {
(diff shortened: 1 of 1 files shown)


neolit123:
nobody objected -> looks good -> merging.

PR #1444, opened 2018-06-28
Dive pictures: adjust rowDDEnd and rowDDStart on picture deletion

bstoeger:
In DivePictureModel, rowDDEnd and rowDDStart specify the range of
pictures in the profile plot. Obviously, these have to be adjusted
when pictures are deleted.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a bug introduced when making picture deletion "smart". The fundamental problem is that we have three lists of pictures:
1) In the C-backend (all dives)
2) In DivePictureModel (all selected dives)
3) In ProfileWidget2 (currently shown dive)

Keeping these in sync is quite hard and very brittle. The old code chose the easy way: on every change recalculate 2) and 3). Ultimately, this will have to be simplified.

### Changes made:

1) Keep track of changes to rowDDStart and rowDDEnd

### Release note:

None needed - bug introduced recently.
### Documentation change:

None needed.

### DIFF
--- a/qt-models/divepicturemodel.cpp
+++ b/qt-models/divepicturemodel.cpp
@@ -138,6 +138,14 @@ static bool removePictureFromSelectedDive(const char *fileUrl)
 	return false;
 }
 
+// Calculate how many items of a range are before the given index
+static int rangeBefore(int rangeFrom, int rangeTo, int index)
+{
+	if (rangeTo <= rangeFrom)
+		return 0;
+	return std::min(rangeTo, index) - std::min(rangeFrom, index);
+}
+
 void DivePictureModel::removePictures(const QVector<QString> &fileUrls)
 {
 	bool removed = false;
@@ -163,6 +171,12 @@ void DivePictureModel::removePictures(const QVector<QString> &fileUrls)
 		beginRemoveRows(QModelIndex(), i, j - 1);
 		pictures.erase(pictures.begin() + i, pictures.begin() + j);
 		endRemoveRows();
+
+		// After removing pictures, we have to adjust rowDDStart and rowDDEnd.
+		// Calculate the part of the range that is before rowDDStart and rowDDEnd,
+		// respectively and subtract accordingly.
+		rowDDStart -= rangeBefore(i, j, rowDDStart);
+		rowDDEnd -= rangeBefore(i, j, rowDDEnd);
 	}
 }
 


dirkhh:
I assume this needs to go into 4.8.0?
Right now, calling out what you want to see in 4.8.0 would make my life much easier :-)
I'm still hoping to cut the release tomorrow...

bstoeger:
> I assume this needs to go into 4.8.0?

I would recommend, yes. But again, if someone would look at the code, that would be very welcome. Not only is this last-minute, it is also rather hairy.

dirkhh:
This is part of the "heck if I know what this /really/ does" code... @tcanabrava any comments?

bstoeger:
Note that `rowDDend` and `rowDDstart` was introduced by @sfuchs79 to support the pictures of multiple dives being shown in the dive-photo tab, while still showing only the pictures of the actual dive in the profile.

Keeping this all in sync is very brittle and I believe that the best way would be to bypass `DivePictureModel` in `ProfileWidget2`. I.e. fetch the pictures directly from the C-backend. But certainly not for 4.8.0.

bstoeger:
This definitely fixes a bug, which is not very hard to reproduce:

Peparation: Have two consecutive dives with pictures on the profile.
Execution:
1) Select the lower dive in the list (with the lower id).
2) Press "shift + arrow up". Both dives will be selected, the upper dive (higher id will be shown in the profile plot).
3) In the photos-tab delete a picture from the lower-id dive.
4) In the photos-tab delete a picture from the higher-id dive.
5) Boom: the wrong picture (if any) will be removed from the profile. 

You can see why this fails with the following patch:
```
diff --git a/profile-widget/profilewidget2.cpp b/profile-widget/profilewidget2.cpp
index 7570e4deb..57cd3fe0d 100644
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -2063,6 +2063,7 @@ void ProfileWidget2::plotPictures()
 {
        DivePictureModel *m = DivePictureModel::instance();
        pictures.resize(m->rowDDEnd - m->rowDDStart);
+printf("plotPictures(): %d %d\n", m->rowDDStart, m->rowDDEnd);
 
        double x, y, lastX = -1.0, lastY = -1.0;
        for (int i = m->rowDDStart; i < m->rowDDEnd; i++) {
diff --git a/qt-models/divepicturemodel.cpp b/qt-models/divepicturemodel.cpp
index 8c66f5974..352548fef 100644
--- a/qt-models/divepicturemodel.cpp
+++ b/qt-models/divepicturemodel.cpp
@@ -148,6 +148,7 @@ void DivePictureModel::removePictures(const QVector<QString> &fileUrls)
        copy_dive(current_dive, &displayed_dive);
        mark_divelist_changed(true);
 
+printf("removePictures() before: %d %d\n", rowDDStart, rowDDEnd);
        for (int i = 0; i < pictures.size(); ++i) {
                // Find range [i j) of pictures to remove
                if (std::find(fileUrls.begin(), fileUrls.end(), pictures[i].filename) == fileUrls.end())
@@ -164,6 +165,7 @@ void DivePictureModel::removePictures(const QVector<QString> &fileUrls)
                pictures.erase(pictures.begin() + i, pictures.begin() + j);
                endRemoveRows();
        }
+printf("removePictures() after: %d %d\n", rowDDStart, rowDDEnd);
 }
 
 int DivePictureModel::rowCount(const QModelIndex&) const
```

Note that `rowDDStart` and `rowDDEnd` tell the `ProfileWidget2`, which of `DivePictureModel`'s pictures it is responsible for. By not updating these you can probably even make the application crash of you manage to provoke a `plotPictures` event.

My conclusion is that we should detach `ProfileWidget2` from `DivePictureModel`, and this will be my next priority item. It was kind-of obvious as long as both displayed the same list of pictures and lists were simply recalculated, but now it's just not viable anymore.

bstoeger:
Another funky error-mode:

Same situation as above:
1) Select upper dive
2) Shift+Down
3) Delete picture of upper (higher-id) dive
4) Drag & drop a remaining picture of higher-id dive on profile
5) Boom: The profile will show a picture of the lower-id dive

Reason: The profile thinks the it still has _n_ instead of _n-1_ pictures and therefore plots the first picture from the next dive.

bstoeger:
> By not updating these you can probably even make the application crash

Well, at least it doesn't crash, because `QAbstractTableModel::index()` handles out-of-bounds gracefully.

dirkhh:
Excellent, with the explanation and the review from @tcanabrava I'll merge it

PR #1447, opened 2018-06-29
core: remove prefs-macros.h where unused

janiversen:
move #include prefs-macros from SettingsObjectWrapper.h to SettingsObjectWrapper.cpp
include dive.h directly (only part of prefs-macros.h used) in preference classes

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

prefs-macros.h is included in many files where not needed due to being include
in SettingsObjectWrapper.h

classes in desktop-widget/preferences did not need prefs-macros.h but dive.h, corrected

### DIFF
--- a/core/subsurface-qt/SettingsObjectWrapper.cpp
+++ b/core/subsurface-qt/SettingsObjectWrapper.cpp
@@ -7,6 +7,7 @@
 #include <QNetworkProxy>
 
 #include "core/qthelper.h"
+#include "core/prefs-macros.h"
 
 DiveComputerSettings::DiveComputerSettings(QObject *parent):
 	QObject(parent)
--- a/core/divecomputer.cpp
+++ b/core/divecomputer.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "divecomputer.h"
+#include "dive.h"
 #include "subsurface-string.h"
 #include "subsurface-qt/SettingsObjectWrapper.h"
 
--- a/core/subsurface-qt/SettingsObjectWrapper.h
+++ b/core/subsurface-qt/SettingsObjectWrapper.h
@@ -5,8 +5,7 @@
 #include <QObject>
 #include <QDate>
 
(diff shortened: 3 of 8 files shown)


bstoeger (approved):
lgtm.

core/subsurface-qt/SettingsObjectWrapper.cpp
```diff
@@ -7,6 +7,7 @@
 #include <QNetworkProxy>
 
 #include "core/qthelper.h"
+#include "../prefs-macros.h"
```
dirkhh: I know we are not consistent in this, but as you work towards consistency, would you mind adding header file references relative to $SRC - so `core/prefs-macros.h` instead of `../prefs-macros.h` ?

dirkhh (changes requested):
I very much like simplifications like this.
Small comment with a request for a change.

core/subsurface-qt/SettingsObjectWrapper.cpp
janiversen: Good point, changed.

janiversen:
@dirkhh hope you like the new version. I will split things as good as I can, but I have about 50 commits piled up in different branches, so I need to get decision on which way we go (or not go) with qPref... I think the basic question is do we
a) keep the very split design we have today (with its high possibility for errors)
b) wait for a new super generator, which may or may not solve all the exceptions.
c) accept my base design, and correct with good suggestions (like the one from bstoeger about macros)

I would really prefer to know where we are heading, so I know if my plans are valid or to be thrown overboard.

dirkhh:
I am ok with the design as you are working on it. If @tcanabrava finds the time to continue the work on the code generator, that will be rather experimental at first and may take some time. I don't believe it makes sense to stop forward progress because of that.

bstoeger:
Since @dirkhh's change request was implemented, I think that we can merge this.

PR #1448, opened 2018-06-29
core: make stuff more readable

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Isolate first part of #1423 in this PR, since these 5 commits make a standalone change.

Sorting/grouping can always be discussed, but having names randomly distributed is not a good starting point for a new developer trying to learn.

### Changes made:

1) sort/group struct preferences in pref.h making it easier to locate variables (when you do not know the name)
2) move PP_GRAPHS_ENABLED from pref.h, since it is only used in 1 cpp file
3) sort CMakeLists.txt, to make it faster to look if a file is included or not (did this after I mistakenly added the same file twice)
4) add qPref.h and qPrefprivate.h, this is the fundament of #1423
5) remove double definition of enum cloud_storage_status, having the same enum defined in several places is error prone.

### DIFF
--- a/core/pref.h
+++ b/core/pref.h
@@ -223,8 +223,6 @@ enum cloud_status {
 
 extern struct preferences prefs, default_prefs, git_prefs;
 
-#define PP_GRAPHS_ENABLED (prefs.pp_graphs.po2 || prefs.pp_graphs.pn2 || prefs.pp_graphs.phe)
-
 extern const char *system_divelist_default_font;
 extern double system_divelist_default_font_size;
 
--- a/core/pref.h
+++ b/core/pref.h
@@ -59,106 +59,147 @@ typedef struct {
 	int download_mode;
 } dive_computer_prefs_t;
 
+// ********** PREFERENCES **********
+// This struct is kept global for all of ssrf
+// most of the fields are loaded from git as
+// part of the dives, but some fields are loaded
+// from local storage (QSettings)
+// The struct is divided in groups (sorted)
+// and elements within the group is sorted
+//
+// When adding items to this list, please keep
+// the list sorted (easier to find something)
 struct preferences {
-	const char *divelist_font;
-	const char *default_filename;
-	const char *default_cylinder;
+	// ********** Animations **********
+	int animation_speed;
+
+	// ********** CloudStorage **********
 	const char *cloud_base_url;
 	const char *cloud_git_url;
-	const char *time_format;
-	const char *date_format;
-	const char *date_format_short;
-	bool time_format_override;
-	bool date_format_override;
-	double font_size;
-	partial_pressure_graphs_t pp_graphs;
-	bool mod;
-	double modpO2;
-	bool ead;
-	bool dcceiling;
-	bool redceiling;
-	bool calcceiling;
-	bool calcceiling3m;
-	bool calcalltissues;
-	bool calcndltts;
-	short gflow;
-	short gfhigh;
-	int animation_speed;
-	bool gf_low_at_maxdepth;
-	bool show_ccr_setpoint;
-	bool show_ccr_sensors;
-	bool show_scr_ocpo2;
-	bool display_invalid_dives;
-	short unit_system;
-	struct units units;
-	bool coordinates_traditional;
-	bool show_sac;
-	bool display_unused_tanks;
-	bool show_average_depth;
-	bool show_icd;
-	bool zoomed_plot;
-	bool hrgraph;
-	bool percentagegraph;
-	bool rulergraph;
-	bool tankbar;
-	bool save_userid_local;
-	const char *userid;
-	int ascrate75; // All rates in mm / sec
-	int ascrate50;
-	int ascratestops;
-	int ascratelast6m;
-	int descrate;
-	int sacfactor;
-	int problemsolvingtime;
-	int bottompo2;
-	int decopo2;
-	enum deco_mode display_deco_mode;
-	depth_t bestmixend;
-	int proxy_type;
-	const char *proxy_host;
-	int proxy_port;
-	bool proxy_auth;
-	const char *proxy_user;
-	const char *proxy_pass;
-	bool doo2breaks;
-	bool drop_stone_mode;
-	bool last_stop;   // At 6m?
-	bool verbatim_plan;
-	bool display_runtime;
-	bool display_duration;
-	bool display_transitions;
-	bool display_variations;
-	bool safetystop;
-	bool switch_at_req_stop;
-	int reserve_gas;
-	int min_switch_duration; // seconds
-	int bottomsac;
-	int decosac;
-	int o2consumption; // ml per min
-	int pscr_ratio; // dump ratio times 1000
-	int defaultsetpoint; // default setpoint in mbar
-	bool show_pictures_in_profile;
-	bool use_default_file;
-	short default_file_behavior;
-	facebook_prefs_t facebook;
-	const char *cloud_storage_password;
-	const char *cloud_storage_newpassword;
 	const char *cloud_storage_email;
 	const char *cloud_storage_email_encoded;
-	bool save_password_local;
-	short cloud_verification_status;
+	const char *cloud_storage_newpassword;
+	const char *cloud_storage_password;
+	const char *cloud_storage_pin;
+	short       cloud_timeout;
+	short       cloud_verification_status;
+	bool        git_local_only;
+	bool        save_password_local;
+	bool        save_userid_local;
+	const char *userid;
+
+	// ********** DiveComputer **********
+	dive_computer_prefs_t dive_computer;
+
+	// ********** Display **********
+	bool        display_invalid_dives;
+	const char *divelist_font;
+	double      font_size;
+	bool        showDeveloper;
+ 	const char *theme;
+
+	// ********** Facebook **********
+	facebook_prefs_t facebook;
+
+	// ********** General **********
+	bool        auto_recalculate_thumbnails;
+	int         defaultsetpoint; // default setpoint in mbar
+	const char *default_cylinder;
+	const char *default_filename;
+	short       default_file_behavior;
+	int         o2consumption; // ml per min
+	int         pscr_ratio; // dump ratio times 1000
+	bool        use_default_file;
+
+	// ********** Geocoding **********
 	geocoding_prefs_t geocoding;
-	enum deco_mode planner_deco_mode;
-	short vpmb_conservatism;
+
+	// ********** Language **********
+	const char *    date_format;
+	bool            date_format_override;
+	const char *    date_format_short;
+	locale_prefs_t  locale; //: TODO: move the rest of locale based info here.
+	const char *    time_format;
+	bool            time_format_override;
+
+	// ********** LocationService **********
 	int time_threshold;
 	int distance_threshold;
-	bool git_local_only;
-	short cloud_timeout;
-	bool auto_recalculate_thumbnails;
-	locale_prefs_t locale; //: TODO: move the rest of locale based info here.
+
+	// ********** Network **********
+	bool        proxy_auth;
+	const char *proxy_host;
+	int         proxy_port;
+	int         proxy_type;
+	const char *proxy_user;
+	const char *proxy_pass;
+
+	// ********** Planner **********
+	int             ascratelast6m;
+	int             ascratestops;
+	int             ascrate50;
+	int             ascrate75; // All rates in mm / sec
+	depth_t         bestmixend;
+	int             bottompo2;
+	int             bottomsac;
+	int             decopo2;
+	int             decosac;
+	int             descrate;
+	bool            display_duration;
+	bool            display_runtime;
+	bool            display_transitions;
+	bool            display_variations;
+	bool            doo2breaks;
+	bool            drop_stone_mode;
+	bool            last_stop;   // At 6m?
+	int             min_switch_duration; // seconds
+	enum deco_mode  planner_deco_mode;
+	int             problemsolvingtime;
+	int             reserve_gas;
+	int             sacfactor;
+	bool            safetystop;
+	bool            switch_at_req_stop;
+	bool            verbatim_plan;
+
+	// ********** TecDetails **********
+	bool                        calcalltissues;
+	bool                        calcceiling;
+	bool                        calcceiling3m;
+	bool                        calcndltts;
+	bool                        dcceiling;
+	enum deco_mode              display_deco_mode;
+	bool                        display_unused_tanks;
+	bool                        ead;
+	short                       gfhigh;
+	short                       gflow;
+	bool                        gf_low_at_maxdepth;
+	bool                        hrgraph;
+	bool                        mod;
+	double                      modpO2;
+	bool                        percentagegraph;
+	partial_pressure_graphs_t   pp_graphs;
+	bool                        redceiling;
+	bool                        rulergraph;
+	bool                        show_average_depth;
+	bool                        show_ccr_sensors;
+	bool                        show_ccr_setpoint;
+	bool                        show_icd;
+	bool                        show_pictures_in_profile;
+	bool                        show_sac;
+	bool                        show_scr_ocpo2;
+	bool                        tankbar;
+	short                       vpmb_conservatism;
+	bool                        zoomed_plot;
+
+	// ********** Units **********
+	bool            coordinates_traditional;
+	short           unit_system;
+	struct units    units;
+
+	// ********** UpdateManager **********
 	update_manager_prefs_t update_manager;
-	dive_computer_prefs_t dive_computer;
 };
+
 enum unit_system_values {
 	METRIC,
 	IMPERIAL,
--- a/core/settings/qPref.cpp
+++ b/core/settings/qPref.cpp
@@ -2,16 +2,15 @@
 #include "qPref_private.h"
 #include "qPref.h"
 
-
-qPref *qPref::m_instance = NULL;
+qPref::qPref(QObject *parent) : 
+	QObject(parent)
+{
+}
 qPref *qPref::instance()
 {
-	if (!m_instance)
-		m_instance = new qPref;
-	return m_instance;
+static qPref *self = new qPref;
+	return self;
 }
-
-
 
 void qPref::loadSync(bool doSync)
 {
--- a/core/settings/qPref.cpp
+++ b/core/settings/qPref.cpp
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "qPref_private.h"
+#include "qPref.h"
+
+
+qPref *qPref::m_instance = NULL;
+qPref *qPref::instance()
+{
+	if (!m_instance)
+		m_instance = new qPref;
+	return m_instance;
+}
+
+
+
+void qPref::loadSync(bool doSync)
+{
+}
--- a/core/settings/qPref.h
+++ b/core/settings/qPref.h
@@ -7,10 +7,10 @@
 
 class qPref : public QObject {
 	Q_OBJECT
+	Q_ENUMS(cloud_status);
 
 public:
-	qPref(QObject *parent = NULL) : QObject(parent) {};
-	~qPref() {};
+	qPref(QObject *parent = NULL);
 	static qPref *instance();
 
 	// Load/Sync local settings (disk) and struct preference
@@ -19,7 +19,6 @@ public:
 public:
 
 private:
-	static qPref *m_instance;
 };
 
 #endif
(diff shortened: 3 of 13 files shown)


core/settings/qPref.h
```diff
@@ -0,0 +1,38 @@
...
+public:
+	qPref(QObject *parent = NULL) : QObject(parent) {};
+	~qPref() {};
+	static qPref *instance();
+
+	// Load/Sync local settings (disk) and struct preference
+	void loadSync(bool doSync);
+
+public:
+	enum cloud_status {
```
bstoeger: I admit that I haven't really looked at the other PR, because the preferences code is the part which I am the least fond of.

So just a random note: Why would `cloud_status` belong to the preferences-class? This seems to me fundamentally wrong. So I wonder whether a better approach would be to declutter the preferences first before doing whatever this PR is doing?

janiversen:
That is a good point, before this PR it was both in pref.h and preferences. I choose qPref because this status is only used in qml as far as I remember.

It is needed as an enum in qml, and I think those enums needs to be in the registred class. However if desired I can try an circumvent that, however having the definition 2 times (as now) is bad.

core/settings/qPref_private.h
```diff
@@ -0,0 +1,74 @@
...
+#include <QVariant>
+
+#include "core/qthelper.h"
+
+
+#define COPY_TXT(name, string)	\
+{		\
+	free((void *)prefs.name);	\
+	prefs.name = copy_qstring(string);	\
+}
```
dirkhh: at least here on GitHub the trailing backslashes seem rather randomly distributed :-)

core/settings/qPref.h
dirkhh: I agree that this is not a preference - but it is stored with the preferences... now you can argue "why do we do that?" and that's an interesting question. Preferences and cloud status are both part of the persistent state that we maintain across runs. We can now consider renaming the thing that we store - but I'm not sure that adds a lot of value.

dirkhh (approved):
Thanks for breaking this out as a smaller subset.
I think this looks good to go in after 4.8.0
Each of these changes makes things simpler and easier to read - and that's a good thing.

core/settings/qPref_private.h
neolit123: aligned spaces or a single space should be used before the `\`, not tabs.

core/settings/qPref.h
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef QPREF_H
+#define QPREF_H
+
+
```
neolit123: extra LF.

core/settings/qPref.h
```diff
@@ -0,0 +1,38 @@
...
+#include <QObject>
+#include "core/pref.h"
+
+
+
```
neolit123: two extra LF.

core/settings/qPref.h
```diff
@@ -0,0 +1,38 @@
...
+public:
+	enum cloud_status {
+		CS_UNKNOWN,
+		CS_INCORRECT_USER_PASSWD,
+		CS_NEED_TO_VERIFY,
+		CS_VERIFIED,
+		CS_NOCLOUD
+	};
+
+
```
neolit123: extra LF.

core/settings/qPref_private.h
```diff
@@ -0,0 +1,74 @@
...
+
+// Header used by all qPref<class> implementations to avoid duplicating code
+
+#include <QObject>
+#include <QSettings>
+#include <QVariant>
+
+#include "core/qthelper.h"
+
+
```
neolit123: extra LF.

core/settings/qPref_private.h
janiversen: Done (changes to a single space and \ not aligned)

core/pref.h
```diff
@@ -59,106 +59,147 @@ typedef struct {
...
+// ********** PREFERENCES **********
+// This struct is kept global for all of ssrf
+// most of the fields are loaded from git as
+// part of the dives, but some fields are loaded
+// from local storage (QSettings)
+// The struct is divided in groups (sorted)
+// and elements within the group is sorted
+//
+// When adding items to this list, please keep
+// the list sorted (easier to find something)
```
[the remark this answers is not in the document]
janiversen: yep, that was tab once again (my editor now inserts tabs whenever it can :-( )

janiversen:
Thanks for the comments, I have pushed a new version, which I hope solves all comments.

No need for this to be merged before the release.

core/settings/qPref.h
bstoeger: >  Preferences and cloud status are both part of the persistent state that we maintain across runs.

So are many other things which are *not* stored in the preferences (last used directories, thumbnails, ...). Just grep for `QSettings`.

> but I'm not sure that adds a lot of value

I guess that depends on your priorities. But since this is a major rewrite/renaming anyway, I think we should ask such questions.

Even the code says:
```
//TODO: Change this to the Cloud Storage Stuff, not preferences.
```

BTW: I really dislike the verbosity of the whole preferences thing, so just seeing `+356 −191` makes me sad. I reckon that's a temporary thing?

core/settings/qPref_private.h
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#define LOADSYNC_INT(name, field) \
+{ \
+	QSettings s; \
+	if (doSync) \
+		s.setValue(group + name, prefs.field); \
+	else \
+		prefs.field = s.value(group + name, default_prefs.field).toInt(); \
+}
```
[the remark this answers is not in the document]
bstoeger: > Is there a reason to create a new object every time?

If you look at #1423, these macros seem to replace a full function body. Which then raises the question, why not generate the whole function with a macro?

I think these macros should go in a PR where their use is seen. Otherwise it's really difficult to judge their usefulness.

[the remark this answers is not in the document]
mfbernardes: Agree they should not be in this PR.

Moreover, even though they replace a function body... they replace a getter, that is also a setter.  and called by the setter, to actually set/save values. This does not make sense to me.

The setter code is always the same anyway, is just one line and can be made more clear by doing just that.

core/settings/qPref.h
```diff
@@ -0,0 +1,25 @@
...
+class qPref : public QObject {
+	Q_OBJECT
+
+public:
+	qPref(QObject *parent = NULL) : QObject(parent) {};
+	~qPref() {};
+	static qPref *instance();
+
+	// Load/Sync local settings (disk) and struct preference
+	void loadSync(bool doSync);
```
mfbernardes: What is the advantage of this loadSync vs just loading as needed (or constructor) and just setting on setter?

I only see a centralized sync making sense if setters are not used (like in the existing code), but this is not the case I see on #1423 where qPref is used.

core/settings/qPref.h
dirkhh: You are correct - this is the time to do things right. So we should move cloud storage out from preferences. That doesn't need to happen in this particular patch series, though.

@tcanabrava had worked on something to create the preferences and all the boilerplate code from much more concise configuration data. I don't know if that code would be an alternative to what Jan is working on.

janiversen:
Just a few short comments from the airport while I wait for a delayed plane.
- I agree with the cloudStorage enum, but I had some problems with the Qt macro if not defined in the class (I assume that is why we 2 definitions right now. I will to move it to oref.h where it really belongs.
- There are only 1 funcion to load/sync on purpose, this is to ensure thatthe same name is used.
- the get/set/disk functions are side by side, so we ensure all 3 are added (look inthe curent code, where get set load sync are split over a lot of lines), I found several problems in the current distributed version, so this is sonething important (at least to me).
- I checked if a parser could be used, but please look at all the exceptions we have, that would make it less efficient than you think.
- bstoeger has a good argument why the macro does not generate the whole function. I did that in an early attempt, but decided sooneone would not like that. my preference is to have 1 macro that generates all 3 functions, so we only have to write the special cases.

note to @bstoeger you can see how the macros are used in the other PR.

@dirkhh I am very willing to make adaptations, because I hope the outcome will be slimmer and more readable, but please lets not reimplement what is already there in the same manner (like e.g. a separate function for load and sync).

core/settings/qPref.h
janiversen: well at least we need a centralized read. And when savi g it would be good to sync as well, as many change struct preferences directly (no set call)

janiversen:
I have remove cloud_status from qPref (now it is only defined in pref.h) and it seems to work with QML.

Regarding the macros, these are a starting point, but the comments from @bestoeger and @mfbernardes  gave me a couple of ideas. My suggestion is that the current macros are accepted as is, and then I will prepare a simple PR that contains extra macros:
- Whole disk function as pr @bstoeger 
- Something similar for the get function (which are 98% identical)
- a single class (Display) that shows how the macros are used.

That enables us to focus the discussion on the macros and the structure of qPref.

core/settings/qPref_private.h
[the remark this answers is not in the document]
dirkhh: It looks like this comment is still unresolved?

dirkhh (changes requested):
I think we are getting closer. It seems the getter / setter semantics still have an open question or two.
The extra line breaks I can fix when I pull it (if I remember).
Thanks

janiversen:
Please help a pour soul, I am unsure which comment is unanswwered, but let me just repeat the headlines:

Getter functions:
Is in this version, normal C++ functions, but as @bstoeger suggest willl be replaced be a single line macro for the 98%).

Setter functions:
Is in this version, normal C++ functions, but as @bstoeger suggest willl be replaced be a single line macro for the 80%). We do however need the e.g. COPY_TXT macros for the remaining 20%

Disk functions
Is in this version, normal C++ function with a macro call, but as @bstoeger suggest willl be replaced be a single line macro for the 75%). We do however need simple macros for the remaining 25%. The creatiion of QSettings will be done as a private variable, but I would like to do that in one commit.

Once this is accepted, I have another PR nearly ready to go, it builds on this one, respecting most of the comments (with the exception I do not fancy 1 diskSync and 1 diskLoad function, since we have to duplicate the storage name). The next PR will contain:
1) amendsments to qPRef_private.h for “all inclusive functions”
2) qPrefDiskplay showing how the macros are used.

If I have missed anything, please tell me what I need to correct/explain in this PR ?

dirkhh:
@bstoeger are you ok with this being merged? I had understood that you had asked some of the changes to be made here, but now reading @janiversen 's summary, it does sound like we are good to go?

core/settings/qPref.h
```diff
@@ -0,0 +1,26 @@
...
+#include <QObject>
+#include "core/pref.h"
+
+class qPref : public QObject {
+	Q_OBJECT
+	Q_ENUMS(cloud_status);
+
+public:
+	qPref(QObject *parent = NULL) : QObject(parent) {};
+	~qPref() {};
```
bstoeger: Any reason to overwrite the default-destructor by an inlined version?
(Note: If you do, then the semicolon is not needed.)

core/settings/qPref.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include <QObject>
+#include "core/pref.h"
+
+class qPref : public QObject {
+	Q_OBJECT
+	Q_ENUMS(cloud_status);
+
+public:
+	qPref(QObject *parent = NULL) : QObject(parent) {};
```
bstoeger: Any reason to inline the constructor?
(Note: If you do, then the semicolon is not needed.)

core/settings/qPref_private.h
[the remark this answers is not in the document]
bstoeger: I still would prefer if these macros were introduced with a PR that also introduces their use. Otherwise it's hard to get the "vision". Note: not necessarily the same commit, but the same PR would be nice.

core/settings/qPref.h
janiversen: The reason was that if I did not make it inline, I got a compile error if it was not declared in the cpp file. But maybe I made something wrong ?

I am happy the remove the {}; if I do not have to declare it in the cpp file.

core/settings/qPref.cpp
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "qPref_private.h"
+#include "qPref.h"
+
+
+qPref *qPref::m_instance = NULL;
+qPref *qPref::instance()
+{
+	if (!m_instance)
```
bstoeger: I think we discussed this already, but I forgot: Any reason not to use the standard singleton idiom:
```
qPref *qPref::instance()
{
	qPref self;
	return &self;
}
```
If this doesn't work for some subtle reason then it needs a comment.

janiversen:
That is precise what I am doing in the following PR, however it is based on this PR, so I either combine the 2 PR´2 (making the second less logical), and we keep it as 2 PR´s.

I understand and agree it is good to have a usecase in parallel to the macros, but please remember not the least due to your comments this is about 1/3 of the total macros.

I will leave it up to you and dirk decide what you prefer (1 or 2 PR, same work for me). I personally think a second commit that includes all macros with a use case is better than mixing the PR

As I wrote earlier you can see how these macros are used in the now outdated PR #1447. But remark I am making more intelligent use of macros due to your suggestions.

janiversen:
Removed merge conflict with latest master and got rid of last extra line breaks. 

Now I wait for a decision on whether or not to merge the 2 PR

core/settings/qPref.h
bstoeger: > The reason was that if I did not make it inline, I got a compile error if it was not declared in the cpp file. 

Of course - if you don't inline, you have to define it somewhere.

> I am happy the remove the {}; if I do not have to declare it in the cpp file.

Not sure that I understand.
1) We define the constructor of every other class in the .cpp file, so why would this be a problem here?
2) You don't pass a parent object anyway, so you can completely remove the declaration and just use the default constructor.

The destructor can also be removed. The default-constructor is good enough.

Note: If you have `{}`, then you don't need `;':
```
	qPref(QObject *parent = NULL) : QObject(parent) {}
```

bstoeger:
> @bstoeger are you ok with this being merged? I had understood that you had asked some of the changes to be made here, but now reading @janiversen 's summary, it does sound like we are good to go?

As I said before: I don't get the big vision of this, but this is due to the fact that I'm not a fan of the preferences code / Qt's property system in general, not due to the PR itself. Apart from the coding-style issues I don't have anything to object, but I don't see the big gain either (well, the formatting of the prefs struct is improved).

janiversen (commented):
I thought an inline constructor ware more efficient, and at least less lines than declaring it in the cpp file. We need the constructor to be sure QObject is called with parent, but I am happy to put it back in the cpp file. The ; is a part of my way of working, even though it is not needed it does harm, and sometimes it catches errors.

I will remove the destructor and let the default destructor take over.

janiversen:
Response to @bstoeger comments:
- removed destructor to allow default destructor
- remove inline constructor and define in cpp file, slightly less efficient and more lines of code, but no subject

I did not change the instance() function, because unless I am mistaken the suggestion does not work (@bstoeger did not yet respond to my comments).

The big vision, might not be so big at all, but look at the current code, and find the missing variables etc, that a “just code cleanup” is a vision.

janiversen:
Seems travis is broken cannot find our domain.

bstoeger:
> I thought an inline constructor ware more efficient, and at least less lines than declaring it in the cpp file. 

Let's not sacrifice consistency for a function that is called once(!) during the application's lifetime.

dirkhh:
Travis isn't having a good day.
@bstoeger we use the same instance pattern in many places. The reason for this is that otherwise Qt and QML create two different instances and things go badly. This may be a bug elsewhere in our code, but unless we have a replacement that is tested and works everywhere, I don't really want to start yet another "oh, we don't need this" followed by a week of me hunting down strange bugs and in the end finding another case where we have two different instances of what we think is the same variable. I have gone through this three times now and the third time I swore that I would kill the next person who breaks this...
@janiversen my biggest concern is "cleanups" that break things. Your "cleanup" of the translation location broke translations. Because you moved them without updating the place where we look for them. And because the iOS simulator once again is broken and doesn't correctly set its locale (whatever I do, I get en-DE instead of de-DE) I have wasted two hours trying to test my fix.
And that's the crux of ALL of this. So many times things get done without being sufficiently tested and thought through, and they end up costing ME hours and hours of work.
I don't mind cleanups. I do mind if for little or now objective incremental value things get broken and I end up being the one who spends the time to fix them, instead of working on new features or things that are actually fun to work on.
Yes, the problem is that this is all incredibly fragile and ugly and hacky and what not. But that is where we are and what we have.
No, I'm not saying NO to this PR. I'll take it once I've beaten Travis into submission and once @bstoeger is reasonably happy.
But I will be rather unhappy if this again ends up breaking things in subtle ways.

janiversen:
Updated to use 
Static qPref *self in instance() instead of a m_instance as a private variable.

I do not see the big difference, but on the other hand I have no firm opinion, I just hope something changes the rest of the system

Git grep m_instance | wc -l give 46.

janiversen:
I am sorry to have caused trouble with the build.sh iOS patch, seems I did not test it sufficiently enough. I do test and when I make mistakes I try to correct them instead of washing my hands ! I am right now limited in being able to help, but I did point you at the potential problem.

This patch does not change the live system, but it is sure if we continue with my qPref changes that there will be some things to correct, the only way to avoid this is 
1) to require test cases (something I will supply to every single qPref class, in the same PR it goes life)
2) test more.

Or of course, do nothing then we only have know errors  :-)

I am honestly not trying to break things, but to help make ssrf better, and of course by doing that, sometimes there are mistakes, I am only human, but I believe not be alone in that aspect.


Ps using m_instance as a static private variable, or having static *self in instance() does not change how qml works. BUT e.g. the current class variables in the qml classes are a problem.

core/settings/qPref.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include "qPref_private.h"
+#include "qPref.h"
+
+qPref::qPref(QObject *parent) : 
+	QObject(parent)
+{
+}
+qPref *qPref::instance()
+{
+static qPref *self;
```
bstoeger: Sorry for being so annoying, but...
1) indentation
2) my point was to make it
`static qPref *self = new qPref;`

The reason: This is guaranteed by the compiler to be thread safe (which admittedly is certainly not an issue here). The compiler does all the locking magic, which the version with `if` does not.

@dirkhh: I don't see how this could ever produce two different `qPref` objects.

bstoeger:
> Git grep m_instance | wc -l give 46.

And I have a patch removing most of them. :)

>  @bstoeger is reasonably happy.

I guess I am. If there turns out to be things that rub me the wrong way, I can always clean them up later. :)

janiversen:
@bstoeger now we are talking, I like that way and have changed it, however you do forget one think, QML does not call instance() it calls the constructor and that is something we cannot control.

The only way to make sure QML (when calling the constructor) does not make problems, is to make sure the class does not have object variables (which I secure in my rewrite).

bstoeger:
> QML does not call instance() it calls the constructor and that is something we cannot control.

Yeah, I forgot. 🙄

That's the reason why my singleton-patch makes the constructors private.

janiversen:
@bstoeger please if you feel I submit things that rub you the wrong way, please tell me first. Sometimes (as with the constructor) I actually know what I do, and sometimes I make mistakes or just plainly have a stupid chain of thoughts.

Ps. I am all in favour of having a private constructor with singleton classes, and have spent time long time ago trying to convince the Qt people that the QML register call (see subsurface-mobile-helper.cpp)  misses a parameter of how to construct the object, but they had different ideas (interface classes).

dirkhh:
> I am honestly not trying to break things, but to help make ssrf better, and of course by doing that, sometimes there are mistakes, I am only human, but I believe not be alone in that aspect.

I know you aren't trying to break things. I wasn't planning to imply that you were.

dirkhh (approved):

dirkhh:
So this looks like everything has been addressed. Let me beat Travis into submission then...

dirkhh:
Travis is really trying to annoy me today. It's succeeding.

dirkhh:
@bstoeger if there was still something that you think would be better if done differently, could you send a PR? It seemed that everything was addressed and I didn't want to keep it hanging here any longer - now we have most of the outstanding things in master and can get all of it tested again.
But it's possible I missed something give just how many comments we had here ;-)

PR #1449, opened 2018-06-30
[RFC] Dive pictures: detach ProfileWidget2 from DivePictureModel

bstoeger:
As long as ProfileWidget2 and DivePictureModel showed the same set of
pictures and any change would lead to a full recalculation of the set,
it made sense to let ProfileWidget2 use DivePictureModel's data.

Recently, keeping the two lists in sync become more and more of a
burden. Therefore, disconnect ProfileWidget2 and DivePictureModel. This
will lead to some code-duplication and perhaps a temporary drop in
UI-performance, but in the end the code is distinctly simpler and also
more flexible.

Thus, for example the DivePhotoTab could be changed to support headings
without having to touch ProfileWidget2 at all.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As mentioned in previous PR's, this disconnects `DivePictureModel` and `ProfileWidget2`. It was much less painful than anticipated. I believe this will not only make things simpler but distinctly more flexible.
### Changes made:

1) Pass a filename instead of a complex struct to `Thumbnailer::fetchThumbnail()`
2) Keep track of image filename and offset in `ProfileWidget2`
3) Directly access the thumbnailer in `ProfileWidget2`
4) Remove `rowDDStart`, `rowDDEnd` and `defaultSite` from `DivePictureModel`
### Release note:

Nothing user visible (hopefully)
### Documentation change:

Nothing user visible (hopefully)

### Mentions:

@tcanabrava @sfuchs79

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -23,6 +23,7 @@
 #include "desktop-widgets/divepicturewidget.h"
 #include "core/qthelper.h"
 #include "core/gettextfromc.h"
+#include "core/imagedownloader.h"
 #endif
 
 #include <libdivecomputer/parser.h>
@@ -152,9 +153,9 @@ ProfileWidget2::ProfileWidget2(QWidget *parent) : QGraphicsView(parent),
 	addActionShortcut(Qt::Key_Left, &ProfileWidget2::keyLeftAction);
 	addActionShortcut(Qt::Key_Right, &ProfileWidget2::keyRightAction);
 
-	connect(DivePictureModel::instance(), &DivePictureModel::dataChanged, this, &ProfileWidget2::updatePictures);
+	connect(Thumbnailer::instance(), &Thumbnailer::thumbnailChanged, this, &ProfileWidget2::updateThumbnail, Qt::QueuedConnection);
 	connect(DivePictureModel::instance(), SIGNAL(rowsInserted(const QModelIndex &, int, int)), this, SLOT(plotPictures()));
-	connect(DivePictureModel::instance(), &DivePictureModel::rowsRemoved, this, &ProfileWidget2::removePictures);
+	connect(DivePictureModel::instance(), &DivePictureModel::picturesRemoved, this, &ProfileWidget2::removePictures);
 	connect(DivePictureModel::instance(), &DivePictureModel::modelReset, this, &ProfileWidget2::plotPictures);
 #endif // SUBSURFACE_MOBILE
 
@@ -2044,49 +2045,70 @@ void ProfileWidget2::clearPictures()
 	pictures.clear();
 }
 
-void ProfileWidget2::updatePictures(const QModelIndex &from, const QModelIndex &to)
+// This function is called asynchronously by the thumbnailer if a thumbnail
+// was fetched from disk or freshly calculated.
+void ProfileWidget2::updateThumbnail(QString filename, QImage thumbnail)
 {
-	DivePictureModel *m = DivePictureModel::instance();
-	for (int picNr = from.row(); picNr <= to.row(); ++picNr) {
-		int picItemNr = picNr - m->rowDDStart;
-		if (picItemNr < 0 || (size_t)picItemNr >= pictures.size())
-			return;
-		if (!pictures[picItemNr])
-			return;
+	// Find the picture with the given filename
+	auto it = std::find_if(pictures.begin(), pictures.end(), [&filename](const PictureEntry &e)
+			       { return e.filename == filename; });
 
-		pictures[picItemNr]->setPixmap(m->index(picNr, 0).data(Qt::UserRole).value<QPixmap>());
+	// If we didn't find a picture, it does either not belong to the current dive,
+	// or its timestamp is outside of the profile.
+	if (it != pictures.end()) {
+		// Replace the pixmap of the thumbnail with the newly calculated one.
+		int size = Thumbnailer::defaultThumbnailSize();
+		it->thumbnail->setPixmap(QPixmap::fromImage(thumbnail.scaled(size, size, Qt::KeepAspectRatio)));
 	}
 }
 
+ProfileWidget2::PictureEntry::PictureEntry (offset_t offsetIn, const QString &filenameIn) : offset(offsetIn),
+	filename(filenameIn),
+	thumbnail(new DivePictureItem)
+{
+}
+
+// Define a default sort order for picture-entries: sort lexicographically by timestamp and filename.
+bool ProfileWidget2::PictureEntry::operator< (const PictureEntry &e) const
+{
+	// Use std::tie() for lexicographical sorting.
+	return std::tie(offset.seconds, filename) < std::tie(e.offset.seconds, e.filename);
+}
+
+// This function resets the picture thumbnails of the current dive.
 void ProfileWidget2::plotPictures()
 {
-	DivePictureModel *m = DivePictureModel::instance();
-	pictures.resize(m->rowDDEnd - m->rowDDStart);
+	pictures.clear();
+	if (currentState == ADD || currentState == PLAN)
+		return;
 
+	// Fetch all pictures of the current dive, but consider only those that are within the dive time.
+	// For each picture, create a PictureEntry object in the pictures-vector.
+	// emplace_back() constructs an object at the end of the vector. The parameters are passed directly to the constructor.
+	FOR_EACH_PICTURE(current_dive) {
+		if (picture->offset.seconds > 0 && picture->offset.seconds <= current_dive->duration.seconds)
+			pictures.emplace_back(picture->offset, QString(picture->filename));
+	}
+	if (pictures.empty())
+		return;
+	// Sort pictures by timestamp (and filename if equal timestamps).
+	// This will allow for proper location of the pictures on the profile plot.
+	std::sort(pictures.begin(), pictures.end());
+
+	// Add the DivePictureItems to the scene, set their pixmaps and filenames
+	// and finaly calculate their positions.
 	double x, y, lastX = -1.0, lastY = -1.0;
-	for (int i = m->rowDDStart; i < m->rowDDEnd; i++) {
-		int picItemNr = i - m->rowDDStart;
-		int offsetSeconds = m->index(i, 1).data(Qt::UserRole).value<int>();
-		// it's a correct picture, but doesn't have a timestamp: only show on the widget near the
-		// information area. A null pointer in the pictures array indicates that this picture is not
-		// shown.
-		if (!offsetSeconds) {
-			pictures[picItemNr].reset();
-			continue;
-		}
-		DivePictureItem *item = pictures[picItemNr].get();
-		if (!item) {
-			item = new DivePictureItem;
-			pictures[picItemNr].reset(item);
-			scene()->addItem(item);
-		}
-		item->setVisible(prefs.show_pictures_in_profile);
-		item->setPixmap(m->index(i, 0).data(Qt::UserRole).value<QPixmap>());
-		item->setFileUrl(m->index(i, 1).data().toString());
+	int size = Thumbnailer::defaultThumbnailSize();
+	for (PictureEntry &e: pictures) {
+		scene()->addItem(e.thumbnail.get());
+		e.thumbnail->setVisible(prefs.show_pictures_in_profile);
+		QImage thumbnail = Thumbnailer::instance()->fetchThumbnail(e.filename).scaled(size, size, Qt::KeepAspectRatio);
+		e.thumbnail->setPixmap(QPixmap::fromImage(thumbnail));
+		e.thumbnail->setFileUrl(e.filename);
 		// let's put the picture at the correct time, but at a fixed "depth" on the profile
 		// not sure this is ideal, but it seems to look right.
-		x = timeAxis->posAtValue(offsetSeconds);
-		if (i == 0)
+		x = timeAxis->posAtValue(e.offset.seconds);
+		if (lastX < 0.0)
 			y = 10;
 		else if (fabs(x - lastX) < 3 && lastY <= (10 + 14 * 3))
 			y = lastY + 3;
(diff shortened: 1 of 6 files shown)


bstoeger:
Perhaps I should detail why this is a relevant change:

Currently, drag & dropping thumbnails on the profile plot is broken. The fundamental problem is that the time-offset will be adjusted, but the order of the images won't. Thus, calculating the picture positions on the profile will fail in funky ways (overlapping images, etc).

Moreover, removing images leaves the positions of the other images unchanged until dives are switched.

I fixed the above two points in #1432. But it turned out that keeping the lists in the profile-widget and the dive-photo-tab in sync became messy. Failure to keep the lists identical would lead to subtle bugs (see e.g. #1444).

Therefore, here the two lists are disconnected. I reimplemented #1432 based on this commit and the result is distinctly less convoluted. I can push the commits on top of this PR or do a separate PR.

In any case, I think this is a significant simplification, which will avoid bugs in the future.

profile-widget/profilewidget2.cpp
```diff
@@ -2044,70 +2045,75 @@ void ProfileWidget2::clearPictures()
...
-	for (int picNr = from.row(); picNr <= to.row(); ++picNr) {
-		int picItemNr = picNr - m->rowDDStart;
-		if (picItemNr < 0 || (size_t)picItemNr >= pictures.size())
-			return;
-		if (!pictures[picItemNr])
-			return;
-
-		pictures[picItemNr]->setPixmap(m->index(picNr, 0).data(Qt::UserRole).value<QPixmap>());
+	auto it = std::find_if(pictures.begin(), pictures.end(), [&filename](const PictureEntry &e)
+			       { return e.filename == filename; });
```
dirkhh: so this is a lambda based implementation that... finds a picture with that given filename?

profile-widget/profilewidget2.cpp
```diff
@@ -2044,70 +2045,75 @@ void ProfileWidget2::clearPictures()
...
+	if (it != pictures.end() && it->thumbnail) {
+		int size = Thumbnailer::defaultThumbnailSize();
+		it->thumbnail->setPixmap(QPixmap::fromImage(thumbnail.scaled(size, size, Qt::KeepAspectRatio)));
```
dirkhh: ...and if we find a thumbnail, we, err, replace it? Why?
This code could really benefit from some comments...

profile-widget/profilewidget2.cpp
```diff
@@ -2044,70 +2045,75 @@ void ProfileWidget2::clearPictures()
...
 	}
 }
 
+bool ProfileWidget2::PictureEntry::operator< (const PictureEntry &e) const
+{
+	return std::tie(offset.seconds, filename) < std::tie(e.offset.seconds, e.filename);
```
dirkhh: what's the semantic of comparing two std::tie ?

profile-widget/profilewidget2.cpp
```diff
@@ -2044,70 +2045,75 @@ void ProfileWidget2::clearPictures()
...
-		DivePictureItem *item = pictures[picItemNr].get();
-		if (!item) {
-			item = new DivePictureItem;
-			pictures[picItemNr].reset(item);
-			scene()->addItem(item);
-		}
-		item->setVisible(prefs.show_pictures_in_profile);
-		item->setPixmap(m->index(i, 0).data(Qt::UserRole).value<QPixmap>());
-		item->setFileUrl(m->index(i, 1).data().toString());
+		e.thumbnail.reset(new DivePictureItem);
```
dirkhh: what does this do?

profile-widget/profilewidget2.cpp
```diff
@@ -2044,70 +2045,75 @@ void ProfileWidget2::clearPictures()
...
+		scene()->addItem(e.thumbnail.get());
```
dirkhh: now we add the thumbnail... which we only set up in the next four lines of code???

profile-widget/profilewidget2.cpp
```diff
@@ -2044,70 +2045,75 @@ void ProfileWidget2::clearPictures()
...
-	first = std::max(0, first - m->rowDDStart);
-	// Note that last points *to* the last item and not *past* the last item,
-	// therefore we add 1 to achieve conventional C++ semantics.
-	last = std::min((int)pictures.size(), last + 1 - m->rowDDStart);
-	if (first >= (int)pictures.size() || last <= first)
-		return;
-	pictures.erase(pictures.begin() + first, pictures.begin() + last);
+	// Delete in ranges instead of single entries to avoid unnecessary copying. Premature optimization?
+	for (auto it1 = pictures.begin(); it1 != pictures.end(); ++it1) {
+		if (std::find(fileUrls.begin(), fileUrls.end(), it1->filename) == fileUrls.end())
```
dirkhh: how / why is this different from the find_if with lambda thingy above at line 2050? It seems to do the same thing? And if not, that too might be worth a comment

profile-widget/profilewidget2.h
```diff
@@ -227,8 +228,17 @@ public
...
-	std::vector<std::unique_ptr<DivePictureItem>> pictures;
+	// The list of pictures in this plot. The pictures are sorted by offset in seconds.
+	// For the same offset, sort by filename.
+	// Pictures that are outside of the dive time are not shown. In this case, thumbnail
+	// is a null-pointer.
+	struct PictureEntry {
+		offset_t offset;
+		QString filename;
+		std::unique_ptr<DivePictureItem> thumbnail;
+		bool operator< (const PictureEntry &e) const;
```
dirkhh: what does it mean to have an operator as part of a struct?

dirkhh:
What's the state of this? I fundamentally can't even read the code. I tried to comment and ask questions, but frankly, I feel rather lost and out of my league.

profile-widget/profilewidget2.h
neolit123: it allows two `PictureEntry` structs to be compared, in the lines of `pictureEntry1 < pictureEntry2`, based on the `bool ProfileWidget2::PictureEntry::operator< (const PictureEntry &e) const` operator implementation.

it isn't considered a bad practice to have operators and methods for structs in C++. the language allows that.

profile-widget/profilewidget2.cpp
neolit123: https://en.cppreference.com/w/cpp/utility/tuple/tie
> std::tie can be used to introduce lexicographical comparison to a struct or to unpack a tuple

it's the C++11 way of saying:
```
return offset.seconds < e.offet.seconds && filename < e.filename;
```

profile-widget/profilewidget2.cpp
neolit123: yes, it goes through `pictures` (contains `PictureEntry`) and returns iterator for the first match based on the lambda. it could be a regular unary function instead of a lambda.

profile-widget/profilewidget2.cpp
bstoeger: This is the "slot" which is asynchronously called by the thumbnailer when a thumbnail was calculated / fetched from disk. The situation is the following:

1) User switches dive.
2) For all images placeholders are shown and UI becomes responsive again.
3) Thumbnailer starts fetching / calculating thumbnails and pushes them one-by-one to the profile and the dive-photo tab.

This made switching dives distinctly snappier (already in previous release). Analogous code is in dive-photo tab, but it is now repeated here in the profile, since this PR is about disconnecting dive-photo tab and profile.

I will add a comment to the function header.

To be honest, I don't know why I used `find_if` here - it could be done by a simple loop just as well. I guess this is simply a habit of thinking more "declarative" than "imperative". No problem with changing to a loop, if you prefer.

profile-widget/profilewidget2.cpp
bstoeger: This is C++'s idiomatic way of doing the following:

```
{
	if (offset.seconds < e.offset.seconds)
		return true;
	if (offset.seconds > e.offset.seconds)
		return false;
	return filename < e.filename;
}
```

The big advantage over hand-coding is that it is much less error-prone (talking from experience here). It also works for an arbitrary number of arguments and compiles to the same code.

profile-widget/profilewidget2.cpp
neolit123: that on compares struct members and needs a special / separate unction (lambda) for the comparison.
this one compares QStrings directly as QString already has a `==` operator required by `std::find`.

profile-widget/profilewidget2.cpp
```diff
@@ -2044,70 +2045,75 @@ void ProfileWidget2::clearPictures()
...
-		item->setPos(x, y);
+		e.thumbnail->setPos(x, y);
 	}
 }
 
-void ProfileWidget2::removePictures(const QModelIndex &, int first, int last)
+// TODO: This does not check for the fact that the same image may be attributed
+// to different dives! Deleting the picture from one dive may therefore remove
+// it from the profile of a different dive.
+void ProfileWidget2::removePictures(const QVector<QString> &fileUrls)
```
neolit123: i think this TODO has to be fixed before we marge, so that we do not forget about it and a user sends us a bug report.

profile-widget/profilewidget2.cpp
bstoeger: This sets the `std::unique_ptr` to a newly allocated object.

I'm using `std::unique_ptr<DivePictureItem>` instead of `DivePictureItem *` here, because it makes it unnecessary to bother with memory management. Grow, shrink, clear the vector - it will take care that no memory is leaked.

I used `std::unique_ptr` instead of Qt's (subpar) version `QScopedPointer`, because the latter didn't work properly in a vector (no move semantics). But with the simplifications in this commit, perhaps `QScopedPointer` works again. I can give it a try.

profile-widget/profilewidget2.cpp
neolit123: i think the purpose of this method is to update the thumbnail of an image, but what happens if it doesn't have one.
is there are separate method to handle the `!it->thumbnail` case? i might have missed it.

profile-widget/profilewidget2.cpp
bstoeger: I can move the call down four lines if you wish. AFAIK it doesn't make any difference, since we're in the UI thread and the objects will only be painted once we're finished(?).

neolit123 (commented):
added some notes. @bstoeger could explain more in detail some of the questions.

i think that one TODO has to be fixed before we merge.

@bstoeger , might be a good idea to add slightly more detailed comments near C++11 code explaining what these do - like `find_if(), finds A in B by using the lambda`, `std::tie() is used here for reason X`.

janiversen:
The comments from @dirkh made me curious. I do not find it hard to read (when being used to C++), actually the code looked good. But I do find the lack of comments makes it a lot harder to read. I had to read parts twice, quessing first time what the code does, and then a second time to get an idea of why do that in this place. A comment in the code itself e.g. before loops makes it a lot simpler.

profile-widget/profilewidget2.cpp
bstoeger: Note that this find is inverted compared to the find above.

In the above function we got a filename and search for the picture with that filename.

Here we got a list of filenames and for every picture we search whether its filename is in the list.

This function is indeed complex. The reason is that it tries to avoid doing repeated copying of vector elements, by deleting the the images in batches. It will find batches of continuous images to be deleted an remove them at once.

profile-widget/profilewidget2.h
bstoeger: It means that the object can be compared with `<`. So you can write
```
	if (e1 < e2) {
		...
	}
```

Here, this is used so that we can sort the images with a default function. As an alternative, we could supply either a lambda or an explicit function to the `sort()` call.

profile-widget/profilewidget2.cpp
bstoeger: > i think the purpose of this method is to update the thumbnail of an image, but what happens if it doesn't have one.
is there are separate method to handle the !it->thumbnail case? i might have missed it.

In the case `!it->thumbnail`, the thumbnail is not on the profile (timestamp outside of dive), therefore there is no point in updating it.

profile-widget/profilewidget2.cpp
bstoeger: > `return offset.seconds < e.offet.seconds && filename < e.filename;`

Note that this code is not correct. If `offset.seconds < e.offset.seconds`, then the filenames will not be compared. Which just proves the point that `std:tie(...)` avoids errors. :)

dirkhh:
Answering from my phone.
My main concern is readability and understanding.
I'm not so much asking you to use different data types or structures. In most cases done comments would be enough.

profile-widget/profilewidget2.cpp
neolit123: yes, makes sense.
how would the above example change if we add a third parameter to the tie?

bstoeger:
> What's the state of this?
I fundamentally can't even read the code. I tried to comment and ask questions, but frankly, I feel rather lost and out of my league.

I must say that you get really proficient at reading C++/STL. :)

I will add comments at the objected places.

> i think that one TODO has to be fixed before we merge.

Note that the problem was there all the time. It is just the TODO that is added. I can try to fix it, but I fear that it will add some complexity.

neolit123:
> I will add comments at the objected places.

should be a general practice for all of us. if we add C++11 code to make sure to comment it, thinking "hey, they are some folks in this project who don't write much C++, so i better explain to them what this does".

EDIT: or even for confusing C - adding a comment is often times better than not adding a comment.

profile-widget/profilewidget2.cpp
bstoeger: Then
`return std::tie(a, b,c) < std::tie(e.a, e.b, e.c)` is equivalent to
```
	if (a < e.a) return true;
	if (a > e.a) return false;
	if (b < e.b) return true;
	if (b > e.b) return false;
	return c < e.c;
```
Not sure if you can write it more concisely without `tie`.

neolit123: i guess `lexicographical comparison` is the key here. `tie` does removes the potential for written errors.

bstoeger:
I added comments and simplified the code even more:

The whole "nullpointer for images outside of the profile" was needed for the old code, where we had to keep the lists in sync. Now, we can simply not keep track of these pictures.

Concerning the TODO: I fear this will be a bit more complicated, as we somehow have to keep track of the dive the deleted images belong to. But as I noted above: this never worked correctly and the whole failure case is a bit obscure: It means that the user added the same picture to different dives (why would you?), had both dives selected at the same time and deleted one of the pictures.

dirkhh:
The TODO documents an outstanding issue that has already been there, so that's acceptable, IMHO. Especially since it is such a rather bizarre corner case to begin with.
With the explanations in the thread and especially with the comments in the code this seems now much easier to understand. Thanks @bstoeger for your patience and updates.

PR #1451, opened 2018-07-01
Split and merge

torvalds:
This fixes dive splitting and merging.

There were two independent bugs in here: one stupid sample pointer arithmetic error (introduced back in 2015 by commit a39038a902: "In free dive mode split dives with only a 10 second surface interval"), and one situation where we incorrectly then considered the two (or more) parts of a dive that had been split to be the exact same dive, so we merged not the parts back into one longer dive, but tried to merge the parts back as if they were two copies of the same (short) dive.

This fixes the dive case that Alberto had with several surface intervals: the his dive now splits into four distinct segments, and you can merge them back again however you like.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -3562,7 +3562,7 @@ int split_dive(struct dive *dive)
 		// the surface start.
 		if (!surface_start)
 			continue;
-		if (!should_split(dc, dc->sample[surface_start].time.seconds, sample[i - 1].time.seconds))
+		if (!should_split(dc, dc->sample[surface_start].time.seconds, sample[-1].time.seconds))
 			continue;
 
 		return split_dive_at(dive, surface_start, i-1);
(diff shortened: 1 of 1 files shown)


bstoeger (approved):
Note that in the Signed-off-by line of the first commit you replaced an 'n' by an 'm'.

torvalds:
Heh. Yes, I misspelled my own email address in the sign-off of the first patch, and in the second patch the commit message has another spealing misstuck: "sive computer" instead of "dive computer".

I think the spelling mistakes can stand.

dirkhh:
Would you like me to fix your two typos when pulling this? Or would you like to fix them and then force push into your branch? I'd hate to see it merged like this, TBH

dirkhh:
I guess @torvalds is still ignoring GitHub notifications :-)
I simply fixed the commit messages and manually merged this.
I cnnot stant tupos.

PR #1454, opened 2018-07-02
Manual split dive

atdotde:

- [x] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This adds a profile widget context menu item to manually split a dive at a given time. Plus it makes sure the cylinder start and end pressures are adopted to a dive split as well (if there is sample data).

I am not 100% sure about what kind of redraw calls/emits are necessary, but without the MainWindow ones, it did not work for me.

### Related issues:

Related to #1451


core/dive.c
```diff
@@ -3570,6 +3582,22 @@ int split_dive(struct dive *dive)
...
+void split_dive_at_time(struct dive *dive, int time)
+{
+	int i = 0;
+	struct sample *sample = dive->dc.sample;
+
+	if (!dive)
+		return;
+	while(sample->time.seconds < time) {
+		++sample;
+		i++;
```
bstoeger: Not that it matters, but that is a curious prefix/postfix inconsistency.

core/dive.c
```diff
@@ -3570,6 +3582,22 @@ int split_dive(struct dive *dive)
...
+		if (dive->dc.samples == i)
+			return;
+	}
```
bstoeger: Any reason to check for `dive->dc.samples==i` but not `i==0`?
Given the call in the next line, `i <= 1` may be even better.

bstoeger (approved):
Apart from two very minor comments - lgtm.
(But didn't try to understand the cylinder thing.)

CHANGELOG.md
```diff
@@ -2,3 +2,5 @@
 # existing entries.
 # Use this layout for new entries:
 # <Area>: <Details about the change> [reference thread / issue]
+
+* Profile: Context menu entry to manually split a dive
```
neolit123: we should add entries on top of the file (above the note).

dirkhh: Yes, please. Even with that the merge sometimes fails strangely enough, but at least chances are better.

core/dive.c
dirkhh: I'd agree. Splitting at sample #0 or #1 simply makes no sense

dirkhh (changes requested):
I think the CHANGELOG and the test that this isn't "outside of the split zone" makes sense. Because just as the first couple samples make no sense, nor do the last couple. Both resulting dives reasonably need to have at least three samples, right?

core/dive.c
bstoeger: Also note that by clicking to the left of the actual profile one can generate timestamps *before* the dive started.

atdotde: Only checking against dc.samples to make sure loop does not run off the end of the sample array. I pushed an updated version that checks the split point is well into the dive (from both ends) in split_dive_at (which has sample numbers).

CHANGELOG.md
atdotde: Fixed in updated version.

dirkhh (approved):

dirkhh:
Thanks for the update. I need to stare at that last commit one more time... the logic for more than one dive computer isn't quite clear to me, yet.

dirkhh:
OK, I convinced myself that this should be correct. Merging manually because of CHANGELOG

dirkhh:
yeah, no, GitHub does that wrong. Oh well. I'll fix it on my system

dirkhh:
merged

Issue #359, opened 2017-04-28
[Idea] Add video file support


atdotde:
In addition to pictures we could show stills from videos taken during the dive. With the advent of cheap video cameras those will become more common.

Adding this need a bit research with the Qt video capabilities. This would be a nice medium size project.

dirkhh:
Go for it :-)

bstoeger:
I looked a little bit into this. Playing videos should not be too much of a problem with `QMediaPlayer`. For extracting thumbnails from videos I see two options and I don't like either:

1) _Use Qt_: There is a rather clumsy, but in principle not too hard to implement way. Play it with `QMediaPlayer` and stream it to a customized `QAbstractVideoSurface`, which stops after capturing a frame. Sounds good, but has a big caveat: At least on Linux, this has to be run in the main thread owing to an old Qt bug: https://bugreports.qt.io/browse/QTBUG-39567 . Thus, we'd have to move the movie fro the worker thread to the main thread. This would probably mean a refactoring of the thumbnailing.

2) _Use external library_: OpenCV comes to mind. With the whole annoyance of adding another dependency.

atdotde:
Have you managed to play a video in Subsurface?

bstoeger:
> Have you managed to play a video in Subsurface?

Yes, that's surprisingly trivial by using `QMediaPlayer` + `QVideoWidget`. A somewhat more complex way is subclassing `QAbstractVideoSurface` and projecting the frames by hand.

To automatically open the system video player instead of the image viewer, no change whatsoever is needed.

Side-note: There seems to be no reliable way of getting the supported video formats of the platform, so I fear we'll have to hard-code them.

bstoeger:
Sadly, whereas video-playback works flawleslly in `QVideoWidget`, it doesn't work with `QGraphicsVideoItem` for me, which would be needed to have it play in the profile-graph. It just flashes up for a fraction of a second and then vanishes. It plays the sound though. My next attempt would be subclassing `QAbstractVideoSurface` and projecting by hand onto the DivePictureItem. But that sounds unnecessarily complicated. :(

PR #1456, opened 2018-07-03
[Help needed] Play videos in profile plot

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Now that we don't hash full files anymore, adding videos works quite nicely. I can add a directory of ~100 videos in only a few seconds. Thumbnail extraction is not yet implemented - to this end I want to make video-playback work smoothly first. [Background: the idea is that for thumbnail extraction we play back the video and snap the first frame.]

But this is exactly where the problem lies: on video playback, memory consumption increases dramatically. I'm not sure what I'm doing wrong here and think that I will need the help of Qt experts. Qt's `QGraphicsVideoItem` did only show the first frame and then went all white (Kubuntu).

### Changes made:

1) Allow adding video files and recognize them as such.
2) Play videos in profile on mouse-over.
### Related issues:

#359

### Mentions:

@thiagomacieira, @tcanabrava: help?


bstoeger:
@dirkhh: I fear this will need some Travis magic. According to Qt's documentation, the classes seem to exist at least since Qt 5.4. But perhaps it will be easiest to compile video support only for a reasonably recent Qt?

neolit123:
> But this is exactly where the problem lies: on video playback, memory consumption increases dramatically. I'm not sure what I'm doing wrong here and think that I will need the help of Qt experts. Qt's QGraphicsVideoItem did only show the first frame and then went all white (Kubuntu).

`QGraphicsVideoItem` is the obvious choice given the parent widget, but have you tried and/or considered `QMediaPlayer` for the sake of the white screen debug or even using that if it causes less issues?

this has to be tested on all desktop OSes too.

> Play videos in profile on mouse-over.

i've been on the "implement me a video player" route so many times over the years and any framework imaginable. first the users will request that the video starts / stops by clicking, (because mouse overs are not good UX in general) and then they will request a full blow set of playback controls.
https://doc.qt.io/qt-5.9/images/mediaplayerex.jpg

bstoeger:
> QGraphicsVideoItem is the obvious choice given the parent widget, but have you tried and/or considered QMediaPlayer for the sake of the white screen debug or even using that if it causes less issues?

The current version does use `QMediaPlayer`. :)

> this has to be tested on all desktop OSes too.

IIRC, @sfuchs79 had reported some success on Windows.

> i've been on the "implement me a video player" route so many times over the years and any framework imaginable. first the users will request that the video starts / stops by clicking, (because mouse overs are not good UX in general) and then they will request a full blow set of playback controls.

It already works in the photo-tab, but I guess when they click in the profile, they likewise should get the system's video player. No need to implement a full fledged video-replay.

As you note, video-playback on mouse-over is not very important. The interesting part will be extracting a thumbnail from the video file via the `QMediaPlayer`.

dirkhh:
The iOS fail is my idiotic setup's fault. So I'll need to fix this on the back (basically speed things up I only upload a partial Qt for iOS installation - and the QAbstractVideoSurface isn't included. I'll fix that - not sure I'll get to it today.
The linux2 build fail is most likely also a missing package, but that can be addressed in the .travis.yml file, I think.
The Android failure confuses me, if we were missing the corresponding package, why isn't cmake complaining. This one is odd.
qt55 just like linux2 has missing packages that need to included in the apt-get arguments.
Does this help @bstoeger  ? I think at least linux2 and qt55 you should be able to fix yourself

neolit123:
> As you note, video-playback on mouse-over is not very important. The interesting part will be extracting a thumbnail from the video file via the QMediaPlayer.

could be tricky, indeed, if HW acceleration is ON:
https://stackoverflow.com/questions/30442759/rendering-a-qgraphicsscene-with-qgraphicsvideoitem-to-qimage

^ of course this is extraction of a thumbnail based on what is on the screen.

bstoeger:
> I think at least linux2 and qt55 you should be able to fix yourself

`linux2` was easy, but for `qt55` I don't even understand where the apt packages are listed...!?

I simply removed video support for mobile. Since they don't support dive pictures to begin with, it seems pointless to compile in video support - no?

> of course this is extraction of a thumbnail based on what is on the screen.

Extraction via `QImage` seems to work quite nicely (on my laptop), otherwise it wouldn't work at all. My main gripe is the exploding memory-usage. Not sure if I'm doing something wrong. Qt's ownership-model is horrible. :(

dirkhh:
I was wondering about mobile myself, but didn't know what you had in your longer term plans. Are you joining the developer call tomorrow? It would be nice to hear what your plans are.

For qt55 the packages are listed in `scripts/qt55/before_install.sh `

Memory consumption is of course a concern. While fewer people these days are using "netbook" style systems, I would still assume that you will run into older laptops with 1G of RAM.

bstoeger:
> Memory consumption is of course a concern. While fewer people these days are using "netbook" style systems, I would still assume that you will run into older laptops with 1G of RAM.

I have 4 GB and playback seems to work reasonably well. More testing, especially on older hardware, would of course be highly appreciated now that it compiles for all targets.

The big problem I see is when we start using `QMediaPlayer` to thumbnail hundreds of videos in a row. I don't really fancy either of
1) looking into the guts of Qt.
2) writing our own video codec to extract frames.
3) using an external frame-extraction tool.
4) looking into KDE's gwenview which thumbnails videos and is based on Qt.

So still hoping for @thiagomacieira to comment / help.

sfuchs79:
>>this has to be tested on all desktop OSes too.

>IIRC, @sfuchs79 had reported some success on Windows.

And I again after a quick test can confirm similar good results compared to the old tests from a few weeks ago. But once again there seems to be one issue under Windows:
Since last video tests I installed a completely fresh Windows 10. When I now tried the new implementation, at the first moment I had no playback but the error message "DirectShowPlayerService::doRender: Unresolved error code 0x80040266 ()".
Then I remembered that already last time I had to install a codec pack to make it work. Now after installing "K lite codec pack" I have playback in the profile again. I saw that there are some Qt bug reports and discussions explaining that issue.

>>i've been on the "implement me a video player" route so many times over the years and any framework imaginable. first the users will request that the video starts / stops by clicking, (because mouse overs are not good UX in general) and then they will request a full blow set of playback controls.

>It already works in the photo-tab, but I guess when they click in the profile, they likewise should get the system's video player. No need to implement a full fledged video-replay.

>As you note, video-playback on mouse-over is not very important. The interesting part will be extracting a thumbnail from the video file via the QMediaPlayer.

I have to say that I like the implementation as it is today. For both pictures and video we have a very simple internal viewer/player and if you want to see the picture full screen, zoom it or see the video full screen or go to a specific position you have to click it to open it in the external viewer of your choice.
So also from my side no need to add video control in Subsurface.

sfuchs79:
Regarding memory consumption:
I'm not sure if or how I could quantify this under Windows. When looking at the simple process view of windows task manager while running Subsurface I can see a obvious increase in memory use from ~240MB to above 550MB for some of my videos during playback. Yes, that is a huge increase. On the other hand my Firefox used 780MB at the same moment while doing almost "nothing" :-)

neolit123:
> For both pictures and video we have a very simple internal viewer/player and if you want to see the picture full screen, zoom it or see the video full screen or go to a specific position you have to click it to open it in the external viewer of your choice.

http://doc.qt.io/qt-5/qdesktopservices.html#openUrl is always the backup plan that doesn't require much effort from us about the implementation and the maintenance.
also removes the possibilities for us to handle things like missing codecs, Qt version bugs, OS specific bugs, OpenGL issues (if HW acceleration is ON) and things like that.

_If you want to play this video, Subsurface will open it for your in your default video player. Make sure that you have all the codecs installed for the format of this video file._

so it's like dodging a software maintainer bullet.

> So also from my side no need to add video control in Subsurface.

but if people start using the internal player, **they will** request controls and also if the internal player doesn't work for them **they will** report issues here at GitHub.

just trying to raise some concerns here. of course, the feature is nice to have.

bstoeger:
> Then I remembered that already last time I had to install a codec pack to make it work. Now after installing "K lite codec pack" I have playback in the profile again. I saw that there are some Qt bug reports and discussions explaining that issue.

Damn. Does that mean that we will have to ship codecs...? Sounds like opening a can of worms (patents, licenses, etc.). :(
Perhaps we can catch the error and direct the user to the appropriate download page..?

> http://doc.qt.io/qt-5/qdesktopservices.html#openUrl 

We use that already and it works. :)

> but if people start using the internal player, they will request controls and also if the internal player doesn't work for them they will report issues here at GitHub.

Sure, but with "just click on the thing", I think we have a pretty good argument.

So my conclusion so far is: Let's try to land recognition and playback first and bother with thumbnailing in due course.

neolit123:
> Damn. Does that mean that we will have to ship codecs...? Sounds like opening a can of worms (patents, licenses, etc.). :(
Perhaps we can catch the error and direct the user to the appropriate download page..?

thread of interest:
https://forum.qt.io/topic/63110/list-of-video-formats-qt-supports/14

> In summary, Qt does not support any video formats by itself, which is why there is no documentation for "list of formats supported". Instead, Qt supports different multimedia backends.

and if a backend doesn't have an installed codec it won't work for users.

on Windows, we can point towards things like "K lite codec pack" (like @sfuchs79 mentioned) and ffdshow. from there the users are on their own. i haven't played video on Linux and OSX.

both mentioned solutions are GPL2, so we can potentially ship them, but we should rather just instruct the users in the docs and on the UI as these codec packs can be rather big.

dirkhh:
I was rather un-enthusiastic about videos from the start. This makes me a lot more unenthusiastic. Two main reasons: a) this is a rabbit hole (Fass ohne Boden) b) most of the hard work I will be stuck with

neolit123:
i propose that we don't support video playback in subsurface and also not generate thumbnails (well, because that would be difficult without codecs).
please, mind that this is just a vote.

quoting myself from earlier:
> so it's like dodging a software maintainer bullet.

the current solution is good enough IMHO -  list the video files attached to a dive with a generic video file icon  + filename and if the user clicks that just do http://doc.qt.io/qt-5/qdesktopservices.html#openUrl.
and possibly add notes in the docs for the above, too.

bstoeger:
> I was rather un-enthusiastic about videos from the start. This makes me a lot more unenthusiastic. Two main reasons: a) this is a rabbit hole (Fass ohne Boden) b) most of the hard work I will be stuck with

You sound like a burnt child. :) I believe / hope that the situation is not that bleak.

In a first step we can only implement importing of videos without playback / thumbnailing (commits 2 and 3 of this PR). New PR incoming.

In a second step, I would simply document in the manual that playback works on Windows only if users install codecs by themselves. Unruly users can then be pointed in the right direction.

For the future, if Qt's multimedia is really in such a bad shape as it looks like right now, I will look into libavcodec. It's GPL2 and supports all major platforms. With this library we can also snap images from inside the video, etc. I have 200 GoPro videos from my last trip and really would like to have thumbnails. Scrolling through pages of dummy icons is not useful.

bstoeger:
> thread of interest:
https://forum.qt.io/topic/63110/list-of-video-formats-qt-supports/14

What this doesn't explain: Windows obviously ships with codecs for its media-player. Why can't Qt use those? Am I naive in thinking that this should work out-of-the-box?

neolit123:
> What this doesn't explain: Windows obviously ships with codecs for its media-player. Why can't Qt use those? Am I naive in thinking that this should work out-of-the-box?

it does explain that Qt uses the Windows DirectShow backend. it has a "starter pack" level of coverage for codecs.

supporting all format out there is impossible. there are hundreds of formats for everything, audio, video, splitters, containers you name it. you buy a new camera from Vendor X and it implements a new amazing format that no longer works on your old Windows 7. i was happy that H.264 became popular, because it centralized the media playback ecosystem, a tiny bit, that is.

the mentioned codec packs have quite good coverage and we need to promote them with our video playback feature; if we add the feature.

but then, how about the other OSes? i see that they need codec packs too, which means that we need to add instructions for all OSes.

if you bring this up on the meeting today we can talk more about it.

dirkhh:
> I was rather un-enthusiastic about videos from the start. This makes me a lot more unenthusiastic. Two main reasons: a) this is a rabbit hole (Fass ohne Boden) b) most of the hard work I will be stuck with
> 
> You sound like a burnt child. :) I believe / hope that the situation is not that bleak.
> 

I sound like a burnt child because I AM a burnt child. I've been doing these releases for half a dozen years now...

> In a first step we can only implement importing of videos without playback / thumbnailing (commits 2 and 3 of this PR). New PR incoming.
> 
> In a second step, I would simply document in the manual that playback works on Windows only if users install codecs by themselves. Unruly users can then be pointed in the right direction.
> 

That seems... something worse thinking about.

> For the future, if Qt's multimedia is really in such a bad shape as it looks like right now, I will look into libavcodec. It's GPL2 and supports all major platforms. With this library we can also snap images from inside the video, etc. I have 200 GoPro videos from my last trip and really would like to have thumbnails. Scrolling through pages of dummy icons is not useful.
> 

As I said on the other PR - I am not set against this. I am worried. If we have a good way to do this sanely, I will be easily (ha) convinced to agree :-)

bstoeger:
So it turns out QMultimedia is not as helpful as I would have hoped to. Let's try ffmpeg/libavcodec next. :)

PR #1459, opened 2018-07-04
Core: add all qPref macros and implement qPrefDisplay as example

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR extend qPref_private with macros for:
The whole get function (when the get function is standard)
The whole set function (when the set function is standard)
The whole get/set/loadsync functions (in case the variable is all standard)

LOADSYNC is updated to use a private QSetting variable instead of making a new instance in 
Every call.

The first commit shows qPrefDisplay using the low level LOADSYNC macros, this will only be used
Where the load/sync does not follow a standard pattern.

The last commit shows qPrefDisplay using the higher level LOADSYNC macros, depending on how
Standarzied the variable is.

REMARK: qPrefDisplay is NOT connected to SettingsObjectWrapper and thus not part of an executable. This is on purpose for 2 reasons:
- We can freely discuss and update the use pattern, without side effects.
- There are no test cases, which needs to be part of every class conversion.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -100,6 +100,7 @@ set(SUBSURFACE_CORE_LIB_SRCS
 
 	# classes to manage struct preferences for QWidget and QML
 	settings/qPref.cpp
+	settings/qPrefDisplay.cpp
 
 	#Subsurface Qt have the Subsurface structs QObjectified for easy access via QML.
 	subsurface-qt/DiveObjectHelper.cpp
--- a/core/pref.h
+++ b/core/pref.h
@@ -95,7 +95,7 @@ struct preferences {
 	bool        display_invalid_dives;
 	const char *divelist_font;
 	double      font_size;
-	bool        showDeveloper;
+	bool        show_developer;
  	const char *theme;
 
 	// ********** Facebook **********
--- a/core/settings/qPref.h
+++ b/core/settings/qPref.h
@@ -3,7 +3,10 @@
 #define QPREF_H
 
 #include <QObject>
+#include <QSettings>
 #include "core/pref.h"
+
+#include "qPrefDisplay.h"
 
 class qPref : public QObject {
 	Q_OBJECT
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "qPref.h"
+#include "qPref_private.h"
+#include "core/subsurface-string.h"
+
+#include <QApplication>
+#include <QFont>
+
+qPrefDisplay::qPrefDisplay(QObject *parent) : QObject(parent)
+{
+}
+qPrefDisplay *qPrefDisplay::instance()
+{
+	static qPrefDisplay *self = new qPrefDisplay;
+	return self;
+}
+
+void qPrefDisplay::loadSync(bool doSync)
+{
+	disk_divelist_font(doSync);
+	disk_font_size(doSync);
+	disk_display_invalid_dives(doSync);
+	disk_show_developer(doSync);
+	disk_theme(doSync);
+}
+
+const QString qPrefDisplay::divelist_font() const
+{
+	return prefs.divelist_font;
+}
+void qPrefDisplay::set_divelist_font(const QString& value)
+{
+	QString newValue = value;
+	if (value.contains(","))
+		newValue = value.left(value.indexOf(","));
+
+	if (newValue != prefs.divelist_font &&
+		!subsurface_ignore_font(qPrintable(newValue))) {
+		COPY_TXT(divelist_font, value);
+		qApp->setFont(QFont(newValue));
+		disk_divelist_font(true);
+		emit divelist_font_changed(value);
+	}
+}
+void qPrefDisplay::disk_divelist_font(bool doSync)
+{
+	LOADSYNC_TXT("/divelist_font", divelist_font);
+}
+
+double qPrefDisplay::font_size() const
+{
+	return prefs.font_size;
+}
+void qPrefDisplay::set_font_size(double value)
+{
+	if (value != prefs.font_size) {
+		prefs.font_size = value;
+		QFont defaultFont = qApp->font();
+		defaultFont.setPointSizeF(prefs.font_size);
+		qApp->setFont(defaultFont);
+		disk_font_size(true);
+		emit font_size_changed(value);
+	}
+}
+void qPrefDisplay::disk_font_size(bool doSync)
+{
+	LOADSYNC_DOUBLE("/font_size", font_size);
+}
+
+bool qPrefDisplay::display_invalid_dives() const
+{
+	return prefs.display_invalid_dives;
+}
+void qPrefDisplay::set_display_invalid_dives(bool value)
+{
+	if (value != prefs.display_invalid_dives) {
+		prefs.display_invalid_dives = value;
+		disk_display_invalid_dives(true);
+		emit display_invalid_dives_changed(value);
+	}
+}
+void qPrefDisplay::disk_display_invalid_dives(bool doSync)
+{
+	LOADSYNC_BOOL("/displayinvalid", display_invalid_dives);
+}
+
+bool qPrefDisplay::show_developer() const
+{
+	return prefs.show_developer;
+}
+void qPrefDisplay::set_show_developer(bool value)
+{
+	if (value != prefs.show_developer) {
+		prefs.show_developer = value;
+		disk_show_developer(true);
+		emit disk_show_developer(value);
+	}
+}
+void qPrefDisplay::disk_show_developer(bool doSync)
+{
+	LOADSYNC_BOOL("/showDeveloper", show_developer);
+}
+
+const QString qPrefDisplay::theme() const
+{
+	return prefs.theme;
+}
+void qPrefDisplay::set_theme(const QString& value)
+{
+	if (value != prefs.theme) {
+		COPY_TXT(theme, value);
+		disk_theme(true);
+		emit theme_changed(value);
+	}
+}
+void qPrefDisplay::disk_theme(bool doSync)
+{
+	LOADSYNC_TXT("/theme", theme);
+}
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -24,10 +24,7 @@ void qPrefDisplay::loadSync(bool doSync)
 	disk_theme(doSync);
 }
 
-const QString qPrefDisplay::divelist_font() const
-{
-	return prefs.divelist_font;
-}
+GET_PREFERENCE_TXT(Display, divelist_font);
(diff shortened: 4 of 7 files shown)


dirkhh:
The iOS Travis error is surprising - is there something in the moc on iOS that's different and that causes us to miss the right `#include` ?

dirkhh (commented):
There are a few things that I don't understand. Please help me out.

- the purpose of the PR is to review the design pattern, not to actually be merged, correct?
 -- if yes, then the title should reflect this
 -- if no, the 2nd (or was it 3rd?) commit has the SOB-message in the middle without an empty line in front, that should be addressed

- can you explain the naming of the macros? Why is everything called _LOADSYNC_ ? I'm sure that this conveys some meaning that I'm just missing.

- the reasons for the differences in names between settings and struct members are mostly stupid, but once thing to pay VERY close attention to is that we CANNOT break existing preference files - which means the names on disk cannot change. I haven't verified that your code doesn't break this (because I wasn't 100% sure what the purpose of this PR is).

- as you said in your PR text, yes, tests would be really important here.

janiversen:
Hmm I have seen that error before (QSettings unknown), on my own system it was a cache problem, let me see if I can do anything.

janiversen:
Funny iOS compiles ok on my remote mac, I tried to solve your SOB message (not too sure what you mean) and pushed a new version.

dirkhh:
> Hmm I have seen that error before (QSettings unknown), on my own system it was a cache problem, let me see if I can do anything.

On Travis things are built from scratch, so it shouldn't be a cash issue, I would think...

> I tried to solve your SOB message (not too sure what you mean) 

I meant that your "Signed-off-by: ..." line wasn't where it usually is: at the end of the commit message, separated by an empty line.

janiversen:
OK I understand the SOB problem, I went through the commits and removed the last empty line, it is something the git adds (maybe it is just my configuration).

dirkhh:
> The idea is to merge this PR, but it does NOT integrate into SettingsObjectWrapper, so it will not affect the live system. QPref_private.h is needed for my future work (once we agree on it), and qPrefDisplay is production ready.
> 

I'm not opposed to that, putting things in place for future work is something several of us have done at times.

> To be honest the “old” macros was called LOADSYNC, because they only did that, and I just continued to expand on that name. Any better name suggestion ? (It is easy for me to change).
> 

I love it when macros have names that really reflect what they do. So ask yourself "what's the best description of the effect / side effect of this macro". Does it get a value? set a value? set a value and persist it (i.e., write to disk)?
Both "load" and "sync" are slightly odd verbs when it comes to values. Does "load" imply that it's always loaded from persistent storage? "sync" implies that there are multiple potentially different values and they are brought to a common value - but it doesn't tell us which one is picked.

> I pay VERY close attention not to change names on disk (that is the LOADSYNC part), because that would make users angry. Everywhere you see “/” in qPrefDisplay.cpp, that is the name taken from SettingsObjectWrapper load and/or sync. The renaming is purely in code !
> 

Thanks

> This PR will not break anything, but we need to agree on the format, e.g. can my set_ and _changed be accepted to unify things. Once that is done and merged, then I will make a PR for each class that contains:
> 
> the class using the macros
> the effect on the rest of the system (name changing)
> test cases (testpreferences will be spilt into multiple files)
> And that is when I hope you and other will actively test the PR to find problems

I hope that @bstoeger and @tcanabrava will also comment on this. Right now I struggle with the names and they distract me. I can see how you change things, I cannot quite tell what the end goal is going to be and how that will make everyone happy (as in: all users of the code get what they need and the semantics are easy to understand and to follow)

dirkhh:
>  OK I understand the SOB problem, I went through the commits and removed the last empty line, it is something the git adds (maybe it is just my configuration).

An empty line at the end is fine (that gets ignored). Your commit message looked like this
```
text explaining things - at least a paragraph of it then no empty line and
Signed-off-by: jan Iversen <...>

more text explaining things after an empty line
more text
```

dirkhh:
Ahh, it's still there. Look at https://github.com/Subsurface-divelog/subsurface/pull/1459/commits/d47df9e734609b3bb319e74d39242bcc5be2711b

dirkhh:
> Ups, I will refer to the danish saying “sometimes you cannot see the wood for all the trees” it is corrected.
> 

Man sieht den Wald vor lauter Bäumen nicht 

> LOADSYNC, described with the low level macros did (LOAD from disk, SYNC to disk),
> 

note how you need the preposition ‘to’ in order to realize that you are syncing from memory to disk and not the other way - which is why sync is a problematic verb here.

> but the others others will be renamed to SET_PREFERENCE* GET_PREFERENCE* HANDLE_PREFERENCE* which tells a bit more what it does. I think like you and prefer names that are understandable.
> 

PREF might be a good abbreviation to make this easier to type :-)

In the meantime I’m going nuts with (a) an iPad set to German which is just so annoying and (b) one Mac that appears to create perfectly fine iOS images, but because it’s a VM it can’t directly talk to an iPad, and another Mac that can talk to an iPad, but creates completely broken iOS builds that give crazy QML errors and none of the buttons work...

janiversen:
Please remember these macros are only used in the qPref files not in the rest of the system, there you use e.g. qPrefDisplay::instance()->theme() 

Depending on the type of vm there might be a solution, vmware used to a remove usb tool, giving you the possibility to attach usb devices from where you have the screen running (not possible with ssh).

My mac made crazy iOS images about a month ago, because Xcode was updated but Xcode command line tools was not updated.

dirkhh:
OK, so the broken images on the iPad were caused by your pref changes. At least some of the weird errors in the logfile seem to point in that direction...
```
QMetaProperty::read: Unable to handle unregistered datatype 'cloud_status' for property 'QMLPrefs::credentialStatus'
qrc:///qml/DiveList.qml:16:33: Unable to assign [undefined] to int
QMetaProperty::read: Unable to handle unregistered datatype 'cloud_status' for property 'QMLPrefs::credentialStatus'
qrc:///qml/Settings.qml:66:11: Unable to assign [undefined] to QString
QMetaProperty::read: Unable to handle unregistered datatype 'cloud_status' for property 'QMLPrefs::credentialStatus'
QMetaProperty::read: Unable to handle unregistered datatype 'cloud_status' for property 'QMLPrefs::credentialStatus'
QMetaProperty::read: Unable to handle unregistered datatype 'cloud_status' for property 'QMLPrefs::credentialStatus'
QMetaProperty::read: Unable to handle unregistered datatype 'cloud_status' for property 'QMLPrefs::credentialStatus'
```
This appears to prevent the No Cloud button from working.
So when creating the updated iOS binary you will get the release branch that you asked for after all :-)

janiversen:
I knew it, remember when we discussed only having one version of the enum cloud_status, and I got “pursuaded” not to have it in qmlPrefs, that is the cause. I will have a patch in some hours.

janiversen:
Try and cherry-pick #1462 that should solve the problem. However I really do not understand why this works for me, but not for you (I will investigate that later, once the release is fixed).

bstoeger:
> I knew it, remember when we discussed only having one version of the enum cloud_status, and I got “pursuaded” not to have it in qmlPrefs, that is the cause. I will have a patch in some hours.

Two comments:
1) I know that I am too brief. What I obviously meant was that ultimately the *variables* should be removed from the preferences, not the enum.
2) This whole thing demonstrates why I dislike QML. I have the impression that Qt tried to combine the disadvantages of all programming languages. Compile times of C++, type safety of scripting languages, elegance of 1970ies macro assembler, XML, plus a dose of NIH syndrome. :(

janiversen:
@bstoeger regarding your 1), this time you really was too brief, I 110% agree to remove the variables, as I have written several time the QML interface is “ackward” and the only way to counter it a bit is to use interface classes, so you will see all qPref classes will not have variables. The current qmlmanager and qmlprefs are a problem waiting to happen, because not only do they have variables, but these are copies of struct preferences.

Regarding your 2) again I totally agree with you, but QML is a major advantage compared to the early Qt version, where 7-80% of your code was just stupid UI handling. Tdoay we have QML, and I believe we can make a secure interconnection between C++ and QML, SettingsObjectWrapper has a lot of good intentions in it that simply was not completed (sorry if I offend anyone, but it starts by having a private constructor, which imidialtly says no qml).

I am trying with qPref, to achieve 2 goals:
- make a very stable interface to qml
- make a qt top of struct preferences, that can be used in standard Qt UI code

janiversen:
An interface class does not have variables but refers to them (in this case to the global pref struct), this means the class can be instanciated n times without problems.

QML likes to instanciate a new object e.g. with every window that uses the class, I have had severe problems with qmlmanager in experimental code (like dynamic windows), because qmlmanager have local variables and if instanciated multiple times, these variables have different states.

I am all for talking about solutions with @tcanabrava and others (I believe @bstoeger have a lot of good opinions as well). I do not pocess the one and only solution, but I do think that what I propose is a major stetp forward.

dirkhh:
Thank you for that explanations. Now I understand what you are saying. And yes, making this a true interface class sounds like the right direction to go.

dirkhh:
@bstoeger @tcanabrava @janiversen - is this ready to merge, does this need more re-working?

janiversen:
As the author I am happy with it, but will do re-work as wanted.

bstoeger:
Sorry - I'm brain-busted from my day-job. On a quick skim I didn't see anything obviously wrong (sure, templates could probably make this more terse - though not a priority now). There's just one thing that threw me off: Do we really want to use a reserved keyword (`class`) as a macro parameter? If nothing else, this will confuse syntax-highlighting, no?

dirkhh:
> Do we really want to use a reserved keyword (class) as a macro parameter? If nothing else, this will confuse syntax-highlighting, no?

How funny that I didn't even catch that. Oops. Yeah, maybe we could use a better name? :-)

janiversen:
Yes of course I did a real bad choice of name (just because it actually is a class), will push a change shortly.

dirkhh:
@janiversen what's the relationship of this one and #1465 - are the independent? Related? Should one go in before the other?
I think this one looks good, but want to make sure I understand your intentions :-)

janiversen:
Let me try to explain.

This PR is about rebuilding the preference system for Qt and QML, the PR was the first step to show the macros and how they are used.

#1465 is the production version of the patch you got for the release. For the release you got a patch that touched minimally in the system, with the consequence that CLOUD_STATUS exist 2 times (again). #1465 defines CLOUD_STATUS in the class qPref (which is needed to get the qml interface working), removes CLOUD_STATUS from pref.h and changes the places where it was used.

There are no “one before the other”, but once one of them is merged, there will most likely be a merge conflict with the other, which I will then solve.

janiversen:
Is this PR pending on something I can solve ?

Would be real nice to be able to continue with this work (and reduce some of current rebasing between my branches :-)

dirkhh:
Nope, not pending. Just waiting to make sure I'm doing what was expected - and then forgot.
Merging now.

PR #1462, opened 2018-07-05
mobile-widgets: solve cloudstatus register problem

janiversen:
Use Q_ENUM instad of Q_ENUMS (which is depreciated) since it does the
meta registration for all Qt platforms.

Q_ENUM require the enum to be defined in the class and cannot refer to
a global class, therefore copied enum to class.

This commit is made to get the release to work, with minimal changes,
this class will be moved to qPref and the double definition solved

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Using Q_ENUMS works even though depreciated on some MAC development setups, however not all. As a temporary fix (for the release) the enum is again added to QMLPref and Q_ENUM is used.

### Mentions:

@dirkhh I do not know why it works (I run with the old code active on my ipad) when building on my mac, and not when you build, but this should solve the problem. 

Remark: This patch is currently NOT tested on iOS, as I work remote and cannot upload to my iPad.

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -269,12 +269,12 @@ void QMLManager::openLocalThenRemote(QString url)
 		 *    no cloud repo solves this.
 		 */
 
-		if (QMLPrefs::instance()->credentialStatus() != CS_NOCLOUD)
-			QMLPrefs::instance()->setCredentialStatus(CS_NEED_TO_VERIFY);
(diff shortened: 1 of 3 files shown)


dirkhh:
I will test this today.
The thing that seems truly broken in master right now is the "No cloud" mode button in the credential settings. Which is one of the first things I always test when installing a test build :-)
I haven't tried with this patch, yet. I should have time over lunch to do so.

dirkhh:
This appears to have fixed the issue for me. I just pushed a new beta to TestFlight

PR #1463, opened 2018-07-05
[RFC] Video-lite

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Because video-playback in #1456 was received with reservations, here is the lite-version: Import videos, but only show a dummy icon. On clicking the icon, the video is shown in the system viewer.

I hope we agree that nowadays this is a must-have feature.

### Changes made:

1) Import dummy video icon.
2) Add list of video extensions and recognize video files.
3) Save to thumbnail file with an appropriate flag signaling "video".

### DIFF
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -69,27 +69,40 @@ void ImageDownloader::saveImage(QNetworkReply *reply)
 	reply->deleteLater();
 }
 
-// Fetch a picture from the given filename. If this is a non-remote filename, fetch it from disk.
-// Remote files are fetched from the net in a background thread. In such a case, the output-flag
-// "stillLoading" is set to true.
+static bool isVideoFile(const QString &filename)
+{
+	// Currently, we're very crude. Simply check if the file exists and if it
+	// is an MP4-style format.
+	QFileInfo fi(filename);
+	if (!fi.exists() && !fi.isFile())
+		return false;
+	metadata md;
+	return get_metadata(qPrintable(filename), &md) == MEDIATYPE_VIDEO;
+}
+
+// Fetch a picture from the given filename and determine its type (picture of video).
+// If this is a non-remote file, fetch it from disk. Remote files are fetched from the
+// net in a background thread. In such a case, the output-type is set to MEDIATYPE_STILL_LOADING.
 // If the input-flag "tryDownload" is set to false, no download attempt is made. This is to
 // prevent infinite loops, where failed image downloads would be repeated ad infinitum.
-// Returns: fetched image, stillLoading flag
-static std::pair<QImage, bool> fetchImage(const QString &filename, const QString &originalFilename, bool tryDownload)
+// Returns: fetched image, type
+Thumbnailer::Thumbnail Thumbnailer::fetchImage(const QString &filename, const QString &originalFilename, bool tryDownload)
 {
-	QImage thumb;
-	bool stillLoading = false;
 	QUrl url = QUrl::fromUserInput(filename);
 	if (url.isLocalFile()) {
-		thumb.load(url.toLocalFile());
+		QString filename = url.toLocalFile();
+		if (isVideoFile(filename))
+			return { videoImage, MEDIATYPE_VIDEO };
+		QImage thumb(filename);
+		return { thumb, thumb.isNull() ? MEDIATYPE_IO_ERROR : MEDIATYPE_PICTURE };
 	} else if (tryDownload) {
 		// This has to be done in UI main thread, because QNetworkManager refuses
 		// to treat requests from other threads. invokeMethod() is Qt's way of calling a
 		// function in a different thread, namely the thread the called object is associated to.
 		QMetaObject::invokeMethod(ImageDownloader::instance(), "load", Qt::AutoConnection, Q_ARG(QUrl, url), Q_ARG(QString, originalFilename));
-		stillLoading = true;
+		return { QImage(), MEDIATYPE_STILL_LOADING };
 	}
-	return { thumb, stillLoading };
+	return { QImage(), MEDIATYPE_IO_ERROR };
 }
 
 // Fetch a picture based on its original filename. If there is a translated filename (obtained either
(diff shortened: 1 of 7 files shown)


neolit123:
@bstoeger 

BTW, we can only include icons and media that comes with a certain license or permissive origin that won't cause legal troubles. @dirkhh can elaborate.

one clean options is that we draw it our selfs.

bstoeger:
> BTW, we can only include icons and media that comes with a certain license or permissive origin that won't cause legal troubles. @dirkhh can elaborate.

Sure, that's why I took an icon of the LPGLed Breeze KDE theme. We already use them for mobile.

> one clean options is that we draw it our selfs.

Definitely not an option if I am supposed to draw it.

dirkhh:
Definitely a good start. I haven't stared at the code long enough, but conceptionally I like this much better.  For the record, I'm not saying NO to video playback, I am voicing concerns. I am still open to being convinced that this is manageable from a packaging perspective.

core/qthelper.cpp
```diff
@@ -1209,11 +1209,19 @@ QString localFilePath(const QString &originalFilename)
 	return localFilenameOf.value(originalFilename, originalFilename);
 }
 
+// TODO: Apparently Qt has no simple way of listing the supported video
+// codecs? Do we have to query them by hand using QMediaPlayer::hasSupport()?
+const QStringList videoExtensionsList = {
+	".avi", ".mp4", ".mpeg", ".mpg", ".wmv"
+};
```
neolit123, line 1216: the list of extensions looks fine to me, as is.

re: "dodging the bullet"
the point is that we shouldn't care about codecs.

obsolete member to list the MIMEs:
http://doc.qt.io/qt-5/qmediaplayer-obsolete.html
support is only an estimate:
http://doc.qt.io/qt-5/qmultimedia.html#SupportEstimate-enum

core/imagedownloader.cpp
```diff
@@ -163,25 +177,30 @@ static QImage getThumbnailFromCache(const QString &picture_filename)
...
 	// Each thumbnail file is composed of a media-type and an image file.
-	// Currently, the type is ignored. This will be used to mark videos.
 	quint32 type;
 	QImage res;
 	stream >> type;
 	stream >> res;
-	return res;
+
+	// Thumbnails of videos currently not supported - replace by dummy
+	// TODO: Correctly extract thumbnails
```
neolit123: should we really keep this TODO?

core/imagedownloader.cpp
```diff
@@ -69,27 +69,40 @@ void ImageDownloader::saveImage(QNetworkReply *reply)
...
-// "stillLoading" is set to true.
+static bool isVideoFile(const QString &filename)
+{
+	// Currently, we're very crude. Simply check if the file exists and if it
+	// is an MP4-style format.
+	QFileInfo fi(filename);
+	if (!fi.exists() && !fi.isFile())
+		return false;
+	metadata md;
+	return get_metadata(qPrintable(filename), &md) == MEDIATYPE_VIDEO;
```
neolit123, line 80: not sure if this would work for all containers out there.
might as well just assume by extension and throw the file to the external player.

dirkhh: If you look at https://github.com/Subsurface-divelog/subsurface/blob/master/core/metadata.cpp#L208 we try to parse the MP4 container - so this seems like a reasonable thing to do for us? This at least doesn't blindly accept any extension.
(Of course I worry about the security implications here... reading that code it should be fairly easy to fool us and to at least crash Subsurface if not worse)

core/imagedownloader.cpp
```diff
@@ -98,25 +111,25 @@ static std::pair<QImage, bool> fetchImage(const QString &filename, const QString
 // was downloaded previously, but for some reason the cached picture was lost. Therefore, in such a
 // case, try the canonical filename. If that likewise fails, give up. For input and output parameters
 // see fetchImage() above.
-static std::pair<QImage, bool> getHashedImage(const QString &filename, bool tryDownload)
+std::pair<QImage, mediatype_t> Thumbnailer::getHashedImage(const QString &filename, bool tryDownload)
 {
 	QImage thumb;
-	bool stillLoading = false;
+	mediatype_t type = MEDIATYPE_UNKNOWN;
```
dirkhh: we set type here to UNKNOWN and a few lines later we test if it is != STILL LOADING. What am I missing?

dirkhh: never mind. std::tie confused me

dirkhh:
It is so embarrassing how much this is already pushing my comfort zone reading code.
I agree with @neolit123 's comment about the TODO - other than this I couldn't find anything wrong, but given that this is clearly not easy to read for me, I'm also not super comfortable with my ability to successfully review this. So I didn't actually approve the PR :-/

core/imagedownloader.cpp
bstoeger: @neolit123 is in principle right - it will parse MP4s, but not AVI. So if that fails, we can still go after extension.

But: even if we fail to identify a video as such, it will simply show a broken icon (we could refine to a format unknown icon). Clicking on it will still open the system video player.

> Of course I worry about the security implications here... reading that code it should be fairly easy to fool us and to at least crash Subsurface if not worse)

I tried to make it at most a denial-of-service, which you can do in various ways anyway. If you think this is unsafe, I'll give it a second look in the next few days.

core/imagedownloader.cpp
bstoeger: As I said in a previous PR:
We can replace `std::pair<QImage, mediatype_t>` by
```
struct SomeSensibleClassName {
	QImage thumbnail;
	mediatype_t type;
};
```

Then the tie will become:
```
	SomeSensibleClassName some_sensible_variable_name = fetchImage(localFilename, filename, tryDownload);
```

So far I was simply too lazy.

core/imagedownloader.cpp
bstoeger: Maybe tone it down to
// TODO: Possibly extract thumbnail
?

core/imagedownloader.cpp
dirkhh: Yeah, that would have more matched my experience.
I keep thinking this is just a sign that I'm not really a C++ developer. But realistically, the struct way of doing this does allow you to call a function like this from C as well (assuming it's linked with C bindings)

core/imagedownloader.cpp
dirkhh: No, I didn't see anything that glaringly unsafe. This was more an idle comment...

core/imagedownloader.cpp
bstoeger: > But realistically, the struct way of doing this does allow you to call a function like this from C as well (assuming it's linked with C bindings)

Except that C will not be happy about the `QImage`... :(

If you prefer the `struct`-version, that's fine by me. The hardest part will be coming up with a sensible name. Perhaps simply `struct Thumbnail`?

bstoeger: So here's the struct version.

sfuchs79:
@bstoeger 
Great that we have the video support in master now! I'll try to continue testing this during the next days.

May I make a few suggestion (again) for fine tuning of the UI:
- Now that we support pictures and video let's rename every occurrence of "pictures" e.g. the tab name, the context menu string,... to s.th. else. I once already proposed "media" but I'm open to any other idea.
- For the file open dialog for the file filter I think it could be nice to have minimum three options: All pictures plus video files, picture files only, video files only.

I can submit a PR for this in a few days or maybe you like to do it but let's first discuss how we want to change it.

BTW: I also would not give up video playback in the profile immediately. I want to try once to recompile my Qt/MXE with the "-wmf-backend" option to see if this leads to s.th.

bstoeger:
> I can submit a PR for this in a few days or maybe you like to do it but let's first discuss how we want to change it.

As you wish. I think the discussion would be ideally conducted at the mailing list, not in a closed PR. My personal opinion: "Media" is fine.

> BTW: I also would not give up video playback in the profile immediately. I want to try once to recompile my Qt/MXE with the "-wmf-backend" option to see if this leads to s.th.

I opened a new PR with the video-playback. Perhaps this helps you with these experiments.

PR #1466, opened 2018-07-05
ssrf: add "make check" to central makefile

janiversen:
extend ssrf/makefile with "make check" option.

This is just a convinience function, to not need to "cd build" first

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

Allow “make check” at subsurface level, just like “make desktop” and “make mobile”

### DIFF
--- a/makefile
+++ b/makefile
@@ -11,17 +11,17 @@ default_target: all
 
 mobile:
 	if test ! -d build-mobile; then (echo "error: please run build.sh before make"; exit -1;); fi
-	cd build-mobile; LIBRARY_PATH=../install_root/lib make 
-	cd build-mobile; LIBRARY_PATH=../install_root/lib make install
+	cd build-mobile; make 
+	cd build-mobile; make install
 
 desktop:
 	if test ! -d build; then (echo "error: please run build.sh before make"; exit -1;); fi
-	cd build; LIBRARY_PATH=../install_root/lib make 
-	cd build; LIBRARY_PATH=../install_root/lib make install
+	cd build; make 
+	cd build; make install
 
 check:
 	if test ! -d build; then (echo "error: please run build.sh before make"; exit -1;); fi
-	cd build; LIBRARY_PATH=../install_root/lib make check
+	cd build; make check
 
 
 all: desktop mobile check
--- a/makefile
+++ b/makefile
@@ -19,4 +19,9 @@ desktop:
 	cd build; LIBRARY_PATH=../install_root/lib make 
 	cd build; LIBRARY_PATH=../install_root/lib make install
 
-all: desktop mobile
+check:
+	if test ! -d build; then (echo "error: please run build.sh before make"; exit -1;); fi
+	cd build; LIBRARY_PATH=../install_root/lib make check
+
+
+all: desktop mobile check


makefile
```diff
@@ -19,4 +19,9 @@ desktop:
 	cd build; LIBRARY_PATH=../install_root/lib make 
 	cd build; LIBRARY_PATH=../install_root/lib make install
 
-all: desktop mobile
+check:
+	if test ! -d build; then (echo "error: please run build.sh before make"; exit -1;); fi
+	cd build; LIBRARY_PATH=../install_root/lib make check
```
dirkhh: does this work? That looks like it would use `subsurface/install_root/lib` as LIBRARY_PATH and that seems wrong... because 
a) it should be `subsurface/../install/root/lib`
b) it should be `DYLD_LIBRARY_PATH=../../install-root/lib `
c) even with that, for me calling the individual tests works, but calling make check fails (and I can't figure out why)

janiversen:
It works on my mac:
Cd ssrf; make check
Cd ssrf/build; make check
Cd ssrf/build/tests; TestMerge

But I just found the reason, I have libgit2.26.dylib in /usr/local/lib

Let me see what I can do to the get rpath included in the exe

janiversen:
Try otool -l TestMerge it shows the problem libgit2.26.dylib is loaded without path (in contrast to the other dylibs).
If you do otool -l subsurface in the app dir it loads ../framework/libgit2.26.dylib which is correct.

Now let us see what can be done about that.

dirkhh:
We fix up the library path for Subsurface in `packaging\macosx\make-package.sh` when creating the DMG; but there must be some other step that gets the locally built binary to be adjusted and I'm trying to find that again (it's been a while since I worked on that particular part of the build process)...
Ahh, I see. It's the make install that fixes it up for `Subsurface.app/Contents/MacOS/Subsurface`.
`macdeployqt` replaces those paths in the binary.
We should be able to simply add an `install_name_tool` run to the build process on Darwin. Or we could try to figure out why the loader path isn't correctly embedded in the test binaries.

janiversen:
Enviado desde mi iPad

> El 6 jul 2018, a las 12:46, Dirk Hohndel <notifications@github.com> escribió:
> 
> We fix up the library path for Subsurface in packaging\macosx\make-package.sh when creating the DMG; but there must be some other step that gets the locally built binary to be adjusted and I'm trying to find that again (it's been a while since I worked on that particular part of the build process)...
> Ahh, I see. It's the make install that fixes it up for Subsurface.app/Contents/MacOS/Subsurface.
> macdeployqt replaces those paths in the binary.
> We should be able to simply add an install_name_tool run to the build process on Darwin. Or we could try to figure out why the loader path isn't correctly embedded in the test binaries.
> 
I think the most correct way it to add a install_name_tool step in case of Darwin. I will look at that later (right now I am writing qTest for qPrefDisplay).

dirkhh:
Hmm - build.sh already tries to fix up `libgit2.dylib` - so I wonder why this failed for me.
Long story short - the `LIBRARY_PATH` in your current commit still seems wrong, but my problem appears to be unrelated.
`LIBRARY_PATH` as you set it should have no impact at all, so does it work if you just remove it?

janiversen:
I copied that from line 468, 469 of build.sh, but it works fine without so I have removed it for all targets.

dirkhh:
In build.sh the variable is set to `$INSTALL_ROOT/lib` - whereas your code set it to `subsurface/install-root/lib` - but `$INSTALL_ROOT` is not below `subsurface`, it's parallel to it.
And the variable name is different by OS, unfortunately. `LIBRARY_PATH` doesn't work on macOS, IIRC.
Either way, if it works without, then let's not have it at all :-)

PR #1477, opened 2018-07-08
ssrf: remove identical if/else in CMake

janiversen:
if/else was identical, remove if/else

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

If/else (for Darwin) was identical removed if/else and only set parameter

If there is a reason for the if/else it should really have a comment, because it is not logical


bstoeger:
Note that the targets had different capitalization - maybe that was on purpose?

janiversen:
Good catch I missed that, but there are no special demand for the mac version to start with a capital letter (at least many of the apps I have downloaded do not).

neolit123:
the capitalization is intended:
https://github.com/Subsurface-divelog/subsurface/commit/7c095419f7f30203995fedcde23da6d541cf078c

/cc @dirkhh

dirkhh:
NAK - on Linux and Windows we wanted this lowercase, on Mac uppercase.
We could argue if that's the right thing to do, but we decided this years ago (this was the case even when we built with Gtk), and changing it now really serves no purpose.

PR #1478, opened 2018-07-08
Fixes / improvements to video-lite PR [#1463]

bstoeger:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A last-minute change in the video-lite PR introduced a bug (of course), which would lead to thumbnails not being properly created if local and original filename were the same (i.e. all newly added images).

Moreover, @neolit123 noted that we do not properly recognize videos that are not based on the QuickTime / MP4 formats.

Fix this and refine the icons: differentiate between "we can't read that file (io error)" and "we can't understand that file (parse error)".

### Changes made:

1) Add "unknown type" icon (LGPL from KDE breeze)
2) If we couldn't parse the video or read as image, look for video file extension
3) If all that fails, show the "unknown type" icon
4) Fix bug where thumbnails would not be properly created
5) Add changelog entry
### Related issues:

#1463
### Release note:

Done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Dive pictures: support addition of videos
 - Cloud storage: fix potential issue with credentials on Linux [#1346]
 - Mobile/iOS: fix missing translations
 - Dive pictures: locate moved pictures based on filename and path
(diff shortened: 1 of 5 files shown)


bstoeger:
Sorry, I'm silly, the bug is introduced in this PR - let's trash this.

PR #1479, opened 2018-07-08
Dive pictures: improve recognition of video files.

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

@neolit123 noted in #1463 that we do not properly recognize videos that are not based on the QuickTime / MP4 formats. Fix this for now by looking at the file extension if we could neither parse as video nor as picture. Later we can implement AVI, etc parsers.

Moreover, refine the icons: differentiate between "we can't read that file (io error)" and "we can't understand that file (parse error)".

### Changes made:

1) Add "unknown type" icon (LGPL from KDE breeze)
2) If we couldn't parse the video or read as image, look for video file extension
3) If all that fails, show the "unknown type" icon
4) Add changelog entry
### Release note:

Add changelog for #1463.

### Documentation change:

Perhaps we should mention videos in the manual?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 - Cloud storage: fix potential issue with credentials on Linux [#1346]
 - Mobile/iOS: fix missing translations
-- Dive pictures: locate moved pictures based on filename and path
+- Dive media: support addition of videos
+- Dive media: locate moved files based on filename and path
 - Profile: Context menu entry to manually split a dive
 - BLE support: fix recognition of Heinrich Weikamp OSTC 2 and OSTC Plus
 ---
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -69,15 +69,12 @@ void ImageDownloader::saveImage(QNetworkReply *reply)
 	reply->deleteLater();
 }
 
-static bool isVideoFile(const QString &filename)
+static bool hasVideoFileExtension(const QString &filename)
 {
-	// Currently, we're very crude. Simply check if the file exists and if it
-	// is an MP4-style format.
-	QFileInfo fi(filename);
-	if (!fi.exists() && !fi.isFile())
-		return false;
-	metadata md;
-	return get_metadata(qPrintable(filename), &md) == MEDIATYPE_VIDEO;
+	for (const QString &ext: videoExtensionsList)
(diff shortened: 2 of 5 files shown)


dirkhh (approved):
Look at all these comments! And that very readable code. :-)
Thanks, this looks good to me, let's see if Travis likes it as well.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Dive pictures: support addition of videos
```
neolit123: i would change this to:
```
Dive media: support addition of videos
```

our base logic and classes should have been called "media" too.
because there was potential for videos even back when photos were first added.

core/imagedownloader.cpp
```diff
@@ -90,11 +87,29 @@ Thumbnailer::Thumbnail Thumbnailer::fetchImage(const QString &filename, const QS
...
+		// For io error or video, return early with the appropriate dummy-icon.
+		if (type == MEDIATYPE_IO_ERROR)
+			return { failImage, MEDIATYPE_IO_ERROR };
+		else if (type == MEDIATYPE_VIDEO)
 			return { videoImage, MEDIATYPE_VIDEO };
+
+		// Try if Qt can parse this image. If it does, use this as a thumbnail.
 		QImage thumb(filename);
-		return { thumb, thumb.isNull() ? MEDIATYPE_IO_ERROR : MEDIATYPE_PICTURE };
+		if (!thumb.isNull())
```
neolit123, line 103: have you observed overhead when doing this on a video file instead of an image:
`QImage thumb("someVideoFile.avi")`

neolit123 (commented):
added a couple of comments, but other than that lgtm.
thanks.

CHANGELOG.md
bstoeger: Done (also changed the other "Dive pictures" entry).

core/imagedownloader.cpp
bstoeger: I added a 1.1 GB avi and it is super-fast. Quite to my embarrassment, the last time I complained about Qt's slow image loading the actual culprit was our own hashing. [Which just shows that the troubles with QMultiMedia during thumbnail-extraction could actually be my fault.]

dirkhh:
Awesome

PR #1481, opened 2018-07-09
[Testing] Play videos in profile widget

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a rebase of #1456 on current master. @sfuchs79 has ideas how playback on Windows can be made run out-of-the-box without the user having to install codecs. This PR is supposed to be a testbed for these experiments.
### Changes made:

1) Play videos in profile plot.
### Related issues:

Rebase of #1456
### Release note:

Currently not needed - test PR.

### Documentation change:

Currently not needed - test PR.


neolit123:
@sfuchs79 @bstoeger @dirkhh 

> This is a rebase of #1456 on current master. @sfuchs79 has ideas how playback on Windows can be made run out-of-the-box without the user having to install codecs

and now we support all codecs out there on Windows?
how about H.264 on Windows 7?
how about other OSes?

i'm still against attempting to play video inside subsurface.
if we end up merging something like this, please mind that i might have reservations for replying to issue reports and for the feature maintenance. :-)

bstoeger:
> and now we support all codecs out there on Windows?

Of course not. And I don't think that we have to. After all, we don't support all picture formats (notably RAW images) either. I think we should strive only for supporting the common formats out-of-the-box.

Before getting nervous about hordes of unruly users, we should see what @sfuchs79's experiments give. If we should then decide that it may be worth it *and* there really turn out to be numerous annoying users, I volunteer to maintain a FAQ. But to be honest, I don't think there will be too many. Those who complain about missing video-playback will complain anyway, no?

neolit123:
> how about other OSes?

this question is arguably more relevant?

bstoeger:
Works for me on plain Kubuntu. I guess(!) it will be the same on other commercially-backed, reasonably modern Linux-distributions.

Mac: no idea.

I understand that you are worried about fragmentation of the user-experience. But this seems to be such a minor feature that I don't think the user-experience is all that different. Playback seems like a nice-thing-to-have, but if it doesn't work then I doubt people will miss it.

In contrast, I see the thumbnailing as a much more significant change. And there we decided that it is OK to demand that the users install ffmpeg. So there we will have the real fragmentation.

dirkhh:
I need to see if I can test this on Mac. I did take a couple of ridiculously bad videos on one of my last dives.
As for user experience... yeah, consistency is great. I am much more worried about packaging and testing. But let's burn that bridge when we get to it. For now I'll happily take a look at the work that has gone into this so far.

sfuchs79:
> Before getting nervous about hordes of unruly users, we should see what @sfuchs79's experiments give.

Yep, and I have results already: No success at all. At least in my MXE based Qt 5.10.1 there is no support for the "use the wmf backend option". Everything regarding this item seems to have disappeared. Plus I found not much info on the web but what I found pointed into the direction that it was never possible to build the wmf backend (wmfengine.dll) with MinGW.
That means for Windows we can only use the DirectShow backend and this means the user needs to have codecs installed.
See also:
https://wiki.qt.io/Qt_5.11_Multimedia_Backends

bstoeger:
Heh - Thanks for your efforts. So it really looks like we should give up on QMultimedia. :/
I still would like to hear the opinion of @thiagomacieira before finally abandoning all attempts in this direction.

thiagomacieira:
I really don't know. I believe some of the qtmultimedia content requires MSVC-specific headers, which may be why you're not getting anything.

bstoeger:
Thx - let's give up on `QMultimedia` for now. :(

PR #1483, opened 2018-07-09
Improve deletion of pictures / drag & drop in profile plot

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a version of #1432 based on the decoupled ProfileWidget2/DivePictureModel code. It turns out that the simplifications in ProfileWidget2 (only keep track of shown images) made the drag&drop code quite a bit more complicated. Because now, there are numerous cases to be considered. Nevertheless, the code should not be too hard to follow. I tested all cases and for me it worked like a charm. On Linux, drag&drop is now much smoother, because not all images are recalculated.

### Changes made:

1) Update pictures on deletion
2) Intelligently update pictures on drag & drop (i.e. don't recalculate all thumbnails)
### Release note:

Not a super-important change, but I think we can add an entry.
### Documentation change:

None needed.

### DIFF
--- a/core/qthelper.h
+++ b/core/qthelper.h
@@ -88,6 +88,32 @@ QString getUserAgent();
 #define TITLE_OR_TEXT(_t, _m) _t, _m
 #endif
 
+// Move a range in a vector to a different position.
+// The parameters are given according to the usual STL-semantics:
+//	v: a container with STL-like random access iterator via std::begin(...)
+//	rangeBegin: index of first element
+//	rangeEnd: index one *past* last element
+//	destination: index to element before which the range will be moved
+// Owing to std::begin() magic, this function works with STL-like containers:
+//	QVector<int> v{ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+//	moveInVector(v, 1, 4, 6);
+// as well as with C-style arrays:
+//	int array[10] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+//	moveInVector(array, 1, 4, 6);
+// Both calls will have the following effect:
+//	Before: 0 1 2 3 4 5 6 7 8 9
+//	After:  0 4 5 1 2 3 6 7 8 9
+// No sanitizing of the input arguments is performed.
+template <typename Vector>
+void moveInVector(Vector &v, int rangeBegin, int rangeEnd, int destination)
+{
+	auto it = std::begin(v);
+	if (destination > rangeEnd)
+		std::rotate(it + rangeBegin, it + rangeEnd, it + destination);
+	else if (destination < rangeBegin)
+		std::rotate(it + destination, it + rangeBegin, it + rangeEnd);
+}
+
 #endif
 
 // 3) Functions visible to C and C++
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -2094,6 +2094,26 @@ bool ProfileWidget2::PictureEntry::operator< (const PictureEntry &e) const
 	return std::tie(offset.seconds, filename) < std::tie(e.offset.seconds, e.filename);
 }
 
+void ProfileWidget2::calculatePictureYPositions()
+{
+	double lastX = -1.0, lastY;
+	for (PictureEntry &e: pictures) {
+		if (!e.thumbnail)
+			continue;
+		// let's put the picture at the correct time, but at a fixed "depth" on the profile
+		// not sure this is ideal, but it seems to look right.
+		double x = e.thumbnail->x();
+		double y;
+		if (lastX >= 0.0 && fabs(x - lastX) < 3 && lastY <= (10 + 14 * 3))
+			y = lastY + 3;
+		else
+			y = 10;
+		lastX = x;
+		lastY = y;
+		e.thumbnail->setY(y);
+	}
+}
+
 // This function resets the picture thumbnails of the current dive.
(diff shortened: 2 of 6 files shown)


core/qthelper.h
```diff
@@ -88,6 +88,32 @@ QString getUserAgent();
...
+// Owing to std::begin() magic, this function works with STL-like containers:
+//	QVector<int> v{ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+//	moveInVector(v, 1, 4, 6);
+// as well as with C-style arrays:
+//	int array[10] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+//	moveInVector(array, 1, 4, 6);
+// Both calls will have the following effect:
+//	Before: 0 1 2 3 4 5 6 7 8 9
+//	After:  0 4 5 1 2 3 6 7 8 9
+// No sanitizing of the input arguments is performed.
```
neolit123, line 106: this sort of detail in comments is what's usually needed for some of the C++11 in our codebase. and not only, some routines are fairly undocumented code-wise.
thank you!

profile-widget/profilewidget2.cpp
```diff
@@ -2094,6 +2101,42 @@ bool ProfileWidget2::PictureEntry::operator< (const PictureEntry &e) const
...
+		// let's put the picture at the correct time, but at a fixed "depth" on the profile
+		// not sure this is ideal, but it seems to look right.
+		double x = e.thumbnail->x();
+		double y;
+		if (lastX < 0.0)
+			y = 10;
+		else if (fabs(x - lastX) < 3 && lastY <= (10 + 14 * 3))
+			y = lastY + 3;
+		else
+			y = 10;
```
neolit123: i think can be written as:
```
if (fabs(x - lastX) < 3 && lastY <= (10 + 14 * 3))
	y = lastY + 3;
else
	y = 10;
```

profile-widget/profilewidget2.cpp
```diff
@@ -2166,23 +2192,87 @@ void ProfileWidget2::dropEvent(QDropEvent *event)
...
+
+		// A picture was drag&dropped onto the profile: We have four cases to consider:
+		//	1a) The image was already shown on the profile and is moved to a different position on the profile.
+		//	    Calculate the new position and move the picture.
+		//	1b) The image was on the profile and is moved outside of the dive time.
+		//	    Remove the picture.
+		//	2a) The image was not on the profile, but belongs to the current dive.
+		//	    Add the picture to the profile if it is during the dive.
+		//	2b) The picture does not belong to the current dive.
+		//	    For now, do nothing. We may think about adding the picture to the dive.
```
neolit123, line 2214: nice, more detailed comments :+1:

neolit123 (commented):
thank for the PR.
added a small note, but overall cannot fault the logic and the code comments are very detailed.

EDIT: BTW i haven't tested this on Windows, i might be able to find time tomorrow.
because we need to test it on all OSes.

profile-widget/profilewidget2.cpp
bstoeger: This is only copy & paste, but yeah - let's just simplify this.

janiversen:
Or maybe describe “what” it does, instead of “how” it does it. To me a comment is more to help understand why the code is there. Having read a no small part of our code I constantly stumble over the “why”. But I think your current approach is really good.

neolit123:
> TBH - I think verbose comments are a double edged sword. For someone who knows the standard library well, explaining std::emplace, etc. will make the code less readable. And the big danger is that comments and code diverge - then they may give the reader the false impression of understanding what's going on, whereas in reality they obscure it.

so keep in mind that the only regularly active person on this team that pushes for us to use the C++ stdlib (and apparently knows it well) is you. :)

if we do a quiz on home many devs from our team have used `std::rotate` before, the results would be 2-3 out of 20.

> So I think we should strive for a sensible compromise.

> Or maybe describe “what” it does, instead of “how” it does it. To me a comment is more to help understand why the code is there. Having read a no small part of our code I constantly stumble over the “why”. But I think your current approach is really good.

yes, i think we should strive for sensible amount of comments on the logic itself with potential notes here and there on how this is done by using the C++'s stdlib.
the subsurface comments should not be a C++ tutorial, that's for sure.

sfuchs79:
> I tested all cases and for me it worked like a charm.

I would like to help by testing under Windows 10. But I'm not 100% sure which are the most important test cases. @bstoeger Could you please give me a few hints?

dirkhh:
Question: I want to cut 4.8.1 very soon to fix some of the problems, especially on Mac.
Should this wait after 4.8.1, or are you comfortable to make this one of the 4.8.1 features?

bstoeger:
This is UI-polishing, so I don't deem it particularly important.

But I am somewhat scared about the recent commits to the profile-thumbnails. Hopefully you won't have to cut a 4.8.2 immediately after 4.8.1. 😮

dirkhh:
No risk no fun. Actually - I have tried this on a couple different machines and it seemed to work fine.
And yes, I'm always willing to do a quick follow up release if needed.

bstoeger:
As you wish. Pushed a CHANGELOG.md change, just in case.

dirkhh:
Based on no good reason whatsoever I decided to wait on this one until after 4.8.1

dirkhh:
manually pulled

PR #1486, opened 2018-07-10
Dive media: Extract thumbnails from videos with ffmpeg

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a test-commit for @atdotde: Rudimentary video-thumbnail extraction with ffmpeg.

### Changes made:

Extract thumbnails from videos via ffmpeg.

### Release note:

Not yet - test-PR
### Documentation change:

Not yet - test-PR

### Mentions:

@atdotde

### DIFF
--- a/desktop-widgets/preferences/preferences_defaults.ui
+++ b/desktop-widgets/preferences/preferences_defaults.ui
@@ -73,7 +73,7 @@
         <item>
          <widget class="QRadioButton" name="noDefaultFile">
           <property name="text">
-           <string>No default file</string>
+           <string>&amp;No default file</string>
           </property>
          </widget>
         </item>
@@ -203,6 +203,79 @@
        <widget class="QSpinBox" name="velocitySpinBox">
         <property name="maximum">
          <number>500</number>
+        </property>
+       </widget>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item>
+    <widget class="QGroupBox" name="groupBox_10">
+     <property name="title">
+      <string>Video thumbnails</string>
+     </property>
+     <layout class="QFormLayout" name="formLayout">
+      <property name="horizontalSpacing">
+       <number>5</number>
+      </property>
+      <property name="verticalSpacing">
+       <number>5</number>
+      </property>
+      <property name="margin">
+       <number>5</number>
+      </property>
+      <item row="1" column="0">
+       <widget class="QLabel" name="ffmpegExectuableLabel">
+        <property name="text">
+         <string>ffmpeg executable</string>
+        </property>
+       </widget>
+      </item>
+      <item row="1" column="1">
+       <layout class="QHBoxLayout" name="horizontalLayout_3b">
+        <item>
+         <widget class="QLineEdit" name="ffmpegExecutable"/>
+        </item>
+        <item>
+         <widget class="QToolButton" name="ffmpegFile">
+          <property name="text">
+           <string>...</string>
+          </property>
+         </widget>
+        </item>
+       </layout>
+      </item>
+      <item row="3" column="0">
+       <widget class="QLabel" name="videoThumbnailPositionLabel">
+        <property name="text">
+         <string>Extract at position</string>
+        </property>
+       </widget>
+      </item>
+      <item row="3" column="1">
+       <widget class="QSlider" name="videoThumbnailPosition">
+        <property name="maximum">
+         <number>100</number>
+        </property>
+        <property name="value">
+         <number>20</number>
+        </property>
+        <property name="orientation">
+         <enum>Qt::Horizontal</enum>
+        </property>
+       </widget>
+      </item>
+      <item row="0" column="0">
+       <widget class="QLabel" name="extractVideoThumbnailsLabel">
+        <property name="text">
+         <string>Extract video thumbnails</string>
+        </property>
+       </widget>
+      </item>
+      <item row="0" column="1">
+       <widget class="QCheckBox" name="extractVideoThumbnails">
+        <property name="text">
+         <string/>
         </property>
        </widget>
       </item>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Dive media: paint duration of videos on thumbnails in the dive-photo tab
+- Dive media: extract video thumbnails using ffmpeg
 - Dive media: implement "open folder of selected media files" [#1514]
 - Dive media: show play-time of videos on the profile plot
 - Dive media: draw thumbnails on top of deco-ceilings
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -95,6 +95,7 @@ set(SUBSURFACE_CORE_LIB_SRCS
 	uemis.c
 	uemis-downloader.c
 	version.c
+	videoframeextractor.cpp
 	windowtitleupdate.cpp
 	worldmap-save.c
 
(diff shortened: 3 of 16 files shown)


sfuchs79:
I tested this and it works nicely here for me under Windows 10.

Simply out of curiosity I even used my "own" cross compiled ffmpeg (version 3.4.2) from MXE. I had to copy ffmpeg.exe and a couple of DLLs to the Subsurface root installation folder.
Just to create no misunderstanding: This is no vote from my side to put this stuff into the Subsurface Windows installer. I can understand that this might not be the preferred solution. In fact it is ~35MB of additional "stuff". Just wanted to tell you that it worked for me in that way :-)

dirkhh:
We can of course *document* how to do this. Here's where you can download things. Here's where you need to copy them.

bstoeger:
> I had to copy ffmpeg.exe and a couple of DLLs to the Subsurface root installation folder.

Cool, then maybe for now we don't even have to use a preferences option. Just tell the user where to put extract the files. For Linux it's simply a matter of having ffmpeg in the PATH, which every sensible distribution should be able to do. For Mac: no idea..!?

Since ffmpeg is GPL + LGPL, I think I could host such an archive.

Note that this is WIP. I want to change the format of the thumbnail files such that for videos we also save the duration (if we know it - see #1494). And it should also be able to contain multiple images (every n-seconds of the video).

sfuchs79:
It may get even better. I'm currently playing with ffmpeg configure options to get a tiny and statically linked ffmpeg.exe which is optimized to only do the specific job we want it to do. Unfortunately with my first try I may have removed too many options - it doesn't work any more (but it is only ~5MB <g>). Could you change your PR in a way that I can see the ffmpeg error message? Currently I only get a "Failed reading ffmpeg output".

bstoeger:
Changed it to print `stderr` in case of failure. The code to print `stdin` is commented out. Perhaps you want to change this to a `printf` so that it is properly formatted. We could also show a `QMessageBox`, perhaps of the collected failures after addition of a bunch of videos.

sfuchs79:
Thanks! With help of the ffmpeg error messages I cross compiled ffmpeg with a few other configurations. I now ended up with a statically linked ffmpeg.exe of about 10MB which does the job of creating thumbnails for me.

Config was (copied from MXE ffmpeg.mk):
         --cross-prefix='$(TARGET)'- \
        --enable-cross-compile \
        --arch=$(firstword $(subst -, ,$(TARGET))) \
        --target-os=mingw32 \
        --prefix='$(PREFIX)/$(TARGET)' \
        --enable-static --disable-shared \
        --enable-small \
        --yasmexe='$(TARGET)-yasm' \
        --disable-debug \
        --disable-pthreads \
        --enable-w32threads \
        --disable-doc \
        --disable-ffplay \
        --disable-ffprobe \
        --disable-ffserver \
        --disable-bsfs \
        --disable-encoders \
        --enable-encoder=mjpeg \
        --disable-filters \
        --enable-filter=scale \
        --disable-muxers \
        --enable-muxer=image2

atdotde:
Of course, you can also use ffmpeg to figure out the duration (or rather its helper ffprobe):

https://superuser.com/questions/650291/how-to-get-video-duration-in-seconds

I don't know if shipping ffmpeg is such a great idea. Why not rather relay on it being there or directing the users to 

https://www.ffmpeg.org/download.html

or using the package manager of their system (on Mac, home-brew happily installs it and places it in the path).

bstoeger:
> Of course, you can also use ffmpeg to figure out the duration (or rather its helper ffprobe)

Yes, but we should strive to give the user the best possible experience without having to install external software as long a the effort stays within reasonable limits. Extracting date and especially duration turns out to be very simple (I'm currently looking at .wmv). I don't want to make ffmpeg a dependency for all video-manipulations. Only for thumbnailing.

bstoeger:
Rebased on master.
Added three preferences fields, which control extraction (quote from commit-message):
- `extract_video_thumbnails` (bool): if true, thumbnails are calculated.
- `extract_video_thumbnail_position` (int 0..100): position in video where thumbnail is fetched.
- `ffmpeg_executable (string)`: path of ffmpeg executable.

If ffmpeg fails to start, thumbnailing is turned off. Preferences-fields not yet wired to frontend!

bstoeger:
Added a "video-file-marker" to mark video thumbnails. The "design" is pure garbage - this will have to be fixed by a more capable person (file: `icons/video_overlay.svg`). Note that the marker is added on-the-fly. The thumbnails on-disk are not messed up. I.e. nothing should be destroyed by the PR as-is.

bstoeger:
Update:
- connected the preferences values to the UI.
- render duration on thumbnails in dive-photo-tab.

@dirkhh, @sfuchs79, @atdotde: From my side this is feature complete, therefore I allowed my self to mark the last commit as fixes #359. This still needs some UI / graphics polishing and there are many options for improvement (@atdotde has the idea of extracting frames every n-seconds; we could also snatch half-a-second of movie, like youtube does). But personally, I'd like to limit my work on this to bug-fixing and polishing, so that I can bother with other things, notably undo.

bstoeger:
I replaced the work-in-progress by the needs-testing label, because from my side, this is ready. 

I would really prefer if this was addressed before the corresponding preferences-code is reworked. Doing it once was painful enough.

janiversen:
If you tell which parts of prefs it is, I will hold thst part back a while.

bstoeger:
```
bool GeneralSettingsObjectWrapper::extractVideoThumbnails() const
{
	return prefs.extract_video_thumbnails;
}
 int GeneralSettingsObjectWrapper::extractVideoThumbnailsPosition() const
{
	return prefs.extract_video_thumbnails_position;
}
 QString GeneralSettingsObjectWrapper::ffmpegExecutable() const
{
	return prefs.ffmpeg_executable;
}
```

That would be very nice - thanks.

desktop-widgets/preferences/preferences_defaults.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>561</width>
-    <height>558</height>
+    <width>588</width>
+    <height>600</height>
```
dirkhh: this always worries me - I assume it's just automatic tool output, but we used to struggle a lot with the preferences window suddenly not fitting in our design goal screen size of 1024x600 and people not being able to change preferences

dirkhh (commented):
I haven't tested this, just reviewed the code. That first commit is massive, but it makes sense as is. I worry about the impact on small screens (see my comment), but I don't think that's a show stopper; just something to test in a VM.
Which brings me to what this really needs.
Testing on Windows, Mac, and Linux.
Oh, and an addition to the user manual :-)

desktop-widgets/preferences/preferences_defaults.ui
bstoeger: Yes - that was my fault. Apparently I played with the window-size in Qt Creator. I wonder: Couldn't the window-size be calculated from the preferred sizes of the widgets...?

I removed the gratuitous `leftMargin`, etc. changes. It's really rude of Qt Creator to mangle these things.

bstoeger:
> Which brings me to what this really needs.
> Testing on Windows, Mac, and Linux.

I tested it quite extensively on Kubuntu. Of course, experience shows that testers who are not the author often find problems in use-cases that the author never thought about.

@sfuchs79 tested an older version on Windows.

I don't have access to Mac. So I'm out in that respect. 

> Oh, and an addition to the user manual :-)

Yes, looked into it a few days ago and quickly gave up. There's quite a lot to change!

dirkhh:
Tested on Mac, both with and without `ffmpeg`
VERY nice work @bstoeger

PR #1488, opened 2018-07-10
Add qml test harness (needed to test the qPref classes) (merge#1)

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add a TestQML exe, that runs all tst_* files in subsurface/tests

TestQML is aware of the qPref object, thus the C++ QML connection can be test

tst_qpref.qml is provided for a base test and an example.

The idea is that all qPref classes will have a tst_ file to ensure the properties are avialable.

In a later PR, the whole registrations process (currently split over 2 files in subsurface, 1 files in desktop widgets and 1 files in tests, will be changed to a single shared function


tests/tst_qPref.qml
```diff
@@ -0,0 +1,17 @@
...
+TestCase {
+	name: "qPref"
+
+	SsrfPrefs {
+		id: prefs
+	}
+
+	function test_register() {
+		var x = prefs.mobile_version
+		compare(x, prefs.mobile_version)
```
dirkhh, line 15: isn't that a tautology?
I'm not quite sure what this tests, TBH

janiversen: it test that the type is registred and can be accessed (which happened to be the problem with my previous PR). Normally you will see test files with a lot more test cases, I kept this simple in order to make it easy to copy.

janiversen:
the second failure needs an update to the PR, will do that tomorrow

dirkhh:
With the switch to Qt 5.11 as our main target, I'm considering moving from Qt 5.5 to Qt 5.6 as oldest supported version. That one is a long term release - Qt 5.5 is actually no longer supported by Qt...
I don't know, however, if that test case wouldn't fail with Qt 5.6 as well.

janiversen:
the failure is a system() definition, I will just build the c string differently.

I would recommend moving to Qt 5.6

janiversen:
Qml testing is not possible in the oldest Qt versions we support therefore added version check to skip test if not supported.

dirkhh:
Same here, I'd like to wait until after 4.8.1 is released

janiversen:
totally agree, and I will use the time to use update the PRs to use the new shared register_qml_types()

dirkhh:
Now that 4.8.1 is out I'll be happy to take those three PRs. Do you want to just collect them into one to avoid the conflicts? Just let me know how you want to handle things.

janiversen:
I think the easiest for me is if you merge them one by one. In cas of conflict simply stop and let me solve it, Your merge of the qmlRegisterType made it possible for me to clean the 3 pr’s so I expect no conflicts.

If you prefer or it is easier for you I can put all the commits in one of the pr.

janiversen:
Give me a couple of hours t¡rhen the changelig.md conflict will be gone

dirkhh:
When I manually merged this one it did something weird with removing code from `subsurface-helper.cpp`
Let me try again and worst case I'll have you do a fixup PR while I'm at Taekwondo this morning :-)

PR #1491, opened 2018-07-11
core: activate qPrefDisplay and add test cases. (merge#2)

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR activates qPrefDisplay which was added as a "showcase" in #1459 

#1459 showed how the new qPref macros was used, in order to discuss the implementation,
and this PR puts qPrefDisplay into production.

Remark with this PR, structure preferences prefs are the only source for display variables !

A number of standard test cases have been added to help secure against unwanted changes:
structure prefs <-> qPrefDisplay
structure prefs <-> disk (remark the disk names remain unchanged)

In order to stabilise the C++ QML Interface a number of QML test cases have been added,
the use of qml test cases help secure that the qml interface works.

This PR depends on #1488 and probably need a manual merge which I will do when #1488 is merged.

This PR will be followed by other PRs one for each "class" of variables in struct preferences.


janiversen:
Capital letter problem, so the question should have been "why is it found on the other platforms" :-)

janiversen:
Now it also compiles with Qt 5.6. But this PR still depend on #1488

dirkhh:
I would like to delay this one also until the weekend...

PR #1494, opened 2018-07-11
Metadata improvements: duration, AVI & WMV

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This adds two new features to the metadata extractor:

1) Extract duration from videos. We might want to display the whole duration of the dive in the profile (@atdotde has some interesting ideas in this regard). Moreover, in the future I want to be able to add photos/videos of a whole dive-trip at once. The file should be associated to the dive it is closest to. This is only possible for videos if we know start *and* end-time.

2) Preliminary support for AVI. Extraction of duration seemed trivial, but unfortunately the creation date does not seem to be standardized. I implemented the two versions that I found on the web. Ultimately, I fear we will have to see what our users need. Perhaps nobody uses AVI anyway. I'm not aware of a camera that outputs such a thing..!?

### Changes made:

1) Extract duration from videos.
2) Preliminary support for AVIs.
### Release note:

Perhaps?
### Documentation change:

Perhaps together with general info on video support?

### DIFF
--- a/core/metadata.cpp
+++ b/core/metadata.cpp
@@ -158,18 +158,26 @@ static bool parseMP4(QFile &f, metadata *metadata)
 			if (f.read(&data[0], atom_size) != static_cast<int>(atom_size))
 				break;
 			uint64_t timestamp = 0;
+			uint32_t timescale = 0;
+			uint64_t duration = 0;
 			// First byte is version. We know version 0 and 1
 			switch (data[0]) {
 			case 0:
 				timestamp = getBE<uint32_t>(&data[4]);
+				timescale = getBE<uint32_t>(&data[12]);
+				duration = getBE<uint32_t>(&data[16]);
 				break;
 			case 1:
 				timestamp = getBE<uint64_t>(&data[4]);
+				timescale = getBE<uint32_t>(&data[20]);
+				duration = getBE<uint64_t>(&data[24]);
 				break;
 			default:
 				// For unknown versions: ignore -> maybe we find a parseable "mdhd" atom later in this file
 				break;
 			}
+			if (timescale > 0)
+				metadata->duration.seconds = lrint((double)duration / timescale);
 			// Timestamp is given as seconds since midnight 1904/1/1. To be convertible to the UNIX epoch
 			// it must be larger than 2082844800.
 			if (timestamp >= 2082844800) {
@@ -194,6 +202,7 @@ static bool parseMP4(QFile &f, metadata *metadata)
 extern "C" mediatype_t get_metadata(const char *filename_in, metadata *data)
 {
 	data->timestamp = 0;
+	data->duration.seconds = 0;
 	data->latitude.udeg = 0;
 	data->longitude.udeg = 0;
 
--- a/core/metadata.cpp
+++ b/core/metadata.cpp
@@ -403,6 +403,89 @@ static bool parseAVI(QFile &f, metadata *metadata)
 	return found_riff;
 }
 
+static bool parseASF(QFile &f, metadata *metadata)
+{
+	f.seek(0);
+
+	// Parse the header of the header object:
+	//	id				(16 bytes)
+	//	size				 (8 bytes)
+	//	number of header objects	 (4 bytes)
+	//	reserved			 (2 bytes)
+	//	------------------------------------------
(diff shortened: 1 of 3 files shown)


neolit123:
> Preliminary support for AVI. Extraction of duration seemed trivial, but unfortunately the creation date does not seem to be standardized. I implemented the two versions that I found on the web. Ultimately, I fear we will have to see what our users need. Perhaps nobody uses AVI anyway. I'm not aware of a camera that outputs such a thing..!?

would that work for other containers different from AVI?

bstoeger:
> would that work for other containers different from AVI?

Sure - in my tests it works nicely for MP4/Quicktime derived formats. I would guess that makes up > 90% of all video files nowaday?

Indeed, the first commit *only* implements duration extraction for these formats. I thought that was obvious, so I didn't mention it.

bstoeger:
Even though this is complex code - I think this can go into 4.8.1. The non-trivial part only affects people uploading AVIs, which we didn't even support before. I threw a some random files at it and nothing bad happened (knock wood).

core/metadata.cpp
```diff
@@ -199,6 +222,187 @@ static bool parseMP4(QFile &f, metadata *metadata)
...
+	}
+
+	// I've also seen "Weekday Mon  Day hh:mm:ss yyyy"(!)
+	QStringList items = s.split(' ', QString::SkipEmptyParts);
+	if (items.size() < 4)
+		return false;
+
+	// Skip weekday if any is given
+	for (const QString &day: weekdays) {
+		if (items[0].startsWith(day, Qt::CaseInsensitive)) {
```
dirkhh, line 255: so this is always encoded in English abbreviation format?
That's crazy

dirkhh:
I agree that this isn't really high risk in the sense of creating a regression. It could cause all kinds of fun issues for users playing with it, but hey...
Should this get a CHANGELOG entry that mentions "experimental support for..."?
Eventually this also needs to be mentioned in the user manual - but that doesn't need to happen for 4.8.1
I'll give others a few more hours to look and comment, but I'm willing to merge this.

core/metadata.cpp
bstoeger: I think there is no standardized way to record the creation date. As far as I understood, the creating software is basically free to write in any human-readable string it wants.

I implemented the two ways I found on the web. I'd really appreciate any real-world AVI examples. But I'm not sure if any modern video equipment produces such a thing.

dirkhh: OK, makes sense. Let's go with what we have then.
I'd still appreciate the CHANGELOG entry :-)

bstoeger:
Heh.. Already found a bug (64 bit values > 0xFFFFFFFF were not read correctly).

bstoeger:
Well - threw in a WMV parser for good measure. By far the easiest so far. Weirdly I didn't find a .wmv file with a valid date online..?

bstoeger:
I tested this with the following procedure
```
#include <QDir>
static void test_metadata(const QDir &dir)
{
	for (const QString &file: dir.entryList(QDir::Files)) {
		QString path = dir.filePath(file);
		printf("%s is ", qPrintable(path));
		metadata md;
		switch(get_metadata(qPrintable(path), &md)) {
		case MEDIATYPE_PICTURE:
			printf("a picture with timestamp: %ld.\n", md.timestamp);
			break;
		case MEDIATYPE_VIDEO:
			printf("a video with timestamp: %ld and duration: %d.\n", md.timestamp, md.duration.seconds);
			break;
		case MEDIATYPE_IO_ERROR:
			printf("not readable.\n");
			break;
		default:
			printf("not supported.\n");
			break;
		}
	}
	for (const QString &dirname: dir.entryList(QDir::NoDotAndDotDot | QDir::Dirs))
		test_metadata(dir.filePath(dirname));
}

extern "C" void test_metadata()
{
	test_metadata(QDir("/home/bs"));
}
```
on all files in my home directory - and believe me, this includes a lot of random garbage. It was reasonably fast and did not crash. So I think apart from erroneous extraction of video-timestamps this code should be rather safe.

dirkhh:
merged

PR #1496, opened 2018-07-12
Qt-script update

jbygdell:
With commit c62ce43 the Qt version where updated to 5.11.1, this updates the qt install script to match the versions of the Qt parts we want.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

When changing Qt version both the variables.sh and qt-installer-noninteractive.qs files needs to be updated. Comment added to the variables file not to forget to also change the qt-install script.

### DIFF
--- a/packaging/android/variables.sh
+++ b/packaging/android/variables.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
-
+# When changing Qt version remember to update the 
+# qt-installer-noninteractive file as well.
 QT_VERSION=5.11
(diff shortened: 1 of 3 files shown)


dirkhh:
Thank you for those. Both of them should have been obvious to me. I am still happier with the Travis Android solution (much faster), but I completely missed that I SHOULD have done this in the first place.
Much appreciated.

PR #1460, opened 2018-07-04
Multiple cylinder edit on mobile

jbygdell:
### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:
We have had multiple requests both here #547 and on the ML to support editing multiple gas mixes and cylinders on the mobile app.
This PR is a stab at enabling that functionality without cluttering the UI.
The solution is to display multiple cylinders as comma separated list on the dive details page.

When editing a dive with multiple cylinders only the used cylinders are displayed on the edit page.

When adding a new dive manually on the mobile app there is a limit to a single cylinder as we don't yet have the option to insert a gas switch event.

### Changes made:
1) Displays all used cylinders when viewing info for a dive.
2) Allow the user the user to edit multiple cylinders using the current GUI, in order not to clutter the UI only the cylinders that are in use are displayed.

### Related issues:
Fixes #547

### Release note:
- Mobile: enable editing of multiple cylinders in a dive

### Documentation change:
The user can edit cylinder information for a dive that uses multiple cylinders on the mobile app.
Adding a dive manually is restricted to a single cylinder since we can not create gas switch events in the mobile app.

PR #1497, opened 2018-07-12
Don't show decimals on pressures

jbygdell:
As per discussion in #1460 there is no point in showing decimal values
for pressures in the equipment tab on desktop or in the dive edit view on 
mobile.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


neolit123:
thanks, this seems like the right thing to do good, as per the discussion in the other thread. :+1:

PR #1498, opened 2018-07-12
core: activate qPrefAnimations and test cases. (merge#3)

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR moves Animations from SettingsObjectWrapper to qPrefAnimations as pr #1459

Remark with this PR, structure preferences prefs are the only source for display variables !

A number of standard test cases have been added to help secure against unwanted changes:
structure prefs <-> qPrefAnimations
structure prefs <-> disk (remark the disk names remain unchanged)

In order to stabilise the C++ QML Interface a number of QML test cases have been added,
the use of qml test cases help secure that the qml interface works.

This PR depends on #1491 and probably need a manual merge which I will do when #1491 is merged.

This PR will be followed by other PRs one for each "class" of variables in struct preferences.


tests/testqPrefDisplay.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void TestQPrefDisplay::test_set_load_struct()
+{
+	// test set func -> load -> struct pref 
+
+	auto display = qPrefDisplay::instance();
+
+	display->set_display_invalid_dives(false);
+	display->set_divelist_font("helvitica");
```
bstoeger, line 62: Perhaps I'm missing a play on words - but the original font is called "Helvetica" (i.e. hailing from Switzerland).

core/settings/qPrefAnimations.h
```diff
@@ -0,0 +1,38 @@
...
+	Q_OBJECT
+	Q_PROPERTY(int animation_speed READ animation_speed WRITE set_animation_speed NOTIFY animation_speed_changed);
+
+public:
+	qPrefAnimations(QObject *parent = NULL);
+	static qPrefAnimations *instance();
+
+	// Load/Sync local settings (disk) and struct preference
+	void loadSync(bool doSync);
+	void load() { loadSync(false); };
```
bstoeger: Pedantic mode: In C (and by inheritance in C++), function definitions don't end in ';'.
What you are writing is the equivalent of
```
struct {
	int a;;
};
```
Sure - it's valid code but also kind of ugly. ;)

tests/testqPrefDisplay.cpp
janiversen: good catch, except this is a test case so I could have called it (nearly) anything.

core/settings/qPrefAnimations.h
janiversen: Pedantic mode is surely ok, but I find it interesting you comment it here and not before (all my PR with class definitions have this).

It is not a big deal for me, except for the monster PR it calls for (to change C++ header files which uses this practice, a quick look on other header files shows this seems to be common practice).

I like to use ; to ensure the statement is terminated and do not find it ugly in contrary :-)

tests/testqPrefDisplay.cpp
janiversen: I have changed it, in the PR dealing with qPrefDisplay #1491 (as stated in the PR this one depends on #1491, which is why you see those commits here).

core/settings/qPrefAnimations.h
janiversen: Could not resist a cross check, please have a look at your own #1486 core/imagedownloader.h  and core/videoframeextractor.h why is it not kind of ugly there :-)

Also if I change the function defintion to:
      void loadSync(bool doSync)
      void load() { loadSync(false); }
I get clang compiler errors

If I change the function call (you wrote definition):
      void load() { loadSync(false)  };
I get clang compiler errors

So what ; do you find kind of ugly ?

dirkhh:
Given the potential for things to go sideways with this, I'd hold this series of pull requests back for a couple of days (I'm guessing until Saturday or Sunday) in order to get 4.8.1 out.
That's not an indication that I don't want these patches, it's an indication that I want to get 4.8.1 out in the next 48 hours.

janiversen:
For sure, holding these changes back is the right thing to do. I had in my mind that you talked about 6 weeks until next release which would be different.

Do NOT get stressed over these PR, next week is fine by me, but please have a look at especially the first 3, so we are sure there are no outstanding issues.

dirkhh:
I merged the one that moved code around. And I will definitely go through them one more time, hopefully later today, in case there is more feedback that I have for you.

Typically I aim for about 6 weeks between releases, unless something is really broken. And right now at least Mac is really broken (crash when opening user manual or FB). There's a fix for a crash on Linux - Windows seems actually ok :-)

dirkhh:
I merged all three of them manually - please check that this is what you wanted and if it isn't, send us another PR :-)

janiversen:
I think it is risky to merge all the commits in one pr to just a merge commit, that makes bisecting a lot more difficult especially since all my commit comments gone.

a very minor detail, you are the author (looked witg “git log”.

If you not mind a lot, then please revert your 3 commits, and I wll make a new PR with all the single commits, deal?

just have to manipulate my master (so I can make the PR while you are studying martial arts, I only got to jui jitsu (self defense))

janiversen:
now I am confused, this is git cheating 

git log shows only your 3 commits, but git rebase shows all my commits, do not understand what is happening here

bstoeger:
> git log shows only your 3 commits

I think I see your commits. It's just that they are farther down the list, because this was a "merge" instead of a "rebase" merge. Or something. I never groked git - my brain's just not built for that.

janiversen:
that makes 2 of us. To me it seems like master history was rewritten.

I finally found the commits, there are on july 12, before the 4.8.1 tag. I wonder if they are still there if I checkout v4.8.1

hope @dirkhh will write a word or two or give an understandabke link (just read the git merge manual).

dirkhh:
>that makes 2 of us. To me it seems like master history was rewritten.

It wasn't. It's just that git log shows commits chronologically by commit time. The merge didn't change your commits at all, so they show up when you made them.
I always recommend using a graphical tool to look at got history, e.g. gitk
That shows things in a much more logical way.
I'm on my phone or I'd paste a picture

Trust me, nothing was lost, nothing was rewritten (or your git pull would have failed as not being a fast forward)

>I finally found the commits, there are on july 12, before the 4.8.1
>tag. I wonder if they are still there if I checkout v4.8.1

Of course not :-)

git is based on set theory, not on a linear history. So git log showing a linear history is just a bad cludge.

>hope @dirkhh will write a word or two or give an understandabke link
>(just read the git merge manual).

Let me know if this helps, happy to explain more

bstoeger:
> I guess I am just used to you merging fast

I think usually @dirkhh does a "rebase" merge, where the branch-to-be-merged is rebased on master first. This seems to give the commits a more recent "commit time" and there is no "merge commit".

> git is based on set theory,

You sure do realize that Cantor went insane over set theory. 😛

dirkhh:
The real difference is the way we merge on GitHub by rebasing. There the commit time stamp of every commit changes (because the person merging becomes the commiter). So the commits show up at the top of the linearized git log.
The way I merged your three pull requests didn't touch your commits at all. You stayed the commiter, your commit time is reflected, so they appear to move back in the linearized log.
I did that because I wanted to keep the merge commits intact so we could see if I messed things up.

dirkhh:
>> git is based on set theory,
>
>You sure do realize that Cantor went insane over set theory. 😛

Have you met the gentleman who designed git and did the initial implementation? 😇
-- 
from my phone.

PR #1501, opened 2018-07-12
map-widget: fix chronology of QML class registration

### LINKED ISSUE #1500: Map not working under Windows 10 with 4.8.0-100
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

With latest continuous build 4.8.0-100 and with my own MXE based Windows build based on master the map is not working.
I only see the (well known) error message in the map window.
I guess this comes from one of the latest commits because my last build with working map can't be longer than 2-3 days ago.

### Operating system:

Windows 10

### Subsurface version:

4.8.0-100

### Mentions:

@dirkhh @janiversen @jbygdell @bstoeger

neolit123:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The QML types needs to be registered before the MainWindow
instance is created on the desktop version, otherwise
the MapWidget instance will be created and it will fail
with a missing QML namespace.

In subsurface-helper.cpp, move register_qml_types() from run_ui()
to init_ui(), as later in init_ui(), MainWindow is instantiated
for the first time in the desktop version.

### Changes made:

move a function call in subsurface-helper.cpp.

### Related issues:

Fixes #1500
https://github.com/Subsurface-divelog/subsurface/commit/ab05fe3cf83526ea63764a6c6587735e8b959527

### Additional information:

NONE

### Release note:

NONE

### Documentation change:

NONE

### Mentions:

@dirkhh @janiversen @sfuchs79

### DIFF
--- a/subsurface-helper.cpp
+++ b/subsurface-helper.cpp
@@ -33,6 +33,7 @@ QObject *qqWindowObject = NULL;
 void init_ui()
 {
 	init_qt_late();
+	register_qml_types();
 #ifndef SUBSURFACE_MOBILE
 	PluginManager::instance().loadPlugins();
 
@@ -97,7 +98,6 @@ void register_qml_types()
 
 void run_ui()
 {
-	register_qml_types();
 
 #ifdef SUBSURFACE_MOBILE
 	QQmlApplicationEngine engine;


janiversen:
good catch. but why does it work on my mac?

neolit123:
> good catch. but why does it work on my mac?

we are instantiating the same sequence on OSX and Windows, so i don't have an explanation.
if you are saying it worked, i trust you. :)

sfuchs79:
Fixes the issue also for me. Thanks!

bstoeger:
What was wrong with the old code [register the types when we know that we need them in the class that needs them]? Seems the right thing to do as opposed to splattering class-internal things all over the code-base. Isn't that quite fundamentally the point of a constructor: Set the things up so that it works?

The more I see of QML the put off by it. And the problem is definitely not us using it "wrongly". :(

neolit123:
> What was wrong with the old code [register the types when we know that we need them in the class that needs them]?

the only benefit of register_qml_types() is that it keeps the registration in one place.
https://github.com/Subsurface-divelog/subsurface/commit/ab05fe3cf83526ea63764a6c6587735e8b959527
the old code in MapWidget was really doing the **register QML types when you need them**.

after that change what happens is:
a MapWidget instance is created in a MainWindow instance. MapWidget loads QML code in it's constructor and needs QML types registered before hand.
registering the QML types in run_ui() is too late as MainWindow is first instantiated in init_ui().

what can be changed here is to load the QML that MapWidget needs, only when MapWidget becomes visible, but let's not do that.

bstoeger:
Could you go the other direction: move these calls into the respective classes (`MapWidgetHelper`, `MapLocationModel` and `MapLocation`) and remove them from the application-global helpers? Or would that create a hen-and-egg problem for mobile, where QML can never access these classes?

janiversen:
it would at least complicate testing qml (with the new test qml). apart from that the other registrations have exactly the same problem since they are referred to as items in mainwindow.qml. The purpose of #1488 is to automate testing of the C++ qml interface something we lack.

they used to be in desktop/mapwidget.cpp which is kind of a strange place since we have subsurface/mapwidget

neolit123:
> they used to be in desktop/mapwidget.cpp which is kind of a strange place since we have subsurface/mapwidget

that's because only desktop-widgets/mapwidget.cpp needs to load QML from C++ and register QML types. on mobile, the map widget QML is loaded directly and only uses C++ as a backend.

i'm ok with all QML registration being in one location.

> Could you go the other direction: move these calls into the respective classes (MapWidgetHelper, MapLocationModel and MapLocation) and remove them from the application-global helpers? Or would that create a hen-and-egg problem for mobile, where QML can never access these classes?

i don't think that would work. the QML registration needs to happen before the mapwidget.qml is loaded.

given this works for me and @sfuchs79 , merging.

PR #1506, opened 2018-07-13
Make sure our libdivecomputer custom IO interfaces have sleep functions

torvalds:
When I switched over from our own custom IO implementation to the new
upstream custom IO model in libdivecomputer, I completely missed the
fact that the libdivecomputer custom IO model also does a custom _sleep_
function.

I'm not entirely sure what the point was, and it broke things even in
libdivecopmputer itself when some of the new sleep functions were
broken.

Anyway, we didn't export any sleep functions at all for the bluetooth,
BLE and FTDI cases, the the libdivecomputer code didn't fall back to any
sane default sleep implementation either, so the end result was no
sleeping at all.

Which didn't matter for most divecomputers.

But it seems like at least some OSTC dive computers did care, at least
in certain situations, and both Miika and Anton had trouble downloading
with their OSTC Sport dive computers.  Using the serial line protocol
and the legacy /dev/rfcomm model worked fine, because then it used the
sleeping functions in the POSIX serial code inside libdivecomputer.

This just adds trivial sleeping functions for the affected download
protocols.  Maybe I should have just made libdivecomputer have a sane
default instead, but this wasn't hard either (the hard part was trying
to figure out why the downloads worked for some people and not for
others).

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/qtserialbluetooth.cpp
+++ b/core/qtserialbluetooth.cpp
@@ -6,6 +6,7 @@
 #include <QEventLoop>
 #include <QTimer>
 #include <QDebug>
+#include <QThread>
 
 #include <libdivecomputer/version.h>
 #include <libdivecomputer/context.h>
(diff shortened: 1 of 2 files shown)


dirkhh:
Derek set milestone: 4.8

dirkhh:
Thanks Linus, waiting for Travis (just because that's the right thing to do) and then I'll merge it.

dirkhh:
This is so strange. Travis status shows '0 builds backlog' for Linux, Linux container builds, and Mac builds. Yet this has been sitting here for 20+ minutes and none of the builds have started...

janiversen:
Wonder are virtual machines entitled to sleep, weekend and vacation :-)

PR #1508, opened 2018-07-13
Disable the WindowContextHelpButtonHint for Qt >=5.10

sfuchs79:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Disables the WindowContextHelpButtonHint by default on Qt::Sheet and
Qt::Dialog widgets. This hides the ? button on Windows, which only
makes sense if you use QWhatsThis functionality.
This value has been added in Qt 5.10.

### Mentions:

@neolit123

### DIFF
--- a/core/qt-init.cpp
+++ b/core/qt-init.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0
 #include <QApplication>
+#include <Qt>
 #include <QNetworkProxy>
 #include <QLibraryInfo>
 #include <QTextCodec>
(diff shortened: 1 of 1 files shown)


neolit123:
i think this is OK,
for older version we could have subclassed our own windows and dialogs and used `Qt::WindowContextHelpButtonHint`, but we never got around to doing that.

neolit123 (approved):

dirkhh:
Since I will build the Mac binaries with Qt 5.11 (but Windows binaries with 5.9) I'll take this one for 4.8.1

PR #1509, opened 2018-07-13
[RFC] Prepare for video thumbnails

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This prepares for video thumbnails. What this basically does is move the thumbnail-writing down in the call chain to where the thumbnails are created and defines a rather general format for thumbnails, which supports multiple images at different locations (@atdotde: this might be interesting to you - are you fine with this format?).

Here's the commit message of the relevant commit:
Video thumbnails are more complex than simple picture thumbnails.
We store a duration and might want to store multiple images.

Therefore, refactor the thumbnailing in imagedownloader.cpp. Move
the thumbnail-writing down in the call chain to where the thumbnails
are created, since we have more information there (i.e. could we
parse the file but not extract an image, etc.).

Split the write-to-cache function into three versions:
  - pictures
  - videos
  - unknown

Define the video-thumbnail on-disk format as
  - uint32  MEDIATYPE_VIDEO
  - uint32  duration of video in seconds
  - uint32  number of pictures
  for each picture:
        - uint32  offset in msec from begining of video
        - QImage  frame

Currently, we write 0 pictures. This will be filled in newer commits.

### Changes made:

1) Move down thumbnail-writing in the call-chain to where thumbnails are generated.
2) Split thumbnail-writing functions into 3 versions (pictures, videos, unknown).
3) Define an on-disk format for video-thumbnails.
### Related issues:

#1494 #1486
### Release note:

When this becomes ready.
### Documentation change:

Needs a general change mentioning videos
### Mentions:

@atdotde

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 - Dive media: implement "open folder of selected media files" [#1514]
+- Dive media: show play-time of videos on the profile plot
+- Dive media: draw thumbnails on top of deco-ceilings
 - Dive media: Change term "photos" and "images" everywhere in the UI to "media files"
 - mobile: show developer menu is now saved on disk and remebered between start of Subsurface
 - tests: add qml test harness
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -93,40 +93,36 @@ Thumbnailer::Thumbnail Thumbnailer::fetchImage(const QString &filename, const QS
 		mediatype_t type = get_metadata(qPrintable(filename), &md);
 
 		// For io error or video, return early with the appropriate dummy-icon.
-		if (type == MEDIATYPE_IO_ERROR) {
-			return { failImage, MEDIATYPE_IO_ERROR };
-		} else if (type == MEDIATYPE_VIDEO) {
-			addVideoThumbnailToCache(originalFilename, md.duration);
-			return { videoImage, MEDIATYPE_VIDEO };
-		}
+		if (type == MEDIATYPE_IO_ERROR)
+			return { failImage, MEDIATYPE_IO_ERROR, 0 };
+		else if (type == MEDIATYPE_VIDEO)
+			return addVideoThumbnailToCache(originalFilename, md.duration);
 
 		// Try if Qt can parse this image. If it does, use this as a thumbnail.
 		QImage thumb(filename);
 		if (!thumb.isNull()) {
-			addPictureThumbnailToCache(originalFilename, thumb);
-			return { thumb, MEDIATYPE_PICTURE };
+			int size = maxThumbnailSize();
+			thumb = thumb.scaled(size, size, Qt::KeepAspectRatio);
+			return addPictureThumbnailToCache(originalFilename, thumb);
 		}
 
 		// Neither our code, nor Qt could determine the type of this object from looking at the data.
 		// Try to check for a video-file extension. Since we couldn't parse the video file,
 		// we pass 0 as the duration.
-		if (hasVideoFileExtension(filename)) {
-			addVideoThumbnailToCache(originalFilename, {0} );
-			return { videoImage, MEDIATYPE_VIDEO };
-		}
+		if (hasVideoFileExtension(filename))
+			return addVideoThumbnailToCache(originalFilename, {0} );
 
 		// Give up: we simply couldn't determine what this thing is.
 		// But since we managed to read this file, mark this file in the cache as unknown.
-		addUnknownThumbnailToCache(originalFilename);
-		return { unknownImage, MEDIATYPE_UNKNOWN };
+		return addUnknownThumbnailToCache(originalFilename);
 	} else if (tryDownload) {
 		// This has to be done in UI main thread, because QNetworkManager refuses
 		// to treat requests from other threads. invokeMethod() is Qt's way of calling a
 		// function in a different thread, namely the thread the called object is associated to.
 		QMetaObject::invokeMethod(ImageDownloader::instance(), "load", Qt::AutoConnection, Q_ARG(QUrl, url), Q_ARG(QString, originalFilename));
-		return { QImage(), MEDIATYPE_STILL_LOADING };
+		return { QImage(), MEDIATYPE_STILL_LOADING, 0 };
 	}
-	return { QImage(), MEDIATYPE_IO_ERROR };
+	return { QImage(), MEDIATYPE_IO_ERROR, 0 };
 }
 
 // Fetch a picture based on its original filename. If there is a translated filename (obtained either
@@ -142,7 +138,7 @@ Thumbnailer::Thumbnail Thumbnailer::getHashedImage(const QString &filename, bool
 	// If there is a translated filename, try that first.
 	// Note that we set the default type to io-error, so that if we didn't try
 	// the local filename first, we will load the file from the canonical filename.
-	Thumbnail thumbnail { QImage(), MEDIATYPE_IO_ERROR };
+	Thumbnail thumbnail { QImage(), MEDIATYPE_IO_ERROR, 0 };
 	if (localFilename != filename)
 		thumbnail = fetchImage(localFilename, filename, tryDownload);
 
@@ -199,7 +195,9 @@ Thumbnailer::Thumbnail Thumbnailer::getVideoThumbnailFromStream(QDataStream &str
 
 	// If reading did not succeed, schedule for recalculation - this thumbnail might
 	// have been written by an older version, which couldn't extract the duration.
-	if (stream.status() != QDataStream::Ok)
+	// Likewise test the duration and number of pictures for sanity (no videos longer than 10 h,
+	// no more than 10000 pictures).
+	if (stream.status() != QDataStream::Ok || duration > 36000 || numPics > 10000)
 		return { QImage(), MEDIATYPE_VIDEO, 0 };
 
 	// Currently, we support only one picture
@@ -211,7 +209,7 @@ Thumbnailer::Thumbnail Thumbnailer::getVideoThumbnailFromStream(QDataStream &str
 	}
 
 	// No picture -> show dummy-icon
-	return { res.isNull() ? videoImage : res, MEDIATYPE_VIDEO, {(int32_t)duration} };
+	return { res.isNull() ? videoImage : res, MEDIATYPE_VIDEO, (int32_t)duration };
 }
 
 // Fetch a thumbnail from cache.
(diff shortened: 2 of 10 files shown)


bstoeger:
I added a proof-of-concept commit, which adds a bar above video-thumbnails representing the video's runtime during the dive.

This needs some styling. Colors are in-your-face (a.k.a. programmer-colors) and on zooming it has horrible behavior.

The z-value is currently messed up (bars below other videos, etc.), but there's no point in fixing this before #1516 is in.

Currently, the video-duration is stored with the thumbnail. I.e. the bar appears with a small delay. We might also save the duration in the log, as we used to do with the hashes (@dirkhh?).

dirkhh:
The purpose of saving the duration to the log file would be to be able to draw this duration bar without access to the video file, correct? This seems reasonable - can this be done as an attribute of the "picture" entry?

bstoeger:
> The purpose of saving the duration to the log file would be to be able to draw this duration bar without access to the video file, correct?

Two advantages:
1) Snappier - bars appear immediately and not only when the thumbnail is fetched from disk. This may also improve placing of the thumbnails if there are overlapping videos / pictures. I'm sure there are people diving with multiple cameras. :)
2) When opening the log on a different computer without transferring the media-files, the bar is shown.

Disadvantage:
Added complexity. For example, if we find out during thumbnailing that the duration changed, we also have to change to corresponding picture-entry in the backend.

> This seems reasonable - can this be done as an attribute of the "picture" entry?

Yes, of course. This should be forwards- and backwards-compatible. But I would add this later. First, we have to decide if we even want that and then polish the UI.

bstoeger:
I fixed the zoom problem. Code is not nice.

Color scheme is still horrible and the z-value messed up (waiting for #1516).

Any opinions? @sfuchs79?

sfuchs79:
@bstoeger 
- I like the idea
- Can't say that much about the code
- The color topic is err... difficult. I tend to say that red is not the right color for the bar but what is the right color?
- Test results:
My starting point is that I already have dives with videos in my log. What happens is that if I select such a dive for the first time, a incorrect (zero duration) bar is displayed for all the videos. I have to do a "recalculate thumbnails" for each video then. Also then the bar is not updated correctly but disappears completely. I then have to unselect and reselect the dive to show the correct bar. Don't know if this is expected behavior at the moment but this needs to be improved. For any "old" video with missing or old thumbnail format the thumbnail plus side information needs to be recreated immediately and the bar needs to be updated immediately.

bstoeger:
> The color topic is err... difficult. I tend to say that red is not the right color for the bar but what is the right color?

That's why I made these easily configurable for anyone to experiment:
```
static const double unscaledDurationLineWidth = 4.0;
static const double unscaledDurationLinePenWidth = 0.5;
static const QColor durationLineFillColor = Qt::red;
static const QColor durationLinePenColor = Qt::black;
```

> What happens is that if I select such a dive for the first time, a incorrect (zero duration) bar is displayed for all the videos. I have to do a "recalculate thumbnails" for each video then.

That should be fixed now. The missing bar thing might be that the bar is actually behind the thumbnail (the z-value problem).

What I noted though is that thumbnailing of normal pictures for some reason became *extremely* slow. I will try to investigate.

bstoeger:
> What I noted though is that thumbnailing of normal pictures for some reason became extremely slow. I will try to investigate.

Found the culprit - the new code wasn't down-scaling thumbnails anymore, so they were written with full size. That's of course something to avoid. Fixed.

sfuchs79:
I tested again. Part of the issue I reported is gone. But the part in bold is still there:
**My starting point is that I already have dives with videos in my log. What happens is that if I select such a dive for the first time, a incorrect (zero duration) bar is displayed for all the videos. I have to do a "recalculate thumbnails" for each video then.** Also then the bar is not updated correctly but disappears completely. I then have to unselect and reselect the dive to show the correct bar. Don't know if this is expected behavior at the moment but this needs to be improved. For any "old" video with missing or old thumbnail format the thumbnail plus side information needs to be recreated immediately and the bar needs to be updated immediately.

But I think I can guess what is not taken into account. I did use a build including #1486 before. So I already have "real" ffmpeg generated thumbnails for my videos. But these old style thumbnails are w/o duration information and have different data format in general. Am I right?
If yes, I don't know if you need to capture this case because it will only hit "power testers" like me during the current testing phase ;-)

sfuchs79:
Two more bugs found:
- On disabling display of media in profile via button bar the new duration bars stay visible
grafik

- When calculated deco ceiling and/or deco ceiling from DC is displayed and media icons overlay this area and then media icons are enlarged on mouse over afterwards media icons are behind deco ceiling  (potentially an old bug?!)
Does #1516 also fix this because it is similar?
Correct:
grafik
Incorrect:
grafik

bstoeger:
> But I think I can guess what is not taken into account. I did use a build including #1486 before. So I already have "real" ffmpeg generated thumbnails for my videos. But these old style thumbnails are w/o duration information and have different data format in general. Am I right?

Yes that explains it - in this case the first bytes of the old thumbnail would be read as the duration which would give non-sensical durations. I don't think it's worth checking for that situation. Ultimately I will rebase #1486 onto the new format. I added two sanity checks that maybe detect your "broken" thumbnails.

bstoeger:
> On disabling display of media in profile via button bar the new duration bars stay visible.

Fixed.

> When calculated deco ceiling and/or deco ceiling from DC is displayed and media icons overlay this area and then media icons are enlarged on mouse over afterwards media icons are behind deco ceiling (potentially an old bug?!)
Does #1516 also fix this because it is similar?

#1516 will not fix it, but it will make it possible to fix it. :) The way I read the code, the thumbnails were always behind the ceiling?

It's just a matter of setting the correct `zValue` for all the things on the profile.

sfuchs79:
>> When calculated deco ceiling and/or deco ceiling from DC is displayed and media icons overlay this area and then media icons are enlarged on mouse over afterwards media icons are behind deco ceiling (potentially an old bug?!)
Does #1516 also fix this because it is similar?

> #1516 will not fix it, but it will make it possible to fix it. :) The way I read the code, the thumbnails were always behind the ceiling?

Hmm, that sounds surprising to me. I thought thumbnails were always in front of the ceiling? But honestly I don't know exactly. What do we want? If I look at the screenshots above I like the first one better:
grafik
If I want to focus on the deco stuff, I'll switch of the media icons... Just my opinion.

I need to get some sleep now. More testing tomorrow. Really very nice work! :-)

bstoeger:
> Hmm, that sounds surprising to me. I thought thumbnails were always in front of the ceiling? 

Maybe I'm reading it wrong, but to me it looks like the tissues are placed at `z=1..16`:
```
        for (int i = 0; i < 16; i++) {
                DiveCalculatedTissue *tissueItem = new DiveCalculatedTissue(this);
                setupItem(tissueItem, profileYAxis, DivePlotDataModel::TISSUE_1 + i, DivePlotDataModel::TIME, 1 + i);
```

And the thumbnails are placed at `z=0` (normal) or `z=5` (magnified).

I changed thumbnails to go to `z=100` (`z=105` magnified). Let's see how this works out. Ideally, all these z-values should be defined at a common place... At the moment this is a bit ... chaotic.

bstoeger:
...and I managed to mess up stacking again. :( Will have to fix that tomorrow as I still have some paper-work to do. :(

bstoeger:
Gah - I give up on `stackBefore()`. Either I'm too stupid or this doesn't behave the way it suggests it does. Therefore, I'm back to using the z-value, which is super-vexing, because it makes #1516 basically useless. :(

My status for this PR:
- UI fine tuning (color, placement, size, ...) is probably better if I leave it to others.
- Saving duration in the log: material for a different PR.

So unless more bugs are found or @atdotde dislikes the thumbnail on-disk format, this is done from my side. What we now need is a decision if this is the right way to proceed.

sfuchs79:
I did a new build and will continue testing during the next days.
Regarding UI fine tuning:
Instead of introducing a "new" color I would reuse the profile background color. BTW: Where or how do we define it?
For the moment I ended up with:
```
static const double unscaledDurationLineWidth = 2.5;
static const double unscaledDurationLinePenWidth = 0.5;
static const QColor durationLineFillColor = QColor("#F2F2E6");
static const QColor durationLinePenColor = Qt::black;
```
That looks like this:
grafik

bstoeger:
> I did a new build and will continue testing during the next days.

That's very much appreciated.

> Instead of introducing a "new" color I would reuse the profile background color. BTW: Where or how do we define it?

The colors are defined in `core/color.cpp`. And you are right - either we use colors from there or we should add the new colors there.

sfuchs79:
Ok, then my proposal would be translated into:
```
static const double unscaledDurationLineWidth = 2.5;
static const double unscaledDurationLinePenWidth = 0.5;
static const QColor durationLineFillColor = SPRINGWOOD1;
static const QColor durationLinePenColor = BLACK1;
```

bstoeger:
I adopted your values, but used `getColor(...)`. This will do the right thing for black&white prints.

dirkhh:
I see the progress - do we have consensus that this is what the three(?) of you want?

bstoeger:
> I see the progress - do we have consensus that this is what the three(?) of you want?

I can only speak for myself. This PR consists of two parts:
1) New on-disk format for video-thumbnails.
2) Indication of duration on the profile

I think the 1) is reasonably open for future developments. Especially since this is not yet cast in stone - we simply save "0" as the number of pictures. Thus, in subsequent commits we might still think about being more complex than only saving "time-offset + frame" for each picture.

Concerning the UI changes in 2) I'll leave that to others to judge.

sfuchs79:
@bstoeger 
Commit message not correct any more "red rectangle".

bstoeger:
> Commit message not correct any more "red rectangle".

Good catch - That's easy to fix. :)

bstoeger:
Fixed the conflict (weirdly, when I rebased locally, there was no conflict [something something 3-way merge]).

Since nobody through their hands up in horror, I think this is good to go..? Finalization of the on-disk structure can be discussed in a subsequent commit. At the moment we're simply writing a "0" for no-pictures.

bstoeger:
I think potential naysayers had enough time by now? I don't think waiting much longer will improve things. Sorry for being impatient, but I'd really like to finish at least *one* feature. 😛

dirkhh:
I was badly distracted by the website translation rathole.
Let me read through this and comment and/or merge.

PR #1510, opened 2018-07-14
Dive media: transport dive-id in drag'n'drop events

bstoeger:
9efb56e2d43161d952efb444d1f13d87bfdd45b5 introduced rather complex
logic for picture drag'n'drop events onto the profile. Among other
things, the code had to check whether the picture actually belongs
to the displayed dive.

This can be simplified by transporting the dive-id in the drag'n'drop
event structure. The flow goes like this:
DivePictureModel--(1)-->DivePictureWidget--(2)-->ProfileWidget

For (1), we can use the Qt::UserRole role. This was used to transport
the picture-offset, but this is not needed anymore since ProfileWidget
was decoupled from DivePictureModel.

For (2), we simply replace the "position" value, which was never used.
Why would the receiver care which pixel was pressed in the media-tab?

This commit also contains a minor cleanup in DivePictureWidget:
QListView::mousePressEvent(event) was called in both branches of an
if and can therefore be removed from the if. This is so trivial,
that it doesn't warrant its own commit.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a rather small cleanup to the drag'n'drop changes in #1483. On drag & drop we should not only send the picture-filename, but also the dive it belongs to. Indeed, we should do this generally, to avoid confusion with images associated to different dives.

### Changes made:

1) Transport dive-id on image drag'n'drop events.
2) Drop drag'n'drop events of images belonging to other dives immediately.

### Release note:

No user-visible change (hopefully)
### Documentation change:

No user-visible change (hopefully)

### DIFF
--- a/desktop-widgets/divepicturewidget.cpp
+++ b/desktop-widgets/divepicturewidget.cpp
@@ -32,7 +32,9 @@ void DivePictureWidget::mouseDoubleClickEvent(QMouseEvent *event)
 void DivePictureWidget::mousePressEvent(QMouseEvent *event)
 {
 	if (event->button() == Qt::LeftButton && event->modifiers() == Qt::NoModifier) {
-		QString filename = model()->data(indexAt(event->pos()), Qt::DisplayPropertyRole).toString();
+		QModelIndex index = indexAt(event->pos());
+		QString filename = model()->data(index, Qt::DisplayPropertyRole).toString();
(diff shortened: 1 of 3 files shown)


dirkhh:
Reading this three times makes me at least think that I understand it. :-/

PR #1515, opened 2018-07-15
core: add qPrefCloudStorage

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Following qPrefDisplay and qPrefAnimations are qPrefCloudStorage, same schema

There are several special test cases to test e.g. that newpassword is not saved on disk.

remark qmlprefs and the qml files are not changed in this PR, as the preference is to deal with qmlPrefs in a single PR.

### DIFF
--- a/core/settings/qPref.cpp
+++ b/core/settings/qPref.cpp
@@ -15,6 +15,7 @@ static qPref *self = new qPref;
 
 void qPref::loadSync(bool doSync)
 {
+	qPrefCloudStorage::instance()->loadSync(doSync);
 	qPrefDisplay::instance()->loadSync(doSync);
 }
 
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -101,6 +101,7 @@ set(SUBSURFACE_CORE_LIB_SRCS
 	# classes to manage struct preferences for QWidget and QML
 	settings/qPref.cpp
 	settings/qPrefAnimations.cpp
+	settings/qPrefCloudStorage.cpp
 	settings/qPrefDisplay.cpp
 	settings/qPrefPrivate.cpp
 
--- a/core/cloudstorage.cpp
+++ b/core/cloudstorage.cpp
@@ -3,7 +3,6 @@
 #include "pref.h"
 #include "qthelper.h"
 #include "settings/qPref.h"
-#include "core/subsurface-qt/SettingsObjectWrapper.h"
 #include <QApplication>
 
 CloudStorageAuthenticate::CloudStorageAuthenticate(QObject *parent) :
@@ -49,10 +48,10 @@ void CloudStorageAuthenticate::uploadFinished()
 
 	QString cloudAuthReply(reply->readAll());
 	qDebug() << "Completed connection with cloud storage backend, response" << cloudAuthReply;
-	CloudStorageSettings csSettings(parent());
+	qPrefCloudStorage csSettings(parent());
 
 	if (cloudAuthReply == QLatin1String("[VERIFIED]") || cloudAuthReply == QLatin1String("[OK]")) {
-		csSettings.setVerificationStatus(qPref::CS_VERIFIED);
+		csSettings.set_cloud_verification_status(qPref::CS_VERIFIED);
 		/* TODO: Move this to a correct place
 		NotificationWidget *nw = MainWindow::instance()->getNotificationWidget();
 		if (nw->getNotificationText() == myLastError)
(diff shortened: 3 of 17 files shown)


janiversen:
ready from my side to be merged, but depends on #1518 (in case of merge problems I will handle it, as soon as I see #1518 is on master)

dirkhh:
So the first few commits here are the same commits as the ones that you posted as #1518 
I commented on the font name there - other than that I think that one might be ok. Now I need to look at the OTHER commits here to see what's up with them - but I'm not sure how you want to proceed... drop #1518 and just clean up this one? Or get #1518 in first and then try again for this one without the duplicate commits?

core/settings/qPref.cpp
```diff
@@ -3,6 +3,8 @@
 #include "qPref.h"
 #include "ssrf-version.h"
 
+QSettings qPref_setting;
+
```
dirkhh: the commit message does NOT describe what you are doing in this commit

dirkhh:
The commit messages are in part wrong (as in, they don't describe what the commit does), and many are poorly formatted (subject line too long, hard to read, missing empty line before SOB.

I can manually pull commits like this and clean them up - but first the questions in my earlier comment need to be addressed

janiversen:
why the need to CAPITALIZE “other commits” this PR depends on #1518, so of course it contains “a version” of #1518 in order to get green light from Travis. If you see that as a problem, simply delay review until it is cleaned.

I thought I wrote how to handle the 2 commits, that this PR depend on #1518, and once #1518is on master I will rebase this one.

I will edit the commits in a moment, but it would for sure be nice to have some documented rules, a while ago my code lines used too much vertical space, and now the commit line is too long. I normally use a 70char soft limit and 80char hard limit for both code and commit lines, but now I have a Q_PROPERTY line with more than 200chars (but it save vertical space).

core/settings/qPref.cpp
janiversen: as far as I can see (maybe wrong) this is not part of the CloudStorage PR, but the included dependency (which will be removed once #1518 is on master). I believe this is lobg gone in #1518.

dirkhh:
capitalization is a way to express stress as one would in spoken words. What I was trying to express was that I had fallen behind and wanted to make sure I look at all of your outstanding PRs.

Typically if you have a series of commits that all build on top of each other you have them in one PR. Having one PR include all the commits that are in another PR is a bit confusing for reviewers, I think.

Rules for code are different from rules for commit messages - commit subjects should be fewer than 70 chars, otherwise in a few tools you get weird wrapping issues (including GitHub). The body of a commit message should generally stay below 78 chars (but if you are quoting code sometimes you have to violate this - still trying to stay within this limit makes things easier to read).

The comments about vertical space in code vs line length... you have been a developer for a long time... it puzzles me that you are so combative about something that has been an issue with no simple answer for at least 40 years in software development.
Humans read code on screens. It used to be those screens were 80 chars wide, so that's where that obsolete limit came from. Very few people still use this line length - and as a result we end up using longer variable names and function names which make code easier to read.
200 chars is VERY long. But depending on context may be the right thing to do. Typically I try to stay under 120, but there are many cases where this creates code that's harder to read, so I break the rules.

If you tell me that you need a hard rule and that every time we have a situation where feedback about your choices makes you combative... I don't know.

Your responses to the back and forth on PRs always make me wonder if you assume an amount of aggression and forcefulness that isn't intended. What I write in comments are my observations when reading the commits. Suggestions how to make them better. I really don't know why it feels to me that at least once a week we miscommunicate so badly that I end up writing long explanations of why I said what I said...

janiversen:
I did have a big PR, with all the qPref classes, but was asked to break it up, now I try to functional grouped PR, but of course I want to test newer work, while waiting for a PR to be merged.

What I will in future is  mark any PR, that are ready but depend in a not merged PR as “doNotMerge” (I use WIP, until I am satisfied).

I am not combating the change if QPR0PERTY, you did not like it nd asked me to change my style and I did it. I am trying to have clear rules, so I can add it in a githook and have it automatically controlled.

janiversen:
Linux2 is unhappy with sonethung that seems unrelelated to this patch. 

This PR is merge ready from my side.

dirkhh:
Sorry @janiversen - I was distracted for a few days and apparently none of the other maintainers dared touch this :-)
I think this is coming along nicely, let's keep all this work moving forward so we have a cleaner QML-C++ interface.
Thanks

PR #1516, opened 2018-07-15
Dive media: fix stacking of thumbnails in profile plot

bstoeger:
9efb56e2d43161d952efb444d1f13d87bfdd45b5 added code to rearrange
pictures on drag'n'drop. To keep the correct repainting order (i.e.
later thumbnails are painted on top of previous thumbnails), the
z-value was set accordingly. In principle a sound idea, but it did
not work out, because on hoverEnter and hoverExit events, the z-values
were modified, thus restoring the wrong drawing order.

Instead, use the QGraphicsItem::packBefore() function. Experimentation
showed that this only works if stacking is done from the back
of the list.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a UI bug introduced in 9efb56e2d43161d952efb444d1f13d87bfdd45b5, which would lead to thumbnails drawn in the wrong order if they were rearranged by drag'n'drop and then magnified by hover-events.

### Changes made:

1) Use `QGraphicsItem::packBefore()` to control drawing order.

### Release note:

None needed - fixes a bug in current release-cycle.

### Documentation change:

None needed, as this is a bug fix.

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -2103,17 +2103,21 @@ bool ProfileWidget2::PictureEntry::operator< (const PictureEntry &e) const
 
 // Calculate the y-coordinates of the thumbnails, which are supposed to be sorted by x-coordinate.
 // This will also change the order in which the thumbnails are painted, to avoid weird effects,
-// when items are added later to the scene. This is simply done by increasing the Z-value.
+// when items are added later to the scene. This is done using the QGraphicsItem::packBefore() function.
+// We can't use the z-value, because that will be modified on hoverEnter and hoverExit events.
 void ProfileWidget2::calculatePictureYPositions()
 {
-	double lastX = -1.0, lastY;
-	double z = 0.0;
-	for (PictureEntry &e: pictures) {
-		if (!e.thumbnail)
+	// Quit early if there are no items. The last loop in this function assumes that the vector is not empty.
+	if (pictures.empty())
+		return;
+
+	double lastX = -1.0, lastY = 0.0;
+	for (auto it = pictures.begin(); it != pictures.end(); ++it) {
+		if (!it->thumbnail)
 			continue;
 		// let's put the picture at the correct time, but at a fixed "depth" on the profile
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
looks reasonable - I haven't tested it, though

profile-widget/profilewidget2.cpp
```diff
@@ -2103,28 +2103,40 @@ bool ProfileWidget2::PictureEntry::operator< (const PictureEntry &e) const
...
-// when items are added later to the scene. This is simply done by increasing the Z-value.
+// when items are added later to the scene. This is done using the QGraphicsItem::packBefore() function.
+// We can't use the z-value, because that will be modified on hoverEnter and hoverExit events.
 void ProfileWidget2::calculatePictureYPositions()
 {
+	// Quit early if there are no items. The last loop in this function assumes that the vector is not empty.
+	if (pictures.empty())
+		return;
+
 	double lastX = -1.0, lastY;
```
sfuchs79: @bstoeger 
Already with the old code but still with the new code there is a warning about lastY:
/home/stefan/Entwicklung/Subsurface/subsurface/profile-widget/profilewidget2.cpp:2123:17: warning: 'lastY' may be used uninitialized in this function [-Wmaybe-uninitialized]
    y = lastY + 3;
Maybe you could silence this with this change?

bstoeger:
> Already with the old code but still with the new code there is a warning about lastY

Yes, that's trivial to fix (`lastY = 0.0;`). But this is a false positive and I'm not sure what the policy is in this regard. I guess in this case it's so trivial that we should just silence the compiler.

profile-widget/profilewidget2.cpp
neolit123: yes, please initialize it to `0.0`.

bstoeger: Ok - stupid compiler. :(

neolit123:
seems good :+1: 
waiting on CI.

PR #1519, opened 2018-07-17
Cleanups to core/color.[c|hpp]

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are - admittedly gratuitous - cleanups. But using the heavy `QVector<>` for what is a static array just rubbed my the wrong way. At least my executable shrunk a bit, even if it's only a drop in the bucket.

### Changes made:

Just repeating the commit-message here:
1) Make the profile_color map of static linkage - it is not
used outside of this file.

2) Remove the third color, which originally was planned for
printing. It was not accessed anywhere.

3) Replace QVector<QColor> by std::array<QColor, 2>. Using a
reference-counted, copy-on-write, dynamic container for static
data seems like overkill. std::array<QColor, 2> has exactly the
same run-time impact as QColor[2], but allows for assignment.

4) Use brace-initialization and remove the unneeded COLOR macro.

5) Remove the fill_profile_color function. Simply use static
initialization.

6) Move #includes from .h to .cpp file.

7) Remove text_render_options(_t), which were not used anywhere.

### Release note:

No.
### Documentation change:

No.


bstoeger:
PS: I would even consider turning the `QMap<>` into a plain C-array, but that would of course mean being careful about having the same order in the `enum` and the array.

PR #1520, opened 2018-07-17
Cleanup: rename dive_getUniqID() and handle double ids gracefully

bstoeger:
dive_getUniqID() actually *set* the unique id of a dive. Therefore,
rename this function to dive_setUniqID(). Weirdly, all callers of the
function used the pattern
	dive.id = dive_getUniqID(&dive);
so the dive-id was assigned twice, once inside and once outside the
function. Change this to
	dive_setUniqId(&dive)
and don't return the id. It can be read from dive.id anyway.

Finally, fix a small inconsistency:
If a dive already had an id, this id would have been registered and
a debug message written. Creation of an already-registered id would
on the other hand raise an assert(). But the latter can only happen
as a consequence of the former - so why be once graceful and once
terminate? Therefore, make both cases graceful.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is just a (very) minor cleanup, but I think the new code is more logical. Seeing something like:
```
        dive.id = dive_getUniqID(&dive);
```
you probably wouldn't have guessed that the function actually *sets* the id by itself. Changing this to
```
        dive_setUniqID(&dive);
```
makes it much more clear what happens.
 
### Changes made:

1) rename function `dive_getUniqID()` to `dive_setUniqID`.
2) don't return id in `dive_setUniqID`.
3) consistently handle double-ids gracefully.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -462,7 +462,7 @@ struct dive *alloc_dive(void)
 	if (!dive)
 		exit(1);
 	memset(dive, 0, sizeof(*dive));
-	dive->id = dive_getUniqID(dive);
+	dive_setUniqID(dive);
 	return dive;
 }
 
@@ -1784,7 +1784,7 @@ struct dive *fixup_dive(struct dive *dive)
 	/* we should always have a uniq ID as that gets assigned during alloc_dive(),
 	 * but we want to make sure... */
 	if (!dive->id)
-		dive->id = dive_getUniqID(dive);
+		dive_setUniqID(dive);
 
 	return dive;
 }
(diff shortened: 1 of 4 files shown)


bstoeger:
Actually - forget that. In all three callers the id is either
1) reset to 0
2) checked to be 0

immediately before the call. The whole `QSet<>` thing makes no sense. Let's just remove it completely.

PR #1521, opened 2018-07-17
Cleanup: simplify dive_getUniqID()

bstoeger:
dive_getUniqID() is used to create unique dive ids, which are
stable during application lifetime. It was passed a dive, checked
that the id was not set (if it was that it is know to the application)
and set a new id (in contradiction to its name!) if it hadn't any.

There were three callers:

alloc_dive(): called the function on a zeroed dive struct.
fixup_dive(): called the function only if the dive had a 0 id.
MainWindow::setupForAddAndPlan(): called the function on a zeroed dive
struct.

Thus, in all three callers the id is guaranteed to be zero and
the whole keeping-track-of-ids logic is moot. Remove the logic,
don't pass a dive struct to dive_getUniqID() and move the function
to the C-backend.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Minor cleanup: remove the whole `dive_getUniqID()` logic, since all callers guaranteed to not have an id set. The dives where either cleared or tested for 0-id right before the call. This function might have made sense previously, but now there seems to be no point. For once, move something from C++ to C. :)
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -454,6 +454,15 @@ double get_weight_units(unsigned int grams, int *frac, const char **units)
 	return value;
 }
 
+// we need this to be uniq. oh, and it has no meaning whatsoever
+// - that's why we have the silly initial number and increment by 3 :-)
+int dive_getUniqID()
+{
+	static int maxId = 83529;
+	maxId += 3;
+	return maxId;
+}
+
 struct dive *alloc_dive(void)
 {
 	struct dive *dive;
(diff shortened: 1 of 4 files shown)


dirkhh:
I always love to say this... "Well, actually..." :-)
So the way my retarded brain works, this does actually get you an ID. If it were a setID function, then you'd pass in the ID, right?
But with that idiotic trivia out of the way, I think you are spot on and this is a great simplification of the code.
Let's on principle wait for Travis, but from my POV this is good to go.

bstoeger:
NB: I think you comment on the version that I trashed? Because the new version, which I prefer, stays with the `dive_getUniqId()` name.

dirkhh:
Yes, this was just a comment about what you had said about what the function does - and that was in the other PR, now that I think about it.
But my point remains. This is good.

PR #1523, opened 2018-07-18
Initial OpenBSD support

kristapsdz:
Add OpenBSD support.  Probably extends to all other BSDs as well.

Signed-off-by: Kristaps Dzonsons <kristaps@bsd.lv>

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a trivial patch to add OpenBSD support to the codebase.  (It will probably work with NetBSD and FreeBSD as well, but that remains untested.)  With this, I can actually build and run the entire system.  (`build.sh` needs to be patched as well, but that's not quite ready yet.)  Alternatively, `linux.c` can also have an ifdef around the /proc filesystem parts and can be compiled for both.

### Changes made:

1) clone linux.c with specific changes for openbsd


dirkhh:
Hi @kristapsdz 
OpenBSD is not a platform we had requests for before :-)
As our GitHub bot already pointed out, please add a Signed-off-by: to your commits - that's a requirement for contributions to Subsurface. Other than this I believe this looks fine.

kristapsdz:
@dirkhh , done!  I've been using Subsurface on OpenBSD for a while for technical dive planning and as an intermediary for Shearwater Perdix AI -> subsurface-mobile -> cloud -> subsurface -> export as XML -> [divecmd](https://kristaps.bsd.lv/divecmd).  I'd like to pull it into OpenBSD's packages, but that shouldn't require any additional patches.

dirkhh:
We do the Signed-off-by in the commits, but now that you have provided it in the PR I can simply manually add them when I pull this. Trying to keep it simple for you :-)

glance-:
Might even be a idea to unify macos.c with that unix.c then. A quick vimdiff between linux.c and macos.c show's that they're quite similar.

dirkhh:
Let's do this in two steps.
@kristapsdz would you mind merging this into a unix.c or would you prefer if we merged this as is and did that in a separate step?
And then the merge with macos.c is something we can look at in the next round.

bstoeger:
From what I've seen there is only need for two `#ifdefs` for *BSD<->Linux. I think it would probably be easiest to just implement them, though perhaps I'm underestimating the complexity of introducing the appropriate `#define` or `-D` compiler switch.

MacOS in contrast differs more. So yes, let's wait with that.

dirkhh:
@kristapsdz did you plan to address the comments, or should we take what you submitted and then make those changes ourselves?

kristapsdz:
@dirkhh, I'm on it!  (Almost time for weekend programming...) :)

dirkhh:
Thanks, no rush - I'm just trying to make sure things don't fall between the cracks... we had many cases where both we (or, frankly, I) and the submitter were waiting for the other to take the next step... so it always seems worth while to 'ping' on these topics

bstoeger (approved):

dirkhh:
I collapsed this into one commit and merged it manually.
Thanks @kristapsdz

PR #1528, opened 2018-07-19
Implement proper Qt-model semantics for DiveTripModel (and fix multi-level undo) [part 1/x]

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As I said in a previous PR, the undo-code is fundamentally broken. I see different ways of fixing that and went with the simpler method: keep pointers to dives and trips stable. This is done by undo-objects taking ownership of the objects. Note that crashes can still be provoked when dives are deleted in other parts of the code (notably split/merge/load). But that should be rather easy to fix. In any case, the new code would need testing:
- delete multiple dives / remove dives from trips
- undo/redo repeatedly
- undo/delete other dives/undo again
- etc..

So far it worked reasonably for me.

Note that the code uses C++ concepts, notably the owning `std::unique_ptr`. But this makes resource-management so much simpler that I'm convinced that it is worth it.

Sadly, this pretty much makes the last two commits of master moot. Could we - as an exception - remove them?

Commit message:
The original undo-code was fundamentally broken. Not only did id leak
resources (copied trips were never freed), it also kept references
to trips or dives that could be changed by other commands. Thus,
anything more than a single undo could lead to crashes.

Two ways of fixing this were considered
1) Don't store pointers, but unique dive-ids and trip-ids.
   Whereas such unique ids exist for dives, they would have to be
   implemented for trips.
2) Don't free objects in the backend.
   Instead, take ownership of deleted objects in the undo-object.
   Thus, all references in previous undo-objects are guaranteed to
   still exist (unless the objects are deleted elsewhere).

After some contemplation, the second method was chosen, because
it is significantly less intrusive. While touching the undo-objects,
clearly separate backend from ui-code, such that they can ultimately
be reused for mobile.

Note that if other parts of the code delete dives, crashes can still
be provoked. Notable examples are split/merge dives. These will have
to be fixed later. Nevertheless the new code is a significant
improvement over the old state.

### Changes made:

1) Create functions in backend that allow taking ownership of dives/trips
2) Take ownership of dives/trips in undo objects.
3) Initial support for undo of dive addition.

### Release note:

Will do once there is undo for split/merge.


bstoeger:
Puh - from what I've seen so far this will be quite an uphill battle. But I'm confident that it can be done and I believe that it will lead to untangling of the frontend/backend interface, because all the divelist-modifying manipulations are collected in one place.

My first goal is to get this right for all operations that create/destroy dives [add, delete, merge, split, import]. For loading of a new logbook I would simply destroy the undo-history for now [though we should keep multiple-document support in mind].

At the moment, the selection-handling is completely messed up, so don't be surprised if that does funky things.

bstoeger:
I made the comments more verbose - though perhaps this should be moved away from the source code to the documentation?

dirkhh:
I was quite distracted for a few days - hopefully I'll be able to look at the PRs this weekend.
I'm so glad that you are tackling this one. I remember way back when this was written that I worried about it - but then other things took over and I never went back...

bstoeger:
Comments on the overall design would of course be highly appreciated. But note that the details are not yet completely fleshed out, so an in-depth review would be premature.

My plan for this PR:
1) Implement all divelist-changing operations as undo-able commands. Only then can this at least *theoretically* work. This should also prove/disprove the viability of the approach.
2) Link with the divelist-model. At the moment we're completely regenerating the list on nearly every action. Quite an embarrassment.
3) Link with the dive-selection code. Currently, the commands that I implemented so far have completely unpredictable effect on the selection.

Mid-range to long-range plans from my side would then be:
1) Remove the undo code from the desktop-version, so that mobile can also use it. There is a lot of code-duplication that could be avoided.
2) Turn it into simple function-calls instead of object generations. I.e. hide implementation details.
3) Make editing operations also undo-able.

bstoeger:
> Would it be possible to ensure this can be used for the mobile version also (clear division between habdling and ui) ?

Yes, that's the plan (see point 1 of the mid-range plans above). I try to strictly separate backend / qt-model and desktop code.

bstoeger:
Please wait with testing - first I want to do the model<->command interaction. I plan to do this via signal (of which I'm not too big of a fan) so that it can be used from mobile and desktop.

Yes, the whole concept of this is to make pointers to `struct dive` and `struct trip` stable as long as there are references to them. This is done by owning them: either in the backend or in the undo-command. If a dive object is deleted, it means that it is a) unknown to the backend and b) no command references it.

dirkhh:
> **Please wait with testing** - first I want to do the model<->command interaction. I plan to do this via signal (of which I'm not too big of a fan) so that it can be used from mobile and desktop.

Hehe... the title of the PR and this comment are not exactly in agreement :-) :-)

bstoeger:
Yeah - that was for the original version which is obsolete. Changed.

glance-:
One random idea which might be worth exploring is actually using the git storage for undo/redo. That way we never need to worry about carefully unwinding and/or tracking changes, just save them, and load a older commit in case of undo.

bstoeger:
Yes, the idea is less crazy than it may seem at a first glance. I was contemplating about a git-inspired system, where two arbitrary divelogs could be related by a "diff". This would make it possible to have non-linear undo, i.e. different "branches". But I've given up on it as I thought it was way over my head.

As for using git directly, perhaps it can be done. The changes in this commit might even be the basis for such a thing, as all divelog-changing commands are isolated (and logic can then be transferred into the C-core if needed). I'm a bit reserved about reloading the whole divelog on every undo/redo. Currently, I'm going the opposite direction: don't reload the whole divelist on every single action, only the parts that changed. But I don't want to exclude the possibility that this is the wrong direction as nowadays everyone has multi-core GHz-clocked CPUs in their pockets. Why not even save every selection, every scrolling, in a git-repository for perfect replay?

In any case, I'll leave that to others as `git` and `libgit` are black-magic to me.

janiversen:
I believe the git idea have a lot of merit. It would really simplify the whole undo/redo process, and actually it is something on my long term plan,  I have already had to manually “play” with git commits, due to “interesting” conflicts when you leave ssrf open on the ipad and start working on a mac.

One relatively simple approach would be to have a branch where all changes (as in save/delete) would be stored and ssrf would not remember every single change.

example:
edit dive:
1. add location
2. add weight
3. save (and new commit to git branch)

If the user want to undo/redo things
1. present user with commits from local branch
2. present user eith extended choise as commits in master branch.

I think if we had that we could remove all undo/redo code.

my priority is currently (of course) to finish the rewrite of properties, but then this is something I could give a go, if we agree that is a simple, safe and sufficient solution.

janiversen:
ps. I also believe the current work that provides a separation between UI and process is impotant and should be added.

dirkhh:
Oh, using git as a means of undo is something that I looked into and started implementing way back when.
It's a wee bit heavy weight for this purpose, and there were a few conceptual problems I ran into that ended up derailing the idea. Which is not to say that it isn't worth re-visiting that idea.

janiversen:
Could you please highlight the conceptual problems? when limiting undo/redo to to save/delete I miss to see them.

dirkhh:
This is years ago - I haven't revisited this idea in since forever...
Let's see what I remember:
- Calling into the git-save code path after every transaction caused unexpected delays on larger dive data; I don't recall the specifics, but IIRC a renumber of a large dive log could cause a double digit second delay
- undo performance is roughly similar to file open performance - also unexpected
- This massively increased the commit count and since libgit2 doesn't support GC it increases the size of the repo at rest (and at initial transfer over the wire);

bstoeger:
Update:

The significant new part is the implementation of proper Qt-model semantics for the `DiveTripModel`. This is a double-monster commit. It is monstrous in size and monstrous in content [template and lambda madness!]. I tried it with classical for loops, but this ended up with pages-upon-pages of repeated code. I see two alternatives to templates and lambdas:
1) Macros.
2) Classical C-style callbacks with state transported via (void *) pointers.

But neither case seems more readable to me.

Before going into detail why the code is so complex, a disclaimer: The current state is known-to-be-broken in two aspects:
1) [Showstopper] Currently, the code does not modify the dive-selection. Therefore, it is quite easy to bring the application in an inconsistent state: Delete, merge or split dives and the old dive will be in the information window. You probably can crash the application rather easily in such a state.
2) [Borderline acceptable] The filters are not updated, if dives are deleted/merged, etc. The filters must be closed and reopened by two consecutive CTRL-F.

What this does is implement Qt-model semantics for the `DiveTripModel`. Currently, on every divelist-changing action, the whole model is rebuilt. What instead should be done is calling `beginInsertRows()`/`endInsertRows()` and related (delete, update) pairs. These functions work on sets of contiguous rows. But the user can choose arbitrary dives in an arbitrary number of trips *and* we can be in two modes: list-view and trip-view.

So the actual complexity arises from the fact that dives are batched twice: in the commands, they are batched by trip and sent via one signal per trip. In the `DiveTripModel`, they are batched according to contiguousness. This capitalizes on the fact that dives are always sorted by date. Now, this is by no means a difficult problem - stepping two lists and collect in batches. But the resulting code is verbose. Therefore, it is split-off in templates. One could, of course, go the easy route and add/delete/update in single dives. But since Qt *does* provide a batch-wise interface I think we should do our best to use it.

This has still potential for improvement: For example, instead of linear search, binary searches could be used, since dives are always ordered by date. Moreover, in the case of moving-between-trips and changing dive-date, I went the "cheating" way: Instead of using `beginMoveRows()`/`endMoveRows()` the dives are removed and re-added. This can likewise be improved at a later date.

So having the disclaimer above in mind, this can be tested / glimpsed at. Testing this fully is rather hard. It will involve performing a sequence of these actions:
```
void addDive(dive *d, bool autogroup);
void deleteDive(const QVector<struct dive*> &divesToDelete);
void shiftTime(const QVector<dive *> &changedDives, int amount);
void renumberDives(const QVector<QPair<int, int>> &divesToRenumber);
void removeDivesFromTrip(const QVector<dive *> &divesToRemove);
void removeAutogenTrips();
void addDivesToTrip(const QVector<dive *> &divesToAddIn, dive_trip *trip);
void createTrip(const QVector<dive *> &divesToAddIn);
void autogroupDives();
void mergeTrips(dive_trip *trip1, dive_trip *trip2);
void splitDives(dive *d, duration_t time);
void mergeDives(const QVector <dive *> &dives);
```
with arbitrary selections (contiguous, non-contiguous, spanning multiple-trips, top-level-dives) and in both views (list, trip) and then doing multiple CTRL-Z, CTRL-SHIFT-Z actions to undo/redo everything.

For me personally, the UI-experience is distinctly improved by not reloading the dive-list on every action.

Next short-term plans:
- Hook up with selection-machinery (though not yet sure where this belongs - command, model or view!?).
- Hook up with filters

Next medium to long-term plans:
- Extend command system to editing (including data, profile, media, ...)
- Extend undo to dive-import (from file, divecomputer)
- Implement in mobile

tcanabrava:
I really like the idea of dropping the dive id. Some background on the Original state of the code:
When I started to do this model (and this was one of the first bit of code on the c++ port, almost 5 years ago) I had to work with the Qt and the Gtk versions of it, as I could not break the code for Gtk while I was porting to Qt, and many of the operations would rearrange the dives in the dive_table c vector, making the pointer inconistent. I have not followed the code last year to be sure that this is not going to happen anymore, but it's something to be aware.
If I could I'd replace the dive_tale to a QList<Dive*> or similar, always making sure that the pointer is alive and well.

bstoeger:
I'm well aware of the difficult situation that you inherited and you did a great job. It's a small wonder that it works as flawless as it does. :)

I audited the code and there seems to be no unknown reallocation of `struct dive *` left, but of course I might have missed something.

desktop-widgets/undocommands.cpp
```diff
@@ -200,47 +254,90 @@ void UndoRenumberDives::redo()
...
+}
 
-	for (auto &pair: divesToAdd)
-		add_dive_to_trip(pair.first, pair.second);
-	divesToAdd.clear();
-	mark_divelist_changed(true);
+UndoAddDivesToTrip::UndoAddDivesToTrip(const QVector<dive *> &divesToAddIn, dive_trip *trip)
+{
+	setText(divesToAddIn.size() == 1 ? gettextFromC::tr("add dives to trip")
+					 : gettextFromC::tr("add %1 dives to trip").arg(divesToAddIn.size()));
```
dirkhh: that messes with the way plurals are handled in the translation. Please check http://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals
reading through other commits it's clear that you just followed the existing - bad - pattern.
It would be nice if you could fix it while you are touching the code - or we could do that later if that's distracting right now.

desktop-widgets/command_divelist.cpp
```diff
@@ -176,14 +182,19 @@ void UndoAddDive::undo()
...
 }
 
-UndoDeleteDive::UndoDeleteDive(const QVector<struct dive*> &divesToDeleteIn) : divesToDelete(divesToDeleteIn.toStdVector())
+DeleteDive::DeleteDive(const QVector<struct dive*> &divesToDeleteIn) : divesToDelete(divesToDeleteIn.toStdVector())
 {
 	setText(divesToDelete.size() > 1 ?
-			gettextFromC::tr("delete %1 dives").arg(divesToDelete.size()) :
-			gettextFromC::tr("delete dive"));
+			tr("delete %1 dives").arg(divesToDelete.size()) :
+			tr("delete dive"));
```
dirkhh: see my previous comment on plurals - I realize that this was wrong in the existing code as well. Shame on us.

desktop-widgets/command_divelist.cpp
```diff
@@ -225,21 +236,26 @@ void UndoShiftTime::undo()
...
 
 
-UndoRenumberDives::UndoRenumberDives(const QVector<QPair<int, int>> &divesToRenumberIn) : divesToRenumber(divesToRenumberIn)
+RenumberDives::RenumberDives(const QVector<QPair<int, int>> &divesToRenumberIn) : divesToRenumber(divesToRenumberIn)
 {
 	setText(divesToRenumber.count() > 1 ?
-			gettextFromC::tr("renumber %1 dives").arg(divesToRenumber.count()) :
-			gettextFromC::tr("renumber dive"));
+			tr("renumber %1 dives").arg(divesToRenumber.count()) :
+			tr("renumber dive"));
```
dirkhh: ditto

desktop-widgets/command_divelist.cpp
```diff
@@ -264,24 +290,24 @@ void UndoTripBase::redo()
...
 	redo();
 }
 
-UndoRemoveDivesFromTrip::UndoRemoveDivesFromTrip(const QVector<dive *> &divesToRemove)
+RemoveDivesFromTrip::RemoveDivesFromTrip(const QVector<dive *> &divesToRemove)
 {
-	setText(divesToRemove.size() == 1 ? gettextFromC::tr("remove dive from trip")
-					  : gettextFromC::tr("remove %1 dives from trip").arg(divesToRemove.size()));
+	setText(divesToRemove.size() == 1 ? tr("remove dive from trip")
+					  : tr("remove %1 dives from trip").arg(divesToRemove.size()));
```
dirkhh: wow, this bug was in many places... I wonder why I didn't catch it back when the code went in

desktop-widgets/command_divelist.cpp
```diff
@@ -291,17 +317,17 @@ UndoRemoveAutogenTrips::UndoRemoveAutogenTrips()
 	}
 }
 
-UndoAddDivesToTrip::UndoAddDivesToTrip(const QVector<dive *> &divesToAddIn, dive_trip *trip)
+AddDivesToTrip::AddDivesToTrip(const QVector<dive *> &divesToAddIn, dive_trip *trip)
 {
-	setText(divesToAddIn.size() == 1 ? gettextFromC::tr("add dives to trip")
-					 : gettextFromC::tr("add %1 dives to trip").arg(divesToAddIn.size()));
+	setText(divesToAddIn.size() == 1 ? tr("add dives to trip")
+					 : tr("add %1 dives to trip").arg(divesToAddIn.size()));
```
dirkhh: and more of the same

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "desktop-widgets/downloadfromdivecomputer.h"
+#include "core/qthelper.h"
+#include "desktop-widgets/mainwindow.h"
+#include "desktop-widgets/command.h"
```
dirkhh, line 3: given the one-liner change, it's surprising that you'd need all these new include files...

desktop-widgets/command_divelist.cpp
```diff
@@ -4,9 +4,41 @@
...
+// Input parameters:
+//	- dives: a vector of trip,dive pairs, which will be sorted and processed in batches by trip.
+//	- action: a function object, taking a trip-pointer and a QVector of dives, which will be called for each batch.
+template<typename Function>
+void processByTrip(std::vector<std::pair<dive_trip *, dive *>> &dives, Function action)
+{
+	// Use std::tie for lexicographical sorting of trip, then start-time
+	std::sort(dives.begin(), dives.end(),
+		  [](const std::pair<dive_trip *, dive *> &e1, const std::pair<dive_trip *, dive *> &e2)
+		  { return std::tie(e1.first, e1.second->when) < std::tie(e2.first, e2.second->when); });
```
dirkhh, line 25: eventually I'll figure out how to read code like this without wanting to run away.
After staring at it for a while, I think I know what it does. And I see why doing it this way makes the code smaller and easier to maintain (I guess?)

desktop-widgets/command_divelist.cpp
```diff
@@ -55,6 +89,22 @@ static std::vector<DiveToAdd> removeDives(std::vector<dive *> &divesToDelete)
...
 
+	// We send one dives-deleted signal per trip (see comments in DiveListNotifier.h).
+	// Therefore, collect all dives in a array and sort by trip.
+	std::vector<std::pair<dive_trip *, dive *>> dives;
+	dives.reserve(res.size());
+	for (const DiveToAdd &entry: res)
+		dives.push_back({ entry.trip, entry.dive.get() });
+
+	// Send signals.
+	processByTrip(dives, [&](dive_trip *trip, const QVector<dive *> &divesInTrip) {
```
dirkhh, line 129: sorry - syntax question... what does the `&` in the `[ ]` mean?
Googling tells me "captures all variables used in the lambda by reference" - that's great. What does that mean again?
Sorry :-(

dirkhh (commented):
This is my second round through these patches. This time I made some comments, most of them are not really related to your code and more related to things that were wrong before or things that I don't understand in fancy C++ 
I will keep going through this PR until I think that I understand it...
This definitely feels like it's going in the right direction.

desktop-widgets/command_divelist.cpp
bstoeger: The advantage of lambdas over function pointers is that they can have state. For this purpose, you feed it variables from the containing scope. These variables (=captures) are given in square brackets.

So, for example
```
auto fun = [when] (dive *d) { return dive->when < when; }
```
would be a function that takes a `dive *` as an input parameter and returns true whenever the start-time is before the local variable `when`:
```
	for_each_dive(i, dive) {
		if (fun(dive))
			return dive;
	}
```
 Obviously, it has to store the `when` somewhere, either by reference or by value. In the case of complex types, the former is more efficient [you don't have to copy the value]. But on the other hand, if the lambda has a longer lifetime, you might end up with dangling references, or changing values.

`[&]` and `[=]` are the lazy options: Don't specify all captured variables, simply say the all used variables are capture by reference (`[&]`) or by value (`[=]`). Hope that clears it up (somewhat).

dirkhh:
More reading, more playing with code. You people will turn me into a C++ developer eventually, I guess. So yeah, I think I understand this better now.

One request - given such a change, it would be really nice if this series came with test cases that ensure that this is actually correct and behaves as intended. Writing tests is annoying and boring, but to me the lack of test coverage is one of my bigger concerns for Subsurface.

bstoeger:
Before thinking about tests, I want to get this "feature-complete". This means fixing the whole selection-topic. At the moment I have a very fundamental problem with the way `QTreeView` and `QItemSelectionModel` work. Notably, there is no trivial way of distinguishing between user-initiated and programmatical selection changes. So far, this was handled with `disconnect`/`connect` pairs of internal(!) signals (c.f. `divelistview.cpp`). I'm exploring ways to circumvent this, but none of the options I came up with are appealing.

Moreover, I will be travelling the whole of August, so my responsiveness and productivity will be distinctly decreased. So don't hold your breath.

dirkhh:
Every good diver knows that you should NEVER hold your breath...
Yeah, the disconnect/connect mess is fun, isn't it? I tried to find ways to this differently for a while and so did @tcanabrava - we couldn't come up with anything.
Enjoy your travels

bstoeger:
> Yeah, the disconnect/connect mess is fun, isn't it? I tried to find ways to this differently for a while and so did @tcanabrava - we couldn't come up with anything.

I studied Qt's source and to me it looks like there is no better solution short of implementing our own ItemSelectionModel (which we might have to do sooner-or-later anyway). I just changed it to use a global-flag (command-executing) instead of `disconnect`/`connect` pairs.

The whole selection topic is more difficult than I anticipated and isolating all this in a proper model is probably the way to go, though too invasive for now. I put it on the long-term to-do list.

Notably, the interaction with the filter is tricky. For example:
- User deletes dives.
- User changes filter, hiding some dives.
- User presses CTRL-Z.
- Dives are restored with their selection, but some of them are not displayed -> inconsistent state (currently the code expects that filtered-out dives are not selected).

You can generate numerous scenarios like that. For now I plan to keep `selected` true even if a dive is not shown and replace all `selected` conditions by `selected && !hidden_by_filter` (hidden in a function, of course).

Another option (actually the easier one) might be making filter-changes undo-able. Thus we would always ensure a consistent selection state. But I don't think this would be right from a UI-perspective.

bstoeger:
Forget last message - I changed my mind (again). Easiest way for now will be saving the full selection with every undo command. Certainly rather inefficient, but for now consistent-state is the top goal.

tcanabrava:
You are touching one of the hardest points in subsurface :(

bstoeger:
I can only do baby-steps at the moment. Selection is half-implemented, but the trip-selection thing needs more thought and renumbering / moving between trips currently loses selection.

During testing I discovered that undo of merging does not work properly because the merge routine modifies the data structures of the to-be-merged dives. The easy way would have been copy the dives and merge the copies. But I thought it can't be hard to make the in-parameters const and fix the fallout. It turned out to be super-annoying.

I fear that I need help with the commit "Undo: don't modify source-dives on merge". This is written completely blind! It seems to work for my simple test-case (one cylinder), but I have *no* idea what I'm doing here. If somebody could review that / find errors, that would be highly appreciated!

core/dive.c
```diff
@@ -270,8 +270,7 @@ enum divemode_t get_current_divemode(struct divecomputer *dc, int time, struct e
 	return *divemode;
 }
 
-/* this returns a pointer to static variable - so use it right away after calling */
```
dirkhh: this comment is still true... why did you delete it?

core/dive.c
```diff
@@ -714,18 +725,18 @@ void copy_cylinders(struct dive *s, struct dive *d, bool used_only)
...
+		d->cylinder[j].gasmix = s->cylinder[i].gasmix;
+		d->cylinder[j].depth = s->cylinder[i].depth;
+		d->cylinder[j].cylinder_use = s->cylinder[i].cylinder_use;
+		d->cylinder[j].manually_added = true;
+
+		// Restore the start and end pressures from original cylinder
+		d->cylinder[i].start.mbar = t[i].start.mbar;
+		d->cylinder[i].end.mbar = t[i].end.mbar;
+		d->cylinder[i].sample_start.mbar = t[i].sample_start.mbar;
+		d->cylinder[i].sample_end.mbar = t[i].sample_end.mbar;
```
dirkhh: that indentation change looks bogus to me

core/dive.c
```diff
@@ -1390,10 +1401,12 @@ static void fixup_airtemp(struct dive *dive)
 
 /* zero out the airtemp in the dive structure if it was just created by
  * running fixup on the dive. keep it if it had been edited by hand */
```
dirkhh: that comment is now misleading... 
```
/* if the air temperature in the dive data is redundant to the one in its
 * first divecomputer (i.e., it was added by running fixup on the dive)
 * return 0, otherwise return the air temperature given in the dive */
```

core/dive.c
```diff
@@ -1836,11 +1849,11 @@ struct sample *add_sample(struct sample *sample, int time, struct divecomputer *
  * that the time in between the dives is at the surface, not some "last
  * sample that happened to be at a depth of 1.2m".
  */
-static void merge_one_sample(struct sample *sample, int time, struct divecomputer *dc)
+static void merge_one_sample(const struct sample *sample, int time, struct divecomputer *dc)
 {
 	int last = dc->samples - 1;
 	if (last >= 0) {
-		static struct sample surface = { .bearing.degrees = -1, .ndl.seconds = -1 };
+		struct sample surface = { .bearing.degrees = -1, .ndl.seconds = -1 };
```
dirkhh: I realize that the pointless initialization was there before, but given that these values will be overwritten just four lines below, it does seem silly to keep it.
I'm trying to figure out why this used to be a static variable, but I'm guessing that made sense only in an earlier incarnation of the code?

dirkhh (commented):
Wait, I'm not done with the review.
GitHub is doing something weird. In order not to lose the comments made so far, I'll "finish" this review as it forces me to do, but I'll continue :-)

core/dive.c
dirkhh: actually, I can't seem to be able to comment on the code that isn't close to the modification (and apparently that's what ended my previous review), but I wonder if the fact that this sample is passed into add_sample() below was the (possibly misguided, need to reed that function again) reason to have the surface sample static
No, thankfully that's not the case, that would have been scary-insane. So yes, long story short, dropping 'static' seems like the right thing to do, and you could also drop the initialization.

core/dive.c
```diff
@@ -1925,18 +1950,23 @@ static void merge_samples(struct divecomputer *res, struct divecomputer *a, stru
...
+			int sensor_id;
+
+			sensor_id = cylinders_map_a[as->sensor[j]];
+			if (sensor_id < 0)
+				continue;
+
+			if (as->pressure[j].mbar)
+				sample.pressure[j] = as->pressure[j];
+			if (as->sensor[j])
+				sample.sensor[j] = sensor_id;
```
dirkhh, line 1973: this looks odd. shouldn't this be sensor_id instead of 'j'? That's the whole point of the map, right?

core/dive.c
```diff
@@ -1999,43 +2029,54 @@ static int same_gas(struct event *a, struct event *b)
...
+		cylinders_map2 = cylinders_map_tmp;
 	}
 
 	a = src1->events;
 	b = src2->events;
-	while (b) {
-		b->time.seconds += offset;
-		b = b->next;
-	}
-	b = src2->events;
```
dirkhh, line 2027: so you remove the event adjustment here because you now do consider the offset as part of the sort... but I'm not quite understanding why you are making that change

dirkhh (commented):
I have a few questions and a couple minor suggestions, but even staring at this for 20+ minutes I'm neither sure I found a bug, nor am I convinced it's correct. I'll try to get @torvalds to take a look as well as this is code he was very involved in writing.

core/dive.c
torvalds: Deleting it makes sense - if it's const and nobody will change it, it doesn't matter that it's static.

core/dive.c
torvalds: Yeah, that's completely bogus

core/dive.c
```diff
@@ -1977,20 +2007,20 @@ static char *merge_text(const char *a, const char *b, const char *sep)
...
+		return a->field < b->field + offset ? -1 : 1
 
-static int sort_event(struct event *a, struct event *b)
+static int sort_event(const struct event *a, const struct event *b, int offset)
 {
-	SORT(a, b, time.seconds);
-	SORT(a, b, type);
-	SORT(a, b, flags);
-	SORT(a, b, value);
+	SORT(a, b, time.seconds, offset);
```
torvalds: This change makes no sense.

Honestly, I see why you do it, but the end result is all kinds of odd. The offset simply doesn't make sense in this context.

Instead of passing in "offset" to sort_event(), just pass in both times. Then this code continues to make sense, and the games with the time will make (more) sense in the context of the caller.

You can see how Dirk is confused about the caller code.

torvalds (commented):
Ugh. I _really_ wish this was split up much more. It's hard to look at. I had to fetch the code locally just to look at it without some of the github noise. 

I really think the end result is likely fine, but it looks like a lot of the "const" additions could have been done independently in an earlier commit, so that there wouldn't be this one big commit that is hard to read.

All the gasmix functions could probably have been made const without much affecting anything else, for example. 

Then, all the event functions..

And only in the very last phase the "struct divecomputer" itself, which is to a large degree what then causes a lot of the actual reorg parts of the patch.

core/dive.c
bstoeger: Also note that the comment was never correct in the general case - only if event is not a gas-change event would it return pointer-to-static. The other case (pointer-to-heap) seems much more dangerous and is probably the one that deserves a comment.

Even though it's not classical C, I wonder if we could simply return by value here. `gasmix` is two integers. I think on common 64-bit platforms this will be returned in register anyway?

For
```
struct X {
        int a;
        int b;
};

struct X f()
{
        struct X x;
        x.a = 1;
        x.b = 2;
        return x;
}
```
I get
```
f:
        movabsq $8589934593, %rax
        ret
```

bstoeger:
> Ugh. I really wish this was split up much more. It's hard to look at. I had to fetch the code locally just to look at it without some of the github noise.

Yes, sorry. :(
I thought it will only be a few lines and was in for a bad surprise. Moreover, I was always terrible at organizing.

I will try to split it up when I find time.

core/dive.c
bstoeger: Hmm... I believe this is correct: this transforms `as->sensor[j]` into `sample.sensor[j]`. `sample_renumber()` [which was split off of `dc_cylinder_renumber()`] seems to do the same thing.

Note that `j` loops over MAX_SENSORS (which effectively means that `j` loops over `{0,1}`), not over the cylinders. The way I understand things, a AI divecomputer can measure at most two pressures at the same time, but the cylinder can change...?

But I'm super-tired at the moment, so I might be seeing things wrongly.

core/dive.c
bstoeger: Copied that verbatim, thanks.

core/dive.c
bstoeger: The whole purpose of this commit is to not modify the original dive data. So we can't adjust the offset of the events of the `src2` dive. :(

bstoeger:
Help, I can't rebase:

```
First, rewinding head to replay your work on top of it...
Applying: Undo: fix multi-level undo of delete-dive and remove-dive-from-trip
Applying: Cleanup: let MainTab::updateDiveSite() take dive * instead of id
Applying: Undo: implement undo of manual dive-creation
Applying: Undo: implement rudimentary support for undo of dive-splitting
Applying: Undo: implement rudimentary support for undo of dive-merging
Applying: Undo: clear undo history on load / import
Applying: Dive editing: don't repeatedly loop through whole dive list
Applying: Undo: make editing of dive-time an undoable operation
Applying: Dive list: replace dive-id by dive pointer
Using index info to reconstruct a base tree...
M       libdivecomputer
Falling back to patching base and 3-way merge...
warning: Failed to merge submodule libdivecomputer (commits don't follow merge-base)
Auto-merging libdivecomputer
CONFLICT (submodule): Merge conflict in libdivecomputer
error: Failed to merge in the changes.
Patch failed at 0009 Dive list: replace dive-id by dive pointer
Use 'git am --show-current-patch' to see the failed patch

Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
```

For some reason, the patch in question contains a libdivecomputer update:
```
Date:   Sun Jul 22 13:40:02 2018 +0200

    Dive list: replace dive-id by dive pointer
    
    The undo-system now guarantees that pointers to dives are stable
    throughout their lifetime. Therefore, replace the unique index by
    pointers. This is a small performance improvement, but much more
    importantly, it will make it more natural to transport a pointer
    to the dive inside QModelIndex's private pointer.
    
diff --git a/libdivecomputer b/libdivecomputer
index f6ea5f514..bb985eedb 160000
--- a/libdivecomputer
+++ b/libdivecomputer
@@ -1 +1 @@
-Subproject commit f6ea5f514abae747dc29a68490c58cd972f3e392
+Subproject commit bb985eedbb3e686e5a71bf2aa234fb49f8cab653
```
I fear that `git rebase --skip` will remove this commit... :(
Should I `git add -i` everything by the subproject commit? Is that even possible?

bstoeger:
OK - got it with some detours.

bstoeger:
Small update: Proof-of-concept connection of dive-list-commands and filters. Currently it's adding dive-by-dive which will be very ineffective for large undo-commands (notably dive-import). Once these commands are implemented, this will have to be reworked a little bit, but it looks rather trivial. This is connected via functions, not signals, because we need determinism: on adding dives, the filters should be populated *before* adding the dives, so that the latter get the correct filter-status on addition.

I'm starting to see light at the end of the tunnel. I know of a few issues which will have to be addressed, then we can start testing / deciding if this is the way to go. But be warned: at the beginning this *will* spectacularly go down in flames.

My revised short-to-long-term priorities:.
1) Split `DiveTripModel` in two models: list-view and tree-view. list-view does not need any mirrored data from the C-core. Switch both models as source of the filter model if the view is changed. We'll have to see how this works out with signals.
2) Make import/download undoable. Adapt the filter code in this regard.
3) Start making editing functions undoable.
4) Port to mobile(?)
5) Make the C-core compatible with the tree-model [i.e. replace lists by sorted arrays]. Remove mirrored data in the C++ layer.

dirkhh:
Nice to see the progress.

bstoeger:
Another small update: Make adding/cloning of planned dives undo-able.

bstoeger:
For me this is now in a state where this can be tested / needs a decision.

Let me recapitulate my disclaimer:
1) This is very complex code.
2) At the beginning this will bring some pain.

What does it do:
1) Make nearly all dive-list modifying operations undo-able (exception: dive-import).
2) Implement Qt model/view interface (that means the dive-list is *not* recalculated every time, which makes the interface significantly more snappy for me).
3) Incrementally update the filter boxes.
4) Start to remove layering violations and attempt to define a proper data-flow all the way down from C via the model up to the views. Notably the selection!

So why is it complex:
1) Qt's model/view interface.
2) Two models in one (List view and tree view).
3) Duplication of data in C and C++ layers.

How can we solve that:
1) Is a fact we'll have to live with. Or give up and do a full reset of the model on every change.
2) Is inherited from old code. I plan to split this in two models (list and tree) and just set the appropriate model if the view-mode is changed. Easy enough.
3) Once the two models are split we can pretty much immediately remove the list-model from C++, as this data is already in the C layer. The tree-model we could likewise push down to C, but that will be more work. It will mean replacing all the linked-lists with dynamic arrays. I.e. implementing `std::vector` using C-macros..? Not looking forward to that. Also I'm not yet sure about the top-level of the (two-level) tree, which is a mix of dives / dive-trips. I will think about that in due course.

Other things for the future:
- Make dive-import undo-able.
- Make dive-editing (also the planner!) undo-able.
- Include in mobile.

dirkhh:
30+ commits, 2600 lines net added. This is a monster PR. Which means that in order to review this I need to find at least a couple of hours of undisturbed time... that's going to be tough... my life rarely presents me with those gems in a way that I know up front that I'll have them.
I glanced through this and it looks like it will move us a huge step forward. So I think this is by and large the right thing to do. But please be patient while I try to find time to review this. Highly unlikely to happen today, given all the family activities going on...

bstoeger:
No problem - hasn't to happen immediately. I will nag you again in due course. ;)

core/divelist.c
```diff
@@ -861,12 +896,27 @@ void add_dive_to_trip(struct dive *dive, dive_trip_t *trip)
 		trip->when = dive->when;
 }
 
+dive_trip_t *alloc_trip()
```
torvalds: Please use proper prototypes.

So "dive_trip_t *alloc_trip(void)"

torvalds:
Only looking at the core C files rather than the UI C++ ones - it looks fine. 

I wonder how much test coverage we have for the complex cases (ie merges etc), but I don't see anything fundamentally wrong.

dirkhh:
OK, first third reviewed, looking good so far. The rest will have to wait for later.

bstoeger:
@dirkhh: I understand Windows/BLE has priority, but if you find time could you perhaps review the second third? Rebasing this on master was very painful (even though 100% self-inflicted, who gave me the crazy idea of messing with dive import while this PR is in review!).

dirkhh:
I guess the WIP in the title keeps making it drop from my high priority review list.
I'm on a plane again, and today's schedule is murder, so realistically I doubt I'll get to it for the next 12+ hours...

bstoeger:
A couple of days more or less won't make a significant difference. Removed WIP, as this is indeed for testing / review.

janmulder:
Yes, its difficult to get these monster PRs going. The first thing I tried did not work as expected, so, I read the message thread above to see if the thing that I was testing was supposed to be included (and it was not: undo of import). This is a problem when getting this tested. Further, reporting issues in here is also a little confusing, as its difficult to keep track of individual issues.

This said: my first issue found in the PR (and probably a simple one). Delete all dives from the divelist. Before, the state of the app was as starting choosing close form the menu. Now, it keeps showing the last selected dive, and the divelist is (correctly) empty. At least, this is inconsistent.

janmulder:
Bug 2: opened my earlier provided test-import.ssrf. File contains 21 dives (without any data filled in, only DC data). Give the first dive a dive site (just type something in the location field). And hit apply in the blue bar above.
1) navigate to a different dive in the divelist. Notice that the entered dive site name keeps populated. Which is wrong as the newly selected dive does not have a site filled in.
2) likely related to 1). Choose file->close. A message is shown: please save or cancel before closing the file. This looks like we are in EDIT mode, but we are not as we applied the edit in 1). In addition, the blue bar is not shown, so show should we save/cancel the edit? Using shortcut <esc> I can discard my edit.

core/subsurface-qt/DiveListNotifier.h
```diff
@@ -30,10 +30,69 @@ class DiveListNotifier : public QObject {
...
+		DiveListNotifier &notifier;
+		bool oldValue;
+		InCommandMarker(DiveListNotifier &);
+		friend DiveListNotifier;
+	public:
+		~InCommandMarker();
+	};
+
+	// Usage:
+	// voif doWork()
```
dirkhh: I like `voif` functions...

desktop-widgets/undocommands.cpp
dirkhh: this still needs to be addressed!

dirkhh (approved):
I had a meeting cancel. And instead of... never mind, so I did another review.
Yeah, this is hard to review. You do an outstanding job commenting and explaining constructs. And now that I read this code the third time I am starting to feel more comfortable with what I'm seeing. In a few commits I still got lost, mainly because I didn't have the time to go back to the full sources and making sure I understand the side effects.
One thing that needs to be fixed (not necessarily before merging, but in a follow up SOON so we don't ship with this) is the handling of plurals in translation strings. I noticed this in a few spots, I commented on the most egregious one, but there are more cases where you just say `"...dives..."` instead of dealing with the plural `"... dive(s)..."`

I asked Linus to please take a look and I hope he does. But at this point I think I'm ok with seeing this merged and pushing hard on getting it tested by our developer base...

janmulder:
Yep, despite the fact that I already identified some possible problems, I also think that merging soon is the best way to get this huge and very useful PR moving. I feel that the amount of developers testing stuff before merge is limited. So, we get more testing and proper bug reporting than letting this sit again.

bstoeger:
@janmulder: Thanks! Your testing is very much appreciated - finally progress. I will fix those when I find time (currently busy with lecturing).

@dirkhh: Thanks for the review! Are you reviewing an intermediate version by chance? I see
```
setText(tr("add %n dives to trip", "", divesToAddIn.size()));
```
I didn't check the history though.

bstoeger:
Concerning the CHANGELOG.md conflict: when I rebase locally it works but github complains. Weird.

dirkhh:
Don't worry about the merge. I'll deal with that.
As for the plurals, I reviewed what GitHub showed me as the most recent... went I went looking for it just now I couldn't find it, so maybe that is resolved. But it's something we should look for in the end result.

torvalds:
I didn't actually look at the individual commits, just the end result. But I like what I saw. 

Admittedly, I'm not actually all that used to all the C++'isms, but this all seems to be the *right* C++'isms. And the end result seems way more powerful than the ad-hoc situation we had before.

I do note how I once again wished that "alloc_trip()" would always take the date to be used for initialization, but that's an entirely independent change, and there may well be reasons why we don't always have a date initially. It was just made more obvious by this series since now it's an explicit helper function.

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -670,21 +671,12 @@ void MainTab::reload()
 }
 
 // tricky little macro to edit all the selected dives
-// loop over all dives, for each selected dive do WHAT, but do it
-// last for the current dive; this is required in case the invocation
-// wants to compare things to the original value in current_dive like it should
-#define MODIFY_SELECTED_DIVES(WHAT)                            \
+// loop ove all DIVES and do WHAT.
+#define MODIFY_DIVES(DIVES, WHAT)                            \
```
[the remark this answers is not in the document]
bstoeger: Sure, but the PR is monstrous as-is. I will think about that once (if ever) we make dive-editing undo-able.

bstoeger:
@dirkhh: Somehow I managed to mess up dive-adding. I consider this a show-stopper. Unfortunately, I will not find time today to fix it. Please give me a day or two.

desktop-widgets/tab-widgets/maintab.cpp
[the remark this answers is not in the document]
torvalds: Yeah, I had no complaints about that. It was more a "other parts of your changes went the template route". 

Everything I looked at looks fine to me.

dirkhh:
@bstoeger no problem. Honest question: is it easier to just merge it and live with this being broken in master for a few days? I want to prevent you from having to keep re-basing such a large PR.

bstoeger:
Fortunately, the problem with dive-adding was rather easy to fix: I had inadvertently deleted the "addDive()" call. In light testing, this seems also to work with tags now.

janmulder:
The bug I called nr 2) yesterday can be solved by:

```
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -1002,6 +1002,7 @@ void MainTab::acceptChanges()
        DivePlannerPointsModel::instance()->setPlanMode(DivePlannerPointsModel::NOTHING);
        MainWindow::instance()->dive_list()->verticalScrollBar()->setSliderPosition(scrolledBy);
        MainWindow::instance()->dive_list()->setFocus();
+       editMode = NONE;
        MainWindow::instance()->exitEditState();
        cylindersModel->changed = false;
        weightModel->changed = false;

```

But this said. Why does `MainWindow::exitEditState()` not reset the `editMode`?

janmulder:
Yep, bug 1 is gone.

bstoeger:
> But this said. Why does MainWindow::exitEditState() not reset the editMode?

Note that `MainWindow::exitEditState()` changes the state of Main*Window*, whereas `editMode` is the state of Main*Tab*.

The bug was introduced in commit "Undo: implement undo of manual dive-creation" where, for reasons I can't determine anymore, I removed one `editMode = NONE;` that should have stayed.

Without your debugging, I think I would have had a very hard time figuring this out.

bstoeger:
To be honest, I'm a bit scared of the fallout... 😛

PR #1531, opened 2018-07-22
 move prefs.dive_computer from SettingsObjectWrapper to qPrefDiveComputer

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add qPrefDivecomputer with test cases.

This is a sub struct in struct preferences so the macros are modified to handle that effect.

getters as functions are overkill, but needed for QML, made these static inline, which saves code, and lines of code to write.

Removed LOADSYNC* macroes in order to reduce the number of macros.

### DIFF
--- a/core/settings/qPrefAnimations.h
+++ b/core/settings/qPrefAnimations.h
@@ -15,8 +15,8 @@ public:
 
 	// Load/Sync local settings (disk) and struct preference
 	void loadSync(bool doSync);
-	void load() { loadSync(false); }
-	void sync() { loadSync(true); }
+	void inline load() { loadSync(false); }
+	void inline sync() { loadSync(true); }
 
 public:
 	static inline int animation_speed() { return prefs.animation_speed; };
--- a/core/settings/qPrefAnimations.h
+++ b/core/settings/qPrefAnimations.h
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
 #ifndef QPREFANIMATIONS_H
 #define QPREFANIMATIONS_H
+#include "core/pref.h"
 
 #include <QObject>
 
@@ -18,7 +19,7 @@ public:
 	void sync() { loadSync(true); }
 
 public:
-	int animation_speed() const;
+	static inline int animation_speed() { return prefs.animation_speed; };
 
 public slots:
 	void set_animation_speed(int value);
--- a/core/settings/qPrefCloudStorage.cpp
+++ b/core/settings/qPrefCloudStorage.cpp
@@ -31,6 +31,7 @@ void qPrefCloudStorage::loadSync(bool doSync)
 
 void qPrefCloudStorage::set_cloud_base_url(const QString& value)
 {
+	QString valueGit = value + "/git";
 	if (value != prefs.cloud_base_url) {
 		// only free and set if not default
 		if (prefs.cloud_base_url != default_prefs.cloud_base_url) {
@@ -39,14 +40,12 @@ void qPrefCloudStorage::set_cloud_base_url(const QString& value)
 		}
 
 		disk_cloud_base_url(true);
+		disk_cloud_git_url(true);
 		emit cloud_base_url_changed(value);
+		emit cloud_git_url_changed(valueGit);
 	}
 }
-void qPrefCloudStorage::disk_cloud_base_url(bool doSync)
-{
-	LOADSYNC_TXT("/cloud_base_url", cloud_base_url);
-	LOADSYNC_TXT("/cloud_git_url", cloud_git_url);
-}
+DISK_LOADSYNC_TXT(CloudStorage, "/cloud_base_url", cloud_base_url)
 
 void qPrefCloudStorage::set_cloud_git_url(const QString& value)
 {
@@ -86,8 +85,12 @@ void qPrefCloudStorage::set_cloud_storage_password(const QString& value)
 }
 void qPrefCloudStorage::disk_cloud_storage_password(bool doSync)
 {
-	if (!doSync || prefs.save_password_local)
-		LOADSYNC_TXT("/password", cloud_storage_password);
+	if (doSync) {
+		if (prefs.save_password_local) 
+			qPrefPrivate::instance()->setting.setValue(group + "/password", prefs.cloud_storage_password);
+	} else {
+		prefs.cloud_storage_password = copy_qstring(qPrefPrivate::instance()->setting.value(group + "/password", default_prefs.cloud_storage_password).toString());
+	}
 }
 
 HANDLE_PREFERENCE_TXT(CloudStorage, "/pin", cloud_storage_pin);
@@ -107,5 +110,8 @@ void qPrefCloudStorage::disk_userid(bool doSync)
 {
 	//WARNING: UserId is stored outside of any group, but it belongs to Cloud Storage.
 	const QString group = QStringLiteral("");
-	LOADSYNC_TXT("subsurface_webservice_uid", userid);
+	if (doSync)
+		qPrefPrivate::instance()->setting.setValue(group + "subsurface_webservice_uid", prefs.userid);
+	else
+		prefs.userid = copy_qstring(qPrefPrivate::instance()->setting.value(group + "subsurface_webservice_uid", default_prefs.userid).toString());
 }
--- a/core/settings/qPrefCloudStorage.cpp
+++ b/core/settings/qPrefCloudStorage.cpp
@@ -29,7 +29,6 @@ void qPrefCloudStorage::loadSync(bool doSync)
 	disk_userid(doSync);
 }
 
-GET_PREFERENCE_TXT(CloudStorage, cloud_base_url);
 void qPrefCloudStorage::set_cloud_base_url(const QString& value)
 {
 	if (value != prefs.cloud_base_url) {
@@ -49,7 +48,6 @@ void qPrefCloudStorage::disk_cloud_base_url(bool doSync)
 	LOADSYNC_TXT("/cloud_git_url", cloud_git_url);
 }
 
-GET_PREFERENCE_TXT(CloudStorage, cloud_git_url);
 void qPrefCloudStorage::set_cloud_git_url(const QString& value)
 {
 	if (value != prefs.cloud_git_url) {
@@ -67,7 +65,6 @@ HANDLE_PREFERENCE_TXT(CloudStorage, "/email", cloud_storage_email);
 
 HANDLE_PREFERENCE_TXT(CloudStorage, "/email_encoded", cloud_storage_email_encoded);
 
-GET_PREFERENCE_TXT(CloudStorage, cloud_storage_newpassword);
 void qPrefCloudStorage::set_cloud_storage_newpassword(const QString& value)
 {
 	if (value == prefs.cloud_storage_newpassword)
@@ -79,7 +76,6 @@ void qPrefCloudStorage::set_cloud_storage_newpassword(const QString& value)
 	emit cloud_storage_newpassword_changed(value);
 }
 
-GET_PREFERENCE_TXT(CloudStorage, cloud_storage_password);
 void qPrefCloudStorage::set_cloud_storage_password(const QString& value)
 {
 	if (value != prefs.cloud_storage_password) {
@@ -106,7 +102,6 @@ HANDLE_PREFERENCE_BOOL(CloudStorage, "/save_password_local", save_password_local
 
 HANDLE_PREFERENCE_BOOL(CloudStorage, "/save_userid_local", save_userid_local);
 
-GET_PREFERENCE_TXT(CloudStorage, userid);
 SET_PREFERENCE_TXT(CloudStorage, userid);
 void qPrefCloudStorage::disk_userid(bool doSync)
 {
--- a/core/settings/qPrefCloudStorage.cpp
+++ b/core/settings/qPrefCloudStorage.cpp
@@ -35,8 +35,8 @@ void qPrefCloudStorage::set_cloud_base_url(const QString& value)
 	if (value != prefs.cloud_base_url) {
 		// only free and set if not default
 		if (prefs.cloud_base_url != default_prefs.cloud_base_url) {
-			COPY_TXT(cloud_base_url, value);
-			COPY_TXT(cloud_git_url, QString(prefs.cloud_base_url) + "/git");
+			qPrefPrivate::copy_txt(&prefs.cloud_base_url, value);
+			qPrefPrivate::copy_txt(&prefs.cloud_git_url, QString(prefs.cloud_base_url) + "/git");
 		}
 
 		disk_cloud_base_url(true);
@@ -55,7 +55,7 @@ void qPrefCloudStorage::set_cloud_git_url(const QString& value)
 	if (value != prefs.cloud_git_url) {
 		// only free and set if not default
 		if (prefs.cloud_git_url != default_prefs.cloud_git_url) {
-			COPY_TXT(cloud_git_url, value);
+			qPrefPrivate::copy_txt(&prefs.cloud_git_url, value);
 		}
 		disk_cloud_git_url(true);
 		emit cloud_git_url_changed(value);
@@ -73,7 +73,7 @@ void qPrefCloudStorage::set_cloud_storage_newpassword(const QString& value)
 	if (value == prefs.cloud_storage_newpassword)
 		return;
 
-	COPY_TXT(cloud_storage_newpassword, value);
+	qPrefPrivate::copy_txt(&prefs.cloud_storage_newpassword, value);
 
 	// NOT saved on disk, because it is only temporary
 	emit cloud_storage_newpassword_changed(value);
@@ -83,7 +83,7 @@ GET_PREFERENCE_TXT(CloudStorage, cloud_storage_password);
 void qPrefCloudStorage::set_cloud_storage_password(const QString& value)
 {
 	if (value != prefs.cloud_storage_password) {
-		COPY_TXT(cloud_storage_password,value);
+		qPrefPrivate::copy_txt(&prefs.cloud_storage_password, value);
 		disk_cloud_storage_password(true);
 		emit cloud_storage_password_changed(value);
 	}
--- a/core/settings/qPrefCloudStorage.cpp
+++ b/core/settings/qPrefCloudStorage.cpp
@@ -108,10 +108,16 @@ HANDLE_PREFERENCE_BOOL(CloudStorage, "/save_userid_local", save_userid_local);
 SET_PREFERENCE_TXT(CloudStorage, userid);
 void qPrefCloudStorage::disk_userid(bool doSync)
 {
-	//WARNING: UserId is stored outside of any group, but it belongs to Cloud Storage.
-	const QString group = QStringLiteral("");
-	if (doSync)
+	if (doSync) {
+		// always save in new position (part of cloud storage group)
 		qPrefPrivate::instance()->setting.setValue(group + "subsurface_webservice_uid", prefs.userid);
-	else
-		prefs.userid = copy_qstring(qPrefPrivate::instance()->setting.value(group + "subsurface_webservice_uid", default_prefs.userid).toString());
+	} else {
+		//WARNING: UserId was  stored outside of any group.
+		// try to read from new location, if it fails read from old location
+		prefs.userid = copy_qstring(qPrefPrivate::instance()->setting.value(group + "subsurface_webservice_uid", "NoUserIdHere").toString());
+		if (QString(prefs.userid) == "NoUserIdHere") {
+			const QString group = QStringLiteral("");
+			prefs.userid = copy_qstring(qPrefPrivate::instance()->setting.value(group + "subsurface_webservice_uid", default_prefs.userid).toString());
+		}
+	}
 }
--- a/core/settings/qPrefCloudStorage.h
+++ b/core/settings/qPrefCloudStorage.h
@@ -27,23 +27,23 @@ public:
 
 	// Load/Sync local settings (disk) and struct preference
 	void loadSync(bool doSync);
-	void load() { loadSync(false); }
-	void sync() { loadSync(true); }
+	void inline load() { loadSync(false); }
+	void inline sync() { loadSync(true); }
 
 public:
-	static inline const QString cloud_base_url() { return prefs.cloud_base_url; }
-	static inline const QString cloud_git_url() { return prefs.cloud_git_url; }
-	static inline const QString cloud_storage_email() { return prefs.cloud_storage_email; }
-	static inline const QString cloud_storage_email_encoded() { return prefs.cloud_storage_email_encoded; }
-	static inline const QString cloud_storage_newpassword() { return prefs.cloud_storage_newpassword; }
-	static inline const QString cloud_storage_password() { return prefs.cloud_storage_password; }
-	static inline const QString cloud_storage_pin() { return prefs.cloud_storage_pin; }
+	static inline QString cloud_base_url() { return prefs.cloud_base_url; }
+	static inline QString cloud_git_url() { return prefs.cloud_git_url; }
+	static inline QString cloud_storage_email() { return prefs.cloud_storage_email; }
+	static inline QString cloud_storage_email_encoded() { return prefs.cloud_storage_email_encoded; }
+	static inline QString cloud_storage_newpassword() { return prefs.cloud_storage_newpassword; }
+	static inline QString cloud_storage_password() { return prefs.cloud_storage_password; }
+	static inline QString cloud_storage_pin() { return prefs.cloud_storage_pin; }
 	static inline int cloud_timeout() { return prefs.cloud_timeout; }
 	static inline int cloud_verification_status() { return prefs.cloud_verification_status; }
 	static inline bool git_local_only() { return prefs.git_local_only; }
 	static inline bool save_password_local() { return prefs.save_password_local; }
 	static inline bool save_userid_local() { return prefs.save_userid_local; }
-	static inline const QString userid() { return prefs.userid; }
+	static inline QString userid() { return prefs.userid; }
 
 public slots:
 	void set_cloud_base_url(const QString& value);
--- a/core/settings/qPrefCloudStorage.h
+++ b/core/settings/qPrefCloudStorage.h
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
 #ifndef QPREFCLOUDSTORAGE_H
 #define QPREFCLOUDSTORAGE_H
+#include "core/pref.h"
 
 #include <QObject>
 
@@ -30,19 +31,19 @@ public:
 	void sync() { loadSync(true); }
 
 public:
-	const QString cloud_base_url() const;
-	const QString cloud_git_url() const;
-	const QString cloud_storage_email() const;
-	const QString cloud_storage_email_encoded() const;
-	const QString cloud_storage_newpassword() const;
-	const QString cloud_storage_password() const;
-	const QString cloud_storage_pin() const;
-	int   cloud_timeout() const;
-	int   cloud_verification_status() const;
-	bool  git_local_only() const;
-	bool  save_password_local() const;
-	bool  save_userid_local() const;
-	const QString userid() const;
+	static inline const QString cloud_base_url() { return prefs.cloud_base_url; }
+	static inline const QString cloud_git_url() { return prefs.cloud_git_url; }
+	static inline const QString cloud_storage_email() { return prefs.cloud_storage_email; }
+	static inline const QString cloud_storage_email_encoded() { return prefs.cloud_storage_email_encoded; }
+	static inline const QString cloud_storage_newpassword() { return prefs.cloud_storage_newpassword; }
+	static inline const QString cloud_storage_password() { return prefs.cloud_storage_password; }
+	static inline const QString cloud_storage_pin() { return prefs.cloud_storage_pin; }
+	static inline int cloud_timeout() { return prefs.cloud_timeout; }
+	static inline int cloud_verification_status() { return prefs.cloud_verification_status; }
+	static inline bool git_local_only() { return prefs.git_local_only; }
+	static inline bool save_password_local() { return prefs.save_password_local; }
+	static inline bool save_userid_local() { return prefs.save_userid_local; }
+	static inline const QString userid() { return prefs.userid; }
 
 public slots:
 	void set_cloud_base_url(const QString& value);
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -35,7 +35,7 @@ void qPrefDisplay::set_divelist_font(const QString& value)
 
 	if (newValue != prefs.divelist_font &&
 		!subsurface_ignore_font(qPrintable(newValue))) {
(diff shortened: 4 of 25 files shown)


dirkhh:
The Travis errors both look like real issues, not Travis flukes. And they both seem somewhat related to the prefs code, so possibly indeed cause by this PR...

janiversen:
yes it is caused by a change in the macros.....hurraah for using test cases

dirkhh (changes requested):
Tiny coding style issues, but other than that it looks good.

core/settings/qPrefCloudStorage.h
```diff
@@ -30,19 +31,19 @@ class qPrefCloudStorage : public QObject {
...
-	const QString cloud_storage_newpassword() const;
-	const QString cloud_storage_password() const;
-	const QString cloud_storage_pin() const;
-	int   cloud_timeout() const;
-	int   cloud_verification_status() const;
-	bool  git_local_only() const;
-	bool  save_password_local() const;
-	bool  save_userid_local() const;
-	const QString userid() const;
+	static inline const QString cloud_base_url() { return prefs.cloud_base_url; };
```
bstoeger: Lots of unnecessary semicolons...

core/settings/qPrefDisplay.cpp
```diff
@@ -66,7 +64,7 @@ void qPrefDisplay::set_font_size(double value)
 void qPrefDisplay::disk_font_size(bool doSync)
 {
 	if (doSync)
-		LOADSYNC_DOUBLE("/font_size", font_size)
+		qPrefPrivate::instance()->setting.setValue(group + "/font_size", prefs.font_size); \
```
bstoeger: backslash?

core/settings/qPrefDiveComputer.h
```diff
@@ -0,0 +1,55 @@
...
+	Q_PROPERTY(QString vendor READ vendor WRITE	set_vendor NOTIFY vendor_changed);
+
+public:
+	qPrefDiveComputer(QObject *parent = NULL);
+	static qPrefDiveComputer *instance();
+
+	// Load/Sync local settings (disk) and struct preference
+	void loadSync(bool doSync);
+	void load() {loadSync(false); }
+	void sync() {loadSync(true); }
```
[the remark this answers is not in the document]
bstoeger: @dirkhh: these *are* inlined functions.
So actually there are two questions to ask:
1) Do we want these functions inlined?
2) Do we want the inlined functions put in class definitions?

core/settings/qPrefPrivate.cpp
```diff
@@ -9,3 +9,10 @@ qPrefPrivate *qPrefPrivate::instance()
 	static qPrefPrivate *self = new qPrefPrivate;
 	return self;
 }
+
+void qPrefPrivate::copy_txt(const char **name, const QString& string)
+{ \
```
bstoeger: backslashes?

core/settings/qPrefDiveComputer.h
```diff
@@ -0,0 +1,55 @@
...
+
+public slots:
+	void set_device(const QString& device);
+	void set_device_name(const QString& device_name);
+	void set_download_mode(int mode);
+	void set_product(const QString& product);
+	void set_vendor(const QString& vendor);
+
+signals:
+	void device_changed(const QString& device);
```
bstoeger, line 40: We put '&' and '*' to the variable name:
```
	void device_changed(const QString &device);
```
Some people don't like it - but that's how the C-inherited grammar works - sorry.

core/settings/qPrefPrivate.cpp
```diff
@@ -9,3 +9,10 @@ qPrefPrivate *qPrefPrivate::instance()
 	static qPrefPrivate *self = new qPrefPrivate;
 	return self;
 }
+
+void qPrefPrivate::copy_txt(const char **name, const QString& string)
```
bstoeger, line 13: 1) Am I missing the member variable that this is using or why is this non-static?
2) Idiomatic C++ would be
```
void qPrefPrivate::copy_txt(const char *&name, const QString& string)
```

core/settings/qPrefPrivate.h
```diff
@@ -14,213 +14,181 @@ class qPrefPrivate : public QObject {
     Q_OBJECT
```
bstoeger, line 14: What's the point of making this class a `QObject`? I don't see any signals or the like?

core/settings/qPrefDisplay.cpp
```diff
@@ -35,7 +34,7 @@ void qPrefDisplay::set_divelist_font(const QString& value)
 
 	if (newValue != prefs.divelist_font &&
 		!subsurface_ignore_font(qPrintable(newValue))) {
-		COPY_TXT(divelist_font, value);
+		qPrefPrivate::instance()->copy_txt(&prefs.divelist_font, value);
```
bstoeger: 1) Why would we need an object-instance to copy a text? This seems fundamentally wrong, unless we count the number of bytes copied or something?
2) Wouldn't it be possible to derive from a private base-class that provides these helper functions?

core/settings/qPrefDiveComputer.h
```diff
@@ -0,0 +1,55 @@
...
+	qPrefDiveComputer(QObject *parent = NULL);
+	static qPrefDiveComputer *instance();
+
+	// Load/Sync local settings (disk) and struct preference
+	void loadSync(bool doSync);
+	void load() {loadSync(false); }
+	void sync() {loadSync(true); }
+
+public:
+	static inline const QString device() {return QString(prefs.dive_computer.device); };
```
bstoeger: `inline` is redundant here.
When I read a class definition I'm not particularly interested in implementation details, so I think @dirkhh has a point.

core/settings/qPrefDiveComputer.h
```diff
@@ -0,0 +1,55 @@
...
+	static inline const QString device_name() {return QString(prefs.dive_computer.device_name); };
+	static inline int download_mode() {return prefs.dive_computer.download_mode; };
+	static inline const QString product() {return QString(prefs.dive_computer.product); };
```
bstoeger: 1) Note that AFAIK this could be simply written as `return prefs.dive_computer.product;`. I don't see what useful information one gets from the explicit conversion constructor.

2) You return a `QString` by *value* (i.e a copy). There is no point in making it `const`. Note that although it looks similar, this is very different from `const char *product()`. The pointer you return in that case is *not* const, but the data you point to *is* const. What you have written would be the equivalent to `const char * const product()`. Note the two `const`. If you would return by reference, then const could make sense: `const QString &product()`, because there the caller does not get a unique copy, but a possibly shared reference. [Returning a reference in this case would of course be a bug - the referenced object is immediately destroyed on function return. I'm just using this as an example].

To make a long story short. If you write
```
	static inline const QString product() {...} 
```
you also have to write
```
	static inline const int download_mode() {...}
```
and, frankly, nobody would do that for good reasons.

core/settings/qPrefDisplay.cpp
janiversen: 1. it is fundamentally wirng, oversight from my side, corrected.
2. yes but then we would create 2 Q_OBJECTS, and have a qsettings variable in the interface classes. We would also introduce multiple inheritance something which I prefer to aviod.

core/settings/qPrefPrivate.cpp
janiversen: see other comment, changed to static was oversight.

core/settings/qPrefPrivate.h
janiversen: I had problems with QSettings if the Q_OBJECT was not present.

core/settings/qPrefCloudStorage.cpp
```diff
@@ -86,8 +85,12 @@ void qPrefCloudStorage::set_cloud_storage_password(const QString& value)
...
 void qPrefCloudStorage::disk_cloud_storage_password(bool doSync)
 {
-	if (!doSync || prefs.save_password_local)
-		LOADSYNC_TXT("/password", cloud_storage_password);
+	if (doSync) {
+		if (prefs.save_password_local) 
+			qPrefPrivate::instance()->setting.setValue(group + "/password", prefs.cloud_storage_password);
+	}
+	else
+        prefs.cloud_storage_password = copy_qstring(qPrefPrivate::instance()->setting.value(group + "/password", default_prefs.cloud_storage_password).toString());
```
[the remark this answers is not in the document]
janiversen: changed

core/settings/qPrefCloudStorage.cpp
```diff
@@ -107,5 +110,8 @@ void qPrefCloudStorage::disk_userid(bool doSync)
 {
 	//WARNING: UserId is stored outside of any group, but it belongs to Cloud Storage.
```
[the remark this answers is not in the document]
janiversen: yes, not to my liking, but the old code did it, so this code does it too, a.o. to avoid upgrade issues.

core/settings/qPrefDisplay.cpp
janiversen: done (copy/paste is a dangerous thing)

core/settings/qPrefDiveComputer.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+static const QString group = QStringLiteral("DiveComputer");
+
+qPrefDiveComputer::qPrefDiveComputer(QObject *parent) : QObject(parent)
+{
+}
+qPrefDiveComputer *qPrefDiveComputer::instance()
+{
+    static qPrefDiveComputer *self = new qPrefDiveComputer;
+    return self;
```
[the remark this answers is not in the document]
janiversen: done

core/settings/qPrefCloudStorage.h
janiversen: corrected, but remark it is not an error to add ; for many it is a natural way to close a statement (just like it mandatory in class definitions)

core/settings/qPrefDisplay.cpp
bstoeger: > yes but then we would create 2 Q_OBJECTS, and have a qsettings variable in the interface classes. We would also introduce multiple inheritance something which I prefer to aviod.

This explanation doesn't make sense to me:
- I don't see why this would mean multiple inheritance (which is in many cases perfectly alright): QObject > BaseClass > DerivedClass.
- The whole point of interface classes is to have some kind-of multiple inheritance in languages which don't allow it (in my opinion for dubious reasons).
- Also note that - strictly speaking - an interface is a class where all functions are pure virtual. So your class is not an interface in the classical OO sense.

Ultimately, I believe that the best course of action would be removing `copy_txt` from class context make it a helper function. This could be in a private context but even `qthelper.cpp` would be fine. As far as I can tell, this function is not only needed  in preferences code. But perhaps `set_string` would be a better name. The copying is only half of the function.

core/settings/qPrefCloudStorage.h
bstoeger: > corrected, but remark it is not an error to add ; for many it is a natural way to close a statement (just like it mandatory in class definitions)

We discussed this before. :) There is *no* semicolon at the end of a function definition. You're simply adding random ';' to your class. The compiler will happily compile this:
```
class A {
    ;;;;;;;;;
};
```
I'm not sure if you are aware, but this is likewise valid outside of class context:
```
void f()
{
};
```

core/settings/qPrefDiveComputer.h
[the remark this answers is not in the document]
janiversen: added inline. for very simple functions like 
- calling another function
- returning a variable
the compiler optimizes the inline function away, so e.g.
->load() ends up being ->loadsync(false)
x = qPrefDiveComputer::instance->janVar(9 ends up being x = prefs.janVar
this is a lot more efficient, and also think of it code maintenance wise, having getters adds 4lines pr variable.

I am against using inline for anything complex (an “if” should e.g. never be used) but for these functions which are really just a macro it does not seem logical to add runtime and maintenance costs.

core/settings/qPrefDiveComputer.h
janiversen: actuslly it is not. This instruct the compiler to use inline where possible (see above) but leave 1 static copy of the function for the linker.

If I remove inline, we will not benefit from the normal compiler optimization.

But I have no special heart here, if the consensus is not to use what clang offers.

core/settings/qPrefDiveComputer.h
janiversen: 1. yes it could but I am no fan a of automatic conversions, I have spent too many hours debugging why “automatic” messed up my code. 
2. no need to spent time learning my basic compiler functions, I know what I did and did it for a reason (clang generates different code in the caller). changed

added new commits to make the same change in other qPref classes

core/settings/qPrefDiveComputer.h
janiversen: This is not a matter of taste, you talk about 2 different situations
The first one &var is old C style, used in statement like x=&y meaning “take address of”
The second is C++ function declaration style, meaning “get reference to” (so you avoid calling with func(&var)

I am aware the actual assembler code generated is the same, but the meaning is different, which is why it is normal to use
type& var in function declarations. 
as an anekdote, this was an idiologic fight to Bjarne, who wanted another symbol to mark the difference clearly.

janiversen:
I have now responded to all comments, changedca lot but all. Some of the comments (especially one) I do not agree with and have instead argumented why my code is as it is.

I have had limited possibility to test the changes, so I am primarely awaiting the verdict of Travis

core/settings/qPrefDisplay.cpp
janiversen: The class would inherit from QObject and the common class. And each class in the list has is own Q_OBJECT, so this only blows ups the runtime size.

it is not true that interface classes need to have pure virtual functions, you forget the very common interface-implement design. Interface classes are often used to hide the implementation details which is by no means dubious, but a sane way to keep a stable interface.

I do agree that set_string or set_text_property would be a better name, I can correct thatin a later commit. If I remove it from the class context it is no longer privat (see my use of friend). 

It is not used anywhere else, but I can easily move it to qthelper if wanted.

core/settings/qPrefCloudStorage.h
janiversen: i am aware of many ways to misuse the compiler, I have wriiten enough backends (p-code to assembler) to see that, and I know we discussed it before.

I am aware you try to remove every extra not strictly needed term, I look at it differently, if it does not generate extra code why not make the code easier to read (think of your automatic conversions as well).

However do note that the code is changed to match your style.

core/settings/qPrefDiveComputer.h
bstoeger: You are completely wrong on this one.

Defining a function in a class makes it inline, period. See first paragraph here: https://en.cppreference.com/w/cpp/language/inline
> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.

It *must* be this way, otherwise your program would fail to llink. I think you misunderstood what the `inline` specifier does: It means that you can define the same function in multiple-translation units without the linker complaining. The compiler is free to choose whether it actually "inlines" the function.

So the `inline` here is 100% redundant. If clang interprets a redundant `inline` as forceinline or alwaysinline attribute (could you please provide a link for that?), then this is voodoo and we should not poison the code-base with such compiler-idiosyncrasies. If you insist on forcing inline, please do it in a compiler independent way using the `forceinline` and `always_inline` attributes.

janiversen: Please I am not completely wrong, it is and was inline (allowing the compiler to choose), however specifying it does not make it wrong ! @dirkhh was in doubt, the word “inline” albeit not needed, makes it clear to everyone.

I was perfectly happy with the old code, but the comments from you and @dirkhh called for a change.

janiversen: You really do not need to explain compiler basics, and I do not want to fight a war over idiology, C++ allows many things that are not 100% needed, but are nice to avoid confusion by people who are less C++ gurus.

I will leave it to @dirkhh to decide what he wants with this PR. It is really a bad motivation, to tell me “I am completely wrong”, when I try hard to react positive to changes requested.

core/settings/qPrefDiveComputer.h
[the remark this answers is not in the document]
bstoeger: > I am against using inline for anything complex (an “if” should e.g. never be used)

Note that following this rule, things like
```
    static inline const QString cloud_base_url() { return prefs.cloud_base_url; }
```
should probably not be inlined. I would wager to guess that this has more than one `if`.

core/settings/qPrefDiveComputer.h
bstoeger: > I was perfectly happy with the old code, but the comments from you and @dirkhh called for a change.

Story of my life... :-/

core/settings/qPrefDiveComputer.h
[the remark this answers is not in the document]
janiversen: Do we read the same code? it is a single return statement, I really wonder where you see any “if” statement !

 the compiler removes the function call, the return and replaces it with the variable, so e.g. an assigment becomes direct x = prefs.cloud_base_url.

and just to follow up, neither p-code nor assembler use “if” they have other names.

core/settings/qPrefDiveComputer.h
bstoeger: Note that the "completely wrong" was with respect to the claim that removing the `inline` keyword disables compiler optimization. I stand by that, even though I could have worded it less strongly. I think you take things too personal, there's no offense intended from my side whatsoever.

janiversen: Well the sentence “you are completely wrong” is kind of personal, and if you read my comment I explained why “static inline” makes sense, not why load() should have the inline keyword.

core/settings/qPrefDiveComputer.h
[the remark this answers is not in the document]
bstoeger: >  it is a single return statement, I really wonder where you see any “if” statement !

In `QString`'s constructor. There is a strlen - if your compiler inlines that (for me it doesn't) then you have a whole loop in there. :)

core/settings/qPrefDiveComputer.h
bstoeger: Sorry, but this is not true. The argument went like this (verbatim quotes):
>> inline is redundant here.

> actuslly it is not. [...] If I remove inline, we will not benefit from the normal compiler optimization.

That's plain and simply wrong; sorry for being so blunt about this. This is neither an ideology nor a style question, simply a fact. And the whole thread proves that someone *does* have to explain language basics. I honestly fail to see the problem with that.

PS: I am very far from being a C++ expert, much less a guru.

dirkhh: _**SIGH**_
We are all working on the same project, in our spare time, because it's fun.
No one is trying to attack anyone else. No one should feel attacked.
Almost all of us are using their second, or third, or fourth language for communication.

I really wish I wasn't starting to dread opening the discussion threads on PRs - but lately that's exactly what I've been feeling. And no, this is not a subtle, backwards way to criticize any one individual, either.

The purpose of the discussions on PRs is to make code better, to make sure your stupid, under-educated maintainer actually understands what's going on, and that overall the project is going in the right direction.

That can and does include disagreements. But those disagreements by definition are about the facts at stake. "you are wrong" sounds personal, but it's actually still about the facts at stake. When Linus tells me "you are an idiot" he is (a) correct and (b) crossing the line towards attacking the person (and he has had my explicit consent to that for 25+ years).

I cannot tell you how much I appreciate @janiversen 's work on cleaning up the C++/QML interface, much of which is broken because I wrote it and had no idea what I was doing. And I cannot tell you how much I appreciate that @bstoeger helps explain things to me that are outside my comfort zone. Because I honestly didn't know the subtleties of const, static, inline, references and what not in these class definitions and declarations. I love it when he corrects my misconceptions. And I appreciate it when he responds to my questions.

Does this mean that I'm potentially the wrong maintainer and should pass on the baton to someone else? I don't know. You tell me.

But none of this is picking on anyone, none of this is personal, none of this is rude or offensive.
It's intended to make the code better.

I don't want to lose the fun that is maintaining Subsurface (the assholes insulting me on FB for the shit software that I apparently force them to look at are bad enough, TBH). So let's all please remind ourselves why we do this and what our goal is. A better C++/QML interface that allows us to work on cool features for Subsurface-mobile (and Subsurface).

I'll get off my soap box. It feels like I said all of this before.

core/settings/qPrefPrivate.cpp
dirkhh: for the idiot here... what's the difference between `const char **name` and `const char *&name` in this context?

core/settings/qPrefPrivate.h
dirkhh: I had this in other classes as well - over time we had a number of classes that ended up all being QObjects because things broke otherwise and I didn't understand why.

core/settings/qPrefDisplay.h
```diff
@@ -22,11 +23,11 @@ class qPrefDisplay : public QObject {
...
-	const QString divelist_font() const;
-	double font_size() const;
-	bool display_invalid_dives() const;
-	bool show_developer() const;
-	const QString theme() const;
+	static inline const QString divelist_font() {return QString(prefs.divelist_font); };
+	static inline double font_size() {return prefs.font_size; };
+	static inline bool display_invalid_dives() {return prefs.display_invalid_dives; };
+	static inline bool show_developer() {return prefs.show_developer; };
+	static inline const QString theme() {return QString(prefs.theme); };
```
dirkhh: so, just to make sure I understand this correctly - the `inline` is redundant but a reminder to the reader that the compiler may decide to inline this. And the fact that we want the compiler to inline this is why the implementation is in the `.h` instead of the `.cpp`, correct?

core/settings/qPrefAnimations.h
```diff
@@ -18,7 +19,7 @@ class qPrefAnimations : public QObject {
 	void sync() { loadSync(true); }
 
 public:
-	int animation_speed() const;
+	static inline int animation_speed() { return prefs.animation_speed; };
```
dirkhh: same here as the last comment.
And, BTW, from a white space perspective, I like this version with a space inside the `{ ` and ` }` better than the `{return...` white space in the earlier commits.
I guess this is something I should add to coding style now that we have more instances of `{` not being followed by a newline...

core/settings/qPrefPrivate.cpp
bstoeger: In this context the difference is purely syntactical - i.e. irrelevant.

In other contexts there are subtle differences between pointers and references. Notably: you can't "reseat" references - they reference the same object throughout their lifetime. And there is no null-reference. From a conceptual point-of-view, a reference is just a name for a different object.

core/settings/qPrefDisplay.h
bstoeger: 1) If you want it inlined *and* used from different parts of the codebase, you should obviously put it in a header file, yes.
2) There are two ways of making a member function inline. Firstly, the common way, outside of the class definition:
```
class {
    int f();
};
[...]
inline int f()
{
    return -1;
}
```
I think this is the cleaner approach, as it separates function declaration from function definition. Some people collect the inlined functions and templates in a separate "*_impl.h" header. Inlining is an implementation detail. Why would the user of the function care if it's inline? That decision should be up to the writer of the function.
But you can also do it by the function-definition in class-definition method:
```
class {
    int f() { return -1; }
};
```
This is implicitly inline. It *has* to be, because otherwise the program couldn't link if you use the class in different translations units. The advantage is less lines of code. @janiversen had a mess of sometimes with and sometimes without the keyword. Unfortunately, he interpreted my comment as a request to add it everywhere (story of my life). Still that's better than inconsistency.
3) I find the whole inlining practice questionable once `QString` is involved. I tried it and `gcc7` is surprisingly smart about *not* inlining too much. For example the `strlen` is not inlined for me. So it really depends on in how many places these functions are called. If they are called in one or no places, inlining is of course a win. If they are called numerous times, it is a loss. In the end, we probably should rely on the compiler to do the right thing.

core/settings/qPrefDisplay.cpp
bstoeger: > The class would inherit from QObject and the common class.

No! The base (common) class derives from `QObject` - then the derived classes are automatically `QObjects`. They don't derive again from `QObject`. That would be a disaster from Qt's object ownership point of view (which is already less-than-pretty to start with).

core/settings/qPrefDiveComputer.h
dirkhh: well, technically it IS a matter of taste, but I'm ok with adding this as a different category to the coding style document.

core/settings/qPrefCloudStorage.cpp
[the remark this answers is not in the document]
dirkhh: We could parse both the old and a new location, and only write a new location.
We have done this in other cases where we wanted to move things.
I am NOT suggesting that this is a request to change this, just pointing out the options.

core/settings/qPrefDiveComputer.h
bstoeger: Please no - with a few exceptions [which I'd like to fix] we use the `*` and `&`  stick to the variable name convention. Putting it to the type is just against reality. From a parser point of view, type declarations with `*` and `&` are the same. That the original elegance of C [`int *i` meaning the object pointed to by `i` is an `int` is lost for reference notation - `int &i` does not mean the address of `i` is an `int`] is irrelevant. Either we choose one or the other, but please not a mix based on dubious reasoning. Consider the line:
```
int&     i=x, j=y;
```
`i` is a reference to int and `j` is an int, which is in contradiction to what the line implies. Sorry, but that's the way it works. I have a feeling that people using the `int& i` syntax are in denial and can't accept the fact that C++, by inheritance from C, works the way it works. To those people, I would recommend moving to a "pure" language of their liking, such as rust.

So for me it's not a big deal if some isolated parts of the code violate our conventions, but let's not make it "official" and end up with a random mess. The code is inconsistent enough as it is.

janiversen: you make examples in east while I make code in west! I wrote and mean how to use & (get by reference) in function declarations.

There is a difference between “call by address” and “reference by address”, fell free to take a look at the parser. The most obvious difference is that the caller do not use & 2 examples:
foo(&x).     void foo(int *x).    this is the traditional C way.
foo(x)           void foo(int& x). is the more elegant (some see it is confusing) way, which was introduced with C++. and you will see good C++ compilers like clang use the difference, so that is the reason to also write it.

I DO NOT talk about statements, where it makes no sense to put the & on the type.

I do not care what your job is, you are surely the local C++ expert in this project, but that hardly makes you the only expert, who have the right to give c++ lessons (as you wrote you fell its needed). I personally have spent many years working on C++ backends, preprocessors and parsers (actually much less than writing application code). I do NOT comment other peoples code on “relidious” issues, nor do I give lessons in C++

I feel for @dirkhh with his statement about reading PR´s and I am sorry for ince again being the cause, but I think everyone gets a bit edgy when ones professional experience is reduced by having to take C++ lessons. You are both collaborators so nothing hinders you from changing the code once (if) this PR is merged (you have the capacity to do it in this PR, but I would see that as a not nice way to make a point).

@dirkhh have pointed at a couple of RELEVANT changes, which will be uploaded in a couple of hours, apart from that I will stand back and watch.


I am sorry writing personally in this comment, I have tried hard to not to be personal (I do NOT wish to offend anyone)j, but my professionalidm is in question here, something I cannot just neglect.

core/settings/qPrefCloudStorage.cpp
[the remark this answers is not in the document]
janiversen: Thanks for the hint I will that in a swperate commit so it is isolated.

core/settings/qPrefDiveComputer.h
bstoeger: Please, stop it.

You fail to acknowledge that `&`, `*` (and recently `&&`) have *different* meaning in expression and type declaration context. You keep mixing it up. Even *if* you prefer to put the `&`, `*` and `&&` *declarators* to the type, then you write
```
void qPrefPrivate::copy_txt(const char** name, const QString& string)
```
If you don't, then you write
```
void qPrefPrivate::copy_txt(const char **name, const QString &string)
```
A mixture of both makes no sense. Not from a parser, not from a language, not from any perspective.

I'm not judging anyone for language-savviness. It doesn't matter to me at all as long as the contributions are good. In my opinion, the only thing that may "reduce your professional experience" is your over-reaction to criticism.

janiversen:
@dirkhh I have added your suggestion on how to change location of the userid, please make an extra review of the last commit (line 189++ in qPrefCloudStorage.cpp) and let me know if It is as you suggested.

@bstoeger you are very wrong in stating that I am not open to critic, I like reviews, because I am only human and make errors, so relevant comments (as you can see) is something I correct fast, likewise when e.g. @dirkhh point out how the codebase use if/else, or even when you point at extra ; (who make sense to some people and is based on a lifetime of experience). But when you explain your way of programming as THE way instead of ONE way I object, due to my past I am sure my way of programming is safe, robust and readable, but I respect other ways because I know my way is not THE way. And to be honest when you try to explain C++ to me as if I am undergradutate I loose both patience and motivation. If prgramming THE (== your) way (e.g. write foo(int &x) instead of foo(int& x) is more important than bringing the project forward, then experienced programmers like me are in the wrong place.

I will let this PR rest for some hours while I train some OW students A way (not THE) way of diving, giving @dirkhh a chance to chime in.

core/settings/qPrefDiveComputer.h
dirkhh: I think we are mixing a coding style (=subjective, personal) discussion with a language semantics discussion.

I looked at this again and I think I would like us to be consistent across both declaration and use, across both reference, address, and dereference. The designator should tie to the variable or argument name, not to the type.

It's easy enough to handle this in cleanup - but from a project coding style perspective, I would prefer this consistency.

This isn't about anyone being right or wrong, this is the same logic by which I request that we use tabs and not spaces and that we write if/else clauses a certain way.

tcanabrava:
People, as **THE** c++ expert in the project I kindly ask that this stop, *now*. If anyone has a problem about the code, enter in contact with me personally and I'll gladly talk, I'll not accept this figth here. and Btw: I'm fine with the changes, I'm fine with the Style. I'm also keeping an eye on the reviews more as promissed to dirk.

core/settings/qPrefDiveComputer.h
tcanabrava: This is *just* a matter of style, the code will work the same, there's no difference *at all*, the assembly generated is completely the same. with all that said, stop. Some people will argue that the correct way is to use QString const &, and again, the assembly generated is the same.

dirkhh (approved):
Thanks for adding the code to migrate the settings to a saner place. That looks reasonable to me.

dirkhh:
Talking off-line with @tcanabrava and  at this point we're both ok with this being merged.
I'll work again on my code formatting script (in my infinite spare time, instead of doing something fun and/or useful) so we can stop worrying about that and simply include this as some kind of hook in GitHub...

janiversen:
Thanks for taking a decision, I expect the remaning groups to be without more changes to the base framework.

I am also trying to hunt down a problem qPrefCloudStorage created (no autoload during startup with cloud profile).

PR #1536, opened 2018-07-25
Cleanup: fix column width calculation

bstoeger:
The DiveListView constructor allocated a "tripModel" and set the
column-widths. The "tripModel" was then immediately deallocated.
Since the columnd-widths are stored in a vector of the object,
they are thus lost. Therefore, remove the whole column-setting
code. This makes the DiveTripModel::setColumnWidth() function
unused. Remove it as well.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Dead code removal. @neolit123: You were the last to touch this code, but as far as I can see - this was a no-op for a much longer time ... ?

The whole idea of storing column-widths in the model seems questionable to me. Aren't column widths very fundamentally part of the view-side?

### Changes made:

Remove dead code.
### Related issues:

None.
### Release note:

Nope.
### Documentation change:

Nope.
### Mentions:

@neolit123

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -26,7 +26,8 @@
 #include "core/metrics.h"
 
 DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelection(false), sortColumn(0),
-	currentOrder(Qt::DescendingOrder), dontEmitDiveChangedSignal(false), selectionSaved(false)
+	currentOrder(Qt::DescendingOrder), dontEmitDiveChangedSignal(false), selectionSaved(false),
+	initialColumnWidths(DiveTripModel::COLUMNS, 50)	// Set up with default length 50
 {
 	setItemDelegate(new DiveListDelegate(this));
 	setUniformRowHeights(true);
@@ -43,63 +44,15 @@ DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelec
 	setSelectionMode(ExtendedSelection);
 	header()->setContextMenuPolicy(Qt::ActionsContextMenu);
 
-	const QFontMetrics metrics(defaultModelFont());
-	int em = metrics.width('m');
-	int zw = metrics.width('0');
-
-	// Fixes for the layout needed for mac
-#ifdef Q_OS_MAC
-	int ht = metrics.height();
-	header()->setMinimumHeight(ht + 4);
-#endif
-
-	// TODO FIXME we need this to get the header names
-	// can we find a smarter way?
-	tripModel = new DiveTripModel(this);
-
-	// set the default width as a minimum between the hard-coded defaults,
-	// the header text width and the (assumed) content width, calculated
-	// based on type
-	for (int col = DiveTripModel::NR; col < DiveTripModel::COLUMNS; ++col) {
-		QString header_txt = tripModel->headerData(col, Qt::Horizontal, Qt::DisplayRole).toString();
-		int width = metrics.width(header_txt);
-		int sw = 0;
-		switch (col) {
-		case DiveTripModel::NR:
-		case DiveTripModel::DURATION:
-			sw = 8*zw;
-			break;
-		case DiveTripModel::DATE:
-			sw = 14*em;
-			break;
-		case DiveTripModel::RATING:
-			sw = static_cast<StarWidgetsDelegate*>(itemDelegateForColumn(col))->starSize().width();
-			break;
-		case DiveTripModel::SUIT:
-		case DiveTripModel::SAC:
-			sw = 7*em;
-			break;
-		case DiveTripModel::PHOTOS:
-			sw = 5*em;
-			break;
-		case DiveTripModel::LOCATION:
-			sw = 50*em;
-			break;
-		default:
-			sw = 5*em;
-		}
-		if (sw > width)
-			width = sw;
-		width += zw; // small padding
-		if (width > tripModel->columnWidth(col))
-			tripModel->setColumnWidth(col, width);
-	}
-	delete tripModel;
-
-
 	header()->setStretchLastSection(true);
 
 	installEventFilter(this);
+
+	// TODO: We use a dummy DiveTripModel to calculated column widths.
+	// Change this to a global object.
+	DiveTripModel tripModel;
+	for (int i = DiveTripModel::NR; i < DiveTripModel::COLUMNS; i++)
+		calculateInitialColumnWidth(tripModel, i);
 }
 
 DiveListView::~DiveListView()
@@ -111,13 +64,52 @@ DiveListView::~DiveListView()
 		if (isColumnHidden(i))
 			continue;
 		// we used to hardcode them all to 100 - so that might still be in the settings
-		if (columnWidth(i) == 100 || columnWidth(i) == tripModel->columnWidth(i))
+		if (columnWidth(i) == 100 || columnWidth(i) == initialColumnWidths[i])
 			settings.remove(QString("colwidth%1").arg(i));
 		else
 			settings.setValue(QString("colwidth%1").arg(i), columnWidth(i));
 	}
 	settings.remove(QString("colwidth%1").arg(DiveTripModel::COLUMNS - 1));
 	settings.endGroup();
+}
+
+void DiveListView::calculateInitialColumnWidth(const DiveTripModel &tripModel, int col)
+{
+	const QFontMetrics metrics(defaultModelFont());
+	int em = metrics.width('m');
+	int zw = metrics.width('0');
+
+	QString header_txt = tripModel.headerData(col, Qt::Horizontal, Qt::DisplayRole).toString();
+	int width = metrics.width(header_txt);
+	int sw = 0;
+	switch (col) {
+	case DiveTripModel::NR:
+	case DiveTripModel::DURATION:
+		sw = 8*zw;
+		break;
+	case DiveTripModel::DATE:
+		sw = 14*em;
+		break;
+	case DiveTripModel::RATING:
+		sw = static_cast<StarWidgetsDelegate*>(itemDelegateForColumn(col))->starSize().width();
+		break;
+	case DiveTripModel::SUIT:
+	case DiveTripModel::SAC:
+		sw = 7*em;
+		break;
+	case DiveTripModel::PHOTOS:
+		sw = 5*em;
+		break;
+	case DiveTripModel::LOCATION:
+		sw = 50*em;
+		break;
+	default:
+		sw = 5*em;
+	}
+	if (sw > width)
+		width = sw;
+	width += zw; // small padding
+	initialColumnWidths[col] = std::max(initialColumnWidths[col], width);
 }
 
 void DiveListView::setupUi()
@@ -137,7 +129,7 @@ void DiveListView::setupUi()
 		if (width.isValid())
 			setColumnWidth(i, width.toInt());
 		else
-			setColumnWidth(i, tripModel->columnWidth(i));
+			setColumnWidth(i, initialColumnWidths[i]);
 	}
 	settings.endGroup();
 	if (firstRun)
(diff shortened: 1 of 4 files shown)


neolit123:
i think that my patches:
https://github.com/bstoeger/subsurface/commit/e3118d915cf49de4cc722522031b0be44014d790#diff-556d1da58f8fe4f17d42c0c3ed5ea14c
https://github.com/bstoeger/subsurface/commit/52445ec8f5bb957a41c6650982e69c96174eee8a#diff-556d1da58f8fe4f17d42c0c3ed5ea14c

didn't remove the `delete tripModel` line and that's a problem.
i wonder if the no-op is actually the bug - i.e. we don't maintain the sizes correctly.

what i think happens with my changes is that they fix the crashes but it doesn't remember the sizes, because of the `delete`.

i think turning DiveTripModel into a singleton is worth the try - e.g. `->instance()`?

> Aren't column widths very fundamentally part of the view-side

well it's a ViewModel and not a DataModel, but it doesn't matter that much.
the same question applies for the column names - `tripModel->headerData()`.

bstoeger:
Note that the `DiveTripModel` is reallocated *every single time* in `DiveListView` without the `columnWidth`s being restored. This is all a bit ... crazy. :)

> i think turning DiveTripModel into a singleton is worth the try - e.g. ->instance()?

Incidentally, this is what I made for my undo branch and this is why it occurred to me that this must be dead code. So please wait with that. It could be that there is a reason that the old code did it this way and it's not worth the risk, since I refactored `DiveTripModel` anyway.

> well it's a ViewModel and not a DataModel, but it doesn't matter that much.
the same question applies for the column names - tripModel->headerData().

I would feel that `tripModel->headerData()` belongs to the model and `tripModel->columnWidth` does not. After all, you *could* have two lists with different font sizes, no? But sure - not a big deal. I will try to come up with something that doesn't interfere too much with my other stuff.

bstoeger:
I think I fixed the issue - please test. Note: to test, you have to delete the old config file. Indeed, the widths were broken for me from the start. Probably you old-timers just don't notice because you use an old config with proper column widths?

bstoeger:
PS: I wonder if the manual setup:
```
        initialColumnWidths[DiveTripModel::NR] = 70;
        initialColumnWidths[DiveTripModel::DATE] = 140;
        initialColumnWidths[DiveTripModel::RATING] = 90;
        initialColumnWidths[DiveTripModel::SUIT] = 70;
        initialColumnWidths[DiveTripModel::SAC] = 70;
        initialColumnWidths[DiveTripModel::PHOTOS] = 5;
        initialColumnWidths[DiveTripModel::LOCATION] = 500;
```
is still necessary now?

bstoeger:
> I wonder if the manual setup [...] is still necessary now?

Matter of fact: if I remove these lines, things seem to work just as well. Am I missing something?

desktop-widgets/divelistview.cpp
```diff
@@ -26,7 +26,8 @@
 #include "core/metrics.h"
 
 DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelection(false), sortColumn(0),
-	currentOrder(Qt::DescendingOrder), dontEmitDiveChangedSignal(false), selectionSaved(false)
+	currentOrder(Qt::DescendingOrder), dontEmitDiveChangedSignal(false), selectionSaved(false),
+	initialColumnWidths(DiveTripModel::COLUMNS, 50)		// Set up with default length 50
```
neolit123: comment should be with a single space after the code line.
missing `;` at the end.

desktop-widgets/divelistview.cpp
```diff
@@ -43,63 +44,17 @@ DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelec
...
 
 	installEventFilter(this);
+
+	// TODO: We use a dummy DiveTripModel to calculated column widths.
+	// Change this to a global object.
+	DiveTripModel tripModel;
+	for (int i = DiveTripModel::NR; i < DiveTripModel::COLUMNS; i++)
+		calculateInitialColumnWidth(tripModel, i);
+
+	setupUi();
```
neolit123: i don't know what is best about `setupUi()`.

desktop-widgets/divelistview.cpp
```diff
@@ -120,6 +75,45 @@ DiveListView::~DiveListView()
...
+	case DiveTripModel::LOCATION:
+		sw = 50*em;
+		break;
+	default:
+		sw = 5*em;
+	}
+	if (sw > width)
+		width = sw;
+	width += zw; // small padding
+	initialColumnWidths[col] = std::max(initialColumnWidths[col], width);
```
neolit123, line 112: `setColumnWidth` and `columnWidths` were added because we had weird out-of-bounds bugs.
the functions had guards.

if all works well a regular array / vector should be fine, of course.

desktop-widgets/divelistview.cpp
```diff
@@ -424,7 +411,7 @@ void DiveListView::reload(DiveTripModel::Layout layout, bool forceSort)
 
 	QSortFilterProxyModel *m = qobject_cast<QSortFilterProxyModel *>(model());
 	QAbstractItemModel *oldModel = m->sourceModel();
-	tripModel = new DiveTripModel(this);
+	DiveTripModel *tripModel = new DiveTripModel(this);
```
neolit123, line 406: memory leak?
`DiveListView::reload()` is called more than once.

neolit123 (commented):
added a couple of comments.

desktop-widgets/divelistview.cpp
bstoeger: > comment should be with a single space after the code line.

Ok - will do

> missing ; at the end.

This is an initialization list. I think adding ';' would not compile..

desktop-widgets/divelistview.cpp
bstoeger: This is a private function - nobody else can call it. It's called in the loop
```
for (int i = DiveTripModel::NR; i < DiveTripModel::COLUMNS; i++)
```
and the array is initialized with
```
initialColumnWidths(DiveTripModel::COLUMNS, 50)
```
In my opinion, this is guaranteed to work. Note that DiveTripModel::COLUMNS is a constant (enum) and can't change during the application's lifetime. Also note that initialization and function call is only in the constructor - I think this is really safe.

desktop-widgets/divelistview.cpp
bstoeger: No, the old model is deleted a few lines later (that's the reason for fetching `oldModel`).

But you are right: this code is really shoddy. I have a commit to replace this with a "singleton". Though since the object is only used in this class, we might even make it a sub-object.

desktop-widgets/divelistview.cpp
bstoeger: Access to the array is only in the `destructor` and `setupUI`. But there, likewise all is in a loop of the above kind.

So I think the out-of-bound accesses must have had a different cause.

neolit123:
@bstoeger 
please, mind that with the current master it's possible to resize a column and on restart the size will be remembered. with this patch all the columns start with the same width and the custom sizes are not remembered.

dirkhh:
Oh my. This is code that has been around for a long time and that was incredibly painful to get to work. Some of this may be working around situations that no longer apply today or bugs in Qt that are no longer there, but please be careful there.
The goal was to have an out of the box experience that looks somewhat reasonable - equally sized columns look terrible and especially on small screens (1024x600 was always a design goal for us) you don't get to see much.
Obviously, the goal was to remember changes that you made to the width of the columns.

I haven't paid attention to this in a while, but I haven't noticed it being broken when trying Subsurface in a new VM or on a new computer...

janiversen:
not remembering changed column width is not a good thing. I use that a lot when accessing my mac vm from my ipad.

desktop-widgets/divelistview.cpp
```diff
@@ -407,13 +401,6 @@ void DiveListView::headerClicked(int i)
 
 void DiveListView::reload(DiveTripModel::Layout layout, bool forceSort)
 {
-	// we want to run setupUi() once we actually are displaying something
-	// in the widget
-	static bool first = true;
-	if (first && dive_table.nr > 0) {
-		setupUi();
-		first = false;
-	}
```
dirkhh, line 416: this was a critical fix way back when it was introduced. Because column widths that you set when nothing was displayed weren't kept.
My guess is that this is a Qt bug that was fixed?

bstoeger:
> please, mind that with the current master it's possible to resize a column and on restart the size will be remembered. with this patch all the columns start with the same width and the custom sizes are not remembered.

Should be fixed by moving `setupUI()` back. Sigh. :/

> Oh my. This is code that has been around for a long time and that was incredibly painful to get to work. Some of this may be working around situations that no longer apply today or bugs in Qt that are no longer there, but please be careful there.

I *am* careful, as in testing different situations (new user / new divelog) that come to mind. But of course I can't guarantee that I'm not inadvertently removing an obscure fix. The code *is* convoluted after all. I didn't test the loading-from-prefs, because I always started fresh. Now I do.

> I haven't paid attention to this in a while, but I haven't noticed it being broken when trying Subsurface in a new VM or on a new computer...

Well, it's not broken in the sense that it's not usable, just the columns don't adapt to the system-font size and are too small for my tastes. In current master, the column widths are fixed-size according to the code in `DiveTripModel`. There is a rather long-winded piece of code in `DiveListView`, which does some heuristics with the current font, but the results of these calculations are trashed. I can reproduce the difference by running an old and a new instance. Note: to see the difference you have to have a log opened, because the `setupUI()` is back to be called only when an actual dive-log is displayed.

This PR is supposed to make the dead-code work. You might of course say that the fixed-size thing is preferred, then we can just remove the dead code. Fine with me.

 TBH, I have some doubts that this is a Qt bug. But I'm not sure investigating this further is worth it.

BTW: Both versions do not adapt to changed font-sizes in the preferences.

dirkhh:
Just in case I wasn't clear.
I'm super excited that you are fixing this - I just wanted to provide some context and point out a few things. I love the idea that maybe this will work better with different font sizes. The broken dead code that doesn't do what it seems to do is almost certainly mine - something that I tried, checked in, and then gave up on. There's way more of that than I'm willing to admit.

Is this in a stage where you say "this is ready, please test", or is this WIP?

bstoeger:
To be honest - it was never the plan to fix/improve this - I just found that there is some dead code while refactoring the whole dive-list-model thing. :/ So don't expect anything except some removed SLOC. Also, it seems the code isn't by you. Not that it would be of any importance - no point in blaming anybody. And I guess the code worked at some point but bit-rotted.

Sure, you can test it. But the only difference to master is that the column-sizes will be different on a fresh install. I'm not sure if we want to debug the `setupUi()` thing?

bstoeger:
Ok - I understand the `setupUi()` thing now and it is
a) different than you think and
b) rather easy to fix.

It has nothing to do with whether data is loaded or not. It is simply that the column widths have to be setup *after* the source-model is set. But the model is set in `reload()`. Thus, of course, you can't call `setupUi()` in the constructor. And even if you set it *the first time at the beginning* in `reload()`, it doesn't work, because the source-model is not yet set. Thus, the actual reason why the "have we loaded something"-check works is that it fails in the first call because it is loaded very early in `MainWindow`. In the first call the source-model is set and on second call it will work (independently of whether something is loaded). So this is one case where we shouldn't blame Qt. :)

So the fix is very simple: set the model (ideally singleton) in the constructor.

bstoeger:
Pushed a temporary fix - as I said numerous times conversion to singleton I have in a different branch and don't want to risk that for master.

neolit123 (approved):
i've built this change locally and the widths are preserved.
also tried with a clean user config.

dirkhh:
This looks good and worked here. Which surprises me, given my recollection of how we got to where we were before your changes. Thanks.

bstoeger:
The old mis-behavior can be explained quite simply:
- Before setting the source model, the model has zero columns.
- When setting the columns-widths in such a state, they will simply be thrown away (arguably rightly so).

So if there is a Qt-bug involved, it's the fact that it doesn't complain about widths of non-existing columns being set.

dirkhh:
Ha - that's actually a reasonable explanation :-)
Thanks

PR #1537, opened 2018-07-25
qPrefFacebook, ready

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Builds system change
- [ ] Documentation change
- [ ] Language translation

next in the qPref series

Second commit is correcting a name in qml
Rest is qPrefFacebook
removed SAVE_FACEBOOK_CREDENTIALS

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -104,6 +104,7 @@ set(SUBSURFACE_CORE_LIB_SRCS
 	settings/qPrefCloudStorage.cpp
 	settings/qPrefDisplay.cpp
 	settings/qPrefDiveComputer.cpp
+	settings/qPrefFacebook.cpp
 	settings/qPrefPrivate.cpp
 
 	#Subsurface Qt have the Subsurface structs QObjectified for easy access via QML.
--- a/core/settings/qPref.h
+++ b/core/settings/qPref.h
@@ -9,6 +9,7 @@
 #include "qPrefCloudStorage.h"
 #include "qPrefDisplay.h"
 #include "qPrefDiveComputer.h"
+#include "qPrefFacebook.h"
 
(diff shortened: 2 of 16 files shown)


janiversen:
@dirkhh thi is right now low priority due to the qPrefCloudStorage problem, but can you please point me where the facebook settings are stored and loaded ? it does not happen in SettingsObjectWrapper nor can I find it in desktop-widgets/plugins/facebook. Thanks in advance for a code pointer.

dirkhh:
I think this is a trick question :-)
We decided very intentionally to NOT persistently store the FB credentials because of fear of other apps being able to abuse them for spamming purposes - and Subsurface getting blamed for that.
So we use the settings in memory (and that's in `desktop-widgets/plugins/facebook/facebookconnectwidget.cpp`), but we don't store this to disk.
Is that what you are asking about, or am I misunderstanding your question?

dirkhh:
Yes, I'm ok with removing this. The code was written and then simply commented out at first when we decided to not save to disk after all.

dirkhh:
The intro message for the PR doesn't seem to be correct anymore. No biggie.
The commits add the `SAVE_FACEBOOK_CREDENTIALS` handling and then remove it again. One could easily reduce the commit count and simplify this series by refactoring things so that this never gets added in the first place.
I don't want to bog you down with mechanics like this - would you like me to do the refactoring for you?
Other than this, this looks ok

janiversen:
Reduced the commits, hope travis is ok with the changes,

btw every day is DM day until september 2nd, but some days are more fun than others.

dirkhh:
Thanks for updating. Travis is happy, I am happy.

PR #1538, opened 2018-07-26
Whitespace

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This simply tries to address a few of the coding style issues. A lot more files need to be cleaned up, but it's a start.

### Changes made:

This is pretty much all done with the whitespace.pl script - in a couple of corner cases it needed small hand edits (and that's the challenge with the script - it's 99.9% perfect or something like this, but every few files there is something that it doesn't get right...
 
### Mentions:

@bstoeger @janiversen @tcanabrava 
none of this should cause concerns, this should all be what's in the CodingStyle - and if it isn't, then it should be.
And if there is something that you feel is really terrible, then (a) explain why and (b) propose a change to both CodingStyle and the script :-)

### DIFF
--- a/core/divecomputer.cpp
+++ b/core/divecomputer.cpp
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "divecomputer.h"
 #include "dive.h"
-#include "subsurface-string.h"
 #include "subsurface-qt/SettingsObjectWrapper.h"
+#include "subsurface-string.h"
 
 DiveComputerList dcList;
 
@@ -38,13 +38,13 @@ bool DiveComputerNode::changesValues(const DiveComputerNode &b) const
 
 const DiveComputerNode *DiveComputerList::getExact(const QString &m, uint32_t d)
 {
-	auto it = std::lower_bound(dcs.begin(), dcs.end(), DiveComputerNode { m, d , {}, {}, {} } );
+	auto it = std::lower_bound(dcs.begin(), dcs.end(), DiveComputerNode{m, d, {}, {}, {}});
 	return it != dcs.end() && it->model == m && it->deviceId == d ? &*it : NULL;
 }
 
 const DiveComputerNode *DiveComputerList::get(const QString &m)
 {
-	auto it = std::lower_bound(dcs.begin(), dcs.end(), DiveComputerNode { m, 0 , {}, {}, {} } );
+	auto it = std::lower_bound(dcs.begin(), dcs.end(), DiveComputerNode{m, 0, {}, {}, {}});
 	return it != dcs.end() && it->model == m ? &*it : NULL;
 }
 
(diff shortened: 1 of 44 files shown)


janiversen:
Looks OK to me, but would be even better, if it was added as a githook (sorry if I missed that it is), so it was done automatically.

core/divecomputer.cpp
```diff
@@ -38,13 +38,13 @@ bool DiveComputerNode::changesValues(const DiveComputerNode &b) const
 
 const DiveComputerNode *DiveComputerList::getExact(const QString &m, uint32_t d)
 {
-	auto it = std::lower_bound(dcs.begin(), dcs.end(), DiveComputerNode { m, d , {}, {}, {} } );
+	auto it = std::lower_bound(dcs.begin(), dcs.end(), DiveComputerNode{m, d, {}, {}, {}});
```
bstoeger, line 41: I think @neolit123 (sorry if I mis-attribute) complained that I should add spaces inside braces.

In consistency with the changes of `core/settings/qPrefDisplay.h` I would propose `DiveComputerNode{ m, d, {}, {}, {} }`, i.e. no space after the class-name, but spaces after `{` and before `}`

neolit123: > I think @neolit123 (sorry if I mis-attribute) complained that I should add spaces inside braces.

i don't think i reviewed this PR:
https://github.com/Subsurface-divelog/subsurface/pull/1386
maybe somewhere else?

the way we initialize in C is:
`int ar[] = { 0, 1, 2 };`
so `DiveComputerNode{ m, d, {}, {}, {} }` seems better.

bstoeger: Yes, I think it was in the context of a lambda, which is a bit of a different situation.

dirkhh: Some of this is of course black magic hidden somewhere in the logic of clang-format.
What I would say is that we tend to have parenthesis without spaces `bla(3, "no")`, but braces with no code behind them (so the opening brace that starts a code block is typically followed by a newline) - but in the rare instances where we went for the "tight" code model `void bla(int x) { return x + 1; }` we typically have spaces there as well.
I like spaces better and this was one of the changes I wasn't as excited about. But that means someone needs to figure out how to teach `whitespace.pl` how to do this right.

dirkhh:
> Looks OK to me, but would be even better, if it was added as a githook (sorry if I missed that it is), so it was done automatically.

I can see why you would think this. But given the sometimes insanely stupid things that the tool wants to do, and the fact that it seems impossible to get it to always do the right thing - or at least something that seems to make sense to me, automating this seems fraught with problems. Because sometimes we would WANT the code submitted to fail that test.

And then, of course, I have NO IDEA how to do this with a hook :-)

core/divecomputer.cpp
neolit123: should we merge this as is?

dirkhh: I think so. And then fix `whitespace.pl` and go over things again to make them consistent.

But then I may be biased.

PR #1540, opened 2018-07-26
fix to qPrefCloudStorage git_url is derived from base_url

janiversen:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

old version of qPrefCloudStorage loaded/saved git_url from/to disk, whereas 
SettingsObjectWrapper derived git_url from base_url (kind of ugly, but)

Code changed to derive git_url from base_url,
test cases updated to test the relationship between base_url and git_url.

### Documentation change:
·

### Mentions:

@dirkhh please test that all your multipke issues are solved
@sfuchs79 I am not sure this solves your case, so please test (I still fo not have a valid local profile).

### DIFF
--- a/core/settings/qPrefCloudStorage.cpp
+++ b/core/settings/qPrefCloudStorage.cpp
@@ -17,7 +17,6 @@ qPrefCloudStorage *qPrefCloudStorage::instance()
 void qPrefCloudStorage::loadSync(bool doSync)
 {
 	disk_cloud_base_url(doSync);
-	disk_cloud_git_url(doSync);
 	disk_cloud_storage_email(doSync);
 	disk_cloud_storage_email_encoded(doSync);
 	disk_cloud_storage_password(doSync);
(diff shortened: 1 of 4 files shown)


janiversen:
the failure on travis qt59, looks like a travis update problem, not caused by this fix

sfuchs79:
Result of very quick (5 seconds) test: Yes, fixes my issue. But will continue testing next week. I'm on the way to the airport...

core/settings/qPrefCloudStorage.cpp
```diff
@@ -16,8 +18,8 @@ qPrefCloudStorage *qPrefCloudStorage::instance()
 
 void qPrefCloudStorage::loadSync(bool doSync)
 {
+	qDebug() << " JAN load was called\n";
```
dirkhh: :-)
I do that all the time...

core/settings/qPrefCloudStorage.cpp
```diff
@@ -2,6 +2,8 @@
 #include "qPref.h"
 #include "qPrefPrivate.h"
 
+#include <QDebug>
+
```
dirkhh: doesn't hurt, but I think once your messages are removed, we don't need it anymore

core/settings/qPrefCloudStorage.cpp
```diff
@@ -28,6 +30,10 @@ void qPrefCloudStorage::loadSync(bool doSync)
 	disk_save_password_local(doSync);
 	disk_save_userid_local(doSync);
 	disk_userid(doSync);
+	qDebug() << "baseurl: " << QString(prefs.cloud_base_url) << " ok\n";
+	qDebug() << "giturl: " << QString(prefs.cloud_git_url) << " ok\n";
+	qDebug() << "git local: " << (prefs.git_local_only ? "true" : "false") << " ok\n";
+
```
dirkhh: If any of these paths should be reported with '-v', then let's do that there (we do report the 'cloud URL' already)

dirkhh (changes requested):
You left a few qDebugs in there...

dirkhh:
If you can't get to it today (I think you are working as DM again this week), let me know and I can edit your commits and remove the debug code... whatever works for you

janiversen:
Thanks for the suggestion, I just returned from a “deep” dive (AW) and a rescue course, so I need to relax with a bit of normal work.

Pushed an update, sorry for letting that slip.

dirkhh:
Where "_deep_" is what... 20m? :-)
I'll look at things over lunch (within the next hour or so)

janiversen:
30m (Padi regulation), max depth for padi advanced divers, hence the “”. But ok all beginnings are difficult, and the student was actually quite good.

dirkhh (approved):

PR #1542, opened 2018-07-28
 qPrefProxy

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

next in the rewrite series
### Changes made:

commit 1) add needed HANDLE_ to qPrefPrivate
commit 2) correct misleading test_multiple in all qPref test classes
commit 3) cleaned register_type in subsurface-helper.
the use of a very dimple macro makes the whole block a lot easier to read.
rest is qPrefProxy.

### Mentions:

@dirkhh please look at commit 3 and see if you like it and compare with the old version.

### DIFF
--- a/core/settings/qPrefProxy.cpp
+++ b/core/settings/qPrefProxy.cpp
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "qPref.h"
+#include "qPrefPrivate.h"
+
+#include <QNetworkProxy>
+
+static const QString group = QStringLiteral("Network");
+
+qPrefProxy::qPrefProxy(QObject *parent) : QObject(parent)
+{
+}
+qPrefProxy *qPrefProxy::instance()
+{
+	static qPrefProxy *self = new qPrefProxy;
+	return self;
+}
+
+void qPrefProxy::loadSync(bool doSync)
+{
+	disk_proxy_auth(doSync);
+	disk_proxy_host(doSync);
+	disk_proxy_pass(doSync);
+	disk_proxy_port(doSync);
+	disk_proxy_type(doSync);
+	disk_proxy_user(doSync);
+}
+
+HANDLE_PREFERENCE_BOOL(Proxy, "/proxy_auth", proxy_auth);
+
+HANDLE_PREFERENCE_TXT(Proxy, "/proxy_host", proxy_host);
+
+HANDLE_PREFERENCE_TXT(Proxy, "/proxy_pass", proxy_pass);
+
+HANDLE_PREFERENCE_INT(Proxy, "/proxy_port", proxy_port);
+
+HANDLE_PREFERENCE_INT_DEF(Proxy, "/proxy_type", proxy_type, QNetworkProxy::DefaultProxy);
+
+HANDLE_PREFERENCE_TXT(Proxy, "/proxy_user", proxy_user);
--- a/core/settings/qPrefProxy.h
+++ b/core/settings/qPrefProxy.h
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef QPREFPROXY_H
+#define QPREFPROXY_H
+#include "core/pref.h"
+
+#include <QObject>
+
+
+class qPrefProxy : public QObject {
+	Q_OBJECT
+	Q_PROPERTY(bool proxy_auth READ proxy_auth WRITE set_proxy_auth NOTIFY proxy_auth_changed);
+	Q_PROPERTY(QString proxy_host READ proxy_host WRITE set_proxy_host NOTIFY proxy_host_changed);
+	Q_PROPERTY(QString proxy_pass READ proxy_pass WRITE set_proxy_pass NOTIFY proxy_pass_changed);
+	Q_PROPERTY(int proxy_port READ proxy_port WRITE set_proxy_port NOTIFY proxy_port_changed);
+	Q_PROPERTY(int proxy_type READ proxy_type WRITE set_proxy_type NOTIFY proxy_type_changed);
+	Q_PROPERTY(QString proxy_user READ proxy_user WRITE set_proxy_user NOTIFY proxy_user_changed);
+
+public:
+	qPrefProxy(QObject *parent = NULL);
+	static qPrefProxy *instance();
+
+	// Load/Sync local settings (disk) and struct preference
+	void loadSync(bool doSync);
+	void inline load() { loadSync(false); }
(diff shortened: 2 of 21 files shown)


core/settings/qPrefProxy.cpp
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "qPref.h"
+#include "qPrefPrivate.h"
+
+#include <QNetworkProxy>
+
+
```
dirkhh: excess empty lines :-)

core/settings/qPrefProxy.cpp
```diff
@@ -0,0 +1,39 @@
...
+HANDLE_PREFERENCE_TXT(Proxy, "/proxy_host", proxy_host);
+
+HANDLE_PREFERENCE_TXT(Proxy, "/proxy_pass", proxy_pass);
+
+HANDLE_PREFERENCE_INT(Proxy, "/proxy_port", proxy_port);
+
+HANDLE_PREFERENCE_INT_DEF(Proxy, "/proxy_type", proxy_type, QNetworkProxy::DefaultProxy);
+
+HANDLE_PREFERENCE_TXT(Proxy, "/proxy_user", proxy_user);
```
dirkhh, line 38: that's pretty :-)

core/settings/qPrefProxy.h
```diff
@@ -0,0 +1,60 @@
...
+
+
+class qPrefProxy : public QObject {
+	Q_OBJECT
+	Q_PROPERTY(bool proxy_auth READ proxy_auth WRITE set_proxy_auth NOTIFY proxy_auth_changed);
+	Q_PROPERTY(QString proxy_host READ proxy_host WRITE set_proxy_host NOTIFY proxy_host_changed);
+	Q_PROPERTY(QString proxy_pass READ proxy_pass WRITE set_proxy_pass NOTIFY proxy_pass_changed);
+	Q_PROPERTY(int proxy_port READ proxy_port WRITE set_proxy_port NOTIFY proxy_port_changed);
+	Q_PROPERTY(int proxy_type READ proxy_type WRITE set_proxy_type NOTIFY proxy_type_changed);
+	Q_PROPERTY(QString proxy_user READ proxy_user WRITE set_proxy_user NOTIFY proxy_user_changed);
```
dirkhh, line 16: am I mistaken or is this a six space indent? 😱

core/settings/qPrefProxy.h
```diff
@@ -0,0 +1,60 @@
...
+
+public:
+	qPrefProxy(QObject *parent = NULL);
+	static qPrefProxy *instance();
+
+	// Load/Sync local settings (disk) and struct preference
+	void loadSync(bool doSync);
+	void inline load() { loadSync(false); }
+	void inline sync() { loadSync(true); };
```
dirkhh: As we all know, thanks to @bstoeger , that second `;` after the `}` isn't needed and should be dropped for consistency...

tests/testqPrefDisplay.cpp
```diff
@@ -109,26 +109,16 @@ void TestQPrefDisplay::test_struct_disk()
...
-	prefs.font_size = 11.0;
-	prefs.show_developer = true;
-	prefs.theme = copy_qstring("myTheme");
-	auto display_direct = new qPrefDisplay;
 
-	prefs.display_invalid_dives = true;
-	prefs.divelist_font = copy_qstring("multipleCharsInString");
-	prefs.font_size = 15.0;
-	prefs.show_developer = false;
-	prefs.theme = copy_qstring("myTheme8");
```
dirkhh, line 124: why did all these get removed? Or why were they there in the first place?

subsurface-helper.cpp
```diff
@@ -147,55 +147,32 @@ static void register_meta_types()
 }
 #endif // not SUBSURFACE_TEST_DATA
 
+#define REGISTER_TYPE(useClass, useQML) \
+	rc = qmlRegisterType<useClass>("org.subsurfacedivelog.mobile", 1, 0, useQML); \
+	if (rc < 0) \
+		qDebug() << "ERROR: Cannot register useQML (class useClass), QML will not work!!";
+
```
dirkhh: that's a good idea. Makes the code much easier to read.

dirkhh (approved):
A few small comments, but overall I think this continues the series

subsurface-helper.cpp
bstoeger: I don't think that works as intended (that's not how the parser / preprocessor tokenizes):

```
#include <cstdio>
#include <typeinfo>

#define TEST1 xyz
#define TEST2 "abc"

struct test
{
        int a;
};

int main()
{       
        test x, *y;
        printf("XYZ TEST1 XYZ\n");
        printf("ABC " TEST2 " ABC\n");
        printf("type-name of int: %s\n", typeid(int).name());
        printf("type-name of x: %s\n", typeid(x).name());
        printf("type-name of y: %s\n", typeid(y).name());
}
```
gives 
```
XYZ TEST1 XYZ
ABC abc ABC
type-name of int: i
type-name of x: 4test
type-name of y: P4test
```

Unfortunately, `typeid.name()` gives (on gcc) mangled names and "demangling" is not portable. On the other hand, for a debug message a mangled name might be enough.

So you could turn that into a template function (untested):
```
template <typename T>
void registerType(const char *useQML)
{
	if (qmlRegisterType<T>("org.subsurfacedivelog.mobile", 1, 0, useQML) < 0)
		qDebug() << QStringLiteral("ERROR: Cannot register %1 (class %2), QML will not work!!".arg(useQML, typeinfo(T).name());
}
```
And it might be a good idea to turn this from `qDebug` to `qWarning`.

core/settings/qPrefProxy.cpp
janiversen: done

core/settings/qPrefProxy.h
janiversen: ? it is a single tab in my version ?

core/settings/qPrefProxy.h
janiversen: done

subsurface-helper.cpp
janiversen: good catch, I forgot the parameter was inside “”. corrected

core/settings/qPrefProxy.h
dirkhh: GitHub misled me. Sorry

Issue #1541, opened 2018-07-28
DCDeviceData.bluetooth_mode true even w/o BT_SUPPORT


kristapsdz:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When BT_SUPPORT is *not* enabled, the "use bluetooth" checkbox is not displayed (cf. desktop-widgets/downloadfromdivecomputer.cpp) and setBluetoothMode is never called.  However, DCDeviceData (core/downloadfromdcthread.cpp) sets bluetooth_mode to true by default.  Since the button does not engage without BT_SUPPORT and is checked by default, the downloader always tries to use BT even if it's not compiled in.  This makes downloading from any DC impossible.

I've locally patched core/downloadfromdcthread.cpp as:

```c
#ifdef BT_SUPPORT
        data.bluetooth_mode = true;
#else
        data.bluetooth_mode = false;
#endif
```

And everything works fine.  This is on OpenBSD, which doesn't have bluetooth at all.

Alternatively, desktop-widgets/downloadfromdivecomputer.cpp could be patched to setBluetoothMode(false) if the CPP check for BT_SUPPORT is false.

(I can make a git pull, but it's a tiny patch and I'm still getting the hang of git...)

dirkhh:
I'd love to see a pull request.
Practice makes perfect.

PR #1543, opened 2018-07-29
When BT_SUPPORT is not enabled, don't default to "true" for using blu…

kristapsdz:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

BT_SUPPORT, if undefined, also needs the bluetooth_mode value in DCDeviceData to be false by default.  Otherwise, it's never changed from true due to the ifdef in DownloadFromDCWidget, which results in the system trying to use bluetooth even if it's not there!

With this patch, I can download dives on OpenBSD 6.3 with the current codebase.

This was originally noted at https://github.com/Subsurface-divelog/subsurface/issues/1541.

### DIFF
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -216,7 +216,11 @@ DCDeviceData::DCDeviceData()
 	data.download_table = nullptr;
 	data.diveid = 0;
 	data.deviceid = 0;
+#if defined(BT_SUPPORT)
 	data.bluetooth_mode = true;
+#else
+	data.bluetooth_mode = false;
+#endif
 	data.force_download = false;
 	data.create_new_trip = false;
 	data.libdc_dump = false;


dirkhh:
Thanks for submitting this.
If you are interested in learning a bit more about submitting PRs, git, and GitHub oddities, here's what I suggest you should do:
- typically, don't do development in your master branch. When you work on something like this, do `git checkout -b fixBt` or some other name that makes sense to you and then push from that branch (no need to fix that in this case, but maybe something to think about in the future)
- there are some best practices about commit messages https://chris.beams.io/posts/git-commit/ - in your case, please shorten the Subject line of the commit; something like `BT support: disable BT download when compiling without BT`
- in your Signed-off-by line, add `<` `>` around the email address, just as I did when I edited https://github.com/Subsurface-divelog/subsurface/commit/51066e5478d76824c5da53f37184e0e0d1f3e4af
- finally, GitHub allows some commands in commit messages. Since this fixes the issue 1541, you can simply have the following in a line by itself in the commit message and it will automagically close the issue once your commit is merged: `Fixes #1541`

Lots of suggestions for a tiny commit - I hope this isn't overwhelming. All of this can be addressed with a single command and some editing, though.
In your development branch (for you, `master`), simply issue this command:
```
git commit --amend
```
git will open your editor and allow you to re-edit the current commit message (which should be the one of this commit). Simply change the Subject line, fix the Signed-off-by line, and add the Fixes line and save the commit message again. Now force push the updated commit to GitHub (the syntax for that depends a bit on how you have things set up, but simply add a `-f` as argument to the push that you did to create this PR) and the rest should work automatically.

If you have questions about any of this, many of us here will be happy to help.
If you'd rather not deal with this, simply let me know and I'll clean it up when manually merging your commit.

Thanks

kristapsdz:
@dirkhh , will do.  I appreciate the suggestions and will do them---I've a few more diffs coming your way, so anything that makes quick merging easier helps!

dirkhh:
... in that case... maybe DO move this to a branch. And please don't do merges in that branch (that was one of the issues with the last PR that also came from master... you pulled from upstream in the middle of things). If at any point you have questions, just ask - I always enjoy getting new developers to contribute to our project :-)

dirkhh:
Excellent - thanks for updating this. Looks perfect now. Since you didn't change the code I see no point in waiting for Travis

PR #1544, opened 2018-07-29
Cleanup: remove DiveListView::fixMessyQtModelBehaviour()

bstoeger:
The function DiveListView::fixMessyQtModelBehaviour() was used to
expand the first columns of dive-trips in the dive-list view.
This function was called everytime that the dive-list was modified.
It is kind of ludicrous that external callers would have to
tell the DiveListView, when it has to update its column headers.

Instead, place this functionality in the overriden reset() and
rowsInserted() functions, as these are the only ways that
rows can be added. Change the DiveTripModel to use the proper
beginResetModel()/endResetModel() pair instead of the previous
full deletion and full repopulation using the beginRemoveRows()/
endRemoveRows() and beginInsertRows()/endInsertRows().

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a small spin-off of a (quite intimidating) monster-commit which implements proper Qt model/view semantics [`beginInsertRows()`, etc. you know what I mean] for DiveTripModel.

I hope breaking out tiny things like this will make it a bit more palatable.

This was tested with random dive-list manipulations (merge trips, delete dives) and setting/removing filters. Seemed to work as expected...
 
### Changes made:

1) Replace `fixMessyQtModelBehaviour()` and replace by the overridden `reset()` and `rowsInserted()` functions.

### Related issues:

None.
### Release note:

Nope.
### Documentation change:

Nope.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -39,7 +39,6 @@ DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelec
 	DiveTripModel *tripModel = new DiveTripModel(this);
 	model->setSourceModel(tripModel);
 	setModel(model);
-	connect(model, SIGNAL(layoutChanged()), this, SLOT(fixMessyQtModelBehaviour()));
 
 	setSortingEnabled(false);
 	setContextMenuPolicy(Qt::DefaultContextMenu);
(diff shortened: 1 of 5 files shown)


dirkhh:
This looks sane to me, but since @tcanabrava wrote this code and fought with this behavior for quite a while to get it right, I'd really love to make sure he's happy with this solution...

tcanabrava:
Seems Sane.
my biggest figth with the models was actually the strange behavior that Dirk and Linus want to "Select a trip = selects the whole dives inside of a trip"
Thanks for making my code better <3

dirkhh:
That Mac build error is very strange. 45 minutes run time and claims that part of the binaries that were built on this machine were built for macOS 10.12? What the heck.
Restarting.

PR #1549, opened 2018-07-30
Cleanup: Make WindowsTitleUpdate a global object

bstoeger:
WindowsTitleUpdate is such a trivial object (a QObject with a single
signal and no own state), that it's not really understandable why
it would need all that "singleton" boiler-plate. Just make it
a default constructed/destructed global object.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

`WindowsTitleUpdate` is such a trivial object, that I don't understand why it would need all that `instance()` boiler-plate. Simply make it a global (`extern`) object.

### DIFF
--- a/core/windowtitleupdate.cpp
+++ b/core/windowtitleupdate.cpp
@@ -1,33 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "windowtitleupdate.h"
 
-WindowTitleUpdate *WindowTitleUpdate::m_instance = NULL;
-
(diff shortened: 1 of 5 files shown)


dirkhh:
I wrote this code (and yes, I did - trust me) and I think I went with the singleton because I had been burned by not doing that too many times before. I have no issue with you cleaning this up :-)

bstoeger:
Seems like Travis is not feeling too well today..

PR #1550, opened 2018-07-31
Cleanup: replace virtual by override where appropriate

bstoeger:
The keyword "virtual" signalizes that the function is virtual,
i.e. the function of the derived class is called, even if the
call is on the parent class.

It is not necessary to repeat the "virtual" keyword in derived
classes. To highlight derived virtual functions, the keyword
"override" should be used instead. It results in a hard compile-
error, if no function is overridden, thus avoiding subtle bugs.

Replace "virtual" by "override" where appropriate. Moreover,
replace Q_DECL_OVERRIDE by override, since we require reasonably
recent compilers anyway.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a purely code-hygiene thing. It *is* a bit dangerous and I hope I didn't unvirtualize a function that's supposed to be virtual, notably `TankItem::modelDataChanged()` and `DivePictureModel::updatePictures()`. But it doesn't look like these classes are inherited from?

BTW: inlining virtual functions as in
```
	void startDownload() override { }
	void startUpload() override { }
	void buttonClicked(QAbstractButton *button) override { Q_UNUSED(button) }
```
seems a bit pointless...
### Changes made:
1) Replace `virtual` by `override` where appropriate.
2) Replace `Q_DECL_OVERRIDE` by `override`.
3) Replace `/*reimp*/` by `override`.

### DIFF
--- a/core/divesitehelpers.h
+++ b/core/divesitehelpers.h
@@ -10,7 +10,7 @@ Q_OBJECT
 public:
 	static ReverseGeoLookupThread *instance();
 	void lookup(struct dive_site *ds);
-	void run() Q_DECL_OVERRIDE;
+	void run() override;
 
 private:
 	ReverseGeoLookupThread(QObject *parent = 0);
--- a/desktop-widgets/btdeviceselectiondialog.h
+++ b/desktop-widgets/btdeviceselectiondialog.h
@@ -39,8 +39,8 @@ public:
 	bool isActive() const;
 	QString errorToString() const;
 	QBluetoothDeviceDiscoveryAgent::Error error() const;
-	virtual void run();
-	virtual void stop();
+	void run() override;
+	void stop();
 
 private:
 	bool running;
--- a/desktop-widgets/configuredivecomputerdialog.h
+++ b/desktop-widgets/configuredivecomputerdialog.h
@@ -26,9 +26,9 @@ public:
 	GasSpinBoxItemDelegate(QObject *parent = 0, column_type type = PERCENT);
 	~GasSpinBoxItemDelegate();
 
-	virtual QWidget *createEditor(QWidget *parent, const QStyleOptionViewItem &option, const QModelIndex &index) const;
-	virtual void setEditorData(QWidget *editor, const QModelIndex &index) const;
-	virtual void setModelData(QWidget *editor, QAbstractItemModel *model, const QModelIndex &index) const;
+	QWidget *createEditor(QWidget *parent, const QStyleOptionViewItem &option, const QModelIndex &index) const override;
+	void setEditorData(QWidget *editor, const QModelIndex &index) const override;
+	void setModelData(QWidget *editor, QAbstractItemModel *model, const QModelIndex &index) const override;
 
 private:
 	column_type type;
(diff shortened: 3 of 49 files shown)


janiversen:
Am I the only one who gets a ton of “override” warnings? When I compile on Mac, I get a couple of pages full of these warnings, I hope this is a sideeffect that will be corrected?

janiversen:
If not, I will make a pr solving this.

janiversen:
@dirkhh wdyt ?

bstoeger:
> If not, I will make a pr solving this.

Please.
I'm not sure what is the way to go here:
1) remove all overrides
2) add missing overrides
3) silence compiler warnings
4) revert

?

In any case, I'd like to meet the person who came up with that warning.

dirkhh:
I don't know what the right answer is. Do the Travis build logs give you some insight into what's happening? Is that a warning that we should (and can) disable on Mac?

bstoeger:
The warning is - if you use `override` for one function, you should use it for all.

I have no idea if that is a sensible proposition (personally, I think it is "peculiar"). If @janiversen doesn't feel like fixing it - just revert. It doesn't matter.

dirkhh:
I don't think it's fair to make this @janiversen 's problem :-)
My guess is that we should simply disable that warning if it makes no sense - and that's really what I was asking about. Does this warning raise an actual issue, or is it one of those warnings that are just dumb... remember, you are the resident C++... no wait, @tcanabrava made himself the C++ guru, so we should ask him about that :-)

bstoeger:
My opinion: the warning is inane. But the whole matter is not important enough to fight the compiler. I can come up with a patch that removes all `override`s.

dirkhh:
I thought you said the `override` is what we are supposed to use on sane compilers.
Frankly - I'm out of my area of competence. I'm looking at you and @tcanabrava to tell me what's the right thing to do. Or maybe @thiagomacieira

thiagomacieira:
I think Clang's warnings are sane: if does not warn that you didn't use the keyword, but it does warn if you used it in some methods but not in others. The pagefuls that  @janiversen must be getting are because of the same headers included multiple times. Fixing a couple of lines should reduce this considerably.

janiversen:
I can confirm that it seems to be about 2 handfull of places. And the warning (at the places I controlled) are due to usage where the keyword is missing, not that other methods are missing the keyword.

I personally do not like the override keyword, but have no problem with it, as long as it is used consistently. And I strongly beleive disabling the warning is the wrong way to go.

bstoeger:
> I thought you said the override is what we are supposed to use on sane compilers.

It's purely a code annotation thing, which has no effect whatsoever on compilation. It prevents you from silly mistakes, where you don't override a virtual function, because you mixed up the types (e.g. missing const).

If clang finds override for *one* function in a class, it wants it for *all* overridden function in that class. There's no point in discussing the merit of this warning; personally I can't follow the thought process that led to this warning _at all_. If you want people to use `override`, then you shouldn't make it as painful as possible. If I were using clang, I would simply remove all `override`s.

thiagomacieira:
The thought process is that you should always annotate with `override`, but old code couldn't use it before the keyword existed. That means clang detects whether you opted in and then asks you to be consistent. That's all.

PR #1551, opened 2018-08-01
Fix clang override-warnings: the nuclear option

bstoeger:
Commit df156a56c08a56eb380711a507ef739d8150a71f replaced "virtual"
by "override" where appropriate. Unfortunately, this had the
unintended consequence of producing numerous clang warnings. If
clang finds a override-modified function in a class definition,
it warns for *all* overriden virtual functions without the override
modifier.

To solve this, go the easy route and remove all overrides. At least
it is consistent.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This would be the easy way of fixing the clang override warnings: simply remove them all. I'm not saying that this is the right thing to do - I'm just throwing that as an option out there.
Three other possibilities:
- Suppress the warning.
- Add the missing overrides.
- Revert df156a56c08a56eb380711a507ef739d8150a71f and go back to the previous messy - but warning-free - state.
### Changes made:

1) Remove all overrides modifiers.

### DIFF
--- a/core/divesitehelpers.h
+++ b/core/divesitehelpers.h
@@ -10,7 +10,7 @@ Q_OBJECT
 public:
 	static ReverseGeoLookupThread *instance();
 	void lookup(struct dive_site *ds);
-	void run() override;
+	void run();
 
 private:
 	ReverseGeoLookupThread(QObject *parent = 0);
--- a/core/downloadfromdcthread.h
+++ b/core/downloadfromdcthread.h
@@ -63,7 +63,7 @@ class DownloadThread : public QThread {
 
 public:
 	DownloadThread();
-	void run() override;
+	void run();
 
 	DCDeviceData *data();
 	QString error;
--- a/desktop-widgets/btdeviceselectiondialog.h
+++ b/desktop-widgets/btdeviceselectiondialog.h
@@ -39,7 +39,7 @@ public:
 	bool isActive() const;
 	QString errorToString() const;
 	QBluetoothDeviceDiscoveryAgent::Error error() const;
-	void run() override;
+	void run();
 	void stop();
 
 private:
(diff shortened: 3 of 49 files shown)


janiversen:
Warnings are gone. Considering that the keyword does not change the compiled code but are more “documentation”, it seems very much like the current discussion about having a not needed ; 

I am all for  removing override.

I looked at the old pr, and could not imidiatly see the diffrence between reverting the old patch or merging this one (I am sure I overlooked something) P

neolit123 (approved):

bstoeger:
> Considering that the keyword does not change the compiled code 

*If* it compiles, that is. `override` *does* prevent a whole class of subtle bugs, so it has some merit. The question is: does the gain outweigh having to fight the compiler-warnings? I'll leave that decision to @dirkhh and @tcanabrava.

> it seems very much like the current discussion about having a not needed ;

The same can be said about any coding-style question.

> I looked at the old pr, and could not imidiatly see the diffrence between reverting the old patch or merging this one (I am sure I overlooked something) P

No `virtual` keyword, where it's not needed...

dirkhh:
I am disappointed that we didn't go the other way, finding the over-ridden functions that were lacking the keyword, given that there is some subtle advantage to having this, but since we didn't have this before, I'm ok with this solution.

bstoeger:
> I am disappointed that we didn't go the other way

Then why did you merge?

I tried adding the `-Wsuggest-override` warning, but I fear adding *all* overrides is too painful. Can I compile under Linux with clang? Then I could just fix the few(?) inconsistent classes.

dirkhh:
Yes, of course you can compile under Linux with clang - I do that all the time, both gcc and clang builds.
The reason I merged this? Because the tons of warnings were distracting for our Mac developers.

dirkhh:
https://stackoverflow.com/questions/7031126/switching-between-gcc-and-clang-llvm-using-cmake

PR #1553, opened 2018-08-01
 add qPrefUnits and remove from SettingsObjectWrapper

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

next class in the qPref series.

This one caused and are still causing problems.

unit_system was defined as a short, but was in reality an enum, instead of adding extra macros, this is corrected in the first commit.

Many if the units are not (and was not) available to qml, due to the enums, which are used in C code as well and cannot be transferred to qPrefUnits. However the ui needs strings, so all set/get functions should accept and retrieve a string.

I need to make an array or similar to store the relatinship string-enum, and then update unitspreferences. I have however decided to make that in another PR, not to mix things.

### Mentions:

@dirkhh if you are unhappy with the ui changes being in another PR, then of course I can append the commits here.
@tcanabrava suggestions as to how we can secure strings and enums stays synchronized are welcome.

### DIFF
--- a/core/settings/qPrefUnit.cpp
+++ b/core/settings/qPrefUnit.cpp
@@ -1,0 +1,75 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "qPref.h"
+#include "qPrefPrivate.h"
+
+
+static const QString group = QStringLiteral("Units");
+
+qPrefUnits::qPrefUnits(QObject *parent) : QObject(parent)
+{
+}
+qPrefUnits *qPrefUnits::instance()
+{
+	static qPrefUnits *self = new qPrefUnits;
+	return self;
+}
+
+
+void qPrefUnits::loadSync(bool doSync)
+{
+	disk_coordinates_traditional(doSync);
+	disk_duration_units(doSync);
+	disk_length(doSync);
+	disk_pressure(doSync);
+	disk_show_units_table(doSync);
+	disk_temperature(doSync);
+	disk_unit_system(doSync);
+	disk_vertical_speed_time(doSync);
+	disk_volume(doSync);
+	disk_weight(doSync);
+}
+
+HANDLE_PREFERENCE_BOOL(Units, "/coordinates", coordinates_traditional);
+
+HANDLE_PREFERENCE_ENUM_EXT(Units, units::DURATION, "/duration_units", duration_units, units.);
+
+HANDLE_PREFERENCE_ENUM_EXT(Units, units::LENGTH, "/length", length, units.);
+
+HANDLE_PREFERENCE_ENUM_EXT(Units, units::PRESSURE, "/pressure", pressure, units.);
+
+HANDLE_PREFERENCE_BOOL_EXT(Units, "/show_units_table", show_units_table, units.);
+
+HANDLE_PREFERENCE_ENUM_EXT(Units, units::TEMPERATURE, "/temperature", temperature, units.);
+
+QString qPrefUnits::unit_system()
+{
+	return 	prefs.unit_system == METRIC ? QStringLiteral("metric") :
+			prefs.unit_system == IMPERIAL ? QStringLiteral("imperial") :
+											QStringLiteral("personalized");
+}
+void qPrefUnits::set_unit_system(const QString& value)
+{
+	short int v = value == QStringLiteral("metric") ? METRIC :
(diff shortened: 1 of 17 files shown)


desktop-widgets/preferences/preferences_units.cpp
```diff
@@ -51,14 +51,14 @@ void PreferencesUnits::syncSettings()
 	QString unitSystem[] = {"metric", "imperial", "personal"};
 	short unitValue = ui->metric->isChecked() ? METRIC : (ui->imperial->isChecked() ? IMPERIAL : PERSONALIZE);
```
bstoeger, line 52: Turn into `int`, please.

core/settings/qPrefUnit.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+QString qPrefUnits::unit_system()
+{
+	return 	prefs.unit_system == METRIC ? QStringLiteral("metric") :
+			prefs.unit_system == IMPERIAL ? QStringLiteral("imperial") :
+											QStringLiteral("personalized");
+}
+void qPrefUnits::set_unit_system(const QString& value)
+{
+	short int v = value == QStringLiteral("metric") ? METRIC :
```
bstoeger, line 52: Also `int`. Especially as a local variable - there's not even any space gain.

desktop-widgets/preferences/preferences_units.cpp
janiversen: I did not touch that, it was a short before, but I agree it should be an int. Thus PR tries only to implement qPrefUnits, and will be followed by a PR cleaning the use of units. I have added your suggestion to that branch.

core/settings/qPrefUnit.cpp
janiversen: thus is my class, but I copied the content from SettingsObjectWrapper 1-1 in order to keep transparency. I will change that as part of the prefs.units cleanup.

btw, not that it matters a lot, but I believe short is defined as 32 bit on a 64 bit system, so we save 2 bytes on the stack :-)

core/settings/qPrefUnit.h
```diff
@@ -0,0 +1,80 @@
...
+	Q_PROPERTY(units::VOLUME volume READ volume WRITE set_volume NOTIFY volume_changed);
+	Q_PROPERTY(units::WEIGHT weight READ weight WRITE set_weight NOTIFY weight_changed);
+
+public:
+	qPrefUnits(QObject *parent = NULL);
+	static qPrefUnits *instance();
+
+	// Load/Sync local settings (disk) and struct preference
+	void loadSync(bool doSync);
+	void inline load() { loadSync(false); }
```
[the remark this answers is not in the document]
janiversen: This is the way it is done in merged qPref classes, so why change it specifically for this class. Once all classes are merged, we can discuss and agree on a common cleanup, which is then done once for all classes.

janiversen:
@dirkhh it seems I am currently the only one adding test cases, and I am not sure how to interpret that. I have seen other PR’s (like the Qt_models), where it should be relative easy to make testcases.

I believe test cases make a system more resistent to sideeffects to changes (especially when bugs are added as test cases, first to show the bug, and second to verify it is gone) are but if I am the only with this pow, then lets use the time for other things.

dirkhh:
I am a huge fan of test cases. We aren't enforcing that they get added - in large part because the existing set is so small and inconsistent. But I really want more people to do so and I guess I should start poking people about it. But by all means, please continue to add them. They are important.

janiversen:
I think you are right, politely poking people in cases (where it is easy as in isolated like qPref and Qt_models) will slowly help us get better coverage which is gold when refactoring code. I am convinced if you do not push gently it will not happen (I will continue though, because it helps me).

dirkhh:
Thanks for your continued work on this.
I think the comments about inline have merit, it would be nice if you could clean this up as you work on this subsystem. But that shouldn't prevent us from merging this.
As for short vs. int... in some areas we try to be super space conscious (and even there we have mostly given up, e.g. the sample data structure has grown quite a bit over time). Here it doesn't really matter - but it does no harm, either.

PR #1554, opened 2018-08-02
Two very minor cleanups

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are two very minor cleanups of code that I stumbled upon.

### Changes made:

1) Move `sort_table` from `subsurface_startup.c` to `divelist.c`
2) Only use 4 (3+1) instead of 7 (2*3+1) chars for three-char month names. Apparently, the old code used to translate once in-place, but now it is done dynamically and the array doesn't change.

### Release note:

Nope.
### Documentation change:

Nope.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -25,6 +25,7 @@
  * void mark_divelist_changed(int changed)
  * int unsaved_changes()
  * void remove_autogen_trips()
+ * void sort_table(struct dive_table *table)
  */
 #include <unistd.h>
 #include <stdio.h>
@@ -1385,7 +1386,6 @@ int get_dive_id_closest_to(timestamp_t when)
 		return dive_table.dives[i]->id;
 }
 
-
 void clear_dive_file_data()
 {
 	while (dive_table.nr)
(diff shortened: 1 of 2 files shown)


core/divelist.c
```diff
@@ -25,6 +25,7 @@
  * void mark_divelist_changed(int changed)
  * int unsaved_changes()
  * void remove_autogen_trips()
+ * void sort_table(struct dive_table *table)
```
dirkhh, line 28: BTW: this is from the very, very earliest days of development. Only a couple of files have these lists of functions at the beginning, and I bet they are ridiculously out of date. We should just remove them.

core/subsurfacestartup.c
```diff
@@ -114,8 +114,7 @@ struct units *get_units()
 /* random helper functions, used here or elsewhere */
 const char *monthname(int mon)
 {
-	static const char month_array[12][7] = {
-		/*++GETTEXT: these are three letter months - we allow up to six code bytes*/
+	static const char month_array[12][4] = {
```
dirkhh, line 117: good catch

dirkhh (approved):
Looks great. Thanks

PR #1555, opened 2018-08-02
core: qPref, remove inline and ; for inline functions

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I would have preferred to wait with this, since it affects all the branches (one for each non-merged class) that are being updated.

Now its done, but please bear over with me if I forget to remove it in one of the next PR in this restructuring series.

### DIFF
--- a/core/settings/qPref.h
+++ b/core/settings/qPref.h
@@ -25,8 +25,8 @@ public:
 
 	// Load/Sync local settings (disk) and struct preference
 	void loadSync(bool doSync);
-	void inline load() { loadSync(false); }
-	void inline sync() { loadSync(true); }
+	void load() { loadSync(false); }
+	void sync() { loadSync(true); }
 
 public:
 	enum cloud_status {
--- a/core/settings/qPrefAnimations.h
+++ b/core/settings/qPrefAnimations.h
@@ -15,11 +15,11 @@ public:
 
 	// Load/Sync local settings (disk) and struct preference
 	void loadSync(bool doSync);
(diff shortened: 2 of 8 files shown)


janiversen:
@dirkhh I am currently trying to understand the reasoning for some strange structs, like e.g. depth_t

I can only see disadvantages by having it in a struct (you need to address the variable within), would you be against converting these to a typedef, which seems to provide the same type of security against accidential mixing types ?

dirkhh:
changing this would be a MASSIVE change to the code base for no actual benefit.
This is a very common way to ensure in C that you handle data structures correctly.
I recommend against trying to change this.

dirkhh:
in C code with a decent compiler it doesn't

bstoeger (approved):

bstoeger:
It depends on the structure layout. On my system
```
struct test2 {
    int16_t t1;
    int16_t t3;
    int32_t t2;
};
```
is 8 bytes, whereas
```
struct test1 {
    int16_t t1;
    int32_t t2;
    int16_t t3;
};
```
is 12 bytes.

But since C programmers usually are low-level programmers I reckon that they have though about this.

dirkhh:
which is why our struct entries used to be carefully ordered.
Of course, since, a few C++ programmers have made everything better by ordering them logically and then alphabetically, types be darned. And a lazy maintainer accepted those changes instead of arguing byte counts...

PR #1556, opened 2018-08-03
 qPrefDiveplanner

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

commit 1: add new macro to handle types to allow e.g. depth_t
commit 2: the qPrefUnit qml test produced a warning, I overlooked.
rest is qPrefDivePlanner

### Mentions:

@dirkhh it would be good if some of the “planners” could test the PR before it gets merged.


janiversen:
If somebody test the PR please checK:
display_deco_mode
planner_deco_mode
and
deco_mode
because that code had a bigger cleanup

janiversen:
rebase to current master.

@dirkhh just for info, I have a couple of PR´s pending, so my rebase chain is growing (but at least linear, due to my new “law” of only submitting one PR at a time).

Should this code show problems, I will solve those as highest priority (which I hope to have proven with earlier PRs).

dirkhh:
I was REALLY hoping that @atdotde would take a look at this one based on your comments above. The planner is his baby and I couldn't even begin to try to test this. That's the only reason for not merging it, TBH.

janiversen:
Something I would like too, then lets do something different, let us park this one until @atdotde gets around to test it, and I will submit another one.

The only catch is, by doing this is that this one should not be merged in parallel with other qPref PR, should I mark it or something?

atdotde:
I have to admit, I ignored all discussions around qPref* so far as I do not have much to add to that discussion. But from the planner perspective, this looks harmless. So: Thumbs up.

atdotde (approved):

janiversen:
Thanks a lot, I on the contrary tried to understand the relationship between
display_deco_mode
planner_deco_mode and
deco_mode

but if you are happy I am too.

@dirkhh if you agree, this one is ready for merge

janiversen:
Thanks for the explanation, that makes a lot of sense.

PR #1561, opened 2018-08-05
Move DC download icons used by mobile app to mobile-widgets/qml/icons/

merinsky:
Signed-off-by: Martin Měřinský <mermar@centrum.cz>

- [ x] Code cleanup

Move two icons to mobile directory where they belong.

### Changes made:
1) Two icons moved and location in qrc updated.


janiversen:
The iOS does not seem related to this PR.

I do have a reason to talk about this patch, I have for a while been working on having all icons in one shared place, allowing easier sharing, and I actually think I moved these icons in an earlier commit.

As I move through the qml (sidetracked at the moment due to qPref) my aim is to share icons as well as code where possible.

@dirkhh it would be good to hear your opinion, on which direction we should go?

merinsky:
Hi janiversen.
I didn't see your work. But I think, until your work will be merged, it is reasonable to keep mobile icons in the mobile directory, where they are now. With a few exceptions, that this patch tries to correct.

If you are talking about sharing icons. You mean sharing between what? Between OS na app. Or between mobile and desktop version?

janiversen:
just for the record this work was merged a while ago, as part of a bigger qrc cleanup. What I still have outstanding is to secure that icons used in both desktop and mobile are only once in the repo.

the DC download to which the icons relate are in core/ and not mobile-widgets/ indicating the code is shared so why not the icons.

If you look at it from a logical perspective,  both desktop and mobile can download (using the shared code), so instead of splitting the icons, it would be logical to ensure only ine set exist.

merinsky:
Right now, all mobile app icons are not in the subsurface/icons directory. Except button-download-dc-black_icon.svg and button-download-dc_icon.svg. These two icons are not used by desktop app. And no icon from subsurface/icons directory is used by mobile app. So right now, I don't see any sharing. Just two icons in the place where they do not belong.

I don't insist on moving them to mobile-widgets directory. They can be moved elsewhere.

While sharing is generally usefull, I'm not quite sure about sharing icons between mobile and desktop app. These are two completely different applications with different GUIs. Are you sure, that icons format and organization can be seamlessly shared between Linux GUI and Androis/iOs?

Code and icons are two different things. And are organized in different manner.

From my perspective, I'm not interested in mobile app, and would like to keep desktop app clean from mobile specific things.

janiversen:
Well my interest is the subsurface application in all its flavours. I strongly do not believe in Splitting things instead the current tendency is making mobile-desktop converge. E.g. on my ipad I would prefer to have the desktop app with a touch interface and not “just” the mobile. In my mind the most logical idea is to have 1 icon directory with its own qrc file.

However I just wrote my opinion, since it influences some of the current/future plans. @dirkhh is the one who needs to decide.

merinsky:
Right now, icons are not shared. Threre are multiple qrc files and multiple icons directories. There are two icons clearly (I wouldn't repeat the reasons) in the wrong place. I proposed another location and would be happy if there is better proposal.

And for the record, you didn't answer the most important question, whether you are sure, that icons format and organization can be seamlessly shared between Linux GUI and Androis/iOs? Because it isn't seamless even between Linux GUIs (Gnome vs. KDE).

Things might change in the future. But let's work with current state and facts (not what we belive or would prefer to have).

For moving two files, this is incredible lot of text ;-)

dirkhh:
I think both of you have a point. I'm not sure what this actually 'fixes', but it's true that we are quite unlikely to use an icon like this in the desktop version...

PR #1562, opened 2018-08-06
Android intent

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Listen to the Intent notification of a device we have registered our interest in being plugged in and log that that has happened.

### Related issues:

Very loosely related to the issue of not being able to download from FTDI on most Android devices

### Mentions:

@glance- I'd love to hear what you think about this.

### DIFF
--- a/android-mobile/src/org/subsurfacedivelog/mobile/SubsurfaceMobileActivity.java
+++ b/android-mobile/src/org/subsurfacedivelog/mobile/SubsurfaceMobileActivity.java
@@ -53,13 +53,7 @@ public class SubsurfaceMobileActivity extends QtActivity
 	public void onNewIntent(Intent intent) {
 		Log.i(TAG + " onNewIntent", intent.getAction());
 		UsbDevice device = (UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE);
-		Log.i(TAG + " onNewIntent device name", device.getDeviceName());
-	//	if (Build.VERSION.SDK_INT > 20) {
-	//		Log.i(TAG + " onNewIntent manufacturer name", device.getManufacturerName());
-	//		Log.i(TAG + " onNewIntent product name", device.getProductName());
-	//	}
 		Log.i(TAG + " onNewIntent toString", device.toString());
-		setDeviceString(device.toString());
 		super.onNewIntent(intent);
 		setIntent(intent);
 		// Intent will be processed, if all is initialized and Qt / QML can handle the event
@@ -88,10 +82,6 @@ public class SubsurfaceMobileActivity extends QtActivity
 		Log.i(TAG + " processIntent", intent.getAction());
 		UsbDevice device = (UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE);
 		Log.i(TAG + " processIntent device name", device.getDeviceName());
-	//	if (Build.VERSION.SDK_INT > 20) {
-	//		Log.i(TAG + " processIntent manufacturer name", device.getManufacturerName());
-	//		Log.i(TAG + " processIntent product name", device.getProductName());
-	//	}
-		Log.i(TAG + " processIntent toString", device.toString());
+		setDeviceString(device.toString());
 	} // processIntent
 }
--- a/android-mobile/src/org/subsurfacedivelog/mobile/SubsurfaceMobileActivity.java
+++ b/android-mobile/src/org/subsurfacedivelog/mobile/SubsurfaceMobileActivity.java
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: GPL-2.0
+
+// Java classes for handling Intents on Android (which in this case means
+// Subsurface-mobile receiving notifaction that an FTDI dive computer was
+// plugged in
+
+// Created while looking at https://github.com/ekke/ekkesSHAREexample
+
+package org.subsurfacedivelog.mobile;
+
+import org.qtproject.qt5.android.QtNative;
+
+import org.qtproject.qt5.android.bindings.QtActivity;
+import android.os.*;
+import android.content.*;
+import android.app.*;
+
+import java.lang.String;
+import android.content.Intent;
+import android.hardware.usb.UsbDevice;
+import android.hardware.usb.UsbManager;
+import android.util.Log;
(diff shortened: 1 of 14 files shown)


glance-:
Looks resonable to me.

Next steps are to figure out how to open the qml download page, when receiving the intent.

dirkhh:
The fun continues. Now we actually try to extract the human readable name of the device (which for my only FTDI device, a broken OSTC 3 gets me "HeinrichtWeikamp OSTC 3") and also open the download page. I ran out of time before heading on this business trip to automatically set the correct vendor / product based on the information received (and I'm guessing that there will be cables where that is garbage).
I didn't take the OSTC 3 and the ancient loaner phone where I can test this code with me, so not sure if I can make meaningful progress the next three of days.
Not sure if this is useful enough to merge as is. Thoughts?

janiversen:
Merging it makes sure it gets tested more (based on my experience with qPref), and should it do something awful wrong (which I do not believe) then the nuclear option of reverting is available :-)

dirkhh:
I think @janiversen is making a valid point. This really shouldn't disrupt other OSs and is reasonably tested on Android. I'll give @glance- a little more time in case he wants to comment, but maybe I should just merge things.

dirkhh:
This has turned into a somewhat longish series...
Based on the testing I think this is close to being ready for merging.
I'd appreciate some reviews... (and comments if you've browsed through the code)
I way too often just push my stuff into master, this is big enough that I'd rather not do that.

core/android.cpp
```diff
@@ -209,3 +239,15 @@ bool subsurface_user_is_root()
 	return false;
 }
 }
+
+/* called from QML manager */
+void checkPendingIntents()
+{
+	QAndroidJniObject activity = QtAndroid::androidActivity();
+	if(activity.isValid()) {
```
neolit123: minor nit:  missing space.

mobile-widgets/qmlmanager.cpp
```diff
@@ -1775,3 +1776,95 @@ int QMLManager::getDetectedProductIndex(const QString &currentVendorText)
...
+	// the rest is based on the vendor and product IDs
+	if (deviceString.contains("HeinrichsWeikamp OSTC 3")) {
+		name = QString("%1;%2;%3")
+				.arg(vendorList.indexOf("Heinrichs Weikamp"))
+				.arg(productList["Heinrichs Weikamp"].indexOf("OSTC 3"))
+				.arg(connectionListModel.indexOf("FTDI"));
+
+	} else if (deviceString.contains("mManufacturerName=ATOMIC AQUATICS") &&
+		   deviceString.contains("mProductName=COBALT")) {
+		if (deviceString.contains("mVersion=2"))
```
neolit123: nit: should be wrapped with `{` since multiline if/else?

android-mobile/src/org/subsurfacedivelog/mobile/SubsurfaceMobileActivity.java
```diff
@@ -0,0 +1,87 @@
...
+		} else {
+			Log.i(TAG + " checkPendingIntents", "nothingPending");
+		}
+	} // checkPendingIntents
+
+
+	private void processIntent() {
+		Intent intent = getIntent();
+		Log.i(TAG + " processIntent", intent.getAction());
+		UsbDevice device = (UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE);
```
neolit123: `null` check for device:
```
...
UsbDevice device = (UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE);
if (device != null) {
	Log.i(TAG + " processIntent device name", device.getDeviceName());
	setDeviceString(device.toString());
} else {
	Log.i(TAG + " processIntent", "device is null");
}
```

android-mobile/src/org/subsurfacedivelog/mobile/SubsurfaceMobileActivity.java
```diff
@@ -0,0 +1,87 @@
...
+			Log.i(TAG + " checkPendingIntents", "checkPendingIntents: true");
+			processIntent();
+		} else {
+			Log.i(TAG + " checkPendingIntents", "nothingPending");
+		}
+	} // checkPendingIntents
+
+
+	private void processIntent() {
+		Intent intent = getIntent();
```
neolit123, line 84: might never happen, but i would still null check:
```
if (intent == null) {
	Log.i(TAG + " processIntent", "intent is null");
	return;
}
```

android-mobile/src/org/subsurfacedivelog/mobile/SubsurfaceMobileActivity.java
```diff
@@ -0,0 +1,87 @@
...
+				isIntentPending = true;
+			}
+		}
+	} // onCreate
+
+	// if we are opened from other apps:
+	@Override
+	public void onNewIntent(Intent intent) {
+		Log.i(TAG + " onNewIntent", intent.getAction());
+		UsbDevice device = (UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE);
```
neolit123, line 55: null check for `device`:

```
UsbDevice device = (UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE);
if (device == null) {
	Log.i(TAG + " onNewIntent toString", "device is null");
	return;
}
```

neolit123 (approved):
lgtm
added some null check comments for the java file.

dirkhh:
We aren't getting any more reviews and people using the test binaries seem happy, so let's merge this and do a new Android Beta.

dirkhh:
BTW: the weird thing that things that I push get built twice by Travis is annoying, but it's part of that whole "I moved subsurface from my account to Subsurface-divelog". So when I push branches to my own repo, somehow GitHub pushes them to Subsurface-divelog instead. So this is built once as a branch and once as a PR. And if one of those two Travis jobs passes and the other fails with a stupid connection issue, it seems wasteful to restart that one... that's why I merged something that Travis shows as "build failed"

PR #1566, opened 2018-08-08
Allow zero length segments in planner

atdotde:
Those are needed to indicate bailout or set point switches
at the beginning of  the ascend.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh please check as the code deleted was originally by you (according to git blame) and I don't understand why it was there in the first place.


dirkhh:
The original algorithm needed us to treat the "magic" segment differently. If your code handles them correctly, no need to skip them.

PR #1569, opened 2018-08-12
qPrefLanguage

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/tests/testpreferences.cpp
+++ b/tests/testpreferences.cpp
@@ -109,43 +109,6 @@ void TestPreferences::testPreferences()
 	TEST(general->o2Consumption(), 1);
 	TEST(general->pscrRatio(), 1);
 	TEST(general->useDefaultFile(), false);
-
-	auto language = qPrefLanguage::instance();
-	language->set_lang_locale("en_US");
-	language->set_language("en");
-	language->set_time_format("hh:mm");
-	language->set_date_format("dd/mm/yy");
-	language->set_date_format_short("dd/mm");
-	language->set_time_format_override(false);
-	language->set_date_format_override(false);
-	language->set_use_system_language(false);
-
-	TEST(language->lang_locale(), QStringLiteral("en_US"));
-	TEST(language->language(), QStringLiteral("en"));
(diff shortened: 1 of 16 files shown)


tests/testpreferences.cpp
```diff
@@ -110,42 +110,56 @@ void TestPreferences::testPreferences()
...
+
+	TEST(location->timeThreshold(), 10);
+	TEST(location->distanceThreshold(), 20);
+
+	location->setTimeThreshold(30);
+	location->setDistanceThreshold(40);
+
+	TEST(location->timeThreshold(), 30);
+	TEST(location->distanceThreshold(), 40);
+>>>>>>> core: activate qPrefLanguage
```
dirkhh: that looks odd... is that the leftover of a merge marker?

tests/testpreferences.cpp
```diff
@@ -159,7 +159,6 @@ void TestPreferences::testPreferences()
 
 	TEST(location->timeThreshold(), 30);
 	TEST(location->distanceThreshold(), 40);
->>>>>>> core: activate qPrefLanguage
```
dirkhh: hehe. there it is again. I'd appreciate if you could refactor the commits to not add this in the first place as it will break compilation in the commits in between which could be annoying when bisecting.

dirkhh (changes requested):
Thanks for another entry in this ongoing series.
I'd appreciate if you could refactor them to never add the merge marker in the first place. Simply so we don't stumble into this when bisecting.
Otherwise this looks good.
Thanks

janiversen:
Done, I accidently pushed when I wanted something else (I still test that it builds with every commit)

dirkhh:
For once this looks like actual test failures instead of Travis errors.
I'll admit that with all the random pointless errors lately, it's easy to miss looking into each of them in detail...

janiversen:
it is real errors, short - int, once again, will solve it tomorrow

dirkhh (approved):

PR #1571, opened 2018-08-14
Move configure dive computer to be called from Download dialog

dirkhh:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This was a suggestion from Linus when quite a few users ended up trying to configure dive computers instead of downloading from them.
This way the entry is gone from the main menu, and the button is grayed out unless a dive computer that we actually CAN configure is selected.

### Release note:

I guess I need to add a CHANGELOG.md entry. Oops :-)

### Documentation change:

This requires changes to the user manual. We need to update the screenshots for the download dialog and change the text to explain that that's from where you can start the configuration.

### Mentions:

@glance- do I list all the supported dive computers? That's now important to get right...

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.cpp
+++ b/desktop-widgets/configuredivecomputerdialog.cpp
@@ -14,21 +14,41 @@
 #include <QProgressDialog>
 #include <QSettings>
 
+static int getTypeFromVendorProduct(QString vendor, QString product)
+{
+	if (vendor == "Heinrichs Weikamp") {
+		if (product == "OSTC Mk2" ||
+		    product == "OSTC 2N" ||
+		    product == "OSTC 2C") {
+			return 0;
+		}
+		if (product == "OSTC 2" ||
+		    product == "OSTC 3" ||
+		    product == "OSTC Sport" ||
+		    product == "OSTC Plus" ||
+		    product == "OSTC cR") {
+			return 1;
+		}
+		if (product == "OSTC 4") {
+			return 2;
+		}
+		return -1;
+	}
+	if (vendor == "Suunto" &&
+	    (product == "Vyper" ||
+	     product == "Stinger" ||
+	     product == "Mosquito" ||
+	     product == "D3" ||
+	     product == "Vytec" ||
+	     product == "Cobra" ||
+	     product == "Gekko" ||
+	     product == "Zoop"))
+		return 3;
+	return -1;
+}
 bool diveComputerConfigurable(QString vendor, QString product)
 {
-	if ((vendor == "Heinrichs Weikamp" &&
-	     product != "OSTC") ||
-	    (vendor == "Suunto" &&
-	     (product == "Vyper" ||
-	      product == "Stinger" ||
-	      product == "Mosquito" ||
-	      product == "D3" ||
-	      product = "Vytec" ||
-	      product == "Cobra" ||
-	      product == "Gekko" ||
-	      product == "Zoop")
-		return true;
-	return false;
+	return getTypeFromVendorProduct(vendor, product) >= 0;
 }
 
 GasSpinBoxItemDelegate::GasSpinBoxItemDelegate(QObject *parent, column_type type) : QStyledItemDelegate(parent), type(type)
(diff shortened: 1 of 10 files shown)


desktop-widgets/configuredivecomputerdialog.cpp
```diff
@@ -14,9 +14,27 @@
 #include <QProgressDialog>
 #include <QSettings>
 
+bool diveComputerConfigurable(QString vendor, QString product)
+{
+	if ((vendor == "Heinrichs Weikamp" &&
+	     product != "OSTC") ||
```
glance-: As far as I've know, the original "OSTC" runs the same firmware as the mk2, Thus it works just fine to configure it.

desktop-widgets/configuredivecomputerdialog.cpp
```diff
@@ -14,21 +14,41 @@
...
+	     product == "Stinger" ||
+	     product == "Mosquito" ||
+	     product == "D3" ||
+	     product == "Vytec" ||
+	     product == "Cobra" ||
+	     product == "Gekko" ||
+	     product == "Zoop"))
+		return 3;
+	return -1;
+}
```
glance-, line 48: Rather than hard coding a bunch of names, whitelist DC_FAMILY_HW_OSTC, DC_FAMILY_HW_OSTC3, and DC_FAMILY_SUUNTO_VYPER.

For odd reasons, the OSTC 4 is in the DC_FAMILY_HW_OSTC3 family, but for this perspective, it works differently.

glance- (commented):
I'm not shure if this UI makes more sense, and I feel like its way more complicated to reach.

Also it feels really odd to to First select vendor, product and maybee device and then click configure and then need to enter device agan, and click connect.

Yea, I know, A lot of folks don't care about the config pages, but I use them, and this feels backwards to me.

It also breaks guides like http://forum.heinrichsweikamp.com/read.php?6,14509 which quite a few folks use to upgrade the firmware in their devices.

dirkhh:
>I'm not shure if this UI makes more sense, and I feel like its way more complicated to reach.

Yeah, I guess that "harder to reach" part is intentional. We've had quite a few people who end up in the configure screen by mistake and send support requests (or give up) because it looks like their dive computer isn't supported. Frankly, the current one is too easy to reach :-(

> Also it feels really odd to to First select vendor, product and maybee device and then click configure and then need to enter device agan, and click connect.

I agree that this could be better. I am pre-selecting the right dive computer family already, I should also pass in the device if selected, I guess. Actually, now that I think about it, these are ALL computers that are connected over FTDI or BT. So maybe there's more opportunity for simplification.
Or maybe this is just a dumb idea in the first place. Linus suggested it and it seemed to make a lot more sense to me... We cannot disable the main menu entry for this without knowing which dive computer a user has... yet we really don't want to show this to someone unless they have one of the supported dive computers it seems.

> Yea, I know, A lot of folks don't care about the config pages, but I use them, and this feels backwards to me.

Understood.

> It also breaks guides like http://forum.heinrichsweikamp.com/read.php?6,14509 which quite a few folks use to upgrade the firmware in their devices.

Admittedly, the changes to that guide would be fairly small.

But yeah, maybe this is just the wrong idea and all we should do is make it way more obvious that this isn't the dialog people are looking for. Which in itself will be ugly and weird and break the guides.

I don't know.

desktop-widgets/configuredivecomputerdialog.cpp
dirkhh: At the point when I make this determination, I have the two strings.
But yeah, I guess I should parse that into a family and use that instead.
Or just not bother. I don't use this. Ever. Why am I fixing it?

glance-:
I was floating the idea in my mind just to throw a popup, the first time a user opens the configure dialog explaining that its is to change settings and upgrade firmware in some dive computers, and its not needed for downloading data from a dive computer.

It can be done quite easily and when we have shown it once, we just write a QSettings thingie that the user has bin warned.

Not exelent UX, but probably a bit simpler.

PR #1573, opened 2018-08-14
qPrefGeneral

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR completes the series converting SettingsObjectWrapper to qPref*

There are a couple of PR related in the pipeline:
1) remove SettingsObjectWrapper totally
2) move all QSettings usage to qPref (there are a couple of conflicting usages)
3) update qml, qmlprefs and qmlmanager to use qPref


dirkhh:
Yay. This was a lot of work.
I don't see anything glaringly wrong in the code, so let's wait for Travis.

Completely unrelated to this PR, I noticed that on my Mac when testing yesterday I kept getting the question about "checking for updates" on several starts in a row...
I was in the middle of testing my own code and didn't investigate and then it stopped. I know, great way to report an issue :-(

janiversen:
I did notice somethung similar the other day but took it as “a glitch”, since you have seen the same I will investigate a bit deeper.

Yes a lot of work, but I strongly believe this part of the code is now more consistent and stable.

dirkhh:
> Yes a lot of work, but I strongly believe this part of the code is now more consistent and stable.

I think the consistency and the accessibility from QML are great. I hope this was the last rewrite of that particular code... and I look forward to the changes that you and others are now able to contribute to the mobile version!

janiversen:
thanks. btw curent master does not show “checking for updates”. can I have disbled it (cannot see that option in preferences) ?

dirkhh:
IIRC the logic is based on whether we have an entry for the next time to check. So is it possible that the dialog is displayed, but then the next time to check isn't set? I have to run to a meeting, so I can't quote the sources right now...

janiversen:
the magic is in desktop-widgets/updatemanager.cpp

I forced next_check back in history, restarted ssrf and got a dialog
“you are running a development version...”
next_check was correctly updated to noeˋ14days
restarted ssrf
no dialogbox

I am not sure why your mac are different (could it be something with the date format?
could you please do a qDebug() << update_manager->next_check() <<  “   “ < QDate::currentDate()
before and after the set_next_check call in updatemanager.cpp

dirkhh:
Interesting thing to ask for because when I did the tests I was running in a German locale as I was testing an annoying glitch in the translations as well (which turned out to be user error, btw).
And of course then once you look it's obvious why this was broken. The dates are stored as localized strings instead of storing them as time_t value. And if you switch locales, comparing those strings will go wrong.

janiversen:
if ok with you I will change that, but I am not sure I can keep compatibility (also not sure it really matters).

PR #1575, opened 2018-08-14
Cleanup: Split out filter-functionality

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is in preparation for making the undo-changes compatible with the filter. It splits out the filter functionality in a function that takes a `dive *`. Since this is independent of the undo-PR I post this as a separate PR.

There *is* one functional change: if divesite-filtering is enabled, this will now also be reflected in the `hidden_by_filter` flag. I'm not sure if the old behavior was an oversight or on purpose. There was no comment.

### Changes made:

1) Remove unused parameters
2) Constify parameters
3) Split out dive-filter function

### Related issues:

#1528 
### Release note:

No user visible changes (hopefully).
### Documentation change:

No user visible changes (hopefully).

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -4202,9 +4202,9 @@ const char *get_dive_country(struct dive *dive)
 	return NULL;
 }
 
-char *get_dive_location(struct dive *dive)
+const char *get_dive_location(const struct dive *dive)
 {
-	struct dive_site *ds = get_dive_site_by_uuid(dive->dive_site_uuid);
+	const struct dive_site *ds = get_dive_site_by_uuid(dive->dive_site_uuid);
 	if (ds && ds->name)
 		return ds->name;
 	return NULL;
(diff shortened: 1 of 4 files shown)


dirkhh:
This looks sane to me. @tcanabrava  any comments?

bstoeger:
Added a somewhat more brutal commit [at airport - no time for substantial things].

This separates core and UI logic: Set `hidden_by_filter` in one place and let the UI only access this flag. The old code was very hard to reason about as the `hidden_by_filter` flag was set somewhere deep in the UI code. Also to me it very much looks like every selection function was called twice once for filtering trips and then dives..?

dirkhh:
OK, giving up on waiting for @tcanabrava :-(

PR #1577, opened 2018-08-14
Automate ftdi

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Clearly, picking the right device is hard. Also, FTDI drivers on the Mac appear to be a mess.
This solves both problems for people with FTDI dive computers (still the majority of them).
Build on Mac against libftdi1, try to guess if the selected dive computer uses FTDI, and then disable the device drop down and just use ftdi, period
The question becomes "what if we guess wrong?" - is it possible that there's a 3rd party non-FTDI cable for one of the typically FTDI computers? Is there a Silicon Labs cable for a Suunto Vyper?
If that was a real possibility, we might need a way to override this. But for now I assume that this would work.

### Release note:

Automatically detect FTDI based dive computers on Mac

### Documentation change:

Yes, this should be mentioned in the user manual, but this should be very easy, something like "for most dive computers Subsurface automatically detects how to connect to the device, for those where it doesn't..." and then lead to the text that describes how you figure out what device name to use.

### Mentions:

@glance- because he always is good at telling me why my ideas are stupid
@atdotde because hopefully you can test this 
@neolit123 because I wonder if this would work under Windows as well... I haven't tried to build against libftdi1 on Windows, but if that was possible, we should definitely consider it.

### DIFF
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -2,6 +2,7 @@
 #include "core/libdivecomputer.h"
 #include "core/qthelper.h"
 #include "core/subsurface-qt/SettingsObjectWrapper.h"
+#include "core/subsurface-string.h"
 #include <QDebug>
 #include <QRegularExpression>
 #if defined(Q_OS_ANDROID)
@@ -62,6 +63,27 @@ void DownloadThread::run()
 	dcs->set_product(internalData->product);
 	dcs->set_device(internalData->devname);
 	dcs->set_device_name(m_data->devBluetoothName());
+}
+
+static const QVector<dc_family_t> ftdiFamilies = {
+	DC_FAMILY_SUUNTO_SOLUTION, DC_FAMILY_SUUNTO_EON, DC_FAMILY_SUUNTO_VYPER, DC_FAMILY_SUUNTO_VYPER2, DC_FAMILY_SUUNTO_D9,
+	DC_FAMILY_OCEANIC_VTPRO, DC_FAMILY_OCEANIC_VEO250, DC_FAMILY_OCEANIC_ATOM2,
+	DC_FAMILY_HW_OSTC, DC_FAMILY_HW_OSTC3, // but careful, the OSTC 4 is DC_FAMILY_HW_OSTC3 but has no serial connector
+	DC_FAMILY_CRESSI_EDY, DC_FAMILY_CRESSI_LEONARDO,
+	DC_FAMILY_DIVERITE_NITEKQ,
+	DC_FAMILY_DIVESYSTEM_IDIVE,
+	DC_FAMILY_TECDIVING_DIVECOMPUTEREU};
+
+bool guessIfFTDI(dc_descriptor_t *descriptor)
+{
+	if (!descriptor)
+		return false;
+	dc_family_t family = dc_descriptor_get_type(descriptor);
+	if (family == DC_FAMILY_HW_OSTC3 && same_string(dc_descriptor_get_product(descriptor), "OSTC 4"))
+		return false;
+	if (ftdiFamilies.contains(family))
+		return true;
+	return false;
 }
 
 static void fill_supported_mobile_list()
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -75,8 +75,10 @@ DownloadFromDCWidget::DownloadFromDCWidget(QWidget *parent, Qt::WindowFlags f) :
 	if (!dc->vendor().isEmpty()) {
 		ui.vendor->setCurrentIndex(ui.vendor->findText(dc->vendor()));
 		productModel.setStringList(productList[dc->vendor()]);
-		if (!dc->product().isEmpty())
+		if (!dc->product().isEmpty()) {
 			ui.product->setCurrentIndex(ui.product->findText(dc->product()));
+			updateDeviceEnabled();
+		}
 	}
 
 	updateState(INITIAL);
@@ -479,6 +481,15 @@ void DownloadFromDCWidget::updateDeviceEnabled()
 
 	// call dc_descriptor_get_transport to see if the dc_transport_t is DC_TRANSPORT_SERIAL
 	if (dc_descriptor_get_transports(descriptor) & DC_TRANSPORT_SERIAL) {
+#if defined(SERIAL_FTDI)
+		if (guessIfFTDI(descriptor)) {
+			// this should be an FTDI device that we'll connect to by using
+			// the magic device name "ftdi"
+			ui.device->setEditText("ftdi");
+			ui.device->setEnabled(false);
+			return;
+		}
+#endif
 		// if the dc_transport_t is DC_TRANSPORT_SERIAL, then enable the device node box.
 		ui.device->setEnabled(true);
 	} else {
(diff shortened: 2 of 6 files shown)


dirkhh:
Just realized that Travis wasn't building against libftdi on Mac which kind-of defeated the purpose of the testing here :-)

core/downloadfromdcthread.cpp
```diff
@@ -64,6 +65,27 @@ void DownloadThread::run()
 	dcs->set_device_name(m_data->devBluetoothName());
 }
 
+QList<dc_family_t> ftdiFamilies = {
```
bstoeger: Unless this is used in other translation units, you can make it `static const`.

Moreover, `QVector` is usually preferred over `QList`. The latter has some very strange semantics (among other things: what it does depends on the size of the type).

dirkhh: right now it's only used here - so yeah, `static const` should be fine.
And QVector vs QList makes no difference to me - all I ever do is use the `.contains()` member as that creates much more readable code than a series of `if(....) ` conditions.

Thanks for the suggestions!

neolit123:
> @neolit123 because I wonder if this would work under Windows as well... I haven't tried to build against libftdi1 on Windows, but if that was possible, we should definitely consider it.

i just built it using cmake. it worked fine:
```
C:\dev\libftdi\build\src>dir /b *.dll *.a
libftdi1.dll
libftdi1.a
libftdi1.dll.a
```
it needed libusb from pkg-config:
```
-- Checking for module 'libusb-1.0'
--   Found libusb-1.0, version
-- Found LIBUSB: /lib/libusb/lib/libusb-1.0.a
```

edit: built from this "mirror":
https://github.com/mdarse/libftdi/
not sure it has an official github mirror?

https://github.com/mdarse/libftdi/blob/master/README.mingw
but the links are out of date there, here is updated info:
https://github.com/libusb/libusb/wiki/Windows#how-to-use-libusb-on-windows

> Recommended: Use the most recent version of Zadig, an Automated Driver Installer GUI application for WinUSB, libusb-win32 and libusbK...

this means that the user has to install the WinUSB driver manually for libusb and libftdi to work?

but libdc already needs libusb on Windows and i don't think we have the WinUSB driver requirement.
does libusb + libdc work for users ATM without it?

dirkhh:
Yes, we already use libusb for the Cobalt. I need to double check but I think we no longer use libusb for the EON Steel/Core.
Thanks for taking a poke at this. I'm updating my Windows build environment right now and things got messed up - that's why I asked you to take a look in the first place.
I don't think I'll get this all sorted before I head on vacation, but it's pretty high on top of my list as right now I cannot create a new Windows release and that sucks :-/

The other option would be to try this on Travis, I guess.

dirkhh:
Oh, which OSTC do you have, Lubomir? Is it a BT only one, or does it have a serial connector on the back as well? Because all the serial enabled OSTC are in fact FTDI devices :-)

neolit123:
> It would be a round silver screw (maybe 12mm diameter - designed to use with a coin or something) in the center of the back...

the back looks like this (charger attached):
https://www.innodive.com/store/upload/010217_124158_PEEL_ABcbgn.jpg

on the side there is a silver screw, but the battery goes in there:
https://www.innodive.com/store/upload/010217_124158_PEEL_UfCkmz.jpg

edit: maybe that's the serial version:
image

dirkhh:
@glance- @atdotde any comments on this one?

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -479,6 +481,15 @@ void DownloadFromDCWidget::updateDeviceEnabled()
 
 	// call dc_descriptor_get_transport to see if the dc_transport_t is DC_TRANSPORT_SERIAL
 	if (dc_descriptor_get_transports(descriptor) & DC_TRANSPORT_SERIAL) {
+#if defined(SERIAL_FTDI)
+		if (guessIfFTDI(descriptor)) {
+			// this should be an FTDI device that we'll connect to by using
+			// the magic device name "ftdi"
+			ui.device->setEditText("ftdi");
+			ui.device->setEnabled(false);
```
glance-, line 489: I think this is the wrong way of doing it.

Ex, a lot of the DC_FAMILY_HW_OSTC3 devices are bluetooth, and not ftdi.

It would be better I guess to leave the field editable, so the user has the oppertunity to choose something else.

core/downloadfromdcthread.cpp
```diff
@@ -64,6 +65,27 @@ void DownloadThread::run()
...
+	DC_FAMILY_DIVERITE_NITEKQ,
+	DC_FAMILY_DIVESYSTEM_IDIVE,
+	DC_FAMILY_TECDIVING_DIVECOMPUTEREU};
+
+bool guessIfFTDI(dc_descriptor_t *descriptor)
+{
+	if (!descriptor)
+		return false;
+	dc_family_t family = dc_descriptor_get_type(descriptor);
+	if (family == DC_FAMILY_HW_OSTC3 && same_string(dc_descriptor_get_product(descriptor), "OSTC 4"))
```
glance-, line 82: There are other devices in the OSTC3 family which are bluetooth.

Maybee you should filter out those with DC_TRANSPORT_BLUETOOTH or DC_TRANSPORT_BLE?

glance- (commented):
I see where you're going with this, and I generally like the idea.

I'd go with something like ftdi_usb_find_all to try to find ftdi devices and go from there.

I think it would be even better if we altered our download flow to probe after what we can find and identify and give the user the shortcut to download from that dive computer. More like the download experience you created for the mobile app.

desktop-widgets/downloadfromdivecomputer.cpp
dirkhh: look above in line 484 - we only do this if the descriptor includes DC_TRANSPORT_SERIAL
And for dual mode dive computers, you still can click on Bluetooth and that overrides anything we do here, anyway.
So I _**think**_ this should work as intended.

core/downloadfromdcthread.cpp
dirkhh: I guess the function name could change... this only gets called for descriptors that we know include serial transport...
But that reminds me, doesn't @jefdriesen consider BT/rfcomm a serial transport? So this may not be sufficient after all now that I think about it...

jefdriesen:
> The question becomes "what if we guess wrong?" - is it possible that there's a 3rd party non-FTDI cable for one of the typically FTDI computers? Is there a Silicon Labs cable for a Suunto Vyper?

Yes, there is. The CustomIdea cable uses a Silicon Labs chip.

core/downloadfromdcthread.cpp
jefdriesen: Dive computers supporting rfcomm will indeed report both DC_TRANSPORT_BLUETOOTH and DC_TRANSPORT_SERIAL. The serial transport is there to indicate support for the serial port emulation mode. There are still many users of that feature, especially on Windows. In some cases it's even the only option. On Windows there exist several alternative bluetooth stacks (e.g. Widcomm and Toshiba) besides the Microsoft stack. But the Windows rfcomm api supports only the Microsoft stack. So if a user happens to have one of the alternative stacks installed, then the serial port emulation mode is the only one that will work.

PR #1578, opened 2018-08-15
clenup SettingsObjectWrapper

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

main purpose is to remove SettingsObjectWrapper files

### Changes made:

some commits do more special things:
1) testpreferences are now empty so removed

2,3,4) add static to qPref methods allowing faster and simpler call
qPrefXYZ::foo() instead of qPrefXYZ::instance()->foo()

5) only include qPref.h where needed, include qPrefXYZ instead

6) move load/sync to qPref

7) add sync() call when app terminates to ensure update of struct preferences

rest is remove SettingsObjectWrapper and use the new qPref syntax.

### Additional information:

Next up is to remove QSettings outside qPref at leadt 2 files save the same variables as in qPref.


dirkhh:
Quite the crowning achievement PR. I didn't see anything glaringly wrong.
I'll merge it so it gets testing... I'm leaving on a vacation trip tomorrow morning, so if I don't merge this today, it might not happen for a few days.

janiversen:
Hve some nice vacation time!!

PR #1580, opened 2018-08-15
divelogexportlogic: show warning on missing 'theme' folder

neolit123:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

show a warning if the user is trying to export HTML, but the "theme" folder is missing for some reason.
not really user facing, unless someone messes with his installation.

### DIFF
--- a/core/divelogexportlogic.cpp
+++ b/core/divelogexportlogic.cpp
@@ -148,8 +148,10 @@ void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes)
 	export_HTML(qPrintable(json_dive_data), qPrintable(photosDirectory), hes.selectedOnly, hes.listOnly);
 
 	QString searchPath = getSubsurfaceDataPath("theme");
-	if (searchPath.isEmpty())
+	if (searchPath.isEmpty()) {
+		report_error(qPrintable(gettextFromC::tr("Cannot find a folder called 'theme' in the standard locations")));
 		return;
+	}
 
 	searchPath += QDir::separator();
 


core/divelogexportlogic.cpp
```diff
@@ -148,8 +148,10 @@ void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes)
 	export_HTML(qPrintable(json_dive_data), qPrintable(photosDirectory), hes.selectedOnly, hes.listOnly);
 
 	QString searchPath = getSubsurfaceDataPath("theme");
-	if (searchPath.isEmpty())
+	if (searchPath.isEmpty()) {
+		report_error(gettextFromC::tr("Cannot find a folder called 'theme' in the standard locations").toUtf8());
```
dirkhh: I thought ::tr always returns utf8... so I'm not sure why you need the toUtf8() call...

core/divelogexportlogic.cpp
```diff
@@ -148,8 +148,10 @@ void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes)
 	export_HTML(qPrintable(json_dive_data), qPrintable(photosDirectory), hes.selectedOnly, hes.listOnly);
 
 	QString searchPath = getSubsurfaceDataPath("theme");
-	if (searchPath.isEmpty())
+	if (searchPath.isEmpty()) {
+		report_error(qPrintable(gettextFromC::tr("Cannot find a folder called 'theme' in the standard locations")));
```
neolit123: switched to `qPrintable`.
`.toUtf8()` converted to `QByteArray` which was then auto-casted to `const char*` which is what `report_error()` needed.

core/divelogexportlogic.cpp
neolit123: switched to `qPrintable`.
`.toUtf8()` converted to `QByteArray` which was then auto-casted to `const char*` which is what `report_error()` needs.

dirkhh:
that looks better... I guess we should let Travis run, though I don't see what could be wrong here.

core/divelogexportlogic.cpp
bstoeger, line 152: If it is not user-facing, is it worth adding a translation string?

neolit123: it can be.

bstoeger: If you think a new translation string  is warranted, I think we can commit this (after all @dirkhh gave his approval).

PR #1582, opened 2018-08-16
Cleanup: const-ify numerous functions

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a split-off of a too-large commit in #1528. The ultimate goal is to make `merge_dive()` not modify the source dives, which is realized by taking `const` pointers. This is a part of the necessary `const`-ification. For `gasmix` I chose to pass by value instead of `const`-pointer. The old passing around pointers to mutable data seemed scary and `gasmix` is only two `int`s, which nicely fit into a 64-bit register.

### Changes made:

1) Make a number of functions `const`-clean.

### Related issues:

#1528

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/deco.c
+++ b/core/deco.c
@@ -479,7 +479,7 @@ void calc_crushing_pressure(struct deco_state *ds, double pressure)
 }
 
 /* add period_in_seconds at the given pressure and gas to the deco calculation */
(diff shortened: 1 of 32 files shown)


dirkhh (approved):
Thanks for breaking this out. This is a good way to make the code cleaner.
At first I was chafing at the gasmix-by-value, but it's hard to argue that this is /wrong/. So yeah, approved.

PR #1583, opened 2018-08-17
 map-widget: add support for filtering of map locations

### LINKED ISSUE #1581: Display divesites after using filter tool
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Running subsurface version 4.8.1-249 on Ubuntu 18.04
After filtering a dive list, only the dives fitting the key are listed in the dive list, as it should be.
However, the divesite map is not updated to agree with the (now shortened) dive list that fit the key of the filter. This actually appears to work on some versions of Subsurface, e.g. apparently in Windows, though I could not verify this.
When I open the attached divelog (testMap.xml) with 6 dives and filter it for "Drysuit", then only the most north-easterly location should show. But all four location remain on the map (see attached testMap.jpg).

### Operating system:

Ubuntu 18.04

Dunno

### Subsurface version:
 V4.8.1-249
 Yes, 4.8.1
 Self-compiled

### Steps to reproduce:

1) open attached dive log.
2) Go to Log->Filter divelist and select the two keys with "Drysuit"
    Only dives from one divesite should be visible in the dive list.
3) Inspect any changes in the map content.

### Current behavior:

As explained above
### Expected behavior:

The dive map should only show one divesite and not four.

### Mentions:

@neolit123

testmap
[testMap.zip](https://github.com/Subsurface-divelog/subsurface/files/2294733/testMap.zip)

neolit123:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

If the dive list is filtered the map should hide dive locations
that do not match the dive list filter.

To achieve that, loop over dives instead of dive sites in
MapWidgetHelper::reloadMapLocations() and discard
dives that are hidden by the filter. Then check if a dive
has a dive site attached to it and re-use the old functionality.

### Changes made:

see above.

### Related issues:

Fixes https://github.com/Subsurface-divelog/subsurface/issues/1581

### Additional information:

none

### Release note:

```
Desktop/Map-Widget: add support for filtering of map locations [#1581]
```

### Documentation change:

doesn't really need it. the change makes is that the users should get the expected map + filtering behavior.

### Mentions:

@willemferguson @dirkhh @bstoeger

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Desktop/Map-Widget: add support for filtering of map locations [#1581]
 - Android: switch to Download page when dive computer is plugged in and try to
   populate vendor / device / connection based on the information we get
 - Mobile: enable multiple cylinder editing for dives that uses more than one cylinder 
--- a/map-widget/qmlmapwidgethelper.cpp
+++ b/map-widget/qmlmapwidgethelper.cpp
@@ -116,12 +116,13 @@ void MapWidgetHelper::centerOnDiveSite(struct dive_site *ds)
 
 void MapWidgetHelper::reloadMapLocations()
 {
-	struct dive_site *ds;
 	int idx;
+	struct dive *dive;
 	QMap<QString, MapLocation *> locationNameMap;
 	m_mapLocationModel->clear();
 	MapLocation *location;
 	QVector<MapLocation *> locationList;
+	QVector<uint32_t> locationUuids;
 	qreal latitude, longitude;
 
 	if (displayed_dive_site.uuid && dive_site_has_gps_location(&displayed_dive_site)) {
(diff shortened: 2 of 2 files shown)


bstoeger:
Just a general comment, which is semi-relevant to this PR:

The location-duplicate-check is weird: If there are three divesites with the same name two of which are close, it will depend on the order in which they are looped over.

With the new code you can then have the following funny situation: dive sites 1 and 2 have the same name, but are distanced by more then `MIN_DISTANCE_BETWEEN_DIVE_SITES_M`.
We loop over dives:
 - dive 1, divesite 1: divesite 1 is registered in `locationNameMap`.
 - dive 2, divesite 2: divesite 2 is registered in `locationNameMap`.
 - dive 3, divesite 1: divesite 1 is considered as new (because divesite 2 is in the `locationNameMap`) and readded.

Thus, divesite 1 is now twice on the map.

bstoeger:
As a quick fix, perhaps also collect the UUIDs in a `QMap` (or, probably more performant, a `QVector`)?

bstoeger (approved):
lgtm. Performance considerations can be addressed later.
@willemferguson: does this work for you?

willemferguson:
@neolit123 @bstoeger @dirkhh 
I tested this quite a bit and it appears pretty good. I tested this on all the filter key fields, both with single selections and multiple selections, changing the order of multiple selections. Good stuff!

PR #1592, opened 2018-08-20
 remove use of QSettings outside qPref

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

There are several places where identical values are read/written to disk in competition with qPref.

this PR removes the need for QSettings outside qPref by extending the qPref classes.

It has a number of commits to ensure each commit is a logical unit that can be tested and compiled on its own.

This is the natural (and final) addition to qPref.


janiversen:
remark the PR is not complete, but I wanted to see what travis thinks

janiversen:
@dirkhh I still miss a few complicated ones, but I prefer to make them in a separate PR, to secure they get a bit more attention.

core/qthelper.cpp
```diff
@@ -1455,19 +1454,12 @@ void init_proxy()
 
 QString getUUID()
 {
-	// This is a correct usage of QSettings,
-	// it's not a setting per se - the user cannot change it
-	// and thus, don't need to be on the prefs structure
-	// and this is the *only* point of access from it,
```
bstoeger, line 1461: I think this comment has a point. This is not a "preferences" field and I is no reason for all the `Q_PROPERTY` bloat. If you want to, for some reason, generalize non-preferences settings (current directories, etc), I would suggest doing it outside of the preferences framework.

janiversen: I can relate to not bloating Q_PROPERTY that isnot needed in this case, however I believe having QSettings not spread around the source is a good thing.

I will remove the Q_PROPERTY for this variable

janiversen:
rebased on newest master, wonde why github did not inform me about the merge problem.

janiversen:
Seems this PR is not going anywhere, even though comments have been next to nothing, which indicates that the PR is only important to me.

It is however hugely disturbing to have this open in my continued work (constant rebasing and problems with nearly every update in my tree). For these reasons I am pulling it back to my local repo
(which allows me to eg. fix the current signal issue with out rebase problems.

dirkhh:
no no, wrong reaction, Jan.
I think you know that I was gone for a week - and as is usual, coming back the backlog at work was crushing.
I tried to quickly deal with the "easy to look at" PRs, but this one simply seemed intimidating based on the number of commits. This is the right thing to do, I just want to find the time to review it thoroughly which hopefully I'll have later today.
Please cut me some slack here :-)

core/qthelper.cpp
bstoeger: I still think that the original comment is correct and this change is conceptually bad. This was perfectly localized logic, which is now spread out over multiple places with strange stuff like
```
	if (!doSync) {
		load_uuidString();
	}
```
I fail to see what we gain from that and why this has to be inside a class. Moreover suddenly, we have an interface (complexity) where there wasn't any.

If, for some reason, we don't want plain `QSettings` in code, why not use trivial get/set functions such as:
```
template <typename T>
T getSettingsValue(const char *group, const char *key, const T &def = T())
{
	QSettings settings;
	settings.beginGroup(group);
	return settings.contains(key) ? settings.value(key).value<T>() : def;
}

template <typename T>
void setSettingsValue(const char *group, const char *key, const T &val)
{
	QSettings settings;
	settings.beginGroup(group);
	settings.setValue(key, QVariant(val));
}
```
and then
```
{
	QString uuidString = getSettingsValue<QString>("UpdateManager", "UUID");
	if (uuidString.isEmpty()) {
		uuidString = QUuid::createUuid().toString();
		setSettingsValue("UpdateManager", "UUID", uuidString);
	}
 	uuidString.replace("{", "").replace("}", "");
 	return uuidString;
}
```
(completely untested).
This can of course be made much more fancy. The strings can be made constants to avoid typos, etc.

This seems so much less complex and so much more reusable for all the paths, etc.

janiversen: you are really taking this out of concept! while you might have a point in this particular case, did you look at the other commiys where QSettings was misused to duplicate what happened in the old SettingsObjectWrapper (Cloud_storage_password is an example, divePlanner is another one where a lot of variables where saved to disk, but not the direct struct preference changes). Many of my changes solves potential or undetected bugs, I believe that is a good reason to sanitize the usage.

Or look at the other things I have simplyfied. I strongly believe that limiting QSettings to one place makes the code more stable! your template actually solves nothing, it is the usage I am looking at.

However I leave it to @dirkhh

janiversen:
@dirkhh I know you had a hopefully nice vacation (although a day short; you did not fancy doing some surfing 👍 )

As I suspected, not everybody thinks it is the right thing to do, and as always there are no clear answer (both arguments have merit), but I do not have the energy for a new word exchange, so there will only be descalation.

ps. I am edging closer to understand the signal problem, which are not only related to qPref.

dirkhh:
Glad to hear that you are figuring out the signal problem. I'm at the horse barn now, which means 120 minutes of SLOW internet but fairly undisturbed time to review PRs :-)

core/qthelper.cpp
dirkhh: If I understand @bstoeger correctly, he points out that we are mixing two things in the 'preferences' - on the one hand these are indeed what one might describe as preferences, user visible settings. And then it's used as persistent storage for certain values (like the UUID or the last directory opened). Conceptually those two uses are indeed different. And I think one could argue either way - that they don't need all the overhead, or that it's better to have one set of well tested, stable code that is used for all circumstances.
I've been looking at this, trying to see both sides, and find myself fairly close to the center :-)
I think what pushes me just a tiny bit into @janiversen 's camp is the thought that if we have such a persistently stored information that we'd want to access from QML, then having this all in the same framework would make things simpler. And if pressed, I think we have so much insane complexity in this code already, why not go all the way and at least have it consistent.
So I will side with @janiversen here - but barely :-)

core/settings/qPrefCloudStorage.cpp
```diff
@@ -7,7 +7,6 @@ static const QString group = QStringLiteral("CloudStorage");
 qPrefCloudStorage::qPrefCloudStorage(QObject *parent) : QObject(parent)
 {
 }
-
```
dirkhh, line 10: that seems like a pointless whitespace change... we typically do have an empty line before the beginning of a new function... not worth redoing things over, so just an idle nag.

dirkhh (approved):
this is just slightly longer than I like for typical PRs, and there were at least a couple that could have easily been moved to their own PR (e.g. the whitespace only commit), but I think overall this moves us in the right direction.
Yes, we can argue about the amoun of overhead, but I think this is a net positive.

janiversen:
I agree this was a bit longer than usual (blame your vacation).

Splitting it into multiple PR gives me a rebase nightmare. PR1 is based on master, PR2 on PR1, PR3 on PR2 etc.

now if you request a change in PR1, rebasing gets real fun...PR1 is simple, but PR2 and later all contains the old PR1 so here I get all kind of conflicts when rebasing. I read up/down but have not found a simple solution. This was 5 branches during primary dev. until I got too frustrated.

core/qthelper.cpp
bstoeger: > If I understand @bstoeger correctly, he points out that we are mixing two things in the 'preferences' - on the one hand these are indeed what one might describe as preferences, user visible settings. And then it's used as persistent storage for certain values (like the UUID or the last directory opened). Conceptually those two uses are indeed different. And I think one could argue either way - that they don't need all the overhead, or that it's better to have one set of well tested, stable code that is used for all circumstances.

That was not my main point. My main point is that simple code with localized logic is now split out over multiple places and replaced by complex code. You're supposed to make life easier, not harder. KISS.

The easy way would be defining trivial helper functions as I sketched above. Then write tests once for the _helper functions_. Don't replace something simple by something complex that you have to write tests for. Just imagine that for every new `report_error()` instance you would have to add a field to some structure and write a new test for it.

The only argument I have heard so far is "access from QML". But I'm neither convinced that this is necessary [that would be a long discussion], nor that it couldn't be done in a less intrusive way. QML can invoke functions, so I'm not convinced that this property stuff is needed. Again: KISS.

Also note that the whole complexity of the preferences system is due to the C/C++ split. Therefore, we have two representations of the whole preferences data, which need to kept in sync. The whole code is riddled with this problem [c.f. `DiveTripModel`, etc.], so I'm not happy if we get more of that.

Finally, I have to remark that @janiversen's tactics of choking off discussion and forcing commits (my way or the high way) leave a really bad taste in my mouth. The only reaction to my specific criticism was
1) But the other parts of the patch are nice (good, but what does that change for this particular case?).
2) Refusal to discuss (no energy for word exchange).

PS: You misspelled `UpdateManager/UUID`.

janiversen: Pardon me! I cannot let this go unanswered:
- I NEVER said anything about “my way or the high way”, nor did I think in that direction! I think I have shown. that I have no problem changing my code in response to reviews.
- I cannot force commits, unlike you I am “just” a contributor. And I am pretty sure that if I tried to force @dirkhh he would (with all rights) react forcefully.
- If you have a bad taste in the mouth, I can only recommend tooth paste.
- I am not choking off discussions, I simply notified @dirkhh that he did not need to be concerned, remember last time he felt a need to call in help. 

to your arguments:
1a. There are min. 3 other places exactly the sane as the ine you noted, so why not those, sounds illogical.
1b. there a general misunderstanding “preferences” does not mean it has to be user changeable, if so “struct preferences” needs to be changes.
2. I apologize I was unaware that you have the right to decide if I will participate in a discussion or not. 

Thanks for the PS. that is constructive review.

janiversen: @dirkhh please do me a favour and revert this PR, it clearly upsets a collaborator, and seems to cause a lot of bad vibrations. you are of course to pick and choose among the single commits as you like, this is not the highway.

I am sorry for the problems I have caused. I have on purpose for a period of time steared away from discussions to avoid confrontations like this one, However I note that the tone in some conversations (I have seens the same in PR’s not from me) is far from what I seek in a project.

For the sake of the project and my sanity, it is better that I take a longer break from the project. Github places me as #10 most active committer, not bad all considering.
Link is here (to avoid comments about how I came to that conclusion) 
https://github.com/Subsurface-divelog/subsurface/graphs/contributors

@dirkhh a personal thank to you for all the help provided. In contrast to what I am used to, I have to admit that your way of handling PR´s works pretty well

PR #1594, opened 2018-08-20
Mobile: default cylinder

### LINKED ISSUE #1535: mobile: does not apply default tank when loading new dives
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Having set a default bottle (in my case 12l, 200atm) in desktop preferences, and sharing profile with mobile, still adds “unknown” bottle to downloaded dives.

I played around and checked that the default is 12l 200ATM in my mobile version.

It can of course be that I have overseen something and default bottle is simply not implemented in mobile (at least you cannot define it in preferences), then my problem is the same, but the solution is different.

### Operating system:

macos and ios

virtual mac and ipad pro

### Subsurface version:

4.8.1 and master

### Steps to reproduce:

1) set a default bottle in desktop preferences
2) secure profile is shared (not totally sure this is needed)
3) download dives in mobile
4) verify that added bottle is not the default.

jbygdell:

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR adds the default cylinder functionality to subsurface mobile.

### Changes made:
1) Add an option to the settings page for the user to select a default cylinder.
2) Make the choice persistent by saving to disk via qPrefGeneral
3) If a default cylinder is defined select that when adding a new dive.

### Related issues:
Fixes #1535

### Release note:
Add default cylinder functionality to Subsurface mobile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Mobile: add default cylinder functionality for new dives [#1535]
 - Desktop/Export: fix bug related to quoted text when exporting to HTML [#1576]
 - Desktop/Map-Widget: add support for filtering of map locations [#1581]
 - Android: switch to Download page when dive computer is plugged in and try to
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -10,6 +10,9 @@ import org.subsurfacedivelog.mobile 1.0
 Kirigami.ScrollablePage {
 	objectName: "Settings"
 	id: settingsPage
+	property alias defaultCylinderModel: defaultCylinderBox.model
+	property alias defaultCylinderIndex: defaultCylinderBox.currentIndex
+
 	title: qsTr("Settings")
 	background: Rectangle { color: subsurfaceTheme.backgroundColor }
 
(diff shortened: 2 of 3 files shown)


jbygdell:
Derek add label: ui-mobile

janiversen (changes requested):
the changes to qmlprefs, are bot really needed and only complicates the code, you can reference qPrefGeneral directly and use the signal from there.

janiversen (changes requested):
If I apply only this commit mobile throws errors, due to e.g. when searching fir errors that each commit work as a unit.

janiversen (changes requested):
do not call loadSync (and also not sync), use the qPrefGeneral vriable directly and the set method saves to disk.

janiversen (changes requested):
why not use rebase to update the commit in question?

janiversen:
I forgot to write, you can make this change without changing C++, see tests/tst_qPrefGeneral for an example of how to use qPref directly.

jbygdell:
Rebased to match current master.

janiversen:
Your update looks good to me.

jbygdell:
@janiversen there is one thing that troubles me, the signal from qPrefs is never received on the QML side. In this case it is not a problem, but for other preferences it will be.

janiversen:
Thanks for noting this, I will add it to the test cases and see what is wrinh.

dirkhh:
This seems like a straight forward and very useful addition (the other discussion notwithstanding).
Thanks @jbygdell

PR #1599, opened 2018-08-25
Move arrays to linked lists while parsing SmartTrak tables

salvadorcunat:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A user reported recently a crash of the importer tool to mailing list.
The crash was due to a too short memory allocation for an array while parsing Fish table.  It could be fixed by simply giving the array a higher (and supposedly) unreachable space, but it doesn't looks like a good practice.
So, I've gone to a major rework and changed most tables to linked lists which seem the simpler and best option.
### Changes made:

1) Do no set CMake build type in CMakeLists, or we will be unable to build a binary with debug symbols.  Unrelated to the main bug but necessary or script selected mode would be useless.
2) Add some helper functions to manage lists (head insert, free, build ...)
3) Apply lists to table parsing  functions.
4) Disassociate table building from dive parsing. Just build the table equivalent list  once and reuse for each dive we parse.

### DIFF
--- a/smtk-import/CMakeLists.txt
+++ b/smtk-import/CMakeLists.txt
@@ -8,7 +8,6 @@ option(COMMANDLINE "Build command line version" OFF)
 set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTOUIC ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
-set(CMAKE_BUILD_TYPE "Release")
 
 set(SSRF_PATH ../)
 set(CMAKE_MODULE_PATH ${SSRF_PATH}cmake/Modules)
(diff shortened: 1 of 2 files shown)


bstoeger:
I think a much better data structure would have been a dynamically (re)allocated array. But this is not a show-stopper, just a bit of a "smell".

janiversen:
I agree with @bstoeger, because not only are linked lists a factor slower for most operations, but are also often causing strange problems. Reallocating a dynimac array the few times there are not enough space sounds a lot simpler to me.

dirkhh:
This is obviously not a part of the project that I spend too much focus on - but I agree with the sentiment here. Linked list are rarely the right answer.
Yet, this is your code, you maintain it, so I see no reason to over-rule you here, @salvadorcunat

PR #1603, opened 2018-08-27
Filter: don't count dives on mouse-hover

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

While working on integrating #1528 with the filters, I noticed that we have some very crazy behavior:

In `FilterModelBase::data()`, when accessing the name of a filter-item, the dives are counted. This is fired - among other things - for every mouse-hover event(!). It's amazing that people with huge dive logs didn't notice (especially since some people are reportedly working with 32-bit machines).

This commit fixes this problem by calculating the counts once on repopulating the filters. Moreover, this is a first step towards proper qt-model semantics for the filter-boxes [insert / move / remove].

### Changes made:

1) Turn the checked-flag into a struct.
2) Add a count item to the struct.
3) Count dives on repopulate and then use the cached value.

### Related issues:

#1528

### Additional information:

None.

### Release note:

Not needed (not really user-visible).

### Documentation change:

Not needed (not really user-visible).

### DIFF
--- a/qt-models/filtermodels.cpp
+++ b/qt-models/filtermodels.cpp
@@ -32,30 +32,41 @@ FilterModelBase::FilterModelBase(QObject *parent) : QStringListModel(parent),
 {
 }
 
-// Update the stringList and the checkState array.
+// Update the stringList and the items array.
 // The last item is supposed to be the "Show Empty Tags" entry.
 void FilterModelBase::updateList(const QStringList &newList)
 {
-	// Keep copy of old checkState array to reimport them later.
-	std::vector<char> oldCheckState = checkState;
-	checkState.resize(newList.count());
-	std::fill(checkState.begin(), checkState.end(), false);
-	anyChecked = false;
+	// Keep copy of the old items array to reimport the checked state later.
+	// Note that by using std::move(), this is an essentially free operation:
+	// The data is moved from the old array to the new one and the old array
+	// is reset to zero size.
+	std::vector<Item> oldItems = std::move(items);
 
-	// Ignore last item, since this is the "Show Empty Tags" entry.
-	for (int i = 0; i < rowCount() - 1; i++) {
-		if (oldCheckState[i]) {
+	// Resize the cleared array to the new size. This leaves the checked
+	// flag in an undefined state (since we didn't define a default constructor).
+	items.resize(newList.count());
+
+	// First, reset all checked states to false
+	anyChecked = false;
+	for (Item &item: items)
+		item.checked = false;
+
+	// Then, restore the checked state.  Ignore the last item, since
+	// this is the "Show Empty Tags" entry.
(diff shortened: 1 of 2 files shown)


qt-models/filtermodels.cpp
```diff
@@ -32,30 +32,38 @@ FilterModelBase::FilterModelBase(QObject *parent) : QStringListModel(parent),
...
-	// Keep copy of old checkState array to reimport them later.
-	std::vector<char> oldCheckState = checkState;
-	checkState.resize(newList.count());
-	std::fill(checkState.begin(), checkState.end(), false);
+	// Keep copy of the old items array to reimport the checked state later.
+	// Note that by using std::move(), this is an essentially free operation:
+	// The data is moved from the old array to the new one and the old array
+	// is reset.
+	std::vector<Item> oldItems = std::move(items);
+	items.resize(newList.count());
```
dirkhh: So this moves the existing `items` to `oldItems` and then resizes `items` to the new length of the `newList`... but then what's the content of `items`? Are all members zeroed out? Or undefined?
In the loop below you just set `.checked` to false, so I assume the members are undefined?

qt-models/filtermodels.cpp
```diff
@@ -32,30 +32,38 @@ FilterModelBase::FilterModelBase(QObject *parent) : QStringListModel(parent),
...
+
+	// First, reset all checked states to false
 	anyChecked = false;
+	for (Item &item: items)
```
dirkhh, line 51: inconsistent whitespace - this is one construct that I don't think we list in CodingStype. Here you use 
`for (type single: vector)` with no space before the `:`, below in line 124 there's a space before the `:`

dirkhh (changes requested):
A couple of small comments - I'd appreciate if you fixed the white space (I realized later that this may have been inconsistent before you touched it)...

qt-models/filtermodels.cpp
bstoeger: The assignment with `std::move` essentially resets the source-vector to zero size (in this case we could also simply have swapped the new and the old vectors - that would have been the pre-C++11 idiomatic thing to do).

The resize leaves the contents of the struct in an undefined state because we didn't specify a default constructor and the PODs [plain old data type] such as `bool` and `int` don't have a default constructor either. Resize is conceptually the same as `malloc`/`realloc`ing a dynamic array.

I extended the comments.

qt-models/filtermodels.cpp
bstoeger: Added a white space commit. The version without space seems to be the more common one.

dirkhh (approved):

PR #1604, opened 2018-08-27
Usb storage branch

torvalds:
I'm finally looking at the Garmin Descent Mk1 now that the merge window is over.

And this does *not* add support for it, but this does add the scaffolding to subsurface for libdivecomputer supporting it. Eventually.

It basically does a first pass of trying to generalize some of the Uemis downloader support as a "USB Storage" support, and remove the special cases that made it Uemis only.

NOTE! Some special cases still remain, eg the whole OS specific "find the Uemis mount point" has just been hacked up into a "find the Uemis or Garmin mount point". 

But the core downloader code now just looks at "what are the supported download protocols of the dive computer" and the device node lookup looks at "do I want DC_PROTOCOL_SERIAL nodes" etc. It used to have a special "dc_type" (which was really "Uemis or not") rather than try to look at what protocols the dive computer supported.

Again - this should not change anything for existing dive computer support, it's purely preparatory for an eventual libdivecomputer entry for the Garmin Descent.

### DIFF
--- a/core/android.cpp
+++ b/core/android.cpp
@@ -81,7 +81,7 @@ const char *system_default_filename(void)
 	return path;
 }
 
-int enumerate_devices(device_callback_t callback, void *userdata, int dc_type)
+int enumerate_devices(device_callback_t callback, void *userdata, unsigned int transport)
 {
 	/* FIXME: we need to enumerate in some other way on android */
 	/* qtserialport maybee? */
--- a/core/display.h
+++ b/core/display.h
@@ -2,6 +2,8 @@
 #ifndef DISPLAY_H
 #define DISPLAY_H
 
+#include "libdivecomputer.h"
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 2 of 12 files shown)


torvalds:
It goes without saying that I haven't tested the Windows or OSX or Android side. But hopefully my "obvious" changes actually work. I guess Travis will tell.

neolit123:
the only thing that Travis complained about is the ios job:
```
/Users/travis/build/Subsurface-divelog/subsurface/core/ios.cpp:71:5: error: conflicting types for 'enumerate_devices'
int enumerate_devices(device_callback_t, void *, int)
    ^
In file included from /Users/travis/build/Subsurface-divelog/subsurface/core/ios.cpp:7:
../../../core/display.h:49:5: note: previous declaration is here
int enumerate_devices(device_callback_t callback, void *userdata, unsigned int transport);
    ^
1 error generated.
```

torvalds:
Ok, I'll fix that up. It just needs to be 'unsigned int'.

I have one more commit anyway to actually do the 'dc_usb_storage_open()' part, so I'll just create a new fixed branch.

PR #1605, opened 2018-08-27
Show new firmware notice for OSTC Plus

janmulder:

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Also show a new firmware notice for a HW OSTC Plus. This appeared to be broken (as can be verified on the current master from today, as HW just released version 2.98 of the firmware).

### Changes made:
See commit

### Additional information:
Notice that there is some confusion in both Subsurface and Libdivecomputer with respect to types of OSTCs. Physically, there exist 2 type of OSTC3's. The first edition with only DC_TRANSPORT_SERIAL (and an USB connector), and a second version that lacks the connector but supports both BT and BLE. Confusingly, the second version is marked OSTC3 on the device, but the user needs to select OSTC Plus (which is basically a 3rd version of the OSTC3, combined with a successor of the OSTC Sport) to download dives using BT/BLE.

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.cpp
+++ b/desktop-widgets/configuredivecomputerdialog.cpp
@@ -236,7 +236,7 @@ OstcFirmwareCheck::OstcFirmwareCheck(QString product) : parent(0)
 {
 	QUrl url;
 	memset(&devData, 1, sizeof(devData));
-	if (product == "OSTC 3" || product == "OSTC 3+" || product == "OSTC cR") {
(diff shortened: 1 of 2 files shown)


dirkhh:
Thanks for catching this.
The model confusion is annoying. I wonder if we should fix this in libdivecomputer... but of course I'm not quite sure HOW to fix this in libdivecomputer... because claiming that the OSTC 3 supports BLE will lead to it being shown as supported on iOS... which of course the older models aren't.

PR #1606, opened 2018-08-27
Updates usb-storage branch

torvalds:
This is the second try at unifying the usb storage dive computer model, now with the iOS build fixed, and with a commit to actually do the "dc_usb_storage_open()" call when a USB storage dive computer exists.

With my current libdivecomputer skeleton driver, this actually enumerates the dive files on the Garmin Descent Mk1. NOTE! It does not actually parse them, and I haven't made my in-progress work public, but I think this is sufficient to support the Garmin Descent on the subsurface side.

There's a lot more work for me on the libdivecomputer side before the Garmin Descent works, but that's just a SMOP. Eventually.

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.cpp
+++ b/desktop-widgets/configuredivecomputerdialog.cpp
@@ -398,11 +398,11 @@ static void fillDeviceList(const char *name, void *data)
 	comboBox->addItem(name);
 }
 
-void ConfigureDiveComputerDialog::fill_device_list(int dc_type)
+void ConfigureDiveComputerDialog::fill_device_list(unsigned int transport)
 {
 	int deviceIndex;
 	ui.device->clear();
-	deviceIndex = enumerate_devices(fillDeviceList, ui.device, dc_type);
+	deviceIndex = enumerate_devices(fillDeviceList, ui.device, transport);
 	if (deviceIndex >= 0)
 		ui.device->setCurrentIndex(deviceIndex);
 }
@@ -1445,12 +1445,12 @@ void ConfigureDiveComputerDialog::on_DiveComputerList_currentRowChanged(int curr
 		return;
 	}
 
-	int dcType = DC_TYPE_SERIAL;
+	unsigned int transport = DC_TRANSPORT_SERIAL;
 
 
 	if (selected_vendor == QString("Uemis"))
-		dcType = DC_TYPE_UEMIS;
-	fill_device_list(dcType);
+		transport = DC_TRANSPORT_USBSTORAGE;
+	fill_device_list(transport);
 }
 
 void ConfigureDiveComputerDialog::checkLogFile(int state)
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -143,7 +143,7 @@ void DownloadFromDCWidget::updateState(states state)
 		return;
 
 	if (state == INITIAL) {
-		fill_device_list(DC_TYPE_OTHER);
+		fill_device_list(~0);
 		ui.progressBar->hide();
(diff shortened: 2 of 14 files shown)


torvalds:
Let's see if the iOS fixup that I had missed the last time makes it all work now.

Of course, maybe my dc_usb_storage_open() commit now broke. Fingers crossed.

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -229,15 +229,17 @@ void DownloadFromDCWidget::updateState(states state)
 	currentState = state;
 }
 
+// FIXME! Get the transport list from libdivecomputer!
```
dirkhh, line 232: not sure I understand the FIXME. Isn't that what we are doing below?

desktop-widgets/configuredivecomputerdialog.cpp
```diff
@@ -1445,12 +1445,12 @@ void ConfigureDiveComputerDialog::on_DiveComputerList_currentRowChanged(int curr
 		return;
 	}
 
-	int dcType = DC_TYPE_SERIAL;
+	unsigned int transport = DC_TRANSPORT_SERIAL;
 
 
 	if (selected_vendor == QString("Uemis"))
```
dirkhh, line 1451: I assume that the reason we don't need to hardcode "Garmin" here is that libdivecomputer will do the right thing for us here?

dirkhh (approved):
This looks really good - there are two small questions, but overall I think this is great to see!

desktop-widgets/downloadfromdivecomputer.cpp
torvalds: Duh. That FIXME is a left-over from something I already fixed.

desktop-widgets/configuredivecomputerdialog.cpp
torvalds: No, it's actually simpler than that - only Uemis supports configuration.

torvalds: That's also why that configuredivecomputerdialog.cpp has the hardcoded transports: we don't support configuration in the general case.

This code would all go away if we made the configuration code be a button in the download window instead. Right now we duplicate a lot of the download setup and computer choice in the configuration window. Just sayin'

torvalds (commented):
Replying to the individual questions

desktop-widgets/configuredivecomputerdialog.cpp
dirkhh: I implemented that. And got rather strong pushback, especially from @glance-

dirkhh:
merged manually to get rid of the obsolete FIXME

desktop-widgets/configuredivecomputerdialog.cpp
glance-: We don't support configuring Uemis computers, and thats rather just a copy paste thingie from when the initial code was written as a part of gsoc2014. All those traces of Uemis can be removed.

It's not all that much that could be removed on a merger. It was way more right after gsoc2014, but I've cleaned it up quite a bit.

In #1571 , My intention was to rather point out which issues I saw with the proposed code, and what was left and needed to be done to have a sane flow.

dirkhh: TBH, the feedback felt much more like "this is my code, I use it all the time, you're making this harder". But maybe I was just in a sensitive mood when I read the comments. Either way, I agree with Linus and I continue not to like having that "configure dive computer" menu setting - but that is exactly what you appear to WANT to keep because it's what's in the video and in the guide and it's what you use. And I didn't feel like fighting over it.

PR #1611, opened 2018-08-29
Dive trips: don't crash on repeated trip-removal

bstoeger:
In the UI it is possible to remove a dive from a trip twice,
which leads to a crash, because trip is NULL (obviously).

Instead of doing a proper fix (don't show the "remove from
trip" entry in the first place), ignore dives without a
trip, since a rewrite of the undo-code is planned for the
medium future anyway.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The commit message says it all: avoid null pointer access on double-trip-removal.

### Changes made:

1) Check if dive belongs to trip before removing it.

### Additional information:

None.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -684,9 +684,12 @@ void DiveListView::removeFromTrip()
 	struct dive *d;
 	QMap<struct dive*, dive_trip*> divesToRemove;
 	for_each_dive (i, d) {
-		if (d->selected)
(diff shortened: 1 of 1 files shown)


dirkhh:
seems rather obvious (in hindsight).
Thanks

Issue #1608, opened 2018-08-28
Subsurface segfaults when opening preferences (Related to  #1401?)


p-neves:
### Describe the issue:
Subsurface segfaults when opening preferences

- [ X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When opening program preferences, Subsurface segfaults

### Operating system:

Arch Linux

### Subsurface version:

4.8.1

Yes. With Subsurface-4.7.8-x86_64.AppImage

Self compiled from master

4.8.1-303-g66aeaddd0f9c

### Steps to reproduce:

1) Start Subsurface
2) Select File/Preferences

### Current behavior:

Subsurface segfaults

### Expected behavior:

Open application preferences

### Additional information:

It is probably related to my logbook (it's getting huge!). Please see issue #1401.

Backtrace follows:

gdb bt:

#0  0x00007ffff1629667 in __strlen_avx2 () at /usr/lib/libc.so.6
#1  0x000055555593a457 in QString::QString(char const*) (ch=0xffffffff <error: Cannot access memory at address 0xffffffff>, this=0x7fffffffcdc0) at /usr/include/qt/QtCore/qstring.h:693
#2  0x000055555593a457 in PreferencesDefaults::refreshSettings() (this=0x5555565af1d0) at /home/xxx/Diveapps/subsurface/desktop-widgets/preferences/preferences_defaults.cpp:72
#3  0x000055555592f0fd in PreferencesDialog::refreshPages() (this=0x55555622aba0) at /home/xxxxx/Diveapps/subsurface/desktop-widgets/preferences/preferencesdialog.cpp:111
#4  0x00005555556b0661 in MainWindow::on_actionPreferences_triggered() (this=<optimized out>) at /home/paneves/Diveapps/subsurface/desktop-widgets/mainwindow.cpp:814
#5  0x000055555575782a in MainWindow::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) (_o=0x555555e245e0, _c=QMetaObject::InvokeMetaMethod, _id=12, _a=0x7fffffffd030)
    at /home/paneves/Diveapps/subsurface/build/desktop-widgets/subsurface_interface_autogen/EWIEGA46WW/moc_mainwindow.cpp:356
[...]

PR #1612, opened 2018-08-30
Filter: unify data structure of FilterBaseModel

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

`FilterBaseModel` is a strange hybrid. It derives from `QStringListModel` to keep track of the items, but has additional data (`checked`, `count`) in a separate list. Make it a "real" Model by also moving the name into the structure. Start implementing model semantics beyond full reset, viz. adding an item and renaming an item.

Ultimately, this should lead to dynamically updated filters on do and undo.

This PR should probably wait until #1607 and #1608 are understood.

### Changes made:

1) Remove a signal that ultimately should be put at a different place.
2) Move the name into the item struct in `FilterModelBase`.
3) Derive `FilterModelBase` from `QAbstractListItem`.
4) Implement `addItem` and rename functionality.
5) Sort the list in `updateList()`, not the individual callers.

### Related issues:

None.

### Additional information:

None.

### Release note:

No.

### Documentation change:

No.

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -211,7 +211,6 @@ MainTab::MainTab(QWidget *parent) : QTabWidget(parent),
 
 	connect(ui.diveNotesMessage, &KMessageWidget::showAnimationFinished,
 					ui.location, &DiveLocationLineEdit::fixPopupPosition);
-	connect(this, SIGNAL(diveSiteAdded(const QString &)), LocationFilterModel::instance(), SLOT(addName(const QString &)));
 
 	// enable URL clickability in notes:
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
I don't think we are any closer to understanding these two issues.
Unless you disagree @bstoeger, I'd go ahead and merge this instead of letting it sit here and age. Best case it magically fixes the bugs :-)

bstoeger:
From my side you can commit this. I don't see how this could fix the bugs - there seems to be some core-data structure corruption. :(

dirkhh:
I wonder if #1616 is possible behind the crash. Dereferencing an invalid QVariant sounds like a bad idea :-)
I'll merge this in the meantime.

Issue #1614, opened 2018-08-31
Valgrind complains about use of uninitialized data in deco-calculation


bstoeger:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I'm not sure if this is an actual bug, but I get the following reports with `valgrind`:

```
==18712== Conditional jump or move depends on uninitialised value(s)
==18712==    at 0x382C7D: tissue_tolerance_calc (deco.c:247)
==18712==    by 0x399111: init_decompression (divelist.c:627)
==18712==    by 0x3D85B4: create_plot_info_new (profile.c:1330)
==18712==    by 0x31BAF5: ProfileWidget2::plotDive(dive*, bool, bool) (profilewidget2.cpp:640)
==18712==    by 0x233D39: MainWindow::current_dive_changed() (mainwindow.cpp:539)
==18712==    by 0x8CBC6CE: QMetaObject::activate(QObject*, int, int, void**) (in /usr/lib/x86_64-linux-gnu/libQt5Core.so.5.9.5)
==18712==    by 0x2DC8FD: DiveListView::selectionChanged(QItemSelection const&, QItemSelection const&) (divelistview.cpp:582)
==18712==    by 0x6B42930: ??? (in /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5.9.5)
==18712==    by 0x8CBC5B4: QMetaObject::activate(QObject*, int, int, void**) (in /usr/lib/x86_64-linux-gnu/libQt5Core.so.5.9.5)
==18712==    by 0x8C39616: QItemSelectionModel::selectionChanged(QItemSelection const&, QItemSelection const&) (in /usr/lib/x86_64-linux-gnu/libQt5Core.so.5.9.5)
==18712==    by 0x8C3DEAF: QItemSelectionModel::emitSelectionChanged(QItemSelection const&, QItemSelection const&) (in /usr/lib/x86_64-linux-gnu/libQt5Core.so.5.9.5)
==18712==    by 0x8C41082: QItemSelectionModel::select(QItemSelection const&, QFlags<QItemSelectionModel::SelectionFlag>) (in /usr/lib/x86_64-linux-gnu/libQt5Core.so.5.9.5)
==18712== 
==18712== Conditional jump or move depends on uninitialised value(s)
==18712==    at 0x382DD7: tissue_tolerance_calc (deco.c:253)
[...]

PR #1615, opened 2018-08-31
Zero initialize deco state struct

atdotde:
Valgrind found use of some uninitialized variable (probably
ds->gf_low_pressure_this_dive ), see #1614 . Zero is the correct
value to start with. Lacking a working version of valgrind I cannot
check this actually fixes the problem.

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh (approved):
That seems like the right thing to do either way

PR #1617, opened 2018-08-31
Cleanup: make DiveTripModel a global object

bstoeger:
DiveTripModel (the model describing the dive-list) was destroyed
and recreated on every reset of the list. This seems excessive.
Instead - in analogy to most other models - make it a single
global object.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a cherry-pick from #1528, which makes the `DiveTripModel` a global object instead of regenerating it all the time. Since this is still based on `TreeModel`, I had to add a `clear()` function to tree model.

The reason for cherry-picking this is that setting the source model seems to make problems under not-yet-understood circumstances (see #1607)

### Changes made:

1) Add `clear()` function to `TreeModel`
2) Move `trips` from `DiveTripModel` class to function.
3) Make `DiveTripModel` a single global object.

### Related issues:

#1528, #1607
### Release note:

No.
### Documentation change:

No.
### Mentions:

@dirkhh @janmulder

### DIFF
--- a/qt-models/treemodel.h
+++ b/qt-models/treemodel.h
@@ -4,6 +4,7 @@
 
 #include <QAbstractItemModel>
 #include <QCoreApplication>
+#include <memory>
 
 struct TreeItem {
 	Q_DECLARE_TR_FUNCTIONS(TreeItemDT)
@@ -25,16 +26,16 @@ class TreeModel : public QAbstractItemModel {
 	Q_OBJECT
 public:
 	TreeModel(QObject *parent = 0);
-	~TreeModel();
 	QVariant data(const QModelIndex &index, int role) const;
 	int rowCount(const QModelIndex &parent = QModelIndex()) const;
 	int columnCount(const QModelIndex &parent = QModelIndex()) const;
 	QModelIndex index(int row, int column, const QModelIndex &parent = QModelIndex()) const;
 	QModelIndex parent(const QModelIndex &child) const;
+	void clear();
 
 protected:
 	int columns;
-	TreeItem *rootItem;
+	std::unique_ptr<TreeItem> rootItem;
 };
 
 #endif
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -36,8 +36,7 @@ DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelec
 	model->setSortRole(DiveTripModel::SORT_ROLE);
 	model->setFilterKeyColumn(-1); // filter all columns
 	model->setFilterCaseSensitivity(Qt::CaseInsensitive);
-	DiveTripModel *tripModel = new DiveTripModel(this);
-	model->setSourceModel(tripModel);
+	model->setSourceModel(DiveTripModel::instance());
 	setModel(model);
 
 	setSortingEnabled(false);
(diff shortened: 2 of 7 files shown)


qt-models/treemodel.h
```diff
@@ -25,16 +26,16 @@ class TreeModel : public QAbstractItemModel {
...
 	QVariant data(const QModelIndex &index, int role) const;
 	int rowCount(const QModelIndex &parent = QModelIndex()) const;
 	int columnCount(const QModelIndex &parent = QModelIndex()) const;
 	QModelIndex index(int row, int column, const QModelIndex &parent = QModelIndex()) const;
 	QModelIndex parent(const QModelIndex &child) const;
+	void clear();
 
 protected:
 	int columns;
-	TreeItem *rootItem;
```
neolit123, line 37: QScopedPointer is the Qt alternative here.

i still don't know if we have a policy on C++ stdlib vs Qt.
like, when to use what? is there are a reason for using one over the other? etc.

my thinking would be - use Qt unless stdlib is better at something.

neolit123 (approved):

janmulder:
Tried this PR in relation to the discussion in #1607. For me, the SIGSEGV when deleting the last dive of a trip is fixed. Not really surprising as that crash seems to be related to recreating the divelist model over and over again.

qt-models/treemodel.h
bstoeger: Here, `QScopedPointer` would work. But in general, I had such a bad experience with `QScopedPointer` that this is the one case where I would generally go with `std::unique_ptr`. The difference is that the latter has move-semantics, which means that you can pass down ownership. For example you can put `std::unique_ptr`s into (stdlib) containers. See also https://stackoverflow.com/questions/40346393/should-i-use-qscopedpointer-or-stdunique-ptr

For `QVector` vs `std::vector`, etc. I think we should decide on a case-by-case basis. I tend to default to the former, even though I much prefer the latter. They are surprisingly different: Qt does reference counting and copy-on-write, whereas stdlib is just syntactic sugar on top of C `malloc`/`realloc`. You pass it by value and it will make a deep copy (which is what passing by value means). Qt devs argue that their containers are more "beginner"-friendly, because you don't have to think about ownership, but I have my doubts because this leads to quite unpredictable behavior. Personally, I think Qt's containers are broken, but that it's not worth fighting them.

Also Qt-versions have convenience functions such as `vector.contains(element)`, vs. in stdlib you would write a single general function for that: `contains(general_container, element)`. The second is much closer to C, no?

neolit123: thank you for the explanation.
the overall move to use std::* more widely in the codebase was a bit of a surprise, mostly because nobody investigated differences between the standard library and Qt containers before.

if one tries to contribute stdlib code in the Qt source code, it would be rejected for consistency reasons - i.e. use the Qt containers.
so from a consistency standing point this is not exactly good for Subsurface, but i guess we can do it here and there as long as the code is readable.

we should keep in mind that the C++ standard library tends to not be very readable.

bstoeger: Yes, with a bit of a heavy heart, I try to use Qt-containers unless COW semantics make it impossible or distinctly more difficult. An example is putting owning pointers in containers. Suppose you had a `QVector<std::unique_ptr>`. Modifying means possible copying and - boom - you would suddenly have two versions of the unique-ptr pointing to the same object, which is by definition not possible. In my opinion this just shows that Qt containers are broken with respect to ownership management, but what can you do. Instead of fixing their containers they decided to invent QML. :(

dirkhh:
I think this is certainly the right thing to do, so let's merge this one and see if this moves the needle on the crashes.

Issue #1619, opened 2018-09-01
New filter model causing new warnings on MAC.


janiversen:
### Describe the issue:

- [ x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When compiling on Mac for Mac or iOS the following warnings are issues quite a lot of times:

In file included from /Users/jani/develop/src/subsurface/desktop-widgets/divelistview.cpp:8:
/Users/jani/develop/src/subsurface/./qt-models/filtermodels.h:42:16: warning: 'flags' overrides a
      member function but is not marked 'override' [-Winconsistent-missing-override]
        Qt::ItemFlags flags(const QModelIndex &index) const;
                      ^
In file included from /Users/jani/develop/src/subsurface/desktop-widgets/divelistview.cpp:8:
/Users/jani/develop/src/subsurface/./qt-models/filtermodels.h:43:11: warning: 'data' overrides a
      member function but is not marked 'override' [-Winconsistent-missing-override]
        QVariant data(const QModelIndex &index, int role = Qt::DisplayRole) const;
                 ^
In file included from /Users/jani/develop/src/subsurface/desktop-widgets/divelistview.cpp:8:
/Users/jani/develop/src/subsurface/./qt-models/filtermodels.h:44:7: warning: 'setData' overrides a
      member function but is not marked 'override' [-Winconsistent-missing-override]
        bool setData(const QModelIndex &index, const QVariant &value, int role = Qt::EditRole);
             ^

Remark: removed lines to make the issue shorter

This is something new that was not present a couple of weeks ago.

### Operating system:

osx newest, xcode newest, Qt 11.1

### Expected behavior:

no warnings.

bstoeger:
Gah. At this point I would opt for passing `-Wno-inconsistent-missing-override` to clang, as this warning actually *hinders* us adding `override`s.

janiversen:
sounds wrong, if these were to be overwritten, they would have been marked so. Supressing warnings just because we use the system different does not sound like a correct implementation.

But of course if we do not care about these kind of warnings, change the makefiles.

bstoeger:
>  if these were to be overwritten, they would have been marked so.

I am not sure if you get the point of the `override` keyword? The functions *are* overridden, since they have the same function signature as the virtual functions in the base class. By adding the `override` keyword failing to do so (because we get the signature subtly wrong) results in a hard compilation error.

[...]

PR #1622, opened 2018-09-01
Build: suppress inconsistent-missing-override warnings on clang

bstoeger:
In a previous commit all override-modifiers were thrown out owing
to warning floods caused by the inconsistent-missing-override
flag on clang.

Re-adding overrides is now very painful because it's an all-or-nothing
thing. Let's disable the warning for now. If we reastablished good
coverage with override, we might think about re-enabling the warning.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Disable the warnings complained about in #1619. We might want to re-add them once we have more coverage. Though in Qt-context there are so many virtual functions that I'm not sure an all-or-nothing approach to `override` is really sensible.

### Changes made:

1) Pass `-Wno-inconsistent-missing-override`

### Related issues:

#1619
### Release note:

No.
### Documentation change:

No.
### Mentions:

@janiversen @dirkhh

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,7 @@ add_definitions(-DSUBSURFACE_SOURCE="${SUBSURFACE_SOURCE}")
 #
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 ")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu99 -Wno-inconsistent-missing-override")
 	if((${CMAKE_SYSTEM_NAME} MATCHES "Darwin") AND
 	   ((${CMAKE_SYSTEM_VERSION} MATCHES "11.4.") OR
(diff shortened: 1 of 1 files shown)


janiversen:
looks good, just need to check if it covers iOS as well

CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ add_definitions(-DSUBSURFACE_SOURCE="${SUBSURFACE_SOURCE}")
 #
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 ")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu99 -Winconsistent-missing-override")
```
dirkhh: shouldn't this be `-Wno-inconsisten-missing-overrides`? Or am I completely misunderstanding how this works?

bstoeger:
Good catch, forgot to `git commit --amend` before push. 🙄

PR #1625, opened 2018-09-02
Fix appdata and missing application icon

Alexander-Wilms:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

The [Freedesktop spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html) recommends naming resources using the reverse DNS format. I changed the file names of icons, the .desktop file and the .appdata.xml file. I also added a call to QGuiApplication's setDesktopFileName in qt-init.cpp, without which a placeholder icon is shown in the titlebar when running Subsurface on KDE w/ Wayland.

### Changes made:

1) Rename application icons, .desktop file and .appdata.xml file
2) Fix appdata to that it passes validation using `appstreamcli validate`
2) Use QGuiApplication::setDesktopFileName() to fix a broken icon when using Subsurface on KDE w/ Wayland


appdata/org.subsurface_divelog.Subsurface.appdata.xml
```diff
@@ -0,0 +1,45 @@
...
+      <image>https://subsurface-divelog.org/screenshots/preferences2.png</image>
+    </screenshot>
+  </screenshots>
+  <url type="homepage">https://subsurface-divelog.org</url>
+  <url type="bugtracker">https://github.com/Subsurface-divelog/subsurface/issues</url>
+  <url type="faq">https://subsurface-divelog.org/faq/</url>
+  <url type="help">https://subsurface-divelog.org/documentation/</url>
+  <url type="translate">https://www.transifex.com/subsurface/subsurface/</url>
+  <releases>
+     <release version="4.8.1" date="2013-07-14" />
```
Alexander-Wilms: Ideally, this field would be automatically updated at the same time that ssrf-version.h is generated.

Alexander-Wilms:
It seems like `linuxdeployqt` has an issue with icon names containing multiple periods. I believe this is a bug.

janiversen:
Apart from being a recommendation, are you having problems with the old filenames?

I would prefer to keep the old shorter icon names without multiple periods.

Alexander-Wilms:
Not sure how relevant it is to you, but Flatpak expects these files to have reverse DNS names. Quite a few upstream projects have adopted this naming scheme.

Alexander-Wilms:
Alternatively, I could limit this PR to fixing the appdata and adding QGuiApplication::setDesktopFileName().

janiversen:
What happens in flatpack if the icons do not follow that naming standard (e.g. LibreOffice have tons of icons that do not follow that standard, and we still produce a flatpack version) ?

Using reverse DNS naming, how do you make 1 icon file available for multiple applications (e.g. desktop and mobile of course with different domains) ?

It is not my call, but I would limit the PR to what is really a problem, and I am still unsure if there are a real problem ?

Alexander-Wilms:
> What happens in flatpack if the icons do not follow that naming standard (e.g. LibreOffice have tons of icons that do not follow that standard, and we still produce a flatpack version) ?

There are options to automically rename icons, appdata and desktop file: https://github.com/flathub/org.subsurface_divelog.Subsurface/blob/master/org.subsurface_divelog.Subsurface.yaml#L8

LibreOffice uses a script which renames the relevant files: https://github.com/LibreOffice/core/blob/master/solenv/bin/assemble-flatpak.sh#L43

> Using reverse DNS naming, how do you make 1 icon file available for multiple applications (e.g. desktop and mobile of course with different domains) ?

I assume one could use symlinks.

> It is not my call, but I would limit the PR to what is really a problem, and I am still unsure if there are a real problem ?

The only actual problems are that the appdata contains some invalid xml tags and that the application icon is not shown when running Subsurface under Wayland. I'll update the PR later.

janiversen:
I know the script that LO uses, and the same could be done here, no need to rename the original files.

Alexander-Wilms:
I assume the one failed build isn't related to these changes.

bstoeger:
> I assume the one failed build isn't related to these changes.

It isn't. If you rebase on current master the problem should be gone.

PR #1626, opened 2018-09-02
qPref: use shadow variable to detect direct changes in struct preferences.

### LINKED ISSUE #1623: Preferences: final sync takes forever
### Describe the issue:

- [X] Bug ?
- [ ] Change request
- [ ] New feature request
- [x] Discussion request

### Issue long description:
Since the new preferences code, after exit of the UI (using the menu-quit), it takes about 15 seconds before my `bash` prompt is shown. Investigation shows that a lot of time is spend in the final preferences sync (`qPref::sync();`). My Arch system does not have a working valgrind at the moment (bug in valgrind) so further investigation where those 15 seconds are spend is not easy. 

### Operating system:
OS: Arch Linux (upto date)

### Subsurface version:
What version of Subsurface are you running? master
Does the same happen on another Subsurface version?: no, seems introduces in the big preferences rewrite

### Steps to reproduce:
Start Subsurface, quit, and check the time the command prompt returns

### Current behavior:
15 seconds spend, `top` shows no relevant cpu usage, so it feels like a 15 second stall.

### Expected behavior:
Minimal delay between UI exit and process end. Writing some preferences to a simple ASCII file cannot be that hard.

@janiversen

### LINKED ISSUE #1624: opening a log file makes subsurface 4.8 use CPU100% and hang up
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
I'm running subsurface 4.8.0-1~xenial on a Debian stable (with some older libssl as the subsurface packages do work with current Debians)...
Opening my version="3" logfile makes the CPU go to 100% and the process of opening the log never finishing. Eventually I have to kill subsurface process.

It works fine again when going back to 4.7.6-1~zesty.

### Operating system:
Debian stable
OSTC4

### Subsurface version:
4.8.0-1~xenial

Works again with 4.7.6-1~zesty.

The "official" release from ppa.launchpad.net/subsurface/subsurface/ubuntu

Any ideas on how to fix that?

janiversen:
Add a shadow variable to all DISK_* macros, in order to check
if the original variable in struct preferences is changed.

Only save to disk if actually changed

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Added a static variable to DISK_* allowing to detect whether the variable in struct preferences 
was changed and thus needs to be saved on disk.

This is a quick fix that hugely removes the number of disk operations in qPref::sync(), however this
is a temporary solution. The correct solution is to guard access to the preference variables.

### Documentation change:

fixes #1623

### Mentions:

@janmulder


janiversen:
I knew this would have side effect, let me fix that

janmulder:
Tried and LGTM. The exit time is normal again for me, and not saving unchanged data is IMHO always good. 

That said ... as I'm the only one that sees this issue, I'm not sure this sort of temporary solution is this the way to go.

janiversen:
thanks a lot for testing. The word "temporary" is not totally correct, this is the first step in providing a guard around the property variables.

We do not know how many else noted a change, after all this is master and not yet a release, so not merging this PR might cause trouble in a later release.

bstoeger:
With respect to this and previous problems, I would like to ask you for a big favor: Could you (and perhaps others) pause for a moment and contemplate on whether *conceptually* we were in a better situation before the preferences-rewrite? I'm not talking about code-quality (I didn't have a closer look), just about overall design. It is normal with new code that bugs show up, but here I see a stream of problems of a class that we didn't have before and therefore I think it is fair to ask if these new problems are justifiable.

I'll try to give a short analysis of the before / now situation. I'd love to be shown wrong.

- We work with the Qt framework that has an interface for storing settings. The interface may be a bit verbose, but that's an implementation detail. The framework is portable, well tested and hides all the loading/synchronizing away from us. Not perfect, but reasonable.

- So I hope that we agree that re-implementing the functionality of QSettings is not desirable, especially if we re-implement QSettings on top of QSettings.

- Then why did we have the previous preferences madness? I think the root-cause for that is that C-code directly accesses the "prefs" structure. So we somehow had to synchronize "prefs" and QSettings. That ended up in a horrible mess.

- So my first conclusion would be that we want to move things *away* from prefs, because all that conceptual complexity would go away by itself. Therefore, I have a big problem moving LastPath, etc *into* prefs.

- What the old code did was read the preferences on startup and fill the prefs structure. Not nice, but what can you do if you don't want to touch the C part?

- Nevertheless, *updating* was left to QSettings. Now the way I read this bug, we also do the writing? If this is the case (perhaps I'm misunderstanding things), then this needs some really good reason. I can't remember anybody complaining about QSettings' syncing. So what is the reason for this new class of problems and this new complexity? What original problem does it fix? Can't we simply remove all that complexity by letting QSettings do its thing? After all, that's the whole point of it.

Then I have a second point to contemplate about (the view of the user - sadly something developers often forget):

- The point of libraries / subsystems is to hide complexity and make it easy for the caller. I consider myself a user of the settings machinery.

- Consider a simple task: I want to remember the last path of video files.

- With the old system this was very easy:
 1) Read from QSettings before opening the file requester.
 2) Write to QSettings after getting a file.

- With the new system:
 1) Call into Q...before opening the file requester.
 2) Call into Q.. after getting a file
 Note that the syntax might be a bit simpler, but *conceptually* it's exactly the same. *But* now I have to additionally at least:
 3) Add a macro in a completely different part of the source
 4) Add fields in  completely different part of the source
 5) Take care that things are loaded / synced
 6) Write new tests

- As a *user* the whole thing put me into a *significantly* worse place. So again it is just fair to ask: What do we gain from the complexity that is dropped onto me as a user. So far, nobody could tell me.

Thank you.

janiversen:
Before responding to you, I think you need to take a good look at the old settingsObjectWrapper and how many problems it had. Just one example it too had a sync() but only synced a fraction of the used properties. Did you ever have a look at how signals work in qml, fact is they do not.

I do not understand your comments about users (unless you confuse users and developers). they do not see what gets called where, they only get frustrated when e.g. their properties are not kept between sessions. I do not see or understand how the user can be in a worse place (apart from the bugs we have seen). If you want to use QSetting directly there are absolutely nothing I can or will do against it, for me e.g. lastDir is a good example to have it standardized and tested!

I highly agree with you that pref needs to be reworked, so that we do not have the current triangle mess...it is implemented and works nicely on my mac, and this PR is part of the road (admitted a forwarded part because @janmulder reported a problem). 

I know where I am heading with qPref, in the same direction I have advertized since the start. I have NEVER written anything about reinventing QSettings that would be lost time, but making our code stable, tested and consistent (e.q. ensuring properties are saved to disk) are a major concern to me.

You are entitled to ask, but I think it is simply to easy to say before qPref everything worked as a charm. I believe qPref is a major step on a road to making the whole system more stable, it gives a lot of advantages like e.g. having lastDir available in QML and having a consistent load/store of properties.

Please remember things take time and commits, I could have shortcutted several times, but comments etc have constantly caused (mostly very good) changes. I have frequently asked @dirkhh if he agree with the direction, because I do not want to force anything on the project.

I see your comment as an open question, but I feel I am the wrong person to respond, let others decide what is the best road for the project.

dirkhh:
It's a very typical idiom to call the developer working ontop of a library or a subsystem the "user" of that library or subsystem. And it seems pretty clear that that's who @bstoeger is talking about.

My understanding of these changes was from the beginning that they weren't about simplifying the handling of QSettings. They were (and are) about the fact that we have three worlds we are trying to get together. C code, C++ code, QML. @janiversen's analysis was that in order to be able to access our preferences in a sane way from QML, we needed to encapsulate them differently. And his main goal from the start has not been to spent three months tediously reinventing the wheel (like that's fun for anyone), but to get to the point where he feels he has the infrastructure available to work on significantly expanding the functionality of the mobile UI. So far we have of course only seen tiny glimpses of this, but that's the crux with things like this, you need to build the foundation before you can build the house.

I have been supportive of these changes because I really want to see that second step.
Do I know if this is the best possible implementation, the best path to that goal? Nope.
Do I feel that I understand the code and can find issues if needed? Yep - as I've already shown.
Could this be done differently with templates and C++20 extensions to who knows what? I'm sure. Do I prefer that? I'm not so sure.

There is one major concern that Berthold raises that I think I agree with, though.

> 3. Add a macro in a completely different part of the source
> 4. Add fields in completely different part of the source
> 5. Take care that things are loaded / synced
> 6. Write new tests

I do worry about the added complexity for the developer here. There is of course a tradeoff (there always is). The central place where all this is managed makes it more obvious when new preferences and persistent data are added. The sprawl of them all over the code is a code quality concern. And conceptually that is a disadvantage.

But I'm not sure I'm willing to agree that any of this proves that "before this change, everything was better".

So to answer Berthold's question: yes, I definitely think it's worth to take a breath and ask if this is an improvement. I worry that the analysis focuses too much on the **what** and **how** instead of the **why**.
The one important question here to me is not "can we deal with the C/C++ issue better?", but can we deal with the C/C++/QML issue better?". And is it worth re-inventing the wheel for the fifth (I think) time to do this differently.

janiversen:
You talk about QSettings as if it solves everything, it is merely functions to get/set properties to/from disk. In order to e.g. load at start time, you need to add a function (or having every single piece of software load when it sees fit), same goes for sync().

Please bear over with me, but QSettings is only the wheel of a car, it takes more to build a whole car.

Regarding test, I highly disagree that code should be written so that tests are not needed. Tests are needed, if nothing else to ensure when changing anything the function stays identical. Thinking of the amazing you do with divelist I would have thought you would love and demand test cases.  With distributed use of QSettings there are even more demand for tests, to ensure all the corner cases.

Please help me, and show me how we can avoid writing a sync(), QSettings do not help in any way here. Remember sync() is there to ensure struct pref, get written to disk, properties who are guarded, are stored (using setValue) when the setter is called. And just for reference sync() was part of SettingsObjectWrapper as well (although missing many variables).

bstoeger:
> Could this be done differently with templates and C++20 extensions to who knows what? I'm sure. Do I prefer that? I'm not so sure.

Heh - talk about snark (thanks for teaching me that word).

I'm NOT advocating anything of that. On the contrary, I'm advocating removing all the unneeded complexity. Look at the absolutely *horrible* macro-mess we have now. You can't tell me that this is easier to read than a simple template (which is nothing but a macro). And we don't even need templates. If you prefer, write a get() and a set() function for every data-type (how many are there - int, double, bool, string?) if that makes it easier for you.

So far I still have not heard any convincing argument why we would need the complexity. :(

janiversen:
@bstoeger you have convinced me, looking forward to see how you will solve the triangle of
- direct access to struct pref (and ensuring consistency)
- QML access
- well behaving code using setter/getter functions.
qPref is not a wonder, it is one path to solve the above problem set. Other developers would most likely do it differently, but there are one big difference, I am doing it and offering it, while others talk.

I will let this PR sit, while I work on other parts.

bstoeger:
> Regarding test, I highly disagree that code should be written so that tests are not needed. 

The best code is no code, because it is guaranteed to be bug free and doesn't need a test.

> You talk about QSettings as if it solves everything, it is merely functions to get/set properties to/from disk.

That's *exactly* what we want. Perfect tool for the job. If you dislike the interface put a wrapper around it.

> In order to e.g. load at start time, you need to add a function (or having every single piece of software load when it sees fit), same goes for sync().

The only reason we need to manually load at start-time is `struct prefs`. Therefore don't add things to it and we don't have to bother. We don't have to sync, because (quote Qt manual):

```
This function is called automatically from QSettings's destructor and by the event loop at regular intervals, so you normally don't need to call it yourself.
```

I think the whole thing is *conceptually* much simpler than it seems (devil lies of course in the details):

1) Properties mirrored in `struct prefs`:
  - load at startup into prefs
  - setter-function: write into prefs and QSettings
  - getter-function: read from prefs
  - prefs is *read only*. setting *must* be called via function.
2) Properties not mirrored in `struct prefs`
  - general setter-function (one per type): write into QSettings
  - general getter-function (one per type): read from QSettings
3) QML access:
  - simply call the above

How could a setter and getter from 1) look like? For example like this:

```
set_preferences_int(PREFERENCES_KEY, 25)
int i = get_preferences_int(PREFERENCES_KEY)
```
Note that this can be called from C!
Internally, this does a lookup in a table indexed with PREFERENCES KEY, which gives a) a type, b) a pointer into struct prefs, c) a string for QSettings and d) a pointer to a signal.

Super-easy! Just one table to be modified. Only a small set of functions to be tested. Using this table, even the loading could be automated.

We could then slowly move 1) -> 2). After all C can call into C++. And perhaps get rid of `struct prefs` and the loading at program start.

So what do you think?

janiversen:
So easy when leaving out the details, of course struct prefs are not ReadOnly by far, and the table you talk about, wonder how QML will handle that since in qml you write:
     <property> = <whatever>
and the Q_PROPERTY translates that to a set_<property>(value) call, no indexes etc.

What happens in this case:
     struct prefs.abc = 116
    terminate application.
I do not think the destructor can detect that...and that is the problem, the current implementation level allows struct prefs to be writeable, the goal removes struct prefs, since it is not needed (use get instead, which due to inline is just as quick).

You also need to consider the Q_PROPERTY statements, and the signal functions.

So when you write a simple table, which is not a bad idea, that needs a parser to generate the needed functions, and of course handling all the exceptions. That was actually the idea of @tcanabrava if I remember correct, but nobody have done it.

If you only talk about the inside of the setter/getter functions, they are already covered with 2 standard functions.

bstoeger:
> of course struct prefs are not ReadOnly by far

Make it `const`, see what fails to compile and fix it. For those parts that are not easily fixable, introduce a `get_prefs_mutable()` function so that all these places are clearly marked.

> You also need to consider the Q_PROPERTY statements, and the signal functions.

So far I refused to look into that, because Qt's "meta-object" system is the worst mess I've seen in a long time. But do we actually need that macro? Is this the *only* way to register properties? Can't it be done programmatically? I mean the whole QML thing is a total disaster when it comes to *not* catching missing properties at runtime - can't we use that for our purposes?

> So when you write a simple table, which is not a bad idea, that needs a parser to generate the needed functions

That is precisely what I do *not* want to do. I'm so fed up with all that MOC stuff. I really hope Qt doesn't force us in that direction.

> ahh forgot to write you also need to consider how your table makes
emit propertyName Changed(value);

Trivial: function pointer.

BTW: For probably 99% of the properties we do *not* need a signal. Why send a signal for a property like `lastPath`. Make no sense what-so-ever. This should be opt-in.

bstoeger:
> BTW: For probably 99% of the properties we do not need a signal. Why send a signal for a property like lastPath. Make no sense what-so-ever. This should be opt-in.

So I wonder if a classical callback would not make much more sense in this particular case. The only disadvantage of a callback would be that it's not automatically transported across thread boundaries. But even that can be solved with that invoke - stuff.

janiversen:
I am sorry to say it, but you need to get deeply involved in both Q_PROPERTY and MOC to solve the triangle. QSetting is basically the easy part because it just works.

Making structure pref const, was basically what I have already done, but submitting those changes to the existing SettingsObjectWrapper was too insecure. I needed to get a basic qPref implementation first (which is now merged), and then do the optimisation...at least that is how I work (step by step, and secure every step works before taking the next).

Function pointer in an emit will not work, because MOC translates it wrongly.

You do not need to use the Q_PROPERTY macro, it is just a simple way of hiding the actual meta calls.

janiversen:
Well sending a signal makes a lot of sense to QML, since that is the way it detects changes.

A classical callback could surely do a lot of the work, apart from QML and apart from the fact, that is not how Qt operates, signal/connect is one of the backbones in Qt. As I wrote earlier I have chosen one path of many. I am currently not in a situation, where I would like to rewrite the whole stuff (make tables etc), but if there are others willing to invest that time, I have no hard feelings.

I think the project need to focus and instead of suggesting potential better solutions, decide if qPref should be completed (with the priorities as I wrote, and reviews as we go along), or it is time to consider another path (e.g. your tables). I for one have chosen, I do no longer think in tables and callbacks, I think in completing the current implementation.

bstoeger:
> I am sorry to say it, but you need to get deeply involved in both Q_PROPERTY and MOC to solve the triangle.

According to the docs you *can* access "dynamic properties" from QML. Not sure how that works with signals, but I would be surprised if the Qt developers would not have thought about that.

> I think in completing the current implementation.

Fair enough. I'd still ask you to reconsider all the complexities like e.g. the syncing.

janiversen:
dynamic properties, needs to be declared and made public in the Qt meta system, that is what Q_PROPERTY is about.

Again, we are NOT syncing guarded variables, only variables from struct prefs, where we do not control how/when they are changed. You actually commented on that in another PR.

You can find a construct
foo::loadSync(bool doSync)
{
     disk_abc(doSync) 
    ....
    if (!doSync)
      load_xyz()

disk_* is for struct prefs variables (both load and sync). load_* is for guarded variables (no sync).

bstoeger:
> dynamic properties, needs to be declared and made public in the Qt meta system, that is what Q_PROPERTY is about.

This is in contradiction to how I read the docs:

> QObject::setProperty() can also be used to add new properties to an instance of a class at runtime. When it is called with a name and a value, if a property with the given name exists in the QObject, and if the given value is compatible with the property's type, the value is stored in the property, and true is returned. If the value is not compatible with the property's type, the property is not changed, and false is returned. But if the property with the given name doesn't exist in the QObject (i.e., if it wasn't declared with Q_PROPERTY()), a new property with the given name and value is automatically added to the QObject, but false is still returned. This means that a return of false can't be used to determine whether a particular property was actually set, unless you know in advance that the property already exists in the QObject.

> Again, we are NOT syncing guarded variables, only variables from struct prefs, where we do not control how/when they are changed. You actually commented on that in another PR.

How many of these are there? The only ones I know are the `git_local_only`-style IO stuff for which I have a (bit-rotted) patch to remove them, as this is clearly a misuse of the preferences thing.

Then there's the fetch preferences from git stuff. But apart from that, it can't be too much, no?

janiversen:
as I wrote earlier Q_PROPERTY is a macro that can be replaced by meta calls, setProperty is one of them...there are 4-5 different calls to choose from.

But e.g. to use setProperty you need a pointer to the qml engine, something that can be quite funny to obtain especially in the test framework.

I do not know, what you want to say, about there are only a few.....I am working towards making the properties guarded and removed from prefs, isn't that enough.

bstoeger:
> as I wrote earlier Q_PROPERTY is a macro that can be replaced by meta calls, setProperty is one of them...there are 4-5 different calls to choose from.

These are different beasts, please read the docs:
http://doc.qt.io/qt-5/properties.html

`Q_PROPERTY` associates properties with *types* at compile time.
`setProperty()` can modify these properties, but with a twist: If no such property exists it will generate what Qt calls a "dynamic property", which is associated with an *object*.
This means that we proably do *not* have to generate the Q_PROPERTY, signals etc. stuff for QML. We can do it at runtime.

I'm trying to find a non-disruptive way out of the current complexity. Replacing Q_PROPERTY by dynamic properties could be very helpful in this regard.

> But e.g. to use setProperty you need a pointer to the qml engine, something that can be quite funny to obtain especially in the test framework.

Why? I want to `setProperty` on the preferences objects, this has nothing to do with QML.

> I do not know, what you want to say, about there are only a few.....I am working towards making the properties guarded and removed from prefs, isn't that enough.

Well, you said we need the sync-complexity because there is uncontrolled writing to the `prefs` struct. Then it is a legitimate questions to ask how much of that there is. Note that it's only the *writing* that is the problem. We don't have to get rid of the prefs struct, just the writing. If there is not much [or irrelevant like the git_local_only stuff] this seems like a rather trivial thing to do.

dirkhh:
Lively discussion... first things first:
> > Could this be done differently with templates and C++20 extensions to who knows what? I'm sure. Do I prefer that? I'm not so sure.
>
> Heh - talk about snark (thanks for teaching me that word).

Yes. Definitely snark. Thanks for calling me out for that. I could claim that it wasn't aimed at you, but that would be maybe not 100% honest... you have puzzled me quite a few times with your code :-)
___

With that out of the way, please let me be _patronizing_ and _annoying_ and thank both of you for a truly constructive and productive discussion. This was very interesting to read and I think it's moving the mutual understanding forward - which is what I am usually looking for in situations like this.

I will not try to summarize where I think we are in this process, because I'll get it wrong for sure.

Having read everything so far and understood (I think) most of is, let me instead ask a few questions... is it fair to say that we'd be a lot closer to a solution that everyone is happy with if we

1. make prefs read only (i.e. const) and clean up all cases where we change prefs from C code to use a setter instead
2. find a clean way to in QML a) access "persistent data" (the preferences, settings, an other random things we have in these structures), b) modify them, and c) be notified when they are modified by other code
3. umm, is there a third thing we need? I'm sure I'm missing something here

Jan's current code is a big step in the right direction (in my humble opinion), but I truly understand and appreciate Berthold's desire to reduce complexity, simplify the code, and make this easier. What I don't know is if this is possible. So, gentlemen, let's please continue this discussion to the point where we either agree to disagree (and then I get the shit job, err, honorable task of picking a "winner"), or where we find a solution that we think gets this done.

And again, thank you.

janiversen:
Well if you say they are different beasts, then I suppose they are....there would not be different calls if they did exact the sane thing. But you still forget qPref with all the Q_PROPERTY is merged, tested and done. The speculatins do not really help the project decide how to continue. If you volunteer to do it, it would be a fair alternative to completing my work. I do not volunteer to redo that work, I want to complete qPref, put it aside and get on buildung on this fundament.

I am totally aware that the project might decide to wait for someone to come around and reimplement my work based on your ideas, that is how projects work.....and in the meantime I will hopefully have the mobile version I want.

janiversen:
I think enough time have been  spent on this patch. @janmulder have the patch, so no need to keep it.

@dirkhh I am not removing the branch for a day or so, should you be interested in merging this

janiversen:
see comment in #1637.

Please continue discussions, even though the issue is closed the root cause remains.

@dirkhh closing this is not due to me being angry or dimilar, it is simply to reduce my rebase work, since you indicated it will be s while until you look at these PR.

dirkhh:
@janiversen I am taking this manually from GitHub's cached history.
Since this conflicts with changes that I made, it's easier if I just do it directly and that way you don't have the extra work of rebasing.
Thanks

janiversen:
@dirkhh be my guest, nice to see that my work after all is not that bad :-)

I have a nag about cleaning up, and in truth sometimes I do it too extensively :-( so now on my local repo I rename the branch to 1OLD_* (the 1 to put it top of the list, since git does not have something like an archive). Renaming a branch that is in a PR never worked for me.

PR #1629, opened 2018-09-02
DO NOT MERGE, this is just a test to see if travis likes my changes, when so it will be closed.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


janiversen:
Finally, closing to continue my work offline. Will be submitted when QML testing works as I want.

dirkhh:
Just curious - do you then do functional tests on the build artifacts, or do you just do these PRs in order to get them compile tested...

dirkhh:
"Undocumented QML" demands. Yuck. Is that the `*_changed` vs `*Changed` ? Is that why the signals weren't seen in the QML code?

janiversen:
yup..and also we silly create a qml object (see main.qml) thst of course do not emit the signals. My test versions gets rid of these extra qml objects in main (which influences startup time) and make the C++ object available in qml.

dirkhh:
A couple of typos make this last comment hard to parse. "we still(?) create a QML object..." ?
If this means you are finding a way to improve startup and improve C++/QML integration, then this sounds promising  indeed. Thanks for the explanations.

janiversen:
ups. yes we create qml objects in main.qml (see qmlprefs) which are instanciated from the C++ class and thus have no relation with c++  object which emit the signals.

The instanciation is one type of c++/qml interaction needed to make qml objects like profile and map, but it is very costly.

There are a couple if other ways to identify the C++ object (not class) in qml, which gives a direct connection between the c++ object and the qml code. This cannot be used to e.g. create graphicsl object but are perfect for sharing variables, and very cheap to use.

I am playing with these ways, to find the best solution for our case. It might however be best to clean up strcut prefs first, since I have some interesting effects in the following scenario:
1) qml reads variable e.g to show on screen.
2) C++ changes var in struct prefs (e.g diwnload thread)
3) user changes var in qml
4) qml tries to update var in C++ which is already changed.

Guarded variables will secure correct handling of this.

dirkhh:
BTW: this is how we got into all of these classes with the `::instance()` semantics because we kept ending up with two objects of the same class instead of using the same object from QML and C++.
So it sounds like there are more cases of this.
I'd love to see **and understand** an actual solution to this. So far it all seems like voodoo to me.

And for the scenario in your last comment... why doesn't 2. change what is shown in QML, i.e. why is this a partial one way street? Should a change in C++ be reflected in the QML UI? I thought that was part of the idea here...

In this context, what exactly does "guarded variable" mean?

janiversen:
it should, but struct prefs is not monitored so no signal is emitted, and thus qml (or any other) does not know about the change until the read struct prefs.

A guarded variable is basically a static qPref private variable that can only be accessed through the set/get method, and therefore a signal will be emitted when changing.

Getting a guarded variable is just as cheap as reading struct prefs variables due to the inlined get method. Setting is more costly but stored on disk (cached in QSetting) and thus no sync() is needed when terminating the app.

dirkhh:
Ahh, so this is all about not writing to prefs and instead using setters.
I think all three of us want that desperately. Have you looked at Bertholds PR where he tries to start towards that? #1630

janiversen:
not a bad idea, but I hve basically no idea to do that and how to keep that setup synchronized with the project setup. I am happy to do it, if I can read somewhere how to replicate the project setup and keep it synchronized.

glance-:
@janiversen It's quite simple. Just login to travis and enable building for your repo. We might need to correct a couple of details in our travis build infra so it just works, but its nearly trivial.

janiversen:
in favt it was simple, builds runs now..THANKS a lot.

PR #1630, opened 2018-09-02
Cleanup: constify get_units()

bstoeger:
get_units() returns a pointer to the units struct in the preferences.
Callers should not modify the preferences via this struct, therefore
make the return value point to const.

This is a small step in constifying the global preferences structure.

Signed-off-by: Berthold Stoeger <bstoeger@mail.tuwien.ac.at>

### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

A tiny const-ification cleanup. Commit message says it all.
 
### Changes made:

1) Make `get_units()` return pointer-to-const.
2) Fix fallout

### Release note:

No.
### Documentation change:

No.

PR #1633, opened 2018-09-02
[RFC] Preferences: make prefs structure const

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As discussed in #1626, we might want to identify all writers to `struct preferences pref`. Therefore, make it `const`.

Note 1: This looks very "undefined behavior"-ish to me. Low-level experts, please have a look! A (less undefined) alternative might be something like:

```
struct preferences prefs_mutable;
#define prefs ((const struct preferences)prefs_mutable)
```

Note 2: This contains the commits of #1630

What I learned from this: It turns out that things look much less scary than I had feared. Most of the things (e.g git_local_only) do not belong into prefs and need not be synced. The rest is startup / preferences code, where it is OK. And then a few things that should be converted. The only thing that needs some thought seems to be the "reset to default" feature. But I think it is no problem, since the dialog will write the changed properties anyway on "Apply" or "Save", isn't it?

If these few things are fixed, we don't need any manual syncing without completely doing away with the prefs-structure, no?

### Changes made:

1) Make prefs const.
2) Provide an accessor for non-const access.
3) Fix fallout.

### Related issues:

#1626
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/appdata/subsurface.appdata.xml
+++ b/appdata/subsurface.appdata.xml
@@ -36,7 +36,7 @@
   <url type="help">https://subsurface-divelog.org/documentation/</url>
   <url type="translate">https://www.transifex.com/subsurface/subsurface/</url>
   <releases>
-    <release version="" date="" />
+    <release version="v4.8.1-402-g5623a9ed45de" date="2018-09-08" />
   </releases>
   <developer_name>The Subsurface development team</developer_name>
   <update_contact>subsurface@subsurface-divelog.org</update_contact>
--- a/core/dive.c
+++ b/core/dive.c
@@ -3795,17 +3795,17 @@ void set_informational_units(const char *units)
 
 }
 
-void set_git_prefs(const char *prefs)
+void set_git_prefs(const char *field)
 {
-	if (strstr(prefs, "TANKBAR"))
+	if (strstr(field, "TANKBAR"))
 		git_prefs.tankbar = 1;
-	if (strstr(prefs, "DCCEILING"))
+	if (strstr(field, "DCCEILING"))
 		git_prefs.dcceiling = 1;
-	if (strstr(prefs, "SHOW_SETPOINT"))
+	if (strstr(field, "SHOW_SETPOINT"))
 		git_prefs.show_ccr_setpoint = 1;
-	if (strstr(prefs, "SHOW_SENSORS"))
+	if (strstr(field, "SHOW_SENSORS"))
 		git_prefs.show_ccr_sensors = 1;
-	if (strstr(prefs, "PO2_GRAPH"))
+	if (strstr(field, "PO2_GRAPH"))
 		git_prefs.pp_graphs.po2 = 1;
 }
 
--- a/core/dive.h
+++ b/core/dive.h
@@ -764,7 +764,7 @@ extern void dump_cylinders(struct dive *dive, bool verbose);
 #endif
 
 extern void set_informational_units(const char *units);
-extern void set_git_prefs(const char *prefs);
+extern void set_git_prefs(const char *field);
 
 extern char *get_dive_date_c_string(timestamp_t when);
 extern void update_setpoint_events(const struct dive *dive, struct divecomputer *dc);
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -915,7 +915,7 @@ static struct git_repository *is_remote_git_repository(char *remote, const char 
 			/* grab the part between "protocol://" and "@" as encoded email address
 			 * (that's our username) and move the rest of the URL forward, remembering
 			 * to copy the closing NUL as well */
-			prefs.cloud_storage_email_encoded = strndup(p, at - p);
+			get_prefs_mutable()->cloud_storage_email_encoded = strndup(p, at - p);
 			memmove(p, at + 1, strlen(at + 1) + 1);
 		}
 	}
--- a/core/pref.h
+++ b/core/pref.h
@@ -211,7 +211,29 @@ struct preferences {
 	update_manager_prefs_t update_manager;
 };
 
-extern struct preferences prefs, default_prefs, git_prefs;
+extern struct preferences default_prefs, git_prefs;
+
+/* In general, we want to avoid uncontrolled writing to the
+ * prefs structure. Therefore, prefs is a macro that
+ * casts the underlying structure to const.
+ * Places that nevertheless write to prefs are supposed
+ * to use the "get_prefs_mutable()" function. This allows
+ * us to easily find these places.
+ *
+ * Preferences code, which legitimately writes to prefs,
+ * can define PREFERENCES_SOURCE before including this file.
+ * prefs will then be defined as mutable structures.
+ */
+extern struct preferences prefs_internal; /* Don't use directly! */
+#ifdef PREFERENCES_SOURCE
+#	define prefs prefs_internal
+#else
+#	define prefs (*(const struct preferences*)&prefs_internal)
+static inline struct preferences *get_prefs_mutable()
+{
+	return &prefs_internal;
+}
+#endif // PREFERENCES_SOURCE
 
 extern const char *system_divelist_default_font;
 extern double system_divelist_default_font_size;
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -470,7 +470,7 @@ QString uiLanguage(QLocale *callerLoc)
 	else
 		uiLang = languages[0];
 
-	prefs.locale.lang_locale = copy_qstring(uiLang);
+	get_prefs_mutable()->locale.lang_locale = copy_qstring(uiLang);
 
 	// there's a stupid Qt bug on MacOS where uiLanguages doesn't give us the country info
 	if (!uiLang.contains('-') && uiLang != loc.bcp47Name()) {
@@ -500,11 +500,11 @@ QString uiLanguage(QLocale *callerLoc)
 		dateFormat.replace("'en' 'den' d:'e'", " d");
 		if (!prefs.date_format_override || empty_string(prefs.date_format)) {
 			free((void *)prefs.date_format);
-			prefs.date_format = copy_qstring(dateFormat);
+			get_prefs_mutable()->date_format = copy_qstring(dateFormat);
 		}
 		if (!prefs.date_format_override || empty_string(prefs.date_format_short)) {
 			free((void *)prefs.date_format_short);
-			prefs.date_format_short = copy_qstring(shortDateFormat);
+			get_prefs_mutable()->date_format_short = copy_qstring(shortDateFormat);
 		}
 	}
 	if (!prefs.time_format_override || empty_string(prefs.time_format)) {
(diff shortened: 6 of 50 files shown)


bstoeger:
I tried a different version and I think that I am more happy about it:
```
extern struct preferences prefs_internal; /* Don't use directly! */
#define prefs (*(const struct preferences*)&prefs_internal)
static inline struct preferences *get_prefs_mutable()
{
        return &prefs_internal;
}
```

janiversen:
I think this is the wrong way to go. It complicates matters without no benefit. The better way is to remove the property variables from struct preferences. Also which benefits does this bring invthe life system, I did a much less intrusive search for non readonly variables.

btw. all qPref test fails not only 3.

A quick look shows it breaks a couple of concepts in the qpref macros. And more importantly breaks totally with my stream of patches. This seems very much like a split from the intentions of qPref, which of course might not be a bad idea, but I think it is important it is said so its not forgotten.

bstoeger:
Yes, I have no idea what is going on here:
```
QElfParser: '/home/travis/build/Subsurface-divelog/subsurface/build/tests/testmerge47+48.ssrf' is not an ELF object
```
I beg your pardon?

In any case, I pushed a bit of a different flavor:
- Preferences code, which legitimately modifies `struct preferences prefs`, defines `PREFERENCES_SOURCE`. `prefs` then becomes mutable.
- All other code has to use the `get_prefs_mutable()` function, which should drive home the message.

To repeat my observation: I think we are very close to "prefs-const-clean", so with a few small modifications, we probably don't really need that sync stuff at all. :)

bstoeger:
> The better way is to remove the property variables from struct preferences.

Please explain. I don't see the problem in having C read from `prefs`. It always worked like that and as long as there isn't write access it seems like a valid interface.

One idea would be to mark those properties that are exported to C. Only for them write to the prefs-structure.

By the way: I'm not sure what "property variables" are. I assume you mean those variables that happen to be also accessible from QML?

> Also which benefits does this bring invthe life system, I did a much less intrusive search for non readonly variables.

1) It makes it very clear that you shouldn't write to `prefs`.
2) We can do a trivial `git grep get_prefs_mutable` to find the few remaining cases.
3) Your explanation of the whole sync complexity was that "we need it because of writing to prefs". Now this commit pretty much proves that this is not the case.

Sounds like a good idea to me. And especially *significantly* less intrusive than removing all of `prefs`, which to me has a bit of the "change for the sake of change" flavor.

bstoeger:
> In any case, I pushed a bit of a different flavor

Which seems to make tests work..? Beats me.

bstoeger:
> There is really not a lot to explain, I tried to write my opinion in a very neutral way, so really no need to beg your pardon.

The only way that I can interpret this statement is that you misread things. My expression of surprise was with respect to the quoted error message. Indeed, the whole posting you seem to refer to was not addressed at you _at all_.

> you use git_local_only as the example, it seems however you did not analyze the code, git_local_only and cloud_status:CS_NO_CLOUD compete and the correct action is to replace git_local_only.

You are very wrong on this one. Matter of fact, as I stated before, many moons ago I wrote a patch to remove `git_local_only`. This is IO-internal state and should not be global at all, neither in prefs, nor static nor anywhere else. The correct fix here is to introduce IO-state (make the IO-machinery reentrant!). In the meanwhile we can isolate it.

> Is it really logical that
pref.zyx = 17
is legal in some sources and others not.

Yes, why not? The perferences-code can by definition write to `prefs`, the rest of the code should not. Makes perfect sense and is commensurate with idiomatic-C++'s use of `public`, `protected`, `private`. C doesn't have these concepts, so we simply invent them. I think that's the C-way of doing things. If you have better ideas how to transport public/private to C, I'm eager to hear them.

> This PR does not adapt the test cases, which also serve as examples of how to use the code, this is at the very least confusing.

This statement seems contradictory to me, as we seem to agree that user-code shouldn't write to `prefs` directly, yet the tests have to. I would suggest putting examples in the documentation and let tests be tests.

> now we store a bunch of consts to disk when the program exists, sounds wrong to me.

This statement makes no sense to me, care to elaborate? Isn't the whole point of the exercise to *not* have to manually sync at program exit?

Let me share my view of things and please let me know where you think that I am wrong.

1) You introduced lots of complexity and this gives problems [this is not an accusation - it's just a matter of fact that complexity gives bugs].
2) I'm sure that we agree that removing complexity is the goal.
3) A lot of complexity, notably syncing, can be effectively removed by making `prefs` read-only or noting that the few odd cases don't need to by synced anyway. At least that's what I understood from the previous discussion.
4) You argue that we should wait with removing complexity until `prefs` is basically completely removed (<- this is the point I'm not sure I get right)
5) Removing prefs is an *extremely* invasive change to the core code.
6) So far there has been no argument brought forward what this very invasive change buys us. Especially I don't see how this makes things less complex for the users (as in callers) of the code. I really think that the writer of a subsystem should always keep this perspective in mind. And it doesn't get much simpler than `prefs.language`.

So I cannot really follow point 4), but perhaps that's not even what you are suggesting. Would it not be the best way forward to start removing complexity, let things settle and then think about whether we really want to remove `prefs`? Removing prefs seems so massive that I'd really like to evaluate other options first. Note that this doesn't mean that the rest of the work is bad. It's just that my feeling is that the path you follow doesn't exactly give us what it promises (stability).

janiversen:
@dirkhh let us make life easy, I recommend merging this PR, since it will effectively end the qPref discussion.

I am trying to find ways to change parts of this PR so that it ruins as little as possible in my prepared work, but it will take a while until I am able to merge master (with this PR) into my other branches.

#1626 will help to eliminate big part of syncing const variables (but really it should be fixed in the code). sync() and load() are still called when stopping/starting the app, and currently all defined variables are read from/written to disk.

bstoeger:
Sorry for being such a nuisance, but I think it is fair and necessary to ask questions about your design. It seems to me that I haven't yet fully understood what's going on.

> I am trying to find ways to change parts of this PR so that it ruins as little as possible in my prepared work, but it will take a while until I am able to merge master (with this PR) into my other branches.

But why is that? The only thing this does is mark those places that write to prefs. You wanted to remove these anyway, so it seems this makes that particular part of the task *easier*. Nothing a simple `git rebase` wouldn't solve, no?

> #1626 will help to eliminate big part of syncing const variables (but really it should be fixed in the code). sync() and load() are still called when stopping/starting the app, and currently all defined variables are read from/written to disk.

Now, wait. My understanding of the discussion in #1626 was that sync-on-exit was only necessary owing to uncontrolled writing to prefs. Now that we identified all occurrences and can hopefully decide whether they need sync-on-exit, wouldn't it be best to *remove* general sync-on-exit. Only do this for the handful of (very) odd cases?

janiversen:
your PR does not remove the sync() call, nor adapt the DISK_* macros, therefore variables will be written as before your PR, so #1636 will help. Of course the ideal solution is you solve the sync problem in the PR.

The only thing this PR does is NOT just mark those places...it e.g. also changes the include sequence in qPref*. A simple git rebase, brings me nothing but tons of conflicts, and I am 99% the end result will not work (due to the pointer introduction, as I manipulate prefs in a later commit).....but we do talk about my private branch, so that is all my problem and not a project problem.

dirkhh:
> git_local_only and cloud_status:CS_NO_CLOUD compete and the correct action is to replace git_local_only. Why did I not do that? simply because my first step was to remove settingsobjectwrapper in a 1-1 operation, so I could do a before/after compare.

Err, no. `cloud_status:CS_NO_CLOUD` and `prefs.git_local_only` do not "compete". They serve complementing purposes. One is the persistent setting defining whether we should contact the cloud server, the other is a temporary override that we use in certain situations to ensure that the `libgit2` code doesn't access the network when we want it to operate only on the local cached copy.

janiversen:
heh then it sounds real logical that git_local_only is used in qmlmanager.cpp. NO_CLOUD also signals not to use the network, but work on the local copy (allthough not temporary).

But who cares, that is not the problem at hand. I made my recommendation, which seems identical to that of @bstoeger (sorry if I interpreted wrong), so once this is done, @bstoeger can continue his divelist work, and I with my experiments.

dirkhh:
> heh then it sounds real logical that git_local_only is used in qmlmanager.cpp. NO_CLOUD also signals not to use the network, but work on the local copy (allthough not temporary).

Actually, that makes a lot of sense. One signifies the *current* state on whether you should contact the remote server, the other signifies the *intended, persistent* state. So depending on what you are trying to do, either may be the right one to look at.

dirkhh:
So this has some good parts and some bad parts.
What I don't like is the `get_prefs_mutable` - instead I'd prefer a setter function for the few members that are being set from C/C++ code. Because then we can make sure that those call send the correct signals, etc.
Wouldn't that make a lot more sense?

bstoeger:
The point of this patch is *not* to fix anything. It is of a *documentary* value. A simple `git grep get_prefs_mutable` will show all potentially problematic uses.

I really dislike the "blindly convert everything" approach. The most obvious example is `git_local_only`, but there may be others. There seem to be so few cases that there is no excuse for not checking each case individually. The one big offender is `qmlmanager.cpp`.

janiversen:
just documenting the cases, seems to leave a big possibility that it will never, or at least not for a long time, be changed., SettingsObjectWrapper seems to have used the same approach of being incomplete in some parts, but at least what was done was done completely.

In order to avoid misunderstanings, I am finished with qPref (apart from bug solving) when my current PR is completed, so do not count on me to do the setter implementation. btw git_local_only was not done blindly bur in purpose, because my goal was to first make a 1 to 1 conversion, making notes while doing it (which includes the list of who writes to prefs.) and then make changes, So please do not call it blindly, it was just another way to the same goal.

bstoeger:
Removing the problematic cases should proceed in three steps:
1) Identify and document.
2) Think about each individual case (is syncing even necessary, should it even be in prefs?).
3) Fix the cases *for which it is necessary*.

Not doing 1) because it doesn't do 3) is counter-productive.

janiversen:
and please do not forget:
4. remove DISK_* macros in qPrefPrivate*
5. correct SET_* macros to write to disk (using the qPrefPrivate function)
6. remove all loadSync() and fill in correct calls in all load() functions
7. update all qPref cpp/h sources to reflect to above changes.
8. update test cases to not write to prefs
9. consider all the places where qPref"class"::"var"() is used to reference a variable
   (since this was the alternative to make prefs. constant, meaning it is 2 ways of doing 
    the same, which is unclean)

This is a rather big editing task, that goes hand in hand with your 3 points.

dirkhh:
That's a nice list of tasks.
I'll start with two that I know shouldn't be in `prefs`: `git_local` and `newpassword`
That should hopefully give us an idea how this process should be tackled moving forward.
I have to drive my kids to school now, I'll stare at this some more when I'm back at my desk :-)

bstoeger:
Rebased on master. Removing `git_local_only` is a big step forward, though I would really like to make the IO-thing reentrant.

> What I don't like is the get_prefs_mutable - instead I'd prefer a setter function for the few members that are being set from C/C++ code.

I fear that we will have to live with something like this for some time. See the funky things that `desktop-widgets/preferences/preferencesdialog.cpp` does. And I still think that we should first mark those places clearly with `get_prefs_mutable()` or other scheme to your liking (any suggestions?) and then work on each case individually.

If something doesn't need syncing, then writing to prefs is not ideal, but not a tragedy either.

dirkhh:
I think we can close this one as well.

PR #1635, opened 2018-09-03
win32: fix writing of logs for non-admin users

neolit123:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

if the user is running the application on Windows, not from a terminal, but from the desktop, the app will write log files. but if the user does not have privileges to write to the Subsurface install path the log files will not be written.

these patches adjust `windows.c` so that the files are written in the user path.

### Changes made:

1) add `utf16_to_utf8()`
1) make `system_default_path_append()` use `wchar_t` only.
1) write logs to user path

### Related issues:

NONE

### Additional information:

https://groups.google.com/d/msgid/subsurface-divelog/7f1b59ec-8b39-4091-be08-d499a03fd83d%40googlegroups.com?utm_medium=email&utm_source=footer

### Release note:

```
Windows: write log files to the user path instead of the path where Subsurface
  is installed.
```

### Documentation change:

NONE

### Mentions:

@dirkhh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Windows: write log files to the user path instead of the path where Subsurface
+  is installed.
 - Desktop: fix issue with dive list row height in case of larger fonts [#1600]
 - Desktop: fix a bug where it is possible for the user to hide all divelist columns [#1600]
 - Mobile: add default cylinder functionality for new dives [#1535]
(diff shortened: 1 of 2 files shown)


dirkhh (commented):
Thanks - this looks good.
Travis is unhappy, though...

dirkhh:
I get a bit frustrated with all the noise in the Travis results - but then again, it does catch real issues often enough to be worth it, IMHO

dirkhh (approved):

PR #1638, opened 2018-09-05
[TEST] Build system: newer versions for libgit2 and openssl

janmulder:
Maybe this fixes build issues for Qt5.6


janmulder:
Ok, forget this. Does not fix the Qt 5.6 build and the iOs build is broken too. Like a said: a test.

PR #1643, opened 2018-09-06
solve #1642

### LINKED ISSUE #1642: Default file name not backwards compatible
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

It just happened to me that an older and a recent version open different default logs.
Looking into the config file I see both versions of the file:

new:
```
[General]
CloudStoragesubsurface_webservice_uid=XXX
GeneralSettingsdefault_filename=XXX
loadFromCloudXXXXXt=true
subsurface_webservice_uid=XXX
```
Note line 4: The email is in the key - is that on purpose? Looks ugly.

old:
```
[GeneralSettings]
auto_recalculate_thumbnails=true
default_cylinder=AL80
default_file_behavior=1
default_filename=XXX
default_git_branch=master
defaultsetpoint=1100
extract_video_thumbnails=true
extract_video_thumbnails_position=84
ffmpeg_executable=ffmpeg
o2consumption=720
pscr_ratio=100
use_default_file=true
```

I'm marking this as a bug, even though I'm not 100% certain it should be considered as such. It would be borderline OK if old->new settings are taken but new->old breaks. Is that the case?

In any case, I'm not sure that I like the split between `GeneralSettings` and `General` as well as the gratuitous change of the config file format.

### Operating system:

Kubuntu 18.04

### Subsurface version:

current master + local patches.

### Steps to reproduce:

Start old and new versions.
### Current behavior:

Different config-file layouts.
### Expected behavior:

Same config-file layouts.
### Mentions:

@janiversen

janiversen:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

add missing / in default_filename

fixes #1642

### DIFF
--- a/core/settings/qPrefCloudStorage.cpp
+++ b/core/settings/qPrefCloudStorage.cpp
@@ -7,6 +7,7 @@ static const QString group = QStringLiteral("CloudStorage");
 qPrefCloudStorage::qPrefCloudStorage(QObject *parent) : QObject(parent)
 {
 }
+
 qPrefCloudStorage *qPrefCloudStorage::instance()
 {
 	static qPrefCloudStorage *self = new qPrefCloudStorage;
@@ -112,13 +113,4 @@ void qPrefCloudStorage::disk_userid(bool doSync)
 			prefs.userid = copy_qstring(qPrefPrivate::propValue(group + "subsurface_webservice_uid", default_prefs.userid).toString());
 		}
 	}
-}
-
-bool qPrefCloudStorage::loadFromCloud(const QString& email)
-{
-	return qPrefPrivate::propValue(QString("loadFromCloud") + email, false).toBool();
-}
-void qPrefCloudStorage::set_loadFromCloud(const QString& email, bool done)
-{
-	qPrefPrivate::propSetValue(QString("loadFromCloud") + email, done);
 }
--- a/core/settings/qPrefCloudStorage.h
+++ b/core/settings/qPrefCloudStorage.h
@@ -45,8 +45,6 @@ public:
 	static bool save_userid_local() { return prefs.save_userid_local; }
 	static QString userid() { return prefs.userid; }
 
-	static bool loadFromCloud(const QString& email);
-
 public slots:
 	static void set_cloud_base_url(const QString &value);
 	static void set_cloud_storage_email(const QString &value);
@@ -60,8 +58,6 @@ public slots:
 	static void set_save_password_local(bool value);
 	static void set_save_userid_local(bool value);
 	static void set_userid(const QString &value);
-
-	static void set_loadFromCloud(const QString& email, bool done);
 
 signals:
 	void cloud_base_url_changed(const QString &value);
--- a/core/settings/qPrefGeneral.cpp
+++ b/core/settings/qPrefGeneral.cpp
@@ -45,7 +45,7 @@ HANDLE_PREFERENCE_BOOL(General, "/auto_recalculate_thumbnails", auto_recalculate
 
 HANDLE_PREFERENCE_TXT(General, "/default_cylinder", default_cylinder);
 
-HANDLE_PREFERENCE_TXT(General, "default_filename", default_filename);
+HANDLE_PREFERENCE_TXT(General, "/default_filename", default_filename);
 
 
 void qPrefGeneral::set_default_file_behavior(enum def_file_behavior value)
--- a/tests/testqPrefCloudStorage.cpp
+++ b/tests/testqPrefCloudStorage.cpp
@@ -266,18 +266,4 @@ void TestQPrefCloudStorage::test_oldPreferences()
 	TEST(cloud->cloud_verification_status(), 1);
 }
 
-void TestQPrefCloudStorage::test_loadFromCloud_var()
-{
-	auto cloud = qPrefCloudStorage::instance();
-
-	cloud->set_loadFromCloud("mail1", true);
-	cloud->set_loadFromCloud("mail2", false);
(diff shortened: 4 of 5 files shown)


janiversen:
It is 2 different things.

The / (second commit) was a real error, that caused problems with compatibility!

I am not sure there are a problem with loadFromCloud, but @bstoeger does not seem to like it or indicate there are a problem, so better revert it. I cannot really judge if it is the right thing to do, that depends a lot on another discussion, but it does not harm.

I have not done any testing against 4.8.0, all my testing have been against 4.8.1, I wanted to automate that, but somehow never to around to finding the file location of QSettings on my Mac.

janiversen:
If you mail me the file behind QSetting from a Mac for 4.8.0 (preferable so I can see the path) then I will do the compare.

dirkhh:
I have spent 20 minutes on this so far and I am not happy.
We are storing everything in the settings, even values that are unchanged from the default which is wrong and broken.

Clearly in order to solve the overall fight between @janiversen and @bstoeger I need to actually fully understand what we do here and how and decide how I want it to be done. I have taken a much too easy-handed approach to this whole mess.
This is an abject failure of me as maintainer and I'll clean it up.

janiversen:
Interesting, I did not find anything in SettingsObjectWrapper::sync() that compared to default...but on the other hand sync() was not complete.

The different setters in SettingsObjectWrapper only checked that the value had actually changed, before writing it. That is why qPref does (what I thought) was the same.

The problem could be that qPref:sync() do not check with defaults, nor do the single setters. Correcting sync to check with default is something like a 5 minute task.

Regarding the overall fight, yes it does seem I did not get sufficient review (e.g. saving parameters that are at default, which not pr se is broken), but yes I think it is a good idea you decide what you want to have.

janiversen:
When thinking, then please remember qPref::sync() is there to TEMPORARY solve the problem of:
prefs.xyz = 123
I submitted a patch to only store changed values, but that was left hanging.
Please also remark that the patch from @bstoeger does not touch sync() so that is 100% the same problem.

My idea is to replace:
prefs.xyz = 124 with qPref...::set_xyz(124) and remove qPref::sync() totally. I actually presented that earlier, but due to the split in many PRs sync() was invented.

bstoeger:
> Clearly in order to solve the overall fight between @janiversen and @bstoeger I need to actually fully understand what we do here and how and decide how I want it to be done. 

Grrrrrr. I am *not* fighting. I am *not* saying that everything is bad. There are aspects that I dislike, some of them quite a bit [e.g. moving non-prefs things into prefs] and there are parts of the process that I dislike [e.g. super-invasive changes without clear benefits being prioritized over simplifications]. I complain about them and try to come up alternatives and yes, sometimes I probably get a bit too grumpy and impatient when I have the feeling that I'm not being understood. Mea culpa. But for me this is all withing the range of normal discussion and very far from fighting. And I don't see any insurmountable dissent.

For now, please let's make this stable and binary compatible with previous versions. Thank you.

bstoeger:
> Please also remark that the patch from @bstoeger does not touch sync() so that is 100% the same problem.

Of course not. The idea was to *isolate* the odd case, so that we can analyze them [which of them have to be *actually* synced] modify these few to use the proper setters and then remove the `sync()` business.

janiversen:
That makes 2 of us, I do not see a fight....but merely 2 different philosofies, where the correct implementation might be in the middle.

I have very rapidly solved every problem I caused with qPref, but now it is as if everything have to named “fight” and other bad words.....seems to me this is more about making a scapegoat, instead of finding a golden middleway, at least that is how I feel it.

bstoeger:
For me this fixes the problem - it now loads the same file as the old version.

There are Travis mis-compilations which seem to be real problems.

Personally, I think that the QMLManager-properties were perfectly fine there and should not have been moved to the prefs. But I don't care enough to have an argument on that.

janiversen:
Thanks for testing and confirming it works.

As for the qmlmanager part, I thought I had reverted that commit (first commit in here), exactly in order to avoid discussions.

bstoeger:
> As for the qmlmanager part, I thought I had reverted that commit (first commit in here), exactly in order to avoid discussions.

I saw that, but I'm not sure if there are more of those..

dirkhh:
@janiversen - from my reading of the Travis output, I think something is broken with building Subsurface-mobile:
```
[ 70%] Building CXX object CMakeFiles/subsurface-mobile.dir/mobile-widgets/qmlmanager.cpp.o
/workspace/subsurface/mobile-widgets/qmlmanager.cpp: In member function 'void QMLManager::syncLoadFromCloud()':
/workspace/subsurface/mobile-widgets/qmlmanager.cpp:1462:20: error: 'loadFromCloud' is not a member of 'qPrefCloudStorage'
  m_loadFromCloud = qPrefCloudStorage::loadFromCloud(prefs.cloud_storage_email);
                    ^
/workspace/subsurface/mobile-widgets/qmlmanager.cpp: In member function 'void QMLManager::setLoadFromCloud(bool)':
/workspace/subsurface/mobile-widgets/qmlmanager.cpp:1467:2: error: 'set_loadFromCloud' is not a member of 'qPrefCloudStorage'
  qPrefCloudStorage::set_loadFromCloud(prefs.cloud_storage_email, done);
  ^
```
All three failing builds are Subsurface-mobile builds.

janiversen:
yes qmlmanger is only compiled for. mobile. I will have a look tomorrow.

dirkhh:
manually merged

PR #1645, opened 2018-09-06
SmartTrak import tool. Back to arrays.

salvadorcunat:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] Code rework

In PR #1599 comments, @bstoeger , @janiversen  and @dirkhh pointed out the benefits of dinamically growing arrays in front of linked lists. While I fully agree in speed concerns, past bad experiences in reallocating arrays make me very cautious with them (too much probably).
Fortunately there is no need to use them, as we can know in advance how many elements will conform the array.
As a bonus, smartrak tables are all indexed, so we can use the index to access directly the array elements without search. The drawback is the "holes" we can find in the tables which have to be allocated too, to port the table index to the array.

In summary, I keep lists to the relations tables (which have to be traversed in full anyway, and are quite short), and take the rest of tables back to arrays.  The speed for a >400 dives file and some list quite long, improves about a 12% (although full import in CLI mode takes only 2 seconds in my "notsopowerful" laptop).

By the way, I've added some other tables to subsurface's dive notes, which were out till now  (weather, surface and underwater).

### DIFF
--- a/smtk-import/smartrak.c
+++ b/smtk-import/smartrak.c
@@ -902,9 +902,8 @@ void smartrak_import(const char *file, struct dive_table *divetable)
 	MdbHandle *mdb, *mdb_clon;
 	MdbTableDef *mdb_table;
 	MdbColumn *col[MDB_MAX_COLS];
-	char *bound_values[MDB_MAX_COLS], *ver;
+	char *bound_values[MDB_MAX_COLS];
 	int i, dc_model;
-	struct types_list *type_list = NULL, *activity_list = NULL, *gear_list = NULL, *fish_list = NULL, *buddy_list=NULL;
 
 	// Set an european style locale to work date/time conversion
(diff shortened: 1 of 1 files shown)


salvadorcunat:
> I assume you tested this with some of your data files to make sure there aren't any off-by-one errors hidden in there?

Yes.  It has been tested with all smartrak files I've avaliable including fake and corrupt files and works pretty well. I'm quite happy with this version.
There are still some tables not imported in subsurface, but they are related to the dive site and I'm still unsure how to manage them  (placing them in dive site notes seems the best option).
Anyway, I will review the full project, including scripts and CMakeLists before you cut 4.8.2 (thinking of next weekend if Willem and Berthold have to add documentation to user manual).

PR #1646, opened 2018-09-06
Fix appdata and icon

Alexander-Wilms:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a follow-up to #1625, since Github displayed this message and pushing new commit didn't help: "Alexander-Wilms:master was force-pushed and no longer has any new commits. Pushing new commits will allow the pull request to be re-opened."
The appdata file didn't pass validation and on KDE Wayland, the fallback icon was used instead of the actual Subsurface icon.

### Changes made:
1) Fix appdata so that it passes validation using `appstreamcli validate`
2) Use QGuiApplication::setDesktopFileName() to fix a broken icon when using Subsurface on KDE w/ Wayland


core/qt-init.cpp
```diff
@@ -22,7 +22,10 @@ void init_qt_late()
 	QTextCodec::setCodecForLocale(QTextCodec::codecForMib(106));
 
 	QCoreApplication::setOrganizationName("Subsurface");
-	QCoreApplication::setOrganizationDomain("subsurface.hohndel.org");
+	QCoreApplication::setOrganizationDomain("subsurface-divelog.org");
```
dirkhh: while this is the correct domain, I wonder if this will break things when upgrading. I.e., won't that cause the settings from previous versions to no longer be found?

appdata/subsurface.appdata.xml
```diff
@@ -1,34 +1,47 @@
...
+      <image>https://subsurface-divelog.org/screenshots/preferences2.png</image>
+    </screenshot>
+  </screenshots>
+  <url type="homepage">https://subsurface-divelog.org</url>
+  <url type="bugtracker">https://github.com/Subsurface-divelog/subsurface/issues</url>
+  <url type="faq">https://subsurface-divelog.org/faq/</url>
+  <url type="help">https://subsurface-divelog.org/documentation/</url>
+  <url type="translate">https://www.transifex.com/subsurface/subsurface/</url>
+  <releases>
+     <release version="4.8.1" date="2013-07-14" />
```
dirkhh: Hardcoding the release here seems unfortunate... this should be implemented differently.

Alexander-Wilms: Would removing these values and adding this to scripts/ be acceptable?

```
#!/bin/sh
version=$(git describe --abbrev=12)
date=$(git log -1 --format="%at" | xargs -I{} date -d @{} +%Y-%m-%d)
sed -e "s|<release version=\"\" date=\"\" />|<release version=\"$version\" date=\"$date\" />|" -i appdata/subsurface.appdata.xml
```

core/qt-init.cpp
Alexander-Wilms: Right, I've changed it back.

appdata/subsurface.appdata.xml
dirkhh: yes, we have a couple of places where we do similar things for example for other builds

Alexander-Wilms: Okay, done

dirkhh:
merged manually in order to make sure we actually call the script to fix the appdata (and to shorten the subject line of the second commit message).
@Alexander-Wilms please test that this now does what you need ;-)

Alexander-Wilms:
Thanks. Yes, it works :+1:

janmulder:
@dirkhh: a build now results in an updated `subsurface.appdata.xml` so a modified file with `git status`. Obviously, we could gitignore this file, but no sure that's a valid solution.

dirkhh:
Darn, hadn't noticed.
We should change this to the ...xml.in -> .xml pattern...

Issue #1648, opened 2018-09-07
Weird/buggy webservice UUID behavior (part 1)


janmulder:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Some time ago, I forgot my dive computer, so my buddy (also Subsurface user, including using GPS logging using mobile) exported the dive from his Subsurface logbook to complete my logbook. Just verified that his XML export contained his webservice UUID. Not immediately sure this is a bug, but it is al least a privacy leak, as I can now see the logged locations for that webservice UUID. Interestingly, this webservice UUID was the only setting that is exported (and I do not see any reason why we might want to export this). It was likely exported with official version 4.7.8. And just tested with the current master: export does export a webservice UUID. To me, this is a bug. Let call it bug 1).

But it is getting better. As I'm using a local git repo for storage of my logbook, I can see that the imported webservice UUID from the buddy ends up in the persistent and shared preferences (the 00-Subsurface file in the git storage). This import was back in May 2018, using Subsurface:4.7.8.82. Up to now, I've not found a way to reset this value to my own webservice UUID (other then commit a change to my git storage). And obviously, as it are persistent and shared preferences that buddies UUID ends up on my mobile devices as well (verified looking in the git repo from the cloud storage on a rooted Android device, and when syncing GPS data in the app, it shows in the log data as well). One part I do not understand here, is, why is my own webservice UUID never stored in the persistent preferences, but an imported one is? This confuses me big time, and lets call this bug 2)

And yes, we should get rid of that webservice UUID altogether (at lease visible to the user), but that is plan B.

dirkhh:
(1) privacy: agreed. I want that whole web service to go away and simply apply the GPS fixes "on device". I haven't taken the plunge to actually implement that (or "unimplement that"), but I think that's the sane direction. I haven't pushed my GPS fixes to the service in ages.
The reason the XML file contains that is by now bogus, as you can now look up your ID from your Cloud Storage account. But in older versions that wasn't the case and there this saved you from having to type it in on every computer or VM that you used.

(2) that's strange; your own user ID (which you enter in the preferences) should then overwrite your buddies. This sounds like a much more serious bug to me than the privacy issue in (1)

We definitely want to fix both issues for 4.8.2 - and we should discuss on the mailing list the idea to just switch to not using the GPS webservice anymore

janmulder:
See also issue #1013 (yes, an old one, but related).

[...]

PR #1649, opened 2018-09-07
Preferences: more missing /

janmulder:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

subsurface_webservice_uid ends up in the wrong group due missing
/. Notice that this does not fix the issues I mentioned in #1648.

### DIFF
--- a/core/settings/qPrefCloudStorage.cpp
+++ b/core/settings/qPrefCloudStorage.cpp
@@ -103,14 +103,14 @@ void qPrefCloudStorage::disk_userid(bool doSync)
 {
 	if (doSync) {
 		// always save in new position (part of cloud storage group)
-		qPrefPrivate::propSetValue(group + "subsurface_webservice_uid", prefs.userid, default_prefs.userid);
+		qPrefPrivate::propSetValue(group + "/subsurface_webservice_uid", prefs.userid, default_prefs.userid);
 	} else {
 		//WARNING: UserId was  stored outside of any group.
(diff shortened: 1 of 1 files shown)


dirkhh:
Thanks for that. Let me try and grep for other cases where we are missing a `/`

janmulder:
Looking trough my .conf file, I do not see more suspicious looking concatenated preference names, so it would surprise me when you find more.

dirkhh:
I was considering something like this:
```
QString keyFromGroupAndName(QString group, QString name)
{
	QString slash = group.endsWith('/') ? "" : "/";
	return group + slash + name;
}
```
and then replace all the places where we do `group + name` to create the key (basically the dozen macros) with `keyFromGroupAndName(group, name)` which would ensure that typos like that have no negative effect. Trying to find all the call sites turned out to be rather painful and error prone...

@janiversen: thoughts?

janiversen:
no thoughts, that will do the job. You can make it simpler by removing the / in the static group variable in all qPref*.cpp

However doing it your way is more secure. Please do remember to check all qPref*.cpp for the few properties that do not use macros.

dirkhh:
Yes, I was in the process of hunting for those. The idea above was intended as the big hammer to solve the majority of them. I'll create a PR and ask you for review, @janiversen

PR #1650, opened 2018-09-07
qPref: use helper function to ensure key/name grouping

dirkhh:
We had a couple of instances of names being incorrectly merged with their
group, this should handle that better. It's a bit of a big hammer to use, but
it seems to work (and it makes it easy to then git grep for cases that don't
use the new helper function.

This should change all invocations that had the `/` as part of the name to drop that (and I think I found a couple where this was still missing). And it makes it fairly easy to use git grep for situations where we aren't using the helper.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Calling this a cleanup is misleading. The existing code was fine, it just opened the door to typos that would cause group and key to be concatenated without a `/` between them. Now the pattern should make it easier to avoid that (and the helper tries to be smart in case I missed a case where the `/` was already present.

### Changes made:

1) add helper function that ensures there's exactly one `/` between `group` and `name` (that's a lie, if you call it with both `group` ending with `/` and `name` starting with `/`, it doesn't remove one of them...)
2) edit the macros and the users of `qPref` to not include the `/` in the name anymore (this one is more for simplification and consistency than anything else)

### Related issues:

#1649 #1643
### Mentions:

@janmulder @bstoeger @janiversen - could the three of you take a look to make sure this is sane and doesn't break anything? It feels like overkill, but I think it should get the job done.

### DIFF
--- a/core/settings/qPrefPrivate.h
+++ b/core/settings/qPrefPrivate.h
@@ -22,15 +22,17 @@ private:
 	qPrefPrivate() {}
 };
 
+// helper function to ensure there's a '/' between group and name
+QString keyFromGroupAndName(QString group, QString name);
 
 //******* Macros to generate disk function
 #define DISK_LOADSYNC_BOOL_EXT(usegroup, name, field, usestruct) \
 	void qPref##usegroup::disk_##field(bool doSync) \
 	{ \
 		if (doSync) \
-			qPrefPrivate::propSetValue(group + name, prefs.usestruct field, default_prefs.usestruct field); \
+			qPrefPrivate::propSetValue(keyFromGroupAndName(group, name), prefs.usestruct field, default_prefs.usestruct field); \
 		else \
-			prefs.usestruct field = qPrefPrivate::propValue(group + name, default_prefs.usestruct field).toBool(); \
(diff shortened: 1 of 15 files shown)


janmulder (commented):
I checked my .conf file and see no strange concatenations of group/name combinations, so that is good. Not sure that this check guarantees 100% completeness, but overall LGTM.

core/settings/qPrefPrivate.h
```diff
@@ -22,15 +22,17 @@ class qPrefPrivate {
 	qPrefPrivate() {}
 };
 
+// helper function to ensure there's a '/' between group and name
+QString keyFromGroupAndName(QString group, QString name);
```
janiversen, line 26: I add "extern" in front, instead of just being a forward declaration....but that might be a matter of taste.

janiversen:
Just looked at the code (in GitHub, did NOT compile), but there all the changes I expected and the changes look correct.

Added one small comment, but that might be due to my C tendency.

PR #1651, opened 2018-09-07
Merged prefs

dirkhh:
This one combines what's in #1650 with a merged version of #1626 (where I also renamed the static variable)
This is basically to make sure all this compiles, once Travis is happy, I'll merge into master, and then I'll ask @bstoeger to consider updating #1633 on top of that new state (because I think that's also the right thing to do).
And then all of us can see if we can simplify things now that we have all the moving parts merged :-)

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh:
Eh, stop, nono... this doesn't have the edited commit message and has uggly merges... What's going on...

PR #1652, opened 2018-09-07
Merged prefs

dirkhh:
This one combines what's in #1650 with a merged version of #1626 (where I also renamed the static variable)
This is basically to make sure all this compiles, once Travis is happy, I'll merge into master, and then I'll ask @bstoeger to consider updating #1633 on top of that new state (because I think that's also the right thing to do).
And then all of us can see if we can simplify things now that we have all the moving parts merged :-)

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh:
OK, this now looks right.
@janiversen I made the one change you suggested (`extern` in the .h file)
@bstoeger I'd love a critical eye on this to make sure I didn't mess anything up... and then we should look at #1633

bstoeger:
Well, it probably doesn't come as a surprise that I dislike the `static type current_state` stuff and I think this should be removed ASAP (but post release). Repeating myself here: we should leave all the sync business to `QSettings`. The only thing we have to do on top of `QSettings` is read at program start to fill the `prefs` structure.

janiversen:
I really think you need to read more on what QSettings::sync() it is no miracle, it only writes to disk, what have been through QSettings::setvalue(), which only caches the change. the current_state is to detect changes in struct preferences caused by
prefs.xyz = ...
once those are gone, qPref::sync() is obsolete.

so leaving the sync business to QSettings in the current state of the sources would have some negative efftects, like setiings not being saved.

so in total someone needs to replace prefs.xyz = ... with qPref<class>.set_xyz(...) and then not remove the current_state stuff, but the whole qPref::sync(), the DISK* macros and a couple of other things.

bstoeger:
> the current_state is to detect changes in struct preferences caused by
prefs.xyz = ...
once those are gone, qPref::sync() is obsolete.

Actually, as the make-prefs-const commit shows, it is obsolete pretty much right now. It is crazy to add a static (=global) variable for *all* settings if there's not even a handful of direct writers to prefs.

> so in total someone needs to replace prefs.xyz = ... with qPref.set_xyz(...) 

Not in general, no. Some of these should be removed from prefs (e.g. `git_local_only`). Some of them probably don't need to be synced anyway. So the first thing "someone" has to do is to assess the situation and don't act blindly. How many of these items *really* have to be synced? The first step is to generate a list of those. Then we can manually sync *only* these without even having to touch the rest of the code.

I'm not a fan of the "we can't make things simpler without rewriting everything" dichotomy.

janiversen:
You are completely right in all your observations, but there are 2 big differences:

1) the replacement of prefs.xyz = ... is future, at least it has not happened and are not part of your const PR, therefore something is needed.
2) it would be a lot more difficult (a LOT more code) to only save that couple on disk, right now the check is added to the macros, and thus duplicated automatically.

I too would prefer another solution, but due to the lack of decision where do we go from here, I stopped working on qPref, so I did not publish the changes in 1). At lot would have been easier if the direction had been settled a while ago. It seems though as if @dirkhh has made up his mind, which is good, even though it is against what I believe to be correct, this at least enables these matters to be solved.

dirkhh:
@bstoeger you are correct, what I committed is a bit extreme but it solves a real problem that users are reporting today (OK, that at least two developers actually **using** master have reported) - and it is a fix until the root cause has been addressed. Which is why I suggested that we continue working on the const-ification of prefs and elimination of the calls that set prefs directly. `prefs.git_local` should move out of prefs, the others need to be replaced by setters.

I didn't want to gate the 4.8.2 release by this, so I applied the brute force workaround. If we can get this completed in the next few days, great, if not, we'll fix it after 4.8.2.

@janiversen I still have no idea where this will go, so I'm curious to hear which wrong decision you believe I have already made.

Everything I've done with qPref in the last two days was to force a state that allows me to release 4.8.2. Because our user base doesn't care one hoot about the internals, they care about support for new dive computers and improvements to the UI. But I can't subject them to massive regressions, either.

janiversen:
@dirkhh ??? in the same message you wrote the @bstoeger to continue working on const-fication, that is, what I have said too often, not a good direction.

Wanting to know which variables are really being written to is a good idea, but better done in a branch and not in the production code (that is what I did).
Apart from that, I believe continuing to use struct preferences, for variables/properties in qPref is not a good idea, much better to replace the few instances (@bstoeger called that intrusive, but bear in mind his redefinition of pref is exactly as intrusive, just more hidden).

dirkhh:
We can disagree about the implementation, but I think idiomatically the logical way to ensure that pref isn't written to directly in the rest of the code is to make it a const. So in that sense I think @bstoeger is on the right track.

janiversen:
just for the record, I have already replaced many places with 2 cases:
xyz = prefs.var --> xyz = qPref<class>::<var>()
and
prefs.var = xyz --> qPref<class>::set_<var>(xyz)

which does exactly the same, but not hidden. 

I have however no problem that you prefer the hidden method, saves me a lot of future work :-)

bstoeger:
> If we can get this completed in the next few days, great, if not, we'll fix it after 4.8.2.

I voiced my opinion on that above ("post release").

> Wanting to know which variables are really being written to is a good idea, but better done in a branch and not in the production code (that is what I did).

Disagree. Having a separate branch seems like a major disadvantage. Good code is *self documenting* and what could better document "don't write to `prefs`" than making it `const`?

> I have however no problem that you prefer the hidden method

A hard compilation error is not hidden.

> xyz = prefs.var --> xyz = qPref::()

1) The C-compiler will not like the double-colon.
2) I still haven't heard a rationale for this change. Accessing structure-elements is a perfectly idiomatic way of obtaining settings. As opposed to write-access (where we want to avoid diverging `prefs` and `QSettings` at all costs) there is absolutely no need for a function-call. But this change is very invasive and probably hard to undo. Let's wait with this until we got a clear view on the way to proceed. Especially, let's remove complexity *first*, before doing gratuitous "cleanups".

dirkhh:
Maintainer time...
In C code we should allow read access to prefs.xyz but we should replace writes to prefs.xyz with a call to a setter function.
The prefs structure should be a const so that the compiler catches violations of that rule.

bstoeger:
> In C code we should allow read access to prefs.xyz but we should replace writes to prefs.xyz with a call to a setter function.

Only in C or did you mean C *and* C++? If we decide on keeping `prefs` for read-access, I see no reason for introducing a C/C++ inconsistency. Currently, the only reason for the getter-functions is the `Q_PROPERTY` thing. Let's not add an additional dependency on the getters, which might make it harder in the long-term to remove them.

PR #1653, opened 2018-09-09
qPref: add signal testing for all properties

janiversen:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In order to use signal in QML, they need to work in C++, this PR
makes update to ensure all signals are valid and tested.

all signals are tested first in C++ and then also in QML, so that future changes are 
automatically controlled.

Furthermore some cleanup tasks have been added.

This PR is now complete and can be merged, depending on review comments.

### Changes made:

1) Cleanup
commit 1: core/setting: change *_changed to *Changed for the sake of QML.0
commit 3: core/settings: add missing empty line.
commit 7: core/tests: change short to int in pref.h
commit 10: core/settings ... : remove qPref* includes in qPref.h

2) error correction:
commit 2: desktop/profile-widget: update signal from _changed to Changed
commit 8: core: declare cloud_status in qPrefCloudStorage
commit 11: core/settings: correct wrong emit in qPrefFacebook

3) register C++ object instead of C++ class (needed to connect correctly)
commit 4 core: add auto registration to qPref*
commit 5: core,tests: change qml register function
commit 6: tests: change testqml to allow direct register of C++ objects
commit 9: tests: update qml test cases
commit 12: core/settings: change bestmixend to int, to allow qml to work
commit 13: tests: add signal test to call testqPref*cpp


dirkhh:
Wow, that's impressive.
I called it a night and have a family focused Sunday, but I'll try to find time to provide feedback tomorrow.
Thanks!

janiversen:
hmmm seem travis tries to build smtk2ssrf, that seems to be something new. I cannot build it on my Mac (missing some libraries) and it seems smtk2ssrf uses the core lib from build, which of course is build with qml support. 

I have no idea how to add qml support to smtk2ssrf and even less on the implications.

If there is a -D variable I can test on, I am happy to add that to qPref.cpp

janiversen:
alternatively I will make a dummy cpp, ugly but it works.

dirkhh:
The question is what makes more sense... a dummy file, or adding a define to the smtk2ssrf builds that you can use to not include the QML code... I think I'd lean towards the latter.

dirkhh:
There are two scripts that are driving those builds, depending on the OS you are on, `scripts/smtk2ssrf-build.sh` and `packaging/windows/smtk2ssrf-mxe-build.sh`
It should be easy to have them insert a `#define SMTK2SSRF` into every compile and then just trigger on that

dirkhh:
Also, your Qt55 and Qt56 builds are failing - but from a quick look it doesn't look like you are using a function that's too new... the error is on `qmlRegisterUncreatableType` which is definitely in Qt 5.5.
I wonder if maybe the implicit / automatic includes changed after 5.6 and you need to add an extra include in your sources? Not sure, but also something to look into, please.

janiversen:
I will try to hack the smtk scripts, but it will be in blind and with crossed fingers.

I know there are some problem with 5.6,  but to be honest I put that a bit on the backburner to first be sure I have all qml signals solved (just need a lot of editing).

dirkhh:
Yeah, Qt 5.5 and 5.6 are nice to have - and I want to keep them around as long as reasonable.
One thing to try is to build `export-html` (I need to add this to Travis I just realize). That might expose the same issues as `smtk2ssrf` with less dependency issues.

dirkhh:
hehe, never mind. SInce it hadn't been in Travis it's broken right now.
I have a rather busy Sunday starting just about now. I'll look into that when I have time, but that may not happen today

janiversen:
When you do add a #define NO_SSRF_QML, I will test that in smtk scripts, but that might also not happen until tomorrow (qml is killing me right now, I have a seg fault in QMLEngine).

dirkhh:
It always puzzles me that I am able to push into YOUR repository. But on GitHub by creating a PR you appear to give me as the maintainer that ability... anyway, I fixed the smtk2ssrf build issue (and made it possible to build on Mac as well, all you need to do is ensure that you have `glib2` installed which you can do via Homebrew).
This should allow you to focus on the core of your code.

packaging/windows/smtk2ssrf-mxe-build.sh
```diff
@@ -222,6 +222,7 @@ i686-w64-mingw32.shared-cmake \
 	-DPKG_CONFIG_EXECUTABLE="/usr/bin/pkg-config" \
 	-DCMAKE_PREFIX_PATH="$BASEDIR"/mxe/usr/i686-w64-mingw32.shared/qt5 \
 	-DCMAKE_BUILD_TYPE=$RELEASE \
+	-DNO_SSRF_QML=ON \
```
dirkhh: you can drop this commit now - I simply added the missing link libraries which seemed much easier and less intrusive.

janiversen: ok will do that tomorrow.

janiversen:
Thanks, when I make a PR I give maintainer right to change, that is easier.....however thos changes does not enter my local repo, so it can lead to funny results.

janiversen:
@dirkhh all yours now

subsurface-helper.cpp
```diff
@@ -155,21 +155,11 @@ static void register_meta_types()
...
-	REGISTER_TYPE(qPrefGeocoding, "SsrfGeocodingPrefs");
-	REGISTER_TYPE(qPrefLanguage, "SsrfLanguagePrefs");
-	REGISTER_TYPE(qPrefLocationService, "SsrfLocationServicePrefs");
-	REGISTER_TYPE(qPrefPartialPressureGas, "SsrfPartialPressureGasPrefs");
-	REGISTER_TYPE(qPrefProxy, "SsrfProxyPrefs");
-	REGISTER_TYPE(qPrefTechnicalDetails, "SsrfTechnicalDetailsPrefs");
-	REGISTER_TYPE(qPrefUnits, "SsrfUnitPrefs");
-	REGISTER_TYPE(qPrefUpdateManager, "SsrfUpdateManagerPrefs");
+
+	(void)rc;
```
dirkhh: that could have gone in an `#else` below, but it's ok

tests/tst_qPrefUpdateManager.qml
```diff
@@ -30,7 +30,7 @@ TestCase {
 		var x4 = tst.next_check
 		var x4_date = Date.fromLocaleString(Qt.locale(), "01-01-2001", "dd-MM-yyyy")
 		tst.next_check = x4_date
-		compare(tst.next_check,  x4_date)
+//TBD		compare(tst.next_check,  x4_date)
```
dirkhh: things like that make me nervous... is this something that you did intentionally?

core/settings/qPrefGeocoding.h
```diff
@@ -17,9 +17,9 @@ class qPrefGeocoding : public QObject {
 	static qPrefGeocoding *instance();
 
 	// Load/Sync local settings (disk) and struct preference
-	void loadSync(bool doSync);
-	void load() { loadSync(false); }
-	void sync() { loadSync(true); }
+	static void loadSync(bool doSync);
+	static void load() { loadSync(false); }
+	static void sync() { loadSync(true); }
```
dirkhh, line 22: This does make the commit **title** a bit misleading - that one had me expect only whitespace changes.
Not critical, but just something I noticed.

core/settings/qPref.cpp
```diff
@@ -39,28 +38,34 @@ void qPref::loadSync(bool doSync)
 	qPrefUpdateManager::loadSync(doSync);
 }
 
-#define REGISTER_QPREF(useClass, useQML) \
-	rc = qmlRegisterType<useClass>("org.subsurfacedivelog.mobile", 1, 0, useQML); \
-	if (rc < 0) \
-		qWarning() << "ERROR: Cannot register " << useQML << ", QML will not work!!";
```
dirkhh: you introduced this one earlier in the series... oh well.

tests/testqml.cpp
```diff
@@ -1,47 +1,29 @@
...
+
+// main loosely copied from QUICK_TEST_MAIN_WITH_SETUP macro
 int main(int argc, char **argv)
 {
-	// QML testing is not supported in the oldest Qt versions we support
-	// if running with Qt version less than 5.10 then skip test
-#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
+#if QT_VERSION < QT_VERSION_CHECK(5, 11, 0)
+	return 0;
+#else
```
dirkhh, line 12: so now this requires 5.11 ? 
That means it won't run on Travis at all.

core/settings/qPrefGeocoding.h
janiversen: hmmm, this should have been in one of the earlier commits, whitespace should be only whitespace, but sometimes git rebase --continue outtrixes me, let me have a look

tests/testqml.cpp
janiversen: Sorry but using contextProperty requires access to the QML engine, and this is not supported in earlier versions of qmltest. the call "with_setup" is very new.

If we really need to support earlier Qt versions for mobile, then I can copy/modify the Qt sources to do the same...whatever you prefer.

tests/testqPrefDisplay.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include <QDate>
 #include <QTest>
+#include <QSignalSpy>
```
dirkhh, line 10: that seems unrelated to this commit?
in general - a commit like this in the middle of this series is understandable, but adds a lot of noise... and makes it harder to review things because I need to make sure that with all the simple renaming there isn't something else in there that slipped through...

tests/tst_qPref.qml
```diff
@@ -1,17 +1,12 @@
 // SPDX-License-Identifier: GPL-2.0
 import QtQuick 2.6
 import QtTest 1.2
-import org.subsurfacedivelog.mobile 1.0
```
dirkhh, line 4: I'm curious why we no longer need to import org.subsurfacedivelog.mobile... isn't that required?

tests/testqml.cpp
dirkhh: I'll work on making sure that our mobile versions are built with Qt 5.11
Not sure how we'll get this into Travis, but I guess I'll work on that as well.

tests/tst_qPref.qml
janiversen: This is because we refer to C++ object not C++ classes. Objects you refer to directly as an extension of the QML language, classes needs an "envelope" with version number etc, so you can use version 1.0, allthough the newest is 2.0

tests/testqml.cpp
dirkhh: I'm trying hard to have Travis run something really similar to what I use to build the app with on iOS.
While I don't want to lose the old Qt versions, for Android and iOS I am much more willing to push for "latest and greatest". The big challenge was that when 5.11 came out packaging the app on both platforms failed. I am once again trying to get this to work, now that I got some hints from another app  Qt/QML app packager. We'll see how this goes (but Tue-Thu I am traveling with reduced time for Subsurface)

janiversen:
I hope I have addressed all your comments. It has been a little nightmare with rebasing, so yes something might have slipped through.

BUT I have test compiled every commit for both desktop and device, and most commit (with a few exception, because it was transition) I have run check on all commits as well.

The CloudCred bug I found is also solved, and had to do with the registration of classes/objects.

When you register an object you cannot refer to the enums inside the class, that needs a class registration....NOT documented.

I really feel we have a very stable qPref C++ <-> QML interface now (and I got a lot of knowledge updated in the process). The next step is to use this interface and reduce qmlmanager a lot.

I am also trying to make a whole new startup sequence (and settings/change cloud creds), because the current one is heavy.

janiversen:
So back in your hands...and myself back to my birthday party.

dirkhh:
Happy Birthday?
I'll go through this one more time (given that I still want to get 4.8.2 out soon and don't want to introduce breakage) and hopefully merge it tonight my time

janiversen:
Just curious, I had rebased the latest upload on master, so wondering why you needed to do a manual merge ?

and yes, it was my birthday yesterday.

janiversen:
I will look at it a bit later, seems strange to me. The register function should register the map widget.

Real funny, I updated my dive locations last week, with this code, but I changed some #ifdef later.

THANKS for bisecting it, makes life easier. I expect to have a fix ready later today and will reference you in the PR.

I can reproduce it with current master, on desktop NOT mobile. This is clearly my problem, and I will supply a fix during today, sorry for the inconvenience.

janmulder:
Same for me, desktop fails, mobile-on-desktop works, did not check mobile on device.

PR #1655, opened 2018-09-09
Fix broken translations in dive list filter plus simplify strings

sfuchs79:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [x] Language translation

Fix some broken translations in the dive list filter UI by replacing
tr() with gettextFromC::tr
Plus simplify some strings given to translation by separating parts
like ": ".

I'm not 100% sure but I guess the broken translations were introduced here: 1a4e7ad0dd29b9eeac2016933cb86bff260dadbb

I didn't do a real bisect but only a git blame plus a "poor mans" bisect by installing 4.7.5 (still OK) and 4.7.6 (NOK).

### Mentions:

@bstoeger @dirkhh

### DIFF
--- a/desktop-widgets/simplewidgets.cpp
+++ b/desktop-widgets/simplewidgets.cpp
@@ -541,23 +541,23 @@ void FilterBase::hideEvent(QHideEvent *event)
 
 TagFilter::TagFilter(QWidget *parent) : FilterBase(TagFilterModel::instance(), parent)
 {
-	ui.label->setText(tr("Tags: "));
+	ui.label->setText(tr("Tags") + QStringLiteral(": "));
 }
 
 BuddyFilter::BuddyFilter(QWidget *parent) : FilterBase(BuddyFilterModel::instance(), parent)
 {
-	ui.label->setText(tr("Person: "));
+	ui.label->setText(tr("Person") + QStringLiteral(": "));
 	ui.label->setToolTip(tr("Searches for buddies and divemasters"));
 }
 
 LocationFilter::LocationFilter(QWidget *parent) : FilterBase(LocationFilterModel::instance(), parent)
 {
-	ui.label->setText(tr("Location: "));
+	ui.label->setText(tr("Location") + QStringLiteral(": "));
 }
 
 SuitFilter::SuitFilter(QWidget *parent) : FilterBase(SuitsFilterModel::instance(), parent)
 {
-	ui.label->setText(tr("Suits: "));
+	ui.label->setText(tr("Suits") + QStringLiteral(": "));
 }
 
 MultiFilter::MultiFilter(QWidget *parent) : QWidget(parent)
--- a/desktop-widgets/simplewidgets.h
+++ b/desktop-widgets/simplewidgets.h
@@ -167,6 +167,7 @@ public:
 };
 
 class FilterBase : public QWidget {
+	Q_OBJECT
 	void addContextMenuEntry(const QString &s, void (FilterModelBase::*)());
 protected:
 	FilterBase(FilterModelBase *model, QWidget *parent = 0);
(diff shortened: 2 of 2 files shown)


bstoeger:
Thanks for looking into this. If this fixes a problem, I'm fine with it, but I'd really like to understand the problem first. Wouldn't the old code simply mean that the translation-strings are in `FilterModelBase`, `TagFilter`, etc. context? If these are too many contexts, perhaps use `FilterModelBase::tr(...)` everywhere?

As for the `gettextFromC::tr("Person") + QString(": ")` construct, it really doesn't matter, but as a matter of principle I think we should use `QStringLiteral(": ")`. Possibly even a single string literal for the whole translation unit:
```
static const QString whatEverYouWantToCallIt = QStringLiteral(": ");
```

sfuchs79:
>Thanks for looking into this. If this fixes a problem, I'm fine with it, but I'd really like to understand the problem first. Wouldn't the old code simply mean that the translation-strings are in FilterModelBase, TagFilter, etc. context? If these are too many contexts, perhaps use FilterModelBase::tr(...) everywhere?

Honestly I was a little bit in a hurry when I submitted this PR. And honestly it's not really clear to me why the simple tr() doesn't work in this case. But I already a couple of times learned that the whole translation stuff sometimes is a little bit tricky. Maybe someone else can explain this in detail?

Updated to using `QStringLiteral(": ")`

dirkhh:
`FilterBase` is derived from `QWidget`, so `tr()` should just work.
Given how much in Qt is done through cascading classes that inherit from each other, I have a hard time believing that this is what's failing here.
Sure, we can force the gettextFromC route (especially to get this to work for 4.8.2), but I think this deserves a bit more analysis to understand _what_ is going wrong and _why_. Because as far as I can tell there's nothing wrong with the way this is currently implemented.

janmulder:
See and old commit of mine: 604971ca0c83e8. Maybe worth trying to see if it is a case like that.

dirkhh:
That's kinda what I was saying. Since that class wasn't derived from QObject, you needed to mark it. I thought with something derived from QWidget we didn't need that.
But yeah, definitely worth trying before we do the `gettextFromC` thingy

bstoeger:
Oh - I think I see it (I was looking at the wrong file).

`FilterBase`, etc are missing the Q_OBJECT line, isn't it?

janmulder:
When you generate the .ts files in the translations directory (uncomment the last 2 lines in the CMakefile.txt there), you will get warnings for all tr() use without a visible O_OBJECT. Q_DECLARE_TR_FUNCTIONS(class) will likely do the trick.

bstoeger:
> I thought with something derived from QWidget we didn't need that.

The way I understand it: Even if you derive from `QObject`, for working signals, etc. you have to add a `Q_OBJECT` macro, which will also define the `tr()` function. If you don't need signals you can use `Q_DECLARE_TR_FUNCTIONS` instead.

If you don't use either of those, `tr()` will call the version inherited from the parent class and therefore use a different context as the parser.

@sfuchs79: Care to add either of the macros to the function definition in the header file? Personally, I have no preference. Well, maybe I'd go for the lighter `Q_DECLARE_TR_FUNCTIONS`, since I'm not a fan of Qt's macro-shenanigans.

bstoeger:
> If we prefer Q_DECLARE_TR_FUNCTIONS(context) we also need to decide which context to choose.

Then perhaps go the `Q_OBJECT` route to prevent a potential source of errors (misspelled contexts).

> Would it be ok for all to still also simplify the strings by separating the trailing ": "?

Sounds like a good idea to me, but ultimately @dirkhh's call as this will produce new translations.

sfuchs79:
Ok, so this would result in this updated version for the moment... ;-)

sfuchs79:
>When you generate the .ts files in the translations directory (uncomment the last 2 lines in the CMakefile.txt there), you will get warnings for all tr() use without a visible Q_OBJECT. 

Following this comment from @janmulder I have a question to @dirkhh :
Would it make sense to change the script you use on your side to re-create the source string in a way so that it fails as soon as we have such warnings? Then you could either fix this immediately or report it back to the developers via an issue.

sfuchs79:
The warnings we currently have when creating the .ts files are maybe not really harmful but interesting :-)
We have two other occurrences of missing Q_OBJECT macro plus some other stuff.
I'm currently trying to fix as many as possible.

dirkhh:
@sfuchs79 thanks for looking into the warnings. One oddity is that I need to post-process the strings to get the plurals right... and yes, I've been lazy following up on these warnings because (last time I checked) the vast majority of them were bogus... but I guess I missed the real ones in the noise.
I'm interested to hear what you find.

bstoeger:
It *does* produce new translation strings though, but I think @dirkhh had enough time to speak up. ;)

PR #1656, opened 2018-09-09
Shutting down the GPS web service (and removing it from Subsurface apps)

janmulder:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

See also the discussion on the ML (same name as this PR). This is the very first start of removal of GPS web service code and UI. As Dirk said: remove support for downloading / applying GPS fixes from the webservice in 4.8.2, I made this simple PR.

Be very careful here as this can break workflow of users that currently have unapplied GPS coordinates sitting in the webservice, and do upgrade to the version where this commit goes in. They have no clean UI way any more to apply those GPS fixes from the desktop.


janmulder:
Yes, valid complain from Travis. Will fix this.

desktop-widgets/mainwindow.ui
```diff
@@ -15,7 +15,16 @@
...
+    <property name="leftMargin">
+     <number>0</number>
+    </property>
+    <property name="topMargin">
+     <number>0</number>
+    </property>
+    <property name="rightMargin">
+     <number>0</number>
+    </property>
+    <property name="bottomMargin">
```
dirkhh: I hate the noise that various versions of the designer create in the output

desktop-widgets/mainwindow.ui
```diff
@@ -53,7 +62,7 @@
      <x>0</x>
      <y>0</y>
      <width>861</width>
-     <height>23</height>
+     <height>30</height>
```
dirkhh: and stuff like this always worries me. You remove things, yet the height value went up. Hmm?

desktop-widgets/mainwindow.ui
```diff
@@ -453,7 +453,7 @@
     <bool>true</bool>
    </property>
    <property name="icon">
-    <iconset resource="../subsurface.qrc">
+    <iconset>
```
dirkhh: And this REALLY worries me.
TBH, I'd much prefer if you edited the diff to just  contain the removal of the actionDownloadWeb and not touch anything else. I know that's paranoid, but...

desktop-widgets/mainwindow.ui
```diff
@@ -719,7 +719,7 @@
     <bool>true</bool>
    </property>
    <property name="text">
-    <string>Cloud storage online</string>
+    <string>Cloud stora&amp;ge online</string>
```
dirkhh: and then stuff like this... did you intentionally do that? I doubt that...

dirkhh:
I think this is the right next step.
Many of the changes done by the designer here worry me. Most likely for no reason, but I'd rather not do them.
This needs a changelog entry
This needs documentation changes.
I *know* that I said we should do this for 4.8.2, I realize that I'm asking for a lot. So if this is too much too quickly, maybe we need to delay until after 4.8.2, but I think this is doable.

desktop-widgets/mainwindow.ui
janmulder: Me too, I even re-did my QtCreator delete a second time, be be very sure I did nothing else then delete the part I wanted. By agree, I will filter the real change a leave everything else untouched.

janmulder:
All agree. Will cleanup that QtCreator crap and do the rest (like changelog), and docs together with @willemferguson

dirkhh:
So this look good, Travis likes it, it has a CHANGELOG, which means the only thing missing are the edits to the usermanual... I'll merge this so it hopefully gets some testing for unintended side effects :-)

PR #1659, opened 2018-09-09
Planner: fix display of "overlapping dives" message

bstoeger:
1) Add a missing `<div>`

2) More importantly: recognize html content via `<div>`-tags instead of
   `<table>`-tags.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a display bug. To test:
CTRL-L. Save.
Open dive in planner. Save as new.
CTRL-L. Save.
--> Message will be shown as HTML gibberish.

I'm not sure if that fix is correct!
### Changes made:

1) Add missing `<div>`-tag
2) Recognize HTML by `<div>` instead of `<table>`-tag.
### Release note:

Probably too trivial.
### Documentation change:

No
### Mentions:

@atdotde @sfuchs79 @dirkhh

### DIFF
--- a/core/plannernotes.c
+++ b/core/plannernotes.c
@@ -120,7 +120,7 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
 	}
 
 	if (diveplan->surface_interval < 0) {
-		put_format(&buf, "<div><b>%s (%s) %s<br>",
+		put_format(&buf, "<div><b>%s (%s) %s<br></div>",
 			translate("gettextFromC", "Subsurface"),
 			subsurface_canonical_version(),
 			translate("gettextFromC", "dive plan</b> (overlapping dives detected)"));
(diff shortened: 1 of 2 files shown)


bstoeger:
screenshot

Attaching a screenshot showing the problem.

> In any case, it is my understanding that the only case in which the notes contain html (rather than ascii text) is output from the planner. For user entered text, it should be plain text, so what's the difference between div an table?

The problem is that currently HTML is identified by the `<table>` tag, which the above message does not contain. Therefore I switched all tests from `<table>` to `<div>` (which *all* planner-generated messages contain), without auditing each case. This makes the problem go away for me.

atdotde (approved):
Ah, you need the overlapping times to trigger this. Fair enough.

bstoeger:
Yes, the steps above are supposed to produce two planned dives withe the same time and then another planned dive. No idea why it doesn't do this for you..?

dirkhh:
@atdotde approved, no one commented on my "both div and table" comment, so I guess this is good to go.

PR #1661, opened 2018-09-10
clean mobile-widgets and use qPref directly in cpp as well as qml

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

With qPref being stable and available in C++ as well as QML, it is time to
reduce qmlPrefs.cpp and qmlManager.cpp, who to a large extent serve as
a gateway between C++ and QML, but in a manner that are costly and do not allow signalling

qmlPrefs and qmlManager are available in QML as the C++ Class, meaning the objects will be
instanciated in QML (this is rather costly and memory consuming). qPref* are available directly
as C++ Object in QML and thus without any significant overhead.

The changes (especially around cloud credentials) are rather delicate, and needs a lot of manual testing, for that reason, each variable will be in a separate commit.

Remark: this PR is not concerned with how the qPref variables are used in the rest of the code, this is part of another discussion outside my scope.

This was the low hanging fruits in qmlmanager and qmlprefs, which all was gateway functions to gateway functions and thus not needed. All of the remaining parts needs more code changes.

Remark: there should be no connection to the desktop 4.8.2 release.
### Changes made:

1) clean cloud_storage_password

### DIFF
--- a/core/gpslocation.cpp
+++ b/core/gpslocation.cpp
@@ -3,6 +3,7 @@
 #include "qt-models/gpslistmodel.h"
 #include "core/pref.h"
 #include "core/qthelper.h"
+#include "core/settings/qPrefLocationService.h"
 #include <time.h>
 #include <unistd.h>
 #include <QDebug>
@@ -36,6 +37,9 @@ GpsLocation::GpsLocation(void (*showMsgCB)(const char *), QObject *parent) :
 	userAgent = getUserAgent();
 	(void)getGpsSource();
 	loadFromStorage();
+
+	// register changes in time threshold
+	connect(qPrefLocationService::instance(), SIGNAL(qPrefLocationService::time_thresholdChanged()), this, SLOT(setGpsTimeThreshold(int seconds)));
 }
 
 GpsLocation *GpsLocation::instance()
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -139,8 +139,7 @@ Kirigami.ScrollablePage {
 				enabled: subsurfaceTheme.currentTheme !== "Blue"
 				onClicked: {
 					blueTheme()
-					prefs.theme = subsurfaceTheme.currentTheme
-					manager.savePreferences()
+					PrefDisplay.theme = subsurfaceTheme.currentTheme
 				}
 			}
 
@@ -192,8 +191,7 @@ Kirigami.ScrollablePage {
 				enabled: subsurfaceTheme.currentTheme !== "Pink"
 				onClicked: {
 					pinkTheme()
-					prefs.theme = subsurfaceTheme.currentTheme
-					manager.savePreferences()
+					PrefDisplay.theme = subsurfaceTheme.currentTheme
 				}
 			}
 
@@ -244,8 +242,7 @@ Kirigami.ScrollablePage {
 				enabled: subsurfaceTheme.currentTheme !== "Dark"
 				onClicked: {
 					darkTheme()
-					prefs.theme = subsurfaceTheme.currentTheme
-					manager.savePreferences()
+					PrefDisplay.theme = subsurfaceTheme.currentTheme
 				}
 			}
 		}
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -363,10 +363,12 @@ Kirigami.ScrollablePage {
 			}
 			SsrfSwitch {
 				id: developerButton
-				checked: prefs.developer
+				checked: PrefDisplay.show_developer
 				Layout.preferredWidth: gridWidth * 0.25
 				onClicked: {
-					prefs.developer = checked
+					console.log("JAN changing developer menu(" + PrefDisplay.show_developer + ")")
+					PrefDisplay.show_developer = checked
+					console.log("JAN changed developer menu(" + PrefDisplay.show_developer + ")")
 				}
 			}
 		}
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -278,11 +278,10 @@ Kirigami.ScrollablePage {
 
 			Controls.TextField {
 				id: distanceThreshold
-				text: prefs.distanceThreshold
+				text: PrefLocationService.distance_threshold
 				Layout.preferredWidth: gridWidth * 0.25
 				onEditingFinished: {
-					prefs.distanceThreshold = distanceThreshold.text
-					manager.savePreferences()
+					PrefLocationService.distance_threshold = distanceThreshold.text
 				}
 			}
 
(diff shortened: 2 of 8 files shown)


janiversen:
@dirkhh I need a bit of help in understanding the difference between
prefs.cloud_storage_email (which is used as userid on the git storage)
and
prefs.userid
they seem conflicting to me.

Secondly we have
prefs.cloud_storage_newpassword (which are not saved to disk)
that seems like a variable that should not exist ?

Ps. I am not doing any general cleanup, that I leave to others, but I am trying to reduce qmlprefs and qmlmanager, and for that reason I am confused.

janmulder:
I would not worry to much about `prefs.userid`. This is (for as far as I know) only used for the GPS webservice thing, which is in the process to be deleted.

janiversen:
thanks that clears part of my confusion.

dirkhh:
And newpassword indeed shouldn't be in prefs - this is a case where it just seemed simple to keep it there - but that's of course bogus.

janiversen:
based on the feedback, I will remove userid and newpassword from mobile-widget (make the latter a class variable).

I will consider removing those 2 from qPref, if it can be done without effects outside mobile-widget and qPref.

I am currently untangling the whole mobile login process, which seems overly complicated (from a source code perspective)l

dirkhh:
I'm not sure how much time I'll have today, but I'll be happy to look into taking care of `newpassword` and `git_local`

janiversen:
Thanks I was not aware of this, you can hold back this PR if you like, and I will amend to it (was preparing for a second PR)

dirkhh:
This all depends on how confident you are that you didn't break anything, doesn't it? :-)
Seriously, I don't see a point in slowing you down as you improve the code, I just don't want to end up breaking things right before cutting releases.

mobile-widgets/qml/Settings.qml
```diff
@@ -363,10 +363,12 @@ Kirigami.ScrollablePage {
...
 			SsrfSwitch {
 				id: developerButton
-				checked: prefs.developer
+				checked: PrefDisplay.show_developer
 				Layout.preferredWidth: gridWidth * 0.25
 				onClicked: {
-					prefs.developer = checked
+					console.log("JAN changing developer menu(" + PrefDisplay.show_developer + ")")
+					PrefDisplay.show_developer = checked
+					console.log("JAN changed developer menu(" + PrefDisplay.show_developer + ")")
```
dirkhh: Entertaining to have in the log, but maybe not in production code :-)

core/gpslocation.cpp
```diff
@@ -36,6 +37,9 @@ GpsLocation::GpsLocation(void (*showMsgCB)(const char *), QObject *parent) :
 	userAgent = getUserAgent();
 	(void)getGpsSource();
 	loadFromStorage();
+
+	// register changes in time threshold
+	connect(qPrefLocationService::instance(), SIGNAL(qPrefLocationService::time_thresholdChanged()), this, SLOT(setGpsTimeThreshold(int seconds)));
```
dirkhh, line 42: the code doesn't seem to match the commit message. Yes, you now listen to a signal, but where is the "do not call" part?
Or am I misunderstanding the commit message?

mobile-widgets/qml/Settings.qml
```diff
@@ -293,11 +293,10 @@ Kirigami.ScrollablePage {
 
 			Controls.TextField {
 				id: timeThreshold
-				text: prefs.timeThreshold
+				text: PrefLocationService.time_threshold / 60
 				Layout.preferredWidth: gridWidth * 0.25
 				onEditingFinished: {
-					prefs.timeThreshold = timeThreshold.text
-					manager.savePreferences()
+					PrefLocationService.time_threshold = timeThreshold.text * 60
```
dirkhh, line 296: not criticism, just curious... why did this change between minutes and seconds?

dirkhh (changes requested):
I **really** like the granularity of these changes.
Easy to understand, should be easy to test as well.
Let's not leave the log message in there, also a couple of questions.

janiversen:
Please have a look at the new commit messages and see if they are more descriptive.

Regarding the "how confident I am", I have tested every commit by:
1) start app, change the value en question
2) restart app, and secure value is actually changed

For the part changed in this PR, I could not see much more testing. Whereas future changes like e.g. the whole cloud identification needs much more testing.

janiversen:
@dirkhh  Seems I hit a snag again, I believed #1097 was dead but to be sure and not step on toes, I. asked and to my surprise got the answer, that it is still valid.

Seems whatever I work on gets blocked, which is anything but motivating. My qPref work was blocked by a PR that seems not to be moving, now my mobile-widget work is blocked by a longterm project, that haven't been touched since February.

I am just happy, I checked that PR, before I invested too much time. I honestly thought that I had proven myself in the area of C++ / QML or detangling IO and UI.

I am (or more correctly was until 1 hour ago) working on detangling the whole CloudCred process, which (pardon me) is a big mess, with several illogical user situations (just one example: try to have a verified cred, then change your password to a faulty one...what happens you get the old divelist, until you restart the ap).

I have no intention of repeating the qPref disaster (for me), then better do something else !

dirkhh:
I think your qPref work is moving and continues to move - we still want to move to making prefs constant, but it seems much more that that work is currently on hold because of the changes you had made.
Frankly, I thought the majority of the qPref work (as far as C++ changes) was done and what was left was qmlmanager/QML code updates. Please let me know what I'm obviously missing.

To be honest, I don't quite understand how you pick the areas that you work on. So all that pref work was ostensibly so that you could make significant updates to the QML UI. I don't see how the cloud credentials can be a blocking item for that. I appreciate that you see code that you don't like and want to change it, but I also see that you dislike fundamentally all of the existing code. And the incremental value of you reimplementing all of the existing code differently is not entirely clear to me. Again, maybe I'm missing something obvious.

dirkhh (approved):

janiversen:
qPref is closed from my side....I have what I need for QML and somebody will implement the missing parts in the rest of the system using this strange const thing.

It is not true that I dislike all code far from, I concentrated on qPref to make a stable platform for QML, and it was natural to integrate qPref in the rest of the system, which probably was my mistake. I think your judgement is a bit unfair, if you look at my commits, they are all (apart from a few stray ones) directed at making the qml-c++ interface robust. It is correct I dislike spaghetti code, which is what I see right now with cloudCred...I understand how it came to life, so it natural it ends more complicated than it needs to be...simplifying existing code is a factor easier than writing it in the first place (I salut those who had the initial task of making the Qml - c++ interface, while learning qml, and getting the app to work.....a real challenge and a nice result).

Now I have climbed the ladder, to demangle qmlmanager/qmlpref or more correctly to simplify. That part will change a lot of QML (and not just renaming). qPref is no blocker at all, but seeing somebody else have a long term project running to partly make the same happen, makes me carefull not to make work for nothing again (I have thrown away quite a number of prepared qPref integration in the rest of the code, due to the const decision).

CloudCred is basically blocked because I thought I was the only one working on detangling the mobile UI - IO, but now I know I am not. In contrary to the other, I keep my number of open PR´s and themes at 1, and I quickly finalize things....having a PR not being touched since february is to me a really loooong term project.

PR #1662, opened 2018-09-10
Optional anonymization upon export

atdotde:
Add a checkbox that triggers replacement of all English characters by
x's in notes, buddy, dive guide and (while we are at it) suit.

This is ment for people sharing logs for debugging that are concious
about privacy issues. It leaves the lenth of strings in tact as well
as special charcters as those might be needed to track down a particular
parsing problem.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

This was discussed in #1624 .

### DIFF
--- a/core/save-xml.c
+++ b/core/save-xml.c
@@ -63,6 +63,35 @@ static void show_utf8(struct membuffer *b, const char *text, const char *pre, co
 	free(cleaned);
 }
 
+static void blankout(char *c)
+{
+	while(*c) {
+		switch (*c) {
+		case 'A'...'Z':
+			*c = 'X';
+			break;
+		case 'a'...'z':
+			*c = 'x';
+			break;
+		default:
+			;
+		}
+		++c;
+	}
+}
+
+static void show_utf8_blanked(struct membuffer *b, const char *text, const char *pre, const char *post, int is_attribute, bool anonymize)
+{
+	if (!text)
+		return;
+	char *copy = strdup(text);
+
+	if (anonymize)
+		blankout(copy);
+	show_utf8(b, copy, pre, post, is_attribute);
+	free(copy);
+}
+
 static void save_depths(struct membuffer *b, struct divecomputer *dc)
 {
 	/* What's the point of this dive entry again? */
@@ -119,12 +148,12 @@ static void save_salinity(struct membuffer *b, struct divecomputer *dc)
 	put_string(b, " />\n");
 }
 
-static void save_overview(struct membuffer *b, struct dive *dive)
+static void save_overview(struct membuffer *b, struct dive *dive, bool anonymize)
 {
-	show_utf8(b, dive->divemaster, "  <divemaster>", "</divemaster>\n", 0);
-	show_utf8(b, dive->buddy, "  <buddy>", "</buddy>\n", 0);
-	show_utf8(b, dive->notes, "  <notes>", "</notes>\n", 0);
-	show_utf8(b, dive->suit, "  <suit>", "</suit>\n", 0);
+	show_utf8_blanked(b, dive->divemaster, "  <divemaster>", "</divemaster>\n", 0, anonymize);
+	show_utf8_blanked(b, dive->buddy, "  <buddy>", "</buddy>\n", 0, anonymize);
+	show_utf8_blanked(b, dive->notes, "  <notes>", "</notes>\n", 0, anonymize);
+	show_utf8_blanked(b, dive->suit, "  <suit>", "</suit>\n", 0, anonymize);
 }
 
 static void put_gasmix(struct membuffer *b, struct gasmix mix)
@@ -443,7 +472,7 @@ static void save_picture(struct membuffer *b, struct picture *pic)
 	put_string(b, "/>\n");
 }
 
-void save_one_dive_to_mb(struct membuffer *b, struct dive *dive)
+void save_one_dive_to_mb(struct membuffer *b, struct dive *dive, bool anonymize)
 {
 	struct divecomputer *dc;
 
@@ -469,7 +498,7 @@ void save_one_dive_to_mb(struct membuffer *b, struct dive *dive)
 			   FRACTION(dive->dc.duration.seconds, 60));
 	else
 		put_format(b, ">\n");
-	save_overview(b, dive);
+	save_overview(b, dive, anonymize);
 	save_cylinder_info(b, dive);
 	save_weightsystem_info(b, dive);
 	save_dive_temperature(b, dive);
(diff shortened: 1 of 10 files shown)


core/save-xml.c
```diff
@@ -119,12 +119,60 @@ static void save_salinity(struct membuffer *b, struct divecomputer *dc)
...
+		++c;
+	}
+}
+
+static void save_overview(struct membuffer *b, struct dive *dive, bool annonymize)
+{
+	char *xxxs;
+	if (annonymize) {
+		if (dive->divemaster) {
+			xxxs = strdup(dive->divemaster);
```
bstoeger: I wonder if a `show_utf8_xxx()` function would make that more readable.

dirkhh:
This breaks the building of non-subsurface binaries linking against our core libraries - see Travis output

core/save-xml.c
```diff
@@ -119,12 +119,60 @@ static void save_salinity(struct membuffer *b, struct divecomputer *dc)
 	put_string(b, " />\n");
 }
 
-static void save_overview(struct membuffer *b, struct dive *dive)
+static void xxxout(char *c)
```
dirkhh: not in love with XXX function names... how about asciiToX()

core/save-xml.c
```diff
@@ -119,12 +119,60 @@ static void save_salinity(struct membuffer *b, struct divecomputer *dc)
...
+		default:
+			;
+		}
+		++c;
+	}
+}
+
+static void save_overview(struct membuffer *b, struct dive *dive, bool annonymize)
+{
+	char *xxxs;
```
dirkhh: again, XXX is unfortunately a standing expression I'd prefer we'd stay away from.

desktop-widgets/divelogexportdialog.ui
```diff
@@ -279,20 +279,26 @@
...
             <widget class="QComboBox" name="CSVUnits_2">
+             <property name="enabled">
+              <bool>false</bool>
+             </property>
              <property name="geometry">
               <rect>
-               <x>30</x>
-               <y>30</y>
+               <x>10</x>
+               <y>20</y>
```
dirkhh: above this line there are a few changes here which I'm not sure about... are these related to this PR?

desktop-widgets/divelogexportdialog.ui
```diff
@@ -279,20 +279,26 @@
...
                <width>102</width>
                <height>27</height>
               </rect>
              </property>
+             <property name="currentText">
+              <string>Metric</string>
+             </property>
```
dirkhh: Oh, and this one.

dirkhh (changes requested):
a) libdivecomputer change: NAK
b) the 'XXX' naming - I'd prefer our software to be PG
c) this is really noisy for a really limited use case - I'm not 100% convinced that this is the right way to go. Why not just do this in post processing in a perl scrip?

bstoeger:
> c) this is really noisy for a really limited use case - I'm not 100% convinced that this is the right way to go. Why not just do this in post processing in a perl scrip?

Wouldn't the version I suggested be much less noisy:
```
static void save_overview(struct membuffer *b, struct dive *dive, bool annonymize)
{
	if (annonymize) {
		show_utf8_XYZ(b, dive->divemaster, "  <divemaster>", "</divemaster>\n", 0);
		show_utf8_XYZ(b, dive->buddy, "  <buddy>", "</buddy>\n", 0);
		show_utf8_XYZ(b, dive->notes, "  <notes>", "</notes>\n", 0);
		show_utf8_XYZ(b, dive->suit, "  <suit>", "</suit>\n", 0);
	} else {
		show_utf8(b, dive->divemaster, "  <divemaster>", "</divemaster>\n", 0);
		show_utf8(b, dive->buddy, "  <buddy>", "</buddy>\n", 0);
		show_utf8(b, dive->notes, "  <notes>", "</notes>\n", 0);
		show_utf8(b, dive->suit, "  <suit>", "</suit>\n", 0);
	}
}
```
It could be condensed even more:
```
static void save_overview(struct membuffer *b, struct dive *dive, bool annonymize)
{
	show_utf8_XYZ(b, dive->divemaster, "  <divemaster>", "</divemaster>\n", 0, anonymize);
	show_utf8_XYZ(b, dive->buddy, "  <buddy>", "</buddy>\n", 0, anonymize);
	show_utf8_XYZ(b, dive->notes, "  <notes>", "</notes>\n", 0, anonymize);
	show_utf8_XYZ(b, dive->suit, "  <suit>", "</suit>\n", 0, anonymize);
}
```
`XYZ` being a placeholder, obviously.

atdotde:
I will of course fix the libdivecomputer thing and the build failures as well as function/variable names. I could manually fix the Designer mishaps.

I am not sure about the show_utf8_XYZ suggestions. I had considered those before but decided against since a) adding the boolean parameter there would mean changing a lot more code since they are called in many places and b) since this is a debugging feature I wanted annqonymized and ordinary code paths to be as similar as possible.

Being a perl monger myself, going for a post processing perl script that modifies the xml was of course my first idea. But since this is supposed to be run by the user before sending us her log, I did not want to assume she has a perl compiler set up. Again, I could offer this as a web service via a cgi-script but I thought privacy conscious users would not like to upload their unencrypted data to a random web site.

Please comment.

bstoeger:
I fear that I can't follow your arguments:

> a) adding the boolean parameter there would mean changing a lot more code since they are called in many places.

No - the rest of the code would continue using `show_utf8` without additional parameter.

> b) since this is a debugging feature I wanted annqonymized and ordinary code paths to be as similar as possible.

But aren't both of my example functions *much* closer to the "ordinary code path" than your version?

Perhaps I'm not clear about my idea. But `show_utf8_XYZ` in the second example would simply look like:

```
static void show_utf8_XYZ(struct membuffer *b, const char *prefix, const char *value, const char *postfix, bool anonymize)
{
	if (!anonymize) {
		show_utf8(b, prefix, value, postfix);
		return;
	}
	[... do what you're doing four times in save_overview ...]
}
```

Having the code only once looks much less error prone, doesn't it?

atdotde:
I guess this update addresses all your concerns (let's see if Travis is now happy) except the general "maybe this should be a perl script".

bstoeger (approved):
lgtm

bstoeger:
There's still the problem with the stand-alone programs, though..

bstoeger:
git grep finds these two instances:
```
smtk-import/smrtk2ssrfc_window.cpp:     save_dives_logic(qPrintable(outputFile), false);
smtk-import/smtk_standalone.cpp:                save_dives_logic(outfile, false);
```
which need to be fixed.

atdotde:
Finally, Travis is happy. @dirkhh do you still prefer a perl script (I would be happy to provide that but see above)?

dirkhh:
I'll review in a little bit. I see the logic why having this in the app is better.

bstoeger:
Just a super-tiny nitpick: "anonym" has only one 'n' between 'a' and 'o'. Yeah, throwing stones while living in a glass house... :-P

atdotde:
I knew, I should have checked the spelling...

dirkhh (approved):

atdotde:
This checkbox is only relevant for export, not for saving. Plus you always get a warning when a file is overwritten (and a backup is saved).

bstoeger:
You're right - that should be fine, especially if a backup is produced.

dirkhh:
No risk no fun. I think we should randomly set this flag for regular saving of data as well and call it a data privacy feature. Especially for all the people who complain about using the Subsurface Cloud...

(I am NOT in a good mood this morning...)

PR #1663, opened 2018-09-10
Remove two prefs

dirkhh:
`git_local_only` and `cloud_new_password` are two examples of abuse of the `prefs` structure. Neither of them should be a preference.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

#1661 #1653 #1633 

### Additional information:

we need to go through the code and find the other places where we use `prefs` and really shouldn't be

### Mentions:

@bstoeger @janiversen

### DIFF
--- a/core/cloudstorage.cpp
+++ b/core/cloudstorage.cpp
@@ -26,6 +26,7 @@ QNetworkReply* CloudStorageAuthenticate::backend(const QString& email,const QStr
 	} else if (!newpasswd.isEmpty()) {
 		requestUrl = QUrl(CLOUDBACKENDUPDATE);
 		payload += QChar(' ') + newpasswd;
+		cloudNewPassword = newpasswd;
 	} else {
 		requestUrl = QUrl(CLOUDBACKENDVERIFY);
 		payload += QChar(' ') + pin;
@@ -64,8 +65,8 @@ void CloudStorageAuthenticate::uploadFinished()
 		report_error(qPrintable(tr("Cloud account verification required, enter PIN in preferences")));
 	} else if (cloudAuthReply == QLatin1String("[PASSWDCHANGED]")) {
 		free((void *)prefs.cloud_storage_password);
-		prefs.cloud_storage_password = prefs.cloud_storage_newpassword;
-		prefs.cloud_storage_newpassword = NULL;
+		prefs.cloud_storage_password = strdup(qPrintable(cloudNewPassword));
+		cloudNewPassword.clear();
 		emit passwordChangeSuccessful();
 		return;
 	} else {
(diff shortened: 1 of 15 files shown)


core/cloudstorage.cpp
```diff
@@ -64,8 +65,8 @@ void CloudStorageAuthenticate::uploadFinished()
 		report_error(qPrintable(tr("Cloud account verification required, enter PIN in preferences")));
 	} else if (cloudAuthReply == QLatin1String("[PASSWDCHANGED]")) {
 		free((void *)prefs.cloud_storage_password);
-		prefs.cloud_storage_password = prefs.cloud_storage_newpassword;
-		prefs.cloud_storage_newpassword = NULL;
+		prefs.cloud_storage_password = strdup(qPrintable(cloudNewPassword));
```
janiversen: qPrefCloudStorage::set_password sounds like the better option, considering intent of making prefs readonly

core/git-access.c
```diff
@@ -27,6 +27,15 @@
 
 bool is_subsurface_cloud = false;
 
+// the mobile app assumes that it shouldn't talk to the cloud
+// the desktop app assumes that it should
```
janiversen, line 31: change is correct, but it sounds stupid, a mobile that prefers local use

core/subsurfacestartup.c
```diff
@@ -12,9 +12,6 @@
 struct preferences prefs, git_prefs;
 struct preferences default_prefs = {
 	.cloud_base_url = "https://cloud.subsurface-divelog.org/",
-#if defined(SUBSURFACE_MOBILE)
-	.git_local_only = true,
```
janiversen, line 16: just a comment, default is as I expected, which makes the earlier comment =false for mobile even more correct.

janiversen (commented):
Added my comments. This PR will cause conflict with my PR (signalTest), but that is life.

janiversen:
now I know, why that commit (constructor private) was removed, it changed a lot of files, which seemed unpopular, so I moved it to my remember_to_do branch.

janiversen:
sounds correct....apart from the “better way”,  I have learned this depend pretty much on the viewer.

but yes your patch is not wrong, it gets the job done... sorry for nick picking.

so just merge it.

dirkhh:
you're not nit-picking, you are providing feedback, and I'm responding to the feedback.
Travis still isn't happy and I'm not sure what's wrong with Travis... because the logs just stop. Or they end with Travis unable to resolve github.com...

dirkhh:
Found the final typo.

janmulder:
I believe I see a (small) regression, that is most likely related to the `git_local_only` thing.

On desktop. Open a local logbook (in my case a git style one). Open "File" menu. Notice that the menu option "Cloud storage online" is grayed out, and the toggle is unselected. Chose "Save to Cloud". It correctly saves to cloud, and this ends in a state where "Cloud storage online" is not grayed out any more, and we are "online", with the toggle selected.

I said "regression", but not sure it is. It is different as before this commit, the toggle "Cloud storage online" was not touched by "Save to Cloud", but now it is. It now brings the cloud storage online. Not sure this is an intended change ...

dirkhh:
Mostly intended. Or to be precise, the old behavior of remembering if last time you started you had connectivity problems was unintended, so this gets us back to where we were a few releases ago.

PR #1660, opened 2018-09-10
Mobile: Remove webservice, remove UI components

janmulder:
### Describe the pull request:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This is the first commit related to the removal of the GPS webservice. It is nothing more then removing 2 buttons from the menu to upload and download from the server, so technically a trivial change.

### Additional information:
As with the desktop application: Be very careful here as this forces our users to use Subsurface-mobile, and a online cloud account as that is the way to transfer GPS data from a mobile device to the desktop.

### Release note:
CHANGELOG.md. added

### Documentation change:
Yes, needed.

PR #1667, opened 2018-09-11
Remove GPS webservice backend

janmulder:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a set of cleanups related to the removal of 2 preferences: save_userid_local and userid. The commits are ordered so that a sane running state remains, should a bisect ever lands here.

Obviously, old logbook file in XML and git storage including these preferences are still read but the userid data is silently ignored, and on write no userid data is written any more. And for the rest: is is just removal of unsed code due to the removal of the 2 prefence values.

### Changes made:
see commits, and testcases are updated.

### Related issues:
#1656, #1660

### Additional information:
No need to get this in 4.8.2, but I believe its safe :-)


dirkhh:
This looks solid. Given that we removed the UI, we might as well get this in as well.
Thanks

PR #1668, opened 2018-09-12
serial_ftdi: use Sleep() on Win32

neolit123:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Windows doesn't have nanosleep() unless libwinpthread is used.

Since the nanosleep() usage in serial_ftdi_sleep():
- does not break in case of EINTR
- has input in milliseconds

the WINAPI Sleep() should be a good alternative.

### Changes made:

1) call Sleep() on Win32.

### Related issues:

NONE

### Additional information:

i must admit, i didn't try building with SERIAL_FTDI, so let's see if CI is happy.

### Release note:

NONE

### Documentation change:

NONE

### Mentions:

@dirkhh

### DIFF
--- a/core/serial_ftdi.c
+++ b/core/serial_ftdi.c
@@ -26,11 +26,16 @@
 #include <string.h>     // strerror
 #include <errno.h>      // errno
 #include <sys/time.h>   // gettimeofday
-#include <time.h>       // nanosleep
 #include <stdio.h>
 
 #include <libusb.h>
 #include <ftdi.h>
+
+#ifdef _WIN32
+#include <windows.h>    // Sleep
+#else
+#include <time.h>       // nanosleep
+#endif
 
 #ifndef __ANDROID__
 #define INFO(context, fmt, ...)	fprintf(stderr, "INFO: " fmt "\n", ##__VA_ARGS__)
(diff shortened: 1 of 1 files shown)


core/serial_ftdi.c
```diff
@@ -107,6 +112,9 @@ static dc_status_t serial_ftdi_sleep (void *io, unsigned int timeout)
 
 	INFO (device->context, "Sleep: value=%u", timeout);
 
+#ifdef __MINGW32__
```
glance-: Why `__MINGW32__` and not `WIN32`?

dirkhh: A quick grep through our sources seems to indicate that we do indeed use WIN32 in situations like this. Would you please change that, @neolit123

dirkhh (changes requested):
WIN32 change

neolit123:
ACK for consistency. will try to updated later today.

neolit123:
updated to use `_WIN32`.

dirkhh:
UH, does this contain a libdivecomputer update by mistake?

neolit123:
>UH, does this contain a libdivecomputer update by mistake?

sorry, fixed.

dirkhh:
Thanks @neolit123

dirkhh (approved):

PR #1671, opened 2018-09-13
Mac: hard code macosx-version-min as 10.11, regardless of available SDKs

### LINKED ISSUE #1654: travis mac build problems.
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

The Travis mac build have started to generate a ton of warnings from libdivecomputer, all like:

ld: warning: object file (/Users/travis/build/Subsurface-divelog/install-root/lib/libdivecomputer.a(version.o)) was built for newer OSX version (10.12) than being linked (10.11)

It seems there is a make problem with libdivecomputer. This is potential dangerous, it we use or start using a API call that are different between 10.12 and 10.11, API are backwards compatible (compile for 10.11 but link for 10.12) but not forward compatible.

dirkhh:
This should never be wrong, and should avoid the warnings on Travis.

Fixes #1654

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@janiversen I think this fixes the issue you raised

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -88,7 +88,7 @@ if [ $PLATFORM = Darwin ] ; then
 		exit 1;
 	fi
 	BASESDK=$(ls $SDKROOT | grep "MacOSX10\.1.\.sdk" | head -1 | sed -e "s/MacOSX//;s/\.sdk//")
-	OLDER_MAC="-mmacosx-version-min=${BASESDK} -isysroot${SDKROOT}/MacOSX${BASESDK}.sdk"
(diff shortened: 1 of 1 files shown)


janiversen:
Quite ok with me. Developers who start installing Xcode today, might experience they lack libraries for version 10.11, but it is easy sorted out.

PR #1672, opened 2018-09-13
Proof of concept / do not apply: Make QMLPrefs a proper singleton

bstoeger:
We had that absolutely crazy construct, where in some QML-accessed
singletons the instance variable was set in the constructor. The
background was that apparently if registered as a normal type, QML
would insist on constructing an object via "new T".

Nevertheless, as a quick look in the documentation shows, QML *does*
support singletons with the qmlRegisterSingletonType() function.
This is a first proof of concept to make the QMLPrefs class a real
singleton.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is merely a proof of concept / request for discussion. QML *does* know singletons. Therefore it seems to me that we should use that for the few QML types that we have (c.f. `subsurface-helper.cpp`). This commit was mostly done blindfolded and I'm sure there is some major breakage. But it runs *and* shows my dives.

Does this fix a user-visible problem? No. Does it remove a major embaressment? Hell yes.

Yes, this used a template. This can of course be converted to a macro.

More Info: this makes the constructor private, which means that QML can in no way construct its own copies.

### Changes made:

Make `QMLPrefs` a singleton.

### Release note:

Just a proof of concept.

### Documentation change:

Just a proof of concept.

### DIFF
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -15,9 +15,9 @@ Item {
 	property string password: password.text;
 
 	function saveCredentials() {
-		prefs.cloudUserName = login.text
-		prefs.cloudPassword = password.text
-		prefs.cloudPin = pin.text
+		QMLPrefs.cloudUserName = login.text
+		QMLPrefs.cloudPassword = password.text
+		QMLPrefs.cloudPin = pin.text
 		manager.saveCloudCredentials()
 	}
 
@@ -53,15 +53,15 @@ Item {
 
 		Controls.Label {
 			text: qsTr("Email")
-			visible: !rootItem.showPin
+			visible: !QMLPrefs.showPin
 			font.pointSize: subsurfaceTheme.smallPointSize
 			color: subsurfaceTheme.secondaryTextColor
 		}
 
 		Controls.TextField {
 			id: login
-			text: prefs.cloudUserName
-			visible: !rootItem.showPin
+			text: QMLPrefs.cloudUserName
+			visible: !QMLPrefs.showPin
 			Layout.fillWidth: true
 			inputMethodHints: Qt.ImhEmailCharactersOnly |
 					  Qt.ImhNoAutoUppercase
@@ -69,15 +69,15 @@ Item {
 
 		Controls.Label {
 			text: qsTr("Password")
-			visible: !rootItem.showPin
+			visible: !QMLPrefs.showPin
 			font.pointSize: subsurfaceTheme.smallPointSize
 			color: subsurfaceTheme.secondaryTextColor
 		}
 
 		Controls.TextField {
 			id: password
-			text: prefs.cloudPassword
-			visible: !rootItem.showPin
+			text: QMLPrefs.cloudPassword
+			visible: !QMLPrefs.showPin
 			echoMode: TextInput.PasswordEchoOnEdit
 			inputMethodHints: Qt.ImhSensitiveData |
 					  Qt.ImhHiddenText |
@@ -87,20 +87,20 @@ Item {
 
 		Controls.Label {
 			text: qsTr("PIN")
-			visible: rootItem.showPin
+			visible: QMLPrefs.showPin
 		}
 		Controls.TextField {
 			id: pin
 			text: ""
 			Layout.fillWidth: true
-			visible: rootItem.showPin
+			visible: QMLPrefs.showPin
 		}
 
 		RowLayout {
 			Layout.fillWidth: true
 			Layout.margins: Kirigami.Units.smallSpacing
 			spacing: Kirigami.Units.smallSpacing
-			visible: rootItem.showPin
+			visible: QMLPrefs.showPin
 			SsrfButton {
 				id: registerpin
 				text: qsTr("Register") 
@@ -126,7 +126,7 @@ Item {
 			Layout.fillWidth: true
 			Layout.margins: Kirigami.Units.smallSpacing
 			spacing: Kirigami.Units.smallSpacing
-			visible: !rootItem.showPin
+			visible: !QMLPrefs.showPin
 
 			SsrfButton {
 				id: signin_register_normal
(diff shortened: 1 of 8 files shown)


janmulder:
> But it runs and shows my dives

Indeed, but for me it aborts with a SIGSEGV on exit of the app in `QmlType::SingletonInstanceInfo::destroy()`

bstoeger:
Cool. :)

https://bugreports.qt.io/browse/QTBUG-33398
Reported 2013(!).

In any case easy enough to fix.

dirkhh:
@bstoeger the template use here does make sense and the code seems readable.
Eventually I'll grow up and get over my template phobia.

You mark this as proof of concept do not merge.
Is there anything conceptually wrong with the code?
Besides being ugly, what's conceptually wrong with the existing code?

bstoeger:
> @thiagomacieira any idea if this bug will be fixed at some point?

It's not a bug (perhaps a documentation/design bug). It's designed as that. The problem is that for QML singleton means one object *per QML engine*. For our C++ code singleton means one object *per application*. QML guarantees that the generator-function is called only once per engine, but expects that we return a new heap-allocated object, which it will delete later on. Fixing this issue would probably mean adding more state to the QML type system.

Therefore, we'll always have some strange code with this respect and we should probably consider a few options. Give me a day or two, please.

With "easy enough to fix" I meant fix in our code not in Qt - which I did. Sorry for being unclear. I should learn to express myself more clearly / detailed.

bstoeger:
> Besides being ugly, what's conceptually wrong with the existing code?

This is purely a code-hygiene thing. And yes, it comes with some risk, because QML is inherently non-compile checked. Who knows where it breaks..! I think ultimately the code will be clearer if we access singletons directly via their name instead of via a sub-object.

But this has no hurry.

bstoeger:
> Is there anything conceptually wrong with the code?

For one, I think we should either convert all or none of the QML-types. And it's more logical than the old code, but not super-pretty either.

dirkhh:
To me this raises the question of where we should spend our effort.
The current way we do this is ugly, but it seems to work.
Elegance is a good goal, but if it introduces more risk of breaking things, I'm not sure that's what I'd encourage - and especially if our vision of how things should be is running counter to the QML philosophy / design ideas. Then I think we're better off to live with ugly and work on making the Subsurface UI better...
just my 2¢

bstoeger:
> and especially if our vision of how things should be is running counter to the QML philosophy / design ideas.

Well it can't be more counter to QML than what we have now. :) The problem is the different views of what a singleton is, that is a sad fact and will not go away.

And it's not purely code hygiene. Doing like I sketched means that we do not depend on QML being the first accessor to the instance. Currently, if other code calls `instance()` first -> boom. If we fix that with an `if`-condition in `instance()` we end up with two different objects. Not an immediate crash, but even worse, because harder to track. It was *you* who told me the stories about different instances being a problem. This is absolutely impossible with the scheme above. It makes us independent from subtle order-of-initialization issues. I hope you agree that this is a good thing.

dirkhh:
I thought the current ugly scheme did the right thing even when instantiated first by C++ code.
Shows you how little I know

bstoeger:
Well, maybe I am missing something, but I don't see how that could work? If QML calls `new T` it *will* get a new object. No way around that. If there already is a different instance flying around, we have two objects... No..? The whole point of the `qmlRegisterSingletonType` thing is that it calls a callback, not `new`.

bstoeger:
> Therefore it seems to me that we should use that for the few QML types that we have (c.f. subsurface-helper.cpp).

Except those that aren't singletons (good!), of course:

- QMLProfile
- DCDownloadThread
- DCImportModel
- Map*

dirkhh:
Yeah, you're right - since we can't get QML to call ::instance this relies on QML calling first.
Which means we should either add code that blast loud and huge warnings if the constructor gets called and m_instance isn't NULL, or we should go down your path of actually cleaning this up.

bstoeger:
For things that are only relevant in mobile context (e.g. `QMLManager`), we could also consider going the other way: construct the object in QML and if C++ code needs it, query the QMLEngine.

dirkhh:
> and if C++ code needs it, query the QMLEngine.

How would that work?

bstoeger:
I have no idea - just throwing ideas out there.. :)

But in principle it should work:
http://doc.qt.io/qt-5/qtqml-cppintegration-interactqmlfromcpp.html

janiversen:
Before starting changing qmlprefs and qmlmanager heavily like making them singletons, it might be a good idea to look at what is already there in qPref. The better way to go is to use setContextProperty, and at the same time remove the objects we create in QML.

Please, do not just look at the C++ side, when doing these kind of changes, the QML part is just as important.

I am working on reducing qmlmanager/qmlpref and making the needed parts more stable, but it seems as if my experience is really not wanted here (or at least nobody cared to consult me, but preferred discussing with what seems to be not full knowledge)....there are ABSOLUTELY no need to call xxx:instance() that is done properly with the setContextProperty call. 

@dirkhh  Could I please have an opinion if my work is wanted or I should just keep it in my fork, this is second time in a couple of days @bstoeger decides to suggest important changes to the C++ QML interface, without considering the QML part of the implications and disrupts my ongoing work.

I have no intention of making a long discussion with @bstoeger about whether or not a Singleton is the way to go. If "just throwing ideas" is better than the already proven concept I have got merged then let it be so!

bstoeger:
I'm perfectly fine with setting a property from C++.
[Refraining from commenting on the tone of the message.]

neolit123 (approved):
usage of `QMLPrefs` like that in the QML seems much better to me.
`register_qml_singleton()` seems OK too.

dirkhh:
Yes, we should go back and clean all of this up.

dirkhh:
This really isn't applicable anymore

Yes, one day we might change our preferences magic/nightmare.

Not today. Not this year.

PR #1673, opened 2018-09-13
Desktop: improve navigation options on edit

### LINKED ISSUE #1445: Pictures shown when manually adding dives
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [x] Discussion request

### Issue long description:

When adding a dive, the dive photo tab shows the pictures of the previously selected dive(s). While this is technically correct (the dive selection is still valid), it seems weird from an UI perspective.

### Operating system:

```
$ uname -a
Linux laptop 4.15.0-23-generic #25-Ubuntu SMP Wed May 23 18:02:16 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
I doubt that this is OS-dependent.
### Subsurface version:

Current master.

### Steps to reproduce:
1) Go to photos-tab with dive having pictures.
2) CTRL-+.
### Current behavior:

Previous pictures are shown.
### Expected behavior:

No pictures shown, unless new pictures are added manually (?)

janmulder:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As described in the referenced issue, we where able to navigate to nonlogical static pages (like information, statistics. extra data) when adding a dive. These are output style pages that make no sense on edit or add. Further, disable access to some pages when entering edit mode.

Notice that the small change in file mainwindow.cpp is simply because this did not work at all, and became superfluous any way.

### Related issues:
Fixes: #1445

### Release note:
Not sure... needed for this trivial thing?

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -270,6 +270,11 @@ void MainTab::toggleTriggeredColumn()
 
 void MainTab::addDiveStarted()
 {
+	ui.tabWidget->setCurrentIndex(0);
+	ui.tabWidget->setTabEnabled(2, false);
+	ui.tabWidget->setTabEnabled(3, false);
+	ui.tabWidget->setTabEnabled(4, false);
+	ui.tabWidget->setTabEnabled(5, false);
 	enableEdition(ADD);
 }
 
@@ -339,6 +344,9 @@ void MainTab::enableEdition(EditMode newEditMode)
 	MainWindow::instance()->dive_list()->setEnabled(false);
 	MainWindow::instance()->setEnabledToolbar(false);
 	MainWindow::instance()->enterEditState();
+	ui.tabWidget->setTabEnabled(2, false);
+	ui.tabWidget->setTabEnabled(3, false);
+	ui.tabWidget->setTabEnabled(5, false);
 
 	if (isTripEdit) {
 		// we are editing trip location and notes
@@ -534,6 +542,8 @@ void MainTab::updateDiveInfo(bool clear)
 			ui.tabWidget->setTabText(0, tr("Notes"));
 			ui.tabWidget->setTabEnabled(1, true);
 			ui.tabWidget->setTabEnabled(2, true);
+			ui.tabWidget->setTabEnabled(3, true);
+			ui.tabWidget->setTabEnabled(4, true);
 			ui.tabWidget->setTabEnabled(5, true);
 			// Recover the tab selected for last dive
 			if (!lastSelectedDive)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Desktop: disable UI elements that make no sense during editing [#1445]
 - Mobil: setting for developer menu entry is remembered across sessions
 - Mobile: remove UI components of the Webservice GPS interaction
 - Desktop: remove UI components of the Webservice GPS import. This functionality will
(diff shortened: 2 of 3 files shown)


bstoeger:
Thanks, that has bothered me for the longest time and I never could motivate myself to dive into the tab thing.

> Not sure... needed for this trivial thing?

Perhaps not needed, but I think it would be a good idea, as it fixes a user-visible bug.

dirkhh:
I would agree - this is a behavioral change. A good one, of course, but we should mention it.
Now let me look at the code...

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -339,6 +344,9 @@ void MainTab::enableEdition(EditMode newEditMode)
 	MainWindow::instance()->dive_list()->setEnabled(false);
 	MainWindow::instance()->setEnabledToolbar(false);
 	MainWindow::instance()->enterEditState();
+	ui.tabWidget->setTabEnabled(2, false);
+	ui.tabWidget->setTabEnabled(3, false);
+	ui.tabWidget->setTabEnabled(5, false);
```
dirkhh, line 349: Simply pattern matching makes me wonder why we disable the fourth tab in one case and not the other - I guess I should actually read the code in context and try to figure it out myself...
It's the Media tab... hmm... I call 'bug'

janmulder: As the media tab makes sense on edit :-)

dirkhh: it does? How?

janmulder: You can add media when editing a dive using the right mouse context menu ...

dirkhh: I guess that shows you how often I use _THAT_ feature.

dirkhh:
I hate to be a pain, but that CHANGELOG... hmm.
How about
`- Desktop: disable UI elements that make no sense during editing`

bstoeger:
Tested this and it works nicely. The changes look very reasonable and it fixes a UI-problem. I think we can commit.

My only *very* minor gripe: When I'm on the statistics tab and edit the dive, it jumps to the media tab. I'd prefer if it goes to the notes tab.

bstoeger:
There's another minor inconsistency, though I think it is not necessary to bother now, I just want it documented:

If in edit mode the user adds a picture and then says "Discard Changes", the picture is obviously still there.

bstoeger:
Yes. My personal vision would be to incorporate all that into the undo-system. But that will be a steep uphill battle. :(

janmulder:
@bstoeger  wrote: 
> My only very minor gripe: When I'm on the statistics tab and edit the dive, it jumps to the media tab. I'd prefer if it goes to the notes tab.

This intrigued me. How can you edit a dive from the statistics tab? Oh wait, using the menu log-edit dive. I have to admit that this was the very first time I ever used that menu entry.  In fact, I do not see _any_ reason for this menu option. @dirkhh: why do we need a menu option to edit a dive, when we can just start typing in a dive to be edited?

bstoeger:
Heh. Just shows how different people are. :) Personally, I wouldn't have a problem with the entry being removed.

dirkhh:
This is a left-over of a long past history, where you could't just edit things in place. Sure, we could remove that.

janmulder:
Ok, then I opt for removing the edit dive from the menu, as fixing the navigation (probably also very simple) is like fixing something for a now dead piece of code. Will do that maybe today or ultimately tomorrow.

PR #1674, opened 2018-09-13
Small changes

tcanabrava:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Small changes in mainwindow.cpp related to code cleanup.
My bit of study trying to remember what I did in the past.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+# brace expansion notation, does not work properly everywhere
+[*.h]
+indent_style = tab
+[*.cpp]
+indent_style = tab
+[*.txt]
+indent_size = 4
+[*.json]
+indent_size = 4
+[*.cmake]
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -103,7 +103,7 @@ extern "C" int updateProgress(const char *text)
 	return progressDialogCanceled;
 }
 
-MainWindow *MainWindow::m_Instance = NULL;
+MainWindow *MainWindow::m_Instance = nullptr;
 
 extern "C" void showErrorFromC(char *buf)
 {
@@ -113,14 +113,14 @@ extern "C" void showErrorFromC(char *buf)
 }
 
 MainWindow::MainWindow() : QMainWindow(),
-	actionNextDive(0),
-	actionPreviousDive(0),
+    actionNextDive(nullptr),
+    actionPreviousDive(nullptr),
 #ifndef NO_USERMANUAL
 	helpView(0),
 #endif
 	state(VIEWALL),
-	survey(0),
-	findMovedImagesDialog(0)
+    survey(nullptr),
+    findMovedImagesDialog(nullptr)
 {
 	Q_ASSERT_X(m_Instance == NULL, "MainWindow", "MainWindow recreated!");
 	m_Instance = this;
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -1341,93 +1341,92 @@ void MainWindow::on_actionHash_images_triggered()
 	findMovedImagesDialog->show();
 }
 
+
 QString MainWindow::filter_open()
 {
-    QString f = tr(
-        "Dive log files"
-        " (*.ssrf"
-        " *.xml"
-        " *.can"
-        " *.db"
-        " *.sql"
-        " *.dld"
-        " *.jlb"
-        " *.lvd"
-        " *.sde"
-        " *.udcf"
-        " *.uddf"
-        " *.dlf"
-        " *.log"
-        " *.txt"
-        " *.apd"
-        " *.dive"
-        " *.zxu *.zxl"
-        ");;"
-    );
+    QString f = tr("Dive log files");
+    f += " (*.ssrf"
+         " *.xml"
+         " *.can"
+         " *.db"
+         " *.sql"
+         " *.dld"
+         " *.jlb"
+         " *.lvd"
+         " *.sde"
+         " *.udcf"
+         " *.uddf"
+         " *.dlf"
+         " *.log"
+         " *.txt"
+         " *.apd"
+         " *.dive"
+         " *.zxu *.zxl"
+         ");;";
 
-	f += tr("Subsurface files") + " (*.ssrf *.xml);;";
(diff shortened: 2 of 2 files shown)


dirkhh:
YAY!!!! Tomaz is back!!!!! 
Did I mention **__YAY__**??
OK, now I have to review this, right?

dirkhh:
Hey @tcanabrava - remember that SOB thing that we do? Our new bot already told you so :-)

desktop-widgets/mainwindow.cpp
```diff
@@ -53,9 +53,9 @@
 #include "profile-widget/profilewidget2.h"
 
 #ifndef NO_PRINTING
-#include <QPrintDialog>
-#include <QBuffer>
-#include "desktop-widgets/printdialog.h"
+    #include <QPrintDialog>
+    #include <QBuffer>
+    #include "desktop-widgets/printdialog.h"
```
dirkhh, line 58: I'm not sure I like that.
But if you do it, do it with a **TAB** :-)

desktop-widgets/mainwindow.cpp
```diff
@@ -77,8 +77,10 @@
     #include "plugins/facebook/facebookconnectwidget.h"
 #endif
 
-QProgressDialog *progressDialog = NULL;
-bool progressDialogCanceled = false;
+namespace {
+    QProgressDialog *progressDialog = nullptr;
+    bool progressDialogCanceled = false;
+}
```
dirkhh: how does this help us?
Also, this commit includes `libdivecomputer` updates, hard NAK on that.

.editorconfig
```diff
@@ -0,0 +1,24 @@
...
+indent_style = tab
+[*.txt]
+indent_size = 4
+[*.json]
+indent_size = 4
+[*.cmake]
+indent_size = 4
+
+[*.md]
+trim_trailing_whitespace = false
```
dirkhh, line 24: which editors care about this type of file?

desktop-widgets/mainwindow.cpp
```diff
@@ -1166,24 +1166,17 @@ void MainWindow::on_actionViewAll_triggered()
...
-	}
 
-	static QList<int> listGlobeSizes;
-	if (listGlobeSizes.empty()) {
-		listGlobeSizes.append(lrint(appW * 0.7));
-		listGlobeSizes.append(lrint(appW * 0.3));
-	}
+    auto _lrint = [](double value) {
+        return static_cast<int>(std::lrint(value));
+    };
```
dirkhh: umm, what's that? How is this better than a macro?

dirkhh (commented):
Interesting mix of indeed nice changes that help with readability and clean up things and a couple of things that I really question.


Most importantly we need Signed-of-by lines and we need to drop the libdivecomputer change

desktop-widgets/mainwindow.cpp
bstoeger: The anonymous namespace is used to make things local to the compilation-unit. I.e. it has the same apparent effect as `static`.

Why did they introduce this strange thing, when there is `static` you may ask. Well, as Linus noted with quite some disdain, you can't declare member-functions or even classes as `static` [technically you can - but it has an entirely different meaning]. Therefore, if you don't want to export symbols of member functions, you have to use the anonymous namespace.

Whether C++ is silly for introducing this kludge and/or C is silly for using `static` for two different meanings, I'll leave to you to decide.

PS: Note that these aren't member functions, so `static` would be completely fine in this case.

desktop-widgets/mainwindow.cpp
```diff
@@ -1159,24 +1166,17 @@ void MainWindow::on_actionViewAll_triggered()
...
+
+    auto mainSizes = QList<int>{ _lrint(appH * 0.7), _lrint(appH * 0.3) };
```
bstoeger: What's in this case the advantage of `auto` over the more traditional
```
QList<int> mainSizes = { lrint(appH * 0.7), lrint(appH * 0.3) };
```
?
Seems only more characters to type..?

desktop-widgets/mainwindow.cpp
dirkhh: (a) this is a lambda function and @tcanabrava know just how much I hate this particular piece of syntax licorice (just like licorice (Lakritze) pretends to be candy but is just flat out disgusting and vomit inducing, the same is IMHO true for this abuse of letters) 
(b) which is why I suggested using a macro (or, if you absolutely have to) a static function
(c) the difference is in rounding behavior, IIRC - we introduced lrint on purpose to fix some really odd bugs

desktop-widgets/mainwindow.cpp
dirkhh: thank you - I was wondering the same but couldn't quite figure out how to ask

desktop-widgets/mainwindow.cpp
bstoeger: > the difference is in rounding behavior, IIRC - we introduced lrint on purpose to fix some really odd bugs

Yeah, `rint` returns a `double` not an `int` as I inferred from it's name.

tcanabrava:
> Hey @tcanabrava - remember that SOB thing that we do? Our new bot already told you so :-)

Sorry, I added to a couple and forgot to another couple. :)

desktop-widgets/mainwindow.cpp
tcanabrava: I blame my editor configuration. I like this because it makes sense: it's a branch inside of the if, even if it's an #if.

.editorconfig
tcanabrava: this thing makes sure that tools that understand editorconfig (quite a lot of them right now, including emacs, vim, atom, visual studio code, kate, qtcreator-editorconfig-plugin) will understand. if you don't like it I'll remove.

desktop-widgets/mainwindow.cpp
tcanabrava: 1 - I'll change this to a proper function
2 - I'll buy you Salmiakki, that's Licorice with amonium chloride <3

desktop-widgets/mainwindow.cpp
tcanabrava: no advantage whatsoever. I'm used doing that because of my work and you know how muscle memory is, I'm using the variant without auto.

desktop-widgets/mainwindow.cpp
bstoeger: Generally, I think perfectly consistent style is a bit overrated. As long as it's readable one can tolerate a bit of individuality.

But if you change it here, perhaps change it globally, so that it doesn't stand out. PS another style to consider might be:
```
#ifndef ...
#       define ...
#       define ...
#endif
```

dirkhh:
This has been merged via #1682

PR #1682, opened 2018-09-15
Tomaz small changes

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Refactored #1674
- dropped the editorconfig thing (which based on all the whitespace damage in the commits doesn't seem to really work, anyway - let's revisit that separately)
- dropped the `#include` indentation change (if we really want this (I don't think I do) then this should be done consistently everywhere and ideally through a script)
- combined related commits
- removed the lambda detour
- updated all of the commit messages
- cleaned up all the whitespace issues... :headdesk:

### Release note:

none - invisible to the user

### Documentation change:

none

### Mentions:

@tcanabrava please fix your editor setup. Please look through the commits here and see what I did with commit messages and with combining commits. This series is not bad, but the time that I spent on it is something that worries me, given the net value (as in: actual fix vs. just making things more pleasing to the developer)

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -142,15 +142,15 @@ MainWindow::MainWindow() : QMainWindow(),
 	// what is a sane order for those icons? we should have the ones the user is
 	// most likely to want towards the top so they are always visible
 	// and the ones that someone likely sets and then never touches again towards the bottom
-	profileToolbarActions << ui.profCalcCeiling << ui.profCalcAllTissues << // start with various ceilings
-				 ui.profIncrement3m << ui.profDcCeiling <<
-				 ui.profPhe << ui.profPn2 << ui.profPO2 << // partial pressure graphs
-				 ui.profRuler << ui.profScaled << // measuring and scaling
-				 ui.profTogglePicture << ui.profTankbar <<
-				 ui.profMod << ui.profNdl_tts << // various values that a user is either interested in or not
-				 ui.profEad << ui.profSAC <<
-				 ui.profHR << // very few dive computers support this
-				 ui.profTissues; // maybe less frequently used
+	profileToolbarActions = { ui.profCalcCeiling, ui.profCalcAllTissues, // start with various ceilings
+				ui.profIncrement3m, ui.profDcCeiling,
+				ui.profPhe, ui.profPn2, ui.profPO2, // partial pressure graphs
+				ui.profRuler, ui.profScaled, // measuring and scaling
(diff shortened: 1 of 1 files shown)


desktop-widgets/mainwindow.cpp
```diff
@@ -13,69 +13,80 @@
...
 
-#include "desktop-widgets/mapwidget.h"
+namespace {
+	QProgressDialog *progressDialog = nullptr;
+	bool progressDialogCanceled = false;
+	int progressCounter = 0;
 
-QProgressDialog *progressDialog = NULL;
-bool progressDialogCanceled = false;
+	int round_int (double value) {
```
bstoeger, line 85: I wonder if  this should be a global C helper-function.
There *are* a few `(int)lrint(...)` instances in different parts of the code.

desktop-widgets/mainwindow.cpp
```diff
@@ -13,69 +13,80 @@
...
-#include "core/isocialnetworkintegration.h"
-#include "core/pluginmanager.h"
 
 #if defined(FBSUPPORT)
 #include "plugins/facebook/facebook_integration.h"
 #include "plugins/facebook/facebookconnectwidget.h"
 #endif
 
-#include "desktop-widgets/mapwidget.h"
+namespace {
```
bstoeger: Is there a recommendation of `namespace` vs. `static`?

I have no preference whatsoever, so in new code I'd simply use whatever is recommended. [Currently `static`, because that is *much* more common.]

desktop-widgets/mainwindow.cpp
dirkhh: Yes, we can do this separately. This was all cleanup in one file :-)

dirkhh:
> Is there a recommendation of namespace vs. static?
> 
> I have no preference whatsoever, so in new code I'd simply use whatever is recommended. [Currently static, because that is much more common.]
> 


I can't find this comment on GitHub.
My thinking would be that in C++ files I'm ok with using a namespace like this.

/D

desktop-widgets/mainwindow.cpp
bstoeger: Should anyone care,
```
int round_int_v1(double d)
{
	return (int)lrint(d);
}

int round_int_v2(double d)
{
	return (int)nearbyint(d);
}
```
gave the same result in 10e7 operations, but v1 seems to be ca. 10% faster.

PR #1689, opened 2018-09-16
Always return status from ftdi_open when called

glance-:
Using dc_serial_open as a fallback to ftdi_open is just wrong, and will
never work, just mask the real error and introduce read herrings.

Always return status from ftdi_open when called


dirkhh:
Funny  - when trying to figure out ftdi on Android I ran into that and I have a commit just like this in my semi-abandoned branch :-)

PR #1693, opened 2018-09-17
Dive media: on import read metadata only once

bstoeger:
On import of dive media, the timestamp is read from the
metadata to check if the image belongs to the selected dives.
The pictures are then listed in a dialog.

Currently, the metadata is read twice if images are outside
of a dive: once in picture_check_valid() and if it turns
out that the picture is not valid again in picture_get_time()
to display the proper timestamp.

Even though metadata-extraction is reasonably fast, this is
a bit of an embarrassment.

Instead, read the timestamps only once in the constructor of
the dialog and from then on only used these timestamps. Keep
the timestamps in a QVector. Rename the picture_check_valid()
function to picture_check_valid_time() and pass a timestamp
instead of a filename.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a code-cleanup, which could be considered a borderline bug-fix. We read the metadata of images outside of dives twice. [Actually thrice, since once imported the metadata are read again]. In the dive-shift dialog only read the metadata once in the constructor and then access the timestamp from a `QVector`

### Changes made:

1) Rename `picture_check_valid` to `picture_check_valid_time()` and pass a timestamp instead of a filename.
2) Read metadata in the constructor of `ShiftImageTimesDialog` and remember the timestamps in a `QVector`

### Release note:

Too trivial. Not needed.
### Documentation change:

Not needed.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -3869,12 +3869,11 @@ bool dive_check_picture_time(const struct dive *d, int shift_time, timestamp_t t
 	return false;
 }
 
-bool picture_check_valid(const char *filename, int shift_time)
+bool picture_check_valid_time(timestamp_t timestamp, int shift_time)
 {
 	int i;
 	struct dive *dive;
 
-	timestamp_t timestamp = picture_get_timestamp(filename);
 	for_each_dive (i, dive)
 		if (dive->selected && dive_check_picture_time(dive, shift_time, timestamp))
 			return true;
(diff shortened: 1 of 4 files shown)


dirkhh:
This does seem like something I could skip until after 4.8.2, correct?

bstoeger:
Sure...

PR #1695, opened 2018-09-18
[TEST] Mobile: use editText instead of currentText in Suit and DiveMaster boxes

### LINKED ISSUE #1694: QML UI issue with empty data fields
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Download from dive computer in Subsurface-mobile.
Edit dives.

> there seems to be some residual information stuck in the build. I put “none” in my log for divemaster, but it came out “Parrots Landing Watersport Part” and I also wrote “none” in the bathing suit space, but it came out “bathing”. 

### Operating system:

iOS

### Subsurface version:

latest test binary

bstoeger:
A user reported problems with editing the Suit and DiveMaster fields.
Apparently, editing does not change the currentText. Without doing
a deeper analysis, simply use editText (a more proper fix might be
changing the currentIndex on editing).

(Parially?) fixes #1694

Reported-by: Mark Powell <mcpowell123@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Attempt at fixing (or at least duct taping) #1694 

@dirkhh: Could you ask the user if this fixes their problems?

### Changes made:

1) Directly use `editText` instead of conditional for Suit and DiveMaster boxes.

### Related issues:

#1694 

### Release note:

Perhaps?

### Documentation change:

No.

### DIFF
--- a/mobile-widgets/qml/DiveDetailsEdit.qml
+++ b/mobile-widgets/qml/DiveDetailsEdit.qml
@@ -110,8 +110,7 @@ Item {
 		// apply the changes to the dive_table
 		manager.commitChanges(dive_id, detailsEdit.dateText, locationBox.editText, detailsEdit.gpsText, detailsEdit.durationText,
 				      detailsEdit.depthText, detailsEdit.airtempText, detailsEdit.watertempText,
-				      suitBox.currentText != "" ? suitBox.currentText : suitBox.editText, buddyBox.editText,
-				      divemasterBox.currentText != "" ? divemasterBox.currentText : divemasterBox.editText,
+				      suitBox.editText, buddyBox.editText, divemasterBox.editText,
 				      detailsEdit.weightText, detailsEdit.notesText, startpressure,
 				      endpressure, usedGas, usedCyl ,
 				      detailsEdit.rating,


jbygdell:
What is the result if you select an already existing entry?
EditText == what the user put in to a editable combobox
CurrentText == text related to a index nr in the combobox model, i.e what you get when you select something  from the dropdown menu.

jbygdell (approved):
This seems safe to do.
In my testing editText will contain the correct data regardless if it the value is typed in by hand or selected from the list.

dirkhh:
PR still says [TEST] - so what's the next step @bstoeger ?

bstoeger:
It would be nice to get confirmation of the original reporter. But with @jbygdell's testing, this seems to be the "right thing" to do anyway (if `editText` contains the correct value, why have a conditional expression?).

As far as I understand on mobile we do not have to support old, potentially mis-behaving Qt versions, right?

dirkhh:
Correct - mobile is typically only built by me and against current Qt.

PR #1696, opened 2018-09-18
Cleanup: make surface sample in merge_one_sample() non-static

bstoeger:
The merge_one_sample() function adds a sample to the destination
dive if dives are merged. For long periods between samples at surface
depths, it adds a surface interval.

To decrease the number of global objects, make the sample structure
non-static. Of course, initialization of an on-stack structure is
slower. Therefore move it into the corresponding if. Thus, the
structure will be initialized only once per surface-interval.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a tiny, mostly irrelevant code-cleanup, which removes a static variable. This is a spin-of coming from hunting strange hangs on dive-merging in the undo-branch.

Note that you would think that the compiler would be smart enough to move the initialization of `surface` into the if. In my tests it was not.

### Changes made:

1) Make `surface` sample non-static
2) Initialize `surface` sample only if needed.

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1828,19 +1828,21 @@ static void merge_one_sample(struct sample *sample, int time, struct divecompute
 {
 	int last = dc->samples - 1;
 	if (last >= 0) {
-		static struct sample surface = { .bearing.degrees = -1, .ndl.seconds = -1 };
 		struct sample *prev = dc->sample + last;
 		int last_time = prev->time.seconds;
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
Yep, that makes sense.

PR #1707, opened 2018-09-21
Remember dive computers

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This remembers the last up to four dive computers that a user downloaded from and offers shortcut buttons to select them.
So in order to test this, you actually have to create that memory. So select a dive computer vendor/product/device combo, click download (doesn't matter if it works or fails). Then select a different combo and again click download (now called 'retry' if the previous download failed).
Now you should see two of these buttons.

### Changes made:

1) added settings magic to remember 4 more DCs
2) added  Desktop UI buttons to pick them and fill in the DC information
3) repeated the same for QML

### Release note:

Will do, once this works

### Documentation change:

Yeah, that should be documented in the manual

### Mentions:

@bstoeger @tomaz @janmulder @jbygdell

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+Desktop & Mobile: remember up to four dive computers that have been used to
+  download dives and provide easy shortcut buttons to switch between them
 Mobile: add units selection to Settings page
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
--- a/Documentation/mobile-manual.html.git
+++ b/Documentation/mobile-manual.html.git
@@ -700,6 +700,9 @@ settings). Android devices also support all Bluetooth dive computers that
 are supported in Subsurface. And both Android and iOS devices allow direct
 download of dive data from a hand full of Bluetooth LE enabled dive
 computers.</p></div>
+<div class="paragraph"><p>Use the <em>Vendor name</em> and <em>Dive Computer</em> drop-downs to pick the correct
+dive computer. When downloading from multiple different dive computers, shortcut
+buttons to easily switch between them will appear below the drop-downs.</p></div>
 <div class="paragraph"><p>The process for download is slightly different between the two OSs.
 In our testing we got the best results on iOS when the dive computer was
 in Bluetooth mode before <em>Subsurface-mobile</em> is started. On most dive
@@ -952,7 +955,7 @@ device and the font characteristics used by <em>Subsurface-mobile</em>.</p></div
 <div id="footer">
 <div id="footer-text">
 Last updated
- 2018-09-15 15:42:56 PDT
+ 2018-09-21 19:29:32 PDT
 </div>
 </div>
 </body>
--- a/Documentation/mobile-manual.txt
+++ b/Documentation/mobile-manual.txt
@@ -244,6 +244,10 @@ are supported in Subsurface. And both Android and iOS devices allow direct
 download of dive data from a hand full of Bluetooth LE enabled dive
 computers.
 
+Use the _Vendor name_ and _Dive Computer_ drop-downs to pick the correct
+dive computer. When downloading from multiple different dive computers, shortcut
+buttons to easily switch between them will appear below the drop-downs.
+
 The process for download is slightly different between the two OSs.
 In our testing we got the best results on iOS when the dive computer was
 in Bluetooth mode before _Subsurface-mobile_ is started. On most dive
--- a/Documentation/user-manual.html.git
+++ b/Documentation/user-manual.html.git
@@ -897,6 +897,18 @@ The <strong>Device or Mount Point</strong> drop-down list contains the USB or Bl
    specific dive
    computer and, in some cases, how to do the correct settings to the operating
    system of the computer on which <em>Subsurface</em> is running.
+   Some dive computers disable this drop-down as they use other means to find and
+   connect to the device. For some dive computers the data entered here is actually
+   the mount point of the USB storage device as which the dive computer identifies
+   itself.
+</p>
+</li>
+<li>
+<p>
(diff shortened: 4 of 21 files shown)


dirkhh:
Brilliant. I now tried this in my Linux VM (also with Qt 5.11.1) and there it works as expected :-/

janmulder:
Tried this on mobile (Android on device). I tested with 2 types of OSTC.
1) It does not remember the connection type. Did not inspect the code to see if this is intentional, but it seems weird from a user perspective.
2) The mobile UI is a monster ...  even 2 buttons do not fit on my 5.5" screen (portrait, in landscape it just fits). As it is no scrollable page (now), it will be impossible to hit computer nr 3 or 4.
3) I believe a delete history is needed (somewhere).

2+3 might be better with a pull down like the model/vendor/connection type above. This list could include a delete list option as well.

jbygdell:
Did a test on an android device, the FTDI connection type is not remembered but the BLE is.
I'll see what I can do about the UI/UX part for the mobile app.

dirkhh:
If live to see a screenshot from the phone where this doesn't fit in the screen @janmulder ... We may need a "shrink" option. On all of the devices I test on (seven at this point) this looks and fits fine.

It's supposed to remember connection, but I hadn't tested this with ftdi...

jbygdell:
One problem is if you have two DCs with long names (HW & Shearwater) the page will expand beyond the screens width.
overflow

jbygdell:
We can however hide the DC selector once we have a DC stored in the prefs, with the options to add more or forget the ones that are saved.

screenshot_1
screenshot_2

dirkhh:
First: would you like an option to shrink everything on your screen by like 20% - simply so it doesn't look so clunky. That would be a global option in the Settings.
Then: I like the Add DC idea in principle, but when I try to think through it, how do I show which DC has been picked? OK, I could do that with different colors, but... I'm not sure.
And outside of testing things, when would I need the ForgetDC button? When I sell / break / lose an old DC, I guess. So I'd have that in the settings as well, I think.
Let's think about this some more :-)

dirkhh:
Oh, and regardless of shrinking and what not. You are of course correct, we need to limit the maximum size of those buttons...

jbygdell:
We don't have to limit the size of the buttons if we skip the grid view and allow the buttons to wrap as needed.

dirkhh:
Cool, I'll play with that as well.

jbygdell:
Move the label to its own RowLayout above the buttons, replace the GridLayout with Flow and set the spacing to Kirigami.Units.smallSpacing and you get this.

screenshot 14

jbygdell:
The "Forget DCs" is just something I added to be able to test this. It should go into the settings page perhaps.

jbygdell:
I'm seeing some caching effects on mobile, need to look through my stashes to find out how I had fixed it.

janmulder:
A lot of progress while I was diving :smile:. I still have no luck in remembering my BT connection. I like the forget in the settings. I also notice that I cannot select FTDI when BT is on, but this might be intentionally, and not related  to this PR. All this on mobile.

dirkhh:
> I'm seeing some caching effects on mobile, need to look through my stashes to find out how I had fixed it.

If I knew what "caching effects" are, I could try to investigate :-)

I guess not remembering the connections is a real concern. So I need to play with that some more.
It would be nice to get some feedback on the desktop implementation as well, but I guess that will happen once we ship it in the next version...

dirkhh:
I have an idea why the connection wasn't matched. That's what I get for testing on the Desktop.
This last commit is largely untested, but it does make sense to me (and it compiles!) :-)

bstoeger:
```
	dive_computer_prefs_t dive_computer;
	dive_computer_prefs_t dive_computer1;
	dive_computer_prefs_t dive_computer2;
	dive_computer_prefs_t dive_computer3;
	dive_computer_prefs_t dive_computer4;
```
Ugh. Personally, I would have saved this in a `QVector<>` + `maxDiveComputers`. But as long as it works - we can fix it later. :)

janmulder:
Tried after commit 3fb7c65. While that commit looks sane, it does not work (desktop-on-mobile) for me. Further, the exact same behavior is in the desktop, and mentioned commit only deals with mobile.

jbygdell:
I have tested this on a android device, with a couple of additions I have something that does the correct thing. I'll create a PR for the needed changes.
What we then need is or someone to test this on a real iOS device as well.

janmulder:
Tried again, now this PR + #1710. On device. Setup 2 DCs from scratch (so deleted any history first). An FTDI OSTC TR is correctly restored, and an OCTC3 with BT/BLE connection string that reads "OSTC 3 <BT address>" does not have its connection string displayed on hitting the shortcut button.

jbygdell:
@janmulder what device iOS or Android?

janmulder:
Android. It seems related to the fact that my OSTC is called "OSTC 3 00:80:..." as a plain BT address is recognized correctly.

dirkhh:
@bstoeger you will need to fight the C programmers on this team until we give up and you become the maintainer :-)
And the thought of adding a non-pref data structure that suddenly gets saved with that dumpster fire that is our preferences code... also not something I wanted to contemplate.
@jbygdell for ftdi, I thought I had changed that match to be case insensitive and just a subset match. Did I get that wrong? I'm about to run to a family activity, so I won't be able to check right now... but it's that new function added to qmlmanager that does that...
And for the name + BT address case... again, that new function takes a sub string match. I'm seriously confused why that doesn't work - it did in my testing.

dirkhh:
So is this, combined with #1710 ready to merge? Or what needs to be fixed?

PR #1709, opened 2018-09-23
POC / don't merge: Scale mobile ui

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

correctly set the default font of the QML UI (and since all other measurements are derived from that, create the basis for scaling it)
add a new preference for the scaling factor
hook up a silly UI in the Settings to change it
ded. -->

### Related issues:

This addresses the "huge UI" issue mentioned in the comments of #1707 

### Additional information:

The user experience isn't great. I can't figure out how to redraw the UI. So you literally have to kill the app and restart it to get the larger font :-(
I'm sure we could do better than that.

### Release note:

needs to be added once this is sufficiently user friendly

### Documentation change:

ditto

### Mentions:

@janmulder @jbygdell

### DIFF
--- a/core/pref.h
+++ b/core/pref.h
@@ -104,6 +104,7 @@ struct preferences {
 	bool        display_invalid_dives;
 	const char *divelist_font;
 	double      font_size;
+	double      mobile_scale;
 	bool        show_developer;
 
 	// ********** Facebook **********
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -30,7 +30,7 @@ QByteArray qPrefDisplay::st_bottomSplitter;
 static const QByteArray st_bottomSplitter_default = "";
 
 bool qPrefDisplay::st_maximized;
-static bool st_maximized_default = false; 
+static bool st_maximized_default = false;
 
 QByteArray qPrefDisplay::st_geometry;
 static const QByteArray st_geometry_default = 0;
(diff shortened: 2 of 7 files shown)


janmulder:
Tried this. Seems to work (with the limitations mentioned, like dynamic redraw) on desktop-on-mobile, but my own Android build does not work at all. I see the setting in Subsurface.conf (on the device), so that's good, but the font does not resize after restart.

dirkhh:
That's frustrating. Very very frustrating given how long it took me to figure out WHY we were getting completely garbled font sizes when trying to scale things before.
I haven't had time to try this on Android (the workflow is so much slower than I typically try to develop the mobile UI as much as possible on desktop). And of course I removed all of the useful debugging info that I had added during development, so I need to re-add that and then see which part of this isn't working :-(
I assume the one added line with information (when it tells us at startup it now not only tells us the font name but also the size) shows the same size regardless of the mobile_scale setting?

janmulder:
> I assume the one added line with information (when it tells us at startup it now not only tells us the font name but also the size) shows the same size regardless of the mobile_scale setting?

Yes, correct. All data, also on the Theme page is identical regardless of the mobile_scale setting.

janmulder:
I'm making some progress on this PR. Key in what I'm trying to do is direct manipulation of the `fontMetrics.font.pointSize` on QML side. This results in immediate redraw of the UI font and derived stuff, and works both on desktop, and Android device. However, it appears that not all UI geometry is derived from the font size, so that seems an additional step needed in this PR.

PR #1710, opened 2018-09-23
Remember DCs extras

jbygdell:

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [X] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Additions to #1707 that fixes some UI issues on Android.

### Changes made:

1) Make the flow object use all of the parent objects width
2) Prevent stale data from showing when entering download DC mode.
3) Disable the forget button when no DCs are saved
4) Clear all DC related data when forgetting saved DCs.

### Related issues:

#1707

### DIFF
--- a/mobile-widgets/qml/DownloadFromDiveComputer.qml
+++ b/mobile-widgets/qml/DownloadFromDiveComputer.qml
@@ -179,7 +179,7 @@ Kirigami.Page {
 			Layout.alignment: Qt.AlignTop
 			Layout.topMargin: Kirigami.Units.smallSpacing * 2
 			spacing: Kirigami.Units.smallSpacing;
-			width: subsurfaceTheme.columnWidth - Kirigami.Units.gridUnit * 4
+			Layout.fillWidth: true
 			function setDC(vendor, product, device) {
 				comboVendor.currentIndex = comboVendor.find(vendor);
 				comboProduct.currentIndex = comboProduct.find(product);
--- a/mobile-widgets/qml/DownloadFromDiveComputer.qml
+++ b/mobile-widgets/qml/DownloadFromDiveComputer.qml
@@ -348,13 +348,8 @@ Kirigami.Page {
 
 		onVisibleChanged: {
 			comboVendor.currentIndex = comboProduct.currentIndex = comboConnection.currentIndex = -1
-			if (visible) {
-				comboVendor.currentIndex = manager.getDetectedVendorIndex()
-				comboProduct.currentIndex = manager.getDetectedProductIndex(comboVendor.currentText)
-				comboConnection.currentIndex = manager.getMatchingAddress(comboVendor.currentText, comboProduct.currentText)
-				manager.DC_vendor = comboVendor.currentText
-				manager.DC_product = comboProduct.currentText
-				manager.DC_devName = comboConnection.currentText
+			if (visible && PrefDiveComputer.vendor !== "" ) {
+				rememberedDCsGrid.setDC(PrefDiveComputer.vendor1, PrefDiveComputer.product1, PrefDiveComputer.device1)
 			}
 		}
 	}
--- a/mobile-widgets/qml/DownloadFromDiveComputer.qml
+++ b/mobile-widgets/qml/DownloadFromDiveComputer.qml
@@ -347,6 +347,7 @@ Kirigami.Page {
 		}
 
 		onVisibleChanged: {
+			comboVendor.currentIndex = comboProduct.currentIndex = comboConnection.currentIndex = -1
 			if (visible) {
 				comboVendor.currentIndex = manager.getDetectedVendorIndex()
 				comboProduct.currentIndex = manager.getDetectedProductIndex(comboVendor.currentText)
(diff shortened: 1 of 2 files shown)


mobile-widgets/qml/DownloadFromDiveComputer.qml
```diff
@@ -179,7 +179,7 @@ Kirigami.Page {
 			Layout.alignment: Qt.AlignTop
 			Layout.topMargin: Kirigami.Units.smallSpacing * 2
 			spacing: Kirigami.Units.smallSpacing;
-			width: subsurfaceTheme.columnWidth - Kirigami.Units.gridUnit * 4
+			Layout.fillWidth: true
```
dirkhh, line 182: when testing on Desktop there are situations where the drop down comboboxes extended to the right of the screen. By making this somewhat narrower I was able to keep the buttons fully visible.
But I basically agree that this is broken and that if this can happen at all, THAT needs to be fixed.

mobile-widgets/qml/DownloadFromDiveComputer.qml
```diff
@@ -348,13 +348,8 @@ Kirigami.Page {
...
 			comboVendor.currentIndex = comboProduct.currentIndex = comboConnection.currentIndex = -1
-			if (visible) {
-				comboVendor.currentIndex = manager.getDetectedVendorIndex()
-				comboProduct.currentIndex = manager.getDetectedProductIndex(comboVendor.currentText)
-				comboConnection.currentIndex = manager.getMatchingAddress(comboVendor.currentText, comboProduct.currentText)
-				manager.DC_vendor = comboVendor.currentText
-				manager.DC_product = comboProduct.currentText
-				manager.DC_devName = comboConnection.currentText
+			if (visible && PrefDiveComputer.vendor !== "" ) {
+				rememberedDCsGrid.setDC(PrefDiveComputer.vendor1, PrefDiveComputer.product1, PrefDiveComputer.device1)
```
dirkhh: but if we don't have a DC remembered, we should still try to guess one... so I'm not sure I understand why you removed that code...

dirkhh (changes requested):
Overall this looks right, but at least one of my comments seems to indicate a problem. First time users should get the DC pre-selected that we guessed based on BT data (or on FTDI Intent - not sure if this still works)

dirkhh:
Travis can't connect to GitHub... oops.

jbygdell:
> First time users should get the DC pre-selected that we guessed based on BT  data (or on FTDI Intent - not sure if this still works)

That is  something I can only test partially as my OSTC 3 presents itself as "BM+SR 3".
On the other hand when I plug in my Suunto, I get the vendor and connection type correctly specified as Suunto and FTDI.

mobile-widgets/qml/DownloadFromDiveComputer.qml
jbygdell: What I did was to add a button with the same width as you had specified, that button used ~75% of the width of my 5" phone I use for testing this. 
Unless the "Layout.fillWidth: true" is set objects that should fit side by side wrap.

dirkhh: Maybe that's another Android vs Desktop issue? Because smaller buttons would happily sit next to each other and a larger one would then stay by itself. So as written this worked as one would expect it. But I am OK with your change if this works on your phone.
I'm very frustrated that the scale thing doesn't work for you, though :-(

dirkhh:
> That is something I can only test partially as my OSTC 3 presents itself as "BM+SR 3".

You could simply add this as a detected string :-)

jbygdell:
> You could simply add this as a detected string :-)

Currently in progress.

mobile-widgets/qml/DownloadFromDiveComputer.qml
jbygdell: Partly to remove stale data.
Example scenario:
Connect DC vi a USB, download data and remove cable.
We now have a saved DC and switching to DC download will give us the correct DC selected.
If we then forget the DCs and go to DC download again we are presented with the vendor of the last used DC. 
This is wrong, if we don't have a DC paired or connected via USB we should be presented with an empty dialog.

jbygdell: Seems we will need that code for BT/BLE DCs...
I'll remove this commit as the rest is ready to be merged together with #1707 .

jbygdell: Now that I removed it I see what's missing..
Getting this to work will take some time, let's merge these as they are right now and I'll fix the automagic part separately.

dirkhh: ok

jbygdell:
@dirkhh this was not merged into master but into the rememberDCs branch.

dirkhh:
Hu?
Drat.
Sorry about that.
I'm on the phone, just boarded a plane.
Can you fix / merge into master?

jbygdell:
On Tue, Sep 25, 2018, 15:01 Dirk Hohndel <notifications@github.com> wrote:

> Hu?
> Drat.
> Sorry about that.
> I'm on the phone, just boarded a plane.
> Can you fix / merge into master?
>
No, also traveling right now.

janmulder:
Manually merged into master.

PR #1711, opened 2018-09-23
Core: split process_dives() in post-import and post-load versions

bstoeger:
process_dives() is used to post-process the dive table after loading
or importing. The first parameter states whether this was after
load or import.

Especially in the light of undo, load and import are fundamentally
different things. Notably, that latter should be undo-able, whereas
the former is not. Therefore, as a first step to make import undo-able,
split the function in two versions and remove the first parameter.

It turns out the the load-version is very light. It only sets the
DC nicknames and sorts the dive-table. There seems to be no reason
to merge dives.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a small detangling in preparation for undo of dive-import (which is a bigger project). Please check - it looks trivial enough, but I might have missed something obvious.

### Changes made:

1) Split `process_dives()` in two versions: one for load, one for import.
### Related issues:

#1528
### Additional information:

No.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -3,6 +3,8 @@
 /* core logic for the dive list -
  * accessed through the following interfaces:
  *
+ * void process_loaded_dives();
+ * void process_imported_dives(bool prefer_imported);
  * dive_trip_t *dive_trip_list;
  * unsigned int amount_selected;
  * void dump_selection(void)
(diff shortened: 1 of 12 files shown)


core/divelist.c
```diff
@@ -1315,24 +1332,17 @@ void process_dives(bool is_imported, bool prefer_imported)
 		delete_single_dive(i + 1);
 		// keep the id or the first dive for the merged dive
 		merged->id = id;
-
-		/* this means the table was changed */
-		did_merge = true;
 	}
+
 	/* make sure no dives are still marked as downloaded */
 	for (i = 1; i < dive_table.nr; i++)
```
bstoeger, line 1338: BTW: Is the `i = 1` here correct?
After all we *can* import dives into an empty log, isn't it..?

dirkhh: But since we access dive_table.dives[i-1] we really don't want to start at 0 :-)
What this loop does is that it walks the whole dive table and compares "neighboring" dives to check if they should be merged.

dirkhh (approved):
I'm not convinced that we have to make the import undoable, but the code seems correct.
I'll ping Linus for a review as well.

core/divelist.c
bstoeger: I think we're talking about different loops.

Note that if we import into an empty log, then `preexisting==0` and with
```
        for (i = preexisting; i < dive_table.nr; i++)
                dive_table.dives[i]->downloaded = true;
```
the `downloaded` flag of all dives will be set.

But with
```
        for (i = 1; i < dive_table.nr; i++)
                dive_table.dives[i]->downloaded = false;
```
it will not be reset for the first dive.

Or perhaps I'm missing something obvious.

torvalds: Yeah, I think that second for-loop should start from 0.

torvalds:
Changes look ok to me. I started wondering why DiveImportedModel::recordDives() uses 'true' for prefer_downloaded, but that's preexisting and independent of this change.

PR #1712, opened 2018-09-23
Qt ble prep

torvalds:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This cleans up the qt-ble GATT service discovery and makes it more generic. 

### Changes made:

1) it uses the WAITFOR() macro to do the "wait for event to complete", which makes for much simpler code, and can improve latency tii. 

2) it moves the special service test logic from the service discovery phase to a later "now that we've discovered all services, let's figure out which one is the preferred one".

That second change in particular makes it much easier to look at what characteristics and descriptors the different services offer, and should make it much clearer how we pick one GATT service to be the preferred communication service.

It also outputs the service information in a much more legible format along with all their characteristics and descriptor information.

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -47,11 +47,6 @@ static int debugCounter;
 	} while (timer.elapsed() < (ms));				\
 } while (0)
 
-static void waitFor(int ms)
-{
-	WAITFOR(false, ms);
-}
-
 extern "C" {
 
 void BLEObject::serviceStateChanged(QLowEnergyService::ServiceState)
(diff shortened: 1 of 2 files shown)


torvalds:
Oh, I guess I should have mentioned: I've tested this with the Shearwater Perdix AI and the Suunto EON Core on my desktop. Those are the main two BLE dive computers I have easy access to. 

The service discovery works for the Garmin Descent (with a libdivecomputer that exposes it as a BLE computer), but I haven't figured out the actual download rules, so that's just a "it makes debug messages clearer, but doesn't really change the fact that it doesn't work".

dirkhh:
Now that I actually read through the commits... yes, this makes sense. Let's see what we learn from it for the Mares (and Garmin) cases.

PR #1713, opened 2018-09-24
2 GPS related commits (cleanup and  fix)

janmulder:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix multiple run-time errors in connect call introduced in 504e912. 1) Set the proper signature of the signal. 2) make the used slot a real slot (so move it to the proper section in the header) and 3) set the proper signature for the slot.

Highly unlikely that normal users notice the run-time errors and possibly unwanted behavior, as this all deals with the subtile GPS service update threshold.

And while we are at it, as a second commit some more stripping of unused code after removal of the GPS webservice.

### Changes made:
See commits

### DIFF
--- a/core/gpslocation.cpp
+++ b/core/gpslocation.cpp
@@ -39,7 +39,7 @@ GpsLocation::GpsLocation(void (*showMsgCB)(const char *), QObject *parent) :
 	loadFromStorage();
 
 	// register changes in time threshold
-	connect(qPrefLocationService::instance(), SIGNAL(qPrefLocationService::time_thresholdChanged()), this, SLOT(setGpsTimeThreshold(int seconds)));
+	connect(qPrefLocationService::instance(), SIGNAL(time_thresholdChanged(int)), this, SLOT(setGpsTimeThreshold(int)));
 }
 
 GpsLocation *GpsLocation::instance()
--- a/core/gpslocation.cpp
+++ b/core/gpslocation.cpp
@@ -11,14 +11,6 @@
 #include <QUrlQuery>
 #include <QApplication>
 #include <QTimer>
-#include <QJsonDocument>
-#include <QJsonObject>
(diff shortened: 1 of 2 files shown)


core/gpslocation.cpp
```diff
@@ -39,7 +31,7 @@ GpsLocation::GpsLocation(void (*showMsgCB)(const char *), QObject *parent) :
 	loadFromStorage();
 
 	// register changes in time threshold
-	connect(qPrefLocationService::instance(), SIGNAL(qPrefLocationService::time_thresholdChanged()), this, SLOT(setGpsTimeThreshold(int seconds)));
+	connect(qPrefLocationService::instance(), SIGNAL(time_thresholdChanged(int)), this, SLOT(setGpsTimeThreshold(int)));
```
bstoeger, line 34: You could use the pointer-to-member function version, to get at lease *some* compile-time verification:
```
connect(qPrefLocationService::instance(), &qPrefLocationService::time_thresholdChanged, this, &GpsLocation::setGpsTimeThreshold);
```

Apart from that minor thing - lgtm.

bstoeger (approved):

core/gpslocation.cpp
dirkhh: Yes, I generally like the new syntax better. I had some problems with it a couple of times when trying to connect to overloaded slots and just recently learned that there's a special, magic helper for that 🙄 
But it's fine to not update this as part of this PR

dirkhh (approved):
NICE. And thanks for catching more of the obsolete code. I keep thinking that on one of my long flights I should just go hunt for things that we still have in the sources and no longer use. There aught to be a TON

Issue #1715, opened 2018-09-24
Desktop: BT/BLE connection is not restored from saved DCs


janmulder:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Like #1714, I see a similar issue on the desktop, but as the root cause is very likely to be different, I open a new issue for this. Starting the application, the download from DC is correctly filled from the settings. That is, it shows the last used DC (and as I have only one DC in use, that is just fine). However, after the first download a shortcut button is added. Hitting this, clears my BT connection string.

Inspecting the code, this seems obvious. 1) the connection pull down only contains the devices that are found on the computer (in my case /dev/ttyS[0-3]). Unlike mobile, there are no BT addresses added to this list. 2) as the value from the shortcut button (correctly coming from the settings) is simply not in this pull down, a findText() in this list of options from the pull down menu is always going to fail. The only way to get the BT address back is doing a new scan in the "get remote BT device) sub-window, which defies the logic behind this new functionality.

So, I'm confused. Has @dirkhh as the main developer of this new functionality ever seen this working?

### Operating system:
 What OS are you running, including OS version and the language that you are running in: up to date Arch Linux

### Subsurface version:
What version of Subsurface are you running? --> master
Does the same happen on another Subsurface version? --> na., first version in master
Are you using official release, test build, or compiled yourself? --> self
Provide Git hash if your are building Subsurface yourself. -->   b45124d5541a

### Steps to reproduce:
Simply hit a BT stored DCs button, and see that it does not populate the connection.

dirkhh:
Yes, I have - because I wrote this in a VM (I intentionally develop on many different machines, some Macs, some Linux) and so it was only tested with serial ports.
The bug report shows how dumb that was, but the fix seems rather obvious - if we don't find it, add it to the model.

janmulder:
Merged PR does the job with a small consistency remark in #1716, but no reason to keep this issue open. Closed.

PR #1716, opened 2018-09-24
Remember DC: add connection

dirkhh:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Hopefully this is the first step towards correctly working shortcut buttons.
If the connection isn't found, it's added.
This still isn't enough for BT/BLE on a Mac as there we seem to have to scan before being able to use such device names (and I don't have a Linux box with working BT adapter - so wasn't able to test it there)
Oh, and I just see that my FTDI change snuck in there, too. 
Unrelated, but I think technically correct.

### Changes made:

Simply add the remembered connection to the model of the drop down

### Related issues:

#1715 

### Mentions:

@janmulder

### DIFF
--- a/core/macos.c
+++ b/core/macos.c
@@ -104,6 +104,9 @@ int enumerate_devices(device_callback_t callback, void *userdata, unsigned int t
 	struct dirent *ep = NULL;
 	size_t i;
 	if (transport & DC_TRANSPORT_SERIAL) {
+		// on Mac we always support FTDI now
+		callback("FTDI", userdata);
+
 		const char *dirname = "/dev";
 		const char *patterns[] = {
 			"tty.*",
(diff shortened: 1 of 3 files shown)


janmulder (commented):
Functionally, this works (on Linux, cannot test on Mac). So ACK.

A small consistency remark. Devices detected by scanning read "BT reported name (BT address)", and that is also shown when entering the download from DC screen. When selecting the shortcut, only the BT address is shown. This is functionally sufficient, but a little inconsistent from a UI standpoint.

That BT reported name was added, to ease identifying your DC in noisy BT environments, which is obviously a different purpose than interfacing with it (for which the BT address is the only relevant part).

dirkhh:
Yeah, working on this blind (as I mentioned, no working Linux system - Qt/BLE is broken in other ways)... I agree that this should try to be consistent. More poking needed, but I guess we can merge this as a step in the right direction??? @janmulder

janmulder:
Merged, as indeed a relevant step to make it useful at all on desktop (and BT/BLE)

PR #1717, opened 2018-09-24
qt-ble: re-organize how we pick the GATT characteristics to read and write

torvalds:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

We used to just blindly pick "first" and "last" characteristic from the
preferred service, and that was stupid but happened to work for the dive
computers we supported.  Note that for some of them, "first" and "last"
was actually the *same* characteristic, since it could be a single one
that supported both.

However, this first/last hack definitely doesn't work for the Mares
BlueLink BLE dongle, and it's really all pretty wrong anyway.

So re-organize the code to actually look at the properties of the
characteristics.  I don't have a BlueLink to test with, but my EON Core
and Shearwater Perdix AI are still happy with this, and the code
conceptually makes a lot more sense.

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -49,17 +49,12 @@ static int debugCounter;
 
 extern "C" {
 
-void BLEObject::serviceStateChanged(QLowEnergyService::ServiceState)
+void BLEObject::serviceStateChanged(QLowEnergyService::ServiceState newState)
 {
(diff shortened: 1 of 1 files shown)


dirkhh:
The diff is a bit hard to read, but I think it all makes sense. Best way to find out is to get people to try it, I guess.

PR #1718, opened 2018-09-25
Mobile: scale UI, part 2

janmulder:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

**WARNING**: this is PR on top of #1709, that I first rebased on master, so Dirks commits are in this PR as well.

Make the font scaling work also on Android, and added a POC for the divelist (which is very simple as it already scaled very nicely). It now dynamically changes correctly (the part that was missing in #1709), and also on Android device.

### Changes made:

By manipulation the used font pointSize property, we can dynamically scale fonts and derived UI objects. At the same time, we have some logic to determine the default font, its size, etc, for example depending on screen properties. The scaling of the UI (and its font) does not need to interfere with those defaults.

However, when we want to reset the pointSize, we alter the default, so a backup of the default is needed. Ok, not al full backup, as the only thing we like to manipulate is the pointSize, to which we want to be able to return.

All this leads to these changes. A basePointSize property is added, that is initialized from the default. Due to the binding logic of the QML engine, it is not a classic initialization, but a binding between the 2 properties. We need to break that binding explicitly, so that the original PointSize is always preserved.

See commits for further details.

### Related issues:
#1709 

### Release note:
as #1709 

### Documentation change:
as #1709

### Mentions:
@dirkhh

### DIFF
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -99,7 +99,7 @@ void qPrefDisplay::disk_divelist_font(bool doSync)
 
 void qPrefDisplay::set_font_size(double value)
 {
-	if (value != prefs.font_size) {
+	if (!IS_FP_SAME(value, prefs.font_size)) {
 		prefs.font_size = value;
 		disk_font_size(true);
 
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -304,6 +304,7 @@ Kirigami.ScrollablePage {
 					wrapMode: Text.WrapAtWordBoundaryOrAnywhere
 					visible: text !== ""
 					font.weight: Font.Bold
+					font.pointSize: subsurfaceTheme.regularPointSize
 					anchors {
 						top: parent.top
 						left: dateBox.right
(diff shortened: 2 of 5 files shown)


dirkhh:
Wow. There are a few things in there that make your eyes glaze over.
The logic with breaking the binding is **so intuitive**. Not.
I'm impressed you figured this out!
This looks good to me, but I'm on the phone, just boarded a plane, so I think it's better if one of the Europeans merged this

janmulder:
@jbygdell: maybe a quick review from you on this QML hackery?

jbygdell:
The code looks sane.
I'm traveling right now so I can't test this on a hiDPI device until tomorrow.

dirkhh:
OK, finally back at a computer. Tried this on an Android device and it looks great.
Just about to upload iOS - and then I noticed that I forgot to bring my iPad on this trip.
Crud.

Are you continuing to work on the other pages so that everything scales? Just checking so we don't end up both working on the same thing (I have a few other branches that I'll happily play with instead).

Again, THANKS. This is a great step in the right direction

janmulder:
Yes, working on the download from DC page first. That is the most crowded one, with all the controls. Some changes are trivial (like the ssrfButton) but others require more trail and error (as always in QML :smile: )

PR #1723, opened 2018-09-26
Improve XML sample time parsing, and allow qt-ble to handle partial packets

torvalds:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a pull request with two fixes (or missing features, depending on how you read it) for behavior reported in the last couple of days.

The XML sample time parsing helps xslt importers from sources that describe sample times in a hh:mm:ss format, like the VMS Redbare CCR. We already supported two formats: plain seconds and 'mm:ss' format. This just extends the existing code to also support 'hh:mm:ss'.

The other fix/feature is allowing libdivecomputer backends to read partial packets, like the Mares downloader wants. Note that the Mares downloader *also* wants the reverse case, where multiple packets are joined into one big buffer, but that is handled internally by libdivecomputer with a recent update.

### DIFF
--- a/core/parse-xml.c
+++ b/core/parse-xml.c
@@ -322,16 +322,21 @@ static void temperature(char *buffer, temperature_t *temperature)
 static void sampletime(char *buffer, duration_t *time)
 {
 	int i;
-	int min, sec;
+	int hr, min, sec;
 
-	i = sscanf(buffer, "%d:%d", &min, &sec);
+	i = sscanf(buffer, "%d:%d:%d", &hr, &min, &sec);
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
Thanks for that. I do appreciate the elegance of the C++ code sometimes. This would be a lot harder to read in pure C (at least for me).
I'll wait for Travis, but both commits look good to me (and I don't see how they could break a build)

PR #1726, opened 2018-09-27
mobile, QML UI: Settings page scalability

janmulder:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A relatively big change for such a simple page. Most relevant changes are:

1) Do not use Kirigami.Header anymore. It appears that this header has word wrapping on (and we cannot override that). This is annoying on this page, as headings seems randomly be split over 2 lines, even in cases where there is more than enough room to display it on one line. And as the Kirigami.Header is just a trivial wrapper of a Text field, we can simple replace it.

2) A lot of the toplevel GridLayouts had width properties set. These are not needed (and confused my debugging code), so they are removed without any visual change. As a general rule, do not try to set properties that are not needed. In general, it can only lead to binding loops or undefined behavior.

3) Add a font size to our Theme. The step from regular to title size was a little too big.

4) And, obviously, numerous font.pointSize lines are added to actually resize the font.

### Changes made:
See above.

### Related issues:
QML interface scaling

### DIFF
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -28,31 +28,28 @@ Kirigami.ScrollablePage {
 		GridLayout {
 			id: cloudSetting
 			columns: 3
-			Layout.bottomMargin: Kirigami.Units.gridUnit
 
-			Kirigami.Heading {
+			Controls.Label {
 				text: qsTr("Cloud status")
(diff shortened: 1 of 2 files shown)


dirkhh:
I'll merge as is and we can iterate if necessary.
Thanks for continuing to work on making this better

PR #1727, opened 2018-09-27
QML UI: newer Kirigami and fallout

janmulder:
Update to the master of today, and no issues detected on mobile-on-desktop
and Android.

Only, the ugly border is back as the magic hack of 0b16b547ae2aefb1 failed
due to the patch file that errored. So that is fixed too.


dirkhh:
Cool, does this bring any fixes / improvements, or is this simply in order to keep us close to upstream (which in and of itself is a good thing)?

janmulder:
Currently, we are using Kirigami 2.2, but the current version is already 2.4, so there is definitely potential for something new (but I did not yet inspect all the commits in between). So that is the next thing I'm going to do. And for this rest, indeed, keeping up with master.

dirkhh:
The iOS build failure indicates that we need to update the qmake stuff we use there...
I'll just merge as is, figure it out locally, and push a fix.
Thanks

janmulder:
Yes, something I cannot test locally. And things that are not tested will fail :-)

dirkhh:
And interestingly, it's again the Kirigami team that didn't update their qmake files...

dirkhh:
I think I have a fix, but something else is stuck. Gotta work on something else, feel free to ignore the broken Travis, I'll fix later today.

dirkhh:
ok, fixed locally, pushed to master and the iOS build succeeded :-)

PR #1729, opened 2018-09-28
Trivial: Minor header cleanup

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Minor header cleanup in `import-csv.h`. Make one local function static. And remove redundant prototypes. These were a bit scary, because in `dive.h` they were declared `extern "C"`, whereas in `import-csv.h` they were not. I have my doubts that this is well defined.
 
### Changes made:

1) Make function static.
2) Remove redundant prototypes from `dive.h`.

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/import-csv.h
+++ b/core/import-csv.h
@@ -17,7 +17,6 @@ enum csv_format {
 
 #define MAXCOLDIGITS 10
 
-void add_sample_data(struct sample *sample, enum csv_format type, double val);
 int parse_csv_file(const char *filename, char **params, int pnr, const char *csvtemplate);
 int try_to_open_csv(struct memblock *mem, enum csv_format type);
 int parse_txt_file(const char *filename, const char *csv);
--- a/core/import-csv.h
+++ b/core/import-csv.h
@@ -17,11 +17,19 @@ enum csv_format {
 
 #define MAXCOLDIGITS 10
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 int parse_csv_file(const char *filename, char **params, int pnr, const char *csvtemplate);
 int try_to_open_csv(struct memblock *mem, enum csv_format type);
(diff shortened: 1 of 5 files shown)


dirkhh:
Declarations pulled in from C code behave the same as `extern "C"` ones in C code.
It's C++ code where this changes things.

core/import-csv.h
```diff
@@ -17,11 +17,19 @@ enum csv_format {
 
 #define MAXCOLDIGITS 10
 
+#ifdef __cplusplus
+extern "C" {
+#endif
```
dirkhh, line 22: interesting - we include this in divelogimportdialog.cpp
I wonder why this didn't cause us problems...

dirkhh: Now I know why. We don't need that include and should remove it.
Which means that technically we don't need that `#ifdef` bracing, but it's not wrong, either.

desktop-widgets/mainwindow.cpp
```diff
@@ -23,6 +23,7 @@
 #include "core/gettextfromc.h"
 #include "core/git-access.h"
 #include "core/isocialnetworkintegration.h"
+#include "core/import-csv.h"
```
dirkhh, line 26: hmm - so why did you add that here? Now I'm confused.

dirkhh: oh never mind, ignore all of my comments. I need more coffee.

dirkhh:
After thoroughly confusing myself, I now get what this actually does. Approved.

core/import-csv.h
bstoeger: It worked because the extern "C" was taken from dive.h.

PR #1732, opened 2018-09-29
Cleanup: split out free_dive() function from delete_single_dive() [Leak fix?]

bstoeger:
Currently, we can only delete dives that are indexed in the main
dive table. In the future, we will have to delete dives outside
of this table (e.g. for undo). Therefore, split out the free_dive()
function from delete_single_dive(), which takes an index into
the main dive table.

In the process, adopt the dive freeing-code from clear_dive(),
which frees more data than the code in delete_single_dive().
This potentially fixes a memory-leak.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

While looking at my undo-branch it occurred to me that the split-out `free_dive()` function frees less than the `clean_dive()` function. I'm not 100% sure, because the code is convoluted - but doesn't that mean that the dive-deletion in `UndoDeleteDive::redo()` (desktop-widgets/undocommands.c) leaks memory?

The mobile version has a `clear_dive()` call, but I couldn't convince my self that all code-paths don't leak memory.

This commit tries to unify the free-dive-memory functions and adds a `free_dive()` function in preparation of the undo-work.

In case it isn't obvious, this is a bit experimental and therefore post-release material. Therefore I assign the `priority-low` tag.
 
### Changes made:

1) Move dive-memory freeing into a local functions.
2) Split out `free_dive()` from `delete_single_dive()`

### Related issues:

None.
### Additional information:

None.
### Release note:

None.
### Documentation change:

None.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -528,11 +528,7 @@ static void copy_tl(struct tag_entry *st, struct tag_entry *dt)
 		*_dptr = 0;                              \
 	}
 
-/* copy_dive makes duplicates of many components of a dive;
- * in order not to leak memory, we need to free those .
- * copy_dive doesn't play with the divetrip and forward/backward pointers
- * so we can ignore those */
-void clear_dive(struct dive *d)
+static void free_dive_structures(struct dive *d)
 {
 	if (!d)
 		return;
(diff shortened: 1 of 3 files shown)


core/dive.c
```diff
@@ -528,14 +528,8 @@ static void copy_tl(struct tag_entry *st, struct tag_entry *dt)
...
 
-/* copy_dive makes duplicates of many components of a dive;
- * in order not to leak memory, we need to free those .
- * copy_dive doesn't play with the divetrip and forward/backward pointers
- * so we can ignore those */
-void clear_dive(struct dive *d)
+static void free_dive_structures(struct dive *d)
 {
-	if (!d)
-		return;
```
dirkhh: why not leave this in there? If you call free_dive_structures(NULL) bad things happen. We may not be at risk of doing this today, but still... it seems cheap insurance...

dirkhh (approved):
small suggestion, but really fine either way :-)

core/dive.c
bstoeger: Yes, that makes sense...

With this perhaps really wait until release? Seems correct, but nevertheless a bit scary.

dirkhh: hehe... you're not paying attention. The release is out :-)

bstoeger: Oops. I admit I didn't follow too closely as there was too much activity.

dirkhh: No problem - I'm just giving you a hard time.
I really want BLE on Windows to work... and I think that's where most of the activity comes from.

PR #1736, opened 2018-09-29
Cleanup: don't derive UserSurveyServices from WebServices

bstoeger:
UserSurveyServices derives from WebServices and therefore has
to define three pure virtual functions [startDownload(),
startUpload(), buttonClicked()] as no-ops. Interestingly,
a comment in the header says
	"need to declare them as no ops or Qt4 is unhappy"
which is of course not true as these functions are not
declared by Qt.

There seems to be no point in deriving from WebServices,
therefore don't do it. These function definitions can then
be removed.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Another very minor code cleanup, see commit-message.
### Changes made:

1) Don't derive UserSurveyServices from WebServices
2) Unvirtualize downloadTimedOut
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/desktop-widgets/subsurfacewebservices.cpp
+++ b/desktop-widgets/subsurfacewebservices.cpp
@@ -793,17 +793,16 @@ void DivelogsDeWebServices::buttonClicked(QAbstractButton *button)
 	}
 }
 
-UserSurveyServices::UserSurveyServices(QWidget *parent, Qt::WindowFlags f) : WebServices(parent, f)
+UserSurveyServices::UserSurveyServices(QWidget *parent, Qt::WindowFlags f) : QDialog(parent, f)
 {
-
 }
 
-QNetworkReply* UserSurveyServices::sendSurvey(QString values)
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
Both of these commits are at the edge of my comfort zone... not in that I'm not comfortable with them, but that I'm not comfortable I understand enough to have been able to make them myself.
With that caveat, they look reasonable to me

neolit123 (approved):
LGTM.
we no longer support Qt4, too.

bstoeger:
> Both of these commits are at the edge of my comfort zone... not in that I'm not comfortable with them, but that I'm not comfortable I understand enough to have been able to make them myself.

I'm also not comfortable with the first one. But I submitted a survey and it still worked. :)

The second one should be obvious. If a function is not overridden, there can be no point in it being virtual.

> we no longer support Qt4, too.

The comment was completely bogus. This has nothing to do with Qt.

PR #1738, opened 2018-09-30
Mobile, Remember DCs: the final part

jbygdell:

- [ ] Bug fix
- [X] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Since we now keep track of up to 4 dive computers the logic when entering download mode needs to change. Instead of displaying the last used DC we need to match the connected DC against the ones we have stored. If we have a match against a DC that have been used before the corresponding button is marked as selected.
Some USB cables only supply Vendor name and not the device, in that case we still try to match against the previously used DCs. If we find a possible match we select that device.

### Related issues:

#1707 #1710

### DIFF
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -291,16 +291,6 @@ QStringList DCDeviceData::getProductListFromVendor(const QString &vendor)
 
 int DCDeviceData::getMatchingAddress(const QString &vendor, const QString &product)
 {
-	if (qPrefDiveComputer::vendor() == vendor &&
-	    qPrefDiveComputer::product() == product) {
-		// we are trying to show the last dive computer selected
-		for (int i = 0; i < connectionListModel.rowCount(); i++) {
-			QString address = connectionListModel.address(i);
-			if (address.contains(qPrefDiveComputer::device()))
-				return i;
-		}
-	}
-
 	for (int i = 0; i < connectionListModel.rowCount(); i++) {
 		QString address = connectionListModel.address(i);
 		if (address.contains(product))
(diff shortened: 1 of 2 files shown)


jbygdell:
Not the first time that happens.

dirkhh:
I broke Windows Travis by adding libftdi1 for the windows build. I'll fix that later today.
I'll be away from my computer for the next few hours, though

dirkhh:
@janmulder since you currently work on the QML UI as well, would you provide a review?
I'm working on fixing the Windows Travis thing in the meantime :-)

janmulder (commented):
Ok, tried this (rebased on current master). The code change looks sane, but I'm a little hesitant  to give a plain ACK.

What does to concept "the connected DC" mean in case of BT/BLE devices? In think that most users open the download from DC screen, select the one that they are **going** to download, and hit the download button. And just before they hit the download button, they put the DC in transfer mode. In other words, there is no connected DC at opening of the download screen.

I see in my case (OSTC3 with BT/BLE, no FTDI) that it correctly sets vendor and connection (and the known issue that for libdc, this version of the OSTC3 needs to be download as OSTC Plus, as the OSTC3 is FTDI only). So I now have to change the product field before I can download, where before this PR (last chosen instead of connected, whatever connected means), I had to select the connection.  Obviously, this is not a problem is this PR but libdc/HW/ssrf weirdness wrt. DC type and it capabilities.

My final conclusion would be: merge as is.

dirkhh:
There are so many permutations of all this. Thanks for the detailed analysis. I'll follow your advice.

jbygdell:
The issue with the early BT/BLE OSTC3s we can fix by label them as OSTC Plus.

@janmulder in your case you should be able to use the quick select button that will appear once you have performed a download.

janmulder:
> in your case you should be able to use the quick select button that will appear once you have performed a download.

Interestingly, that does not work. Hitting the quick select button clears the connection. Notice that a fake quick select button (for a device that I do not have, so with a fake BT connection address) is perfectly recalled. As this fake one is just a xx:xx:xx:xx address and my OSTC3 address is like "OSTC 3 xx:xx:xx:xx" it feels like some issue with substring compare. But have not tracked down where.

dirkhh:
Yes, I have noticed that I messed something up with BT based device recall on mobile.
It worked when I tested it, but then I must have "improved" it. :-(
I'm working on too many things at once right now, so I don't think I'll have time to look into THAT any time soon.

jbygdell:
Well, I started it so I might as well continue.

jbygdell:
@bstoeger  and @janmulder have a look at #1744 and report if it fixes the issues you see.

PR #1739, opened 2018-09-30
QML UI: newer Kirigami and fallout (revisited)

janmulder:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Go to the current master of Kirigimi, and deal with the fallout. The first attempt in #1727 resulted in a error in the mobile app, and as we were very close the a release, the safest was to revert. Later, it appeared that the issue could be resolved with 1 line of code. So that is added now. Also, as Kirigami fixed the qmake build issue, we now do not need to dirty-patch that at our end any more.

Further, bump Kirigami includes to 2.4. Testing shows no artifacts. 

### Changes made:
See commits

### Related issues:
#1727


dirkhh:
OK, let's try this again and see how it looks.
Thanks for working on this!

PR #1742, opened 2018-10-01
[RFC] Import: import Poseidon MkIV logs via DiveLogImportDialog

bstoeger:
Poseidon MkIV logs (.txt) were special cased in MainWindow.cpp,
which led to a user-interface inconsistency. In some cases
[user chooses ".txt" (non-Poseidon) and ".csv"], *two*
import-dialogs were shown.

Move handling of Poseidon MkIV logs into DiveLogImportDialog.
There are already other "special" cases handled in this dialog.

At the moment, this shows the first 10 depth-values, which is
kind of useless, as this will all be at surface level. We
might think about something more useful.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This moves the handling of Poseidon logs into `DiveLogImportDialog` in an attempt to make the user-interface more consistent and also the code more consistent. This will ease the detangling of the import code (think undo, don't unnecessarily import dive-sites, remove divesite-uuids).

Currently, the display is quite useless (first 10 depth values); I'm all ears for other ideas.

### Changes made:

1) Move handling of Poseidon into `DiveLogImportDialog`
### Related issues:

None.
### Additional information:

None.
### Release note:

Will do if code is accepted.
### Documentation change:

Will check if code is accepted.
### Mentions:

@mturkia

### DIFF
--- a/desktop-widgets/divelogimportdialog.cpp
+++ b/desktop-widgets/divelogimportdialog.cpp
@@ -15,7 +15,24 @@ static QString subsurface_index = "subsurface/csvindex";
 
 #define SILENCE_WARNING 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ""
 
-const DiveLogImportDialog::CSVAppConfig DiveLogImportDialog::CSVApps[CSVAPPS] = {
+struct CSVAppConfig {
+	QString name;
+	int time;
+	int depth;
+	int temperature;
+	int po2;
+	int sensor1;
+	int sensor2;
+	int sensor3;
+	int cns;
+	int ndl;
+	int tts;
+	int stopdepth;
+	int pressure;
+	int setpoint;
+	QString separator;
+};
+static const CSVAppConfig CSVApps[] = {
 	// time, depth, temperature, po2, sensor1, sensor2, sensor3, cns, ndl, tts, stopdepth, pressure, setpoint
 	// indices are 0 based, -1 means the column doesn't exist
 	{ "Manual import", SILENCE_WARNING },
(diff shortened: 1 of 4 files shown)


mturkia (commented):
Seems reasonable but I have not tested this yet. Also the Windows build seems to fail on missing library or something. (I don't see how this patch would cause the failure)

dirkhh:
The Windows build is fixed in master.
But now we see the failed Max builds where autotools mis-recognizes available functions...
I'll try to remember how I fixed that locally...

bstoeger:
> The Windows build is fixed in master.

I reckon that means that I should rebase this on master?

@mturkia: Could you please also check #1750? Sorry for being a nuisance, but I'm a bit nervous because my local branches diverge more and more from master.. :-P

dirkhh:
I'm trying my fix for the Mac build error.

Having your branch diverge is not a good sign.
How much of this should be merged upstream?

bstoeger:
Rebased on master, edited commit message (MkIV -> MkVI) and added two very minor commits.

The second commit is a minor code-cleanup, which boils down to taste. In my opinion, the resulting code is simpler (one constant less, no sentinel), but if you want to keep the old version, that's fine by me.

The third commit makes the import-dialog modal (according to @neolit123 this is already the case on Windows), which allows removal of the `deleteLater()` kludge.

Concerning diverging branches, what scares me most is how my own branches diverge. Notably import and undo both mess with dive-merging in a different way. This PR seems pretty low risk to me  (how many users actually import from this Poseidon thing?) and would allow progress on the import-branch. Having the latter upstream would allow me to unify the two diverging branches.

dirkhh (approved):
This looks fine.
We should try to get a Poseidon user to test this :-)

mturkia:
I tested one of the MKVI sample logs that I have and the logs looked to be the same. This test was only visual, so not entirely sure the imports are identical. Having a Poseidon user to test it would be great, as he&she would actually know what is the original dive.

PR #1744, opened 2018-10-01
Mobile correctly match BLE DCs

### LINKED ISSUE #1714: Mobile: BT/BLE connection is not restored from saved DCs
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
See below.

### Operating system:
What OS are you running, including OS version and the language that you are running in: Android 8.1, LineageOS 15.1, US language.
What device are you using? Wileyfox Swift

### Subsurface version:
What version of Subsurface are you running? Mobile 2.1.4 (4.8.2.39) from play store beta channel
Does the same happen on another Subsurface version? first version with this feature
Are you using official release, test build, or compiled yourself? official beta

### Steps to reproduce:
See screencast. First rename to mp4, as that is not allowed to upload here.
[untitled.mp4.txt](https://github.com/Subsurface-divelog/subsurface/files/2410382/untitled.mp4.txt)

1) goto Download page. In my case the Device/Vendor/Connection are filled in, and at the start of the video, no saved DCs are present.
2) Hit download. I did not connect a DC, so that times out. After timeout, the new button is presented.
3) Hit the new button, and see that in my mp4, the connection is cleared.
4) Select the connection again, and fill in the wanted connection
5) Hit the new button again, and see that the connection is cleared.

### Current behavior:
See above

### Expected behavior:
Connection is restored when hitting the button.

jbygdell:

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Improve the matching logic for DCs that the app remembers.
### Changes made:

1) Enable matching of previously used USB DCs.
2) Correctly match BT/BLE address sub-string.
3) Save Vendor + Product + Address when remembering DCs
4) Reset the "Download" button when changing DC.
### Related issues:

Fixes #1714 
### Mentions:

@janmulder test this and see if it solves your issues.

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -18,8 +18,7 @@ static dc_descriptor_t *getDeviceType(QString btName)
 
 	if (btName.startsWith("OSTC")) {
 		vendor = "Heinrichs Weikamp";
-		if (btName.mid(4,2) == "3#") product = "OSTC 3";
-		else if (btName.mid(4,2) == "3+") product = "OSTC 3+";
+		if (btName.mid(4,1) == "3") product = "OSTC Plus";
 		else if (btName.mid(4,2) == "s#") product = "OSTC Sport";
 		else if (btName.mid(4,2) == "s ") product = "OSTC Sport";
 		else if (btName.mid(4,2) == "4-") product = "OSTC 4";
--- a/core/connectionlistmodel.cpp
+++ b/core/connectionlistmodel.cpp
@@ -53,6 +53,6 @@ void ConnectionListModel::removeAllAddresses()
 
 int ConnectionListModel::indexOf(QString address)
 {
-	const QRegExp re(address, Qt::CaseInsensitive);
+	const QRegExp re(".*" + address + ".*", Qt::CaseInsensitive);
 	return m_addresses.indexOf(re);
 }
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -27,12 +27,12 @@ static QString str_error(const char *fmt, ...)
 
 static void updateRememberedDCs()
 {
-	QString current = qPrefDiveComputer::vendor() + " - " + qPrefDiveComputer::product();
+	QString current = qPrefDiveComputer::vendor() + " - " + qPrefDiveComputer::product() + " - " + qPrefDiveComputer::device();
 	QStringList dcs = {
-		qPrefDiveComputer::vendor1() + " - " + qPrefDiveComputer::product1(),
-		qPrefDiveComputer::vendor2() + " - " + qPrefDiveComputer::product2(),
-		qPrefDiveComputer::vendor3() + " - " + qPrefDiveComputer::product3(),
-		qPrefDiveComputer::vendor4() + " - " + qPrefDiveComputer::product4()
+		qPrefDiveComputer::vendor1() + " - " + qPrefDiveComputer::product1() + " - " + qPrefDiveComputer::device1(),
+		qPrefDiveComputer::vendor2() + " - " + qPrefDiveComputer::product2() + " - " + qPrefDiveComputer::device2(),
+		qPrefDiveComputer::vendor3() + " - " + qPrefDiveComputer::product3() + " - " + qPrefDiveComputer::device3(),
+		qPrefDiveComputer::vendor4() + " - " + qPrefDiveComputer::product4() + " - " + qPrefDiveComputer::device4()
 	};
 	if (dcs.contains(current))
 		// already in the list
--- a/mobile-widgets/qml/DownloadFromDiveComputer.qml
+++ b/mobile-widgets/qml/DownloadFromDiveComputer.qml
@@ -197,6 +197,7 @@ Kirigami.Page {
 							break
 						}
 					}
+					download.text = qsTr("Download")
 
 				}
 			}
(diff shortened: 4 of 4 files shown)


janmulder:
Ok, I believe this could work (with only reading the code). But at more or less at the same time, I also have a 1 line fix :-) 

```
--- a/core/connectionlistmodel.cpp
+++ b/core/connectionlistmodel.cpp
@@ -53,6 +53,6 @@ void ConnectionListModel::removeAllAddresses()
 
 int ConnectionListModel::indexOf(QString address)
 {
-       const QRegExp re(address, Qt::CaseInsensitive);
+       const QRegExp re(".*" + address + ".*", Qt::CaseInsensitive);
        return m_addresses.indexOf(re);
 }
```
Obviously, this does not handle the OSTC 3 vs OSTC Plus confusion, but I think that that is something different. This as a replacement for your commit a4625d0.

To be sure: the funky BT address (BM+SR) seems to me a non-production version (ie . no serial number). Is this correct? If this is true, I think we should not have this in our code (but convenient for your local testing).

And the automagic renaming of OSTC3 over BT/BLE to OSTC Plus ... yes, that looks fine. I'm not familiar with FTDI, but its clear that a FTDI OSTC3 shall never be renamed to OSTC Plus.

Now a test ...

janmulder:
... and the test: fail. It now fully correctly fills my OSTC 3 as OSTC Plus, including populated connection. So that is good. But my (fake) Shearwater-Predator with connection string xx:xx:xx:xx fails. The reason is obvious. Comparing with the model name included assumes that the discovery constructed the proper model name (which is not the case for my fake, non existing, Predator). I believe this is fundamentally dangerous, as we cannot assume that we can always derive the model name from the BT name it advertises.

So, I like my 1-liner better.

Conclusion: 
commit e9b0363 => looks like development code 
commit 8f46aec => LGTM
commit a4625d0 => see above.

jbygdell:
The first commit is indeed a workaround for my OSTC 3+, a very, very early model.
The thing is that if we detect something as a DC it automatically gets the model prepended to the address.
I'll cook up a new version during the evening.

janmulder:
Further corner case: my Android device reports my OSTC 3 as BT and BLE able (which is correct). BLE is sitting under 1 shortcut button, BT under a next. BLE LE: style address is not restored.

jbygdell:
@janmulder if you insert this as line 184 of DownloadFromDiveComputer.qml
`console.log( eval("PrefDiveComputer.product" + i) + ":" + eval("PrefDiveComputer.device" + i).toUpperCase())`
Then we might get a hint of why it does not work.

jbygdell:
Well I found some other bugs that need fixing.
But one thing that I can't reproduce is the issue @janmulder is seeing with two shortcut buttons to the same DC.

janmulder:
> But one thing that I can't reproduce is the issue @janmulder is seeing with two shortcut buttons to the same DC.

Correct that you could not reproduce ... I cheated here by manually adding a BLE one next to the BT one in the settings file on disk/device. It appears that from the UI you cannot add a duplicate for the same vendor/product combination. The question is here: is this a valid use case (more than 1 connection per DC)? 

What brings me to the question: how are things supposed to work with the remembered DCs for users that have 2 identical DCs and like a shortcut for both? Currently, this cannot work as the "key" to lookup the connection data is vendor/product. This seems a valid use case.

Further, I notice that "forget" does not clear the data in the settings file. Not sure this is wrong, but it will pull up data from before a "forget".

jbygdell:
> What brings me to the question: how are things supposed to work with the remembered DCs for users that have 2 identical DCs and like a shortcut for both? Currently, this cannot work as the "key" to lookup the connection data is vendor/product. This seems a valid use case.

Yes, and we always return the first match we find. which for dualstack DCs mean we always return BT-classic. 
Should we turn this around and match vendor + address to a product?

> Further, I notice that "forget" does not clear the data in the settings file. Not sure this is wrong, but it will pull up data from before a "forget".

Yepp, I noticed that to, I have the fixes in my repo.

janmulder:
> Yes, and we always return the first match we find. which for dualstack DCs mean we always return BT-classic.
> Should we turn this around and match vendor + address to a product?

Or maybe even further. Why not match from address to vendor + product? Basically, the address is the only unique key here (BT/BLE case, not FTDI). That could (very likely) even support multiple buttons per DC.

jbygdell:
First of we need to start saving the device address, now I can actually get two buttons to the same DC with different addresses.

janmulder:
Tried this with the last commit added (rebased on master). All scenarios I could think of seem to work.

dirkhh:
Conceptually, outside the desire of wanting to test BLE on purpose, I cannot imagine a scenario where a regular user would want to download over (much slower) BLE if BT is available.
BTW: dual stack dive computers appear to be extremely rare. The relatively short-lived Petrel 2 dual stack, and I think a couple of OSTC flavors. No other dive computer I'm aware of supports both.

jbygdell:
This was more of a way to not create two quick button to the same DC, we should use the reported address but since it is the same DC we should not add a new quick button for it.

jbygdell:
Anyway I'll clean up what's in here and generate a proper PR for this.

dirkhh:
Cool - and if you rebase on latest master (as of 2 minutes ago) you'll even get meaningful Travis results :-)

dirkhh (approved):
looks good

PR #1747, opened 2018-10-02
Mobile: simpify startup logic

janmulder:

- [x] Bug fix (visual only)
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When starting the mobile app, I noticed a short display of an empty page with title "Cloud credentials" just before showing the divelist. Simply a not nice visual effect.

This commit simplifies some logic and resolves this. 

### Additional information:
As the code in this part is fragile, this is tested for normal and clean startup of the app, switching credentials, from no cloud to valid account (which even nicely imports the no cloud dives: this surprised me as I have never seen this working).

### DIFF
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -334,12 +334,8 @@ Kirigami.ScrollablePage {
 		visible: opacity > 0
 		Behavior on opacity { NumberAnimation { duration: Kirigami.Units.shortDuration } }
 		function setupActions() {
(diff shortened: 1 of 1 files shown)


janmulder:
Well, not sure what is to fix on the no cloud -> cloud transition. It imported my 1 dive from the no cloud repo correctly into the cloud account I opened. 

Oh wait ... reading again. You fixed it already :-)

PR #1752, opened 2018-10-04
[Core] Rework dive-importing interface

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This changes the dive-importing interface such that dives are imported into a distinct table and then passed down to `process_imported_dives()`. For dive-computer downloads this was already effectively the case (downloads were into a downloadTable), this PR just finalizes the interface. This allows removal of the `preexisting` variable of `struct dive_table` and the `downloaded` flag of `struct dive`, which had a weird hybrid meaning anyway.

There is a functional change, which is described in the commit message as:
```
    This actually introduces (at least) two functional changes:
    1) If a new dive spans two old dives, it will only be merged to the
       first dive. But this seems like a pathological case, which is of
       dubious value anyway.
    2) Dives unrelated to the import will not be merged. The old code
       would happily merge dives which were not even close to the
       newly imported dives. A surprising behavior.
```

The uemis downloaded used (misused?) dive->downloaded to mark deleted dives. This is changed to use (misuse?) dive->hidden_by_filter.

This work is in context of undo and removal of divesite uuids.

### Changes made:

1) Pass dive_table to process_imported_dives.
2) Remove preexisting from struct dive_table.
3) Remove dive->downloaded.

### Release note:

Only user visible in pathological cases.
### Documentation change:

Only user visible in pathological cases.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -2759,11 +2759,6 @@ static int match_dc_dive(const struct divecomputer *a, const struct divecomputer
 	return 0;
 }
 
-static bool new_without_trip(const struct dive *a)
-{
-	return a->downloaded && !a->divetrip;
-}
-
 /*
  * Do we want to automatically try to merge two dives that
  * look like they are the same dive?
@@ -2802,16 +2797,13 @@ static int likely_same_dive(const struct dive *a, const struct dive *b)
 	    same_string(b->dc.model, "manually added dive"))
 		return 0;
 
-	/* Don't try to merge dives with different trip information */
-	if (a->divetrip != b->divetrip) {
-		/*
-		 * Exception: if the dive is downloaded without any
-		 * explicit trip information, we do want to merge it
-		 * with existing old dives even if they have trips.
-		 */
-		if (!new_without_trip(a) && !new_without_trip(b))
-			return 0;
-	}
+	/* Don't try to merge dives with different trip information
+	 * Exception: if the dive is downloaded without any
+	 * explicit trip information, we do want to merge it
+	 * with existing old dives even if they have trips.
+	 */
+	if (a->divetrip != b->divetrip && b->divetrip)
+		return 0;
 
 	/*
 	 * Do some basic sanity testing of the values we
@@ -2843,6 +2835,11 @@ static int likely_same_dive(const struct dive *a, const struct dive *b)
  * This could do a lot more merging. Right now it really only
  * merges almost exact duplicates - something that happens easily
  * with overlapping dive downloads.
+ *
+ * If new dives are merged into the dive table, dive a is supposed to
+ * be the old dive and dive b is supposed to be the newly imported
+ * dive. If the flag "prefer_downloaded" is set, data of the latter
+ * will take priority over the former.
  */
 struct dive *try_to_merge(struct dive *a, struct dive *b, bool prefer_downloaded)
 {
@@ -3328,16 +3325,20 @@ int count_dives_with_suit(const char *suit)
  *     two must have a different start time, and "offset" is the relative
  *     time difference between the two.
  *
- * (a) two different dive computers that we might want to merge into
+ * (b) two different dive computers that we might want to merge into
  *     one single dive with multiple dive computers.
  *
  *     This is the "try_to_merge()" case, which will have offset == 0,
  *     even if the dive times might be different.
+ *
+ * If new dives are merged into the dive table, dive a is supposed to
+ * be the old dive and dive b is supposed to be the newly imported
+ * dive. If the flag "prefer_downloaded" is set, data of the latter
+ * will take priority over the former.
  */
 struct dive *merge_dives(struct dive *a, struct dive *b, int offset, bool prefer_downloaded)
 {
 	struct dive *res = alloc_dive();
-	struct dive *dl = NULL;
 
 	if (offset) {
 		/*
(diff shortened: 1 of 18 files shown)


bstoeger:
Damn, that was stupid.

Currently I can't compile mobile:
```
bs@laptop:~/src/subsurface/build-mobile$ make
-- GCC optimization flags: -O2
-- Building Subsurface-mobile requires BT support
-- system name Linux
-- Configuring done
CMake Error at CMakeLists.txt:307 (add_executable):
  Cannot find source file:

    mobile-widgets/qml/kirigami/src/scenepositionattached.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
  .hxx .in .txx


-- Generating done
-- Build files have been written to: /home/bs/src/subsurface/build-mobile
Makefile:1324: recipe for target 'cmake_check_build_system' failed
make: *** [cmake_check_build_system] Error 1
```
Will investigate.

bstoeger:
Nevermind, running `build.sh` seems to have solved the problem.

janmulder:
Yes, use the `scripts/build.sh -both` once in a while. You are missing (at least) Kirigami updates that get pulled in when using this script.

janmulder:
Found a bug, but did not try to diagnose (and/or solve) it further (yet).

Imported a massive 110 dive set of dives, that I've sitting around as a dump from my OSTC, and which I can import using the libdivecomputer simulator. As most of these dives are already in my logbook, an import should duplicate the profile, and I should see in the bottom of the profile 1 of 2 (and the possibility to see the individual profiles, that are obviously identical in this case). All this works.

But, as I have trips, and in these trips I merged and deleted dives (in the past), things get more complicated. For example, a trip with 11 dives, shows after import 19 dives, and 9 shown **in the header**. This is definitely a bug in this PR as a replay on master nicely shows 11 dives for this trip in the header with the same imported set. I say explicitly, in the header, as the dives itself seem correct (the bug manifests itself in the trip header data). The 19 is clear where that is coming  from. There are 19 profiles in the set of 11 dives: so it counts profiles now, instead of dives which is wrong. Where the 9 shown is coming from ... no idea at this point. There is no filter active so 19 is to be expected.

janmulder:
These 2 files demonstrate the issue. Open the test.ssrf and import from file  test-import.ssrf

[test-import.ssrf.txt](https://github.com/Subsurface-divelog/subsurface/files/2446957/test-import.ssrf.txt)
[test.ssrf.txt](https://github.com/Subsurface-divelog/subsurface/files/2446958/test.ssrf.txt)

bstoeger:
Thanks! Test cases make this so much easier. Unfortunately, I will not be able to handle this today.

The behavior does not come as a big surprise, as we parse into a distinct table, but it is quite possible that the dives are added to the global trip-list. If they are not properly removed from there, on merge this could lead to the observed behavior.

core/dive.c
```diff
@@ -2792,6 +2787,9 @@ static bool new_without_trip(const struct dive *a)
  * the dates are *too* far off the user will have to join two
  * dives together manually. But this tries to handle the sane
  * cases.
+ *
+ * Dive a and b are supposed to be the previously existing and a
```
mturkia: This comment is confusing. Do you mean that (a) already exists and (b) is new, or vise versa? Or this could even mean that a and be are already existing. Removing the 'a' from the 'a newly imported dive' might make it more clear, but I would suggest being still more explicit on the comment.

janmulder:
And found a different bug in this PR. Start with an empty (0 dives) logbook. Do import from DC, import some dives. This all works. Now hit ok, to transfer them to the (still empty) logbook. No dives are transferred. Notice that import in an empty logbook from a file works just fine.

janmulder:
And the reason for bug 2 is obvious (file downloadfromdivecomputer.cpp, line 510). 

```
dive = get_dive(dive_table.nr - 1);
	if (dive != NULL) {
```

This code makes (some) sense when importing dives in 1 table, but makes no sense al all when separating things.

bstoeger:
Ooops. Bug #2 is already in master, isn't it? In any case, looks trivial to fix.

janmulder:
No, bug 2 is not in master. Which very surprised me when I was only code reading.

bstoeger:
Yes I see - it calls `record_dive`, which add the dive to the global table. master is fine.

bstoeger:
Pushed a quickfix for bug#2 (untested) and updated the comment objected by @mturkia. Will look at bug #1 later.

janmulder:
I can confirm that bug 2 is fixed.

bstoeger:
Ok, bug#1 is also clear: when we merge dives, we have to remove the old dive from the trip and add the new one. Added a quick-fix. This has some code-duplication, I will factor this out into a (temporary) function.

bstoeger:
Split out common code into a new function. Code's not pretty, but for the moment should fulfill its purpose.

janmulder:
And also here: fix is confirmed.

That said. I found another merge related bug, but I have to admit that this is not (all) introduced in this PR. I say "all" because this PR seems to change something in the values.  So maybe its better to open an issue for this, but merging is a significant part of this PR, I just mention it here.

The bug is: merging messes up the surface interval values. Look at the intervals of the dives in the files I posted above, and see what happens after merge. Suddenly we have (some) negative intervals. I did not actually verify the values, but those negative ones are definitely wrong.

janmulder:
>  I found another merge related bug

Never mind. While there might something strange going on: its not a merge related bug. And its even unsure that it is a bug at all. I use a re-import, where the original has been modified (with previous merges and delete of bogus dives that I did not want to show up as part of any dive). When importing the raw DC data again, we end up with (sometimes partially) overlapping dives. In this setting, the computation of surface interval gets confused, as there is definitely some (implicit?) assumption about the end of the previous dive is before that start of the current one.

janmulder:
I agree that there could be a more sane way of handling surface intervals of overlapping dives. However, I cannot think of a relevant use case. In other words: something very low priority. Its definitely not related to this PR, as I see the same strange negative surface intervals in the current master (after importing an overlapping set of dives).

torvalds:
This all looks fine to me too.

dirkhh:
merged manually

PR #1754, opened 2018-10-04
Cleanup: don't access downloadTable in libdivecomputer.c

bstoeger:
If no dives were downloaded in do_libdivecomputer_import(), an
error message would be produced. To check for downloaded dives,
the function would access the global downloadTable instead of
the actual table the dives are imported to (at the moment the
same - but the interface allows for a different table).

Move the error-creation to the caller to avoid this situation.
An alternative option would be to check the actual table the
dives were supposed to be downloaded to. But from a program-logic
point of view "no dives" does not seem like an error condition.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Another trivial cleanup, which removes an access to the global downloadTable. A different solution is schematized in the commit message [test the actual table]. Tested it briefly and I still get the "no dives downloaded" message.
### Changes made:

1) Move creation of "no dives downloaded" error from libdivecomputer.c to the caller.
### Additional information:

None.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -91,6 +91,8 @@ void DownloadThread::run()
 		error = str_error(errorText, internalData->devname, internalData->vendor, internalData->product);
 		qDebug() << "Finishing download thread:" << error;
 	} else {
+		if (!downloadTable.nr)
+			error = tr("No new dives downloaded from dive computer");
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
ok

PR #1755, opened 2018-10-06
[Please Test] Dive information: fix surface interval calculation

bstoeger:
The old surface interval calculation had fundamental issues:

1) process_all_dives(), which calculates the statistics over *all*
   dives was used to get the pointer to the previous dive.
2) If two dives in the table had the same time, one of those would
   have been considered the "previous" dive.
3) If the dive, for which the surface interval is calculated is
   not yet in the table, no previous dive would be determined.

Fix all this by creating a get_surface_interval() function and
removing the "get previous dive" functionality of process_all_dives().
Remove the process_all_dives() call from TabDiveInformation::updateData().

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The surface interval calculation was broken in a spectacular way [*all* statistics were calculated just to get the pointer to the previous dive - and even that didn't work properly]. Try to fix this. See commit message. This could need some testing.

N.B.: The `process_all_dives()` call in `MainTab::updateDiveInfo` also seems dubious, as in the general case yearly stats will *not* be displayed. Why should this be calculated on switching dives?

### Changes made:

1) Create `get_surface_interval()` function.
2) Remove "get previous dive" functionality from `process_all_dives()`
3) Remove `process_all_dives()` call from `TabDiveInformation::updateData()`
### Additional information:

None.
### Release note:

Will do.
### Documentation change:

No.
### Mentions:

@janmulder

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1420,3 +1420,32 @@ void sort_table(struct dive_table *table)
 {
 	qsort(table->dives, table->nr, sizeof(struct dive *), sortfn);
 }
+
+/*
+ * Calculate surface interval for dive starting at "when". Currently, we
+ * might display dives which are not yet in the divelist, therefore the
+ * input parameter is a timestamp.
+ * If the given dive starts during a different dive, the surface interval
+ * is 0. If we can't determine a surface interval (first dive), <0 is
+ * returned. This does *not* consider pathological cases such as dives
+ * that happened inside other dives. The interval will always be calculated
+ * with respect to the dive that started previously.
+ */
+timestamp_t get_surface_interval(timestamp_t when)
+{
+	int i;
+	timestamp_t prev_end;
+
+	/* find previous dive. might want to use a binary search. */
+	for (i = dive_table.nr - 1; i >= 0; --i) {
+		if (dive_table.dives[i]->when < when)
+			break;
+	}
+	if (i < 0)
+		return -1;
+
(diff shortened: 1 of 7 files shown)


janmulder:
Tried this on my overlapping test set, and I'm impressed. Even in that very unrealistic set of dives, all intervals look sane. Sane in the sense that I understand the value presented.

```
/* find previous dive. might want to use a binary search. */
for (i = 0; i < dive_table.nr; ++i) {
```
Practically, most users will work on recent dives, even when they have a huge logbook. So, why don't you just walk the `dive_table` backwards? That will find the previous dive much faster in most use cases.

And finally. Yes, I found something else (not related to this PR), as I was triggered by the change in the HTML export here. When exporting a subset of dives (which works correctly), the statistic HTML shows the statistics for the complete logbook. I believe this is a bug.

bstoeger:
> Practically, most users will work on recent dives, even when they have a huge logbook. So, why don't you just walk the dive_table backwards? 

Good point - done. Ultimately we should do a binary (or interpolation) search. We also need this in other places (e.g. import).

> When exporting a subset of dives (which works correctly), the statistic HTML shows the statistics for the complete logbook. 

Yes, the statistics code needs some overhaul. Looking into it.

torvalds:
Looks good to me.

PR #1762, opened 2018-10-07
Recent fixes

### LINKED ISSUE #1761: serial_ftdi_read() read timeout potentially too long
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

The function serial_ftdi_read() in core/serial_ftdi.c is not handling a read timesout properly.
This can cause the timeout  period while waiting for read data to be 10x or even 100x longer than it should be when there are read issues on the data cable particularly when using Android and USB OTG cables.
i.e. a  5 second read timeout for not receiving data can be as long as 7 minutes.

When things are working normally without communication errors, this bug does not create any problems for downloading from the dive computer; However, when there are communication issues, an extended timeout can cause libdivecomputer retries to fail as the data cable and/or divecomputer can timeout long before libdivecomputer can re-send the command to the dive computer to retry the command.

The issue is that the code has a loop that checks for characters every 1ms.
It loops 1 time for each 1ms in the desired timeout. The issue is that the the code inside the loop is not accounting for the amount of time consumed by the code inside the loop so the actual timeout period can be much longer than the desired timeout period.
The main culprit for the added time is the ftdi_read_data() call. This time used by the function can vary substantially and can be as long as 140ms when using libusb on Android with a USB OTG adapter and FTDI data cable.  - (as is used when using a FTDI data cable on Pelagic dive computers)
The solution is to still sleep for 1ms between polls for characters but not blindly loop for the number of milliseconds in the timeout period but rather break out of the loop when the elapsed time in the loop exceeds the desired timeout period.

### Operating system:

Pretty much all but I noticed issues when using Android 4.4.4 and Android 8.0.0

Galaxy S4 (Android 4.4.4) Galaxy S7 (Android 8.0.0)
Atmos Aeris AI using Oceanic 04.9600 USB data cable

### Subsurface version:

Pretty much any subsurface version, but the extra long timeouts are most noticeable on the older versions of subsurface mobile on Android from late summer of 2017.

### Steps to reproduce:

It can be difficult to fully reproduce the super extra long timeouts as the excess time used by ftdi_read_data() can vary substantially depending on the version of Android and the version of Subsurface mobile.
To reproduce / test simply connect the FTDI data cable to the phone but not to a Dive computer and try to do a download from an Aeris Atmos AI dive computer.
The Subsurface libdivecomputer code has read timeouts and does retries and the timeouts can be much longer than expected - but it depends on the version of Subsurface mobile.
The earlier Subsurface versions from the summer of 2017 were much better at demonstrating the issue.

### Current behavior:

Latest versions of Subsurface Mobile do not work on Galaxy S4 with Android 4.4.4
It can't get permission to use the USB device as the USB device dialog does not come up anymore when the USB cable is plugged in on the past few Subsurface mobile builds from the Play Store.

Subsurface Mobile Version 2.1.3(4.8.2.5) on a Galaxy S7 seems to timeout without an excessively long timeout period when the dive computer is not attached. However, even though the existing code is not currently creating/triggering an excessive timeout, the issue/bug still exists.

### Expected behavior:

serial_ftdi_read() returns after the specified timeout period with a TIMEOUT status if all the data wasn't read.

### Additional information:

I had a solution that worked on Android that I was going to commit and generate a pull request to fix it, but it depends on gettimeofday() to be able to track elapsed time. For some reason WIN32 does not include this call. I found a substitute for this for WIN32 but I have no idea about the other OSes.
I'm not sure what the final soultion needs to be and whether or not Subsurface is interested in pulling in this WIN32 gettimeofday() function.

Below is the serial_ftdi_read() code I was using:
```
static dc_status_t serial_ftdi_read (void *io, void *data, size_t size, size_t *actual)
{
	ftdi_serial_t *device = io;

	if (device == NULL)
		return DC_STATUS_INVALIDARGS;

	// The total timeout.
	long timeout = device->timeout;

	// Simulate blocking read as 10s timeout
	if (timeout == -1)
		timeout = 10000;

	int backoff = 1;
	unsigned int nbytes = 0;

        struct timeval tvstart, tvnow;
        gettimeofday(&tvstart, 0);

	while (nbytes < size) {
		int n = ftdi_read_data (device->ftdi_ctx, (unsigned char *) data + nbytes, size - nbytes);
		if (n < 0) {
			if (n == LIBUSB_ERROR_INTERRUPTED)
				continue; //Retry.
			ERROR (device->context, "%s", ftdi_get_error_string(device->ftdi_ctx));
			return DC_STATUS_IO; //Error during read call.
		} else if (n == 0) {
                        gettimeofday(&tvnow, 0);
                        // check to see if too much time has elapsed
                        if( ((tvnow.tv_sec - tvstart.tv_sec) * 1000) + ((tvnow.tv_usec - tvstart.tv_usec)/1000) >= timeout) {
				ERROR(device->context, "%s", "FTDI read timed out.");
				return DC_STATUS_TIMEOUT;
			}
			serial_ftdi_sleep (device, backoff);
		}

		nbytes += n;
	}

	INFO (device->context, "Read %d bytes", nbytes);

	if (actual)
		*actual = nbytes;

	return DC_STATUS_SUCCESS;
}
```
Here is a WIN32 gettimeofday() function if that is desired route:
```
// for win32 a gettimeofday() is needed.
// one can be found here: https://stackoverflow.com/questions/1676036/what-should-i-use-to-replace-gettimeofday-on-windows
#ifdef _WIN32
/*
 * gettimeofday.c
 *    Win32 gettimeofday() replacement
 *
 * src/port/gettimeofday.c
 *
 * Copyright (c) 2003 SRA, Inc.
 * Copyright (c) 2003 SKC, Inc.
 *
 * Permission to use, copy, modify, and distribute this software and
 * its documentation for any purpose, without fee, and without a
 * written agreement is hereby granted, provided that the above
 * copyright notice and this paragraph and the following two
 * paragraphs appear in all copies.
 *
 * IN NO EVENT SHALL THE AUTHOR BE LIABLE TO ANY PARTY FOR DIRECT,
 * INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES, INCLUDING
 * LOST PROFITS, ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS
 * DOCUMENTATION, EVEN IF THE UNIVERSITY OF CALIFORNIA HAS BEEN ADVISED
 * OF THE POSSIBILITY OF SUCH DAMAGE.
 *
 * THE AUTHOR SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING, BUT NOT
 * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
 * A PARTICULAR PURPOSE.  THE SOFTWARE PROVIDED HEREUNDER IS ON AN "AS
 * IS" BASIS, AND THE AUTHOR HAS NO OBLIGATIONS TO PROVIDE MAINTENANCE,
 * SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
 */

#include "c.h"

#include <sys/time.h>


/* FILETIME of Jan 1 1970 00:00:00. */
static const unsigned __int64 epoch = ((unsigned __int64) 116444736000000000ULL);

/*
 * timezone information is stored outside the kernel so tzp isn't used anymore.
 *
 * Note: this function is not for Win32 high precision timing purpose. See
 * elapsed_time().
 */
static int
gettimeofday(struct timeval * tp, struct timezone * tzp)
{
    FILETIME    file_time;
    SYSTEMTIME  system_time;
    ULARGE_INTEGER ularge;

    GetSystemTime(&system_time);
    SystemTimeToFileTime(&system_time, &file_time);
    ularge.LowPart = file_time.dwLowDateTime;
    ularge.HighPart = file_time.dwHighDateTime;

    tp->tv_sec = (long) ((ularge.QuadPart - epoch) / 10000000L);
    tp->tv_usec = (long) (system_time.wMilliseconds * 1000);

    return 0;
}

#endif
```

torvalds:

- [x] Bug fix
- [x] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Various random misc fixes 

### Changes made:

1) implement the "get_name()" function the BLE iostream, needed for the Aqualung i770R and i300C

2) make the ftdi user-space driver hopefully get the right timeouts across different operating systems

3) fix a build warning

### Related issues:

Fixes #1761 

### Mentions:

@bperrybap

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1348,7 +1348,6 @@ static bool try_to_merge_into(struct dive *dive_to_add, int idx, bool prefer_imp
 void process_imported_dives(struct dive_table *import_table, bool prefer_imported, bool downloaded)
 {
 	int i, j;
-	struct dive *old_dive, *merged;
 	int preexisting;
 	bool sequence_changed = false;
 
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -70,6 +70,7 @@ void DownloadThread::run()
 	auto internalData = m_data->internalData();
 	internalData->descriptor = descriptorLookup[m_data->vendor() + m_data->product()];
 	internalData->download_table = &downloadTable;
+	internalData->btname = strdup(m_data->devBluetoothName().toUtf8());
 #if defined(Q_OS_ANDROID)
 	// on Android we either use BT, a USB device, or we download via FTDI cable
 	if (!internalData->bluetooth_mode && (same_string(internalData->devname, "FTDI") || same_string(internalData->devname, "")))
--- a/core/libdivecomputer.h
+++ b/core/libdivecomputer.h
@@ -30,7 +30,7 @@ typedef struct dc_user_device_t
 {
 	dc_descriptor_t *descriptor;
 	const char *vendor, *product, *devname;
-	const char *model;
+	const char *model, *btname;
 	unsigned char *fingerprint;
 	unsigned int fsize, fdiveid;
 	uint32_t libdc_firmware, libdc_serial;
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -542,4 +542,10 @@ dc_status_t qt_ble_write(void *io, const void* data, size_t size, size_t *actual
 	return ble->write(data, size, actual);
 }
 
+const char *qt_ble_get_name(void *io)
+{
+	BLEObject *ble = (BLEObject *) io;
+	return ble->get_name();
+}
+
 } /* extern "C" */
--- a/core/qt-ble.h
+++ b/core/qt-ble.h
@@ -23,6 +23,7 @@ public:
 	inline void set_timeout(int value) { timeout = value; }
 	dc_status_t write(const void* data, size_t size, size_t *actual);
 	dc_status_t read(void* data, size_t size, size_t *actual);
+	inline const char *get_name() { return device->btname; }
 
 	inline QLowEnergyService *preferredService() { return preferred; }
 	inline int descriptorWritten() { return desc_written; }
@@ -63,6 +64,7 @@ dc_status_t qt_ble_set_timeout(void *io, int timeout);
 dc_status_t qt_ble_read(void *io, void* data, size_t size, size_t *actual);
 dc_status_t qt_ble_write(void *io, const void* data, size_t size, size_t *actual);
 dc_status_t qt_ble_close(void *io);
+const char *qt_ble_get_name(void *io);
 }
 
 #endif
--- a/core/qtserialbluetooth.cpp
+++ b/core/qtserialbluetooth.cpp
@@ -424,6 +424,7 @@ ble_packet_open(dc_iostream_t **iostream, dc_context_t *context, const char* dev
 		NULL, /* purge */
 		qt_custom_sleep, /* sleep */
 		qt_ble_close, /* close */
+		qt_ble_get_name, /* get_name */
 	};
 
 	rc = qt_ble_open(&io, context, devaddr, (dc_user_device_t *) userdata);
--- a/core/serial_ftdi.c
+++ b/core/serial_ftdi.c
@@ -103,6 +103,20 @@ static dc_status_t serial_ftdi_get_transmitted (ftdi_serial_t *device)
 	return DC_STATUS_UNSUPPORTED;
 }
 
+/*
+ * Get an msec value on some random base
+ */
+static unsigned int serial_ftdi_get_msec(void)
+{
+#ifdef _WIN32
+	return GetTickCount();
+#else
+	struct timeval tv;
+	gettimeofday(&tv, NULL);
+	return tv.tv_sec * 1000 + tv.tv_usec / 1000;
+#endif
+}
+
 static dc_status_t serial_ftdi_sleep (void *io, unsigned int timeout)
 {
 	ftdi_serial_t *device = io;
@@ -390,8 +404,7 @@ static dc_status_t serial_ftdi_read (void *io, void *data, size_t size, size_t *
 	if (timeout == -1)
 		timeout = 10000;
 
-	int backoff = 1;
-	int slept = 0;
+	unsigned int start_time = serial_ftdi_get_msec();
 	unsigned int nbytes = 0;
 	while (nbytes < size) {
 		int n = ftdi_read_data (device->ftdi_ctx, (unsigned char *) data + nbytes, size - nbytes);
@@ -401,12 +414,11 @@ static dc_status_t serial_ftdi_read (void *io, void *data, size_t size, size_t *
 			ERROR (device->context, "%s", ftdi_get_error_string(device->ftdi_ctx));
 			return DC_STATUS_IO; //Error during read call.
 		} else if (n == 0) {
-			if (slept >= timeout) {
+			if (serial_ftdi_get_msec() - start_time > timeout) {
 				ERROR(device->context, "%s", "FTDI read timed out.");
 				return DC_STATUS_TIMEOUT;
 			}
-			serial_ftdi_sleep (device, backoff);
-			slept += backoff;
+			serial_ftdi_sleep (device, 1);
 		}
 
 		nbytes += n;


neolit123:
LGTM.

as a side note, Dirk gave me his Perdix AI and i'm getting timeout issues with the Qt stack on Windows, so i need to figure out why.
it fails somewhere after downloading two dives.

the OSTC+ works fine.

dirkhh:
@neolit123  - I'm excited that you get the Perdix AI to talk at all. And yes - one thing that Linus' work seems to have taught us over the past two months is that BLE is rather sensitive to timing and timeouts...

torvalds:
@neolit123 so the BLE code connects fine and actually starts downloading and has successful IO?

Usually, once things get that far, BLE tends to be pretty reliable for me. But I've seen two different failure modes:

 (a) just lost packets, which need retries (and libdivecomputer doesn't necessarily always do that well, it depends on the particular back-end)

 (b) an actual lost connection where the divecomputer itself gives up

and particularly for the Shearwater, that second case is what I've occasionally seen. 

If you can get a bluetooth packet trace, that would be interesting. I suspect, for example, that we do the wrong thing in the BLE reading side: we just ignore the first two bytes of a BLE packet, but I *think* it contains a "packet sequence number", and if we have packet loss, I'm not sure we do the right thing.

But honestly, the shearwater really tends to be pretty good once it has connected and the data transfer has started, so I wonder if it's some Qt Windows BLE oddity.

torvalds:
Ugh. Travis just failed on this. But it *looks* like that may just be a build stall. I'm not seeing any actual build errors, but it might also be that I'm just Travis-incompetent.

dirkhh (approved):
Thanks for taking care of these!

dirkhh:
Travis failure was a network operation timing out. Restarted "just because" - I know that this will pass since all the other ones passed.

neolit123:
@dirkhh 
our users would have to update to the latest version of Windows 10.
my laptop decided to update today (outside of my control), it picked (https://en.wikipedia.org/wiki/Windows_10_version_history#Version_1803_(April_2018_Update)) and after that:
- the discovery of BLE devices is much more stable at least for the OSTC+, while the Perdix often needs OS restart to be paired again.
- i wrote a C# console app that can be called to perform programmatic discovery and paring on a BT(LE) devices. before this Windows update it was not possible to pair outside of a UWP app, but after the update a regular C# app can do it too.

i was thinking about invoking this C# app from Subsurface somehow.

torvalds:
@neolit123 that packet trace looks perfectly fine.

Yes, we ignore the first two bytes of each BLE packet, but they are "number of packets in this reply" and "packet sequence number within this reply", so it's not actually all that critical unless you have packets being re-ordered or something. 

And nothing is re-ordered. It's literally that we stop receiving replies. The last command we send gets the first packet back, and that packet says "there will be three packets in this reply, this is the first of them". And then nothing after that.

And three seconds after getting that partial reply, we give up and send the "bye bye" command.

Now, the Shearwater back-end is one of the libdivecomputer backends that doesn't even try to do any retries, and I guess I could try to add something like that now that we have reasonable timeout behavior. But that code has _never_ tried to do retries, and it's actually been pretty reliable if it just gets started, so I don't think the lack of retries is the problem here.

But yeah, BLE on the desktop has always been pretty badly supported. Windows may just be the worst example of a fairly general issue. 

The actual bluetooth chip you use might matter too. My laptop tends to connect better than my desktop, for example (but I do all my BLE development on the desktop anyway, I only use the laptop when traveling)

dirkhh:
> our users would have to update to the latest version of Windows 10.
> 

Yeah, I was getting this impression as well.

> my laptop decided to update today (outside of my control), it picked (https://en.wikipedia.org/wiki/Windows_10_version_history#Version_1803_(April_2018_Update) <https://en.wikipedia.org/wiki/Windows_10_version_history#Version_1803_(April_2018_Update)>) and after that:
> 
> the discovery of BLE devices is much more stable at least for the OSTC+, while the Perdix often needs OS restart to be paired again.
> i wrote a C# console app that can be called to perform programmatic discovery and paring on a BT(LE) devices. before this Windows update it was not possible to pair outside of a UWP app, but after the update a regular C# app can do it too.

That's the version I'm on as well.

> i was thinking about invoking this C# app from Subsurface somehow.
> 

As you know, I'm ok with an idea like that if it makes the user experience better. I continue to be hopeful that we can get SOMETHING to work.

PR #1763, opened 2018-10-08
Mobile: show temperature graph when no gas graph

janmulder:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In commit bd0c99dfb71 (display pO2 and Setpoint for CCR dives) the choice was made: when CCR than never a temperature graph. While this seems reasonable, there is small group of mCCR divers that do not log their po2 digitally, so the only po2 they can display is the setpoint that is reported by the used DC. As this is a bit of a dull flat line, most of these divers do not display this. And this leaves room on the small mobile display for the temperature data.

So effectively: show temperature or po2 graphs.

Suggested-by: Peter Zaal <pzaal@xs4all.nl>

### Related issues:
commit bd0c99dfb71

### Release note:
Well ... this is a change for a tiny group mCCR divers ... I dot not feel this is needed.


dirkhh (approved):
That seems like a reasonable thing to do.

PR #1764, opened 2018-10-08
[BUGFIX] Map: zoom on dive sites when flipping through dive site list

bstoeger:
The dive site list was connected to centerOnDiveSite(). Apparently,
the currently selected dive site should have been shown in the map.
Yet, this never worked, because the actual dive site of the selected
dive had precedence in centerOnDiveSite().

It seems that centerOnDiveSite() had actually to purposes:
1) center on the passed in dive site
2) center on the dive sites of the selected dives

Therefore, split this function in two separate functions for
each of these use-cases. This allows us to remove some pre-processor
magic (mobile vs. desktop) and to remove a parameter from the
MainTab::diveSiteChanged() signal.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a code cleanup / bugfix: skipping through the divelist did not switch zoom on the divesites as the code apparently intended. Fix this by splitting `centerOnDiveSite` in two functions. Note: I didn't test mobile. Do some minor refactoring while at it. Notably, remove one vector and a signal-argument.

### Changes made:

1) Split `centerOnDiveSite` in two versions
2) Refactor the new `centerOnSelectedDiveSite` slightly
3) On change of current dive site call the new `centerOnSelectedDiveSite`

### Additional information:

None
### Release note:

Will do.
### Documentation change:

No.
### Mentions:

@neolit123: seems like you touched the code last. Please have a look, especially concerning the mobile-version.

### DIFF
--- a/desktop-widgets/locationinformation.cpp
+++ b/desktop-widgets/locationinformation.cpp
@@ -360,7 +360,6 @@ bool DiveLocationFilterProxyModel::lessThan(const QModelIndex &source_left, cons
 	return source_left.data().toString() < source_right.data().toString();
 }
 
-
 DiveLocationModel::DiveLocationModel(QObject*)
 {
 	resetModel();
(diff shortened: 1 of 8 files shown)


neolit123:
the change looks OK to me, but i didn't compile / test it, as i'm busy with other tasks ATM.

> Note: I didn't test mobile.

please do.

dirkhh (approved):
The change looks reasonable.
I don't see how this could break mobile (famous last words)

bstoeger:
> please do.

It seems to still work on mobile-on-desktop. The problem is that I'm not even fully aware of all the mobile features, though.

dirkhh:
I got distracted noticing that Mobile on Desktop doesn't compile for me with cmake 3.12 on a Mac.
The only thing you really would need to test is if the map shows up correctly when selecting it from dives

bstoeger:
Yes, that still works.

dirkhh:
Thanks for the testing.

PR #1765, opened 2018-10-08
Update BT/BLE flow on Macs

dirkhh:

- [ ] Bug fix
- [X] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

On the Mac we cannot connect to a BT/BLE device unless it was found via scan during THIS session. That does seem to be an odd limitation and it seems other software doesn't suffer from this limitation, but it is what we appear to see. So this PR changes our behavior on macOS to always start scanning once the download dialog opens and to wait before trying to connect with a remembered dive computer until it has been found.

### Changes made:

1) add a couple of helper functions to make life easier
2) change the code to always scan on all OSs except for Android
3) wait until the dive computer was found in a scan before trying to connect to it

### Release note:

Not sure how to describe this. In a way this is user visible, of course, but it actually just makes macOS behave more like the other platforms.
Something like:

macOS: don't require explicit BT/BLE scan for known devices

### Documentation change:

I need to check if we explicitly mention that on a Mac you **always** have to scan before downloading from a BT/BLE device

### Mentions:

@atdotde @jbygdell (? - I keep forgetting who has a Mac for testing)


jbygdell:
I can't test on mac, I scrapped mine once it could not be updated anymore. That must be 2 or 3 OS revisions ago.

dirkhh:
@mfbernardes this would be one where testing on Mac would be really important (assuming you have a BT or BLE dive computer... which of course makes that whole "testing matrix" even more complicated)

atdotde:
Will give it a try. Please give me a bit of time.

atdotde:
Download to Mac works for me both for BL and BLE from OSTC Sport. I still had to start the scan manually. Using the "used recently" button worked without rescanning (of BT, BTE was not offered). Anything else to try?

dirkhh:
> Download to Mac works for me both for BL and BLE from OSTC Sport. I still had to start the scan manually.

That sounds like it didn't work as intended. If you quit Subsurface and start again, and it keeps the BT address from last time, simply hitting Download should add a fairly long delay (since it scans in the background), but it should WORK without manual scanning

mfbernardes:
@dirkhh Tested with a Perdix AI on Mojave.

In general works as intended, BUT when you press "Download" it shows "Connecting to dive computer" for quite a while (giving the impression something is wrong) then after a while it connects and downloads correctly.

If I open the download dialog and wait a bit it just work directly. So, it's just the time to get the scan finished (or at least discover the expected dive computer).

The annoying part is the unusual wait with the "Connecting to dive computer". Also, in case you get annoyed by the wait and hit "Cancel download", the entire dialog is blocked until the scan finishes, with the dialog Cancel button disabled and the entire application unresponsive.

Note: this branch didn't build with cmake 3.12 so I first rebased with master.

dirkhh:
Oh, and on the user experience issue... hmm... I wonder what we could do to make this seem less "hung". Maybe change the text to not just "connecting" but "trying to find your dive computer (this can take 30 seconds)"?

mfbernardes:
> Yes, I actually added the little activity visualization to deal with that long delay... and you are right - that's the time to finish the scan. I wish I knew how some other apps manage to avoid that scan or to at least speed it up...
> 
> And there is a cmake 3.12 problem in Subsurface-mobile, but I hadn't seen one in the desktop app... that's pretty odd because the very system I worked on this branch on is also the system (with cmake 3.12) where I found and fixed the Subsurface-mobile problem.
> Can you tell me what the error was?

```
[ 63%] Linking CXX executable Subsurface.app/Contents/MacOS/Subsurface
ld: library not found for -lcrypto
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [Subsurface.app/Contents/MacOS/Subsurface] Error 1
make[1]: *** [CMakeFiles/Subsurface.dir/all] Error 2
make: *** [all] Error 2
```

I didn't dig into this, but since master built fine I just rebased and it built... but checking again even on master I get this error sometimes.

atdotde:
Once more, I did not expressly clearly what I meant: yes, after restarting subsurface, the BT address is still there and I can directly connect without starting a manual scan.

But in the discovery sub-dialogue, even when the list of devices is empty, it does not start a scan by itself but I still have to press the scan button.

neolit123:
> On the Mac we cannot connect to a BT/BLE device unless it was found via scan during THIS session. That does seem to be an odd limitation and it seems other software doesn't suffer from this limitation, but it is what we appear to see. So this PR changes our behavior on macOS to always start scanning once the download dialog opens and to wait before trying to connect with a remembered dive computer until it has been found.

i wonder if this is something that can be patched in Qt and we should ask the maintainers.

this is a bit odd on the OS side.
i'm going to assume that it just maps a random temporary address each time a device is scanned.

AKA - on demand virtual address mapping as a security precaution?
one needs to request a virtual address to be able to connect to a paired device.

haven't looked much at the code in the Qt backend for OSX.

PR #1771, opened 2018-10-10
Mobile: prevent crash and more cosmetics

janmulder:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

2 simple commits. It was possible to crash the mobile app by selecting download from dc, and have the product/vendor not filled in. Found this by accident. Not having a DC paired and BT switched off on the device, shows the DL screen without product/vendor. Hit download results in SIGSEGV. The reason is simple, and made more safe here too. 

And while we are at it. Disable the rescan button when BT is off. It makes no sense.


dirkhh (approved):
Thanks for finding this

PR #1775, opened 2018-10-12
Edit: prevent not needed change marking

janmulder:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

While trying to understand more of the big change from PR 1528, I found that the divelist was marked changed while it was not changed at all. Reason is simple. The MODIFY_DIVES code assumes
its called only for truly changed data. But in case of saving tagged strings, it was not.

This fixes this. And I do not believe this has any visual effects.

### Additional information:
The whitespace change is separated. When preferred this may be squashed.

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -1280,47 +1280,47 @@ void MainTab::saveTaggedStrings(const QVector<dive *> &selectedDives)
 	struct dive *cd = current_dive;
 
 	if (diffTaggedStrings(cd->buddy, displayed_dive.buddy, addedList, removedList))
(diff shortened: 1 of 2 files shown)


bstoeger (approved):
Makes sense.

Recently I was also observing a strange behavior where, when tabbing through the fields of the dive, it would add an empty entry to the dive-master and buddy fields. But not always. Could this be related?

bstoeger:
Indeed unrelated. I think you are on to something: when I hit tab it probably interprets this as "enter dive-buddy starting with `\t`" The strange thing: once I click on "discard changes", this doesn't happen anymore.

In any case, this PR seems obviously correct. I think we can commit.

PR #1776, opened 2018-10-12
QML, cleanup: remove stackView

janmulder:
There is no reason (any more?) to have a property defined that basically renames the global pageStack into a local pageView. Just cleanup.


dirkhh (approved):
I kept stumbling over that, but always suspected that there was a reason that I was missing :-/

PR #1777, opened 2018-10-12
Cleanup: Turn widget accessor-functions into simple pointers

bstoeger:
The keeps track of different sub widgets needed by other parts
of the code, notably:
	MainTab
	PlannerDetails
	PlannerSettingsWidget
	ProfileWidget2
	DivePlannerWidget
	DiveListView
Access to these widgets was provided with accessor functions.
Now these functions were very weird: instead of simply returning
pointers that were stored in the class, they accessed a data
structure which describes the different application states.
But this data structure was "duck-typed", so there was an
implicit agreement at which position the pointers to the
widgets were put inside. The widgets were then down-cast by
the accessor functions. This might make sense if the individual
widgets could for some reason be replaced by other widgets
[dynamic plugins?], but even then it would be strange, as one
would expect to get a pointer to some base class.

Therefore, directly store the properly typed pointers to the
widgets and simply remove the accessor functions. Why bother?

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a small cleanup which removes "duck typing" in `MainWindow`. It touches a hairy part of the code though. The reason for this cleanup is that I want to add access to the dive-location edit widget outside of `MainWindow`'s constructor. But I couldn't be motivated to add one of those strange accessor-functions, which might have been useful at one point. But now, I think they just needlessly complicate things.

### Changes made:

1) Replace widget-accessors by simple pointers
### Related issues:

None.
### Additional information:

No.
### Release note:

None.
### Documentation change:

None.

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -137,14 +137,14 @@ MainWindow::MainWindow() : QMainWindow(),
 	// widgets will change on the mainwindow.
 
 	// for the "default" mode
-	MainTab *mainTab = new MainTab();
-	DiveListView *diveListView = new DiveListView();
-	ProfileWidget2 *profileWidget = new ProfileWidget2();
+	information = new MainTab(this);
+	dive_list = new DiveListView(this);
+	graphics = new ProfileWidget2(this);
 	MapWidget *mapWidget = MapWidget::instance();
 
-	PlannerSettingsWidget *plannerSettings = new PlannerSettingsWidget();
-	DivePlannerWidget *plannerWidget = new DivePlannerWidget();
-	PlannerDetails *plannerDetails = new PlannerDetails();
+	divePlannerSettingsWidget = new PlannerSettingsWidget(this);
+	divePlannerWidget = new DivePlannerWidget(this);
+	plannerDetails = new PlannerDetails(this);
 
 	// what is a sane order for those icons? we should have the ones the user is
 	// most likely to want towards the top so they are always visible
@@ -170,7 +170,7 @@ MainWindow::MainWindow() : QMainWindow(),
 	profLayout->setMargin(0);
 	profLayout->setContentsMargins(0,0,0,0);
 	profLayout->addWidget(toolBar);
-	profLayout->addWidget(profileWidget);
+	profLayout->addWidget(graphics);
 	profileContainer->setLayout(profLayout);
 
 	LocationInformationWidget * diveSiteEdit = new LocationInformationWidget();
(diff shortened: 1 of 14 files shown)


desktop-widgets/mainwindow.cpp
```diff
@@ -137,14 +137,14 @@ MainWindow::MainWindow() : QMainWindow(),
 	// widgets will change on the mainwindow.
 
 	// for the "default" mode
-	MainTab *mainTab = new MainTab();
-	DiveListView *diveListView = new DiveListView();
-	ProfileWidget2 *profileWidget = new ProfileWidget2();
+	information = new MainTab(this);
```
atdotde: what's wrong with calling this mainTab? I feel a bit uneasy with these automatically typed variables but maybe only because I am not used to these and do not immediately spot those as declarations.

atdotde (commented):
The general theme of removing the accessors sounds very reasonable to me.

desktop-widgets/mainwindow.cpp
bstoeger: `mainTab` would be the right name indeed. I just didn't want to change too much in this commit and for some (historical?) reason this was named `information` before.

> I feel a bit uneasy with these automatically typed variables but maybe only because I am not used to these and do not immediately spot those as declarations.

Not sure what you mean here - the declaration is in the `mainwindow.h` file. I was thinking about moving these initializations into the initializer list:
```
MainWindow::MainWindow() :
...
	mainTab(new MainTab(this)),
	diveListView(new diveListView(this)),
...
```
Perhaps this would make it more explicit that these are member-variables?

bstoeger: I pushed the variable renaming as a second commit. But probably this should be squashed into the first.

Then I noted that actually we can't initialize the sub-widgets in the initializer-list, because the constructors of the widgets use `MainWindow::instance()`, which is not yet set! To me the whole thing looks very brittle or even undefined-behavior-ish: The widgets happily use `MainWindow` during construction of the latter.

dirkhh:
I'm not 100% sure, but I think this goes back to the ability to have different widgets in the different quadrants of the overall app window. @tcanabrava should be able to refresh our memories.
The changes are squarely in the "heck if I know, sure, looks good, but I worry that there was a reason" category...

bstoeger:
There are two aspects to this:

1) Accessor functions. There is a valid argument for those vs. pointers: other code should be able to access the sub-widgets but should not be able to modify the pointers. Accessors are one way of realizing that [other possibilities are e.g. const-pointers or references]. For me, personally, this is a "Q: Doctor it hurts if ... A: don't do it then."-situation. I (again purely subjectively) don't see the value in code only for what feels like finger-wagging. But in the end it is a style thing and therefore I would not have a problem with the functions, nor would I touch them if it weren't for the second point:
2) Pointer storage. While I see how the fetch from the state-arrays might have looked like a cool idea, I cannot fathom a situation were this could make sense. Think about it: A change in the location of the widgets makes it necessary to adapt the accessor functions!? Not even talking about weak typing in a strongly typed language.

In summary: accessor functions ok; location-depended and weakly typed pointer store: very bad idea.

PR #1778, opened 2018-10-12
Desktop: enable tab-changes-focus for TagWidget

bstoeger:
When tabbing through the dive-info screen, a dive-master and a
dive-buddy would be added. The reason is that pressing tab would
be interpreted as a text-input. Disable this behavior by calling
setTabChangesFocus(true) in the TagWidget constructor.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

There was a weird behavior on tabbing through the dive-info fields: an empty dive-buddy and a dive-master would be added, because TAB would be registered as text-input. This can be prevented by setting the tabChangesFocus flag.
### Release note:

Done.
### Documentation change:

No.
### Mentions:

@janmulder: see our previous discussion on the topic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Desktop: don't add dive-buddy or dive-master when tabbing through fields
 - Filter: don't recalculate all filters on dive list-modifying operations
 - Desktop: don't recalculate full dive list on dive list-modifying operations
 - Undo: implement undo/redo for all dive list-modifying operations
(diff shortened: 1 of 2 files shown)


janmulder:
Nice. Tried this, and works correctly. 

Interestingly, before this fix, I could not force the problem in a deterministic way. Only fast tabbing sometimes introduced the /t character in the tab field. Tried the same after this fix, and I cannot reproduce the issue any more, whatever I try.

PR #1779, opened 2018-10-13
Dive site: remove displayed_dive_site object

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is the final part of a sub-series that removes the `displayed_dive_site` object. This makes it distinctly easier to replace dive-site uuids by pointers. Indeed, the `displayed_dive_site` object is replaced by a pointer.

### Changes made:

1) Remove external updates to `displayed_dive_site`
2) Make `displayed_dive_site` local to `LocationInformationWidget`
3) Don't write taxonomy and GPS data into `displayed_dive_site`
4) Replace `displayed_dive_site` by pointer to edited site.
### Related issues:

This introduces a pointer to the LocationInformationWidget in Mainwindow according to the simplified scheme in #1777 instead of doing the weird fetch from application state descriptor array.
### Release note:

Not needed.
### Documentation change:

Not needed.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -18,7 +18,6 @@
  * it's used in the UI, but it seems to make the most sense to have it
  * here */
 struct dive displayed_dive;
-struct dive_site displayed_dive_site;
 
 struct tag_entry *g_tag_list = NULL;
 
--- a/core/dive.h
+++ b/core/dive.h
@@ -439,7 +439,6 @@ struct dive_table {
 
 extern struct dive_table dive_table, downloadTable;
 extern struct dive displayed_dive;
-extern struct dive_site displayed_dive_site;
 extern unsigned int dc_number;
 extern struct dive *current_dive;
 #define current_dc (get_dive_dc(current_dive, dc_number))
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1677,7 +1677,6 @@ void clear_dive_file_data()
 		delete_dive_site(get_dive_site(0)->uuid);
 
 	clear_dive(&displayed_dive);
-	clear_dive_site(&displayed_dive_site);
 
 	reset_min_datafile_version();
 	saved_git_id = "";
--- a/core/divesite-helper.cpp
+++ b/core/divesite-helper.cpp
@@ -3,11 +3,11 @@
 #include "pref.h"
 #include "gettextfromc.h"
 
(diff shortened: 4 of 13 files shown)


dirkhh:
On a plane (yay, dive trip!)
The original code was crazy fragile because it tried to deal with aborted edits. So this might deserve a careful review (which I can't really do until tomorrow, I guess). Happy to have others take a closer look, so I'm not asking that you wait for me, just pointing out that this went through a lot of debugging when it was introduced and it's not code that I'm proud of..

bstoeger:
I hope that the new code is distinctly less fragile because it doesn't rely on global objects. I tested the interactions with the map and geo-lookup, but I'm certain that others will come up with situations that I didn't expect.

Take your time and enjoy your trip; it's not a pressing issue at all.

janmulder:
First remark of something that is **not** introduced here. When selecting the globe to edit the site, the site panel opens with the blue bar already open. But nothing is edited yet (so there i noting to apply or discard, so to be consistent it needs to be hidden.

**Edit**: never mind this remark.We need a way to get back to the main dive edit page. An apply/cancel is logical. I was just distracted by the big blue bar, that I associate with edited.

dirkhh:
Layover in Houston, but didn't unpack my laptop.
Typically a lot of Subsurface work happens on dive trips 😁

dirkhh (approved):
I vaguely remember that there was an issue when you opened a dive, edited the dive site, and then continued to edit the dive, and somehow that could mess things up if you discarded your changes.
But I can't figure out how exactly that caused a problem and I can't seem to recreate the error case.
So I'll merge this with some slight uneasiness...

bstoeger:
> I vaguely remember that there was an issue when you opened a dive, edited the dive site, and then continued to edit the dive, and somehow that could mess things up if you discarded your changes.

That's probably the reason that the edit-dive-site button is disabled during dive-edit..?

PS: This PR fixes a user-visible bug, from which I don't know when it was introduced and therefore whether the fix deserves a CHANGELOG entry:
1) Edit name of dive site.
2) Do reverse geo-lookup.
3) Result: Name is reset.

PPS: At least @mturkia's CSV-import fix also deserves a CHANGELOG entry.

dirkhh:
I'd suggest adding CHANGELOG entries when in doubt :-)

PR #1780, opened 2018-10-14
Autogroup only when selected

janmulder:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Comits f427226b3b and 43c3885249f of the undo series introduced 2 calls of autogroup_dives() without checking the autogroup global boolean. This is a bug. An import from DC (for example) then triggers an autogrouping, the divelist is autogrouped, and the UI button is off.

This commit solves this. I've chosen for a guard in the autogroup_dives() that now is a no-op when called when the user did not select autogrouping. In addition, simplified the other calls to this function, as we do
not need to check before calling any more.

### Related issues:
bug introduced in  f427226b3b and 43c3885249f 

### Documentation change:
No

### Mentions:
@bstoeger

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1023,7 +1023,8 @@ dive_trip_t *get_dives_to_autogroup(int start, int *from, int *to, bool *allocat
 }
 
 /*
- * Walk the dives from the oldest dive, and see if we can autogroup them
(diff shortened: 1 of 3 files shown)


bstoeger (approved):
lgtm, thanks.

PR #1782, opened 2018-10-14
Desktop: Enable next/prev DCs only for applicable dives

janmulder:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Does not solve any problem, but might help users that are confused about the next/prev DC menu items, to select a different profile for the currently selected dive. So, enable these menu items only for dives where more than one DC is used.

Signed-off-by: Jan Mulder <jlmulder@xs4all.nl

### Additional information:
Yes, trivial.


bstoeger (approved):
That makes sense.

It was interesting to learn that the null-dive also has one dive-computer.

PR #1784, opened 2018-10-14
Two more tiny parser-related cleanups.

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Two more tiny parser-related cleanups (dead-code removal + unglobalization of a variable). Looks obvious to me, but perhaps I'm missing something. Notably the `sanitize_cylinder_type()` function now looks a bit naked and I wonder if it should be folded into `match_standard_cylinder()`. But I'm not familiar enough with the whole `fixup_dive()` code to do that.

### Changes made:

1) Remove dead code from `sanitize_cylinder_type()`
2) Un-globalize the now non-global `xml_parsing_units`
### Related issues:

None.
### Additional information:

No.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1202,7 +1202,6 @@ static void match_standard_cylinder(cylinder_type_t *type)
 	type->description = p;
 }
 
-
 /*
  * There are two ways to give cylinder size information:
  *  - total amount of gas in cuft (depends on working pressure and physical size)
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
Looks good, I asked @torvalds to take a look as well

torvalds:
Looks fine to me as well.

We don't ever seem to set XML parsing units to IMPERIAL_units, so there is similar dead code wrt the weight and length units, but that seems to be more sensibly in the actual  XML parsing code, so I guess it can/should stay in case we ever see an imperial xml file.

At which point we probably need to do the cylinder sizing again too, but it probably should be done at xml parsing time, so the code removed here might as well go away.

PR #1785, opened 2018-10-14
Parser: fix leakage of dive-site and dive data

bstoeger:
Dive site data was collected in "cur_dive_site", which was then
merged into an existing or a new dive site. But only the struct
dive_site pointed to by "cur_dive_site" and the taxonomy data
were freed, not the textual data such as name or description.
Therefore, split out the approrpriate free-ing from the
delete_dive_site() function and call that instead of a simple
free().

A similar situation occured for dives that would not be added
to the dive-table because they were deemed incomplete. Use
free_dive() here instead of a simple free() too.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A borderline bug-fix: the dive-site parsing was leaking memory of name and description. Likewise, but a much less common case, incomplete dives could perhaps leak data(?). This is all a bit hairy, but at least normal loading still seems to work.

### Changes made:

1) split out free_dive_site().
2) use free_dive_site() and free_dive() instead of plain free().
### Additional information:

None.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/divesite.c
+++ b/core/divesite.c
@@ -40,7 +40,6 @@ uint32_t get_dive_site_uuid_by_gps(degrees_t latitude, degrees_t longitude, stru
 	return 0;
 }
 
-
 /* to avoid a bug where we have two dive sites with different name and the same GPS coordinates
  * and first get the gps coordinates (reading a V2 file) and happen to get back "the other" name,
  * this function allows us to verify if a very specific name/GPS combination already exists */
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
This looks correct. In order not to confuse myself I'll wait merging until I verified that the import crash is fixed.

PR #1786, opened 2018-10-14
Enable smtk-import windows build in containers

salvadorcunat:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [X] Build system change
- [ ] Documentation change
- [ ] Language translation

Quite simple modifications to enable building smtk2ssrf windows package in containerized environment.
At this moment seems impossible to make a shared build of mdbtools (see [mxe's build matrix](https://mxe.cc/build-matrix.html)) so I've built a lighter prebuilt binaries tarball including just Glib-2.0 and its dependencies.
This way we relay on building mdbtools from source.
BTW, libftdi is included, as we are not building a lightweight subsurface in windows.  Curiously in regular windows build libftdi is excluded on purpose.

Changes have been tested in travis (except after_success.sh) and seem to work well. Sadly I haven't figured out yet how to fix the issues I'm having with grantlee and googlemaps in my system when running the container scripts.

### DIFF
--- a/packaging/windows/smtk2ssrf-mxe-build.sh
+++ b/packaging/windows/smtk2ssrf-mxe-build.sh
@@ -162,6 +162,9 @@ fi
 		     --enable-shared=no \
 		     --disable-man \
 		     --disable-gmdb2
+# hack to make mdbtools build outsource
+ln -vs "$BUILDDIR"/mdbtools/include/mdbver.h "$BASEDIR"/mdbtools/include/mdbver.h
+
 make $JOBS >/dev/null && make install || \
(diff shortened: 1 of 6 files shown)


dirkhh:
Nice. With that we can retire the non-container build...
Thanks

dirkhh (approved):
Thanks for this @salvadorcunat 
The symbolic link to find the include file is a bit weird, but it seems to work.
I assume you have tested the resulting binary?

salvadorcunat:
> I assume you have tested the resulting binary?

Yes I've finally managed to build an installer locally using the containers and it works.  BTW I'm going to place an issue about the problems to build locally, although I think they are related to my system as the containerized build works pretty well in travis.

PR #1788, opened 2018-10-15
Dive list: fix crash on dive import

bstoeger:
In commit 8c2383b4952fa22d41745d29484462ed6a67112b dive merging was
changed to not modify the original dive. On import, dives were then
merged and the original deleted. The merge_weightsystem_info() was
not adapted accordingly (deep copy of string instead of pointer),
leading to a use-after-free crash.

Resolve this by doing a deep copy.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix use-after-free (see commit message).

### Changes made:

1) Do a deep copy of the weight-system-description
### Release note:

No - bug introduced in current release cycle.
### Mentions:

@dirkhh: This trivially fixes at least one crash condition, I will audit for more after work.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -2130,7 +2130,8 @@ static void merge_weightsystem_info(weightsystem_t *res, const weightsystem_t *a
 {
 	if (!a->weight.grams)
 		a = b;
-	*res = *a;
+	res->weight = a->weight;
+	res->description = copy_string(a->description);
 }
 
 /* get_cylinder_idx_by_use(): Find the index of the first cylinder with a particular CCR use type.
--- a/core/dive.c
+++ b/core/dive.c
@@ -3475,7 +3475,7 @@ struct dive *merge_dives(const struct dive *a, const struct dive *b, int offset,
 	MERGE_MAX(res, a, b, number);
 	MERGE_NONZERO(res, a, b, cns);
 	MERGE_NONZERO(res, a, b, visibility);
-	MERGE_NONZERO(res, a, b, picture_list);
+	STRUCTURED_LIST_COPY(struct picture, a->picture_list ? a->picture_list : b->picture_list, res->picture_list, copy_pl);
 	taglist_merge(&res->tag_list, a->tag_list, b->tag_list);
 	merge_cylinders(res, a, b, cylinders_map_a, cylinders_map_b);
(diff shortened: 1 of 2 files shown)


dirkhh (commented):
You're the man. I'll test this locally in a minute :-)

janmulder:
And this does not fix my endless loop. (looping in divetripmodel.cpp line 64).

But the fix itself looks sane.

dirkhh:
Ditto here - my crash is still there, and still looks like a memory corruption.
in `process_imported_dives()` we walk the `dive_table.dives[]` array and check if a new dive starts during the dive time of an old dive. And as we do this, the pointers in that array itself appear to get overwritten.

bstoeger:
> And this does not fix my endless loop. (looping in divetripmodel.cpp line 64).

Yes, I also see that - but that too shouldn't be too hard to fix - somehow the trip-list is ill.

bstoeger:
And another one, which also doesn't explain the crash, but should remove (rightful) `valgrind` noise...

bstoeger:
@dirkhh: My problem is that I can't test this any further because I don't have two dive computers, much less two computers with the same dive(s). Could you perhaps create two separate minimal log-files which I can import to see the problem? Or simply post a new valgrind trace without the artifacts of the issues that were fixed in this PR? Thanks!

dirkhh (approved):
These all look reasonable. Merging, then testing again

dirkhh:
Still the same crash. I'm about to head out for diving again - I'll try to work on this as well tonight.

janmulder:
@bstoeger. I can reproduce a crash with the the 2 files I provided earlier (I can attach them again, but you still might have  them):
- opened my normal logbook, and close it
- open the test.ssrf
- import the test-import.ssf
- import the initial test.ssrf again.

On the current master (including your 4 fixes as just merged by Dirk), I get persistent memory corruption style crashes related to the dc list in the dive struct.

bstoeger:
Cool - I can reproduce that - will look into it after dinner.
Since I don't have multiple dive computers the dc-list is hard to test for me. :(

bstoeger:
Does not crash for me with #1794. Could you please check?

PR #1794, opened 2018-10-15
[RFC] ** TESTING DO NOT MERGE ** Core: don't overwrite next in copy_dc_renumber()

bstoeger:
copy_dc_renumber() is an internal function to copy dive computers
and renumber the cylinders. Since the structure was copied this
would also copy the next pointer and thus effectively copy the
whole remaining list.

This meant that on dive-merge two dives shared the same computer
and if one of them is freed, use-after-free crashes would happen.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] Testing

This is a test commit, in which I try to fix a crash reported by @janmulder. In the C-core I'm in full panic mode (what am I doing!?), so please have a close look. @janmulder: This fixes the crash you described, but I fear this is not @dirkhh's crash condition.

Is is correct to restore the ->next pointer?

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -489,14 +489,24 @@ static void copy_dc(const struct divecomputer *sdc, struct divecomputer *ddc)
 
 static void dc_cylinder_renumber(struct dive *dive, struct divecomputer *dc, const int mapping[]);
 
-/* copy dive computer and renumber the cylinders */
+/* copy dive computer list and renumber the cylinders
+ * space for the first divecomputer is provided by the
+ * caller, the remainder is allocated */
 static void copy_dc_renumber(struct dive *d, const struct divecomputer *sdc, struct divecomputer *ddc, const int cylinders_map[])
 {
-	*ddc = *sdc;
-	ddc->model = copy_string(sdc->model);
-	copy_samples(sdc, ddc);
-	copy_events(sdc, ddc);
-	dc_cylinder_renumber(d, ddc, cylinders_map);
+	for (;;) {
+		*ddc = *sdc;
+		ddc->model = copy_string(sdc->model);
+		copy_samples(sdc, ddc);
+		copy_events(sdc, ddc);
+		dc_cylinder_renumber(d, ddc, cylinders_map);
+		if (!sdc->next)
+			break;
+		sdc = sdc->next;
+		ddc->next = calloc(1, sizeof(struct divecomputer));
+		ddc = ddc->next;
+	}
+	ddc->next = NULL;
 }
 
 /* copy an element in a list of pictures */


dirkhh:
I just posted a test case for my crash to the other thread.

dirkhh:
Hehe... this fixes my test case. Let's try the "real" case (my full dive log)... yep, fixed as well. So now I'll need to read the code :-)
But first in indication: 👍

core/dive.c
```diff
@@ -492,7 +492,9 @@ static void dc_cylinder_renumber(struct dive *dive, struct divecomputer *dc, con
 /* copy dive computer and renumber the cylinders */
 static void copy_dc_renumber(struct dive *d, const struct divecomputer *sdc, struct divecomputer *ddc, const int cylinders_map[])
 {
+	struct divecomputer *old_next = ddc->next;
 	*ddc = *sdc;
+	ddc->next = old_next;
```
dirkhh: So that's rather scary. I don't think that's what we should do. Because now the _second_ and all further DCs are those of the original dive.

Here's how that C data structure beauty works...
struct dive CONTAINS a struct divecomputer - so the first dive computer is always part of the dive structure.
And any **additional** dive computer is then attached via the linked list.
So if you copy the `->next` pointer, you basically attach whatever second and up dive computer of ddc were before this function was called to the new dive. I need to read this a couple more times in context, but that does feel wrong.

bstoeger: Yes I understand the data structure. :)
I was hoping that copy_dc_renumber was used to extend the divecomputer-list of the newly generated destination dive. So shouldn't ddc->next always be NULL (because freshly generated)?

If this:
```
        if (a->model && !b->model) {
                copy_dc_renumber(d, a, res, cylinders_map_a);
                return;
        }
```
is supposed to copy all dive computers, then clearly the dc_renumber has to be put in a loop, isn't it....?

dirkhh: Sorry, didn't intend to sound condescending.
I don't think we can assume that ddc->next is always NULL because we could be importing (instead of downloading) and an import could contain more than one DC per dive (which isn't the case in my example, though).

dirkhh: I see something else that's scary, though.

`dive.c:3445 	const struct dive *dl = NULL;`
next use is
`dive.c: 3487  		join_dive_computers(res, &res->dc, &dl->dc, &a->dc, cylinders_map_b, cylinders_map_a, 1);`
Umm, that does seem like a bad idea...

bstoeger: > Sorry, didn't intend to sound condescending.

You didn't.

> I don't think we can assume that ddc->next is always NULL

It should be, because ddc belongs to a freshly constructed dive.

But you are correct - the code is wrong. This all needs to be put in loops to copy the whole list.

dirkhh: oh, and that was introduced in your 8c2383b495

bstoeger: > I see something else that's scary, though.

Good catch - that should by `&b->dc`.

bstoeger: > I get the feeling that you might need to copy the whole chain.

Yes, this is definitely the case. I tried to implement something, but I am not sure I did this correctly. Please check if you find time.

>  And if you feel like you want to rewrite dive.c as C++, I think that's ok. 

Hell no. For many reasons, one of them that I don't even know a small fraction of the use cases (CCR, multi-cylinder, multi-dive computer, deco, ...).

dirkhh: > > And if you feel like you want to rewrite dive.c as C++, I think that's ok.
> 
> Hell no.

That made me laugh :-)

core/dive.c
```diff
@@ -489,14 +489,24 @@ static void copy_dc(const struct divecomputer *sdc, struct divecomputer *ddc)
...
+		copy_samples(sdc, ddc);
+		copy_events(sdc, ddc);
+		dc_cylinder_renumber(d, ddc, cylinders_map);
+		if (!sdc->next)
+			break;
+		sdc = sdc->next;
+		ddc->next = calloc(1, sizeof(struct divecomputer));
+		ddc = ddc->next;
+	}
+	ddc->next = NULL;
```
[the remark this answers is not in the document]
torvalds: This second version at least handles the "merge two and two" case right.

dirkhh:
I think this should remove the TESTING name :-)

core/dive.c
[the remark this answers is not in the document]
torvalds: Hmm. I see a different bug, but it's not new.

The

        *ddc = *sdc;

thing is wrong for the extradata. It happens to work (because I suspect we never change that list, and we never free it), but I think we should do "copy_extra_data()" the same way we do "copy_events()" etc.

Similarly, we do

        ddc->model = copy_string(sdc->model);

to copt the model string, but don't do the same for "serial" and "fw_version".

But as mentioned, this is all pre-existing, and the new copy_dc_renumber() function just duplicates the (mostly harmless) bug that existed in "copy_dc()".

So I think this commit looks fine.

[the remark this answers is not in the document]
bstoeger: > but don't do the same for "serial" and "fw_version".

Yes, I noticed that. Then I realized that these fields are also not freed in free_dc(). So I assumed they were some global things. If this is not the case, this certainly should be cleaned up.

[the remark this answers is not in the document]
dirkhh: I'm happy to entertain a follow up PR to clean up those more "harmless" issues.
Yes, we treat the strings in extra data like const (i.e., we never edit them and never free them), so because of that it's fine to share them, but it would be much more consistent to actually have hem private for each dive computer structure.
But I'd also love to see Travis finish on this one so we can get it merged.
I have some data downloading I want to do and that does merge into dives with multiple DCs...

PR #1796, opened 2018-10-15
deco: _really_ make sure the deco state is fully initialized

torvalds:
I incorrectly thought that 'ci_pointing_to_guiding_tissue' was the only
missing initialization, because that is the only one valgrind pointed at.

... that is, until I started looking at a few more dives, which showed
that there were other parts tht weren't initialized either, like

        double tolerated_by_tissue[16];
        double tissue_inertgas_saturation[16];
        double crushing_onset_tension[16];            // total inert gas tension in the t* moment

so just make sure to clear the whole data structure, to avoid any random
behavior due to uninitialized deco state.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/deco.c
+++ b/core/deco.c
@@ -540,6 +540,8 @@ void clear_vpmb_state(struct deco_state *ds) {
 void clear_deco(struct deco_state *ds, double surface_pressure)
 {
 	int ci;
+
+	memset(ds, 0, sizeof(*ds));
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
yeah, that will do it

PR #1803, opened 2018-10-16
Parser: don't leak extra data key/value pairs

bstoeger:
On parsing of dive computer extra data, key/value pairs are stored
in global state. They are added to the dive computer with
add_extra_data(), which makes a copy of the string. The local
copies of the strings are never freed.

free() the strings after storing them. The data still leaks in case
of unfinished parsing of extra_data tags, but this will be
taken care of in a subsequent commit.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A simple leak-fix, see commit message. Looks obvious, but as usual the devil is in the details.

### Changes made:

1) Free extra-data strings after adding them to the dive-computer.

### Related issues:

None.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/parse-xml.c
+++ b/core/parse-xml.c
@@ -270,6 +270,9 @@ static void extra_data_end(void)
 	// don't save partial structures - we must have both key and value
 	if (cur_extra_data.key && cur_extra_data.value)
 		add_extra_data(get_dc(), cur_extra_data.key, cur_extra_data.value);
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
That looks correct

PR #1807, opened 2018-10-17
Profile: show correct info box for multiple DCs

### LINKED ISSUE #1802: 2 DCs, different deco, information box
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
See [2dc-deco.ssrf.txt](https://github.com/Subsurface-divelog/subsurface/files/2481831/2dc-deco.ssrf.txt)

This is a file with 1 dive with a tricky history. The ultimate effect is that it is a dive with 2 DCs, but the profile is identical, as it is 1 dive imported over itself, where the imported dive had a minor tweak (setting a different DC name, which is just fake). First, I noticed that the deco was a little different (the ssrf computed one) for both DCs. **This is correct**. When I imported the dive back in 2016, the libdc/ssrf code was not able to correctly read the OSTCs setpoint at the start, and also not during a dive, so I added them by hand. The 2nd profile imported again with the current  libdc/ssrf, and sets the setpoint correctly. So, the setpoint changes are different for both profiles, resulting in different deco.

But, while looking into this, I noticed that the information box is **not updated** when you switch using the keyboard shortcuts (and not using the mouse). This is wrong, and seems a simple UI glitch.

### Operating system:
Linux, but not relevent

### Subsurface version:
current master

### Steps to reproduce:
See above

### Current behavior:
See above

### Expected behavior:
Obvious: repaint the information box on DC change

### Additional information:
example file attached.

janmulder:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Update the dive info box when toggling the chosen DC using the keyboard. The keyboard toggle does not generate a mouse move event, so this data was not repainted. For this, a sub-optimization of not repainting when not moving the mouse had to be removed. This does not impact ant performance as 99.9999% of the calls of the repaint are caused by mouse movement anyway.

### Related issues:
Fixes: #1802

### Release note:
not needed IMHO

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -828,6 +828,8 @@ void ProfileWidget2::plotDive(struct dive *d, bool force, bool doClearPictures)
 		clearPictures();
 	else
 		plotPictures();
+
+	toolTipItem->refresh(mapToScene(mapFromGlobal(QCursor::pos())));
 #endif
 
 	// OK, how long did this take us? Anything above the second is way too long,
@@ -1066,7 +1068,8 @@ void ProfileWidget2::scrollViewTo(const QPoint &pos)
 void ProfileWidget2::mouseMoveEvent(QMouseEvent *event)
 {
 	QPointF pos = mapToScene(event->pos());
-	toolTipItem->refresh(pos);
+	toolTipItem->refresh(mapToScene(mapFromGlobal(QCursor::pos())));
+
 	if (zoomLevel == 0) {
 		QGraphicsView::mouseMoveEvent(event);
 	} else {
--- a/profile-widget/divetooltipitem.cpp
+++ b/profile-widget/divetooltipitem.cpp
@@ -240,8 +240,6 @@ void ToolTipItem::refresh(const QPointF &pos)
 	refreshTime.start();
 
 	int time = lrint(timeAxis->valueAt(pos));
-	if (time == lastTime)
-		return;
 
 	lastTime = time;
 	clear();


profile-widget/profilewidget2.cpp
```diff
@@ -828,6 +828,8 @@ void ProfileWidget2::plotDive(struct dive *d, bool force, bool doClearPictures)
 		clearPictures();
 	else
 		plotPictures();
+
+	toolTipItem->refresh(mousePos);
```
bstoeger: Perhaps instead of saving mousePos
```
	QPoint pos = mapFromGlobal(QCursor::pos());
	toolTipItem->refresh(pos);
```
?
Matter of fact - if I quickly move the mouse cursor out of the profile, with the current code I get random values [where the cursor was last registered on the widget] on key-right,left.

janmulder: Agree. The storing of mousePos is the part I do not like as well. That said ... it should be something like `mapToScene(mapFromGlobal(QCursor::pos()))` but that gives a small error compared to `mapToScene(event->pos())` small as in 2 pixels. That results in a small jump of the data in the info box when pressing left/right. Let me see if I can understand where that small difference is coming from.

janmulder: force pushed a new version.

dirkhh (approved):
Just wondering if the commit message should mention the change to the position transformation.

profile-widget/profilewidget2.cpp
```diff
@@ -1066,7 +1068,8 @@ void ProfileWidget2::scrollViewTo(const QPoint &pos)
 void ProfileWidget2::mouseMoveEvent(QMouseEvent *event)
 {
 	QPointF pos = mapToScene(event->pos());
-	toolTipItem->refresh(pos);
+	toolTipItem->refresh(mapToScene(mapFromGlobal(QCursor::pos())));
```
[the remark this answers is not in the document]
janmulder: There is a lot of position magic going on. 

The `event->pos()` and the `mapToScene(mapFromGlobal(QCursor::pos()))` are almost the same. When moving the mouse slowly, I see on my system a (2, 2) pixel offset between the two. So when we would do the 2 refresh calls differently, a tiny jump of info box might occur. Obviously, we do not have an event to pull the pos() from in the plotDive function. 

When moving (very) fast, inside, and inside->outside with the mouse, the differences can be bigger (and even very big). This is also in the current master (without this PR, and likely forever, but nobody ever noticed). Try the following: move fast inside, sudden stop, and click a mouse button (without moving the mouse). Carefully watch the time in the info box. This click repaints the info fox. There is a very big chance that the time will change. In fact, the data in the box was wrong, and is now correct after the repaint. It is wrong in the sense that the mouse it at point A, but the data shows data from point B. Repaint causes data from point A to be drawn.

I found this behavior because it annoyed me, that selecting a different DC (causing a repaint) can change the time of the data in the info box (and all other values). But after realizing that it always has been this way, I sort of accepted this. It seems a fundamental problem that the `QCursor::pos())` and the `event->pos()` can be different when moving fast (just depending when al the handlers are called in the Qt's event loop).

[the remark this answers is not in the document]
janmulder: To obfuscate it a little more :-) The toolTip is drawn at position: `toolTipItem->pos()`. The data inside in it is drawn for the location of the mouse pointer which is controlled by `toolTipItem->refresh(<mouse position>)`. Basically, these are unrelated. Above, I was only talking about the latter, and Dirk's question was about the first.

PR #1808, opened 2018-10-17
Mobile filter

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is mainly code originally done by @janmulder  and then massaged into place by me.
It adds filtering to the mobile UI.
So far only tested on mobile-on-desktop. I'll look at it on device between dives later, but I wanted to get it out for the Europeans to play with.

### Changes made:

1) add a UI action to toggle the filter (this is all Jan's work)
2) add a backend to apply the filtering (mostly Jan's)
3) changed that backend to do full text search instead of just dive sites
4) added a count of filtered dives to the header (and made it an overlay)

### Additional information:

The count on top should be correct, but the count that is part of the trip headers is wrong - it shows the count of all dives (filtered or not) in the trip. That's a bit confusing and needs to be fixed.

### Release note:

TODO

### Documentation change:

TODO

### Mentions:

@janmulder @jbygdell and anyone else who plays with the mobile UI
Which soon will be @mturkia :-)

### DIFF
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -20,6 +20,8 @@ Kirigami.ScrollablePage {
 	property int horizontalPadding: Kirigami.Units.gridUnit / 2 - Kirigami.Units.smallSpacing  + 1
 	property string activeTrip
 	property bool showBusy: false
+	property QtObject diveListModel: diveModel
+	property string numShownText
 
 	supportsRefreshing: true
 	onRefreshingChanged: {
@@ -409,7 +411,6 @@ Kirigami.ScrollablePage {
 				anchors.right: parent.right
 				anchors.leftMargin: Kirigami.Units.gridUnit / 2
 				anchors.rightMargin: Kirigami.Units.gridUnit / 2
-				onVisibleChanged: numShown.text = diveModel.shown()
 				Controls.TextField  {
 					id: sitefilter
 					z: 10
@@ -424,7 +425,6 @@ Kirigami.ScrollablePage {
 						diveModel.setFilter(text)
 						console.log("back from setFilter")
 						showBusy = false
-						numShown.text = diveModel.shown()
 					}
 					onEnabledChanged: {
 						// reset the filter when it gets toggled
@@ -438,10 +438,7 @@ Kirigami.ScrollablePage {
 					id: numShown
 					z: 10
 					verticalAlignment: Text.AlignVCenter
-					// when this is first rendered, the model is still empty, so
-					// instead of having a misleading 0 here, just don't show a count
-					// it gets set whenever visibility or the search text changes
-					text: ""
+					text: numShownText
 				}
 			}
 		}
@@ -452,7 +449,7 @@ Kirigami.ScrollablePage {
 		anchors.fill: parent
 		opacity: 1.0 - startPage.opacity
 		visible: opacity > 0
-		model: diveModel
+		model: page.diveListModel
 		currentIndex: -1
 		delegate: diveDelegate
 		header: filterHeader
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -362,7 +362,7 @@ Kirigami.ScrollablePage {
 		anchors.fill: parent
 		horizontalAlignment: Text.AlignHCenter
 		verticalAlignment: Text.AlignVCenter
-		text: qsTr("No dives in dive list")
+		text: diveListModel ? qsTr("No dives in dive list") : qsTr("Please wait, filtering dive list")
 		visible: diveListView.visible && diveListView.count === 0
 	}
 	Component {
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -415,6 +415,10 @@ Kirigami.ScrollablePage {
 					diveModel.setFilter(text)
 					numShown.text = diveModel.shown()
 				}
+				onVisibleChanged: {
+					// reset the filter when it gets toggled
+					text = ""
+				}
 			}
 			Controls.Label {
 				id: numShown
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -19,7 +19,6 @@ Kirigami.ScrollablePage {
 	property color secondaryTextColor: subsurfaceTheme.secondaryTextColor
 	property int horizontalPadding: Kirigami.Units.gridUnit / 2 - Kirigami.Units.smallSpacing  + 1
 	property string activeTrip
-	property bool showBusy: false
 	property QtObject diveListModel: diveModel
 	property string numShownText
 
@@ -327,13 +326,6 @@ Kirigami.ScrollablePage {
 		}
 	}
 
-	Controls.BusyIndicator {
-		running: showBusy
-		z: 10
-		anchors.fill: parent
-		anchors.margins: Kirigami.Units.gridUnit
-	}
-
 	StartPage {
 		id: startPage
 		anchors.fill: parent
@@ -419,12 +411,9 @@ Kirigami.ScrollablePage {
 					text: ""
 					placeholderText: "Full text search"
 					onAccepted: {
-						showBusy = true
-						console.log("show busy")
 						rootItem.filterPattern = text
 						diveModel.setFilter(text)
 						console.log("back from setFilter")
-						showBusy = false
 					}
 					onEnabledChanged: {
 						// reset the filter when it gets toggled
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -363,14 +363,58 @@ Kirigami.ScrollablePage {
 		text: qsTr("No dives in dive list")
 		visible: diveListView.visible && diveListView.count === 0
 	}
-
 	Component {
 		id: filterHeader
 		RowLayout {
+			id: filterBar
+			states: [
+				State {
+					name: "isVisible"
+					when: rootItem.filterToggle
+					PropertyChanges { target: filterBar; visible: true; height: sitefilter.implicitHeight }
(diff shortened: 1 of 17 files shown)


janmulder (commented):
1) I think we need an apply button in the search bar. I do not like the behavior when typing letters, sometimes the focus gets lost, and we might be spending to much CPU time resetting that search string for every letter typed. So the user-experience will be of a sluggish UI.

2) I would like a search panel that animates from the top down, and pushes the dive list down. This is pure UI thing.

3) I'm not sure I like the full text search case insensitive. For example. Search for buddy Rick, also results in the dives with buddy Patrick.

dirkhh:
1. as I said, only tested on laptop where performance and focus weren't an issue. Adding an apply button would be very easy
2. also doable, simply requires a transition for the visual effect; what's harder is to have it push the list down and not cover the first dive
3. this turns into a user preference / number of options thing; I personally would want the search to be case insensitive. If I search for "frog fish" I don't care if it's at the beginning of a sentence or not. I also know that people will start arguing about searching the beginning of certain text strings only (we had that argument about location names before). Etc. I'm happy to entertain certain options here, but it has to stay reasonable. Especially if we add an 'apply' button we will already start getting tight on space - and the last thing I want to do is turn this into two lines.

mturkia:
Isn't the full text missing notes?

janmulder:
1) I have focus issues on desktop as well. Not tried it on device yet.
2) Like I said: pure UI, not the most important thing at this point.
3) Agree. Not a multi line thing.

new point 4) set filter on. Search, select a dive far away (old), set filter off. As the dive is far out of the view presented, it starts scrolling. Which takes forever on a big logbook.

new point 5) see 4), but now we arrived at the chosen dive. The trip where that dive is in is collapsed, so you do not see the dive you have selected before. Very confusing.

janmulder:
@mturkia: currently it does not look in notes. Not sure this is by design or just forgotten.

bstoeger:
Not directly related to this PR - but can someone explain why we need a separate list-model for mobile?

dirkhh:
- clear button - let's reset the filter when toggling and see how that works.
- focus issues - I may need help with that, let's see
- I'm sure how you get 52 results for "tes" on mobile but not in Subsurface - but I have a suspicion. I'll play with that.
Thanks to both of you for the testing. Really appreciated

dirkhh:
As always with QML, things that should be easy take an inordinately long time while you try to figure out why it doesn't do what you want it to do.
Here's what I have implemented so far:
- filter animation: this should work as it exactly matches other animations we use, but at least when run on the desktop it doesn't look the way I expected it to look. It also still doesn't push down the list (so it covers the first item)... now that I write this I have a brilliant idea... why is this a Header and not a Footer... maybe that would look better as well. Oh, wait, but that then collides with the action button, doesn't it? darn.
- filter clears when toggled
- filtering only starts when you hit enter
- I added notes to the full text search and also added a separator between the concatenated strings. My suspicion is that when @jbygdell got weird match counts it was matching the end of one string plus the beginning of the next string. E.g., do you have buddies with last letter 't' and dive masters starting with 'es' (or 'te' and 's')?

As usual, I'd love to get this in more people's hands earlier, so let's please focus on what has to be fixed before this can go into master / create beta test binaries... (but let's still keep track of everything else that needs to be addressed... especially the scrolling issue worries me)

dirkhh:
@bstoeger 
> Not directly related to this PR - but can someone explain why we need a separate list-model for mobile?

I'm guessing this was because of the complexity of the list/tree model we use on the desktop

jbygdell (commented):
This fixed the faulty search result.

jbygdell:
I spent some time last night trying to fix the focus issues, one is still not fixed.
This is the behavior I'm seeing on my Android device:
When the filter view becomes visible, it is not active and the keyboard is not visible.
So now I have to tap in the text input for this to happen.

dirkhh:
With the latest push I have the keyboard opening reliably, but sometimes (on Android) the text input field isn't shown until I scroll "up" a tiny bit in the dive list (i.e., drag my finger down to show newer dives). And there is an ugly UI glitch where it stays frozen for a bit with the keyboard removed but the dive list not repainted while it waits to match the right dives.
It should be easy to have it show "something else" there, but as usual with QML, my first couple of attempts didn't work :-)

dirkhh:
OK, lots of updates. This might be ready to be merged. It's far from perfect, but I think it's viable.
Known issues:
- on Android I usually need to pull down the dive list a tiny bit before the input field for the filter (the ListView header) is shown. No idea why - I tried MANY things to fix that, no luck).
- the trip headers still show the unfiltered number of dives - I think it may be better with the filtered number, but that's sadly compute intensive to implement.
- sometimes on Android I don't get the virtual keyboard. I tried to hack around that, but it's still not 100%

Please test and comment!

mobile-widgets/qml/DiveList.qml
```diff
@@ -421,6 +421,16 @@ Kirigami.ScrollablePage {
 							forceActiveFocus()
 						}
 					}
+					onFocusChanged: {
```
jbygdell: This block gives another issue, the keyboard is not hidden when pressing to hide the filter view.
Removing the onFocusChanged block solves this issue for me.

dirkhh: That's wild since I actually ask to hide it RIGHT THERE.
But without this block I could easily reproduce a situation where I had no keyboard and tapping on the input field didn't get me a keyboard, either. So I'm not sure what to do here... maybe don't try to explicitly hide it?

dirkhh: Yeah, that's what bugs me the most. How can we write an app that works for all (or at least most) of our users???
Not sure what to do here - is there another way to ensure that the keyboard is shown when the TextField is entered?

jbygdell: Activating the textfield by tapping on it will always bring up the keyboard in my tests. And the forceActiveFocus fixes the issue with displaying the keyboard when opening the filter view.

dirkhh:
OK, now performance is MUCH better. That's more like it.
All we have to figure out is that pesky keyboard / focus issue. I dropped the onFocusChanged commit.

dirkhh:
I think this is ready to move to master. @jbygdell @bstoeger @janmulder any concerns?

mobile-widgets/qmlmanager.cpp
```diff
@@ -1847,6 +1849,17 @@ void QMLManager::showDownloadPage(QString deviceString)
 	emit pluggedInDeviceNameChanged();
 }
 
+void QMLManager::setFilter(const QString filterText)
+{
+	// show that we are doing something, then do something in another thread in order not to block the UI
+	QMetaObject::invokeMethod(qmlWindow, "showBusy");
+	QFuture<void> future = QtConcurrent::run(QThreadPool::globalInstance(),
```
bstoeger, line 1856: `future` is never used - am I missing something?

qt-models/divelistmodel.h
```diff
@@ -12,11 +12,21 @@ class DiveListSortModel : public QSortFilterProxyModel
...
 public slots:
 	int getDiveId(int idx);
 	int getIdxForId(int id);
+	void setFilter(QString f);
+	void resetFilter();
+	int shown();
+protected:
+	bool filterAcceptsRow(int source_row, const QModelIndex &source_parent) const;
+private:
+	std::vector<bool> filteredRows;
```
bstoeger, line 27: Not important, but note that `std::vector<bool>` has a design mistake which turns it into a bitfield (!). Thus, it tends to be much less efficient than one might expect.

bstoeger:
At a quick glance it looks reasonable, but I'm no well-versed enough in QML to form a proper opinion.

qt-models/divelistmodel.h
dirkhh: Oh, so I should make this a std::vector<char> or something like that?

mobile-widgets/qmlmanager.cpp
dirkhh: Good catch. I thought I'd use a FutureWatcher to know when this is done - but then ended up simply doing the call to hide in the lambda

(BTW: I can't believe that no one has called me out of using a lambda in my implementation - I was certain that I'd get some serious ribbing over this one...)

qt-models/divelistmodel.h
dirkhh: The reason I'm asking, in case that's not obvious... the main reason to "roll my own" implementation of the filter was performance, so I really wouldn't want to to then use a poor implementation of a vector and lose some of the advantage I'm trying to get...

bstoeger: > Oh, so I should make this a std::vector or something like that?

Out of principle, probably yes. Realistically, the microsecond more-or-less will not be perceivable.

dirkhh:
@janmulder thanks for the review and encouragement

at 1 & 2) you don't think it's ADORABLE? Ok, maybe not. Yeah, that thing is obviously entirely unstyled and there is a ton of room for improvement. No argument.
at 3) that's odd - it should have height zero when not in use. There was some visual oddity when I made it invisible early in my tests, but again, easy enough to play with some more.
at 4) yes. I actually looked into fixing that and that gave me the idea for the much faster filter... and then I forgot to fix those numbers.
at 4) (you had two 4s) I hadn't noticed that it wasn't reset - should also be easy to fix. I'd love to hear other design ideas how that result should be presented (but I think it's an important one because that way you can easily answer questions like "how many dives did I do in my dry suit")

dirkhh:
manually merged. I couldn't figure out how to change the color of the busy indicator without completely rolling my own. What the heck? But at least I made it smaller.
And I fixed the bug with the count of found elements not being reset.
So (1) and the the first half of the second (4) above are addressed. The rest still need to be done. But I merged it into master for now. I'll be spending most of tomorrow on airplanes, so feel free to play with this.

PR #1810, opened 2018-10-19
Dive site: select dive site when entering edit mode

### LINKED ISSUE #1809: Add/Edit location problem
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Issue for @bstoeger. Use my simple 1 dive/2DC logbook example. Select the globe to edit the location. The map window says: drag the selected dive location. But the flag is not selected so you cannot drag. Problem introduced in commit 0aef04352.

### Operating system:
Linux (but not OS dependent)

### Subsurface version:
introduced by commit 0aef04352

### Steps to reproduce:
See above

### Current behavior:
See above

### Expected behavior:
As before this commit. The flag shall be selected to it can be moved.

### Additional information:
maybe some relation to #1804, but as that one a a very non-specific issue, that raises more questions, I decided to make this specific new issue.

### Mentions:
@bstoeger

bstoeger:
Commit 0aef04352a3210a6024f860758af466ea774dd5e made it impossible
to move new dive sites on the map. When entering dive-site-edit mode,
the filter would be instructed to show only dives of the corresponding
dive site [which didn't yet exist] and therefore all dive sites
on the map were deselected.

Fix this by explicitly centering on the dive site to be edited in
MapWidgetHelper::enterEditMode().

Fixes #1809

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a bug which made it impossible to move the dive site on the map. See commit message.

This one-liner seems obviously correct, but I wonder if the whole thing needs a bit of a refactoring. Notably, I'm not sure if it is a good idea that all other dive sites are hidden. Since usually I don't have precise GPS coordinates, I'd like to get the dive sites right at least in a relative manner (DS 1 was slightly to the east of DS2, etc.).

### Changes made:

1) Center on dive site when going into edit mode.
### Related issues:

Fixes #1809.

### Release note:

Not needed - bug introduced during release cycle(?).
### Documentation change:

No.
### Mentions:

@janmulder reported the problem.

### DIFF
--- a/map-widget/qmlmapwidgethelper.cpp
+++ b/map-widget/qmlmapwidgethelper.cpp
@@ -302,6 +302,7 @@ void MapWidgetHelper::enterEditMode(quint32 uuid)
 	} else {
 		coord = exists->coordinate();
 	}
+	centerOnDiveSiteUUID(uuid);
 	emit coordinatesChanged(degrees_t { (int)lrint(coord.latitude() * 1000000.0) },
 				degrees_t { (int)lrint(coord.longitude() * 1000000.0) });
 	emit editModeChanged();


janmulder:
ACK for this one-liner.

About showing all sites when editing. Yes, that seems more logical to me. I would also help in the scenario that you are about to add a double site because you did not realize you had that site already on an earlier dive.

dirkhh:
I agree with both of you - showing all dive sites around the proposed (?) GPS location seems like the right thing to do. I think I'll merge this one as it fixes a bug, but the suggestion for future work includes showing all of them :-)

PR #1812, opened 2018-10-20
Add 'location_t' data structure

torvalds:
Instead of having people treat latitude and longitude as separate
things, just add a 'location_t' data structure that contains both.

Almost all cases want to always act on them together.

This is really just prep-work for adding a few more locations that we
track: I want to add a entry/exit location to each dive (independent of
the dive site) because of how the Garmin Descent gives us the
information (and hopefully, some day, other dive computers too).

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/qt-models/divelocationmodel.cpp
+++ b/qt-models/divelocationmodel.cpp
@@ -43,8 +43,8 @@ QVariant LocationInformationModel::getDiveSiteData(const struct dive_site *ds, i
 		switch(column) {
 		case UUID: return ds->uuid;
 		case NAME: return ds->name;
-		case LATITUDE: return ds->latitude.udeg;
-		case LONGITUDE: return ds->longitude.udeg;
+		case LATITUDE: return ds->location.lat.udeg;
+		case LONGITUDE: return ds->location.lon.udeg;
 		case COORDS: return "TODO";
 		case DESCRIPTION: return ds->description;
 		case NOTES: return ds->name;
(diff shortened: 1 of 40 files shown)


dirkhh (approved):
Nice. Makes things more readable.

neolit123:
:+1: 
seems like the smtk-import tool needs an update too:
https://api.travis-ci.org/v3/job/444161031/log.txt
(errors near the bottom of the log)

neolit123 (approved):

qt-models/divelocationmodel.cpp
```diff
@@ -123,28 +123,25 @@ bool GPSLocationInformationModel::filterAcceptsRow(int sourceRow, const QModelIn
...
+	location = locationIn;
 	invalidate();
 }
 
-void GPSLocationInformationModel::setCoordinates(degrees_t latitudeIn, degrees_t longitudeIn)
+void GPSLocationInformationModel::setCoordinates(const location_t &locationIn)
 {
-	latitude = latitudeIn;
-	longitude = longitudeIn;
+	location = locationIn;
```
bstoeger: `location = location` is a noop.

bstoeger (approved):
lgtm - excuse my erroneous comment.

Intuitively I would pass by value instead of const reference, since both fit into a 64-bit register. But that's an irrelevant detail.

torvalds:
Yeah, I'll fix that nasty smartrak thing, and also add a helper to create locations from floating point latitude/longitude, since there are a number of places that do that (I created one for the special case of the qmlmapdigethelper, but I'll do the general case).

torvalds:
Ok, pushed updated branch. Further slight cleanups, and the smartrak thing hopefully is right now too.

PR #1813, opened 2018-10-21
mxe-based-build.sh comments: Also build libftdi1

sfuchs79:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Add libftdi1 in comments in MXE build script.

### DIFF
--- a/packaging/windows/mxe-based-build.sh
+++ b/packaging/windows/mxe-based-build.sh
@@ -27,7 +27,7 @@
 #
 # now you can start the build
 #
-# make libxml2 libxslt libusb1 libzip libssh2 curl qt5 nsis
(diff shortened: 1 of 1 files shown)


neolit123 (approved):
LGTM, but it's only needed for FTDI=ON?

dirkhh:
Given that we have a command line option to turn building against libftdi off, I guess we should have language that explains that in the comment... I mean your patch isn't wrong because by default we do need this, but it's also potentially misleading because you can still build Subsurface with MXE without libttdi1 installed...

dirkhh:
slightly edited and merged

PR #1815, opened 2018-10-21
Mobile UI: make SsrfSwitch resizable

jbygdell:

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Back in 8ab8a67f the checkbox where made resizable.
This applies the same functionality to the selector switch and makes the
two objects match in vertical size.

### DIFF
--- a/mobile-widgets/qml/SsrfSwitch.qml
+++ b/mobile-widgets/qml/SsrfSwitch.qml
@@ -6,8 +6,8 @@ import org.kde.kirigami 2.4 as Kirigami
 Switch {
 	id: root
 	indicator: Rectangle {
-		implicitWidth: Kirigami.Units.largeSpacing * 2.2
-		implicitHeight: Kirigami.Units.largeSpacing * 0.75
+		implicitWidth: 36 * PrefDisplay.mobile_scale
+		implicitHeight: 12 * PrefDisplay.mobile_scale
 		x: root.leftPadding
 		y: parent.height / 2 - height / 2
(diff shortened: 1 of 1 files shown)


jbygdell:
Derek add label: ui-mobile

jbygdell:
Derek add label: complexity-low

mobile-widgets/qml/SsrfSwitch.qml
```diff
@@ -18,8 +18,8 @@ Switch {
 		Rectangle {
 			x: root.checked ? parent.width - width : 0
 			y: parent.height / 2 - height / 2
-			width: Kirigami.Units.largeSpacing * 1.1
-			height: Kirigami.Units.largeSpacing * 1.1
+			width: 20 * PrefDisplay.mobile_scale
+			height: 20 * PrefDisplay.mobile_scale
```
dirkhh, line 22: not disagreeing with the change, but what I find odd is that line 9 to line 21 used to have about 2x difference, vs. it now is a 1.8 difference.
I assume there's a good reason for that and it now looks better? :-)

jbygdell:
Given that the current dize of the switch is a pinhead in comparison to the checkbox. The logical choice was to make them equal in height, inner and outer. While keeping as close as possible to the original ratio of the switch.

dirkhh:
That makes sense. Thanks

PR #1816, opened 2018-10-21
Parser: Make parser mostly reentrant

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Before doing the dive-site parsing changes [replace UUIDs by pointers] it seemed appropriate to isolate the whole global parser state. After this change-set the parser still accesses a few global variables [such as autogroup], but that should be variables that are accessed from other parts of the code.

This was mostly done "peephole" style. But it seems to read my own log and runs the tests.

There is one potentially debatable thing noted in the commit-message:
> A somewhat controversial point might be that the individual parsing functions are split in those that need parser-state and those that don't. This means that there are now two versions of the MATCH macro, viz. one for the former and one for the latter.

### Changes made:

1) Make two free_*() functions NULL aware.
2) Introduce a parser_state structure.

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -765,9 +765,11 @@ void insert_trip_dont_merge(dive_trip_t *dive_trip)
 /* free resources associated with a trip structure */
 void free_trip(dive_trip_t *trip)
 {
-	free(trip->location);
-	free(trip->notes);
-	free(trip);
+	if (trip) {
+		free(trip->location);
+		free(trip->notes);
+		free(trip);
+	}
 }
 
 /* remove trip from the trip-list, but don't free its memory.
--- a/core/import-cobalt.c
+++ b/core/import-cobalt.c
@@ -86,23 +86,8 @@ static int cobalt_location(void *param, int columns, char **data, char **column)
 {
 	UNUSED(columns);
 	UNUSED(column);
-	struct parser_state *state = (struct parser_state *)param;
-
-	static char *location = NULL;
-	if (data[0]) {
-		if (location) {
-			char *tmp = malloc(strlen(location) + strlen(data[0]) + 4);
-			if (!tmp)
-				return -1;
-			sprintf(tmp, "%s / %s", location, data[0]);
-			free(location);
-			location = NULL;
-			state->cur_dive->dive_site_uuid = find_or_create_dive_site_with_name(tmp, state->cur_dive->when);
-			free(tmp);
-		} else {
-			location = strdup(data[0]);
-		}
-	}
+	char **location = (char **)param;
+	*location = data[0] ? strdup(data[0]) : NULL;
 	return 0;
 }
 
@@ -116,6 +101,7 @@ static int cobalt_dive(void *param, int columns, char **data, char **column)
 	struct parser_state *state = (struct parser_state *)param;
 	sqlite3 *handle = state->sql_handle;
 	char *err = NULL;
+	char *location, *location_site;
 	char get_profile_template[] = "select runtime*60,(DepthPressure*10000/SurfacePressure)-10000,p.Temperature from Dive AS d JOIN TrackPoints AS p ON d.Id=p.DiveId where d.Id=%d";
 	char get_cylinder_template[] = "select FO2,FHe,StartingPressure,EndingPressure,TankSize,TankPressure,TotalConsumption from GasMixes where DiveID=%d and StartingPressure>0 and EndingPressure > 0 group by FO2,FHe";
 	char get_buddy_template[] = "select l.Data from Items AS i, List AS l ON i.Value1=l.Id where i.DiveId=%d and l.Type=4";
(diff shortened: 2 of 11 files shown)


bstoeger:
For documentation purposes, a list of static variables in the parser after this PR:

1) In `parse_float()`, `first_time` is used to emit a single warning if floating point has a comma. A comment suggests that this could be replaced with a test for subsurface files (which is now available in the parser_state structure).
2) In `add_dive_site()`, `suffix` is used to create dive site numbers. This is questionable anyway [you want a variable per dive site name, not a global one], but let's have a closer look when refactoring the dive site code.
3) In `cobalt_location()` we remember the string of the location. This should probably be moved to `parser_state`.

core/parse.h
```diff
@@ -31,67 +20,95 @@ struct parser_settings {
...
+	int lastcylinderindex, next_o2_sensor;
+	int o2pressure_sensor;
+	struct extra_data cur_extra_data;
+	struct units xml_parsing_units;
+	struct dive_table *target_table;	/* non-owning */
+
+	sqlite3 *sql_handle;			/* for SQL based parsers */
+	event_allocation_t event_allocation;
+};
+#define cur_event event_allocation.event
```
neolit123, line 66: add empty line above this line.

core/import-cobalt.c
```diff
@@ -73,20 +73,19 @@ extern int cobalt_buddies(void *handle, int columns, char **data, char **column)
...
 
-extern int cobalt_location(void *handle, int columns, char **data, char **column)
+static int cobalt_location(void *param, int columns, char **data, char **column)
 {
-	UNUSED(handle);
 	UNUSED(columns);
 	UNUSED(column);
+	struct parser_state *state = (struct parser_state *)param;
 
 	static char *location = NULL;
```
neolit123: i guess it makes sense to be moved in the state as a generic "location" or possibly "cobalt_location" param.

core/parse.h
```diff
@@ -31,67 +20,94 @@ struct parser_settings {
...
+void sample_start(struct parser_state *state);
+void sample_end(struct parser_state *state);
+void divecomputer_start(struct parser_state *state);
+void divecomputer_end(struct parser_state *state);
+void userid_start(struct parser_state *state);
+void userid_stop(struct parser_state *state);
 void utf8_string(char *buffer, void *_res);
 
-void add_dive_site(char *ds_name, struct dive *dive);
+void add_dive_site(char *ds_name, struct dive *dive, struct parser_state *state);
```
neolit123, line 110: +1 for moving suffix to state - e.g. dive_site_suffix.

neolit123 (approved):
thanks for the refactor. added some minor comments.
the global state of the parser has not caused us serious problems, but this seems like a good RFC.

also +1 for removing static variables.
i have not examined if CI passes.

core/parse.h
bstoeger: Stupid question: Is this `SSRF_WS` thing the webservice that was recently removed or is this something completely different? I.e. could we simply remove the code?

bstoeger: Stupid question: Is this `SSRF_WS` thing the webservice that was recently removed or is this something completely different? I.e. could we simply remove the code?

bstoeger: Stupid question: Is this `SSRF_WS` thing the webservice that was recently removed or is this something completely different? I.e. could we simply remove the code?

bstoeger: Stupid question: Is this `SSRF_WS` thing the webservice that was recently removed or is this something completely different? I.e. could we simply remove the code?

neolit123: ^ seems like so, but would defer to @dirkhh on the decision.
(github is spazzing today i see comments repeated 6 times)

neolit123:
> Stupid question: Is this SSRF_WS thing the webservice that was recently removed or is this something completely different? I.e. could we simply remove the code?

seems like so, but would defer to @dirkhh on the decision.
(github is spazzing today i see comments repeated 6 times)

core/parse.h
bstoeger: Stupid question: Is the whole `SSRF_WS` business related to the web-service that was removed recently? I.e. could the whole code simply be deleted?

bstoeger: Argh.. Sorry. I tried again and again and it would not display anything for me. Perhaps my browser.

bstoeger (commented):
I think the problem was that I inadvertenly started a review. Trying to close that now.. :/

core/parse.h
dirkhh: no, GitHub was unwell - I saw the emails about you commenting, but couldn't see the comments on GitHub. Maybe some of their distributed infrastructure had outdated caches or something.
Anyway, yes, that SSRF_WS was indeed about the soon-to-be-discontinued GPS web service. I merged the commit removing it.

core/divelist.c
```diff
@@ -765,9 +765,11 @@ void insert_trip_dont_merge(dive_trip_t *dive_trip)
 /* free resources associated with a trip structure */
 void free_trip(dive_trip_t *trip)
 {
-	free(trip->location);
-	free(trip->notes);
-	free(trip);
+	if (trip) {
```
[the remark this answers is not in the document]
bstoeger: Yes. But I tried to model after the other `free_*()` functions, which seem to prefer the 
```
	if (x) {
		free(...)
	}
```
scheme. I would convert them all in one go, if this is desired.

core/import-cobalt.c
bstoeger: I added a second commit, which makes the code more explicit. But this is only compile tested! Do we have Cobalt files? I don't see any obvious things in the dives-directory.

bstoeger:
Replying via e-mail since for some reason I can't post on github!?

> -void add_dive_site(char *ds_name, struct dive *dive);
> +void add_dive_site(char *ds_name, struct dive *dive, struct parser_state
> *state);
> 
> +1 for moving suffix to state - e.g. dive_site_suffix.

Stupid question: Is the whole `SSRF_WS` business related to the web-service 
that was removed recently? I.e. could the whole code simply be deleted?

dirkhh:
Typo in the last commit message: on vs. one
I'm still reading and re-reading the big commit. And with travel and jet lag and now conference I haven't quite found enough time to do a credible review. Sorry. Hopefully later today.

core/parse.h
```diff
@@ -31,67 +20,95 @@ struct parser_settings {
...
+};
+
+/*
+ * parser_state is the state needed by the parser(s). It is initialized
+ * with init_parser_state() and resources are freed with free_parser_state().
+ * "owning" marks pointers to objects that are freed in free_parser_state().
+ * In contrast, "non-owning" marks pointers to objects that are owned
+ * by other data-structures.
+ */
+struct parser_state {
```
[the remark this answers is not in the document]
dirkhh: there is no rule. or I should say "there no longer is"
Linus initially had some ideas that having a XXX_t meant that this was a type by itself vs a convenient collection of variables (or something like this - I'd have to go through my email archive for more detail), but I think this has long long gone out the window.

dirkhh (approved):
Somehow reading this a third time it suddenly all made sense. No idea why I was stuck the first two times I tried.
Great effort. I hate huge commits like this, but this was the only way to do it, I think.
I agree with the MATCH/MATCH_STATE split.

dirkhh:
last commit message slightly edited and all of it merged.

PR #1818, opened 2018-10-23
Remove (most) dive-site UUIDs in UI code

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Replace UUID-references to dive-sites by classical pointers. But don't change the UUID logic itself (i.e. the UUID encodes name + first timestamp).

This will make implementation of undo in analogy to the dive list (using owning pointers) easier. Moreover, it will make it possible to remove the UUID logic, which is somewhat questionable in the first place.

Briefly tested on mobile & desktop using a modest dive log.

### Changes made:

Too long to list - see commit-message.

### Release note:

 
Not needed.

### Documentation change:

Not needed.


bstoeger:
Another piece of the puzzle: remove UUID from the uemis_helper struct. Unfortunately, I have no way of testing this..

bstoeger:
...aaand also remove UUIDs from the Location-models...

This commit is a bit larger and features a funny logic bug (see commit description).

bstoeger:
Replaced UUID by pointer in mobile code. But cannot test this:
```
$ ./subsurface-mobile
INFO: QQmlApplicationEngine failed to load component
INFO: qrc:///qml/main.qml:548 Type DiveList unavailable
qrc:///qml/DiveList.qml:3 module "QtQuick.Controls" version 2.4 is not installed

can't create window object
```
Seems like a system update upgraded to QtQuick.Controls 2.5!? Running `build.sh` does not help. :(

dirkhh:
Which version of Qt do you have installed?
I don't think you NEED version 2.4 of Controls2, but I don't recall why I revved the version in the first place. My first suggestion would be to go back to 2.2 and see if it works

bstoeger:
Going back to 2.2 works, that should do for testing. Hopefully this doesn't get included in a commit (like all the libdivecomputer commits).

I have all the recent packages installed, I think:
```
bs@laptop:~/src/subsurface/build-mobile$ sudo apt install qml-module-qtquick-controls2
Reading package lists... Done
Building dependency tree       
Reading state information... Done
qml-module-qtquick-controls2 is already the newest version (5.9.5-0ubuntu2).
0 upgraded, 0 newly installed, 0 to remove and 11 not upgraded.
bs@laptop:~/src/subsurface/build-mobile$ sudo apt install qml-module-qtquick-controls
Reading package lists... Done
Building dependency tree       
Reading state information... Done
qml-module-qtquick-controls is already the newest version (5.9.5-0ubuntu1).
0 upgraded, 0 newly installed, 0 to remove and 11 not upgraded.
bs@laptop:~/src/subsurface/build-mobile$ sudo apt install qtquickcontrols2-5-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
qtquickcontrols2-5-dev is already the newest version (5.9.5-0ubuntu2).
0 upgraded, 0 newly installed, 0 to remove and 11 not upgraded.
```
I hope that an upgrade to Kubuntu 10.10 will fix that..?

bstoeger:
Added a commit to replace the UUIDs in MapLocationModel. This one is a bit weird, because with my QML-version all QVariant<void *> compare as equal. I'm not sure if this is my fault or that QML is broken beyond any repair. So what it does is use uintptr_t instead. Wow. But it seems to work.

Not yet tested on mobile(-on-desktop). Will do later.

bstoeger:
> Not yet tested on mobile(-on-desktop). Will do later.

Tested it - and of course it breaks the map. 😛

bstoeger:
There were more lingering `void *`s. Mobile seems to work now.

bstoeger:
Cool - Travis found some parser bugs. No time today - will look into it tomorrow.

bstoeger:
@neolit123, @dirkhh: Any comments? This is necessary for integrating dive-sites into the undo system. While this in itself is not a top-priority issue I'd hate to have this PR bit-rot. The chances for that are high, as it touches many parts of the code. Moreover, there was discussion on the ML with respect to removing the dive-site UUIDs altogether. This PR does most of the grunt work in this direction.

dirkhh:
I'll share a secret with you. When in a rush (which I am roughly 95% of the time I ever look at Subsurface), I tend to ignore PRs that are WIP, TEST, RFC, etc, and focus on those that indicate that they are ready to go in. So I've peeked at this one a couple of times, but never really read through it.
I know that's wrong, especially since it explicitly asked for testing, but that's just the reality...

Anyway, I should have time later today to go through it and review it.

bstoeger:
> Anyway, I should have time later today to go through it and review it.

Cool, thanks.

neolit123 (approved):
had a look at all commits, SGTM.
this needs beta testing.

core/datatrak.c
```diff
@@ -204,7 +204,7 @@ unsigned char *dt_dive_parser(unsigned char *runner, struct dive *dt_dive)
 	ds = get_dive_site_by_name(buffer);
 	dt_dive->dive_site_uuid = ds ? ds->uuid : 0;
 	if (dt_dive->dive_site_uuid == 0)
-		dt_dive->dive_site_uuid = create_dive_site(buffer, dt_dive->when);
+		dt_dive->dive_site_uuid = create_dive_site(buffer, dt_dive->when)->uuid;
```
dirkhh: I was gonna say "but what about a NULL pointer?" - and then I read the code of `alloc_or_get_dive_site()` and the fact that we simply do an exit(1) if we run out of memory. Oh. OK. That makes sure there's never a NULL pointer returned, I guess... 😱

core/divesite.c
```diff
@@ -145,13 +145,15 @@ int nr_of_dives_at_dive_site(struct dive_site *ds, bool select_only)
...
 }
 
-bool is_dive_site_used(uint32_t uuid, bool select_only)
+bool is_dive_site_used(struct dive_site *ds, bool select_only)
 {
 	int j;
 	bool found = false;
 	struct dive *d;
+	if (!ds)
+		return 0;
```
dirkhh: we should return `false`

desktop-widgets/locationinformation.cpp
```diff
@@ -521,7 +521,7 @@ void DiveLocationLineEdit::itemActivated(const QModelIndex &index)
 		qDebug() << "Setting a Existing dive site";
 	if (view->isVisible())
 		view->hide();
-	emit diveSiteSelected(currUuid);
+	emit diveSiteSelected();
```
dirkhh, line 520: this comment is for the commit message, not for this code... not sure how GitHub will present this comment, so this is about commit https://github.com/Subsurface-divelog/subsurface/pull/1818/commits/3fb3b11f0c0e2f7ad4d953e3f6d0b7041edb1e27 
The commit message says "... receiving slot would not use that signal" - I think it should say "... would not use that argument".

desktop-widgets/locationinformation.cpp
```diff
@@ -507,15 +506,14 @@ void DiveLocationLineEdit::focusOutEvent(QFocusEvent *ev)
...
 	QModelIndex idx = index;
-	if (index.column() == LocationInformationModel::UUID)
+	if (index.column() == LocationInformationModel::DIVESITE)
 		idx = index.model()->index(index.row(), LocationInformationModel::NAME);
 
-	QModelIndex uuidIndex = index.model()->index(index.row(), LocationInformationModel::UUID);
-	uint32_t uuid = uuidIndex.data().toInt();
-	currType = uuid == 1 ? NEW_DIVE_SITE : EXISTING_DIVE_SITE;
-	currUuid = uuid;
+	dive_site *ds = (dive_site *)index.model()->index(index.row(), LocationInformationModel::DIVESITE).data().value<void *>();
```
dirkhh: Yuck. That is ugly as heck. And if I run into this line while chasing a bug, I'll claw my eyes out.
Can you do me a favor and add a comment what this actually DOES?

desktop-widgets/mapwidget.cpp
```diff
@@ -66,8 +67,8 @@ void MapWidget::centerOnDiveSite(struct dive_site *ds)
 void MapWidget::centerOnIndex(const QModelIndex& idx)
 {
 	CHECK_IS_READY_RETURN_VOID();
-	struct dive_site *ds = get_dive_site_by_uuid(idx.model()->index(idx.row(), 0).data().toUInt());
-	if (!ds || !dive_site_has_gps_location(ds))
+	struct dive_site *ds = (struct dive_site *)idx.model()->index(idx.row(), LocationInformationModel::DIVESITE).data().value<void *>();
```
dirkhh: same here. This may be a sign of my limitations, but that's REALLY hard to parse for me.

desktop-widgets/modeldelegates.cpp
```diff
@@ -452,9 +453,8 @@ void LocationFilterDelegate::paint(QPainter *painter, const QStyleOptionViewItem
 	QString diveSiteName = index.data().toString();
 	QString bottomText;
 	QIcon icon = index.data(Qt::DecorationRole).value<QIcon>();
-	struct dive_site *ds = get_dive_site_by_uuid(
-		index.model()->data(index.model()->index(index.row(),0)).toInt()
-	);
+	struct dive_site *ds = (struct dive_site *)
+		index.model()->data(index.model()->index(index.row(), LocationInformationModel::DIVESITE)).value<void *>();
```
dirkhh: ditto

qt-models/divelocationmodel.cpp
```diff
@@ -118,24 +119,23 @@ GeoReferencingOptionsModel::GeoReferencingOptionsModel(QObject *parent) : QStrin
 
 bool GPSLocationInformationModel::filterAcceptsRow(int sourceRow, const QModelIndex &parent) const
 {
-	uint32_t uuid = sourceModel()->index(sourceRow, LocationInformationModel::UUID, parent).data().toUInt();
-	if (uuid == ignoreUuid || uuid == RECENTLY_ADDED_DIVESITE)
+	struct dive_site *ds = (struct dive_site *)sourceModel()->index(sourceRow, LocationInformationModel::DIVESITE, parent).data().value<void *>();
```
dirkhh: yeah, it's that same construct I keep stumbling over. Maybe it will start making sense to me. I there a way to make this two steps that makes it clearer?

core/subsurface-qt/DiveObjectHelper.cpp
```diff
@@ -129,7 +129,7 @@ QString DiveObjectHelper::gps_decimal() const
 
 QVariant DiveObjectHelper::dive_site() const
 {
-	return QVariant::fromValue((void *)get_dive_site_by_uuid(m_dive->dive_site_uuid));
+	return QVariant::fromValue((uintptr_t)get_dive_site_by_uuid(m_dive->dive_site_uuid));
```
dirkhh: I think I'm missing the significance of the `uintptr_t` here. Help?

core/parse-xml.c
```diff
@@ -558,6 +558,13 @@ static void hex_value(char *buffer, uint32_t *i)
 	*i = strtoul(buffer, NULL, 16);
 }
 
+static void dive_site(char *buffer, struct dive_site **ds)
+{
+	uint32_t uuid;
+	hex_value(buffer, &uuid);
+	*ds = get_dive_site_by_uuid(uuid);
```
dirkhh, line 565: this dereference doesn't need a NULL check?

dirkhh (approved):
First: wow. You don't shy away from tedious work.
I have a few comments and questions, but overall this looks like the right thing to do.
I think a couple of my comments are worth a refresh of the series, but the underlying changes look ok.

desktop-widgets/locationinformation.cpp
bstoeger: Yes, I hate it too. We already have a few of those in the code base, notably when getting pointers to dives and trips out of the models. Therefore I thought let's continue using this idiom, as ugly as it is.

What happens:
When you get data out of a model, it is returned in a `QVariant`, which is kind of lame in that it doesn't support arbitrary types out-of-the-box. So what we did is cast to (void *) transport that in a `QVariant` and then cast back. Indeed ugly as hell.

This example here is even worse, because we access the model from the index and the make an index out of the index via the model and .... never mind.

Note that I didn't come up with that just added the void * casting - I'll try to come up with something more palatable.

====

That being said, the whole situation becomes even more crazy when QML is involved, because there, all QVariant<void *> seem to compare as equal - that revelation was quite a WTF-moment. What I therefore did was cast to uintptr_t(!)

====

One way out of this mess would perhaps be registering the types as meta-types with Qt's preprocessor directives. I believe that you can then put them in QVariants, but so far I haven't explored that option, as it also strikes me as not very elegant.

Does that clear it up somewhat?

core/subsurface-qt/DiveObjectHelper.cpp
bstoeger: There's a comment in a later commit, which should probably be copied here. Short story: QVariant<void *> compare as equal in QML on my setup, whereas QVariant<uintptr_t> don't. I couldn't believe it either.

core/parse-xml.c
bstoeger: None of the output parameters in the parser are checked for NULL. Therefore I would think that for the sake of consistency we shouldn't add a NULL-check here either.

core/datatrak.c
bstoeger: Moreover note that the dereference is removed in a later commit anyway.

core/divesite.c
bstoeger: done (locally).

desktop-widgets/locationinformation.cpp
bstoeger: true - fixed (locally).

bstoeger:
Sigh. I just found a bug:
Moving the flag on the map produces:
```
qrc:/qml/MapWidget.qml:78: Error: Unknown method parameter type: const dive_site*
```

But I will only be able to analyze this tomorrow. Sigh.

neolit123:
`Q_DECLARE_METATYPE` might solve that.

bstoeger:
It seems to work now again.

By declaring `dive_site *` a metatype, all the crazy `uintptr_t` and `void *` constructs are gone now. There's still the `index.model().index(index ...)` insanity, but that was there before.

Of course the history is now super-ugly with constructs being added and removed..

Ultimately, `struct dive *` and `struct dive_trip *` should probably also be made Q_METATYPEs so that more `void *`-casting can be removed

desktop-widgets/locationinformation.cpp
dirkhh: thanks for the explanations - every time I work with you on one of your PRs I learn something. Simply awesome.

dirkhh:
Nice.
I don't know why, but without the extra void * in the middle the code does seem much more logical to me.
I'm ok with making dive and dive_trip Q_METATYPES as well. But I think for now this series is ready to be merged and to get a lot more testing.

PR #1819, opened 2018-10-23
Show coordinates if no tags from reverse geolookup

mturkia:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This will display the dive site coordinates after the Location label if
there are no associated tags from reverse geolookup. Thus it'll be
clearer for users that we do have GPS location stored for the dive, and
might be able to see more easily if the recorded coordinates are
correct.

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -419,6 +419,12 @@ void MainTab::updateDiveInfo(bool clear)
 		if (ds) {
 			ui.location->setCurrentDiveSiteUuid(ds->uuid);
 			ui.locationTags->setText(constructLocationTags(&ds->taxonomy, true));
+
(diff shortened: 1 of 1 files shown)


dirkhh:
I think it's fair to say that Travis isn't a fan of this PR...

bstoeger:
28e3413ff66552f392fecee25068d634cdfe59fc replaced latitude / longitude pairs by location_t.

Note that it also introduced a `has_location()` helper, which might be of use for this PR.

mturkia:
should have pulled the latest master, 2 days old repo seems to be ancient in Subsurface. Force pushed a commit that should be ok. And uses the has_location()

dirkhh (approved):
I like it

PR #1822, opened 2018-10-24
Mobile filter2

dirkhh:

- [x] Bug fix
- [x] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Not sure if I should call this a bug fix or a new feature :-)
We now update the trip count on mobile to reflect filtering.

### Changes made:

1) two changes to core code: a new member in the dive trip structure to hold the number of dives shown (vs total nr of dives) in a trip
2) and the ability to get to the underlying dive structure from the DiveObjectHelper. This one I was worried about - one could see it as a hack. But then adding an accessor to get the dive_trip member seemed equally strange. And adding accessors to get to the showndives member of the trip was just ridiculous. I'm curious to hear what people think would be the right abstraction layer here
3) update the filter model to track how many dives of each trip are shown
### Related issues:

See comments in #1808 where this change is requested

### Release note:

That reminds me, I need to add an entry for the whole mobile filtering thingy

### Documentation change:

still to be done

### Mentions:

@janmulder @jbygdell 

(I'm on a plane, connected via WiFi, 11km or so above the ground... I don't think it's a good idea for me to try and upload Android binaries...)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Mobile: add full text filtering of the dive list
 - Desktop: don't add dive-buddy or dive-master when tabbing through fields
 - Filter: don't recalculate all filters on dive list-modifying operations
 - Desktop: don't recalculate full dive list on dive list-modifying operations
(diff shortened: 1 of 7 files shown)


jbygdell:
LGTM

PR #1834, opened 2018-10-29
Implement custom sort function

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Implement a custom sort function for more flexibility (and potential for optimization). I repeat the commit message of the major commit here:
```
The dive list was sorted using the default-sorter of
QSortFilterProxy model. This is mighty inflexible as it
considers only one column. This has the funky effect that
for rows with identical elements, the sort order depends
on the previous sorting.

Implement a lessThan() function in the MultiFilterSortModel,
which simply hands the sorting down to the actual model.
This might be considered a layering violation, but it makes
things so much easier.

Sadly, it seems like the column-to-be-sorted is transported
in the provided indices. Therefore, the comparison is chosen
using a switch for *every* comparison. It would seem much
more logical to set a function pointer once and use that.
Further investigations are necessary.

The new code uses the "strcasecmp" function instead of Qt's
string function. This might have effect on the collation
of non-ASCII UTF-8 strings. Let's see.
```
From a user's perspective, this provides stable sorting.
But note that the actual reason for this change is an attempt in splitting the `DiveTripModel` in two, viz. a Tree and a List mode.
 
RFC Material: This now uses `strcasecmp`. I am not sure if that is as "good" as Qt's string comparison / collation.

### Changes made:

1) Change control flow to match data flow: core --> DiveTripModel --> MultiFilterSortModel --> DiveListView.
2) Detangle code in DiveListView.
3) Implement custom comparison function.

### Release note:

Probaby should do.
### Documentation change:

No.

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -320,9 +320,6 @@ DiveTripModel::DiveTripModel(QObject *parent) :
 	connect(&diveListNotifier, &DiveListNotifier::divesSelected, this, &DiveTripModel::divesSelected);
 	connect(&diveListNotifier, &DiveListNotifier::divesDeselected, this, &DiveTripModel::divesDeselected);
 	connect(&diveListNotifier, &DiveListNotifier::currentDiveChanged, this, &DiveTripModel::currentDiveChanged);
-
-	// Update trip headers if filter finished
-	connect(MultiFilterSortModel::instance(), &MultiFilterSortModel::filterFinished, this, &DiveTripModel::filterFinished);
 }
 
 int DiveTripModel::columnCount(const QModelIndex&) const
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -53,9 +53,6 @@ QVariant DiveTripModel::tripData(const dive_trip *trip, int column, int role)
 	if (role == TRIP_ROLE)
 		return QVariant::fromValue<void *>((void *)trip); // Not nice: casting away a const
 
-	if (role == SORT_ROLE)
-		return (qulonglong)trip->when;
-
 	if (role == Qt::DisplayRole) {
 		switch (column) {
 		case DiveTripModel::NR:
@@ -143,46 +140,6 @@ QVariant DiveTripModel::diveData(const struct dive *d, int column, int role)
 	switch (role) {
 	case Qt::TextAlignmentRole:
 		return dive_table_alignment(column);
-	case SORT_ROLE:
-		switch (column) {
-		case NR:
-			return (qlonglong)d->when;
-		case DATE:
-			return (qlonglong)d->when;
-		case RATING:
-			return d->rating;
-		case DEPTH:
-			return d->maxdepth.mm;
-		case DURATION:
-			return d->duration.seconds;
-		case TEMPERATURE:
-			return d->watertemp.mkelvin;
-		case TOTALWEIGHT:
-			return total_weight(d);
-		case SUIT:
-			return QString(d->suit);
-		case CYLINDER:
-			return QString(d->cylinder[0].type.description);
-		case GAS:
-			return nitrox_sort_value(d);
-		case SAC:
-			return d->sac;
-		case OTU:
-			return d->otu;
-		case MAXCNS:
-			return d->maxcns;
-		case TAGS:
-			return get_taglist_string(d->tag_list);
-		case PHOTOS:
-			return countPhotos(d);
-		case COUNTRY:
-			return QString(get_dive_country(d));
-		case BUDDIES:
-			return QString(d->buddy);
-		case LOCATION:
-			return QString(get_dive_location(d));
-		}
-		break;
 	case Qt::DisplayRole:
 		switch (column) {
 		case NR:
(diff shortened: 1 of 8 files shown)


dirkhh (commented):
Several typos in the first commit message.
The switch in lessThan function does smell like something that could cause performance issues.
The strcasecmp vs QString comparison. Hmm. Not sure where that might bite us. Ideally we'd test with non ASCII languages. What does this do in Bulgarian @neolit123 ? Or in Russian?

bstoeger:
> The switch in lessThan function does smell like something that could cause performance issues.

Note that this reduces the number of switches per comparison from two to one. Before, each value to be compared would have been extracted with the data() function and SORT_ROLE.

But yes, maybe we should simply set a function-pointer when we change sort order and ignore Qt's column / role shenanigans.

qt-models/divetripmodel.cpp
```diff
@@ -1106,3 +1060,84 @@ void DiveTripModel::filterFinished()
...
+	return diff1 < 0 || (diff1 == 0 && diff2 < 0);
+}
+
+static int strCmp(const char *s1, const char *s2)
+{
+	if (!s1)
+		return !s2 ? 0 : -1;
+	if (!s2)
+		return 1;
+	return strcasecmp(s1, s2); // TODO: support for non-ASCII UTF-8!
```
neolit123: yep, strcasecmp cannot be used for UTF-8. QString already has this support.

neolit123:
i think this is a good change, as we need the flexibility of a custom sorter.

qt-models/divetripmodel.cpp
bstoeger: Ok, so let's go via `QString` at first and then see if we can avoid the unnecessary copying by using `std::collate` or something. But of course this makes only sense with benchmarking.

Today I will not find time to work on this, hopefully tomorrow.

bstoeger:
Benchmarks are in:
Using a ~4000 dives file and the following code:
```
QTime timer;
timer.start();
                sortByColumn(i, currentOrder);
int milli = timer.elapsed();
printf("Milliseconds: %d\n", milli);
```
I repeatedly clicked the "date" header. The reason to do it this way is to not test any string-comparisons.
I get for master:
```
Milliseconds: 7
Milliseconds: 8
Milliseconds: 7
Milliseconds: 7
Milliseconds: 8
Milliseconds: 8
Milliseconds: 8
Milliseconds: 8
Milliseconds: 9
Milliseconds: 8
Milliseconds: 8
Milliseconds: 6
Milliseconds: 9
Milliseconds: 7
Milliseconds: 7
Milliseconds: 7
Milliseconds: 8
Milliseconds: 9
Milliseconds: 8
Milliseconds: 7
Milliseconds: 8
```
and for this PR:
```
Milliseconds: 3
Milliseconds: 4
Milliseconds: 3
Milliseconds: 4
Milliseconds: 4
Milliseconds: 3
Milliseconds: 4
Milliseconds: 4
Milliseconds: 4
Milliseconds: 3
Milliseconds: 3
Milliseconds: 4
Milliseconds: 4
Milliseconds: 3
Milliseconds: 4
Milliseconds: 4
Milliseconds: 4
Milliseconds: 3
Milliseconds: 4
Milliseconds: 4
```

So ... it's twice as fast and yet totally irrelevant because below a Human's detection threshold. 😀 Computers simply have gotten too fast to care about code optimization.

PS: Obviously don't commit yet, as this still need's proper string-comparison.

dirkhh:
The commit message of https://github.com/Subsurface-divelog/subsurface/pull/1834/commits/9a9c7e046d6def9be61125e967d1e2e1f31cc624 should be updated. You no longer use strcasecmp :-)

dirkhh:
Also, the commit message of https://github.com/Subsurface-divelog/subsurface/pull/1834/commits/09d7490fbbb716e0770dd9ba266e8b20523852ab might benefit from some gentle editing:

> on the other hand,differs

Would prefer a space before differs

> In out particular case

should be "In our particular case"

bstoeger:
Reworded the commit-messages - thanks for pointing these out.

neolit123 (approved):
LGTM, thanks for the locale fix.
unicode comparisons sound like pure magic to me, but let's trust what Qt has.

dirkhh (approved):
Thanks

janmulder:
There is one aspect I do not like: the default ordering. The application opens with the oldest dive on top, and most divers will be interested in the newest dive. Also when clicking the date column, the oldest is shown first.

dirkhh:
I completely agree. We always want the latest dive on top by default - unless the user explicitly reverts the order.

bstoeger:
Or actually - it does make sense:
We start with
```
currentOrder(Qt::DescendingOrder)
```
but the model itself is filled in descending order - so in total you get ascending order. :P

So perhaps this is the correct fix for now:
```
diff --git a/desktop-widgets/divelistview.cpp b/desktop-widgets/divelistview.cpp
index b669eb077..cf47faa47 100644
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -27,7 +27,7 @@
 #include "core/subsurface-qt/DiveListNotifier.h"
 
 DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelection(false), sortColumn(DiveTripModel::NR),
-	currentOrder(Qt::DescendingOrder), dontEmitDiveChangedSignal(false), selectionSaved(false),
+	currentOrder(Qt::AscendingOrder), dontEmitDiveChangedSignal(false), selectionSaved(false),
 	initialColumnWidths(DiveTripModel::COLUMNS, 50)	// Set up with default length 50
 {
 	setItemDelegate(new DiveListDelegate(this));
@@ -465,11 +465,9 @@ void DiveListView::headerClicked(int i)
 	/* No layout change? Just re-sort, and scroll to first selection, making sure all selections are expanded */
 	if (currentLayout == newLayout) {
 		// If this is the same column as before, change sort order. Otherwise, choose a default
-		// sort order (descending for NR and DATE, ascending elsewise).
+		// sort order (ascending).
 		if (sortColumn == i)
 			currentOrder = (currentOrder == Qt::DescendingOrder) ? Qt::AscendingOrder : Qt::DescendingOrder;
-		else if (i == DiveTripModel::NR || i == DiveTripModel::DATE)
-			currentOrder = Qt::DescendingOrder;
 		else
 			currentOrder = Qt::AscendingOrder;
 		sortByColumn(i, currentOrder);
```

PR #1835, opened 2018-10-29
Fixed get_gas_at_time for equal times

atdotde:
This fixes a subtle bug introduced in 5c4569247a31c which
unified two functions finding the gasmix at a given time
during the dive. There was a slight difference, though:
Does a gaschange exactly at that time count or not? For
the planner to work, the answer has to be in the affirmative.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a bug reported by @willemferguson by mail in planned dives that do gas changes 
in the manually entered parts.
### Additional information:

To test, open dives/test51.xml and edit it in the planner. If it turns red for a ceiling violation, the bug is present.

### Mentions:
@willemferguson this should fix your bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fix a bug in planner for dives with manual gas changes.
 - Mobile: add full text filtering of the dive list
 - Desktop: don't add dive-buddy or dive-master when tabbing through fields
 - Filter: don't recalculate all filters on dive list-modifying operations
--- a/core/dive.c
+++ b/core/dive.c
@@ -4447,7 +4447,7 @@ struct gasmix get_gasmix(const struct dive *dive, const struct divecomputer *dc,
 		res = gasmix;
 	}
 
-	while (ev && ev->time.seconds < time) {
+	while (ev && ev->time.seconds <= time) {
 		res = get_gasmix_from_event(dive, ev);
 		ev = get_next_event(ev->next, "gaschange");
 	}
--- a/core/dive.c
+++ b/core/dive.c
@@ -4433,6 +4433,7 @@ int dive_has_gps_location(const struct dive *dive)
 	return dive_site_has_gps_location(dive->dive_site);
 }
 
+/* When evaluated at the time of a gasswitch, this returns the new gas */
 struct gasmix get_gasmix(const struct dive *dive, const struct divecomputer *dc, int time, const struct event **evp, struct gasmix gasmix)
 {
 	const struct event *ev = *evp;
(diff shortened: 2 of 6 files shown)


dirkhh:
I see you added a new test dive. Is there a test to go along with that?
And thanks for bisecting / fixing.

bstoeger (approved):
If it fixes a bug then go for it.

From a logical point, I guess it makes sense, but I'm not sure. 😛 Perhaps this deserves a comment?

bstoeger:
Or wait a bit - The big question is: does some part of the code depend on this returning the gasmix *at* the given time (I.e. before the events).

Perhaps the more secure version right now is simply adding 1 to `time.seconds` in the call of `get_gasmix`. But then this needs a comment on the different semantics of `get_gas_mix` and `get_gasmix_at_time`.

atdotde:
@dirkhh yesterday, it was too late for a test, but here it is.

@bstoeger given the granularity of our time variables, adding one would have also done the job, but after looking thru the use cases of get_gasmix and get_gasmix_at_time I think (with commit d3f3ac8) all uses are fine with the now fixed semantics of these functions. For short intervals, the difference is minimal, anyway, only the planner has longer segments (I hope).

That said, this is another instance of confusion arising from mixing time segments and instances of time (possibly being identified with the segment of time before or after). This has bitten us a number of times in the planner with segments vs. waypoints and is ultimately responsible for the terribly difficult logic in planner notes.c

But as so many times before, I now think I fixed it. At least, I added comments to clarify this issue and there is a test for the planner.

But while we are at it: Somebody should have a look at the OTU calculation. For a long time, I have been under the impression that it gives very different results than other programs computing it...

willemferguson:
On 2018/10/30 13:28, Robert C. Helling wrote:
>
> But while we are at it: Somebody should have a look at the OTU 
> calculation. For a long time, I have been under the impression that it 
> gives very different results than other programs computing it...
>
I can confirm this. I hate to use Multideco as a reference, but if it 
might be helpful I can prepare a few plans in Multideco, showing OTU & 
CNS, and then do exactly the same for Subsurface. This does not mean 
Mutideco does it correctly, but at least one could compare the 2 pieces 
of software.

I have not used other logging software for years (Subsurface satisfies 
all my dive modes and equipment), but one could do similar comparisons 
between Subsurface and, let's say, Divinglog 5. What do you think?

However, I am diving till Sunday, so I could not get to that before then.

Kind regards,

willem



-- 
This message and attachments are subject to a disclaimer.

Please refer to 
http://upnet.up.ac.za/services/it/documentation/docs/004167.pdf 
<http://upnet.up.ac.za/services/it/documentation/docs/004167.pdf> for
full 
details.

atdotde:
Willem, that would be great. Ideally, these are very simple plans where one can do the calculation by hand to see at which part we go wrong.

bstoeger:
Thanks for debugging this - Travis seems to be happy about the new test, therefore I think that we can commit.

PR #1840, opened 2018-11-02
Mobile copy

mturkia:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Implement copy-paste for Subsurface-mobile. With a long press on divelist, you get the option to copy some dive details and then paste them to another dive. Currently the details to be copied are hardcoded, but the plan is to implement configuration possibility for this (seems to take quite a long time for me to do this, so decided to get the pull request done for the basic implementation).

### Changes made:

copy-paste support for mobile version

### Additional information:

Following info is copied, if exist:
	what.divemaster = true;
	what.buddy = true;
	what.suit = true;
	what.tags = true;
	what.cylinders = true;
	what.weights = true;

### Release note:

TODO

### Documentation change:

TODO
Now there are two other icons next to the dive delete icon when pressing long on divelist (mobile). And data can be copied to other dives.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Desktop: Indicate sort direction in list header
 - Desktop: Use defined sort-order when switching sort column
 - Fix a bug in planner for dives with manual gas changes.
+- Mobile: add initial copy-paste support
 - Mobile: add full text filtering of the dive list
 - Desktop: don't add dive-buddy or dive-master when tabbing through fields
 - Filter: don't recalculate all filters on dive list-modifying operations
--- a/Documentation/mobile-manual.txt
+++ b/Documentation/mobile-manual.txt
@@ -233,6 +233,27 @@ instruction to delete the dive.
 
 image::mobile-images/RedDustbin.jpg["FIGURE: delete dive from list",align="center"]
 
+== Copying dive details
+
+To copy selected dive details over to another dive, long-press a dive in the
+divelist. This will show a three icons on the right side of the selected dive
+as shown on the image below. The left image will copy details from the selected
+dive and middle button will paste what has been copied before. Naturally you
+should long-press a different dive after copying the details to perform the
+paste operation.
+
+image::mobile-images/CopyPaste.jpg["FIGURE: copy dive details in list",align="center"]
+
+Currently the details to be copied are hard-coded and include the following
+information:
+
+- divemaster
+- buddy
+- suit
+- tags
+- cylinders
+- weights
+
 [[S_Download]]
 == Download dives from a dive computer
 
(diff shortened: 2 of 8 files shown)


dirkhh (approved):
This looks great. Nicely done, @mturkia 
On the one hand I'd love to have this in the upcoming release, on the other hand I worry how much this might break (plus documentation need)... thoughts?

dirkhh:
If nothing else it DOES need a CHANGELOG entry. But I really think this needs a couple of lines in the mobile user manual as well.

mturkia:
Changelog and mobile-manual updated.

I am quite clueless when it comes to the mobile version, so I have no idea whether this might break something or not. For sure, the UI should be polished, but in my testing it is working.

Future work should include configuration option for what is copied over, I have been thinking that long press on copy button would open up the configuration options to avoid the settings pane to overflow even more. And possibly visual indication when we have something in paste buffer.

Anyway, I would like to have this feature in next release, but will of course understand if it does not make it. Also someone else should test it out to see if something is borken.

janmulder:
Tried it after a dive today. What I'm missing is visual confirmation that anything has happened. As you are doing this from the dive list, and all the pasted data is in the details only.

mturkia:
> Tried it after a dive today. What I'm missing is visual confirmation that anything has happened. As you are doing this from the dive list, and all the pasted data is in the details only.

Good point. Just added notification text for both copy and paste operations.

mturkia:
> On 4 Nov 2018, at 10.56, Jan Mulder <notifications@github.com> wrote:
> 
> I'm not a real fan of toast messages, but this does the job. So ACK.
> 
> Another thing. Reading the code, I see that not all dive_components are mentioned. Like notes, divesite are missing. Is this by intention or just because this is an initial new capability?
> 

This is intentional. I usually don’t want to copy that info :)
Configuration UI is really needed, but as long as this is hard-coded, IMO it does not make sense to copy i.e. divesite. (I rarely dive in same spot multiple times in a row)

janmulder:
> This is intentional. I usually don’t want to copy that info :) Configuration UI is really needed, but as long as this is hard-coded, IMO it does not make sense to copy i.e. divesite.

I agree in the current state (without config UI), but I see no reason to leave anything out (that is already supported on the desktop copy-paste) in the state with config UI. Copying notes seems useless to me in any case, but copying divesite might be useful for numerous users of the app. Obviously, the copy-paste can be used for any historic data (not only dives "in a row"), that is now easy to find with the filter capability.

janmulder:
> Just added notification text for both copy and paste operations.

I'm not a real fan of toast messages, but this does the job. So ACK.

Another thing. Reading the code, I see that not all `dive_components` are mentioned. Like notes, divesite are missing. Is this by intention or just because this is an initial new capability?

dirkhh:
The conflict is trivial to deal with.
But I wonder about @janmulder 's last comment. Should this go in as is?

janmulder:
I'm fine with merging as is. At the same time, I believe that a configuration option to set the fields one likes to copy/paste a required addition to this functionality, as users will find scenarios that we are not thinking of at this point.

mturkia:
I am working on ui for this. And am somewhat optimistic I might be able to pull it off. Not sure how long it will take.

dirkhh:
Ok, so I'll merge it as is as now is the best time to get new stuff in :-)

PR #1841, opened 2018-11-02
Dive list: invert default sort order for sort by date / number

bstoeger:
Commit 6dc1d239f81f16133d5954bd2b21ffa0c2dcc755 introduced a
well-defined sort order in the case of equal contents. It changed
the code for sorting by date to simply use the order of the
source model.

BUT: The source-model was already sorted in descending order
on date. Thus setting the default order on descening by date,
the data was then presented as *ascending* by date.

Change this back to descending by always using default-ascending
in the filter model.

Ultimately, the source model should simply reflect the ordering
of the core-data (ascending on date), but such a change is
too invasive shortly before release.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes Bug introduced in 6dc1d239f81f16133d5954bd2b21ffa0c2dcc755. See commit message.
### Changes made:

1) Change default sort order when sorting by date or number.

### Related issues:

Fixes bug introduced in #1834 

### Release note:

No (recently introduced).
### Documentation change:

No.
### Mentions:

@janmulder: please test / comment.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -27,7 +27,7 @@
 #include "core/subsurface-qt/DiveListNotifier.h"
 
 DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelection(false), sortColumn(DiveTripModel::NR),
-	currentOrder(Qt::DescendingOrder), dontEmitDiveChangedSignal(false), selectionSaved(false),
+	currentOrder(Qt::AscendingOrder), dontEmitDiveChangedSignal(false), selectionSaved(false),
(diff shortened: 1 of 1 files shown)


janmulder (commented):
ACK.

And yes, the code as it is, is confusing but that is indeed something for after the coming release.

dirkhh (approved):
I always hate the idea that we don't do "the right thing" because of an "upcoming release". With my goal of releases at least every other month, there's almost always a release that is "upcoming". But I agree that some changes are better done "right after" such a release to maximize the time for testing and fixing unintended consequences of such change.

PR #1842, opened 2018-11-03
Undo: fix addition of multiple dives

bstoeger:
The generic addInBatches() function is used to add batches of
contiguous sets of dives to the dive-list models. The loop
searching for the end of the batch used the wrong index and
would therefore not properly cut the batches.

Fix this.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a horrible bug in the qt-models code introduced with the undo work.
To reproduce the bug:
1) select non contiguous set of dives
2) delete dives
3) undo

The dives will be added as a contiguous set. Thus, the list will not have proper chronological order and all hell breaks loose!

The bug is a release-blocker, therefore I'm tagging `priority-high`.
### Changes made:

1) Use proper index for finding end of the range

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -753,9 +753,9 @@ void DiveTripModel::addDivesToTrip(int trip, const QVector<dive *> &dives)
 // before the trip in the case of equal timestamps.
 bool DiveTripModel::dive_before_entry(const dive *d, const Item &entry)
 {
-	// Dives at the same time come before trips, therefore use the "<=" operator.
+	// Dives at the same time come before trips, therefore use the ">=" operator.
 	if (entry.trip)
-		return d->when <= entry.trip->when;
+		return d->when >= entry.trip->when;
 	return !dive_less_than(d, entry.getDive());
 }
 
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -717,7 +717,7 @@ void addInBatches(Vector1 &v1, const Vector2 &v2, Comparator comp, Inserter inse
 			// We were at end -> insert the remaining items
 			j = v2.size();
 		} else {
-			for (j = i + 1; j < (int)v2.size() && comp(v2[i], v1[idx]); ++j)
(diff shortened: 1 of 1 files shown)


janmulder:
Tried this both before and after this fix. While I fully believe that this fix can be merged, I do not see "the hell breaking lose" without this fix. In fact, after this fix, I still see a strange behavior when deleting and undoing.

Using my trivial test.ssrf logbook from before (13 dives, 11 in one trip, 1 dive before and 1 dive after the trip). Delete the dives outside the trip, delete 1 dive in the trip). Undo. The dive in the trip shows up in the expected spot. The dives before/after are now both below the trip. Re-sort on the number => dives go both on top or on bottom of the trip. Re-sort on date => dives are back at their chronological position.

I did not bisect this scenario, so I do not know if this has always been this way, but this feels like a bug that is related to this PR.

bstoeger:
Yes - maybe that was a bit of a hyperbole as the user can reset the model by switching between tree and list view. In any case, a wrong order is something we want to avoid at all costs, because it will lead to wrong insertion points and a total tohubohu down the road.

What you describe sounds like the `when` field of the trip-structure is not updated properly. I will analyze that later today.

bstoeger:
Argh. It comes from the fact that the `DiveTripModel` is in *reverse* chronological order and in reverse order with respect to the core model. So my brain always gets confused with the comparison-direction. Pushed a fix on top of this PR.

bstoeger:
PS: In the short run (shortly after release), I'd really like to adapt the order of the model to mirror the data in the core. In the long run, I'd like to get rid of the redundant representation in the Qt-model altogether.

dirkhh:
Based on the discussion I'm a bit unclear whether this should be merged prior to release or not.

bstoeger:
Definitely, as it fixes two bugs.

dirkhh (approved):
Both changes seem correct

PR #1846, opened 2018-11-06
Dive list: invert sort-direction to reflect core

bstoeger:
Traditionally, the DiveTripModel has its data sorted in opposite
direction to the core-data (chronologically descending vs. ascending).
This bring a number of subtle problems. For example, when filling
the model, trips are filled according to the *last* dive, whereas
later insertion points are according to the ->when value from the
core, which depends on the *first* dive.

As a start of fixing these subtleties, change the sort direction
to reflect the core-data. Ideally, this should lead to a removal
of the redundant data-representation.

Since the model is now sorted in ascending order, sorting has to
be enabled in the DiveListView constructor to reflect the
default-descending order.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

On closer look, the `DiveTripModel` code has a number of problems concerning overlapping trips and dives that happened during a trip, but are not part of a trip. This PR does not fix all of these problems, but is a first step in making them fixable. It also is a first step in removing the redundant representation of list-mode.

I have difficulties giving this a risk / gain rating. It certainly needs testing, comparisons are infamously tricky.

### Changes made:

1) Reverse order in DiveTripModel to reflect core.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1712,8 +1712,12 @@ void clear_table(struct dive_table *table)
  * probably want to unify the models.
  * After editing a key used in this sort-function, the order of
  * the dives must be re-astablished.
- * Currently, this does a lexicographic sort on the (start-time, id)
- * tuple. "id" is a stable, strictly increasing unique number, that
+ * Currently, this does a lexicographic sort on the
+ * (start-time, trip-time, id) tuple.
+ * trip-time is defined such that dives that do not belong to
+ * a trip are sorted *after* dives that do. Thus, in the default
+ * chronologically-descending sort order, they are shown *before*.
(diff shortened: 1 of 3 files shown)


dirkhh:
I'll wait until after release - simply because that's such a weird part of the code with so many strange side effects.

bstoeger:
Makes sense. Just to be up-front about this: These subtleties didn't exist before, because the model was reset after every change. Of course keeping model and core incrementally in sync is a bit tricky. I really wish there was only one representation of that data. But I don't have yet a fleshed out plan in that direction.

PR #1850, opened 2018-11-10
More accurate CNS calculations

willemferguson:
Update table of maximum oxygen exposure durations, used in CNS calulations.
This table shows the official NOAA maximum O2 exposure limits
(in seconds) for different PO2 values. It also gives
slope values for linear interpolation for intermediate PO2 values
between the tabulated PO2 values in the 1st column.
Top & bottom rows are inserted that are not in the NOAA table:
(1) For PO2 > 1.6 the same slope value as between
1.5 & 1.6 is used. This exptrapolation for PO2 > 1.6 likely
gives an underestimate above 1.6 but is better than the
value for PO2=1.6 (45 min). (2) The NOAA table only
tabulates values for PO2 >= 0.6. Since O2-uptake occurs down to
PO2=0.5, the same slope is used as for 0.7 > PO2 > 0.6.
This gives a conservative estimate for 0.6 > PO2 > 0.5. To
preserve the integer structure of the table, all slopes are
given as slope*10: divide by 10 to get the valid slope.
The columns below are:
po2 (mbar), Maximum Single Exposure (seconds), single_slope,
Maximum 24 hour Exposure (seconds), 24h_slope */

Then update Calculations of the CNS for a single dive  -
this only takes the first divecomputer into account.
The previous version of the code did a table lookup and
used the max O2 exposure for the next-higher PO2 category.
This gave a shorter max O2 exposure time and a higher CNS
contribution for a specific dive segment, resulting in a
slightly conservative value of CNS, often some 2 - 3 % too high.
This code does an interpolation for PO2 values inbetween
PO2 entries in the lookup table and therefore results in a more
accurate maximum O2 exposure time for that PO2.
The maximum O2 exposure duration for each segment
is also calculated based on the mean depth of the two
samples (start & end) that define each segment. The CNS
contribution of each segment is found by dividing the
time duration of the segment by its maximum exposure duration.
The contributions of all segments of the dive are summed to
get the total CNS% value. This is a partial implementation
of the proposals in Erik Baker's document "Oxygen Toxicity Calculations" */

Overall, this PR does not radically alter the existing CNS calculation,
it only makes it more accurate and more consistent by doing
interpolation and by using mean segment depth to find PO2.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@atdotde @glance-

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -181,54 +181,84 @@ static int calculate_otu(const struct dive *dive)
 	}
 	return lrint(otu);
 }
-/* calculate CNS for a dive - this only takes the first divecomputer into account */
-int const cns_table[][3] = {
-	/* po2, Maximum Single Exposure, Maximum 24 hour Exposure */
-	{ 1600, 45 * 60, 150 * 60 },
-	{ 1500, 120 * 60, 180 * 60 },
-	{ 1400, 150 * 60, 180 * 60 },
-	{ 1300, 180 * 60, 210 * 60 },
-	{ 1200, 210 * 60, 240 * 60 },
-	{ 1100, 240 * 60, 270 * 60 },
-	{ 1000, 300 * 60, 300 * 60 },
-	{ 900, 360 * 60, 360 * 60 },
-	{ 800, 450 * 60, 450 * 60 },
-	{ 700, 570 * 60, 570 * 60 },
-	{ 600, 720 * 60, 720 * 60 }
+/* Table of maximum oxygen exposure durations, used in CNS calulations.
+   This table shows the official NOAA maximum O2 exposure limits (in seconds) for different PO2 values. It also gives
+   slope values for linear interpolation for intermediate PO2 values between the tabulated PO2 values in the 1st column.
+   Top & bottom rows are inserted that are not in the NOAA table: (1) For PO2 > 1.6 the same slope value as between
+   1.5 & 1.6 is used. This exptrapolation for PO2 > 1.6 likely gives an underestimate above 1.6 but is better than the
+   value for PO2=1.6 (45 min). (2) The NOAA table only tabulates values for PO2 >= 0.6. Since O2-uptake occurs down to
+   PO2=0.5, the same slope is used as for 0.7 > PO2 > 0.6. This gives a conservative estimate for 0.6 > PO2 > 0.5. To
+   preserve the integer structure of the table, all slopes are given as slope*10: divide by 10 to get the valid slope.
+   The columns below are: 
+   po2 (mbar), Maximum Single Exposure (seconds), single_slope, Maximum 24 hour Exposure (seconds), 24h_slope */
+int const cns_table[][5] = {
+	{ 1600,  45 * 60, 456, 150 * 60, 180 },
+	{ 1550,  83 * 60, 456, 165 * 60, 180 },
+	{ 1500, 120 * 60, 444, 180 * 60, 180 },
+	{ 1450, 135 * 60, 180, 180 * 60,  00 },
+	{ 1400, 150 * 60, 180, 180 * 60,  00 },
+	{ 1350, 165 * 60, 180, 195 * 60, 180 },
+	{ 1300, 180 * 60, 180, 210 * 60, 180 },
+	{ 1250, 195 * 60, 180, 225 * 60, 180 },
+	{ 1200, 210 * 60, 180, 240 * 60, 180 },
+	{ 1100, 240 * 60, 180, 270 * 60, 180 },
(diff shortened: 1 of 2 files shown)


dirkhh:
> @dirkhh: the Travis build error on Android is valid. Some recent commits in qt-android-cmake break our build.

Great. Something else to investigate. Thanks for tracking this down, @janmulder

dirkhh:
Robert, this is outside my comfort zone. Comments?

core/divelist.c
```diff
@@ -181,31 +181,48 @@ static int calculate_otu(const struct dive *dive)
...
+   of the proposals in Erik Baker's document "Oxygen Toxicity Calculations" */
 static double calculate_cns_dive(const struct dive *dive)
 {
 	int n;
 	size_t j;
 	const struct divecomputer *dc = &dive->dc;
 	double cns = 0.0;
-
-	/* Caclulate the CNS for each sample in this dive and sum them */
+	/* Calculate the CNS for each sample in this dive and sum them */
```
atdotde, line 228: Seems like whitespace only.

core/divelist.c
```diff
@@ -214,21 +231,20 @@ static double calculate_cns_dive(const struct dive *dive)
 		t = sample->time.seconds - psample->time.seconds;
 		if (sample->setpoint.mbar) {
 			po2 = sample->setpoint.mbar;
-		} else {
+		} else {	// Find FiO2 as well as PO2 at mean depth of dive segment:
 			int o2 = active_o2(dive, dc, sample->time);
-			po2 = lrint(o2 * depth_to_atm(sample->depth.mm, dive));
+			po2 = lrint(o2 * depth_to_atm((int)round((sample->depth.mm + psample->depth.mm) / 2.0), dive));
```
atdotde: why do you go to float? Integer should be fine.

core/divelist.c
```diff
@@ -214,21 +231,20 @@ static double calculate_cns_dive(const struct dive *dive)
...
+		/* Don't increase CNS when below 500 matm */
+		if (po2 <= 500)
 			continue;
-		/* Find what table-row we should calculate % for */
-		for (j = 1; j < sizeof(cns_table) / (sizeof(int) * 3); j++)
+		/* Find the table-row for calculating the maximum exposure at this PO2 */
+		for (j = 1; j < sizeof(cns_table) / (sizeof(int) * 5); j++)
 			if (po2 > cns_table[j][0])
 				break;
-		j--;
```
atdotde, line 228: Did you leave out this j-- on purpose? I don't know which direction your slope (to the left or to the right) is actually going to work.

core/divelist.c
```diff
@@ -214,21 +231,20 @@ static double calculate_cns_dive(const struct dive *dive)
...
 		}
-		/* CNS don't increse when below 500 matm */
-		if (po2 < 500)
+		/* Don't increase CNS when below 500 matm */
+		if (po2 <= 500)
 			continue;
-		/* Find what table-row we should calculate % for */
-		for (j = 1; j < sizeof(cns_table) / (sizeof(int) * 3); j++)
+		/* Find the table-row for calculating the maximum exposure at this PO2 */
+		for (j = 1; j < sizeof(cns_table) / (sizeof(int) * 5); j++)
```
atdotde: These indices could be slightly more readable if you typedef'ed an enum with the column names 

typedef enum {PO2, MAX_SINGLE_EXPOSURE, SINGLE_SLOPE, MAX_DAILY_EXPOSURE, MAX_DAILY_SLOPE, LAST} cns_headers;

then this 5 becomes LAST and you can use names when accessing the table elements.

atdotde (commented):
Only the j-- comment might actually be problematic (but maybe I don't get it), the rest is just coding style.

core/divelist.c
willemferguson: j-- dropped intentionally

willemferguson:
The updated code above hopefully addresses the issues raised. Thanks for your attention to this.

core/divelist.c
willemferguson: Done

core/divelist.c
willemferguson: Dropped open line

atdotde (approved):

dirkhh:
Just FYI, I'm _literally_ in the middle of doing the 4.8.4 release, so this will be merged after that goes out

bstoeger:
I would think this and #1851 should have a CHANGELOG.md entry? Also the multiple-divemaster thing?

PR #1857, opened 2018-11-16
New filter widget

tcanabrava:
Change the old filter widget with, with one that has a simpler code and it's easier to maintain.
The functionality also changed, now you can search by stars, dates, temperature, tags, people, etc.
all searches based on name accepts multiple entries separated by a comma.
ctrl + f to open, ctrl + f to close the dialog.


tcanabrava:
I forgot to sign off, if you guys can, review it anyway, I'll update this as soon as possible.

desktop-widgets/filterwidget2.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+	// TODO: unhide this when we discover how to search for equipment.
+	ui->equipment->hide();
+	ui->labelEquipment->hide();
+	ui->invertFilter->hide();
+
+	ui->to->setDate(data.to.date());
+
+	connect(ui->maxRating, &StarWidget::valueChanged,
+		[this]{ updateFilter(); }
```
bstoeger: Personally, I dislike this kind of gratuitous lambdas. What's wrong with a pointer-to-member function? On the other hand I don't care much.

desktop-widgets/filterwidget2.cpp
```diff
@@ -0,0 +1,8 @@
+#include "filterwidget2.h"
+
+FilterWidget2::FilterWidget2(QWidget* parent)
+: QWidget(parent)
+, ui(new Ui::FilterWidget2())
+{
+    ui->setupUi(this);
```
dirkhh: oh good, first line of code that isn't tool generated, and indentation is done with spaces instead of tabs...

desktop-widgets/filterwidget2.h
```diff
@@ -13,7 +13,8 @@ namespace Ui {
 
 class FilterWidget2 : public QWidget {
     Q_OBJECT
-    FilterWidget2(QWidget *parent);
+public:
+    explicit FilterWidget2(QWidget *parent = 0);
```
dirkhh: this really should be just squashed into the previous commit. After it's switched to tab indent, that is

desktop-widgets/mainwindow.h
```diff
@@ -88,6 +88,7 @@ class MainWindow : public QMainWindow {
 	ProfileWidget2 *graphics;
 	DivePlannerWidget *divePlannerWidget;
 	DiveListView *diveList;
+    FilterWidget2 *filterWidget2;
```
dirkhh: whitespace

desktop-widgets/filterwidget2.cpp
```diff
@@ -4,5 +4,5 @@ FilterWidget2::FilterWidget2(QWidget* parent)
 : QWidget(parent)
 , ui(new Ui::FilterWidget2())
 {
-    ui->setupUi(this);
+	ui->setupUi(this);
```
dirkhh, line 10: when you submit a new series, it would be really nice to just not go through this. Sure, I could clean those up when applying, but rebasing and squashing those commits should be really easy for you.

desktop-widgets/mainwindow.cpp
```diff
@@ -194,7 +194,7 @@ MainWindow::MainWindow() : QMainWindow(),
 	registerApplicationState("PlanDive", divePlannerWidget, profileContainer, divePlannerSettingsWidget, plannerDetails );
 	registerApplicationState("EditPlannedDive", divePlannerWidget, profileContainer, diveList, mapWidget );
 	registerApplicationState("EditDiveSite", diveSiteEdit, profileContainer, diveList, mapWidget);
-	registerApplicationState("FilterDive", mainTab, profileContainer, diveListView, filterWidget2);
+	registerApplicationState("FilterDive", mainTab, profileContainer, diveList, filterWidget2);
```
dirkhh, line 198: this, too, shouldn't be a separate commit and should just be done correctly in the first place. Otherwise the series fails when trying `git bisect` and hitting one of the earlier commits.

desktop-widgets/CMakeLists.txt
```diff
@@ -141,3 +141,4 @@ add_library(subsurface_generated_ui STATIC ${SUBSURFACE_UI_HDRS})
 target_link_libraries(subsurface_generated_ui ${QT_LIBRARIES})
 add_library(subsurface_interface STATIC ${SUBSURFACE_INTERFACE} ${SUBSURFACE_UI_SRCS})
 target_link_libraries(subsurface_interface ${QT_LIBRARIES} ${GRANTLEE_LIBRARIES} subsurface_desktop_preferences)
+set_target_properties(subsurface_interface PROPERTIES CXX_STANDARD 14)
```
dirkhh: ugh.
I'm not sure I want that.
What, specifically, can we do if we have that.
What does it break - e.g., what's the minimum gcc required (we build Windows with gcc 4.9 IIRC)

desktop-widgets/filterwidget2.cpp
```diff
@@ -1,8 +1,25 @@
...
+
+#include <QDoubleSpinBox>
 
 FilterWidget2::FilterWidget2(QWidget* parent)
 : QWidget(parent)
 , ui(new Ui::FilterWidget2())
 {
 	ui->setupUi(this);
+
+	connect(ui->maxRating, &StarWidget::valueChanged, [this]{ updateFilter();});
```
dirkhh: yuck that's ugly. Come on, @tcanabrava - you know I hate this stuff. Just use a function pointer.

desktop-widgets/filterwidget2.cpp
dirkhh: Yes. See my comment two commits earlier. NAK on this. And I think this is the only reason to add C++14? Heck no.

desktop-widgets/filterwidget2.cpp
```diff
@@ -68,23 +75,24 @@ FilterWidget2::FilterWidget2(QWidget* parent)
...
+	data.maxWaterTemp = ui->maxWaterTemp->value();
+	data.minAirTemp = ui->minAirTemp->value();
+	data.maxWaterTemp = ui->maxWaterTemp->value();
+	data.from = ui->from->dateTime();
+	data.to = ui->to->dateTime();
+	data.tags = ui->tags->text().split(",", QString::SkipEmptyParts);
+	data.people = ui->people->text().split(",", QString::SkipEmptyParts);
+	data.location = ui->location->text().split(",", QString::SkipEmptyParts);
+	data.equipment = ui->equipment->text().split(",", QString::SkipEmptyParts);
+	data.invertFilter = ui->invertFilter->isChecked();
```
dirkhh, line 88: again, this should be cleaned up. You introduce new code, then you introduce a new data structure, then you change the code you introduced 6 commits ago to use the new data structure. Let's not do that.

qt-models/filtermodels.h
```diff
@@ -16,16 +16,17 @@ struct dive_trip;
...
-	double maxAirTemp;
+	bool validFilter = false;
+	int minVisibility = 0;
+	int maxVisibility = 5;
+	int minRating = 0;
+	int maxRating = 5;
+	double minWaterTemp = 0;
+	double maxWaterTemp = 100;
+	double minAirTemp = 0;
+	double maxAirTemp = 100;
```
dirkhh, line 27: same with these initializations... just do them as you first introduce that code.

dirkhh: BTW: initialization in a struct definition... I didn't know that you can do that. Or is this a C++ only thing?

qt-models/filtermodels.cpp
```diff
@@ -39,6 +39,55 @@ void MultiFilterSortModel::setLayout(DiveTripModel::Layout layout)
 
 bool MultiFilterSortModel::showDive(const struct dive *d) const
 {
+	if (! m_filterData.validFilter) {
+		return true;
+	}
```
dirkhh: no curly braces

qt-models/filtermodels.cpp
```diff
@@ -39,6 +39,55 @@ void MultiFilterSortModel::setLayout(DiveTripModel::Layout layout)
...
+
+	bool shouldShow = true;
+	if (d->visibility < m_filterData.minVisibility || d->visibility > m_filterData.maxVisibility) {
+		shouldShow = false;
+	}
```
dirkhh: ditto. and more of the same below

qt-models/filtermodels.cpp
```diff
@@ -75,7 +75,7 @@ bool MultiFilterSortModel::showDive(const struct dive *d) const
 		bool found_tag = false;
 		for (const auto& filter_tag : m_filterData.tags) {
 			for (const auto& dive_tag : dive_tags) {
-				if (filter_tag == dive_tag) {
+				if (filter_tag.trimmed() == dive_tag.trimmed()) {
```
dirkhh: this code was introduced in the previous commit. Just fix the previous commit, please

qt-models/filtermodels.cpp
```diff
@@ -39,6 +39,55 @@ void MultiFilterSortModel::setLayout(DiveTripModel::Layout layout)
...
+	if (m_filterData.to.isValid() && d->when > m_filterData.to.toTime_t()) {
+		shouldShow = false;
+	}
+
+	// tags.
+	if (!m_filterData.tags.isEmpty()) {
+		auto dive_tags = get_taglist_string(d->tag_list).split(",");
+		bool found_tag = false;
+		for (const auto& filter_tag : m_filterData.tags) {
+			for (const auto& dive_tag : dive_tags) {
```
dirkhh: a) I dislike use of auto here - why not just use the correct type
b) we tie the `&` to the variable, no the type

qt-models/filtermodels.cpp
```diff
@@ -75,7 +75,7 @@ bool MultiFilterSortModel::showDive(const struct dive *d) const
 		bool found_tag = false;
 		for (const auto& filter_tag : m_filterData.tags) {
 			for (const auto& dive_tag : dive_tags) {
-				if (filter_tag.trimmed() == dive_tag.trimmed()) {
+				if (dive_tag.trimmed().toUpper().contains(filter_tag.trimmed().toUpper())) {
```
dirkhh: and again, you introduced this two commits ago.
your commit series is more a "train of thought" exercise :-)

qt-models/filtermodels.cpp
```diff
@@ -104,6 +104,30 @@ bool MultiFilterSortModel::showDive(const struct dive *d) const
...
+		if (d->divetrip) {
+			diveLocations.push_back(QString(d->divetrip->location));
+		}
+		if (d->dive_site) {
+			diveLocations.push_back(QString(d->dive_site->name));
+		}
+
+		bool found_location = false;
+		for (const auto& filter_location : m_filterData.location) {
+			for (const auto& dive_location : diveLocations) {
```
dirkhh: see earlier comment

qt-models/filtermodels.cpp
```diff
@@ -158,15 +158,6 @@ bool MultiFilterSortModel::showDive(const struct dive *d) const
...
 	return shouldShow;
-
-	if (curr_dive_site) {
-		dive_site *ds = d->dive_site;
-		if (!ds)
-			return false;
-		return ds == curr_dive_site || same_string(ds->name, curr_dive_site->name);
-	}
-
-	return true;
```
dirkhh: why was this ever introduced?

qt-models/filtermodels.cpp
```diff
@@ -39,6 +39,55 @@ void MultiFilterSortModel::setLayout(DiveTripModel::Layout layout)
...
+			}
+		}
+
+		if (!found_tag) {
+			shouldShow = false;
+		}
+	}
+
+	return shouldShow;
+
```
dirkhh, line 137: this is an unconditional return - so the code below is dead code, correct?
ahh, ten commits later (or so) you remove the dead code.

dirkhh (changes requested):
So far I have just read the code, no attempt to compile / run it.
Lots of small things, see all the comments. Also, 679d46e98224544dcd74e37683cc97ae0b038544 introduces libdivecomputer changes. Err, please don't do that.

dirkhh:
The only desktop build that completes is the Qt57 / Fedora 26 one. Odd. I wonder why that one passes if all the others fail.

qt-models/filtermodels.cpp
```diff
@@ -15,605 +15,150 @@
...
 
-void BuddyFilterModel::diveAdded(const dive *d)
-{
-	QStringList buddies = getDiveBuddies(d);
-	if (buddies.empty()) {
-		increaseCount(QString());
-		return;
+	bool shouldShow = true;
+	if (d->visibility < m_filterData.minVisibility || d->visibility > m_filterData.maxVisibility) {
+		shouldShow = false;
```
bstoeger: I really dislike this kind of control flow. Why not just do a `return false` here? Shorter and much easier to reason about. This is idiomatic C, because in C you often have to do some cleanups, but in C++ calling the destructors is left to the compiler.

desktop-widgets/filterwidget2.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+	connect(ui->maxVisibility, &StarWidget::valueChanged,
+		[this]{ updateFilter(); }
+	);
+
+	connect(ui->minVisibility, &StarWidget::valueChanged,
+		[this]{ updateFilter();}
+	);
+
+	connect(ui->maxAirTemp, qOverload<double>(&QDoubleSpinBox::valueChanged),
```
bstoeger: According to the docs, `qOverload` is >= Qt 5.7. http://doc.qt.io/qt-5/qtglobal.html#qOverload

This one might be a legitimate use of lambdas:
```[this](double){ updateFilter(); }```
with a big fat comment that Qt is trying to be as painful as possible by overloading the `valueChanged` signal with double and string parameters. [What the hell!?]

I would welcome upping the requirement to 5.7+, but this doesn't seem like a particular convincing use case. ;)

File: desktop-widgets/mainwindow.h (hunk 4 above)
Comment (bstoeger) [reply desktop-widgets/mainwindow.h]: Is this actually used outside of `MainWindow`'s constructor? If not, there seems to be no point to introduce another member-variable.

File: qt-models/filtermodels.h [hunk 21]
```diff
@@ -137,14 +50,14 @@ public
...
-
+	void filterDataChanged(const FilterData& data);
 signals:
 	void filterFinished();
 private:
 	MultiFilterSortModel(QObject *parent = 0);
-	QList<FilterModelBase *> models;
 	struct dive_site *curr_dive_site;
 	DiveTripModel *model;
+	FilterData m_filterData;
```
bstoeger: None of the other member variables have the `m_` prefix. Even though we are super-inconsistent with respect to this, at least stay consistent inside of a class ... ?

tcanabrava:
Hey, just read all the conversation and I’ll fix all the issues raised. Thanks for the feedback, it’s valuable.

dirkhh:
Got it to compile.
```
diff --git a/desktop-widgets/printdialog.cpp b/desktop-widgets/printdialog.cpp
index bbbade2e4..4c863fbea 100644
--- a/desktop-widgets/printdialog.cpp
+++ b/desktop-widgets/printdialog.cpp
@@ -10,6 +10,7 @@
 #include <QShortcut>
 #include <QSettings>
 #include <QMessageBox>
+#include <QDialogButtonBox>
 
 #define SETTINGS_GROUP "PrintDialog"
 
diff --git a/desktop-widgets/simplewidgets.h b/desktop-widgets/simplewidgets.h
index bd1b8a2dd..a814519c4 100644
--- a/desktop-widgets/simplewidgets.h
+++ b/desktop-widgets/simplewidgets.h
@@ -20,7 +20,6 @@ class FilterModelBase;
 #include "ui_urldialog.h"
 #include "ui_divecomponentselection.h"
 #include "ui_listfilter.h"
-#include "ui_filterwidget.h"
 #include "core/exif.h"
 #include "core/dive.h"
 
```

I like the concept a lot. I found a couple of small issues. E.g., 
- how do I reset the lower bound of the star filter to "no star"?
- the temperature ranges feel strange with the spin boxes set to 0.00 and 99.99 - that's both ugly and misleading (while unlikely, at least in degF you could have a water temp above 100... it would make a lot more sense to populate those with the highest and lowest temp found in the dive list
- it would be nice to somewhere show the total number of dives that match

But overall... NICE.

tcanabrava:
"how do I reset the lower bound of the star filter to "no star"?"
clicking in the first star twice should clear it  (if it's not doing this is a error in my part, in the stars widget implementation.) 

The temperature idea is good, I'll try to implement it.

The total amount of dives is displayed in a per-trip basis.

qt-models/filtermodels.cpp
tcanabrava: That's a true statement.

desktop-widgets/filterwidget2.cpp
tcanabrava: I would say that this is not gratuitous, to connect with a pointer to member I need to pass a method that receives the same arguments as the valueChanged, so I'd need to have a DateTimeSomethingChanged, StringSomethingChanged, DoubleSomethingChanged and IntSomethingChanged. while with lambdas I can basically ignore the type that's being send. I'm thinking on how to do this.

tcanabrava: it seems that I was wrong and I can now connect with different parameters, nice.

bstoeger: Indeed I was confused about your explanation, because note how the lambda you defined above doesn't take a parameter either.

desktop-widgets/mainwindow.h
dirkhh: hey, still whitespace... and what about @bstoeger 's comment?

desktop-widgets/mainwindow.cpp
```diff
@@ -1876,13 +1875,7 @@ void MainWindow::on_paste_triggered()
...
 void MainWindow::on_actionFilterTags_triggered()
 {
-	if (ui.multiFilter->isVisible()) {
-		ui.multiFilter->closeFilter();
-		ui.actionFilterTags->setChecked(false);
-	} else {
-		ui.multiFilter->setVisible(true);
-		ui.actionFilterTags->setChecked(true);
-	}
+	setApplicationState(getCurrentAppState() == "FilterDive" ? "Default" : "FilterDive");
```
dirkhh, line 1873: not to quibble, but the text of the commit message (sorry, can't comment on commit messages, so commenting in the code) of "dropping tons of code" seems like a bit of an overstatement?

qt-models/filtermodels.h
dirkhh: yeah, let's call this `filterData`

qt-models/filtermodels.cpp
dirkhh: still no curly braces

dirkhh (changes requested):
Much better. Nice work, @tcanabrava 
As I noted in this review, there are a few review comments you didn't address (but you addressed most of them). And overall the commit messages aren't great in guiding someone who reads the series.
Given the submodule hiccup you'll have to do another revision, anyway, so I'd appreciate if you could address those remaining concerns and maybe work on commit messages. I'm happy to help with those as well.

desktop-widgets/filterwidget.ui
```diff
@@ -70,10 +70,10 @@
      <item>
       <widget class="QToolButton" name="clear">
        <property name="toolTip">
-	<string>Reset filters</string>
+        <string>Reset filters</string>
        </property>
        <property name="icon">
-        <iconset resource="../subsurface.qrc">
+        <iconset>
```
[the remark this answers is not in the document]
tcanabrava: qt-designer mishaps.

tcanabrava:
please don't let this rot. :)

dirkhh (approved):
@tcanabrava thanks for your patience and for your willingness to continue to improve this PR. I really like where we ended up. Well done.

dirkhh:
And of course, who doesn't love a series with these stats:
+559 −1,040
:-)
Now let's look at Travis :-(

dirkhh:
Ah, those are the two issues fixed in master. I'll merge as is.

bstoeger:
Something went wrong with the last version - the `filterDataChanged` signal and slot are not `connect()`ed.

Also dive-site editing breaks (doesn't show the dives belonging to the selected dive site). Not a big deal, but mildly annoying since I expressly mentioned that in a comment.

Personally, I miss the full-text search features of the old filter-widget (you could get a list of items by fulltext search), but
1) My log isn't even long enough to make the filtering necessary.
2) I'm "special" when it comes to UI. I happen to like bells and whistles. MacOS, which many consider the pinnacle of UI design, drives me insane. :-P

bstoeger:
PS: Not to give a wrong impression - overall I think this is an improvement.

bstoeger:
Also note that now all the `count_dives_with_*` functions can be removed. Is it planned to have this feature (how many dives did I have with this and that buddy, suit, etc.) resurrected somewhere?

dirkhh:
@tcanabrava something broke here when I merged it - I think you said you were going to fix this, any chance you could take a look? Thanks

bstoeger:
If @tcanabrava doesn't find time, I can try to look at the issues but I'd  like to avoid stepping on his toes. Especially concerning the signal thing. Also note that I can't reproduce one of @janmulder's reported problems.

janmulder:
@bstoeger: what problem are you referring to (that you cannot reproduce)?

bstoeger:
In #1889:

> 3) Tried to filter on a tag. Ok, something is filtered, but the end result is a far to small set of dives. I have at least 400 dives with that tag, and only about 20 are selected by the entered tag.
> 4) Now empty the entered tag. Expected the full divelist to be shown again, but only a few extra dives are added again.

PR #1858, opened 2018-11-17
More dive-list churn

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is more churn that comes out of the divelist / undo work. The dive-list is not yet fully under control, but I'm trying to get there.

Firstly, there's a detangling of ownership concerning trips: The core would merge trips with the same date. But that is problematic with upcoming changes to the import code: the importer is responsible for deciding when it wants to merge trips. Also it seems ill defined, as the date of a trip is a fluctuating value. Adding / removing / editing trips changes trip date. Sadly, the git-storage uses the trip date as an id. I'll try to see what happens with two trips on the same date [A situation that can already be constructed in the old code].

Secondly, this tries to move the comparison functions into a single place (the core) to get this mess under control. Not completely there yet, but hopefully a step in the right direction.

Thirdly, this converges core and model data structures: Dives in trips are now also kept in table form in the core. The long-term goal here is to remove the double representation of the data. Ideally, the models simply present the core-data. But that's still a long way to go.

That's all a bit hairy therefore shortly after release is probably the best time to get this a bit of exposure. I'm sure that there are a number of problems lurking, but hopefully nothing fundamental.

### Release note:

Nothing fundamentally user visible (unless the user has strange use-cases like overlapping trips, etc). So probably not.

### Documentation change:

No.
### Mentions:

The usual suspects. :)

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -40,6 +40,7 @@
  * void remove_autogen_trips()
  * bool dive_less_than(const struct dive *a, const struct dive *b)
  * bool trip_less_than(const struct dive_trip *a, const struct dive_trip *b)
+ * bool dive_or_trip_less_than(struct dive_or_trip a, struct dive_or_trip b)
  * void sort_table(struct dive_table *table)
  * bool is_trip_before_after(const struct dive *dive, bool before)
  * void delete_dive_from_table(struct dive_table *table, int idx)
@@ -1559,7 +1560,7 @@ void process_imported_dives(struct dive_table *import_table, bool prefer_importe
 		struct dive *dive_to_add = import_table->dives[i];
 
 		/* Find insertion point. */
-		while (j < dive_table.nr && dive_table.dives[j]->when < dive_to_add->when)
+		while (j < dive_table.nr && dive_less_than(dive_table.dives[j], dive_to_add))
 			j++;
 
 		/* Try to merge into previous dive. */
@@ -1742,13 +1743,11 @@ bool dive_less_than(const struct dive *a, const struct dive *b)
 	return comp_dives(a, b) < 0;
 }
 
-/* Trips are compared according to the first dive in the trip.
- * Even though it shouldn't happen, take care about "empty" trips.
- * Since a dive can only belong to one trip, no two trips should
- * compare as equal
- */
+/* Trips are compared according to the first dive in the trip. */
 static int comp_trips(const struct dive_trip *a, const struct dive_trip *b)
 {
+	/* This should never happen, nevertheless don't crash on trips
+	 * with no (or worse a negative number of) dives. */
 	if (a->dives.nr <= 0)
 		return b->dives.nr <= 0 ? 0 : -1;
 	if (b->dives.nr <= 0)
@@ -1759,6 +1758,33 @@ static int comp_trips(const struct dive_trip *a, const struct dive_trip *b)
 bool trip_less_than(const struct dive_trip *a, const struct dive_trip *b)
 {
 	return comp_trips(a, b) < 0;
+}
+
+/* When comparing a dive to a trip, use the first dive of the trip. */
+static int comp_dive_to_trip(struct dive *a, struct dive_trip *b)
+{
+	/* This should never happen, nevertheless don't crash on trips
+	 * with no (or worse a negative number of) dives. */
+	if (b->dives.nr <= 0)
+		return -1;
+	return comp_dives(a, b->dives.dives[0]);
+}
+
+static int comp_dive_or_trip(struct dive_or_trip a, struct dive_or_trip b)
+{
+	if (a.dive && b.dive)
+		return comp_dives(a.dive, b.dive);
+	if (a.trip && b.trip)
+		return comp_trips(a.trip, b.trip);
+	if (a.dive)
+		return comp_dive_to_trip(a.dive, b.trip);
+	else
+		return -comp_dive_to_trip(b.dive, a.trip);
+}
+
+bool dive_or_trip_less_than(struct dive_or_trip a, struct dive_or_trip b)
+{
+	return comp_dive_or_trip(a, b) < 0;
 }
 
 static int sortfn(const void *_a, const void *_b)
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -30,6 +30,7 @@
  * dive_trip_t *combine_trips_create(struct dive_trip *trip_a, struct dive_trip *trip_b)
  * struct dive *unregister_dive(int idx)
  * void delete_single_dive(int idx)
+ * void add_dive_to_table(struct dive_table *table, int idx, struct dive *dive)
  * void add_single_dive(int idx, struct dive *dive)
  * struct dive *merge_two_dives(struct dive *a, struct dive *b)
  * void select_dive(struct dive *dive)
@@ -1111,34 +1112,42 @@ struct dive **grow_dive_table(struct dive_table *table)
 }
 
 /* get the index where we want to insert the dive so that everything stays
- * ordered reverse-chronologically */
-int dive_get_insertion_index(struct dive *dive)
+ * ordered according to dive_less_than() */
+int dive_get_insertion_index(struct dive_table *table, struct dive *dive)
 {
 	/* we might want to use binary search here */
-	for (int i = 0; i < dive_table.nr; i++) {
-		if (dive->when <= dive_table.dives[i]->when)
+	for (int i = 0; i < table->nr; i++) {
+		if (dive_less_than(dive, table->dives[i]))
 			return i;
 	}
-	return dive_table.nr;
+	return table->nr;
 }
 
-/* add a dive at the given index. if the index is negative, the dive will
- * be added according to reverse chronological order */
-void add_single_dive(int idx, struct dive *dive)
+/* add a dive at the given index to a dive table. if the index is negative,
+ * the dive will be added according to dive_less_than() order */
+void add_dive_to_table(struct dive_table *table, int idx, struct dive *dive)
 {
 	int i;
 	if (idx < 0)
-		idx = dive_get_insertion_index(dive);
-	grow_dive_table(&dive_table);
-	dive_table.nr++;
-	if (dive->selected)
-		amount_selected++;
+		idx = dive_get_insertion_index(table, dive);
+	grow_dive_table(table);
+	table->nr++;
 
-	for (i = idx; i < dive_table.nr; i++) {
-		struct dive *tmp = dive_table.dives[i];
-		dive_table.dives[i] = dive;
+	for (i = idx; i < table->nr; i++) {
+		struct dive *tmp = table->dives[i];
+		table->dives[i] = dive;
 		dive = tmp;
 	}
+}
+
+/* add a dive at the given index in the global dive table and keep track
+ * of the number of selected dives. if the index is negative, the dive will
+ * be added according to dive_less_than() order */
+void add_single_dive(int idx, struct dive *dive)
+{
+	add_dive_to_table(&dive_table, idx, dive);
+	if (dive->selected)
+		amount_selected++;
 }
 
 bool consecutive_selected()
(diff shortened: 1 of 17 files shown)


bstoeger:
Heh... And Git storage indeed fails. Will check it out.

bstoeger:
Ok - with latest commit the git-test now runs locally. Perhaps that commit should be moved to the begin of the series to not have points where the tests fail.

dirkhh:
Oh lovely. More churn. Yay. :-)
(just kidding, starting a review now... but the PR text made me giggle)

core/divelist.c
```diff
@@ -844,41 +846,39 @@ struct dive *last_selected_dive()
...
-	pprev = dive->pprev;
-	*pprev = next;
-	if (next)
-		next->pprev = pprev;
-
+	idx = get_idx_in_table(&trip->dives, dive);
+	if (idx)
+		unregister_dive_from_table(&trip->dives, idx);
 	dive->divetrip = NULL;
-	if (was_autogen)
```
dirkhh: I notice that you no longer handle the `was_autogen` flag (which has become unused).
That seems like it would be a problem.

core/divelist.c
```diff
@@ -1737,6 +1738,25 @@ bool dive_less_than(const struct dive *a, const struct dive *b)
...
 }
 
+/* Trips are compared according to the first dive in the trip.
+ * Even though it shouldn't happen, take care about "empty" trips.
+ * Since a dive can only belong to one trip, no two trips should
+ * compare as equal
+ */
+static int comp_trips(const struct dive_trip *a, const struct dive_trip *b)
+{
+	if (a->dives.nr <= 0)
```
dirkhh, line 1750: when would the number be negative?

core/divelist.c
```diff
@@ -1757,6 +1758,20 @@ bool trip_less_than(const struct dive_trip *a, const struct dive_trip *b)
...
 }
 
+static int comp_dive_or_trip(struct dive_or_trip a, struct dive_or_trip b)
+{
+	if (a.dive && b.dive)
+		return comp_dives(a.dive, b.dive);
+	if (a.trip && b.trip)
+		return comp_trips(a.trip, b.trip);
+	return a.dive ? 1 : -1;
+}
```
dirkhh, line 1782: The first two cases make sense. But then the last one might deserve some explanation. So if you compare a dive to a trip (why would you ever?) the result depends on whether a is a dive? I don't get it.

dirkhh (commented):
I so hate that I cannot comment on commit messages in GitHub.
First commit - read the commit message again. You're adding a dive to a dive trip, not a dive to a dive.

d75df5a4819e3c731ea000669ea56b6db581db5c the commit message states that the merging of trips is now the callers responsibility. But I don't see code that actually handles that case added to the caller.

af84c763c5d5d8d4d74bddc0d715df9ff795ffe9 this is translating too literally from German... Start with "Because of a copy-..."

I worry about the implication of the file format change... if you use this version to write a data file, what happens if you then open that data file with 4.8.4 or earlier?

bstoeger:
> I worry about the implication of the file format change... if you use this version to write a data file, what happens if you then open that data file with 4.8.4 or earlier?

File format doesn't (at least not deliberately) change - trips are written just as before with date. Only on *reading* the date is ignored, because it is implicitly calculated from the first dive. Thus the file-format is backwards compatible.

core/divelist.c
bstoeger: Yeah, code is bogus (as in missing something) - will fix and comment.

core/divelist.c
bstoeger: Hopefully never. Actually it should also never be zero - this is just "defensive" coding so I thought lets catch all the bogus cases. But I can change it to `== 0` if you want.

core/divelist.c
bstoeger: Yes, I will have to recheck that change.

bstoeger:
> d75df5a the commit message states that the merging of trips is now the callers responsibility. But I don't see code that actually handles that case added to the caller.

Yes, that's a misleading formulation - will reword. The merging will come with changes to the import code, which is not yet ready. I still think the current behavior can be removed as the only usecase I see is when importing a divelog onto itself. Perhaps for two diverging log-files? I wonder if there should be two options: "Import log files" and "Merge log files". The latter would merge overlapping trips..? This still needs some thought.

core/divelist.c
dirkhh: I'm ok with defensive coding. I love a comment that says "should never happen, but heck, just to be safe" or something...

core/divelist.c
dirkhh: Yay, I actually caught an issue :-)

dirkhh:
> Yes, that's a misleading formulation - will reword. The merging will come with changes to the import code, which is not yet ready. I still think the current behavior can be removed as the only usecase I see is when importing a divelog onto itself. Perhaps for two diverging log-files? I wonder if there should be two options: "Import log files" and "Merge log files". The latter would merge overlapping trips..? This still needs some thought.

So we made the decision very early on that whenever we import dives we should merge them. Can you tell me a scenario where you think not merging is the right thing to do? Just trying to understand your logic.

bstoeger:
> So we made the decision very early on that whenever we import dives we should merge them. Can you tell me a scenario where you think not merging is the right thing to do? Just trying to understand your logic.

That's fine by me. We just have to be careful that we don't talk about different notions of merging. What the code currently does (or rather attempts to do) is merge trips with the *same* date. I.e. trips where the first dive has exactly the same start time. The only way I see that happen is if you import a log file onto itself or two log files that diverged from a single parent. But even that won't work if the date of the first dive of a trip was edited.

But note that this doesn't even work right now: When I try to import a dive log onto itself, I got all trips duplicated with the exception of a single dive. It's quite likely that I broke that recently, but I'm not sure if it's worth debugging...? Probably I should, as it seems to be a regression.

What I was hinting at above was rather a more general merge: If two trips overlap [end time of last dive of trip A is after start time of first dive of trip B], then join A and B to a single trip. This would have the advantage that it would also work if the first dive of a trip was edited. But I'm not sure that you want to *always* do that. And of course it brings the classic problem: Dive log 1 has disjoint trips A and B, Divelog 2 has a trip C that connects A and B.

But maybe I'm overthinking things...? 😛

dirkhh (approved):
These all seem to make sense. Merging

janmulder:
> > I worry about the implication of the file format change... if you use this version to write a data file, what happens if you then open that data file with 4.8.4 or earlier?
> 
> File format doesn't (at least not deliberately) change - trips are written just as before with date. Only on _reading_ the date is ignored, because it is implicitly calculated from the first dive. Thus the file-format is backwards compatible.

Hmm. I just open my logbook after this PR, and get a red bar "Unmatched action 'time'" coming from `load_git()`. Issued for every trip I have in my logbook. So, on reading, its not ignored, but causes complaints from the parser.

bstoeger:
Gah. Sorry - will fix. I really should start using git-storage. I have no use for it, because I simply put my log on my onwCloud account, so I depend on Travis. But that obviously can't find UI issues. 🙁

bstoeger:
@dirkhh: What do you think about demoting the "Unmatched action" message to a debug-message? That would make forward- and backward-compatibility so much easier.

janmulder:
Ok, I'm not Dirk, but my 2 cents :-) Given this example, the data format is unchanged (date/time of a trip are still saved, but not supposed to be interpreted on read). That is strange in itself, as why do we write it out, and not use it on read any more? But when we are writing it, that now bogus data will stay forever in the logbook. And when we hide this from the UI (aka making it a debug message), nobody will ever notice, and we might end up with all kinds of unused crap in the logbook data.

For me, forward compatibility should be invisible to the user (on trivial cases like this), or with some explicit conversion warning we had once with a big change in format. Backward compatible ... well, the data is still read even with warnings, and that seems enough to me.

bstoeger:
> That is strange in itself, as why do we write it out, and not use it on read any more?

We write it so that the file can be read by older versions. Seems especially important for git storage, which will be used from different versions.

We don't read it, because the information is redundant.

> But when we are writing it, that now bogus data will stay forever in the logbook.

It's not any more bogus than before - it's just redundant (always was) and still needed for older versions.

>  And when we hide this from the UI (aka making it a debug message), nobody will ever notice, and we might end up with all kinds of unused crap in the logbook data.

But what's the difference to handling the value and ignoring it? That's hiding it as well. And the chances that someone will note in the future are not much bigger. It could even be that a debug-message is more visible than a code-comment.

> For me, forward compatibility should be invisible to the user (on trivial cases like this), or with some explicit conversion warning we had once with a big change in format. Backward compatible ... well, the data is still read even with warnings, and that seems enough to me.

I'm not sure that I understand the point: In this case the same file does not produce messages with old versions but does with new versions. We can't remove the field, because old versions will fail and not just produce a message.

dirkhh:
On my phone, so email response.
I think writing the data is fine for a couple of days so people can switch back and forth (but make sure you don't write bogus data), but fundamentally we should just not write it anymore.
And no, please don't hide those messages.

My concern about file format changes was more along the lines of
- understand what an old version does if it sees and XML or git file in the new format
- document the change
- think about whether we need to rev the file format version number

bstoeger:
> I think writing the data is fine for a couple of days

A couple of *days*? That would mean once you upgrade to the new version on one device you have to update *all* devices. Doesn't that make the whole git thing a bit useless..? I was thinking more in terms of a couple of *years*. 😁

dirkhh:
That's why I said I want to play with what this does when opened in an older version.

But yes, I understand your concern.

bstoeger:
> That's why I said I want to play with what this does when opened in an older version.

Both parsers (xml & git) seem to handle it gracefully. I'm not sure if the `trip->when` field will be filled out correctly and if not, if that even matters.

This is the patch on top of master that removes the field from both files:

```
diff --git a/core/save-git.c b/core/save-git.c
index c53c2bb97..e5d0a583f 100644
--- a/core/save-git.c
+++ b/core/save-git.c
@@ -731,17 +731,12 @@ static void create_trip_name(dive_trip_t *trip, struct membuffer *name, struct t
 	put_string(name, "trip");
 }
 
-static int save_trip_description(git_repository *repo, struct dir *dir, dive_trip_t *trip, struct tm *tm)
+static int save_trip_description(git_repository *repo, struct dir *dir, dive_trip_t *trip)
 {
 	int ret;
 	git_oid blob_id;
 	struct membuffer desc = { 0 };
 
-	put_format(&desc, "date %04u-%02u-%02u\n",
-		   tm->tm_year, tm->tm_mon + 1, tm->tm_mday);
-	put_format(&desc, "time %02u:%02u:%02u\n",
-		   tm->tm_hour, tm->tm_min, tm->tm_sec);
-
 	show_utf8(&desc, "location ", trip->location, "\n");
 	show_utf8(&desc, "notes ", trip->notes, "\n");
 
@@ -786,7 +781,7 @@ static int save_one_trip(git_repository *repo, struct dir *tree, dive_trip_t *tr
 	free_buffer(&name);
 
 	/* Trip description file */
-	save_trip_description(repo, subdir, trip, tm);
+	save_trip_description(repo, subdir, trip);
 
 	/* Make sure we write out the dates to the dives consistently */
 	first = MAX_TIMESTAMP;
diff --git a/core/save-xml.c b/core/save-xml.c
index 4abb3cc12..7cb9ff385 100644
--- a/core/save-xml.c
+++ b/core/save-xml.c
@@ -520,7 +520,6 @@ static void save_trip(struct membuffer *b, dive_trip_t *trip, bool anonymize)
 	struct dive *dive;
 
 	put_format(b, "<trip");
-	show_date(b, trip_date(trip));
 	show_utf8(b, trip->location, " location=\'", "\'", 1);
 	put_format(b, ">\n");
 	show_utf8(b, trip->notes, "<notes>", "</notes>\n", 0);
```

bstoeger:
I have to admit that I am a bit confused. Intuitively I would have guessed that in designing a parser:
1) If there are attributes that you don't know: Ignore them, as they might be an extension.
2) If an attribute you need is missing complain.

It turns out that the current parser works exactly the *opposite* way. That means adding an optional attribute will always give tons of user-visible warnings in older versions.

Can you confirm that this is the intended behavior?

PR #1859, opened 2018-11-18
Mobile copy ui

mturkia:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

UI for the mobile copy-paste functionality.
### Release note:

added
### Documentation change:

added
### Mentions:

@dirkhh @janmulder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 - Fix bug in cloud save after removing dives from a trip
 - Dive: Perform more accurate OTU calculations, and include
   OTU calculations for rebreather dives [#1851 & #1865].
+- Mobile: UI for copy-paste
 - Mobile: add initial copy-paste support
 - Desktop: translate trip date
 ---
--- a/Documentation/mobile-manual.txt
+++ b/Documentation/mobile-manual.txt
@@ -244,8 +244,9 @@ paste operation.
 
 image::mobile-images/CopyPaste.jpg["FIGURE: copy dive details in list",align="center"]
 
-Currently the details to be copied are hard-coded and include the following
-information:
+To choose what dive details to copy, long-press the copy button. This will open
+up a configuration page where you can toggle the details you want to copy over
+to the destination. By default, the following fields are copied:
 
 - divemaster
 - buddy
--- a/mobile-widgets/qml/CopySettings.qml
+++ b/mobile-widgets/qml/CopySettings.qml
@@ -1,0 +1,168 @@
+// SPDX-License-Identifier: GPL-2.0
+import QtQuick 2.6
+import QtQuick.Controls 2.2 as Controls
+import QtQuick.Window 2.2
+import QtQuick.Dialogs 1.2
+import QtQuick.Layouts 1.2
+import org.kde.kirigami 2.4 as Kirigami
+import org.subsurfacedivelog.mobile 1.0
+
+Kirigami.ScrollablePage {
+	objectName: "CopySettings"
+	id: settingsCopy
+
+	title: qsTr("Copy Settings")
+	background: Rectangle { color: subsurfaceTheme.backgroundColor }
+
+	property real gridWidth: settingsCopy.width - Kirigami.Units.gridUnit
+
+	ColumnLayout {
+		width: gridWidth
+
+		GridLayout {
+			id: copy_settings
+			columns: 2
+			Controls.Label {
+				text: qsTr("Selection for copy-paste")
+				font.pointSize: subsurfaceTheme.headingPointSize
+				font.weight: Font.Light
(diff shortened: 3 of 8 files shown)


janmulder:
Tried this. I like the way things are "hidden" in the UI. No cluttering for non interested users, but you need to read the user manual to see all the possibilities of the copy/paste capability. Obviously, some developers and users might think the opposite (the UI should show all capabilities and the user shall be guided by intuition, and not needing a manual ever, but whatever. I like this hidden style).

I believe that the chosen copy/paste preferences need to be stored and retrieved on next sessions. For users that like a different than default copy/paste behavior, not storing them is a nuisance.

It does not seem to paste the location. Did not investigate why, but this seems a bug to me.

mturkia:
I just tested the copy of divesite / location and it seems to be working for me. However, there are two peculiarities.
- Location on divelist is not updated
- I was not able to manually edit the location of an existing dive (not related to this patch). Neither selecting an existing site nor modifying the text worked.

janmulder:
Ok, next test. Its getting confusing ... tested on device. My first Qt 5.12, clang and arm64 build, including this PR (obviously). 

All is copy/pasted (that is selected, so also the location + gps data). But, non is shown. Not on the divelist, not in the details. Scrolling away (and thus invalidating QML cached data), reveals finally all data pasted. But please keep in mind ... a very experimental build context.

dirkhh:
Haha, talk about hard to discover UI. I figured I'd just try it. Nope, couldn't figure it out. Read the code. Yeah, that makes sense...
Locations aren't copied for me. Cached QML pages stay unmodified - so user experience is really unexpected.
@mturkia are you actively working on this? Are you looking for help with some of this?

mturkia:
I have tried a couple of times to get the info updated on screen automatically and failed. So a bit of help would be appreciated.

janmulder:
When I add the following code, automatic updating of the divelist does work for me (both on mobile-on-desktop, and own Android build), notice, I rebased this PR first on current master first, so not sure that the line numbers below are correct. This code is at the end of `pasteDiveData`.

```
------------------------ mobile-widgets/qmlmanager.cpp ------------------------
index 775305698..425c29ad2 100644
@@ -1440,0 +1441,4 @@ void QMLManager::pasteDiveData(int id)
+
+	DiveListModel *dm = DiveListModel::instance();
+	int modelIdx = dm->getDiveIdx(d->id);
+	DiveListModel::instance()->updateDive(modelIdx, d);
```

This said. There is still something weird left. The `what` data used in `pasteDiveData` does not always correspond to the setting in the UI. Did not invesitigate this further. Furthermore, my earlier remark "I believe that the chosen copy/paste preferences need to be stored and retrieved on next sessions. For users that like a different than default copy/paste behavior, not storing them is a nuisance." is also still valid.

mturkia:
> When I add the following code, automatic updating of the divelist does work for me

Thanks, I'll take a look at this next week.

I agree that it would be convenient if the selection what to copy-paste was saved. The desktop version does not save it either, instead it presents the selection dialog whenever someone is copying data.

mturkia:
The update of divelist seems to be working now with Jan's trick.

I am wondering about the defaults. The desktop version does not save what one wants to copy-paste over sessions. It instead shows a dialog with nothing selected when one does the copy operation. Would this option be actually feasible for the mobile as well? This would, at least, be easy to find unlike the long press on copy button apparently is. Or should this memory be implemented in both versions?

janmulder:
Wrt. the defaults. I'm not sure on this either. I think that saving the list of fields in the preferences on mobile only (and leave desktop as it is) is acceptable. One could argue that this is inconsistent, but I can easily point out 100 other things things that are different in mobile and desktop. The difference in processing power, screen real-estate, and way of interfacing with the application (no mouse of mobile) makes this valid: it are simply different apps, that use a common code base (as much as possible).

I think that I would opt for stored in preferences on mobile.

dirkhh:
I feel bad that this has been sitting for a month. Between a crazy travel schedule, hurting my foot and being in quite a bit of pain when sitting (which cut down on the time I was willing to spend sitting at my computer in the evening), and then getting one heck of a cold... I just haven't spent nearly as much time on Subsurface in the past six weeks than I really wanted.

I think giving the more challenging interaction with the UI on mobile, having mobile remember the choice does make more sense. But I'll admit that I'm slightly lost where we are overall with this PR...

janmulder:
> Why is Travis failing?

I kicked the builds, and all went well. That "mac" build is very close to the max job length (50 min) of Travis. And its fully unclear to me why that build takes so long.

I'm happy to get this merged.

PR #1860, opened 2018-11-18
[Cleanup] Simplify tripflag logic

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a trivial cleanup to the tripflag logic. The only use of tripflag is to mark dives that have been explicitly removed from a trip and avoid them from being autogrouped. Remove the other uses.
### Changes made:

1) Remove unused macros
2) Expand a macro that was used in only one function and add a comment
3) Remove a redundant trip flag

### Release note:

No.
### Documentation change:

No.


core/divelist.c
```diff
@@ -1020,7 +1020,9 @@ dive_trip_t *get_dives_to_autogroup(int start, int *from, int *to, bool *allocat
 			continue;
 		}
 
-		if (!DIVE_NEEDS_TRIP(dive)) {
+		/* Only consider dives that have not been explicitly removed from
+		 * a dive trip by the user in which case tripflag would be NO_TRIP. */
```
neolit123: it's also OK to make the marco divelist.c local.
but this change is also fine.

neolit123 (approved):
LGTM

bstoeger:
I just found a subtle thing that might mean that the second commit is wrong:

In `get_preferred_trip` there is the code:
```
        /*
         * The larger tripflag is more relevant: we prefer
         * take manually assigned trips over auto-generated
         * ones.
         */
        if (a->tripflag > b->tripflag)
                return a;

        if (a->tripflag < b->tripflag)
                return b;
```
So actually it *does* make a difference whether the tripflag is `IN_TRIP` or `ASSIGNED_TRIP`.

Weirdly enough, the XML and git loaders seem to behave differently: the former explicitly sets the lower-priority `IN_TRIP`, the latter doesn't.

bstoeger:
But it seems like that worked only partially anyway (perhaps destroyed by me?):
In `create_and_hookup_trip_from_dive()` `IN_TRIP` is set, but immediately afterwards it gets overwritten in `add_dive_to_trip()`.

I replaced the trip-selection logic by `trip->autogen`. Seems a bit more logical to me. In any case I will remove the complexity-low tag, as this all seems to be rather subtle in nature!

Edit: the offending code seems to be from around 2012-ish and so far nobody complained. :)

bstoeger:
Ok - Completely removed the `tripflag` enum and replaced it by a `notrip` boolean. Only compile&run tested.

dirkhh:
Uh. Yeah. Old code. I'll ask Linus to comment.

bstoeger:
Jeez. I have that talent of stepping on my own foot. Rebased - fingers crossed that I got it right.

torvalds:
Looks good to me. The ASSIGNED_TRIP vs IN_TRIP thing might have made some difference for some odd case, but is not worth worrying about. Let's simplify and see if anybody ever notices anything.

dirkhh (approved):
If you go back in time to c804c4e02e0530889697fab524cba294a55ffeeb (more than six years ago) you can see why we had the trip flag the way it is. The idea was that we wanted to be able to tell apart if a trip was automatically generated or if it was explicitly (in the UI - or by saving autogenerated trips to disk) set. So you could play with trip auto generation, without losing explicit assignment.
But I think no one besides me ever thought this was important because people pretty much consistently either HATE trips or always just use autogenerated ones.
So yeah, whatever.

PR #1861, opened 2018-11-18
[Cleanup] silence warnings

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes noisy warning I get since upgrading to Kubuntu 18.10 and accordingly to gcc 8.2.
The first one is a bit dubious - an alternative would be to increase the filename buffers, but I understand that this is supposed to generate MS-DOS 8.3 filenames!?

### Changes made:

Silence warnings.

### DIFF
--- a/core/uemis-downloader.c
+++ b/core/uemis-downloader.c
@@ -516,6 +516,22 @@ static void uemis_increased_timeout(int *timeout)
 	usleep(*timeout);
 }
 
+static char *build_ans_path(const char *path, int filenr)
+{
+	char *intermediate, *ans_path, fl[13];
+
+	/* Clamp filenr into the 0..9999 range. This is never necessary,
+	 * as filenr can never go above UEMIS_MAX_FILES, but gcc doesn't
+	 * recognize that and produces very noisy warnings. */
+	filenr = filenr < 0 ? 0 : filenr % 10000;
+
+	snprintf(fl, 13, "ANS%d.TXT", filenr);
+	intermediate = build_filename(path, "ANS");
+	ans_path = build_filename(intermediate, fl);
+	free(intermediate);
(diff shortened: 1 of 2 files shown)


core/uemis-downloader.c
```diff
@@ -583,7 +593,7 @@ static bool uemis_get_answer(const char *path, char *request, int n_param_in,
 		if (import_thread_cancelled)
 			return false;
 		progress_bar_fraction = filenr / (double)UEMIS_MAX_FILES;
-		snprintf(fl, 13, "ANS%d.TXT", filenr - 1);
+		snprintf(fl, 13, "ANS%d.TXT", sanitize_filenr(filenr - 1));
```
neolit123: how about turning this in a utility function:
```
void write_filenr(char *buf, int bufsz, int filenr) 
{
    filenr = filenr < 0 ? 0 : filenr % 10000;
    snprintf(buf, bufsz, "ANS%d.TXT", filenr);    
}

// usage
write_filenr(fl, sizeof(fl), filenr - 1)
```

bstoeger: Yes, fine by me. The question remains: do we actually want to code around compiler warnings?

dirkhh: I'd like the `write_filenr()` helper better...

dirkhh (approved):
I like this in general - but I think having a helper function to actually just write that ANS file would be more natural.

bstoeger:
Alright - had a go at putting this in an extra function. Problem: I have no device to test!

dirkhh (approved):

PR #1864, opened 2018-11-20
Git parser: don't produce user-visible error on git parsing

bstoeger:
In 64e6e435f82801f4f440ef5b1caf58a91a7c9929 the when field of struct
trip was removed. Accordingly it was not read from git repositories.
This produced a large amount of user-visible error messages.

Reinstate the trip-date and time parsing functions, but ignore
the value.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes error messages barfed at the user that are due to ignoring trip-date and time.

See discussion in #1858. To me, the behavior of the parser (complaining about unknown attributes) is very surprising. But from the discussion so far I understand that it's supposed to be that way..

### Changes made:

1) Add trip date and time parsing functions but ignore the value.

### Mentions:

@janmulder @dirkhh

### DIFF
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -799,6 +799,14 @@ static void parse_dc_event(char *line, struct membuffer *str, void *_dc)
 	}
 }
 
+/* Not needed anymore - trip date calculated implicitly from first dive */
+static void parse_trip_date(char *line, struct membuffer *str, void *trip)
+{ UNUSED(line); UNUSED(str); UNUSED(trip); }
+
+/* Not needed anymore - trip date calculated implicitly from first dive */
+static void parse_trip_time(char *line, struct membuffer *str, void *trip)
+{ UNUSED(line); UNUSED(str); UNUSED(trip); }
+
 static void parse_trip_location(char *line, struct membuffer *str, void *_trip)
 { UNUSED(line); dive_trip_t *trip = _trip; trip->location = get_utf8(str); }
 
@@ -1001,7 +1009,7 @@ static void site_parser(char *line, struct membuffer *str, void *_ds)
 struct keyword_action trip_action[] = {
 #undef D
 #define D(x) { #x, parse_trip_ ## x }
-	D(location), D(notes),
+	D(date), D(location), D(notes), D(time),
 };
 
 static void trip_parser(char *line, struct membuffer *str, void *_trip)


bstoeger:
This version seems more logical to me:
```
commit eb774a905e6d4f6cdbb2fa05499b794c4df7a66a
Author: Berthold Stoeger <bstoeger@mail.tuwien.ac.at>
Date:   Mon Nov 19 19:47:57 2018 +0100

    Git loader: demote "Unmatched action" message to debug-message
    
    In 66c3663fbdd7f3322f8f98a780a2fc6a1ba571b6 reading of the
    trip-date and time was removed, as this is redundant information.
    
    But this would produce "Unmatched action" errors. Demote to debug
    message to ease forward- and backward-compatibility.
    
diff --git a/core/load-git.c b/core/load-git.c
index 1447becb2..f85276dc0 100644
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -463,7 +463,8 @@ static int match_action(char *line, struct membuffer *str, void *data,
 		else
 			low = mid+1;
 	}
-report_error("Unmatched action '%s'", line);
+	if (verbose > 1)
+		fprintf(stderr, "Unmatched action '%s'\n", line);
 	return -1;
 }
```

dirkhh:
I understand what you are trying to do. Please give me time to think this through and look at the alternatives.

bstoeger:
Take your time - I just didn't want to give the impression that I don't care about the bugs that I introduced. 🙂

torvalds:
I disagree.

If you switch to an older version, that doesn't understand some new field that you created using a never version of subsurface, you will be losing data.

That's not some "hide it from the user" warning case. That's a serious problem. People need to be warned about it.

torvalds:
Side note: it's perhaps worth noting that the git model of dive object caching means that losing data is *not* necessarily obvious. 

You may lose data when parsing the new state, but then when you save your changes back, only dives that were actively changed will be written back (the others will match the git object and the save will be optimized away). So the loss of data can be hidden: you can add new dives with an old version, and never notice that the parsing missed data on older dives, *unless* you made some other unrelated change to an older dives (like updated a buddy or did a dive renumber or whatever).

Regardless, I'd *much* rather fix the "trip.time" issue by simply keeping the parsing of it, and igmoring the data as redundant, rather than silencing a very valid warning in general.

bstoeger:
Point taken - I didn't think about the "overwrite log with old version" case..

dirkhh (approved):
Based on @torvalds' review and my perception, the actual PR is the right thing to do. So let's merge it

PR #1870, opened 2018-11-22
Cleanup: make "struct dive *" and "struct dive_trip *" Qt metatypes

bstoeger:
Just as we did for pointer to struct dive_site, make pointers to
struct dive and struct dive_trip "Qt metatypes". This means that
they can be passed through QVariants without taking a detour via
void *.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Make `struct dive *` and `struct dive_trip *` Qt metatypes, so that they can be passed through `QVariants` and later through `QML`. I hope that especially the latter will make it possible to remove some of the idiosyncrasies of the mobile dive list. For example, the way I read the code, the trip header is formatted for *every* dive and then the trip headers are compared. It would appear much more logical to simply compare the `dive_trip *` pointers.

Note: this is a bit hairy, because `QVariant`s often cases don't give compile time errors, but simply return null pointers! E.g. if in a `QVariant` returning functions you do
```
   return trip;
```
On readout this will give NULL.

### Changes made:

1) Make `struct dive *` and `struct dive_trip *` Qt metatypes

### DIFF
--- a/core/dive.h
+++ b/core/dive.h
@@ -758,16 +758,22 @@ extern void set_git_prefs(const char *prefs);
 extern char *get_dive_date_c_string(timestamp_t when);
 extern void update_setpoint_events(const struct dive *dive, struct divecomputer *dc);
 
-#ifdef __cplusplus
-}
-#endif
-
 extern weight_t string_to_weight(const char *str);
 extern depth_t string_to_depth(const char *str);
 extern pressure_t string_to_pressure(const char *str);
 extern volume_t string_to_volume(const char *str, pressure_t workp);
 extern fraction_t string_to_fraction(const char *str);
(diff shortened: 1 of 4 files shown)


bstoeger:
> For example, the way I read the code, the trip header is formatted for every dive and then the trip headers are compared. It would appear much more logical to simply compare the dive_trip * pointers.

And now I understand the reason - the section is defined as a *string* and can't be anything we want.
https://doc-snapshots.qt.io/qt5-5.11/qml-qtquick-listview.html

This means that it will not work the way that I hoped. Nevertheless, this still makes the desktop-models a tiny bit more palatable.

PR #1872, opened 2018-11-22
[RFC] Mobile: don't format trip heading for all dives (Warning: contains #1870)

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I checked the source code of "QuickListView" and I found no obvious way of passing anything other than a string as "section". Perhaps I'm simply blind. So this is the best I could come up with to remove that crazy format the trip-header for every dive instead of every trip. Note that this is not so much a performance thing but a code-cleanliness / flexibility thing. I'm not very happy about it, but it is a step forward I think. Especially if someone can tell us how this is really planned to be used.

I know that it worked before, but this is better for my mental health. 😛

### Changes made:

1) Replace tripMeta by tripId and only pass the pointer, not the fully formatted strings.
2) Use distinct functions for formatting the trip header and short-date.
### Related issues:

Contains #1870 

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/dive.h
+++ b/core/dive.h
@@ -758,16 +758,22 @@ extern void set_git_prefs(const char *prefs);
 extern char *get_dive_date_c_string(timestamp_t when);
 extern void update_setpoint_events(const struct dive *dive, struct divecomputer *dc);
 
-#ifdef __cplusplus
-}
-#endif
-
 extern weight_t string_to_weight(const char *str);
 extern depth_t string_to_depth(const char *str);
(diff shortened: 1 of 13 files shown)


bstoeger:
Added two commits in an attempt to converge desktop and mobile. I know it's a hopeless endeavor, but one can at least try.

dirkhh (approved):
I can't find anything wrong with this.

PR #1875, opened 2018-11-26
Dive list: fix off-by-two bug in DiveTripModel

bstoeger:
Commit 911edfca712a046944de6d033cc4b8dd50cedfc3 changed the dive list
on desktop to update positions of trips when adding/removing dives.
A very unlikely case, but necessary for consistency.

For a trip to be moveable down, its index has to be one-less than
the maximum index, which is "items - 1". The code was doubly wrong:
it forget the "- 1" and checked for less-or-equal instead less-than.
Thus this was effectively an off-by-two error. Fix it.

Signed-off-by: Berthold Stoeger <bstoeger@mail.tuwien.ac.at>

### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This fixes a recently introduced, easily pulled off crash condition: delete dives from or add dives to either the last or the second to last dive.

Sometimes I wonder where my brain is. 😛 
### Changes made:

1) Fix silly off-by-two error.

### Release note:

No.
### Documentation change:

No.

PR #1876, opened 2018-12-03
Invalidate git cache when removing dive from trip

atdotde:
... otherwise this change is not saved when saving to git.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a bug reported in the google group 
https://groups.google.com/d/msgid/subsurface-divelog/aa812413-8fcc-4817-ab3b-964144e2046e%40googlegroups.com?utm_medium=email&utm_source=footer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Fix bug in cloud save after removing dives from a trip
 - Dive: Perform more accurate OTU calculations, and include
   OTU calculations for rebreather dives [#1851 & #1865].
 - Mobile: add initial copy-paste support
--- a/desktop-widgets/command_divelist.cpp
+++ b/desktop-widgets/command_divelist.cpp
@@ -92,6 +92,7 @@ dive *DiveListBase::addDive(DiveToAdd &d)
 	res->hidden_by_filter = !show;
 
 	add_single_dive(d.idx, res);		// Return ownership to backend
+	invalidate_dive_cache(res);		// Ensure that dive is written in git_save()
 
 	// If the dive to be removed is selected, we will inform the frontend
 	// later via a signal that the dive changed.
@@ -191,7 +192,7 @@ std::vector<dive *> DiveListBase::addDives(std::vector<DiveToAdd> &divesToAdd)
 
 // This helper function renumbers dives according to an array of id/number pairs.
 // The old numbers are stored in the array, thus calling this function twice has no effect.
-// TODO: switch from uniq-id to indexes once all divelist-actions are controlled by undo-able commands
+// TODO: switch from uniq-id to indices once all divelist-actions are controlled by undo-able commands
 static void renumberDives(QVector<QPair<dive *, int>> &divesToRenumber)
 {
 	for (auto &pair: divesToRenumber) {
@@ -199,6 +200,7 @@ static void renumberDives(QVector<QPair<dive *, int>> &divesToRenumber)
 		if (!d)
 			continue;
 		std::swap(d->number, pair.second);
+		invalidate_dive_cache(d);
 	}
 
 	// Emit changed signals per trip.
@@ -239,6 +241,7 @@ static OwningTripPtr moveDiveToTrip(DiveToTrip &diveToTrip)
 	// Store old trip and get new trip we should associate this dive with
 	std::swap(trip, diveToTrip.trip);
 	add_dive_to_trip(diveToTrip.dive, trip);
+	invalidate_dive_cache(diveToTrip.dive);		// Ensure that dive is written in git_save()
 	return res;
 }
 


bstoeger:
Ouch. I think that this is not the correct position for `invalidate_dive_cache()`. Because consider the following (admittedly synthetic) situation:

1) Remove dive from trip
2) Save to git (<- new git_id will be set)
3) Undo
4) Save to git (<- will still have new git_id, not saved)

Therefore, I think that `invalidate_dive_cache()` should be called in the corresponding `redo()` and `undo()` functions.

But even more correctly, the undo-machinery should remember the git_id in the case of undo.

atdotde:
@bstoeger where do you suggest to put it? Maybe in 
struct dive_trip *unregister_dive_from_trip(struct dive *dive, short was_autogen)
and
void add_dive_to_trip(struct dive *dive, dive_trip_t *trip)

(the logic being, whenever you move dives around you need to invalidate the cache)? Am I mussing something?

bstoeger:
As a short-term fix this seems like the most consistent place. After all, most of the other calls to `invalidate_dive_cache` are at this level too.

atdotde:
Here is the change suggested by @bstoeger. It seems to me that the Travis build failure for android is unrelated to this PR.

bstoeger:
> struct dive_trip *unregister_dive_from_trip(struct dive *dive, short was_autogen)

Note that you added the call to `remove_dive_from_trip()`, which calls `unregister_dive_from_trip()` and not vice-versa. This will therefore still not work, because the undo command calls the latter directly.

Moreover, I wonder if a invalidate-cache call also has to be added to `renumberDives` in `command_divelist.cpp`.

atdotde:
Included @bstoeger 's suggestions. I have no overview if this is needed other places.

bstoeger (approved):
lgtm

bstoeger:
Damn, just discovered a different problem:

At least the normal parser will use `add_dive_to_trip()` to add the dive, which will first call `remove_dive_from_trip()`, which will then call `invalidate_dive_cache()`.

Thus, the current version will invalidate all git-ids on load. :( So perhaps this is not the best place to do this after all - sorry. I will try to get a grip on this invalidate_dive_cache thing tonight.

dirkhh:
Linus doesn't get @torvalds tags on GitHub (because WAY too many people tag him). So if you want a review from him, please send him email (something like "Please review this Subsurface PR" with a link to the PR in the body and maybe a word or two why his review would be good...

atdotde:
> On 7. Dec 2018, at 20:28, bstoeger <notifications@github.com> wrote:
> 
> Yes, the way this PR stands it will invalidate all git-ids on load. Sorry for the suggestion.
> 
> This patch on top of this PR works for me but independent confirmation would be nice:
> 
> 

I have the impression, this needs further work: With your patch, starting from an empty log, I can make it crash (in invalidate_dive_cache) by opening the planner and immediately saving that dive. This crash does not happen with  my last version.

In addition, both versions (but not the 4.8.4 release) crash when in my log I try to remove the earlier of two dives in a trip (this crashes in dive_or_trip_less_than according to the crash report).

Unfortunately, I have somehow borked my debugger at the moment so it would be great if you could take a look.

Thanks
Robert

--
.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oOo.oO
Robert C. Helling     Elite Master Course Theoretical and Mathematical Physics
                      Scientific Coordinator
                      Ludwig Maximilians Universitaet Muenchen, Dept. Physik
print "Just another   Phone: +49 89 2180-4523  Theresienstr. 39, rm. B339
    stupid .sig\n";   http://www.atdotde.de

bstoeger:
Yes, the patch was bogus. Here a better version:
```
diff --git a/core/divelist.c b/core/divelist.c
index 1af0ccdda..ee7986466 100644
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -902,7 +902,6 @@ struct dive_trip *unregister_dive_from_trip(struct dive *dive, short was_autogen
 	if (idx >= 0)
 		unregister_dive_from_table(&trip->dives, idx);
 	dive->divetrip = NULL;
-	invalidate_dive_cache(dive);
 	return trip;
 }
 
@@ -920,7 +919,6 @@ void add_dive_to_trip(struct dive *dive, dive_trip_t *trip)
 	remove_dive_from_trip(dive, false);
 	add_dive_to_table(&trip->dives, -1, dive);
 	dive->divetrip = trip;
-	invalidate_dive_cache(dive);
 }
 
 dive_trip_t *alloc_trip(void)
diff --git a/desktop-widgets/command_divelist.cpp b/desktop-widgets/command_divelist.cpp
index 6f613cce9..ba126947a 100644
--- a/desktop-widgets/command_divelist.cpp
+++ b/desktop-widgets/command_divelist.cpp
@@ -92,6 +92,7 @@ dive *DiveListBase::addDive(DiveToAdd &d)
 	res->hidden_by_filter = !show;
 
 	add_single_dive(d.idx, res);		// Return ownership to backend
+	invalidate_dive_cache(res);		// Ensure that dive is written in git_save()
 
 	// If the dive to be removed is selected, we will inform the frontend
 	// later via a signal that the dive changed.
@@ -240,6 +241,7 @@ static OwningTripPtr moveDiveToTrip(DiveToTrip &diveToTrip)
 	// Store old trip and get new trip we should associate this dive with
 	std::swap(trip, diveToTrip.trip);
 	add_dive_to_trip(diveToTrip.dive, trip);
+	invalidate_dive_cache(diveToTrip.dive);	// Ensure that dive is written in git_save()
 	return res;
 }
 ```

The second problem I couldn't reproduce with my test file. Does it still happen with this patch?

bstoeger:
> The second problem I couldn't reproduce with my test file. Does it still happen with this patch?

This could be the bug fixed in #1875. Please try to rebase on master.

atdotde:
Yes, indeed rebasing helped. This is your latest version. Let's wait for Travis.

PR #1882, opened 2018-12-08
Core: don't remove dive from trip in add_dive_to_trip()

bstoeger:
All callers of add_dive_to_trip() work on freshly generated dives,
with one exception, that was redundant anyway. Therefore it is not
necessary to remove the dive from a potential previous trip. Move the
responsibility of removing the dive from a trip to the caller,
respectively remove the redundant call. Add a warning message in the
case that trip is set.

Background: On import (either download or file-import) we might not
want to add trips to the global trip-list. For example to enable undo
of import but more generally to detangle that data flow. Thus,
add_dive_to_trip() should not mingle with the global trip-list,
which it has to do if a trip is deleted because the old dive was
removed.

Analysis of the add_dive_to_trip() callers:

1) core/dive.c

pick_trip():
    called on freshly generated merged dive.

finish_split():
    called on two freshly generated split dives.

2) core/divelist.c

create_and_hookup_trip_from_dive():
    called on freshly downloaded dive in dive_cb().
    called on freshly downloaded dive in record_uemis_dive().

autogroup_dives():
    called on dive from get_dives_to_autogroup(), which only
    finds dives that are outside of trips.

combine_trips():
    unused - removed in sibling commit.

try_to_merge_into():
    this call was actually erroneous - dive was already added
    to trip in try_to_merge(). Remove call.

3) core/libdivecomputer.c

dive_cb():
    called on freshly downloaded dive.

4) core/uemis_downloader.c

record_uemis_dive():
    called on freshly downloaded dive.

5) core/load_git.c

create_new_dive():
    called on freshly allocated dive.

6) core/parse.c

dive_end():
    called on freshly parsed dive.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This moves the responsibility of removing a dive from a trip from `add_dive_to_trip()` to the caller. Actually, this function was only called on non-trip dives anyway, so no change needed. Add a warning printf in case the function is called on a dive inside a trip.

While this may look like idle code-reshuffling, it does actually simplify the work-in-progress on dive-import. The problem with automatically removing the dive from the trip is that if the trip becomes empty it is deleted from the global trip-list. But we might not want to add all trips to the global trip-list. For example to make download/import undo-able.

### Changes made:

1) Don't remove dive from trip in add_dive_to_trip().

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -19,7 +19,6 @@
  * void insert_trip(dive_trip_t *dive_trip_p)
  * void unregister_trip(dive_trip_t *trip)
  * void free_trip(dive_trip_t *trip)
- * void remove_dive_from_trip(struct dive *dive)
  * void remove_dive_from_trip(struct dive *dive, bool was_autogen)
  * void add_dive_to_trip(struct dive *dive, dive_trip_t *trip)
  * dive_trip_t *create_and_hookup_trip_from_dive(struct dive *dive)
@@ -912,11 +911,14 @@ void remove_dive_from_trip(struct dive *dive, short was_autogen)
 		delete_trip(trip);
 }
 
+/* Add dive to a trip. Caller is responsible for removing dive
+ * from trip beforehand. */
 void add_dive_to_trip(struct dive *dive, dive_trip_t *trip)
 {
 	if (dive->divetrip == trip)
 		return;
-	remove_dive_from_trip(dive, false);
+	if (dive->divetrip)
+		fprintf(stderr, "Warning: adding dive to trip that has trip set\n");
 	add_dive_to_table(&trip->dives, -1, dive);
 	dive->divetrip = trip;
 }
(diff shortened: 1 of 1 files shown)


bstoeger:
Another small disentanglement of the same flavor: Don't remove dive from trip in `unregister_dive()` as the direct callers of that function already did that.

dirkhh:
The two Travis failures both seem to be network related. Restarting.

bstoeger:
Travis-Mac really has an attitude at the moment. :-/

atdotde:
It sees, for some reason, brew update kicks of a huge thing: I restarted your job which was hung on 

Migrating tap caskroom/cask to homebrew/cask

The restarted job gets past that stage but then appears to be downloading a million and one things from home-brew...

dirkhh (approved):
This does look sane.

PR #1883, opened 2018-12-10
Unglobalize downloadTable

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a long-standing pet peeve of mine. While working on import I finally got a reason to un-globalize the `downloadTable` and make it a sub-object of `DownloadThread`. The problem was QML's object-system. I'm not particularly happy about the result, but it seems like a step forward.

This is only compile & run tested, as I don't have a device with dives ATM. The only device I have is BLE anyway, which doesn't seem to be supported by my mobile-on-desktop build.

### Changes made:

1) Cleanup the singleton mess of DCDeviceData. It seems that incorporation of DCDeviceData into QMLManager made this strange pattern useless. In my short tests, the constructor was called at most once.
2) Unifiy DiveImportedModel::setImportedDivesIndexes and DiveImportedModel::repopulate.
3) Change DiveImportedModel::checkStates to a vector. The old code was in-principle wrong, because it assumed that sizeof(bool)=1 [which is the case for all supported architectures though].
4) unglobalize downloadTable

### Release note:

No.
### Documentation change:

No.
### Mentions:

@janmulder: would you care to give this a quick download test?

### DIFF
--- a/core/dive.h
+++ b/core/dive.h
@@ -278,10 +278,10 @@ struct divecomputer {
 #define W_IDX_PRIMARY 0
 #define W_IDX_SECONDARY 1
 
-struct dive_table {
+typedef struct dive_table {
 	int nr, allocated;
 	struct dive **dives;
-};
+} dive_table_t;
 
 typedef struct dive_trip
 {
@@ -425,7 +425,7 @@ extern const struct units SI_units, IMPERIAL_units;
 extern const struct units *get_units(void);
 extern int run_survey, verbose, quit, force_root;
 
-extern struct dive_table dive_table, downloadTable;
+extern struct dive_table dive_table;
 extern struct dive displayed_dive;
 extern unsigned int dc_number;
 extern struct dive *current_dive;
@@ -759,10 +759,14 @@ extern void average_max_depth(struct diveplan *dive, int *avg_depth, int *max_de
 #ifdef __cplusplus
 }
 
-/* Make pointers to dive and dive_trip "Qt metatypes" so that they can
- * be passed through QVariants. */
+/* Make pointers to dive, dive_trip and dive_table "Qt metatypes" so that they can
+ * be passed through QVariants and through QML.
+ * Note: we have to use the typedef "dive_table_t" instead of "struct dive_table",
+ * because MOC removes the "struct", but dive_table is already the name of a global
+ * variable, leading to compilation errors. */
 Q_DECLARE_METATYPE(struct dive *);
 Q_DECLARE_METATYPE(struct dive_trip *);
+Q_DECLARE_METATYPE(dive_table_t *);
 
 #endif
 
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -77,9 +77,6 @@ dive_trip_t *dive_trip_list;
 
 unsigned int amount_selected;
 
-// We need to stop using globals, really.
-struct dive_table downloadTable;
-
 #if DEBUG_SELECTION_TRACKING
 void dump_selection(void)
 {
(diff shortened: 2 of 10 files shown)


janmulder:
Tried a download from my OSTC3. Both with desktop, and with Android device using the mobile app. Both work as before.

bstoeger:
Thanks for testing! That's a wonder... 😛

bstoeger:
Now Travis-Mac went on strike...

core/dive.h
```diff
@@ -425,7 +425,7 @@ extern const struct units SI_units, IMPERIAL_units;
 extern const struct units *get_units(void);
 extern int run_survey, verbose, quit, force_root;
 
-extern struct dive_table dive_table, downloadTable;
+extern struct dive_table dive_table;
```
dirkhh, line 428: maybe we should just make it a typedef as dive_table_t ?

core/dive.h
```diff
@@ -759,10 +759,15 @@ extern void average_max_depth(struct diveplan *dive, int *avg_depth, int *max_de
...
 }
 
-/* Make pointers to dive and dive_trip "Qt metatypes" so that they can
- * be passed through QVariants. */
+/* Make pointers to dive, dive_trip and dive_table "Qt metatypes" so that they can
+ * be passed through QVariants and through QML.
+ * Note: we have to use a typedef for "struct dive_table", because MOC removes the
+ * "struct", but dive_table is already the name of a global variable, leading to
+ * compilation errors. */
+typedef struct dive_table *dive_table_ptr;
```
dirkhh: and then this becomes a `dive_table_t *` and that should work, right?

core/dive.h
```diff
@@ -759,10 +759,15 @@ extern void average_max_depth(struct diveplan *dive, int *avg_depth, int *max_de
...
 Q_DECLARE_METATYPE(struct dive *);
 Q_DECLARE_METATYPE(struct dive_trip *);
+Q_DECLARE_METATYPE(dive_table_ptr);
```
dirkhh: btw, why does this work with `struct dive *` above? That seems inconsistent with what you said in the commit message.
Just curious

dirkhh (commented):
I think this is all good and in a way I'm ok with merging it as is, but I have a couple of comments and I think we need to either decide that we can't have struct types that aren't typedef'ed (and typedef all of them), or that this does actually work and not mess with them, or explain to me what I'm missing :-)

core/dive.h
bstoeger: Yes, I think that should work. Will do it this way if you prefer.

core/dive.h
bstoeger: > btw, why does this work with struct dive * above? 

Because there is no global variable with the name `dive`, but there is a global variable with the name `dive_table`.

What happens is that MOC declares parameters as `dive_table *xyz`, which is perfectly fine C++. Unless there is a variable with the name `dive_table`, which then takes precedence, so `dive_table *xyz` gets parsed as "multiply the object `dive_table` with `xyz`. Apparently the Qt developers didn't have classical C code in mind, where you can't leave out the struct and therefore variables with the same name as structs are rather common.

dirkhh: BTW: I always thought that it was odd that Linus decided against typedefs for some of the structs. He had his logic (typedefs were for "simple" structures that were ok to do, say, assignments on... but given the way we use structs and of course given all the C++ influx, this really doesn't make sense anymore...

dirkhh (approved):
I guess there isn't really anything outstanding in this one - let's merge it.

Issue #1844, opened 2018-11-04
Adding dive media files for multiple dives - conceptional issue


sfuchs79:
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [ ] New feature request
- [x] Discussion request

### Issue long description:

When adding my photos and videos for my last dive trip I found a kind of conceptional issue with our current add dive media function. Example scenario:
- 3 dives A, B and C
- 3 sets of media files I, II and III
- Time stamps of media file set I fits dive time of dive A +-30min
- Time stamps of media file set II fits dive time of dive B +-30min
- Time stamps of media file set III fits dive time of dive C but there are files outside the +-30min range for this dive
- I select the dives A, B and C
- I do "Load media files from disk" for these dives
- I select all files (set I, II and III)
- I receive the warning that some files are outside the +-30min range
- I select "Load media files even if the time does not match the dive time"
- Result: For each of the dives A, B and C all the files from set I plus II plus III are added.

Not the desired result but also not unexpected. We have no logic to find out to which dive the files outside the +-30min range belong to and also the files from the other dives are then treated in the same way as files outside the range.
From my point of view it's not so simple to find a "perfect" solution for this.
There is one simple fix: Disable (gray out) the "Load media files even if the time does not match the dive time" as soon as multiple dives are selected. The user would need to add the files outside the +-30min range for every single dive. That sounds acceptable for me.

Any better ideas?

### Operating system:

All

### Subsurface version:

Current master

atdotde:
Sounds good to me.

bstoeger:
Yes - this has annoyed me for a very long time.

bstoeger:
> From my point of view it's not so simple to find a "perfect" solution for this.

Last time I thought about this my plan was to associate the media-file with the dive it's closest to. Of course a problem arises if we can't extract the meta-data properly. For this case, perhaps add a 24-h safeguard?

mturkia:
Usually when I add pics to dives on a dive trip, I just want the matching files to be associated with dives. I select all the dives from the trip, all the pictures during the time and want the pics that occurred during the dives to be matched. In this scenario I am just happy if the pics that are not directly associated with a dive are ignored. Anyway, if it is not possible to select multiple dives, the media handling becomes quite tedious task.

bstoeger:
@sfuchs79: If you find time, could you perhaps check if current master is fine for you? Thx.

bstoeger:
[...]

PR #1886, opened 2018-12-11
Dive media: add media to closest dive

bstoeger:
Currently, when selecting "Load media files even if time does not
match the dive time", the media are added to *all* selected dives.
Instead add it to the closest dive.

This seems like the less surprising behavior. Of course now if the
user really wants to add a media file to multiple dives, they will
have to do it manually.

To avoid a messy interface, this is solved by moving the iterate-
over-selected-dives loop to the core. Thus, a helper-function can
be made local to its translation unit.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is my idea for #1844: Add media not to *all* selected dives, but to the nearest. This seems to be the more common use case. But now users wanting to add the same file to multiple dives have more work.

Only lightly tested. This is mostly a RFC.

Since functionality moved to the core, one might think about enhancing the add-picture tests: Have multiple dives and multiple pictures and make sure that they are attributed correctly.
### Changes made:

1) Upload image only to closest selected dive.

### Related issues:

#1844
### Release note:

Yes - but let's discuss this first.

### Documentation change:

Perhaps - but let's discuss this first.

### Mentions:

@sfuchs79

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -4004,20 +4004,51 @@ static bool new_picture_for_dive(struct dive *d, const char *filename)
 	return true;
 }
 
+/* Return distance of timestamp to time of dive. Result is always positive, 0 means during dive. */
+static timestamp_t time_from_dive(const struct dive *d, timestamp_t timestamp)
+{
+	timestamp_t end_time = dive_endtime(d);
+	if (timestamp < d->when)
+		return d->when - timestamp;
+	else if (timestamp > end_time)
+		return timestamp - end_time;
+	else
+		return 0;
+}
+
 // only add pictures that have timestamps between 30 minutes before the dive and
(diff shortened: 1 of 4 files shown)


neolit123 (approved):
the change simplifies the code, so LGTM.

i guess folks should comment if this is the desired UX.
"media is added for closest dives" seems good from my POV.

dirkhh:
I think that's the correct semantic

dirkhh (approved):

atdotde:
Looks fine

atdotde (approved):
Much better than what we had before.

Issue #1854, opened 2018-11-13
Average depth in yearly statistics


plxflx:
### Describe the issue:

- [ ] Bug
- [x] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Average depth in Yearly statistics tend to get very low when you do a lot of deco diving. On any given deco dive an average depth tends to be between 23-25m because of the time spent on shallow depth in the end of the dive. I want average depth to be based on maximum depth. Average depth for two given dives should end up some where between the two dives maximum depth.
![skarmbild fran 2018-11-13 22-57-59](https://user-images.githubusercontent.com/43491057/48446344-4b0dac80-e799-11e8-8f43-d5955586ac44.png)
In this example the average depth for month 8 is lower then minimum depth. I think it would make more sense to have an average between minimum and maximum. 

### Subsurface version:

4.8.3

### Current behavior:

Average depth is the average of average depth for the given number of dives.

### Expected behavior:

Average depth is to be the average of maximum depth for the given number of dives.

PR #1890, opened 2018-12-16
Average max depth

atdotde:

- [ ] Bug fix
- [x] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When doing statistics over more than one dive, the average of the max depths of the dives makes an interesting datum. So display it in the statistics tab as well as the yearly statistics. We removed the average depth of all selected dives from the statistics tab (as this "average" was often confusing and tended to be below the "min" value) but I guess this is much better.

### Related issues:

Implements suggestion in #1854


bstoeger (approved):
Compiled and quickly tested - lgtm.
Minor point: Perhaps rename "Depth" to "Max. depth" in stats tab?

atdotde:
Good idea @bstoeger, updated accordingly.

bstoeger:
Seems simple enough, implements a user-requested feature and shouldn't hurt anyone. Let's commit.

PR #1897, opened 2018-12-23
Improve error reporting when exporting a selection of 0 dives.

jakkarth:
If no dives are selected when trying to export a selection, a message
is shown that no dives were selected, but it's immediately hidden
behind a message saying that a temporary file could not be created.
In fact, the creation of the temporary file wasn't never attempted,
so the message that the user actually sees is misleading.

The solution chosen here is to duplicate the check that at least some
dives are selected, and abort early if that case is detected, rather
than continuing on to show the additional misleading message. Not
elegant, but it gets the job done.

Better solutions to this include refactoring prepare_dives_for_divelogs
to return something more descriptive than a bool, remove that check
from prepare_dives_for_divelogs entirely since it doesn't seem to be
a good fit there, or switch to exceptions for handling these problems
rather than return values. I don't have sufficient familiarity with
the codebase to attempt these more invasive changes, but they
should be considered in the future.

On a final note, some of the other error messages in this file start
with a capital letter, but the one relevant to this particular PR
does not. Again, I'm not familiar enough with the codebase (or
translations) to know if that's safe to change, so I'll leave that
for another time or another developer.

Reported-by: John Plaxco
Signed-off-by: John Plaxco <john@johnplaxco.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/desktop-widgets/subsurfacewebservices.cpp
+++ b/desktop-widgets/subsurfacewebservices.cpp
@@ -484,6 +484,10 @@ void DivelogsDeWebServices::prepareDivesForUpload(bool selected)
 {
 	/* generate a random filename and create/open that file with zip_open */
 	QString filename = QDir::tempPath() + "/import-" + QString::number(qrand() % 99999999) + ".dld";
+	if (!amount_selected) {
+		report_error(tr("no dives were selected").toUtf8());
+		return;
+	}
 	if (prepare_dives_for_divelogs(filename, selected)) {
 		QFile f(filename);
 		if (f.open(QIODevice::ReadOnly)) {


dirkhh:
Hi @jakkarth and welcome as a Subsurface developer.
Thanks for the contribution - much appreciated.
As you can tell from the code base, we usually request a "Signed-off-by" line that includes the developer's email address. Is that something you'd be willing to do?

The code change itself looks reasonable. You are making good points in the text above, though. The error message _should_ start with a capital letter (that will have implications for the translations, but those are easily caught before release). And then, of course, the bigger question of whether the code overall should be refactored to be more logical - I'm certainly open to proposals.

If you are able to fix the Signed-off-by line, I'd take the PR as provided and would encourage you to make other changes in a follow up PR.

Thanks

/D

jakkarth:
I'm happy to update the line to be my email address instead of my name if that's what's required. Reading through https://subsurface-divelog.org/documentation/contributing/ and https://github.com/Subsurface-divelog/subsurface/blob/master/CONTRIBUTING.md, it seemed pretty clear that "Your Name" meant I should put my name there, not my email address. Other commits in the log show Name \<email\>, so I'll adopt that convention, but perhaps a doc update would be in order to clarify that for future contributors :)

Issue #1891, opened 2018-12-19
Linux desktop: BT and BLE weirdness


janmulder:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [x] Discussion request

### Issue long description:
I see miscellaneous issues related to BT/BLE. I group them in this issue, just to document them now, and when splitting off separate issues due to unrelated technical parts is felt needed: that is perfectly fine.

1) Given my infrastructure, I cannot download from my dual stack OSTC3 aka Plus over BT. This seems related to QTBUG-70536. BLE does work correctly (after a re-pair session). I see a related error in the console: _Connecting to port is not supported (Uuid required)_  and opening the device fails.

2) When playing with BT/BLE downloads I found that UI settings like "force classical" or "force BLE" do not really do what they suggest. I expect that force means: no never attempt the other way of interfacing. Its not like that.

3) What confuses things further is that the choice made as "Bluetooth mode" (auto, force BT, force BLE) in the remote Bluetooth device selection is not preserved between sessions. It always defaults to "auto" on startup. I do not suggest to put this in the preferences, but the relation with the shortcut DC section buttons, and the LE: addressing for BLE is at least confusing. Without looking to logdata it's simpy unclear how the DC data came in (BT vs BLE). 

Obviously, point 2) and 3) are most prominent on dual stack DCs.

### Operating system:
What OS are you running, including OS version and the language that you are running in: up to date Arch Linux desktop. For the people that do not know Arch, things like kernel, bluez version, Qt version are all pretty new. For example: Qt 5.12.0, bluez 5.5, kernel 4.19.8

### Subsurface version:
What version of Subsurface are you running? master
Does the same happen on another Subsurface version? unsure (no bisects done)
Are you using official release, test build, or compiled yourself?: self
Provide Git hash if your are building Subsurface yourself: bfe69239df6a

janmulder:
Let me elaborate a little more on point 1) above (while Qt 5.12.0 is compiling from source in the background).

[...]

PR #1898, opened 2018-12-23
Core: POC fix BT on Linux, workaround possible Qt bug on 5.12

janmulder:

- [x] Bug fix (proof of concept)
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

After upgrading to Qt 5.12.0, download over BT from a DC did not work
any more. On the console the message "Connecting to port is not
supported (Uuid required)". Linus noticed earlier that we do rather
strange processing in this part of the code related to selecting port 1
or port 5. This all seems not needed (any more), but broader testing is
advised. This being stripped from the code, the mentioned error from Qt
persisted. That is strange in itself, as we did not reference port
numbers any more (but we do no service discovery, possibly confusing 
Qt's internals).

Step 2 in this commit is actually using an uuid to the call to
connectToService. Choosing an uuid seems relatively straightforward as
we can use the same one we already use for Android. That is the default
BT RFCOMM Serial Port Profile uuid. Interestingly, when changing to this
uuid we run immediately in a Qt runtime error telling us "QDBusPendingReply:
type ManagedObjectList is not registered with QtDBus.". For these 2
unexpected Qt messages, QTBUG-72742 was made. Studying the Qt source
code at this point reveals a possible workaround. Simply create a
QBluetoothLocalDevice object, which, behind the scenes registers the Qt
internal ManagedObjectList with QtDBus. This can definitely be
considered a hack, and that is why this commit is marked proof of
concept.

In the end, interfacing over BT works again.

### Changes made:
See commit.

### Related issues:
#1891 (bullet 1)

### Additional information:
QTBUG-72742 is made in relation to this PR

### Release note:
No.

### Mentions:
@ablasche

### DIFF
--- a/core/qtserialbluetooth.cpp
+++ b/core/qtserialbluetooth.cpp
@@ -3,6 +3,7 @@
 
 #include <QtBluetooth/QBluetoothAddress>
 #include <QtBluetooth/QBluetoothSocket>
+#include <QBluetoothLocalDevice>
 #include <QEventLoop>
 #include <QTimer>
 #include <QDebug>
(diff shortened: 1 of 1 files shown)


dirkhh:
I read through this a few times; it's so easy to get confused with the #if clauses that were removed.
Based on my understanding what's going on I **_think_** this is right... and we certainly no longer need to worry about Qt versions older than 5.6 (and I'm willing to slowly creep up that minimum).

I'd love to hear from @ablasche TBH

janmulder:
And QTBUG-72742 is now a P1:Critical  for Qt. So, I'm pretty sure it will be fixed.

ablasche (commented):
The patch looks good. As a minor optimization you could create the QBluetoothLocalDevice instance on the stack rather than on the heap.

Once I have patched Qt (probably 5.12.2), you can drop the instance creation altogether as I'll add the registration to the internals of QBluetoothSocket::connectToService()

janmulder:
Fixed the heap vs stack issue Alex correctly noticed, and reworded the commit message slightly to reflect the fact that Qt will solve this.

As 5.12.2 is still some time away, I suggest to merge this workaround.

janmulder:
The Android 5.12 branch. As the apk does not run... maybe sort that out first?

janmulder:
@bstoeger: agree but this is perfectly fine for the thing we want to achieve here. Just trapping a call to an internal function to bypass the Qt error

janmulder:
The fix of QTBUG-72742 will land in Qt 5.12.1

PR #1904, opened 2018-12-23
Make dive-import undoable

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This makes dive-import undo-able. Yes, this is horribly complex and the history is a mess. I tested a few use-cases (merging import - not downloading).

One reason for the complexity was the fact that dive-trips were always added to a global table. That had to be split out. Actually, I have a feeling that the whole table is unnecessary, so perhaps we can remove all that. The only place where the table is used is for saving (xml and git) and there it could be generated on the fly.

The other reason is the handling of trips on merging. This did not work reliably at all. The new code therefore does not merge single dives but whole trips. This is a distinct functional change. Time will tell if this works as expected or if we'll have to give the user the option merge overlapping trips / don't merge overlapping trips. The behavior implemented here is: merge trips on import, don't merge on download (to respect the "make new trip" option).

### Changes made:

1) Merge trip-wise on import
2) Make import undoable
3) Minor cleanups

### Release note:

Yes.
### Documentation change:

Not sure? Let's wait for a more complete undo-system?


dirkhh (approved):
That's a rather, err, massive PR. I read through it a couple of times and I tried it. It seems to work.
I'd love a review from @torvalds as this touches some very old code...

bstoeger:
Thanks - I'll admit that the history is garbage. 8193f5dddf3a50f43bc9f353b5d436822e53e755 adds some very subtle code that is replaced later.

dirkhh:
Linus checked this and was ok with merging it.

PR #1906, opened 2018-12-27
Dive list: split list and tree-models

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Currently, we have one model that does both - tree and list view. This PR splits them in two distinct models that derive from a common base class. The idea is that - in the distant future - perhaps we can use the list-model for desktop and mobile and then can use the undo-system also for mobile.

This makes for more lines of code as there is now some code duplication. Also it has a somewhat strange behavior which makes that a model reset leads to a new instance returned by the `instance()` function - so this is not a classical "singleton" anymore. As an alternative, perhaps always keep both models up-to-date and just switch the source model in the filter-model?

### Changes made:

1) Split list and tree-versions of the DiveTripModel.
2) Straighten control flow: DiveListView -> MultiFilterSortModel -> DiveTripModel.
### Release note:

No - purely internal.
### Documentation change:

No - purely internal.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -28,20 +28,20 @@
 #include "desktop-widgets/simplewidgets.h"
 
 DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelection(false),
-	currentLayout(DiveTripModel::TREE), dontEmitDiveChangedSignal(false), selectionSaved(false),
-	initialColumnWidths(DiveTripModel::COLUMNS, 50)	// Set up with default length 50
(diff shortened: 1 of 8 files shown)


dirkhh (approved):
This looks sane to me - but this is the type of code where I always feel a bit uncomfortable.
I guess at this point @bstoeger is our expert on that code... I requested a review from @tcanabrava but I don't know how much time Tomaz has these days... so if he doesn't find time on Friday I'll be ok with merging it.

PR #1907, opened 2018-12-28
Mobile: fix editing dive site (Quickfix for #1887)

### LINKED ISSUE #1887: Editing location on mobile does not work
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When editing a dive on subsurface-mobile, the location is not saved.

### Operating system:

subsurface-mobile on desktop, probably the native iOS and Android builds as well, but haven't tested

### Subsurface version:

master

### Steps to reproduce:

- Open divelist.
- Choose a dive and edit the location
- Save
- Check if the location was changed on divelist and dive details

### Current behavior:

The location sticks to the original one on both dive list and details

### Expected behavior:

Location is updated to the new value

bstoeger:
Commit 68961a169efc37039cd3fda334efb9ad9927444f made it impossible
to edit a dive site on mobile if a dive-site was already set: If
divesite was non-null, no actions were taken. Remove the conditional.

Reported-by: Miika Turkia <miika.turkia@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a quick-fix for #1887. Do not apply yet, as I'm unsure what happens when location is empty, but coordinates are set. Then `d->dive_site` is NULL, but what happens if the GPS coordinates are set? Crash? Will have to check later.

### Changes made:

1) Remove bogus test for `ds == NULL`.

### Related issues:

Perhaps fixes #1887.
### Release note:

First have a closer look at corner cases.
### Documentation change:

No
### Mentions:

@mturkia 
@dirkhh

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -887,11 +887,9 @@ bool QMLManager::checkLocation(DiveObjectHelper *myDive, struct dive *d, QString
 	qDebug() << "checkLocation" << location << "gps" << gps << "dive had" << myDive->location() << "gps" << myDive->gas();
 	if (myDive->location() != location) {
 		diveChanged = true;
-		if (!ds)
-			ds = get_dive_site_by_name(qPrintable(location));
-		if (!ds && !location.isEmpty()) {
+		ds = get_dive_site_by_name(qPrintable(location));
+		if (!ds && !location.isEmpty())
 			ds = create_dive_site(qPrintable(location), d->when);
-		}
 		d->dive_site = ds;
 	}
 	// now make sure that the GPS coordinates match - if the user changed the name but not


mturkia:
I only did a very quick test on simulator without any GPS coordinates and that worked. As you have done more testing, and I assume it covers the more interesting case you describe in the pull request, this PR should be good to merge.

dirkhh (commented):
So what happens to the dive site that was there? Does it become an orphan?
I guess we don't have a garbage collector for those and then simply ignore it at the end of execution?

bstoeger:
> So what happens to the dive site that was there? Does it become an orphan?

At first, it stays in the `dive_site_table`.

But it seems to me that both the xml and git writers have code to delete unused dive-sites. See functions `save_divesites()` and `save_dives_buffer()`.

bstoeger:
Uh... It just occurred to me that this is problematic:

1) delete dive with single-use dive site.
2) save. Dive site will be deleted.
3) undo. Boom, dive has pointer to freed dive site.

So we do need proper reference counting.

bstoeger:
Sure, that would be a quick-fix, but somehow not entirely satisfying. 😛

In the long run I'm also not sure if not saving is the right thing to do. What about a list of dive-sites with manual delete and an option to purge dive-sites with 0 dives?

dirkhh:
Right.
The famous dive site management feature.
Yes, we need to fix our handling of dive sites.
We have thousands of emails on that topic, but never came up with a design that we were happy with.
I don't think that's something we'll fix as a small side project...
So yes, by all means, let's tackle this. But please don't underestimate the complexity.
Of course I'm talking to the person fixing undo...

/D

PR #1915, opened 2019-01-01
Dive sites: don't delete unused dive sites on save

bstoeger:
Unused dive sites were deleted on save. This clashed with the undo
system in the following scenario:

1) Delete single-use dive site.
2) Save (dive site deleted)
3) Undo (reference to freed dive site)

Therefore, as a quick-fix, keep the referenced dive site around.
Note that this also means that empty dive sites must not be
deleted, as it might refer to a dive in the undo system. Instead
only clear references to empty dive sites in the global dive
table. Factor this functionality out, as it was common to the
XML and git savers.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes an easily reproduced crash:
1) Delete dive with single-use dive site
2) Save
3) Undo

The reason is that on save "unused" dive sites were deleted, but might still be references from the deleted dive.

### Changes made:

1) Don't delete unused dive sites on save.
2) Factor out common purge empty dive site code in the XML and git savers.
### Related issues:

#1907 
### Release note:

Perhaps?
### Documentation change:

No.

### DIFF
--- a/core/divesite.c
+++ b/core/divesite.c
@@ -346,6 +346,23 @@ struct dive_site *find_or_create_dive_site_with_name(const char *name, timestamp
 	return create_dive_site(name, divetime);
 }
 
+void purge_empty_dive_sites()
+{
+	int i, j;
+	struct dive *d;
+	struct dive_site *ds;
+
+	for (i = 0; i < dive_site_table.nr; i++) {
+		ds = get_dive_site(i);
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
Looks right to me

Issue #1926, opened 2019-01-04
Unable to connect & Download from Cloud Storage on RaspberryPi


keithgrimes:
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Compilation of software completed on a RaspberryPi (as no existing distribution available). Software starts up ok. I can download software from my dive computer but am unable to connect to cloud storage  due errors raised (see below). 

### Operating system:
Debian Stretch
RaspberryPi
Have tried twice (including on a clean installation). Same error is reported. 
I also run 4.8.4 on iOS with no problems (but this is not a self compiled version). 

### Subsurface version:
Version Built: 4.8.4
Does not happen on pre-built versions of 4.8.4 (I run the packages on iOS & OSX without problem). 
Self Compile
Git Commit Ref used: 0903bef9db8659313c4bb533c295505c0c3851bf

### Steps to reproduce:
Execute the software. 
Configure the Cloud Storage account in Preferences
Select File>Open Cloud Storage

### Current behavior:
On first execution the error message ‘Creating Commit Failed’
On second execution the error message ‘unable to lookup revision <email>’ is displayed

### Expected behavior:
Dive Log already stored in the cloud is downloaded and displayed.

### Additional information:
Build Log available if needed.

keithgrimes:
Forgot to mention I have reviewed other entries with similar messages and tried deleting the directory ~/.subsurface/cloudstorage but this does not resolve the issue.

mturkia:
Have you installed all the dependencies listed in INSTALL file?

We do provide Debian packages from our Launchpad repository that should be compatible with Stretch, thus there should not be real reason to compile yourself just to get the software running.

Instructions to use the Launchpad .deb are available in https://subsurface-divelog.org/download/

keithgrimes:
As stated attached are build files, logs and an updated CMakeLists.txt file required to ensure the build completes (as instructed on an article on the User Forum). 

Please note two of the dependencies conflict (I have tried compilation with both). The other does not exist.

[config.txt](https://github.com/Subsurface-divelog/subsurface/files/2729374/config.txt)
[build.log](https://github.com/Subsurface-divelog/subsurface/files/2729375/build.log)
[CMakeLists.txt](https://github.com/Subsurface-divelog/subsurface/files/2729376/CMakeLists.txt)

janmulder:
[...]

PR #1928, opened 2019-01-08
Build: build libgit2 from source when less than 0.26.0

janmulder:

- [x] Likely bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup/maintenance
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

We currently require a minimal version of libgit2 of 0.24.0. From
issue #1926 it seems that this version is too old. So, a simple test on
Linux to see the behavior with such an old libgit2, I tried that.
Interestingly, with the current version of openssl that old libgit2
version does not even compile from source (known error in libgit2).

So, bump our minimal version of libgit2 to 0.26.0. That is also the
version we currently use on the Travis and official builds, so well
tested.

### Changes made:
See commit (trivial)

### Related issues:
issue #1926 seems to show something related to an old libgit2

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -197,7 +197,7 @@ else
 		LIBGIT=$(pkg-config --modversion libgit2 | cut -d. -f2)
 		LIBGIT2_FROM_PKGCONFIG="-DLIBGIT2_FROM_PKGCONFIG=ON"
 	fi
-	if [[ "$LIBGIT" < "24" ]] ; then
+	if [[ "$LIBGIT" < "26" ]] ; then
 		# maybe there's a system version that's new enough?
 		LIBGIT=$(ldconfig -p | grep libgit2\\.so\\. | awk -F. '{ print $NF }')
 	fi
@@ -237,7 +237,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	fi
 fi
 
-if [[ "$LIBGIT" < "24" ]] ; then
+if [[ "$LIBGIT" < "26" ]] ; then
 	LIBGIT_ARGS=" -DLIBGIT2_INCLUDE_DIR=$INSTALL_ROOT/include -DLIBGIT2_LIBRARIES=$INSTALL_ROOT/lib/libgit2.$SH_LIB_EXT "
 
 	cd $SRC


atdotde:
Looks good

janmulder:
Well ... the older Travis Qt builds have a problem with this change. Qt56 fails in a valid way. Interestingly, qt55 succeeds, but care looking in the build log shows a problem (git checkout v0.26.0
error: Your local changes to the following files would be overwritten by checkout:). So, its feels that there is still something wrong here. In addition, I did not inspect any of the the used docker images used to build those old versions. That qt55 case is weird.

janmulder:
After a little code reading in the Travis Qt5x series, I think its time for @dirkhh to take a look. All those build are with different distributions or OS versions (ubuntu, fedora, opensuse). I think this is intentional to also verify building for those distributions. But it feels a bit of a double-edged sword to me. I always (incorrectly) assumed that the Qt5x Travis series are only different with respect to qt version, but its way more complex than that (maybe rename them to reflect its real purpose?).

dirkhh:
I'm traveling today. Will take a look this evening.
It's in a way less complicated than you think.
I was simply looking for distros with those Qt binaries available. So this is indeed primarily about Qt versions. The rest is side effects, but worth looking into.

dirkhh:
Apparently OpenSUSE switched to libressl and even though libgit2 0.26 should include the build fix for that, it seems to fail. not exactly sure why.
```
$ git tag --contains 42ad85ef01e372f1d65b9c
v0.26.0
```
seems to indicate that this should be addressed. But maybe there are more fixes needed? Is there a reason you didn't switch to the last 0.26 tag (0.26.8)?

dirkhh:
nope, that didn't do the trick. I need to download the image locally and play with this some more...  it has got to be possible to compile a newer libgit2 on OpenSUSE 42.3

janmulder:
> Is there a reason you didn't switch to the last 0.26 tag (0.26.8)?

Well ... not really. 

Reading the scripting it looks looks its supposed to work like this:
- check what version the OS does offer, but look at our install-root first
- when older than (was 0.24.0, this commit only changes that to 0.26.0) install from source
- pull the source version that we specified in the get-dep-lib.sh script (which is 0.26.0 currently) and use that

The main reason I did not touch the version in get-dep-lib.sh is that I like the relation between the check (< 26) and the used version, as the old way looks strange (when < 24 use 26), and as we currently seem to use 0.26.0 for production builds as well.

janmulder:
Played around on the test branch and managed to get Qt56 compiled using openssl and not libressl. 

There is also an issue on the Qt55 Travis build that still needs to be addressed ...

janmulder:
Ok, the issue with the Qt55 build is some git magic. Apparently git on Ubuntu is configured differently with respect to crlf conversions, and the ligbit2 repo has a *text=auto in its .gitattributes, causing a dirty working directory on clone, in which we cannot reset the branch to the one we like. Lacking time right now to find a fix.

dirkhh:
> Ok, the issue with the Qt55
I don't see a build failure with Qt55... 4923.8 passed https://travis-ci.org/Subsurface-divelog/subsurface/jobs/476749202

janmulder:
Look in the log right at the top. Not sure how long that failure is already there, as nobody looks at logs when its marked green.

dirkhh:
Fun. We should make sure that the build actually FAILS when something like this happens.
Let's figure out how to teach git to do the right thing. I have a few minutes right now and will look into that.

Issue #1925, opened 2019-01-03
Desktop Crash When Importing Dives to New Trip From Descent


UncleConker:
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Desktop application crashes when importing dives into a new trip from the import dialog box but not when importing without a new trip.

### Operating system:
Microsoft Surface Pro 3
Operating system: Windows 10 (10.0)
CPU architecture: i386
OS CPU architecture: x86_64
Language: en-US

### Subsurface version:
Version: Subsurface 4.8.4.0
Does the same happen on another Subsurface version?: Unsure
Are you using official release, test build, or compiled yourself?: Official

### Steps to reproduce:
1) Hover over Import tab
2) Click Import from Dive Computer
3) Select "Garmin" and "Descent Mk 1"
4) Enter appropriate USB port
5) Tick "Download into new trip" option
6) Click "Download"
7) Select dives you wish to import
8) Hit "OK"

After a brief load and spinning cursor the application will crash.

### Current behavior:
Application crashes when downloading dives into a new trip but not when downloading without creating a new trip.

### Expected behavior:
Dives download successfully and create a new trip containing the new dives.

### Additional information:
**When I went to run the dump log the import was successful and created a new trip as expected the first time.**
[Log File 1](https://drive.google.com/file/d/1tpbxmFCW3lYjkrfIxGefShvz-Ik7x-CV/view?usp=sharing)

**The second time I ran a Log file the application crashed as well**
[Log File 2](https://drive.google.com/file/d/1tpbxmFCW3lYjkrfIxGefShvz-Ik7x-CV/view?usp=sharing)

Dump File: When I try to create a dump file I get a  "Dive Import Error"

dirkhh:
A lot of that code has been massively rewritten, lately. Would you be willing to test a current development version to see if you can still see the problem there?
You usually can find our latest binary here: 
https://github.com/Subsurface-divelog/subsurface/releases/tag/continuous 
just grab the subsurface-4.8.4-*.exe (as I write this it's https://github.com/Subsurface-divelog/subsurface/releases/download/continuous/subsurface-4.8.4-126-g0903bef9db86.exe but that will change once we merge the next pull request)

janmulder:
[...]

PR #1930, opened 2019-01-10
Import: fix download into new trip

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The recent undo-of-import changes were buggy when importing into a new trip:
1) The trip table was not passed correctly.
2) The process_imported_dives() logic was faulty.

Fix both. Add a unused-variable fix for good measure.

### Changes made:

1) Pass correct trip table to downloader.
2) Fix silly logic mistake in merge-trips loop.

### Related issues:

#1925 
#1904 

### Release note:

No.
### Documentation change:

No.
### Mentions:

@dirkhh @janmulder


dirkhh (approved):
LGTM

bstoeger:
Factored out the merge-trip-into-existing-trips functionality to make the loop a bit more palatable.

But: I believe that the way we do this is wrong anyway, from a UI perspective. What happens is that "download into new trip" is performed during download. So consider the following situation:

1) User selects "download into new trip".
2) Downloads dives. Dives appear in the dialog and are added to a new trip.
3) Deselects "download into new trip".
4) Clicks "OK".

The *old* version of the "download into new trips" flag will be used, because all that was done during download time. Therefore I think that this logic should be moved from the download-code to the import-code, isn't it?

dirkhh:
Linus doesn't do GitHub... I'll ask him to chime in. This was his code... :-)

torvalds:
I think @bstoeger is right, it probably makes more sense to do as a import-time thing.

The only important part is that because the new dives are in a new trip, they shouldn't then be merged with any pre-existing dives. But yes, the "should trip be assigned" makes more sense at import time than download time.

dirkhh:
@bstoeger your call if you want to merge as is or make it "more correct".
Linus and I are in Fiji with poor internet...

bstoeger:
> Linus doesn't do GitHub... I'll ask him to chime in. This was his code... :-)

The garbage logic was mine.

I pushed the "move new-trip logic to import function" part on top. Now history is weird because the first fix adds a trip-table to the libdivecomputer state, which is then promptly removed.

Please don't commit yet - the code doesn't yet ensure to never merge dives if add-to-new-trip is set.

> Linus and I are in Fiji with poor internet...

That sounds like a cruel fate. :) Fiji is where I did my first dive ... spoiled forever. If only it weren't so hard to reach from Europe!

bstoeger:
From my side this is ready (though additional testing can't hurt).

With create-new-trip and download-all-dives I was able to download the same dive into two trips. I guess that's how it's meant to be...

dirkhh (approved):
Reading through the code this looks very nice.
I appreciate all the cleanups that you do, @bstoeger

PR #1889, opened 2018-12-16
Filter: connect widget and filter

bstoeger:
In the latest version of the new filter-widget the connection
between widget and filter was lost. Connect both - but use a
simple function call instead of a signal, since it is not
immediately obivous were the connection should be made.

Signed-off-by: Berthold Stoeger <bstoeger@mail.tuwien.ac.at>

### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This fixes the new filter-widget by connecting it to the filter. This is a temporary fix until @tcanabrava finds time to do it like it was intended. It is supposed to let people play with the new widget. I replaced the signal by a simple function as I was too lazy to think about where the `connect` call should be placed [I always find it weird when an object connects is own signal - unless it is supposed to be a call across thread boundaries.]
### Changes made:

1) Call filter from filter-widget.
### Mentions:

@janmulder: This should make it possible to test.

PR #1938, opened 2019-01-17
Filter: connect filter and widget

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a reopening of #1889. But I can't reopen it, because I force-pushed changes [reopen-then-push works, but push-then-reopen doesn't?].

Sorry for bull-in-a-china-shop mode - at least this allows people to test the widget.

@janmulder: I think I know why #1889 crashes for you on exit - please confirm whether the new commit fixes this. [Note: The fact that the destructor of the QObject base class calls hide() on sub-objects (according Qt's object hierarchy, not the C++ object hierarchy) lets me seriously question their design decisions. The base-object is now in a semi-destructed state and it seems very logical for sub-objects to call into the base-object. Very annoying.]

### Changes made:

1) Update filter if user enters data.

### Release note:

Not for this PR, but the original PR.

### Documentation change:

Not for this PR, but the original PR.

### DIFF
--- a/qt-models/filtermodels.cpp
+++ b/qt-models/filtermodels.cpp
@@ -111,10 +111,12 @@ bool MultiFilterSortModel::showDive(const struct dive *d) const
 
 	// TODO: get the preferences for the imperial vs metric data.
 	// ignore the check if it doesn't makes sense.
-	if (d->watertemp.mkelvin < C_to_mkelvin(filterData.minWaterTemp) || d->watertemp.mkelvin > C_to_mkelvin((filterData.maxWaterTemp)))
+	if (d->watertemp.mkelvin &&
+	    (d->watertemp.mkelvin < C_to_mkelvin(filterData.minWaterTemp) || d->watertemp.mkelvin > C_to_mkelvin((filterData.maxWaterTemp))))
 		return false;
 
-	if (d->airtemp.mkelvin < C_to_mkelvin(filterData.minAirTemp) || d->airtemp.mkelvin > C_to_mkelvin(filterData.maxAirTemp))
+	if (d->airtemp.mkelvin &&
+	    (d->airtemp.mkelvin < C_to_mkelvin(filterData.minAirTemp) || d->airtemp.mkelvin > C_to_mkelvin(filterData.maxAirTemp)))
 		return false;
 
 	if (filterData.from.isValid() && d->when < filterData.from.toTime_t())
--- a/qt-models/filtermodels.cpp
+++ b/qt-models/filtermodels.cpp
@@ -16,57 +16,48 @@
 #include <algorithm>
 
 namespace {
-	bool hasTag(const QStringList tags, const struct dive *d)
+	// Check if a string-list contains at least one string containing the second argument.
+	// Comparison is non case sensitive and removes white space.
+	bool listContainsSuperstring(const QStringList &list, const QString &s)
 	{
-		if (!tags.isEmpty()) {
-			auto dive_tags = get_taglist_string(d->tag_list).split(",");
-			bool found_tag = false;
-			for (const auto& filter_tag : tags)
-				for (const auto& dive_tag : dive_tags)
-					if (dive_tag.trimmed().toUpper().contains(filter_tag.trimmed().toUpper()))
-						found_tag = true;
-
-			return found_tag;
-		}
-		return true;
+		return std::any_of(list.begin(), list.end(), [&s](const QString &s2)
+				   { return s2.trimmed().contains(s.trimmed(), Qt::CaseInsensitive); });
 	}
 
-	bool hasPerson(const QStringList people, const struct dive *d)
+	bool hasTags(const QStringList &tags, const struct dive *d)
 	{
-		if (!people.isEmpty()) {
-			QStringList dive_people = QString(d->buddy).split(",", QString::SkipEmptyParts)
-				+ QString(d->divemaster).split(",", QString::SkipEmptyParts);
+		if (tags.isEmpty())
+			return true;
+		QStringList dive_tags = get_taglist_string(d->tag_list).split(",");
 
-			bool found_person = false;
-			for(const auto& filter_person : people)
-				for(const auto& dive_person : dive_people)
-					if (dive_person.trimmed().toUpper().contains(filter_person.trimmed().toUpper()))
-						found_person = true;
-
-			return found_person;
-		}
-		return true;
+		return std::all_of(tags.begin(), tags.end(), [&dive_tags](const QString &tag)
+				   { return listContainsSuperstring(dive_tags, tag); });
 	}
 
-	bool hasLocation(const QStringList locations, const struct dive *d)
+	bool hasPersons(const QStringList &people, const struct dive *d)
 	{
-		if (!locations.isEmpty()) {
-			QStringList diveLocations;
-			if (d->divetrip)
-				diveLocations.push_back(QString(d->divetrip->location));
+		if (people.isEmpty())
+			return true;
+		QStringList dive_people = QString(d->buddy).split(",", QString::SkipEmptyParts)
+			+ QString(d->divemaster).split(",", QString::SkipEmptyParts);
 
-			if (d->dive_site)
-				diveLocations.push_back(QString(d->dive_site->name));
+		return std::all_of(people.begin(), people.end(), [&dive_people](const QString &person)
+				   { return listContainsSuperstring(dive_people, person); });
+	}
 
-			bool found_location = false;
-			for (const auto& filter_location : locations)
-				for (const auto& dive_location : diveLocations)
-					if (dive_location.trimmed().toUpper().contains(filter_location.trimmed().toUpper()))
-						found_location = true;
+	bool hasLocations(const QStringList &locations, const struct dive *d)
+	{
+		if (locations.isEmpty())
+			return true;
+		QStringList diveLocations;
+		if (d->divetrip)
+			diveLocations.push_back(QString(d->divetrip->location));
 
-			return found_location;
-		}
-		return true;
+		if (d->dive_site)
+			diveLocations.push_back(QString(d->dive_site->name));
+
+		return std::all_of(locations.begin(), locations.end(), [&diveLocations](const QString &location)
+				   { return listContainsSuperstring(diveLocations, location); });
 	}
 
 	// TODO: Finish this iimplementation.
@@ -130,15 +121,15 @@ bool MultiFilterSortModel::showDive(const struct dive *d) const
 		return false;
 
 	// tags.
-	if (!hasTag(filterData.tags, d))
+	if (!hasTags(filterData.tags, d))
 		return false;
 
 	// people
-	if (!hasPerson(filterData.people, d))
+	if (!hasPersons(filterData.people, d))
 		return false;
 
 	// Location
-	if (!hasLocation(filterData.location, d))
+	if (!hasLocations(filterData.location, d))
 		return false;
 
 	if (!hasEquipment(filterData.equipment, d))
--- a/qt-models/filtermodels.cpp
+++ b/qt-models/filtermodels.cpp
@@ -100,6 +100,10 @@ void MultiFilterSortModel::resetModel(DiveTripModelBase::Layout layout)
 
 bool MultiFilterSortModel::showDive(const struct dive *d) const
 {
+	// If curr_dive_site is set, we are in a special dive-site editing mode.
+	if (curr_dive_site)
+		return d->dive_site == curr_dive_site;
+
 	if (!filterData.validFilter)
 		return true;
 
--- a/qt-models/filtermodels.cpp
+++ b/qt-models/filtermodels.cpp
@@ -228,7 +228,7 @@ bool MultiFilterSortModel::lessThan(const QModelIndex &i1, const QModelIndex &i2
 	return DiveTripModelBase::instance()->lessThan(i1, i2);
 }
 
-void MultiFilterSortModel::filterDataChanged(const FilterData& data)
+void MultiFilterSortModel::filterDataChanged(const FilterData &data)
 {
 	filterData = data;
 	myInvalidate();
--- a/desktop-widgets/filterwidget2.cpp
+++ b/desktop-widgets/filterwidget2.cpp
@@ -3,112 +3,108 @@
 
 #include <QDoubleSpinBox>
 
-FilterWidget2::FilterWidget2(QWidget* parent)
-: QWidget(parent)
-, ui(new Ui::FilterWidget2())
+FilterWidget2::FilterWidget2(QWidget* parent) : QWidget(parent)
 {
-	ui->setupUi(this);
+	ui.setupUi(this);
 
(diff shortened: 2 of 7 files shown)


janmulder:
Let me copy/paste the remarks I made on December 17th:

1. the filter UI does not look "nice". Especially, the 2 items on the top (visibility and rating) are badly aligned.
2. Open app, ctrl-f to open the filter pane, close app => crash on exit from the application
3. Tried to filter on a tag. Ok, something is filtered, but the end result is a far to small set of dives. I have at least 400 dives with that tag, and only about 20 are selected by the entered tag.
4. Now empty the entered tag. Expected the full divelist to be shown again, but only a few extra dives are added again.

I can confirm that the crash I mentioned under 2) is fixed now. 

Needles to say that points 1), 3), 4) are unchanged.

bstoeger:
Let me add to that:

5. On dive-site edit filtering does not work anymore (it used to display only dives for the currently edited dive-site).
6. Negation not functional anymore.

These all feel like show-stoppers to me, but before fixing them, we have to make the interface respond. The question is: wait for @tcanabrava's as-intended version or go with my butchered version?

Finally, let me add a (highly synthetical) situation where the old filter-widget was superior in my opinion. I can't judge if it is relevant though as I don't even have enough dives to make the filter-widget necessary:

You have two buddies "Jack" and "Jackie". You want to filter all dives without "Jack" - that will also remove all dives with "Jackie". In the old widget entering "Jack" would list both and you could select one (or both). Of course this also works in the other direction: filtering for "Jack" gives additional dives for "Jackie", but showing too many dives seems less of a problem to me than hiding dives. To solve this we would need another option: "Match only full words", where one then has to wonder whether the new widget is really that much simpler...

bstoeger:
> Needles to say that points 1), 3), 4) are unchanged.

Just to be sure: Can you check that current master is not affected by 3) and 4)? I *did* work on the filter-code recently (in the course of dive-site uuid removal), so maybe that's not even related to the new widget...

janmulder:
> > Needles to say that points 1), 3), 4) are unchanged.
> 
> Just to be sure: Can you check that current master is not affected by 3) and 4)? I _did_ work on the filter-code recently (in the course of dive-site uuid removal), so maybe that's not even related to the new widget...

Yes, verified this PR on top of current master, and 3) and 4) (and 1) are still present.

bstoeger:
Jeez - I'm stupid. I forgot that the widget is already in current master. What I *actually* wanted to ask is if you are positive that  #1857 introduced the problem and that it didn't exist before. But instead of compiling an old version, could you perhaps produce a minimal example? It just doesn't happen for my humble set of dives.

janmulder:
Ok, checked out 19f6e8810cdaedcbec (the commit just before merging the new Tomaz stuff). Ok, the UI is fully different, but selecting a tag results in the proper set of dives, de-selecting again results again in the proper set.

Notice that I say "tag", but just found out that filtering on a temperature range does not work either. And I inclined to say that none of the filter options work as intended.

janmulder:
This 2 dive "logbook" shows, for me, the filtering does not work. Try: max temp = 9.00, or tag = cave. In both cases, 1 dive should be shown. I see 0 (that does only returns to the ldive list on exit of the filter). 

[test-filter.ssrf.txt](https://github.com/Subsurface-divelog/subsurface/files/2773038/test-filter.ssrf.txt)

janmulder:
Ok, the first issue is that it does not deal correctly for empty air or water temperatures.

bstoeger:
Thanks - that should be easy to fix and independent of this PR.

bstoeger:
> Ok, the first issue is that it does not deal correctly for empty air or water temperatures.

Pushed a fix to this problem.

@dirkhh: you might want to take the last commit (32c4b474853302f525b385710210bafd10191a6b) independently of the others, as it appears "obviously correct and fixes a bug".

bstoeger:
According to the code-comments, I think we should add another point:

7) No support for imperial units.

bstoeger:
Added fix for 5)
@dirkhh: I think you can also safely take that commit (88d09c7a7c4bc8b44498fbdaf69eace0f76801a0).

bstoeger:
Removed one alignment issue - for the rest I'm not going to fight with all these box/alignment things. :-P

janmulder:
Ok, that (trivial) temperature temperature commit solves (at least) my issues 3) and 4). 

And a new point:

8. filter on multiple tags or people. It now does OR between the tags (or people). I think we should do AND. That seems way more useful (not sure that the old filter did).

bstoeger:
> 8. filter on multiple tags or people. It now does OR between the tags (or people). I think we should do AND. That seems way more useful (not sure that the old filter did)

Added a commit to change from OR to AND. Replaced the loops by algorithms. Still not happy about the liberal use of `QStringList`.

I'm getting more and more weary about the substring comparisons. Is that *really* what we want? I would think we should at least do `startsWith`, not `contains`?

Moreover, I think we need a decision on how to proceed. I'd hate wasting my time on this if @tcanabrava  already has most of these fixes..

janmulder:
> Moreover, I think we need a decision on how to proceed. I'd hate wasting my time on this if @tcanabrava already most of these fixes..

Fully agree. So @dirkhh: what is your opinion on this?

tcanabrava:
no, please, by all means, merge this.
my time is unfortunately quite low so I appreaciate the help here.

dirkhh:
Linus and I are leaving Fiji today, so hopefully tomorrow I'll be back working on Subsurface. Having just tethering to my phone as internet and being out on the water 8+ hours a day slowed me down this past week 
a) we should work jointly on fixing the bugs in the widget. It's unfortunate that @tcannabrava doesn't have the time right now, but such is life.
b) whenever two or more of you agree to push commits or whole PRs into master, go ahead and do it, don't wait for me
c) I really hope we can get this widget into working order (or revert it) as we need to cut a new release fairly soon in order to have all the DCs supported

qt-models/filtermodels.cpp
```diff
@@ -16,57 +16,48 @@
...
-			for (const auto& filter_tag : tags)
-				for (const auto& dive_tag : dive_tags)
-					if (dive_tag.trimmed().toUpper().contains(filter_tag.trimmed().toUpper()))
-						found_tag = true;
-
-			return found_tag;
-		}
-		return true;
+		return std::any_of(list.begin(), list.end(), [&s](const QString &s2)
+				   { return s2.trimmed().contains(s.trimmed(), Qt::CaseInsensitive); });
```
dirkhh, line 24: ok, this is a use of lambda I approve of. Even I can read and understand this one. :-)

dirkhh (approved):
OK, broke out the computer and tethered again.
I think these changes all look fine, so I'm ok with merging. There is an Android build error with the SDK license not being accepted. that's weird as that worked before I left. I wonder what changed...
But I'm not going to mess with Gigabytes of containers over a tethered internet connection :-/
I'll merge with the build error and the goal to look at this once we're in New Zealand :-)

bstoeger:
> Ok, that (trivial) temperature temperature commit solves (at least) my issues 3) and 4).

Sadly, I think that it is wrong. Because now, any dives without recorded temperature will not be shown. But if you *do* filter for temperature I guess those dives should *not* be shown.

Therefore, instead of the default 0-99 range we need a "filter off" state. I experimented briefly with `setSpecialValueText()` , but the resulting UI experience was quite horrible. To turn off the filter, the user had to enter a special value (in my case -11). So we'd have to implement our own widget that overrides `interpretText()`.

Side note: the default min-max values 0-99 are no good. Ice-divers will experience air-temperatures well below 0°C and in Icelandic thermal pools you will have distinctly above 100°F.

janmulder:
> > Ok, that (trivial) temperature temperature commit solves (at least) my issues 3) and 4).
> 
> Sadly, I think that it is wrong. Because now, any dives without recorded temperature will not be shown. But if you _do_ filter for temperature I guess those dives should _not_ be shown.

Not fully sure what you are saying here. What I see now, this that filtering on a temp range also shows dives with empty temperature. That surprised me a little (and I think that this is what you mean). But I'm not sure that this is flat wrong. In my 1000+ dives logbook, I found 4 dives without water temperature, and those 4 dives are manually entered (because I forgot my DC, or the DC failed in some way during a dive, like empty battery).  So, no water temp seems rare (unless users are manually entering dives and do not bother to enter this data). In fact, I like this aspect to find "strange" logbook entries.

> 
> Therefore, instead of the default 0-99 range we need a "filter off" state. I experimented briefly with `setSpecialValueText()` , but the resulting UI experience was quite horrible. To turn off the filter, the user had to enter a special value (in my case -11). So we'd have to implement our own widget that overrides `interpretText()`.

As I said above, I'm not sure the new temperature filter is wrong.
> 
> Side note: the default min-max values 0-99 are no good. Ice-divers will experience air-temperatures well below 0°C and in Icelandic thermal pools you will have distinctly above 100°F.

And fully agree on this.

And as a side note: I'n currently trying to change the data/time range to the way we use date/time in the rest of the application. So, consistency related.

henrik242:
On Sun, Jan 20, 2019, 11:38 bstoeger <notifications@github.com wrote

> Side note: the default min-max values 0-99 are no good. Ice-divers will
> experience air-temperatures well below 0°C and in Icelandic thermal pools
> you will have distinctly above 100°F
>
Also I have a lot of (salt water) dives with *water* temperatures of -1 and
-2°C

Henrik

>

bstoeger:
> Not fully sure what you are saying here. What I see now, this that filtering on a temp range also shows dives with empty temperature. That surprised me a little (and I think that this is what you mean).

Yes, that's exactly what I meant. Sorry for being unclear. If this is fine for you, let's keep it for now.

I'm still weary of physically meaningless temperatures (e.g. the user misplaced the decimal separator). This will never be shown if the filter is active. Of course, these can be searched for by opening the filter and going in negation mode.

Same thing for time. Is using the current time really an appropriate default for latest-time?

janmulder:
> 
> Same thing for time. Is using the current time really an appropriate default for latest-time?

Valid question. I do not think that many users are really interested in time based filtering, so, for me, any default for the latest time default seems ok-ish. And then we have basically 2 options. Midnight 00:00, or current time. Current time seems slightly better, as a possible open filter and an import from DC on the current day does not hide the just imported dives (and 00:00 does).

bstoeger:
Speaking of negation mode ... Currently there is a (hidden) overall "invert filter" flag. I'm not sure that this is what we want.

1) According to de Morgan this turns a logical AND of multiple conditions into a logical OR of multiple negated conditions. So if a user filters on buddy X and tag Y, they now get all dives without buddy X *or* without tag Y. Does not make much sense, does it?
2) Inverting "min temperature" makes little sense as this turns it into "max temperature", which already exists. Same for "from" and "min rating", etc...
3) Inverting temperature- and time-*range* is at least semantically reasonable: Show me all dives not during this particular month or year. But is it really that useful?

In conclusion, I wonder if we shouldn't add individual "negate" buttons for selected fields (viz. Tags, People, Location). Luckilly, for these there seems to be enough space in the UI.

dirkhh:
Yes, I think it makes a lot more sense to have individual "negate" buttons.
I'd love it if you could keep working on getting the filters right.
I do have the problem of fairly urgently wanting to release a new version. I'll post about it on the mailing list to discuss options...

bstoeger:
I think we more-or-less addressed most of the points discussed in this PR, even if not 100% satisfyingly (e.g. the negate thing needs UI work). Therefore, with a bit of additional testing, perhaps it could even go into release? Depending on how pressing it is of course.

I'd just add another "regression", though not sure how important it is:

9) Missing reset-filter button.

bstoeger:
> Current time seems slightly better, as a possible open filter and an import from DC on the current day does not hide the just imported dives (and 00:00 does).

An alternative would be 23:59:59 for "to"?

janmulder:
> I think we more-or-less addressed most of the points discussed in this PR, even if not 100% satisfyingly (e.g. the negate thing needs UI work). Therefore, with a bit of additional testing, perhaps it could even go into release? Depending on how pressing it is of course.
> 
> I'd just add another "regression", though not sure how important it is:
> 
>     9. Missing reset-filter button.

And 10. Missing "filter shows x of y dives" . That is something I actually use for questions like: how many cave dives did I make with buddy A.



> > Current time seems slightly better, as a possible open filter and an import from DC on the current day does not hide the just imported dives (and 00:00 does).
> 
> An alternative would be 23:59:59 for "to"?

Yes, that is even better, or date + 1 day and time 00:00. That current time setting is actually "broken" at the moment. It's set at creation of the widget, and that is startup of the application. That can be just a little early.

tcanabrava:
On Tue, Jan 22, 2019 at 10:07 AM Jan Mulder <notifications@github.com>
wrote:

> I think we more-or-less addressed most of the points discussed in this PR,
> even if not 100% satisfyingly (e.g. the negate thing needs UI work).
> Therefore, with a bit of additional testing, perhaps it could even go into
> release? Depending on how pressing it is of course.
>
> I'd just add another "regression", though not sure how important it is:
>
> 9. Missing reset-filter button.
>
> And 10. Missing "filter shows x of y dives" . That is something I actually
> use for questions like: how many cave dives did I make with buddy A.
>

That's hard to do as the filter acts upon a row. but you have the amount of
dives that are shown in the list.

Current time seems slightly better, as a possible open filter and an import

dirkhh:
Yes, we definitely need the number of dives visible somewhere.
The reason for the urgency with a release right now is that we have a couple of bugs that really need fixing. I may cut the release as early as this week.
And then I'm perfectly happy to do another one two or three weeks later :-)

bstoeger:
Another potential issue:

11. Filter doesn't display in list mode. The old filter worked in that mode (which makes sense IMHO).

tcanabrava:
That should be easy to fix. I’ll look at the code tomorrow

Em sex, 25 de jan de 2019 às 22:22, bstoeger <notifications@github.com>
escreveu:

dirkhh:
I think it would be useful to track all the different issues with the filter code as GitHub issues... that way commits that fix them can close the issue and we see what's left to do...

dirkhh:
And independent from that last comment. I think the filter does work in list mode - at least it seemed to do the right thing in a quick test I just did (which happens to include your latest PR, @bstoeger). Can you let me know what doesn't work? Thanks

bstoeger:
With list mode I mean when you go "View > Dive List".

Moreover, the old filter had a "hide" mode, where it would only show the number of selected dives, but hide the rest of the UI. This could be useful if e.g. you want to show all wall dives on the map.

True, we might want to start using GitHub issues.

dirkhh:
Ahh - and here I thought Tree vs. List mode in the dive list.
Yeah, if you go full screen on the dive list, the new filter placement is a bit awkward.
Let's see what @tcanabrava can come up with.
And yes, we should have the number of shown dive over in the dive list widget, not in the filter widget so we can hide that one more easily.
Please create issues for those. I'm heading to Taekwondo this morning, but I am planning to spend some time on all this in a couple of hours

PR #1940, opened 2019-01-20
Desktop UI Filter improvements

janmulder:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

1) simple commit taking care of the poor layout on the star widgets on top of the filter widget.
2) On all (most?) places we use separate date/time fields for the time of a dive, and we follow the setting from the preferences to format those. Make the new filter widget consistent, with respect to the to and from
interval.
3) improved the tab order of the filter widget

### Changes made:
See above

### DIFF
--- a/desktop-widgets/filterwidget2.cpp
+++ b/desktop-widgets/filterwidget2.cpp
@@ -24,7 +24,13 @@ FilterWidget2::FilterWidget2(QWidget* parent) : QWidget(parent)
 	ui.labelEquipment->hide();
 	ui.invertFilter->hide();
 
-	ui.to->setDate(data.to.date());
+	ui.fromDate->setDisplayFormat(prefs.date_format);
+	ui.fromTime->setDisplayFormat(prefs.time_format);
+
+	ui.toDate->setDisplayFormat(prefs.date_format);
+	ui.toTime->setDisplayFormat(prefs.time_format);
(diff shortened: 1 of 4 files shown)


janmulder:
Force pushed as it did not compile on Travis for older Qt than 5.8. I used a too new function.

dirkhh:
I think we should consider raising the minimum Qt version. I don't think there aren't supported distros older than 5.8 (but I could be wrong)

janmulder:
> I think we should consider raising the minimum Qt version. I don't think there aren't supported distros older than 5.8 (but I could be wrong)

Agree, but this case was very trivial to fix (and shows the weirdness that goes on in Qt versions). `QDateTime::toMSecsSinceEpoch()` since Qt 4.7, but the almost identical `QDateTime::toSecsSinceEpoch() ` seconds variant since 5.8.

dirkhh (approved):
Code looks fine, haven't tested (dealing with Travis instead of having fun with code... oh how I suffer)

dirkhh:
BTW - I was wrong... Ubuntu 16.04 (which quite a few people are still on and which has two years left in its official life cycle) is on Qt 5.5. So unless we have to, it would make sense to continue supporting Qt 5.5 for a while longer...

PR #1941, opened 2019-01-20
Filter: implement imperial units.

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

<!-   - Describe your pull request in detail. -->
Make the temperature-fields aware of imperial units. This is far from perfect, as the values are not converted and taken as-is. Ultimately, we want the min and max values depend on the selected units. And we probably also want to store the value in a unit-independent way.

But this should make the filter usable for imperial unit users. Please test.

### Changes made:

1) Use the correct conversion function
2) Add suffix
3) Extend range of valid temperatures

### DIFF
--- a/desktop-widgets/filterwidget2.cpp
+++ b/desktop-widgets/filterwidget2.cpp
@@ -1,5 +1,7 @@
 #include "desktop-widgets/filterwidget2.h"
 #include "desktop-widgets/simplewidgets.h"
+#include "core/qthelper.h"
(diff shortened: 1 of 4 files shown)


dirkhh:
ok, tested, but it doesn't merge cleanly.
Would you mind re-creating this on top of current master, @bstoeger ?

PR #1943, opened 2019-01-22
Filter: quick implementation of negation

bstoeger:
Add negate buttons to the Tags, People, Location and Equipment
filters. Currently, if nothing is entered the filter is ignored
whether negate is on or off. One might think about filtering all
dives without tags, etc. instead.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Quick & Dirty implementation of individual negate buttons in the filter. See commit-message. I didn't find out how to properly align the UI fields.

### Changes made:

1) Add negate buttons to four filter items.

### Mentions:

@dirkhh @janmulder

### DIFF
--- a/desktop-widgets/filterwidget2.cpp
+++ b/desktop-widgets/filterwidget2.cpp
@@ -92,13 +92,18 @@ FilterWidget2::FilterWidget2(QWidget* parent) : QWidget(parent), ignoreSignal(fa
 	connect(ui.locationNegate, &QToolButton::toggled,
 		this, &FilterWidget2::updateFilter);
 
-	connect(ui.logged, SIGNAL(stateChanged(int)), this, SLOT(updateLogged(int)));
+	connect(ui.logged, &QCheckBox::stateChanged,
+		this, &FilterWidget2::updateLogged);
 
-	connect(ui.planned, SIGNAL(stateChanged(int)), this, SLOT(updatePlanned(int)));
+	connect(ui.planned, &QCheckBox::stateChanged,
+		this, &FilterWidget2::updatePlanned);
 
 	// Update temperature fields if user changes temperature-units in preferences.
-	connect(qPrefUnits::instance(), &qPrefUnits::temperatureChanged, this, &FilterWidget2::temperatureChanged);
-	connect(qPrefUnits::instance(), &qPrefUnits::unit_systemChanged, this, &FilterWidget2::temperatureChanged);
+	connect(qPrefUnits::instance(), &qPrefUnits::temperatureChanged,
+		this, &FilterWidget2::temperatureChanged);
+
+	connect(qPrefUnits::instance(), &qPrefUnits::unit_systemChanged,
+		this, &FilterWidget2::temperatureChanged);
 
 	// Update counts if dives were added / removed
 	connect(MultiFilterSortModel::instance(), &MultiFilterSortModel::countsChanged,
--- a/desktop-widgets/filterwidget2.cpp
+++ b/desktop-widgets/filterwidget2.cpp
@@ -20,8 +20,8 @@ FilterWidget2::FilterWidget2(QWidget* parent) : QWidget(parent), ignoreSignal(fa
 
 	// TODO: unhide this when we discover how to search for equipment.
 	ui.equipment->hide();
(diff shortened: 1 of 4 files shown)


janmulder:
Tried it. Layout looks ok for me. That is, it feels like the filter widget keeps "growing" (which I do not like), so I think we have to be careful not making it (much) bigger.

Further, I think I see something strange. Filter on people: a AND b, now negate. I would expect NOT (a AND b) == NOT a OR NOT b, but I see a dive with only a. So it seems to filter a AND NOT b (or something).

bstoeger:
> Tried it. Layout looks ok for me. That is, it feels like the filter widget keeps "growing" (which I do not like), so I think we have to be careful not making it (much) bigger.

Yes, that's what I meant with bad layout: the new fields shouldn't extend the whole widget, but shrink the text-fields. These seem amply spaced to me anyway.

> I would expect NOT (a AND b) == NOT a OR NOT b, but I see a dive with only a

But a dive with only a *does* fulfill "NOT a OR NOT b", no? Or do you mean that negation should negate the individual entries? So would you rather have "NOT a AND NOT b"?

Instead of the buttons we could also make this a "power-feature": Add a "~" in front of a buddy's name so that  dives with this buddy are not shown. This would give ultimate flexibility: "a,b", "~a,b", "a,~b", "~a,~b".

janmulder:
> 
> > I would expect NOT (a AND b) == NOT a OR NOT b, but I see a dive with only a
> 
> But a dive with only a _does_ fulfill "NOT a OR NOT b", no? 

Yes, you are right. I was confused, and can see how a regular user will get confused by this boolean logic :-)

> Or do you mean that negation should negate the individual entries? So would you rather have "NOT a AND NOT b"?

That as indeed what I was expecting, but that does say more about my expectation, than the formally correctness of the logic.
> 
> Instead of the buttons we could also make this a "power-feature": Add a "~" in front of a buddy's name so that dives with this buddy are not shown. This would give ultimate flexibility: "a,b", "~a,b", "a,~b", "~a,~b".

I like this. Flexible and intuitive from a users perspective IMHO.

dirkhh:
Careful with mixing "correct logic" with "what users expect"
I think when people filter for "a AND b" and then invert the filter, what they expect may either
NOT (a AND b) (so all dives that don't have both of those people - the actual negation)
or (more likely, IMHO)
(NOT a) AND (NOT b) (all dives that have neither of those people)
I'm not opposed to introducing cool syntax for the pros, but easy things should be easy and behave as you'd expect. And part of that is how this is documented.
Instead of an 'negate' or 'invert' button the button could have a switch / radio button 'include' and 'exclude' a second one that says 'any' / 'all' (but frankly, I don't think that second one is worth it)

bstoeger:
> I'm not opposed to introducing cool syntax for the pros, but easy things should be easy and behave as you'd expect. And part of that is how this is documented.

What about a mouse-hover alt-text: "prepend '~' to exclude person / tag / location"?
 
> Instead of an 'negate' or 'invert' button the button could have a switch / radio button 'include' and 'exclude' a second one that says 'any' / 'all' (but frankly, I don't think that second one is worth it)

Sure, that's all easy to implement, but the two-button / four-states version seems excessive even to me (and I like configurable interfaces). Moreover, I fear that "include/exclude" takes up a bit much screen real estate...

bstoeger:
I changed to logic to "exclude all". Would be cool if someone else could do the UI part.

dirkhh (approved):
This looks good.
Clarification: what do you mean by "do the UI part"?

bstoeger:
> Clarification: what do you mean by "do the UI part"?

1) Align the buttons so that they don't stretch the widget, but compress the text-fields.
2) Replace the buttons by combo box / checkbox, ...

dirkhh:
The fix to not stretch the widget is easy, I can add that.
My proposal would be that we get the overall functionality to the point that we want, and then do a pass over the UI to make it look nice.

BTW: merging the other PR now causes a conflict here. I'm sure I can fix this with a local merge.
Would you like me to merge this now, or are there functionality changes you want to do first?

PR #1944, opened 2019-01-22
Filter: Negation with '~' [Alternative to #1943]

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As discussed in #1943, this is a finer-grained negation mechanism, which lets the user prepend '~' to filter dives *without* that term.

What we still have is the "substring" problem, where e.g. one buddy name is included in another, so negation would filter out too many dives. Possible solution: Different modes: substring, starts-with, whole-word-match?

### Changes made:

1) Implement negation by prepending '~'

### Related issues:

#1943
### Documentation change:

I guess that should be documented, as it is quite obscure!
### Mentions:

@dirkhh @janmulder

### DIFF
--- a/desktop-widgets/filterwidget2.cpp
+++ b/desktop-widgets/filterwidget2.cpp
@@ -21,7 +21,6 @@ FilterWidget2::FilterWidget2(QWidget* parent) : QWidget(parent), ignoreSignal(fa
 	// TODO: unhide this when we discover how to search for equipment.
 	ui.equipment->hide();
 	ui.labelEquipment->hide();
-	ui.invertFilter->hide();
 
 	ui.fromDate->setDisplayFormat(prefs.date_format);
 	ui.fromTime->setDisplayFormat(prefs.time_format);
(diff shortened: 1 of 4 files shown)


janmulder:
I like this :-) 

And agree about putting something in the documentation (but that is true anyway for the new filter widget).

bstoeger:
> My preference would be sub-string on location but starts-with on people.

This sounds like a reasonable and easily implemented compromise.

Perhaps for people we should make it whole-word, but with auto-completion. Say you enter a letter and it automatically fills out the first buddy (alphabetically) with that letter. If not happy, enter a second letter, etc. But right now I'm not ready to invest the time necessary for such a feature.

henrik242:
@dirkhh I agree, only programmers will understand this.  Using `NOT` would be much better.

bstoeger:
With #1943 in, little point in keeping this one here up-to-date.

PR #1950, opened 2019-01-25
Filter: Implement counts, reset button and close button

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Reimplements functionality of the old filter. This is all more hairy and than one would think:

1) The counts have to be actualized on addition / removal of dives (think undo while filter is active).
2) Resetting text fields causes signals which have to be ignored.
3) The from date has to be set explicitly to make reset of from date work.
4) One can't simply hide the widget, one has to reset the "application state".

All in all not fun at all.

UI is horrible, but I don't know how to handle that grid layout. Help, please.

Note: for undo I noticed a horrible UI behavior where on pressing CTRL^Z, the text field in the filter *and* the last dive action were undone. This is not acceptable and it seems that we're doing something wrong concerning hooking to CTRL^Z.
 
### Changes made:

1) Implement counts
2) Update counts on dive addition / removal
3) Implement reset
4) Implement close

### DIFF
--- a/desktop-widgets/command_divelist.cpp
+++ b/desktop-widgets/command_divelist.cpp
@@ -141,14 +141,6 @@ std::vector<dive *> DiveListBase::addDives(DivesAndTripsToAdd &toAdd)
 	std::vector<dive *> res;
 	res.resize(toAdd.dives.size());
 
-	// First, tell the filters that new dives are added. We do this here
-	// instead of later by signals, so that the filter can set the
-	// checkboxes of the new rows to its liking. The added dives will
-	// then appear in the correct shown/hidden state.
-	QVector<dive *> divesForFilter;
-	for (const DiveToAdd &entry: toAdd.dives)
-		divesForFilter.push_back(entry.dive.get());
-
 	// Now, add the dives
 	// Note: the idiomatic STL-way would be std::transform, but let's use a loop since
 	// that is closer to classical C-style.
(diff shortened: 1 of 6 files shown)


dirkhh (approved):
As always, your code looks good.
I'll try to find time today (just returned from Oceania, I'm mildly jetlagged) to see if I can make the UI better. But that can be done after this is merged, I think.
The one comment I didn't understand was about ^Z - what do you expect it to do here?
I would disable ^Z for the text boxes...

bstoeger:
If I'm currently entering something in the people filter box, I would expect ^Z to undo my last change to that.

What I definitely do not expect (but what happened) is that it undid my last change in the filter box (removed buddy) *and* undid the deletion of a dive. The result was very confusing. It's surprisingly hard to let the filter-box lose focus: pressing ESC does not do it.

dirkhh:
I guess I'm not assuming (like you are) that ^Z undoes my typing.
If that's what you want to have, just disconnect the keyboard shortcut for the overall ^Z/undo functionality as one of the text fields gains focus (and reconnect afterwards).
Easy in theory, but I remember struggling with this overall in the past.
But I think that can be done in a separate PR, so I'll merge this one.

PR #1953, opened 2019-01-26
Filter UI update

dirkhh:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Instead of having the negate button with the logical symbol for negation, have a button with a text that explains what we do with the text / tags entered, and a tooltip that tells the use that they can toggle that action to its negation. Visually this makes a lot more sense than this explanation, I hope :-)

Also added a spacer at the bottom so the UI isn't so weirdly stretched out

### Documentation change:

Obviously the whole new filter UI needs documentation, once it's done.

### Mentions:

@bstoeger @tcanabrava

### DIFF
--- a/desktop-widgets/filterwidget2.cpp
+++ b/desktop-widgets/filterwidget2.cpp
@@ -20,7 +20,7 @@ FilterWidget2::FilterWidget2(QWidget* parent) : QWidget(parent), ignoreSignal(fa
 
 	// TODO: unhide this when we discover how to search for equipment.
 	ui.equipment->hide();
-	ui.equipmentNegate->hide();
+	ui.equipmentMode->hide();
 	ui.labelEquipment->hide();
 
 	ui.fromDate->setDisplayFormat(prefs.date_format);
@@ -50,6 +50,11 @@ FilterWidget2::FilterWidget2(QWidget* parent) : QWidget(parent), ignoreSignal(fa
 	connect(ui.minVisibility, &StarWidget::valueChanged,
 		this, &FilterWidget2::updateFilter);
 
+	// We need these insane casts because Qt decided to function-overload some of their signals(!).
+	// QDoubleSpinBox::valueChanged() sends double and QString using the same signal name.
+	// QComboBox::currentIndexChanged() sends int and QString using the same signal name.
+	// Qt 5.7 provides a "convenient" helper that hides this, but only if compiling in C++14
+	// or higher. One can then write: "QOverload<double>(&QDoubleSpinBox::valueChanged)", etc.
 	connect(ui.maxAirTemp, static_cast<void (QDoubleSpinBox::*)(double)>(&QDoubleSpinBox::valueChanged),
 		this, &FilterWidget2::updateFilter);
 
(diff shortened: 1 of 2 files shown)


bstoeger (approved):
Code is fine with me.

But the UI behaves a bit strangely:
1) The "none of" mode is dark grey, the "all of" mode is light grey.
2) Switching between "matches" and "doesn't match" resizes the widget, which gives a bit of a "fluttery" feel.

Therefore, I wonder if a combo-box would be more appropriate.

dirkhh:
I ran out of time working on this - I have the sizes for the button fixed, but I'm still trying to find a clean and simple way to make them not change color.

bstoeger:
I tried to replace the buttons by combo-boxes, c.f. #1954. At least programmatically, this makes things easier as we don't have to (re)implement combo-box semantics. Moreover, it opens the possibility to add an "any of" mode, though I find that somewhat questionable. The UI feels a bit clunky though. I'll let others decide..

bstoeger:
> but I'm still trying to find a clean and simple way to make them not change color.

I think that should be rather easy: remove the `checkable` property and replace the `toggled` by the `clicked` signal.

bstoeger:
(And in the clicked signal, don't read the checked property, but invert the value in the filter state.)

dirkhh:
This has what was in #1954 except with the correct author for the third commit plus a commit that does a tiny bit of UI cleanup.
If this is what we want to go with (would love to hear from @janmulder and @henrik242 , among others) I can do some more cleanup work. One thing I'll do is move the "count" of selected dives to the dive list widget so that it's visible when the dive list goes full window...

janmulder:
LGTM (except for a trivial spelling error in the last commit title: ceanup). Obviously, I like the nerdy ~ negate way still better, but that is voted down :smile:

dirkhh:
@janmulder : send me a PR that adds a "Nerdmode" option to the settings and accepts the expanded syntax in the filter. I'm not AGAINST having that. I am against confusing most of our users. Big difference.
PS: will fix commit title...
PPS: also trying to understand why the Linux build failed...

dirkhh:
Hmm, @probonopd - I think your latest update to `linuxdeployqt` breaks our build. Well, actually, no it doesn't break our build, it just now requires some modules that we didn't need before... why are `libqgtk` and `libqgtkstyle` required now? I think it's this commit that causes the issue https://github.com/probonopd/linuxdeployqt/commit/eba669d39b7f5d7ee67cf1e4ee4e7c457d19430e and this allegedly fixes a bug... but I still don't understand why we now need those two additional libraries...

probonopd:
Hi @dirkhh what I intended to do is bundle libqgtk and libqgtkstyle _if they are there_. This is to make the application look native on Gtk+ systems. 

But the breakage actually has nothing to do with those, but my failed attempt at bundling all iconengines and imageformats. Sorry it broke your build. Reverting.

dirkhh:
Ok, I'm merging this so we can continue to build on top of it.

PR #1955, opened 2019-01-27
[TRIVIAL] Cleanup: remove redundant conditional

bstoeger:
In MainTab::acceptChanges there was a "editMode != ADD" condition
inside a else block to "editMode == ADD", which is therefore
redundant. Remove.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a trivial code-cleanup that removes an always-true conditional.

### Changes made:

1) Remove tautological conditional.


dirkhh (approved):
Sometimes when code moves around things like that happen.
But yeah, this looks correct

PR #1956, opened 2019-01-29
Add glib static build on Dockerfile

salvadorcunat:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [X] Build system change
- [ ] Documentation change
- [ ] Language translation

First commit should build glib statically -on the docker image used to build and pack the mxe libraries for the windows binaries- which is needed to build mdbtools under mxe environment.
Second commit can be dropped if you think it is superfluous.

### Changes made:

1) Build glib (and its dependencies) statically.
2) pass an argument from the docker build command line to avoid manually editing the Dockerfile each time the image is rebuilt.

### Related issues:

Please see comments on commit 3738879.

### Mentions:

@dirkhh

### DIFF
--- a/scripts/docker/mxe-build-container/Dockerfile
+++ b/scripts/docker/mxe-build-container/Dockerfile
@@ -43,6 +43,7 @@ RUN apt-get update  &&  apt-get upgrade -y && \
     git checkout 9f6b9c6f58510bedfa0bf9f87de9a214abe6b653 ; \
     mv ../settings.mk . ; \
     make -j 6 2>&1 | tee build.log ; \
+    make MXE_TARGETS=i686-w64-mingw32.static glib -j 6 2>&1 |tee build.log ; \
     mkdir -p neolit ; cd neolit ; git clone -b wip/win http://code.qt.io/cgit/qt/qtconnectivity.git ; \
     cd qtconnectivity ; \
     PATH=/win/mxe/usr/bin/:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /win/mxe/usr/i686-w64-mingw32.shared/qt5/bin/qmake qtconnectivity.pro ; \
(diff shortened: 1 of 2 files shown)


scripts/docker/mxe-build-container/Dockerfile
```diff
@@ -1,5 +1,10 @@
-From ubuntu:18.04
+# Build the image using the --build-arg option, e.g.:
+# docker build -t boret/myimage:0.1 --build-arg=mxe_sha=123ABC456 .
```
dirkhh, line 2: nice! I like that. I just learned something new :-)

salvadorcunat:
Hi @dirkhh.
Finally could build the image lowering the parallelism to the bare minimum and it can build both subsurface and smtk2ssrf binaries. I see you realized that the final purge removes most tools needed for the build.
I've pushed another commit to add -a param to tee in the second make command which is currently overwriting the build log; and another one to remove the download of the static libraries package in windows-container/before_install.sh
BTW It could be good to add another --build-arg to the Dockerfile for the image tag (0.7 in the current Dockerfile).
Best regards.

dirkhh:
It looks like you aren't looking at my branch where I pushed similar changes...
https://github.com/Subsurface-divelog/subsurface/tree/windowsQt5.12

thanks for doing all this work. I'll look at this later this morning.

dirkhh:
Your build is failing with the same error that I see in my branch:
`configure: error: C compiler cannot create executables`
I don't see this error when running locally with the same Docker image, so I'm still trying to figure out how I managed to break that...

salvadorcunat:
Hmmm.  The only difference I can see is:
##### local log
`checking for i686-w64-mingw32.static-strip...  i686-w64-mingw32.static-strip`
##### travis log
`checking for i686-w64-mingw32.static-strip... no`

Have you updated the image in docker-hub?  travis log claims that before_install.sh is running  `dirkhh/mxe-build-container:0.6`, otherwise, the script should point to something like `dirkhh/mxe-build-container:0.7`
But 0.7 image lacks of git, so it will not work.

dirkhh:
I incorporated (and slightly updated) these commits here into #2098, so I'm closing this PR...

PR #1957, opened 2019-01-30
[TRIVIAL] Minor temperature changes

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The first commit is a trivial code-cleanup (directly return `temperature_t` from the temperature parser). The second is pure pedantry: Kelvins are not degrees.

### Release note:

Obviously no.

### Documentation change:

Obviously no.


atdotde:
This is one of the few places where unsigned actually makes sense... LGTM

bstoeger:
Changed my mind - for consistency, we would have to convert all parsing functions. Not worth it.

bstoeger:
Heh, commented at the same time. :)

> This is one of the few places where unsigned actually makes sense...

But then we should also check user-input for >= 0 K, which we currently don't.

And what about temperature differences, these can be negative, isn't it?

PR #1962, opened 2019-02-02
Export profile image and clipboard text

atdotde:
With Facebook support gone, we should offer a way to export
the profile image. This has been part of the TeX support
but this makes it explicit.

Furthermore, copy dive items to the system clipboard so they can be pasted in other applications.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

### Documentation change:

Needs a new screen of the profile dialogue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Export option for profile picture
 - Desktop: remove support for the "Share on Facebook" feature.
   Rationale: It is fairly easy to share images on Facebook, thus it was decided
   that this feature is redundant and should be removed from Subsurface.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 - Export option for profile picture
+- Export: fix picture thumbnails [#1963]
 - Desktop: remove support for the "Share on Facebook" feature.
   Rationale: It is fairly easy to share images on Facebook, thus it was decided
   that this feature is redundant and should be removed from Subsurface.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-- Export option for profile picture
+- Desktop: Add export option for profile picture [#1962]
 - Export: fix picture thumbnails [#1963]
 - Desktop: remove support for the "Share on Facebook" feature.
   Rationale: It is fairly easy to share images on Facebook, thus it was decided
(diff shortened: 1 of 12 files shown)


desktop-widgets/simplewidgets.cpp
```diff
@@ -490,6 +491,50 @@ void DiveComponentSelection::buttonClicked(QAbstractButton *button)
 		COMPONENT_FROM_UI(cylinders);
 		COMPONENT_FROM_UI(weights);
 		selective_copy_dive(&displayed_dive, targetDive, *what, true);
+		QClipboard *clipboard = QApplication::clipboard();
```
neolit123, line 494: we might want to add a static text in the dialog that data is copied to the clipboard too.
otherwise right now it's unclear.

neolit123 (commented):
thanks for this extension @atdotde !
added some minor comments.

atdotde:
This now includes @neolit123 's suggestions as well as @bstoeger 's commits from #1965 to to fix the images in the profile export reported in #1963 (manually merged to work with this line of history).

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+- Add export option for profile picture
```
neolit123: sorry that i didn't suggest last time, we might want to prefix this with `Desktop:` and also link to the PR `[ #.... ]` ?

desktop-widgets/divelogexportdialog.cpp
```diff
@@ -229,6 +235,36 @@ void DiveLogExportDialog::export_depths(const char *filename, const bool selecte
...
+	int count = 0;
+	if (!filename.endsWith(".png", Qt::CaseInsensitive))
+		filename = filename.append(".png");
+	QFileInfo fi(filename);
+
+	for_each_dive (i, dive) {
+		if (selected_only && !dive->selected)
+			continue;
+		if (count)
+			saveProfile(dive, fi.path() + "/" + fi.completeBaseName().append(QString("-%1.").arg(count)) + fi.suffix());
```
neolit123: Windows will probably work with the `/` here, but we might want to use `QDir::separator()`

neolit123 (commented):
thanks for the update.
added two final comments on my side.

CHANGELOG.md
atdotde: Will amend the @neolit123 suggestions commit accordingly. I don't think we should have issue numbers in the CHANGELOG as that is end user facing.

dirkhh:
Actually, we have had issue numbers in the CHANGELOG - that really is developer facing. When it gets moved into the release notes, and especially for the announcements, that's when I drop them, typically.

willemferguson:
I am aware of the image changes required for the user manual. I will finalise this once the PR has been pulled into master. I would need information on the text info that is copied onto the clipboard. In prelim tests of this PR I could not access the text info.

Issue #1951, opened 2019-01-26
Filter widget not shown in fullscreen dive-list mode


bstoeger:
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [x] New feature request
- [ ] Discussion request

### Issue long description:

When going "View > Dive list", the filter widget is not accessible. But in this view the filter makes very much sense. Map view is another view where the filter might be handy.

### Operating system:

Any desktop OS.

### Subsurface version:

Master.

### Steps to reproduce:

1) View > Dive list
2) CTRL^F
### Current behavior:

No visible changes.
### Expected behavior:

Filter widget is shown.
### Mentions:

@dirkhh @janmulder

PR #1970, opened 2019-02-10
Filter for maximized divelist

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Show the filter pane with the maximized dive list

### Changes made:

Simply enable the bottom right pane in dive list maximized mode, depending on whether we are filtering or not.

### Related issues:

#1951 

### Release note:

Not needed as the filter hasn't been released, yet.

### Documentation change:

Not sure if this deserves a mention, @willemferguson 

### Mentions:

@bstoeger


bstoeger (approved):
Tested it - and works good for me.

Can't really comment on the code, as I have never looked into the state thing. It looks reasonable though. Personally, I find it a bit weird to define a local variable as const - but many people think that this should be the default, so it's perfectly fine.

PR #1974, opened 2019-02-14
Only show dive computers

dirkhh:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This had bugged me for a while, but it became especially annoying when Linus and I were trying to track down a bug in a libdivecomputer merge that caused a crash on a Teric. We were at the linux.conf.au conference and Subsurface found 200+ devices and the device selection dialog became completely unusable. But why are we showing all these devices - in most typical circumstances we know which devices are the potential dive computers. So let's just show those (and add a way to temporarily override this on both desktop and mobile.

### Related issues:

Not sure if this might help with #1862 

### Release note:

Yeah, this would have to be added to the CHANGELOG

### Documentation change:

... and to the documentation

### Mentions:

I'd love some feedback @bstoeger @janmulder @atdotde @willemferguson @glance- 
This is LIGHTLY tested...

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -82,7 +82,12 @@ static dc_descriptor_t *getDeviceType(QString btName)
 	if (!vendor.isEmpty() && !product.isEmpty())
 		return descriptorLookup.value(vendor + product);
 
-	return NULL;
+	return nullptr;
+}
+
+bool matchesKnownDiveComputerNames(QString btName)
+{
+	return getDeviceType(btName) != nullptr;
 }
 
 BTDiscovery::BTDiscovery(QObject*) : m_btValid(false),
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -91,6 +91,7 @@ bool matchesKnownDiveComputerNames(QString btName)
 }
 
 BTDiscovery::BTDiscovery(QObject*) : m_btValid(false),
+	m_showNonDiveComputers(false),
 	discoveryAgent(nullptr)
 {
 	if (m_instance) {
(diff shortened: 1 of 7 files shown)


janmulder:
I like the idea and I'm sure many users in a noisy BT environment will like it. Funny is that Matthias from HW told me this already 2 times at the BOOT show. So, I gave it a try ...

First, my BT environment is rather unstable at my Arch box. So, I did 2 tries on desktop. Self build (Qt 5.12.1) and the appImage. The self build is very unstable in BT scanning/switching controller on and of (from Subsurface this does not work, and from the KDE Plama desktop tools it does work but Subsurface does not see that change). So: self build is useless for reliable testing. But the filter seems to work ...

The AppImage (with Qt 5.9.3) does work a bit better, but is also influenced by the Arch stuff. Errors like "Device discovery aborted due to unexpected adapter changes from another process.". This makes testing also rather problematic there.

And a quick test on mobile (Android, self build) => all seems to work. The only minor thing is the slightly long string in the settings overlapping the toggle button (so UI only).

dirkhh:
Thanks for testing, @janmulder 
Good catch on the overly long string on mobile. I did change the string on desktop to be partially in the tooltip. Can't do that on mobile :-)
I guess I'll split it into two lines.

bstoeger (approved):
Didnt test (on phone), but code and feature lgtm.

dirkhh:
I did some more testing, including testing on device. Apple devices sometimes don't resolve names and then you need to try twice (but otherwise you'd have to KNOW which one of the random hex strings is the right one, so this isn't REALLY an issue). I guess I'll merge it and we'll see where it takes us.
Oh, and I'll add a CHANGELOG for it as well.

PR #1976, opened 2019-02-18
Edit: use correct offset when changing dive times

### LINKED ISSUE #1975: Changing dive time/ date behaviour
[ ] Bug

### Issue long description:
I've just updated to latest Mac version 4.8.5. 

I have a number of different models of computers, and have been aligning date and time for some dives where the times weren't aligned.

Previously I was able to change date and time in the Notes tab successfully. Since update to latest version, if I try to amend the time or date, when I select 'accept changes' the time and date in the dive list changes to a date time which bear no resemblance to the date / time changed to. Initially the date and time display as the correct revised date time in the notes page, but if I click into another dive and back in, the notes page also changes to the incorrect date time. 

As an example, I have tried several times to make a time change (which was initially supposed to be a change of an hour).  Each time it makes it worse. e.g. current dive date time in Notes tab is Thu 21 Feb 2019 22.28, on trying to change it to Fri 15th Feb 11:00 it changes to Thurs 28 Feb 2019 9:56. If I try again to set it to 15th Feb at 11:00 it now changes to Web 13th Marc 2019 8:52... (and so on). 

### Operating system:

Mac Mojave 10.14.2

Multiple dive computers. This example happens to be a Perdix, but the same applies to multiple models. 

### Subsurface version:

Latest Mac release 4.8.5 

official release

### Steps to reproduce:

1) Upload a dive 
2) Enter notes tab
3) Change the time
4) Accept changes
5) See date/time in the dive list. 

### Current behavior:

As above

### Expected behavior:

Date time should change correctly.

bstoeger:
The undo-work reversed the direction of the offset. This was apparently
only fixed when using the menu entry, but not when editing dives directly.
Invert the offset to get the correct time.

While doing so, remove a redundant if: First it checked whether the
dates are the same, then whether the offset is non-zero.

Fixes #1975.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Editing the date or time shifted the dive-time in the wrong direction. Invert the offset and remove a redundant conditional.

### Changes made:

1) Fix direction of time-offset.
2) Remove redundant if-clause.
### Related issues:

Fixes #1975.

### Release note:

Yes, probably.
### Documentation change:

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Desktop: Fix editing of dive-time [#1975]
 - Desktop: Add export option for profile picture [#1962]
 - Export: fix picture thumbnails [#1963]
 - Desktop: remove support for the "Share on Facebook" feature.
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -892,11 +892,9 @@ void MainTab::acceptChanges()
 			}
 		}
 
-		if (displayed_dive.when != cd->when) {
-			timestamp_t offset = cd->when - displayed_dive.when;
-			if (offset)
-				Command::shiftTime(selectedDives, (int)offset);
-		}
+		timestamp_t offset = displayed_dive.when - cd->when;
+		if (offset)
+			Command::shiftTime(selectedDives, (int)offset);
 	}
 	if (editMode == MANUALLY_ADDED_DIVE) {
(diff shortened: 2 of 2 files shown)


dirkhh (approved):
Thanks for fixing this so quickly...

I'll wait for Travis on principle, but this looks obviously correct.

PR #1978, opened 2019-02-18
[WIP] Make editing in notes-tab undoable

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Implement undo of editing in the dive-tab. This is a bit hairy. For one, some widgets don't send `editingFinished` signals, so we only get informed on every key-press, which seems way too fine-grained. So high-jack the focus-out events in these cases. Moreover, some editing commands act differently: depth and time only edit the current dives, the other fields all selected dives. The code tries to keep the old behavior as far as possible.

This also includes the dive-sites in the undo-system. That should now be more save.

A few points are broken:
1) No more warning if multiple dives edited. Not sure how to proceed here. Show a message as long as the field has focus?
2) Breaks copy & paste. Copy & past will have to be integrated in the undo-system.
3) The whole ^Z treatment is strange. If a text field has focus, ^Z will only act on that field. Even if there are no changes. So it's hard to actually undo things when a field is active: the user has to press ESC first to remove focus from any input field.
4) Editing of depth and duration does not enter edit-mode anymore. To manually change the profile of an entered dive, the user now has to go via the edit-dive menu.

### Changes made:

1) Make editing of fields on the notes tab undoable.
2) Include dive-sites in undo-system.
3) Minor cleanups.

### Release note:

Yes.
### Documentation change:

Not sure. Depends on how detailed the documentation is.
### Mentions:

The usual suspects.


dirkhh:
You are a brave man. The inconsistent behavior of the various widgets has driven me to rage in the past. Which signal to use when to indicate what. ARGL.

desktop-widgets/command.cpp
```diff
@@ -2,10 +2,12 @@
 
 #include "command.h"
 #include "command_divelist.h"
+#include "command_edit.h"
 
 namespace Command {
 
 // Dive-list related commands
+
```
dirkhh: why an empty?

Also, there's a typo in the commit message, and as we all know by now you can't comment on those, so I'll comment here that the commit message of "Undo: implement rudimentary undo of dive-notes editing" using "editig" instead of "editing"...

desktop-widgets/command_edit.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+	std::swap(old, value);
+
+	mark_divelist_changed(true);
+}
+
+// The constructors of the base class will be called directly by functions
+// generating the derived classes, owing to the "using" declaration.
+// But they don't have access to the implementation so can't instantiate
+// them. Therefore, explicitly intantiate the constructors here.
```
dirkhh: I understand all the words here... I don't think I actually understand what you are telling me.

#InOverMyHead

desktop-widgets/command_edit.cpp
```diff
@@ -0,0 +1,89 @@
+// SPDX-License-Identifier: GPL-2.0
+
+#include "command_edit.h"
+#include "core/divelist.h"
+
+namespace Command {
+
+template<typename T>
+EditBase<T>::EditBase(const QVector<dive *> &divesIn, T newValue, T oldValue) :
+	value(std::move(newValue)),
```
dirkhh: I looked ahead and it seems we only ever instantiate this for QString. So why is this a template class?
Or am I missing something?
 (OK, that DOES seem rather likely)

Yeah, coming back to state that I found this in all of the following commits... :-/

bstoeger:
A word on the implementation: This uses a very old-school "object oriented" style, which seems old-fashioned nowadays: Commands are implemented using subclasses with virtual functions. This needs quite a lot of ugly repetitive boiler plate. But firstly, this feels quite appropriate in a Qt-context, which uses a similar approach. And secondly, it turned out that adding new commands was rather trivial.

desktop-widgets/command_edit.cpp
```diff
@@ -93,4 +96,43 @@ DiveField EditNotes::fieldId() const
...
+// The index "i" will be stored as an additional payload with the command.
+// Thus, we need an explicit constructor. Since the actual handling is done
+// by the base class, which knows nothing about this index, it will not be
+// sent via signals. Currently this is not needed. Should it turn out to
+// become necessary, then we might either
+//	- Not derive EditMode from EditBase.
+//	- Change the semantics of the mode-editing.
+// The future will tell.
+EditMode::EditMode(const QVector<dive *> &dives, int indexIn, int newValue, int oldValue)
+	: EditBase(dives, newValue, oldValue), index(indexIn)
```
dirkhh: Ah, I get it. That previous commit smartly set things up for the later ones to add the non-QString implementations

core/dive.h
```diff
@@ -169,7 +169,7 @@ struct sample                         // BASE TYPE BYTES  UNITS    RANGE
 	duration_t rbt;                   // int32_t    4  seconds  (0-34 yrs)             remaining bottom time
 	depth_t depth;                    // int32_t    4    mm     (0-2000 km)            dive depth of this sample
 	depth_t stopdepth;                // int32_t    4    mm     (0-2000 km)            depth of next deco stop
-	temperature_t temperature;        // uint32_t   4  mdegrK   (0-4 MdegK)            ambient temperature
+	temperature_t temperature;        // uint32_t   4    mK     (0-4 MK)               ambient temperature
```
dirkhh, line 172: :-D

I love it

Especially in the middle of this rather dense (for me) series... A commit that I completely and fully understand...

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -1124,41 +1124,13 @@ void MainTab::divetype_Changed(int index)
...
-	if (editMode == IGNORE || acceptingEdit == true)
+	if (editMode == IGNORE || acceptingEdit == true || !current_dive)
 		return;
-	displayed_dive.watertemp.mkelvin = parseTemperatureToMkelvin(text);
-	markChangedWidget(ui.watertemp);
-	validate_temp_field(ui.watertemp, text);
-}
-
-void MainTab::validate_temp_field(QLineEdit *tempField, const QString &text)
-{
```
dirkhh: I'm sure this will be obvious to me in a moment, but can you explain one more time why you no longer want to validate the field?

desktop-widgets/command_edit.h
```diff
@@ -40,6 +35,11 @@ class EditBase : public Base {
 	EditBase(const QVector<dive *> &dives, T newValue, T oldValue);
 
 protected:
+	T value; // Value to be set
+	T old; // Previous value
+	void undo() override;
+	void redo() override;
+
```
dirkhh, line 42: why are these now protected? Is this a logical part of this commit?

desktop-widgets/command_edit.cpp
```diff
@@ -295,4 +295,126 @@ DiveField EditMode::fieldId() const
...
+		//: remove the part in parantheses for %n = 1
+		setText(tr("Edit %1 (%n dive(s))", "", num_dives).arg(fieldName()));
+
+	return num_dives;
+}
+
+void EditTagsBase::undo()
+{
+	if (dives.empty())
+		return; // Shouldn't happen
```
dirkhh: I typically like to print out  a warning in those "shouldn't happen" cases... makes debugging easier if they ever do happen :-)

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -204,8 +204,6 @@ MainTab::MainTab(QWidget *parent) : QTabWidget(parent),
 	// enable URL clickability in notes:
 	new TextHyperlinkEventFilter(ui.notes);//destroyed when ui.notes is destroyed
 
-	acceptingEdit = false;
```
dirkhh, line 205: this causes me flashbacks... I think we added this because there was some way that we ended up recursing into our own code and without this flag we did rather weird stuff. I am not sure that this is still true and I am definitely not at all sure that your code is wrong... I just remember that this was a major source of pain early on in the Qt code.

dirkhh (commented):
I repeat. Oh wow. That's once again quite a series of patches.
This has a few mostly trivial or curious comments, I didn't see anything that I can point my finger at and say "this is wrong"... but it definitely needs more review and then testing.

The parts that I understand and was able to follow looked good. But given that this is code that I have always considered rather fragile I'm both excited to see it get rewritten and worried about hidden issues...

dirkhh:
I wonder if we should do away with the "this is being edited, accept / discard" thing.
Since each edit can be undone, there really isn't a point anymore, is there?
Would make that things easier or harder for you?

bstoeger:
> I wonder if we should do away with the "this is being edited, accept / discard" thing.

That's what this PR does (but only for the notes tab). I want to do one tab after the other until finally reaching the planner (the thought gives me cold feet).

The problem with doing away with the edit-mode is that the user is not warned if they edit multiple dives. I'm a bit scared about inadvertent batch-edits. My idea was to show a temporary message as long as a text-field is edited or after performing an edit.

desktop-widgets/tab-widgets/maintab.cpp
bstoeger: Note that this commit does *not* remove the `acceptingEdit` functionality. Rather it unifies it with the `editMode == IGNORE` functionality that does exactly the same thing.

desktop-widgets/command_edit.h
bstoeger: Because the dive-site editing does some special stuff like deleting non-existing dive sites and therefore "hijacks" the `undo()` call. To be able to call the `undo()` function of the parent class that has to be made protected instead of private. Of course, I could change the history to make them protected from the start...

desktop-widgets/tab-widgets/maintab.cpp
bstoeger: Because the field is rewritten after editing anyway. That does in principle the same as this validate call.

desktop-widgets/command.cpp
bstoeger: Removed.

desktop-widgets/command_edit.cpp
bstoeger: Tried to reword the comment - not sure if the new one is more easily understood. It *is* subtle. Moving the constructor in a header file would probably remove the need of this instantiation, but I'm not sure we want that.

desktop-widgets/command_edit.cpp
bstoeger: done.

dirkhh:
> The problem with doing away with the edit-mode is that the user is not warned if they edit multiple dives. I'm a bit scared about inadvertent batch-edits. My idea was to show a temporary message as long as a text-field is edited or after performing an edit.

I think we already mark any field that is being edited. Once focus shifts / editing ends, that mark should be removed (as now the edit is finished). I like the idea of a separate warning when the user edits multiple fields. Not sure exactly how that would work with things where the edit is a single click (rating, for example). But I guess having a transient warning at the bottom or something would work?

Maybe THIS is a question we should raise on the mailing list where more people see it.

bstoeger:
Grrr... Not sure how to make `qWarning() << ...` work on Qt <= 5.6. I'm tempted to replace it by an `fprintf(stderr, ...)`!

dirkhh:
It's frustrating that we still need Qt 5.6.
But every time I consider dropping it, I'm reminded that some Linux distros ship ancient libraries.

Feel free to use a `#ifdef` for this so that everywhere else we get the `qWarning()`

henrik242:
@dirkhh Those users can still use older versions of Subsurface IMHO.  Development should have some kind of priority :)

dirkhh:
> @dirkhh Those users can still use older versions of Subsurface IMHO. Development should have some kind of priority :)

I agree with the sentiment. But Ubuntu 16.04 is still one of our more commonly used OSs.
That said - it's a tiny minority compared to all the people on Windows & Mac (those two are about 85% of our desktop users). Which is why I suggested to use an `#ifdef` here. This is about a warning message, right? So ok, users on Ubuntu 16.04 won't get that. But everyone else will. That should force us to not ship on 16.04 (i.e., they get all the other new features).

bstoeger:
Next attempt...

dirkhh:
Yay, that did it :-)
With that out of the way, where are we with this PR? Any more insights on how we want to handle multi dive edits?

bstoeger:
The next thing I will look into is fixing copy & paste of dives. I think that's the biggest show-stopper.

atdotde:
Can we ifdef our own definition of some qWarning() that defaults to printf for 5.6?

bstoeger:
You just have to know how to use `qWarning` correctly:

`qWarning("...")` generally works (if `<QtGlobal>` is included, which is pretty much universally the case).
`qWarning() << "..."` works generally for Qt > 5.6 and for Qt <= 5.6 only if `<QDebug>` is included.

At least that's what I got from reading the documentation. So I was at fault for using the second version.

bstoeger:
While looking at the copy & paste code, I found an interesting way to crash the application:
1) Copy dive with a unique dive site.
2) Delete that dive -> dive site unused and will be deleted.
3) Paste dive -> *boom* freed dive site is now referenced again.

Not sure how to best handle that case. The easy way out would seem to be invalidating the "clipboard" if the copied dive gets deleted. But that seems kind of ... cheap? Reminds me of those bad old days in X where you would copy text, close editor, paste -> nothing there!?

dirkhh:
Hey, X was one of the early projects I worked on (and the first project where I became a maintainer...)

Invalidating the clipboard seems better than crashing... a clean solution might be convoluted.

henrik242:
Or never delete dive sites automatically?  I some times create dive sites before a trip (e.g. with gps or descriptions and stuff, so that I can find them), and hook them up to the dive(s) later.  I would hate it if dive sites were deleted automatically (for *any* reason).

(I don't know if this is possible now with Subsurface, though, I stopped using it as my primary dive logger long ago because of the lacking dive site stuff.  But hey, it might be a lot better now!)

bstoeger:
@dirkhh: Off-topic, but you have to admit that that behavior (loss of clipboard on application close) was slightly infuriating and hard to relate to newcomers.

A proper dive site management is on the TODO list. I'm currently quite busy with work related stuff, but I can prioritize it up. First though I want to finish this PR and undo of trip, weightsystem and cylinders.

In any case, this would only move, not solve the problem:
1) Copy dive
2) Manually delete dive site
3) Paste

We need some form of invalidation of the clipboard or copy the full divesite and readd/merge on paste.

dirkhh:
@bstoeger heck, yes. It caught me more than once by surprise and I may have yelled bad words...

I think the correct solution is to NOT delete dive sites until the app shuts down. The memory consumption is unlikely to pose a huge problem, but let's just keep them around with some form of permanent owning pointer and be done with that.

As to @henrik242 's comment... many of us would LOVE to see better dive site management. This has been a source of considerable pain over the last few years. I will devote whatever is necessary from my side to make this happen. And I would suggest that we have a careful and detailed discussion on the mailing list on how this should look before we wade into this. Because we had SO MANY false starts there...

dirkhh:
So I took this to the mailing list. Of course it's late in Europe, but let's see what we get over the weekend... it was a funny coincidence that Doug (whom I know from real life) ended up asking me such a related question just today...

bstoeger:
Alright - implemented undo of paste. Code is not pretty at all, but it light testing it did what it was supposed to do.

bstoeger:
Quickly implemented undo of dive-trip fields.

Open points:

1) No warning if multiple dives edited. Will tackle this in the upcoming days.
2) Weird focus problems. Unclear whether ^Z acts globally or on the current field. This needs a Qt expert.
3) Edit dive profile is now a bit less easy to reach: Context menu on profile. Not a problem for me, but perhaps others?

In a next step, I will work on undo of dive-site editing and a dive site list. Without touching the basic functionality, that is.

For me this PR is at a point where it could do with some testing.

dirkhh:
I did some testing and it didn't eat my cat. I'd love for some other to test as well before merging...

bstoeger:
Please don't merge yet - the dive site stuff in this PR is conflicting with the dive site stuff in 
#1991. I will have to harmonize these two.

dirkhh:
No problem. I would have asked you if you are ready to remove the [WIP] before I do that.
I try really hard to pay attention to that so people can indeed post WIP here without risk of things getting merged pre-maturely.

bstoeger:
All further development is done in #2023 - close.

PR #1980, opened 2019-02-19
Filter: implement any-of mode

bstoeger:
Add an additional mode to the tags, people and location filters: any_of.
Replace the original invert-bool by an enum.
Move the common code into a distinct function.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Implement any-of mode in filters, as suggested by @willemferguson. This is mostly untested.

### Release note:

Nope - part of new filter.
### Documentation change:

I don't think the new filter is already documented.

### Mentions:

@willemferguson: please check with your use-case as described on the mailing list. Thanks.

### DIFF
--- a/qt-models/filtermodels.cpp
+++ b/qt-models/filtermodels.cpp
@@ -24,28 +24,37 @@ namespace {
 				   { return s2.trimmed().contains(s.trimmed(), Qt::CaseInsensitive); } );
 	}
 
-	bool hasTags(const QStringList &tags, const struct dive *d, bool negate)
+	// Check whether either all, any or none of the items of the first list is
+	// in the second list as a super string.
+	// The mode is controlled by the second argument
+	bool check(const QStringList &items, const QStringList &list, FilterData::Mode mode)
+	{
+		bool negate = mode == FilterData::Mode::NONE_OF;
+		bool any_of = mode == FilterData::Mode::ANY_OF;
+		auto fun = [&list, negate](const QString &item)
+			   { return listContainsSuperstring(list, item) != negate; };
+		return any_of ? std::any_of(items.begin(), items.end(), fun)
+			      : std::all_of(items.begin(), items.end(), fun);
+	}
+
+	bool hasTags(const QStringList &tags, const struct dive *d, FilterData::Mode mode)
 	{
 		if (tags.isEmpty())
 			return true;
 		QStringList dive_tags = get_taglist_string(d->tag_list).split(",");
-
-		return std::all_of(tags.begin(), tags.end(), [&dive_tags, negate](const QString &tag)
-				   { return listContainsSuperstring(dive_tags, tag) != negate; } );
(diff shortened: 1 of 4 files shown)


qt-models/filtermodels.cpp
```diff
@@ -24,28 +24,37 @@ namespace {
...
+	// in the second list as a super string.
+	// The mode is controlled by the second argument
+	bool check(const QStringList &items, const QStringList &list, FilterData::Mode mode)
+	{
+		bool negate = mode == FilterData::Mode::NONE_OF;
+		bool any_of = mode == FilterData::Mode::ANY_OF;
+		auto fun = [&list, negate](const QString &item)
+			   { return listContainsSuperstring(list, item) != negate; };
+		return any_of ? std::any_of(items.begin(), items.end(), fun)
+			      : std::all_of(items.begin(), items.end(), fun);
```
dirkhh, line 37: My head just exploded... don't mind me... what the... err... I have no idea what that does... I'm sure it's obvious...
Ok, I can guess what it does, but... oh my...

dirkhh (commented):
First, thanks for implementing that.
Second, everything looks great except for that any/all of not implicit calls another thing lambda that I'm sure is brilliant and efficient but I'm not quite sure I understand...

qt-models/filtermodels.cpp
bstoeger: That's weird, because for the old code you said this is one use of lambdas you can get behind (paraphrasing, not searching the actual comment).

It's exactly the same, just that it now uses `any_of` instead of `all_of` for one of the modes. And instead of typing the lambda twice I just assigned it to a variable.

Perhaps the small part that is missing is that the `auto fun = ...` line assigns a lambda and not the result of a lambda to `fun`? After all, a lambda is just a function pointer + some state. The state being given in square brackets. One has to use `auto` in this case, because lambdas have a private type that is only known to the compiler.

dirkhh: The things that I read and that make sense and the things that I read and... can't quite wrap my mind around aren't logically separated, I guess.
This is a perfectly fine use of a lambda. 
I just struggled looking at the different values possible and understanding that this really does the right thing in all cases. Part of that may have been limited time to look at it (I tried to comment while I had some time and then had to run...)
I.e., I am not asking you to change this, instead I'll think about it some more :-)

dirkhh (approved):
...and just like that, it suddenly makes perfect sense.
I think the disaggregation of a tri-state to two booleans was throwing me off - especially combined with the syntax pieces that I'm less comfortable with.
No, this is correct. Thanks.

bstoeger:
@willemferguson: there should be a combo-box to the left of the fields were you can choose all_of / any_of / none_of. The defaults are of course debatable.

PR #1981, opened 2019-02-20
Filter: change default mode for location to any-of

bstoeger:
Since a dive has only one location all-of makes little sense. It
*can* make sense if the user enters two substrings (e.g. Tofo and Reef),
but generally it won't. Therefore change the default to any-of.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Change default for location-filtering from all-of to any-of. Since a dive can have only one location, the latter seems a much more common use-case than the former.

### Release note:

No.
### Documentation change:

No.
### Mentions:

@willemferguson

### DIFF
--- a/qt-models/filtermodels.h
+++ b/qt-models/filtermodels.h
@@ -44,7 +44,7 @@ struct FilterData {
 	QStringList equipment;
 	Mode tagsMode = Mode::ALL_OF;
 	Mode peopleMode = Mode::ALL_OF;
-	Mode locationMode = Mode::ALL_OF;
+	Mode locationMode = Mode::ANY_OF;
 	Mode equipmentMode = Mode::ALL_OF;
 	bool logged = true;
 	bool planned = true;


dirkhh:
I actually thought about this when you first submitted your pull request. And I thought it was intentional. Because typical location names are "Golden Drop" or "Blue Hole"... so all-of actually seems like the reasonable default... if I enter "Blue Hole" I would be surprised if by default I got all sites with either "Blue" or "Hole" in the name...

@henrik242 - you tend to have strong feelings about our location handling :-) thoughts?

dirkhh:
Ugh - the quotation marks in my previous reply could be misleading... I'm not suggesting that we use those as part of the syntax for the filter widget... they were just intended to indicate the user input...

henrik242:
> if I enter "Blue Hole" I would be surprised if by default I got all sites with either "Blue" or "Hole" in the name...

I would be surprised too.  And slightly annoyed, so I would have come here and reported it as a bug :)

dirkhh:
You are correct - the comma is the magic separator. Which means that this does act as expected and this PR is indeed correct. Thanks for reminding me of the logic :-/

PR #1984, opened 2019-02-23
Append the dive mode as a tag that is seen by the filter tool.

willemferguson:
This allows one to filter dives by divemode, e.g. by typing
"CCR" or "Open circuit" in the tags textbox of the filter tool.
Quite useful if one dives using more than one dive mode. For the
purpose of the filter tool only the dive mode attribute is added
to the list of tags for the specific dive being considered. The
tag list for the same dive (in the XML dive log) is not affected
in any way.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@bstoeger @dirkhh

### DIFF
--- a/qt-models/filtermodels.cpp
+++ b/qt-models/filtermodels.cpp
@@ -42,6 +42,7 @@ namespace {
 		if (tags.isEmpty())
 			return true;
 		QStringList dive_tags = get_taglist_string(d->tag_list).split(",");
+		dive_tags.append(gettextFromC::tr(divemode_text_ui[d->dc.divemode]));
 		return check(tags, dive_tags, mode);
 	}
 


qt-models/filtermodels.cpp
```diff
@@ -42,6 +42,7 @@ namespace {
 		if (tags.isEmpty())
 			return true;
 		QStringList dive_tags = get_taglist_string(d->tag_list).split(",");
+		dive_tags.append(divemode_text_ui[d->dc.divemode]);
```
bstoeger: This should probably be translated by `gettextFromC::tr(divemode_text_ui[d->dc.divemode])`, for now. We can cache the translations later.

willemferguson: Done.

bstoeger (approved):
Code looks good to me. Featurewise, I'll let others decide.

dirkhh:
I only dive OC, so I am not sure this is what people would want, but it sounds reasonable.
I see no harm done here.

PR #1990, opened 2019-03-03
Import: fix infinite loop

bstoeger:
Owing to a variable reuse in a nested loop, importing dive logs
with new trips could lead to an infinite loop. Use a fresh index "j".

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a very stupid bug, which would lead to an infinite loop on import. :(
We really should enhance our tests to test all these corner-cases...

### Release note:

Should probably do, yes.
### Documentation change:

No.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1673,7 +1673,7 @@ void process_imported_dives(struct dive_table *import_table, struct trip_table *
 			    struct dive_table *dives_to_add, struct dive_table *dives_to_remove,
 			    struct trip_table *trips_to_add)
 {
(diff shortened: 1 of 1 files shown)


dirkhh:
Agree on better tests.
I thought we had a test that covered import, bit maybe it's too simplistic to catch this?
Feel free to add a test that would have failed without this fix.

dirkhh (approved):

PR #1991, opened 2019-03-03
[WIP] Make undo-code dive site aware (part 1)

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a first step in including the dive site handling in the undo system.

The most notable change is that it completely does away with the dive site UUIDs. Meaning these are not stable anymore. I don't know how git storage reacts to that (yet). Well, actually we don't change the sort order, so the UUIDs should only be reset once. Still I haven't tested this, so beware.

### Changes made:

1) Don't parse into global dive site table on load
2) Merge dive sites on import
3) Undo dive site changes on import and add dives
4) Reset dive site UUIDs on save
5) Only set dive site UUIDs on save or load

### Related issues:

#1978
### Release note:

After the series.
### Documentation change:

After the series.


bstoeger:
BTW: I wonder if we should combine `struct dive_table`, `struct trip_table` and `struct dive_site_table` into a combined struct (e.g. `struct divlog`?). Quite a few functions take all three tables as argument...

bstoeger:
Of course, the UUID could also be made stable, that would mean slightly more complex code. The way it would work: every time a dive site is added to the global dive site table the UUID is only changed if it clashes with an existing dive.

bstoeger:
@dirkhh: Any comments on making the dive site UUIDs unstable? Will this affect git / cloud storage? Thanks.

bstoeger:
@dirkhh: Related question: When do we actually have to call `invalidate_dive_cache()`?

dirkhh:
Sorry for being completely drowned by a corporate "leadership" event the last couple of days...
So `invalidate_dive_cache()` is needed to let the git logic know that it needs to actually recreate the tree to check if dives have changed. So whenever you change something inside a single dive, this needs to be called. When in doubt, email to Linus.
As for UUIDs... I'm not sure what we are gaining by doing away with them. This will cause massive thrashing in the git repo because the dive sites are all stored under `01-Divesites/Site-<UUID>` and then referenced by UUID in each dive. Why WOULD we want to change that in the storage format?

bstoeger:
> As for UUIDs... I'm not sure what we are gaining by doing away with them. 

It just makes things simpler. One example: you have two diverging logs and edited a dive site in one of them. The new import code will recognize them as different, but they have same UUID. You have to take care of that.

> This will cause massive thrashing in the git repo because the dive sites are all stored under 01-Divesites/Site-<UUID> and then referenced by UUID in each dive.

Understood. I will change the dive site UUIDs to be retained (unless it would generate duplicated UUIDs).

Concerning your mail to the ML: This *is* in preparation of the dive-site rework. But I'm heading off to a conference and have a new lecture to prepare, so I don't find much time currently. Nevertheless, there is some progress.

dirkhh:
Since I clearly suck at expressing myself in an easily understandable way, let me be very clear... I was NOT trying to appear critical of you, @bstoeger - and yes, I realize that this is connected to the dive site rearchitecture. I just noticed that the discussion had faded without a clear "let's do A, B, and then C"...
no implication that this is your job or that you didn't do anything that I expected you to do.

bstoeger:
Oh, I didn't understand the mail as criticism. I was just giving a status update and was too lazy write a mail. :)

dirkhh:
Fun fact... this of course writes to the shared ssrftest account git repo :-)

bstoeger:
Ooops. I hope nothing broke...? I will try to fix the UUID thing soon.

dirkhh:
I don't see any breakage (which in itself is some form of good news?)

bstoeger:
Added ref-counting of dive-sites modeled according to ref-counting of dive-trips.
Scanning the whole dive list to see if a dive site is used is a bit embarrassing. The ref-counts are kept up-to-date but are not yet used.

UUIDs are still trashed, so better don't try this on real logs for now. [Side-thought: stable UUIDs are not a bad idea, because then we can keep the dive-sites sorted by UUID and do quick lookups with binary search]

bstoeger:
Now makes sure that
a) UUIDs are preserved if possible
b) No two dive sites with the same UUID can be in the same table

Does not (yet?) keep the dive site table in ordered-by-UUID state.

I wonder if the same should be done for dives.

bstoeger:
FYI: I added a dive site tab. Yes, that's a silly location, but it was the easiest for me to start with. This can be moved into a dialog or view later.

It's still extremely rudimentary and only used for display. Only some very basic features are implemented:
1) deletion of dive sites
2) editing of names in the table
3) sorting by clicking on the table head

The starting column widths are non-sensical, as this uses the code of `TableView`. The latter only uses the header strings.

When a dive site is deleted, the dives there are set to "no dive site". This is not yet automatically updated in the UI, as this needs infrastructure from #1978.

Currently, this can easily be crashed by editing a name in the table then deleting the dive site by other means and then undo. The classic. Therefore, this is far from ready.

bstoeger:
Update: you can now add a dive site and edit the description directly in the table. And of course undo these actions.

The initial column widths are still broken!

Someone care to give this a try? Warning: still very rudimentary, but I hope it gives an impression of the direction this takes.

dirkhh:
I'll play with it tonight.
Thank you for keeping working on this

bstoeger:
Thanks - but don't be disappointed: there's only a minute or so to play with. :)

bstoeger:
Started undo-izing the "DiveLocationWidget". This still has focus issues (will connect ESC to lose focus as in the dive editing PR) and fetching geo-tags is not yet undoable.

But you can do funky things such as moving the flag on the map and then undoing it. :)

djunkins:
I built this branch and started playing with it. I didn't run into any problems with the undo code or the divesite list functionality as you described it above.

One comment though... In the change to the divesite editing screen, you have removed the "Discard Changes" button. I think it may be useful to have a "Cancel" button that undos everything back to the point that you entered the edit screen, rather then having to repeatedly undo the changes until you get back to where you think you were.

dirkhh:
I didn't find anything that I was able to break - but I found the focus behavior really irritating.
Cmd-Z is restricted to your current text edit until you somehow manage to get focus away from that. And that often took me several tries :-(

bstoeger:
Thanks to both of you for testing this.

> One comment though... In the change to the divesite editing screen, you have removed the "Discard Changes" button. I think it may be useful to have a "Cancel" button that undos everything back to the point that you entered the edit screen, rather then having to repeatedly undo the changes until you get back to where you think you were.

That's a good idea indeed, but more tricky than it sounds at first. With more-and-more undoized actions, it's quite easy to sneak in undo-commands while editing a dive site. Currently e.g. import dives. Cancel would then also undo these commands. Therefore, I'd wait with that until this is all more settled.

> I didn't find anything that I was able to break - but I found the focus behavior really irritating.
Cmd-Z is restricted to your current text edit until you somehow manage to get focus away from that. And that often took me several tries :-(

Yes, that's vexing. I'm not sure if this is our or Qt's fault. I force-pushed changes where I managed to hijack ESC events to steal focus by focusing the MainWindow. Very hackish, but my brain's simply not wired for Qt's event system. Perhaps a Qt/KDE person wants to have a look?

Other new features:
- Implemented undo of reverse geo lookup.
- Implemented undo of dive site merge.
- Add edit button to dive site table.
- Maximize "name" and "description" fields. This should be replaced by proper size management, but for testing purposes it is good enough.

Missing:
- "Purge unused dive sites" functionality.
- Implicit addition of dive site by adding picture with geo-information is not yet supported. You might confuse and even crash the application, if you try to mix this with undo. My goal here would be to not add a new dive site, but the generic GPS information discussed on the ML.
- Switch to dive site page on undo (?). Because currently you may press CTRL-Z and have no idea what actuall changed...!
- Update of dive fields and filter on changes to dive sites.

The last point depends on infrastructure of the dive edit undo PR (#1978). I'm playing with the thought of merging the two to a monster-PR. But I fear that such an enormous PR would bring maintenance issues. What do you think?

dirkhh:
I'm ok with a monster PR. And I'm ok with getting #1978 to a point where we can merge it first. Your call.

bstoeger:
All further development is done in #2023 - close.

PR #1994, opened 2019-03-05
Add support to compile from source on a Raspberry Pi

djunkins:
Signed-off-by: Doug Junkins <junkins@foghead.com>

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [x] Documentation change
- [ ] Language translation

Updating to support compiling from source on a Raspberry Pi. Tested on a 
Raspberry Pi Model B+ running Raspian Stretch and on a Raspberry Pi 3
Model B running Raspian Stretch and Ubuntu Mate.

### Changes made:

1) Update CMakeLists.txt to test if the CMAKE_SYSTEM_PROCESSOR variable is
either armv6l or armv7l. If so, add libGLESv2 to the list of libraries
to be linked to support openGL on embedded systems.

2) Update INSTALL file to list the packages needed to be installed on
Raspian Stretch or Ubuntu Mate to successfully compile subsurface from
source.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,13 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
         # in some builds we appear to be missing libz for some strange reason...
 	# Add ssh2 at the end for openSUSE builds (for recent cmake?)
         set(SUBSURFACE_LINK_LIBRARIES ${SUBSURFACE_LINK_LIBRARIES} -lssh2 -lz -lpthread)
+
+        # Test for ARM processor (Raspberry Pi) and add libGLESv2 if found
+        if (CMAKE_SYSTEM_PROCESSOR STREQUAL "armv7l" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "armv6l")
+                message (STATUS "Found ARM processor. Adding libGLESv2")
+                set(SUBSURFACE_LINK_LIBRARIES ${SUBSURFACE_LINK_LIBRARIES} -lGLESv2)
+                set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+        endif()
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
         execute_process(
                 COMMAND sh scripts/get-version linux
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -186,6 +186,13 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
         # in some builds we appear to be missing libz for some strange reason...
 	# Add ssh2 at the end for openSUSE builds (for recent cmake?)
         set(SUBSURFACE_LINK_LIBRARIES ${SUBSURFACE_LINK_LIBRARIES} -lssh2 -lz -lpthread)
+
+        # Test for ARM processor (Raspberry Pi) and add libGLESv2 if found
+        if (CMAKE_SYSTEM_PROCESSOR STREQUAL "armv7l" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "armv6l")
+                message (STATUS "Found ARM processor. Adding libGLESv2")
+                set(SUBSURFACE_LINK_LIBRARIES ${SUBSURFACE_LINK_LIBRARIES} -lGLESv2)
```
dirkhh, line 193: Is that true for ALL arm v7l/v6l systems? Or should there be an additional test for being on a Raspberry?
I don't know enough about the myriad of ARM systems out there. But let's assume I build on Android system - it seems there we don't need that additional library. (then again, I haven't built Subsurface on Android in a long time, these days all I test is Subsurface-mobile).

dirkhh (commented):
Thanks for your first contribution, Doug.
I have a question below - I simply don't know the answer and would like to find out what you think about this...

djunkins:
Hey Dirk,

My experience with ARM processors is somewhat limited as well. I put the test for the ARM processor inside the elseif that tests that the CMAKE_SYSTEM_NAME is “Linux” to detect a desktop operating system. This was to avoid changing the existing Android build parameters. My understanding is that the GLESv2 library is required for openGL support on any of the ARM processors, but it may be included in the system libraries on the Android.

Two alternatives I see are to write an external test script that will perform a more exhaustive test to detect that the target is a Raspberry Pi, or to have a explicit configuration option to include the GLESv2 library. If you’d like, I can explore those options in more depth.

Thanks for the feedback.

Cheers,

-Doug

dirkhh:
Thanks for the more detailed explanation. I think this is likely to be fine. We haven't seen much interest in ARM builds outside Android and Raspberries. Of course that may change in the not so distant future as we see more ARM laptops show up - but those will be ARM64. So I think this is OK.

PR #1995, opened 2019-03-07
Cmake cleanups

DerDakon:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

Some more cleanups in CMake code.


dirkhh (approved):
I love these. Please keep them coming. This is making our build system better and easier to understand.
Thanks

DerDakon:
Things could be a bit cleaner if a newer CMake version is used. The next step would be 3.1, which allows compiler independent selection of the C/C++ standard used. That would be the same as I already did for the Intel compiler, this would just move before all that compiler if's.

dirkhh:
I believe all of our platforms at this point use at least CMake 3.1
More than happy to bump the minimum version required to some thing like 3.1 (but not MUCH newer as for example Ubuntu 16.04 which we'll support for at least a couple more years has only 3.5.1)

PR #1996, opened 2019-03-08
Add stats by max depth and min temp to yearly statistics widget

djunkins:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Adds breakdown by Max Depth and by Min Temperature to the Yearly statistics widget. Dives are added to the stats buckets based on the maximum depth of the dive and on the minimum temperature of the dive. Buckets are allocated based on the constants STATS_MAX_DEPTH, STATS_DEPTH_BUCKET, STATS_MAX_TEMP, and STATS_TEMP_BUCKET defined in core/statistics.h

### Changes made:

In core/statistics.c:

1) allocate memory for stats_t structures for both depth and temperature and initialize the memory blocks to zeros.

2) Process each dive in the dive list and to add the dive to the appropriate depth or temperature bucket. Dives with depth or temperature that exceed the maximums defined are included in the last bucket.

3) After the dive list is processed, identify the buckets that have at least one dive in them to be displayed by setting is_trip=true.

In qt-models/yearlystatisticsmodel.cpp:

4) Add labels for the depth and temperature buckets in the location field

5) Add a line to the statistics display for each of the buckets that have at least one line added.

### Additional information:

Translation fields have not been updated for the labels included for the depth and temperature statistics, but the units are set based on the unit preferences.

### DIFF
--- a/core/statistics.c
+++ b/core/statistics.c
@@ -14,6 +14,7 @@
 #include "display.h"
 #include "divelist.h"
 #include "statistics.h"
+#include "units.h"
 
 static void process_temperatures(struct dive *dp, stats_t *stats)
 {
@@ -95,6 +96,7 @@ char *get_minutes(int seconds)
 void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 {
 	int idx;
+	int t_idx, d_idx, r;
 	struct dive *dp;
 	struct tm tm;
 	int current_year = 0;
@@ -104,7 +106,7 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 	int prev_month = 0, prev_year = 0;
 	int trip_iter = 0;
 	dive_trip_t *trip_ptr = 0;
-	unsigned int size, tsize;
+	size_t size, tsize, dsize, tmsize;
 	stats_t stats = { 0 };
 
 	if (dive_table.nr > 0) {
@@ -119,17 +121,24 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 
 	size = sizeof(stats_t) * (dive_table.nr + 1);
 	tsize = sizeof(stats_t) * (NUM_DIVEMODE + 1);
+	dsize = sizeof(stats_t) * ((STATS_MAX_DEPTH / STATS_DEPTH_BUCKET) + 1);
+	tmsize = sizeof(stats_t) * ((STATS_MAX_TEMP / STATS_TEMP_BUCKET) + 1);
 	free_stats_summary(out);
 	out->stats_yearly = malloc(size);
 	out->stats_monthly = malloc(size);
 	out->stats_by_trip = malloc(size);
 	out->stats_by_type = malloc(tsize);
-	if (!out->stats_yearly || !out->stats_monthly || !out->stats_by_trip || !out->stats_by_type)
+	out->stats_by_depth = malloc(dsize);
+	out->stats_by_temp = malloc(tmsize);
+	if (!out->stats_yearly || !out->stats_monthly || !out->stats_by_trip ||
+		  !out->stats_by_type || !out->stats_by_depth || !out->stats_by_temp)
 		return;
 	memset(out->stats_yearly, 0, size);
 	memset(out->stats_monthly, 0, size);
 	memset(out->stats_by_trip, 0, size);
 	memset(out->stats_by_type, 0, tsize);
+	memset(out->stats_by_depth, 0, dsize);
+	memset(out->stats_by_temp, 0, tmsize);
 	out->stats_yearly[0].is_year = true;
 
 	/* Setting the is_trip to true to show the location as first
@@ -144,6 +153,12 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 	out->stats_by_type[3].is_trip = true;
 	out->stats_by_type[4].location = strdup(translate("gettextFromC", divemode_text_ui[FREEDIVE]));
 	out->stats_by_type[4].is_trip = true;
+
+	out->stats_by_depth[0].location = strdup(translate("gettextFromC", "All (by max depth stats)"));
+	out->stats_by_depth[0].is_trip = true;
+
+	out->stats_by_temp[0].location = strdup(translate("gettextFromC", "All (by min. temp stats)"));
+	out->stats_by_temp[0].is_trip = true;
 
 	/* this relies on the fact that the dives in the dive_table
 	 * are in chronological order */
@@ -173,6 +188,30 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 
 		process_dive(dp, &(out->stats_by_type[dp->dc.divemode + 1]));
 		out->stats_by_type[dp->dc.divemode + 1].selection_size++;
+
+		/* stats_by_depth[0] is all the dives combined */
+		out->stats_by_depth[0].selection_size++;
+		process_dive(dp, &(out->stats_by_depth[0]));
+
+		d_idx = dp->maxdepth.mm / (STATS_DEPTH_BUCKET * 1000);
+		if (d_idx < 0)
+			d_idx = 0;
+		if (d_idx >= STATS_MAX_DEPTH / STATS_DEPTH_BUCKET)
+			d_idx = STATS_MAX_DEPTH / STATS_DEPTH_BUCKET - 1;
+		process_dive(dp, &(out->stats_by_depth[d_idx + 1]));
+		out->stats_by_depth[d_idx + 1].selection_size++;
+
+		/* stats_by_temp[0] is all the dives combined */
+		out->stats_by_temp[0].selection_size++;
+		process_dive(dp, &(out->stats_by_temp[0]));
+
+		t_idx = ((int)mkelvin_to_C(dp->mintemp.mkelvin)) / STATS_TEMP_BUCKET;
+		if (t_idx < 0)
+			t_idx = 0;
+		if (t_idx >= STATS_MAX_TEMP / STATS_TEMP_BUCKET)
+			t_idx = STATS_MAX_TEMP / STATS_TEMP_BUCKET - 1;
+		process_dive(dp, &(out->stats_by_temp[t_idx + 1]));
+		out->stats_by_temp[t_idx + 1].selection_size++;
 
 		if (dp->divetrip != NULL) {
 			if (trip_ptr != dp->divetrip) {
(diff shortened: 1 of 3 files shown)


core/statistics.c
```diff
@@ -104,7 +108,7 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 	int prev_month = 0, prev_year = 0;
 	int trip_iter = 0;
 	dive_trip_t *trip_ptr = 0;
-	unsigned int size, tsize;
+	unsigned int size, tsize, dsize, tmsize;
```
bstoeger: Idle comment: It's in the original code, but "unsigned int" seems like a strange choice to me. `size_t` [type of `sizeof(x)`] or `int` would be more logical to me.

core/statistics.c
```diff
@@ -119,17 +123,24 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
...
 	free_stats_summary(out);
 	out->stats_yearly = malloc(size);
 	out->stats_monthly = malloc(size);
 	out->stats_by_trip = malloc(size);
 	out->stats_by_type = malloc(tsize);
-	if (!out->stats_yearly || !out->stats_monthly || !out->stats_by_trip || !out->stats_by_type)
+	out->stats_by_depth = malloc(dsize);
+	out->stats_by_temp = malloc(tmsize);
+	if (!out->stats_yearly || !out->stats_monthly || !out->stats_by_trip ||
+		  !out->stats_by_type || !out->stats_by_depth || !out->stats_by_temp)
```
bstoeger, line 134: Also in original code: We're in a OOM situation here so this will never happen. But shouldn't this do something like
```
free_stats_summary(out);
```
?

core/statistics.c
```diff
@@ -174,6 +191,26 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 		process_dive(dp, &(out->stats_by_type[dp->dc.divemode + 1]));
 		out->stats_by_type[dp->dc.divemode + 1].selection_size++;
 
+		/* stats_by_depth[0] is all the dives combined */
+		out->stats_by_depth[0].selection_size++;
+		process_dive(dp, &(out->stats_by_depth[0]));
+
+		d_idx = dp->maxdepth.mm / (STATS_DEPTH_BUCKET * 1000);
+		if (d_idx < 0) { d_idx = 0; }
```
bstoeger: Here and below:
Coding style: no braces for single statement `if`, `while`, `for`.

core/statistics.c
```diff
@@ -207,6 +244,24 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 		prev_month = current_month;
 		prev_year = current_year;
 	}
+
+	/* add labels for depth ranges up to maximum depth seen */
+	if (out->stats_by_depth[0].selection_size) {
+		d_idx = out->stats_by_depth[0].max_depth.mm;
+		if (d_idx > STATS_MAX_DEPTH * 1000) { d_idx = STATS_MAX_DEPTH * 1000; }
+		for (r=0; r * (STATS_DEPTH_BUCKET*1000) < d_idx; ++r) {
```
bstoeger: Here and below:
Coding style: space before and after operators (`+`, `*`, `=`).

qt-models/yearlystatisticsmodel.cpp
```diff
@@ -225,4 +227,42 @@ void YearlyStatisticsModel::update_yearly_stats()
 		rootItem->children.append(item);
 		item->parent = rootItem.get();
 	}
+
+	/* Show the statistic sorted by dive depth */
+	if (stats.stats_by_depth != NULL && stats.stats_by_depth[0].is_trip == true &&
+			stats.stats_by_depth[0].selection_size) {
+		YearStatisticsItem *item = new YearStatisticsItem(stats.stats_by_depth[0]);
```
bstoeger, line 233: Idle comment: Please don't take this as an example how memory management is done in C++ ☹
But it's consistent with the rest of the code, so OK.

bstoeger:
Apart from a few stylistic issues, this looks good to me.

A note on the original code: It seems to me that
1) This uses a new `stats_summary` object on every invocation.
2) Never calls `free_stats_summary`

So it leaks memory. And yes, that's me to blame: 97991e2b9fff4254c2b40417bf6d7496ba0d849f

djunkins:
Thanks for the feedback. I will fix the code style issues and add the free_stats_summary call where appropriate and update.

dirkhh:
I'm not sure I understand what you are trying to tell me here:

> Translation fields have not been updated for the labels included for the depth and temperature statistics, but the units are set based on the unit preferences.

core/statistics.c
dirkhh: ... and with very rare exceptions, we prefer the conditionally executed block on the next line.
So this should be
```
if (d_idx < 0)
        d_idx = 0;
```

core/statistics.c
```diff
@@ -15,6 +15,8 @@
 #include "divelist.h"
 #include "statistics.h"
 
+#include <stdio.h>
```
dirkhh: I believe this is only needed for `asprintf()`

core/statistics.c
```diff
@@ -207,6 +226,14 @@ void calculate_stats_summary(struct stats_summary *out, bool selected_only)
 		prev_month = current_month;
 		prev_year = current_year;
 	}
+
+	/* add labels for depth ranges up to maximum depth seen */
+	for (dr=0; dr * 10000 <= out->stats_by_depth[0].max_depth.mm; ++dr)
+	{
+		asprintf(&label, "%dm - %dm", dr * 10, (dr+1) * 10);
```
dirkhh: ... and didn't we agree not to use that but to use our version?

dirkhh (changes requested):
Hey @djunkins -- you are getting a bit of a rough ride here... please don't think we are picking on you.
There are quite a few other coding style issues here. Instead of forcing you through a line by line explanation of all the ways in which we like things... most of them can be done automatically using Qt Creator - the xml file to use as coding style definition for that is mentioned in our CodingStyle document. Happy to help if necessary.
I won't hold up the PR for coding style issues, but I thought it would be worth looking into `asprinft()` one more time...

core/statistics.c
djunkins: @dirkhh I think this was an outdate version of statistics.c. The next commit moved the label creation out of statistics.c and into qt-models/yearlystatisticsmodel.cpp where I could do away with the asprintf altogether and use the get_depth_string and get_temperature_string functions defined in qthelper.cpp to create QString objects.

core/statistics.c
djunkins: Agreed. This was an artifact to when I was using asprintf. I will remove it.

djunkins:
@dirkhh Thanks for the feedback -- no problem with the rough ride. TBH, my coding is a bit rusty and this is a good exercise to help me understand how to best contribute to this project. I will look into using QTCreator to help with the coding style and will review the coding style document.

I'll make the changes suggested and update the pull request.

djunkins:
@dirkhh Regarding the comments about the translation fields, I meant that I have not added anything into the foreign language translation (.ts) files for the strings I used in the labels, so when the language is set to something other than English, the labels are still in English. I just wanted to be explicit that someone else would need to add the string translations.

djunkins:
> Apart from a few stylistic issues, this looks good to me.
> 
> A note on the original code: It seems to me that
> 
> 1. This uses a new `stats_summary` object on every invocation.
> 2. Never calls `free_stats_summary`
> 
> So it leaks memory. And yes, that's me to blame: [97991e2](https://github.com/Subsurface-divelog/subsurface/commit/97991e2b9fff4254c2b40417bf6d7496ba0d849f)

@bstoeger Reviewing this more, I don't think it is actually leaking memory. When calculate_stats_summary is called, it first calls free_stats_summary(out) to free the previously allocated space before calling malloc for each type of stat. I think this means that, while it is retaining the memory used to create the stats summary after the stats widget is closed, it is only keeping one block of memory rather than consuming more memory each time the statistics are displayed.

This may still not be ideal, but it is not as bad as continue to use more memory each time the stats are displayed.

bstoeger:
> Reviewing this more, I don't think it is actually leaking memory. 

Indeed, it doesn't leak, but not for the reason you state. :)

The thing is that this doesn't simply use `struct stats_summary`. It uses `stats_summary_auto_free`, which is defined in `statistics.h` and has a constructor and a destructor taking care about initializing and freeing. That's pretty much an iodiomatic C++ wrapper around C. Apparently I wrote that code and totally forgot about it. Heh.

core/statistics.c
bstoeger: Answering to my self here: No, not needed. The caller will do that using the C++ wrapper.

djunkins:
@bstoeger Got it. That begs the question, then, whether the free_stats_summary(out) that is called at the beginning of calculate_stats_summary() should be removed.

bstoeger:
I don't think so, because it enables reuse of  a `stats_summary` object. We don't do this at the moment, but why not...?

qt-models/yearlystatisticsmodel.cpp
djunkins: For my education and if you have free cycles, can you expand on why this is a bad example of C++ memory management?

dirkhh (commented):
Ahh, I see.
One of the thing we like to do is rewrite the local history when we develop code so that what goes upstream is actually clean and as intended on every commit. Right now you have older commits that introduce functions (like `asprintf()`) or coding style issues which you then fix in later commits.
Once something is in upstream, we of course never rewrite history - but during development it often makes sense to go back and clean things up a bit more.
That is a somewhat more advanced use of `git`, but it's not all that complicated.
We have a couple of options
(a) I pull things as is and we live with the back and forth in the history
(b) I pull things manually and clean it up locally and then push a slightly prettier version
(c) I can help you in doing the history rewrite yourself in your working tree
Happy to do either - let me know what you prefer.

djunkins:
Oops. I missed that step in contributing.

If you tell me the easiest way to rewrite history in my working tree, I will do that. Since there are only three modified files, I could probably just do it by saving them to a temp directory and creating a new branch from master, but if there is a better way to do it in the future, I'd like to learn.

Does this also mean closing the pull request and opening a new one once I have pushed the cleaned up version?

dirkhh:
There are many, many ways of doing this. With such a small set of changes, what I would do is create a new branch
`git checkout -b redoThis`
and then do a
`git reset HEAD~6`
which gives you master with all of your changes in it.
Now you can use
`git add -p <filename>`
to select your changes that you want to have in the first commit
`git commit -s`
to create that commit and message
repeat until you have a reasonable number of commits that make sense for what you want to do.
For this case it would be almost reasonable to simply do ONE commit with everything in it. Or you could break it down in the backend part and the UI part. Or many other way.
Please ask if this was too terse an explanation :-)

djunkins:
@dirkhh I have pushed a cleaned up commit to my repository as a new branch. Do I need to create a new pull request?

dirkhh:
Ahh, forgot the last step in the description.
Now force push the new branch to the remote old branch:
`git push -f origin redoThis:stats-update`
and then this PR will update with the new commits

mturkia:
On Sat, Mar 9, 2019 at 8:33 PM Doug Junkins <notifications@github.com>
wrote:

> @dirkhh <https://github.com/dirkhh> Regarding the comments about the
> translation fields, I meant that I have not added anything into the foreign
> language translation (.ts) files for the strings I used in the labels, so
> when the language is set to something other than English, the labels are
> still in English. I just wanted to be explicit that someone else would need
> to add the string translations.
>

No need to touch the .ts files. Translations are marked within the sources,
and then the .ts files are generated via transifex (web based translation
tool). CodingStyle.md mentions the tr() function that is used within the
sources.

bstoeger:
I think all comments by @dirkhh were addressed. Let's commit - fine adjustments can be done later.

This probably deserves a CHANGELOG.md entry?

djunkins:
Is the CHANGELOG.md entry something you need me to write and commit, or is taken from the pull request automatically?

bstoeger:
Simply add a line at the top of CHANGELOG.md. Example: 3502afad10055117fa9bb265d344f5e2c5ad7b9a

PS: does this perhaps also need a documentation update?

djunkins:
I've updated CHANGELOG.md and pushed it to the my djunkins:stats-update branch. My GitHub experience level is low, but I think that since this pull request was already closed I will need to do another one for the change. Alternatively, I can generate a patch file if that is easier. Which is preferred?

dirkhh:
A new pull request is the easiest from a project perspective

PR #1997, opened 2019-03-10
Desktop: ensure cloud storage email address is all lower case

dirkhh:
We already do that on mobile and I was certain we used to do this for
desktop as well, but apparently that got lost somewhere...

This should solve the problems we are seeing for people with mixed case
email addresses.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

Simply force the email string to contain no upper case letters

### Release note:

done

### Documentation change:

not needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Desktop: make sure cloud storage email addresses are lower case only
 - Desktop: Fix editing of dive-time [#1975]
 - Bluetooth: only show recognized dive computers by default
 - Desktop: Add export option for profile picture [#1962]
--- a/desktop-widgets/preferences/preferences_network.cpp
+++ b/desktop-widgets/preferences/preferences_network.cpp
@@ -53,7 +53,7 @@ void PreferencesNetwork::syncSettings()
 	proxy->set_proxy_user(ui->proxyUsername->text());
 	proxy->set_proxy_pass(ui->proxyPassword->text());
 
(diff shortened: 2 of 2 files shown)


dirkhh:
I'd appreciate if one of the other Collaborators would do a sanity check and merge this

atdotde (commented):
Shall we enforce that everywhere before sending? Like converting also stored addresses to lower and even more on the receiving end?

dirkhh:
We are already forcing things to be lower case of the server.
The problem is the branch name is derived from the email. And that's why people can't read their data.

dirkhh:
So this is converting the address to lower as it is extracted from the UI.
Maybe I'm not understanding your comment, @atdotde - where else would you want to make sure that we convert it?

atdotde:
I wasn't really aware of how this is handled: I meant, we take cloud_storage_email and cloud_storage_email_encoded from the preferences (in particular in getCloudUrl() in qthelper.c) and thought it might be a good idea to force lower case in those places as well. But I am no longer sure this is such a good idea as it might collide with branch names of checked out repositories.

dirkhh:
I don't want to risk breaking things for people who have working setups. So this is designed to only behave differently if someone adds new cloud credentials.
I have looked through this again and am reasonably certain it's correct.
Merging.

PR #2000, opened 2019-03-14
Map: create correct index in updateMapLocationCoordinates()

bstoeger:
When updating the coordinates of a dive site, the MapLocationModel
is updated. The code created a (col, row) index with col = 0.
[The idea of course being col = x, row = y]. Alas, that's not
how Qt works - its models want (row, col) indices. The code
worked, because the only time when the dive site locations were
updated was in dive site edit mode, when only one site is visible,
i.e. there is only one row leading to the correct (0, 0) index.

Fix this so that we can also change dive site positions if more
than one site is displayed.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Found a funny big when playing with dive site undo. Description is in the commit message.

### Changes made:

1) Fix bug in map code, that was hidden up to now.
### Release note:

No, not user visible.
### Documentation change:

No, not user visible.

### DIFF
--- a/qt-models/maplocationmodel.cpp
+++ b/qt-models/maplocationmodel.cpp
@@ -155,7 +155,7 @@ void MapLocationModel::updateMapLocationCoordinates(const struct dive_site *ds, 
 	foreach(location, m_mapLocations) {
 		if (ds == location->divesite()) {
 			location->setCoordinateNoEmit(coord);
-			emit dataChanged(createIndex(0, row), createIndex(0, row));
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
Cute. Because y,x is how all of us think about coordinates?
Great catch

dirkhh:
I wonder if there are other instances of similar bugs in the code...

PR #2006, opened 2019-03-17
Another set of lgtm fixes

dirkhh:

- [ x Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are less obviously bugs and more potential cause for confusion.
Still seemed worth fixing.


Subsurface-divelog:
This pull request **fixes 11 alerts** when merging 38bf9187f2aaf0e7c91aa38c3fd82e242979f4dd into 412c242c10bba270314fe0bee9a14a0547f97014 - [view on LGTM.com](https://lgtm.com/projects/g/Subsurface-divelog/subsurface/rev/pr-7dea683952e81869880b50bb2d7efd2f32ea7541)

**fixed alerts:**

* 9 for Local variable hides global variable
* 2 for Multiplication result converted to larger type

---

core/datatrak.c
```diff
@@ -32,12 +32,12 @@ static unsigned long four_bytes_to_long(unsigned char x, unsigned char y, unsign
 	return ((long)x << 24) + ((long)y << 16) + ((long)z << 8) + (long)t;
 }
 
-static unsigned char *byte_to_bits(unsigned char byte)
+static unsigned char *byte_to_bits(unsigned char byte_value)
```
bstoeger, line 35: Hmm.. I generally find these fixes of not too much value, because I don't see how such a short function would be confusing.

In this case though, wouldn't it make *much* more sense to simply remove the global `byte` variable?

[Rant: It's really vexing that you can't return `char [8]` in C or C++.]

dirkhh: I agree that your approach is much better - remove the global.

The value of static checkers has been hotly debated for 20+ years. I finally found one that can be reasonably integrated with github and it seemed a relatively small price to pay to get a bit more robustness in place.

BTW: there are a couple Alerts in its findings that I would love for you to take a look.
https://lgtm.com/projects/g/Subsurface-divelog/subsurface/alerts/?mode=list
e.g. "Redefined default parameter", "Resource not released in destructor", "Virtual call from constructor to destructor"
I am happy to ignore some of the warnings (like the ones about FIXME comments), but I think it's worth while to at least look at the others and understand if they represent an actual concern or not.
Thanks :-)

atdotde: I am currently looking at some of the alerts, not the ones you mentioned, though...

dirkhh: to avoid duplicating work, feel free to point out what you are looking at :-)

atdotde:
I am looking at Large object passed by value, Declaration hides parameter, FIXME comment: This should be configurable, FIXME comment: This needs a function to find the divemode at the end of the dive like in

PR #2007, opened 2019-03-18
Minor cleanups in datatrak.c

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Minor code cleanups, mostly removal of global variables.

Warning: I have no device of this kind, so can't test!

By the way: I believe that datatrak.h should be merged into datatrak.c, but I'm leaving it as is. It's only included once, so it doesn't matter anyway.

### Changes made:

1) Remove global variables.
2) Remove complex bit field handling code.

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/datatrak.c
+++ b/core/datatrak.c
@@ -16,12 +16,6 @@
 #include "file.h"
 #include "ssrf.h"
 
-unsigned char tmp_1byte, *byte;
-unsigned int tmp_2bytes;
-char is_nitrox, is_O2, is_SCR;
-unsigned long tmp_4bytes;
-long maxbuf;
-
 static unsigned int two_bytes_to_int(unsigned char x, unsigned char y)
 {
 	return (x << 8) + y;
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
I am always amazed how willing you are to jump in and fix code.
This looks good to me, but I would really hope we can find someone with such a device so we can see it tested...

core/datatrak.c
```diff
@@ -26,13 +27,9 @@ static unsigned long four_bytes_to_long(unsigned char x, unsigned char y, unsign
 	return ((long)x << 24) + ((long)y << 16) + ((long)z << 8) + (long)t;
 }
 
-static unsigned char *byte_to_bits(unsigned char byte_value)
+static bool bit_set(unsigned char byte, int bit)
```
atdotde, line 30: Does the compiler figure this is a good candidate for inlining? What about making the arguments static (once you start worrying about these kinds of things...)?

atdotde:
Looks good to me, as long as the compiler does not complain about undefined stuff...

core/datatrak.c
bstoeger: > Does the compiler figure this is a good candidate for inlining?

Compiles down to a single instruction for me (gcc with -O2):

```
        .loc 1 32 2 discriminator 14 view .LVU462
        .loc 1 361 5 is_stmt 0 discriminator 14 view .LVU463
        testb   $4, %r15b
        jne     .L238
```

dirkhh: hehe - I'd consider that fairly decent optimization :-)

salvadorcunat:
I've been testing on some old datatrak files and everything works fine (no crashes and the resulting .xml files are exactly the same imported with current master) so thumbs up for me.
Best regards.

PR #2015, opened 2019-03-19
let CMake select the language level compiler flags

DerDakon:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Require CMake 3.1, which allows selecting the correct compiler flags for C++11 and C99 in a compiler independent way.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,8 @@
 # Uncomment his to see all commands cmake actually executes
 # set(CMAKE_VERBOSE_MAKEFILE ON)
 
-project(Subsurface)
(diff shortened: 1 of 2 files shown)


dirkhh (commented):
Thanks for your continued work on improving our cmake environment.
Let's see if the switch in ordering in the second commit makes Travis happy!

DerDakon:
The fix is not the switch, but the
```cmake
set(PKG_CONFIG_USE_CMAKE_PREFIX_PATH FALSE)
```

PR #2016, opened 2019-03-19
Cleanup: Pass const-reference instead of pointers

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As suggested by @dirkhh, let's make the recent LGTM warning fixes a bit more idiomatic C++ by passing const-references instead of pointer.


dirkhh (approved):
Thanks - I think it's good to see how this /should/ be done.

PR #2018, opened 2019-03-20
Core: simplify ConnectionListModel

dirkhh:
The complicated setup with the AddressRole is unnecessary. All we want to be
able to do is get the index of a specific text in the list. In hindsight I am
puzzled why I implemented this in such a complex fashion.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [X] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I tried to figure out why I went this route when I created this code two years ago. And I have no idea. The code was complicated and redundant (I did that whole search twice, using two different techniques?). And of course it bothered LGTM - which is of course why I went back to look at it in the first place.
This implementation seems much simpler and painfully obvious.

### Changes made:

1) remove the AddressRole
2) use the existing indexOf method instead of manually iterating the addresses

### Mentions:

@bstoeger you are the person with the good taste when it comes to C++ and (my apologies) Qt. Am I missing something here? Or was the existing code just stupid in the first place?

### DIFF
--- a/core/connectionlistmodel.cpp
+++ b/core/connectionlistmodel.cpp
@@ -7,27 +7,13 @@ ConnectionListModel::ConnectionListModel(QObject *parent) :
 {
 }
 
-QHash <int, QByteArray> ConnectionListModel::roleNames() const
-{
-	QHash<int, QByteArray> roles;
-	roles[AddressRole] = "address";
-	return roles;
-}
-
 QVariant ConnectionListModel::data(const QModelIndex &index, int role) const
 {
 	if (index.row() < 0 || index.row() >= m_addresses.count())
 		return QVariant();
-	if (role != AddressRole)
+	if (role != Qt::DisplayRole)
 		return QVariant();
 	return m_addresses[index.row()];
-}
-
-QString ConnectionListModel::address(int idx) const
-{
-	if (idx < 0 || idx >> m_addresses.count())
-		return QString();
-	return m_addresses[idx];
 }
 
 int ConnectionListModel::rowCount(const QModelIndex&) const
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -287,12 +287,8 @@ QStringList DCDeviceData::getProductListFromVendor(const QString &vendor)
 
 int DCDeviceData::getMatchingAddress(const QString &vendor, const QString &product)
 {
-	for (int i = 0; i < connectionListModel.rowCount(); i++) {
-		QString address = connectionListModel.address(i);
-		if (address.contains(product))
-			return i;
-	}
-	return -1;
+	Q_UNUSED(vendor)
+	return connectionListModel.indexOf(product);
 }
 
 DCDeviceData *DownloadThread::data()
--- a/core/connectionlistmodel.h
+++ b/core/connectionlistmodel.h
@@ -6,13 +6,8 @@
 class ConnectionListModel : public QAbstractListModel {
 	Q_OBJECT
 public:
-	enum CLMRole {
-		AddressRole = Qt::UserRole + 1
-	};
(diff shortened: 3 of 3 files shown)


Subsurface-divelog:
This pull request **fixes 1 alert** when merging cf666c861d9b261b4e728324653a4e501f2a5f2b into b6c7abc1a76f4a05ea564ae994f19d5fc467699a - [view on LGTM.com](https://lgtm.com/projects/g/Subsurface-divelog/subsurface/rev/pr-18fb3c5c131f96747c3301c6c4f9470d4a334c9a)

**fixed alerts:**

* 1 for Redefined default parameter

---

core/connectionlistmodel.cpp
```diff
@@ -7,29 +7,15 @@ ConnectionListModel::ConnectionListModel(QObject *parent) :
...
 		return QVariant();
-	if (role != AddressRole)
+	if (role != Qt::DisplayRole)
 		return QVariant();
 	return m_addresses[index.row()];
 }
 
-QString ConnectionListModel::address(int idx) const
-{
-	if (idx < 0 || idx >> m_addresses.count())
```
bstoeger, line 28: `idx >> m_addresses.count()` ??

dirkhh: Yep, wasn't that brilliant?
I have no excuse.
And of course no static checker ever noticed this perfectly valid statement.

core/connectionlistmodel.cpp
```diff
@@ -7,29 +7,15 @@ ConnectionListModel::ConnectionListModel(QObject *parent) :
 {
 }
 
-QHash <int, QByteArray> ConnectionListModel::roleNames() const
```
bstoeger, line 10: Seems like something that might be used by QML, but a quick grep did not reveal anything accessing "address". Except "btPairedDevice", but that's something different. Can probably be removed.

core/downloadfromdcthread.cpp
```diff
@@ -287,12 +287,8 @@ QStringList DCDeviceData::getProductListFromVendor(const QString &vendor)
...
 int DCDeviceData::getMatchingAddress(const QString &vendor, const QString &product)
 {
-	for (int i = 0; i < connectionListModel.rowCount(); i++) {
-		QString address = connectionListModel.address(i);
-		if (address.contains(product))
-			return i;
-	}
-	return -1;
+	Q_UNUSED(vendor)
+	return connectionListModel.indexOf(product);
```
bstoeger, line 291: I think you replaced a substring by a full-match search. Not sure what is correct here.

dirkhh: look at our implementation of `indexOf()`:
```
int ConnectionListModel::indexOf(QString address)
{
	const QRegExp re(".*" + address + ".*", Qt::CaseInsensitive);
	return m_addresses.indexOf(re);
}
```

core/connectionlistmodel.cpp
```diff
@@ -7,29 +7,15 @@ ConnectionListModel::ConnectionListModel(QObject *parent) :
...
-	roles[AddressRole] = "address";
-	return roles;
-}
-
 QVariant ConnectionListModel::data(const QModelIndex &index, int role) const
 {
 	if (index.row() < 0 || index.row() >= m_addresses.count())
 		return QVariant();
-	if (role != AddressRole)
+	if (role != Qt::DisplayRole)
```
bstoeger, line 14: Old code with that default parameters was very weird - this seems more logical to me.

If it still works, I'm all for it.

bstoeger (approved):
lgtm

core/connectionlistmodel.cpp
dirkhh: I tested it a bit and it seems to work just as well.

PR #2019, opened 2019-03-20
Desktop: KMessageWidget: release resources

dirkhh:
Simply add a trivial destructor to the private class.

Addresses an LGTM warning.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [X] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

simply release the resources created in `init()`

### DIFF
--- a/desktop-widgets/kmessagewidget.cpp
+++ b/desktop-widgets/kmessagewidget.cpp
@@ -40,6 +40,7 @@ class KMessageWidgetPrivate
 {
 public:
     void init(KMessageWidget *);
+    ~KMessageWidgetPrivate();
 
     KMessageWidget *q;
     QFrame *content;
@@ -100,6 +101,13 @@ void KMessageWidgetPrivate::init(KMessageWidget *q_ptr)
     closeButton->setDefaultAction(closeAction);
 
     q->setMessageType(KMessageWidget::Information);
+}
+
+KMessageWidgetPrivate::~KMessageWidgetPrivate()
+{
+    delete iconLabel;
+    delete textLabel;
+    delete closeButton;
 }
 
 void KMessageWidgetPrivate::createLayout()


dirkhh:
(there are moments when I despair about my stupidity... those first two attempts were a clear sign that I need more coffee)

bstoeger:
Not sure about this one, as it seems to use Qt's object hierarchy (that predates universal usage of smart-pointers):
```
iconLabel = new QLabel(content);
```

Makes `iconLabel` a subobject of `content`. I.e. if content is destroyed, `iconLabel` will be too.

```
content = new QFrame(q)
```

Same here: `content` will be destroyed with the `KMessageWidget`.

Therefore it's well possible that these are false positives.

bstoeger:
The new code will not lead to double free, as the destructor of the private object is run first and will remove the objects from the object hierarchy. Nevertheless: Most probably useless code.

dirkhh:
Yeah, I was surprised by this one. Especially since it only caught two of the three.
I think there's nothing WRONG with this change, it just should be unnecessary.
There's always the balance of taking those checkers seriously vs. silly changes just to silence them.
I think this is just barely a change that's worth doing...

tcanabrava:
bstoeger is right, there's no leaks there.
the way qt deals with object destruction is quite simple:

for(auto child : children) {
    child->deleteLater();
}

each child will be deleted before the parent is deleted.

dirkhh:
What I'd want your help with is the one remaining "Warning" in the report...
https://lgtm.com/projects/g/Subsurface-divelog/subsurface/alerts/?mode=list&severity=warning

Call to virtual function settingsChanged which is overridden in DiveCalculatedTissue. 

That one I don't fully understand

bstoeger:
This warning is legit / very bad style - I will look at a proper fix later today or tomorrow.

dirkhh:
>This warning is legit / very bad style - I will look at a proper fix later today or tomorrow.

Thanks

bstoeger:
> There's always the balance of taking those checkers seriously vs. silly changes just to silence them.

Let's not forget the third possibility: fixing the checker. 😀

dirkhh:
> Let's not forget the third possibility: fixing the checker. 😀

Good point. Let's put this on hold while I check in with the LGTM folks

tcanabrava:
That warning is because virtual functions are not virtual in the construtor (the parent is constructed *after* the child), so the call is calling the child, not the parent. *in that case* there's no issue at all, as the parent virtual is empty and it's calling the correct function.

dirkhh:
But why WOULDN'T we explicitly call the child function?

Subsurface-divelog:
This pull request **fixes 2 alerts** when merging 3cbeec9668ae9c1852fb660a5af92ec29f3ab7e5 into c69ca4df80c9c74aa842b7f1fb3c44b22ae3232e - [view on LGTM.com](https://lgtm.com/projects/g/Subsurface-divelog/subsurface/rev/pr-23ba5598c9cf18493e1d2becbb733dd0018dd0d5)

**fixed alerts:**

* 2 for Resource not released in destructor

---

tcanabrava:
We can, it will silence the warning and the result will be the same. it is probably a mistake when I wrote the code.

bstoeger:
One might also ask why we need a virtual function at all? It seems to be only used in three (out of numerous) cases - these particular cases could connect individually to the signal.

So it is a bit "smelly". But simply adding the suggested "Class::" qualifier (perhaps with a one-liner comment) is fine with me.

dirkhh:
As for the issue we started out with (the deletion of resources in the destructor), feel free to comment over here as well: https://discuss.lgtm.com/t/false-positive-for-qt-object-hierarchy/1922

bstoeger:
You said everything there is to say. :)

I can see how static checkers will have a problem with this: How can they know that in `QSomeRandomWidget x(this)` the argument is actually passed through to `QObject` if the definition of `QSomeRandomWidget`'s constructor is not available? I fear this will either have to work with a database of classes or heuristics.

For better or worse, the whole QObject system is showing its age.

dirkhh:
> You said everything there is to say. :)

Yes, I parroted well. But now there's a follow up question and there I need an actual expert...

dirkhh:
now fixed in LGTM, closing

PR #2023, opened 2019-03-22
Dive site and dive edit undo

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a combination of #1978 and #1991. For more details, see there. This is mostly only compile tested - beware!

### Changes made:

1) Make dive site manipulations undoable.
2) Make editing in the main dive-tab undoable.
### Related issues:

#1978 and #1991
### Release note:

Will do.
### Documentation change:

Perhaps...?

@dirkhh: is there a way to test-compile all commits in a PR? Rebasing was a bit hairy and I'm not convinced that I got it right for all commits.


dirkhh:
Sweeeeet 
82 commits. I love it.
I really appreciate your work on this.
It will take me a bit of time to go through this and provide a review - so please don't take "no immediate response" as a sign that this isn't exactly what I was hoping that you would do.

Subsurface-divelog:
This pull request **introduces 3 alerts** when merging 1e9269f25fe6ed33dabc5ff4be37ccf5403c016a into 8facd937f86d913d94275abb80bfa4d2b181bbc6 - [view on LGTM.com](https://lgtm.com/projects/g/Subsurface-divelog/subsurface/rev/pr-be0bef98d22b28d78fed9a5f1360046bad364918)

**new alerts:**

* 2 for Declaration hides parameter
* 1 for Missing header guard

---

core/dive.h
```diff
@@ -746,12 +747,12 @@ extern void average_max_depth(struct diveplan *dive, int *avg_depth, int *max_de
  * Note: we have to use the typedef "dive_table_t" instead of "struct dive_table",
  * because MOC removes the "struct", but dive_table is already the name of a global
  * variable, leading to compilation errors. Likewise for "struct trip_table" and
- * "struct dive_site_table". */
+ * "struct dive_site_table" (defined in "divesite.h"). */
+#include <QObject>
```
dirkhh, line 754: why do we suddenly need to include  `QObject` here?

core/divesite.c
```diff
@@ -382,6 +382,29 @@ void purge_empty_dive_sites(struct dive_site_table *table)
 	}
 }
 
+void add_dive_to_dive_site(struct dive *d, struct dive_site *ds)
+{
+	int idx;
+	if (d->dive_site == ds)
+		return;
+	if (d->dive_site)
+		fprintf(stderr, "Warning: adding dive that has dive site set to dive site\n");
```
dirkhh: that is an, err, interesting message...

dirkhh:
Since stupidly you cannot comment on commit messages themselves, here's a comment in the overall conversation stream to mention issues I have with commits that aren't about lines of code:

https://github.com/Subsurface-divelog/subsurface/pull/2023/commits/d5fc663625718b72ec33a85f241d524072af6eed : typo in the commit message: `global divre site`

https://github.com/Subsurface-divelog/subsurface/pull/2023/commits/625776c534017b4292c89e884d485f98109fe899 : this commit seems to make the compile orgy worse. `dive.h` is now included in `divesite.h`, many more files include `divesite.h` which is likely to slow down our compiles even more. I'm not saying this is wrong - I'm just pointing out that we seem to gravitate towards a model where most source files include "everything"...

core/divesite.c
```diff
@@ -140,6 +140,19 @@ void add_dive_site_to_table(struct dive_site *ds, struct dive_site_table *table)
...
+{
+	struct dive_site *ds;
+	ds = calloc(1, sizeof(*ds));
+	if (!ds)
+		exit(1);
+	ds->uuid = rand() & 0xff;
+	ds->uuid |= (rand() & 0xff) << 8;
+	ds->uuid |= (rand() & 0xff) << 16;
+	ds->uuid |= (rand() & 0xff) << 24;
+	return ds;
```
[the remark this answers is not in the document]
dirkhh: ahh, I see that this is addressed in a later commit

desktop-widgets/tab-widgets/TabDiveSite.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include <qt-models/divecomputerextradatamodel.h>
+
+TabDiveSite::TabDiveSite(QWidget *parent) : TabBase(parent)
+{
+	ui.setupUi(this);
+	ui.diveSites->setTitle(tr("Dive sites"));
+	ui.diveSites->setModel(LocationInformationModel::instance());
+
+	// Show only the first few rows
```
dirkhh: did you mean columns?

core/divesite.c
```diff
@@ -10,6 +10,7 @@
 
 struct dive_site_table dive_site_table;
 
+// TODO: keep table sorted by UUID and do a binary search?
```
dirkhh: Since we don't match by uuid, I'm not sure how that would be reasonable. What am I missing?

dirkhh (commented):
Made it through the first 25 commits, up to 'Coding style: add tab before Q_OBJECT'.
Running out of time here.
So far I haven't seen anything that's wrong. A few nit-picks, a few questions.

bstoeger:
> [625776c](https://github.com/Subsurface-divelog/subsurface/commit/625776c534017b4292c89e884d485f98109fe899) : this commit seems to make the compile orgy worse. `dive.h` is now included in `divesite.h`, many more files include `divesite.h` which is likely to slow down our compiles even more. I'm not saying this is wrong - I'm just pointing out that we seem to gravitate towards a model where most source files include "everything"...

I was hoping that it was the other way round:
- All translation units that include "divesite.h" also need "dive.h".
- Not all translation units that include "dive.h" need "divesite.h".

If these assumptions are true, then the new way should be a (tiny) improvement, no?
Is there a way to benchmark this?

core/divesite.h
```diff
@@ -24,10 +24,10 @@ struct dive_site
 	struct taxonomy_data taxonomy;
 };
 
-struct dive_site_table {
+typedef struct dive_site_table {
 	int nr, allocated;
 	struct dive_site **dive_sites;
-};
+} dive_site_table_t;
```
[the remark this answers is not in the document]
bstoeger: Sadly, we need this typedef to make `struct dive_site_table` a "Qt Metatype". Which in turn we need to pass `dive_site_table` through QML without defining a silly wrapper class. The reason is that MOC can't handle the fact that there is a global variable called `dive_site_table`.

I believe this to be a defect in Qt, but I'm not going to argue and it wouldn't help for now anyway.

core/divesite.c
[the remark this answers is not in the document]
bstoeger: Yes, this is just temporary.

core/dive.h
bstoeger: `#include <QObject>` is needed for the metatype-macros.
Previously, `dive.h` included `divesite.h` which includes `QObject`. Now that this is no longer the case, `dive.h` has to include `QObject` explicitly.

core/divesite.c
bstoeger: When parsing and when creating new UUIDs, we *do* search by UUID. In this case a binary search might help. It's just that
1) The comment precedes the wrong function.
2) With the later changes the table *does* stay sorted by UUID. Therefore the comment is not correct.

I will check that out.

bstoeger:
> Made it through the first 25 commits, up to 'Coding style: add tab before Q_OBJECT'.
> Running out of time here.

I'll be away the whole next week anyway...!

core/cochran.c
```diff
@@ -791,10 +791,11 @@ static void cochran_parse_dive(const unsigned char *decode, unsigned mod,
 	free(buf);
 }
 
-int try_to_open_cochran(const char *filename, struct memblock *mem, struct dive_table *table, struct trip_table *trips)
+int try_to_open_cochran(const char *filename, struct memblock *mem, struct dive_table *table, struct trip_table *trips, struct dive_site_table *sites)
```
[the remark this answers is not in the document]
bstoeger: Saying that I put any thought into these function arguments would be a lie. Nevertheless, the fundamental difference is that the former functions are dive-site only and therefore `table` is clear from context. The latter are parser-functions which already take two other table-arguments (viz. dive and trip).

I changed the dive-site functions to take a `ds_table`. That was so painful (I'm very clumsy with git) that I'd prefer changing the parser functions with an additional commit. Note that for consistency, we'd also have to change the `trips` parameter to `trip_table`. Which will cause lgtm.com to complain, because `trip_table` is the name of a global variable.

Anyway, I entertain the though to replace the dive-, trip- and site-table triple by a
```
struct dive_log {
        struct dive_table dives;
        struct trip_table trips;
        struct dive_site_table sites;
};
```
 That would simplify these function signatures *and* replace three global variables by one.

core/divelist.c
```diff
@@ -1600,21 +1606,23 @@ bool try_to_merge_trip(struct dive_trip *trip_import, struct dive_table *import_
...
  *	1) Dives to be added
  *	2) Dives to be removed
  *	3) Trips to be added
+ *	4) Dive sites to be added
  * The dives to be added are owning (i.e. the caller is responsible
  * for freeing them).
- * The dives and trips in "import_table" and "import_trip_table" are
- * consumed. On return, both tables have size 0.
+ * The dives, trips and sites in "import_table", "import_trip_table"
+ * and "import_sites_table" are. On return, the tables have size 0.
```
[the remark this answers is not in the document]
bstoeger: Fixed.

core/divesite.c
```diff
@@ -291,6 +315,30 @@ static void merge_string(char **a, char **b)
 	free(s1);
 }
 
+/* Used to check on import if two dive sites are equivalent.
+ * Since currently no merging is performed, be very conservative
+ * and only consider equal dive sites that are exactly the same.
+ * Taxonomy is not compared, as not taxonomy is generated on
```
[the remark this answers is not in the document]
bstoeger: Fixed.

core/divesite.c
bstoeger: Reworder - better?
[This is a shouldn't happen warning - so the fact that it is printed is probably more of a concern than it being cryptic.]

desktop-widgets/tab-widgets/TabDiveSite.cpp
bstoeger: Yes - fixed.

core/divesite.c
bstoeger: Moved it to the correct function.

bstoeger:
Added a quick&dirty fulltext filter on the dive site tab. UI is only a mock up.

bstoeger:
@willemferguson: On the mailing list you noted that it's very hard to merge dive sites.

The last commit in this series add a field on the dive site edit form where you can enter up to which distance you want other dive sites to be listed. These dive sites can then be merged into the current site via the context menu.

Does this help?

I'm travelling, so I currently can't partake in mailing list discussions...

bstoeger:
This series has a major show-stopper:
When manually adding / planning a dive, the fields are set in the current dive, not the new dive. I will have to think about that!

Ultimately I want to get rid of the "edit mode" thing, but I'm not sure we're there yet.

bstoeger:
This PR is getting more and more scary. Pushing FYI. Currently editing equipment crashes.

Other things that will have to be fixed:
- No warning if multiple dives edited.
- No '*' shown if things were edited.
- Audit for `invalidate_dive_cache()` (!!)

Especially the last item gives me a stomach ache.

djunkins:
I haven't had a chance to follow too much of the discussion on this since I was traveling this week, but I built the monster branch and when testing it out noticed that when selecting a dive, details like Buddy are not being loaded into the main window. As you can see from the screenshot below, the buddy is still shown in the dive list, but not in the main window. The same is true for other fields like Divemaster and Suit. Tags and Notes do seem to be populated correctly. Is this by design?

Screen Shot 2019-03-30 at 7 37 04 AM

bstoeger:
No - this is a bug in the most recent commits. I will look into it. Thanks for testing!

bstoeger:
Yeah, that was a misplaced '!' in the last commit - should be fixed now.

desktop-widgets/tab-widgets/TabDiveSite.cpp
```diff
@@ -8,7 +8,10 @@ TabDiveSite::TabDiveSite(QWidget *parent) : TabBase(parent)
 {
 	ui.setupUi(this);
 	ui.diveSites->setTitle(tr("Dive sites"));
-	ui.diveSites->setModel(LocationInformationModel::instance());
+	ui.diveSites->setModel(&model);
```
dirkhh, line 15: looking at just THIS commit, you add the private `model` and set that model, but I can't see us ever setting that member to anything useful (like `LocationInformationModel::instance()`)
This may be in a future commit, but right here this raises alarm bells

qt-models/divelocationmodel.cpp
```diff
@@ -183,6 +169,56 @@ void LocationInformationModel::diveSiteDeleted(struct dive_site *, int idx)
...
+	// Kind of dirty, but less effort.
+	struct dive_site *ds1 = get_dive_site(i1.row(), &dive_site_table);
+	struct dive_site *ds2 = get_dive_site(i2.row(), &dive_site_table);
+	if (!ds1 || !ds2) // Invalid dive sites compare as different
+		return false;
+	switch (i1.column()) {
+	case LocationInformationModel::NAME:
+	default:
+		return QString::localeAwareCompare(QString(ds1->name), QString(ds2->name)) < 0; // TODO: avoid copy
+	case LocationInformationModel::DESCRIPTION: {
```
dirkhh, line 204: sorting by description seems... rather academic?

qt-models/divelocationmodel.cpp
```diff
@@ -183,6 +169,56 @@ void LocationInformationModel::diveSiteDeleted(struct dive_site *, int idx)
...
+		// the names are listed as ascending.
+		return cmp != 0 ? cmp < 0 :
+		       QString::localeAwareCompare(QString(ds1->name), QString(ds2->name)) < 0; // TODO: avoid copy
+	}
+	}
+}
+
+DiveSiteSortedModel::DiveSiteSortedModel()
+{
+	setSourceModel(LocationInformationModel::instance());
```
dirkhh, line 221: so this sets the `DiveSiteSortedModel` - but I don't see how this would propagate into the `TabDiveSite::model`

dirkhh (commented):
Interrupting my current review early because stupid GitHub makes it such a pain to comment on commit messages. I found a couple of easily parsed typos so far, but this one I cannot figure out:
`We probably never show a single column per taxonomy field, but on column cooking something out of all taxonomy fields.` -- that's https://github.com/Subsurface-divelog/subsurface/pull/2023/commits/16eebfc1f0497c0da81c4310bbfaee9d7fff4808
Other than that, small comments, one of which might identify a bug(?).
I love the detailed comments and I'm surprised that at this point I am reading most of this code as if I understood what it was doing :-) :-)

dirkhh (commented):
Further review process. Not finding much to complain about - except that https://github.com/Subsurface-divelog/subsurface/pull/2023/commits/074abbd3277fa35e1ed0f78b2d9fa12cc251625c is one of those commits that still push my limits in reading modern C++ code.
I think I get what it does - but I realize that I would have a hard time trying to fix anything in this code - which is frustrating as a sign of my limitations. And you are commenting in a lot of detail, so that helps.

bstoeger:
> that's [16eebfc](https://github.com/Subsurface-divelog/subsurface/commit/16eebfc1f0497c0da81c4310bbfaee9d7fff4808)

Yes, that was neither English, nor any other language known to mankind. I tried to reword it.

desktop-widgets/tab-widgets/TabDiveSite.cpp
bstoeger: As you noted below, `LocationInformationModel::instance()` is set as source model in `DiveSiteSortedModel`. This should be fine (and it appears to work).

qt-models/divelocationmodel.cpp
bstoeger: Sure, but since the column header is there it should be clickable for consistency reasons, no? Consider it as an example of how to implement this, should someone like to add more columns.

qt-models/divelocationmodel.cpp
bstoeger: The details are implemented by Qt's `QSortFilterProxyModel` class. You just have to set the source model and that's it. Out filter-code could definitely be simplified.

bstoeger:
> Further review process. Not finding much to complain about - except that [074abbd](https://github.com/Subsurface-divelog/subsurface/commit/074abbd3277fa35e1ed0f78b2d9fa12cc251625c) is one of those commits that still push my limits in reading modern C++ code.
> I think I get what it does - but I realize that I would have a hard time trying to fix anything in this code - which is frustrating as a sign of my limitations. And you are commenting in a lot of detail, so that helps.

Note that this commit is mostly just moving code to a common header file. And I want to remove it anyway - after this series.

I'll try to explain:

The original undo code worked with the dive list in trip (i.e. tree) mode. Therefore the modifications were sent in batches per trip. But now, more and more systems are using these notifications and this makes no sense anymore. So my plan is to send the notifications as full lists of dives and only the list in trip mode should partition it. This will make things distinctly simpler and make this crazy code local to the dive list.

qt-models/divelocationmodel.h
```diff
@@ -17,7 +17,7 @@ class LocationInformationModel : public QAbstractTableModel {
 public:
 	// Common columns, roles and accessor function for all dive-site models.
 	// Thus, different views can connect to different models.
-	enum Columns { REMOVE, NAME, DESCRIPTION, NUM_DIVES, COORDS, NOTES, DIVESITE, TAXONOMY, COLUMNS};
+	enum Columns { REMOVE, NAME, DESCRIPTION, NUM_DIVES, LOCATION, NOTES, DIVESITE, TAXONOMY, COLUMNS};
```
dirkhh: There's something weird going on with your merge of the two PRs. This change appears to be in two of the commits... I'm trying to figure out what happened here, but I'm a bit confused staring at what I see in git...
Ah, I see... yeah, this is confusing, but it looks right

dirkhh (commented):
Freaky - something changed in the PR while I was reviewing it... need to stop now and figure out why some of my comments suddenly are to outdated commits... arg

bstoeger:
> Freaky - something changed in the PR while I was reviewing it... need to stop now and figure out why some of my comments suddenly are to outdated commits... arg

Sorry... Probably a force-push by me which invalidated all subsequent commits? In any case, I'm reading all comments.

dirkhh (commented):
OK, done with the review. Confused myself in the middle. But that's par for the course.
The last couple of commits need more time from me - removing the add dive and the change from `displayed_dive` to `current_dive` have deep implications that I'm not sure I have spent enough time on, but I need to stop staring at this right now. I'll get back to it, I guess.

Overall this series is incredible. Such an amount of work and massive improvements.
Now we need to iron out the bugs and get it into master - and I'd like to see it in master, soon, so we can work towards 4.9 (which I keep talking about but then hold off because of the outstanding PRs)

bstoeger:
Thanks for your time. Probably better to do 4.9 *before* this PR? There might be some crazy bugs, especially with respect to non-undoized things like equipment, etc.

desktop-widgets/command_edit.cpp
```diff
@@ -8,16 +8,28 @@
...
+}
+
 template<typename T>
-EditBase<T>::EditBase(const QVector<dive *> &divesIn, T newValue, T oldValue) :
+EditBase<T>::EditBase(T newValue, bool currentDiveOnly) :
 	value(std::move(newValue)),
-	old(std::move(oldValue)),
-	dives(divesIn.toStdVector())
+	dives(getSelectedDives(currentDiveOnly)),
+	current(current_dive)
```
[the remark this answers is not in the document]
bstoeger: This remembers the current dive so that on undo it can be made the current dive again. Note that we don't know which of the selected dives was the current dive.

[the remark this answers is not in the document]
dirkhh: That at the same time is scary and makes sense :-)

bstoeger:
Ok - tried to implement the unsaved-changes as sensible as possible given that we now have a two-kinds of changes (undoable and non-undoable) system.

dirkhh:
This is really clever - I like it.

djunkins:
I like how this is coming along. One more thing I saw was that if I add a new divesite to the list, but don't assign any dives to it, it does not get saved. I think it is lines 886-888 in save-git.c that are preventing this for cloud storage:
`		/* Only write used dive sites */`
`		if (!is_dive_site_used(ds, false))`
`			continue;`


But I can't find where it would be affecting local .ssrf files.

djunkins:
I also don't know if there are compatibility issues with older versions if we start writing unused divesites to the cloud storage.

djunkins:
Looks like it is line 600-605 in save-xml.c that would be preventing unused sites from being saved to the local xml file. The logic is a little different than save-git.c since there is the option to only export the selected fives.

dirkhh:
Yes, today we don't save unused dive sites because we had no other way to purge them. So that code should eventually be removed.

dirkhh:
Loading an unused dive site from storage should not cause any issues.

djunkins:
Removing those checks seems to have worked to save unused dive sites to both .ssrf files and to the cloud storage. I have a patch, but can't figure out how to add it to the comment here, so I'll send it to the mailing list shortly.

Next issue I'm running into when testing is that when I have a dive selected and go to the dive site tab and edit a different dive site, the map window centers on the previously saved location of the dive site being edited, but the "drag position" flag is set at the location of the dive site for the currently selected dive.

djunkins:
Patch to remove the checks for whether a dive site is used before saving.

-Doug

bstoeger:
Thank you - I will add that later (currently at work PC doing paper work).

bstoeger:
> Next issue I'm running into when testing is that when I have a dive selected and go to the dive site tab and edit a different dive site, the map window centers on the previously saved location of the dive site being edited, but the "drag position" flag is set at the location of the dive site for the currently selected dive.

Weirdly, this use-case works for me. But I have another problem:
1) Select a dive.
2) Edit a different dive site.
3) Now a dive with that site is selected.

This is due to how the filter works: If it is activated it selects a dive in its list. When editing a dive site, the filter shows only those dives which are at that site and selects one of them. Not sure how to ideally handle this situation...

djunkins:
@bstoeger Your comment made me realize why it may be the case that I had a different experience -- the dive site I was editing was one I had created that is not associated with any dives. I tried editing a site that has at least one dive associated with it and the flag was positioned properly.

Regarding your use-case, is there a way to store the dive that was selected when you entered the dive site edit mode and restore the selected dive to that when you are done editing?

bstoeger:
> Your comment made me realize why it may be the case that I had a different experience -- the dive site I was editing was one I had created that is not associated with any dives. I tried editing a site that has at least one dive associated with it and the flag was positioned properly.

Nice, I can reproduce that now. Will look into it.

> Regarding your use-case, is there a way to store the dive that was selected when you entered the dive site edit mode and restore the selected dive to that when you are done editing?

Yes, either that or instruct the filter not to change selection when going into dive site edit mode.

djunkins:
@bstoeger Two more things I noticed while testing normal actions with the new dive edit undo code:

1) Changes made on the Notes tab are not getting saved to the cloud storage. If I add a Tag, add a Buddy, edit the Notes, etc. I can save the changes to a local file, but anything I save to Cloud Storage reverts to the old version the next time I load from the cloud. I have checked that the asterisk is appearing in the title bar to indicate a save is needed and that it disappears when I click "Save to cloud storage" .

2) This may be more of a design issue, but if I make changes to a field on the Notes tab, they are lost unless I change focus to another field before saving. In this case, the asterisk does not appear in the title bar until I change focus to a different field. This is especially apparent when editing the notes field on a dive, where the user can't just hit Tab to move to another field before hitting <command-S> or <control-S> to save.

bstoeger:
> Changes made on the Notes tab are not getting saved to the cloud storage.

Yes, I worried about that. This confirms that dives are not saved if `invalidate_dive_cache()` is not called. I added a commit that should take care of that, but a more complete audit is needed. Moreover, the commit should be squashed into the appropriate earlier commits.

> This may be more of a design issue, but if I make changes to a field on the Notes tab, they are lost unless I change focus to another field before saving. In this case, the asterisk does not appear in the title bar until I change focus to a different field. This is especially apparent when editing the notes field on a dive, where the user can't just hit Tab to move to another field before hitting or to save.

In the notes field you can press ESC. Clicking on the menu does automatically remove focus, so that's fine. But pressing CTRL-S does not and the current changes don't get saved. That's certainly not acceptable.

The proper fix would probably be adding an undo-command for every key press, but accumulating the commands as long as a field has focus. The short-term fix might be force-removal of focus on save...?

dirkhh:
I was trying to think through the scenarios in which we might end up with unsaved edits in the notes field. So far Ctrl/CMD-S is the only one I can see. Picking a different dive already removes focus, as does interaction with the menu.
So wouldn't it be the easiest to "simply" remove focus when Ctrl/CMD-S is pressed and ensure that the edits are saved to the dive before then saving the dives storage?
We do NOT want an undo for every keypress.

bstoeger:
> We do NOT want an undo for every keypress.

That was not what I meant. My idea was adding an undo command for the first keystroke but subsequent keystrokes would amend that command (accumulate was the wrong expression as that implies multiple commands). Of course the devil lies in the detail: for notes this is trivial, but for tagged entry fields this might be rather tricky the way it currently works.

Force-removing focus on save is probably the best quick-fix. Will try that when I find time.

djunkins:
@bstoeger It looks like the commit that calls `invalidate_dive_cache()` did the trick for saving changes to fields in the Notes tab to the cloud storage.

I'll keep poking at this and see what other issues I can find. I think it is looking good, though.

desktop-widgets/command_edit.cpp
```diff
@@ -228,6 +228,11 @@ DiveField EditAirTemp::fieldId() const
 void EditWaterTemp::set(struct dive *d, int value) const
 {
 	d->watertemp.mkelvin = value > 0 ? (uint32_t)value : 0u;
+
+	// re-populate the temperatures - easiest way to do this is by calling fixup_dive
+	// TODO: Is this necessary?
+	fixup_dive(d);
+	invalidate_dive_cache(d);
```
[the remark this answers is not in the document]
bstoeger: Removed that comment.

bstoeger:
> @bstoeger Your comment made me realize why it may be the case that I had a different experience -- the dive site I was editing was one I had created that is not associated with any dives. I tried editing a site that has at least one dive associated with it and the flag was positioned properly.

I added two commits that hopefully fix that problem.

This raises the question: do we want to display dive sites without dives or not? Or add an option?

djunkins:
> This raises the question: do we want to display dive sites without dives or not? Or add an option?

Do you mean display the dive sites on the map view, or in the table view of dive sites? I would definitely like to see them in the table view for the "Dive sites" tab, but do not have a strong opinion about whether unused sites should appear on the map.

bstoeger:
I meant whether they should be shown as a flag on the map. Perhaps grayed out?

djunkins:
I think grayed out may be a good alternative.

dirkhh:
We already have two different flags - we could add a third that is visually different enough, maybe a light blue instead of red or something.

dirkhh:
93 commits into this journey... what's on your list before this should be merged into master?
It does have a couple of usability issues, still, but overall it looks like a MASSIVE improvement to me...

djunkins:
Two things I would like to see are:

1) The ability to export dive sites (either all of them or selected dive sites) as an XML file.

2) The ability to import the XML file of dive sites to a different user's instance of Subsurface.

I am willing to work on those myself once the existing stuff is merged into master. With those, I can handle my use case of being able to share dive sites amongst my family when we dive together.

bstoeger:
> what's on your list before this should be merged into master?

I would suggest doing 4.9 first to have a version with the recent bug fixes and without potential catastrophes.
Two points before merging come to mind:
1) fix the error that causes the parsing-test to fail.
2) the multiple-dive-editing warning. Not having that warning sounds like a recipe for disaster to me.

bstoeger:
> I am willing to work on those myself once the existing stuff is merged into master. With those, I can handle my use case of being able to share dive sites amongst my family when we dive together.

If you want to do that, I can write the "QUndoCommand" for importing a list of dive sites.

dirkhh:
> I would suggest doing 4.9 first to have a version with the recent bug fixes and without potential catastrophes.

🤦‍♂️ 
I consider this a key feature of 4.9

So let me change course and do a 4.8.6 instead then.

bstoeger:
> 1. fix the error that causes the parsing-test to fail.

Well, that sucks. The dive site UUIDs are now generated randomly because they are no longer used to check for dive site equivalence. But that is not exactly a good precondition for systematic testing. Seems like I'll have to reinstate the old UUID generation.

bstoeger:
Can someone explain why iOS fails but Android works? Something moc related? I don't have access to that platform.

atdotde:
There is some lib path not found that has hohndel as some subdirectory name.  Maybe Travis doesn’t know about that.

bstoeger:
It seems like `core/subsurface-qt/DiveListNotifier.cpp` isn't built on iOS. Does that use a different build system than the other mobile platforms? mobile-on-desktop builds fine for me. @dirkhh?

dirkhh:
iOS is the only one that uses `qmake`. Please don't get me started. It's frustrating, but it's also not worth the amount of pain and random breakage to try to use `cmake` for the iOS build.
So, you have to _also_ edit https://github.com/Subsurface-divelog/subsurface/blob/master/packaging/ios/Subsurface-mobile.pro

bstoeger:
Wow - it runs through Travis. :)

dirkhh:
> Wow - it runs through Travis. :)

Quick, let's merge it, it MUST be perfect.
Seriously, I wonder at what point we are better off with all this in master.

bstoeger:
Well, I found a funny usability bug:

1) Enter filter and choose some settings
2) Press CTRL-+
3) New dive will not be displayed, because it doesn't match filter. Only the count in the window title is increased.

dirkhh:
That is a fun bug. Could easily be fixed after merging.
I wonder if there's anything in here that would stop us from merging, so stuff that completely breaks usability of master, that might destroy your data file, that eats kittens, etc

dirkhh:
After finishing 4.8.6 I went back to this, built it and played with it. Many ideas what we should add / improve - but I don't see any show stoppers that would prevent us from merging at this stage. But because of the complexity, I'd really like a "yep, go ahead" from @bstoeger

dirkhh (approved):
Lots of discussion, great work over quite a long time - I think it's time to get this into master.

dirkhh:
BTW: one of the next features I'd love to see... when I'm on the Dive sites tab, if I click on a site, the dive list below could be filtered by dive site. Including multi selection support.
But seriously, this is a great patch series.

bstoeger:
Well, things will definitely break, but I think merging will make fixing things easier. Since Easter holidays start next week I should also have a bit more time than the last few. So let's go for it. But I would suggest merging the Qt-cleanups first.

dirkhh:
Merged the Qt cleanups - which created conflicts here, as expected.
I'll wait for you to address those... once you have this branch back to mergeable and Travis-building, I'll merge this.
BTW: did you want to merge Doug's PR #2042 into this branch, or should we merge it into master afterwards?

bstoeger:
> BTW: did you want to merge Doug's PR #2042 into this branch

I would, but I don't know how to do that.

bstoeger:
Added a small filter simplification. Please check it out before merging.

bstoeger:
> BTW: one of the next features I'd love to see... when I'm on the Dive sites tab, if I click on a site, the dive list below could be filtered by dive site. Including multi selection support.

Pushed an attempt at this, please check it out. UI is not very smooth, tbh.

dirkhh:
@bstoeger - you are an animal... and I mean that in the very, very best possible way.
I can't believe you already did an implementation of my humble suggestion.
Travis just completed. I looked at the filter simplification and the new code for dive site selection.
I'll try the latter in a moment and if nothing seems broken, I'll merge this...
After that we should have a conversation on the developer mailing list about the new features / open issues, documentation changes, etc...

dirkhh:
I think this exactly what I was looking for. The dive site list isn't beautiful, but the functionality is really good. Very well done. Thank you.

djunkins:
> > BTW: did you want to merge Doug's PR #2042 into this branch
> 
> I would, but I don't know how to do that.

Would it be easiest if I create a new pull request against the newly merged `master` branch with these changes?

djunkins:
Regarding the PR #2042, I saw that you created a new branch to merge it.

There are some usability issues with the dive site filter around selecting a dive site that has no dives associated with it, but all in all it looks good. I'll try to document the issues around the filter issues and open a bug on it.

dirkhh:
Excellent plan. And yes, I already created #2044 and will close #2042

djunkins:
Issue report #2045 covers the dive site filtering issue. It's a little worse than I realized since it is being triggered when you try to add a new dive site to the table as well.

PR #2024, opened 2019-03-24
random Qt cleanups

DerDakon:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Additional information:
**This drop support for Qt before 5.5.**

It uses the new style connect syntax, which is more efficient as it does not do runtime string lookups. There are 3 places where this seems to have shown real bugs, where no compatible slot exists. Until now this would only cause a runtime warning.


core/connectionlistmodel.cpp
```diff
@@ -21,7 +21,7 @@ int ConnectionListModel::rowCount(const QModelIndex&) const
 	return m_addresses.count();
 }
 
-void ConnectionListModel::addAddress(const QString address)
+void ConnectionListModel::addAddress(const QString &address)
```
bstoeger, line 24: This change seems correct to me.

But be careful with slots - they might cross thread boundaries in which case you might end up with a dangling reference.

Also note that `QString`s have copy-on-write semantics, so copying is very cheap.  That forced COW is a bad idea anyway is a different matter...

DerDakon: No, passing things to a slot in a different thread is safe: it will be copied when it is transferred between the even loops, so when you get the reference passed in it is from a QVariant.toString(), which will not change while you access it. The reference avoids one of this reference count up-/down cycles (which is otherwise cheap).

bstoeger: I'm quite sure that I got crashes when doing the video thing, but I'll take your word for it...

desktop-widgets/configuredivecomputerdialog.cpp
```diff
@@ -344,10 +344,10 @@ void OstcFirmwareCheck::saveOstcFirmware(QNetworkReply *reply)
...
 	dialog->setAutoClose(true);
 	ConfigureDiveComputer *config = new ConfigureDiveComputer();
-	connect(config, SIGNAL(message(QString)), dialog, SLOT(setLabelText(QString)));
-	connect(config, SIGNAL(error(QString)), dialog, SLOT(setLabelText(QString)));
-	connect(config, SIGNAL(progress(int)), dialog, SLOT(setValue(int)));
-	connect(dialog, SIGNAL(finished(int)), config, SLOT(dc_close()));
+	connect(config, &ConfigureDiveComputer::message, dialog, &QProgressDialog::setLabelText);
+	connect(config, &ConfigureDiveComputer::error, dialog, &QProgressDialog::setLabelText);
+	connect(config, &ConfigureDiveComputer::progress, dialog, &QProgressDialog::setValue);
+// 	connect(dialog, &QProgressDialog::finished, config, &ConfigureDiveComputer::dc_close); TODO: this is wrong
```
DerDakon: One of the things that probably have not worked even before this path

desktop-widgets/subsurfacewebservices.cpp
```diff
@@ -353,11 +356,11 @@ void WebServices::updateProgress(qint64 current, qint64 total)
 
 void WebServices::connectSignalsForDownload(QNetworkReply *reply)
 {
-	connect(reply, SIGNAL(finished()), this, SLOT(downloadFinished()));
-	connect(reply, SIGNAL(error(QNetworkReply::NetworkError)),
-		this, SLOT(downloadError(QNetworkReply::NetworkError)));
-	connect(reply, SIGNAL(downloadProgress(qint64, qint64)), this,
-		SLOT(updateProgress(qint64, qint64)));
+// 	connect(reply, &QNetworkReply::finished, this, &WebServices::downloadFinished); TODO: wrong
```
DerDakon: And these connects wont work either because there are no such slots in WebServices.

bstoeger: As far as I know, the webservice thing was discontinued. In any case, thanks for all the cleanups / fixes!

bstoeger: Actually, perhaps this did work before your patch:

note how `DivelogsDeWebServices` does have a `DownloadFinished` slot. Not sure how Qt handles that case with the string-signature thing.

Easy fix would be adding a pure virtual function. 

But I would prefer merging these two classes - what is the point of a parent class when there is a single subclass...?

core/qtserialbluetooth.cpp
```diff
@@ -106,14 +109,15 @@ static dc_status_t qt_serial_open(qt_serial_t **io, dc_context_t*, const char* d
 
 	// Wait until the connection succeeds or until an error occurs
 	QEventLoop loop;
-	loop.connect(serial_port->socket, SIGNAL(connected()), SLOT(quit()));
-	loop.connect(serial_port->socket, SIGNAL(error(QBluetoothSocket::SocketError)), SLOT(quit()));
+	QObject::connect(serial_port->socket, &QBluetoothSocket::connected, &loop, &QEventLoop::quit);
+	QObject::connect(serial_port->socket, QOverload<QBluetoothSocket::SocketError>::of(&QBluetoothSocket::error),
+		&loop, [&loop](QBluetoothSocket::SocketError) { loop.quit(); });
```
dirkhh: Here we reached the point where I KNOW what the old code does and I have not the faintest clue what the new code does.
Sure, you know a lot more about Qt and C++ programming than I do, but for now I assume that I'll have to maintain this syntax salad.
Ugh.

desktop-widgets/CMakeLists.txt
```diff
@@ -81,6 +81,7 @@ set(SUBSURFACE_INTERFACE
 	command_base.cpp
 	command_divelist.cpp
 	locationinformation.cpp
+	qoverload.h
```
dirkhh: why add a `.h` file in the middle of the `C++` files?

desktop-widgets/configuredivecomputerdialog.cpp
dirkhh: I don't like the idea of commenting this out with a TODO. Either leave the old code, or figure out what we want / need there. But removing something and replacing it with nothing plus a comment that there's work left to do... not so cool.

desktop-widgets/divelistview.cpp
```diff
@@ -541,7 +541,7 @@ void DiveListView::reloadHeaderActions()
 			a->setChecked(shown);
 			a->setProperty("index", i);
 			a->setProperty("settingName", settingName);
-			connect(a, SIGNAL(triggered(bool)), this, SLOT(toggleColumnVisibilityByIndex()));
+			connect(a, &QAction::triggered, this, [this](int) { toggleColumnVisibilityByIndex(); });
```
dirkhh: again, no clue what this does

desktop-widgets/diveplanner.cpp
```diff
@@ -461,57 +466,54 @@ PlannerSettingsWidget::PlannerSettingsWidget(QWidget *parent, Qt::WindowFlags f)
...
-	connect(ui.bestmixEND, SIGNAL(valueChanged(int)), this, SLOT(setBestmixEND(int)));
-	connect(ui.bottomSAC, SIGNAL(valueChanged(double)), this, SLOT(bottomSacChanged(double)));
-	connect(ui.decoStopSAC, SIGNAL(valueChanged(double)), this, SLOT(decoSacChanged(double)));
+	auto p = plannerModel;
+	connect(ui.recreational_deco, &QRadioButton::clicked, plannerModel,
+		[p, this]() { p->setDecoMode(RECREATIONAL); disableDecoElements(RECREATIONAL); });
+	connect(ui.buehlmann_deco, &QRadioButton::clicked, plannerModel,
+		[p, this]() { p->setDecoMode(BUEHLMANN); disableDecoElements(BUEHLMANN); });
+	connect(ui.vpmb_deco, &QRadioButton::clicked, plannerModel,
+		[p, this]() { p->setDecoMode(VPMB); disableDecoElements(VPMB); });
```
dirkhh: I am repeating myself, but I intensely dislike this. No idea what it does, makes debugging harder as I can't set breakpoints. Not a fan.

desktop-widgets/locationinformation.cpp
```diff
@@ -24,18 +24,18 @@ LocationInformationWidget::LocationInformationWidget(QWidget *parent) : QGroupBo
 	ui.diveSiteMessage->setCloseButtonVisible(false);
 
 	acceptAction = new QAction(tr("Apply changes"), this);
-	connect(acceptAction, SIGNAL(triggered(bool)), this, SLOT(acceptChanges()));
+	connect(acceptAction, &QAction::triggered, this, [this](bool) { acceptChanges(); });
```
dirkhh: see above

desktop-widgets/locationinformation.cpp
```diff
@@ -24,18 +24,18 @@ LocationInformationWidget::LocationInformationWidget(QWidget *parent) : QGroupBo
...
 
 	rejectAction = new QAction(tr("Discard changes"), this);
-	connect(rejectAction, SIGNAL(triggered(bool)), this, SLOT(rejectChanges()));
+	connect(rejectAction, &QAction::triggered, this, [this](bool) { rejectChanges(); });
```
dirkhh: see above

desktop-widgets/mainwindow.cpp
```diff
@@ -169,16 +172,13 @@ MainWindow::MainWindow() : QMainWindow(),
 	diveSiteEdit = new LocationInformationWidget(this);
 	connect(diveSiteEdit, &LocationInformationWidget::endEditDiveSite,
 			this, &MainWindow::setDefaultState);
-	connect(diveSiteEdit, SIGNAL(endEditDiveSite()), this, SLOT(refreshDisplay()));
+	connect(diveSiteEdit, &LocationInformationWidget::endEditDiveSite,
+			this, [this] { refreshDisplay(); });
```
dirkhh: more of the same

dirkhh (changes requested):
I'm happy with the first commit. Definitely something on my todo list.
The `indexOf()` replacement seems obviously broken to me.
And the replacement of the connect syntax is a mega commit that I have a lot of issues with.
Sure, in many cases this seems to be the obvious and reasonable thing to do.
But in quite a few of them (I gave up after half a dozen) I can't make sense of the new code and it seems to be an orgy in C++ syntax that causes me pain.
Please note, I'm not saying that your code in all of these instances is wrong. I'm saying that I am not smart enough to look at it and say "yeah, I get it".
I'd much prefer that we deal with the easy ones (ideally in smaller, easier to digest chunks) and then figure out what the right thing is for the remaining ones that require lambda bla something this something else seven special characters TADA!!! in order to work - vs the existing, easily understandable code.

dirkhh:
Forgot to say:
I appreciate that you are trying to clean this up.
Please don't see this as a rejection of your work; see this as an admission that you have exceeded **my** limitations of dealing with modern C++ code... and my "per commit size limit" :-)

desktop-widgets/divelistview.cpp
bstoeger: NB: In this case I also don't understand the point of the lambda. This should work without, no?

bstoeger: Or is it the default parameter to the `triggered` signal that makes overload resolution fail?

desktop-widgets/filterwidget2.cpp
```diff
@@ -52,21 +55,16 @@ FilterWidget2::FilterWidget2(QWidget* parent) :
...
-	// QDoubleSpinBox::valueChanged() sends double and QString using the same signal name.
-	// QComboBox::currentIndexChanged() sends int and QString using the same signal name.
-	// Qt 5.7 provides a "convenient" helper that hides this, but only if compiling in C++14
-	// or higher. One can then write: "QOverload<double>(&QDoubleSpinBox::valueChanged)", etc.
-	connect(ui.maxAirTemp, static_cast<void (QDoubleSpinBox::*)(double)>(&QDoubleSpinBox::valueChanged),
+	connect(ui.maxAirTemp, QOverload<double>::of(&QDoubleSpinBox::valueChanged),
 		this, &FilterWidget2::updateFilter);
 
-	connect(ui.minAirTemp, static_cast<void (QDoubleSpinBox::*)(double)>(&QDoubleSpinBox::valueChanged),
+	connect(ui.minAirTemp, QOverload<double>::of(&QDoubleSpinBox::valueChanged),
```
bstoeger: These are a distinct improvement, though. The `static_cast<>` trick was crazy.

desktop-widgets/modeldelegates.cpp
```diff
@@ -84,10 +87,10 @@ const QSize& StarWidgetsDelegate::starSize() const
 ComboBoxDelegate::ComboBoxDelegate(QAbstractItemModel *model, QObject *parent, bool allowEdit) : QStyledItemDelegate(parent), model(model)
 {
 	editable = allowEdit;
-	connect(this, SIGNAL(closeEditor(QWidget *, QAbstractItemDelegate::EndEditHint)),
-		this, SLOT(revertModelData(QWidget *, QAbstractItemDelegate::EndEditHint)));
-	connect(this, SIGNAL(closeEditor(QWidget *, QAbstractItemDelegate::EndEditHint)),
-		this, SLOT(fixTabBehavior()));
+	connect(this, &ComboBoxDelegate::closeEditor,
```
bstoeger: These two could now be made a single line.

desktop-widgets/qoverload.h
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: (GPL-3.0 OR LGPL-3.0)
+#pragma once
```
bstoeger: We tend to use classical include guards. Let's keep things consistent.
(I would have no problem with changing everything to pragma, but also don't see the advantage).

DerDakon: I was just lazy.

desktop-widgets/CMakeLists.txt
DerDakon: Pure nonsense, my bad.

dirkhh: You have not the slightest idea how relieved I am to hear that... :-)

desktop-widgets/filterwidget2.cpp
dirkhh: Completely agree

DerDakon:
I'll rework this to leave out the offending parts.

dirkhh:
> I'll rework this to leave out the offending parts.

Nothing here was "offending".
What I would love to see is that MASSIVE commit broken down into smaller ones that are obvious
And then maybe as a separate PR with the more complex ones - with explanations for the the not so C++ savvy maintainer.
@bstoeger has done a great job in the past to simply help me over the hump by adding excessive comments or stronger commit messages.
I apologize for the extra work this causes - but I think long term this is the right approach for the project.

Makes sense?

DerDakon:
All fine, that was just how I read this. "Controversial" would have been a better wording I guess. I just dropped the Qt-connect-style commit for now and will rework that. It was more of a "what do you think about that" at the end anyway. I guess the other commits should be fine.

core/connectionlistmodel.cpp
```diff
@@ -37,8 +37,10 @@ void ConnectionListModel::removeAllAddresses()
...
 
-int ConnectionListModel::indexOf(QString address)
+int ConnectionListModel::indexOf(const QString &address) const
 {
-	const QRegExp re(".*" + address + ".*", Qt::CaseInsensitive);
-	return m_addresses.indexOf(re);
+	for (const QString &a : m_addresses)
+		if (a.contains(address, Qt::CaseInsensitive))
+			return true;
+	return false;
```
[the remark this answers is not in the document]
DerDakon: You are absolutely right, I misread the code. Will fix.

bstoeger:
> I just dropped the Qt-connect-style commit for now

If it isn't too much work, perhaps just leave in the instances that need neither `qOverload` nor lambdas.

I think we all agree that this would be a win. Maybe we should even add a line to the coding style document to use the member-function syntax.

core/connectionlistmodel.cpp
```diff
@@ -37,8 +37,10 @@ void ConnectionListModel::removeAllAddresses()
...
 }
 
-int ConnectionListModel::indexOf(QString address)
+int ConnectionListModel::indexOf(const QString &address) const
 {
-	const QRegExp re(".*" + address + ".*", Qt::CaseInsensitive);
-	return m_addresses.indexOf(re);
+	for (int i = 0; i < address.count(); i++)
+		if (address.at(i).contains(address, Qt::CaseInsensitive))
+			return true;
```
bstoeger: This should be `return i`. And of course twice `m_addresses` instead of `address`, but the compile failures already tell you that...

core/divelogexportlogic.cpp
```diff
@@ -128,21 +128,19 @@ void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes)
...
-	QString exportFiles = file.fileName() + "_files";
+	const QString exportFiles = file.fileName() + "_files" + QDir::separator();
+	mainDir.mkdir(exportFiles);
 
-	QString json_dive_data = exportFiles + QDir::separator() + "file.js";
-	QString json_settings = exportFiles + QDir::separator() + "settings.js";
-	QString translation = exportFiles + QDir::separator() + "translation.js";
-	QString stat_file = exportFiles + QDir::separator() + "stat.js";
-	exportFiles += "/";
+	const QString json_dive_data = exportFiles + "file.js";
```
bstoeger: Is there any point in making these const? We generally don't follow a const-by-default style for local variables.

core/connectionlistmodel.cpp
DerDakon: sorry, pushed wrong version.

core/divelogexportlogic.cpp
DerDakon: I'm just used to that.

desktop-widgets/locationinformation.cpp
DerDakon: This is not needed, I fixed it.

desktop-widgets/divelistview.cpp
DerDakon: This goes wrong if the signal has no parameters, and the slot has a lot, but all have default parameters. You sadly can't just connect them in that case, so I took the "simple" solution of adding an intermediate step that just forwards it. But I applied that pattern a bit too often.

core/qtserialbluetooth.cpp
DerDakon: The second part is luckily not needed at all, one can directly connect here.

core/connectionlistmodel.cpp
bstoeger: Attention: this still says `return true` instead of `return i`!

DerDakon: Apparently I'm unable to do a simple squash. Let's try again.

dirkhh (approved):
These all look good. Thanks!

dirkhh:
I'd just like to repeat what I said before, and what @bstoeger said here - please don't give up on the connection syntax change. Most of it we all agree on, and the ones where my head explodes because I need to practice my C++ reading skills simply require a bit of care and feeding of the maintainer. Ideally in smaller chunks :-)

DerDakon:
Yes, the rest will come. It's just that I sometimes get distracted by real life or other side projects ;)

dirkhh:
Ha. I know THAT feeling. No worries. Just over-communicating to avoid misunderstandings.

PR #2025, opened 2019-03-24
Cleanup: fix printGPSCoords signature and leaks

bstoeger:
The printGPSCoords() function returns a copied C-style string. Since
the owndership is transferred to the caller, the correct return type
is "char *" instead of "const char *".

Thus a number of casts when calling free can be removed.

Moreover a number of callers didn't free the string and thus were
leaking memory. Fix them. Ultimately we might want two versions
of the function: one for QString, one for C-style strings.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes return code of `printGPSCoords` and a few leaks.

### Changes made:

1) Make `printGPSCoords` return `char *`.
2) Free memory returned by `printGPSCoords`.

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/gpslocation.cpp
+++ b/core/gpslocation.cpp
@@ -137,7 +137,9 @@ QString GpsLocation::currentPosition()
 		if (delta < 300) {
 			// we can simply use the last position that we tracked
 			gpsTracker gt = m_trackers.last();
-			QString gpsString = printGPSCoords(&gt.location);
+			char *gps = printGPSCoords(&gt.location);
+			QString gpsString = gps;
+			free(gps);
 			qDebug() << "returning last position" << gpsString;
 			return gpsString;
 		} else {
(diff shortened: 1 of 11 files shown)


core/gpslocation.cpp
```diff
@@ -137,7 +137,9 @@ QString GpsLocation::currentPosition()
 		if (delta < 300) {
 			// we can simply use the last position that we tracked
 			gpsTracker gt = m_trackers.last();
-			QString gpsString = printGPSCoords(&gt.location);
+			char *gps = printGPSCoords(&gt.location);
+			QString gpsString = gps;
+			free(gps);
```
dirkhh, line 142: we have this pattern three times - I guess it's worth while to have a little C++ wrapper that does this. But that could easily be in a later commit.

dirkhh (approved):
This looks good. Small comment that agrees with you that it might be worth while to improve things with a Qt wrapper for the Qt users...

PR #2027, opened 2019-03-25
Heat Map at Surface

atdotde:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Part of the decompression is done at the surface while breathing air again. This PR adds the possibility to show this in the planner by extending the dive by a final segment at zero depth breathing air. This way, tissue saturations and in particular the heat map becomes visible after reaching the surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Planner: Allow for a final segment at the surface to display further desaturation
 - Desktop: Add stats by depth and temperature ranges to yearly stats [#1996]
 - Desktop: make sure cloud storage email addresses are lower case only
 - Desktop: Fix editing of dive-time [#1975]
--- a/core/planner.c
+++ b/core/planner.c
@@ -1085,6 +1085,7 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 			// Switch to an empty air cylinder for breathing air at the surface
 			// If no empty cylinder is found, keep using last deco gas
 			current_cylinder = i;
+			dive->cylinder[i].cylinder_use = NOT_USED;
 			break;
 		}
 	plan_add_segment(diveplan, prefs.surface_segment, 0, current_cylinder, 0, false, OC);
--- a/core/planner.c
+++ b/core/planner.c
@@ -1087,7 +1087,7 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 			current_cylinder = i;
 			break;
 		}
-	plan_add_segment(diveplan, 600, 0, current_cylinder, 0, false, OC);
+	plan_add_segment(diveplan, prefs.surface_segment, 0, current_cylinder, 0, false, OC);
 	create_dive_from_plan(diveplan, dive, is_planner);
 	add_plan_to_notes(diveplan, dive, show_disclaimer, error);
 	fixup_dc_duration(&dive->dc);
--- a/core/planner.c
+++ b/core/planner.c
@@ -1080,6 +1080,14 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 		diveplan->eff_gflow = lrint(100.0 * (regressiona() * first_stop_depth + regressionb()));
 	}
 
+	for (int i = 0; i < MAX_CYLINDERS; i++)
+		if (cylinder_nodata(&dive->cylinder[i])) {
+			// Switch to an empty air cylinder for breathing air at the surface
+			// If no empty cylinder is found, keep using last deco gas
+			current_cylinder = i;
+			break;
+		}
+	plan_add_segment(diveplan, 600, 0, current_cylinder, 0, false, OC);
 	create_dive_from_plan(diveplan, dive, is_planner);
 	add_plan_to_notes(diveplan, dive, show_disclaimer, error);
 	fixup_dc_duration(&dive->dc);
--- a/core/plannernotes.c
+++ b/core/plannernotes.c
@@ -24,9 +24,12 @@ static int diveplan_duration(struct diveplan *diveplan)
 {
 	struct divedatapoint *dp = diveplan->dp;
 	int duration = 0;
+	int lastdepth = 0;
 	while(dp) {
-		if (dp->time > duration)
+		if (dp->time > duration && (dp->depth.mm > SURFACE_THRESHOLD || lastdepth > SURFACE_THRESHOLD)) {
 			duration = dp->time;
+			lastdepth = dp->depth.mm;
+		}
 		dp = dp->next;
 	}
 	return (duration + 30) / 60;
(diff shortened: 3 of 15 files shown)


core/planner.c
```diff
@@ -1080,6 +1080,13 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 		diveplan->eff_gflow = lrint(100.0 * (regressiona() * first_stop_depth + regressionb()));
 	}
 
+	for (int i = 0; i < MAX_CYLINDERS; i++)
+		if (cylinder_nodata(&dive->cylinder[i])) {
+			// Switch to an empty air cylinder for breathing air at the surface
+			current_cylinder = i;
+			break;
+		}
```
dirkhh, line 1090: obvious question... what happens if the user has all available cylinders defined and none of them are air?
I know, sounds like a corner case, but I also know that those are the things that tend to bite us.

core/planner.c
```diff
@@ -1080,6 +1080,13 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
...
+	plan_add_segment(diveplan, 600, 0, current_cylinder, 0, false, OC);
```
dirkhh: adding this unconditional, instead of as an option. Hmm. Interesting. I didn't expect that.

core/plannernotes.c
```diff
@@ -193,6 +196,8 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
 		    !rebreatherchange_before &&
 		    !rebreatherchange_after)
 			continue;
+		if (lastdepth == 0 && dp->depth.mm == 0 && !dp->next)
+			continue;
```
dirkhh, line 201: so that's the end of the plan - why do we skip the rest of the loop? A comment here might be helpful when we look at this code in a few years

profile-widget/diveeventitem.cpp
```diff
@@ -254,20 +255,34 @@ bool DiveEventItem::shouldBeHidden()
 	return false;
 }
 
+int DiveEventItem::depthAtTime(int time)
+{
+	QModelIndexList result = dataModel->match(dataModel->index(0, DivePlotDataModel::TIME), Qt::DisplayRole, time);
+	if (result.isEmpty()) {
+		Q_ASSERT("can't find a spot in the dataModel");
+		hide();
+		return -666;
```
dirkhh: The number of the beast as an error number is cute... but I'd prefer something with a symbolic name that is then checked by the caller... because the caller above uses this as a depth (admittedly just to compare against a constant, but still... worst case return "-1" which you define as "ABOVE_SURFACE")

profile-widget/diveeventitem.cpp
```diff
@@ -254,20 +255,34 @@ bool DiveEventItem::shouldBeHidden()
...
+	}
+	return dataModel->data(dataModel->index(result.first().row(), DivePlotDataModel::DEPTH)).toInt();
+}
+
 void DiveEventItem::recalculatePos(bool instant)
 {
 	if (!vAxis || !hAxis || !internalEvent || !dataModel)
 		return;
 
+
```
dirkhh: a second empty line?

profile-widget/diveeventitem.cpp
```diff
@@ -254,20 +255,34 @@ bool DiveEventItem::shouldBeHidden()
...
 	QModelIndexList result = dataModel->match(dataModel->index(0, DivePlotDataModel::TIME), Qt::DisplayRole, internalEvent->time.seconds);
 	if (result.isEmpty()) {
 		Q_ASSERT("can't find a spot in the dataModel");
 		hide();
 		return;
 	}
+	int depth = depthAtTime(internalEvent->time.seconds);
+	if (depth < 0)
```
dirkhh: and then here you could compare with `==` to your symbolic constant

core/planner.c
```diff
@@ -1086,7 +1086,7 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 			current_cylinder = i;
 			break;
 		}
-	plan_add_segment(diveplan, 600, 0, current_cylinder, 0, false, OC);
+	plan_add_segment(diveplan, prefs.surface_segment, 0, current_cylinder, 0, false, OC);
```
dirkhh, line 1091: Ahh, this makes it conditional, in a way. That works.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Planner: Allow for a final dive segment at the surface
```
dirkhh: I'm being picky, but is this a "dive" segment or is this a "deco" segment?

dirkhh (changes requested):
Typo in the last commit message `suface` vs `surface`.
My review is more nit-picky than usual, my apologies. I wish I'd always do such thorough reviews - this just happens to be something that was easy to understand for me, so I felt ambitious :-)

core/planner.c
atdotde: I thought about this as well. First of all, this should not happen(™). Remember the discussion when we increased MAX_CYLIDERS to its current value and some senior people complained that "this is far too big and occupies too much memory"? But we increased it anyway to have room when merging dives into different dive computers of a single dive (and in that case, it's not a problem as they would share the "surface air" gas). But we could increase MAX_CYLINDERS by one...

That said, what happens now if we run out of gases to have an additional surface air gas is to continue to breath the last deco gas. I admit, this could be confusing but it is failing gravely.

Possible alternatives: when hitting MAX_CYLINDERS, forcing the extra time to 0, effectively turning this feature off. That would be a bit more code but probably a nicer failure mode. Or I could stop modelling this surface segment breathing surface air by an additional cylinder and making this segment "more special" with the implicit meaning of breathing air. But I guess that would introduce "if (segment.type == SPECIAL)" cases in many many places.

core/planner.c
atdotde: UI is added in second commit as you will see. This first commit is to get the logic right.

core/plannernotes.c
atdotde: We skip the rest as the surface segment should not show up in the notes. Will add a comment.

profile-widget/diveeventitem.cpp
atdotde: I needed some out of bounds special value to signal error (before, when this was part of the other function which is of type void, this was simply a return;.

-1 would not really work (yes, I just realised that the conditional below should take that into account) as that is a valid depth value of a somewhat noisy depth sensor. So I needed a more negative number. But yes, that should be a constant...

profile-widget/diveeventitem.cpp
atdotde: oops

profile-widget/diveeventitem.cpp
atdotde: I should definitely test against the constant (which should be negative enough not to be a possible value of a depth sensor)

CHANGELOG.md
atdotde: Will do. Expect an update tonight (old world time).

core/planner.c
dirkhh: You mean it's failing gracefully? I agree, that's not a terrible failure mode. And indeed, the case is unlikely to happen.
And on the overall tradeoff of this vs extra code for special handling... let's keep it simple.

atdotde:
Rebased on current master and addressed comments.

dirkhh (commented):
Still that typo in the final commit message:
"Make suface air..."
"suface" isn't a word I know...

dirkhh (approved):
LGTM

PR #2028, opened 2019-03-27
CMake: add headers to project

DerDakon:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Add the headers to the CMake target so they show up in IDE projects.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -23,104 +23,183 @@ endif()
 if(BTSUPPORT)
 	add_definitions(-DBT_SUPPORT)
 	set(BT_SRC_FILES desktop-widgets/btdeviceselectiondialog.cpp)
-	set(BT_CORE_SRC_FILES qtserialbluetooth.cpp btdiscovery.cpp)
+	set(BT_CORE_SRC_FILES
+		btdiscovery.cpp
+		btdiscovery.h
+		qtserialbluetooth.cpp
+	)
 endif()
 
 if(BLESUPPORT)
 	add_definitions(-DBLE_SUPPORT)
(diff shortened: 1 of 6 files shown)


atdotde:
I guess the Travis failure was a fluke. I restarted that build. Besides that it looks good, but your commit messages are, eh, somewhat dense. Why is this an improvement?

dirkhh:
Similar question as @atdotde ... 
Why is a function better than a macro?
Why would we want to add the headers as targets (so far it seems they are picked up as dependencies, anyway)?
I always love to learn how we can do things better, @DerDakon :-)

DerDakon:
A functions is better than a macro for the same reason as in C: it adds a scope. Also macro arguments, also they look like being variables, are not really variables, as documented here: https://cmake.org/cmake/help/latest/command/macro.html#argument-caveats

The headers are found as dependencies, but they do not show up as project members when e.g. using MSVC, so it is not that comfortable to open a header that is part of the project.

dirkhh:
ok, makes sense

bstoeger:
I'm not convinced about adding the headers to the cmake file. Isn't the whole point of cmake that it can find these dependencies? This is a step away from automation. Moreover, omitting a .c or .cpp file will result in a failed build, but omitting a .h will not. Keeping this consistent looks like a maintenance headache. Do at least other IDEs profit from that?

DerDakon:
It finds dependencies to headers automatically. Making them part of the project makes just sure that they show up in the IDEs.

PR #2035, opened 2019-04-03
Cleanup: simplify QPref class

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I was bothered by the fact that the qPref.cpp file was constantly recompiling without me touching it. It turns out that it defined access functions to the version strings, which were never used. I'm not sure why a *preferences* object would provide *version* information...? Therefore I removed it. Then I noticed that the whole object actually has no state whatsoever. Therefore there seems to be no obvious point in generating an instance of it (well, nothing involving QML is obvious). But after removing this, both desktop and mobile survive a quick compile & start test...

### Changes made:

1) Remove unsused functions
2) Make registerQML static
3) Remove qPref singleton
4) make qPref a non-Q_OBJECT class

### Release note:

No.
### Documentation change:

No.
### Mentions:

@dirkhh: you were the last person touching this code. Perhaps I'm trampling down something subtle..?


DerDakon:
I suggest either explicitely deleting the constructor to prevent creating an instance of this entirely, or just making it a namespace.

dirkhh (approved):
This should work. And I want to throw a stick of dynamite into the preference handling eventually. In my infinite spare time

dirkhh:
> I suggest either explicitely deleting the constructor to prevent creating an instance of this entirely, or just making it a namespace.

I believe this might mess with the integration with QML

PR #2039, opened 2019-04-04
Cleanup: provide our own qOverload<> implementation.

bstoeger:
This is only in Qt 5.7 and therefore can't be used in Qt 5.5 and 5.6
builds. Moreover, we can't simply reuse Qt's version owing to
licensing concerns.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Hide that horrible and unwieldy `static_cast` hack to resolve the connect-mess caused by function overloading of signals.
### Changes made:

1) Implement a trivial alternative to qOverload without all the special cases.

### Related issues:

#2036
### Release note:

No.
### Documentation change:

No.
### Mentions:

@DerDakon @dirkhh

### DIFF
--- a/core/qthelper.h
+++ b/core/qthelper.h
@@ -116,6 +116,21 @@ void moveInVector(Vector &v, int rangeBegin, int rangeEnd, int destination)
 		std::rotate(it + destination, it + rangeBegin, it + rangeEnd);
 }
 
+// Qt overloads some signals(!) which can't therefore be used in connect() calls with
+// pointers-to-member functions where the signatures of signal and slot don't match perfectly.
+// For this case, Qt 5.7 provides the qOverload<> helper.
+// Since we still support Qt 5.5 let's reimplement it here.
+#if QT_VERSION < 0x050700
+template <typename... Args>
+struct QOverload {
+	template <typename Ret, typename Class>
+	static Ret (Class::*of(Ret (Class::*fun)(Args...)))(Args...)
+	{
+		return fun;
+	}
+};
+#endif
+
 #endif
 
 // 3) Functions visible to C and C++
--- a/desktop-widgets/filterwidget2.cpp
+++ b/desktop-widgets/filterwidget2.cpp
@@ -52,21 +52,16 @@ FilterWidget2::FilterWidget2(QWidget* parent) :
 	connect(ui.minVisibility, &StarWidget::valueChanged,
 		this, &FilterWidget2::updateFilter);
 
-	// We need these insane casts because Qt decided to function-overload some of their signals(!).
-	// QDoubleSpinBox::valueChanged() sends double and QString using the same signal name.
-	// QComboBox::currentIndexChanged() sends int and QString using the same signal name.
-	// Qt 5.7 provides a "convenient" helper that hides this, but only if compiling in C++14
-	// or higher. One can then write: "QOverload<double>(&QDoubleSpinBox::valueChanged)", etc.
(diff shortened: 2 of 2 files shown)


core/qthelper.h
```diff
@@ -116,6 +116,21 @@ void moveInVector(Vector &v, int rangeBegin, int rangeEnd, int destination)
...
 }
 
+// Qt overloads some signals(!) which can't therefore be used in connect() calls with
+// pointers-to-member functions where the signatures of signal and slot don't match perfectly.
+// For this case, Qt 5.7 provides the qOverload<> helper.
+// Since we still support Qt 5.5 do our own implementation here. The template arguments
+// are the function parameters of the member function to pick.
+// Usage: "disambiguate<double>::get(&QDoubleSpinBox::valueChanged)".
+template <typename... Args>
+struct disambiguate {
```
DerDakon: I would just call it qOverload and wrap it in an #ifdef, so we simply use the Qt version if it is present.

bstoeger: Hmm.. I called it differently on purpose to make clear that it is a new implementation, not a copy. Is there any advantage in using Qt's version?

DerDakon: It gives us the chance to just drop the new implementation once Qt 5.7 can be a required version. And it makes it clear for every other Qt programmer what this is about. That is is not a copy is IMHO clear for everyone looking at the implementation.

dirkhh: I like @DerDakon 's approach here. Call it the same and only use it if we are on a Qt version that's too old (<5.7 I believe). And then, once we decide to drop Ubuntu 16.04 (and other Qt 5.5 based distros) we can simply remove the code again.

bstoeger: Fine with me - but obviously I can't test it, because I only have Qt 5.11.1. I'm a bit weary of code that compiles differently for different build environments.

Weirdly, it turns out that it compiles without `#ifdef` even on Qt 5.11. I'm too lazy to find out why. Let's just wait what clang thinks - it seems to be the more picky compiler.

bstoeger: Well... Seems to compile in all test environments.

bstoeger: > Weirdly, it turns out that it compiles without #ifdef even on Qt 5.11.

I see - because I misspelled it. The template is QOverload, not qOverload. Stupid me.

core/qthelper.h
```diff
@@ -116,6 +116,19 @@ void moveInVector(Vector &v, int rangeBegin, int rangeEnd, int destination)
...
+// For this case, Qt 5.7 provides the qOverload<> helper.
+// Since we still support Qt 5.5 let's reimplement it here.
+template <typename... Args>
+struct qOverload {
+	template <typename Ret, typename Class>
+	static Ret (Class::*of(Ret (Class::*fun)(Args...)))(Args...)
+	{
+		return fun;
+	}
+};
```
dirkhh, line 131: so why doesn't this conflict on Qt5.7 and newer?
I am so confused...

bstoeger: Because typo (qOverload vs. QOverload). Will change once I find time...

dirkhh (approved):
I think should go in - any reason not to merge?

bstoeger:
>  any reason not to merge?

Unnecessary complexity for code cosmetics? Your call...

dirkhh:
Given that we can drop the top part once we are on Qt5.7+, I think this does make things easier to read. So I'll merge.

PR #2040, opened 2019-04-06
Merge extra_data list when merging two dives of the same dive computer

torvalds:
When merging two dives into a longer one, merge the dive computer
extra_data list too.

We just pick all the extra-data (but avoid entirely duplicate key/value
entries).

Note that this can cause confusing extra-data, in that both dives migth
have things like "battery percentage at beginning/end of dive" keys, and
if the values are different, you'll now get *both* of those values, but
that's better than randomly just taking one of them.

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -2009,6 +2009,58 @@ static void merge_samples(struct divecomputer *res,
 	}
 }
 
+/*
+ * Does the extradata key/value pair already exist in the
+ * supplied dive computer data?
+ *
+ * This is not hugely efficient (with the whole "do this for
+ * every value you merge" it's O(n**2)) but it's not like we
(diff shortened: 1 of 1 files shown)


torvalds:
Note that it might be more efficient to do something clever like sorting the lists, but with an average of maybe 5-10 entries, i don't think efficiency matters. And right now we keep the list in "reported by dive computer order", so sorting might have odd visual effects if the dive computer reports related key/value pairs together.

So I'm not claiming this is smart or beautiful, but it seems to work for me.

bstoeger (approved):
Looks good to me.

PR #2042, opened 2019-04-10
Add export option to export dive sites.

djunkins:
Add export option to export dive sites.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add export option to export dive sites.

Selecting "Selected dives" exports the dive sites for the selected dives. Selecting "All dives" exports all dive sites.

XML format is the subsection of the divelog XML that describes the sites headed with a <divesites> section like:
```
      <divesites program='subsurface' version='3'>
      </divesites>
```


core/save-xml.c
```diff
@@ -798,3 +798,64 @@ int export_dives_xslt(const char *filename, const bool selected, const int units
...
 }
+
+void save_dive_sites_buffer(struct membuffer *b, const bool select_only, bool anonymize)
+{
+	int i;
+	put_format(b, "<divesites program='subsurface' version='%d'>\n", DATAFORMAT_VERSION);
+
+	/* save the dive sites - to make the output consistent let's sort the table, first */
+	sort_dive_site_table(&dive_site_table);
+	purge_empty_dive_sites(&dive_site_table);
```
dirkhh: why sort before purging? I know, silly point, but still...

core/save-xml.c
```diff
@@ -798,3 +798,64 @@ int export_dives_xslt(const char *filename, const bool selected, const int units
...
+		try_to_backup(filename);
+		error = -1;
+		f = subsurface_fopen(filename, "w");
+	}
+	if (f) {
+		flush_buffer(&buf, f);
+		error = fclose(f);
+	}
+	if (error)
+		report_error("Save failed (%s)", strerror(errno));
```
dirkhh: I note that we don't translate the error messages. Oh well, I guess that's not terrible...
Maybe a better error message `"Failed to save divesites to %s (%s)", filename, strerror(errno)`

dirkhh (changes requested):
I like this. A couple of minuscule comments.
I'm always frustrated how noisy the `.ui` file changes end up being - but that's a tool problem.

core/save-xml.c
bstoeger: Note that the `add_dive_site_to_table()` function keeps the array sorted. Therefore I think that we can remove the `sort_dive_site_table()` call entirely.

Moreover, I'm not sure that the `purge_empty_dive_sites()` call still makes sense, now that the user can explicitly delete dive sites.

djunkins: I will remove the call to `sort_dive_site_table()` from this function as well as the `save_dives_buffer()` function it was copied from.

Looking at the `purge_empty_dive_sites()` code, it looks like like it removes the sites that have no name, description, notes, or location. I think it is reasonable to remove those empty sites before exporting to XML.

bstoeger: > Looking at the purge_empty_dive_sites() code, it looks like like it removes the sites that have no name, description, notes, or location. I think it is reasonable to remove those empty sites before exporting to XML.

Sure, but the check should probably be moved into the loop:
```
	for (int i = 0; i < dive_site_table.nr; i++) {
		struct dive_site *ds = get_dive_site(i, &dive_site_table);
		if (is_empty_dive_site(ds))
				continue;
		...
	}
```

The current code can lead to crashes:

1) Clear all dive site fields -> old values remembered in undo command.
2) Save -> This deletes the dive site.
3) Undo -> booom, writing to non-existing dive site.

The problem is that once the dive-site thing is undoized, other code shouldn't add/remove dive sites directly.

djunkins: Got it. That makes sense. I'll make that change and update the pull request. I'll also change it in the existing code to export dives since the same logic exists.

core/save-xml.c
djunkins: I updated the error messages and added the translate function.

bstoeger (approved):
I'm very tired - but lgtm. There is still one dangerous `purge_empty_dive_sites` in `core/save-git.c`. And actually the whole function can be removed. But this all need not be done in this PR.

Thank you.

dirkhh:
replaced by #2044

Issue #2045, opened 2019-04-12
Dive site filtering on a site with no dives is broken


djunkins:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When in the dive site tab, if you select a dive site that has zero dives associated with it, the dive list is empty, but both the main window tabs and the dive site table are grayed out which does not allow the user to unselect the site to reset the filter. The same problem exists when you use the `+` button to add a dive site to the table since the moment the dive site is created it has no dives associated with it.

The only workaround I found was to use <command-F> to bring up the filter pane and reset the filter there.
### Operating system:

MacOS 10.14.4

### Subsurface version:

This is on the recently merged `master 8695d8bdb` branch with the dive site undo features.

### Steps to reproduce:

1) Navigate to `Dive sites` tab
2) Use `+` button to create a dive site in the dive site table
3) Struggle to find out how to edit or unselect the dive site or reset the filter

djunkins:
I also just noticed and verified that after triggering the bug above, MacOS gives an error saying Subsurface closed unexpectedly when the user exits from the program through the normal process (Command-Q).

dirkhh:
That might be the one I'm describing in #2046

djunkins:
Yes, the crash on exit sounds the same. I did not attempt to isolate it to filtering on an unused dive site vs. filtering on any dive site.

bstoeger:
That shouldn't be too hard to fix. But while doing so, I found a cognate bug: On File>Close, the dive site table keeps its entries and adding a new dive site will add it at the end.

bstoeger:
Should be fixed in master.

Issue #2046, opened 2019-04-12
Crash on exit when filtering dive list by selecting a dive site


dirkhh:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

This is easy to reproduce.
Pick a dive site in the dive site tab - the dive list is now filtered to only show dives at that site.
Now quit Subsurface.
Crash

### Operating system:

Tested on Linux

### Subsurface version:

latest master 

### Additional information:

stack trace seems to indicate that in the `MainWindow` destructor we end up calling down to `MultiFilterSortModel::myInvalidate()` which emits `filterFinished()` which then calls `mainTab->updateDiveInfo()` which dereferences `MainWindow::instance()->graphics->isPlanner()` and crashes.

### Mentions:

@bstoeger

bstoeger:
> This is easy to reproduce.

Doesn't happen for me... :)

But we had that bug before when the filter was active... I'll try to remember what that was.

bstoeger:
Oh yes, now I remember: somehow some Qt version changed and would call hide() on child objects during destruction(!). I was pretty upset back then, because a destructor should free memory and *not* call methods that might invoke things in a partially destructed object tree. If this is the same thing, I consider it a bug in Qt. Evidently, we still have to work around it. The last time the solution was transforming child-objects into sub-objects, I believe.

Might also be something completely different, of course...

Could you post the backtrace?

dirkhh:
What I list above is the back trace - with a few Qt internal functions in the middle. But those are all of our functions that we call.

bstoeger:
> What I list above is the back trace - with a few Qt internal functions in the middle. 

Those functions and objects would be important, because they explain why it doesn't happen for me.

bstoeger:
I think I understand what's happening:

New Qt-versions call hide on child-objects in the destructor. If the dive-site tab is open, this calls hide, which resets the filter, which accesses the semi-destructed windows object.

I believe this to be a Qt bug: It seems quite normal that child objects would call into their parent on hiding. Calling hide() in the destructor of the *base* class violates what you would expect from a C++ destructor. You're supposed to free resources and nothing else. It seems to me that the whole Qt-object system reverses logical order of destruction: the child objects should be destroyed *first*.

Perhaps we simply should make the maintab a smart-pointer. That will destroy the maintab before the mainwindow and remove it from Qt's object hierarchy.

[...]

PR #2047, opened 2019-04-12
Don't go into UI deadlock when creating a new site

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a quick-fix for the bug reported in #2045. It is not ideal, because the fundamental problem is that the dive site list is implemented as a dive tab, when it isn't. The "fix" is to disable all tabs but the dive site tab.
 
### Changes made:

1) Don't disable all tabs when no dive is selected.
2) Reset dive location mode on file-close.
### Related issues:

#2045
### Release note:

No - bug not in release.
### Documentation change:

No.
### Mentions:

@djunkins @dirkhh

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -653,6 +653,7 @@ void MainWindow::cleanUpEmpty()
 	diveList->reload();
 	diveList->setSortOrder(DiveTripModelBase::NR, Qt::DescendingOrder);
 	MapWidget::instance()->reload();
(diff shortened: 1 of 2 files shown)


djunkins:
This looks good as a fix for #2045. As you are aware the crash-on-exit, #2046, is due to something else.

dirkhh (approved):
this looks right

PR #2049, opened 2019-04-13
Desktop: destroy maintab before destructing mainwindow

bstoeger:
Apparently, in some Qt-versions the destructor of the base class
calls hide on child-objects (according to Qt's object hierarchy).
This is obviously called after the derived class has been destructed.

In a concrete case, the base class of the destructed MainWindow
would hide the TabDiveSite object. That would reset the filtering
if a dive site was selected, which would indirectly access the
MainWindow, which is already partially destroyed.

Therefore, destroy the MainTab before destroying the MainWindow.
Do this by keeping it as a std::unique_ptr subobject. Thus, it
will be destroyed before the MainWindow and remove itself from
Qt's object hierarchy.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is an attempt at fixing #2046. I can't reproduce the crash, so this is blindfold-mode. Moreover I believe we had problems with `std::unique_ptr`, where on some platforms you had to include the full class definition of the pointed-to object. That would be very unfortunate, as it would mean that including MainWindow.hpp would suck in MainTab.hpp, which is also not exactly light.

### Changes made:

1) Make MainTab a smart-pointer to have it deleted *before* the mainwindow.

### Related issues:

#2046
### Release note:

Bug not in release.
### Documentation change:

No.
### Mentions:

@dirkhh @djunkins 
Please test, I'm only guessing here.

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -133,7 +133,7 @@ MainWindow::MainWindow() : QMainWindow(),
 	// widgets will change on the mainwindow.
 
 	// for the "default" mode
-	mainTab = new MainTab(this);
+	mainTab.reset(new MainTab);
 	diveList = new DiveListView(this);
 	graphics = new ProfileWidget2(this);
 	MapWidget *mapWidget = MapWidget::instance();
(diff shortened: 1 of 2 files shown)


bstoeger:
> where on some platforms you had to include the full class definition of the pointed-to object.

That's only the case if the destructor is inline, so it should be fine here.

neolit123 (commented):
LGTM

djunkins:
I am no longer having the crash on exit after applying this patch.

dirkhh (approved):
Fixes the crash. Thanks

PR #2050, opened 2019-04-13
Desktop: remove edit-check in tag-widget completer

bstoeger:
The tag-widget was only showing the completer if we were in edit mode.
The edit mode does not exist anymore - therefore remove the check.
Hopefully this has no unintended consequences, like the completer
not disappearing when it should.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The completer was not shown when editing tags. This hopefully fixes it without follow-up bugs.
### Changes made:

1) Remove check for edit mode in tag-widget completer.

### Release note:

Bug not in release.

### Documentation change:

No.

### DIFF
--- a/desktop-widgets/tagwidget.cpp
+++ b/desktop-widgets/tagwidget.cpp
@@ -89,14 +89,6 @@ void TagWidget::reparse()
 	if (pos.first >= 0 && pos.second > 0)
 		currentText = text().mid(pos.first, pos.second - pos.first).trimmed();
 
-	/*
-	 * Do not show the completer when not in edit mode - basically
-	 * this returns when we are accepting or discarding the changes.
-	 */
-	if (MainWindow::instance()->mainTab->isEditing() == false) {
-		return;
-	}
-
 	if (m_completer) {
 		m_completer->setCompletionPrefix(currentText);
 		if (m_completer->completionCount() == 1) {


djunkins:
Looks good to me.

desktop-widgets/tagwidget.cpp
```diff
@@ -89,14 +89,6 @@ void TagWidget::reparse()
 	if (pos.first >= 0 && pos.second > 0)
 		currentText = text().mid(pos.first, pos.second - pos.first).trimmed();
 
-	/*
-	 * Do not show the completer when not in edit mode - basically
-	 * this returns when we are accepting or discarding the changes.
-	 */
-	if (MainWindow::instance()->mainTab->isEditing() == false) {
```
dirkhh, line 96: so, if edit mode no longer exists, why do we have an isEditing() function - which is checked in almost a dozen places in current master...?

dirkhh (commented):
I don't think this is wrong, but I'd like to understand why we still HAVE the isEditing() thingy in the first place...

desktop-widgets/tagwidget.cpp
bstoeger: Edit mode still exists when editing cylinders and weights. Therefore the accesses in MainWindow are probably valid.

But in this particular case it was used to close the completer when exiting edit mode(?). Which effectively meant never showing it, because edit mode is not activated for the dive-notes tab.

Perhaps this needs to be reinstated in some other form (on focus out?), but this check is clearly bogus. At least it survived light testing.

dirkhh: I noticed on my Mac that I didn't get a tag completer at all. I'll merge this because yes, this makes sense - and will investigate a bit more why I'm not seeing a completer.

bstoeger: Two points:

1) The commit message is not entirely correct - I will fix it later.
2) The editMode will perhaps stay for some time, because it is used when editing the profile of a manually added dive. Making every planner action undoable ... ouch. Though we could move it away from the maintab to the profile and thus contain it better.

dirkhh: I guess I was a bit pre-mature in merging it, then. Oh well. Too late to fix the commit message.

bstoeger: Well... It's not totally wrong, just a bit imprecise. Should be OK.

PR #2055, opened 2019-04-14
Write dive data as video subtitles

atdotde:
This commit adds an entry to the dive media context
menu which offers to write a subtitle file. This
creates an .ass file for the selected videos.

VLC is able to show these subtitles directly, they
can be integrated into the video file with ffmpeg.

What is still missing:
A check that the selected media file is actually a video and not
an image.
Run ffmpeg to determine the length of the video (currently
hard coded to 5 min).

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


core/save-profiledata.c
```diff
@@ -21,6 +22,15 @@ static void put_double(struct membuffer *b, double val)
 	put_format(b, "\"%f\" ", val);
 }
 
+static void put_secs(struct membuffer *b, int secs)
```
dirkhh: that seems oddly named - I would assume that this writes the # of seconds. Isn't this much more `put_timestamp()`?

core/save-profiledata.c
```diff
@@ -148,13 +158,45 @@ static void put_headers(struct membuffer *b)
...
+	put_secs(b, (entry+1)->sec - offset < length ? (entry+1)->sec - offset : length);
+	put_format(b, "Default,,0,0,0,,");
+	put_format(b, "%d:%02d ", FRACTION(entry->sec, 60));
+	value = get_depth_units(entry->depth, &decimals, &unit);
+	put_format(b, "D=%02.2f %s ", value, unit);
+	if (entry->temperature) {
+		value = get_temp_units(entry->temperature, &unit);
+		put_format(b, "T=%.1f%s ", value, unit);
+	}
+	if (entry->ndl < 3600) {
```
dirkhh: interesting choice... I'd say that deserves a comment?

desktop-widgets/tab-widgets/TabDivePhotos.cpp
```diff
@@ -106,6 +109,36 @@ void TabDivePhotos::recalculateSelectedThumbnails()
...
+	for (const auto &photo: indexes) {
+		if (photo.isValid()) {
+			QString fileUrl = photo.data(Qt::DisplayPropertyRole).toString();
+			if (!fileUrl.isEmpty()) {
+				QFileInfo fi = QFileInfo(fileUrl);
+				QFile subtitlefile;
+				subtitlefile.setFileName(QString(fi.path()) + "/" + fi.completeBaseName() + ".ass");
+				int offset = photo.data(Qt::WhatsThisRole).toInt();
+				struct membuffer b = { 0 };
+				save_subtitles_buffer(&b, &displayed_dive, offset, 300);
```
dirkhh: That hard coded duration is just weird. And definitely deserves a comment :-)

dirkhh (changes requested):
I think this is a great idea. I have a few comments. And I _*really*_ think that having this only do a hardcoded 5 minutes is not sufficient...
How hard would it be to determine the length of the video?

core/save-profiledata.c
atdotde: It does, but in the format the subtitle file likes. Will rename if you prefer.

bstoeger:
I'm not sure I understand what this does. Does this write general subtitles or for a selected video?

In the former case, why would we need the length in the latter case we already *do* know start time and length after thumbnailing. So I am a bit confused about the 5 min thing...

core/save-profiledata.c
atdotde: The comment is in the commit message: this is a todo. I wrote this on a train and wanted to get it out before the train arrives.

desktop-widgets/tab-widgets/TabDivePhotos.cpp
```diff
@@ -106,6 +109,36 @@ void TabDivePhotos::recalculateSelectedThumbnails()
...
+void TabDivePhotos::saveSubtitles()
+{
+	QVector<QString> selectedPhotos;
+	if (!ui->photosView->selectionModel()->hasSelection())
+		return;
+	QModelIndexList indexes = ui->photosView->selectionModel()->selectedRows();
+	if (indexes.count() == 0)
+		indexes = ui->photosView->selectionModel()->selectedIndexes();
+	selectedPhotos.reserve(indexes.count());
+	for (const auto &photo: indexes) {
```
bstoeger, line 121: Ok - now I see what this is trying to do. I'll have to remember how we store date & length information for the added media. I'm sure we have that information, so scanning the files seems pointless.

atdotde:
This last force-push includes a determination of the video length from the thumbnail cache.

core/save-profiledata.c
dirkhh: I would prefer if you'd rename it. We should try to have reasonable function names, especially when introducing something new (so we don't fight with any legacy understanding)

core/save-profiledata.c
dirkhh: I understand the desire to get a PR out. But I don't see an explanation for the choice between NDL and TTS in the commit message, either...

core/save-profiledata.c
```diff
@@ -148,13 +158,46 @@ static void put_headers(struct membuffer *b)
...
+	put_format(b, "\n");
+}
 static void save_profiles_buffer(struct membuffer *b, bool select_only)
 {
 	int i;
 	struct dive *dive;
 	struct plot_info pi;
 	struct deco_state *planner_deco_state = NULL;
 
+
```
dirkhh: double empty line

dirkhh (changes requested):
Thanks for addressing the video duration, a couple other comments remain that I would love to see addressed.

Thanks

core/qthelper.cpp
```diff
@@ -984,6 +984,28 @@ QString thumbnailFileName(const QString &filename)
 	return thumbnailDir() + hash.result().toHex();
 }
 
+int video_duration(const QString &filename)
```
bstoeger: Sorry, but I really dislike the fact that the thumbnailer internals are reproduced here. I'd rather either use a thumbnailer function to fetch the thumbnail or extend the `PictureEntry` with a duration field. I think I would prefer the second option. I'll look into it when I find some time...

desktop-widgets/tab-widgets/TabDivePhotos.cpp
```diff
@@ -106,6 +110,40 @@ void TabDivePhotos::recalculateSelectedThumbnails()
...
+		if (photo.isValid()) {
+			QString fileUrl = photo.data(Qt::DisplayPropertyRole).toString();
+			if (!fileUrl.isEmpty()) {
+				QFileInfo fi = QFileInfo(fileUrl);
+				QFile subtitlefile;
+				subtitlefile.setFileName(QString(fi.path()) + "/" + fi.completeBaseName() + ".ass");
```
bstoeger, line 127: The official extension for subtitles is ".ass"? Jeez.

bstoeger (changes requested):

qt-models/divepicturemodel.cpp
```diff
@@ -98,6 +98,9 @@ QVariant DivePictureModel::data(const QModelIndex &index, int role) const
 		case Qt::DisplayPropertyRole:
 			ret = QFileInfo(entry.filename).filePath();
 			break;
+		case Qt::WhatsThisRole:
```
bstoeger: When the user goes into "what's this role" this will now display the offset in seconds...?

Better use 'Qt::UserRole + 1'. If you prefer you can also give it a symbolic name in the class definition like:
```
        enum {
                OffsetRole = Qt::UserRole + 1,
                LengthRole = Qt::UserRole + 2
        };
```

core/qthelper.cpp
bstoeger: My proposal would be to extend the `PictureEntry` struct by a `duration` field, which is initialized to 0. Then, fill this field in `DivePictureModel::updateThumbnail()` and access it like the offset with a `Qt::UserRole + X`.

This would mean the user has to wait until the thumbnails have loaded, but I think that's acceptable.

qt-models/divepicturemodel.cpp
```diff
@@ -59,7 +59,7 @@ void DivePictureModel::updateDivePictures()
 		if (dive->selected) {
 			int first = pictures.count();
 			FOR_EACH_PICTURE(dive)
-				pictures.push_back({ dive->id, picture, picture->filename, {}, picture->offset.seconds });
+					pictures.push_back({ dive->id, picture, picture->filename, {}, picture->offset.seconds, {.seconds = 0}});
```
dirkhh: silly white space

dirkhh (approved):
I think this looks good now. One tiny silly whitespace thing with an additional leading tab.
@bstoeger - are you happy as well?

atdotde:
Extra tab removed.

qt-models/divepicturemodel.cpp
```diff
@@ -59,7 +59,7 @@ void DivePictureModel::updateDivePictures()
 		if (dive->selected) {
 			int first = pictures.count();
 			FOR_EACH_PICTURE(dive)
-				pictures.push_back({ dive->id, picture, picture->filename, {}, picture->offset.seconds });
+				pictures.push_back({ dive->id, picture, picture->filename, {}, picture->offset.seconds, {.seconds = 0}});
```
bstoeger, line 62: Interesting ... I thought this will only be valid C++ starting with C++20...?

bstoeger (approved):
Haven't tested it, but lgtm. Only that aggregate initialization is surprising to me. I don't think this is valid pre-C++20. But if the compilers are happy with it - why not.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Desktop: For videos, add sive data export as subtitle file
```
bstoeger: "sive" is a typo, right?

qt-models/divepicturemodel.cpp
bstoeger: Well, Travis likes it, so this is good to go for me (except the typo in CHANGELOG).

atdotde:
Changelog typo fixed.

bstoeger:
Thanks - let's merge.

PR #2058, opened 2019-04-14
make sure to clear dives on dive split failure

torvalds:
The dive splitting code returns an error code when splitting fails, but
it turns out that the C++ UI code doesn't actually look at the error
code, and instead expected the resulting dives to be NULL if an error
happened and the split didn't succeed for whatever reason.

Which is kind of lazy of it, but we might as well clear the resulting
dives and make the UI code happy.  This should fix the problem that
Celia Marlowe reported on the Subsurface google groups forum.

Reported-by: Celia Marlowe
Acked-by: Berthold Stoeger <bstoeger@mail.tuwien.ac.at>
Fixes: 145f70aab5 ("Undo: implement split-out of dive computer")

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -3596,8 +3596,6 @@ static int split_dive_at(const struct dive *dive, int a, int b, struct dive **ou
 	struct divecomputer *dc1, *dc2;
 	struct event *event, **evp;
 
-	*out1 = *out2 = NULL;
-
 	/* if we can't find the dive in the dive list, don't bother */
 	if ((nr = get_divenr(dive)) < 0)
 		return -1;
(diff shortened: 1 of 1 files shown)


torvalds:
I've not actually been able to test that this makes a difference, because it all worked for me even before. But it does look _likely_ to be the root cause of the problem that Celia reported.

bstoeger:
Looks good, thank you.

bstoeger (approved):

dirkhh (approved):
look reasonable, waiting for @bstoeger to approve

bstoeger:
I approved ca. 30 min ago. :)

dirkhh:
Eh, well, err, right. I meant to say, err, waiting for Travis.
Right. That's what I was going to say.

PR #2059, opened 2019-04-14
Dive site: display proper text for the two special options

bstoeger:
The dive-site line edit box features two special entries for adding
new dive sites. These should display different texts depending on
whether the current dive has a dive site or not.

The current check is wrong, because it used displayed_dive, but
since the last set of undo-changes, this might not be filled out
correctly anymore. Instead the code should check the actual current
dive.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a very minor bug, but still a bug. The dive site box might have shown the wrong text in specific circumstances (create dive site, vs. create dive site from current site).

### Changes made:

1) use current_dive to fill out text in dive site box.


dirkhh (approved):
that makes sense

PR #2060, opened 2019-04-15
Don't show surface gas in gas use list

atdotde:
Skip gas use calculation for the gas that was used in
the surface segment added by the planner in the end.

Reported-by: philippe@philmassart.net

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh (approved):
That should fix it. Thanks for jumping on that one :-)

PR #2061, opened 2019-04-15
A few dive site selection widget cleanups, e. g. enable sorting

bstoeger:
DiveLocationLineEdit stored a pointer to itself in a global variable
so that the DiveLocationModel can access it to access the filter text.

Instead, on change simply pass the filter text down from DiveLocationLineEdit
to DiveLocationModel.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes a hack that has irked me for the longest time. Removes a global variable by passing down a filter text instead of letting the filter access the parent object via dubious methods.

### DIFF
--- a/desktop-widgets/locationinformation.cpp
+++ b/desktop-widgets/locationinformation.cpp
@@ -262,6 +262,7 @@ void DiveLocationFilterProxyModel::setFilter(const QString &filterIn)
 {
 	filter = filterIn;
 	invalidate();
+	sort(LocationInformationModel::NAME);
 }
 
 bool DiveLocationFilterProxyModel::filterAcceptsRow(int source_row, const QModelIndex&) const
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
This look reasonable

bstoeger:
Another small cleanup. Removes a `qobject_cast`.

bstoeger:
No, sorry. I misread - removing second commit again.

bstoeger:
Now it should make sense...

bstoeger:
...and enable sorting (currently by name, later by distance from current position?).

Moreover don't show the special entries if there is no text (easily done by entering something and then deleting it).

dirkhh (approved):
Small steps in the right direction.

PR #2064, opened 2019-04-17
qt-ble: return DC_STATUS_TIMEOUT rather than DC_STATUS_IO on timeout

torvalds:
This didn't use to matter, because none of the BLE-using backends did
retry on timeout until recently.

But Jef started doing packet sending retry for the Mares Icon backend,
and now we should make sure to distinguish the "IO failed" from "IO
timed out" cases.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh (approved):
This seems "obviously correct" (famous last words)

PR #2067, opened 2019-04-24
Update (and fix) device mobile ID's for Scubapro dive computers

torvalds:
I'm not sure this is actually relevant for anything any more, but this
adds the USB device ID's for the Scubapro G2 Console and HUD versions.
It also fixes things to use the proper vendor name (a bit too much
cut-and-paste, where the code said "Suunto" instead of "Scubapro").

The real device ID changes are in libdivecomputer, this is just the
Android xml list for recognized USB devices that likely nobody really
uses.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -1938,17 +1938,19 @@ void QMLManager::showDownloadPage(QString deviceString)
 				.arg(connectionListModel.indexOf("Suunto"))
 				.arg(productList["Suunto"].indexOf("EON Core"))
 				.arg(connectionListModel.indexOf("USB device"));
-	} else if (deviceString.contains("mVendorId=11884") && // 0x2e6c / 0x3201
-		   deviceString.contains("mProductId=12801")) {
+	} else if (deviceString.contains("mVendorId=11884") && // 0x2e6c / 0x3201,0x3211,0x4201
+		   (deviceString.contains("mProductId=12801") ||
+		    deviceString.contains("mProductId=12817") ||
+		    deviceString.contains("mProductId=16897"))) {
 		name = QString("%1;%2;%3")
 				.arg(connectionListModel.indexOf("Scubapro"))
-				.arg(productList["Suunto"].indexOf("G2"))
+				.arg(productList["Scubapro"].indexOf("G2"))
 				.arg(connectionListModel.indexOf("USB device"));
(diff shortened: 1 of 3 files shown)


torvalds:
NOTE! This does *not* contain the actual libdivecomputer submodule update that is needed to get the new device ID for the G2 HUD. 

The libdivecomputer repo has already been updated, it just needs to get the submodule update in subsurface.

dirkhh (approved):
This looks good. I'm surprised you didn't update the submodule as well. No reason for you not to include that, right?

mobile-widgets/qmlmanager.cpp
```diff
@@ -1938,17 +1938,19 @@ void QMLManager::showDownloadPage(QString deviceString)
 				.arg(connectionListModel.indexOf("Suunto"))
 				.arg(productList["Suunto"].indexOf("EON Core"))
 				.arg(connectionListModel.indexOf("USB device"));
-	} else if (deviceString.contains("mVendorId=11884") && // 0x2e6c / 0x3201
-		   deviceString.contains("mProductId=12801")) {
+	} else if (deviceString.contains("mVendorId=11884") && // 0x2e6c / 0x3201,0x3211,0x4201
+		   (deviceString.contains("mProductId=12801") ||
+		   (deviceString.contains("mProductId=12817") ||
+		   (deviceString.contains("mProductId=16897"))) {
```
dirkhh: too much cut and paste... you don't need all of these leading opening parenthesis, only the first of the three... and this is currently causing a syntax error because you close one too few...

dirkhh (changes requested):
This is why travis fails

mobile-widgets/qmlmanager.cpp
torvalds: Oops. Will fix.

torvalds: Ok, I've force-pushed the obvious fix, but still don't do a mobile build, so... 

Does github automatically redo things after the force-push?

dirkhh: Yes it does :-)

bstoeger (approved):
I don't know this code as I don't use mobile - but the changes certainly appear sensible to me.

bstoeger:
Changes look "obviously correct". Let's merge.

PR #2068, opened 2019-04-25
[RFC] Use actual GPS data in dive site handling

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This changes the way new dive sites are added and dive sites are sorted in the dive site selection widget:

1) If there is a GPS1 extra data entry, new dive sites will be created with that entry.
2) If there is such an entry, dive sites will be sorted by distance to this entry.

This can be extended to divecomputer events. Because such an operation is slow, it is only done sparingly, e.g. when changing the current dive.

The use case I had in mind is the following:
1) Download dives from a place where you haven't been. Go through dives and add new dive site names. These will automatically get correct GPS data.
2) Download dives from a place where you have been. To select the correct dive site, press the new button (currently only labeled with "...") and choose the appropriate site. The best fitting site should be at the top.

I don't have a GPS-enabled dive computer, therefore this needs testing / feedback.

### Changes made:

1) Create function that extracts the GPS1 extra data and falls back to the current dive site
2) Use this function to set location for new dive sites.
3) Use this function to sort dive sites with respect to distance to current dive site.
4) A few minor cleanups.

### Release note:

Should be done, yes.
### Documentation change:

Perhaps.
### Mentions:

@dirkhh @torvalds (pinged via webmail).

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -4500,6 +4500,53 @@ int dive_has_gps_location(const struct dive *dive)
 	return dive_site_has_gps_location(dive->dive_site);
 }
 
+/* Extract GPS location of a dive computer stored in the GPS1
+ * or GPS2 extra data fields */
+static location_t dc_get_gps_location(const struct divecomputer *dc)
+{
+	location_t res = { };
+
+	for (struct extra_data *data = dc->extra_data; data; data = data->next) {
+		if (!strcmp(data->key, "GPS1")) {
+			parse_location(data->value, &res);
+			/* If we found a valid GPS1 field exit early since
+			 * it has priority over GPS2 */
+			if (has_location(&res))
+				break;
+		} else if (!strcmp(data->key, "GPS2")) {
+			/* For GPS2 fields continue searching, as we might
+			 * still find a GPS1 field */
+			parse_location(data->value, &res);
+		}
+	}
+	return res;
+}
+
+/* Get GPS location for a dive. Highest priority is given to the GPS1
+ * extra data written by libdivecomputer, as this comes from a real GPS
+ * device. If that doesn't exits, use the currently set dive site.
+ * This function is potentially slow, therefore only call sparingly
+ * and remember the result.
+ */
+location_t dive_get_gps_location(const struct dive *d)
+{
+	location_t res = { };
+
+	for (const struct divecomputer *dc = &d->dc; dc; dc = dc->next) {
+		res = dc_get_gps_location(dc);
+		if (has_location(&res))
+			return res;
+	}
+
+	/* No libdivecomputer generated GPS data found.
+	 * Let's use the location of the current dive site.
+	 */
+	if (d->dive_site)
+		res = d->dive_site->location;
+
+	return res;
+}
+
 /* When evaluated at the time of a gasswitch, this returns the new gas */
 struct gasmix get_gasmix(const struct dive *dive, const struct divecomputer *dc, int time, const struct event **evp, struct gasmix gasmix)
 {
--- a/core/dive.h
+++ b/core/dive.h
@@ -452,9 +452,6 @@ void split_divecomputer(const struct dive *src, int num, struct dive **out1, str
 #define for_each_dc(_dive, _dc) \
 	for (_dc = &_dive->dc; _dc; _dc = _dc->next)
 
-#define for_each_gps_location(_i, _x) \
-	for ((_i) = 0; ((_x) = get_gps_location(_i, &gps_location_table)) != NULL; (_i)++)
-
 extern struct dive *get_dive_by_uniq_id(int id);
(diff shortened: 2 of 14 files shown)


torvalds:
Looks sane to me.

The one thing I'd suggest is to use GPS2 if you don't find GPS1. That's the exit location (not wonderful, I know), and at least with the Garmin Descent (which is the only dive computer to give GPS location right now), for many dives I only have GPS2 because I forgot to wait for a GPS lock before jumping into the water...

On a non-drift dive, the exit location is usually fine. And even on a drift dive, the exit location is at least _close_ to the dive site, so the whole "sort by distance to show exisiting ones" should work fairly well.

I didn't test it. You should already have an XML file of mine with several GPS coordinates..

bstoeger:
> The one thing I'd suggest is to use GPS2 if you don't find GPS1.

Done. GPS1, if it exists, is given priority over GPS2.

> I didn't test it. You should already have an XML file of mine with several GPS coordinates..

I tested the feature using this file (removed all dives from their site). It seems to work, but I can't tell if the "work flow" is improved.

torvalds:
Right now my workflow for GPS locations has been to just look at the map and match up the name with previous dives (when I've had them) that way. I doubt the new model can be worse, but it's one of those things where it's probably obvious only once you start using it actively and noticing what works and doesn't..

dirkhh (approved):
This all seems really cool and useful. I'll need to play with it with a few real dives. Quick, let's book a dive trip. Or, more boringly, I could delete some dives from my dive log and download them again from my Garmin Descent. Oh well. But from reading the code this does seem an improvement.

dirkhh:
Screen Shot 2019-04-29 at 1 18 48 PM
I'm not sure what I did wrong here... but downloading fresh dives from a Garmin Descent I get these really odd suggestions. Look at the distances claimed. The dive that was downloaded was on Fiji. The closest suggestions are La Herradura and Gibraltar (and a site that's actually close by on Fiji).

bstoeger:
Two possibilities spring to mind: these dive sites have no GPS location or integer overflow... Will check that out tomorrow. Currently busy with administration... Brr...

dirkhh:
They have GPS for sure - it's gotta be some overflow as these are about half way around the globe apart...

bstoeger:
Problem seems to be in `get_distance()`:

When calculating the distance from 36.734910, -3.737600 to -17.316583, 177.962119
I get:
```
lat2_r = -0.302231
lat_d_r = -0.943377
lon_d_r = 3.171258
```
which gives
```
a = 1.117674
```
That doesn't fly too well with `sqrt(1.0 - a)`. NaN is "rounded" by lrint to 0.

Spherical geometry is actually part of my job, but not today, sorry...

djunkins:
Here is a patch that updates the algorithm in get_distance() to one I have used before. Testing with the coordinates in Dirk’s earlier email seems to work (it gives a distance of 11050 miles between Fiji and La Herradura).

-Doug

djunkins:
[0001-Bugfix-for-algorithm-in-get_distance.patch.txt](https://github.com/Subsurface-divelog/subsurface/files/3129454/0001-Bugfix-for-algorithm-in-get_distance.patch.txt)

Here is the patch file previously mentioned.

dirkhh:
How about a nice little pull request?
Pretty please?
I mean, it has an SOB, so I guess I could just apply it... hrmpf.
Of course - it would be fun to point to a source for this formula, too :-)

dirkhh:
Ah. It's the Spherical Law of Cosines
https://www.movable-type.co.uk/scripts/latlong.html

djunkins:
I'll put it into a pull request. Thanks for researching the source. I had it in some python code from years ago, so it was a quick translation into C. I would have had to search for the source of the algorithm.

djunkins:
Quoting from http://www.longitudestore.com/haversine-formula.html

"When using these formulae, care must be taken to ensure 
that h does not exceed 1 due to a floating point error (d is 
only real for h from 0 to 1). h only approaches 1 for 
antipodal points (on opposite sides of the sphere) — in this 
region, relatively large numerical errors tend to arise in the 
formula when finite precision is used. However, because d 
is then large (approaching πR, half the circumference) a 
small error is often not a major concern in this unusual case"

bstoeger:
Rebased on master - that hopefully fixes these strange issues!

dirkhh:
Trying to catch up with the various PRs we have.
This does seem like a very nice improvement. Any reason not to merge?

bstoeger:
>  Any reason not to merge?

The bug you reported is hopefully fixed, but you might want to check again.

dirkhh:
Looks all good to me

PR #2074, opened 2019-04-27
Support for Scubapro G2 HUD discovery via bluetooth

william-perry:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a small patch that adds bluetooth discovery of the new Scubapro G2 HUD computer.  Unfortunately the device name is simply 'HUD' with no other references to G2 or Scubapro.

### Changes made:
1) Simply look for the 'HUD' tag in bluetooth discovery and mark it as a possible 'G2 HUD'  product from vendor 'Scubapro'.

### Release note:

Adding a new dive computer seems worthy of an entry in the changelog.

### Documentation change:
Would need to add G2 HUD to the list of supported bluetooth dive computers 

### Mentions:

@torvalds

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -56,9 +56,10 @@ static dc_descriptor_t *getDeviceType(QString btName)
 		product = "EON Core";
 	}
 
-	if (btName.startsWith("G2")  || btName.startsWith("Aladin")) {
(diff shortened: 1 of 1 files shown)


bstoeger:
LGTM. The repeated `startsWith` is a bit weird, but let's keep it like this for now.

Issue #2076, opened 2019-04-28
Can not edit dive site location on map


djunkins:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When editing a dive site from the dive site tab, I can no longer drag the flag on the map to update the lat/long of the dive site location.

### Operating system:

OSX 10.14.4
### Subsurface version:

* master          9bb7bbef8 [origin/master] Support for Scubapro G2 HUD discovery via bluetooth

### Steps to reproduce:

1) Navigate to dive site tab
2) Select the site to be edited
3) Enter edit mode
4) Attempt to drag flag on the site
### Additional information:

It looks like the grayed out flag for sites that do not have any dives associated with them is missing and the map is not moving to the location of those sites when you attempt to edit them as well.

### Mentions:

@bstoeger

bstoeger:
Analysis:
1) On the dive-site tab a filter is active.
2) Switching the tab to the dive-site-edit tab resets the filter.
3) This emits `filterFinished`
4) `filterFinished` is connected to `MapWidget::reload()`
5) `MapWidget::reload()` exits edit mode and therefore the flag can't be dragged.

I think 4) is wrong, but I'm quite certain that this was introduced to fix problems with the map-edit mode. This will probably take some thinking. That's all horrible spaghetti code!

bstoeger:
Actually I think 5) is wrong.

PR #2078, opened 2019-04-29
Various updates to flags on map

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a partial fix for #2076, which tries to make the behavior of the map edit mode more consistent. Since we get spurious reload signals - these have to be disabled during edit mode.

The problem is that now it works as supposed and only dive sites of dives which are not filtered away are shown. That means that on the dive site tab only the selected dive sites are shown, which is less than optimal. We probably have to add two filter modes
1) show only dive sites of filtered dives
2) show all dive sites even though filtering is active
?
### Changes made:

1) Don't exit edit mode when reloading map
2) Don't reload map when in edit mode
### Related issues:

#2076
### Release note:

Bug not in release.
### Mentions:

@djunkins

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -277,9 +277,7 @@ void DiveListView::restoreSelection()
 		return;
 
 	selectionSaved = false;
-	dontEmitDiveChangedSignal = true;
-	unselectDives();
-	dontEmitDiveChangedSignal = false;
+	QList<int> divesToSelect;
 	Q_FOREACH (dive_trip_t *trip, selectedDives.keys()) {
 		QList<int> divesOnTrip = getDivesInTrip(trip);
 		QList<int> selectedDivesOnTrip = selectedDives.values(trip);
@@ -289,8 +287,9 @@ void DiveListView::restoreSelection()
 			selectTrip(trip);
 			selectedDivesOnTrip.removeAll(-1);
 		}
-		selectDives(selectedDivesOnTrip);
+		divesToSelect += selectedDivesOnTrip;
 	}
+	selectDives(divesToSelect);
 }
 
 // This is a bit ugly: we hook directly into the tripChanged signal to
@@ -417,6 +416,10 @@ void DiveListView::selectDives(const QList<int> &newDiveSelection)
 		return;
 
 	dontEmitDiveChangedSignal = true;
+
+	// First, clear the old selection
+	unselectDives();
+
 	// select the dives, highest index first - this way the oldest of the dives
 	// becomes the selected_dive that we scroll to
 	QList<int> sortedSelection = newDiveSelection;
--- a/desktop-widgets/locationinformation.cpp
+++ b/desktop-widgets/locationinformation.cpp
@@ -182,7 +182,6 @@ void LocationInformationWidget::acceptChanges()
 	MainWindow::instance()->diveList->setEnabled(true);
 	MainWindow::instance()->setEnabledToolbar(true);
 	MainWindow::instance()->setApplicationState("Default");
-	MapWidget::instance()->endGetDiveCoordinates();
 	MapWidget::instance()->repopulateLabels();
 	MultiFilterSortModel::instance()->stopFilterDiveSites();
 }
@@ -200,7 +199,6 @@ void LocationInformationWidget::initFields(dive_site *ds)
 		filter_model.set(0, location_t { degrees_t{ 0 }, degrees_t{ 0 } });
 		clearLabels();
 	}
-	MapWidget::instance()->prepareForGetDiveCoordinates(ds);
 }
 
 void LocationInformationWidget::on_diveSiteCoordinates_editingFinished()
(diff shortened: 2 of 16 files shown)


bstoeger:
First attempt at showing all dive sites in dive site mode.

I'd still like to have all other dive sites grayed-out when on the dive site tab.

bstoeger:
Trying to detangle the whole dive-selection / site-selection mess and I have a feeling that I'm running in circles... ☹

User visible changes:
- Now multiple sites are highlighted on the map if multiple dives/sites are selected.
- You can move sites while on the dive-site-table screen.
- Tons of warnings spewed on the console:
```
"Passing incompatible arguments to C++ functions from JavaScript is dangerous and deprecated."
"This will throw a JavaScript TypeError in future releases of Qt!"
"Could not convert argument 0 at"
         "expression for source@qrc:/qml/MapWidget.qml:62"
```

I believe the last point is due to Qt not being smart enough to turn null-QVariants into null-pointers. We'll probably have to turn the strongly typed function into a weakly typed one taking a QVariant. And convert it in the function, because why easy and compile-time evaluated when you can make it hard and run-time evaluated. ☹

bstoeger:
Fixed the warnings by turning a proper typed parameter into a `QVariant`.

@djunkins: If you find time, could you try out if this works for you?

djunkins:
@bstoeger This seems to work to edit the dive site flag location from the dive sites tab with two caveats:

1) Once you click the edit button to edit the dive site fields you can no longer drag the flag to a new location on the map, including after you click the Done button to return to the dive site list. Only after you select another dive site and come back to the original site can you move the flag again.

2) Sites that don't have dives associated with them are still not working. They don't appear on the map and when selecting a site with no dives, you can not edit it's location on the map.

bstoeger:
Sigh - this is so horrible. When clicking the edit button, we reload the map four (sic!) times:

1) Click on the edit button - Qt selects the corresponding row.
2) We instruct the filter to filter on a different dive site.
3) This changes the selection, because the old selection was unselected.
4) The changed selection reloads the map (I).
5) After the filter has done its thing, it sends a filter finished signal.
6) That reloads the map (II).
7) We get the the click and switch into dive-site-edit mode.
8) The dive site tab instructs the filter to filter on the dive site.
9) Filter does its thing, sends a filter-finished signal and reloads the map (III).
10) The dive site list gets its hide signal and stops edit mode.
11) Filter is reset and reloads the map (IV).

10 is the reason why flag dragging does not work anymore.
I will try to improve this mess, but not today.

bstoeger:
Added reference-counting to filter's dive site mode. It's awful, but at least now switching from table to edit view and back should work.

I'll see what is wrong with 0-dives dive sites, but probably not today.

djunkins:
This is working much better for sites that have at least 1 dive associated with them. Interestingly, when you select both a site with dives and one without dive, the map zooms and scrolls to include coverage for the site without dives even though there is no flag there.

dirkhh:
It's so obvious that the original code was written by multiple people with non-identical vision for the code flow - and maybe non-optimal understanding of the underlying mechanics. I recognize some of the abominations that you removed as hacks I did to fix something that I didn't fully understand.
Thank you so much for digging in and cleaning this mess up.
I read through all the commits and can't find anything to bicker about. @djunkins obviously has a point with the dive sites with no dives - but I wonder if we should merge this now and deal with that issue in a follow on PR - in the past long running PRs that we didn't bring into master often turned into rather significant time sinks.
Your call, @bstoeger

bstoeger:
Do not commit yet, please. I think the whole dive selection vs. dive site selection vs. filter vs. map thing is still fundamentally wrong. But currently I have no time to wrap my head around it...

dirkhh:
Marked as `[WIP]` to document that this isn't ready for merge

neolit123:
thank for working on this refactor. i went trough the commits and this seems good.

when the QML map code was created it was a direct port of the old Marble (map globe widget) and for a short period of time we were able to toggle between the two map widgets on compile time. their backends were very similar in terms of public methods and signal/slot connectivity.

the reload-on-anything logic felt wrong, but was cheap and didn't cause any trouble (back then), so i didn't bother changing it.

but seems like the filtering of dive-sites with the combination of dive edit opened the can of worms.

> We probably have to add two filter modes
> 1. show only dive sites of filtered dives
> 2. show all dive sites even though filtering is active

1 is more intuitive, but if it creates complications 2 is fine.
i wouldn't add both modes or a toggle between the two.

bstoeger:
Added commits that at least show unused dive sites on the map when in edit mode.

Still very non-intuitive behavior. For example, editing the name of a dive site does not immediately update the name on the map.

Another problem: Selecting multiple dive sites whereby one has no GPS locations will put the flag of the latter at the old map position, but the map will zoom on the former.

bstoeger:
Name is now updated on the map.

@djunkins: is this usable for you?

It still has some funky issues with undo, etc. But we can take care about them later...

djunkins:
Sorry. I’ve been buried with my day job. I’ll try to build and test this tonight.

djunkins:
@bstoeger This seems to be working well. So far I have not run into any issues with it not working as expected.

bstoeger:
Thanks. There still are UI quirks, but I hope that they are acceptable for a first version. Therefore removing the WIP tag.

dirkhh (approved):
Great series. I would have merged as is... but the typo in the commit message bugs me.
Maybe the best way to deal with this is for me to merge it manually instead of waiting for @bstoeger to force push...

PR #2080, opened 2019-05-01
Desktop user manual update: dive site management

willemferguson:
Include text about dive site management. 5 new figures
are introduced. Writing this enabled me to appreciate just how
much time must have gone into writing the code.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

### Mentions:

@bstoeger @dirkhh


Documentation/user-manual.txt
```diff
@@ -711,12 +711,12 @@ be used).
...
 [icon="images/icons/warning2.png"]
 [WARNING]
-Dive locations are managed as a *separate* part of the dive log.
+Dive locations are managed as a *separate* part of the dive log (See the section entitled xref:S_DiveSiteManagement[Manipulating Dive Site Information]).
 The dive information in the *Notes* and *Equipment* tabs can therefore NOT be edited at the same
 time as the dive site information. Save all the other dive information (e.g.
 divemaster, buddy, protective gear, notes about the dive) by selecting _Apply changes_ on the
 *Notes* tab BEFORE editing the dive site information. Then supply a dive site name in the
-textbox labeled _Location_ on the *Notes* tab.
+textbox labeled _Location_ on the *Notes* tab.
```
bstoeger: There is a slight inconsistency between `labeled` and `labelled`. I believe both are OK (US vs. commonwealth), but perhaps unify in the same chapter.

Documentation/user-manual.txt
```diff
@@ -1407,6 +1407,91 @@ size of the _CSV_ file might cause problems. Importing 100 dives at a time
...
+
+Select the dive site to be merged by right-clicking it. A confirmation
+message is presented (see image above). Clicking the confirmation message merges the selected dive with
+the dive named at the top of the panel. Clicking the confirmation also returns you to the dive sites management panel.
+
+==== Add a dive site
+
+At the top right of the dive sites management table is a round button with a "+". Clicking that button inserts a
+new site entitled "New dive site" into the list (image below). Edit this new site by selecting its edit icon and provide a name,
+coordinates and, if preferred, a description (if required, the new site can also be deleted, e.g. after erroneously
```
bstoeger: Not sure if this should be stated here - but you can also simply undo with CTRL-Z.

bstoeger (approved):
Thank you - I spotted only one (very) minor point.

Documentation/user-manual.txt
willemferguson: When I add a new dive site and hit Ctrl-Z, the new site is not removed, so that operation is not undone. I am probably misunderstanding you. Please explain?

bstoeger: Yes, the problem is that as long as a text-field is active, CTRL-Z goes only to that text field.

Try first ESC to remove focus from the text field and then undo. Alternatively, you can directly select Undo in the application-menu.

I'm not sure if this is a Qt problem or if we're doing something wrong. One day I'll have to analyze how other Qt applications behave.

dirkhh:
Travis is irrelevant for this (and it's again just the issue with the one openSUSE distro update timing out). @bstoeger is this now at the point where you are happy? :-)

bstoeger:
I think it's great, yes.

PR #2082, opened 2019-05-02
Convert air pressure in the Information Tab to be editable

willemferguson:
The Information tab shows the atmospheric pressure. Make this value editable
and also ensure that changes to it are undo-able.

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@bstoeger @dirkhh @janmulder

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1317,9 +1317,10 @@ static struct event *find_previous_event(struct divecomputer *dc, struct event *
 	return previous;
 }
 
-static void fixup_surface_pressure(struct dive *dive)
+pressure_t calculate_surface_pressure(const struct dive *dive)
 {
-	struct divecomputer *dc;
+	const struct divecomputer *dc;
+	pressure_t res;
 	int sum = 0, nr = 0;
 
 	for_each_dc (dive, dc) {
@@ -1328,8 +1329,24 @@ static void fixup_surface_pressure(struct dive *dive)
 			nr++;
 		}
 	}
-	if (nr)
-		dive->surface_pressure.mbar = (sum + nr / 2) / nr;
+	res.mbar = nr ? (sum + nr / 2) / nr : 0;
+	return res;
+}
+
+static void fixup_surface_pressure(struct dive *dive)
+{
+	dive->surface_pressure = calculate_surface_pressure(dive);
+}
+
+/* if the surface pressure in the dive data is redundant to the calculated
+ * value (i.e., it was added by running fixup on the dive) return 0,
+ * otherwise return the air temperature given in the dive */
+pressure_t un_fixup_surface_pressure(const struct dive *d)
+{
+	pressure_t res = d->surface_pressure;
+	if (res.mbar && res.mbar == calculate_surface_pressure(d).mbar)
+		res.mbar = 0;
+	return res;
 }
 
 static void fixup_water_salinity(struct dive *dive)
@@ -1823,7 +1840,8 @@ struct dive *fixup_dive(struct dive *dive)
 		fixup_dive_dc(dive, dc);
 
 	fixup_water_salinity(dive);
-	fixup_surface_pressure(dive);
+	if (!dive->surface_pressure.mbar)
+		fixup_surface_pressure(dive);
 	fixup_meandepth(dive);
 	fixup_duration(dive);
 	fixup_watertemp(dive);
--- a/core/dive.h
+++ b/core/dive.h
@@ -535,6 +535,8 @@ extern bool dive_or_trip_less_than(struct dive_or_trip a, struct dive_or_trip b)
 extern void sort_dive_table(struct dive_table *table);
 extern void sort_trip_table(struct trip_table *table);
 extern struct dive *fixup_dive(struct dive *dive);
+extern pressure_t calculate_surface_pressure(const struct dive *dive);
+extern pressure_t un_fixup_surface_pressure(const struct dive *d);
 extern void fixup_dc_duration(struct divecomputer *dc);
 extern int dive_getUniqID();
 extern unsigned int dc_airtemp(const struct divecomputer *dc);
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -77,6 +77,13 @@ static weight_t get_weight(const char *line)
 	return w;
 }
 
+static pressure_t get_airpressure(const char *line)
+{
+	pressure_t p;
+	p.mbar = lrint(ascii_strtod(line, NULL));
+	return p;
+}
+
 static pressure_t get_pressure(const char *line)
 {
 	pressure_t p;
@@ -245,6 +252,9 @@ static void parse_dive_airtemp(char *line, struct membuffer *str, void *_dive)
 
 static void parse_dive_watertemp(char *line, struct membuffer *str, void *_dive)
 { UNUSED(str); struct dive *dive = _dive; dive->watertemp = get_temperature(line); }
+
+static void parse_dive_airpressure(char *line, struct membuffer *str, void *_dive)
+{ UNUSED(str); struct dive *dive = _dive; dive->surface_pressure = get_airpressure(line); }
 
 static void parse_dive_duration(char *line, struct membuffer *str, void *_dive)
 { UNUSED(str); struct dive *dive = _dive; dive->duration = get_duration(line); }
@@ -981,7 +991,7 @@ static void divecomputer_parser(char *line, struct membuffer *str, void *_dc)
 struct keyword_action dive_action[] = {
 #undef D
 #define D(x) { #x, parse_dive_ ## x }
-	D(airtemp), D(buddy), D(cylinder), D(divemaster), D(divesiteid), D(duration),
+	D(airpressure), D(airtemp), D(buddy), D(cylinder), D(divemaster), D(divesiteid), D(duration),
 	D(gps), D(location), D(notes), D(notrip), D(rating), D(suit),
 	D(tags), D(visibility), D(watertemp), D(weightsystem)
 };
--- a/core/parse-xml.c
+++ b/core/parse-xml.c
@@ -1306,6 +1306,8 @@ static void try_to_fill_dive(struct dive *dive, const char *name, char *buf, str
 		return;
 	if (MATCH("visibility.dive", get_rating, &dive->visibility))
 		return;
+	if (MATCH_STATE("airpressure.dive", pressure, &dive->surface_pressure))
+		return;
 	if (state->cur_ws_index < MAX_WEIGHTSYSTEMS) {
 		if (MATCH("description.weightsystem", utf8_string, &dive->weightsystem[state->cur_ws_index].description))
 			return;
--- a/core/save-git.c
+++ b/core/save-git.c
@@ -425,10 +425,13 @@ static void save_dc(struct membuffer *b, struct dive *dive, struct divecomputer 
  */
 static void create_dive_buffer(struct dive *dive, struct membuffer *b)
 {
+	pressure_t surface_pressure = un_fixup_surface_pressure(dive);
 	if (dive->dc.duration.seconds > 0)
 		put_format(b, "duration %u:%02u min\n", FRACTION(dive->dc.duration.seconds, 60));
 	SAVE("rating", rating);
 	SAVE("visibility", visibility);
+	if (surface_pressure.mbar)
+		SAVE("airpressure", surface_pressure.mbar);
 	cond_put_format(dive->notrip, b, "notrip\n");
 	save_tags(b, dive->tag_list);
 	if (dive->dive_site)
--- a/core/units.h
+++ b/core/units.h
@@ -253,6 +253,17 @@ static inline int mbar_to_PSI(int mbar)
 	return to_PSI(p);
 }
 
+static inline int32_t altitude_to_pressure(int32_t altitude) 	// altitude in mm above sea level
+{						// returns atmospheric pressure in mbar
+	return (int32_t) (1013.0 * exp(- altitude / 7800000.0));
+}
+
+
+static inline int32_t pressure_to_altitude(int32_t pressure)	// pressure in mbar
+{						// returns altitude in mm above sea level
+	return (int32_t) (log(1013.0 / pressure) * 7800000);
+}
+
 /*
  * We keep our internal data in well-specified units, but
  * the input and output may come in some random format. This
--- a/desktop-widgets/command_edit.cpp
+++ b/desktop-widgets/command_edit.cpp
@@ -248,6 +248,27 @@ DiveField EditWaterTemp::fieldId() const
 	return DiveField::WATER_TEMP;
 }
 
+// ***** Atmospheric pressure *****
+void EditAtmPress::set(struct dive *d, int value) const
+{
+	d->surface_pressure.mbar = value > 0 ? (uint32_t)value : 0u;
+}
+
+int EditAtmPress::data(struct dive *d) const
+{
+	return (int)d->surface_pressure.mbar;
+}
+
+QString EditAtmPress::fieldName() const
+{
+	return tr("Atm. pressure");
+}
+
+DiveField EditAtmPress::fieldId() const
+{
+	return DiveField::ATM_PRESS;
+}
+
 // ***** Duration *****
 void EditDuration::set(struct dive *d, int value) const
 {
--- a/desktop-widgets/tab-widgets/TabDiveInformation.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveInformation.cpp
@@ -2,15 +2,24 @@
 #include "TabDiveInformation.h"
 #include "ui_TabDiveInformation.h"
 #include "../tagwidget.h"
+#include "core/units.h"
+#include "core/dive.h"
+#include "desktop-widgets/command.h"
 
 #include <core/qthelper.h>
 #include <core/statistics.h>
 #include <core/display.h>
 
+#define COMBO_CHANGED 0
+#define TEXT_EDITED 1
+
 TabDiveInformation::TabDiveInformation(QWidget *parent) : TabBase(parent), ui(new Ui::TabDiveInformation())
 {
 	ui->setupUi(this);
 	connect(&diveListNotifier, &DiveListNotifier::divesChanged, this, &TabDiveInformation::divesChanged);
+	QStringList atmPressTypes { "mbar", get_depth_unit() ,"use dc"};
+	ui->atmPressType->insertItems(0, atmPressTypes);
+	pressTypeIndex = 0;
 }
 
 TabDiveInformation::~TabDiveInformation()
@@ -32,7 +41,7 @@ void TabDiveInformation::clear()
 	ui->averageDepthText->clear();
 	ui->waterTemperatureText->clear();
 	ui->airTemperatureText->clear();
-	ui->airPressureText->clear();
+	ui->atmPressVal->clear();
 	ui->salinityText->clear();
 }
 
@@ -74,6 +83,17 @@ void TabDiveInformation::updateProfile()
 			" ", current_dive->dc.divemode == FREEDIVE));
 
 	ui->sacText->setText( mean[0] ? SACs : QString());
+
+	if (current_dive->surface_pressure.mbar == 0) {
+		ui->atmPressVal->clear();			// If no atm pressure for dive then clear text box
+	}
+	else {
+
+		ui->atmPressVal->setEnabled(true);
+		QString pressStr;
+		pressStr.sprintf("%d",current_dive->surface_pressure.mbar);
+		ui->atmPressVal->setText(pressStr);		// else display atm pressure
+	}
 }
 
 // Update fields that depend on start of dive
@@ -99,15 +119,15 @@ void TabDiveInformation::updateData()
 	ui->waterTemperatureText->setText(get_temperature_string(current_dive->watertemp, true));
 	ui->airTemperatureText->setText(get_temperature_string(current_dive->airtemp, true));
 
-	if (current_dive->surface_pressure.mbar) 	/* this is ALWAYS displayed in mbar */
-		ui->airPressureText->setText(QString("%1mbar").arg(current_dive->surface_pressure.mbar));
-	else
-		ui->airPressureText->clear();
-
 	if (current_dive->salinity)
 		ui->salinityText->setText(QString("%1g/ℓ").arg(current_dive->salinity / 10.0));
 	else
 		ui->salinityText->clear();
+
+	ui->atmPressType->setEditable(true);
+	ui->atmPressType->setItemText(1, get_depth_unit());  // Check for changes in depth unit (imperial/metric)
+	ui->atmPressType->setEditable(false);
+	ui->atmPressType->setCurrentIndex(0);  // Set the atmospheric pressure combo box to mbar
 }
 
 // This function gets called if a field gets updated by an undo command.
@@ -130,6 +150,9 @@ void TabDiveInformation::divesChanged(dive_trip *trip, const QVector<dive *> &di
 	case DiveField::WATER_TEMP:
 		ui->waterTemperatureText->setText(get_temperature_string(current_dive->watertemp, true));
 		break;
+	case DiveField::ATM_PRESS:
+		ui->atmPressVal->setText(ui->atmPressVal->text().sprintf("%d",current_dive->surface_pressure.mbar));
+		break;
 	case DiveField::DATETIME:
 		updateWhen();
 		break;
@@ -137,3 +160,47 @@ void TabDiveInformation::divesChanged(dive_trip *trip, const QVector<dive *> &di
 		break;
 	}
 }
+
+void TabDiveInformation::on_atmPressType_currentIndexChanged(int index) { updateTextBox(COMBO_CHANGED); }
+
+void TabDiveInformation::on_atmPressVal_editingFinished() { updateTextBox(TEXT_EDITED); }
+
+void TabDiveInformation::updateTextBox(int event) // Either the text box has been edited or the pressure type has changed.
+{                                       // Either way this gets a numeric value and puts it on the text box atmPressVal,
+	pressure_t atmpress = { 0 };    // then stores it in dive->surface_pressure.The undo stack for the text box content is
+	double altitudeVal;             // maintained even though two independent events trigger saving the text box contents.
+	if (current_dive) {
+		switch (ui->atmPressType->currentIndex()) {
+		case 0:		// If an atm pressure has been specified in mbar:
+			if (event == TEXT_EDITED)         // this is only triggered by on_atmPressVal_editingFinished()
+				atmpress.mbar = ui->atmPressVal->text().toInt();    // use the specified mbar pressure
+			break;
+		case 1:		// If an altitude has been specified:
+			if (event == TEXT_EDITED) {	// this is only triggered by on_atmPressVal_editingFinished()
+				altitudeVal = (ui->atmPressVal->text().toFloat());    // get altitude from text box
+				if (prefs.units.length == units::FEET)         // if altitude in feet
+					altitudeVal = feet_to_mm(altitudeVal); // imperial: convert altitude from feet to mm
+				else
+					altitudeVal = altitudeVal * 1000;     // metric: convert altitude from meters to mm
+				atmpress.mbar = altitude_to_pressure((int32_t) altitudeVal); // convert altitude (mm) to pressure (mbar)
+				ui->atmPressVal->setText(ui->atmPressVal->text().sprintf("%d",atmpress.mbar));
+				ui->atmPressType->setCurrentIndex(0);    // reset combobox to mbar
+			} else { // i.e. event == COMBO_CHANGED, that is, "m" or "ft" was selected from combobox
+				ui->atmPressVal->clear();   // Clear the text box so that altitude can be typed
+			}
+			break;
+		case 2:          // i.e. event = COMBO_CHANGED, that is, the option "Use dc" was selected from combobox
+			atmpress = calculate_surface_pressure(current_dive);	// re-calculate air pressure from dc data
+			ui->atmPressVal->setText(QString::number(atmpress.mbar)); // display it in text box
+			ui->atmPressType->setCurrentIndex(0);          // reset combobox to mbar
+			break;
+		default:
+			atmpress.mbar = 1013;    // This line should never execute
+			break;
+		}
+		if (atmpress.mbar)
+			Command::editAtmPress(atmpress.mbar, false);      // and save the pressure for undo
+	}
+}
+
+
--- a/dives/TestAtmPress.xml
+++ b/dives/TestAtmPress.xml
@@ -1,0 +1,50 @@
+<divelog program='subsurface' version='3'>
+<settings>
+  <autogroup state='1' />
+</settings>
+<divesites>
+<site uuid='4fdcf19b' name='Loc1'>
+</site>
+<site uuid='b15ad032' name='Loc2'>
+</site>
+</divesites>
+<dives>
+<trip date='2019-04-09' time='09:15:16' location='Scottburgh'>
+<dive number='524' tripflag='NOTRIP' divesiteid='b15ad032' date='2019-04-09' time='09:15:16' airpressure='1.012 bar' duration='42:08 min'>
+  <cylinder size='10.0 l' workpressure='232.0 bar' description='10ℓ 232 bar' o2='30.0%' depth='42.96 m' />
+  <cylinder size='10.0 l' workpressure='232.0 bar' description='10ℓ 232 bar' o2='30.0%' depth='42.96 m' />
+  <weightsystem weight='3.0 kg' description='belt' />
+  <divecomputer model='Uwatec Galileo Trimix' deviceid='462769ce' diveid='14106bf6'>
+  <depth max='30.068 m' mean='21.394 m' />
+  <temperature air='24.6 C' water='23.2 C' />
+  <water salinity='1025 g/l' />
+  <sample time='0:00 min' depth='0.0 m' temp='24.4 C' rbt='99:00 min' />
+  <sample time='0:04 min' depth='2.439 m' temp='24.8 C' rbt='99:00 min' />
+  <sample time='0:08 min' depth='3.395 m' temp='24.4 C' rbt='99:00 min' />
+  <sample time='0:12 min' depth='3.902 m' temp='24.8 C' pressure='208.5 bar' rbt='99:00 min' />
+  <sample time='45:16 min' depth='4.488 m' pressure='208.25 bar' rbt='98:00 min' />
+  </divecomputer>
+</dive>
+<dive number='525' divesiteid='4fdcf19b' date='2019-04-10' time='08:14:01' airpressure='0.991 bar' duration='41:24 min'>
+  <cylinder size='10.0 l' workpressure='232.0 bar' description='10ℓ 232 bar' o2='31.0%' depth='41.25 m' />
+  <cylinder description='unknown' o2='31.0%' depth='41.25 m' />
+  <weightsystem weight='3.0 kg' description='belt' />
+  <divecomputer model='Uwatec Galileo Trimix' deviceid='462769ce' diveid='04c0cf94'>
+  <depth max='26.166 m' mean='19.561 m' />
+  <temperature air='24.8 C' water='23.6 C' />
+  <water salinity='1025 g/l' />
+  <sample time='0:00 min' depth='0.0 m' temp='26.8 C' rbt='99:00 min' />
+  <sample time='0:04 min' depth='3.61 m' rbt='99:00 min' />
+  <sample time='0:08 min' depth='4.741 m' temp='25.6 C' pressure='217.25 bar' rbt='99:00 min' />
+  <sample time='0:12 min' depth='6.478 m' pressure='216.75 bar' rbt='97:00 min' />
+  <sample time='0:16 min' depth='7.785 m' rbt='94:00 min' />
+  <sample time='44:04 min' depth='0.0 m' rbt='99:00 min' />
+  <sample time='44:08 min' depth='0.02 m' rbt='99:00 min' />
+  <sample time='44:12 min' depth='0.098 m' rbt='99:00 min' />
+  <sample time='44:16 min' depth='0.059 m' rbt='99:00 min' />
+  <sample time='44:20 min' depth='0.059 m' rbt='99:00 min' />
+  </divecomputer>
+</dive>
+</trip>
+</dives>
+</divelog>
--- a/tests/testAirPressure.cpp
+++ b/tests/testAirPressure.cpp
@@ -1,0 +1,63 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "testAirPressure.h"
+#include "core/divesite.h"
+#include "core/divelist.h"
+#include "core/file.h"
+#include "core/dive.h"
+#include <QString>
+#include <core/qthelper.h>
+
+void TestAirPressure::initTestCase()
+{
+	/* we need to manually tell that the resource exists, because we are using it as library. */
+	Q_INIT_RESOURCE(subsurface);
+}
+
+void TestAirPressure::get_dives()
+{
+	struct dive *dive;
+	verbose = 1;
+
+	QCOMPARE(parse_file(SUBSURFACE_TEST_DATA "/dives/TestAtmPress.xml", &dive_table, &trip_table, &dive_site_table), 0);
+	dive = get_dive(0);
+	dive->selected = true;
+	QVERIFY(dive != NULL);
+}
+
+void TestAirPressure::testReadAirPressure()
+{
+	struct dive *dive;
+	dive = get_dive(0);
+	QVERIFY(dive != NULL);
+	dive->selected = true;
+	QCOMPARE(1012, dive->surface_pressure.mbar);
+	dive = get_dive(1);
+	QVERIFY(dive != NULL);
+	dive->selected = true;
+	QCOMPARE(991, dive->surface_pressure.mbar);
+}
+
+void TestAirPressure::testConvertAltitudetoAirPressure()
+{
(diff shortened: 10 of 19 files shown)


desktop-widgets/command_edit.cpp
```diff
@@ -248,6 +249,29 @@ DiveField EditWaterTemp::fieldId() const
 	return DiveField::WATER_TEMP;
 }
 
+// ***** Atmospheric pressure *****
+void EditAtmPress::set(struct dive *d, int value) const
+{
+	divecomputer *dc = get_dive_dc(d,dc_number); 
+	dc->surface_pressure.mbar = value > 0 ? (uint32_t)value : 0u;
+	fixup_surface_pressure(d);  // create new dive->surface-pressure
```
bstoeger: Note that this has strange semantics if you have more than one DC:
You display the average pressure of all DCs, but edit only the current DC.

That means to actually set a given value, the user has to flip through all DCs.

willemferguson: Given that dive->surface_pressure is a calculated value and not an original data item, this is the only way of doing it in practice. There are a few others that logically should behave in a similar way, e.g. airtemp which also supplied by the dive computer. This would mean that, when viewing the Information tab, at least airtemp, watertemp and surface_pressure represent the values of the currently-shown dive computer. I need to indicate that in the user manual. They all appear to have a similar logic with equivalent fixup() functions ind dive.c. I hope my interpretation is correct.

bstoeger: I think that air-temperature currently behaves fundamentally different: This is saved with the dive. Also note that `fixup_airtemp` only acts if the temperature for the dive is not yet set, so a user-provided value will not be overwritten. Finally, it only seems to use the temperature of the first dive computer, not an average.

I see different ways of handling this:
1) Making pressure behave more like temperature, i.e. save it to the file and don't overwrite it with dive-computer values if the user changed it.
2) Show a non-editable pressure average over all dive computers and an per-divecomputer editable value.

Since I wouldn't dive with more than one DC (Segal's law 😁), I don't have an opinion on the matter.

willemferguson: I would definitely prefer the first option.  This would mean that the dc-specific values are not viewable if more than one dc is used. However, dc-specific data are still stored in the XML and are, in principle available to be processed.

willemferguson: The code has been changed so that: If the used edits surface pressure in the Information tab, it is saved as dive.surface_pressure. If the dive is saved it is stored in XML as a dive-related attribute. Upon initialising a dive from XML, if dive.surface_pressure has a value, this is used and not calculated from the dive computer. The use case for editing the surface pressure is for dives at altitude when the dive computer does not measure atmospheric pressure. In the corner case when the dive computer does measure pressure and the user modifies the pressure, this is stored and saved as dive.surface_pressure, but the pressure derived from dive computer(s) can be recovered and stored in dive.surface_pressure by typing a zero-value in the text box in the Information tab.

core/dive.c
```diff
@@ -453,6 +453,20 @@ double get_weight_units(unsigned int grams, int *frac, const char **units)
...
 
+int32_t altitude_to_pressure(int altitude) 		// altitude in meters above sea level
+{						// function returns atmospheric pressure in mbar
+	int i, alt;
+	static const int pressval[21] = { 1013, 984, 955, 926, 899, 872, 846, 820, 795, 771, 747, 724, 701, 679, 658, 637, 616, 597, 577, 559, 540 };
+	// Calculated atmospheric pressure values at 250m increments in altitude from sea level to 4999m,
+	// calculated following the US Standard Atmosphere 1976, US Government Printing Office, Washington DC, 1976.
+	// For intermediate altitudes, linear interpolation is performed here.
+	alt = altitude;
+	if (altitude < 0) alt = 0;
```
bstoeger: Minor nit-pick: The dead sea goes down to -430 m (growing).

bstoeger (approved):
LGTM - very well done.

The only thing I noticed is that maybe that altitude-to-pressure function should be in units.c instead of dive.c?

core/dive.c
willemferguson: Implemented this. IMPORTANT: this function cannot go to units.c because there is no such file. Unit.h comprises typedefs and simple definitions. Moving the altitude_to_pressure() function there causes untold problems. I hope my interpretation is correct?

bstoeger: > Implemented this.

Nice, thank you.

> IMPORTANT: this function cannot go to units.c because there is no such file. 

Ok, perhaps we should one day generate such a file, but not now.

dirkhh:
The failed tests are caused by the change to the file format. So this is a case where we need to fix the tests :-)

dirkhh (changes requested):
A couple small comments in the code.
This has support for saving to / reading from XML (need to fix the test, though)
It's lacking support for saving to / reading from git (that's critical to have, of course)

core/dive.h
```diff
@@ -390,6 +390,7 @@ extern int depth_to_mbar(int depth, const struct dive *dive);
 extern double depth_to_bar(int depth, const struct dive *dive);
 extern double depth_to_atm(int depth, const struct dive *dive);
 extern int rel_mbar_to_depth(int mbar, const struct dive *dive);
+extern int32_t altitude_to_pressure(int altitude);
```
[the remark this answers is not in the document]
willemferguson: I get this when the function is in units.h. I have no idea what the linker is trying to communicate. I do not see any multiple definitions:
`~/src/subsurface$ grep -r "altitude_to_pressure("

core/units.h:int32_t altitude_to_pressure(int altitude) 		// altitude in meters above sea level (range -500 to 4999)
desktop-widgets/tab-widgets/TabDiveInformation.cpp:			atmpress = altitude_to_pressure(atmpress);	// convert altitude to pressure
`

`[ 78%] Linking CXX executable TestQPrefUnits
[ 78%] Linking CXX executable TestQPrefProxy
[ 78%] Linking CXX executable `TestQPrefUpdateManager`
../core/libsubsurface_corelib.a(subsurfacestartup.c.o): In function `altitude_to_pressure':
/home/willem/src/subsurface/core/units.h:267: multiple definition of `altitude_to_pressure'
CMakeFiles/TestQPrefPartialPressureGas.dir/testqPrefPartialPressureGas.cpp.o:/home/willem/src/subsurface/./core/units.h:267: first defined here
../core/libsubsurface_corelib.a(qPrefPartialPressureGas.cpp.o): In function `altitude_to_pressure':
/home/willem/src/subsurface/./core/units.h:267: multiple definition of `altitude_to_pressure'
CMakeFiles/TestQPrefPartialPressureGas.dir/testqPrefPartialPressureGas.cpp.o:/home/willem/src/subsurface/./core/units.h:267: first defined here
../core/libsubsurface_corelib.a(qPrefPrivate.cpp.o): In function `altitude_to_pressure':
/home/willem/src/subsurface/./core/units.h:267: multiple definition of `altitude_to_pressure'
CMakeFiles/TestQPrefPartialPressureGas.dir/testqPrefPartialPressureGas.cpp.o:/home/willem/src/subsurface/./core/units.h:267: first defined here
../core/libsubsurface_corelib.a(unix.c.o): In function `altitude_to_pressure':
/home/willem/src/subsurface/core/units.h:267: multiple definition of `altitude_to_pressure'
`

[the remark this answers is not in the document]
bstoeger: If you put a function in a header file, you have to mark it as `inline`.

willemferguson:
Git access implemented.
Code style corrected.

core/units.h
```diff
@@ -253,6 +253,21 @@ static inline int mbar_to_PSI(int mbar)
 	return to_PSI(p);
 }
 
+static inline int32_t altitude_to_pressure(int altitude) 	// altitude in meters above sea level (range -500 to 4999)
```
atdotde: Sorry, I am very late to the party, but: In the planner, we already have functionality to enter surface pressure either as an actual pressure or as an altitude (assuming standard pressure at sea level), see dive planner.cpp around lines 225ff. Why do it differently here? If this is superior, you should change the behaviour in the planner as well

core/parse-xml.c
```diff
@@ -535,6 +535,11 @@ static void get_uint16(char *buffer, uint16_t *i)
 	*i = atoi(buffer);
 }
 
+static void get_uint32(char *buffer, int32_t *i)
```
atdotde: What is your reason to use uint32 and not good old int as we do everywhere else?

atdotde (changes requested):

core/units.h
bstoeger: For those not familiar with the planner: Could you state which function transforms altitude to pressure? I only see `depth_to_atm`, but that is depth under water, isn't it?

core/parse-xml.c
bstoeger: That's apparently a typo, as the output parameter is `int32_t`, not `uint32_t`.

core/units.h
willemferguson: Of course I did not think at all of the implementation in the planner (and I use the planner regularly!). The philosophy behind my design of the converter was that it should not take more screen space than the QString that was always displayed there (i.e. the non-editable version). The implementation in the planner takes double the space but it more user friendly. With my implementation the conversion is dependent on the user selecting the unit (mbar/m) first before typing the number. The conversion is triggered when the user types Enter after typing the number to be converted. The user can type in the number, then select the unit but selecting unit does not do the conversion: click on the number and hit Enter will do it.
For Robert's calculation see planner.cpp line 252.

core/parse-xml.c
willemferguson: Typo

dirkhh:
...you still need to fix the tests :-)

core/save-git.c
```diff
@@ -194,11 +194,6 @@ static void save_temperatures(struct membuffer *b, struct divecomputer *dc)
 	put_temperature(b, dc->watertemp, "watertemp ", "°C\n");
 }
 
-static void save_airpressure(struct membuffer *b, struct divecomputer *dc)
-{
-	put_pressure(b, dc->surface_pressure, "surfacepressure ", "bar\n");
-}
-
```
dirkhh: so we saved the pressure before, now we save it differently. Can you elaborate why this is necessary?

core/units.h
atdotde: > For those not familiar with the planner: Could you state which function transforms altitude to pressure? I only see `depth_to_atm`, but that is depth under water, isn't it?

It doesn't have its own function but it is done here:
https://github.com/Subsurface-divelog/subsurface/blob/14a763a6a0148a42fad06717bd2532181a54c626/desktop-widgets/diveplanner.cpp#L252
and in reverse here:
https://github.com/Subsurface-divelog/subsurface/blob/14a763a6a0148a42fad06717bd2532181a54c626/desktop-widgets/diveplanner.cpp#L238

core/save-git.c
bstoeger: See the discussion here:
https://github.com/Subsurface-divelog/subsurface/pull/2082#discussion_r280376552
https://github.com/Subsurface-divelog/subsurface/pull/2082#discussion_r280669523

The old code saved the value of the *dive computer*. The point here is to save the value entered by the *user*.

core/units.h
atdotde: I wasn't so much thinking about the UI presentation but more about the computation.

core/parse-xml.c
atdotde: Fine, but why a fixed precision type?

core/units.h
bstoeger: > It doesn't have its own function but it is done here

Thanks. Indeed, both places should use the same global function.

core/parse-xml.c
bstoeger: Because that reflects the `pressure_t` type. That makes sense to me.

core/save-git.c
bstoeger: Of course, the air pressure of the dc should still be saved, so this part of the code shouldn't be removed...

willemferguson: Real use case: I dive my Galileo at 1200m altitude. Galileo does not report atm. pressure. I can now type in the altitude, gets converted to mbar and saved as the dive.surface_pressure. On the other hand if I use my Petrel I do not want to be in the situation as to override the surface_pressure reported by the Petrel. The original datum is too valuable. I am considering disabling the write access to the text box and combobox if the dive computer has a surface_pressure.

core/units.h
willemferguson: I have no particular allegiance to the database from where I do my calculations. In this case the physicist amongst us can only have by far the best opinion.

core/save-git.c
bstoeger: @willemferguson: The problem is that here you remove saving the value reported by the dive computer(s). This code should stay in - you only want to write the user supplied value *in addition*.

core/units.h
willemferguson: Robert, you have probably seen the inline function in units.h. I think you are in by far the best position to make a recommendation about the calculation. I am fairly certain your calculation is from first principles and I am quite positive towards such an approach. ok?

core/save-git.c
willemferguson: Ha, I understand

core/units.h
atdotde: Have a look at 
image The dots are your datapoint, the solid line is the formula from diveplanner.cpp line 252 (barometric formula like https://en.wikipedia.org/wiki/Barometric_formula ). Looks good enough for me.

atdotde: Note that my formula is for altitudes in mm (as all internal length values in Subsurface). I guess you should use mm as well and only convert them to other units when displaying to the user (using the usual helper functions for depth).

willemferguson: Using mm and helpers ok. If you were to look at the existing two ways of doing the calculation, which would be your preferred? I suggest using your preferred, which I will write as an inline function in units.h and using it in the Information tab as well as in the planner.

atdotde: My personal taste is always in favour of a formula instead of a look-up table. And yes, please make this a helper function (as well as the reverse) and use that everywhere.

willemferguson:
New altitude-to-pressure functions implemented in units.h (inverse function as well).
Omission in save-git.c corrected.
Test code added. Here is output from test code:
********* Start testing of TestAirPressure *********
Config: Using QtTest library 5.9.5, Qt 5.9.5 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 7.3.0)
PASS   : TestAirPressure::initTestCase()
PASS   : TestAirPressure::get_dives()
PASS   : TestAirPressure::testReadAirPressure()
PASS   : TestAirPressure::testWriteReadBackAirPressure()
PASS   : TestAirPressure::cleanupTestCase()
Totals: 5 passed, 0 failed, 0 skipped, 0 blacklisted, 17ms
********* Finished testing of TestAirPressure *********

core/dive.c
```diff
@@ -1328,8 +1328,12 @@ static void fixup_surface_pressure(struct dive *dive)
 			nr++;
 		}
 	}
-	if (nr)
+	if (nr) {
 		dive->surface_pressure.mbar = (sum + nr / 2) / nr;
+		dive->userflags = true;
+	}
+	else
```
dirkhh: well, that doesn't work with our coding style, right?
```
} else { 
      dive->userflags = false;
}
```

core/dive.c
```diff
@@ -1328,8 +1328,12 @@ static void fixup_surface_pressure(struct dive *dive)
 			nr++;
 		}
 	}
-	if (nr)
+	if (nr) {
 		dive->surface_pressure.mbar = (sum + nr / 2) / nr;
+		dive->userflags = true;
```
dirkhh: so this new `userflags` member puzzles me. It's a plural name, but just a boolean value. And what exactly does it tell me. That the surface pressure is calculated, not set? So should this be `dive->surface_pressure_calculated` ?

core/load-git.c
```diff
@@ -246,6 +253,9 @@ static void parse_dive_airtemp(char *line, struct membuffer *str, void *_dive)
 static void parse_dive_watertemp(char *line, struct membuffer *str, void *_dive)
 { UNUSED(str); struct dive *dive = _dive; dive->watertemp = get_temperature(line); }
 
+static void parse_dive_airpressure(char *line, struct membuffer *str, void *_dive)
+{ UNUSED(str); struct dive *dive = _dive; dive->surface_pressure = get_airpressure(line); }
```
dirkhh, line 257: wouldn't we have to set that `userflag` or better `surface_pressure_calculated` flag to `false` here?
I guess we could rely on it being zero initialized (I'd have to spend more time thinking through the code to make sure that this is indeed the case)

core/parse-xml.c
```diff
@@ -1306,6 +1306,8 @@ static void try_to_fill_dive(struct dive *dive, const char *name, char *buf, str
 		return;
 	if (MATCH("visibility.dive", get_rating, &dive->visibility))
 		return;
+	if (MATCH_STATE("airpressure.dive", pressure, &dive->surface_pressure))
```
dirkhh, line 1309: same comment as before about the new flag... this may be correct if we are sure this is initialized to `false`

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -74,6 +79,18 @@ void TabDiveInformation::updateProfile()
 			" ", current_dive->dc.divemode == FREEDIVE));
 
 	ui->sacText->setText( mean[0] ? SACs : QString());
+
+	if (current_dive->surface_pressure.mbar == 0)
+		ui->atmPressVal->clear();			// If no atm pressure for dive then clear text box
+	else {
```
dirkhh, line 90: again, coding style. If one side of an `else` requires braces, we force braces on both sides.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -130,10 +147,35 @@ void TabDiveInformation::divesChanged(dive_trip *trip, const QVector<dive *> &di
...
 		break;
 	default:
 		break;
 	}
 }
+
+void TabDiveInformation::on_atmPressVal_editingFinished()
+{
+	int32_t atmpress;
+	if ((current_dive) && (!current_dive->userflags))
```
dirkhh: Hmm - I may be misunderstanding the code... if the pressure was calculated from the dive computer(s), then you aren't able to overwrite it here. Is that what you intended?
Wouldn't it be more logical to not allow editing instead of silently ignoring the value the user entered? Or is this doing something else and I'm just confused?

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -130,10 +147,35 @@ void TabDiveInformation::divesChanged(dive_trip *trip, const QVector<dive *> &di
...
+	{
+		if (ui->atmPressType->currentIndex() == 1) {		// If altitude has been specified:
+			atmpress = (int)(0.5 + ui->atmPressVal->text().toFloat());	// get altitude from text box
+			if (prefs.units.length == units::FEET)		// if altitude in feet
+				atmpress = (int)(atmpress / 3.28084); 	// 	convert feet -> meters
```
dirkhh: we have a `feet_to_mm()` helper. I'd prefer to use that then to hardcode another value here in the middle of the code

dives/TestAtmPress.xml
```diff
@@ -0,0 +1,1385 @@
...
+  <event time='18:00 min' type='25' flags='2' name='gaschange' cylinder='1' o2='30.0%' />
+  <event time='33:36 min' type='25' flags='1' name='gaschange' cylinder='0' o2='30.0%' />
+  <event time='34:17 min' type='8' name='Deepstop' />
+  <event time='37:36 min' type='8' name='bookmark' />
+  <sample time='0:00 min' depth='0.0 m' temp='24.4 C' rbt='99:00 min' />
+  <sample time='0:04 min' depth='2.439 m' temp='24.8 C' rbt='99:00 min' />
+  <sample time='0:08 min' depth='3.395 m' temp='24.4 C' rbt='99:00 min' />
+  <sample time='0:12 min' depth='3.902 m' temp='24.8 C' pressure='208.5 bar' rbt='99:00 min' />
+  <sample time='0:16 min' depth='4.488 m' pressure='208.25 bar' rbt='98:00 min' />
+  <sample time='0:20 min' depth='5.795 m' rbt='97:00 min' />
```
dirkhh: This adds ~1300 lines of samples... but I don't think those samples are really relevant or interesting for the purpose of this test. Wouldn't it make more sense to have just 5 samples for a simple profile and make it easy to spot what matters here, the `airpressure`?

tests/testAirPressure.cpp
```diff
@@ -0,0 +1,56 @@
...
+	dive->selected = true;
+	QVERIFY(dive != NULL);
+}
+
+void TestAirPressure::testReadAirPressure()
+{
+	struct dive *dive;
+	dive = get_dive(0);
+	dive->selected = true;
+	QVERIFY(dive != NULL);
```
dirkhh, line 31: Well, maybe we should do this test before dereferencing `dive` in the line above?

tests/testAirPressure.cpp
```diff
@@ -0,0 +1,56 @@
...
+	QCOMPARE(QString::number(1012), QString::number(dive->surface_pressure.mbar));
+	dive = get_dive(1);
+	dive->selected = true;
+	QVERIFY(dive != NULL);
```
dirkhh, line 35: ditto for ordering of check / dereferencing

tests/testAirPressure.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+void TestAirPressure::testReadAirPressure()
+{
+	struct dive *dive;
+	dive = get_dive(0);
+	dive->selected = true;
+	QVERIFY(dive != NULL);
+	QCOMPARE(QString::number(1012), QString::number(dive->surface_pressure.mbar));
```
dirkhh: `QCOMPARE()` happily compares numbers. Why the conversion to strings?

tests/testAirPressure.cpp
```diff
@@ -0,0 +1,56 @@
...
+	dive = get_dive(1);
+	dive->selected = true;
+	QVERIFY(dive != NULL);
+	if (dive->surface_pressure.mbar)
+	QCOMPARE(QString::number(991), QString::number(dive->surface_pressure.mbar));
```
dirkhh: ditto for `QCOMPARE()` and integers

tests/testAirPressure.cpp
```diff
@@ -0,0 +1,56 @@
...
+void TestAirPressure::testWriteReadBackAirPressure()
+{
+	struct dive *dive;
+	int32_t ap = 1111;
+	dive = get_dive(0);
+	dive->selected = true;
+	QVERIFY(dive != NULL);
+	dive->surface_pressure.mbar = ap;
+	QCOMPARE(save_dives("./testout.ssrf"), 0);
+	QCOMPARE(parse_file("./testout.ssrf", &dive_table, &trip_table, &dive_site_table), 0);
```
dirkhh, line 56: since you didn't clear the data in between, this actually tests what happens when you merge the file into itself. Maybe not what you had in mind?

tests/testAirPressure.cpp
```diff
@@ -0,0 +1,56 @@
...
+	dive = get_dive(1);
+	dive->selected = true;
+	QVERIFY(dive != NULL);
+	if (dive->surface_pressure.mbar)
+	QCOMPARE(QString::number(991), QString::number(dive->surface_pressure.mbar));
+}
+
+void TestAirPressure::testWriteReadBackAirPressure()
+{
+	struct dive *dive;
```
dirkhh, line 48: just like above, you have the ordering of dereferencing and non-null tests as well as the odd conversion into string before using `QCOMPARE()`. Oddly, you are happy to use `QCOMPARE()` on numbers when you check the return value of functions... :-)

dirkhh (changes requested):
Thanks for your continued work on this.
I feel bad to once again request changes, but there are a few things here that need some more work. Coding style issues, and several issues in the way you created the test.
Additionally, the name of the flag and then some questions about the use of the flag...

tests/testAirPressure.cpp
willemferguson: Do not feel bad at all. This is what is required to get the code up to standard. Your comments are appreciated. Maybe we should talk about the flag. It is an extra variable in the dive structure and I would like to limit adding extra variables to that structure. I do not want to run ahead of things and try to limit the discussion to the airpressure, but it is possible that other similar coding might appear in the near future. I am thinking of salinity in the case of manually-entered dives. I need to get your opinion about adding more booleans if other similar cases in coding do appear in future. My idea was basically to turn the userflag into a byte and do some bit-setting and bit-reading using | and & operators and thus avoid having to add additional variables to the dive structure. If you are not too worried about adding a couple of additional booleans if the need may arise in future, then probably using a byte is not necessary. I am affected by my programming past where every byte counted. What is your feeling?

I am working on your other requests.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: At the moment, if the airpressure was supplied by the dive computer, the edit box is non-editable and the combobox is disabled. I am heavily influenced by my interaction with the darktable software where the original photo is sacrosanct: you never change anything about the original photo and any editing is in xml outside the original. In the same way I am biased towards keeping the downloaded dive log as "pure" as possible. If you have other feelings I would be more than happy to fit in. See line 94 of this file.

bstoeger: > At the moment, if the airpressure was supplied by the dive computer, the edit box is non-editable and the combobox is disabled. I am heavily influenced by my interaction with the darktable software where the original photo is sacrosanct: you never change anything about the original photo and any editing is in xml outside the original.

I'm not sure that the analogy is fitting. After all, the whole point of darktable is to edit the image, just that the original is left untouched. Making the field editable without overwriting the original data would be the proper equivalent.

I see the use cases for editing the pressure makes sense: For example if you have a dive computer with a broken air-pressure sensor.

tests/testAirPressure.cpp
bstoeger: > Maybe we should talk about the flag. It is an extra variable in the dive structure and I would like to limit adding extra variables to that structure.

I'm not saying that the flag is wrong. I'd just like to note that for air temperature, there is a different mechanism: There is an `un_fixup_airtemp` function that returns 0 if the airtemp corresponds to the one supplied by the dive computer.

The analog would look something like this:
```
diff --git a/core/dive.c b/core/dive.c
index 114ca31db..55498ebdd 100644
--- a/core/dive.c
+++ b/core/dive.c
@@ -1317,9 +1317,10 @@ static struct event *find_previous_event(struct divecomputer *dc, struct event *
 	return previous;
 }
 
-static void fixup_surface_pressure(struct dive *dive)
+static pressure_t calculate_surface_pressure(const struct dive *dive)
 {
-	struct divecomputer *dc;
+	pressure_t res;
+	const struct divecomputer *dc;
 	int sum = 0, nr = 0;
 
 	for_each_dc (dive, dc) {
@@ -1328,8 +1329,13 @@ static void fixup_surface_pressure(struct dive *dive)
 			nr++;
 		}
 	}
-	if (nr)
-		dive->surface_pressure.mbar = (sum + nr / 2) / nr;
+	res.mbar = nr ? (sum + nr / 2) / nr : 0;
+	return res;
+}
+
+static void fixup_surface_pressure(struct dive *dive)
+{
+	dive->surface_pressure = calculate_surface_pressure(dive);
 }
 
 static void fixup_water_salinity(struct dive *dive)
@@ -1424,6 +1430,17 @@ static void fixup_airtemp(struct dive *dive)
 		dive->airtemp.mkelvin = dc_airtemp(&dive->dc);
 }
 
+/* if the surface pressure in the dive data is redundant to the calculated
+ * value (i.e., it was added by running fixup on the dive) return 0,
+ * otherwise return the air temperature given in the dive */
+pressure_t un_fixup_surface_pressure(const struct dive *d)
+{
+	pressure_t res = d->surface_pressure;
+	if (res.mbar && res.mbar == calculate_surface_pressure(d).mbar)
+		res.mbar = 0;
+	return res;
+}
+
 /* if the air temperature in the dive data is redundant to the one in its
  * first divecomputer (i.e., it was added by running fixup on the dive)
  * return 0, otherwise return the air temperature given in the dive */
diff --git a/core/dive.h b/core/dive.h
index 7a6cf4099..f1f2cc36b 100644
--- a/core/dive.h
+++ b/core/dive.h
@@ -536,6 +536,7 @@ extern void sort_dive_table(struct dive_table *table);
 extern void sort_trip_table(struct trip_table *table);
 extern struct dive *fixup_dive(struct dive *dive);
 extern void fixup_dc_duration(struct divecomputer *dc);
+extern pressure_t un_fixup_surface_pressure(const struct dive *d);
 extern int dive_getUniqID();
 extern unsigned int dc_airtemp(const struct divecomputer *dc);
 extern unsigned int dc_watertemp(const struct divecomputer *dc);
diff --git a/core/save-git.c b/core/save-git.c
index ffbeca1d0..c6cb0c969 100644
--- a/core/save-git.c
+++ b/core/save-git.c
@@ -419,11 +419,12 @@ static void save_dc(struct membuffer *b, struct dive *dive, struct divecomputer
  */
 static void create_dive_buffer(struct dive *dive, struct membuffer *b)
 {
+	pressure_t surface_pressure = un_fixup_surface_pressure(dive);
 	if (dive->dc.duration.seconds > 0)
 		put_format(b, "duration %u:%02u min\n", FRACTION(dive->dc.duration.seconds, 60));
 	SAVE("rating", rating);
 	SAVE("visibility", visibility);
-	if (dive->surface_pressure.mbar)
+	if (surface_pressure.mbar)
 		SAVE("airpressure", surface_pressure.mbar);
 	cond_put_format(dive->notrip, b, "notrip\n");
 	save_tags(b, dive->tag_list);
diff --git a/core/save-xml.c b/core/save-xml.c
index 027c6f4f2..8b5e3a7e6 100644
--- a/core/save-xml.c
+++ b/core/save-xml.c
@@ -473,6 +473,7 @@ static void save_picture(struct membuffer *b, struct picture *pic)
 void save_one_dive_to_mb(struct membuffer *b, struct dive *dive, bool anonymize)
 {
 	struct divecomputer *dc;
+	pressure_t surface_pressure = un_fixup_surface_pressure(dive);
 
 	put_string(b, "<dive");
 	if (dive->number)
@@ -488,8 +489,8 @@ void save_one_dive_to_mb(struct membuffer *b, struct dive *dive, bool anonymize)
 		put_format(b, " divesiteid='%8x'", dive->dive_site->uuid);
 	}
 	show_date(b, dive->when);
-	if (dive->surface_pressure.mbar)
-		put_format(b, " airpressure='%d'", dive->surface_pressure.mbar);
+	if (surface_pressure.mbar)
+		put_format(b, " airpressure='%d'", surface_pressure.mbar);
 	if (dive->dc.duration.seconds > 0)
 		put_format(b, " duration='%u:%02u min'>\n",
 			   FRACTION(dive->dc.duration.seconds, 60));
```

But let me reiterate that I'm *not* saying that this is preferred over a flag. I only want to mention that as an alternative...

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: A broken pressure sensor would render the dive computer unusable because it would not be able to read depth (well, more strictly, water pressure). But, Berthold I am totally open-minded about this. I would like to hear Dirk's or any one else's opinion. If one can over-write dive->surface_pressure even when dc->surface_pressure also exists, it would eliminate the need for any flag and it would simplify things. I used to have a facility that if one typed "0" into the airpressure text box and if dc->surface_pressure does exist, it puts dc->surface_pressure back into the text box. I removed this but can easily put it back. It was intended as mechanism for getting back the dc->surface_pressure after one has over-typed airpressure in the text box.

bstoeger: Ok - I thought the altimeter was a different sensor, as this is a value that is notoriously hard to measure.

Another use case might be for those who use more than one dive computer and trust one over than the other. Or perhaps for people who care more about the geographic altitude than the actual pressure.

willemferguson: Thanks for this, Berthold. Unless there are any other opinions, I will make dive->surface_pressure editable, even if dc->surface_pressure does exist. Would this be ok?

willemferguson:
This is now fully functional except for one undo setup that is still required.
Dive surface pressure can now always be edited.
Combobox has 3rd option "Use dc". If this is selected, the surface_pressure from the dc is inserted into the text box.
Programming style corrections made.
Test program re-ordered in more logical flow of instructions.
Test program number-strings changed: QCOMPARE now compares numbers, not strings.

Berthold, I need your help with setting up the undo for on_atmPressType_currentIndexChanged() in TabDiveInformation.cpp. I left a commented instruction there. The difficulty is that, using the combobox selection, I want to implement undo for the text int the text box. So its no use writing a set of undo functions for the combo box. The undo has to happen for the text in the text box. I hope this sounds halfway understandable. If the solution is basic, please give me a few pointers that I can implement.

tests/testAirPressure.cpp
dirkhh: I really like this... better than carrying more flags. And of course more consistent

willemferguson:
I have now included Berthold's suggestions with respect to unfixup_air_pressure and the code around this.

willemferguson:
These changes appear to fix the undo stack for the atm pressure text box.
The challenge was to channel two totally independent events QComboBox indexchange and QLineEdit editingFinished into a single stream within a single undo stack for the QLineEdit. I am convinced this is a very hamfisted way of doing it, but this was the only solution I could think of. Berthold, I would appreciate it if you would be prepared to check this code carefully.
If these changes are ok, then we are probably close to a finalised form for the code.

core/parse-xml.c
atdotde: Could you please make sure dive planner.cpp also uses your new helper functions? 

The other thing is (I said this before and @bstoeger did not really agree so please make of it what you like): I find using fixed precision types like int32_t strange unless you are doing some bit fiddling where you need control over how many bits you really have. Why not use int or our own semantically dimensional types like pressure_t? And yes, there might be good reasons to use unsigned types (like: pressure cannot be negative) but using those has bitten us already in other places (as unsigned int u = 0; int i = -1; if (u < i) printf("WTF!\n") does print since i is cast to unsigned int (and thus is very big) rather than casting u to int.

willemferguson: I am trying to limit the code changes to deal with atmospheric pressure in the UI and it is much more invasive than I expected. Would it be ok if I dealt with that as a different PR? Once this one is done, I will do the changes to the planner immediately. Ok?

The mbar member of pressure_t is int32_t, that was the logic. But the airpressure is now being parsed using the pressure() function that operates on a pressure_t structure.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -130,10 +147,64 @@ void TabDiveInformation::divesChanged(dive_trip *trip, const QVector<dive *> &di
...
 }
+
+void TabDiveInformation::updateTextBox(int i)
+{
+	pressure_t atmpress = { 0 };
+	double altitudeVal;
+	if (current_dive) {
+		switch (ui->atmPressType->currentIndex())
+		case 0:		// If an atm pressure has been specified in mbar:
+			if (i == 1) {	// this is only triggered by on_atmPressVal_editingFinished()
```
bstoeger: Something is very weird here -> this braces should be after the switch!

The fact that this compiles is scary and amazing, but the way C (and by inheritance C++) works!

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -130,10 +147,64 @@ void TabDiveInformation::divesChanged(dive_trip *trip, const QVector<dive *> &di
...
+				altitudeVal = (ui->atmPressVal->text().toFloat());	// get altitude from text box
+				if (prefs.units.length == units::FEET)		// if altitude in feet
+					altitudeVal = feet_to_mm(altitudeVal); // imperial: convert altitude from feet to mm
+				else
+					altitudeVal = altitudeVal * 1000;	// metric: convert altitude from meters to mm
+				atmpress.mbar = altitude_to_pressure((int32_t) altitudeVal); // convert altitude (mm) to pressure (mbar)
+				ui->atmPressVal->setText(ui->atmPressVal->text().sprintf("%d",atmpress.mbar));
+				ui->atmPressType->setCurrentIndex(0);		// reset combobox to mbar
+			}
+			if (i == 0)
```
bstoeger: This could also be written as `else...` or `else if...`

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -130,10 +147,64 @@ void TabDiveInformation::divesChanged(dive_trip *trip, const QVector<dive *> &di
...
+		break;
 	case DiveField::DATETIME:
 		updateWhen();
 		break;
 	default:
 		break;
 	}
 }
+
+void TabDiveInformation::updateTextBox(int i)
```
bstoeger: Why not something more descriptive, such as `bool typeChanged` or similar?

bstoeger:
> Berthold, I would appreciate it if you would be prepared to check this code carefully.
> If these changes are ok, then we are probably close to a finalised form for the code.

I have only a few coding-style comments. If you tested it and it does what it is supposed to do - from my side it's good to go.

core/parse-xml.c
bstoeger: `alloc_dive()` does a `memset()` to 0. I believe that we depend in multiple places on the fact that pointers are zeroed.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -130,10 +147,64 @@ void TabDiveInformation::divesChanged(dive_trip *trip, const QVector<dive *> &di
...
+void TabDiveInformation::on_atmPressType_currentIndexChanged(int index)
+{
+	updateTextBox(0);
+}
+
+void TabDiveInformation::on_atmPressVal_editingFinished()
+{
+	updateTextBox(1);
+}
+
```
bstoeger: Coding style: unnecessary vertical white space...

core/parse-xml.c
atdotde: With "fixed" I meant defined as in "platform independent". With int you get what is natural for the platform (or the compiler more specifically), the "natural" thing whereas specifying the number of bits forces that on the compiler. But this really not worth losing sleep about. Just don't forget to adopt dive planner.cpp to use the helper as well soon.

atdotde (approved):

core/parse-xml.c
bstoeger: > With "fixed" I meant defined as in "platform independent". 

Yes, I got that now, but with your comment 9 days ago I was a bit confused, as I thought you were arguing against integer-arithmetic... I think the proper name for the int*_t types is "exact width" or "fixed width" as opposed to "fixed precision", which for me means a fixed number of decimal places.

> But this really not worth losing sleep about.

True. 😁

willemferguson:
I corrected the curly braces and the if statement. Within the main function I used two #defines to make the logic as clear as possible. I hope this responds to the last issues raised by Berthold.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: Corrected

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: Done

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: Done

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: The purpose of the function is to consider the index of the combobox, store an appropriate number in the text box and save it in the dive structure. A *change* in index is only one of the triggers for this function. The more important trigger is editing of the text box. I cannot really think of a more descriptive name for the function. "typeChanged" would be less accurate.

willemferguson:
I think I have now responded to all the requests.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
bstoeger: I meant the parameter, not the function. :)

willemferguson: I think the #defines that I used make it more clear.

bstoeger:
@dirkhh: Are you OK with merging? There seem to be a few minor points - but from my side they could be done as cleanup-commits.

core/dive.c
```diff
@@ -1328,8 +1329,24 @@ static void fixup_surface_pressure(struct dive *dive)
...
+}
+
+static void fixup_surface_pressure(struct dive *dive)
+{
+	dive->surface_pressure = calculate_surface_pressure(dive);
+}
+
+/* if the surface pressure in the dive data is redundant to the calculated
+ * value (i.e., it was added by running fixup on the dive) return 0,
+ * otherwise return the air temperature given in the dive */
```
dirkhh, line 1343: the comment should say `surface pressure` instead of `air temperature`, correct?

desktop-widgets/tab-widgets/TabDiveInformation.cpp
dirkhh: this one is still there

dirkhh (approved):
Two minuscule things left - I'll fix them myself real quick after I merge this

dirkhh:
cleanup pushed as well

PR #2088, opened 2019-05-10
Simplification of application-state code

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This tries to make the app-state thing a bit more explicit and less dynamic.

The backstory: It's three times that I had to figure our how the widgets are disabled in dive-site-edit mode. It's clever, but very hard to follow: dynamic property lists in arrays. And since we only use a single property (enabled) over-engineered.

I replaced that by some old-fashioned flags and Boolean expressions. Boring, yes.

Moreover, I replaced the QByteArray index by a simple enum. Thus, typos will be caught at runtime.

The use of an `enum class` construct may be questionable to some, as it means that it has to be cast to int to be used as an array index. We could also use something like:

```
struct ApplicationState {
	enum State {
		...
	};
};
```
The disadvantage here would be that the type of the enum is `ApplicationState::State`. I'm a bit undecided on that point.

Overall, the new code is not very pretty, but at least to me it is more palatable. YMMV.

### Changes made:

1) Turn application state index from QByteArray to enum.
2) Use boring flags instead of property lists to enable widgets.
3) Use function instead of macro to set the individual quadrants.

### DIFF
--- a/core/applicationstate.cpp
+++ b/core/applicationstate.cpp
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "applicationstate.h"
+
+static ApplicationState appState = (ApplicationState)-1; // Set to an invalid value
+
+ApplicationState getAppState()
+{
+	return appState;
+}
+
+void setAppState(ApplicationState state)
+{
+	appState = state;
+}
+
--- a/core/applicationstate.h
+++ b/core/applicationstate.h
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef APPLICATIONSTATE_H
+#define APPLICATIONSTATE_H
+
+// By using an enum class, the enum entries don't polute the global namespace.
+// Moreover, they are strongly typed. This means that they are not auto-converted
+// to integer types if e.g. used as array-indices.
+enum class ApplicationState {
+	Default,
+	EditDive,
+	PlanDive,
+	EditPlannedDive,
+	EditDiveSite,
+	FilterDive,
+	Count
+};
+
+ApplicationState getAppState();
+void setAppState(ApplicationState state);
+
+#endif
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -12,6 +12,7 @@
 #include "divecomputer.h"
 #include "time.h"
 #include "gettextfromc.h"
+#include "applicationstate.h"
 #include "metadata.h"
 #include <sys/time.h>
 #include "exif.h"
@@ -1362,21 +1363,9 @@ extern "C" void parse_display_units(char *line)
 	qDebug() << line;
 }
 
-static QByteArray currentApplicationState;
-
-QByteArray getCurrentAppState()
-{
-	return currentApplicationState;
-}
-
-void setCurrentAppState(const QByteArray &state)
-{
-	currentApplicationState = state;
-}
-
 extern "C" bool in_planner()
 {
-	return currentApplicationState == "PlanDive" || currentApplicationState == "EditPlannedDive";
+	return getAppState() == ApplicationState::PlanDive || getAppState() == ApplicationState::EditPlannedDive;
 }
 
 extern "C" enum deco_mode decoMode()
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -171,26 +171,18 @@ MainWindow::MainWindow() : QMainWindow(),
 
 	diveSiteEdit = new LocationInformationWidget(this);
 
-	std::pair<QByteArray, QVariant> enabled = std::make_pair("enabled", QVariant(true));
-	std::pair<QByteArray, QVariant> disabled = std::make_pair("enabled", QVariant(false));
-	PropertyList enabledList;
-	PropertyList disabledList;
-	enabledList.push_back(enabled);
-	disabledList.push_back(disabled);
-
-	registerApplicationState(ApplicationState::Default, mainTab.get(), profileContainer, diveList, mapWidget );
-	registerApplicationState(ApplicationState::EditDive, mainTab.get(), profileContainer, diveList, mapWidget );
-	registerApplicationState(ApplicationState::PlanDive, divePlannerWidget, profileContainer, divePlannerSettingsWidget, plannerDetails );
-	registerApplicationState(ApplicationState::EditPlannedDive, divePlannerWidget, profileContainer, diveList, mapWidget );
-	registerApplicationState(ApplicationState::EditDiveSite, diveSiteEdit, profileContainer, diveList, mapWidget);
-	registerApplicationState(ApplicationState::FilterDive, mainTab.get(), profileContainer, diveList, &filterWidget2);
-
-	setStateProperties(ApplicationState::Default, enabledList, enabledList, enabledList, enabledList);
-	setStateProperties(ApplicationState::EditDive, enabledList, enabledList, enabledList, enabledList);
-	setStateProperties(ApplicationState::PlanDive, enabledList, enabledList, enabledList, enabledList);
-	setStateProperties(ApplicationState::EditPlannedDive, enabledList, enabledList, enabledList, enabledList);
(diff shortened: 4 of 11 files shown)


core/applicationstate.cpp
```diff
@@ -0,0 +1,14 @@
+#include "applicationstate.h"
```
dirkhh, line 2: we like having an SPDX comment on the top of all files

core/applicationstate.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+static ApplicationState appState = (ApplicationState)-1; // Set to an invalid value
```
dirkhh, line 4: space around the `-` ?

core/applicationstate.h
```diff
@@ -0,0 +1,21 @@
...
+// to integer types if e.g. used as array-indices.
+enum class ApplicationState {
+	Default,
+	EditDive,
+	PlanDive,
+	EditPlannedDive,
+	EditDiveSite,
+	FilterDive,
+	Count
+};
```
dirkhh, line 16: I don't think we are consistent with this - don't we use all caps for things like this in other cases? For example in `enum cylinderuse`...

Looking further down I realize why you are using mixed case. This makes the changes smaller (or at least easier). I still think all caps would be better.

core/qthelper.cpp
```diff
@@ -1362,21 +1363,9 @@ extern "C" void parse_display_units(char *line)
...
-
-void setCurrentAppState(const QByteArray &state)
-{
-	currentApplicationState = state;
-}
-
 extern "C" bool in_planner()
 {
-	return currentApplicationState == "PlanDive" || currentApplicationState == "EditPlannedDive";
+	return getAppState() == ApplicationState::PlanDive || getAppState() == ApplicationState::EditPlannedDive;
```
dirkhh, line 1368: I like this much better.

desktop-widgets/mainwindow.cpp
```diff
@@ -369,17 +369,17 @@ void MainWindow::setupSocialNetworkMenu()
 {
 }
 
-void MainWindow::setStateProperties(const QByteArray& state, const PropertyList& tl, const PropertyList& tr, const PropertyList& bl, const PropertyList& br)
+void MainWindow::setStateProperties(ApplicationState state, const PropertyList& tl, const PropertyList& tr, const PropertyList& bl, const PropertyList& br)
 {
-	stateProperties[state] = PropertiesForQuadrant(tl, tr, bl, br);
+	stateProperties[(int)state] = PropertiesForQuadrant(tl, tr, bl, br);
```
dirkhh: agreed, that's a bit awkward, but... meh. Overall it's a win.

desktop-widgets/mainwindow.cpp
```diff
@@ -1835,9 +1822,15 @@ void MainWindow::showFilterIfEnabled()
 		ui.bottomSplitter->setSizes({ EXPANDED, COLLAPSED });
 	}
 }
-void MainWindow::registerApplicationState(ApplicationState state, QWidget *topLeft, QWidget *topRight, QWidget *bottomLeft, QWidget *bottomRight)
+void MainWindow::registerApplicationState(ApplicationState state, QWidget *topLeft, QWidget *topRight, QWidget *bottomLeft, QWidget *bottomRight,
+					  int topLeftFlags, int topRightFlags, int bottomLeftFlags, int bottomRightFlags)
```
dirkhh: I know this is a weird question... but what makes more sense:
`(state, widget1, widget2, widget3, widget4, flag1, flag2, flag3, flag4)`
or
`(state, widget1, flag1, widget2, flag2, widget3, flag3, widget4, flag4)`
?
For some reason I first thought the second one would be better. But after writing this, I'm not sure anymore. So consider this food for thought...

desktop-widgets/mainwindow.cpp
```diff
@@ -1835,9 +1822,15 @@ void MainWindow::showFilterIfEnabled()
...
 {
-	applicationState[(int)state] = WidgetForQuadrant(topLeft, topRight, bottomLeft, bottomRight);
+	applicationState[(int)state] = {
+		{ topLeft, topLeftFlags},
+		{ topRight, topRightFlags},
+		{ bottomLeft, bottomLeftFlags},
+		{ bottomRight, bottomRightFlags}
```
dirkhh: And this is why I thought the first order was better... because we use them in these pairs...

dirkhh (changes requested):
Mostly questions, but at least one actual change request (the missing SPDX header)

core/applicationstate.cpp
bstoeger: Done.

core/applicationstate.cpp
bstoeger: That's a cast and a unary '-'. We usually don't add space for either of those. This could also be written as:
```
static ApplicationState appState = ApplicationState(-1); // Set to an invalid value
```
Perhaps that's more clear?

core/applicationstate.h
bstoeger: The reason I didn't got wall all-caps is that normal enums work more like macros - the values are valid in the outer scope. I was under the impression that that's why typically all-caps are used: to avoid name conflicts. Here it's more like member-variables.

But I can change to all-caps, no problem. Up to you.

desktop-widgets/mainwindow.cpp
bstoeger: First I was also going with `state, widget` pairs. But then I thought these lines:
```
        registerApplicationState(ApplicationState::Default, mainTab.get(), profileContainer, diveList, mapWidget,
                                                            FLAG_NONE, FLAG_NONE, FLAG_NONE, FLAG_NONE);
```
were more in line with the old code. I'm perfectly fine with changing it.

bstoeger: Well, I just realized that when going the (widget, flag) way, we can arrange the quadrants in the source code like they are on the screen. Pure genius - let's do this. 😀

core/applicationstate.cpp
dirkhh: :faceplant:
Yeah. Right.
Not having the space is the correct thing.

core/applicationstate.h
dirkhh: That makes sense. I'm fine with it.

desktop-widgets/mainwindow.cpp
dirkhh: 💯

desktop-widgets/mainwindow.cpp
```diff
@@ -171,26 +171,18 @@ MainWindow::MainWindow() : QMainWindow(),
...
+	registerApplicationState(ApplicationState::EditDive, { { mainTab.get(), FLAG_NONE }, { profileContainer, FLAG_NONE },
+							       { diveList, FLAG_NONE },       { mapWidget, FLAG_NONE } });
+	registerApplicationState(ApplicationState::PlanDive, { { divePlannerWidget, FLAG_NONE },         { profileContainer, FLAG_NONE },
+							       { divePlannerSettingsWidget, FLAG_NONE }, { plannerDetails, FLAG_NONE } });
+	registerApplicationState(ApplicationState::EditPlannedDive, { { divePlannerWidget, FLAG_NONE }, { profileContainer, FLAG_NONE },
+								      { diveList, FLAG_NONE },          { mapWidget, FLAG_NONE } });
+	registerApplicationState(ApplicationState::EditDiveSite, { { diveSiteEdit, FLAG_NONE }, { profileContainer, FLAG_DISABLED },
+								   { diveList, FLAG_DISABLED }, { mapWidget, FLAG_NONE } });
+	registerApplicationState(ApplicationState::FilterDive, { { mainTab.get(), FLAG_NONE }, { profileContainer, FLAG_NONE },
+								 { diveList, FLAG_NONE },      { &filterWidget2, FLAG_NONE } });
```
dirkhh, line 185: That is awesome.

dirkhh (approved):

PR #2090, opened 2019-05-11
Core: consider dive-number on sorting

### LINKED ISSUE #2086: Divelist sorting is wrong when date/time info is the same
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I just imported a lot of dives from a CSV file. The dives have Dive# and date, but no time information, so the import uses the current time as the dive time. This means that I end up with several dives with the same date/time info but different dive#.
Even though I select the Dive# as sort option, the list is still sorted by the date/time info:
image
 
The only way to get the sorting correct is to change the time info on the dive.

### Operating system:

Running Windows 10 1803 English.

### Subsurface version:

4.8.6

### Steps to reproduce:

Import a list of dives with several dives on the same date and no time info.

bstoeger:
A user reports a problem when dives have the same time but different
numbers. The dives appear sorted randomly (effectively they are sorted
by an internal unique-id).

Try to sort by number for dives at the same date in this case.

Fixes #2086

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Sorts dives by number if on same date, as requested by @RasmusAaen.

### Related issues:

Fixes #2086

### Release note:

Done.

### Documentation change:

No.
### Mentions:

@RasmusAaen

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -780,7 +780,7 @@ struct dive *last_selected_dive()
  * After editing a key used in this sort-function, the order of
  * the dives must be re-astablished.
  * Currently, this does a lexicographic sort on the
- * (start-time, trip-time, id) tuple.
+ * (start-time, trip-time, number, id) tuple.
  * trip-time is defined such that dives that do not belong to
  * a trip are sorted *after* dives that do. Thus, in the default
  * chronologically-descending sort order, they are shown *before*.
@@ -805,6 +805,10 @@ static int comp_dives(const struct dive *a, const struct dive *b)
 		if (trip_date(a->divetrip) > trip_date(b->divetrip))
 			return 1;
 	}
+	if (a->number < b->number)
+		return -1;
+	if (a->number > b->number)
+		return 1;
 	if (a->id < b->id)
 		return -1;
 	if (a->id > b->id)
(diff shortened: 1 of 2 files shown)


core/divelist.c
```diff
@@ -805,6 +805,10 @@ static int comp_dives(const struct dive *a, const struct dive *b)
 		if (trip_date(a->divetrip) > trip_date(b->divetrip))
 			return 1;
 	}
+	if (a->number < b->number)
+		return -1;
+	if (a->number > b->number)
+		return 1;
```
dirkhh-vmware, line 811: Honest question... should that also work if one of them has number 0 (i.e., isn't numbered)?

dirkhh-vmware (commented):
Neither request for change nor approval - I don't know what the right answer is...

dirkhh:
Hehe. Oops. Submitting reviews with my work ID that I rarely ever use...

core/divelist.c
bstoeger: Dives without numbers are sorted before dives with numbers (that means the other way around by default, since we default to descending).

We could reverse that (no numbers after numbers) but I don't think that's a use-case that we should spend too much thought on. Even the whole same time, but different number case appears a bit obscure to me.

dirkhh: The same time case was actually a failure to import CVS dive times correctly... but this does seem the right thing to do... and I agree with your analysis.

dirkhh (approved):

PR #2091, opened 2019-05-11
[RFC] Cleanup: remove UTF8 macros

bstoeger:
At some places we use UTF8 string literals. Therefore, we effectively
only support UTF8 build systems. We might just as well remove all
the other UTF_* macros and use direct string literals.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In some places we depend on UTF-8 string literals. Therefore, we might just remove the UTF8_* macros altogether, isn't it. I'm not sure how this works with strings that are translated, though...

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -316,10 +316,10 @@ double get_temp_units(unsigned int mk, const char **units)
 
 	if (units_p->temperature == FAHRENHEIT) {
 		deg = mkelvin_to_F(mk);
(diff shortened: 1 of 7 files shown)


dirkhh (approved):
I hadn't realized that we are using UTF8 in the source - but I guess it's 2019, so let's merge this.

PR #2092, opened 2019-05-15
Sort equal dives by divecomputer model

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a small improvement, whereby on splitting out dive computers, the dives were sorted apparently randomly (actually by a unique id).

### Release note:

No - the feature is not in a release.

### Documentation change:

No.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -771,6 +771,29 @@ struct dive *last_selected_dive()
 	return ret;
 }
 
+/* Like strcmp(), but don't crash on null-pointers */
+static int safe_strcmp(const char *s1, const char *s2)
+{
+	return strcmp(s1 ? s1 : "", s2 ? s2 : "");
+}
+
+/* Compare a list of dive computers by model name */
+static int comp_dc(const struct divecomputer *dc1, const struct divecomputer *dc2)
+{
+	int cmp;
+	while (dc1 || dc2) {
+		if (!dc1)
+			return -1;
+		if (!dc2)
+			return 1;
+		if ((cmp = safe_strcmp(dc1->model, dc2->model)) != 0)
+			return cmp;
+		dc1 = dc1->next;
+		dc2 = dc2->next;
+	}
+	return 0;
+}
+
 /* This function defines the sort ordering of dives. The core
  * and the UI models should use the same sort function, which
  * should be stable. This is not crucial at the moment, as the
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
Looks reasonable. I restarted the LGTM analysis that somehow failed

core/divelist.c
```diff
@@ -771,6 +771,29 @@ struct dive *last_selected_dive()
 	return ret;
 }
 
+/* Like strcmp(), but don't crash on null-pointers */
+static int safe_strcmp(const char *s1, const char *s2)
+{
+	return strcmp(s1 ? s1 : "", s2 ? s2 : "");
+}
```
dirkhh, line 778: have you looked at `same_string()` https://github.com/Subsurface-divelog/subsurface/blob/2feedf46fa86c19095a9f302d1d7a44d3416384d/core/subsurface-string.h#L33

dirkhh (changes requested):
Now I have a request for a change :-)

core/divelist.c
bstoeger: > have you looked at same_string()

Yes, but that returns bool (true, false) whereas we want a ternary digit (<0, 0, >0) for sorting..?

dirkhh: eh. then this shouldn't be named `..._strcmp` but `safe_string_less_than` or something like that.

bstoeger: I don't think so. `string_less_than(a,b)` should return a bool that is true if string a is less than string b (i.e. false if a>=b).

The function here, on the other hand, does exactly what the man page of `strcmp` says:
> The  strcmp()  function  compares  the two strings s1 and s2.  It returns an integer less than, equal to, or greater than zero if s1 is found, respectively, to be less than, to match, or be greater than s2.

We could call it `safe_comp_string` or something...?

atdotde: Suggestion: We could use the fact that the empty string is less than any other and not invoke strcmp on those as in 

return s1 ? ( s2 ? strcmp(s1, s2) : 1) : s2 ? -1 : 0;

or even

return s2 ? ( s1 ? strcmp(s1, s2) : -1) : s2  != NULL;

bstoeger: Sure, that's a legit way to do it. But I think it is harder to read and I'm not convinced that the resulting code is better: https://godbolt.org/z/Haz218
This optimizes for the uncommon case (string is null), isn't it?

dirkhh (approved):

PR #2095, opened 2019-05-16
Add support for the new Suunto D5

torvalds:
This is just a trivial update to recognize the BT name, and a
libdivecomputer submodule update to get the updates to libdivecomputer
to recognize the new USB IDs etc.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -56,6 +56,11 @@ static dc_descriptor_t *getDeviceType(QString btName)
 		product = "EON Core";
 	}
 
+	if (btName.startsWith("Suunto D5")) {
+		vendor = "Suunto";
+		product = "D5";
+	}
+
 	if (btName.startsWith("G2")  || btName.startsWith("Aladin") || btName.startsWith("HUD")) {
 		vendor = "Scubapro";
 		if (btName.startsWith("G2")) product = "G2";


torvalds:
Note: I've verified that the USB download works. I haven't gotten the BLE side to pair with my desktop, but I suspect it's the same thing that makes it very awkward with the EON Core and EON Steel too: those dive computers want to pair using a secure host key, and bluez doesn't support that by default (apparently it breaks some mice and keyboards).

That likely means that BLE is most useful on mobile. Big surprise.

torvalds:
I take all the BLE comments back. I just tried again, and it worked. The D5 does *not* need any magic BLE setup, but I think my phone had paired with it because I tried to use the Suunto app.

So I had to do "forget mobile" on the D5, and then pair on the desktop, and it all seemed to work.

torvalds:
I'm not sure why the "Analysis failed" thing is happening. Looking at the logs, it seems to have trouble fetching the libdivecomputer update, and says

  ERROR: Error searching for sub-modules
  > Missing unknown 1194585595f7a26f3ec7f96b5ad6e8f689a70258

but I've pushed out that commit to the upstream libdivecomputer repo.

It looks like it tries to get the libdivecomputer files from somewhere else. This was the first time I tried the libdivecomputer submodule update, and clearly I did something wrong.

dirkhh:
I need to switch LGTM to use the GitHub app integration. No time to look into that right now.
I don't think you need to worry too much about it

/D

On May 16, 2019 2:59:57 PM PDT, Linus Torvalds <notifications@github.com> wrote:
>I'm not sure why the "Analysis failed" thing is happening. Looking at
>the logs, it seems to have trouble fetching the libdivecomputer update,
>and says
>
>  ERROR: Error searching for sub-modules
>  > Missing unknown 1194585595f7a26f3ec7f96b5ad6e8f689a70258
>
>but I've pushed out that commit to the upstream libdivecomputer repo.
>
>It looks like it tries to get the libdivecomputer files from somewhere
>else. This was the first time I tried the libdivecomputer submodule
>update, and clearly I did something wrong.
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/Subsurface-divelog/subsurface/pull/2095#issuecomment-493246650

-- 
From my phone

bstoeger:
Travis is happy, therefore I think we can commit.

rumplezZiI:
有点东西

PR #2099, opened 2019-05-19
Travis: build macOS test against Qt 5.12.3 and switch cloud storage

dirkhh:
The .app.zip should once again run on any Mac (ignoring the security issue of
unsigned binaries).

The download location has changed to free storage on Google Drive (not sure how
well that will work). The Qt binaries in that archive include the jpeg and png
libraries that were missing in the Qt 5.11.1 binaries we used until now.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

I created a new archive with a highly customized Qt 5.12.3 install (including QtWebKit, excluding all debug libraries, excluding anything outside of the `clang_64` directory). The QtWebKit version in that install was modified to use the also included `libjpeg` and `libpng`. As a result the `.app.zip` created as part of a Travis build should once again work on any Mac (10.13 or newer - that's a Qt requirement).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ matrix:
 
     - env: SUBSURFACE_PLATFORM='mac'
       os: osx
-      osx_image: xcode10.1
+      osx_image: xcode9.2
       language: c++
       addons:
         homebrew:
--- a/scripts/ios/before_install.sh
+++ b/scripts/ios/before_install.sh
@@ -31,7 +31,7 @@ pushd ${TRAVIS_BUILD_DIR}/..
 
 echo "Get custom Qt build and unpack it"
 curl --output ./Qt-5.11.1-ios.tar.xz \
-                https://storage.googleapis.com/travis-cache/Qt-5.11.1-ios.tar.xz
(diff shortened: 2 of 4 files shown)


dirkhh:
OK, this is now tested and works. The build host is macOS 10.13 which means the resulting `Subsurface.app` doesn't open  from the Finder on a macOS 10.12 system - but funnily enough it appears to work ok when started directly from Terminal.

dirkhh:
And now this all works without a hitch on macOS 10.12 as well. Final step before merging - move the cloud storage to a cheaper provider to save some money... who would have thought that just the silly downloads of those tar balls would add up that much...

henrik242:
Can't you just host the tar balls as releases on GitHub?

dirkhh:
Interesting idea. I haven't thought about it. They are 70M and 150M, quite hefty.

On May 20, 2019 6:23:56 AM GMT+02:00, henrik242 <notifications@github.com> wrote:
>Can't you just host the tar balls as releases on GitHub?
>
>-- 
>You are receiving this because you modified the open/close state.
>https://github.com/Subsurface-divelog/subsurface/pull/2099#issuecomment-493834332

-- 
From my phone

PR #2100, opened 2019-05-19
Implement undo of dive computer addition / moving

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a tiny step in undoization: Make dive computer moving / deletion undoable. Code is primitive, but should be quite robust: Two copies of the dives are kept. This is distinctly easier than manipulating the dives and moving them in the list if necessary. Let's see how it fares.

Also contains minor cleanups.

@dirkhh: I'll be away for 4 weeks (writing this from a TGV), but will be reading emails. Trying to implement the basics of your suggestion concerning multi-edit is the next point I'll attempt.

### Changes made:

1) Make dive computer moving undoable
2) Make dive computer deletion undoable
3) Minor cleanups

### Release note:

Done.
### Documentation change:

Probably not needed...

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -637,7 +637,7 @@ static void copy_dive_onedc(const struct dive *s, const struct divecomputer *sdc
  * this is specifically so we can create a dive in the displayed_dive and then
  * add it to the divelist.
  * Note the difference to copy_dive() / clean_dive() */
-struct dive *clone_dive(struct dive *s)
+struct dive *move_dive(struct dive *s)
 {
 	struct dive *dive = alloc_dive();
 	*dive = *s;			   // so all the pointers in dive point to the things s pointed to
--- a/core/dive.c
+++ b/core/dive.c
@@ -4189,11 +4189,10 @@ struct dive *make_first_dc(const struct dive *d, int dc_number)
 	return res;
 }
 
-/* always acts on the current dive */
-unsigned int count_divecomputers(void)
+int count_divecomputers(const struct dive *d)
 {
 	int ret = 1;
-	struct divecomputer *dc = current_dive->dc.next;
+	struct divecomputer *dc = d->dc.next;
 	while (dc) {
 		ret++;
 		dc = dc->next;
@@ -4229,7 +4228,7 @@ static void delete_divecomputer(struct dive *d, int num)
 
 	/* If this is the currently displayed dive, we might have to adjust
 	 * the currently displayed dive computer. */
-	if (d == current_dive && dc_number >= count_divecomputers())
+	if (d == current_dive && dc_number >= count_divecomputers(d))
 		dc_number--;
 	invalidate_dive_cache(d);
 }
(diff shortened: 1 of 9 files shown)


dirkhh:
I'm still trying to figure out what else we need before we can release 4.9 @bstoeger 
I finally found some time the last few days and am frantically trying to clean up build infrastructure issues (fixed the Windows container build, just about to fix the Mac build, got a new iOS build on the way, Android will be next)... but I'm worried about what we are missing before we can reasonably get ready for 4.9 (and there wasn't much discussion on the mailing list).

Oh well. Let's review this one :-)

desktop-widgets/command_divelist.cpp
```diff
@@ -802,6 +803,57 @@ SplitDiveComputer::SplitDiveComputer(dive *d, int dc_num) : SplitDivesBase(d, sp
...
 
+MoveDiveComputerToFront::MoveDiveComputerToFront(dive *d, int dc_num)
+{
+	setText(tr("move dive computer to front"));
+
+	diveToRemove.dives.push_back(d);
+
+	dive *new_dive = make_first_dc(d, dc_num);
+	if (!new_dive)
+		return;
```
dirkhh, line 809: wouldn't this deserve some error? Because you've now deleted the existing dive (which isn't really the same as moving dive computers within the dive...)

core/dive.c
```diff
@@ -4234,10 +4234,22 @@ static void delete_divecomputer(struct dive *d, int num)
 	invalidate_dive_cache(d);
 }
 
-/* always acts on the current dive */
-void delete_current_divecomputer(void)
+/* Clone a dive and delete goven dive computer */
```
dirkhh, line 4236: the commit message has a cut and paste error: 
https://github.com/Subsurface-divelog/subsurface/pull/2100/commits/872539e6974244c064b3d62e4561e0e6cba3995a says `Undo: make "delete dive computer front" undoable` and that `front` in there seems wrong.

dirkhh (changes requested):
I'm not sure about the error message since the code is used two different ways... but reading the flow it seems that if you want to move a DC to the front and that DC isn't found, we end up deleting the dive and not warn about it. But maybe I'm misreading that.

I would ask you to check the commit message, though.

desktop-widgets/command_divelist.cpp
bstoeger: Yes, this is unfortunate, but harmless:

Note that the old dive is added to a dive-to-remove list. But if we don't add an element to the dive-to-add list, the `workToBeDone()` function will return false and the command will never be executed.

This is solved in a later commit (the if is moved to the front), but I will also fix it here.

core/dive.c
bstoeger: Fixed.

desktop-widgets/command_divelist.cpp
dirkhh: thx

dirkhh (approved):

PR #2106, opened 2019-05-25
[RFC] First stab at multi-edit warning

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a first attempt at showing a warning when multiple dives are edited.

It uses a KMessageWidget. Sadly, the animated show/hide can't be used as it seems to be buggy: an animatedHide()/animatedShow() pair may actually *hide* the widget! Therefore the UI experience has to be improved. This is mostly meant to sketch how a solution could look like. Someone feels like improving the UI?

During testing I found a fundamental problem, which is in my opinion a release-blocker:

1) Go to a dive where the water-temperature was rounded (e.g. multiple dive computers or °F-to-°C conversion).
2) Edit air-temperature, press tab -> the air-temperature is edited focus switches to water-temperature
3) Try to undo the air-temperature editin via the application menu
3a) The water-temperature field loses focus
3b) We get an editing-finished signal
3c) Since only one decimal place is shown this gives a differently rounded value than the recorded one and is interpreted as a real edit
3d) An undo command is created
3e) By chosing undo, the invisible edit is undone!
4) User confused since nothing happened -> goto 3). Endless loop!

I was thinking about always rounding to a single decimal place in the undo-command. But that would mean that the command is aware of F vs. C setting. So currently I'm planning to create an input field that only sends editing-finished signals if the value changed compared to the last setting of the value! Sounds like a crazy proposition, but probably the easiest way out...?

PS: Using my mobile as hot-spot, therefore very brief.

### Documentation change:

Perhaps needed?
### Mentions:

@dirkhh

### DIFF
--- a/core/subsurface-qt/DiveListNotifier.h
+++ b/core/subsurface-qt/DiveListNotifier.h
@@ -81,6 +81,12 @@ signals:
 	void diveSiteDiveCountChanged(dive_site *ds);
 	void diveSiteChanged(dive_site *ds, int field); // field according to LocationInformationModel
 	void diveSiteDivesChanged(dive_site *ds); // The dives associated with that site changed
+
+	// This signal is emited every time a command is executed.
+	// This is used to hide an old multi-dives-edited warning message.
+	// This is necessary, so that the user can't click on the "undo" button and undo
+	// an unrelated command.
+	void commandExecuted();
 public:
 	// Desktop uses the QTreeView class to present the list of dives. The layout
 	// of this class gives us a very fundamental problem, as we can not easily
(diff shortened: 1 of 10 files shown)


atdotde:
Can't you check if the current value in the UI is the same as the stored value when rounded, i.e. by what we would fill the UI from the stored value?

dirkhh (approved):
Not beautiful, but does the job. I'll merge it so we can start working towards 4.9

PR #2111, opened 2019-05-28
Check if dive_site exists before accessing its name

### LINKED ISSUE #2109: Mac: crash when copying dive site of dive created with pasted dive site
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I was entering four dives manually, and had begun creating them locally (without saving, though that doesn't seem to affect the issue). The dives had an almost identical profile, so I wanted to use copy and paste to save time. After the first dive, I copied and pasted the details (including dive site) to create the second, editing only the fields that had changed. Before creating the third dive, I tried to copy the details from the second, including the dive site. When I clicked 'OK' in the Component selection dialog, the app crashed. As a result, all unsaved dives (in this case, all of them) were lost.

### Operating system:

macOS 10.14.5 English

MacBook Pro (Retina, 15-inch, Mid 2014)

### Subsurface version:

Mac 4.8.6

Official release

### Steps to reproduce:

1) Select existing dive from dive list
2) Cmd C to open copy component selection menu
3) Select dive site and click OK
4) Cmd + to add dive
5) Cmd V to paste dive site (add other details as desired)
6) Click Apply changes
7) Select newly created dive from dive list
8) Cmd C to open copy component selection menu
9) Select dive site (should already be selected) and click OK

### Current behavior:

Immediate crash

### Expected behavior:

Copying the dive site of the recently created dive

atdotde:
... in copy dive to clipboad.

Fixes #2109

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Additional information:

It would be great to have a test for such a UI function like Copy. I have no idea how that would be implemented.

### DIFF
--- a/desktop-widgets/simplewidgets.cpp
+++ b/desktop-widgets/simplewidgets.cpp
@@ -496,7 +496,7 @@ void DiveComponentSelection::buttonClicked(QAbstractButton *button)
 		QTextStream text;
 		QString cliptext;
 		text.setString(&cliptext);
-		if (what->divesite)
+		if (what->divesite && displayed_dive.dive_site)
 			text << tr("Dive site: ") << displayed_dive.dive_site->name << "\n";
 		if (what->divemaster)
 			text << tr("Dive master: ") << displayed_dive.divemaster << "\n";


bstoeger:
This makes sense - though I don't understand how the reporter got a null-divesite. There seems to be an additional problem?

Also I wonder if current_dive should be used here instead of displayed_dive.

In any case - commit.

atdotde:
In my current set-up, debugger usage is broken. So I had to investigate this by more traditional means. I knew in which function it was crashing and my first suspicion was that it was a displayed_dive vs current_dive issue after @bstoeger 's recent changes. But that did not fix it and the crash only disappeared after checking the dive_site to be non-null. So I went back to master and changed only that. But locally, I still have a branch using current_dive. Once more, I am not really sure what is the correct thing to use given that the user doing copy would do that with respect to what is currently shown, wouldn't she? If you want, I can turn that branch into a PR as well.

bstoeger:
Yes, I think the reason that I didn't convert it myself is that the cylinders and weight systems might be currently edited by the user. Better wait until the equipment tab is converted to the undo system.

PR #2114, opened 2019-05-30
Cleanup: Move tag functions into separate translation unit

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A few random cleanups - see commit messages. Biggest change: move tag functions in their own translation unit to make dive.h a tiny bit smaller.

### Release note:

No - nothing user visible.


core/dive.c
```diff
@@ -3429,18 +3429,6 @@ struct tag_entry *taglist_added(struct tag_entry *original_list, struct tag_entr
 	return added_list;
 }
 
-void dump_taglist(const char *intro, struct tag_entry *tl)
```
atdotde, line 3432: I am not sure if this will be needed once somebody wants to debug something around tags.

bstoeger: If somebody wants to debug, they could simply temporarily add two lines of the type
```
printf("Tags of dive %d\n", d->number);
for(struct tag_entry *tag = d->tag_list; tag; tag = tag->next) printf("%s (%s)\n", tag->tag->name, tag->tag->source);
```
at strategic positions, no?

I'm not convinced we want to drag around that function for the few cases when it might be needed. At least it should be ifdef-ed out.

bstoeger:
A bit more header splitting - certainly very far from perfect, but it can't hurt to start separating all that.

bstoeger:
@dirkhh: I need your help here. I'm trying to debug a crash condition which is trivially reproduced, but sadly happens deep in Qt code. It's also in master and related to dive selection / the DiveTripModel. Somehow the selection model gets messed up by our (well, my) use of beginMoveRows/endMoveRows. Problem is that the crashes are deep within Qt. Sometimes there's *only* Qt code in the (40+ functions deep) stack backtrace!

To at least know where it exactly crashes I wanted to build Qt from source. Took a few hours (literally). But all I get is:
```
$ ./subsurface
Could not find the Qt platform plugin "xcb" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: linuxfb, minimal, offscreen, vnc, webgl.

Aborted (core dumped)
```

When I configure with "-xcb" it says
```
ERROR: Feature 'xcb' was enabled, but the pre-condition 'features.thread && features.xkbcommon && libs.xcb' failed.
```
I installed all development packages featuring xcb - nothing.

In the config.log I see:
```
Trying source 0 (type pkgConfig) of library xkbcommon ...
+ /usr/bin/pkg-config --exists --silence-errors xkbcommon '>=' 0.5.0
pkg-config did not find package.
  => source produced no result.
Trying source 1 (type inline) of library xkbcommon ...
None of [libxkbcommon.so libxkbcommon.a] found in [] and global paths.
  => source produced no result.
test config.qtbase_gui.libraries.xkbcommon FAILED
```

But it is clearly there:
```
$ find /usr -name *xkbcommon*
/usr/include/xkbcommon
/usr/include/xkbcommon/xkbcommon-compat.h
/usr/include/xkbcommon/xkbcommon-x11.h
/usr/include/xkbcommon/xkbcommon-names.h
/usr/include/xkbcommon/xkbcommon.h
/usr/include/xkbcommon/xkbcommon-keysyms.h
/usr/include/xkbcommon/xkbcommon-compose.h
/usr/lib/x86_64-linux-gnu/pkgconfig/xkbcommon.pc
/usr/lib/x86_64-linux-gnu/pkgconfig/xkbcommon-x11.pc
/usr/lib/x86_64-linux-gnu/libxkbcommon.so.0.0.0
/usr/lib/x86_64-linux-gnu/libxkbcommon.so
/usr/lib/x86_64-linux-gnu/libxkbcommon-x11.so
/usr/lib/x86_64-linux-gnu/libxkbcommon.so.0
/usr/lib/x86_64-linux-gnu/libxkbcommon-x11.a
/usr/lib/x86_64-linux-gnu/libxkbcommon.a
/usr/lib/x86_64-linux-gnu/libxkbcommon-x11.so.0
/usr/lib/x86_64-linux-gnu/libxkbcommon-x11.so.0.0.0
/usr/lib/i386-linux-gnu/libxkbcommon.so.0.0.0
/usr/lib/i386-linux-gnu/libxkbcommon.so.0
/usr/share/doc/libxkbcommon-dev
/usr/share/doc/libxkbcommon0
/usr/share/doc/libxkbcommon-x11-dev
/usr/share/doc/libxkbcommon-x11-0
/usr/share/apport/package-hooks/source_libxkbcommon.py
```

I tried also with -xcb-xlib, but that just produces the same error as before (no backend found).

Very frustrating. :(

Is there anything I can do?

bstoeger:
Never mind - that would be too easy anyway. It should be possible to deduce the problem without debugger.

bstoeger:
I think I understand one of the crash-conditions, but there seem to be more of them:

1) Select first dive in trip and say "create trip above".
2) Select last dive in the old trip
3) Undo

What happens: the undo command deletes the newly generated trip. It removes it from the model with:
```
                        beginRemoveRows(QModelIndex(), idx, idx);
                        items.erase(items.begin() + idx);
                        endRemoveRows();
```
That triple of calls mangles the current selection.
Before, there were 17 selection items (one per column). Afterwards, only column 0 is still valid, the others are invalid (row = -1, column = -1, model = null).
I think this happens because the items are moved(?).

When the old selection is cleared, this crashes somewhere in the QSortFilterProxyModel, which is called by QTreeView::isIndexHidden.

What a disaster. :(

I tried clearing the whole selection before every command and that made crashes significantly more rare. But they still happen.

I'm not sure to what extent we are at fault (I didn't find any invalid memory access with valgring) or whether this whole model/view thing is just madness. Probably I'm simply too stupid to understand it.

bstoeger:
No real progress here since I don't know whether this is a follow-up error. But at least I can localize the problem to one call:
In `DiveTripModelTree::divesDeleted()` we remove a whole trip with the triple given above.
The `QAbstractItemModel::endRemoveRows()` call frees memory via `AbstractItemModel::rowsRemoved()`.

After this call returns, calling the `QItemSelectionModell::selection()` function accesses said freed memory.
Notably it seems to generate a `QModelIndex` with an `internalPointer()` to a `QSortFilterProxyModel` mapping that points to the above freed memory. The access is then in `
QSortFilterProxyModel::parent(QModelIndex &) const`

I guess we're using the model-interface incorrectly, but how so? I really don't feel like digging deeper, especially without a custom-built qt version.

In general, funny things happen to the selection on add/remove rows...

@thiagomacieira ?

thiagomacieira:
I don't know almost anything about ItemViews.

dirkhh:
Ugh, slowly coming up for air.
Building Qt from source is a bit of a trip, but the issues above don't sound familiar - @thiagomacieira any chance this rings a bell for you?
The trip selection (and our augmentation of the way Qt tracks selection) has always been a pain. In the past I had given up and tracked selection separately in our data structures, but IIRC you removed some of that. It's code that I haven't looked at in a while.
First thing to try would be to save the selection (i.e. copy it into our data structures), then unselect everything and try to re-create the selection based on the data you saved above.
Sometimes the unselect all crashes and last time we had a problem there that pointed us in the right direction (IIRC it was an object that we freed too early that was still referenced). Not sure if this will help you here.

thiagomacieira:
The problem building from source is that your xkbcommon is probably too old for Qt (minimum 0.5.0). We've recently stopped shipping our own copy bundled inside the Qt sources, which is why you're seeing this when you previously didn't.

I don't know almost anything about the ItemViews error. But if you isolate which of the cleanup commits caused the problem, I can take a look and see if there's anything I know.

bstoeger:
> The problem building from source is that your xkbcommon is probably too old for Qt (minimum 0.5.0). We've recently stopped shipping our own copy bundled inside the Qt sources, which is why you're seeing this when you previously didn't.

apt reports 0.8.2. It appears that somehow the package is broken since pkg-config doesn't know anything of it.

> I don't know almost anything about the ItemViews error. But if you isolate which of the cleanup commits caused the problem, I can take a look and see if there's anything I know.

The problem is also in master. I'm just discussing it here because in this branch it is very easily triggered. For personal reasons I won't be able to work on this for at least a week.

dirkhh (approved):
So many cleanups. Wow.
Nothing stands out as problematic, so instead of carrying this as PR, let's merge and continue to hunt for the crashes.

dirkhh:
I can reproduce the crash in master, so it shouldn't be discussed here (and shouldn't hold up merging this branch).
The one thing that I notice is that when I change the selection back to what it was right after the trip was created, undoing the trip doesn't crash. So apparently there is some side effect of having a different selection when undoing the command than was the result of the command. I was trying to track down what that might be, but the code has become so elegant (i.e., so unreadable to me) that I struggle with understanding what triggers what and how things get called. But I think that's one possible string to pull on.

bstoeger:
Thanks for looking into this. I think a have a rather good grasp of what happens.

The fundamental problem is the duplication of data in the core and the Qt models. This will have to go for at least two reasons, namely performance and maintainability. The way things work right now is insane.

What happens in this concrete case on undo:
1) first we modify the core: move the dive from the split-out trip into the old trip
2) then, we update all the model things.

The problem: the filter-sort-model uses the core structures to decide whether a trip is hidden. Since the trip we removed the dive from has no more dives (in the core) it is suddenly considered as hidden. This completely messes with the mapping between source and filtered model. I don't know if there is also a Qt-bug involved, but what we're doing is definitely wrong.

You can see this by making the MultiFilterSortModel::filterAcceptsRow function simply return true.

dirkhh:
I hear what you are saying.
A bit of historic context. We were unable to create the semantics for selection that we wanted in Qt. So we ended up with the duplication so that we could "fix-up" things to match our desired outcome. Obviously that's a bad thing if it can lead to crashes - but I worry that moving everything to only live in Qt models will allow us to maintain the functionality that we have today.

bstoeger:
Actually, I'm talking about removing the data in the qt-models, not in the core. Don't worry. :)

I'm not a fan of the Qt model/view system (I think it is broken at many levels - for example its way of expressing ranges, accessing data, etc.). But its whole point is connecting the interface to arbitrary data. It should *not* duplicate data. Currently, we're doing things in a fundamentally wrong way.

The whole selection thing is a bit of a red herring - that is not the fundamental problem. The problem is that the trips and dives were at the same time in model and core - in different states. Apparently the standard qt filter model can't handle changes of visibility. I'm not certain if this is a Qt bug or not - but it looks at least like a quality-of-implementation thing to me.

I have a patch that fixes the problem - but the code is appalling. And it has some strange behavior with the selection that I still have to look into.

As for selection, yes I doubt that that should be in the core. What if you want (for what ever reason) two list-views? It just seems illogical for me to save selection with the dives. But let's bother with the pressing issues first!

dirkhh:
> Actually, I'm talking about removing the data in the qt-models, not in the core. Don't worry. :)

Yay. That is a direction I endorse :-)

PR #2115, opened 2019-05-31
Add BLE detection for the new Oceanic Geo 4.0

torvalds:
Note that we don't really have libdivecomputer support for it yet, only
newly added model numbers etc.  But the name detection should make it
easier for people to at least download a memory dump.

In addition to the libdivecomputer model number updates, this also has a
merge of Jef's upstram libdivecomputer changes.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -85,6 +85,11 @@ static dc_descriptor_t *getDeviceType(QString btName)
 		product = "Pro Plus X";
 	}
 
+	if (btName.contains(QRegularExpression("^FS\\d{6}$"))) {
+		vendor = "Oceanic";
+		product = "Geo 4.0";
+	}
+
 	if (!vendor.isEmpty() && !product.isEmpty())
 		return descriptorLookup.value(vendor + product);
 


torvalds:
Note that this is entirely untested, because I don't have the required hardware.

This is mostly preparatory - the actual support will need more libdivecomputer changes. But in the meantime the BLE device should now be correctly detected and named, and hopefully that in turn will make future reports about this dive computer better (the initial reporter tried to download from entirely the wrong device).

bstoeger:
Looks good to me.

I reckon this regex means "string starts with FS, followed by 6 digits, then ends"?

I'm not sure if you are aware, but C++ has raw string literals. Might be handy for more complicated regexs.

torvalds:
Yeah, the BLE naming of Oceanic devices (well, the Plegic devices that can be Oceanic, Aqualung. etc) is the model number ("FS", although libdivecomputer treats it as the hex number 0x4653) followed by six digits of serial number.

I just duplicated the existing matching logic we have for other similar devices (Aqualung i770R and the Oceanic Pro Plus X).

We could maybe have some generic set of regexps for these things (with a mapping from regex to vendor/name), but I'm not sure it's worth it yet.

bstoeger:
Yes, perhaps having precompiled regexps is a good idea in the long run. Since Travis is happy, let's commit.

bstoeger:
Sorry, can't commit:

> Pull requests that have a failing status can’t be merged on a phone.

😐

neolit123:
looks like there is a new CI tool "LGTM analysis: C/C++".
https://lgtm.com/projects/g/Subsurface-divelog/subsurface/alerts/?mode=list

PR #2121, opened 2019-06-03
Add delta-pO2 display in Information Box

willemferguson:
Add display of the difference between pO2 in rebreather loop and the
equivalent OC pO2: this is the oxygen drop over the mouthpiece for
SCR dives. Obviously this is only displayed for SCR dives.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


atdotde:
The data is there, yes, why not display it? I am not a PSCR diver, I cannot judge how useful this is. How is pressures.o2 determined, does this come from a sensor? Or is it calculated using the steady state approximation? I could not quickly determine this. But yes, looks good to me.

atdotde (approved):
If this gives interesting data for Willem's dives, then yes, why not.

bstoeger:
I find the whole proposition of a rebreather very scary - but if it is useful, then so be it.

PR #2123, opened 2019-06-06
Printing: use regexp in preprocessTemplate()

bstoeger:
preprocessTemplate() replaces variables of the kind "dive.weight0"
by "dive.weights.0". Replace the old code by regexps. This not
only makes the code significantly shorter, it also makes it independent
from the name of the dive variable (i.e. "dive").

Moreover, it removes a dependency on MAX_WEIGHTSYSTEMS and MAX_CYLINDERS,
which might help in removing these restrictions.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This replaces the manual variable-replacement code in templatelayout.cpp by regular expressions.
Not only is the resulting code shorter, it is also more flexible in two ways:
1) Can use other dive variables than `dive`
2) Is independent of MAX_WEIGHTSYSTEMS and MAX_CYLINDERS, should we decide to make these dynamic. (BTW: This is the actual reason for the patch.)

But to be honest, I don't see the point in these replacements at all. Why can't the user simply write `weights.5` instead? And wouldn't it be more logical to simply loop over `weights` than to access them individually (and thus fail if MAX_WEIGHTSYSTEMS is increased)? I'd really like to deprecate these substitutions.
### Release note:

No.

### Documentation change:

Should at least replace weight0-weight5 by weight0-weightn.

### Mentions:

@neolit123

### DIFF
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -3,7 +3,6 @@
 #include "core/qthelper.h"
 #include "core/settings/qPrefDiveComputer.h"
 #include <QDebug>
-#include <QRegularExpression>
 #if defined(Q_OS_ANDROID)
 #include "core/subsurface-string.h"
 #endif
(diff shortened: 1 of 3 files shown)


neolit123:
LGTM thanks.

neolit123 (approved):

PR #2127, opened 2019-06-21
use QStringRef instead of QString instances as intermediate values

DerDakon:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This avoids allocating temporary strings when the original stays constant so the original memory can be reused. There are also some other minor cleanups along the way, e.g. avoiding to search the string twice for a character position.

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -25,11 +25,11 @@ static dc_descriptor_t *getDeviceType(QString btName)
 	if (btName.startsWith("OSTC")) {
 		vendor = "Heinrichs Weikamp";
 		if (btName.at(4) == '3') product = "OSTC Plus";
-		else if (btName.mid(4,2) == "s#") product = "OSTC Sport";
-		else if (btName.mid(4,2) == "s ") product = "OSTC Sport";
-		else if (btName.mid(4,2) == "4-") product = "OSTC 4";
-		else if (btName.mid(4,2) == "2-") product = "OSTC 2N";
-		else if (btName.mid(4,2) == "+ ") product = "OSTC 2";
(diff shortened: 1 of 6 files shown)


atdotde:
Let me raise a conservative point: Even though this might be technically correct, I am not sure this is solving any actual problem. Let me quote from the Qt documentation for QStringRef: 

 

> For most users, there are no semantic benefits to using QStringRef instead of QString since QStringRef requires attention to be paid to memory management issues, potentially making code more complex to write and maintain.
> 
> Warning: A QStringRef is only valid as long as the referenced string exists. If the original string is deleted, the string reference points to an invalid memory location.
> 
> We suggest that you only use this class in stable code where profiling has clearly identified that performance improvements can be made by replacing standard string operations with the optimized substring handling provided by this class.

This sounds to me you are introducing a complication in the source code to gain a few cycles in places that are not performance critical at all. 

Convince me otherwise!

DerDakon:
That's why I used them only for temporaries, in most cases never storing them to a variable at all. I separated out the QStringRef changes from the other cleanups now, so you have 2 distinct commits to look at.

dirkhh (changes requested):
I appreciate the thought here - but if I look at the code overall, I cannot claim that I find the new code  introduced in the first commit is easier to read, easier to maintain, or sufficiently better. We can certainly argue about each of the changes but overall - I'm not convinced.

The second commit is simply a NAK from me. This code is not time critical and this is such an odd thing to worry about - and it definitely risks adding fragility, as shown with the added `.toString()` in line 382 - that's a sign of the code becoming harder to maintain. So no. Again, thanks for working on this, but this isn't what needs work.

PR #2130, opened 2019-06-22
Desktop: don't use planner to manually add dive

### LINKED ISSUE #2128: Bugs in manual dive entry profile and dive duration.
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When adding a dive by hand, one should be able to edit the dive profile on the profile panel as described in section 5.1.1 of the user manual. This currently does not happen. In fact it appears there are two bugs:

1) See attached image. The default profile of 15m for 45 minutes shows a crazy profile that is not useful to anyone. The descent rate to 15m should be in the order of 20m/min, NOT 15m/22min. Similarly the default ascent rate should be around 10m/min. The profile should have a square shape, not a triangular shape.

2) The normal red cross-haired cursor and the white points on the vertices of the profile are gone so that the default profile cannot be edited on the profile panel. See section 5.1.1 in user manual.

The dive time value in the Notes tab appears to be involved in these anomalies. If one retypes the 45 minutes, then the dive profile becomes "normal".

### Operating system:
 Ubuntu 18.04
 i7 laptop

### Subsurface version:
 latest master

 compiled amster

### Steps to reproduce:

Start up Subsurface
Select Log -> Add dive

dirkhh:
Instead of calling into the planner, simply create the dive computer
information right there, using the existing helper function we have to
create simple profiles.

Fixes #2128

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As the todo in the comment said - calling into the planner really didn't make any sense there.
Instead we create the initial dive (and profile) right there.

### Related issues:

Fixes #2128 

### Release note:

not needed

### Documentation change:

not needed

### Mentions:

@willemferguson

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -965,18 +965,17 @@ void MainWindow::on_actionAddDive_triggered()
 
 	// create a dive an hour from now with a default depth (15m/45ft) and duration (40 minutes)
 	// as a starting point for the user to edit
-	clear_dive(&displayed_dive);
-	displayed_dive.id = dive_getUniqID();
-	displayed_dive.when = QDateTime::currentMSecsSinceEpoch() / 1000L + gettimezoneoffset() + 3600;
-	displayed_dive.dc.duration.seconds = 40 * 60;
-	displayed_dive.dc.maxdepth.mm = M_OR_FT(15, 45);
-	displayed_dive.dc.meandepth.mm = M_OR_FT(13, 39); // this creates a resonable looking safety stop
-	displayed_dive.dc.model = strdup("manually added dive"); // don't translate! this is stored in the XML file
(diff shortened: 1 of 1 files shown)


bstoeger:
Interesting... For some reason I didn't get this to work without the planner.

Could we now switch to a local dive-object, as in:
```
	dive d = { 0 };
	d.id = dive_getUniqID();
	...
```
?

I'd really like to move the whole `displayed_dive` object down to the planner.

dirkhh:
Excellent idea. Pushed a commit to do just that.

bstoeger (approved):
Didn't test it, but definitely approve.

bstoeger:
Tested - looks good, thanks.

PR #2131, opened 2019-06-23
Dive list: cache shown flag in model (quick-fix for undo-crash)

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a quick-fix for an easily reproduced crash condition. The details are in the first commit-message. The code is appalling, but seems to be the easy way out for now. We really have to get rid of the duplication of code in core and qt-model and the QSortFilterProxyModel. This is just not viable.

This unifies selection behavior for a better UI experience. And pushes the batching of signals by trips to the place where it is needed (the tree view).

### Changes made:

1) Cache the shown flag in the qt-models.
2) Access the shown flag via qt-models, not directly in core.
3) Remove batching of signals by trip.
4) Unify selection in divelist commands.

### Mentions:

@dirkhh: I only slightly tested this - but you said you were able to reproduce the crash, so hopefully you can confirm that this fixes it.

### DIFF
--- a/desktop-widgets/command_divelist.cpp
+++ b/desktop-widgets/command_divelist.cpp
@@ -89,6 +89,37 @@ dive *DiveListBase::addDive(DiveToAdd &d)
 	return res;
 }
 
+// Some signals are sent in batches per trip. To avoid writing the same loop
+// twice, this template takes a vector of trip / dive pairs, sorts it
+// by trip and then calls a function-object with trip and a QVector of dives in that trip.
+// The dives are sorted by the dive_less_than() function defined in the core.
+// Input parameters:
+//	- dives: a vector of trip,dive pairs, which will be sorted and processed in batches by trip.
+//	- action: a function object, taking a trip-pointer and a QVector of dives, which will be called for each batch.
+template<typename Function>
+void processByTrip(std::vector<std::pair<dive_trip *, dive *>> &dives, Function action)
+{
+	// Sort lexicographically by trip then according to the dive_less_than() function.
+	std::sort(dives.begin(), dives.end(),
+		  [](const std::pair<dive_trip *, dive *> &e1, const std::pair<dive_trip *, dive *> &e2)
+		  { return e1.first == e2.first ? dive_less_than(e1.second, e2.second) : e1.first < e2.first; });
+
+	// Then, process the dives in batches by trip
+	size_t i, j; // Begin and end of batch
+	for (i = 0; i < dives.size(); i = j) {
+		dive_trip *trip = dives[i].first;
+		for (j = i + 1; j < dives.size() && dives[j].first == trip; ++j)
+			; // pass
+		// Copy dives into a QVector. Some sort of "range_view" would be ideal, but Qt doesn't work this way.
+		QVector<dive *> divesInTrip(j - i);
+		for (size_t k = i; k < j; ++k)
+			divesInTrip[k - i] = dives[k].second;
+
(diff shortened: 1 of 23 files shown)


qt-models/divetripmodel.cpp
```diff
@@ -683,8 +689,35 @@ dive *DiveTripModelTree::diveOrNull(const QModelIndex &index) const
...
+		// An invalid parent means that we're at the top-level
+		Item &item = items[idx.row()];
+		item.shown = shown;
+		if (item.d_or_t.dive)
+			filter_dive(item.d_or_t.dive, shown);
+	} else {
+		filter_dive(items[parent.row()].dives[idx.row()], shown);
+	}
+
+	return true;
```
dirkhh, line 705: read through this function a couple of times... I'm still not entirely sure what it does, conceptually. I understand the code, but... 
Maybe I need to stare at all this a little longer, but I'll get on an airplane in a few minutes...

desktop-widgets/command_divelist.cpp
```diff
@@ -89,6 +89,37 @@ dive *DiveListBase::addDive(DiveToAdd &d)
...
+// twice, this template takes a vector of trip / dive pairs, sorts it
+// by trip and then calls a function-object with trip and a QVector of dives in that trip.
+// The dives are sorted by the dive_less_than() function defined in the core.
+// Input parameters:
+//	- dives: a vector of trip,dive pairs, which will be sorted and processed in batches by trip.
+//	- action: a function object, taking a trip-pointer and a QVector of dives, which will be called for each batch.
+template<typename Function>
+void processByTrip(std::vector<std::pair<dive_trip *, dive *>> &dives, Function action)
+{
+	// Use std::tie for lexicographical sorting of trip, then start-time
```
dirkhh: is that comment correct? I don't see you use `std::tie` here...

dirkhh (commented):
Started reviewing, got lost a few times, need to get on an airplane.
Not sure if I'll have time during my layover @ LAX before the Shanghai flight or when I'll be able to continue to work on this - so this is mainly a heads up that it could be 20+ hours before I get back to it.

dirkhh:
Oh, first commit message has a word in there that doesn't exist:
 > 3) That lead firt to a rearrangement of the trips.

qt-models/divetripmodel.cpp
bstoeger: Added a few comments and split a line in two. That's hopefully more clear now. Maybe we should cache the flag in the divetrip core-structure, but that can be done later. At the moment it's most important to fix that annoying crash.

desktop-widgets/command_divelist.cpp
bstoeger: No, that comment was outdated. Replaced it by proper comment.

qt-models/divetripmodel.cpp
dirkhh: WiFi on the flight... the crash is fixed. I had two different ways to crash things (and one other situation where undo lost the selection). All of them work reliably now. Well done.

dirkhh (approved):
after testing and more code reading, I think this is good to go.
@bstoeger please merge (you were working on it just recently, I don't want to risk that I merge something that's out of sync)

bstoeger:
I detest the code - but let's try to get a stable 4.9 and then clean this mess.

PR #2132, opened 2019-06-23
[RFC] Merge dives: use higher number

### LINKED ISSUE #2126: Merging dives discards dive number
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

After fixing time on one DC (Garmin messed up local time as they changed to daylight saving time erroneously), the dive number is set to zero instead of keeping the original value. The dive number appears to be taken from the earlier dive, not the one with proper value.

### Subsurface version:

git master

### Steps to reproduce:

Start with dives from different DCs on same log, with different start time (they were not merged automatically on import due to time difference). One dive should have normal dive number, the other is zero (as happens on download on this situation).

- Shift the time on the dive number 0 to correct one.
- Select both dives
- Merge these dives manually on the pop-up menu

### Current behavior:

Dive number on the merged dive is 0 (if the dive with number 0 was earlier).

### Expected behavior:

Dive number is taken from the dive that had non-zero value.

bstoeger:
The merge_dives() function made the output dive use the higher number.

That was overwritten in MergeDives::MergeDives() with the number of
the first dive.

Remove the overwriting and use the number provided by the core.

Fixes #2126

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

On dive merge, don't overwrite the dive-number produced by the core.

This means that the higher number is used. Marking as RFC, since I'm not sure why the old code was there and I'm not sure what the correct policy is. In any case, this moves the decision to the `merge_dives()` function where it should be.

### Release note:

Not sure.

@mturkia @dirkhh

### DIFF
--- a/desktop-widgets/command_divelist.cpp
+++ b/desktop-widgets/command_divelist.cpp
@@ -918,9 +918,6 @@ MergeDives::MergeDives(const QVector <dive *> &dives)
 	d->divetrip = nullptr;
 	d->dive_site = nullptr;
 
-	// The merged dive gets the number of the first dive
-	d->number = dives[0]->number;
-
 	// We will only renumber the remaining dives if the joined dives are consecutive.
 	// Otherwise all bets are off concerning what the user wanted and doing nothing seems
 	// like the best option.


bstoeger:
Yeah, given the renumber logic right afterwards, I see how the dive should probably get the number of the first dive. Let me change that.

PR #2133, opened 2019-06-23
[RFC] Dive merge: don't use unset dive numbers

### LINKED ISSUE #2126: Merging dives discards dive number

bstoeger:
On merging, don't use the number of the first dive if it is 0.
Use the first non-zero number.

Fixes #2126

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Use the first non-0 number when merging dives.

### Release note:

Maybe?

### Documentation change:

No.
### Mentions:

@mturkia @dirkhh

### DIFF
--- a/desktop-widgets/command_divelist.cpp
+++ b/desktop-widgets/command_divelist.cpp
@@ -918,8 +918,13 @@ MergeDives::MergeDives(const QVector <dive *> &dives)
 	d->divetrip = nullptr;
 	d->dive_site = nullptr;
 
-	// The merged dive gets the number of the first dive
-	d->number = dives[0]->number;
+	// The merged dive gets the number of the first dive with a non-zero number
+	for (const dive *dive: dives) {
+		if (dive->number) {
+			d->number = dive->number;
+			break;
+		}
+	}
 
 	// We will only renumber the remaining dives if the joined dives are consecutive.
 	// Otherwise all bets are off concerning what the user wanted and doing nothing seems


dirkhh (approved):
I'd like it better without the `auto`, but it seems fine

bstoeger:
OK - will replace by classical loop tomorrow.

desktop-widgets/command_divelist.cpp
```diff
@@ -918,8 +918,13 @@ MergeDives::MergeDives(const QVector <dive *> &dives)
 	d->divetrip = nullptr;
 	d->dive_site = nullptr;
 
-	// The merged dive gets the number of the first dive
-	d->number = dives[0]->number;
+	// The merged dive gets the number of the first dive with a non-zero number
+	for (const dive *dive: dives) {
+		if (dive->number) {
+			d->number = d->number;
```
dirkhh: that does seem a bit too tautological... I think you meant `d->number = dive->number`...

dirkhh (changes requested):

dirkhh (approved):
that looks right

dirkhh:
So this passes tests, code looks great - why is this an RFC? 
OK, let's comment: I think it makes sense when merging dives to generally pick the earlier number (think merging your last two dives), but only pick a valid number...

bstoeger:
In the `merge_dives()` core-function we determine the *largest* number. Since the code sends different messages, I thought it was prudent to get opinions. Taking the first (non-zero) number appears more logical to me.

mturkia:
I think of two slightly different use cases (most common ones)
- 0 and non-zero dive numbers, and that merge should obviously be non-zero
- two numbered dives: this case would logically warrant for the smaller number. In case we are merging the latest two dives, this will keep the dives numbered without gaps.

What is the first non-zero? How are the dives ordered?

dirkhh:
I'll merge as is as I now twice ran into a case that this code would have fixed.
If we want to fine tune the logic later, that's ok with me.

PR #2134, opened 2019-06-24
Desktop: fix time editing

dirkhh:
We were shifting in the wrong direction. Which caused the field to be marked as
'edited' again, which meant we shifted the wrong way and twice the distance.
This seems to fix the problem for both date and time editing.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

It felt like we just fixed this, right? I'm confused.

### Mentions:

@bstoeger - I wrote and tested this on the flight (yay for ridiculously overpriced trans-Pacific WiFi).
It seems too simple, but it seems to work :-)

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -814,7 +814,7 @@ static void shiftTime(QDateTime &dateTime)
 
 	timestamp_t when = dateTime.toTime_t();
 	if (current_dive && current_dive->when != when) {
-		timestamp_t offset = current_dive->when - when;
+		timestamp_t offset = when - current_dive->when;
 		Command::shiftTime(dives, (int)offset);
 	}
 }


bstoeger (approved):
Makes sense.

It fixes the direction problem. But it only hides the double-signal problem: On undo/redo it will still generate a new UndoCommand. It's only that this command detects that the value is not changed and therefore does not execute. It's not a user-visible problem. Nevertheless, I think we should fix it to avoid weird consequences in corner cases.

dirkhh:
I had code that used `editMode - IGNORE` around the undo commands - but ended up throwing it away. But yeah, I guess that is what we should do to avoid double executions in case of weird rounding issues, for example.

bstoeger:
Yes, these modes always scare me. For the other fields I solved this by subclassing the Qt widgets and implementing different signals for user updates and programmatical updates or using a has_changed flag. Some of the Qt widgets make this distinction, some don't. Some have a has_changed flag, some don't. Quite inconsistent.

PR #2136, opened 2019-06-24
Dive site edit fix

### LINKED ISSUE #2135: Dive flag not updated by redo of GPS edit
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When undoing and then redoing an edit that adds a dive location / GPS coordinates, the flag icon on the map and the globe icon on the dive list are correctly removed by the undo, but not added back by the redo. After selecting a different dive they are shown correctly.
The dive site name is correctly removed and added back in the dive list. Only the little logo indicating GPS coordinates (and the flag on the map) are missing after redo.

### Operating system:

Tested on macOS - but I doubt this is OS specific

### Subsurface version:

with latest master (2e230da361)

### Steps to reproduce:

1) Start from a dive with no GPS / location
2) Add a location, click on the GPS symbol to open dive site management, drag the flag to a new location
3) Click done --> Flag is shown
4) Undo --> Flag disappears
5) Redo --> Flag is still gone
6) Select different dive (nearby, so you see the newly added dive location) --> Flag is shown

Additionally, the little globe symbol in the dive list is also not added back until you change dives (it is removed when you undo adding the dive location). The name of the dive site, however, is correctly added back by the redo.

### Current behavior:

see above

### Expected behavior:

the flag and globe icon in the dive list should be restored in the redo

### Mentions:

@bstoeger I'll look at this - this is a good way for me to figure out more of the inner workings of all that amazing undo code that you've written. I may give up at some point and ask for help :-)

dirkhh:
Make sure that edits to dive sites are correctly reflected both in the dive list and on the map

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

First a silly consistency fix
Second commit makes sure the map gets notified if a dive site is changed by the undo command
Third commit makes sure that the dive list is notified as well - this one seems a bit odd as we now notify the dive list about both the changes and dive sites and the fact that this changes the dives at these dive sites. I wasn't sure if I wanted to add the code that does that here or on the signal receiving end (i.e. in the DIveTripModel). I ended up doing it here since the code seemed much simpler this way.

### Related issues:

Fixes #2135 

### Mentions:

@bstoeger - I'd love to hear your thoughts about sending two signals vs. making the `DiveTripModel` a recipient of the `diveSiteChanged` signal and doing the handling there

### DIFF
--- a/desktop-widgets/command_divesite.cpp
+++ b/desktop-widgets/command_divesite.cpp
@@ -186,6 +186,18 @@ static void swap(char *&c, QString &q)
 	q = s;
 }
 
+// Helper function: collect the dives that are at the given dive site
+static QVector<dive *> getDivesForSite(struct dive_site *ds)
+{
+	QVector<dive *> diveSiteDives;
+
+	for (int i = 0; i < ds->dives.nr; ++i)
+		diveSiteDives.push_back(ds->dives.dives[i]);
+
+	return diveSiteDives;
+}
+
+
 EditDiveSiteName::EditDiveSiteName(dive_site *dsIn, const QString &name) : ds(dsIn),
 	value(name)
 {
@@ -201,6 +213,7 @@ void EditDiveSiteName::redo()
 {
 	swap(ds->name, value);
 	emit diveListNotifier.diveSiteChanged(ds, LocationInformationModel::NAME); // Inform frontend of changed dive site.
+	emit diveListNotifier.divesChanged(getDivesForSite(ds), DiveField::DIVESITE); // dive site name can be shown in the dive list
 }
 
 void EditDiveSiteName::undo()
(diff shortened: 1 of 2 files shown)


desktop-widgets/command_divesite.cpp
```diff
@@ -186,6 +186,18 @@ static void swap(char *&c, QString &q)
 	q = s;
 }
 
+// Helper function: collect the dives that are at the given dive site
+static QVector<dive *> getDivesForSite(struct dive_site *ds)
+{
+	QVector<dive *> diveSiteDives;
```
bstoeger, line 192: Premature optimization, but why not:
```
	diveSiteDives.reserve(ds->dives.nr);
```
?

bstoeger:
Looks sensible. But I will have to look into it in more detail when I find time.

bstoeger:
> I'd love to hear your thoughts about sending two signals vs. making the DiveTripModel a recipient of the diveSiteChanged signal and doing the handling there.

I think I prefer the current version, because it is simpler.

desktop-widgets/command_divesite.cpp
dirkhh: agreed. I should use that.

bstoeger (approved):
Looks like the pragmatic thing to do.

PR #2137, opened 2019-06-25
User manual update (update 1 of 3)

willemferguson:
Text has been added for the undo facility and for the specification
of atmospheric pressure.

Two images added.

Please check where I have added text for undo and tell me if you agree that this is at a good position in the user manual. There are two parts of the manual 1) manual dive entry and 2) dives from dc. Reference to undo needs to be make on both of these parts. I hope this does justice to an enormous mount of code to implement undo.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @bstoeger


bstoeger (approved):
Fine with me, thank you.

Documentation/user-manual.txt
```diff
@@ -19,7 +19,7 @@ image::images/Subsurface4Banner.jpg["Banner",align="center"]
 Linus Torvalds, Miika Turkia, Amit Chaudhuri, Jan Schubert, Salvador Cuñat, Pedro Neves,
 Stefan Fuchs
 
-[blue]#_Version 4.9, February 2019_#
+[blue]#_Version 4.9, June 2019_#
```
dirkhh, line 22: It will surely be July - but we can edit this when we release it :-)

dirkhh (approved):
I think this is a great start for the undo documentation.
Maybe we should mention the keyboard shortcuts? And this also deserves a bit more language about the fact that most operations can be undone. And finally a few words about the oddity that tanks and weights now work differently than the rest of the editing.
I'll merge this as is as it is a distinct step in the right direction.

PR #2140, opened 2019-06-26
Remove MAX_WEIGHTSYSTEMS restriction

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes the restriction to 6 weightsystems. It is a small first step in undoization of the equipment tab.

A side effect is that it temporarily makes the edit of multiple dives more strict: The dive will only be changed if the weightsystems correspond precisely to the shown dive. The old system appeared very ill defined, concerning addition / removal of weightsystems, so I wonder if anybody used that? This will have to be rethought in more detail when implementing undo.

Also note that this is only tested very superficially.

### Release note:

Should do?
### Mentions:

@dirkhh

### DIFF
--- a/core/equipment.h
+++ b/core/equipment.h
@@ -39,11 +39,24 @@ typedef struct
 	const char *description; /* "integrated", "belt", "ankle" */
 } weightsystem_t;
 
+/* Table of weightsystems. Attention: this stores weightsystems,
+ * *not* pointers * to weightsystems. This has two crucial
+ * consequences:
+ * 1) Pointers to weightsystems are not stable. They may be
+ *    invalidated if the table is reallocated.
+ * 2) add_to_weightsystem_table(), etc. takes ownership of the
+ *    weightsystem. Notably of the description string */
+struct weightsystem_table {
+	int nr, allocated;
+	weightsystem_t *weightsystems;
+};
+
 #define MAX_CYLINDERS (20)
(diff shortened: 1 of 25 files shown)


core/equipment.h
```diff
@@ -39,13 +39,27 @@ typedef struct
...
+ *    invalidated if the table is reallocated.
+ * 2) add_to_weightsystem_table(), etc. takes ownership of the
+ *    weightsystem. Notably of the description string */
+struct weightsystem_table {
+	int nr, allocated;
+	weightsystem_t *weightsystems;
+};
+
 #define MAX_CYLINDERS (20)
-#define MAX_WEIGHTSYSTEMS (6)
```
neolit123, line 43: does this mean that user can now add infinite weight systems?

bstoeger: Yes.

dirkhh (commented):
Should this wait until after 4.9?

bstoeger:
> Should this wait until after 4.9?

I'd think so, yes.

dirkhh (approved):
Now that 4.9 is out, let's merge this

dirkhh:
of course this now has a conflict. merged locally

PR #2142, opened 2019-06-27
Cleanup: remove includes from extern "C" block

bstoeger:
In "core/save-html.h", the "core/dive.h" header was included in the
extern "C" block. This is invalid, because "core/dive.h" included
from C++ code contains Qt macros that expand to C++ templates. These
in turn must not have extern "C" linkage, since a plain C-linker
cannot handle such things.

The only reason this worked is that in all cases "core/save-html.h"
was included after "core/dive.h". The include of the latter in the
former had therefore not effect.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes an invalid chain of includes, see commit message.

### Release note:

No.

### DIFF
--- a/core/save-html.h
+++ b/core/save-html.h
@@ -2,12 +2,12 @@
 #ifndef HTML_SAVE_H
 #define HTML_SAVE_H
 
+#include "dive.h"
+#include "membuffer.h"
+
 #ifdef __cplusplus
 extern "C" {
 #endif
-
-#include "dive.h"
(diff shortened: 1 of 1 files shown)


atdotde:
Looks good

dirkhh (approved):
That looks correct. Good catch

Issue #1922, opened 2019-01-02
Preferences > General > Video thumbnails


seanwhe:
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Installed on a system that does not yet have FFMPEG installed, the "Extract video thumb nails" checked by default.

### Operating system:

Linux 4.15.0-43-generic #46-Ubuntu SMP Thu Dec 6 14:45:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

### Subsurface version:

v4.8.4

### Steps to reproduce:

1. Install on a system that does have FFMPEG installed
2. File > Preferences > General

### Current behavior:

Default checked when FFMPEG is not found

### Expected behavior:

Default unchecked when FFMPEG is not found

seanwhe:
I think it should be unchecked and should warn that FFMPEG is found or not.

If the user tries to check it, it should attempt to detect ffmpeg and if fail give some message to say, install it.

IMHO cosmetics count in a big way to usability.

bstoeger:
We probably shouldn't nag users that will never add a video (the vast majority?) with `ffmpeg` related messages. We could show a message the first time they attempt to add videos and `ffmpeg` doesn't work though. Likewise if they actively try to turn on the option.

@dirkhh: What do you think?

dirkhh:
I would be annoyed if Subsurface babbled about ffmpeg when I have no interest to ever attach a video to a dive. So no, this should happen the first time you use it.

bstoeger:
We now warn if the user attempts to set a non-working ffmpeg and give a better error message if the user tries to add videos without having ffmpeg installed.

It's not exactly what the reported wanted, but I think it's OK from an UI point of view?

PR #2159, opened 2019-07-05
Warn on non-executable ffmpeg

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This warns when the user sets a non-existing ffmpeg. Moreover, the error-message when ffmpeg fails to execute is extended to instruct the user to set the proper executable in the preferences. This is in response to #1922, though probably not exactly what the reporter wanted.

### DIFF
--- a/desktop-widgets/preferences/preferences_defaults.cpp
+++ b/desktop-widgets/preferences/preferences_defaults.cpp
@@ -8,6 +8,8 @@
 #include "core/settings/qPrefDiveComputer.h"
 
 #include <QFileDialog>
+#include <QProcess>
+#include <QMessageBox>
 
 PreferencesDefaults::PreferencesDefaults(): AbstractPreferencesWidget(tr("General"), QIcon(":preferences-other-icon"), 0 ), ui(new Ui::PreferencesDefaults())
 {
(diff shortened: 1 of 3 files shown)


dirkhh (commented):
Looks good... but adds new strings.
I'm trying to decide if this is important enough to re-open strings for 4.9 or if we should merge this right after... or if we should merge it now and accept potentially untranslated strings for those errors... thoughts?

desktop-widgets/preferences/preferences_defaults.cpp
```diff
@@ -45,13 +47,37 @@ void PreferencesDefaults::on_localDefaultFile_toggled(bool toggle)
...
+
+	// If the user didn't provide a string they probably didn't intend to run ffmeg,
+	// so let's not give an error message.
+	if (s.isEmpty())
+		return;
+
+	// Try to execute ffmpeg. But wait at most 2 sec for startup and execution
+	// so that the UI doesn't hang unnecessarily.
+	QProcess ffmpeg;
+	ffmpeg.start(s);
```
neolit123, line 62: not sure if someone will bother with exploits, but i'm hoping that we didn't introduce an attack vector by executing a third party binary without the consent of the user. the functionality should be off by default and if the user picks a binary we should only then enable it.

then again, someone might play with adding the value in settings/registry.
ideally, there should be some sort of validation that this is truly ffmpeg, but not exactly easy to do as the user can pick any version.

neolit123 (commented):
not having these strings translated seems ok for planned release.

desktop-widgets/preferences/preferences_defaults.cpp
atdotde: Why not just fstat() and check the executable bit? Is there a Windows version of that? I would guess, actually running it is overkill. We might even try to see if it is an actual ffmpeg by asking it to encode something and checksum the result ;-)

atdotde:
I would say an untranslated string is still better than silently failing.

desktop-widgets/preferences/preferences_defaults.cpp
bstoeger: > not sure if someone will bother with exploits, but i'm hoping that we didn't introduce an attack vector by executing a third party binary without the consent of the user.

This sounds a bit academic. The executable is only run when the user explicitly edited the field. We could check for `changed()` to make sure that the user didn't just tab through. This means that the user plans to upload videos, so the ffmpeg executable will be run anyway. And if an attacker managed to install a backdoored ffmpeg binary, they could also have installed a backdoored subsurface binary in the first place.

> Why not just fstat() and check the executable bit?

As far as I know, Windows is POSIX, so it should work. Feel free to submit a patch. I'm not convinced that this solves anything, but I also don't mind.

bstoeger: > Why not just fstat() and check the executable bit?

On second thought - perhaps not such a good idea: you would have to scan the PATH in a platform independent way. And do it in such a way that it finds the same executable as QProcess. I'm not sure we want to open that can of worms.

neolit123: > The executable is only run when the user explicitly edited the field. 

if that's the case then this is safer.

> As far as I know, Windows is POSIX, so it should work. Feel free to submit a patch. I'm not convinced that this solves anything, but I also don't mind.

Windows has it's own "Windows" standard and is certainly not POSIX.
i don't think we need to do anything else other than only calling the "ffmpeg" binary after the user has enabled (picked it) it in settings.

atdotde: I meant just checking the executable bit instead of trying to run it.

dirkhh: I think that is a valid concern, but also, as @bstoeger points out, an attack vector that doesn't pass the smell test. I wouldn't worry about it.

bstoeger: > Windows has it's own "Windows" standard and is certainly not POSIX.

Ok - good to know.

> I meant just checking the executable bit instead of trying to run it.

Yes, I understand. But first you have to find the same binary that `QProcess` would use, since that is what is used by the thumbnail-generator. I haven't looked at the code of `QProcess`, but I can imagine that it's not completely trivial.

dirkhh:
Given the sum of comments, merging.

PR #2160, opened 2019-07-05
Cleanups concerning the pressure values on the profile

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are a few cleanups in preparation of making the whole thing more dynamic (see #2140). That in turn is preparation for undo of equipment editing actions.

Usage of inline functions makes this somewhat more verbose than the macros before. Sadly, C doesn't know references that can be returned from functions, so a setter function is introduced.

### DIFF
--- a/core/gaspressures.c
+++ b/core/gaspressures.c
@@ -352,7 +352,7 @@ static void debug_print_pressures(struct plot_info *pi)
 	int i;
 	for (i = 0; i < pi->nr; i++) {
 		struct plot_data *entry = pi->entry + i;
-		printf("%5d |%9d | %9d || %9d | %9d |\n", i, SENSOR_PRESSURE(entry), INTERPOLATED_PRESSURE(entry), DILUENT_PRESSURE(entry), INTERPOLATED_DILUENT_PRESSURE(entry));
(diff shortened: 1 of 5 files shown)


dirkhh:
Amy idea why the LGTM build would fail?
Also, I assume this is post-4.9, just like #2140 ?

bstoeger:
I don't understand, no - it warns about an undefined inlined function which would mean that the proper include-file was not included. By why then does Travis pass...?

dirkhh:
Is it possible that LGTM builds with a combination of settings that we don't test on Travis? I need to stare a bit more at that build output


-- 
From my phone

bstoeger:
Seems like making the inline functions `static inline` did the trick. Is this a C thing? I don't think I ever had to do this in C++. We also have a number of non-static inline functions in the C++ part.

It appears backward to me: wouldn't `static inline` mean that if the compiler decides to not inline then the function is repeated for every translation unit? I always thought that inline means that the compiler is free to inline or not but the linker will not complain about multiple definitions. Weird.

bstoeger:
Well, according to a quick Internet search, it appears that `inline` has indeed different meanings in C and C++. Good to know.

dirkhh (approved):
These all look good

PR #2164, opened 2019-07-06
Cleanup: fix memory management of the plot data

bstoeger:
There was a global variable last_pi_entry_new, which stored the
recently allocated plot data. This was freed when new plot data
was generated.

A very scary proposition: You can never have two plot datas at
the same time! But exactly that happens when you export for
example subtitles.

The only reason why this didn't lead to very crazy behavior
is that at least on my Linux machine, the calloc() call would
just return the previously freed memory.

Fix this mess by removing the global variable and freeing the
data in the callers.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes a variable and also fixes a use after free bug: When making a new plot-info the old data was released even if someone still used it! This worked only out of shear luck: the libc would hand out the just free()d memory on the subsequent calloc() call!

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -398,11 +398,10 @@ static void check_setpoint_events(const struct dive *dive, struct divecomputer *
 }
 
 
-struct plot_info calculate_max_limits_new(struct dive *dive, struct divecomputer *given_dc)
+static void calculate_max_limits_new(struct dive *dive, struct divecomputer *given_dc, struct plot_info *pi)
 {
 	struct divecomputer *dc = &(dive->dc);
 	bool seen = false;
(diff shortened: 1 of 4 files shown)


bstoeger:
Or perhaps it worked because the DivePlotDataModel immediately makes a copy of the data and only that copy is accessed?

That's all a bit convoluted...

dirkhh (approved):
This looks reasonable. I don't quite follow the conversation about why things may or may not have worked before, but the changes seem to make sense.

Issue #2170, opened 2019-07-10
Export of dive profile images (PNG) problem when exporting multiple dives


h3llra1ser:
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When you export profile images of multiple dives all profile images have the same Y axis and also wrong scale and ticks of the X axis.
It seems that the grid gets initialized only for the first exported dive (which equals to the profile shown in subsurface) - but the other profiles dont match the axis ticks.
However, the profile grap itself is drawn correct if you ignore X and Y axis in the image :)

### Operating system:

Win7 and Linux - both behave the same

### Subsurface version:

V 4.8.6 official

### Steps to reproduce:

Export more than one dive profile image at once using the menu File / Export / Dive profile

I noticed the same problem when you run TeX export - those profile images are created the same wrong way!

### Current behavior:

X and Y axis correct for the first exported dive, all others have the same axis set which is wrong:

### Expected behavior:

The X and Y axis should be reinitialized for each PNG export in the same loop

atdotde:
I cannot confirm this with the current development version. Could you please try it with a binary from https://github.com/Subsurface-divelog/subsurface/releases/tag/continuous

atdotde:
Checking again, it seems for me, only the y axis is concerned.

h3llra1ser:
Tested again with
Subsurface-4.8.4-679-g9eb860d45d65-x86_64.AppImage

same here....
Yes, as i wrote, its only the axis labels and ticks which are wrong

atdotde:
This is weird. Somehow the profileYAxis is not recomputed. Could not figure out why. Have to run.

bstoeger:
Try setting the "animation speed" in the preferences to 0 (i.e. no animations). I have a feeling that the image is written before the animation has finished.

To me it seems fundamentally flawed that the profile widget of the main window is reused.

bstoeger:
I'm not sure that my analysis is correct. If it is, it is hopefully fixed by #2171, so please check that out.

h3llra1ser:
Tested again with a set of 50 dives.
Disabling animations (set speed to 0) fixes this issue!

bstoeger:
Thanks - hopefully #2171 fixes the issue even without disabling animations.

atdotde:
It does

PR #2171, opened 2019-07-10
Profile: disable animations on export

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is my attempt at fixing #2170. It seems to have worked in my tests, but testing was very light. More testing needed.
### Changes made:

1) Store animation speed in profile object
2) Disable animations on export

### Related issues:

#2170
### Release note:

Done.
### Documentation change:

No.
### Mentions:

@atdotde

### DIFF
--- a/profile-widget/divepixmapitem.cpp
+++ b/profile-widget/divepixmapitem.cpp
@@ -4,6 +4,7 @@
 #include "qt-models/divepicturemodel.h"
 #include "core/pref.h"
 #include "core/qthelper.h"
+#include "core/settings/qPrefDisplay.h"
 #ifndef SUBSURFACE_MOBILE
 #include "desktop-widgets/preferences/preferencesdialog.h"
 #endif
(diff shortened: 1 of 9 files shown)


atdotde:
On the phone, haven’t tried this patch. But I had the same suspicion about animations, so I inserted  1 sec of sleep between dives. This did not help.

bstoeger:
> But I had the same suspicion about animations, so I inserted 1 sec of sleep between dives.

What kind of sleep did you use? I think you have to make sure that Qt's main loop runs. So a simple libc `sleep()` will probably not do it...

profile-widget/divepixmapitem.cpp
```diff
@@ -108,9 +109,9 @@ void DivePictureItem::setFileUrl(const QString &s)
 
 void DivePictureItem::hoverLeaveEvent(QGraphicsSceneHoverEvent*)
 {
-	Animations::scaleTo(this, 0.2);
+	Animations::scaleTo(this, 0.2, qPrefDisplay::animation_speed());
```
atdotde: I think you got the order of arguments wrong (at least I get a compiler warning), should be: this, speed, 0.2

atdotde (changes requested):
I can confirm this fixes the issue.

atdotde:
I used QThread::mssleep(1000) and did not check it runs the event loop. It seems it doesn't.

profile-widget/divepixmapitem.cpp
bstoeger: Ooops. Good catch, thanks! Will fix later.

bstoeger: Should be fixed now.

atdotde (approved):

PR #2175, opened 2019-07-12
Android: fix BT/BLE dowload for arm64 by upgrading Qt, OpenSSL

janmulder:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

This set of commits upgrades the Qt version for Android build to 5.12.4. All the remaining work can be considered fallout.

The Travis build will fail, as this PR does not upgrade Travis and Docker work needed. 

Notice that the generated apks show up in a slightly different place (1 directory deeper, making distinction in debug ans release builds).

### Changes made:
See commits

### Related issues:
No official open issue, but this fixes BT/BLE download on arm64 builds. Which is required by Google soon.

### Additional information:
Travis/Docker work to be done (@dirkhh)

### Release note:
No

### DIFF
--- a/android-mobile/build.gradle
+++ b/android-mobile/build.gradle
@@ -7,10 +7,11 @@ buildscript {
     repositories {
         jcenter()
         maven { url "https://dl.bintray.com/android/android-tools/" }
+	google()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.3'
+        classpath 'com.android.tools.build:gradle:3.2.1'
     }
 }
 
@@ -18,6 +19,7 @@ allprojects {
     repositories {
         jcenter()
         maven { url "https://dl.bintray.com/android/android-tools/" }
+	google()
     }
 }
 
--- a/packaging/android/build.sh
+++ b/packaging/android/build.sh
@@ -242,22 +242,29 @@ if [ "$QUICK" = "" ] ; then
 		mkdir -p openssl-build-"$ARCH"
 		cp -r openssl/* openssl-build-"$ARCH"
 		pushd openssl-build-"$ARCH"
-		perl -pi -e 's/install: all install_docs install_sw/install: install_docs install_sw/g' Makefile.org
 		perl -pi -e 's/-mandroid//g' Configure
+		export ANDROID_NDK_HOME=$ANDROID_NDK_ROOT
 		# Use env to make all these temporary, so they don't pollute later builds.
-		env SYSTEM=android \
-			CROSS_COMPILE=${BUILDCHAIN}- \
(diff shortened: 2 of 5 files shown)


janmulder:
Interestingly, the build fails in a different way as I expected. `libgit2` compiles (ok, with a million of warnings, but its always like that). And a bit later the log says `Checking for module 'libgit2'  Package 'openssl', required by 'libgit2', not found. `

But, long before that, `libzip` is successfully compiled and logging:  `Found OpenSSL: /android/ndk-arm/sysroot/usr/lib/libcrypto.so (found version "1.1.1c")  `

Hmm. Strange.

janmulder:
Ok, first build error was fixed by the force pushed change ...

But know we have a new one at the very end generating the apk: `Unable to find toolchain: /android/subsurface/../android-ndk-r18b/toolchains/x86-64-4.9/prebuilt`. A very subtle one, as on my own build that directory has a slightly different name. The `x86-64-4.9` from Travis reads `x86_64-4.9` on my filesystem (watch the dash vs. underscore).

dirkhh:
I can look at that today. Those usually aren't too hard to fix
-- 
From my phone

dirkhh:
It appears that it is looking for part of the toolchain that I manually remove because in the past we didn't need it.
Let me create a Qt 5.12.4 based build container which contains those missing pieces (I'm on the road, I need to do that at home where I have two orders of magnitude faster internet... I hope I'll get to it today)

janmulder:
Curious to see what you come up with :-) 

Comparing a locally running docker container, I see that `android-ndk/android-ndk-r18b/toolchains/x86-4.9` is missing, but that is not what the compile error suggests.

But a new container is useful anyway due to the new Qt version.

janmulder:
Superseded by #2182. Close this PR.

dirkhh:
Here's what bothers me. The Travis log is for 4.8.4.702
Yet when I pull that exact branch at home it's 4.8.4.687
What extra 15 commits is Travis building? Am I just confusing myself?

janmulder:
I never looked into much detail to the output of `git describe ` command, but I agree that this seems strange. When is even more strange is that full output of the describe is: v4.8.4-702-g90caa4219, and that the abbreviated SHA is not a commit in our repo.

But, I think we are confusing ourselves here. Linus can probably explain what is likely going on here :smile:

dirkhh:
OK, poking at this some more it makes sense. Travis merges the branch on top of master before building it. That should make no difference whatsoever - but I'll rebase my local branch and push that so that we are really building the same thing on both ends.

PR #2183, opened 2019-07-14
Misc fixes for loading/saving

torvalds:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes two bugs:

 - avoid a crash when saving invalid cylinder usage information

 - save RBT data properly in the XML format

The RBT data fix came from me testing the "load then save" fix, and noticing that the RBT in the save didn't match the RBT in the load.

Of course, because we mis-saved the RBT data originally, it now means that every old file has bogus RBT data that we will mis-load, but at least going forward we should be ok. And the RBT problem really only happens when RBT goes down to zero, which it obviously newer should - you'd be a bad diver if your dive computer reports RBT as zero.

What can I say? "Palau" and "there were still things to see".

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -23,7 +23,7 @@
  * here */
 struct dive displayed_dive;
 
-const char *cylinderuse_text[] = {
+const char *cylinderuse_text[NUM_GAS_USE] = {
 	QT_TRANSLATE_NOOP("gettextFromC", "OC-gas"), QT_TRANSLATE_NOOP("gettextFromC", "diluent"), QT_TRANSLATE_NOOP("gettextFromC", "oxygen"), QT_TRANSLATE_NOOP("gettextFromC", "not used")
 };
 
--- a/core/dive.h
+++ b/core/dive.h
@@ -23,7 +23,7 @@ extern int last_xml_version;
 
 enum divemode_t {OC, CCR, PSCR, FREEDIVE, NUM_DIVEMODE, UNDEF_COMP_TYPE};	// Flags (Open-circuit and Closed-circuit-rebreather) for setting dive computer type
 
-extern const char *cylinderuse_text[];
+extern const char *cylinderuse_text[NUM_GAS_USE];
 extern const char *divemode_text_ui[];
 extern const char *divemode_text[];
 
--- a/core/save-git.c
+++ b/core/save-git.c
@@ -140,6 +140,7 @@ static void save_cylinder_info(struct membuffer *b, struct dive *dive)
 		cylinder_t *cylinder = dive->cylinder + i;
 		int volume = cylinder->type.size.mliter;
 		const char *description = cylinder->type.description;
+		int use = cylinder->cylinder_use;
 
 		put_string(b, "cylinder");
 		if (volume)
@@ -150,8 +151,8 @@ static void save_cylinder_info(struct membuffer *b, struct dive *dive)
 		put_gasmix(b, cylinder->gasmix);
 		put_pressure(b, cylinder->start, " start=", "bar");
 		put_pressure(b, cylinder->end, " end=", "bar");
-		if (cylinder->cylinder_use != OC_GAS)
-			put_format(b, " use=%s", cylinderuse_text[cylinder->cylinder_use]);
+		if (use > OC_GAS && use < NUM_GAS_USE)
+			show_utf8(b, " use=", cylinderuse_text[use], "");
 		if (cylinder->depth.mm != 0)
 			put_milli(b, " depth=", cylinder->depth.mm, "m");
 		put_string(b, "\n");
--- a/core/save-xml.c
+++ b/core/save-xml.c
@@ -180,6 +180,7 @@ static void save_cylinder_info(struct membuffer *b, struct dive *dive)
 		cylinder_t *cylinder = dive->cylinder + i;
 		int volume = cylinder->type.size.mliter;
 		const char *description = cylinder->type.description;
+		int use = cylinder->cylinder_use;
 
 		put_format(b, "  <cylinder");
 		if (volume)
@@ -189,8 +190,8 @@ static void save_cylinder_info(struct membuffer *b, struct dive *dive)
 		put_gasmix(b, cylinder->gasmix);
 		put_pressure(b, cylinder->start, " start='", " bar'");
 		put_pressure(b, cylinder->end, " end='", " bar'");
-		if (cylinder->cylinder_use != OC_GAS)
-			show_utf8(b, cylinderuse_text[cylinder->cylinder_use], " use='", "'", 1);
+		if (use > OC_GAS && use < NUM_GAS_USE)
+			show_utf8(b, cylinderuse_text[use], " use='", "'", 1);
 		if (cylinder->depth.mm != 0)
 			put_milli(b, " depth='", cylinder->depth.mm, " m'");
 		put_format(b, " />\n");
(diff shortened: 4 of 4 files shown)


bstoeger:
I'm just heading off to dinner, will have a look later. Just two very quick comments:

1) Shouldn't an unparseable cylinderuse already be caught at load time (i.e. in core/load-git.c:367)?
2) Do I understand correctly that the cylinderuse saved is a translated version, i.e. save with one locale and load with another and the data is lost? Sounds like a terrible idea.

dirkhh (approved):
Looks good. Waiting for Travis

dirkhh:
> Do I understand correctly that the cylinderuse saved is a translated version, i.e. save with one locale and load with another and the data is lost? Sounds like a terrible idea.

I need to stare at the code some more. And no, we should NEVER save translated strings for these things...

torvalds:
> I'm just heading off to dinner, will have a look later. Just two very quick comments:
> 
> 1. Shouldn't an unparseable cylinderuse already be caught at load time (i.e. in core/load-git.c:367)?

Yes. But it sets 'use' to -1, which then will cause problem when saving, because it will make us load the usage string from cylinderuse_text[-1], which gets us a likely horribly bad pointer, and then when we try to "print" that, we SIGSEGV.

Could we just ignore it entirely and just set usage to zero? Yes.

> 2. Do I understand correctly that the cylinderuse saved is a translated version, i.e. save with one
> locale and load with another and the data is lost? Sounds like a terrible idea.

No, we should never use locale when saving the data files. That is indeed a terrible idea.

But we've done it by mistake several times. Sometimes it's been about incorrect translation strings ("it's English and user visible, so we should translate it"), and sometimes it's been about other subtler things (we had several cases where we used floating point, and then had commas in the floating point values in non-US locales, which doesn't work so well when you use commas as delimiters).

So we don't translate when saving, but we test against the original _and_ translated string when loading, because we've clearly incorrectly translated at save time at some point (I assume - I didn't actually go and check)..

bstoeger (approved):
Look reasonable to me.

bstoeger:
Oooops. Current master now crashes for me on startup with
```

0x00007fffce9f716e in QDeclarativeGeoMapItemView::setDelegate(QQmlComponent*) () from /usr/lib/x86_64-linux-gnu/libQt5Location.so.5
(gdb) where
#0  0x00007fffce9f716e in QDeclarativeGeoMapItemView::setDelegate(QQmlComponent*) () at /usr/lib/x86_64-linux-gnu/libQt5Location.so.5
#1  0x00007fffcea779d8 in  () at /usr/lib/x86_64-linux-gnu/libQt5Location.so.5
#2  0x00007fffcea77b13 in QDeclarativeGeoMapItemView::qt_metacall(QMetaObject::Call, int, void**) () at /usr/lib/x86_64-linux-gnu/libQt5Location.so.5
#3  0x00007ffff33df4b5 in QQmlObjectCreator::setPropertyBinding(QQmlPropertyData const*, QV4::CompiledData::Binding const*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#4  0x00007ffff33e0825 in QQmlObjectCreator::setupBindings(bool) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#5  0x00007ffff33e17f4 in QQmlObjectCreator::populateInstance(int, QObject*, QObject*, QQmlPropertyData const*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#6  0x00007ffff33e260b in QQmlObjectCreator::createInstance(int, QObject*, bool) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#7  0x00007ffff33dfcbe in QQmlObjectCreator::setPropertyBinding(QQmlPropertyData const*, QV4::CompiledData::Binding const*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#8  0x00007ffff33e0825 in QQmlObjectCreator::setupBindings(bool) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#9  0x00007ffff33e17f4 in QQmlObjectCreator::populateInstance(int, QObject*, QObject*, QQmlPropertyData const*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#10 0x00007ffff33e260b in QQmlObjectCreator::createInstance(int, QObject*, bool) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#11 0x00007ffff33dfcbe in QQmlObjectCreator::setPropertyBinding(QQmlPropertyData const*, QV4::CompiledData::Binding const*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#12 0x00007ffff33e0825 in QQmlObjectCreator::setupBindings(bool) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#13 0x00007ffff33e17f4 in QQmlObjectCreator::populateInstance(int, QObject*, QObject*, QQmlPropertyData const*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#14 0x00007ffff33e260b in QQmlObjectCreator::createInstance(int, QObject*, bool) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#15 0x00007ffff33e2c42 in QQmlObjectCreator::create(int, QObject*, QQmlInstantiationInterrupt*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#16 0x00007ffff33542d0 in QQmlComponentPrivate::beginCreate(QQmlContextData*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#17 0x00007ffff33527af in QQmlComponent::create(QQmlContext*) () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#18 0x00007ffff458dc2b in QQuickWidget::continueExecute() () at /usr/lib/x86_64-linux-gnu/libQt5QuickWidgets.so.5
#19 0x00007ffff458e178 in  () at /usr/lib/x86_64-linux-gnu/libQt5QuickWidgets.so.5
#20 0x00005555556c9740 in MapWidget::MapWidget(QWidget*) (this=0x55555619cb70, parent=<optimized out>) at /home/bs/src/subsurface/desktop-widgets/mapwidget.cpp:32
#21 0x00005555556c9ba7 in MapWidget::instance() () at /home/bs/src/subsurface/desktop-widgets/mapwidget.cpp:107
#22 0x00005555556df70d in MainTab::MainTab(QWidget*) (this=0x555555f69580, parent=<optimized out>) at /home/bs/src/subsurface/desktop-widgets/tab-widgets/maintab.cpp:87
#23 0x00005555556bd16c in MainWindow::MainWindow() (this=0x555555df2b30) at /home/bs/src/subsurface/desktop-widgets/mainwindow.cpp:139
#24 0x00005555556b1e20 in init_ui() () at /home/bs/src/subsurface/subsurface-helper.cpp:167
#25 0x00005555556ac599 in main(int, char**) (argc=<optimized out>, argv=<optimized out>) at /home/bs/src/subsurface/subsurface-desktop-main.cpp:92
```
:(

bstoeger:
> So we don't translate when saving, but we test against the original and translated string when loading, because we've clearly incorrectly translated at save time at some point (I assume - I didn't actually go and check)..

Indeed, that's what the code seems to (correctly) do. Then I wonder how Steve managed to get an invalid use string. Perhaps it's a different memory corruption after all...?

bstoeger:
> Oooops. Current master now crashes for me on startup with

Unrelated to this PR - also crashes when checking out an older commit.

dirkhh:
That crash could be a mix with a googlemapwidget built against a different Qt version.
Some distros (like arch) just recently updated to Qt 5.13. Just to avoid being caught in something like that - is this a remote possibility?
A full build against Qt 5.13 on Arch Linux of latest master doesn't crash for me.

PR #2186, opened 2019-07-17
Undo: make editing of dive number undoable

bstoeger:
When pressing F2 in the dive list, the number can be edited.
Make this action undoable by implementing a EditNumber command.

This command is differs from the other undo commands, as not the
currently selected dives are changed. This means that the EditCommand
needs an alternative constructor taking a single dive. This constructor
was implemented in the base class so that all edit commands can now
be called with a single dive.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Somehow I missed that particular dive-edit command. Make it undoable as well. It *is* well hidden though - it works for me only if I press "F2". That's quite obscure!

### Changes made:

1) Implement EditNumber undo command.
2) Use undo command for setting the dive number.
### Release note:

No.
### Documentation change:

Is the F2 thing noted anywhere?

### DIFF
--- a/core/cochran.c
+++ b/core/cochran.c
@@ -787,7 +787,6 @@ static void cochran_parse_dive(const unsigned char *decode, unsigned mod,
 	}
 
 	record_dive_to_table(dive, table);
-	mark_divelist_changed(true);
 
 	free(buf);
 }
--- a/core/dive.c
+++ b/core/dive.c
@@ -3215,8 +3215,6 @@ static int split_dive_at(const struct dive *dive, int a, int b, struct dive **ou
 			d2->number = 0;
 	}
 
-	mark_divelist_changed(true);
-
 	*out1 = d1;
 	*out2 = d2;
 	return nr;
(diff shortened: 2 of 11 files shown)


dirkhh:
F2? What the heck? I do not believe that this is an intentional user interface. I wonder if this is something we get by default from Qt and we just never noticed and didn't turn it off?
I did some googling and can't find it, though. Nor can I find a reference to F2 in our sources...
Puzzled

bstoeger:
The #-field should be editable by clicking into it, and I'm sure that it was at some point. For some reason now it works only when pressing F2.

See the section "key bindings" here: https://doc.qt.io/qt-5/qtreeview.html

henrik242:
> The #-field should be editable by clicking into it, and I'm sure that it was at some point. For some reason now it works only when pressing F2.

F2 is used for "edit cell" in Excel and other spreadsheets, so it's probably somewhat related.

PR #2191, opened 2019-07-24
Make git parser (nearly) reentrant

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a split-off from the cylinder-generalization code. I wanted to make sure that sensor readings have valid cylinder ids. This needs access to the dive when adding a sample. That was trivial in the XML parser, but hard in the git parser. There, we only have a pointer to the dive computer.

Therefore, this PR adopts a similar state structure as in the XML parser. In any case, removing global state seems like a good idea. As an added bonus, there is less weak typing by passing through "void *" pointers. I'm not sure if the initial idea is even valid - after all it means that the cylinders must be parsed before the samples.

### Changes made:

1) Use a state struct instead of global variables.
2) Minor whitespace fixes.

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1488,7 +1488,7 @@ void clear_dive_file_data()
 	clear_dive(&displayed_dive);
 
 	reset_min_datafile_version();
-	saved_git_id = "";
+	clear_git_id();
 }
 
 bool dive_less_than(const struct dive *a, const struct dive *b)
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
This all looks good to me. Just because of which code this is, would you mind poking Linus for an ACK as well?

torvalds:
Looks fine to me. So ACK.

PR #2192, opened 2019-07-25
[RFC] Core: split copy_cylinders() in two functions

bstoeger:
copy_cylinders() copied the cylinders of one dive onto another dive
and then reset to the original gas values. Presumably, when copy and
pasting cylinders from one dive to another, only the types should
be copied, not the gases.

Moreover, the function could either copy all or only the used cylinders.

Firstly, the code was bogus: when restoring the pressures the indices
were mixed up: the old indices were used. Thus, when there where
uncopied cylinders, not all pressure values were restored.

Secondly, it is not clear that all callers actually want to restore
the pressure data. It rather appears the two (out of three) callers
actually just want to copy the cylinders.

Therefore, split the function in
  1) copy_cylinders(): copy the cylinders with pressure data
  2) copy_cylinder_types(): copy only the cylinder information

Since there is only one caller of copy_cylinder_types(), the "used_only"
argument can be removed. Since all cylinders are copied there is
no point in storing the pressure data. Don't overwrite it in
the first place.

The resulting two functions should be distinctly easier to understand.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a borderline bug fix. The original `copy_cylinders()` function was clearly bogus. The original pressure values were not restored correctly if cylinders were skipped.

I hope that the split functions are much more clear. Compile & run tested.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -433,6 +433,7 @@ struct dive *move_dive(struct dive *s)
 		d->_component = copy_string(s->_component)
 
 // copy elements, depending on bits in what that are set
+static void copy_cylinder_types(const struct dive *s, struct dive *d);
 void selective_copy_dive(const struct dive *s, struct dive *d, struct dive_components what, bool clear)
 {
 	if (clear)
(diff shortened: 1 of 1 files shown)


bstoeger:
Got a "Looks fine to me" from @torvalds by mail.

bstoeger:
Readded an inadvertently removed `free()`.

dirkhh (approved):
I see nothing wrong here - but this isn't code that I'm very familiar with

PR #2196, opened 2019-07-27
Avoid possible crash while loading maps

dirkhh:
We ended up with a NULL location and happily dereferenced it. Not good.

Signed-off-by: Dirk Hohndel <dirk@hohndel.org>

### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Using the dive data from one of the reporters of the 4.9 crash I was able to reproduce the crash and this fixes it
### Mentions:

@bstoeger  would you do a sanity check, please?

PR #2197, opened 2019-07-27
Crash fix: prevent crash in MapLocationModel::reload

bstoeger:
Commit 0c387549164d7eec3ea6647c54ada2fba7f8d5e6 introduced a
bug in MapLocationModel::reload() by setting an entry in the
name-to-location map before the location was initialized.

Move the setting of the map entry back where it was before:
after the assignment of the location variable.

Moreover, define the location variable directly on allocation
of the location to avoid thus bugs in the future.

Why did we not get a "might be used unitialized" warning
anyway?

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Attempt at crash-fix as requested in #2196.
### Mentions:

@dirkhh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-
+- fix crash when (re-)loading maps
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
 * Use this layout for new entries: `[Area]: [Details about the change] [reference thread / issue]`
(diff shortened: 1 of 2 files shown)


bstoeger:
This all raises the question why we didn't get warnings of the compiler or static checkers? Are the warning levels too low?

dirkhh (approved):
Yep, this makes more sense.

PR #2202, opened 2019-07-29
[RFC] Add Statistics Graphs for Monthly and Yearly Statistics

djunkins:
This is the first step in adding graphs to visualize statistics about the dive log. I would appreciate feedback on the UI decisions I made and the coding choices for the model and widget before I proceed with adding additional graphs. Next on the list will be a graph showing the number of dives that include each Tag in the global tag list.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Adds a new menu option in the "View" menu for Statistics graphs. Clicking the menu option will open a modal window with stats graphs for Yearly stats and Monthly stats. Additional graphs can be added as new tabs in the window.

### Changes made:

1) Create DateStatsTableModel to store count of dives in each month and year in the dive log
2) Create ChartWidget with a tab widget and a statistics chart on each tab
3) Adds "View->Statistics graph" menu item with a keyboard shortcut of Alt-G

### DIFF
--- a/desktop-widgets/chartwidget.h
+++ b/desktop-widgets/chartwidget.h
@@ -1,0 +1,26 @@
+#ifndef CHARTWIDGET_H
+#define CHARTWIDGET_H
+
+#include <QWidget>
+#include "ui_chartwidget.h"
+#include "qt-models/datestatsmodel.h"
+
+namespace Ui {
+class ChartWidget;
+}
+
+class ChartWidget : public QWidget
+{
+	Q_OBJECT
+
+public:
+	explicit ChartWidget(QWidget *parent = nullptr);
+	void DrawYearChart();
+	void DrawMonthChart();
+
+private:
+	Ui::ChartWidget ui;
+	DateStatsTableModel model;
+};
+
+#endif // STATSCHARTWIDGET_H
--- a/desktop-widgets/chartwidget.h
+++ b/desktop-widgets/chartwidget.h
@@ -7,7 +7,9 @@
 #include <QVBarModelMapper>
 #include <QHBarModelMapper>
 #include "ui_chartwidget.h"
+#include "qt-models/buddystatsmodel.h"
 #include "qt-models/datestatsmodel.h"
+#include "qt-models/tagstatsmodel.h"
 #include "core/subsurface-qt/DiveListNotifier.h"
 
 class ChartWidget : public QWidget
@@ -19,6 +21,8 @@ public:
 	~ChartWidget();
 	QtCharts::QChart *drawYearChart();
 	QtCharts::QChart *drawMonthChart();
+	QtCharts::QChart *drawTagChart();
+	QtCharts::QChart *drawBuddyChart();
 
 public
 slots:
@@ -27,10 +31,16 @@ slots:
 private:
 	Ui::ChartWidget ui;
 	DateStatsTableModel *model;
+	TagStatsListModel *tagModel;
+	BuddyStatsListModel *buddyModel;
 	QtCharts::QChartView monthChartView;
 	QtCharts::QChartView yearChartView;
+	QtCharts::QChartView tagChartView;
+	QtCharts::QChartView buddyChartView;
 	QtCharts::QHBarModelMapper yearMapper;
 	QtCharts::QVBarModelMapper monthMapper;
+	QtCharts::QVBarModelMapper tagMapper;
+	QtCharts::QVBarModelMapper buddyMapper;
 };
 
 #endif // STATSCHARTWIDGET_H
--- a/desktop-widgets/chartwidget.h
+++ b/desktop-widgets/chartwidget.h
@@ -29,6 +29,8 @@ slots:
 	void filterFinished();
 
 private:
+	int calcYAxisTicks(int maxDiveCount);
+	int calcYAxisMax(int maxDiveCount);
 	Ui::ChartWidget ui;
 	DateStatsTableModel *model;
 	TagStatsListModel *tagModel;
--- a/desktop-widgets/chartwidget.h
+++ b/desktop-widgets/chartwidget.h
@@ -2,12 +2,13 @@
 #define CHARTWIDGET_H
 
 #include <QWidget>
+#include <QChart>
+#include <QChartView>
+#include <QVBarModelMapper>
+#include <QHBarModelMapper>
 #include "ui_chartwidget.h"
 #include "qt-models/datestatsmodel.h"
-
-namespace Ui {
-class ChartWidget;
-}
+#include "core/subsurface-qt/DiveListNotifier.h"
 
 class ChartWidget : public QWidget
 {
@@ -15,12 +16,21 @@ class ChartWidget : public QWidget
 
 public:
 	explicit ChartWidget(QWidget *parent = nullptr);
-	void DrawYearChart();
-	void DrawMonthChart();
+	~ChartWidget();
+	QtCharts::QChart *drawYearChart();
+	QtCharts::QChart *drawMonthChart();
+
+public
+slots:
+	void filterFinished();
 
 private:
 	Ui::ChartWidget ui;
-	DateStatsTableModel model;
+	DateStatsTableModel *model;
+	QtCharts::QChartView monthChartView;
+	QtCharts::QChartView yearChartView;
+	QtCharts::QHBarModelMapper yearMapper;
+	QtCharts::QVBarModelMapper monthMapper;
 };
 
 #endif // STATSCHARTWIDGET_H
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -1027,7 +1027,7 @@ void MainWindow::on_actionStatsCharts_triggered()
 {
 // Check if charts are being included
 #ifndef NO_CHARTS
-	statsChartWidget = new ChartWidget();
+	statsChartWidget.reset(new ChartWidget(this));
 	statsChartWidget->show();
 #endif
 }
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -217,6 +217,11 @@ MainWindow::MainWindow() : QMainWindow(),
 	plannerDetails->printPlan()->hide();
 	ui.menuFile->removeAction(ui.actionPrint);
 #endif
+
+#ifdef NO_CHARTS
+	ui.menuView->removeAction(ui.actionStatsCharts);
+#endif
+
 	enableDisableCloudActions();
 
 	ui.mainErrorMessage->hide();
@@ -1016,6 +1021,15 @@ void MainWindow::on_actionYearlyStatistics_triggered()
 	d.setWindowTitle(tr("Yearly statistics"));
 	d.setWindowIcon(QIcon(":subsurface-icon"));
 	d.exec();
+}
+
+void MainWindow::on_actionStatsCharts_triggered()
+{
+// Check if charts are being included
+#ifndef NO_CHARTS
+	statsChartWidget = new ChartWidget();
+	statsChartWidget->show();
+#endif
 }
 
 void MainWindow::toggleCollapsible(bool toggle)
--- a/desktop-widgets/mainwindow.h
+++ b/desktop-widgets/mainwindow.h
@@ -223,7 +223,7 @@ private:
 	LocationInformationWidget *diveSiteEdit;
 
 #ifndef NO_CHARTS
-	ChartWidget *statsChartWidget;
+	std::unique_ptr<ChartWidget> statsChartWidget;
 #endif
 
 	bool plannerStateClean();
--- a/desktop-widgets/mainwindow.h
+++ b/desktop-widgets/mainwindow.h
@@ -23,6 +23,10 @@
 #include "core/gpslocation.h"
 #include "core/dive.h"
 
+#ifndef NO_CHARTS
+#include "desktop-widgets/chartwidget.h"
+#endif
+
 #define NUM_RECENT_FILES 4
 
 class QSortFilterProxyModel;
@@ -116,6 +120,7 @@ slots:
 	void on_actionRenumber_triggered();
 	void on_actionAutoGroup_triggered();
 	void on_actionYearlyStatistics_triggered();
+	void on_actionStatsCharts_triggered();
 
 	/* view menu actions */
 	void on_actionViewList_triggered();
@@ -216,6 +221,10 @@ private:
 	bool filesAsArguments;
 	UpdateManager *updateManager;
 	LocationInformationWidget *diveSiteEdit;
+
+#ifndef NO_CHARTS
+	ChartWidget *statsChartWidget;
+#endif
 
 	bool plannerStateClean();
 	void setupForAddAndPlan(const char *model);
--- a/qt-models/datestatsmodel.cpp
+++ b/qt-models/datestatsmodel.cpp
@@ -1,0 +1,125 @@
+#include "datestatsmodel.h"
+#include "core/dive.h"
+#include "core/divelist.h"
+#include "core/time.c"
+#include <QtCore/QVector>
+
+DateStatsTableModel::DateStatsTableModel(QObject *parent) :
+	QAbstractTableModel(parent)
+{
+	int idx;
+	struct dive *dp;
+	struct tm tm;
+	int current_year = -1;
+	QVector<int> yearVec(YEAR + 1, 0);
+
+	// Iterate through dives and count for each month and year
+	for_each_dive(idx, dp) {
+		utc_mkdate(dp->when, &tm);
+		// Advance current_year to the current dive, adding an empty
+		// QVector for each year
+		while (tm.tm_year != current_year) {
+			// Check if this is the first dive
+			if (current_year == -1) {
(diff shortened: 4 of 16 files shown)


djunkins:
There is no rush to review or merge this pull request given the focus on the 4.9 release and getting the 4.9.1 bug fixes out.

dirkhh:
For obvious reasons I'd love to hear what @tcanabrava thinks about this :-)

My goal for this week is to get 4.9.1 out (I'm currently planning to do this on Thursday), so the timing is perfect to review this in the meantime.

desktop-widgets/charts/chartwidget.cpp
```diff
@@ -0,0 +1,139 @@
...
+        model = new DateStatsTableModel;
+
+        DrawYearChart();
+        DrawMonthChart();
+}
+
+ChartWidget::~ChartWidget()
+{
+        delete ui;
+        delete model;
```
bstoeger: Can we please stop using 1980ies style memory management?
Simple way, perhaps longer compilation times:
- make `ui` and `model` a member-object and initialize in initializer list.

More complex way:
- make `ui` and `model` of type `std::unique_ptr<...>`.

qt-models/datestatsmodel.cpp
```diff
@@ -0,0 +1,141 @@
...
+                return QVariant();
+
+        if (orientation == Qt::Horizontal)
+                // Return QString with Year lable in it.
+                return QString("%1").arg(m_data[section]->at(YEAR));
+        else
+                // Return labels for months and total.
+                switch (section) {
+                case TOTAL:
+                        return QString(tr("Total"));
```
bstoeger: This casts a QString to a QString, which is redundant.

qt-models/datestatsmodel.cpp
```diff
@@ -0,0 +1,141 @@
...
+                        break;
```
bstoeger: This break will never be reached.

qt-models/datestatsmodel.cpp
```diff
@@ -0,0 +1,141 @@
...
+QVariant DateStatsTableModel::headerData(int section, Qt::Orientation orientation, int role) const
+{
+        if (role != Qt::DisplayRole)
+                return QVariant();
+
+        if (orientation == Qt::Horizontal)
+                // Return QString with Year lable in it.
+                return QString("%1").arg(m_data[section]->at(YEAR));
+        else
+                // Return labels for months and total.
```
bstoeger: According to coding style we add braces for complex multi-line statements.

qt-models/datestatsmodel.h
```diff
@@ -0,0 +1,42 @@
...
+        int rowCount(const QModelIndex &parent = QModelIndex()) const;
+        int columnCount(const QModelIndex &parent = QModelIndex()) const;
+        QVariant headerData(int section, Qt::Orientation orientation, int role = Qt::DisplayRole) const;
+        QVariant data(const QModelIndex &index, int role = Qt::DisplayRole) const;
+        int axisMax(Qt::Orientation orientation) const;
+
+private:
+        QList<QVector<int> * > m_data;
+        int m_columnCount;
+        int m_rowCount;
```
bstoeger: This member-variable appears redundant. Can't we just return the correct value in the function?

qt-models/datestatsmodel.h
```diff
@@ -0,0 +1,42 @@
...
+
+        int rowCount(const QModelIndex &parent = QModelIndex()) const;
+        int columnCount(const QModelIndex &parent = QModelIndex()) const;
+        QVariant headerData(int section, Qt::Orientation orientation, int role = Qt::DisplayRole) const;
+        QVariant data(const QModelIndex &index, int role = Qt::DisplayRole) const;
+        int axisMax(Qt::Orientation orientation) const;
+
+private:
+        QList<QVector<int> * > m_data;
+        int m_columnCount;
```
bstoeger: This member variable is not used anywhere...?

qt-models/datestatsmodel.h
```diff
@@ -0,0 +1,42 @@
...
+        explicit DateStatsTableModel(QObject *parent = 0);
+
+        int rowCount(const QModelIndex &parent = QModelIndex()) const;
+        int columnCount(const QModelIndex &parent = QModelIndex()) const;
+        QVariant headerData(int section, Qt::Orientation orientation, int role = Qt::DisplayRole) const;
+        QVariant data(const QModelIndex &index, int role = Qt::DisplayRole) const;
+        int axisMax(Qt::Orientation orientation) const;
+
+private:
+        QList<QVector<int> * > m_data;
```
bstoeger: A QList of *pointers* to QVectors...
Note that this indirection is completely unnecessary as QList will store pointers to non-trivial objects anyway.

qt-models/datestatsmodel.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+DateStatsTableModel::DateStatsTableModel(QObject *parent) :
+        QAbstractTableModel(parent)
+{
+        int idx;
+        struct dive *dp;
+        struct tm tm;
+        m_rowCount = YEAR;
+        int current_year = -1;
+        QVector<int>* yearVec;
```
bstoeger: No pointer here, please.

qt-models/datestatsmodel.cpp
```diff
@@ -0,0 +1,141 @@
...
+                        // Check if this is the first dive
+                        if (current_year == -1) {
+                                yearVec = new QVector<int>(YEAR + 1);
+                                yearVec->replace(YEAR, tm.tm_year);
+                                current_year = tm.tm_year;
+                        } else {
+                                m_data.append(yearVec);
+                                yearVec = new QVector<int>(YEAR + 1);
+                                current_year++;
+                                yearVec->replace(YEAR, current_year);
```
bstoeger: The idiomatic way would be a simple `yearVec[YEAR]=current_year`.

qt-models/datestatsmodel.cpp
bstoeger: If I'm interpreting the code correctly, I would initialize this by a fixed size vector
```
QVector<int> yearVec(YEAR + 1, 0);
```
Then add the vector with
```
m_data.append(yearVec);
```
and reset it with
```
yearVec.fill(0);
yearVec[YEAR] = xxx;
```

bstoeger:
Only a few notes, will look closer later that week.

@djunkins: Just so that I interpret the code correctly - could it be that you come from a Java background? In "modern" C++ `new` is a "code smell" and it's all about value semantics. This is not meant as an offense - I'm just wondering about the habit of allocating everything.

Sadly, Qt's containers are completely broken with respect to value semantics. :(

djunkins:
@bstoeger I don't have much of a background in java -- mostly C, python and, a long time ago, perl. My habit of allocating everything and relying on pointers probably comes from C. I based this code on the sample code from the QtCharts pages which is apparently outdated in style. I think I understand the point you are making about focusing on the values that you want to pass rather than worrying about allocating the object itself. I'll take another stab at this and will address your comments above.

djunkins:
LGTM C/C++ analysis is failing due to some features I used that were introduced in Qt 5.12. I'll have to figure out a different way to do the Y-axis ticks that is compatible back to Qt 5.8.

bstoeger:
> I don't have much of a background in java -- mostly C, python and, a long time ago, perl. My habit of allocating everything and relying on pointers probably comes from C. 

Many thanks for the info. It's interesting for me to see where certain habits come from and I hope I didn't appear condescending. Yes, much of the code base is very traditionally C-ish and it's not easy to find the ideal compromise between consistency and a more modern style that avoids memory leaks, etc.

I will try to have a closer look at this PR this week.

djunkins:
Travis is failing on some checks due to a lack of the Qt5Charts library on some test environments. On Ubuntu 18.04, adding the  libqt5charts5 and libqt5charts5-dev packages resolved this. The Qt5Charts library was already included in my Qt package on Mac OSX, but that may have been because I downloaded everything when I installed Qt.

djunkins:
The qt59 travis check is now working, but I'm not sure how to add QtCharts to the custom Qt builds that are used for the other checks.

dirkhh:
The question how far back to support Qt is consistently driven by Linux distros. We (well, I) build with our own Qt on Windows, Mac, Android and iOS - and I think all of those are currently 5.11 or 5.12 (it's surprisingly painful to upgrade all my build environments...).
But Linux distros are what's really holding us back. And frankly, Linux is a fairly small part of our audience. In the early days it was 100% - then I ported Subsurface to Windows and Mac and Linux quickly dropped to about 1/5th. Today it's maybe 1/10th of our user base.
Ubuntu 16.04 is based on Qt5.5 - and while most people have moved on to 18.04 or newer, many have not...

djunkins:
There are other options for charting available (QWT and Javascript are two that I have run across), but they may be as difficult to integrate as QtCharts.

atdotde:
I had to manually build the charts target, it seems there is something missing an a CMakeLIsts.txt file. 

Why is this window modal? Why can't I put it in the background and for example modify the selection?

dirkhh:
> There are other options for charting available (QWT and Javascript are two that I have run across), but they may be as difficult to integrate as QtCharts.

I think QtCharts is the way to go, frankly...

djunkins:
@atdotde Did you have to add the QtCharts library or package in order to get it to build, or was there something else you had to modify to "manually build the charts target"? With the addition of the QtCharts package on Ubuntu 18.04 and including it in my Qt build on Mac OSX, the build script worked for me with the additions to the CMakeLists.txt files that were included in the pull request.

Regarding the modal window, as of now the graphs are drawn for the complete dive list and not for the filtered or selected dives, so my thought was that the window would be closed before changes to the dive log were made. Making the window non-modal and allowing the charts to be built for a subset of the dives in the dive log is something that can be added.

atdotde:
I had to run ccmake to tell it where to find the QtCharts library (I already had it installed). But then, it would not create the ui_chartwidget.h when building subsurface (make install or "Build" in creator), in creator I had to right click on "hammer subsurface_charts" to "Build subsurface_charts".

I think modal windows should be avoided as much as possible as they tend to be in the way and prevent the user from doing what they want.

bstoeger:
> But Linux distros are what's really holding us back. And frankly, Linux is a fairly small part of our audience.

Can't you conditionally compile the charts? If someone insists on using an outdated Linux distribution they can probably live without fancy statistics, no?

desktop-widgets/mainwindow.cpp
```diff
@@ -1018,6 +1019,12 @@ void MainWindow::on_actionYearlyStatistics_triggered()
 	d.exec();
 }
 
+void MainWindow::on_actionStatsCharts_triggered()
+{
+        statsChartWidget = new ChartWidget();
```
bstoeger: This has spaces instead of a tab.

desktop-widgets/mainwindow.h
```diff
@@ -217,6 +219,8 @@ public
 	UpdateManager *updateManager;
 	LocationInformationWidget *diveSiteEdit;
 
+	ChartWidget *statsChartWidget;
```
bstoeger: I don't understand the point of this member variable. It is set but never read and it is not initialized.

If the dialog is supposed modal, you could simply use a local variable:
```
	ChartWiddget chart;
	chart.show()
```
If the dialog is not supposed to be modal, but you can have multiple dialogs, you could use a local pointer:
```
	ChartWidget *chart = new Chart(this);
	chart->show();
```
If the dialog is non-modal and there is only supposed to be one dialog at a time, then this member variable is needed, but has to be initialized to null. Or use the magic of smart pointers:
```
	std::unique_ptr<ChartWidget> chart;
```
That will automatically set to null on creation, free on destruction and with a simple
```
	chart.reset(new ChartWidget(this))
```
will destroy the old object (if any) and replace by a new one.

henrik242:
> Ubuntu 16.04 is based on Qt5.5

Do we really need to support that old distros?  If they really like their old OS versions, they're probably just as fine with old (and well-working!) versions of Subsurface.

djunkins:
> Can't you conditionally compile the charts? If someone insists on using an outdated Linux distribution they can probably live without fancy statistics, no?

I was just thinking about this yesterday. I will see if I can add a check for the QtCharts library and compile if it exists (kind of like the printing features if QtWebKit is present.)

djunkins:
I made the addition of the Stats Charts conditional on the Qt5Charts library being found by cmake and fixed some of the other build issues @atdotde identified. I have not yet changed the window to a non-modal dialog.

djunkins:
It looks like the Mac travis check failed due to a timeout downloading a homebrew bottle, but the other checks are now working whether or not the Qt5Charts library is available using the conditional compile.

bstoeger:
Restarted the Mac build.

desktop-widgets/chartwidget.h
```diff
@@ -0,0 +1,26 @@
+#ifndef CHARTWIDGET_H
+#define CHARTWIDGET_H
+
+#include <QWidget>
+#include "ui_chartwidget.h"
+#include "qt-models/datestatsmodel.h"
+
+namespace Ui {
+class ChartWidget;
```
bstoeger: This forwad declaration makes no sense anymore, now that you include `ui_chartwidget.h`.

qt-models/datestatsmodel.cpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+QVariant DateStatsTableModel::headerData(int section, Qt::Orientation orientation, int role) const
+{
+	if (role != Qt::DisplayRole)
+		return QVariant();
+
+	if (orientation == Qt::Horizontal) {
+		// Return QString with Year lable in it.
+		return QString("%1").arg(m_data[section].at(YEAR));
```
bstoeger: 1) This should probably use `QStringLiteral`. Not that it is performance critical, just as a matter of principle.
2) We try to be locale aware. If this string is shown to the user, you should probably use "%L1". 

Alternative: use Locale::toString().

djunkins:
Updated the ChartWidget to be non-modal and to filter the dives being reported in the charts based on the current filter. When the filter changes or dives are added, deleted or edited, the charts will update with the new data.

bstoeger:
> When the filter changes or dives are added, deleted or edited, the charts will update with the new data.

Then in principle you don't have to regenerate the dialog, but could simply call show() on it. That said, I am fine with the current version. Haven't tested it though.

dirkhh:
I don't want to lose this one out of sight while battling other fires...
is this still work in progress, do you consider it ready for full review / merging?

djunkins:
> I don't want to lose this one out of sight while battling other fires...
> is this still work in progress, do you consider it ready for full review / merging?

I have another commit to push that will add two more tabs to the chart widget:

1) Bar chart with number of dives with each buddy
2) Bar chart with number of dives with each tag

Once I have that commit pushed, I think it will be ready for full review and merging. 

The open question in my mind is how to make sure the QtCharts library is installed. I think the changes I made to the build scripts and CMakeLists.txt files will prevent errors if the library is not present, but it would help to hear if people are able to successfully add the library and build with the charts enabled.

djunkins:
I pushed the commit with the two new charts over the weekend. Looking for feedback now on how to finish the review and merging of this into master.

dirkhh:
I built the latest of this, applied on top of latest master (just in case that might cause issues).
Yearly stats shows me two columns. Year 0 with no dives, and year 2019 with no dives (technically no dives in year 0 isn't wrong, but... and no dives this year is definitely wrong).
Months shows just the current year. The three months during which I went diving appear to have the correct numbers, all in brown. Below is a legend that shows 2019 as brown. All other colors are matched to year 0 and show no dives.
Tags looks about right, except that the font for the y axis is unreadably small (I have ~600 dives in my dive log)
Buddies has unreadable fonts for both x and y axis, so I can't even tell if the data are correct.

djunkins:
@dirkhh Thanks for the feedback, as discouraging as it was. I think that's the first time anyone has actually built the charts besides myself and using a log other than my own or the test logs in the subsurface/dives directory. I think for tags and buddies, a solution may be to limit to the top-25 (or so) tags and buddies. I will also work on making the y-axis ticks more adaptive to large numbers of dives.

I'm not sure where to start with the problems on the date-based charts, but I will try to build some test log files that hit any corner cases you may be running in to. If you have an anonymized log file you can share that is similar to yours, it would be useful.

Obviously there is a ways to go on this pull request.

dirkhh:
First, my apologies. I __NEVER__ want the feedback I provide to be discouraging.
Yes, testing with limited data is always the first path to overlook problems.
We have a massively huge data file that we use for testing... look at the end of `scripts/linux/before_install.sh` for how to access that :-)
When limiting the screen to N items you are almost certain to get requests to be able to scroll to a second / third / etc page...

djunkins:
@dirkhh No worries. I just meant that it was discouraging when it didn't work like I expected, but given the limited number of test logs I used, that is not unexpected. I'll figure out how to download the huge test set and do some more work on this. I'll let you know when I think it is in a better place.

Cheers.

djunkins:
I'm going to close this pull request for now. I need to go back and look at how to do this more cleanly with the scaling for large numbers of dives, buddies and tags. I'll keep working and will open a new pull request when I have a better solution coded.

PR #2205, opened 2019-07-31
fix site/location build issue

salvadorcunat:
[SmartTrak import]
We were assuming these tables were sorted with their indexes, but it
happens to be false, under some circustances at least, and results in merged or wrong locations/sites.

Simply check the index fixes the issue.

ATM I don't know what makes the index to be unsorted, may be a big number of dive sites, may be editing an old dive point ...

Reported-by: Andreas Hagberg <scubasoft@gmail.com>
Signed-off-by: Salvador Cuñat <salvador.cunat@gmail.com>

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@atdotde

### DIFF
--- a/smtk-import/smartrak.c
+++ b/smtk-import/smartrak.c
@@ -334,8 +334,9 @@ static void smtk_build_location(MdbHandle *mdb, char *idx, struct dive_site **lo
 	if (!table)
 		return;
 
-	for (i = 1; i <= atoi(idx); i++)
+	do {
 		mdb_fetch_row(table);
+	} while (strcasecmp(col[0]->bind_ptr, idx));
 	loc_idx = copy_string(col[2]->bind_ptr);
 	site = copy_string(col[1]->bind_ptr);
 	loc = create_location(strtod(col[6]->bind_ptr, NULL), strtod(col[7]->bind_ptr, NULL));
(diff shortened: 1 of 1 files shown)


bstoeger:
Looks good to me in principle, but what happens if for some crazy reason the index is not found? Looks like an infinite loop to me.

I suggest that we try to handle that situation gracefully, as that shouldn't be too hard.

salvadorcunat:
Hi @bstoeger .
You are right there would be an infinite loop. Take a look to the other patch I've just send.
Nevertheless, the only reasons I can think of for the failure would be a damaged db which should be catch by the previous smtk-open-table() call or a dive reffering to a non existent index which should be really weird as the .slg file would fail on the smarttrak application too.

bstoeger:
> Nevertheless, the only reasons I can think of for the failure would be a damaged db which should be catch by the previous smtk-open-table() call or a dive reffering to a non existent index which should be really weird as the .slg file would fail on the smarttrak application too.

Sure, but since handling it is not hard, I think we should try it. Your patch leaks memory, but it's not terribly important as this is a "shouldn't happen" situation.

atdotde:
I installed the original patch on the web service and this indeed fixes the reporter's problem.

atdotde:
Restarted the Mac build.

PR #2208, opened 2019-08-04
Remove MAX_CYLINDERS limit

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In analogy to the weightsystems, keep cylinders in dynamically resized tables. Since only few people will use 10+ cylinders, this should actually use less memory, especially for the profile display, where we allocated data for 30 cylinders for every point on the graph(!). Obviously, this was distinctly more work than the weightsystem change. This is preparation for the undoization of the equipment tab.

Warning 1: this is only compile & run tested. Most of the changes were done blind. It *will* crash and burn in horrific ways. Especially the planner is completely untested and I'm sure it will explode at the slightest touch.

Warning 2: C programmers will cringe when they see that the `cylinder_t` structure is passed on the stack in the table functions. As explained in the commit message, this is due to how the table functions currently work. I plan to rework them in the future and then address the problem. For now, I will just note that cylinder_t is 64 bytes, that makes 8 long words on 64-bit architectures. Passing it on the stack is probably not even that horrible compared to passing around pointers.

Note: this now also theoretically supports 0-cylinder dives [e.g. free dives]. At least you can manually add a dive without a cylinder.

### Changes made:

1) Allocate all cylinder arrays dynamically
2) Remove MAX_CYLINDERS limit.
3) Some random cleanups.

### Release note:

Done.
### Documentation change:

Have to check.
### Mentions:

@dirkhh @atdotde


bstoeger:
> 1 for Multiplication result converted to larger type

If that multiplication overflows signed int we have a more serious problem (viz. a dive with tens of millions of cylinders).

dirkhh (approved):
Wow.
Thank you for a very well structured series.
It's very, very invasive. It cleans up the code and the new code seems overall better (if not always easier to understand).
I tried to review all of it, but the mega commit towards the end made my eyes cross.
Passing 64bytes on the stack - we do worse in the CVS style parsing functions.
The FIXME should be addressed by simply removing all of that code.
The multiplication overflow is one of those cases where static analysis is at it's limit. I know that there are comments we can add to ensure that LGTM is happy again. I can look that up... I agree that this should never overflow.
Of course we could simply call it an error if a dive has more than a thousand cylinders which would also work :-)
I'm approving this change - but I have NOT done any testing. Especially the use in the planner that @bstoeger mentioned deserves to be tested @atdotde

bstoeger:
> I know that there are comments we can add to ensure that LGTM is happy again.

I think we could simply cast the argument to size_t before multiplication.

My strategy with FIXME so far was to rename it to TODO.

bstoeger:
> I tried to review all of it, but the mega commit towards the end made my eyes cross.

Yeah, this could have been split in two approximately equal-sized parts by first changing everything to use the `get_cylinder()` accessor function and then converting to the table. Hindsight, etc. :-/

core/equipment.h
```diff
@@ -33,6 +33,17 @@ typedef struct
 	bool bestmix_he;
 } cylinder_t;
 
+/* Table of cylinders. Attention: this stores cylinders,
+ * *not* pointers to cylinders. This has two crucial consequences:
+ * 1) Pointers to cylinders are not stable. They may be
+ *    invalidated if the table is reallocated.
```
atdotde, line 39: This comment makes me worry: As far as I can see, this is currently not thread save, there are no locks. Even if it does not bite us now, it could at any point in the future introduce very subtle bugs. With this, it seems we should get rid of all cylinder_t pointers in the rest of the code (in favour of passing around indices) and also the construct get_cylinder(index)->member looks dangerous. Furthermore, if the table can grow and shrink (can it actually shrink? maybe that's not a good idea) at random times, keeping nr around for any delayed use can lead to problems. This clearly is not a full review, just a thing that popped to my mind when reading the code.

core/planner.c
```diff
@@ -764,23 +767,23 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 	vpmb_start_gradient(ds);
 	if (decoMode() == RECREATIONAL) {
 		bool safety_stop = prefs.safetystop && max_depth >= 10000;
-		track_ascent_gas(depth, &dive->cylinder[current_cylinder], avg_depth, bottom_time, safety_stop, divemode);
+		track_ascent_gas(depth, get_cylinder(dive, current_cylinder), avg_depth, bottom_time, safety_stop, divemode);
```
atdotde: Let's make this take an index rather than a pointer.

atdotde: Or even better, make current cylinder a copied struct cylinder_t rather than a pointer.

core/plannernotes.c
```diff
@@ -446,16 +446,14 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
 	free(temp);
 
 	/* Print gas consumption: This loop covers all cylinders */
-	for (int gasidx = 0; gasidx < MAX_CYLINDERS; gasidx++) {
+	for (int gasidx = 0; gasidx < dive->cylinders.nr; gasidx++) {
```
atdotde, line 464: Maybe introduce a FOREACH_CYLINDER() ?

bstoeger:
> This comment makes me worry: As far as I can see, this is currently not thread save, there are no locks.

Oh, but the code has never been thread safe and I don't see how it ever could be. All user interactions are suspended while we're calculating something in the UI thread. Which non-UI thread do you think can grow or shrink the cylinder table?

The comment above only means that you must not hold a reference to a cylinder across operations that might grow / shrink the table. That is very much the idiomatic behavior for vector-like containers.

And of course the table can shrink. Namely when the user deletes a cylinder. But the necessary remapping of cylinder-ids is explicitly handled in the CylinderModel code. I hope that I didn't butcher that too much. :)

core/planner.c
bstoeger: I don't get it. `track_ascent_gas()` does not store a reference to the cylinder, does it? Moreover it seems to modify the cylinder. So you *have* to pass in a pointer to the cylinder. Using a copy would defeat the purpose or am I missing something?

What I can imagine has to be done is add checks that current_cylinder is still valid after the user deleted cylinders.

atdotde:
The planner runs in a background thread when we compute plan variations.

atdotde:
But maybe this is ok as it uses its own dive and thus its own table.

atdotde:
I guess, at least in the planner, by making current_cylinder an actual copy rather than in index that requires going to the table for every access would make me sleep much better.

bstoeger:
To be honest - looking at the `lock_planner` calls, I have my doubts that this is all OK. The planner seems to access displayed_dive, which is not under lock as far as I can see. Can't the user modify the cylinders while the planner is running?

But this is not a problem of this patch - it might only make it more visible.

bstoeger:
> I guess, at least in the planner, by making current_cylinder an actual copy rather than in index that requires going to the table for every access would make me sleep much better.

Without having looked closer - the correct fix appears to me to move all actions that modify `displayed_dive` into a planner-locking critical section. But I would have to take a much closer look.

bstoeger:
> But maybe this is ok as it uses its own dive and thus its own table.

That is the crucial point - what we have to fix is the few remaining accesses to displayed_dive in the planner. Changing from pointer to index will not help with thread-safety at all. It will only make it fail in more obscure ways.

dirkhh:
So where are we with this branch?

bstoeger:
I would postpone this until the thread-issues of the planner are fixed (i.e. #2211 and #2213). And then it still needs a lot of testing.

bstoeger:
Rebased on current master. Only compile tested.

bstoeger:
Rebased on master, "fixed" new LGTM alerts.
Removed [WIP] tag - this now needs testing, ideally from someone using multiple cylinders and the planner. @atdotde, @mturkia?

bstoeger:
Travis-mac is currently acting up for all PRs. I think that build can be ignored.

core/planner.c
```diff
@@ -1099,14 +1101,12 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
...
-				// If no empty cylinder is found, keep using last deco gas
-				current_cylinder = i;
-				dive->cylinder[i].cylinder_use = NOT_USED;
-				break;
-			}
+		// Switch to an empty air cylinder for breathing air at the surface
+		// If no empty cylinder is found, keep using last deco gas
+		cylinder_t cyl = { 0 };
+		cyl.cylinder_use = NOT_USED;
+		add_to_cylinder_table(&dive->cylinders, dive->cylinders.nr, cyl);
```
atdotde, line 1108: maybe you can make this return a pointer to the newly added cylinder (so you can combine this with the next line)?

atdotde (approved):
Given how huge this is, I mainly looked at the planner as far as it is concerned by this. I like this. Most of the calls there are really for the gas mix of the current cylinder (which is only used for read access). Maybe this could be simplified by storing the gas mix. But I can add this later once this is merged. You have my go.

core/planner.c
bstoeger: That wouldn't make much sense here as `current_cylinder` is an integer not a pointer.

Anyhow, the whole table-functionality should be reworked and returning an iterator to the `add_to_*_table()` functions is a good idea. Let's keep that for a future PR.

bstoeger:
Time to revive this. @dirkhh: do you want to cut a release before merging this - it certainly will break at some places?

dirkhh:
I'm constantly updating Android these days, but we have nothing interesting to release for desktop, except a few bug fixes in libdivecomputer. So right now I didn't have a release planned.
Go ahead and merge when ready.

bstoeger:
Not ready - conflicts with the recent changes. Will look into it.

bstoeger:
Fixed a bug, and rebased on master, but multi-cylinder handling still broken (last segment not shown) - will continue debugging.

dirkhh:
So this is still WIP, correct? The last comment seems to imply that there is an issue with multi cylinder handling.

bstoeger:
I fixed all the problems that I found. Nevertheless, there are a few UI fixes in master. That makes me wonder if this risky commit should go in?

dirkhh:
> I fixed all the problems that I found. Nevertheless, there are a few UI fixes in master. That makes me wonder if this risky commit should go in?

I am not planning an imminent Subsurface release (focused on -mobile right now), and typically bugs are more likely to be addressed after things got merged (we get a LOT more testing of master than of PR code). So I'll do another code review and if I find nothing that bugs me, I'll merge.

If you are aware of specific areas that people should test, sending an email to the developer mailing list might be useful

core/profile.c
```diff
@@ -808,12 +809,12 @@ static void calculate_sac(struct dive *dive, struct divecomputer *dc, struct plo
 	const struct event *ev = NULL;
 	bool *gases, *gases_scratch;
 
-	gases = malloc(MAX_CYLINDERS * sizeof(*gases));
-	memset(gases, 0, MAX_CYLINDERS * sizeof(*gases));
+	gases = malloc(pi->nr_cylinders * sizeof(*gases));
+	memset(gases, 0, pi->nr_cylinders * sizeof(*gases));
```
torvalds: I'm not a huge fan of this commit, simply because it's so *huge*. But the only real comment I have about it is that I wish you didn't do that blind malloc() conversion.

This should probably have been a calloc() even originally, but at least the size used to be a constant. Now it isn't, and it most definitely should be a calloc().

So those two lines should be

     gases = calloc(pi->nr_cylinders, sizeof(*gases));

instead, which initializes things to zero, and will probably also avoid the warning about possible multiplication overflow.

core/dive.c
```diff
@@ -704,54 +704,71 @@ void fixup_dc_duration(struct divecomputer *dc)
...
+		}
 	}
 
 	/* We know about the explicit first cylinder (or first) */
-	mask |= 1 << explicit_first_cylinder(dive, dc);
+	idx = explicit_first_cylinder(dive, dc);
+	if (idx > 0 && used_and_unknown[idx]) {
+		used_and_unknown[idx] = false;
+		num--;
+	}
```
torvalds: This seems to change behavior. Why the "if (idx > 0" test?

It very much intentionally knew that 0 is a valid value for the case of not explicitly having a first cylinder - then cylinder zero _is_ the first cylinder. 

So that discard of 0 seems just wrong, and is an unexplained change from what the old code did.

core/dive.c
```diff
@@ -761,7 +778,8 @@ void per_cylinder_mean_depth(const struct dive *dive, struct divecomputer *dc, i
 	uint32_t lasttime = 0;
 	int lastdepth = 0;
 	int idx = 0;
-	unsigned int used_mask, known_mask;
+	bool *used_cylinders;
+	int num_used_cylinders;
```
torvalds, line 754: I'm not entirely convinced about the whole "use bool array instead of bits". It's kind of sad. We really don't care about more than 32 cylinders.

But hey, I guess I don't care that deeply.

torvalds (commented):
It looks mostly good to me. One huge commit that I'm not entirely happy about, a couple of small nits I found, and a lot of code that _looked_ like it didn't change behavior but I might have missed something.

I'm ok with this, except for the explicit_first_cylinder() mis-conversion. The bitmap thing I can live with, even if it's not wonderful.

core/profile.c
bstoeger: Thanks for review - I will change that to calloc. I'm not familiar with C-style memory management.

core/dive.c
bstoeger: Good catch - the condition is wrong.

I reckon that I was thinking about the no-cylinders case and this was intended to read `idx >= 0`.

Nevertheless, neither does `explicit_first_cylinder` return `< 0` in the no-cylinder case, nor can we reach that place, since `per_cylinder_mean_depth()` returns early if there are no cylinders.

core/dive.c
dirkhh: On that specific note... the world record 330m dive that took 15 hours was done with 9 tanks. Even if you want to plan and have the real dive in the same dive, 32bits allows for 16 tanks... I just don't think this is something that's worth changing.

bstoeger: On the other hand - who knows what strange use-case the planner comes up with. If we remove the hard-limit, let's remove it completely. Otherwise we'd have to add sanity checks and then how do we treat the failure case, etc.

dirkhh: now THAT is the winning argument. Adding all the checks for numCylinders < whatever limit is just silly.

PR #2211, opened 2019-08-05
More include detangling

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

More splitting out of declarations from dive.h
As usual: certainly not perfect, but hopefully a step in the right direction.

### Changes made:

1) Split out error reporting functions.
2) Split out file related functions.
3) Split out planner declarations.

### Mentions:

@atdotde: This mostly concerns the planner.

### DIFF
--- a/core/android.cpp
+++ b/core/android.cpp
@@ -2,6 +2,7 @@
 /* implements Android specific functions */
 #include "dive.h"
 #include "display.h"
+#include "file.h"
 #include "qthelper.h"
 #include <string.h>
 #include <sys/types.h>
@@ -10,6 +11,7 @@
 #include <libusb.h>
 #include <errno.h>
 #include <unistd.h>
+#include <zip.h>
 
 #include <QtAndroidExtras/QtAndroidExtras>
 #include <QtAndroidExtras/QAndroidJniObject>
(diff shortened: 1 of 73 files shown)


bstoeger:
PS: This will fail on the platform specific files.

bstoeger:
@atdotde: if you find time, please have a look at commit 5836751d8a5ffb01eb280f6dae1e58e09218844e. I have *no* idea what this is doing, but it shouldn't change the results. It's all a bit ominous and I wonder if that regression code can be moved out of the function.

bstoeger:
Ok - different approach: move the regression manipulation out of the tissue_tolerance_calc() function. Very cryptic.

bstoeger:
This now removes all `displayed_dive` accesses from planner.c and deco.c and I don't see any obvious indirect accesses. Therefore I hope this removes the thread-safety doubts about  #2208.

atdotde:
I did not have time to look at this in detail, yet. But the approach is good: The planner just needs some dive to operate on (replacing the use of displayed_dive), it can be passed to it in the call.

The regression data (that is/was/should be used to compute effective gradient factors for VPM-B schedules) is simply deco state and should rather go to struct deco_state rather than getting its own struct.

I hope I find some time later today to look into this more closely.

dirkhh:
Closing this after merging #2213

PR #2214, opened 2019-08-07
Limit gas compressibility argument range to halfway sane values

torvalds:
The curve fitting for our gas compressibility was only done in the sane
range of 0-500 bar, which is what a scuba cylinder can reasonably be
expected to perhaps have.

But the planner ends up happily using negative cylinder pressures when
you run out of gas, and then the compressibility gives nonsensical
results.

That's clearly a planner bug, but the nonsensical gas compressibility
values made it harder to see what could be wrong.

So we just clamp the inpot range to the range we have verified against
experimental data.  If you try to get compressibility for negative
pressures, you get the compressibility for an ideal and imaginary gas.
And if you try to get compressibility for pressures over 500 bar, we'll
just assume that it's 500 bar.

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/gas-model.c
+++ b/core/gas-model.c
@@ -47,6 +47,14 @@ double gas_compressibility_factor(struct gasmix gas, double bar)
 	double x1, x2, x3;
 	double Z;
 
+	/*
+	 * The curve fitting range is only [0,500] bar.
+	 * Anything else is way out of range for cylinder
+	 * pressures.
+	 */
+	if (bar < 0) bar = 0;
+	if (bar > 500) bar = 500;
+
 	o2 = get_o2(gas);
 	he = get_he(gas);
 


torvalds:
This limits the damage from

https://github.com/Subsurface-divelog/subsurface/issues/2212

in case anybody cares about the history of this fix.

bstoeger (approved):
Looks reasonable to me.

@atdotde says that negative pressures are actually useful (the way I understand: -400 bar meaning I need two more cylinders). I don't plan on ever diving either in deco or with multiple cylinders, so I can't judge this.

torvalds:
> @atdotde says that negative pressures are actually useful (the way I understand: -400 bar meaning I need two more cylinders).

Hmm. Maybe as a very rough approximation that does end up being the case. But exactly because of gas compressibility, you can't really do math that way (ie "-400 bar" doesn't really mean "two extra cylinders at 200 bar would work").

But I guess as a starting point for "let's try what the planner says if I add more cylinders" thing it may work. I've never used the planner either - I'm deco certified, but if I never do another planned dive in my life I'm probably happy.

bstoeger:
> Hmm. Maybe as a very rough approximation that does end up being the case. But exactly because of gas compressibility, you can't really do math that way (ie "-400 bar" doesn't really mean "two extra cylinders at 200 bar would work").

True - for 200 bar I get a 1.036 factor, which is higher than I expected, but still a good estimate. With 400 bar I get already 1.2, which starts to be highly non-linear.

In any case, the patch is obviously correct, as for negative pressures we simple should assume an ideal gas (could return -1 right away). Let's commit.

PR #2216, opened 2019-08-07
Cleanup: replace macro by inline function in gas-model.c

bstoeger:
Replace a macro calculating a degree-three polynomial by an
inline function.

Moreover, calculate the powers 1, 2 and 3 of the pressure inside
the function. The compiler will be smart enough to optimize this
to the same code. The only important thing is to write "x*x*x*coeff"
instead of "coeff*x*x*x". The compiler can't optimize the latter
because ... wonderful floating point semantics.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A tiny code-cosmetics thing: turn macro into inline function don't do the compilers job. I verified that both versions produce the same assembly on compiler explorer. The crucial point is just to write x*x*x*coeff instead of coeff*x*x*x. The joys of floating point calculations!

### Release note:

No.
### Documentation change:

No.
### Mentions:

@atdotde

### DIFF
--- a/core/gas-model.c
+++ b/core/gas-model.c
@@ -6,7 +6,10 @@
 #include "dive.h"
 
 /* "Virial minus one" - the virial cubic form without the initial 1.0 */
-#define virial_m1(C, x1, x2, x3) (C[0]*x1+C[1]*x2+C[2]*x3)
+static double virial_m1(const double coeff[], double x)
+{
+	return x*coeff[0] + x*x*coeff[1] + x*x*x*coeff[2];
+}
 
 /*
  * Z = pV/nRT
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
This looks good to me. I don't even know why we got a PUSH Travis build. I turned those off a long time ago. The PR Travis build completed without error...

dirkhh:
@atdotde would you please review?

atdotde:
I like it.

atdotde:
Restarted failed Travis builds.

dirkhh:
OK, looks like this is ready to merge...

henrik242:
> The crucial point is just to write xxxcoeff instead of coeffxxx. 

Should that be added as a comment?

Issue #2199, opened 2019-07-28
Surface Mobile - App Crashes downloading dives


SousaMartins:
### Describe the issue:

- [ x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

The surface mobile version 2.2.0(4.9.0.30) crashes after finish download the dives from a Shearwater perdix computer. It worked fine with the previous version

### Operating system:
Android 9

Samsung Galaxy J7

### Subsurface version:

 surface mobile version 2.2.0(4.9.0.30)

It does not happened with the previous version

I am using official release

### Steps to reproduce:

1) The Bluetooth connection is establish
2) It starts download the dives data
3) The application crashes

### Current behavior:

crashes the application

dirkhh:
are you able to connect to the device with `adb` (if you don't know what that means, that's ok, I'm simply asking on the off chance that you are somewhat familiar with debugging issues on an Android device).
We have a small number of users who report similar problems, but none of the developers experience this issue which makes it a lot harder to track down. Are you on our beta channel (i.e., do you get the more frequent updates)? I'll make a newer version available there which fixes a different potential crash, just on the off chance that that's the one you see (and you just see it with a different path leading you there)

SousaMartins:
Hi thank you for your email

 

Unfortunately I am not familiar with  debugging issues on an Android device.

I am sending you two log files I have downloaded from the Android device. I hope this will help to identify and solve the problem

 

Thanks once again

SousaMartins:
No I am not on the beta channel

dirkhh:
Since this fixed the problem for two other users, I just pushed the new build to production - you should get an update to Subsurface, soon. I would appreciate if you could report back here if that does (or, worst case, does not) fix your problem.
Thanks

SousaMartins:
Sorry the problem is not fixed.

 

I am sending you the new log file.

 

Thanks

SousaMartins:
[subsurface.log](https://github.com/Subsurface-divelog/subsurface/files/3451333/subsurface.log)

bstoeger:
Looks like a duplicate of #2167. :(

Issue #2167, opened 2019-07-08
mobile: SIGSEGV on BT download from DC on Android


janmulder:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Using the current master, and a self build (arm) Android APK, I have a reproducible SIGSEGV in `DiveImportedModel::repopulate`. Interestingly, I cannot reproduce this crash using mobile-on-desktop (which complicates debugging).

To reproduce: download from a BT/BLE DC. And look at the logcat, and see it crash just after the finish of a successful download thread.

### Operating system:
OS: Android 9, LineageOS 16.0 

### Subsurface version:
What version of Subsurface are you running? current master (4.8.4.681)
Does the same happen on another Subsurface version? not reproducible on 4.8.4.577
Are you using official release, test build, or compiled yourself? self compiled. 

### Steps to reproduce:
See above.

bstoeger:
Does it crash directly in `DiveImportedModel::repopulate()` or in one of the two function calls?

Because I don't see how this could crash in this function unless either the `this` pointer is invalid or there is some preceding memory corruption (which would explain why it doesn't happen on mobile-on-desktop).

Could you perhaps try running mobile-on-desktop with valgrind and see if there is some memory corruption reports? Thanks.

bstoeger:
Well, for me mobile-on-desktop crashes right at startup, so there seems to be some fundamental issue. :)

bstoeger:
Never mind - that was an unrelated issue (see #2168).

bstoeger:
@dirkhh: shouldn't we try to fix this before 4.9? Sounds like a show-stopper to me.

janmulder:
In the meantime, I did valgrind mobile-on-desktop, and saw nothing suspicious there (with this scenario). Notice, not the exact same SHA this issue is written against (but a little newer). Unfortunately, I do not have  access to my DCs (everything is seriously packed for a cave diving exploration starting very soon), so I cannot verify things right now. My understanding is that 4.9 desktop is already tagged, and build for release (so no more changes, unless ...). Further, 4.9 desktop is not per definition the release of 2.2.0 mobile.

Also, I did numerous BT/BLE downloads over the past couple of days, mainly using arm64 builds. And this issue is not there. I'm sure I found it on the arm build. So, I'm not sure it can be reproduced any more.

bstoeger:
[...]

PR #2219, opened 2019-08-08
DO NOT MERGE Android crash fixes

dirkhh:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Lots of crashes with the latest Android release, tons of unhappy users.
Looking through the stack traces I found the three most common crash sites and added code that should prevent us from crashing even if the tables are NULL.

### Related issues:

#2167 
#2199

### DIFF
--- a/qt-models/diveimportedmodel.cpp
+++ b/qt-models/diveimportedmodel.cpp
@@ -136,10 +136,14 @@ void DiveImportedModel::repopulate(dive_table_t *table, struct dive_site_table *
 	diveTable = table;
 	sitesTable = sites;
 	firstIndex = 0;
-	lastIndex = diveTable->nr - 1;
-	checkStates.resize(diveTable->nr);
-	std::fill(checkStates.begin(), checkStates.end(), true);
-
+	if (diveTable) {
+		lastIndex = diveTable->nr - 1;
+		checkStates.resize(diveTable->nr);
+		std::fill(checkStates.begin(), checkStates.end(), true);
+	} else {
+		lastIndex = -1;
+		checkStates.resize(0);
+	}
 	endResetModel();
 }
 
--- a/qt-models/divelocationmodel.cpp
+++ b/qt-models/divelocationmodel.cpp
@@ -224,10 +224,12 @@ DiveSiteSortedModel::DiveSiteSortedModel()
 QStringList DiveSiteSortedModel::allSiteNames() const
 {
 	QStringList locationNames;
-	int num = rowCount();
-	for (int i = 0; i < num; i++) {
-		int idx = mapToSource(index(i, 0)).row();
-		locationNames << QString(dive_site_table.dive_sites[idx]->name);
+	if (dive_site_table.dive_sites) {
+		int num = rowCount();
+		for (int i = 0; i < num; i++) {
+			int idx = mapToSource(index(i, 0)).row();
+			locationNames << QString(dive_site_table.dive_sites[idx]->name);
+		}
 	}
 	return locationNames;
 }
--- a/core/file.c
+++ b/core/file.c
@@ -263,7 +263,7 @@ int check_git_sha(const char *filename, struct git_repository **git_p, const cha
 	struct git_repository *git;
 	const char *branch = NULL;
 
-	char *current_sha = strdup(saved_git_id);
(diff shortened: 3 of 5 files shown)


bstoeger:
Good that you could localize the crashes. Could you post the backtraces - I'd really like to understand the root cause for those. Though I expect it to be somewhere in Javascript, I.e. hard to figure out.

qt-models/divelocationmodel.cpp
```diff
@@ -224,10 +224,12 @@ DiveSiteSortedModel::DiveSiteSortedModel()
 QStringList DiveSiteSortedModel::allSiteNames() const
 {
 	QStringList locationNames;
-	int num = rowCount();
-	for (int i = 0; i < num; i++) {
-		int idx = mapToSource(index(i, 0)).row();
-		locationNames << QString(dive_site_table.dive_sites[idx]->name);
+	if (dive_site_table.dive_sites) {
```
bstoeger, line 227: This one here looks wrong. It still can crash if rowCount and nr don't match. You want to loop over the correct number. But it would be important to find out why model and core are out if sync.

qt-models/diveimportedmodel.cpp
```diff
@@ -136,10 +136,14 @@ void DiveImportedModel::repopulate(dive_table_t *table, struct dive_site_table *
 	diveTable = table;
 	sitesTable = sites;
 	firstIndex = 0;
-	lastIndex = diveTable->nr - 1;
-	checkStates.resize(diveTable->nr);
-	std::fill(checkStates.begin(), checkStates.end(), true);
-
+	if (diveTable) {
```
bstoeger, line 139: How can table be null here? Even if this doesnt crash, import still will not work, isnt it?

bstoeger (changes requested):

dirkhh:
> Good that you could localize the crashes. Could you post the backtraces - I'd really like to understand the root cause for those. Though I expect it to be somewhere in Javascript, I.e. hard to figure out.

That gives me a hopefully brilliant idea...

qt-models/divelocationmodel.cpp
dirkhh: I agree with both points. 

The new version hasn't been out long enough to confidently claim that this fixed #2221, but at least I haven't seen it recur with the new version.

qt-models/diveimportedmodel.cpp
dirkhh: I'm frankly as perplexed as you are... but this was the only thing that I thought could be crashing in that function with the stack trace in #2221

bstoeger:
Perhaps retract the mobile version? There seems no point if it is unusable. It's a problem that I don't use mobile, but the undo changes obviously mess with the core a lot. :(

dirkhh:
> Perhaps retract the mobile version? There seems no point if it is unusable. It's a problem that I don't use mobile, but the undo changes obviously mess with the core a lot. :(

I don't think it's unusable. I use it successfully, no crashes. So do many others.
I'm trying to stem some of the more common crashes, that's all.

qt-models/divelocationmodel.cpp
bstoeger: Looking at this again - this must be bogus.

`dive_site_table` is a global object. If for some reason `dive_site_table.nr` is > 0 and `dive_site_table.dive_sites` is null, the core structures are corrupted. The application will crash on the next attempt to access a dive sites. If not here - elsewhere.

What could happen though is that the model was not updated properly and that `mapToSource` returns an invalid index. Therefore I suggest adding a
```
		if (idx < 0 || idx >= dive_site_table.nr) {
			log_some_warning_visible_on_mobile();
			continue;
		}

bstoeger: E.g. `setupDivesite()` in qmlmanager.c does not appear to reset the DiveSiteSortedModel - which will lead to invalid indexes on mapping and could explain that crash. That should be reproducible.

dirkhh:
The two string fixes are a separate PR now #2232 
The other three are all questionable and need better fixes elsewhere

PR #2227, opened 2019-08-10
Fix typo in CNS calculation

atdotde:
There was a sign error in the formula comuting the
interpolation for the CNS table.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@willemferguson please confirm.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -231,6 +231,12 @@ static double calculate_cns_dive(const struct dive *dive)
 		}
 		po2i = (po2i + po2f) / 2;	// po2i now holds the mean po2 of initial and final po2 values of segment.
 		/* Don't increase CNS when po2 below 500 matm */
+		if (po2i > 1600) {
+			if (po2i > 1650)
+				return 0;
+			else
+				po2i = 1600;
+		}
 		if (po2i <= 500)
 			continue;
 		/* Find the table-row for calculating the maximum exposure at this PO2 */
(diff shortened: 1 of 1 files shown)


dirkhh:
Admittedly, that's an improvement... but based on the discussion on the mailing list, I'm not sure this is the way we should go... We either should refuse to calculate CNS, or we should try something along the lines of the approximations suggested by you or actually maybe the ones Linus came up with.

willemferguson:
Robert, my suggestion is:
1) use the 4th order polynomial to calculate the CNS toxicity. The relevant root for that equation is close to 1648.
2) expand the dive structure with a po2GT1640 boolean.
3) if any of the po2 values are greater than 1640, make the po2 for calculation 1640
4) In the planner if po2GT1640 then print a warning with the calculated CNS
5) Once the po2GT1640 value exists I will then correct the CNS value In the Information tab in a similar way.

atdotde:
I need to look into this but today I am traveling and essentially offline. I want to understand how much better the fourth order approximation is and if it is worthwhile doing it given that the original table data is probably not that precise. But I would definitely approximate the log of the table rather than the data directly to avoid zero crossings.  Please give me some time.

atdotde:
In the meantime, maybe somebody could trace back the original source of the table data to figure out how much massaging has already been done to the data and how much effort we should spend on precisely getting it right.

willemferguson:
The table comes from the original Calk and Lambertsen publication. Clearly there is some interpolation already at the 0.05 bar intervals. Erik Baker correctly used the information in his table. I suspect the polynomial is a pretty good fit to the data.

dirkhh:
closing since #2246 was merged

Issue #2221, opened 2019-08-09
Android crash in DiveImportedModel::repopulate


dirkhh:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Backtrace:
```
backtrace:
  #00  pc 00000000001798aa  /data/app/org.subsurfacedivelog.mobile-Ac2S9LYxrLIi7igc4P96lQ==/lib/arm/libsubsurface-mobile.so (DiveImportedModel::repopulate(dive_table*, dive_site_table*)+33)
  #01  pc 0000000000183989  /data/app/org.subsurfacedivelog.mobile-Ac2S9LYxrLIi7igc4P96lQ==/lib/arm/libsubsurface-mobile.so (DiveImportedModel::qt_metacall(QMetaObject::Call, int, void**)+52)
  #02  pc 00000000000f673b  /data/app/org.subsurfacedivelog.mobile-Ac2S9LYxrLIi7igc4P96lQ==/lib/arm/libQt5Qml.so
```

alternative stack trace from older Android version 6.0/7.0
```
  #00  pc 00000000001798aa  /data/app/org.subsurfacedivelog.mobile-1/lib/arm/libsubsurface-mobile.so (_ZN17DiveImportedModel10repopulateEP10dive_tableP15dive_site_table+33)
  #01  pc 0000000000183989  /data/app/org.subsurfacedivelog.mobile-1/lib/arm/libsubsurface-mobile.so (_ZN17DiveImportedModel11qt_metacallEN11QMetaObject4CallEiPPv+52)
  #02  pc 00000000000f673b  /data/app/org.subsurfacedivelog.mobile-1/lib/arm/libQt5Qml.so
```
This one made me think that it was indeed the `dive_table` that was the issue

### Operating system:

Most of the crash reports (more than a hundred) are on Android 8.1 and 9
The older stack trace is from 6.0 and 7.0

### Subsurface version:

4.9.0.35

### Mentions:

#2219 was supposed to fix this one - and I haven't seen this crash reported in the latest build based on #2219

bstoeger:
Oh my ... These backtraces aren't very helpful. :(
We seem to have a very fundamental problem, but how to debug this? I reckon this can't be reproduced on mobile-on-desktop...?

bstoeger:
The fix makes no sense to me. The call in `mobile-widgets/qml/DownloadFromDiveComputer.qml` is
```
[...]

PR #2229, opened 2019-08-10
[RFC] Alternative fix to #2221

bstoeger:
There are reported crashes on Android that suggest a null
"tables" attribute in DownloadDCThread. This should never
happen, as the table() function connected to this attribute
returns the address of a subobject. Thus, even if the original
DownloadThread is null, this would not return a null pointer
(the subobject is not at address 0).

Catch these null-object accesses and write a warning message
to the console. Hopefully, this will help in localizing the
problem.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

@dirkhh: Could you please try this alternative fix to #2221? I'm currently travelling and have no device to test on. :( This QML stuff is mind-boggling.

### DIFF
--- a/mobile-widgets/qml/DownloadFromDiveComputer.qml
+++ b/mobile-widgets/qml/DownloadFromDiveComputer.qml
@@ -28,6 +28,10 @@ Kirigami.Page {
 		id: downloadThread
 
 		onFinished : {
+			if (!table || !sites) {
+				console.warn("DCDownloadThread::onFinished(): table or sites is null!")
+				return
+			}
 			importModel.repopulate(table, sites)
 			progressBar.visible = false
 			if (dcImportModel.rowCount() > 0) {


dirkhh:
The problem? I cannot reproduce these crashes, either. All I can do is post new versions for our 6k+ active users to try and see what happens...

bstoeger:
Testing whether download is still possible with this patch would be helpful. If that doesn't fix #2221, I really don't know how to proceed...

dirkhh:
That I can do.

dirkhh:
OK, the iOS build failure is that idiotic `FILE` problem. I can ignore this.
I'll test this branch as it is to see if there's any impact on the ability to download from dive computer on Android.
Should this go on top of my bogus fixes? Or should we abandon those that are completely pointless based on your analysis? I'm trying to figure out how to continue with this in the most productive manner.
(still not a single crash report for the latest binary - but sadly they don't show up reliably... sometimes I get none for days and then a whole bunch, many of them older... but just a few minutes ago I got half a dozen crash reports from within the last hour, but all on older App versions)

bstoeger:
I would definitely apply the string functions. That can't hurt.

I would wait with the others to see if we manage to fix them in a more controlled way. At the risk of angering some users.

dirkhh:
OK I can do that. Seems reasonable

dirkhh:
Trying to download from my EON Steel crashes the phone hard. But I don't think that has anything to do with your changes. There's once again something crazy going on with the BLE stack on the EON Steel. I made the mistake of pairing it with a different device (my iPad) to test things there, and every time you un-pair and try to re-pair things go catastrophic.

I successfully downloaded one dive from the Shearwater Teric. So while that is only a tiny bit of testing, I'd call this successful.

dirkhh (approved):
Let's merge this

Issue #2222, opened 2019-08-09
Android crash in DiveSiteSortedModel::allSiteNames


dirkhh:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Backtrace:
```
  #00  pc 000000000022d2a8  /data/app/org.subsurfacedivelog.mobile-2DwhJ8aKhN2G5g4SWRa4rA==/lib/arm64/libsubsurface-mobile.so (DiveSiteSortedModel::allSiteNames() const+204)
  #01  pc 000000000020686c  /data/app/org.subsurfacedivelog.mobile-2DwhJ8aKhN2G5g4SWRa4rA==/lib/arm64/libsubsurface-mobile.so
  #02  pc 00000000002086a4  /data/app/org.subsurfacedivelog.mobile-2DwhJ8aKhN2G5g4SWRa4rA==/lib/arm64/libsubsurface-mobile.so (QMLManager::qt_metacall(QMetaObject::Call, int, void**)+88)
  #03  pc 0000000000259d28  /data/app/org.subsurfacedivelog.mobile-2DwhJ8aKhN2G5g4SWRa4rA==/lib/arm64/libQt5Qml.so
```

### Operating system:

Android 9

### Subsurface version:

4.9.0.35

### Mentions:

#2219 was supposed to fix this one - haven't seen it recur in the new version

bstoeger:
There is now code that prevents us from accessing invalid indexes. If it still crashes, this means that the internal model state got messed up. This happened before with Qt proxy-models, so it's not completely unlikely.

Leaving open.

PR #2230, opened 2019-08-10
[RFC] Attempt at fixing #2222

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is an attempt at fixing #2222. Problem: neither was I able to reproduce the crash, nor do I understand why the problem would become active with 4.9. It should have given problem in previous versions. So I have my doubts that this actually fixes the problem, but I'm grasping straws here.

### Changes made:

1) Don't crash on invalid mappings of the DiveSiteSortedModel.
2) Try to keep DiveSiteSortedModel in sync.

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -333,13 +333,19 @@ void QMLManager::openLocalThenRemote(QString url)
 	updateAllGlobalLists();
 }
 
+void QMLManager::updateSiteList()
+{
+	LocationInformationModel::instance()->update();
+	emit locationListChanged();
+}
+
 void QMLManager::updateAllGlobalLists()
 {
 	buddyModel.updateModel(); emit buddyListChanged();
 	suitModel.updateModel(); emit suitListChanged();
 	divemasterModel.updateModel(); emit divemasterListChanged();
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
These both look obviously reasonable, so much so that I'd be tempted to just merge them.
Any concerns with that?

bstoeger:
I could still set the dive-site on mobile-on-desktop with this applied, but I lost all confidence with respect to the mobile app.

bstoeger:
We should wait for travis - who knows if fprintf is declared everywhere, etc.

dirkhh:
> I could still set the dive-site on mobile-on-desktop with this applied,

Sorry. can't parse that sentence.

>  but I lost all confidence with respect to the mobile app.

And yeah, I have serious confidence issues with the mobile app. Yet it works well for 95+% of its users, so abandoning it seems wrong, too

I'll definitely wait for Travis

bstoeger:
With this patch applied I can still set the dive-site using mobile-on-desktop.

I was talking about the confidence in me understanding what is going on.

dirkhh:
Travis just shows the known iOS issue that's fixed in master

PR #2231, opened 2019-08-10
Mobile: fix editing of weight system

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Removal of MAX_WEIGHTSYSTEM made it impossible to edit the weight on mobile for dives that had no weightsystem set (e.g. newly created dives). This PR fixes this.

### Changes made:

1) Enable editing of weight for 0-weightsystem dives.
2) Add a new weightsystem for 0-weightsystem dives.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Mobile: allow editing weights for dives that have no weight
 - Desktop: fix display of trip time
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -254,7 +254,7 @@ QStringList DiveObjectHelper::weights() const
 
 bool DiveObjectHelper::singleWeight() const
 {
-	return m_dive->weightsystems.nr == 1;
+	return m_dive->weightsystems.nr <= 1;
 }
 
 QString DiveObjectHelper::weight(int idx) const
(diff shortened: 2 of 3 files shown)


dirkhh (approved):
Thanks for fixing this one. Looks good

bstoeger:
Should we use something different than the empty string as description?

dirkhh:
> Should we use something different than the empty string as description?

we could do something like `tr("weight")`

dirkhh:
Does that look right, @bstoeger ?
I'll merge locally if you approve (and if Travis likes it)

dirkhh:
does this look better, @bstoeger ?

bstoeger:
Yes, that looks reasonable to me.
Before commit, perhaps change to `qUtf8Printable` as the documentation sounds to me as if that was more secure: https://doc.qt.io/qt-5/qtglobal.html#qUtf8Printable
But I'm not sure if I read the documentation right.

dirkhh:
We use `qPrintable` everywhere because we set Subsurface up to always use Utf8
```
void init_qt_late()
{
	QApplication *application = qApp;
	// tell Qt to use system proxies
	// note: on Linux, "system" == "environment variables"
	QNetworkProxyFactory::setUseSystemConfiguration(true);

	// Set the locale codec to UTF-8.
	// This makes QFile::encodeName() work on Windows and qPrintable() is equivalent to qUtf8Printable().
	QTextCodec::setCodecForLocale(QTextCodec::codecForMib(106));
```

bstoeger:
Ah yes, I think it was even me who removed all the qUtf8Printable things. Heh.

dirkhh:
merged manually

PR #2234, opened 2019-08-11
[RFC] Cleanup: remove CylinderObjectHelper

bstoeger:
The CylinderObjectHelper class was introduced to enable structured
access to cylinder data in Grantlee templates. The idea is sound, but
the implementation was flawed:

1) A list of these objects were created for every DiveObjectHelper.
   This is wasteful - the list should only be generated on demand.
2) It is a reference wrapper, making it inherently insecure. It would
   be much better to store the values in QStrings, etc. and pass the
   value-based object to grantlee.

Moreover, this seems to never have been used. Remove for now and if
the need arises, implement it as on-demand generated value-objects.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [X] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes the `CylinderObjectHelper` which appears to never have been used. It could be though that "third party" grantlee templates use it. Therefore marking as RFC. In that case I propose to reimplement this object in an inherently more secure way, as on-demand generated value objects. References are evil.

### Changes made:

1) Remove CylinderObjectHelper

### Release note:

Done.

### Documentation change:

I didn't see any mention in the docs - so no...?

### Mentions:

@dirkhh


dirkhh:
I'm pretty sure that Tim Wootton uses this in his templates. He introduced this for that purpose in bd40ef7f424604013e1fe1b5feea648ab6304b72

bstoeger:
Are these templates available somewhere so that I can test an alternative solution?

dirkhh:
I'd suggest contacting him via email...

bstoeger:
Ok - close for now.

bstoeger:
Argh. Qt's metaclass system is *so* frustrating. Q_OBJECTs can't be default copied and having a simple wrapper around cylinder_t be a full Q_OBJECT is wrong anyway. Turn it into a Q_GADGET and grantlee can't use it anymore. You have to define these silly GRANTLEE_BEGIN_LOOKUP blocks.

I'm getting more and more frustrated with Qt - hopefully they fix this mess with Qt 6.0.

bstoeger:
@dirkhh: I reckon that grantlee has added support for Q_GADGET recently:
https://github.com/steveire/grantlee/commit/20b62ee39b17ce49aaaf9b8a2f85e184cc708019
Since we compile grantlee, we are not restricted to old versions, are we?

dirkhh:
Qt is far from perfect - but it's the best we have available.
I have learned not to hope that they "fix" things in new versions.
As for new Grantlee versions... that's fascinating. When we adopted Grantlee it seemed abandoned - but it looks like it's under development again. YAY.
Absolutely, I'm happy to work with a newer version of that!!

bstoeger:
> Qt is far from perfect - but it's the best we have available.

Yes, I know.

> Absolutely, I'm happy to work with a newer version of that!!

Cool - for now I'll write the GRANTLEE_BEGIN_LOOKUP thing and let's schedule it for removal once a new stable version of grantlee is out.

dirkhh:
works for me

PR #2237, opened 2019-08-12
Desktop: show local git repos in recent files

### LINKED ISSUE #2236: Cannot open existing local git repo
When saving a local git repo, one can afterwards go to File->Open and the git repo is in the list of recently used dive logs. However, if one exits Subsurface and restarts it, the repo is not listed in the list of recently used dive logs any more. When trying to open an existing repo, there is (at least in Gnome) not a facility to type any file name. This makes it impossible to open the repo after saving it.

### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Operating system:

Subsurface V 4.9.1 on Ubuntu 18.04

### Mentions:

@dirkhh

dirkhh:
But don't show our cloud storage entry (as that is already in the File menu,
anyway).

This is extremely useful because while you can manually enter a file name to
save to (and therefore can use the 'magic' git repo syntax), on most OSs there
is no way to enter that non-existing 'file name' (which is the git branch in
square brackets) in the file open dialog.

Fixes: #2236

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

Fixes #2236 

### Release note:

This is such a corner case use case that I wouldn't document this

### Documentation change:

This is such a corner case use case that I wouldn't document this

### Mentions:

@willemferguson

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -1457,8 +1457,20 @@ void MainWindow::loadRecentFiles()
 		if (!key.startsWith("File_"))
 			continue;
 		QString file = s.value(key).toString();
-		if (QFile::exists(file))
+
+		// never add our cloud URL to the recent files
+		if (!same_string(prefs.cloud_git_url, "") && file.startsWith(prefs.cloud_git_url))
+			continue;
+		// but allow local git repos
+		QRegularExpression gitrepo("(.*)\\[[^]]+]");
+		QRegularExpressionMatch match = gitrepo.match(file);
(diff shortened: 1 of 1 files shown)


willemferguson:
The only way to do this from within Subsurface is to set it up in the Preferences panel as the local dive log file. Not necessarily convenient but it is in principle accessible.

neolit123:
LGTM

PR #2238, opened 2019-08-12
Turn CylinderObjectHelper into value type

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Turns the CylinderObjectHelper class into a value type, with the following advantages:
1) No possibilities of dangling references
2) No explicit memory management

Moreover, generate the list of CylinderObjectHelper only when needed (on printing). Currently such an object is generated for every dive.

Disadvantage:
grantlee does not directly support Q_GADGET based value types. Therefore currently a GRANTLEE_*_LOOKUP block is needed. This is already fixed in current grantlee master.

For me, the code is now much more logical, but as always YMMV.

Tested on a dummy template.

### Changes made:

1) Make CylinderObjectHelper a value type.
2) Generate CylinderObjectHelpers only on demand.

### Release note:

No - no user visible change.
### Documentation change:

Not concerning this commit - but the whole feature is not mentioned in the docs...?
### Mentions:

@dirkhh

### DIFF
--- a/core/subsurface-qt/CylinderObjectHelper.cpp
+++ b/core/subsurface-qt/CylinderObjectHelper.cpp
@@ -3,36 +3,16 @@
 #include "../qthelper.h"
 
 static QString EMPTY_CYLINDER_STRING = QStringLiteral("");
-CylinderObjectHelper::CylinderObjectHelper(cylinder_t *cylinder) :
-	m_cyl(cylinder) {
-}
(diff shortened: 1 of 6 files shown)


dirkhh (approved):
This does look more obvious to me. Waiting for Travis

PR #2239, opened 2019-08-12
Qt513 mac

dirkhh:
Use new Qt tar.xz file with Qt 5.13
That requires a fix in Grantlee - and while we are looking at Grantlee, let's just update to 5.1

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

Qt 5.13 for Mac
Grantlee 5.1 on all OSs

### Mentions:

@bstoeger

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -1,0 +1,84 @@
+name: Mac
+on: push
+
+jobs:
+  desktopBuild:
+    runs-on: macOS-latest
+    steps:
+    - name: checkout sources
+      uses: actions/checkout@v1
+    - name: setup Homebrew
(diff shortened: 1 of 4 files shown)


dirkhh:
I am baffled by the Mac build failure. Homebrew decides to upgrade `gdal` and that in return wants to upgrade `qt` and then we try to download the Qt sources and of course things time out (there's no way you could build Qt every time you do a Travis build).
I need to understand why this suddenly happens :-(

Subsurface-divelog:
Travis seems completely hosed.
I'll work on moving the other builds over to GitHub Actions as well over the next few days.

PR #2240, opened 2019-08-12
Cleanup: remove default argument from DiveObjectHelper constructor

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes the default argument from the DiveObjectHelper constructor. This is only a documentation thing - setting the default argument to NULL suggests that this is a valid argument, when clearly it will crash immediately.


dirkhh (approved):
This seems obviously correct.

PR #2242, opened 2019-08-13
[TEST]Pass DiveObjectHelper as values and remove DiveListModel indirection (Contains #2238)

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This does two things:
1) Turns DiveObjectHelper into a Q_GADGET object and pass it around as value, not as a reference.
2) Remove the list of DiveObjectHelpers in DiveListModel and generate the DiveObjectHelpers on-demand.

This should make the possibility of DiveObjectHelper corruption impossible (unless I missed places where DiveObjectHelper is passed around as pointers).

But even though I believe that this is a step in the right direction, please note that:
1) This does not solve the fundamental problem: DiveObjectHelper is a thin wrapper around "dive *". This means that it represents a reference to a dive that may be freed and we have no control over DiveObjectHelpers existing in QML.
2) Code that transformed QVariants into `DiveObjectHelper *` will now get default-initialized DiveObjectHelper objects and crash. If these fragments still exist (I didn't find any) they have to be converted to cast to `DiveObjectHelper`.
3) We now have to be very careful that core and DiveListModel do not get out of sync. Otherwise this will lead to invalid DiveObjectHelpers. But at least these crash immediately and shouldn't give strange memory corruption bugs.

### Mentions:

@dirkhh: If you find time during your travels, it would be very helpful if you could find crashes or weird behavior in this PR during usual work (download, new, edit, delete dives, etc). This is not yet a solution, but it would be good to see if it is a realistic route. For some reason bluez doesn't run on my new laptop and starting the mobile app takes a minute or two, making testing a pain.

### DIFF
--- a/core/subsurface-qt/CylinderObjectHelper.cpp
+++ b/core/subsurface-qt/CylinderObjectHelper.cpp
@@ -3,36 +3,16 @@
 #include "../qthelper.h"
 
 static QString EMPTY_CYLINDER_STRING = QStringLiteral("");
-CylinderObjectHelper::CylinderObjectHelper(cylinder_t *cylinder) :
-	m_cyl(cylinder) {
-}
+CylinderObjectHelper::CylinderObjectHelper(cylinder_t *cylinder)
+{
+	if (!cylinder)
+		return;
 
-CylinderObjectHelper::~CylinderObjectHelper() {
-}
-
-QString CylinderObjectHelper::description() const {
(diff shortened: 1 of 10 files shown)


bstoeger:
Close, since the plan is to go with #2244.

PR #2243, opened 2019-08-13
Mobile: decouple full text search from DiveObjectHelper

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This avoids generating DiveObjectHelpers for the purpose of full text search. The reason is actually twofold:
1) If we turn DiveObjectHelper into a value-based object we don't want to generate tons of them in a tight loop. Conceptually, they represent the full data of a dive, as displayed on the dive-details tab or on print-templates. Not temporary throw-away objects.
2) In the longer run we want to move full text search to the core, where we can build a classical word-index and thus also make it available for the desktop app.

### Mentions:

@dirkhh: please check.


bstoeger:
Close, since the plan is to go with #2244.

PR #2244, opened 2019-08-13
[TEST] DiveObjectHelper: the nuclear option [contains #2242 and #2243]

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR is intended for @dirkhh to test the "nuclear option" concerning the mobile crashes that we are observing. This turns DiveObjectHelper into a value type. Yes, the code is crazy and will probably give horrible side effects. But it presents the opportunity to test the theory that the random crashes are due to dangling references concerning DiveObjectHelper and (indirectly) struct dive.

Ideally, we get this PR to act normally and then let it test users who experience these random crashes. Yes, I am aware that the crash data is anonymized, but perhaps we can find such a user. If the data-corruption crashes still happen, then I am really stumped.

### Changes made:

1) Turn DiveObjectHelper into value-object.

@dirkhh: it would be great if you could find bugs in this PR. In my short testing I didn't hit anything obvious, but I'm not using mobile on a regular basis.

Addendum: apparently undo of dive-deletion doesn't update the list properly, will check that out.

### DIFF
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -67,16 +67,6 @@ DiveObjectHelper::DiveObjectHelper(struct dive *d) :
 {
 }
 
-DiveObjectHelper::operator bool() const
-{
-	return !!m_dive;
-}
-
-bool DiveObjectHelper::operator!() const
-{
-	return !m_dive;
-}
-
 int DiveObjectHelper::number() const
 {
 	return m_dive->number;
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -55,6 +55,13 @@ static QString getPressures(struct dive *dive, int i, enum returnPressureSelecto
 	return fmt;
 }
 
+// Qt's metatype system insists on generating a default constructed object,
+// even if that makes no sense. Usage of this object *will* crash.
+DiveObjectHelper::DiveObjectHelper() :
+	m_dive(nullptr)
+{
+}
+
 DiveObjectHelper::DiveObjectHelper(struct dive *d) :
 	m_dive(d)
 {
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -290,13 +290,6 @@ QStringList DiveObjectHelper::cylinders() const
 	return cylinders;
 }
 
-QString DiveObjectHelper::cylinder(int idx) const
-{
-	if ( (idx < 0) || idx > MAX_CYLINDERS)
-		return QString();
-	return getFormattedCylinder(m_dive, idx);
-}
-
 QVector<CylinderObjectHelper> DiveObjectHelper::cylinderObjects() const
 {
 	QVector<CylinderObjectHelper> res;
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -313,11 +313,6 @@ QVector<CylinderObjectHelper> DiveObjectHelper::cylinderObjects() const
 	return res;
 }
 
-QString DiveObjectHelper::tripId() const
-{
-	return m_dive->divetrip ? QString::number((quint64)m_dive->divetrip, 16) : QString();
-}
-
 int DiveObjectHelper::tripNrDives() const
 {
 	struct dive_trip *dt = m_dive->divetrip;
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -87,11 +87,6 @@ int DiveObjectHelper::id() const
 	return m_dive->id;
 }
 
-struct dive *DiveObjectHelper::getDive() const
-{
-	return m_dive;
-}
-
 QString DiveObjectHelper::date() const
 {
 	QDateTime localTime = QDateTime::fromMSecsSinceEpoch(1000*m_dive->when, Qt::UTC);
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -236,7 +236,6 @@ DiveObjectHelper::DiveObjectHelper(const struct dive *d) :
 	singleWeight(d->weightsystems.nr <= 1),
 	suit(d->suit ? d->suit : QString()),
 	cylinders(formatCylinders(d)),
-	cylinderObjects(makeCylinderObjects(d)),
 	maxcns(d->maxcns),
 	otu(d->otu),
 	sumWeight(get_weight_string(weight_t { total_weight(d) }, true)),
@@ -244,6 +243,16 @@ DiveObjectHelper::DiveObjectHelper(const struct dive *d) :
 	startPressure(getStartPressure(d)),
 	endPressure(getEndPressure(d)),
 	firstGas(getFirstGas(d))
+{
+}
+
+DiveObjectHelperGrantlee::DiveObjectHelperGrantlee()
+{
+}
+
+DiveObjectHelperGrantlee::DiveObjectHelperGrantlee(const struct dive *d) :
+	DiveObjectHelper(d),
+	cylinderObjects(makeCylinderObjects(d))
 {
 }
 
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -313,12 +313,6 @@ QVector<CylinderObjectHelper> DiveObjectHelper::cylinderObjects() const
 	return res;
 }
 
-int DiveObjectHelper::tripNrDives() const
-{
-	struct dive_trip *dt = m_dive->divetrip;
-	return dt ? dt->dives.nr : 0;
-}
-
 int DiveObjectHelper::maxcns() const
 {
 	return m_dive->maxcns;
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -247,13 +247,6 @@ bool DiveObjectHelper::singleWeight() const
 	return m_dive->weightsystems.nr <= 1;
 }
 
-QString DiveObjectHelper::weight(int idx) const
-{
-	if ((idx < 0) || idx >= m_dive->weightsystems.nr)
-		return QString();
-	return getFormattedWeight(m_dive, idx);
-}
-
 QString DiveObjectHelper::suit() const
 {
 	return m_dive->suit ? m_dive->suit : QString();
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -67,6 +67,16 @@ DiveObjectHelper::DiveObjectHelper(struct dive *d) :
 {
 }
 
+DiveObjectHelper::operator bool() const
+{
+	return !!m_dive;
+}
+
+bool DiveObjectHelper::operator!() const
+{
+	return !m_dive;
+}
+
 int DiveObjectHelper::number() const
 {
 	return m_dive->number;
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -12,9 +12,9 @@
 
 enum returnPressureSelector {START_PRESSURE, END_PRESSURE};
 
-static QString getFormattedWeight(struct dive *dive, unsigned int idx)
+static QString getFormattedWeight(const struct dive *dive, unsigned int idx)
 {
-	weightsystem_t *weight = &dive->weightsystems.weightsystems[idx];
+	const weightsystem_t *weight = &dive->weightsystems.weightsystems[idx];
 	if (!weight->description)
 		return QString();
 	QString fmt = QString(weight->description);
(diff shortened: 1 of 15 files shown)


core/subsurface-qt/DiveObjectHelper.cpp
```diff
@@ -409,3 +409,21 @@ QString DiveObjectHelper::fullTextNoNotes() const
...
+	if (!d)
+		return false;
+	return contains(get_dive_location(d), filterstring, cs) ||
+	       (d->divetrip && contains(d->divetrip->location, filterstring, cs)) ||
+	       contains(d->buddy, filterstring, cs) ||
+	       contains(d->divemaster, filterstring, cs) ||
+	       contains(d->suit, filterstring, cs) ||
+	       get_taglist_string(d->tag_list).contains(filterstring, cs) ||
+	       (includeNotes && contains(d->notes, filterstring, cs));
+}
```
dirkhh: the implementation is fine - why is this part of DiveObjectHelper? Shouldn't this be in `qthelper.cpp`

core/subsurface-qt/DiveObjectHelper.cpp
```diff
@@ -55,6 +55,13 @@ static QString getPressures(struct dive *dive, int i, enum returnPressureSelecto
 	return fmt;
 }
 
+// Qt's metatype system insists on generating a default constructed object,
+// even if that makes no sense. Usage of this object *will* crash.
+DiveObjectHelper::DiveObjectHelper() :
```
dirkhh: which makes me wonder if there's a way to let us know that such an object was created and who was the caller / backtrace to get us there - and to do so in a way that we can then ask the user to send us THAT log.

core/subsurface-qt/DiveObjectHelper.cpp
```diff
@@ -67,6 +67,16 @@ DiveObjectHelper::DiveObjectHelper(struct dive *d) :
 {
 }
 
+DiveObjectHelper::operator bool() const
```
dirkhh: this comment is about the commit message for "Mobile: Generate DiveObjectHelpers on the fly" - note the typo "On the flip side, NOT the DiveListModel..."

qt-models/divelistmodel.cpp
```diff
@@ -218,6 +217,12 @@ int DiveListModel::getDiveIdx(int id) const
 	return get_idx_by_uniq_id(id);
 }
 
+QModelIndex DiveListModel::getDiveQIdx(int id)
+{
+	int idx = getDiveIdx(id);
+	return idx >= 0 ? createIndex(idx, 0) : QModelIndex();
```
dirkhh, line 200: I read this and the block above for `getIdxForId` three times and they still confuse me. Could you add a couple of comments that help the slow learners what this does? 
`getDiveQIdx` vs `getDiveIdx`?

core/subsurface-qt/DiveObjectHelper.cpp
```diff
@@ -208,176 +108,187 @@ QString DiveObjectHelper::gas() const
...
+	tripId(d->divetrip ? QString::number((quint64)d->divetrip, 16) : QString()),
+	tripNrDives(d->divetrip ? d->divetrip->dives.nr : 0),
+	maxcns(d->maxcns),
+	otu(d->otu),
+	sumWeight(get_weight_string(weight_t { total_weight(d) }, true)),
+	getCylinder(formatGetCylinder(d)),
+	startPressure(getStartPressure(d)),
+	endPressure(getEndPressure(d)),
+	firstGas(getFirstGas(d))
+{
```
dirkhh, line 246: So we use the constructor to render everything that we might need. Instead of rendering it on demand later. Now... how often do we call this constructor? Hmmm easy enough to test, I guess.

dirkhh (commented):
I find this series surprisingly likeable. Yes, the last commit with that monster constructor is wild, but... this should solve many / most of the annoying crashes and frankly, that alone is worth it, assuming the performance impact isn't so bad that we kill the usability of the app. Definitely worth playing with.

Have you done any mobile-on-desktop testing at all? I could of course create an Android app and push it into beta, but given that I'm likely not as responsive as I usually am these next two weeks, even that makes me a bit nervous. If I find time today I'll try this on the desktop and see if I can get at least a rough idea of the impact...
Very nicely done, @bstoeger

dirkhh:
Building straight from this branch and running it on a high end Mac with my ~600 dive dive log there is a noticeable (~13 second) delay after the dive data are opened. Then I get 600 warnings `INFO: DiveListModel::getDive(): accessing invalid dive with id XXX` and then a window opens with the wrong theme (pink instead of blue) and an empty dive list. So this may have received less testing than I thought... :-)

dirkhh:
Starting to drill down on this some more... we appear to access `DiveListModel::getDive()` with indices that are NR..2*NR-1 - that's where all the warnings about invalid dives come from.

core/subsurface-qt/DiveObjectHelper.cpp
bstoeger: The object is created when registering the metatype at application-startup.

qt-models/divelistmodel.cpp
bstoeger: One returns an integer, one a QModelndex. Will add comments.

core/subsurface-qt/DiveObjectHelper.cpp
bstoeger: > Now... how often do we call this constructor?

As you found out - often when showing the dive-list. This should be converted to "classical" Qt model/view semantics. Will do that.

bstoeger:
> Building straight from this branch and running it on a high end Mac with my ~600 dive dive log there is a noticeable (~13 second) delay after the dive data are opened. Then I get 600 warnings INFO: DiveListModel::getDive(): accessing invalid dive with id XXX and then a window opens with the wrong theme (pink instead of blue) and an empty dive list. So this may have received less testing than I thought... :-)

Weird - it worked for me with a dummy divelog (no trips). I will see if I can reproduce these issues.

BTW: Until Friday in one week, I'll be busy at a workshop and conference, so don't expect any speed from me either.

bstoeger:
Moreover, I'd like to split the DiveObjectHelper into a small version for the mobile app and a larger version (which derives from the small version) for printing. E.g. the mobile app doesn't use the CylinderObjectHelper thing, etc.

dirkhh:
Here's how far I got - my workday begins in a few moments and will likely prevent me from doing much more for the next 10 hours (but things sometimes change I and I get a break).
```
1   DiveListModel::getDive(int)                                                                                                 divelistmodel.cpp  278 0x10006c4f8 
2   DiveListSortModel::filterAcceptsRow(int, QModelIndex const&) const                                                          divelistmodel.cpp  54  0x10006c4c8 
3   QSortFilterProxyModelPrivate::source_items_inserted(QModelIndex const&, int, int, Qt::Orientation)                                                 0x101f5d636 
4   QSortFilterProxyModelPrivate::_q_sourceRowsInserted(QModelIndex const&, int, int)                                                                  0x101f614c2 
5   QMetaObject::activate(QObject *, int, int, void * *)                                                                                               0x101fb579a 
6   QAbstractItemModel::endInsertRows()                                                                                                                0x101f3f012 
7   DiveListModel::addDive(QList<dive *> const&)                                                                                divelistmodel.cpp  150 0x10006c8a7 
8   DiveListModel::addAllDives()                                                                                                divelistmodel.cpp  160 0x10006c80f 
```                                                                                                                                                     
So `filterAcceptsRow()` gets called with source_row in the range of NR .. 2*NR-1
I haven't figured out why, yet (because as you can tell, those calls come from the Qt innards)

bstoeger:
Thanks - looks easy to figure out, but I have to prepare my slides. :(

bstoeger:
Weird. This is the diff that makes it break, but only if there are trips:
```
diff --git a/core/subsurface-qt/DiveObjectHelper.cpp b/core/subsurface-qt/DiveObjectHelper.cpp
index 4aaedbc07..11a48ca09 100644
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -55,6 +55,13 @@ static QString getPressures(struct dive *dive, int i, enum returnPressureSelecto
 	return fmt;
 }
 
+// Qt's metatype system insists on generating a default constructed object,
+// even if that makes no sense. Usage of this object *will* crash.
+DiveObjectHelper::DiveObjectHelper() :
+	m_dive(nullptr)
+{
+}
+
 DiveObjectHelper::DiveObjectHelper(struct dive *d) :
 	m_dive(d)
 {
diff --git a/core/subsurface-qt/DiveObjectHelper.h b/core/subsurface-qt/DiveObjectHelper.h
index 75e3e88ef..dd0169e76 100644
--- a/core/subsurface-qt/DiveObjectHelper.h
+++ b/core/subsurface-qt/DiveObjectHelper.h
@@ -9,8 +9,8 @@
 #include <QVector>
 #include <QVariant>
 
-class DiveObjectHelper : public QObject {
-	Q_OBJECT
+class DiveObjectHelper {
+	Q_GADGET
 	Q_PROPERTY(int number READ number CONSTANT)
 	Q_PROPERTY(int id READ id CONSTANT)
 	Q_PROPERTY(int rating READ rating CONSTANT)
@@ -50,6 +50,7 @@ class DiveObjectHelper : public QObject {
 	Q_PROPERTY(QStringList endPressure READ endPressure CONSTANT)
 	Q_PROPERTY(QStringList firstGas READ firstGas CONSTANT)
 public:
+	DiveObjectHelper(); // This is only to be used by Qt's metatype system!
 	DiveObjectHelper(struct dive *dive);
 	int number() const;
 	int id() const;
@@ -97,6 +98,7 @@ public:
 private:
 	struct dive *m_dive;
 };
+	Q_DECLARE_METATYPE(DiveObjectHelper)
 	Q_DECLARE_METATYPE(DiveObjectHelper *)
 
 #endif
diff --git a/qt-models/divelistmodel.cpp b/qt-models/divelistmodel.cpp
index 59339cbd0..a45760c0c 100644
--- a/qt-models/divelistmodel.cpp
+++ b/qt-models/divelistmodel.cpp
@@ -241,7 +241,7 @@ QVariant DiveListModel::data(const QModelIndex &index, int role) const
 
 	DiveObjectHelper *curr_dive = m_dives[index.row()];
 	switch(role) {
-	case DiveRole: return QVariant::fromValue<QObject*>(curr_dive);
+	case DiveRole: return QVariant::fromValue<DiveObjectHelper>(*curr_dive);
 	case DiveDateRole: return (qlonglong)curr_dive->timestamp();
 	}
 	return QVariant();
```
Not sure what is wrong here - debugging QML is so annoying.

bstoeger:
> This is the diff that makes it break, but only if there are trips

Indeed, if I make tripNrDives return 0, I see all dives. So in principle data transport via the Q_GADGET thing works, but for some reason not for the trip thing. This is so weird. I would like to play it via normal model anyway, but still I'd like to understand what is going on here!

bstoeger:
For some reason - when switching to Q_GADGET -- the section property stays empty. Giving up for today.

bstoeger:
Ok, I had to do it - moved the `tripId` and `tripNrDives` data from DiveObjectHelper to the model and also implemented the other list-view data in the model.

Now for some reason the trip display works. The QML properties system is completely meshugge. I'm not being facetious - this is insane.

It should also startup faster because the list now shouldn't generate any DiveObjectHelpers (for me it takes a minute or two because it searches and doesn't find Bluetooth - very annoying).

Next steps:
1) Split out grantlee-only functionality
2) Don't get fresh DiveObjectHelpers for every datum in the DiveDetails*.qml files (could use some help there).
3) Address the comments.

I didn't test dive-adding, etc. I'm sure there are some bad surprises there, but hopefully all fixable. Though that very obscure Q_GADGET behavior gives me a bad feeling.

bstoeger:
@dirkhh: did you have time looking at the new version? We need a decision here before I can continue work on cylinder and weightsystem undo.

dirkhh:
I was focused on things going into 4.9.3
I'll look today and let you know

core/subsurface-qt/DiveObjectHelper.h
```diff
@@ -88,6 +86,21 @@ class DiveObjectHelper {
 	QStringList firstGas;
 	static bool containsText(const struct dive *d, const QString &filterstring, Qt::CaseSensitivity cs, bool includeNotes);
 };
-	Q_DECLARE_METATYPE(DiveObjectHelper)
+
+// This is an extended version of DiveObjectHelper that also keeps track of cylinder data.
+// It is used by grantlee to display structured cylinder data.
+// Note: this grantlee feature is undocumented. If htere turns out to be no users, we migh
```
dirkhh: typo `htere`

dirkhh (commented):
I really like this series. It touches on some many places where when I wrote that code I had no idea what it did on the Qt side and basically stumbled around blind, following some examples. And then others took this and added more complex objects on the Qt side. What you did is invasive but seems to take us to a MUCH better place.
I found one ridiculous typo and nothing else to comment on. Well done.

bstoeger:
Thanks - then I will continue cleaning this up.

core/subsurface-qt/DiveObjectHelper.cpp
bstoeger: Moved it to qthelper.cpp.

core/subsurface-qt/DiveObjectHelper.cpp
bstoeger: Fixed.

core/subsurface-qt/DiveObjectHelper.h
bstoeger: Fixed.

dirkhh:
So... now that 4.9.3 is out, is this ready to be merged?
I haven't figured out how to tell what has changed since the last time I reviewed the code...

qt-models/divelistmodel.cpp
bstoeger: Added comments. Not sure if they are all that helpful though.

bstoeger:
Fixed one commit description - if Travis passes, I will merge even though merging my own PR feels a bit weird. But there seems to be consensus that this is the way forward.

[Side note: I'm a bit anxious that it will turn out that the observed memory corruption is actually happening elsewhere. But then at least we have ruled out one particular source and perhaps get more sensible backtraces.]

bstoeger:
Ok - let's merge. If all hell breaks lose, we still can revert.

PR #2246, opened 2019-08-14
Replace table interpolation by two line fit for CNS

atdotde:
[FitCNS.pdf](https://github.com/Subsurface-divelog/subsurface/files/3500845/FitCNS.pdf)
We used a table lookup for CNS equivalent times. Turns
out the log of this table falls pretty much on a straight
line for po2 <= 1.5bar. We now fit this tabel two two
lines, one for <= 1.5 bar and one above. This four
parameter fit has half the sum of errors squared
than the five parameter fit using a fourth order
polynomial. (For details see attached pdf of mathematica session).

Fitting the log has the advantage that this never
crosses 0, which would have the bad effect of
resulting in negative CNS values as we divide
by the table value.

We refuse to compute CNS values for pO2 exceeding 1.7bar
since this is way beyond the range of the table used. This is signalled by returning a negative CNS value. I hope I caught all occurrences where we need to replace this with an appropriate user facing message.

Superseeds #2227 

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@willemferguson please double check.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -161,35 +161,6 @@ static int calculate_otu(const struct dive *dive)
 	return lrint(otu);
 }
 
-/* Table of maximum oxygen exposure durations, used in CNS calulations.
-   This table shows the official NOAA maximum O2 exposure limits (in seconds) for different PO2 values. It also gives
-   slope values for linear interpolation for intermediate PO2 values between the tabulated PO2 values in the 1st column.
-   Top & bottom rows are inserted that are not in the NOAA table: (1) For PO2 > 1.6 the same slope value as between
-   1.5 & 1.6 is used. This exptrapolation for PO2 > 1.6 likely gives an underestimate above 1.6 but is better than the
-   value for PO2=1.6 (45 min). (2) The NOAA table only tabulates values for PO2 >= 0.6. Since O2-uptake occurs down to
-   PO2=0.5, the same slope is used as for 0.7 > PO2 > 0.6. This gives a conservative estimate for 0.6 > PO2 > 0.5. To
-   preserve the integer structure of the table, all slopes are given as slope*10: divide by 10 to get the valid slope.
-   The columns below are: 
-   po2 (mbar), Maximum Single Exposure (seconds), single_slope, Maximum 24 hour Exposure (seconds), 24h_slope */
-enum cns_table_headers {PO2VAL, SINGLE_EXP, SINGLE_SLOPE, DAILY_EXP, DAILY_SLOPE, NO_COLUMNS};
-static int const cns_table[][5] = {
-	{ 1600,  45 * 60, 456, 150 * 60, 180 },
-	{ 1550,  83 * 60, 456, 165 * 60, 180 },
-	{ 1500, 120 * 60, 444, 180 * 60, 180 },
-	{ 1450, 135 * 60, 180, 180 * 60,  00 },
-	{ 1400, 150 * 60, 180, 180 * 60,  00 },
-	{ 1350, 165 * 60, 180, 195 * 60, 180 },
-	{ 1300, 180 * 60, 180, 210 * 60, 180 },
-	{ 1250, 195 * 60, 180, 225 * 60, 180 },
-	{ 1200, 210 * 60, 180, 240 * 60, 180 },
-	{ 1100, 240 * 60, 180, 270 * 60, 180 },
-	{ 1000, 300 * 60, 360, 300 * 60, 180 },
-	{ 900,  360 * 60, 360, 360 * 60, 360 },
-	{ 800,  450 * 60, 540, 450 * 60, 540 },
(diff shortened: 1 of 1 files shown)


willemferguson:
I am pretty happy with it. For the planner it should work perfectly. My strong reservation is for calculating CNS for dive logs, because it will refuse to calculate CNS of the pO2 is > 1.7. When I do deco at 6m using oxygen, especially in the sea, the pO2 goes > 1.7 all the time, granted it is for pretty short periods only, but in many cases this code would refuse to calculate pO2. Keep in mind that the expected pO2 at 6m in the sea > 1.6. And it is precisely in these cases when using high levels of oxygen that CNS is important to evaluate. So, two requests, please?
1) Let us use the value for either 1.6 or 1.7 when the pO2 is above 1.7.
2) In this case print an warning message in the bottom red bar indicating that CNS value is approximate because some pO2 values are above 1.7.

atdotde:
The value of 1.7 bar is somewhat arbitrary and I am happy to move it a bit (to 1.8bar for example). But the data from the table does not warrant at all to keep the value of 1.6bar above that partial pressure as the table is suggesting the the allowable time is decreasing fast with increased partial pressure of O2. We should use the interpolation function rather than keeping a constant value as that would (if you trust the whole CNS-percentage story at all) be vastly too aggressive. We should err on the side of conservatism rather than otherwise.

For example look at this table: https://www.ivans-dive-team.de/pdf/Sauerstoffuhr.pdf At 1.8bar they consider the maximal exposure time to be 2min.

core/divelist.c
```diff
@@ -233,23 +204,26 @@ static double calculate_cns_dive(const struct dive *dive)
...
 			continue;
-		/* Find the table-row for calculating the maximum exposure at this PO2 */
-		for (j = 1; j < sizeof(cns_table) / (sizeof(int) * NO_COLUMNS); j++)
-			if (po2i > cns_table[j][PO2VAL])
-				break;
-		/* Increment CNS with simple linear interpolation: 100 * time / (single-exposure-time + delta-PO2 * single-slope) */
-		cns += (double)t / ((double)cns_table[j][SINGLE_EXP] - ((double)po2i - (double)cns_table[j][PO2VAL]) * (double)cns_table[j][SINGLE_SLOPE] / 10.0) * 100;
+		if (po2i > 1700)
+			return -1.0;     // We don't have reliable data for too high ppO2
+
```
bstoeger, line 207: "two lines"?

glance-:
I'd rather see that we emit a warning that the cns value is inaccurate for high ppO2's, rather then not calculating them at all.

Its not all that uncommon to end up with high cns values after "big" dives, but its still relevant to know if its 100, 200 or 500.

dirkhh:
(1) let's warn the user once pO2 goes above 1.7
(2) let's try and compute "something" up to 1.8 and refuse above that
I can see the argument that "stuff happens" and your 6m deco stop on O2 can get you to 1.7 or slightly above. If you get past 1.8 giving you any value at all seems mostly misleading. Actually, as I write this... I think I'd go for an area based cutoff. If the area under the graph of "time above 1.7" is larger than X (where I have no idea what X should be). One sample with a pO2 of 2.0 really doesn't make much difference. 1 minute at 2.0 means we have no idea what you CNS would be.
Does that make sense?

willemferguson:
I think this really makes sense. No really accurate area measurement is required, let's say 1 minute above 1.8. For the Galileo that would be 15 samples (4s apart). For the Shearwater it would be 6 samples (10s apart). Subsurface must have the sampling interval to interpret the profile. Robert, thank you very much for your time. I think you thought this was going to be a quickie, and now you have everyone commenting.

atdotde:
For the window thing: we could just continue to use this formula. The result will be that the cns percentage goes through the roof.  Which one could argue is meaningful in itself. Otherwise just keep this approach and just change 1.7 to 1.8. as far as warning Gieses gave always done that.

core/divelist.c
```diff
@@ -233,12 +204,12 @@ static double calculate_cns_dive(const struct dive *dive)
...
+		// This formula is the result of fitting to lines to the Log of the NOAA CNS table
```
bstoeger: Still not sure if this should read "two lines" as per commit description?

willemferguson:
Line 207. Please make that 1800 or make it a user-settable preference. As explained previously, it is not a useable value when using oxygen for decompression. I also support doing some timing for the value above 1800, e.g. 1 minute.

atdotde:
The limit is gone and finally I managed to replace s/to/two/. 

I am on vacation with extremely bad internet, so don't expect much interaction from my side.

glance- (approved):

PR #2255, opened 2019-08-24
Update libdivecomputer to support the Aqualung i200c

torvalds:
I got confirmation from Tiago Thedim Dias that my libdivecomputer patch
makes BLE downloading work from the i200c, and already pushed out the
libdivecomputer changes earlier.  This updates the subproject in
subsurface to have those changes.

This also adds the bluetooth name patterns for the i300c and a few other
Aqualung dive computers we hadn't added yet.  That should make them show
up in the bleutooth device list even without having to check the "Show
all bluetooth devices" check-box.

Tiago claims he didn't need that, and I wonder if we have some overly
permissive match somewhere, but it's the right thing to do regardless.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -78,9 +78,28 @@ static dc_descriptor_t *getDeviceType(QString btName)
 		product = "Quad";
 	}
 
+	// The Pelagic dive computers (generally branded as Oceanic or Aqualung)
+	// show up with a two-byte model code followed by six bytes of serial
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
These look good. I'll have to wait for Travis but I definitely want this in 4.9.2

PR #2258, opened 2019-08-25
Map: catch null divesites in map widget selection code

bstoeger:
Just to be sure, refuse to add null divesites to the selection.

Moreover, refuse to call the setSelected function on a null-divesite.
I got an unfriendly Qt-Warning there:

"Passing incompatible arguments to C++ functions from JavaScript is
dangerous and deprecated."
"This will throw a JavaScript TypeError in future releases of Qt!"

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is supposed to make handling of the map selection code a bit more robust. I doubt that it fixes the crash reported for snap packages (#2248). Ultimately we probably should switch back to a `QVariant` instead of a `divesite *` parameter. The C++ to QML interface is a real pain. :(

### Related issues:

Related to #2248 but I'm nearly certain that this isn't the problem.

### DIFF
--- a/map-widget/qml/MapWidget.qml
+++ b/map-widget/qml/MapWidget.qml
@@ -69,7 +69,7 @@ Item {
 						drag.target: (mapHelper.editMode && mapHelper.model.isSelected(model.divesite)) ? mapItem : undefined
 						anchors.fill: parent
 						onClicked: {
-							if (!mapHelper.editMode)
+							if (!mapHelper.editMode && model.divesite)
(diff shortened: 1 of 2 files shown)


Saviq:
@bstoeger confirmed this doesn't fix #2248.

dirkhh (approved):
Seems like a reasonable improvement, even if it doesn't fix the crash

PR #2260, opened 2019-08-25
[TEST] Map: replace reset-model by remove-all and add-all pair

bstoeger:
!!! TEST COMMIT !!!

@Saviq: Could you please test if this for some crazy reason fixes the crash?

Thanks.


bstoeger:
Never mind - I tested the Travis generated AppImage and it still crashes.

PR #2263, opened 2019-08-26
[WIP] Bugfix planner notes (div tags and incorrect string)

### LINKED ISSUE #2261: Gas consumption in planner notes not displayed corretly - put_format() bug?!
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Gas consumption in planner notes not displayed correctly. Part of the string missing:
grafik

I did some debugging and I have some doubts that I found a bug in put_format()
It may be that the function is processing some strings incorrectly.

I added this code to plannernotes.c:
`/* Gas consumption: Now finally print all strings to output */
printf("----- This goes to buf using put_format():\n");
printf("%s%s%s<br>\n\n\n", temp, warning, mingas);
put_format(&buf, "%s%s%s<br>", temp, warning, mingas);
printf("----- This is the buf after:\n");
printf("%s\n\n\n", buf.buffer);
free(temp);`

And I see in the output the part of one string is not transferred correctly to the buffer:
[debug_planner_notes_buffer.txt](https://github.com/Subsurface-divelog/subsurface/files/3538527/debug_planner_notes_buffer.txt)

### Operating system:

Windows 10

### Subsurface version:

Latest master

### Mentions:

@bstoeger

sfuchs79:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Correct non-balanced div tags in the planner notes.

Change last existing snprintf() functions to asprintf() functions
the have more consistent code.

Change to the _loc version of the function to fix a strange bug
where the string was not copied correctly to the buffer.

### Related issues:

Fixes #2261 

### Additional information:

In fact what really fixes #2261 on my side is this very small change from put_format() to put_format_loc(). Unfortunately my guess would be that put_format() should work there as well and I have still no glue why it fails. Any ideas?

### Mentions:

@bstoeger @atdotde

### DIFF
--- a/core/plannernotes.c
+++ b/core/plannernotes.c
@@ -539,7 +539,7 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
 			}
 		}
 		/* Gas consumption: Now finally print all strings to output */
-		put_format(&buf, "%s%s%s<br>", temp, warning, mingas);
+		put_format_loc(&buf, "%s%s%s<br>", temp, warning, mingas);
 		free(temp);
 		free(warning);
 		free(mingas);
--- a/core/plannernotes.c
+++ b/core/plannernotes.c
@@ -449,14 +449,16 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
 	for (int gasidx = 0; gasidx < MAX_CYLINDERS; gasidx++) {
 		double volume, pressure, deco_volume, deco_pressure, mingas_volume, mingas_pressure, mingas_d_pressure, mingas_depth;
 		const char *unit, *pressure_unit, *depth_unit;
-		char warning[1000] = "";
-		char mingas[1000] = "";
+		char *warning;
+		char *mingas;
 		cylinder_t *cyl = &dive->cylinder[gasidx];
 		if (cyl->cylinder_use == NOT_USED)
 			continue;
 		if (cylinder_none(cyl))
 			break;
 
+		asprintf(&warning, "");
+		asprintf(&mingas, "");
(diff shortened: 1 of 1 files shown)


atdotde:
I have no clue what causes the bug you observe but these changes look good.

bstoeger:
> I have no clue what causes the bug you observe but these changes look good.

Disagree. Neither the new allocations (which I believe leak memory) nor replacing put_format by put_format_loc for no apparent reason look good.

core/plannernotes.c
```diff
@@ -449,14 +449,16 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
...
-		char mingas[1000] = "";
+		char *warning;
+		char *mingas;
 		cylinder_t *cyl = &dive->cylinder[gasidx];
 		if (cyl->cylinder_use == NOT_USED)
 			continue;
 		if (cylinder_none(cyl))
 			break;
 
+		asprintf(&warning, "");
```
bstoeger, line 460: This allocates an empty string that is not freed in the case of other asprintf calls.
If you really want to write it that way, you could do a `warning = strdup("");` and a `free(warning)` before additional `asprintf` calls.

But really, I'd prefer finding the source of the original bug. Is 1000 too small a buffer size?

atdotde:
Does anybody except @sfuchs79 see this bug?
Is it restricted  to the German locale? Does it go away when we increase the buffer size?

dirkhh:
Similar question here: what's the state of this? And I assume I shouldn't worry about it until after 4.9.3?

sfuchs79:
Abandon this because #2266 will bring the solution in a better way.

PR #2266, opened 2019-08-28
Clean plannernotes

glance-:
This introduces some debug ability to planner notes, and then cleans and corrects the html it spits out a bit.

This is quite related to #2261 and #2263

### DIFF
--- a/cmake/Modules/HandleFindGrantlee.cmake
+++ b/cmake/Modules/HandleFindGrantlee.cmake
@@ -10,5 +10,10 @@ else()
 		find_package(Grantlee5 REQUIRED)
 		set(GRANTLEE_LIBRARIES Grantlee5::Templates)
 	endif()
+	# Not really grantlee, but rather printing in general
 	LIST(APPEND QT_EXTRA_COMPONENTS PrintSupport)
+	# Because Qt5WebKitWidgets isn't a part of the "regular" Qt5, we can't get it the normal way
+	#LIST(APPEND QT_EXTRA_COMPONENTS WebKitWidgets)
+	find_package(Qt5WebKitWidgets REQUIRED)
+	set(GRANTLEE_LIBRARIES ${GRANTLEE_LIBRARIES} Qt5::WebKitWidgets)
 endif()
--- a/core/plannernotes.c
+++ b/core/plannernotes.c
@@ -616,4 +616,7 @@ finished:
 	mb_cstring(&buf);
 	free(dive->notes);
 	dive->notes = detach_buffer(&buf);
+#ifdef DEBUG_PLANNER_NOTES
+	printf("<!DOCTYPE html>\n<html>\n\t<head><title>plannernotes</title><meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\"/></head>\n\t<body>\n%s\t</body>\n</html>\n", dive->notes);
+#endif
 }
--- a/core/plannernotes.c
+++ b/core/plannernotes.c
@@ -120,7 +120,7 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
 		return;
 
 	if (error) {
-		put_format(&buf, "<span style='color: red;'>%s </span> %s<br>",
+		put_format(&buf, "<span style='color: red;'>%s </span> %s<br/>",
 				translate("gettextFromC", "Warning:"),
 				translate("gettextFromC", "Decompression calculation aborted due to excessive time"));
 		goto finished;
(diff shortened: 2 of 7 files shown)


bstoeger:
@sfuchs79: This probably collides with your private changes, so please check it out. Thanks.

sfuchs79:
I'm for sure in general ok with cleaning up the non-balanced tags. I would have done part of this as well.

But this change is not ok because it introduces a huge issue with the layout of the planner generated notes as soon as you save the dive in the planner and go back to the main window.
Here are the results - this is correct looking dive w/o the change after saving the dive in the planner:
grafik


This is with the change after saving the dive in the planner:
grafik

I already made the very same experience when (re)formatting the planner notes and as far as I know at this time I learned that it is no good idea:
- to use `\n`
- to use `<br>` in some specific ways

Also I think that `<br/>` is s.th. that should not be used here or at least is not needed for anything.

Now, regarding #2261 hmm... thats a difficult story. I have to admit that this change removes the issue with the truncated string but I have doubts if it really fixes the root cause or maybe only hides the issue. I still have doubts that there is an issue with put_format()

glance-:
Two oddities here.
The notes field renders the planner notes differently than the dive plan details in the planner. I'd never thought that.
The other one is that only the profile gets refreshed, and not the other field get refreshed directly after saving the plan. This causes that one doesn't directly see the planner notes after saving them.

glance-:
Found the real odd one: desktop-widgets/tab-widgets/maintab.cpp:372 , MainTab::updateNotes()
```tmp.replace(QString("\n"), QString("<br>"));```

Probably some odditie from way back in time.

**Comment by glance-** [comment]
Ah, its you @sfuchs79 who wrote that back in b7806d1b945851a6a1713e7d619e1b3f4ddee464

I think that's wrong. \n doesn't mean <br/> in html.

And, `<br>` is the pre html4 variant of the tag. In xhtml, the correct use is the empty tag `<br/>`.

**Comment by sfuchs79** [comment]
Ups, you are for sure right. This `tmp.replace(QString("\n"), QString("<br>"));` is not nice.

But I remember why I did it:
Could you please now also try printing the planned dive via the std. print function from main window with and w/o your change?
I have these results w/o change (old):
[Print_current.pdf](https://github.com/Subsurface-divelog/subsurface/files/3554534/Print_current.pdf)

With change (new):
[Print_new.pdf](https://github.com/Subsurface-divelog/subsurface/files/3554537/Print_new.pdf)


Printing is not able to handle the '\n' or anything (?) else correctly. But maybe we should improve/fix this on the printing side.
All in all I'm not against these improvements. They are clearly correct. But we have to preserve old functionality everywhere...

**Comment by atdotde** [comment]
The other thing you need to preserve is copy to clipboard so you can passte the plan to a word processor to adjust and print wet notes.

**Comment by glance-** [comment]
I'm building the QWebView beast right now so i can test the printing thingie. I'll continue to look at it tomorrow.

@atdotde I'm not shure what you mean right now. If you mean the "Log" -> "Copy dive components" , choose notes, that works, but you get the notes as html whereever you paste it.

**Comment by glance-** [comment]
Ok. Finally I managed to hack togeather a working subsurface with Qt5::WebKitWidgets , and yes, the printing is b0rked. But the printing was borked from before, and didn't care about tables or nothing. The hack to replace all <br/>'s with \n's did hide some of it, but the rest looks quite bad still.

The woke printing problem boils down to the fact that the templates we shipp expects plain text in notes, and the planner generates html-ish. Eg:
```html
                        <table class="notes_table_class">
                            <tbody>
                            <tr>
                                <td class="fieldTitle">
                                    <h1> Notes </h1>
                                </td>
                            </tr>
                            <tr>
                                <td class="fieldData">
                                    <div class="textArea">
                                        <p> {{ dive.notes|safe }} </p>
                                    </div>
                                </td>
                            </tr>
                            </tbody>
                        </table>
```
That will never behave well when notes contains a snippet of html.

A couple of ways to do this might be:
* to rewrite the templates to accept html or ascii
* pre-render the html into ascii and use that
* render the html into a image and inject that as notes
* expose the fact that notes might be html to the template and let the template decide what to do with that

A first variant would probably be to wrap the notes filed in grantle and mangle the html. If its html remove all the \n's , and replace all the <br/>'s with \n's. That would make it behave as bad as its done before but not make it any worse.

glance-:
@atdotde Now when i built subsurface with printing support (for the first time in years...) I tested the print button in the planner and it looked correct.

dirkhh:
Now that I finished the upgrade of our AppImage to Qt 5.12... what is the consensus here - is this still under development? What is it waiting for?

I'm embarrassed to say that I still need to actually do another release (4.9.3) and then announce it (as 4.9.2 was never announced)... I assume this should wait until after?

glance-:
Status is that I've boiled it down to the fact that our printing templates are the last issue. They don't handle planner notes in a good way.

I found more planner notes rewriting in DiveObjectHelper::notes() , so thats patched so we still can generate prints. They are about as ugly as before, but now they don't regress.

sfuchs79:
@glance- 
Did a quick test and this looks good for me now. Brings only improvements, no more drawbacks. Thanks for working on this.

One question: You wrote:
> Two oddities here.
> [...]
> The other one is that only the profile gets refreshed, and not the other field get refreshed directly after saving the plan. This causes that one doesn't directly see the planner notes after saving them.

Honestly I even didn't recognize this behavior in the beginning but now that you mentioned it - this is now still there, isn't it? Did you have a chance to look into this as well?

glance-:
Nope. File a separate issue for that.

dirkhh (approved):
This looks sane - I haven't printed in ages, but a quick sanity check seemed fine.

PR #2267, opened 2019-08-28
 Fix broken windows build with latest MXE

buxtonpaul:
Replaces some enums with names that do not clash with windows #defines.
Specifically:
ERROR -> ERRORED, PASCAL->PASCALS, IGNORE->IGNORED,FLOAT->FLOATVAL

Signed-off-by Paul Buxton paulbuxton.mail@googlemail.com

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

The MXE build for windows pulls in preprocessor defines for ERROR, IGNORE, FLOAT and PASCAL which override the enumerated type definitions. This patch renames the conflicting enums to avoid this. We also remove the #undef PASCAL which was presumably the original attempt to avoid this problem for the PASCAL case.

### DIFF
--- a/core/configuredivecomputer.h
+++ b/core/configuredivecomputer.h
@@ -26,7 +26,7 @@ public:
 		FWUPDATE,
 		CANCELLING,
 		CANCELLED,
-		ERROR,
+		ERRORED,
 		DONE,
 	};
 
--- a/core/dive.c
+++ b/core/dive.c
@@ -3404,7 +3404,7 @@ void set_informational_units(const char *units)
 		if (strstr(units, "PSI"))
 			git_prefs.units.pressure = PSI;
 		if (strstr(units, "PASCAL"))
-			git_prefs.units.pressure = PASCAL;
+			git_prefs.units.pressure = PASCALS;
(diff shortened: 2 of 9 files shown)


buxtonpaul:
Agree that some of the names are less concise than the original forms.
Happy for better alternatives...

buxtonpaul:
Regarding it being a Subsurface vs MXE issue. To be fair I suspect it is more likely one of the packages we are using through MXE rather than MXE itself. 

I will spend some time this weekend to see if I can see which packages, and why. I think it is probably someone has been lazy and included a headerfile in a header rather than just in the c/cpp files that require it.

I notice the travis build for OSX has failed, I cannot see if that is down to my change or not..

dirkhh:
The OSX failure is a timeout - unrelated to your work, of course.

dirkhh (approved):
This passes Travis, it isn't wrong, and it fixes an issue. I'll merge as is and look forward to more fixes from you!

Issue #2264, opened 2019-08-27
Feature of identifying old planner notes in dive notes defect


sfuchs79:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

There was once a feature which automatically identified old planner notes in planned dive notes.
It was/is implemented here:
https://github.com/Subsurface-divelog/subsurface/blob/master/qt-models/diveplannermodel.cpp#L1130

But this feature needs the global variable "disclaimer" holding the translated disclaimer text.

And filling this global variable was removed here:
https://github.com/Subsurface-divelog/subsurface/commit/a4834e198d55d50a44d6f80e2137777b64652ce7#diff-0df10959bbcbe5342528360ce6f541a5L116

We please should have a look together how to make this work again :-)

### Subsurface version:

Current master

### Mentions:

@atdotde @bstoeger

bstoeger:
Do I understand correctly that it is supposed to remove the old disclaimer when replanning a dive?

But how did that even work if it depends on the type of the previously planned dive? That appears extremely brittle to me.

Do you have any suggestion how to fix that?

sfuchs79:
Fixed by #2274

PR #2274, opened 2019-09-01
[WIP] Make disclaimer removal work again

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

There was code that removed the old disclaimer from the planner notes. The functionality was unintentionally removed. Reinstate it, but instead of using a global variable with the recently generated disclaimer be smarter and recognize disclaimers with different deco models.
### Changes made:

1) Reinstate removal of dislaimer
2) Remove global variable
### Related issues:

#2264
### Release note:

Will do once finished.
### Documentation change:

No.
### Mentions:

@sfuchs79: Please check this out. It is different from the old code as it only removes the disclaimer, not the whole thing. You know best what to do, so feel free to adapt.

### DIFF
--- a/core/planner.h
+++ b/core/planner.h
@@ -56,7 +56,6 @@ extern char *get_planner_disclaimer_formatted();
 extern void free_dps(struct diveplan *diveplan);
 extern struct dive *planned_dive;
 extern char *cache_data;
-extern char *disclaimer;
 
 struct divedatapoint *plan_add_segment(struct diveplan *diveplan, int duration, int depth, int cylinderid, int po2, bool entered, enum divemode_t divemode);
 struct divedatapoint *create_dp(int time_incr, int depth, int cylinderid, int po2);
--- a/core/planner.h
+++ b/core/planner.h
@@ -50,6 +50,7 @@ extern int get_cylinderid_at_time(struct dive *dive, struct divecomputer *dc, du
 extern int get_gasidx(struct dive *dive, struct gasmix mix);
 extern bool diveplan_empty(struct diveplan *diveplan);
 extern void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_disclaimer, int error);
+extern const char *get_planner_disclaimer();
 
 extern void free_dps(struct diveplan *diveplan);
 extern struct dive *planned_dive;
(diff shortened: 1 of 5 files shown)


core/planner.h
```diff
@@ -55,7 +55,6 @@ extern const char *get_planner_disclaimer();
 extern void free_dps(struct diveplan *diveplan);
 extern struct dive *planned_dive;
 extern char *cache_data;
-extern char *disclaimer;
```
dirkhh, line 57: The build failure on Windows seems to imply that this is still used somewhere in `mainwindow.cpp`

bstoeger:
Yes, for printing. Obviously this was dysfunctional, as the global variable wasn't set. I will have to look into this when I have a clear mind.

dirkhh:
Ah, and this one needs some attention as well :-)

atdotde (changes requested):
The point of the old code was (and at some point that worked) not so much to remove the disclaimer but actually the old dive plan but not additional stuff that had been manually entered above the plan. The disclaimer was only used as the anchor to find the start of the dive plan (before the disclaimer text got the %s). So I would chop off everything from disclaimerPositionStart.

bstoeger:
OK - changed it in that way. Completely untested, please check it out.

Also removed the global disclaimer variable from the plan-printing routine. Is that still needed?

atdotde:
Hmm, I liked the printed plan having a disclaimer (even if it's just snake oil). The complication with the mentioning of the deco algorithm together with our mixture of C/Qt with %s vs. %1 for .arg() are unfortunately preventing an easy solution. What do you think?

bstoeger:
Ok - added a function that returns disclaimer with current deco-mode.

Please test.

core/planner.h
bstoeger: Should be fixed.

atdotde:
Looks, good. I restarted the Mac build. One could now worry that the plan is not removed when upon replanning the model is switched between VPM-B an Bühlmann (to prevent that only match the first bit of the disclaimer), but I am happy with the current state.

atdotde (approved):

bstoeger:
> One could now worry that the plan is not removed when upon replanning the model is switched between VPM-B an Bühlmann (to prevent that only match the first bit of the disclaimer)

I don't understand: In contrast to the old code, the new code matches beginning and end of the disclaimer, but ignores the deco-type. Now it should work when changing the model as opposed to the old code, no?

atdotde:
You are absolutely right. I only read a singe commit and did not realise you now have two functions, one with %s and one with the actual deco model string.

PR #2276, opened 2019-09-04
Trusty qt512

dirkhh:
Locally I was able to build a Qt 5.12 based AppImage with this.
I bet this won't work on Travis on the first try :-)

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

- create a Trusty container that includes Qt 5.12 with QtWebKit and all the things we need to build Subsurface
- use that container on Travis to build an AppImage

### Related issues:

#2248 

### Mentions:

@bstoeger I'm about to go to bed, busy day tomorrow, but at least I wanted to get this first attempt out there...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,12 @@ matrix:
             - libglib2.0-dev
             - mdbtools-dev
 
+    - env: SUBSURFACE_PLATFORM='linux-trusty-qt512'
+      os: linux
+      language: c++
+      services:
+          - docker
+
     - env: SUBSURFACE_PLATFORM='linux2'
       os: linux
       dist: trusty
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -28,6 +28,10 @@ while [[ $# -gt 0 ]] ; do
 			# force Bluetooth support off
 			BTSUPPORT="OFF"
 			;;
+		-quick)
+			# only build libdivecomputer and Subsurface - this assumes that all other dependencies don't need rebuilding
+			QUICK="1"
+			;;
 		-build-deps)
 			# in order to build the dependencies on Mac for release builds (to deal with the macosx-version-min for those
 			# call this script with -build-deps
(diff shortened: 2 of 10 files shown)


dirkhh:
So this is brilliant. Not. Yes, I can build things in the Docker container, but I can't run the AppImage I need to create the AppImage.

@probonopd is there a trick I am missing?

Do I need to figure out how run this in the Trusty VM inside which the Trusty Docker container is running...? which brings me back to having to manually figure out which shared objects need to be included... it feels like I am So. Close.

probonopd:
Changing

https://github.com/Subsurface-divelog/subsurface/pull/2276/files#diff-d946a51eef804c61365b1b2117f66ceeR33

to

```
./linuxdeployqt*.AppImage --appimage-extract-and-run ./appdir/usr/share/applications/*.desktop -bundle-non-qt-libs -qmldir=./subsurface/map-widget/ -verbose=2
```

should do the trick.

---

Background: Although Docker can be configured to allow FUSE to run inside it, some consider it a security issue and prefer not to use FUSE inside Docker. In those cases, one can still run (the contents of) AppImages using the `--appimage-extract-and-run` switch. 

Reference: https://github.com/AppImage/AppImageKit/wiki/FUSE#docker

dirkhh:
The lack of a useful Travis log on why the `after_success.sh` fails is just frustrating.
I'm sure it's a silly typo, or more likely "thinko" but I can't see it.
So let's make it painfully step by step.

GRMBL. So much time wasted on this. CRAZY

bstoeger:
This is all Chinese to me (well since I know ca. 10 Kanji I probably can make more sense out of Chinese).

But before you get insane over this, I hope you realize that we have a (hopefully) working AppImage: https://transfer.sh/SR2Kg/Subsurface-4.9.2-37-ga0f5a20ec288-x86_64.AppImage
Yes, something else might break, but better than nothing.

probonopd:
> The lack of a useful Travis log on why the after_success.sh fails is just frustrating.

(Temprarily) moving everything from `after_success` to `script` may help in such cases?

PR #2281, opened 2019-09-08
[RFC] Planner: When re-planning a dive update old dive and display correctly

### LINKED ISSUE #2280: Planner: When re-planning existing dive most info not saved after planner exit
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Spin-off from #2266 
When re-planning and existing dive and exiting the planner via "Save", most of the updated dive info is not copied/replaced correctly.

### Operating system:

Windows 10

### Subsurface version:

4.9.2-42 from Travis and my own MXE build

### Steps to reproduce:

- Open example file
[Planner_save_bug.zip](https://github.com/Subsurface-divelog/subsurface/files/3587562/Planner_save_bug.zip)
- Select the single planned dive
- Do "Logbook->Edit dive in planner"
- Change dive e.g. runtime of bottom leg
- Do "Save"
--> Back in main window you can see that profile is updated but all other dive info is not updated.
Even if you unselect - select the dive the update is not done.

### Mentions:

@glance- 
@bstoeger 
@atdotde

sfuchs79:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When editing an existing dive in the planner and after some updates
overwrite the existing dive by pressing "Save" the old dive is
"re-planned".
When doing this, update the old dive correctly and also update the
display of the main window.

Proposal to fix #2280 

### Related issues:

Fixes #2280 

### Mentions:

@bstoeger @atdotde

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -9,6 +9,7 @@
 #include "core/divelist.h" // for mark_divelist_changed()
 #include "core/settings/qPrefDivePlanner.h"
 #include "desktop-widgets/command.h"
+#include "desktop-widgets/mainwindow.h"
 #include "core/gettextfromc.h"
 #include "core/deco.h"
 #include <QApplication>
@@ -1154,8 +1155,13 @@ void DivePlannerPointsModel::createPlan(bool replanCopy)
 		Command::addDive(&displayed_dive, false, false);
 	} else {
 		// we were planning an old dive and rewrite the plan
-		mark_divelist_changed(true);
+		displayed_dive.maxdepth.mm = 0;
+		displayed_dive.dc.maxdepth.mm = 0;
+		fixup_dive(&displayed_dive);
 		copy_dive(&displayed_dive, current_dive);
+		mark_divelist_changed(true);
+		emit updateDiveInfo();
+		MainWindow::instance()->refreshDisplay();
 	}
 
 	// Remove and clean the diveplan, so we don't delete
--- a/qt-models/diveplannermodel.h
+++ b/qt-models/diveplannermodel.h
@@ -111,6 +111,7 @@ signals:
 	void recreationChanged(bool);
 	void calculatedPlanNotes();
 	void variationsComputed(QString);
+	void updateDiveInfo();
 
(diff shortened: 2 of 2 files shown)


sfuchs79:
@bstoeger 
While looking at this again I think a better solution would be to create a new `Command::updateDive` or maybe work with `deleteDive` plus `addDive` (with same old ID).

bstoeger:
Yes, I will do that in due course.

qt-models/diveplannermodel.cpp
```diff
@@ -1154,8 +1155,13 @@ void DivePlannerPointsModel::createPlan(bool replanCopy)
 		Command::addDive(&displayed_dive, false, false);
 	} else {
 		// we were planning an old dive and rewrite the plan
-		mark_divelist_changed(true);
+		displayed_dive.maxdepth.mm = 0;
+		displayed_dive.dc.maxdepth.mm = 0;
+		fixup_dive(&displayed_dive);
 		copy_dive(&displayed_dive, current_dive);
+		mark_divelist_changed(true);
+		emit updateDiveInfo();
```
bstoeger, line 1163: That signal seems not to be caught anywhere...?

bstoeger: @sfuch79: could you please shed light on this signal thing? I don't see where this signal is caught - am I missing something?

Moreover, I think this could use a CHANGELOG.md entry as it fixes a bug.

bstoeger (changes requested):

dirkhh:
So are we moving back to using this one?
If yes, @sfuchs79 would you please address the comments?

bstoeger:
> So are we moving back to using this one?

I changed my mind again. 😀
The other PR contains some fixes, like updating the statistics tab on undo. Therefore, I think we should go forward with it. I just have to find a way to replicate @sfuchs79's problem.

dirkhh:
And this is therefore obsolete, correct?
Closing.

PR #2282, opened 2019-09-08
Show correct notes entry when switching on stops

atdotde:
When gas switching only on stops is selected, the notes
showed an extra line at the not realized stop depth. This
eliminates it. It also makes sure there are no 0 second
spurious entries. And gas switching takes more than zero
time (otherwise we would have to print a line of zero
duration for at the gas switch depth).

Reported-by: tormento <turment@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

The logic for the planner notes is quite brittle. I did some testing but not necessarily enough. Maybe @RickMWalsh could check as well.

### DIFF
--- a/core/planner.c
+++ b/core/planner.c
@@ -933,10 +933,6 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 			if (gi >= 0 && stoplevels[stopidx] <= gaschanges[gi].depth) {
 				/* We have reached a gas change.
 				 * Record this in the dive plan */
-				if (is_final_plan)
-					plan_add_segment(diveplan, clock - previous_point_time, depth, current_cylinder, po2, false, divemode);
-				previous_point_time = clock;
-				stopping = true;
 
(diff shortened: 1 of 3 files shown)


dirkhh:
Hrmpf. I am just about to cut 4.9.3 - and this clearly says "needs more testing"... skipping :-(

RickMWalsh:
Sorry Dirk and Robert,
The code looks sane but I'm currently lacking a development environment and
not in a position to test. Given this fixes an rather minor bug - printing
an unnecessary line in some dive plans - nobody is going to get hurt if it
doesn't make it into 4.9.3.

atdotde:
@dirkhh I wasn't expecting anything else. This fixes a minor bug (that has been around for at least two years) that is even hard to notice.

dirkhh (approved):
The changes seem to make sense, 4.9.3 is out, so let's merge

PR #2287, opened 2019-09-11
Report and handle libdivecomputer errors better

torvalds:
This tries to do the right thing wrt libdivecomputer parse errors, as exemplified by the recent report of an inability to download anything at all from an OSTC2 because one dive was corrupted.

Instead of stopping downloading when one dive fails to parse, we'll continue to the next dive. 

However, that ends up also resulting in a need for better error reporting, since otherwise the parse errors are basically almost entirely silent (they would get reported on the progress bar, but overwritten by the next dive being imported).

So this further tries to separate out the concept of an informational update and an actual download error report, and make the latter be reported using the subsurface error reporting mechanism.

- [x] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -770,7 +770,7 @@ static int dive_cb(const unsigned char *data, unsigned int size,
 	rc = create_parser(devdata, &parser);
 	if (rc != DC_STATUS_SUCCESS) {
 		dev_info(devdata, translate("gettextFromC", "Unable to create parser for %s %s"), devdata->vendor, devdata->product);
-		return false;
+		return true;
 	}
 
 	rc = dc_parser_set_data(parser, data, size);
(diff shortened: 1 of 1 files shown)


torvalds:
NOTE! I'm traveling without any dive computers. 

It's a failure, I know. Long flights across the Atlantic without any diving? Not good. But the end result is that I really can't test this in _any_ sane way, and can only say that it looks somewhat reasonable. But what it looks like if you have hundreds of parse errors and just keep on downloading more dives? Who knows. Hopefully that never happens, and hopefully even if it happens our general error reporting deals with it somewhat gracefully, but ..

dirkhh:
YOU KNOW that you can easily test this with the libdivecomputer emulator, right?
Which runs nicely under Linux and can simulate any errors you'd like to test for...

dirkhh (approved):
This looks like it's an improvement.
I restarted the Mac built that failed because of DNS issues

Issue #2224, opened 2019-08-09
Android crash in trip_shown_dives


dirkhh:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

backtrace
```
  #00  pc 00000000002a8abc  /data/app/org.subsurfacedivelog.mobile-_noeo-4a2W3q-VScJ2MvdQ==/lib/arm64/libsubsurface-mobile.so (trip_shown_dives+64)
  #01  pc 000000000023211c  /data/app/org.subsurfacedivelog.mobile-_noeo-4a2W3q-VScJ2MvdQ==/lib/arm64/libsubsurface-mobile.so (DiveListSortModel::tripTitle(QVariant const&)+192)
  #02  pc 0000000000237d9c  /data/app/org.subsurfacedivelog.mobile-_noeo-4a2W3q-VScJ2MvdQ==/lib/arm64/libsubsurface-mobile.so
  #03  pc 0000000000238024  /data/app/org.subsurfacedivelog.mobile-_noeo-4a2W3q-VScJ2MvdQ==/lib/arm64/libsubsurface-mobile.so (DiveListSortModel::qt_metacall(QMetaObject::Call, int, void**)+76)
  #04  pc 00000000002ffe3c  /data/app/org.subsurfacedivelog.mobile-_noeo-4a2W3q-VScJ2MvdQ==/lib/arm64/libQt5Qml.so
```

### Operating system:

Android 8.1

### Subsurface version:

4.9.1.16 based on #2219

bstoeger:
This one here makes me very nervous.

In tripTitle() we explicitly check that trip is a valid reference. And still it crashes. This means that the core data structures are corrupted.

dirkhh:
I can't run Valgrind on Android. I can run Valgrind on mobile-on-Desktop and am happy to try this again, frustrating as it generally appears to be.
Family appears to be taking up most of my time today - I'll try to get to this, soon.

bstoeger:
Sadly, the last time I tried to find a data corruption bug (loading a log with many weightsystems on old version), valgrind didn't help.

bstoeger:
I tried to run mobile-on-desktop under valgrind and didn't manage to hit any invalid memory accesses (apart from the usual rendering stuff). I added, removed, edited dives and used the undo functionality. Moreover I didn't see anything fundamentally wrong in the dive-manipulation code of qmlmanager.cpp.

[...]

PR #2288, opened 2019-09-14
Mobile: don't pass string-encoded trip pointers through QML

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a desperate attempt at fixing #2224. I doubt that it will do so, but it should make the code more robust. At least we will be sure that there are no stale pointers flying around. If the crash still happens, we have confirmation of data corruption of the core structures.
### Changes made:

1) Add a unique id to trips.
2) Pass unique ids instead of pointers through QML.

### Mentions:

@dirkhh

### DIFF
--- a/core/trip.c
+++ b/core/trip.c
@@ -122,7 +122,9 @@ void remove_dive_from_trip(struct dive *dive, struct trip_table *trip_table_arg)
 
 dive_trip_t *alloc_trip(void)
 {
-	return calloc(1, sizeof(dive_trip_t));
+	dive_trip_t *res = calloc(1, sizeof(dive_trip_t));
+	res->id = dive_getUniqID();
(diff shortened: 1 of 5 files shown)


bstoeger:
Travis fails appear bogus to me.

dirkhh:
> Travis fails appear bogus to me.

Yep - Travis sometimes suffers odd networking issues and then fails like this are the result. I restarted again

dirkhh (approved):
This does look like it would make the code much more robust.

dirkhh (approved):
Sorry to let this dangling for so long. Merging now.

PR #2289, opened 2019-09-14
If only one dive selected, only one temperature in stats tab

avances123:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In stats tab, when only one dive is selected,
on one stat, only average is shown, except temperature which
3 same temps for max, min and avg are shown.

### Changes made:
Check if combined_count is more than one to show max/min temp

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveStatistics.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveStatistics.cpp
@@ -69,8 +69,10 @@ void TabDiveStatistics::updateData()
 	else
 		ui->sacLimits->setAverage("");
 
-	ui->tempLimits->setMaximum(get_temperature_string(stats_selection.max_temp, true));
-	ui->tempLimits->setMinimum(get_temperature_string(stats_selection.min_temp, true));
+	if (stats_selection.combined_count > 1){
+		ui->tempLimits->setMaximum(get_temperature_string(stats_selection.max_temp, true));
+		ui->tempLimits->setMinimum(get_temperature_string(stats_selection.min_temp, true));
+	}
 	if (stats_selection.combined_temp.mkelvin && stats_selection.combined_count) {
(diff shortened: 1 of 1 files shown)


desktop-widgets/tab-widgets/TabDiveStatistics.cpp
```diff
@@ -69,8 +69,11 @@ void TabDiveStatistics::updateData()
 	else
 		ui->sacLimits->setAverage("");
 
-	ui->tempLimits->setMaximum(get_temperature_string(stats_selection.max_temp, true));
-	ui->tempLimits->setMinimum(get_temperature_string(stats_selection.min_temp, true));
+	if (stats_selection.combined_count > 1){
```
bstoeger, line 72: What happens if the selection is changed from multiple to one dive? Then the old values remain, no?

Unless I'm missing something, I suggest adding an else branch that clears the fields in the case of a single dive.

bstoeger (changes requested):
Thank you for your contribution.

Please merge the two commits into a single commit. There is no point in a whitespace-fix commit if you could simply just not introduce the whitespace problem in the first place.

desktop-widgets/tab-widgets/TabDiveStatistics.cpp
avances123: When you change from multiple to one dive, the new value is the only which is shown,

avances123:
You can select the option "squash and merge"  and get this job done. Anyways I have rebased it for you.

bstoeger (approved):
Please add space before braces, apart from that - looks good to me.

dirkhh:
fixed whitespace and merged this. Thanks for your contribution.

PR #2297, opened 2019-09-22
Cleanup: Unify code for idiosyncratic singletons

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

We have these weird singletons owing to the way we use them in QML: The instance pointer is set in the constructor. This PR generates a base-class for these singletons. This is done for two reasons:
1) Documentation. It is hoped that we replace these in the long run.
2) Improved debugging. We will now see wanted and unwanted generation of these singletons in the log.

This is a preparation-PR for an attempt to avoid passing table-pointers through QML. The idea is that the download thread updates the model of the downloaded dives itself. But for that the latter has to be one of these strange singletons because it is used from QML and from desktop.

### Changes made:

1) Implement a singleton class that allows generation via new and instance()
2) Convert the weird singletons
3) Un-singletonize GpsLocation

### Mentions:

@dirkhh: this could need some testing, though it seemed to run properly for me (download not tested as I still don't have working BT).

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -15,7 +15,6 @@ extern QMap<QString, dc_descriptor_t *> descriptorLookup;
 namespace {
 	QHash<QString, QBluetoothDeviceInfo> btDeviceInfo;
 }
-BTDiscovery *BTDiscovery::m_instance = NULL;
 
 static dc_descriptor_t *getDeviceType(QString btName)
 // central function to convert a BT name to a Subsurface known vendor/model pair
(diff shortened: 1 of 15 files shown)


bstoeger:
@dirkhh: since #2299 was trashed, this is now optional. Up to you.

dirkhh:
Yes, let's at least get this part in.

PR #2298, opened 2019-09-22
Fix  path of machine-id in dbus library appimage.

buxtonpaul:
When building dbus within the appimage, cmake picks up the installation
path of various files dbus uses through the GNUInstallDirs package,
however this doesn't work under the appimage build.
So we replace the variable with the normal location of this file.

Signed-off-by: Paul Buxton <paubuxton.mail@googlemail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

On the mailing list David Tillotson raised an issue with the Appimage running on Devuan distribution. Where dbus was failling to locate the file machine-id. This file is normally created by systemd and is located in /etc/machine-id. If dbus is installed on a system without this file it will generate a machine-id in /var/lib/dbus/machine-id. 

The root cause of this is that it looks like dbus doesn't typically get built/installed using the default  gnu install paths, specifically it is built with --localstatedir=/var 
see http://www.linuxfromscratch.org/lfs/view/systemd/chapter06/dbus.html

### DIFF
--- a/scripts/docker/trusty-qt512/Dockerfile
+++ b/scripts/docker/trusty-qt512/Dockerfile
@@ -55,7 +55,8 @@ RUN rm -rf qtwebkit
 ADD dbus-1.12.16.tar.gz .
 RUN mkdir dbus-1.12.16/build && \
     cd dbus-1.12.16/build && \
-    cmake ../cmake -DDBUS_INSTALL_SYSTEM_LIBS=1 && \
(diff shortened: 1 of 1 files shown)


dirkhh:
This looks like it would be the right thing. I'll create a combined fresh pull request of the two PRs and push the resulting Docker image so we can then create AppImages for testing (yeah, a bit complicated).

dirkhh:
merged via #2300

PR #2299, opened 2019-09-22
[DO NOT MERGE] Attempt at keeping the DiveImportedModel consistent (contains #2297)

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

There were numerous ways in which the DiveImportedModel could go out of sync. For example by sharing a single dive table with the download thread(!). Also the import-functions consume the tables without the model being adapted. I see no obvious way how this can lead to crashes. But let's clean this up as a first step.

This also removes moving pointers via QML, which has us caused terrible problems before.

This is FYI. It is only compile tested. If some brave person compiles and tests download that would be very helpful (desktop and mobile). Currently, I can't test this as I haven't found time to install Bluetooth.

### Changes made:

Too many to list in detail, sorry.
1) Make DiveImportedModel the authoritative source of the imported dives.
2) Implement move semantics to move downloaded dives from DownloadThread to DiveImportedModel.
3) Keep model in sync.

### Related issues:

Related to #2292 and #2221, but probably doesn't fix them. They should give different backtraces, though.

### Mentions:

@dirkhh: FYI - can't work on this anymore today.

### DIFF
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -3,6 +3,7 @@
 #include "core/qthelper.h"
 #include "core/settings/qPrefDiveComputer.h"
 #include "core/divelist.h"
+#include "qt-models/diveimportedmodel.h"
 #include <QDebug>
 #if defined(Q_OS_ANDROID)
 #include "core/subsurface-string.h"
@@ -63,6 +64,10 @@ DownloadThread::DownloadThread() : downloadTable({ 0 }),
 	diveSiteTable({ 0 }),
 	m_data(DCDeviceData::instance())
 {
+	// By connecting the finished signal in the constructor, we make sure that we are called first.
+	// Thus, we can update the model and all other recipients of the signal (notably QML) can
+	// check the model for new dives. Yes, this is somewhat subtle, but well defined.
+	connect(this, &QThread::finished, this, &DownloadThread::reloadModel);
 }
 
 void DownloadThread::run()
@@ -108,6 +113,11 @@ void DownloadThread::run()
 	qPrefDiveComputer::set_device_name(m_data->devBluetoothName());
 
 	updateRememberedDCs();
+}
+
+void DownloadThread::reloadModel()
+{
+	DiveImportedModel::instance()->repopulate(&downloadTable, &diveSiteTable);
 }
 
 static void fill_supported_mobile_list()
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -318,16 +318,6 @@ DCDeviceData *DownloadThread::data()
 	return m_data;
 }
 
-struct dive_table *DownloadThread::table()
-{
-	return &downloadTable;
-}
-
-struct dive_site_table *DownloadThread::sites()
-{
-	return &diveSiteTable;
-}
-
 QString DCDeviceData::vendor() const
 {
 	return data.vendor;
--- a/core/downloadfromdcthread.cpp
+++ b/core/downloadfromdcthread.cpp
@@ -117,7 +117,8 @@ void DownloadThread::run()
 
 void DownloadThread::reloadModel()
 {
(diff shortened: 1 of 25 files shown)


bstoeger:
PS: In case you wonder - compiling the tests fails because core/downloadfromdcthread.cpp now directly accesses the DivesImportedModel.

The former is linked with the tests, the latter isn't. Either remove the thread from the core lib or link the tests with the models...

qt-models/diveimportedmodel.h
```diff
@@ -4,8 +4,9 @@
 #include <QAbstractTableModel>
```
dirkhh, line 4: The subject line of the commit has a typo: `DiveImporedModel` should be `DiveImportedModel`
And the description in the commit message has two verbs "fill reload"

core/downloadfromdcthread.cpp
```diff
@@ -63,6 +64,10 @@ DownloadThread::DownloadThread() : downloadTable({ 0 }),
 	diveSiteTable({ 0 }),
 	m_data(DCDeviceData::instance())
 {
+	// By connecting the finished signal in the constructor, we make sure that we are called first.
+	// Thus, we can update the model and all other recipients of the signal (notably QML) can
+	// check the model for new dives. Yes, this is somewhat subtle, but well defined.
+	connect(this, &QThread::finished, this, &DownloadThread::reloadModel);
```
dirkhh, line 70: I love this. Brilliant.

mobile-widgets/qml/DownloadFromDiveComputer.qml
```diff
@@ -28,11 +28,6 @@ Kirigami.Page {
 		id: downloadThread
 
 		onFinished : {
-			if (!table || !sites) {
-				console.warn("DCDownloadThread::onFinished(): table or sites is null!")
-				return
-			}
-			importModel.repopulate(table, sites)
```
dirkhh, line 35: not having to do this from QML is So. Much. Better.

dirkhh (approved):
Read through this twice and can see several spots where this prevents corruption that could happen with the current code. Thinking this through I'm now almost surprised that the current code works at all in many instances, as the correct tables were only available if the functions happened to be called in a certain order.
A couple of small comments. And I'll look at the build failure as well as I would love to see this merged so we can test this on Android (and iOS).

dirkhh:
Hrmpf. The fix to compiling the tests seems straight forward. Just like for the regular desktop app, ad the `subsurface_models_desktop` library in `tests/CMakeLists.txt`
Except that fails with an undefined reference to `DiveImportedModel::repopulate()` - an issue we don't have when linking Subsurface - only with the tests.
Right now I'm feeling like I can't see the forest for the trees...

bstoeger:
No worries - I will change it so that the model connects to the thread. Then things should compile again.

BTW: While I believe that the changes in this PR are in principle correct, I doubt that it will fix any crashes. The `data()` function was quite resilient against returning garbage. Let's see.

dirkhh:
The latest Subsurface-mobile Android beta is built with this included - and one of the people who had crashes on import reports that he didn't see those crashes with this version

bstoeger:
Interesting... I hope you realize that this was all done completely blind with the only testing being compile and run. 😮

dirkhh:
> Interesting... I hope you realize that this was all done completely blind with the only testing being compile and run. 😮

I didn't. I reviewed the code and it seemed to make sense and I could identify two situations around download thread where this seems much more likely to do the right thing...

bstoeger:
Ok - that's good news. At least we have a grasp of what we should never do:
1) Pass C-pointers through QML.
2) Modify models in QML context.

I have an idea how to solve this in a much nicer way, which will mean completely rewriting this PR. Please give me a few days - currently I have some mineralogical problems to tackle.

bstoeger:
Yes, let's redo this in a more sensible way...

PR #2302, opened 2019-09-25
Please Test: Make DownloadThread subobject of DiveImportedModel [Alternative to #2299]

bstoeger:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This is an alternative way of implementing #2299: Since downloading goes via the dive-imported model anyway, we might as well make the data flow:
  UI<->Model<->Thread
This gives much less headache than all three modules talking to each other and trying desperately to stay in sync. This should also make debugging much easier as there should be less dreaded QML in the backtraces. This does not go full in: The desktop version still directly accesses the thread in places, for example to read out the error code. We might make this still much cleaner, but it is a start.

Note: this is mostly untested. The first version will crash and burn. Testing would be highly appreciated!

### Changes made:

1) Make DownloadThread a subobject of DiveImportedModel
2) Don't move table-pointers through QML

### Related issues:

#2292 and #2221 report crashes in parts of the code that are removed here. So hopefully this improves the situation!

### Release note:

Let's wait if this actually fixes something or makes it worse.

### Documentation change:

No.
### Mentions:

@dirkhh: please test if you find time.

Issue #2295, opened 2019-09-22
Significant performance regression introduced in f8c5c8b


dirkhh:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

latest master, at least with Subsurface-mobile, shows a significant performance regression at start with my ~650 dive log

### Operating system:

Tried both with mobile on Desktop under Linux as well as with Android

### Subsurface version:

latest master
a bisect shows that the issue is introduced by f8c5c8bedf

### Steps to reproduce:

start Subsurface=mobile
significant delay (~15-20 seconds) after dive log is opened

### Mentions:

@bstoeger - this one is yours...

bstoeger:
Strange - I get that long wait even with a very short test file. But once it's loaded it seems to behave normally.

dirkhh:
Yes, it's just at the start. But the delay on Android is so long that I don't want to have people think that the app is hung at start - that was my first perception when trying. It takes a LONG time where it just sits at the logo.

dirkhh:
I'm trying to profile this and am confused.
The gprof output lists a total time of 2.64 seconds for startup - yet it actually takes 28 seconds to start up.
GRMBL
I'm traveling the next 48 hours... I guess I'll have to release a Subsurface-mobile beta with that performance regression...

bstoeger:
It happens here:
```
void DiveListModel::reload()
{
        beginResetModel();
        endResetModel();
}
```
For me that generates 60(!) full DiveObjectHelper objects per dive(!).
Of course, the stack dump only shows QML. Brilliant.

bstoeger:
Here's the stack back trace:
```
#0  DiveListModel::data (this=0x7fffffffdb80, index=..., role=257) at /home/bs/src/subsurface/qt-models/divelistmodel.cpp:211
#1  0x00007ffff59e4939 in QSortFilterProxyModel::data(QModelIndex const&, int) const () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#2  0x00007ffff65adaad in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
#3  0x00007ffff66d3044 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5
[...]

PR #2303, opened 2019-09-26
Mobile: Don't access dive-id via DiveObjectHelper

bstoeger:
There is already a role to do that. Query the model directly to
avoid creating a full DiveHelperObject.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Avoid creating an unnecessary DiveObjectHelper just to fetch the dive-id. The id is already exported by the model.

### DIFF
--- a/mobile-widgets/qml/DiveDetailsView.qml
+++ b/mobile-widgets/qml/DiveDetailsView.qml
@@ -397,7 +397,7 @@ Item {
 		}
 		Component.onCompleted: {
 			qmlProfile.setMargin(Kirigami.Units.smallSpacing)
-			qmlProfile.diveId = model.dive.id;
+			qmlProfile.diveId = model.id;
 			qmlProfile.update();
 		}
 	}
--- a/mobile-widgets/qml/DownloadFromDiveComputer.qml
+++ b/mobile-widgets/qml/DownloadFromDiveComputer.qml
@@ -391,7 +391,7 @@ Kirigami.Page {
 					manager.appendTextToLog("Save downloaded dives that were selected")
 					importModel.recordDives()
 					manager.saveChangesLocal()
-					diveModel.reload()
+					diveModel.resetInternaData()
 					pageStack.pop();
 					download.text = qsTr("Download")
 					divesDownloaded = false
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -275,7 +275,7 @@ void QMLManager::applicationStateChanged(Qt::ApplicationState state)
 
 void QMLManager::openLocalThenRemote(QString url)
 {
-	clear_dive_file_data();
+	DiveListModel::instance()->clear();
 	setNotificationText(tr("Open local dive data file"));
 	QByteArray fileNamePrt = QFile::encodeName(url);
 	bool glo = git_local_only;
@@ -722,7 +722,7 @@ void QMLManager::loadDivesWithValidCredentials()
 	// if we aren't switching from no-cloud mode, let's clear the dive data
 	if (!noCloudToCloud) {
 		appendTextToLog("Clear out in memory dive data");
-		clear_dive_file_data();
+		DiveListModel::instance()->clear();
 	} else {
 		appendTextToLog("Switching from no cloud mode; keep in memory dive data");
 	}
--- a/qt-models/divelistmodel.cpp
+++ b/qt-models/divelistmodel.cpp
@@ -168,6 +168,13 @@ void DiveListModel::updateDive(int i, dive *d)
 	insertDive(i);
 }
 
+void DiveListModel::clear()
+{
+	beginResetModel();
+	clear_dive_file_data();
+	endResetModel();
+}
+
 void DiveListModel::reload()
 {
 	beginResetModel();
--- a/qt-models/divelistmodel.cpp
+++ b/qt-models/divelistmodel.cpp
@@ -73,7 +73,7 @@ int DiveListSortModel::getIdxForId(int id)
 void DiveListSortModel::reload()
 {
 	DiveListModel *mySourceModel = qobject_cast<DiveListModel *>(sourceModel());
-	mySourceModel->reload();
+	mySourceModel->resetInternalData();
 }
 
 // In QtQuick ListView, section headings can only be strings. To identify dives
@@ -177,8 +177,16 @@ void DiveListModel::clear()
 
 void DiveListModel::reload()
 {
-	beginResetModel();
-	endResetModel();
+	// Note: instead of doing a (logical) beginResetModel()/endResetModel(),
+	// we add the rows (if any). The reason is that a beginResetModel()/endResetModel()
+	// pair resulted in the DiveDetailsPage being renedered for *every* dive in
+	// the list. It is unclear whether this is a Qt-bug or intended insanity.
+	// Therefore, this function must only be called after having called clear().
+	// Otherwise the model will become inconsistent!
+	if (dive_table.nr > 0) {
+		beginInsertRows(QModelIndex(), 0, dive_table.nr - 1);
+		endInsertRows();
+	}
 }
 
 void DiveListModel::resetInternalData()
--- a/qt-models/divelistmodel.h
+++ b/qt-models/divelistmodel.h
@@ -62,6 +62,7 @@ public:
 	QHash<int, QByteArray> roleNames() const;
 	QString startAddDive();
 	void resetInternalData();
+	void clear(); // Clear all dives in core
 	Q_INVOKABLE DiveObjectHelper at(int i);
 };
 
--- a/qt-models/divelistmodel.h
+++ b/qt-models/divelistmodel.h
@@ -53,7 +53,7 @@ public:
 	void removeDive(int i);
 	void removeDiveById(int id);
 	void updateDive(int i, dive *d);
-	void reload();
+	void reload(); // Only call after clearing the model!
 	struct dive *getDive(int i);
 	int rowCount(const QModelIndex &parent = QModelIndex()) const;
 	int getDiveIdx(int id) const;


bstoeger:
Added a quick-fix for #2295. This needs to be audited/cleaned and made compatible with #2302.
Note: I'm very compelled to never touching anything QML-related again. 😠 This is pure madness.
Note 2: When switching to a dive this still does four reloads of the DiveDetails page!

dirkhh (approved):
Thanks for tracking this down. Reading the code it looks... well, maybe not exactly sane, but with the explanations I get it. I'll test this in a moment.

dirkhh:
> Note: I'm very compelled to never touching anything QML-related again. 

@bstoeger I have said the same thing many times, but the fact of the matter is... rewriting the two mobile apps in Java and ObjC/Swift isn't really a feasible alternative. There simply isn't a better framework that I have found that abstracts out so many things for us and allows us to have an app with a largely consistent code base across Windows/Mac/Linux/iOS/Android.

So as much as I sympathize with your sentiment, can I bribe you into not abandoning us to our fate here? Because frankly, you are the #1 person in helping me figure out some of the more insane breakage here. I'd be lost without your help.

dirkhh:
I need to dig into this some more. This seems to fix the issue when building on the desktop.
In my first attempt to test this on Android I get a crash 
```
09-26 09:41:14.522 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Detected OnePlus device, trying to force bundled font "Roboto2001047"
09-26 09:41:14.539 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Qt reports default font is set as "Roboto2001047"
09-26 09:41:14.704 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Generating second instance of singleton 8QMLPrefs
09-26 09:41:14.704 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Generated singleton 8QMLPrefs
09-26 09:41:14.704 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Generating second instance of singleton 10QMLManager
09-26 09:41:14.704 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Generated singleton 10QMLManager
09-26 09:41:24.296 30515 30515 F DEBUG   : pid: 30300, tid: 30344, name: qtMainLoopThrea  >>> org.subsurfacedivelog.mobile <<<
09-26 09:41:24.299 30515 30515 F DEBUG   :     #01 pc 00000000001c4494  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libQt5Core.so
09-26 09:41:24.299 30515 30515 F DEBUG   :     #02 pc 00000000001c656c  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libQt5Core.so (QIODevice::write(char const*, long long)+156)
09-26 09:41:24.299 30515 30515 F DEBUG   :     #03 pc 00000000001d541c  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libsubsurface-mobile.so (QMLManager::writeToAppLogFile(QString)+136)
09-26 09:41:24.299 30515 30515 F DEBUG   :     #04 pc 00000000001d5284  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libsubsurface-mobile.so (writeToAppLogFile(QString)+220)
09-26 09:41:24.299 30515 30515 F DEBUG   :     #05 pc 000000000023889c  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libsubsurface-mobile.so (MessageHandlerModel::addLog(QtMsgType, QString const&)+596)
09-26 09:41:24.299 30515 30515 F DEBUG   :     #06 pc 00000000000b1834  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libQt5Core.so
09-26 09:41:24.299 30515 30515 F DEBUG   :     #07 pc 00000000000ad4e8  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libQt5Core.so
09-26 09:41:24.299 30515 30515 F DEBUG   :     #08 pc 00000000000ad438  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libQt5Core.so (QMessageLogger::debug(char const*, ...) const+132)
09-26 09:41:24.299 30515 30515 F DEBUG   :     #09 pc 00000000001bdc74  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libsubsurface-mobile.so (QMLManager::QMLManager()+244)
09-26 09:41:24.299 30515 30515 F DEBUG   :     #10 pc 0000000000201488  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libsubsurface-mobile.so (void QQmlPrivate::createInto<QMLManager>(void*)+16)
09-26 09:41:24.299 30515 30515 F DEBUG   :     #11 pc 00000000002c6dd4  /data/app/org.subsurfacedivelog.mobile-Lu-XxrkTKqKAFR-clG0bJA==/lib/arm64/libQt5Qml.so (QQmlType::create(QObject**, void**, unsigned long) const+104)
```
I'm a bit suspicious why this would happen only on Android and will do a full rebuild and try again, but wanted to provide this to you just in case.

dirkhh:
Did a completely clean rebuild from scratch and can reliably reproduce this crash at start

bstoeger:
Oh jeez - just what we needed: A device-only crash.

To me this looks like a problem of the previous PR (#2297). Look at the messages:
```
09-26 09:41:14.704 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Generating second instance of singleton 8QMLPrefs
```
Is this the whole log? Where was the first instance created? Have you tried running current master on device?

I'll look into this later this evening.

bstoeger:
This is the log I get when running mobile-on-desktop. Note how there is no mention of `second instance` this or that:
```
Generated singleton 13DiveListModel
Generated singleton 12GpsListModel
INFO: Generated singleton 8QMLPrefs
INFO: Generated singleton 10QMLManager
INFO: "0.000: Starting Subsurface-mobile:2.2.1(4.9.3.76):Linux (Ubuntu 19.04):x86_64:en-US"
INFO: "0.000: built with libdivecomputer v0.7.0-devel-Subsurface-NG (ce6d9896a79afaa82641132e338f8744714c8593)"
INFO: "0.000: built with Qt Version 5.12.2, runtime from Qt Version 5.12.2"
INFO: "0.000: built with libgit2 0.27.7"
INFO: "0.000: Running on Ubuntu 19.04"
INFO: Generated singleton 11BTDiscovery
INFO: Cannot find a running Bluez. Please check the Bluez installation.
INFO: BTDiscoveryReDiscover: localBtDevice.isValid() false
INFO: localBtDevice isn't valid or not connectable
INFO: "Created position source geoclue2"
INFO: "75.079: Created position source geoclue2"
INFO: "Set GPS service update interval to 300 s"
INFO: "75.079: Set GPS service update interval to 300 s"
INFO: "75.079: location service is not available"
INFO: qrc:/org/kde/kirigami/Page.qml:276:5: QML Loader: Binding loop detected for property "active"
INFO: Window width changed to 510
INFO: Column width 510 root item width 510
INFO: width in Grid Units 30 original gridUnit 17 now 17
INFO: Model size of -1 is less than 0
INFO: Both point size and pixel size set. Using pixel size.
INFO: "75.389: Synchronising data file"
INFO: "75.392: Load dives from local cache"
INFO: "75.434: Successfully opened dive data"
INFO: "76.725: 57 dives loaded from cache"
INFO: "76.725: have cloud credentials, but user asked not to connect to network"
INFO: Running on AU Optronics  eDP-1
INFO: Column width 510 root item width 510
INFO: width in Grid Units 30 original gridUnit 17 now 17
INFO: Using the following font: Sans at 10pt with mobile_scale: 1
INFO: qrc:/org/kde/kirigami/GlobalDrawer.qml:258: TypeError: Cannot read property 'length' of undefined
INFO: qrc:/org/kde/kirigami/GlobalDrawer.qml:293: TypeError: Cannot read property 'length' of undefined
INFO: qrc:/org/kde/kirigami/GlobalDrawer.qml:292: TypeError: Cannot read property 'length' of undefined
INFO: qqwindow devicePixelRatio 1 1
INFO: Supported dive computers:
INFO: "Aeris: 500 AI (SERIAL), A300 (SERIAL), A300 AI (SERIAL), A300CS (SERIAL), Atmos 2 (SERIAL), Atmos AI (SERIAL), Atmos AI 2 (SERIAL), Compumask (SERIAL), Elite (SERIAL), Elite T3 (SERIAL), Epic (SERIAL), F10 (SERIAL), F11 (SERIAL), Manta (SERIAL), XR-1 NX (SERIAL), XR-2 (SERIAL)"
INFO: "Apeks: Quantum X (SERIAL)"
INFO: "Aqualung: i100 (SERIAL), i200 (SERIAL), i200C (SERIAL, BLE), i300 (SERIAL), i300C (SERIAL, BLE), i450T (SERIAL), i550 (SERIAL), i550C (SERIAL, BLE), i750TC (SERIAL, BT), i770R (SERIAL, BLE)"
INFO: "Atomic Aquatics: Cobalt (USB), Cobalt 2 (USB)"
INFO: "Beuchat: Mundial 2 (SERIAL), Mundial 3 (SERIAL), Voyager 2G (SERIAL)"
INFO: "Citizen: Hyper Aqualand (SERIAL)"
INFO: "Cochran: Commander I (SERIAL), Commander II (SERIAL), Commander TM (SERIAL), EMC-14 (SERIAL), EMC-16 (SERIAL), EMC-20H (SERIAL)"
INFO: "Cressi: Cartesio (SERIAL), Drake (SERIAL), Edy (SERIAL), Giotto (SERIAL), Goa (SERIAL), Leonardo (SERIAL), Newton (SERIAL)"
INFO: "Deepblu: Cosmiq+ (BLE)"
INFO: "Dive Rite: NiTek Q (SERIAL), NiTek Trio (SERIAL)"
INFO: "DiveSystem: Orca (SERIAL), iDive DAN (SERIAL), iDive Deep (SERIAL), iDive Easy (SERIAL), iDive Free (SERIAL), iDive Pro (SERIAL), iDive Reb (SERIAL), iDive Stealth (SERIAL), iDive Tech (SERIAL), iDive X3M (SERIAL)"
INFO: "Garmin: Descent Mk1 (USBSTORAGE)"
INFO: "Genesis: React Pro (SERIAL), React Pro White (SERIAL)"
INFO: "Heinrichs Weikamp: Frog (SERIAL, BT), OSTC (SERIAL), OSTC 2 (SERIAL, BT, BLE), OSTC 2 TR (SERIAL, BT, BLE), OSTC 2C (SERIAL), OSTC 2N (SERIAL), OSTC 3 (SERIAL), OSTC 4 (SERIAL, BT, BLE), OSTC Mk2 (SERIAL), OSTC Plus (SERIAL, BT, BLE), OSTC Sport (SERIAL, BT, BLE), OSTC cR (SERIAL)"
INFO: "Hollis: DG02 (SERIAL), DG03 (SERIAL), TX1 (SERIAL)"
INFO: "Mares: Airlab (SERIAL), Darwin (SERIAL), Darwin Air (SERIAL), Genius (SERIAL, BLE), Icon HD (SERIAL), Icon HD Net Ready (SERIAL), M1 (SERIAL), M2 (SERIAL), Matrix (SERIAL), Nemo (SERIAL), Nemo Air (SERIAL), Nemo Apneist (SERIAL), Nemo Excel (SERIAL), Nemo Steel (SERIAL), Nemo Titanium (SERIAL), Nemo Wide (SERIAL), Nemo Wide 2 (SERIAL), Puck (SERIAL), Puck 2 (SERIAL), Puck Air (SERIAL), Puck Pro (SERIAL, BLE), Quad (SERIAL, BLE), Quad Air (SERIAL, BLE), Smart (SERIAL, BLE), Smart Air (SERIAL, BLE), Smart Apnea (SERIAL, BLE)"
INFO: "Oceanic: Atom 1.0 (SERIAL), Atom 2.0 (SERIAL), Atom 3.0 (SERIAL), Atom 3.1 (SERIAL), Datamask (SERIAL), F10 (SERIAL), F11 (SERIAL), Geo (SERIAL), Geo 2.0 (SERIAL), Geo 4.0 (SERIAL, BLE), OC1 (SERIAL), OCS (SERIAL), OCi (SERIAL), Pro Plus 2 (SERIAL), Pro Plus 2.1 (SERIAL), Pro Plus 3 (SERIAL), Pro Plus X (SERIAL, BLE), VT 4.1 (SERIAL), VT Pro (SERIAL), VT3 (SERIAL), VT4 (SERIAL), VTX (SERIAL), Veo 1.0 (SERIAL), Veo 180 (SERIAL), Veo 2.0 (SERIAL), Veo 200 (SERIAL), Veo 250 (SERIAL), Veo 3.0 (SERIAL), Versa Pro (SERIAL)"
INFO: "Ratio: iDive Color Deep (SERIAL), iDive Color Easy (SERIAL), iDive Color Fancy (SERIAL), iDive Color Free (SERIAL), iDive Color Pro (SERIAL), iDive Color Reb (SERIAL), iDive Color Tech+ (SERIAL), iDive Deep (SERIAL), iDive Easy (SERIAL), iDive Fancy (SERIAL), iDive Free (SERIAL), iDive Pro (SERIAL), iDive Reb (SERIAL), iDive Tech+ (SERIAL), iX3M GPS Deep (SERIAL, BT), iX3M GPS Easy (SERIAL, BT), iX3M GPS Fancy (SERIAL, BT), iX3M GPS Pro  (SERIAL, BT), iX3M GPS Reb (SERIAL, BT), iX3M GPS Tech+ (SERIAL, BT), iX3M Pro Deep (SERIAL), iX3M Pro Easy (SERIAL), iX3M Pro Fancy (SERIAL), iX3M Pro Pro (SERIAL), iX3M Pro Reb (SERIAL), iX3M Pro Tech+ (SERIAL)"
INFO: "Reefnet: Sensus (SERIAL), Sensus Pro (SERIAL), Sensus Ultra (SERIAL)"
INFO: "Scubapro: Aladin H Matrix (BLE), Aladin Sport Matrix (BLE), Aladin Square (USBHID), Chromis (SERIAL), G2 (USBHID, BLE), G2 Console (USBHID, BLE), G2 HUD (USBHID, BLE), Mantis (SERIAL), Mantis 2 (SERIAL), Meridian (SERIAL), XTender 5 (SERIAL)"
INFO: "Seac: Guru (SERIAL), Jack (SERIAL)"
INFO: "Seemann: XP5 (SERIAL)"
INFO: "Shearwater: Nerd (SERIAL, BT), Nerd 2 (BLE), Perdix (SERIAL, BT, BLE), Perdix AI (BLE), Petrel (SERIAL, BT), Petrel 2 (SERIAL, BT, BLE), Predator (SERIAL, BT), Teric (BLE)"
INFO: "Sherwood: Amphos (SERIAL), Amphos Air (SERIAL), Insight (SERIAL), Insight 2 (SERIAL), Vision (SERIAL), Wisdom (SERIAL), Wisdom 2 (SERIAL), Wisdom 3 (SERIAL)"
INFO: "Subgear: XP-Air (SERIAL)"
INFO: "Suunto: Cobra (SERIAL), Cobra 2 (SERIAL), Cobra 3 (SERIAL), D3 (SERIAL), D4 (SERIAL), D4f (SERIAL), D4i (SERIAL), D5 (USBHID, BLE), D6 (SERIAL), D6i (SERIAL), D9 (SERIAL), D9tx (SERIAL), DX (SERIAL), EON Core (USBHID, BLE), EON Steel (USBHID, BLE), Eon (SERIAL), Gekko (SERIAL), HelO2 (SERIAL), Mosquito (SERIAL), Solution (SERIAL), Solution Alpha (SERIAL), Solution Nitrox (SERIAL), Spyder (SERIAL), Stinger (SERIAL), Vyper (SERIAL), Vyper 2 (SERIAL), Vyper Air (SERIAL), Vyper Novo (SERIAL), Vytec (SERIAL), Zoop (SERIAL), Zoop Novo (SERIAL)"
INFO: "Tecdiving: DiveComputer.eu (SERIAL, BT)"
INFO: "Tusa: Element II (IQ-750) (SERIAL), IQ-700 (SERIAL), Talis (SERIAL), Zen (IQ-900) (SERIAL), Zen Air (IQ-950) (SERIAL)"
INFO: "Uemis: Zurich (USBSTORAGE)"
INFO: "Uwatec: Aladin Air Twin (SERIAL), Aladin Air Z (SERIAL), Aladin Air Z Nitrox (SERIAL), Aladin Air Z O2 (SERIAL), Aladin Pro (SERIAL), Aladin Pro Ultra (SERIAL), Aladin Sport Plus (SERIAL), Memomouse (SERIAL)"
INFO: "Zeagle: N2iTiON3 (SERIAL)"
INFO: qqwindow screen has ldpi/pdpi 96.1263 141.951
INFO: "77.236: AppState changed to active with no save ongoing and no unsaved changes"
INFO: Window width changed to 800
INFO: Column width 800 root item width 800
INFO: width in Grid Units 47 original gridUnit 17 now 17
INFO: "79.049: AppState changed to inactive with no save ongoing and no unsaved changes"
```

dirkhh:
The log is mildly filtered for some of the noise, but contains all of the output from Subsurface
And yes, I hate having device only crashes. Is there an easy way to dump the name of the calling function? I wonder why suddenly on Android we see a duplicate instance - what's odd is that we get the report on the second instance before we get the one for the first. That could be random, of course, but it's consistent every time I run this...

dirkhh:
And yes, it works fine when running mobile on desktop

bstoeger:
> What's odd is that we get the report on the second instance before we get the one for the first. That could be random, of course, but it's consistent every time I run this...

Not surprising. Look at the code in `core/singleton.h`: It warns of generating the second instance before generating the instance. There should be some mention of generating the first instance. If not, then the instance pointer is not initialized to null...!? Is this a different compiler than mobile-on-desktop? There *are* some crazy optimizations with respect to variable initialization!

Could you test current master on device so that we know whether it is this PR or the PR with the singletons? Thanks.

bstoeger:
Check this out:
https://blog.tchatzigiannakis.com/undefined-behavior-can-literally-erase-your-hard-disk/

dirkhh:
Funny. Not.
And yes, Android is compiled with clang, desktop by default with gcc.
I will build current master and test on device.
Do you actually have an Android device?

dirkhh:
> > What's odd is that we get the report on the second instance before we get the one for the first. That could be random, of course, but it's consistent every time I run this...
> 
> Not surprising. Look at the code in `core/singleton.h`: It warns of generating the second instance before generating the instance. There should be some mention of generating the first instance. If not, then the instance pointer is not initialized to null...!? Is this a different compiler than mobile-on-desktop? There _are_ some crazy optimizations with respect to variable initialization!

```
09-26 09:41:14.704 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Generating second instance of singleton 8QMLPrefs
09-26 09:41:14.704 30300 30344 D /android/subsurface/qt-models/messagehandlermodel.cpp: INFO: Generated singleton 8QMLPrefs
```
by the way - since you twice asked about the mention of generating the first instance... you did see that they are in there back to back, right?

bstoeger:
Yes, I see that. But that's only one instantiation. See the code:
```
template<typename T>
SillySingleton<T>::SillySingleton()
{
        if (self)
                qWarning("Generating second instance of singleton %s", typeid(T).name());
        self = static_cast<T *>(this);
        qDebug("Generated singleton %s", typeid(T).name());
}
```
Could you send me a disassembly of `QMLPrefs::QMLPrefs`? Or even the contents of `qmlprefs.s`? I think this can be generating by passing -S to the compiler.

I have a Honor 8. But I'm pretty much a dinosaur when it comes to using smart phones.

dirkhh:
My point is that testing on device is really simple thanks to me making it really simple.
i.e. you can easily create builds and put them on your device to test
You may decide that you don't want to do that as you are already doing too many things (and I'd find that hard to argue with), but if you are interested, there's the option.
On the flip side because of the way I compile things, it may take me a moment to get you that disassembly... I'll try to find it and send it to you in a moment.

dirkhh:
yes, current master also crashes on device
no, current master compiled with clang does not crash on ArchLinux

dirkhh:
interestingly, I had the now discarded PR that included the singleton pattern built before you told me that this PR wasn't tested at all (I think that was #2299) and that did NOT crash on device

bstoeger:
Last time I tried compiling with `clang` on x86 I failed. But if you tell me how to generate an ARM(?)-executable, I'll try to look at the assembly myself.

dirkhh:
I'll send you an email with a script to do so locally

bstoeger:
> yes, current master also crashes on device

Well, then at least it's not this PR, which makes this somewhat less mysterious.

bstoeger:
I looked at this on godbolt and both gcc and clang produce correct code on x86-64. Perhaps this is a clang ARM64 miscompilation?

In any case, if it turns out to be #2297, it's probably best to just revert. I still like the idea, but not worth fighting for..

dirkhh:
I actually really like the idea in there and since as part of #2299 it didn't have that effect, I wonder what's different.
I sent you the huge disassembly (sorry, should have compressed that - 7MB email... let me know if you didn't get that). And my script for local builds

dirkhh:
oh, and my testing is all with arm64 as that's what is used on almost all modern phones

bstoeger:
> I actually really like the idea in there and since as part of #2299 it didn't have that effect, I wonder what's different.

That is indeed mind boggling.

> I sent you the huge disassembly (sorry, should have compressed that - 7MB email... let me know if you didn't get that). And my script for local builds

Got it - thanks.

bstoeger:
Ok - let's move the discussion here.

1) I checked the assembly using the -S option. Looks reasonable.
2) Installed current master with a few additional `printf`s. Runs just fine.

How can I have a look at the logs? I see the "copy logs to clipboard" option - but I don't see my `printf`s in there? Could it be that I somehow got the wrong version (it says 4.9.3.76)?

dirkhh:
I removed the Singleton commits for the latest beta. Need to revert from master as well.
Is this a version that you have compiled?
Printf you only see on adb logcat
Qdebug you can see on the app log
(Sorry. On my phone, back on the computer within the hour)

On September 27, 2019 12:09:58 PM PDT, bstoeger <notifications@github.com> wrote:
>Ok - let's move the discussion here.
>
>1) I checked the assembly using the -S option. Looks reasonable.
>2) Installed current master with a few additional `printf`s. Runs just
>fine.
>
>How can I have a look at the logs? I see the "copy logs to clipboard
>option - but I don't see my `printf`s in there? Could it be that I
>somehow got the wrong version (it says 4.9.3.76)?
>
>-- 
>You are receiving this because you commented.
>https://github.com/Subsurface-divelog/subsurface/pull/2303#issuecomment-536063315

-- 
From my phone

bstoeger:
Please wait with revert until I do some more debugging. There is something extremely ominous going on. When I log self before doing the check:
```
self: 0x0
Generating second instance of singleton 8QMLPrefs
Generated singleton 8QMLPrefs
self: 0x7defd16a80
Generating second instance of singleton 10QMLManager
Generated singleton 10QMLManager
```
1) Apparently `self` is 0 and yet the warning is emitted!
2) `QMLManager::self` is not 0. Something smashed the .bss segment!?

Perhaps we have a very different problem and this is only a symptom? BTW: Application runs fine for me with a few hundred dummy dives. But this is debug, not release.

dirkhh:
Wait what?
```
if (self)
                qWarning("Generating second instance of singleton %s", typeid(T).name());
```
and we print the warning with `self == 0`???
I know this is silly, but to avoid weird race condition issues... can you print self in the qWarning?
And of course - what does the assembler output look like????

bstoeger:
Oh jeez - Now I get it. The template code is totally wrong!

`this` points to the singleton sub-object. This is *not* at the beginning of the allocated object. Thus, when accessing via `instance()` you will get a pointer somewhere *into* the object.

I'm not sure how this produces the observed effect, but matter of fact this is totally broken.

Please revert and let's not talk about this anymore. :(

Well, on the positive side of things - at least I can now test on device..

dirkhh:
Stuff goes wrong. And you are normally so much better in understanding the intricacy of object hierarchies, I never doubted the code. It _looks_ right...
I'll revert. And I agree with you, the fact that you can now test on device is definitely a major positive side effect :-)

dirkhh:
So this whole discussion was odd topic for this PR. And this PR is independent of the singleton change and I believe this PR to be fundamentally correct. So I'm merging this.

bstoeger:
To my defense, it should have worked after all. Up-casting takes care of the fact that the subobject may not be located at the beginning of the derived object:
```
class B1 {
    int i;
    int j;
};
class B2 {
    int i;
    int j;
};
class D: public B1, public B2 {
};

D *upcast(B2 *b)
{
    return static_cast<D *>(b);
}
```
Compiles on ARM64 gcc to:
```
upcast(B2*):
  sub x1, x0, #8
  cmp x0, 0
  csel x0, x1, xzr, ne
  ret
```
I.e `return b ? b - 8 : 0`

Still, let's not waste more time with this.

bstoeger:
Moreover, it should have worked anyway, because an empty base class has the same address as the derived class.

Thus
```
class B1 {
    int i;
    int j;
};

class D;
class B2 {
    static D *d;
public:
    B2();
};

class D: public B1, public B2 {
    D();
};

B2::B2()
{
    d = static_cast<D *>(this);
}

D::D()
{
}

D *B2::d;
```
compiles to the perfect
```
B2::B2() [base object constructor]:
  adrp x1, .LANCHOR0
  str x0, [x1, #:lo12:.LANCHOR0]
  ret
D::D() [base object constructor]:
  adrp x1, .LANCHOR0
  str x0, [x1, #:lo12:.LANCHOR0]
  ret
B2::d:
  .zero 8
```

bstoeger:
Houston, we have a problem.

Because this is still bothering me, I changed the singleton code to:
```
template<typename T>
SillySingleton<T>::SillySingleton()
{
        qDebug("self1: %p", self);
        qDebug("self2: %p", self);
        if (self)
                qDebug("Generating second instance of singleton %s", typeid(T).name());
        self = static_cast<T *>(this);
        qDebug("Generated singleton %s at %p (%p)", typeid(T).name(), self, this);
}
```

This gives the output:
```
self1: 0x0
self2: 0x7de2e48d00
Generating second instance of singleton 8QMLPrefs
Generated singleton 8QMLPrefs at 0x7de207bee0 (0x7de207bee0)
self1: 0x7deea97a80
self2: 0x7deea97a80
Generating second instance of singleton 10QMLManager
Generated singleton 10QMLManager at 0x7de2f7f000 (0x7de2f7f000)
```
(!!)

bstoeger:
mobile on desktop gives the expected:
```
self1: 0x0
self2: 0x0
Generated singleton 13DiveListModel at 0x7fff2247cf50 (0x7fff2247cf50)
self1: 0x0
self2: 0x0
Generated singleton 12GpsListModel at 0x7fff2247cfa0 (0x7fff2247cfa0)
INFO: self1: 0x0
INFO: self2: 0x0
INFO: Generated singleton 8QMLPrefs at 0x55bc440c0550 (0x55bc440c0550)
```
@dirkhh: something is wrong with our android build.

buxtonpaul:
Is this a case of the static initialisation fiasco?
Most singleton implementations I have seen put the instance handle as a
static variable in the getinstance function rather than as a member
variable of the class, which I believe is to avoid the static variable
conflict.

bstoeger:
> Is this a case of the static initialisation fiasco?

I don't think so - after all we initialize the static variable to the `nullptr` constant. To my understanding the "static initialization order fiasco" only happens for global objects with constructors. Then, the call-order is undefined.

> Most singleton implementations I have seen put the instance handle as a
static variable in the getinstance function rather than as a member
variable of the class, which I believe is to avoid the static variable
conflict.

We do this likewise in many cases. The reason is that the compiler guarantees thread-safety for function-local static variable initialization.

Unfortunately, here we can't do this because the way we use QML, QML generates the object with `new`.

What is really striking is how after the first `QDebug` the `self` variable seems to be overridden!? Perhaps this is some multi-threading thing. Note how we don't see all the debug-messages. I'll try to see if we can initialize the logging earlier. Still - this makes our android crashes appear in a very different light. I'm somewhat concerned that all my recent mobile work was pointless because the problem is in the build itself. :(

dirkhh:
I don't think it was pointless. You cleaned up code that was highly suspect because of the way it used pointers to things that were potentially reallocated elsewhere. That was broken, period.
But the debug output of the self1 self2 version is really making me question my sanity...
Or the sanity of the compiler.
What happens if you assign `self` to a function local variable as the first line of the function.
Does that function local variable also change?

bstoeger:
```
template<typename T>
SillySingleton<T>::SillySingleton()
{
        T *self_local = self;
        static T *self_local_static = self;
        qDebug("1. self: %p, self_local: %p, self_local_static: %p", self, self_local, self_local_static);
        qDebug("2. self: %p, self_local: %p, self_local_static: %p", self, self_local, self_local_static);
        if (self)
                qDebug("Generating second instance of singleton %s", typeid(T).name());
        self = static_cast<T *>(this);
        qDebug("Generated singleton %s at %p (%p)", typeid(T).name(), self, this);
}
```
Gives:
```
1. self: 0x0, self_local: 0x0, self_local_static: 0x0
2. self: 0x7debe4ec80, self_local: 0x0, self_local_static: 0x0
Generating second instance of singleton 8QMLPrefs
Generated singleton 8QMLPrefs at 0x7de1e28d40 (0x7de1e28d40)
1. self: 0x7deeaa7a80, self_local: 0x7deeaa7a80, self_local_static: 0x0
2. self: 0x7deeaa7a80, self_local: 0x7deeaa7a80, self_local_static: 0x0
Generating second instance of singleton 10QMLManager
Generated singleton 10QMLManager at 0x7debfd6200 (0x7debfd6200)
```

This is getting more and more ominous! Apparently there is only one function-local static variable.

bstoeger:
Erm... In the second case, how can self_local_static be 0 when we just assigned a different value to it?

dirkhh:
So with this pattern wouldn't this code work?
```
{
        static T *self_local_static = nullpointer;
        // one of the next two lines should ensure that self is correct
        T *self_local = self;
        qDebug("consuming incorrect self %p, self);
        if (self_local_static != nullpointer && self_local_static != self)
                qWarning() << "called twice";
        self_local_static = self;

...
}
```
and then use that as designed for your singleton? Or am I missing yet another corner case?

Issue #2301, opened 2019-09-23
Subsurface-mobile 2.2.1(4.9.3.30): no menu


ghost:
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When I start the app, the menu shows itself just for a very brief time (probably less than a second) before the list of dives gets drawn over it.

### Operating system:

Android 8.1

bq Aquaris Pro

### Subsurface version:

Subsurface-mobile 2.2.1(4.9.3.30)

It did work before the update.

official release

### Steps to reproduce:

Open app

### Current behavior:

Menu only show for a very short time

### Expected behavior:

Menu shown constantly

dirkhh:
This is a new regression on a small number of devices that we are working on fixing.
It appears that as a workaround you can rotate your device sideways, wait for Subsurface-mobile to redraw the screen, and then rotate it back to the original position. So far for everyone encountering this problem that was sufficient to go back to a correctly drawn screen

ghost:
I can confirm that this works.

PR #2305, opened 2019-09-28
Combine all the recent work

dirkhh:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a combination of #2302 and all the work I had done to deal with the Android UI issues, plus a number of small build system changes and other random cleanup
Not very structured, but it's a very close match for the latest Android beta; this one has one merge commit left that I had flattened in my internal branch that I used to create the Android binary - that's why the version numbers don't match - but having that merge seemed to make things easier to figure out in the git history.

### Changes made:

1) @bstoeger 's fixes for the way we deal with the table of downloaded dives
2) some very hacky QML code that tries to be clever about the spurious screen size change that we get on some devices... I think it is related to the way I change the font and related information in certain situations, but without that Subsurface-mobile is unusable on certain small devices with large default fonts
3) a few random cleanups and build system modifications
4) a new version number for Subsurface-mobile

### Related issues:

#2292
#2295 
#2301 

### Mentions:

@bstoeger I'd appreciate a review :-/


dirkhh:
I think the only commit that isn't signed off is the merge commit. Hrmpf.

bstoeger:
I'll have a look later today.
Just note that the downloadTable thing was written before we realized that model-reset may cause a rendering of the dive-detail page for *all* dives. So this might need some tuning. Nevertheless, probably best to merge first and then optimize.

dirkhh:
As I said, the latest Android Beta is based on this branch. So this is getting real world testing.
I played with it on half a dozen of my devices and tested downloading on a couple of them - I didn't notice any horrible hangs for recalculating profiles...

mobile-widgets/qml/main.qml
```diff
@@ -552,8 +554,16 @@ if you have network connectivity and want to sync your data to cloud storage."),
 		property color contrastAccentColor: "#FF5722" // used for delete button
 		property color lightDrawerColor: "#FFFFFF"
 		property color darkDrawerColor: "#424242"
+		property int initialWidth: rootItem.width
 		Component.onCompleted: {
-			setupUnits()
+			// break the binding
+			initialWidth = initialWidth * 1
```
bstoeger, line 562: That is very obscure to someone not familiar with QML / JavaScript and might be worthy of a comment.
Is that a way to convert to int? How about `initialWidth + 0`?

subsurface-helper.cpp
```diff
@@ -61,8 +61,20 @@ void exit_ui()
...
-
 #ifdef SUBSURFACE_MOBILE
+#if defined(Q_OS_ANDROID)
+	if (getAndroidHWInfo().contains("/OnePlus/")) {
+		QFontDatabase db;
+		int id = QFontDatabase::addApplicationFont(":/fonts/Roboto-Regular.ttf");
+		QString family = QFontDatabase::applicationFontFamilies(id).at(0);
+		QFont newDefaultFont;
+		newDefaultFont.setFamily(family);
+		(static_cast<QApplication *>(QCoreApplication::instance()))->setFont(newDefaultFont);
```
bstoeger, line 73: That cast is ugly. If I read the docs correctly, you can either user `QApplication::instance()` or `qApp` without cast.

bstoeger (approved):
Only two minor comments - let's merge and proceed...

mobile-widgets/qml/main.qml
dirkhh: Actually, the comment is there. This is about breaking the binding in QML. Normally initialWidth would be "bound" to the value it got initialized with as a property - so whenever width changes, initialWidth would change. By making it the result of a computation, we are breaking this binding. I'm sure adding 0 would work the same.
https://doc.qt.io/qt-5/qtqml-syntax-propertybinding.html for more information

subsurface-helper.cpp
dirkhh: stupidly we have a qApp macro that conflicts with the "normal" way of doing this :-(
Yes, the right solution would be to change that macro. I'll add fixing that to my long list and merge as is (as you suggested).

dirkhh:
Thank you so much for taking the time to review, @bstoeger 
I always feel MUCH better if someone else takes another look.

mobile-widgets/qml/main.qml
bstoeger: > Actually, the comment is there. This is about breaking the binding in QML. Normally initialWidth would be "bound" to the value it got initialized with as a property - so whenever width changes, initialWidth would change. By making it the result of a computation, we are breaking this binding. I'm sure adding 0 would work the same.

Whaat? I don't think I want to understand that. :-P

subsurface-helper.cpp
bstoeger: > stupidly we have a qApp macro that conflicts with the "normal" way of doing this :-(

`git grep qApp` just shows a number of normal uses of `qApp`.

jaybny: is there a macro?

dirkhh: > is there a macro?

random question on an eight months old PR?

PR #2306, opened 2019-09-29
Cleanup: modify a few classes to use the classical singleton pattern

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is just a code cleanup and not strictly necessary. Some of our singletons are not allocated by QML. So instead of generating them by `new`, let's just use `instance()` function. Thus, they are generated at first use, which is the whole point of the singleton pattern. This makes the code more consistent.

### Related issues:

#2297

### Release note:

Not needed - just a minor code cleanup.
### Documentation change:

Not needed - just a minor code cleanup.

### Mentions:

@dirkhh


dirkhh (approved):
These all look good.
I can't wait to figure out the creative ways in which this will break stuff.

PR #2313, opened 2019-10-06
Update Subsurface-mobile to latest Kirigami release

dirkhh:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

We haven't updated Kirigami in ages - and it moved forward a LOT with monthly releases.
This picks up their last release and appears to get me to something that appears to work, but hasn't seen nearly enough testing.
I'm sure there are pages and actions that don't work as expected. Help would be very appreciated.

### Mentions:

@janmulder 
Jan, I think you are the main person who is likely to be able to help here
@bstoeger 
Berthold, I know how much you hate QML, but maybe you could play with this on device to see if I broke anything?
@glance- 
Anton, you are the other one who I know can easily build on Android and test and who has a very solid understanding of this code - if not exactly love for QML
@notmart 
Finally, Marco - not sure if you have time to look at projects USING your project. But of course, as always, any pointers and suggestions from you would be MOST welcome.


janmulder:
Some remarks to start with:

- I build it (arm64 style) myself, and I'm not using the wrapper build script (but only the normal build.sh), as I have all dependencies installed on my development machine. The first failure was due to not using the wrapper script.  I did not have the `./android-mobile/Roboto-Regular.ttf` file, that is copied over from the used NDK when using the script. But as I'm likely one of the very few self builders not using the docker stuff (by default), never mind ... 

- The annoying small white border around the menu is back.

- Cannot access the map.

- The breadcrumb line is not colored according to the used theme. and the forward/backward arrows are not there

- There are icons missing. The Back arrow in the menu does not show.

- I had no luck downloading over BT/BLE. My logcat shows 2 dives downloaded, but the app is frozen, the 2 downloaded dives never show in the app, and I have to force close it. Possibly related is the message  `qrc:/qml/main.qml:666:2: QML ScrollablePage: Binding loop detected for property "contentHeight`

dirkhh:
Yeah, I agree, the dependency on the wrapper script is a bit annoying. But at this point I think it has become fairly hard to set up a working build environment without that - so I felt that as long as that script was able to get things in the right places, that might be acceptable.
Any suggestions how to better document this?

I'll look into that white border - I think that may be intentional, but maybe I can figure out how to turn it off. I hadn't even noticed it.

I know that the breadcrumbs aren't quite behaving as intended - this has completely changed in its implementation on the Kirigami side - you know longer explicitly set a header, instead it's automatically provided by the `PageRow`. I hadn't even thought to test theming. Definitely something I need to work on. I remember seeing reference to the forward/backward buttons in the code, but you are right, they are missing.

The map works here, so I wonder what's different for you. What do you get in the AppLog when you try to access the map?

Missing icons - another thing that shouldn't be hard to track down.

While the binding loop is definitely something to address, I don't think that's the reason for the hang. I assume you get the accept button, you tap it but nothing appears to happen? That's a bug that I thought we fixed just a few days ago. I wonder if it somehow crept back / I forgot a commit when merging my Android work. If this is the bug you are encountering, then if you kill Subsurface-mobile and restart it, the downloaded dives should be there... can you check that?

Thank you SO MUCH for testing. I typically get into this point of "tunnel vision" when working on something like this. Lots of things are broken, when it finally seems to work, I can't come up with more things to test (I guess subconsciously I don't want to find more things that are broken?)

janmulder:
> Yeah, I agree, the dependency on the wrapper script is a bit annoying. But at this point I think it has become fairly hard to set up a working build environment without that - so I felt that as long as that script was able to get things in the right places, that might be acceptable.
> Any suggestions how to better document this?

Well, we have the file: `packaging/android/README` which looks like the place to put this, but that file is untouched since 2016, so that will definite need more work to make this more sensible.

> 
> I'll look into that white border - I think that may be intentional, but maybe I can figure out how to turn it off. I hadn't even noticed it.

That white border was hacked out earlier using one of these diff.patch scripts.

> The map works here, so I wonder what's different for you. What do you get in the AppLog when you try to access the map?

I see stuff like:

```
<Unknown File>:7: TypeError: Cannot read property 'contentX' of null
QPainter::begin: Paint device returned engine == 0, type: 3
QPainter::setCompositionMode: Painter not active
QPainter::end: Painter not active, aborted
```
But just found out that I can swipe to the map page. But clicking on the "map it" button, the map menu item, or the map icon in the bottom buttons, does not put focus on the map. There is some animation going on (like it seems that the map is selected in the `PageRow`, but the focus returns immediately, and no map is shown). It's very likely some `PageRow` navigational thing.

> 
> While the binding loop is definitely something to address, I don't think that's the reason for the hang. I assume you get the accept button, you tap it but nothing appears to happen? That's a bug that I thought we fixed just a few days ago. I wonder if it somehow crept back / I forgot a commit when merging my Android work. If this is the bug you are encountering, then if you kill Subsurface-mobile and restart it, the downloaded dives should be there... can you check that?

Never mind this one. I recompiled all, installed it again, paired the devices again, en all works now as intended. Really do not understand why it failed like that in the first place.

> 
> Thank you SO MUCH for testing. I typically get into this point of "tunnel vision" when working on something like this. Lots of things are broken, when it finally seems to work, I can't come up with more things to test (I guess subconsciously I don't want to find more things that are broken?)

Yes, tunnel vision it a common problem here when there is a plethora of small things :-)

buxtonpaul:
Hmm,
I  just sent a pull request that I thought was fixing an issue with the
Dark Theme, but from the comments here, it sounds like the dark theme
(before Kirigami updates) was working fine?
I was getting unreadable text due to it being black on black, is that only
me?

dirkhh:
@buxtonpaul dark theme never really worked. I don't use dark themes - no one else cared, so it stayed partially implemented. Patches VERY welcome.

dirkhh:
@janmulder I don't see errors like the ones you list when opening the map.
Correction, I do see the first one about line 7 of the unknown file - but not the others.
Is it possible that your googlemaps plugin wasn't recompiled for your Qt version?

That said, I found that a couple of my hacks to deal with the screen width issues are definitely wrong and it is possible that those might cause the map not being scrolled into view. I'll push an updated series, soon.

I'll keep plugging away on this :-)

dirkhh:
@notmart - any idea how I can change the background color of the default global tool bar? I've been searching for the answer for about a couple of hours now and went absolutely nowhere :-(

janmulder:
> Is it possible that your googlemaps plugin wasn't recompiled for your Qt version?

No, I installed new Qt 5.13.1, and saw that the maps plugin got compiled and installed correctly.

dirkhh:
Progress, of sorts.
I can now reproduce the "map not shown" issue - maybe that will allow me to then FIX that issue :-)
I also have re-enabled your old fix for the white border around the global drawer.
Still no idea why it doesn't show the back button - it's there, you can tap on it and it works. You just need to guess where it might be :-)

bstoeger:
Will test it tomorrow. What's the idiomatic way of updating my docker? Fresh download or should I simply use git inside the docker?

dirkhh:
When you run the build script it will pull the new container. Ironically the very things I'm doing to make the download for Travis smaller and faster prevent Docker from downloading only the new bits for you. On a slow link this 3.5GB download might be a bit painful. On a typical Gbit Internet connection it takes maybe 30 seconds.

dirkhh:
OK, the subject of that last commit is very true. That hack is _insane_
But it appears to work :man_shrugging: 
@notmart - I have now spent a 3 hour flight reading Kirigami sources and have played with every color setting that I could find - I still have a global tool bar with an un-themed background.
And I the forward / backward button icons don't show up in that tool bar as well. Which makes me think that I'm doing something wrong.
As usual, I of course don't have a simple example to reproduce this - do you happen to have code that themes that tool bar with a different color? Maybe that will point me in the right direction.
The examples unfortunately still appear to use the old ApplicationHeader which is now actually deprecated.

janmulder:
Well, let me refrain from any comment on the _insane hack_ commit ... but at my end, the map now shows. But, I immediately encounter something I do not like: open the app, select a dive in the dive list (its details are shown), hit "map it" (the map is shown). Now, using the breadcrumbs, go back to the dive list. There are now 3 pages on the stack (divelist, details of a dive, map). Now, select the dive again (in the divelist). expected to see the details again, but the map is shown instead. 

All said, I see more things related to the map page, but I did not try to bisect any of it, so I can't really tell where it got introduced.

- The selected dive site used to show up on the map with a different icon (more bright red than the not selected sites). This does not work any more. I suspect that this is not in this PR.
- On the map page, there is the small map menu that says "selected visible dive locations". This does not work any more, but this is not related to this PR.

dirkhh:
@janmulder the error case that you describe is something I thought I had tested and fixed. I'll try again - but that shouldn't be too hard to fix (err, hack around).
And I agree, several of the map features that used to work were broken at some point, but they appear broken in master before this PR. We should, however, attempt to fix them.
I'm traveling this week and will try to find time to continue to work on this.

dirkhh:
The last two commits should fix the issue reported by @janmulder when you switch to the dive list and then back to the dive.

dirkhh:
So it seems that at some point we broke the tracking of which dive is currently selected in the mobile UI - and with that we didn't highlight that dive flag anymore.
And also, when changing dives in the dive list, we didn't update the map at all.
Especially when using Subsurface-mobile on a tablet in landscape mode, this behavior now seems MUCH nicer.
@bstoeger can you check that I didn't make a mess of things with my selection code.
Also, there is this weird `selectedDiveTimestamp` thing that I don't think we use anywhere...
@janmulder please keep poking at this. I would love to make this more usable.

janmulder:
Ok, I see lots of improvement, but I find again something of which I do not know its related to this PR.

When I try to use the filter capability, the app freezes, Android trows a notification about the app not responding, I close the app, look at the logcat, and see numerous messages like `qrc:/qml/main.qml:666:2: QML ScrollablePage: Binding loop detected for property "contentHeight"`. Its seems some kind of endless processing loop.

And this said. I believe that the 2 issues mentioned before (theme setting, missing icons) are basically show stoppers. So, I truly hope @notmart will come up with something.

dirkhh:
And here is another brutal hack.
It seems that @notmart is busy (or he's ignoring us), so I figured out a truly brutal way to deal with the issue that the theme color keeps getting overwritten in the colorSync() slot in Kirigami. Instead of spending even more time trying to figure out how this is supposed to work (extensive googling shows absolutely nothing, can't find anything useful in the docs, etc), I'm simply hard-coding the subsurfaceTheme color that I want in the Kirigami code for the breadcrumbs toolbar. That is a terrible and stupid thing to do, but, oh well. For me this works.

So... I think the final thing left to fix (err, I mean the last thing on the list of things that we already KNOW are broken) is the back button icon...

dirkhh:
Hahahahahaaha.....
can you hear the hysterical laughter?
Yeah, well, I guess I have it mostly working.
For mobile-on-desktop all of the back/forward icons are shown with this branch, on device the ones in the toolbar show up now, the ones in the GlobalDrawer are missing... so still some more work needed
But ignoring just how hacky all this is, I think it's getting close to usable.
More testing would be extremely welcome.

@bstoeger - any input from you?

@notmart - now that I figured out how to hack around the issues, is that maybe enough to make it easier to help us figure out if this is an actual kirigami bug or if there is an idiomatic way to do this?

janmulder:
Yes, getting closer and closer. But what I'm really hesitant about, is, why upgrade Kirigami with the cost of this massive series of hacks, and no obvious benefits (other than keeping in sync with upstream).

And still some loose ends:

- the forward/backward arrows in the menu are still missing (re-reading, that is what Dirk already said)
- the breadcrumb line is higher than before (and a bit fat)
- there is some font size/type weirdness in the menu's "back" buttons. They look a little different than the other menu items

dirkhh:
> Yes, getting closer and closer. But what I'm really hesitant about, is, why upgrade Kirigami with the cost of this massive series of hacks, and no obvious benefits (other than keeping in sync with upstream).

The longer we delay, the harder it gets. And there's a chance that we'll help uncover Kirigami bugs.
But the real reason is that I hate working on new features for the mobile app against an old version of Kirigami. Typically things improve with newer versions.
 
> And still some loose ends:
> 
>     * the forward/backward arrows in the menu are still missing (re-reading, that is what Dirk already said)

Correct. I see them on Mobile-on-desktop, but not on Android. I'm hoping that I can figure this on out as well.

>     * the breadcrumb line is higher than before (and a bit fat)

That should be easy to fix.
 
>     * there is some font size/type weirdness in the menu's "back" buttons. They look a little different than the other menu items

I'm confused, I thought you can't see them. Or do you mean in the titlebar / toolbar?
Could you provide a screenshot to annotate?

janmulder:
> > * there is some font size/type weirdness in the menu's "back" buttons. They look a little different than the other menu items
> > ```
> 
> I'm confused, I thought you can't see them. Or do you mean in the titlebar / toolbar?
> Could you provide a screenshot to annotate?

Screenshot_20191009-075503_Subsurface-mobile

I mean the font size of the "back" text.

dirkhh:
Excellent. I hadn't even noticed.
All the other text comes from us, using our fonts, the "Back" text comes from Kirigami (in `GlobalDrawer.qml`)because we have a multi-level menu. 
In my experiments with Mobile on Desktop I see that switching font sizes doesn't fully propagate through the app until a restart, but once you restart, it seems to work.
But when playing with it on Android I see that this isn't the case there (which of course makes things harder to debug). 
I'll look into the options of getting all of the sizes consistent :-)

dirkhh:
I am open to accusations of insanity. This surely is NOT the way to do it. But. Umm. It works?
I now have Back entries in the GlobalDrawer on Android with correct font and correct icon

janmulder:
Obviously, I can still find something :-)

- Not yet addressed: the breadcrumb line is higher than before (and a bit fat)
- There is no forward arrow in the global menu for items that have a sub-menu

dirkhh:
> Obviously, I can still find something :-)

I would be disappointed if you didn't

>     * Not yet addressed: the breadcrumb line is higher than before (and a bit fat)

Yes. I am working on this as we speak but am running into something else that looks like a Kirigami bug. Because I can easily get the toolbar to be the same hight it was before, but at least on Android the font size isn't adjusted (even though the code to do so is clearly in there - since @notmart doesn't seem to respond here, I've reached out in email to see if I can get some help)

>     * There is no forward arrow in the global menu for items that have a sub-menu

Ugh. That works on Mobile-on-Desktop. I bet it's the same issue that I'm hacking around for the back button. Somehow on Android we don't find some of the built-in kirigami icons when accessing them from Kirigami code. I need to figure out the ACTUAL issue instead of just hacking around that. But I think this is something that I can fix... it's just so much more painful to test things that I cannot reproduce with Mobile-on-Desktop...

mobile-widgets/qmlmanager.cpp
```diff
@@ -1375,6 +1375,15 @@ bool QMLManager::undoDelete(int id)
 	return true;
 }
 
+void QMLManager::selectDive(int id)
```
bstoeger, line 1378: `amount_selected` doesn't seem to be used on mobile. Nevertheless, I'd suggest that this function keeps it up to date to avoid surprises.

Something like:
```
	amount_selected = 0;
	for_each_dive (i, dive) {
		dive->selected = (dive->id == id);
		if (dive->selected)
			++amount_selected;
	}
```

I can't comment on the rest of the PR, since I don't understand it.

BTW: Would it be possible to change the details-view so that only one DiveObjectHelper object is generated per page load? I don't understand that property thing.

dirkhh: I think you are exaggerating if you say you  don't "understand" the rest of the PR. It's really not that hard - just awkward and backwards.

As to your code snippet. That seems a somewhat convoluted way to set `amount_selected = 1` since we sure hope that those `id`s are unique and only one is passed in, right?

And limiting the number of `DiveObjectHelper`s generated per page load... right now I'm mostly aiming for "have this stupid thing work at all". I agree that it would nice to have it also be efficient, but one miracle at a time, please.

bstoeger: > As to your code snippet. That seems a somewhat convoluted way to set amount_selected = 1 since we sure hope that those ids are unique and only one is passed in, right?

Unless an unknown id is passed in. Shouldn't happen, but who knows...

bstoeger: For the sake of robustness you could even add a
```
	if (amount_selected == 0)
		qWarning("QManager::selectDive(): unknown id")
```
or similar at the end of the function.

dirkhh:
Famous last words, but I think with this latest push I have all the outstanding issues addressed.
- all the next/previous icons seem to work
- the titlebar correctly shrinks and the font size and alignment in the title bar appears correct

@janmulder - here's my challenge to you to find more things that are broken.

Once we think that this is fine, I can get started on what I was intending to work on last week. A couple new features for the mobile UI.

janmulder:
> here's my challenge to you to find more things that are broken.

Tried to test all things I mentioned before, and this time I do *not* find anything obliviously broken. So, for me, things are ready to merge.

I do see 1 thing I've not seen before. Have 3 pages open (divelist, details, map), go to details, and flip trough some pages. Sometimes there is edge of the map visible, before showing the (intended) details of the next/prev dive. This is a minor visible glitch not worth fixing (and I'm almost convinced that this is something in Kirigami, and likely not simple to fix).

PR #2314, opened 2019-10-06
Windows build: Changesto build using mxebased hidapi,libusb and grantlee

buxtonpaul:
Remove hidapi, grantlee,qtconnectivity  and libzip from manually built components and use the mxe based ones instead.

Use recently added grantlee package from MXE (this is based on the last released grantlee plus the changelist)
that allows it to build with qt5.13.1
Remove libzip and qtconnectivity as these are handled by mxe packages.

Also hide vscode files from git as they should not be tracked.

Signed-off-by: Paul Buxton <paulbuxton.mail@googlemail.com>

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

I added grantlee to the mxe project based on the last official releast of grantlee + the changelist that allows it to build on qt 5.13.1
We also use hidapi, libzip and qtconnectivity from the mxe build as well.

### Additional information:

This will need a container rebuild


dirkhh:
Haven't looked at this - but the reason we use our own QtConnectivity is that we need the branch with the Windows/BLE changes - and I thought they won't show up until Qt 5.14 in upstream. Are they in Qt 5.13.1 already?
Without those changes BLE won't work on Windows.

buxtonpaul:
Ah, didn't spot that! I will check in case that is in 5.13.1, but likely I
will revert the qtconnectivity change.

dirkhh:
A bit of Subsurface / GitHub philosophy that Dawn might appreciate (or not... :-) )
Some projects squash PRs into a single commit - which loses all of the commit history and a lot of IMHO very important information when you later try to bisect issues. Subsurface doesn't do that, so when we merge PRs, we typically merge them as they are.
Your PR now introduces a change and then undoes that change in the immediate next commit. The preferred way for the PR author to deal with this is to squash the two commits in your branch (since the second one isn't useful in understanding the sequence of changes, it simply removes an error introduced in your PR). This puts a little more onus on the contributor, but in the end I think it creates a better project with a more meaningful git history.

Having said all this, I'll simply do this myself and push the changes manually instead of waiting for another back and forth with you. But please take this as food for thought for next time (and I am hoping for many more contributions from you!)

buxtonpaul:
Thanks. Duly noted, I hadn't seen any specific references to if you prefer
squashed commits or not, I know some projects strongly prefer things one
way or th other

Don't merge these yet!

I am looking at the build some more, as when I backtracking the
qtconnectivity change I am failing to get a container to build. I think
that it is due to the fact we are taking the head of the qtconnectivity
win/wip branch rather than a fixed version. I will try and verify that, it
is slow progress as rebuilding containers takes a loooong time :-)

I am also spending a bit of time brushing up on my Qt and looking into how
QML works. My attempts to make the combo-boxes in the dark theme work are
also being problematic.

Paul

dirkhh:
That was entirely my fault - I merged this without thinking through that I needed to actually build that container, first. I'll revert the commit and wait for the go-ahead from you.
In the meantime you can see in master the changes that I did to your two commits and especially the commit message. Maybe that's useful.

buxtonpaul:
Yes that is useful thanks.

Annoyingly both the head of mxe and qtconnectivity are broken. I have
identified the revisions we should be using and am just checking it builds
now.

dirkhh:
The head of MXE appears to be broken more often than not. That's why I tend to pick a commit that works for me and hard-code it. qtconnectivity is a different topic - maybe @neolit123 can help pick a good commit there.

buxtonpaul:
I have gone back to the one which I think you used for the previous
container build (based on what would have been the head on that date).
Was the mxe version hard coded? It looked like the change to make it
reusable also made it use the head by default.
I am changing this to use a known value for both mxe and qtconnectivity but
allowing a build time override, so we know that anyone trying to use the
container locally will get the same version unless the specifically
override it.

dirkhh:
please look at scripts/docker/mxe-build-container/Dockerfile to see how the MXE version is picked. 
The logic of making the version an argument to the Dockerfile is nice for testing different version, but then it's easy to lose track of the one ultimately in play.
My local scripts seem to show that I used 9f6b9c6 for the latest container build. That should correspond to this commit: https://github.com/mxe/mxe/commit/9f6b9c6f58510bedfa0bf9f87de9a214abe6b653

neolit123:
> Haven't looked at this - but the reason we use our own QtConnectivity is that we need the branch with the Windows/BLE changes - and I thought they won't show up until Qt 5.14 in upstream. Are they in Qt 5.13.1 already?
Without those changes BLE won't work on Windows.

indeed, i see the changes that the Subsurface/Windows BLE backend needs in the 5.14 branch, but not in 5.13.

> The head of MXE appears to be broken more often than not. That's why I tend to pick a commit that works for me and hard-code it. qtconnectivity is a different topic - maybe @neolit123 can help pick a good commit there.

a commit that can be used is:
https://github.com/qt/qtconnectivity/commit/8b7b52d66f2616040ca4aaae3f2732be96e19ab8

it still requires configuration:
> The use of the Win32 backend must be selected by the -native-win32-bluetooth configuration option.

scripts/get-dep-lib.sh
```diff
@@ -15,7 +15,7 @@ CURRENT_LIBXML2="v2.9.4"
 CURRENT_LIBFTDI="1.3"
 CURRENT_KIRIGAMI="8691651c1f0d09430416ba5fe2130647554b06a9"
 CURRENT_BREEZE_ICONS=""
-CURRENT_GRANTLEE="v5.0.0"
+CURRENT_GRANTLEE="master"
```
neolit123, line 18: i don't think we should pin to master, but instead to a stable tag.
v5.0.0 works for us and we shouldn't move away from it if possible.

packaging/windows/mxe-based-build.sh
```diff
@@ -20,14 +20,15 @@
...
+# # note this currently doesn't build on mxe version @180304a
 # # qtbase_CONFIGURE_OPTS=-debug-and-release
 #---
 # (documenting this in comments is hard... you need to remove
 # the first '#' of course)
 #
 # now you can start the build
-#
-# make libxml2 libxslt libusb1 libzip libssh2 libftdi1 curl qt5 nsis
+# look at the setting.mk ffrom scripts/docker/mxe-build-container/settings.mk
```
neolit123, line 30: typo in ffrom.

scripts/get-dep-lib.sh
buxtonpaul: Agree, will remove this.

packaging/windows/mxe-based-build.sh
buxtonpaul: Thanks

buxtonpaul:
> 9f6b9c6

So the Dockerfile as it currently stands doesn't pick a specific commit, it defaults to master and you have to build it specifying the in the docker build command in order to override that.

What I am proposing is changing the dockerfile to specify specific revisions, but allow someone doing local development to specify an override when they build the container, but if they don't it should use the same versions as the tagged container that the Travis builds are using.

Also moving to the mxe revision you suggest will not have the updated Grantlee and hidapi builds in the mxe toolchain, which was the original point of this :-)

neolit123:
> Also moving to the mxe revision you suggest will not have the updated Grantlee and hidapi builds in the mxe toolchain, which was the original point of this :-)

are there important fixes in Grantlee or hidapi that we want?

buxtonpaul:
Not fixes.

Simply trying to reduce build time by moving things from the build script
to the container. I am not emotionally attached to this, so if we decide it
is more effort than it is worth I have no issues with scapping it.To some
extent this was my way of getting a handle on how the builds are done.

Paul

neolit123:
i see the potential for exposing such functionality as pluggable in the bash scripts.

i.e. if someone wants to do it locally they can use the prebuild packages from mxe, else the default option (to not change the current behavior) to build from source at a fixed commit.

dirkhh:
Of course, completely unintentionally I've thrown you a curveball by moving the test build to GitHub Actions. The scripts are extremely similar - so while this branch can no longer be used, I would appreciate if you continued your work on addressing the underlying issue.
Qt 5.14 Beta 2 should come out in a day or two... and the release maybe a month later. Hard to say when it will be in MXE, but my guess is that it won't be too long after that. So whatever we do today will likely need to get updated again once that is out. As @neolit123 points out, 5.14 will bring his qtconnectivity changes into the release and resolve the need to have our own qtconnectivity build.

buxtonpaul:
No worries :-)

dirkhh:
Hey @buxtonpaul  -- any updates? We have a much better build environment for the containers now, so maybe this will be much easier now?

buxtonpaul:
I will try and have a look at this soon. Just of on holiday then away for
Christmas, so likely in the new year I will try and spend more time on
development

dirkhh:
Still hoping that @buxtonpaul will come back to this and start actively maintaining our Windows build infrastructure... new versions of key libraries and all that good stuff :-)

buxtonpaul:
I will try and get back to this 😁

buxtonpaul:
Ok looking through this, and the current versions of things.
The head of MXE looks to be updated to 5.14.2
It contains
hidapi @ the current head of hidapi
grantlee @ 5.2.0
qtconnectivity from qt 5.14.2
libzip @ 1.5.2

So I should be able to remove the fetch of these from the windows build before_isntall.
Update the version of MXE we use, and ensure these packages are in the windows container, unless anyone has any concerns?

dirkhh:
That sounds like the right overall direction... way back when I started all this, the versions in MXE had issues (or were missing like hidapi). So what we REALLY need is someone to update, get this working, and then stick around to maintain it... :-)

buxtonpaul:
Closing this as the newer one would be the better place to continue any discussion.

PR #2318, opened 2019-10-09
Cloud storage: remove outdated certificate hack

dirkhh:
The old server certificates where not recognized on some older platform,
so we hardcoded the hex digest of the valid certificate and ignored the
error.

Those certificates have been replaced last week, so there is no point to
this hack anymore - also, we should always show the SSL error, not just
in verbose mode.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

We no longer use that certificate, so no point to this anymore

### Changes made:

1) remove the override for the certificate that's no longer in use
2) always show the SSL errors, not just in verbose mode

### Mentions:

This should be an obvious change, but I'd appreciate a quick review, @bstoeger or @atdotde or @neolit123

### DIFF
--- a/core/checkcloudconnection.cpp
+++ b/core/checkcloudconnection.cpp
@@ -78,24 +78,9 @@ bool CheckCloudConnection::checkServer()
 
 void CheckCloudConnection::sslErrors(const QList<QSslError> &errorList)
 {
-	if (verbose) {
-		qDebug() << "Received error response trying to set up https connection with cloud storage backend:";
-		for (QSslError err: errorList) {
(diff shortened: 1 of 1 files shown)


neolit123:
SGTM, i've restarted the failed Mac and 5.7 builds.

neolit123:
i think the mac job is just timing out, while i do not understand why the 5.7 job is red. :thinking: 
merging based on the other jobs.

PR #2320, opened 2019-10-14
Keep text field that user wants to edit visible

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

One of my main frustration editing dives on mobile has always been that I tap on a field, the keyboard pops up and hides the field I wanted to edit. Grmbl.
I figured out how to prevent that.
Testing this ONLY makes sense on device - you can't really see the effect when running Mobile-on-Desktop

### Changes made:

1) create our own TextField that sets some of the repetitive defaults (because of this the net lines added by this PR is tiny)
2) create a timer that triggers 300/600ms after we tap on the text field and makes sure that the text field is indeed visible
3) redo those changes for the TextArea on the DiveDetailsEdit page (we have only one, so I didn't create a new type for that)

### DIFF
--- a/mobile-widgets/qml/DiveDetails.qml
+++ b/mobile-widgets/qml/DiveDetails.qml
@@ -187,7 +187,8 @@ Kirigami.Page {
 
 	actions.main: Kirigami.Action {
 		icon {
-			name: state !== "view" ? ":/icons/document-save.svg" : ":/icons/document-edit.svg"
+			name: state !== "view" ? ":/icons" + subsurfaceTheme.iconStyle + "/document-save.svg" :
+						 ":/icons" + subsurfaceTheme.iconStyle + "/document-edit.svg"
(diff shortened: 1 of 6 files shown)


dirkhh:
Oh, and that other commit to fix the edit/save icon.
I'm not even sure when this broke, TBH. I'm thinking it may have been a new version of the Breeze icons where the foreground was no longer white?

bstoeger:
Weird compilation error on linux2 Travis. I reckon that is a now unsupported Qt-version for mobile..?

dirkhh:
Exactly. That's what got me to do the other PR

dirkhh:
Since I don't think any of the other maintainers really feels comfortable reviewing QML code... let's merge this.

PR #2321, opened 2019-10-14
Stop supporting Qt < 5.9.1 -- require Qt 5.11 for mobile builds

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Having moved the AppImage to 5.12 there really wasn't much point in supporting anything prior to 5.9.1 - picking that one gives us BLE everywhere. And anyone on a distro that doesn't have 5.9.1 should probably update their distro before complaining to us

Additionally, one thing I didn't notice when switching to Kirigami 5.62 in #2313 was that that requires Qt 5.11. Which is fine since we build our iOS and Android app on fairly current Qt, anyway, in order to get all of their recent bug fixes

Let's see how many things I broke by doing this...


bstoeger (approved):
I can't judge how relevant Qt < 5.9 is in the grand scheme of things, but this certainly makes things more convenient.

So approval from me.

dirkhh:
The Mac build is known to be broken right now - so I'll take the rest as a go ahead...
I'm sure there will be plenty of complaining once this hits master

mturkia:
I just noticed that Subsurface cannot be built anymore in Ubuntu 18.04 LTS (EOL 2024). So just checking if we really need to enforce this version? @dirkhh 

Of course, AppImage can be used on those older releases, but it would be nice if we could provide deb packages on all version of Ubuntu that are still supported. If not, we need to update Subsurface download page to state the fact as we still recommend our PPA for xenial and Mint 18 in there.
 
```
CMake Error at CMakeLists.txt:261 (find_package):
  Could not find a configuration file for package "Qt5" that is compatible
  with requested version "5.9.1".

  The following configuration files were considered but not accepted:

    /usr/lib/x86_64-linux-gnu/cmake/Qt5/Qt5Config.cmake, version: 5.5.1
```

dirkhh:
It's 16.04 we don't support anymore, which even back when I pushed the change had (based on what I see on the cloud server) basically no users.

We pushed to require higher versions because we wanted to use some more modern methods. But I don't recall how critical that was. I just didn't expect a lot of people to stay on an outdated desktop version...

PR #2325, opened 2019-10-17
Planner: copy deco state before passing it to worker thread

bstoeger:
The planner has a computeVariations() function that can be run
in a worker thread. The code was not thread safe: a deco_state
object allocated on the stack of the caller was passed down to
the worker thread. It's well possible that the object would go
out of scope before the thread run.

Therefore, when running in the background, copy the object first
and free it in the worker thread.

Side note: Qt makes proper memory management again as difficult
as possible: You can't pass a std::unique_ptr<> to QtConcurrent::run,
because move-only objects are not supported. Not very friendly!

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When (re)planning a dive I got a valgrind alert:
```
==6818== Invalid read of size 8
==6818==    at 0x399E6F: DivePlannerPointsModel::computeVariations(diveplan*, deco_state*) (diveplannermodel.cpp:1016)
==6818==    by 0x39EEF4: non-virtual thunk to QtConcurrent::RunFunctionTask<void>::run() (qtconcurrentrunbase.h:140)
==6818==    by 0x9712BD0: ??? (in /usr/lib/x86_64-linux-gnu/libQt5Core.so.5.12.2)
==6818==    by 0x970F611: ??? (in /usr/lib/x86_64-linux-gnu/libQt5Core.so.5.12.2)
==6818==    by 0x4C89181: start_thread (pthread_create.c:486)
==6818==    by 0x9FE3B1E: clone (clone.S:95)
==6818==  Address 0x1ffeffde20 is on thread 1's stack
==6818==  1248 bytes below stack pointer
```

If I'm reading the code correctly, then valgrind is correct - the code is not thread safe. I suppose that `QtConcurrent` runs the member function independently in a different thread. If this is not the case, this PR can be rejected.

This is a quick-fix which tries to not touch other code-paths. With this, I don't get the alert anymore.

### Changes made:

1) Copy deco_state and free in worker-thread.

### Release note:

No - too obscure.

### Documentation change:

No.

### Mentions:

@atdotde: You might want to check this out.

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -919,7 +919,10 @@ void DivePlannerPointsModel::createTemporaryPlan()
 		cloneDiveplan(&diveplan, plan_copy);
 		unlock_planner();
 #ifdef VARIATIONS_IN_BACKGROUND
-		QtConcurrent::run(this, &DivePlannerPointsModel::computeVariations, plan_copy, &plan_deco_state);
+		// Since we're calling computeVariations asynchronously and plan_deco_state is allocated
+		// on the stack, it must be copied and freed by the worker-thread.
+		struct deco_state *plan_deco_state_copy = new deco_state(plan_deco_state);
+		QtConcurrent::run(this, &DivePlannerPointsModel::computeVariationsFreeDeco, plan_copy, plan_deco_state_copy);
 #else
 		computeVariations(plan_copy, &plan_deco_state);
 #endif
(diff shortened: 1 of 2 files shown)


atdotde:
The observation is certainly correct. An alternative solution would have been to call computeVariations always on a freshly malloc'ed deco_state and let itself free it in the end.

Thumbs up from me.

bstoeger:
> An alternative solution would have been to call computeVariations always on a freshly malloc'ed deco_state and let itself free it in the end.

Yes, that was my first attempt. I made it properly by using `std::unique_ptr` and then discovered that `QtConcurrent::run()` does not support move-only types. So I said whatever and went with the least intrusive route that touches the caller as little as possible.

dirkhh (approved):
This looks like a reasonable solution to the problem to me.
The build failure is a known Travis issue which is gone in master (since that Travis build has been removed there)

dirkhh:
I so wish GitHub would auto-create a compare link for force-pushes. Using  https://github.com/Subsurface-divelog/subsurface/compare/21abf1e..aa355d5 it's nice to see the fixes to the comments that you pushed :-)

bstoeger:
Two approvals and it fixes a subtle potential use after free problem - let's merge.

PR #2327, opened 2019-10-20
avoid creating hundreds of DiveObjectHelpers

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is an odd one. Not sure how to classify this.
It duplicates functionality from the DiveObjectHelper in the divelistmodel.
Instead of creating hundreds of DiveObjectHelpers with all the data of the dive, only to use one element of that and then discard it, return that one element directly from the model.
This adds a ton more roles and duplicates code, but wow does it reduce the number of DiveObjectHelpers that are being created. By seriously a factor of ten! ~130 per dive to 12 per dive.

Of course this raises the question... is the DiveObjectHelper maybe the wrong thing to use from QML, period?

### Mentions:

@bstoeger I'd love to hear what you think. And "this is totally bogus and stupid" is a valid response :-)

### DIFF
--- a/core/subsurface-qt/DiveObjectHelper.cpp
+++ b/core/subsurface-qt/DiveObjectHelper.cpp
@@ -10,6 +10,14 @@
 #include "core/subsurface-string.h"
 #include "qt-models/tankinfomodel.h"
 
+#if defined(DEBUG_DOH)
+#include <execinfo.h>
+#include <stdio.h>
+#include <unistd.h>
+static int callCounter = 0;
+#endif /* defined(DEBUG_DOH) */
+
+
 enum returnPressureSelector {START_PRESSURE, END_PRESSURE};
 
 static QString getFormattedWeight(const struct dive *dive, unsigned int idx)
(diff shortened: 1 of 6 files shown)


bstoeger:
I believe(!) that indeed QML should access the model directly instead of going via the DiveObjectHelper. But as you know, I don't "get" QML, so take my opinion with a grain of salt. DiveObjectHelper is still needed for Grantlee printing.

Just one quick remark: the call-counter can be made static, isn't it?

dirkhh:
> I believe(!) that indeed QML should access the model directly instead of going via the DiveObjectHelper. But as you know, I don't "get" QML, so take my opinion with a grain of salt. 

I would like to say that you and I constantly learn more about QML. And given that we don't have any QML experts participating in this project, I appreciate having someone check my logic.
There are a few more calls to the DiveObjectHelper from QML and I will fix those today or tomorrow. Those will bring more code duplication (and I was getting tired last night), so I pushed this as RFC to get feedback from you.
Of course my room neighbors headed out at 4am (i.e., now) this morning and have been up making lots of noise for an hour, so I had more opportunity to look at this PR and figured out a completely unrelated logic flaw in the new GitHub Action code. I pushed a fix to this branch, but I'll merge that independently and take it out of this branch again once I'm fully awake :-)

> DiveObjectHelper is still needed for Grantlee printing.

Yes, of course. I wasn't planning to suggest that it should be removed - just that the way we used it from QML caused potential performance issues. But I see that the PR description is poorly phrased. I'll update that in a moment.

> Just one quick remark: the call-counter can be made static, isn't it?

Of course. Initially this was just going to be a quick hack, but then I realized that checking it in for future reference might be worth it.

PR #2336, opened 2019-10-25
Mares: add BT discovery logic

torvalds:
We don't have the "show all dive computers" logic on mobile, so we need
something like this.

Possibly we should use the libdivecomputer matching code if it exists,
but that's a much bigger change, let's do this incremental one for now.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -71,11 +71,14 @@ static dc_descriptor_t *getDeviceType(QString btName)
 		if (btName.startsWith("Aladin")) product = "Aladin Sport Matrix";
 	}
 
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
Ugly but useful. We have this problem with several vendors.

dirkhh:
And since we currently don't get builds since I'm an idiot, let's just merge as this looks soooo obvious,

zzz6519003:
"Genius"? lol

PR #2340, opened 2019-10-26
First couple of Coverity cleanups

dirkhh:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Both fixes should be super obvious

### Additional information:

The second commit raises some questions that need investigation. Why do we have two different out of bounds values here. Is this intentional? If yes, can we please verify that every use is indeed correct and add a bit more commenting? If not, can we please consolidate?

### Release note:

none

### Documentation change:

none

### Mentions:

@atdotde @bstoeger @sfuchs79 your comments on the second commit would be welcome.

### DIFF
--- a/core/parse.c
+++ b/core/parse.c
@@ -33,7 +33,6 @@ void free_parser_state(struct parser_state *state)
 	free_picture(state->cur_picture);
 	free((void *)state->cur_extra_data.key);
 	free((void *)state->cur_extra_data.value);
-	free((void *)state->cur_settings.dc.nickname);
 	free((void *)state->cur_settings.dc.model);
 	free((void *)state->cur_settings.dc.nickname);
 	free((void *)state->cur_settings.dc.serial_nr);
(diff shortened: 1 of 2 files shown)


bstoeger:
Suggestion:
Define `UNDEF_COMP_TYPE` to `-1` and add the comment from `qt-models/diveplannermodel.cpp` to `core/divemode.h`:
```
// UNDEF_COMP_TYPE is an internal flag meaning "take the dive mode from the previous planner point".
```

Or something to that regard.

bstoeger (approved):
Both changes appear to be obviously correct. I'm surprised that the double-free didn't lead to crashes / debug messages.

dirkhh:
I'll merge as is, but would like @atdotde to please look at @bstoeger 's suggestion above.

PR #2341, opened 2019-10-26
Coverity fixes

dirkhh:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I triaged a few more - some false positives marked as such, and three reasonably obvious fixes.

### Mentions:

@bstoeger - the last one was introduce by you, I think. Can you check my logic just in case I'm missing something subtle here?

### DIFF
--- a/core/ostctools.c
+++ b/core/ostctools.c
@@ -44,7 +44,8 @@ void ostctools_import(const char *file, struct dive_table *divetable, struct tri
 	FILE *archive;
 	device_data_t *devdata = calloc(1, sizeof(device_data_t));
 	dc_family_t dc_fam;
-	unsigned char *buffer = calloc(65536, 1), *uc_tmp;
+	unsigned char *buffer = calloc(65536, 1);
+	unsigned char uc_tmp[2];
 	char *tmp;
 	struct dive *ostcdive = alloc_dive();
 	dc_status_t rc = 0;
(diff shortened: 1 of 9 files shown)


core/ostctools.c
```diff
@@ -97,7 +97,6 @@ void ostctools_import(const char *file, struct dive_table *divetable, struct tri
 	// Read dive's raw data, header + profile
 	if (fseek(archive, 456, 0) == -1) {
 		report_error(failed_to_read_msg, file);
-		free(uc_tmp);
```
bstoeger, line 100: 1) Obviously correct, but that raises the question why `uc_tmp` is allocated on the heap in the first place
2) For better encapsulation, the `free(ostcdive)` calls should probably be `free_dive(ostcdive)`

core/parse.c
```diff
@@ -246,12 +246,13 @@ void dive_end(struct parser_state *state)
 {
 	if (!state->cur_dive)
 		return;
-	if (!is_dive(state))
+	if (!is_dive(state)) {
 		free_dive(state->cur_dive);
-	else
+	} else {
```
bstoeger, line 251: Yes, that fixes a subtle bug.

core/ostctools.c
dirkhh: I think I implemented both of these suggestions in my force push.
I'd appreciate another look :-)

dirkhh:
now transfer.sh is stalling. I'll merge this and fix the PR artifacts. Grmbl

PR #2342, opened 2019-10-27
GitHub Actions: switch back to the GitHub artifacts for PRs

dirkhh:
transfer.sh has suddenly started to time out. And for the rather
occasional need of having a binary to test from a PR, this should be
good enough.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

It was great while it lasted.
Facepalm
This is the official GitHub mechanism. It's not super convenient, but it should work well enough

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -39,10 +39,15 @@ jobs:
         BIN1: ./Subsurface-mobile.apk
         BIN2: ./Subsurface-mobile.arm64.apk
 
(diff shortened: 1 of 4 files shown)


dirkhh:
OK, this worked and the artifacts are available on the individual build's Details page.
I think that's even easier than what we used to do with transfer.sh.
merging

PR #2345, opened 2019-10-27
Don't expose 'detach_buffer()' to membuffer users

torvalds:
The native buffer of a membuffer is not NUL-terminated, so when you want
to detach it and use it as a C string, you had to first do
'mb_cstring()' that adds the proper termination/

This was all documented in the header files, and all but two users did
it correctly.

But there were those two users, and the exported interface was
unnecessarily hard to use.  We do want the "just detach the raw buffer"
internally in the membuffer code, but let's not make the exported
interface be that hard to use.

So this switches the exported interface to be 'detach_cstring()', which
does that 'mb_cstring()' for you, and avoids the possibility that you'd
use a non-terminated memory buffer as a C string.

The old 'detach_buffer()' is now purely the internal membuffer
implementation, and not used by others.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/errorhelper.c
+++ b/core/errorhelper.c
@@ -22,8 +22,7 @@ int report_error(const char *fmt, ...)
 		return -1;
 
 	VA_BUF(&buf, fmt);
-	mb_cstring(&buf);
-	error_cb(detach_buffer(&buf));
+	error_cb(detach_cstring(&buf));
 
 	return -1;
 }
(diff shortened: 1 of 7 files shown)


dirkhh (approved):
This looks good.
It won't make Coverity happy, but that's ok because I can just mark that as "ignore".
Waiting for the compiles to finish

PR #2347, opened 2019-10-27
More Coverity cleanup

dirkhh:
I'm planning to push more into this branch, but there's one commit that I'd like someone else (@bstoeger ?) to review. It's the third (Cleanup: better handling of NULL dive in setCurrentDiveSite).

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1478,6 +1478,18 @@ static int comp_dive_to_trip(struct dive *a, struct dive_trip *b)
 
 static int comp_dive_or_trip(struct dive_or_trip a, struct dive_or_trip b)
 {
+	/* we should only be called with both a and b having exactly one of
+	 * dive or trip not NULL. But in an abundance of caution, make sure
+	 * we still give a consistent answer even when called with invalid
+	 * arguments, as otherwise we might be hunting down crashes at a later
+	 * time...
+	 */
+	if (!a.dive && !a.trip && !b.dive && !b.trip)
+		return 0;
+	if (!a.dive && !a.trip)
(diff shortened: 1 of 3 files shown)


bstoeger (approved):
Fine with me.

core/divelist.c
```diff
@@ -1478,6 +1478,10 @@ static int comp_dive_to_trip(struct dive *a, struct dive_trip *b)
 
 static int comp_dive_or_trip(struct dive_or_trip a, struct dive_or_trip b)
 {
+	if (!a.dive && !a.trip)
```
bstoeger, line 1489: I'm not very happy about that. In principle, `!a.dive && !a.trip` is something that shouldn't happen.

If you want to handle this gracefully anyway, then it would be good to add
```
	if (!a.dive && !a.trip && !b.dive && !b.trip)
		return 0;
```
at the top. As it is, this doesn't define a (weak) ordering. If you can have as situation where 'a < b'  and 'b < a', `std::sort` *will* crash. We've been there before. :-P

dirkhh: Funny, I actually wondered about that, and then thought it was overkill. But you are right about the potential to cause havoc if a<b and b<a are both true.
Thanks for catching this.
Oh, and I have learned the really hard way that "this never happens" is generally not a good strategy. Years later the code may be used in subtly different way. Quite a few of our hard to find bugs resulted from things like that...

PR #2356, opened 2019-11-01
Profile: remove ProfileWidget2::recalcCeiling()

bstoeger:
The ProfileWidget2::recalcCeiling() function is used in one place,
namely when an undo-command changes the mode. It recalculates
decompression data and repaints the ceilings and thus avoids a
full profile-redraw.

This is smart, but it becomes problematic when the dive is changed
and the ceiling is recalculated before the profile is redrawn.
The DivePlotDataModel then still has data from the previous dive
but cylinders of the new dive are accessed.

This kind of situation may arise if multiple dive fields are
updated, as for example when replanning a dive.

Currently, this only causes a temporary mis-calculation. When
removing MAX_CYLINDERS this will lead to crashes.

One might attempt to fix the whole data-dependency mess. This
commit goes the cheap route and simply redraws the profile when
the mode is changed. Yes, it is in a way ineffective, but we
do worse things. The ProfileWidget2::recalcCeiling() thus becomes
unused and is removed.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I'm not sure what to classify this as - it is not a cleanup, nor does it change behavior, nor fix a current bug. It fixes a future bug: #2208 may crash on undo of dive-replanning without this. Details are in the commit message.

### Changes made:

Do a full profile reload when dive mode changes - not only a recalculation of the ceilings. This prevents out-of-sync data.

### Related issues:

Needed for #2208

### Release note:

Nope.
### Documentation change:

Nope.
### Mentions:

@atdotde: Please comment.

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -355,7 +355,7 @@ void MainTab::updateNotes(const struct dive *d)
 void MainTab::updateMode(struct dive *d)
 {
 	ui.DiveType->setCurrentIndex(get_dive_dc(d, dc_number)->divemode);
-	MainWindow::instance()->graphics->recalcCeiling();
+	MainWindow::instance()->graphics->replot();
 }
 
 static QDateTime timestampToDateTime(timestamp_t when)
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
Yeah, big hammer, but I can see your point.

dirkhh:
@atdotde - any comments from you?

dirkhh:
no comments from @atdotde so I merged this - easy enough to clean up if there are concerns, but I dislike having things bitrott here

PR #2357, opened 2019-11-02
Reorganise the Information tab of the Info panel

willemferguson:
1)  Reorganise the existing widgets in the Information tab
2)  Move divemode widget and visibility widget from Notes tab to
    Information tab
3)  Add 4 additional star widgets to cater for additional environmental
    data (wavesize, current, surge, chill). They are not connected to
    the underlying code yet.
4)  Translate water density to a word indicating water type
5)  Reorganise the Notes tab to compensate for the moving the
    divemode and visibility widgets to the Information tab
6)  Add corrections suggested by Dirk and Berthold
    Colour hex codes to colour names
    Calculate font size for star widget subscripts
    Change sone strings to tr("string")
    Remove instructions that have been commented out
    C++ stylistic changes
7)  Remove the problems in showing a QGroupBox in Qt Windows. I do
    this by removing the CSS specifying border characteristics

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Documentation change:

 This implies changes to the user manual.

### Mentions:

@dirkh @bstoeger @sfuchs79

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveInformation.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveInformation.cpp
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "TabDiveInformation.h"
 #include "ui_TabDiveInformation.h"
+#include "desktop-widgets/mainwindow.h" // TODO: Only used temporarilly for edit mode changes
+#include "profile-widget/profilewidget2.h"
 #include "../tagwidget.h"
 #include "core/units.h"
 #include "core/dive.h"
@@ -12,6 +14,7 @@
 
 #define COMBO_CHANGED 0
 #define TEXT_EDITED 1
+#define CSS_SET_HEADING_BLUE "QLabel { color: mediumblue;} "
 
 TabDiveInformation::TabDiveInformation(QWidget *parent) : TabBase(parent), ui(new Ui::TabDiveInformation())
 {
@@ -20,6 +23,28 @@ TabDiveInformation::TabDiveInformation(QWidget *parent) : TabBase(parent), ui(ne
 	QStringList atmPressTypes { "mbar", get_depth_unit() ,"use dc"};
 	ui->atmPressType->insertItems(0, atmPressTypes);
 	pressTypeIndex = 0;
+	// This needs to be the same order as enum dive_comp_type in dive.h!
+	QStringList types;
+	for (int i = 0; i < NUM_DIVEMODE; i++)
+		types.append(gettextFromC::tr(divemode_text_ui[i]));
+	ui->diveType->insertItems(0, types);
+	connect(ui->diveType, SIGNAL(currentIndexChanged(int)), this, SLOT(diveModeChanged(int)));
+	QString CSSSetSmallLabel = "QLabel { color: mediumblue; font-size: " +                        /* // Using label height ... */
+		QString::number((int)(0.5 + ui->diveHeadingLabel->geometry().height() * 0.66)) + "px;}"; // .. set CSS font size of star widget subscripts
+	ui->scrollAreaWidgetContents_3->setStyleSheet("QGroupBox::title { color: mediumblue;} ");
+	ui->diveModeBox->setStyleSheet("QGroupBox{ padding: 0;} ");
+	ui->diveHeadingLabel->setStyleSheet(CSS_SET_HEADING_BLUE);
+	ui->gasHeadingLabel->setStyleSheet(CSS_SET_HEADING_BLUE);
+	ui->environmentHeadingLabel->setStyleSheet(CSS_SET_HEADING_BLUE);
+	ui->groupBox_visibility->setStyleSheet(CSSSetSmallLabel);
+	QAction *action = new QAction(tr("OK"), this);
+	connect(action, &QAction::triggered, this, &TabDiveInformation::closeWarning);
+	ui->multiDiveWarningMessage->addAction(action);
+	action = new QAction(tr("Undo"), this);
+	connect(action, &QAction::triggered, Command::undoAction(this), &QAction::trigger);
+	connect(action, &QAction::triggered, this, &TabDiveInformation::closeWarning);
+	ui->multiDiveWarningMessage->addAction(action);
+	ui->multiDiveWarningMessage->hide();
 }
 
 TabDiveInformation::~TabDiveInformation()
@@ -34,7 +59,6 @@ void TabDiveInformation::clear()
 	ui->maxcnsText->clear();
 	ui->oxygenHeliumText->clear();
 	ui->gasUsedText->clear();
-	ui->dateText->clear();
 	ui->diveTimeText->clear();
 	ui->surfaceIntervalText->clear();
 	ui->maximumDepthText->clear();
@@ -43,6 +67,22 @@ void TabDiveInformation::clear()
 	ui->airTemperatureText->clear();
 	ui->atmPressVal->clear();
 	ui->salinityText->clear();
+	ui->waterTypeText->clear();
+}
+
+void TabDiveInformation::divesEdited(int i)
+{
+	// No warning if only one dive was edited
+	if (i <= 1)
+		return;
+	ui->multiDiveWarningMessage->setCloseButtonVisible(false);
+	ui->multiDiveWarningMessage->setText(tr("Warning: edited %1 dives").arg(i));
+	ui->multiDiveWarningMessage->show();
+}
+
+void TabDiveInformation::closeWarning()
+{
+	ui->multiDiveWarningMessage->hide();
 }
 
 // Update fields that depend on the dive profile
@@ -87,7 +127,6 @@ void TabDiveInformation::updateProfile()
 	if (current_dive->surface_pressure.mbar == 0) {
 		ui->atmPressVal->clear();			// If no atm pressure for dive then clear text box
 	} else {
-
 		ui->atmPressVal->setEnabled(true);
 		QString pressStr;
 		pressStr.sprintf("%d",current_dive->surface_pressure.mbar);
@@ -98,7 +137,6 @@ void TabDiveInformation::updateProfile()
 // Update fields that depend on start of dive
 void TabDiveInformation::updateWhen()
 {
-	ui->dateText->setText(get_short_dive_date_string(current_dive->when));
 	timestamp_t surface_interval = get_surface_interval(current_dive->when);
 	if (surface_interval >= 0)
 		ui->surfaceIntervalText->setText(get_dive_surfint_string(surface_interval, tr("d"), tr("h"), tr("min")));
@@ -108,10 +146,19 @@ void TabDiveInformation::updateWhen()
 
 void TabDiveInformation::updateSalinity()
 {
-	if (current_dive->salinity)
+	if (current_dive->salinity) {                 // Set up the salinity string:
 		ui->salinityText->setText(QString("%1g/ℓ").arg(current_dive->salinity / 10.0));
-	else
+		if (current_dive->salinity < 10050)   // Set water type indicator:
+			ui->waterTypeText->setText(tr("Fresh"));
+		else if (current_dive->salinity < 10190)
+			ui->waterTypeText->setText(tr("Salty"));
+		else if (current_dive->salinity < 10210) // (EN13319 = 1.019 - 1.021 g/l)
+			ui->waterTypeText->setText(tr("EN13319"));
+		else ui->waterTypeText->setText(tr("Salt"));
+	} else {
 		ui->salinityText->clear();
+		ui->waterTypeText->clear();
+	}
 }
 
 void TabDiveInformation::updateData()
@@ -125,12 +172,11 @@ void TabDiveInformation::updateData()
 	updateWhen();
 	ui->waterTemperatureText->setText(get_temperature_string(current_dive->watertemp, true));
 	ui->airTemperatureText->setText(get_temperature_string(current_dive->airtemp, true));
-	updateSalinity();
-
-	ui->atmPressType->setEditable(true);
 	ui->atmPressType->setItemText(1, get_depth_unit());  // Check for changes in depth unit (imperial/metric)
-	ui->atmPressType->setEditable(false);
-	ui->atmPressType->setCurrentIndex(0);  // Set the atmospheric pressure combo box to mbar
+	ui->atmPressType->setCurrentIndex(0);                // Set the atmospheric pressure combo box to mbar
+	updateMode(current_dive);
+	updateSalinity();
+	ui->visibility->setCurrentStars(current_dive->visibility);
 }
 
 // This function gets called if a field gets updated by an undo command.
@@ -141,6 +187,10 @@ void TabDiveInformation::divesChanged(const QVector<dive *> &dives, DiveField fi
 	if (!current_dive || !dives.contains(current_dive))
 		return;
 
+	if (field.visibility)
+		ui->visibility->setCurrentStars(current_dive->visibility);
+	if (field.mode)
+		updateMode(current_dive);
 	if (field.duration || field.depth || field.mode)
 		updateProfile();
 	if (field.air_temp)
@@ -149,10 +199,27 @@ void TabDiveInformation::divesChanged(const QVector<dive *> &dives, DiveField fi
 		ui->waterTemperatureText->setText(get_temperature_string(current_dive->watertemp, true));
 	if (field.atm_press)
 		ui->atmPressVal->setText(ui->atmPressVal->text().sprintf("%d",current_dive->surface_pressure.mbar));
-	if (field.datetime)
-		updateWhen();
 	if (field.salinity)
 		updateSalinity();
+}
+
+
+void TabDiveInformation::on_visibility_valueChanged(int value)
+{
+	if (current_dive)
+		divesEdited(Command::editVisibility(value, false));
+}
+
+void TabDiveInformation::updateMode(struct dive *d)
+{
+	ui->diveType->setCurrentIndex(get_dive_dc(d, dc_number)->divemode);
+	MainWindow::instance()->graphics->recalcCeiling();
+}
+
+void TabDiveInformation::diveModeChanged(int index)
+{
+	if (current_dive)
+		divesEdited(Command::editMode(dc_number, (enum divemode_t)index, false));
 }
 
 void TabDiveInformation::on_atmPressType_currentIndexChanged(int index) { updateTextBox(COMBO_CHANGED); }
@@ -193,7 +260,7 @@ void TabDiveInformation::updateTextBox(int event) // Either the text box has bee
 			break;
 		}
 		if (atmpress.mbar)
-			Command::editAtmPress(atmpress.mbar, false);      // and save the pressure for undo
+			divesEdited(Command::editAtmPress(atmpress.mbar, false));      // and save the pressure for undo
 	}
 }
 
--- a/desktop-widgets/tab-widgets/maintab.ui
+++ b/desktop-widgets/tab-widgets/maintab.ui
@@ -298,114 +298,19 @@
             </layout>
            </item>
            <item>
-            <layout class="QGridLayout" name="gridLayout_8" columnstretch="0,0,1">
-             <property name="leftMargin">
+            <layout class="QHBoxLayout" name="gridLayout_9">
+             <property name="Spacing">
               <number>5</number>
              </property>
-             <property name="rightMargin">
-              <number>5</number>
-             </property>
-             <property name="horizontalSpacing">
-              <number>5</number>
-             </property>
-             <property name="verticalSpacing">
-              <number>0</number>
-             </property>
-             <item row="0" column="0">
-              <widget class="QLabel" name="RatingLabel">
-               <property name="sizePolicy">
-                <sizepolicy hsizetype="MinimumExpanding" vsizetype="Preferred">
-                 <horstretch>0</horstretch>
-                 <verstretch>0</verstretch>
-                </sizepolicy>
-               </property>
-               <property name="text">
-                <string>Rating</string>
-               </property>
-              </widget>
-             </item>
-             <item row="0" column="1">
-              <widget class="QLabel" name="visibilityLabel">
-               <property name="sizePolicy">
-                <sizepolicy hsizetype="MinimumExpanding" vsizetype="Preferred">
-                 <horstretch>0</horstretch>
-                 <verstretch>0</verstretch>
-                </sizepolicy>
-               </property>
-               <property name="text">
-                <string>Visibility</string>
-               </property>
-              </widget>
-             </item>
-             <item row="1" column="0" alignment="Qt::AlignVCenter">
-              <widget class="StarWidget" name="rating" native="true">
-               <property name="sizePolicy">
-                <sizepolicy hsizetype="MinimumExpanding" vsizetype="Preferred">
-                 <horstretch>0</horstretch>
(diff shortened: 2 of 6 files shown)


bstoeger:
The code is fine with me. If needed, we can do cleanups later.

I'll leave the UI discussion to others as I have no strong opinion on this. I still wonder if we should define different views, e.g. one for recreational divers, one for professionals, one for free divers, etc.

glance-:
Why do we want editable fields in the information tab? Isn't that supposed to be calculated/display only data?

I'm really not sure we would like more star widgets. Whats the meaning of a 3 star wavesize? What is even the meaning of surge and chill? I'm against more of these widgets.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -1,17 +1,21 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "TabDiveInformation.h"
 #include "ui_TabDiveInformation.h"
+#include "desktop-widgets/mainwindow.h" // TODO: Only used temporarilly for edit mode changes
```
glance-, line 4: What do this TODO mean? temporarily until what?

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -125,12 +169,27 @@ void TabDiveInformation::updateData()
 	updateWhen();
 	ui->waterTemperatureText->setText(get_temperature_string(current_dive->watertemp, true));
 	ui->airTemperatureText->setText(get_temperature_string(current_dive->airtemp, true));
-	updateSalinity();
-
-	ui->atmPressType->setEditable(true);
```
glance-, line 130: Why is this setEditable removed?

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -125,12 +169,27 @@ void TabDiveInformation::updateData()
...
+	if (current_dive->salinity)
+		ui->salinityText->setText(QString("%1g/ℓ").arg(current_dive->salinity / 10.0));
+	else
+		ui->salinityText->clear();
```
glance-: Why is updateSalinity() removed, and a version inlined here?

glance- (changes requested):

desktop-widgets/tab-widgets/TabDiveInformation.cpp
bstoeger: This will be removed once everything is undo-ized and we can remove the edit-mode. I will take care of that.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: Because it is set to FALSE at the start and because we do not want the dropdown list to be user-editable. Makes sense?

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: Fixed, thank you.

dirkhh:
> Why do we want editable fields in the information tab? Isn't that supposed to be calculated/display only data?

This was discussed a while ago. The tradeoff was "way too many fields on the Notes tab" or "add editable fields on the info tab". I ended up siding with Willem on adding editable fields on the info tab and at the same time removing some of the redundant data between the two tabs and making things more logical.

> I'm really not sure we would like more star widgets. Whats the meaning of a 3 star wavesize? What is even the meaning of surge and chill? I'm against more of these widgets.

I think I voiced the same opinion, but there wasn't really much discussion. My biggest problem is that these are all things that I would never document, but I hear that more scientifically slanted divers care about things like this. In my mind these would better be handled by free text entries in the Notes field -- especially since once we add this here, the next person will say "I want to be able to filter for all dives with 3 star waves" -- and what the heck are 3 star waves in the first place...

But attempts to discuss this on the developer mailing list lead to complete silence. No one cares enough to even comment.

willemferguson:
If the star widgets are removed then we would need to remove ALL the star widgets, including that for visibility. Some divers would like to retain the visibility star widget, this came out in the previous discussion. There were several positive comments at the time, e.g. to indicate on these widgets where the bestcase and worstcase extremes of each widget are. Then my question would be what makes the visibility widget more important than current, swell surface waves or chill? From a dive safety point of view it is more useful to have information on the non-visibility environmental characteristics. I often refer back to these data, to see how sufficient my suit was at a specific site and/or for a specific month or season. Therefore my point of view is: if we remove the star widgets then all of them should be removed. At this point we are not really pressed for screen space due to the fact that the information is now presented in a much more compact way.
Does anyone have any suggestions?

dirkhh:
> If the star widgets are removed then we would need to remove ALL the star widgets, including that for visibility. 

I don't understand that logic. We can retain what we have and not add new ones.

> Some divers would like to retain the visibility star widget, this came out in the previous discussion. There were several positive comments at the time, e.g. to indicate on these widgets where the bestcase and worstcase extremes of each widget are. Then my question would be what makes the visibility widget more important than current, swell surface waves or chill? 

Simple - we've had those since forever (at least version 3). And most every dive log I've played with has some form of a rating widget.

Having said that - if I were to start Subsurface as a new project with what I know about MY use of it... I wouldn't add either of those two widgets. Way too much trouble for way too little value (to me)

> From a dive safety point of view it is more useful to have information on the non-visibility environmental characteristics. I often refer back to these data, to see how sufficient my suit was at a specific site and/or for a specific month or season. Therefore my point of view is: if we remove the star widgets then all of them should be removed. At this point we are not really pressed for screen space due to the fact that the information is now presented in a much more compact way.
> Does anyone have any suggestions?

Again, you are missing the point. This is not an all or nothing conversation. The status quo ante is that we have those two widgets. So the discussion needs to be "do we need a change from that and why". I.e. you would need to present strong arguments to either remove the two existing ones or to add new ones. You do not need to present arguments to keep what we have...

willemferguson:
My argument is simply that visibility is not by any means the most important environmental characteristic to log. Information about surface, surge, current and perceived chill provide an important context for understanding the dive, answering questions such as "Why did I not add a 3m deco stop to the dive.. large surface waves?"; "How efficient was my suit to keep me warm?"; "Why was my SAC unexpectedly high.. the current?". Frequently we cannot find a definitive answer to such questions but having a widget allows one to eliminate some possibilities. This information is more useful than visibility info.

bstoeger:
From my side the additional fields wouldn't hurt - I'm not even using the old ones, but they appear to be rather popular. 🤷 I must say I do see the point in current, though.

The problem is that this sets a precedent and where should we stop? It's probably a silly example, but a recreational warm-water photographer/videographer may for example be more interested in rating topography, fish diversity, or fish density than swell or chill. It's hard to please all.

I'd suggest splitting this PR in two: reorganization of the existing fields and addition of new fields. The former seems to be finished, whereas the latter still needs backend support (parsers, etc).

desktop-widgets/tab-widgets/TabDiveInformation.cpp
glance-: But what did this code do to begin with? This is one of the problems with huge commits containing multiple changes combined into one, that there are no information on _WHY_ a certain chang was done.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
glance-: No, its not. The copied version is still there.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -7,7 +7,6 @@
 #include "core/units.h"
 #include "core/dive.h"
 #include "desktop-widgets/command.h"
-#include "desktop-widgets/command_edit.h"
```
glance-: Another unrelated change?

glance- (changes requested):
Don't write that it "takes into account a concerns" , rather write what those concerns where and how this takes care of them.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -188,6 +187,7 @@ void TabDiveInformation::updateData()
 		ui->waterTypeText->clear();
 	}
 	updateMode(current_dive);
+	updateSalinity();
```
glance-, line 178: Now this code is run twice. Once in the copied version, and once in the function called.

glance- (changes requested):

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: Yup, that header is not required to build this file

glance-: Then that's unrelated to "Do updatesalinity upon screen refresh", and should have bin a separate commit stating why its removed, or rather, update the commit where you add it so it doesn't add it at all.

willemferguson: Fixed. I put the line back. I will submit that change as a separate PR.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: My misunderstanding. Fixed now.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
glance-: No, Not a separate PR, correct the commit where you added it instead.

willemferguson: whatever  you want  ......

willemferguson: I am not up to that level in git. I have now deleted that line. This is a change in code, therefore, as far as I can see, requiring an additional change to the present PR. If you would like me to do anything else, please send  me the git instructions to do elsewise.

glance-: https://git-scm.com/book/en/v2/Git-Branching-Rebasing
https://help.github.com/en/github/using-git/about-git-rebase

bstoeger: @willemferguson: If you want to merge two commits, first find out the commit id of the commit you want to merge into with
```
git log
```
Then
```
git rebase -i COMMIT-ID^
```
(the ^ is important, because you will be presented the commits *after* the commit you specified).
This will open an editor.
Copy the line of the commit you want to merge below the line of the commit you want to merge into. Then change the first letter of the commit you want to merge to `s` for "squash into previous commit".

Then save and exit the editor and it should be done.

I suggest doing this in a test-branch, which you can create by
```
git checkout -b branch-name
```

If this doesn't work out for you, I'm sure someone of us can do this for you before merging.

willemferguson: I have now submitted a unified PR which hopefully comprises only one commit and which I hope adresses all the issues raised.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: It sets up the dropdown list used for selecting the appropriate unit for specifying atmospheric pressure. The function initialises the values in the Information tab.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: My oversight. duplication removed now.

atdotde:
Hi,

> On 3. Nov 2019, at 15:49, Dirk Hohndel <notifications@github.com> wrote:
> 
> Again, you are missing the point. This is not an all or nothing conversation. The status quo ante is that we have those two widgets. So the discussion needs to be "do we need a change from that and why". I.e. you would need to present strong arguments to either remove the two existing ones or to add new ones. You do not need to present arguments to keep what we have...

I have not really followed this recent discussion, but here are my two cents: I almost exclusively dive at local fresh water spots. That means: Same drysuit all year round, never any signifiant waves or current or similar but visibility is changing a lot during the year (mainly based on season, recent weather and some longer term environmental trends like the local farmer spreading too much fertilisers over his fields). So I like the visibility widget as it helps tracking longer term trends and I like it as stars as I am not good at making quantitative guesses. The “rating” widget is really how much fun I had (depending on a million factors). 

In short, the existing widgets make sense to me. Additional factors that might have been important to that specific dive I would put in the notes or tags.

But maybe I am one fashioned and like the status quo.

Best
Robert

willemferguson:
I dive caves, freshwater and sea water using four types of suits and three types of underware depending on the situation from 1m depth to 100m depth using recreational, sidemount, twinset and rebreather gear. I have quite few divers around me (3 of them using subsurface) that also dive varied dives. Being able to refer back to previous dives becomes much more useful in these circumstances. Currently we actually put most of the data in the notes field. The fact is that subsurface would be more useful if it had these items as star widgets instead of having to remember to add these items after every dive. Would it make subsurface any less useful if these star widgets were indeed kept? Every person has a right to personal preference but which person's personal preference should be followed? I guess the yard stick should be to maximise the functionality of the software and make it as adaptable as possible.

willemferguson:
I apologise if my comment above sounds like a rant. The reason is that I get so depressed by the quality of argumentation (even my own). The argument should be with respect to the advantages as counterweighed by the disadvantages of some new feature and whether someone actually cares enough to write the code for this new feature. Personal points of view that, not substantiated by fact that allow progress in an argument to proceed are simply not helpful. There is a fact that seems to be overlooked. I gave plenty of advanced notice of what I intended (several demo screenshots) and there were at the time several positive comments on the issue of star widgets. Two of these comments I actually implemented. My point of view is that, if these fundamental problems do exist, they should have been raised at the time. Not now after I committed significant time to implement what was raised during the proposals. I get depressed when the argument just falls back to a starting position as if this is the first time the discussion is being brought up.

dirkhh:
Willem, not getting feedback on the mailing list is one of my main frustrations with Subsurface. Your experience is unfortunately very consistent with my own. I am very grateful that some of the key developers at least comment here. Yes, that's frustrating, yes, that feels like it's late in the process, but I am still grateful because every little bit helps - otherwise it's just me and that is not sustainable.

We have run into similar scenarios before. Every user, and especially every developer sees THEIR use case as the critical one, and as the one that Subsurface should cater for. You may remember the endless flame wars about the planner which resulted in me actually disabling the planner (that I had originally written myself) for a while. The discussions about OC vs CCR. The discussions about asset management (the people who want to track the serial number of their regulator). This isn't new. This isn't easy to solve.

We have more than 10k users these days. How many of them do you think dive in ways that are similar to you? How many of them dive mainly in very similar environments. I used to dive both cold water / dry suit and warm water. I haven't been in cold water in over a year and I don't think I'll be back any time soon. Most divers tend to find the thing that they like and keep doing that. For a small group that could be "I dive caves, freshwater and sea water using four types of suits and three types of underware depending on the situation from 1m depth to 100m depth using recreational, sidemount, twinset and rebreather gear." - I just don't think that's the majority of our users.

So how much screen real estate do we want to devote to that group. Vs the people tracking their regulators. Vs marine biologists who want drop downs for all the species that they have found on their dives. And the people who want a global, shared dive site database.

So... I can see current as something that's very similar to visibility. If I think back about some of the dives where the current pissed me off, and all the dives where I was glad that there was none - yeah, that's generally useful outside of pool dives.

I have a hard time seeing how I would even rate waves/swell - that seems something much better handled by tags / free text.

And I think "chill" is just something that doesn't quite fit in with the other things we track. If this is important to you, use tags chill1 through chill5 or something.

Yes, I didn't send that feedback on the mailing list. I'm pretty sure I pushed back about these and your rating of them at some point, but I should have been more definitive. Part of this is that I don't want to always be the lone dictator. Because I often get stuff wrong. But it seems at least among the few people commenting here there is some consensus on that.

willemferguson:
Thank you for that comment. You yourself suggested that each star widget should be consistently arranged with the least comfortable value on the left and the most comfortable value at the right. That did not appear to be a pushback.

But these things are really beside the point at the moment. Let's rather look at the future. What should happen to this PR now? Should i just abandon it?

dirkhh:
Not at all.
Almost everything you are doing here had broad support.
The question seems to be mainly about whether a couple of the new star widgets are going in the right direction. So let's get the overall PR into shape and then consider what we'll do with those widgets.
Ideally we could have the additional widgets in their own commits. I can help with that.

willemferguson:
This has been done. The current PR only includes the reorganisation of existing data fields. As indicated a few km up the screen after all these discussions, the star widgets constitute a separate PR. As far as I can see the PR is in shape.

dirkhh:
My apologies, I will find time today to go through this again... I should have checked the commits.

dirkhh:
There's something odd going on with the GitHub actions.
Internet speed at the conference I'm at today is rather, err, disappointing, so I'm having a hard time debugging what's going on there :-(

dirkhh (commented):
The commit message reads like the log of what happened to the PR.
When I come back in the future, it makes a lot more sense for the commit message to describe what's in the commit. So I'd drop item 3) and 8) and reconsider 6) and 9) (I think all of us are ok to not be mentioned in the commit)

willemferguson:
Commit message is now updated to be correct.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
glance-: Move the updateSalinity back where it were before, then, if there is no reason for moving it.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
glance-: You created the setEditable(true) , setItemText, setEditable(false) dance back in 1bdf00b2b472078a0b24f1c269782d822cb96e02 , and now you remove the setEditable(true) bit of it. Either leave it there, or remove both the setEditable(true) and setEditable(false) bits and write about why in the commit message.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
glance-: Also, the whole waterTypeText block should go into the updateSalinity method, so all calls to updateSalinity also refreshes the waterTypeText. Otherwise they will diverge on divesChanged.

glance-:
```git am aeb541f2e42407f87d7f0156496c7687be319e2f.patch
Applying: Reorganise the Information tab of the Info panel
.git/rebase-apply/patch:110: trailing whitespace.
		if (current_dive->salinity < 10050) 
.git/rebase-apply/patch:1064: trailing whitespace.
           <item>  
.git/rebase-apply/patch:1099: trailing whitespace.
           <item>  
warning: 3 lines add whitespace errors.```

File: desktop-widgets/tab-widgets/TabDiveInformation.cpp (hunk 2 above)
Comment (willemferguson) [reply desktop-widgets/tab-widgets/TabDiveInformation.cpp:130]: I removed the setEditable statements. Initially I had some problems with setItemText and I experimented with setEditable to get the setItemText to work. Clearly setEditable is not appropriate.

File: desktop-widgets/tab-widgets/TabDiveInformation.cpp (hunk 5 above)
Comment (willemferguson) [reply desktop-widgets/tab-widgets/TabDiveInformation.cpp:178]: This is a good suggestion that I implemented. Thank you, @glance

**Comment by dirkhh** [comment]
@glance-  I think I slightly lost track here - are all of your concerns addressed?

File: desktop-widgets/tab-widgets/TabDiveInformation.cpp (hunk 5 above)
Comment (glance-) [reply desktop-widgets/tab-widgets/TabDiveInformation.cpp:178]: Not seeing that code, did you forget to push?

File: desktop-widgets/tab-widgets/TabDiveInformation.cpp (hunk 2 above)
Comment (glance-) [reply desktop-widgets/tab-widgets/TabDiveInformation.cpp:130]: But why is the setEditable(false) still there then, and why isn't this mentioned in the commit message?

File: desktop-widgets/tab-widgets/TabDiveInformation.cpp [hunk 6]
```diff
@@ -125,12 +164,23 @@ void TabDiveInformation::updateData()
...
 	ui->waterTemperatureText->setText(get_temperature_string(current_dive->watertemp, true));
 	ui->airTemperatureText->setText(get_temperature_string(current_dive->airtemp, true));
-	updateSalinity();
-
-	ui->atmPressType->setEditable(true);
 	ui->atmPressType->setItemText(1, get_depth_unit());  // Check for changes in depth unit (imperial/metric)
 	ui->atmPressType->setEditable(false);
 	ui->atmPressType->setCurrentIndex(0);  // Set the atmospheric pressure combo box to mbar
+	if (current_dive->salinity) {			// Set water type indicator (EN13319 = 1.020 g/l)
+		if (current_dive->salinity < 10050)
```
glance-: Trailing whitespace

desktop-widgets/tab-widgets/maintab.ui
```diff
@@ -298,100 +298,15 @@
...
-                </size>
-               </property>
-              </spacer>
-             </item>
-            </layout>
-           </item>
-           <item>
-            <layout class="QGridLayout" name="gridLayout_9">
-             <property name="horizontalSpacing">
+           <item>
```
glance-: Trailing whitespaces. Was this hand edited?

desktop-widgets/tab-widgets/maintab.ui
```diff
@@ -439,6 +344,41 @@
              </item>
             </layout>
            </item>
+           <item>
```
glance-: More trailing whitespaces.

glance-:
Nope.

I'd also loved to see each of the items on the list in the commit message as a separate commit, but that is probably hoping for to much.

desktop-widgets/tab-widgets/maintab.ui
willemferguson: done

desktop-widgets/tab-widgets/maintab.ui
willemferguson: done

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: The previous push did not go as expected. I cannot see white space. Can you still see it? If so then please indicate explicitly

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: The previous push did not go as expected. Please look now? solved?

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: Can you still see it? I cannot.

glance-: Now its gone.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
glance-: Great.

For future reference, in every case where you end up writing a list in the commit message, you should split it in different commits instead. One commit per logical change.

glance- (approved):

dirkhh:
Thank you @glance- for the in depth review - thanks to both of you for your patience and follow-through.
This really makes my life easier!
I'll merge this manually since GitHub sees a conflict.

PR #2359, opened 2019-11-03
Ios updates

dirkhh:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

A few small changes needed in order to get an iOS build going on the latest Xcode

And a few overall mobile cleanups and additions

### Mentions:

@atdotde - I think you are one of the few people who occasionally build their own iOS binaries?

PR #2360, opened 2019-11-06
[RFC] Mobile dive list refactor

dirkhh:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a first working draft of what I want to do with the dive list on mobile.
Having two nested filter proxy models seems fragile, but it seems to work well with excellent performance, both mobile-on-desktop and more importantly when testing on a fairly old iPhone 6.
I need to go through the code a couple more time to see if we still need all the hacks and if the models can be simplified.
Right now I just flat out removed the animation of expanding/collapsing trips - that should be easy to add back.
Having said all this, in my "modest" testing (three different types of dive lists that hopefully cover all the various combinations of trips, dives, notrip dives, mobile-on-desktop and on device) this seems to work as designed with a significant performance improvement.

### Changes made:

1) add second `QSortFilterProxyModel` for the vertical dive list that hides all but the first dive in any non-expanded trip
2) use that model and simplify the visibility decision by implementing that in C++ code
3) ensure that the filters get updated when things change
4) change the way we pick which dive is marked as selected in the vertical dive list (going back to the core data structures instead of assuming that indices between the models align)
5) connect that new model to the horizontal dive list (the dive details you can flip through) by translating dive id into the corresponding index

### Related issues:

many, many performance complaints

### Release note:

need to add to CHANGELOG

### Documentation change:

This should be user transparent (except for the better performance)

### Mentions:

@bstoeger I'd love your comments on the model implementation. I hope I didn't do anything too terrible
@atdotde please test on your iDevice (and also look at the code)
@janmulder if you have the time... I'm interested in your comments about my QML code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: fix subtle potential new data loss bug in first session connecting to a cloud account
 Planner: Add checkbox on considering oxygen narcotic
 Planner: Improve rounding of stop durations in planner notes
 Desktop: register changes when clicking "done" on dive-site edit screen
--- a/core/dive.h
+++ b/core/dive.h
@@ -144,6 +144,9 @@ struct dive {
 	struct dive_trip *divetrip;
 	bool selected;
 	bool hidden_by_filter;
+#if defined(SUBSURFACE_MOBILE)
+	uint8_t collapsed; /* four values: 0 = don't show, 1 = show as dive, 2 = show corresponding trip, 3 = show dive and trip */
+#endif
 	timestamp_t when;
 	struct dive_site *dive_site;
 	char *notes;
--- a/core/dive.h
+++ b/core/dive.h
@@ -139,23 +139,17 @@ struct dive_site_table;
 struct dive_trip;
 struct trip_table;
 struct dive {
-	int number;
-	bool notrip; /* Don't autogroup this dive to a trip */
 	struct dive_trip *divetrip;
-	bool selected;
-	bool hidden_by_filter;
-#if defined(SUBSURFACE_MOBILE)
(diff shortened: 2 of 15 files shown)


dirkhh:
eh, I missed that this includes #2359 - ignore that part, the new commits start with "add collapsed state to the dive data structure"

dirkhh:
@bstoeger I know you said you are busy today - given that I am shipping this code to beta testers on both iOS and Android I think I'd prefer to merge it, even if your tree model based code may quickly supersede it. Thoughts?

bstoeger:
Yes, I'm not very happy about having display-related data in the dive structure, but for now I'm fine with this.

bstoeger (approved):

dirkhh:
> Yes, I'm not very happy about having display-related data in the dive structure, but for now I'm fine with this.

We've had `selected` in there forever and `hidden_by_filter` for almost exactly 5 years to the day: 19ca90f1a0 

That doesn't mean that this is the right thing to do, just that there is precedent :-)

(and we'll ignore the inconvenient fact that it is me setting that precedent, with `hidden_by_filter`... at least `selected` was introduced by Linus in 2012... 972669d636 )

bstoeger:
Yes, I know. Nevertheless, I'd prefer a vector of selected dives, because that would be much closer to Qt's representation of the selection. Not a high priority though.

dirkhh:
I am very, very open to investigating different implementations for both the filter state and the selection state. I am in no way married to the current way of doing things. I find that the current way of doing things has that charming property of "it seems to work"... :-)

PR #2361, opened 2019-11-07
scripts: update get-deps to allow clean build

janiversen:
libssh2 depends on openssl, therefore it is important that openssl is
build before libssh2.

The old get-deps would cause errors in 2 situations:
1) In a clean build, make of libssh2 would fail
2) In a normal build, where openssl changed version, make of libssh2 would
depend on old build.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -237,6 +237,17 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	make install
 	popd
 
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
I have actually stopped building my own dependencies this way, but I believe this change is correct

PR #2363, opened 2019-11-08
Undo of weightsystem editing

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

We have to progress on that front - this implements undo of the weight system actions (add, delete, edit). That's the easy part of the equipment tab, but it gave me some ideas on how to handle the much harder part: cylinders. The problem with cylinders is that they are used in two places: normal editing and planner. That will be fun.

Overall, this was a bit harder than expected, because when hovering over the weightsystem-types, the old code would overwrite the displayed dive. The logic was changed so that the model now has a "temporary" weightsystem line that it can display. This can be set, committed or reset.

I don't like how the model calls the undo-command, but that's the easiest way for now.

This is only lightly tested.

### Release note:

Done.
### Documentation change:

Unsure.

### DIFF
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -278,6 +278,11 @@ int removeWeight(int index, bool currentDiveOnly)
 	return execute_edit(new RemoveWeight(index, currentDiveOnly));
 }
 
+int editWeight(int index, weightsystem_t ws, bool currentDiveOnly)
+{
+	return execute_edit(new EditWeight(index, ws, currentDiveOnly));
+}
+
 // Trip editing related commands
 void editTripLocation(dive_trip *trip, const QString &s)
 {
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -273,6 +273,11 @@ int addWeight(bool currentDiveOnly)
 	return execute_edit(new AddWeight(currentDiveOnly));
 }
 
+int removeWeight(int index, bool currentDiveOnly)
+{
+	return execute_edit(new RemoveWeight(index, currentDiveOnly));
+}
+
 // Trip editing related commands
 void editTripLocation(dive_trip *trip, const QString &s)
 {
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -268,6 +268,11 @@ void editProfile(dive *d)
 	execute(new ReplanDive(d, true));
 }
 
+int addWeight(bool currentDiveOnly)
+{
+	return execute_edit(new AddWeight(currentDiveOnly));
+}
+
 // Trip editing related commands
 void editTripLocation(dive_trip *trip, const QString &s)
 {
--- a/commands/command.h
+++ b/commands/command.h
@@ -83,6 +83,7 @@ void replanDive(dive *d); // dive computer(s) and cylinder(s) will be reset!
 void editProfile(dive *d); // dive computer(s) and cylinder(s) will be reset!
 int addWeight(bool currentDiveOnly);
 int removeWeight(int index, bool currentDiveOnly);
+int editWeight(int index, weightsystem_t ws, bool currentDiveOnly);
 
 // 5) Trip editing commands
 
--- a/commands/command.h
+++ b/commands/command.h
@@ -82,6 +82,7 @@ void pasteDives(const dive *d, dive_components what);
 void replanDive(dive *d); // dive computer(s) and cylinder(s) will be reset!
 void editProfile(dive *d); // dive computer(s) and cylinder(s) will be reset!
 int addWeight(bool currentDiveOnly);
+int removeWeight(int index, bool currentDiveOnly);
 
 // 5) Trip editing commands
 
--- a/commands/command.h
+++ b/commands/command.h
@@ -81,6 +81,7 @@ int editDiveMaster(const QStringList &newList, bool currentDiveOnly);
 void pasteDives(const dive *d, dive_components what);
 void replanDive(dive *d); // dive computer(s) and cylinder(s) will be reset!
 void editProfile(dive *d); // dive computer(s) and cylinder(s) will be reset!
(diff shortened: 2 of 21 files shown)


dirkhh (commented):
Code review looks good. You have a couple of //: comments that point out that the text there would be better if part of the string would be removed for n=1... I think it would be cleaner to just have conditional code there that uses different strings for n=1 and n>1 -- relying on the translation mechanism to take care of that seems a bit excessive.

I have NOT tested this at all - hoping to do so later today. That's why this is just a comment and not an Approval

dirkhh:
Hmm - I'm having a weird issue.
Dive without weight.
I add weight called "integrated" it autocompletes to "3.6kg" (I apparently used that name before).
I try to update this to 4kg
It refuses to update, switches back to 3.6kg
I edit other dives
I try again. still can't change the amount for that weight

bstoeger:
Thanks for testing - I will look into that.

Off-topic: I started with the linear-model thing, but was side-tracked by other things (e.g. reading a thesis). Then lost interest when I realized that in QML we use roles, on desktop columns. Will try to do more today.

bstoeger:
> I think it would be cleaner to just have conditional code there that uses different strings for n=1 and n>1 -- relying on the translation mechanism to take care of that seems a bit excessive.

Fine with me - but note that last time we had an if, you said that we use the translation thing for that. In that case though it was only the plural form that changed.

But perhaps we should even leave in the "weight edited (1 dive)"? Up to you.

dirkhh:
> Off-topic: I started with the linear-model thing, but was side-tracked by other things (e.g. reading a thesis). Then lost interest when I realized that in QML we use roles, on desktop columns. Will try to do more today.

Ahh, missed that comment earlier - see my email to you :-)
Yes, roles vs. columns was one of the things I noticed as well, but that seemed like a rather straight forward translation (famous last words). I'm quite eager to see this - knowledge in my brain appears to work like OS paging algorithms... so right now the understanding of the QML dive list is "paged in" and I have worked on understanding the models and would love to put this "warm cache" situation to good use... before something else displaces it again and I need to start reading everything again...

dirkhh:
> > I think it would be cleaner to just have conditional code there that uses different strings for n=1 and n>1 -- relying on the translation mechanism to take care of that seems a bit excessive.
> 
> Fine with me - but note that last time we had an if, you said that we use the translation thing for that. In that case though it was only the plural form that changed.
> 
> But perhaps we should even leave in the "weight edited (1 dive)"? Up to you.

Ahh. Yeah - this is subtle. If all we do is "%n dive(s)" then using the plural algorithm in the translation is really easiest. The situation here seemed different in that you wanted no text when n is 1. Because we still need the plural thing for the n>1 case (as some languages have a different form for 2 than for more than 2). But for the n=1 case we just want to drop part of the text and that's harder to get right for the translators. It's doable - but I think it would be easier to separate that out in code.

I realize writing this that maybe I'm not fully consistent in my reasoning here :-)

bstoeger:
> Hmm - I'm having a weird issue.

I can reproduce this and it's an UI issue. The crucial point is that when you choose a different weight type, you don't click on it, but move away with the mouse. By explicitly clicking on a (different) weight-type it will work again.

The problem is, if the user doesn't click on an item, we don't get the correct closeEditor signal that says "commit this". This is annoying, because if we react to every closeEditor signal, we get two signals when the user clicks on an item. I will have to think about that.

bstoeger:
I hope I fixed it, but this whole delegate stuff is very obscure and I'm not sure how stable it is across Qt versions. We might have to adjust the Hint-switch statement.

bstoeger:
Argh. Stopped working again - will check that out tomorrow. We're getting a spurious setModelData call after having committed the change.

bstoeger:
Alright, back to this...

Rebased on current master (surprisingly painful) and changed the weightsystem-model logic. I couldn't get the editor-hint and setValue calls under control (too many different cases: clicked, tab-pressed, esc-pressed, etc.). Therefore, I chose a different approach: if the item is set to the already existing item, we exit (or don't enter) temporary-mode.

@dirkhh: please test if you find time.

dirkhh:
And now this one conflicts as well.
I need to be faster merging things so I don't end up causing all this bit rot to add extra cycles.
:-(

dirkhh:
I merged this locally and the resulting build behaves strangely.
Simple test file with a could of dives with no weights.
Click the plus to add a weight system.
No visual change.
Click the plus again.
Nothing.
Click the plus for a tank. Ahh, that works. Discard.
Now the two empty weigh systems show up...

Second attempt, same file
add weight
nothing
switch to different dive
try to switch back
app hangs.

bstoeger:
Interesting. Doesn't happen for me (with unmerged PR).

bstoeger:
Oh - found a way to reproduce this - you mustn't change dive first. Appears to be an initialization thing.

bstoeger:
Note that even if you *do* have weight and cylinder - before you switch dive they aren't shown.

bstoeger:
This particular problem should be fixed by #2400 - let's get that in first and then rebase.

bstoeger:
> I merged this locally and the resulting build behaves strangely.

Rebased on master and that particular problem should be fixed.

bstoeger:
@dirkhh: Added a commit to move the "leave-out parentheses from text for 1 dives" logic to C++.

I did this as a separate commit because we already used the pattern elsewhere and so we stay consistent.

I'm not 100% convinced that we should leave out the parentheses for single-dive edits and could also simply remove the translator comments - feel free to leave out the last commit.

dirkhh (approved):
Thank you. This seems to work now.

PR #2364, opened 2019-11-09
Mobile: add ability to directly open support email

dirkhh:
This is even easier and more obvious than copying the logs on the About page
and then manually creating an email and pasting those logs.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

tested on iOS, Android, and mobile-on-desktop Mac

### Release note:

done

### Documentation change:

done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: add menu entry to start a support email, using native email client
 Mobile: fix subtle potential new data loss bug in first session connecting to a cloud account
 Planner: Add checkbox on considering oxygen narcotic
 Planner: Improve rounding of stop durations in planner notes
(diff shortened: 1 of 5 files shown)


atdotde (commented):
Me likes a lot. I had tried to add this feature ages ago. At the time I tried to use http rather than email but never got that to work. Let’s see how much traffic this generates and how concerned people will be about privacy issues.

dirkhh:
I don't see any privacy concerns... this opens an email in the user's default email app. And they control what they send...
As for traffic - to avoid the obvious moderation issues right now these go just to me. Let's see how much more spam I get because of that...

dirkhh:
It's a pretty straight forward change and no one seems to hate it, so let's merge.

PR #2366, opened 2019-11-11
Show altitude corresponding to surface pressure

willemferguson:
In the information tab, presenting atmospheric pressure is a bit
unintuitive because the diver cannot easily relate that to altitude.

For the Atm. Pressure widget in the Information tab this code does:
If the atmospheric pressure for a dive exists and the user selects
the 'm' or 'ft' option from the combobox, then the estimated altitude
is shown in the text box.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @glance @bstoeger

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveInformation.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveInformation.cpp
@@ -232,9 +232,11 @@ void TabDiveInformation::updateTextBox(int event) // Either the text box has bee
 	double altitudeVal;             // maintained even though two independent events trigger saving the text box contents.
 	if (current_dive) {
 		switch (ui->atmPressType->currentIndex()) {
-		case 0:		// If an atm pressure has been specified in mbar:
+		case 0:		// If atm pressure in mbar has been selected:
 			if (event == TEXT_EDITED)         // this is only triggered by on_atmPressVal_editingFinished()
 				atmpress.mbar = ui->atmPressVal->text().toInt();    // use the specified mbar pressure
+			else                              // if no pressure has been typed, then show existing dive pressure
+				ui->atmPressVal->setText(QString::number(current_dive->surface_pressure.mbar));
 			break;
 		case 1:		// If an altitude has been specified:
 			if (event == TEXT_EDITED) {	// this is only triggered by on_atmPressVal_editingFinished()
@@ -247,7 +249,15 @@ void TabDiveInformation::updateTextBox(int event) // Either the text box has bee
 				ui->atmPressVal->setText(ui->atmPressVal->text().sprintf("%d",atmpress.mbar));
 				ui->atmPressType->setCurrentIndex(0);    // reset combobox to mbar
 			} else { // i.e. event == COMBO_CHANGED, that is, "m" or "ft" was selected from combobox
-				ui->atmPressVal->clear();   // Clear the text box so that altitude can be typed
+				 // Show estimated altitude
+				bool ok;
+				double convertVal = 0.0010;	// Metric conversion fro mm to m
+				int pressure_as_integer = ui->atmPressVal->text().toInt(&ok,10);
+				if (ok && ui->atmPressVal->text().length()) {  // Show existing atm press as an altitude:
+					if (prefs.units.length == units::FEET) // For imperial units
+						convertVal = mm_to_feet(1);    // convert from mm to ft
+					ui->atmPressVal->setText(QString::number((int)(pressure_to_altitude(pressure_as_integer) * convertVal)));
+				}
 			}
 			break;
 		case 2:          // i.e. event = COMBO_CHANGED, that is, the option "Use dc" was selected from combobox


desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -247,7 +247,12 @@ void TabDiveInformation::updateTextBox(int event) // Either the text box has bee
 				ui->atmPressVal->setText(ui->atmPressVal->text().sprintf("%d",atmpress.mbar));
 				ui->atmPressType->setCurrentIndex(0);    // reset combobox to mbar
 			} else { // i.e. event == COMBO_CHANGED, that is, "m" or "ft" was selected from combobox
-				ui->atmPressVal->clear();   // Clear the text box so that altitude can be typed
+				 // Show estimated altitude
+				bool ok;
+				int i = ui->atmPressVal->text().toInt(&ok,10);
```
dirkhh: `i` seems like an odd variable name here - that's typically an index in a loop or something.
And because of this I'm not sure I understand the next part of the code.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -247,7 +247,12 @@ void TabDiveInformation::updateTextBox(int event) // Either the text box has bee
...
+				if (ok && ui->atmPressVal->text().length()) {
+					ui->atmPressVal->setText(QString::number(pressure_to_altitude(i)/1000));
+				}
```
dirkhh, line 260: so the user changed the label to 'm' or 'ft', and then you convert the pressure to an altitude, so `i` should have been named `pressureValue` or somethings like that. But that's where I'm confused, because if the user changed this to 'm' or 'ft', don't we assume that the input value is in meters or feet?

Or is the logic that there is a pressure value in the field, the user switches to 'm', we convert that value into an estimated altitude, and THEN the user edits the value? I guess that would make more sense, but then a comment here in the code would be helpful (`// convert existing pressure value into estimated altitude`)

dirkhh (commented):
Small comment - I think the code is right in principle, just could use better variable name and maybe a comment

desktop-widgets/tab-widgets/TabDiveInformation.cpp
willemferguson: I discovered I did not provide for the imperial case. Here it is.
The logic is as follows:
1) Dive has no surface pressure value. However, diver knows altitude (e.g. 3000 ft)
2) Select ft from the combo box and type in the value of 3000 in the text box. Hit ENTER or click mouse elsewhere.
3) The estimated atm pressure for 3000 ft is calculated and shown (900 mbar)
4) Now let us assume the dive had an atm pressure of 900 mbar (maybe even from elsewhere e.g. dive computer)
    We are curious to know to what altitude this relates to.
5) The text box shows 900 and the combobox shows mbar.
6) Select ft in combobox. The estimated altitude in ft is now shown (some value near 3000 ft)
7) Selecting mbar from the combobox shows 900 mbar again, as previously.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -12,6 +12,9 @@
 #include "core/display.h"
 #include "core/divelist.h"
 
+#include <stdlib.h>
+#include <stdio.h>
+
```
dirkhh: why do we need these two headers? Did you have some `fprintf()` in there for testing and kept the `#include` lines? Nothing in the code you added should need these headers...

desktop-widgets/tab-widgets/TabDiveInformation.cpp
dirkhh: yeah, the explanation makes sense. Obviously this conversion is a "rough guess", but I can see how some people might appreciate it.
See my other comment about the include lines added (I think) by mistake...

willemferguson:
My fault. I tried to do two in-place edits on github and commit them within github. Of course it did not work as expected.
The changes are done as requested and pushed from local repository onto github.

dirkhh:
I find editing on GitHub with the Web UI almost impossible to get right. I almost always do things locally and then `push / push -f`

dirkhh (approved):
I think this looks good

dirkhh:
@glance- since you did such great reviews recently... would you have time to look?

glance- (approved):
Nice work Willem! Concise, on point and one single feature in this commit.

We could have a tad cleaner code if we had a nice helper who converted altitude to a display unit based on user settings. I also don't like that altitude_to_pressure uses plain uint32_t and not a typedef'ed type with the unit clear, but neither of those tings are problems with this code, but rather problems with our code in general, and I ain't going to put it on Willem to fix that for this, but its rather a note to self about things we can improve in the future.

PR #2368, opened 2019-11-11
 Notes tab: Use top RH for depth and duration fields

willemferguson:
Currently the top righthand part of the notes tab is used for
    showing and editing air temperature and water temperature. But
    these fields were moved over to the Information tab and are not
    required in the Notes tab any more. Rather use this space for the
    depth and duration data for manually-entered dives. Currently
    extra vertical space is created in the Notes tab for showing this
    field, resulting in inefficient use of screen space and
    inelegant layout. This code moves the Duration and Depth fields
    into the top righthand of the Notes tab.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @bstoeger @glance @sfuchs79

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -186,7 +186,6 @@ MainTab::~MainTab()
 void MainTab::hideMessage()
 {
 	ui.diveNotesMessage->animatedHide();
-	updateTextLabels(false);
 }
 
 void MainTab::closeMessage()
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
Also a great size for a commit.
I'm curious why you didn't combine the two commits into one PR, though. Is that a "how do I do this in git" issue? Or did you assume that they were better in separate PRs? Because these two are functionally indeed separate, but of course they ARE related, right?

PR #2372, opened 2019-11-11
Two trivial(?) cleanups

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

@dirkhh: This is a preparation PR for the unified list model. The idea is to try to avoid passing pointers through QML. As you probably remember, we had a myriad of problems with that.
### Changes made:

1) Fix typos
2) Introduce a bool IS_TRIP_ROLE.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -348,8 +348,7 @@ void DiveListView::clearTripSelection()
 
 	// we want to make sure no trips are selected
 	Q_FOREACH (const QModelIndex &index, selectionModel()->selectedRows()) {
-		dive_trip_t *trip = index.data(DiveTripModelBase::TRIP_ROLE).value<dive_trip *>();
-		if (!trip)
+		if (!index.data(DiveTripModelBase::IS_TRIP_ROLE).toBool())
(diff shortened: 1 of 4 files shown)


desktop-widgets/divelistview.cpp
```diff
@@ -348,8 +348,7 @@ void DiveListView::clearTripSelection()
 
 	// we want to make sure no trips are selected
 	Q_FOREACH (const QModelIndex &index, selectionModel()->selectedRows()) {
-		dive_trip_t *trip = index.data(DiveTripModelBase::TRIP_ROLE).value<dive_trip *>();
-		if (!trip)
+		if (!index.data(DiveTripModelBase::IS_TRIP_ROLE).toBool())
```
dirkhh, line 351: I like this!

dirkhh (approved):
Yes, I think that's the way to go here. Add the roles that we need to communicate the information.

I'll wait for GitHub actions before merging...

bstoeger:
Checks OK, so let's merge.

PR #2375, opened 2019-11-12
Fixes in tab widget UI layout

sfuchs79:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Three minor changes in tab widget UI layout and txt:
- Remove leading space in string "Gas name"
- Remove duplicate <item> entry
- Correct "leftMargin" and "rightMargin" to 0 everywhere

### Mentions:

@willemferguson

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveInformation.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveInformation.cpp
@@ -32,7 +32,6 @@ TabDiveInformation::TabDiveInformation(QWidget *parent) : TabBase(parent), ui(ne
 	QString CSSSetSmallLabel = "QLabel { color: mediumblue; font-size: " +                        /* // Using label height ... */
 		QString::number((int)(0.5 + ui->diveHeadingLabel->geometry().height() * 0.66)) + "px;}"; // .. set CSS font size of star widget subscripts
 	ui->scrollAreaWidgetContents_3->setStyleSheet("QGroupBox::title { color: mediumblue;} ");
-	ui->diveModeBox->setStyleSheet("QGroupBox{ padding: 0;} ");
 	ui->diveHeadingLabel->setStyleSheet(CSS_SET_HEADING_BLUE);
 	ui->gasHeadingLabel->setStyleSheet(CSS_SET_HEADING_BLUE);
 	ui->environmentHeadingLabel->setStyleSheet(CSS_SET_HEADING_BLUE);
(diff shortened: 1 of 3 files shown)


sfuchs79:
@willemferguson 
There is still one issue in the new layout, which I was not able to understand and fix:

grafik

dirkhh:
> @willemferguson
> There is still one issue in the new layout, which I was not able to understand and fix:
> 


my guess is that there's a line break in the string (translated or otherwise). Can you check?

willemferguson:
The qml (i.e. TabDiveInformation.ui) is clear. No line break. I also checked ui_tabDiveInformation.h, generated by the macro-compiler and no issue. But this is on Linux where the issue is not clearly shown as in Win. But if I understand correctly, ui_tabDiveInformation.h for a Win executable is exactly the same file used for a Linux build. There is not a Win macro-compiler, is there? Note that the headings for Qgroupboxes are inline with the frame in Win, but above the frame in Linux.

dirkhh:
Totally guessing, haven't had a chance to build it and try (and I have to head to a board meeting now, so I don't know when I'll have time to play with this), but I notice that `Dive mode` appears to be the only one in that row with `Qt::AlignBottom`
https://github.com/sfuchs79/subsurface/blob/e61ed5f9488068d8b1718382ff29a91eacf080c6/desktop-widgets/tab-widgets/TabDiveInformation.ui#L82

willemferguson:
This is likely a secondary problem that arose from the heading position problem. The important question is why is all the headings are placed more or less one text line lower in Windows than in Linux. Once this one is solved the Win problem with the dive mode QGroupBox is likely to be tractable.

willemferguson:
The Qt::AlignBottom keeps the combobox in the correct vertical position.

sfuchs79:
I found the reason for the wrong layout of the "dive mode" box. This line:
https://github.com/Subsurface-divelog/subsurface/blob/master/desktop-widgets/tab-widgets/TabDiveInformation.cpp#L35

When I remove it, everything looks nice:

grafik

willemferguson:
Thanks, Stefan. I can confirm that dropping line 35 results in an expected display of the tab in Linux. I would be happy if you included that within this PR. The issue that remains is why Windows displays the groupbox titles **in-line** with the frame of the box, but in Linux, **above** the box. I would love to see what it looks like on Mac. But that would probably be material for a different PR, not this one.

sfuchs79:
Done. Pushed a second commit with this fix.

dirkhh (approved):
Cool, glad you found the issue. This looks good to go

PR #2378, opened 2019-11-13
Core: set gas use to zero if no values are given

### LINKED ISSUE #2376: Dive information tab shows meaningless and wrong values for SAC and gas consumption
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

By clicking around in my dive list I can randomly produce insane values for SAC and gas consumption in the dive information tab.

Examples:
grafik
grafik
grafik

These are all screenshots from the very same dive. Produced by unselecting/selecting this dive. The wrong values happen for gases, where there is no cylinder pressure info at all. So expected would be having printed no info at all.

### Mentions:

@dirkhh @bstoeger @willemferguson

bstoeger:
In get_gas_used() the use was left uninitialized if there are neither
user- nor computer-supplied values. This gives random SACs in the UI.
Initialize to 0.

Fixes #2376.

Reported-by: Stefan Fuchs <sfuchs@gmx.de>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is supposed to fix #2376. All info in the commit message.
### Mentions:

@sfuchs79 @willemferguson

### DIFF
--- a/core/statistics.c
+++ b/core/statistics.c
@@ -376,6 +376,8 @@ volume_t *get_gas_used(struct dive *dive)
 		end = cyl->end.mbar ? cyl->end : cyl->sample_end;
 		if (end.mbar && start.mbar > end.mbar)
 			gases[idx].mliter = gas_volume(cyl, start) - gas_volume(cyl, end);
+		else
+			gases[idx].mliter = 0;
 	}
 
 	return gases;


core/statistics.c
```diff
@@ -376,6 +376,8 @@ volume_t *get_gas_used(struct dive *dive)
 		end = cyl->end.mbar ? cyl->end : cyl->sample_end;
 		if (end.mbar && start.mbar > end.mbar)
 			gases[idx].mliter = gas_volume(cyl, start) - gas_volume(cyl, end);
+		else
+			gases[idx].mliter = 0;
```
atdotde, line 380: I am not sure this is the correct thing to do. We simply don't know the value but that does not make it 0. Using zero when it was actually non-zero but unknown would result in a displayed SAC rate that is too small. We should rather compute the SAC rate only for those parts of the dive where a cylinder was breathed where we know how much.

atdotde: For example, I might be doing a dive with back gas (for which I have usage information) plus a deco gas (for which for example I forgot to memorise the initial pressure). Then I would like to get my SAC rate only for the part of the dive that used the back gas. And no value at all for the SAC rate if I had no usable gas information.

bstoeger: Both callers of that function explicitly treat `mliter == 0` as "don't consider this cylinder". If you think that a different mechanism should be used, feel free to submit a patch.

This PR just restores old behavior.

dirkhh: I think that @bstoeger  is right. @atdotde can we close this and merge?

dirkhh: ping @atdotde

bstoeger:
Feedback from @sfuchs79 (i.e. does this fix the issue?) would also be welcome.

sfuchs79:
@bstoeger 
Please give me time until the weekend. Then I will be able to test...

atdotde:
Thumbs up!

sfuchs79:
Tested -> bug fixed. Thanks!
And yes, I see consumption and SAC for every tank where I have entered pressure info. So e.g. if I used 3 tanks and have pressure info for 2 tanks, I see results for 2 tanks.

BTW: In the statistics tab with the global dive SAC rate the situation it different: There you won't see a SAC rate unless you have pressure for all used cylinders. So there with the example above I won't see a SAC rate. Only if I enter also pressure info for the 3rd cylinder I see a SAC there. But honestly one could argue that this is a feature because you can't present global dive SAC if a part of the dive is missing.

bstoeger:
Thanks for testing and approval - will commit.

Since I only use single cylinders, I have no opinion on the statistics tab. It shouldn't be hard to change in any case.

PR #2379, opened 2019-11-13
[RFC] Refactoring: move undo commands to top level

bstoeger:
In the future we might want to use undo-commands for mobile as
well (even if not implementing undo).

Therefore, move the undo-command source from desktop-widgets
to their own commands top-level folder.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

@dirkhh: this moves the command-code one level up, so that we might potentially use it from mobile. Please check it out. My first time with cmake - it wasn't painful, but I can't say that I enjoyed it either.


commands/command.cpp
```diff
@@ -1,10 +1,10 @@
...
 
 #include "command.h"
-#include "command_divelist.h"
-#include "command_divesite.h"
-#include "command_edit.h"
-#include "command_edit_trip.h"
+#include "divelist.h"
+#include "divesite.h"
+#include "edit.h"
+#include "edit_trip.h"
```
dirkhh: I don't think I'd do this - now we have two `divelist.h` files in `core` and in `commands`. That's just asking for trouble.

commands/divelist.cpp
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
 
-#include "command_divelist.h"
-#include "command_private.h"
+#include "divelist.h"
+#include "private.h"
 #include "desktop-widgets/mainwindow.h"
 #include "desktop-widgets/divelistview.h"
```
dirkhh: ahh... desktop UI include files...

dirkhh (changes requested):
So right now we are including `mainwindow.h` and that's a bit of a problem, right?
We can make this conditional to being built for desktop (and have the correct path to make it actually work - did you test compile this?), or we can go the extra mile and make the commands truly independent from the UI.

As for `cmake` - I have worked with a lot of different build systems. The hillariously terrible `imake` that was part of the X Window System 30 years ago. `qmake`, `autotools`, `meson`, `scons` and lately `bazel`. I find `cmake` to be one of the more tolerable systems that actually mostly make sense... can't say that for too many of the others :-)

commands/command.cpp
bstoeger: What do you suggest? `commands/command_divelist.h`?

Note that the only header included from outside is `commands/command.h`. Everything else is purely local to the commands-folder.

commands/divelist.cpp
bstoeger: Done.

bstoeger:
> As for cmake - I have worked with a lot of different build systems. The hillariously terrible imake that was part of the X Window System 30 years ago. qmake, autotools, meson, scons and lately bazel. I find cmake to be one of the more tolerable systems that actually mostly make sense... can't say that for too many of the others :-)

Off-topic: I have a tiny Qt application and so far I am quite happy with `qmake`. If I'd like to make it run on e.g. Windows is there a (meta-)build system that you would recommend?

dirkhh:
> Off-topic: I have a tiny Qt application and so far I am quite happy with `qmake`. If I'd like to make it run on e.g. Windows is there a (meta-)build system that you would recommend?

You can make it run on Windows with `qmake`. When we first started the Subsurface Qt port that's what we used as well. Unless you have more complex dependencies, there's no reason to go to `cmake` if you already have `qmake` working. That said, at this point I think I'm more familiar with `cmake` as that's what I spend most of my time fighting with :-)

commands/command.cpp
dirkhh: Yes, I think `#include "commands/command_divelist.h"` makes sense.
Admittedly, a bit verbose, but certain to avoid confusion :-)

bstoeger: Fine with me.

dirkhh (approved):
thanks

dirkhh:
I assume that we can drop the 'RFC' and merge this now. To avoid time lag issues, please feel free to merge if you agree, @bstoeger

bstoeger:
Cool - thanks.

PR #2382, opened 2019-11-14
/: Corrected wrong Qt dependency from Kirigami

janiversen:
Kirigami 5.62 uses QtQuick.Controls 2.5, which were introduced in
Qt 5.12 (see https://doc.qt.io/qt-5/qtquickcontrols-index.html)

update CMakeList.txt to demand min. 5.12 when building SUBSURFACE-MOBILE
update INSTALL text

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [x] Documentation change
- [ ] Language translation

if in doubt look at (as one of many files that use 2.5):
kirigami/src/controls/ListSectionHeader.qml:
import QtQuick.Controls 2.5 as QQC2

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -62,8 +62,8 @@ Getting Qt5
 
 We use Qt5 in order to only maintain one UI across platforms.
 
-Qt5.9.1 is the oldest version supported if ONLY building Subsurface (desktop version)
-Qt5.11 is the oldest version supported if also building Subsurface-mobile (ios/android)
+Qt5.9.1 is the oldest version supported if ONLY building Subsurface
+Qt5.12 is the oldest version supported if also building Subsurface-mobile
 
 Download the Open Source version from https://www.qt.io/download and 
 follow the Qt instructions to install it or alternatively follow the
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,8 +251,8 @@ set(QT_FIND_COMPONENTS Core Concurrent Widgets Network Svg Positioning Quick Loc
 if (SUBSURFACE_TARGET_EXECUTABLE MATCHES "DesktopExecutable")
 	find_package(Qt5 5.9.1 REQUIRED COMPONENTS ${QT_FIND_COMPONENTS} LinguistTools Test QuickTest)
 else()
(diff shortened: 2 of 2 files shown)


dirkhh:
I vaguely remember that it does, indeed, work with Qt 5.11.0 (as that was one of the platforms I tested it on). And it's possible that I manually patched the import to use 2.4.
But that's mostly a historical statement. I see absolutely no reason not to simply require Qt 5.12.
Subsurface-mobile can be built on the desktop for testing purposes, but its main use is of course iOS and Android, and there we build against Qt 5.12 and 5.13 for the official binaries

INSTALL
```diff
@@ -62,8 +62,8 @@ Getting Qt5
 
 We use Qt5 in order to only maintain one UI across platforms.
 
-Qt5.9.1 is the oldest version supported if ONLY building Subsurface (desktop version)
-Qt5.11 is the oldest version supported if also building Subsurface-mobile (ios/android)
+Qt5.9.1 is the oldest version supported if ONLY building Subsurface
+Qt5.12.5 is the oldest version supported if also building Subsurface-mobile
```
dirkhh: since I am building the Android binaries against 5.12.0 right now I believe that this statement is incorrect, though

dirkhh (changes requested):
While it makes sense to build against the latest update to 5.12, I haven't bothered to update my Android build environment... so we shouldn't claim that it requires a newer version that what's used in the official binaries.

janiversen:
Updated text.

FYI: I am actually only building/testing:
- macOS (Catalina) Desktop
- macOS (Catalina) MobileExecutable
- Mobile for IOS Simulator
- Mobile for IOS device
with the official master as well as a major new version.

I thought it would be good to mention which patch version it was tested against, and I presumed (wrongly) that the newest patch version is what is supported (Qt do no support old patch versions).

dirkhh (approved):
Thanks - I SHOULD be building the apps on the very latest Qt versions... but because of the awkward infrastructure issues every update causes a lot of work and so I tend to skip versions :-(

PR #2383, opened 2019-11-15
Fix use before init cases found by Coverty

atdotde:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/uemis-downloader.c
+++ b/core/uemis-downloader.c
@@ -603,7 +603,7 @@ static bool uemis_get_answer(const char *path, char *request, int n_param_in,
 			free(ans_path);
 			return false;
 		}
-		if (read(ans_file, tmp, 100) < 0) {
+		if (read(ans_file, tmp, 100) < 3) {
 			free(ans_path);
 			close(ans_file);
 			return false;
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -53,7 +53,8 @@ MainTab::MainTab(QWidget *parent) : QTabWidget(parent),
 	lastSelectedDive(true),
 	lastTabSelectedDive(0),
 	lastTabSelectedDiveTrip(0),
-	currentTrip(0)
+	currentTrip(0),
+	modified(false)
 {
 	ui.setupUi(this);
 
--- a/core/dive.c
+++ b/core/dive.c
@@ -246,9 +246,11 @@ void add_extra_data(struct divecomputer *dc, const char *key, const char *value)
 enum divemode_t get_current_divemode(const struct divecomputer *dc, int time, const struct event **evp, enum divemode_t *divemode)
 {
 	const struct event *ev = *evp;
-	if (*divemode == UNDEF_COMP_TYPE) {
+	if (*divemode == UNDEF_COMP_TYPE && dc) {
 		*divemode = dc->divemode;
-		ev = dc ? get_next_event(dc->events, "modechange") : NULL;
+		ev = get_next_event(dc->events, "modechange");
+	} else {
+		ev = NULL;
 	}
 	while (ev && ev->time.seconds < time) {
 		*divemode = (enum divemode_t) ev->value;
--- a/core/import-csv.c
+++ b/core/import-csv.c
@@ -543,6 +543,7 @@ int parse_txt_file(const char *filename, const char *csv, struct dive_table *tab
 		cyl.type.workingpressure.mbar = 200000;
 		cyl.type.description = "3l Mk6";
 		cyl.gasmix.o2.permille = 1000;
+		cyl.manually_added = true;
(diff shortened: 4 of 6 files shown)


glance-:
You shoud add Fixes CID XXX to each of the commit messages.

atdotde:
I wasn't aware of that. Actually, I _hope_ it fixes the issue (and the uninitialised structures are not accessed anywhere further down the line). Does the assertion "Fixes CID XXXX" imply this is not looked into anymore?

glance-:
I have some vague remembrance that when signing up for coverity you agreed to give them credit for findgin the issue, but I can't find that any more so it might be a ghost in my brain.

Anyhow, what it does is to flipp the state to "Fix submitted" and if the issue reoccurs one can find it in the log how it was fixed the previous time.

atdotde:
Spent some more time on Coverty results and fixed some obvious ones.

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -75,6 +75,8 @@ MainTab::MainTab(QWidget *parent) : QTabWidget(parent),
 
 	memset(&displayed_dive, 0, sizeof(displayed_dive));
 
+	modified = false;
```
bstoeger: Put this in initializer list, please.

atdotde: updated. Now I am a real C++ programmer ;-)

dirkhh:
We as a project want to thank Coverity for the free use of their service by giving them credit. So whenever we fix a bug that was found by Coverity (or, sometimes, that we found because of a false positive in Coverity sent us chasing for why it thought something was wrong, and in the process find something different that is actually wrong) we want to give them credit in the commit message. That does NOT mask the bug.

qt-models/weightmodel.cpp
```diff
@@ -102,7 +102,7 @@ bool WeightModel::setData(const QModelIndex &index, const QVariant &value, int r
 			if (!ws->description || gettextFromC::tr(ws->description) != vString) {
 				// loop over translations to see if one matches
 				int i = -1;
-				while (i < MAX_WS_INFO && ws_info[++i].name) {
+				while (i < MAX_WS_INFO - 1 && ws_info[++i].name) {
```
dirkhh: My guess is that this will change once we no longer have a fixed array of weight systems

core/uemis-downloader.c
```diff
@@ -603,7 +603,7 @@ static bool uemis_get_answer(const char *path, char *request, int n_param_in,
 			free(ans_path);
 			return false;
 		}
-		if (read(ans_file, tmp, 100) < 0) {
+		if (read(ans_file, tmp, 100) < 100) {
```
dirkhh: I haven't played with the UEMIS downloader in ages because all of the Uemis dive computers that I had access to have long since died.
Every time I look at the code my eyes bleed...
I am not sure if we are guaranteed that a 100 characters can be read here. Reading the code we only ever appear to look at the first 4 bytes of this buffer? Am I missing something?
So this one I don't think is necessarily right and it could change behavior of Subsurface if indeed we get fewer than 100 bytes back in a valid read...

dirkhh (commented):
That last one I have concerns with. Can you take another look? Are we accessing the rest of that buffer anywhere?

qt-models/weightmodel.cpp
bstoeger: Careful: this is not an array of weightsystems, but an array of weightsystem infos. The fixed array of weightsystems is long gone. :)

But yes, that loop is horrible, and if I remember correctly, I removed it in #2363. Which makes me wonder if it is worth generating a merge-conflict here.

And which reminds me that I have to look at #2363 again...

core/uemis-downloader.c
atdotde: You are right, checking against 100 is definitely wrong. I was too quick and only saw the tmp[100] in the debug code before not even realising that that is writing rather than reading. I also have to admit I confused the return value of read with fread. But what would be the correct value, maybe 3 (as we are using up to tmp[2] below?

qt-models/weightmodel.cpp
atdotde: I am happy to leave it as it is. Even if we access out of bounds, that is only read access, so no danger of overwriting other data.

dirkhh: It felt wrong when I wrote my comment above. Urgs. Sorry @bstoeger 

Yes, let's drop this one and assume that #2363 will take care of it.

core/uemis-downloader.c
dirkhh: I re-read the code. I believe we need to get at least 3 bytes to be able to process the buffer.

dirkhh (approved):
thanks

PR #2384, opened 2019-11-15
Dive list: signal correct trip in DiveTripModelTree::topLevelChanged

bstoeger:
DiveTripModelTree::topLevelChanged() has pretty complex code, as
it has to handle the fact that when adding/removing a dive from
a trip, the trip can change it's position.

It got the common case (trip stays where it is) wrong: it sent
a signal to redraw the next trip. Catch that special case.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a bug fix for a bug detected when adapting the model code for mobile. Weirdly enough, desktop works even with this bug.
### Changes made:

1) Signal correct trip, not the trip after the correct trip.

### Release note:

No - for some reason not visible on desktop. Apparently, the header is updated automatically if a child is updated.

### Documentation change:

No.
### Mentions:

@dirkhh

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -759,6 +759,11 @@ void DiveTripModelTree::topLevelChanged(int idx)
 		endMoveRows();
 	}
 
+	// If we moved the object backwards in the array, we have to
+	// subtract one from the index to account for the removed object.
+	if (newIdx > idx)
+		--newIdx;
+
 	// Finally, inform UI of changed trip header
 	QModelIndex tripIdx = createIndex(newIdx, 0, noParent);
 	dataChanged(tripIdx, tripIdx);


qt-models/divetripmodel.cpp
```diff
@@ -752,6 +752,10 @@ void DiveTripModelTree::topLevelChanged(int idx)
 			++newIdx;
 	}
 
+	// Moving an object behind itself is a noop
+	if (newIdx == idx + 1)
```
dirkhh: I think that comment is slightly misleading. Or could be read in a misleading way. If my objects are
``1 2 3 4 5``
and I move `3` to spot `4` then that isn't a no-op, it results in 
``1 2 4 3 5``
But here in this code `newIdx == idx +1` is actually the indicator that we don't need to move the entry, based on the code above it.

dirkhh (changes requested):
I agree with your code, that's obviously correct, but I think your comment is misleading.

qt-models/divetripmodel.cpp
bstoeger: Well, but in your case 3 is moved behind 4, not behind itself. Or would you read "move behind" as "move anywhere behind"?

dirkhh: I'm reading the code and the comment in a rather literal way.
The new index is one bigger than the current index.
That's why I say your code is right, but your comment could be misleading.
How about
```// check if we need to move the object at all```

bstoeger: Well, thinking more about this, it turns out that code, comment *and* commit message were all wrong. :)

Hopefully correct now.

dirkhh (approved):

PR #2387, opened 2019-11-15
Partially revert 3025e0630de2b9ced3d0aadd1513cb0e1c93b159

bstoeger:
This commit did the "right" thing by implementing Qt mode semantics
as intended, but for unknown reasons the profile is not properly
cleared on close-file anymore. This code is so convoluted that there
is not point in fighting it at the moment. Revert to remove-rows
instead of reset-model.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a recently introduced bug by doing a semi-revert. The code is too convoluted to find out what is actually happening.

### Release note:

No - bug not in release.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1446,6 +1446,7 @@ void clear_dive_file_data()
 {
 	while (dive_table.nr)
 		delete_single_dive(0);
+	current_dive = NULL;
 	while (dive_site_table.nr)
 		delete_dive_site(get_dive_site(0, &dive_site_table), &dive_site_table);
 	if (trip_table.nr != 0) {
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -658,7 +658,6 @@ void MainWindow::closeCurrentFile()
 	/* free the dives and trips */
 	clear_git_id();
 	clear_dive_file_data();
-	current_dive = nullptr;
 	setCurrentFile(nullptr);
 	graphics->setEmptyState();
 	mainTab->clearTabs();
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -657,12 +657,8 @@ void MainWindow::closeCurrentFile()
 {
 	/* free the dives and trips */
 	clear_git_id();
-	clear_dive_file_data();
+	DiveTripModelBase::instance()->clear();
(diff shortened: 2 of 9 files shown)


dirkhh (commented):
No idea why one works and the other doesn't. I trust you here :-)

bstoeger:
Added another cleanup in the same area (clearing of data).

bstoeger:
Argh - this still crashes if you move with the mouse over the profile. Very frustrating!

bstoeger:
Alright - that was a different ancient bug that became active due to the new cylinder code. I wonder how many of these surprises are still lurking there...

bstoeger:
@dirkhh: The last commit introduces a `DiveTripModelBase::instance()->clear()` function. If you call that instead of the disconnect-model thing, there should be no way of invalid accesses. The cleared data should bubble up the different layers. If crashes still happen, please let me know so that I can try to figure out what's going on.

dirkhh:
> Alright - that was a different ancient bug that became active due to the new cylinder code. I wonder how many of these surprises are still lurking there...

Many. Many.

dirkhh:
> @dirkhh: The last commit introduces a `DiveTripModelBase::instance()->clear()` function. If you call that instead of the disconnect-model thing, there should be no way of invalid accesses. The cleared data should bubble up the different layers. If crashes still happen, please let me know so that I can try to figure out what's going on.

That sounds cool. And that would make the QML code so much cleaner.
I will play with this for sure!
THANKS

core/divelist.c
```diff
@@ -1446,6 +1446,7 @@ void clear_dive_file_data()
 {
 	while (dive_table.nr)
 		delete_single_dive(0);
+	current_dive = NULL;
```
dirkhh, line 1449: the commit message seems wrong, though. Or at least it confuses me.
In general, we expect current_dive to be non-NULL, right?

dirkhh (commented):
This all looks great. One commit message seems to have a typo in the title, one confuses me with the reference to non-NULL current_dive. Could you review those (I know it's late in Austria)

core/divelist.c
bstoeger: Better?

dirkhh (approved):
Excellent work...
Much better this way.

PR #2389, opened 2019-11-17
Attempt at filter detangling

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is an attempt at making the filter code
1) somewhat more clear
2) unified for mobile and desktop

I'm not yet very happy about the results. I'm sure that this can be done in a more controlled way. Data flow is still a complete mess. But I hope that it is a step in the right direction.

What this basically does:
1) Create a class that is only responsible for filtering.
2) Move the logic to set the hidden_by_filter flag closer to the core. Still, this is spread out over DiveTripModel, undo-commands and core in a rather nasty way.

@dirkhh: If you agree with the general direction I will try if I can make the new mobile code run with this. But I will only be able to do so tomorrow evening.

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -376,6 +376,7 @@ void DiveTripModelBase::clear()
 {
 	beginResetModel();
 	clear_dive_file_data();
+	emit diveListNotifier.divesSelected({}, nullptr); // Inform profile, etc of changed selection
 	endResetModel();
 }
 
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -1377,7 +1377,6 @@ void DiveTripModelList::filterFinished()
 	// In list mode, we don't have to change anything after filter finished.
 }
 
-
 // Simple sorting helper for sorting against a criterium and if
 // that is undefined against a different criterium.
 // Return true if diff1 < 0, false if diff1 > 0.
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -376,6 +376,7 @@ void DiveTripModelBase::clear()
 {
 	beginResetModel();
 	clear_dive_file_data();
+	clearData();
 	emit diveListNotifier.divesSelected({}, nullptr); // Inform profile, etc of changed selection
 	endResetModel();
 }
@@ -605,6 +606,11 @@ int DiveTripModelTree::rowCount(const QModelIndex &parent) const
 	// Only trips have items
 	const Item &entry =  items[parent.row()];
 	return entry.d_or_t.trip ? entry.dives.size() : 0;
+}
+
+void DiveTripModelList::clearData()
+{
+	items.clear();
 }
 
 static const quintptr noParent = ~(quintptr)0; // This is the "internalId" marker for top-level item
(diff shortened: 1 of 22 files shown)


dirkhh (commented):
I hate that I can't comment on commit messages - I brought this up with GitHub again this week. A couple of them are just slightly awkwardly phrased. But nothing catastrophic.
The direction seems valid. The last commit is something that I need to read one more time, I'm not 100% sure I understand what you did there.
From a file system layout perspective (recognize a bike-shedding attempt, please) - I don't think the `divefilter` classes necessarily should be in `mobile-widgets` and `desktop-widgets`. They simply aren't UI widgets. I think they'd be more logically in the models code or the core code, with distinct file names but identical class names. But again, that's not really a critical comment.
The fundamental idea of what you are doing here makes a ton of sense to me. Because I really struggled tracing through the various interdependent classes what happened where. I think this will make the code much cleaner.

bstoeger:
> From a file system layout perspective (recognize a bike-shedding attempt, please) - I don't think the divefilter classes necessarily should be in mobile-widgets and desktop-widgets.

Yes, I was also thinking about that - but I didn't know where to put it. Perhaps even in `core` and with a big `#ifdef` for mobile vs. desktop?

dirkhh:
> Yes, I was also thinking about that - but I didn't know where to put it. Perhaps even in `core` and with a big `#ifdef` for mobile vs. desktop?

Honestly, I think that would be my preference.

bstoeger:
Quick rebase - I hope I didn't mess up things.

dirkhh (commented):
Reviewing the changes: looking good. I like the way this is layered now much better.
I haven't had a chance to build / try this, so not quite ready to "approve" :-)

qt-models/divetripmodel.cpp
```diff
@@ -376,6 +376,7 @@ void DiveTripModelBase::clear()
 {
 	beginResetModel();
 	clear_dive_file_data();
+	clearData();
```
dirkhh, line 379: Ahhh - that was why things were crashing for me on mobile without disconnecting the model.
We were using the pointers in the items...

bstoeger: Ultimately, there should be no reason to disconnect the model.

bstoeger:
>  I like the way this is layered now much better.

Sometimes I have a feeling of spinning in circles, but let's see...

qt-models/divetripmodel.cpp
dirkhh: So there were two distinct reasons:
(a) when applying the GPS fixes, the ListViews (both of them) didn't notice that there was a change in the dives and the GPS fixes wouldn't show up until after you restarted the app. I tried figuring out the data changed signals back then and couldn't. I think you have that addressed (at least I read that in one of the commits in one of your branches). I haven't tested that, yet.
(b) when changing dive files we were dereferencing bogus pointers that came from the items array. I think this should be fixed with this.

dirkhh:
> > I like the way this is layered now much better.
> 
> Sometimes I have a feeling of spinning in circles, but let's see...

What I would love, when we are done, is a nice diagram how the models and proxies and filters are layered and which functions one should call at which level to do what.
Because I'm still thoroughly confused when I need to talk to the SortFilterProxy, when I talk to the base class, when I talk to one of the classes in the middle.
I think laying this out clearly would be extremely helpful to make sure the two of us understand if what we got to was what we wanted.

dirkhh:
Just so you don't think I have forgotten this... I tested this on the desktop and it seems to work fine, I wasn't able to find a way too break things.
So from my side this is a tentative "I think this is ok".
You still have this marked as "RFC", so I won't merge :-)

bstoeger:
Removing RFC tag - I'm planning a more controversial followup-PR in the same area.

PR #2392, opened 2019-11-21
Bt updates

### LINKED ISSUE #2358: Subsurface Mobile "Show all bluetooth devices..." is not sticky over reboots
### Describe the issue:
- [ x] Bug
After a reboot, the bluetooth setting "Show all devices..." is unchecked.

### Issue long description:

In order to get a list of bluetooth devices to show up in the download screen you may need to enable this option. All known bluetooth devi ec then appear in the connection list. After a reboot of the phone, this option is turned off and no bluetooth devices appear in the connection list.

Note that in my world none of the bluetooth devices were "recognized" dive computers. this is a secondary issue. the real problem is that the option turned itself off, leading to inconsistent behaviour for the end user.
 
### Operating system:

Android 10

Essential PH-1

### Subsurface version:

Subsurface mobile 2.2.2 (4.9.3.119)

Subsurface mobile downloaded from android store thingy

### Steps to reproduce:

set the option
restart phone
option is no longer set...
### Expected behavior:

Option should be retained over restarts

dirkhh:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

two small fixes.
One subtle update to the detection code for the OSTC family of computers - picking OSTC 3 ended up causing us to not recognize certain BLE versions of the OSTC on iOS as dive computers.
And a change to the language on the settings page that clarifies that the "show all BT devices" settings is not persistent.

### Related issues:

Fixes #2358 

### Release note:

done

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -30,9 +30,10 @@ static dc_descriptor_t *getDeviceType(QString btName)
 		else if (btName.mid(4,2) == "4-") product = "OSTC 4";
 		else if (btName.mid(4,2) == "2-") product = "OSTC 2N";
 		else if (btName.mid(4,2) == "+ ") product = "OSTC 2";
-		// all OSTCs are HW_FAMILY_OSTC_3, so when we do not know,
-		// just try this
-		else product = "OSTC 3"; // all OSTCs are HW_FAMILY_OSTC_3
+		// all BT/BLE enabled OSTCs are HW_FAMILY_OSTC_3, so when we do not know,
+		// just use a default product that allows the codoe to download from the
+		// user's dive computer
+		else product = "OSTC 2";
 	}
 
 	if (btName.startsWith("Predator") ||
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: ensure that all BT/BLE flavors of the OSTC are recognized as dive computers [#2358]
 Desktop: allow copy&pasting of multiple cylinders [#2386]
 Desktop: don't output random SAC values for cylinders without data [#2376]
 Mobile: add menu entry to start a support email, using native email client
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -559,7 +559,7 @@ Kirigami.ScrollablePage {
 			}
 
 			Controls.Label {
-				text: qsTr("Show all bluetooth devices \neven if not recognized as dive computers")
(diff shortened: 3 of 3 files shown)


Subsurface-divelog:
@glance- since you know the OSTC family of dive computers better than I do... does this seem reasonable to you? This code will only run if the dive computer was detected, so we should only assign vendor/product pairs that will then show up as available, i.e. we need too pick BT/BLE enabled products...

core/btdiscovery.cpp
```diff
@@ -31,8 +31,8 @@ static dc_descriptor_t *getDeviceType(QString btName)
 		else if (btName.mid(4,2) == "2-") product = "OSTC 2N";
 		else if (btName.mid(4,2) == "+ ") product = "OSTC 2";
 		// all OSTCs are HW_FAMILY_OSTC_3, so when we do not know,
```
glance-: This comment isn't that correct. Not all OSTC's are HW_FAMILY_OSTC_3, but, rather, all with bluetooth are HW_FAMILY_OSTC_3.

glance-: But, there's also the Frog, which is a HW computer with bluetooth, which isn't a HW_FAMILY_OSTC_3, but there weren't that many of them made, and I've actually never seen one, so lets ignore that =)

core/btdiscovery.cpp
```diff
@@ -31,8 +31,8 @@ static dc_descriptor_t *getDeviceType(QString btName)
...
-		// just try this
-		else product = "OSTC 3"; // all OSTCs are HW_FAMILY_OSTC_3
```
glance-, line 35: Yea, the "OSTC 3" didn't have bluetooth, so thats probably why it fell away somewhere along the way. The "OSTC Plus" (3+) all as bluetooth.

glance-:
Looks reasonable. I commented about some odds and ends to try to clarify things.

glance- (approved):

core/btdiscovery.cpp
Subsurface-divelog: I kept it as "OSTC 2" as the default product name = I saw no advantage naming it "OSTC Plus"
But I updated the comment to be clearer.

core/btdiscovery.cpp
Subsurface-divelog: I had one conversation with a Frog owner a few years ago... but yes, they are being ignored here because I don't even know the BT name... :-)

PR #2394, opened 2019-11-21
AppImage: don't bundle our libdbus

Subsurface-divelog:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

simply delete the offending library before bundling

### Related issues:

Fixes: #2370

### Release note:

DONE

### Mentions:

@glance- can you please check that the resulting AppImage works for you?
Since this is a PR, you'll need to go to the Action for Linux Qt5.12 and download it from the Artifacts drop down in the upper right area of the page

PR #2395, opened 2019-11-21
Build container images in GitHub actions

glance-:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Build our ci containers in our ci system

### Changes made:
This moves the build tasks of manually building our containers for our ci system to building them on our ci, github actions.

### Related issues:
When we where working with #2394 for example, its quite clear that changes to our ci containers are painfull and hard to do, and can't be done by anyone else than @dirkhh , so this solves that by moving it to github actions, and that way opening up for others to help out.

### Additional information:
Because docker hates my github username because it contains a "-" in it, I can't really test this, so there might be bugs.

### Mentions:
@dirkhh

### DIFF
--- a/.github/workflows/dockerimages.yml
+++ b/.github/workflows/dockerimages.yml
@@ -1,0 +1,72 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - scripts/docker/android-build-container/Dockerfile
+    - scripts/docker/mxe-build-container/Dockerfile
+    - scripts/docker/trusty-qt512/Dockerfile
+
+
+jobs:
+  android-build-container:
+    runs-on: ubuntu-latest
+    env:
+      VERSION: 5.13.01
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Get our pre-reqs
+      run: |
+          cd scripts/docker/android-build-container
+          bash download.sh
+          sed -ie 's/^sudo/#sudo/' setup-docker.sh
+          bash setup-docker.sh
+
+    - name: Build the Docker image
+      run: docker build ./scripts/docker/android-build-container/ --tag docker.pkg.github.com/${GITHUB_REPOSITORY}/android-build-container:${VERSION}
+
+    - name: Push the Docker image
+      run: |
+        docker login docker.pkg.github.com -u ${GITHUB_REPOSITORY/\/*/} -p ${{ secrets.GITHUB_TOKEN }}
+        docker push docker.pkg.github.com/${GITHUB_REPOSITORY}/android-build-container:${VERSION}
+
+  mxe-build-container:
+    runs-on: ubuntu-latest
+    timeout-minutes: 720
+    env:
+      VERSION: 0.9
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Build the Docker image
+      run: docker build ./scripts/docker/mxe-build-container/ --tag docker.pkg.github.com/${GITHUB_REPOSITORY}/mxe-build-container:${VERSION}
+
+    - name: Push the Docker image
+      run: |
+        docker login docker.pkg.github.com -u ${GITHUB_REPOSITORY/\/*/} -p ${{ secrets.GITHUB_TOKEN }}
+        docker push docker.pkg.github.com/${GITHUB_REPOSITORY}/mxe-build-container:${VERSION}
+
+  trusty-qt512:
+    runs-on: ubuntu-latest
+    env:
+      VERSION: 0.7
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Get our pre-reqs
+      run: |
+        cd scripts/docker/trusty-qt512
+        bash getpackages.sh
+
+    - name: Build the Docker image
+      run: docker build ./scripts/docker/trusty-qt512/ --tag docker.pkg.github.com/${GITHUB_REPOSITORY}/trusty-qt512:${VERSION}
+
+    - name: Push the Docker image
+      run: |
+        docker login docker.pkg.github.com -u ${GITHUB_REPOSITORY/\/*/} -p ${{ secrets.GITHUB_TOKEN }}
+        docker push docker.pkg.github.com/${GITHUB_REPOSITORY}/trusty-qt512:${VERSION}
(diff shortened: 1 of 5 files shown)


dirkhh:
There is the option of creating a second GitHub account without a `-`. I'll pay the GitHub charges for that in return for you help and contribution here :-)

.github/workflows/dockerimages.yml
```diff
@@ -6,7 +6,7 @@ on:
     - master
     paths:
     - scripts/docker/android-build-container/Dockerfile
-    - scripts/docker/mxe-build-container/Dockerfile
+#    - scripts/docker/mxe-build-container/Dockerfile
```
dirkhh: can this be done in stages? I.e. create a first partial build and push it and use that as the base for a second part?
Awkward, but hey, you know - I like the underlying idea of this not just being something only I can do **quite a bit**

dirkhh (commented):
Obviously we wouldn't want to introduce something and then take immediately change it in the next commit, but as I said in my comment... is there a way around that 6h limit by splitting things up?

.github/workflows/dockerimages.yml
glance-: I've thought about it, but I haven't gotten around to testing it, and after not getting around to it for long enough I'd just rather go with the bits that works, and leave the mxe-bits "as-is" to be continued on at some other time (or by someone else).

Subsurface-divelog: That makes sense.
I'm curious, do you have a partial run of this (i.e., where was the process at the 6h mark)?

dirkhh: (sorry, logged into the wrong account - I need to be Subsurface-divelog to play with the webhook, but I usually try to comment as "me")

glance-: https://github.com/glance-/subsurface/runs/281752387

One if the big thingies is spending 3h building qtwebkit.

dirkhh: > https://github.com/glance-/subsurface/runs/281752387
> 
> One if the big thingies is spending 3h building qtwebkit.

Thanks for sharing that. So to me this looks like there are three easy to identify logical blocks:

- create stage 1 container with qt
- create stage 2 container with qtwebkit
- create final container with everything ready

The mechanics to do that are very simple, I want to try that this weekend. The biggest question really is how to trigger this as actions so that we don't get bitten by the 6h limit without having to do very unnatural things to our repo. I'm reading up on the rules for GitHub actions to understand that better...

dirkhh: So according to https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions the rule is "Each job in a workflow can run for up to 6 hours of execution time."
That should allow a *really* simple setup for this.
I'm not sure I'll have the time to try this out today, but I'll definitely find time over the weekend - this may be quite simple

glance-:
One thing I'm fighting here is githubs templating and dockers ideas about how a string should look.

 tried to write the workflow thingies repo intependant, so one could develop on new images in a separate repo, and it would run them there, so I'm using templating to configure which repo to pull the images from, but "Subsurface-divelog/subsurface" contains a capital "S" and docker doesn't like that. Also my username ends with a - and docker also hates that.

There's no toLower()/lower() in the github actions templating thingie, and I'm starting to loose hope of getting those bits working as i thought, so soon I might just give up and hard code "subsurface-divelog/subsurface"

dirkhh:
I commented about your name before. I think the capital 'S' is just cosmetic - is there anything that fails if you use a lower case 's'? Do I need to switch the name of the org (I can do that if it makes a difference - but we'd have to make sure that that doesn't break other things depending on us)

glance-:
Another blocker that I didn't notice is:
https://github.community/t5/GitHub-Actions/Github-Actions-new-Pulling-from-private-docker-repositories/m-p/32007#M973
https://github.community/t5/GitHub-Actions/Use-docker-images-from-GitHub-Package-Registry/m-p/31353#M777

dirkhh:
There does appear to be a workaround, though: https://github.community/t5/GitHub-Actions/Use-docker-images-from-GitHub-Package-Registry/m-p/36809#M2692
I need to understand the downsides of that better...

.github/workflows/dockerimages.yml
dirkhh: > So according to https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions the rule is "Each job in a workflow can run for up to 6 hours of execution time."
> That should allow a _really_ simple setup for this.
> I'm not sure I'll have the time to try this out today, but I'll definitely find time over the weekend - this may be quite simple

I did kick the test off last night and no, this is not simply a "per step" maximum - splitting things up is insufficient to do this. On the flip side I was able to have it push to Docker Hub, so at least one of the issues is solvable. The issue will be one of documentation as this requires the person playing with this to create their own docker hub account (and no, that does NOT have to have the same user name as your GitHub account) and put the credentials into secrets on GitHub.

Now I need to come up with a creative way around the time limit problem. 

I do wonder, thought, how much help this will bring towards updating these images as that is still a remarkably painful experience. And in many ways it is less painful doing this on your own machine as that is likely more powerful and therefore quicker than that GitHub Action VMs. So why would someone interested in helping do that as an Action in the first place instead of just doing that on their own laptop? But I guess the principle of having this happen "out in the open" instead of "behind closed doors on Dirk's machine" by itself is valuable.

glance-:
I tink we can do something with events. Workflow 1 builds a "stage 1" and push it out, and trigger a event. Workflow 2 start on that event and builds "stage 2" on top of "stage 1" and so on.
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows

dirkhh:
Uhh, that's neat. I think I have all the pieces in place to use that, I just need to figure out how to set up the glue to trigger those. Googling for examples got me some ideas.
I have a few errands to run, but should have time to play with this this afternoon.
Thanks @glance- that's a really clever way around the limitations...

dirkhh:
I made a ton of progress.
It's not working completely, yet, but those are syntax issues and me fighting with yaml.
I have a prototype in a different repo that I'm cleaning up and I think I should have enough time tomorrow to merge your work and mine and make this available. Fingers crossed.

dirkhh:
Given the progress on #2401 I assume that you're ok with me closing this one.

PR #2396, opened 2019-11-23
Added export functionality to subsurface-mobile (only file export)

janiversen:

- [ ] Bug fix
- [x ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add the possibility to do export directly in Subsurface-mobile.

The dialogs are kept close to the desktop version, with needed differences
(f.x. Subsurface-mobile to not have "selected dives" so all dives are exported).

The functionality is in principle the same as the desktop, and the "old" export 
functions have been moved from desktop-widgets to core in order to be shared.

There were a couple of problems found during implementation:

1) generating a PNG file (export profile) is intertwined with the UI and needs to 
be done completely different in mobile. Provisions are made to have 2 different
implementations, but the mobile implementation is open.

2) DiveFilter have 2 different implementations, and the mobile one is very limited, therefore
"exporting dive sites" would have needed a completely different implementation. The better
solution is to make the full functionality of DiveFilter available to Subsurface-mobile, but
that is outside the scope of this PR.

3) Upload to DiveShare and DiveLogs.de, the current web service implementation is 
intertwined with the UI and need to be detangled.

4) The design could use improvements.

These will be addressed later by me, or others.

### Changes made:

See Commits, which each are small, and considered documented.

### Additional information:

Allowed myself to include a commit (first one) even though it is not related to the patch, but
administrative important.

### Release note:

Added change to CHANGELOG.md

### Documentation change:

GlobalDrawer contains a new entry "Export"
New Export page
New File dialog (standard QT version)

### DIFF
--- a/core/exportfuncs.cpp
+++ b/core/exportfuncs.cpp
@@ -1,0 +1,311 @@
+// SPDX-License-Identifier: GPL-2.0
+#include <QFileInfo>
+#include <QDir>
+#include <QtConcurrent>
+#include "core/membuffer.h"
+#include "core/divesite.h"
+#include "core/tag.h"
+#include "core/file.h"
+#include "core/errorhelper.h"
+#include "core/divefilter.h"
+#include "exportfuncs.h"
+
+
+exportFuncs *exportFuncs::instance()
+{
+    static exportFuncs *self = new exportFuncs;
+    return self;
+}
+
+void exportFuncs::exportProfile(QString filename, const bool selected_only)
+{
+	struct dive *dive;
+	int i;
+	int count = 0;
+	if (!filename.endsWith(".png", Qt::CaseInsensitive))
+		filename = filename.append(".png");
+	QFileInfo fi(filename);
+
+	for_each_dive (i, dive) {
+		if (selected_only && !dive->selected)
+			continue;
+		if (count)
+			saveProfile(dive, fi.path() + QDir::separator() + fi.completeBaseName().append(QString("-%1.").arg(count)) + fi.suffix());
+		else
+			saveProfile(dive, filename);
+		++count;
+	}
+}
+
+
+void exportFuncs::export_TeX(const char *filename, const bool selected_only, bool plain)
+{
+	FILE *f;
+	QDir texdir = QFileInfo(filename).dir();
+	struct dive *dive;
+	const struct units *units = get_units();
+	const char *unit;
+	const char *ssrf;
+	int i;
+	bool need_pagebreak = false;
+
+	struct membuffer buf = {};
+
+	if (plain) {
+		ssrf = "";
+		put_format(&buf, "\\input subsurfacetemplate\n");
+		put_format(&buf, "%% This is a plain TeX file. Compile with pdftex, not pdflatex!\n");
+		put_format(&buf, "%% You will also need a subsurfacetemplate.tex in the current directory.\n");
+	} else {
+		ssrf = "ssrf";
+		put_format(&buf, "\\input subsurfacelatextemplate\n");
+		put_format(&buf, "%% This is a plain LaTeX file. Compile with pdflatex, not pdftex!\n");
+		put_format(&buf, "%% You will also need a subsurfacelatextemplate.tex in the current directory.\n");
+	}
+	put_format(&buf, "%% You can download an example from http://www.atdotde.de/~robert/subsurfacetemplate\n%%\n");
+	put_format(&buf, "%%\n");
+	put_format(&buf, "%% Notes: TeX/LaTex will not render the degree symbol correctly by default. In LaTeX, you may\n");
+	put_format(&buf, "%% add the following line to the end of the preamble of your template to ensure correct output:\n");
+	put_format(&buf, "%% \\usepackage[utf8]{inputenc}\n");
+	put_format(&buf, "%% \\usepackage{gensymb}\n");
+	put_format(&buf, "%% \\DeclareUnicodeCharacter{00B0}{\\degree}\n"); //replaces ° with \degree
+	put_format(&buf, "%%\n");
+
+	/* Define text fields with the units used for export.  These values are set in the Subsurface Preferences
+	 * and the text fields created here are included in the data fields below.
+	 */
+	put_format(&buf, "\n%% These fields contain the units used in other fields below. They may be\n");
+	put_format(&buf, "%% referenced as needed in TeX templates.\n");
+	put_format(&buf, "%% \n");
+	put_format(&buf, "%% By default, Subsurface exports units of volume as \"ℓ\" and \"cuft\", which do\n");
+	put_format(&buf, "%% not render well in TeX/LaTeX.  The code below substitutes \"L\" and \"ft$^{3}$\",\n");
+	put_format(&buf, "%% respectively.  If you wish to display the original values, you may edit this\n");
+	put_format(&buf, "%% list and all calls to those units will be updated in your document.\n");
+
+	put_format(&buf, "\\def\\%sdepthunit{\\%sunit%s}", ssrf, ssrf, units->length == units::METERS ? "meter" : "ft");
+	put_format(&buf, "\\def\\%sweightunit{\\%sunit%s}",ssrf, ssrf, units->weight == units::KG ? "kg" : "lb");
+	put_format(&buf, "\\def\\%spressureunit{\\%sunit%s}", ssrf, ssrf, units->pressure == units::BAR ? "bar" : "psi");
+	put_format(&buf, "\\def\\%stemperatureunit{\\%sunit%s}", ssrf, ssrf, units->temperature == units::CELSIUS ? "centigrade" : "fahrenheit");
+	put_format(&buf, "\\def\\%svolumeunit{\\%sunit%s}", ssrf, ssrf, units->volume == units::LITER ? "liter" : "cuft");
+	put_format(&buf, "\\def\\%sverticalspeedunit{\\%sunit%s}", ssrf, ssrf, units->length == units::METERS ? "meterpermin" : "ftpermin");
+
+	put_format(&buf, "\n%%%%%%%%%% Begin Dive Data: %%%%%%%%%%\n");
+
+	for_each_dive (i, dive) {
+		if (selected_only && !dive->selected)
+			continue;
+
+		saveProfile(dive, texdir.filePath(QString("profile%1.png").arg(dive->number)));
+		struct tm tm;
+		utc_mkdate(dive->when, &tm);
+
+		dive_site *site = dive->dive_site;
+		QRegExp ct("countrytag: (\\w+)");
+		QString country;
+		if (site && ct.indexIn(site->notes) >= 0)
+			country = ct.cap(1);
+		else
+			country = "";
+
+		pressure_t delta_p = {.mbar = 0};
+
+		QString star = "*";
+		QString viz = star.repeated(dive->visibility);
+		QString rating = star.repeated(dive->rating);
+
+		int i;
+		int qty_cyl;
+		int qty_weight;
+		double total_weight;
+
+		if (need_pagebreak) {
+			if (plain)
+				put_format(&buf, "\\vfill\\eject\n");
+
+			else
+				put_format(&buf, "\\newpage\n");
+		}
+		need_pagebreak = true;
+		put_format(&buf, "\n%% Time, Date, and location:\n");
+		put_format(&buf, "\\def\\%sdate{%04u-%02u-%02u}\n", ssrf,
+		      tm.tm_year, tm.tm_mon+1, tm.tm_mday);
+		put_format(&buf, "\\def\\%shour{%02u}\n", ssrf, tm.tm_hour);
+		put_format(&buf, "\\def\\%sminute{%02u}\n", ssrf, tm.tm_min);
+		put_format(&buf, "\\def\\%snumber{%d}\n", ssrf, dive->number);
+		put_format(&buf, "\\def\\%splace{%s}\n", ssrf, site ? site->name : "");
+		put_format(&buf, "\\def\\%sspot{}\n", ssrf);
+		put_format(&buf, "\\def\\%ssitename{%s}\n", ssrf, site ? site->name : "");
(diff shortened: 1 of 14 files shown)


janiversen:
Sorry Xcode make a mess (again), when I get back to my other machine, I will upload the correct Patch1 again.

janiversen:
Dropped the team work in Xcode, and use a simple rsync, that should prevent problems in the future

dirkhh:
I am feeling a bit ambivalent about this one.
Who would use this and why? And how (given how hard both iOS and Android are working to avoid making user created files in one app accessible in another app)?
On Android we will soon no longer be able to write XML files to a directory that is accessible to the user, period. Unless the device is rooted, I believe that this will not be possible with Android 11.
I'm less familiar with iOS, but typically Apple and Google are on very similar trajectories here.

Which gets me back to the underlying question. Someone who really wants that export should simply sync their data with a Mac and do the export there. I'll bet dollars for donuts that this is overall a LOT easier for the user - and it already works.

So what am I missing? Can you explain the use case (and the target audience) a bit more?

Thanks

janiversen:
i have little experience with Android (or more correctly very old experience), but it seems very counterproductive if they close the possibility to share files. In iOS and the new iPadOS things are very different.

The use case is simple, the iPad is more and more been seen as a computer, so many people do not have a Mac. I upload my dives to DiveLogs.de (something I am working on backporting to the official subsurface-mobile) and most of my friends here do the same. I use the subsurface export file, to import in another program.

Right now I use it nearly daily to control my version of DivePlanner.

How about the following solution:
- I move the menu entry to "Developer", making it less visible
- I prioritise backporting the Web service (which surely must to supported by Android)
- We would also make it a iOS only function, but that seems odd to me

dirkhh:
So here's where Android is going:
https://developer.android.com/training/data-storage#scoped-storage
Apps can only write to app specific directories which on un-rooted devices other apps cannot read.
Apps can only write certain file types (like .jpg) to directories that are readable by other apps.
And all this comes with a complete break of the existing APIs and new APIs that of course don't work on older devices. Which means maintaining two completely code paths for a LONG time; we still have active users on Android 4.4 (which current versions of the app no longer supports, but I see them accessing cloud storage, so they are USING the app). Android 5 was introduced more than 5 years ago. So we will have apps not supporting scoped storage for at least another 5 years...

I don't think this belongs into "Developer". This is a feature. I haven't looked at your code (my apologies -- too much on my plate, I'll try to get to it today), but I'd rather have this as something that's in the main menu under Dive management. That's the logical place for it.
Your argument about iPad makes sense - I know a number of people who do that.
Let's get this to work well on an iPad and then look at what is possible on an iPhone (are there issues with this on an iPhone?) and on Android.

I agree that the integration into another web service (DiveLogs.de in addition to our cloud storage) is a very nice thing to add to the mobile app.

janiversen:
I have not had issues with this on the following devices:
iPad Pro (1st generation), iPadOs 13.2
iPhone 6s, iOS 12.4 (not rooted)

It has also been tested (but in a subsurface-mobile version, with a lot more patches:
iPhone 11
iPad Air
iPhone 8

I can install the official subsurface-mobile with this patch on my iPad again, just to check again.
I am having an issue, installing a patches subsurface-mobile alongside the official one, it does not work (which is why my personal version carries a slightly different name).

Should I do a check "#ifdef OS_IOS" and the equivalent in QML ?

dirkhh:
QML doesn't have the ability to do #ifdef OS :-(

I will try to build this on Android in the next day or three and see how well this can be used prior to Android enforcing scoped storage.
Please bear with me as I try to take care of many different issues around Subsurface

mobile-widgets/qmlmanager.cpp
```diff
@@ -2125,3 +2128,70 @@ void QMLManager::appInitialized()
...
+	// TBD
+}
+
+void QMLManager::exportToWEB(export_types type, QString userId, QString password, bool anonymize)
+{
+	switch (type)
+	{
+		case EX_DIVELOGS_DE:
+//			DivelogsDeWebServices::instance()->prepareDivesForUpload(ui->exportSelected->isChecked());
+		case EX_DIVESHARE:
```
dirkhh, line 2197: just reading this code is an invitation to forget adding the `break;` at the end of each case statement.
Please don't do that.

core/exportfuncs.cpp
```diff
@@ -0,0 +1,311 @@
+// SPDX-License-Identifier: GPL-2.0
+#include <QFileInfo>
```
dirkhh, line 2: From a reviewer's perspective it is rather challenging to have one commit apparently add new export code that I need to go through line by line to understand, and then have almost the same code removed elsewhere two commits later.

What you should do is have one commit `move export functions into separate compilation unit` and allow the reviewer to easily see that this is the same code that just moved into a different file.

packaging/ios/Subsurface-mobile.pro
```diff
@@ -28,6 +28,7 @@ SOURCES += ../../subsurface-mobile-main.cpp \
 	../../core/divesitehelpers.cpp \
 	../../core/errorhelper.c \
 	../../core/exif.cpp \
+	../../core/exportfuncs.cpp \
```
dirkhh, line 31: this also should be part of the one commit that moves the code around

mobile-widgets/qml/Export.qml
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: GPL-2.0
+import QtQuick 2.12
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
+import org.subsurfacedivelog.mobile 1.0
+import org.kde.kirigami 2.4 as Kirigami
+
+Kirigami.ScrollablePage {
+	title: qsTr("Export Divelog information")
```
dirkhh, line 11: while adding a dummy page might make sense in your logical step by step process to develop this code, when I read through the PR to review things this feels odd. Why don't you just have the final Export.qml page right here - or at least the skeleton with all the UI elements (I haven't read all the way to the end, yet, just in case there are interfaces that the final page will use that aren't implemented yet - you could comment them out at this stage).
But adding a dummy page that is filled out a few commits later seems... odd - at least when adding a new feature like this.

mobile-widgets/qml/Export.qml
```diff
@@ -12,6 +12,20 @@ Kirigami.ScrollablePage {
 
 	property int selectedExport: ExportType.EX_DIVE_XML
 
+	FileDialog {
```
dirkhh, line 15: this I can see as its own commit (just continuing on the comments about not starting out with a "dummy" page

mobile-widgets/qml/Export.qml
```diff
@@ -26,6 +26,50 @@ Kirigami.ScrollablePage {
 		}
 	}
 
+	Dialog {
```
dirkhh, line 29: same here - this as its own commit is reasonable.

subsurface-helper.cpp
```diff
@@ -111,6 +111,8 @@ void run_ui()
 	ctxt->setContextProperty("connectionListModel", &connectionListModel);
 	ctxt->setContextProperty("logModel", MessageHandlerModel::self());
 
+	qmlRegisterUncreatableType<QMLManager>("org.subsurfacedivelog.mobile",1,0,"ExportType","Enum is not a type");
```
dirkhh, line 114: the order of commits means that the previous few commits won't actually work.
Why isn't this part of the commit that adds that enum?

subsurface-helper.cpp
```diff
@@ -120,6 +122,7 @@ void run_ui()
 		engine.load(QUrl(QStringLiteral("qrc:///qml/main.qml")));
 	}
 #else
+
```
dirkhh: why are you adding a blank line?

packaging/ios/Subsurface-mobile.pro
```diff
@@ -81,6 +81,7 @@ SOURCES += ../../subsurface-mobile-main.cpp \
 	../../core/btdiscovery.cpp \
 	../../core/connectionlistmodel.cpp \
 	../../core/qt-ble.cpp \
+	../../core/save-profiledata.c \
```
dirkhh, line 84: that should be done earlier in the series

dirkhh (changes requested):
did a code review, haven't played with this, yet. That's on my todo list.
I'm surprised that you are adding all different flavors of export here, but I guess given the goal to make Subsurface-mobile on an iPad a complete Subsurface replacement I can see where you are coming from. As usual, I worry about the user experience on smaller screens...

Most of my comments are about the way this pull request is structured. I prefer things to be structured differently which makes it easier to follow as a reviewer, but also easier to have every commit in itself create a build that works. `git bisect` is a tool that I use a lot and whenever we have intermediate steps that don't compile or don't start, that adds additional work for other developers later.

janiversen:
Before I address your comments, I have a couple of general comments, to explain my commits:
1) I test every single commit, by building desktop, mobileExecutable and iOS, exactly to make sure that "git bisect" will work. In the rare case it is not buildable, it will be noted in the commit text.
2) I try to keep my commit small and inside 1 directory, to make it easier to review and test.
3) I have a preference (whenever possible) when moving functionality, to copy the functionality (add in the new place), test that I did not break anything, then remove the old functionality and test again. I would have thought it makes review easier since the second commit is a "removal only".

Having said that, I can see you have a lot of good comments, which I will address one by one.

mobile-widgets/qmlmanager.cpp
```diff
@@ -2125,3 +2128,70 @@ void QMLManager::appInitialized()
 	checkPendingIntents();
 #endif
 }
+
+void QMLManager::exportToFile(export_types type, QString dir, bool anonymize)
+{
+	// add standard filename
+	QString fileName = dir.right(dir.size() - 7) + "/Subsurface_export";
```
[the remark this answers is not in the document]
janiversen: Added a comment (reason is "file://" which subsurface_open does not accept.

mobile-widgets/qmlmanager.cpp
```diff
@@ -2125,3 +2128,70 @@ void QMLManager::appInitialized()
...
+	switch (type)
+	{
+		case EX_DIVELOGS_DE:
+//			DivelogsDeWebServices::instance()->prepareDivesForUpload(ui->exportSelected->isChecked());
+		case EX_DIVESHARE:
+//			DiveShareExportDialog::instance()->prepareDivesForUpload(ui->exportSelected->isChecked());
+		default:
+			break;
+	}
+	qDebug() << "export to WEB type " << type << " using " << userId << "/" <<  password << " remove names " << anonymize;
```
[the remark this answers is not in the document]
janiversen: Changed

mobile-widgets/qmlmanager.cpp
janiversen: corrected

core/exportfuncs.cpp
janiversen: I have updated the commit text, to state it is a 1-1 copy from divelogexportdialog. But I really believe (also from a reviewers perspective) to not make commit across directories, and in this case split the adding/removing (to keep the removing as simple as possible). If you do not agree I will squash the commits.

packaging/ios/Subsurface-mobile.pro
janiversen: This has really nothing to do with moving the code around, it is about making the new code available to qmlmanager, so you could argue it should be part of the qmlmanager commit.

I can see however that this PR makes changes 2 times to the file, and that is not needed so corrected.

mobile-widgets/qml/Export.qml
janiversen: no more "dummy" page.

mobile-widgets/qml/Export.qml
janiversen: ok

core/exportfuncs.cpp
dirkhh: That's interesting. I see "not make commit across directories" as fairly unimportant. To me it matters more to have both parts of the move in the same commit so I can see how this affects things.

But I do of course see your point that this way you can introduce the new code, then switch to using it, then remove the old code, which also has a lot of advantages.

No need to squash the commits, what you did isn't wrong.

subsurface-helper.cpp
janiversen: What does not work ?
The commits are in this order:
1) add enum to qmlmanager.h (222ab86cc mobile-widgets: add export functionality to qmlmanager)
2) make enum available to qml (a66adbbb5 startup mobile: make ExportType enum available to QML)
3) use it in QML (bd091859c mobile-widgets/qml: add export page)

It is not part of the commit that adds the enum, because having qmlmanager.h and subsurface-helper.cpp in the same commit seems wrong to me (from a bisect pow).

subsurface-helper.cpp
janiversen: corrected

packaging/ios/Subsurface-mobile.pro
janiversen: yes

janiversen:
@dirkhh I hope to have addressed all your concerns, if not (or if you have new ones) please let me know.

subsurface-helper.cpp
dirkhh: I must have confused myself reading the code. Sorry

dirkhh:
I've settled down with a glass of wine...
I'll take another look at the code tomorrow and try it on an iPad as well as on Android

It's Thanksgiving week in the US which means a bit less work and a LOT more family obligations

janiversen:
In Spain it is "black-friday week", so it is impossible to do shopping. 

Enjoy your wine, and please do not stress with this PR, I am in no rush (I am busy changing 2 upcoming PR´s based on your comments).

dirkhh:
This fails to build for me when merged on top of master - there seems to be a disconnect regarding the kirigami version in use. I need to track this down some more...

dirkhh:
Finally got a chance to play with this. Sorry this took so long.
The user experience seems a bit weird as the export page isn't removed after you exported something. When running mobile on desktop the file picker picks a folder to place the export file into, but the user has to guess the (admittedly obvious) file name.
Neither of those seems like a show stopper (and the second one may still be what we want to do in order to simplify the UI), so I'm inclined to merge this so it doesn't sit here for too long.

dirkhh (approved):

PR #2397, opened 2019-11-23
Correct CMake to no do "cp -a" for missing QML parts.

janiversen:
The "cp -a" gives a "cannot overwrite" error because macdeployqt
is doing the job.

only copy if Qt version is between 5.11.0 and 5.12.5

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

This patch have been tested with Qt 5.12.5 and Qt 5.13.2

### Changes made:

see commit

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -440,7 +440,12 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
 	install(CODE "execute_process(COMMAND cp -a ${_qt5Core_install_prefix}/qml/QtPositioning ${CMAKE_BINARY_DIR}/${APP_BUNDLE_DIR}/Contents/Resources/qml)")
 	if(NOT Qt5Core_VERSION VERSION_LESS 5.11.0)
 		# and with Qt 5.11 we need another library that isn't copied by macdeployqt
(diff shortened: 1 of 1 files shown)


dirkhh (commented):
I simply need to test this on the platforms that I usually build.
Usually the trade-off "annoying error message that doesn't hurt" vs "I need to make sure this still works for all the different ways I build this" makes me lean towards ignoring annoying errors...

janiversen:
Your POW is very understandable, feel free to close this PR if you feel it does not help the project.

janiversen:
On a second thought lets all save some work, since this do not have any influence on the work I have made.

dirkhh:
As you know, I always want to respect the work that contributors do.
And of course what you were trying to do was correct in the sense that it fixed a bug.
The real challenge is the complexity of all the different platforms and the various versions against which I end up building - so in the end the issue is that this project still relies way too much on ME doing so many different things. And every time we manage to automate another piece (thanks to @glance- for the work on liberating the building of containers from being dependent on me), overall things get easier to change.
So hopefully it won't be long before PRs like this will feel far less daunting to me.
Thanks for your understanding

PR #2398, opened 2019-11-23
Allow fast changes (easy) changes to qml.

janiversen:
UPDATE, made it even simpler to use.

- [ ] Bug fix
- [ ] Functional change
- [ x] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

This patch allow subsurface-mobile for desktop (mobileExecutable) to load qml as files
instead of as resource (which are in the executable).

Using this feature, allows to compile subsurface-mobile once, and then simply restart subsurface-mobile to see the effect. This reduces turnaround times for qml changes by a big factor.

Use case:

--> run & test loop
1) ./build-mobile/Subsurface-mobile.app/Contents/MacOS/Subsurface-mobile
--testqml=.../subsurface/mobile-widgets/qml
2) <make modifications to qml in qml directory
3) restart subsurface-mobile (continue with 1)

Subsurface and Subsurface-mobile (used without switch) remain identical to earlier.

### Documentation change:

Not sure where this should be documented.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,6 +179,10 @@ elseif (SUBSURFACE_TARGET_EXECUTABLE MATCHES "MobileExecutable")
 	endif()
 	list(APPEND QT_EXTRA_COMPONENTS QuickControls2)
 	add_definitions(-DSUBSURFACE_MOBILE)
+
+	# add definition to seperate mobile for devices and for desktop
+	add_definitions(-DSUBSURFACE_MOBILE_DESKTOP)
+
 	message(STATUS "Building Subsurface-mobile requires BT support")
 	set(BTSUPPORT ON)
 endif()
--- a/core/subsurfacestartup.c
+++ b/core/subsurfacestartup.c
@@ -106,6 +106,9 @@ struct preferences default_prefs = {
 };
 
 int run_survey;
+#ifdef SUBSURFACE_MOBILE_DESKTOP
+char *testqml = NULL;
+#endif
 
 const struct units *get_units()
 {
@@ -183,6 +186,9 @@ static void print_help()
 	printf("\n --version             Prints current version");
 	printf("\n --survey              Offer to submit a user survey");
 	printf("\n --user=<test>         Choose configuration space for user <test>");
+#ifdef SUBSURFACE_MOBILE_DESKTOP
+	printf("\n --testqml=<dir>       Use QML files from <dir> instead of QML resources");
+#endif
 	printf("\n --cloud-timeout=<nr>  Set timeout for cloud connection (0 < timeout < 60)\n\n");
 }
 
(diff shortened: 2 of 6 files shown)


dirkhh:
I don't understand why many of your commit messages start with `/:`
I didn't notice this last time, but I find it odd when I look through our git history.

CMakeLists.txt
```diff
@@ -179,6 +179,10 @@ elseif (SUBSURFACE_TARGET_EXECUTABLE MATCHES "MobileExecutable")
 	endif()
 	list(APPEND QT_EXTRA_COMPONENTS QuickControls2)
 	add_definitions(-DSUBSURFACE_MOBILE)
+
+	# add definition to seperate mobile for devices and for desktop
+	add_definitions(-DSUBSURFACE_MOBILE_DESKTOP)
+
```
dirkhh, line 185: Commenting on the commit messages for "/: add CMake definition SUBSURFACE_MOBILE_DESKTOP"...
I think this should be "Build system: add SUBSURFACE_MOBILE_DESKTOP as build type"

core/subsurfacestartup.c
```diff
@@ -241,6 +247,13 @@ void parse_argument(const char *arg)
 				++force_root;
 				return;
 			}
+#ifdef SUBSURFACE_MOBILE_DESKTOP
+			if (strncmp(arg, "--testqml=", sizeof("--testqml=") - 1) == 0) {
+				testqml = malloc(strlen(arg) - sizeof("--testqml=") +1);
```
dirkhh: white space: `"--testqml=") + 1);`

mobile-widgets/qml/qmldir
```diff
@@ -0,0 +1,3 @@
+MapWidget ../../map-widget/qml/MapWidget.qml
+MapWidgetContextMenu ../../map-widget/qml/MapWidgetContextMenu.qml
+MapWidgetError ../../map-widget/qml/MapWidgetError.qml
```
dirkhh, line 11: just for my understanding - this used to work because we bundled things as resources... why is this needed now?
My guess is so that if you point to the "in tree" directory things automagically work?

dirkhh (changes requested):
Overall this looks like a good change. I added a few comments, especially on the commit messages.

CMakeLists.txt
janiversen: Seems I misunderstood the guideline to use e.g. core: or mobile-widgets: when you make changes in that directory. Will update the commit messages.

mobile-widgets/qml/qmldir
janiversen: A good question, in order to let others know as well I have added a comment in the file.

In short QQmlApplicationEngine allows imports to be from different directories, but Components (other qml files that you reference like e.g. SsrfButton needs to be in the same directory). The qrc file actually make the same aliases.

janiversen:
> I don't understand why many of your commit messages start with `/:`
> I didn't notice this last time, but I find it odd when I look through our git history.

Hmm "Resolve conversation" seems to delete my response, not what I expected.

I understood the guideline to use e.g. core: mobile-widget: when the commit was centered around that part of the code, hence /: when working at top level. I have updated the commit messages.

mobile-widgets/qml/qmldir
dirkhh: excellent. thanks.

dirkhh (approved):
This looks great. Thanks for addressing my comments!

PR #2401, opened 2019-11-24
[WIP] Containers in GitHub actions (replaces #2395)

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a first attempt to complete the work @glance- started in #2395 
It's been heavily tested and painfully slowly pieced together in a stub repo, just to get the mechanics right. And I have every expectation that this will fail in many creative ways, just based on the number of attempts it took me to get the hacked-together version working

### Changes made:

1) use docker hub and create a new subsurface account there
2) store the credentials for that account as secrets in this account
3) split the MXE build into two stages with the completion of the first one triggering the second one through a repository dispatch event
4) deal with the yaml inability to do even the most basic string manipulations (or my inability to figure out how to do this more easily)
5) finally use those new images in our actions

### Related issues:

#2395 

### Additional information:

I expect the CICD run to fail because the new images aren't there, yet. And as I said above, I'm sure there are more bugs in here. So most likely this will take a few restarts and possibly a few fixes before it's ready. Hence "WIP"

### Release note:

doesn't really affect users

### Documentation change:

doesn't really affect users

### Mentions:

@glance- deserves much of the credit... and I would appreciate your keen eye on this to catch all the mistakes I snuck in

Part of the complication is that for security reasons I don't want this to run on pull requests, only on pushes to branches in this repo... so testing might be a bit awkward :-)


dirkhh:
Ahh, this is harder to test than I thought.
So the repository dispatch event only triggers if the "receiving" workflow is present on the `master` branch, it seems. So in order to test if this would work, we'd have to commit it to master...
That's not a great user experience and could cause a lot of noise on the master branch...

glance-:
Awsome work.

I'd suggest you inc the container version numbers and then can safely push everything up to 301a7744c9e81e9fb360e951c9aa6fc3399d88d7 to master, without risking anything.

Then we can test the last bits in separate

dirkhh:
> I'd suggest you inc the container version numbers and then can safely push everything up to [301a774](https://github.com/Subsurface-divelog/subsurface/commit/301a7744c9e81e9fb360e951c9aa6fc3399d88d7) to master, without risking anything.
> 
> Then we can test the last bits in separate

Yes, I came to a similar conclusion. But I realized that in order to be able to test changes to the stage 2 Docker file I needed to deliver the branch name as part of the payload... otherwise when pushing to a different branch, we'd still just get the stage 2 as implemented in master built. Admittedly, this may not be a huge deal in real life, but I'd like to get this part right as well.
This means figuring out how to either use the payload, or how to extract the name from the event name (where we right now only encode the version we are building).
The other thing that occurred to me in this context is that it might make sense to encode the branch name in the docker image name as well - but then that causes issues in our use of those new images (i.e. how we can have the right name to use when testing a pull request vs. testing master.
I'll need to do some more thinking to make sure this is all reasonably well designed :-)

dirkhh:
OK, new version that should do the right thing in a branch. And of course will still fail because the corresponding YAML isn't in master, yet. But at least once it is merged it will be slightly more flexible.

Again, @glance- I'd love to know if you can spot anything that looks fishy.
I spent quite a bit of time on this (because each iteration takes a long time - and of course in my tests I didn't actually build the full docker images). I hope the value we gain from it makes up for the time spent :-)

As discussed, this is dropping the commit that switches to the new images so we can do this in a follow-on PR once this was merged.

dirkhh:
I was wrong. This didn't fail, it just sends a repository dispatch event that we are ignoring, so the MXE container doesn't get built until this gets merged into master.
I'll give others more time to comment :-)

glance-:
We should make the "pushing" conditional on being on master, or encode the branch name in the version name if the branch not master.

You could push the images to github packages to, for future experimentation.

dirkhh:
> We should make the "pushing" conditional on being on master, or encode the branch name in the version name if the branch not master.

The reason I am not making it conditional on master is that otherwise the CI pipeline never runs until AFTER a merge into master. Because of security concerns we still restrict things to "push" (which means people who are trusted as collaborators) vs. "pull_request" (which is anyone).
I thought about adding the branch name to the image name. I can still do that, no problem.

> You could push the images to github packages to, for future experimentation.

True - but since we aren't using them, it would feel like intentionally wasting GitHub's resources at this point. Once we have a plan what to do with them (or as soon as GitHub fixes the issues with using their own registry in their Actions) I'll be happy to do that.

glance- (approved):

dirkhh:
Because things haven't be weird enough here... 

Ubuntu 14.04 is of course long EOL and extended security maintenance (ESM) is available for a fee (and free for individuals). Which requires authentication. The credentials for which we of course can't add to an open source repo. And since they fixed a glib2 bug `apt-get update` now points to a package that's only in ESM and because of that the Trusty setup fails.
I'm not sure that I agree with having a container that isn't set up for ESM fail a simple `apt-get install` if no ESM credentials are available. But as far as I can see that's where we are right now.

None of my ideas how to work around this are straight forward or easy to implement. Am I missing something?

henrik242:
> Ubuntu 14.04 is of course long EOL

Controversial, I know, but how about just switching to the latest LTS?  The few users that are stuck on older systems can still use their old Subsurface installations...

dirkhh:
It's been accepted as a Bug by Canonical and should be addressed shortly. There's even a workaround available (but since I'm in the middle of my work day, I haven't looked into that, yet)

https://bugs.launchpad.net/ubuntu/+source/ubuntu-advantage-tools/+bug/1854202

dirkhh:
OK, this seems to have worked. So let's merge it and then add using those new builds.

PR #2404, opened 2019-11-26
qPref constructor private (singleton) and all access functions static.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Making all get/set functions static are slightly faster (since the functions are addressed without a vptr) but more importantly making the constructor private secures the object is a singleton.

This patch obsoletes #1672, and secures that both C++ and QML works efficiently.

This patch do not change functionality as such.

It also contains a single commit to update .gitignore with some untracked files, this strictly seen does not belong to this PR.

### Changes made:

see commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ Documentation/docbook-xsl.css
 Documentation/user-manual*.html
 Documentation/user-manual*.pdf
 Documentation/user-manual*.text
+Documentation/mobile-images/mobile-images
 packaging/windows/subsurface.nsi
 packaging/macos/Info.plist
 *.kdev4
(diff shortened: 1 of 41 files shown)


dirkhh:
I had so hoped to never see a PR about qPref anything ever again in my life... oh well, not all hopes can work out.

I'll add going through this to my todo list. Not today.

dirkhh (approved):
First, I'm not in love with the qPref code that we have.
And I really would rather leave it alone unless something is broken and needs fixing.
Yes, one day we might all be bored and replace this with something more elegant. But today is not that day.

With that out of the way - this series seems reasonable.

I would have preferred commits 2 & 3 to be a separate cleanup PR (they would have been merged within the hour I bet). And I might have preferred more verbose commit messages; having a bit of the "why" and "so what" of a change in the commit messages is always useful. Also look for line length in the commit subject lines...

But that's all poking on minor marginal things - there's nothing wrong with this and I see why it's an improvement, so I'll merge it.

PR #2405, opened 2019-11-26
mac: create Xcode project for mobile on a desktop, and desktop versions

janiversen:
This is a convenience PR for those debugging on a Mac.

Create Xcode project, allowing easy editing of all sources.
the project uses "make mobile" and "make desktop" to generate the project,
so all CMake changes are automatically updated

the project uses folder references, so adding/removing a source is
automatically updated. The project only needs to be updated when a new folder is created
directly in the subsurface directory.

The project expect build-mobile and build to be existing as well
as external libraries, so after cloning do a normal build before
using the project.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ x] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/packaging/macosx/Subsurface-mobile.xcodeproj/.gitignore
+++ b/packaging/macosx/Subsurface-mobile.xcodeproj/.gitignore
@@ -1,0 +1,2 @@
+project.xcworkspace
+xcuserdata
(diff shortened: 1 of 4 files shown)


dirkhh:
so this isn't something I'm familiar with at all. Is there a reason why this adds a new top level directory?
In other words, is this something that could live under `packaging/macosx` ?

PR #2406, opened 2019-11-26
Cylinders: Fix crash when removing cylinders

bstoeger:
Change the remove() function of the cylinder and weight models
to take the index by value. The code used to take it by reference
and the reference would be invalidated when removing rows from
the model!

Reported-by: Gaetan Bisson <bisson@archlinux.org>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I'm currently bedridden owing to a nasty cold. Therefore, no comments besides commit message.

### Changes made:

Pass model indexes by value instead of reference. References are evil.
### Mentions:

@vesath

### DIFF
--- a/qt-models/cylindermodel.cpp
+++ b/qt-models/cylindermodel.cpp
@@ -486,7 +486,7 @@ Qt::ItemFlags CylindersModel::flags(const QModelIndex &index) const
 	return QAbstractItemModel::flags(index) | Qt::ItemIsEditable;
 }
 
-void CylindersModel::remove(const QModelIndex &index)
+void CylindersModel::remove(QModelIndex index)
 {
 
 	if (index.column() == USE) {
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
Sorry to hear you are sick - that sucks.
The changes look good, I'll wait until we hear back from @vesath if this fixes his problem (which I guess it will)

vesath:
Wow, that was fast! I confirm that the proposed patch indeed fixes my problem. Thanks Berthold, get better soon.

bstoeger:
Thanks, it's already going much better.

With things like this, I always wonder if it would be worth to file a defect-report. It's a *very* surprising behavior that we get passed a const-reference to an index, which is then invalidated under our feet! Even though we mustn't access data via the index once we deleted the row, invalidating it appears rather unfriendly. However, constructing a minimal example showing the behavior would be quite some work.

PR #2408, opened 2019-11-27
[RFC] Desktop: Improve speed of selecting multiple (or all) dives

bstoeger:
When selecting all dives via CTRL-A or manually and the trips
were not expanded, the QSelectionModel sends a single
selectionChanged signal per trip. We are reloading the map
in every call, making this very slow.

I couldn't figure out how to make QSelectionModel behave more
nicely, therefore I chose the nuclear option: Remove the map
reloading from selectionChanged() and hook into all functions
that do selection changes. In these functions, first call the
original code and then do the selection-changed operations.

This will certainly need some tuning.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Willem reported slow selection of the whole divelog. The main cause was way too many map-reloads. I didn't find out how to solve this properly, therefore I propose this very dirty hack, which hooks into Qt's innards directly. Ideally a Qt-export would look into this, but I reckon that this is unrealistic.

This needs testing, as it will certainly break things.

Note: This doesn't fix the fundamental madness of the selection code.
### Changes made:

1) Hook into selection changing functions.
2) Move map-reloading into kidnapped functions.
3) Remove dontEmitDiveChangedSignal, since the this is now solved differently, i.e. the code at the end of selection changes is called explicitly.

### Release note:

done.

### Documentation change:

no.
### Mentions:

@willemferguson: You complained about the selection problem, therefore I would appreciate your testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 Mobile: (desktop only) new switch --testqml, allows to use qml files instead of resources.
+Desktop: increase speed of multi-trip selection
 Mobile: ensure that all BT/BLE flavors of the OSTC are recognized as dive computers [#2358]
 Desktop: allow copy&pasting of multiple cylinders [#2386]
 Desktop: don't output random SAC values for cylinders without data [#2376]
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -31,7 +31,7 @@
 #include "desktop-widgets/mapwidget.h"
 
 DiveListView::DiveListView(QWidget *parent) : QTreeView(parent), mouseClickSelection(false),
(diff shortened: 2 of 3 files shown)


dirkhh (commented):
That looks harmless. Um. Not. Another thing to my growing list of things that I promise to test and have no idea when I'll get to it.
It's Thanksgiving. And my family expects me to NOT spend the next two days staring at my computer.
Starting right about now...

bstoeger:
> That looks harmless. Um. Not.

Like the rest of the selection code. In my quick testing it is a massive improvement, though.

Enjoy your holidays - this can wait. Let's see what @willemferguson says first.

dirkhh:
Yes, indeed. HUGE improvement... I like it!

PR #2412, opened 2019-11-29
Add Export divesites to mobile.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Found a way to bypass DiveFilter (which are not activated for mobile), using a C macro
for_each_dive_site

This PR extends the export functionality in #2396, with dive sites export.

### Changes made:

see commits

### Related issues:

sits on top of #2396, and need to be rebased when (if) #2396 is accepted.

### Documentation change:

Mobile doc. should be extended with the Export facility, including dive site export.

### DIFF
--- a/core/exportfuncs.cpp
+++ b/core/exportfuncs.cpp
@@ -8,6 +8,7 @@
 #include "core/file.h"
 #include "core/errorhelper.h"
 #include "core/divefilter.h"
+#include "core/divesite.h"
 #include "exportfuncs.h"
 
 
@@ -300,6 +301,12 @@ std::vector<const dive_site *> exportFuncs::getDiveSitesToExport(bool selectedOn
 			continue;
 		res.push_back(ds);
 	}
+#else
+	/* walk the dive site list */
+	int i;
+	const struct dive_site *ds;
+	for_each_dive_site (i, ds, &dive_site_table)
+		res.push_back(get_dive_site(i, &dive_site_table));
 #endif
 	return res;
 }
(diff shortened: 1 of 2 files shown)


dirkhh:
#2396 was merged.
But I don't like the filter hack. I would prefer to make sure we do this right in the context of the larger filter rework @bstoeger and I have been working on for a few weeks (but that I got completely sidetracked from for the last two weeks... so I need to reconnect on the status of that)

janiversen:
Happy to use diveFilrer when available, that is the right thing to do.

janiversen:
I had a second look, and I believe it is less a "hack" as you think, and I cannot see what I can do in this PR to change it 

The loop for_each_dive_site, is actively used several places in the code:
Please look at core/divesite.c (that is where I copied it from) line 21, 33, 44, 56, 70, 101, 311, 334

Apart from that in the code merged there are a number of for_each_dive, which surely also one day can be replaced by the new filter function.

The real "hack" is in core/divefilter.cpp line 5++, that is where it is accepted that mobile uses a different version (in this case in the divest) than the desktop.

Please advice which changes you expect me to make. Alternatively since there are many places still open to change, would it not be sensible to just count this as one more ?

dirkhh (approved):
It works. I'll merge it and when we redo filtering, we just need to make sure we don't break this one.

PR #2414, opened 2019-11-30
Undo: make profile-editing undoable

bstoeger:
Recently, undo of dive-replanning was introduced. Therefore,
it appears logical to do the same thing for editing of the
profile of manually added dives.

For now, use the same undo-command, just change the displayed
text from "replan dive" to "edit profile". Move the fixup dive
call into the undo-command.

Eventually, every action on the profile should be made undoable.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This implements a small oversight of making replanning undoable: profile editing is virtually the same and should likewise be undoable. This PR most certainly conflicts with the divelist and weightsystem undo PRs rhat are currently open.

### Release note:

Not necessary, since undo of dive replanning is already mentioned.

### DIFF
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -240,7 +240,12 @@ void pasteDives(const dive *d, dive_components what)
 
 void replanDive(dive *d)
 {
-	execute(new ReplanDive(d));
+	execute(new ReplanDive(d, false));
+}
+
+void editProfile(dive *d)
+{
+	execute(new ReplanDive(d, true));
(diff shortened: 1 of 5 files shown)


dirkhh (approved):
I had to read the changes to `maintab.cpp` twice, but I think this makes a lot of sense.

PR #2426, opened 2019-12-07
mobile: upload to divelogs.de

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR activates "upload to divelogs.de" in the export menu.

All dives will always be uploaded (select in not supported in the current divest).

### Changes made:

see commits.

### Release note:

Added line to CHANGELOG.md


janiversen:
There is a small problem outstanding about closing the Export page, so just checking if build is ok on all platforms.

janiversen:
All done (except +1 from Travis). 

Once (if) accepted, I will submit the dive-share upload PR, and in the meantime I will continue back porting Diveplanner and get it to work in this mobile version (in parallel to the desktop version, which is a nice challenge).

janiversen:
This is new:

/win/mxe/usr/lib/gcc/i686-w64-mingw32.shared/5.5.0/include/c++/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.

Are there no support for C++ 2011 on Windows ? seems very strange!
@dirkhh could this be related to your ongoing build work ?

dirkhh:
> This is new:
> 
> /win/mxe/usr/lib/gcc/i686-w64-mingw32.shared/5.5.0/include/c++/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
> 
> Are there no support for C++ 2011 on Windows ? seems very strange!
> @dirkhh could this be related to your ongoing build work ?

I think we JUST got a working container again a few hours ago - but that one hasn't been verified, yet (I am about to kick off a test build) - so right now I am not confident that the Windows builds are indeed equivalent to what we had a week or so ago. Let me look into that.
I'll also try to review this PR today

mobile-widgets/uploadDiveLogsDE.cpp
```diff
@@ -0,0 +1,296 @@
...
+		return;
+	}
+
+	// And upload it
+	uploadDives(filename, userid, password);
+}
+
+
+bool uploadDiveLogsDE::prepareDives(bool selected, const QString &filename)
+{
```
dirkhh: do this function is remarkably similar to `DivelogsDeWebServices::prepare_dives_for_divelogs()` except that the arguments are reversed...
I haven't gone through the whole series, just starting the review, but I immediately wonder if it wouldn't make more sense to have a shared class in `core/` to take care of these tasks. Is there something fundamentally different that I haven't noticed, yet?

mobile-widgets/uploadDiveLogsDE.cpp
```diff
@@ -0,0 +1,296 @@
...
+		se = zip_error_code_system(error);
+#else
+		zip_error_get(zip, &ze, &se);
+#endif
+		report_error(qPrintable(tr("error writing zip file: %s zip error %d system error %d - %s")),
+					 qPrintable(QDir::toNativeSeparators(filename)), ze, se, zip_strerror(zip));
+		return false;
+	}
+	return true;
+}
```
dirkhh: yeah, having compared the complete function the differences feel very very subtle to me - to the point where I'm not quite understanding WHY they are different.

mobile-widgets/uploadDiveLogsDE.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+
+void uploadDiveLogsDE::uploadDives(const QString &filename, const QString &userid, const QString &password)
+{
+	QHttpPart part1, part2, part3;
+	static QNetworkRequest request;
+	QString args;
+
+
```
dirkhh: double empty line

mobile-widgets/uploadDiveLogsDE.cpp
```diff
@@ -0,0 +1,296 @@
...
+	// connect signals from upload process
+	connect(reply, SIGNAL(finished()), this, SLOT(uploadFinished()));
+	connect(reply, SIGNAL(error(QNetworkReply::NetworkError)), this,
+		SLOT(uploadError(QNetworkReply::NetworkError)));
+	connect(reply, SIGNAL(uploadProgress(qint64, qint64)), this,
+		SLOT(updateProgress(qint64, qint64)));
+	connect(&timeout, SIGNAL(timeout()), this, SLOT(uploadTimeout()));
+
+	timeout.start(30000); // 30s
```
dirkhh: this function also borrows heavily from the `DivelogsDeWebServices` class, only here you combined a rather complex multi function approach into something that is easier to read - I don't know if it's better, need to spend more time with the code, but again the real question is "why would it be worth to have this code twice"?

core/settings/qPrefCloudStorage.cpp
```diff
@@ -77,3 +77,24 @@ HANDLE_PREFERENCE_INT(CloudStorage, "timeout", cloud_timeout);
 HANDLE_PREFERENCE_INT(CloudStorage, "cloud_verification_status", cloud_verification_status);
 
 HANDLE_PREFERENCE_BOOL(CloudStorage, "save_password_local", save_password_local);
+
+QString qPrefCloudStorage::divelogsde_userid()
+{
+	return qPrefPrivate::propValue(keyFromGroupAndName(group, "divelogsde_userid"), "").toString();
```
dirkhh: this confuses me - we already have `divelogde_user` and `divelogde_pass` - why create new keys for this?

core/settings/qPrefCloudStorage.cpp
```diff
@@ -77,3 +77,24 @@ HANDLE_PREFERENCE_INT(CloudStorage, "timeout", cloud_timeout);
...
+void qPrefCloudStorage::set_divelogsde_userid(const QString &value)
+{
+	qPrefPrivate::propSetValue(keyFromGroupAndName(group, "divelogsde_userid"), value, "");
+	emit instance()->divelogsde_useridChanged(value);
+}
+
+
+QString qPrefCloudStorage::divelogsde_password()
+{
+	return qPrefPrivate::propValue(keyFromGroupAndName(group, "divelogsde_password"), "").toString();
```
dirkhh: ditto

mobile-widgets/uploadDiveLogsDE.cpp
```diff
@@ -257,6 +261,8 @@ void uploadDiveLogsDE::updateProgress(qint64 current, qint64 total)
 
 void uploadDiveLogsDE::uploadFinished()
```
dirkhh: the declaration (and the use in the previous hunk) have two arguments, this has no arguments. That seems wrong? Or am I missing some overloading here?

mobile-widgets/uploadDiveLogsDE.cpp
```diff
@@ -301,10 +305,14 @@ void uploadDiveLogsDE::uploadTimeout()
 	QString err(tr("divelogs.de not responding"));
 	report_error(err.toUtf8());
 	emit uploadFinish(false, err);
+	timeout.stop();
```
dirkhh: you add this after every `emit uploadFinish()` - wouldn't it make more sense to have the `timeout.stop()` in that?
Or does that not work for other reasons?

mobile-widgets/uploadDiveLogsDE.cpp
```diff
@@ -257,7 +257,18 @@ void uploadDiveLogsDE::uploadDives(const QString &filename, const QString &useri
...
+	if (total <= 0 || current <= 0)
+		return;
+
+	// Calculate percentage
+	// And signal whoever wants to know
+	qreal percentage = (float)current / (float)total;
+	emit uploadProgress(percentage);
+
+	// reset the timer: 30 seconds after we last got any data
+	timeout.start();
```
dirkhh: that's a **LONG** timeout

mobile-widgets/qml/Export.qml
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0
 import QtQuick 2.12
-import QtQuick.Controls 2.12
+import QtQuick.Controls 2.13
```
dirkhh: why do we need 2.13? It's always nice to have a comment in the commit message that explains why we need a newer version of any dependency

dirkhh (commented):
Lots of comments, some rather fundamental.
I like the idea of offering this alternative way to upload dives for the people who hate our cloud or who simply don't even want to have a laptop/desktop anymore

mobile-widgets/uploadDiveLogsDE.cpp
janiversen: Yup, copied from the desktop version. What would you like as timeout ?

mobile-widgets/qml/Export.qml
janiversen: It might just be me, I am using 2.13 normally, let me test that.

janiversen:
I avoided making a shared version because the desktop version is very entangled in the UI. But I will move the code and make a shared version.

I will however do the needed work and make a shared version. No guarantee that I can share the whole lot, but let me come back on this.

mobile-widgets/uploadDiveLogsDE.cpp
```diff
@@ -0,0 +1,296 @@
...
+		return false;
+	}
+
+	// remove zip file if previous version exist
+	QFile f(filename);
+	if (f.open(QIODevice::ReadOnly)) {
+		f.close();
+		f.remove();
+	}
+
```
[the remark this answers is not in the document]
janiversen: Corrected.

mobile-widgets/uploadDiveLogsDE.cpp
janiversen: The should not be any relevant differences.

mobile-widgets/uploadDiveLogsDE.cpp
janiversen: I cleaned some return parts, that is all. This is ready as a shared function.

mobile-widgets/uploadDiveLogsDE.cpp
janiversen: corrected

mobile-widgets/uploadDiveLogsDE.cpp
janiversen: Reason is that it is a multi version in the desktop is that the second part is triggered by the button, however I might be able to bypass that.

core/settings/qPrefCloudStorage.cpp
janiversen: The desktop version uses QSettings directly, so it has a different behavior than the other settings. But of course the names should be identical, corrected.

dirkhh:
In the meantime I replaced the broken Windows build container with a working one and restarted that build, just to get the bogus red dot removed.

As for code sharing - I think it would make sense to have something that is similar to what you created here in core and have the code from both desktop and mobile call into that - so the part that prepares the zip file and talks to the webservice on the other side.

I think it's perfectly fine to think about this as just the parts that are easy to share. I don't want to push you to a massive rework of existing code - I just don't think this almost complete duplication was the right answer, either.

If you want to focus on mainly the mobile UI, I can take a stab at moving this code into core if that would make your life easier.

core/settings/qPrefCloudStorage.cpp
janiversen: corrected

mobile-widgets/uploadDiveLogsDE.cpp
dirkhh: Ha. That's funny. It shows you just how inconsistent reviews here are. We have a command line option to increase the timeout when accessing the subsurface cloud.
Maybe we could reuse that here? `prefs.cloud_timeout`.
This is set to 5 seconds by default in the desktop app, but 10 seconds in the mobile app, because we felt that you'd be more likely to have bad internet on a device than on your laptop (not always true, though). And then there is the `--cloud-timeout=` command line option to overwrite this.

mobile-widgets/uploadDiveLogsDE.cpp
janiversen: nope, "uploadFinish(a, b)" is a signal, without implementation, and uploadFinished() is a SLOT with implementation. Agree naming could have been more different, but they do the same thing, so I kept them close.

mobile-widgets/uploadDiveLogsDE.cpp
janiversen: uploadFinish() is a signal and to not have an implementation, therefore I need to stop the timeout here.

mobile-widgets/qml/Export.qml
janiversen: Found the problem, the ProgressBar have some problems solved in 2.13, but I have already included the workaround for 2.12

mobile-widgets/uploadDiveLogsDE.cpp
dirkhh: Ahhh, I had the feeling that I was missing something - and the name confusion with the slot made it worse for me as I read the changes. Thanks for clearing this up.

mobile-widgets/qml/Export.qml
dirkhh: I can never tell which version goes with which Qt version... in general, for the mobile app I'm MUCH MORE willing to require a reasonably new Qt version. Being able to use at least the last two (so right now Qt5.12/13) would be nice, but if there is a good reason to require Qt5.13 (is this is what 2.13 maps to) then I'm also ok with this, I just ask that it's documented in the commit WHY we made that choice.
Thanks

janiversen:
Do not worry, I will do the rework, but how about we make a new directory "sharedUI" since a lot of this is not really core.

I have 5-6 classes in the PR queue, that are shared, so this class will not be alone.

dirkhh:
> Do not worry, I will do the rework, but how about we make a new directory "sharedUI" since a lot of this is not really core.
> 
> I have 5-6 classes in the PR queue, that are shared, so this class will not be alone.

if we have indeed shared UI code then I think `shared-ui-classes` would fit a little better with `qt-models`, `desktop-widgets`, `mobile-widgets` ... agreed?

Having said this... this isn't really UI code and should never have gone into a `desktop-widget` in the first place. This is code that talks to a webservice backend (and that prepares the data to send there). I _think_ that's `core` code. But we can certainly come up with a different shared folder if this gets a little too far from the strict meaning of "core". But let's not call it UI unless it's really UI related.

Or am I again missing something?

janiversen:
You are not missing anything, in this case, but it my other classes are, lets call them facilitators for the UI (in the way that core/settings also contain UI close code). 

I will move the code, after I have reworked it, based on your comments, and once moved I will integrate it into the existing desktop-widget code.

Instead of reworking, compiling and testing all commits, I will do the move on top of the series, I hope that is something you can accept.

dirkhh:
> Instead of reworking, compiling and testing all commits, I will do the move on top of the series, I hope that is something you can accept.

You mean on top of this series? Sure, that works for getting the code ready. I may then end up combining a few commits as I merge things to make the whole thing easier to understand - or not, depending on how it looks :-)
We'll figure it out.

janiversen:
This should do away with the review comments, a fast test shows it still works.

Now to the "funny" part integration, that will cause code changes, but that should be seen as 2 separate issues.

janiversen:
@dirkhh prepareDives integrated, quite a tough job to get the function identical, and document all differences.

I will continue with a couple of other functions, but that will be a lot harder to document differences (think of the complex division of the actual upload, which I have combined into a more simple one).

janiversen:
@dirkhh done (but it was no small task, you asked for).

Now it is down to review and testing.

Mobile works, but I still have to test a couple of failure situation
Desktop is basically untested, but I will test later (when I get to my currently remote Mac).

subsurfacewebservice could do with a general workover also for download, but that is for another day (and developer).

janiversen:
upload works for me on a Mac.

but keyboard does not :-( I have tested desktop version, mobileExecutable and mobile (iPad).

dirkhh:
> @dirkhh done (but it was no small task, you asked for).

Remember, I offered that I'd do it.

dirkhh:
Playing with this on a Mac I seem to run into odd hangs, both when importing from divelog and when exporting to divelog. I'll admit that I never use this functionality, so let me try this over on a Linux box later this morning (as debugging things is an order of magnitude easier on Linux).

janiversen:
I do not forget that.

At least you have somewhere else to debug. Since I upgraded my mac to Catalina I have been unable to set a breakpoint (nor lldb or xcode works). Seems Subsurface-mobile gets build without debug symbols.

dirkhh:
Here's what I see when trying to export to divelogs.de from the desktop:
```
QObject::connect: No such signal uploadDiveLogsDE::uploadDiveLogsDE::uploadFinish() in /home/hohndel/src/subsurface/desktop-widgets/subsurfacewebservices.cpp:284
QObject::connect:  (receiver name: 'WebServices')
QObject::connect: No such signal uploadDiveLogsDE::uploadDiveLogsDE::uploadProgress() in /home/hohndel/src/subsurface/desktop-widgets/subsurfacewebservices.cpp:286
QObject::connect:  (receiver name: 'WebServices')
QObject::connect: No such signal uploadDiveLogsDE::uploadDiveLogsDE::uploadStatus() in /home/hohndel/src/subsurface/desktop-widgets/subsurfacewebservices.cpp:288
QObject::connect:  (receiver name: 'WebServices')
```
the export process doesn't seem to ever finish (it's been sitting there for several minutes with no further output)

janiversen:
hmmm I can provoke the same on this Mac.

How I would love to be able to debug, do you know of any build system changes lately that could have affected the debug capability

Xcode does not stop at any breakpoints
and when I do
--> lldb Subsurface
and in lldb
--> list main
it tells me there are no function main.

janiversen:
This problem is solved, however there seems to be another one. Actually very embarrassing I tested with the wrong executable, no wonder it worked. Let me test a bit more, now I found a way to debug

I have pinpointed my debug problem, when Subsurface (similar Subsurface-mobile) is first created in Subsurface.app/Contents, it is 7358784 bytes (with symbols etc) after make is completed it is only 5487120 bytes (and no symbols) so something is calling "strip".

dirkhh:
On Mac we use an odd combination of cmake / Qt tools / Xcode to create the final binary. On Linux when you build in debug mode things don't get stripped, but I experienced exactly what you describe today when first trying to debug this on my Mac... I'd love to hear if you can figure out what is causing us to strip that binary :-/

janiversen:
Now it works on my Mac. If uploading with no dives in the divelist, the error text could be better.

@dirkhh can you please test on your Linux platforms, and of course pass any review comments, you might have.

core/uploadDiveLogsDE.cpp
```diff
@@ -261,13 +261,14 @@ void uploadDiveLogsDE::slot_updateProgress(qint64 current, qint64 total)
...
+	
 	if (total <= 0 || current <= 0)
 		return;
 
-	// Calculate percentage
+	// Calculate percentage as 0.x (values between 0.0 and 1.0)
 	// And signal whoever wants to know
 	qreal percentage = (float)current / (float)total;
-	emit uploadProgress(percentage);
+	emit uploadProgress(percentage, 1.0);
```
dirkhh, line 275: why not just pass `uploadProgress(current, total)` ?

core/uploadDiveLogsDE.cpp
```diff
@@ -246,18 +246,18 @@ void uploadDiveLogsDE::uploadDives(const QString &filename, const QString &useri
 	reply = manager()->post(request, multipart);
 
 	// connect signals from upload process
-	connect(reply, SIGNAL(finished()), this, SLOT(uploadFinished()));
+	connect(reply, SIGNAL(finished()), this, SLOT(slot_uploadFinished()));
```
dirkhh: idiomatically I think I would prefer `uploadFinishedSlot()`, but that's trivial to change later, so this is fine.

desktop-widgets/subsurfacewebservices.cpp
```diff
@@ -493,7 +359,8 @@ void DivelogsDeWebServices::prepareDivesForUpload(bool selected)
 		report_error(tr("No dives were selected").toUtf8());
 		return;
 	}
-	if (prepare_dives_for_divelogs(filename, selected)) {
+
+	if (uploadDiveLogsDE::instance()->prepareDives(filename, selected)) {
```
dirkhh: I love seeing this. NICE!

core/uploadDiveLogsDE.cpp
```diff
@@ -11,6 +11,9 @@
 #include "core/membuffer.h"
```
dirkhh, line 11: comment on the commit message for commit a4db3d0162dae2ec2d
the subject line is way too long, the line number references feel odd to me, simply go through it in order and the reader can easily figure it out
Other than that, I think is the right thing to do.

dirkhh (changes requested):
added review comments for the merging of the backend code.
This was tricky and I think it's mostly well done - I think I found a couple of issues, though.
As is often the case, the PR is getting complex.

I am NOT suggesting that you redo everything, I'm just thinking out loud here.
What would be nicer is to
- create the core/shared implementation and switch desktop to use it (one PR)
- use that shared implementation on mobile (and add the UI) (second PR)

But of course that's not how this came into existence and I think refactoring it that way would be more work than is justifiable.
I reserve the right to muck around with it when I merge the code, thought :-)

Overall we're headed in a good direction, I think.

core/uploadDiveLogsDE.cpp
janiversen: Corrected

core/uploadDiveLogsDE.cpp
```diff
@@ -49,15 +51,13 @@ void uploadDiveLogsDE::doUpload(bool selected, const QString &userid, const QStr
 
 	// Make zip file, with all dives, in divelogs.de format
 	if (!prepareDives(filename, selected)) {
-		report_error(tr("Failed to create upload file %s\n").toUtf8(), qPrintable(filename));
-		emit uploadFinish(false, err);
+		emit uploadFinish(false, tr("Cannot prepare dives, none selected?"));
```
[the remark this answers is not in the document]
janiversen: Yes there are several errors behind this one, and the are all done with report_error. However showing these technical errors to the users proved confusing (e.g. when you have no dives the technical error message is "could not create zip file").

core/uploadDiveLogsDE.cpp
janiversen: done for all slots.

core/uploadDiveLogsDE.cpp
janiversen: Because then I would need to do the calculation on the QML site. The old desktop code also did not use current/total directly. Instead of having arbitrary values it makes sense to me to use percentage, the total is needed to tell if the range is 0-1 or 0-100

core/uploadDiveLogsDE.cpp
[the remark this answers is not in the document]
janiversen: hmmm I commented this earlier, so you might see my comments twice.

Yes there are several technical errors, but they proved to confuse the user. E.g. if you do not have any dives, the technical error message is "cannot create zip file". All errors within the function are reported with report_error, and should go into the log.

desktop-widgets/subsurfacewebservices.cpp
```diff
@@ -344,14 +276,17 @@ void DivelogsDeWebServices::startUpload()
...
+	uploadDiveLogsDE::instance()->doUpload(useSelectedDives,
+							   qPrefCloudStorage::divelogde_user(),
+							   qPrefCloudStorage::divelogde_pass());
+
+	connect(uploadDiveLogsDE::instance(), SIGNAL(uploadFinish(bool, const QString &)),
+			this, SLOT(uploadFinished(bool, const QString &)));
+	connect(uploadDiveLogsDE::instance(), SIGNAL(uploadProgress(qreal, qreal)),
+			this, SLOT(updateProgress(qreal, qreal)));
+	connect(uploadDiveLogsDE::instance(), SIGNAL(uploadStatus(const QString &)),
+			this, SLOT(uploadStatus(const QString &)));
```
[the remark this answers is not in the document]
janiversen: ups! corrected

core/uploadDiveLogsDE.cpp
janiversen: Then I would need to do calculations in QML. The old desktop code also did manipulate current/total. I found it logical to use percentage, and total is there for convenience (and to signal that the range is 0-1 and not 0-100)

core/uploadDiveLogsDE.cpp
janiversen: corrected

desktop-widgets/subsurfacewebservices.cpp
janiversen: thanks

core/uploadDiveLogsDE.cpp
janiversen: Corrected

janiversen:
"I reserve the right to muck around with it when I merge the code, thought :-)"
I have spent quite a lot of time to secure every single commit compiles and works on desktop, desktopMobile and mobile, if you change the order of commit there are not guarantees that bisecting will work. 
If you change the code, a review might be a good idea.

Having said that you are welcome to much around as you want.

core/uploadDiveLogsDE.cpp
dirkhh: makes sense. Thanks for the explanation!

core/uploadDiveLogsDE.cpp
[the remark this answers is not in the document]
dirkhh: yes, a couple of your comments showed up twice here. I have this problem with GitHub sometimes where I need to force a reload for things to show up...

I guess the real question here should have been "are there other potential errors other than 'no dives selected' that the user should see?" - like "out of space"?

dirkhh:
Testing with mobile on desktop:
- I get "Network Error 5"
- then I get upload successful
- and the dialog stays around until I click 'Cancel'

Desktop looks unchanged to before

(again I find bugs in the upload, but again they are consistent with our behavior before your changes)

core/uploadDiveLogsDE.cpp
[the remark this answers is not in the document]
janiversen: Out of space might be one, but I believe having a “simple” error message for the user and having everything in the log. However if you prefer it I will show the technical error.

janiversen:
The dialog stays open pr. design choice so you see it is completed, I can very easily close it in case there are no errors, if you prefer that?

divelogs.de seems to have some temporary connect problems.

dirkhh:
> The dialog stays open pr. design choice so you see it is completed, I can very easily close it in case there are no errors, if you prefer that?

Can we change the name of the button from `Cancel` to `Done` in that case? That would seem more logical.

janiversen:
Changing the text of a standard button was not as easy as I thought, I close the dialog instead.

I have rebased, added 1 commit (last one, to close dialog) and done a push -f.

Hope this was the last hurdle.

dirkhh:
> 
> Hope this was the last hurdle.

I'm disappointed that you are describing the review as a 'hurdle'.

dirkhh (approved):

janiversen:
Sorry I did not mean the review as such, it is actually something I appreciate a lot !!!! This PR had a lot of unplanned changes (like the whole integration) and effects of that, so I am a bit tired of the "git rebase -I HEAD~25" command :-)

As you know I very quickly implement your review suggestions, that is simply because I agree with the reason. Sometimes I believe you not to be correct, and then we talk, which often gives a third (and better) solution. Very few times we simply disagree, but still find a way (like I think you have accepted my feelings about not mixing directories in one commit).

So please do NOT be disappointed, and sorry for not expressing me more correctly.

PR #2428, opened 2019-12-10
A few more model cleanups

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

We had that absolutely horrible way that the pointer returned from `DiveTripModel::instance()` was invalidated if the model was reset, because a new model object was generated.

Make this logic local to the MultiFilterSortModel. This straightens up data and control flow, because now all interactions of the UI are via the MultiFilterSortModel.

The main reason for this PR is that on mobile we want two different proxy-models translating the DiveTripModel and that didn't work out with the old method.

Finally, this PR fixes a few bugs concerning selection and signals.

PS: This still doesn't make it possible to have a DiveTripList and a DiveTripTree model at the same time. The reason being that on filter-reset the model directly modifies the hidden_by_filter flag of the core. The reason for that is that the model sends signals for all dives where hidden_by_filter changed. To move the filter-recalculation to the core, the model would have to remember the old filter-status and I didn't want to do that for now.

### Mentions:

@dirkhh

### DIFF
--- a/core/divefilter.cpp
+++ b/core/divefilter.cpp
@@ -193,7 +193,7 @@ void DiveFilter::startFilterDiveSites(QVector<dive_site *> ds)
 		// When switching into dive site mode, reload the dive sites.
 		// We won't do this in myInvalidate() once we are in dive site mode.
 		MapWidget::instance()->reload();
-		DiveTripModelBase::instance()->recalculateFilter();
+		emit diveListNotifier.filterReset();
 	}
 }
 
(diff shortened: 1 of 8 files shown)


dirkhh (commented):
Reviewed the code and liked it all - but am on an airplane so testing it will be a bit harder. I'll try and if I don't stumble across anything I might merge - or might leave it for later...

dirkhh:
I was able to do some testing and couldn't find anything that broke. But my testing was far from extensive. But I think it makes sense to merge and move forward.

PR #2429, opened 2019-12-10
build-system: avoid stripping executable on Mac.

janiversen:
Add -no-strip to macdeployqt, to ensure the executable contains
debug symbols

Signed-off-by: Jan Iversen <jan@casacondor.com>

### Describe the pull request:

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This PR allows debugging (desktop and mobileExecutable) on a Mac.

PR #2430, opened 2019-12-10
mac: correct Xcode project schemes

janiversen:
As a side effect of #2429 the Xcode project schemes are updated.

Now it is possible to build/run/debug mobileExecutable as
well as desktop from within Xcode.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh:
So this should be merged after #2429 , correct?

janiversen:
it can be done before, but debugging is not really possible until the other is merged

dirkhh:
elsewhere in READMEs and in some comments we seem to assume that things are in `~/src/subsurface` but since I think you are likely the main / only user of this, I'm ok with the `~/develop/src` base path... it's easy enough to either fix locally or add a symlink. And there's nothing inherently "better" about `~/src` vs. `~/develop/src`

PR #2432, opened 2019-12-11
[RFC] build-system: allow shared source across multiple hosts

dirkhh:
This attempts to allow sharing a host directory across multiple builds, target
use case is to have a shared source directory on a VM host and be able to build
from that in a number of VMs without those builds stepping on top of each
other.

Instead of subsurface/build, subsurface/mobile-build,
subsurface/libdivecomputer/build, use a prefix path to allow having true out of
tree builds.

The one shortcoming is that the autotools need to be run in the libdivecomputer
directory - that means this will be run on the first system that starts a
build. But that seems to cause no harm in my testing.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

It's fairly straight forward, trying hard not to break things for people who aren't using the prefix (i.e. 99% of the users), while giving this extra flexibility to people like me who build on several different platforms on the same system.

### Mentions:

not sure if any of you would use this. but I'd love critical eyes on this to make sure I didn't break anything. @glance- you tend to be good at these things...

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -9,22 +9,40 @@
 #
 # it installs the libraries and subsurface in the install-root subdirectory
 # of the current directory (except on Mac where the Subsurface.app ends up
-# in subsurface/build
+# in subsurface/build)
+#
+# by default it puts the build folders in
+# ./subsurface/llibdivecomputer/build (libdivecomputer build)
+# ./subsurface/build                  (desktop build)
+# ./subsurface/build-mobile           (mobile build)
 #
 # there is basic support for building from a shared directory, e.g., with
 # one subsurface source tree on a host computer, accessed from multiple
 # VMs as well as the host to build without stepping on each other - the
 # one exceptioin is running autotools for libdiveconputer which has to
 # happen in the shared libdivecomputer folder
-
-# create a log file of the build
+# one way to achieve this is to have ./subsurface be a symlink; in that
+# case the build directories are libdivecomputer/build, build, build-mobile
+# alternatively a build prefix can be explicitly given with -build-prefix
+# that build prefix is directly pre-pended to the destinations mentioned
+# above - if this is a directory, it needs to end with '/'
 
 # don't keep going if we run into an error
 set -e
 
+# create a log file of the build
+
 exec 1> >(tee build.log) 2>&1
 
 SRC=$(pwd)
+
+if [[ -L subsurface && -d subsurface ]] ; then
+	# ./subsurface is a symbolic link to the source directory, so let's
+	# set up a prefix that puts the build directories in the current directory
+	# but this can be overwritten via the command line
+	BUILD_PREFIX="${SRC}/"
+fi
+
 PLATFORM=$(uname)
 
 BTSUPPORT="ON"
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -2,7 +2,7 @@
 #
 # this should be run from the src directory which contains the subsurface
 # directory; the layout should look like this:
-#.../src/subsurface
+# .../src/subsurface
 #
 # the script will build Subsurface and libdivecomputer (plus some other
 # dependencies if requestsed) from source.
@@ -10,6 +10,12 @@
 # it installs the libraries and subsurface in the install-root subdirectory
 # of the current directory (except on Mac where the Subsurface.app ends up
 # in subsurface/build
+#
+# there is basic support for building from a shared directory, e.g., with
+# one subsurface source tree on a host computer, accessed from multiple
+# VMs as well as the host to build without stepping on each other - the
+# one exceptioin is running autotools for libdiveconputer which has to
+# happen in the shared libdivecomputer folder
 
 # create a log file of the build
 
@@ -40,6 +46,14 @@ while [[ $# -gt 0 ]] ; do
 			# in order to build the dependencies on Mac for release builds (to deal with the macosx-version-min for those
 			# call this script with -build-deps
 			BUILD_DEPS="1"
+			;;
+		-build-prefix)
+			# instead of building in build & build-mobile in the current directory, build in <buildprefix>build
+			# and <buildprefix>build-mobile; notice that there's no slash between the prefix and the two directory
+			# names, so if the prefix is supposed to be a path, add the slash at the end of it, or do funky things
+			# where build/build-mobile get appended to partial path name
+			shift
+			BUILD_PREFIX="$1"
 			;;
 		-build-with-webkit)
 			# unless you build Qt from source (or at least webkit from source, you won't have webkit installed
@@ -73,7 +87,7 @@ while [[ $# -gt 0 ]] ; do
 			;;
 		*)
 			echo "Unknown command line argument $arg"
-			echo "Usage: build.sh [-no-bt] [-quick] [-build-deps] [-build-with-webkit] [-mobile] [-desktop] [-both] [-create-appdir] [-release]"
+			echo "Usage: build.sh [-no-bt] [-quick] [-build-deps] [-build-prefix <PREFIX>] [-build-with-webkit] [-mobile] [-desktop] [-both] [-create-appdir] [-release]"
 			exit 1
 			;;
 	esac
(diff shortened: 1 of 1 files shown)


janiversen:
👍 This is very similar to what I have been using for a very long time. I took a slightly different (but easier approach).

In every build directory I have a link to src (e.g. buildNewest/src/sub..... ) the script changes are a lot easier when the relationship between build dir and src are fixed.

I it useful not only for multi host but also when you build for different scenarios, because I can keep the builds ready.

glance-:
I haven't had the time to propperly look at this, I'll try to find time for it.

Meanwile, I'm just mentioning another approach I used in https://github.com/Subsurface-divelog/subsurface/blob/master/packaging/android/build.sh , is that the script can figure out where it is, thus where the source-tree is, and then just builds in its current working directory.

dirkhh:
I am more than happy to consider different approaches.
One of my concerns was that we typically call this script from the parent directory of the source directory and sadly have now long established that by default the builds are in sub-directories of the source directory. This is of course annoying - but how many things will break if we suddenly change that default - not just in GitHub Actions but in the tooling of many of the random developers working on this, in packaging scripts, etc. That's why I didn't want to change that default.
But if there's an easier way to achieve that, even better.

janiversen:
It would be easy to check if the directory subsurface exist in current directory meaning build “old way” and else build alternative way (any of them)

dirkhh:
> It would be easy to check if the directory subsurface exist in current directory meaning build “old way” and else build alternative way (any of them)

I'm not sure what this would gain us, though. Maybe I'm missing your point.

So far I have seen a few comments on other ways to do things, but no actual feedback on the proposed code here in the PR...

dirkhh:
The commit that I just pushed does the Right Thing (tm) if `./subsurface` is a symlink. This should make things much easier for the scenarios that I think @glance- was talking about.

janiversen:
Elegant solution, works for me (and do not break any of the tools I use).

scripts/build.sh
```diff
@@ -149,7 +181,7 @@ export INSTALL_ROOT
 # make sure we find our own packages first (e.g., libgit2 only uses pkg_config to find libssh2)
```
glance-, line 185: Wtf, You can't do comments on the bits between the hunks. Why github.

Anyway, shouldn't the install-root-bits end up being prefixed to? otherwise all those bits will conflict and overwrite each other.

scripts/build.sh
```diff
@@ -2,23 +2,47 @@
...
 
 SRC=$(pwd)
+
+if [[ -L subsurface && -d subsurface ]] ; then
+	# ./subsurface is a symbolic link to the source directory, so let's
+	# set up a prefix that puts the build directories in the current directory
+	# but this can be overwritten via the command line
+	BUILD_PREFIX="${SRC}/"
+fi
+
```
glance-, line 45: I'd suggest that the subsurface-source-check should be moved from lower to here:
```suggestion

if [[ ! -d "subsurface" ]] ; then
	echo "please start this script from the directory containing the Subsurface source directory"
	exit 1
fi
```

glance-: Also, this could be rather turned into a thingie to find the source tree instead.
```suggestion

if [ ! -d subsurface ] ; then
    pushd "$(dirname "$0")/../../"
    SRC=$PWD
    popd
fi
```
<s>And use that to reference the SUBSURFACE_SOURCE variable to reference source later.</s>

scripts/build.sh
```diff
@@ -41,6 +65,14 @@ while [[ $# -gt 0 ]] ; do
 			# call this script with -build-deps
 			BUILD_DEPS="1"
 			;;
+		-build-prefix)
+			# instead of building in build & build-mobile in the current directory, build in <buildprefix>build
+			# and <buildprefix>build-mobile; notice that there's no slash between the prefix and the two directory
+			# names, so if the prefix is supposed to be a path, add the slash at the end of it, or do funky things
+			# where build/build-mobile get appended to partial path name
+			shift
+			BUILD_PREFIX="$1"
```
glance-, line 74: No reason to quote $1 here, No globbing happens on assignment.

scripts/build.sh
```diff
@@ -149,7 +181,7 @@ export INSTALL_ROOT
 # make sure we find our own packages first (e.g., libgit2 only uses pkg_config to find libssh2)
 export PKG_CONFIG_PATH=$INSTALL_ROOT/lib/pkgconfig:$PKG_CONFIG_PATH
 
-echo Building in $SRC, installing in $INSTALL_ROOT
+echo Building from $SRC, installing in $INSTALL_ROOT
```
glance-: Quote the whole thing as a string to prevent globbing.

scripts/build.sh
```diff
@@ -340,19 +372,17 @@ if [ ! -d libdivecomputer/src ] ; then
...
 
-cd libdivecomputer
-
-mkdir -p build
-cd build
+mkdir -p "${BUILD_PREFIX}libdivecomputer/build"
+cd "${BUILD_PREFIX}libdivecomputer/build"
 
-if [ ! -f ../configure ] ; then
+if [ ! -f $SRC/subsurface/libdivecomputer/configure ] ; then
```
glance-: Quote the argument to [ here. This will break if $SRC contains any "funny" chars or a " " for example.

scripts/build.sh
```diff
@@ -340,19 +372,17 @@ if [ ! -d libdivecomputer/src ] ; then
...
 	# this is not a typo
 	# in some scenarios it appears that autoreconf doesn't copy the
 	# ltmain.sh file; running it twice, however, fixes that problem
-	autoreconf --install ..
-	autoreconf --install ..
+	autoreconf --install $SRC/subsurface/libdivecomputer
+	autoreconf --install $SRC/subsurface/libdivecomputer
```
glance-: More quoting needed.

scripts/build.sh
```diff
@@ -340,19 +372,17 @@ if [ ! -d libdivecomputer/src ] ; then
...
 fi
-CFLAGS="$OLDER_MAC -I$INSTALL_ROOT/include $LIBDC_CFLAGS" ../configure --prefix=$INSTALL_ROOT --disable-examples
+CFLAGS="$OLDER_MAC -I$INSTALL_ROOT/include $LIBDC_CFLAGS" $SRC/subsurface/libdivecomputer/configure --prefix=$INSTALL_ROOT --disable-examples
```
glance-: Quote here to.

scripts/build.sh
```diff
@@ -451,22 +481,22 @@ fi
 
 set -x
 
-cd $SRC/subsurface
 for (( i=0 ; i < ${#BUILDS[@]} ; i++ )) ; do
 	SUBSURFACE_EXECUTABLE=${BUILDS[$i]}
 	BUILDDIR=${BUILDDIRS[$i]}
 	echo "build $SUBSURFACE_EXECUTABLE in $BUILDDIR"
 
+	cd $SRC/subsurface
```
glance-: Quote argument to cd

scripts/build.sh
```diff
@@ -451,22 +481,22 @@ fi
...
 	# pull the plasma-mobile components from upstream if building Subsurface-mobile
 	if [ "$SUBSURFACE_EXECUTABLE" = "MobileExecutable" ] ; then
-		cd $SRC/subsurface
 		bash ./scripts/mobilecomponents.sh
 	fi
 
-	mkdir -p $SRC/subsurface/$BUILDDIR
-	cd $SRC/subsurface/$BUILDDIR
+	mkdir -p $BUILDDIR
+	cd $BUILDDIR
```
glance-: Quote variable here to

scripts/build.sh
```diff
@@ -451,22 +481,22 @@ fi
...
 	export CMAKE_PREFIX_PATH="$INSTALL_ROOT/lib/cmake;${CMAKE_PREFIX_PATH}"
-	cmake -DCMAKE_BUILD_TYPE=$DEBUGRELEASE .. \
+	cmake -DCMAKE_BUILD_TYPE=$DEBUGRELEASE $SRC/subsurface \
```
glance-: And another case where quoting is needed.

scripts/build.sh
```diff
@@ -2,23 +2,47 @@
...
 # don't keep going if we run into an error
 set -e
 
+# create a log file of the build
+
 exec 1> >(tee build.log) 2>&1
 
 SRC=$(pwd)
+
+if [[ -L subsurface && -d subsurface ]] ; then
```
glance-, line 39: No reason to use [[ here, [ is enough. 

Also, use -a rather then &&, or split it into two [-blocks like:
```suggestion
if [ -L subsurface ] && [ -d subsurface ] ; then
```

glance- (commented):
This script needs quiet a bit of quoting to prevent splitting or globbing.

I'd suggest a propper haul through shellcheck. I generally don't use this script, but I might get around to it some day, and meanwhile I'd suggest everyone who writes shell scripts plumbs shellcheck up as a linter in your editor.

scripts/build.sh
dirkhh: Yes, I think that makes sense. Will add.

scripts/build.sh
dirkhh: not sure what you mean by that second set of comments.

"a thingie to find the source tree instead" .. "and use that to reference the SUBSURFACE_SOURCE variable to reference source later"... ?

scripts/build.sh
dirkhh: I am not a connoisseur of $shell. I believe you :-)

dirkhh:
Thanks @glance- for the detailed feedback. _really_ appreciated.
I'll rework and update

scripts/build.sh
glance-: Copy paste error. I copied code from packaging/android/build.sh but later rewrote the suggested code but not the comment. Just strike the last line and it all makes more sense.

dirkhh:
`shellcheck` and `ale` installed. Many, many warnings addressed.
`INSTALL_ROOT` now considers `BUILD_PREFIX` as well.

PR #2433, opened 2019-12-11
Dive list: implement DiveTripModelBase::reset()

bstoeger:
On desktop, resetting the model is realized by generating a new
model object. This is due to the fact that we have two different
models (tree and list) and for switching between those, we have
to create a new object.

On mobile, currently there are no plans to support the list-mode.
Therefore, there is no reason the recreate the object. Instead,
implement a reset() function that reloads the core data.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is another (tiny) preparation for mobile and desktop unification. On desktop, we reset the models by doing a full regeneration of the model-object. Since we don't support list-mode on mobile (yet?) add a function to reset the model.
### Mentions:

@dirkhh

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -379,6 +379,15 @@ void DiveTripModelBase::clear()
 	endResetModel();
 }
 
+void DiveTripModelBase::reset()
+{
+	beginResetModel();
+	clearData();
+	populate();
+	endResetModel();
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
NICE.

PR #2434, opened 2019-12-11
share upload dive-share code (preparing for mobile)

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

moving the actual upload code from desktop-widgets to core.

This PR only rearranges the code to make it useable for the mobile version.

A new PR will make the mobile integration.

### DIFF
--- a/core/uploadDiveShare.cpp
+++ b/core/uploadDiveShare.cpp
@@ -79,9 +79,9 @@ void uploadDiveShare::slot_uploadFinished()
 	reply->deleteLater();
 	timeout.stop();
 	if (reply->error() != 0)  {
-		emit uploadFinish(false, reply->errorString());
+		emit uploadFinish(false, reply->errorString(), reply->readAll());
 	} else {
-		emit uploadFinish(true, tr("Upload successful"));
+		emit uploadFinish(true, tr("Upload successful"), reply->readAll());
 	}
 }
 
@@ -95,7 +95,7 @@ void uploadDiveShare::slot_uploadTimeout()
 	}
 	QString err(tr("dive-share.com not responding"));
 	report_error(err.toUtf8());
-	emit uploadFinish(false, err);
+	emit uploadFinish(false, err, QByteArray());
 }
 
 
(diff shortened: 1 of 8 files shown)


core/uploadDiveShare.cpp
```diff
@@ -0,0 +1,84 @@
...
+	timeout.setSingleShot(true);
+}
+
+
+void uploadDiveShare::doUpload(bool selected, const QString &uid, bool noPublic)
+{
+}
+
+
+void uploadDiveShare::slot_updateProgress(qint64 current, qint64 total)
```
dirkhh, line 60: silly bike-shedding comment about the name. Again, no need to refactor for this, can always simply be renamed later.

desktop-widgets/diveshareexportdialog.cpp
```diff
@@ -105,34 +106,16 @@ void DiveShareExportDialog::finishedSlot()
...
-		request.setUrl(QUrl(DIVESHARE_BASE_URI "/upload"));
-
-	request.setRawHeader("User-Agent", getUserAgent().toUtf8());
-	if (ui->txtUID->text().length() != 0)
-		request.setRawHeader("X-UID", ui->txtUID->text().toUtf8());
-
-	reply = manager()->put(request, json_data);
-
-	QObject::connect(reply, SIGNAL(finished()), this, SLOT(finishedSlot()));
+	uploadDiveShare::instance()->doUpload(exportSelected, uid, noPublic);
```
dirkhh, line 117: this follows the pattern of splitting a code move across two commits.
I understand your reason for it. maybe I shouldn't even comment on it.
this is simply a preference on my part to have that in the same commit...

dirkhh (approved):
I haven't tested the code - I'm sure you did. But from a review perspective I think this is good to go.
I assume there's no reason to wait for the mobile code, correct?

dirkhh:
Thank you for following my suggestion from yesterday to split this into a smaller PR just factoring out the code. This makes the review so much easier.
I should have time later this morning to do a quick sanity test on this one, but unless you tell me otherwise, I expect to merge this today.

dirkhh:
OK, tested and works exactly as before (which doesn't mean that it works well... for example in my test it exported the end pressure of a tank, but not the start pressure - but that is the case with both your code and master, so it has nothing to do with this PR)
Ready to merge from my perspective.

janiversen:
No reason to wait for the other, it can be merged. Once merged I need to rebase the other and do a push -f, but I will not make changes

PR #2435, opened 2019-12-11
upload to dive-share for mobile

janiversen:
### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Allows mobile users to upload to dive-share.

This is now thoroughly tested, the new parts work, however the way it works in general (old code) can be questioned.

This PR concludes the export integration in mobile.

PR #2437, opened 2019-12-12
Small patches to upload* in core

janiversen:

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

During the qml integration, the extended testing and the resulting debugging a couple of issues
was found, some more serious than others.

The commits are small, to include only 1 problem.

This must be merged BEFORE #2435 is merged

### DIFF
--- a/mobile-widgets/qml/Export.qml
+++ b/mobile-widgets/qml/Export.qml
@@ -75,7 +75,6 @@ Kirigami.ScrollablePage {
 		onApply: {
 			if (selectedExport === ExportType.EX_DIVELOGS_DE) {
 				if (fieldUserID.text !== PrefCloudStorage.divelogde_user) {
-
 					PrefCloudStorage.divelogde_user = fieldUserID.text
 				}
 				if (fieldPassword.text !== PrefCloudStorage.divelogde_pass)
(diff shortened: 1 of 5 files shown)


mobile-widgets/qml/Export.qml
```diff
@@ -75,7 +75,6 @@ Kirigami.ScrollablePage {
 		onApply: {
 			if (selectedExport === ExportType.EX_DIVELOGS_DE) {
 				if (fieldUserID.text !== PrefCloudStorage.divelogde_user) {
```
dirkhh, line 77: interestingly, our coding style would suggest to remove the curly braces here...

dirkhh: not asking you to change this, just rambling

dirkhh (approved):
thanks for this PR. I love the granularity of the commits.

mobile-widgets/qml/Export.qml
janiversen: I will bear that in mind, to be honest I tend to use {} too often (I have too often seen If if else go wrong :-)

Issue #2154, opened 2019-07-01
New feature: track pool dives


kittykat:
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [x] New feature request
- [ ] Discussion request

### Issue long description:

For the training that I am doing right now, I need to track my pool sessions as well as my open water dives. Obviously my pool sessions don't count towards my total dive numbers or bottom time.
In general, outside of my current training, I also like to keep track of when I practice which skills in the pool and with whom.

### Operating system:

Linux desktop

### Subsurface version:

4.8.6
4.8.4-663-g5c1446a87acb

### Current behavior:

I've been using "freedive" to mark both my freediving dives and pool session as I don't care about tracking freediving stats

### Expected behavior:

Have another "dive mode" or some sort of "tag" that can be used to track pool sessions which doesn't contribute to total dive numbers or bottom times (or can be excluded from the stats as a setting in the settings menu or a tickbox the stats dialog)

atdotde:
Same answer as on the other topic: Use the filter on tags and select the remaining dives. This should give you the statistics you are looking for. Alternatively, you could use a separate log book file for your pool dives.

bstoeger:
A dive mode does not make sense to me: A pool dive can be either OC, rebreather or freedive, no?

Perhaps really use the free-form tags + filter?

kittykat:
> A dive mode does not make sense to me: A pool dive can be either OC, rebreather or freedive, no?

True

> Perhaps really use the free-form tags + filter?

There's really two problems that I see:
1) If I log my pool dives, they count towards my total and therefore the numbering of open water dives does not match the ones I have to keep in paper form. e.g. if I was just starting training, my paper log open dive 1 would be something like dive 4 in subsurface if I'd had 3 pool sessions.
2) If I use tags and filters, I lose the "yearly stats" functionality (including all the shiny new features in the upcoming version) and have to rely on the "Statistics" tab

I'm not sure what the correct solution would be… ideally, I'd have a tickbox when editing a dive for "don't number or count or include this dive in any way other than keeping a record of it".

With the way I'm tracking pool dives right now (bit less than ideal), it would also work to have a "dive type" checkbox on the yearly stats page to filter there by dive types.

mturkia:
[...]

PR #2440, opened 2019-12-13
Implement dive-invalid flag. Needs some UI work and fine-tuning.

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This implements the "mark dive invalid" feature that has been requested a few times.

This implements the core code. However, the UI code is only a proof of concept: Invalid dives, if shown, are marked by a gray background. This is unfortunate as this is not visible when selected. I'd ask someone else to do the UI part, if there is interest.

The invalid flag is considered for the statistics and when calculating a new dive number. This probably needs some fine-tuning. For example, what about residual N2 in the planner?

### Changes made:

1) Implement invalid-flag
2) Save/load invalid-flag
3) Implement undo-command to set/reset the invalid-flag
4) Update filter when changing the show-invalid-dives preferences
5) Ignore invalid dives in statistics / when calculating new dive numbers

### Related issues:

This feature was suggested as a potential solution for #2154.

### Release note:

Let's wait for approval / refinement.

### Documentation change:

Should be done, yes.
### Mentions:

@dirkhh @willemferguson @atdotde

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -107,6 +107,9 @@ QVariant DiveTripModelBase::tripData(const dive_trip *trip, int column, int role
 	case MobileListModel::TripNotesRole: return QString(trip->notes);
 	}
 #endif
+	// Set the font for all trips alike
+	if (role == Qt::FontRole)
+		return defaultModelFont();
 
 	if (role == TRIP_ROLE)
 		return QVariant::fromValue(const_cast<dive_trip *>(trip)); // Not nice: casting away a const
@@ -187,6 +190,15 @@ static QString displayWeight(const struct dive *d, bool units)
 		return s + gettextFromC::tr("lbs");
 }
 
+static QFont struckOutFont()
+{
+	QFont font;
+	font.setStrikeOut(true);
+	return font;
+}
+static QBrush invalidForeground(Qt::gray);
+static QFont invalidFont = struckOutFont();
+
 QVariant DiveTripModelBase::diveData(const struct dive *d, int column, int role)
 {
 #ifdef SUBSURFACE_MOBILE
@@ -235,6 +247,10 @@ QVariant DiveTripModelBase::diveData(const struct dive *d, int column, int role)
 	}
 #endif
 	switch (role) {
+	case Qt::FontRole:
+		return d->invalid ? invalidFont : defaultModelFont();
+	case Qt::ForegroundRole:
+		return d->invalid ? invalidForeground : QVariant();
 	case Qt::TextAlignmentRole:
 		return dive_table_alignment(column);
 	case Qt::DisplayRole:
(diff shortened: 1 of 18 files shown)


bstoeger:
Sorry for the color-explosion-banner. I just scrolled through the tag-list and checked everything that applies. 😛

dirkhh:
I think this should best be looked at by someone who actually expressed interest in this. I always found this a very strange concept.

bstoeger:
I'll ask on the mailing list. If there is no interest, we can also remove the stub and the obscure preferences entry. Fine with me.

janiversen:
just for information, we have a case in issue #2555 where the invalid flag could be useful.

Also I think, maybe wrongly, that #2513 might be related to this PR as well, or more correctly some work regarding invalid flag already merged.

bstoeger:
Rebased on master - that was somewhat painful with all the prefs churn. On the mailing list, there were two or three positive, though not ecstatic, comments and no negatives. In a quick test, the main problem was that if a dive is selected you don't see whether it is invalid or not. So this would need some UI work. Also it is unclear if "invalid" dives should be considered by the planner to calculate remaining gas load, or whatever magic it does. :-P

In any case, I'd really like to come to a decision here. I'd prefer to not have to rebase again.

bstoeger:
Update: Make the font italicized for invalid dives. Looks absolutely horrid, but at least one now recognized invalid dives when selected. The best I can come up with (apart from writing a custom "delegate", which I *really* don't feel like doing).

dirkhh:
I am hoping to find time to play with this today... sorry to not have provided feedback earlier.
Oddly not going out hasn't improved the net number of available hours for Subsurface...

dirkhh:
OK, played with this.
You are a UI genius. If this professor thing doesn't work for you, consider a career as a visual designer. That use of italics is inspired and game changing. I'm all in.
:-)
The feature seems to work as intended, I wasn't able to break anything or to spot a scenario where it seemed broken. My testing wasn't comprehensive, I'm sure, but I tried what felt like the obvious scenarios.
Hopefully later this afternoon I'll try to come up with something even more inspired concerning the UI, but with that one caveat, I don't have a lot to complain about.
Thank you.

bstoeger:
New attempt: use struck out font instead of italics - I think that's universally understood as "invalid".

dirkhh:
Looked at your new UI approach. Definitely better. That is indeed intuitive.
One thing that I found strange when playing with it was the "make valid" wasn't the top entry in the context menu for an invalid dive. That's of course a pain because that means having two entries there, exactly one of which is visible at a time - but when I played with this I had the pretty strong feeling that the "make valid" entry should be the first.

Now we need to implement something similar for the mobile UI - but that's not required to merge this part, I guess. Oh - my mobile build is broken with this branch - but it seems to work on GitHub Actions, so I'm guessing it's something that I borked here.
**Edited to add:** _yep, my mobile build directory was completely messed up, no idea how I did that... blew it away and did a clean build, no issues - except that we have no UI support for invalid dives, of course_

qt-models/divetripmodel.cpp
```diff
@@ -187,6 +190,15 @@ static QString displayWeight(const struct dive *d, bool units)
 		return s + gettextFromC::tr("lbs");
 }
 
+static QFont struckOutFont()
+{
+	QFont font;
+	font.setStrikeOut(true);
```
dirkhh, line 196: good choice - but doesn't match the commit message :-)

dirkhh (approved):
I almost clicked merge, then I noticed that the commit message of 0d45e5c8 may need some work :-)

dirkhh:
for some reason I no longer get notifications when you push things...
yes, this is good. thanks!

PR #2441, opened 2019-12-13
Coding style: use std::equal_to instead of lambda

bstoeger:
Use std::equal_to instead of lambdas that compare two dive pointers.
One could argue over which version is more readable. For whatever
it's worth, std::equal_to is more compact and expressive.

This removes an old erroneous comment that stated that std::equal_to
is only available since C++14.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Random trivial cleanup. All information is in the commit message.

### Changes made:

1) Use std::equal_to instead of lambda-expression.


dirkhh (approved):
certainly easier on the eyes

PR #2442, opened 2019-12-13
Diveplanner for mobile, first PR of a larger number of PR´s

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Diveplanner is probably the most wanted feature in mobile, this PR is the first of a longer series to make this happen.

The work to make planner available in a fashion compatible with the desktop version will be made in phases:
- Remove functional code from UI and move to a plannerShared (to avoid duplicated code)
- Planner Setting UI
- Planner Edit UI
- Planner Edit from dive UI
- Planner View UI
- Planner manage dive plans UI

The biggest challenge is to integrate the planner in a non-intrusive mode, and for each step secure that the desktop functionality is unchanged.

Focus is on not breaking the desktop, which is the cause for smaller commits and tests after each commit. 

The review comments from the "export in mobile" PR´s have been incorporated.

### Changes made:

Move conversion of ascent/descend settings to shared code.

### Mentions:

This PR is really pure code move, but it would be nice if one of dive planner developers keep an eye on this and comming PR´s, at least review comments are more than welcome.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -72,6 +72,8 @@ set(SUBSURFACE_CORE_LIB_SRCS
 	divelist.h
 	divelogexportlogic.cpp
 	divelogexportlogic.h
+	plannershared.cpp
+	plannershared.h
 	divesite-helper.cpp
 	divesite.c
 	divesite.h
--- a/core/plannershared.cpp
+++ b/core/plannershared.cpp
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "plannershared.h"
+
+
+plannerShared *plannerShared::instance()
+{
+    static plannerShared *self = new plannerShared;
+    return self;
+}
--- a/core/plannershared.cpp
+++ b/core/plannershared.cpp
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "plannershared.h"
+#include "core/pref.h"
+#include "core/settings/qPrefDivePlanner.h"
 
 
 plannerShared *plannerShared::instance()
@@ -7,3 +9,53 @@ plannerShared *plannerShared::instance()
     static plannerShared *self = new plannerShared;
     return self;
 }
+
+// Used to convert between meter/feet and keep the qPref variables independent
+#define TO_MM_BY_SEC ((prefs.units.length == units::METERS) ? 1000.0 / 60.0 : feet_to_mm(1.0) / 60.0)
+
+// Converted meter/feet qPrefDivePlanner values
+int plannerShared::ascratelast6m()
+{
+	return lrint(prefs.ascratelast6m / TO_MM_BY_SEC);
+}
+void plannerShared::set_ascratelast6m(int value)
+{
+	qPrefDivePlanner::set_ascratelast6m(value * TO_MM_BY_SEC);
+}
+
+int plannerShared::ascratestops()
+{
+	return lrint(prefs.ascratestops / TO_MM_BY_SEC);
+}
+void plannerShared::set_ascratestops(int value)
+{
+	qPrefDivePlanner::set_ascratestops(value * TO_MM_BY_SEC);
+}
+
+int plannerShared::ascrate50()
+{
+	return lrint(prefs.ascrate50 / TO_MM_BY_SEC);
+}
+void plannerShared::set_ascrate50(int value)
+{
+	qPrefDivePlanner::set_ascrate50(value * TO_MM_BY_SEC);
+}
+
+int plannerShared::ascrate75()
+{
+	return lrint(prefs.ascrate75 / TO_MM_BY_SEC);
+}
+void plannerShared::set_ascrate75(int value)
+{
+	qPrefDivePlanner::set_ascrate75(value * TO_MM_BY_SEC);
+}
+
+int plannerShared::descrate()
+{
+	return lrint(prefs.descrate / TO_MM_BY_SEC);
+}
+void plannerShared::set_descrate(int value)
+{
+	qPrefDivePlanner::set_descrate(value * TO_MM_BY_SEC);
+}
+
--- a/core/plannershared.h
+++ b/core/plannershared.h
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef PLANNERSHARED_H
+#define PLANNERSHARED_H
+#include <QObject>
+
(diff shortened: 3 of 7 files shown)


janiversen:
"build dependencies" does not seem like a problem in this code ?

glance-:
"Conflicting distribution: https://packages.microsoft.com/repos/microsoft-ubuntu-bionic-prod bionic InRelease (expected bionic but got stable)" , some bug in azure-mirror. Restarted the check.

Edit: Still the same issue, so it wasn't just transient, but its not a issue in your code.

janiversen:
Thanks, I assume it is the same problem in the two other PR´s.

core/plannerShared.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+
+plannerShared *plannerShared::instance()
+{
+    static plannerShared *self = new plannerShared;
+    return self;
+}
+
+// Used to convert between meter/feet and keep the qPref variables independent
+#define UNIT_FACTOR ((prefs.units.length == units::METERS) ? 1000.0 / 60.0 : feet_to_mm(1.0) / 60.0)
```
atdotde: Maybe use a name that signals this is a velocity unit and that it is internal velocity / user velocity (so it is obvious if you have to multiply or divide).

core/plannerShared.h
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef PLANNERSHARED_H
+#define PLANNERSHARED_H
+#include <QObject>
+
+// This is a shared class (mobile/desktop), and contains the core of the diveplanner
```
atdotde: Do I understand correctly, this is still the user facing part of the planner (i.e. the C++ part) not the internals (in C)?

atdotde (commented):
I think it is great you started this. Of course the million dollar question is: How do you envision the user interaction on mobile without mouse and keyboard? Figuring this out has always been my show stopper for going in this direction. Just to state my expectation: The USP of the Subsurface dive planner is that it is so interactive, that you can immediately see the consequences of making changes to the profile, gases, deco parameters etc. There are already enough mobile deco planners where you have to enter the dive plan in a cumbersome way and then press some "calculate button" to spit out the plan. I think we need to do much better than that to make it worth the effort. On a tablet, something not too far from what we do on the desktop might be possible (which would need a profile which is actually a widget and not just an image of a widget as is currently on mobile) but for phones I can imagine coming up with a design that is not totally unpractical is a tough one.

core/plannerShared.cpp
janiversen: How about <variable name>Unit ? 

I totally agree that different names would be best, but I failed to come up with a convincing name.

core/plannerShared.h
janiversen: It it sitting in the middle, with the C code on one side and the UI/QML on the other.

E.g. UNIT_FACTOR, would need to be rebuild for qml and thus duplicated, this class avoids the duplication.

core/plannerShared.cpp
atdotde: what about TO_MM_BY_SEC ?

janiversen:
Thanks for your comments, I agree getting the UI right is a challenge. We have a version running but made without any thought of back porting, here the iPhone (or lets say devices with small screen) only can view dive plans, the iPad can also edit. The real pain in the neck is getting the graphic to work.

As I do more PR´s it would be nice to discuss the user interface (and of course also the implementation).

glance-:
I was thrilled when I saw the subject of this PR. I've bin thinking about this for a long time, but never gotten around to it, and I think you're starting in the right end, by refactoring the current code to make the infrastructure available in the mobile apps.

I don't think we need to solve the whole UI problem right now. We can start out with a "engineering-UI" with just depth/time that produces a table as a result, and it doesn't matter. When we figure out the UI, we can tackle that part of the problem.

core/plannerShared.cpp
janiversen: Just to be sure, I do not misunderstand you.

You talk ONLY about the macro, not the variable names (ascratelast6m etc.) which are exposed externally to plannerShared ?

If you only talk about the macro, I will make the change immediately.

core/plannerShared.h
glance-: Yea, this UNIT_FACTOR convertion thingie I didn't like at all. I know It's there since before, but I think we should clean that up.

We have a velocity_t in core/profile.h , but thats rather just colors for the profile. I think we should have a propper velocity_t which describes this and convertion helpers around that.

glance-: I think mm/sec is the right unit for a velocity_t for the planner.

core/plannerShared.cpp
glance-: Yea, this UNIT_FACTOR convertion thingie I didn't like at all. I know It's there since before, but I think we should clean that up.

We have a velocity_t in core/profile.h , but thats rather just colors for the profile. I think we should have a propper velocity_t which describes this and convertion helpers around that.

I think mm/sec is the right unit for a velocity_t for the planner.

janiversen:
@atdotde I have changed the macro name, and I look forward to hear your thoughts as we go further.

core/plannerShared.cpp
janiversen: I have put a "velocity_t" on my agenda, but I would prefer to have this simple step first, because it is very easy to review, that I have not broken anything, and then in a second step replace the macro with a proper velocity_t.

glance-: I'm completely fine with that.

henrik242:
Awesome initiative!

henrik242:
> There are already enough mobile deco planners where you have to enter the dive plan in a cumbersome way and then press some "calculate button" to spit out the plan

I find this method works fairly well in e.g. MultiDeco, although they have implemented it terribly. We could do *so* much better in Subsurface.

janiversen:
I believe to have reacted in a positive manner to all comments and review. Please advice what more is expected. Thanks in advance.

glance- (approved):

janiversen:
Sorry for the number of force pushes, based on a reviewer comment (that I highly agree with) I changed filenames from plannerShared.* to plannershared.*  and of course the corresponding CMake and include changes.

NO functionality was changed.

This PR is ready to be merged.

dirkhh (approved):
Excellent!
Thank you for the exhaustive responses to all the feedback.
Much appreciated.

PR #2444, opened 2019-12-13
qml fixes: default cylinder and cloud username

janiversen:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Corrected default_cylinder
Corrected cloud Username.

### DIFF
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -45,7 +45,7 @@ Kirigami.ScrollablePage {
 				color: subsurfaceTheme.textColor
 			}
 			Controls.Label {
-				text: prefs.credentialStatus === CloudStatus.CS_NOCLOUD ? qsTr("Not applicable") : prefs.cloudUserName
+				text: PrefCloudStorage.credentialStatus === CloudStatus.CS_NOCLOUD ? qsTr("Not applicable") : PrefCloudStorage.cloud_storage_email
 				font.pointSize: subsurfaceTheme.regularPointSize
 				Layout.preferredWidth: gridWidth * 0.60
 				color: subsurfaceTheme.textColor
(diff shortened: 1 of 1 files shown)


janiversen:
third "build dependencies" clearly something is not right.

mobile-widgets/qml/Settings.qml
```diff
@@ -45,7 +45,7 @@ Kirigami.ScrollablePage {
 				color: subsurfaceTheme.textColor
 			}
 			Controls.Label {
-				text: prefs.credentialStatus === CloudStatus.CS_NOCLOUD ? qsTr("Not applicable") : prefs.cloudUserName
+				text: prefs.credentialStatus === CloudStatus.CS_NOCLOUD ? qsTr("Not applicable") : PrefCloudStorage.cloud_storage_email
```
dirkhh: overall I still find the way we access these values from QML highly inconsistent and confusing.
Here we are now using `prefs` and `PrefCloudStorage` on the same line...

dirkhh (commented):
on the whole I don't see an advantage of one over the other.
The access to C++ data is still the worst part of working with QML

mobile-widgets/qml/Settings.qml
janiversen: Sadly enough prefs.credentialstatus is a calculated value in mobile-widgets/prefs.cpp, otherwise I would have replace that. I am working on a bigger cleanup of the login procedure to highly simplify it (but keep exact the same functionality), and this commit is basically part of that.

I agree that the QML - C++ interface not is the best in the world, I am slowly trying to make it better, by e.g. eliminating prefs (which are a leftover from the early days of qPref and thus in reality a parallel implementation to qPref just to add confusion.

I do believe with every patch the interface get clearer and more precisely defined.

janiversen:
Changed commits to tell the problem/intention directly.

Also changed the second commit, so pref and PrefCloudStorage are not on the same line.

dirkhh (approved):

PR #2445, opened 2019-12-14
Create new root directory "backend-shared" sitting between desktop-widgets/mobile-widgets and the rest

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ x] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR sits on top of  #2442, and that needs to be merged first.

It was decided to put the backend functions (exportfuncs, plannerShared) in core, which is
already a shared lib. However e.g. plannerShared needs access to qt-models, which are not 
build when building core.

As a consequence of this a new root directory is created "backend-shared" to contain classes that sits between the pure functionality (like core and qt-models) and the UI (like desktop-widgets and mobile-widgets).


janiversen:
@atdotde, @glance- just to let you know, the dive planner shared files have changed place.

A question:
Looking at diveplanner in the desktop version, the parts:
Rates, Planning, Gas options.

Those all seem to be something a diver set (in a SIMPLE dive planner) once and not for every dive plan, so it could be seen as an extension of preferences. Am I wrong in this assumption ?

janiversen:
Resolved rebase conflict in Xcode project.

dirkhh (changes requested):
See my email to the mailing list. As the first hunk in the first commit shows, this is changing our established pattern of lower case filenames. I'd prefer if we didn't do that.

janiversen:
I have not subscribed to the subsurface mailing for personal reasons, but I will try and find the file in question and change it.

janiversen:
The problem is actually in the first PR (diveplanner branch), I have done a push -f on that, now I just need to rebase this one.

janiversen:
Sorry for the number of force pushes, based on a reviewer comment (that I highly agree with) I changed filenames from plannerShared.* to plannershared.* and of course the corresponding CMake and include changes.

NO functionality was changed.

This PR is ready to be merged.

dirkhh:
I think based on the recent merges of older PRs we now have merge conflicts here.
Since you appear to be actively working on things and responding, I think it's best if I wait for a clean push from you, do another review and then hopefully merge this so we stop being stalled on older PRs that have brought agreement...
Thanks

core/plannershared.cpp
```diff
@@ -1,61 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0
-#include "plannershared.h"
-#include "core/pref.h"
-#include "core/settings/qPrefDivePlanner.h"
+#include "plannerShared.h"
```
dirkhh: ahh, this seems to still have camelCase

backend-shared/plannershared.cpp
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-2.0
-#include "plannerShared.h"
+#include "plannershared.h"
```
dirkhh: and here you clean it up :-)

dirkhh (changes requested):
took a quick look:
- there is a camelCase include that later gets fixed - maybe those could just be merged into one commit
- you don't need to warn about multi-directory commits. I continue to believe that you are the only one worried about those and continue not to view them as a big deal at all

backend-shared/plannershared.cpp
janiversen: Should be corrected ??

janiversen:
please advice if there are any open issues on this PR ?

dirkhh (approved):
Thanks
An example for what I consider an awesome PR
A hand full of small commits
Addresses one thing
Easy to wrap my mind around and understand.

PR #2446, opened 2019-12-15
plannerShared, add planning variables.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Added first part of the planning variables (see desktop diveplanner window).
This integrates calls to plannerModel and therefore additional qt_models is needed

Since commands (undo) is not available for mobile, these calls (only a handful) have been
left out using a #ifndef SUBSURFACE_MOBILE

### Mentions:

@atdotde @glance- this continues the first PR you reviewed, it would be nice if you would cast an eye on this as well.


janiversen:
Very strange, this iOS failure.

When running "cd packaging/iOS; ./build.sh" it works, but travis do not like it.

Could it be travis have some cached Cmake setup ?

janiversen:
My bad, I had forgotten how fragile this iOS build is. Let me see if I can find some time and bring my tools in sync with this.

dirkhh:
you mentioned elsewhere (I think) that you were considering moving the iOS build to Cmake. Not sure that's worth it, but also not opposed. It will cause me work to make sure that I can still create packages that I can push to the app store, but that's a price I'm willing to pay for not having to worry about the two parallel build systems (or, actually, three, as Xcode is mixed in with that as well)
Thanks

janiversen:
I believe your comment have no relation to this PR, but anyhow I have duly noted your concerns. You asked if I could activate the qt compiler in CMake which I already have, but as part of a IOS integration.

However as mentioned this had a very low priority for me, and now even less priority.

janiversen:
I believe this PR is ready to get merged, if there are anything outstanding please advice.

dirkhh:
I have been called to participate in family activities, so I need to pause here for now, likely for today.
I will pick all of this up as soon as I have the time.

janiversen:
Heh, I am in quite the contrary situation, I am all alone (apart from the family dog), so expect some changes from me the coming days.

But I hope you will have a pleasant family time.

janiversen:
resubmitting in smaller chunks

PR #2448, opened 2019-12-17
build-system iOS: clean build.sh to follow same rules as scripts/build.sh

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ x] Build system change
- [ ] Documentation change
- [ ] Language translation

change build.sh to follow the same location for build directories as scripts/build.sh.

This is part of a larger change in the iOS build system, which currently are very
fragile and error prune (e.g. all file names must be specified in Subsurface-mobile.pro).

### DIFF
--- a/packaging/ios/build.sh
+++ b/packaging/ios/build.sh
@@ -145,8 +145,8 @@ for ARCH in $ARCHS; do
 	autoreconf --install
 	popd
 	if [ ! -e $PKG_CONFIG_LIBDIR/libxslt.pc ] ; then
-		mkdir -p build-ios/libxslt-build-$ARCH_NAME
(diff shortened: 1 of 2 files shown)


packaging/ios/build.sh
```diff
@@ -48,7 +48,7 @@ popd
 # prepare build dir
 mkdir -p build-ios
 
-IOS_QT=${TOP}/Qt
+IOS_QT=~/Qt
```
dirkhh, line 51: Funny story... the reason for this is that I used to build macOS and iOS from the same tree - and at times had different Qt versions. I no longer do that, so this is fine with me.

dirkhh (approved):

PR #2450, opened 2019-12-18
mobile: smaller changes to login screen

janiversen:

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

the "hamburger" was visible in the login screen, allowing the user to bypass login, without setting the proper status.

Save password local is default false, change to true (securing the user do not need to change it, remark the login screen does not contain this checkbox).

remove a couple of unused functions.

This PR is preparation to remove a big part of qmlprefs, but due to size of PR this will come afterward.

### DIFF
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -26,11 +26,9 @@ Kirigami.ApplicationWindow {
 	pageStack.globalToolBar.preferredHeight: Math.round(Kirigami.Units.gridUnit * (Qt.platform.os == "ios" ? 2 : 1.5))
 	pageStack.globalToolBar.maximumHeight: Kirigami.Units.gridUnit * 2
 
-	property alias oldStatus: prefs.oldStatus
 	property alias notificationText: manager.notificationText
 	property alias locationServiceEnabled: manager.locationServiceEnabled
 	property alias pluggedInDeviceName: manager.pluggedInDeviceName
-	property alias showPin: prefs.showPin
 	property alias defaultCylinderIndex: settingsWindow.defaultCylinderIndex
 	property bool filterToggle: false
 	property string filterPattern: ""
@@ -249,8 +247,8 @@ Kirigami.ApplicationWindow {
 					if (prefs.credentialStatus == CloudStatus.CS_UNKNOWN) {
 						// the user has asked to change credentials - if the credentials before that
 						// were valid, go back to dive list
-						if (oldStatus == CloudStatus.CS_VERIFIED) {
-							prefs.credentialStatus = oldStatus
+						if (prefs.oldStatus == CloudStatus.CS_VERIFIED) {
+							prefs.credentialStatus = prefs.oldStatus
 						}
 					}
 					returnTopPage()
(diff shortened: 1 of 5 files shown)


janiversen:
force pushed with latest subsurface-master, ready to be merged.

subsurface-mobile-main.cpp
```diff
@@ -64,6 +65,7 @@ int main(int argc, char **argv)
 
 	// some hard coded settings
 	qPrefDisplay::set_animation_speed(0); // we render the profile to pixmap, no animations
+	qPrefCloudStorage::set_save_password_local(true);
```
dirkhh, line 68: this value isn't used at all in Subsurface-mobile.
https://github.com/Subsurface-divelog/subsurface/blob/967f0e56d79c4c25e54619ef8f56770baa9edaa8/mobile-widgets/qmlmanager.cpp#L540
Therefor setting it here makes no sense, nor does the commit message that assumes that this is a thing.

mobile-widgets/qmlmanager.cpp
```diff
@@ -537,11 +536,9 @@ void QMLManager::saveCloudCredentials()
...
 		return;
 
-	s.beginGroup("CloudStorage");
-	s.setValue("email", QMLPrefs::instance()->cloudUserName());
-	s.setValue("password", QMLPrefs::instance()->cloudPassword());
-	s.setValue("cloud_verification_status", QMLPrefs::instance()->credentialStatus());
-	s.sync();
+	qPrefCloudStorage::set_cloud_storage_email(QMLPrefs::instance()->cloudUserName());
+	qPrefCloudStorage::set_cloud_storage_password(QMLPrefs::instance()->cloudPassword());
+	qPrefCloudStorage::set_cloud_verification_status(QMLPrefs::instance()->credentialStatus());
```
dirkhh, line 541: I see absolutely zero value in changes like this.
We have code that works. You split over two commits changes that make that code work differently.
And that I now need to spend time figuring out if it breaks something or has unintended side effects.
How is that good use of my time?

mobile-widgets/qml/main.qml
```diff
@@ -52,8 +52,8 @@ Kirigami.ApplicationWindow {
 		id: fontMetrics
 		Component.onCompleted: {
 			manager.appendTextToLog("Using the following font: " + fontMetrics.font.family
-				    + " at " + subsurfaceTheme.basePointSize + "pt" +
-				    " with mobile_scale: " + PrefDisplay.mobile_scale)
+					+ " at " + subsurfaceTheme.basePointSize + "pt" +
+					" with mobile_scale: " + PrefDisplay.mobile_scale)
```
dirkhh: random whitespace change in the middle of an unrelated commit

dirkhh (changes requested):
a couple of useful changes, mixed in with other things that I would have to think a lot more about than I really want to right now.
I really don't understand how you decide what goes into one PR. And how you split commits.
How can the first and third commit in this series not be one commit. I wouldn't want to take that commit anyway, but at least it would have made sense together.

subsurface-mobile-main.cpp
janiversen: that is actually not true. if save_password_local == false it will ask you for a password every time. it is however true it is not used in QML directly, but as mentioned indirectly.

mobile-widgets/qmlmanager.cpp
janiversen: I see where you are heading...might be better to abandon the whole thing. 

Actually the code that "works" does not respond correctly to default values, something you (thanks) introduced into qPrefs.

dirkhh: that's a fair criticism of my comment.

I withdraw the statements that there's "absolutely zero value".

mobile-widgets/qml/main.qml
janiversen: Corrected, sorry for the problem.

dirkhh (approved):

PR #2451, opened 2019-12-20
mobile: remove QMLPrefs::cloudUserName / cloudPassword (change of user experience!)

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

replaced cloudUserName and cloudPasword with qPrefCloudStorage::

Please note that the first commit "do not revert to old credentials." caused a change in the user experience in a special situation (when cancelling PIN setup).

This PR sits on top of #2450, and should not be merged before #2450 is merged

### DIFF
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -225,7 +225,7 @@ Kirigami.ApplicationWindow {
 						visible: text.length > 0
 						level: 3
 						color: "white"
-						text: prefs.cloudUserName
+						text: PrefCloudStorage.cloud_storage_email
 						wrapMode: Text.NoWrap
 						elide: Text.ElideRight
 						font.weight: Font.Normal
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -244,13 +244,6 @@ Kirigami.ApplicationWindow {
 				text: qsTr("Dive list")
 				onTriggered: {
 					manager.appendTextToLog("requested dive list with credential status " + prefs.credentialStatus)
-					if (prefs.credentialStatus == CloudStatus.CS_UNKNOWN) {
-						// the user has asked to change credentials - if the credentials before that
-						// were valid, go back to dive list
-						if (prefs.oldStatus == CloudStatus.CS_VERIFIED) {
-							prefs.credentialStatus = prefs.oldStatus
-						}
-					}
 					returnTopPage()
 					globalDrawer.close()
 				}
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -50,8 +50,8 @@ Kirigami.ApplicationWindow {
 		id: fontMetrics
 		Component.onCompleted: {
 			manager.appendTextToLog("Using the following font: " + fontMetrics.font.family
-				    + " at " + subsurfaceTheme.basePointSize + "pt" +
-				    " with mobile_scale: " + PrefDisplay.mobile_scale)
+					+ " at " + subsurfaceTheme.basePointSize + "pt" +
+					" with mobile_scale: " + PrefDisplay.mobile_scale)
 		}
 	}
 	visible: false
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -74,7 +74,7 @@ Item {
 
 		Controls.TextField {
 			id: password
-			text: prefs.cloudPassword
+			text: PrefCloudStorage.cloud_storage_password
 			visible: !rootItem.showPin
 			echoMode: TextInput.PasswordEchoOnEdit
 			inputMethodHints: Qt.ImhSensitiveData |
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -58,7 +58,7 @@ Item {
 
 		Controls.TextField {
 			id: login
-			text: prefs.cloudUserName
+			text: PrefCloudStorage.cloud_storage_email
 			visible: !prefs.showPin
 			Layout.fillWidth: true
 			inputMethodHints: Qt.ImhEmailCharactersOnly |
@@ -101,7 +101,7 @@ Item {
 			visible: prefs.showPin
 			SsrfButton {
 				id: registerpin
-				text: qsTr("Register") 
+				text: qsTr("Register")
 				onClicked: {
 					saveCredentials()
 				}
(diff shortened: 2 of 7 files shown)


mobile-widgets/qml/main.qml
```diff
@@ -244,13 +244,6 @@ Kirigami.ApplicationWindow {
 				text: qsTr("Dive list")
 				onTriggered: {
 					manager.appendTextToLog("requested dive list with credential status " + prefs.credentialStatus)
-					if (prefs.credentialStatus == CloudStatus.CS_UNKNOWN) {
-						// the user has asked to change credentials - if the credentials before that
-						// were valid, go back to dive list
-						if (prefs.oldStatus == CloudStatus.CS_VERIFIED) {
-							prefs.credentialStatus = prefs.oldStatus
-						}
-					}
```
dirkhh, line 253: well, this is a fundamental design change. And not one that I necessarily agree with.

Either way, having this in the middle of a PR that says that it removes a certain variable seems... suboptimal

mobile-widgets/qml/main.qml
```diff
@@ -165,6 +165,8 @@ Kirigami.ApplicationWindow {
 	globalDrawer: Kirigami.GlobalDrawer {
 		id: gDrawer
 		height: rootItem.height
+		enabled: (prefs.credentialStatus === CloudStatus.CS_NOCLOUD ||
+			 prefs.credentialStatus === CloudStatus.CS_VERIFIED)
```
dirkhh: this is a useful change. It has nothing to do with this PR.

dirkhh (changes requested):
This PR mixes useful cleanup with behavior change with changes that I think add more noise than value (which is, of course, an expression of my personal perception vs. what is clearly your perception since you are submitting them)

This should be at least three PRs. And you may get at least some questions on a couple of them.

mobile-widgets/qml/main.qml
janiversen: The commit message, I believe is pretty clear. Added text in the PR to make it even more clear.

The word "fundamental" seems a bit "over the top". In essence this is caused by the fact, that the old version allowed the user to escape the login, without ever having done a login or choose "no-cloud", something I find is a fundamental design problem. With this PR the user is locked in the login page, until one of 3 things happen:
1) enter verified credentials (divelist is loaded from cloud)
2) select no-cloud mode (divelist is loaded from local)
3) enter new credentials, PIN page shows, enter PIN or cancel (back to enter credentials).

Cancel PIN behaves strongly in the old version. A user enters new credentials, sees the PIN screen, but does not receive the email immediately, so works with other applications (causing the mobile app to close). Having received the email with the PiN, the user starts mobile again, BUT does not see the PIN screen, instead the old credentials are used.

If revert to old credentials is a feature, then at least it should work in all situations (just my opinion).

But of course having read your highly de-motivating opinion of my work, I understand your comment.

mobile-widgets/qml/main.qml
janiversen: Your are totally correct, it is part of the 3 PR´s that provides a more uniform login (e.g. the user cannot escape, without having chosen how to load the divelst).

I apologize ! it should have been in newlogin3, I wanted to make the 3 login PR´s as one, because they really work together, but split it in 3 to keep the number of commits in each low.

I can move it, if you really want it, but it need to be in one of the 3 PRs because it is a fundamental limitation. Please advice ?

janiversen:
Seems something funny have happened with the comments on this PR. I remember having responded to:
quote <<yet another round of "I decided that we need to access these values a certain way, let's change everything so we only use that one certain way".<<
and tried to do it in a polite way, but it seems gone but the cause of my comments are still here.

I am still offended by having this hanging on my back, and see it as a rather unfair comment on my work, however I would understand if the comment was made in "the heat of the moment", but then it should be edited just as it happened to my primary response.

I am inclined to close the "new login" series as they obviously causes a lot of dismay, but I will leave it open for now.

dirkhh:
All my comments here are "in the heat of the moment" as I typically feel "behind".
I have had this happen to me before that after a force push or some seemingly unrelated change some of my comments / responses disappear. But it's always one of those "didn't I respond to this?" moments and of course you can't go back and check.
I seem to remember seeing a response from you.

Please don't take these comments any more negative than they are meant. They express an emotional response as I read patches. Not a verdict that I consider you an insufficient human being.

I try to be very specific and criticize actions, not people. If this comes across differently, then I need to improve my writing. But given how hard it is to stay on top of all of these various PRs, that will slow me down even more.

dirkhh:
I've tried to merge all the PRs where I felt the current state is reasonable, even if sometimes I have small remaining issues but where I feel those aren't worth arguing over (often it's how you split things that I really think should be one commit across two commits -- you and I simply disagree about the value (or lack thereof) of having all changes in a commit restricted to one directory :-)

I will go back over the remaining open ones that aren't marked as "don't merge" to see if I missed anything, and to remove any comments that may come across as rude.

mobile-widgets/qml/main.qml
dirkhh: Yes, "fundamental" is over the top, agreed.

I think your argument here for the behavior change has merit.
As you refactor this (I hope, see my other comments), would you add a shortened version of this to the relevant commit message so that this argument is maintained in the git record?
Thanks

mobile-widgets/qml/main.qml
dirkhh: I think this change was in one of the other PRs that I just merged. Or am I starting to confuse myself?

janiversen:
point taken, it actually came across very differently (as something personal), but let history be history.

dirkhh (commented):
OK, updated / added some comments.

I would appreciate some rework based on the updated comments. Especially capturing the rationale in one of the commit messages would be useful. And maybe splitting into smaller PRs if that isn't too much work. And of course it needs to be rebased :-)

dirkhh:
> point taken, it actually came across very differently (as something personal), but let history be history.

If at all possible (and I know this is hard), can you please assume that my comments are always written with a focus on the software and on the process, an NEVER, EVER about the person.

janiversen:
I will try my best but I am a burned child (not sure if this expression translated well into US version of the English language).

I hope that slowly you get more trust in my commits (although I still like a thorough review to point out the silly errors). I have worked a year together with my local club to create a (in our opinion) highly more modern version of mobile, and I want to share our work instead of keeping it local.

btw. should you ever come close to Granada, Spain you have a standing invitation of some interesting dives (and local dinner).

mobile-widgets/qml/main.qml
janiversen: tomorrow

janiversen:
closed for rework

PR #2453, opened 2019-12-22
clean CMake for mobile

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

1) corrected potential confusing spelling error in build.sh 
2) Added CMakeList.txt files to root directories that did not have one.

This PR makes the use of CMakeLists.txt consistent (each root directory have its own
CMakeLists.txt) and at the same time cleans the central CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -279,6 +279,7 @@ add_subdirectory(qt-models)
 add_subdirectory(commands)
 add_subdirectory(profile-widget)
 add_subdirectory(map-widget)
+add_subdirectory(mobile-widgets)
 
 if (SUBSURFACE_TARGET_EXECUTABLE MATCHES "DesktopExecutable")
 	add_subdirectory(desktop-widgets)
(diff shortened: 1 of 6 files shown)


scripts/build.sh
```diff
@@ -12,7 +12,7 @@
 # in subsurface/build)
 #
 # by default it puts the build folders in
-# ./subsurface/llibdivecomputer/build (libdivecomputer build)
+# ./subsurface/libdivecomputer/build (libdivecomputer build)
```
dirkhh, line 15: not sure this qualifies as "disturbing" ("causing anxiety; worrying"), but yeah, typo.

CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 # cmake based build of Subsurface
 
+
```
dirkhh: why are we adding another new line here? That seems like leftover from some other changes maybe?

map-widget/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# map widget to show locations
+
+# library used by mobile build
+set(SUBSURFACE_MAPWIDGET_SRCS
+	qmlmapwidgethelper.cpp
+)
```
dirkhh, line 6: This is truly just a question... but... is it worth adding this for one file? In other words, besides logical consistency and what not, what do we gain by doing this?

desktop-widgets/CMakeLists.txt
```diff
@@ -55,7 +55,6 @@ set (SUBSURFACE_UI
 
 # the interface, in C++
 set(SUBSURFACE_INTERFACE
-	../map-widget/qmlmapwidgethelper.cpp
```
dirkhh, line 58: shouldn't this be part of the last commit? Aren't we getting a compile error otherwise (haven't tried)

dirkhh (approved):
I understand that this is cleaner. I'm not sure that's where I would have focused my time.
It doesn't seem to break anything, so "approved"...

scripts/build.sh
janiversen: updated text in PR (it is not present in the commit text).

CMakeLists.txt
janiversen: Corrected.

map-widget/CMakeLists.txt
janiversen: That argument is valid for all CMakeList.txt files in subdirectories. Why not just make one big central file (e.g. mobile-widgets, contained a lot of entries in the central file).

We do not gain performance or anything, but consistency means easier to understand and thereby easier to maintain (at least for me). The other gain is that each root directory is a library.

And I "believe" that in a short future, there will be a PR with a dynamic map that allows the user to change locations (at least the current experiments looks good).

desktop-widgets/CMakeLists.txt
janiversen: I build and test every single commit, so there are no compile errors and it works.

I have no idea how CMake deals with the conflict, but please remember the old central CMake reference qmlmapwidgethelper.cpp, so this problem have existed for a long time.

dirkhh (approved):

PR #2454, opened 2019-12-22
activate qml compiler for iOS.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR activates the compiler for iOS, and should not have affect on other platforms.

eval() is not allowed in qml so replaced all references.

### DIFF
--- a/mobile-widgets/qml/DownloadFromDiveComputer.qml
+++ b/mobile-widgets/qml/DownloadFromDiveComputer.qml
@@ -162,21 +162,47 @@ Kirigami.Page {
 					elide: Text.ElideRight
 				}
 				onCurrentTextChanged: {
+					var curVendor
+					var curProduct
+					var curDevice
 					dc1.enabled = dc2.enabled = dc3.enabled = dc4.enabled = true
 					for (var i = 1; i < 5; i++) {
+						switch (i) {
+							case 1:
+								curVendor = PrefDiveComputer.vendor1
+								curProduct = PrefDiveComputer.product1
+								curDevice = PrefDiveComputer.device1
+								break
+							case 2:
+								curVendor = PrefDiveComputer.vendor2
+								curProduct = PrefDiveComputer.product2
+								curDevice = PrefDiveComputer.device2
+								break
+							case 3:
+								curVendor = PrefDiveComputer.vendor3
+								curProduct = PrefDiveComputer.product3
+								curDevice = PrefDiveComputer.device3
+								break
+							case 4:
+								curVendor = PrefDiveComputer.vendor4
+								curProduct = PrefDiveComputer.product4
+								curDevice = PrefDiveComputer.device4
+								break
+						}
+
 						if (comboProduct.currentIndex === -1 && currentText === "FTDI"){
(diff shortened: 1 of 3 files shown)


dirkhh (commented):
This looks very straight forward.
I am way, way, WAY behind on reviewing pull requests.
There are quite a few pending and I don't think I'll get to them today - I just picked this one since it's something that I had been wanting to play with for a long time...
I /should/ have a few hours tomorrow to focus on Subsurface. If there is a preferred order to look at your requests and test them, please let me know.

dirkhh:
Did a quick test (actually just dealing with getting my iOS build chain back to work - why does Apple break my setup about once a month???) and this seems to work well - have you done any benchmarks / size comparisons (not required, just curious). Also, do you happen to know how to enable this on the cmake side so we'd have this on Android as well?

janiversen:
I think the easiest way is to review in PR order, that way all PR´s could rebase automatically. Right now the dive planner is where I have the biggest queue pending, which happens to be the oldest PRs.

I do have an a solution for CMake, which is on my pending list to make available (low priority, since it does not change any user experience), which targets is to have CMake generate iOS as well (adding 3 extra targets, 1 pr architecture). The solution I have works for generating the libraries, but it has a manual step for linking/adding resources.

The size change (simulator build) is less than I expected between 4-5Mb out of 41Mb, looking at the qmlc files (see e.g. in tests)
-rw-r--r--  1 jan  staff   4,1K 12 sep  2018 tst_qPrefGeneral.qml
-rw-r--r--  1 jan  staff   8,4K  2 feb  2019 tst_qPrefGeneral.qmlc
I expected more.

Performance:
without bluetooth, the startup is faster (the load call is reduced to 1/3), however independent of this bluetooth is causing a 2-3 second black screen on some iPhone (like my 6s).
Divelist manipulation is a lot more fluid, especially on my overloaded 6s.

Talking about breaking tool chains, it seems your latest libdivecomputer update called for a new version of Automake (min. 1.6 required). Causing some confusion in our builds.

mobile-widgets/qml/DownloadFromDiveComputer.qml
```diff
@@ -162,21 +162,47 @@ Kirigami.Page {
...
+								curProduct = PrefDiveComputer.product3
+								curDevice = PrefDiveComputer.device3
+								break
+							case 4:
+								curVendor = PrefDiveComputer.vendor4
+								curProduct = PrefDiveComputer.product4
+								curDevice = PrefDiveComputer.device4
+								break
+						}
+
```
janiversen, line 192: Having read especially one of your comments (in another PR), I see the need to explain this change. I could have choosen to make "yet another round of ......", by making a changing vendor etc in qPrefDiveComputer to a qStringList, which would have been the pure solution,  but I choose NOT to do that and instead made a more ugly implementation.

I hope this demonstrates, that I do NOT make changes for reason of purity or just because I have a firm opinion, I make changes because it moves subsurface-mobile forward in a positive manner.

PR #2455, opened 2019-12-23
plannerShared: add remaining planning variables

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add the last planning variables to plannerShared and update desktop-widgets.


dirkhh:
So this removed the do-not-merge label.
Once the conflicts are resolved I'll do a review 

And I totally agree with your comments in one of the many PRs how frustrating it is to split things into smaller PRs that then cause these conflicts -- yet at the same time this makes progress on the "easy" ones so much faster and you don't have to carry around all these commits which is off-putting to reviewers.
Quite literally, I often find myself with "I have 15 minutes to look at something". A PR with three or four small commits that address exactly one thing for an obvious or well explained reason is really easy to look at, think about, approve, and merge.
But a PR with 30+ long commits... I won't even start reading this if all I have is a few minutes.

So in the end the process that I suggest is optimizing for the scarce resource (my time) and tries to make sure that I am less likely to become a blocker (because many of the other maintainers might merge a small, obvious change - but typically they leave the more convoluted things to me)

janiversen:
Agreed, but in respect of mobile, I am like to have the challenges of the very convoluted items :-)

janiversen:
resubmitting in smaller chunks

PR #2460, opened 2019-12-23
(Trivial?) cleanup: Turn QMLProfile::diveId from QString to int

bstoeger:
This property is used to render the profile of a given dive.
Weirdly, even though the diveId is an integer, it was stored
as a string. It is not clear why that is the case. Therefore,
turn into the more natural int and avoid unnecessary conversion.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is what appears to be an obvious tiny cleanup.

### Changes made:

1) Turn a `QString` that represents an `int` into an `int`

### Mentions:

@dirkhh: Perhaps I'm missing something obvious? In a first test things still seemed to work.

[Side note: in the same test it became obvious that the current mobile-list in master does rather funky things when trips overlap - but this is not related to this PR].

### DIFF
--- a/profile-widget/qmlprofile.cpp
+++ b/profile-widget/qmlprofile.cpp
@@ -80,17 +80,17 @@ void QMLProfile::setMargin(int margin)
 	m_margin = margin;
 }
 
-QString QMLProfile::diveId() const
+int QMLProfile::diveId() const
 {
 	return m_diveId;
 }
 
-void QMLProfile::setDiveId(const QString &diveId)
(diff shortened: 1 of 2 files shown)


janiversen:
Works for me.

dirkhh (approved):
Thanks. Great, logical, concise PR with obvious value.

PR #2464, opened 2019-12-24
(planner 1) include planner in mobile build

janiversen:
### Describe the pull request:

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Added a #ifdef because commands are not available for mobile. It would be nice if there was a guideline saying new "core" functionality should be available for all platforms (e.g. commands and diveFilter).

Added needed models etc. to include plannerModel in mobile build.

PR #2465, opened 2019-12-24
[do not merge, need rebase] (planner 2) add planning variables to plannerShared

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

** Warning  ** this pr is based on #2464 and thus contains those commits

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "plannershared.h"
-#include "core/pref.h"
 #include "core/settings/qPrefDivePlanner.h"
-
+#include "core/settings/qPrefTechnicalDetails.h"
+#include "qt-models/diveplannermodel.h"
 
 plannerShared *plannerShared::instance()
 {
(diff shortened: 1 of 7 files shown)


backend-shared/plannershared.cpp
```diff
@@ -113,3 +113,13 @@ void plannerShared::set_vpmb_conservatism(int value)
...
 }
+
+bool plannerShared::dobailout()
+{
+	return qPrefDivePlanner::dobailout();
+}
+void plannerShared::set_dobailout(bool value)
+{
+	qPrefDivePlanner::set_dobailout(value);
+	DivePlannerPointsModel::instance()->emitDataChanged();
```
dirkhh, line 124: so why is this done differently from the cases in the previous commit, where you have this asymmetric pattern of reading from qPref and calling the setter in `DivePlannerPointsModel` ?
Calling that signal in the unrelated class seems odd. But maybe there's a good reason to do it this way that I just don't see?

dirkhh (commented):
one small comment about something that looks like an inconsistency to me.
overall I'd prefer a symmetric setup where we do both setters and getters through the same class, even if that causes an extra function call / class dereference.

backend-shared/plannershared.cpp
janiversen: I too find it odd, but desktop-widgets/diveplanner.cpp have a number of case like this:

before:
void PlannerSettingsWidget::setBackgasBreaks(bool dobreaks)
 {
       qPrefDivePlanner::instance()->set_doo2breaks(dobreaks);
       plannerModel->emitDataChanged();
 }
after:
void PlannerSettingsWidget::setBackgasBreaks(bool dobreaks)
 {
       plannerShared::set_doo2breaks(dobreaks);
 }

Where you see I move the functionality 1-1 to backend-shared.

Actually I do not know if there is a good reason.

dirkhh: That's a good explanation. @atdotde @bstoeger - any insights into why we do things that way?
(I think Berthold is offline, now that I wrote this)

dirkhh:
manually merged

PR #2466, opened 2019-12-24
[do not merge, need rebase] (planner 3) add last planning variables

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

** Warning ** this pr is based on #2465 and thus contains those commits

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -150,3 +150,13 @@ void plannerShared::set_switch_at_req_stop(bool value)
 {
 	DivePlannerPointsModel::instance()->setSwitchAtReqStop(value);
 }
+
+bool plannerShared::doo2breaks()
+{
+	return qPrefDivePlanner::doo2breaks();
+}
+void plannerShared::set_doo2breaks(bool value)
+{
+	qPrefDivePlanner::set_doo2breaks(value);
+	DivePlannerPointsModel::instance()->emitDataChanged();
+}
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "plannershared.h"
-#include "core/pref.h"
 #include "core/settings/qPrefDivePlanner.h"
-
+#include "core/settings/qPrefTechnicalDetails.h"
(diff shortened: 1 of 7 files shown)


dirkhh (commented):
similar comment as before.
This seems like we are moving things around in odd and not particularly obvious ways.
If those variables are already in a different model, why do we need all these "proxy" getters and setters? Shouldn't we just access that different model? And if that's the wrong answer, then the explanation should be in the commit message.
The way this reads to me it's just not clear why we need all of these in the plannerShared class instead of just going to the class we get them from.

janiversen:
Because I am trying to make a backend that hides the complexity of multiple models etc, that way the different models to not need to interface QML. The UI should not need to know how the classes are structured internally, the UI should have a simple consistent interface.

But I see what you mean, and following your argument through, there are no real reason for this backend (but please do not complain, when the QML interface gets even more complicated).

dirkhh:
I don't think that was what I was suggesting.

dirkhh:
I like the idea of having a shared backend. And having an structured, sane way to access things from QML.
My comments were about the code that ended up in that backend.
And if the answer is "this is done like that on the desktop side" then that's, err, odd, but clearly no longer an issue that you necessarily need to fix in this PR.
Let me finish my work on the other PR and I'll get back to this series and poke more at the desktop code to get myself comfortable with your solution.
So let's put this on hold (hopefully just an hour or three) and ignore my comments for now

janiversen:
quote "And if the answer is "this is done like that on the desktop side" then that's, err, odd, but clearly no longer an issue that you necessarily need to fix in this PR."

yes the functions are 1-1 copies from the desktop code. A couple of function have had code copied from a couple of places in the desktop-widget. The only modification is pure naming like plannerModel-> (seeing it now I could of course have made a class variable with the same name).

dirkhh:
... and power went out at my house.
Sending this from my phone so you don't wonder why I stopped working on the PRs
-- 
From my phone

janiversen:
Christmas is time for candle light! and no I did not even notice nor am I assuming about how and when you work.

I would still like to know, how I can do branches on branches without the second branch including the commits from the first one when I push it.

dirkhh:
manually merged

PR #2467, opened 2019-12-24
[do not merge, need rebase] (planner 4) add gas variables

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

** Warning ** this pr is based on #2466 and thus contains those commits

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -150,3 +150,13 @@ void plannerShared::set_switch_at_req_stop(bool value)
 {
 	DivePlannerPointsModel::instance()->setSwitchAtReqStop(value);
 }
+
+bool plannerShared::doo2breaks()
+{
+	return qPrefDivePlanner::doo2breaks();
+}
+void plannerShared::set_doo2breaks(bool value)
+{
+	qPrefDivePlanner::set_doo2breaks(value);
+	DivePlannerPointsModel::instance()->emitDataChanged();
+}
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -59,6 +59,7 @@ void plannerShared::set_descrate(int value)
 	qPrefDivePlanner::set_descrate(value * TO_MM_BY_SEC);
 }
 
+
 // Planning values
 deco_mode plannerShared::planner_deco_mode()
 {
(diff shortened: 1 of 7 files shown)


backend-shared/plannershared.cpp
```diff
@@ -59,6 +59,7 @@ void plannerShared::set_descrate(int value)
 	qPrefDivePlanner::set_descrate(value * TO_MM_BY_SEC);
 }
 
+
```
dirkhh, line 63: second new line?

backend-shared/plannershared.cpp
```diff
@@ -170,3 +171,43 @@ void plannerShared::set_min_switch_duration(int value)
...
+double plannerShared::bottomsac()
+{
+	return qPrefDivePlanner::bottomsac() / 1000.0;
+}
+void plannerShared::set_bottomsac(double value)
+{
+	// NO conversion, this is done in the planner model.
+	DivePlannerPointsModel::instance()->setBottomSac(value);
+}
```
dirkhh, line 184: same asymetric setup - only worse with the factor of 1000 being hidden somewhere I can't see reading the commit.

backend-shared/plannershared.cpp
```diff
@@ -211,3 +212,15 @@ void plannerShared::set_sacfactor(double value)
...
+
+bool plannerShared::o2narcotic()
+{
+	return qPrefDivePlanner::o2narcotic();
+}
+void plannerShared::set_o2narcotic(bool value)
+{
+	qPrefDivePlanner::set_o2narcotic(value);
+	DivePlannerPointsModel::instance()->emitDataChanged();
+	CylindersModel::instance()->updateBestMixes();
```
dirkhh, line 224: just reading this... we call a setter in `plannerShared` which calls a setter in `qPrefDivePlanner`, emits a signal from `DivePlannerPointsModel` and then calls a random update function in `CylinderModel`.

That looks like a complete mess, not like a clean design (and use) of those classes.
Anyone trying to debug this in a year will be thoroughly confused, I think.

dirkhh (commented):
I continue to worry about the approach here. This is crazy convoluted.

dirkhh:
manually merged

PR #2469, opened 2019-12-24
Desktop: Import dive coordinates directly from GPS

willemferguson:
This allows Subsurface to obtain the coordinates of a dive directly
from a GPS track. It parses a GPX file (GPX V1.0 or V1.1) from
a GPS to locate the trackpoint immediatedly after the start of a
dive. There is an additional "Use GPS file" button in the Edit Dive
Site panel that is selected from the Notes tab. Image:
Dive location edit screen

This allows one to select a GPX file, bringing up the Import GPS
dialog. Image:
GPS synchronisation dialog

There is extensive provision for cross-checking that the dive track
synchronises with the dive start and end. If the Save button in the
dialog is pressed the dive coordinates are copied into the Dive
Coordinates text box in the Edit Dive Site panel. The map moves
to indicate the location of the dive site.

The bulk of the work is done in importgps.cpp. The code is
pretty intergrated: I tried to break it up in smaller commits but that
was not feasible.

The code was tested on GPX generated by using a Garmin Etrix 30 
GPS as well as by two Android apps.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @bstoeger @atdotde


atdotde:
Without looking at a single line of code yet: You could provide a sample kml file with a matching dive in the dives directory.

willemferguson:
Attached a single-dive log and a GPX file. GPX is, as far as I can gather, more or less the standard format of geographers and outdoor recreational GPS machines. KML is the Google-specific format that is additionally supported by a few GPS machines. I have not seen a GPS machine that uses KML as a a primary file storage format.
[gpstest.zip](https://github.com/Subsurface-divelog/subsurface/files/3999014/gpstest.zip)
This is a freshwater dive with the GPS located closeby on land, and not on the water!

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+	connect(ui.diff_forward, SIGNAL(toggled(bool)), this, SLOT(changeDiffForward()));
+	connect(ui.GPSbuttonBox, SIGNAL(clicked(QAbstractButton *)), this, SLOT(buttonClicked(QAbstractButton *)));
+	coords.settingsDiff_offset = 0;
+	coords.timeZone_offset = 0;
+	coords.lat = 0;
+	coords.lon = 0;
+	pixmapSize = (int) (ui.diveDateLabel->height() / 2);
+}
+
+void  ImportGPS::buttonClicked(QAbstractButton *button)
```
neolit123: extra space after `void`

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+		LocationUI->diveSiteCoordinates->setText(QString::number(coords.lat) + ", " + QString::number(coords.lon));
+		LocationUI->diveSiteCoordinates->editingFinished();
+	} else {
+		close();
+	}
+}
+
+// Extract text from the present position in the file until
+//  the first 'delim' character is encountered.
+int ImportGPS::get_substring(FILE *f, char *buf, char delim)
```
neolit123: is this a naming convention that has settled?
should all C++ member functions be `pascalCase` e.g. `ImportGPS::getSubstring`?

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+	}
+	buf[j] = 0x0; // terminate the string
+	return 0;
+}
+
+// Find the next occurence of a specified GPX element in the file,
+// characerised by a "<xxx " or "<xxx>" character sequence.
+// s specifies the name of the element searched for.
+// termc is the ending character of the element name search = '>' or ' '.
+int ImportGPS::find_xml_element(FILE *f, const char *s, char *buf, char termc)
```
neolit123: instead of more complex C logic it might be better to leverage Qt's utilities such as QChar.
this is not in the C backend and performance of QChar should be pretty good.

this is not strictly a blocker, but makes the CPP code more readable instead of the C / C++ mixture within a single CPP file.

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+// Find the coordinates at the time specified in coords.start_dive
+// by searching the gpx file "fileName"
+int ImportGPS::getCoordsFromFile() 
+{
+	struct tm tm1;
+	time_t when = 0;
+	double lon, lat;
+	char buf[BUFFER_SIZE];
+	char strbuf[50];
+	int line=0;
```
neolit123: missing space around `=`

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+				break;
+		}
+		buf[j] = 0x0;	// terminate the xml element name
+		if (!(strcmp(buf, "/trk"))) // End of GPS track found: return EOF
+			return 1;
+		if (c == skipdelim) continue; // if wrong delimiter for this element was found
+		if (strcmp(s, buf))       // Compare xml element name from gpx file with the
+			match = false;    // the element searched for.
+		else
+			match = true;
```
neolit123: can be `match = strcmp(s, buf) ? true : false`

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+	double lon, lat;
+	char buf[BUFFER_SIZE];
+	char strbuf[50];
+	int line=0;
+	int64_t time_offset = coords.settingsDiff_offset + coords.timeZone_offset;
+	time_t divetime;
+	bool first_line = true;
+	bool found = false;
+	QByteArray local8bitBAString1 = fileName.toLocal8Bit(); // Do these stupid gymnastics to
+	char * fname = local8bitBAString1.data();               // convert QString to a C string filename
```
neolit123: `char *fname`

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+	divetime = coords.start_dive;
+	FILE *f1;
+	if (!(f1 = fopen(fname, "r"))) {
+		fprintf(stderr, "File open error %s\n", fname);
+		return 1;
+		}
```
neolit123: `}` should be indented below `if`

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+	utc_mkdate(divetime, &time);
+	int dyr,dmon,dday,dhr,dmin;
+	dyr = time.tm_year;
+	dmon = time.tm_mon;
+	dday = time.tm_mday;
+	dhr = time.tm_hour;
+	dmin = time.tm_min;
+#endif
+
+	do {
```
neolit123, line 130: similar comment here, the buffer can be processed using Qt constructs.

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+	ui.diveDateLabel->setText("Dive date = " + QString(datestr) + QString::number(time.tm_year));
+	ui.startTimeSyncLabel->setText( QString::number(time.tm_hour) + ":" + QString::number(time.tm_min)); // dive start time
+	utc_mkdate(coords.end_dive, &time);
+	ui.endTimeSyncLabel->setText(QString::number(time.tm_hour) + ":" + QString::number(time.tm_min));  // dive end time
+
+	// This section implements extensive warnings in case there is not synchronisation between dive and GPS data:
+	if (gps_day != dive_day)
+		problemString = "(different dates)";
+	else 
+		problemString = "";
```
neolit123: can be `problemString = (gps_day != dive_day) ?  .....`

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+	ui.badIconLabel->setPixmap(badResultIcon.scaled(pixmapSize,pixmapSize,Qt::KeepAspectRatio));
+	ui.badIconLabel->setVisible(false);
+	// Show information or warning message as well as synch quality icon
+	if (coords.start_dive < coords.start_track) {
+		ui.resultLabel->setStyleSheet("QLabel { color: red;} ");
+		ui.resultLabel->setText("PROBLEM: Dive started before the GPS track "+ problemString);
+		ui.badIconLabel->setVisible(true);
+	} else {
+		if (coords.start_dive > coords.end_track) {
+		ui.resultLabel->setStyleSheet("QLabel { color: red;} ");
```
neolit123: should be indented to the right.

desktop-widgets/locationinformation.cpp
```diff
@@ -219,6 +221,21 @@ void LocationInformationWidget::initFields(dive_site *ds)
 	}
 }
 
+
+void LocationInformationWidget::on_GPSbutton_clicked()
+{
+	QString fileName = QFileDialog::getOpenFileName(  this, tr("Select GPS file to open"),
```
neolit123: `getOpenFileName(  this...` -> `getOpenFileName(this...`

neolit123:
@willemferguson 
i've added some general coding style / C vs C++ comments. i have not tested the code.

desktop-widgets/importgps.cpp
bstoeger: Or, perhaps even simpler remove the else branch. `problemString` was initialized to the empty string anyway, no?

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+#include "core/dive.h"
+
+#include <time.h>
+#include <QFileDialog>
+#include <QProcess>
+#include <QDateTime>
+#include <QDate>
+#include <stdlib.h>
+#include <stdio.h>
+#include <QDebug>
```
bstoeger: I'm always a bit weary about QDebug includes. Is it needed?

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+   3) Allow modification of the coordinates dependent on international time zone and
+              on differences in local time settings between the GPS and the dive computer.
+   4) Saving the coordinates into the Coordinates text box in the Dive Site Edit panel
+              and which which causes the map to show the location of the dive site. 
+   The structure coords is used to store critical information.  */
+ImportGPS::ImportGPS(QWidget *parent, QString fileName, class Ui::LocationInformation *LocationUI) : QDialog(parent),
+	fileName(fileName), LocationUI(LocationUI)
+{
+	ui.setupUi(this);
+	connect(ui.timeDiffEdit, SIGNAL(timeChanged(const QTime)), this, SLOT(timeDiffEditChanged()));
```
bstoeger, line 25: I would so much prefer the pointer-to-member version, because it gives compile-time type checking. Though we can do that later.

desktop-widgets/importgps.cpp
bstoeger: Where's the user of that function?
In C++ code one might use QString or std::string instead of fixed-size buffers and all that brittle stuff.

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,329 @@
...
+			match = false;    // the element searched for.
+		else
+			match = true;
+	} while (match == false);
+	return 0;
+}
+
+// Extract characters in src from first to last position 
+// e.g. substr("ascdef",dest,3,5) yields dest = "cde".
+char* ImportGPS::substr(const char *src, char *dest, int first, int last)
```
bstoeger: Per CodingStyle, we put the `*` to the object, not the type. I.e. `char *ImportGPS...`

willemferguson: Done

desktop-widgets/importgps.cpp
willemferguson: Please give a bit more detailed information. I would like to correct this now.

desktop-widgets/importgps.cpp
willemferguson: done.

desktop-widgets/locationinformation.cpp
willemferguson: done.

desktop-widgets/importgps.cpp
willemferguson: done.

desktop-widgets/importgps.cpp
willemferguson: done.

desktop-widgets/importgps.cpp
willemferguson: Done.

desktop-widgets/importgps.cpp
willemferguson: Done.

desktop-widgets/importgps.cpp
willemferguson: Done

desktop-widgets/importgps.cpp
willemferguson: @bstoeger's comment taken for this one. The else part has been deleted now.

desktop-widgets/importgps.cpp
willemferguson: Done

desktop-widgets/importgps.cpp
willemferguson: Done. Qt constructs were used where possible.

desktop-widgets/importgps.cpp
willemferguson: Buffer converted to QString.

desktop-widgets/importgps.cpp
willemferguson: Done.

willemferguson:
I am not sure why the divelistview.cpp & divelistview.h files are included in this push now. They were not listed as tracked or changed in the commit.

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,308 @@
...
+#include "desktop-widgets/locationinformation.h"
+#include "core/dive.h"
+
+#include <time.h>
+#include <QFileDialog>
+#include <QProcess>
+#include <QDateTime>
+#include <QDate>
+#include <stdlib.h>
+#include <stdio.h>
```
neolit123: can we group the imports a little:

perhaps Q* imports at the bottom and C time/std* imports separated 

e.g.
```
#include "ui_importgps.h"
#include "desktop-widgets/importgps.h"
#include "desktop-widgets/locationinformation.h"
#include "core/dive.h"

#include <time.h>
#include <stdlib.h>
#include <stdio.h>

#include <QFileDialog>
#include <QProcess>
#include <QDateTime>
#include <QDate>
```

also aren't these imported from dive.h already?
```
#include <time.h>
#include <stdlib.h>
#include <stdio.h>
```

in general, if we can get away with not using C methods in this CPP file it would be beter.

desktop-widgets/locationinformation.cpp
```diff
@@ -219,6 +220,21 @@ void LocationInformationWidget::initFields(dive_site *ds)
 	}
 }
 
+
+void LocationInformationWidget::on_GPSbutton_clicked()
+{
+	QString fileName = QFileDialog::getOpenFileName(this, tr("Select GPS file to open"),
+								tr("/home/willem/Desktop"),
```
neolit123: we shouldn't hardcode this directory. we should use the subsurface "application directory".

neolit123:
@willemferguson i saw a big comment with QFile / import related questions in my email notification, but it's not here any more. possibly you resolved the issue.

added a couple of more comments, but this is looking better.

desktop-widgets/locationinformation.cpp
willemferguson: You are absolutely correct. The application directory may differ between versions of subsurface. Specifically, in the Linux release versions (e.g. Ubuntu repository) the application directory may be in the system area of Linux (e.g. in the snap directory of Ubuntu). The safest option is probably the home directory. This is what I did.

desktop-widgets/importgps.cpp
willemferguson: Done

desktop-widgets/locationinformation.cpp
willemferguson: Another solution that is OS-independent would be to use the path provided by system_default_filename(). This would normally point to the directory containing the Subsurface settings.

neolit123: this should use `QString(system_default_directory())`
`/home` is not portable.

also no need to `tr(..)` the path.

willemferguson: I did exactly that in my latest commit.

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,302 @@
...
+	time_t divetime;
+	bool first_line = true;
+	bool found = false;
+	divetime = coords.start_dive;
+	QString buf;
+	QFile f1;
+	f1.setFileName(fileName);
+	if (!f1.open(QIODevice::ReadOnly | QIODevice::Text)) {
+		QByteArray local8bitBAString1 = fileName.toLocal8Bit();
+		char * fname = local8bitBAString1.data();   // convert QString to a C string filename
```
neolit123: `char * fname` -> `char *fname`

willemferguson: Corrected

dirkhh:
So what's the status of this? I see quite a bit of discussion, but I'm not sure what to do...
@bstoeger @neolit123 - are your review concerns addressed?
@willemferguson is the the version you want me to rebase and review?
I'm sorry to have dropped the ball here... I got a little burned out on reviews last week and decided to use my Subsurface time to write code instead...

desktop-widgets/importgps.cpp
dirkhh: there are two different ways to do connect calls. the one you have here with the string based syntax, and the "modern" one which uses type checking at compile time. The difference is well explained in the Qt Wiki: https://wiki.qt.io/New_Signal_Slot_Syntax

desktop-widgets/divelistview.cpp
```diff
@@ -49,9 +49,6 @@ DiveListView::DiveListView(QWidget *parent) : QTreeView(parent),
 
 	resetModel();
 
-	// Update selection if all selected dives were hidden by filter
-	connect(&diveListNotifier, &DiveListNotifier::filterReset, this, &DiveListView::filterFinished);
-
```
dirkhh, line 54: why is this removed?

desktop-widgets/divelistview.cpp
```diff
@@ -273,26 +270,6 @@ void DiveListView::selectTrip(dive_trip_t *trip)
...
-	auto marker = diveListNotifier.enterCommand();
-
-	// we want to make sure no trips are selected
-	Q_FOREACH (const QModelIndex &index, selectionModel()->selectedRows()) {
-		if (!index.data(DiveTripModelBase::IS_TRIP_ROLE).toBool())
-			continue;
-		selectionModel()->select(index, QItemSelectionModel::Deselect);
-	}
-}
-
```
dirkhh, line 295: similarly. why is this removed?

desktop-widgets/divelistview.cpp
```diff
@@ -1019,19 +996,6 @@ void DiveListView::loadImageFromURL(QUrl url)
...
-	// (but without updating the selection state of the dives... this just cleans
-	// up an oddity in the filter handling)
-	clearTripSelection();
-
-	// If there are no more selected dives, select the first visible dive
-	if (!selectionModel()->hasSelection())
-		select_newest_visible_dive();
-	emit divesSelected();
-}
-
```
dirkhh, line 1034: ditto

desktop-widgets/locationinformation.cpp
dirkhh: so I need to combine those commits. No problem.

desktop-widgets/locationinformation.cpp
```diff
@@ -223,8 +223,9 @@ void LocationInformationWidget::initFields(dive_site *ds)
 
 void LocationInformationWidget::on_GPSbutton_clicked()
 {
+	QFileInfo finfo(system_default_filename());
 	QString fileName = QFileDialog::getOpenFileName(this, tr("Select GPS file to open"),
-								tr("/home"),
+								finfo.absolutePath(),
```
dirkhh, line 228: so this will default to the `username.xml` filename we use as default for the data file. I think that's a bit confusing.

dirkhh (commented):
I did a quick run through. Clearly these commits need to be combined - I can easily do this. But I do have a few questions.

neolit123:
> @neolit123 - are your review concerns addressed?

i had some C/C++ style concerns, which seem resolved.
but no real context if this feature is desired.

dirkhh:
> ...but no real context if this feature is desired.

Oh, very very much desired! This is something that we have been asked to implement several times.

desktop-widgets/locationinformation.cpp
willemferguson: I have completed all updates that I can see.  Yes, this points to the default folder containing the dive log. This is the only remaining issue. What default directory to use in a platform-independent way?

desktop-widgets/divelistview.cpp
willemferguson: This is a gremlin. I have no idea how this became part of the PR. Should I submit a commit that undoes these three changes?

dirkhh: I can take care of that when merging.

dirkhh:
moved this to a new PR. That needs review and more checking.
Still uses the old connect syntax. I ran out of time

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,302 @@
...
+		bufptr->append(QChar(c));
+	} while (c != delim);
+	return 0;
+}
+
+// Find the next occurence of a specified target GPX element in the file,
+// characerised by a "<xxx " or "<xxx>" character sequence.
+// 'target' specifies the name of the element searched for.
+// termc is the ending character of the element name search = '>' or ' '.
+int ImportGPS::findXmlElement(QFile *fileptr, QString target, QString *bufptr, char termc)
```
buxtonpaul, line 69: Hmm.
Why are we manually parsing XML? 
Our CMake file looks like we link against libxml2.0 which suggests we should be able to parse the track with that?

buxtonpaul (commented):
I had considered looking at using GPX file as a source of both location information as well as potentially heart rate data, as the type of watch which records GPX data is often able to also track heart rate, and can export the data in the GPX file using an extension to the schema.
Not sure if this is likely to work underwater though...

desktop-widgets/importgps.cpp
willemferguson: In fact, the dive log files in xml are parsed using a much more powerful framework than this gpx interpretation. However, when extracting a small amount of very specific information from xml it is, in my opinion, overkill to use one of the larger xml parsing frameworks. The existing code is pretty compact.

willemferguson:
This is off-topic, but heart rate is normally dealt with as part of an existing dive log, not as a separate data type (at least this has been the experience to date). The hardware works underwater using the regular Polar chest strap and Scubapro (incorporating heart rate into deco calculations) has been using this approach. Heart rate data only make sense within the context of a dive profile (i.e. the dive log)

desktop-widgets/importgps.cpp
buxtonpaul: Agree overkill if we were including it just for this. But if the library is already there why not just use it?

dirkhh:
> I had considered looking at using GPX file as a source of both location information as well as potentially heart rate data, as the type of watch which records GPX data is often able to also track heart rate, and can export the data in the GPX file using an extension to the schema.
> Not sure if this is likely to work underwater though...

We have a number of supported dive computers that record heart rate data, e.g. the Garmin Descent Mk1. I can't speak to the accuracy of that data, but it does "work" :-)

desktop-widgets/importgps.cpp
dirkhh: I'll let @willemferguson speak for himself - for me it's often that simply parsing XML as strings seems "approachable" and dealing with all the machinations of these frameworks is often painful.

@buxtonpaul are you offering to convert Willem's code to using either libxml2 or the Qt XML functions? I don't want to make this a requirement for Willem before his code is merged...

@willemferguson one thing I really would like to have (and this __should__ be a separate commit) is a dive log XML and corresponding `.gpx` file so that I am able to actually test this :-)

And of course, this code should have been structured in a way that it is clearly separated into core and UI so that it can be called from mobile as well (another thing that I should have looked for earlier... I'm not doing a good job as maintainer these days, I'm afraid). So I guess I am offering to take care of that... :-(

dirkhh:
Oh, and reminder - we should please base any future work on #2545 where I rebased, squashed, and partially improved this code on top of a much more recent master

willemferguson:
will do

desktop-widgets/importgps.cpp
willemferguson: Keeping Core and UI separate:
As is, the core and UI are reasonable separate. I created the coords structure with the aim to keep the core separate; otherwise the members of coords could have been defined within the object definition in the header file. BUT, some of the values in the coords structure are obtained from the UI.

willemferguson:
Please remove this PR?? Already merged to master.

dirkhh:
you should be able to close your own pull requests, but as maintainer I can close any PR, so I'll just close this...

PR #2470, opened 2019-12-24
build system: do not name libraries <name> _desktop and _mobile

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

No need to name libraries <name>_desktop and <name>_mobile, they
are stored in 2 different trees (build and build-mobile).

Having different names for the same library, just adds need for more
ifdef in the different CMakeList.txt files

If there are reasons to make the libraries different, just close this PR and e.g. tests/CMakeLists.txt will need more ifdef.

Generally we should move in a direction where all libraries (apart from desktop-widgets and mobile-widgets) are shared code between desktop and mobile.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,7 +313,7 @@ if (SUBSURFACE_TARGET_EXECUTABLE MATCHES "MobileExecutable")
 		subsurface_profile
 		subsurface_mapwidget
 		subsurface_backend_shared
-		subsurface_models_mobile
+		subsurface_models
 		subsurface_corelib
 		${SUBSURFACE_LINK_LIBRARIES}
 	)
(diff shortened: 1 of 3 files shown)


dirkhh (changes requested):
this is one of those PRs.
technically there is nothing wrong with it.
but I ask myself "WHY"?
There was nothing wrong with what we had before.
Sure, someone might argue that the distinction in names isn't needed. But maybe to some of us those names make more sense.
The only reason that this commit removes lines is because you remove things that are commented out. Other than that this is not shorter, not better.
On the contrary, I think stating explicitly that there are two different sets of models (and hopefully soon, two different sets of commands), depending on what you are building for is useful and a feature.

NAK

dirkhh:
The review only has "request changes", not a flat out NAK. So I'll "close and comment" instead.

PR #2475, opened 2019-12-25
change CodingStyle.md file name and add 2 "rules"

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ x] Documentation change
- [ ] Language translation

"This is one of those commits", but I hope I am not the only one seeing the irony in the fact that the first file not to follow the styles are CodingStyle.md

Commit 1: rename CodingStyle.md -> CODINGSTYLE.md

Having read CODINGSTYLE.md closely, I believe there are a couple of items to be clarified/amended:

commit 2:
- cpp/h filenames are identical to class name, but in lowercase

commit 3:
- the filename rule, applies only to new files  (?? my preference)
or would it be acceptable to submit a patch that e.g. corrects all filenames in tests,
this can be done without affecting the rest, but e.g. changing filenames in core/settings
affect a lot of files.

It is on purpose made as 3 commits, so that the maintainers can choose to all/part/none of this PR.


dirkhh (approved):
Of course commit 1 is then in violation of commit 3, right? :-)
And commit 2 is phrasing this very strongly - we have a few exceptions to this where we have multiple, strongly related classes in one file with a more generic name.
These comments are of course nit picking.
I'll slightly edit commit 2 and merge this as well

janiversen:
yes of course commit 3 is a violation of 1 :-)

I did it as 3 commits so you could pick and choose, and/or to edit the text.

It is also ok if you squash the 3, because there are not reason for it being in 3.

dirkhh:
merged
(hey, it being three ups your commit count if nothing else)

Editorial comment, NOT related to this PR but to the other ones I merged this morning.

The multi-part, building on top of each other, but able to discuss in smaller "increments" and therefore easier to review methodology that you settled on with the 7-part planner series works really well for me, now that I clued in to how you are doing it and have adjusted the way I apply them.

So in case you are wondering, yes, this will continue to work for me moving forward.

I'm excited to see the progress Subsurface-mobile is making.

janiversen:
I will continue the same pattern.

I too am happy to see about 1 years work get out to a broader usage than our little 25 person club locally.

I see now that I should have been more careful with my changes, because backporting is not the easiest thing.

janiversen:
but no problem I can reopen and rebase

PR #2479, opened 2019-12-25
[do not merge, need rebase] (planner 6) prepare for first part of UI

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

All new pages in QML will use Template* instead of the class (e.g. TemplateLabel not label),
using this approach allows easy central change of styles (and sometimes more), and saves a lot
of writing in each page.
The current system already contains a couple of templates (like e.g. SSRFButton) but they do not
cover a lot, and the current use is at least not in all places.

The introduction of these templates (and more to come) does NOT imply that existing pages will be modified. That is far from the case, and would really not bring anything but a lot of work. Existing
pages will be modified when/if they are redesigned.

Especially TemplateSection offer new interesting functionality.

The decision when a use of a class is one-off or be converted in a template is simple:
- class has a direct graphical representation (not just layout like e.g. column)
- class is used more than 3 times, with the same e.g. color/font
Our implementation contains currently 14 templates with 2 more on the way, but will not be
introduced here until needed.

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -150,3 +150,13 @@ void plannerShared::set_switch_at_req_stop(bool value)
 {
 	DivePlannerPointsModel::instance()->setSwitchAtReqStop(value);
 }
+
+bool plannerShared::doo2breaks()
+{
+	return qPrefDivePlanner::doo2breaks();
+}
+void plannerShared::set_doo2breaks(bool value)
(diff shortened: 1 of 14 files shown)


dirkhh:
manually merged

PR #2482, opened 2019-12-26
bug fix: make PIN page work as expected.

janiversen:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When pressing "register" in the pin page, it did not switch to divelog page
- wrong function called in onClick event
When pressing "no cloud" in the credential page qml reported an error
- missing parameters in function call.

Cleanup: qml function saveCredentials() just called manager.saveCredentials()
- call manager.saveCredentials() direct
prefs.showPin = pin.text removed since it was wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: secure user selects operational mode, before leaving login
 Mobile: use of qml compiler, smaller footprint and faster UI
 Mobile: support for "upload to divelogs.de"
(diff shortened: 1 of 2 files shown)


janiversen:
"checkout sources" that must be something outside of my domain.

dirkhh:
manually merged

PR #2485, opened 2019-12-27
(qmlpref 1) limit use of qmlprefs

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Removed functions cancelCredentialsPinSetup() and clearCredentials() as part of the ongoing
effort to reduce (eliminate) the use of qmlprefs

The whole startup is very enveloped and seems to have been expanded piece by piece, at least 
it is hard to see the design. This series will focus on removing qmlprefs, but by doing so it is 
necessary to change some startup functions in qmlmanager.

Because the startup is complicated together with the many partly overlapping functions, removal will be done in small steps allowing a full test after each step.

### DIFF
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -109,8 +109,10 @@ Item {
 				id: cancelpin
 				text: qsTr("Cancel")
 				onClicked: {
-					prefs.cancelCredentialsPinSetup()
-					rootItem.returnTopPage()
+					PrefCloudStorage.cloud_verification_status = CloudStatus.CS_UNKNOWN
+					prefs.cloudCredentials = CloudStatus.CS_UNKNOWN
+					manager.startPageText = qsTr("Check credentials...");
+					prefs.showPin = false;
 				}
 			}
 		}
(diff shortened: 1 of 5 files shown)


janiversen:
@dirkhh whenever I start building a list of PR, the title starts with
(<name> <number>)
where name is my reference to the series
and number is the merge sequence.

dirkhh (commented):
this looks reasonable

Right now this looks like it might add complexity to the QML code just to remove a class - so I'm not entirely convinced, yet, that this is the right thing to do.

janiversen:
Yes it adds complexity to the QML code, but (call me coward) I am not prepared to make all changes in one go. As you can see I am slowly removing qmlprefs (prefs.).

This is a step in this removal, you will seen see a PR that removes prefs. in QML, making it simpler (assuming 1 assignment is simpler than 1 function call).

janiversen:
Added a commit, which will cause push force to the remainder of the series.

dirkhh (approved):
ok, let's start working on merging this series

dirkhh:
manually merged

PR #2476, opened 2019-12-25
Bugfix: correct ratings calculations in plannerShared

janiversen:
### Describe the pull request:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Calculations were off by +- 1 due to rounding errors.

Error was found during implementation of automated test #2477

PR #2492, opened 2019-12-27
Planner test, added last test of conversions

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Added test of gas variables.

There is a problem with one of the test cases for decopo2. Bottompo2 contains exact the same
code and tests but do not have problems. This is something I will investigate a bit more.

### DIFF
--- a/tests/testplannershared.cpp
+++ b/tests/testplannershared.cpp
@@ -15,18 +15,51 @@ void TestPlannerShared::initTestCase()
 	plannerShared::instance();
 }
 
+
+// test values have been researched with official subsurface 4.9.3
+//
+// research have been done by using the UI of subsurface, while
+// checking the content of org.hohndel.subsurface.Subsurface.plist
+// (securing UI value it not the default)
+//
+// All test have been done on a iMac Pro with Catalina 10.15.2, values in the plist
+// might be different on other systems.
+// if you experience such problems, please report these so a proper adjustment can
+// be made.
+//
+// meters pr time unit and feet pr time unit
+// UI (meters) - plist value       UI (feet) - plist value
+//    16m      -    267               33f    -     168
+//     7m      -    117               27f    -     137
+//     8m      -    133               40f    -     203
+//    10m      -    167               35f    -     178
+//
+// liter pr time unit and cuft pr time unit
+// UI (liters) - plist value       UI (cuft) - plist value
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
as I look at that I realize that we don't appear to run `make check` as part of the GitHub Actions anymore. I need to fix that.

janiversen:
I have noted that too. Remark the TestPlannerShared is only available in build-mobile (due to link problems in the desktop, commands references mainWindow::instance(), not the best separation backend/UI).

dirkhh:
> I have noted that too. Remark the TestPlannerShared is only available in build-mobile (due to link problems in the desktop, commands references mainWindow::instance(), not the best separation backend/UI).

definitely something to fix at some point - I'll look into it.
In the meantime I'll make sure we test in both desktop and mobile builds.

dirkhh:
That said, I have some tests failing right now :-/

janiversen:
Yes because you have a bug fix PR outstanding #2476.

And as noted in the PR, I have a strange problem with decopo2 and some very specific values. I am right controlling the Qt sources to see how they do fuzzy double compare.

janiversen:
Now all the new test cases works here. I will remove the //TEMP once I find out why
bottompo2 can test 1.1 but decopo2 cannot.

dirkhh:
I have unrelated tests failing... e.g. `TestGitStorage` -- which is one that I __REALLY__ don't like to see failing...
I am not saying that this has anything to do with your code. Maybe I shouldn't have added that comment here and instead on my PR.
Please ignore

janiversen:
@dirkhh interesting, when I do
cd build
make check
-- or --
cd build-mobile
make check

It builds the exe and then I get some strange errors:

[100%] Built target TestQML_autogen
[100%] Built target TestQML
make[3]: *** No rule to make target `../tests/if', needed by `tests/CMakeFiles/check'.  Stop.
make[2]: *** [tests/CMakeFiles/check.dir/all] Error 2
make[1]: *** [tests/CMakeFiles/check.dir/rule] Error 2
make: *** [check] Error 2


I do not know if it is related to your problem, but wanted to inform you.

dirkhh (approved):

PR #2493, opened 2019-12-27
Mobile ui next steps

dirkhh:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

- remove `gDrawer` which provided a second name to refer to the `globalDrawer`
- tiny whitespace cleanup
- restructure user manual / contact support into sub-menu with distinct icons
- add sub-menu item for cloud password reset

### Release note:

included

### Documentation change:

this will need updates to the screen shots - but so much is happening now that will require such updates...

### Mentions:

@janiversen for the `gDrawer` change and overall to take a look


mobile-widgets/qml/main.qml
```diff
@@ -472,6 +472,16 @@ if you have network connectivity and want to sync your data to cloud storage."),
...
 					}
 				}
+				Kirigami.Action{
+					icon {
+						name: ":/icons/account_circle.svg"
+					}
+					text: qsTr("Reset Subsurface Cloud password")
+					onTriggered: {
+						Qt.openUrlExternally("https://cloud.subsurface-divelog.org/passwordreset")
+						globalDrawer.close()
```
janiversen, line 483: If you reset the password you should also do:
qPrefCloudStorage::set_cloud_storage_password("")
qPrefCloudStorage::set_cloud_verification_status(CS_UNKNOWN)
Leaving the 2 systems in sync, this will activate the startPage and ask for new credentials.

but I think it would be nice if the user got OK/not OK, and a short text describing what to do next.
You can use qmlManager:set_startPageText

mobile-widgets/qml/main.qml
```diff
@@ -161,7 +161,6 @@ Kirigami.ApplicationWindow {
 	}
 
 	globalDrawer: Kirigami.GlobalDrawer {
-		id: gDrawer
```
janiversen, line 164: I agree in changing the id, but please do not just delete the line, change it to "id: globalDrawer" so we do not depend on the default id

janiversen (changes requested):
I like these changes, have added a few couple of comments.

mobile-widgets/qml/main.qml
dirkhh: So this external link isn't about changing the password - it's about recovering it / setting a new one if the user can't log in. It gets you to a website which sends you an email with a different link in it (to authenticate you), on which you then can set a new password.
I assume that this will be highly asynchronous to running the app. And most likely the user isn't correctly authenticated when using that functionality.
So I'm having a hard time guessing what the right action is. Forcing a log out (in case the user was, in fact, logged in) seems a little harsh...

mobile-widgets/qml/main.qml
dirkhh: ok

mobile-widgets/qml/main.qml
janiversen: I totally agree, I thought it was a password reset, maybe the text should "Recover Subsurface Cloud password", in which case I agree with your changes.

janiversen (approved):
Let me just make sure my review does not block anything.

PR #2494, opened 2019-12-27
GitHub Actions: run the tests

dirkhh:
We lost running the tests when migrating to GitHub Actions...

### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

running locally there are broken tests. so this needs fixing as part of this PR

### Changes made:

simply try to enable tests both in mobile and desktop context, both on Linux and Mac

PR #2499, opened 2019-12-28
tests: deactivate testQML

janiversen:
Qt 5.13 (and maybe 5.12) introduced new macros for QUICK_TEST_MAIN_WITH_SETUP
and as testQML is loosely based on this macro it needs to be redesigned.

Before Qt 5.11 it was not possible to do QML testing that depended on
classes being registred with  the macro QUICK_TEST_MAIN_WITH_SETUP,
therefore it was nessecary to make an independent implementation of
main()

for now, let the main do a "return 0", to keep the test in the loop.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/tests/testqml.cpp
+++ b/tests/testqml.cpp
@@ -7,7 +7,8 @@
 // main loosely copied from QUICK_TEST_MAIN_WITH_SETUP macro
 int main(int argc, char **argv)
 {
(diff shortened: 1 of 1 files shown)


dirkhh:
I added this to #2494

PR #2502, opened 2019-12-29
(qmlpref 1, last round) qmlprefs::setCredentials

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I hope the commit messages describes the changes.

It is made in small parts to enable testing in between.

### DIFF
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -110,7 +110,6 @@ Item {
 				text: qsTr("Cancel")
 				onClicked: {
 					PrefCloudStorage.cloud_verification_status = CloudStatus.CS_UNKNOWN
-					prefs.cloudCredentials = CloudStatus.CS_UNKNOWN
 					manager.startPageText = qsTr("Check credentials...");
 					prefs.showPin = false;
 				}
@@ -140,7 +139,8 @@ Item {
 				onClicked: {
 					manager.setGitLocalOnly(true)
 					PrefCloudStorage.cloud_auto_sync = false
-					prefs.credentialStatus = CloudStatus.CS_NOCLOUD
+					prefs.oldStatus = PrefCloudStorage.cloud_verification_status
+					PrefCloudStorage.cloud_verification_status = CloudStatus.CS_NOCLOUD
 					manager.saveCloudCredentials("", "", "")
 					manager.openNoCloudRepo()
 				}
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -315,14 +315,14 @@ void QMLManager::openLocalThenRemote(QString url)
 		auto credStatus = qPrefCloudStorage::cloud_verification_status();
 		if (credStatus != qPrefCloudStorage::CS_NOCLOUD &&
 		    credStatus != qPrefCloudStorage::CS_INCORRECT_USER_PASSWD)
-			QMLPrefs::instance()->setCredentialStatus(qPrefCloudStorage::CS_NEED_TO_VERIFY);
+			qPrefCloudStorage::set_cloud_verification_status(qPrefCloudStorage::CS_NEED_TO_VERIFY);
 	} else {
 		// if we can load from the cache, we know that we have a valid cloud account
 		// and we know that there was at least one successful sync with the cloud when
 		// that local cache was created - so there is a common ancestor
 		setLoadFromCloud(true);
 		if (qPrefCloudStorage::cloud_verification_status() == qPrefCloudStorage::CS_UNKNOWN)
-			QMLPrefs::instance()->setCredentialStatus(qPrefCloudStorage::CS_VERIFIED);
+			qPrefCloudStorage::set_cloud_verification_status(qPrefCloudStorage::CS_VERIFIED);
 		if (git_prefs.unit_system == IMPERIAL)
 			qPrefUnits::set_unit_system("imperial");
 		else if (git_prefs.unit_system == METRIC)
@@ -470,7 +470,7 @@ void QMLManager::finishSetup()
 			appendTextToLog(QString("working in no-cloud mode, finished loading %1 dives from %2").arg(dive_table.nr).arg(existing_filename));
 		}
 	} else {
-		QMLPrefs::instance()->setCredentialStatus(qPrefCloudStorage::CS_UNKNOWN);
+		qPrefCloudStorage::set_cloud_verification_status(qPrefCloudStorage::CS_UNKNOWN);
 		appendTextToLog(tr("no cloud credentials"));
 		setStartPageText(RED_FONT + tr("Please enter valid cloud credentials.") + END_FONT);
 	}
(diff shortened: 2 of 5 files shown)


mobile-widgets/qmlmanager.cpp
```diff
@@ -444,7 +444,7 @@ void QMLManager::finishSetup()
 {
 	// Initialize cloud credentials.
 	git_local_only = !prefs.cloud_auto_sync;
-	QMLPrefs::instance()->setCredentialStatus((qPrefCloudStorage::cloud_status) prefs.cloud_verification_status);
+
```
dirkhh, line 447: the commit message refers to `the "if" will not execute` which I'm sure made sense when you wrote the series and thought about the changes, but is really hard to understand when you read this commit. What `if` and where?
Would it be better to describe in words what cannot happen?

mobile-widgets/qmlmanager.cpp
```diff
@@ -456,7 +456,9 @@ void QMLManager::finishSetup()
 		openLocalThenRemote(url);
 	} else if (!empty_string(existing_filename) &&
 				qPrefCloudStorage::cloud_verification_status() != qPrefCloudStorage::CS_UNKNOWN) {
-		QMLPrefs::instance()->setCredentialStatus(qPrefCloudStorage::CS_NOCLOUD);
+		QMLPrefs::instance()->setOldStatus((qPrefCloudStorage::cloud_status)qPrefCloudStorage::cloud_verification_status());
+		set_filename(NOCLOUD_LOCALSTORAGE);
+		qPrefCloudStorage::set_cloud_verification_status(qPrefCloudStorage::CS_NOCLOUD);
```
dirkhh, line 461: again, I struggle to correlate the commit message to the commit.
I can see that you are calling `setOldStatus()` and `set_filename()`, having that in the commit message is kinda pointless. What the commit message should tell me is why we are adding these calls.
And then the commit message talks about `saveCredentials()` which I cannot see in this commit. And it doesn't mention at all that we are switching from `setCredentialStatus()` to `set_cloud_verification_status()` and what the implications of that may be.

I'm not saying this code is wrong. I am just confused what you are doing here and the commit message is increasing that confusion.

mobile-widgets/qmlmanager.cpp
```diff
@@ -526,7 +526,7 @@ void QMLManager::saveCloudCredentials(const QString &newEmail, const QString &ne
 	if (qPrefCloudStorage::cloud_verification_status() != qPrefCloudStorage::CS_NOCLOUD &&
 		!cloudCredentialsChanged) {
 		// just go back to the dive list
-		QMLPrefs::instance()->setCredentialStatus(QMLPrefs::instance()->oldStatus());
+		qPrefCloudStorage::set_cloud_verification_status(QMLPrefs::instance()->oldStatus());
```
dirkhh, line 531: the commit message tells me that you are `setting oldStatus() directly to qPrefCloudStorage`
This may be a language issue, but based on that message I would assume code that says
`oldStatus = qPrefCloudStorage`
which is of course not what you are trying to tell me.
Did you mean something like `restore the previous state in qPrefCloudStorage instead of QMLPrefs` and then continue on to explain why that is the right thing to do?

mobile-widgets/qmlprefs.cpp
```diff
@@ -36,13 +35,13 @@ QMLPrefs *QMLPrefs::instance()
 
 qPrefCloudStorage::cloud_status QMLPrefs::credentialStatus() const
 {
-	return m_credentialStatus;
+	return (qPrefCloudStorage::cloud_status)qPrefCloudStorage::cloud_verification_status();
```
dirkhh: so the commit message tells me that you are `removing m_cloudCredentials`.
That appears to be incorrect. The thing that you are removing is called `m_credentialStatus`
And what the commit message _should_ tell the reader is actually what the commit is really doing.
Maybe something like `replace m_credentialStatus with accesses to cloud_verification_status in the qPrefCloudStorage class, because... in order to... preparing for...`
I want the commit message to tell me why you are doing what you are doing. And if it talks about what you are doing, it should correctly summarize what you are doing.

mobile-widgets/qml/CloudCredentials.qml
```diff
@@ -140,7 +139,8 @@ Item {
 				onClicked: {
 					manager.setGitLocalOnly(true)
 					PrefCloudStorage.cloud_auto_sync = false
-					prefs.credentialStatus = CloudStatus.CS_NOCLOUD
+					prefs.oldStatus = PrefCloudStorage.cloud_verification_status
```
dirkhh, line 142: So this is net new code that doesn't appear to be explained in the commit message. Why are you saving the `oldStatus` here?

dirkhh: or is this the odd reference to `does not hit the "if"` - which I have no idea what that tells me...

mobile-widgets/qmlmanager.cpp
```diff
@@ -315,14 +315,14 @@ void QMLManager::openLocalThenRemote(QString url)
 		auto credStatus = qPrefCloudStorage::cloud_verification_status();
 		if (credStatus != qPrefCloudStorage::CS_NOCLOUD &&
 		    credStatus != qPrefCloudStorage::CS_INCORRECT_USER_PASSWD)
-			QMLPrefs::instance()->setCredentialStatus(qPrefCloudStorage::CS_NEED_TO_VERIFY);
+			qPrefCloudStorage::set_cloud_verification_status(qPrefCloudStorage::CS_NEED_TO_VERIFY);
```
dirkhh, line 318: `qmlprefs::setCredentialStatus() only makes something special in case of CS_NOCLOUD.`
I'm sure that made sense to you when you wrote it. But to the reader that's at best a riddle.
The commit message only confuses me - so you are obviously replacing the `setCredentialStatus()` calls with `set_cloud_verification_status()`. Can you describe why that's the right thing to do? And if there is a special case that is worth explaining (`makes something special`), then maybe be more explicit about that?

dirkhh (changes requested):
Obviously the code that you are cleaning up is a total mess.
But I struggle with this series because the explanations in the commit messages don't help me much to understand what you are doing and why that's the right thing to do.
Your commit messages often are very transactional. They tell me what you are doing in the commit - which usually I can see. Sometimes they are actually wrong and don't describe what the commit is doing - I pointed those out. But what I __REALLY__ want to see in the commit messages is the "why" and the "so what" and the "this is why this matters" or "this is why this is correct" or "this is where this is leading".
As you are writing the code and the commits I'm sure that's all in your mind and perfectly clear to you, so you are just quickly describing what the commit does. But when I read the code I'm in the opposite situation. I can see that a commit replaces one call with a different one. But I don't know why.

mobile-widgets/qmlmanager.cpp
janiversen: Refrased

mobile-widgets/qmlmanager.cpp
janiversen: Refrased

mobile-widgets/qmlmanager.cpp
janiversen: refrased

mobile-widgets/qmlprefs.cpp
janiversen: refrased.

mobile-widgets/qml/CloudCredentials.qml
janiversen: explained (hopefully in an understandable way).

mobile-widgets/qmlmanager.cpp
janiversen: refrased

dirkhh:
I'm heading out to do family things again... I'll look through this again in a few hours and hopefully will be able to merge it so  we can move on to more fun stuff :-)

I don't say this often enough: THANKS for cleaning up (admittedly: my) mess.

janiversen:
have a nice time...thanks is not need (but very welcome).

dirkhh:
merged

PR #2503, opened 2019-12-29
(qmlpref 2, last round) remove showPin

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

continued cleanup of qmlprefs.

### DIFF
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -110,7 +110,6 @@ Item {
 				text: qsTr("Cancel")
 				onClicked: {
 					PrefCloudStorage.cloud_verification_status = CloudStatus.CS_UNKNOWN
-					prefs.cloudCredentials = CloudStatus.CS_UNKNOWN
 					manager.startPageText = qsTr("Check credentials...");
 					prefs.showPin = false;
 				}
@@ -140,7 +139,8 @@ Item {
 				onClicked: {
 					manager.setGitLocalOnly(true)
 					PrefCloudStorage.cloud_auto_sync = false
-					prefs.credentialStatus = CloudStatus.CS_NOCLOUD
+					prefs.oldStatus = PrefCloudStorage.cloud_verification_status
+					PrefCloudStorage.cloud_verification_status = CloudStatus.CS_NOCLOUD
 					manager.saveCloudCredentials("", "", "")
 					manager.openNoCloudRepo()
 				}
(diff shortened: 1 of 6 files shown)


mobile-widgets/qml/CloudCredentials.qml
```diff
@@ -46,31 +46,31 @@ Item {
 
 		Controls.Label {
 			text: qsTr("Email")
-			visible: !prefs.showPin
+			visible: PrefCloudStorage.cloud_verification_status !== CloudStatus.CS_NEED_TO_VERIFY
```
dirkhh, line 49: this isn't wrong.
but is it better?

mobile-widgets/qmlmanager.cpp
```diff
@@ -339,10 +339,6 @@ void QMLManager::openLocalThenRemote(QString url)
 		appendTextToLog(QStringLiteral("%1 dives loaded from cache").arg(dive_table.nr));
 		setNotificationText(tr("%1 dives loaded from local dive data file").arg(dive_table.nr));
 	}
-	if (qPrefCloudStorage::cloud_verification_status() == qPrefCloudStorage::CS_NEED_TO_VERIFY) {
-		appendTextToLog(QStringLiteral("have cloud credentials, but still needs PIN"));
```
dirkhh, line 343: so we are missing this fairly useful piece of information in the log

dirkhh (commented):
Yes, this removes some code.
I am not convinced that this is an improvement.
`showPin` is very easy to understand.
I'll go along with it, but this seems fairly arbitrary to me.
I sometimes like to have a variable somewhere that describes a state that the app is in.

janiversen:
I would agree with your statement, if showPin was a cover functions for something like
return (PrefCloudStorage.cloud_verification_status === CloudStatus.CS_NEED_TO_VERIFY), but it is a separate variable, meaning it is difficult to know if there are situations where:
showPin == true and
PrefCloudStorage.cloud_verification_status !== CloudStatus.CS_NEED_TO_VERIFY)

Of course I could have chosen to make such a cover function, but I preferred a clean replace, which is easy to review, and then maybe later (once the oldStatus issue is solved) make something more elegant.

mobile-widgets/qmlmanager.cpp
janiversen: it's back.

dirkhh:
good explanation. Makes sense.

dirkhh:
merged

PR #2504, opened 2019-12-29
(qmlpref 3, last round), solve oldStatus problem, remove qmlprefs.*

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

With this PR, the qmlpref series is closed, and qmlpref is removed.

After 2 attempts to clean oldStatus from qmlmanager, it was decided to move that signal into qmlmanager.

oldStatus is only used to see if user changed from offline to online, in order to merge the dives. However this approach is not complete:
- oldStatus is (and was) not stored on disk, so if the user ask to change credentials and then closes the application, any offline dives will be lost.

The right use (for a mobile application) would be, not to have 2 file locations (offline and cached) but simply use offline as cache when online. That way 
- changing offline -> online, would merge dives to the cloud automatically
  just like happens with dive entered in the cache
- changing online -> offline, would keep the dives available.

Why would a user go offline, our case is simple, be aboard a ship pretty far from any mobile masts, then going offline saves some startup time (no try to connect). This is something I have wanted to dig into for a while but the complexity is very high, especially when I compare the desktop login to the mobile login.

@dirkhh you very skeptical about parts of the commit, but does the result (a simpler C++/QML interface, and a little simpler login) weigh up the questionable commits ?


janiversen:
@dirkhh I forgot, in order to test PIN etc, I have created a number of sub accounts in the cloud.

My main test account is jancasacondor@gmail.com the sub accounts are named jancasacondor+test?@gmail.com

The sub account can be deleted. Btw is there a way (like recover password) where I can delete these accounts myself ?

dirkhh:
You appear to be mixing terms here - or maybe I don't understand what you are trying to tell me.
Part of the problem is that in many cases in your previous commits you have incorrectly called "no cloud" mode "offline". Which is highly misleading.

If you have a cloud account associated with your app, then regardless of whether you are online or offline, i.e. regardless of whether you are syncing to the cloud or just making changes locally on your device and not even attempting to connect to the cloud, the files are in the same location. But that location is derived from your email address that you are using for the cloud account.

If you explicitly tell Subsurface-mobile that you don't want a cloud account (which is something that for reasons I cannot explain is extremely important to some people who happily will use an Apple or Google phone, but who tell me that cloud=BAD), then there is no email address associated with the cloud storage and therefore your local cache (which is also your only storage of dives) is at a location that is different from the location the files will be at if you later decide that maybe you do want a cloud account after all. And that is the case where we need to move things around.

Now either I have told you something you already knew and completely missed the point you were talking about, or maybe I have helped you clear up a misunderstanding... :-)

dirkhh:
> @dirkhh I forgot, in order to test PIN etc, I have created a number of sub accounts in the cloud.
> 
> My main test account is [jancasacondor@gmail.com](mailto:jancasacondor@gmail.com) the sub accounts are named jancasacondor+test?@gmail.com
> 
> The sub account can be deleted. Btw is there a way (like recover password) where I can delete these accounts myself ?

I can delete those. And no, there is no backend for a user to delete their own account. That's too much open heart surgery in a somewhat fragile system for my taste.

mobile-widgets/qmlmanager.cpp
```diff
@@ -154,7 +154,8 @@ QMLManager::QMLManager() : m_locationServiceEnabled(false),
 	m_selectedDiveTimestamp(0),
 	alreadySaving(false),
 	m_pluggedInDeviceName(""),
-	m_showNonDiveComputers(false)
+	m_showNonDiveComputers(false),
+	m_oldStatus(qPrefCloudStorage::CS_UNKNOWN)
```
dirkhh, line 157: the commit message makes no sense. 
`Removal of oldStatus in qmlmanager proved to be complicated,
so instead the variable is moved into qmlmanager.`
Can you rephrase this?

dirkhh (commented):
nice, we removed a class.
that was a lot of open heart surgery just to do that.
I'm not arguing that this is the wrong thing to do.
I'm asking myself "how much better is Subsurface-mobile now". What did the user gain? Or, maybe, what did the developer gain? How has this become significantly easier?

Because you touched a ton of code that sucked and was convoluted and bad. But apparently mostly worked. Which means besides trying to make sense of your changes and how they relate to the commit messages, I now also need to make sure that none of this had unintended side effects. That it didn't break things that maybe you haven't thought of or haven't come across.

So, net gain for the user? I'd argue zero.
Net gain for the code? yeah, I think it's cleaner.
Cost for me? Several hours, many still ahead of me.

This isn't wrong. Maybe I shouldn't complain about it and just silently accept it.
I'm just trying to point out to you that in your quest to make Subsurface-mobile better, I'm not sure if this is the best way to spend your time and mine. We are lucky in that this is falling during the holidays and I have more time as my day job is fairly quiet and this is more competing for my time with fixing bugs, adding features, and spending time with my family.

mobile-widgets/qmlmanager.cpp
janiversen: refrased.

janiversen:
I too do not understand why users would want to the NO_CLOUD mode permanently...but I see we have 3 modes:
1) real NO_CLOUD mode (no credentials etc).
2) local cache cloud mode (credentials verified, but no network connection or automatic sync
not enabled)
3) cloud mode ((credentials verified and network connection)

1), 2) are offline in my terminology, but with the very important difference that 2) tries to connect to your cloud server (you get a message using local cache)
3) is of course online.

I would like a combination of 1) and 2) so that when NO_CLOUD mode is selected, the user have all the dives, but not network activity.

janiversen:
You sure know how to motivate a person, I guess I brought it on myself by asking. 

To me qmlmanager (and earlier also qmlprefs) is one big monolith of confusing functions, that makes it very hard to implement new functionality in a non-confusing manner, and that is surely worth my time. I cannot talk about your time, but I would have thought part-cleaning one of the most enveloped parts of qmlmanager should be worth your time too.

I agree the net gain for the user directly is zero, but with your argument we can freeze mobile as it is. 

I do not make these changes just for the fun, but maintaining the QML code with the old C++ interface is no fun.

I sense your preference is, without saying it, do not make such changes it is not worth the effort. If I sense correct I foresee bigger problems as I try to get more changes into the code.

dirkhh:
> I too do not understand why users would want to the NO_CLOUD mode permanently...but I see we have 3 modes:
> 
>     1. real NO_CLOUD mode (no credentials etc).
> 
>     2. local cache cloud mode (credentials verified, but no network connection or automatic sync
>        not enabled)
> 
>     3. cloud mode ((credentials verified and network connection)
> 
> 
> 1), 2) are offline in my terminology, but with the very important difference that 2) tries to connect to your cloud server (you get a message using local cache)
> 3) is of course online.
> 
> I would like a combination of 1) and 2) so that when NO_CLOUD mode is selected, the user have all the dives, but not network activity.

Fun fact - git always operates on your local git tree. There is no "remote only" or "online" way of doing things in git. So the way git storage works, there is no "online" or "offline". You have a local git repo. You can chose to sync that with the server at any point in time.
That's why I find that terminology that you use confusing. Especially since some of the developers think about this as "no-cloud" == offline and "have-cloud-account" == online -- and at least in some of the code comments that logic is visible.

Now, we already have what you are asking for. You have all your dives. They are in a local git repo.
If you now add cloud credentials to push them to the cloud, that's when things get tricky. Because that fundamentally changes your local git repo. Part of that are of course design decisions Linus made when first implementing all this (e.g. the branch name being your email address). But untangling that now with about ten thousand people using it and using it using dozens of current and old versions of our software on six different platforms (we have a handful of users on FreeBSD and OpenBSD)... that's more pain that I'm ready for.

So please don't describe this as "we don't have this" or "I want this". Because we have it and it works. It's just a bit fragile and you identified a situation where the current code breaks. But that's an implementation detail that can be fixed.

dirkhh:
> You sure know how to motivate a person, I guess I brought it on myself by asking.

Yeah, I'm a barrel of laughs sometimes. Especially when waking up at 5 in the morning after 4h of sleep, yet unable to go back to sleep because my brain is spinning.

I am not trying to demotivate you. I'm trying to share with you what I'm thinking. I painstakingly try to make sure that you don't see this as a NAK on your code (as we had this problems repeatedly in the past). But also, we learned in the past that I need to be very very explicit in what I want and don't want. So my words are typically chosen in the hope that they don't require much interpretation. They are questions, they are comments. They are not judgments unless they are phrased as judgment, i.e., "this is wrong" or "this makes no sense".
Or, as you have seen once or twice "NAK, I don't want that change".

> To me qmlmanager (and earlier also qmlprefs) is one big monolith of confusing functions, that makes it very hard to implement new functionality in a non-confusing manner, 

completely and 100% agreed.

> and that is surely worth my time. I cannot talk about your time, but I would have thought part-cleaning one of the most enveloped parts of qmlmanager should be worth your time too.

Also agreed. That doesn't mean I have to enjoy it, right?

> I agree the net gain for the user directly is zero, but with your argument we can freeze mobile as it is.
> 
> I do not make these changes just for the fun, but maintaining the QML code with the old C++ interface is no fun.
> 
> I sense your preference is, without saying it, do not make such changes it is not worth the effort. If I sense correct I foresee bigger problems as I try to get more changes into the code.

No, here you are interpreting my words and are trying to find subcontext that really isn't there.
I looked at this series and I thought "Jan is working on the planner - how does that help with the planner?".
I absolutely want the mobile app to get better. And I absolutely agree that this will require to clean up the cestpool of madness that I have created there.

And I apologize for phrasing my comments to you in a  way that appeared to you to blame you for the fragility and the issues around that. I was simply venting my frustration that this is crazy fragile code... and yes, I was in effect saying that I would have rather not touched this. I guess it's like going to the dentist. I of course KNOW that this is the right thing to do and that I should go to the dentist and that I should encourage you to clean up this code. That doesn't mean I enjoy either of those. And at times (especially early Sunday morning, before coffee) that may result in me being grumpy and whiny and complainy about the person who causes me the pain - which in that state of mind appears to be the dentist. Or you. And not me, myself, I, by not brushing my teeth enough and by having written and maintained such a disastrously poorly done mess in the first place.

Makes more sense?

Anyway, please do NOT get discouraged by my comments. Let's address the substance and clean up the code and create an amazing mobile app.

janiversen:
Just for the record, I am working on the planner but also a couple of other things, we need to have in place for the planner/new statistics/new navigation to work.

I am not discouraged and I like then you speak directly, backside is that it makes me also speak directly :-)

I totally agree, you once told me "a full fledged mobile version, will not be before 2020" (remember?) and I intent to prove it will be in 2020. The current division of models/filters is making it a bit more difficult, because I have modified the old models, but really like what I see.

dirkhh:
merged

PR #2509, opened 2020-01-01
Git storage refactor

dirkhh:
Make it possible to use different cloud credentials in TestGitStorage - and then do that in our GitHub Actions.

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


neolit123:
the test and build workflow changes LGTM.

dirkhh:
> the test and build workflow changes LGTM.

turns out they only LOOK good. :-(
Pull requests can't use the secrets. I knew that. DUH.
So this needs to be done differently.

glance-:
A Idea on how to fix this is to test it against a local git repo, and point subsurface tests against a file:// url instead.

dirkhh:
> A Idea on how to fix this is to test it against a local git repo, and point subsurface tests against a file:// url instead.

I fixed it by hard-coding credentials again. Oh well. Seems to work. And does exercise the live connection to the cloud storage server, which really is the point here.

PR #2512, opened 2020-01-01
mobile-widgets/qml: reintroduce showPin

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

reintroduce showPint the proper way, as a direct check on
PrefCloudStorage and without a shadow variable.

This secures PrefCloudStorage is only tested once and not
n times. As a sideeffect it saves typing and control.

### DIFF
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -13,6 +13,7 @@ Item {
 
 	property string username: login.text;
 	property string password: password.text;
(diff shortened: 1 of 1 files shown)


janiversen:
TestGitStorage failed (again), that is clearly not caused by this patch.

dirkhh:
TestGitStorage is a bit temperamental. I'm trying to figure out why but right now the information is a wee bit confusing. So I started a branch using different accounts to avoid it being race conditions. But that still hasn't explained it. So for now assume that those failures are acceptable.

dirkhh (approved):
looks good

PR #2514, opened 2020-01-02
Bugfix: solved smaller connect problems in desktop-widgets/diveplanner.cpp

janiversen:
### Describe the pull request:

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

There was a couple of typos and the QSignalMapper caused problems.

This was detected while testing mobile and comparing to desktop.

PR #2515, opened 2020-01-02
code cleanup: change connect() to new syntax

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

change all connect() in diveplanner.cpp to new syntax.

this PR sits on top of #2514


janiversen:
@dirkhh This is really a deep deep hole, because a lot of the widgets return e.g. QString, even though it is declared as (int). QSpinBox is a good example, I have to read a bit more about how to overcome this (without rewriting the whole).

janiversen:
@atdotde when you start looking, you might look at this PR as well. It turns out changing all connect() to new syntax is a bit of research (parameter search).

dirkhh:
And maybe this isn't the thing to spend time on right now. 
I remarked on this in a different review as something that we should investigate - I am not saying that you are the one who should do this, nor am I saying that now is the time to do it.
And as we have seen before, in some circumstances the old syntax is simply superior and if that's the case then maybe we shouldn't change it.

janiversen:
I know you did say it was not my problem, but that should not prevent me from trying, as it seems to be a general challenge, and I wanted to know how to solve these kind of connect().

The QOverload() template, helps a lot in avoiding a fallback to old syntax.

I will stop working on this for now, but at least we have a good chunk converted.

dirkhh (approved):
This looks good to me and clearly catches a bug or two.

janiversen:
Would not be bad to have this and #2514 in master. if so I promise to repair/change anything Robert mentions when he gets time.

I am currently doing 3 way testing
- Release desktop version
- Master desktop version
- Master++ mobile version
And everytime I have to include these 2 PRs in the build.

dirkhh:
My point is that 2515 includes the commits of 2514

janiversen:
yes it does, I was just trying to make a joke.

atdotde (approved):
Looks sane (had to educate myself about QOverload).

bstoeger (approved):
LGTM. We may think about upping the requirement to C++14 and use the slightly less verbose qOverload, though that's cosmetics.

I didn't follow the conversation - what is the reason for non-complete conversion (e.g. set_bottomsac)?

dirkhh:
I think it makes sense to merge this as is and then continue on top of it.
Thanks @janiversen  and thanks everyone else for the reviews

PR #2516, opened 2020-01-02
(Planner 1): add mechanism to change between m/min and ft/min 

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add code to qml, to start testing online change of METERS <-> FEET.

### DIFF
--- a/mobile-widgets/qml/DivePlannerSetup.qml
+++ b/mobile-widgets/qml/DivePlannerSetup.qml
@@ -10,6 +10,7 @@ import org.kde.kirigami 2.4 as Kirigami
 Kirigami.ScrollablePage {
 	title: qsTr("Dive planner setup")
 
+	property string speedUnit: (PrefUnits.length === "meters") ? qsTr("m/min") : qsTr("ft/min")
 	Column {
 		width: parent.width
 		spacing: 1
@@ -39,7 +40,7 @@ Kirigami.ScrollablePage {
 					stepSize: 1
 					value: Planner.ascrate75
 					textFromValue: function (value, locale) {
-						return value + qsTr("m/min")
+						return value + speedUnit
 					}
 					onValueModified: {
 						Planner.ascrate75 = value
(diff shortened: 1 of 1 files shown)


mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -10,6 +10,7 @@ import org.kde.kirigami 2.4 as Kirigami
 Kirigami.ScrollablePage {
 	title: qsTr("Dive planner setup")
 
+	property string textLength: (PrefUnits.length === "meters") ? qsTr("m/min") : qsTr("ft/min")
```
dirkhh: I an not in love with the name of that property. Wouldn't this be `speedUnit` or something along those line?

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -10,7 +10,14 @@ import org.kde.kirigami 2.4 as Kirigami
...
 	title: qsTr("Dive planner setup")
 
-	property string textLength: (PrefUnits.length === "meters") ? qsTr("m/min") : qsTr("ft/min")
+	property string textLength: (PrefUnits.unit_system === "metric") ? qsTr("m/min") : qsTr("ft/min")
+//	property string textLength: PrefUnits.unit_system
+	onTextLengthChanged: {
+		console.log("--> New text << " + textLength )
+		spinAscrate75.value++
+		spinAscrate75.value--
+	}
```
dirkhh, line 26: I don't understand what you are trying to do here. The increment/decrement looks like this is test code?

dirkhh (changes requested):
Two comments, one on the property name, one questioning if this is production code :-)

mobile-widgets/qml/DivePlannerSetup.qml
janiversen: changed

mobile-widgets/qml/DivePlannerSetup.qml
janiversen: it is not test code, added a larger comment as to why.

dirkhh (approved):
that ++/-- is one of those things where my jaw just hits the floor. QML is a strange beast for sure.
Thanks for the explanation

dirkhh:

PR #2518, opened 2020-01-02
(Planner 3): make qPrefUnit work and signal correct.

janiversen:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Secure all signals are sent when changing IMPERIAL <-> METRICS

### DIFF
--- a/core/settings/qPrefUnit.cpp
+++ b/core/settings/qPrefUnit.cpp
@@ -94,13 +94,16 @@ QString qPrefUnits::unit_system()
 }
 void qPrefUnits::set_unit_system(const QString& value)
 {
-	short int v = value == QStringLiteral("metric") ? METRIC :
-					value == QStringLiteral("imperial")? IMPERIAL :
-							PERSONALIZE;
-	if (v == METRIC) {
+	set_unit_system(value == QStringLiteral("metric") ?  METRIC :
+					value == QStringLiteral("imperial")? IMPERIAL : PERSONALIZE);
+	emit instance()->unit_systemStringChanged(value);
+}
+void qPrefUnits::set_unit_system(unit_system_values value)
+{
+	if (value == METRIC) {
 		prefs.unit_system = METRIC;
 		prefs.units = SI_units;
-	} else if (v == IMPERIAL) {
+	} else if (value == IMPERIAL) {
 		prefs.unit_system = IMPERIAL;
 		prefs.units = IMPERIAL_units;
 	} else {
--- a/core/settings/qPrefUnit.cpp
+++ b/core/settings/qPrefUnit.cpp
@@ -125,6 +125,15 @@ void qPrefUnits::set_unit_system(unit_system_values value)
 	}
 	disk_unit_system(true);
 	emit instance()->unit_systemChanged(value);
+	emit instance()->unit_systemStringChanged(unit_system());
+	emit instance()->volumeChanged(prefs.units.volume);
+	emit instance()->volumeStringChanged(volume());
+	emit instance()->weightChanged(prefs.units.weight);
+	emit instance()->weightStringChanged(weight());
+	emit instance()->lengthChanged(prefs.units.length);
+	emit instance()->lengthStringChanged(length());
+	emit instance()->temperatureChanged(prefs.units.temperature);
+	emit instance()->temperatureStringChanged(temperature());
 }
 DISK_LOADSYNC_ENUM(Units, "unit_system", unit_system_values, unit_system);
 
--- a/core/settings/qPrefUnit.cpp
+++ b/core/settings/qPrefUnit.cpp
@@ -36,10 +36,9 @@ QString qPrefUnits::duration_units()
 }
 void qPrefUnits::set_duration_units(const QString& value)
 {
-	prefs.units.duration_units = value == QStringLiteral("hours") ? units::DURATION::ALWAYS_HOURS :
-								 value == QStringLiteral("minutes")? units::DURATION::MINUTES_ONLY :
-																	 units::DURATION::MIXED;
-	disk_duration_units(true);
+	set_duration_units(value == QStringLiteral("hours") ?  units::DURATION::ALWAYS_HOURS :
+					   value == QStringLiteral("minutes")? units::DURATION::MINUTES_ONLY :
+														   units::DURATION::MIXED);
 	emit instance()->duration_unitsStringChanged(value);
 }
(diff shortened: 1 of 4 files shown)


core/settings/qPrefUnit.cpp
```diff
@@ -94,13 +94,17 @@ QString qPrefUnits::unit_system()
 }
 void qPrefUnits::set_unit_system(const QString& value)
 {
-	short int v = value == QStringLiteral("metric") ? METRIC :
+	set_unit_system(value == QStringLiteral("metric") ?  METRIC :
 					value == QStringLiteral("imperial")? IMPERIAL :
-							PERSONALIZE;
-	if (v == METRIC) {
+							                             PERSONALIZE);
```
dirkhh: that looks like odd whitespace - I think this may be Xcode introducing its own ideas of continuation line indentation...

core/settings/qPrefUnit.h
```diff
@@ -14,7 +14,7 @@ class qPrefUnits : public QObject {
 	Q_PROPERTY(QString pressure READ pressure WRITE set_pressure NOTIFY pressureStringChanged)
 	Q_PROPERTY(bool show_units_table READ show_units_table WRITE set_show_units_table NOTIFY show_units_tableChanged)
 	Q_PROPERTY(QString temperature READ temperature WRITE set_temperature NOTIFY temperatureStringChanged)
-	Q_PROPERTY(QString unit_system READ unit_system WRITE set_unit_system NOTIFY unit_systemChanged)
+	Q_PROPERTY(QString unit_system READ unit_system WRITE set_unit_system NOTIFY unit_systemStringChanged)
```
dirkhh, line 17: doesn't that remove the `enum` based version?
or is this something where both flavors work because of overloading? In that case, why does the signal name have to change?

core/settings/qPrefUnit.h
```diff
@@ -77,7 +78,8 @@ public slots:
 	void show_units_tableChanged(bool value);
 	void temperatureChanged(int value);
 	void temperatureStringChanged(const QString& value);
-	void unit_systemChanged(const QString& value);
+	void unit_systemChanged(int value);
```
dirkhh, line 81: it seems like we don't reference this `int` based version elsewhere...

core/settings/qPrefUnit.cpp
```diff
@@ -104,9 +104,35 @@ void qPrefUnits::set_unit_system(unit_system_values value)
...
+		prefs.units.volume = units::VOLUME::CUFT;
+		set_volume(units::VOLUME::LITER);
+		prefs.units.weight = units::WEIGHT::LBS;
+		set_weight(units::WEIGHT::KG);
+		prefs.units.length = units::LENGTH::FEET;
+		set_length(units::LENGTH::METERS);
+		prefs.units.pressure = units::PRESSURE::PSI;
+		set_pressure(units::PRESSURE::BAR);
+		prefs.units.temperature = units::TEMPERATURE::FAHRENHEIT;
+		set_temperature(units::TEMPERATURE::CELSIUS);
```
dirkhh, line 108: that code looks really, really weird.
I'm not sure I have a better idea to ensure that the signals get sent... maybe the `set_xxx()` functions should have an optional argument to force sending the signal? But this just looks crazy.

Also, the commit message refers to `unit_meterChanged()` which should be `unit_lengthChanged`, I hope.

dirkhh (commented):
Ugh. This is all pretty ugly with having int and string versions.
This may be the best way to do things, but if it is then I wonder if there isn't something else that is broken in our design right now.

janiversen:
> Ugh. This is all pretty ugly with having int and string versions.
> This may be the best way to do things, but if it is then I wonder if there isn't something else that is broken in our design right now.

It is pretty normal to have string and int functions, normally they have the same name, and the C++ compiler chooses between them. The int is frequently used for internal compares while the string is frequently used for presentation.

We could avoid the string version in case of qPrefUnit but that would mean duplicating the enums used (use Q_ENUMS in the class, while keeping the original available for C).

The bigger design problem here is really to get the enums in units.h etc available to QML without duplicating them. I did work on a solution but stopped due to the amount of changes it caused.

Some day, I hope we can make the Q_ENUMS directly in units.h etc, but checking if we are compiling for C++, but that is a bigger change, that touches a lot of sensitive code.

core/settings/qPrefUnit.cpp
janiversen: removed

core/settings/qPrefUnit.h
janiversen: No we never used the enum here. I am changing the signal to <name>StringChanged(value) to be consistent with the other signals.

Normally the signal name would be the same and simply have different parameters. But having overloaded functions which causes problems in connect, and forced me to use the SIGNAL/SLOT macros, was something you did not accept, hence 2 different function names.

If you have changed your mind, I will make a patch PR afterwards and correct all of qPrefUnit in one go.

core/settings/qPrefUnit.h
janiversen: no not in this PR, but it will come in the next PR.

core/settings/qPrefUnit.cpp
janiversen: I have added a bigger comment explaining why it is needed, and also added a longer text in the commit.

It was my judgement that changing the macros (with possible side effects) or writing all these setters directly was a far bigger change.

Yes, my errorr, changed to unit_lengthChanged

janiversen:
Changed to a more understandable solution in qPrefUnit when changing between METRICS <-> IMPERIAL

core/settings/qPrefUnit.h
dirkhh: I think I followed up when we went through this and said something to the extend of "oh, with that explanation this makes sense" and then we talked about how you sometimes don't advocate well for your changes :-)
I am OK with the `SIGNAL()/SLOT()` syntax where it provides clear value. I.e. if switching to the new syntax causes ugly problems, then don't. This here is an example where using the new syntax appears to make things more complicated - in another commit you worked around that with a lambda function. If that's not an option here, then feel free to go with the old syntax.

dirkhh (commented):
I'll re-read all of this later today - the next few hours are back to back meetings so it will be a while

dirkhh:

PR #2519, opened 2020-01-02
(Planner 4): Secure planner model is recalculated and text is changed

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When changing METERS <-> FEET, secure that the model is recalculated and that the 
UI text is changed.

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -2,6 +2,7 @@
 #include "plannershared.h"
 #include "core/settings/qPrefDivePlanner.h"
 #include "core/settings/qPrefTechnicalDetails.h"
+#include "core/settings/qPrefUnit.h"
 #include "qt-models/diveplannermodel.h"
 #include "qt-models/cylindermodel.h"
 
@@ -10,6 +11,12 @@ plannerShared *plannerShared::instance()
     static plannerShared *self = new plannerShared;
     return self;
 }
+plannerShared::plannerShared()
+{
+	// Be informed when user switched METER <-> FEET
+	connect(qPrefUnits::instance(), &qPrefUnits::lengthChanged, this, &unit_lengthChangedSlot);
+}
+
 
 // Used to convert between meter/feet and keep the qPref variables independent
 #define TO_MM_BY_SEC ((prefs.units.length == units::METERS) ? 1000.0 / 60.0 : feet_to_mm(1) / 60.0)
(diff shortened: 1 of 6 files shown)


backend-shared/plannershared.cpp
```diff
@@ -301,3 +309,9 @@ void plannerShared::set_display_variations(bool value)
 {
 	DivePlannerPointsModel::instance()->setDisplayVariations(value);
 }
+
+// Handle when user change length measurement type
+void plannerShared::unit_lengthChangedSlot(int value)
+{
+	qDebug() << "---> change METER <-> FEET";
```
dirkhh: I don't think we want to check this into production code...

backend-shared/plannershared.cpp
```diff
@@ -313,5 +313,10 @@ void plannerShared::set_display_variations(bool value)
 // Handle when user change length measurement type
 void plannerShared::unit_lengthChangedSlot(int value)
 {
-	qDebug() << "---> change METER <-> FEET";
+	// Provoke recalculation of model and send of signals
+	set_ascratelast6m(ascratelast6m());
+	set_ascratestops(ascratestops());
+	set_ascrate50(ascrate50());
+	set_ascrate75(ascrate75());
+	set_descrate(descrate());
```
dirkhh, line 320: please collapse this with the previous commit and remove the `#include <QDebug>`

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -11,11 +11,19 @@ Kirigami.ScrollablePage {
...
 		spinAscrate75.value--
+		spinAscrate50.value++
+		spinAscrate50.value--
+		spinAscratestops.value++
+		spinAscratestops.value--
+		spinAscratelast6m.value++
+		spinAscratelast6m.value--
+		spinDescrate.value++
+		spinDescrate.value--
+		console.log("---> new values READY")
```
dirkhh: again, this doesn't look like the final code. If the increment/decrement pairs are needed, then we should have comments there explaining this

dirkhh (changes requested):
two of the commits should be combined... and more questions

backend-shared/plannershared.cpp
janiversen: It was just there to prove that the signal got received, I had some serious problems with qPrefUnit. 

Gone

backend-shared/plannershared.cpp
janiversen: done

mobile-widgets/qml/DivePlannerSetup.qml
janiversen: Added a longer comment in one of the previous commits. Apart from the console.log (which is gone) this is the final code. At least until I find a better way to make the spinBox update its text.

janiversen:
I should have replied to all comment in the 4 PRs, and in most cases done the requested changes.

At some point we need to make a design decision on how to handle C based enums in QML. QML requires the enum is defined in a class with an Q_OBJECT macro. This is pretty easy to do, and the C code can still access it (because we compile it as C++ code), but it is a change that does a search/replace in a LOT of files.

The other alternative (as used in qPrefCloudStorage) is to duplicate the enum, and add a big warning. That solution is pretty easy but also dangerous (to a degree, because how often do we change these fundamental enums). I have done that in my alternative implementation, and are happy to submit that code as a new PR (which includes fixing e.g. qPrefUnit).

dirkhh:

PR #2520, opened 2020-01-02
Fix get_divemode logic

atdotde:
Commit dbb504 tried to prevent an uninitialized dc pointer
from being dereferenced. But I screwed up the logic always
setting the event pointer to NULL. This fixes this error.

Reported-by: willemferguson@zoology.up.ac.za

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

Reported by @willemferguson

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -246,9 +246,11 @@ void add_extra_data(struct divecomputer *dc, const char *key, const char *value)
 enum divemode_t get_current_divemode(const struct divecomputer *dc, int time, const struct event **evp, enum divemode_t *divemode)
 {
 	const struct event *ev = *evp;
-	if (*divemode == UNDEF_COMP_TYPE && dc) {
-		*divemode = dc->divemode;
-		ev = get_next_event(dc->events, "modechange");
+	if (dc) {
+		if (*divemode == UNDEF_COMP_TYPE) {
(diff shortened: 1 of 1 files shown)


core/dive.c
```diff
@@ -246,9 +246,11 @@ void add_extra_data(struct divecomputer *dc, const char *key, const char *value)
...
 {
 	const struct event *ev = *evp;
-	if (*divemode == UNDEF_COMP_TYPE && dc) {
-		*divemode = dc->divemode;
-		ev = get_next_event(dc->events, "modechange");
+	if (dc) {
+		if (*divemode == UNDEF_COMP_TYPE) {
+			*divemode = dc->divemode;
+			ev = get_next_event(dc->events, "modechange");
+		}
```
dirkhh, line 253: so with this new code in case `dc != NULL && *divemode != UNDEF_COMP_TYPE` we don't assign `ev` at all. Is that what you want?

atdotde: Yes, that is exactly the case that needed fixing. It reuses the ev from the previous invocation. The first invocation should be with *divemode set to UNDEF_COMP_TYPE.

PR #2523, opened 2020-01-03
[WIP] Implement rudimentary yearly statistics for mobile

atdotde:
This is my first attempt that is actually going somewhere on creating something new for the mobile app. It brings the annual statistics to mobile, so far only as a text.

Comments welcome.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/mobile-widgets/qml/Statistics.qml
+++ b/mobile-widgets/qml/Statistics.qml
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0
 import QtQuick 2.12
-import QtQuick.Controls 2.12
-import QtQuick.Controls 1.4
-import QtQuick.Controls.Styles 1.4
+import QtQuick.Controls 2.12 as Controls
+//import QtQuick.Controls 1.4
+//import QtQuick.Controls.Styles 1.4
 import QtQuick.Layouts 1.12
 import QtQuick.Dialogs 1.3
 import org.subsurfacedivelog.mobile 1.0
@@ -11,23 +11,72 @@ import org.kde.kirigami 2.4 as Kirigami
 Kirigami.ScrollablePage {
 	title: qsTr("Annual statistics")
 	objectName: "statistics" // we mainly use this in debug messages to the log
-
-	TextArea {
-		style: TextAreaStyle {
-			textColor: subsurfaceTheme.textColor
-			selectedTextColor: subsurfaceTheme.contrastAccentColor
+	property QtObject statisticsModel: mobileStatisticsModel
+	Component {
+		id: yearDelegate
+		Kirigami.AbstractListItem {
+			leftPadding: Kirigami.Units.largeSpacing
+			rightPadding: Kirigami.Units.largeSpacing
 			backgroundColor: subsurfaceTheme.backgroundColor
+			width: parent.width
+			Column {
+				Row {
+					spacing: 2 * Kirigami.Units.gridUnit
+					TemplateLabel { text: year }
+					TemplateLabel { text: dives + " " + qsTr("dive(s)") }
+					TemplateLabel { text: qsTr("Total dive time:") + " " + totalTime }
+				}
+				GridLayout {
+					columns: 4
+					width: parent.width
+					// Row 0 -- this should create consistent width, but doesn't
+					Rectangle {
+						width: yearDelegate.width / 4
+						height: 1
+					}
+					Rectangle {
+						width: yearDelegate.width / 4
+						height: 1
+					}
+					Rectangle {
+						width: yearDelegate.width / 4
+						height: 1
+					}
+					Rectangle {
+						width: yearDelegate.width / 4
+						height: 1
+					}
+
+					// Row 1
+					TemplateLabel { }
+					TemplateLabel { text: qsTr("AVG") }
+					TemplateLabel { text: qsTr("MIN") }
+					TemplateLabel { text: qsTr("MAX") }
+					// Row 2
+					TemplateLabel { text: qsTr("time") }
+					TemplateLabel { text: averageTime }
+					TemplateLabel { text: shortestTime }
+					TemplateLabel { text: longestTime }
+					// Row 3
+					TemplateLabel { text: qsTr("depth") }
+					TemplateLabel { text: averageDepth }
+					TemplateLabel { text: minDepth }
+					TemplateLabel { text: maxDepth }
+					// Row 4
+					TemplateLabel { text: qsTr("SAC") }
+					TemplateLabel { text: averageSac }
+					TemplateLabel { text: minSac }
+					TemplateLabel { text: maxSac }
+				}
+			}
 		}
-		width: parent.width
-		height: parent.height
-		readOnly: true
-		textFormat: TextEdit.RichText
-		text: manager.statisticsPageText
 	}
 
-	ColumnLayout {
-		width: parent.width
-		spacing: 1
-		Layout.margins: 10
+	ListView {
+		id: statisticsListView
+		anchors.fill: parent
+		model: statisticsModel
+		delegate: yearDelegate
+		bottomMargin: Kirigami.Units.iconSizes.medium + Kirigami.Units.gridUnit
 	}
 }
--- a/mobile-widgets/qml/Statistics.qml
+++ b/mobile-widgets/qml/Statistics.qml
@@ -14,6 +14,7 @@ Kirigami.ScrollablePage {
 	    width: parent.width
 	    height: parent.height
 	    readOnly: true
+	    textFormat: TextEdit.RichText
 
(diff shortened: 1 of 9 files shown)


dirkhh (commented):
If you look at this online with "hide whitespace changes" you see that we are losing comments in the qrc file... is that a tool issue?

atdotde:
Hmm, I edited the qrc file in Qt Creator. So all non trivial changes (except for adding a file) are produced by that tool. But I could easily redo it in vi.

dirkhh:
While I typically love QtCreator, I find it frustrating how it handles .qrc files.
I'd prefer if those changes where done manually, TBH

dirkhh:
This is a thing of beauty. If your work as professor no longer fulfills you, maybe you should consider a future as graphical designer... :-) :-) :-)

Ok, that was mean and uncalled for. I will now punish myself accordingly.

I'm thrilled that you are working on this. And in a way this textual representation is certainly better than what we have today (which is nothing). The question becomes: where do we want to take it from here. Productize the textual representation (that could use some fine tuning to make it look better), or is this just a stop-gap before you will dazzle us with brilliant graphics? I'm just trying to understand how I can best help you.

atdotde:
Dirk,

> On 4. Jan 2020, at 20:12, Dirk Hohndel <notifications@github.com> wrote:
> 
> This is a thing of beauty. If your work as professor no longer fulfills you, maybe you should consider a future as graphical designer... :-) :-) :-)
> 
> 

Thanks for the praise. In fact, I have considered alternative careers if all this ballet dancing does not work out in the end.

But honestly, I would say it is not much worse than the yearly statistics of the desktop version...
> I'm thrilled that you are working on this. And in a way this textual representation is certainly better than what we have today (which is nothing). The question becomes: where do we want to take it from here. Productize the textual representation (that could use some fine tuning to make it look better), or is this just a stop-gap before you will dazzle us with brilliant graphics? I'm just trying to understand how I can best help you.
> 
> 
Of course, in the end we want colorful graphics, don’t we?

But in the meantime, as I said above, I consider this comparable (maybe after some optical tweaking) to the desktop version. So it could work of the purpose to show some dive operator you have done so many dives in the last 12 months or similar.

What I also wanted to do is to see if I can actually get something in QML to work (this is admittedly rather rudimentary) and get the data from the core to the UI (seems to work).

I had a brief look into QtCharts, but soon realized that I needed some extra libraries for that so settled for the textual representation for the time being. But that would be the natural next step. Alternatively, I could try the approach of the profile, i.e. draw on some invisible widget turn that into an image and display that. But that sounds quite roundabout to me.

Best
Robert

dirkhh:
OK, would you like me to help with the visual design, or would you like to continue this by yourself?

atdotde:
> On 5. Jan 2020, at 00:02, Dirk Hohndel <notifications@github.com> wrote:
> 
> OK, would you like me to help with the visual design, or would you like to continue this by yourself?

Sure, any improvement is welcome!

R.

dirkhh:
So after finally having dug out of the salinity nightmare, I have actually LOOKED at your code instead of just building / running it.
The good news is that I understand what you are doing and why.
The bad news is that this is _really_ not how one would do statistics in QML.
The way you would do this is to create a model and to expose the data that we currently have in a bunch of arrays to QML through that model. Which is days of boilerplate and frustration and... yeah, it's a lot of fun. Really. Trust me.

I know how to do this and I am tempted to do it - but that brings me back to the question of "what's the longer term plan" because I would hate to do all that work and then we don't use it longer term.

So short term, making this look better is a completely different project from what I expected it to be. That said, I'll give it a try :-) 

OK, got a tiny bit done, but because the TextArea only parses a relatively small subset of HTML, there isn't really a lot we can do here.

Anyway, trivial update (and it always makes me queasy that I can push into your repo...)

atdotde:
Dirk,

> On 5. Jan 2020, at 05:18, Dirk Hohndel <notifications@github.com> wrote:
> 
> So after finally having dug out of the salinity nightmare, I have actually LOOKED at your code instead of just building / running it.
> The good news is that I understand what you are doing and why.
> The bad news is that this is really not how one would do statistics in QML.
> The way you would do this is to create a model and to expose the data that we currently have in a bunch of arrays to QML through that model. Which is days of boilerplate and frustration and... yeah, it's a lot of fun. Really. Trust me.
> 
> I know how to do this and I am tempted to do it - but that brings me back to the question of "what's the longer term plan" because I would hate to do all that work and then we don't use it longer term.
> 
> So short term, making this look better is a completely different project from what I expected it to be. That said, I'll give it a try :-)
> 
> OK, got a tiny bit done, but because the TextArea only parses a relatively small subset of HTML, there isn't really a lot we can do here.
> 
> Anyway, trivial update (and it always makes me queasy that I can push into your repo...)
> 
> 
I realized that do to anything better than showing a static text would involve a model and eventually I will need this. But let’s try to achieve something else first (which will later also need a model…): Let’s get charts to work. What I intended was to first copy&paste (this is how I always proceed…) one of the QtCharts examples and then later modify it so that it shows what we want. Specifially, I added a

	ChartView {
	    id: chart
	    title: "Top-5 car brand shares in Finland"
	    anchors.fill: parent
	    legend.alignment: Qt.AlignBottom
	    antialiasing: true

	    PieSeries {
		id: pieSeries
		PieSlice { label: "Volkswagen"; value: 13.5 }
		PieSlice { label: "Toyota"; value: 10.9 }
		PieSlice { label: "Ford"; value: 8.6 }
		PieSlice { label: "Skoda"; value: 8.2 }
		PieSlice { label: "Volvo"; value: 6.8 }
	    }
	}



to Statistics.qml as well as as

 import QtCharts 2.0

on the top. But at runtime, this complains that QtCharts is not installed (it is installed on my System as this example builds and runs from the Qt Examples). Since you have more experience with this, how do I have to modify CMakeLists.txt (I guess that is the place to look) to install QtCharts?

Best
Robert

dirkhh:
Fun fact. We already have a model for yearly statistics. I am playing with using that in the mobile app...

In the meantime... your issue with QtCharts is likely that you didn't run `make install` as that is what moves the required QML files and shared libraries into a spot where the app can find it. Can you check if that's what happened? If not, can you paste the actual error output?

dirkhh:
in the meantime I realized that while we have a model for yearly statistics, that one is tied to the treemodel, which we are not (yet) using in mobile. That's something that @bstoeger and I have been working on on and off (me lately mostly off) and that is still not ready for prime time...
Which of course makes me even less interested in cloning this based on the model used in the mobile app as that is likely to go away and make all this work redundant... decisions, decisions...

dirkhh:
umm, no, that's not it, either. It has its own `TreeModel`. What the heck? I'm starting to get lost in all of these models...

bstoeger:
`TreeModel` is a convenience model by @tcanabrava. It does the memory management for a model representing a tree with an arbitrary number of levels. Since the statistics-model is the only user and has a fixed two level depths, we might as well fold one into the other. However, I don't see this as a priority. It works after all.

dirkhh:
Thanks - I figured that out and built it, but the model doesn't quite work from QML - that needs a little more work, I'm afraid. I'm playing with it as we speak :-)

bstoeger:
Could it be that QML models work via roles whereas on Desktop we use columns? That's one incompatibility that is very annoying.

dirkhh:
That is exactly it... and it's driving me unhappy right now.
I'm doing a new model from scratch rather than trying to fix this.

dirkhh:
I pushed a very simple model based implementation.
Instead of removing your commits (that seemed rude since it's a branch in your repo) I added more commits and the last one replaces the text based statistics with the model based ones.
Of course in a final PR we should avoid that detour and simplify the sequence of commits.

The UI is no better than yours - arguably maybe worse. But it's now in QML and we can style it there.

atdotde:
@dirkhh I like what you did there, except for the wrong order of years (which I corrected). Do you have any idea regarding my QtCharts problem? My guess is, once that is solved, replacing the table by pretty viewgraphs is a piece of cake now that there is a model.

djunkins:
@atdotde Not sure if it will help, but when I was working to add charts to the desktop version I used the following logic in CMakeLists.txt to include the Charts library at build time:

```
        find_package(Qt5Charts)
        if (Qt5Charts_FOUND)
                list(APPEND QT_EXTRA_COMPONENTS Charts)
                set(NO_CHARTS OFF)
        else()
                set(NO_CHARTS ON)
                message ("Qt5Charts not found -- disabling statistics charts")
        endif(Qt5Charts_FOUND)

```

dirkhh:
given current ongoing discussions, I think we should close this

PR #2524, opened 2020-01-04
make robust and simple interface for using enums in QML

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add a header file with a class that contains the enums used both in C/C++ and QML. These
enums are in effect duplicated and needs to be updated when the C/C++ variant changes.

Use of 1 class for making enums available to QML, makes problem searching a lot easier, and
the interface becomes simpler.

### DIFF
--- a/backend-shared/CMakeLists.txt
+++ b/backend-shared/CMakeLists.txt
@@ -5,6 +5,7 @@ set(BACKEND_SRCS
 	exportfuncs.h
 	plannershared.cpp
 	plannershared.h
+	typeenum.h
 )
 
 add_library(subsurface_backend_shared STATIC ${BACKEND_SRCS})
--- a/backend-shared/typeenum.h
+++ b/backend-shared/typeenum.h
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef TYPEENUM_H
+#define TYPEENUM_H
+
+// *** REMARK ***
+// This feature is currently not used
+// but is left to allow for future use.
+
+//FUTURE This header file can be included twice
+//FUTURE 1 time as enum only (TYPEENUM_COMPILE_AS_ENUM defined)
+//FUTURE 1 time as class (TYPEENUM_COMPILE_AS_ENUM not defined)
+//FUTURE #if !defined(TYPEENUM_ALLOW_ENUM) || !defined(TYPEENUM_ALLOW_CLASS)
+//FUTURE #if defined(TYPEENUM_COMPILE_AS_ENUM)
+//FUTURE #define TYPEENUM_ALLOW_ENUM
+//FUTURE #else
+//FUTURE #define TYPEENUM_ALLOW_CLASS
+//FUTURE #endif
+
+//FUTURE This header is shared between the C/C++ part of subsurface and
+//FUTURE indirectly with QML.
+//FUTURE it secures that shared enums are defined identical for QML
+//FUTURE and C++ by having the shared enums defined only once in this
+//FUTURE header.
+//
+//FUTURE TYPEENUM_COMPILE_AS_ENUM is set where it is included, to NOT
+//FUTURE generate the class (which is only needed for the QML interface)
+//FUTURE that can be registred to make the enums available in QML
+
+//FUTURE #ifndef TYPEENUM_COMPILE_AS_ENUM
+//FUTURE  This part is only relevant when generating the QML accessible
+//FUTURE  class.
+#include <QObject>
+
+class typeEnum : public QObject {
+	Q_OBJECT
+
+public:
+
+private:
+	typeEnum() {}
+
+public:
+//FUTURE #endif // TYPEENUM_COMPILE_AS_ENUM
+
+	// Shared enums
+	//FUTURE accessed in C/C++ just with the name.value
+	// accessed in QML with TypeEnum.name.value
+	enum CLOUD_STATUS {
+		CS_UNKNOWN,
+		CS_INCORRECT_USER_PASSWD,
+		CS_NEED_TO_VERIFY,
+		CS_VERIFIED,
+		CS_NOCLOUD
+	};
+
+	// Do not activate these enums until code is changed
+	enum LENGTH {
+		METERS,
+		FEET
+	};
+	enum VOLUME {
+		LITER,
+		CUFT
+	};
+	enum PRESSURE {
+		BAR,
+		PSI,
+		PASCALS
+	};
+	enum TEMPERATURE {
+		CELSIUS,
+		FAHRENHEIT,
+		KELVIN
+	};
+	enum WEIGHT {
+		KG,
+		LBS
+	};
+	enum TIME {
+		SECONDS,
+		MINUTES
+	};
+	enum DURATION {
+		MIXED,
+		MINUTES_ONLY,
+		ALWAYS_HOURS
+	};
+
+//FUTURE #ifndef TYPEENUM_COMPILE_AS_ENUM
+	Q_ENUM(CLOUD_STATUS);
+	Q_ENUM(LENGTH);
+	Q_ENUM(VOLUME);
+	Q_ENUM(PRESSURE);
+	Q_ENUM(TEMPERATURE);
+	Q_ENUM(WEIGHT);
+	Q_ENUM(TIME);
+	Q_ENUM(DURATION);
+//FUTURE #endif
+};
+//FUTURE #endif // TYPEENUM_COMPILE_AS_ENUM
+
+//FUTURE  Clear TYPEENUM_COMPILE_AS_ENUM to allow multple includes
+//FUTURE #undef TYPEENUM_COMPILE_AS_ENUM
+#endif
--- a/core/settings/qPref.cpp
+++ b/core/settings/qPref.cpp
@@ -74,7 +74,6 @@ void qPref::registerQML(QQmlEngine *engine)
 	}
 
 	// Register special types
-	qmlRegisterUncreatableType<qPrefCloudStorage>("org.subsurfacedivelog.mobile",1,0,"CloudStatus","Enum is not a type");
 	qRegisterMetaType<deco_mode>();
 	qRegisterMetaType<def_file_behavior>();
 	qRegisterMetaType<taxonomy_category>();
--- a/core/settings/qPrefCloudStorage.h
+++ b/core/settings/qPrefCloudStorage.h
@@ -38,7 +38,6 @@ public:
 		CS_VERIFIED,
 		CS_NOCLOUD
 	};
-	Q_ENUM(cloud_status);
 
 	static bool cloud_auto_sync() { return prefs.cloud_auto_sync; }
 	static QString cloud_base_url() { return prefs.cloud_base_url; }
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -13,7 +13,7 @@ Item {
 
 	property string username: login.text;
 	property string password: password.text;
-	property bool showPin: (PrefCloudStorage.cloud_verification_status === CloudStatus.CS_NEED_TO_VERIFY)
+	property bool showPin: (PrefCloudStorage.cloud_verification_status === TypeEnum.CS_NEED_TO_VERIFY)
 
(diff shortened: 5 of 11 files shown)


dirkhh:
This is fascinating and feels rather clever to me. It's one of those things where I'd really love to hear what Berthold thinks about it...

janiversen:
I have no problem hearing his opinion, but maybe it would be easier for him, when I publish the whole example.

janiversen:
@dirkhh how can I silence LGTM. It indicates that typeenum.h does not have a guard, which is false (see code).

Finally I got it to work as I wanted, it compiles and is integrated in units.h

dirkhh:
I thought about this some more and I would rather not do this.
It is going to be insanely invasive with thousands of changes and way too many opportunities for unintended side effects.
I'd rather deal with QML being the garbage that it is and pass things around as strings.

I appreciate your offer to take on this task, but I'd rather have all of us focus on things that create user visible value.

janiversen:
No problem, then I will just use the alternative, which is simpler, but more risky and less robust.

I will take remove the change to units.h, and thus leave the C/C++ side of the system unchanged, however I still want to have shared (in reality duplicates on the C++ side) enums, in order to make the QML cleaner. That way any enum problem in QML can be looked for in one place.

Doing that allows me to avoid the strings and int arguments in QML and make that part better.

The nice side effect is that we can anytime later to these invasive changes to the C/C++ side.

bstoeger:
I have not tried it. However, according to the third comment in this thread: https://forum.qt.io/topic/59869/q_enum-enum_name-works-with-only-enums-declared-inside-class-srsly

you can Q_DECLARE_METATYPE a global enum. Thus, you should be able to pass it in function and use it as properties, no?

dirkhh:
> I have not tried it. However, according to the third comment in this thread: https://forum.qt.io/topic/59869/q_enum-enum_name-works-with-only-enums-declared-inside-class-srsly
> 
> you can Q_DECLARE_METATYPE a global enum. Thus, you should be able to pass it in function and use it as properties, no?

That does sound like a much cleaner solution. Something to play with.

janiversen:
yes the namespace approach is not bad at all, it allows you to use the enum in signal/slots, but having played with it, you still need a class with Q_OBJECT and the q_enums macro to make it work in qml (at least that is my experience after having actually tried it).

While someone plays and tries (hopefully successfully) we still need a solution now that works in QML, to allow me to continue the work, and this PR solves that, without invading the rest of the code.

dirkhh (commented):
I'm sorry. I still look at this series and think to myself "this doesn't add anything that really makes things better".  It's lots of duplication, lots of boilerplate, and the QML code didn't get one bit better.

Yes, I know, you would like to avoid the duplication but changing literally every C++ file in the project to add that shared include file and redo all the enums everywhere. I find that the wrong solution.

QML sucks. It's interfaces are poorly thought out. Let's deal with that in QML, not by making everything else suffer the consequences.

I'm not sold on this idea.

janiversen:
when you build a house without having made a good fundament, what happens? do not say the QML interface is poorly thought out, you should argue that the subsurface interface was never designed, but just happened, and with that fundament you expect a simple interface.

Well this PR is NOT about changing C/C++, I changed the title. There are no changes to the C/C++ code (apart from parts interfacing QML), this PR is solely about the QML interface.

Yes enums will continue to be duplicated, as already done,  but it simplifies the life in QML. The alternative to making these enums available to QML, is to make at least the double amount of string functions, is that really more robust and easier to maintain?

dirkhh:
Jan, what I always find interesting when working with you is how challenging things get when I disagree with you.

Reading this comment above makes me want to step back again from dealing with Subsurface for a while and focus on my entertaining day job. Subsurface is supposed to be fun, to keep my mind off of the frustrations and stress at work. It's not about getting yelled at and treated like an idiot.

I do not appreciate being lectured on how I am allowed to talk about the project that I maintain.

janiversen:
I apologize if you feel offended, but you wrote about QML and its interfaces, a place where it seems we highly disagree. I do not know if we disagree on qmlmanager. I am not lecturing you, but I too can sometimes have an opinion (even though I normally suppress it).

The main issue here, seems far more to be, that I followed your guideline, and stopped pursuing the more elegant but also riskier solution, and have submitted a PR that solely concentrates on QML.

If you do not like the PR, it will not be included, and not a negative word from me about that.

dirkhh:
Jan, we are back to where we were in September.

The tone of your messages may be entirely a language issue, but it is interesting that many of the other Subsurface developers seem to all feel the same way. The others simply decline to interact with you. I don't have that option unless I decide to again ask you not to continue to contribute to Subsurface. Which I really don't want to do because I see value in the things that you are working on.

The passive aggressive comments from you are just really hard to take. _"I too can sometimes have an opinion (even though I normally suppress it)."_ and many others along those lines. And then there is the flat out condescension: _"do not say the QML interface is poorly thought out, you should argue that the subsurface interface was never designed, but just happened, and with that fundament you expect a simple interface."_ 
I am too busy with my real life to deal with this. We are both too old to see a likelihood of change in this.

Whenever people don't fall in line with your vision we get these comments. And there's not a single other contributor with whom I have this issue. We have some very, very accomplished people working on this projects. And they all manage to get along. I find it really frustrating that I don't appear to be able to find a way to make this work for you as well.

dirkhh:
BTW:
This is not an apology: _"I apologize if you feel offended"_

https://blogs.psychcentral.com/love-matters/2018/06/the-top-12-fake-apologies-and-what-makes-for-an-authentic-apology/

It's not that I feel offended. It's that you offend people. Deliberately and frequently.

janiversen:
No problem, consequence is clear. You are allowed to call things “garbage” and I am not allowed to defend it, even in a polite tone.

dirkhh:
Did you design QML when I wasn't watching? Were you part of the Qt team in the early 2000s?
I mean, I talked a lot to TrollTech and then Nokia in those days but I admit I didn't check if you were an engineer on that team and I just didn't realize.
Otherwise, I don't understand what your comparison even means.
I do not believe I have ever called your code garbage or your design garbage.
I go out of my way not to talk down to you or any other developer, ever.
Your default mode of conversation appears to be a combination of passive aggressive complaining and condescension.

janiversen:
Once we have all of diveplanner settings ready, I will present it as a PR.

bstoeger:
My 2c:

- At the core we are a C project, therefore there are no classes. Nothing is cast in stone, but changing the core to C++ would need some serious discussion out of respect with the developers who are not comfortable with C++. 

- Even if the core were in C++, a class only for enums would appear strange to me.

- Even if the core were in C++, I feel we shouldn't have to make our core classes `Q_OBJECT`s or `Q_GADGET`s.

- QML/Qt's reflection/metatype system is broken in that automatic generation of enum symbols does not work for enums outside of `Q_*` class context.

- Duplicating the enums as two *different* types (outside / inside of class context) is neither simple nor robust. You rely on archaic C-semantics. In C++11-style scoped enums (enum class), you couldn't even assign between these two types without casting.

- QML is the display layer - there should be no significant performance impact for using strings over enums.

- If there is a performance problem with QML, it's probably because it is a weakly typed scripting language - let's move that part to C++.

- Readability of the QML layer is not negatively affected by using strings instead of enums.

- If you want a "self-documenting" API, you still can pass in and out enums by making them `Q_METATYPE`s.

- For many enums we already have functions convert enums from/into strings for load/save. You could adapt those to compare enums in QML [`status_to_string(status) == 'offline'`].

Overall, I see more downsides than upsides to `Q_ENUM` for core-enums. Let's not do it.

janiversen:
The majority of the arguments are actually something I highly agree with, I have made a couple of short comments,
which are intended to be read positive please.

I do not want to lecture anybody or offend anybody, but when looking at the arguments for using strings in QML, I see 2 big issues, thatI do not see reflected in the arguments:

First issue) Using string in QML, makes a type of bug searching a lot harder.
Consider the 2 snippets:
    A === “length” // correct compare
    A === “lenght” // faulty.
Or
    A === Enums.LENGTH // correct compare
    A === Enums.LENGHT // faulty

Using enums secures QML generates an warning/error, with line number etc, so it is very obvious that there are a problem and where.
In case of strings no warnings are generated and you have to first see the problem, then search for it.

Second issue) Using strings nearly doubles the number of getters/setters.

Currently the system works nicely in both C++ and QML with int based getters/setters, if using strings especially qPref* will nearly double the number of getters/setters
(Keeping the int for C++ and implementing string functions for QML).

This is to me a lot of work for really nothing. Why not keep the getters/setters we have and use them. Currently I cannot use a lot of them because the enum values are not available in QML.

Having said this, please remark the PR is withdrawn and will not be resubmitted (also because it was noted that the PR did not contain any user or qml improvements). At a point later a new PR will be offered, with the total dive planner setting, making it easy to see the user benefit as well as the benefit of not writing a lot of new getters/setters.

rgds
Jan I.

> 	• At the core we are a C project, therefore there are no classes. Nothing is cast in stone, but changing the core to C++ would need some serious discussion out of respect with the developers who are not comfortable with C++.

Totally agree, and I for one question if the gain of doing it is higher than the pain to do it.

> 
> 	• Even if the core were in C++, a class only for enums would appear strange to me.

For the C++ part I agree, but it is not that uncommon in backend APIs (at least as a base class).

> 
> 	• Even if the core were in C++, I feel we shouldn't have to make our core classes Q_OBJECTs or Q_GADGETs.

Totally agree.
> 
> 	• QML/Qt's reflection/metatype system is broken in that automatic generation of enum symbols does not work for enums outside of Q_* class context.

It is not only QML/Qt but also Qt. This is a frequent topic on blogs about Qt (and QML). It is a feature highly wished, but seems very hard to implement. Q_NAMESPACE solves part of the problem,
But creates its own set of problems, especially when using a C compiler.

> 
> 	• Duplicating the enums as two different types (outside / inside of class context) is neither simple nor robust. You rely on archaic C-semantics. In C++11-style scoped enums (enum class), you couldn't even assign between these two types without casting.

I do not quite understand, why we continue on this theme. In the original PR I  pointed out the risk and I was told (which I actually agree with) that it was not worth the risk, to change nearly all files.
As a consequence the PR was changed to be only about the C++/QML interface.

> 
> 	• QML is the display layer - there should be no significant performance impact for using strings over enums.
Correct
> 
> 	• If there is a performance problem with QML, it's probably because it is a weakly typed scripting language - let's move that part to C++.
There are no real performance problem relating to this issue. But there might be extra hours spent on bug hunting, due to typo´s
> 
> 	• Readability of the QML layer is not negatively affected by using strings instead of enums.
Correct
> 
> 	• If you want a "self-documenting" API, you still can pass in and out enums by making them Q_METATYPEs.
That is not really a goal for me, my main goal is to the QML to tell me when I have a typo (non existing enum).

> 
> 	• For many enums we already have functions convert enums from/into strings for load/save. You could adapt those to compare enums in QML [status_to_string(status) == 'offline’].
I did not find many of those functions, on the contrary I have found string lists in desktop-widget and the same string list in qPrefUnits.

> 
> Overall, I see more downsides than upsides to Q_ENUM for core-enums. Let's not do it.
> 


Cloud_Status have been duplicated in qPrefCloudStorage, for more than a year, without problems, so why not continue on that.

bstoeger:
I understand the intention, I just don't see the gain. If the error would be detected at compile-time, I could see the point. However, being dynamically typed, you would have to hit the error first in which case the source of the problem should be trivial to find. An additional message in the log will not be crucial in finding the typo. This appears to me putting lipstick on a pig (QML).

If you want strong typing, you have to switch to C++ anyway. Instead of comparing the enum in QML, you could for example export a boolean property that tests the enum. Sure, not ideal, but certainly less convoluted than the type-duplication thing.

janiversen:
Even though we have a farmhouse with production I do not have experience with lipstick on pigs, so I trust your knowledge.

This pig (QML) as you call it, are actually rather good at screaming. The pages are controlled (apart from the script code, which should be a very small part) when being loaded or when being compiled (in this project only active for ios, the PR I made was fast, but does not fully do what I wanted).

But independent, I do not believe I can add further information to this topic. At the end of the day the decision are made by the maintainers and I look forward to see how my work can continue.

PR #2511, opened 2020-01-01
BUG: fix overwritten salinity values

willemferguson:
…/subsurface into salinity1

This hopefully fixes a problem that @dirkhh encountered when running this code, indicated
overwritten salinity values when that was not the case.

Also minor cleanup of whitespace.

Signed-off-by: willemferguson <willemferguson@zoology.up.ac.za>

### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh

PR #2439, opened 2019-12-13
Make salinity data in the Information tab editable. (rebased / merged #2391)

dirkhh:
### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

mainly so that it's easier for @willemferguson to look into the issue mentioned in #2391

PR #2525, opened 2020-01-05
Salinity merge

dirkhh:

- [ ] Bug fix
- [x] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This takes @willemferguson 's old PR, the two "_fixes_" he posted since, plus some of my own changes, rebases it on top of master.
In the process I took a large sharp instrument to the series of commits and massively restructured things to avoid some of the back and forth and to try to make things more logical.
The resulting serious of commits still sucks, there are still issues with the final code and I'm still not at all happy with this, but this cost me about six hours of work to get there so I didn't want to just throw it away.

### Changes made:

1) take Willem's changes to add preference entry, file I/O, UI and rebase them on top of current master and in the process condense them into 4 commits, fixing a few bugs in the process (and possibly introducing a few new ones)
2) add some code cleanups that are unrelated but that were causing warnings and were just ugly
3) clean up the way we deal with dives that simply don't have any salinity information
4) change way we decide to show the warning that dive salinity and dc salinity aren't the same

### Related issues:

#2511
#2439 

### Additional information:

This needs a LOT more testing

### Release note:

missing

### Documentation change:

missing - we need to describe how this works and add screenshots - but so much in the tabs has changed...

### Mentions:

@willemferguson 
but this needs testing from more people. Volunteers, please!


willemferguson:
There are three problems:

1) The simplification of the PR code does not provide for the case where user_salinity is 0; as is typically the case when one opens a dive log where the user has not manually changed the salinity. And when user-salinity != dc-> salinity, the warning icon is shown. Consequently the warning icon is shown with all dives. There is a simple fix to that.

2) This is semantic and one should decide on the approach required. With your latest simplification, the discrepancy between the salinity of salt water (1.025 for libdivecomputer; 1.03 for Subsurface) is not taken into account. The 1.025 value (if my memory is correct from my scratching around in the libdivecomputer code some time ago) comes from libdivecomputer and NOT from the dive computer. The 1.025 value shows for both my Galileo and my Shearwater. Consequently, if one selects "Salt water" from the combobox and the dive computer also shows salt water, then the warning icon is shown because of this discrepancy between 1.03 and 1.025. My opinion is that the warning should not be shown since the appropriate water type is correctly selected.

3) More significantly, the original code switched off the warning icon when a manually-entered dive is being edited. Currently it does not and the warning icon remains on for manually-entered dives. This is not an appropriate arrangement.

It is simple to fix these issues. Would you like me to submit a commit for these?

dirkhh:
re 1.: I saw the warning icon as complaining about dive salinity being not equal to dc salinity. Easily fixed to use the semantics you want
re 2.: there are two problems with your statement. One, the 1025 is not a random value that libdc uses for all salt water. It's the value that comes from the documentation of a couple of dive computers as the value the dc vendor assumes. libdc always tries to use the value the dc has used to convert pressure to depth. And in my case I have dive computers that report 1025, 1030, and 1020.
Additionally, the code that you state fixed that did indeed hide that problem. It also resulted in fresh water and salty water being treated as the same.
Anyway, since we have to trust libdivecomputer (or fix it), I see no point in trying to change the values around. If you want the DC value, use the DC value. If you want a different value (e.g. 1030 instead of 1025) then deal with the fact that you changed the value and accept the warning icon.
re 3.: I don't understand what you are trying to tell me here

I will happily fix what's broken in this PR

willemferguson:
First, thank you for being involved in code that I should have completed. Given your long todo list, I feel privileged to have this much attention.

re 2: I accept your explanation and am more than happy to go along with it. It is rather a corner case and I think that your approach is the most robust. Thank you for looking at that.

re 3. If I go Log-> Add dive -> Information then the salinity is blank as it should be. Now, if I change it to Salt, then the warning icon shows. It shows for any salinity I may choose. For a manually-entered dive it does not make sense to show the warning icon at all because there is no dc->salinity. I think it is useful if the warning icon is disabled for manually-entered dives. Does this make sense at all?

dirkhh:
Thanks for the explanation. Now I get what you meant.
Please check if the commits that I just pushed address your issue.
The first one simply deals with a bug in the code that I had missed before - you always checked `current_dive->dc.salinity` which is the salinity of the first dive computer. What we should compare with is the salinity to of the currently shown dive computer (which in my case is often different because the different dive computers I use use different values for salt water).
The second one tries to address both of the errors in my code that you pointed out. "No user change" and "manual dive".

willemferguson:
Tested the code and threw a few curved balls but everything appears perfect. Thank you so much for our time.

dirkhh:
admittedly this still needs more testing, but it now has a CHANGELOG entry and a simplistic reference in the user manual, so I'll merge as is (no point in waiting for the checks as that last commit just changed text files)

PR #2527, opened 2020-01-06
Cleanups

dirkhh:
Trying to do something productive and have some fun.
What better to do than address random warnings and Coverity reports.
Yes, I know how to throw a party. Envy me.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

loads of small changes. I tried to make sure that I didn't break anything (heck, famous last words) and that things still compiled with Qt 5.9. But that last part is what GitHub Actions are for. The former problem is left as an exercise to the reader.

### Release note:

N/A

### Documentation change:

N/A

### Mentions:

@bstoeger if you are bored... some of these fall into your wheelhouse.

### DIFF
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -469,10 +469,10 @@ ImportDives::ImportDives(struct dive_table *dives, struct trip_table *trips, str
 	// this only matters if undoit were called before redoit
 	currentDive = nullptr;
 
-	struct dive_table dives_to_add = { 0 };
-	struct dive_table dives_to_remove = { 0 };
-	struct trip_table trips_to_add = { 0 };
-	struct dive_site_table sites_to_add = { 0 };
+	struct dive_table dives_to_add = { 0, 0, 0 };
+	struct dive_table dives_to_remove = { 0, 0, 0 };
+	struct trip_table trips_to_add = { 0, 0, 0 };
+	struct dive_site_table sites_to_add = { 0, 0, 0 };
 	process_imported_dives(dives, trips, sites, flags, &dives_to_add, &dives_to_remove, &trips_to_add, &sites_to_add);
 
 	// Add trips to the divesToAdd.trips structure
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -469,10 +469,10 @@ ImportDives::ImportDives(struct dive_table *dives, struct trip_table *trips, str
 	// this only matters if undoit were called before redoit
 	currentDive = nullptr;
 
-	struct dive_table dives_to_add = { 0, 0, 0 };
-	struct dive_table dives_to_remove = { 0, 0, 0 };
-	struct trip_table trips_to_add = { 0, 0, 0 };
-	struct dive_site_table sites_to_add = { 0, 0, 0 };
+	struct dive_table dives_to_add = empty_dive_table;
+	struct dive_table dives_to_remove = empty_dive_table;
+	struct trip_table trips_to_add = empty_trip_table;
+	struct dive_site_table sites_to_add = empty_dive_site_table;
 	process_imported_dives(dives, trips, sites, flags, &dives_to_add, &dives_to_remove, &trips_to_add, &sites_to_add);
 
 	// Add trips to the divesToAdd.trips structure
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -544,7 +544,7 @@ void ImportDives::undoit()
 
 DeleteDive::DeleteDive(const QVector<struct dive*> &divesToDeleteIn)
 {
-	divesToDelete.dives = divesToDeleteIn.toStdVector();
+	divesToDelete.dives = std::vector<dive *>(divesToDeleteIn.begin(), divesToDeleteIn.end());
 	setText(tr("delete %n dive(s)", "", divesToDelete.dives.size()));
 }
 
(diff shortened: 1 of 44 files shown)


commands/command_divelist.cpp
```diff
@@ -469,10 +469,10 @@ ImportDives::ImportDives(struct dive_table *dives, struct trip_table *trips, str
 	// this only matters if undoit were called before redoit
 	currentDive = nullptr;
 
-	struct dive_table dives_to_add = { 0 };
-	struct dive_table dives_to_remove = { 0 };
-	struct trip_table trips_to_add = { 0 };
-	struct dive_site_table sites_to_add = { 0 };
+	struct dive_table dives_to_add = { 0, 0, 0 };
```
bstoeger: This is good - we had that `{ 0 }` construct in some places and I copied it.

Perhaps we should add `empty_*_table` objects so that we can write
`struct dive_table dives_to_add = empty_dive_table`

IIRC it was not completely obvious to make this work with C and C++ at the same time...

desktop-widgets/locationinformation.cpp
```diff
@@ -264,12 +264,12 @@ void LocationInformationWidget::reverseGeocode()
...
 	if (!diveSite || !has_location(&location))
 		return;
-	taxonomy_data taxonomy = { 0 };
+	taxonomy_data taxonomy = { 0, 0 };
 	reverseGeoLookup(location.lat, location.lon, &taxonomy);
 	Command::editDiveSiteTaxonomy(diveSite, taxonomy);
 }
 
-DiveLocationFilterProxyModel::DiveLocationFilterProxyModel(QObject *) : currentLocation({0, 0})
+DiveLocationFilterProxyModel::DiveLocationFilterProxyModel(QObject *) : currentLocation({{0}, {0}})
```
bstoeger: Good.

Here I would also suggest to define an `empty_location` object.

commands/command_edit.cpp
```diff
@@ -910,10 +910,16 @@ void PasteDives::redo()
 	undo();
 }
 
-// ***** Paste *****
+// ***** ReplanDive *****
```
bstoeger, line 920: Good.

commands/command_edit.cpp
```diff
@@ -910,10 +910,16 @@ void PasteDives::redo()
...
 ReplanDive::ReplanDive(dive *source, bool edit_profile) : d(current_dive),
+	when(0),
```
bstoeger, line 922: Not sure about these - I think @atdotde already made these and I objected.

If `d` is NULL, `workToBeDone` returns false and the object will be destroyed immediately. Of course, Coverty isn't that smart.

Up to you.

commands/command_edit.cpp
```diff
@@ -112,7 +112,11 @@ void EditBase<T>::undo()
 
 	// Send signals.
 	DiveField id = fieldId();
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+	emit diveListNotifier.divesChanged(QVector<dive *>(dives.begin(), dives.end()), id);
```
bstoeger, line 116: What? Why is the old method deprecated? Isn't the old one much more readable? Sigh.

commands/command_edit.cpp
```diff
@@ -865,8 +865,7 @@ void PasteState::swap(dive_components what)
 }
 
 // ***** Paste *****
-PasteDives::PasteDives(const dive *data, dive_components whatIn) : what(whatIn),
-	current(current_dive)
+PasteDives::PasteDives(const dive *data, dive_components whatIn) : what(whatIn)
```
bstoeger, line 868: Good.

commands/command_divelist.cpp
```diff
@@ -544,7 +544,7 @@ void ImportDives::undoit()
 
 DeleteDive::DeleteDive(const QVector<struct dive*> &divesToDeleteIn)
 {
-	divesToDelete.dives = divesToDeleteIn.toStdVector();
+	divesToDelete.dives = std::vector<dive *>(divesToDeleteIn.begin(), divesToDeleteIn.end());
```
bstoeger, line 547: Concerning commit message: fools seldom differ. :-D

bstoeger (approved):
LGTM.

My only criticism would be the ReplanDive thing, but if you prefer the full initialization, I'm fine with it.

commands/command_divelist.cpp
bstoeger: Updated previous comment - should of course read "not completely obvious".

commands/command_divelist.cpp
dirkhh: The new syntax is unintuitive and fragile.
But oh well. It's symmetric in both directions and apparently that's how one is supposed to deal with vectors...

commands/command_edit.cpp
dirkhh: I looked at this and went back and forth. It's silly to initialize things.
I could list it as a false positive and ignore it in the Coverity output. In a way, this is the lazy way out...

dirkhh:
I like the idea of creating `const` empty objects that replace all these assignments. 
That shouldn't be too hard, right? I'll play with it :-)

dirkhh:
OK, did a quick one to see what GitHub Actions think about this...
If this looks useful, I'll do the tables and locations as well.

dirkhh:
umm, err, yeah, let's forget about that one. How about that second attempt?

bstoeger:
Fine with me, but then I'm not familiar with idiomatic C. In C++ you would obviously just write a default constructor. Perhaps get @torvalds' opinion? The definition of `empty_cylinder` does look somewhat scary. :)

bstoeger:
PS: That reminds me that I wanted to refactor the table macros. Ooof.

PR #2528, opened 2020-01-09
(part 1) Foundation for additional work.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ x] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR is about to quote @bstoeger (in #2524) “This appears to me putting lipstick on a pig (QML)”. This is the foundation for the Planner module as well as the upcoming settings module.

The reason for continuing with enum (solely for the C++/QML interface, NO influence on the rest of the code) are as follows:

- Using strings are in every aspect just as much “duplicating enums” just more enveloped. An enum value is addressed “meter” instead of METER, and strings provide no extra benefits.
- Using enums gives better (but not perfect) error detection when compiling or loading the qml files.
- Using strings would require a lot of new getters/setters, especially in qPref*, with a lot of compare statements. This seems to be a lot of work, without any benefit to the user experience or the QML code (actually it would divert man days from work that enhance the user experience).
- The int getters/setters are ready, and with this PR, problem is solved, without needing to wait for someone to implement the string functionality.

The PR contains:
Commit 1-5, implement TypeEnum and convert existing duplicated enum (CloudStatus).
Commit 6-, deals with removing the duplicated string setters/getters in qPrefUnit, which proved to be more delicate to remove, than expected, hence the bigger number of commits.

The last 2 commits deal with
- Solving a signal problem
- Removing no longer used code

The series should show both user benefit and improving in the QML code (this was noted to be lacking in the previous commit).

As usual I look forward to receive review comment and correct my mistakes.

### DIFF
--- a/mobile-widgets/CMakeLists.txt
+++ b/mobile-widgets/CMakeLists.txt
@@ -1,6 +1,7 @@
 # mobile backend functions
 
 set(SUBSURFACE_MOBILE_SRCS
+	qmlinterface.cpp
 	qmlmanager.cpp
 	qml/kirigami/src/columnview.cpp
 	qml/kirigami/src/delegaterecycler.cpp
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -13,7 +13,7 @@ Item {
 
 	property string username: login.text;
 	property string password: password.text;
-	property bool showPin: (PrefCloudStorage.cloud_verification_status === CloudStatus.CS_NEED_TO_VERIFY)
+	property bool showPin: (Backend.cloud_verification_status === Enums.CS_NEED_TO_VERIFY)
 
 	ColumnLayout {
 		id: outerLayout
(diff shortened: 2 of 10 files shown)


janiversen:
Test of gitstorage failed, and I cannot see any test failing that relate to this PR.

janiversen:
An example of how the qml Compiler actually catches a lot of stuff:

Error compiling qml file: ../../../mobile-widgets/qml/DivePlannerManager.qml:27:1: error: Syntax error
(from iOS build in Travis). So even though qml is a dynamic language, the compiler makes it a lot more static.

janiversen:
I am VERY pleased to say that we have now a way to have strong enum typing in QML.

This PR is a bit limited (I should have done qPrefUnits) as well, but I wanted to show the progress.

The PR contains:
- the pure interface (QMLInterface) with enum for CloudStatus and units
- PrefCloudStorage.cloud_verification_status changed to Backend.cloud_verification_status
and CloudStatus. changed to Enums.
BUT much more important look at qmlinterface.h the setters and getters do not use int, but a reference to the enums.

To be honest, I am enjoying a glass of wine, because to me this is a major step forward in getting QML more stable. I hope you like it, and look forward to review comments.

dirkhh (approved):
This looks really useful.

I wish we didn't need the duplicate enums, but we have to live with what we can get from C++/QML without massive rewrites everywhere.

I will manually apply this.

PR #2529, opened 2020-01-09
(1) Planner gas, first part

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR is about making the GAS section, behave as it should.

This is first badge, of making planning variables available in QML and updating DivePlannerSetup.
In order to keep the number of commits low, this total is split in 2 PR.

This part concentrates on the simple variables. Using the new QML interface, there are very little
reason to have these in plannerShared.

As usual I look forward to receive review comment and correct my mistakes.

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -214,51 +214,6 @@ void plannerShared::set_bestmixend(int value)
 	CylindersModel::instance()->updateBestMixes();
 }
 
-bool plannerShared::display_runtime()
-{
-	return qPrefDivePlanner::display_runtime();
-}
-void plannerShared::set_display_runtime(bool value)
-{
-	DivePlannerPointsModel::instance()->setDisplayRuntime(value);
-}
-
-bool plannerShared::display_duration()
-{
-	return qPrefDivePlanner::display_duration();
-}
-void plannerShared::set_display_duration(bool value)
-{
-	DivePlannerPointsModel::instance()->setDisplayDuration(value);
-}
-
-bool plannerShared::display_transitions()
-{
-	return qPrefDivePlanner::display_transitions();
-}
-void plannerShared::set_display_transitions(bool value)
-{
-	DivePlannerPointsModel::instance()->setDisplayTransitions(value);
-}
-
-bool plannerShared::verbatim_plan()
-{
-	return qPrefDivePlanner::verbatim_plan();
-}
-void plannerShared::set_verbatim_plan(bool value)
-{
-	DivePlannerPointsModel::instance()->setVerbatim(value);
-}
-
-bool plannerShared::display_variations()
-{
-	return qPrefDivePlanner::display_variations();
-}
-void plannerShared::set_display_variations(bool value)
-{
-	DivePlannerPointsModel::instance()->setDisplayVariations(value);
-}
-
 // Handle when user changes length measurement type
 void plannerShared::unit_lengthChangedSlot(int value)
 {
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -11,11 +11,6 @@ plannerShared *plannerShared::instance()
     static plannerShared *self = new plannerShared;
     return self;
 }
-plannerShared::plannerShared()
-{
-	// Be informed when user switches METER <-> FEET
-	connect(qPrefUnits::instance(), &qPrefUnits::lengthChanged, this, &unit_lengthChangedSlot);
-}
 
 // Planning values
 deco_mode plannerShared::planner_deco_mode()
@@ -212,9 +207,4 @@ void plannerShared::set_bestmixend(int value)
 {
 	qPrefDivePlanner::set_bestmixend(units_to_depth(value).mm);
 	CylindersModel::instance()->updateBestMixes();
-}
-
-// Handle when user changes length measurement type
-void plannerShared::unit_lengthChangedSlot(int value)
-{
 }
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -461,10 +461,10 @@ PlannerSettingsWidget::PlannerSettingsWidget(QWidget *parent, Qt::WindowFlags f)
 
 	connect(ui.lastStop, &QAbstractButton::toggled, plannerShared::instance(), &plannerShared::set_last_stop);
 	connect(ui.lastStop, &QAbstractButton::toggled, this, &PlannerSettingsWidget::disableBackgasBreaks);
-	connect(ui.verbatim_plan, &QAbstractButton::toggled, plannerShared::instance(), &plannerShared::set_verbatim_plan);
-	connect(ui.display_duration, &QAbstractButton::toggled, plannerShared::instance(), &plannerShared::set_display_duration);
-	connect(ui.display_runtime, &QAbstractButton::toggled, plannerShared::instance(), &plannerShared::set_display_runtime);
-	connect(ui.display_transitions, &QAbstractButton::toggled, plannerShared::instance(), &plannerShared::set_display_transitions);
+	connect(ui.verbatim_plan, &QAbstractButton::toggled, plannerModel, &DivePlannerPointsModel::setVerbatim);
+	connect(ui.display_duration, &QAbstractButton::toggled, plannerModel, &DivePlannerPointsModel::setDisplayDuration);
+	connect(ui.display_runtime, &QAbstractButton::toggled, plannerModel, &DivePlannerPointsModel::setDisplayRuntime);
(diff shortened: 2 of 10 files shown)


janiversen:
@atdotde I moved drop_stone_mode to the ratings section, because it seems to work (as I expected it to) by overriding the descent rate, and thus more logical to have those 2 together. Can you please confirm that my assumption is correct. Thanks in advance.

janiversen:
TestGitStorage failed, no other test, so I assume it's ok.

backend-shared/plannershared.cpp
```diff
@@ -199,7 +199,9 @@ void plannerShared::set_min_switch_duration(int value)
 
 double plannerShared::bottomsac()
 {
-	return qPrefDivePlanner::bottomsac() / 1000.0;
+	return (qPrefUnits::volume() == units::LITER) ?
+				qPrefDivePlanner::bottomsac() / 1000.0 :
+				ml_to_cuft(qPrefDivePlanner::bottomsac());
```
dirkhh: I find this an odd choice. We try to do everything in our native units and then do the conversions close to the UI layer. We often use helper functions to do this.
Of course that isn't consistent and the logic has been broken many times all over the code. And the return value here is `double` which isn't implying any unit in the first place. But still, this to me seems unintuitive.

desktop-widgets/diveplanner.cpp
```diff
@@ -559,8 +559,8 @@ void PlannerSettingsWidget::settingsChanged()
 		ui.bottomSAC->setSingleStep(0.1);
 		ui.decoStopSAC->setDecimals(2);
 		ui.decoStopSAC->setSingleStep(0.1);
-		ui.bottomSAC->setValue(ml_to_cuft(prefs.bottomsac));
-		ui.decoStopSAC->setValue(ml_to_cuft(prefs.decosac));
+		ui.bottomSAC->setValue(plannerShared::bottomsac());
+		ui.decoStopSAC->setValue(plannerShared::decosac());
```
dirkhh: case in point - this code is now the same between the `if` clause and the `else` clause. So having it twice is even more confusing.

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -23,8 +23,8 @@ Kirigami.ScrollablePage {
 			spinBestmixend.value = Planner.bestmixend
 		}
 		onVolumeChanged: {
-			spinBottomsac.value = Planner.bottomsac
-			spinDecosac.value = Planner.decosac
+			spinBottomsac.value = Planner.bottomsac * ((PrefUnits.volume === Enums.LITER) ? 1 : 100)
+			spinDecosac.value = Planner.decosac * ((PrefUnits.volume === Enums.LITER) ? 1 : 100)
```
dirkhh: Ugh. So we are getting adjusted values back, but we still end up having to have conditional code here for scaling? That's just strange and ugly. This should return a metric value and we should have a JS function that converts it to and from the display value.

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -265,7 +265,7 @@ Kirigami.ScrollablePage {
 					text: qsTr("SAC factor")
 				}
 				TemplateSpinBox {
-					from: 20
+					from: 1
```
dirkhh: when would it make sense for this factor to be `<1`?

dirkhh (changes requested):
All of this is of course a matter of taste and preferences. I would prefer if we didn't do the unit conversion the way you have implemented it. I'm sure there are other examples in the code, but to me anything that is done in "values" or "numbers" should be done in the scaled SI units that we use all over the code, and that conversions to the unit system the user prefers should be done in the UI layer or if done in helper functions should be delivered as text.

backend-shared/plannershared.cpp
janiversen: Please remember most of this is copied 1-1 from the desktop-widgets/diveplanner. And this is actually very close to the UI layer, unless you want to do it twice once for desktop-widgets and once for mobile-widgets. 
Actually in this particular case it might make sense to move it, because while the desktop UI spin box handles double, QML does not so that value need an extra conversion, just to complicate matters.

desktop-widgets/diveplanner.cpp
janiversen: now this is what I call a good catch, of me sleeping while working !! of course these needs to be moved out after the if/else (that contains other parts).

mobile-widgets/qml/DivePlannerSetup.qml
janiversen: I agree this is ugly, and needs to be corrected !! This is a case where a shared conversion does not help anything at all.

but for the general argument the case for making plannershared.* was to share the conversions between desktop and mobile. If your preference have changed, then we double quite a number of conversions, is that what is wanted ?

mobile-widgets/qml/DivePlannerSetup.qml
janiversen: for reference the value is presented with 1 decimal, so the range on the screen is 0.1 to 9.90.

The desktop allows a range of 2.0 to 10.0.

I have never heard of a negative SAC factor, it would imply you do not consume air, but produce air. Maybe it is something that e.g. @atdotde can tell us.

dirkhh:
I think the key issue for me is one that goes back to other conversations about strings versus integers or more generically strings vs. numbers. In most cases (at least the way things are intended, I should say) the conversions to user visible output (i.e. unit conversions) are done either directly in the UI code or are done in helper functions returning text.
I understand that this may not be the best way to do it, and I acknowledge that this isn't done consistently. But that "thought" or "vision" is what prompted me to comment. I feel weird to have a function `bottomsac()` that returns an untyped value that the consumer has to know is actually in display units (or in one case is NOT in display units and still needs to be scaled by a factor of 100).

It's entirely possible that this is the best way to do it. Then maybe the name of the function should be changed to indicate that `bottomsacDisplay()` or something like that? I don't know.

The following will either make you laugh or cry. At one point when we converted the code to C++ and added a Qt UI I had started to write classes to use for distance, weights and volumes. And those classes internally stored things in our scaled SI units and had setters and getters that allowed you to get and set the value with numbers and strings in either unit. Something like `var.core` and `var.ui`, overloaded to allow all kinds of arguments.
I abandoned the idea after talking to Linus who called me names :-)

I am trusting you that this is the right layer to do this. Let's settle on just having function names that indicate the return value is in UI units or something like that.

mobile-widgets/qml/DivePlannerSetup.qml
dirkhh: still curious about this one because I don't understand the reason for this change... I could see allowing factors `>= 1.0`, but I think I don't understand `>= 0.1`

janiversen:
I am not laughing, I had a developer surprise me once by doing something overnight overnight, and the project was not for quite a time.

bottomSac and decoSac are wrong we should not have double conversion that needs to be corrected.

And I like your idea of using <name>Display to signal the use.

I have added WIP to the title, and will make the changes.

janiversen:
Added a couple of commit at the start to change fixed numbers (like 10 20 etc) in spacing/margin.

janiversen:
I have also updated the calculations so that they are more logical and not in QML.

dirkhh (approved):
I think I'll refactor the commits locally to make this easier to consume, but as far as the code changes go, I will take this

janiversen:
Having read your in the last PR, about collapsing all into 3 commits. I think it is better to close these big PR, and I will do a bit of refactoring and then submit them one by one, because there are some to me important blocks which I would not like to see collapsed as one big commit (but some of the commits can surely be squashed).

I wanted to show the result which was why I pushed all PR. There should not be commits that e.g. adds a function and then a later remove the same function.

l will await your response before doing anything.

dirkhh:
please don't close them. See my response in the other PR.
I meant this as a way for me to understand what you are doing.

But please also read my comments there about the way many open source projects do things. I don't think there should be an assumption that a contributor controls how commits are merged into a project...

dirkhh:
I have merged a somewhat refactored version of this.
The commits now reflect what I would consider reasonable units of change.
If you change something from A to B, this doesn't have a commit that adds B, then a commit that changes the use from A to B, and then a commit that removes A, it just has a commit that changes from A to B which is much easier to look through and understand and appreciate.
Similarly, when you make the same type of change (hard coded number to use of Kirigami.Units) to three different types of value, that is still one consistent change, not three changes.

I understand that all these are matters of taste and opinion. But I feel fairly strongly that what I am doing here is reasonable editorial control as a maintainer of a project.

janiversen:
For what it is worth, you have been doing “reasonable editorial control” of this PR

PR #2530, opened 2020-01-09
(2) Planner planning

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add last part of mobile dive planner Settings.

In order to keep the PR short, this is only the graphics and connection to Backend.

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -105,10 +105,18 @@ void plannerShared::set_decosac(double value)
 
 double plannerShared::sacfactor()
 {
-	return qPrefDivePlanner::sacfactor() / 100.0;
+	return qPrefDivePlanner::sacfactor() /
+#ifdef SUBSURFACE_MOBILE
+			10.0;
+#else
+			100.0;
+#endif
 }
 void plannerShared::set_sacfactor(double value)
 {
+#ifdef SUBSURFACE_MOBILE
+	value /= 10.0;
+#endif
 	// NO conversion, this is done in the planner model.
 	DivePlannerPointsModel::instance()->setSacFactor(value);
 }
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -61,6 +61,17 @@ void plannerShared::set_min_switch_duration(int value)
 	DivePlannerPointsModel::instance()->setMinSwitchDuration(value);
 }
 
+int plannerShared::surface_segment()
+{
+	return qPrefDivePlanner::surface_segment() / 60;
+}
+void plannerShared::set_surface_segment(int value)
+{
+	// NO conversion, this is done in the planner model.
+	DivePlannerPointsModel::instance()->setSurfaceSegment(value);
+	emit instance()->surface_segmentChanged(surface_segment());
+}
+
 double plannerShared::bottomsac()
 {
 	return (qPrefUnits::volume() == units::LITER) ?
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -63,20 +63,42 @@ void plannerShared::set_min_switch_duration(int value)
 
 double plannerShared::bottomsac()
 {
-	return qPrefDivePlanner::bottomsac() / 1000.0;
+	return (qPrefUnits::volume() == units::LITER) ?
+				qPrefDivePlanner::bottomsac() / 1000.0 :
+				ml_to_cuft(qPrefDivePlanner::bottomsac()
+#ifdef SUBSURFACE_MOBILE
+					* 100 // cuft without decimals (0 - 300)
+#endif
+				);
 }
 void plannerShared::set_bottomsac(double value)
 {
+#ifdef SUBSURFACE_MOBILE
+	if (qPrefUnits::volume() == units::CUFT)
+		value /= 100; // cuft without decimals (0 - 300)
+#endif
+
 	// NO conversion, this is done in the planner model.
 	DivePlannerPointsModel::instance()->setBottomSac(value);
 }
 
 double plannerShared::decosac()
 {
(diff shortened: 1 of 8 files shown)


janiversen:
TestGitStorage failed, no other test, so I assume it is ok.

backend-shared/plannershared.cpp
```diff
@@ -74,16 +74,30 @@ void plannerShared::set_descrate(int value)
 }
 
 // Planning values
-deco_mode plannerShared::planner_deco_mode()
+int plannerShared::planner_deco_mode()
```
dirkhh: so this moves the function return value from the `enum` to an `int`
why is that the right thing to do?

dirkhh: or, in other words, isn't this a case where this "unification" makes the core code "less typed"?

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -158,12 +158,16 @@ Kirigami.ScrollablePage {
 				}
 				TemplateComboBox {
 					editable: false
+					currentIndex: Planner.dive_mode === Enums.OC ? 0 :
+								  Planner.dive_mode === Enums.CCR ? 1 : 2
```
dirkhh: so QML cannot assign an enum to current index and vice versa?
That makes me think that `dive_mode` being an enum is NOT an advantage here. It if was an int we could just say `currentIndex: Planner.dive_mode` correct? Likewise the other way around below.

backend-shared/plannershared.cpp
```diff
@@ -100,12 +100,13 @@ void plannerShared::set_dive_mode(int value)
 
 int plannerShared::reserve_gas()
 {
-	return qPrefDivePlanner::reserve_gas();
+	return qPrefUnits::pressure() == units::BAR ? prefs.reserve_gas / 1000 :
+												  mbar_to_PSI(prefs.reserve_gas);
```
dirkhh: again something odd going on with your indentation...

dirkhh (commented):
a few comments / questions

mobile-widgets/qml/DivePlannerSetup.qml
janiversen: yes QML allows currentIndex = Enums.OC

I did it the more complicated way, because I did not want to UI code to impose a specific order in the enum (which you can argue is being too careful)

if Enums.OC changes value, the text in the combobox needs to be chsnged as well.

I will change to use currentIndex = Enum.* I think this happens in a couple of places.

backend-shared/plannershared.cpp
janiversen: I wonder if I have a wrong setting in qtCreator which I normally use for qml files

mobile-widgets/qml/DivePlannerSetup.qml
dirkhh: we have many instances where we have string arrays and corresponding enums that need to stay in sync. All this requires, I think, is a comment that says "hey, this needs to stay in sync with that".

backend-shared/plannershared.cpp
dirkhh: we have sample settings for QtCreator in https://github.com/Subsurface-divelog/subsurface/blob/master/CODINGSTYLE.md#qtcreator 
Those may not be perfect, but they seem to work for me?
(BTW: this is a .cpp file, so maybe this was Xcode?)

backend-shared/plannershared.cpp
janiversen: no it does not affect the core code, it makes it available to QML. Deco_mode is not known to 
QML. The duplicated enum in typeenums.h are available but we do not want to use that in desktop-widgets.

This is btw. a good example where having a pure interface class solely for qml makes a lot of sense. Let me rework foundation1 to make plannerShared type secure with the normal enums and at the same time make the functions available to QML.

janiversen:
Is this a problem with this PR, or something with the android build ? I am a bit confused

dirkhh:
> Is this a problem with this PR, or something with the android build ? I am a bit confused

don't know what you are referring to - but then you force-pushed since writing this.

janiversen:
Yes I found the problem....and I have sort of lost confidence in Travis. The bug was in (1) of the series but not reported until here. In essence I had a Q_PROPERTY(double ...) where the variable was int. Took me a while to find, because I saw (1) - (2) as green and ok, so I only searched in this PR.

Anyhow (1) - (4) are all ready to go, of course depending any review comments.

dirkhh:
> Yes I found the problem....and I have sort of lost confidence in Travis. 

Not to belabor a detail - but you keep referring to Travis which we haven't used in a while. We use GitHub Actions for our CI.

> The bug was in (1) of the series but not reported until here. In essence I had a Q_PROPERTY(double ...) where the variable was int. Took me a while to find, because I saw (1) - (2) as green and ok, so I only searched in this PR.
> 
> Anyhow (1) - (4) are all ready to go, of course depending any review comments.

I'll try to get to this today. Work is keeping me way too busy.

janiversen:
Thanks for correcting me,  it goes to show how little I understand the git world (I have a set of notes with how-to, that is how I survive).

No stress, weekend is around the corner, and with my PR for the build script, I just switch directory to continue working on another part (git worktree, really saves a lot of time switching branches, no need to compile again and again).

janiversen:
squashed a bit and rebase on top of plannerGas2

dirkhh (approved):
Thank you for adapting the granularity of commits.
This is much easier to read for me.
I still don't love the odd "unit-free" scaled double return values, but I'll merge this as is and we'll see if this comes back to cause pain later.

PR #2531, opened 2020-01-11
(3) add design to planner setup and cleanup plannershared

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [x ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Change some of the templates to look better (and be smaller) in order to better fit on the
real estate of a mobile, without considering but also not blocking different sizes on different
devices.

Before image (look at section and spin box):

before

after image

after

The changes are within the limitations of the current subsurfaceTheme (and especially the font
handling). In a later PR subsurfaceTheme will be replaced by one that adjust depending on the
size of the real estate (as well as day/night).

Remark the changes are far more than just colouring, especially the spin box have a tendency to stretch across all the space it thinks are available, which leads to part of the box being off-screen.

Moved some calculations from plannershared to diveplannermodel in order to not split calculations over 2 functions (and files).

### Changes made:

Commit 1-4: Design updates
Commit 5 -: plannershared cleanup

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -89,6 +89,9 @@ void plannerShared::set_bottomsac(double value)
 
 double plannerShared::decosac()
 {
+// Mobile and desktop use the same values when using units::LITER,
+// however when using units::CUFT desktop want 0.00 - 3.00 while
+// mobile wants 0 - 300, therefore multiply by 100 for mobile
 	return (qPrefUnits::volume() == units::LITER) ?
 				qPrefDivePlanner::decosac() / 1000.0 :
 				ml_to_cuft(qPrefDivePlanner::decosac()
@@ -105,6 +108,9 @@ void plannerShared::set_decosac(double value)
 
 double plannerShared::sacfactor()
 {
+// mobile want 0 - 100 which are shown with 1 decimal as 0.0 - 10.0
+// whereas desktop wants 0.0 - 10.0
+// as a consequence divide by 10 or 100
 	return qPrefDivePlanner::sacfactor() /
 #ifdef SUBSURFACE_MOBILE
 			10.0;
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -105,10 +105,18 @@ void plannerShared::set_decosac(double value)
 
 double plannerShared::sacfactor()
 {
-	return qPrefDivePlanner::sacfactor() / 100.0;
+	return qPrefDivePlanner::sacfactor() /
+#ifdef SUBSURFACE_MOBILE
+			10.0;
+#else
+			100.0;
+#endif
 }
 void plannerShared::set_sacfactor(double value)
 {
+#ifdef SUBSURFACE_MOBILE
+	value /= 10.0;
+#endif
 	// NO conversion, this is done in the planner model.
 	DivePlannerPointsModel::instance()->setSacFactor(value);
 }
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -61,6 +61,17 @@ void plannerShared::set_min_switch_duration(int value)
 	DivePlannerPointsModel::instance()->setMinSwitchDuration(value);
 }
 
+int plannerShared::surface_segment()
+{
+	return qPrefDivePlanner::surface_segment() / 60;
+}
+void plannerShared::set_surface_segment(int value)
+{
+	// NO conversion, this is done in the planner model.
+	DivePlannerPointsModel::instance()->setSurfaceSegment(value);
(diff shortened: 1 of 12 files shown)


dirkhh (commented):
this is a PR where a before/after picture would help

dirkhh (commented):
only a small part of this PR is about the design.
And one of the other commits does yet another round of changes to those weird scaling issues we have discussed before.

I think I'll collapse these 40 commits across 4 PRs into something like 3 to be able to see the actual changes that this makes and to understand this as a whole.
Not sure when I'll have the time to do that.

janiversen:
collapse 40 commits into 3 seems very excessive to me and we loose quite a lot of history, that at least are important to me.

I have done quite a job into splitting this into commits that do one thing and not multiple. Surely it can be reduced to something like 30.

In case you really want to reduce it to 3 commits, I think it is better to close all 4 PR´s unmerged.

janiversen:
Look at the existing code to see the same scaling issues. I spent quite sometime getting these moved into simple functions, functions that are shared between the desktop UI and QML, which seems like the obvious thing to do. I think the decision here is whether or not we want to share e.g. these scaling between the desktop or not.

dirkhh:
Hi Jan, I wasn't suggesting that I merge this as 3 commits. And even if that was my plan, I think a response of "then we shouldn't merge it at all" is kinda strange. There is a surprisingly large subset of open source projects that ALWAYS squashes PRs when merging. So many, that this is one of the three options that GitHub offers (merge as branch, rebase and merge (what we do), and squash PR)
You tend to break things down very very aggressively. To the point that I find it distracting and confusing. Especially when in a 40 commit series you add things, then change them, then remove them. Which is reasonable as a "this shows how I made this change", but creates a ton of noise.

To your other comment... please read very carefully what I wrote there. I am not blaming you for the underlying challenge of scaling. I am expressing what I feel when I see the code. One thing that would help me, I think, is if there was a comment for these functions that defined the unit that is used for those un-typed return values. So something like "for metric, this is lightyears per milisecond, for imperial this is miles per century, but on mobile it's inches per minute"

janiversen:
well you did write “ I think I'll collapse these 40 commits across 4 PRs into something like 3 ” which to me reads 40 into 3, sorry if I misunderstood that.

I actually did no write “then we shouldn't merge it at all", I actually wrote close these big PRs and I will refactor another bit (which I agreed to being a good thing) and then resubmit them one by one with the intention of them being easier to review.

you asked me to submit smaller PRs in order to be easier to review, which I have done, but it seems you consider all 4 PRs as one, which is basically the contrary to what you asked of me. But more importantly each PR covers one part of the setup and are in that sense really independent.

May I politely suggest we concentrate on getting (1) in a shape where you agree to merge it, and then continue with (2).

dirkhh:
As usual, communication is hard.

I really appreciate to have these series broken down into parts. Much easier to review.
And I appreciate your systematic approach.
I am a bit worried about the end-point of (1), or actually the second to last commit. Adding a ton of commented out lines to then remove the comments bit by bit... I think that's where your preferences and mine are just fairly different. I like things broken down into bite size chunks, but not into individual molecules. So adding a group of variables and their use to me seems easier to follow.
I think getting just (1) into master is one approach, I think I want to get all of this into master as I know that it's the combination of all four of them really moves things forward.

My goal here is not to make you wait and not to keep telling you do things differently.

Frankly, I'd prefer looking at the code than writing more messages here... because the time that I thought I could use for that today is pretty much gone now

janiversen:
removing the commit that concerns you was on my list, it was needed in order not to forget anything, but then got forgotten.

I will watch and see what you end up merging, important to me is the grouping Gas and Planning, within each of these surely commit can be combined without loosing anything important.

janiversen:
rebased PR. Remark 4 of 7 commits are about design (you wrote "only a small part of this PR is about the design.") and the title reflect 2 purposes design and cleanup, so it looks fairly balanced to me.

janiversen:
once again "testGitStorage", lets see if a restart helps.

janiversen:
and this time it is a kirigami error:

CMake Error at mobile-widgets/CMakeLists.txt:25 (add_library):
  Cannot find source file:
    qml/kirigami/src/columnview.cpp

This passed just moments ago, who does it not pass now (there were absolutely no changes to the comments in the last force push).

I suppose using "Github actions" provides some advantages over travis, but to me it seems, build stability is not one of those advantages.

janiversen:
Added the comments requested, and thus try once again to build.

dirkhh (approved):
very nice
sorry it took so long to get this merged, but I think the result is much improved

dirkhh:
I have no idea why this was showing a conflict - I was easily able to merge it locally, so I did that and pushed the result.

PR #2532, opened 2020-01-13
[RFC]Sorting: use strcoll() instead of Qt-function

bstoeger:
Instead of letting the QSortFilterProxyModel, which is placed
on top of our dive-list models, do the sorting we should keep
our dive list sorted in the core. This will make the code of
the new mobile-models much simpler, as we don't have to invert
the direction.

To do so, we shouldn't use Qt's QString::localeAwareCompare()
function, but rather libc's strcoll(). A further advantage is
that we don't have to transform C-strings into QStrings first.
Thus, this might even be a performance gain.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I'd like to move the sorting of the dive-list models closer to the core. The main reason being that I'd like the lower models already be sorted in the correct direction to avoid complexities in potential rewrites mobile models.

As a first step, I'd like to replace Qt's locale-aware comparison by the libc strcoll(). The problem is that I'm not sure how well this works across different OSs and locales. I _feel_ that it should be reliable today, but who knows. Therefore, I'd appreciate comments from developers with different OS and locales (@neolit123 comes to mind).

An alternative would be the C++ stdlib `std::collate`, but the interface looks more involved than `strcoll()`.

This would avoid unnecessary creation of Qt UTF16 strings.

### Changes made:

1) Replace `QString::localeAwareCompare` by `strcoll`.

### Release note:

No.
### Documentation change:

No.
### Mentions:

@neolit123 @dirkhh

### DIFF
--- a/qt-models/divelocationmodel.cpp
+++ b/qt-models/divelocationmodel.cpp
@@ -198,18 +198,18 @@ bool DiveSiteSortedModel::lessThan(const QModelIndex &i1, const QModelIndex &i2)
 	switch (i1.column()) {
 	case LocationInformationModel::NAME:
 	default:
-		return QString::localeAwareCompare(QString(ds1->name), QString(ds2->name)) < 0; // TODO: avoid copy
+		return strcoll(ds1->name, ds2->name) < 0;
 	case LocationInformationModel::DESCRIPTION: {
-		int cmp = QString::localeAwareCompare(QString(ds1->description), QString(ds2->description)); // TODO: avoid copy
+		int cmp = strcoll(ds1->description, ds2->description);
 		return cmp != 0 ? cmp < 0 :
-		       QString::localeAwareCompare(QString(ds1->name), QString(ds2->name)) < 0; // TODO: avoid copy
+		       strcoll(ds1->name, ds2->name) < 0;
 	}
 	case LocationInformationModel::NUM_DIVES: {
 		int cmp = ds1->dives.nr - ds2->dives.nr;
 		// Since by default nr dives is descending, invert sort direction of names, such that
 		// the names are listed as ascending.
 		return cmp != 0 ? cmp < 0 :
-		       QString::localeAwareCompare(QString(ds1->name), QString(ds2->name)) < 0; // TODO: avoid copy
+		       strcoll(ds1->name, ds2->name) < 0;
 	}
 	}
 }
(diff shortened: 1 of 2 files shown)


neolit123:
> I feel that it should be reliable today, but who knows. Therefore, I'd appreciate comments from developers with different OS and locales (@neolit123 comes to mind).

i need to update to my build to test this change, so not sure when i will be able to do that.
possibly _some_ weekend.

my intuition is telling me that relying on `*strcoll` is not a good idea.
Qt as a framework, gives us a function that should be consistent on all OSes.

if the sorting is moved to the backend the backend needs to be able to call a Qt utility with `QString::localeAwareCompare()`.

> This would avoid unnecessary creation of Qt UTF16 strings.

if `strcoll` is decided as the way forward, for Windows it needs wrappers around  `wcscoll` in `windows.c` because Windows is not UTF-8.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strcoll-wcscoll-mbscoll-strcoll-l-wcscoll-l-mbscoll-l?view=vs-2019

so the UTF-16<-> UTF-8 conversion cannot be avoided on Windows.

dirkhh (commented):
I'm confused by the commit message - because what I see in the commit doesn't seem to match most of the commit message. The commit is just replacing the `QString::localeAwareCompare()` with strcoll - it isn't moving the sorting to the core.
There are several parts here that deserve a comment:
(a) yes, from a performance perspective using a C function to compare C strings is much smarter
(b) the trade-off of "runtime performance" vs. "making sure this does the right thing on all supported platform and OSs"... I know which way I'd lean. If we need to be able to call this code from C, then create a helper function in `qthelper.cpp`
(c) I haven't done an audit on this, so take it with a grain of salt, but at least at many points in the past there have been parts of the core code that relied on the dive list table being sorted ascending by time.  Whether that's the calculations for surface interval between dives, whether it's the deco computations, or whether it's the algorithm that matches dives with GPS fixes. There is a very strong assumption that the dive at `n-1` is the immediately previous dive and the dive at `n+1` is the immediately next dive. It's possible that much of this could be fixed... but the risk of getting something wrong here is huge, IMHO.

So unless I am completely misunderstanding what you are trying to do here, I think this may be a bad idea...

bstoeger:
> I'm confused by the commit message

Yeah, that was more motivation than what the commit actually does.

I'll start with the rework of the sorting and do the performance things later.

I'll close this for now. @neolit123: If it doesn't cost you too much time I'd nevertheless be very interested if/how this works on Windows.

PR #2533, opened 2020-01-13
Very minor table cleanups

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Originally, I wanted to refactor the table code:
1) Only have a single set of functions that takes the size of the object as parameter
2) Generate a number of inline-stubs that passes the correct size to these functions and casts the arguments/results

This turned out to be more tedious than expected owing to the need to pass sorting functions. Therefore I decided that it is not worth the hassle right now. Here is just two very minor cleanups.

### Changes made:

1) Remove unused typedef.
2) Use three-way comparison instead of less-than comparison, since the latter is trivially expressed in terms of the former. This allows removal of a few redundant stubs.

### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/dive.h
+++ b/core/dive.h
@@ -128,10 +128,10 @@ struct divecomputer {
 	struct divecomputer *next;
 };
 
-typedef struct dive_table {
+struct dive_table {
 	int nr, allocated;
 	struct dive **dives;
-} dive_table_t;
+};
 
 struct picture;
 struct dive_site;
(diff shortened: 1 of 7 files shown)


bstoeger:
Oops, appears like we crossed commits. @dirkhh introduced a user of the typedef that I removed. Close for now.

PR #2534, opened 2020-01-13
(Theme 1) base to allow user selectable colours etc.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The theme "thing" is moved to C++ for several reasons:
- let QML do UI and not a lot more
- Prepare for personalized themes (user can select colors freely)
- automatic day/night theme shift.

this PR moves theme colors to C++ in the least invasive manner. subsurfaceTheme still exist in 
main.qml, to cope with fonts, and skeleton color properties (to avoid doing a big search/replace).

### Changes made:

commit 1) remove columnWidth as it is undefined.
commit 2) add themeInterface  in C++
commit 3) register themeInterface, but with a temporary name
commit 4) add missing colors in the different themes
commit 5) add support for changing theme
commit 6) integrate ThemeNew in QML

### Additional information:

What are the reasons to move something that works (main.qml - subsurfaceTheme) to C++, a good question that deserves a longer answer:

Arguments:
1) User should be able to select colours freely, and not just choose between themes. The idea is to keep the themes as "templates" for user selections.

settingsColor

This can be done in QML, but the code will grow and QML is anything but a good environment for implementation code.

1a) day/night shift which is very popular among iOS applications.
Adding an alert in C++ was easy, it surely can be done in QML too but (I think) far less easy.

2) Use should be able to free select font sizes (not just small, normal, large):

settingsFont

This can also be done in QML, but when the font size depend on the available real estate, it becomes complicated in QML.

3) Addon, do not need to be part of theme, it was just easy when developing it. The user should be able to choose language, time format and most importantly how to use '.' and ','. There seems to be a problem with number formatting in QML (it does not understand ES_en), so an added formatter helps doing that.

settingsLanguage

Side note (not relevant for this PR):
The screen shots are not the official subsurface-mobile, but our version. PageStack is long gone and replaced by a toolbar at the bottom, giving immediate access to the important pages, the rest are in globaldrawer. 

4) main.qml is overloaded with code, that would be logically in other places. All reductions of code, make main.qml more easy to understand, and thus maintain.

All of this and more are on my list to be integrated piece by piece (if wanted of course) in the official version.

### DIFF
--- a/mobile-widgets/qml/DiveDetails.qml
+++ b/mobile-widgets/qml/DiveDetails.qml
@@ -53,7 +53,7 @@ Kirigami.Page {
 	property alias cylinderModel4: detailsEdit.cylinderModel4
 
 	title: currentItem && currentItem.modelData && currentItem.modelData.location && "" !== currentItem.modelData.location ?
-		       currentItem.modelData.location : qsTr("Dive details")
+			   currentItem.modelData.location : qsTr("Dive details")
 	state: "view"
 	leftPadding: 0
 	topPadding: Kirigami.Units.gridUnit / 2
@@ -188,8 +188,8 @@ Kirigami.Page {
 
 	actions.main: Kirigami.Action {
 		icon {
-			name: state !== "view" ? ":/icons" + subsurfaceTheme.iconStyle + "/document-save.svg" :
-						 ":/icons" + subsurfaceTheme.iconStyle + "/document-edit.svg"
+			name: state !== "view" ? ThemeNew.iconStyle + "/document-save.svg" :
+						 ThemeNew.iconStyle + "/document-edit.svg"
 			color: subsurfaceTheme.primaryColor
 		}
 		text: state !== "view" ? qsTr("Save edits") : qsTr("Edit dive")
--- a/mobile-widgets/themeinterface.cpp
+++ b/mobile-widgets/themeinterface.cpp
@@ -17,12 +17,57 @@ void themeInterface::setup()
 
 void themeInterface::set_currentTheme(const QString &theme)
 {
+	m_currentTheme = theme;
 	qPrefDisplay::set_theme(m_currentTheme);
+	update_theme();
 	emit currentThemeChanged(theme);
 }
 
 void themeInterface::update_theme()
 {
+	if (m_currentTheme == "Blue") {
+		m_backgroundColor = m_blueBackgroundColor;
+		m_contrastAccentColor = m_blueContrastAccentColor;
+		m_darkerPrimaryColor = m_blueDarkerPrimaryColor;
+		m_darkerPrimaryTextColor = m_blueDarkerPrimaryTextColor;
+		m_drawerColor = m_blueDrawerColor;
+		m_lightDrawerColor = m_blueDrawerColor;
+		m_lightPrimaryColor = m_blueLightPrimaryColor;
+		m_lightPrimaryTextColor = m_blueLightPrimaryTextColor;
+		m_primaryColor = m_bluePrimaryColor;
+		m_primaryTextColor = m_bluePrimaryTextColor;
+		m_secondaryTextColor = m_blueSecondaryTextColor;
+		m_textColor = m_blueTextColor;
+		m_iconStyle = ":/icons";
+	} else if (m_currentTheme == "Pink") {
+		m_backgroundColor = m_pinkBackgroundColor;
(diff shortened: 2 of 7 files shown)


dirkhh (commented):
This is one of those PRs that I don't understand.
So this adds a ton of new properties, another class that is shared (didn't you just open an issue that complained about already having too many of those)?
And most importantly, it seems to only replicate something that we already have in QML in C++.

janiversen:
> This is one of those PRs that I don't understand.

I have tried to make a longer explanation with screen shots in the top part. I believe all could be done in QML, but my preference is really not to have implementations in QML but solely UI.

I was very tempted to add the "theme" to "Backend" when I back ported the lot, but I resisted, because there are no point in being that pure. themeinterface like qmlmanager sits in mobile-widgets and thus are pr definition already backend classes.

janiversen:
I decided to keep the WIP, because there are a couple of things I want to check again, because I might be able to make them smarter.

janiversen:
and 2 times testGitTorage.

dirkhh:
> and 2 times testGitTorage.

Disabled again in  `master` - I truly don't understand why that one randomly fails. And over-proportionally for you.

mobile-widgets/themeinterface.cpp
```diff
@@ -0,0 +1,30 @@
...
+void themeInterface::setup()
+{
+	// get current theme
+	m_currentTheme = qPrefDisplay::theme();
+	update_theme();
+}
+
+void themeInterface::set_currentTheme(const QString &theme)
+{
+	qPrefDisplay::set_theme(m_currentTheme);
```
dirkhh, line 21: is this correct? I would have expected `::set_theme(theme)`

mobile-widgets/themeinterface.cpp
```diff
@@ -0,0 +1,30 @@
...
+	emit currentThemeChanged(theme);
+}
+
+void themeInterface::update_theme()
+{
+	m_contrastAccentColor = "#FF5722";
+	m_lightDrawerColor = "#FFFFFF";
```
dirkhh: those hard coded colors surprise me. If this is correct, I think it deserves a comment.

mobile-widgets/themeinterface.cpp
```diff
@@ -23,8 +23,6 @@ void themeInterface::set_currentTheme(const QString &theme)
 
 void themeInterface::update_theme()
 {
-	m_contrastAccentColor = "#FF5722";
-	m_lightDrawerColor = "#FFFFFF";
```
dirkhh: ah, here the two lines disappear. Easy to refactor.

mobile-widgets/qml/DiveDetails.qml
```diff
@@ -53,7 +53,7 @@ Kirigami.Page {
 	property alias cylinderModel4: detailsEdit.cylinderModel4
 
 	title: currentItem && currentItem.modelData && currentItem.modelData.location && "" !== currentItem.modelData.location ?
-		       currentItem.modelData.location : qsTr("Dive details")
+			   currentItem.modelData.location : qsTr("Dive details")
```
dirkhh, line 56: odd whitespace change?

dirkhh (approved):
Thanks for the much more detailed explanation.
Yes, let's do this.
I will merge locally and simplify the commit sequence - I think at least the first and third commit should be combined into one... see comments

dirkhh:
I apologize if my combining of commits this time was a little more aggressive.
In the end this is always a matter of personal preferences. Here your series of commits would start with code that was arguably wrong, only to be fixed a couple of commits later. So I basically combined three commits into one.

Merged

janiversen:
yes it really was more aggressive, but no problem, as long as it doesn't become a habit :-)

I can see that I actually forgot to update the PR, because the missing colours was already merged, so the PR really deserved to be squashed.

janiversen:
> > and 2 times testGitTorage.
> 
> Disabled again in `master` - I truly don't understand why that one randomly fails. And over-proportionally for you.

The simple reason could be that I push more than most contributors/committers. I actually did some testing a while ago, and tried to write it down see #2538 
Seen from the outside (and as you know with no knowledge about how the internals works) it looks like a "run multiple instances simultaneously" problem. I might be totally wrong, but when I only did one push, wait until test was passed, then another push, it failed a lot less.

PR #2539, opened 2020-01-16
Bugfix: several Pref* registration was registered to wrong object.

janiversen:

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

qPref::registerQML contained, what posible is a paste error:
PrefEquipment, PrefMedia, PrefLog
was all registered to the object qPrefUpdateManager, which surely is wrong.

PrefClearDc was also registered to qPrefUpdateManager but unused.

Settings.qml and main.qml also referenced PrefGeneral.default_cylinder, changed to
PrefEquipment.default_cylinder.

I hope this change is correct (at least the functions work better), because it seems strange not to have been catched in testing or review.

### DIFF
--- a/core/settings/qPref.cpp
+++ b/core/settings/qPref.cpp
@@ -67,10 +67,10 @@ void qPref::registerQML(QQmlEngine *engine)
 		ct->setContextProperty("PrefTechnicalDetails", qPrefTechnicalDetails::instance());
 		ct->setContextProperty("PrefUnits", qPrefUnits::instance());
 		ct->setContextProperty("PrefUpdateManager", qPrefUpdateManager::instance());
-		ct->setContextProperty("PrefEquipment", qPrefUpdateManager::instance());
-		ct->setContextProperty("PrefMedia", qPrefUpdateManager::instance());
+		ct->setContextProperty("PrefEquipment", qPrefEquipment::instance());
+		ct->setContextProperty("PrefMedia", qPrefMedia::instance());
 		ct->setContextProperty("PrefClearDc", qPrefUpdateManager::instance());
-		ct->setContextProperty("PrefLog", qPrefUpdateManager::instance());
+		ct->setContextProperty("PrefLog", qPrefLog::instance());
 	}
 
 	// Register special types
--- a/core/settings/qPref.cpp
+++ b/core/settings/qPref.cpp
@@ -69,7 +69,6 @@ void qPref::registerQML(QQmlEngine *engine)
 		ct->setContextProperty("PrefUpdateManager", qPrefUpdateManager::instance());
 		ct->setContextProperty("PrefEquipment", qPrefEquipment::instance());
 		ct->setContextProperty("PrefMedia", qPrefMedia::instance());
-		ct->setContextProperty("PrefClearDc", qPrefUpdateManager::instance());
 		ct->setContextProperty("PrefLog", qPrefLog::instance());
 	}
 
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -387,7 +387,7 @@ Kirigami.ScrollablePage {
 				inputMethodHints: Qt.ImhNoPredictiveText
 				Layout.fillWidth: true
 				onActivated: {
-					PrefGeneral.default_cylinder = defaultCylinderBox.currentText
+					PrefEquipment.default_cylinder = defaultCylinderBox.currentText
 				}
 			}
 		}
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -150,7 +150,7 @@ Kirigami.ApplicationWindow {
 		detailsWindow.cylinderModel2 = manager.cylinderInit
 		detailsWindow.cylinderModel3 = manager.cylinderInit
 		detailsWindow.cylinderModel4 = manager.cylinderInit
-		detailsWindow.cylinderIndex0 = PrefGeneral.default_cylinder == "" ? -1 : detailsWindow.cylinderModel0.indexOf(PrefGeneral.default_cylinder)
+		detailsWindow.cylinderIndex0 = PrefEquipment.default_cylinder == "" ? -1 : detailsWindow.cylinderModel0.indexOf(PrefEquipment.default_cylinder)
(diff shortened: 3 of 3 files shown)


dirkhh:
So basically, this whole complex mess was broken but it made no difference that it was broken.
And you wonder why so many of us consider the QML/C++ data interface "suboptimal" (to avoid being accused of using negative language).

At some point in the future I'm sure I'll again feel like looking at any of this.

Today is not that day.

janiversen:
Your time is your time, so of course you decide when you do what.

I am simply trying to fix issues, I am being made aware of, and present the fixes here, just as I present new features.

janiversen:
I truly understand why you and others feel like you do, I however see it as a challenge as something to improve, my "offline" work during the last year showed me it can be changed into something a lot cleaner, but far from perfect.

I also think I understand why you write "at some point in the future", at least I can understand if you feel frustrated (not in anyway implying that this is the case). Reviewing hundreds of commits in a suboptimal (I see that word as a positive understatement) part of the code, can hardly be categorized as fun hobby, but more as hard work.

I do strongly believe, that sharing the work I have done during the last year is the right thing to do, instead of keeping it restricted to the little group of local apple users. Remark "sharing" does not mean accepting my work uncritically but opening it for suggestions and alterations (like I did in PR "theme 3", by adding screen shots of the new settings page).

I am a bit frustrated because "At some point in the future" casts a big question mark on whether my work is wanted or merely accepted. Looking back at our discussions (which should not have happened and I apologize for not being more patient and accepting) I see this and "my believe that I can make the interface better", being the center cause, with some added language misunderstandings (both in writing and reading).

I believe it is important that you keep your motivation intact, which to me means you do (mostly) what is fun hobby to you.

I think there are several ways forward, I will leave the PRs sitting for a while, as I work on back porting a new page called "Dive Summary", which is a page containing the information often
requested by the dive centers (together with some numbers to make the user see how he/she progresses).

Just a stray thought nothing more, one way forward could be that you nominate a "junior maintainer" ("junior" meaning you are in charge, but "junior" can offload some of the work) with very limited scope, something like e.g.:
- limited to PRs that target mobile-widgets and backend-shared or limited to the QML/C++ interface and QML.
- All PRs must be public for at least 7 days before being merged
- Any commit that include files outside mobile-widgets and backend-shared, follow the usual way.

That could offload you a bit and keep the desktop version untouched while allowing the mobile to grow more rapidly and keeping you in control (during the 7 days, a single comment from you, would block the PR).

I hope you will read this, in the same positive spirit as it was written. This is meant as a sincere try to resolve an unpleasant situation and there are no intention of "lecturing you" or in any other way provoke more problems. I might unknowingly have used language that due to my lack of understanding or culture differences can be viewed as offensive/aggressive but please accept that is in no way intentional on the contrary.

In the hope that we can continue changing the mobile version into something similar great as the desktop version, I look forward to read your comments.

janiversen:
I have made a proposal to ease the complexity of the C++/QML interface, something which hopefully could make people happier. Please when you feel motivated take a look at #2541 

Thanks in advance
Jan I.

dirkhh:
Thanks for your thoughtful comments. I will comment on #2541 over there. But I want to address a few key themes here:

(a) your work is welcome and wanted.
(b) we have a multi-maintainer model already. Unfortunately a couple of them stepped away from Subsurface development and a couple others prefer not to engage with your PRs, so in the end it's usually me who ends up working on them.
(c) a big part of the frustration for me is the tone of your responses when people disagree with your direction or parts of your implementation. I can argue with @atdotde or @bstoeger for hours about things we disagree about and I never feel talked down to and I never feel insulted. I can't say the same about my interactions with you... I understand and believe you when you say that isn't your intention. And I don't think it makes a lot of sense for me to try to illustrate the point to you - I've gone through that exercise in the past and I don't think it has helped our conversations in any way, shape, or form. I need to simply develop a filter when I read your comments, I guess.

dirkhh (approved):
LGTM

janiversen:
Thanks for your positive reading and comment. I will just comment on one thing, the reason why I keep asking in different forms if my work is really wanted.

You write "I never feel talked down to and I never feel insulted" from others, I do not know if you ever wondered what could be the cause of this unwanted chain reaction, I have. I really feel like the lowest underdog of all in this project, for 2 simple reasons, which far too often puts me in defense mode.

1) my commits are scrutinized down to details like typos in the commit messages and getting review message telling me that I try to smuggle commits in. Comparing that to what I see being merged from others (like allowing a reference to MainWindow in core files, or the cause of this bug fix), makes me feel pretty low life. I have said it before, but it needs to be said again, I really appreciate the positive review messages, that find my stupid bugs or suggest alternative/better ways to solve problems.

2) I often spent hours implementing different solutions (like the Q_ENUM) and continue (with the help of Google) until I have something that actually works, which I then proudly present. It has happened more than once, that a committer have made a google search and then suggest that as a better alternative, stating he has not actually tried it, and you often make that the right solution. Nobody even cares to hear, that many times I have already read the same google pages and actually tried it, which makes me feel my experience is totally worthless.

But maybe I too just need to put a heavy filter on.

Alternatively we both try to make a fresh start, cut each other a bit of slack or at least the benefit of doubt. Is it time to store the axes and light the peace pipe ?

dirkhh:
Our reviews are not consistent and I do a very poor job ensuring that each PR gets the attention it deserves. All of us do this in our spare time and that time is dwindling for many of us, which makes that inconsistency problem worse. I could point you to several recent PRs from other contributors that have received significantly more pushback than yours and have at times been completely rewritten (and I ended up introducing bugs when rewriting). So I don't think your characterization is fully balanced. But I will admit that there are certain contributors whom I trust more than others and whose code I tend to accept without the same level of in depth review that it deserves. And occasionally I pay (or I should say, the project pays) the price for that.

dirkhh:
The only real improvement here is not "less review and scrutiny for everyone" but instead "more consistent review and scrutiny".
And BTW, I still occasionally push my own changes straight into master and I still struggle to get strong review for some of my own PRs, so even with the above in place, that still won't create a truly fair system. But then, this isn't a foundation project with a governing board. This is a scrappy little program that contrary to the fact that it by and large sucks has gotten a few users.

PR #2542, opened 2020-01-18
mobile: small cleanups in main.qml 

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

commit 1:  remove subsurfaceTheme.columnWidth from Log.qml
it was quite confusing that Log.qml referenced a non-existing property.

commit 2: remove M126 warnings in main.qml
I try to remove QML warnings etc. whenever I touch a QML file

commit 3: move initialWidth and initialHeight to rootItem
Trying to clean subsurfaceTheme to what is actually used.

commit 4: secure block of used colors is correct
This is just to avoid confusion, subsurfaceTheme contains a
block of "used colors", but 2 colors were defined outside this block
and a dark theme color was defined outside the dark theme block

commit 5: add missing colors to blue theme
Complete blue with the missing colors, in order to ensure that all
3 themes and used colors contain the same number of colors.

commit 6: add missing colors to pink theme
same as 5 but for pink theme.

commit 7: add missing colors to dark theme
same as 6 but for dark theme. And this theme actually
defined one color more than the other themes.

### DIFF
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -676,6 +676,8 @@ if you have network connectivity and want to sync your data to cloud storage."),
 		property color textColor
 		property color secondaryTextColor
 		property color drawerColor
+		property color contrastAccentColor: "#FF5722" // used for delete button
+		property color lightDrawerColor: "#FFFFFF"
 
 		// colors for the blue theme
 		property color blueDarkerPrimaryColor: "#303F9f"
@@ -709,12 +711,10 @@ if you have network connectivity and want to sync your data to cloud storage."),
 		property color darkBackgroundColor: "#303030"
 		property color darkTextColor: darkPrimaryTextColor
 		property color darkSecondaryTextColor: "#757575"
-
-		property color contrastAccentColor: "#FF5722" // used for delete button
-		property color lightDrawerColor: "#FFFFFF"
 		property color darkDrawerColor: "#424242"
 		property int initialWidth: rootItem.width
 		property int initialHeight: rootItem.height
+
 		Component.onCompleted: {
 			// break the binding
 			initialWidth = initialWidth * 1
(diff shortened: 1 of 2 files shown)


mobile-widgets/qml/main.qml
```diff
@@ -737,12 +735,12 @@ if you have network connectivity and want to sync your data to cloud storage."),
 
 	onWidthChanged: {
 		manager.appendTextToLog("Window width changed to " + width + " orientation " + Screen.primaryOrientation)
-		if (subsurfaceTheme.initialWidth !== undefined) {
-			if (width !== subsurfaceTheme.initialWidth && rootItem.firstChange) {
+		if (initialWidth !== undefined) {
+			if (width !== initialWidth && rootItem.firstChange) {
```
dirkhh: so it is entirely possible that I am full of nonsense here, but IIRC the reason this was done in the horrible ugly way it was done was to ensure that things weren't set too early. There is this very odd thing that is happening on some devices where the width of the screen changes on us about half way through the initialization phase. And it took me several weeks of testing with all of my devices and with the help of several Subsurface-mobile users to come up with a sequence of steps that appeared to be completely brain-dead, but that lead to us having the correct width on all of the various phones (I seem to remember that this was mostly Android devices, but I'm no longer sure).

So while I think that on the surface this sounds logical, I am very worried about breaking something that took me so much time and effort to get right.

dirkhh (changes requested):
If I could talk you into dropping that one commit that I fear will cause a lot of re-testing and possibly break the screen layout on some weird Android devices, the rest I think is all good to go.

mobile-widgets/qml/main.qml
janiversen: I will drop that commit and force push a bit later

janiversen: However it is really weird that initialWidth is broken but initialHeight is not...especially because there no explanation.

dirkhh: Really weird does not begin to cover it. This whole thing was driving me nuts.
IIRC I only track the width because they always changed together, and all I try to track with this is "when do we have the 'real' starting dimensions so that we can set the `gridUnit` and the `columnWidth` and all that".

It's entirely possible (even likely) that I'm just doing something fundamentally wrong here. But I'm sure you can see that I am rather nervous to risk having to go all of this again...

janiversen:
I can follow your arguments, some day we might find the problem, but as you like to state "today is not the day". Just force pushed an update without this problematic commit.

dirkhh (approved):
thank you

PR #2543, opened 2020-01-19
(1) last part of foundation (making qPrefUnit look "normal").

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add qPrefUnit to the new pure interface, and then remove string functions.

I am sorry, that I did it in small steps (commits), but the small differences in each change, made me not only build but also run "make check". The commits can be squashed, if you want, but for bisecting have them separate makes kind of sense.

Next up, before rebasing the lot of WIP, is to change names in plannerShared as you suggested. Now that have a pure interface for QML, I will also start using that.

### Changes made:

commit 1: allow qPrefUnit strongly typed in QML
commit 2: use strongly typed units in QML
commit 3: allow signals to be strongly typed in C++
- Remark, this is something new, not only cleanup, the old (including very old) used int !
commit 4: change qPrefUnits::unit_system() to strongly typed and no strings
commit 5: remove string func. for duration_units
commit 6: remove string func. for length
commit 7: remove string func. for pressure
commit 8: remove string func. for temperature
commit 9: remove string func. for vertical_speed_time
commit 10: remove string func. for volume
commit 11: correct signal error.
- this surely must have caused spurious errors at least in the mobile.
commit 12: removed not needed if.

### DIFF
--- a/mobile-widgets/qmlinterface.cpp
+++ b/mobile-widgets/qmlinterface.cpp
@@ -20,4 +20,20 @@ void QMLInterface::setup(QQmlContext *ct)
 	// relink signals to QML
 	connect(qPrefCloudStorage::instance(), &qPrefCloudStorage::cloud_verification_statusChanged,
 		[=] (int value) { emit instance()->cloud_verification_statusChanged(CLOUD_STATUS(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::duration_unitsChanged,
+			[=] (int value) { emit instance()->duration_unitsChanged(DURATION(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::lengthChanged,
+			[=] (int value) { emit instance()->lengthChanged(LENGTH(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::pressureChanged,
+			[=] (int value) { emit instance()->pressureChanged(PRESSURE(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::temperatureChanged,
+			[=] (int value) { emit instance()->temperatureChanged(TEMPERATURE(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::unit_systemChanged,
+			[=] (int value) { emit instance()->unit_systemChanged(UNIT_SYSTEM(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::vertical_speed_timeChanged,
+			[=] (int value) { emit instance()->vertical_speed_timeChanged(TIME(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::volumeChanged,
+			[=] (int value) { emit instance()->volumeChanged(VOLUME(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::weightChanged,
+			[=] (int value) { emit instance()->weightChanged(WEIGHT(value)); });
 }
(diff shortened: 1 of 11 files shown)


dirkhh:
You have so many PRs in flight... I think I lost track of the intended ordering... the name of this one suggests that I should look at this next?

mobile-widgets/qmlinterface.cpp
```diff
@@ -20,4 +20,20 @@ void QMLInterface::setup(QQmlContext *ct)
 	// relink signals to QML
 	connect(qPrefCloudStorage::instance(), &qPrefCloudStorage::cloud_verification_statusChanged,
 		[=] (int value) { emit instance()->cloud_verification_statusChanged(CLOUD_STATUS(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::duration_unitsChanged,
+			[=] (int value) { emit instance()->duration_unitsChanged(DURATION(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::lengthChanged,
+			[=] (int value) { emit instance()->lengthChanged(LENGTH(value)); });
```
dirkhh, line 26: I sense inconsistency here - and it seems that we already had this before...
`duration_unitsChanged()` vs `lengthChanged()`
and reading through the code it's even more confusing. The one that is marked as `...unitsChanged` is one that in fact changes a display preference (always show hours), and the others are the ones that really change units.

Not pretty. Actually, rather confusing. Bad maintainer / review work back then.

Out of scope for this PR for sure.

mobile-widgets/qmlinterface.h
```diff
@@ -102,12 +111,36 @@ class QMLInterface : public QObject {
...
 public slots:
 	void set_cloud_verification_status(CLOUD_STATUS value) {  qPrefCloudStorage::set_cloud_verification_status(value); }
+	void set_duration_units(DURATION value) { qPrefUnits::set_duration_units((units::DURATION)value); }
+	void set_length(LENGTH value) { qPrefUnits::set_length((units::LENGTH)value); }
+	void set_pressure(PRESSURE value) { qPrefUnits::set_pressure((units::PRESSURE)value); }
+	void set_temperature(TEMPERATURE value) { qPrefUnits::set_temperature((units::TEMPERATURE)value); }
+	void set_unit_system(UNIT_SYSTEM value) { qPrefUnits::set_unit_system((unit_system_values)value); }
+	void set_vertical_speed_time(TIME value) { qPrefUnits::set_vertical_speed_time((units::TIME)value); }
+	void set_volume(VOLUME value) { qPrefUnits::set_volume((units::VOLUME)value); }
+	void set_weight(WEIGHT value) { qPrefUnits::set_weight((units::WEIGHT)value); }
```
dirkhh, line 132: yeah, so here we see this in use... and I guess it's not as bad as it's used as `...units...set_weight()` so the `units` part may have felt repetitive. 
So the only one that might eventually deserve being renamed is `duration_units` into something like `duration_always_hours`

dirkhh (approved):
I think this is ok. I need to play with it locally. And I think it will be easier to understand if I combine the commits into one where all the conversions from string happen. I'll play with it and merge soon.

dirkhh:
ok, this one has been refactored (with no code changes whatsoever - but to me the result is a lot easier to read and understand) and merged

PR #2545, opened 2020-01-19
Desktop: Import dive coordinates directly from GPS

dirkhh:
This allows Subsurface to obtain the coordinates of a dive directly
from a GPS track. It parses a GPX file (GPX V1.0 or V1.1) from
a GPS to locate the trackpoint immediatedly after the start of a
dive. There is an additional "Use GPS file" button in the Edit Dive
Site panel that is selected from the Notes tab. Image:

This allows one to select a GPX file, bringing up the Import GPS
dialog.

There is extensive provision for cross-checking that the dive track
synchronises with the dive start and end. If the Save button in the
dialog is pressed the dive coordinates are copied into the Dive
Coordinates text box in the Edit Dive Site panel. The map moves
to indicate the location of the dive site.

The bulk of the work is done in importgps.cpp. The code is
pretty intergrated: I tried to break it up in smaller commits but that
was not feasible.

The code includes responses to the comments by @neolit123 and
@bstoeger. The C-based file input was replaced with Qt-based
code using QChar, QString and QFile.

[Dirk Hohndel: fixed several small issues in the .ui file, removed
               various headers includes that weren't needed and
               fixed printing of minutes as zero padded]

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is my refactor of #2469 

### Changes made:

squashed all the commits
removed several `#include` that weren't needed
dealt with copy paste errors in `.ui` file
added code to correctly zero-pad minutes
added code not to call the importer if no file name was given

### Release note:

still needed

### Documentation change:

still needed

### Mentions:

@willemferguson


desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,294 @@
...
+//  Read text from the present position in the file until
+//  the first 'delim' character is encountered.
+int ImportGPS::getSubstring(QFile *file, QString *bufptr, char delim)
+{
+	char c;
+	bufptr->clear();
+	do {
+		if (file->read(&c, 1) <= 0) // EOF
+			return 1;
+		if ( c == delim) break;
```
neolit123: extra space after `(`

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,294 @@
...
+		bufptr->append(QChar(c));
+	} while (c != delim);
+	return 0;
+}
+
+// Find the next occurence of a specified target GPX element in the file,
+// characerised by a "<xxx " or "<xxx>" character sequence.
+// 'target' specifies the name of the element searched for.
+// termc is the ending character of the element name search = '>' or ' '.
+int ImportGPS::findXmlElement(QFile *fileptr, QString target, QString *bufptr, char termc)
```
neolit123: a nice to have would be to have unit tests for the XML parsing logic here.
but as someone commented here https://github.com/Subsurface-divelog/subsurface/pull/2469#discussion_r368404123
i think this logic can be replaced with Qt's XML objects.

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,294 @@
...
+		tm1.tm_mday = buf.mid(8,2).toInt(&ok,10);
+		tm1.tm_hour = buf.mid(11,2).toInt(&ok,10);
+		tm1.tm_min  = buf.mid(14,2).toInt(&ok,10);
+		tm1.tm_sec  = buf.mid(17,2).toInt(&ok,10);
+		when = utc_mktime(&tm1) + time_offset;
+		if (first_line) {
+			first_line = false;
+			coords.start_track = when;   // Local time of start of GPS track
+		}
+		if ((when > divetime) && (found == false)) {   // This GPS local time corresponds to the start time of the dive
```
neolit123: can be `(when > divetime && found == false)`

desktop-widgets/importgps.cpp
```diff
@@ -0,0 +1,294 @@
...
+	int dyr,dmon,dday,dhr,dmin;
+	dyr = time.tm_year;
+	dmon = time.tm_mon;
+	dday = time.tm_mday;
+	dhr = time.tm_hour;
+	dmin = time.tm_min;
+#endif
+
+	do {
+		line++;  // this is the sequence number of the trkpt xml element processed
```
neolit123: this is again feels like doing some XML parsing that can be done with Qt objects.

neolit123 (commented):
had another pass at the diff.

i guess these are comments for @willemferguson 

@willemferguson a trick to get this latest patch is the following:
```
# pull the changes from this PR in a local patch
wget https://github.com/Subsurface-divelog/subsurface/pull/2545.patch
# you can then checkout your local branch "importgps1"
# remove you last few commits with a tool like "gitk" or "git reset --hard HEAD~N"
# and apply this patch using
git am 2545.patch

# you can then force push the local change to your remote branch
# which will update your other PR
git push -f <remote-name> importgps1
```

desktop-widgets/importgps.cpp
willemferguson: I agree. But let's just get this code into master first to avoid the need for another rebase. Then we can add the trimmings. There is also some work to do with respect to the default directory with gpx files.

dirkhh:
I added a commit to move the GPX parsing code into core.
I still don't have a dive XML plus corresponding GPX file, so I still can't test that this actually works.
once I have that I can add a test case and after that maybe we can move this to no longer hand-parse XML...

PR #2547, opened 2020-01-21
(1) Rest of Gas variables.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Reduced plannerShared by removing simple variables,
added all variables to QMLInterfaces
update calculations in plannerShared where needed.

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -105,10 +105,18 @@ void plannerShared::set_decosac(double value)
 
 double plannerShared::sacfactor()
 {
-	return qPrefDivePlanner::sacfactor() / 100.0;
+	return qPrefDivePlanner::sacfactor() /
+#ifdef SUBSURFACE_MOBILE
+			10.0;
+#else
+			100.0;
+#endif
 }
 void plannerShared::set_sacfactor(double value)
 {
+#ifdef SUBSURFACE_MOBILE
+	value /= 10.0;
+#endif
 	// NO conversion, this is done in the planner model.
 	DivePlannerPointsModel::instance()->setSacFactor(value);
 }
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -63,20 +63,42 @@ void plannerShared::set_min_switch_duration(int value)
 
 double plannerShared::bottomsac()
 {
-	return qPrefDivePlanner::bottomsac() / 1000.0;
+	return (qPrefUnits::volume() == units::LITER) ?
+				qPrefDivePlanner::bottomsac() / 1000.0 :
+				ml_to_cuft(qPrefDivePlanner::bottomsac()
+#ifdef SUBSURFACE_MOBILE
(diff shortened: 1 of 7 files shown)


backend-shared/plannershared.cpp
```diff
@@ -63,20 +63,42 @@ void plannerShared::set_min_switch_duration(int value)
 
 double plannerShared::bottomsac()
 {
-	return qPrefDivePlanner::bottomsac() / 1000.0;
+	return (qPrefUnits::volume() == units::LITER) ?
+				qPrefDivePlanner::bottomsac() / 1000.0 :
+				ml_to_cuft(qPrefDivePlanner::bottomsac()
+#ifdef SUBSURFACE_MOBILE
+					* 100 // cuft without decimals (0 - 300)
+#endif
```
dirkhh, line 71: so we are returning an un-typed floating point number, with different scaling depending on the UI.

I know we talked about this before - the idea that we prepare UI text in shared code.

Hmm - trying to figure out where to take this comment.

I guess all I want to express is that I struggle with this. It feels wrong. But I don't know if I should push for something else. And what exactly to push for.

backend-shared/plannershared.cpp
```diff
@@ -115,10 +115,18 @@ void plannerShared::set_problemsolvingtime(int value)
 
 double plannerShared::sacfactor()
 {
-	return qPrefDivePlanner::sacfactor() / 100.0;
+	return qPrefDivePlanner::sacfactor() /
+#ifdef SUBSURFACE_MOBILE
+			10.0;
+#else
+			100.0;
+#endif
```
dirkhh, line 113: this feels equally strange.

dirkhh (commented):
the flow in the two parts is broken in to way too many small commits (I never thought I'd say that) and it seems that we add things here, then change them later, then remove them... I'm sure that's not what's actually happening, but I struggle with reviewing this as a whole.

Speaking of struggle - see my comments on those un-typed, scaled floating point return values.
Those make me really uncomfortable. But I haven't figured out if I want something else instead. And what it is that I would want.

Once again, I need to find the time to merge this locally, refactor the commits so they make more sense to me, and then look at the resulting code and how it is used both from desktop and mobile.

@atdotde - do you have thoughts about this, since the planner has mostly been your baby?

janiversen:
Updated to contain fewer commits, but of course each more complicated.

dirkhh:
I'm now not sure where this goes or if I maybe merged the wrong PRs.
Sorry if I got confused and got this wrong. I'll stop here and wait for your comments where to continue. Thanks

janiversen:
There was a reason why they were numbered 1-2-3 :-) I think you already merged these commits, let me check.

janiversen:
It is merged with either (2) or (3).

next up is “theme 1”. but that is a different topic and might need more conversation.

PR #2549, opened 2020-01-22
(anytime) build-system: add support "git worktree"

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add option "-src-dir <source directory>" to build options.

When calling build.sh without -src-dir it uses src/subsurface as usual, but when called
with -src-dir <source directory>, it uses src/<source directory> as source basis.

This is a needed option, when working with "git worktree", which is used when working on
different branches in parallel (e.g. master and my-feature-branch), because it allows a
build directory in each worktree, and thus much faster when switching work.

### Changes made:

1) add variable SRC_DIR, with default value "subsurface"
2) replace "subsurface" with SRC_DIR in whole script
3) add option -src-dir
4) expand usage explanation.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -47,6 +47,7 @@ PLATFORM=$(uname)
 
 BTSUPPORT="ON"
 DEBUGRELEASE="Debug"
+SRC_DIR="subsurface"
 
 # deal with all the command line arguments
 while [[ $# -gt 0 ]] ; do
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
makes sense

PR #2558, opened 2020-01-26
Update to new libdivecomputer version

torvalds:
Jef has changed the libdivecomputer iostream layer and extended it in
two different ways:

 - iostram's now have a 'poll()' method, which does what the name
   implies: waits for data to be available with a timeout.

 - iostreams now have a 'ioctl()' method, which can be used to implement
   miscellaneous operations. Right now the two ones that you can do are
   "set latency" (this replaces the old 'set_latency()' method) and "get
   BLE name" (this replaces our 'get_name()' method that was never part
   of the upstream libdivecomputer interfaces)

Neither of these is all that complicated, and the transition is fairly
obvious.

HOWEVER.

I have absolutely no idea how to do 'poll()' on Windows sockets, and I
have no intention of figuring it out.  We use a direct socket interface
to implement the (non-BLE) RFCOMM bluetooth serial protocol, and I'm not
sure why Windows is so special here.  I suspect - but cannot test - that
we should just switch the Windows RFCOMM implementation over to the use
the same QtBluetooth code that we use on other platforms.

I assume that the Windows Bluetooth support was originally not
sufficiently good for that, but these days we depend on Qt doing BLE for
us even on Windows, so presumably FRCOMM works too.

That would be a nice cleanup, and would make 'poll()' work on RFCOMM
under Windows too.  However, since I can't test it, I've not done that,
but instead just made the Windows RFCOMM 'poll()' method always return
success.  That may or may not get the thing limping along.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/qtserialbluetooth.cpp
+++ b/core/qtserialbluetooth.cpp
@@ -12,6 +12,7 @@
 #include <libdivecomputer/version.h>
 #include <libdivecomputer/context.h>
 #include <libdivecomputer/custom.h>
+#include <libdivecomputer/serial.h>
 
 #if defined(Q_OS_WIN)
 	#include <winsock2.h>
@@ -302,6 +303,33 @@ static dc_status_t qt_serial_write(void *io, const void* data, size_t size, size
 	return DC_STATUS_SUCCESS;
 }
 
+static dc_status_t qt_serial_poll(void *io, int timeout)
+{
+	qt_serial_t *device = (qt_serial_t*) io;
+
+	if (!device)
+		return DC_STATUS_INVALIDARGS;
+
+#if defined(Q_OS_WIN)
+	// FIXME FIXME FIXME!! But how ?
+	// I have no idea about windows socket programming - Linus
+	// We'll just pretend it's always readable, and hope for the best
+	// Why is the windows side not using QBluetoothSocket?
+	return DC_STATUS_SUCCESS;
+#else
+	if (!device->socket)
+		return DC_STATUS_INVALIDARGS;
+	if (device->socket->waitForReadyRead(timeout))
+		return DC_STATUS_SUCCESS;
+	return DC_STATUS_TIMEOUT;
+#endif
+}
+
+static dc_status_t qt_serial_ioctl(void *io, unsigned int request, void *data, size_t size)
+{
+	return DC_STATUS_UNSUPPORTED;
+}
+
 static dc_status_t qt_serial_purge(void *io, dc_direction_t)
 {
 	qt_serial_t *device = (qt_serial_t*) io;
@@ -385,20 +413,20 @@ ble_packet_open(dc_iostream_t **iostream, dc_context_t *context, const char* dev
 
 	static const dc_custom_cbs_t callbacks = {
 		qt_ble_set_timeout, /* set_timeout */
-		NULL, /* set_latency */
 		NULL, /* set_break */
 		NULL, /* set_dtr */
 		NULL, /* set_rts */
 		NULL, /* get_lines */
 		NULL, /* get_received */
 		NULL, /* configure */
+		qt_ble_poll, /* poll */
 		qt_ble_read, /* read */
 		qt_ble_write, /* write */
+		qt_ble_ioctl, /* ioctl */
 		NULL, /* flush */
 		NULL, /* purge */
 		qt_custom_sleep, /* sleep */
 		qt_ble_close, /* close */
-		qt_ble_get_name, /* get_name */
 	};
 
 	rc = qt_ble_open(&io, context, devaddr, (dc_user_device_t *) userdata);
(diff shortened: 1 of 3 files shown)


core/qtserialbluetooth.cpp
```diff
@@ -302,6 +303,33 @@ static dc_status_t qt_serial_write(void *io, const void* data, size_t size, size
...
+
+	if (!device)
+		return DC_STATUS_INVALIDARGS;
+
+#if defined(Q_OS_WIN)
+	// FIXME FIXME FIXME!! But how ?
+	// I have no idea about windows socket programming - Linus
+	// We'll just pretend it's always readable, and hope for the best
+	// Why is the windows side not using QBluetoothSocket?
+	return DC_STATUS_SUCCESS;
```
neolit123, line 318: i do not remember the reason why we are not using QBluetoothSocket on Windows.
EDIT: i think the Qt BT classic backend was simply not ready yet.

this is the native Windows way:
https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsapoll

```
WSAPOLLFD fdarray = { 0 };
fdarray.fd = device->socket;
fdarray.events = POLLRDNORM;
	
if (WSAPoll(&fdarray, 1, timeout) == SOCKET_ERROR) {
	qDebug() << "Error polling socket:" << WSAGetLastError();
	return DC_STATUS_TIMEOUT;
}
return DC_STATUS_SUCCESS;
```

WSAPoll needs Vista or later. only compile tested with Mingw 5.3*.

looks like Subsurface is already linking ws2_32:
https://github.com/Subsurface-divelog/subsurface/blob/master/CMakeLists.txt#L236
which includes WSAPoll.

but above this line:
https://github.com/Subsurface-divelog/subsurface/blob/df5040c6aec41fb324a9712f8da8b0b079dd09aa/core/qtserialbluetooth.cpp#L18

the following must be added.
```
#ifndef _WIN32_WINNT
#define _WIN32_WINNT 0x0600
#endif
```
maybe some of the Qt includes is already doing that.

dirkhh: Frankly, I have no idea. All of the Windows code has been done by others. I'm happy to get patches / pull requests from people who either know or have the time to figure these things out.
If this waits until I find the time and opportunity to work on this with my Windows Laptop (which I haven't booted in maybe half a year) then I guess we're screwed.

torvalds: Is there somebody who can test? Either Lubomir's "implement poll()" or just the "switch over to QtBluetooth on Windows too"?

I'd actually prefer (strongly) to try to switch Windows over to QtBluetooth, because it would clean that file up enormously and we'd have the same code-base. But I obviously cannot test it..

dirkhh: @janmulder I think you tested BT stuff on Windows in the path? Or am I confused?
@neolit123 has only BLE devices, I think.
See my comment above about my ability to test things.

One thing that might work is to create a pull request with the code that you want tested and then send email to the developer mailing list - we have several hundred people on that list. Someone ought to have a windows box and a BT dive computer

neolit123:
> @neolit123 has only BLE devices, I think.

i only have Perdix AI, which AFAIK is BTLE only and this code path will not be executed on it?

> One thing that might work is to create a pull request with the code that you want tested and then send email to the developer mailing list - we have several hundred people on that list. Someone ought to have a windows box and a BT dive computer

we had success with that in the past.
with the new GitHub actions changes i'm not sure if possible to download the installer from a CI build:
https://github.com/Subsurface-divelog/subsurface/pull/2558/checks?check_run_id=409801119
with Travis it was possible to get the download URL from the logs.

core/qtserialbluetooth.cpp
neolit123: > @neolit123 has only BLE devices, I think.

i only have Perdix AI, which AFAIK is BTLE only and this code path will not be executed on it?

> One thing that might work is to create a pull request with the code that you want tested and then send email to the developer mailing list - we have several hundred people on that list. Someone ought to have a windows box and a BT dive computer

we had success with that in the past.
with the new GitHub actions changes i'm not sure if possible to download the installer from a CI build:
https://github.com/Subsurface-divelog/subsurface/pull/2558/checks?check_run_id=409801119
with Travis it was possible to get the download URL from the logs.

torvalds: Here's the "enhanced" pull request with this same base, but with the addition of windows direct socket code removal.

https://github.com/Subsurface-divelog/subsurface/pull/2559

Whether it works or not, who knows...

dirkhh:
Yes, you can download the installer. 
In this case they are here:
https://github.com/Subsurface-divelog/subsurface/actions/runs/31632044

(you can click on Actions on GitHub, then find the build for the right branch, click on Windows... there you go :-) )

dirkhh:
merged the other one, so this one should be redundant, right?

glance-:
Yea, the old windows code written by Claudiu back in GSOC 2015 was because QtBluetooth didn't support windows back then. If QtBluetooth works good enough on windows nowadays It's way better to use ut so we don't need the windows special case code any longer.

core/qtserialbluetooth.cpp
zzz6519003: add test

PR #2559, opened 2020-01-27
Update libdivecomputer and make Windows use qtbluetooth

torvalds:
This is the same old "update to new libdivecomputer version", with the added twist that it also has a commit to make the Windows rfcomm code use QtBluetooth like all other platforms.

That's a nice cleanup, but I don't know if modern Qt is good enough on Windows these days..

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/qt-ble.h
+++ b/core/qt-ble.h
@@ -23,7 +23,13 @@ public:
 	inline void set_timeout(int value) { timeout = value; }
 	dc_status_t write(const void* data, size_t size, size_t *actual);
 	dc_status_t read(void* data, size_t size, size_t *actual);
-	inline const char *get_name() { return device->btname; }
+	inline dc_status_t get_name(char *res, size_t size)
+	{
+		if (!device->btname) return DC_STATUS_UNSUPPORTED;
+		strncpy(res, device->btname, size);
+		return DC_STATUS_SUCCESS;
+	}
+	dc_status_t poll(int timeout);
 
 	inline QLowEnergyService *preferredService() { return preferred; }
 	inline int descriptorWritten() { return desc_written; }
@@ -61,10 +67,11 @@ private:
 extern "C" {
 dc_status_t qt_ble_open(void **io, dc_context_t *context, const char *devaddr, dc_user_device_t *user_device);
 dc_status_t qt_ble_set_timeout(void *io, int timeout);
+dc_status_t qt_ble_poll(void *io, int timeout);
 dc_status_t qt_ble_read(void *io, void* data, size_t size, size_t *actual);
 dc_status_t qt_ble_write(void *io, const void* data, size_t size, size_t *actual);
+dc_status_t qt_ble_ioctl(void *io, unsigned int request, void *data, size_t size);
 dc_status_t qt_ble_close(void *io);
-const char *qt_ble_get_name(void *io);
 }
 
 #endif
--- a/core/qtserialbluetooth.cpp
+++ b/core/qtserialbluetooth.cpp
@@ -12,6 +12,7 @@
 #include <libdivecomputer/version.h>
 #include <libdivecomputer/context.h>
 #include <libdivecomputer/custom.h>
+#include <libdivecomputer/serial.h>
 
 #if defined(Q_OS_WIN)
 	#include <winsock2.h>
@@ -302,6 +303,33 @@ static dc_status_t qt_serial_write(void *io, const void* data, size_t size, size
 	return DC_STATUS_SUCCESS;
 }
 
+static dc_status_t qt_serial_poll(void *io, int timeout)
+{
+	qt_serial_t *device = (qt_serial_t*) io;
+
+	if (!device)
+		return DC_STATUS_INVALIDARGS;
+
+#if defined(Q_OS_WIN)
+	// FIXME FIXME FIXME!! But how ?
+	// I have no idea about windows socket programming - Linus
+	// We'll just pretend it's always readable, and hope for the best
+	// Why is the windows side not using QBluetoothSocket?
+	return DC_STATUS_SUCCESS;
+#else
+	if (!device->socket)
+		return DC_STATUS_INVALIDARGS;
+	if (device->socket->waitForReadyRead(timeout))
+		return DC_STATUS_SUCCESS;
+	return DC_STATUS_TIMEOUT;
+#endif
+}
+
+static dc_status_t qt_serial_ioctl(void *io, unsigned int request, void *data, size_t size)
+{
+	return DC_STATUS_UNSUPPORTED;
+}
+
 static dc_status_t qt_serial_purge(void *io, dc_direction_t)
 {
 	qt_serial_t *device = (qt_serial_t*) io;
@@ -385,20 +413,20 @@ ble_packet_open(dc_iostream_t **iostream, dc_context_t *context, const char* dev
 
 	static const dc_custom_cbs_t callbacks = {
 		qt_ble_set_timeout, /* set_timeout */
-		NULL, /* set_latency */
 		NULL, /* set_break */
 		NULL, /* set_dtr */
 		NULL, /* set_rts */
 		NULL, /* get_lines */
 		NULL, /* get_received */
 		NULL, /* configure */
+		qt_ble_poll, /* poll */
 		qt_ble_read, /* read */
 		qt_ble_write, /* write */
+		qt_ble_ioctl, /* ioctl */
 		NULL, /* flush */
 		NULL, /* purge */
 		qt_custom_sleep, /* sleep */
 		qt_ble_close, /* close */
-		qt_ble_get_name, /* get_name */
 	};
 
 	rc = qt_ble_open(&io, context, devaddr, (dc_user_device_t *) userdata);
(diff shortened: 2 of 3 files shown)


dirkhh:
Frankly, I have no idea why the AppImage build fails. This is very odd.
But more importantly the Windows build succeeded, so now people should be able to test the artifact...

core/qtserialbluetooth.cpp
```diff
@@ -295,45 +196,51 @@ static dc_status_t qt_serial_write(void *io, const void* data, size_t size, size
...
+	if (!device)
 		return DC_STATUS_INVALIDARGS;
-#if !defined(Q_OS_WIN)
-	if (device->socket == NULL)
+
+	if (!device->socket)
 		return DC_STATUS_INVALIDARGS;
-#endif
-	// TODO: add implementation
+	if (device->socket->waitForReadyRead(timeout))
```
neolit123: i was checking if Qt for Windows has implementation for this and found:

https://doc.qt.io/qt-5/qbluetoothsocket.html

> Note: QBluetoothSocket does not support synchronous read and write operations. Functions such as waitForReadyRead() and waitForBytesWritten() are not implemented. I/O operations should be performed using readyRead(), read() and write().

there is also this comment in the underlying (abstract) implementation for Windows:

https://doc.qt.io/qt-5/qabstractsocket.html#waitForReadyRead

> Note: This function may fail randomly on Windows. Consider using the event loop and the readyRead() signal if your software will run on Windows.

so i think we should be doing something similar to what we have already here:
https://github.com/Subsurface-divelog/subsurface/blob/e28be280cd73fce5f41215ccb476ce97bc4073a7/core/qtserialbluetooth.cpp#L152

```
static dc_status_t qt_serial_poll(void *io, int timeout)
{
	qt_serial_t *device = (qt_serial_t*) io;

	if (!device)
		return DC_STATUS_INVALIDARGS;

	if (!device->socket)
		return DC_STATUS_INVALIDARGS;

	QEventLoop loop;
	QTimer timer;
	timer.setSingleShot(true);
	loop.connect(&timer, SIGNAL(timeout()), SLOT(quit()));
	loop.connect(device->socket, SIGNAL(readyRead()), SLOT(quit()));
	timer.start(device->timeout);
	loop.exec();
	
	if (!timer.isActive())
		return DC_STATUS_SUCCESS;
	return DC_STATUS_TIMEOUT;
}
```

neolit123:
@dirkhh

> Frankly, I have no idea why the AppImage build fails. This is very odd.

it seems as if `dc_custom_cbs_t` is not accepting the right function pointers.
could this build be using an out of date libdc?

<details>

```
/__w/***/***/core/qtserialbluetooth.cpp: In function 'dc_status_t ble_packet_open(dc_iostream_t**, dc_context_t*, const char*, void*)':
/__w/***/***/core/qtserialbluetooth.cpp:300:2: error: invalid conversion from 'dc_status_t (*)(void*, int)' to 'dc_status_t (*)(void*, unsigned int, unsigned int, dc_parity_t, dc_stopbits_t, dc_flowcontrol_t)' [-fpermissive]
  };
  ^
/__w/***/***/core/qtserialbluetooth.cpp:300:2: error: invalid conversion from 'dc_status_t (*)(void*, unsigned int, void*, size_t) {aka dc_status_t (*)(void*, unsigned int, void*, long unsigned int)}' to 'dc_status_t (*)(void*)' [-fpermissive]
/__w/***/***/core/qtserialbluetooth.cpp:300:2: error: invalid conversion from 'dc_status_t (*)(void*, unsigned int)' to 'dc_status_t (*)(void*)' [-fpermissive]
/__w/***/***/core/qtserialbluetooth.cpp:300:2: error: invalid conversion from 'dc_status_t (*)(void*)' to 'const char* (*)(void*)' [-fpermissive]
/__w/***/***/core/qtserialbluetooth.cpp: In function 'dc_status_t rfcomm_stream_open(dc_iostream_t**, dc_context_t*, const char*)':
/__w/***/***/core/qtserialbluetooth.cpp:334:2: error: invalid conversion from 'dc_status_t (*)(void*, size_t*) {aka dc_status_t (*)(void*, long unsigned int*)}' to 'dc_status_t (*)(void*, unsigned int*)' [-fpermissive]
  };
  ^
/__w/***/***/core/qtserialbluetooth.cpp:334:2: error: invalid conversion from 'dc_status_t (*)(void*, int)' to 'dc_status_t (*)(void*, unsigned int, unsigned int, dc_parity_t, dc_stopbits_t, dc_flowcontrol_t)' [-fpermissive]
/__w/***/***/core/qtserialbluetooth.cpp:334:2: error: invalid conversion from 'dc_status_t (*)(void*, unsigned int, void*, size_t) {aka dc_status_t (*)(void*, unsigned int, void*, long unsigned int)}' to 'dc_status_t (*)(void*)' [-fpermissive]
/__w/***/***/core/qtserialbluetooth.cpp:334:2: error: invalid conversion from 'dc_status_t (*)(void*, dc_direction_t)' to 'dc_status_t (*)(void*, unsigned int)' [-fpermissive]
/__w/***/***/core/qtserialbluetooth.cpp:334:2: error: invalid conversion from 'dc_status_t (*)(void*, unsigned int)' to 'dc_status_t (*)(void*)' [-fpermissive]
/__w/***/***/core/qtserialbluetooth.cpp:334:2: error: invalid conversion from 'dc_status_t (*)(void*)' to 'const char* (*)(void*)' [-fpermissive]
make[2]: *** [core/CMakeFiles/***_corelib.dir/qtserialbluetooth.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [core/CMakeFiles/***_corelib.dir/all] Error 2
make: *** [all] Error 2
```

</details>

dirkhh:
> @dirkhh
> 
> > Frankly, I have no idea why the AppImage build fails. This is very odd.
> 
> it seems as if `dc_custom_cbs_t` is not accepting the right function pointers.
> could this build be using an out of date libdc?

Yes, that's exactly what it looks like, yet at the top of the log you see it updating the submodule and building that (of course I don't recall seeing the actual sha, so maybe that's where something is hosed)...

neolit123:
> of course I don't recall seeing the actual sha, so maybe that's where something is hosed

the SHA matches:

> Submodule path 'libdivecomputer': checked out 'f2f775b9aaec76abf9a534b9385dfd2d5fe5058d'

https://github.com/Subsurface-divelog/libdc/commit/f2f775b9aaec76abf9a534b9385dfd2d5fe5058d

odd.

core/qtserialbluetooth.cpp
neolit123: i can send a follow up for this once we resolve the build problem for AppImage.

dirkhh: Let's ignore this for the moment and assume that it will work once merged...
You should be able to push additional commits into the other PR

dirkhh: sorry, got confused. THIS PR

neolit123: i pushed a commit.

dirkhh:
So I finally got my Windows PC to work. Tried this prior to Lubomir's commit with the binary based on e28be28 - the only BT dive computer I have is a Shearwater Petrel which is known to be a PITA. This one is the BT only version. I can pair it on Windows. But a download attempt immediately fails.
Somehow the logfiles in AppData\Roaming\Subsurface are both empty (which seems wrong)

dirkhh:
Take it back. I forgot the most important step whenever you deal with a Windows issue.

REBOOT

After a reboot, I was able to download from my Petrel. Very clearly a successful BT (not BLE - two orders of magnitude faster than BLE, and this Petrel doesn't support BLE, anyway) connection.

YAY!!!

What additional "thing" does your commit add, Lubomir? Should I have expected my download to fail?

neolit123:
> What additional "thing" does your commit add, Lubomir? Should I have expected my download to fail?

it really depends what libdc wants to do with it's poll callback. maybe it's something for some DCs but not for others. i do not know the underlying details.

but reading the Qt docs they mention that we should not be using waitForReadyRead(), thus the change i've added.

dirkhh:
Are you still up at 6 or already up at 6?

I guess I'll merge the PR...

On January 26, 2020 8:05:49 PM PST, "Lubomir I. Ivanov" <notifications@github.com> wrote:
>> What additional "thing" does your commit add, Lubomir? Should I have
>expected my download to fail?
>
>it really depends what libdc wants to do with it's poll callback. maybe
>it's something for some DCs but not for others. i do not know the
>underlying details.
>
>but reading the Qt docs they mention that we should not be using
>waitForReadyRead(), thus the change i've added.
>
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/Subsurface-divelog/subsurface/pull/2559#issuecomment-578584463

-- 
From my phone

neolit123:
> Are you still up at 6 or already up at 6?

going to bed soon.

core/qt-ble.h
```diff
@@ -23,7 +23,13 @@ class BLEObject : public QObject
 	inline void set_timeout(int value) { timeout = value; }
 	dc_status_t write(const void* data, size_t size, size_t *actual);
 	dc_status_t read(void* data, size_t size, size_t *actual);
-	inline const char *get_name() { return device->btname; }
+	inline dc_status_t get_name(char *res, size_t size)
```
bstoeger, line 26: From what I can tell, this all looks good. However, note that this is way out of my comfort zone.

My only stylistic comment would be to un-inline this function. I'm not a fan of mixing function declaration and function definition. Moreover, the `inline` keyword is redundant here. These stylistic issues can be addressed in a later commit, though.

bstoeger (approved):

core/qt-ble.h
dirkhh: yeah, let's deal with that later and merge now... mainly to see if that deals with the weird build issue that I still don't understand. I'm actually worried that there is some older include file that is part of the build container and that somehow that gets picked over the one that is built during the GitHub Action run. If that's the case, then I need to fix / update the build container in order to address that :-/

dirkhh (approved):
let's merge

core/qt-ble.h
torvalds: This pr is closed, but just commenting on this part anyway for posterity: I agree, but it would probably be best to just make the whole "ioctl()" function be the thing we uninline and make a public function.

But yeah, for this round I just did the minimal changes for just the interface change.

PR #2560, opened 2020-01-27
build container workaround: outdated header files

dirkhh:
try to track down the build error


dirkhh (commented):
just updated the commit messages.
This fixes half of the current build problems on GitHub Actions.

PR #2565, opened 2020-01-29
(1b) Fix Kirigami error.

janiversen:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

For a short while this error have been reported:
INFO: qrc:/org/kde/kirigami/ScrollablePage.qml:187: TypeError: Cannot assign to read-only property "parent".

It boiled down to a ListModel that was defined directly at page level, something which is very normal and allowed in QtQuick 2. Kirigami.scrollPage does not allow non graphical elements to be defined directly.

Kirigami did an excellent job, with QtQuick 1, but with QtQuick 2 in reality offers very little and poses limitations compared to using QtQuick 2 directly. I believe strongly it is a good idea to isolate Kirigami in Template*, but not in a "Big Bang" commit, but as we slowly update the pages.

### DIFF
--- a/mobile-widgets/qml/DiveSummary.qml
+++ b/mobile-widgets/qml/DiveSummary.qml
@@ -11,23 +11,6 @@ Kirigami.ScrollablePage {
 	background: Rectangle { color: subsurfaceTheme.backgroundColor }
 	title: qsTr("Dive summary")
 
-	ListModel {
-		id: monthModel
-		ListElement {text: qsTr("Total")}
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
I am open to completely removing Kirigami if that is a reasonable direction. For us, when we got started, it provided something that seemed more useful than "just" QML. And we obviously use quite a few of its features today. That doesn't mean that I am determined to keep it

PR #2566, opened 2020-01-29
mobile/dive summary: rewrite statistics code

dirkhh:
There were quite a few issues with the code - clearly a complete failure of
code review.

- all values were '??' if a period contained no dives
- imperial units were not calculated at all
- significant truncation and data loss in the way totals were added as meters
  and minutes instead of the higher precision data that is available
- several issues in striing conversion methodology, e.g. missing zero padding
  for minutes
- missing maxSac
- incorrectly calculated avgSac
- incorrectly claculated number of EANx dives
- hard to read code with most variables named 'temp'

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I clearly completely failed in my code review. I have no excuse.

### Changes made:

too many to list. Basically a rewrite of the statistics code

### Release note:

not necessary as this doesn't add anything meaningfully new

### Mentions:

@janiversen

### DIFF
--- a/backend-shared/divesummary.cpp
+++ b/backend-shared/divesummary.cpp
@@ -9,9 +9,9 @@ QStringList diveSummary::diveSummaryText;
 
 timestamp_t diveSummary::firstDive, diveSummary::lastDive;
 int diveSummary::dives[2], diveSummary::divesEAN[2], diveSummary::divesDeep[2], diveSummary::diveplans[2];
-long diveSummary::divetime[2], diveSummary::depth[2], diveSummary::sac[2];
-long diveSummary::divetimeMax[2], diveSummary::depthMax[2], diveSummary::sacMin[2];
-int diveSummary::divesSAC[2];
+long diveSummary::divetime[2], diveSummary::depth[2];
+long diveSummary::divetimeMax[2], diveSummary::depthMax[2], diveSummary::sacMin[2], diveSummary::sacMax[2];
+long diveSummary::totalSACTime[2], diveSummary::totalSacVolume[2];
 
 void diveSummary::summaryCalculation(int primaryPeriod, int secondaryPeriod)
 {
@@ -28,12 +28,12 @@ void diveSummary::summaryCalculation(int primaryPeriod, int secondaryPeriod)
 
 	// prepare stringlist
 	diveSummaryText.clear();
-	diveSummaryText << "??" << "??" << "??" << "??" << "??" <<
+	diveSummaryText << "??" << "??" << QObject::tr("no dives in period") << QObject::tr("no divies in period")  << "??" <<
 						 "??" << "??" << "??" <<
 						 "?:??" << "?:??" << "?:??" <<
 						 "?:??" << "?:??" << "?:??" <<
 						 "??" << "??" << "??" << "??" << "??" <<
-						 "??" << "??" << "??" << "??" << "??";
+						 "??" << "??" << "??" << "??" << "??" << "??" << "??";
 
 	// set oldest/newest date
 	if (firstDive) {
@@ -62,10 +62,11 @@ void diveSummary::loopDives(timestamp_t primaryStart, timestamp_t secondaryStart
 	dives[0] = dives[1] = divesEAN[0] = divesEAN[1] = 0;
 	divesDeep[0] = divesDeep[1] = diveplans[0] = diveplans[1] = 0;
 	divetime[0] = divetime[1] = depth[0] = depth[1] = 0;
-	sac[0] = sac[1] = 0;
 	divetimeMax[0] = divetimeMax[1] = depthMax[0] = depthMax[1] = 0;
 	sacMin[0] = sacMin[1] = 99999;
-	divesSAC[0] = divesSAC[1] = 0;
+	sacMax[0] = sacMax[1] = 0;
+	totalSACTime[0] = totalSACTime[1] = 0;
+	totalSacVolume[0] = totalSacVolume[1] = 0;
 
 	for_each_dive (i, dive) {
 		// remember time of oldest and newest dive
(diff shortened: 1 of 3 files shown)


dirkhh:
oh, found one more issue.
switching units in the settings doesn't update the dive summary.
once you change one of the time periods it gets recalculated and the correct units are used.

janiversen:
Just a few comments:

- all values were '??' if a period contained no dives
That was a design decision, to indicate clearly that there are not data.

- imperial units were not calculated at all
hmmm, that can only be a presentation issue, because we do not store imperial units extra.

- significant truncation and data loss in the way totals were added as meters
That was a design decision, normally divers do not keep mm when keeping a manual log.

- and minutes instead of the higher precision data that is available
Also a design decision, the desktop version actually also use meters in several cases.

- several issues in striing conversion methodology, e.g. missing zero padding
- for minutes
correct.

- missing maxSac
Who is interested in maxSac, that makes no sense. If you like me have a transmitter maxSac is somewhere around 40 l/min, but just because it starts calculating from the first second, and thus catches e.g. the full inflation of my BC.

- incorrectly calculated avgSac
I will have a look at your code, I made an extra variable to count it correctly.

- incorrectly claculated number of EANx dives
I will have a look at your code

- hard to read code with most variables named 'temp'
that is because they are temp ! just instead of repeating the calculations.

Let me walk through the code.

backend-shared/divesummary.cpp
```diff
@@ -9,9 +9,9 @@ QStringList diveSummary::diveSummaryText;
 
 timestamp_t diveSummary::firstDive, diveSummary::lastDive;
 int diveSummary::dives[2], diveSummary::divesEAN[2], diveSummary::divesDeep[2], diveSummary::diveplans[2];
-long diveSummary::divetime[2], diveSummary::depth[2], diveSummary::sac[2];
-long diveSummary::divetimeMax[2], diveSummary::depthMax[2], diveSummary::sacMin[2];
-int diveSummary::divesSAC[2];
+long diveSummary::divetime[2], diveSummary::depth[2];
+long diveSummary::divetimeMax[2], diveSummary::depthMax[2], diveSummary::sacMin[2], diveSummary::sacMax[2];
```
janiversen, line 13: sacMax is really not needed. and if needed how come minDepth is not needed ?

backend-shared/divesummary.cpp
```diff
@@ -28,12 +28,12 @@ void diveSummary::summaryCalculation(int primaryPeriod, int secondaryPeriod)
 
 	// prepare stringlist
 	diveSummaryText.clear();
-	diveSummaryText << "??" << "??" << "??" << "??" << "??" <<
+	diveSummaryText << "??" << "??" << QObject::tr("no dives in period") << QObject::tr("no divies in period")  << "??" <<
```
janiversen, line 31: That is a nice change, I should have thought of that.

backend-shared/divesummary.cpp
```diff
@@ -94,83 +95,90 @@ void diveSummary::loopDives(timestamp_t primaryStart, timestamp_t secondaryStart
...
 	dives[inx]++;
 
 	// sum dive in minutes and check for new max.
-	temp = dive->duration.seconds / 60;
-	divetime[inx] += temp;
-	if (temp > divetimeMax[inx])
-		divetimeMax[inx] = temp;
+	divetime[inx] += dive->duration.seconds;
+	if (dive->duration.seconds > divetimeMax[inx])
+		divetimeMax[inx] = dive->duration.seconds;
```
janiversen: Remark if you compare the value to a manual log you will see differences.

backend-shared/divesummary.cpp
```diff
@@ -94,83 +95,90 @@ void diveSummary::loopDives(timestamp_t primaryStart, timestamp_t secondaryStart
...
+		if (dive->sac > sacMax[inx])
+			sacMax[inx] = dive->sac;
 	}
 
 	// EAN dive ?
 	for (int j = 0; j < dive->cylinders.nr; ++j) {
-		if (dive->cylinders.cylinders[j].gasmix.o2.permille > 210)
+		if (dive->cylinders.cylinders[j].gasmix.o2.permille > 210) {
 			divesEAN[inx]++;
+			break;
```
janiversen, line 127: ups very good catch.

dirkhh:
The desktop does all calculations at full precision and then rounds for the presentation layer.
This patch creates the same output as the desktop does.
We have divers argue with us about whether salt water has a specific weight of 1025g/l or 1030g/l and they want the ability to adjust that to get "more accurate" depth readings. So no, truncating dives down the the next full meter and calculating dive durations based on minutes will not work.

the highest SAC for a dive of course makes sense when trying to get a range of typical air consumption. Just like minimum SAC isn't the lowest SAC during a dive but the lowest dive SAC over all (selected) dives, the same logic is used for maximum SAC.

Calculating avg SAC as sum(sac(dive))/#dives is matematically not useful. The code in the commit uses the same logic as the desktop app (treat all selected dives as one long "dive" and correctly calculate the SAC for that. I.e., SAC is weighted by dive duration.

As a tech diver with multiple EANx cylinders your code gave me more EANx dives than total dives.

mobile-widgets/qml/DiveSummary.qml
```diff
@@ -181,7 +181,7 @@ Kirigami.ScrollablePage {
 			text: Backend.diveSummaryText[19]
 		}
 		TemplateLabel {
-			text: qsTr("avg. SAC")
+			text: qsTr("max. SAC")
```
janiversen, line 184: Does that mean we no longer present average sac ? That is bad, the average sac is used to see the a diver becomes better over timer (== consumes less air).

janiversen (commented):
I hope by having selected "comment" not to have started a review. You have made quite a number of good changes to the code, but e.g. adding maxSac (and not consequently add minDepth) and (I did not test compile) remove average sac, makes it a lot less useful.

janiversen:
"oh, found one more issue.
switching units in the settings doesn't update the dive summary.
once you change one of the time periods it gets recalculated and the correct units are used."

That is a mistake, in order to catch that different unit signals need to be caught (a connect to a slot method).

backend-shared/divesummary.cpp
dirkhh: I guess you and I just differ in our opinion on what is useful information.
The shallowest dive I have done? I wouldn't know who cares about this.
The highest SAC rate I had on a dive? Yeah, when planning dives it is extremely useful to have an upper boundary for the SAC rate to make sure you don't run out of gas. I think that value is much more useful in real life than the minimum SAC rate (which is basically a vanity number - yay, my min SAC is 7l/min. Great. So what?)

mobile-widgets/qml/DiveSummary.qml
dirkhh: please read the full patch :-(

janiversen:
Remark calculating the average sac as if it is one long dive, does not really tell if a diver gets betters. We use the average sac for each to dive, to secure a diver over time gets a lower consumption, which seems to be different from you what think of. This is especially true for relative new recreational divers, and for technical divers when switching equipment (e.g. I switches to dive with side mount, and it took me a while to get my consumption down).

Yes it seems we have different opinion on which values are needed, for specialists we should have a real statistical page. My intent was to help relative new divers when they come to a dive center and to show them how they get better.

But no problem, if you prefer other content.

dirkhh:
Jan.
I have one hundred 30 second dives with a SAC rate of 1l/min and ten 1h dive with a SAC rate of 10l/min

Would you say my average SAC rate is closer to 1l/min or closer to 10l/min?

Math matters. The calculation in my code calculates the average SAC rate across the selected time.
Maybe you want the value that you are calculated in your code. Maybe you could call it the average of the per dive SAC rate, or something like that. But it is definitely not the average SAC rate.

janiversen:
Now I remember why the original page removes all dives less than 5 minutes and less than 4 meters (swimming pool) from being summed/calculated. 

You could consider doing something similar.

dirkhh:
Given the underlying goal to have the statistics on desktop and mobile provide consistent results, I'll merge this as is.

PR #2568, opened 2020-01-30
small adjustments to Divesummary

janiversen:

- [ ] Bug fix
- [x ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Commit 1:
Secures page is updated when user changes units (length or volume)

Commit 2:
introduces limit to what is calculated.

1) Add a note telling how the calculation is done
2) Exclude dives shorter than 5 minutes and shallower than 4 meters (swimming pool)
Remark: the page is intended to be shown in dive centres, and a dive center will not consider
these dives as real dives
3) Exclude SAC < 4 l/min from SAC calculation
thinking of your 30 1 l/min dives
4) Exclude SAC > 40 l/min from SAC calculation
Any diver breathing more than 40 l/min have a serious problem :-)
And this excludes my real life experience with 2 regulators stuck in free flow at 30m when returning on a deco dive (luckily my O2 tank worked at 6m)

Having digested your changes of the calculation, I agree they make a lot sense, including the SAC average.

### DIFF
--- a/backend-shared/divesummary.cpp
+++ b/backend-shared/divesummary.cpp
@@ -95,6 +95,13 @@ void diveSummary::loopDives(timestamp_t primaryStart, timestamp_t secondaryStart
 
 void diveSummary::calculateDive(int inx, struct dive *dive)
 {
+	// Check for dive validity
+	// Swimmingpool dives (4 meters) are not counted, nor
+	// are very short dives (5 minutes)
+	if (dive->maxdepth.mm < 4000 ||
+		dive->duration.seconds < 5 * 60)
+		return;
+
 	// one more real dive
 	dives[inx]++;
 
@@ -111,7 +118,9 @@ void diveSummary::calculateDive(int inx, struct dive *dive)
 		divesDeep[inx]++;
 
 	// sum SAC, check for new min/max.
-	if (dive->sac) {
+	// Do not include dives with SAC > 40 l/min (probably a free flow regulator)
+	// and SAC < 4 l/min (probably a faulty transmittor)
+	if (dive->sac <= 40000 && dive->sac >= 4000) {
 		totalSACTime[inx] += dive->duration.seconds;
 		totalSacVolume[inx] += dive->sac * dive->duration.seconds;
 		if (dive->sac < sacMin[inx])
@@ -175,9 +184,14 @@ void diveSummary::buildStringList(int inx)
 	unitText = (qPrefUnits::volume() == units::LITER) ? " l/min" : " cuft/min";
 	diveSummaryText[18+inx] = volumeString(sacMin[inx]) + unitText;
 	diveSummaryText[20+inx] = volumeString(sacMax[inx]) + unitText;
+
 	// finally the weighted average
-	long avgSac = totalSacVolume[inx] / totalSACTime[inx];
-	diveSummaryText[22+inx] = volumeString(avgSac) + unitText;
+	if (totalSACTime[inx]) {
+		long avgSac = totalSacVolume[inx] / totalSACTime[inx];
+		diveSummaryText[22+inx] = volumeString(avgSac) + unitText;
+	} else {
+		diveSummaryText[22+inx] = QObject::tr("no dives");
+	}
 
 	// Diveplan(s)
 	diveSummaryText[24+inx] = QStringLiteral("%1").arg(diveplans[inx]);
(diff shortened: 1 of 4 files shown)


PeterZaal:
Regarding 2: So if I dive in a shallow cave for almost 2 hours and it's only 3 meters deep, that is not a real dive?
Regarding 3: So all my rebreather dives are not taken into account? They are for me typically around 0,8 l/min.

janiversen:
That is a good point, but then again this is not a general statistic module (something that are in the works), and I assume being a cave diver, you do not need to convince a dive center that you qualify for a guided "normal" dive.

The idea is not to present the perfect statistic, but to be a page which the dive center can look at when especially divers with less experience wants to participate in a for them more challenging dive.

That all being said, I will add an option to bypass the limitation.

janiversen:
Checkbox added, so now your dives count as real dives :-)

But please remember the page is oriented at relative new divers. #2523 could be more useful for you.

backend-shared/divesummary.cpp
```diff
@@ -95,6 +95,13 @@ void diveSummary::loopDives(timestamp_t primaryStart, timestamp_t secondaryStart
 
 void diveSummary::calculateDive(int inx, struct dive *dive)
 {
+	// Check for dive validity
+	// Swimmingpool dives (4 meters) are not counted, nor
+	// are very short dives (5 minutes)
+	if (dive->maxdepth.mm < 4000 ||
+		dive->duration.seconds < 5 * 60)
+		return;
```
dirkhh, line 104: I don't think arbitrary thresholds like this are a good idea.
THIS is the real reason to have an invalid dive flag.
If the user wants to exclude certain dives, we should give them the option to do so.
But we should absolutely not create a "summary" that arbitrary excludes dives from the summary.
So NAK on this change.

backend-shared/divesummary.cpp
```diff
@@ -111,7 +118,9 @@ void diveSummary::calculateDive(int inx, struct dive *dive)
 		divesDeep[inx]++;
 
 	// sum SAC, check for new min/max.
-	if (dive->sac) {
+	// Do not include dives with SAC > 40 l/min (probably a free flow regulator)
+	// and SAC < 4 l/min (probably a faulty transmittor)
+	if (dive->sac <= 40000 && dive->sac >= 4000) {
```
dirkhh: I have several dives with a SAC below 7l, a very good friend of mine has many dives around 5l (she is not a big person). And as a tech dive master, teaching advanced classes in ~8C water, I have seen many students with SAC rates north of 35l
So again. No. We should not have such arbitrary cut-offs. If a user wants to exclude a dive from the summary, give them a way to do so. Do not arbitrary exclude them.
NAK

backend-shared/divesummary.cpp
```diff
@@ -175,9 +184,14 @@ void diveSummary::buildStringList(int inx)
...
+
 	// finally the weighted average
-	long avgSac = totalSacVolume[inx] / totalSACTime[inx];
-	diveSummaryText[22+inx] = volumeString(avgSac) + unitText;
+	if (totalSACTime[inx]) {
+		long avgSac = totalSacVolume[inx] / totalSACTime[inx];
+		diveSummaryText[22+inx] = volumeString(avgSac) + unitText;
+	} else {
+		diveSummaryText[22+inx] = QObject::tr("no dives");
+	}
```
dirkhh, line 195: great catch

dirkhh (changes requested):
NAK as submitted
While personally I wouldn't use it, I might consider an inverted approach.
Have correct calculation by default and offer people who would like to algorithmically exclude some of the dives the ability to set those thresholds and enable them.

dirkhh:
Since I expect that this may take a day or two to sort out, I'll manually extract the two significant fixes from this PR.

janiversen:
"NAK as submitted",

No problem, this is not an important change to me. Once we can tough the models again, we have some ideas on how to make a more general statistic model (replacing the yearly statistic model) and with flexibility to sort/group on columns.

dirkhh:
As indicated, I grabbed the indisputably strong changes in the PR and pushed them to master

janiversen:
oooh forgot you wrote that, sorry if my close caused you trouble. Adding that commit to master is of course ok.

PR #2570, opened 2020-01-30
(1) Theme: change color layout.

janiversen:

- [ ] Bug fix
- [ ] Functional change
- [ x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

Show the user all colours:

Captura de pantalla 2020-01-30 a las 18 09 06


mobile-widgets/qml/TemplateButton.qml
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: GPL-2.0
```
dirkhh: Can we do a single commit that does indeed do the git move and replaces all the uses?
I think that would be preferrable.

mobile-widgets/themeinterface.cpp
```diff
@@ -26,6 +26,7 @@ void themeInterface::setup(QQmlContext *ct)
 	instance()->set_currentScale(qPrefDisplay::mobile_scale());
 }
 
+
```
dirkhh: that's an interesting commit. I think my small cleanup caused your tooling to add this...
obviously this commit should be dropped

mobile-widgets/themeinterface.cpp
```diff
@@ -26,7 +26,6 @@ void themeInterface::setup(QQmlContext *ct)
 	instance()->set_currentScale(qPrefDisplay::mobile_scale());
 }
 
-
```
dirkhh: and that's the other half of it. Please drop as well.

mobile-widgets/qml/Settings.qml
```diff
@@ -139,201 +139,221 @@ TemplatePage {
...
+					currentIndex: (ThemeNew.currentTheme === "Blue") ? 0 :
+								  (ThemeNew.currentTheme === "Pink") ? 1 : 2
+					model: ListModel {
+						   ListElement {text: qsTr("Blue")}
+						   ListElement {text: qsTr("Pink")}
+						   ListElement {text: qsTr("Dark")}
+					}
+					onActivated:  {
+						ThemeNew.currentTheme = currentIndex === 0 ? "Blue" :
+												currentIndex === 1 ? "Pink" : "Dark"
```
dirkhh, line 157: odd whitespace

dirkhh (commented):
a couple of small suggestions for cleanup. but overall good

mobile-widgets/qml/Settings.qml
janiversen: line 149 is alignment (at least in Qt Creator)
line 151-153 is also alignment

Which I thought to be ok (at least you have corrected me, for changing that earlier).

mobile-widgets/qml/TemplateButton.qml
janiversen: Done.

janiversen:
Pushed new version, that hopefully contain all the changes you requested.

janiversen:
Seems I managed to break something by removing the 2 commits, will look at it tomorrow

dirkhh:
looks like you still have a dependency on the `SsrfButton.qml` in your `.qrc` file.

janiversen:
Could not sleep (which is bad because I have a 95m dive tomorrow or Saturday depending on the current), kept thinking about this error. Thanks for pointing me in the right direction.

Force pushed an update, lets see if it compiles, and then back to the sack. It going to be an interesting dive  2x 3mix using sidemount, 1x Air transport (up/down) and 1x 85% O2 for deco. Going in a group of 5.

janiversen:
Looks good, at least no general errors. Going zzz offline.

janiversen:
I assume this PR does not really make sense anymore. I will let the branch stay for a couple of days should I be wrong in my assumption.

PR #2572, opened 2020-01-31
[RFC] Remove plannerShared::instance()

bstoeger:
This class contains only static functions (i.e. it does not contain
any state). There does not seem to be a reason to have an instance
of that class. Therefore, remove the instance() function.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The `plannerShared`[*] class has no state, but rather is a collection of static functions. Currently, there appears to be no reason to keep a global instance of this class. If QML needs it, it can generate its own instance.

Should the plan be to add state to this class, this needs some discussion. The planner state is already splattered over a number of classes, I'm not convinced that splitting this up more is the way to go.

[*] virtually all of our class-names are PascalCase, I see no reason to break this rule.
### Changes made:

Remove `plannerShared::instance()`

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -6,12 +6,6 @@
 #include "qt-models/diveplannermodel.h"
 #include "qt-models/cylindermodel.h"
 
-plannerShared *plannerShared::instance()
-{
-    static plannerShared *self = new plannerShared;
-    return self;
-}
-
 // Planning values
 deco_mode plannerShared::planner_deco_mode()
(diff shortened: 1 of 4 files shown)


janiversen:
Yes it can be removed, instance() is a leftover from when plannerShared interacted directly with QML. Thanks for catching that.

bstoeger:
Merge since trivial and no objections.

PR #2574, opened 2020-01-31
Cleanup: remove QMLInterface::instance()

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

First of all, I must say that I think the QMLInterface class is quite horrible. Especially these enum things strike me as a remedy that is much worse than the disease.

Let's remove a few of the particularly illogical/inconsistent bits: an instance which is only used once; connecting in a helper function when you could just do in the constructor as we do nearly everywhere else.

This is only compile tested. Given how insane QML is, I wouldn't wonder if it can't handle non-heap allocated objects or the connects() have to be called in a certain way.

Let's see.

### DIFF
--- a/mobile-widgets/qmlinterface.cpp
+++ b/mobile-widgets/qmlinterface.cpp
@@ -81,16 +81,11 @@ QMLInterface::QMLInterface()
 			this, &QMLInterface::display_variationsChanged);
 }
 
-QMLInterface *QMLInterface::instance()
-{
-	static QMLInterface *self = new QMLInterface;
-	return self;
-}
-
 void QMLInterface::setup(QQmlContext *ct)
 {
 	// Register interface class
-	ct->setContextProperty("Backend", QMLInterface::instance());
+	static QMLInterface self;
+	ct->setContextProperty("Backend", &self);
 
 	// Make enums available as types
 	qmlRegisterUncreatableType<QMLInterface>("org.subsurfacedivelog.mobile",1,0,"Enums","Enum is not a type");
@@ -99,7 +94,6 @@ void QMLInterface::setup(QQmlContext *ct)
 	// this is needed in order to load the divesummary page
 	diveSummary::summaryCalculation(0, 3);
 }
-
 
 void QMLInterface::summaryCalculation(int primaryPeriod, int secondaryPeriod)
 {
--- a/mobile-widgets/qmlinterface.cpp
+++ b/mobile-widgets/qmlinterface.cpp
@@ -2,6 +2,85 @@
 #include "qmlinterface.h"
 #include <QQmlEngine>
 
+QMLInterface::QMLInterface()
+{
+	// relink signals to QML
+	connect(qPrefCloudStorage::instance(), &qPrefCloudStorage::cloud_verification_statusChanged,
+		[this] (int value) { emit cloud_verification_statusChanged(CLOUD_STATUS(value)); });
+	connect(qPrefUnits::instance(), &qPrefUnits::duration_unitsChanged,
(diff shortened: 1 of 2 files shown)


mobile-widgets/qmlinterface.cpp
```diff
@@ -2,95 +2,98 @@
...
 
-QMLInterface *QMLInterface::instance()
-{
-	static QMLInterface *self = new QMLInterface;
-	return self;
-}
-
-void QMLInterface::setup(QQmlContext *ct)
+QMLInterface::QMLInterface()
 {
```
janiversen: This is a change I am very uncomfortable with. I had severe problems when testing, which was why I moved the connect to setup. What can indeed be discussed is if setup() should be static.

Also I do not see any special savings by doing it in the constructor.

mobile-widgets/qmlinterface.h
```diff
@@ -81,9 +81,7 @@ class QMLInterface : public QObject {
 	Q_PROPERTY(QStringList diveSummaryText READ diveSummaryText NOTIFY diveSummaryTextChanged);
 
 public:
-	static QMLInterface *instance();
-
-	// function to do the needed setup and do connect of signal/signal
+	// function to do the needed setup
```
janiversen: This is not wrong, but merely a matter of taste. I actually use the public instance() quite a lot when I test and prepare new commits.

janiversen (changes requested):
Deleted to avoid more discussion

bstoeger:
> Seriously if it strikes you as being much worse than a disease

This was an old proverb, from which I thought it was obvious that it is not to be taken literally. The disease standing for passing around strings.

mobile-widgets/qmlinterface.cpp
bstoeger, line 6: Show us these problems and then we can either
1) fix the root cause or
2) document why these connects can't be done in the constructor as we do it virtually everywhere else in the code base unless there are some very specific reasons.

A valid reason might for example be that the connects are dynamic and can be set from multiple places.

janiversen: I am sorry, I am not here to play, QMLInterface represent quite an investment in time and trial/error from my side and it works. I politely offered my opinion, based on my experience in code written by me, nothing more.

But anyhow this is way over my head, I am just a contributor, while you are a committer, giving us quite different levels in the project. If you feel this is the right thing to do, then please go ahead and merge it.

bstoeger:
Jeez. I really dislike merging my own PRs because it violates the four eyes principle. However, I dislike the drama even more - so I will simply merge this. If things break for reasons that I don't yet understand, we'll sort it out, one way or the other. Perhaps with a (partial) revert. At least then things are documented.

Just for the record: I also dislike the way enums are treated in QMLInterface and how every signal has to be caught and reposted using the scheme:
```
        connect(qPrefUnits::instance(), &qPrefUnits::temperatureChanged,
                        [] (int value) { emit instance()->temperatureChanged(TEMPERATURE(value)); });
```
I simply don't understand what problem this is solving compared to simply passing ints or strings. QML is not strongly typed no matter what. With this new ultra-complicated scheme - if we hit a typo - we get an additional message in the log. I'm not seeing a convincing scenario where this would be of crucial importance. If an if-statement takes the wrong turn it should be rather easy to figure out where that happened, isn't it?

If you want strong typing, don't use QML, it's that simple. Use QML as display layer that displays stuff and feed it from C++ where you theoretically have strong typing. I say theoretically, because Qt model/view system is very much based on QVariant. :(

PR #2576, opened 2020-02-01
Help needed: Add CylindersModelFiltered

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Currently, the handling of unused cylinders is completely broken: In `CylindersModel` the right number of cylinders are shown, but the code doesn't account for the fact that the unused cylinders might not be the last cylinders.

As a comment suggests, this should be solved by placing a proxy-model on top of the CylindersModel. This PR does just that. It is a bit inconsistent in how some calls are routed through (those taking indexes as parameter must be of course - the index has to be mapped). For other functions the caller has to access the base model. This is up for debate.

I suspect that this completely breaks the planner, as the planner probably supposes that the indexes are with respect to the list in the core...? Therefore, I need help with testing this.

Moreover, we should discuss if this feature is even worth the complexities. Perhaps we should add a "delete unused cylinders" button (with undo functionality, so no risk) and an "import unused cylinders" flag on import instead?

One advantage of the proxy model as realized here is that we can easily implement sorting of the cylinder list if something like that is even useful?

### Changes made:

1) Implement CylindersModelFiltered
2) Replace CylindersModel by CylindersModelFiltered

### Release note:

Perhaps?

### Mentions:

@atdotde @dirkhh

### DIFF
--- a/backend-shared/plannershared.cpp
+++ b/backend-shared/plannershared.cpp
@@ -126,7 +126,7 @@ void PlannerShared::set_o2narcotic(bool value)
 {
 	qPrefDivePlanner::set_o2narcotic(value);
 	DivePlannerPointsModel::instance()->emitDataChanged();
-	CylindersModelFiltered::instance()->model()->updateBestMixes();
+	DivePlannerPointsModel::instance()->cylindersModel()->model()->updateBestMixes();
 }
 
 double PlannerShared::bottompo2()
@@ -136,7 +136,7 @@ double PlannerShared::bottompo2()
 void PlannerShared::set_bottompo2(double value)
 {
 	qPrefDivePlanner::set_bottompo2((int) (value * 1000.0));
-	CylindersModelFiltered::instance()->model()->updateBestMixes();
+	DivePlannerPointsModel::instance()->cylindersModel()->model()->updateBestMixes();
 }
 
 double PlannerShared::decopo2()
@@ -148,8 +148,8 @@ void PlannerShared::set_decopo2(double value)
 	pressure_t olddecopo2;
 	olddecopo2.mbar = prefs.decopo2;
 	qPrefDivePlanner::instance()->set_decopo2((int) (value * 1000.0));
-	CylindersModelFiltered::instance()->model()->updateDecoDepths(olddecopo2);
-	CylindersModelFiltered::instance()->model()->updateBestMixes();
+	DivePlannerPointsModel::instance()->cylindersModel()->model()->updateDecoDepths(olddecopo2);
+	DivePlannerPointsModel::instance()->cylindersModel()->model()->updateBestMixes();
 }
 
 int PlannerShared::bestmixend()
@@ -159,5 +159,5 @@ int PlannerShared::bestmixend()
 void PlannerShared::set_bestmixend(int value)
 {
 	qPrefDivePlanner::set_bestmixend(units_to_depth(value).mm);
-	CylindersModelFiltered::instance()->model()->updateBestMixes();
+	DivePlannerPointsModel::instance()->cylindersModel()->model()->updateBestMixes();
 }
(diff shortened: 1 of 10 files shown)


dirkhh (commented):
I think conceptually this is the right thing to do.
I have NOT tested this, just read through the code (and didn't find anything obviously wrong)

atdotde:
I have to admit, I have not been able to get the consequences of this fully into my head. And probably I don't  understand something very basic. 
I have not been able to create a situation in the context of the planner in which actually a cylinder is filtered out because it is not used. So I am not sure if I am actually testing anything.
But my suspicion is that in the context of the planner, it does not really make sense to filter unused cylinders: In order to create a new cylinder to be used in the planner, you need to click on the plus button in the cylinder list and then quasi by definition it is not used until we either make a manual segment use it or it is picked up for the ascent by the planner. So it should be visible so one can select it, shouldn't it?

I guess, I need some more instructions how to actually test this.

That said, maybe it is helpful to realise the data flow in the planner (I should say originally conceived by @dirkhh): There is struct dive plan (with items diveplannerpoint for the segments which feature an int cylinderid that refers to the cylinder breathed during that segment). This is also the basis for the diveplannerpointsmodel. And there is the struct dive (with gas change events at t=0 to signal available gases) and both get translated back and forth in the process of planning. The code also uses dive.cylinders[dp.cylinderid] so they indices between the divplan and the dive should be in sync.

I don't know if this is helpful. Maybe we should just speak on the phone to clarify these issues.

bstoeger:
Thanks - I pinged you because I likewise don't even know how to systematically test this. It just happened to me that while testing I suddenly got such an unused cylinder and things went weird then.

The way I understand is that when you add cylinders via the UI, they should be marked as "manually_added" and these cylinders are never hidden. Therefore, I reckon that the only way to get unused cylinders is by import from a dive computer that knows about such things...? So probably this is not even relevant for the planner and we should switch back to the unfiltered model?

Perhaps we really should think about removing the feature in its current form - it apparently didn't work properly and nobody ever complained. Instead, allow the user to remove unused tanks directly at import or later...?

atdotde:
Ah, the planner initializes the cylinders by copying them from the current dive (this way you can have standard sets of deco cylinders and don't have to enter them for each dive anew when you plan them. So, we should create a dive in the log with an unused cylinder and start the planner from there to see if we manage to break something.

atdotde:
A simple way out  might be to ignore the preference for hiding unused cylinders whenever the planner is active and this way force the cylindermodel to be identical to the filtered model?

bstoeger:
> A simple way out might be to ignore the preference for hiding unused cylinders whenever the planner is active and this way force the cylindermodel to be identical to the filtered model?

Or simply use the unfiltered model in the planner to start with? I'm not convinced that we should ever hide cylinders in the planner. I will try to change the PR to do just that.

One thing that makes me wonder is that there hasn't been a single complaint about cylinders missing in the planner - so how common is that phenomenon?

bstoeger:
Alright - here's my attempt at using the filtered version only on the equipment tab, but not in the planner.

Admittedly, the history is now very weird.

I'm still unsure if all of this is the right thing to do - but in principle it shouldn't make things worse tp before.

dirkhh:
@atdotde can you play with this? I never use the planner, so I don't think I would even know how to test this.

bstoeger:
It would be nice if we could come to a decision with this PR. Next logical step would be starting with undo of cylinder-editing. Currently, I'm planning to have undo only in effect for the equipment tab, not the planner. Once that works we can still decide if we need undo at the planner level.

So far nobody commented on the idea of removing the "hide unused cylinder" feature altogether. Perhaps bring it up on the mailing list?

dirkhh:
So based on @atdotde 's comments further up I think we should move forward here - give him a moment to yell stop and then just merge it...

atdotde (approved):
No reason to yell. Go ahead!

PR #2577, opened 2020-02-01
[RFC] Move sorting closer to the core

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] Refactoring of core logic

Attempts to generate "flattened" versions of the tree model for mobile are complicated by the fact that the tree model is sorted in ascending order. To avoid these mental gymnastics, it would be nice to have the base model sorted in the "correct" way. However, we can't change the sort-order of the core dive list, because the planner needs to know the nitrogen loading incurred by previous dives and assumes ascending order. Therefore, this PR moves the sorting from down from the filter-model to the tree-model. That is of course pretty hairy so comments / testing are needed. Especially the interaction with the undo-commands that add / remove / modify dives needs testing.

Currently, when changing the sorting, the whole base model is reset. This sounds quite ineffective. However, on my run-of-the-mill laptop a full reload including sorting on buddies of a ~4000 dives test log takes less than 40 microseconds.

### Changes made:

1) Move the sorting of the dive model from filter-proxy model to base model

### Release note:

No.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1370,7 +1370,7 @@ static int comp_dive_to_trip(struct dive *a, struct dive_trip *b)
 	return comp_dives(a, b->dives.dives[0]);
 }
 
-static int comp_dive_or_trip(struct dive_or_trip a, struct dive_or_trip b)
+int comp_dive_or_trip(struct dive_or_trip a, struct dive_or_trip b)
 {
 	/* we should only be called with both a and b having exactly one of
 	 * dive or trip not NULL. But in an abundance of caution, make sure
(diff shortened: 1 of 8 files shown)


dirkhh (commented):
Brain went 'tilt'.
This trip has been absolutely brutal on my sleep patterns - I don't think I'll be able to review this tonight.
I'm heading back home tomorrow (but on a flight with no internet - boo), so it may take a day or two before I get to this
Sorry.

bstoeger:
No worries - I think there are some fundamental flaws with this approach anyway. I'll have to sleep over this.

bstoeger:
After having pondered over the matter, let's not do this for now. While reimplementing QSortFilterProxyModel may be intellectually stimulating, it's currently just not worth it. Especially since for work related reasons my brain's already busted by Markov chains, geometric series and other entry-level maths, which should be trivial unless you haven't done it for a long time... :-P

PR #2580, opened 2020-02-03
Dive site: ignore dive sites without location in proximity search

bstoeger:
When editing a dive site, the user can search for close dive sites
to merge duplicates. Dive sites without location are treated as
being located at 0N0E. This makes no sense, because:

When selecting a dive site without location, we shouldn't list
dive sites close to 0N0E.

Likewise when having a dive site close to 0N0E, we shouldn't list
dive sites that have no location.

Therefore, ignore these cases.

This also means that now dive sites without location are not
considered as close to other dive sites without location. That
might be a debatable point.

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes what IMHO is a bug: dive sites without location are considered as located at 0N0E and treated as such. By ignoring dive sites without location this changes the semantics of the close-dive site search: dive sites without location are now not considered as being at the same place. Not sure about that.
### Changes made:

1) Ignore dive sites without location
### Release note:

Done.
### Documentation change:

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Desktop: ignore dive sites without location in proximity search
 Mobile: add personalized option for units
 Mobile: add Dive Summary page
 Mobile: option to reset default cylinder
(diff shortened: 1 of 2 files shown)


dirkhh:
So it appears that now all of the Ubuntu 18.04 based containers are failing.
And the errors that we see make no sense to me.
Why me?

dirkhh (approved):
I think that's more logical than what we have today.

PR #2581, opened 2020-02-03
Cleanup: Dive sites: select each dive site only once

bstoeger:
After selecting dives, the selected dive sites are collected.
This was done using the selectionModel()->selection().indexes(),
which is wrong, because it gives one index per row *and* column.
Accordingly, every dive site was added numerous times to the
array of dive sites to be selected. Change this to
selectionModel()->selectedRows(), which gives one entry per row.

Moreover, if multiple dives with the same site were selected,
this site was also added to the array multiple times. Therefore,
check the array before adding sites.

Note that all this should not change the user experience in
any way, it is only a code-hygiene thing.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is only a code-hygiene thing. Don't select the same dive site numerous times. Details in the commit description.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -555,10 +555,10 @@ void DiveListView::selectionChangeDone()
 	// by the selected dives.
 	if (!DiveFilter::instance()->diveSiteMode()) {
 		QVector<dive_site *> selectedSites;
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
that looks good

PR #2582, opened 2020-02-03
Cleanup: don't initialize DivePlannerPointsModel at startup

bstoeger:
The whole point of having X::instance() functions is to solve
the infamous "Static Initialization Order Fiasco": When having
global objects in different translation units, their order
of initialization is undefined. Thus, when these objects access
each other one cannot guarantee the correct order of
initialization. The X::instance() functions generate the objects
on first use.

DivePlannerPointsModel has such an instance() function. However,
for convenience(?) in diveplanner.cpp we find the global variable
 static DivePlannerPointsModel* plannerModel =
	DivePlannerPointsModel::instance();
Thus, the DivePlannerPointsModel constructor is run before main(),
negating the whole purpose of the instance() function.

Let's remove this line to avoid hard-to-debug startup issues.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When making the CylindersModel a subobject of the DivePlannerPointsModel, the application crashed on startup. Turns out we initialize DivePlannerPointsModel before main() leading to the infamous "Static Initialization Order Fiasco".

Fix it.

### DIFF
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -24,8 +24,6 @@
 #define MAX_DEPTH M_OR_FT(150, 450)
 #define MIN_DEPTH M_OR_FT(20, 60)
 
-static DivePlannerPointsModel* plannerModel = DivePlannerPointsModel::instance();
-
 DiveHandler::DiveHandler() : QGraphicsEllipseItem()
 {
 	setRect(-5, -5, 10, 10);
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
That seems very reasonable

bstoeger:
Could we commit? I'd like to "unglobalize" the CylindersModel at some point.

atdotde:
This is beyond my payscale. So, yes, go ahead.

atdotde (approved):

PR #2584, opened 2020-02-03
Cleanup: lower-case filenames in core/subsurface-qt/

bstoeger:
We tend to use lower-case filenames. Let's do it for these files
as well. Simple search & replace.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Another idle code-cleanup: turn PascalCase filenames to lower case, since this is what we do most of the time. I compiled it on desktop and it ran through.


dirkhh (commented):
I have mixed feelings about massive search and replace changes like this.
They often create issues for others who are working on longer projects (and make merging open PRs a bit harder).
Yes, consistency is good.
Let me think about this while I try to deal with the GitHub Action breakage and all the other pending PRs...

bstoeger:
Yes, I've shot myself in the foot with this kind of cleanup numerous times. 😛

bstoeger:
That being said - I wouldn't consider a 40 lines diff as being massive and it only concerns includes, so the chance of a rebase-conflict is rather slim. In the case that it happens, it should be rather trivial to fix.

PR #2586, opened 2020-02-03
Dive site: inform map of selection in DiveFilter::setFilterDiveSite

bstoeger:
When starting / changing the dive-site filter, inform the map of
the changed dive site selection by calling
	MapWidget::instance()->selectionChanged();
This fixes a bug, where on clicking dive sites in the dive site
tab the dive sites from the *previous* click were highlighted.

Perhaps the selectionChanged() call should be put into the
setSelected() call. But the data flow between the different
parts of the dive-site and map code are so convoluted that I
don't want to risk anything!

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes an ugly UI bug, where clicking on dive-sites on the dive-site tab would highlight the dive-site from the *previous* click on the map. I'm not sure that the fix is correct - see commit message. But the whole interaction of filter / divesite / map is so complex that I don't dare to touch this for now. And yes, that's at least partially my fault.

### DIFF
--- a/core/divefilter.cpp
+++ b/core/divefilter.cpp
@@ -217,6 +217,7 @@ void DiveFilter::setFilterDiveSite(QVector<dive_site *> ds)
 
 	emit diveListNotifier.filterReset();
 	MapWidget::instance()->setSelected(dive_sites);
+	MapWidget::instance()->selectionChanged();
 	MainWindow::instance()->diveList->expandAll();
 }
 


dirkhh (approved):
good catch

Issue #2037, opened 2019-04-04
Not able to parse CSV in 4.8.5


jofan:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

In the latest version I'm no longer able to import CSV files that worked fine in previous versions.
This is an example of rows in the CSV (#, location, date, max depth, duration, time, notes, buddy):

```
1202	LCU	2005-02-19	14	0:21:00	10:00:00	Fisheye photography	Johanna
1203	Makaha	2005-02-19	10	0:40:00	12:00:00	Intros for Captain Bruce	Johanna
```
I have tried with different date and time formats but not been able to make it work.

### Operating system:

macOS Mojave
English
MacBook Pro 2015

### Subsurface version:

4.8.5

I does not happen to earlier versions.

I'm using the official release

### Steps to reproduce:

1. Copy and paste log rows from Excel into Google Sheets
2. In Google Sheets: select "Download as Tab-separated values"
3. Rename downloaded file from *.tsv to *.csv
4. Select "Import log files" in Subsurface
5. Drag and drop header tags appropriately and import
6. Error message is displayed saying that the CSV could not be parsed.

### Current behavior:

Error message generated that the CSV cannot be parsed

### Expected behavior:

Parsing CSV and importing logs

### Additional information:

Example file attached (rename .txt to .csv)

### Mentions:

[Divelogexport - Sheet22.txt](https://github.com/Subsurface-divelog/subsurface/files/3042302/Divelogexport.-.Sheet22.txt)

jofan:
Having tested some more it seems like it's specific to this file. I'll try to find the row that causes the issue.

atdotde:
Possibly any line with quotation marks or an apostrophe?

jofan:
Thanks @atdotde, it was a & sign that caused the issue.

atdotde:
Ampersand should be a legal character in CSV. This is a bug.

mturkia:
As we are using XSLT to parse CSV, it looks like this cannot be fixed without re-writing the CSV import in C or C++.

dirkhh:
Can this be handled by a pre-parsing step? Where we quote special characters that XSLT can't handle?

PR #2587, opened 2020-02-04
CSV import: fix importing ampersand character

mturkia:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As we do XSLT parsing for the CSV import, ampersand characters need to
be encoded with &amp; for the parsing to succeed.

### Related issues:

#2037 

### Additional information:

Add ampersand character into notes field in CSV file and import it.


dirkhh (approved):
I had to read the encoding / moving things in the buffer around code roughly five times to make sure this looks right... but I think it does

PR #2590, opened 2020-02-04
[RFC] Summary cleanups

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

To be honest, I'm not very happy about how the summary code uses a completely unstructured QStringList to display structured data. I understand that it's annoying but writing this in model-form would seem to be more appropriate and also much more flexible, as you could have an arbitrary number of rows / columns. The way the code looks right now it doesn't seem very appealing to use it on desktop.

Even though this PR doesn't address my fundamental problem with the code - it makes it at least a bit more palatable, because it removes a lot of global state (why are we still adding these?). The only global state left is a cache of the summary in the QML interface. I'm not going to touch that, as DiveSummary.qml gives me vertigo.

### Changes made:

Remove global state; attempt to make the code more logical.

### DIFF
--- a/mobile-widgets/qmlinterface.cpp
+++ b/mobile-widgets/qmlinterface.cpp
@@ -95,8 +95,15 @@ void QMLInterface::setup(QQmlContext *ct)
 	diveSummary::summaryCalculation(0, 3);
 }
 
+static QStringList diveSummaryTextCached;
+
 void QMLInterface::summaryCalculation(int primaryPeriod, int secondaryPeriod)
 {
-	diveSummary::summaryCalculation(primaryPeriod, secondaryPeriod);
-	emit diveSummaryTextChanged(diveSummary::diveSummaryText);
+	diveSummaryTextCached = diveSummary::summaryCalculation(primaryPeriod, secondaryPeriod);
+	emit diveSummaryTextChanged(diveSummaryTextCached);
+}
+
+const QStringList QMLInterface::diveSummaryText()
+{
+	return diveSummaryTextCached;
 }
--- a/mobile-widgets/qmlinterface.cpp
+++ b/mobile-widgets/qmlinterface.cpp
@@ -92,14 +92,14 @@ void QMLInterface::setup(QQmlContext *ct)
 
 	// calculate divesummary first time.
 	// this is needed in order to load the divesummary page
-	diveSummary::summaryCalculation(0, 3);
+	self.summaryCalculation(0, 3);
 }
 
 static QStringList diveSummaryTextCached;
 
 void QMLInterface::summaryCalculation(int primaryPeriod, int secondaryPeriod)
 {
-	diveSummaryTextCached = diveSummary::summaryCalculation(primaryPeriod, secondaryPeriod);
+	diveSummaryTextCached = ::summaryCalculation(primaryPeriod, secondaryPeriod);
 	emit diveSummaryTextChanged(diveSummaryTextCached);
 }
 
--- a/mobile-widgets/qmlinterface.h
+++ b/mobile-widgets/qmlinterface.h
@@ -217,7 +217,7 @@ public:
 	bool verbatim_plan() { return prefs.verbatim_plan; }
 	bool display_variations() { return prefs.display_variations; }
 
-	const QStringList &diveSummaryText() { return diveSummary::diveSummaryText; }
+	const QStringList diveSummaryText();
 
 public slots:
 	void set_cloud_verification_status(CLOUD_STATUS value) {  qPrefCloudStorage::set_cloud_verification_status(value); }
--- a/backend-shared/divesummary.cpp
+++ b/backend-shared/divesummary.cpp
@@ -1,11 +1,154 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "divesummary.h"
+#include "core/dive.h"
 #include "core/qthelper.h"
 #include "core/settings/qPrefUnit.h"
 
 #include <QDateTime>
+#include <array>
 
-QStringList diveSummary::summaryCalculation(int primaryPeriod, int secondaryPeriod)
+struct Stats {
+	Stats();
+	int dives, divesEAN, divesDeep, diveplans;
+	long divetime, depth;
+	long divetimeMax, depthMax, sacMin, sacMax;
+	long divetimeAvg, depthAvg, sacAvg;
+	long totalSACTime, totalSacVolume;
+};
+
+Stats::Stats() :
+	dives(0), divesEAN(0), divesDeep(0), diveplans(0),
+	divetime(0), depth(0), divetimeMax(0), depthMax(0),
+	sacMin(99999), sacMax(0), totalSACTime(0), totalSacVolume(0)
+{
+}
+
+static void calculateDive(struct dive *dive, Stats &stats)
+{
+	if (is_dc_planner(&dive->dc)) {
+		stats.diveplans++;
+		return;
+	}
+
+	// one more real dive
+	stats.dives++;
+
+	// sum dive in minutes and check for new max.
+	stats.divetime += dive->duration.seconds;
+	if (dive->duration.seconds > stats.divetimeMax)
+		stats.divetimeMax = dive->duration.seconds;
+
+	// sum depth in meters, check for new max. and if dive is a deep dive
+	stats.depth += dive->maxdepth.mm;
+	if (dive->maxdepth.mm > stats.depthMax)
+		stats.depthMax = dive->maxdepth.mm;
+	if (dive->maxdepth.mm > 39000)
+		stats.divesDeep++;
+
+	// sum SAC, check for new min/max.
+	if (dive->sac) {
+		stats.totalSACTime += dive->duration.seconds;
+		stats.totalSacVolume += dive->sac * dive->duration.seconds;
+		if (dive->sac < stats.sacMin)
+			stats.sacMin = dive->sac;
+		if (dive->sac > stats.sacMax)
+			stats.sacMax = dive->sac;
+	}
+
+	// EAN dive ?
+	for (int j = 0; j < dive->cylinders.nr; ++j) {
+		if (dive->cylinders.cylinders[j].gasmix.o2.permille > 210) {
+			stats.divesEAN++;
+			break;
+		}
+	}
+}
+
+// Returns a (first_dive, last_dive) pair
+static std::array<timestamp_t, 2> loopDives(timestamp_t primaryStart, timestamp_t secondaryStart, Stats &stats0, Stats &stats1)
+{
+	struct dive *dive;
+	int i;
+	timestamp_t firstDive = 0, lastDive = 0;
+
+	for_each_dive (i, dive) {
+		// remember time of oldest and newest dive
+		if (i == 0)
+			firstDive = dive->when;
+		if (dive->when > lastDive)
+			lastDive = dive->when;
+
+		// check if dive is newer than primaryStart (add to first column)
+		if (dive->when > primaryStart)
+			calculateDive(dive, stats0);
+
+		// check if dive is newer than secondaryStart (add to second column)
+		if (dive->when > secondaryStart)
+			calculateDive(dive, stats1);
+	}
+	return { firstDive, lastDive };
+}
+
+static QString timeString(long duration)
+{
+	long hours = duration / 3600;
+	long minutes = (duration - hours * 3600) / 60;
+	if (hours >= 100)
+		return QStringLiteral("%1h").arg(hours);
+	else
+		return QStringLiteral("%1:%2").arg(hours).arg(minutes, 2, 10, QChar('0'));
+}
+
+static QString depthString(long depth)
+{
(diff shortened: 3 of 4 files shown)


dirkhh:
The Android build failure appears to be a network issue where the git clone of kirigami failed. This was surprisingly hard to find in the log :-)

janiversen (commented):
This looks to me as a quite good commit, and it surely make the code clearer.

janiversen (commented):
this I believe is a matter of taste, but it cuts lines of code, so why not.

There is a typo in the commit text though, that should be corrected.

janiversen (commented):
The QML interface does not need a global variable as such, but I wanted to avoid putting Q_OBJECT in this class, therefore it cannot be used in the Q_PROPERTY of QMLInterface.

mobile-widgets/qmlinterface.cpp
```diff
@@ -95,8 +95,15 @@ void QMLInterface::setup(QQmlContext *ct)
 	diveSummary::summaryCalculation(0, 3);
 }
 
+static QStringList diveSummaryTextCached;
+
```
janiversen, line 99: The intention of this class was not to carry data or implementations. Starting to add variables is a slippery road, where we might end up with a new monolith like QMLManager. I recommend not to do this.

I know it is a global variable local to QMLInterface.cpp and not in the class, but the argument is the same.

mobile-widgets/qmlinterface.cpp
```diff
@@ -95,8 +95,15 @@ void QMLInterface::setup(QQmlContext *ct)
...
 void QMLInterface::summaryCalculation(int primaryPeriod, int secondaryPeriod)
 {
-	diveSummary::summaryCalculation(primaryPeriod, secondaryPeriod);
-	emit diveSummaryTextChanged(diveSummary::diveSummaryText);
+	diveSummaryTextCached = diveSummary::summaryCalculation(primaryPeriod, secondaryPeriod);
```
janiversen: it seems we are copying the stringlist here, is that really needed ?

mobile-widgets/qmlinterface.h
```diff
@@ -217,7 +217,7 @@ class QMLInterface : public QObject {
 	bool verbatim_plan() { return prefs.verbatim_plan; }
 	bool display_variations() { return prefs.display_variations; }
 
-	const QStringList &diveSummaryText() { return diveSummary::diveSummaryText; }
+	const QStringList diveSummaryText();
```
janiversen, line 220: As far as I can see, the code is just move into the cpp file (of course due to the static variable), but it breaks with how the rest is implemented, without (it seems to me) no real reason.

janiversen (commented):
Surely it is only a bunch of functions, but something that might expand over time, most of the more modern code in this project uses classes as limitations. Starting to use namespaces competing with classes is not a way to make the code easier to understand. I have a strong preference to keep it as a class (and expand that as we go along).

mobile-widgets/qmlinterface.cpp
bstoeger: > it seems we are copying the stringlist here

Perhaps surprisingly we aren't. This is for two reasons:
1) Qt's containers use CopyOnWrite semantics. An assignment of two containers does no deep copy, but simply increases a reference count. A deep copy will only be performed if you modify one of the containers. This has some nasty implications, by the way. You will even find recommendations to pass QString, etc. by value in signals to avoid subtle time-of-life or concurrent-access bugs. Matter of fact, we did have horrible-to-debug issues because the QModelIndex we got in a signal suddenly changed under our feet. Lesson learned: Qt's proxy models aren't reentrant.
2) Even if this were standard non-COW containers, which I strongly prefer, this would be subject to return value optimization and copy elision: https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization
Take for example the code
```
#include <cstdio>

struct X {
	X() { printf("Construct\n"); }
	X(const X &) { printf("Copy construct\n"); }
	X &operator=(const X &) { printf("Copy assign\n"); return *this; }
	X &operator=(X &&) { printf("Move\n"); return *this; }
};

X f()
{
	return X();
}

int main()
{
	X x1 = f();
	printf("--\n");
	x1 = f();
}
```
This outputs:
```
Construct
--
Construct
Move
```
There is not a single copy (by assignment or construction) performed. A move operation of a container is usually more-or-less free. I could imagine that on modern CPUs it is even more efficient than the indirection caused by references! But don't quote me on that.

janiversen: thanks for the explanation, then the only remaining problem is that QMLInterface is no longer a pure interface, but carries data.

dirkhh: I remember our conversation that you felt that it being a pure interface was important, but if asked I don't think I could reconstruct that argument... in other words, would you please re-explain that to me (I know you already did... I am simply very slow when it comes to internalizing object oriented programming subtleties).

bstoeger: > [...] pure interface  [...] object oriented programming  subtleties.

Sorry for being that annoying person, but `QMLInterface` is *not* an interface class in the sense that you would typically use the word in OO programming (especially Java and other languages that don't have multiple inheritance).

In OO context, an interface *defines* methods that a class has to provide to fulfill that interface but it does not *implement* them. Classes can implement one or more interfaces by providing the appropriate functions.

In C++ terms, an interface class would consist only of pure virtual functions, i.e. function of the type
```
   virtual void f() = 0;
```
`QMLInterface` is, in contrast, very much an *implementation* of a number of functions.
Of course that all has nothing to do with your question, it's just clarification of terminology. Also note that I expressly do *not* suggest to rename the thing.

janiversen: please QMLInterface does not contain implementations, but merely calls to classes/functions that contain the actual implementation, in that respect it is a pure interface. This type of interfaces are often found as API, allowing a library to only expose the methods it wants to expose. But please this is not a terminology discussion, making the right choice is far more  important, because the choice have consequences. 

My concern is to avoid another monolith like qmlmanager, if we start allowing implementation code in qmlinterface, the simplicity will soon be gone, right now it serves one purpose and one purpose only, to be interface to QML, it hold the Q_PROPERTY etc. definitions.

If we allow implementations we can start to move quite a lot of mobile specific code from e.g. diveplannerpointsmodel into qmlinterface and then we end having a mixture of actual implemention code and interface calls, which soon becomes interesting to maintain.

The goal was to simplify the QML interface and make it clear what code is interface. If we move away from that we should move qmlinterface into qmlmanager to at least avoid building 2 momoliths.

dirkhh: Please allow me to try to insert myself as a translator here :-)

Ignoring for a moment the correct definition of what an "interface class" is...

I believe what Jan is talking about makes sense. Having a "clean class" (to invent a new term) between the UI and the backend seems useful. This exposes the methods and signals and accessors that the QML code needs, and simply then calls shared code (or sometimes mobile specific code that isn't UI related) to collect the data, but doesn't hold any of that data itself.

Can I claim that this is absolutely necessary for Subsurface-mobile to be created?
No, I don't think I can.
Can I claim that I understand that this is a useful construct to make things cleaner and easier to understand?
Yeah, possibly.

Maybe this helped move the conversation forward :-)

If not, feel free to curse me and move on arguing :-(

dirkhh:
One thing that I pointed out to Jan in my initial review when the DiveSummary was merged was that I felt that this unstructured `QStringList` was a bit scary and extremely fragile.

@janiversen - how hard would it be to turn this into a clean interface (to reuse my term) as well?
So something that can be used from QML like
`diveSummary1.deepDives` and `diveSummary2.sacMax`, etc?
Which would then be connected to the two structures that @bstoeger is proposing in this RFC?

janiversen:
I think the 2 structures make a lot of sense (as pointed out in my review comments. It is not hard but a lot of typing to make each variable available in qml.

I would prefer not to register 2 new interfaces, as I am trying to cut down on the many we already have, but I believe it can be done a a subclass (class within a class) of qmlinterface. So you would reference like backend.divesummary1.<variable>

dirkhh:
> I think the 2 structures make a lot of sense (as pointed out in my review comments. It is not hard but a lot of typing to make each variable available in qml.

I'm happy to give this a try... having something to implement would be a fun distraction :-)

> I would prefer not to register 2 new interfaces, as I am trying to cut down on the many we already have, but I believe it can be done a a subclass (class within a class) of qmlinterface. So you would reference like backend.divesummary1.

That seems reasonable - and a property alias could hide it in the QML code

janiversen:
It should be pretty easy, define a class within qminterface with Q_PROPERTY corresponding to the struct and of course get/set functions as well as signals.

Then you need to add 2 Q_PROPERTY in qmlinterface e.g. divesummary1 and divesummary2, without a notify. the get/set functions refer 2 instances of the class defined previously.

Disclaimer: This is my thoughts not implementation, there are surely a couple of details I have forgotten.

If you make it as a PR or branch I can help you with tips as you get the code done.

janiversen:
After a good nights sleep, I think the design we made yesterday caused quite a lot of extra code for no good reason.

The better solution is to:
- Make class divesummary a simple model, delivering row of name,value1,value2. Doing that gives a number of advantages:
Display  in both desktop and mobile, becomes independent of the actual content, by using a list view and a delegate.
The content is consistent between desktop and mobile
name is only translated once
no big additions to qmlinterface, since the model (based on abstract model) is automatically available to QML.

Furthermore the calculation function should be void. In the current proposal (where it returns the string list), a calculation will be done every time the screen updates. Think visible attribute.

The challenge in this way is to find an elegant way to connect the model (basically a list) to the struct or change the struct into something different.

This has a huge impact on this PR, meaning most of the commit needs to be changed.

bstoeger:
Both suggestions of structured objects and a model appear viable and are definitely preferred over unstructured QStringList. Both come with a rather annoying amount of boilerplate, but such is the nature of Qt's model/view and property systems. I have implemented a number of models, therefore I believe that I can easily come up with a sketch of both solutions for comparison purpose.

> delivering row of name,value1, value2

Name would conceptually be part of the header, in this case vertical. This avoids having to do nasty formatting tricks, as the header is rendered differently. But this is also where the problems start: on desktop we might prefer a horizontal layout, etc.

The fundamental question is how close are the layout of desktop and mobile. If they are very close, a model is probably preferred, if they differ, structured result objects provide more flexibility at the cost of slightly higher maintenance when adding new fields. I.e. the field has to be added to the UI.

janiversen:
If you make a sketch, you need to make the QML part as well, otherwise it is only a half sketch.
Using a model makes the QML part a play in the sun (list view with a delegate), while using single names means a lot more typing. Making a sketch is of course the maximum usage of time :-)

With my experience, I would not make a sketch but immediately make a model. Dirk seemed to want something to implement and at least making the QML might be fun for him (I would presume making the model would also be fun).

Name is not part of the header, see image

Captura de pantalla 2020-02-05 a las 12 30 22

each row have its own name, and the header is fixed. You might note that there are separations between some rows, that could be easily implemented by name = "", and testing for that condition
in the delegate.

Divesummary is not yet implemented in the desktop, so yes it can be assumed to be 100% identical.

Forgot one thing, the model should add e.g. l/min - cuft/min to value1 and value2 so the UI use the name, value1, value2 directly.

bstoeger:
> Name is not part of the header, see image

Note that I said "conceptually", as I'm not sure how QML handles vertical headers. And yes, the names are very much a vertical header. Conceptually, that is.

> Divesummary is not yet implemented in the desktop, so yes it can be assumed to be 100% identical.

I can't parse that sentence. Do you suggest that the UI has to be the same because it is not yet implemented?

> that could be easily implemented by name = "", and testing for that condition
in the delegate

That is an argument against the model - with objects you get maximum flexibility in formatting, if that is what you prefer, at a rather small cost.

janiversen:
name are not implemented as a header, but merely as a normal label with font bold. name = "" was just my way of making life simpler, of course the delegate can make the spacing without that. If at some time name should be formatted as a header, that is something to be done locally in the delegate without changing the model (who is basically only providing the data in a structured row by row format). I know a model can do special header handling, but that is really not needed.

I suggested that the desktop UI could very well be the same, since it is not implemented. I did not suggest it have to be the same. But in order to make progress we need to make assumption of the parts not yet implemented, and it strikes me as a normal assumption to assume they are identical seem from the POW of the model.

Your "rather small cost" is not so small after all. Adding a field would involve:
- in case of a struct.
1) add field to divesummary and of course the calculation
2) add row to desktop including name (both ui and cpp files)
3) get name translated for the desktop
4) add field to QMLInterface
5) add row to QML including name
6) get name translated for QML
- in case of a model:
1) add field to divesummary and of course the calculation
2) translate name in divesummary
- no changes to the delegate of desktop and QML, since they automatically present the rows the 
model delivers.

Considering that big difference in maintenance costs, I have no doubt a model is the better choice.
And with a model you also have a lot of flexibility in how to present it, by programming the delegate.

PR #2591, opened 2020-02-05
bugfix: solve spurious loss of signal/slots.

janiversen:

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR solves a problem of missing signal/slots from QMLInterface. With the connect() being in the constructor that constructs a global object, there are no guarantee that the referenced objects are constructed, and thus sometimes leads to the connect not being successful. Unfortunate Qt does not report an error when this happens.

Solution was to move the connect() back into setup(), while keeping the nice lambda changes.

In contrary to what I wrote earlier, we (a student who writes a paper on clang and myself) have, out of scientific interest, examined the clang compiler (Xcode 11.3.1) compiling for iOS, other compilers might behave differently.

We found that the global space is created in 3 steps:
1) all public const (we are still debating if a class const is added at this step, it seems to depend on compiler switches). These are created in a special const segment, and are created by the linker.
2) all public global variables, including classes. The sequence seems very random, changing the compile order is no guarantee of changing the sequence of constructor calls. Linker allocates the global space needed, and the constructors are called as part of the startup code.
3) all private static variables, including those hidden in e.g. instance(). Because instance()
does:
       static <class> *self = new <class>;
       return self;

self is a hidden (compiler scope is local) global object, and it is constructed before instance() is called. We have seen hints that some optimization switches might cause a delayed construct, but for debug it is constructed as part of the global space creation.
4) main is called

Based on these finding, it seems like a good idea, never to reference parallel global objects in a global constructor (actually this is also taught here in the C++ class).

Some of the reasons why this has not been detected before can be:
1) most of the connects relate to the UI, and a missing change in a single graphical element is easily overlooked
2) There are very little formal testing of the signal/slot (apart from qPref)
3) Bulk of the connects are in desktop-widget, that connect a library (like e.g. core or backend-shared) to the UI, while QMLInterface interconnect signals.

remark, this was a practical study using Xcode 11.3.1 and IPadOS 13.1 simulator. I am sure reading the C++ specs contains some descriptions as well, but theory and practice are often two different pair of shoes.

Any positive comment and/or suggestions are mostly welcome.

### DIFF
--- a/mobile-widgets/qmlinterface.h
+++ b/mobile-widgets/qmlinterface.h
@@ -86,7 +86,7 @@ public:
 	static QMLInterface *instance();
 	
 	// function to do the needed setup
-	static void setup(QQmlContext *ct);
+	void setup(QQmlContext *ct);
 
 	// Duplicated enums, these enums are properly defined in the C/C++ structure
 	// but duplicated here to make them available to QML.
@@ -319,6 +319,6 @@ signals:
 
 	void diveSummaryTextChanged(QStringList);
 private:
-	QMLInterface();
+	QMLInterface() {}
 };
 #endif // QMLINTERFACE_H
--- a/mobile-widgets/qmlinterface.h
+++ b/mobile-widgets/qmlinterface.h
@@ -81,6 +81,10 @@ class QMLInterface : public QObject {
 	Q_PROPERTY(QStringList diveSummaryText READ diveSummaryText NOTIFY diveSummaryTextChanged);
 
 public:
+	// function to address the object, which is not needed in production code,
+	// but highly needed when testing signal/slot.
+	static QMLInterface *instance();
+	
 	// function to do the needed setup
 	static void setup(QQmlContext *ct);
 
--- a/mobile-widgets/qmlinterface.cpp
+++ b/mobile-widgets/qmlinterface.cpp
@@ -2,6 +2,12 @@
 #include "qmlinterface.h"
 #include <QQmlEngine>
 
+QMLInterface *QMLInterface::instance()
+{
+	static QMLInterface *self = new QMLInterface;
+	return self;
+}
+
 QMLInterface::QMLInterface()
 {
 	// relink signals to QML
@@ -84,8 +90,7 @@ QMLInterface::QMLInterface()
 void QMLInterface::setup(QQmlContext *ct)
 {
 	// Register interface class
-	static QMLInterface self;
-	ct->setContextProperty("Backend", &self);
+	ct->setContextProperty("Backend", instance());
 
 	// Make enums available as types
 	qmlRegisterUncreatableType<QMLInterface>("org.subsurfacedivelog.mobile",1,0,"Enums","Enum is not a type");
--- a/mobile-widgets/qmlinterface.cpp
+++ b/mobile-widgets/qmlinterface.cpp
@@ -8,11 +8,21 @@ QMLInterface *QMLInterface::instance()
 	return self;
 }
 
-QMLInterface::QMLInterface()
+void QMLInterface::setup(QQmlContext *ct)
 {
+	// Register interface class
+	ct->setContextProperty("Backend", instance());
+
+	// Make enums available as types
+	qmlRegisterUncreatableType<QMLInterface>("org.subsurfacedivelog.mobile",1,0,"Enums","Enum is not a type");
+
+	// calculate divesummary first time.
+	// this is needed in order to load the divesummary page
+	diveSummary::summaryCalculation(0, 3);
+
 	// relink signals to QML
 	connect(qPrefCloudStorage::instance(), &qPrefCloudStorage::cloud_verification_statusChanged,
-		[this] (int value) { emit cloud_verification_statusChanged(CLOUD_STATUS(value)); });
+			[this] (int value) { emit cloud_verification_statusChanged(CLOUD_STATUS(value)); });
 	connect(qPrefUnits::instance(), &qPrefUnits::duration_unitsChanged,
 			[this] (int value) { emit duration_unitsChanged(DURATION(value)); });
 	connect(qPrefUnits::instance(), &qPrefUnits::lengthChanged,
@@ -85,19 +95,6 @@ QMLInterface::QMLInterface()
 			this, &QMLInterface::verbatim_planChanged);
 	connect(qPrefDivePlanner::instance(), &qPrefDivePlanner::display_variationsChanged,
 			this, &QMLInterface::display_variationsChanged);
-}
-
-void QMLInterface::setup(QQmlContext *ct)
-{
-	// Register interface class
-	ct->setContextProperty("Backend", instance());
-
-	// Make enums available as types
-	qmlRegisterUncreatableType<QMLInterface>("org.subsurfacedivelog.mobile",1,0,"Enums","Enum is not a type");
-
-	// calculate divesummary first time.
-	// this is needed in order to load the divesummary page
-	diveSummary::summaryCalculation(0, 3);
 }
 
 void QMLInterface::summaryCalculation(int primaryPeriod, int secondaryPeriod)
--- a/subsurface-helper.cpp
+++ b/subsurface-helper.cpp
@@ -188,7 +188,7 @@ void register_qml_types(QQmlEngine *engine)
 		QQmlContext *ct = engine->rootContext();
 
 		// Register qml interface classes
-		QMLInterface::setup(ct);
+		QMLInterface::instance()->setup(ct);
 		themeInterface::setup(ct);
 	}
 


bstoeger:
> self is a hidden (compiler scope is local) global object, and it is constructed before instance() is called. We have seen hints that some optimization switches might cause a delayed construct, but for debug it is constructed as part of the global space creation.

This is not how C++ works.
`self` is *constructed* (in your case copy-constructed with a pointer generated by new) in a thread safe manner the first time the function is called.
If you suspect that your compiler is broken, I suggest adding `printf`s to the constructor of the generated object. If your suspicion is substantiated, file a defect report to your compiler vendor. This seems like a rather massive breakage as this is a very common idiom, so you might want to contact the appropriate channels.

bstoeger:
To expand on that - if you believe that order-of-initialization is the culprit, first identify the signal that is presumably lost. Then add something like
```
        printf("QMLInterface::QMLInterface(): %p\n", this)
```
to the constructors of the respective objects.
And also `printf` the pointers to the objects before the connect() calls.
Finally, `printf` in the slots of the lost signals to see if they arrive.

That's should prove any order-of-initialization issues and we can figure out if it is
1) A compiler bug (unlikely)
2) A bug in Qt (a bit mire likely)
3) A bug in our code base (most likely)

mobile-widgets/qmlinterface.h
```diff
@@ -315,6 +319,6 @@ public slots:
 
 	void diveSummaryTextChanged(QStringList);
 private:
-	QMLInterface();
+	QMLInterface() {};
```
bstoeger: Could we stop inlining constructors without reason, please?
And sorry for being pedantic: The semi-colon after a function-definition here is superfluous, by the way.

janiversen: The semicolon was my error (I participate in 2 other projects, where it is demanded), it is gone.

What is the technical reason for not inlining the constructor ?

Our code have plenty of examples (all of qPref f.x.). It is a very convenient way to declare the constructor private (avoid construction of objects) and at the same time tell it is empty.

janiversen:
Are you really suggesting to use printf ? We use qDebug() as written in the PR. In iOS it is not recommended to use printf.

It seems to be different signals, but we have not spent time on pointing the actual signals. Since it is a spurious behaviour we can never be sure to catch all problems.

But earlier today while rebasing branches we identified a new problem with connect() in the constructor.

Our upcoming profilewidget2 connect signals from QMLInterface to slots in Profilewidget2, at the same time QMLInterface connect a few signals from ProfileWidget2, and not not forget that also is a good reason to have instance(), which was removed.

Interconnecting classes is perfectly legal in Qt and is one of the purposes of signal/slot instead of a simple call. However if connect are in the constructor this becomes interesting. Following your definition of how C++ works, lets assume QMLInterface::interface() is called first, this will cause QMLInterface::QMLInterface() to be called.
QMLInterface() calls ProfileWidget2::instance() which will cause ProfileWidget2::ProfileWidget2() to be called.
ProfileWidget2() call QMLInterface::instance() but the QMLInterface object has not finished its construction, probably leading to interesting problems.

We have not tried this because having a setup() is a lot simpler and safe, without needing to consider these behaviours.

dirkhh:
I think was Berthold is pointing out that the behavior that you are describing doesn't seem to be consistent with the fix that is proposed. And that using `printf` is a very simple and quick way to figure out what's going wrong.
I find it likely that there is some circular object dependency in your code (it doesn't appear to be present in ours, but you obviously have a ton more code in your fork) which you are breaking by doing this two phase initialization - which isn't really a fix, it's simply a way to hide the problem.

Your specific example of `QMLInterface` calling `ProfileWidget` which is calling `QMLInterface` sounds very much like a circular dependency to me. And that I think is where we should start addressing the problem.

janiversen:
well my point was exactly that having the connect() in the constructor introduces the circular dependency, when connect() is in setup() there are no circular dependency.

One of the ideas behind signal/slot compared to a simple call is to allow classes to use each other without creating a circular dependency. QMLInterface and ProfileWidget2 is good example if that, it is not bad code that they use each others signal, it only becomes bad when the connect() is in the constructor because that introduces a dependency.

A 2phase initialization is something used very frequently and per se not a bad design.

janiversen:
I can see that my arguments, are not written in an understandable way, since it seems you are all searching for errors, while the current demand of having connect() in the constructor in reality is a feature limitation.

Let me try with a variant of a classical Qt example (advocating for signal slot)

Consider this code snippet:

```
class backend {
   void bigCalculation()
   {
       .....
       UIInterface::instance()->WeDidIt();
   }
}

class UIInterface {
    void doBigCalculation()
    {
        // from UI execute button
        ....
        backend:instance()->:bigCalculation()
   }

    void WedDidIt()
    {
        // make waving flag visible in UI
    }
} 
```

This is perfectly valid code (also used this project in e.g. the exportFuncs, and in several places in desktop-widgets) and it works without problems.
However it has one big drawback the UI is unresponsive while calculating, in order to make
the calculation asynchronous we decide to use signal/slot after all that is one of the main reasons they were introduced. The following example follow the current rules of this project:

```
class backend {
   backend()
  {
     // this constructor need the UIInterface object (constructor call completed)
      connect(UIInterface::instance(), pleaseCalculate, this, doBigCalculation)
   }

   void bigCalculation()
   {
       .....
       emit calculationFinished();
   }
}

class UIInterface {
    UIInterface()
    {
        // this constructor need the UIInterface object (constructor call completed)
        connect(backend::instance(), calculationFinished, this, WeDidIt)
    }

    void doBigCalculation()
    {
        // from UI execute button
        ....
       emit pleaseCalculate()
   }

    void WedDidIt()
    {
        // make waving flag visible in UI
    }
} 
```

and we have introduced a circular dependency in the constructors which are causing "funny" problems, so in simple words, this type of normal async. connections are not permitted in this project.

This PR offers a simple solution (which can be found in a lot of Qt applications):

```
class backend {
   backend()
  {
    // this constructor need the UIInterface object (constructor call completed)
    // which is no problem since the UIInterface constructor is empty
       connect(UIInterface::instance(), pleaseCalculate, this, doBigCalculation)
   }

   void bigCalculation()
   {
       .....
       emit calculationFinished();
   }
}

class UIInterface {
    setup()
    {
       // the UIInterface constructor no longer depends on the backend object,
      // and thus the circular dependency is broken
        connect(backend::instance(), calculationFinished, this, WeDidIt)
    }

    void doBigCalculation()
    {
        // from UI execute button
        ....
       emit pleaseCalculate()
   }

    void WedDidIt()
    {
        // make waving flag visible in UI
    }
} 
```

I trust this example is understandable. I do hope that you can also see, why this is a perfectly normal use of signal/slots and not a design error that needs to be corrected.

The PR solves the 2 problems with the current QMLInterface:
a) it reintroduces instance() which among others allow backend classes to catch signals
b) it moves the connect() back into a setup() to avoid the circular dependency.

I assume we (I do) want a user interface that is highly responsive, and not being blocked when doing heavier things, which can be as simple as applying a new sort in a list.

Please reconsider this limiting demand that connect() MUST be in the constructor, only exception is when it is a dynamic connect (ref. comments from bstoeger).

bstoeger:
> it seems you are all searching for errors

Understandably so, after all the title of the PR reads as "bugfix: solve spurious loss if signal/slot" and in your initial post you accuse clang of breaking C++ in a very fundamental way. These are very disconcerting points (we *do* rely on deterministic order of construction using this idiom in quite a lot of places), and of course they demand some investigation.

```
class backend {
   backend()
  {
     // this constructor need the UIInterface object (constructor call completed)
      connect(UIInterface::instance(), pleaseCalculate, this, doBigCalculation)
   }
}
class UIInterface {
    UIInterface()
    {
        // this constructor need the UIInterface object (constructor call completed)
        connect(backend::instance(), calculationFinished, this, WeDidIt)
    }
}
```
NAK! I have semi-jokingly referred to `connect()` as a version of COMEFROM, but this is taking it to new levels. If you establish an (biderectional or not) connection between objects, do it in one place. I have spent quite some energy on detecting and removing these kind of cyclic dependencies and don't want them reintroduced. Avoiding this kind of chaos is exactly why I do not want two-phase initialization (unless there are very good reasons).

It is true that connect() can be used to transport data into a different thread, but I'm not seeing anything of that kind here. In your example, why can't `UIInterface` simply call into `doBigCalculation()` and then get a call back later? This all appears to me very badly thought out. And it also makes me somewhat skeptical concerning the QMLInterface class. I thought this should only be a "pure interface", I don't see how this fits this description.

> I assume we (I do) want a user interface that is highly responsive

I'm sure that you don't mean it that way, just be aware that this comes across as very disingenuous. Executing code in the background does not mean that there has to be these kind of cyclic dependencies.

In summary:
1) ACK on the instance() part, though I find it a bit sad and it also gives me the impression that QMLInterface is growing to something that it was not originally meant to be.
2) NAK on the two-phase initialization. Now more than before. If there are good arguments, I will reconsider that - but the given example is precisely what I want to avoid.

janiversen:
OK, I respect your arguments even though I clearly think differently.

Closing PR, allowing me to concentrate on other matters.

PR #2594, opened 2020-02-06
Fix mobile settings ui

dirkhh:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

As I started preparing for the mobile beta, I realized that once again I did a shitty job reviewing the code. This PR deals _only_ with the theme settings.

### Changes made:

1) fixed the visual appearance of the swatches. See the commits for details
2) fixed the scale buttons to be placed where they belong, instead of on top of on top of the theme drop down

### Mentions:

@janiversen : I admit that I was rather disappointed when I tried to do the betas and realized that this was broken in an incredibly obvious manner - one that made me doubt that you had tested this at all. 
You keep asking me to trust you because you test every commit and things are always perfect. And I try to be open to your ideas and under time pressure (of my own doing) at times merge things just reading the code and not running / testing in sufficient detail.
This PR fixes the issues with one small part of the visual appearance that I found. But there are a number more. For example the very very fragile "get the width right across all the different devices" code is, unfortunately, broken.
Which means that I will once again spend time on trying to fix this - and we risk that our users who were happy that things finally worked, may end up with a broken release again.

This is why I absolutely HATE to have us change working code just because it doesn't match your ideas of how it should work.

### DIFF
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -321,23 +321,23 @@ TemplatePage {
 					spacing: Kirigami.Units.largeSpacing
 					TemplateButton {
 						text: qsTr("smaller")
-						enabled: ThemeNew.currentScale !== 0.85
+						enabled: subsurfaceTheme.currentScale !== 0.85
 						onClicked: {
-							ThemeNew.currentScale = 0.85
+							subsurfaceTheme.currentScale = 0.85
 						}
 					}
 					TemplateButton {
 						text: qsTr("regular")
-						enabled: ThemeNew.currentScale !== 1.0
+						enabled: subsurfaceTheme.currentScale !== 1.0
 						onClicked: {
-							ThemeNew.currentScale = 1.0
+							subsurfaceTheme.currentScale = 1.0
 						}
 					}
 					TemplateButton {
 						text: qsTr("larger")
-						enabled: ThemeNew.currentScale !== 1.15
+						enabled: subsurfaceTheme.currentScale !== 1.15
 						onClicked: {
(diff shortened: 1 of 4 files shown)


janiversen:
I do test every commit, but not after each of the often many rebase you ask for (very often with good reason) and surely not after the changes you make (I am not suggesting that you are introducing bugs! but merely telling what I test).

However, this is all history, apart from 575 commits from me that are already merged, I will not be your problem anymore.

dirkhh:
> However, this is all history, apart from 575 commits from me that are already merged, I will not be your problem anymore.

Understood. Thank you for your contributions.

With that, I think we will change course and remove quite a few of the things that went in to prepare for future work on the planner, etc. No point in leaving those around. That doesn't affect this PR (as the folding settings are kinda neat), but I will spend the next couple of days on removing all those things that make no sense to have hanging around.

janiversen:
Why not simply revert the last bunch of theme commits (back to where the C++ version was activated), just an idea and bring the working code back. It is surely a lot faster than testing everything again, and I suspect nobody will want to complete the work anyhow. I am very sure that if you revert back to before QMLInterface bstoeger will be very happy.

ps. I gave unsubscribed from notifications unless my name is mentiomed.

Subsurface-divelog:
OK, this seems to deal with most of the obvious UI issues.
I'm planning to spend more time on my own handful of devices to make sure I've caught at least the worst of the worst.
For those of you who can build for devices - this may be a good time to do so :-)

PR #2596, opened 2020-02-06
Cleanup: un-classify export library

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is bike-shedding, but I see no reason for a class that is only composed of functions. The only state it had was a QFuture, but that seems to be better kept at the caller's site anyway. Reasoning: Why not allow multiple notifications that are closed automatically?

The one thing I have doubts is the namespace thing - we didn't have one before, so why would we need it now. On the other hand it doesn't seem to hurt. On the third hand, `Export::export_TeX` does feel a bit redundant - so one might keep the namespace and remove the prefix, etc. I really don't know.

I tested this on desktop and still could export files.

@neolit123 @dirkhh: You might have opinions on the namespace thing.

### DIFF
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -12,7 +12,7 @@
 #include "core/divesite.h"
 #include "exportfuncs.h"
 
-void exportProfile(QString filename, const bool selected_only)
+void exportProfile(QString filename, bool selected_only)
 {
 	struct dive *dive;
 	int i;
@@ -33,7 +33,7 @@ void exportProfile(QString filename, const bool selected_only)
 }
 
 
-void export_TeX(const char *filename, const bool selected_only, bool plain)
+void export_TeX(const char *filename, bool selected_only, bool plain)
 {
 	FILE *f;
 	QDir texdir = QFileInfo(filename).dir();
@@ -233,7 +233,7 @@ void export_TeX(const char *filename, const bool selected_only, bool plain)
 
 }
 
-void export_depths(const char *filename, const bool selected_only)
+void export_depths(const char *filename, bool selected_only)
 {
 	FILE *f;
 	struct dive *dive;
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -311,8 +311,8 @@ std::vector<const dive_site *> exportFuncs::getDiveSitesToExport(bool selectedOn
 	return res;
 }
 
-void exportFuncs::exportUsingStyleSheet(QString filename, bool doExport, int units,
+QFuture<int> exportFuncs::exportUsingStyleSheet(QString filename, bool doExport, int units,
 	QString stylesheet, bool anonymize)
 {
-	future = QtConcurrent::run(export_dives_xslt, filename.toUtf8(), doExport, units, stylesheet.toUtf8(), anonymize);
+	return QtConcurrent::run(export_dives_xslt, filename.toUtf8(), doExport, units, stylesheet.toUtf8(), anonymize);
 }
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -4,6 +4,7 @@
 #include <QtConcurrent>
 #include "core/membuffer.h"
 #include "core/divesite.h"
+#include "core/gettextfromc.h"
 #include "core/tag.h"
 #include "core/file.h"
 #include "core/errorhelper.h"
@@ -11,14 +12,7 @@
 #include "core/divesite.h"
 #include "exportfuncs.h"
 
-
-exportFuncs *exportFuncs::instance()
-{
-    static exportFuncs *self = new exportFuncs;
-    return self;
-}
-
-void exportFuncs::exportProfile(QString filename, const bool selected_only)
+void exportProfile(QString filename, const bool selected_only)
 {
 	struct dive *dive;
 	int i;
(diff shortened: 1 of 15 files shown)


backend-shared/exportfuncs.cpp
```diff
@@ -4,21 +4,17 @@
...
 #include "core/divesite.h"
+#include "core/gettextfromc.h"
 #include "core/tag.h"
 #include "core/file.h"
 #include "core/errorhelper.h"
 #include "core/divefilter.h"
 #include "core/divesite.h"
 #include "exportfuncs.h"
 
+namespace Export {
```
neolit123: a global singleton pattern seems to be preferred in Subsurface.
`class->instance()->doFunc()`
no strong opinions on my side.

desktop-widgets/divelogexportdialog.cpp
```diff
@@ -131,6 +131,9 @@ void DiveLogExportDialog::on_exportGroup_buttonClicked(QAbstractButton*)
 	showExplanation();
 }
 
+// We use one future for all exports. This is no problem as we also have only one notification widget.
+static QFuture<void> future;
```
neolit123: can this be a member of the dialog instance?

bstoeger: I think so, yes. However, KMessageWidget is "part of the KDE libraries" according to the file header, so I didn't want to touch it.

backend-shared/exportfuncs.cpp
bstoeger: Heh. I'm always arguing the other way round. :-P

Note that these functions have no global state -> why would we want an instance and even more strangely an instance that can send/receive signals, i.e. a QObject. The analogy I see is more qthelper.cpp than mainwindow.cpp.

Checking the history - indeed these functions were lifted the `DiveLogExportDialog` class. But note that this class does *not* have an instance method. This makes sense to me: why wouldn't you allow the user to open more than one of these dialogs (on OSs that tend to be non-modal)?

desktop-widgets/divelogexportdialog.cpp
bstoeger: Oh - I think I misunderstood - you suggest to make it a member of DiveLogExportDialog, not of KMessageWidget. Sure, this can be done.

backend-shared/exportfuncs.cpp
neolit123: maybe these can just be public static functions in a class.

one problem i had with namespaces is that they can be defined in multiple files and if the project is poorly organized one has to `grep` to see what goes in the namespace.

bstoeger: My gut feeling would be to simply remove the namespace. A user of these functions `#include`s the header file and just calls them. They are marked with export_ anyway so there should be no risk of name-clash and the names can be `grep`ed.

desktop-widgets/divelogexportdialog.cpp
bstoeger: Stupid question: Do we even have to keep a QFuture object? Once we've passed it down to the notification widget - can't we destroy the original? I'm not familiar with copy-semantics of QFuture (I assume it is copied in the QFutureWatcher).

neolit123: i do not know for sure. examples like this one are keeping it around as a class member: http://codejourneys.blogspot.com/2008/06/qt-simple-example-of-use-of.html

backend-shared/exportfuncs.cpp
neolit123: possibly better, yes.

Subsurface-divelog: I find namespaces frustrating (we have a couple) as they break my pattern of grepping for stuff. In a class at least the .cpp file contains `ClassName::method()` so I can find that easily. In a namespace I can't.
So my preference would be that we keep this a class - but maybe not a `QObject`?
Otherwise I just don't think this adds a lot of value

Subsurface-divelog (approved):
Reasonable cleanup. Needs testing.
Not sure that it's a massive improvement.
Let's call this a SOFT approval.

desktop-widgets/divelogexportdialog.cpp
bstoeger: This example has a number of problems - new but no delete...

According to the Qt docs, QFuture is designed to be passed around as value:
```
QFuture is a lightweight reference counted class that can be passed by value.
```
I tested it on a huge log, and it worked flawlessly. Let's remove that global variable.

backend-shared/exportfuncs.cpp
bstoeger: @dirkhh, @neolit123:I am aware that this is idle bike-shedding!

> In a class at least the .cpp file contains ClassName::method() so I can find that easily. In a namespace I can't.

Fair enough. Though I must say that personally I never `grep` for the namespace - `grep`ing for the function itself has always been sufficiently unambiguous.

My take on namespaces: Their use is primarily for libraries of *different* vendors to avoid name-clashes between exported functions. In C you would prefix every function with gtk_, in C++ you define a namespace Gtk. So I don't really see the use for a non-library project. For the undo-commands, I created a namespace for convenience/laziness - I didn't want to prefix functions that outside code isn't supposed to touch anyway with with command_

Here I went the namespace route to make the changes less drastic. However, the functions are prefixed with export anyway, so why bother with a class *or* namespace? As I said further up: If you want to use the export functions, include the appropriate header and just use them.

Has anyone ever complained about all our core-C functions not being stuck in classes? I certainly haven't missed that.

Therefore, I propose to simply remove the class. I have no problem with defining a class full of static functions. But frankly - I don't see the appeal.

dirkhh: I'm perfectly happy with not having class or namespace

dirkhh (approved):
Remove the namespace completely and be done with it.

dirkhh:
I assume you'll refactor this one as well and force push it so it can be merged?

bstoeger:
Rebased on master. From my side, this is ready to go, though low priority because only a code hygiene thing.

That said, I wonder if the `backend-shared` directory is all that meaningful. Isn't everything that's outside of `mobile-widgets` and `desktop-widgets` shared by definition?

dirkhh:
Happy to entertain changes to any of that. This was all introduced in order to allow for the now abandoned planner.

dirkhh (approved):
merging as is

PR #2599, opened 2020-02-08
mobile: prevent crash adding dives

dirkhh:
When the cylinders became a dynamic data structure, a get_cylinder() call
suddenly could return a NULL pointer. So use get_or_create_cylinder() for the
first call when parsing the user's data.

Also, deal with an oddity where the string lists look different because an
empty list technically isn't the same as a list with one empty string.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

1) normalize (for some definition of that term) the `QStringList` we get from QML
2) first time we access a cylinder, make sure we create it if it doesn't exist

### Release note:

I don't think we have a public version with that bug, but I added an entry, just in case

### Documentation change:

no

### Mentions:

@bstoeger

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: fix potential crash when adding / editing dives
 Mobile: automatically scroll the dive edit screen so that the notes edit cursor stays visible
 Desktop: ignore dive sites without location in proximity search
 Mobile: add personalized option for units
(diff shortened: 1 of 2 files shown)


bstoeger (approved):
Not tested, but looks good to me.
Turning the cylinders into dynamic arrays was a massive infrastructure change and some breakage was to be expected. :-/

PR #2601, opened 2020-02-08
Single column portrait

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a feature request from a user. When we changed the screen scaling to work correctly on a lot more devices, inadvertently their screen suddenly had 2 columns in portrait mode and they really disliked that because it made the profile picture so much smaller.

So this adds an option to force single column mode in portrait orientation.

### Changes made:

1) add preference
2) add settings
3) change size calculations

### Release note:

done

### Documentation change:

missing

### Mentions:

@willemferguson for the user manual
@bstoeger for the settings and the QML code (remember, you're now our QML expert...) :-)

### DIFF
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -41,6 +41,9 @@ static const QByteArray st_windowState_default;
 int qPrefDisplay::st_lastState;
 static int st_lastState_default = false;
 
+bool qPrefDisplay::st_singleColumnPortrait;
+static bool st_singleColumnPortrait_default = false;
+
 qPrefDisplay *qPrefDisplay::instance()
 {
 	static qPrefDisplay *self = new qPrefDisplay;
@@ -66,7 +69,7 @@ void qPrefDisplay::loadSync(bool doSync)
 		load_geometry();
 		load_windowState();
 		load_lastState();
-
+		load_singleColumnPortrait();
 	}
 }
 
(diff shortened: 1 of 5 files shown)


core/settings/qPrefDisplay.cpp
```diff
@@ -41,6 +41,9 @@ static const QByteArray st_windowState_default;
 int qPrefDisplay::st_lastState;
 static int st_lastState_default = false;
 
+bool qPrefDisplay::st_singleColumnPortrait;
+static bool st_singleColumnPortrait_default = false;
```
bstoeger, line 45: You appear to be following a scheme in this file, but shouldn't it be
```static const bool st_singleColumnPortrait_default = false;```
or even
```static const constexpr bool st_singleColumnPortrait_default = false;```
?
Not a practical issue though - any modern compiler will recognize that you never change the thing and remove it.

bstoeger (approved):
I don't see anything wrong with this PR. However, this doesn't mean anything as I'm not familiar with any part of the code it touches.

> remember, you're now our QML expert...

I'm thrilled. Can I also become qPref expert, please?

dirkhh:
> > remember, you're now our QML expert...
> 
> I'm thrilled. 

I thought you would be.

> Can I also become qPref expert, please?

You are hereby anointed to be expert and sole ruler of qPref and related messes.

core/settings/qPrefDisplay.cpp
dirkhh: I really just followed the pattern. I assume that you will rip all of this out, soon.

bstoeger: > I assume that you will rip all of this out, soon.

There are more important things, but it is on the todo list, yes. :-P

PR #2602, opened 2020-02-08
Dive list: be more careful on when updating the UI after selection

### LINKED ISSUE #2595: Dive indive list does not get scope when selected
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

1) Open a dive log
2) Select a trip heading. The trip notes are displayed.
3) Select a dive from this trip. The Notes tab does not change, still shows trip notes.
    The dive notes are not shown. I would expect it show the dive notes, not the trip notes.

### Mentions:

@dirkhh @bstoeger

bstoeger:
Fix two issues:

1) When narrowing the selection, we didn't get setSelection()
   calls. Only, when the user released the mouse button was
   the selection updated. Therefore, hook into the mouse-release-
   event and update the UI if the selection changed.

2) We updated the ui in setSelection(). However, this was called
   on mouse-move even if the actual selection didn't change.
   Therefore, compare selection before and after processing of
   the event and only refresh the UI if there are changes.

Clearly, this can only be a quick stopgap solution and we
should find out how to properly hook into the selection change
machinery. Though see commit 4928c4ae0421193bbd371cb0924091a970489611
for the reason why we do things as we do them.

Fixes #2595

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes the selection bug reported in #2595. The code is a catastrophe, but is the only way I found to fix the problem reported in commit 4928c4ae0421193bbd371cb0924091a970489611. Sigh - I'm not getting the selection problems under control. :-/

### Changes made:

1) Hook into selection changes in mouseReleaseEvent
2) Only update UI when selection actually changes

### Related issues:

Fixes #2595.
### Release note:

Not needed - bug not in release.
### Documentation change:

Not needed - only bug fix.
### Mentions:

@willemferguson @dirkhh

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -533,10 +533,44 @@ void DiveListView::currentChanged(const QModelIndex &current, const QModelIndex&
 	scrollTo(current);
 }
 
+void DiveListView::mouseReleaseEvent(QMouseEvent *event)
+{
+	// Oy vey. We hook into QTreeView's setSelection() to update the UI
+	// after selection changes. However, there is a case when this doesn't
+	// work: when narrowing the selection. If multiple dives are selected,
+	// and the user clicks on one of them, the selection is unchanged.
+	// Only on mouse-release the selection is changed, but then
+	// setSelection() is not called.
+	// Notably, this happens when the user selects a trip and the clicks
+	// on a dive in the same trip.
+	// To solve this, we hook into the mouseReleseEvent here and detect
+	// selection changes changes by comparing the selection before and after
+	// processing of the event.
+	// We really should find another way to solve this.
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
You will remember that way back when in the ancient days when people still dealt with their own problems and didn't expect "toolkits" or "frameworks" to do things for them, we had implemented our own selection code completely independent from Qt.
Then a brave knight entered the arena and said "behold, I will fix this mess" and started using the Qt selection framework.
Yeah, Tomaz and I (and at some point even Linus) were fighting the illogical behavior of this for quite a while. I look forward to the permanent solution to all this - likely when we port to Qt 8. :-P

PR #2603, opened 2020-02-08
[TRIVIAL] Cleanup: remove erroneous comments

bstoeger:
Remove two erroneous comments stating that a function-local
QSettings variable should not be static because it is initialized
too early. Scoped static variables are initialized when execution
first hits the statement.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is an idle removal of an erroneous comment, which I stumbled upon when git grep-ing for some qPref stuff.
In case someone doesn't believe me and is too lazy to test, here's the corresponding part of a working draft of the ISO standard:
https://timsong-cpp.github.io/cppwp/n3337/stmt.dcl
On "block-scope variables": "Otherwise [when it can't be statically initialized, rem.] such a variable is initialized the first time control passes through its declaration; such a variable is considered initialized upon the completion of its initialization."

### Changes made:

1) Remove comments.

### DIFF
--- a/core/settings/qPrefPrivate.cpp
+++ b/core/settings/qPrefPrivate.cpp
@@ -18,10 +18,6 @@ QString keyFromGroupAndName(QString group, QString name)
 
 void qPrefPrivate::propSetValue(const QString &key, const QVariant &value, const QVariant &defaultValue)
 {
-	// REMARK: making s static (which would be logical) does NOT work
-	// because it gets initialized too early.
-	// Having it as a local variable is light weight, because it is an
-	// interface class.
 	QSettings s;
 	bool isDefault = false;
 	if (value.isValid() && value.type() == QVariant::Double)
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
So the interesting thing is that in one of the PRs that lead to this `s` was static and things went very weird. I'm not disagreeing with your statement here, I'm just saying that this comment was based on an observation.

This changes no code, so I'll merge without waiting for GitHub Actions to finish

bstoeger:
> So the interesting thing is that in one of the PRs that lead to this s was static and things went very weird.

Perhaps a QSession object doesn't get updates when other QSession objects change things? Or it keeps some locks? Or it can't be accessed from different threads? Could be many things.

PR #2607, opened 2020-02-09
Break git pull command in CONTRIBUTING.md into three separate commands

jb2cool:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The contribution instructions originally had the cd, git checkout and git pull all on one command
this is not allowed and gave a syntax/command error. When i looked at the CONTRIBUTING.md file
the commands were on three separate lines it was just the markdown backticks that were missing from showing up as three separate lines. 

### Changes made:

I have simply wrapped each of these lines in its own backticks.

### Additional information:

Signed-off-by: Jason Bramwell <jb2cool@gmail.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,9 +13,9 @@ If you would like to contribute patches that fix bugs or add new features, that 
 Here is a very brief introduction on creating commits that you can either send as [pull requests][6] on our main GitHub repository or send as emails to the mailing list. Much more details on how to use git can be found at the [git user manual][7].
 
 Start with getting the latest source (look at the [Building Page][8] to find out how).  
-`cd subsurface  
-git checkout master  
-git pull`  
+`cd subsurface`  
+`git checkout master`  
+`git pull`  
 ok, now we know you're on the latest version. Create a working branch to keep your development in:  
 `git checkout -b devel`  
 Edit the code (or documentation), compile, test… then create a commit:  
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ If you would like to contribute patches that fix bugs or add new features, that 
 
 Here is a very brief introduction on creating commits that you can either send as [pull requests][6] on our main GitHub repository or send as emails to the mailing list. Much more details on how to use git can be found at the [git user manual][7].
 
-Start with getting the latest source (look at the [Building Page][8] to find out how).  
+Start with getting the latest source.  
 `cd subsurface`  
 `git checkout master`  
 `git pull`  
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -13,9 +13,9 @@ If you would like to contribute patches that fix bugs or add new features, that
 Here is a very brief introduction on creating commits that you can either send as [pull requests][6] on our main GitHub repository or send as emails to the mailing list. Much more details on how to use git can be found at the [git user manual][7].
 
 Start with getting the latest source (look at the [Building Page][8] to find out how).  
-`cd subsurface  
-git checkout master  
-git pull`  
+`cd subsurface`  
+`git checkout master`  
+`git pull`
```
dirkhh, line 18: an alternative would be to have triple backticks at the beginning and end (and I'm not sure how to even represent this in markdown...)

dirkhh (approved):
I think this is fine. The .csv files should indeed be just comments - this isn't parsed anywhere.
The markdown could be done differently, but either way is fine.

CONTRIBUTING.md
```diff
@@ -84,8 +84,7 @@ In order to make reviews simpler and have contributions merged faster in the cod
...
 [7]: https://www.kernel.org/pub/software/scm/git/docs/user-manual.html
-[8]: https://subsurface-divelog.org/building/
-[9]: https://gerrit-review.googlesource.com/Documentation/user-signedoffby.html
-[10]: https://subsurface-divelog.org/wp-content/uploads/2011/10/Screenshot-gitk-subsurface-1.png "Example with gitk"
-[11]: https://github.com/Subsurface-divelog/subsurface/blob/master/CodingStyle
-[12]: https://github.com/Subsurface-divelog/subsurface/blob/master/CONVENTIONS.md
+[8]: https://gerrit-review.googlesource.com/Documentation/user-signedoffby.html
+[9]: https://subsurface-divelog.org/wp-content/uploads/2011/10/Screenshot-gitk-subsurface-1.png "Example with gitk"
+[10]: https://github.com/Subsurface-divelog/subsurface/blob/master/CodingStyle
+[11]: https://github.com/Subsurface-divelog/subsurface/blob/master/CONVENTIONS.md
```
dirkhh: this changes the link references, without changing the references to those links (i.e., where we reference `[9]` in the file we should reference `[8]` now, correct?

jb2cool: Damnit, you're right. I'll fix that now.

jb2cool: Ok, that should now be fixed.

Not making a great first impression really am i, a bit like an overenthusiastic puppy!

dirkhh: You are making an outstanding impression

#1 thing I value in a contributor? Positive response to feedback on a pull request. And then, ideally, updates based on that feedback...

dirkhh:
Now... that commit message for that last commit is a bit oddly formatted on github, mainly because the subject (first line of the commit) is too long.
And ideally, really, those last two commits should be one.
So, in your branch, locally, do this:
```
git rebase -i HEAD~3
```
An editor will open with three lines starting with `pick`. On the third of those lines change `pick` into `f` and save the file. Now the last two commits will be combined into one and the offending commit message will be discarded.
Now force push to github :-)

jb2cool:
Thanks for your help with this Dirk, i think we are done (for now at least). Is there anything else that i need to do from my side or is all good and just waiting for merging from your side?

dirkhh:
The commit message is still a bit too wide - have you looked at it on GitHub?
Instead of playing "bring me another rock", let me fix that locally (instead of merging on GitHub) and then you can look at it when it hits master in a few minutes...

PR #2610, opened 2020-02-10
Move filtering closer to the core

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

To avoid ProxyModel-Lasagna, Dirk suggested filtering the dives directly at the core. Indeed, the DiveTripModels have to handle filtering anyway. We hope that this helps in unifying the desktop and mobile models.

This is an attempt at doing so. It is only very lightly tested though, so it *will* break widely. Testing would be very welcome. Note that this is not as easy as one might think. The interesting parts are when modifying dives changes the filter-criterion; when whole trips are hidden and the interaction with the undo-system.

The main commit is admittedly a tad big and should probably be split up.

### Changes made:

1) Move filtering to the DiveTripModels.
2) Remove filtering from the MultiFilterProxyModel (which now has an inappropriate name, btw).
3) Remove some now unused code.

### Release note:

No.
### Documentation change:

No.
### Mentions:

@dirkhh @neolit123

### DIFF
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -68,11 +68,9 @@ dive *DiveListBase::addDive(DiveToAdd &d)
 	}
 	dive *res = d.dive.release();		// Give up ownership of dive
 
-	// Set the filter flag according to current filter settings
-	bool show = DiveFilter::instance()->showDive(res);
-	res->hidden_by_filter = !show;
-	if (show)
-		++shown_dives;
+	// When we add dives, we start in hidden-by-filter status. Once all
+	// dives have been added, their status will be updated.
+	res->hidden_by_filter = true;
 
 	int idx = dive_table_get_insertion_index(&dive_table, res);
 	add_to_dive_table(&dive_table, idx, res);	// Return ownership to backend
@@ -185,18 +183,20 @@ DivesAndSitesToRemove DiveListBase::addDives(DivesAndTripsToAdd &toAdd)
 		  [](const DiveToAdd &d, const DiveToAdd &d2)
 		  { return dive_less_than(d.dive.get(), d2.dive.get()); });
 
-	// Remember old number of shown dives
-	int oldShown = shown_dives;
-
 	// Now, add the dives
 	// Note: the idiomatic STL-way would be std::transform, but let's use a loop since
 	// that is closer to classical C-style.
 	auto it2 = res.rbegin();
+	QVector<dive *> divesToFilter;
+	divesToFilter.reserve(toAdd.dives.size());
 	for (auto it = toAdd.dives.rbegin(); it != toAdd.dives.rend(); ++it, ++it2) {
 		*it2 = addDive(*it);
 		dives.push_back({ (*it2)->divetrip, *it2 });
+		divesToFilter.push_back(*it2);
 	}
 	toAdd.dives.clear();
+
+	ShownChange change = DiveFilter::instance()->update(divesToFilter);
 
 	// If the dives belong to new trips, add these as well.
 	// Remember the pointers so that we can later check if a trip was newly added
@@ -224,7 +224,7 @@ DivesAndSitesToRemove DiveListBase::addDives(DivesAndTripsToAdd &toAdd)
 		emit diveListNotifier.divesAdded(trip, createTrip, divesInTrip);
 	});
 
-	if (oldShown != shown_dives)
+	if (!change.newShown.empty() || !change.newHidden.empty())
 		emit diveListNotifier.numShownChanged();
 
 	return { res, sites };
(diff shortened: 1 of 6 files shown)


dirkhh:
Conceptual question. If a dive is hidden, then how does a change make it unhidden?
Maybe some undo or something? Just trying to figure out if we are making our lives too hard?
For example, wouldn't it be much easier to do this:
- connect to the "something changed in the divelist" signal
- remove filter and add filter with same criterion
- prosper

bstoeger:
> Conceptual question. If a dive is hidden, then how does a change make it unhidden?
Maybe some undo or something?

Precisely.
1) Change buddy "A" to "B".
2) Filter for buddy "B".
3) Undo. <- dive has to disappear

Or:

1) Filter for buddy "A".
2) Change buddy "A" to "B" <- dive has to disappear
3) Undo <- dive has to appear

> remove filter and add filter with same criterion

I reckon you suggest to fully reload the filter after every change to dives? I'm not sure it is that simple, but I will think about it.

bstoeger:
Also note, that's pretty much what the code attempts to do anyway:
```
	ShownChange shownChange = updateShown(dives);
	divesShown(trip, shownChange.newShown);
	divesHidden(trip, shownChange.newHidden);
```
For all changed dives, recalculate the filter and add if newly shown, remove if newly hidden. That's in principle the same as when the whole filter is reset, just limited to the changed dives.

That brings me to another complication when testing: tree vs flat mode. :-P

dirkhh:
Here are some of the tests I tried (all on desktop, of course)
(1) tree view, filter by rating
- change rating of currently shown dive
-- dive list updates (dive is now hidden), including the totals for the trip
-- dive itself stays shown (it's debatable if that's a bug)
- undo
-- dive list updates
-- dive is still shown

(2) same thing in flat view (sort by date)
- same results

(3) tree view, filter by tag
- delete a dive that is shown
-- dive is deleted, count is updated, next dive is shown
- undo
- dive returns, counts are updated, un-deleted dive is selected

(4) filter for different tag which makes the deleted dive hidden
- undo / redo sequence
-- no visible changes, as it should be

(5) filter by temperature
- currently visible dive is one that gets hidden, but stays visible (same as above)
-- deleting that dives via redo now changes things so that the first visible dive is selected
-- undo keeps that visible dive selected (that seems like the correct behavior)

(6) now add filter by person
-- correct dives stay visible
- invert the filter
-- different dive becomes visible, but its trip isn't expanded in the dive list (that's a bug, but it seems like a corner case and not really an issue)

So by and large it seems like this is on the right track, @bstoeger

dirkhh:
iOS builds are having timeout issues right now.
Funnily enough, in a completely different context I have been talking to GitHub management about some of the concerns around GitHub Actions that run on macOS environments (like the iOS builds) and I understand their pain.
Right now I'm selectively restarting these to ensure that we don't break iOS by mistake, but I realize that this may be an issue for a while

bstoeger:
First of all, thanks for testing!

> dive itself stays shown (it's debatable if that's a bug)

Since we *do* hide dives when the filter is edited, I consider this as a bug. Should be fixed now.

> different dive becomes visible, but its trip isn't expanded in the dive list (that's a bug, but it seems like a corner case and not really an issue)

Yes, the whole trip-expansion business. I have to admit that I haven't really thought about it up to now.

dirkhh:
Where are we with this? More testing needed? Consider for merging?

bstoeger:
Depends on how important the trip-expansion is to you. Also, if you could double-check that the current dive is now set correctly, that would be great (I'm already quite tired for today).

dirkhh:
not super important :-)
And yes,will test some more
Get some sleep

bstoeger:
Added a small code-reshuffling commit: Loop in the actual filter code, not the model. This makes the way we filter more flexible (think fulltext index) and also is, arguably, more logically organized.

dirkhh (approved):
Did some more testing - I think this looks good.
If you feel this is ready to merge, please change the Subject. After that I'll read through the changes one more time (because I really did a poor job with reviews, lately) and then hopefully merge

dirkhh:
I guess @bstoeger is enjoying a weekend with family... but I'm hesitant to merge this without him chiming in :-)

@atdotde , did you have a chance to test this? Anyone else?

bstoeger:
Sorry - I was distracted. Merge in principle fine with me, I just improved the last commit. Therefore, let's wait for CI. If you have doubts concerning the last commit, feel free to leave it out. Thanks.

dirkhh:
So this could already be merged in order to make #2625 smaller? Because I think this one is certainly fine...

bstoeger:
> So this could already be merged in order to make #2625 smaller? Because I think this one is certainly fine...

It *is* somewhat scary, but yes - from my side it's ready to go.

PR #2617, opened 2020-02-12
mobile UI: restructure menu

dirkhh:
This changes the 'GPS' entry to be called 'Location', moves the 'Map' entry
below that, and reuses the map icon for both the map inside this submenu and
for the main menu item.

It moves the 'About' entry under 'Help' and both 'Export' and 'Dive summary'
under 'Dive management'.

This way we have only five (or with 'Developer' enabled, six) entries in the
main menu making it much more appropriate for really tiny screens.

Additionally, the entrieis moved into sub-menus are ones that are not all that
commonly used.

Screen Shot 2020-02-12 at 8 55 30 AM

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

discussion on the developer mailing list

### Release note:

not sure if this is worth an entry?

### Documentation change:

obviously this will once again require new pictures and possibly some text changes...

### Mentions:

@willemferguson

### DIFF
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -241,15 +241,6 @@ Kirigami.ApplicationWindow {
 			},
 			Kirigami.Action {
 				icon {
-					name: ":/icons/map-globe.svg"
-				}
-				text: mapPage.title
-				onTriggered: {
-					showMap()
-				}
-			},
-			Kirigami.Action {
-				icon {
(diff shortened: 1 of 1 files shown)


willemferguson:
I think this is pretty useful. I will do user manual after this has been released as a beta.

PR #2618, opened 2020-02-12
[for testing] GitHub Actions/iOS: make less noisy

dirkhh:
It seems that our iOS builds are killed because they create too much output
(and frankly, the amount of output this creates is ridiculous, I should have
caught this months ago).

- [x] Build system change


dirkhh:
Yeah, that's not it.
It's the silly "you have to log in to you Qt account" requirement that they just added.
Yikes. Stupid.
OK, plan B

PR #2619, opened 2020-02-13
GitHub Actions/iOS: shift to using a pre-packaged Qt installation

dirkhh:
Since the official Qt binaries can no longer be installed without disclosing
credentials (well, sure, that could be done through secrets), I decided that
we should go back to packaging just the part of the iOS Qt SDK that we need.

- [x] Build system change

This will hopefully fix the iOS GitHub Action
(yeah, as if I ever get that right on the first try...)

### Changes made:

instead of installing from the official Qt binaries (which now requires logging into a Qt account) we go back to packaging just a small (~160MB) tar file with the binaries that we need to build for iOS.
There is a lot of room for further optimization to only pack those libraries that we actually touch, but I didn't feel motivated to go down too far into this rat hole.

### Related issues:

All of the currently failing iOS GitHub Actions


dirkhh:
Yay, this worked. Merging so that future iOS builds based on the new master will no longer show red.

PR #2621, opened 2020-02-14
Fix git-load string handling

### LINKED ISSUE #2613: Divemode changes become bookmarks after saving and reopening
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I manually added a bailout event to my last dive in subsurface. The icon on the profile is the little curved arrow with a line through it.

I saved, closed and reopened subsurface the icon for the bailout had disappeared and was replaced with a dive flag ('normal' bookmark with a note of 'OC @26:20')

This also applies if I switch to pSCR mode and then back to CCR mode.

### Operating system:

Windows 10, english uk

Shearwater petrel 2 (hardwired), firmware version 71

### Subsurface version:
subsurface  4.9.3
 Does the same happen on another Subsurface version? don't know 
 official release

### Steps to reproduce:
- Diving on CCR (Shearwater) in CCR mode
- During the dive I bailout and switch to BO mode on my computer.

1. When I import the dive into subsurface, I don't see the bailout event.
2. When I manually add a change divemode event (bailout to OC, back on CCR ) the event appears and looks fine and shows the OC divemode (bailout) icon. The Surface GF at end of the dive reflects the bailout accurately.
3. I save and quit.
4. When I open subsurface and look at the dive, I see a 'dive flag' bookmark in the profile, but no bailout icon. Further, the surface GF is calculated as if I had stayed on CCR through the whole dive
5. I remove the event (the note says 'OC @ [time]')
6. Repeat steps 2,3,4 - same result.

### Current behavior:
see description

### Expected behavior:
divemode change event persists as  between saves and affects tissue calculation (surface GF etc.)

torvalds:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This makes it much easier to handle multiple strings when parsing key value pairs, and in the process fixes the problem that Robert Helling @atdotde reported where event lines with a divemode key would get a corrupted name (the divemode string would becomes the name of the event).

It also adds a fixup to such corruption, since we know what the event name must be ("modechange") if it has a divemode string in it.

### Related issues:

Fixes #2613

### DIFF
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -761,6 +761,7 @@ static int get_divemode(const char *divemodestring) {
 struct parse_event {
 	const char *name;
 	struct event ev;
+	int has_divemode;
 };
 
 static void parse_event_keyvalue(void *_parse, const char *key, const char *value)
@@ -778,6 +779,7 @@ static void parse_event_keyvalue(void *_parse, const char *key, const char *valu
 		parse->name = value;
 	} else if (!strcmp(key,"divemode")) {
 		parse->ev.value = get_divemode(value);
+		parse->has_divemode = 1;
 	} else if (!strcmp(key, "cylinder")) {
 		/* NOTE! We add one here as a marker that "yes, we got a cylinder index" */
 		parse->ev.gas.index = 1 + get_index(value);
@@ -833,6 +835,10 @@ static void parse_dc_event(char *line, struct membuffer *str, struct git_parser_
 			break;
 		line = parse_keyvalue_entry(parse_event_keyvalue, &p, line, str);
 	}
+
+	/* Only modechange events should have a divemode - fix up any corrupted names */
+	if (p.has_divemode && strcmp(p.name, "modechange"))
+		p.name = "modechange";
 
 	ev = add_event(state->active_dc, p.ev.time.seconds, p.ev.type, p.ev.flags, p.ev.value, p.name);
 
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -45,21 +45,6 @@ struct keyword_action {
 
 static git_blob *git_tree_entry_blob(git_repository *repo, const git_tree_entry *entry);
 
-static char *get_utf8(struct membuffer *b)
-{
-	int len = b->len;
-	char *res;
-
-	if (!len)
-		return NULL;
-	res = malloc(len + 1);
-	if (res) {
-		memcpy(res, b->buffer, len);
-		res[len] = 0;
-	}
-	return res;
-}
-
 static temperature_t get_temperature(const char *line)
 {
 	temperature_t t;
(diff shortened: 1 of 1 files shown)


core/load-git.c
```diff
@@ -725,28 +750,41 @@ static int get_divemode(const char *divemodestring) {
...
-		event->gas.index = 1 + get_index(value);
+		parse->ev.gas.index = 1 + get_index(value);
 	} else if (!strcmp(key, "o2")) {
-		event->gas.mix.o2 = get_fraction(value);
+		parse->ev.gas.mix.o2 = get_fraction(value);
 	} else if (!strcmp(key, "he")) {
-		event->gas.mix.he = get_fraction(value);
+		parse->ev.gas.mix.he = get_fraction(value);
 	} else
 		report_error("Unexpected event key/value pair (%s/%s)", key, value);
```
dirkhh, line 791: not your code, but a coding style violation... should have `{ }` around the `report_error()`

dirkhh (commented):
Very fragile code in so many ways.
I did a review by reading the code, but realistically that's unlikely to catch any other subtle issues.
I really should figure out how to create a test for this :-(
I will do some manual testing later today, I hope others (especially @atdotde ) do the same.

core/load-git.c
```diff
@@ -834,6 +836,10 @@ static void parse_dc_event(char *line, struct membuffer *str, struct git_parser_
 		line = parse_keyvalue_entry(parse_event_keyvalue, &p, line, str);
 	}
 
+	/* Only modechange events should have a divemode - fix up any corrupted names */
+	if (p.has_divemode && strcmp(p.name, "modechange"))
+		p.name = "modechange";
```
dirkhh, line 841: after fixing things up we should mark the dive list as modified and make sure that the cache for this dive is invalidated. Otherwise this fix only gets written out if the user changes anything else in that dive...

dirkhh (changes requested):
doing some testing shows that by and large this seems to work well... but the fact that we normally don't write out a fixed version is unfortunate... maybe you can address that?

core/load-git.c
torvalds: My thinking was that it didn't matter on disk, since we have the code to fix it up anyway.

The important part is that it's fixed up in our data structures, so that we show the event correctly etc.

We have other places where we do somewhat similar fixups when loading. In fact, some of them are in that same function (look for the "Older logs might mark the dive to be CCR by having .." comment just below).

There are actually many more subtle changes we do, where we might save in a different format than we loaded in (because we have various "don't save if the value hasn't changed" heuristics etc, and those heuristics have changed over the years).

So I don't think we need to mark it dirty.

But it certainly wouldn't be *wrong* to mark it dirty, so I'm not against it. It's just that I don't really expect anybody to read the git save format directly outside of subsurface and care about it...

dirkhh: that's a valid point. I was thinking about this as a "get the stored format correct" exercise, but in reality that's not relevant. What matters is that we can read what's on disk and do the right thing with it.

dirkhh (approved):
agreeing with Linus' argument... this is good as far as I can see.

dirkhh:
Oh, could you add
```
Fixes #2613 
```
to the second commit's commit message?

atdotde (approved):
This indeed fixes the problem without obviously introducing new ones. But please have a look at #2622 which might be a lot simpler.

PR #2622, opened 2020-02-14
Fix loading bailout events

### LINKED ISSUE #2613: Divemode changes become bookmarks after saving and reopening

atdotde:
The parsing of events mistakes the new divemode name
as the event name. Let's use the fact that dive modes
have distinct names from other events and simply
fix those events whose name matches a divemode.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a bit of a hack since it relies on the "only one string per line" feature of git loading. But it is simple. If a dive mode name ends up as an event name, make that a mode change event and fix the value.

### Related issues:

Fixes #2613

### DIFF
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -722,7 +722,7 @@ static int get_divemode(const char *divemodestring) {
 		if (!strcmp(divemodestring, divemode_text[i]))
 			return i;
 	}
-	return 0;
+	return -1;
 }
 
 static void parse_event_keyvalue(void *_event, const char *key, const char *value)
@@ -799,6 +799,11 @@ static void parse_dc_event(char *line, struct membuffer *str, struct git_parser_
 	name = "";
 	if (str->len)
 		name = mb_cstring(str);
+	int divemode = get_divemode(name);
+	if (divemode >= 0) {
+		name = strdup("modechange");
+		event.value = divemode;
+	}
 	ev = add_event(state->active_dc, event.time.seconds, event.type, event.flags, event.value, name);
 
 	/*


dirkhh:
Would you agree that it makes sense to use #2621 instead?
Also, would you review that one, please?

atdotde:
Yes, #2621 attacks the problem at a more fundamental level. But this here might be "good enough".

dirkhh:
You are correct that #2621 is more complex - but it also fixes a bigger problem and increases the likelihood that the next time we add things to the format we don't re-break things (knock wood).

My preference would be to go with that, TBH.

atdotde:
Fine with me

PR #2627, opened 2020-02-17
Mobile profile enhancements

dirkhh:

- [x] Functional change
- [x] New feature

1) allow temporary switching to verbose mode from QML UI
2) add x/y offset to the QML profile
3) enable pinching for scale and dragging in order to pan the zoomed in profile

### Release note:

done

### Documentation change:

done

### Mentions:

@bstoeger not much C++ to review - but remember, you are the designated QML guru :-)
@willemferguson I know you are traveling, but if you see this and want to check the documentation update, that would be neat
@atdotde can you test this?
anyone else? @janmulder has been very quiet lately :-(

### DIFF
--- a/mobile-widgets/qml/DiveDetailsView.qml
+++ b/mobile-widgets/qml/DiveDetailsView.qml
@@ -218,20 +218,24 @@ Item {
 		rowSpacing: Kirigami.Units.smallSpacing * 2
 		columnSpacing: Kirigami.Units.smallSpacing
 
-		QMLProfile {
-			id: qmlProfile
-			visible: !noDive
+		Rectangle {
 			Layout.fillWidth: true
 			Layout.preferredHeight: Layout.minimumHeight
 			Layout.minimumHeight: width * 0.75
 			Layout.columnSpan: 3
-			clip: false
-			Rectangle {
-				color: "transparent"
-				opacity: 0.6
-				border.width: 1
-				border.color: subsurfaceTheme.primaryColor
+			clip: true
+
+			QMLProfile {
+				id: qmlProfile
+				visible: !noDive
 				anchors.fill: parent
+				Rectangle {
+					color: "transparent"
+					opacity: 0.6
+					border.width: 1
+					border.color: subsurfaceTheme.primaryColor
+					anchors.fill: parent
+				}
 			}
 		}
 		Controls.Label {
--- a/mobile-widgets/qml/DiveDetailsView.qml
+++ b/mobile-widgets/qml/DiveDetailsView.qml
@@ -1,8 +1,5 @@
 // SPDX-License-Identifier: GPL-2.0
-import QtQuick 2.6
-/*
-import QtWebView 1.0
-*/
+import QtQuick 2.10
 import QtQuick.Dialogs 1.2
 import QtQuick.Layouts 1.2
 import QtQuick.Controls 2.2 as Controls
@@ -229,12 +226,77 @@ Item {
 				id: qmlProfile
 				visible: !noDive
 				anchors.fill: parent
+				clip: true
+				property real lastScale: 1.0 // final scale at the end of previous pinch
 				Rectangle {
 					color: "transparent"
 					opacity: 0.6
 					border.width: 1
 					border.color: subsurfaceTheme.primaryColor
 					anchors.fill: parent
+				}
+				PinchArea {
+					anchors.fill: parent
+					pinch.dragAxis: Pinch.XAndYAxis
+					onPinchStarted: {
+						if (manager.verboseEnabebled)
+							manager.appendTextToLog("pinch started w/ previousScale " + qmlProfile.lastScale)
+					}
+					onPinchUpdated: {
+						if (pinch.scale * qmlProfile.lastScale < 1.0)
+							qmlProfile.lastScale = 1.0 / pinch.scale // this way we never shrink and the changes stay smooth
+						// the underlying widget deals with the scaling, no need to send an update request
+						qmlProfile.scale = pinch.scale * qmlProfile.lastScale
+						if (manager.verboseEnabled)
+							manager.appendTextToLog("pinch updated to scale " + qmlProfile.scale);
+					}
+					onPinchFinished: {
+						// remember the final scale value so we can continue from there next time the user pinches
+						qmlProfile.lastScale = pinch.scale * qmlProfile.lastScale
+					}
+
+					MouseArea {
+						anchors.fill: parent
+						drag.target: qmlProfile
+						drag.axis: Drag.XAndYAxis
+						drag.smoothed: true
+						pressAndHoldInterval: 150 // very short - feels about right
+						// cursor/finger position as we start dragging
+						property real initialX
+						property real initialY
+						// the offset previously used to show the profile
+						property real oldXOffset
+						property real oldYOffset
+						// this indicates that we are actually dragging
+						property bool dragging: false
+						onPressAndHold: {
+							dragging = true;
+							oldXOffset = qmlProfile.xOffset
+							oldYOffset = qmlProfile.yOffset
+							initialX = mouse.x
+							initialY = mouse.y
+							if (manager.verboseEnabled)
+								manager.appendTextToLog("press and hold at mouse" + Math.round(10 * mouse.x) / 10 + " / " + Math.round(10 * mouse.y) / 10)
+							// give visual feedback to the user that they now can drag
+							qmlProfile.opacity = 0.5
+						}
+						onPositionChanged: {
+							if (dragging) {
+								var x = (mouse.x - initialX) / qmlProfile.scale
+								var y = (mouse.y - initialY) / qmlProfile.scale
+								if (manager.verboseEnabled)
+									manager.appendTextToLog("drag mouse "  + Math.round(10 * mouse.x) / 10 + " / " + Math.round(10 * mouse.y) / 10 + " delta " + Math.round(x) + " / " + Math.round(y))
+								qmlProfile.xOffset = oldXOffset + x
+								qmlProfile.yOffset = oldYOffset + y
+								qmlProfile.update()
+							}
+						}
+						onReleased: {
+							// reset things
+							dragging = false
+							qmlProfile.opacity = 1.0
+						}
+					}
 				}
 			}
 		}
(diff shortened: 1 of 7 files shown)


profile-widget/qmlprofile.cpp
```diff
@@ -23,6 +27,9 @@ QMLProfile::QMLProfile(QQuickItem *parent) :
 
 void QMLProfile::paint(QPainter *painter)
 {
+	QElapsedTimer timer;
+	timer.start();
```
bstoeger: Since the result is only read if `verbose == true`, perhaps also put `timer.start()` in a `if (verbose)`?

profile-widget/qmlprofile.cpp
```diff
@@ -85,19 +94,24 @@ int QMLProfile::diveId() const
...
+		qDebug() << reason << "dive #" << d->number << "x" << QString::number(m_xOffset, 'f', 1) << "y" << QString::number(m_yOffset, 'f', 1);
 	m_profileWidget->plotDive(d, true);
 }
 
+void QMLProfile::setDiveId(int diveId)
+{
+	m_diveId = diveId;
+	if (m_diveId < 0)
+		return;
+	updateProfile(QString("setDiveId"));
```
bstoeger: Is the `QString` even needed. I would imagine that `const char *` is automatically cast to `QString` when passed as an argument.

If you want to use the `QString`, use `QStringLiteral` instead for the 1-2 μs gain. :-P

profile-widget/qmlprofile.cpp
```diff
@@ -113,6 +127,34 @@ void QMLProfile::setDevicePixelRatio(qreal dpr)
 	}
 }
 
+qreal QMLProfile::xOffset() const
```
bstoeger: Do we really need the getter function? Couldn't we simply declare the PROPERTY as MEMBER?

I'm not a fan of getter functions that do nothing.

profile-widget/qmlprofile.cpp
```diff
@@ -113,6 +127,34 @@ void QMLProfile::setDevicePixelRatio(qreal dpr)
...
+// don't update the profile here, have the user update x and y and then manually trigger an update
+void QMLProfile::setXOffset(qreal value)
+{
+	if (IS_FP_SAME(value, m_xOffset))
+		return;
+	m_xOffset = value;
+	emit xOffsetChanged();
+}
+
+qreal QMLProfile::yOffset() const
```
bstoeger: See above: getter function needed?

profile-widget/qmlprofile.h
```diff
@@ -10,6 +10,8 @@ class QMLProfile : public QQuickPaintedItem
 	Q_OBJECT
 	Q_PROPERTY(int diveId MEMBER m_diveId WRITE setDiveId)
 	Q_PROPERTY(qreal devicePixelRatio READ devicePixelRatio WRITE setDevicePixelRatio NOTIFY devicePixelRatioChanged)
+	Q_PROPERTY(qreal xOffset READ xOffset WRITE setXOffset NOTIFY xOffsetChanged)
+	Q_PROPERTY(qreal yOffset READ yOffset WRITE setYOffset NOTIFY yOffsetChanged)
```
bstoeger: See above: MEMBER instead of READ?

mobile-widgets/qml/DiveDetailsView.qml
```diff
@@ -218,20 +215,89 @@ Item {
...
+				clip: true
+				property real lastScale: 1.0 // final scale at the end of previous pinch
+				Rectangle {
+					color: "transparent"
+					opacity: 0.6
+					border.width: 1
+					border.color: subsurfaceTheme.primaryColor
+					anchors.fill: parent
+				}
+				PinchArea {
```
bstoeger, line 238: I just learned what "pinching" is in mobile UIs. Interesting.

bstoeger (commented):
Just a few minor comments, after skimming over the code. Will test later. If I find out how to simulate the "pinching gesture" on desktop-on-mobile, that is.

profile-widget/qmlprofile.cpp
dirkhh: I thought about it but was too lazy. Will rectify that tardiness.

profile-widget/qmlprofile.cpp
dirkhh: The string was useful when I started debugging the code... and frankly, if we ever run into a situation where we get runaway update requests, we might be grateful for it. But I can of course optimize this with `QStringLiterals`

profile-widget/qmlprofile.cpp
dirkhh: I thought that if I do `MEMBER` then I can have neither setter nor getter. And since the setter does something other than update the value, this seemed like the way to go. And yes, we need to be able to read this value from QML, so we can't make it write-only :)

profile-widget/qmlprofile.cpp
dirkhh: same answer :-)

mobile-widgets/qml/DiveDetailsView.qml
dirkhh: Ha. I don't believe this for one second.
Unless, of course, you still have a flip phone.
Given some of your other comments earlier that does seem like a possibility :-P

bstoeger: I knew the "motion", I didn't know what it was called. Off-topic: I detest when Google maps turns the north somewhere else because I'm not doing a perfectly vertical motion. Going to search how to prevent that right now...

profile-widget/qmlprofile.cpp
bstoeger: We do have a few cases of `MEMBER` + `WRITE`:
```
$ git grep MEMBER | grep WRITE
mobile-widgets/qmlmanager.h:    Q_PROPERTY(bool locationServiceEnabled MEMBER m_locationServiceEnabled WRITE setLocationServiceEnabled NOTIFY locationServiceEnabledChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(bool locationServiceAvailable MEMBER m_locationServiceAvailable WRITE setLocationServiceAvailable NOTIFY locationServiceAvailableChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(bool loadFromCloud MEMBER m_loadFromCloud WRITE setLoadFromCloud NOTIFY loadFromCloudChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(QString startPageText MEMBER m_startPageText WRITE setStartPageText NOTIFY startPageTextChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(bool verboseEnabled MEMBER m_verboseEnabled WRITE setVerboseEnabled NOTIFY verboseEnabledChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(QString notificationText MEMBER m_notificationText WRITE setNotificationText NOTIFY notificationTextChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(QString progressMessage MEMBER m_progressMessage WRITE setProgressMessage NOTIFY progressMessageChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(bool btEnabled MEMBER m_btEnabled WRITE setBtEnabled NOTIFY btEnabledChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(bool showNonDiveComputers MEMBER m_showNonDiveComputers WRITE setShowNonDiveComputers NOTIFY showNonDiveComputersChanged)
mobile-widgets/qmlmanager.h:    Q_PROPERTY(qPrefCloudStorage::cloud_status oldStatus MEMBER m_oldStatus WRITE setOldStatus NOTIFY oldStatusChanged)
mobile-widgets/themeinterface.h:        Q_PROPERTY(QString currentTheme MEMBER m_currentTheme WRITE set_currentTheme NOTIFY currentThemeChanged)
profile-widget/qmlprofile.h:    Q_PROPERTY(int diveId MEMBER m_diveId WRITE setDiveId)
```

profile-widget/qmlprofile.cpp
dirkhh: fixed

profile-widget/qmlprofile.cpp
dirkhh: actually, now that there are only two calls to `updateProfile()` left, there really isn't a point anymore. So I removed the whole reason argument.

bstoeger: I think you misunderstood my point:
Couldn't you simply write
```
        updateProfile("setDiveId");
```
I would presume that since `QString` has a conversion constructor from `const char *`, this should work. It could of course also lead to an error owing to ambiguity.

mobile-widgets/qml/DiveDetailsView.qml
dirkhh: there are so many small UI issues to be fixed.
not sure that would make the top 10 of my annoyances - but every bit of help is welcome :-)

profile-widget/qmlprofile.cpp
dirkhh: Cool beans. I'll happily try that and simplify the code then. Update coming in a few minutes, I want to get betas for iOS and Android out with the new profile... :-)

profile-widget/qmlprofile.cpp
dirkhh: I understood your point. I just decided that dropping that 'reason' argument was the more appropriate thing to do. Initially all of the setters called `updateProfile()` and the debug output got really confusing. That's why I added the `reason` argument in the first place.
But now that I don't do that anymore, it's not needed. So why clutter the code and waste critical µs...

dirkhh:
Yep, `MEMBER` does indeed work.
Also added a completely unrelated commit that fixes some styling of the export dialog (needed for Android)

bstoeger (approved):
I compiled and ran it on mobile-on-desktop. I can zoom on the map, but I could do this before. So approve with the caveat that I didn't test it on device.

dirkhh:
> I compiled and ran it on mobile-on-desktop. I can zoom on the map, but I could do this before. So approve with the caveat that I didn't test it on device.

- this is of course about the dive profile, not the map
- on a Mac with a touchpad you can do the pinch zoom even on desktop -- no idea about Linux
- drag/pan should work on any desktop platform
- I have now tested this on four different Android and iOS devices, from a brand new iPad Pro to an old Nexus 5 and it seems to work on all of them

Also, both the iOS and Android betas that were pushed yesterday include this code, and so far we haven't seen any major complaints :-)

PR #2629, opened 2020-02-19
[TRIVIAL] mobile: remove DiveListSortModel::getFilterString()

bstoeger:
This member function was never used.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Trivial(?) code cleanup.

### DIFF
--- a/qt-models/divelistmodel.cpp
+++ b/qt-models/divelistmodel.cpp
@@ -191,11 +191,6 @@ DiveListSortModel *DiveListSortModel::instance()
 	return &self;
 }
 
-QString DiveListSortModel::getFilterString() const
-{
-	return filterString;
-}
-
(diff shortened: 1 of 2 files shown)


dirkhh:
Given that you removed that model in our other branch, is it worth messing with it? This just makes merging harder...

dirkhh (commented):
So this is obviously correct, but I wonder if we really want to add this noise if we are removing this model, anyway.

bstoeger:
Yes - for itself it does more harm than good. However, I may need this for #2625, so let's move discussion there.

bstoeger:
Included in #2625 - individual PR is pointless.

PR #2630, opened 2020-02-19
undo: don't add dive to null-trip

bstoeger:
In moveDiveToTrip(), the dive was first removed from its old trip
and then added to the new trip. This function is also used to
remove the dive from its trip (by moving it to the "null-trip"
if you whish). Even in that case add_dive_to_trip() was called.
The only reason why this didn't crash is that add_dive_to_trip()
checks whether old and new trip are the same. This is the case
when adding to the "null-trip", since the dive was removed from
the trip just before.

To cut a long story short, to trust on add_dive_to_trip() not
crashing if moving from the null-trip to the null-trip is
way to subtly. If we remove a dive from its trip, don't call
add_dive_to_trip() in the first place.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a code cleanup that removes a horrible subtlety in the way the remove-dive-from-trip code is handled. See commit message.
### Mentions:

@dirkhh

### DIFF
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -271,7 +271,8 @@ static OwningTripPtr moveDiveToTrip(DiveToTrip &diveToTrip)
 
 	// Store old trip and get new trip we should associate this dive with
 	std::swap(trip, diveToTrip.trip);
-	add_dive_to_trip(diveToTrip.dive, trip);
+	if (trip)
+		add_dive_to_trip(diveToTrip.dive, trip);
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
This does seem fairly obvious;
The undo code overall is very complex and I need to spend more time reading it to make sure that I finally really understand it.

PR #2632, opened 2020-02-24
build-system: use working version of Breeze icons

dirkhh:
As of today some of the icons that we need are no longer in the git repoo but
instead created via script. Instead of making this work everywhere, let's just
use a working SHA...

- [X] Bug fix
- [X] Build system change

### Mentions:

@bstoeger


dirkhh:
As expected, this works, so merging my own PR :-P

PR #2634, opened 2020-02-24
Mobile profile tuning

dirkhh:

- [x] Functional change

Two small changes that push things a bit for the profile
- panning with basically no wait; we had a number of beta testers confused by the 150ms hold time before they were able to drag. This is near-instanteneous.
- 25% smaller font in the profile. That's getting quite small by default, but looks a lot less cluttered. And of course now we can zoom in...


dirkhh:
No comments - but then, no one else is working on the mobile app anymore, so I guess I'll just push ahead.

PR #2635, opened 2020-02-24
Erase deco obligation when NDL is positive again

### LINKED ISSUE #1351: Drawing of the computer-reported ceiling doesn't "end" when deco ceiling is cleared
### Describe the issue:
- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Drawing (on chart) of the computer-reported ceiling doesn't seem to end, at least for Ratio iX3M Tech+, after deco/ceiling has cleared.
dive2
dive2_r

Attached are screen shots of Subsurface compared to Ratio's (horrible) DiveLogger software.

### Operating system:
Windows 10 Pro (English) x64 Build 17134
Desktop - Intel Core i7-2600K 3.4Ghz & 32GB
Haven't tried on other devices/OSes yet, but I can/will if requested.

### Subsurface version:
4.7.8 (official release from subsurface-divelog.org)
Haven't tried on other version.

### Steps to reproduce:
Imported dive log (from computer) and toggled computer-reported ceiling

### Current behavior:
Computer-reported ceiling/deco doesn't show as "cleared" on graph, even after computer showed clear (and Ratio DiveLogger shows clear.)

### Expected behavior:
Expect the computer-reported deco to match DiveLogger.

### Additional information:
Included relevant dive from log.
[gextyr.zip](https://github.com/Subsurface-divelog/subsurface/files/2066430/gextyr.zip)

Note - given the data, it seems that (due to ocean swell and current) I didn't hold at exactly 20ft/6m, which causes numerous recalculations of stop time.  However, the computer and DiveLogger both behave as expected.

atdotde:
Logs from Ratio dive computers appeared to never clear
dc reported deco. According to Jef, the end of deco
is communicated by libdivecomputer as a non-zero
NDL. So this patch should implement this.

I don't have access to any Ratio dive computer, so
I cannot test, this actually works. Neither do I know
if some other divecomputer might confusingly report
both a positive NDL and a deco obligation (but it shouldn't)

Fixes #1351

One could argue that the problem is not in Subsurface but in libdivecomputer but let's not.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -334,8 +334,14 @@ sample_cb(dc_sample_type_t type, dc_sample_value_t value, void *userdata)
 		// values will be overwritten by new data if available
 		sample->in_deco = in_deco;
 		sample->ndl.seconds = ndl;
-		sample->stoptime.seconds = stoptime;
-		sample->stopdepth.mm = stopdepth;
+		// A positive NDL erases any previous deco obligations
+		if (ndl > 0) {
+			sample->stoptime.seconds = 0;
+			sample->stopdepth.mm = 0;
+		} else {
+			sample->stoptime.seconds = stoptime;
+			sample->stopdepth.mm = stopdepth;
+		}
 		sample->setpoint.mbar = po2;
 		sample->cns = cns;
 		sample->heartbeat = heartbeat;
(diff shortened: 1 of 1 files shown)


dirkhh:
Based on the conversation on the mailing list this is only part of the solution. I think we should wait until we understand what @jefdriesen implements in libdivecomputer before moving forward here.

dirkhh:
Is this still the right thing to do on top of the changes in libdivecomputer?
I haven't paid enough attention to those discussions as this is squarely the overlap of Linus & Robert in my mind... one less thing to worry about :-)

jefdriesen:
No, this shouldn't be necessary anymore. I tested the libdivecomputer fix with subsurface, and it produces the expected result now.

PR #2637, opened 2020-02-25
documentation: additional explanations around Windows USB drivers

dirkhh:
This is adapted from an email by Jef.

Suggested-by: Jef Driesen <jef@libdivecomputer.org>

- [x] Documentation change

### Documentation change:

This *is* a documentation change :-)

### Mentions:

@willemferguson 
@jefdriesen

### DIFF
--- a/Documentation/user-manual.txt
+++ b/Documentation/user-manual.txt
@@ -4406,7 +4406,18 @@ computer prefers (e.g. Bluetooth, USB, infra-red).
 
 	* On Windows, the OS should offer to download the correct
 	  driver once the user connects the dive computer to the USB port and
-	  operating system sees the equipment for the first time.
+	  operating system sees the equipment for the first time. In some instances this
+	  seems to fail, but regardless, every USB device (with the exception of HID devices
+	  like the Suunto EON Steel family) needs a driver. For many divecomputer this means
+	  downloading a cable specific driver (these tend to be for example FTDI, Silicon Labs,
+	  or Prolific drivers). In other cases (for example for dive computers with a "real" USB
+	  interface like the Atomics Cobalt) it is sufficient to have the WinUSB (or the libusbK)
+	  driver installed. In current Windows versions the WinUSB binary is already part of Windows
+	  and the "driver" is a simple inf file that tell Windows to use the WinUSB driver. But it
+	  still needs to be "installed". Often it is sufficient to install the dive log app that is
+	  provided by the dive computer vendor to trigger the driver installation. But in cases
+	  where that doesn't work, some googling to find the correct driver for a specific download
+	  cable may be necessary.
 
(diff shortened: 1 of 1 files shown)


jefdriesen:
Just a small remark. My explanation about WinUSB is only for dive computers with a real USB interface (like the Atomic Cobalt). For dive computers using a usb-serial interface you still need the corresponding driver from FTDI, Prolific, Silabs, etc and not the WinUSB driver. If you install the WinUSB driver for an usb-serial device, it will not work (except maybe ftdi devices with libftdi, because that uses libusb under the hood).

dirkhh:
> Just a small remark. My explanation about WinUSB is only for dive computers with a real USB interface (like the Atomic Cobalt). For dive computers using a usb-serial interface you still need the corresponding driver from FTDI, Prolific, Silabs, etc and not the WinUSB driver. If you install the WinUSB driver for an usb-serial device, it will not work (except maybe ftdi devices with libftdi, because that uses libusb under the hood).

Yeah, I thought that I was making this sound to easy. We do build the Windows binary with libftdi, but this could be misread that no Prolific of SiLabs drivers are needed, either. I'll give it another try.
Thanks for the feedback, that's really helpful!

dirkhh:
OK, second attempt pushed.

jefdriesen:
Has anyone ever tried libftdi on Windows? Installing the WinUSB driver means the original FTDI usb-serial driver gets replaced. The result is that the device will no longer show up as a serial port. The normal serial transport will stop working, and the libftdi based transport is now the only option. Also the official application of the dive computer manufacturer will no longer work. So in general it's probably a very bad advice to try to install the WinUSB driver for usb-serial dive computers. Especially when we're not sure the libftdi actually works.

The new text "In many cases..."  and the fact that it's listed first, makes it look like installing the WinUSB driver is always the best option. But there is actually only one dive computer, the Atomic Cobalt, where it is really needed. For all others, installing the correct usb-serial driver is the recommended option. And the easiest way to do that is by installing the application of the dive computer manufacturer.

Even if libftdi works fine on Windows, I recommend to stick to the standard ftdi usb-serial driver. It works very well, and doesn't break other applications. On other platforms, there is no such problem, because libusb can access USB devices without having to install another driver. On Windows usb-serial and libusb/libftdi are mutually exclusive :-(

This is also the reason why libdivecomputer prefers using hidapi over libusb for USB HID devices on Windows. Hidapi can use the standard Microsoft USB HID driver and works out of the box, while libusb requires installing the WinUSB drivers, which breaks all other applications.

dirkhh:
I'm happy to follow your recommendation here and rephrase again.
Yes, I've tried libftdi on Windows and it seemed to work fine - but those were rather limited tests with the one and only semi-working FTDI dive computer that I have.
I do remember the conversation about hidapi and that's what we are using for our Windows build

PR #2638, opened 2020-02-27
Minor cleanups concerning the equipment tab

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

These are just minor cleanups as a consequence of the undo work. They don't (shouldn't) change the user interaction. Note that the "are you sure to revert" message is not properly shown if the profile is edited. However, that problem also exists without this PR.
1) Remove an unused member variable.
2) Remove pointless checking for changed weightsystems.
3) Avoid double update of tabs on reject-changes.


dirkhh (approved):
looks good

PR #2639, opened 2020-02-27
Fix crash in planner

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a crash in the planner introduced with the "CylindersModelFiltered": Since the planner doesn't use this model, the delegate must not cast to this model. To solve this dilemma, use Qt's model-view mechanism instead to store/load data. It's a bit hairy, so testing would be welcome.

To reproduce the crash simply go to the planner and select a different cylinder type.

### Changes made:

1) Access data via Qt's model-view system instead of custom functions.
2) Don't downcast the model.

### Release note:

Not in release - no.

### Mentions:

@atdotde @dirkhh

### DIFF
--- a/desktop-widgets/modeldelegates.cpp
+++ b/desktop-widgets/modeldelegates.cpp
@@ -238,7 +238,7 @@ static struct RevertCylinderData {
 
 void TankInfoDelegate::setModelData(QWidget*, QAbstractItemModel*, const QModelIndex&) const
 {
-	CylindersModelFiltered *mymodel = qobject_cast<CylindersModelFiltered *>(currCombo.model);
+	QAbstractItemModel *mymodel = currCombo.model;
 	TankInfoModel *tanks = TankInfoModel::instance();
 	QModelIndexList matches = tanks->match(tanks->index(0, 0), Qt::DisplayRole, currCombo.activeText);
 	int row;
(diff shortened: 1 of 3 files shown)


dirkhh (commented):
Heck if I know. Seriously - no idea.

bstoeger:
> Heck if I know. Seriously - no idea.

? Is this comment in the correct PR ?

dirkhh:
> > Heck if I know. Seriously - no idea.
> 
> ? Is this comment in the correct PR ?

Yes. I do not fully understand what you are doing here and why what was there before would cause a crash.

dirkhh:
OK, let me phrase this differently.
I believe your code to be correct. I cannot point at something and say "this seems wrong".
But I don't understand why this crashed, nor do I have enough understanding of how all this is supposed to work to tell you if this is the right approach.
Which lead me to the somewhat flippant "heck if I know" comment - which maybe wasn't very professional nor was it especially supportive of you as the developer. Apologies for that.

bstoeger:
This:
``` qobject_cast<CylindersModelFiltered *>(currCombo.model);```
returns NULL if the model is not of `CylindersModelFiltered` type. Since in the planner we use the non-filtered `CylindersModel`, it crashes. We can't change the cast to the latter, because then the EquipmentTab would crash.

We could derive `CylindersModelFiltered` and `CylindersModel` from a common base class that has virtual `passInData` and `cylinderAt` functions, which are reimplemented in the leaf classes.

We could define different delegates for the planner and the equipment tab.

Or we could, if the first cast fails, do the other cast.

Or, as in this PR, we could squeeze the data through Qt's model/view system with special roles and columns as realized in this PR.

Hope that clears it up.

> Which lead me to the somewhat flippant "heck if I know" comment - which maybe wasn't very professional nor was it especially supportive of you as the developer. Apologies for that.

\*shrug*
I didn't feel belittled. I was just confused as I read it as an answer and I wasn't sure to which question.

bstoeger:
Added a leak fix and a (very minor) code prettification. That said, beauty is in the eye of the beholder, obviously.

dirkhh (approved):
This looks fine - thanks for the explanation

PR #2641, opened 2020-02-27
Filter26

dirkhh:
This is simply re-casting #2625 on top of current master to see if this maybe can fix the odd GitHub Action issue with the Android build - which fails at the git checkout phase over there.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a primitive first implementation of a fulltext index. There are two reasons for this:
1) The filterwidget on desktop is getting horrendously complex and I wonder if many usecases couldn't be replaced by a simple fulltext search. My suggestion would be to show only the fulltext search first and then let the user expand "advanced options".
2) IIRC we do fulltext search on mobile in a background thread, which appears to be very brittle. Therefore, I suggest using the same fulltext code on mobile and desktop.

Be warned that this is at concept stage, though it worked in my preliminary tests. It is 100% unoptimized and there is ample room for optimization. However, when testing on a ~4000 dives testlog, the actual fulltext search was always faster than 10 ms. If the search-words reach three letters or more, it typically got even <1 ms. I wonder if optimizing the actual search is even worth it.

There is one big downside: currently, loading the testlog mentioned above is slowed by ~750 ms, because the index is generated on loading. A first attempt at generating the index in a background thread failed: The profile widget clears the `displayed_dive`, which accesses the cache and thus waits for the index generation. Moreover, the background thread would somehow have to lock on the dive-list, which we don't even support. Ugh.

If someone suggests an external fulltext-library that is easy to hook into, I'm fine with that too!

### Release note:

Let's stabilize this first.
### Documentation change:

Let's stabilize this first.
### Mentions:

@dirkhh @willemferguson @atdotde

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 Mobile: enable profile zoom and pan
+Mobile: add people- and tags-filter modes
+Desktop: fix tab-order in filter widget
+Desktop: implement fulltext search
 Mobile: fix saving newly applied GPS fixes to storage
 Desktop: add starts-with and exact filter modes for textual search
 Mobile: add option to only show one column in portrait mode
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -88,6 +88,8 @@ set(SUBSURFACE_CORE_LIB_SRCS
 	file.h
 	format.cpp
 	format.h
+	fulltext.cpp
+	fulltext.h
 	gas.c
 	gas.h
 	gas-model.c
--- a/core/dive.c
+++ b/core/dive.c
@@ -17,6 +17,10 @@
 #include "tag.h"
 #include "trip.h"
 #include "structured_list.h"
+#ifndef SUBSURFACE_MOBILE
+#include "fulltext.h"
+#endif
+
 
 /* one could argue about the best place to have this variable -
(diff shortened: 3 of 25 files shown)


dirkhh:
Surprise, surprise - no build issues.

bstoeger:
> Surprise, surprise - no build issues.

Well to me it *is* surprising... Anyhow, that raises two questions:
1) What do we learn from this?
2) How do we proceed?

😛

dirkhh:
> > Surprise, surprise - no build issues.
> 
> Well to me it _is_ surprising... Anyhow, that raises two questions:
> 
> 1. What do we learn from this?

Somehow GitHub got stuck in a place where the Android build wanted to check out a reference that wasn't a tree. It seemed clear to me that we needed to change the commit SHAs in order to bump this along - rebasing on the latest master seemed like the easiest solution.

> 2. How do we proceed?

I think I'm ready to merge this, unless you have concerns.

Fundamentally I want to push to get our outside branch merged so that we can work towards new releases. It's been way too long since we made releases.

bstoeger:
> I think I'm ready to merge this, unless you have concerns.

Fine with me.

PR #2642, opened 2020-02-28
Desktop: fix scrolling through the dive list with cursor keys

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes scrolling through the dive list with cursor keys. I'm really at a loss here - I don't see how to fix this mess without hooking into all events that can change the selection.

As a recapitulation: The whole problem is that when selecting all items in the tree-view, we get a selection event for every trip. If we have a few hundred trips, the map is repainted a few hundred times, thus freezing the UI for many seconds. Sadly, we don't know when the last selection event is sent.

### Mentions:

@dirkhh @willemferguson

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -554,6 +554,19 @@ void DiveListView::mouseReleaseEvent(QMouseEvent *event)
 		selectionChangeDone();
 }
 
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
horrible... that's a good description of the selection mess.

PR #2643, opened 2020-03-01
First steps of cylinder-editing undo

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are the first steps at implementing undo of cylinder editing. The goal is to enable undo only for the equipment tab and leave the planner as-is. The second goal was to touch the code as little as possible to avoid messing things up, which I don't understand. The result is ... not exactly pretty.

What is still missing is manipulation of cylinders via the profile. Currently, you probably can crash the app by adding a cylinder - adding a cylinder switch - undo addition of cylinder. I will look into that next.

### Changes made:

1) Implement cylinder-undo commands.
2) Remove some of the edit-mode cruft that is now no longer needed.

### Release note:

Will do once ready.
### Documentation change:

Not sure..?

### Mentions:

@atdotde: I will need your help here with testing, especially with respect to interactions with the planner.


dirkhh:
There may be some issues remaining :-)
```
Undefined symbols for architecture x86_64:
"Command::editCylinder(int, cylinder_t, bool)", referenced from:
   CylindersModel::setData(QModelIndex const&, QVariant const&, int) in libsubsurface_models_mobile.a(cylindermodel.cpp.o)
```
I'll look through the changes.

bstoeger:
Yes, we must comment that out for mobile as long as we don't have the undo-commands on mobile. Later - since I'm just heading out for dinner...

core/planner.c
```diff
@@ -206,6 +206,16 @@ void fill_default_cylinder(const struct dive *dive, cylinder_t *cyl)
...
+cylinder_t create_new_cylinder(const struct dive *d)
+{
+	cylinder_t cyl = empty_cylinder;
+	fill_default_cylinder(d, &cyl);
+	cyl.start = cyl.type.workingpressure;
+	cyl.manually_added = true;
+	cyl.cylinder_use = OC_GAS;
+	return cyl;
+}
+
```
dirkhh: this really should be in `equipment.c`, shouldn't it?
Let's move `fill_default_cylinder()` to the right file as well, instead of having these functions handing out in the planner...

core/qthelper.cpp
```diff
@@ -1653,3 +1653,33 @@ char *copy_qstring(const QString &s)
 {
 	return strdup(qPrintable(s));
 }
+
+// Generate a cylinder-renumber map for use when the n-th cylinder
+// of a dive with count cylinders is removed. It fills an int vector
+// with 0..n, -1, index..count-1. This probably should be moved
```
dirkhh: you are using bother `index` and `n` here. That's a cut'n paste error, right?

core/qthelper.cpp
```diff
@@ -1653,3 +1653,33 @@ char *copy_qstring(const QString &s)
...
+// to the C-core, but using std::vector is simply more convenient.
+// This assumes that n < count!
+std::vector<int> get_cylinder_map_for_remove(int count, int n)
+{
+	// 1) Fill mapping[0]..mapping[index-1] with 0..index-1
+	// 2) Set mapping[index] to -1
+	// 3) Fill mapping[index+1]..mapping[count-1] with index..count-2
```
dirkhh: again the references to `index` even thought that's nowhere in the code
And... instead of explaining what the code does, it would be even better to explain WHY that's what you do here (I think I have reached the point in my `std::` learning journey that I can actually read and understand that code... yay me)

dirkhh (commented):
reviewed until `undo:reorder cylinders on remove...`
submitting so I can come back and continue the review later

qt-models/cylindermodel.cpp
```diff
@@ -350,100 +351,116 @@ bool CylindersModel::setData(const QModelIndex &index, const QVariant &value, in
 	QString vString = value.toString();
 	bool changed = vString != data(index, role).toString();
 
+	std::string newType; // If we allocate a new type string, this makes sure that it is freed at the end of the function
```
dirkhh, line 355: that seems a bit subtle (or maybe that's just me).
Why not `strdup()` it and free it at the end? I think that would be more easily readable for me.
But maybe I'm missing other benefits of doing it this way?

dirkhh (commented):
OK, gone through all of it.
One thing I wonder about with the whole issue of two cylinder models, etc.
What happens if the user is working in the planner, and then uses undo commands?
Should we just disable the undo stack while the user is in the planner? I think that would make sense...

I'm not approving this because I had a few questions (including this one here) - but I think this is very much going the right way.

Also, for the record, let me thank you for the way you structure these PRs. Methodical, fairly easy to follow... makes my life so much easier!

core/planner.c
bstoeger: > this really should be in equipment.c, shouldn't it?

Yes, it should. The last time I wanted to move `fill_default_cylinder` there, I didn't because it takes a `dive *` argument and I couldn't decide if this is a dive-related or a cylinder-related function, so I left the declaration in `dive.h`. I guess `equipment.h` is more logical after all.

core/qthelper.cpp
bstoeger: Yes, that's bogus.

bstoeger:
> Should we just disable the undo stack while the user is in the planner? 

Yes, definitely. Just as we disable the undo stack when the user is editing the profile.

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -210,21 +210,6 @@ void MainTab::enableEdition(EditMode newEditMode)
...
-
-
-
-		// FIXME : can we get rid of this recursive crap?
-
-
-
-		MainWindow::instance()->editCurrentDive();
-		return;
-	}
```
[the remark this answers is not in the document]
bstoeger: To my understanding, the previous commits made this dead code:
The only caller of `enableEdition()` is `MainWindow::editCurrentDive()` and that *always* calls `enableEdition()` with `mode == MANUALLY_ADDED_DIVE` for manually added dives. So we can never reach this if-branch (note that it would lead to an infinite recursion).

This was only necessary when `enableEdition()` was called from other parts of the code.

desktop-widgets/mainwindow.cpp
```diff
@@ -1792,21 +1792,19 @@ void MainWindow::editCurrentDive()
...
 		mainTab->enableEdition(MainTab::MANUALLY_ADDED_DIVE);
 	} else if (defaultDC == "planned dive") {
+		disableShortcuts();
 		DivePlannerPointsModel::instance()->setPlanMode(DivePlannerPointsModel::PLAN);
 		setApplicationState(ApplicationState::EditPlannedDive);
 		DivePlannerPointsModel::instance()->loadFromDive(d);
 		mainTab->enableEdition(MainTab::MANUALLY_ADDED_DIVE);
-	} else {
-		setApplicationState(ApplicationState::EditDive);
-		mainTab->enableEdition();
```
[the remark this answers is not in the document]
bstoeger: OK - good to know. We can drop this commit.

With my personal dives I can't select "Edit the Profile" in the context menu though?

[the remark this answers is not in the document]
bstoeger: Looking closer:
In bbfa2b655bac88306f8e7805f2eb61b2f801ad0a you implemented the context-menu, however only for manually added dives.

Perhaps you are confusing this with the "Edit dive in planner"  function, which also works for downloaded dives. Though that is different from the whole "editCurrentDive()" function.

Therefore currently, it is dead code that is removed here.

qt-models/cylindermodel.h
```diff
@@ -31,7 +31,9 @@ class CylindersModel : public CleanerTableModel {
 	};
 
 	enum Roles {
-		PASS_IN_ROLE = Qt::UserRole + 1 // For setting data: don't do any conversions
+		TEMP_ROLE = Qt::UserRole + 1, // Temporarily set data, but don't store in dive
+		COMMIT_ROLE, // Save the temporary data to the dive. Must be set with Column == TYPE.
+		REVERT_ROLE // Revert to original data from dive. Must be set with Column == TYPE.
```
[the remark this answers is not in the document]
bstoeger: Yes, that is super-horrible. Sadly, it was the quickest solution I could conjure that works despite the fact that the planner and the equipment tab use different models.

qt-models/cylindermodel.cpp
bstoeger: > Why not strdup() it and free it at the end?

That's functionally equivalent, however...

> But maybe I'm missing other benefits of doing it this way?

Yes - the crucial advantage is that it doesn't leak when someone adds an early return from the function. No need for `goto out` or additional `free()`, etc. The compiler is much better at such things.

desktop-widgets/mainwindow.cpp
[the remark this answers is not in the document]
dirkhh: You are correct.

core/planner.c
bstoeger: Moved both functions to equipment.c

core/qthelper.cpp
bstoeger: replaced "index" by "n".

core/qthelper.cpp
bstoeger: Replaced "index" by "n". Added a small comment to the function header what this does. Note that all this cylinder-remapping is not from me.

bstoeger:
> > Should we just disable the undo stack while the user is in the planner?
> 
> Yes, definitely. Just as we disable the undo stack when the user is editing the profile.

Added two new commits that connect the disabling of undo to the profile status. Thus, undo should now be disabled in the planner, though this could need some testing. Removed the EDIT status.

bstoeger:
I added a bunch of cleanup commits that will make my life easier when incorporating the event manipulations in the undo system. This will be necessary if we want cylinder undo. Example:
1) add cylinder
2) add gas-switch to new cylinder
3) undo cylinder addition -> boom we have an invalid gas switch

Yes, this adds a bunch of lambdas, but I hope it is clear that this simplifies code significantly!

bstoeger:
> Yes, this adds a bunch of lambdas, but I hope it is clear that this simplifies code significantly!

To be fair though, only a part of the simplification is due to lambdas. A big part is that the timestamp of an event is not recalculated anymore inside every action from the QPoint which is transported via the event.

bstoeger:
Small update: started with undo of event-editing. Adding bookmark, divemode and setpoint was surprisingly easy. Gas switch is the one that makes me feel very uneasy and which I will do last. I simply do not know what that can change and if it would perhaps be easiest to simply exchange two full dives...?

bstoeger:
Ok - tried adding GasChange-undo. I have _no idea_ what I'm doing here.
@janmulder, @atdotde: I fear I really need help with testing here, as all this functionality is way over my head.

Things that are known broken:
1) On event-deletion we have to do the same thing as on gas-change. After all, one can delete a gas-change event!
2) On undo/redo we should switch to the dive & divecomputer affected by the command. Otherwise the interface is very intransparent.

bstoeger:
Removing the [WIP] tag, as from my side, this is now feature complete.

Obviously, I could have missed something. This now would need testing, but I reckon there isn't too much interest. Not sure how to proceed - from a UI perspective, this removes the edit mode for cylinders that stuck out like a sore thumb.

dirkhh:
I will test over the weekend. Posting on the mailing list with a pointer to the binaries that this needs testing is also worth while.
We have a number of people who typically test things like this -- Robert, Miika (who is traveling), Jan, Linus (also traveling), Willem...

dirkhh:
Started testing. This is on a Mac, with this branch (so no merge ontop of anything else).
When I added a new tank and changed the O2 or He values for the gas mix, those entries weren't shown in the Cylinders widget. But when I tried to reproduce that and added a third tank, all the values where shown again. I tried a few things to see if I could reproduce it, but no such luck.
I think I'll mark this up as "oddity" unless I run into it again...

dirkhh:
Ahh, here is something that I **can** reproduce.
Edit a cylinder and try to hit CMD-Z (this is a Mac) for undo and nothing happens. Menu works. Giving focus to a different widget works. But editing something, hitting tab, and then hitting CMD-Z to undo the previous edit doesn't work.
Not a show-stopper, but worth noting

dirkhh:
Another thing (may not have been introduced by your changes). If I change the gas mix the Bot.MOD isn't always updated.  I haven't quite reverse engineered what triggers the update of the widget. When doing undo/redo the change is immediate - it's entering the data in the widget where it isn't.

dirkhh:
Ugh, another "I can't reproduce this" report.
I have a dive with one tank. Add a second tank. Add a gas switch to it. Add another tank.
Now I remove that gas switch. I had it also remove that second tank when I removed the event. And add it back when I did an undo. I don't think the event removal should remove the tank.
And when I try to reproduce things, it doesn't :-(

bstoeger:
> Ahh, here is something that I can reproduce.
Edit a cylinder and try to hit CMD-Z (this is a Mac) for undo and nothing happens. Menu works. Giving focus to a different widget works. But editing something, hitting tab, and then hitting CMD-Z to undo the previous edit doesn't work.
Not a show-stopper, but worth noting

That's a known issue with the undo of the whole application, not only the cylinders: When you hit Tab, the next text field in the table gets the focus. Henceforth CTRL-Z is captured by the text field and not forwarded to the app, even if there is nothing to undo. To me this appears to be a general Qt problem?

I will look at you other comments later today, thanks for testing.

bstoeger:
For the records, one problem pertains to multi-cylinder edits:
In my freshly imported dives I wanted to change all cylinder types. However, only a bunch of them got changed. The reason is that we only change those cylinders that we consider the "same". Since my computer exports the %O₂ settings, only the dives with the same mix as the displayed dive were updated. This is of course stupid.

Nevertheless, I would like to fix this later and keep it as a known defect for now. We need to properly define what we consider the "same" cylinder. E.g. when changing the oxygen fraction, should we consider the dives with the same oxygen fraction or with the same oxygen *and* helium fraction...?

mturkia:
On Sat, Mar 7, 2020 at 9:35 PM bstoeger <notifications@github.com> wrote:

> For the records, one problem pertains to multi-cylinder edits:
> In my freshly imported dives I wanted to change all cylinder types.
> However, only a bunch of them got changed. The reason is that we only
> change those cylinders that we consider the "same". Since my computer
> exports the %O₂ settings, only the dives with the same mix as the displayed
> dive were updated. This is of course stupid.
>

I think I noticed the same today. With Subsurface
v4.9.3-1034-g97e26fd51b5e, only one cylinder gets new info when trying to
edit multiple dives. All the O2 info should be the same on these dives.

> Nevertheless, I would like to fix this later and keep it as a known defect
> for now. We need to properly define what we consider the "same" cylinder.
> E.g. when changing the oxygen fraction, should we consider the dives with
> the same oxygen fraction or with the same oxygen *and* helium fraction...?
>
IMO when editing type of a cylinder, all types of the selected dives'
cylinders should be updated. It does not matter what the O2 or pressures
are.

miika

dirkhh:
Just to make sure my expectations are reasonable...
Is anyone planning to test this? 
It seems silly to let it sit here if no one is planning to do anything with it.

bstoeger:
> IMO when editing type of a cylinder, all types of the selected dives'
cylinders should be updated. It does not matter what the O2 or pressures
are.

Yes, that makes things a bit more complicated, because we have to treat type and gas-mix changes with different undo-commands.

For consistency, when changing the type, I would still only edit those cylinders that have the same type as the shown one. This is what we do with the rest of the editing commands. Moreover, we should show an "editing XXX dives dialog" - I wonder why this isn't happening.

dirkhh:
My preference would be that we don't try to mix behavioral changes with this.
Yes, the logic when and how we change things right now may not be what you find intuitive.
But it's what is documented and what people are used to... and changing that (and coming up with a new policy) should be done after a discussion on the mailing list (which of course assumes that anyone cares enough to comment - but usually at least Linus and Miika respond to things like this... as does Robert)

bstoeger:
The problem is that this PR *does* currently introduce behavioral changes, at least that's what I get from @mturkia's message. If it doesn't - all the better, because that means less work for now.

dirkhh:
__*<cough. cough>*__
if anyone cares what I think, I have a 105% preference that you not spend time on **this** PR right now, but instead that we finish _that other branch_...
__*<cough. cough>*__

dirkhh:
GitHub is acting up, the website keeps timing out on me...
Now that the other stuff is merged, let's return here.
@atdotde did you get a chance to test this branch? I think this is the next thing we should merge...

atdotde:
Not yet, but will do.

atdotde:
OK, here is something, which does not look right:
Start with an empty log. Open the planner, change the default dive to 40m, make the first cylinder a D12 with air, add a second cylinder, say AL80 with EAN50. Then save the planned dive to the dive list.
Go to the equipment tab, edit the "AL80" string to something like "AL80lala", press TAB to move focus, press Ctrl-Z to undo.

Nothing happens. Then click on the "D12" field to move the focus and without editing click somewhere else again to move the focus once more: "D12" becomes "AL80lala".

dirkhh:
> Go to the equipment tab, edit the "AL80" string to something like "AL80lala", press TAB to move focus, press Ctrl-Z to undo.

That doesn't work as discussed earlier. Ctrl-Z is text widget specific while you are inside one of the text widgets. So you need to click somewhere else (or use the menu system) to test undo in that scenario

> Nothing happens. Then click on the "D12" field to move the focus and without editing click somewhere else again to move the focus once more: "D12" becomes "AL80lala".

That, on the other hand, sounds like a bug. I'll need to build this branch to try and reproduce this - @bstoeger , are you looking into this as well?

bstoeger:
Currently on the move. There's one known bug which might be responsible for the confusion: the tabs are not properly cleared when the current dive is set to null (e.g. add dive followed by undo). That used to work, sigh.

bstoeger:
> Then click on the "D12" field to move the focus and without editing click somewhere else again to move the focus once more: "D12" becomes "AL80lala".

I can reproduce this now, it seems like the "tank delegate" is not initialized correctly.

atdotde (commented):
This is what I found in a first round of reading through the code. Now, I also need to understand it.

core/equipment.c
```diff
@@ -377,6 +392,46 @@ cylinder_t *get_or_create_cylinder(struct dive *d, int idx)
 	return &d->cylinders.cylinders[idx];
 }
 
+/* if a default cylinder is set, use that */
+void fill_default_cylinder(const struct dive *dive, cylinder_t *cyl)
+{
+	const char *cyl_name = prefs.default_cylinder;
+	struct tank_info_t *ti = tank_info;
+	pressure_t pO2 = {.mbar = 1600};
```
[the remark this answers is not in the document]
bstoeger: > I thought, pO2max is a preference.

Note that this is not my code, it is only moved. I can add a patch that uses `prefs.modpO2` though. I just wonder why we use floating point here, when usually we use fixed point.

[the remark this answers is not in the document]
bstoeger: Done - please check.

commands/command_event.cpp
```diff
@@ -0,0 +1,210 @@
...
+AddEventBookmark::AddEventBookmark(struct dive *d, int dcNr, int seconds) :
+	AddEventBase(d, dcNr, create_event(seconds, SAMPLE_EVENT_BOOKMARK, 0, 0, "bookmark"))
+{
+	setText(tr("Add bookmark"));
+}
+
+AddEventDivemodeSwitch::AddEventDivemodeSwitch(struct dive *d, int dcNr, int seconds, int divemode) :
+	AddEventBase(d, dcNr, create_event(seconds, SAMPLE_EVENT_BOOKMARK, 0, divemode, QT_TRANSLATE_NOOP("gettextFromC", "modechange")))
+{
+	setText(tr("Add dive mode switch to %1").arg(gettextFromC::tr(divemode_text_ui[OC])));
```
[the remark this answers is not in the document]
bstoeger: Doh, of course.

commands/command_edit.cpp
```diff
@@ -1155,4 +1158,190 @@ void EditWeight::undo()
...
+{
+	// Get the old cylinder, bail if index is invalid
+	if (!current || index < 0 || index >= current->cylinders.nr) {
+		dives.clear();
+		return;
+	}
+	cyl = clone_cylinder(current->cylinders.cylinders[index]);
+
+	// Deleting a cylinder from multiple dives is semantically ill-defined.
+	// What we will do is trying to delete the same weightsystem if it exists.
```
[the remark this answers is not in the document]
bstoeger: Yes, of course.

bstoeger:
> Then click on the "D12" field to move the focus and without editing click somewhere else again to move the focus once more: "D12" becomes "AL80lala".

That should be fixed with the latest commit. That said, the whole ComboBoxDelegate code makes my head spin.

dirkhh:
So this one is ready to merge, correct?
Do I merge this before or after we do the Subsurface-mobile 3.0 release...

bstoeger:
I attempted to fix the issues that @atdotde found. No sure if that makes it "ready to merge".

bstoeger:
To me there are two issues:
- No multi-dive-edit is shown if cylinders are edited (same problem for weights).
- Very non-intuitive behavior that when changing e.g. type only dives with the same gasmix are changed.

bstoeger:
@mturkia, @atdotde, @dirkhh: could we please reconsider this? I tried to make the editing more fine-grained to avoid surprising effects and now a multi-dive warning is shown. Unfortunately, this means that more testing is necessary. Sadly, the code is very complex because the same model is used for planner (without undo) and non-planner (with undo).

dirkhh:
I would really like to see this go in.
Multi dive editing of cylinders has never been super consistent and logical.
And I know that if people will find objectively incorrect behavior that you will track it down and fix it.
I haven't had time to play with the latest iteration, and I haven't used the planner in several years, so I may not be the best judge as to whether this is bug free - but in my limited testing last week it did what I expected it to do.

mturkia:
I am seeing something weird, when testing this branch. Somewhat
unpredictable behaviour. My log had maybe 15 dives with no cylinder type
set. And here are a few observations while changing the type.
- Setting type to AL80 for the oldest of the unknown cylinders => all the
unknown cylinders got the new type of AL80, even though only one was
selected
  - When saving the changes, the only difference committed to cloud storage
is switching the order of two dive computers. No info is updated to any of
the cylinders, not the selected dive, nor the other dives that had the
change done in the GUI (unintentionally)
  - It might be that some of the cylinder information is not updated on
screen, when switching dives (with arrow keys)
- When hitting undo, no description of any of these cylinders was changed
back to unknown
  - Edit menu had option to redo the undo
- After reloading my divelog, setting the type of cylinder to one dive and
a few selected dives seemed to work, however, setting the cylinder to last
~10 selected dives resulted in one of the dives not getting the cylinder
type set
- It seems that when changing the cylinder type, I am unable to store the
change into git at all.

No idea what causes these random irregularities, otherwise the undo-redo
seems to be working properly.

BTW have you noticed issue
https://github.com/Subsurface-divelog/subsurface/issues/2676 ? Copy-paste
of cylinder information is broken, IMO.

bstoeger:
A small improvement: no more inconsistent state. However another problem appeared: You can't set an arbitrary name while the combobox is active, i.e. a different name that starts with the same characters exists. The delegate code is very frustrating.

bstoeger:
I hope that all but one of the issues reported by @mturkia are fixed now. Who knows what new breakage there is though - the delegate code is very convoluted.

The only one, which is not fixed is:
> setting the cylinder to last ~10 selected dives resulted in one of the dives not getting the cylinder
type set

This is probably caused by this cylinder having had a different type then the others. Would be good to double check that.

bstoeger:
PS: The commit-history is horrible, but I don't think reworking it is worth the hassle.

dirkhh:
are we waiting for more testing from @mturkia and others, or would you prefer that we move ahead, merge as is and deal with additional fall-out later?

I just noticed that I get crashes in the cylinder code in export-html on the new cloudserver... turns out it was something that was recently fixed in master, but not in the version that happened to be used on the server :-/

bstoeger:
> are we waiting for more testing from @mturkia and others, or would you prefer that we move ahead, merge as is and deal with additional fall-out later?

Would be nice if we get confirmation from @mturkia.

> I just noticed that I get crashes in the cylinder code in export-html on the new cloudserver... turns out it was something that was recently fixed in master, but not in the version that happened to be used on the server :-/

I reckon this means I don't have to act? Perhaps add this failure mode to the tests?

dirkhh:
> > I just noticed that I get crashes in the cylinder code in export-html on the new cloudserver... turns out it was something that was recently fixed in master, but not in the version that happened to be used on the server :-/
> 
> I reckon this means I don't have to act? Perhaps add this failure mode to the tests?

No action necessary.
When I started setting up the new cloud storage server I used whatever was latest master that day, which happened to be just before 62adc24d15 went in... which caused us to dereference a non-existing cylinder. The author of said commit in the commit message indicated that changes to that code would be forthcoming :-)

dirkhh:
I think I'll move forward and merge this. I want us to get a 4.9.4 out and this is one of the biggest pending things to go in. Once this is merged I'll ask for more testing on the mailing list...

PR #2644, opened 2020-03-03
desktop: don't close application if pressing CTRL-Q in setpoint dialog

bstoeger:
The CTRL-Q signal in the setpoint dialog was connected to the close
slot of the main-window instead of the dialog. Thus, pressing CTRL-Q
would close the application, not the dialog.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a funny bug: pressing CTRL-Q closes the application not the setpoint dialog. This commit is implicitly contained in a commit of #2643. I'm not sure if we want the extra commit here for documentation purposes. @dirkhh: up to you. Feel free to trash this PR if it appears too trivial.

### Changes made:

1) connect CTRL-Q to the appropriate receiver.
### Release note:

No - too trivial / obscure.

### DIFF
--- a/desktop-widgets/simplewidgets.cpp
+++ b/desktop-widgets/simplewidgets.cpp
@@ -205,7 +205,7 @@ SetpointDialog::SetpointDialog(QWidget *parent) : QDialog(parent),
 	QShortcut *close = new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_W), this);
 	connect(close, SIGNAL(activated()), this, SLOT(close()));
 	QShortcut *quit = new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_Q), this);
-	connect(quit, SIGNAL(activated()), parent, SLOT(close()));
+	connect(quit, SIGNAL(activated()), this, SLOT(close()));
 }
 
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
he, yeah, that looks better

bstoeger:
Ooops I take that back - this was the intended behavior:
1306c0fecbd958656ba312be7e8ba1428f876183

Sorry for noise.

dirkhh:
this may be what our GSOC student intended 6 years ago and I indeed signed off on that.
I'm not sure this is what we should want... or maybe it is.
Today is not a good day for me to make judgment calls like that.

bstoeger:
If in doubt, let's leave it like it is.

I removed the `connects` and what happens is that the application doesn't react to CTRL-Q in modal dialogs. I'm not going to judge if that is right or wrong. Anyhow, if we decide that CTRL-Q should also work from modal dialogs, I would suggest creating our own Dialog base-class to avoid all that duplicate code. Very low priority, of course.

PR #2645, opened 2020-03-04
[TRIVIAL] cleanup: make version_printed variable static and local to function

bstoeger:
The version_printed variable is used to print version information
only once. It was a global variable, but never used outside of
its function. Therefore, move it into the function and make it
static. Since this is a plain old datatype (POD), it makes no
no difference whatsoever whether the static variable is in block
scope or not. Indeed, it is initialized in the data segment). Well,
we are in C mode and therefore everything has to be POD by definition.
I tested this on gcc and clang.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a completely trivial idle code-cleanup. It could shorten the symbol-table by a byte or two.

### Mentions:

@dirkhh

### DIFF
--- a/core/subsurfacestartup.c
+++ b/core/subsurfacestartup.c
@@ -128,9 +128,9 @@ const char *monthname(int mon)
  */
 bool imported = false;
 
-bool version_printed = false;
 void print_version()
 {
+	static bool version_printed = false;
 	if (version_printed)
 		return;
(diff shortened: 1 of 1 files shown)


bstoeger:
Too trivial to lose time over it. Since CI was successful - commit.

PR #2646, opened 2020-03-05
Cherry-picks from new mobile list branch

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

@dirkhh: these are the first of a few of the more trivial commits, with slightly reworded commit-messages. If we get these into master, I hope that cleaning up the history will be easier.

### DIFF
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -24,7 +24,7 @@ void deleteDive(const QVector<struct dive*> &divesToDelete)
 	execute(new DeleteDive(divesToDelete));
 }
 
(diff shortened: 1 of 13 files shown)


dirkhh (approved):
That last commit is utter garbage from an "it actually does what it claims to do" perspective - but it's useful when testing mobile on the desktop...

But I'm ok with merging all of this.

bstoeger:
Good - thanks. I will merge and rebase our branch so that we can slowly can get rid of commits.

PR #2649, opened 2020-03-07
More cherry-pickery

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] preparation for mobile models

Some more cherry-pickery to reduce the size of the mobile models branch. One of the commits is a completely unrelated keyhole optimization, the other necessary for getting the filter-counts right.


dirkhh (approved):
I'm of course happy with those two commits going in separately.
I do question the value of taking the effort of extracting these vs. just merging the 90 commits in our branch.
But if it really is the number of commits that you worry about, then maybe we should remove all the context menu, trip edit, etc stuff from that branch...

PR #2656, opened 2020-03-08
[testing needed] usb-serial-for-android: Implement timeout-handling

charno:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Since the Android USB stack and subsequently the usb-serial-for-android
driver have problems with read-timeouts, the read-timeout is now
implemented in AndroidSerial.java. Also, DC_STATUS_TIMEOUT is returned
if there are less bytes returned than expected.

Different chipsets seem to behave differently with
usb-serial-for-android. On CP210x the read blocks until there is some
data here, but on FTDI the chip seems to return whatever is currently in
the buffer (so 0 bytes if the buffer is empty). This different behaviour
should be mitigated by the changes by this commit.

### Changes made:
1) Return DC_STATUS_TIMEOUT if less bytes than expected are read
2) Use a loop to read bytes from the serial device until enough bytes are present or until timeout runs out.

### Additional information:
Tested devices:
- [x] Mares Puck Pro

### Mentions:
@dirkhh

### DIFF
--- a/android-mobile/src/org/subsurfacedivelog/mobile/AndroidSerial.java
+++ b/android-mobile/src/org/subsurfacedivelog/mobile/AndroidSerial.java
@@ -19,6 +19,7 @@ import java.util.Queue;
 import java.util.List;
 import java.util.LinkedList;
 import java.lang.Math;
+import java.util.Date;
 
 public class AndroidSerial {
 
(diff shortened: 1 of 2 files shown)


dirkhh:
This looks good. Right now I have only one FTDI divecomputer and I have never had timeout issues with that one. But I'll test (but since it looks reasonable, I think I'll merge it later... gotta run now to meet a friend for breakfast)

dirkhh:
tested and works. merging

PR #2659, opened 2020-03-09
Android updates

dirkhh:

- [x] Bug fix
- [x] Functional change

This tries to get the dive computers listed on Android right...
And it fixes the issue where we wouldn't show the download page if Subsurface-mobile gets started by plugging in the dive computer

### Changes made:

1) toss out the manual selection of FTDI dive computers (which was bogus, anyway, as the cable defines the chip, not the dive computer, and some dive computers are supported by multiple cables with different chips)
2) try to select only those transports we support
3) make sure we check if we need to switch to the download page after finished processing main.qml

### Mentions:

@charno 
I haven't tested this enough... but it looked right when I wrote it :-)

### DIFF
--- a/android-mobile/res/xml/device_filter.xml
+++ b/android-mobile/res/xml/device_filter.xml
@@ -20,20 +20,6 @@
     <!-- Mares Icon HD Custom PID -->
     <usb-device vendor-id="0xFFFF" product-id="0x0005"/>
 
-    <!-- USB devices -->
-    <!-- EON Steel -->
-    <usb-device vendor-id="0x1493" product-id="0x30"/>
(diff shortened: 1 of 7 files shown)


dirkhh:
Hmm - this isn't right yet. Dang it. Needs more work. Maybe tomorrow.
Leaving it here for now, but definitely not ready to merge.

dirkhh:
This is now tested and appears to do what I want it to do... waiting for the CI
I'd love it if people were to review this :-)

dirkhh:
oops, two of these commits were dropped when I merged everything locally -> #2663

PR #2660, opened 2020-03-09
[RFC] Remove dive-number editing code

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

While looking at the divelist commits, it occurred to me that the dive-number can't be edited directly anymore (since when)? Instead, one has to use the "renumber dives" context menu. Since no-one complained, let's just remove to corresponding code.

### Changes made:

1) Remove dead code.
### Mentions:

@dirkhh @mturkia @atdotde

### DIFF
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -163,11 +163,6 @@ int editMode(int index, int newValue, bool currentDiveOnly)
 	return execute_edit(new EditMode(index, newValue, currentDiveOnly));
 }
 
-int editNumber(int newValue, bool currentDiveOnly)
-{
-	return execute_edit(new EditNumber(newValue, currentDiveOnly));
-}
-
 int editSuit(const QString &newValue, bool currentDiveOnly)
 {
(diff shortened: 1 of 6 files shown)


dirkhh:
I was pretty sure that I just added that feature on mobile...
Yes, on the desktop you use renumber, even for a single dive. But on mobile that's something we allow.

bstoeger:
Ok - then we should drop the second commit here. Waiting for the PR with the mobile-undo-changes.

dirkhh:
So this no longer applies cleanly because of 35b33b8c6a - one of the changes to make things compile, I guess.
Having said this, I took the time to read through my code and realized that you are of course right - this should be merged because we aren't calling `Command::editNumber` in the mobile app, instead we are doing this in `QMLManager::commitChanges()` which reasonably tracks "whole dive changes".
Would you like me to just manually merge this?

dirkhh:
merged manually

PR #2661, opened 2020-03-09
cleanup: fix NOCLOUD_LOCALSTORAGE leak

bstoeger:
The macro NOCLOUD_LOCALSTORAGE creates the path to the local git
repository as a C-string. None of the users were freeing the string
and thus leaking memory.

Replace the macro by an inline function that creates a QString
and pass down to C-functions using the qPrintable() macro.
Note that every qPrintable() invocation does a UTF16->UTF8
conversion. This could be avoided by either using a std::string
or a QByteArray. However, we are talking about microseconds of
CPU time in operations that typically take seconds. Not worth
it.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a leak concerning the local git repository.

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -374,13 +374,17 @@ void QMLManager::updateAllGlobalLists()
 	updateSiteList();
 }
 
+static QString nocloud_localstorage()
+{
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
this makes sense

PR #2663, opened 2020-03-10
Android updates (two missing commits)

dirkhh:
Somehow these two commits were dropped when I merged the Android updates branch... #2659

- [x] Functional change

don't offer to open non-USB-serial devices on Android. We don't support those on Android.
do away with the broken special-casing for FTDI dive computers


charno:
Just installed and ran it.
Since the rest already worked fine for me, the only thing I really can say is that the app went straight to the download page when started for the first time with the "USB Device Connected"-intent.

PR #2664, opened 2020-03-10
Add timestamps to libdivecomputer.log

charno:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The added timestamps in the libdivecomputer.log can be used mainly to figure out timing problems when talking to a divecomputer. The motivation for this change are the current communication problems with the usb-serial-for-android implementation in respect to Suunto computers.

### Changes made:
1) copy timer.{c,h} from libdivecomputer into core/
2) use the timestamps from said timer in the logfunc in libdivecomputer.c

### Release note:

Probably not?

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -159,6 +159,8 @@ set(SUBSURFACE_CORE_LIB_SRCS
 	taxonomy.c
 	taxonomy.h
 	time.c
+	timer.c
+	timer.h
 	trip.c
 	trip.h
 	uemis-downloader.c
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -21,6 +21,7 @@
 #include "display.h"
 #include "errorhelper.h"
 #include "sha1.h"
+#include "timer.h"
 
 #include <libdivecomputer/version.h>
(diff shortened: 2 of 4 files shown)


dirkhh:
I assume that you have tested this. It's kinda freaky to see the code that clearly expects the `#define`s as provided by `autoconf` in our sources - but I guess you took me very literally when I suggested to use Jef's code :-)

dirkhh:
Tested on Android. Thank you

charno:
Hi @dirkhh 

Yes I took it very literal, since I already did some tries with just copying the code relevant to get current time in microseconds, bud I got lost in the #ifdef hell and realized that I would never manage to get it right without being able to test on all platforms. (Jef's method does calculate the delta to the start timepoint also in the platform-dependend code).

And now that you mention it... I did also test this on Android. I just wanted to download the linux version to check on the desktop as well, but I couldn't find the build artifacts from master. Since this is clearly different for different platforms, maybe it would make sense to run at least a basic "it does not crash when downloading" test on the other platforms. I only have linux...

Issue #2657, opened 2020-03-08
Android-Mobile USB-Serial Devices: Allow precise device selection.


charno:
### Describe the issue:

- [ ] Bug
- [x] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
This is the attempt to write a somewhat coherent description (at least how I see it ;-)) of the results from discussing on the mailing list.

The current behavior is that a dummy device "usb-serial" is used. When this device is selected, the device driver is using a (VID/PID-Based) detection mechanism to attach a Chip-specific driver-implementation to the first found USB-to-Serial interface. This works fine as long as there is only one USB-to-Serial interface attached and the VID/PID is known.

Two scenarios are not working with the current implementation:
- More than one known USB-to-Serial interface is attached
- A USB-to-Serial interface without known VID/PID pair, but which is supported by usb-serial-for-android is used.

Both this scenarios are solved with my proposal.

**User-Facing change 1**: Listing actual USB devices. This allows for the usecase with multiple attached USB-to-Serial interface.
- Instead of populating the device-list with a dummy "usb-serial" interface, actually display all connected USB-Devices (without knowing if they are USB-to-Serial interfaces) in the device-list. Something like:
  - "USB Device: CP2102 USB to UART Bridge Controller"
  - "USB Device: FT230X Basic UART"
- The user can then select which USB device he/she wants to use.

**User-Facing change 2**: Allow for Driver selection. This allows for the usecase with unknown VID/PID pairs.
- Provide a Setting like "Allow for USB driver selection" in the app settings.
- When this setting is enabled: Instead of populating the device-list with a dummy "usb-serial" interface, actually display all connected USB-Devices (without knowing if they are USB-to-Serial interfaces) and all drivers known to usb-serial-for-android in the device-list. Something like:
  - "USB Device: CP2102 USB to UART Bridge Controller (autoselect driver)" // This would be the same behavior as in usecase 1
  - "USB Device: CP2102 USB to UART Bridge Controller (CdcAcmSerialDriver)"
  - "USB Device: CP2102 USB to UART Bridge Controller (Ch34xSerialDriver)"
  - "USB Device: CP2102 USB to UART Bridge Controller (Cp21xxSerialDriver)"
  - "USB Device: CP2102 USB to UART Bridge Controller (FtdiSerialDriver)"
  - "USB Device: CP2102 USB to UART Bridge Controller (ProlificSerialDriver)"
[...]

PR #2668, opened 2020-03-11
android usb serial: Prepare device / driver select

charno:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR contains the serial_android_usb part of the changes proposed
in issue #2657.

This commit keeps compatibility with the current UI-Code in the case that only one USB-Device is connected. If two devices are connected, only the first one is tried (instead of all are tried, and the first for which usb-serial-for-android finds a driver is used)

### Changes made:

1) A data structure that contains all the data that can be used to describe an usb device (including user-facing string).
2) A function to get a list of all attached usb devices (optionally with selectable driver class).
3) Changes in the serial_android_usb_open-function and in the Java part to use the information about the usb device and optionally selected driver when connecting.

### Related issues:
#2657

### Additional information:

There are still some small things to do:
- Change the user-facing string to something more descriptive.
- Parts which aren't uesd anymore when the UI-Part is implemented are simply marked as obsolete (to keep compatibility for now).

### Release note:

When the UI changes are done.

### Documentation change:

When the UI changes are done.

### Mentions:

@dirkhh

### DIFF
--- a/android-mobile/AndroidManifest.xml
+++ b/android-mobile/AndroidManifest.xml
@@ -92,6 +92,7 @@
         android:normalScreens="true"
         android:smallScreens="true" />
 
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
     <!--
          The following comment will be replaced upon deployment with default permissions based on the dependencies of the application.
          Remove the comment if you do not require these default permissions.
--- a/android-mobile/src/org/subsurfacedivelog/mobile/AndroidSerial.java
+++ b/android-mobile/src/org/subsurfacedivelog/mobile/AndroidSerial.java
@@ -14,6 +14,8 @@ import android.content.Intent;
 import org.subsurfacedivelog.mobile.SubsurfaceMobileActivity;
 
 import java.lang.System;
+import java.lang.Class;
+import java.lang.reflect.Constructor;
 import java.lang.Thread;
 import java.util.Queue;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


dirkhh:
I started working on this but unfortunately got side-tracked with real life. I'll try to find time to continue to work on it today.
Right now my plan is not to merge it until I have at least a vague idea that this is what we need :-)

charno:
I think it only makes sense to merge this once the UI changes are also done. The current status does not improve usability at all. In fact, when this patch is applied the edge-case that somebody has two USB-devices (one recognized usb-serial deivce, and one other device) is attached might break depending on the order of enumeration.

dirkhh:
Making progress.
3 dive computers connected, it already picks the one that sent the intent...
Screenshot_20200314-131644

dirkhh:
Next I need to think about how to hide the specific drivers in the list.
Here's my current thinking:
- if we recognize the VID/PID, then we know the correct driver, so there isn't a point in showing the driver choices
- if we have another device which we don't recognize, then it makes sense to offer the different drivers
Does that sound logical @charno ? Am I missing something?

I also still need to implement the intent that you stubbed out, but that shouldn't take too long.

charno:
Looking good! :-)

That's one way to do it. You mean show the driver selection in the case that at least one device is attached for which we don't know the VID/PID pair, right?

My proposal was that there could be a setting to enable / disable advanced driver selection. This would maybe be more easy to implement, and maybe the users then would have more initiative to send their VID/PID pair, which then could be added to the list of known pairs.

But I don't feel strongly one way or the other...

dirkhh:
What I mean is this: 
- if a device has one of the known VID/PID pairs, then show it only once with no driver info
- if there's a device that we don't recognize, show it 5 times with the various drivers
This does the right thing in the normal case (only one device, known VID/PID), and also in the complex cases (multiple devices and possibly one or more of them with unknown VID/PID).
I should have a screen shot to show what I mean in a few minutes...

charno:
Aaaah...

Did you see the bool parameter of the `serial_usb_android_get_devices(bool driverSelection)`? This bool does basically what you try to achieve, but over all devices.

What it does is this:
- If false:
  - Only one device? Set `uiRepresentation` to `"USB Connection"`, don't set driver class. Only one entry.
  - Multiple devices? Set `uiRepresentation` to the format in your screenshots without the parentheses at the end. Don't set driver class. One entry per device.
- If true:
  - Show 6 Entries per Device. For each device, one without driver class, 5 with different driver classes. `uiRepresentation` is in the format from the screenshots.

So my thought was to set the driver selection option over all devices, not on specific ones.

charno:
This was done in the anticipation to have a setting to enable / disable the driver selection feature. 

But maybe it makes sense to have multiple entries for devices where we don't know the PID/VID (and the setting is off). This could be quite easily implemented in the `serial_usb_android_get_devices(bool driverSelection)` function.

dirkhh:
Our preferences where to implement logic is driven by where we are comfortable.
I typically cannot read Java code. It literally takes me several minutes of staring at things and usually some googling to make sense of the functions you implement.
So no, I didn't notice that that was what you intended there.
And given that usually I end up being the one having to maintain things in the long run, I shy away from relying on code that I feel entirely incapable of fixing.

So let me play with my vision here for a moment longer - the positive side effect of my vision is that I don't need another preference setting. And we have already 2x the number of settings that we SHOULD have in order to keep the settings screen useful. And since I assume that 0.001% of our users might ever run into this issue, if I can avoid adding a setting, that's the path I'd be likely to prefer...

charno:
I see, I'm also much more fluent in C++, my Java experience is quite rusty and was not used for several years. This (and that Subsurface is implemented mainly in C/C++) is why I implemented most of the functionality in C++, not Java.

The `serial_usb_android_get_devices()` is actually completely in C++, with some JNI calls to Android Services only. No Subsurface-specific Java-Implementation is used in this functionality. (https://github.com/charno/subsurface/blob/usb-serial-new-interface/core/serial_usb_android.cpp#L288).

I agree that it's preferable to have no new preference setting.

Currently you seem to always call the `serial_usb_android_get_devices` function with driverSelection set to true. Can you try to set it to false and check if this is more to what you envision?

I can implement the changes you propose tomorrow; so I would:
- Remove the bool parameter (since it's not used when no user-preference is available)
- For each device with known VID/PID pair: Display one entry (`FTDI [1:4]`), no `className` set.
- For each device with unknown VID/PID pair: Display five entries (`FTDI [1:4] (CdcAcmSerialDriver)`, `FTDI [1:4] (CdcAcmSerialDriver)` etc.)

Did I miss something?

When done like this, the whole logic for filling the list would be in one place, and no additional filter needs to be applied.

What do you think?

dirkhh:
I'll keep playing with this. Thanks to COVID19 and a surprise snow event in Portland I have some time to spend on this today... let's see what I end up with :-)

charno:
Can you maybe push the results somewhere and send me a link to the branch? When I'll do some work tomorrow it'll be still in the middle of the night for you, and maybe I can then check what changes would make sense.

dirkhh:
yes, I'll push this in a branch here when I'm done for today...

dirkhh:
So this is what I'm doing now...

Needs lots of cleanup, but also seems reasonable

Screenshot_20200314-144734

dirkhh:
The "Plantronics" is a BLE headset plugged into my USB hub, the other two are an ancient Mares Puck Air (forgot I had that - had to buy a new battery) and the Icon HD (I want to switch out the name that we display for that one).

charno:
Looks good! :-)

Have you seen my comment above? (https://github.com/Subsurface-divelog/subsurface/pull/2668#issuecomment-599136357)

I'm just asking because you answered very shortly afterwards and I'm not sure if it got lost in the crossfire. The result would be the same.

dirkhh:
> I see, I'm also much more fluent in C++, my Java experience is quite rusty and was not used for several years. This (and that Subsurface is implemented mainly in C/C++) is why I implemented most of the functionality in C++, not Java.
> 
> The `serial_usb_android_get_devices()` is actually completely in C++, with some JNI calls to Android Services only. No Subsurface-specific Java-Implementation is used in this functionality. (https://github.com/charno/subsurface/blob/usb-serial-new-interface/core/serial_usb_android.cpp#L288).
> 
> I agree that it's preferable to have no new preference setting.
> 
> Currently you seem to always call the `serial_usb_android_get_devices` function with driverSelection set to true. Can you try to set it to false and check if this is more to what you envision?
> 
> I can implement the changes you propose tomorrow; so I would:
> 
> * Remove the bool parameter (since it's not used when no user-preference is available)
> * For each device with known VID/PID pair: Display one entry (`FTDI [1:4]`), no `className` set.
> * For each device with unknown VID/PID pair: Display five entries (`FTDI [1:4] (CdcAcmSerialDriver)`, `FTDI [1:4] (CdcAcmSerialDriver)` etc.)
> 
> Did I miss something?
> 
> When done like this, the whole logic for filling the list would be in one place, and no additional filter needs to be applied.
> 
> What do you think?

yes, I saw that. Given that it's very late in Switzerland and just 15:00 over here I figured that I'd just make those changes myself - I was confused and thought that the logic was in Java. Since you already have that in C++ I'll just move my code into that function. Of course now I need to switch things from `QString` which I'm very familiar with to `std::string` which I am not very familiar with, but I'll just consider this yet another learning opportunity...

charno:
Ok, perfect, that's fine for me.

Feel free to change the type in the struct to QString, there isn't any std::string specific stuff going on, just passing around some strings.

dirkhh:
this has been superseded by #2674

PR #2673, opened 2020-03-13
iostream: fix incorrect rfcomm error case when reading

torvalds:
We had two independent bugs here, both of which needed to fire for this
to cause a problem.  This fixes both of them.

The first bug was that our rfcomm code would return DC_STATUS_SUCCESS
with a zero-sized read when a timeout happened, or when the rfcomm
socket had disconnected.  That makes absolutely no sense.  We should
return DC_STATUS_TIMEOUT on timeout, and DC_STATUS_IO if the socket has
disconnected without any data.

The fix to this is to make the whole rfcomm iostream read logic much
simpler: there's no need to loop at all for partial results, because the
libdivecomputer iostream side will do the loop for us (and handle
partial results much better: it knows if the target backend can handle
those partial results or not).

The second bug was in our libdivecomputer iostream read() function,
which reacted very badly to this bad return value.  This updates our
libdivecomputer branch to one that is more careful about things.

Reported-by: linuxcrash <albin@mrty.ch>
Debugged-by: Jef Driesen <jef@libdivecomputer.org>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/qtserialbluetooth.cpp
+++ b/core/qtserialbluetooth.cpp
@@ -172,32 +172,27 @@ static dc_status_t qt_serial_write(void *io, const void* data, size_t size, size
 {
 	qt_serial_t *device = (qt_serial_t*) io;
 
-	if (device == NULL || device->socket == NULL)
+	if (device == NULL || device->socket == NULL || !actual)
 		return DC_STATUS_INVALIDARGS;
 
-	size_t nbytes = 0;
-	int rc;
+	*actual = 0;
+	for (;;) {
+		int rc;
 
-	while(nbytes < size && device->socket->state() == QBluetoothSocket::ConnectedState)
-	{
-		rc = device->socket->write((char *) data + nbytes, size - nbytes);
+		if (device->socket->state() != QBluetoothSocket::ConnectedState)
+			return DC_STATUS_IO;
+
+		rc = device->socket->write((char *) data, size);
 
 		if (rc < 0) {
 			if (errno == EINTR || errno == EAGAIN)
-			    continue; // Retry.
-
-			return DC_STATUS_IO; // Something really bad happened :-(
-		} else if (rc == 0) {
-			break;
+				continue;
+			return DC_STATUS_IO;
 		}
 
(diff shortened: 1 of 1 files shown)


torvalds:
NOTE! I do not actually have any dive computer that uses rfcomm. 

So this is entirely untested, although it _looks_ obvious enough. Before merging it, somebody should really check with a dive computer that uses legacy BT communication.

And all credit for debugging this goes to @jefdriesen in 

https://github.com/libdivecomputer/libdivecomputer/issues/13

linuxcrash:
@torvalds I will test this as soon as I get some time on my laptop to recompile and run Subsurface with my Ratio DC using RFCOMM.

Its a honor to get this fixed by the master himself (hope you don't mind my nickname). ;-)

torvalds:
@linuxcrash @jefdriesen it turns out the exact same issue exists on the rfcomm writing side too, but it's likely that nobody actually cares. 

I do have patches for that side too, but I'd like the reading side to be tested before I then commit and push the write side changes.

torvalds:
On Dirk's suggestion, I just updated the branch to have the write-side change too.

That makes the branch name a bit misleading, but might as well just get both fixes tested at the same time (and assuming testing is successful, eventually merged at the same time).

dirkhh:
@linuxcrash any luck testing this?

linuxcrash:
@torvalds @dirkhh @jefdriesen,
I just tested it again using my Ratio i3XM Tech+ to download 99 dives via bluetooth and it all works as expected. I can herewith confirm that this bug has fixed as expected!

I'd like to thank you guys making this work!

[subsurface_fix.log](https://github.com/Subsurface-divelog/subsurface/files/4333131/subsurface_fix.log)

dirkhh:
That's excellent.
Thanks for reporting and testing, and special thanks to @jefdriesen for tracking down a bug in our fork. That's truly above-and-beyond!

PR #2674, opened 2020-03-15
Android updates

dirkhh:
This builds on top of #2668 - but the commits in that PR have been mildly adjusted (mostly for coding style / white space).

- [x] Functional change
- [x] New feature

With this we now can download from a specific USB device on Android (if multiple devices are connected via a hub). We also have better UI strings, we do a better job detected connected devices, we don't spread knowledge about USB devices in various places... many good things.
This also supports re-scanning the USB devices when tapping on `Rescan` and fixes a number of semi-related bugs in the code.

This isn't ready to be merged, I need to re-read the code, make sure that I have removed all code that was replaced by this, make sure things work as expected when mixing BT and USB dive computers, etc.

And it definitely could use some more testing ; that said: I tried this with up to three dive computers and one non-DC connected to a USB hub. And tried a number of scenarios when it comes to downloads.
I even tested the new permission intent (where you try to download from a USB where you hadn't provided permission, yet). It seems to work well here...

### Related issues:

#2668 
#2657 

### Release note:

missing

### Documentation change:

needed

### Mentions:

@bstoeger @charno


charno:
Hi Dirk,

Thank you very much, I love it! :-) With this, the user experience seems very polished!

I've done two small changes in https://github.com/Subsurface-divelog/subsurface/pull/2675:
- remove "autodetect"-option for unknown PID/VIDs. This makes no sense to have, since we can't detect the driver for it anyways.
- remove /dev/ttyS* for android from the default list again. They snuck in when you changed the `#if` in subsurface-mobile-main.cpp

One question:
- The "danger danger"-comment in core/serial_usb_android.cpp:353, what do you mean by that? there's no index there, but directly the Usb Device to download from. What's the danger you see there?

Some improvements that can be done:
When the app is first startet, and the download-screen is opened (without using the usb intent), the device list should also be refreshed. Currently, only the default entries (FTDI, and before my change the /dev/ttyS* devices) are present. Only after pressing the refresh button the currently connected devices are scanned and displayed.

On Permission granted-intent: Don't only open the download page again (we're already there...!), but directly restart the download (since now we have the permission on the device). 
The only place where the permission is requested is when the user presses the download-button, and permission is not-yet-granted. 
In the permission granted-intent, we also get the UsbDevice, which might be helpful. (`setUsbDevice((UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE, true)` instead of `setUsbDevice(null)`)


Also, you said somewhen earlier that you want to remove the old, not-really-working FTDI driver from the android build. It's your call, but maybe this would be the right time to do it?

To keep the PRs / branches somewhat clean, it might make sense that you just cherry-pick (and probably interactive-rebase) these changes into your branch, and close my new PR.

Thanks a lot, it's really nice to see my work getting polished form an ui perspective!

dirkhh:
Clearly I hadn't built this for anything but Android. Oops. I think I fixed this :-)

dirkhh:
> I've done two small changes in #2675:
> 
> * remove "autodetect"-option for unknown PID/VIDs. This makes no sense to have, since we can't detect the driver for it anyways.
> * remove /dev/ttyS* for android from the default list again. They snuck in when you changed the `#if` in subsurface-mobile-main.cpp

Thanks for those two - I'll cherry-pick them into my branch (but I'll away from my computer for the next couple of hours)

> One question:
> 
> * The "danger danger"-comment in core/serial_usb_android.cpp:353, what do you mean by that? there's no index there, but directly the Usb Device to download from. What's the danger you see there?

That is a reminder to myself that I need to think through the memory management model a bit more. Every time we pass a pointer to an object to something, we need to make sure that the object stays around and is still valid. We get these UsbDevice objects from the Java code - what's their life cycle? Do they just stay around? How do they become invalid? Etc.

> Some improvements that can be done:
> When the app is first startet, and the download-screen is opened (without using the usb intent), the device list should also be refreshed. Currently, only the default entries (FTDI, and before my change the /dev/ttyS* devices) are present. Only after pressing the refresh button the currently connected devices are scanned and displayed.

Some first use thing... that makes sense.

> On Permission granted-intent: Don't only open the download page again (we're already there...!), but directly restart the download (since now we have the permission on the device).
> The only place where the permission is requested is when the user presses the download-button, and permission is not-yet-granted.

I can add that, no problem. And it is logical to do so.

> In the permission granted-intent, we also get the UsbDevice, which might be helpful. (`setUsbDevice((UsbDevice) intent.getParcelableExtra(UsbManager.EXTRA_DEVICE, true)` instead of `setUsbDevice(null)`)

I'm about to run  out of the house - need to look at the sources to make sure I understand what you are saying.
 
> Also, you said somewhen earlier that you want to remove the old, not-really-working FTDI driver from the android build. It's your call, but maybe this would be the right time to do it?

Yes, I will do that next

> To keep the PRs / branches somewhat clean, it might make sense that you just cherry-pick (and probably interactive-rebase) these changes into your branch, and close my new PR.

That's indeed what I'll do
 
> Thanks a lot, it's really nice to see my work getting polished form an ui perspective!

:-)

charno:
> That is a reminder to myself that I need to think through the memory management model a bit more. Every time we pass a pointer to an object to something, we need to make sure that the object stays around and is still valid. We get these UsbDevice objects from the Java code - what's their life cycle? Do they just stay around? How do they become invalid? Etc.

The QAndroidJniObject itself does reference-counting of the underlying Java-Object, and releases the reference once no more QAndroidJniObject pointing to the Java-Object is present. Then Java garbage collection takes over. So no need to release the UsbDevice-Object itself, but only the QAndroidJniObject (and obviously the descriptor-struct).

That's one of the big improvements of QAndroidJniObject in comparison to plain JNI.

dirkhh:
So things that cause me to be concerned...
`getDescriptor()` has a local variable `descriptor` that initializes a std::string from a local  `char buffer[128]` and others by copying strings it gets from JNI (and which it then releases) and all that is returned to the caller. Since we return this as a struct, we copy this on assignment, but if I'm not mistaken, that would be a shallow copy - so with pointers to `char` this would be broken by design, but I'm uncertain how an assignment of `std::string` works in this case.
I assume that this will do the right thing and do a deep copy of the string? And then release the memory when the struct goes out of scope?
`serial_usb_android_get_devices()` creates a `std::vector` of those structures. I assign that `std::vector` to a static `std::vector` - and I overwrite previous values. What's the memory management semantic there? And if I have a pointer to an element in that `std::vector<android_usb_serial_device_descriptor>` and I overwrite that `std::vector` with a new assignment, what happens to that pointer?

This is where I flat out admit that C++ memory management still confuses me. @bstoeger is usually the person who tells me when I'm wrong :-)

bstoeger:
> I assume that this will do the right thing and do a deep copy of the string?

Of course - if you copy a std::string, it does a deep copy. Value semantics - from one make two.

> I assign that std::vector to a static std::vector - and I overwrite previous values. What's the memory management semantic there?

The old data is freed, the new data is copied, unless you use references instead of strings, obviously.

I haven't looked, but perhaps you want to make a
```
        static_vector = std::move(new_vector)
```
That would *not* copy data, but simply change the first,last,capacity pointers of static_vector to that of new_vector and initialize new_vector to the empty vector.

> And if I have a pointer to an element in that std::vector<android_usb_serial_device_descriptor> and I overwrite that std::vector with a new assignment, what happens to that pointer?

The pointer will become invalid. Attention: even inserting into a vector can make all your pointers/interators in that vector invalid. std::vector codifies the realloc/array C-idiom. Often, the best thing you can do is to never use pointers and references, but always values. Though I haven't looked at this PR to decide whether that applies here.

dirkhh:
@bstoeger I love your use of the word "obviously". This reminds me of a great line from a great movie...  https://images.app.goo.gl/WTFD7BTrAFETfSMo9
Anyway... none of this is even remotely obvious to me :-)
And right now we pass a `void *` pointer that points to an element of that `std::vector` through C code back to C++ code that then turns it back into a structure and uses an `QAndroidJNIObject` that's inside that structure.
Which is why that code has the comment `danger, danger` attached :-)
I'm getting the distinct feeling that I should do better than this to prevent this from going KABOOM.

bstoeger:
> And right now we pass a void * pointer that points to an element of that std::vector through C code back to C++ code that then turns it back into a structure and uses an QAndroidJNIObject that's inside that structure.

Yes, depending on the lifetime of that `void *` pointer, that sounds like a bad idea. I will have a look, but probably not today.

dirkhh:
> Yes, depending on the lifetime of that `void *` pointer, that sounds like a bad idea. I will have a look, but probably not today.

I'm quite sure this is a Bad Idea<sup>TM</sup>

charno:
@dirkhh I now see why there's a "danger danger" comment :)

Thanks for @bstoeger for your comments. That's really helpful, and I do learn some things from this. Normally I work in a C++ Only-Codebase with the luxury of using all the fancy shared-pointers etc, so no  manual memory management for me :-)

Currently I'm wondering why it didn't go kaboom until now! ;-)

I'm currently playing around with a possible solution, which is to copy the currently active android_usb_serial_device_descriptor into a separate field so the pointer is not invalidated when the vector changes.

bstoeger:
To me it appears that if we want to stay consistent with the current way things work, then `device_data_t` should be extended by the two(?) text-fields that you need.

That said ... where are all the textual fields in device_data_t free()d? Am I missing something or are we leaking that stuff?

charno:
> To me it appears that if we want to stay consistent with the current way things work, then `device_data_t` should be extended by the two(?) text-fields that you need.

Things are a bit more complicated. What we need to pass around it this:
```
struct android_usb_serial_device_descriptor {
	QAndroidJniObject usbDevice; /* the UsbDevice */
	std::string className; /* the driver class name. If empty, then "autodetect" */
	std::string uiRepresentation; /* The string that can be used for the user interface. */

	// Device information
	std::string usbManufacturer;
	std::string usbProduct;
	std::string manufacturer;
	std::string product;
	uint16_t pid;
	uint16_t vid;
};
```

Or at least the `QAndroidJniObject usbDevice` and `std::string uiRepresentation`. Since one of these is a c++ object, the problem still remains the same

dirkhh:
The reason this currently doesn't go KABOOM is that we are helping our luck out a little by allocating these objects right before we call into the download function and not messing with that `std::vector` at all (and therefore not causing a `re-alloc` or `free`) until after we are finished. But that's not a sane way of doing this.
My first idea would be to have a static copy of the `struct` in the calling function and pass a pointer to that into the C code. That should ensure that the pointer stays valid

charno:
> The reason this currently doesn't go KABOOM is that we are helping our luck out a little by allocating these objects right before we call into the download function and not messing with that `std::vector` at all (and therefore not causing a `re-alloc` or `free`) until after we are finished. But that's not a sane way of doing this.

Aaaah... :-) I tried to follow the ways when things were called but got lost in qt/qml magic.

> My first idea would be to have a static copy of the `struct` in the calling function and pass a pointer to that into the C code. That should ensure that the pointer stays valid

That sounds good to me. Very similar to my proposal, but implemented in a more local way.

dirkhh:
with that last force push I have implemented (and tested) the two things @charno requested: granting the permission simply continues the download. And the first time you enter the download screen, USB devices are queried

charno:
https://github.com/Subsurface-divelog/subsurface/commit/671faae7584aa6061e68d282e8719e32ee1f1a4a

VERY similar actually. I didn't test yet.

bstoeger:
> Things are a bit more complicated. What we need to pass around it this:

All of it!? Wow.

> My first idea would be to have a static copy of the struct in the calling function and pass a pointer to that into the C code. That should ensure that the pointer stays valid

Ugh. Can't we simply create a copy on the heap and `delete` that in the yet-to-be-written(?) code that also `free()`s the rest of the data?

Not a fan of non-reentrant code, to be honest.

dirkhh:
@bstoeger as you know, memory mismanagement is my personal specialty.
The code that is being called is DECIDEDLY non-reentrant. So I don't think @charno 's solution makes it any worse. But yes, this could just as easily done on the heap. I'll think about this some more after lunch.
Thanks for helping us figure this out!

charno:
> All of it!? Wow.

At least the QAndroidJniObject usbDevice and std::string uiRepresentation. Since one of these is a c++ object, the problem still remains the same.

Have a good lunch :)

bstoeger:
Can't you store the "static" version right next to `device_data_t data` in the `DCDeviceData` class? That would make it reentrant. Just say no to variables with static storage. :-P

charno:
> with that last force push I have implemented (and tested) the two things @charno requested: granting the permission simply continues the download. And the first time you enter the download screen, USB devices are queried

Just tested as well, works nice, thanks.

dirkhh:
@bstoeger I think what's in e4ad519 is sane and works.

bstoeger:
> I think what's in e4ad519 is sane and works.

That's just wrong.

*If* you want to go that way you have to use `new` and `delete` so that the destructor is called!

However I still don't understand why you don't store your `android_usb_serial_device_descriptor` as a subobject in DCDeviceData alongside the `data` member?

That seems so simple, no malloc, new or anything...?

dirkhh:
> > I think what's in [e4ad519](https://github.com/Subsurface-divelog/subsurface/commit/e4ad519b07b9655d0c430c70e8ee671866ab451b) is sane and works.
> 
> That's just wrong.
> 
> _If_ you want to go that way you have to use `new` and `delete` so that the destructor is called!

Sweet. It's a struct. I can't use new on a struct.
Did I mention just how much I am not comfortable with memory management. It's not even funny.

> However I still don't understand why you don't store your `android_usb_serial_device_descriptor` as a subobject in DCDeviceData alongside the `data` member?
> 
> That seems so simple, no malloc, new or anything...?

Ok, I'll see if I understand how that would work.

dirkhh:
No, that cannot work. I cannot include std::whatever objects and Qwhatever objects inside a structure that is used in C code. And no, I will not re-implement `libdivecomputer.c` in C++. Thank you for not asking.

bstoeger:
> Sweet. It's a struct. I can't use new on a struct.

Why would you think so? The difference between `class` and `struct` is purely syntactic sugar: If not specified otherwise, elements in `class` default to `private:`, in `struct` to `public:`. That's it.

In C you cant use new and delete, that's true. However all the code in your last commit was C++.

dirkhh:
Not to sound too stupid (that's often hard for me).
I have no idea what you are talking about. I have no idea what the words even mean. As in
"what does `bla = new struct whatever` mean? what does it do?"
not the slightest clue. Nor do I know how that would make any difference when copying the value over.
I think I should just stop pretending that I can be the maintainer of a project that is using language constructs that were developed a decade or two after I stopped learning things

bstoeger:
> No, that cannot work. I cannot include std::whatever objects and Qwhatever objects inside a structure that is used in C code.

`DCDeviceData` is a C++ class. Just put the `android_usb_serial_device_descriptor` in *there* as a subobject, *not* a pointer to an object.

In the set function simply assign to that object and set the `void *` pointer in the C class.

Subobjects without pointers are the best way of memory management, because it means that you don't have to do memory management. :)

dirkhh:
I will grab a nice hammer and smash it into my forehead. That sounds equally pleasant compared to dealing with this.

bstoeger:
> bla = new struct whatever

I guess bla is a pointer?
`new` is (conceptually) nothing more than malloc() + call constructor.
`delete` is (conceptually) nothing more than call destructor + free().

But I still think that you simply shouldn't allocate that stuff on the heap. Keep it together with the device_data_t thing - since it appears to have the _same lifetime_.

> I think I should just stop pretending that I can be the maintainer of a project that is using language constructs that were developed a decade or two after I stopped learning things

I hope you realize that new, delete etc were invented in Chernobyl-explosion time frame (I think). :-P

> I will grab a nice hammer and smash it into my forehead. That sounds equally pleasant compared to dealing with this.

I really don't understand where you have a problem right now...:

```
class DCDeviceData {
...
private:
        device_data_t data;
#ifdef ANDROID
        android_usb_serial_device_descriptor android_stuff
#endif
};
```
and
```
void DCDeviceData::set_android_stuff(const android_usb_serial_device_descriptor &descriptor)
{
        android_stuff = descriptor;
        data.androidUsbDeviceDescriptor = &android_stuff;
}
```

What is wrong with that?

dirkhh:
You really don't understand how fundamentally uncomfortable and unfamiliar I am with things that are completely normal and obvious to you.
I really appreciate your patience and willingness to help me out.
Some days my frustration gets away from me and I express myself less grateful that I should
My apologies

dirkhh:
So how about a9ce191
I actually wrote that just before I saw your last comment... and it's remarkably similar to what you suggested. Maybe there is hope for me.
Nah, whom am I kidding.

bstoeger:
> So how about a9ce191

That makes sense to me.
I just would pass a const-reference to `DCDeviceData::setUsbDevice()` to avoid an unnecessary deep copy of all the strings.

It still bothers me that apparently the device_data_t data is never freed - but apparently that has been like that forever and nobody ever complained.

dirkhh:
There are many parts of the code that don't properly free storage.
I believe I have updated this to do what you suggested

charno:
Hi @dirkhh and @bstoeger

Thanks for all your work! I just installed this version on my phone and did a bit of an extended test run. During this, I found three more points which might make sense to change before releasing:
- When I plugin my computer while subsurface is in the foreground, and then use the device connected-intent, the app freezes.
- When the download page is opened, or when I press the refresh-button, no entry in the connection field is selected (it shows as empty). It might make sense to select the first entry by default? 
- Maybe it makes sense to list USB devices before bt-devices? Plugging in an usb device is most likely done with the purpose of downloading from this device, so showing it more prominently than the bt-devices might make sense. Also, this would pair nicely with selecting the first entry by default. (I couldn't test if this is already the case since I don't own a bt divecomputer, but from the code it looks like bt would be listed first)

Otherwise it works like a charm!

Dirk, I feel your pain and frustration. Besides coding I'm maintaining a yocto-based linux distribution for work, that's also a task where the strong urge to pull out lots of hair is permanently present. But the more frustration, the bigger the relief once everything runs as planned :-)

PR #2677, opened 2020-03-16
git: return strdup()ed empty string on error in pop_cstring

bstoeger:
The pop_cstring() function is used by the git parser to
duplicate a quoted string. On error, it returns an empty
string literal. Since the caller expects a copied string
and takes ownership of that string, it will ultimately
been freed.

Concrete example: a log with erroneous cylinder data was opened
getting such an empty string literal as description. On closing or
syncing with the cloud, the dive is freed, leading to a free
of the string literal -> crash.

Return a copy of the empty string instead.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a crash when opening and closing erroneous git logs. The whole proposition of the code appears scary to me: the parsing functions are either called with copies of strings or pointers into shared data depending on whether the textual data is quoted or not.

### Changes made:

1) Return copies of the empty string from `pop_cstring()` on error to be consistent with the non-error case.

### Release note:

No - this is an obscure bug that can only be hit with broken git-logs.

### Mentions:

@dirkhh @torvalds

### DIFF
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -334,11 +334,11 @@ static char *pop_cstring(struct membuffer *str, const char *err)
 
 	if (!str) {
 		report_error("git-load: string marker without any strings ('%s')", err);
-		return "";
+		return strdup("");
 	}
 	if (!str->len) {
 		report_error("git-load: string marker after running out of strings ('%s')", err);
-		return "";
+		return strdup("");
 	}
 	len = strlen(mb_cstring(str)) + 1;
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
That looks like the right thing to do.
Paging @torvalds

bstoeger:
Got an ack from Linus by mail.

PR #2678, opened 2020-03-16
Android usb ui tuning

dirkhh:
simply implementing the two suggestions @charno made (well, one with a twist)

- [x] Functional change
- [x] New feature

- show USB devices before BT/BLE devices on Android
- try to keep the same connection selected when user taps refresh - if the previous selection isn't available anymore, pick first one in the list


charno:
Even better, thanks! :-)

The code is very obvious. I can test it tomorrow if you think it's necessary to have a second test?

dirkhh:
I tested it on three devices with a number of scenarios. I'll merge as is :-)

PR #2680, opened 2020-03-17
RFC: Qmlmanager cleanup

dirkhh:

- [x] Bug fix
- [x] Code cleanup

This is a rather drastic change. Mostly driven by the fact that the more I read the code, the less sense it made. I simply wasn't able to construct a use case where that whole convoluted second run through carefully validating credentials, syncing with the remote, etc, made any sense at all.
Maybe back when we synced the webservice ID there was a point to it? But then, why re-sync the git repo?
Clearly, this needs massive amounts of testing. And I'm not sure how to make sure that happens without creating an Android beta - and thereby risking that we mess things up (potentially badly) for people who are mostly just trying to get the latest features / USB Android, etc.

### Changes made:

rip out that whole second round of connecting to the backend

### Mentions:

@bstoeger @charno @mturkia @janmulder @atdotde @glance- @torvalds 
This really, really needs some testing of Subsurface-mobile. I think it should be fine to do that testing as mobile-on-desktop, but of course testing on device wouldn't hurt, either.

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -327,11 +327,12 @@ void QMLManager::applicationStateChanged(Qt::ApplicationState state)
 
 void QMLManager::openLocalThenRemote(QString url)
 {
+	// clear out the models and the fulltext index
 	MobileModels::instance()->clear();
 	setNotificationText(tr("Open local dive data file"));
-	if (verbose)
-		appendTextToLog(QString("Open local dive data file %1").arg(url));
+	appendTextToLog(QString("Open dive data file %1 - git_local only is %2").arg(url).arg(git_local_only));
 	QByteArray fileNamePrt = QFile::encodeName(url);
+
 	/* if this is a cloud storage repo and we have no local cache (i.e., it's the first time
 	 * we try to open this), parse_file will ALWAYS connect to the remote and populate the cache.
 	 * Otherwise parse_file will respect the git_local_only flag and only update if that isn't set */
@@ -494,6 +495,7 @@ QString QMLManager::getCombinedLogs()
 
 void QMLManager::finishSetup()
 {
+	appendTextToLog("finishSetup called");
 	// Initialize cloud credentials.
 	git_local_only = !prefs.cloud_auto_sync;
 
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -388,9 +388,6 @@ void QMLManager::openLocalThenRemote(QString url)
 		if (qPrefCloudStorage::cloud_verification_status() != qPrefCloudStorage::CS_NOCLOUD)
 			appendTextToLog(QStringLiteral("have cloud credentials, but user asked not to connect to network"));
 		alreadySaving = false;
-	} else {
-		appendTextToLog(QStringLiteral("have cloud credentials, trying to connect"));
-		tryRetrieveDataFromBackend();
 	}
 	updateAllGlobalLists();
 }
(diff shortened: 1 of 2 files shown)


mobile-widgets/qmlmanager.cpp
```diff
@@ -797,6 +690,7 @@ void QMLManager::loadDivesWithValidCredentials()
 	// if we aren't switching from no-cloud mode, let's clear the dive data
 	if (!noCloudToCloud) {
 		appendTextToLog("Clear out in memory dive data");
+		clear_dive_file_data();
```
bstoeger: Unnecessary - MobileModels::instance()->clear() will take care of that.

mobile-widgets/qmlmanager.cpp
```diff
@@ -516,6 +516,8 @@ void QMLManager::finishSetup()
 		qPrefCloudStorage::set_cloud_verification_status(qPrefCloudStorage::CS_NOCLOUD);
 		saveCloudCredentials(qPrefCloudStorage::cloud_storage_email(), qPrefCloudStorage::cloud_storage_password(), qPrefCloudStorage::cloud_storage_pin());
 		appendTextToLog(tr("working in no-cloud mode"));
+		clear_dive_file_data();
```
bstoeger: unnecessary - the next line will take care of that.

mobile-widgets/qmlmanager.cpp
```diff
@@ -327,11 +327,13 @@ void QMLManager::applicationStateChanged(Qt::ApplicationState state)
 
 void QMLManager::openLocalThenRemote(QString url)
 {
+	// clear out the dive table, dive data, models and the fulltext index
+	clear_dive_file_data();
```
bstoeger: Unnecessary - the next line will take care of that.

dirkhh:
You can tell, I was grasping at straws there. I'll drop all that, then.

bstoeger:
Can't really test because even current master hangs when I edit a divesite..?

bstoeger:
I like it: when trying to add a divesite to the divesite-table twice:
```
        while (ds->uuid == 0 || get_dive_site_by_uuid(ds->uuid, ds_table) != NULL)
                ++ds->uuid;
```
of course is an infinity loop. :-D Something is wrong in the edit-code, because we must not add an already existing dive site to the table.

bstoeger:
Ok - understood:
In `create_dive_site_with_gps` we add the dive site to the table and then add it again in the undo command. we'll have to split that in two functions.

dirkhh:
So these things are broken in master, not broken by this PR, correct?

bstoeger:
Yes - PR in a few min.

dirkhh:
merged

dirkhh:
dropped the redundant commit and rebased ontop of master

bstoeger:
Hm.. Now you also removed the `MobileModels::instance()->clear()` lines that you had added previously. However, I haven't checked whether those are necessary. My point was that `MobileModels::instance()->clear()` does a `clear_dive_file_data()` and therefore the *combination* is redundant.

dirkhh:
I had added one such call and was able to verify that that one was redundant as that code path could ONLY be hit when first opening the app.

bstoeger:
OK - I see a potentially very fundamental problem:

This function:
```
void QMLManager::mergeLocalRepo()
{
        struct dive_table table = empty_dive_table;
        struct trip_table trips = empty_trip_table;
        struct dive_site_table sites = empty_dive_site_table;
        parse_file(qPrintable(nocloud_localstorage()), &table, &trips, &sites);
        add_imported_dives(&table, &trips, &sites, IMPORT_MERGE_ALL_TRIPS);
}
```
attempts to read into fresh tables and then add the dives to the divelog.

However, if `parse_file()` is called on a git repository, the git code will load directly into the global tables!

Can you explain the meaning of this function - I'm not sure I understand its purpose.

dirkhh:
The logic is that we have a local git repo, we parse the content into those tables and then we import those into the currently open dive list.
The idea is that the user starts out (for whatever reason) with no-cloud on their phone, so they use a local repo with no cloud backing. Now they want to switch to a cloud account. We should open that cloud account and sync with the remote, and then import the local git repo into that.
If there's a better way to do that, I'm all ears.

bstoeger:
How can I test that?
I edited a dive, manually synced with cloud and only got:
```
INFO: "28.908: Save changes to local cache"
INFO: "28.911: local save requested with no unsaved changes"
INFO: "28.911: Synchronising data file"
INFO: "28.912: Sync with cloud storage"
INFO: "29.864: Waiting for cloud connection (1 second(s) passed)"
INFO: "29.994: Successful cloud connection, fetch remote"
INFO: "31.350: Push local changes to cloud storage"
INFO: "33.445: Transfer to storage (0/7)"
INFO: "35.060: Transfer to storage (7/7)"
INFO: "35.062: Done syncing with cloud storage"
already have loaded SHA 46d04c8944cd93e2d4b36d5256b04abdfe7f0a7a - don't load again
INFO: "35.063: Cloud sync shows local cache was current"
INFO: "41.994: AppState changed to inactive with no save ongoing and no unsaved changes"
```
I don't understand that cloud business. :(

bstoeger:
Lol - when I change to "no cloud mode" it crashes!

dirkhh:
> Lol - when I change to "no cloud mode" it crashes!

Not quite seeing the humor in this.
I tested 'no cloud' mode in several scenarios this morning. Can you say a bit more about how you can trigger this crash? Were you able to run this in a debugger to pinpoint where we are crashing?

bstoeger:
> Lol - when I change to "no cloud mode" it crashes!

More precisely: It crashes on master, but does *not* crash with this PR. So it seems that fixes at least *some* bugs.

dirkhh:
> How can I test that?
> I edited a dive, manually synced with cloud and only got:
> 
> ```
> INFO: "28.908: Save changes to local cache"
> INFO: "28.911: local save requested with no unsaved changes"
> INFO: "28.911: Synchronising data file"
> INFO: "28.912: Sync with cloud storage"
> INFO: "29.864: Waiting for cloud connection (1 second(s) passed)"
> INFO: "29.994: Successful cloud connection, fetch remote"
> INFO: "31.350: Push local changes to cloud storage"
> INFO: "33.445: Transfer to storage (0/7)"
> INFO: "35.060: Transfer to storage (7/7)"
> INFO: "35.062: Done syncing with cloud storage"
> already have loaded SHA 46d04c8944cd93e2d4b36d5256b04abdfe7f0a7a - don't load again
> INFO: "35.063: Cloud sync shows local cache was current"
> INFO: "41.994: AppState changed to inactive with no save ongoing and no unsaved changes"
> ```
> 
> I don't understand that cloud business. :(

So we are talking to the back end server and are waiting to see it identify itself (that's the waiting for cloud connection business). Then we fetch the remote to make sure we have all commits in our local git repo. After we merge those (if necessary), we push the new state back out to the cloud server. Since that sync could have changed the local state, we then check if what we have in the local repo still matches what we have in memory.

Seems reasonably logical and straight forward to me. How can I help you understand this better?

bstoeger:
> Can you say a bit more about how you can trigger this crash?

When changing to no-cloud mode, reliably.

> Were you able to run this in a debugger to pinpoint where we are crashing?

```
#0  0x00007ffff576b8d2 in __GI___libc_free (mem=<optimized out>) at malloc.c:3124
#1  0x000055555573cde0 in free_cylinder (c=...) at /home/bs/src/subsurface/core/equipment.c:34
#2  clear_cylinder_table (table=table@entry=0x55555f996120) at /home/bs/src/subsurface/core/equipment.c:63
#3  0x000055555572720a in free_dive_structures (d=0x55555f9960f0) at /home/bs/src/subsurface/core/dive.c:371
#4  0x0000555555727242 in free_dive_structures (d=0x55555f9960f0) at /home/bs/src/subsurface/core/dive.c:379
#5  free_dive (d=0x55555f9960f0) at /home/bs/src/subsurface/core/dive.c:379
#6  0x0000555555731bd1 in delete_dive_from_table (table=0x555555a95ce0 <dive_table>, idx=idx@entry=0) at /home/bs/src/subsurface/core/divelist.c:743
#7  0x0000555555731ca1 in delete_single_dive (idx=idx@entry=0) at /home/bs/src/subsurface/core/divelist.c:774
#8  0x0000555555732057 in clear_dive_file_data () at /home/bs/src/subsurface/core/divelist.c:1347
#9  0x00005555556f4f25 in DiveTripModelBase::clear (this=0x555555a93fa0 <MobileModels::instance()::self>) at /home/bs/src/subsurface/qt-models/divetripmodel.cpp:475
#10 0x00005555556e6389 in MobileModels::clear (this=<optimized out>) at /home/bs/src/subsurface/qt-models/mobilelistmodel.cpp:942
#11 0x000055555565c93a in QMLManager::openLocalThenRemote (this=0x555555c21c40, url=...) at /home/bs/src/subsurface/mobile-widgets/qmlmanager.cpp:330
#12 0x000055555565dee6 in QMLManager::openNoCloudRepo (this=this@entry=0x555555c21c40) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qrefcount.h:60
#13 0x000055555567e0f2 in QMLManager::qt_static_metacall (_o=0x555555c21c40, _c=<optimized out>, _id=<optimized out>, _a=0x7fffffffac60)
    at /home/bs/src/subsurface/build-mobile/mobile-widgets/subsurface_mobile_autogen/EWIEGA46WW/moc_qmlmanager.cpp:673
#14 0x0000555555680113 in QMLManager::qt_metacall (this=0x555555c21c40, _c=QMetaObject::InvokeMetaMethod, _id=38, _a=0x7fffffffac60)
    at /home/bs/src/subsurface/build-mobile/mobile-widgets/subsurface_mobile_autogen/EWIEGA46WW/moc_qmlmanager.cpp:1088
#15 0x00007ffff6cd34e5 in ?? () from /lib/x86_64-linux-gnu/libQt5Qml.so.5
```

bstoeger:
> Seems reasonably logical and straight forward to me. How can I help you understand this better?

I tried to ask how I can hit the mergeLocalRepo function, though I guess I found it.

dirkhh:
With my local build I can happily switch from a cloud account to a nocloud account.
I do get a ton of scary `DiveSiteSortedModel:allSiteNames invald index` message - but no crash

bstoeger:
> With my local build I can happily switch from a cloud account to a nocloud account.

With master? As I said above: that happens for me on master, not this PR.

> I do get a ton of scary DiveSiteSortedModel:allSiteNames invald index message - but no crash

Clearly, data got inconsistent. That's a bug.

bstoeger:
PS: When I switch to no-cloud all dives disappear. Is that by design? I would have expected that it saves the cloud dives to the local repository?

Also I'm not sure I understand the point of no-cloud mode: Couldn't I just turn auto-sync off and not sync with the cloud for the same effect..?

Just trying to understand, not criticizing - there are basically three repositories:
1) no-cloud
2) cloud local
3) cloud remote

?

dirkhh:
Lost power here for a moment. How timely, my new UPS just arrived...

> PS: When I switch to no-cloud all dives disappear. Is that by design? I would have expected that it saves the cloud dives to the local repository?

No, this is a one way street. There is no point in the other direction. You already have everything in a local repo (it just happens to have a cloud backing).

> Also I'm not sure I understand the point of no-cloud mode: Couldn't I just turn auto-sync off and not sync with the cloud for the same effect..?

People are idiots. I mean, err, users didn't like the idea of having an account connected to their email address. And since we use the email address both as user name and as branch name (btw - that has very odd historical reason and should be changed... but it's kinda too late for that)... we created this special mode that doesn't even attempt to connect with the server.

And, second point... when you enter cloud credentials Subsurface-mobile insists on at least ONE successful connection to the cloud server in order to make sure that your local cache and the cloud backend have a shared base commit to work with.

> Just trying to understand, not criticizing - there are basically three repositories:
> 
> 1. no-cloud

bla/whatever/cloudstorage/localstorage[master]

> 2. cloud local

bla/whatever/cloudstorage/<hexhash.>[your-email-address]

> 3. cloud remote

same as number 2, so the exact same git repository, simply written to a remote server

dirkhh:
So the crash that you posted above happens with master but not with this branch?

That would explain why I wasn't able to reproduce this, ++testing this branch++

dirkhh:
Reading through my epic failure in comprehension again... yeah, this seems to be fixing a bug. At least one. And I cannot find anything that it breaks. So I'll merge it, I guess.

PR #2681, opened 2020-03-17
mobile/edit: don't add dive site twice to table

bstoeger:
When editing a dive on mobile we might have to create a new
dive site. That site is added to the global dive site table
in the undo command. However, the code in QMLManager created
the dive site with create_dive_site*() functions, which already
adds it to the table. The undo command then added the dive
site again leading to a hang of the application.

To solve this problem, create new alloc_dive_site*()
functions that do the same as create_dive_site*()
but do not add it to the table.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a hang when editing a dive site in mobile. Details in the commit-description.

### Changes made:

1) Don't add dive site to global table twice.

### Release note:

No - introduced with divelist changes.

### Mentions:

@dirkhh: should be put in beta quickly, since current master hangs when editing dive site!

### DIFF
--- a/core/divesite.c
+++ b/core/divesite.c
@@ -171,6 +171,21 @@ struct dive_site *alloc_dive_site()
 	return ds;
 }
 
+struct dive_site *alloc_dive_site_with_name(const char *name)
+{
+	struct dive_site *ds = alloc_dive_site();
+	ds->name = copy_string(name);
+	return ds;
+}
+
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
looking good to me... the Mac failure appears to be a GitHub issue - checking out the git tree fails

PR #2683, opened 2020-03-18
Fix divesite-model data inconsistency

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

On mobile the divesite model was not properly reset when resetting the dive list. To avoid this mistake in the future, simply do the reset in the divelist model resetting function.
### Mentions:

@dirkhh: Please check if this fixes the invalid-index warnings. I only gave this a short compile & run test.


dirkhh (approved):
This seems reasonable.
I do think that it might make sense to have a central function that updates all of the models in a safe manner, in the right order, etc. And ensure that it's the only one that can call those update functions to prevent ourselves from getting into more such messes...

tested locally and I can no longer reproduce the invalid index warning

well done.

PR #2685, opened 2020-03-19
Typo: UCT to UTC

mturkia:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [x] Language translation

Fixing misspelled UTC

### DIFF
--- a/desktop-widgets/importgps.ui
+++ b/desktop-widgets/importgps.ui
@@ -305,7 +305,7 @@
               <enum>Qt::LeftToRight</enum>
              </property>
              <property name="text">
-              <string>By how many hours did the international time zone at the dive site differ from UCT (also regarded as Greenwich Mean Time)? Normally one would specify this value in whole hours (h:00):</string>
+              <string>By how many hours did the international time zone at the dive site differ from UTC (also regarded as Greenwich Mean Time)? Normally one would specify this value in whole hours (h:00):</string>
              </property>
              <property name="wordWrap">
               <bool>true</bool>


PeterZaal:
Yes, I noticed the same incorrect spelling in the translations on Transifex. I guess this means this new string has to be pushed to Transifex also. But maybe/probably there are more strings updated elsewhere, so guess strings will be pushed to Transifex anyway shortly before release so translators can get everything updated to the latest.

mturkia:
> Yes, I noticed the same incorrect spelling in the translations on Transifex. I guess this means this new string has to be pushed to Transifex also. But maybe/probably there are more strings updated elsewhere, so guess strings will be pushed to Transifex anyway shortly before release so translators can get everything updated to the latest.

Yep, this is how it is supposed to go. Dirk will need to push the new strings to Transifex, and pull translated strings before release.

dirkhh:
While I know this, I keep stumbling over it...
https://en.wikipedia.org/wiki/Coordinated_Universal_Time#Etymology
Thanks and Ill push updated strings in a moment.
Historically we tended to always have a couple of late changes, usually only very few strings. Why would this release be any different :-)

dirkhh:
The string is pushed - it was the only new string, and since it was trivial to do, I fixed it in all languages where it was already translated (and almost all of them had already used the correct abbreviation, anyway, so all I had to do was to fix the match).
I also fixed a couple of places where we had the same typo in code comments :-)

Issue #2686, opened 2020-03-20
Android: Small UI issues


charno:
Finally found time to test the current mobile beta. I found some small UI nuisances, probably not blockers but probably also very easy to fix.

### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

1)
When switching storage (Cloud -> local, local -> local was tested), the buttons in the divelist are strangely positioned.
![subsurface_buttonbar](https://user-images.githubusercontent.com/2691293/77180254-f0413f00-6ac9-11ea-8fcf-c9714fb53c6a.jpeg)

2)
When a divecomputer is disconnected and reconnected (with the DEVICE_CONNECTED intent) while the download page is open, the connection list is inconsistent. The Connection from the intent is used, but when I click on the device list the old connection (bus/device id) is shown.
![subsurface_connectionlist](https://user-images.githubusercontent.com/2691293/77180747-a73dba80-6aca-11ea-82f4-62d093a14b45.gif)

3)
The save button in the dive editing dialog should probably be white (as it is on the blue background on the divelist)
![grafik](https://user-images.githubusercontent.com/2691293/77180448-38f8f800-6aca-11ea-9304-e7c8f457d3ed.png)

### Operating system:
Android 10 on galaxy s10

### Subsurface version:
3.0.0(4.9.3.1186) from Play store beta

dirkhh:
I have seen the first and have tried to debug it... but haven't been able to figure out _why_ the buttons shift right. Their position is based on the width of the page - so if that was wrong, then the whole page layout should be wrong, but it isn't (at least in my experiments).

The little screen video is a bit confusing to me, I'm not quite sure where it starts and what exactly you are showing there. Am I understanding you correctly that you get an incorrect connection listed? Does the download work?
I have seen a situation where a device was listed twice (once with an incorrect bus address), but tapping on rescan / Neu scannen fixed that in my case. Does it fix the situation for you?

Yes, the incorrectly colored action button is on my "to be fixed before release" list.

Thank you so much for this report - incredibly helpful to see what is not working for our users.

charno:
Hi Dirk,

Downloading does work, as said it's just a little UI annoyance.

Steps:
- Start subsurface with computer already attached to phone
- Go to download screen (btw., for me the only existing entry was not selected...)
- Disconnect the dc
- Connect the dc
- select "Subsurface-mobile" as the app to handle the attached USB device

Now the status is this:
[...]

PR #2687, opened 2020-03-20
Mobile ui fixes

dirkhh:

- [x] Bug fix

This is my attempt to deal with two of the three issues in #2686 -- the first one I still haven't figured out.

### Changes made:

1) simplify the way we pick the icons we use. White works well as foreground color for all of our primary colors. So consistently use that
2) try harder to ensure that we select the correct entry in the connection drop down in the various scenarios regarding when a user plugs in their dive computer
3) Ugggh. And I also added a fix for the completely f-ed up layout of the Settings page. I cannot believe I merged that. What a mess.

### Related issues:

#2686 

### Release note:

not needed - these weren't issues in the last release

### Documentation change:

This will affect the yet to be done screen shots

### Mentions:

@charno

### DIFF
--- a/mobile-widgets/qml/DiveDetails.qml
+++ b/mobile-widgets/qml/DiveDetails.qml
@@ -220,8 +220,8 @@ Kirigami.Page {
 
 	actions.main: Kirigami.Action {
 		icon {
-			name: state !== "view" ? subsurfaceTheme.iconStyle + "/document-save.svg" :
-						 subsurfaceTheme.iconStyle + "/document-edit.svg"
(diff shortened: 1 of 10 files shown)


dirkhh:
pushed two more changes into this branch. One to fix the trip summary edit (it didn't hide the keyboard / close the page after save), the other one is just a long long long overdue fix for our Android logging. This has bugged me for forever...

dirkhh:
not sure who might even want to review this. Oh well... merging my own stuff again, I guess.

PR #2690, opened 2020-03-21
translation updates

dirkhh:

- [x] Bug fix
- [x] Build system change

en_US is our source language - but for the plural handling to work correctly, we need a _translation_ for our strings - and all we need to add there are the strings with plural forms. In the past I did this manually and have consistently gotten things wrong. This hopefully addresses that by using the correct tooling for it.

In the process of testing this I noticed that most of the commands had no working translations. The way the classes were constructed didn't play well with the way the Qt tools work...

### Changes made:

1) special-case the creation of the `subsurface_en_US.ts` file in the (usually commented out) cmake commands
2) update all of those plural forms
3) fully qualify the class used for the `tr()` call for the commands
4) update sources and translations accordingly

### Mentions:

@bstoeger


dirkhh:
repushed to show that this is only two net new commits

dirkhh:
Now I've added a special commit just for @bstoeger  -- except of course he doesn't use de_AT as his locale :-)

dirkhh:
With no comments, I'll just merge :-)

PR #2691, opened 2020-03-21
Revert "core: use set_lang_locale to modify prefs.locale.lang_locale"

bstoeger:
This reverts commit 573a4a5e2da9531fbaa9e82da57131edc691a851.

The commit broke setting the language in the desktop preferences:
Instead of setting the locale in the prefs struct, the locale
is set via qPrefLanguage. However, that saves the default language
(extracted from the system) to disk. Now when the language is
read from the preferences, we get that default value.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This reverts a commit that made setting the language in the preferences ineffective. The original commit says that it fixes an issue on mobile, though is not clear about which issue. Therefore - to be sure - we might conditionally compile this for the moment until we get a proper fix. However that bears the risk of having this conditional compilation forever without anybody knowing why. So I'd prefer to fix the issue right away.

### Release note:

No - bug introduced in release cycle.

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -471,7 +471,7 @@ QString uiLanguage(QLocale *callerLoc)
 	else
 		uiLang = languages[0];
 
-	qPrefLanguage::set_lang_locale(uiLang);
+	prefs.locale.lang_locale = copy_qstring(uiLang);
 
 	// there's a stupid Qt bug on MacOS where uiLanguages doesn't give us the country info
 	if (!uiLang.contains('-') && uiLang != loc.bcp47Name()) {


dirkhh (approved):
since we aren't storing a locale on mobile, I'm happy with your fix.

bstoeger:
I believe(!) the problem that the original commit tried to fix is that if QML accesses the language-property before the uiLanguage() function is run, then it will get the default value and not realize if the value has changed on startup. Thus, different parts of the UI might have different ideas about the language.

The proper fix in that case would be to run uiLanguage() earlier on startup.

However, if you don't see any breakage, let's keep it that way.

PR #2693, opened 2020-03-22
cleanup: don't call uiLanguage() in qPref::loadSync()

bstoeger:
uiLanguage() initializes the language fields from the preferences
values. It is unclear why this function is called in qPref::loadSync()
*before* the fields are loaded from disk. It can only initialize to
the default values anyway. After qPref::loadSync() uiLanguage()
is called again so that everything can be initialized with the
correct perferences values.

Remove the first call. If things break, let's fix them in a sensible
way.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

There appears to be no reason to initialize the language system twice at startup..?

### Changes made:

1) Remove apparently superfluous call to `uiLanguage()`

### Release note:

Not needed.

### Documentation change:

Not needed.

### Mentions:

@dirkhh: I ran desktop and mobile and everything *appeared* to be correct.


bstoeger:
A few more cleanups making the code hopefully more easy to follow. I'm not sure about the MacOS thing - perhaps I am missing something subtle?

dirkhh (approved):
This all looks sane.
Let's see what it breaks in real life.

PR #2694, opened 2020-03-22
Coverity fixes

dirkhh:

- [x] Code cleanup

marked as code cleanup instead of fixes as all of these are theoretical, imho.
Yes, the dive table could be completely corrupt, but then everything else will blow up as well.
There were a few other new coverity reports about similar things that were just too obviously correct to even deal with.
Static analysis is great, but the false positives can be tedious.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -464,7 +464,7 @@ int init_decompression(struct deco_state *ds, struct dive *dive)
 		struct dive *pdive = get_dive(i);
 		/* we don't want to mix dives from different trips as we keep looking
 		 * for how far back we need to go */
-		if (dive->divetrip && pdive->divetrip != dive->divetrip) {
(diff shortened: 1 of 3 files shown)


bstoeger:
No. 1 is a real bug. No. 2 and 3 make me sad.

dirkhh:
I was torn whether I should address 2&3. There are half a dozen more along those lines that were even more blatantly stupid. Maybe I should drop 2&3 as well.

glance-:
I'm resonably shure that we can configure coverity to "know" that get_dive() always returns a valid pointer.

bstoeger:
Careful - it doesn't. We use the idiom "increase index until get_dive() returns NULL" at some points.

It's probably not worth losing too much time over this.

dirkhh:
I decided to cherry pick the actual fix and will for now ignore the rest.

PR #2696, opened 2020-03-23
mobile/profile: don't pan unless we are zoomed in

dirkhh:
This way someone trying to swipe from dive to dive won't inadvertantly
pan the profile instead. And panning it really only makes sense when
zoomed in in the first place.

This could leave us in a situation where we zoom in, pan, zoom out and
now the profile is out of whack and we cannot correct it. A simple click
on the profile fixes that.

The real solution would be some constraining / adjusting as we zoom and
pan to ensure we keep things correctly positioned. Maybe I'll figure out
the correct way to do this later...

- [X] Bug fix

Truly dreadful and disgusting hack around the problem @mturkia described.
I hate so much about this - and of course this still doesn't deal with the fact that you can pan the profile around unconstrained. I tried to get the math to work so that the profile wouldn't pan past it's own edges, but there is so much black magic in the render function, and such a weird interaction between the scaling and the transformation in the render function that I gave up on getting this right... instead I did this hack to address the larger issue that one wasn't able to swipe from dive to dive

### Release note:

no, this is a net new feature

### Mentions:

@mturkia

### DIFF
--- a/mobile-widgets/qml/DiveDetailsView.qml
+++ b/mobile-widgets/qml/DiveDetailsView.qml
@@ -256,19 +256,35 @@ Item {
 					}
 
 					MouseArea {
-						anchors.fill: parent
-						drag.target: qmlProfile
-						drag.axis: Drag.XAndYAxis
-						drag.smoothed: true
-						pressAndHoldInterval: 50 // very short - feels about right
+						// we want to pan the profile if we are zoomed in, but we want to immediately
+						// pass the mouse events through to the ListView if we are not. That way you
+						// can swipe through the dive list, even if you happen to swipe the profile
+						property bool isZoomed: qmlProfile.scale - 1.0 > 0.02
+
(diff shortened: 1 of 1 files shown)


dirkhh:
Since typically no one even looks at the QML only changes, let's merge and see if this gets feedback from the beta testers

mturkia:
This is definitely a lot better. Looking at the code didn't really do any good, but according to my testing, it seems to work as described. Would it be possible to centre the dive profile automatically when it is zoomed out? Clicking on the profile did not reset the position for me after zooming out. Testing on iPhone, if that matters.

dirkhh:
What I really really want, is to figure out the math so that as we zoom / pan, we never show the blank space around the profile. The math should be simple, but when I played with it yesterday, with so many different pieces working together, I got stuck...
I'll give it another go.

PR #2697, opened 2020-03-24
divelist: don't initialize invalidFont at startup

bstoeger:
To mark invalid dives, we use a struck out font, which was a static
variable at translation unit scope, i.e. initialized at application
startup. Sadly, this crashes on iOS.

It is unclear when we can initialize fonts. Try to move initialization
to the constructore of DiveTripModelBase and make the font a member
of that class. For consistency, also make the invalidBrush a member
of this class.

This now means that the diveData function cannot be static anymore,
since it needs access to the font and brush. But OK.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This attempts to fix a startup-crash on iOS. Apparently you can't initialize fonts on application startup, so defer to when the model is created. Let's hope that this is "late enough".

### Mentions:

@dirkhh

### DIFF
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -190,16 +190,7 @@ static QString displayWeight(const struct dive *d, bool units)
 		return s + gettextFromC::tr("lbs");
 }
 
-static QFont struckOutFont()
-{
-	QFont font;
-	font.setStrikeOut(true);
-	return font;
-}
-static QBrush invalidForeground(Qt::gray);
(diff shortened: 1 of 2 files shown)


dirkhh (commented):
This looks logical, I'll build an iOS version and see if it works :-)

dirkhh (approved):
Tested and no crash
Thank you so much for fixing this right away...

PR #2699, opened 2020-03-24
Better pan zoom

dirkhh:

- [x] Functional change

As @mturkia suggested, we shouldn't allow the user to pan/zoom the profile in a way that shows the empty space outside of the profile.
This took an embarrassing number of attempts to get right, but in the end the code is really clean and simple, I think.

### Changes made:
1) allow simulation of zoom on the desktop (so much easier for debugging)
2) implement upper and lower constraints for the x and y offsets
3) ensure that we enforce those constraints when zooming

### Release note:

none, this is improving a features that hasn't been released, yet

### Documentation change:

I don't think that's needed. This seems fairly intuitive

### Mentions:

@mturkia 
@bstoeger -- just because I want to show off using std::min and std::max for clamping the value... (and I even looked up that this will be available as std::clamp in C++17 which we aren't quite ready for, yet)

### DIFF
--- a/profile-widget/qmlprofile.cpp
+++ b/profile-widget/qmlprofile.cpp
@@ -24,7 +24,18 @@ QMLProfile::QMLProfile(QQuickItem *parent) :
 	m_profileWidget->setPrintMode(true);
 	m_profileWidget->setFontPrintScale(fontScale);
 	connect(QMLManager::instance(), &QMLManager::sendScreenChanged, this, &QMLProfile::screenChanged);
+	connect(this, &QMLProfile::scaleChanged, this, &QMLProfile::triggerUpdate);
 	setDevicePixelRatio(QMLManager::instance()->lastDevicePixelRatio());
+}
+
+// we need this so we can connect update() to the scaleChanged() signal - which the connect above cannot do
+// directly as it chokes on the default parameter for update().
+// If the scale changes we may need to change our offsets to ensure that we still only show a subset of
+// the profile and not empty space around it, which the paint() method below will take care of, which will
+// eventually get called after we call update()
+void QMLProfile::triggerUpdate()
+{
+	update();
 }
 
 void QMLProfile::paint(QPainter *painter)
--- a/profile-widget/qmlprofile.cpp
+++ b/profile-widget/qmlprofile.cpp
@@ -62,6 +62,25 @@ void QMLProfile::paint(QPainter *painter)
 	// shift the painter (taking its existing transformation into account)
 	QTransform profileTransform = QTransform();
 	profileTransform.scale(sx, sy);
+
+	// this is a bit complex because of the interaction with the Qt widget scaling
+	// logic. We use the default setup where the scale happens with the center of the
+	// widget as transformation center.
+	// The QML code *should* already ensure that we don't shrink the profile, but for
+	// the math below to work, let's make sure
+	qreal profileScale = scale();
+	if (profileScale < 1.0) {
+		profileScale = 1.0;
+		setScale(profileScale);
+	}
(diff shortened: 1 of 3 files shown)


profile-widget/qmlprofile.cpp
```diff
@@ -24,9 +24,18 @@ QMLProfile::QMLProfile(QQuickItem *parent) :
...
+	connect(this, &QMLProfile::scaleChanged, this, &QMLProfile::triggerUpdate);
 	setDevicePixelRatio(QMLManager::instance()->lastDevicePixelRatio());
 }
 
+// we need this so we can call update() (which eventually redraws the profile and calls paint()
+// from the scaleChanged() signal -- this way as we scale out we can ensure that we shift the offsets
+// to not show "empty space" outside the profile when panning
+void QMLProfile::triggerUpdate()
+{
+	emit update();
```
bstoeger: According to the docs, update is not a signal and therefore the `emit` is conceptually wrong here:
https://doc.qt.io/qt-5/qquickpainteditem.html

Practically, I believe `emit` is simply replaced by the empty string. However, some static checkers may complain.

profile-widget/qmlprofile.cpp
```diff
@@ -62,6 +71,18 @@ void QMLProfile::paint(QPainter *painter)
...
 	QTransform profileTransform = QTransform();
 	profileTransform.scale(sx, sy);
+
+	// this is a bit complex because of the interaction with the Qt widget scaling
+	// logic. We use the default setup where the scale happens with the center of the
+	// widget as transformation center.
+	// When zooming and panning we want to ensure that we always show a subset of the
+	// profile and not the "empty space" around the profile.
+	// a bit of math on a piece of paper shos that our offsets need to stay within +/- dx and dy
+	qreal dx = width() * (scale() - 1) / (2 * dpr * scale());
```
bstoeger: I reckon that `scale` is limited to `>=1` (i.e. only zoom in, no zoom out), otherwise `dx` and `dy` become negative and the clamping below will become funny.

profile-widget/qmlprofile.cpp
```diff
@@ -62,6 +71,18 @@ void QMLProfile::paint(QPainter *painter)
 	// shift the painter (taking its existing transformation into account)
 	QTransform profileTransform = QTransform();
 	profileTransform.scale(sx, sy);
+
+	// this is a bit complex because of the interaction with the Qt widget scaling
+	// logic. We use the default setup where the scale happens with the center of the
+	// widget as transformation center.
+	// When zooming and panning we want to ensure that we always show a subset of the
+	// profile and not the "empty space" around the profile.
+	// a bit of math on a piece of paper shos that our offsets need to stay within +/- dx and dy
```
bstoeger: `shos` -> `shows`

bstoeger (approved):
C++-part fine with me - only minor comments.

profile-widget/qmlprofile.cpp
dirkhh: oops - I can't even tell why I was confused about this... you are right.

profile-widget/qmlprofile.cpp
dirkhh: Correct, but that's done in QML, so it might make sense to enforce that here as well. I'll think of a good solution for that.

profile-widget/qmlprofile.cpp
dirkhh: thanks

dirkhh:
Very useful review - much appreciated. I'll have a new version shortly.

dirkhh:
I believe I have addressed all three comments. Waiting for CI and then I'll merge

PR #2702, opened 2020-03-27
Properly update fulltext filter when editing a dive

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is supposed to fix a bug reported on the ML. When editing a dive, the fulltext filter must be updated. The desktop versions do this, but not the mobile version.

I'm not 100% convinced of the resolution, but already too tired to think about this.

### Changes made:

1) Update hidden_by_filter flag
2) Update fulltext filter
### Mentions:

@dirkhh

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1248,8 +1248,11 @@ void EditDive::exchangeDives()
 	// it by swapping the dive data.
 	newDive->hidden_by_filter = oldDive->hidden_by_filter;
 
-	// Bluntly exchange dive data by shallow copy
+	// Bluntly exchange dive data by shallow copy.
+	// Don't forget to unregister the old and register the new dive!
+	fulltext_unregister(oldDive);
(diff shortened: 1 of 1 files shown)


dirkhh:
Question: you say that the desktop version "does this"...
So should the corresponding code in the desktop version be removed since this is now done in the command infrastructure?

bstoeger:
No - desktop and mobile use different commands for editing dives. The former are much more fine-grained.

What I meant was that the desktop version does this already - no change needed.

bstoeger:
To be more clear: when I said the "desktop versions do this", I meant "the desktop versions [of the undo commands] do this".
See the `fulltext_register()` calls in `EditTagsBase::undo()` and `EditBase<T>::undo()`. This PR adds the code to the coarse `EditDive` mobile command.

Hope that explains it.

PR #2703, opened 2020-03-28
Save fix

dirkhh:

- [x] Bug fix

CRUD.
In local sync only mode we can get into a state where `alreadySaving` is `true` and never gets reset. As a result no changes that the user makes are ever saved :-(

This significantly redesigns the use of that flag to make it much tighter, just around the actual potential storage access.

### Release note:

still need to add those

### Mentions:

@bstoeger  I would REALLY appreciate your review - trying to release this within an hour if possible as that could really cause trouble for users.

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -1272,12 +1272,12 @@ void QMLManager::changesNeedSaving()
 }
 
 void QMLManager::openNoCloudRepo()
-/*
- * Open the No Cloud repo. In case this repo does not (yet)
- * exists, create one first. When done, open the repo, which
- * is obviously empty when just created.
- */
 {
+	/*
+	 * Open the No Cloud repo. In case this repo does not (yet)
+	 * exists, create one first. When done, open the repo, which
+	 * is obviously empty when just created.
+	 */
 	QString filename = nocloud_localstorage();
 	const char *branch;
 	struct git_repository *git;
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -305,13 +305,22 @@ void QMLManager::applicationStateChanged(Qt::ApplicationState state)
 	stateText.append((unsaved_changes() ? QLatin1String("") : QLatin1String("no ")) + QLatin1String("unsaved changes"));
 	appendTextToLog(stateText);
 
-	if (!alreadySaving && state == Qt::ApplicationInactive && unsaved_changes()) {
-		// FIXME
-		//       make sure the user sees that we are saving data if they come back
-		//       while this is running
+	if (state == Qt::ApplicationInactive && unsaved_changes()) {
+		// saveChangesCloud ensures that we don't have two conflicting saves going on
+		appendTextToLog("trying to save data as user switched away from app");
 		saveChangesCloud(false);
-		appendTextToLog("done saving to git local / remote");
+		appendTextToLog("done trying to save to git local / remote");
 	}
+}
+
+int QMLManager::openAndMaybeSync(const char *filename)
+{
+	// parse_file will potentially sync with the git server - so make sure we don't start
+	// a save in the middle of that (for example if the user switches away from the app)
+	alreadySaving = true;
+	int error = parse_file(filename, &dive_table, &trip_table, &dive_site_table);
+	alreadySaving = false;
+	return error;
 }
 
 void QMLManager::openLocalThenRemote(QString url)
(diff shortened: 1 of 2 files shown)


bstoeger:
These changes  make no sense to me, sorry. But perhaps I'm simply missing something.

The `alreadySaving` flag is only tested in `QMLManager::saveChangesLocal()` and `QMLManager::saveChangesCloud()`. Therefore constructs like
```
        alreadySaving = true;
        int error = parse_file(filename, &dive_table, &trip_table, &dive_site_table);
        alreadySaving = false;
```
or
```
        alreadySaving = true;
        if (git != dummy_git_repository) {
                appendTextToLog(QString("have repository and branch %1").arg(branch));
                error = git_load_dives(git, branch, &dive_table, &trip_table, &dive_site_table);
        } else {
                appendTextToLog(QString("didn't receive valid git repo, try again"));
                error = parse_file(fileNamePrt.data(), &dive_table, &trip_table, &dive_site_table);
        }
        alreadySaving = false;
```
look like no-ops to me, assuming that there is no callback from the git-loader to QMLManager.

You are right though that the handling of `alreadySaving` appears fishy. The code would be much more easy to follow if the flag was set and cleared in the same function. Though I guess you want to move it to a higher, not to a lower level.

For example, the code in `finishSetup()` should perhaps read
```
                alreadySaving = true;
                openLocalThenRemote(url);
                alreadySaving = false;
```
???
My fundamental problem is that I really don't know what this flag protects against.

dirkhh:
The thing that you are missing is that if the user switches away from the app, we get notified in a separate thread: `QMLManager::applicationStateChanged()`
And there we check if there are unsaved changes and call `saveChangesCloud()`.
And that's what all this is supposed to protect against.
Not our main thread doing things twice, but that thread (which only happens when the user switches away) doing a conflicting save.
History here is that if you initiated a save and switched away you could reliably corrupt your local cache.

dirkhh:
The reason I moved this LOWER, not higher, is that I want to only protect the actual storage access - and I want to minimize the risk that we somehow end up in a state where you set the flag and then exit a function without clearing it.

bstoeger:
> The thing that you are missing is that if the user switches away from the app, we get notified in a separate thread: QMLManager::applicationStateChanged()

If that is the *only* reason then we simply shouldn't call `saveChangesCloud()` from there, but instead send a signal to the `QMLManager` so that the UI thread picks up from there.

To my knowledge, `connect` defaults to `auto-connection`, which means that it will queue the signal if sent from a different thread. This is one of Qt's idiosyncratic IPC mechanisms.

Having the saving run from different threads appears very risky to me, in more than one way. The dive data could change right under our feet.

> The reason I moved this LOWER, not higher, is that I want to only protect the actual storage access - and I want to minimize the risk that we somehow end up in a state where you set the flag and then exit a function without clearing it.

That's gambling on the fact that the saving spends most of the time in the IO routines. You can still create two threads that will save at the same time leading to weird results.

dirkhh:
When trying to call things in the main thread, I had that fail when the user switches away. The UI Thread apparently doesn't get rescheduled if it isn't already active doing something.

I'm not sure I understand your last comment.

dirkhh:
My issue right now is that 3.0.0 is broken and I can reproduce the issue with it and it prevents any changes the user makes from being saved.
This may not be the correct fix (and I'd love to find a better fix), but it (a) fixes the problem and (b) appears to be incrementally better compared to what we did before.
Given that I know that we have only one UI thread and potentially more than one threads being triggered by the user switching between apps. I _believe_ that this is save, but would of course appreciate any scenario that might break things.
I just pushed a 3.0.1 beta for Android (and am waiting for iOS to allow me to do so - which can take a day or three sometimes). Thankfully not a lot of people are on dive trips right now so I don't think we are causing a lot of data loss for people, but I'd still prefer to have a version that fixes the bug that people are seeing right now with the released version.

bstoeger:
> My issue right now is that 3.0.0 is broken and I can reproduce the issue with it and it prevents any changes the user makes from being saved.

If this whole thing is reproducible on mobile-on-desktop I can try to have a look. However, not today.

> When trying to call things in the main thread, I had that fail when the user switches away. The UI Thread apparently doesn't get rescheduled if it isn't already active doing something.

That's weird. To my understanding, since `applicationStateChange()` is called via a signal, it should run from QMLManager's thread. So I'm still not sure that I understand the reason of the flag.

> I'm not sure I understand your last comment.

The code gambles on the fact that when the test for `alreadySaving` is made, the concurrent thread is inside one of the `alreadySaving=true,false` sections. It's not a bad gamble, as most of the time the thread will be doing IO, but certainly not guaranteed.

dirkhh:
> If this whole thing is reproducible on mobile-on-desktop I can try to have a look. However, not today.

Yes, the overall behavior is the same when run on the DT. Switch away from the app, this function is called.
 
> > When trying to call things in the main thread, I had that fail when the user switches away. The UI Thread apparently doesn't get rescheduled if it isn't already active doing something.
> 
> That's weird. To my understanding, since `applicationStateChange()` is called via a signal, it should run from QMLManager's thread. So I'm still not sure that I understand the reason of the flag.

That sounds logical and I don't really have a good answer right now. It's been years since I battled this particular stupidity.
 
> > I'm not sure I understand your last comment.
> 
> The code gambles on the fact that when the test for `alreadySaving` is made, the concurrent thread is inside one of the `alreadySaving=true,false` sections. It's not a bad gamble, as most of the time the thread will be doing IO, but certainly not guaranteed.

The logic I used was this: the user initiates some action, then switches away. The time this takes to initiate the switch should mean that the app is either idle or is in the middle of saving something.
You are right, there is the chance that the user manages to switch away BEFORE we set the alreadySaving flag. And a way around this would be to actually implement a real lock that is used everywhere.

I guess I'll tackle that one next.

PR #2704, opened 2020-03-29
Mutex

dirkhh:

- [x] Code cleanup

Realistically this is actually a bug fix.
The old code relied on the fact that it takes the user a moment to switch away from the app, and that therefore the main thread would have already locked the critical section before we try to save once notified that the user switched away.

### Changes made:

1) replace the `bool` with a `QMutex`
2) clean up some of the use cases to make the code easier to read
3) the log message when receiving the 'switched away' signal no longer can tell if a save is ongoing (I could of course have added an additional bool for that but that seemed silly)

### Mentions:

@bstoeger -- I think that this is a much cleaner solution. Or maybe I'm just on drugs.

This is much easier to look at ignoring whitespace


bstoeger:
That makes more sense to me. I still wonder if the lock shouldn't be held at the entry points to this functionality. To give a more in-depth opinion I will have to create a call-graph. This is all still very opaque to me.

dirkhh:
Yes, of course. We could hold the mutexes higher up in the call graph. This is a rather straight forward adaption. Definitely room for more improvement. What makes this one attractive to me is that it is evident that all calls that access storage are indeed protected. The code is such a horrible, disgusting mess that I am less convinced that I could ensure the same higher up in the stack.

Would you suggest that we merge this as is (as a clear improvement), or should we try to push this up higher in the stack before merging?

bstoeger:
If you are positive that this cannot deadlock, go ahead. When I find time, I will try to generate a call-graph.

dirkhh:
I don't see how the current system could deadlock. But then again, I'm an idiot.
Speaking of which, looking at the call graph is entirely at your own risk - descending into insanity seems like a likely outcome.
Also, keep in mind that the QML code calls into QMLManager with `manager.method()` calls - there are a couple of methods in the class that are ONLY called from QML which can be confusing when trying to figure out what's going on.
And, of course, we get notifications from the OS (e.g. when switching away, or when the user plugs in a device on Android) :-)

PR #2707, opened 2020-03-30
Minor mobile cleanups

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are just a few code-hygiene things.

The idea of the first is to make the QMLManager interface more clear. It *could* break things, since QML is not compiled and some parts of the code might use the archaic runtime connect-syntax. So please check.

The other commits fix warnings that annoyed me for the longest time.

### Changes made:

1) Unslotify functions that are neither used in signals nor called from QML.
2) Fix compiler warnings.

### Mentions:

@dirkhh: please check, especially the first commit.


dirkhh (approved):
a couple of those were slots because I thought I might need to be able to call them from QML

PR #2708, opened 2020-03-30
Cleanups in themeinterface.cpp

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Some minor code cleanups concerning the ThemeInterface class. All over the place from coding style to not exporting local constants.

Not much to say - more detailed info is in the commit messages.

### DIFF
--- a/mobile-widgets/themeinterface.cpp
+++ b/mobile-widgets/themeinterface.cpp
@@ -5,44 +5,44 @@
 #include "core/settings/qPrefDisplay.h"
 #include <QFontInfo>
 
-const QColor BLUE_BACKGROUND_COLOR = "#eff0f1";
-const QColor BLUE_CONTRAST_ACCENT_COLOR = "#FF5722";
-const QColor BLUE_DARKER_PRIMARY_COLOR = "#303F9f";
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
This all looks great.
I'm 99% sure that the change regarding constructor / setup / when things are called are correct.
On the flip side, I have many scars caused by timing issues around types that QML decides it needs to create and that then trying to use values that aren't initialized and other such nonsense.
As I said, this looks sane and correct to me. But it needs careful testing on actual devices to make sure that this isn't one of those stupid QML gotchas.

bstoeger:
> On the flip side, I have many scars caused by timing issues around types that QML decides it needs to create and that then trying to use values that aren't initialized and other such nonsense.

That shouldn't be an issue here, as the order of calls remain be the same.

Only in the newer commits, which I just pushed, the order-of-initialization is slightly changed. However, we never hand over control to QML, just register global objects, so I hope that it doesn't matter (*knock wood*).

dirkhh:
I assume you refer to 2a50f95 - that one had already been pushed when I posted my review.
I'm happy to merge and see what happens. Worst case it breaks the current betas.

PR #2711, opened 2020-03-31
cleanup: make QMLManager member functions private

bstoeger:
The
	- mergeLocalRepo()
	- openLocalThenRemove()
functions were not invoked from outside QMLManager.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Weird - why did I miss those yesterday...?

### DIFF
--- a/mobile-widgets/qmlmanager.h
+++ b/mobile-widgets/qmlmanager.h
@@ -215,8 +215,6 @@ public slots:
 	void copyAppLogToClipboard();
 	bool createSupportEmail();
 	void finishSetup();
-	void openLocalThenRemote(QString url);
-	void mergeLocalRepo();
 	QString getNumber(const QString& diveId);
 	QString getDate(const QString& diveId);
 	QString getCurrentPosition();
@@ -279,6 +277,8 @@ private:
 	void loadDivesWithValidCredentials();
 	void revertToNoCloudIfNeeded();
 	void consumeFinishedLoad();
+	void mergeLocalRepo();
+	void openLocalThenRemote(QString url);
 
 #if defined(Q_OS_ANDROID) || defined(Q_OS_IOS)
 	QString appLogFileName;


dirkhh (changes requested):
obviously correct - except for the commit message that mentions openLocalThenRemove instead of openLocalThenRemote :-)

bstoeger:
Yes, duh.

With this PR I get the following call graph, whereby I omitted leaf functions that do not access the git-store (notably `verifyCredentials`, `revertToNoCloudIfNeeded` and `consumeFinishedLoad`). These omitted functions need to be protected as well, though (e.g. `consumeFinishedLoad` resets the core data structures).

However, before doing so, I'd like to understand why that problem exists in the first place. To me it looks like all the accesses in this call-graph should come from the UI thread. I will investigate when I find time.

Disclaimer: the call graph is hand-made, it probably contains errors.
call

dirkhh:
My understanding (and way, WAY back when I first hacked this pile of poop together, I clearly remember that I DID encounter corruption) is that the `applicationStateChanged` signal can come in completely asynchronously, even in the middle of you doing I/O. Specifically, I had a back end issue that would cause all cloud accesses to be serialized, which meant your cloud save would sometimes take minutes. I switched away from Subsurface-mobile to check my email, and a second save happened and caused my git repo to be corrupted. Now granted, this was likely with Android 4.x - we are talking a while ago. But realistically this should be somewhat easy to reproduce / verify...

dirkhh (approved):

bstoeger:
> I switched away from Subsurface-mobile to check my email, and a second save happened and caused my git repo to be corrupted. Now granted, this was likely with Android 4.x - we are talking a while ago. But realistically this should be somewhat easy to reproduce / verify...

I can't reproduce this:
```
bs@laptop:~/src/subsurface/build-mobile$ ./subsurface-mobile 2>/dev/null 
        ------ QMLManager::openLocalThenRemote(): 0x7fd523e455c0
        ------ QMLManager::openAndMaybeSync(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::changesNeedSaving(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::changesNeedSaving(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
        ------ QMLManager::changesNeedSaving(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
        ------ QMLManager::changesNeedSaving(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
        ------ QMLManager::loadDivesWithValidCredentials(): 0x7fd523e455c0
        ------ QMLManager::applicationStateChanged(): 0x7fd523e455c0
        ------ QMLManager::saveChangesLocal(): 0x7fd523e455c0
```
It shouldn't matter whether it's old Android or not. Since the `applicationStateChanged` comes in via an auto-signal, it must run in the UI-thread. All things QML should also come via the UI-thread. I don't see where we would call these functions from a different thread? Do you know where we schedule such things in the background?

dirkhh:
excellent. let's rip the whole shit out. Yay.

bstoeger:
> excellent. let's rip the whole shit out. Yay.

Careful. Just because *I* couldn't reproduce it, doesn't mean that it never happens.

dirkhh:
Correct. But right now we have users not able to save data they download from a dive computer.
See #2718 
Which I was able to reproduce and from the log it seems that once again we have the storage locked. I don't understand, yet, why we ended up in a state where it is locked, but that's what I see in the log file. So right now I'm leaning towards removing that lock, TBH

PR #2712, opened 2020-03-31
[RFC] mobile: don't call saveChangesLocal() twice on non-iOS

bstoeger:
QMLManager::changesNeedSaving() behaves differently on iOS:
it only saves locally with saveChangesLocal(), whereas all
other OS save to cloud with saveChangesCloud(). Nevertheless,
even for other OS saveChangesLocal() is called even though
that will be called in saveChancesCloud anyway. Therefore,
compile the saveChangesLocal() call in changesNeedSaving
conditionally.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Unless I'm missing something, there appears to be no reason to call saveChangesLocal() twice on non-iOS?

### Changes made:

1) Conditionally compile saveChangesLocal() - this makes the code a bit more logical to me, YMMV.

The call-graph then becomes:
call2

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -1267,8 +1267,9 @@ void QMLManager::changesNeedSaving()
 	// on iOS
 	// on all other platforms we just save the changes and be done with it
 	mark_divelist_changed(true);
+#if defined(Q_OS_IOS)
 	saveChangesLocal();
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
looks right

dirkhh:
The call graph is... wow.

PR #2713, opened 2020-03-31
[TEST] cleanup: don't compile subsurface-helper.cpp for tests

bstoeger:
It is unclear why the subsurface-helper.cpp file was linked with the
tests. In any case, it led to comlicated ifdef-ery, so let's remove
that for now and readd later if the need arises.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is mostly for CI. I don't understand why the subsurface-helper.cpp file was compiled for the tests, since most of the file was ifdef-ed out anyway. Perhaps CI will tell me.

### DIFF
--- a/subsurface-helper.cpp
+++ b/subsurface-helper.cpp
@@ -33,7 +33,6 @@ QString getAndroidHWInfo(); // from android.cpp
 #include <QFontDatabase>
 #endif /* Q_OS_ANDROID */
 
-#ifndef SUBSURFACE_TEST_DATA
 QObject *qqWindowObject = NULL;
 
 // Forward declaration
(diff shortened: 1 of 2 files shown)


bstoeger:
Is the TestQML thing even compiled / run by CI?

dirkhh:
> Is the TestQML thing even compiled / run by CI?

One option to find out would be to look at the log of the Mac or either the Ubuntu 18.04 or 19.10 builds, as those contain a call to `make check` in their `.yml`. And it is very clear that `TestQML` is both built and run on those.

dirkhh (approved):
I think those tests are mostly bogus and have never really been all that useful.
I do not believe that they really test anything that is relevant to what we do. And I don't recall them ever failing in a manner that pointed to a bug in our code - instead the times they are failing are when we make changes in our code and forget to update the tests... which makes me think that they are useless garbage.

PR #2717, opened 2020-04-01
Next round mobile fixes

dirkhh:

- [x] Bug fix
- [x] Functional change
- [x] New feature

This maybe should be the first step towards a 3.1 release, I guess, as some of these changes are a bit more invasive - but for now it calls itself 3.0.3

### Changes made:

1) cleanup all the copyright years to 2020
2) add yet another notification mechanism to update the user on progress early during startup. Yes, this feels redundant, but it is really a game changer on a slowish device or with a massively huge dive log. This feels so much more responsive to me.
3) remove the locking of git storage - testing seems to indicate that all access comes from the same thread, so the locking is unnecessary and indeed harmful as it creates situations where users can't save their changes to storage
4) dive list performance improvements

### Release note:

done

### Documentation change:

not needed

### Mentions:

@bstoeger updating the comment here since the performance issue that I first listed here has in fact been resolved.


bstoeger:
> 13 seconds (!!) for Qt to finish setting up the model

Have you profiled it? Perhaps numerous map reloads as we had at one point?

I have a 3500 dives file and it starts in a few s on my slow laptop (mobile on desktop starts nearly instantaneously). Although that has only very few trips, so that might be the culprit.

dirkhh:
I have not profiled it. And I haven't checked if there are call-backs into the rest of our code. But as I tried to indicate above, we spend those 13 seconds on `endResetModel();`

dirkhh:
Here are some timing data from my 32bit phone (old Sony Xperia):
```
INFO: "4.036: showProgress: Open local dive data file"
INFO: "9.616: showProgress: Successfully opened dive data"
INFO: "9.696: showProgress: Create full text index"
INFO: "10.297: showProgress: \r608 dives processed"
INFO: "10.309: showProgress: populate data model"
INFO: "10.916: showProgress: \r608 dives processed"
INFO: "11.073: showProgress: finish populating data store"
INFO: "12.746: showProgress: setting up internal data structures"
INFO: "13.241: showProgress: done setting up internal data structures"
INFO: "14.557: showProgress: 608 dives loaded from local dive data file"
```
this is with my normal dive file (actually, an older version of it I just realize). With only 600 dives the times look much more balanced. 4 seconds until we finally start opening the local repo, 5.6 seconds to load the git repo, half a second for the full text index, half a second to feed the data into the model, and only 1.7 seconds spent on the `endResetModel()` call...

I'll try the huge data file during my next work break :-)

core/qthelper.cpp
```diff
@@ -1642,6 +1642,15 @@ char *copy_qstring(const QString &s)
 	return strdup(qPrintable(s));
 }
 
+// function to call to allow the UI to show updates for longer running activities
+void (*uiNotificationCallback)(QString msg) = nullptr;
+
+void uiNotification(const QString &msg)
+{
+	if (uiNotificationCallback != nullptr)
+		uiNotificationCallback(QString(msg));
```
bstoeger: Casting QString to QString here. Seems redundant.

mobile-widgets/qmlmanager.cpp
```diff
@@ -108,6 +109,24 @@ extern "C" int gitProgressCB(const char *text)
 	return 0;
 }
 
+// this callback is used from the uiNotification() function
+// the detour via callback allows us to keep the core code independent from QMLManager
```
bstoeger, line 96: Not criticizing the code - but that comment is not very convincing. You could achieve the same by defining two different uiNotification functions on desktop and mobile:
```
#ifdef SUBSURFACE_MOBILE
inline void uiNotification(const QString &msg)
{
        showProgress(msg)
}
#else
inline void uiNotification(const QString &msg)
{
}
#endif
```
This way, the compiler *might* remove the whole call on desktio. I believe it won't though because it can't know whether creation/destruction of the QString has side-effects. If you want it to be zero-cost on desktop you would have to define it as a macro.

The callback is needed if you want to change the recipient at runtime. E.g. if such messages might be sent when QMLManager is not yet ready.

Of course, the classical Qt way for that particular use-case would be sending a signal via a class.

mobile-widgets/qmlmanager.cpp
```diff
@@ -108,6 +109,24 @@ extern "C" int gitProgressCB(const char *text)
...
+// I'm not sure it makes sense to have three different progress callbacks,
+// but the usage models (and the situations in the program flow where they are used)
+// are really vastly different...
+// this is mainly intended for the early stages of the app so the user sees that
+// things are progressing
+extern void showProgress(QString msg)
+{
+	static QMLManager *self;
```
bstoeger: There appears to be no reason to make this a static variable. This function should probably read as
```
        QMLManager *self = QMLManager::instance();
        if (self)
                self->setNotificationText(msg);
```

mobile-widgets/qmlmanager.cpp
```diff
@@ -108,6 +109,24 @@ extern "C" int gitProgressCB(const char *text)
...
 }
 
+// this callback is used from the uiNotification() function
+// the detour via callback allows us to keep the core code independent from QMLManager
+// I'm not sure it makes sense to have three different progress callbacks,
+// but the usage models (and the situations in the program flow where they are used)
+// are really vastly different...
+// this is mainly intended for the early stages of the app so the user sees that
+// things are progressing
+extern void showProgress(QString msg)
```
bstoeger: This should read `static`, no?

mobile-widgets/qmlmanager.cpp
```diff
@@ -518,6 +541,8 @@ void QMLManager::finishSetup()
 		appendTextToLog(tr("no cloud credentials"));
 		setStartPageText(RED_FONT + tr("Please enter valid cloud credentials.") + END_FONT);
 	}
+	m_initialized = true;
+	emit initializedChanged();
```
bstoeger, line 528: Does Qt really not provide a setProperty macro that sets the property and sends the signal? That feels so lame!

mobile-widgets/qmlmanager.cpp
dirkhh: I'm not worried about the overhead of those calls. Yes, you are 100% correct that I could do all kinds of wonderful things to avoid a few hundred cycles. That is NOT the problem that needs addressing.
I'm happy to rip out the whole notification garbage that is in there. All of it. From everywhere. And create a reasonable way to deal with that and distributing the notifications to the right recipients. Because what we have today is a set of horrible hacks, implemented by different people at different times trying to address exactly their one problem at that time - just like this PR is.

What we are doing today is broken. The user on the mobile app gets no useful feedback and stares at what looks like a hung app for 15 seconds for a fairly modestly sized dive list or 2:30 minutes for the 8000 dive data set (with 98 seconds spent on that `endResetModel()` call).

Please understand that I am NOT saying that there is anything wrong with your code. I am trying to address a serious issue for people actually using the mobile app. I have received half a dozen support requests already where people believe the app has died when in fact it's just processing data before showing the UI.

core/qthelper.cpp
dirkhh: that's because initially I had considered doing what you asked for an have a single notification mechanism, which required it being able to be called from C code, so I passed in a `const char *` that was then cast to `QString`.
And then that whole thing pissed me off because we would use `tr()` to create a localized string (since this is user visible), then extract the `char *` from that, only to then pass it around again as `QString`.

mobile-widgets/qmlmanager.cpp
dirkhh: we have that same code a few lines above. The idea was to not constantly call the `instance()` method and instead just cache the value locally. But yes, you are of course correct. No reason to do this.

Editorial comment... I always find it interesting to see where people worry about wasting cycles and where they don't care. And what patterns are causing them to look for alternatives. Nothing wrong with any of the approaches, it's just interesting how you and I get 'triggered' by very different patterns :-)

mobile-widgets/qmlmanager.cpp
dirkhh: I need to look at that again... I think this is yet another left over from my initial idea of what I wanted to do there.

mobile-widgets/qmlmanager.cpp
bstoeger: huh?
I was commenting on a non-sequitur code-comment - you went off on a rant about something completely different?

Notably, I'm not interested in cycles here, that was purely an FYI remark. I just pointed out that the comment is not convincing in justifying the complexities of a call-back.

mobile-widgets/qmlmanager.cpp
dirkhh: The documentation EXPLICITLY states that if you use `MEMBER` that setting the variable will emit the signal. Except that in my tests it doesn't.
So yes, I can implement this in the beautifully elegant way envisioned by Qt and simply accept the fact that often the QML UI doesn't get notified... that didn't seem as useful to me as it may sound at first.
Or maybe my testing is broken and this all works in reality. Who knows.
The amount of ugly hacks that we have all over the place because Qt/QML doesn't do what realistically one thinks it should do. Like rows/columns/roles when it comes to ListViews.
But I'm happy to spend more time on trying to coax it into doing this differently... or simply just writing my own setter function. Which seems so stupid as there is exactly ONE TIME when I change that value from false to true.
Maybe that whole thing should be rethought and instead I should do one of those bizarre calls into QML instead of abusing yet another `bool` value to notify QML of a change.

Yet another thing where our code is chock full of ugly hacks that were added one at a time to fix one issue at one point in time.

dirkhh:
Oh, btw, and completely unrelated to this PR. Apparently my attempt to rip out crud that piled up in the code over time managed to break the no-cloud-to-cloud transition. I now have two reports that we no longer keep the no-cloud data around when switching to cloud mode.

Which is why I hate every time I "clean up" something that is obviously stupid and broken, but then end up realizing that the reason it is why it is because that was a weird ill documented hack around some specific situation..
BTW: I notice that my comments sound angry and bitter. The current global pandemic is not interacting well with my depression and I notice myself being a lot less positive and cheerful overall. So please don't take any of this as attacks on you, This really isn't about you at all. And I appreciate your feedback as it helps me question if I am doing the right things here.

mobile-widgets/qmlmanager.cpp
bstoeger: > The idea was to not constantly call the instance() method and instead just cache the value locally. But yes, you are of course correct. No reason to do this.

Note that this strategy will not help: Since compilers do whole-translation unit optimizations, they will simply inline that `instance()` method, which is a trivial access to a global variable.

OTOH, the version with the local static variable might do some locking, so it might actually be less effective. In this case it probably won't do locking because you are const-initializing the variable and then these things behave like a "global" static. Yes, C++ is complicated. :-P

mobile-widgets/qmlmanager.cpp
dirkhh: OK, tested this again. If I remove the emit, the UI never receives the update.

Which means this line `Qt emits automatically that signal when needed for MEMBER properties that do not have an explicit setter.` on this page https://doc.qt.io/qt-5/properties.html is flatly incorrect.

Would you prefer that I implement a setter in order to hide this ugliness from the `finishSetup()` function?

bstoeger: > The documentation EXPLICITLY states that if you use MEMBER that setting the variable will emit the signal. 

It can't when you do a simple assignment. It would have to somehow override `operator=()` for the variable. However, since the variable is simply a bool, this is not possible.

So with `setting the property` they surely mean a special function or macro. That's what I was asking for.

There actually is the `setProperty` function. But that is dynamic, so it is now very appealing.

> Maybe that whole thing should be rethought and instead I should do one of those bizarre calls into QML instead of abusing yet another bool value to notify QML of a change.

I think the Q_PROPERTY thing is the way it is supposed to work. As weird as it may be...

bstoeger: > OK, tested this again. If I remove the emit, the UI never receives the update.

I know - it can't auto-generate that emit. :)

> Would you prefer that I implement a setter in order to hide this ugliness from the finishSetup() function?

Not really. Even more lines of code. But if you feel it is "cleaner", "clearer" or something, it wouldn't bother me either.

mobile-widgets/qmlmanager.cpp
dirkhh: fixed both instances of this and decided to have one call the other because... why not.
which immediately showed another bug that I'm not tracking down...
we get a stray notification that we should get during initialization, and clearly libgit isn't doing a good job sending us update notification as we open local and remote repos.

mobile-widgets/qmlmanager.cpp
dirkhh: Also, thinking about this some more... I'm not convinced that I wouldn't want some form of notification on the desktop as well. Much of the backend code is indeed shared, which is why I have that function pointer mess in the first place.

core/qthelper.cpp
dirkhh: fixed

dirkhh:
Force pushed updates that address most of your comments. Still has the `emit` that seems needed - I asked above if you preferred me to hide that inside a single use setter - or use a completely different mechanism to notify the QML UI that we are done with the initialization.

mobile-widgets/qmlmanager.cpp
dirkhh: > huh?
> I was commenting on a non-sequitur code-comment - you went off on a rant about something completely different?
> 
> Notably, I'm not interested in cycles here, that was purely an FYI remark. I just pointed out that the comment is not convincing in justifying the complexities of a call-back.

Ignore my rants today, please. I think I said this elsewhere. I am not doing well right now as far as my mental health is concerned.

As for the complexity of a call-back. As a C programmer thinking about the cycles that we take for the callback vs the creation of any one random Qt data structure...
right, you aren't commenting on cycles. so I guess I just don't know what you were really commenting on. Anyway, apologies that my responses are a lot more aggressive than they should be

dirkhh:
oh the fun never stops.
Finally figured out what the HECK was going on. No idea, exactly, WHY it was going on, but at least I have now hacked around it.
My change that allowed the user to actually see the notifications as a side effect didn't load the DiveList into the pageStack until after we had initialized the models. And strangely that caused the ListView to instantiate _every single delegate_ - which meant that for the model with 8k dives we had about 100k model data accesses. Part of that number was my criminally naive implementation of the invalid dive handling which caused EIGHT model data accesses for each delegate, but even without that we still accessed the model 60k times. Crazy.
Simply not setting the model of the ListView until after the model was populated fixed this and made that maddening long delay go away. From 13 seconds to 7/100th of a second...

dirkhh:
What new hell... iOS? We appear to build `libxslt` but not the `libxml2` which that needs as a dependency... how did that ever work?

dirkhh:
ok, let's try this.
Dropped the "force lower case email" commit.
Tried a hack to fix iOS build. Which made me really mad at the iOS build script, so I tried to clean that one up while I was looking at it
Hopefully this will build everywhere and then I can push 3.0.3 betas

dirkhh:
not in love with the GitHub Action for iOS right now. It builds just fine here for me. So what the HECK is going on... ignoring that, the rest still deserves a review.

dirkhh:
I can successfully build for iPhone/iPad locally, but the build for the simulator that we do on GitHub Actions obviously fails right now with odd errors in the libxml2 configuration.
I certainly want to fix that, but that's not top of my priorities right now.
I'll try to get out a 3.0.3 beta soon.

mobile-widgets/qml/DownloadFromDiveComputer.qml
```diff
@@ -467,7 +467,7 @@ Kirigami.Page {
 					importModel.recordDives()
 					// it's important to save the changes because the app could get killed once
 					// it's in the background - and the freshly downloaded dives would get lost
-					manager.saveChangesLocal()
```
bstoeger, line 470: Note that this does *not* the same (except on iOs). `changesNeedSaving()` saves to the cloud, whereas `saveChangesLocal()` saves only locally. Not sure if you intend that.

dirkhh: it saves to the cloud if auto sync is enabled.
So yes, that's what I intended.
I think :-)

dirkhh:
there are so many loose ends I need to fix. The iOS build, three other bugs that I'm aware of.

If we ever get to the point where people were to actually truly test the betas that I do before a release, then maybe that feeling of frantic trying to race to catch up with the bugs we find in the production release wouldn't make me so reluctant to do production releases in the first place.

I must be delirious even thinking that.
Time to go to bed

mobile-widgets/qml/DownloadFromDiveComputer.qml
bstoeger: Ok - Then, while you're at it, please make saveChangesLocal() private to QMLManager, as I think that was the last external caller. When doing the call graph I thought there was a reason for doing this differently here.

bstoeger: A allowed myself to make that change and force-push. Please pull before going ahead to avoid complications.

dirkhh:
rebased on latest master, which will hopefully fix the iOS build.
Hoping to get @bstoeger 's other fixes merged so I can get new betas out that maybe aren't as horribly broken as the crap I'm torturing our users with right now :-)
Maybe they'll even save changes to storage...

dirkhh:
ok, all passed. Merging locally since I got bored and improved a couple of commit messages, but didn't touch the code and see no point in going through an hour of running checks **again**

PR #2719, opened 2020-04-04
Use the undo stack to check for unsaved changes on mobile

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Mobile doesn't use the undo stack when determining whether there are unsaved changes, which lead to not saving, when it should have saved. Therefore move this functionality from mainwindow.cpp to qthelper.cpp.

This currently breaks the tests, since I didn't figure out how to compile the undo-commands for the tests. Note that we will have to do this in the future anyway, because we *want* to test the undo commands and their interaction with the mobile models.

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -582,7 +582,7 @@ void MainWindow::on_actionCloudstoragesave_triggered()
 		return;
 
 	setCurrentFile(qPrintable(filename));
-	setFileClean();
+	Command::setClean();
 }
 
 void MainWindow::on_actionCloudOnline_triggered()
@@ -638,12 +638,6 @@ bool MainWindow::okToClose(QString message)
 	return true;
 }
 
-void MainWindow::setFileClean()
-{
-	mark_divelist_changed(false);
-	Command::setClean();
-}
-
 void MainWindow::closeCurrentFile()
 {
 	/* free the dives and trips */
@@ -655,7 +649,7 @@ void MainWindow::closeCurrentFile()
 	if (!existing_filename)
 		setTitle();
 	disableShortcuts();
-	setFileClean();
+	Command::setClean();
 
 	clear_events();
 
@@ -1568,7 +1562,7 @@ int MainWindow::file_save_as(void)
 		return -1;
 
 	setCurrentFile(qPrintable(filename));
-	setFileClean();
+	Command::setClean();
 	addRecentFile(filename, true);
 	return 0;
 }
@@ -1605,7 +1599,7 @@ int MainWindow::file_save(void)
 	}
 	if (is_cloud)
 		hideProgressBar();
-	setFileClean();
+	Command::setClean();
 	addRecentFile(QString(existing_filename), true);
 	return 0;
 }
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -585,14 +585,6 @@ void MainWindow::on_actionCloudstoragesave_triggered()
 	setFileClean();
 }
 
-// Currently we have two markers for unsaved changes:
-// 1) unsaved_changes() returns true for non-undoable changes.
-// 2) Command::isClean() returns false for undoable changes.
-static bool unsavedChanges()
-{
-	return unsaved_changes() || !Command::isClean();
-}
-
 void MainWindow::on_actionCloudOnline_triggered()
 {
 	bool isOffline = !ui.actionCloudOnline->isChecked();
@@ -612,7 +604,7 @@ void MainWindow::on_actionCloudOnline_triggered()
 	git_local_only = isOffline;
 	if (!isOffline) {
 		// User requests to go online. Try to sync cloud storage
-		if (unsavedChanges()) {
+		if (!Command::isClean()) {
 			// If there are unsaved changes, ask the user if they want to save them.
 			// If they don't, they have to sync manually.
 			if (QMessageBox::warning(this, tr("Save changes?"),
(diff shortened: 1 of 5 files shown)


dirkhh (commented):
God that's ugly and confusing.

It's the right thing to do, but the naming confusion is just horrible.

I should merge this, of course, but I worry about the tests. And I'm too tired to handle that tonight. Will look at it tomorrow

bstoeger:
Next attempt, which doesn't slurp in the undo-machinery into the core-lib.

Thereby, I found a fundamental problem:

In `parse_file()` we check whether the same git-repository was loaded *and* there are no changes. If so, it refuses to load. This is currently broken, because it doesn't consider the undo-commands. However, checking the undo-commands at this place would mean that the undo-commands are linked to basically everything (smtk-import, etc), which we probably don't want.

I haven't tried it, but this probably doesn't work anymore on desktop:
1) load from cloud
2) edit dives
3) decide that this was garbage - reload from cloud

I wonder whether it wouldn't be best to move the check up a layer. After all, the IO-layer is supposed to load files when it is told to, not to have its own mind on when to load things. Do you remember why you added this here in the first place in commit 8c1cc4524d19e05b816f3388448cacd735d64504? I guess it is to avoid aggressive syncing by the mobile code?

bstoeger:
Added a commit that removes this check. In the worst case this should give unneeded loading from the repository. However, I believe that mobile only syncs if there are changes anyway? If on desktop the user forces a sync, why not let them?

dirkhh:
For some reason it seems you dislike call back functions. This to me appears like the perfect use. Create a `registerChangedCheck(&function)` infrastructure where various pieces of code can register a C calling convention function that should be added to the list of functions called to check if there are changes. Have the undo code register a function there in its constructor. You don't link against undo, that function isn't called...
To me that is the cleanest way to deal with optional parts of the code - which is why I implemented the uiNotification that way. The cost of checking if a function pointer is 0 is absolutely negligible, even if you do it a million times. A single call to the undo stack is ten times more code executed.

desktop-widgets/mainwindow.cpp
```diff
@@ -612,7 +604,7 @@ void MainWindow::on_actionCloudOnline_triggered()
 	git_local_only = isOffline;
 	if (!isOffline) {
 		// User requests to go online. Try to sync cloud storage
-		if (unsavedChanges()) {
+		if (!Command::isClean()) {
```
dirkhh, line 607: yeah, why would the undo stack become the center of the universe here? This should be a core function, ideally callable from C code.

dirkhh (changes requested):
No, this isn't the way to do this. Because fundamentally you'll get to the point where everything relies on the undo code, export_html, smtk2ssrf, everything. A check whether the dive list is clean should NOT require UI code.

dirkhh:
It's late in Austria. Are you still working on this? Or would you prefer if I tried to do "my version" of this while you get your beauty sleep?

bstoeger:
This is not necessary at the moment, since you changed to `changesNeedSaving()`. I will care about that another time.

dirkhh:
> This is not necessary at the moment, since you changed to `changesNeedSaving()`. I will care about that another time.

I worry that there are other paths in the mobile code where we end up not checking for the state of the undo stack...

mobile-widgets/qmlmanager.cpp
```diff
@@ -301,10 +301,10 @@ void QMLManager::applicationStateChanged(Qt::ApplicationState state)
 	}
 	stateText.prepend("AppState changed to ");
 	stateText.append(" with ");
-	stateText.append((unsaved_changes() ? QLatin1String("") : QLatin1String("no ")) + QLatin1String("unsaved changes"));
+	stateText.append((!Command::isClean() ? QLatin1String("") : QLatin1String("no ")) + QLatin1String("unsaved changes"));
 	appendTextToLog(stateText);
 
-	if (state == Qt::ApplicationInactive && unsaved_changes()) {
+	if (state == Qt::ApplicationInactive && !Command::isClean()) {
```
dirkhh, line 307: so this here, for example, might be incorrect - unless we carefully made sure that every path that uses the undo code also set our unsaved_changes() flag...

mobile-widgets/qmlmanager.cpp
```diff
@@ -1729,7 +1729,7 @@ void QMLManager::screenChanged(QScreen *screen)
 
 void QMLManager::quit()
 {
-	if (unsaved_changes())
+	if (!Command::isClean())
```
dirkhh, line 1732: or what about this one?

bstoeger:
You have to look the other way round:

After which Command::... there is no call of changesNeedSaving():

1) editInvalid
2) pasteDive
3) addDive - but that's only transient so there shouldn't be any saving

The correct fix for now is to add the changesNeedSaving() there too, as it seems inconsistent not to do so.

If you still fear that we might miss something, I can come up with a stop-gap solution in 10 min.

I still believe that the test in parse_file() is bogus.

dirkhh:
I agree that I can look at the code paths that don't set the dive list as changed.
But isn't it cleaner to use the information that we already have before risking to miss a path?
So yeah, I would really prefer to add belts and suspenders here :-)
(doppelt gemoppelt hält besser).
I will take another look at the `parse_file()` issue right now

dirkhh:
Yeah... I think you convinced me. The optimization is `parse_file()` felt clever when I made it, but it is odd that the i/o layer would refuse to load things like that...

dirkhh:
I believe we can close this one

PR #2720, opened 2020-04-04
cleanup: copy saved_git_id, don't use local buffer

bstoeger:
In an attempt to reduce the number of global variables, don't use
a local buffer to store the currently loaded git-id. The git-id
itself is still a global variable, which in the future can hopefully
be encapsulated in a "struct File" or similar.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A tiny step in reducing the number of global variables.

### DIFF
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -1677,15 +1677,17 @@ static int load_dives_from_tree(git_repository *repo, git_tree *tree, struct git
 
 void clear_git_id(void)
 {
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
I will admit that I don't quite see how this makes any difference at all, but I trust you.

PR #2721, opened 2020-04-04
undo/mobile: keep track of dive sites

bstoeger:
When editing the dive site of a dive, the dive-table of the
corresponding dive site was not properly updated by the undo
commands. Try to get this right.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a data-inconsistency when editing the dive site on mobile: the dive-table of the dive site was not updated properly. It's a small wonder that that hasn't caused any crashes. This tries to do it right and in my few tests I couldn't get it to crash.

This code is pretty hairy. I really wish that we could use desktops' fine grained undo to avoid having to get all these subtle things right twice. Idea: do fine grained undo, but only save if the user exits the edit-screen. The question then is: do we save after every undo as well?

PS: I found a weird behavior: If a dive-site has GPS, the user can't simply remove the dive from that site: If the user erases the name, then the code will detect that there are GPS coordinates and attempt to create a new dive site. For that purpose it takes the name of the *old* dive site, which the user wanted to clear:

```
                        setupDivesite(res, d, ds, lat, lon, qPrintable(myDive.location));
```

That appears wrong. But I don't know what happens if we pass the empty string here.

### Changes made:

1) Add and remove dive from dive_site in EditDive command.

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1004,13 +1004,14 @@ void EditWeight::undo()
 EditDive::EditDive(dive *oldDiveIn, dive *newDiveIn, dive_site *createDs, dive_site *editDs, location_t dsLocationIn)
 	: oldDive(oldDiveIn)
 	, newDive(newDiveIn)
+	, newDiveSite(newDiveIn->dive_site)
 	, changedFields(DiveField::NONE)
 	, siteToRemove(nullptr)
 	, siteToAdd(createDs)
 	, siteToEdit(editDs)
 	, dsLocation(dsLocationIn)
 {
-	if (!oldDive || ! newDive)
+	if (!oldDive || !newDive)
 		return;
 
 	setText(Command::Base::tr("Edit dive [%1]").arg(diveNumberOrDate(oldDive)));
@@ -1061,6 +1062,8 @@ EditDive::EditDive(dive *oldDiveIn, dive *newDiveIn, dive_site *createDs, dive_s
 		changedFields |= DiveField::NOTES;
 	if (oldDive->salinity != newDive->salinity)
 		changedFields |= DiveField::SALINITY;
+
+	newDive->dive_site = nullptr; // We will add the dive to the site manually and therefore saved the dive site.
 }
 
 void EditDive::undo()
@@ -1097,9 +1100,16 @@ void EditDive::exchangeDives()
 
 	// Bluntly exchange dive data by shallow copy.
 	// Don't forget to unregister the old and register the new dive!
+	// Likewise take care to add/remove the dive from the dive site.
 	fulltext_unregister(oldDive);
+	dive_site *oldDiveSite = oldDive->dive_site;
+	if (oldDiveSite)
+		unregister_dive_from_dive_site(oldDive); // the dive-site pointer in the dive is now NULL
 	std::swap(*newDive, *oldDive);
 	fulltext_register(oldDive);
+	if (newDiveSite)
+		add_dive_to_dive_site(oldDive, newDiveSite);
+	newDiveSite = oldDiveSite; // remember the previous dive site
 	invalidate_dive_cache(oldDive);
 
 	// Changing times may have unsorted the dive and trip tables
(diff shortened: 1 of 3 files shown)


dirkhh (commented):
That's less code than I expected, given the PR description.
So I'm guessing that I am missing something subtle here.
More coffee.

dirkhh:
(1) about using the fine grained undo code. Remember, we did that at first. And it created a dozen entries (or something along that line) in the undo stack when you added a dive. And the order of those entries was the order of execution in the code, not the order in which the user made changes. And I tried to figure out how to do that with macro commands and how to remove shit from the undo stack and it's all a major mess because there aren't really any good ways to make modifications to the undo stack itself.
(2) about saving after every edit. HECK NO. on a large dive list with an older device a simple one line change takes a couple of seconds. The UI will become utterly unusable.
(3) we should treat receiving back an empty name as a special case that says "I don't want a dive site for this dive". I briefly considered adding a warning about losing the GPS location as I was typing this, but hey, we have undo, so if this isn't what the user wanted, they can undo the operation.

bstoeger:
> about using the fine grained undo code. Remember, we did that at first. And it created a dozen entries (or something along that line) in the undo stack when you added a dive. And the order of those entries was the order of execution in the code, not the order in which the user made changes. And I tried to figure out how to do that with macro commands and how to remove shit from the undo stack and it's all a major mess because there aren't really any good ways to make modifications to the undo stack itself.

This is not what I'm thinking about. The idea was to post undo commands for every user initiated edit action, not numerous commands at once to *simulate* a bigger command. It works fine on desktop, so I'm not convinced that it is a disaster on mobile.

> about saving after every edit. HECK NO.

Note that I have expressly written:

> but only save if the user exits the edit-screen

That means precisely as many saves as now. The question then was how we treat *undo/redo*. One could for example mark those undo commands were a save has happened. And save when one steps across these. Just throwing ideas out there - because maintaining these subtle things twice is not viable.

> we should treat receiving back an empty name as a special case that says "I don't want a dive site for this dive"

Well, that should be trivial to implement.

dirkhh:
> > about using the fine grained undo code. Remember, we did that at first. And it created a dozen entries (or something along that line) in the undo stack when you added a dive. And the order of those entries was the order of execution in the code, not the order in which the user made changes. And I tried to figure out how to do that with macro commands and how to remove shit from the undo stack and it's all a major mess because there aren't really any good ways to make modifications to the undo stack itself.
> 
> This is not what I'm thinking about. The idea was to post undo commands for every user initiated edit action, not numerous commands at once to _simulate_ a bigger command. It works fine on desktop, so I'm not convinced that it is a disaster on mobile.

Yes, I think I understand now. Instead of doing the multiple undo commands in the `commitChanges()` function, I assume you mean whenever the user leaves a field in the UI.

The problem with that is that if the user cancels the edit, all of these actions are still on the undo stack. And it's really confusing.

I have tried hard to find a way to remove things from the undo stack and I wasn't able to find a way to do that. If you can solve THAT and allow us to remove all the changes when a user cancels out of an edit, then I'm willing to figure out how to do all of this in QML and create multiple, fine grained call back functions in `QMLManager`.

> That means precisely as many saves as now. The question then was how we treat _undo/redo_. One could for example mark those undo commands were a save has happened. And save when one steps across these. Just throwing ideas out there - because maintaining these subtle things twice is not viable.

I don't understand what you are proposing here.

> > we should treat receiving back an empty name as a special case that says "I don't want a dive site for this dive"
> 
> Well, that should be trivial to implement.

So many things should be trivial to implement. And then I try to implement them and it turns out that I'm wrong. That is quite possibly the best summary of my life as software developer that I've ever written.

dirkhh:
I don't think that's something we can get done "today" (for some value of "today"). So I think I should merge what you have done here and try to get an actually working 3.0.3 out to our poor users.

Thankfully, thanks to COVID few people are on dive trips and the fact that changes aren't saved isn't really bothering a lot of people right now...

dirkhh:
OK, did some testing and convinced myself that this change should go in.

PR #2722, opened 2020-04-04
GitHub Actions: fix iOS build

dirkhh:
This cherry picks a few commits from my Mobile fixes branch as apparently more stuff has changed while I wasn't looking. Hopefully this will get us to a working build.

- [x] Build system change

Yeah, this is wild speculation with a bold commit message in case I guessed right

### Changes made:

- Install Homebrew pkg-config on the iOS build VMs
- add libxml2 to the iOS build (I'm actually not sure how this ever worked without that -- maybe it used host header files and doesn't actually link against the library? I'm not sure)
- fix the libxslt build to find those header files
- clean up the mess that was our iOS build script (at least a little - there's more room for cleanup there)

### Related issues:

every iOS GitHub action run failing

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -14,7 +14,7 @@ jobs:
     - name: checkout sources
       uses: actions/checkout@v1
     - name: setup Homebrew
-      run: brew install autoconf automake libtool
+      run: brew install autoconf automake libtool pkg-config
     - name: set our Qt build
       run: |
         env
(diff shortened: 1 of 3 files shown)


dirkhh:
Well, I guess with the iOS build having successfully completed I can call this a successful pull request.
I will wait for the remaining ones that are running, **just because**.
Once I merge this I'll rebase my ui fixes branch on top of this - might make sense to rebase other on top of this as well so we stop having all those red 'x'es.

PR #2723, opened 2020-04-04
mobile: dive site fixes

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Two dive-site-editing fixes. Not certain about the second one.

### Changes made:

1) Don't lose dive site change events it the dive is set to a different site without editing the site
2) Use the new dive site name when creating new gps dive sites.


dirkhh (commented):
This is confusing me. Need to look at this in QtCreator and look at git blame to figure out why it is what it is.
I'm not at all saying that your fix is wrong. Just that I would like to understand why we have the code that we have.

dirkhh (approved):
Yeah, that was weird. Your code seems obviously "more correct"...

There's still something odd here as in my testing I sometimes get gasmix related texts from the qDebug() in this code...?????

PR #2729, opened 2020-04-07
Minor changes to divelist-changed code

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Two changes we discussed previously:

### Changes made:

1) Don't test for git SHA in parse_file().
2) Unify duplicate code in mobile and desktop.

### Mentions:

@dirkhh: take only the first commit if you feel unsure about the code-unification thing. I think it doesn't matter a lot as the functions are tiny anyway.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -807,7 +807,7 @@ void mark_divelist_changed(bool changed)
 	updateWindowTitle();
 }
 
-int unsaved_changes()
+bool is_divelist_changed()
 {
 	return dive_list_changed;
 }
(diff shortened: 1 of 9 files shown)


bstoeger:
Ok - the mobile tests fail. However, that is not worth fighting for now. Let's reconsider later, when we include the undo-commands in the tests, which we should really do in the near future.

PR #2730, opened 2020-04-07
git: don't check for git sha in parse_file()

bstoeger:
parse_file() refused to load from a git repository if we already
had that repository and there were no changes. However, this only
checked the global divelist_changed flag, which does not track
undo-commands. Thus, after editing dives the user couldn't reload
from git.

Remove this check. It is somewhat questionable that the io layer
refuses to load from a repository anyway. Let the caller decide.
There appears to be a check_git_sha function for that purpose(?).

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a borderline bug-fix as a user might not be able to reload from cloud after doing some changes. See previous discussion.

### DIFF
--- a/core/file.c
+++ b/core/file.c
@@ -12,7 +12,6 @@
 
 #include "dive.h"
 #include "subsurface-string.h"
-#include "divelist.h"
 #include "errorhelper.h"
 #include "file.h"
 #include "git-access.h"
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
Let's see if this is another one where much later we figure out that there are unintended side effects. For now I just can't see how there would be.

PR #2733, opened 2020-04-08
Fulltext: attempt to not crash when regenerating the fulltext cache

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is code hygiene and hopefully a crash-fix.
### Changes made:

1) The fulltext_reload() function was supposed to create the initial fulltext cache. Therefore rename it to fulltext_populate()
2) It was crashing if called on a dive list that had was already indexed previously. Fix that by not clearing the word chache.

### Mentions:

@dirkhh: Please try if this fixes your problem with the git-test. However, I fear that the whole parse directly into the existing dive-list is problematic.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -831,7 +831,7 @@ void process_loaded_dives()
 	/* Autogroup dives if desired by user. */
 	autogroup_dives(&dive_table, &trip_table);
 
-	fulltext_reload();
+	fulltext_populate();
 }
 
 /*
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
This makes sense and is a much better name.

dirkhh:
I agree with your evaluation of the use of parse_file in the GitStorage test, so I'll fix that separately as I try to figure out how to make that test workable at scale.

PR #2735, opened 2020-04-09
Core: Support libgit2 v1 when using system libgit2

flokli:
This ifdefs doesn't properly handle libgit2 >=1.0
(where LIBGIT2_VER_MAJOR == 1)

Invert the logic like done in other places.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -359,10 +359,10 @@ static int try_to_git_merge(git_repository *repo, git_reference **local_p, git_r
 	}
 
 	git_merge_init_options(&merge_options, GIT_MERGE_OPTIONS_VERSION);
-#if !LIBGIT2_VER_MAJOR && LIBGIT2_VER_MINOR > 23
-	merge_options.flags = GIT_MERGE_FIND_RENAMES;
-#else
+#if !LIBGIT2_VER_MAJOR && LIBGIT2_VER_MINOR < 22
 	merge_options.tree_flags = GIT_MERGE_TREE_FIND_RENAMES;
+#else
+	merge_options.flags = GIT_MERGE_FIND_RENAMES;
 #endif
 	merge_options.file_favor = GIT_MERGE_FILE_FAVOR_UNION;
 	merge_options.rename_threshold = 100;


core/git-access.c
```diff
@@ -359,10 +359,10 @@ static int try_to_git_merge(git_repository *repo, git_reference **local_p, git_r
 	}
 
 	git_merge_init_options(&merge_options, GIT_MERGE_OPTIONS_VERSION);
-#if !LIBGIT2_VER_MAJOR && LIBGIT2_VER_MINOR > 23
-	merge_options.flags = GIT_MERGE_FIND_RENAMES;
-#else
+#if !LIBGIT2_VER_MAJOR && LIBGIT2_VER_MINOR < 22
```
dirkhh, line 362: So this is obviously incorrect as it switches from `>23` to in effect `!<22` which is very wrong.
But the bigger question is "why would we still support anything older than libgit2 0.26.
I'll need to go through the source and make that happen.

dirkhh (changes requested):
You are welcome to adjust your PR, but I think it might be easier if I just fix this for all of the builds.

core/git-access.c
flokli: There's only 3 occurences - I can update the PR to get rid of the old usages.

dirkhh:
That's part of it.
The other part is fixing the build scripts and cmake dependencies to ensure we find 0.26 or newer...

flokli:
I found some libgit 0.20 comments in android and macos files…

flokli:
Hm, it also seems the `pkg_config_library` abstraction makes it a bit harder to pass a version boundary to `pkg_check_modules` - I think I'll leave that up to you. Thanks!

dirkhh:
OK, I'll take care of it in the next couple of days.
Those comments are part of what needs cleaning up. All the builds that I drive use libgit2 >= 0.26, including Mac and Android. And I believe that at this point all even remotely reasonable Linux distributions ship a version that's newer than that (and the scripts are catching it if they don't - except that those version checks are ALSO broken).
So yeah, this is a slightly bigger task to get right :(

Thanks for bringing it to my attention, that's really useful and appreciated.

PR #2737, opened 2020-04-10
Libgit cleanup

dirkhh:

- [x] Bug fix
- [x] Functional change
- [x] Code cleanup
- [x] Build system change

This is a replacement for #2735, but it goes much further

### Changes made:

1) remove conditional code for older libgit2
2) cleanup version checks for libgit2 in the build system
3) while we are at it, improve the commit message for merges

### Release note:

not needed, this really isn't user visible

### Documentation change:

not needed, this really isn't user visible

### Mentions:

@flokli

### DIFF
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -349,6 +349,7 @@ static int try_to_git_merge(git_repository *repo, git_reference **local_p, git_r
 	git_commit *local_commit, *remote_commit, *base_commit;
 	git_index *merged_index;
 	git_merge_options merge_options;
+	struct membuffer msg = { 0, 0, NULL};
 
 	if (verbose) {
 		char outlocal[41], outremote[41];
@@ -430,10 +431,12 @@ static int try_to_git_merge(git_repository *repo, git_reference **local_p, git_r
 		goto write_error;
 	if (git_tree_lookup(&merged_tree, repo, &merge_oid))
 		goto write_error;
-	if (git_signature_default(&author, repo) < 0)
-		if (git_signature_now(&author, "Subsurface", "noemail@given") < 0)
-			goto write_error;
-	if (git_commit_create_v(&commit_oid, repo, NULL, author, author, NULL, "automatic merge", merged_tree, 2, local_commit, remote_commit))
+	if (get_authorship(repo, &author) < 0)
+		goto write_error;
+	const char *user_agent = subsurface_user_agent();
+	put_format(&msg, "Automatic merge\n\nCreated by %s\n", user_agent);
+	free((void *)user_agent);
+	if (git_commit_create_v(&commit_oid, repo, NULL, author, author, NULL, mb_cstring(&msg), merged_tree, 2, local_commit, remote_commit))
 		goto write_error;
 	if (git_commit_lookup(&commit, repo, &commit_oid))
 		goto write_error;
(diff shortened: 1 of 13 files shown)


flokli (commented):
Small nit: the last patch in the series needs a verb, maybe s/we can which/we can see which/.

I however wasn't able to compile this, as `core/unix.c` doesn't apply on top of 4.9.3, and master needs a `libdivecomputer/ble.h` not yet part of libdivecomputer-subsurface-branch-4.9.3.tgz.

core/git-access.c
```diff
@@ -434,10 +430,13 @@ static int try_to_git_merge(git_repository *repo, git_reference **local_p, git_r
 		goto write_error;
 	if (git_tree_lookup(&merged_tree, repo, &merge_oid))
 		goto write_error;
-	if (git_signature_default(&author, repo) < 0)
-		if (git_signature_now(&author, "Subsurface", "noemail@given") < 0)
-			goto write_error;
-	if (git_commit_create_v(&commit_oid, repo, NULL, author, author, NULL, "automatic merge", merged_tree, 2, local_commit, remote_commit))
+	if (get_authorship(repo, &author) < 0)
+		goto write_error;
```
bstoeger, line 435: I'm surprised that in C you can jump across the definition of a variable. In C++ that wouldn't be accepted, because it would mean that you destroy the object without having initialized it.

That raises the question: what happens if the code jumps to `write_error` here? Will `free_buffer` be called on an unitialized `msg`?

dirkhh:
> Small nit: the last patch in the series needs a verb, maybe s/we can which/we can see which/.

Ahh - will fix

> I however wasn't able to compile this, as `core/unix.c` doesn't apply on top of 4.9.3, and master needs a `libdivecomputer/ble.h` not yet part of libdivecomputer-subsurface-branch-4.9.3.tgz.

This patch is relative to latest master, of course.
As for getting the right libdivecomputer, that's super easy:
- clone the branch for this PR
- `cd subsurface`
- `git submodule init && git submodule update`
- `mkdir -p libdivecomputer/build`
- `cd libdivecomputer`
- `autoreconf --install`
- `cd build`
- `../configure`
- `make install`

But, in reality, I would **STRONGLY** encourage you to simply use the `scripts/build.sh` to create a working build as that deals with quite a few additional pitfalls.

core/git-access.c
dirkhh: yes, and that would be BAD. I need to move the declaration / initialization to the top of the function.

dirkhh:
Fix to both comments pushed. Thanks for the feedback.
Interesting dilemma. The `membuf` never gets touched before all those jumps to `diverged_error:` so I don't have a `free_buffer(&msg)` there... which of course looks inconsistent and there's a theoretical risk that in the future we could have a goto there later in the function (or use `msg` earlier. Calling `free_buffer()` on an unused `membuf` is cheap... but looking at today's code is also clearly unnecessary. I decided not to add that call, but am interested in what others think :-)

flokli:
>This patch is relative to latest master, of course.
>As for getting the right libdivecomputer, that's super easy:
>- clone the branch for this PR
>- `cd subsurface`
>- `git submodule init && git submodule update`

Thanks, I wasn't aware a compatible libdivecomputer is also shipped as a
git submodule.  This is much easier than fetching a tarball from a
branch for each release, thanks :-)

>- `mkdir -p libdivecomputer/build`
>- `cd libdivecomputer`
>- `autoreconf --install`
>- `cd build`
>- `../configure`
>- `make install`
>
>But, in reality, I would **STRONGLY** encourage you to simply use the
>`scripts/build.sh` to create a working build as that deals with quite a
>few additional pitfalls.

Yeah, I'd agree, but in [our case](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/misc/subsurface/default.nix),
there's also some qt-specifics, separate builds of libdivecomputer etc,
which is why I can't just invoke `build.sh`.

However, both qmake and autoconf-style projects are well - supported,
and as subsurface supports pkgconfig to pass in dependencies, it's all
fine :-)

Picking libdivecomputer from the git submodule, I was able to build and
run this PR with libgit2 1.0.0.

flokli:
Nice, thanks! ~~Are you planning to do a release soon, or should I backport these patches on top of 4.9.3 to fix the build in nixpkgs?~~

Edit: Saw your commit on the nixpkgs PR - I'll see if I can fix it locally.

Issue #2739, opened 2020-04-10
Petrel 2 not recognized as dive computer


ncgator82:
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
iOS 13 and Shearwater Petrel2 will not connect to Subsurface mobile. Toggled verbose 
logging.  Also toggled on  temporarily show all bluetooth devices.  App log shows this:
Found new device = ""Petrel" "LE:{4447fea5-ffbb-7015-3be1-d5c9a2b0a33f}" Not
recognized as dive computer.  Device is actually Petrel2 but I am unable to achieve a connection between Subsurface Mobile and the Shearwater.  My current workaround is by connecting to a Windows PC, then performing a cloud sync.

### Operating system:
iOS 13.4.1 English
 iPhone 11 

### Subsurface version:
3.0.3(4.9.3.1328)

dirkhh:
I'm puzzled that this wouldn't be recognized as dive computer...
```
	if (btName.startsWith("Predator") ||
	    btName.startsWith("Petrel") ||
```
Very strange.
There is a settings option to show all BT/BLE connections. If you enable that one (sorry, this is "per session only", so it does NOT stay enable between runs of the software), can you download from your Petrel2?
Please give that a try, and right after the download attempt (regardless of whether it's successful or not), please tap on Help->Ask for support in the app and send me the log that this will create as an email (the default recipient address that it will populate is just going to me, so this stays private in case you worry about that).

ncgator82:
This release fixed my issue. Thanks so much!!

Rick Holcombe

PR #2740, opened 2020-04-10
iOS fixes

dirkhh:

- [x] Bug fix
- [x] Code cleanup

This fixes one potential crash that I've only seen on iOS so far, and also fixes a bug where we didn't correctly detect a Shearwater Petrel 2 on iOS #2739 
The last commit is some code cleanup so we stop doing comparisons once we have a match.

### Related issues:

#2739 

### Release note:

done

### Mentions:

@ncgator82

### DIFF
--- a/core/connectionlistmodel.cpp
+++ b/core/connectionlistmodel.cpp
@@ -39,9 +39,12 @@ void ConnectionListModel::addAddress(const QString &address)
 
 void ConnectionListModel::removeAllAddresses()
 {
-	beginRemoveRows(QModelIndex(), 0, rowCount());
+	if (rowCount() == 0)
+		return;
+
+	beginResetModel();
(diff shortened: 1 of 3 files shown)


core/connectionlistmodel.cpp
```diff
@@ -39,9 +39,12 @@ void ConnectionListModel::addAddress(const QString &address)
 
 void ConnectionListModel::removeAllAddresses()
 {
-	beginRemoveRows(QModelIndex(), 0, rowCount());
+	if (rowCount() == 0)
```
bstoeger, line 42: With `beginResetModel()`/`endResetModel()` the check is redundant.

dirkhh: it's redundant. Is it wrong?

PR #2743, opened 2020-04-12
delegates: remove TankInfoDelegate::createEditor()

bstoeger:
When creating a TankInfoDelegate editor, reploting of the profile
was disabled to avoid replotting when the user scrolls through
the tank-info list. Since the code was changed to only set the
tank-info when the editor is closed, this became unnecessary
(hopefully). Indeed the clearing of the flag was removed in a
previous flag. This means that we also have to remove the setting
of the flag. Since this is all the TankInfoDelegate::createEditor()
function was doing, we can remove the whole function.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a fallout of the cylinder-undo code. The clearing of the don't replot-flag was removed. Therefore, the setting should also be removed.

### DIFF
--- a/desktop-widgets/modeldelegates.cpp
+++ b/desktop-widgets/modeldelegates.cpp
@@ -247,13 +247,6 @@ void TankInfoDelegate::editorClosed(QWidget*, QAbstractItemDelegate::EndEditHint
 		mymodel->setData(IDX(CylindersModel::TYPE), currCombo.activeText, CylindersModel::COMMIT_ROLE);
 }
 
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
Overall approved, but please fix the typo in the first commit message - the last 'flag' should be 'commit'...

bstoeger:
Done.

Is it possible to edit commit messages directly on Github? Doing it locally is workflow-breaking. :(

dirkhh:
That's a feature request that I've made for repeatedly - and it apparently was there and then was removed because... reasons...
So yeah, I think you need to do that locally and force push :-(

PR #2744, opened 2020-04-12
[TRIVIAL] cleanup: remove unnecessary profile update after adding new dive

bstoeger:
MainWindow::on_actionAddDive_triggered() updated the profile after
calling Command::addDive(). However, that is redundant because the
undo-machinery does the profile reload. Remove the call.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a trivial cleanup that removes a redundant profile replot. Tried it and things still work as normal.

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -86,7 +86,6 @@ namespace {
 	};
 }
 
-
 extern "C" int updateProgress(const char *text)
 {
 	if (verbose)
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
I'm pretty sure we have many instances where we have redundant calls like this... :-(

PR #2745, opened 2020-04-12
Three fixes

atdotde:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I was made aware that editing a dive in the planner loses the dive computer events attached to the dive. Commit 2/3 adds code to preserve those. The other two commits fix two complaints by the compiler (the fixes look right to me but are not in parts of the code that i know too well).

### Changes made:

1) Reshuffle members of a struct to make the variable length part last again
2) After editing a dive in the planner, copy events from the original dive until the point where we modified the dive by deleting waypoints
3) Remove an extra argument in a function call

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -127,6 +127,11 @@ int event_is_gaschange(const struct event *ev)
 		ev->type == SAMPLE_EVENT_GASCHANGE2;
 }
 
+bool event_is_divemodechange(const struct event *ev)
+{
+	return same_string(ev->name, "modechange");
+}
+
 struct event *create_event(unsigned int time, int type, int flags, int value, const char *name)
 {
 	int gas_index = -1;
@@ -571,6 +576,31 @@ void copy_events(const struct divecomputer *s, struct divecomputer *d)
 		ev = ev->next;
 	}
 	*pev = NULL;
+}
+
+/* copies all events from all dive computers before a given time
+   this is used when editing a dive in the planner to preserve the events
+   of the old dive */
+void copy_events_until(const struct dive *sd, struct dive *dd, int time)
+{
+	if (!sd || !dd)
+		return;
+
+	const struct divecomputer *s = &sd->dc;
+	struct divecomputer *d = &dd->dc;
+
+	while (s && d) {
+		const struct event *ev;
+		ev = s->events;
+		while (ev != NULL) {
+			// Don't add events the planner knows about
+			if (ev->time.seconds < time && !event_is_gaschange(ev) && !event_is_divemodechange(ev))
+				add_event(d, ev->time.seconds, ev->type, ev->flags, ev->value, ev->name);
+			ev = ev->next;
+		}
+		s = s->next;
+		d = d->next;
+	}
 }
 
 int nr_cylinders(const struct dive *dive)
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -1287,7 +1287,9 @@ static struct dive *get_last_valid_dive()
  */
 int get_dive_nr_at_idx(int idx)
 {
-	struct dive *last_dive = get_last_valid_dive(dive_table.nr - 1);
+	if (idx < dive_table.nr)
+		return 0;
+	struct dive *last_dive = get_last_valid_dive();
 	if (!last_dive)
 		return 1;
 	return last_dive->number ? last_dive->number + 1 : 0;
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -771,8 +771,8 @@ static int get_divemode(const char *divemodestring) {
  */
 struct parse_event {
 	const char *name;
-	struct event ev;
 	int has_divemode;
+	struct event ev;
 };
 
 static void parse_event_keyvalue(void *_parse, const char *key, const char *value)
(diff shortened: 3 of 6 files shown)


core/divelist.c
```diff
@@ -1287,7 +1287,7 @@ static struct dive *get_last_valid_dive()
  */
 int get_dive_nr_at_idx(int idx)
 {
-	struct dive *last_dive = get_last_valid_dive(dive_table.nr - 1);
+	struct dive *last_dive = get_last_valid_dive();
```
bstoeger, line 1292: The whole function appears to be buggy.
There seems to be a
```
        if (idx < dive_table.nr)
                return 0;
```
missing, isn't it? At least that's what I read from the comment.

dirkhh: but that condition is true for all dives. What am I missing?

bstoeger: I don't think so - if you add at the end of the table then `idx == dive_table.nr`. However, the error condition is probably a not very common case - you would have to have dives in the future and then add a new dive with the current date, or something. All a bit synthetic.

core/load-git.c
```diff
@@ -771,8 +771,8 @@ static int get_divemode(const char *divemodestring) {
  */
 struct parse_event {
 	const char *name;
-	struct event ev;
 	int has_divemode;
+	struct event ev;
```
bstoeger, line 775: Not that it makes a difference, but the compiler is wrong here. The variable-length array is not accessed.

dirkhh (commented):
I'm confused about the third commit and what that function does... need to read that overall code some more.
In the meantime, could you fix the typo in the commit message for the second commit (delete)

core/divelist.c
dirkhh: very common case - all manually added dives tend to be in the future by default.
But as I said in my review comment, I need to read the calling code to actually understand what this does

bstoeger: Not sure if you're being facetious.

The common case is that the newly added dive (which gets the current timestamp) is the last in the list. If this it isn't the case (which is probably very uncommon), then this code fails.

What you have to do to hit this bug:
1) Add dive (gets current timestamp).
2) Set dive date to the future.
3) Add another dive -> boom wrong number.

core/load-git.c
atdotde: My compiler (clang) complains that this is only allowed in a GNU extension of the standard to have a variable length element not at the end. So why not reshuffle in a way that the standard is happy?

bstoeger: Sorry - I didn't want to imply that we shouldn't shut up the compiler - I was just idly pointing out that this is not a real bug. Note that in C++ mode this would cause a hard error, not a warning [because different struct members must have different addresses, which is not possible if the variable sized array has 0 size].

BTW, we use numerous GNU extensions, such as these weird case labels with ranges.

core/divelist.c
atdotde: @bstoeger wrote that in 0d78fe45c3

bstoeger: Thanks - I already noticed that. :-D

bstoeger (approved):
Overall I agree with these changes, though I'm not comfortable enough with the planner to properly assess the first commit. I will fix the function of the last commit if nobody else wants to.

core/dive.c
```diff
@@ -573,6 +573,30 @@ void copy_events(const struct divecomputer *s, struct divecomputer *d)
...
+
+	const struct divecomputer *s = &sd->dc;
+	struct divecomputer *d = &dd->dc;
+
+	while (s && d) {
+		const struct event *ev;
+		ev = s->events;
+		while (ev != NULL) {
+			if (ev->time.seconds < time && !event_is_gaschange(ev))
+				add_event(d, ev->time.seconds, ev->type, ev->flags, ev->value, strdup(ev->name));
```
bstoeger: Just one nitpick - this seems to leak memory: `add_event` copies the name (because the name is part of the event-structure). Thus, the `strdup` is redundant, isn't it?

dirkhh:
My suggestion would be that you and @atdotde collaborate to fix this (including memory leaks, etc).
sometimes I find the back & forth on GitHub useful and productive, sometimes it feels disconnected and oddly confrontational even if I know that's not how it is intended.

bstoeger:
The only contentious point from my side is the `strdup()`, though I haven't tested the planner fixes.

I'm already too tired to look at the dive-numbering thing today. Will do that tomorrow. In any case, removal of the parameter is fine.

atdotde:
This push adds the if clause @bstoeger suggested. Please confirm if this is the fix you intended. It also gets rid of the strdup and expcludes dive mode change events (as the planner has already added those).

bstoeger:
Yes, I think this is good, thanks.

[Side note: Strictly speaking, this could still be improved by considering invalid dives. However, I think that is such an obscure case that it's not worth bothering about for now.]

bstoeger:
I reckon that you have tested the planner part, so I see no reason to keep this back.

PR #2746, opened 2020-04-13
Remove fine-grained startup notifications

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes fine grained startup notifications and subjectively makes startup much faster for me. Part of that is due to an objective speed-up, see commit message.

Note that this is only mobile-on-desktop. On device YMMV.
### Changes made:

1) Don't do fine grained ui-notifications on startup.

### DIFF
--- a/core/fulltext.cpp
+++ b/core/fulltext.cpp
@@ -148,14 +148,9 @@ void FullText::populate()
 	uiNotification(QObject::tr("start processing"));
 	int i;
 	dive *d;
-	for_each_dive(i, d) {
-		// this makes sure that every once in a while we allow the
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
I'm not going to question your measurements.
Looking at the code being executed the changes in processing time seem... surprising.
But, whatever. On most devices, with most dive lists the updates were not providing a lot of objective value, and given your measurements, this is the right thing to do.

PR #2747, opened 2020-04-13
mobile: don't call main loop for notifications once initialized

bstoeger:
Calling qApp->processEvents() in QMLManager::setNotificationText()
caused crashes, because it could lead to the context-menu that
initialized the call being deleted. Something that QML apparently
doesn't like.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes the reproducible crash that I reported yesterday on the mailing list. I simply have to delete a dive from the dive context menu (long click on dive in list).

Perhaps this impedes notification? Alternatively, we could simply clear the notification callback once initialization is done, but I feel that version would not be welcomed.

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -1710,7 +1710,12 @@ void QMLManager::setNotificationText(QString text)
 	appendTextToLog(QStringLiteral("showProgress: ") + text);
 	m_notificationText = text;
 	emit notificationTextChanged();
-	qApp->processEvents();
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
We keep going back and forth on this one. It seems possible that the crash is indeed a bug in Kirigami code (since it came back when we added the context menu), but I obviously prefer a sluggish UI to crashes. The problem here is that we again are likely to not get updates during slow storage / network operations to the user, but so be it. Better than having the app crash.

PR #2748, opened 2020-04-13
desktop: remove special QGroupBox stylesheet for non-Windows systems

bstoeger:
That style-sheet made things look really ugly on most Linux themes.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes an uglification of the the DiveInfoTab on Linux.
 
### Changes made:

1) Remove special non-Windows QGroupBox stylesheet.
### Mentions:

@willemferguson: It seems you wrote that stylesheet for Windows, but actually it was only included for non-Windows?

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveInformation.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveInformation.cpp
@@ -38,11 +38,7 @@ TabDiveInformation::TabDiveInformation(QWidget *parent) : TabBase(parent), ui(ne
 	connect(ui->diveType, SIGNAL(currentIndexChanged(int)), this, SLOT(diveModeChanged(int)));
 	QString CSSSetSmallLabel = "QLabel { color: mediumblue; font-size: " +                           // Using label height
 		QString::number((int)(0.5 + ui->diveHeadingLabel->geometry().height() * 0.66)) + "px;}"; // .. set CSS font size of star widget subscripts
-#if defined(Q_OS_WIN)
 	ui->scrollAreaWidgetContents_3->setStyleSheet("QGroupBox::title { color: mediumblue;} ");
-#else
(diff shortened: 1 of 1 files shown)


desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -38,11 +38,7 @@ TabDiveInformation::TabDiveInformation(QWidget *parent) : TabBase(parent), ui(ne
 	connect(ui->diveType, SIGNAL(currentIndexChanged(int)), this, SLOT(diveModeChanged(int)));
 	QString CSSSetSmallLabel = "QLabel { color: mediumblue; font-size: " +                           // Using label height
 		QString::number((int)(0.5 + ui->diveHeadingLabel->geometry().height() * 0.66)) + "px;}"; // .. set CSS font size of star widget subscripts
-#if defined(Q_OS_WIN)
```
dirkhh, line 41: so if the suspicion was that this check is inverted (i.e., that the other style sheet is needed on Windows), have you given that a try (err, do you have access to a Windows system to try it)?
I need to start up my Windows VM, deal with an hour of mandatory updates and then see if one loops better than the other before merging this (yes, I realize that this wasn't used on Windows in the current code - I'm wondering if this would be a net improvement if it WAS used)

dirkhh (commented):
Certainly we want either this or an inversion of the condition.
Give me a moment to try that.

dirkhh:
Ugh. Wow. Apparently I tested the German localization in that VM last time I ran Subsurface there :-)
Anyway, the look without the customization is kinda wasteful and really bad. So that's what we get with this PR (and what's the standard already today). Clearly room for improvement.
The second screenshot is with PR #2754 -- yeah, let's go with yours :-)

(also, the German translations for some of these terms is REALLY verbose and long...)

Screen Shot 2020-04-13 at 10 09 55 AM
Screen Shot 2020-04-13 at 10 10 21 AM

PR #2750, opened 2020-04-13
divelist: when removing/adding a dive unregister/register fulltext

bstoeger:
This fixes a crash: when the undo commands removed a dive from
the list, the fulltext cache was not cleared. If now the divelist
is reset and then the undo-command deleted, deletion of the owned
dive tries to remove it's fulltext cache, which doesn't exist
anymore.

For reasons of symmetry, when readding the dive, its fulltext
has to be registered.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Properly unregister fulltext cache when removing dive from the core. This hopefully fixes a crash reported on mobile.

### DIFF
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -73,6 +73,7 @@ dive *DiveListBase::addDive(DiveToAdd &d)
 	res->hidden_by_filter = true;
 
 	int idx = dive_table_get_insertion_index(&dive_table, res);
+	fulltext_register(res);				// Register the dive's fulltext cache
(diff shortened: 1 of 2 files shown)


janmulder:
Tested this PR, including PRs #2747 and #2746, and I cannot make things crash any more (with was rather easy before).

dirkhh (approved):
Looks reasonable. Thanks for testing @janmulder

PR #2760, opened 2020-04-14
Git storage testing

dirkhh:

- [x] Bug fix
- [x] Functional change
- [x] Code cleanup
- [x] Build system change

This should fix the core issue of why TestGitStorage kept failing for us
The code implicitly relied on `.git/HEAD` pointing to the user's branch in the local cache, yet whether that was the case depended on how / when the local cache was created. And if this was wrong, the user could in fact lose data.

### Changes made:

1) ensure that `.git/HEAD` points to our branch
2) create ten test accounts and use random branch names
3) clean up after the test runs
4) re-enable the test

### Related issues:

need to see if we have some that relate to this

### Release note:

Given that this fixes a potentially bad user visible problem, yes, I should add that

### Mentions:

Not sure who'd be brave enough to look at this...
I guess @torvalds for the git changes
@bstoeger to look at the test (if you are bored)

### DIFF
--- a/.github/workflows/linux-bionic-5.9.yml
+++ b/.github/workflows/linux-bionic-5.9.yml
@@ -37,9 +37,6 @@ jobs:
         bash -x subsurface/scripts/build.sh -desktop -build-with-webkit
 
     - name: test desktop build
-      env:
-        SSRF_USER_EMAIL: ssrftest-u18@hohndel.org
-        SSRF_USER_PASSWORD: geheim-u18
       run: |
         # and now run the tests - with Qt 5.9 we can only run the desktop flavor
         echo "------------------------------------"
--- a/.github/workflows/linux-eoan-5.12.yml
+++ b/.github/workflows/linux-eoan-5.12.yml
@@ -44,9 +44,6 @@ jobs:
         bash -e -x subsurface/scripts/build.sh -mobile
 
     - name: test mobile build
-      env:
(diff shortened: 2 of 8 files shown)


dirkhh:
Ahh. `QRandomGenerator` is Qt 5.10 or newer. And Ubuntu 18.04 is on 5.9.
So I'll need an alternative implementation until we can stop supporting Qt 5.9
lovely

glance-:
rand() or just if-def the test case for qt >= 5.10 , and ignore old bionic. focal is coming on the 23rd and it isn't far until we can bionic users to the app image instead.

dirkhh:
> rand() or just if-def the test case for qt >= 5.10 , and ignore old bionic. focal is coming on the 23rd and it isn't far until we can bionic users to the app image instead.

yeah, I will do an ifdef'ed `rand()` based implementation and keep what I have for Qt >= 5.10
I really like using Qt functionality for things like this as I then know that it's expected to work consistently on all platforms. But I'll do another one until we can use that more broadly.

Even with Focal coming out next week, I expect people to keep using Bionic for at least another year. I seem to remember that people were mad at me when I killed Trusty support in right after Bionic was released... based on that I would say we should expect to keep Bionic support around for another couple of years :-)

bstoeger:
> @bstoeger to look at the test (if you are bored)

Way over my head, sorry. If it works for you, then excellent.

PR #2761, opened 2020-04-14
Git cache branch name

torvalds:
This should fix the issue with the local cached git repo not having the right branch name.

I tested the "local repo was on 'master'" case by hand, but I haven't actually tested the initial creation of an empty repository. I'm expecting that Dirk's git storage tests would catch that one, and this pull request is more of a "if Dirk finds it easier to do a pull than take the patch from my email".

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -703,7 +703,7 @@ static git_repository *create_and_push_remote(const char *localdir, const char *
 {
 	git_repository *repo;
 	git_config *conf;
-	char *variable_name, *merge_head;
+	char *variable_name, *head;
 
 	if (verbose)
 		fprintf(stderr, "git storage: create and push remote\n");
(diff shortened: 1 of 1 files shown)


core/git-access.c
```diff
@@ -664,6 +665,22 @@ static git_repository *update_local_repo(const char *localdir, const char *remot
...
 	}
+
+	/* Check the HEAD being the right branch */
+	if (!git_repository_head(&head, repo)) {
+		const char *name;
+		if (!git_branch_name(&name, head)) {
+			if (strcmp(name, branch)) {
+				char *branchref = format_string("refs/heads/%s", branch);
+
+				report_error("Setting cache branch from '%s' to '%s'", name, branch);
```
dirkhh, line 676: I wonder if this is something where we should show a RED ERROR to the user, or if we shouldn't just silently fix it for them... I mean, what would they do with the knowledge that the cache branch changed?
Imagine Andy and Wendie seeing that message... :-)

dirkhh:
merged via #2760

PR #2764, opened 2020-04-17
Fix CSV for profile export

atdotde:
The function to print a double did not print a comma while
the lines ended in a comma (and now in an empty field).

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/save-profiledata.c
+++ b/core/save-profiledata.c
@@ -13,14 +13,24 @@ static void put_int(struct membuffer *b, int val)
 	put_format(b, "\"%d\", ", val);
 }
 
+static void put_int_with_nl(struct membuffer *b, int val)
+{
+	put_format(b, "\"%d\"\n", val);
+}
+
 static void put_csv_string(struct membuffer *b, const char *val)
 {
 	put_format(b, "\"%s\", ", val);
 }
 
+static void put_csv_string_with_nl(struct membuffer *b, const char *val)
+{
+	put_format(b, "\"%s\"\n", val);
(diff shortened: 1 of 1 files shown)


bstoeger:
Fine with me. The `put_double` change is clearly a bug fix. I'm not sure if there is a difference of ending a line with `,` vs. `,""`...? Note: I'm not against it, just wondering whether there is a practical difference.

atdotde:
Regarding the trailing comma: It just looked weird to me. I tried importing to LibreOffice and it had no problem with a trailing comma. But since you asked, I did some research: As quoting is optional for fields not containing commas, spaces or quotation marks, a trailing comma should probably be identical to ,"" (i.e. interpreted as an empty last field). But even as CSV is probably mostly defined by folklore and what various programs can parse, it turns out, there is an RFC: https://tools.ietf.org/html/rfc4180 And it says in section 2.4 "The last field in the record must not be followed by a comma.". So, to adhere with that we could add the "" in the end. 

My searching for this RFC also revealed that Excel (which I always considered the de facto reference implementation of CSV) is an offender often adding multiple trailing commas to lines (and there are many web pages discussing how to get rid of those). 

Anyway, you once more caught me at a hack. And the more proper version is not even more complicated. So, here it is.

bstoeger (approved):
Looks good, thanks.

PR #2765, opened 2020-04-17
SSL override

dirkhh:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

It turns out that (at least) on Android libgit2 doesn't appear to (always?) correctly respond to valid SSL certificates. Or maybe it's our openssl that doesn't get it right? It's rather frustrating to try to figure out what exactly is going on.

The SSL code in Qt correctly rejects an invalid SSL certificate, but libgit2 often tells us that our certificate on the cloud server is wrong. We used to have code in `git-access.c` that was supposed to only override things for our old certificate that we haven't used in years, but a bug in that code made it still work. So this PR brings back the certificate callback which now unapologetically acceps any certificate for our cloud server, relying on the fact that we earlier called `canReachCloudServer()` which will have rejected an invalid certificate.

The PR also includes some long overdue cleanup to make debug printouts work on Android.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: fix false rejection of Subsurface cloud SSL certificate
 Mobile: fix misdetection of Shearwater Petrel 2 on iOS
 Desktop: fix creation of new cylinder types (names couldn't be the start of already existing names)
 Mobile: fix no-cloud to cloud transition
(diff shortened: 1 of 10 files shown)


dirkhh:
I have no idea why that Android build fails

PR #2767, opened 2020-04-18
Reorganise and reword divelog export panel

willemferguson:
Perform slight reorganisation and rewording of the divelog export
panel. Specifically the wording is changed and expanded to make clear
the 2 types of export of profile data. All the CSV export options are
grouped together. With respect to profile export to CSV format, it is
made clear that one option concerns export of dive computer-related
profile data, whereas the other option involves export of computed
dive profile data from the Dive Profile panel.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @atdotde

### DIFF
--- a/desktop-widgets/divelogexportdialog.cpp
+++ b/desktop-widgets/divelogexportdialog.cpp
@@ -84,7 +84,7 @@ void DiveLogExportDialog::showExplanation()
 	if (ui->exportUDDF->isChecked()) {
 		ui->description->setText(tr("Generic format that is used for data exchange between a variety of diving related programs."));
 	} else if (ui->exportCSV->isChecked()) {
-		ui->description->setText(tr("Comma separated values describing the dive profile."));
+		ui->description->setText(tr("Comma separated values describing the dive profile as downloaded from dive computer."));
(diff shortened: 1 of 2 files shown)


dirkhh (commented):
This creates two new strings for the translators, but that's not necessarily a show stopper.
I never use this feature, so I find it hard to come to a clear conclusion how significant / important this change may be...

willemferguson:
In theory there should not be an extra string. I just edited six of the existing strings and moved one item in the .ui file to group all the  CSV exports together. I have no need for this specific facility either. Came across it while doing some work for the user manual. With the present version the existing labels are very confusing and this is an attempt to remove that ambiguity to make the UI a bit more intuitive.

willemferguson:
Why not leave this until after the release?

dirkhh:
They'll still cause noise in the translations, but since they are already there somewhere else, it should be very minimal. If you think that this makes the UI less confusing, I see no reason not to merge this.

PR #2777, opened 2020-04-20
profile: avoid invalid accesses in setup_gas_sensor_pressure

bstoeger:
Since we removed MAX_CYLINDERS, we have to possibility of dives
with no cylinders. In such a case, setup_gas_sensor_pressure()
would do invalid read- and write-accesses. Therefore, return
early in such a case.

Reported-by: Chirana Gheorghita Eugeniu Theodor <office@adaptcom.ro>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes an invalid access, which is probably the cause of a crash reported on the mailing list. Note: I have no time and did not think this through. In any case it shouldn't hurt.

### Changes made:

1) Return early from setup_gas_sensor_pressure() if there are no cylinders.
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -961,10 +961,13 @@ static int same_rounded_pressure(pressure_t a, pressure_t b)
  * tell us what the first gas is with a gas change event in the first sample.
  * Sneakily we'll use a return value of 0 (or FALSE) when there is no explicit
  * first cylinder - in which case cylinder 0 is indeed the first cylinder.
- * We likewise return 0 if the event concerns a cylinder that doesn't exist. */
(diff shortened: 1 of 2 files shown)


bstoeger:
To answer my question raised in the commit message: the other callers of ` explicit_first_cylinder` bail early if there are no cylinders. Nevertheless, I think it is wrong to return 0 from this function if there are no cylinders.

dirkhh (approved):
This seems reasonable to me. Let's merge and see if it fixes the crash reported on the mailing list

PR #2778, opened 2020-04-21
desktop: don't access first data element if no cylinders

bstoeger:
TabDiveInformation::updateProfile() does some statistics via the
per_cylinder_mean_depth function. It passes down arrays with one
entry per cylinder, which are allocated by means std::vector.

To pass the array, the expression "&vector[0]" is used. It seems
like some compilers through an assertion violation if vector
has no elements. They are technically correct in that this is
undefined, but still this appears like very unfriendly behavior.
After all, std::vector should behave just like a dynamic C-array
that is automatically freed, when going out of scope.

Replace the "&vector[0]" by "vector.data()" and don't do the
call if there aren't any cylinders for good measure.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Another possible crash fix concerning the profile.
### Changes made:

- Don't do `&vector[0]` for empty arrays.

### DIFF
--- a/desktop-widgets/divelogexportdialog.cpp
+++ b/desktop-widgets/divelogexportdialog.cpp
@@ -180,7 +180,7 @@ void DiveLogExportDialog::on_buttonBox_accepted()
 					filename.append(".xml");
 				QByteArray bt = QFile::encodeName(filename);
 				std::vector<const dive_site *> sites = getDiveSitesToExport(ui->exportSelected->isChecked());
-				save_dive_sites_logic(bt.data(), &sites[0], (int)sites.size(), ui->anonymize->isChecked());
+				save_dive_sites_logic(bt.data(), sites.data(), (int)sites.size(), ui->anonymize->isChecked());
 			}
 		} else if (ui->exportImageDepths->isChecked()) {
 			filename = QFileDialog::getSaveFileName(this, tr("Save image depths"), lastDir);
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
This looks reasonable - kinda silly to have to do that... I wonder what compiler flag on Fedora might trigger this...

bstoeger:
> I wonder what compiler flag on Fedora might trigger this...

Well, we don't have confirmation that this is the actual culprit. If it is, it must be some debug-setting:

```
      reference
      operator[](size_type __n) _GLIBCXX_NOEXCEPT
      {
        __glibcxx_requires_subscript(__n);
        return *(this->_M_impl._M_start + __n);
      }
```
This is check is compiled if _GLIBCXX_ASSERTIONS is set according to `debug/assertions.h`

I really don't get the point of this check. The [] operator is precisely for when you do *not* want to check the index. And the function signature says that a reference is returned . Sure, it is undefined behavior to return a reference to a non-existing object, however crashing the program in such a case appears to be a quality-of-implementation issue to me.

I wonder if release versions should be compiled with settings that make the program crash on seemingly benign conditions, though.

dirkhh:
I agree with you - but I'll admit that this is outside of my area of expertize (and somewhat outside of my control).

In completely unrelated news, I've gotten some help from the GitHub Actions experts and I now understand why the builds are failing. We are running out of space in those VMs. As it turns out, the free space in those VMs has been slowly declining over time, and because we switched to building the containers through actions one of my hacks to shrink the container size no longer is being applied. So I will need to create new containers and I'll go back to doing that locally so they are less than half the size than the current versions...

At the same time I'll try to switch the Android one to Qt 5.14.2. I have started doing that and that does - surprise, surprise - of course break everything again as it requires a newer NDK which no longer uses the same paths and there are a bunch of other changes that break our build script...

Issue #2329, opened 2019-10-21
Add dive mode to filter function


plxflx:
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [X] New feature request
- [ ] Discussion request

### Issue long description:
Can we please add a "Dive mode" field to the dive filter function?

### Operating system:
Windows and Ubuntu

dirkhh:
I haven't seen a lot of requests for this. This would not be excessively hard to do, but it will clutter up the filter screen a bit. And I think the vast majority of users will never use this as most people only ever dive OC.
I'm not going to close this just yet. If we get more requests on this issue, I'll consider it. But right now I'm leaning towards 'no'.

plxflx:
I think its doable. I guess the majority of the user base are OC divers but as there is the options for CCR and free dives (and pSCR) I think this should be supported all the way. Further CCR training often requires a certain number of hours on the unit. The ability to filter this out make it possible to show this to an instructor. The same goes for continued cave/mine diving training.

A simple layout would be to just add the same drop down from log header to filter layout and just add the option of 'all'.

adamel:
This would be very useful.

bstoeger:
Should be in release (even though UI is not perfect). Closing.

PR #2779, opened 2020-04-21
Add dive mode filtering

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Per request, add dive-mode filtering. The new filter widget is becoming slightly bloated.

### Changes made:

1) Add filter option for dive mode
2) set maximum time seven days in the future to avoid confusing behavior while testing: manually added dives don't show up.
### Related issues:

Implements #2329
### Release note:

Yes, should probably do.
### Documentation change:

Not sure?
### Mentions:

requested by @plxflx and @adamel

### DIFF
--- a/core/divefilter.cpp
+++ b/core/divefilter.cpp
@@ -349,6 +349,10 @@ bool DiveFilter::showDive(const struct dive *d) const
 	if (!filterData.planned && !has_planned(d, false))
 		return false;
 
+	// Dive mode
+	if (filterData.diveMode >= 0 && d->dc.divemode != (divemode_t)filterData.diveMode)
+		return false;
+
 	return true;
 }
 
--- a/core/divefilter.h
+++ b/core/divefilter.h
@@ -101,6 +101,7 @@ struct FilterData {
 	StringFilterMode equipmentStringMode = StringFilterMode::SUBSTRING;
 	bool logged = true;
 	bool planned = true;
+	int diveMode = -1; // -1: don't filter, >= 0: corresponds to divemode_t
 };
 
 class DiveFilter {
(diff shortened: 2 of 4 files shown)


dirkhh (approved):
I will admit that I struggle with the filter widget and the insane complexity and user hostility that it provides.
And that's not the fault of this PR, but that's a failure of vision when we first engaged down this road.
I personally find it completely unusable and find that I have pretty much stopped using it because it's so confusing and complex.
But then, maybe I'm not the target audience, anyway.

dirkhh:
In unrelated news, I'm on day three of trying to figure out a solution to the Android build failure.
What I have learned is that Qt 5.14 is a complete nonstarter for Android.
It breaks so many things it is truly stunning.
And cmake builds for Android with Qt 5.14 appear to work for no one. Even qmake builds seem to be very much hit and miss. So after wasting a day and a half in that direction, I'm now back to trying to get a clean 5.13.2 build going with a massively smaller docker image.

bstoeger:
Idea: If we improve the fulltext search, I hope that we can cover >90% of the use cases just with that. Then show the rest only in an "advanced mode", sorted by relevance. This advanced mode will have to be moved out of its tiny quadrant though.

adamel:
One thing that really annoys me with the search functionality is that the keyword search uses substring match so that when serching for "cave" it matches "cavern".

bstoeger:
That should be fixed for a long time in master: you can change the search mode to "exact" on a field-by-field basis.

dirkhh:
> Idea: If we improve the fulltext search, I hope that we can cover >90% of the use cases just with that. Then show the rest only in an "advanced mode", sorted by relevance. This advanced mode will have to be moved out of its tiny quadrant though.

I think that's a good next step direction.
I'm willing to merge this before 4.9.4 - was that your intention as well?

bstoeger:
Yes, I think this is low-risk and doesn't make the UI worse than it already is.

PR #2780, opened 2020-04-23
[RFC] Presumably trivial dc cleanups

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are some idle cleanups. However, I'm not 100% sure about the last one. It seems that the divecomputer list is not accessed from mobile anyway, so it shouldn't matter..?

### Changes made:

1) Move function to appropriate translation unit
2) fix whitespace
3) clear divecomputer and model name lists automatically when clearing dive file data
### Release note:

No.
### Documentation change:

No.

### DIFF
--- a/core/device.h
+++ b/core/device.h
@@ -14,6 +14,7 @@ extern void set_dc_deviceid(struct divecomputer *dc, unsigned int deviceid);
 extern void create_device_node(const char *model, uint32_t deviceid, const char *serial, const char *firmware, const char *nickname);
 extern void call_for_each_dc(void *f, void (*callback)(void *, const char *, uint32_t,
 						       const char *, const char *, const char *), bool select_only);
+extern void clear_device_nodes();
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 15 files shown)


bstoeger:
TestParse fails for some reason - I will check that out tomorrow.

bstoeger:
I see - comment in the tests says:
```
        /*
         * DC is not cleared from previous tests with the
         * clear_dive_file_data(), so we do have an additional DC nick
         * name field on the log.
         */
```
However, it doesn't say whether this is intended...?

dirkhh (approved):
Oh my. That's something that we wrote fairly early on (originally).
I have no recall if that was intentionally... maybe there was a use case where you'd do this to keep an existing nickname? Makes no sense to me...

dirkhh:
after sleeping on it, I still can't figure out why this would be the wrong thing to do.
Let's merge.

dirkhh:
(oh -- one thing I noticed... doing multiple merges back to back clearly gets us in trouble with the CI... the way we post releases isn't re-entrant and has several race conditions... a trivial way to avoid some of that is to wait between merges... not ideal... the other option would be to do all the merges manually / locally and doing one push at the end of a series of merges... which I think is what I may switch to)

Issue #2776, opened 2020-04-20
Desktop: select visible dive locations takes very long


janmulder:
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Given my logbook (1230 dives of which about 400 on 1 dive site) it can take 60 seconds to "select visible dive locations" on the map. At first, I thought that I found an endless loop but it simply takes unexpectedly long. 

Make sure the frequently visited dive site is on the map, and select the "select visible dive locations" from the map menu. Running with -v I see that for all matching dives a "Set the current dive site" line is printed. 

### Operating system:
OS: Arch Linux (up to date)
What device are you using? rather normal desktop machine

### Subsurface version:
What version of Subsurface are you running? current master
Does the same happen on another Subsurface version? don't no
Are you using official release, test build, or compiled yourself? self compiled
Provide Git hash if your are building Subsurface yourself. b0dfc0d0ffe7

### Steps to reproduce:
See above

### Current behavior:
Action takes very long.

### Expected behavior:
I would expect that selecting 400 dives out of a list of 1200 dives using the map method would take no more time than (multi-) selecting those same dives directly in the dive list.

atdotde:
My log file is much smaller, so I could not reproduce any significant delays. Maybe you could share your log with me (export and annonymize) by mail?
I played a bit with a profiler, though. It seems, the selection is changed one dive at a time which results in the main window tabs being recalculated quite often. Maybe this is something to look into.

atdotde:
Jan sent me his log file and I ran it through the profiler:
<img width="659" alt="Screenshot 2020-04-25 at 13 14 23" src="https://user-images.githubusercontent.com/863957/80278459-b087fb80-86f6-11ea-8234-c1cc40c7f89b.png">
It seems, 

MapWidgetHelper::selectedDivesChanged

is called almost 60.000 times, each time updating all the information in the main window information tab.

Unfortunately, this is not directly called but through Qt's signalling model so I am lost at tracing the root of this, it might well be inside the QML part of the code. 

@neolit123 @bstoeger @dirkhh you seem to have written these parts of the code. Do you have any idea how to control how often this signal is emitted? Or alternatively, stop the action in the divest until the complete change of selection is done? It seems, the MapWidget already has such a mechanism in the isReady variable that prevents similar things happening at load time. Maybe we can set this to false during the update of the selection?

bstoeger:
[...]

PR #2784, opened 2020-04-25
More selection streamlining

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This removes the peculiar way in which the map would call the dive list instead of the core function to select dives. For me, this makes selecting all visible dive sites of a huge dive-log usable. It's still dog-slow, but at least it doesn't hang from more than a minute!

This needs some serious testing!

### Changes made:

1) An assortment of code cleanups concerning the selection code.

### Related issues:

Hopefully improves #2776. I don't want to say fixes - because it is still dog-slow, but for me it is a distinct improvement.

### Mentions:

@janmulder, @atdotde, @dirkhh: this needs some serious testing - This part of the code is hairy and I want to add that it's only half our fault. The Qt selection code is ominous.


dirkhh (approved):
In my experience the only way this will get testing is by merging it.
I did a local build and can't say that I found anything that this breaks, but - meh... it's not something I ever use...
I'm planning to merge this all locally so we have just one push to the CI machinery

desktop-widgets/divelistview.cpp
```diff
@@ -306,73 +306,6 @@ QList<dive_trip_t *> DiveListView::selectedTrips()
 	return ret;
 }
 
-void DiveListView::selectDive(int i)
```
neolit123, line 309: this code goes back to after the gtk -> qt port and the addition of the marble map widget.

desktop-widgets/mapwidget.cpp
```diff
@@ -94,7 +93,15 @@ void MapWidget::selectionChanged()
 void MapWidget::selectedDivesChanged(const QList<int> &list)
 {
 	CHECK_IS_READY_RETURN_VOID();
-	MainWindow::instance()->diveList->selectDives(list);
```
neolit123, line 97: this was the result of the marble -> qml map port.
at some point we had a compiler macro to smooth switch between the two.

desktop-widgets/divelistview.cpp
neolit123: is this and the plural not uses elsewhere outside of map logic?

desktop-widgets/mapwidget.cpp
```diff
@@ -94,7 +93,15 @@ void MapWidget::selectionChanged()
...
+	// We get a list of dive indices, but the selection code wants a list of dives.
+	// Therefore, transform them here.
+	std::vector<dive *> selection;
+	selection.reserve(list.size());
+	for (int idx: list) {
+		if (dive *d = get_dive(idx))
+			selection.push_back(d);
+	}
+	setSelection(selection, current_dive);
```
neolit123, line 104: this is the first time i see:
https://github.com/Subsurface-divelog/subsurface/blob/7387b10b2944c0126999a3911d6ecf705d243a8b/core/selection.cpp#L151

if this now the way to change currentDive, LGTM.

dirkhh: so you're good with merging this, @neolit123 ?

neolit123: yes

dirkhh:
manually merged

PR #2785, opened 2020-04-26
Another (small) selection speed improvement

bstoeger:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This is another small speed improvement. Sadly not nearly as much as I hoped.
### Changes made:

1) Make sure to not treat programmatical selection changes as user-initiated selection changes.
2) Make functions private to make reasoning easier.
### Related issues:

Slightly improves #2776.

### Mentions:

@janmulder, @atdotde: This also needs testing.

PR #2786, opened 2020-04-26
More selection speed improvements [contains #2785]

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This selects all dives at once, which is a further improvement. Contains #2785. Posting as an extra PR, because the changes are becoming more and more "experimental".

For discussion see #2785, #2776 and the commit message.

### Mentions:

@janmulder, @atdotde.

### DIFF
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -381,7 +381,6 @@ void DiveListBase::finishWork()
 
 void DiveListBase::undo()
 {
-	auto marker = diveListNotifier.enterCommand();
 	initWork();
 	undoit();
 	finishWork();
@@ -389,7 +388,6 @@ void DiveListBase::undo()
 
 void DiveListBase::redo()
 {
-	auto marker = diveListNotifier.enterCommand();
 	initWork();
 	redoit();
 	finishWork();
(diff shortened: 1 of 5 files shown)


bstoeger:
Another one which gives quite a boost in some cases, though not all. I will have to think about the trip-selection problem.

dirkhh (approved):
That is quite a series of changes to the very innards of our code.
A week before a release is not my preferred time for something like this to go in, but frankly, the only way we'll find issues is if people test this - so having it sit on GitHub with zero users and bit rot is the wrong direction.

Are you still adding more changes, or is this a good point to merge, @bstoeger ?

bstoeger:
> Are you still adding more changes

Not here and not today. Fixing the dive-trip disaster is probably much more tricky and should be done separately.

dirkhh:
OK. Then I'll merge this and see if we are closer to having working binaries for all platforms.
I believe the update macOS binaries should now populate (this merge will be the test). The various Linux builds seem to be all working and getting some testing.
The Windows binary is the one where I'm debating whether I should keep it as is (as that is relatively well tested) or whether I should update the Qt version. Right now (given the ridiculous amount of pain and wasted time this has caused on the other platforms) I'm leaning towards updating to a newer Qt after the release...

Issue #2788, opened 2020-04-27
Bogus air cylinders in planner


atdotde:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Start from an empty log, enter the planner. Though not in the cylinder table there are a number of air cylinders. You can see them by clicking on the "used gas" column in the dive planner points table in the selector. Or by adding a waypoint and right clicking on the handle in the profile window. The planner thinks those cylinder are real and tries to switch to air in dives that should not have air cylinders.

### Subsurface version:

Current master

### Mentions:

@bstoeger I guess this is a fallout of the new cylinder handling. To you see immediately where this could come from? Otherwise I will in investigate this further.

bstoeger:
I will check that out later today.

bstoeger:
That said, I couldn't reproduce that easily. Tried:
- Started without log. Went to planner - only saw one air cylinder.
- Started with a test log. Went into planner - got the same gases as the last dive.
- Started with a test log. Closed log. Went into planner - only saw one air cylinder.

In all three cases The plannerpoints and the profile context menu were consistent with the cylinder list.

atdotde:
Just realized: You have to have the surface segment >0. That used to work having air there as a special gas.

bstoeger:
Thanks - I can reproduce this now, will try to debug later.

bstoeger:
The problem was introduced in 7c9f46acd202121e67557bb634961ef17a9f6c1f. A cylinder is added every time the surface interval is changed. This is obviously wrong. However, the fundamental problem is that the old code was pointing to a non-existing cylinder to simulate breathing non-compressed air(?). This cylinder was not shown to the user. Since removing the MAX_CYLINDERS restriction, we obviously don't support this anymore. I will have to think about this. Should we add a cylinder and mark it as "don't show"? Or somehow use a special index meaning "uncompressed air"?

atdotde:
Yes, I thought we had settled on something like this. I would call it "surface air".

atdotde:
Looking at the code, we add a cylinder with cylinder_use = NOT_USED. You think, it's this cylinder?

bstoeger:
Yes. My quick fix was to check if we have an empty air cylinder and only add a new cylinder if that doesn't exist.

However, that still had two problems: the cylinder-model didn't realize that there was a new cylinder. And even if we fix that, the user then sees an ugly empty cylinder in the list.

[...]

PR #2794, opened 2020-04-27
planner: horror-hack: create fake cylinder at end of cylinder table

bstoeger:
When we had fixed-sized cylinder arrays, the planner used the last
empty cylinder for "surface air". This was not recognized by the UI
as a separate cylinder, because "empty cylinder" was the sentinel for
the end of the table. The conversion to dynamically sized cylinder
tables broke this code: everytime the surface segment is changed,
a new dummy cylinder is added, which is visible in the UI.

As a very temporary stop-gap fix, emulate the old code by creating
a cylinder and then setting the end-of-table to before that cylinder.
This means that we have to loosen the out-of-bound checks.

That's all very scary and should be removed as soon as possible.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is supposed to sketch a temporary fix to #2788. I feel very bad for this and would prefer a proper fix. However, that might be too short before release.

### Changes made:

Create dummy cylinder at end of cylinder table for "surface air".

### Related issues:

see discussion in #2788

### Mentions:

@atdotde

### DIFF
--- a/core/planner.c
+++ b/core/planner.c
@@ -272,7 +272,11 @@ static void create_dive_from_plan(struct diveplan *diveplan, struct dive *dive, 
 		if (dp->cylinderid != lastcylid) {
 			/* need to insert a first sample for the new gas */
 			add_gas_switch_event(dive, dc, lasttime + 1, dp->cylinderid);
-			cyl = get_or_create_cylinder(dive, dp->cylinderid);
+			cyl = get_cylinder(dive, dp->cylinderid); // FIXME: This actually may get one past the last cylinder for "surface air".
+			if (!cyl) {
+				report_error("Invalid cylinder in create_dive_from_plan(): %d", dp->cylinderid);
+				continue;
+			}
 			sample = prepare_sample(dc);
 			sample[-1].setpoint.mbar = po2;
 			sample->time.seconds = lasttime + 1;
@@ -1070,12 +1074,12 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 	}
 
 	if (prefs.surface_segment != 0) {
-		// Switch to an empty air cylinder for breathing air at the surface
-		// If no empty cylinder is found, keep using last deco gas
-		cylinder_t cyl = empty_cylinder;
-		cyl.cylinder_use = NOT_USED;
-		add_cylinder(&dive->cylinders, dive->cylinders.nr, cyl);
-		current_cylinder = dive->cylinders.nr - 1;
+		// Switch to an empty air cylinder for breathing air at the surface.
+		// FIXME: This is incredibly silly and emulates the old code when
+		// we had a fixed cylinder table: It uses an extra fake cylinder
+		// past the regular cylinder table, which is not visible to the UI.
+		// Fix this as soon as possible!
+		current_cylinder = dive->cylinders.nr;
 		plan_add_segment(diveplan, prefs.surface_segment, 0, current_cylinder, 0, false, OC);
 	}
 	create_dive_from_plan(diveplan, dive, is_planner);
--- a/core/planner.c
+++ b/core/planner.c
@@ -1074,7 +1074,7 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 		// If no empty cylinder is found, keep using last deco gas
 		cylinder_t cyl = empty_cylinder;
 		cyl.cylinder_use = NOT_USED;
-		add_to_cylinder_table(&dive->cylinders, dive->cylinders.nr, cyl);
+		add_cylinder(&dive->cylinders, dive->cylinders.nr, cyl);
 		current_cylinder = dive->cylinders.nr - 1;
 		plan_add_segment(diveplan, prefs.surface_segment, 0, current_cylinder, 0, false, OC);
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Desktop: fix profile display of planned dives with surface segments
 maps: show the dive site as marker when opening Google Maps
 Mobile: fix false rejection of Subsurface cloud SSL certificate
 Mobile: fix misdetection of Shearwater Petrel 2 on iOS
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1167,7 +1167,7 @@ void RemoveCylinder::undo()
 {
 	for (size_t i = 0; i < dives.size(); ++i) {
 		std::vector<int> mapping = get_cylinder_map_for_add(dives[i]->cylinders.nr, indexes[i]);
-		add_to_cylinder_table(&dives[i]->cylinders, indexes[i], clone_cylinder(cyl[i]));
+		add_cylinder(&dives[i]->cylinders, indexes[i], clone_cylinder(cyl[i]));
 		emit diveListNotifier.cylinderAdded(dives[i], indexes[i]);
 		invalidate_dive_cache(dives[i]); // Ensure that dive is written in git_save()
 	}
--- a/core/cochran.c
+++ b/core/cochran.c
@@ -682,7 +682,7 @@ static void cochran_parse_dive(const unsigned char *decode, unsigned mod,
 			cyl.gasmix.o2.permille = (log[CMD_O2_PERCENT] / 256
 				+ log[CMD_O2_PERCENT + 1]) * 10;
 			cyl.gasmix.he.permille = 0;
-			add_to_cylinder_table(&dive->cylinders, 0, cyl);
+			add_cylinder(&dive->cylinders, 0, cyl);
 		} else {
 			dc->model = "Commander";
 			dc->deviceid = array_uint32_le(buf + 0x31e);	// serial no
@@ -692,7 +692,7 @@ static void cochran_parse_dive(const unsigned char *decode, unsigned mod,
 				cyl.gasmix.o2.permille = (log[CMD_O2_PERCENT + g * 2] / 256
 					+ log[CMD_O2_PERCENT + g * 2 + 1]) * 10;
 				cyl.gasmix.he.permille = 0;
-				add_to_cylinder_table(&dive->cylinders, g, cyl);
+				add_cylinder(&dive->cylinders, g, cyl);
 			}
 		}
 
@@ -739,7 +739,7 @@ static void cochran_parse_dive(const unsigned char *decode, unsigned mod,
 			cyl.gasmix.he.permille =
 				(log[EMC_HE_PERCENT + g * 2] / 256
 				+ log[EMC_HE_PERCENT + g * 2 + 1]) * 10;
-			add_to_cylinder_table(&dive->cylinders, g, cyl);
+			add_cylinder(&dive->cylinders, g, cyl);
 		}
 
 		tm.tm_year = log[EMC_YEAR];
--- a/core/dive.c
+++ b/core/dive.c
@@ -227,7 +227,10 @@ struct event *add_event(struct divecomputer *dc, unsigned int time, int type, in
 void add_gas_switch_event(struct dive *dive, struct divecomputer *dc, int seconds, int idx)
 {
 	/* sanity check so we don't crash */
-	if (idx < 0 || idx >= dive->cylinders.nr) {
+	/* FIXME: The planner uses a dummy cylinder one past the official number of cylinders
+	 * in the table to mark no-cylinder surface interavals. This is horrendous. Fix ASAP. */
+	//if (idx < 0 || idx >= dive->cylinders.nr) {
+	if (idx < 0 || idx >= dive->cylinders.nr + 1 || idx >= dive->cylinders.allocated) {
 		report_error("Unknown cylinder index: %d", idx);
 		return;
 	}
--- a/core/dive.c
+++ b/core/dive.c
@@ -339,14 +339,16 @@ enum divemode_t get_current_divemode(const struct divecomputer *dc, int time, co
 
 struct gasmix get_gasmix_from_event(const struct dive *dive, const struct event *ev)
 {
-	struct gasmix dummy = gasmix_air;
 	if (ev && event_is_gaschange(ev)) {
 		int index = ev->gas.index;
+		// FIXME: The planner uses one past cylinder-count to signify "surface air". Remove in due course.
+		if (index == dive->cylinders.nr)
+			return gasmix_air;
 		if (index >= 0 && index < dive->cylinders.nr)
 			return get_cylinder(dive, index)->gasmix;
 		return ev->gas.mix;
 	}
-	return dummy;
+	return gasmix_air;
 }
 
 // we need this to be uniq. oh, and it has no meaning whatsoever
--- a/core/equipment.c
+++ b/core/equipment.c
@@ -56,7 +56,7 @@ MAKE_CLEAR_TABLE(weightsystem_table, weightsystems, weightsystem)
 //static MAKE_GET_IDX(cylinder_table, cylinder_t, cylinders)
 static MAKE_GROW_TABLE(cylinder_table, cylinder_t, cylinders)
 //static MAKE_GET_INSERTION_INDEX(cylinder_table, cylinder_t, cylinders, cylinder_less_than)
-MAKE_ADD_TO(cylinder_table, cylinder_t, cylinders)
+static MAKE_ADD_TO(cylinder_table, cylinder_t, cylinders)
 MAKE_REMOVE_FROM(cylinder_table, cylinders)
 //MAKE_SORT(cylinder_table, cylinder_t, cylinders, comp_cylinders)
 //MAKE_REMOVE(cylinder_table, cylinder_t, cylinder)
@@ -144,11 +144,23 @@ cylinder_t clone_cylinder(cylinder_t cyl)
 	return res;
 }
 
+void add_cylinder(struct cylinder_table *t, int idx, cylinder_t cyl)
+{
+	add_to_cylinder_table(t, idx, cyl);
+	/* FIXME: This is a horrible hack: we make sure that at the end of
+	 * every single cylinder table there is an empty cylinder that can
+	 * be used by the planner as "surface air" cylinder. Fix this.
+	 */
+	add_to_cylinder_table(t, t->nr, empty_cylinder);
+	t->nr--;
+	t->cylinders[t->nr].cylinder_use = NOT_USED;
+}
+
 /* Add a clone of a cylinder to the end of a cylinder table.
  * Cloned in means that the description-string is copied. */
 void add_cloned_cylinder(struct cylinder_table *t, cylinder_t cyl)
 {
-	add_to_cylinder_table(t, t->nr, clone_cylinder(cyl));
+	add_cylinder(t, t->nr, clone_cylinder(cyl));
 }
 
 bool same_weightsystem(weightsystem_t w1, weightsystem_t w2)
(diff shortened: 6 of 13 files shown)


atdotde:
I don't think this is the correct approach. I think the problem is in handling the cylinders back and forth between dive and dive plan.

Please give me some more time to debug this.

bstoeger:
> I don't think this is the correct approach.

I am sure that it is not. However, it emulates the old code that did kind of work.

> Please give me some more time to debug this.

Sure - take your time.

atdotde (commented):
After a night of sleep, I came to terms with this approach, maybe it's not so bad after all. But I would do the semantics slightly differently: Let's start from the statement "there is always air at the surface and it comes in an infinite amount". So it makes some sense to have it as a default entry in a cylinder table that is never explicitly shown. I would make the table always contain element [cylinders.nr], i.e. one bigger but that entry understood as read only filled with only 0s. This could then as well be used for surface segments. But I would make it always there, not to be initialised from the planner. You can just refer to it explicitly by making cylinder id = cylinders.nr. 
Yes there is a cost in RAM, reserving space for something that is always 0, but I think it's worth it, so if you have to read form that cylinder, there is something there.

I would like to play with this a little more.

core/planner.c
```diff
@@ -272,7 +272,9 @@ static void create_dive_from_plan(struct diveplan *diveplan, struct dive *dive,
 		if (dp->cylinderid != lastcylid) {
 			/* need to insert a first sample for the new gas */
 			add_gas_switch_event(dive, dc, lasttime + 1, dp->cylinderid);
-			cyl = get_or_create_cylinder(dive, dp->cylinderid);
+			cyl = get_cylinder(dive, dp->cylinderid); // FIXME: This actually may get one past the last cylinder for "surface air".
+			if (!cyl)
+				continue;
```
atdotde, line 278: This should not happen. Silently going to the next dp is not the right thing. There should at least be a warning.

core/planner.c
```diff
@@ -1072,10 +1074,17 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
...
+		// we had a fixed cylinder table: It makes sure that there is at least
+		// one cylinder beyond the used cylinders in the cylinder table.
+		// It then lets the cylinder point past the end of the table. Thus,
+		// this fake-cylinder can be used by the planner, but is not visible to
+		// the UI. Fix this as soon as possible!
 		cylinder_t cyl = empty_cylinder;
 		cyl.cylinder_use = NOT_USED;
 		add_to_cylinder_table(&dive->cylinders, dive->cylinders.nr, cyl);
 		current_cylinder = dive->cylinders.nr - 1;
+		dive->cylinders.nr--;
```
atdotde: As I said, I would vote for always having an extra 0 cylinder at the end, so we are not decreasing the number in every run but simply use that last cylinder.

bstoeger:
I still think this approach is utterly broken. Since we are under time-pressure I can indeed simply always have the cylinder table have one more entry (memory usage is not an issue at all - we used to have 20 cylinders for *every* dive and nobody ever complained). I will push something to that effect in the later afternoon. However, this opens a can of worms:

What happens when the dive is saved and loaded? Are there gas change events to the one-past-last cylinder? What will the git and xml loaders do in such a case? What will fixup_dive() do? We have to be extremely careful not to end up with broken dives on disk.

Ultimately, I really think there should be the possibility to represent gas switch to surface air. In a quick test this seemed to kind-of work. However, there were a few problems. Notably in some parts of the code cylinderid = -1 is interpreted as "use the same cylinder". So either we need a flag or a different magic number. In any case, certainly not for this release.

bstoeger:
Since I found a few seconds, here a short update where there is always an empty cylinder at the end of the array.

As I feared, saving does not work properly, at least to XML:
- Plan a dive with surface segment
- Save, close application load
- Replan dive

The surface segment is now missing, probably because the invalid cylinder id confuses the code.

bstoeger:
> The surface segment is now missing, probably because the invalid cylinder id confuses the code.

Ok, so more testing gives a strange behavior: When replanning a dive, the surface interval is set to the last used value, not the value of the replanned dive.

So for example:
1) plan dive, set surface interval to 10 min
2) plan dive, set surface interval to 20 min
3) replan first dive, surface interval is now 20 min, because that was the last value used in the planner.

Is that intended behavior?

dirkhh:
structurally I really struggle with the idea that we special case a cylinder without having a flag that clearly says "THIS IS NOT A CYLINDER". Which is different from "NOT USED". Having a magic last cylinder will require changes in a ton of places - at least when we make these changes, let's give ourselves a clear way of knowing that this is indeed just a magic cylinder that doesn't actually exist. I believe that doing that would then also allow us to not force this as the last cylinder which would likely make things easier elsewhere?
But that of course means that the model knows about the cylinder and the UI layer and the storage layer know not to show / save the data.

I still wonder if it wouldn't be easier to simply use '-1' (or heck, '-42') to indicate 'surface air' and to instead special case THAT in the deco calculations. That sounds like a much more self contained set of changes.

bstoeger:
> structurally I really struggle with the idea that we special case a cylinder without having a flag that clearly says "THIS IS NOT A CYLINDER". Which is different from "NOT USED". Having a magic last cylinder will require changes in a ton of places - at least when we make these changes, let's give ourselves a clear way of knowing that this is indeed just a magic cylinder that doesn't actually exist. I believe that doing that would then also allow us to not force this as the last cylinder which would likely make things easier elsewhere?

From your message, I'm not 100% sure if you are aware of what this PR does: The point is that this cylinder is _not seen_ by the rest of the code, because it is beyond `cylinders.nr`. This is what the old code _effectively_ did, so this here is probably be the least invasive version that is currently possible. Since the cylinder is not seen by the rest of the system, there is little point in arguing over how we should mark it - it purely virtual. That's why this PR has to weaken some bound-checks.

> I still wonder if it wouldn't be easier to simply use '-1' (or heck, '-42') to indicate 'surface air' and to instead special case THAT in the deco calculations. That sounds like a much more self contained set of changes.

Yes, that's what I'm saying all the time. However, that will be distinctly more invasive than this PR. I have a proof of concept that seems to kind-of work. It will need some auditing though, because some parts of the code treat '< 0' as use the same cylinder, so this might give some skewed results (e.g. wrong cylinder use, etc.)

dirkhh:
> From your message, I'm not 100% sure if you are aware of what this PR does: The point is that this cylinder is _not seen_ by the rest of the code, because it is beyond `cylinders.nr`. This is what the old code _effectively_ did, so this here is probably be the least invasive version that is currently possible. Since the cylinder is not seen by the rest of the system, there is little point in arguing over how we should mark it - it purely virtual. That's why this PR has to weaken some bound-checks.

You are correct. I did not realize that that's what the PR does. Need to re-read.

bstoeger:
> I still wonder if it wouldn't be easier to simply use '-1' (or heck, '-42') to indicate 'surface air' and to instead special case THAT in the deco calculations. That sounds like a much more self contained set of changes.

One problem with that is that we can't use normal gas-change events, because `validate_gaschange()`, which is called during `fixup_dive()` will try to "correct" these event.

One thing we could do is misuse the `SAMPLE_EVENT_SURFACE` event to mean switch to surface air in planner context. The problem is that apparently some dive computers emit these events when the diver briefly breaches the surface and we don't want to interpret these as "stopped using gas".

Or use gas index "-1" as "unknown" and "-2" as "no gas". These will end up in the on-disk logs as "0" and "-1", respectively.

bstoeger:
@atdotde, @dirkhh: If Dirk wants to do a release this week, we will not be able to properly fix this and we need a stop-gap measure. I believe that this PR is the least intrusive that we can achieve in a short time. Still we need to do some auditing, concerning e.g. what is written/loaded to/from XML and git.

I can tone down the comments and commit messages if you prefer. I can also try to summarize (my view of) the problem and different ideas that I have how we can fix it in the long run. However for new, we need a decision of how to proceed in the short term.

dirkhh:
I think this is a reasonable approach. I'm not too worried about the tone anywhere. I'm worried about getting this tested and making sure we don't store nonsense to either git or XML that later bites us. That should be easy to test, I hope. Add some plans with surface intervals, mess around with a few things, store, inspect what was stored.

dirkhh:
I'll ask you to merge this yourself when you think it's ready

atdotde:
I am currently looking at this.

atdotde:
To me, this seems to work. I propose merging for the time being.

bstoeger:
I tried this and two points of note:

1) When saving I get the following events in the XML file: `<event time='21:43 min' type='11' value='21' name='gaschange' cylinder='0' />` at the beginning of the surface segment. Which is arguably wrong, but I'm not sure if that is different from what the code did before.

2) I can easily mess up the profile when replanning a dive. For some reason, I can then switch between dives and the events, but not the profile data is updated!?

bstoeger:
Arghl - never mind. I was testing the wrong branch (my event-changes).

bstoeger:
Ok - next test, this time the correct branch:

- We now indeed write cylinder change events to a non-existing cylinder (one past last). That still raises the question: is this what we did before? (Ubuntu ships with a broken version!)

- The profile is still easy to mess up by replanning a dive:
screenshot

bstoeger:
You don't even need to replan a dive for that bug. A long surface segment is sufficient to trigger it. See attached log: switch between both dives.
[test2.zip](https://github.com/subsurface/subsurface/files/4560548/test2.zip)

dirkhh:
so... is this bug in 4.9.3?
I'm trying to get 4.9.4 out. I'd prefer if we didn't regress, but we also can't make "every bug we find is fixed" a precondition...
grabbing a 4.9.3 build and will test...

dirkhh:
Yep, same broken profile in 4.9.3 on my Mac.

So let's file this as a bug and move on...

bstoeger:
Since you have a 4.9.3 at hand (mine doesn't run since it was linked against the wrong libgit2), could you plan a dive with surface long surface segment and look what gas-change event you get in the XML?

Thanks.

atdotde:
Maybe I am stating the obvious, but what seems to happen with the too long surface interval is that part of another dive is plotted.

dirkhh:
> Since you have a 4.9.3 at hand (mine doesn't run since it was linked against the wrong libgit2), 

I shouldn't laugh... but this is exactly why I like Mac/Windows apps and AppImages. I can go back and run ancient versions of Subsurface with those :-)

> could you plan a dive with surface long surface segment and look what gas-change event you get in the XML?

I get none at all, but maybe I did that wrong...
```
<dive number='3' date='2020-04-30' time='23:40:07' duration='15:53 min'>
[deleted notes]
  <cylinder size='11.094 l' workpressure='206.843 bar' description='AL80' start='206.843 bar' end='126.805 bar' depth='65.73 m' />
  <divecomputer model='planned dive' last-manual-time='20:00 min'>
  <depth max='20.0 m' mean='13.453 m' />
  <surface pressure='1.013 bar' />
  <sample time='0:00 min' depth='0.0 m' pressure='206.843 bar' />
  <sample time='0:52 min' depth='20.0 m' pressure='203.598 bar' />
  <sample time='1:47 min' depth='20.0 m' pressure='198.46 bar' />
  <sample time='6:51 min' depth='20.0 m' pressure='170.147 bar' />
  <sample time='8:55 min' depth='0.0 m' pressure='162.562 bar' />
  <sample time='15:46 min' depth='0.0 m' pressure='150.018 bar' />
  <sample time='17:37 min' depth='15.0 m' pressure='144.128 bar' />
  <sample time='20:00 min' depth='15.0 m' pressure='133.317 bar' />
  <sample time='21:08 min' depth='4.8 m' pressure='129.853 bar' />
  <sample time='22:44 min' depth='0.0 m' pressure='126.805 bar' />
  </divecomputer>
</dive>
```

bstoeger:
> Maybe I am stating the obvious, but what seems to happen with the too long surface interval is that part of another dive is plotted.

Yes, if you resize the window, the correct data is plotted. Strange.

bstoeger:
And when I remove the "new" gas-change event with one-past-last cylinder, the display bug goes away. Therefore, it *is* a regression introduced with this PR. :(

bstoeger:
> I get none at all, but maybe I did that wrong...

Then what I don't understand: if there is no gas-change event, how does the code know that we are at surface air? Does it special-case 0 depth? And if that is the case: why do we have to switch to a virtual cylinder in the first place? I don't get it... Robert?

bstoeger:
Ok, I see - if we don't change to a fake cylinder it keeps consuming gas. Still don't understand how the old code worked. Sigh.

bstoeger:
@dirkhh: The 4.9.3 appimage is already broken with respect to surface intervals. Where can I get older versions?

dirkhh:
> @dirkhh: The 4.9.3 appimage is already broken with respect to surface intervals. Where can I get older versions?

https://subsurface-divelog.org/downloads/
all the way back to 4.5.2 :-)

bstoeger:
Yep - even 4.9.1 is broken:
1) Started with an empty log.
2) Plan dive - add surface interval. Looks good.
3) Accept plan - got the strange profile from the attached screen shot.

4) There is a spurious cylinder in the equipment tab.
5) Save log. There is indeed a second cylinder with "unused" in the log.
6) Restart application and load log. The second cylinder is gone. Profile looks good.
7) Save log. The second cylinder is now gone from the XML file. The funny thing though: the profile looks fine now. No more air consumption in the surface segment...

Thus the question remains: how can we achieve 7) directly without doing cylinder tricks..?

In conclusion, this probably never really worked properly. (Note: this is not supposed to be an excuse for the mess I produced when removing MAX_CYLINDERS - I totally misread the code.)

dirkhh:
does that mean that we should merge this and consider ourselves ready for 4.9.4?
And then fix the longstanding issue over the next few weeks?

bstoeger:
Not yet please - this is better than master but worse than 4.9.3, since it produces 1) rendering errors on the profile and 2) kind-of invalid XML files with cylinder-ids past the last cylinder.

dirkhh:
thanks for that clarification

bstoeger:
> this is better than master but worse than 4.9.3

Well, I was wrong. This appears to produce *exactly* the behavior of earlier versions. With 4.9.1 and the steps above I get the XML file:
```
<dive number='1' date='2020-05-01' time='00:02:55' duration='21:54 min'>
  [...]
  <cylinder size='11.1 l' workpressure='207.0 bar' description='unknown' start='207.0 bar' end='109.727 bar' depth='66.0 m' />
  <divecomputer model='planned dive' last-manual-time='20:00 min'>
  <depth max='15.0 m' mean='14.064 m' />
  <surface pressure='1.013 bar' />
  <event time='21:55 min' type='11' value='21' name='gaschange' cylinder='1' />
  <sample time='0:00 min' depth='0.0 m' pressure='207.0 bar' />
  <sample time='1:00 min' depth='15.0 m' pressure='203.723 bar' />
  <sample time='20:00 min' depth='15.0 m' pressure='114.996 bar' />
  <sample time='20:42 min' depth='10.8 m' pressure='112.549 bar' />
  <sample time='21:54 min' depth='0.0 m' pressure='109.727 bar' />
  <sample time='21:55 min' depth='0.0 m' />
  <sample time='43:54 min' depth='0.0 m' />
  </divecomputer>
</dive>
```
Note that there is an gaschange-event to cylinder 1, which doesn't exist. And if you switch between dives, the profile is broken:


This is therefore probably indeed the right thing to do for now, as it is not more broken than earlier releases (and not too many people complained). Nevertheless, please give me some more time to think about this. At least the profile bug should be fixed...

bstoeger:
Oh please - debugging this is *not* fun.

When I remove the following lines from `calculate_max_limits_new()` in `profile.c`:
```
                /* Make sure we can fit all events */
                ev = dc->events;
                while (ev) {
                        if (ev->time.seconds > maxtime)
                                maxtime = ev->time.seconds;
                        ev = ev->next;
                }
```
This decreases the `maxtime` variable by 1(!) second. This single second has the effect that the profile is properly updated. In return, the surface segment is not shown anymore. Indeed, a few lines up we read:
```
                        if ((depth > SURFACE_THRESHOLD || lastdepth > SURFACE_THRESHOLD || in_planner()) &&
                            s->time.seconds > maxtime)
                                maxtime = s->time.seconds;
```
Thus, apparently surface segments should not be shown?

Thus, my current theory is that the profile gets confused by the fact that `maxtime` is one second inside the last segment, but that the segment extends significantly farther...?

atdotde:
Indeed, surface segments (defined w.rt. SURFACE_THRESHOLD) should not be show for logged dives (apparently, some dc's record those at the end of the dive before turning off). Or put differently, (in my understanding), it's ok if those extend out of the display. This is different in the planner and you could argue about planned dives in the log. But even for those, outside the planner, you see that the heat-map is not proper for the surface segment (it doesn't change with time during the segment as there, we don't do interpolation). I always considered that not worrisome as it wasn't really visible in the log (as opposed to the planner).

atdotde:
What I am currently looking at is, I inserted a
 maxime /= 2;
after the code segment you quoted that increases the maxtime to see if it has any effect besides not making the dive fit the profile widget. It does (it doesn't when you *=2), for example we force going to the surface at maxim... and the artefacts look pretty much like what we are discussing here.

atdotde:
Have you noticed that also the information in the inbox when you move the mouse is corrupted (look for example at the time)?

bstoeger:
> Indeed, surface segments (defined w.rt. SURFACE_THRESHOLD) should not be show for logged dives (apparently, some dc's record those at the end of the dive before turning off). Or put differently, (in my understanding), it's ok if those extend out of the display. 

Understood - thanks for the information.

> Have you noticed that also the information in the inbox when you move the mouse is corrupted (look for example at the time)?

Indeed, a segment that starts before but extends beyond `maxtime` seems to somehow confuse the profile.

atdotde:
I am currently looking at populate_plot_entries in profile.c . I suspect that is not happy when you do weird stuff to maxtime (is in: it stops computing plot entries).

bstoeger:
Yes, it appears like it inherently can't handle the last segment extending much beyond `maxtime`. When allocating the points:
```
        nr = dc->samples + 6 + maxtime / 10 + count_events(dc);
```
It may not even allocate enough points for the last segment, because it is based on maxtime. Thus, I think the calculation of `maxtime` should be changed to include the last sample after the last event...?

bstoeger:
I added a commit, which tries to fix the `maxtime` calculation. Apparently, `maxtime` should be rounded to a segment boundary. This seems to work in a first quick test.

atdotde:
Yes, this looks good. As a bonus, it also gets the heat map for the surface segment right.

bstoeger:
Few, that's a relief - thanks for testing. Hopefully this doesn't break other use cases...!

atdotde:
By the way, I now keep getting the can't find a spot in the dataModel warning. Not exactly sure what causes that.

bstoeger:
Surely a fallout of #2800 - another thing to checkout. Help.

bstoeger:
@dirkhh wanted us to commit this thing when ready, so let's do it.

atdotde:
It seems it happens during animations when flipping between dives. Looking into this but I am not too worried.

atdotde:
Fingers crossed

bstoeger:
> It seems it happens during animations when flipping between dives. Looking into this but I am not too worried.

Interesting. I'm quite sure that always happened, we just haven't seen the message so far.

atdotde:
I agree.

atdotde:
My suspicion is that while animating, we try to figure out the position for events that are no longer there: There is still the event marker (with its time) that is trying figure out where to go but the underlying dive already changed so it cannot determine the corresponding sample.

PR #2797, opened 2020-04-30
Avoid empty cylinder description in git saving

atdotde:
The git parser complains with an error message when
reading an empty cylinder description string. So turn
the empty string into "unknown" upon saving.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/save-git.c
+++ b/core/save-git.c
@@ -144,6 +144,8 @@ static void save_cylinder_info(struct membuffer *b, struct dive *dive)
 		const char *description = cylinder->type.description;
 		int use = cylinder->cylinder_use;
 
+		if (description && !*description)
+			description = "unknown";
 		put_string(b, "cylinder");
 		if (volume)
 			put_milli(b, " vol=", volume, "l");


atdotde:
OK, tests were not happy, maybe this is better.

atdotde:
Not better. I will come back to this later tonight. Need to make dinner.

core/save-git.c
```diff
@@ -144,6 +144,8 @@ static void save_cylinder_info(struct membuffer *b, struct dive *dive)
 		const char *description = cylinder->type.description;
 		int use = cylinder->cylinder_use;
 
+		if (description && !*description)
```
bstoeger, line 147: note: we have the `empty_string()` helper function.

dirkhh:
The reason the tests fail is because we now change the data when written out. So you'll need to make appropriate adjustments to the test.

bstoeger:
I'm sure that I'm missing something, but this appears a bit backwards to me. I see two possibilities:
1) We support empty cylinder descriptions. Then the git parser should not complain.
2) We do not support empty cylinder descriptions. Then this should be handled at a higher level. It would be weird to allow it in XML files but disallow it in git repositories, no?

atdotde:
Forget this, this was fixing a bogon fluctuation bug

atdotde:
It seems, the git parser is fine with having no description but it does not like 
description=""
I somehow managed to end up with the latter, though, but I cannot reproduce this anymore.

bstoeger:
> I somehow managed to end up with the latter, though, but I cannot reproduce this anymore.

It's the same bug as #2788: Add a surface interval and you get a cylinder with empty description. The bug is already in 4.9.3... :(

PR #2798, opened 2020-04-30
desktop: don't replot profile when populating dive information tab

bstoeger:
When the dive mode is changed, the profile has to be replot. This
is by a function of the TabDiveInformation. However, that function
was also executed when populating the tab. Thus, when changing dive,
the profile was plot twice.

Move the profile plotting out of the function. Ultimately, the profile
should listen to the appropriate signals itself.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Avoid double profile reloads for *every* selection change. Changing dive-mode still replots the profile as expected.
### Changes made:

1) Don't replot profile when populating the dive information tab.
### Release note:

No.

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveInformation.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveInformation.cpp
@@ -316,6 +316,7 @@ void TabDiveInformation::divesChanged(const QVector<dive *> &dives, DiveField fi
 	if (!current_dive || !dives.contains(current_dive))
 		return;
 
+	bool replot = false;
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
This looks good.
At one point we drew the plot like six times for a single change. The architecture for that has become so convoluted and weird. So I'm glad that you are dis-entangling that...

I'll look at what's wrong with the Mac build

PR #2799, opened 2020-04-30
Github Actions: don't bundle mac app

dirkhh:
There's no point in doing that since the app directory this creates is broken
on older macOS versions, anyway (and we create a working DMG through a
webhook).

Additionally, lately this has started to fail on GitHub, so let's just rip this
out.

- [x] Build system change

It's silly to create that bundle given that it doesn't work on most systems.
And apparently what used to work to simply create a .zip by handing over a directory is now considered an error so GitHub Actions fail.
Let's just rip that out

### Mentions:

@bstoeger if you rebase on top of this, I bet your PR will pass GitHub Actions again


dirkhh:
genius... the mac build is failing. But now it's a timeout to BackBlaze (where we host our cached Qt install because that's SO MUCH CHEAPER than S3 or Google Storage)

dirkhh:
ok, this is just a simple fix for GitHub actions, merging it myself

PR #2801, opened 2020-05-01
profile: properly update lastgasmix when populating events

bstoeger:
When populating the events of a profile, a pointer to the current
gasmix was passed around to properly calculate isobaric_counterdiffusion.
The DiveEventItem::setupToolTipString() function updated this gasmix
when processing gas change events.

I inadvertently broke the code when replacing gasmix-pointers by
values. We could of course simply revert this part of the commit.
However, the data flow was horrible anyway: for example is supposed
that the setup functions were called in the correct order (i.e.
DiveEventItem::setupToolTipString() is called after all other
functions using the gasmix). Not exactly easy to follow.

Therefore, keep passing around the gasmix as value to make it clear
that the functions don't modify it. Keep the gasmix up-to-date at
the caller's site in ProfileWidget2::plotDive().

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This should fix a bug, which I noted while debugging the profile: Wrong gasses were used when calculating `isobaric_counterdiffusion`. Notably, the current gas wasn't updated since I turned pointers to values. This *looks* right, but checking by someone who actually knows what "isobaric counterdiffusion" is would be great.

### Changes made:

1) update gasmix when populating event items on profile.
### Mentions:

@atdotde

### DIFF
--- a/profile-widget/diveeventitem.cpp
+++ b/profile-widget/diveeventitem.cpp
@@ -195,7 +195,6 @@ void DiveEventItem::setupToolTipString(struct gasmix lastgasmix)
 			name += QString::fromUtf8(mb.buffer, mb.len);
 			free_buffer(&mb);
 		}
-		lastgasmix = mix;
 	} else if (same_string(internalEvent->name, "modechange")) {
 		name += QString(": %1").arg(gettextFromC::tr(divemode_text_ui[internalEvent->value]));
 	} else if (value) {
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -802,6 +802,8 @@ void ProfileWidget2::plotDive(const struct dive *d, bool force, bool doClearPict
 		item->setZValue(2);
 		scene()->addItem(item);
 		eventItems.push_back(item);
+		if (event_is_gaschange(event))
+			lastgasmix = get_gasmix_from_event(&displayed_dive, event);
 		event = event->next;
 	}
 


dirkhh (commented):
Ugh, that looks so harmless, but the commit message makes me want to look at that code in context. Give me a few minutes. I don't doubt that you are correct in your analysis, but I'd love to understand it myself and understand how it got broken...
BTW: it's nice to have a pointer to the commit that broke things (or the beginning of the series that broke it) as that makes it easier to put context around things...

bstoeger:
Breaking commit: 360f07e4533d4ede7ba494724382fc8dbcb4649c

dirkhh:
> Breaking commit: [360f07e](https://github.com/subsurface/subsurface/commit/360f07e4533d4ede7ba494724382fc8dbcb4649c)

August 2018? Wow

dirkhh (approved):
OK, I looked at this a bit more and while I still don't quite understand the error case, I can't see how this could make things worse.

@atdotde - do you have a chance to look at this this evening?

bstoeger:
The error case is that even though you switch gas, isobaric_counterdiffusion is still calculated with the first cylinder's gas. Apparently, this has an effect on the displayed icons and the tooltip of the gas-change events.

However, I don't even know what isobaric counterdiffusion is, so I am certainly the wrong person to judge whether this is fixed now.

dirkhh:
> The error case is that even though you switch gas, isobaric_counterdiffusion is still calculated with the first cylinder's gas. Apparently, this has an effect on the displayed icons and the tooltip of the gas-change events.

Tracing the code some more, I now see this. Thanks.

> However, I don't even know what isobaric counterdiffusion is, so I am certainly the wrong person to judge whether this is fixed now.

So this was introduced by @willemferguson in f072f78c29 - I hope that either he or our resident physicist (@atdotde ) can comment on this some more...

atdotde:
Just sat down to look at this...

atdotde:
This is definitely the right thing to do. The only thing that surprised me was that the gas change event tooltip ends up in the info box. I wasn't aware of that before.

atdotde:
If you wanted to know about isobaric counter diffusion, I could recommend some diving blog that discusses these kinds of things: https://thetheoreticaldiver.org/wordpress/index.php/tag/icd/

atdotde (approved):
Setting lastgasmix and never looking at it again is clearly wrong. Better change it further up the call chain.

dirkhh:
> If you wanted to know about isobaric counter diffusion, I could recommend some diving blog that discusses these kinds of things: https://thetheoreticaldiver.org/wordpress/index.php/tag/icd/

Yeah, but who knows what **THAT GUY** knows about all this... 😀

PR #2802, opened 2020-05-01
minor cleanups concerning profile

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Code hygiene things: Debugging is so much easier when you know that a parameter is not changed, therefore do some const-ification.
[Side-note: sadly `const` is broken in C. If you have something of the kind
```
struct plot_info {
        struct entry *table;
};
```
And you pass in "const *plot_info", then the compiler will not complain if you modify the table-entries. Thus it is much harder to get such const-ifications right, compared to C++.]
### Changes made:

1) Constification
2) Localization
3) Code-hygiene: make a C function explicitly `external`.

### DIFF
--- a/core/display.h
+++ b/core/display.h
@@ -34,7 +34,7 @@ extern int is_default_dive_computer(const char *, const char *);
 
 typedef void (*device_callback_t)(const char *name, void *userdata);
 
-int enumerate_devices(device_callback_t callback, void *userdata, unsigned int transport);
+extern int enumerate_devices(device_callback_t callback, void *userdata, unsigned int transport);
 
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
We are very inconsistent with the 'extern' in declarations. You can usually go back and find the original author (or whose code this was copied from) because some of us tend to write `extern` because they are old and you used to need that, and many of us don't :-)
Anyway, nice cleanups, I'll take it.

PR #2803, opened 2020-05-01
some include disentangling

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Notably, this decouples dive.h and divelist.h. A tiny step in removing interdependencies.

### DIFF
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -7,6 +7,7 @@
 #include "core/divesite.h"
 #include "core/gettextfromc.h"
 #include "core/tag.h"
+#include "core/subsurface-time.h"
 #include "core/file.h"
 #include "core/errorhelper.h"
 #include "core/divefilter.h"
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -3,6 +3,7 @@
 #include <QDir>
 #include <QtConcurrent>
 #include "core/membuffer.h"
+#include "core/dive.h"
 #include "core/divesite.h"
 #include "core/gettextfromc.h"
 #include "core/tag.h"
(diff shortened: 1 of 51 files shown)


dirkhh (approved):
- generally I don't love random changes right before a release
- that said, something like this, if it compiles everywhere (assuming our testing is broad enough) we should be fine... I can't remember if we build `smtk2ssrf` anywhere. I do think I added `export_html` to at least one of the builds
- on top of that - I want to establish a rule that header file changes are ONLY allowed, if they reduce the number of files that include `dive.h` :-)

bstoeger:
> generally I don't love random changes right before a release

Sure, this can wait. Just throwing it out there.

> on top of that - I want to establish a rule that header file changes are ONLY allowed, if they reduce the number of files that include dive.h

Please add "directly or indirectly". If we remove dive.h from a header file, then it can be that numerous translation units now need an additional `#include "dive.h"` even though the total number of TUs that (indirectly) include `dive.h` decreases.

bstoeger:
PS: also removing declarations from `dive.h` can be very useful even though it doesn't reduce the number of `dive.h` dependencies. Because changing respective declaration than doesn't cause a full recompile!

dirkhh:
You took my comment exactly the wrong way. I was NOT criticizing your PR at all. I was trying to make some lighthearted comments.
I think this is a net positive change that you are proposing here.
And I'm merging this -- see my approval. I'm just trying to leave a few minutes between merges as not doing so confuses the way we create artifacts.
Yes, that's sadly broken, but it is what we have right now.

bstoeger:
> You took my comment exactly the wrong way.

Not really - I was just clarifying your statements.

PR #2805, opened 2020-05-02
[Post-Release] Move selection of trips to core

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In analogy to dives, move the management of trip selection to the core. The main reason is that checking whether a single trip is selected can be painfully slow, which in turn slows down updating of the main tab.

For a 3500 dive test file, this PR reduced the maintab update from 130 to 5 ms, when selecting all dives.

Sadly, the first CTRL-A is still very slow. It appears that the proxy-model has to populate its cache or something. I will have to investigate further.

### Changes made:

1) Introduce a trip->selected flag in analogy to dive->selected.
2) Use said flag to improve speed when many dives are selected.

### Release note:

Probably not.
### Documentation change:

No.

### DIFF
--- a/core/selection.cpp
+++ b/core/selection.cpp
@@ -3,12 +3,13 @@
 
 #include "selection.h"
 #include "divelist.h"
-#include "display.h" // for amount_selected
+#include "trip.h"
 #include "subsurface-qt/divelistnotifier.h"
 
 #include <QVector>
 
 int amount_selected;
+static int amount_trips_selected;
 
 extern "C" void select_dive(struct dive *dive)
 {
@@ -155,6 +156,11 @@ void setSelection(const std::vector<dive *> &selection, dive *currentDive)
 	QVector<dive *> divesToSelect;
 	divesToSelect.reserve(selection.size());
 
+	// Since we select only dives, there are no selected trips!
+	amount_trips_selected = 0;
+	for (int i = 0; i < trip_table.nr; ++i)
+		trip_table.trips[i]->selected = false;
+
 	// TODO: We might want to keep track of selected dives in a more efficient way!
 	int i;
 	dive *d;
@@ -230,4 +236,20 @@ extern "C" void select_newest_visible_dive()
 
 	// No visible dive -> deselect all
 	select_single_dive(nullptr);
+}
+
+extern "C" void select_trip(struct dive_trip *trip)
+{
+	if (trip && !trip->selected) {
+		trip->selected = true;
+		amount_trips_selected++;
+	}
+}
+
+extern "C" void deselect_trip(struct dive_trip *trip)
+{
+	if (trip && trip->selected) {
+		trip->selected = false;
+		amount_trips_selected--;
+	}
 }
--- a/core/selection.cpp
+++ b/core/selection.cpp
@@ -253,3 +253,16 @@ extern "C" void deselect_trip(struct dive_trip *trip)
 		amount_trips_selected--;
 	}
 }
+
+extern "C" void clear_selection(void)
+{
+	current_dive = nullptr;
+	amount_selected = 0;
+	amount_trips_selected = 0;
+	int i;
+	struct dive *dive;
+	for_each_dive (i, dive)
+		dive->selected = false;
+	for (int i = 0; i < trip_table.nr; ++i)
+		trip_table.trips[i]->selected = false;
+}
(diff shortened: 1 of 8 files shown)


core/selection.cpp
```diff
@@ -231,3 +237,19 @@ extern "C" void select_newest_visible_dive()
 	// No visible dive -> deselect all
 	select_single_dive(nullptr);
 }
+
+extern "C" void select_trip(struct dive_trip *trip)
```
dirkhh, line 241: so in this commit I only see a call to `deselect_trip()` and no calls to `select_trip()` which makes me wonder... but most likely that's simply in the next commits?

desktop-widgets/divelistview.cpp
```diff
@@ -566,9 +569,11 @@ void DiveListView::selectionChanged(const QItemSelection &selected, const QItemS
 		const QAbstractItemModel *model = index.model();
 		struct dive *dive = model->data(index, DiveTripModelBase::DIVE_ROLE).value<struct dive *>();
 		if (!dive) { // it's a trip!
+			dive_trip *trip = model->data(index, DiveTripModelBase::TRIP_ROLE).value<dive_trip *>();
+			select_trip(trip);
```
dirkhh, line 555: ahhh, here it is

core/selection.cpp
```diff
@@ -254,6 +254,17 @@ extern "C" void deselect_trip(struct dive_trip *trip)
...
 
+extern "C" struct dive_trip *single_selected_trip()
+{
+	if (amount_trips_selected != 1)
+		return NULL;
+	for (int i = 0; i < trip_table.nr; ++i) {
+		if (trip_table.trips[i]->selected)
+			return trip_table.trips[i];
+	}
+	return NULL; // shouldn't happen
```
dirkhh, line 266: print out a warning?

core/selection.cpp
```diff
@@ -231,3 +237,19 @@ extern "C" void select_newest_visible_dive()
 	// No visible dive -> deselect all
 	select_single_dive(nullptr);
 }
+
+extern "C" void select_trip(struct dive_trip *trip)
+{
+	if (trip && !trip->selected) {
+		trip->selected = true;
+		amount_trips_selected++;
```
dirkhh, line 245: in our UI logic, selecting a trip always implies selecting all the dives in the trip.
Would it make sense to do this hear, or would this confuse other layers and we really just worry about the helper data structure for trips?
As I write the question, I think I know the answer... :-)

dirkhh (approved):
I think this looks good.
I'm happy to see you work on all these post-release things already...
I cannot wait to get this release out, deal with the almost guaranteed random breakage that we didn't catch (because... 1500+ commits), and move forward with cool new stuff.

core/selection.cpp
bstoeger: Yes, that was the original idea. However, I'm not sure about the correct way to implement this, so first I wanted to start with the "easy" version. The problem is that we get a loop:

UI selection change -> dive list -> Core selects more items -> dive list

We end up with a selection change in the divelist, while a selection change is active. What we could do is
1) a "apply_selection_change()" function that returns a list of additional items that the dive list has to select.
2) simply intercepting all UI selection changes, sending them to the core and let the core rebuild the selection. I.e. never apply selections directly.

dirkhh: I can see the advantage of the much simpler approach to not deal with causing new selections

dirkhh:
OK, this is good to go - unless you want to add the warning that I suggested.
And I'll wait for #2758

core/selection.cpp
bstoeger: done.

dirkhh: it must be my day to be blind... I don't see it...

bstoeger: Forgot "-a" in "commit --amend".

dirkhh:
This is so frustrating. TestGitStorage failed. But only in the mobile build on Ubuntu.
No useful errors. Works in the other builds. And anyway, this shouldn't affect anything in the storage test.
I try to rerun the test, and get the (%*&@(#% this isn't a tree error again. What the heck.

PR #2806, opened 2020-05-02
Add test for profile data

atdotde:
There was a TestProfile but that was a stub that did not test anything.

We have an export function that serialises the profile data
(including lots of derived data like deco information and
cylinder pressure interpolation). So here is now a simple
tests that can detect regressions in the profle.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### Additional information:

This adds 10MB of reference profile data. This should test a lot of our features. We could save space by just checking against a check sum but this would not tell us what changed.

### DIFF
--- a/dives/abitofeverything.ssrf
+++ b/dives/abitofeverything.ssrf
@@ -1,0 +1,11585 @@
+<divelog program='subsurface' version='3'>
+<settings>
+<divecomputerid model='Uwatec Galileo Sol' deviceid='00abcdef' nickname='Uwatec Galileo Sol (abcdef)'/>
+<divecomputerid model='Heinrichs Weikamp OSTC 3' deviceid='01234567' nickname='Heinrichs Weikamp OSTC 3 (1234567)'/>
+<divecomputerid model='Aeris A300CS' deviceid='2b012345' serial='035025'/>
+<divecomputerid model='Heinrichs Weikamp OSTC 3' deviceid='4ee3d5e1' nickname='Heinrichs Weikamp OSTC 3 (4ee3d5e1)'/>
+<divecomputerid model='Shearwater Petrel' deviceid='64cb6f5d' serial='39384572' firmware='28'/>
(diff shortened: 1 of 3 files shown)


dirkhh (commented):
I think that's a really cool idea. Thank you.
As for the size of the data set... would it make sense to maybe use just a couple of dives that are fairly extreme (long, deep, deco)? I know that git compresses the data, so this shouldn't really add 10MB, but it seems like it would make sense to have something slightly targeted here instead of a bunch or "random" dives?

dirkhh:
Having TestGitStorage fail is not a good sign.
Can you reproduce this locally?

atdotde:
This is now with a somewhat smaller reference file (that contains a few more special cases). I hope this covers pretty much all the features of the profile.

TestGitStorage fails for me locally but does so also for current master:
`FAIL!  : TestGitStorage::cleanupTestCase() Compared values are not the same
   Actual   (QString(branch)): "https://cloud.subsurface-divelog.org/git/gitstorage1@hohndel.org[52004ec8a8f8]"
   Expected (randomBranch)   : "52004ec8a8f8"
   Loc: [/Users/Helling/src/subsurface/tests/testgitstorage.cpp(59)]`

while there is a comment in the source saying 
'	// this odd comparison is used to tell that we were able to connect to the remote repo;
	// in the error case we get the full cloudTestRepo name back as "branch"'
which seems to indicate there is some connection problem.

dirkhh:
Yes, that indeed means that there is again something wrong with connecting to cloud storage.
Perfect for the day I released a new version...
I'll look into this some more.

dirkhh:
storage server ran out of inodes in its /tmp dir.
So TestGitStorage runs again locally for me.
I don't understand why trying to restart these tests fails - the error I see there is rather odd (it can't check out the sources)

dirkhh:
OK, I looked at this some more and I think it's a great idea

PR #2807, opened 2020-05-03
desktop: ask user when quitting application while planning

### LINKED ISSUE #1078: Sigdev crash when exiting from dive planner
Crash on non-intended exit from dive planner

### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Current behavior:

V4.7.6 on Ubuntu 16.04
To leave the dive planner after doing some planning, one either has to cancel the dive plan or save it. If one by accident closes the application by clicking the close window icon on the window frame, the program crashes:
_subsurface: Fatal IO error 2 (No such file or directory) on X server :0.
Segmentation fault (core dumped)_
This is not ideal at all. There used to be a dialog box with a message that said "Save your data or cancel the dive plan" or something similar.

### Expected behavior:

Provide a dialog box with a warning or exit gracefully.

bstoeger:
We used to cancel a plan (or profile edit) when the user quit
the application while planning. This is inconsistent with
respect to closing or opening a different log, where the user
was asked for confirmation.

Thus, for consistency and to avoid loss of a planned dive,
use the okToClose() function in on_actionClose_triggered() of
MainWindow. As an added bonus, this saves a few SLOC.

Fixes #1078

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a UI issue reported by @willemferguson in #1078. The crash of the report I couldn't reproduce though (on Kubuntu 20.04).
### Changes made:

1) Ask user instead of canceling a plan when closing the application.

### Related issues:

Fixes #1078.

### Release note:

I don't think so.

Willem, please let me know if this behavior is what you had in mind.

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -741,21 +741,9 @@ void MainWindow::on_actionPreferences_triggered()
 
 void MainWindow::on_actionQuit_triggered()
 {
-	if (mainTab->isEditing()) {
-		mainTab->rejectChanges();
-		if (mainTab->isEditing())
-			// didn't discard the edits
-			return;
-	}
-	if (DivePlannerPointsModel::instance()->currentMode() != DivePlannerPointsModel::NOTHING) {
-		DivePlannerPointsModel::instance()->cancelPlan();
-		if (DivePlannerPointsModel::instance()->currentMode() != DivePlannerPointsModel::NOTHING)
-			// The planned dive was not discarded
-			return;
-	}
-
-	if (unsavedChanges() && (askSaveChanges() == false))
+	if (!okToClose(tr("Please save or cancel the current dive edit before quiting the application.")))
 		return;
+
 	writeSettings();
 	QApplication::quit();
 }


dirkhh (approved):
Of course I immediately worry WHY we did it this way, but yeah, this change makes sense.
Not quite sure if this would indeed fix what looked more like a Ubuntu Qt bug, but I assume you tested that somehow?

dirkhh:
I tried to restart the GitHub Actions (as the cloud server is fixed), and get `reference is not a tree: da0fe4bfbc294b524dbc9af8d7c27ba6d68ef21b`
Something odd going on with GitHub, I guess

bstoeger:
Yes, I tested this before and after and couldn't reproduce the crash [Kubuntu 20.04].

It appeared to me that the main complaint was not so much the crash (the worst thing in that case is that some settings are not saved), but the loss of the plan on accidental hitting the close-icon:

> This is not ideal at all. There used to be a dialog box with a message that said "Save your data or cancel the dive plan" or something similar.

That I could reproduce and it now doesn't happen anymore.

dirkhh:
I'm ok merging this, but as requested will wait to go with #2758 first...

bstoeger:
> I'm ok merging this, but as requested will wait to go with #2758 first...

Please don't wait for #2758 with this or other PRs. I was only ranking by level of interest not by some kind of dependency scheme.

We might wait for @willemferguson's opinion, though.

dirkhh:
I think it makes sense to merge, then. We can always tweak if this isn't sufficient for @willemferguson 's issue

PR #2819, opened 2020-05-05
just testing if we once again catch missing DCO

dirkhh:
nothin to see here


dirkhh:
yes, the DCO bot works. Great.

PR #2815, opened 2020-05-05
Update cylinder UI fields when cylinders are edited

bstoeger:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This fixes #2814 and related bugs.

### Changes made:

1) update dive->sac when cylinders are edited
2) update dive list when cylinders are edited
3) update information tab when cylinders are edited
4) update statistics tab when cylinders are edited
### Related issues:

Fixes #2814.

### Release note:

done.

### Documentation change:

no.

Issue #2814, opened 2020-05-04
Looks like SAC isn't updated in the dive list itself


dirkhh:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

From Scubaboard:

> Hi @dirkhh. I just noticed a minor bug in the latest desktop version (4.9.4) in windows 10.
>
> Once I add tank and start/end pressures for a new dive, the calculated SAC value is shown in the corresponding area in the Information Tab, but it is not shown in the list with the dives.

more details and screenshot here: https://www.scubaboard.com/community/threads/subsurface-4-9-4-has-been-released.595025/page-2#post-8999551

### Operating system:

Windows 10

### Subsurface version:

4.9.4 release binaries

### Current behavior:

SAC in the dive list table doesn't update until restart

### Expected behavior:

should update once tank info is filled in

### Additional information:

I have NOT attempted to reproduce this, yet - just posting a bug that was posted on Scubaboard so it doesn't get lost

PR #2820, opened 2020-05-05
[RFC] Change control-flow when resetting core data

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The way the application is reset on loading / closing irked me for quite some time. This tries to disentangle inter-module dependencies by creating a "data reset" signal that is emitted by the core. The various modules should then reset themselves.

On desktop at least a quick test seemed to work fine.
### Changes made:

1) Use common signal to reset models.
2) Call into core directly to reset data. Don't go via models.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -835,6 +835,9 @@ void process_loaded_dives()
 	autogroup_dives(&dive_table, &trip_table);
 
 	fulltext_populate();
+
+	/* Inform frontend of reset data. This should reset all the models. */
+	emit_reset_signal();
 }
 
 /*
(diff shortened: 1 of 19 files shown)


dirkhh (commented):
Talk about an invasive change... :-)
Yeah, I think this is the right thing to do, it will need a lot of testing - usually the only way to get this is to merge it. But with this I want to do at least some testing on mobile (and especially on device) to be at least somewhat confident that this doesn't completely break things...
BTW: I don't think it's related at all, but have you seen #2814 ? That's another place where apparently a change in the core isn't propagating to a widget...

bstoeger:
> BTW: I don't think it's related at all, but have you seen #2814 

#2815

dirkhh (approved):
Spent more time staring at this. I believe this is the right thing to do.

dirkhh:
happy to merge this as well - can do the merge locally if you want.

bstoeger:
rebase on master

dirkhh:
GitHub is unwell - I need to reach out to them again. This is happening more and more frequently. Source checkout fails with a bogus error that the SHA isn't a tree

PR #2822, opened 2020-05-05
Remove Bühlmann factor cache

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

While looking at the `inertgas` enum I stumbled over this cache and had my doubts that a thread-synchronized hashmap would be significantly faster than an `exp(...)` function. Indeed, after making the function global and calling it from a different translation unit with
```
        QElapsedTimer t;
        t.start();
        double res = 0.0;
        for (int i = 0; i < 1'000'000; ++i) {
                for (int ci = 0; ci < 16; ++ci) {
                        res += factor(200, ci, HE);
                        res += factor(200, ci, N2);
                }
        }
        printf("elapsed: %llu res: %lf\n", t.elapsed(), res);
```
I got a speed increase from 604 ms to 266 ms by removing the cache. Moreover, note that these are 32 million calls that take only 266 ms. It therefore seems that optimizing this particular function is not worth it.

### Changes made:

1) remove Bühlmann factor cache.
2) move `inertgas` enum to `deco.c` source file.

### Release note:

No user visible change.

### DIFF
--- a/core/deco.c
+++ b/core/deco.c
@@ -30,6 +30,7 @@
 #include "planner.h"
 #include "qthelper.h"
 
+enum inertgas { N2, HE };
 #define cube(x) (x * x * x)
 
 // Subsurface until v4.6.2 appeared to produce marginally less conservative plans than our benchmarks.
--- a/core/deco.c
+++ b/core/deco.c
@@ -290,18 +290,10 @@ double tissue_tolerance_calc(struct deco_state *ds, const struct dive *dive, dou
 }
 
 /*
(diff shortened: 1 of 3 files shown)


atdotde (approved):
When the fixed one second cache was introduced that gave a significant speed up. I agree that without the caching the UI responsiveness even for stupidly long deco dives is at least not noticeably worse. So let's do this simplification.

PR #2823, opened 2020-05-06
desktop: avoid spurious undo commands for date/time editing

bstoeger:
The date and time fields of the main tab posted undo events
for every date/timeChanged signal. Thus, when changing the
day of the month to e.g. 21, this would result in two date
change events: one to the 2nd and one to the 21st. This is
very irritating.

Instead listen to editingFinished() events, which thankfully
exist for these widgets.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a UI nuisance, see commit description.

PS: I dislike the on_*() methods as the signal signature is only runtime checked.
### Release note:

I guess not.

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -607,23 +607,23 @@ static void shiftTime(QDateTime &dateTime)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
that looks great.

PR #2824, opened 2020-05-06
Remove displayed_dive dependency of ExtraDataModel

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The main goal is to display data of current_dive instead of displayed_dive for consistency and so that we can move displayed_dive to the planner. Thus, there is no risk of forgetting to copy changes to displayed_dive.

While doing so, perform a few assorted cleanups for consistency with the other model code.

### Release note:

No - should not be user visible.

### DIFF
--- a/qt-models/divecomputerextradatamodel.cpp
+++ b/qt-models/divecomputerextradatamodel.cpp
@@ -4,8 +4,7 @@
 #include "core/metrics.h"
 
 
-ExtraDataModel::ExtraDataModel(QObject *parent) : CleanerTableModel(parent),
-	rows(0)
+ExtraDataModel::ExtraDataModel(QObject *parent) : CleanerTableModel(parent)
 {
 	//enum Column {KEY, VALUE};
 	setHeaderDataStrings(QStringList() << tr("Key") << tr("Value"));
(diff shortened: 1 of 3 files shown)


qt-models/divecomputerextradatamodel.cpp
```diff
@@ -20,7 +20,7 @@ void ExtraDataModel::clear()
 
 QVariant ExtraDataModel::data(const QModelIndex &index, int role) const
 {
-	struct extra_data *ed = get_dive_dc(&displayed_dive, dc_number)->extra_data;
+	struct extra_data *ed = dc ? dc->extra_data : nullptr;
```
dirkhh: what happens to us if the `dc` pointer becomes invalid? Do we know that any operation that could reallocate our dive table clears the model? Specifically, I don't see any code that resets the `dc` variable to nullptr.

dirkhh (commented):
Maybe I don't understand the lifecycle of this `ExtraDataModel`, but it looks to me like we are creating dangling pointers here...

dirkhh (approved):
so assuming we do the right thing and call `clear()` when things change (which is what you have been diligently working on), this looks good now.

bstoeger:
In principle yes - however, I will change this to cache the QStrings. Thus, no potential of dangling pointers.

dirkhh:
> In principle yes - however, I will change this to cache the QStrings. Thus, no potential of dangling pointers.

cool, so I'll hold off on merging.

bstoeger:
Data now cached.

dirkhh (approved):
I really like this approach.

PR #2826, opened 2020-05-08
add Berthold to the list of explicitly mentioned developers

dirkhh:
Number 3 overall committer with currently 9% of total commits and 27% of all
commits in the last three years.

- [x] Code cleanup
- [x] Documentation change

Should be entirely obvious.

### Mentions
@bstoeger since you never use the Umlaut, I assume you prefer the 'oe'?


bstoeger:
Thanks. I somehow feel that most of the commits are code-reshuffling of stuff that is already there and not something substantially new.

> since you never use the Umlaut, I assume you prefer the 'oe'?

When it comes to co-authorship, I usually use the German spelling with Umlaut - it's just that many of the American and UK publishers and databases butcher it. My e-mail is configured to "oe" because experience shows that non-ASCII still gives problems with many clients. Or at least it used to until recently. I have no strong preference of one-way over the other.

PR #2828, opened 2020-05-11
Mobile tags

dirkhh:

- [x] Functional change
- [x] New feature

Calling it a 'new feature' seems a bit excessive - but yeah. We already allowed the user to filter by tags, but we didn't even show tags. That makes no sense.

### Changes made:

- improve vertical 'density' of the DiveDetails - there was just too much empty space
- make tags string available to mobile UI
- add a row to the DiveDetails (but combined with the vertical densification, the overall space needed stayed about the same

### Release note:

done

### Documentation change:

yeah, this will require new pictures and maybe a word or two in the manual

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,53 +1,12 @@
+Mobile: show dive tags on dive details page
 Desktop: update SAC fields and other statistics when editing cylinders
 Desktop: Reconnect the variations checkbox in planner
 Desktop: add support for dive mode on CSV import and export
 Desktop: fix profile display of planned dives with surface segments
-maps: show the dive site as marker when opening Google Maps
-Mobile: fix false rejection of Subsurface cloud SSL certificate
-Mobile: fix misdetection of Shearwater Petrel 2 on iOS
-Desktop: fix creation of new cylinder types (names couldn't be the start of already existing names)
-Mobile: fix no-cloud to cloud transition
-Mobile: remove locking of data storage access
-Mobile: performance improvements for dive list
-Mobile: user notification of progress during startup
-Mobile: ensure filter text field has focus when opened
-Mobile: dive summary: make time drop downs smaller
-Mobile: close menus with Android back button on dive details
-Mobile: add ability to toggle dive invalid flag from dive details context menu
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,53 +1,7 @@
-Desktop: update SAC fields and other statistics when editing cylinders
-Desktop: Reconnect the variations checkbox in planner
-Desktop: add support for dive mode on CSV import and export
-Desktop: fix profile display of planned dives with surface segments
```
bstoeger, line 4: The top four deleted items are bug fixes, which are not in current release.

qt-models/divetripmodel.cpp
```diff
@@ -222,6 +222,7 @@ QVariant DiveTripModelBase::diveData(const struct dive *d, int column, int role)
 	case MobileListModel::SumWeightRole: return get_weight_string(weight_t { total_weight(d) }, true);
 	case MobileListModel::DiveMasterRole: return d->divemaster ? d->divemaster : QString();
 	case MobileListModel::BuddyRole: return d->buddy ? d->buddy : QString();
+	case MobileListModel::TagsRole: return d->tag_list ? get_taglist_string(d->tag_list) : QString();
```
bstoeger: `get_taglist_string()` should support empty lists. If it doesn't, let's fix `get_taglist_string()`

CHANGELOG.md
dirkhh: oops. thanks for catching that.

qt-models/divetripmodel.cpp
dirkhh: it does. But it's painfully convoluted. It creates a new `struct membuffer`, allocates one byte for the trailing `'\0'`, hands that 'C-string' back to an intermediary function, which then converts it to a `QString` and returns it to us.
But yeah, whatever. It works. Or I could move the `NULL` test one level down into `get_taglist_string()`... I'll take a look.

dirkhh: after reading through this some more and looking at other callers I decided to file this under 'pointless optimizations' and moved on

PR #2831, opened 2020-05-12
Remove 'string marker after running out of strings' warning

torvalds:
This warning is unnecessarily scary - we had a problem with parsing
multiple strings on the same line, but it should be all solved, and
while it does mean that people may have old incorrect git save files
with empty strings, scaring users about it isn't going to help.

And with the warning removed, we can just remove the whole test for an
empty string, because the normal code sequence handles that case just
fine.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -348,10 +348,6 @@ static char *pop_cstring(struct membuffer *str, const char *err)
 		report_error("git-load: string marker without any strings ('%s')", err);
 		return strdup("");
 	}
-	if (!str->len) {
-		report_error("git-load: string marker after running out of strings ('%s')", err);
-		return strdup("");
-	}
 	len = strlen(mb_cstring(str)) + 1;
 	return remove_from_front(str, len);
 }


dirkhh (approved):
Yeah, with an empty string, `remove_from_front()` is a slightly convoluted way to create a `char *` pointing to a single `'\0'` byte. So this is a reasonable fix to the problem.

torvalds:
Well, it actually does something else than create a regular 'char *' from it: the remove_from_front() logic ends up using "detach_cstring(str)" for the "oh, you're removing the whole thing" case, if the empty string was the last thing in the buffer (which is generally should be).

So in a sense it's a bit more efficient than doing the strdup(), because it will simply take over the membuffer allocation directly.

In _another_ sense it means that we now potentially then hit the "there is no string allocation at all" case next time around if there were multiple strings on the line.

But that's actually the right thing to do: if there really are multiple strings, and the string buffer had just a single NUL character in it, we'll now report that "string marker without any strings" warning for the second string. 

And that's a real bug, because it means that there's something wrong in the string parsing.

So the change to just treat the empty string as a normal string can actually change behavior in other ways, but those other ways should be strictly improvements.

I've actually run that commit locally since April (the author date on that thing is Apr 3 for me - I've rebased it multiple times and I wrote a proper commit message only today). But since I haven't been diving my "testing" has been mainly just rebasing and building and checking that my dive log still loads.

PR #2835, opened 2020-05-14
Save SAC, OTU and CNS in the XML export

torvalds:
We do _not_ read them back, since they are calculated values, although I
guess we could aim to do that too at some point in case we have an
import from somewhere else that has these values but not the profile (or
gas use) to actually calculate them.

Requested-by: Miika Turkia <miika.turkia@gmail.com>

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/save-xml.c
+++ b/core/save-xml.c
@@ -502,6 +502,16 @@ void save_one_dive_to_mb(struct membuffer *b, struct dive *dive, bool anonymize)
 		put_format(b, " chill='%d'", dive->chill);
 	if (dive->invalid)
 		put_format(b, " invalid='1'");
+
+	// These three are calculated, and not read when loading.
+	// But saving them into the XML is useful for data export.
+	if (dive->sac > 100)
+		put_format(b, " sac='%d.%03d l/min'", FRACTION(dive->sac, 1000));
+	if (dive->otu)
+		put_format(b, " otu='%d'", dive->otu);
+	if (dive->maxcns)
+		put_format(b, " cns='%d%%'", dive->maxcns);
+
 	save_tags(b, dive->tag_list);
 	if (dive->dive_site)
 		put_format(b, " divesiteid='%8x'", dive->dive_site->uuid);
--- a/dives/TestDiveDM4.xml
+++ b/dives/TestDiveDM4.xml
@@ -5,7 +5,7 @@
 <divesites>
 </divesites>
 <dives>
-<dive number='1' date='2013-02-02' time='16:14:08' duration='59:20 min'>
+<dive number='1' otu='45' cns='15%' date='2013-02-02' time='16:14:08' duration='59:20 min'>
   <notes>Notes are here</notes>
   <cylinder workpressure='200.0 bar' o2='33.0%' start='205.11 bar' />
   <divecomputer>
--- a/dives/TestDiveDM5.xml
+++ b/dives/TestDiveDM5.xml
@@ -5,7 +5,7 @@
 <divesites>
 </divesites>
 <dives>
-<dive number='1' date='2013-02-02' time='16:14:08' duration='59:20 min'>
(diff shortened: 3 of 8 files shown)


torvalds:
I gave this a _very_ quick test, and it seems to export sane values for me.

But I'd suggest Miika actually test this with some exporter of his to see that it works for him too.

Note that my testing really was very cursory, but the change certainly looks trivial.

dirkhh (approved):
Looks reasonable and then will allow @mturkia to do his magic with xslt...

Thanks

Subsurface-divelog:
> On May 14, 2020, at 9:54 AM, Linus Torvalds <notifications@github.com> wrote:
> 
> 
> I gave this a very quick test, and it seems to export sane values for me.
> 
> But I'd suggest Miika actually test this with some exporter of his to see that it works for him too.
> 
> Note that my testing really was very cursory, but the change certainly looks trivial.
> 
> 

They do. And they work. And our tests work as well in that they caught that you changed something in the output format and that therefore the expected test results no longer match.

Are you willing to fix this (a simply 'make check' will show the error locally), or would you prefer that and indentured slave work on that instead?

/D

torvalds:
Oh, it fails the 'testparse' tests.

I'm not entirely sure how that is set up, it looks like it's just verifying that reading and then writing out some test dive, and verifies that it gets the same results?

So it gets

   Actual   (written.takeFirst().trimmed()): "<dive number='417' otu='135%' cns='54'  divesiteid='deadbeef' date='2015-05-23' time='13:23:00' duration='100:00 min'>"
   Expected (readin.takeFirst().trimmed()) : "<dive number='417' divesiteid='deadbeef' date='2015-05-23' time='13:23:00' duration='100:00 min'>"

which is different because now we write out that otu/cns number.

So the failure looks benign and trivial, but I'm not sure exactly what we should do to fix it.

Nor did I even see where it gets that test data from (and the OTU looks scary).

Dirk?

Subsurface-divelog:
On Thu, May 14, 2020 at 10:04 AM Dirk Hohndel <dirk@hohndel.org> wrote:
>
> Are you willing to fix this (a simply 'make check' will show the error locally), or would you prefer that and indentured slave work on that instead?

I can fix it. But should I just change the test dive data or what?

And I found where that 'dive 417' comes from (dives/test40-42.xml),
and realized that while an OTU of 135 is much bigger than anything I
see in my own data, it's fine (not a percentage like CNS, and "limit
to 300 per day" says 135 isn't that bad)

               Linus

torvalds:
Ok, try #2 with an ugly "just update the test XML files" manual fixup for the syntax change.

torvalds:
Note to self: don't use 'hash for 'number'. That "try #2" ends up pointing to the second pull request because hashtag is magic. LOL.

torvalds:
Grr.

I just realized that when I added the percent sign to the CNS value, I screwed up and added it to the OTU value instead.

So now the markers are the wrong way around in this branch. CNS is a percentage, OTU is not. 

And I didn't react to this mix-up even though I literally looked at that OTU of 135 in Miika's dive, and went "ok, an OTU of 135 is fine, because it's just a random unit, and CNS is the percentage. And then I just did the cut-and-paste fix of the test-file updates without reacting to the percentage sign being on the wrong value.

Does this matter? No. Is it wrong? Yes.

I'll just re-do this branch. Aaarghh.

torvalds:
Third time's the charm.. I hope.

dirkhh (approved):
This has the very cool side effect that it will likely catch any time we break (or change) our SAC, CNS, OTU calculations :-)

bstoeger:
To me this feels a bit like throwing out the baby with the bathwater. It introduces a smelly asymmetry between XML and git logs. Are we supposed to write any calculated value that can be exported to the XML?

Oh well, *shrug*

PR #2836, opened 2020-05-14
core/bt: match DC descriptor in lower case

dirkhh:
This fixes a rather subtle bug.

In btdiscovery.cpp we are detecting dive computers based on their BT name and
are setting up product+vendor as the key for that lookup. QMap always uses case
sensitive comparisons and a tiny inconsistency snuck into our code.
libdivecomputer names for the Aqualung dive computers i200C / i300C / i550C end
in an upper case C (as matches the official branding), but in btdiscovery.cpp
we have those names with lower case c. And therefore didn't recognize these
dive computers.

Obviously this is easy to fix by fixing those three strings, but I decided that
it was silly to set ourselves up for similar oversights in the future. So
instead I switched the matching of the descriptor to simply be allways all
lower case.

- [x] Bug fix
- [x] Code cleanup

compare names case insensitively. Since `QMap` doesn't support that (really??) simply store the keys and do the lookups after converting to lower case.

### Related issues:

user support request for Subsurface-mobile on iOS

### Release note:

done

### Documentation change:

not needed

### Mentions:

@bstoeger for sanity check :-)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: fix failure to recognize several Aqualung BLE dive computers
 Mobile: show dive tags on dive details page
 Desktop: update SAC fields and other statistics when editing cylinders
 Desktop: Reconnect the variations checkbox in planner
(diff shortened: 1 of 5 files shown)


neolit123:
i double checked if these are all instances of getting a value from "descriptorLookup" and i think these are all of them.
LGTM, thanks.

atdotde:
Just for the record: My understanding of the QMap documentation is that you can have keys of a type that has its own operator== and operator< . But I agree that storing keys in lower case is simpler in the first place.

PR #2837, opened 2020-05-15
follow up on BT/BLE detection

dirkhh:

- [x] Bug fix
- [x] Functional change

This is a follow up to the previous PR dealing with misrecognized dive computers.
In the process of debugging that I stumbled across the fact that we didn't show the names of devices that weren't recognized as dive computers (which is really quite stupid - did we expect the user to figure out by the magic hex numbers which device to pick)? And that we mixed devices with name (and ALL dive computers that we have seen so far have a name) with those without name.
This rectifies both of those issues

### Release note:

done

### Documentation change:

not needed

### Mentions:

@neolit123 , I guess :-)


neolit123:
LGTM

PR #2841, opened 2020-05-16
Idle whitespace and other cleanups concerning model delegates

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are some idle cleanups, but they have bothered me for the longest time and I feel better now. The only somewhat tricky commit that could lead to a change is the last one. Reason being that it was not really well defined what it was doing before. However, I think that both of the overloaded signals were emitted, so it shouldn't change a thing, except that the functions might now be called only once.


dirkhh (approved):
I can't see anything obviously wrong here.

PR #2844, opened 2020-05-16
BLE: add list of known good/bad BLE GATT services

torvalds:
We've tried to do this "automagic" service discovery, and it mostly
works, but then occasionally it doesn't.

Making things worse, I think different platforms end up enumerating
services differently, so our "pick the first service that looks like it
might be a serial service" ends up working on some platforms, but not
necessarily on others.  Because "first" might be different.

So start a list of known good/bad services - and fall back to the old
logic when you can't decide reliably.

This fills in juat a few cases that I can easily check myself, and the
"details" field for them may be incomplete.  For example, I know Nordic
Semiconductor has their vendor-specific UUIDs, and they can be found in
different devices, so calling them "Nordic UART" and "Nordic Flash"
services makes sense.

But the "Scubapro i770R" service? It might indeed be specific to the
Scubapro i770R.  Or it might be a general service UUID that Pelagic
uses.  Or it might be the service UUID of a particular chip, and found
in dive computers from other designs too (and not necessarily in all
i770R's either).

So this is a preliminary first stab at this, and I'm sure we'll extend
the list and possibly improve on the explanations.

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -101,31 +101,118 @@ void BLEObject::writeCompleted(const QLowEnergyDescriptor&, const QByteArray&)
 	desc_written++;
 }
 
+struct uud_match {
+	const char *uuid, *details;
+};
+
+static const char *match_service(const QBluetoothUuid &service, const struct uud_match *array)
+{
+	const char *uuid;
+
+	while ((uuid = array->uuid) != NULL) {
+		if (service == QUuid(uuid))
+			return array->details;
+		array++;
+	}
+	return NULL;
+}
+
+//
+// Known BLE GATT service UUID's that we should prefer for the serial
+// emulation.
+//
+// The Bluetooth SIG is a disgrace, and never standardized any serial
+// communication over BLE. They should just have specified a standard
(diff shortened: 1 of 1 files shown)


torvalds:
This should fix the problem that Paul Rostykus <sherpa@mind.net> reported with the i200C.

Well, who knows? It should at least avoid that flash upgrade service, and thus *maybe* work.

Looking at logs, I see that other people have hit that same flash update service UUID with other dive computers too, and it looks like it might be iOS (and OS X?) that ends up doing service discovery in a different order and thus hitting this when other platforms might not be.

But I have not been able to test this with the i200C, of course. I did test the Deepblu Cosmic+, the Suunto EON Steel, and the Aqualung i770R. It didn't break those ones, but no promises..

dirkhh (commented):
This looks very reasonable. I'd love to wait until we hear from Paul if indeed this works now on the i300C

dirkhh (approved):
we have confirmation that this works

PR #2847, opened 2020-05-17
BLE: trivial updates to the GATT services lists

torvalds:
Add a couple of known services (Scubapro G2 and Shearwater), and update
the names of others that turn out to be used for multiple dive
computers.  Also add another Broadcom upgrade service UUID.

While at it, sort the services numerically to make it easier to see that
a UUID already exists, since these service numbers do get used across
multiple different devices.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh (commented):
This looks great. Of course I just created releases for iOS and Android, but oh well :-)

torvalds:
Not a big deal on the release.

The main important UUIDs that really matters for several dive computers are the two blacklisted ones we already had. 

The new blacklist entry may help the Mares Bluelink Pro. Again, only relevant on iOS, and that thing is so flaky and slow anyway that it's almost non-functional even when it works.

PR #2855, opened 2020-05-23
mobile/credentials: email address must be lower case

dirkhh:
I could have sworn that I have fixed this several times in various places,
but apparently (as shown by todays support emails) it's still possible to
setup a mixed case email address. So let's try to solve this problem at
the very top.

- [x] Bug fix
- [x] Functional change

Force the email address to be lower case.

### Release note:

done

### Documentation change:

not needed. This is well documented everywhere. And I thought we enforced this.
But clearly we didn't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: convert entered email to lower case
 desktop: update date and time fields on maintab if user changes preferences
 mobile: small improvements to usability with dark theme
 Core: improve service selection for BLE, adding white list and black list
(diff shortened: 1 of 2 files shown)


mturkia:
LGTM
As I am unfamiliar with this part of the code, is saveCloudCredentials called always, when they are inserted. Or is it possible, e.g. to not call this function when apply button is hit instead of save? And end up using mixed case emali?

dirkhh:
I have tried to audit this - but to the best of what I can find this is the only place we set up the email address in the mobile app. I'll merge this locally.

PR #2856, opened 2020-05-24
Trivial: localize water density (salinity) string

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This localizes the salinity string for consistency with the rest of the application. For users of the German locale, the density reads as 1 g / l, which is less than that of air.

### Changes made:

1) Localize water density string.

### Release note:

Done.
### Documentation change:

Not necessary.
### Mentions:

@willemferguson

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: localize salinity string with respect to thousands separators
 desktop: update date and time fields on maintab if user changes preferences
 mobile: small improvements to usability with dark theme
 Core: improve service selection for BLE, adding white list and black list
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
this look fine 
I wouldn't be surprised if there were more spots where we get the thousand separator wrong - except as you pointed out, not a lot of our values tend to go into that range. tank pressures in psi for example.

But then comes the real question. When and why do we print thousand separators in the first place? Are we consistent with that? Not something I have ever wondered about.

PR #2859, opened 2020-05-27
Idle cleanup:  remove dialog singletons

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is an idle code-cleanup: remove dialog singletons. Create the dialogs only when they are needed instead of keeping a global object. This reduces global state and dependencies. These are modal dialogs and now all dialogs in `simplewidget.cpp` are un-singletonized.

I tested all three invocations and they still work.

### Changes made:

1) Remove two dialog singletons.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -647,8 +647,8 @@ void DiveListView::splitDives()
 
 void DiveListView::renumberDives()
 {
-	RenumberDialog::instance()->renumberOnlySelected();
-	RenumberDialog::instance()->show();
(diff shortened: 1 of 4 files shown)


dirkhh (changes requested):
Nice changes.
The commit message of the second commit seems to have the wrong subject (refers to Renumber when that commit is about ShiftTimes)

bstoeger:
Doh - copy&paste error. Fixed.

dirkhh (approved):

PR #2865, opened 2020-05-29
mobile/dive-edit: fix broken editable combo boxes

dirkhh:
The changes in commit 99438121c4 ("mobile/dive-edit: use template components
and theme colors") unintentionally broke our ability to actually edit the
content of one of those fields. 

The more time I wasted trying to get this back to work the clearer it became that the ComboBox theming is flat out broken and the documentation is flat out wrong.

I should care enough to file a bug report.

- [x] Bug fix

Yuck. See why I hate the dark theme? That change was made to accommodate the dark theme. And ended up breaking things badly.
That said, the documentation of Customizing a ComboBox https://doc.qt.io/qt-5/qtquickcontrols2-customize.html#customizing-combobox could mention the fact that this flat out breaks the ability to edit the text...

I simply reverted the part of the change that broke the ComboBox. Even though it implemented EXACTLY what was suggested in the linked documentation. Whatever. So don't use the dark theme. It's simple.

### Changes made:

Add another template type to simplify our code.

### Related issues:

I got quite a few support requests when that got broken.

### Release note:

done

### Mentions:

Since no one seems to ever test the QML stuff that I push, I'll make a new Android beta release instead and wait for people to yell at me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: fix broken editing of location, suit, buddy, and dive master
 Mobile: fix missing translations on Android
 CSV export: support for multiple cylinders
 CSV import: support for multiple cylinders for Subsurface CSV files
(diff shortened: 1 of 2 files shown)


dirkhh:
After having tested this on device, I'd call it a step in the right direction, but it's definitely not there yet.
Now I cannot reliably pick an entry from the drop down... :-(

PR #2867, opened 2020-05-31
Updates to MXE build process [WIP] [Do not merge]

buxtonpaul:
Update the Windows Container build to use newer MXE toolchain to enable Qt 5.14 and remove bespoke versions of qtconnectivity, hidapi, mdbtools and grantlee libraries
 

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Update to use a known tagged version of MXE toolchain, which contains builds qt 5.14 and includes grantlee, mdbtools, hidapi. This enables us to drop pulling and building these modules ourselves. Use of qt5.14 also enables use of QTConnectivity from MXE as well.
Also adds some utility/build scripts to make development with containers more consistant with the containers used in the github actions.

### Changes made:

1) Add a container prep script and update windows.yaml to use this instead of running the individual steps manually
2) Add a container build script to make it easier to build a container locally using the same process as the github actions
3) Move modules from manually built to being fetched from mxe
4) Patch cmake findpkgconfig.cmake to correctly interpret PKG_CONFIG_PATH when cross compiling
5) Tweak cmake file to correctly link in some modules which were failing to link correctly with newer cmake version

### DIFF
--- a/.github/workflows/scripts/windows-container-prep.sh
+++ b/.github/workflows/scripts/windows-container-prep.sh
@@ -1,8 +1,7 @@
 #!/bin/bash
 # abstract the prepare commands for the windows build into a script that can be reused
 # instead of a yaml file
-cd /win
-ln -s /__w/subsurface/subsurface .
+
 echo "downloading sources for fresh build"
 bash subsurface/scripts/get-dep-lib.sh single . libzip
 bash subsurface/scripts/get-dep-lib.sh single . googlemaps
--- a/.github/workflows/scripts/windows-container-prep.sh
+++ b/.github/workflows/scripts/windows-container-prep.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+# abstract the prepare commands for the windows build into a script that can be reused
+# instead of a yaml file
+cd /win
+ln -s /__w/subsurface/subsurface .
+echo "downloading sources for fresh build"
+bash subsurface/scripts/get-dep-lib.sh single . libzip
+bash subsurface/scripts/get-dep-lib.sh single . googlemaps
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -11,7 +11,7 @@ jobs:
   buildInContainer:
     runs-on: ubuntu-latest
     container:
-      image: docker://subsurface/mxe-build-container:1.0
+      image: docker://paulbuxton/mxe-build-container:1.1
 
     steps:
     - name: checkout sources
@@ -20,6 +20,8 @@ jobs:
     - name: get other dependencies
       run: |
         echo "Running script to install additional dependancies into container"
+        cd /win
+        ln -s /__w/subsurface/subsurface .
         bash -x subsurface/.github/workflows/scripts/windows-container-prep.sh 2>&1 | tee pre-build.log
     - name: run build
       run: |
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -19,20 +19,8 @@ jobs:
 
     - name: get other dependencies
       run: |
-        echo "creating the link from /win/subsurface"
-        cd /win
-        ln -s /__w/subsurface/subsurface .
-        ls -l
-        ls -l subsurface/scripts
-        echo "installing missing container components"
-        apt-get install -y ca-certificates libtool
-        echo "downloading sources for fresh build"
-        bash subsurface/scripts/get-dep-lib.sh single . libzip
-        bash subsurface/scripts/get-dep-lib.sh single . hidapi
-        bash subsurface/scripts/get-dep-lib.sh single . googlemaps
-        bash subsurface/scripts/get-dep-lib.sh single . grantlee
-        bash subsurface/scripts/get-dep-lib.sh single . mdbtools
-
+        echo "Running script to install additional dependancies into container"
+        bash -x subsurface/.github/workflows/scripts/windows-container-prep.sh 2>&1 | tee pre-build.log
     - name: run build
       run: |
         cd /win
--- a/desktop-widgets/groupedlineedit.cpp
+++ b/desktop-widgets/groupedlineedit.cpp
@@ -32,6 +32,9 @@
 #include <QScrollBar>
 #include <QTextBlock>
 #include <QPainter>
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+#include <QPainterPath>
+#endif 
 #include <QApplication>
 #include <QStyle>
 #include <QStyleOptionFocusRect>
--- a/scripts/docker/mxe-build-container/Dockerfile-stage1
+++ b/scripts/docker/mxe-build-container/Dockerfile-stage1
@@ -53,7 +53,8 @@ ADD settings-stage1.mk /win/settings.mk
 RUN cd /win ; git clone git://github.com/mxe/mxe ; \
     cd mxe ; \
     git checkout ${_ver} ;
-# Patch the qtconnectivity build to enable native-win32-bluetooth
+# Patch the qtconnectivity build to explicilty enable native-win32-bluetooth and ensure another
+# backend is not picked
 ADD qtconnectivity-1.patch /win/qtconnectivity-1.patch
 RUN mv /win/qtconnectivity-1.patch /win/mxe/src
 RUN mv /win/settings.mk /win/mxe
--- a/scripts/docker/mxe-build-container/Dockerfile-stage1
+++ b/scripts/docker/mxe-build-container/Dockerfile-stage1
@@ -52,7 +52,10 @@ RUN mkdir -p /win
 ADD settings-stage1.mk /win/settings.mk
 RUN cd /win ; git clone git://github.com/mxe/mxe ; \
     cd mxe ; \
-    git checkout ${_ver} ; 
+    git checkout ${_ver} ;
+# Patch the qtconnectivity build to enable native-win32-bluetooth
+ADD qtconnectivity-1.patch /win/qtconnectivity-1.patch
+RUN mv /win/qtconnectivity-1.patch /win/mxe/src
 RUN mv /win/settings.mk /win/mxe
 RUN cd /win/mxe ; \
     make -j 6 2>&1 | tee build.log ;
--- a/scripts/docker/mxe-build-container/Dockerfile-stage1
+++ b/scripts/docker/mxe-build-container/Dockerfile-stage1
@@ -8,8 +8,8 @@ ARG mxe_sha=master
 ENV _ver=${mxe_sha}
 
 # update and set up the packages we need for this cross build
-RUN apt-get update  &&  apt-get upgrade -y
-RUN apt-get install -y \
+RUN apt-get update  &&  apt-get upgrade -y && \
+apt-get install -y \
     autoconf \
     automake \
     autopoint \
@@ -17,6 +17,7 @@ RUN apt-get install -y \
     binutils \
     bison \
     bzip2 \
+    ca-certificates \
     flex \
     g++ \
     g++-multilib \
@@ -28,6 +29,7 @@ RUN apt-get install -y \
     libgdk-pixbuf2.0-dev \
     libltdl-dev \
     libssl-dev \
+    libtool \
     libtool-bin \
     libxml-parser-perl \
     make \
@@ -50,7 +52,7 @@ RUN mkdir -p /win
 ADD settings-stage1.mk /win/settings.mk
 RUN cd /win ; git clone git://github.com/mxe/mxe ; \
     cd mxe ; \
-    git checkout ${_ver} ;
+    git checkout ${_ver} ; 
 RUN mv /win/settings.mk /win/mxe
 RUN cd /win/mxe ; \
     make -j 6 2>&1 | tee build.log ;
--- a/scripts/docker/mxe-build-container/build-container.sh
+++ b/scripts/docker/mxe-build-container/build-container.sh
@@ -6,7 +6,7 @@ SCRIPTPATH=$(dirname $0)
 
 export VERSION=1.1
 pushd $SCRIPTPATH
-docker build -t subsurface/mxe-build-container:$VERSION --build-arg=mxe_sha=1ee37f8 -f Dockerfile-stage1 . 
+docker build -t subsurface/mxe-build-container:$VERSION --build-arg=mxe_sha=1ee37f8 -f Dockerfile-stage1 .
 docker build -t subsurface/mxe-build-container:$VERSION --build-arg=VERSION=$VERSION -f Dockerfile-stage2 .
 popd
  
--- a/scripts/docker/mxe-build-container/build-container.sh
+++ b/scripts/docker/mxe-build-container/build-container.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -x
+set -e
+
+SCRIPTPATH=$(dirname $0)
+
+export VERSION=1.1
+pushd $SCRIPTPATH
+docker build -t subsurface/mxe-build-container:$VERSION --build-arg=mxe_sha=1ee37f8 -f Dockerfile-stage1 . 
+docker build -t subsurface/mxe-build-container:$VERSION --build-arg=VERSION=$VERSION -f Dockerfile-stage2 .
+popd
+ 
--- a/scripts/docker/mxe-build-container/build-container.sh
+++ b/scripts/docker/mxe-build-container/build-container.sh
@@ -9,4 +9,3 @@ pushd $SCRIPTPATH
 docker build -t subsurface/mxe-build-container:$VERSION --build-arg=mxe_sha=1ee37f8 -f Dockerfile-stage1 .
 docker build -t subsurface/mxe-build-container:$VERSION --build-arg=VERSION=$VERSION -f Dockerfile-stage2 .
 popd
- 
--- a/scripts/docker/mxe-build-container/settings-stage2.mk
+++ b/scripts/docker/mxe-build-container/settings-stage2.mk
@@ -9,7 +9,32 @@ JOBS := 2
 MXE_TARGETS :=  i686-w64-mingw32.shared
 
 # The three lines below makes `make` build these "local packages" instead of all packages.
-LOCAL_PKG_LIST := qtbase qtconnectivity qtdeclarative qtimageformats qtlocation qtmultimedia qtquickcontrols qtquickcontrols2 qtscript qtsvg qttools qttranslations qtwebview qtwebkit libxml2 libxslt libusb1 libgit2 nsis curl libzip libftdi1
+LOCAL_PKG_LIST := curl \
+                  grantlee \
+                  hidapi \
+                  libftdi1 \
+                  libgit2 \
+                  libusb1 \
+                  libxml2 \
+                  libxslt \
+                  libzip \
+                  mdbtools \
+                  nsis \
+                  qtbase \
+                  qtconnectivity \
+                  qtdeclarative \
+                  qtimageformats \
+                  qtlocation \
+                  qtmultimedia \
+                  qtquickcontrols \
+                  qtquickcontrols2 \
+                  qtscript \
+                  qtsvg \
+                  qttools \
+                  qttranslations \
+                  qtwebkit \
+                  qtwebview \
+                  zstd
 .DEFAULT local-pkg-list:
 local-pkg-list: $(LOCAL_PKG_LIST)
 
--- a/smtk-import/CMakeLists.txt
+++ b/smtk-import/CMakeLists.txt
@@ -92,7 +92,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
         set(SMTK_LINK_LIBRARIES ${SMTK_LINK_LIBRARIES} -lssh2 -lz -lpthread)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
-				remove_definitions(-DUNICODE)
+	remove_definitions(-DUNICODE)
         add_definitions(-mwindows -D_WIN32)
 endif()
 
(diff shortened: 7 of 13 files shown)


.github/workflows/scripts/windows-container-prep.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# abstract the prepare commands for the windows build into a script that can be reused
+# instead of a yaml file
+cd /win
+ln -s /__w/subsurface/subsurface .
+ls -l
+ls -l subsurface/scripts
```
dirkhh: hehe - I know all about lines like this to make sure my script is in the right spot... 
and then I keep forgetting to remove them before pushing things to master :-/

.github/workflows/scripts/windows-container-prep.sh
```diff
@@ -0,0 +1,12 @@
...
+echo "installing missing container components"
+apt-get install -y ca-certificates libtool
```
dirkhh: BTW: that was a hack because for some reason right when I worked on this new certificates were pushed and things were failing in the container. And then when I realized I had missed libtool I added this here as well.
The certificate thing might be needed - don't know. But libtool could certainly be fixed in the base container.

scripts/docker/mxe-build-container/settings-stage2.mk
```diff
@@ -9,7 +9,31 @@ JOBS := 2
 MXE_TARGETS :=  i686-w64-mingw32.shared
 
 # The three lines below makes `make` build these "local packages" instead of all packages.
-LOCAL_PKG_LIST := qtbase qtconnectivity qtdeclarative qtimageformats qtlocation qtmultimedia qtquickcontrols qtquickcontrols2 qtscript qtsvg qttools qttranslations qtwebview qtwebkit libxml2 libxslt libusb1 libgit2 nsis curl libzip libftdi1
+LOCAL_PKG_LIST := curl \
+									grantlee \
```
dirkhh: whitespace looks a little odd :-)

dirkhh (commented):
This looks great.
Thank you so much for working on it.  A few small comments, but I think you are on the right track!

.github/workflows/scripts/windows-container-prep.sh
buxtonpaul: Will tidy that up in the base container then.

scripts/docker/mxe-build-container/settings-stage2.mk
buxtonpaul: Hmm it does look odd in the review. My intention was to put them on separate lines (and sorted) so it was easy to see what packages are in there.
I will clean that up.

buxtonpaul:
Ok, this now seems to be building ok with my container on dockerhub.
Build artifacts look good as far as I am able to test them....

dirkhh:
Timing is awkward since I'm just about to make a release. Would you mind if I waited looking more into this until after the weekend?

buxtonpaul:
No worries.

desktop-widgets/groupedlineedit.cpp
```diff
@@ -32,6 +32,9 @@
 #include <QScrollBar>
 #include <QTextBlock>
 #include <QPainter>
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+#include <QPainterPath>
+#endif
```
dirkhh, line 37: you didn't need the version check here - and I believe this is already in master by now, anyway

scripts/docker/mxe-build-container/Dockerfile-stage1
```diff
@@ -50,7 +52,7 @@ RUN mkdir -p /win
 ADD settings-stage1.mk /win/settings.mk
 RUN cd /win ; git clone git://github.com/mxe/mxe ; \
     cd mxe ; \
-    git checkout ${_ver} ;
+    git checkout ${_ver} ;
```
dirkhh: trailing space?

scripts/docker/mxe-build-container/build-container.sh
```diff
@@ -1,9 +1,12 @@
 #!/bin/bash
+set -x
+set -e
+
 SCRIPTPATH=$(dirname $0)
 
 export VERSION=1.1
 pushd $SCRIPTPATH
-docker build -t subsurface/mxe-build-container:$VERSION --build-arg=mxe_sha=d63639f -f Dockerfile-stage1 .
+docker build -t subsurface/mxe-build-container:$VERSION --build-arg=mxe_sha=1ee37f8 -f Dockerfile-stage1 .
```
dirkhh: another trailing space

.github/workflows/windows.yml
```diff
@@ -20,6 +20,9 @@ jobs:
     - name: get other dependencies
       run: |
         echo "Running script to install additional dependancies into container"
+        pwd
+        ls 
+        ls /
```
dirkhh: I love these :-)

dirkhh:
So it looks like you have things building now - have you asked people to test the resulting installer? I haven't seen discussion on the mailing list about this. I do have an actual Windows box here and will give it a try today.

As for the PR - typically I ask submitters to clean up the commit history "within reason", so for example adding a file and later deleting it is just adding noise... Different people have different degrees of familiarity and different comfort levels when it comes to rebasing commit history, so I always offer that I can do the cleanup if you want me to, but my preference would of course be a clean, concise set of changes from you. :-)
And before we merge that I would then push the container built from the new tooling to the official docker account and use that.

Yeah, a bit of a mess and a bit of a multi step process :-(

dirkhh:
I won't call this an "in depth" test, but I verified that HID support works (with a Suunto EON Steel) and that the user manual and printing work (so that's QtWebKit being functional, as well as Grantlee).
But I cannot seem to get BT/BLE to work.
And the error log gives me a number of messages of "Unimplemented code." as I am trying to connect to the dive computer - which is a string that doesn't appear in Subsurface. So I wonder if there is still an issue with the Windows BT/BLE implementation in Qt 5.15.
@neolit123 - have you played with that at all?

buxtonpaul:
I am happy to clean up the history, as well as tidy up the other bits and
pieces.

The Bluetooth issues are the thing I was afraid of, as I have no way of
testing that.

My understanding was that the fixes we required in qtconnectivity were
merged into 5.14. I can do some digging around and try and verify that some
more.

neolit123:
> So I wonder if there is still an issue with the Windows BT/BLE implementation in Qt 5.15.

my guess is that MXE does not build the QtConnectivity module with the -native-win32-bluetooth option:
https://github.com/qt/qtconnectivity/blob/2b9b462cb81fabcbfc7a94774dae5182101f7c5e/src/bluetooth/doc/src/bluetooth-index.qdoc#L81-L86

buxtonpaul:
Agree, it doesn't look to be building with that option, but as far as I can
see we weren't previously either...

That option is not the default, so we would need to modify the build
options that are applied by MXE. I can get something sorted for that, and
ask for testers in the forum?

dirkhh:
If by 'forum' you mean 'Subsurface developer mailing list', then yes.
Obviously I can test as well, but it's always good to have more testers to get better coverage.

As for build options - I haven't looked into this at all, so I'm simply trusting Lubomir who wrote the underlying code and who helped me get this to work in the first place.

buxtonpaul:
Yeah. i am not especially familiar with either Qt, or Bluetooth, but should
be able to add some #pragma message to confirm what build variants are
being set in each case before I ask anyone to test something. Based on the
git log on the qt connectivity, everything from the wip/win branch has been
integrated into 5.14, but there are another 220 or so other commits on top
of that which may have changed things, although the build configuration
looks the same based on looking at the .pro and configuration files.

Paul

buxtonpaul:
@neolit123 
Hi, so I am trying to enable the  -native-win32-bluetooth. MXE doesn't make this exactly easy. I have ended up creating a patch to qtconnecticity/src/bluetooth/configure.json that sets enabled: True in the native-win32-bluetooth feature.
This looks like it enables the feature, based on the log in MXE.
```
(cd '/win/mxe/tmp-qtconnectivity-i686-w64-mingw32.shared/qtconnectivity-everywhere-src-5.15.0' && patch -p1 -u) < /win/mxe/src/qtconnectivity-1.patch
patching file src/bluetooth/configure.json
cd '/win/mxe/tmp-qtconnectivity-i686-w64-mingw32.shared/qtconnectivity-everywhere-src-5.15.0' && '/win/mxe/usr/i686-w64-mingw32.shared/qt5/bin/qmake'
Info: creating stash file /win/mxe/tmp-qtconnectivity-i686-w64-mingw32.shared/qtconnectivity-everywhere-src-5.15.0/.qmake.stash
Info: creating cache file /win/mxe/tmp-qtconnectivity-i686-w64-mingw32.shared/qtconnectivity-everywhere-src-5.15.0/.qmake.cache

Running configuration tests...
Checking for BlueZ... no
Done running configuration tests.

Configure summary:

Qt Bluetooth:
  BlueZ .................................. no
  BlueZ Low Energy ....................... no
  Linux Crypto API ....................... no
  Native Win32 Bluetooth ................. yes
  WinRT Bluetooth API (desktop & UWP) .... no
  WinRT advanced bluetooth low energy API (desktop & UWP) . no
```
Qt is now configured for building. Just run 'make'.
Once everything is built, you must run 'make install'.
Qt will be installed into '/win/mxe/usr/i686-w64-mingw32.shared/qt5'.


and looking further down in the MXE log I see compilation steps referencing the _win varients of the files e.g. qbluetoothdevicediscoveryagent_win.cpp which I I am understanding the qt.pro file at all is the one that I would expect to see being used. However when I compare the resulting dll's against the one that was built without the native flag. I get the same binaries.
In fact looking at the MXE build for that version although it is configured not to use the native build, it actually still does end up using it.
The log suggests it isn't configured
```
Qt Bluetooth:
  BlueZ .................................. no
  BlueZ Low Energy ....................... no
  Linux Crypto API ....................... no
  Native Win32 Bluetooth ................. no
  WinRT Bluetooth API (desktop & UWP) .... no
  WinRT advanced bluetooth low energy API (desktop & UWP) . no
```
But I beleive because it has been unable to configure WinRT, that the native ends up as the fallback. 

So I think there might be two problems.
1) WinRT I suspect should suceed.
2) Native isn't working?

I may be hopelessly misunderstanding the qt build process however!
Added the build logs from MXE for QtConnectivity with the native flag and without, in case that has anything useful.
[native.log](https://github.com/subsurface/subsurface/files/4808169/native.log)
[no_native.log](https://github.com/subsurface/subsurface/files/4808170/no_native.log)

dirkhh:
So WinRT doesn't help us, IIRC. We need the Native Win32 Bluetooth to be enabled.
@neolit123 - can you help with this?

neolit123:
@dirkhh 

> But I cannot seem to get BT/BLE to work.
And the error log gives me a number of messages of "Unimplemented code." as I am trying to connect to the dive computer - which is a string that doesn't appear in Subsurface. So I wonder if there is still an issue with the Windows BT/BLE implementation in Qt 5.15.

i wonder if those messages were the actual cause for the failures.

there are still some functions which remain unimplemented: https://github.com/qt/qtconnectivity/blob/2be747ece1872d9d4652f82734d782320cf833ac/src/bluetooth/qbluetoothlocaldevice_win.cpp#L168
but that was always the case.

@buxtonpaul 

looking at both the logs i see [`QT_WIN_BLUETOOTH`](https://github.com/qt/qtconnectivity/blob/dev/src/bluetooth/bluetooth.pro#L226), also all the _win* files being compiled into the DLL, which tells me that the native backend must be enabled for both cases.

> I may be hopelessly misunderstanding the qt build process however!

i'd like to say the same for myself. during the work on the BT(LE) backend, i did experience some weird build failures and inconsistencies which forced me to refresh my local copy of Qt multiple times.

> In fact looking at the MXE build for that version although it is configured not to use the native build, it actually still does end up using it.

looking at https://github.com/qt/qtconnectivity/blob/dev/src/bluetooth/bluetooth.pro that might be the case.

to make sure the DLL is **actually** compiled with native BT(LE) support try searching for linkage to these functions in the DLL object code:
```
objdump -D the-dll-path | grep "SetupDiGetDeviceInterfaceDetail\|BluetoothFindNextRadio"
```

also, if there is a way for you to upload a full Subsurface build that has these MXE produced DLLs, i'd be happy to test it.

buxtonpaul:
So objdump doesn't appear to see symbol names. I had q look at the qtbluetooth.dll with the windows dependancywalker, and I think it was trying to call those functions, but I couldn't be 100% sure I was usingit correctly. So I patched the files qbluetoothlocaldevice_win.cpp, qbluetoothdiscoveragent_win.cpp and qlowenergyconrtroller_win.cpp to fail to compile around the points that they call the SetupDiGetDeviceInterfaceDetail and BluetoothFindNextRadio, and tried building the library with MXE again, and it failed as expected. So i am pretty sure the library is being compiled with Native BT support.

The Build that Dirk tested is still available here
https://github.com/subsurface/subsurface/suites/784358614/artifacts/8307551

What would be the setup I would need to try and debug this locally? Would a pair of laptops supporting Bluetooth LE and some form of serial terminal be enough?

dirkhh:
I have never managed to simulate a ble connection like that.
But what happened for me was that the binary didn't see any ble devices at all. So tell Subsurface to show all devices, even non dive computers. And it should show your fitbit, or some of the many other ble devices on your house

buxtonpaul:
Ok, now I am really confused. Scanning for any bluetooth device looks to work for me. It scans and lists a bunch of BLE devices. However I am not sure if these are just ones that it has found in my saved Bluetooth connections, or they are active signals it has detected....

dirkhh:
Sorry I'm so slow responding. Still suffering from migraines and their side effects. I thought I was better, but I guess I was wrong.
For me the app couldn't see any BLE devices. If it can see them for you, then we should once again try to get more people to test to see if this is just something that's messed up on my Windows machine (which would be very embarrassing - I hate sending people down on time-wasting debugging trips).
At this point we have a couple people on the developer list who have BLE dive computers and who use WIndows - so there should be people who can reliably test this. I have a spare laptop with WIndows that I use every few months to do something for Subsurface - it's entirely possible that something got hosed in the OS and I didn't notice. For one thing, I didn't test with 4.9.5 if THAT could see the BLE dive computer - and we know that that version works for quite a few people out there.
I don't think I can stare at a screen long enough today to try that, though. Hopefully, the drugs will kick in and I'll be better tomorrow

buxtonpaul:
I don't think you are slow to respond, and even if you were, your health is
more important! I know that migraines are not fun! I will send out a call
for tested on the mailing list. Hope you feel better soon.

Paul

neolit123:
thanks for the download link for this installer. i was able to search for BLE devices, find the Perdix AI and download dives from it.

as per usual, it did require some trial and error - first unpairing the device from the Windows Bluetooth panel and pairing it again.

dirkhh:
OK, so this shows that it's my Windows box that is hosed and that I sent you on a wild goose chase.
My sincerest apologies @buxtonpaul 

I think what Lubomir's results indicate is that we should switch to your build.
I will find time today to pull this and do the magic fiddling to get things working with the official binaries.

dirkhh (commented):
Do you want to refactor the commits yourself, or should I just do this when I make everything happen (I will of course retain your author-ship and SOB - just remove some of the noise and deal with updating/using the official Subsurface Docker image)

buxtonpaul:
I have refactored the commits and will push them to my github branch,
unless that will break the comment thread...

dirkhh:
> I have refactored the commits and will push them to my github branch, unless that will break the comment thread...

Not at all. That's exactly the right thing to do.
THANK YOU

neolit123:
> I have ended up creating a patch to qtconnecticity/src/bluetooth/configure.json that sets enabled: True in the native-win32-bluetooth feature.

the rather strange fallback to native BT, which conflates with the documentation of "-native-win32-bluetooth", make we wonder if we should keep the above workaround - i.e. have a patched configure.json to request the native backend explicitly, in case the build system is modified in the future to not have the same fallback.

buxtonpaul:
So based on the feedback from the guys testing suggests this isn't ready for prime time. Looking at a diff of the current qt 5.15 version and the wip/win branch, most of the differences are fairly recent commits. I could wind the mxe version of qtconnectivity back to a slightly earlier one and try that?

@neolit123 Agree that sounds a good idea.

neolit123:
i think the reported issues are related to drivers and overall instability of the native Windows BLE backend. i'd ask the reporters to test an older version of Subsurface (built from the wip/win branch), just to confirm that BLE used to work for them before.

> Looking at a diff of the current qt 5.15 version and the wip/win branch, most of the differences are fairly recent commits

i didn't see anything suspicious:
https://github.com/qt/qtconnectivity/compare/wip/win...5.15.0

buxtonpaul:
So based on the feedback that this is no more broken than the old version I have submitted by cleaned up version of the branch, and added the addition of the patch to ensure we build with the native back-end as per @neolit123 suggestion.

dirkhh:
super busy morning here (Dawn was in two of my six(!) meetings so far today 😄 ) - I'll work on this after 3pm my time (so 3.5h from now)

scripts/docker/mxe-build-container/build-container.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+SCRIPTPATH=$(dirname $0)
+
+export VERSION=1.1
+pushd $SCRIPTPATH
+docker build -t subsurface/mxe-build-container:$VERSION --build-arg=mxe_sha=1ee37f8 -f Dockerfile-stage1 . 
+docker build -t subsurface/mxe-build-container:$VERSION --build-arg=VERSION=$VERSION -f Dockerfile-stage2 .
+popd
+
```
neolit123: missing new line at EOF

smtk-import/CMakeLists.txt
```diff
@@ -92,8 +92,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
         set(SMTK_LINK_LIBRARIES ${SMTK_LINK_LIBRARIES} -lssh2 -lz -lpthread)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
-        set(SMTK_LINK_LIBRARIES ${SMTK_LINK_LIBRARIES} -lwsock32 -lws2_32 -lole32 -limm32 -lwinmm)
-	remove_definitions(-DUNICODE)
+				remove_definitions(-DUNICODE)
```
neolit123: i think the indentation here should be one tab?

also line 93 (`set(SMTK_LINK_LIBRARIES...`) can be changed to use a single tab.

smtk-import/CMakeLists.txt
```diff
@@ -109,6 +108,13 @@ set(SMTK_LINK_LIBRARIES ${SMTK_LINK_LIBRARIES} ${SUBSURFACE_LINK_LIBRARIES} ${SS
 source_group("SmartTrak Import libs" FILES ${SMTK_IMPORT_SRCS})
 set(SMTK_IMPORT_TARGET smtk2ssrf)
 add_library(smtk_import STATIC ${SMTK_IMPORT_SRCS})
+
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
+	# Setting the link target this way resolves an issue with not importing timegettime correctly fron winmm
+	target_link_libraries(smtk_import PRIVATE wsock32 ws2_32 ole32 imm32 winmm)
```
neolit123, line 115: timeGetTime() comes from winmm, i vaguely remember issues around ordering of link libraries, but can you explain what happens here?

scripts/docker/mxe-build-container/Dockerfile-stage1
```diff
@@ -52,7 +52,10 @@ RUN mkdir -p /win
 ADD settings-stage1.mk /win/settings.mk
 RUN cd /win ; git clone git://github.com/mxe/mxe ; \
     cd mxe ; \
-    git checkout ${_ver} ; 
+    git checkout ${_ver} ;
+# Patch the qtconnectivity build to enable native-win32-bluetooth
```
neolit123: i'd expand here a little on the "why" - e.g.

> Patch the qtconnectivity build to enable native-win32-bluetooth explicitly and ensure another backend is not picked

neolit123:
thanks for the updates @buxtonpaul 
some remaining comments in https://github.com/subsurface/subsurface/pull/2867/files, but this LGTM.

scripts/docker/mxe-build-container/build-container.sh
buxtonpaul: I will fix this.

smtk-import/CMakeLists.txt
buxtonpaul: With the latest QT version, running with the original link libraries line the app complains about missing symbol.  I tried reordering things but it didn't seem to help. I beleive this is caused by the updated cmake changing how it handles transitive dependancies. This seemed to be the only way I found to resolve issue.

neolit123: ok. yes, reordering libraries to put the problem one last solves this from my experience.

desktop-widgets/groupedlineedit.cpp
dirkhh: I will just drop this hunk in the merge as we already have this in master

scripts/docker/mxe-build-container/Dockerfile-stage1
dirkhh: still has a trailing space - yeah, nit picking

.github/workflows/windows.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   buildInContainer:
     runs-on: ubuntu-latest
     container:
-      image: docker://subsurface/mxe-build-container:1.0
+      image: docker://paulbuxton/mxe-build-container:1.1
```
dirkhh, line 14: I will adjust this part during the merge

dirkhh (approved):
This looks great. Thank you!
I have kicked off a build here (which, as I'm sure you know, takes a while).
It may take a day until I get this all clean and ready and setup for the merge, but as far as I'm concerned, this is good to go. The little white space things I can fix while merging (since I need to touch the commits, anyway)
I guess you could remove the 'draft' status :-)

dirkhh:
basically merged through #2885

Issue #2830, opened 2020-05-12
Issues with CSV summary dive details - multitank dives 


p-neves:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When exporting summary dive details as csv, information on multiple tanks (if used) is not exported. Specifically, information regarding cylinders used (cylinder size, startpressure, endpressure, o2 and he) is only shown for the first cylinder on the equipment tab. 

### Operating system:

Arch Linux
en_US.UTF-8

### Subsurface version:

4.9.4

Compiled myself

gce-7cc7970540

### Steps to reproduce:

1) Select dives (in this case I've selected dives with more than one tank)
2) Select File/Export and then "CSV summary dive details". Check the selection button "Selected dives"
3) Open the exported csv file and verify the missing info

### Current behavior:

The file only shows information (cylinder size, startpressure, endpressure, o2 and he) for the first cylinder on the equipment tab

### Expected behavior:

Information on all used cylinders should be exported (including -- feature request -- "SAC" for each cylinder.

dirkhh:
@mturkia that sounds like it would fall into your wheelhouse (as does #2829)

mturkia:
Jup, I'll try to take a look. The initial implementation was indeed for the first cylinder only as that is the amount of cylinders I use.

p-neves:
When I export my data I get hundreds of cylinder columns, which then exceeds the column limit for e.g. libreoffice. Can't we limit the number of tanks to a reasonable number (say 10?).

thanks

mturkia:
Can you send me a sample log in subsurface xml? That would demonstrate the issue? I think there is a bug somewhere.

PR #2868, opened 2020-05-31
Csv multitank limit

mturkia:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Ensure the cylinder amount on header line comes from a single dive. This patch takes the last dive in log to be exported as reference.
Export at most as many cylinders as the header line has and print empty strings for non-existent cylinders if one dive has less cylinders than the header.
See #2830 @p-neves commented about unreal number of cylinders in export.

### Changes made:

1) Use exported cylinder amount from a single dive
2) Print at most as many cylinders as the header has
3) Print empty cylinders if required

### Related issues:

#2830 

### Release note:

no need
### Mentions:

@p-neves

### DIFF
--- a/xslt/xml2manualcsv.xslt
+++ b/xslt/xml2manualcsv.xslt
@@ -240,68 +240,68 @@
       <xsl:choose>
         <xsl:when test="$cylinders &lt; position()"/>
         <xsl:otherwise>
-      <xsl:choose>
-        <xsl:when test="@start|@end != ''">
-          <xsl:apply-templates select="."/>
-        </xsl:when>
-        <xsl:otherwise>
-          <xsl:value-of select="$fs"/>
-          <xsl:text>&quot;</xsl:text>
           <xsl:choose>
-            <xsl:when test="$units = 1">
-              <xsl:value-of select="concat(format-number((substring-before(@size, ' ') div 14.7 * 3000) * 0.035315, '#.#'), '')"/>
+            <xsl:when test="@start|@end != ''">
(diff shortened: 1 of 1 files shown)


xslt/xml2manualcsv.xslt
```diff
@@ -38,7 +38,7 @@
           $fs)"/>
 
         <!-- Print cylinder info according to the amount of cylinders in dive -->
-        <xsl:for-each select="dive/cylinder|trip/dive/cylinder">
+        <xsl:for-each select="dive[position()=last()]/cylinder|trip[position()=last()]/dive[position()=last()]/cylinder">
```
dirkhh, line 44: so if the last dive has only one cylinder, we don't get additional cylinders from an earlier dive?

dirkhh (approved):
I think this is an improvement, so I'll merge - but look at my comment... I know nothing of XSLT, but is it possible to determine the max number of cylinders over all the dives being exported?

xslt/xml2manualcsv.xslt
mturkia: yep, that is how it goes. I think using the last dive as reference is more logical than first dive. And with XSLT it is not feasible to look for the max amount of cylinders in any of the dives to be exported (due to poor performance). So this chosen compromise appeared to make most sense to me.

p-neves:
Seems ok with my divelog. Thanks Miika!

PR #2870, opened 2020-06-06
Cloud cache import

dirkhh:

- [x] New feature

This came out of my work with a user debugging an error situation they had
run into, specifically due to some issues with creating the cloud storage repo
on the back end, that was overwritten with an empty dive log, again, and subsequently
we moved the valid local cache on their mobile device out of the way and made all the
dives that they entered inaccessible to them.

This is a bit of a hack, but I think it should be very safe to use. I still decided to hide
it in the developer options, because actually explaining how it works basically requires
that the user understands how our git connection with the cloud works. But it's quite
easy for me to walk a user through how to recover otherwise lost dive data - without
`adb` or rooting an iPhone.

All it does is add a way to show the user the local cache dirs that are available (including
the local no-cloud repo, if it exists) and then import the dives in such a directory into the
current dive list. Which magically recovers otherwise lost data.

### Related issues:

This tends to always be in support emails with me about "lost" data. Be it failed no-cloud conversions or the error I described above.

### Release note:

done

### Documentation change:

No, I don't think I want to try to document this :-)

### Mentions:

I'd love if someone could take a look - I'll ping Linus as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: add advanved option to import dives from local cloud cache directories
 Mobile: fix broken editing of location, suit, buddy, and dive master
 Mobile: fix missing translations on Android
 CSV export: support for multiple cylinders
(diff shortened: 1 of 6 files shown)


dirkhh:
Linus looked at this and found it odd but not wrong.

PR #2871, opened 2020-06-08
mobile/dive-details: restrict width of tags field

dirkhh:
Having a lot of tags (or more precisely, a tags string that is very long) could
cause the width of the dive details view to extend past the width of the the
page. The txtTags label was missing a maximum width, and to make the result
more useful, I also added correct wrapping and elide to the mix (stupidly, we
had the wrap and width for the fixed name of the field ('Tags'), but not for
the user determined content of that field).

- [x] Bug fix

This should fix the UI issues for people with very long tags string

### Changes made:

simply restrict the width of that field (and add wrap / elide for good measure)

### Release note:

done

### Documentation change:

not needed

### Mentions:

@mturkia

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Mobile: correct UI issue with tags that were longer than page width
 Mobile: add advanved option to import dives from local cloud cache directories
 Mobile: fix broken editing of location, suit, buddy, and dive master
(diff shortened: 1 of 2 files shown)


dirkhh:
Two users report it fixes their problem. Since this is QML only, I'll simply merge

PR #2873, opened 2020-06-13
Fix includes for building against Qt-5.15

vesath:
Signed-off-by: Gaetan Bisson <bisson@archlinux.org>


bstoeger:
Thanks - I didn't even know that we had that file.

Since the tests pass, this should be safe to merge.

PR #2875, opened 2020-06-13
git storage: significantly expand logging to stderr

dirkhh:
In many cases we did not log the issues the code ran into to stderr which made
remote debugging user problems much harder. This hopefully will help with that.

Since I was looking at the code, I also made the existing messages more
consistent.

- [x] Code cleanup

This doesn't address the issue that we report a successful remote sync to the user on mobile even if that isn't what happened.

### Release note:

not really user visible

### DIFF
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -160,10 +160,11 @@ static int check_clean(const char *path, unsigned int status, void *payload)
 	status &= ~GIT_STATUS_CURRENT | GIT_STATUS_IGNORED;
 	if (!status)
 		return 0;
+	SSRF_INFO("git storage: local cache dir %s modified, git status 0x%04x", path, status);
 	if (is_subsurface_cloud)
 		report_error(translate("gettextFromC", "Local cache directory %s corrupted - can't sync with Subsurface cloud storage"), path);
 	else
-		report_error("WARNING: Git cache directory modified (path %s) status %0x", path, status);
+		report_error("WARNING: Git cache directory modified (path %s) status 0x%04x", path, status);
 	return 1;
 }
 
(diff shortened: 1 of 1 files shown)


core/git-access.c
```diff
@@ -843,18 +865,21 @@ static git_repository *create_local_repo(const char *localdir, const char *remot
...
 	opts.checkout_branch = branch;
-	if (is_subsurface_cloud && !canReachCloudServer())
+	if (is_subsurface_cloud && !canReachCloudServer()) {
+		SSRF_INFO("git storage: cannot reach remote server");
 		return 0;
+	}
 	if (verbose > 1)
 		SSRF_INFO("git storage: calling git_clone()\n");
 	error = git_clone(&cloned_repo, remote, localdir, &opts);
 	if (verbose > 1)
```
mturkia, line 875: Shouldn't this be inside the next "if (error)" clause?

mturkia (approved):
LGTM, but should still consider moving the one error message to be displayed only if there is actually an error.

core/git-access.c
dirkhh: I'm not sure I understand what you are saying.
Maybe I don't see what specific change you are commenting on. line 868-871 basically state "if this is the Subsurface cloud and we can't reach the cloud, then don't bother and return without an error, but at least log that we can't reach the cloud"

Or are you commenting on something else?

mturkia: I see the comment to be for line 875. Anyway, it should be for the following snippet that prints error message always when in verbose mode, even if no error occurred. Or if we want to print this message out even when there was no error, then we should not claim it to be error.
```
	if (verbose > 1)
		SSRF_INFO("git storage: returned from git_clone() with error %d\n", error);
	if (error) {
		SSRF_INFO("git storage: clone of %s failed", remote);
```
Of course, you did not touch this particular line, but it still jumped to my eye when skimming through this diff.

dirkhh: right. got it. Let me contemplate how to improve the text :-)

dirkhh: this should be more consistent now

dirkhh:
merged via #2876

PR #2878, opened 2020-06-15
[TRIVIAL] cleanup: remove indirection when passing arguments to parser functions

bstoeger:
For unkown reasons, the dive site and trips to be parsed into were
passed as pointers to pointers. A simple pointer seems to be enough,
since the object is not allocated by the function.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a trivial cleanup, which removes a pointer indirection in the parser. Probably an artifact.

### Release note:

No.

### DIFF
--- a/core/parse-xml.c
+++ b/core/parse-xml.c
@@ -1364,11 +1364,9 @@ static void try_to_fill_dive(struct dive *dive, const char *name, char *buf, str
 }
 
 /* We're in the top-level trip xml. Try to convert whatever value to a trip value */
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
My guess is that at some point in the past we allocated the trip or dive site in that parsing function?
I didn't try to go back and check that theory, though.

PR #2879, opened 2020-06-16
Branch updates

dirkhh:

- [x] preparing for git repo changes

Had to invent a new category for the pull request.
Words matter. The term 'master' for the default branch is offensive to many people. Renaming our default branch seems like a very small thing that we can do to signal that we care.

### Changes made:

This changes the references in documentation and scripts that I was able to find.
I am fairly certain that this will break a number of downstream scripts (including some of my own ones that I use to create the official binaries). But the changes in this PR should capture at least the ones that are self-contained in the repo.

### Release note:

Since this isn't user visible, I wasn't going to mention it.

### DIFF
--- a/.github/workflows/android-dockerimage.yml
+++ b/.github/workflows/android-dockerimage.yml
@@ -33,10 +33,10 @@ jobs:
     - name: set env
       run: |
         v=${{ env.VERSION }}
-        b=${{ github.ref }} # -BRANCH suffix, unless the branch is master
+        b=${{ github.ref }} # -BRANCH suffix, unless the branch is main
         b=${b/refs\/heads\//}
         b=${b,,} # the name needs to be all lower case
-        if [ $b = "master" ] ; then b="" ; else b="-$b" ; fi
+        if [ $b = "main" ] ; then b="" ; else b="-$b" ; fi
         echo "::set-env name=NAME::subsurface/android-build-container${b}:${v}"
 
     - name: Build and Publish Linux Docker image to Dockerhub
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -2,10 +2,10 @@ name: Android
 on:
   push:
     branches:
-    - master
+    - main
   pull_request:
     branches:
-    - master
+    - main
 
 jobs:
   buildInContainer:
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -2,10 +2,10 @@ name: iOS
 on:
   push:
     branches:
-    - master
+    - main
   pull_request:
     branches:
-    - master
+    - main
 
 jobs:
   mobileBuild:
--- a/.github/workflows/linux-bionic-5.9.yml
+++ b/.github/workflows/linux-bionic-5.9.yml
@@ -2,10 +2,10 @@ name: Ubuntu 18.04 / Qt 5.9
 on:
   push:
     branches:
-    - master
+    - main
   pull_request:
     branches:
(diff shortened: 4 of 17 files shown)


dirkhh:
There is of course no point in running the docker images (so I killed those), and because the trigger now switched to being pull requests against 'main', we don't get any of the other checks running. So I guess I could call the tests 'passed' :-)
I'd appreciate if some of you could look and see if I missed anything where we still refer to our 'master' branch or if there are other workflows that need adjustment...
@neolit123 @bstoeger @atdotde @buxtonpaul

buxtonpaul:
It all looks sane to me, but I guess the tricky bit is the git commands that assume the master branch is being used, which are the ones that don't show up doing a grep. However given that CI builds work fairly well for dev branches, I think it is probably good.

dirkhh:
Yeah, I worried about that part as well.
The frustrating part will be that everyone who has `master` checked out (so all the developers) will have to rename that and change the upstream. I am actually thinking about adding that to the README or something.
And of course everyone who has tooling to build distro packages, etc...
I still want to do this. It matters.

neolit123:
if a software version control "master" branch is offensive, is a "dive master" offensive too?
we have instances of that in a few places.

in this search:
https://github.com/subsurface/subsurface/search?q=master&unscoped_q=master
i see some references of "master" branches under:
-  /scripts/*
- https://github.com/subsurface/subsurface/blob/da953fa18a954379cc52fec85b2f30fa368aa940/packaging/OBS/make-package.sh#L12 this must change in the original repo.
- this one is interesting as i'm wondering what sqlite would do about "sqlite_master" https://github.com/subsurface/subsurface/blob/d8b92a18520a58646180e6b91cafd89dbb997fcb/core/file.c

overall, i think the git branch change is fine as source code consumption of Subsurface is fairly limited. for larger projects out there with multiple interconnected repositories, the situation is more complicated.

dirkhh:
I was wondering about "dive master" vs "dive guide". I'll see if we get comments on that.

As for references to "master" in other repos that we rely on, we obviously depend on what those maintainers do.

Thanks for the review, Lubo...

PeterZaal:
> I was wondering about "dive master" vs "dive guide". I'll see if we get comments on that.
 
"dive guide" or "dive leader" would be better anyway imo. Dive master usually refers to a certain level of training and certification. This field in Subsurface is used for the person that guides or leads the dive. That can be a dive master, an instructor, or someone with just "dive leader" training (eg GUE has this, and this is certainly not equivalent to dive master training in other organisations) etc.
Actually in Dutch the term "dive master" is already translated as "duikleider" which would be "dive leader".

dirkhh:
Yeah, I thought about "dive leader" as well. MIght as well do that, I guess. It will change a lot of strings for the translators. And then I ask myself - should I rename that structure member as well? The branch we use isn't user visible, it's developer visible. So by that logic... I should change the variable names...

bstoeger:
If you feel that this is the right thing to do - I would suggest you go ahead. FWIW I never encountered the term "dive leader" - that person was consistently called the "dive guide". However, I must note that I never dove in any English-speaking country and only recreationally, so my experience is probably irrelevant.

dirkhh:
I've taken a deep breath and am considering what I really want to do here.
To your specific point - I think dive guide is better. Especially given the connotation that leader (and its most obvious German translation) can have.
But I also have talked to a fair number of people who suggest not using 'main' as the new default branch but to instead use the project name.
So... not sure what I'll end up doing.

PR #2880, opened 2020-06-17
serial_ftdi: adjust callbacks structure

dirkhh:
The set_latency has been removed from libdivecomputer and poll and ioctl have
been added. This caused the callbacks to no longer be aligned correctly and the
functions were called with the wrong arguments through the wrong function
pointers, leading to crashes.

- [x] Bug fix

UGH... this is bad. REALLY bad.
When the function pointer table changed in libdivecomputer, we didn't update serial_ftdi
So this is broken on all platforms that link against that; I need to double check what the current state is on Windows (I think I disabled it there), so this MAY be only Mac, and there, of course, it was missing from the last to releases by mistake... but still.
UGH

### Release note:

need to add that. don't merge until then

### Mentions:

@torvalds

### DIFF
--- a/core/qtserialbluetooth.cpp
+++ b/core/qtserialbluetooth.cpp
@@ -284,10 +284,18 @@ ble_packet_open(dc_iostream_t **iostream, dc_context_t *context, const char* dev
 
 	static const dc_custom_cbs_t callbacks = {
 		.set_timeout	= qt_ble_set_timeout,
+		.set_break	= nullptr,
+		.set_dtr	= nullptr,
+		.set_rts	= nullptr,
+		.get_lines	= nullptr,
+		.get_available	= nullptr,
+		.configure	= nullptr,
 		.poll		= qt_ble_poll,
 		.read		= qt_ble_read,
 		.write		= qt_ble_write,
 		.ioctl		= qt_ble_ioctl,
+		.flush		= nullptr,
+		.purge		= nullptr,
 		.sleep		= qt_custom_sleep,
 		.close		= qt_ble_close,
 	};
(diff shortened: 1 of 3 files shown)


torvalds (approved):
We should use named initializers, I guess, but this looks like an obvious fix in the meantime

dirkhh:
It appears that at least Windows and the Snap are shipping with serial_ftdi enabled. This is IMHO enough to do a 4.9.6 release very soon.

dirkhh:
Thanks @torvalds for pointing out that this would be much better with named initializers

dirkhh:
After some more discussions, switched over to Linus' patch and attributed the commit to him (as his version included the bluetooth fix as well)

dirkhh:
More cringing. Older g++ (including Ubuntu 18.04 and the compiler we use for MXE) don't like the way the initializers were done. Let's see if the workaround recommended on StackExchange is enough to fix that...

torvalds:
That older g++ workaround is disgusting and makes the named initializers almost pointless. Oh well. Hopefully it at least avoids the problem with structure mismatches in the future, which is the big deal.

I'm assuming it's some odd C++ issue, because we've used C named initializers in the kernel forever, without needing any of that. Oh well

dirkhh:
Indeed. It's a C++ thing only. Works for C

core/qtserialbluetooth.cpp
```diff
@@ -284,10 +284,18 @@ ble_packet_open(dc_iostream_t **iostream, dc_context_t *context, const char* dev
 
 	static const dc_custom_cbs_t callbacks = {
 		.set_timeout	= qt_ble_set_timeout,
+		.set_break	= nullptr,
```
buxtonpaul, line 287: Another way of solving this might be to use the 'fieldname: value' syntax https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html

Although that has apparently been obsolete since gcc 2.5. I only came across it after googling the lack of support in g++ for the dot syntax.

dirkhh: Yes. I looked at that. It creates rather verbose warnings with clang which tell you that you are using a non-standard gcc language extension and helpfully shows you the standards compliant correct way to do this.
So yeah, this one is on the gcc community, but since we still use the older compilers on many platforms, it will be a few years before there is a clean solution. For now I think I'll merge this as is and do a 4.9.6 release over the weekend.

PR #2888, opened 2020-06-28
Fix dive mergeing with multi-ple cylinders

torvalds:
We did something really horribly wrong when merging cylinders.  It's
been broken since commit 7c9f46a ("Core: remove MAX_CYLINDERS
restriction"), and used some really strange logic.

This rewrites the logic to be (I think) a bit more easy to understand.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -2204,15 +2204,16 @@ static int different_manual_pressures(const cylinder_t *a, const cylinder_t *b)
  * same cylinder use (ie OC/Diluent/Oxygen), and if pressures
  * have been added manually they need to match.
  */
-static int match_cylinder(const cylinder_t *cyl, const struct dive *dive, const bool *used)
+static int match_cylinder(const cylinder_t *cyl, const struct dive *dive, const bool *try_match)
 {
 	int i;
 
 	for (i = 0; i < dive->cylinders.nr; i++) {
 		const cylinder_t *target;
 
-		if (!used[i])
+		if (!try_match[i])
 			continue;
+
 		target = get_cylinder(dive, i);
 		if (!same_gasmix(cyl->gasmix, target->gasmix))
 			continue;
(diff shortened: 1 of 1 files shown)


dirkhh:
TestMerge fails. Which makes me think that we either have an incorrect reference result (i.e. the test needs to be fixed), or that you broke something.

torvalds:
Ok, fixed stupid thinko. In fact, there were two:

 - when traversing over all the cylinders from dive 'b', we didn't actually traverse to b->nr. We user "res->nr" instead. 

 - we never set "try_match[i]" to false after we had a successful match, so we could match the same cylinder in dive 'a' with multiple cylinders in dive 'b'.

but it looks sane now.

NOTE! The exact order of the resulting cylinders can be different now from what it used to be, but I think the merging logic is simpler, and the new order actually makes more sense (we first copy all cylinders from 'a', and then for each cylinder in 'b' we try to match it against one of those cylinders and if that doesn't work we add it to the list).

torvalds:
It would be good if @bstoeger looked over the logic too, just to make sure it makes sense to him as well.

Except I already see that the Mac / desktopBuild fails again.

torvalds:
Oh fudge. A last-minute edit of mine broke for stupid reasons.

Now fixed for real.

core/dive.c
```diff
@@ -2232,36 +2233,57 @@ static int match_cylinder(const cylinder_t *cyl, const struct dive *dive, const
...
-	res->end.mbar = a->end.mbar ?
-		a->end.mbar : b->end.mbar;
+static void merge_one_cylinder(cylinder_t *a, const cylinder_t *b)
+{
+	if (!a->type.size.mliter)
+		a->type.size.mliter = b->type.size.mliter;
+	if (!a->type.workingpressure.mbar)
+		a->type.workingpressure.mbar = b->type.workingpressure.mbar;
+	if (empty_string(a->type.description))
+		a->type.description = copy_string(b->type.description);
```
bstoeger, line 2243: Probably a stupid comment, however note that empty_string returns true for a pointer to "". That would result in a tiny memory leak. I'm not sure if that can actually happen.

bstoeger:
From what I can tell, this looks good, thanks. Note that for planned dives we have a very horrible "cylinder past last cylinder" representing the "surface interval" for historical reasons, which will be lost. However, one could argue that one shouldn't merge planned dives.

@dirkhh: This deserves a CHANGELOG entry.

dirkhh:
Thanks for this. I'll add a CHANGELOG as well.

PR #2889, opened 2020-06-29
profile: correctly display dives with multiple pressure sensors

### LINKED ISSUE #2877: Multiple tank, pressure issue
### Describe the issue:
- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
(I'm new to reporting an issue, so I hope this is ok):

I use two tanks with a pressure sensor on each.  It seems the second tank's data is not available due to the following bug:

The following line resets the pi->nr value after loading a profile into memory
https://github.com/subsurface/subsurface/blob/b01361170770a1cd263c57e856437e5f22c794df/core/profile.c#L645

but as a result, breaks array access in the following set/get functions. This breaks future access to pressure data for any additional tanks.

https://github.com/subsurface/subsurface/blob/b01361170770a1cd263c57e856437e5f22c794df/core/profile.h#L110-L113

### Subsurface version:
Came across the issue with an official release , tested code changes with v4.9.4-122-g87c91af82

### Steps to reproduce:
Load a log with two pressure sensors, look at the pressure graph when the log is selected

### Current behavior:
First transmitter shows correctly, second tank has no data.

### Additional information:
I've replace line 645 with the following, as a temporary workaround, but its just a hack, so I didn't want to submit it.
```
//Re-order the pressures...
    int nrActual = pi->nr;
    pi->nr = idx;
    for( int cyl = 0; cyl < pi->nr_cylinders; cyl++){
        for(int sample = 0; sample < pi->nr; sample++){
            for(int sensor = 0; sensor < NUM_PLOT_PRESSURES; sensor++){
                pi->pressures[cyl * pi->nr + sample].data[ sensor ] =
                        pi->pressures[cyl * nrActual + sample].data[sensor];

            }
        }
    }
```

bstoeger:
When removing the MAX_CYLINDERS restriction, the population of the
pressure profile data has been broken. The pressure data is organized
as a nr_cylinders x nr matrix, where nr is the number of samples. However,
an oversized array was allocated and then the nr_entries value set
to the actually needed number of entries. Obviously, this messed up the
matrix accesses.

Do a quick fix by reallocating the array at the end of the function.
We can't simply keep the old nr value, because the profile would attempt
to render too many entries. Probably not a big disaster, because the
entries are empty anyway, but still...

Fixes #2877

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Attempt at a bug fix for #2877, see discussion there.

### Changes made:

1) Realloc the array at the end of `populate_plot_entries` to ensure correct indexing.

### Related issues:

Fixes #2877.

### Release note:

Should be done yes, together with Linus' fixes to dive merging.

### Documentation change:

No.
### Mentions:

@BartJV @dirkhh

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -512,6 +512,7 @@ static void populate_plot_entries(struct dive *dive, struct divecomputer *dc, st
 	int idx, maxtime, nr, i;
 	int lastdepth, lasttime, lasttemp = 0;
 	struct plot_data *plot_data;
+	struct plot_pressure_data *final_pressure_data;
 	struct event *ev = dc->events;
 	maxtime = pi->maxtime;
 
@@ -642,7 +643,20 @@ static void populate_plot_entries(struct dive *dive, struct divecomputer *dc, st
 	/* Add two final surface events */
 	plot_data[idx++].sec = lasttime + 1;
 	plot_data[idx++].sec = lasttime + 2;
+
+	/* We probably used less plot_data entries than needed.
+	 * "pi->nr" is the number of the allocated entries and "idx" is the number
+	 * of needed entries. Now, let's alloc an array of the actually needed size
+	 * and continue using that.
+	 */
+	final_pressure_data = calloc(idx * (size_t)pi->nr_cylinders, sizeof(struct plot_pressure_data));
+	for (int i = 0; i < pi->nr_cylinders; ++i) {
+		for (int j = 0; j < idx; ++j)
+			final_pressure_data[i * idx + j] = pi->pressures[i * pi->nr + j];
+	}
+	free(pi->pressures);
 	pi->nr = idx;
+	pi->pressures = final_pressure_data;
 }
 
 /*


core/profile.c
```diff
@@ -642,7 +643,20 @@ static void populate_plot_entries(struct dive *dive, struct divecomputer *dc, st
 	/* Add two final surface events */
 	plot_data[idx++].sec = lasttime + 1;
 	plot_data[idx++].sec = lasttime + 2;
+
+	/* We probably used less plot_data entries than needed.
```
dirkhh, line 647: my favorite pet peeve in the English language. If you can count it, it's 'fewer', otherwise 'less'. I give you less respect, but fewer dollars.

dirkhh (approved):
This looks reasonable. I'll point Linus at this as well.

Disheartening how many issues with the `MAX_CYLINDERS` change we keep finding :-(

bstoeger:
> Disheartening how many issues with the MAX_CYLINDERS change we keep finding :-(

Expecting otherwise would have been naive. We really should add multi-sensor dives to the test-set.

mturkia (approved):
LGTM

bstoeger:
Let's go with the alternative, for the simple reason that it is less code even if it is somewhat subtle.

PR #2899, opened 2020-07-14
Disable download button

dirkhh:

- [x] Bug fix

weirdly we had the code to do this, but the condition was spilt across two lines and that appears to fail silently

### Release note:

added CHANGELOG entries for this and for the last two updates

### Documentation change:

not needed


mturkia (approved):
LGTM

PR #2902, opened 2020-07-24
Seacsync

jwobser:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Implement parser for SeacSync databases.
### Changes made:

1) Created import-seac.c
2) Added test in parse.c for seac type DB
3) Added sample DB
4) Added helper function to get local time from UTC + Offset
5) Added 1/100 °C to mkelvin conversion function

### Additional information:

Sample database available. Includes modified dive to simulate gas change.

### Mentions:

@mturkia @dirkhh


core/import-seac.c
```diff
@@ -0,0 +1,360 @@
...
+		month -= 12;
+	}
+
+	divetime.tm_year = year - 1900;
+	divetime.tm_mon = month - 1;
+	divetime.tm_mday = day;
+	divetime.tm_hour = hour;
+	divetime.tm_min = mins;
+	// WARNING: Possibly Non portable
+	divetime.tm_sec = secs - timezone; // counteract mktime() assumption of local time
```
mturkia: I am not really familiar with this timezone variable. However, it seems to be defined as seconds west of UTC (man tzset). If this is the case, UTC+1 would mean timezone == -3600. Your code suggests that I have misunderstood the whole thing. So explanation would be in order. Or fixing the code :D 

Of course, it must also be ensured that this works on Windows also.

core/import-seac.c
```diff
@@ -0,0 +1,360 @@
...
+	// 6 = dive_type
+	// TODO: confirm freedive recorded as type 2
+	if (data[6]) {
+		switch (atoi(data[6])){
+		case 1:
+			state->cur_dive->dc.divemode = OC;
+			break;
+		case 2:
+			state->cur_dive->dc.divemode = FREEDIVE;
+			break;
```
mturkia: Does this software/DC support CCR or Gauge mode?

core/import-seac.c
```diff
@@ -0,0 +1,360 @@
...
+		}
```
mturkia, line 147: I wonder if there should be a default clause to print the unknown value when verbose mode is enabled. This might make it easier to add divemodes if we receive a user report certain mode is missing.

core/import-seac.c
```diff
@@ -0,0 +1,360 @@
...
+	// as 150 for both.
+	if (data[8]) {
+		switch (atoi(data[8])){
+			case 150:
+				state->cur_dive->salinity = 0;
+				break;
+			case 100:
+				state->cur_dive->salinity = 1;
+				break;
+		}
```
mturkia, line 178: Also here a default clause to print the unknown salinity value might help future development.

core/import-seac.c
```diff
@@ -0,0 +1,360 @@
...
+	if (data[1]) {
+		utf8_string(data[1], &state->cur_settings.dc.serial_nr);
+		state->cur_settings.dc.deviceid = atoi(data[1]);
+	}
+
+	dc_settings_end(state);
+	settings_end(state);
+
+	if (data[11]) {
+		state->cur_dive->dc.maxdepth.mm = 10*atoi(data[11]);
```
mturkia: spaces around the asterisk

core/import-seac.c
```diff
@@ -0,0 +1,360 @@
...
+	// Catch a bad query
+	retval = sqlite3_step(sqlstmt);
+	if (retval == SQLITE_ERROR) {
+		fprintf(stderr,"%s","Getting dive data from SeacSync DB failed.\n");
+		return 1;
+	}
+
+	// Create first cylinder
+	cylinder_t *curcyl = get_or_create_cylinder(state->cur_dive, 0);
+	curcyl->gasmix.o2.permille = 10 *sqlite3_column_int(sqlstmt,4);
```
mturkia: space after asterisk, and after comma

mturkia (commented):
Looks good. I mostly have some style / nitpicking comments to make.

A sample database with one or two dives would be nice addition to our test data. We have these stored in dives directory and e.g. TestDiveDM5.db is one sample from Suunto DM5.

A test case for this import (tests/testparse.cpp) would also be great. TestParse::testParseDM5 would be good example for this.

There are quite a few space issues. E.g. after comma, there should be a space. Just browse through the code or grep them...

I prefer multiline comments with /* ... */, not starting each line with //. I guess this is not mentioned in coding style, but I think it is common practice in Subsurface.

core/import-seac.c
```diff
@@ -0,0 +1,360 @@
...
+ * to query the dive_data table for the current dive_number.
+ * @param[in] *param The current dive state
+ * @return Always 0
+ */
+static int seac_gaschange(void *param, sqlite3_stmt *sqlstmt)
+{
+	struct parser_state *state = (struct parser_state *)param;
+
+	event_start(state);
+	state->cur_event.time.seconds = sqlite3_column_int(sqlstmt,1);
```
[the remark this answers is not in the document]
jwobser: Fixed! sed to the rescue!

core/import-seac.c
```diff
@@ -0,0 +1,360 @@
...
+	const char *get_samples = "SELECT dive_number,runtime_s,depth_cm, temperature_mCx10,active_O2_fr, first_stop_depth_cm, first_stop_time_s, ndl_tts_s,cns FROM dive_data WHERE dive_number = ? ORDER BY runtime_s ASC";
+		//  0 = dive_number
+		//  1 = runtime_s
+		//  2 = depth_cm
+		//  3 = temperature_mCx10  - eg dC
+		//  4 = active_O2_fr
+		//  5 = first_stop_depth_cm
+		//  6 = first_stop_time_s
+		//  7 = ndl_tts_s
+		//  8 = cns
```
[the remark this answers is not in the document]
jwobser: Converted to block style. I was added columns to the SQL query one at a time when doing initial development.

core/import-seac.c
jwobser: I added a large comment block in the next commit.

timezone comes from <time.h>, and you're correct, it's seconds west of UTC.
Unfortunately mktime() assumes it is passed local time in the tm struct, so it applies the offset
of the local machine when returning time_t . Have to compensate for this as the timezone
for the dive doesn't necessarily match timezone of user's machine.

core/import-seac.c
jwobser: Good idea for the default case. Added some output to stderr if verbose flag is set.

core/import-seac.c
jwobser: Also added default case with sdterr out in verbose mode.

core/import-seac.c
jwobser: The DC supports gauge mode, but I do not believe it records samples. I will have to trial this.

core/import-seac.c
atdotde: Just a quick remark: In other places (the image importer), we use Qt's functions for date/time calculations with time zones etc, so if this doesn't work on windows, you could try that as well. The problem is, you cannot directly call Qt functionality from pure C (as opposed to C++). So when we need to do that, we go through a helper function in qthelper.cpp.

atdotde (changes requested):
Please rewrite history to squash things like this that delete stuff you introduced in a previous commit.

git rebase -i 

is your friend for things like this.

core/import-seac.c
mturkia: 10*atoi => 10 * atoi

core/import-seac.c
```diff
@@ -0,0 +1,407 @@
...
+
+	// Create sql_stmt type to query DB 
+	retval = sqlite3_prepare_v2(handle, get_samples, -1, &sqlstmt, 0);
+	if (retval != SQLITE_OK) {
+		fprintf(stderr, "%s", "Preparing SQL object failed when getting SeacSync dives.\n");
+		return 1;
+	}
+
+	// Bind current dive number to sql statement
+	sqlite3_bind_int(sqlstmt, 1,state->cur_dive->number);
```
mturkia: space after ,

mturkia:
The iOS build failure seems to be caused by this change. Check out the build log..

```
Undefined symbols for architecture x86_64:
  "_parse_seac_buffer", referenced from:
      _parse_file in file.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

** BUILD FAILED **


The following build commands failed:
	Ld Release-iphonesimulator/Subsurface-mobile.app/Subsurface-mobile normal x86_64
(1 failure)
make: *** [xcodebuild-release-simulator] Error 65
##[error]Process completed with exit code 2.
```

core/import-seac.c
jwobser: I've looked into mktime() further, and it seems there are even more pitfalls involving if the local machine is currently in DST. The times end up being 1hr off.

I think I'm going to have to dig into the QT DateTime library a little bit.

jwobser:
> 
> 
> The iOS build failure seems to be caused by this change. Check out the build log..
> 

Is it possible to build for iOS on Linux? I know Apple likes to lock down their dev tools.
I couldn't find any glaring errors in the code, and I'm not sure if the Travis tool allows exporting .o files.

mturkia:
> > The iOS build failure seems to be caused by this change. Check out the build log..
> 
> Is it possible to build for iOS on Linux? I know Apple likes to lock down their dev tools.
> I couldn't find any glaring errors in the code, and I'm not sure if the Travis tool allows exporting .o files.

Maybe @dirkhh  or @atdotde have some idea about this.

However, I don't think we are going to allow one to import Seac logs on mobile as it does not really make sense there. So you could ifdef the seac import to be excluded on mobile builds. Unless there are better suggestions.

dirkhh:
Sorry to be so slow in engaging here. Real life is keeping me focused on other things right now.
The iOS build failure is very easy to explain. For <censored> reasons, the iOS build is the only one that is NOT created using `cmake`. So whenever you add a file to the `CMakeLists.txt` file, it also needs to be added to `packaging/ios/Subsurface-mobile.pro` - the `qmake` file that we use to create the iOS build.

dirkhh:
But to @mturkia 's comment - I agree that this doesn't really make sense to have on mobile. Dealing with files on mobile is a major pain and basically will be impossible on Android in the next Android release. There are ways to do this on iOS, but again, not something that I am really interested in supporting. So most likely the better approach is to exclude this for mobile builds.

mturkia:
What is your take on the timezone currently? I am mainly asking as you have a TODO there. Should we just go ahead and merge this as is or are you still investigating that part?

jwobser:
Actually I'm out diving this weekend, trying different timezones. Did +5:45
today, plan to get -12 and +14 tomorrow. That should be a pretty through
test. Also going to try a gas switch to make sure that works properly.

I've also added a key value pair for serial number in extra data, but
haven't pushed that change yet.

Still need to look into calculating a hash for the dc, as it's an
alphanumeric string. Also, in the dc struct it's a pointer, and I don't see
a function to allocate a new one. Is that typically done with Cpp to avoid
memory leaks?

dirkhh:
> Actually I'm out diving this weekend, trying different timezones. Did +5:45 today, plan to get -12 and +14 tomorrow. That should be a pretty through test. 

How did that go?

> Also going to try a gas switch to make sure that works properly. 

ditto :-)

> I've also added a key value pair for serial number in extra data, but haven't pushed that change yet. 

Do you need help here?

> Still need to look into calculating a hash for the dc, as it's an alphanumeric string. Also, in the dc struct it's a pointer, and I don't see a function to allocate a new one. Is that typically done with Cpp to avoid memory leaks?

It's manually allocated where needed. The first one is always part of the dive structure, and subsequent ones are then `malloc()`'ed

jwobser:
> 
> 
> > Actually I'm out diving this weekend, trying different timezones. Did +5:45 today, plan to get -12 and +14 tomorrow. That should be a pretty through test.
> 
> How did that go?

It works great! The QT Date Time library did the trick, just had to make a helper function that took an ISO-formatted string. All the offsets are actually inverse in the code, because the database has UTC, and we need to go back to local. But all the timezones I tested match what's displayed for local time on the dive computer itself.

> 
> > Also going to try a gas switch to make sure that works properly.
> 
> ditto :-)

Worked great. Did a 22% EANx safety stop, haha!

> 
> > I've also added a key value pair for serial number in extra data, but haven't pushed that change yet.
> 
> Do you need help here?
> 
> > Still need to look into calculating a hash for the dc, as it's an alphanumeric string. Also, in the dc struct it's a pointer, and I don't see a function to allocate a new one. Is that typically done with Cpp to avoid memory leaks?
> 
> It's manually allocated where needed. The first one is always part of the dive structure, and subsequent ones are then `malloc()`'ed

I think I got a little mixed up here. Wrote that reply on the dive boat. Extra data does have a helper function, and I'm using that to store the GF-H and GF-L that were in use by the DC for that dive. I'm not sure if there is a more appropriate location to store that info.

The firmware, serial, and model strings are all pointers in the dc struct. But I looked at the utf8_string function I had seen @mturkia use in the Suunto importer, and it looks like behind the scenes it just uses strdup to malloc a new string. So I reused that code, except for the model string. Used strdup directly on a string literal.

I'm assuming on program close the dc structs free all the pointers it contains?

One other point of interest, the DB actually contains the nitrogen loading for all 16 compartments for each sample. The model is supposed to be straight Buhlmann, it might be fun to compare the dc model to Subsurface's. For now, though, the importer is pretty complete. It pulls in all the data that I could find a place for.

dirkhh:
> > > Still need to look into calculating a hash for the dc, as it's an alphanumeric string. Also, in the dc struct it's a pointer, and I don't see a function to allocate a new one. Is that typically done with Cpp to avoid memory leaks?
> > 
> > 
> > It's manually allocated where needed. The first one is always part of the dive structure, and subsequent ones are then `malloc()`'ed
> 
> I think I got a little mixed up here. Wrote that reply on the dive boat. Extra data does have a helper function, and I'm using that to store the GF-H and GF-L that were in use by the DC for that dive. I'm not sure if there is a more appropriate location to store that info.

Those should definitely be in extra data - that's where it ends up when downloaded e.g. from a Shearwater dive computer

> The firmware, serial, and model strings are all pointers in the dc struct. But I looked at the utf8_string function I had seen @mturkia use in the Suunto importer, and it looks like behind the scenes it just uses strdup to malloc a new string. So I reused that code, except for the model string. Used strdup directly on a string literal.
>

Correct.
 
> I'm assuming on program close the dc structs free all the pointers it contains?

Well, on program close everything goes 'poof' anyway. But yes, we try to free most of the data in an orderly fashion.
 
> One other point of interest, the DB actually contains the nitrogen loading for all 16 compartments for each sample. The model is supposed to be straight Buhlmann, it might be fun to compare the dc model to Subsurface's. For now, though, the importer is pretty complete. It pulls in all the data that I could find a place for.

We don't have a structure that would reasonably hold that data **per sample** (wow, that's a LOT of data)

mturkia (approved):
Did you get to confirm if freedive has a record type of 2?

LGTM

Edit: This will need also a changelog entry, and instructions to user manual how to obtain the seacsync database to import

jwobser:
> 
> 
> Did you get to confirm if freedive has a record type of 2?
> 
> LGTM
> 
> Edit: This will need also a changelog entry, and instructions to user manual how to obtain the seacsync database to import

Glad I check the freedive type today. It is actually type 3, and the gauge mode is type 2.

Added instruction to get database on windows; I don't know where the program stores the DB on Mac.

dirkhh:
Sorry - I got distracted with real life. Where are we with this?
@atdotde you had requested changes...?

jwobser:
> Sorry - I got distracted with real life. Where are we with this?
> @atdotde you had requested changes...?

He had requested a rebase to get rid of the chain of fix-up commits. I force pushed the rebase, but now I can't resolve his change request.

@mturkia helped me work through some timezone handling issues and finish up some open TODO's I had left.

I have been added dives to my log with it, and have been pleased with it. Correctly grabs gas changes, comments made from the proprietary app and all the data I could find a home for.

mturkia:
manually merged

PR #2905, opened 2020-08-07
RFD: Improve resolution of profile export

atdotde:
The way we export the profile image (as direct export but
also used for printing) is that we render the profile
from the screen to a Pixmap and save that to a file. Unfortunately
this results in very bad resultion and a blurred image.

This is an attempt to improve that situation but it's still far
from perfect: Rather than a QPixmap and grab, I now use a QImage
(where I can set the size) and render, and indeed the picture resolution
(when vied at fixed size) get's better this way. The disadvantage
is that icons and text get smaller at the same rate und so
there is a natural limit on how big we can get. Maybe somebody
with better Qt knowledge can take off from here. In my opinion
this is already a step in the right direction.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -796,6 +796,9 @@ void ProfileWidget2::plotDive(const struct dive *d, bool force, bool doClearPict
 		item->setModel(dataModel);
 		item->setEvent(event, lastgasmix);
 		item->setZValue(2);
+#ifndef SUBSURFACE_MOBILE
+		item->setScale(printMode ? 4 :1);
+#endif
 		scene()->addItem(item);
 		eventItems.push_back(item);
 		if (event_is_gaschange(event))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ core: add support for Shearwater Peregrine (requires firmware V79 or newer)
 core: fix renumbering of imported dives [#2731]
 mobile: fix editing tank information
 planner: Handle zero length segments gracefully when replanning
+profile: improve resolution in printing and export
 mobile: disable download button if no connection is selected
 mobile: fix incorrect time stamps on GPS track points created via location service
 core: correctly recognize A1 as BLE dive computer
--- a/desktop-widgets/divelogexportdialog.cpp
+++ b/desktop-widgets/divelogexportdialog.cpp
@@ -225,11 +225,11 @@ void DiveLogExportDialog::on_buttonBox_accepted()
 void exportProfile(const struct dive *dive, const QString filename)
 {
 	ProfileWidget2 *profile = MainWindow::instance()->graphics;
-	profile->plotDive(dive, true, false, true);
 	profile->setToolTipVisibile(false);
 	profile->setPrintMode(true);
 	double scale = profile->getFontPrintScale();
 	profile->setFontPrintScale(4 * scale);
+	profile->plotDive(dive, true, false, true);
 	QImage image = QImage(profile->size() * 4, QImage::Format_RGB32);
 	QPainter paint;
 	paint.begin(&image);
@@ -238,4 +238,5 @@ void exportProfile(const struct dive *dive, const QString filename)
 	profile->setToolTipVisibile(true);
 	profile->setFontPrintScale(scale);
 	profile->setPrintMode(false);
+	profile->plotDive(dive, true);
 }
--- a/desktop-widgets/divelogexportdialog.cpp
+++ b/desktop-widgets/divelogexportdialog.cpp
@@ -227,7 +227,15 @@ void exportProfile(const struct dive *dive, const QString filename)
 	ProfileWidget2 *profile = MainWindow::instance()->graphics;
 	profile->plotDive(dive, true, false, true);
 	profile->setToolTipVisibile(false);
-	QPixmap pix = profile->grab();
+	profile->setPrintMode(true);
+	double scale = profile->getFontPrintScale();
+	profile->setFontPrintScale(4 * scale);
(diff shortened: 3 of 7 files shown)


dirkhh:
It's been a very long time since I worked on this code, but the core issue that you are describing isn't new.
We already have the mechanism in place via `fontPrintScale` to make sure fonts are larger in print - I'm wondering if this somehow got broken at some point. And I think we should be able to use this for icons as well.
But yeah, this is entirely fixable.

atdotde:
I know this existed for a long time, but I thought this might be the time for fixing this. I wasn't aware fo fontPrintScale but this indeed helps and I did something similar for the event icons. I am a bit worried, though, that this screws up the mobile version (which I currently cannot build here) as that seems to use print mode even for on device.

I am currently looking at the export->dive profile case, but originally wanted to cover printing. Too bad, as of now, I don't have a current grantee on my laptop and thus build without printing. And this make Qt Creator think, the source files for printing are not part of the project. So this has to wait for another day.

dirkhh:
Why can't you build mobile? Anything I can help with? I build and run it on macOS all the time...
And yes, that uses the same hack...

atdotde:
No, it's just that I do it so rarely that it to work takes longer to get it going again (missing include paths, icon files in the right place etc etc). I just invested about an hour in this, I am making progress but I am not there yet again. 

Or in short: just running

subsurface/scripts/build.sh -mobile 

Is too optimistic.

Subsurface-divelog:
Remove the build-mobile directory and running the script is enough.
If not, I need to fix this. But it worked for me just a few days ago

atdotde:
As so often: Of course you are right. Started from a fresh directory with a brand new clone of subsurface. Running the build script almost just worked: It produced a Subsurface-mobile binary which then crashed immediately after start-up (this is much further than where I got earlier today). And I had seen this before, going to the build-mobile directory and running

make -j2 install

again produces a working binary. I have no clue what kind of artefacts in my old tree prevented me from getting anything earlier. Thanks for the encouragement!

dirkhh:
> As so often: Of course you are right. 

I apologize. I know this is a bad habit of mine.

> Started from a fresh directory with a brand new clone of subsurface. Running the build script almost just worked: It produced a Subsurface-mobile binary which then crashed immediately after start-up (this is much further than where I got earlier today). And I had seen this before, going to the build-mobile directory and running
> 
> `make -j2 install`
> 
> again produces a working binary. 

Yes, I'm aware of this. There's some issue where we need to change some library paths which we do in the install step. It's super annoying since that means you have to run install after every build, which really slows down development :-(

> I have no clue what kind of artefacts in my old tree prevented me from getting anything earlier. Thanks for the encouragement!

Usually it is hard paths to older include files or more likely libraries. And you don't need to do a new clone of `subsurface`. All it takes is to delete the whole `build-mobile` directory and then run the build script again. Not ideal, but it works.

profile-widget/profilewidget2.cpp
```diff
@@ -796,6 +796,9 @@ void ProfileWidget2::plotDive(const struct dive *d, bool force, bool doClearPict
 		item->setModel(dataModel);
 		item->setEvent(event, lastgasmix);
 		item->setZValue(2);
+#ifndef SUBSURFACE_MOBILE
+		item->setScale(printMode ? 4 :1);
```
dirkhh, line 800: should this be a fixed scale, or is there a reasonable way to calculate a "good" scale?
I'm not trying to make this harder for you, I'm really just asking

atdotde: I did a bit trial and error and 4 is about right for me. But of course this depends on screen resolution and your target resolution for the image file. Prescribing a fixed image size and adopting the scale factor accordingly does not really work, since the profile has the same aspect ration as on the screen (which is variable). One could try to be clever about this (like fixing either target width or height). What do you think?

neolit123: ~why do we have to set printMode on mobile?~
~(it should not matter as printing is not supported on mobile, or at least it wasn't the last time i've checked)~

NVM: this is under `ifndef`.

neolit123: i think what will happen here is that for the profile image export functionality (divelogexportdialog.cpp # exportProfile()) the scale will end up being adjusted properly, given the QImage scales x4 that is also in this diff.

but for this functionality:
https://github.com/subsurface/subsurface/blob/5b078539a74ad65bd8e984bdb6f692ae4a2ae6e4/desktop-widgets/printer.cpp#L44

which renders the profile without the x4 scale, the above change will make event icons look big (out of scale) on the profiles that are printed via the file -> print func.

dirkhh:
@atdotde - what are the next steps here?

dirkhh (commented):
I'd love to hear what you think about Lubo's comments...
I really like the idea of creating better printed profiles, so I'm eager to merge this, but I also think that on both commits he has a point :-)

dirkhh:
The connection for your resolution spin box is still wrong. Let's make it much simpler. Let's do this:
```
connect(ui.resolution, QOverload<int>::of(&QSpinBox::valueChanged), [this](int value) {
               printOptions->resolution = value;
       });
```
and remove the whole `on_resultion_edited()` slot.

atdotde:
I hope this works now, there was some non-trivial git gymnastics involved. As far as I can see, there are signed-off statements despite what DCO thinks.

dirkhh:
I am also puzzled why GitHub seems to think there's a DCO missing. I'll merge things locally so I can take an extra look...

PR #2906, opened 2020-08-08
Workaround mdbtools crash

salvadorcunat:

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

memcpy() called from mdbtool's mdb_ole_read_full() was throwing a segfault.
As the issue seems mdbtools or even glibc related, we need a workaround to avoid the crashing call.  Patch 2 simply avoids the faulting call although means a wider rework for some utilities  and functions.
### Changes made:

1) smtk_open_table() now builds an array of strings and an array for the lengths of the db fields (this second one on demand, just make the pointer NULL to avoid building the lengths array).  The lengths can't be built along with the MdbColumn, so we drop this last one and will build it where needed (just in main function to get dc's header and profile)
2) As we know in advance the length of the profile we can avoid calling the faulty function on zero-length profiles which seemed to cause the segfault (weirdly enough, as zero-length profiles are very common in older Aladin DC models and we never had such crashes).
3) Except for the main function smartrak_import(), use the values bounded by smtk_open_table(). In the main func we can still use them through the MdbColumn pointer as the columns are built for this function.

### DIFF
--- a/scripts/smtk2ssrf-build.sh
+++ b/scripts/smtk2ssrf-build.sh
@@ -151,7 +151,7 @@ cmake   -DBTSUPPORT=OFF \
 	build
 cd build || aborting "Couldn't cd into $SSRF_PATH/build directory"
 make clean
-make "$JOBS" || STATUS=1
+LIBRARY_PATH=$INSTALL_ROOT/lib make "$JOBS" || STATUS=1
 
 # Restore initial state of subsurface building system:
 echo "----> Restoring Subsurface tree state"
--- a/smtk-import/smartrak.c
+++ b/smtk-import/smartrak.c
@@ -182,7 +182,7 @@ static time_t smtk_timegm(struct tm *tm)
  * Returns an opened table given its name and mdb. outcol and outbounder have to be allocated
  * by the caller and are filled here.
  */
-static  MdbTableDef *smtk_open_table(MdbHandle *mdb, char *tablename, MdbColumn **outcol, char **outbounder)
+static  MdbTableDef *smtk_open_table(MdbHandle *mdb, char *tablename, char **outbounder, int *outlens[])
 {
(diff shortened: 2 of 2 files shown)


dirkhh (commented):
embarrassingly I don't understand the code well enough to really comment. This isn't something that I've ever looked into.

@atdotde does this fix the crash you reported?

atdotde (approved):
I also have no clue what this code is doing, but it parses the file that crashed the old version, so thumbs up from my side.

I just became aware that there are people trying to fork mdbtools to reanimate it, supply patches and harden it. Maybe it's worthwhile to have a look at these efforts, like https://github.com/evanmiller/mdbtools

salvadorcunat:
Hi Dirk, Robert.
Thanks for the mdbtools fork link. It's amazing to see the project active again, it has even a no-glib branch in progress.   I've been playing around this afternoon and it doesn't have the segfault in Ivar's file we were having with the original mdbtools project.
So this opens a new posibility ... Leave the current code in smartrak.c as is right now, and change the build for libmdb to this fork ... The obvious drawback of this approach is that the mdbtools packages from linux distros (at least debian and ubuntu, and probably all other) ship the 0.7.1 original version.
Best regards.

dirkhh:
Just checking - did you want to switch `mdbtools` versions instead of this PR, or should we go ahead and merge this PR?

salvadorcunat:
Hi Dirk.
The patch is harmless for both versions and I think should be applied.  If you want to wait a bit before switching mdbtools, I'm fine with it, I want to take a look at other forks before moving away from the original. But in the long run we will need to switch as the original project is completely stalled.

PR #2910, opened 2020-08-13
Add an "Edit Gas Change" right-click option.

mwerle:
This new option allows a user to select a new destination tank for an existing Gas Change event. This is useful when Subsurface's heuristics get tanks wrong after an import from a divecomputer.

### Feature Discussion
This PR is mainly to discuss whether or not such a feature is of interest.  Posting on the dev mailing list garnered little  responses save to submit the PR and see.

The "proper" way to implement this would be to fix the import heuristics. However, that is prone to danger as it might break existing workflows and heuristics. Hence this "safe" approach is proposed which gives the user full control of any changes made to their data and which will have zero impact on existing functionality.

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add UI to make it easier to fix an edge use-case when diving sidemount with two AI transmitters as well as one (or more)  deco cylinders containing a different gas mix.  Subsurface only generates 2 cylinders in this situation and gets very confused as regards sensor data vs gas switch events from the dive computer. The current desktop UI does not have an easy way to fix this situation.

The new UI menu item allows manually changing the destination cylinder of a gas switch event. It might be useful to allow more fields to be edited (eg, the timestamp; not so useful for divecomputer-supplied gas switch events, but might be nice for manually-added events).

The proposed workflow is now as follows:
1. Import from divecomputer.
2. Use UI to add additional tank(s) and fix gas mixture in all cylinders.
3. Use the new menu item to fix gas switch events to the correct cylinders.

### Changes made:

1) Add a new "Edit Gas Change" action to  ProfileWidget2
2) Implement the action to open a dialog allowing the user to choose a new cylinder.

### Related issues:

None, but has been discussed in the Google Forum as well as briefly mentioned on the developer list.

### Additional information:

TODO if PR is considered useful.

### Release note:

TODO if PR is considered useful.

### Documentation change:

TODO if PR is considered useful.

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1399,6 +1399,18 @@ struct int ProfileWidget2::getEntryFromPos(QPointF pos)
 #endif
 
 #ifndef SUBSURFACE_MOBILE
+/// Prints cylinder information for display.
+/// eg : "Cyl 1 (AL80 EAN32)"
+static QString printCylinderDescription(int i, const cylinder_t *cylinder)
+{
+	QString label = gettextFromC::tr("Cyl") + QString(" %1").arg(i+1);
+	if( cylinder != NULL ) {
+		QString mix = get_gas_string(cylinder->gasmix);
+		label += QString(" (%2 %3)").arg(cylinder->type.description).arg(mix);
+	}
+	return label;
+}
+
 void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
 {
 	if (currentState == ADD || currentState == PLAN) {
@@ -1440,12 +1452,25 @@ void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
 	QPointF scenePos = mapToScene(mapFromGlobal(event->globalPos()));
 	qreal sec_val = timeAxis->valueAt(scenePos);
 	int seconds = (sec_val < 0.0) ? 0 : (int)sec_val;
-	if (current_dive && current_dive->cylinders.nr > 1) {
+	DiveEventItem *item = dynamic_cast<DiveEventItem *>(sceneItem);
+
+	// Add or edit Gas Change
+	if (current_dive && item && event_is_gaschange(item->getEvent())) {
+		int eventTime = item->getEvent()->time.seconds;
+		QMenu *gasChange = m.addMenu(tr("Edit Gas Change"));
+		for (int i = 0; i < current_dive->cylinders.nr; i++) {
+			const cylinder_t *cylinder = get_cylinder(current_dive, i);
+			QString label = printCylinderDescription(i, cylinder);
+			gasChange->addAction(label, [this, i, eventTime] { changeGas(i, eventTime); });
+		}
+	} else if (current_dive && current_dive->cylinders.nr > 1) {
 		// if we have more than one gas, offer to switch to another one
 		QMenu *gasChange = m.addMenu(tr("Add gas change"));
-		for (int i = 0; i < current_dive->cylinders.nr; i++)
-			gasChange->addAction(QString(current_dive->cylinders.cylinders[i].type.description) + tr(" (cyl. %1)").arg(i + 1),
-					     [this, i, seconds] { changeGas(i, seconds); });
+		for (int i = 0; i < current_dive->cylinders.nr; i++) {
+			const cylinder_t *cylinder = get_cylinder(current_dive, i);
+			QString label = printCylinderDescription(i, cylinder);
+			gasChange->addAction(label, [this, i, seconds] { changeGas(i, seconds); });
+		}
 	}
 	m.addAction(tr("Add setpoint change"), [this, seconds]() { ProfileWidget2::addSetpointChange(seconds); });
 	m.addAction(tr("Add bookmark"), [this, seconds]() { addBookmark(seconds); });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: add an "Edit Gas Change" right-click option [#2910]
 core: add support for Shearwater Peregrine (requires firmware V79 or newer)
 core: fix renumbering of imported dives [#2731]
 mobile: fix editing tank information


dirkhh:
The developer mailing list is sadly very "hit and miss". Some topics create very animated discussions, many others garner no responses. 
I think this is a useful addition to the UI. There are a few issues with the PR, though.
- the commit message isn't formatted well; please keep line length to ~74 characters
- I'm not a huge fans of lambdas, but I think I lost that argument a while ago and your use of the lambda actually makes sense (that was hard to write :-) )
- your code appears to be indented with 4 spaces - we use tabs. See https://github.com/subsurface/subsurface/blob/master/CODINGSTYLE.md
- there are a few other white space issues, e.g. `if (condition)` note the space we want to see before the opening parenthesis, but no space inside the parenthesis
- in the list of cylinders to switch to, would it make sense to include the gas mix that the cylinder holds?

@bstoeger is there anything else here that I'm missing? I'm think in the context of the Undo machinery in particular.

bstoeger:
Generally I'm not very happy about the way the cylinder index is retrieved via strings. Is there no idiomatic Qt way of getting the index of the chosen item?

That said, in principle this should work, since the strings are unique. However, I wonder if it does give the correct index. In the loop the current cylinder is skipped. If the user selects a cylinder after that, wouldn't the index be off by one?

What about an alternative UI where instead of opening a dialog, the alternative cylinders are implemented as a sub-menu? This would make the whole logic more robust. On the flip side, the UI may appear funky if there is only one alternative cylinder.

bstoeger:
> I'm think in the context of the Undo machinery in particular.

I hope you don't expect me to remember what I did there...

dirkhh:
> > I'm think in the context of the Undo machinery in particular.
> 
> I hope you don't expect me to remember what I did there...

Well, err, akshually...

mwerle:
Thank you very much for your comments/replies.

> I think this is a useful addition to the UI. There are a few issues with the PR, though.
> 
I knew the PR wasn't quite "production ready" but didn't want to spend more time polishing things until I knew the core idea was ok to go ahead with.  That said, I didn't realise it was quite as bad (eg, whitespace) as it is.. (Haven't used QtCreator in a very long time)

> * in the list of cylinders to switch to, would it make sense to include the gas mix that the cylinder holds?
> 
It's something I also considered, but then decided to copy the formatting of the "Add Gas Change" UI entry for consistency.  Personally I would think that yes, this would be a good change to the UI, but should then be applied to the other action as well

Given "permission" I'm happy to modify both the existing action as well as the new one to add the gas percentage to each cylinder description.

It doesn't really make sense to switch to the cylinder currently in use, so ideally that should be stripped from the list of possible cylinders from the "Add Gas Change" as well, but would require the menu to be constantly rebuilt as the user moves the mouse.

> @bstoeger is there anything else here that I'm missing? I'm think in the context of the Undo machinery in particular.
>
Undo appears to be working due to my re-use of the existing event handling code.

> Generally I'm not very happy about the way the cylinder index is retrieved via strings. Is there no idiomatic Qt way of getting the index of the chosen item?
> 
There does not appear to be a way of retrieving an integer index from the dialog. If someone knows of a way I'd be more than happy to change this as I also think the string-matching is very clunky.

> That said, in principle this should work, since the strings are unique. However, I wonder if it does give the correct index. In the loop the current cylinder is skipped. If the user selects a cylinder after that, wouldn't the index be off by one?
> 
This is rather embarassing, you are, of course, correct. I wonder why it seemed to work when I tested it.
This also means it's completely impossible to use the current approach of string matching since there's no way of knowing how many, or which, cylinders are skipped.  An alternate approach I thought of was to parse the cylinder index directly from the string description.

> What about an alternative UI where instead of opening a dialog, the alternative cylinders are implemented as a sub-menu? This would make the whole logic more robust. On the flip side, the UI may appear funky if there is only one alternative cylinder.
>
Again, something I initially considered (for consistency with the "Add Gas Change"), but this would require the logic to rebuild the list constantly as the user moves the mouse unless we just list all cylinders all the time.  I was also thinking ahead to a more comprehensive edit feature in the future which might include being able to edit the time of the event, or possibly add a description etc.

I'm happy to implement either way.

dirkhh:
> That said, I didn't realise it was quite as bad (eg, whitespace) as it is.. (Haven't used QtCreator in a very long time)

There are sample settings for QtCreator in the  https://github.com/subsurface/subsurface/blob/master/CODINGSTYLE.md document. Just save as XML and import into QtCreator and it will mostly do the right things.

> > * in the list of cylinders to switch to, would it make sense to include the gas mix that the cylinder holds?
> 
> It's something I also considered, but then decided to copy the formatting of the "Add Gas Change" UI entry for consistency. Personally I would think that yes, this would be a good change to the UI, but should then be applied to the other action as well
> 
> Given "permission" I'm happy to modify both the existing action as well as the new one to add the gas percentage to each cylinder description.

Yes please. As a separate commit, please.

> It doesn't really make sense to switch to the cylinder currently in use, so ideally that should be stripped from the list of possible cylinders from the "Add Gas Change" as well, but would require the menu to be constantly rebuilt as the user moves the mouse.

👍 
 
> > Generally I'm not very happy about the way the cylinder index is retrieved via strings. Is there no idiomatic Qt way of getting the index of the chosen item?
> There does not appear to be a way of retrieving an integer index from the dialog. If someone knows of a way I'd be more than happy to change this as I also think the string-matching is very clunky.

Umm, err, I don't think so. But I'm no Qt Guru by any stretch of the imagination

> > That said, in principle this should work, since the strings are unique. However, I wonder if it does give the correct index. In the loop the current cylinder is skipped. If the user selects a cylinder after that, wouldn't the index be off by one?
> 
> This is rather embarassing, you are, of course, correct. I wonder why it seemed to work when I tested it.
> This also means it's completely impossible to use the current approach of string matching since there's no way of knowing how many, or which, cylinders are skipped. An alternate approach I thought of was to parse the cylinder index directly from the string description.

The index (+1) is already included in the string, right?

> > What about an alternative UI where instead of opening a dialog, the alternative cylinders are implemented as a sub-menu? This would make the whole logic more robust. On the flip side, the UI may appear funky if there is only one alternative cylinder.
> 
> Again, something I initially considered (for consistency with the "Add Gas Change"), but this would require the logic to rebuild the list constantly as the user moves the mouse unless we just list all cylinders all the time. I was also thinking ahead to a more comprehensive edit feature in the future which might include being able to edit the time of the event, or possibly add a description etc.
> 
> I'm happy to implement either way.

I think let's stay consistent with what we have.

Thanks for working on this!

mwerle:
Oops, sorry @dirkhh didn't check to see if you'd commented prior to pushing an update.  Again, please consider this a "Work In Progress" and I'm more than happy to clean it all up and start from scratch with a new pull request. (First time contributing here)

I've made the following changes:
* Removed the pop-up dialog; it was just getting too messy trying to extract the cylinder number (especially if strings end up translated) and changed "Edit Gas Change" to a cascading menu similar to the "Add Gas Change" as suggested by @bstoeger 
  - It turned out the menu was getting built from scratch every time the user right-clicks anyway..
* Only "Add Gas Change" _or_ "Edit Gas Change" is visible at any one time (doesn't make sense to have both)
* Changed the cylinder descriptions for both menu entries:
  - eg, "Cyl 1 (AL80 air)", "Cyl 2 (AL80 air)", "Cyl 3 (AL40 EAN50)", "Cyl 4 (AL40 33/10)" 
  - personally I find this layout better, but I'm happy to revert to the existing one and just add the gas mix to it.
* Fixed the whitespace

TODO:
* Currently the cylinder description string is not getting translated.  I can either duplicate the code between "Add" and "Edit" (eww) or add it as a member function (which means pulling equipment.h into the header file).  
 - Preference? Or is there a "utility class" somewhere for this kind of thing?
* I'm not filtering the list of cylinders (all cylinders are shown all the time); not sure whether it's more or less confusing either way.  Ideally we'd also filter out the "currently active" cylinder but I don't think there's any way to do that if there hasn't been a prior gas change event. Finally, I've no idea how "doubles" are treated in Subsurface which might confuse things even more.

bstoeger:
> Currently the cylinder description string is not getting translated. I can either duplicate the code between "Add" and "Edit" (eww) or add it as a member function (which means pulling equipment.h into the header file).

equipment.h is quite light-weight, so I wouldn't have a problem with that. But do you really have to include it? You could simply forward declare any struct you need in the header file. IIRC, enums are different though - you can't forward declare them.

> Preference? Or is there a "utility class" somewhere for this kind of thing?

We have a number of string-formatting in "qthelper.cpp" (which is a horrendous source file). I was contemplating of moving all the Qt-string formatting to a separate source file to clean that one up a little bit.

mwerle:
Thanks. I found the gettextfromC class which exposes "tr".  So I've left the formatting function inside profilewidget2.cpp for now since that's the only place using it.

Going diving this weekend, but I think this micro-feature is reasonably complete for now.  Please let me know if I should create a new PR with better formatted commit messages etc. and I'll get onto it next week.  

It also still needs documentation and changelog and perhaps a sample dive to demonstrate/test it.

Have a good one!

dirkhh:
The test failure is in GitStorage which despite all of my attempts to get it robust **still** occasionally shows random failures. Grmbl. Restarted - this code cannot possibly have broken git storage.

profile-widget/profilewidget2.cpp
```diff
@@ -1399,6 +1399,26 @@ struct int ProfileWidget2::getEntryFromPos(QPointF pos)
...
+	QString mix = "air";
+	int o2 = cylinder->gasmix.o2.permille / 10;
+	int he = cylinder->gasmix.he.permille / 10;
+	if (o2 > 0) {
+		if(he > 0) {
+			mix = QString("%1/%2").arg(o2,he);
+		} else {
+			mix = QString("EAN%1").arg(o2);
+		}
+	}
```
dirkhh, line 1410: instead of figuring this all out here, how about you use `get_gas_string()` from `qthelper.cpp` -- that one already deals with creating consistent names, with translations, etc.

dirkhh (changes requested):
We already have a helper function for gas strings, let's use that.
And also, I think this should all be one commit.
If you are comfortable using `git`, then `git rebase -i HEAD~3` is your friend. If not, let me know and I'll happily handle that part for you.
I still need to stare at the code some more to convince myself that this is doing the right thing :-)

profile-widget/profilewidget2.cpp
atdotde: Even for calculations you should not access members of a struct gasmix directly but use the get_o2 and get_he helpers as they treat the special case of air correctly.

mwerle: Thank you. I _did_ try searching for "EAN" because I was sure there'd be a helper function around for it, but couldn't find it. I probably did a file-local search instead of project-wide in QtCreator. In future I'll grep. ;)

Committed the changes and squashed all the commits into one, as requested, as well as keeping to the 74-char line-limit in the commit message.

dirkhh (approved):
This looks great to me. Thanks for addressing all the feedback!
@bstoeger or @atdotde - any comments?

mwerle:
> This looks great to me. Thanks for addressing all the feedback!
> @bstoeger or @atdotde - any comments?

Thank you for your patience while I muddled through this!

dirkhh:
> Thank you for your patience while I muddled through this!

Oh no, not at all - I love seeing new contributors and you are very receptive to feedback, so you made this really easy.

profile-widget/profilewidget2.cpp
```diff
@@ -1440,12 +1449,25 @@ void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
...
 	int seconds = (sec_val < 0.0) ? 0 : (int)sec_val;
-	if (current_dive && current_dive->cylinders.nr > 1) {
+	DiveEventItem *item = dynamic_cast<DiveEventItem *>(sceneItem);
+
+	// Add or edit Gas Change
+	if (item && event_is_gaschange(item->getEvent())) {
+		int eventTime = item->getEvent()->time.seconds;
+		QMenu *gasChange = m.addMenu(tr("Edit Gas Change"));
+		for (int i = 0; i < current_dive->cylinders.nr; i++) {
+			cylinder_t *cylinder = &current_dive->cylinders.cylinders[i];
```
bstoeger: I suggest
```
        const cylinder_t *cylinder = get_cylinder(current_dive, i);
```
Also I'm not sure that in the planner `current_dive` is correct - do we have to use `&displayed_dive` there? I wish we could get finally rid of displayed_dive. :-/

bstoeger: Ok - I reckon that these gas_change events don't even exist in the planner, so it should be fine.

Still, I would prefer if a check for `current_dive` was added to the if above so that there is not even a theoretical chance of a null-pointer access here.

profile-widget/profilewidget2.cpp
```diff
@@ -1399,6 +1399,15 @@ struct int ProfileWidget2::getEntryFromPos(QPointF pos)
 #endif
 
 #ifndef SUBSURFACE_MOBILE
+/// Prints cylinder information for display.
+/// eg : "Cyl 1 (AL80 EAN32)"
+static QString printCylinderDescription(int i, cylinder_t *cylinder)
```
bstoeger: `const cylinder_t *cylinder`

profile-widget/profilewidget2.cpp
```diff
@@ -1440,12 +1449,25 @@ void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
...
+			gasChange->addAction(label, [this, i, eventTime] { changeGas(i, eventTime); });
+		}
+	} else if (current_dive && current_dive->cylinders.nr > 1) {
 		// if we have more than one gas, offer to switch to another one
 		QMenu *gasChange = m.addMenu(tr("Add gas change"));
-		for (int i = 0; i < current_dive->cylinders.nr; i++)
-			gasChange->addAction(QString(current_dive->cylinders.cylinders[i].type.description) + tr(" (cyl. %1)").arg(i + 1),
-					     [this, i, seconds] { changeGas(i, seconds); });
+		for (int i = 0; i < current_dive->cylinders.nr; i++) {
+			cylinder_t *cylinder = &current_dive->cylinders.cylinders[i];
```
bstoeger: Since you're changing this, perhaps switch to the more palatable `get_cylinder(current_dive, i)`.

bstoeger:
I didn't test it - in general fine with me, up to a few minor comments.

profile-widget/profilewidget2.cpp
mwerle: Seems a bit inefficient to have a loop which then uses an accessor function which re-iterates over the same array, just to ensure a check is made which doesn't apply in this situation,  I know this isn't performance-critical code, but still..

All that get_cylinder() currently does is a check in the Planner due to some hack (according to the comments) - however, the Planner doesn't even have a right-click menu, so this code will never even get triggered in the Planner.

Happy to change it regardless if you really think it improves the codebase. Although the "correct" way to fix it would be to properly encapsulate any types whose data should not be accessed directly. That would then force all code accessing the data to be updated instead of having an inconsistent codebase where some parts use accessors and other parts direct access.

bstoeger: > Seems a bit inefficient to have a loop which then uses an accessor function which re-iterates over the same array, just to ensure a check is made which doesn't apply in this situation, I know this isn't performance-critical code, but still..

This is an absolutely negligible cost compared to the string manipulation that you do in the next line - not even talking about the expense of showing the menu on the screen. That's a complete non-issue.

However, we get:
- More readable code.
- More flexible code - we can change the layout of the cylinders in the dive.
- Safer code - it does bound check. You say it "doesn't apply in this situation", which is correct, but only if you didn't make a typo (e.g. loop over "j" and access "i").
- More consistent code - the number of "get_cylinder()" calls outnumbers the direct accesses.

So overall this appears to be a win.

> All that get_cylinder() currently does is a check in the Planner due to some hack (according to the comments) - however, the Planner doesn't even have a right-click menu, so this code will never even get triggered in the Planner.

You misread the comment (probably because it was poorly written): The check is not there for the planner. The check was _relaxed_ for the planner: You can now access one past the last "official" number of cylinders. There is always one "reserve" cylinder, which represents air. It's a sad story.

> Although the "correct" way to fix it would be to properly encapsulate any types whose data should not be accessed directly. That would then force all code accessing the data to be updated instead of having an inconsistent codebase where some parts use accessors and other parts direct access.

I would only do this if we change the core to C++, because C++ has all the necessary encapsulation features. I'm against "simulating" C++ in C - it can be done, but is very annoying. We could then simply change to `std::vector<cylinder_t>`. I checked and there don't seem to be many direct accesses to the cylinders array left. I will write a cleanup commit.

mwerle: Sorry, yes I know the performance hit is absolutely negligible; it's just the _obviousness_ of traversing the array twice that triggered my OCD.

As for safer, definitely, no arguments.

As for more readable... well.. given we're iterating in a loop which does a deep access of the data structures, to me personally, it's less readable. More readable would be if a we also had a get_num_cylinders() or, ideally, a get_first_cylinder() and get_next_cylinder().  
Your clean-up commit primarily accesses which were iterating over the cylinders instead of just accessing a particular cylinder, so it seems this iteration over cylinders is quite a common pattern in the codebase.

Just my 2c.  Happy to make the change with no further argument needed :)

bstoeger: Sure - feel free to add a `get_num_cylinders()` function or even a `for_each_cylinder` macro akin to the `for_each_dive` macro.

We *could* define `begin()` and `end()` functions for C++ code, so that one can write
```
   for(const cylinder_t &cyl: d->cylinders) {
   }
```

However, if we want to start introducing C++-isms into core, that would need a discussion first.

bstoeger (approved):
Looks good to me. I don't consider the cylinder access important enough to hold this back.

mwerle:
> Looks good to me. I don't consider the cylinder access important enough to hold this back.

There was the potential null reference as well.

Just woke up (insomnia, yayy), so figured I may as well finish this.  I rebased the PR on the current head and fixed the remaining concerns.  If ok, I will merge the 3 commits into one, or revert to the approved commit if not.

bstoeger:
From my side, this can be merged. I think the commits can be squashed, but let's see what @dirkhh says.

dirkhh (commented):
please refactor this so it's one or two commit(s) - I usually like small incremental commits and I'd be certainly happy with the third one staying separate - but really, if I look at the whole change as one diff, that makes the most sense to me. 
Once you push that I can merge.

mwerle:
> please refactor this so it's one or two commit(s) - I usually like small incremental commits and I'd be certainly happy with the third one staying separate - but really, if I look at the whole change as one diff, that makes the most sense to me.
> Once you push that I can merge.

Thank you, I just wanted confirmation that the overall  change is ok before squashing. Easier to revert things while they are still separate commits :)

mwerle:
Also added CHANGELOG entry.

dirkhh (approved):
this looks good. in an abundance of caution I'll wait for the builds to complete, but I expect that I'll merge that later today.

PR #2912, opened 2020-08-15
Fix event merging when merging dives

torvalds:
The merge_events() function was subtly and not-so-subtly broken in a
couple of ways:

 - in commit 8c2383b49 ("Undo: don't modify source-dives on merge"), we
   stopped walking the event list after we merged the first event from a
   dive when the other dive computer had run out of events.

   In particular, this meant that when merging consecutive dives, the
   second dive only had the first event copied over to the merged dive.

   This happened because the original code just moved the whole old list
   over when there was nothing left from the other dive, so the old code
   didn't need to iterate over the event list. The new code didn't
   realize that the pointer movement used to copy the whole rest of the
   list, and also stopped iterating.

   In all fairness, the new code did get the time offset right, which
   the old code didn't. So this was always buggy.

 - similarly, the "avoid redundant gas changes" case was not handled for
   the "we ran out of events for the other dive computer" case.

This fixes both issues.

Cc: Berthold Stoeger <bstoeger@mail.tuwien.ac.at>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1962,20 +1962,15 @@ static void merge_events(struct dive *d, struct divecomputer *res,
 		const int *cylinders_map;
 		int event_offset;
 
-		if (!b) {
-			*p = clone_event(a);
-			event_renumber(*p, cylinders_map1);
-			break;
-		}
-		if (!a) {
-			*p = clone_event(b);
-			(*p)->time.seconds += offset;
-			event_renumber(*p, cylinders_map2);
-			break;
-		}
+		if (!b)
+			goto pick_a;
+
+		if (!a)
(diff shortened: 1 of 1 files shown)


bstoeger (approved):
That looks very reasonable to me.

dirkhh (approved):
Does this fix the crash as a side effect that I don't see? Or is this unrelated to the crash and just fixes other bugs.
The code looks reasonable, so I'll merge.

Issue #2913, opened 2020-08-15
Subsurface-mobile loses gas info when editing start/end pressure in multi cylinder dive


dirkhh:
### Describe the issue:

- [x] Bug

### Issue long description:

user reported bug

### Operating system:

Android 10

### Subsurface version:

3.0.10

### Steps to reproduce:

- download multi gas dive (in user's case, from OSTC 2)
- edit start end pressure

### Current behavior:

- gas info per tank is lost, all tanks same gas as first tank

### Expected behavior:

- don't mess with the tank info

PR #2914, opened 2020-08-15
Fix mobile multi tank

dirkhh:

- [x] Bug fix
- [x] Code cleanup

Triggered by a bug report for Subsurface-mobile I played with the dive editing code and multiple cylinders and fixed the bug that was reported and a second one that I found in the process of debugging the initial bug

### Changes made:

1) fix cut and paste issue that made us lose the gas information of all but the first cylinder
2) fix localized parsing of pressure values

### Related issues:

#2913 

### Release note:

included

### Documentation change:

not needed

### Mentions:

@bstoeger would appreciate a second pair of eyes - or any of the other maintainers...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+mobile: fix editing tank information
 mobile: disable download button if no connection is selected
 mobile: fix incorrect time stamps on GPS track points created via location service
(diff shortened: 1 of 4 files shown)


dirkhh:
No comments, but even after re-reading the changes I can't see how this would be wrong, so let's merge.

PR #2916, opened 2020-08-17
Zerolength

atdotde:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 mobile: fix editing tank information
+planner: Handle zero length segments gracefully when replanning
 mobile: disable download button if no connection is selected
 mobile: fix incorrect time stamps on GPS track points created via location service
 core: correctly recognize A1 as BLE dive computer
--- a/desktop-widgets/locationinformation.h
+++ b/desktop-widgets/locationinformation.h
@@ -10,7 +10,7 @@
 #include <QAbstractListModel>
 #include <QSortFilterProxyModel>
 
-class dive_site;
+struct dive_site;
 
 class LocationInformationWidget : public QGroupBox {
 	Q_OBJECT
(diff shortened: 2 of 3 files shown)


dirkhh (approved):
I had wanted to fix that compiler warning so many times... and always get distracted.
Change looks reasonable.

bstoeger:
If you add ten seconds here, then I think you should also not allow generating such degenerate segments in the first place. I think infinite ascent / descent rates should be disallowed. These are just calling for trouble.

Moreover, I'm not sure how to feel about the arbitrary 10 sec? It would feel much more logical to me to limit the ascent / descent rate to some sane value.

bstoeger:
> I had wanted to fix that compiler warning so many times... and always get distracted.

Personally, that's just one of those warning that I would disable - because ... what does it achieve?

atdotde:
I thought about not allowing zero length transitions in the first place but decided against this as our time granularity of integer minutes in the planner is not very flexible (and is also in the way of having good limits on the velocities. Maybe we can weaken that granularity without forcing users to always type :00 when they don't want to.

bstoeger:
Up to you - you certainly know the planner better than anyone else. Just consider that you could still add the same "fixup" code when creating the segment in the first place. To me, it just doesn't feel right that the planned dive is modified once you load it. Either these segments are a problem, or they aren't... (personally I think they shouldn't exist, because they are unphysical)?

dirkhh (commented):
The last commit message cracks me up: "You cannot be at the same depth at different times"... I'm pretty sure you can... on the other hand, being at different depths at the same time, that seems harder :-) :-) :-)

bstoeger (approved):
Looks good to me. The logic with the `secs` and `duration` is a bit confusing to me. I reckon that you have thought this through.

dirkhh (approved):
thanks for addressing the small issues. There's one more typo in the commit message, but instead of sending you back, I'll just fix it myself and merge locally.

dirkhh:
merged

PR #2920, opened 2020-08-20
cleanup: consistently use get_cylinder() accessor

bstoeger:
get_cylinder(d, i) is more readable than d->cylinders.cylinders[i].
Moreover, it does bound checking and is more flexible with respect to
changing the core data structures. Most places already used this accessor,
but some still accessed the cylinders directly.

This patch unifies the accesses by consistently switching to get_cylinder().
The affected code is in C++ and accesses the cylinder as reference or
object, whereas the get_cylinder() function is C and returns a pointer.
This results in funky looking "*get_cylinder(d, i)" expressions.
Arguably still better than the original.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A minor code cleanup, which switches to `get_cylinder()`. Details in the commit message. I left `profilewidget2.cpp` alone - to avoid conflict with #2910. Only compile-tested.

### Changes made:

1) Switch the few remaining direct cylinder accesses to `get_cylinder()`.

### Related issues:

See discussion in #2910.

### Release note:

No.

### Documentation change:

Hell, no.

### Mentions:

@dirkhh @mwerle

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -668,7 +668,7 @@ PasteState::PasteState(dive *dIn, const dive *data, dive_components what) : d(dI
 		// 2) For cylinders that do not yet exist in the destination dive, we set the pressures to 0, i.e. unset.
 		//    Moreover, for these we set the manually_added flag, because they weren't downloaded from a DC.
 		for (int i = 0; i < d->cylinders.nr && i < cylinders.nr; ++i) {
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
Yay - bounds checking everywhere. What a concept. We should do more of that.
Thanks.

PR #2922, opened 2020-08-20
[semi-serious] Proof of concept: add C++ iterators to dive_table

bstoeger:
This is not intended for merging, therefore I'm not writing a
full commit message. It proves how trivial it is to add C++
style iterators to our table structures.

In this case, there is not much gain (except less polution of
outer scope). However, for cylinders, weights, etc. this might
make the code distinctly nicer.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is not intended for merging - just for discussion: In #2910, we talked about better looping through C-tables. This demonstrates how it could be done. For the dive_table it doesn't make much of a difference code wise, but for cylinders, etc. It might give distinctly more readable code. We could even auto generate the iterator functions in the table-macros.

### Changes made:

1) Add C++ iterators to dive_table

### Mentions:

@dirkhh @mwerle


bstoeger:
To see the advantage of the notation, see the following diff:
```
-       for (int i = 0; i < trip1->dives.nr; ++i)
-               divesToMove.divesToMove.push_back( { trip1->dives.dives[i], newTrip } );
-       for (int i = 0; i < trip2->dives.nr; ++i)
-               divesToMove.divesToMove.push_back( { trip2->dives.dives[i], newTrip } );
+       for (dive *d: trip1->dives)
+               divesToMove.divesToMove.push_back( { d, newTrip } );
+       for (dive *d: trip2->dives)
+               divesToMove.divesToMove.push_back( { d, newTrip } );
```
Distinctly more expressive and less error prone.

bstoeger:
Converted the creation of the iterators into a macro and converted some dive_site-table loops.

dirkhh (commented):
I am certainly inclined to merge a set of changes that implement this for our tables.

mwerle:
> I suggest that you use `get_cylinder()` as that is consistent with the rest of the code, especially now that #2920 has been merged. If you later want to add a `for_each_cylinder` macro, I suggest that you also add a `for_each_used_cylinder` macro, since a quick survey suggests that this is approximately half of the use cases.
> 
Will do, thank you.

> I have no idea why you guys are making this a C vs C++ issue. When I said "more expressive" I was explicitly comparing these two non-macro snippets:
> 
I wasn't making it an issue; merely stating that this particular C++ syntax is something I'm not familiar with.  While I can read it, I wouldn't have been able to write it. 

Despite that, overall I'd have been in favour of having added this proposed change. Anyway.

bstoeger:
The thing is - the part where I thought it would bring the most (looping over cylinders) it did bring the least, because usually the index is needed to check whether the cylinder is used (because the cylinder index is stored in divecomputer samples). Of course one could simply write an iterator adapter such that one can do
```
for (const cylinder_t &cyl: used_iterator(d->cylinders))
        do_something(cyl);
```
This leaves a few cases where indexes are needed to access some flag-array. Here again, one could write a simple adapter such that one can use structured bindings à la
```
for (const auto & [idx, cyl]: idx_iterator(d->cylinders))
        do_something(idx, cyl);
```
But that would need C++17 or awkward `std::tie` statements.

And then, this all works only for C++ parts of the code. So I wonder if it is worth it. If there really is interest, I can try to come up with something, but I think we have other top-priority issues at the moment.

dirkhh:
I don't think this is a high priority. 

A quick comment on your WHO comment above, though... you know... I guess we'll just have to disagree because I can tell right away what the first code does, and I have to think what the second one does because that's just not how my mind is used to think about working with arrays. So yes, my brain may be damaged, but unless you would like to take over as Subsurface maintainer (trust me, I have considered looking for a new maintainer quite a bit over the past year), some patience for my addled old brain might still be required.

That said, NO, I don't think that this is the wrong direction. I'm sure my brain will get used to that, just as I am slowly (about six years after I rejected the first such patch in disgust) am wrapping my mind around lambdas (even though, clearly, I'm still doing it wrong). So I'm open to a thoughtful introduction of such iterators because the one aspect where I do agree with you is that this pattern is less fragile and less error prone.

bstoeger:
Sorry about that comment - it was very stupid even if it was made in jest. Just because I tend to be a self-deprecating cynic, doesn't allow me to treat others like that. Sometimes I get frustrated when I feel that people don't respond to what I write (which is often due to my communication being very brief).

I think we're still talking cross-purposes here. Whether a notation is expressive or not is independent from how familiar you are with it. The whole point of the range-based for is that it abstracts from the underlying container. It simply says: for every item in the container do this. If we change the array to a linked list, the for loop can stay the same. We just have to change the iterator.

However, that is completely independent from which notation is easier to read to you. To give an example from mathematics: Leibnitz' differentiation notation is superior to Newtons'. Still, I mostly use the latter, because I never got used to the former and very rarely need multi-variable calculus.

PR #2923, opened 2020-08-20
Update to current libdivecomputer

torvalds:
The libdivecomputer internals changed for USB devices, and now we need
to scan the USB devices before calling libdivecomputer.  That's the same
pattern as for USBHID and IRDA, so let's just regularize this all.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -25,6 +25,7 @@
 
 #include <libdivecomputer/version.h>
 #include <libdivecomputer/usbhid.h>
+#include <libdivecomputer/usb.h>
 #include <libdivecomputer/serial.h>
 #include <libdivecomputer/irda.h>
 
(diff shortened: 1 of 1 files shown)


bstoeger:
It seems like the windows build is not happy about the `interface` identifier. To be honest, I can't make any sense of the error message. Could `interface` have some weird special meaning on Windows?

bstoeger:
For reference: according to stackoverflow, it does
```
# define interface  struct
```
How weird is that. :-D

dirkhh (approved):
looks nice - I can't get my Cobalt to take a charge right now, so I'll merge without being able to test that part :-(

PR #2925, opened 2020-08-21
[TRIVIAL] cleanup: remove obsolete logic in getFormattedCylinder()

bstoeger:
getFormattedCylinder() is a helper function to format a list
of cylinders. It had that weird logic that it would skip
cylinders without description unless it is the first, which
would instead be written as "unkown".

The reason was the old statically sized cylinder array,
where it wasn't clear if a cylinder was actually in use.

This became obsolete when switching to a variable size
cylinder array. Firstly, all cylinders in the array were added
by the user. Secondly, we now also support dives without
cylinders, i.e. the first cylinder is not any different from
the rest.

Thus, remove the logic and format any cylinder without
description as being of type "unknown".

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a trivial cleanup that removes now obsolete logic.

### Changes made:

1) Format cylinders without description as "unknown", not only the first cylinder.

### Mentions:

@dirkhh

### DIFF
--- a/core/subsurface-qt/diveobjecthelper.cpp
+++ b/core/subsurface-qt/diveobjecthelper.cpp
@@ -35,8 +35,6 @@ static QString getFormattedCylinder(const struct dive *dive, int idx)
 {
 	const cylinder_t *cyl = get_cylinder(dive, idx);
 	const char *desc = cyl->type.description;
-	if (!desc && idx > 0)
-		return QString();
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
this seems very reasonable

PR #2927, opened 2020-08-27
DataTrak import: Add support for WLog extensions.

salvadorcunat:

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

WLog is a Win32 based ancient shareware program whose target was:
1) fully support divelogs coming from DataTrak (DOS or Win)
2) fill some meaningful data which wasn't supported by Uwatec software
3) have a more user-friendly GUI than Datatrak had
The problem achieving goals 1) and 2) at the same time was solved by
adding a complementary file with .add extension and - mandatory - same
base name than .log file (including directory tree).

This .add file has a fixed structure composed of a 12 bytes header,
including file type check and Nº of dives following; then a fixed 850
bytes size for each dive in the log file. Data fields size and position
are fixed inside these blocks and heavily zero padded, so they are easy
to parse.

A serious restriction imposed to the WLog user was *Do not edit the logs
with other software than Wlog*; this was due the order of dives in .log
file being the same than the order of dives in .add file. Thought you
could show a WLog divelog in Datatrak, editing it resulted in mixing all
extended data for dives following the edited one.
Thus, we have to trust files are correct and is to the user ensure this
is so. If extended data are mangled, they are mangled in WLog too and we
are not trying to fix the mess, just importing.

On the technical side, we try to be smart about tank names as neither
DataTrak nor WLog record them. So we just take the first tank in users
list matching the volume recorded in WLog.
For weights we add a translatable "unknown" string as an empty string
results in weight not being shown in subsurface-mobile (which could be a
reportable issue, BTW).

### DIFF
--- a/core/datatrak.c
+++ b/core/datatrak.c
@@ -582,7 +582,7 @@ bail:
  * Parses the header of the .add file, returns the number of dives in
  * the archive (must be the same than number of dives in .log file).
  */
-static unsigned int wlog_header_parser (struct memblock *mem)
+static int wlog_header_parser (struct memblock *mem)
 {
 	int tmp;
 	unsigned char *runner = (unsigned char *) mem->buffer;
--- a/core/datatrak.c
+++ b/core/datatrak.c
@@ -128,7 +128,7 @@ static int dtrak_prepare_data(int model, device_data_t *dev_data)
  * Just get the first in the user's list for given size.
  * Reaching the end of the list means there is no tank of this size.
  */
-static char *cyl_type_by_size(int size)
+static const char *cyl_type_by_size(int size)
 {
 	struct tank_info_t *ti = tank_info;
 
--- a/core/datatrak.c
+++ b/core/datatrak.c
@@ -124,6 +124,23 @@ static int dtrak_prepare_data(int model, device_data_t *dev_data)
 }
 
 /*
+ * Return a default name for a tank based on it's size.
+ * Just get the first in the user's list for given size.
+ * Reaching the end of the list means there is no tank of this size.
+ */
+static char *cyl_type_by_size(int size)
+{
+	struct tank_info_t *ti = tank_info;
+
+	while (ti->ml != size && ti < tank_info + MAX_TANK_INFO)
+		ti++;
+	if (ti == tank_info + MAX_TANK_INFO)
+		return "";
+	else
+		return copy_string(ti->name);
+}
+
+/*
  * Reads the size of a datatrak profile from actual position in buffer *ptr,
  * zero padds it with a faked header and inserts the model number for
  * libdivecomputer parsing. Puts the completed buffer in a pre-allocated
@@ -320,7 +337,7 @@ static unsigned char *dt_dive_parser(unsigned char *runner, struct dive *dt_dive
 	if (tmp_2bytes != 0x7FFF) {
 		cylinder_t cyl = empty_cylinder;
 		cyl.type.size.mliter = tmp_2bytes * 10;
-		cyl.type.description = "";
+		cyl.type.description = cyl_type_by_size(tmp_2bytes * 10);
 		cyl.start.mbar = 200000;
 		cyl.gasmix.he.permille = 0;
 		cyl.gasmix.o2.permille = 210;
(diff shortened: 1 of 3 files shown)


core/datatrak.c
```diff
@@ -123,6 +123,23 @@ static int dtrak_prepare_data(int model, device_data_t *dev_data)
...
+ * Reaching the end of the list means there is no tank of this size.
+ */
+static char *cyl_type_by_size(int size)
+{
+	struct tank_info_t *ti = tank_info;
+
+	while (ti->ml != size && ti < tank_info + MAX_TANK_INFO)
+		ti++;
+	if (ti == tank_info + MAX_TANK_INFO)
+		return "";
```
bstoeger, line 138: The memory management here is messed up. Below you return a copy of a string, here a static string - that can't be right. This should give a fat compiler warning (conversion of `const char *` to `char *`).

PS: Yes, I am aware that this condition is nearly impossible to reach. However, that makes it even more severe - it may lead to very obscure crashes.

bstoeger: Reading the code below - this particular line is probably correct, but you should loose the copy_string below. The cylinder appears to be cloned anyway. Or you copy the string this function and then use `add_cylinder` instead of `add_cloned_cylinder`. Of course that should be commented. The joys of C-style memory management.

core/datatrak.c
```diff
@@ -561,11 +578,105 @@ static unsigned char *dt_dive_parser(unsigned char *runner, struct dive *dt_dive
 	free(devdata);
 	return NULL;
 }
+/*
+ * Parses the header of the .add file, returns the number of dives in
+ * the archive (must be the same than number of dives in .log file).
+ */
+static unsigned int wlog_header_parser (struct memblock *mem)
```
bstoeger: The function is declared as `unsigned int`, but the result is later assigned to an `int`.

Generally, we should avoid unsigned arithmetic like the plague.

core/datatrak.c
```diff
@@ -561,11 +578,105 @@ static unsigned char *dt_dive_parser(unsigned char *runner, struct dive *dt_dive
...
+{
+	int tmp;
+	unsigned char *runner = (unsigned char *) mem->buffer;
+	if (!runner)
+		return -1;
```
bstoeger, line 590: Return `-1` in an `unsigned int` function. That's obscure..?

core/datatrak.c
```diff
@@ -561,11 +578,105 @@ static unsigned char *dt_dive_parser(unsigned char *runner, struct dive *dt_dive
...
+	} else if (wlog_notes) {
+		dt_dive->notes = copy_string(wlog_notes);
+	}
+	free(buffer);
+	free(wlog_notes);
+
+	/*
+	 * Weight in Kg * 100
+	 */
+	tmp = (runner[pos_weight + 1] << 8) + runner[pos_weight];
```
bstoeger: Perhaps create a helper function to access little endian 16-bit values? If that doesn't exist already.

core/datatrak.c
salvadorcunat: Yes, using add_cloned_cylinder() works well.  I don't think this should be moved to add_cylinder() as we are not freeing the original string anyway.
On the previous comment, I'm not getting any warning because of "" const char * being returned.

core/datatrak.c
salvadorcunat: The function doesn't really need to be unsigned as I don't expect anyone to have such a big number of dives logged. Returning a negative value on error is mandatory, BTW, as 0 and positives can be expected correct values.

core/datatrak.c
salvadorcunat: Yes I had one previously defined and widely used along the hole file. Just forgot about it.

salvadorcunat:
Just added 3 patches which fix Berthold's concerns.

dirkhh:
waiting to make sure @bstoeger is happy :-)

core/datatrak.c
```diff
@@ -123,6 +123,23 @@ static int dtrak_prepare_data(int model, device_data_t *dev_data)
 	return g_models[i].libdc_num;
 }
 
+/*
+ * Return a default name for a tank based on it's size.
+ * Just get the first in the user's list for given size.
+ * Reaching the end of the list means there is no tank of this size.
+ */
+static char *cyl_type_by_size(int size)
```
bstoeger: Sorry for being pedantic: But could we change the function signature to:
```
static const char *cyl_type_by_size(int size)
```
?
Returning `char *` softly implies that the caller owns the string and should free it. Returning `const char *` implies that this is owned by others and caller should duplicate it if it wants to take ownership. There are exceptions though.

bstoeger:
Looks fine to me - though I didn't check memory management in detail.

core/datatrak.c
salvadorcunat: > Sorry for being pedantic

You are not, it's a good practice.
Have just sent another patch.

core/datatrak.c
```diff
@@ -588,10 +588,9 @@ static int wlog_header_parser (struct memblock *mem)
 	unsigned char *runner = (unsigned char *) mem->buffer;
 	if (!runner)
 		return -1;
-	tmp = (runner[1] << 8) + runner[0];
-	if (tmp == 0x0252) {
+	if (!memcmp(runner, "\x52\x02", 2)) {
 		runner += 8;
-		tmp = (runner[1] << 8) + runner[0];
+		tmp = (int) two_bytes_to_int(runner[1], runner[0]);
```
dirkhh, line 593: signed vs unsigned integer math is always something that makes me nervous. If someone for some stupid reason has a dive number 33000 (and that is a LOT of diving), this will do weird things. But... yeah.
I would prefer to have an error check for this situation, TBH, but we have many other spots with similar issues. And I consider this an exceptionally rare use case for Subsurface in the first place.

PR #2931, opened 2020-09-02
[Bug #2930] Fix crash bug in LocationInformationWidget

### LINKED ISSUE #2930: Crash bug in reverseGeocode
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

While editing dive sites I kept having Subsurface crash (see also #2929 ).  Ran a debug version of Subsurface in QTEditor and hit a SIGSEGV in command_divesite.cpp:330.

Turns out "ds" is NULL, which, based on the stack trace, makes no sense unless another thread has interfered in the meantime since there is a check for this in the calling chain (localinformation.cpp:317)

So possibly a missing mutex somewhere.

### Operating system:

Debian GNU/Linux 10, british english

Desktop PC

Yes, Windows 10

### Subsurface version:

Debian: 4.9.6 compiled from f5ba42e933ae4701f40ba8b8b8058e5dd8c82cc0

Yes, Windows 10 4.9.6 official release

### Steps to reproduce:

No specific repro steps; edit dive sites doing geo lookups. Sooner or later you should get this crash.

### Current behavior:

Program crashes and exits, losing all edits.

### Expected behavior:

Program handles the issue in some sane manner which does not result in a crash nor loss of data.

### Additional information:

[subsurface_reverseGeocode_crash_stacktrace.txt](https://github.com/subsurface/subsurface/files/5160450/subsurface_reverseGeocode_crash_stacktrace.txt)

mwerle:
If a user exits the LocationInformationWidget (Edit  Dive Site)
while a reverseGeocode lookup is in progress, the object's
diveSite variable is set to null.

When the reverseGeocode lookup completes, this variable is
dereferenced causing an application crash.

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixed by keeping a local reference to the divesite.  This means that when the reverseGeocode lookup completes, the information is entered into the divesite.

An alternative approach would be to check the variable again (or only check it after the geocode lookup) and abort the process if it is now null.

### Related issues:

Fixes #2930 

### Additional information:

To repro and test:

1. Edit a dive site.
2. Click on the "Reverse Geocode Lookup" button
3. As quickly as possible, click the "Done" button

### Release note:

N/A

### Documentation change:

N/A

### DIFF
--- a/desktop-widgets/locationinformation.cpp
+++ b/desktop-widgets/locationinformation.cpp
@@ -313,12 +313,15 @@ void LocationInformationWidget::on_diveSiteDistance_textChanged(const QString &s
 
 void LocationInformationWidget::reverseGeocode()
 {
+	dive_site *ds = diveSite; /* Save local copy; possibility of user closing the widget while reverseGeoLookup is running (see #2930) */
 	location_t location = parseGpsText(ui.diveSiteCoordinates->text());
-	if (!diveSite || !has_location(&location))
+	if (!ds || !has_location(&location))
 		return;
 	taxonomy_data taxonomy = { 0, 0 };
 	reverseGeoLookup(location.lat, location.lon, &taxonomy);
-	Command::editDiveSiteTaxonomy(diveSite, taxonomy);
+	if (ds != diveSite)
+		return;
+	Command::editDiveSiteTaxonomy(ds, taxonomy);
 }
 
 DiveLocationFilterProxyModel::DiveLocationFilterProxyModel(QObject *) : currentLocation(zero_location)


atdotde:
I am not sure this is the correct treatment of the race condition (which seems to arise from the last line of 
`void LocationInformationWidget::acceptChanges()`. When I press "done" I don't expect future changes due to geolookups still taking place. And strictly speaking, how do we know that that ds pointer is still valid if the geo lookup returns hours later?
Could we stop all longing lookups before invalidating divesite?
Maybe @bstoeger could comment, I guess he wrote that comment in acceptChanges.

mwerle:
@atdotde I agree, hence my comment in the pull request.

I see several plausible ways to fix it:
1) As I have done
2) Disable the "Done" button (and any other way to exit the widget) while an operation is in progress.
3) Add mutexes around checks (and subsequent uses of the checked state) wherever an operation could result in such a race condition (effectively also disables the "Done" button, but in an invisible manner to the user).
4) Add an internal "Cancel" operation which is triggered when "Done" is pressed which all operations must continuously check for and honour, aborting and reverting any partially-completed processes.

My approach was merely the simplest and quickest, but I do agree that it might result in some "magical" unexpected background updates.

ED: That said, IMHO a crash bug which loses people's work is of a "must-fix-now" priority even if the fix is suboptimal until a more complete solution can be engineered.

bstoeger:
> I am not sure this is the correct treatment of the race condition (which seems to arise from the last line of
void LocationInformationWidget::acceptChanges(). When I press "done" I don't expect future changes due to geolookups still taking place. And strictly speaking, how do we know that that ds pointer is still valid if the geo lookup returns hours later?
Could we stop all longing lookups before invalidating divesite?

Agree - dangling pointers are not a viable solution. Technically, this could be solved by shared and weak pointers: We could keep a weak pointer to the corresponding undo command and update that if it is still the last command in flight. However, from an UI perspective I find this asynchronous behavior very weird.

I believe this should be made synchronous: don't accept changes while a geo-lookup is in flight and give the possibility of stopping geo-lookup. Alternatively, when the user presses "accept", trash any running geo-lookup. This should be realizable with QFuture, etc., no?

@mwerle: Check QFuture and related things - I believe that this is much easier to solve.

bstoeger:
Also, you say that the crash is caused by a null-pointer dereference. Why not just add a check at the respective place instead of generating dangling pointers?

atdotde:
As long as the check is not atomic, it does not help with a race condition.

bstoeger:
I understand that - I was commenting with respect to a "quick fix", where I think the nullptr check is preferable over a dangling pointer. Though I haven't looked closely and am not even sure what is saved in the geo-thread.

Correct fix is QFuture et al. I might try that tonight. It's not like we get a new release any minute now...

atdotde:
I am pretty sure, that the quick fix eliminates 99.9% of the crashes. It's just that those hacks tend to last forever before being properly addressed.

bstoeger:
And I believe that the nullptr-check will fix even more crashes. After all, the only way to crash with a nullptr-check is the situation where the geo-thread reads out the pointer (meaning that the dive site exists) and the dive site disappears within the few ns that it takes to update the string, no? A very unlikely scenario.

With the dangling pointer, a hung thread could be revived after a few seconds which gives a much larger window where the pointer could become invalid.

The only disadvantage would be that the *wrong* divesite could be edited. To protect from that you could compare the old and the new pointer:
```
void LocationInformationWidget::reverseGeocode()
{
	dive_site *old_ds = diveSite; // Save old dive site so we are sure that it didn't disappear under our feet
	location_t location = parseGpsText(ui.diveSiteCoordinates->text());
	if (!diveSite || !has_location(&location))
		return;
	taxonomy_data taxonomy = { 0, 0 };
	if (old_ds != diveSite)
		return;
	Command::editDiveSiteTaxonomy(diveSite, taxonomy);
}
```
Thinking about this - this could even be the proper fix, without QFuture.

bstoeger:
> And I believe that the nullptr-check will fix even more crashes. After all, the only way to crash with a nullptr-check is the situation where the geo-thread reads out the pointer (meaning that the dive site exists) and the dive site disappears within the few ns that it takes to update the string, no? A very unlikely scenario.

Or probably even an impossible scenario - we're running this code in the main thread, isn't it? That makes all discussion about mutexes, etc moot. I will check that out closer tonight.

bstoeger:
In my tests, the setting and clearing of the `diveSite` variable was all done in the same thread. Therefore, I think the whole talk about mutexes, critical sections, etc. is a red herring.

I would suggest the following fix:
```
diff --git a/desktop-widgets/locationinformation.cpp b/desktop-widgets/locationinformation.cpp
index b8a642557..85196b196 100644
--- a/desktop-widgets/locationinformation.cpp
+++ b/desktop-widgets/locationinformation.cpp
@@ -317,7 +317,17 @@ void LocationInformationWidget::reverseGeocode()
 	if (!diveSite || !has_location(&location))
 		return;
 	taxonomy_data taxonomy = { 0, 0 };
+
+	// Remember the dive site so that we can check whether the dive site changed
+	// while we were doing the reverse geo-lookup in a separate thread.
+	const dive_site *oldDs = dive_site;
+
+	// The currently edited dive site changed while we did the reverse
+	// geo lookup. Let's throw out the results.
+	if (oldDs != diveSite)
+		return;
+
 	Command::editDiveSiteTaxonomy(diveSite, taxonomy);
 }
```

That said - I couldn't reproduce the problem, since for me reverse-geo lookup is instantaneous. Perhaps I'm misunderstanding the whole crash condition though, so take with a grain of salt.

bstoeger:
PS: I suggest disabling the geolocation button while a geo lookup is running, so that the user can't start multiple lookups, which is just asking for trouble.

atdotde:
What @bstoeger says.

mwerle:
> In my tests, the setting and clearing of the `diveSite` variable was all done in the same thread. Therefore, I think the whole talk about mutexes, critical sections, etc. is a red herring.
> 
It's not the setting and cleraring of the variable that's the issue, it's the use of the variable while another thread is clearing it.

If there aren't multiple threads involved, how can the program modify a variable in another function while one function is executing? [1]

> I would suggest the following fix:
> 
This just shifts the check (and reduces the possibility  of the error significantly) but does not eliminate the possible race condition. It's a question of timing and will, in the future, be almost impossible to reproduce due to the precision of the timing required to trigger it. 

> 
> That said - I couldn't reproduce the problem, since for me reverse-geo lookup is instantaneous. Perhaps I'm misunderstanding the whole crash condition though, so take with a grain of salt.

Try setting a dive site on Lady Musgrave Island, Australia (23°54'16.344"S 152°24'28.721"E) - it currently fails to retrieve geo information for this site (it used to) and hence takes much longer than normal.

The crash actually occurs here:
```
void EditDiveSiteTaxonomy::redo()
{
	std::swap(value, ds->taxonomy); /** actually the crash occurs inside std::swap for as the location 0x38 (IIRC) is attempted to be read; the null-dereference itself weirdly enough doesn't trigger a SEGFAULT */
	emit diveListNotifier.diveSiteChanged(ds, LocationInformationModel::TAXONOMY); // Inform frontend of changed dive site.
}
```
Even if we add a null-check before the dereference:
```
void EditDiveSiteTaxonomy::redo()
{
	if(!ds) { return; }
	std::swap(value, ds->taxonomy); /** actually the crash occurs inside std::swap for as the location 0x38 (IIRC) is attempted to be read; the null-dereference itself weirdly enough doesn't trigger a SEGFAULT */
	emit diveListNotifier.diveSiteChanged(ds, LocationInformationModel::TAXONOMY); // Inform frontend of changed dive site.
}
```
, as I commented on the proposed fix, it just shifts the timing issue, it doesn't actually fix it.  We have no control over when control returns to this execution thread after 'ds' is set to null [1].

We either have to use a temporary local variable for the very short period of time this executes, or we have to use mutexes.

As for QFutures etc, I have no experience/knowledge of QT framework programming, so have no idea what they do and, I'm sorry, no time at present to learn so.


[1]
NB:  This analysis and my comments up to this point are based on the assumption that everything runs in a pre-emptive thread scheduling environment. If, on the other hand, the QT framework only ever passes control explicitly between threads, this doesn't apply.  In the latter case, the proposed fix is fine, as would simply moving the check until after the call to 'reverseGeoLookup'. 
As I said, I have zero knowledge of how QT works.

mwerle:
So on the assumption that control can only switch explicitly, I've tried the following and it seems to work (could no longer repro, but again, if it's pre-emptive threading, that is purely luck):

```
void LocationInformationWidget::reverseGeocode()
{
	bool buttonWasEnabled = ui.geoCodeButton->isEnabled();
	ui.geoCodeButton->setEnabled(false);
	dive_site *ds = diveSite; /* Save local copy; possibility of user closing the widget while reverseGeoLookup is running (see #2930) */
	location_t location = parseGpsText(ui.diveSiteCoordinates->text());
	taxonomy_data taxonomy = { 0, 0 };
	if (ds && has_location(&location)) {
	}
	if(ds == diveSite) {
		Command::editDiveSiteTaxonomy(ds, taxonomy);
	}
	ui.geoCodeButton->setEnabled(buttonWasEnabled);
}
```

mwerle:
I've pushed the updated fix; if ok, I can merge the individual commits (left them separate for clarity for now)

bstoeger:
> So on the assumption that control can only switch explicitly.

You can have different threads. However, in this case, the code in locationinformation.cpp is only executed using the UI thread, because the geo-lookup calls into the event loop. At least that's how I read the code. It's horrible and could certainly be made much clearer (apparently I wasn't aware that the geo lookup calls into the event loop). Good practice would be to return a future, not to call into the event loop! But let's not touch it as long as it works.

> could no longer repro, but again, if it's pre-emptive threading, that is purely luck

According to my testing, it's not. You can convince yourself by adding lines of the type:
```
#include <QThread>
        ...
        printf("Initializing diveSite variable: %p\n", QThread::currentThreadId());
```

Note that if you use normal `connect` statements, signals will cross thread boundaries. This makes signals trivially implemented IPC calls. It has some subtle implications though: references are treated differently whether you are sending a signal through thread-boundary or not. One of the many terrible design decisions of the Qt framework, which always follows the same pattern: try to make it superficially "easy" to first-timers, but in reality make it very subtle and hard to control.

> Try setting a dive site on Lady Musgrave Island, Australia (23°54'16.344"S 152°24'28.721"E) - it currently fails to retrieve geo information for this site (it used to) and hence takes much longer than normal.

Still no luck - returns immediately as for other locations with no geo lookup.

desktop-widgets/locationinformation.cpp
```diff
@@ -313,15 +313,18 @@ void LocationInformationWidget::on_diveSiteDistance_textChanged(const QString &s
 
 void LocationInformationWidget::reverseGeocode()
 {
-	dive_site *ds = diveSite; /* Use local copy in case the object-copy is NULL'd by another thread while we perform the geo lookup */
+	bool buttonWasEnabled = ui.geoCodeButton->isEnabled();
+	ui.geoCodeButton->setEnabled(false);
```
bstoeger: This will not work properly, as the `geoCodeButton` is set at different places.
What could happen:
- User starts long running geo lookup. Button is disabled.
- User switches to dive without dive site. Button is disabled.
- Geo lookup finishes. Button is enabled, even if it shouldn't be.

This needs reference counting - I can do it if you want.

mwerle:
> According to my testing, it's not. You can convince yourself by adding lines of the type:

It's fine, I believe you :) Just wanted to make it very clear we're on the same page here.

> Still no luck - returns immediately as for other locations with no geo lookup.

Maybe try putting a sleep(5) or whatever the equivalent is in place to make it take longer?

desktop-widgets/locationinformation.cpp
```diff
@@ -313,15 +313,18 @@ void LocationInformationWidget::on_diveSiteDistance_textChanged(const QString &s
...
-	if (!ds || !has_location(&location))
-		return;
 	taxonomy_data taxonomy = { 0, 0 };
-	reverseGeoLookup(location.lat, location.lon, &taxonomy);
-	if (ds != diveSite)
-		return;
-	Command::editDiveSiteTaxonomy(ds, taxonomy);
+	if (ds && has_location(&location)) {
+		reverseGeoLookup(location.lat, location.lon, &taxonomy);
+	}
```
bstoeger: Per coding guide, we don't use braces for single line code blocks ... *shrug*

desktop-widgets/locationinformation.cpp
```diff
@@ -313,15 +313,18 @@ void LocationInformationWidget::on_diveSiteDistance_textChanged(const QString &s
...
+	if(ds == diveSite) {
```
bstoeger: This looks like a new possible null-pointer access.

Suppose the function is entered with `diveSite == nullptr`.
- `ds` is set to null.
- no geo lookup is performed.
- but the condition 'ds == diveSite` is true and it will call the edit command with a nullptr, isn't it?

desktop-widgets/locationinformation.cpp
mwerle: That is (IMHO) a rather poor coding guideline.  Agreed, the braces do add an extra line of code, but poterntially prevent coding mistakes. I recently encountered a logic error precisely because of this code style in another project (admittedly the source file also had broken indentation and other issues, but still).

That said, I will, of course, adhere to the project guidelines and change it if I can't convince you to change the guidelines ;)

bstoeger:
> Maybe try putting a sleep(5) or whatever the equivalent is in place to make it take longer?

I thought about that, but from a cursory glance it appears that only Qt code is called in the separate thread...? However, I didn't try too hard, to be honest. Maybe I should just try to add sleep to random places and look what happens.

desktop-widgets/locationinformation.cpp
mwerle: Good catch, bug introduced due to refactoring for disabling the button.

My original refactor replaced the "return" with "goto done" but figured that probably wasn't an acceptable coding style in this project. (That said, I used that style in another pull request... it's a useful pattern if you want to force a single exit from a function, but don't want an ever-increasing level of indentation. )

bstoeger:
> That is (IMHO) a rather poor coding guideline.

I think that comes from kernel programming. From my side, I'm fine with both ways. You would have to check with @dirkhh on how strictly he wants to enforce this.

bstoeger:
> My original refactor replaced the "return" with "goto done" but figured that probably wasn't an acceptable coding style in this project.

Actually, we have a lot of those in C-code. Personally, I detest it because I think it is poor-man's RAII. The compiler should do the these trivial yet error-prone things. :-) As I said - I can do the button disabling thing if you wish.

desktop-widgets/locationinformation.cpp
mwerle: Ok, sure.  It's kindof out-of-spec for this bugfix/pull-request anyway, and UI code isn't my forte.

Sorry, didn't see this comment until just now (only saw the comments in the main thread) so ignore my latest commit.  I'll revert this pull request to just focus on the bugfix itself.

mwerle:
> As I said - I can do the button disabling thing if you wish.

Sorry, didn't see your comment suggesting that until just now.  I'll revert this pull request to just focus on the issue at hand.

mwerle:
..and squashed the commits.

bstoeger (approved):
Looks good to me. Will merge once CI passes.

dirkhh:
> > That is (IMHO) a rather poor coding guideline.
> 
> I think that comes from kernel programming. From my side, I'm fine with both ways. You would have to check with @dirkhh on how strictly he wants to enforce this.

Coding style is subjective. Linus started this project, so it started with his preferred coding style. I took over, and I also have 20+ years of being used to that coding style.
We have many coding style violations in the code - but I generally try to encourage people to follow what we have written down.

The fundamental challenge that we have here is this.
- the code was started by C programmers
- most of the code is now C++ and written by C++ programmers
- the two worlds typically don't meet happily

I have no interest and no energy to rewrite all of this to follow some 'new' style. I strongly believe that having one consistent style helps massively with readability. Therefore I generally push towards sticking with what we have.

Rarely, very rarely, over many eons and many frustrating discussions, it happens that I can be talked into changing my mind. Lambdas for very simple connection targets is one of those areas where I finally changed my mind. But realize that you are pushing an elephant up a steep hill here.

PR #2932, opened 2020-09-02
Planner: Properly initialize salinity

atdotde:
When the dive has no explicity salinity, our conversion
between pressure and depth assumed salt water. Make this
explicity by using the corresponding macro.

When the planner starts and no salinity is set explicity,
set the water type chooser to salt water to reflect
our default assumption.

This was pointed out on scuba board.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+planner: properly initialize salinity
 desktop: add an "Edit Gas Change" right-click option [#2910]
 core: add support for Shearwater Peregrine (requires firmware V79 or newer)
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
looks reasonable

dirkhh:
it might be nice as future work to add a little macro to deal with the gravity conversion :-) 
(even though that's only used three or four times across the code from what I can see)

PR #2933, opened 2020-09-02
Macro for "gravity conversion"

atdotde:
This adds a common macro to convert salinity (which is
given as a density in terms of g per 10l) to a specific
weight with units of mbar / mm = bar / m that is used
to translate between pressures and depths.

The weired factor of 10 (from the unusual unit of salinity)
is included in the macro. It is there for historical reasons,
as it goes back to 05b55542c8 from 2012 where it was introduced
in code for downloading from Uemis dive computers.

Now, salinity appears in too many places to easily remove
this unconventional factor of 10 everywhere without breaking
to many things (including various dive computer downloads).

As suggested by @dirkhh in #2932

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -38,6 +38,8 @@ const char *divemode_text_ui[] = {
 // For writing/reading files.
 const char *divemode_text[] = {"OC", "CCR", "PSCR", "Freedive"};
 
+static int calculate_depth_to_mbar(int depth, pressure_t surface_pressure, int salinity);
+
 /*
  * Adding a cylinder pressure sample field is not quite as trivial as it
  * perhaps should be.
@@ -3688,10 +3690,20 @@ int get_surface_pressure_in_mbar(const struct dive *dive, bool non_null)
 	return mbar;
 }
 
+/* This returns the conversion factor that you need to multiply
+ * a (relative) depth in mm to obtain a (relative) pressure in mbar.
+ * As everywhere in Subsurface, the expected unit of a salinity is
+ * g/10l such that sea water has a salinity of 10300
+ */
+static double salinity_to_specific_weight(int salinity)
+{
+	return salinity * 0.981 / 100000.0;
+}
+
 /* Pa = N/m^2 - so we determine the weight (in N) of the mass of 10m
  * of water (and use standard salt water at 1.03kg per liter if we don't know salinity)
  * and add that to the surface pressure (or to 1013 if that's unknown) */
-int calculate_depth_to_mbar(int depth, pressure_t surface_pressure, int salinity)
+static int calculate_depth_to_mbar(int depth, pressure_t surface_pressure, int salinity)
 {
 	double specific_weight;
 	int mbar = surface_pressure.mbar;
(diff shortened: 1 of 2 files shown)


dirkhh:
I don't think this was ever about the mass of 10 liters of water. 
This all started with the pressure to depth conversation where 1bar is the equivalent of 10m, not 1m.
And then laziness (likely on my part) had us end up with the salinity factors normalized to 10000g instead of 1000g
I agree that this might be very painful to clean up, but what I wanted to point out here is that no where did I think that it made sense to look at the mass of 10l of water :-)

core/dive.c
```diff
@@ -3688,6 +3688,8 @@ int get_surface_pressure_in_mbar(const struct dive *dive, bool non_null)
 	return mbar;
 }
 
+#define SALINITY_TO_SPECIFIC_WEIGHT(s) ((s) * 0.981 / 100000.0)
```
dirkhh: ok, now I'm confused. Why are you dividing by 100 thousand?
What are the expected units in and out for this macro?

core/dive.c
```diff
@@ -3728,13 +3730,11 @@ double depth_to_atm(int depth, const struct dive *dive)
...
-	int cm;
-	double specific_weight = SEAWATER_SALINITY * 0.981 / 10000.0;
+	double specific_weight = SALINITY_TO_SPECIFIC_WEIGHT(SEAWATER_SALINITY);
 	if (dive->dc.salinity)
-		specific_weight = dive->dc.salinity / 10000.0 * 0.981;
+		specific_weight = SALINITY_TO_SPECIFIC_WEIGHT(dive->dc.salinity);
 	/* whole mbar gives us cm precision */
-	cm = (int)lrint(mbar / specific_weight);
-	return cm * 10;
+	return (int)lrint(mbar / specific_weight);
```
dirkhh, line 3747: I find this calculation really confusing.
```
1bar ~= 10m
1mbar ~= 10mm == 1cm
```
so that made sense. That extra factor of ten that you have in there... I must be missing something?

core/dive.c
atdotde: This takes a salinity in the kg/10l unit that we use for salinity and returns a number in mbar/mm that can directly be used to convert between our beloved units of mbar and mm. To me this looks like the only sane way (besides changing the unit of salinity by a factor of 10 everywhere).

dirkhh: err, but mbar are not mm. See below. mbar are cm 
so you are now doing deci-mili bar?

dirkhh: so we had mbar and cm and that mostly made sense to me. but 1/10mbar and mm doesn't, really.

atdotde: No no no. The task at hand is: Convert a depth (which is usually expressed mm) to a pressure (usually expresses in mm). 
For this, we need a salinity. This macro takes a salinity (in our weird units where sea water has a salinity of 10300) and spits out the conversion factor that does the actual job above without further ad hoc factors.

so, if you like it is 0.981 / 100000 (mbar / mm) / (g / 10l) as our units are mbar for pressure, mbar for depth and g/10l for salinity. 

I think this is natural given our choice of units.

bstoeger:
Make this a `static` function, please. This documents the expected input type, makes it more readable, and the compiler will inline it anyway.

Moreover, I suggest making `calculate_depth_to_mbar` static (i.e. local to the source file) as well, since it doesn't seem to be used outside of `dive.c`?

core/dive.c
bstoeger: To avoid confusion, I would suggest to put the units of measure of the input and output values in a comment...?

dirkhh:
@atdotde sorry I caused you to enter this morass. Entirely my fault.
I agree with @bstoeger 's suggestions around static functions, units, types, etc.
But I'm perfectly fine if you say "I never wanted to be on the hook for this" and walk away :-)

atdotde:
I addressed @bstoeger 's comments.

bstoeger (approved):

PR #2935, opened 2020-09-03
[Bug #2934] Geo Lookup - support for remote dive sites

### LINKED ISSUE #2934: Remote divesites don't have reverse geo lookup information
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
For an example of a remote dive site see Lady Musgrave Island, Australia (23°54'16.344"S 152°24'28.721"E)

The current implementation uses the geonames "findNearbyPlaces" API, which only looks for populated places and is limiting itself to within 50km. In this case there aren't any.

Perhaps using "findNearby" instead, or using "findNearby" if "findNearbyPlaces" came up empty might be a useful fallback.

http://api.geonames.org/findNearbyJSON?lat=-23.9&lng=152.4&username=demo
http://api.geonames.org/findNearbyPlaceNameJSON?lat=-23.9&lng=152.4&username=demo

### Operating system:
N/A; all versions

### Subsurface version:

4.9.6 (both self-compiled as well as official release) 

Does the same happen on another Subsurface version?
My save-file contains a reverse-geocode lookup of the country for this dive  site, so at some point (a long time ago) this site used to return some information. There was some rework and simplification about 3 years ago.

### Steps to reproduce:
1) Create a new dive site in a remote location
2) Attempt a reverse-geocode lookup
3) ...
4) fail

### Current behavior:
No geo code information retrieved.

### Expected behavior:
At least country/state information made available.

mwerle:
Some remote dive sites have no populated places (towns, cities)
nearby. For such sites, we now fall back to looking up
unpopulated place names, such as the reef or island name.

Also refactor the code; the actual network access is now
encapsulated in its own function making the reverseGeoLookup
function much more readable.

- [ ] Bug fix
- [X] Functional change
- [ ] New feature
- [X] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

If 'geonames.findNearbyPlaceNameJSON' returns no data, a call to 'geonames.findNearbyJSON' is now made, which will (always?) return some data for the lookup.

### Changes made:
1) Moved the already duplicated network code to its own function 'doAsyncRest'.
2) Refactored 'reverseGeoLookup' to use 'doAsyncRest'
3) Added a lookup to http://api.geonames.org/findNearbyJSON if http://api.geonames.org/findNearbyPlaceNameJSON returns no data

### Related issues:
Fixes #2934 

### Additional information:

Use Lady Musgrave Island, Australia (23°54'16.344"S 152°24'28.721"E) as an example of a remote dive site.

### DIFF
--- a/core/divesitehelpers.cpp
+++ b/core/divesitehelpers.cpp
@@ -22,7 +22,8 @@
 #include <QTimer>
 #include <memory>
 
-void reverseGeoLookup(degrees_t latitude, degrees_t longitude, taxonomy_data *taxonomy)
+/** Performs a REST get request to a service returning a JSON object. */
+static QJsonObject doAsyncRESTGetRequest(const QString& url, int msTimeout)
 {
 	// By making the QNetworkAccessManager static and local to this function,
 	// only one manager exists for all geo-lookups and it is only initialized
@@ -30,16 +31,13 @@ void reverseGeoLookup(degrees_t latitude, degrees_t longitude, taxonomy_data *ta
 	static QNetworkAccessManager rgl;
 	QNetworkRequest request;
 	QEventLoop loop;
-	QString geonamesURL("http://api.geonames.org/findNearbyPlaceNameJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
-	QString geonamesOceanURL("http://api.geonames.org/oceanJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
 	QTimer timer;
 
 	request.setRawHeader("Accept", "text/json");
 	request.setRawHeader("User-Agent", getUserAgent().toUtf8());
 	QObject::connect(&timer, SIGNAL(timeout()), &loop, SLOT(quit()));
 
-	// first check the findNearbyPlaces API from geonames - that should give us country, state, city
-	request.setUrl(geonamesURL.arg(getUiLanguage().section(QRegExp("[-_ ]"), 0, 0)).arg(latitude.udeg / 1000000.0).arg(longitude.udeg / 1000000.0));
+	request.setUrl(url);
 
 	// By using a std::unique_ptr<>, we can exit from the function at any point
 	// and the reply will be freed. Likewise, when overwriting the pointer with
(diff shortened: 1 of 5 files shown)


core/divesitehelpers.cpp
```diff
@@ -22,24 +22,20 @@
 #include <QTimer>
 #include <memory>
 
-void reverseGeoLookup(degrees_t latitude, degrees_t longitude, taxonomy_data *taxonomy)
+/** Performs a REST request to a service returning a JSON object. */
+QJsonObject doAsyncRest(QString url, int msTimeout)
```
bstoeger: 1) Is this function used outside of this source file? If not, make it `static`, please.

2) Not that it matters, since `QString` has COW & refcounting semantics. However, I would still prefer `const QString &` unless you really need a copy here.

core/divesitehelpers.cpp
```diff
@@ -22,24 +22,20 @@
...
 {
-	// By making the QNetworkAccessManager static and local to this function,
-	// only one manager exists for all geo-lookups and it is only initialized
-	// on first call to this function.
```
bstoeger, line 29: Did this comment become meaningless? If not, please keep it.

core/divesitehelpers.cpp
```diff
@@ -48,119 +44,116 @@ void reverseGeoLookup(degrees_t latitude, degrees_t longitude, taxonomy_data *ta
...
+	timer.start(msTimeout);
 	loop.exec();
 
 	if (timer.isActive()) {
 		timer.stop();
 		if (reply->error() > 0) {
-			report_error("got error accessing geonames.org: %s", qPrintable(reply->errorString()));
-			return;
+			report_error("got error accessing %s: %s", qPrintable(url), qPrintable(reply->errorString()));
+			goto done;
```
bstoeger: This makes code less readable, not more. Early return is idiomatic C++, please keep it.

core/divesitehelpers.cpp
bstoeger: REST appears to be an acronym. Therefore, I would suggest `doAsyncREST`.

bstoeger (changes requested):
Only a few minor comments and one major gripe: Please don't do the `goto done` thing in C++ code.

Actually, just remove the `rv` variable.

When you want to exit without result simply do a
```
        return QJsonObject();
```
or even
```
        return {};
```
if Qt accepts that.

In the single(!) case where you return a result you can simply do:
```
        return jsonDoc.object();
```

This code here is not performance critical, so it doesn't matter. Just note that it is idiomatic, because it avoids an unnecessary copy and initialization operation.

The way you write it:
1) It initializes `rv`
2) `jsonDoc.object()` initializes a temporary variable.
3) The temporary variable is copied over `rv`

When you simply do as I suggested above:
`jsonDoc.object()` initializes the memory provided by the caller.

So in a performance critical context (which this clearly is not) it is much more efficient and the code is (once you're used to it) more readable. Let's stick to the idiomatic way of doing things.

core/divesitehelpers.cpp
mwerle: Sorry, did not mean to drop this comment. Copy-pasta failure.

mwerle:
> So in a performance critical context (which this clearly is not) it is much more efficient and the code is (once you're used to it) more readable. Let's stick to the idiomatic way of doing things.

Based on some brief reading, my impression was that a modern compiler would optimise what I did anyway and result in a zero-memcpy codepath returning the final object back to the caller.

"once you're used to it"
-> there's a very good reason why I primarily stick to C in my dayjob  ;)
This is nowhere near readable (to me).  I like a function to have a single entry-point, and a single exit-point, and be able to deal with stuff explicitly rather than implicitly. Returning part-way through a function can lead to all sorts of resource leaks (depending on what a function does, of course). Of course, auto_ptr and things like that are "magic" to me, which I don't trust ;)

I'm sure for C++ programmers, the C idioms are probably just as unreadable.. :)

----
Anyway, requested changes about to be committed. Sorry for the unnecessary waffle - not trying to argue!

mwerle:
Reading through the code again, and taking the "return early" idiom to heart, each error-condition should return asap.

So basically the
```
if (timer.isActive()) {
   ... main body
} else {
  ...error
}
```
_should_ be:
```
if(!timer.isActive()) {
  ... error stuff;
  return;
}
main body
```
That would leave the final, successful, return with the actual data at the bottom of the function, instead of hidden somewhere half-way through.

Worth refactoring, or leave as-is to minimise the code changes?

core/divesitehelpers.cpp
```diff
@@ -48,119 +46,115 @@ void reverseGeoLookup(degrees_t latitude, degrees_t longitude, taxonomy_data *ta
...
-			qDebug() << "no reverse geo lookup; geonames returned\n" << fullReply;
+		}
+		taxonomy->nr = ri;
+		l3 = taxonomy_index_for_category(taxonomy, TC_ADMIN_L3);
+		lt = taxonomy_index_for_category(taxonomy, TC_LOCALNAME);
+		if (l3 == -1 && lt != -1) {
+			// basically this means we did get a local name (what we call town), but just like most places
+			// we didn't get an adminName_3 - which in some regions is the actual city that town belongs to,
+			// then we copy the town into the city
+			taxonomy->category[ri].value = copy_string(taxonomy->category[lt].value);
```
bstoeger: In the code above, the old string is freed but not here. Is this on purpose?

core/divesitehelpers.cpp
```diff
@@ -48,119 +46,115 @@ void reverseGeoLookup(degrees_t latitude, degrees_t longitude, taxonomy_data *ta
...
+		report_error("error parsing JSON response: %s", qPrintable(errorObject.errorString()));
+		return QJsonObject{};
+	}
+	// Success, return JSON response from server
+	return jsonDoc.object();
+}
+
+void reverseGeoLookup(degrees_t latitude, degrees_t longitude, taxonomy_data *taxonomy)
+{
+	const QString geonamesNearbyURL("http://api.geonames.org/findNearbyJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
```
bstoeger: Bikeshedding: Since you're refactoring this source file, you could turn those strings into `QStringLiteral`s. I'm fine with leaving them as is.

dirkhh:
This seems like a good set of changes. @bstoeger are all your concerns addressed (obviously we'll want to squash that into one commit)?

mwerle:
(So it appears I don't know how to use GitHub for reviewing and hadn't submitted my comments...)

As further information, these days reverseGeoLookup() is only called from a single location, which always passes in an empty taxonomy_data object.  The results are then copied over the top of any existing taxonomy_data for the dive site.

So any and all comments about merging of data etc. can be ignored.  I've also simplified the whole function significantly since all the special handling appears to have been leftovers from legacy implementations which no longer apply, although I've still coded it in such a way that it takes into account a caller passing in a non-empty taxonomy_data object.

Please see latest commit.

core/divesitehelpers.cpp
bstoeger: > Overall might be better to entirely discard the old taxonomy and start from a new one. Or to actually clear out the data where it says it does. And move the OCEAN to the front to stop it being such a special case.

Or, since this is called from C++ only, simply return a `taxonomy_data` structure instead of passing it in. Thus it becomes clear who is responsible for object construction.

BTW: It just occurs to me that your last PR introduced a (very minor) resource leak:

```
	taxonomy_data taxonomy = { 0, 0 };
	if (ds != diveSite)
		return; // <--- taxonomy data is now leaked!
```

C-style memory management is so pointlessly painful. :-/

What I did in the undo commands to make this less painful is to introduce "owning pointers", which are `unique_ptr`s with a deleter function. You could then declare the function as
```
OwningTaxonomyPtr reverseGeoLookup(degrees_t latitude, degrees_t longitude);
```
and use it like thus:
```
    OwningTaxonomyPtr taxonomy = reverseGeoLookup(...);
    if (!taxonomy || ds != diveSite)
        return;
    Command::editDiveSiteTaxonomy(ds, *taxonomy);
```

The trick is: when the pointer goes out of scope, the taxonomy is automatically freed! Nothing to worry about!

Note: I explicitly do NOT suggest that you implement such a scheme, it was just a FYI.

mwerle: > Or, since this is called from C++ only, simply return a taxonomy_data structure instead of passing it in. 

That doesn't (historically) appear to have been the case, hence the rather convoluted code in the original function.  These days it's only called from a single place and yes, it would reduce the scope of the function and make it very clear exactly what it does (returns the reverse geocode lookup results and doesn't try anything "clever" by merging data etc)

> BTW: It just occurs to me that your last PR introduced a (very minor) resource leak:

Indeed..  *sigh*  Although that was someone's suggested approach, not my original solution (I didn't know that reverseGeoLookup() allocated memory at that point).

>What I did in the undo commands to make this less painful is to introduce "owning pointers", which are unique_ptrs with a deleter function. You could then declare the function as

Ok, how does that work? Given that taxonomy_data.category is a malloc'd pointer.  I'm guessing the "OwningTaxonomyPtr" is a wrapper-class which takes care of it?  Can you give an example of where you implemented one of these?

mwerle: Ok, so I found the 'OwningPtr' implementations, but the only way to implement it as you suggest is if reverseGeoLookup() allocates not only 'taxonomy->category', but the 'taxonomy_data' structure as well. That makes for even messier handling inside the function.

I think I'll have to leave it as per my latest commit (pending review) unless we want to completely revamp everything and turn things into proper C++ classes. Or change "taxonomy_data.category" into being an array inside the struct instead of a pointer, and then just allocate "taxonomy_data".

Which, I think, is getting well beyond the scope for this (originally relatively simple) pull request.  I'm sorry, but my time is quite limited otherwise I'd love to help with more comprehensive work.

bstoeger: As I said - I'm not expecting you to implement such a scheme and I am fine with the `free_taxonomy` call. I will go through it once more today, but it should be fine now.

bstoeger (approved):
If I'm reading things correctly, these two blocks:
```
	if (!taxonomy.category)
		return taxonomy;
```
and
```
		if (taxonomy.category == nullptr)
			taxonomy.category = alloc_taxonomy();
```
became no-ops.

I'm fine with this PR if these are removed, or it is explained why they are still necessary.

mwerle:
> If I'm reading things correctly, these two blocks:
> ...
> became no-ops.
> 
> I'm fine with this PR if these are removed, or it is explained why they are still necessary.

The former is the standard check for out-of-memory (malloc returning NULL).  I know it's incredibly rare on most modern systems, but that's still the spec for malloc and since my dayjob mostly involves coding on systems where out-of-memory is a distinct possibility, I always check for it.

man malloc (Debian GNU/Linux 10):
"... On error, these functions return NULL.  NULL may also be returned by a successful  call to malloc() with a size of zero ..."

Conversely if you can convince me why it's a no-op on all of our supported systems (ie, malloc() is guaranteed not to return NULL), I'm happy to remove it.

The second is an oversight; thank you.  It will be removed. Thank you.

bstoeger:
> Conversely if you can convince me why it's a no-op on all of our supported systems (ie, malloc() is guaranteed not to return NULL), I'm happy to remove it.

That's a nice thought, but it is also completely pointless - most of the allocations are done in C++ code (indirectly) via "new", which will throw on allocation failure. However, owing to major shortcomings of the Qt framework and our own code base, we are not exception safe, so the application will crash inevitably (e.g. with the next string operation). You will not make this application even close to resilient towards allocation failure with any reasonable amount of effort.

On allocation failure, I would suggest just making it crash right away with a null-pointer access. At least that should free up some memory.

mwerle:
> That's a nice thought, but it is also completely pointless

*shrug* You're the boss :) 

Will remove both checks and squash, ready for final review.

bstoeger:
> You're the boss :)

No, absolutely not! Final say goes to @dirkhh.

mwerle:
> No, absolutely not! Final say goes to @dirkhh.

Well, going by precedent, I don't see anywhere else which checks for malloc() failures. So... committed final push adhering to project standards.

bstoeger:
> Well, going by precedent, I don't see anywhere else which checks for malloc() failures.

There are a few here and there in very old parts of the code. See for example `alloc_dive()`. Note however that it calls `exit(1)` on failure, so it might just crash with a null-pointer access.

> So... committed final push adhering to project standards.

Thanks.

bstoeger:
There are still a few minor coding-style issues (braces for one-line block; reference `&` associated with base type not variable name). That said, readability-wise this PR is a huge improvement and therefore I'll merge.

@mwerle: I suggest you read the coding-style document. ;)

PR #2940, opened 2020-09-05
cleanup: leak fix in taxonomy_set_country() [+more assorted cleanups]

bstoeger:
When overwriting a country, the old string was not freed. Fix this.
Contains an unrelated coding-style fix: use braces if code block
contains more than one line.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes what looks like a small resource leak in C code. Only compile-and-run tested. Please have a check whether my interpretation is correct.

### Changes made:

1) Free string before overwriting it.
2) Add braces as per coding style.

### Related issues:

Found while trying to understand #2935.

### Mentions:

@mwerle @dirkhh

### DIFF
--- a/commands/command_divesite.cpp
+++ b/commands/command_divesite.cpp
@@ -268,7 +268,7 @@ bool EditDiveSiteCountry::workToBeDone()
 void EditDiveSiteCountry::redo()
 {
 	QString old = taxonomy_get_country(&ds->taxonomy);
-	taxonomy_set_country(&ds->taxonomy, copy_qstring(value), taxonomy_origin::GEOMANUAL);
+	taxonomy_set_country(&ds->taxonomy, qPrintable(value), taxonomy_origin::GEOMANUAL);
 	value = old;
 	emit diveListNotifier.diveSiteChanged(ds, LocationInformationModel::TAXONOMY); // Inform frontend of changed dive site.
 }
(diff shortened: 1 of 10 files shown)


mwerle:
free_taxonomy also has a potential leak (although with my recent changes less likely).

Still it would be safer to iterate up to TC_NR_CATEGORIES instead of just taxonomy->nr.

(I found and fixed it in #2935 but it must've gotten lost in one of the many rewrites)

mwerle:
Secondly, if you're going into that function already, IMHO you should really remove the search loop and replace it with a call to taxonomy_index_for_category(), then perform the free() if index != -1. Or just call free() regardless.

bstoeger:
Added a number of cleanups, which are supposed to make the whole taxonomy interface more palatable. The users of the taxonomy system *really* shouldn't have to care about the memory management!

dirkhh (approved):
This all looks useful.
Is there any functionality change / bug fix that is worth noting in the CHANGELOG?
I don't think so, but I wanted to call it out in case I missed something

bstoeger:
> Is there any functionality change / bug fix that is worth noting in the CHANGELOG?
> I don't think so, but I wanted to call it out in case I missed something

Nothing user-visible should have changed (knock wood).

PR #2941, opened 2020-09-06
[TRIVIAL?] cleanup: when planning a dive, set dive computer to first dc

bstoeger:
When planning a dive, dc_number was set to 1, which actually is
the second dc! The code seems to handle this gracefully, however
it appears weird. Let's set dc_number to 0 instead.

Originally, the assignment was introduced in a422957cd6525b9753
and moved later in 4f5621c4c6acc3a.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a cosmetic change of a code-oddity. This shouldn't be user visible.

### Changes made:

1) On planning set displayed DC to first, not second (which probably doesn't exist).

### Mentions:

@atdotde: Perhaps I'm missing something and this was intended...?

### DIFF
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -905,7 +905,7 @@ MoveDiveComputerToFront::MoveDiveComputerToFront(dive *d, int dc_num)
 }
 
 DeleteDiveComputer::DeleteDiveComputer(dive *d, int dc_num)
-	: DiveComputerBase(d, clone_delete_divecomputer(d, dc_num), std::min(count_divecomputers(d) - 1, dc_num))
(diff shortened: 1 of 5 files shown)


atdotde:
Looks correct.

atdotde (approved):
Where is this value used later? What you write sounds correct.

bstoeger:
> Where is this value used later?

It is passed to `get_dive_dc()`, which does the arguably correct thing and returns the first dc for out of range arguments. In principle, I believe that the line could be removed, but it also doesn't hurt.

bstoeger:
Added a few more cleanups, which are a bit more hairy, but also seem to be rather obvious...?

dirkhh (approved):
yes, this seems right.

PR #2946, opened 2020-09-12
desktop: on weight type change, don't overwrite weight if already set

### LINKED ISSUE #2938: Changing weight type resets weight amount
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
When changing the weight type for a dive the weight amount field is reset to either 0 or to last used weight amount of the given weight type. 

### Operating system:
Windows 7 English

### Subsurface version:
Subsurface 4.9.6 official

### Steps to reproduce:
Go to a dive log entry with a unknown weight type and a specified weight amount set
Change the weight type to any other type. 
the specified amount is changed to 0 for unused weight types, or to the previously used amount for the specified weight type.

### Current behavior:
weight amount is set along with weight type

### Expected behavior:
weight type and amount should be independently editable

bstoeger:
When importing from other software, it happens that weights are imported
without their type. When the user changes the type, the imported weight
is overwritten, which is not exactly a friendly behavior.

On the other hand, when changing the type after creation of a weight
entry, it is preferrable to set a default weight. This is convenient
for people who commonly use the same weight.

As a compromise, set the default weight only if it was unset. We
recognize this by a weight value of 0 g.

Fixes #2938

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This implements the compromise suggested in #2938. Code is in principle trivial, but needs some real world testing.
### Changes made:

1) When weight is already set, don't fill with default value when changing weight type

### Related issues:

Fixes #2938.

### Release note:

Done.
### Documentation change:

No.
### Mentions:

@sveinmagnus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: don't overwrite already set weights when changing weight type [#2938]
 desktop: do substring search for equipment types
 planner: properly initialize salinity
 desktop: add an "Edit Gas Change" right-click option [#2910]
--- a/core/equipment.c
+++ b/core/equipment.c
@@ -127,7 +127,7 @@ void add_weightsystem_description(const weightsystem_t *weightsystem)
 
 weightsystem_t clone_weightsystem(weightsystem_t ws)
 {
-	weightsystem_t res = { ws.weight, copy_string(ws.description) };
+	weightsystem_t res = { ws.weight, copy_string(ws.description), ws.auto_filled };
 	return res;
 }
 
--- a/core/equipment.h
+++ b/core/equipment.h
@@ -50,9 +50,10 @@ typedef struct
 {
 	weight_t weight;
 	const char *description; /* "integrated", "belt", "ankle" */
+	bool auto_filled; /* weight was automatically derived from the type */
 } weightsystem_t;
 
-static const weightsystem_t empty_weightsystem = { { 0 }, 0 };
+static const weightsystem_t empty_weightsystem = { { 0 }, 0, false };
 
 /* Table of weightsystems. Attention: this stores weightsystems,
  * *not* pointers * to weightsystems. This has two crucial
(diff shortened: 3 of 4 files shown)


dirkhh:
When the heck did adding a new weight / tank on macOS break? I tried to test this on my Mac and stumbled across this unpleasant surprise...
Switching to my Linux box in order to test this PR I see a very distinct change in behavior:
Download a fresh dive (or manually create it)
Go to Equipment tab, add a weight system
As I curser up / down through the 'Type', the weight changes.
Tab to 'Weight', tab back to 'Type'.
As I curser up / down through different types, the weight no longer changes.

Is this significant? I'm not sure. It's definitely a behavioral change that I'm not thrilled about.
I can see that the behavior in #2938 is likely more common to cause frustration, but I wonder if there's a way we could keep the old behavior in this case without too much 'magic'

bstoeger:
> When the heck did adding a new weight / tank on macOS break?

I don't know - I haven't really used MacOS in a decade, since it feels like a straitjacket to me.

>  I wonder if there's a way we could keep the old behavior in this case without too much 'magic'

Introduce an `auto_filled` flag to the weight system, which is not saved to the log?

dirkhh:
It's a very pretty straight jacket. I use it all the time and I enjoy it quite a bit.
So this works with latest master and Qt 5.14 on my Mac, and fails with Qt 5.15.
I'll try to dig more into this 🤷🏼‍♂️

dirkhh:
> > I wonder if there's a way we could keep the old behavior in this case without too much 'magic'
> 
> Introduce an `auto_filled` flag to the weight system, which is not saved to the log?

I don't quite see how you envision this, can you say a bit more, please?

bstoeger:
> It's a very pretty straight jacket. I use it all the time and I enjoy it quite a bit.

Good for you.

> I don't quite see how you envision this, can you say a bit more, please?

Something as the newly pushed commit? Alternatively we could use the heuristic "if weight equals default-value, then it probably was auto-filled".

dirkhh:
I think this looks reasonable.
Let me build it and test it, but I guess this would solve the issue I mentioned

dirkhh (approved):
ok, this leaves the old functionality for people adding new dives, so I'm ok with it.

PR #2947, opened 2020-09-13
desktop: more Qt 5.15 fixes

dirkhh:
This now gets me a working Qt 5.15 based build - even on BigSur

- [x] Bug fix
- [x] Build system change

I don't love the way the TableView change looks, but at least the widget is functional again.
I'm not sure what a better fix would do, though.

The other commits help me create a self contained build that runs on 10.15/11.0 (but not on older versions).

### DIFF
--- a/desktop-widgets/tableview.cpp
+++ b/desktop-widgets/tableview.cpp
@@ -60,7 +60,10 @@ TableView::TableView(QWidget *parent) : QGroupBox(parent)
 		iconSize = btnSize - 2*min_gap;
 	}
 	plusBtn->setIconSize(QSize(iconSize, iconSize));
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
+	// with Qt 5.15, this leads to an inoperable button
 	plusBtn->resize(btnSize, btnSize);
+#endif
 	connect(plusBtn, SIGNAL(clicked(bool)), this, SIGNAL(addButtonClicked()));
 }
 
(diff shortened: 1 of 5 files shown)


desktop-widgets/tableview.cpp
```diff
@@ -60,7 +60,10 @@ TableView::TableView(QWidget *parent) : QGroupBox(parent)
 		iconSize = btnSize - 2*min_gap;
 	}
 	plusBtn->setIconSize(QSize(iconSize, iconSize));
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
```
bstoeger, line 63: And if you remove it unconditionally?

dirkhh: can you try it on your Linux system without it. For me it looks better on Linux with this line... and I would love to figure out how to make it look ok **and** work. But of course you are right, we could just remove it, period.

bstoeger: Yes - definitely looks super ugly without this line. That's a bit infuriating. Perhaps ask an Qt expert?

bstoeger: According to a quick search, you should use a `QToolButton` instead of a `QPushButton`. However, that made it look even worse. Giving up - sorry.

dirkhh: yeah - I got that far as well.
I figured for 4.9.7 I'd go with the compromise in this PR, and then I'll think about a better solution afterwards.
Or maybe I find time this week to poke at it some more. But I think this one is "good enough". Any other comments, or good to merge?

bstoeger: The build system stuff is Chinese to me. So fine with me.

PR #2949, opened 2020-09-13
Disable SAC factor setting for CCR dives

atdotde:
The SAC factor is only used for minimal gas calculations which
don't make sense in the CCR context.

Additionally, make bailout stop for at least minimum switch
time or problem solving time.

This is a small addition that had been sitting on my hard drive for a while but which could go into 4.9.7

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/planner.c
+++ b/core/planner.c
@@ -825,11 +825,12 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 	if ((divemode == CCR || divemode == PSCR) && prefs.dobailout) {
 		divemode = OC;
 		po2 = 0;
+		int bailoutsegment = MAX(prefs.min_switch_duration, 60 * prefs.problemsolvingtime);
 		add_segment(ds, depth_to_bar(depth, dive),
 			get_cylinder(dive, current_cylinder)->gasmix,
-			prefs.min_switch_duration, po2, divemode, prefs.bottomsac);
-		plan_add_segment(diveplan, prefs.min_switch_duration, depth, current_cylinder, po2, false, divemode);
-		clock += prefs.min_switch_duration;
+			bailoutsegment, po2, divemode, prefs.bottomsac);
+		plan_add_segment(diveplan, bailoutsegment, depth, current_cylinder, po2, false, divemode);
+		clock += bailoutsegment;
 		last_segment_min_switch = true;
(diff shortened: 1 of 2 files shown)


desktop-widgets/diveplanner.cpp
```diff
@@ -606,6 +606,8 @@ void PlannerSettingsWidget::setBackgasBreaks(bool dobreaks)
 void PlannerSettingsWidget::setBailoutVisibility(int mode)
 {
 		ui.bailout->setDisabled(!(mode == CCR || mode == PSCR));
+		ui.sacFactor->setDisabled(mode == CCR);
+		ui.sacfactor->setDisabled(mode == CCR);
```
bstoeger: Because once is not enough? ;)

core/planner.c
```diff
@@ -825,11 +825,12 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 	if ((divemode == CCR || divemode == PSCR) && prefs.dobailout) {
 		divemode = OC;
 		po2 = 0;
+		int bailoutsegment = prefs.min_switch_duration < prefs.problemsolvingtime * 60 ? prefs.problemsolvingtime * 60 : prefs.min_switch_duration;
```
bstoeger: What about
```
		int bailoutsegment = MAX(prefs.problemsolvingtime * 60, prefs.min_switch_duration);
```
to make this understandable without mental gymnastics?

[Completely unrelated: the things you have to do to define a proper MAX macro ... jeez. And that's probably not even standard C.]

desktop-widgets/diveplanner.cpp
dirkhh: That way the silly thing is REALLY disabled 🤣

dirkhh:
Happy to merge this before 4.9.7 - maybe after you look at @bstoeger 's comments :-)

desktop-widgets/diveplanner.cpp
atdotde: I think you could be more trusting with respect to the optimisations modern compilers can do.

dirkhh (approved):

PR #2951, opened 2020-09-13
cleanup: some reshuffling of the divecomputer code

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is just some cleanup / code reshuffling of the divecomputer code. It is the beginning of a bigger refactoring. The only somewhat contentious part should be the last commit - which solves the device.h/divecomputer.h mess by merging them. That's why I'm posting this - if you want split of the headers / source files, let me know where they should be split and how they should be called. Thanks.

Note: ultimately, the DiveComputerList class should disappear from the header and be an implementation detail in the source file.

### Changes made:

1) Remove unused code
2) Merge device.h and divecomputer.h

### Mentions:

@dirkhh


dirkhh (approved):
not sure this really needs to go in before 4.9.7... on the other hand... meh.
builds and tests pass...

PR #2958, opened 2020-09-16
McLean UUID matching trial

torvalds:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Let's see if this causes us to use the right characteristics in the vendor service for the McLean Extreme BLE code..

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -101,11 +101,11 @@ void BLEObject::writeCompleted(const QLowEnergyDescriptor&, const QByteArray&)
 	desc_written++;
 }
 
-struct uud_match {
+struct uuid_match {
 	const char *uuid, *details;
 };
 
-static const char *match_service(const QBluetoothUuid &service, const struct uud_match *array)
+static const char *match_service(const QBluetoothUuid &service, const struct uuid_match *array)
 {
 	const char *uuid;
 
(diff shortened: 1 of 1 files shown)


core/qt-ble.cpp
```diff
@@ -247,9 +247,32 @@ BLEObject::~BLEObject()
...
+ *            TX characteristic: 49535343-1e4d-4bd9-ba61-23c647249616
+ *            RX characteristic: 49535343-8841-43f4-a8d4-ecbe34729bb3
+ */
+static const struct uuid_match skip_characteristics[] = {
+	{ "49535343-6daa-4d02-abf6-19569aca69fe", "McLean Extreme Avoid" },
+	{ "49535343-aca3-481c-91ec-d85e28a60318", "McLean Extreme Avoid" },
+	{ "49535343-026e-3a9b-954c-97daef17e26e", "McLean Extreme Avoid" },
+	{ "49535343-4c8a-39b3-2f49-511cff073b7e", "McLean Extreme Avoid" },
+	{ NULL, }
+};
```
dirkhh, line 269: I'm curious - why list the ones we don't want and not just list the ones that we want?

dirkhh: I guess it makes the code below easier? It just seems like it would overall be more error prone?

dirkhh: thinking more about how a positive flow would work, I think I like what you did better

PR #2962, opened 2020-09-19
Update Shearwater sample time calculation

mturkia:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

At least Shearwater Cloud seems to use multiple formats for sample time
recorded in the database. Sometimes the time is in seconds, sometimes in
milliseconds, and sometime it is something I have no idea about. Thus
switching to calculating the sample id myself and using sample interval
to calculate the actual sample time. Seems to be more reliable than
trying to guess what format Shearwater is using for this specific dive.

### Changes made:

Do not even try to decipher sample time provided by Shearwater, instead calculate it using sample interval and sample count.

### DIFF
--- a/core/import-shearwater.c
+++ b/core/import-shearwater.c
@@ -87,8 +87,20 @@ static int shearwater_profile_sample(void *param, int columns, char **data, char
 	int d6, d7;
 
 	sample_start(state);
-	if (data[0])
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
this is far more complex than I'd like, but I think it's an improvement

PR #2963, opened 2020-09-19
media: read timestamp from mvhd header of MP4/QuickTime videos

bstoeger:
ExifTools (and probably other meta-data editors) modifies the
mvhd creation date, but leaves the individual creation dates
in the tracks unchanged. Therefore, use the mvhd atom.

Reported-by: Eric Tanguy <erictanguy2@orange.fr>

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This reads the movie creation date from the "mvhd" header, as this is what ExifTool writes and I guess what users expect.

### Changes made:

1) Read creation date from "mvhd" header.

### Related issues:

See discussion on mailing list.

### Release note:

Done.
### Documentation change:

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-
+- media: read timestamp from mvhd atom of QuickTime/MP4 style videos
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
 * Use this layout for new entries: `[Area]: [Details about the change] [reference thread / issue]`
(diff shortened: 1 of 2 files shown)


dirkhh:
This looks reasonable and shouldn't break what works for others already.
Is there a reason not to merge this before 4.9.7?
(while Linus and I try to fix the rfcomm mess)

bstoeger:
> This looks reasonable and shouldn't break what works for others already.

Unless they have a file where they want the creation date of the first track, not of the whole video extracted? That sounds like a very unlikely scenario, though. Therefore, I believe that this fixes more cases than it breaks. I don't even understand the point of having multiple creation times saved in the same video file. You would have to ask a videographer.

> Is there a reason not to merge this before 4.9.7?

Not that I know of.

dirkhh (approved):
I think that makes a lot of sense. And is unlikely to break things

Issue #2961, opened 2020-09-18
Dive removed from dive list when updating Current or Visibility


fmatray:
### Describe the issue:
- [X ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
When I change Visibility or Current in info tab, the dive disappear from the dive list.
I don't know if it's only a display bug or if divelog is saved without the missing dive.

### Operating system:
MAC OS

### Subsurface version:
4.9.6

### Steps to reproduce:
1) Select a Dive
2) Click on "Information"
3) Change Current or Visibility

### Current behavior:
Selected dive disappears.

### Expected behavior:
It would stay :-)

fmatray:
I also tried with Atm pressure/Altitude with the same result.

dirkhh:
Ugh, that's bad.

I just tried this with the latest development build (as that's what I have easily available on my system); modifying a dive there doesn't make it disappear.
Encouraged by this I tried the official 4.9.6 build for macOS and there, too, I can make sure edits without making the dive go missing.
Now I'm confused.
Can you create a screen video (QuickTime Player makes that easy) that shows the effect? Feel free to share it just with me (dirk at hohndel dot org) if you don't want to post it here. I'm trying to understand what's going on so we can fix it. But without being able to reproduce it, that's pretty hard...

fmatray:
I sent you a video screenshot on your email.

I guess it's just a display problem, with some kind of filter not refreshed correctly.
If I go on the tab "Dive sites" and select all dives, it appears again. But when I try to unselect, all dives disappear.

Is there a way to refresh the dive list, please ?

dirkhh:
So the video seems to imply that you have an active filter going. Can you hit Command-F and disable any filters that may be active? That's the only way I can explain this.

fmatray:
I tried several times and I get the same result.

I tried the reset button and manually checking inputs.

dirkhh:
@bstoeger I don't get this one. How can an edit on the info tab suddenly switch us into 'filtered' mode???

bstoeger:
>  I don't get this one. How can an edit on the info tab suddenly switch us into 'filtered' mode???

Are you sure about that analysis? Can I see the video?

Note that closing the filter is not enough - you have to reset it.

bstoeger:
I can partially reproduce this - when making a dive go away, the a dive is made current, but not selected in the list. Edits are then not registered. I will look into this.

[...]

PR #2965, opened 2020-09-20
dive list: on reload update filter status

bstoeger:
At some time, when introducing the global reset signal the filter
stopped being reloaded when loading a new log. This leads to very
strange UI behavior: dives disappear when editing fields unrelated
to the filter.

Therefore, when reloading the model, reset the filter. One might
argue whether this is the correct place. On the other hand, we
might even make the filter a sub-object of the dive-list model.
Let's think about this.

Partially solves #2961

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The filter-status of individual dives was not reset when loading a file, leading to very strange UI behavior.

### Changes made:

1) Explicitly apply filter when reloading dive list.

### Related issues:

See discussion in #2961.
### Release note:

Done.
### Documentation change:

No.
### Mentions:

@fmatray @dirkhh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- make sure current dive is selected [#2961]
 - update filter status when loading file [#2961]
 
 ---
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-
+- update filter status when loading file [#2961]
 
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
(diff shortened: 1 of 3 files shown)


bstoeger:
Added a second commit, which (hopefully) fixes a severe bug: Make a dive invisible by editing it - a new dive is made `current_dive` and thus shown. However, it is not selected and therefore edit-actions are ignored.

This could explain instances of users reporting that edits were lost!

I reckon this should go into current release if possible, though it certainly needs careful review / testing.

dirkhh (approved):
I agree that this might want to be a child object of the dive list.
The place where you reset the filter is odd (but then, the whole Tree/List thing is odd). What about mobile?
I played with it and it seems to do the right thing - and I definitely think it should be in 4.9.7

dirkhh:
I want to move the CHANGELOG entries straight into the ReleaseNotes (and maybe rephrase the second one slightly), so I'll merge locally...

PR #2967, opened 2020-09-21
Bailout segment is part of bottom time

atdotde:
The clock is only valid in ascent. In each cycle of the
'critial volume algorithm' it re-initialized to the
bottom time at the beginning of deco. So the time spent
on bailout should be added to this bottom time.

Thanks to Coverty for spotting this.

Coverity-scan:  CID-362079

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh (approved):
sometimes Coverity is useful...

PR #2968, opened 2020-09-21
translations: initialize water type strings at run time

### LINKED ISSUE #2966: No translation of options for water type/density in information tab (used to work)
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When manually adding a dive in a non english language, the translation of the options in water type/density remains untranslated (e.g. Brackish, Salt, Fresh, Use DC) This used to work and I tracked it back to working in 4.9.4 but has been broken in 4.9.5 and 4.9.6 (tested diffrent non-english languages with the same result)

### Operating system:

macOS Catalina 10.15.6

### Subsurface version:

Latest development build, but the issue also exists in 4.9.5 and 4.9.6.
The translation works in 4.9.4.

### Steps to reproduce:

- Change to a non english language
- Restart Subsurface
- Manually add a dive
- In the information tab,  open the drop down menu of water type/density and verify that the text remains untranslated

### Current behavior:

No translation 
### Expected behavior:

Translation of options (text) in drop down menu
### Mentions:

@dirkhh

dirkhh:
We cannot initialize strings with translated texts before the translations are loaded.
This was introduced in commit 95e6792c4f ("Grantlee: Add salinity and water
type to grantlee variables") and wasn't caught in the code review. Oops.

- [x] Bug fix

### Changes made:

create an init function for the string list
this isn't ideal, but it's the least disruptive approach so that strings don't move into a different compile unit which would cause translation disruption right before a release.

### Related issues:

Fixes: #2966

### Documentation change:

done
### Mentions:

@bstoeger


bstoeger:
I don't like the additional initialization function. This shouldn't be necessary - I'll do an alternative PR soon.

dirkhh:
> I don't like the additional initialization function. This shouldn't be necessary - I'll do an alternative PR soon.

I don't like it either. I wanted a way to keep the `tr()` calls in the same file (to avoid noise in the translations), and to initialize the `QStringList` at runtime. This was my 'quick and dirty' solution. I'm thrilled if you have a better one ;-)

bstoeger:
> I'm thrilled if you have a better one ;-)

"better" is relative. I have a solution that is consistent with what we usually do, namely that Q_TRANSLATE_NOOP thing. It's a bit convoluted, because we export that array. So I wouldn't say that it is objectively better, just less global variables.

As an alternative I would suggest making the list a "singleton", i.e. a function-local static variable, which guarantees that it is initialized on first call:
```
static QStringList initList()
{
        QStringList res;
        ... init list ...
        return res;
}

const QStringList &getTranslatedList()
{
        static QStringList list = initTranslatedList()
        return list;
}

**Comment by dirkhh** [comment]
yeah, better really is relative here :-)
I think the singleton idea is not terrible. It replaces a global function and global variable with a global function that includes a static variable... yay?

**Comment by bstoeger** [comment]
I will post both solutions and will let you choose. The QT_TRANSLATE_NOOP thing is a bit more consistent with what we usually do ... but *shrug*.

File: core/qthelper.cpp [hunk 1]
```diff
@@ -443,6 +443,16 @@ QString getUiLanguage()
 	return prefs.locale.lang_locale;
 }
 
+// the water types need to match the watertypes enum
+// since the translations aren't loaded until after the app starts, we can't simply initialize
+// that variable as a const, we need to have an init function that does this at runtime.
+QStringList waterTypes;
+
+void setupTranslatedStrings() {
+	waterTypes = QStringList() << gettextFromC::tr("Fresh") << gettextFromC::tr("Brackish") << gettextFromC::tr("EN13319") << gettextFromC::tr("Salt") << gettextFromC::tr("Use DC");
```
bstoeger, line 452: BTW: I never understood that pattern. Seems a) less readable and b) gives less chance of optimization when compared to
```
    waterTypes = QStringList {
        gettextFromC::tr("Fresh"), gettextFromC::tr("Brackish"), ...
    };
```
In the latter case, the array can be initialized with the correct size from the start.

dirkhh:
merged #2969 instead

PR #2970, opened 2020-09-21
translations: initialize water type strings on demand [second alternative to #2968]

bstoeger:
The water type strings were global static and therefore passed through
gettextFromC::tr() before main(). One would hope to get a warning
in such a case, but this is not the case.

Therefore, make the array local to a function, which guarantees
that it will be initialized on first invocation in a thread safe
manner. This is in principle the idiomatic C++ "singleton" pattern.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Second alternative to #2968. The diff is distinctly smaller. Completely untested.

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -626,8 +626,17 @@ QString get_salinity_string(int salinity)
 	return QStringLiteral("%L1%2").arg(salinity / 10.0).arg(gettextFromC::tr("g/ℓ"));
 }
 
+const QStringList &getWaterTypesAsString()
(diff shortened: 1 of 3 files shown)


dirkhh (commented):
this is slightly cleaner, but also slightly more subtle than #2969 
Given that #2969 is more consistent with what we do elsewhere, I think I'd lean that way.

PR #2972, opened 2020-09-22
cleanup: silence two compiler warnings in git-access.c

bstoeger:
gcc complained about two constructs of the kind
	remote_id && SSRF_INFO("...");
And while I am not a fan of excessive warnings, I must say
it has a point here. That's just code obfuscation. In fact,
it appears that the condition was wrong - the SSRF_INFO
should probably be invoked if remote_id is NULL. The way
it was written it would be invoked if it was *not* NULL.

Change both instances to unfancy if statements.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Silence two warnings that have annoyed me for the longest time. With this, I get warning-free compiles on gcc - I believe that we have too low warning levels, TBH. I regularly trip over code that *should* give a warning.

### Changes made:

1) Turn "fancy" boolean expression with side-effect into run-of-the-mill if statement.

### Mentions:

@dirkhh: This actually inverts the condition, please check.

### DIFF
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -505,8 +505,10 @@ static int try_to_update(git_repository *repo, git_remote *origin, git_reference
 	remote_id = git_reference_target(remote);
 
 	if (!local_id || !remote_id) {
-		local_id && SSRF_INFO("git storage: unable to get local SHA");
-		remote_id && SSRF_INFO("git storage: unable to get remote SHA");
+		if (!local_id)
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
agreed. this is better

PR #2973, opened 2020-09-23
mobile/dive-list: correctly update view when changing dive date

### LINKED ISSUE #2971: Subsurface-mobile: adding a dive manually with date in the past causes wrong dive to be shown after saving
### Describe the issue:

- [x] Bug

### Issue long description:

Manually create a new dive, change the dive date to be in the past, before existing dives in the log
save the dive
notice that the dive isn't shown in dive details

### Operating system:

Tested mobile-on-desktop on macOS, tested on Android

### Subsurface version:

current master 4.9.6.125

### Mentions:

@bstoeger - I know you never use the mobile version... and I'm not sure it has anything to do with the model. But I thought I'd add you just in case... :-)

dirkhh:
If the dive timestamp changes, the dive could move in the dive list. But the
current dive actually doesn't change (it's still the same dive, right?). Yet
we need to update the dive list as well as the shown dive (especially if this
is after adding a dive, which is first inserted with the current time and then
updated with whatever the user enters).

Fixes: #2971

- [x] Bug fix

This was first noticed when manually adding dives that were prior to dives that are already in the dive list.
In that case we didn't show that new dive after completing the edit. The reason was that the current dive stayed the same, so our role update didn't get propagated.

### Changes made:

- make sure the added dive is indeed selected
- on mobile, propagate updates even if the current dive didn't change

### Related issues:

Fixes #2971 

### Release note:

done 

### Documentation change:

N/A

### Mentions:

@bstoeger

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-
+mobile: fix manually adding dives in the past [#2971]
 
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -1414,7 +1414,7 @@ void QMLManager::selectDive(int id)
 			amount_selected++;
 	}
 	if (amount_selected == 0)
-		qWarning("QManager::selectDive() called with unknown id");
+		qWarning("QManager::selectDive() called with unknown id %d",id);
 }
 
 void QMLManager::deleteDive(int id)
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -538,9 +538,6 @@ static ShownChange updateShownAll()
 
 void DiveTripModelBase::currentChanged()
 {
-	if (oldCurrent == current_dive)
-		return;
-
 	// On Desktop we use a signal to forward current-dive changed, on mobile we use ROLE_CURRENT.
 	// TODO: Unify - use the role for both.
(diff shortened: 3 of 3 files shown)


dirkhh:
Since this is all about mobile, I'll wait to hear from @bstoeger - no need to merge before 4.9.7

mobile-widgets/qml/DiveDetailsEdit.qml
```diff
@@ -120,6 +120,7 @@ Item {
 
 		Qt.inputMethod.hide()
 		// now make sure we directly show the saved dive (this may be a new dive, or it may have moved)
+		manager.selectDive(dive_id)
```
bstoeger: Do you have a test case where this is necessary? The undo command should select the affected dive.

dirkhh: I added debug output when I was first trying to track this down and thought that I saw an unselected dive and that's when I added that line (which by itself wasn't enough to fix the problem, of course). I have now tried to remove it and can't reproduce a problem without it, so... I guess it's not needed after all. ?!?
Dunno.

bstoeger: If it is not needed - please don't add it. Sooner or later it will cost us time trying to figure out whether it can be removed.

dirkhh: ok

dirkhh:
ok, tested, works as intended, merging

Issue #2980, opened 2020-09-29
Garmin Descent MK1 Import - Location as Entry or Exit point?


brysconsulting:
### Describe the issue:

- [x] Bug
- [x] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When importing dive data from Garmin Descent MK1, should the Subsurface geo coordinates LOCATION value come from water Entry or Exit field? We see coordinates to be importing as the exit location, it would be better to record the Entry location instead.

torvalds:
Subsurface imports both, and will use the entry point as the main GPS location for the dive.

The exit point will only be visible as "GPS2" in the "Extra Info" tab of the dive (with the entry point as "GPS1")

HOWEVER.

The most common problem with the Garmin Descent is that you don't have any GPS information for the entry point at all, because you didn't go into dive mode and wait for the GPS indicator (the ring around the screen) to turn green. It can take up to a minute.

So then the Garmin won't have an entry point at all, because it never got a GPS lock before you jumped into the water.

And if that happens, subsurface will obviously use the exit point GPS value for the dive location.

In other words, the solution is to always go into dive mode before going into the water, and waiting to make sure you have GPS lock on the Descent.

brysconsulting:
Looks like my data in Subsurface is not matching your description. Please see attached screenshots. 

 

Yellow Marking – looks like the application is taking the GPS2 value and not GPS1.

 

Red Marking – if there is no Exit location value, the Subsurface Exit location is recorded with an erroneous value -0.8272…, -08727… 

 

 



 

Regards,

Martin Brys

brysconsulting:
![Screenshot 2020-09-29 133613](https://user-images.githubusercontent.com/24417638/94595554-fffbc600-0258-11eb-997a-2c58f25425dc.jpg)

Looks like my data in Subsurface is not matching your description. Please see attached screenshots. 

Yellow Marking – looks like the application is taking the GPS2 value and not GPS1.

Red Marking – if there is no Exit location value, the Subsurface Exit location is recorded with an erroneous value -0.8272…, -08727…

torvalds:
Strange. This is our code to parse that thing:

                if (!strcmp(data->key, "GPS1")) {
                        parse_location(data->value, &res);
                        /* If we found a valid GPS1 field exit early since
                         * it has priority over GPS2 */
                        if (has_location(&res))
                                break;
                } else if (!strcmp(data->key, "GPS2")) {
[...]

PR #2982, opened 2020-09-29
Prefer "GPS1" for divecomputer download dive site resolution

torvalds:
I think we only have one dive computer that supports GPS data right now:
the Garmin Descent Mk1.  It reports the dive coordinates as "GPS1" and
"GPS2" for the entry point and exit point respectively.

Often GPS1 is missing, because the dive computer may not have gotten a
GPS lock before the diver jumped into the water, so when that happens
we'll use GPS2 for the dive site location.  But when GPS1 exists, we
should prefer that.

And that's what we already did in logic in dc_get_gps_location(), but
for the initial dive site created at download time, we just picked any
divecomputer reported string that started with "GPS".  And since GPS2 is
reported after GPS1 by the Garmin Descent, it would end up overwriting
the entry point that we _should_ have preferred.

Add the same kind of "explicitly prefer GPS1" logic to the initial dive
download case as we already had elsewhere.

Reported-by: @brysconsulting

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -607,6 +607,15 @@ static void parse_string_field(device_data_t *devdata, struct dive *dive, dc_fie
 		char *line = (char *) str->value;
 		location_t location;
 
+		/* Do we already have a divesite? */
+		if (dive->dive_site) {
+			/*
+			 * "GPS1" always takes precedence, anything else
+			 * we'll just pick the first "GPS*" that matches.
+			 */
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
As we see in the comments to #2980 this clearly is a step in the right direction, so let's merge this.

PR #2985, opened 2020-09-29
[RFC] Add dive(s) to any trip

bstoeger:

- [x] Functional change
- [ ] Bug fix
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A user requested on the mailing list that dives be added to any trip. Indeed, the restriction to add dives only to the trip above/below seems illogical. This adds the possibility to add dive(s) to any trip. The trip is selected in a new dialog.

### Changes made:

1) Add trip selection dialog.
2) Move trip header formatting to qt-helper.cpp so that code can be reused by the new dialog.
3) Add context menu to add dives to any trip.

### Additional information:

1) Mostly untested
2) I noticed that the add to trip above/below feature is currently broken: Remove a dive from the middle of the top-most trip. The context menu will feature a "add to trip above" entry even though there is *no* trip above.
3) The context menu is getting absurdly long - this is in need of a redesign.
### Release note:

Done.
### Documentation change:

Perhaps?
### Mentions:

@dirkhh, @willemferguson for possible documentation change...?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- desktop: allow adding dives to arbitrary trips
 - desktop: respect page-up, page-down, home and end keys for selection change [#2957]
 - Use pO2 from prefernces for MOD display in equipment tab
 - filter: more flexible filtering system based on individual constraints
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -4,6 +4,7 @@
 #include "subsurface-string.h"
 #include "device.h"
 #include "errorhelper.h" // for verbose flag
+#include "selection.h"
 #include "core/settings/qPrefDiveComputer.h"
 
 /*
@@ -305,12 +306,8 @@ extern "C" void call_for_each_dc (void *f, void (*callback)(void *, const char *
 	for (const DiveComputerNode &node : values) {
 		bool found = false;
 		if (select_only) {
-			int j;
-			struct dive *d;
-			for_each_dive (j, d) {
+			for (dive *d: getDiveSelection()) {
 				struct divecomputer *dc;
-				if (!d->selected)
-					continue;
 				for_each_dc (d, dc) {
 					if (dc->deviceid == node.deviceId) {
 						found = true;
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -1018,15 +1018,23 @@ extern "C" char *get_current_date()
 	return copy_qstring(current_date);
 }
 
-QString get_trip_date_string(timestamp_t when, int nr, bool getday)
+QString get_trip_string(const dive_trip *trip)
 {
+	if (!trip)
+		return QString();
+
+	int nr = trip->dives.nr;
+	timestamp_t when = trip_date(trip);
+	bool getday = trip_is_single_day(trip);
+
 	QDateTime localTime = timestampToDateTime(when);
 
+	QString prefix = !empty_string(trip->location) ? QString(trip->location) + ", " : QString();
 	QString suffix = " " + gettextFromC::tr("(%n dive(s))", "", nr);
 	if (getday)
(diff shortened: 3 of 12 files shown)


dirkhh (commented):
this sounds logical - I need to play with the UI, though.

dirkhh:
The UI isn't super slick, but, yeah, this is a start.
But, in my testing, I was able to reproduce this odd effect.
- dive list has a few trips, then one top level dive, a trip with one dive, another top level dive, more trips
- add these two top level dives to a trip
- undo
- now in the dive list the two top level dives are next to each other, and the trip that was between them is instead after them

bstoeger:
Can you test whether this:
```
iff --git a/qt-models/divetripmodel.cpp b/qt-models/divetripmodel.cpp
index 539cf98ff..bc2c2ea46 100644
--- a/qt-models/divetripmodel.cpp
+++ b/qt-models/divetripmodel.cpp
@@ -1325,8 +1325,8 @@ void DiveTripModelTree::divesMovedBetweenTrips(dive_trip *from, dive_trip *to, b
 	// Cheating!
 	// Unfortunately, removing the dives means that their selection is lost.
 	// Thus, remember the selection and re-add it later.
-	divesAdded(to, createTo, dives);
 	divesDeletedInternal(from, deleteFrom, dives); // Use internal version to keep current dive
+	divesAdded(to, createTo, dives);
 }
 
 void DiveTripModelTree::divesTimeChanged(timestamp_t delta, const QVector<dive *> &dives)
```
fixes the problem?

Unrelated: The German translation is confused: "(1 Tauchgang(e))" vs. "(6 Tauchgäng(e))". Weird.

willemferguson:
I will keep an eye on this with the view of documentation changes

dirkhh:
> Unrelated: The German translation is confused: "(1 Tauchgang(e))" vs. "(6 Tauchgäng(e))". Weird.

Sweet.
That's verbatim in our translations. Oops.
```
        <translation><numerusform>(%n Tauchgang(e))</numerusform><numerusform>(%n Tauchgäng(e))</numerusform></translation>
```
looking through the various translations with `numerusform` I found a second one that's wrong. Will fix that one as well.

dirkhh:
> Can you test whether this:
> 
> ```
> iff --git a/qt-models/divetripmodel.cpp b/qt-models/divetripmodel.cpp
> index 539cf98ff..bc2c2ea46 100644
> --- a/qt-models/divetripmodel.cpp
> +++ b/qt-models/divetripmodel.cpp
> @@ -1325,8 +1325,8 @@ void DiveTripModelTree::divesMovedBetweenTrips(dive_trip *from, dive_trip *to, b
>  	// Cheating!
>  	// Unfortunately, removing the dives means that their selection is lost.
>  	// Thus, remember the selection and re-add it later.
> -	divesAdded(to, createTo, dives);
>  	divesDeletedInternal(from, deleteFrom, dives); // Use internal version to keep current dive
> +	divesAdded(to, createTo, dives);
>  }
>  
>  void DiveTripModelTree::divesTimeChanged(timestamp_t delta, const QVector<dive *> &dives)
> ```
> 
> fixes the problem?
> 

It does fix the problem

bstoeger:
Added fix and fixed CHANGELOG.md conflict.

Note that this is probably not the perfect fix. And it will probably be easiest to simply send two signals from the core (remove-dive/add-dive) instead of one (move-dive) to avoid sorting inconsistencies.

dirkhh (approved):
LGTM

dirkhh:
You are of course encouraged to send a better fix in a future PR :-)

dirkhh:
ignoring the GitStorage fail - there clearly is something odd happening on the backend. Which once again I cannot reproduce.
Oh how I love this test.

PR #2987, opened 2020-09-29
Use correct pO2 when computing MOD in equipment tab

### LINKED ISSUE #2984: MOD information in equipment tab doesn't consider Max pO2 setting
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
In a nitrox dive, the MOD information on equipment tab doesn't consider max pO2 setting in preferences. It is considering always a 1.40 max pO2

### Operating system:
Windows 8. Not tested in another OS

### Subsurface version:
4.9.7

### Steps to reproduce:

1) Set the max pO2 setting in Preferences > Tech setup to 1.60 bar;
2) Setup a nitrox dive with 32% O2, for example;
3) Compare MOD value showed in equipment tab with MOD value in profile information. In this example, profile information is correct

### Current behavior:
Values are different
mod

### Expected behavior:
Same values. In the example above, 40m

atdotde:
The cylinder model is used both in the planner and the
equipment tab. We have three preferences for the pO2 that
is used to compute MOD: In the planner, there is one for
the bottom part of the dive and another one for deco.
Those are set in the planenr UI. There is another value,
controlled in the Tec Prefernces. That one should be
used in the equipment tab rather than the one from
the planner.

Fixes #2984

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Use pO2 from prefernces for MOD display in equipment tab
 mobile: fix manually adding dives in the past [#2971]
 
 ---
--- a/core/dive.c
+++ b/core/dive.c
@@ -3628,8 +3628,9 @@ int get_depth_at_time(const struct divecomputer *dc, unsigned int time)
 fraction_t best_o2(depth_t depth, const struct dive *dive)
 {
 	fraction_t fo2;
+	int po2 = in_planner() ? prefs.bottompo2 : prefs.modpO2 * 1000;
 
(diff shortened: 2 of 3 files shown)


dirkhh (approved):
that looks good. Thanks for jumping on it

PR #2988, opened 2020-09-30
Remove empty vertical space from filter QListWidget

willemferguson:
In the filter the dropdown lists for selecting dive mode or day-of-week
have a lot of white space at the bottom. This PR removes that white space.
Actually the white space at the bottom of a QListWidget appears to
be a known bug (actually an omission) for the current Qt V15. The above
solution is a brute-force workaround to achieve the same end result.
The active line is actually the setFixedSize(). The other line, however,
comprises good QT layout policy to minimise widget size. The setFixedSize()
function does: "scale the widget to the width of the only column and to
n-times the height of one row and add the size of the 'frame' twice".

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@bstoeger @dirkhh

### DIFF
--- a/desktop-widgets/filterconstraintwidget.cpp
+++ b/desktop-widgets/filterconstraintwidget.cpp
@@ -45,7 +45,9 @@ static QListWidget *makeMultipleChoice(const QModelIndex &index, int role)
 		return nullptr;
 	QListWidget *res = new QListWidget;
 	res->addItems(list);
+	res->setSizePolicy(QSizePolicy::Maximum,QSizePolicy::Maximum);
 	res->setSelectionMode(QAbstractItemView::ExtendedSelection);
+	res->setFixedSize(res->sizeHintForColumn(0) + 2 * res->frameWidth(), res->sizeHintForRow(0) * res->count() + 2 * res->frameWidth());
 	return res;
 }
 


dirkhh (changes requested):
I'm always nervous when we do these `FixedSize` things as we keep learning that between all the different OSs and screen resolutions those often end up biting us later.
Could you add a comment that explains how you arrived at the formula? This way if in a year one of us looks at the code it likely will make more sense...

bstoeger:
In all fairness, the code *reads* sensible to me. It says "scale it to the width of the only column and to n-times the height of one row and add the size of the 'frame' twice". Now, if that is portable across architectures ... who knows?

willemferguson:
Yup, Berthold is correct. I updated the commit message accordingly. I know, I hate this approach as well. I've seen the problems it creates in the Information panel of Subsurface. But at the moment it is better to try and do something about the problem than just leaving it unattended. Under normal circumstances the setSizePolicy should do it but we are dependent on the Qt team correcting this deficiency and it looks like it is not high on their priority list.

PR #2989, opened 2020-09-30
Update libdivecomputer to fix Garmin GPS recording

torvalds:
The Garmin Descent parser had a bug that would re-use possibly stale GPS
locations between dives (and in theory other data fields too, although
in practice I think only GPS data was ever leaked between dives).

This updates libdivecomputer to a fixed version.

Reported-by: @brysconsulting
Link: https://github.com/subsurface/subsurface/issues/2980

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


torvalds:
Oh. For some reason "TestGitStorage" fails. And yes, it fails locally here too. Very odd, since I didn't change anything in that area.

torvalds:
Hmm. TestGitStorage fails after 15s (timeout?) even without this change, so I think that failure is entirely unrelated.

torvalds:
Forced an update just with a changed commit date, to see if it perhaps was just the power outage at Dirk's that caused the failre.

Maybe there's some "force redo of checks" button that I didn't find and could have tried instead.

dirkhh:
Git storage failed because of my infrastructure being down. Because the cloud storage was down.
You should be able to just restart the tests, but a force push does that as well.
Sorry for the inconvenience

torvalds:
Well, there may be something still wrong. The GitStorage test still fails here locally. I should have tried that out before force-pushing, because I suspect the github tests will fail again.

But as mentioned, I don't think the test failure has anything to do with this merge request. Maybe something still messed up with the infrastructure?

dirkhh:
always possible. bringing this back up in the middle of a busy work day is not really a great plan...

dirkhh:
So this is very strange. Cloud storage seems to work, based on a few quick tests between meetings, but the `TestGitStorage` times out and I can't tell from the logs why. I'm sure it's something incredibly stupid and once I figure it out you'll hear a loud groan from this end... but right now... err... Hu?

dirkhh:
For me the test completes locally - just oddly slow. I'm still completely baffled why it fails here. But given that this change truly cannot be the cause of that, let's just get over it and merge it...

PR #2991, opened 2020-10-01
membuffer: be defensive about bad C library vsnprintf implementations

torvalds:
Dirk reports that some Windows users have had odd corruption in the
commit messages in the cloud storage.  They make no sense at all unless
there is some very weird Windows library bug.

The prime suspect is 'vsnprintf()' returning a negative error when the
target buffer is too small (rather than the proper "this is how much
space it would need").  That is a very traditional C library bug that I
thougth had been fixed everywhere, but there doesn't really seem to be a
lot of other likely causes.

So let's make our membuffer code be defensive against bad libraries that
return negative error numbers from vsnprintf.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/membuffer.c
+++ b/core/membuffer.c
@@ -110,6 +110,18 @@ void put_vformat(struct membuffer *b, const char *fmt, va_list args)
 		len = vsnprintf(target, room, fmt, copy);
 		va_end(copy);
 
+		// Buggy C library?
+		if (len < 0) {
+			// We have to just give up at some point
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
I think this is certainly not wrong, so let's merge it and hope that it helps us to make the weird partial commit messages go away

Issue #1787, opened 2018-10-15
Create filter option for tank size


jkingsman:
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [x] New feature request
- [x] Discussion request

### Issue long description:
As a diver, I sometimes want to compare my weighting based on a given suit and tank size (e.g. I know I've dived XXX suit with an HP100 before; how much weight did I use?). Adding a filtering option for tank size (along with the already-present tags, person, location, and suit filtering) would solve this problem easily.

One potential risk is that it pushes the boundaries further for "what about a filter for XYZ" to the potential nth degree; not sure how big of a concern that is.

If this is a feature that there is interest in, I'm happy to do my best on a first draft -- I know it would require doc updates, etc. I've been shadowing the mailing list for a while; hopefully this is the right place to post an idea for feedback and discussion -- feel free to delete or let me know a better place to raise this if I've misstepped 🙂

janmulder:
As a current workaround, you can add the weight + cylinder to the divelist displayed columns and sort on those. Notice that the divelist only displays the first cylinder (for multi cylinder dives).

> One potential risk is that it pushes the boundaries further for "what about a filter for XYZ" to the potential nth degree; not sure how big of a concern that is.

Well, this is always some sort of concern. There seems room for it in the UI, so in this respect this could be added.

That said, my main concern is that the cylinder type is just a tiny part in the whole buoyancy equation. For example, water type, cylinders with same size/type but different weight, amount of undergarment when dry-suit diving, and even the comfort level of this diver on that dive. So, in this respect: its not useful to me to add this to the UI.

jkingsman:
That's definitely a fair point re: not actually solving the problem of buoyancy factors; this might not be as helpful as I imagined.

PR #2993, opened 2020-10-02
Implement rudimentary filtering for gas mixes

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This implements rudimentary filtering for gas mixes as requested on the mailing list. Currently, it only implements a "any-of" mode. Moreover, beware: virtually untested.

This includes some reshuffling of the gas handling code, which now features a general "give me arbitrary gas fraction" function.

### Changes made:

1) Implement general get-gas-component-fraction function.
2) Implement gas-component-percentage numerical constraint.

### Related issues:

See mailing list posting.

### Release note:

Already included.

### Documentation change:

Filter documentation does not yet exist to my knowledge.

### Mentions:

@dirkhh
@atdotde: see gas-function changes
@willemferguson


dirkhh:
minuscule comment - feel of course free to ignore. There is a tiny, marginal, odd advantage for **my** process when testing PRs that makes me like it better if you create the PR branches in the `subsurface/subsurface` repo vs. the `bstoeger/subsurface` repo. I only noticed because recently you have done the former a few times but now you went back to doing the latter, and I figured it might be worth telling you which one I prefer :-)

bstoeger:
> I only noticed because recently you have done the former a few times but now you went back to doing the latter, and I figured it might be worth telling you which one I prefer :-)

That was just for my convenience,  because I didn't want to checkout another branch besides the huge filter thing. I can of course do future development in the official repository - no problem.

dirkhh (approved):
This is... absolutely awesome. And a testament to the strength of the new filter input method.
BTW, we should post something on the developer mailing list that explains how the new filter system works and which parts of it are worth testing...

dirkhh:
> > I only noticed because recently you have done the former a few times but now you went back to doing the latter, and I figured it might be worth telling you which one I prefer :-)
> 
> That was just for my convenience, because I didn't want to checkout another branch besides the huge filter thing. I can of course do future development in the official repository - no problem.

To be clear, this is a minor convenience for me. Only do this if it makes no difference to you. When you have the branches in the main repo it saves me a bunch of commands locally. And we all know that I am lazy, right?

core/gas.h
```diff
@@ -41,12 +43,18 @@ static inline int get_he(struct gasmix mix)
 	return mix.he.permille;
 }
 
+static inline int get_n2(struct gasmix mix)
```
atdotde, line 46: This should have been added long long ago. I am sure, we will find many other places where it can replace a convoluted 1000 - o2 -he construction.

core/deco.c
```diff
@@ -293,7 +292,7 @@ double tissue_tolerance_calc(struct deco_state *ds, const struct dive *dive, dou
 /*
  * Return Buehlmann factor for a particular period and tissue index.
  */
-static double factor(int period_in_seconds, int ci, enum inertgas gas)
+static double factor(int period_in_seconds, int ci, enum gas_component gas)
 {
 	if (period_in_seconds == 1) {
 		if (gas == N2)
```
atdotde, line 298: Now, that there are more than two possible values, maybe this should be a switch statement (with the O2 case either throwing an error or giving a result of 1.0 corresponding to an infinite half-time for oxygen).

atdotde (approved):
I like this a lot.

bstoeger:
Added a filter constraint for cylinder size as requested in #1787.

core/deco.c
bstoeger: Fair point. However, since this is a `static` function, it is not accessible outside the source file, so it should be easy to check that no invalid values are provided.

dirkhh (approved):
I so love the new filter code. This is just brilliant.

PR #2997, opened 2020-10-03
planner: limit depth and time of segments to sane values

### LINKED ISSUE #2762: Depth field allows large values
### Description
When entering depth application freezes. See the provided screenshot.
- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
The "Final depth" field in Dive planner points allows meaningless values to be entered.

### Operating system:
Windows 10

### Subsurface version:
Desktop Subsurface 4.9.3

### Steps to reproduce:
1. Create dive manually in planner
2. Enter values from screenshot in depth field fill "50000" meters
3. Application freezes and shows meaningless chart

### Current behavior:
Application freezes and shows meaning-less chart.
Not checked on another OS.

### Expected behavior:
Nothing is calculated, error popup is shown (based on other fields error behavior), chart isn't refreshed (or is empty).

### Additional information:
See screenshot with current behavior.
subsurface_bug_depth

### Mentions:
N/A

bstoeger:
Using non-sensical depth and times for segments in the planner may
lead to an unresponsive UI. Therefore limit depth to 1000 m/3000 ft
and time to 600 min. Limiting of depth is done twice: once to a
value of 3000 in the UI, once in the DivePlannerModel. The reason
being that the user might switch units and we don't want to restrict
them to 1000 ft, when they switch from m. Time is simply restricted
in the UI, since we (thankfully) support only one time unit.

Fixes #2762.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This implements the restriction of depth in the planner as suggested in #2762, see discussion therein.

As a scientist myself, I understand Robert's stance ("garbage in - garbage out" or "play stupid games win stupid prices"). However, from a UI point of view, I think the reporter has a fair point. Why allow the calculation of plans the planner is not made for (I guess), or worse - which are unphysical, as in deeper as the deepest point on earth? Overall, I feel that the UI inconvenience of a sluggish UI outweighs the gain of planning unrealistic dives.

If someone really wants to create such crazy plans, I suggest adding a preferences option "max planner depth".

### Changes made:

1) Limit depth and time of planner segments.

### Related issues:

Fixes #2762.

### Release note:

Probably not needed.

### Documentation change:

No.

### Mentions:

@atdotde

### DIFF
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -169,9 +169,10 @@ DivePlannerWidget::DivePlannerWidget(QWidget *parent, Qt::WindowFlags f) : QWidg
 	connect(closeKey, &QShortcut::activated, plannerModel, &DivePlannerPointsModel::cancelPlan);
 
 	// This makes shure the spinbox gets a setMinimum(0) on it so we can't have negative time or depth.
-	ui.tableWidget->view()->setItemDelegateForColumn(DivePlannerPointsModel::DEPTH, new SpinBoxDelegate(0, INT_MAX, 1, this));
+	// Limit segments to a depth of 1000 m/3300 ft and a duration of 100 h. Setting the limit for
+	// the depth will be done in settingChanged() since this depends on the chosen units.
 	ui.tableWidget->view()->setItemDelegateForColumn(DivePlannerPointsModel::RUNTIME, new SpinBoxDelegate(0, INT_MAX, 1, this));
-	ui.tableWidget->view()->setItemDelegateForColumn(DivePlannerPointsModel::DURATION, new SpinBoxDelegate(0, INT_MAX, 1, this));
+	ui.tableWidget->view()->setItemDelegateForColumn(DivePlannerPointsModel::DURATION, new SpinBoxDelegate(0, 6000, 1, this));
 	ui.tableWidget->view()->setItemDelegateForColumn(DivePlannerPointsModel::CCSETPOINT, new DoubleSpinBoxDelegate(0, 2, 0.1, this));
 
 	/* set defaults. */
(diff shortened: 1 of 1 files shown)


atdotde:
I can live with the depth limit as this is clearly far beyond that the deco models are made for (and human divers probably as well). But I personally know subsurface users who do dives that last longer than 10 hours so that restriction is too short (if we really need it).

bstoeger:
> But I personally know subsurface users who do dives that last longer than 10 hours so that restriction is too short (if we really need it).

Upped to 100 h. If someone really needs more than that - they can add multiple 100 h segments. There should be *some* maximum - I just managed to crash the application by setting a huge value. The application shouldn't be crashed that easily.

atdotde:
It should not crash. We should investigate this. (Thanks to the binary search for deco times, the decompression algorithm should take log(dive time) steps).

bstoeger:
I think it crashed on an out-of-memory condition. Perhaps too large "polygon" or something. At least in a previous test I got Qt messages complaining about that.

desktop-widgets/diveplanner.cpp
```diff
@@ -169,9 +169,11 @@ DivePlannerWidget::DivePlannerWidget(QWidget *parent, Qt::WindowFlags f) : QWidg
 	connect(closeKey, &QShortcut::activated, plannerModel, &DivePlannerPointsModel::cancelPlan);
 
 	// This makes shure the spinbox gets a setMinimum(0) on it so we can't have negative time or depth.
-	ui.tableWidget->view()->setItemDelegateForColumn(DivePlannerPointsModel::DEPTH, new SpinBoxDelegate(0, INT_MAX, 1, this));
+	// Limit segments to a depth of 3000 m/ft and a duration of 100 h. The DivePlannerPointsModel will
+	// sanitize depth according to the current units.
+	ui.tableWidget->view()->setItemDelegateForColumn(DivePlannerPointsModel::DEPTH, new SpinBoxDelegate(0, 3000, 1, this));
```
dirkhh: I know this is ridiculous nit picking... but... we should do 1000m/3000ft or 3000m/9000ft - having a different max depth depending on metric or imperial just... bugs me.

qt-models/diveplannermodel.cpp
```diff
@@ -334,6 +334,9 @@ bool DivePlannerPointsModel::setData(const QModelIndex &index, const QVariant &v
 		case DEPTH:
 			if (value.toInt() >= 0) {
 				p.depth = units_to_depth(value.toInt());
+				// Restrict depth to 1000 m - this is way beyond sane anyway.
+				if (p.depth.mm > 1000 * 1000)
+					p.depth.mm = 1000 * 1000;
```
dirkhh: so we could use `M_OR_FT(1000,3300)` for the SpinBox above.

dirkhh (changes requested):
I think it's sane. 1000m and 100h are certainly at the very very edge of what the algorithms used have even the slightest bit of testing for. So this makes sense to me.

See my comment on potential improvement.

desktop-widgets/diveplanner.cpp
bstoeger: Sure, we can catch it here, but then the "item delegate" would have to be changed when the user changes the units. Probably no big deal - I will see what I can do.

bstoeger: Done.

dirkhh (approved):
that works

dirkhh:
given that @atdotde 's concern was addressed (and that TestGitStorage is back in "fails randomly" territory), I'll merge

Issue #2999, opened 2020-10-03
HID devices like the Suunto Eon Core/Steel don't work on macOS


dirkhh:
### Describe the issue:

- [x] Bug

### Issue long description:

on macOS (tested on 10.15) we can't seem to open the HID device, even though the app is linked against the HIDAPI dylib

### Operating system:

macOS 10.15.7

### Subsurface version:

both 4.9.7 and current master

dirkhh:
this should be addressed in the current builds (and in 4.9.8)

PR #3000, opened 2020-10-03
Better libdc debug

dirkhh:

- [x] Functional change

make the `dev_info()` text available on `stderr` in verbose mode
try harder to correctly show the transports that we are trying to use to talk to the dive computer

### Related issues:

#2999 - turns out the claim in that issue appears wrong, but this is related as it helped debug what's going on


dirkhh:
I'm getting really tired of those `TestGitStorage` failures that make no sense to me.
And tracing them when running them locally shows me nothing (of course, the git storage backend is also here locally - so if it's some weird timing thing, that doesn't help).

PR #3001, opened 2020-10-04
Use get_n2 helper function

atdotde:
Now, that we have this helper function that should have been
introduced long ago, we can make some more expressions
more idiomatic.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/gas.c
+++ b/core/gas.c
@@ -16,7 +16,7 @@ bool isobaric_counterdiffusion(struct gasmix oldgasmix, struct gasmix newgasmix,
 {
 	if (!prefs.show_icd)
 		return false;
-	results->dN2 = get_he(oldgasmix) + get_o2(oldgasmix) - get_he(newgasmix) - get_o2(newgasmix);
+	results->dN2 = get_n2(newgasmix) - get_n2(oldgasmix);
 	results->dHe = get_he(newgasmix) - get_he(oldgasmix);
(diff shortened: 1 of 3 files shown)


core/gas.c
```diff
@@ -16,7 +16,7 @@ bool isobaric_counterdiffusion(struct gasmix oldgasmix, struct gasmix newgasmix,
 {
 	if (!prefs.show_icd)
 		return false;
-	results->dN2 = get_he(oldgasmix) + get_o2(oldgasmix) - get_he(newgasmix) - get_o2(newgasmix);
+	results->dN2 = get_n2(oldgasmix) - get_n2(newgasmix);
```
bstoeger: Doesn't that invert the sign of dN2? I.e. shouldn't that be
```
	results->dN2 = get_n2(newgasmix) - get_n2(oldgasmix);
```
?

bstoeger:
Looks good to me, up to one comment.

core/gas.c
atdotde: Oops.

bstoeger:
Looks good - thanks (one might argue that "inert = 1000 - o2" is safer with respect to adding other "intert" gases such as H2, but I reckon that this is the slightest of the problems).

PR #3003, opened 2020-10-04
clenup: remove bogus connect statement.

bstoeger:
In OstcFirmwareCheck::saveOstcFirmware() we find the connect() call
    connect(dialog, SIGNAL(finished(int)), config, SLOT(dc_close()));
whereby "config" is of the type "ConfigureDiveComputer".
However, the function signature of ConfigureDiveComputer::dc_close
reads as
    void dc_close(device_data_t *data);
and indeed "data" is accessed inside the function. I don't understand
how this doesn't crash, but clearly something is amiss.

Let's remove that connect statement.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I found this super-ominous connect statement (see commit description). I cannot make any sense of this and don't understand how this does not crash. Does one get console messages when updating firmware concerning an invalid connect statement?

Can we remove this, or clear this up, please?

Obviously, I have no device of that kind.

### Changes made:

1) Remove apparently bogus connect statement

### Mentions:

@glance- @dirkhh

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.cpp
+++ b/desktop-widgets/configuredivecomputerdialog.cpp
@@ -347,7 +347,6 @@ void OstcFirmwareCheck::saveOstcFirmware(QNetworkReply *reply)
 	connect(config, SIGNAL(message(QString)), dialog, SLOT(setLabelText(QString)));
 	connect(config, SIGNAL(error(QString)), dialog, SLOT(setLabelText(QString)));
 	connect(config, SIGNAL(progress(int)), dialog, SLOT(setValue(int)));
-	connect(dialog, SIGNAL(finished(int)), config, SLOT(dc_close()));
 	config->dc_open(&devData);
 	config->startFirmwareUpdate(storeFirmware, &devData);
 }


atdotde:
I hope to have a working OSTC again later this week. Will check.

glance-:
I agree this statement looks bogus. The new's doesn't have any corresponding delete's either, so somethings fishy.

I guess the right thing to do is to have ConfigureDiveComputer constructor take  pointer to device_data_t and it uses that for its calls. That would make more sense in this code.

bstoeger:
Yes, this thing leaks left and right. I think that was to get the ownership-mess under control, see commit 4559a26e316931695f892b205257979f0ea13afa. Apparently, the dialog can be less long lived than the worker threads...?

I don't dare to touch that at the moment, since I don't have the corresponding devices and I don't have an overview of the data/control flow.

In any case, I think that particular connect statement should be removed.

dirkhh (approved):
UGH. Yeah... no, just no.
First line of `dc_close(device_data_t *data)`:
```
	if (data->device)
```
So there is no way this goes well when called from that signal.

PR #3007, opened 2020-10-07
Replacegrantlee

dirkhh:

- [x] Code cleanup
- [x] Build system change

**from @atdotde 's original PR**
I was looking into our printing capabilities (which so far, I have never used) and stumbled getting Grantlee to work for me. Instead of wasting more time to get it to work I decided to reimplement the small subset of its features that we actually use directly.

Even if this code will not make it to master (which I still hope it will eventually do as I consider this a significant simplification), for me this was a nice exercise in implementing a parser for this very limited programming language.

Given that the amount of code to implement the whole thing is not much bigger than the code we used to call Grantlee I think this is a worthwhile simplification. The whole Grantlee library is far too heavy weight  and complex for those few replacements we actually use in our templates.

In the process, I also learned that our code actually circumvents QObject's introspection interface that Grantlee was supposed to make use of so using this heavy lifting wasn't really justified in the first place.

I made sure that at least those templates that we ship with the source work as before.

### Related issues:

this is a replacement for #2924 

### Release note:

missing

### Documentation change:

missing - we do mention Grantlee a lot in the documentation, surprisingly

### Mentions:

@atdotde


dirkhh:
I cancelled the docker image runs - they aren't useful at this stage. And I restarted the one that once again failed on TestGitStorage...

atdotde:
I cloned this in a pristine directory and it seems it work well (building using build.sh). Only that at first I forgot the  -build-with-webkit option and then it failed to compile mainwindow.cpp complaining it cannot include QPrintdialog (which is protected by #ifndef NO_PRINTING but that seems to be not effective (even though without webkit printing is likely going to fail).

atdotde (approved):
For me the gas change icons tend to be on the smaller side
Screenshot 2020-10-08 at 19 30 35

dirkhh:
> For me the gas change icons tend to be on the smaller side

That is fascinating. I know we talked about the ratio between icon size and profile size and text size in the past... Definitely a different PR, for sure. But something we need to look into.

In the meantime, I'll merge this and we can see what it breaks :-)

PR #3011, opened 2020-10-10
Make device table sortable

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This makes the device-table sortable. This may seem like a complete waste of time, and indeed, that's not the point of the PR. The real point lies in the second commit: sort the device-table in core in the way core needs it, not how the UI wants it. Let the UI do its own sorting. Implementing arbitrary sorting then was just "low hanging fruit". Ultimately, the goal is to replace the call-back by run-of-the-mill loops.
 
### Changes made:

1) Add prox-model on top of DiveComputerModel to make the list sortable by arbitrary column
2) Sort core table according to id/model
3) Remove copy/sort of core table in call_for_each_dc()

### Related issues:

See previous device.cpp cleanups.

### Release note:

Too trivial/unimportant.

### Documentation change:

No.

### DIFF
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -210,7 +210,7 @@ bool device::operator!=(const device &a) const
 
 bool device::operator<(const device &a) const
 {
-	return std::tie(model, deviceId) < std::tie(a.model, a.deviceId);
+	return std::tie(deviceId, model) < std::tie(a.deviceId, a.model);
 }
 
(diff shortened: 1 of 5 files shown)


dirkhh (approved):
I'm not quite sure I understand the final goal (yes, I read about replacing the call back... but I need to look at this in context to understand why that matters), but this all makes sense.

PR #3012, opened 2020-10-11
Set the NO_PRINTING variable properly

atdotde:
This fixes building without printing support.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh (approved):
The comment was exactly right.
I HATE the `NO_PRINTING` negative logic here. Because I had looked at the definition of the option and I remember thinking "ok, this defaults to "off", so we need to make sure we handle the "on" case". Which is of course where this all went wrong. Because it defaults to `NO_PRINTING` as `OFF`, so printing is on. ARGL.

dirkhh:
the test is once again TestGitStorage - so until I find the time to (for the fourth time?) figure out what's wrong there, I'll just allow myself to ignore that.

PR #3017, opened 2020-10-13
core: fix detection of duplicate device names

bstoeger:
Recently (c9b8584bd2) the sort criteria of the device-table
was changed from  (model/id) to (id/model). However, that
messed with the detection of duplicate device names: there,
the code searched for the first element greater or equal
to (model / 0).

With the reversal of the sort criteria, this would now
always give the first element.

Therefore, do a simple non-binary search, which is much
more robust. The binary search was a silly and pointless
premature optimization anyway - don't do such things
if not necessary!

Since only one place in the code search for existence
for a model-name, fold the corresponding function into
that place.

Moreover, change the code to do a case-insensitive compare.
This is consistent with the dc_match_serial() code in
core/libdivecomputer.c, where matching models is
case-insensitive!

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a bug recently introduced in c9b8584bd2. Note that the whole binary-search thing was silly to begin with. This also includes a functional change: the search for an existing model is now case-insensitive. The code in `core/libdivecomputer.c` is also case-insensitive, so let's try to be consistent.

### Changes made:

1) Replace binary search by linear search for model string.
2) Make comparison case-insensitive.

### Additional information:

I did not test this!

### Release note:

No - bug not in release.

### DIFF
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -219,12 +219,6 @@ static const device *getDCExact(const QVector<device> &dcs, const divecomputer *
 	return it != dcs.end() && it->model == dc->model && it->deviceId == dc->deviceid ? &*it : NULL;
 }
 
-static const device *getDC(const QVector<device> &dcs, const divecomputer *dc)
-{
-	auto it = std::lower_bound(dcs.begin(), dcs.end(), device{dc->model, 0, {}, {}, {}});
-	return it != dcs.end() && it->model == dc->model ? &*it : NULL;
-}
-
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
makes sense

dirkhh:
merged

PR #3018, opened 2020-10-13
build-system/macOS: remove Qt SQL plugins and suppress some errors

dirkhh:
Technically with this the app might be ready for AppStore inclusion. I don't see myself
spending the energy on that, TBH.

- [X] Build system change

There were a bunch of odd errors and warning when running `make install`
This should take care of most of them, and in the process removes a few plugins that we used to bundle on macOS and never needed, resulting in a marginally smaller app.

### Changes made:

- explicitly remove some directories before manually copying over them
- explicitly remove the SQL plugins
- explicitly request app store compliance


dirkhh:
merging locally to avoid GitHub Action / test build onslaught...

PR #3022, opened 2020-10-14
More device code reshuffling

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Continued work to unglobalize parser and downloader-state. This makes the device-list accessible from C and adds device-table pointers to functions. Thus, every place that accesses the global device-table can now be easily grep-ed for.

To make things accessible from C, turn QStrings into UTF8 std::strings. The latter are stored as C strings and don't have to be converted. However, that requires more defensive code, since you can't construct an std::string with a null-pointer.

This also contains a functional change: models are now compared as case-insensitive for consistency with the old code in core/libdivecomputer.cpp

### Related issues:

See previous changes.

### Additional information:

This needs some testing! I will test the export-part in the upcoming days, but I can't easily test the download from different devices of the same model part.

### Release note:

No.
### Documentation change:

No.
### Mentions:

@dirkhh

### DIFF
--- a/core/datatrak.c
+++ b/core/datatrak.c
@@ -565,7 +565,7 @@ static unsigned char *dt_dive_parser(unsigned char *runner, struct dive *dt_dive
 		dt_dive->dc.deviceid = 0;
 	else
 		dt_dive->dc.deviceid = 0xffffffff;
-	create_device_node(dt_dive->dc.model, dt_dive->dc.deviceid, "", "", dt_dive->dc.model);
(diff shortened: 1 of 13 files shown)


dirkhh (approved):
took me a moment longer to review
Your attention to detail when cleaning up the mess that we have in our code continues to leave me in awe.

dirkhh:
merging locally to avoid GitHub Action / test build onslaught...

PR #3026, opened 2020-10-16
parser: fix potential crash when parsing DAN files(?)

bstoeger:
The last two parameters of the parse_dan_format() function were
mixed up: sites should come before filter_presets.

This should have caused crashes, for DAN files with dive sites.
I don't understand why this didn't cause compiler warnings.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In my opinion, this should fix a crash when importing DAN files with dive-sites (is there such a thing?).
However, what I really don't understand here: why don't I get compiler warnings - neither with gcc nor clang? Am I missing something?

### Changes made:

1) Fix parameters in parse_dan_format()

### Mentions:

@dirkhh: Can you explain the lack of compiler warnings!?

### DIFF
--- a/core/import-csv.c
+++ b/core/import-csv.c
@@ -104,8 +104,8 @@ static char *parse_dan_new_line(char *buf, const char *NL)
 
 static int try_to_xslt_open_csv(const char *filename, struct memblock *mem, const char *tag);
 static int parse_dan_format(const char *filename, char **params, int pnr, struct dive_table *table,
-			    struct trip_table *trips, filter_preset_table_t *filter_presets,
-			    struct dive_site_table *sites)
+			    struct trip_table *trips, struct dive_site_table *sites,
+			    filter_preset_table_t *filter_presets)
 {
 	int ret = 0, i;
 	size_t end_ptr = 0;


core/import-csv.c
```diff
@@ -104,8 +104,8 @@ static char *parse_dan_new_line(char *buf, const char *NL)
 
 static int try_to_xslt_open_csv(const char *filename, struct memblock *mem, const char *tag);
 static int parse_dan_format(const char *filename, char **params, int pnr, struct dive_table *table,
-			    struct trip_table *trips, filter_preset_table_t *filter_presets,
-			    struct dive_site_table *sites)
+			    struct trip_table *trips, struct dive_site_table *sites,
+			    filter_preset_table_t *filter_presets)
```
dirkhh, line 108: based on your own include file, for C code `filter_preset_table_t *` is the same as `void *` thanks to https://github.com/subsurface/subsurface/blob/2e5913d2ba6c5b5c384121692045be52836ac6fa/core/filterpreset.h#L31
And since `void *` is compatible with any pointer in C...

dirkhh (approved):
definitely the right thing to do. Surprised that didn't crash. Maybe it no one used this since we added the new filter code?

core/import-csv.c
bstoeger: Argh. That's stupid. We have to change that.

PR #3028, opened 2020-10-17
Cleanup: make filter_preset_table a regular struct

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This makes `filter_preset_table` an anonymous struct instead of void in C code, so that warnings are generated when passing the wrong arguments. Moreover, remove the typedef `filter_preset_table_t`, since we generally don't use that for the other tables.
### Changes made:

1) make `filter_preset_table` an anonymous struct.
2) remove `filter_preset_table_t` typedef.

### DIFF
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -18,7 +18,7 @@ void addDive(dive *d, bool autogroup, bool newNumber)
 }
 
 void importDives(struct dive_table *dives, struct trip_table *trips, struct dive_site_table *sites,
-		 filter_preset_table_t *presets, int flags, const QString &source)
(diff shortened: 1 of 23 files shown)


bstoeger:
+ a few more minor cleanups / fixes.

dirkhh (approved):
I think these are great cleanups and fixes. Thank you.
I'll merge the two PRs locally and then push to master so we get only one set of new builds

PR #3029, opened 2020-10-17
Refactor the xml-params code

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The xml-parameter code leaks memory right and left. Instead of trying to fix this (a very daunting task), I decided to write a small C++ helper class and implement a very simple policy: the function that allocs the parameters is responsible for freeing them. In C++, this is automatic, by making the parameters block a simple value type. In C, call `alloc_xml_params()` then `free_xml_params()` in the same function.

Note: this is totally untested. FWIW, TestParse completes (not much, because TestParse only compares the number of imported dives, not their content!).

### Changes made:

1) Create and use C++ xml-parameter helper class..
2) Remove unused function.

### Mentions:

@mturkia @dirkhh

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -189,6 +189,8 @@ set(SUBSURFACE_CORE_LIB_SRCS
 	worldmap-options.h
 	worldmap-save.c
 	worldmap-save.h
+	xmlparams.cpp
+	xmlparams.h
 	xmp_parser.cpp
 	xmp_parser.h
 
--- a/core/import-csv.c
+++ b/core/import-csv.c
@@ -13,6 +13,7 @@
 #include "gettext.h"
 #include "import-csv.h"
 #include "qthelper.h"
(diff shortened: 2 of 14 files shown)


bstoeger:
PS: The date/time code in `parse_seabear_csv_file()` is *extremely* obscure to me. Please check.

mturkia:
Something like this should have been done years ago! After initial reading, this looks reasonable, but I will still have to take another look and do a few tests as well.

bstoeger:
Thanks for looking into this - take your time.

I changed `std::unique_ptr` to `std::vector` because that gives us stronger stability guarantees, which are now documented. Moreover, it arguably makes the code easier to read.

dirkhh (approved):
definitely better than what we had.

PR #3031, opened 2020-10-17
[TEST] try whether compilation with C++17 works

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] Test PR

Out of curiosity, try whether compiling with C++17 runs smoothly.

Let's face it: C++11 can nowadays be considered as archaic. There is nothing ground-breaking in C++14 and C++17 for our purposes (as the language goes in a very different direction than the way we use it). However, there are a few nice features from the "syntactic sugar" department here and there.


dirkhh:
Looks like "YES IT WORKS" to me.
So, pray tell... what are we gaining with that?

dirkhh:
So this is marked as TEST.
Should we merge it? Or are you closing the PR now that you know this works?

bstoeger:
Since the oldest Qt Version we support is Qt 5.9 (released 2017), I suggest increasing the C++ dialect to at least C++14, which was well established back then.

Yes, we can live without its features. However, C++11 is becoming a more-and-more outdated dialect. This has caught me off-guard a few times.

dirkhh:
since the builds appear to all pass and those include our oldest build environments (I think), why don't we merge this and go straight to C++17 ?

dirkhh:
merged with modified commit message

bstoeger:
Thanks - if someone complains we can still downgrade to C++14.

PR #3032, opened 2020-10-19
[RFC] libdc: free value strings given by libdc's dc_parser_get_field()

bstoeger:
Apparently libdc gives us copies of strings. The API is very
scary, because (at least according to my reading of the code),
the key/value pair may be stored in a cache. Thus on free()ing
the string in the cache becomes invalid and we must not access
it twice. Very obscure.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes what I believe to be a memory hole in the libdc-interface. However, it could be that I'm reading the code totally wrong. Only compile tested, as I have no device on me.
### Changes made:

1) copy serial-number provided by libdc in analogy to firmware
2) free value-strings provided by libdc

### Mentions:

@torvalds, @dirkhh: Please tell me that I'm missing something obvious.

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -564,7 +564,7 @@ static void set_dc_serial(struct divecomputer *dc, const char *serial)
 {
 	const struct device *device;
 
-	dc->serial = serial;
+	dc->serial = strdup(serial);
 	if ((device = get_device_for_dc(&device_table, dc)) != NULL)
 		dc->deviceid = device_get_id(device);
 
@@ -723,6 +723,7 @@ static dc_status_t libdc_header_parser(dc_parser_t *parser, device_data_t *devda
 		if (!str.desc || !str.value)
 			break;
 		parse_string_field(devdata, dive, &str);
+		free((void *)str.value); // libdc gives us copies of the value-string.
 	}
 
 	dc_divemode_t divemode;


torvalds:
You're not missing anything obvious, but the libdivecomputer strings just aren't all that organized. The general rule is that the key is owned by libdivecomputer, and the value is a malloc that can/should be free'd, but as you say, you can ask for the key/value pair multiple times, so ...

dirkhh:
> You're not missing anything obvious, but the libdivecomputer strings just aren't all that organized. The general rule is that the key is owned by libdivecomputer, and the value is a malloc that can/should be free'd, but as you say, you can ask for the key/value pair multiple times, so ...

I think we have only two options:
(a) the value is strdup'ed every time it is passed back and the user (Subsurface) then frees it
(b) the value is owned by libdivecomputer as well (just like the key) and therefore the freeing needs to happen when the parser is destroyed on the libdc side

bstoeger:
> You're not missing anything obvious, but the libdivecomputer strings just aren't all that organized. The general rule is that the key is owned by libdivecomputer, and the value is a malloc that can/should be free'd, but as you say, you can ask for the key/value pair multiple times, so ...

It appears that if you *don't* read out and free the strings, they are leaked? So in practice, the string-interface is not optional at all for the parsers that go via the cache.

I know that with cell-phones having GBs of RAM, this isn't an issue. But still...

> (b) the value is owned by libdivecomputer as well (just like the key) and therefore the freeing needs to happen when the parser is destroyed on the libdc side

The problem with this is that some parsers don't use the cache and therefore this would put additional burden on them. Disclaimer: if I read the code correctly.

dirkhh (approved):
I think we really need to redo the whole memory management of the strings.
But this at least doesn't appear to be wrong

PR #3033, opened 2020-10-20
devices: remove new devices on undo of import/download

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

So far, download / import would directly add to the global device table. This means that on undo of download / import, the devices stayed there. This changes the download / import code to parse into a separate table, which is then imported by the ImportDives undo-command. Thus, direct write-access of global state is reduces (there is still some left - e.g. autogroup).

Warning: I quickly tested import, but download is untested and will probably crash. I will try to do that in the upcoming days.

Note: In the not-so-distant future, I plan to collect all the tables in a `struct divelog`. This should make the whole parser-state thing more palatable.

### Changes made:

1) Add devices to non-global table
2) Make adding of devices undoable

### Release note:

Not sure.
### Documentation change:

No.

### DIFF
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -18,9 +18,10 @@ void addDive(dive *d, bool autogroup, bool newNumber)
 }
 
 void importDives(struct dive_table *dives, struct trip_table *trips, struct dive_site_table *sites,
-		 struct filter_preset_table *presets, int flags, const QString &source)
+		 struct device_table *devices, struct filter_preset_table *presets,
+		 int flags, const QString &source)
 {
-	execute(new ImportDives(dives, trips, sites, presets, flags, source));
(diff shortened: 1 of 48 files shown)


bstoeger:
PS: This could be simplified distinctly, if the "suunto" special case in the downloader would not add a device directly. However, I didn't dare to touch that, since I don't know in which order the callbacks are called.

dirkhh (dismissed):
looks reasonable

dirkhh:
but did you notice that you broke the export_html build?

bstoeger:
> but did you notice that you broke the export_html build?

fixed.

dirkhh:
merged

PR #3038, opened 2020-10-24
test: manually run gitStorageTest

dirkhh:
This way we should see the output and hopefully be able to figure
out why that silly test keeps randomly failing.

- [x] Build system change

I still have no idea why this keeps randomly failing :-(

### Changes made:

simply run the test with `-v2` manually before running `make check`.
This way we should actually see the full output of the test and be able to tell what exactly fails.
Fingers crossed.

### Related issues:

19 out of 20 GitHub Action failures in the past few weeks


dirkhh:
So in both cases running `TestGitStorage` manually passed, but then as part of `make check` it failed.
FFS.

PR #3039, opened 2020-10-24
Shut up some Coverity warnings

dirkhh:

- [x] Code cleanup

mostly in the 'noise' category.

### DIFF
--- a/core/trip.c
+++ b/core/trip.c
@@ -74,10 +74,12 @@ bool is_trip_before_after(const struct dive *dive, bool before)
 {
 	int idx = get_idx_by_uniq_id(dive->id);
 	if (before) {
-		if (idx > 0 && get_dive(idx - 1)->divetrip)
+		const struct dive *d = get_dive(idx - 1);
+		if (d && d->divetrip)
 			return true;
 	} else {
-		if (idx < dive_table.nr - 1 && get_dive(idx + 1)->divetrip)
+		const struct dive *d = get_dive(idx + 1);
+		if (d && d->divetrip)
 			return true;
 	}
 	return false;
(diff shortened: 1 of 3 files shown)


core/trip.c
```diff
@@ -74,10 +74,12 @@ bool is_trip_before_after(const struct dive *dive, bool before)
 {
 	int idx = get_idx_by_uniq_id(dive->id);
 	if (before) {
-		if (idx > 0 && get_dive(idx - 1)->divetrip)
+		struct dive *d = get_dive(idx - 1);
```
bstoeger: You could write this as `const struct dive *`, to make clear that die is not modified.

core/trip.c
```diff
@@ -74,10 +74,12 @@ bool is_trip_before_after(const struct dive *dive, bool before)
...
+		if (d && d->divetrip)
 			return true;
 	} else {
-		if (idx < dive_table.nr - 1 && get_dive(idx + 1)->divetrip)
+		struct dive *d = get_dive(idx + 1);
+		if (d && d->divetrip)
```
bstoeger, line 82: ditto

qt-models/divetripmodel.cpp
```diff
@@ -717,7 +717,7 @@ void DiveTripModelTree::populate()
 	for (int i = 0; i < dive_table.nr; ++i) {
 		dive *d = get_dive(i);
 		update_cylinder_related_info(d);
```
bstoeger, line 721: If `d` were NULL, wouldn't `update_cylinder_related_info()` crash? I.e. shouldn't the null-pointer check be moved in front of this calls? As it is, it makes no sense...

dirkhh: `update_cylinder_related_info()` does its own NULL check...
This way it was less of a code change. But for clarity, year, you're right.

core/trip.c
dirkhh: good! will do

bstoeger (approved):
lgtm

dirkhh:
merging locally

PR #3040, opened 2020-10-25
Add device management to undo system

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This includes the device management in the undo system. The UI is only a stub - for example, it doesn't save column widths. It uses a new tab instead of the old dialog. The reason is that on undo, the user should get some visual feedback. I plan to switch to the corresponding tab in the future. This seems more appropriate than opening a dialog. However, we're starting to get too much tabs and something has to be done about that.

This now makes `mark_divelist_changed` unused on desktop and thus is somewhat of a "psychological milestone".
### Changes made:

1) Include device management in undo system
2) Move `mark_divelist_changed' to mobile, because desktop relies now only on the undo-commands.

### Release note:

Done.

### Documentation change:

Haven't checked.


dirkhh (approved):
A few comments:
- yes, the UI needs to be rethought; in so many ways
- congrats on completing the undo system for desktop :-)
- let me know how I can help with the mobile code to complete this

bstoeger:
> congrats on completing the undo system for desktop :-)

Only those that were annotated with `mark_divelist_changed`. For example setting `autogroup`, which is saved to the log, is not undoable. Admittedly, that's a very minor thing.

> let me know how I can help with the mobile code to complete this

You could look at the mark_divelist_changed()/unsaved_changes() code in qmlmanager.cpp. It appears to me that it has nothing to do with undo and therefore can probably be removed in its current form...?

dirkhh:
merging locally

PR #3044, opened 2020-10-25
More cleanups

dirkhh:

- [x] Code cleanup
- [x] Documentation change

Coverity fixes
Build warnings
Obsolete Grantlee reference

### Mentions:

@bstoeger - would appreciate a critical look ESPECIALLY at the C++ changes

### DIFF
--- a/core/datatrak.c
+++ b/core/datatrak.c
@@ -602,9 +602,11 @@ static int wlog_header_parser (struct memblock *mem)
 	}
 }
 
+#define NOTES_LENGTH 256
+#define SUIT_LENGTH 26
 static void wlog_compl_parser(struct memblock *wl_mem, struct dive *dt_dive, int dcount)
 {
-	int strlong = 256, tmp = 0, offset = 12 + (dcount * 850),
+	int tmp = 0, offset = 12 + (dcount * 850),
 	    pos_weight =  offset + 256,
 	    pos_viz = offset + 258,
 	    pos_tank_init = offset + 266,
@@ -616,9 +618,10 @@ static void wlog_compl_parser(struct memblock *wl_mem, struct dive *dt_dive, int
 	 * Extended notes string. Fixed length 256 bytes. 0 padded if not complete
 	 */
 	if (*(runner + offset)) {
-		wlog_notes = calloc(strlong + 1, 1);
-		wlog_notes = memcpy(wlog_notes, runner + offset, 256);
-		wlog_notes = to_utf8((unsigned char *) wlog_notes);
+		char wlog_notes_temp[NOTES_LENGTH + 1];
+		wlog_notes_temp[NOTES_LENGTH] = 0;
+		(void)memcpy(wlog_notes_temp, runner + offset, NOTES_LENGTH);
+		wlog_notes = to_utf8((unsigned char *) wlog_notes_temp);
 	}
 	if (dt_dive->notes && wlog_notes) {
 		buffer = calloc (strlen(dt_dive->notes) + strlen(wlog_notes) + 1, 1);
@@ -664,11 +667,11 @@ static void wlog_compl_parser(struct memblock *wl_mem, struct dive *dt_dive, int
 	 * Dive suit, fixed length of 26 bytes, zero padded if shorter.
 	 * Expected to be preferred by the user if he did the work of setting it.
 	 */
-	strlong = 26;
 	if (*(runner + pos_suit)) {
-		wlog_suit = calloc(strlong, 1);
-		wlog_suit = memcpy(wlog_suit, runner + pos_suit, strlong);
-		wlog_suit = to_utf8((unsigned char *) wlog_suit);
+		char wlog_suit_temp[SUIT_LENGTH + 1];
+		wlog_suit_temp[SUIT_LENGTH] = 0;
+		(void)memcpy(wlog_suit_temp, runner + pos_suit, SUIT_LENGTH);
+		wlog_suit = to_utf8((unsigned char *) wlog_suit_temp);
 	}
 	if (wlog_suit)
 		dt_dive->suit = copy_string(wlog_suit);
--- a/core/uemis-downloader.c
+++ b/core/uemis-downloader.c
@@ -819,7 +819,7 @@ static void parse_tag(struct dive *dive, char *tag, char *val)
 	} else if (!strcmp(tag, "altitude")) {
 		uemis_get_index(val, &dive->dc.surface_pressure.mbar);
 	} else if (!strcmp(tag, "f32Weight")) {
-		weightsystem_t ws;
+		weightsystem_t ws = empty_weightsystem;
 		uemis_get_weight(val, &ws, dive->dc.diveid);
 		add_cloned_weightsystem(&dive->weightsystems, ws);
 	} else if (!strcmp(tag, "notes")) {
--- a/desktop-widgets/about.h
+++ b/desktop-widgets/about.h
@@ -9,7 +9,7 @@ class SubsurfaceAbout : public QDialog {
 	Q_OBJECT
 
 public:
-	explicit SubsurfaceAbout(QWidget *parent = 0, Qt::WindowFlags f = 0);
+	explicit SubsurfaceAbout(QWidget *parent = 0);
 private
 slots:
 	void on_licenseButton_clicked();
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -17,6 +17,7 @@
 #include <QMessageBox>
 #include <QHeaderView>
 #include "commands/command.h"
+#include "commands/command_base.h"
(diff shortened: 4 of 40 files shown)


core/datatrak.c
```diff
@@ -616,9 +616,10 @@ static void wlog_compl_parser(struct memblock *wl_mem, struct dive *dt_dive, int
 	 * Extended notes string. Fixed length 256 bytes. 0 padded if not complete
 	 */
 	if (*(runner + offset)) {
-		wlog_notes = calloc(strlong + 1, 1);
-		wlog_notes = memcpy(wlog_notes, runner + offset, 256);
-		wlog_notes = to_utf8((unsigned char *) wlog_notes);
+		char *wlog_notes_temp = calloc(strlong + 1, 1);
```
bstoeger: This one is a bit weird. `strlong` appears to be a constant 256. Which means that this could be allocated on the stack. Then in the next line the 256 is hardcoded!

For now this fix is fine, but this deserves some closer look.

core/datatrak.c
```diff
@@ -666,9 +667,10 @@ static void wlog_compl_parser(struct memblock *wl_mem, struct dive *dt_dive, int
 	 */
 	strlong = 26;
 	if (*(runner + pos_suit)) {
-		wlog_suit = calloc(strlong, 1);
-		wlog_suit = memcpy(wlog_suit, runner + pos_suit, strlong);
-		wlog_suit = to_utf8((unsigned char *) wlog_suit);
+		char *wlog_suit_temp = calloc(strlong, 1);
```
bstoeger: As above. In C++ you would simply declare `strlong` as `constexpr`, which guarantees that you can use it as an index in a static array. In C you would probably #define it and then use a static array.

I'm fine with this fix for now, but this really could use some rework.

core/uemis-downloader.c
```diff
@@ -198,6 +198,7 @@ static void uemis_get_weight(char *buffer, weightsystem_t *weight, int diveid)
 		lbs_to_grams(ascii_strtod(buffer, NULL)) :
 		lrint(ascii_strtod(buffer, NULL) * 1000);
 	weight->description = translate("gettextFromC", "unknown");
+	weight->auto_filled = false;
```
bstoeger: What about either
```
    *weight = empty_weightsystem;
```
at the top of this function
or
```
    weightsystem_t ws = empty_weightsystem;
```
where it is generated?
That would avoid future versions of this bug in case more fields are added.

PS: `uemis_get_weight_unit` could be made local.

desktop-widgets/about.h
```diff
@@ -9,7 +9,7 @@ class SubsurfaceAbout : public QDialog {
 	Q_OBJECT
 
 public:
-	explicit SubsurfaceAbout(QWidget *parent = 0, Qt::WindowFlags f = 0);
+	explicit SubsurfaceAbout(QWidget *parent = 0, Qt::WindowFlags f = QFlag(0));
```
bstoeger: Changes in this commit are fine for now. Ultimately, I would remove these default parameters. Because either:
- We want these dialogs be generated with different flags. Then we shouldn't provide default parameters.
- We want these dialogs be generated with the same flags. Then just hardcode them in the constructor.

desktop-widgets/divelistview.cpp
```diff
@@ -563,7 +563,7 @@ void DiveListView::selectionChanged(const QItemSelection &selected, const QItemS
 			select_dives_in_trip(trip);
 			if (model->rowCount(index)) {
 				QItemSelection selection;
-				selection.select(index.child(0, 0), index.child(model->rowCount(index) - 1, 0));
+				selection.select(model->index(0, 0, index), model->index(model->rowCount(index) - 1, 0, index));
```
bstoeger, line 567: Holy-moly, Qt's model/view API is bizarre. FWIW, It looks right.

dirkhh:
excellent review comments. Thank you so much @bstoeger - this is so helpful!

dirkhh:
I forgot from the last time I did that... the new `QVector` range constructor has only been available since Qt 5.14
Groan.

dirkhh:
I think I have made the code a little better, based on your feedback, @bstoeger 
I also added a couple more changes to deal with a few other warts that I noticed while working on this (hopefully addressing the TestGitStorage issue in the process).

desktop-widgets/divelistview.cpp
```diff
@@ -645,7 +645,12 @@ void DiveListView::addDivesToTrip()
 	std::vector<dive *> dives = getDiveSelection();
 	if (!t || dives.empty())
 		return;
-	Command::addDivesToTrip(QVector<dive *>::fromStdVector(dives), t);
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
```
bstoeger: That looks surprisingly ugly.

What would you think about a general function of the kind
```
template<typename T>
QVector<T> stdToQt(const std::vector<T> &v)
{
#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
	return QVector<T>(v.begin(), v.end());
#else
	return QVector<T>::fromStdVector(c);
#endif
}
```
in `command_base.h`?
Totally untested, but something like this should work. We could extend this to other types, if needed.

dirkhh:
loads more changes - I implemented your idea, @bstoeger - crap, forgot to give you credit. Will do that in my local branch but won't confuse all the test builds by another force push.
This was compile tested on Mac for desktop and mobile on both Qt 5.12 and 5.15 - so all the stuff that's version conditional should be correct. As for all the other compiler variations, I'll wait to see what the compile tests here result in.

With this branch we get a 'nearly' warning free build. Still quite a few Coverity issues left to triage, though.

profile-widget/profilewidget2.cpp
```diff
@@ -1822,12 +1822,20 @@ int ProfileWidget2::fixHandlerIndex(DiveHandler *activeHandler)
 	if (index > 0 && index < handles.count() - 1) {
 		DiveHandler *before = handles[index - 1];
 		if (before->pos().x() > activeHandler->pos().x()) {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 13, 0)
+			handles.swapItemsAt(index, index - 1);
```
bstoeger: What about
```
std::swap(handles[index], handles[index - 1]);
```
? That should work for all Qt versions...

dirkhh: won't that mess with Qt's weird ref counting semantics?

profile-widget/profilewidget2.cpp
```diff
@@ -1022,7 +1022,11 @@ void ProfileWidget2::wheelEvent(QWheelEvent *event)
 		scale(1.0 / zoomFactor, 1.0 / zoomFactor);
 		zoomLevel--;
 	}
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
```
bstoeger, line 1025: > Again, replacement was just recently added.

This is surprisingly unfriendly! You would think you wait a few years with deprecating a function after adding the alternative.

profile-widget/profilewidget2.cpp
bstoeger: OMG - I forgot about that particular insanity. Here, it shouldn't matter because when getting the first non-const reference,  the list goes into "detached" state.

profile-widget/profilewidget2.cpp
dirkhh: This is all in preparation for Qt 6.0 - in the past they were far more reasonable with the speed of deprecating APIs.
I expect that it will be quite a while before we can stop supporting Qt 5.12 :-)

profile-widget/profilewidget2.cpp
dirkhh: TBH, I'd be tempted to keep the ugliness that I have here as that doesn't risk mixing std::... and Qt memory management assumptions. Just doesn't seem worth it

bstoeger: std::swap(&a,&b) literally just does
```
_tmp = a;
a = b;
b = _tmp;
```
If _that_ fails, we have more serious problems!

Here is the relevant snippet from qlist.h:
```
inline T &QList<T>::operator[](int i)
{ Q_ASSERT_X(i >= 0 && i < p.size(), "QList<T>::operator[]", "index out of range");
detach(); return reinterpret_cast<Node *>(p.at(i))->t(); }
```
With the first access, the list is copied (if it was shared) and therefore the second access goes to the same list.

This is safe. Do you have any particular reason to suspect that it isn't?

dirkhh: no, just my regular Qt paranoia :-)

dirkhh: ok, implemented that and pushed otherwise unmodified branch

Should I merge this once the CI passes?

bstoeger: Go ahead...

PR #3045, opened 2020-10-25
filter: use 64-bit integer literal for generating bit fields

bstoeger:
For multiple-choice constraints we use a bit field of type
uint64_t. This means we theoretically support up to 64 items.
Currently use at most seven.

Coverity complained (correctly) that we use the expression
"1 << x" to generate the bitfields. However 1 is a 32-bit
literal on most platforms, which makes this undefined
behavior for x >= 32. Change the integer literal to 64-bit
1ULL.

Moreover, when detecting items with an index >= 64, don't
even attempt to set the according bit, since this is
undefined behavior and the compiler is free to do as it
pleases in such a case.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix of a latent bug, which solves a Coverity warning. Not visible now, but could become active if we have multiple-choice constraints with more than 31 entries. 
### Changes made:

1) Use 64-bit integer literals to create bit fields.
2) Don't attempt to create bit fields with invalid indexes.

### DIFF
--- a/core/filterconstraint.cpp
+++ b/core/filterconstraint.cpp
@@ -1059,7 +1059,7 @@ static bool check_year_range(const filter_constraint &c, const struct dive *d)
 
 static bool check_multiple_choice(const filter_constraint &c, int v)
 {
-	bool has_bit = c.data.multiple_choice & (1 << v);
+	bool has_bit = c.data.multiple_choice & (1ULL << v);
 	return has_bit != c.negate;
 }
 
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
nice. thanks

dirkhh:
the two build errors are just the known "random" connectivity issue with `TestGitStorage` that I eventually need to figure out a solution for...

PR #3048, opened 2020-10-27
cleanup: refactor subsurfacesysinfo.cpp

bstoeger:
This used to be a copy of QSysInfo. However, once the requirement
was raised to Qt5.4, this was replaced by a subclass of the original
QSysInfo - which made the whole file mostly obsolete.

Just use QSysInfo directly where needed.

Only for windows.c, which can't call directly into Qt, keep the
isWin7Or8() helper function.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Idle code-hygiene thing: This removes a rather bizarre artifact. There used to be a copy of QSysInfo, which accordingly had full Qt attribution. Since that was removed, the whole thing became somewhat pointless. Just access QSysInfo directly where needed (qthelper.cpp). Reminds me that it is time to split qthelper.cpp into sensible pieces (e.g. string-formatting and other helpers).

I hope I didn't mess up the user string and the windows code!

### Mentions:

@dirkhh

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -8,7 +8,6 @@
 #include "gettextfromc.h"
 #include "statistics.h"
 #include "membuffer.h"
-#include "subsurfacesysinfo.h"
 #include "version.h"
 #include "errorhelper.h"
 #include "planner.h"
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
This looks reasonable... 
I'm as always worried that this will suddenly change things in unexpected ways and as a result break things because we rely on 'the number of colons in such and such string' or something equally obscure.
Been there, done that.
But... no risk, no glory

Issue #3020, opened 2020-10-14
macOS: Offer possibility to toggle between light and dark theme of Subsurface


ghost:
### Describe the issue:

- [ ] Bug
- [x ] Change request
- [ ] New feature request
- [ ] Discussion request

### Operating system:

macOS 10.15.7 on iMac

### Subsurface version:

4.9.7 official release

### Steps to reproduce:

Set macOS to the dark theme, start Subsurface

### Current behavior:

Dark theme in Subsurface does not look good or even consistent with macOS.

### Expected behavior:

More pretty ;-)

### Suggestion:
Offer possibility to toggle between light and dark theme of Subsurface in Preferences. I'm fine with the light them of Subsurface even when using the dark them of macOS. I went back to 4.9.5.

dirkhh:
It appears that none of the developers use dark theme.
I certainly don't use it - to me, personally, the user experience is terrible and causes eye strain across the board. I am old enough to remember why we all switched to light background in our windows in the 1980s...

I'm happy to entertain changes to Subsurface (i.e., someone sending patches) that don't affect it in regular mode and that help people who want to use dark theme, but I don't see anyone working on that.
As for 'turning it off' - that's explicitly discouraged by Apple. I don't even know if it's possible using Qt. We do quite literally NOTHING to adjust Subsurface or detect if the user is using dark theme.

dirkhh:
I think this is significantly improved in the latest builds

PR #3053, opened 2020-10-29
Simple Mac OSX Dark Mode Fix

djunkins:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Small change to the Desktop tab widgets to fix two minor UI issues: 1) medium blue labels on the Dive Information tab were difficult to read in Mac OS X dark mode, and 2) the new device names tab was missing a label for the dive computer table which caused a default label of "GroupBox" to be displayed.

Medium blue labels in the Dive Information tab were inconsistent with the other tabs.

### Changes made:

1) removed the CSS_SET_HEADING_BLUE macro from TabDiveInformation.cpp
2) removed calls to the setStyleSheet(CSS_SET_HEADING_BLUE) method for the dive information labels
3) modified the CSSSetSmallLabel string to remove the 'color: mediumblue;' style directive for QLabels
4) added a "title" property to the table widget in TabDiveComputer.ui

### Related issues:

Does not fully address issue #3020 since it does not allow a user to toggle between dark and light modes


djunkins:
Please don't merge the pull request yet -- just realized that the edit icon on the dive sites tab needs to be changed also.

dirkhh:
That iOS build error was just bizarre. Let me restart that and see if it goes away.

dirkhh (commented):
@willemferguson will be sad to see his blue text gone... 
But that's clearly the better choice over all. Is there a way to tell if we are in light or dark mode and to change colors depending on that? What happens with a running app when the user changes from light to dark or back?

dirkhh:
> Please don't merge the pull request yet -- just realized that the edit icon on the dive sites tab needs to be changed also.

No problem. I really appreciate you working on this!!!

djunkins:
@dirkhh From what I found, the Qt signal that is supposed to be sent when the mode changes from light to dark is broken, so it does not seem to be possible to explicitly set colors when the mode changes. Removing the explicit color definition was a simple fix, but I will continue to see if there is another way to handle this.

dirkhh:
> @dirkhh From what I found, the Qt signal that is supposed to be sent when the mode changes from light to dark is broken, so it does not seem to be possible to explicitly set colors when the mode changes. Removing the explicit color definition was a simple fix, but I will continue to see if there is another way to handle this.

Do you have a link the the QT_BUG for that?

djunkins:
> Do you have a link the the QT_BUG for that?

No, I have not found the QT_BUG yet. I saw a reference to the issue on another site when I was trying to figure out the best way to address the issue. I will continue to research whether it is still an issue and whether we can catch the signal when the mode changes so we can optimize the style for each.

djunkins:
Please review now. I think I've addressed the most pressing issues with dark mode (the labels on the information tab being unreadable and the edit icon on the dive sites tab being all black.)

Ultimately it will make sense to have two different palettes defined for light and dark mode, but that will take a lot more work including figuring out how to catch the signal when the mode changes.

dirkhh (approved):
I'll build locally and test in light and dark mode.
Once I've done that, I'll merge

dirkhh:
OK, I still don't understand why someone would want to use dark mode... but that's just me being old.
It's really hard on my eyes - almost all of the web pages I go to are black on white. Qt Creator is black on white. Any pdf that I look at is black on white. So my eyes constantly have to switch between positive and negative which is hard on them.
So right now the occasional web page designed by a cool person that is white on black is odd to me - but everything else on my screen looks consistent.
That childish (or old-person) rant out of the way, your PR creates a significant improvement in dark mode. And Subsurface correctly switches between dark and light while running. So "well done".
I notice a couple more icons that aren't perfect, but I'll merge this as is and will leave you the picture below in case you want to do a separate PR to address those
Screen Shot 2020-10-29 at 2 56 24 PM

djunkins:
I'll take a look and see if I can do further improvements on the icons. It also looks like some buttons change background color, but not the font color in dark mode -- they're legible, but not ideal. This PR should help with the worst issues with dark mode, though.

PR #3054, opened 2020-10-29
Updates to new statistics tab

willemferguson:
This PR contains the latest changes I made to the source code, mostly in QML. Because this is saved as a new branch it also contains the changes made for the previous PR. The stats tab show to graphics: a histogram and a scatterplot.
1) To see the histogram click on Create Graph. This is a graph of no.-of-dives against depth.
2) To see the scatterplot, select SAC from the top combobox and click Create Graph. This shows SAC plotted against depth.
At the moment I will not be doing much more until we are ina position to test the models of @bstoeger . Then we will get an impression of how further integration of qml and C++ might proceed.
Its a given that there will be problems in the build tests of this code. It builds perfectly on my system. I might try to trace at least some of these problems and perhaps do some corrections if these are reasonably easy to do.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @bstoeger


dirkhh:
so this has long been superseded by #3052 - so let's close this

PR #3056, opened 2020-11-01
core: on import remove merged dives from trip/divesite

bstoeger:
When dives were merged on import, they were not unregistered
from their dive site and trip before being deleted. Thus, these
tables had stale pointers, which would ultimate lead to crashes.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a crash on dive import. The code is incredibly fragile - we have to do something about this. And we also have to improve our testing of this part of the code! 
### Changes made:

1) Remove merged dives from trip and divesite tables before deleting them.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -857,6 +857,10 @@ static void merge_imported_dives(struct dive_table *table)
 			merged->dive_site = NULL;
 			add_dive_to_dive_site(merged, ds);
 		}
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
that looks correct.
Arg. We should both have caught that.

PR #3058, opened 2020-11-01
parser: don't crash when parsing <weight> tags [Alternative 1]

bstoeger:
When encountering a <weight> tag, we would parse into the last
weightsystem. However, we only create weightsystems when
encountering <weightsystem> tag. Therefore, this code would
either crash or overwrite the previous weightsystem.

Instead, create a new weightsystem for each <weight> tag.

Moreover, make sure that inside a <weightsystem> tag a
weightsystem actually exists. This should be the case,
but who knows...?

Reported-by: Nihal Gabr <gabr.nihal@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a crash on parsing XMLs with <weight> tags - see commit description.

### DIFF
--- a/core/parse-xml.c
+++ b/core/parse-xml.c
@@ -1223,7 +1223,10 @@ static void try_to_fill_dive(struct dive *dive, const char *name, char *buf, str
 {
 	char *hash = NULL;
 	cylinder_t *cyl = dive->cylinders.nr > 0 ? get_cylinder(dive, dive->cylinders.nr - 1) : NULL;
+	weightsystem_t *ws = dive->weightsystems.nr > 0 ?
+		&dive->weightsystems.weightsystems[dive->weightsystems.nr - 1] : NULL;
 	pressure_t p;
+	weight_t w;
 	start_match("dive", name, buf);
 
 	switch (state->import_source) {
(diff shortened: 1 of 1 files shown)


dirkhh (commented):
Since we have no other user for the `get_or_create...` function, this seems a reasonable way to do this. But it's not consistent with what we do for cylinders.

dirkhh:
Actually, after going back to the emails, I think I like #3059 better.

bstoeger:
I'm not sure if you realize the difference between both alternatives:
This one adds multiple weights if there are multiple `<weight>` tags - the other one sums all weight tags into a single value.

I think the correct way depends on how the `<weight>` tag is used. Since I'm not familiar with the various XML formats, it's impossible for me to tell.

dirkhh:
You can just export one of your dives with a weightsystem in XML and see it:

`<weightsystem weight='3.0 kg' description='belt' />`

The `weight` is only ever inside a `weightsystem`, and there's only supposed to be one such tag.

bstoeger:
But that's not the issue here. That case is handled by
```
                if (MATCH_STATE("weight.weightsystem", weight, &ws->weight))
                        return;
```
We're talking about top-level `<weight>` tags in XMLs that are _not_ produced by Subsurface!

torvalds:
Ok, having looked at both, I think this one is the right one.

It has some iOS build issue, but I think that's unrelated.

dirkhh:
I'll work on the iOS issue today let's merge or now

PR #3059, opened 2020-11-01
parser: don't crash when parsing <weight> tags [Alternative 2]

bstoeger:
When encountering a <weight> tag, we would parse into the last
weightsystem. However, we only create weightsystems when
encountering <weightsystem> tag. Therefore, this code would
either crash or overwrite the previous weightsystem.

Instead, add the weight to the first weightsystem. Create that
if it doesn't exist using the new get_or_create_weightsystem()
function, which works like its cylinder-analogue.

Moreover, make sure that inside a <weightsystem> tag a
weightsystem actually exists. This should be the case,
but who knows...?

Reported-by: Nihal Gabr <gabr.nihal@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a crash on parsing XMLs with <weight> tags - see commit description.

### DIFF
--- a/core/parse-xml.c
+++ b/core/parse-xml.c
@@ -1223,7 +1223,10 @@ static void try_to_fill_dive(struct dive *dive, const char *name, char *buf, str
 {
 	char *hash = NULL;
 	cylinder_t *cyl = dive->cylinders.nr > 0 ? get_cylinder(dive, dive->cylinders.nr - 1) : NULL;
+	weightsystem_t *ws = dive->weightsystems.nr > 0 ?
+		&dive->weightsystems.weightsystems[dive->weightsystems.nr - 1] : NULL;
 	pressure_t p;
+	weight_t w;
 	start_match("dive", name, buf);
 
 	switch (state->import_source) {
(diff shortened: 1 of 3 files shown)


core/parse-xml.c
```diff
@@ -1326,12 +1329,17 @@ static void try_to_fill_dive(struct dive *dive, const char *name, char *buf, str
...
-	if (MATCH_STATE("weight", weight, &dive->weightsystems.weightsystems[dive->weightsystems.nr - 1].weight))
+	if (ws) {
+		if (MATCH("description.weightsystem", utf8_string, &ws->description))
+			return;
+		if (MATCH_STATE("weight.weightsystem", weight, &ws->weight))
+			return;
+	}
+	if (MATCH_STATE("weight", weight, &w)) {
+		weightsystem_t *ws = get_or_create_weightsystem(dive, 0);
+		ws->weight.grams += w.grams;
```
dirkhh, line 1340: why are we adding to the weight here, instead of setting it?

dirkhh (commented):
This seems very much the same as the cylinder case, which appeals to me for its consistency

torvalds:
Let's do the other fix. 

This is really only an issue with odd external xml, but I agree with @bstoeger tghar the simpler fix to just always add a weightsystem when you see a weight (without a weightsystem tag) seems to be the right thing after having given it some thought.

PR #3060, opened 2020-11-01
Manipulate text colour to accomodate dark mode displays

willemferguson:
The Preferences has an option to set the colour of the text on the
Information tab to either MediumBlue, LightBlue or Black. The last two of these
colours are meant to enable areadable font contrast on displays with dark mode.
The choice is saved with the other preferences.

Important. Subsurface needs to be restarted for the change to take effect. This
is because the text colour is part of the initialisation of the TabDiveInformation
object. To redo text colour on the fly would require destructing the object and
creating a new TabDiveInformation object.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @bstoeger

### DIFF
--- a/core/pref.h
+++ b/core/pref.h
@@ -66,6 +66,12 @@ enum unit_system_values {
 	PERSONALIZE
 };
 
+enum darkmode_colour_values {
+	MEDIUMBLUE,
+	LIGHTBLUE,
+	BLACK
+};
+
 // ********** PREFERENCES **********
 // This struct is kept global for all of ssrf
 // most of the fields are loaded from git as
@@ -105,6 +111,7 @@ struct preferences {
 	double      font_size;
 	double      mobile_scale;
 	bool        show_developer;
+	enum darkmode_colour_values darkmode_colour;
 
 	// ********** Equipment tab *******
 	const char *default_cylinder;
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -55,6 +55,7 @@ void qPrefDisplay::loadSync(bool doSync)
 	disk_mobile_scale(doSync);
 	disk_display_invalid_dives(doSync);
 	disk_show_developer(doSync);
+	disk_darkmode_colour(doSync);
 	if (!doSync) {
 		load_tooltip_position();
 		load_theme();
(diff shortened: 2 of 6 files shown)


dirkhh:
So this to me makes very limited sense. I'm ok with wanting different colors, but then I would suggest to make it consistent across the app. Not just in one of the tabs

willemferguson:
The same issue with headings is going to come up with the planned statistics tab. I would personally like blue headings in the current statistics tab if you think it would be useful for me to reorganise that tab and come up with a proposal. The point is that this issue currently only concerns the information tab that benefits from coloured headings because of the many items presented and where a distinction between headings and data promotes instant readability. My feeling is that, yes, one should move towards a consistent presentation of the UI, but let's start in one place and when it works well in the info tab one could expand the same principle across other parts of the UI. I hope this sounds halfway logical.

dirkhh:
If we want to follow accepted behavior on macOS, we need to either have a fixed set of colors that work in all modes, or we need to successfully react to the light/dark change signal (which @djunkins found out isn't really working at this point). macOS will (stupidly) change from light to dark when it's "evening" (unless you tell it to stop that nonsense) and the user experience in that case is really poor otherwise.

A setting that requires a restart in order to change the font color is really not a good user experience, either. Why wouldn't we be able to redraw the widget when the user changes the font color?

willemferguson:
I got the code to work in real time, no problem.That is the short answer. Here is the long answer I hope you have time to read. The problem was (and still is) communication between core/settings/qPrefDisplay and desktop-widgets/tab-widgets/TabdiveInformationTab. If this communication cannot happen, every time an updatedata() is executed, a whole lot of time-consuming CSS commands and setup is executed: highly inefficient because mostly there is no update of text colour to do (maybe once in a million executions). I fixed this up in an extremely old-fashioned way. Have a copy of prefs.darkmode_colour defined in class TabDiveInformation. The constructor saves the prefs.darkmode_colour value. Lets say a user changes the font colour, bringing about  a change in prefs.darkmode_colour in the prefs structure. Now, every time updatedata() executes, it checks whether pref.darkmode_colour equals the stored value in TabDiveInformation. If not, the text colours are updated with all the CSS overhead. Now, this could be done much more efficiently if one could use the colour_changed signal generated by qPrefDisplay and connect() that with a socket in TabDiveInformation. But the 2 objects I want to connect here are at opposite sides of the application and I have no idea how to do the connect(). Do you or @bstoeger have any suggestion about connecting such distant objects? I will not even start griping about having to bough to and accommodate the bugs in Apple and Microsoft and which become our problems.

dirkhh:
> Here is the long answer I hope you have time to read. 

I'm always interested in the real / long answers :-)

> The problem was (and still is) communication between core/settings/qPrefDisplay and desktop-widgets/tab-widgets/TabdiveInformationTab. If this communication cannot happen, every time an updatedata() is executed, a whole lot of time-consuming CSS commands and setup is executed: highly inefficient because mostly there is no update of text colour to do (maybe once in a million executions). I fixed this up in an extremely old-fashioned way. Have a copy of prefs.darkmode_colour defined in class TabDiveInformation. The constructor saves the prefs.darkmode_colour value. Lets say a user changes the font colour, bringing about a change in prefs.darkmode_colour in the prefs structure. Now, every time updatedata() executes, it checks whether pref.darkmode_colour equals the stored value in TabDiveInformation. If not, the text colours are updated with all the CSS overhead. Now, this could be done much more efficiently if one could use the colour_changed signal generated by qPrefDisplay and connect() that with a socket in TabDiveInformation. But the 2 objects I want to connect here are at opposite sides of the application and I have no idea how to do the connect(). Do you or @bstoeger have any suggestion about connecting such distant objects? I will not even start griping about having to bough to and accommodate the bugs in Apple and Microsoft and which become our problems.

Story of our lives. Most of the code we have in place deals with bugs and problems in Qt, Linux, Windows, or macOS. And they are all equally problematic, just each in different areas.
I'll take a look at this today - I should have time over lunch.

dirkhh:
ok, started to look at things... we have decided a long time ago to go with US English as the default spelling, including for our variables and things. So let's not do `colour`, please.
also, quite a few whitespace issues.
the setting of colors needs to happen in a separate function that's a slot and that can be connected to the settings signal.
I'll try to clean all of this up and see if I can make it work.

willemferguson:
I have no issues with this. If I can be of help, you can leave decided-upon points e.g. US terms, to me to rectify in this PR.

dirkhh:
OK, we'll bring this in via #3063 - I'll close this one

PR #3061, opened 2020-11-01
build-systsem: hack around Xcode 12 issue

dirkhh:
It appears that Xcode 12 applies some rather self defeating logic when picking
build architectures in release builds for the simulator. It adds aarch64 by
default and I can't find a way to turn that off from the command line. At the
same time, you can't link against the simulator if you have build with aarch64
as the aarch64 simulator doesn't exist, yet.

Some people claim that building in debug mode fixes this. Let's see if that
works.

- [x] Build system change

simply try to build in debug mode for iOS

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -9,8 +9,10 @@ on:
 
 jobs:
   mobileBuild:
-    runs-on: macOS-latest
+    runs-on: macOS-10.15
     steps:
+    - name: switch to Xcode 11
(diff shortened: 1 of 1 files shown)


dirkhh:
The openSUSE issue is an upstream problem in their repo, it never even starts the install of development libraries in the container, so unrelated to our code. And the Ubuntu thing is just a GitTestStorage failure, not even worth restarting
So I'll merge this

Issue #3057, opened 2020-11-01
Segfault on unset DCs


salvadorcunat:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I'm having a segfault (shortened bt follows) coming from get_device_for_dc(), which is present at current master b2b3544f3:

#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff569d537 in __GI_abort () at abort.c:79
#2  0x00007ffff5a357ec in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff5a40966 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff5a409d1 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff5a40c65 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff5a37faa in std::__throw_logic_error(char const*) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x0000555555816aaa in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*> (
    __end=0xffffffffffffffff <error: Cannot access memory at address 0xffffffffffffffff>, __beg=0x0, this=0x7fffffffd740) at /usr/include/c++/10/bits/basic_string.tcc:206
#8  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct_aux<char const*> (
    __end=0xffffffffffffffff <error: Cannot access memory at address 0xffffffffffffffff>, __beg=0x0, this=0x7fffffffd740) at /usr/include/c++/10/bits/basic_string.h:247
#9  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*> (
    __end=0xffffffffffffffff <error: Cannot access memory at address 0xffffffffffffffff>, __beg=0x0, this=0x7fffffffd740) at /usr/include/c++/10/bits/basic_string.h:266
#10 std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string<std::allocator<char> > (__a=..., __s=0x0, this=0x7fffffffd740)
[...]

PR #3062, opened 2020-11-01
smtk-import: Add support for new devices model

salvadorcunat:
Include devices Ids in the DC data.
Ensure we always set a DC model string for manual dives or unsupported devices.

Signed-off-by: Salvador Cuñat <salvador.cunat@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

Related to #3057

### DIFF
--- a/smtk-import/smartrak.c
+++ b/smtk-import/smartrak.c
@@ -35,6 +35,7 @@
 #include "core/divesite.h"
 #include "core/membuffer.h"
 #include "core/tag.h"
+#include "core/device.h"
 
 /* SmartTrak version, constant for every single file */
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
this looks reasonable.
the openSUSE build failure is actually an upstream issue where their repos are currently out of sync and not installable - so that one is safe to ignore

dirkhh:
the iOS failure is also a known issue, one that hopefully I'll be able to fix in master in a few minutes

PR #3067, opened 2020-11-04
automatically pick a good title color for light and dark mode

dirkhh:

- [x] Functional change

This is a much better replacement for #3063 -- which really shouldn't have been merged :-(

### Changes made:

rip out the preferences to pick this one color. that was just so user unfriendly
instead figure out if we are in a light or dark color and pick a reasonable title color.
and notice if the theme changes and update this choice if necessary

### Related issues:

#3060 #3063 #3053 

### Mentions:

@djunkins @bstoeger @willemferguson

### DIFF
--- a/core/pref.h
+++ b/core/pref.h
@@ -66,12 +66,6 @@ enum unit_system_values {
 	PERSONALIZE
 };
 
-enum headerstyle_color_values {
-	MEDIUMBLUE,
-	LIGHTBLUE,
-	BLACK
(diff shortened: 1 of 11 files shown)


dirkhh:
This seems to do the right thing on macOS as well as on my openSUSE installation. My Ubuntu VM unfortunately appears messed up, there dark mode doesn't work at all, so I can't blame this PR for that

PR #3069, opened 2020-11-07
Fixes concerning column-widths

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes column widths of various tables.

### Changes made:

1) Maximize last column in filter preset table and make the trash column small.
2) Save columns of device-table using a sensible name.
3) Actually save columns of device-table and dive-site-table.

### DIFF
--- a/desktop-widgets/filterwidget.cpp
+++ b/desktop-widgets/filterwidget.cpp
@@ -30,6 +30,8 @@ FilterWidget::FilterWidget(QWidget* parent) :
 	ui.presetTable->setModel(FilterPresetModel::instance());
 	ui.presetTable->setSelectionBehavior(QAbstractItemView::SelectRows);
 	ui.presetTable->setSelectionMode(QAbstractItemView::SingleSelection);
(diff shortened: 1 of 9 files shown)


dirkhh (approved):
I love the editorial comments in the commit message.
I wonder what Qt experts would say to all this.
@thiagomacieira - is this something that has come up before?

thiagomacieira:
Yes.

The fact that children are destroyed after the parent has begun but not finished destruction is a known issue. C++ has no two-stage destruction, so this is something that happens. We have some workarounds in pace such as making QWidget destroy the children so they are still run before the QWidget ceases being a QWidget, but there's no way to avoid the general issue.

The order is something I'd never seen discussed before. The order is not order of construction, but the order of parenting. Moreover, that order can change since in a QWidget, the order of the children is also the Tab order. Destroying them in reverse order is a reasonable request, but I just don't know it will have much effect.

Using direct objects so they are destroyed before the containing object itself begins destruction is a frequent solution.

PR #3070, opened 2020-11-07
filter: don't escape '&' in filter-preset menu

bstoeger:
When constructing an action, '&' is used as the keyboard shortcut
marker. Since this mangles preset names, use the setIconText()
function of the action instead.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes the bug reported on the ML. I totally forgot about that menu - maybe we should remove it?

### Changes made:

1) Don't escape ampersands in filter-preset menu.


dirkhh (approved):
This appears to be better than my solution.

PR #3071, opened 2020-11-07
desktop/filter: show '&' in dropdown for loading saved filters

dirkhh:
Apparently this is mis-interpreted as a shortcut key unless the ampersand is
doubled.  The documentation is frustratingly quiet about this.

- [x] Bug fix

### Release note:

maybe?

### Mentions:

@bstoeger


bstoeger:
This might be problematic: https://bugreports.qt.io/browse/QTBUG-63361
However, I don't know if we even support such old Qt-versions?

dirkhh:
we support 5.9, but the mac binaries are build with 5.15 (for 10.15/11.0) and 5.13 for macOS 10.13/14

PR #3074, opened 2020-11-08
Fix translations

### LINKED ISSUE #3068: Translation of undo command when editing cylinders in equipment tab is not working
### Describe the issue:

- [ x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When adding/removing/editing cylinders in equipment tab the translation of the undo command (Edit in main menu) doesn't work as expected. The word "Undo" gets translated but not the rest of the string which describes the command. For example:
"Undo Add cylinder" doesn't translate **Add cylinder**" 
"Undo Edit cylinder" doesn't translate **Edit cylinder**" 
Tested on several different languages with the same result. So all non English languages are affected.

### Operating system:

Verified on Linux and MacOS

### Subsurface version:

Using latest development build

### Steps to reproduce:

- Change to a non English language
- Restart Subsurface
- Add a new dive
- Go to Equipment tab
- Add a new cylinder 
- Verify result under **Edit** in the main menu

### Current behavior:

Part of string get translated

### Expected behavior:

Translation of entire string

dirkhh:

- [x] Bug fix

I hate doing this so close to a release, but this makes the translations actually work for me, and they clearly didn't work before, so this seems like a net win :-)

### Changes made:

As @bstoeger suggested, simply use the `Command::Base::tr()` function everywhere. And then, of course, make the corresponding adjustments to the translations themselves.

### Related issues:

Fixes #3068

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1030,9 +1030,9 @@ AddCylinder::AddCylinder(bool currentDiveOnly) :
 	if (dives.empty())
 		return;
 	else if (dives.size() == 1)
-		setText(tr("Add cylinder"));
+		setText(Command::Base::tr("Add cylinder"));
 	else
-		setText(tr("Add cylinder (%n dive(s))", "", dives.size()));
+		setText(Command::Base::tr("Add cylinder (%n dive(s))", "", dives.size()));
 	cyl = create_new_cylinder(dives[0]);
 }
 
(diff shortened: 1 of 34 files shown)


dirkhh:
There's a chance that pulling translations from Transifex might undo the changes, I'll need to pay attention to that right before making the release

PR #3075, opened 2020-11-11
filter: avoid Windows crash

dirkhh:
The scope confusion between s (the for loop variable) and s (the function
argument) caused a crash in the s.split() on Windows.

- [x] Bug fix

interesting to see this work on all platforms except Windows

### Release notes
added

### Mentions:

@bstoeger

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Windows: avoid crash when setting filter values for tag-style fields
 - mobile: add location service warning as required by Google Play
 - mobile: fix manually adding dives in the past [#2971]
 
--- a/core/filterconstraint.cpp
+++ b/core/filterconstraint.cpp
@@ -686,8 +686,8 @@ void filter_constraint_set_stringlist(filter_constraint &c, const QString &s)
 		return;
 	}
 	c.data.string_list->clear();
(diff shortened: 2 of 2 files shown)


dirkhh:
I guess this is worth a 4.9.9 release :-(
But I'd love to see your take on it (of course, it's 23:00 in Austria...)

bstoeger:
If that fixes it ... good. However, unless I'm missing something, this looks like a frightening mis-compilation, no?

bstoeger (approved):

dirkhh:
> If that fixes it ... good. However, unless I'm missing something, this looks like a frightening mis-compilation, no?

We both know that I know very little about standards and stuff. From a scope perspective the old code looks unfortunate.
And my guess is that the compiler did indeed something stupid here. MXE is gcc based, I just realize that the two other platforms I tested on prior to the fix (Mac and my ArchLinux) both use clang

bstoeger:
We depend on the compiler producing correct code, or if it can't at least emit a fat warning. I get the correct behavior with gcc: https://godbolt.org/z/KffrKo
Does windows use an obscure version of gcc?

dirkhh:
I compile all of our bits from source, so this is MXE. I'll need to go through the logs but I'm pretty sure that this is gcc 8.
(MXE provides multiple version of gcc)

PR #3078, opened 2020-11-12
desktop: update completion models if dive was edited

### LINKED ISSUE #3077: Completer lists don't get refilled at runtime
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

This was reported on Scubaboard, I managed to reproduce it locally.
Imagine a dive log with three buddies Alice, Bob, and Charlie, then typing A, B, or C is enough to get them to pop up as autocompletion suggestions.
Now add a dive with David.
When you add a second dive, you would expect D to pop up an autocompletion for David. But that doesn't happen unless you quit and restart the app
 
### Operating system:

Tested on macOS, report was on Windows

### Subsurface version:

4.9.8

### Mentions:

@bstoeger is this something that looks like a trivial fix? I'll release 4.9.9 today in order to get the Windows crash fixed. So if this one of those "duh, I know what causes that" moments, I'd appreciate a quick PR

bstoeger:
In the main-tab, when changing tag, buddy or divemaster,
update the corresponding completion model.
This is a quick-fix and the wrong thing to do. It works only
if the currently shown dive is changed, which is not a given.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Update auto-complete models, quickfix. See commit description.
### Related issues:

Fixes #3077.

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -261,12 +261,18 @@ void MainTab::divesChanged(const QVector<dive *> &dives, DiveField field)
 	}
 	if (field.divesite)
 		updateDiveSite(current_dive);
-	if (field.tags)
+	if (field.tags) {
+		tagModel.updateModel(); // TODO: Don't do this here
 		ui.tagWidget->setText(get_taglist_string(current_dive->tag_list));
-	if (field.buddy)
+	}
+	if (field.buddy) {
+		buddyModel.updateModel(); // TODO: Don't do this here
 		ui.buddy->setText(current_dive->buddy);
-	if (field.divemaster)
+	}
+	if (field.divemaster) {
+		diveMasterModel.updateModel(); // TODO: Don't do this here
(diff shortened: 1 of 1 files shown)


dirkhh:
the test failures were caused by my dead server - restarted

dirkhh (commented):
would it make more sense to update the models in the commands when we actually edit a buddy / tag / divemaster ?

dirkhh:
once again the GitStorage tests are failing, but also it seems that GitHug Action runs are taking for bleeping ever.
The Cloud storage is up and running and I continue to not understand why the tests are failing

bstoeger:
> would it make more sense to update the models in the commands when we actually edit a buddy / tag / divemaster ?

Yes, but not as a "quickfix".

dirkhh:
merged locally
I gave up and turned of TestGitStorage again. I just don't have the mental ability to deal with that noise

PR #3086, opened 2020-11-17
core: fix incorrect QString::asprintf usage

dirkhh:
This is a static function, it cannot be used as a method on an object to
construct that object.

This caused us to not print GPS coordinates in the UI.

- [x] Bug fix

that's what happens when I 'fix' warnings and don't carefully test the resulting code because it's "obviously right". :-(

### Release note:

done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- core: fix missing GPS coordinate strings in the UI
 - core: Gracefully handle infinte MND for oxygen
 - mobile: add location service warning as required by Google Play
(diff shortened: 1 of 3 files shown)


bstoeger (approved):
LGTM. The instance in downloadfromdcthread.cpp should also be changed to `QString::vasprintf()` though. It is a static function and thus does not operate on an object. Therefore, the line is misleading.

dirkhh:
> LGTM. The instance in downloadfromdcthread.cpp should also be changed to `QString::vasprintf()` though. It is a static function and thus does not operate on an object. Therefore, the line is misleading.

agreed.

PR #3088, opened 2020-11-18
[WIP] Android build system redo - switch to Qt 5.15, latest NDK, create Android Bundle

dirkhh:

- [x] Functional change
- [x] Build system change

Our existing way to build Android apks stopped working with Qt 5.14, so all our Android binaries were still built against Qt 5.13.
In order to switch to a new Qt version a lot has to change in the way we build Subsurface-mobile for Android

I spent way too much time trying to make this work with `cmake` but in the end gave up and switched back to `qmake`, just as we had already done for iOS.

I submitted an AAB  (Android Application Bundle - the strongly encouraged new format to submit apps) to Google Play for closed testing to see if this actually results in a working binary. In my local testing it seemed to work fine, but that's unfortunately not the same because locally one always installs APKs

### Changes made:

- upgrade key components, including SDK, NDK, OpenSSL, Qt
- switch to `qmake` based build
- complete rewrite of dependency building
- still needs more work to create a new build container (no point in creating that until we know that the way we build actually works)

### Related issues:

As I create the new build container, that will include QtCharts support which will be needed for #3052 

### Additional information:

Having two different `Subsurface-mobile.pro` files is silly, so this needs to be consolidated with the iOS build.
Also, this update showed that we really need to upgrade our Kirigami version (which is quite painful as well)

### Release note:

missing

### Documentation change:

likely not needed

### Mentions:

@janmulder @glance-


dirkhh:
oh, the Android CI build is of course bogus as that is not going to use the new setup... I need to have a working android build container before that makes any sense to upgrade. What I most care about with the CI right now is that this didn't break any OTHER builds.

dirkhh:
As expected. A painful waste of time. Oh how I love this part of what I do...
Of course, this still isn't working, still isn't ready for prime time.
But at least we are slowly inching closer.

dirkhh:
merged locally

PR #3092, opened 2020-11-25
Some desktop code reshuffling and cleanups

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are small steps towards better encapsulation and less inter-dependencies. It is mostly moving code around, notably out of the MainWindow. More to come.

### Changes made:

1) Create a global settings-changed signal and connect in the appropriate classes.
2) Simplify the preferences-widget code.
3) Move planner-code from MainWindow to diveplanner.cpp. Collect the three widgets in a new class.
4) Move DiveHandler to profilewidget folder. If we want to have a planner on mobile, this must be generally accessible.
5) Some dead code removal.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -679,15 +679,7 @@ void DiveListView::mergeTripBelow()
 
 void DiveListView::removeFromTrip()
 {
-	//TODO: move this to C-code.
-	int i;
-	struct dive *d;
-	QVector<dive *> divesToRemove;
-	for_each_dive (i, d) {
-		if (d->selected && d->divetrip)
-			divesToRemove.append(d);
-	}
-	Command::removeDivesFromTrip(divesToRemove);
+	Command::removeDivesFromTrip(stdToQt(getDiveSelection()));
 }
 
 void DiveListView::newTripAbove()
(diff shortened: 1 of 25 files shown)


desktop-widgets/diveplanner.cpp
```diff
@@ -19,6 +19,11 @@
 #include <QMessageBox>
 #include <QSettings>
 #include <QShortcut>
+#ifndef NO_PRINTING
```
dirkhh, line 19: at some point, someone will have enough free time to decide it is worth undoing this abomination.
`#ifndef NO_PRINTING` really, really, REALLY should be `#if PRINTING`
🤦🏼‍♂️

dirkhh (approved):
All these cleanups make me very happy.
The more we untangle things, the better in the long run.
I have gone through the changes, but not line by line... it's possible that some unintended side effects slipped through. But conceptually they are the right thing to do, and the code that I looked at in greater detail looked right.

atdotde (approved):
I browsed through the diff and all looks good. I have not checked, though, if all settingsChanged combinations got properly moved. If not, we will hopefully notice.

desktop-widgets/divelistview.cpp
```diff
@@ -679,30 +679,15 @@ void DiveListView::mergeTripBelow()
 
 void DiveListView::removeFromTrip()
 {
-	//TODO: move this to C-code.
-	int i;
-	struct dive *d;
-	QVector<dive *> divesToRemove;
-	for_each_dive (i, d) {
-		if (d->selected && d->divetrip)
```
bstoeger, line 687: Attention: the `d->divetrip` condition is lost and must now be moved to the undo-command.

dirkhh: So, I don't think you implemented that on the undo-command side, or did I miss that in my review?
Which means we shouldn't merge, yet, correct?

bstoeger: Done.

dirkhh (approved):
OK, did another pass through the commits. This looks good. I think it's ready to be merged.

PR #3093, opened 2020-11-26
cleanup: remove unused function in templatelayout.cpp

bstoeger:
This was removed with grantlee.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Trivial cleanup: removal of unused function.
### Mentions:

@atdotde: please check whether that is really not needed anymore or the call was inadvertently removed.

### DIFF
--- a/desktop-widgets/templatelayout.cpp
+++ b/desktop-widgets/templatelayout.cpp
@@ -100,31 +100,6 @@ TemplateLayout::TemplateLayout(print_options *printOptions, template_options *te
 	this->templateOptions = templateOptions;
 }
 
-/* a HTML pre-processor stage. acts like a compatibility layer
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
Since we no longer use Grantlee... I can see how that could have become obsolete?

atdotde (approved):
Yes this can go. The reason is that none of our templates used the whole list of cylinders or weights or cylinders/weights with numbers, so the grantee replacement only implements looping over dives or cylinders (so far, looping over weights is not implemented as, once more, our templates don't use that. But it could be implemented parallel to cylinders).

PR #3096, opened 2020-12-03
mobile updates, including newer Kirigami

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Next step in my attempts to build Android with Qt 5.15 and current Kirigami
This hasn't been tested on iOS **at all**
Works reasonably well on Android (that's what the build 3038 beta is based on)
Works ok for me on Linux

### Changes made:

- switch to Kirigami 5.76
- change the way we build Kirigami and how we include it in our source tree
- update our Kirigami hacks to be relative to 5.76 and maintained as commits

### Mentions:

@janmulder in case you want to take a look at this fun progression


janmulder:
To be sure, this is the version that was pushed to Google Play today, right?

Its a bit off-topic here, but I've been trying lately to build the Android version myself again using the new docker image with Qt 5.15 but that mysteriously fails on my Arch Linux desktop at the very final gradle step to build the aab:

* What went wrong:
Could not determine the dependencies of task ':compileDebugJavaWithJavac'.
java.nio.file.FileSystemException: /subsurface/../proc/self/task/65921/root/proc/65364/task/65364/net: Invalid argument

Something familiar to you @dirkhh?

janmulder:
And did a quick test of the 4.9.10.26 (the one that came from Google Play this morning).

1) the top menu bar does not have the < and > navigation arrows any more
2) the top menu bar is not themed any more (color)
3) tried download over BT and BLE from OSTC Plus => all working. 
4) there is something weird when changing portrait to landscape. The divelist stays on the same width (but only with the start of the app in portrait mode, when started in landscape mode, all seems to work wrt. resizing).

Overall, it seems a proper step forward. I still remember the pain a new Kirigami causes.

dirkhh:
> To be sure, this is the version that was pushed to Google Play today, right?
> 
> Its a bit off-topic here, but I've been trying lately to build the Android version myself again using the new docker image with Qt 5.15 but that mysteriously fails on my Arch Linux desktop at the very final gradle step to build the aab:
> 
> * What went wrong:
>   Could not determine the dependencies of task ':compileDebugJavaWithJavac'.
>   java.nio.file.FileSystemException: /subsurface/../proc/self/task/65921/root/proc/65364/task/65364/net: Invalid argument
> 
> Something familiar to you @dirkhh?

That is not an error I have ever seen. Ugh.
And gradle is typically hard to debug (at least for me) because the error messages are often extremely misleading.

Full disclosure: while the container is basically equivalent to a local VM that I have, in order to make my debugging / turnaround experience better, I have been building on that VM, not with the container, lately. But I'll switch back to it very soon.

dirkhh:
> And did a quick test of the 4.9.10.26 (the one that came from Google Play this morning).

Thank you, much appreciated!
 
> 1. the top menu bar does not have the < and > navigation arrows any more

Yes, they were a victim of my attempts to get things back to working state. Right now I simply disable them.
I need to re-apply my hacks to Kirigami to make them work.

> 2. the top menu bar is not themed any more (color)

I noticed that after I pushed it out. I believe that's another Kirigami hack that needs to be moved to the new tooling.

> 3. tried download over BT and BLE from OSTC Plus => all working.

Yay.

> 4. there is something weird when changing portrait to landscape. The divelist stays on the same width (but only with the start of the app in portrait mode, when started in landscape mode, all seems to work wrt. resizing).

We do some very, very ugly magic hacking there at the start to deal with a number of devices where we get spurious resize notifications and were getting the screen size totally wrong (you may remember that from a couple of years ago). As a result we now have some very fragile logic that tries to figure out whether it should resize the column width or whether it should leave it alone. It sounds like this is getting tripped up for you.
 
> Overall, it seems a proper step forward. I still remember the pain a new Kirigami causes.

Yes, it was not a nice experience. Again. But we had to move to Qt 5.15 in order to be able to build .aab - and for that I needed to upgrade Kirigami :-(

dirkhh:
I love this game. Android and Linux happily build here locally. Now I need to track down the subtle difference between what I have in my VMs and what is there when things are run on GitHub Actions.

dirkhh:
Well... iOS is still broken and I'm mostly at a loss how to fix this. I have tried so many things, it's not even funny.
But the rest of them seem to be reasonably functional at this point.

dirkhh:
Hey @janmulder - I have now checked the instructions in `packaging/android/README` in this branch and started from a fresh Ubuntu 20.04 install with both the docker image based build and the local build system build. Both complete successfully, creating apk's that you could install on your device locally.
I'd love to understand why this fails for you.

dirkhh:
rebased on master

dirkhh:
a quick test on iOS shows that all the icons are missing, but hey, baby steps in the right direction!

dirkhh:
OK, some cleanup and streamlining of the commits to remove a bit of the noise. Still could benefit from more house keeping, but of course there is always the issue with diminishing returns.
First let's see if this actually finally passes all the build tests.
Then fix the icons on iOS.
Then worry about the sequence of commits being pleasing.

dirkhh:
This is now lightly tested on iOS and the icons are back as well.
I will try to find some time to clean up the history a bit, but I think we are getting closer to being able to merge this.
Would love review / feedback - but understand that people are strapped for time and otherwise distracted
@janmulder @glance- @tcanabrava @mturkia

dirkhh:
For those following along... I am now tracking down UI issues that I can identify on
1) iOS
2) Android
3) mobile on desktop

You will be thrilled to learn that the behavior is NOT consistent between the three (of course not) and that there are especially differing visual issues/glitches between them. Including things that I have seen and fixed before and that somehow magically have come back.
This is So. Much. Fun.

dirkhh:
I'm pretty sure no one is actually looking at this branch :-)
But at least now we have the same behavior across Android/iOS/macOS/Linux again.
(based on my admittedly not super exhaustive testing).
The history is much more to my liking.
So at this point I think I'm ready to merge this.
I'll give the Europeans (most all of you) time to look at this during your day tomorrow, but unless I hear some comments, I'll merge this...

mturkia:
Skimming the changes through on a phone didn’t produce anything to complain about. But then again, I am not familiar with kirigami...

dirkhh:
> Skimming the changes through on a phone didn’t produce anything to complain about. But then again, I am not familiar with kirigami...

Thanks for taking a look. I really appreciate this. I think I'll merge this monster later this morning.

PR #3100, opened 2020-12-11
Add surge etc to printing template variables

atdotde:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


core/subsurface-qt/diveobjecthelper.h
```diff
@@ -15,6 +15,10 @@ class DiveObjectHelper {
 	Q_PROPERTY(int id MEMBER id CONSTANT)
 	Q_PROPERTY(int rating MEMBER rating CONSTANT)
 	Q_PROPERTY(int visibility MEMBER visibility CONSTANT)
+	Q_PROPERTY(int wavesize MEMBER wavesize CONSTANT)
```
bstoeger, line 18: Now that grantlee is gone do we actually need that DiveObjectHelper thing?

bstoeger:
Looks correct and compiles - let's merge.

bstoeger:
Looking at the code, I'm not 100% sure, but the loop thing appears fishy to me:

Suppose that you have a list without elements (for example, we now support dives with no cylinders). Since the loop is not entered a single time, the "pos" variable will not be increased. This should output one empty loop content and then exit early when it reaches the end for, isn't it?

I'll try to fix this.

core/subsurface-qt/diveobjecthelper.h
atdotde: True, we don't a lot of the bells and whistles and if I built it from scratch I wouldn't go this way. But we still need to access all the stuff in the dive list via string names (as they come from the template), so I don't think we can get much leaner here.

dirkhh:
I love the order of operation. Merged, then question.
I do that all the time :-)

In case you are wondering "what's Dirk doing these days..." --- still driving himself to tears with the Kirigami / Qt update.
@tcanabrava is now trying to help me.

atdotde:
Yay! Welcome back @tcanabrava !

Qt update? Are you an early adopter of Qt 6.0?

PR #3102, opened 2020-12-12
Printing code cleanups

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Now that grantlee is gone, we can slowly turn the printing code into an acceptable state. Warning: only lightly tested!

### Changes made:

1) Bike shedding: replace pointer by reference.
2) Replace the weird QVariant map by a strongly typed state structure. This was bizzare: a member variable was stored and read from a QVariant map.
3) Fixes a bug in the loop code: This would break for empty loops. Makes this code easier to reason about and formats an if-block only if it is needed.
4) Rename "dive master" in the default templates to "dive guide".

### Release note:

No - code hygiene and obscure bug.

### Mentions:

@atdotde @dirkhh: see dive guide change.


dirkhh (commented):
The second commit message contains a typo.
I read through the code, but I barely understood the old code and really don't feel like I have a handle on whether this is the right thing to do. Nothing stands out as wrong, but...
I'd love to see @atdotde 's comments

dirkhh:
@atdotde any chance you'll have a moment to look at this PR?

bstoeger:
This now contains a more radical culling of the DiveObjectHelper code, which was never a really good idea and became obsolete with the introduction of the mobile models and removal of grantlee. Complex objects simply shouldn't live in QML context - only data with value semantics (IMHO).

atdotde:
I thought I had commented on this before. Will do now.

desktop-widgets/templatelayout.cpp
```diff
@@ -286,35 +284,47 @@ static QRegularExpression forloop(R"(\s*(\w+)\s+in\s+(\w+))");	// Look for "VAR
...
-		++i;
-		parser(tokenList, pos, out, state);
+		state.forloopiterator = i++;
+		parser(tokenList, from, to, out, state);
 	}
 	act = old;
 	state.forloopiterator = olditerator;
 }
 
-void TemplateLayout::parser(QList<token> tokenList, int &pos, QTextStream &out, State &state)
```
atdotde: Implementing this in this  more procedural look ahead way rather than the functional one I had implemented is a matter of personal preference I would say. Also the functional approach could produce better error messages by testing for this in the handling of endfor/endif etc. But this definitely works, so fine with me.

atdotde (approved):
As far as printing is concerned this is great (maybe with a reservation about style re parsing of loops/conditionals but no reservation re functionality). 

I am not qualified to comment on possible consequences for mobile/QML  from removing DiveObjectHelper.

desktop-widgets/templatelayout.cpp
bstoeger: Note that it wasn't only bad diagnostics, but - if I read it correctly - the old code was buggy for empty lists - which can happen for cylinders. If the loop body was not executed once, `pos` was not increased accordingly, isn't it?

Not trying to get into pointless semantic discussions, but the old code was *not* functional, and that was why it was hard to reason about. The whole advantage (some may argue disadvantage) of functional programming is that functions have no side effects. For the same input a functions *must* give the same output and it *must never* change the input parameters. Therefore, the line that you quote is precisely an example of what functional programming is not. For another example, `i = i + 1` makes no sense in functional programming.

That said, the correct way would of course be constructing a parse-tree, not a list of tokens. However, for now that's probably not worth it.

dirkhh:
> I am not qualified to comment on possible consequences for mobile/QML from removing DiveObjectHelper.

I believe the QML code no longer uses `DiveObjectHelper` (it's surprisingly hard to be SURE of that, but I'm pretty sure).
Oddly, there is a bit of use of `DiveObjectHelper` left in `qmlmanager.cpp` and all of it seems bogus and stupid and wrong.
I'll need to go back to a much older version to see how we got here - but from what I can tell this should all just be ripped out.
Given that my 5+ week nightmare of messing with Qt, QML, and Kirigami is coming to a conclusion when I merge the **newKirigami** branch today, that is the perfect starting point to dig into `QMLManager` and figure out ways to make it less horrible... 😆

bstoeger:
Spare yourself the trouble.

The reason DOH was there is quite logical: In the old code, a DOH was generated from the `dive *` pointer that was passed through the model (a horrible idea to begin with). This DOH was used to display the dive data.

On saving, to figure out which fields changed, the code generated a new DOH to check user-input and old data, which kind of makes sense.

Then, we replaced the DOH in the display layer by the new mobile-models, which directly produced the appropriate strings. They used the same formatting functions as DOH, so things continued to work.

This PR now replaces DOH by a direct call to the formatting functions and thus makes DOH useless. However, as stated in the commit message: the correct fix would be to re-query the model to prevent these two parts of the code ever getting out-of-sync.

PS: Obviously, don't merge yet, since Android and iOS don't compile. For some reason I am too stupid to correctly add the new string-format.cpp to the qmake files.

dirkhh:
> Spare yourself the trouble.

:-)

> The reason DOH was there is quite logical: In the old code, a DOH was generated from the `dive *` pointer that was passed through the model (a horrible idea to begin with). This DOH was used to display the dive data.
> 
> On saving, to figure out which fields changed, the code generated a new DOH to check user-input and old data, which kind of makes sense.

Yes, I went back to a much older version of the code and read through it and now understand.

> Then, we replaced the DOH in the display layer by the new mobile-models, which directly produced the appropriate strings. They used the same formatting functions as DOH, so things continued to work.
> 
> This PR now replaces DOH by a direct call to the formatting functions and thus makes DOH useless. However, as stated in the commit message: the correct fix would be to re-query the model to prevent these two parts of the code ever getting out-of-sync.

I haven't gotten to reviewing the new code. Later this morning, I promise
 
> PS: Obviously, don't merge yet, since Android and iOS don't compile. For some reason I am too stupid to correctly add the new string-format.cpp to the qmake files.

:-)
Let's see if this gets worse once I merge newKirigami (among other things, that moves the `Subsurface-mobile.pro` file...

dirkhh (approved):
Always enjoyable to review your PRs.
Snarky commit messages, editorials on programming styles, interesting factoids on the inner workings of various language constructs. There's something for everyone.
I have now convinced myself that this is reasonable and sane and I haven't found anything that appears wrong (or at least no worse than it was).
This will require a rebase on top of newKirigami. Let me know if you want me to deal with that.

dirkhh:
The build errors - I could fix those for you if you want :-)

bstoeger:
Thanks, that would be very welcome!

dirkhh:
merged via #3106

PR #3106, opened 2020-12-17
Print

dirkhh:
simply to run the CI - this is just a rebase of #3102 with the `.pro` file fixed.


dirkhh:
Ok, this seems to have dealt with the CI issues, so merging this rebase and mildly edited (`string-format.cpp` instead of `string-format.c`) version of #3102

PR #3107, opened 2020-12-17
profile: remove clearHandlers function (fixes crash)

bstoeger:
Recently (674c20227b2), the call to ProfileWidget::clearHandlers()
was moved from PlannerWidgets::replanDive() to ProfileWidget2.
This cause a crash, because the code assumes that the number
of elements in the handles-vector the divepointplanner model
is the same.

Clearing the handles violates this assumption. It turns out
that the clearHandlers() function is broken anyway: it clear
the handles-vector, but not the gases-vector, which should
likewise have the same number of elements. It appears that
the clearHandlers() function is an artifact and it is
mysterious how this has worked so far. Remove.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a crash in current master, which can be triggered by entering the planner twice.

This part of the code really needs a refactor.
### Changes made:

1) Remove the `clearHandlers` function, as it appears to have bit-rotted.

### Release note:

No - crash recently introduced.

### Mentions:

@dirkhh @atdotde

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1318,17 +1318,6 @@ void ProfileWidget2::setProfileState()
 }
 
 #ifndef SUBSURFACE_MOBILE
-void ProfileWidget2::clearHandlers()
-{
-	if (handles.count()) {
-		foreach (DiveHandler *handle, handles) {
-			scene()->removeItem(handle);
-			delete handle;
-		}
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
makes sense

PR #3108, opened 2020-12-19
Small profile-picture cleanups / fixes

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a very silly bug, where pictures were deleted twice. It looks like the code handled this gracefully, but still it looks scary.

### Changes made:

1) Improve debug messages: add \n.
2) Don't delete pictures twice.

### Release note:

Bug probably not severe enough.
### Documentation change:

No.
### Mentions:

@dirkhh

### DIFF
--- a/commands/command_pictures.cpp
+++ b/commands/command_pictures.cpp
@@ -119,7 +119,7 @@ static std::vector<PictureListForDeletion> addPictures(std::vector<PictureListFo
 	return res;
 }
 
-RemovePictures::RemovePictures(const std::vector<PictureListForDeletion> &pictures) : picturesToRemove(pictures)
+RemovePictures::RemovePictures(const std::vector<PictureListForDeletion> &pictures)
 {
 	// Filter out the pictures that don't actually exist. In principle this shouldn't be necessary.
 	// Nevertheless, let's play it save. This has the additional benefit of sorting the pictures
--- a/commands/command_pictures.cpp
+++ b/commands/command_pictures.cpp
@@ -24,7 +24,7 @@ void SetPictureOffset::redo()
 {
 	picture *pic = dive_get_picture(d, filename);
 	if (!pic) {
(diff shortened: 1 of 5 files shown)


bstoeger:
Added a commit that pushes `displayed_dive` from a helper class to the profile, with the ultimate goal of removing global state from the profile.

commands/command_pictures.cpp
```diff
@@ -24,7 +24,7 @@ void SetPictureOffset::redo()
 {
 	picture *pic = dive_get_picture(d, filename);
 	if (!pic) {
-		fprintf(stderr, "SetPictureOffset::redo(): picture disappeared!");
+		fprintf(stderr, "SetPictureOffset::redo(): picture disappeared!\n");
```
dirkhh, line 27: my understanding was that `stderr` wasn't line buffered. This doesn't appear to be part of any standard, that seems to say it is implementation dependent, but I saw several references to this as "the convention" and lots of code appears to assume that this is indeed the case.
All that said, the new line makes this much more readable.

profile-widget/divepixmapitem.cpp
```diff
@@ -5,13 +5,9 @@
 #include "core/qthelper.h"
 #include "core/settings/qPrefDisplay.h"
 #include "core/subsurface-qt/divelistnotifier.h"
-#ifndef SUBSURFACE_MOBILE
-#include "core/dive.h" // for displayed_dive
-#include "commands/command.h"
-#endif
```
dirkhh, line 11: typo in the commit message... "which know**S** which dive it represents"

dirkhh (approved):
This looks good, a couple small comments.
Feel free to merge after fixing the commit message.

bstoeger:
Interesting thing about stderr vs. stdout. I guess that makes sense. Changed the commit messages - will merge.

PR #3109, opened 2020-12-19
downloader: Fix typo in libdivecomputer include

henrik242:
Signed-off-by: Henrik Brautaset Aronsen <subsurface@henrik.synth.no>

PR #3110, opened 2020-12-19
Downloader ci

dirkhh:

- [x] Build system change

simply add the downloader to `build.sh` and then call that in one of the CI builds

### Related issues:

#3109 points out that the CI didn't test the downloader builds

### Additional information:

doing this I noticed that https://github.com/subsurface/subsurface/blob/master/packaging/headless/HOWTO might need a rewrite and that maybe we should rename that subdirectory of `packaging` to `downloader` for consistency?
 
### Release note:

N/A

### Documentation change:

needed - see above

### Mentions:

@atdotde would you be able to update the documentation in `packaging/headless`?

### DIFF
--- a/.github/workflows/linux-tumbleweed-qt-latest.yml
+++ b/.github/workflows/linux-tumbleweed-qt-latest.yml
@@ -35,11 +35,14 @@ jobs:
     - name: build Subsurface
       run: |
         echo "--------------------------------------------------------------"
-        echo "building desktop"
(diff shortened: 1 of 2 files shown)


dirkhh:
Ahhh, I forgot that when enabling a mobile build on GitHub, we need to add the git user/email. Oops.

dirkhh:
OK, so this seems to have worked, let's merge it

PR #3111, opened 2020-12-19
Mobile UI

dirkhh:

- [x] Functional change

My favorite topics, combined. Dark theme! QML! Kirigami!
Found several interesting "features" in Kirigami that make all of this much harder.
Admittedly, we don't do theming the way Kirigami wants it done AT ALL
But then, several things about the way Kirigami wants it done are odd (an app can't theme independent from the OS)
And a few things that are used in one part of Kirigami aren't implemented at all in other parts of Kirigami

### Related issues:

every single complaint that the mobile app has a broken dark theme. There were MANY of those

### Release note:

I guess - I should add that, but with the odd way I release the mobile app updates, this is of limited value

### Documentation change:

This implements something that we claimed to have for ages - so... no

### Mentions:

@janmulder @glance- 
I know, no one has time. Me neither.


dirkhh:
Doing code review on your own PR is odd, but I did find a few things that needed changing, so new version pushed.

PR #3112, opened 2020-12-20
[RFC] Remove unused/redundant code in profile

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The control flow in the profile code is a complete mess. For example, in my testing, every call to plotDive() replots the profile items thrice(!). The code tries to be smart and fails horribly. Fixing this will take a long time and I start with a rather simple thing: There seems to be obsolete calls to deco-recalculation, since the deco-info is recalculated for every call to plotDive() anyway. This removes on of those cases. In my tests, this all appeared to work properly, but of course I might have missed something. Please check it out.


bstoeger:
Argh. I messed up the history here. To be redone.

PR #3113, opened 2020-12-21
Fix mobile profile redraw

dirkhh:

- [x] Bug fix
- [x] Code cleanup

If the user edits a dive in a way that changes the profile (things like duration, depth, or gas mix), the mobile app actually didn't redraw the profile until the dive view got invalidated (so scroll forward a couple of dives and then scroll back). That's because the way we tried to do that was completely bogus, fundamentally misunderstanding how delegates work. I'm kinda surprised that this didn't get us at least a QML warning

### Changes made:

- clean up the way we set the dive id for the profile: the existing code kinda worked but also was odd
- update the rendering of the profile if the underlying dive was changed

### Release note:

missing - again, mobile changes are very poorly tracked in that file

### Mentions:

@bstoeger the second commit is one where I'd love to get your feedback. I think I am connecting things correctly (and in my testing this works fine), but this code is really strange. I tried to check if any of the relevant fields were changed, but couldn't figure out how to check the fields argument for a specific set of bits.

### DIFF
--- a/mobile-widgets/qml/DiveDetails.qml
+++ b/mobile-widgets/qml/DiveDetails.qml
@@ -411,6 +411,7 @@ Kirigami.Page {
 				DiveDetailsView {
 					id: diveDetails
 					width: internalScrollView.width
+					myId: model.id
 				}
 				ScrollBar.vertical: ScrollBar { }
 			}
--- a/mobile-widgets/qml/DiveDetailsEdit.qml
+++ b/mobile-widgets/qml/DiveDetailsEdit.qml
@@ -115,9 +115,6 @@ Item {
 				      endpressure, usedGas, usedCyl,
 				      detailsEdit.rating,
 				      detailsEdit.visibility, state)
-		// trigger the profile to be redrawn
-		QMLProfile.diveId = dive_id
-
 		Qt.inputMethod.hide()
 		// now make sure we directly show the saved dive (this may be a new dive, or it may have moved)
 		clearDetailsEdit()
--- a/mobile-widgets/qml/DiveDetailsView.qml
+++ b/mobile-widgets/qml/DiveDetailsView.qml
@@ -12,6 +12,7 @@ Item {
 	property real col1Width: gridWidth * 0.40
 	property real col2Width: gridWidth * 0.30
 	property real col3Width: gridWidth * 0.30
+	property int myId: -1
 
 	width: diveDetailsPage.width - diveDetailsPage.leftPadding - diveDetailsPage.rightPadding
 	height: divePlate.implicitHeight + bottomLayout.implicitHeight + Kirigami.Units.iconSizes.large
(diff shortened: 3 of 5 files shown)


bstoeger (approved):
Fine with me. Though I can't comment on the details. Is this called for every dive or only for visible dives?

In the long run I'd like to get rid of dive->id, but for now it's fine.

You don't need the Q_UNUSED, simply omit the argument name.

Personally, I would use a more functional style ("std::find_if(...)"), but that doesn't have any real advantage. I would just recommend to add a `return` in the if-block.

For testing the fields see e.g. `MainTab::divesChanged`:

```
        // If duration or depth changed, the profile needs to be replotted
        if (field.duration || field.depth)
                MainWindow::instance()->graphics->plotDive(current_dive, true);
```

dirkhh:
> Fine with me. Though I can't comment on the details. Is this called for every dive or only for visible dives?
> 
> In the long run I'd like to get rid of dive->id, but for now it's fine.

That would kill the mobile app.

> You don't need the Q_UNUSED, simply omit the argument name.

OK

> Personally, I would use a more functional style ("std::find_if(...)"), but that doesn't have any real advantage. I would just recommend to add a `return` in the if-block.

Yeah, that's just not a construct that is part of my active language skills. I can look at it and convince myself that it might be doing what is needed in a certain situation, but, yeah, no.

I think a `return` in the `if`-block might be ok - but I need to convince myself that there aren't any operations that can affect multiple dives in the mobile app. I can't think of any right now, but I learned that it's usually these situations where I implement subtle bugs that I then spend days hunting later.
  
> For testing the fields see e.g. `MainTab::divesChanged`:
> 
> ```
>         // If duration or depth changed, the profile needs to be replotted
>         if (field.duration || field.depth)
>                 MainWindow::instance()->graphics->plotDive(current_dive, true);
> ```

Ah, I tried to `||`-combine the flags and compare with that value and that fails.
But I realize that there isn't a field/flag for changing of gases, yet the current code works correctly when I change the gas mix, so I don't think I'll implement such a change right now.

dirkhh:
> Fine with me. Though I can't comment on the details. Is this called for every dive or only for visible dives?

Oh, forgot to comment on this one. We typically have up to three dives either visible or cached off screen. The current dive, and then one dive on either side so that the slide-scroll thing works. As far as I can tell that means that there will be at most three `QMLProfile` objects instantiated that have this handler registered.
But remember, we are talking about QML/C++ code, so as far as I know there are gremlins somewhere doing weird shit.

dirkhh:
> I think a `return` in the `if`-block might be ok - but I need to convince myself that there aren't any operations that can affect multiple dives in the mobile app. I can't think of any right now, but I learned that it's usually these situations where I implement subtle bugs that I then spend days hunting later.

That's a remarkably stupid comment, even for me. I blame the lack of coffee.
Yeah, of course there should be a return. Because if we have more than one dive with the same id we are completely screwed.

PR #3122, opened 2021-01-01
Statistics: Implement statistic variables

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The first in a number of PRs for reviewing the statistics module. This implements the concept of "dive variable" that can be plotted as dependent or independent variable.

Yes, the code is template heavy and at time convoluted. It looks worse than it is, because it doesn't use any TMP (template meta programming) only old-fashioned templates in the sense of type-checked macros. I'm not very proud of it and I'm sure it can be made more space efficient with runtime-polymorphism. However, it allowed me to rather quickly implement this, without too much planning. It has proven to be quite easily expandable.

If any of the constructs are unclear (I suspect there are a few of them), feel free to ask.

### Changes made:

1) A few preparatory commits (the SVG thing is not necessary, but let's bring it in now).
2) Add the statsvariable.cpp source file and compile it.

### Mentions:

@dirkhh

### DIFF
--- a/core/gas.c
+++ b/core/gas.c
@@ -22,7 +22,7 @@ bool isobaric_counterdiffusion(struct gasmix oldgasmix, struct gasmix newgasmix,
 	return get_he(oldgasmix) > 0 && results->dN2 > 0 && results->dHe < 0 && get_he(oldgasmix) && results->dN2 > 0 && 5 * results->dN2 > -results->dHe;
 }
 
-static bool gasmix_is_invalid(struct gasmix mix)
+bool gasmix_is_invalid(struct gasmix mix)
 {
 	return mix.o2.permille < 0;
 }
--- a/core/gas.c
+++ b/core/gas.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "gas.h"
 #include "pref.h"
+#include "gettext.h"
 #include <stdio.h>
 #include <string.h>
 
@@ -138,4 +139,33 @@ void fill_pressures(struct gas_pressures *pressures, const double amb_pressure, 
 			pressures->n2 = get_n2(mix) / 1000.0 * amb_pressure;
 		}
 	}
+}
+
+enum gastype gasmix_to_type(struct gasmix mix)
+{
+	if (gasmix_is_air(mix))
+		return GASTYPE_AIR;
+	if (mix.o2.permille >= 980)
+		return GASTYPE_OXYGEN;
+	if (mix.he.permille == 0)
+		return mix.o2.permille >= 230 ? GASTYPE_NITROX : GASTYPE_AIR;
+	if (mix.o2.permille <= 180)
+		return GASTYPE_HYPOXIC_TRIMIX;
+	return mix.o2.permille <= 230 ? GASTYPE_NORMOXIC_TRIMIX : GASTYPE_HYPEROXIC_TRIMIX;
+}
+
+static const char *gastype_names[] = {
+	QT_TRANSLATE_NOOP("gettextFromC", "Air"),
+	QT_TRANSLATE_NOOP("gettextFromC", "Nitrox"),
+	QT_TRANSLATE_NOOP("gettextFromC", "Hypoxic Trimix"),
+	QT_TRANSLATE_NOOP("gettextFromC", "Normoxic Trimix"),
+	QT_TRANSLATE_NOOP("gettextFromC", "Hyperoxic Trimix"),
+	QT_TRANSLATE_NOOP("gettextFromC", "Oxygen")
+};
+
+const char *gastype_name(enum gastype type)
+{
+	if (type < 0 || type >= GASTYPE_COUNT)
+		return "";
+	return translate("gettextFromC", gastype_names[type]);
 }
--- a/core/gas.h
+++ b/core/gas.h
@@ -69,6 +69,7 @@ extern fraction_t get_gas_component_fraction(struct gasmix mix, enum gas_compone
 extern void fill_pressures(struct gas_pressures *pressures, double amb_pressure, struct gasmix mix, double po2, enum divemode_t dctype);
 
 extern bool gasmix_is_air(struct gasmix gasmix);
+extern bool gasmix_is_invalid(struct gasmix mix);
 extern enum gastype gasmix_to_type(struct gasmix mix);
 extern const char *gastype_name(enum gastype type);
 
--- a/core/gas.h
+++ b/core/gas.h
@@ -22,6 +22,16 @@ struct gasmix {
 static const struct gasmix gasmix_invalid = { { -1 }, { -1 } };
 static const struct gasmix gasmix_air = { { 0 }, { 0 } };
 
+enum gastype {
+	GASTYPE_AIR,
+	GASTYPE_NITROX,
+	GASTYPE_HYPOXIC_TRIMIX,
+	GASTYPE_NORMOXIC_TRIMIX,
+	GASTYPE_HYPEROXIC_TRIMIX,
+	GASTYPE_OXYGEN,
+	GASTYPE_COUNT
+};
+
 struct icd_data { // This structure provides communication between function isobaric_counterdiffusion() and the calling software.
 	int dN2;      // The change in fraction (permille) of nitrogen during the change
 	int dHe;      // The change in fraction (permille) of helium during the change
@@ -59,6 +69,8 @@ extern fraction_t get_gas_component_fraction(struct gasmix mix, enum gas_compone
 extern void fill_pressures(struct gas_pressures *pressures, double amb_pressure, struct gasmix mix, double po2, enum divemode_t dctype);
 
 extern bool gasmix_is_air(struct gasmix gasmix);
+extern enum gastype gasmix_to_type(struct gasmix mix);
+extern const char *gastype_name(enum gastype type);
 
 #ifdef __cplusplus
 }
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -153,11 +153,11 @@ Thumbnailer::Thumbnail Thumbnailer::getHashedImage(const QString &filename, bool
 	return thumbnail;
 }
 
-Thumbnailer::Thumbnailer() : failImage(renderSVGIcon(":filter-close", maxThumbnailSize())), // TODO: Don't misuse filter close icon
-			     dummyImage(renderSVGIcon(":camera-icon", maxThumbnailSize())),
-			     videoImage(renderSVGIcon(":video-icon", maxThumbnailSize())),
+Thumbnailer::Thumbnailer() : failImage(renderSVGIcon(":filter-close", maxThumbnailSize(), false)), // TODO: Don't misuse filter close icon
+			     dummyImage(renderSVGIcon(":camera-icon", maxThumbnailSize(), false)),
+			     videoImage(renderSVGIcon(":video-icon", maxThumbnailSize(), false)),
 			     videoOverlayImage(renderSVGIconWidth(":video-overlay", maxThumbnailSize())),
-			     unknownImage(renderSVGIcon(":unknown-icon", maxThumbnailSize()))
+			     unknownImage(renderSVGIcon(":unknown-icon", maxThumbnailSize(), false))
 {
 	// Currently, we only process one image at a time. Stefan Fuchs reported problems when
 	// calculating multiple thumbnails at once and this hopefully helps.
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -11,7 +11,6 @@
 #include <QString>
 #include <QImageReader>
 #include <QDataStream>
-#include <QSvgRenderer>
 #include <QPainter>
 
 #include <QtConcurrent>
@@ -154,33 +153,11 @@ Thumbnailer::Thumbnail Thumbnailer::getHashedImage(const QString &filename, bool
 	return thumbnail;
 }
 
-static QImage renderIcon(const char *id, int size)
-{
-	QImage res(size, size, QImage::Format_RGB32);
-	res.fill(Qt::white);
-	QSvgRenderer svg{QString(id)};
-	QPainter painter(&res);
-	svg.render(&painter);
-	return res;
-}
-
-// As renderIcon, but render to a fixed width and scale height accordingly
-// and have a transparent background.
-static QImage renderIconWidth(const char *id, int size)
-{
-	QSvgRenderer svg{QString(id)};
-	QSize svgSize = svg.defaultSize();
-	QImage res(size, size * svgSize.height() / svgSize.width(), QImage::Format_ARGB32);
-	QPainter painter(&res);
-	svg.render(&painter);
-	return res;
-}
-
-Thumbnailer::Thumbnailer() : failImage(renderIcon(":filter-close", maxThumbnailSize())), // TODO: Don't misuse filter close icon
-			     dummyImage(renderIcon(":camera-icon", maxThumbnailSize())),
-			     videoImage(renderIcon(":video-icon", maxThumbnailSize())),
-			     videoOverlayImage(renderIconWidth(":video-overlay", maxThumbnailSize())),
-			     unknownImage(renderIcon(":unknown-icon", maxThumbnailSize()))
+Thumbnailer::Thumbnailer() : failImage(renderSVGIcon(":filter-close", maxThumbnailSize())), // TODO: Don't misuse filter close icon
+			     dummyImage(renderSVGIcon(":camera-icon", maxThumbnailSize())),
+			     videoImage(renderSVGIcon(":video-icon", maxThumbnailSize())),
+			     videoOverlayImage(renderSVGIconWidth(":video-overlay", maxThumbnailSize())),
+			     unknownImage(renderSVGIcon(":unknown-icon", maxThumbnailSize()))
 {
 	// Currently, we only process one image at a time. Stefan Fuchs reported problems when
 	// calculating multiple thumbnails at once and this hopefully helps.
--- a/stats/legend.cpp
+++ b/stats/legend.cpp
@@ -1,0 +1,125 @@
+// SPDX-License-Identifier: GPL-2.0
+#include "legend.h"
+#include "statscolors.h"
+#include <QFontMetrics>
+#include <QGraphicsSceneMouseEvent>
+#include <QGraphicsWidget>
+#include <QPen>
+
+static const double legendBorderSize = 2.0;
+static const double legendBoxBorderSize = 1.0;
+static const double legendBoxScale = 0.8;		// 1.0: text-height of the used font
+static const double legendInternalBorderSize = 2.0;
+static const QColor legendColor(0x00, 0x8e, 0xcc, 192); // Note: fourth argument is opacity
+static const QColor legendBorderColor(Qt::black);
+
+Legend::Legend(QGraphicsWidget *chart, const std::vector<QString> &names) :
+	QGraphicsRectItem(chart), chart(chart), displayedItems(0), width(0.0), height(0.0)
+{
+	setZValue(30.0); // On top of everything else.
+	entries.reserve(names.size());
+	int idx = 0;
+	for (const QString &name: names)
+		entries.emplace_back(name, idx++, (int)names.size(), this);
+
+	// Calculate the height and width of the elements
+	if (!entries.empty()) {
+		QFontMetrics fm(entries[0].text->font());
+		fontHeight = fm.height();
+		for (Entry &e: entries)
+			e.width = fontHeight + 2.0 * legendBoxBorderSize +
+				  fm.size(Qt::TextSingleLine, e.text->text()).width();
+	} else {
+		fontHeight = 0.0;
+	}
+	setPen(QPen(legendBorderColor, legendBorderSize));
+	setBrush(QBrush(legendColor));
+
+	resize(); // Draw initial legend
+}
+
+Legend::Entry::Entry(const QString &name, int idx, int numBins, QGraphicsItem *parent) :
+	rect(new QGraphicsRectItem(parent)),
+	text(new QGraphicsSimpleTextItem(name, parent))
+{
+	rect->setZValue(30.0);
+	rect->setPen(QPen(legendBorderColor, legendBoxBorderSize));
+	rect->setBrush(QBrush(binColor(idx, numBins)));
+	text->setZValue(30.0);
+	text->setBrush(QBrush(darkLabelColor));
+}
+
+void Legend::hide()
+{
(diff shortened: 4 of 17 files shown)


core/imagedownloader.cpp
```diff
@@ -154,19 +154,9 @@ Thumbnailer::Thumbnail Thumbnailer::getHashedImage(const QString &filename, bool
...
-	QPainter painter(&res);
-	svg.render(&painter);
-	return res;
-}
-
-// As renderIcon, but render to a fixed width and scale height accordingly
+// As renderSVGIcon(), but render to a fixed width and scale height accordingly
 // and have a transparent background.
-static QImage renderIconWidth(const char *id, int size)
+static QImage renderSVGIconWidth(const char *id, int size)
```
dirkhh: I wonder if it wouldn't be more consistent to move this function into qthelper as well - those two seem rather closely related, having one here, one there seems... odd

core/gas.c
```diff
@@ -139,3 +140,26 @@ void fill_pressures(struct gas_pressures *pressures, const double amb_pressure,
...
 	}
 }
+
+enum gastype gasmix_to_type(struct gasmix mix)
+{
+	if (gasmix_is_air(mix))
+		return GASTYPE_AIR;
+	if (mix.he.permille == 0)
+		return mix.o2.permille >= 230 ? GASTYPE_NITROX : GASTYPE_AIR;
+	return mix.he.permille <= 300 ? GASTYPE_NORMOXIC : GASTYPE_TRIMIX;
```
dirkhh: So I am confused by that. 
The definition of **normoxic** that I am familiar with is a Trimix gas that has roughly normal oxygen content. Every agency thinks about the details differently, but a quick google search landed me here http://www.ddcbuildsite.com/index.php/easyblog/entry/normoxic-trimix-diving-diving-deep-with-a-clearer-head which suggests O2 values between 18-21% -- I was used to 19-22%
I have never heard of the term **normoxic** (which, etymologically seems to imply something like 'normal oxygen') to be triggered by the H2 percentage as in this formula...
If we want to go that route, then one other term that I hear thrown around a lot with trimix divers (full disclosure, I'm a TDI tech dive master and have dive-mastered (shouldn't say 'taught') advanced trimix classes in the past) is **hypoxic** (so hypo or below oxygen) for trimix gasses with an O2 content below whatever you pick for the lower threshold above - in my training 18/40 was considered **hypoxic**, but 19/40 was considered **normoxic**

I am NOT saying that these are the definitions we have to use - all I'm saying is that the definition above has me rather confused...

core/gas.h
```diff
@@ -22,6 +22,14 @@ struct gasmix {
 static const struct gasmix gasmix_invalid = { { -1 }, { -1 } };
 static const struct gasmix gasmix_air = { { 0 }, { 0 } };
 
+enum gastype {
+	GASTYPE_AIR,
+	GASTYPE_NITROX,
+	GASTYPE_NORMOXIC,
+	GASTYPE_TRIMIX,
+	GASTYPE_COUNT
```
dirkhh, line 32: I realize that this is missing another significant gas. **Oxygen** (O2 >= 98%)

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+	return !res.isValid();
+}
+
+struct gas_bin_t {
+	enum class Type {
+		Air,
+		Oxygen,
+		EAN,
+		Trimix
+	} type;
```
dirkhh, line 119: so three commits ago you added the gas types. There you didn't have **oxygen**, but you had **normoxic** (and were IMHO also missing **hypoxic**)
Maybe that's a different use case, but that would be at least worth a comment, right?

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+		return { Type::EAN, o2, 0 };
+	}
+	static gas_bin_t trimix(int o2, int h2) {
+		return { Type::Trimix, o2, h2 };
+	}
+};
+
+static bool is_invalid_value(const gas_bin_t &)
+{
+	return false;
```
dirkhh, line 138: not sure if this is permanently prevented somehow, but `o2+he > 1000` or either of the individual values `< 0` or `> 1000` all seem like valid `invalid` criteria

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+	default:
+	case gas_bin_t::Type::Oxygen:
+	case gas_bin_t::Type::Air:
+		return false;
+	case gas_bin_t::Type::EAN:
+		return t1.o2 < t2.o2;
+	case gas_bin_t::Type::Trimix:
+		return std::tie(t1.o2, t1.he) < std::tie(t2.o2, t2.he);
+	}
+}
```
dirkhh, line 174: I think it's fine to implement this in two places, but this should mention 
https://github.com/subsurface/subsurface/blob/edd19d4725153e6428e7b2e1b6594b4168601a39/core/divelist.c#L32-L39

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+
+std::vector<StatsValue> StatsVariable::values(const std::vector<dive *> &dives) const
+{
+	std::vector<StatsValue> vec;
+	vec.reserve(dives.size());
+	for (dive *d: dives) {
+		double v = toFloat(d);
+		if (is_invalid_value(v))
+			continue;
+		vec.push_back({ v, d });
```
dirkhh: I know, bike shedding, but wouldn't it be much easier to just do
```
if (!is_invalid_value(v))
        vec.push_back({ v, d });
```

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+		res.push_back({ std::move(b), std::move(value)});
+	}
+	return res;
+}
+
+template<typename Binner, typename Bin>
+std::vector<StatsBinDives> SimpleBinner<Binner, Bin>::bin_dives(const std::vector<dive *> &dives, bool fill_empty) const
+{
+	// First, collect a value / dives vector and then produce the final vector
+	// out of that. I wonder if that is permature optimization?
```
dirkhh: premature

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+
+	// Now, turn that into our result array with allocated bin objects.
+	return value_vector_to_bin_vector<Bin>(*this, value_bins, fill_empty);
+}
+
+template<typename Binner, typename Bin>
+std::vector<StatsBinCount> SimpleBinner<Binner, Bin>::count_dives(const std::vector<dive *> &dives, bool fill_empty) const
+{
+	// First, collect a value / counts vector and then produce the final vector
+	// out of that. I wonder if that is permature optimization?
```
dirkhh: premature

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+// (Note that calendar week is defined differently in different parts of the world and therefore omitted for now)
+
+double date_to_double(int year, int month, int day)
+{
+	struct tm tm = { 0 };
+	tm.tm_year = year;
+	tm.tm_mon = month;
+	tm.tm_mday = day;
+	timestamp_t t = utc_mktime(&tm);
+	return t / 86400.0; // Turn seconds since 1970 to days since 1970, if that makes sense...?
```
dirkhh, line 817: stupid bike shedding comment... leap seconds...?

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+		QLocale loc;
+		return StatsTranslations::tr("in %1 %2 steps").arg(loc.toString(bin_size),
+								   get_depth_unit());
+	}
+	QString unitSymbol() const override {
+		return get_depth_unit();
+	}
+	int to_bin_value(const dive *d) const {
+		return metric ? d->maxdepth.mm / 1000 / bin_size
+			      : lrint(mm_to_feet(d->maxdepth.mm)) / bin_size;
```
dirkhh, line 950: I guess this uses the preference, not a variable... but it seemed that I was never given an option to use anything but what's used in the preferences, so...
https://github.com/subsurface/subsurface/blob/edd19d4725153e6428e7b2e1b6594b4168601a39/core/units.h#L27

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+							    ml_to_cuft(d->sac);
+	}
+	std::vector<StatsOperation> supportedOperations() const override {
+		return { StatsOperation::Median, StatsOperation::Mean, StatsOperation::TimeWeightedMean };
+	}
+};
+
+// ============ Water and air temperature, binned in 2, 5, 10, 20 °C/°F bins ============
+
+// We use the same binner for °C and °F. We simply output the current temperature unit.
```
dirkhh: that's... surprising... but I guess it qualifies as "good enough"

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+	}
+	QString name() const override {
+		return StatsTranslations::tr("Air temperature");
+	}
+	double toFloat(const dive *d) const override {
+		return tempToFloat(d->airtemp);
+	}
+};
+
+// ============ Weight, binned in 1, 2, 5, 10 m or 2, 4, 10 or 20 ft bins ============
```
dirkhh: we are binning weight in meters or feet? I doubt that

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+
+struct WeightType : public StatsVariableTemplate<StatsVariable::Type::Numeric> {
+	QString name() const override {
+		return StatsTranslations::tr("Weight");
+	}
+	QString unitSymbol() const override {
+		return get_weight_unit();
+	}
+	int decimals() const override {
+		return prefs.units.weight == units::KG ? 1 : 0;
```
dirkhh: I always thought this was an odd choice (and it's been around forever)

stats/statsvariables.cpp
```diff
@@ -0,0 +1,1674 @@
...
+
+// ============ Gas type, in 2%, 5%, 10% and 20% steps  ============
+// This is a bit convoluted: We differentiate between four types: air, pure oxygen, EAN and trimix
+// The latter two are binned in x% steps. The problem is that we can't use the "simple binner",
+// because a dive can have more than one cylinder. Moreover, the string-binner might not be optimal
+// because we don't want to format a string for every gas types, when there are thousands of dives!
+// Note: when the same dive has multiple cylinders that fall inside a bin, that cylinder will
+// only be counted once. Thus, depending on the bin size, the number of entries may change!
+// In addition to a binner with percent-steps also provide a general-type binner (air, nitrox, etc.)
+
```
dirkhh, line 1338: that gas binner is... crazy... when testing the code I kept trying to figure out how it bins my gases. Just wow.

dirkhh (commented):
That is one monster commit at the end. But it makes sense to just get it over with.
It's making my brain melt and while I did add quite a few comments, I can't claim to have reviewed the correctness of every single line. It uses quite a bit of syntax constructs that I do understand, but that aren't comfortable, so I keep reminding myself what they actually mean. Which makes review harder.

There are a couple of issues that I mention in comments that stop me from approving.

core/gas.c
bstoeger: I originally had hypoxic, but I believe that @willemferguson objected.
See https://github.com/subsurface/subsurface/pull/3052#issuecomment-731947533 and follow-up discussion.
Perhaps I misunderstood things. The good thing is that this is trivial to change by changing gas.h and gas.c.

stats/statsvariables.cpp
bstoeger: Yes, the former was for binning with a "general type". This is for formatting bins that use the actual composition of the gases. Which is fun, because it means that we have two-dimensional bins. I shortly entertained the idea, but then refrained from doing 3D histograms (O2 vs H2 vs Count). 😀

stats/statsvariables.cpp
bstoeger: I'm not qualified enough to make a comment on that. I.e. I don't know how leap seconds are treated by the C library.

In any case, the deviation from linearity should be negligible.

stats/statsvariables.cpp
bstoeger: Yes - we never bin invalid gases, so invalid gas-bins will never exist:
```
                        struct gasmix mix = d->cylinders.cylinders[i].gasmix;
                        if (gasmix_is_invalid(mix))
                                continue;
```
Added a comment.

stats/statsvariables.cpp
bstoeger: These are related, but not the same: `get_dive_gas` gets the "maximum" gas. The `operator<` is used to establish a total order of gas bins.

stats/statsvariables.cpp
bstoeger: done.

stats/statsvariables.cpp
bstoeger: done.

stats/statsvariables.cpp
bstoeger: done.

stats/statsvariables.cpp
bstoeger: This macro does something completely different, no? It is used of you want to provide two "nice" numbers.

In any case - the code was very inconsistent : Sometimes it would respect the "metric" variable, sometimes use the preferences flag. After some thinking I concluded that the correct thing to do is using the variable. For now, this is an obscure difference. But what if we save the statistics into presets similar to the filters? The user may end up with loading a binner for a different unit system than in the preferences. Therefore the binners now all format according to their proper type, just to be sure. The numeric representations don't have that problem and therefore use the preferences flag.

stats/statsvariables.cpp
bstoeger: The unit is now made explicit.

stats/statsvariables.cpp
bstoeger: done.

stats/statsvariables.cpp
bstoeger: I think it kind of makes sense, because standard leads come in half-pound and full-kg units, no?

stats/statsvariables.cpp
bstoeger: It is a bit brutal, indeed. There are two dimensions to this:
- You can bin coarsely by type and finely by percentage.
- The gas is selected depending on the variable (max o2, max he). Not that it is too useful, but maybe someone wants to plot bottom gas vs. deco gas?

core/imagedownloader.cpp
bstoeger: done. That said, qthelper.cpp really needs to be split up.

core/gas.h
bstoeger: I added oxygen, but would suggest that you come to a conclusion with respect to hypoxic with @willemferguson.

bstoeger:
PS: added a weekday-binner, as that appeared to be low-hanging fruit.

bstoeger:
It is not really related, but since it is so trivial I added the colorscheme commit, so that we do some progress.

Waiting for @dirkhh's review, I will try to improve the highlighting of the scatter plots.

bstoeger:
Added the legend, since it is also small and self-contained.

@dirkhh: if this is too fast and you'd prefer to merge things in small pieces, let me know.

bstoeger:
Also let me know if you would prefer the legend box to have rounded corners, a different color, etc. These are all things that are trivial to implement.

dirkhh:
Everything needs to have rounded corners, right? That looks almost always better.
I just got up and as my family is still asleep, I'll try to get as far as I can with this. I'll look through your responses and the new commits right now.

core/gas.c
dirkhh: I can look through that discussion again and see how we got here - I'll focus on the rest, first, to optimize time :-)

stats/statsvariables.cpp
dirkhh: It was meant as a joke among all the others... mainly because you divided by `86400.0` which of course was intended to force floating point math, but to my silly eyes looked like a statement of precision...

stats/statsvariables.cpp
dirkhh: Yes, that's a good point. The ability to have bins in a different unit system than the one currently set in preferences might avoid pain later.

stats/statsvariables.cpp
dirkhh: not really. leads almost never weigh what people claim they do... and they are typically multiples of US pounds or half KG. So 10 lbs are 4.5kg. And because Linus and I both 'thought' metric a lot, but dove mostly with US weights, we wanted it to look correct both ways. Which is of course dumb, because if you dive with 3kg in Egypt then that's 6.6lbs :-)

stats/statsvariables.cpp
dirkhh: There are so many options there. I'm not convinced I fully understood everything you offer.

stats/statsvariables.cpp
dirkhh: Right - I was wondering if you want to comment that we use a similar order elsewhere and refer to it. Simply for consistency. I did get that these are different functions :-)

core/gas.c
dirkhh: Ok, found it: https://github.com/subsurface/subsurface/pull/3052#issuecomment-732253281
> With respect to the hypoxic gas class, I would recommend that you merge it with trimix. Full trimix dives are characterised by a hypoxic oxygen concentration (< 18%) and Helium < 30%. In contrast, normoxic dives are usually defined by oxygen around 20% and helium <= 30%. I would say that in a grey area the critical distinguishing factor is the oxygen concentration.

So this is very much what I talked about above, just with `< 18` instead of `<= 18` as in my training.
But I think his underlying comment kinda was to drop hypoxic (and not to add normoxic) and just have Air, Nitrox, Trimix (and maybe Oxygen as extreme Nitrox).
That said, I think it would be interesting to list all the classes (we can always add 12 million configuration options later... on second thought, let's NOT add all those options).
Outside of training dives I don't think you'll find a lot of trimix with O2 above 22%, so we could call that "Trimix" and the other two "Normoxic Trimix" and "Hypoxic Trimix". Makes sense?

stats/legend.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+Legend::Entry::Entry(const QString &name, int idx, int numBins, QGraphicsItem *parent) :
+	rect(new QGraphicsRectItem(parent)),
+	text(new QGraphicsSimpleTextItem(name, parent))
+{
+	rect->setZValue(30.0);
+	rect->setPen(QPen(legendBorderColor, legendBoxBorderSize));
+	rect->setBrush(QBrush(binColor(idx, numBins)));
+	text->setZValue(30.0);
+	text->setBrush(QBrush(darkLabelColor));
```
dirkhh, line 49: I realized later that this was a bad choice in dark mode. But we can fix that once everything is merged, I don't want to hold this up here.

stats/legend.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+void Legend::resize()
+{
+	if (entries.empty())
+		return hide();
+
+	QSizeF size = chart->size();
+
+	// Silly heuristics: make the legend at most half as high and half as wide as the chart.
+	// Not sure if that makes sense - this might need some optimization.
```
dirkhh, line 69: that does seem like a LOT. But you are right, easy to fine tune later.

dirkhh (approved):
I think the minor questions left are easily resolved.
Please consider the comments that I made about the gases, but once you decided on what to do (just Trimix, or three flavors of Trimix) and fixed the determination to be purely based on the O2 content, this is good to go

core/gas.c
bstoeger: > Outside of training dives I don't think you'll find a lot of trimix with O2 above 22%, so we could call that "Trimix" and the other two "Normoxic Trimix" and "Hypoxic Trimix". Makes sense?

Fine with me. Note that I think that @willemferguson's argument was that, precisely because hypoxic is the standard Trimix, mixtures <= 18% should be called "Trimix" without further qualifier. I will just qualify all three and you can then discuss which one is the "Trimix" version. FWIW, according to Wikipedia hyperoxic Trimix is also called "TriOx".

stats/statsvariables.cpp
bstoeger: Oh, I just realize that it is exactly the other way around than I would have expected. Should this condition be inverted?

stats/statsvariables.cpp
bstoeger: Comment added.

stats/statsvariables.cpp
dirkhh: What you have here is what we have elsewhere. One decimal for metric, no decimal for pounds. It's inconsistent and stupid. We should lead the way and simply always offer one decimal :-)

bstoeger: I inverted the condition: it is now one decimal for kg and no decimals for lbs. Of course we might simply change to one decimal fraction for all.

core/gas.c
```diff
@@ -139,3 +140,32 @@ void fill_pressures(struct gas_pressures *pressures, const double amb_pressure,
...
+{
+	if (gasmix_is_air(mix))
+		return GASTYPE_AIR;
+	if (mix.o2.permille >= 980)
+		return GASTYPE_OXYGEN;
+	if (mix.he.permille == 0)
+		return mix.o2.permille >= 230 ? GASTYPE_NITROX : GASTYPE_AIR;
+	if (mix.o2.permille <= 180)
+		return GASTYPE_HYPOXIC_TRIMIX;
+	return mix.o2.permille <= 230 ? GASTYPE_NORMOXIC_TRIMIX : GASTYPE_HYPEROXIC_TRIMIX;
```
dirkhh, line 154: I like it.
I'm not sure if there are people diving with hypoxic nitrox, TBH - you call that air here (and I hadn't mentioned it earlier). But again, I think this is fine.

bstoeger:
> Everything needs to have rounded corners, right? That looks almost always better.

Ok, let's do that later. I just realized that this means overwriting the `paint()` function, which I don't feel like doing currently. See `ToolTipItem::paint()`.

dirkhh (approved):
In order to keep moving, I think we should merge this once the CI passes...

stats/statsvariables.cpp
bstoeger: > What you have here is what we have elsewhere. One decimal for metric, no decimal for pounds. 

Which is funny, because I'm quite sure that this was just a mix-up from my side. Replaced by `return 1;`.

bstoeger:
> In order to keep moving, I think we should merge this once the CI passes...

Fine - will merge. There will surely be issues popping up, but we can fix them later.

PR #3123, opened 2021-01-01
fix potential crash in GPS code

dirkhh:
If we don't have a GpsLocation instance, we shouldn't dereference it.

- [x] Bug fix

tons of crashes on Android :-(

### Mentions:

@bstoeger, I think that should work - it's the only user of `::instance()` it seems


bstoeger:
That's horrible. Because if that is the cause of the crash it means that it is called before
```
        locationProvider = new GpsLocation(&appendTextToLogStandalone, this);
```
in `QMLManager::QMLManager()`.

Which probably means that this signal:
```
        connect(qobject_cast<QApplication *>(QApplication::instance()), &QApplication::applicationStateChanged, this, &QMLManager::applicationStateChanged);
```
is delivered asynchronously before initialization has finished. And this whole code is very much not thread safe. Unless of course the code enters the event loop somewhere. In any case, this all looks very wrong.

dirkhh:
> That's horrible. Because if that is the cause of the crash it means that it is called before
> 
> ```
>         locationProvider = new GpsLocation(&appendTextToLogStandalone, this);
> ```

Yes, that was the part that threw me as well.
But we have more than 240 crashes with this stack trace...

> 
> in `QMLManager::QMLManager()`.
> 
> Which probably means that this signal:
> 
> ```
>         connect(qobject_cast<QApplication *>(QApplication::instance()), &QApplication::applicationStateChanged, this, &QMLManager::applicationStateChanged);
> ```
> 
> is delivered asynchronously before initialization has finished. And this whole code is very much not thread safe. Unless of course the code enters the event loop somewhere. In any case, this all looks very wrong.

Agreed - but is my workaround a reasonable approach? I can't see how this would be wrong, but you know that code much better...

dirkhh:
I threw in the other fix as well - I want to get a new Android build out today if possible. This is affecting a lot of our users unfortunately.

bstoeger:
Is there a reason for the early connect()? If that fires asynchronously, then there seems to be a more fundamental problem.

Or turn the GpsLocation thing into a "normal" singleton that is generated on demand? I understand that you pass some weird logging callback, but that in itself appears to be wrong. Shouldn't the GpsLocation just call a generic logging function, which logs to the right place?

I mean you can see if that fixes this particular crash and then work on fixing the root cause. But I can only repeat myself: this looks all very wrong.

dirkhh:
Agreed. My immediate concern is to get this crash fixed or at least hacked around.
I will then look more closely at what I did there all these years ago and what would be a better way to do that.
Thanks for the feedback. The logging thing is from when this was intended to be used in both desktop and mobile - which it isn't, it's mobile only. I'll try to make this less crazy :-)

bstoeger:
> The logging thing is from when this was intended to be used in both desktop and mobile - which it isn't, it's mobile only.

Time to remove `GpsLocation *locationProvider;` and `#include "core/gpslocation.h"` from `mainwindow.h`?

bstoeger:
Added commit that removes these.

dirkhh:
ok, merging this and will work on a real fix for the `GpsLocation` mess next.

PR #3124, opened 2021-01-01
Cleanup gps location code

dirkhh:

- [x] Bug fix
- [x] Functional change
- [x] Code cleanup

This was much easier than I thought - but then maybe I didn't go as 'deep' fixing it as you likely expected me to :-)

### Changes made:

1) move the connect for the applicationStateChanged signal to the very end of the `QMLManager` constructor to ensure that the object is fully set up before that slot can be called
2) change the GpsLocation class to a more typical singleton with a static variable in the constructor
3) instead of completely redoing the log logic, simply add another method to set up the callback function

### Mentions:

@bstoeger -- good enough?

### DIFF
--- a/core/gpslocation.cpp
+++ b/core/gpslocation.cpp
@@ -15,17 +15,12 @@
 #include <QApplication>
 #include <QTimer>
 
-GpsLocation *GpsLocation::m_Instance = NULL;
-
-GpsLocation::GpsLocation(void (*showMsgCB)(const char *), QObject *parent) :
-	QObject(parent),
+GpsLocation::GpsLocation() :
 	m_GpsSource(0),
+	showMessageCB(0),
 	waitingForPosition(false),
 	haveSource(UNKNOWN)
 {
-	Q_ASSERT_X(m_Instance == NULL, "GpsLocation", "GpsLocation recreated");
-	m_Instance = this;
-	showMessageCB = showMsgCB;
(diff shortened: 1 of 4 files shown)


bstoeger (approved):
Looks good to me, but this needs some serious testing. This whole application-state-change thing is very obscure to me.

Moreover, I wonder if it wouldn't be more save to replace all `locationProvider` accesses by `GpsLocation::instance()` now?

dirkhh:
> Looks good to me, but this needs some serious testing. This whole application-state-change thing is very obscure to me.

That's the OS telling  the app that it is no longer in the foreground. Which on a mobile OS means "you can be killed any moment now"

> Moreover, I wonder if it wouldn't be more save to replace all `locationProvider` accesses by `GpsLocation::instance()` now?

Sure. Will do.

dirkhh:
Spent more time on this code and want to respond differently to your comments:
 - yes, the `applicationStateChange` code is weird and fragile, but the proposed change here doesn't make it any more weird and fragile. So while this is in fact code that I need to rethink at some point, I'd argue that this change is at least not making it worse
 - removing the `locationProvider` creates more than a dozen places where we call `GpsLocation::instance()`. I'm not sure that's an improvement, but I just pushed a commit to do that so we can decide if that's what we want.

dirkhh:
force pushed to replace a stray use of the old `connect()` syntax

PR #3126, opened 2021-01-02
cleanup: don't save PASCAL pressure units to git

bstoeger:
The way I understand, the PASCAL pressure unit is used to parse
obscure dive logs. However, there is no support in the UI for
using Pa as pressure unit. Therefore remove reading / writing
this unit to git divelogs.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Pa pressures are not supported by the UI(?) - don't write to / read from git logs.

### Release note:

No - shouldn't be user visible.

### DIFF
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -46,6 +46,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/file.c \
 	core/fulltext.cpp \
 	core/subsurfacestartup.c \
+	core/pref.c \
 	core/profile.c \
 	core/device.cpp \
 	core/dive.c \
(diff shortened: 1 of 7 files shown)


dirkhh:
I was happy with the first two commits. Not concerned with the third.
Now I need to make coffee before I decide to approve the fourth. I hate it when we touch the preferences.
But it compiles, so it must be correct, right?

dirkhh (approved):
43c9b4e379e957dbd1830f1c327a16161c2091dd

Ok, I convinced myself that this is the right thing to do.

dirkhh:
I don't think this needs a review from Linus - all you are removing is the code that saves the options - which is something that IIRC was actually written by me. But it's not changing the git storage format (which has always been `mbar` as _native_ unit.
So I'll merge without pinging him.

PR #3127, opened 2021-01-02
[cleanup] some profile code straightening

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

These are the first steps in making the profile widget reentrant, so that we can e.g. print without these weird ramifications, but also implement a life chart-widget in mobile. So far, this is mostly bike-shedding so that I can wrap my head around the control flow, which happens to be quite ... convoluted. We should really refrain from doing control flow via signals.

But it also improves the code in the sense that it halves the repainting of the profile items: They were painted twice per `plot()` call. In my tests I noticed that there is a rendering bug in the planner, but that exists at least since May 2020. It also removes some pointless optimizations: if the settings change there is no point in trying to be smart about when to replot the chart. The chart plotting must be instantaneous anyway if it should be useable in the planner.

### Changes made:

1) Numerous small cleanups.

### Additional information:

This needs some testing.

### Release note:

No - purely preparatory and internal stuff.

### DIFF
--- a/profile-widget/diveprofileitem.cpp
+++ b/profile-widget/diveprofileitem.cpp
@@ -19,32 +19,12 @@ AbstractProfilePolygonItem::AbstractProfilePolygonItem(const DivePlotDataModel &
 	setCacheMode(DeviceCoordinateCache);
 }
 
-void AbstractProfilePolygonItem::replot()
-{
-	modelDataChanged();
-}
-
 void AbstractProfilePolygonItem::setVisible(bool visible)
 {
 	QGraphicsPolygonItem::setVisible(visible);
 }
 
-bool AbstractProfilePolygonItem::shouldCalculateStuff(const QModelIndex &topLeft, const QModelIndex &bottomRight)
-{
-	if (dataModel.rowCount() == 0)
-		return false;
-	if (hDataColumn == -1 || vDataColumn == -1)
-		return false;
-	if (topLeft.isValid() && bottomRight.isValid()) {
-		if ((topLeft.column() >= vDataColumn || topLeft.column() >= hDataColumn) &&
-		    (bottomRight.column() <= vDataColumn || topLeft.column() <= hDataColumn)) {
-			return true;
-		}
-	}
-	return true;
-}
-
-void AbstractProfilePolygonItem::modelDataChanged(const QModelIndex&, const QModelIndex&)
+void AbstractProfilePolygonItem::replot()
 {
 	// Calculate the polygon. This is the polygon that will be painted on screen
 	// on the ::paint method. Here we calculate the correct position of the points
@@ -116,13 +96,11 @@ int DiveProfileItem::maxCeiling(int row)
 	return max;
 }
 
-void DiveProfileItem::modelDataChanged(const QModelIndex &topLeft, const QModelIndex &bottomRight)
+void DiveProfileItem::replot()
 {
 	bool eventAdded = false;
-	if (!shouldCalculateStuff(topLeft, bottomRight))
-		return;
 
-	AbstractProfilePolygonItem::modelDataChanged(topLeft, bottomRight);
+	AbstractProfilePolygonItem::replot();
 	if (polygon().isEmpty())
 		return;
 
@@ -219,17 +197,13 @@ DiveHeartrateItem::DiveHeartrateItem(const DivePlotDataModel &model, const DiveC
 	connect(qPrefTechnicalDetails::instance(), &qPrefTechnicalDetails::hrgraphChanged, this, &DiveHeartrateItem::setVisible);
 }
 
-void DiveHeartrateItem::modelDataChanged(const QModelIndex &topLeft, const QModelIndex &bottomRight)
+void DiveHeartrateItem::replot()
 {
 	int last = -300, last_printed_hr = 0, sec = 0;
 	struct {
 		int sec;
 		int hr;
 	} hist[3] = {};
-
-	// We don't have enougth data to calculate things, quit.
-	if (!shouldCalculateStuff(topLeft, bottomRight))
-		return;
 
 	qDeleteAll(texts);
 	texts.clear();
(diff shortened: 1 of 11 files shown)


dirkhh:
I was honestly under the impression that in Qt signals were the way to do control flow. 
Shows you how little I know.

bstoeger:
IMHO, Qt signals are a callback-mechanism, not a general control-flow mechanism and therefore should be used sparingly. The cases where I think Qt signals are useful / needed:
- When you need dynamic run-time determined dispatch instead of static compile-time determined dispatch.
- If you want to send data across thread boundaries - though Qt's thread model is "special".
- If you interact with Qt widgets that only define a signal interface.

But then I reckon that some people are fans of COMEFROM, so who am I to judge.

profile-widget/tankitem.cpp
```diff
@@ -111,6 +110,4 @@ void TankItem::replot()
 void TankItem::setHorizontalAxis(DiveCartesianAxis *horizontal)
 {
 	hAxis = horizontal;
-	connect(hAxis, SIGNAL(sizeChanged()), this, SLOT(replot()));
```
dirkhh: IIRC when you use profile edit functionality (right click on the profile) and drag the final handle around so that the dive gets longer, that will change the hAxis and that in return should create this signal. But I didn't try this.

dirkhh (commented):
fb0f227259d0515cf3a109a1f4423db84b2dfece
As usual, I admire your braveness. This code is so incredibly fragile - the interactions between things written by Tomaz and then 'improved' by me with half a dozen other developers hacking on it as well is... scary.
I have a couple real questions in the comments. And of course I wonder how much this has been tested - but generally your grasp of these things tends to be way better than mine...

profile-widget/profilewidget2.cpp
```diff
@@ -96,6 +96,14 @@ static struct _ItemPos {
 static const double thumbnailBaseZValue = 100.0;
 #endif
 
+template<typename T, class... Args>
+T *ProfileWidget2::createItem(const DiveCartesianAxis &vAxis, int vColumn, Args&&... args)
+{
+	T *res = new T(*dataModel, *timeAxis, DivePlotDataModel::TIME, vAxis, vColumn,
+		       std::forward<Args>(args)...);
```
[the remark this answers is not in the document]
bstoeger: `std::forward<>` does more than its name suggests.

The thing is that there are numerous ways of passing data from one function to the other (by value, by reference, by rvalue reference). The latter means that the caller is fine with the callee taking ownership of the data (i.e. move semantics):
```
        f(std::move(v)); // <-- poof, the vector may be empty (moved-from) after the call, because callee took ownership.
```

The gist of `std::forward(x)` is that it does the right thing depending on the type of x. It's mostly irrelevant in our case, because we don't use any of these subtleties, but I reckon it is "good style".

profile-widget/tankitem.cpp
bstoeger: Thanks for pointing this out - I will test this tomorrow.

dirkhh: OK, I'll leave it sitting here then :-)

bstoeger:
This code is indeed scary. A new gem I found: the DiveProfileItem adds an event to the dive:
```
                                        add_event(&displayed_dive.dc, entry->sec, SAMPLE_EVENT_CEILING, -1, max / 1000,
                                                QT_TRANSLATE_NOOP("gettextFromC", "planned waypoint above ceiling"));
```
Wow.

dirkhh:
That makes total sense. Every display layer should randomly change the underlying data, right?
Umm. Let me look for that in the code... ahh
https://github.com/subsurface/subsurface/blob/master/profile-widget/diveprofileitem.cpp#L178-L189
Yeah, that's not the place to do it, but it makes sense what it's trying to do... the culprit is indeed yours truly - in 2014(!!)
edited to add a link to the commit: fbd00c6b0da55b4c7add823c80e3d4afaf452544

dirkhh:
BTW, I am looking into enabling ceiling calculation for mobile. And it turns out that I ran into the disastrous mess that is our profile recalculation logic in the process. So I now have new appreciation for this PR (and will wait until this is ready to be merged for that work)...

profile-widget/diveprofileitem.cpp
```diff
@@ -278,14 +278,20 @@ DivePercentageItem::DivePercentageItem(const DivePlotDataModel &model, const Div
...
 	// Ignore empty values. a heart rate of 0 would be a bad sign.
 	QPolygonF poly;
+	colors.clear();
 	for (int i = 0, modelDataCount = dataModel.rowCount(); i < modelDataCount; i++) {
-		sec = dataModel.index(i, hDataColumn).data().toInt();
+		int sec = dataModel.index(i, hDataColumn).data().toInt();
 		QPointF point(hAxis.posAtValue(sec), vAxis.posAtValue(64 - 4 * tissueIndex));
 		poly.append(point);
+
+		double value = dataModel.index(i, vDataColumn).data().toDouble();
```
dirkhh, line 282: comment on the commit message: the second paragraph starts `Do all the calculations in paint()` which obviously isn't what you are doing (and would contradict the first paragraph).

dirkhh (commented):
This continues to look very nice. Small comment on a commit message that gets confused.
What I need to look at again is HOW the plot now realizes that it needs to be redone - just for my own understanding of how these things are connected. And to make sure I know which signals to send :-)

profile-widget/tankitem.cpp
dirkhh: btw, did you have a chance to look at this question? As usual, it's entirely possible that I'm wrong...

bstoeger: Note that I didn't look at this recently, I just pushed old commits that were lying around.

But when I looked at it a few days ago, the axis resized properly when moving the last marker.

dirkhh: that's encouraging.
I could test this some more and move towards merging (after fixing commit messages) - or is this something that you are still actively working on?
It's so hard to see from here where you are currently spending your energy :-)

bstoeger: Feel free to commit if nothing breaks.

dirkhh:
rebased this on current master and pushed into a different branch so the CI can run. I will then put my deco/mobile changes on top of that and mush that whole thing into my other PR... which means I can close this PR... but I'll keep it open until I actually did all that so if I get distracted this doesn't get lost :-)

bstoeger:
Attention: 1849521b8fadeac3f21ec90e30fca59d21575ca6 is buggy and doesn't properly clear the profile if no dive is selected. I will investigate.

dirkhh:
> Attention: [1849521](https://github.com/subsurface/subsurface/commit/1849521b8fadeac3f21ec90e30fca59d21575ca6) is buggy and doesn't properly clear the profile if no dive is selected. I will investigate.

Thanks - in the meantime I am learning that our disgusting pile of preprocessor macros that creates our QML accessible preferences doesn't actually emit `settingsChanged()`. So what I hoped would be easy, is turning harder :-(

dirkhh:
This has now been folded into #3133 
I am aware of your warning that part of this is buggy (and will add that over there for good measure).

dirkhh:
With this I believe I can close this one

PR #3129, opened 2021-01-03
statistics: don't reset binner if binner is set

bstoeger:
Recently code was added to reset variable 1 binner if the second
variable does not support an unbinned first variable.

It forgot to check whether a binner was already set. Do this.
But validate the old binner first!

This code is extremely fragile and will have to be redone.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix for bug reported on the ML. The code must be redone - the old way is not compatible with the new way.

### DIFF
--- a/stats/statsstate.cpp
+++ b/stats/statsstate.cpp
@@ -557,9 +557,11 @@ void StatsState::validate(bool varChanged)
 	if (var1 == var2)
 		var2 = nullptr;
 
+	validateBinner(var1Binner, var1, false);
+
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
This looks logical, and based on a quick test it does appear to fix both the issue that I reported with the temperatures as well as the the one mentioned by Christof

PR #3130, opened 2021-01-04
[WIP] Mobile stats

dirkhh:

- [X] non working playground

This doesn’t do anything useful.
It shows how I think we can load the statsview QtChart from QML.
Now the question becomes: how do I talk to this? I guess I need to reimplement something like the `statsview.cpp`?

### Mentions:

@bstoeger before I spend too much time on this, I wanted to run this by you in case it turns out this is completely the wrong way to go about this. My assumption was that I’d create a QML UI that feeds a C++ class that is analogous to `StatsView` with more or less the same information which will then result in the same time of statistics being show...???

### DIFF
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -302,7 +302,8 @@ HEADERS += \
 
 RESOURCES += mobile-widgets/qml/mobile-resources.qrc \
 		mobile-widgets/3rdparty/icons.qrc \
-		map-widget/qml/map-widget.qrc
+		map-widget/qml/map-widget.qrc \
+		stats/qml/statsview.qrc
 
 android {
 	SOURCES += core/android.cpp \
--- a/mobile-widgets/qml/StatisticsPage.qml
+++ b/mobile-widgets/qml/StatisticsPage.qml
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: GPL-2.0
+import QtQuick 2.6
+import org.subsurfacedivelog.mobile 1.0
+import org.kde.kirigami 2.4 as Kirigami
+
+Kirigami.Page {
+	id: statisticsPagge
+	objectName: "StatisticsPage"
+	title: qsTr("Statistics")
+	leftPadding: 0
+	topPadding: 0
+	rightPadding: 0
+	bottomPadding: 0
+	width: rootItem.colWidth
+	StatsView {
+		id: statisticsWidget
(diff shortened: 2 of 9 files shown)


bstoeger:
> My assumption was that I’d create a QML UI that feeds a C++ class that is analogous to StatsView with more or less the same information which will then result in the same time of statistics being show...???

No - this is not at all what I intended. StatsView should be common for both. I promised that I'll do the (basic) mobile UI - but today I'll have to catch up with work.

bstoeger:
Well - I messed that one up. I wasn't paying attention and derived from the wrong class. As the name implies `QQuickWidget`  is only for desktop apps - it should have been `QQuickItem`.

Therefore, the plan now is to convert `StatsView` to be a `QQuickItem` and use a `QQuickWidget` on desktop to display that. That would be a matter of minutes by deriving from QtCharts' DeclarativeChart. Sadly, this is declared in a private header file. I've had enough of this and will first get rid of the QtCharts module - it was useful but is now causing too much pain. This will take at least a week. I understand that you can't await having this on mobile, but let's do it right. We'll be better off in the long run.

dirkhh:
To avoid confusion... Clear priority for now is to get the desktop version ready. I was literally just playing around to understand how this would work.

So fixing the regression line and other issues that people might find really seems more important.

It may of course be that that still leaves replacement of QtCharts as best next step - I haven't tried to track that issue down.
All this is rather ironic, considering just how much effort was caused by trying to have QtCharts and its QML modules available on all of our platforms 🤷‍♂️

bstoeger:
> It may of course be that that still leaves replacement of QtCharts as best next step.

I think so, yes. For example, I couldn't make the legend movable by mouse. This should be trivial without QtCharts (knocks wood).

> All this is rather ironic, considering just how much effort was caused by trying to have QtCharts and its QML modules available on all of our platforms

I don't think it is ironic. Without this effort we wouldn't have a testable beta-version by now.

dirkhh:
As I usually do - I trust you on the advantages of removing QtCharts. Clearly we aren't using much of this anymore at this stage. Given the pain of making things work on desktop and mobile I had hoped that what they have would help us get there quicker -- and you are correct, it certainly helped us to get where we are right now -- and so I worry about how many pitfalls we'll stumble into as we create our own QML modules to replace what QtCharts already has... but maybe this will all turn out to be much simpler than what I currently fear.

PR #3131, opened 2021-01-04
stats: fix line segment intersection math

dirkhh:
Linear algebra class was a while ago, but somehow this does look more
logical to me.

- [x] Bug fix

I'll have the actual scientists review my math... but what was there was quite clearly not right.
What I changed it to makes sense to me and creates correct regression lines for me in all cases.

### Mentions:

@bstoeger


bstoeger (approved):
Yes - looks obviously correct.

PR #3132, opened 2021-01-04
Add dive rating and visibility to statistics variables

atdotde:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/stats/statsvariables.cpp
+++ b/stats/statsvariables.cpp
@@ -1681,6 +1681,58 @@ struct DayOfWeekVariable : public StatsVariableTemplate<StatsVariable::Type::Dis
 	}
 };
 
+// ============ Rating ============
+struct RatingBinner : public SimpleBinner<RatingBinner, IntBin> {
+	QString format(const StatsBin &bin) const override {
+		return QString("🌟").repeated(derived_bin(bin).value);
+	}
+
+	int to_bin_value(const dive *d) const {
+		int res = (int)d->rating;
+		return res;
+	}
+};
+
+static RatingBinner rating_binner;
+struct RatingVariable : public StatsVariableTemplate<StatsVariable::Type::Discrete> {
+	QString name() const override {
+		return StatsTranslations::tr("Rating");
+	}
+	QString diveCategories(const dive *d) const override {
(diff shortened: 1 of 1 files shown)


stats/statsvariables.cpp
```diff
@@ -1681,6 +1681,58 @@ struct DayOfWeekVariable : public StatsVariableTemplate<StatsVariable::Type::Dis
...
+	QString diveCategories(const dive *d) const override {
+		int rating = (int)d->rating;
+		return QString("🌟").repeated(rating);
+	}
+	std::vector<const StatsBinner *> binners() const override {
+		return { &rating_binner };
+	}
+};
+
+// ============ Rating ============
```
bstoeger: Visibility

stats/statsvariables.cpp
```diff
@@ -1681,6 +1681,58 @@ struct DayOfWeekVariable : public StatsVariableTemplate<StatsVariable::Type::Dis
...
+struct VisibilityBinner : public SimpleBinner<RatingBinner, IntBin> {
```
bstoeger: I'm surprised that this compiles. This should read VisibilityBinner. I will have to add some TMP (template meta programming) to catch this kind of error.

bstoeger (approved):
Two small copy & paste errors. If this compiles and works, it is fine with me.

One could think about a common base class or template for all the "star" variables, but it's probably not worth it. On a rainy afternoon maybe.

I really should remember to catch that wrong template argument, though.

dirkhh:
> Two small copy & paste errors. If this compiles and works, it is fine with me.
> 
> One could think about a common base class or template for all the "star" variables, but it's probably not worth it. On a rainy afternoon maybe.

BTW, not suggesting that this should be added, but we do have more of those star values. Waves, surge, and chill IIRC

> I really should remember to catch that wrong template argument, though.

Oh that will be fun. I can't wait to read THAT code 😉

atdotde:
Fixed the c&p errors. Decided against surf until somebody complains ;-)

PR #3133, opened 2021-01-05
Profile cleanups plus adding mobile calculated ceiling

dirkhh:

- [x] Functional change
- [x] New feature

This now includes @bstoeger 's profile update #3127 as part of this PR

_These are the first steps in making the profile widget reentrant, so that we can e.g. print without these weird ramifications, but also implement a life chart-widget in mobile. So far, this is mostly bike-shedding so that I can wrap my head around the control flow, which happens to be quite ... convoluted. We should really refrain from doing control flow via signals._

_But it also improves the code in the sense that it halves the repainting of the profile items: They were painted twice per `plot()` call. In my tests I noticed that there is a rendering bug in the planner, but that exists at least since May 2020. It also removes some pointless optimizations: if the settings change there is no point in trying to be smart about when to replot the chart. The chart plotting must be instantaneous anyway if it should be useable in the planner._

On top of that, the rest isfairly simple.
Don't disable the calculations.
Don't disable compiling in the drawing of the ceiling
Add a setting to enable the two types of ceiling data on device
Add some magic so the profile gets redrawn when those settings change (otherwise it feels weird when a dive that has a ceiling is shown e.g. on a tablet or with mobile-on-desktop)

### Additional information:

Initial benchmarking seems to indicate that the time to calculate the deco information on a reasonably long deep (65m / 80min) adds less than 10% to the total processing time of the profile - both when run on desktop as well as when run on one of the older phones that I have access to (Nexus 5X from 2015).
More importantly, the profile drawing takes **LESS** time even with ceiling enabled, compared to before this PR.

### Release note:

still needed

### Documentation change:

also still needed

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -1026,7 +1026,7 @@ static void calculate_deco_information(struct deco_state *ds, const struct deco_
 		for (i = 1; i < pi->nr; i++) {
 			struct plot_data *entry = pi->entry + i;
 			int j, t0 = (entry - 1)->sec, t1 = entry->sec;
-			int time_stepsize = 20;
+			int time_stepsize = 20, max_ceiling = -1;
 
 			current_divemode = get_current_divemode(dc, entry->sec, &evd, &current_divemode);
 			gasmix = get_gasmix(dive, dc, t1, &ev, gasmix);
@@ -1096,6 +1096,8 @@ static void calculate_deco_information(struct deco_state *ds, const struct deco_
 				double m_value = ds->buehlmann_inertgas_a[j] + entry->ambpressure / ds->buehlmann_inertgas_b[j];
 				double surface_m_value = ds->buehlmann_inertgas_a[j] + surface_pressure / ds->buehlmann_inertgas_b[j];
 				entry->ceilings[j] = deco_allowed_depth(ds->tolerated_by_tissue[j], surface_pressure, dive, 1);
+				if (entry->ceilings[j] > max_ceiling)
+					max_ceiling = entry->ceilings[j];
 				double current_gf = (ds->tissue_inertgas_saturation[j] - entry->ambpressure) / (m_value - entry->ambpressure);
 				entry->percentages[j] = ds->tissue_inertgas_saturation[j] < entry->ambpressure ?
 					lrint(ds->tissue_inertgas_saturation[j] / entry->ambpressure * AMB_PERCENTAGE) :
@@ -1105,6 +1107,20 @@ static void calculate_deco_information(struct deco_state *ds, const struct deco_
 				double surface_gf = 100.0 * (ds->tissue_inertgas_saturation[j] - surface_pressure) / (surface_m_value - surface_pressure);
 				if (surface_gf > entry->surface_gf)
 					entry->surface_gf = surface_gf;
+			}
+
+			// In the planner, if the ceiling is violated, add an event.
+			// TODO: This *really* shouldn't be done here. This is a contract
+			// between the planner and the profile that the planner uses a dive
+			// that can be trampled upon. But ultimately, the ceiling-violation
+			// marker should be handled differently!
+			// Don't scream if we violate the ceiling by a few cm.
+			if (planner_ds && !pi->waypoint_above_ceiling &&
+			    entry->depth < max_ceiling - 100 && entry->sec > 0) {
+				struct dive *non_const_dive = (struct dive *)dive; // cast away const!
+				add_event(&non_const_dive->dc, entry->sec, SAMPLE_EVENT_CEILING, -1, max_ceiling / 1000,
+					  translate("gettextFromC", "planned waypoint above ceiling"));
+				pi->waypoint_above_ceiling = true;
 			}
 
 			/* should we do more calculations?
--- a/core/profile.c
+++ b/core/profile.c
@@ -991,7 +991,8 @@ static void calculate_ndl_tts(struct deco_state *ds, const struct dive *dive, st
 
 /* Let's try to do some deco calculations.
  */
-void calculate_deco_information(struct deco_state *ds, const struct deco_state *planner_ds, const struct dive *dive, const struct divecomputer *dc, struct plot_info *pi, bool print_mode)
+static void calculate_deco_information(struct deco_state *ds, const struct deco_state *planner_ds, const struct dive *dive,
+				       const struct divecomputer *dc, struct plot_info *pi, bool print_mode)
 {
 	int i, count_iteration = 0;
 	double surface_pressure = (dc->surface_pressure.mbar ? dc->surface_pressure.mbar : get_surface_pressure_in_mbar(dive, true)) / 1000.0;
--- a/core/profile.c
+++ b/core/profile.c
@@ -1414,7 +1414,7 @@ struct divecomputer *select_dc(struct dive *dive)
 	return get_dive_dc(dive, i);
 }
 
-static void plot_string(const struct plot_info *pi, int idx, struct membuffer *b)
+static void plot_string(const struct dive *d, const struct plot_info *pi, int idx, struct membuffer *b)
 {
 	int pressurevalue, mod, ead, end, eadd;
 	const char *depth_unit, *pressure_unit, *temp_unit, *vertical_speed_unit;
@@ -1429,7 +1429,7 @@ static void plot_string(const struct plot_info *pi, int idx, struct membuffer *b
 		int mbar = get_plot_pressure(pi, idx, cyl);
 		if (!mbar)
 			continue;
-		struct gasmix mix = get_cylinder(&displayed_dive, cyl)->gasmix;
+		struct gasmix mix = get_cylinder(d, cyl)->gasmix;
 		pressurevalue = get_pressure_units(mbar, &pressure_unit);
 		put_format_loc(b, translate("gettextFromC", "P: %d%s (%s)\n"), pressurevalue, pressure_unit, gasname(mix));
 	}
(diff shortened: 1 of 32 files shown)


dirkhh:
Still needs CHANGELOG and documentation changes.
@willemferguson -- should I leave the documentation changes to you, or should I just add those myself?

dirkhh:
Warning from #3127... @bstoeger writes

Attention: 3d383658 is buggy and doesn't properly clear the profile if no dive is selected. I will investigate.

(I updated the SHA to point to the rebased commit)

bstoeger:
I rebased this on my current state and fixed the problem mentioned above. This was a bit hairy, so please check the result.

dirkhh:
I tried a quick diff to my last version and there are several new things that went in... so this will take a bit longer to review, and I am off to do some socially distant social things... so in a few hours.

dirkhh:
What I can say after a quick check is that settings changes no longer refresh the dive profile on mobile (specifically, my version from last night redrew the profile if you turned on the calculated ceiling, this one doesn't).

I'll dig into this when I come back.

dirkhh:
Oh dear. So strangely if I do a diff between the last version that I had that worked and what is now in this branch there is a TON of noise with random cleanups and changes and stuff, so clearly you had a rather divergent version of this branch laying around somewhere... and for whatever reason (and I cannot figure out what that reason might be, to be honest), with this version right here my trick to get the updated profile shown doesn't work. Even though the new profile is calculated (as the debug messages show me), it doesn't get rendered on screen.

I'm a bit at a loss as to what to do - since the branches are completely divergent very early in the series `git bisect` doesn't work at all, so I don't know which change is the one that 'breaks' the behavior that I wanted. I'll beat my head against the wall a little longer, I guess. Maybe something will click and I'll understand what's different. Because it makes no sense to me how any of the changes that I can see would have this effect. So likely it's one of these "QML randomly acts differently" moments where me pointing fingers at you is actually the WRONG answer.

bstoeger:
Don't worry about it, I will have a look later.

dirkhh:
To cleanse my brain I have no switched to reviewing the statistics/QtQuick branch instead. That should help :-)

dirkhh:
There... doing something else and coming back was the right thing to do.
Instead of the nonsense hackery that I had there that tried to force a reload of the page and depended on the underlying objects being released as the page was popped (which apparently happened in my earlier version, but no longer after your updated) this now is doing things in what I believe to be the right way.
Establish a signal in the root page that indicates a settings change - and explicitly trigger an update of that the profile if that signal is emitted.
This seems to work reliably
I'll do some more testing, but with this I believe this PR is ready to be merged as well... after the CI completes, of course.

dirkhh:
merged locally to add CHANGELOG.

core/profile.c
```diff
@@ -1107,6 +1109,20 @@ static void calculate_deco_information(struct deco_state *ds, const struct deco_
 					entry->surface_gf = surface_gf;
 			}
 
+			// In the planner, if the ceiling is violated, add an event.
+			// TODO: This *really* shouldn't be done here. This is a contract
```
atdotde, line 1112: Maybe a (historical) comment is in order: I view the ceilings only as a visualisation of the tissue loadings (which is the underlying data), so computing them in the rendering could be justified. 

One would think that the planner should not produce plans that violate the ceilings (pretty much my definition of what the planner is supposed to do) but it turns out it can: This happens mostly when using VPM-B where the deco model of the profile slightly differs from the deco model that the planner uses as the planner has more information available: in particular this the distinction between bottom phase of the dive and ascent phase which is well defined only in the planner and not for real dives. But we decided, the profile should use only information that is available also for logged dives and so it has to make a guess about that distinction (we use the point of the deepest ceiling). So it can happen, that with the information of the planner a point is below the ceiling but for the profile it is above. To understand that effect better, we wanted a visual representation of when this happens so we added the event.

Another possibility for this to happen is that for some reason (e.g. excessive computation time) the planner gives up and then one also wants to know up to which point to trust the plan. 

And yes, one could move the whole deco calculations of the profile to an earlier point (where one could still legitimately modify the dive) but I think this would end up being pretty ugly as well.

bstoeger: > Maybe a (historical) comment is in order: I view the ceilings only as a visualisation of the tissue loadings (which is the underlying data), so computing them in the rendering could be justified.

I understand that. The comment was concerning the addition of an event to the dive. Instead of adding an event to the dive, the code should add a "warning at point X" flag to the `profile_info` structure.

djunkins:
I started looking at this on the latest iOS image from Testflight. Is there anyway to see what GFs were used in the ceiling calculation or to edit them? On the desktop version I have GFs set at 40/85 and the ceiling shown on the mobile version is quite a bit deeper than in the profile on my desktop.

dirkhh:
Right now we don't support setting the GF on mobile, but that's on my todo list before we release that new version.
If you would like to look into that (should really be straight forward - but it is QML), I'd love that. Just let me know so we don't duplicate the work

djunkins:
> Right now we don't support setting the GF on mobile, but that's on my todo list before we release that new version.
> If you would like to look into that (should really be straight forward - but it is QML), I'd love that. Just let me know so we don't duplicate the work

I don't yet have a pipeline built to build an iOS target, but that is something I have wanted to do. Let me make sure I can get that working and I will let you know.

dirkhh:
You can simply build mobile-on-desktop and test on the desktop. I spend 90% of my time with that and then at the end create Android and iOS versions to see where they behave differently. But for basic UI work, this is just fine.

djunkins:
I'll give that a try tonight and will take a stab at adding preference fields for gradient factors in the advanced settings. What are opinions about only displaying the GF fields when the toggle to display calculated ceiling is on?

dirkhh:
> I'll give that a try tonight and will take a stab at adding preference fields for gradient factors in the advanced settings. What are opinions about only displaying the GF fields when the toggle to display calculated ceiling is on?

That seems like the right thing to do, now that I think about it.
Please reach out if you need help with mobile-on-desktop or with where to start for this project (`mobile-widgets/qml/Settings.qml` will be the first response 😀 )

djunkins:
OK. I've make some progress on this, but I'm having trouble figuring out what signal to emit to have the ceiling recalculated when the GF values change. I am currently sending a "settingsChanged" signal when the editing of the GF fields are complete (using the onEditingFinished signal from the qml text field), but while the label of the gflow/gfhigh on the top of the profile changes, the ceiling is not recalculated and the green shading remains for the old GF values. If I restart the app, the new ceiling shows up properly.

I've looked at the desktop preferences code and the only signal I can see being sent when the preferences are saved is the settingsSaved signal and yet somehow the ceiling is recalculated.

dirkhh:
The handling of these signals has changed. I thought that `settingsChanged()` would be sufficient to force a recalculation / redraw of the ceiling, but your testing appears to show differently.
@bstoeger - any thoughts on why that might be?

djunkins:
Digging into this, I'm trying to figure out whether the tissue recalculations is disabled in profilewidget2.cpp when SUBSURFACE-MOBILE is defined. It may be that the tissues are being calculated when the dive loads initially, but not updated when settings are changed on the mobile platform.

dirkhh:
That seems plausible. There's quite a bit of hackery going on there, and it's entirely possible that I missed an section where the `#ifndef SUBSURFACE_MOBILE` should have been removed...

djunkins:
I was not able to identify a misplaced `#ifndef SUBSURFACE_MOBILE` that was preventing tissue recalculations. I tried moving the #ifndef to allow calculation of allTissues in void ProfileWidget2::addItemsToScene(), but it did not help. Now I'm trying to trace the `settingsChanged()` signal from QML into the DiveListNotifer signal.

This is my first foray into QML, so it's taking me a bit of time to figure out what is going on.

bstoeger:
Please don't use `DiveListNotifier::settingsChanged`! This is an extremely heavy signal the reboots the whole application.

There are more fine grained signals for the individual settings. To my understanding the ceilings are recalculated for every replot, no? Thus a simple plot() call should be enough. But probably you are talking about something completely different?

djunkins:
OK. I'll focus on getting slots added to handle the gflowChanged and gfhighChanged signals in a way that recalculates the ceiling. I have to do some actual job-related work now, so I will try this tonight or tomorrow.

bstoeger:
If you give me a precise error description I can look at it. Currently I don't even know where to change these values on the mobile app.

I just note that when changing the depth of a dive, the profile doesn't update either.

dirkhh:
UGH, did that get broken again? I had that fixed a month or so ago.
That's when we started calling `update()` on the QMLProfile after an edit changed a dive and that was sufficient.

If we are back to not updating the profile after changing the dive data, that's bad :-(

djunkins:
Part of that may be related to the syntax for QML connections changing between Qt 5.12 and Qt 5.15. The following now gives a warning that it's been deprecated. In DiveDetailsView.qml we have:
```	
        Connections {
		target: rootItem
		onSettingsChanged: {
			qmlProfile.update()
		}
	}
```
The new format should be:
```	
        Connections {
		target: rootItem
		function onSettingsChanged() {
			qmlProfile.update()
		}
	}
```
https://github.com/qt/qtdeclarative/commit/a2eef6b511988b2435c4e39b6b5551e857ce7775

### Sorry for all the edits.... ###

djunkins:
> Part of that may be related to the syntax for QML connections changing between Qt 5.12 and Qt 5.15. The following now gives a warning that it's been deprecated. In DiveDetailsView.qml we have:
> 

This is the warning produced at runtime with Qt 5.15.2:

INFO: qrc:/qml/DiveDetailsView.qml:20:2: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... }

dirkhh:
I'm aware of this change, but Kirigami throws quite a few of those warnings, yet still appears to work.
My understanding is that this will break with Qt 6
But I would OF COURSE appreciate if you could fix that (assuming we aren't in one of those glorious situations where the fix is only available with Qt 5.15 - I want people on Linux distros that are still on Qt 5.12 to still be able to build Subsurface-mobile...

PR #3134, opened 2021-01-05
statistics: replace QChart::*Axis

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The easy step in removal of the QtCharts module: render our own axes. It's a bit ugly, but I tried to make it full configurable with constants in the various source files. It still needs some optimization concerning label sizes - it doesn't yet recognize when category labels are too large and should be replaced by ellipsis.

### Changes made:

1) Render custom axes
2) Render custom grid

### Mentions:

@dirkhh: This is only very lightly tested, but I don't think any fundamental problems will show up. Certainly a few things to optimize though.

### DIFF
--- a/stats/statsaxis.cpp
+++ b/stats/statsaxis.cpp
@@ -10,7 +10,7 @@
 #include <QFontMetrics>
 #include <QLocale>
 
-StatsAxis::StatsAxis(bool horizontal) : horizontal(horizontal)
+StatsAxis::StatsAxis(QtCharts::QChart *chart, bool horizontal) : chart(chart), horizontal(horizontal)
 {
 }
 
@@ -28,7 +28,7 @@ std::pair<double, double> StatsAxis::minMax() const
 // maximum-size strings especially, when using proportional fonts or for
 // categorical data. Therefore, try to err on the safe side by adding enough
 // margins.
-int StatsAxis::guessNumTicks(const QtCharts::QChart *chart, const QtCharts::QAbstractAxis *axis, const std::vector<QString> &strings) const
+int StatsAxis::guessNumTicks(const QtCharts::QAbstractAxis *axis, const std::vector<QString> &strings) const
 {
 	QFont font = axis->labelsFont();
 	QFontMetrics fm(font);
@@ -51,7 +51,8 @@ int StatsAxis::guessNumTicks(const QtCharts::QChart *chart, const QtCharts::QAbs
 	return std::max(numTicks, 2);
 }
 
-ValueAxis::ValueAxis(double min, double max, int decimals, bool horizontal) : StatsAxisTemplate(horizontal),
+ValueAxis::ValueAxis(QtCharts::QChart *chart, double min, double max, int decimals, bool horizontal) :
+	StatsAxisTemplate(chart, horizontal),
 	min(min), max(max), decimals(decimals)
 {
 }
@@ -66,7 +67,7 @@ static QString makeFormatString(int decimals)
 	return QStringLiteral("%.%1f").arg(decimals < 0 ? 0 : decimals);
 }
 
-void ValueAxis::updateLabels(const QtCharts::QChart *chart)
+void ValueAxis::updateLabels()
 {
 	using QtCharts::QValueAxis;
 
@@ -79,7 +80,7 @@ void ValueAxis::updateLabels(const QtCharts::QChart *chart)
 	QLocale loc;
 	QString minString = loc.toString(min, 'f', decimals);
 	QString maxString = loc.toString(max, 'f', decimals);
-	int numTicks = guessNumTicks(chart, this, { minString, maxString});
+	int numTicks = guessNumTicks(this, { minString, maxString});
 
 	// Use full decimal increments
 	double height = max - min;
@@ -105,16 +106,17 @@ void ValueAxis::updateLabels(const QtCharts::QChart *chart)
 	setTickCount(num + 1);
 }
 
-CountAxis::CountAxis(int count, bool horizontal) : ValueAxis(0.0, (double)count, 0, horizontal),
+CountAxis::CountAxis(QtCharts::QChart *chart, int count, bool horizontal) :
+	ValueAxis(chart, 0.0, (double)count, 0, horizontal),
 	count(count)
 {
 }
 
-void CountAxis::updateLabels(const QtCharts::QChart *chart)
+void CountAxis::updateLabels()
 {
 	QLocale loc;
 	QString countString = loc.toString(count);
-	int numTicks = guessNumTicks(chart, this, { countString });
+	int numTicks = guessNumTicks(this, { countString });
 
 	// Get estimate of step size
 	if (count <= 0)
@@ -150,13 +152,14 @@ void CountAxis::updateLabels(const QtCharts::QChart *chart)
 	setTickCount(numTicks);
 }
 
-CategoryAxis::CategoryAxis(const std::vector<QString> &labels, bool horizontal) : StatsAxisTemplate(horizontal)
+CategoryAxis::CategoryAxis(QtCharts::QChart *chart, const std::vector<QString> &labels, bool horizontal) :
+	StatsAxisTemplate(chart, horizontal)
 {
 	for (const QString &s: labels)
 		append(s);
 }
 
-void CategoryAxis::updateLabels(const QtCharts::QChart *)
+void CategoryAxis::updateLabels()
 {
 }
 
@@ -185,7 +188,8 @@ QString LabelDisambiguator::transmogrify(const QString &s)
 	}
 }
 
-HistogramAxis::HistogramAxis(std::vector<HistogramAxisEntry> bins, bool horizontal) : StatsAxisTemplate(horizontal),
+HistogramAxis::HistogramAxis(QtCharts::QChart *chart, std::vector<HistogramAxisEntry> bins, bool horizontal) :
+	StatsAxisTemplate(chart, horizontal),
 	bin_values(std::move(bins))
 {
 	if (bin_values.size() < 2) // Less than two makes no sense -> there must be at least one category
@@ -223,7 +227,7 @@ std::pair<double, double> HistogramAxis::minMax() const
 // If labels are skipped, try to skip it in such a way that a recommended label is shown.
 // The one example where this is relevant is the quarterly bins, which are formated as (2019, q1, q2, q3, 2020, ...).
 // There, we obviously want to show the years and not the quarters.
-void HistogramAxis::updateLabels(const QtCharts::QChart *chart)
+void HistogramAxis::updateLabels()
 {
 	if (bin_values.size() < 2) // Less than two makes no sense -> there must be at least one category
 		return;
@@ -238,7 +242,7 @@ void HistogramAxis::updateLabels(const QtCharts::QChart *chart)
 	strings.reserve(bin_values.size());
 	for (auto &[name, value, recommended]: bin_values)
 		strings.push_back(name);
-	int maxLabels = guessNumTicks(chart, this, strings);
+	int maxLabels = guessNumTicks(this, strings);
 
 	int step = ((int)bin_values.size() - 1) / maxLabels + 1;
 	if (step < preferred_step) {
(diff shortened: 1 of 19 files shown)


stats/statsaxis.cpp
```diff
@@ -45,31 +70,127 @@ int StatsAxis::guessNumTicks(const QtCharts::QAbstractAxis *axis, const std::vec
 		minSize = minSize * 3 / 2;
 	else
 		minSize *= 2;
-	QRectF chartSize = chart->plotArea();
-	double availableSpace = horizontal ? chartSize.width() : chartSize.height();
-	int numTicks = lrint(availableSpace / minSize);
+	int numTicks = lrint(size / minSize);
```
dirkhh, line 88: this confuses me - you have `size` as loop local variable in the for loop above (line 62), but which `size` are you accessing here?
Oh, going back I realize that there's a member named `size` now. Then maybe we should rename the variable in the loop above? I know that this is syntactically and semantically valid, but... ugh.

dirkhh (approved):
Just a couple small comments.
As expected, we are gaining a lot of code that we hoped not to need to maintain ourselves by using QtCharts. Oh well - I understand why we got here.
As you requested I will test this a little more before merging.

dirkhh:
Ok, here's some feedback based on side-by-side comparison with the QtChart axes code...
Your axes are significantly thicker, making them stand out in a way that I don't find particularly appealing, especially side-by-side. Even the text seems 'fatter' and more 'squat' - is this the same font?
The grid lines are now dotted instead of just of a rather faint color - that change I don't mind as much and wouldn't have noticed without the direct comparison.
I wasn't able to make your code fail, though.

dirkhh:
Screen Shot 2021-01-05 at 9 30 04 AM

tcanabrava:
That is beautiful.

tcanabrava:
I wonder if it's not possible to use the Profile code for this too, we
already have a way to specify grids and axis there.
We need a way to specify bars - shouldn't be too hard.
and with that we can present some statistics on top of the Dive Profile.

dirkhh:
> I wonder if it's not possible to use the Profile code for this too, we already have a way to specify grids and axis there. We need a way to specify bars - shouldn't be too hard. and with that we can present some statistics on top of the Dive Profile.

I agree that we now have two different places where we have code that deals with creation of axes, labels, ticks, grids.
Not ideal, worth looking into in the future.

The idea of statistics on top of the dive profile... I don't immediately see how that is useful, but that could be simply a sign of my limitations 😄  -- more seriously, though, I think the profile is already crazy busy, I'm not sure that making it more busy is helpful - and the profile conceptually is about one dive, the statistics are always about multiple dives, right?

dirkhh:
As I wanted to comment on Tomaz'  "that is beautiful" comment and show one of my favorite charts... I noticed that with the new axes code we lost the regression line...
Screen Shot 2021-01-05 at 9 58 40 AM

tcanabrava:
On Tue, Jan 5, 2021 at 5:56 PM Dirk Hohndel <notifications@github.com>
wrote:

> I wonder if it's not possible to use the Profile code for this too, we
> already have a way to specify grids and axis there. We need a way to
> specify bars - shouldn't be too hard. and with that we can present some
> statistics on top of the Dive Profile.
>
> I agree that we now have two different places where we have code that
> deals with creation of axes, labels, ticks, grids.
> Not ideal, worth looking into in the future.
>
> The idea of statistics on top of the dive profile... I don't immediately
> see how that is useful, but that could be simply a sign of my limitations
> 😄 -- more seriously, though, I think the profile is already crazy busy,
> I'm not sure that making it more busy is helpful - and the profile
> conceptually is about one dive, the statistics are always about multiple
> dives, right?
>

The way I designed the Profile code is that it's easy to remove the dive
and place anything there, so we could "reuse" the widget
- The profile was initially constructed as a way to visualize multiple
dives, this was removed in subsequent console versions (because we used to
have a `one` dive_profile active all the time)

my initial idea - years and years ago
Select a few dives: All of them are visible on the profile (in miniature
format)
A graph appears below them, with the statistics.

(strangely, those graph pictures are somewhat similar to what luiza mocked
up for us years ago)

Tomaz

stats/statsaxis.cpp
```diff
@@ -10,7 +10,7 @@
 #include <QFontMetrics>
 #include <QLocale>
 
-StatsAxis::StatsAxis(bool horizontal) : horizontal(horizontal)
```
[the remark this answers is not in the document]
bstoeger: Wow - that was a  dyslexic commit message. I tried to reformulate it.

stats/statsaxis.cpp
bstoeger: Renamed loop variable to `labelSize`.

bstoeger:
> Ok, here's some feedback based on side-by-side comparison with the QtChart axes code...
Your axes are significantly thicker, making them stand out in a way that I don't find particularly appealing, especially side-by-side. Even the text seems 'fatter' and more 'squat' - is this the same font?
The grid lines are now dotted instead of just of a rather faint color - that change I don't mind as much and wouldn't have noticed without the direct comparison.

Yes, the axes looked fat. To be honest, these were just random values I came up with, in the hope that someone else would optimize this. I made the axes an ticks thinner and removed the dotting. The font is what we get from `QFont`'s default constructor. No idea.

bstoeger:
> I wonder if it's not possible to use the Profile code for this too, we
already have a way to specify grids and axis there.
We need a way to specify bars - shouldn't be too hard.
and with that we can present some statistics on top of the Dive Profile.

In hindsight, this might have been a good idea. But - not knowing what I was doing - I started with a `QtCharts` detour. Before unifying these things, I want to get an insight into the whole QQuickItem problem space, and decide how to proceed. Also in the light of bringing a dynamic chart to mobile. What I see so far is off-putting, but oh well.

dirkhh:
> Yes, the axes looked fat. To be honest, these were just random values I came up with, in the hope that someone else would optimize this. I made the axes an ticks thinner and removed the dotting. The font is what we get from `QFont`'s default constructor. No idea.

Being explicit helps me. You saw this a couple of days ago. "@dirkhh - the axes look too fat and there are other things that look off, if you have a chance, would you look at that" -- that tells me that you expect me to address that and that I won't step on something that you are already working on. As I did for example for the icons for the charts.

I always want to balance helping and not stepping on people's toes 🤷🏼‍♂️

dirkhh:
> > I wonder if it's not possible to use the Profile code for this too, we
> > already have a way to specify grids and axis there.
> > We need a way to specify bars - shouldn't be too hard.
> > and with that we can present some statistics on top of the Dive Profile.
> 
> In hindsight, this might have been a good idea. But - not knowing what I was doing - I started with a `QtCharts` detour. Before unifying these things, I want to get an insight into the whole QQuickItem problem space, and decide how to proceed. Also in the light of bringing a dynamic chart to mobile. What I see so far is off-putting, but oh well.

Many of our detours are painful - some of them have proved helpful.
I am so thrilled with the result of the statistics work that I will not question the steps that got us here.

Please let me know where / how I can help with the process of getting this into the mobile app. Because yes, that has always been the source of countless pain points. And my guess is that there are many ways to get to what we are looking for. So let's find the path that's least 'off-putting'.

bstoeger:
> As I wanted to comment on Tomaz' "that is beautiful" comment and show one of my favorite charts... I noticed that with the new axes code we lost the regression line...

I added a "quick fix". However, I think this is not the correct fix, so let me think about this tomorrow.

bstoeger:
The regression line and median markers should be fixed now, please test. (To test the median/mean markers: plot duration with binning against "none"). BTW: I think the colors of these markers could be made to stand out less.

> As expected, we are gaining a lot of code that we hoped not to need to maintain ourselves by using QtCharts. Oh well - I understand why we got here.

I don't think it is so much and it buys us lots of flexibility (e.g. subtics) that QtCharts doesn't have. Moreover note that some code also disappeared, e.g.the crazy make-labels-unique code by adding zero-width unicode spaces.

> Even the text seems 'fatter' and more 'squat' - is this the same font?

It is bold:
```
        labelFont(QFont()),     // make this configurable
        titleFont(QFont(labelFont.family(), labelFont.pointSize(), QFont::Bold, false)), // same as label, but bold
```
feel free to change the weight, size, etc and force push.

dirkhh:
> The regression line and median markers should be fixed now, please test. (To test the median/mean markers: plot duration with binning against "none"). 

will test over lunch

> BTW: I think the colors of these markers could be made to stand out less.
> 
> > As expected, we are gaining a lot of code that we hoped not to need to maintain ourselves by using QtCharts. Oh well - I understand why we got here.
> 
> I don't think it is so much and it buys us lots of flexibility (e.g. subtics) that QtCharts doesn't have. Moreover note that some code also disappeared, e.g.the crazy make-labels-unique code by adding zero-width unicode spaces.

Yeah, I noticed that. I repeat - I understand how we got here. I should just stop with these editorial comments 😇

> > Even the text seems 'fatter' and more 'squat' - is this the same font?
> 
> It is bold:
> 
> ```
>         labelFont(QFont()),     // make this configurable
>         titleFont(QFont(labelFont.family(), labelFont.pointSize(), QFont::Bold, false)), // same as label, but bold
> ```
> 
> feel free to change the weight, size, etc and force push.

I'll do that over lunch - I'll try to be ready before you usually disappear for the day.

stats/statsview.cpp
```diff
@@ -716,54 +719,70 @@ void StatsView::QuartileMarker::updatePosition()
...
 
-	// If not fully inside drawing region, do clipping. With the check above this guarantees that a != 0,
-	// but owing to floating point imprecision, let's test again.
+	// If not fully inside drawing region, do clipping.
 	if ((y1 < minY || y1 > maxY || y2 < minY || y2 > maxY) && fabs(a) > 0.0001) {
 		// Intersections with y = minY and y = maxY lines
 		double intersect_x1 = (minY - b) / a;
 		double intersect_x2 = (maxY - b) / a;
-		if (intersect_x1 > intersect_x2)
+		if (intersect_x1 < intersect_x2)
```
dirkhh: So you switched this comparison back... that was one of the things that I changed to fix the clipping.
With this branch I again get things like this
Screen Shot 2021-01-06 at 10 31 47 AM

bstoeger: merge error - corrected.

dirkhh:
That was funny. I pulled the branch, added my fixes (< 5min) and tried to push it and the push failed.
What the heck? I had pulled something like 30 seconds before your force push...
I added the changes to the weight of labels and ticks and stuff in a separate commit to make it easier for you to look at. I do think this should simply be fixed earlier in the series...
Either way, with that I think this is ready to be merged. If you agree with the above (and the thinner look) then I can just do that locally.

bstoeger:
Fine with me.

dirkhh:
folded my code into yours were it belonged (none of that was worth a separate commit) and merged into master

bstoeger:
Does not directly belong to this PR. I read a bit of the QtQuick documentation and this is utterly bizarre.

On the one hand the Qt creators recommend QtQuick as being more "modern" than QGraphicsScene, but when you look at the source code and examples it's "the 1980ies called and want their type system back". Not to mention the impossibility of deriving from QtQuick classes (apparently that is improved with Qt 6.0).

But worse, the thing is ridiculously incomplete: There is no TextItem: https://bugreports.qt.io/browse/QTBUG-43272 You must be kidding!?

So there goes the plan to implement the thing using QtQuick primitives and we'll just render into a QGraphicsScene and project that into QtQuick, similar to the profile.

This is all very sad. But what makes this really really weird to me is that you will find lots of praise for QtQuick. I don't get it. It's archaic, unfinished, non-extensible...what is the upside?

And it also raises the question how to do the dynamic chart (planner). Render most of the chart using QtQuick primitives and the axes using QML Text items? Jeez.

bstoeger:
PS: The documentation is also ridiculous. https://doc.qt.io/qt-5/qsgimagenode.html

```
Detailed Description
Warning: The image node class must have a texture before being added to the scene graph to be rendered.
```

Very detailed, thanks. :-/

dirkhh:
So, let's throw this all out and start from scratch. Do what Blender does and write our own libraries, right? How hard can it be?

Kidding, of course. Watching the coup unfold in DC is making me twitchy

bstoeger:
We could port the application to the Blender libraries. They probably support text items.

PR #3136, opened 2021-01-05
Fix update check

dirkhh:

- [x] Bug fix
- [x] Code cleanup

This has been broken for so long. And it bugged me every now end then that I'd get the question whether I wanted to allow the update check. But of course I was always testing something else and just clicking 'Accept' was quick and easy and I never spent the time to actually figure out _why_ this was broken.

Turns out that this got messed up quite a while ago when we changed the preferences code.

### DIFF
--- a/core/pref.h
+++ b/core/pref.h
@@ -47,7 +47,6 @@ enum def_file_behavior {
 
 typedef struct {
 	bool dont_check_for_updates;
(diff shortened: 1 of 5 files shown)


dirkhh:
I guess I'm the only one who cares about this stuff... being down to just a couple of developers makes this all a lot harder.

PR #3137, opened 2021-01-06
[snap] add daily USN check

Saviq:

- [x] Build system change

This workflow will download the current snaps published in the `candidate`
channel for all architectures and check them for packages with published
Ubuntu Security Notices. If it finds any, it will trigger a build of the
snap recipe:

https://code.launchpad.net/~subsurface/+snap/subsurface-stable

This will rebuild the snap with patched packages and publish it to the
`candidate` channel.

### Changes made:
1) add `check_usns.py` script
2) add workflow to run the above daily

### Additional information:
The workflow requires the result of the following (contents of `lpcreds`) be added as the `LAUNCHPAD_CREDENTIALS` secret:
```python
from launchpadlib.launchpad import Launchpad, UnencryptedFileCredentialStore
Launchpad.login_with(
  "subsurface-ci",
  "production",
  version="devel",
  credential_store=UnencryptedFileCredentialStore("lpcreds"),
)
```

It would be best if these were one for a Launchpad account dedicated to CI. Something along the lines of https://launchpad.net/~subsurface-ci-bot.

I simulated the cron run in these three Actions runs:
- [No USNs found](https://github.com/Saviq/subsurface/runs/1656876855?check_suite_focus=true#step:7:8)
- [USNs found](https://github.com/Saviq/subsurface/runs/1656908685?check_suite_focus=true#step:7:11)
- [builds pending](https://github.com/Saviq/subsurface/runs/1656925534?check_suite_focus=true#step:7:9)

### Mentions:
@dirkhh

### DIFF
--- a/.github/workflows/scripts/check_usns.py
+++ b/.github/workflows/scripts/check_usns.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python3
+# coding: utf-8
+
+import json
+import logging
+import multiprocessing
+import os
+import pathlib
+import pprint
+import sys
+import subprocess
+import tempfile
+
+from launchpadlib import errors as lp_errors  # fades
+from launchpadlib.credentials import RequestTokenAuthorizationEngine, UnencryptedFileCredentialStore
+from launchpadlib.launchpad import Launchpad
+import requests  # fades
+
+
+logger = logging.getLogger("subsurface.check_usns")
+logger.addHandler(logging.StreamHandler())
+logger.setLevel(logging.INFO)
+
+APPLICATION = "subsurface-ci"
+LAUNCHPAD = "production"
+RELEASE = "bionic"
+TEAM = "subsurface"
+SOURCE_NAME = "subsurface"
+SNAPS = {
+    "subsurface": {"candidate": {"recipe": "subsurface-stable"}},
+}
+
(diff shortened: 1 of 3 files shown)


dirkhh:
This is one of those moments when I admit that I'm not 100% sure what a PR actually does. I need to do some more reading and googling - I have no doubt that you are submitting the right things here, I just feel like I have an obligation to understand better what it does. So any delay is simply a wait for me to find time to dig into this some more.
Thanks for submitting it, it's stated goal is certainly something I strongly agree with

Saviq:
> So any delay is simply a wait for me to find time to dig into this some more.

Of course, no problem. The script may be a bit more complex than it needs to be, as I've used a version of it for more complex situations.

Let me know if you'd like me to go over it with you, IRC or otherwise.

dirkhh:
I think I understand now what this does... and I created the secret, but I wasn't 100% sure what to encrypt there.
The python script above created a file that looks basically like this
```
consumer_key = CI Bot for Launchpad
consumer_secret = 
access_token = someTOKEN
access_secret = someSECRET
```
Do I add all four lines to the secret? Or only the access_secret?

Saviq:
> Do I add all four lines to the secret? Or only the access_secret?

The whole thing, yes - there should also be a `[1]` at the top.

dirkhh:
I missed the [1]. I'll update the secret and merge this.

dirkhh:
Now let's see if it works as expected :-)

PR #3138, opened 2021-01-06
[snap] add Linux Snap workflow

Saviq:

- [x] Build system change

This will build and verify the snap on pull requests and pushes to `master`.

### Changes made:
1) added the Linux Snap workflow

### Additional information:
The `ccache` part of this workflow is obviously optional, it shaves up to 30% off the build time in exchange for extra complexity. Let me know if you'd rather drop it.

### Mentions:
@dirkhh

### DIFF
--- a/.github/workflows/linux-snap.yml
+++ b/.github/workflows/linux-snap.yml
@@ -1,0 +1,54 @@
+name: Linux Snap
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  Snap:
+    runs-on: ubuntu-latest
+
+    env:
+      SNAPCRAFT_BUILD_INFO: 1
+
+    timeout-minutes: 60
+    steps:
+    - name: Install Snapcraft
+      uses: samuelmeuli/action-snapcraft@v1
+      with:
+        use_lxd: true
+
+    - name: Check out code
+      uses: actions/checkout@v2
+      with:
+        # Needed for version determination to work
+        fetch-depth: 0
+
+    - name: Build
+      run: |
+        # Build the `subsurface` part.
+        sg lxd -c '/snap/bin/snapcraft build --use-lxd subsurface'
+
+    - name: Build and verify the snap
+      id: build-snap
+      env:
+        SNAP_ENFORCE_RESQUASHFS: 0
+      run: |
+        # Actually build the snap.
+        sg lxd -c '/snap/bin/snapcraft --use-lxd'
+
+        sudo snap install review-tools
+        /snap/bin/review-tools.snap-review *.snap
+
+        echo "::set-output name=snap-file::$( ls *.snap )"
+
+    - name: Upload the snap
+      uses: actions/upload-artifact@v2
+      with:
+        name: ${{ steps.build-snap.outputs.snap-file }}
+        path: ${{ steps.build-snap.outputs.snap-file }}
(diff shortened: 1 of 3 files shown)


.github/workflows/linux-snap.yml
```diff
@@ -0,0 +1,43 @@
+name: Linux Snap
+
+on:
+  pull_request:
+    branches:
+    - master
```
dirkhh, line 9: So this is built on pulls, but then not built on pushes. Which means we'll have tests on what's in pull requests, but not a build for the latest version in master, am I reading that right?
One thing the people on our developer list like to have the ability to do is to quickly test "the current version". While this is a developer mailing list in name, a lot of people aren't really comfortable (or, frankly, able) to build from source - so having a 'current master' Snap would be a great thing for those people - I already create 'current master' AppImages (and macOS and Windows installers).
Would this be a reasonable add, in your opinion?
```
on:
  push:
    branches:
    - master
  pull_request:
    branches:
    - master
```

**Edited to add**:
and of course as I reach the 3rd commit in my review, I see you add that there.
That should really be added here, right?

.github/workflows/linux-snap.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    env:
+      SNAPCRAFT_BUILD_INFO: 1
+
+    timeout-minutes: 60
+    steps:
+    - name: Install Snapcraft
+      uses: samuelmeuli/action-snapcraft@v1
+      with:
+        snapcraft_token: ${{ secrets.SNAPCRAFT_TOKEN }}
```
dirkhh: that secret doesn't exist - what needs to be in that?

.github/workflows/linux-snap.yml
```diff
@@ -0,0 +1,43 @@
...
+        # Build the `subsurface` part.
+        sg lxd -c '/snap/bin/snapcraft build --use-lxd subsurface'
+
+    - name: Build and verify the snap
+      id: build-snap
+      env:
+        SNAP_ENFORCE_RESQUASHFS: 0
+      run: |
+        # Actually build the snap.
+        sg lxd -c '/snap/bin/snapcraft --use-lxd'
```
dirkhh, line 67: that looks... odd... but I'll take your word for it :-)

dirkhh (approved):
So the main suggestion turned out to be moot - but I left it there because I thought this would better be done all in the first commit.
Which leaves me with the question about the secret

.github/workflows/linux-snap.yml
dirkhh: A snap is a single file in the end, correct? So this could be uploaded either as a GitHub artifact or be pushed onto my server. Having it in a 'testing' channel on the store would of course be best (as in - easiest for users to interact with), but I understand that's a bit harder.

.github/workflows/linux-snap.yml
dirkhh: the point was that in the previous invocation of basically the same command line, there's a reference to `subsurface`, vs here it's just `/snap/bin/snapcraft --use-lxd` without further argument.

Saviq: Ah yes, this is implicitly `snapcraft snap`, kinda like `make` makes the default target.

Saviq:
All fixed @dirkhh.

dirkhh (approved):
looks good

PR #3140, opened 2021-01-07
[EXPERIMENTAL] Turn statistics widget into QQuickItem

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a first step in turning the StatsView into a QtQuick thing. The code is currently completely insane: for every change, the full scene is rendered into a QImage and then converted into a texture to be shown on the item. This happens for every single mouse-move, since that might move the "info-box".

And of course, it flickers, tears, has antialiasing artifacts owing to int/real conversion issues. But it does show something on mobile: go to statistics and click on "plot".

The plan is now twofold:
1) Slowly convert the different chart features (legend, infobox, series, etc.) to independent QQuickItems.
2) Try to fix the visual artifacts. The QtCharts code looks quite daunting.

I predict this will be tedious, therefore I will probably take a few days off from this topic.

### Mentions:

@dirkhh: please check if you can see the chart on mobile.


bstoeger:
PS: If it wasn't clear - this removes the QtCharts dependency, which should now be removed from the build system.

PPS: For some reason I couldn't move the qt5_add_resources thing of the desktop-only file to desktop-only cmake file.

dirkhh:
Wow.
Quite something...
The overall design how C++ code is supposed to display things in a QML UI seems a little weird. I've done some reading and there seem to be better ways to do this, some fairly recent (I think 5.14), but I'll need to go back and find this again

dirkhh:
Had a quick break between meetings - I think the way this is designed to be done is to have a `QQuickPaintedItem` where you implement your own `paint()` function that updates the visual when needed. I haven't spent the time to read through your code in detail to compare that approach to what you do. I had started to look into doing this for our profile widget and never really found the time to dig into the approach enough to evaluate if this was a better approach than the `QImage`/`QPixmap` rendering approach that we use there.

bstoeger:
> Had a quick break between meetings - I think the way this is designed to be done is to have a QQuickPaintedItem where you implement your own paint() function that updates the visual when needed.

Yes, I've seen that in the profile code, and I think essentially this does the same thing (render onto a texture and show that).

However - and I was not clear enough - this is precisely what I do *not* want to do. The idea is to render every chart element (axes, legend, infobox, etc) into its own QQuickItem (OpenGL, whatever texture thing) and then only move these around. Thus, moving the legend and the infobox would basically not use the CPU at all, only the GPU. I believe that this is how QtQuick UIs are supposed to work...? It could of course be that I'm wrong - this whole thing is so weird that it's not obvious how things should be done. It could also be that the QQuickItem primitives are so ... primitive ... that that's not a viable way anyway.

If it works, we should also try convert the profile to QtQuick. One problem that I found mentioned while briefly browsing the web is that this might be incompatible with printing though.

In any case, I will only look at this next week, because I'm a bit fed up with the topic at the moment. I will try to do some comboboxes for demo reasons, but that's it for a few days.

dirkhh:
> > Had a quick break between meetings - I think the way this is designed to be done is to have a QQuickPaintedItem where you implement your own paint() function that updates the visual when needed.
> 
> Yes, I've seen that in the profile code, and I think essentially this does the same thing (render onto a texture and show that).

The profile code today does NOT use a `QQuickPaintedItem`. It uses the print functionality to literally print a static picture of the profile and then simply scales and moves that static image around. Good enough but also... ugh.

> However - and I was not clear enough - this is precisely what I do _not_ want to do.

I understood that :-)

> The idea is to render every chart element (axes, legend, infobox, etc) into its own QQuickItem (OpenGL, whatever texture thing) and then only move these around. Thus, moving the legend and the infobox would basically not use the CPU at all, only the GPU. I believe that this is how QtQuick UIs are supposed to work...? It could of course be that I'm wrong - this whole thing is so weird that it's not obvious how things should be done. It could also be that the QQuickItem primitives are so ... primitive ... that that's not a viable way anyway.

I think what I was trying to point out is that QML has that `QQuickPaintedItem` as intermediate step - better that just rendering a static pixmap/image, but not as complex as breaking it all down into individual QQuickItems.
I personally don't know what the right answer is  but will be happy to help with experimentation along those lines.
 
> If it works, we should also try convert the profile to QtQuick. One problem that I found mentioned while briefly browsing the web is that this might be incompatible with printing though.

I don't think so - the desktop code can do essentially what it does today - render the widget (in that case the QtQuick widget) and then print that rendered version.

> In any case, I will only look at this next week, because I'm a bit fed up with the topic at the moment. I will try to do some comboboxes for demo reasons, but that's it for a few days.

No worries. We still haven't seen much feedback on what we have, so I'll try to poke at the mailing list some more.

What I would appreciate is if you have the energy to poke at the open items in #3127 -- I thought this was really close to being ready to merge, and it gates my mobile / deco code (as I want to benefit from all the great improvements that you have in that PR)

bstoeger:
> The profile code today does NOT use a QQuickPaintedItem.

Ok...?

```
class QMLProfile : public QQuickPaintedItem
{
        ...
}
```

> I don't think so - the desktop code can do essentially what it does today - render the widget (in that case the QtQuick widget) and then print that rendered version.

The profile widget renders into a QGraphicsScene, which is shown in a QGraphicsView. The latter can render onto images. However, this is fundamentally incompatible with QtQuick, unless you want to do the ugly thing we do now.

If you want to implement the planner without eating all the battery, turning this into a QtQuick widget is probably the way to go.

dirkhh:
> > The profile code today does NOT use a QQuickPaintedItem.
> 
> Ok...?

Clearly I have no idea what I'm talking about?

I completely misunderstood how this worked - and while I have spent way too much time looking at the `.cpp` file, I never noticed this in the `.h` file?

> > I don't think so - the desktop code can do essentially what it does today - render the widget (in that case the QtQuick widget) and then print that rendered version.
> 
> The profile widget renders into a QGraphicsScene, which is shown in a QGraphicsView. The latter can render onto images. However, this is fundamentally incompatible with QtQuick, unless you want to do the ugly thing we do now.

Yeah, I clearly need to go back and read some more about how this works. I must be seriously confused here.
What else is new.
 
> If you want to implement the planner without eating all the battery, turning this into a QtQuick widget is probably the way to go.

planner?

bstoeger:
> planner?

Yes, the planner-mode of the profile. I thought that was one of the long term goals?

dirkhh:
I don't see us ever supporting the 'move the handles around to draw a profile' mode on mobile. That just doesn't seem realistic - that's not how mobile interfaces work - it's super painful with a mouse.

The interactiveness of 'I make changes in the text table and see updates on the graph' seems more realistic, especially on a tablet. But that similarly has a much lower rate of updates.

So maybe I'm missing your point again? I appear to be doing a lot of that

dirkhh:
BTW: yesterday afternoon / evening went in a completely different directions at my house, so I haven't had a time to play with this, yet. Will definitely look at it today.

dirkhh:
Just for my understanding... the comment "_it flickers, tears, has antialiasing artifacts_" -- I should expect **NOT** to see that on the desktop, only on mobile, correct?
Because for me this looks basically perfect on the desktop and, ignoring the rather, err, unfortunate UI implementation, still extremely good on mobile-on-desktop... but I haven't tried it on device, yet...

bstoeger:
> Just for my understanding... the comment "it flickers, tears, has antialiasing artifacts" -- I should expect NOT to see that on the desktop, only on mobile, correct?

I observe some flickering when resizing the desktop window and at times what I believe to be antialiasing artifacts, where ticks are not perfectly equal-sized. But yes, the comment was perhaps a bit over-dramatic.

bstoeger:
> I don't see us ever supporting the 'move the handles around to draw a profile' mode on mobile. That just doesn't seem realistic - that's not how mobile interfaces work - it's super painful with a mouse.
> 
> The interactiveness of 'I make changes in the text table and see updates on the graph' seems more realistic, especially on a tablet. But that similarly has a much lower rate of updates.
> 
> So maybe I'm missing your point again? I appear to be doing a lot of that

You were right, I was talking about directly moving the handles on mobile. The students persuaded me to get one of those tablets with a stylus for the online lecture. I feel that  the interactive planner would work even better than with a mouse - if done right. However, it would be an incredible amount of tedious work (QtQuick is not fun to work with) for a feature that I never use. Therefore it is probably not worth it for now.

What could however be useful though, is ro render the static parts of the profile (those which will also be printed) using a QGraphicsScene and the moving parts (the information box, cross, ruler) using these weird OpenGL node things. That would give an GPU accelerated interactive profile while retaining most of the code and also the ability to print.

dirkhh:
> You were right, I was talking about directly moving the handles on mobile. The students persuaded me to get one of those tablets with a stylus for the online lecture.

Those kids - nothing but trouble.

> I feel that the interactive planner would work even better than with a mouse - if done right. However, it would be an incredible amount of tedious work (QtQuick is not fun to work with) for a feature that I never use. Therefore it is probably not worth it for now.

Yes, that would be a really cool experience - for 0.0x% of our users... I don't think that's where we should spend our effort.
I am having a TON of fun with this branch on mobile. Will push my updates soon. This is completely amazing, IMHO.
Its usability far far surpasses what you indicated here.
 
> What could however be useful though, is ro render the static parts of the profile (those which will also be printed) using a QGraphicsScene and the moving parts (the information box, cross, ruler) using these weird OpenGL node things. That would give an GPU accelerated interactive profile while retaining most of the code and also the ability to print.

Yes, there is so much we could do better with the profile. Interactivity of the profile on mobile has been a common request for years. Baby steps. Let's get the statistics right. Let's get Dirk to the point where he understands how to control column width on mobile so the statistics experience is a good one. I'm working on that as we speak.

tcanabrava:
porting the profile to opengl is an awesome idea.
(I'm trying to get back to subsurface, slowly, and one of the things I'm
doing on my spare time is learning opengl. maybe I try that)

dirkhh:
Please try this, @bstoeger -- it creates perfectly reasonable statistics for me on mobile.
I am absolutely stunned.
We still need to implement the chart selection, there are a few other UI issues, but, err, yeah. This is already totally usable.
Caveat: haven't tried this on device, yet. That's my next step.

dirkhh:
Ha, we can't quite remove the dependency on QtCharts... QScatterSeries is part of QtCharts :-)

bstoeger:
Artifact - I'm not force pushing, since you seem to be working on this.

dirkhh:
> Artifact - I'm not force pushing, since you seem to be working on this.

yeah, finding quite a few more of those... cleaning up right now indeed :-)

dirkhh:
Fun. Now let's see if this all builds and if I can test on Android :)

dirkhh:
Very preliminary testing on a 5 year old phone with small screen shows that as usual we have problems with getting the column widths right - my guess is that I need to break down and actually figure out a better way to really fix this instead of just always trying to hack around this. Fun times.

But the more interesting result is that the charts look amazing and all of this mostly works. Which just blows my mind.

Well done, @bstoeger 
Very well done

bstoeger:
> porting the profile to opengl is an awesome idea. (I'm trying to get back to subsurface, slowly, and one of the things I'm doing on my spare time is learning opengl. maybe I try that)

Whatever floats your boat. :) Please note that I started making the profile reentrant (#3127) and I guess this will take ca. 2 more weeks. To avoid that we step on our toes, perhaps let me finish that first. If you are in a hurry, let me know and I will prioritize this up.

bstoeger:
> But the more interesting result is that the charts look amazing and all of this mostly works.

Note that we still have a looong way to got. That full re-rendering on every mouse-hover is not acceptable. Perhaps I find time on the week end to make the QML combo-boxes work properly (can the currentindex be connected to a property?) and connect to the chart-list model. However, I'll leave styling of the "delegate" to you.

dirkhh:
mouse hover isn't a thing on mobile - yes, it's a thing on mobile-on-desktop, but to the best of my knowledge QML does not deliver hover events on Android or iOS.
I __think__ hover might be supported with a pen - I'll need to check that on my iPad Pro.
I'm not saying that I'm ready to cut a release. I'm saying I'm impressed how usable this already is.
And as usual, I will ram my head against the concrete blocks that are 'overall QML UI' while you work on getting the backend right... so yeah, this is a long ways away from shipping.

In the meantime we have actually gotten some indication that people are using the statistics on the desktop and the feedback overall has been reasonably positive, I think.

As for the combo boxes - I believe those always must use a model. You can get access to the currentIndex via a property-alias, if that's sufficient for your use case.

The current code already has our existing (lame) combo box styling. That really could use some upgrade.
But first I need to fix the whole width / columnWidth mess that we are in right now

dirkhh:
So the UI seems semi-useful now for mobile. Do you ever test on device, @bstoeger ?

Something like this is fairly realistic (and allows you to test QML changes without recompiling, just restarting, as it picks up QML from the file system instead of the qrc:); portrait on my Pixel 5 is about 
```
SUBSURFACE_MOBILE_WIDTH=393 SUBSURFACE_MOBILE_HEIGHT=750 ./subsurface-mobile --testqml=../mobile-widgets/qml
```
 landscape is
```
SUBSURFACE_MOBILE_WIDTH=750 SUBSURFACE_MOBILE_HEIGHT=360 ./subsurface-mobile --testqml=../mobile-widgets/qml
```

I need to test the screen size code on all of my other phones (I have, err, a few) and make sure that this didn't break anything on iOS, either. So this will take a while. But from my perspective I am at a stopping point here.
I sent you a screen shot in email that seems to indicate something going weird with grid and data points. I can reproduce this with mobile-on-desktop as well, see below.
And I still have the issue that when I rotate the screen while statistics are shown, the statistics aren't redrawn. I'm sure that's easy to fix and is just one of those "duh" moments. Because any change to a drop down is enough to 'fix' it.
Screen Shot 2021-01-09 at 10 10 53 AM

bstoeger:
> mouse hover isn't a thing on mobile - yes, it's a thing on mobile-on-desktop, but to the best of my knowledge QML does not deliver hover events on Android or iOS.

Ok, but this is re-render the whole scene on hover is also unacceptable on desktop.

I'll try to look at your changes tomorrow.

dirkhh:
So this works reasonably well now on mobile, the layout is not terrible, it responds correctly to rotation, and is otherwise fairly reasonable. I still need to implement the choice of chart - my first attempt failed badly.
The issue with the grid vs. graph remains, though. I have not tried to track that down, focusing on QML things instead.

dirkhh:
I think I have a reasonably clever idea how to deal with the chart type on mobile, but that requires some more complex QML programming. Getting the charts list into QML isn't hard - but having a fifth dropdown really makes the UI too painful on the small screen of a phone. So I'm trying to move this into the context menu. But generating the actions for a dynamic list of strings is so far "just out of reach"... it feels like I'm close, though. I'll try some more tomorrow

bstoeger:
Ouch... I don't know how that happened, but I just overwrote master with this branch!!

I hope I have fixed it by going back to what I think was the original master head and force pushed that.

Sorry..!

bstoeger:
OK, apart from that shocker, the mobile code now correctly sets the indexes and there is an "operation" combo box for selecting the variable 2 operation.

Note that mobile currently eats 100% CPU on the chart-page, so that really has to be fixed.

I'm a bit torn on when to merge this into master. Probably rather early, to avoid diverting branches if @atdotde starts to add more UI refinements.

dirkhh:
> Ouch... I don't know how that happened, but I just overwrote master with this branch!!

Sweet - `push -f` can be dangerous :-)

> I hope I have fixed it by going back to what I think was the original master head and force pushed that.

A quick check seems to indicate that indeed you went back to where it was last night.

dirkhh:
> OK, apart from that shocker, the mobile code now correctly sets the indexes and there is an "operation" combo box for selecting the variable 2 operation.

I'll look at that in a moment

> Note that mobile currently eats 100% CPU on the chart-page, so that really has to be fixed.

It didn't for me yesterday. I wonder what changed.

> I'm a bit torn on when to merge this into master. Probably rather early, to avoid diverting branches if @atdotde starts to add more UI refinements.

Well, you almost already merged it :-)
In all seriousness, I'm willing to merge it aggressively because this is the direction we are going, we might as well fix things in master. Yesterday (?) you said there was something about the hover implementation on the desktop that you wanted to fix (or maybe I'm misremembering that), but I'll be happy to do another run through this branch to make sure everything looks reasonable (including commit messages) and then merge it.

dirkhh (approved):
There is one commit in there that literally says 'squash me' as part of the commit message.
But other than that I found very little that causes me to say "don't merge".
Do you want me to clean that up and merge?

dirkhh:
Ha, sorry, I responded before finishing the build for on device testing. I also need to fix your QML code to get the layout right :-)

bstoeger:
> Do you want me to clean that up and merge?

Fine with me, though perhaps we should warn people of regressions to be expected.

Personally I would be more aggressive about squashing, but really up to you.

dirkhh:
OK, force push incoming that rather drastically squashes this into just a hand full of commits, one of them ended up being shared things that were created from commits you did and commits I did.

dirkhh:
OK, coming back to this one. I think this should go in - I'll send a caveat email to the mailing list.

PR #3145, opened 2021-01-09
statistics: add dive-# and mean-depth variables.

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Adds two new variables, as requested on the mailling list.

### Changes made:

1) Add dive-# variable.
2) Add mean-depth variable.

### Mentions:

@dirkhh

### DIFF
--- a/stats/statsvariables.cpp
+++ b/stats/statsvariables.cpp
@@ -930,11 +930,18 @@ struct DateVariable : public StatsVariableTemplate<StatsVariable::Type::Continuo
 	}
 };
 
-// ============ Dive depth, binned in 5, 10, 20 m or 15, 30, 60 ft bins ============
+// ============ Dive depth (max and mean), binned in 5, 10, 20 m or 15, 30, 60 ft bins ============
+
+static int dive_depth_in_mm(const dive *d, bool mean)
+{
+	return mean ? d->meandepth.mm : d->maxdepth.mm;
+}
 
 struct DepthBinner : public IntRangeBinner<DepthBinner, IntBin> {
 	bool metric;
-	DepthBinner(int bin_size, bool metric) : IntRangeBinner(bin_size), metric(metric)
+	bool mean;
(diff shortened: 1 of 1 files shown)


stats/statsvariables.cpp
```diff
@@ -1270,6 +1302,40 @@ struct WeightVariable : public StatsVariableTemplate<StatsVariable::Type::Numeri
...
+	}
+	int to_bin_value(const dive *d) const {
+		return d->number / bin_size;
+	}
+};
+
+static DiveNrBinner dive_nr_binner_10(10);
+static DiveNrBinner dive_nr_binner_50(50);
+static DiveNrBinner dive_nr_binner_100(100);
+static DiveNrBinner dive_nr_binner_200(200);
```
dirkhh, line 1343: interesting set of bins - because of our privacy policy I don't know what the actual distribution of total number of dives is, but when people give me access to fix things, we seem to have a lot of people in the <200 dives group - so having 5, 10, 20, 50 as bin sizes would seem useful. Having 100 & 200 for the people with tons and tons of dives isn't wrong, but I think they are more of an outlier (yes, one person that asked me for help had dive numbers past 6000, but they, too, had only about 800 dives in Subsurface)

dirkhh (approved):
Good addition. See my comment about bin sizes - I'm happy to merge as is, or if you agree and want to do a quick update, that's fine too

stats/statsvariables.cpp
bstoeger: The list of binners is now different for logs with < 1000 and > 1000 dives. For the latter small bins give horrible results.

Ultimately, the memory management should be changed to allow for dynamically generated bins. Not a big deal, but also not a high priority.

The code is also improved by now starting the bins at 10*n+1, since 0 means no number. Moreover, dives with number 0 are ignored.

PR #3151, opened 2021-01-10
Make sure to sanitize the gasmix after editing it

### LINKED ISSUE #3149: O2 % for Cylinder with 21% is not saved / shown
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
I have some dives where I did enter 21% for the O2 % field in a row in the Cylinder grid. When reopening the dive log from cloud storage (this is after some time) the 21% is missing, it's empty. Also, when I save the log to a local file there is initially no o2 attribute in the xml. But even than after re-entering 21% and saving locally, I can see the o2='21.0%' in the xml, but after restarting Subsurface and reopening this local logbook the O2 percentage shows empty.
I can imagine that 21% is ofcourse some kind of default, but to me there is a big difference between empty meaning 'I dont know the percentage' or meaning 'it's the default, air'.
If I know I dived with air I enter 21%, if I don't know or don't remember any more what it was I leave it empty. An entered value should always be left as-is anyway and not be changed or removed.

### Operating system:
Windows 10

### Subsurface version:
Official 4.9.10

### Steps to reproduce:
- In equipment tab use a Cylinder and enter 21 in the O2 % field.
- Save changes and close Subsurface
- Restart Subsurface and (if needed) reopen the logbook
- Notice that the O2 field is empty

### Current behavior:
An entered value of 21% for Cylinder O2 % is not shown or removed

### Expected behavior:
An entered Cylinder O2 % value should always be remained

torvalds:
Otherwise we end up showing the gasmix in a different form after editing
for the usual air percentages, because we haven't normalized the gasmix
values and store them back in the dive in the non-normalized format.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1241,6 +1241,7 @@ EditCylinder::EditCylinder(int index, cylinder_t cylIn, EditCylinderType typeIn,
 			cyl[i].gasmix = cylIn.gasmix;
 			cyl[i].bestmix_o2 = cylIn.bestmix_o2;
 			cyl[i].bestmix_he = cylIn.bestmix_he;
+			sanitize_gasmix(&cyl[i].gasmix);
 			break;
 		}
 	}


bstoeger (approved):
That looks correct to me. It sanitizes user input on creation of the command-object.

I suggest to merge if CI passes.

PR #3156, opened 2021-01-11
call invalidate_dive_cache() when editing weights

### LINKED ISSUE #3150: Weight system not saved in cloud storage
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
So this is really weird. I have some older dives where I know for sure I entered the weight system, type and weight. When I now look at these dives the weight type is missing, it's empty. Even when I select again a type, save the change to cloud, it is not preserved. After restarting Subsurface it's empty again!
This also happens with trying to change a non-empty weight (type or kg), or adding a weight system, it is not preserved. After restarting Subsurface the old value is back (or not preserved at all).
This only seems to happen when saving and retrieving data from cloud storage. When I use a local logbook the type is saved and preserved after restarting.
This is really annoying because it seems like a lot of my dives of which I am 100% sure I entered the weight type (literally 100's and 100's of dives) are now missing the weight type.

### Operating system:
Windows 10

### Subsurface version:
Official 4.9.10

### Steps to reproduce:
- For a dive with an existing weight system, change the weight type
- Save the change to the cloud storage and close Subsurface
- Restart Subsurface and notice the weigth type of that dive is unchanged

### Current behavior:
Weight system is not preserved on saving to the cloud storage

### Expected behavior:
Weight system should be saved in the cloud storage

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Bug fix for #3150: call invalidate_dive_cache() at the appropriate places.

This also removes redundant invalidate_dive_cache() calls outside of the undo command. By moving them all to the undo machinery, the problem becomes more localized making it easier to come up with a general solution: Only ever store / pass around const-pointers to dives. To write to a dive, force a call to `get_mutable_dive`, which invalidates the dive cache.

### Changes made:

1) Add invalidate_dive_cache() where needed.
2) Remove invalidate_dive_cache() where redundant.

### Related issues:

Fixes #3150.

### Release note:

Done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- undo: save to git after editing weights [#3159]
 - desktop: complete rewrite of the statistics code, significantly expanding capabilities
 - desktop: add preferences option to disable default cylinder types
 - mobile: add ability to show fundamentally the same statistics as on the desktop
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -838,6 +838,7 @@ void ReplanDive::undo()
 	std::swap(d->duration, duration);
 	std::swap(d->salinity, salinity);
 	fixup_dive(d);
+	invalidate_dive_cache(d); // Ensure that dive is written in git_save()
 
 	QVector<dive *> divesToNotify = { d };
 	// Note that we have to emit cylindersReset before divesChanged, because the divesChanged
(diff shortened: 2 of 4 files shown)


dirkhh (approved):
These look good.
I keep wondering if there is a programatic way that we can test that every single change that we can make in the software also calls `invalidate_dive_cache()`. But I guess that's just a different way of asking for the `const dive *` solution...

dirkhh:
This seems important enough to justify making a 4.9.11 release (in a branch, I guess, as current master is quite a few weeks from being ready to be released). What do you think?

dirkhh:
merged locally

PR #3158, opened 2021-01-11
undo: set dive mode to CCR in undo command, not profile code

bstoeger:
When setting a CCR setpoint, the profile code(!) would turn
the dive into a CCR dive. Not only should the display layer
not alter dives, this also means that the action is not
undoable.

Move that to the appropriate undo command, where it makes
more sense, but obviously also makes things more complicated.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The divemode was not reset on undo of a setpoint addition, because the divemode was changed in the profile code of all things. This moves this to the undo code which, of course, makes things more complicated.

### Additional information:

FWIW, it worked in my limited testing, though I'm not even totally sure, what this should do.

### Release note:

Obscure, but added.

### DIFF
--- a/commands/command_event.h
+++ b/commands/command_event.h
@@ -5,6 +5,7 @@
 #define COMMAND_EVENT_H
 
 #include "command_base.h"
+#include "core/divemode.h"
 
 // We put everything in a namespace, so that we can shorten names without polluting the global namespace
 namespace Command {
@@ -35,10 +36,11 @@ private:
 class AddEventBase : public EventBase {
 public:
 	AddEventBase(struct dive *d, int dcNr, struct event *ev); // Takes ownership of event!
-private:
-	bool workToBeDone() override;
+protected:
 	void undoit() override;
 	void redoit() override;
+private:
+	bool workToBeDone() override;
 
 	OwningEventPtr eventToAdd;	// for redo
 	event *eventToRemove;		// for undo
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
This is one of those cases where reading the commit itself is insufficient to grasp whether this change can work.
Are we sure that in each cases in which the `profile.c` code would have changed the `divemode` indeed the `AddEventSetpointChange()` was called? I tried to think through the code in QtCreator, following how we get from A to B and I'm reasonably sure. It seems we only ever create such events in the SetpointDialog... but... yeah.

dirkhh:
@atdotde you seem to have slightly more experience with the CCR code (and if not, who among the 'reviewers' does?)

atdotde:
I promise I will look into this. Hopefully later tonight.

commands/command_event.h
```diff
@@ -57,6 +59,10 @@ class AddEventDivemodeSwitch : public AddEventBase {
 class AddEventSetpointChange : public AddEventBase {
 public:
 	AddEventSetpointChange(struct dive *d, int dcNr, int seconds, pressure_t pO2);
+private:
+	divemode_t divemode;	// Wonderful: this may change the divemode of the dive to CCR
```
atdotde, line 63: It does because historically, there was no divemode, a dive was a CCR dive by definition if it had an event setting a non-zero set-point anywhere.

atdotde (approved):
I guess this is ok, but I have to admit I am not au courant with the whole undo/redo/command control flow.

dirkhh:
OK, merging this locally (there isn't really a conflict, but whatever)

PR #3160, opened 2021-01-11
[cleanup] mostly constification and other profile code straightening

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is the last preparatory PR before making the profile reentrant. Mostly constification. Some people reallly seem to dislike const. Not much to say, really. Ultimately this should lead to a dynamic profile on mobile (we then can have multiple profile views at the same time) and less insane printing.

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -623,7 +623,7 @@ void ProfileWidget2::plotDive(const struct dive *d, bool force, bool doClearPict
 		maxdepth = newMaxDepth;
 	}
 
-	dataModel->setDive(&displayed_dive, plotInfo);
+	dataModel->setDive(plotInfo);
 #ifndef SUBSURFACE_MOBILE
 	toolTipItem->setPlotInfo(plotInfo);
 #endif
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -1321,7 +1321,7 @@ void ProfileWidget2::setPlanState()
 }
 #endif
 
-bool ProfileWidget2::isPlanner()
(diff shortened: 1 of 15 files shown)


profile-widget/profilewidget2.cpp
```diff
@@ -575,14 +575,7 @@ void ProfileWidget2::plotDive(const struct dive *d, bool force, bool doClearPict
 	if (currentState == EMPTY)
 		setProfileState();
 
-	// next get the dive computer structure - if there are no samples
-	// let's create a fake profile that's somewhat reasonable for the
-	// data that we have
 	struct divecomputer *currentdc = select_dc(&displayed_dive);
-	Q_ASSERT(currentdc);
-	if (!currentdc || !currentdc->samples)
-		fake_dc(currentdc);
```
dirkhh, line 586: would it make sense to print out a `qWarning()` if there aren't any samples?
I'm sure you checked all the paths, but still... or am I too paranoid?

dirkhh (commented):
all the constifications (is that a word??) are great. I am mildly worried about the `fake_dc()` thingy... what are your thoughts on my suggestion?

bstoeger:
Forget this PR for now - it is broken.

bstoeger:
Fixed the problem - time to reopen.

dirkhh:
Interestingly I don't see a new commit pushed... so the problem wasn't in the code?

dirkhh (approved):
I continue to think this is great.
The one fairly invasive change is of course the initial phase of dive planning - but if that were to break something that should be pretty obvious, and in my testing it seemed to work fine.

PR #3159, opened 2021-01-11
Mobile UI tuning

dirkhh:
### Describe the pull request:

- [x] Bug fix
- [x] Functional change
- [x] Code cleanup

### Pull request long description:

- fixes a crash in the mobile statistics code
- changes most of the drop down combo boxes to be "more modern looking"
- various small layout fixes to make better use of the space that's available
- crops the header image in the global drawer on small displays in landscape mode (which I think more people will be using for the statistics)

PR #3161, opened 2021-01-11
Mobile chart selection

dirkhh:

- [x] New feature

Add the ability to pick the chart used to the mobile UI

### Changes made:

1) create a helper in the `StatsState` to get a list of usable charts - prefer `Good`, use `Undesirable` if there are none
2) encode the magic chart ID in that list
3) make all this available to the QML code via the `StatsManager` and allow the selection of the chart via magic ID
4) dynamically create a list of actions at run time that are shown in the context menu of the stats page and allow the user to switch charts

### Related issues:

This includes #3159 almost by mistake - but the UI looks like crap without that one

### Documentation change:

clearly this need to be captured in the as yet to be written manual

### Mentions:

@bstoeger can I get you to definitely take a look at the two C++ helpers?

### DIFF
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -386,17 +386,15 @@ Kirigami.ScrollablePage {
 			anchors.right: parent.right
 			anchors.leftMargin: Kirigami.Units.gridUnit / 2
 			anchors.rightMargin: Kirigami.Units.gridUnit / 2
-			TemplateComboBox {
+			TemplateSlimComboBox {
 				visible: filterBar.height === sitefilter.implicitHeight
 				id: sitefilterMode
-				editable: false
 				model: ListModel {
 					ListElement {text: qsTr("Fulltext")}
 					ListElement {text: qsTr("People")}
 					ListElement {text: qsTr("Tags")}
 				}
 				font.pointSize: subsurfaceTheme.smallPointSize
-				Layout.preferredWidth: parent.width * 0.2
 				Layout.maximumWidth: parent.width * 0.3
 				onActivated:  {
 					manager.setFilter(sitefilter.text, currentIndex)
--- a/mobile-widgets/qml/DiveSummary.qml
+++ b/mobile-widgets/qml/DiveSummary.qml
@@ -12,7 +12,7 @@ Kirigami.ScrollablePage {
 	DiveSummaryModel { id: summaryModel }
 	property string firstDive: ""
 	property string lastDive: ""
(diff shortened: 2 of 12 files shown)


dirkhh:
Haha... this one was fun...
For someone who has no idea what he's doing with QML or JS or much of anything I'm kinda surprised that I got this to work. I almost gave up on the dynamically created list of actions... but now that I figured it out I'm sorta proud of it :-)

dirkhh:
Spent way too much time trying to add a way to toggle the legend on and off. Problem is, for that I need to be able to tell if the current chart can have a legend, and be able to turn it on and off. And right now both information is in the same variable. Yes, I can set `state.legend` to false and the legend goes away - but now I've lost the information if that chart was even able to have a legend (and therefore if I should even show that toggle). I tried to extract that from the chart features, but those are all private to `statsview.cpp` - I then hacked all that together and the result looked like a tank rolled through @bstoeger 's beautiful code.

So I decided to trash it all and to instead say "_umm, hey Berthold... wouldn't it be cool if the StatsManager could tell me if the current chart supports having a legend, and gave me a QML invokable method to toggle the legend?_" -- I'm betting it'll take you 5% of the time I spent on this and it will look beautiful in the code :-)

bstoeger:
I'm not going to interfere with your QML, but this is not at all how I planned to do this.

Both, desktop and QML were supposed to access the ChartListModel, which provides the icons, sorting and all that stuff. At least the list of charts should be constructed from that model and not via a new function, which duplicates code. I will see what I can do.

bstoeger:
But this will take some time, as I will have to get up to speed with QML - and you know how much I enjoy that. So don't hold your breath.

dirkhh:
I understood that - the problem was that I couldn't get that to work with the context menu - and I really like this as a UI metaphor for this action. I can play with this some more in order to get the icons as well, but if you look at the magic that is used to create the array of actions that go into context menu, you may be able to appreciate that this one is kinda different.

bstoeger:
That contextualAction thing looks very weak to me (no headings?). What about doing a `ListView` in a `Popup`?

dirkhh:
> That contextualAction thing looks very weak to me (no headings?). What about doing a `ListView` in a `Popup`?

Ha - the plight of UI design... I was super excited about it, you clearly don't like it.

The design idea of that drawer from the right is that it brings in additional UI elements. A popup would be hideous. But I can't do headers and I can't do a ListView without fundamentally rewriting the way the context drawer is done - it simply assumes that you give it an array of actions. So I'll need to figure out a way to replace that whole content item with a ListView.

Back to the drawing board, I guess :-)

PR #3162, opened 2021-01-12
Export ChartListModel to QML

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Make the ChartListModel accessible from QML. I have no idea if this should be done this way, but it appears to work. This is obviously only a proof-of-concept. I didn't bother to fight with QML.

### Mentions:

@dirkhh: FYI - I think something like this is the way forward, though I understand that making this functional will be extremely painful.

### DIFF
--- a/mobile-widgets/qml/StatisticsPage.qml
+++ b/mobile-widgets/qml/StatisticsPage.qml
@@ -19,6 +19,9 @@ Kirigami.Page {
 	StatsManager {
 		id: statsManager
 	}
+	ChartListModel {
+		id: chartListModel
+	}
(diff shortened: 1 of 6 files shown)


dirkhh:
Thanks for doing that, I was about halfway there but had forgotten about the stupid hash of names... 

I may have some ideas to slightly modify the UI, but it is basically on par with everything else that we do UI wise, so there is that

bstoeger:
superseded by #3163

PR #3163, opened 2021-01-12
Mobile chart selection

dirkhh:

- [x] Functional change
- [x] New feature

With all due respect to @bstoeger and #3162 - I think this is maybe slightly more reasonable

I gave up on doing this in the context drawer since Berthold thought this was a terrible idea, instead I took some liberty with his idea of a popup driven by a button, including actually showing the icons that he so carefully exported, styling the whole thing, placing it in the right spot in the layout, etc...

### DIFF
--- a/mobile-widgets/statsmanager.cpp
+++ b/mobile-widgets/statsmanager.cpp
@@ -3,10 +3,6 @@
 
 StatsManager::StatsManager() : view(nullptr)
 {
-	// Test: show some random data. Let's see what happens.
-	state.var1Changed(2);
-	state.var2Changed(3);
-	state.binner2Changed(2);
 	updateUi();
 }
 
--- a/mobile-widgets/statsmanager.cpp
+++ b/mobile-widgets/statsmanager.cpp
@@ -1,7 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "statsmanager.h"
+#include "stats/chartlistmodel.h"
 
-StatsManager::StatsManager() : view(nullptr)
+StatsManager::StatsManager() : view(nullptr), charts(nullptr)
 {
 	updateUi();
 }
@@ -10,18 +11,19 @@ StatsManager::~StatsManager()
 {
 }
 
-void StatsManager::init(StatsView *v, QObject *o)
+void StatsManager::init(StatsView *v, ChartListModel *m)
 {
 	if (!v)
 		fprintf(stderr, "StatsManager::init(): no StatsView - statistics will not work.\n");
+	if (!m)
+		fprintf(stderr, "StatsManager::init(): no ChartListModel - statistics will not work.\n");
 	view = v;
+	charts = m;
 }
 
 void StatsManager::doit()
 {
-	if (!view)
-		return;
-	view->plot(state);
+	updateUi();
 }
 
 static void setVariableList(const StatsState::VariableList &list, QStringList &stringList, int &idx)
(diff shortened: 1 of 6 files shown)


mobile-widgets/statsmanager.cpp
```diff
@@ -1,31 +1,33 @@
...
 
-StatsManager::StatsManager() : view(nullptr)
+StatsManager::StatsManager() : view(nullptr), charts(nullptr)
 {
-	// Test: show some random data. Let's see what happens.
-	state.var1Changed(2);
-	state.var2Changed(3);
-	state.binner2Changed(2);
+	// Default chart is dives per year
+	state.var1Changed(0);
```
bstoeger: If you want the default, you can remove these three lines.

bstoeger (approved):
Yes, looks much better.

There obviously is room for improvement (centring, resize the popup according to content), but it's a good start.

mobile-widgets/statsmanager.cpp
dirkhh: ahh, didn't realize that. will do

dirkhh:
> Yes, looks much better.
> 
> There obviously is room for improvement (centring, resize the popup according to content), but it's a good start.

It turns out that QML makes it hard to dynamically size the popup. Which is of course rather stupid, but, well, it's QML...
I can poke at this some more.
Right now I am working on fixing some other annoying issues with the QML UI - we are approaching the cross over point... roughly the same number of people have installed the recent iOS and Android updates as the total number of desktop users for 4.9.5 and higher... (admittedly that latter number could be under-counting with people who don't connect to our servers after an update). As a result I think it makes sense to pay more attention to things in the mobile UI that suck...

dirkhh:
Ok, this is looking better. Please check my addition of that property - I couldn't make it work without adding that.

bstoeger:
I believe the "correct" way to do this is to create a QML Model based on `QQMLInstanceModel`, which (of course!) is a private implementation detail of QML. Gee, thanks!

I would not another member variable, since this is redundant information and just risks getting inconsistent. Instead try something like
```
	Q_PROPERTY(int count READ rowCount NOTIFY countChanged)
```
(completely untested)

dirkhh:
and how would that `countChanged()` signal fire? The same way as it fires today. Which means nothing changes from an __inconsistency__ point of view.
In reality this is really an implementation detail, but I'm happy to change accordingly.

bstoeger:
Unless the code grows and we have to handle dynamic addition/removal of rows and forget to update count. Poof - `count != rowCount()`.

Of course, with a seasoned C programmer I would argue that it spares us 4 bytes of RAM. 😁

dirkhh:
Please look at the modified implementation I just pushed

bstoeger:
I think you forgot to commit, because nothing changed...?

dirkhh:
Oh GitHub. Pushed again, now it seems to have accepted the force push. The interesting commit is 465a505

bstoeger:
Sigh. I forgot that rowCount() takes a parameter. Still better than redundant state, IMHO.

dirkhh:
> Sigh. I forgot that rowCount() takes a parameter. Still better than redundant state, IMHO.

Yeah, I see your point there. I didn't even realize this was possible.
I'm learning more from you every day, I guess :-)

Are you ok with merging this?

bstoeger:
Sure, go ahead...

PR #3165, opened 2021-01-12
Statistics: Convert QGraphicsScene to QSG

bstoeger:
In order not to waste CPU by constantly rerendering the chart,
we must use these weird OpenGL QSGNode things. The interface
is appallingly low-level and unfriendly.

As a first test, try to convert the legend. Create a wrapper
class that represents a rectangular item with a texture
and that will certainly need some (lots of) optimization.

From that derive a class that draws a rounded rectangle
for every resize. The child class of that must then paint
on the rectangle after every resize.

That looks all not very fortunate, but it displays a
legend and will make it possible to move the legend
without and drawing operations, only shifting around
an OpenGL surface.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is my first experiment with drawing the legend as a QSGNode. The QSG* interface is horrible (not in a crazy QML sense, but in a very unfriendly and rudimentary sense).

### Changes made:

1) Add experimental wrapper class around QSGNode.
2) Implement drawing the legend on such a node.


bstoeger:
Refactored the thing to only access low-level QSG nodes in the rendering thread as per documentation.

This currently does no memory management, which means that chart items cannot be removed from the scene individually, only all at once.

Implemented moving of the legend with the mouse. Please test.

bstoeger:
Transformed the info-box and added code taking care of z-Values, so that the Infobox is below the legend, even if added later.

dirkhh:
I'm curious about the OpenGL story here... I thought more and more platforms are abandoning that, and Qt6 is moving away from it as well...

dirkhh:
Playing with this for a little, it looks fine.
As a bike shedding comment - would it make sense to allow pushing the legend out across the chart border so that it is clipped and only partially visible?
I did one path through the code and need to spend more time with it to fully understand the complexities, but it's definitely an encouraging start.
Have you looked at how this affects CPU load? I remember you saying that this was part of what you felt you needed to do to address that...

bstoeger:
> I'm curious about the OpenGL story here... I thought more and more platforms are abandoning that, and Qt6 is moving away from it as well...

Always read "OpenGL" as "OpenGL or whatever the QSG thing uses". I wish they would spend more time on making the API less horrible than replacing the rendering technology again.

> Have you looked at how this affects CPU load? I remember you saying that this was part of what you felt you needed to do to address that...

For now this doesn't change anything, because the scene is still rerendered on every update.

dirkhh:
I wish they would continue to maintain Qt 5.15 in open source. My comments on that are... not polite.
That said, yeah, even after a second round reading the code I still have a few questions that I need to try to understand... so not quite ready to approve, yet.

stats/statsview.cpp
```diff
@@ -80,17 +80,40 @@ void StatsView::mouseReleaseEvent(QMouseEvent *)
...
+}
+
 QSGNode *StatsView::updatePaintNode(QSGNode *oldNode, QQuickItem::UpdatePaintNodeData *)
 {
 	// The QtQuick drawing interface is utterly bizzare with a distinct 1980ies-style memory management.
 	// This is just a copy of what is found in Qt's documentation.
-	QSGImageNode *n = static_cast<QSGImageNode *>(oldNode);
-	if (!n)
-		n = rootNode = window()->createImageNode();
+	RootNode *n = static_cast<RootNode *>(oldNode);
```
dirkhh, line 115: that whole thing is... something. Well designed, Qt people. Well designed.

stats/statsview.cpp
```diff
@@ -80,17 +80,40 @@ void StatsView::mouseReleaseEvent(QMouseEvent *)
...
+	if (!n) {
+		n = rootNode = new RootNode;
+		n->imageNode = window()->createImageNode();
+		n->zNodes[(int)ChartZValue::Series]->appendChildNode(n->imageNode);
```
dirkhh: so the nature of 'z' is simply a side effect of the order of these nodes, not something that the Qt code actually understands as a concept - or am I misunderstanding how this is supposed to work?

stats/informationbox.cpp
```diff
@@ -11,74 +12,65 @@ static const int informationBorder = 2;
...
+
+	ChartRectItem::resize(QSizeF(width, height));
 
-	// Setting the position will also set the proper size
-	setPos(pos);
+	painter->setPen(QPen(darkLabelColor)); // QPainter uses QPen to set text color!
+	double y = 2.0 * informationBorder;
+	for (size_t i = 0; i < widths.size(); ++i) {
+		QRectF rect(2.0 * informationBorder, y, widths[i], fontHeight);
+		painter->drawText(rect, text[i]);
```
dirkhh, line 41: ah, so I misunderstood what this was actually doing. It's not actually doing OpenGL - it still uses the same drawing primitives, etc. I was confused there, clearly.
This makes a lot more sense.

stats/statsview.cpp
```diff
@@ -55,6 +56,30 @@ StatsView::~StatsView()
 {
 }
 
+void StatsView::mousePressEvent(QMouseEvent *event)
+{
+	// Currently, we only support dragging of the legend. If other objects
+	// should be made draggable, this needs to be generalized.
```
dirkhh, line 62: that works well on the desktop - I need to test this on mobile. I think drag behaves the same, but I have learned to assume nothing...

dirkhh (commented):
OK, I think I finally understand what's actually happening here. Took me three runs through. I guess I'm just slow.
A few comments and questions.

stats/statsview.cpp
bstoeger: According to the docs, this is designed to be fast, not easy to write for. I fear that the opposite is true: actual applications will be rather inefficient, because it is very hard to get right.

stats/statsview.cpp
bstoeger: This is right. We have n top-level nodes, which are rendered according to creation time and to these we add the child nodes at the various z-levels. The effect is that nodes with a lower z-value are rendered below nodes with a higher z-value by the shader. Inside a z-level the nodes are rendered according to creation time. Therefore, this simulates setZValue of QGraphicsScene.

stats/statsview.cpp
dirkhh: I don't doubt that it is __designed__ to be fast. Some of the folks in the Qt team are focused really on the embedded space and there performance on relatively low end silicon is everything.

stats/informationbox.cpp
bstoeger: Actually it does use OpenGL (or whatever is the shader language du-jour).

I paints once onto a QImage with the usual QPainer API, then transforms this into a texture and that is blitted with OpenGL when the box is moved.

For the regression line, etc. I'm not yet sure if we also should use textures or directly use OpenGL (or whatever) primitives.

stats/statsview.cpp
dirkhh: Yes, that's what I figured. So instead of an 'actual' Z value implementation in the Qt infrastructure we are simulating Z value using that 'fast infrastructure' in our implementation. Which is fine, and I like the naming that you put in place there.

stats/informationbox.cpp
dirkhh: But what I mean is that OpenGL is an implementation detail that we hopefully can rely on being handled in the background. We don't need to write OpenGL code. And this will still work on platforms that are abandoning OpenGL (as macOS is).
I really hope we will not be sucked into writing OpenGL code here - because there be dragons.

dirkhh:
OK, just for my planning - what's next here? It's labeled as 'EXPERIMENTAL', so I assume this isn't something I should merge just yet...

bstoeger:
Failed experiment: using a Texture to blit the axes looks ugly, because the lines drawn with QPainter are rasterized differently than the lines of the QSG thing and the ticks and the grid now don't match. I'll change this to draw the labels on a pixmap and the ticks using QSG.

bstoeger:
Ok, I like this much better. I'm not sure if this is generally doing the right thing: render everything using QSG (OpenGL, or whatever) graphic primitives or it would be better to render every z-layer into a pixmap and blit than onto the screen.

bstoeger:
Converted the regression line and confidence area.

@atdotde: Could you please check the result? The confidence area appears very pale on my laptop screen and I'm not sure if it is intended that way.

This renders onto a pixmap and blits that onto the screen. We could also use graphics primitives - turning the polygon into a triangle strip is a mildly interesting intellectual exercise. But there are some special cases to consider and therefore I didn't bother. Moreover, I have no idea if it is preferred to paint 202 triangles or just blit a single texture.

atdotde:
The paleness indicates the r^2 value (how much of the variance of the date is explained by the regression), the closer the dots to the regression line the darker it gets (actually it is the alpha channel that is affected). It is paler if it is more like a dot cloud.

I believe turning the inside of a polygon into triangles is a non-trivial problem even if we have the additional information about the general form of the thing and the fact it is concave does not make it easier. Well, maybe it could be done by using rectangles from x to x+dx....

bstoeger:
> The paleness indicates the r^2 value (how much of the variance of the date is explained by the regression), the closer the dots to the regression line the darker it gets (actually it is the alpha channel that is affected). It is paler if it is more like a dot cloud.

Ok, that might be the case in my example data set. I was wondering, because the confidence area was very thin. But now I understand that these are two issues: Apparently I have a good linear correlation with a huge spread.

> I believe turning the inside of a polygon into triangles is a non-trivial problem even if we have the additional information about the general form of the thing and the fact it is concave does not make it easier. Well, maybe it could be done by using rectangles from x to x+dx....

Oh, but in this case it is absolutely trivial, because you calculate in pairs with the same x-value. Effectively you only have feed the points alternately (top, bottom, top, bottom,...) and you automatically get the needed data structure for OpenGL: https://en.wikipedia.org/wiki/Triangle_strip

The only special case to consider is when both, the lower and the upper bound are outside the visible area (to the same side).

bstoeger:
Hrmpf. Converted the BarSeries and now the bright labels on dark bars look terrible. It seems that QPainter does not render properly into transparent backgrounds? Will have to investigate.

bstoeger:
Jeez - the QSG API is killing me. It's just wrong on so many levels. I'll just quote the latest commit message:
```
To replace the QGraphicsScene, we need the possibility of
showing and hiding items.

Turns out, the QSG API is completely insane.

Whether an item should be shown is queried by the virtual
function isSubtreeBlocked(), which is supposed to be
overriden by the derived classes.

However, the common nodes for rectangles and pixmaps are
supposed to be created by QQuickWindow, for hardware
optimization. This gives nodes that cannot be derived
from and therefore whether the item is shown or not cannot
be controlled.

There are therefore two distinct cases to consider: The
node is allocated by the code directly or indirectly by
QQuickWindow.

In the latter case, we use a proxy node with the only
purpose of having a "visible" flag and add the obtained
node as a child.

This madness is performed with template trickery to get
unified code.
```
Please tell me that I'm silly and it is not really that way!

dirkhh:
Hey @thiagomacieira - I don't think this is code that you maintain, but I assume you'd know whom Berthold could talk to about this? It seems somewhat unlikely that this is the intended design...

thiagomacieira:
Yeah, it definitely isn't my area and it definitely is pretty low-level.  Git log says the best people to talk to are Laszlo Agocs and Andy Nichols: https://github.com/qt/qtdeclarative/tree/5.15/src/quick.

Emails are their names AT qt.io.

bstoeger:
The statistics are now fully converted to QSG and there is no more pointless repainting of the scene graph.

Two known issues:
- Light colored labels on dark background look horrible. There is something wrong with drawing text onto transparent background - I will have to investigate.
- The application crashes on exit, because apparently you can't have QSGTextures as static objects. Destruction after the QApplication was destroyed gives crashes. I already observed similar things with the map widget. Mildly infuriating, but we'll probably just have to leak these things.

bstoeger:
Another issue: Mobile crashes in the rendering loop. Sigh.

Edit: The memory management of this PR will have to be redone - the QSGNodes must be deleted on the rendering thread. Sigh.

dirkhh:
> 
> - Light colored labels on dark background look horrible. There is something wrong with drawing text onto transparent background - I will have to investigate.

Back in the olden days there were issues that anti-aliasing worked differently when you painted on the final color vs. when paining on transparent backgrounds. I thought this was long fixed, but what do I know.

> - The application crashes on exit, because apparently you can't have QSGTextures as static objects. Destruction after the QApplication was destroyed gives crashes. I already observed similar things with the map widget. Mildly infuriating, but we'll probably just have to leak these things.

If we leak one set of them at exit, that's fine. If we constantly leak them in every statistic that we draw, that could be more problematic.

As for the crash on mobile - that's a bit of a show stopper :-)

bstoeger:
OMG - I refactored the memory management and this turned reasonable code into a steaming pile of madness. This could be made much more sensible, but currently I completely lack energy and motivation to do that. In my tests it run fine on mobile - please check.

> Back in the olden days there were issues that anti-aliasing worked differently when you painted on the final color vs. when paining on transparent backgrounds. I thought this was long fixed, but what do I know.

I have no idea what I'm doing wrong. The rendering of the text is in `stats/chartitems.cpp` if you want to have a look. I tried different flags from the documentation to no avail. Perhaps we will have to render against the background color of the bar for labels on top of bars / pie slices.

bstoeger:
This now makes the regression line and confidence area opt-in.

@willemferguson, @atdotde: Please check the wording of the options when selecting a scatter-plot.

willemferguson:
It is good as it is. Two small issues:
1) The options could conceivably fit into the Preferences->Display panel.
2) The confidence area needs to relate to a level of confidence, normally given as a fraction (or %). 2 * sigma has no meaning in terms of confidence. So, instead of 2 * sigma I would prefer 1.96 * sigma which is equivalent to a confidence of 0.95 or 95%.

willemferguson:
A question I have about  the colour of the confidence region. I asked this question on the mail list and did not get a reply. 

I suspect that the narrower the confidence region (i.e. closer the edges of the region is to the regression line), the darker the colour of the region. But this can be counterproductive, see attached image.
regline
The colour is so light that at first sight one does not even see the confidence region. Granted its a pathetic dataset I use here. If the colour of the confidence region expresses the narrowness of the region, I suggest using one standard colour for all regions. The width of the region is immediately evident when looking at the graph and there is no need to need to have a second indicator of the with of the confidence region.

atdotde:
If you have a normal distribution, the 2 sigma region is (up to the factor 1.96 vs 2 which I guess is less than a pixel for our plot)  is the 95% confidence level. In fact, this is why most of the time you plot 68% or 95% confidence levels (and not 90% or 98%): Those are close to one and two sigma. If that makes you sleep better, we can replace the 2 by 1.96. In fact, for a finite number of data points, one should use Student's t-distribution where the t-values are tabulated https://en.wikipedia.org/wiki/Student%27s_t-distribution#Table_of_selected_values but I am sure this would be overkill given that we are discussing precision of error bars of typically not very well measured numbers.

As I said in another PR where you asked this before: The color intensity (the alpha value to be precise) denotes the r^2 value. This is the fraction of the variance that is explained by the regression, how much of the scatter goes away, when you don't compare the y values to the mean of y but rather to the regression line. In a formula

r^2 = 1 - <(y-(ax+b))^2> /  <(y-<  y > )^2)>

willemferguson:
I understand that 1.96 and 2.0 will not make a noticeable difference to the graph. But, within a statistical context, the 1.96 value is the proper one to use. As far a small-sample graphs, I would be happy to qualify the value of the confidence region for small-sample data (i.e. n < 30) IN THE USER MANUAL. Alternatively I have a function that calculates the 95% two-tailed t-values. I would be happy with whatever your contention is.

bstoeger:
Since I didn't get the text rendering under control, this now renders onto a rectangle with the color of the background pie slice / bar fragment. I tried to look into how QGraphicsScene does this and it appears to use a private `QWidgetTextControl` class..? No idea.

From my side, this PR is "feature complete", though it is only slightly tested.

There are a number of dissatisfying points. The history is ugly, the memory management is a disaster and the drawing of the text seems like a bad idea: what will happen in a "dark" mode? We cant use the same technique, because the text rectangle will be visible on top of the chart grid.

However, I lack motivation and energy to polish this PR much more.

dirkhh:
Wait, you are saying that banging your head against a suboptimally designed API and fighting odd implementation details isn't fulfilling and rewarding in itself? 😳🤣😇
I will do a review pass in the next couple of hours and do some more testing on desktop and mobile and let you know if there are any changes I think we need before merging...
Thanks for all of your effort on this. The results are amazing.

Quick question on outcome vs. goals going in. I remember that you were able to create 100% CPU load with the initial implementation. How has this changed?

stats/statsview.cpp
dirkhh: Very odd effect on iOS (this seemed best to test on a tablet, and that's the only working tablet I have).
Dragging works reasonably well in the `y` direction, but in `x` direction it stops after a few dozen pixels. You can lift the finger and start again, but something is screwy with this :-)

stats/zvalues.h
```diff
@@ -2,19 +2,9 @@
 // Defines the z-values of features in the chart view.
 // Objects with higher z-values are painted on top of objects
 // with smaller z-values. For the same z-value objects are
-// drawn in order of addition to the graphics scene.
+// drawn in order of addition to the scene.
 #ifndef ZVALUES_H
 
-struct ZValues {
-	static constexpr double grid = -1.0;
-	static constexpr double series = 0.0;
```
dirkhh, line 10: comment on the commit message... are those 'flot' values running fast? (sorry, bad German joke)

bstoeger:
Yes, CPU usage is now much better. Still rather high when moving the legend, but that could be due to software rendering(?).

dirkhh (approved):
- I find the history not terrible and likely not worth cleaning up; the couple of detours help understand the code, IMHO
- I find the memory management and the template code used there... eye watering (that is NOT criticism of you or your code... an editorial comment decrying that this is what we ended up with)
- there are a couple of commit messages that deserve edits - I pointed out one, but I noticed a couple more that deserve some TLC
- I tried this on desktop and mobile. With the exception of moving the legend on mobile (where this is likely a highly desirable feature) everything appears to work fine
- as discussed on the ML, it would be great if the labels inside bar charts were a little smarter and would try to shrink themselves so that double digit labels will fit better - the idea would be to make it less likely that you have only single digit percentages shown
- the regression confidence interval color choice is a bit too aggressive when the confidence is high. See picture :-)
Screen Shot 2021-01-19 at 9 43 56 AM

dirkhh:
> Yes, CPU usage is now much better. Still rather high when moving the legend, but that could be due to software rendering(?).

I'm super impressed with all of this. The one thing that I would love to see fixed (after merge is fine) is the issue moving the legend on mobile. I'm hopeful this is easy to fix - it's one of those **so close** things :-)
Not sure if you want to address the commit messages or if you want me to do that as I merge... I vaguely remember that there was one (and I didn't flag it, that's my fault) where I actually wasn't 100% sure what the correct phrasing would be

oh, the color -- also can be fixed post merge

PR #3166, opened 2021-01-13
mobile/GPS: fix two errors in the GPS handling

dirkhh:
First, the time zone adjustment was wrong - this as written could only
ever have worked in UTC or by pure chance.

Second, the order of alerting the UI of the availability of a GPS fix
was also incorrect creating a race between the UI and our data
structures.

- [x] Bug fix

I'm doing a much larger rewrite of the mobile dive edit, and while playing with that I noticed that the 'use current location' thing was completely broken and cannot possibly have worked.

### Changes made:

- fix timezone calculation
- don't notify of new GPS fix until it has been added to the data structure where the UI code will look for it

### Release note:

included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - undo: reset dive-mode on undo of set-point addition
 - desktop: complete rewrite of the statistics code, significantly expanding capabilities
 - desktop: add preferences option to disable default cylinder types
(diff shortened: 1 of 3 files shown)


dirkhh:
tested on device and actually works (which is very different from what we had before this PR), so I'll merge

PR #3171, opened 2021-01-17
Selectfromstats

atdotde:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I have no idea how to do this for the other types of plots.

### DIFF
--- a/stats/scatterseries.cpp
+++ b/stats/scatterseries.cpp
@@ -10,6 +10,7 @@
 #include "core/dive.h"
 #include "core/divelist.h"
 #include "core/qthelper.h"
+#include "core/selection.h"
 
 #include <QGraphicsPixmapItem>
 #include <QPainter>
@@ -124,6 +125,17 @@ std::vector<int> ScatterSeries::getItemsUnderMouse(const QPointF &point) const
 			res.push_back(it - items.begin());
 	}
 	return res;
+}
+
+void ScatterSeries::selectItemsUnderMouse(const QPointF &point)
+{
+	std::vector<struct dive *> selected;
+	struct dive *new_current;
+
+	for(int idx: getItemsUnderMouse(point)) {
+		selected.emplace_back(new_current = items[idx].d);
+	}
+	setSelection(selected, new_current);
 }
 
 static QString dataInfo(const StatsVariable &var, const dive *d)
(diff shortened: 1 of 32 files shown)


atdotde:
oops. This is mainly for discussion, so ignore the fact that it contains contains the commit by @bstoeger, this is about the last commit.

bstoeger:
I wouldn't do this on double click, but single click. Ultimately probably also with "lasso" or "rectangle" selection.

But attention: On mobile we do not currently support selection of multiple dives, so this might lead to inconsistent state and probably should be turned off anyway.

I think in the other plots the dives are not saved per bin, because so far there was no reason to do so.

I will have a look later how to do this in a more general way. Please give me some time as I'm currently busy with other things.

bstoeger:
Also note that currently the statistics are not recalculated on undo/redo, so the `dive *` pointers need to be validated!

stats/scatterseries.cpp
```diff
@@ -126,6 +127,17 @@ std::vector<int> ScatterSeries::getItemsUnderMouse(const QPointF &point) const
 	return res;
 }
 
+void ScatterSeries::selectItemsUnderMouse(const QPointF &point)
+{
+	std::vector<struct dive *> selected;
+	struct dive *new_current;
+
+	for(int idx: getItemsUnderMouse(point)) {
+		selected.emplace_back(new_current = items[idx].d);
```
bstoeger, line 136: Two small comments on this function:

1) This should read `push_back(...)`, not `emplace_back(...)`. the former adds a value to the vector, the later constructs an object in place. `emplace_back()` makes space for the object and then calls the constructor with the given parameters on the unitialized space. A pointer clearly has value semantics, so `push_back` is preferred, even though the latter works.

2) It appears to me that if getItemsUnderMouse() returns no element, then `new_current` is uninitialized - I wonder why the compiler doesn't complain about that...?

bstoeger:
@atdotde: I reckon that you have seen #3175? Do you want me to port this PR?

atdotde:
I am somewhat busy at the moment. I can will try to work on this but it will take a while. If you are in a rush, feel free...

bstoeger:
The code of this PR is included in #3175. I reckon we should close this?

PR #3172, opened 2021-01-19
mobile: add GF fields for ceiling calculation

djunkins:
Signed-off-by: Doug Junkins <douglas.junkins@gmail.com>

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Adds fields to the advanced preferences page to modify GFLow and GFHigh for
the Buhlmann decompression model for calculating ceilings. Also updates the 
set methods for the GF preferences to update the parameters for the Buhlmann
algorithm in core/deco.c.

### Changes made:

1) Added text fields for GFLow and GFHigh to settings.qml
2) Added connections to profilewidget2.cpp to redraw profile when the GF values are changed
3) Added set_gf() calls to the set_gflow() and set_gfhigh() methods in qPrefTechnicalDetails.cpp
4) Removed set_gf() call from the desktop preference widget code in preferences_graph.cpp

### Release note:

mobile: add GF fields to adjust Buhlmann algorithm parameters for calculated ceiling

### DIFF
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -622,7 +622,7 @@ TemplatePage {
 					checked: PrefTechnicalDetails.dcceiling
 					onClicked: {
 						PrefTechnicalDetails.dcceiling = checked
-						rootItem.settingChanged()
+						rootItem.settingsChanged()
 					}
 				}
 				TemplateLabel {
@@ -632,6 +632,30 @@ TemplatePage {
 					checked: PrefTechnicalDetails.calcceiling
 					onClicked: {
 						PrefTechnicalDetails.calcceiling = checked
+						rootItem.settingsChanged()
+					}
+				}
+				TemplateLabel {
+					text: qsTr("GFLow")
+				}
+				TemplateTextField {
+					id: gfLow
+					text: PrefTechnicalDetails.gflow
+					inputMask: "99"
+					onEditingFinished: {
+						PrefTechnicalDetails.gflow = gfLow.text
+						rootItem.settingsChanged()
+					}
+				}
+				TemplateLabel {
+					text: qsTr("GFHigh")
+				}
+				TemplateTextField {
+					id: gfHigh
+					text: PrefTechnicalDetails.gfhigh
+					inputMask: "99"
+					onEditingFinished: {
+						PrefTechnicalDetails.gfhigh = gfHigh.text
 						rootItem.settingsChanged()
 					}
 				}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- mobile: add GF fields to adjust Buhlmann algorithm parameters for calculated ceiling
 - undo: save to git after editing weights [#3159]
 - undo: reset dive-mode on undo of set-point addition
 - desktop: complete rewrite of the statistics code, significantly expanding capabilities
--- a/core/settings/qPrefTechnicalDetails.cpp
+++ b/core/settings/qPrefTechnicalDetails.cpp
@@ -62,6 +62,7 @@ void qPrefTechnicalDetails::set_gfhigh(int value)
 	if (value != prefs.gfhigh) {
 		prefs.gfhigh = value;
 		disk_gfhigh(true);
+		set_gf(-1, prefs.gfhigh);
 		emit instance()->gfhighChanged(value);
 	}
 }
@@ -82,6 +83,7 @@ void qPrefTechnicalDetails::set_gflow(int value)
 	if (value != prefs.gflow) {
 		prefs.gflow = value;
 		disk_gflow(true);
+		set_gf(prefs.gflow, -1);
 		emit instance()->gflowChanged(value);
 	}
 }
--- a/desktop-widgets/preferences/preferences_graph.cpp
+++ b/desktop-widgets/preferences/preferences_graph.cpp
@@ -68,7 +68,6 @@ void PreferencesGraph::syncSettings()
 	prefs.planner_deco_mode = ui->buehlmann->isChecked() ? BUEHLMANN : VPMB;
 	qPrefTechnicalDetails::set_gflow(ui->gflow->value());
 	qPrefTechnicalDetails::set_gfhigh(ui->gfhigh->value());
-	set_gf(ui->gflow->value(), ui->gfhigh->value());
 	qPrefTechnicalDetails::set_vpmb_conservatism(ui->vpmb_conservatism->value());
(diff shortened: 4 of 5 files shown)


mobile-widgets/qml/Settings.qml
```diff
@@ -622,7 +622,7 @@ TemplatePage {
 					checked: PrefTechnicalDetails.dcceiling
 					onClicked: {
 						PrefTechnicalDetails.dcceiling = checked
-						rootItem.settingChanged()
+						rootItem.settingsChanged()
```
dirkhh, line 625: UGH. I hate that there isn't better syntax chcking... I have had so many errors like this over the years in the QML code :-(

mobile-widgets/qml/Settings.qml
```diff
@@ -635,6 +635,30 @@ TemplatePage {
...
+				TemplateLabel {
+					text: qsTr("GFLow")
+				}
+				TemplateTextField {
+					id: gfLow
+					text: PrefTechnicalDetails.gflow
+					inputMask: "99"
+					onEditingFinished: {
+						PrefTechnicalDetails.gflow = gfLow.text
+						rootItem.settingsChanged()
```
dirkhh, line 647: OK, that looks right. I like the use of the `inputMask` -- there's so much more we could do to make the QML UI better...

mobile-widgets/qml/Settings.qml
```diff
@@ -635,6 +635,30 @@ TemplatePage {
...
+					}
+				}
+				TemplateLabel {
+					text: qsTr("GFHigh")
+				}
+				TemplateTextField {
+					id: gfHigh
+					text: PrefTechnicalDetails.gfhigh
+					inputMask: "99"
+					onEditingFinished: {
```
dirkhh, line 657: One thing I remember from other encounters with TextFields... make sure that this signal always fires when you stop editing. I.e., if you just tap on a different page in the menu bar, does this signal fire? What if you use the main menu to switch to the dive list. Etc. I'm not saying that this is the wrong signal - I'm just remembering that this has caused us pain in the past.

mobile-widgets/qmlmanager.cpp
```diff
@@ -323,6 +324,10 @@ QMLManager::QMLManager() : m_locationServiceEnabled(false),
 	connect(Command::getUndoStack(), &QUndoStack::undoTextChanged, this, &QMLManager::undoTextChanged);
 	connect(Command::getUndoStack(), &QUndoStack::redoTextChanged, this, &QMLManager::redoTextChanged);
 
+  // Add connections to monitor for changes to the Deco settings
+	connect(qPrefTechnicalDetails::instance(), &qPrefTechnicalDetails::gflowChanged, this, &QMLManager::gflowChanged);
+	connect(qPrefTechnicalDetails::instance(), &qPrefTechnicalDetails::gfhighChanged, this, &QMLManager::gfhighChanged);
```
dirkhh: I'm not convinced that this is the best way to do this - but I'd need to do more digging in the code.
Look at this... https://github.com/subsurface/subsurface/blob/25cfd852b910c7bbd53f0a1a5244c659761c2cc5/core/settings/qPrefTechnicalDetails.cpp#L60-L78
doesn't that seem to imply that `set_gf()` already gets called?

dirkhh (commented):
This is great. Thank you for working on that.
I cannot tell you how excited I am to try to get someone else to help with the QML UI.
I added a few comments and questions.
On almost everything here I'm in the "I think I remember this works kinda like this" stage of "knowledge" - so please consider all of this as me asking questions and it's entirely possible that the answer to these questions is "No, Dirk, you are wrong, because..."
But I'd love to have that conversation so we both learn more about how this really SHOULD work...

mobile-widgets/qmlmanager.cpp
djunkins: I can only find disk_gfhigh() called in two places -- once in qPrefTechnicalDetails::loadSync and once in qPrefTechnicalDetails::set_gfhigh which you included above. 

set_gfhigh() is the method called when the preference is updated either on the mobile app or the desktop app. In that case as you can see, disk_gfhigh is called with doSync=true which means from the if(doSync) statement that set_gf does not get called, since it is only in the "else" code block. I was not sure of the implications of moving the set_gf outside the if/then/else block since it would impact both the desktop and mobile versions, so I looked for another solution that I could limit to only the mobile app.

What I did was to catch the gfhighChanged signal that is emitted on the next line after the call to disk_gf(true) in the qmlmanager.cpp code since that is limited to the mobile app.

djunkins: Also, as I mentioned in our other email thread, on the desktop app set_gf() is called in the desktop specific code in desktop-widgets/preferences/preferences-graph.cpp line 71 after the preferences are updated.

mobile-widgets/qml/Settings.qml
djunkins: There are two signal handlers  available for the TextFields, "onAccepted" and "onEditingFinished", with the difference being that the Accepted signal is only emitted when enter or return is hit, while EditingFinished is also emitted when the text field loses focus. In both cases the signal is only supposed to be emitted if the text in the field meets the inputMask and any validators that have been included return true.

mobile-widgets/qmlmanager.cpp
djunkins: I've now looked at core/settings/README.md and it looks like the doSync variable is meant to distinguish between when the preference are loaded at startup (doSync==false) or sync'd to save all settings to disk (doSync=true). It appears that currently set_gf() is only called when the preferences are loaded, but it should be possible to also call it when syncing changed values for GFLow and GFHigh to disk and eliminate the call from preferences_graph.cpp and from the addition I made to qmlmanager.cpp.

Can you think of any downside to setting the GFs in the Buhlmann algorithm each time they are sync'd to the disk?

djunkins: I moved the set_gf call to the set_gfhigh and set_gflow methods to cover both desktop and mobile apps and force pushed the commit. It appears everything still works as expected and it did not impact the planner in any way on the desktop app as far as I can tell. It might be good to have another set of eyes look at it though.

dirkhh: I think this is the right direction. I was reading through the code and got distracted a few times (day job getting in the way) - but yes, I don't know why the `set_gf()` call would only be in the `doSync == false` case. That doesn't seem to make any sense. Why would you want to set the GF and then not want it applied?
@atdotde -- any comment on this?

dirkhh (approved):
OK, I think this should be merged.

PR #3176, opened 2021-01-20
fix delete crash

dirkhh:
Apparently in some cases there is a race condition where the action to
delete a dive is still running while a corresponding QML object (but not
that action itself) is being deleted, likely because the action is
becoming invisible once the dive is actually deleted. I'm speculating
that this may be somewhere inside the delegates for the repeater showing
the actions in the context drawer.

This commit is a bit of a convoluted hack. We create a property to hold
the dive id to be deleted and then call a function to delete that dive
via a timer. This of course creates a blatant race condition in itself -
but the user is extremely unlikely to be able to use the context menu to
delete two different dives within 100ms. The code tries to keep that
race window as small as possible.

There has got to be a better way to do this.

- [x] Bug fix

### Related issues:

The crash that so far only Theodor has been able to reproduce

### Release note:

not needed, I think - this has never been reported in a release version (even though it might have happened there as well? maybe?)

### DIFF
--- a/mobile-widgets/qml/DiveList.qml
+++ b/mobile-widgets/qml/DiveList.qml
@@ -278,11 +278,40 @@ Kirigami.ScrollablePage {
 		visible: currentItem && currentItem.myData && !currentItem.myData.isTrip
 		onTriggered: manager.toggleDiveInvalid(currentItem.myData.id)
 	}
+	Timer {
+		// this waits a tenth of a second and then deletes the dive -
+		// this way the Action in the contextDrawer can complete its signal handler before
+		// having the visual QML element going away and being deleted (which could cause a crash)
+		id: delayedDelete
+		interval: 100
+		onTriggered: {
+			var myDiveToDelete = diveToDelete // try to minimize the window for a race
+			diveToDelete = -1
+			if (myDiveToDelete !== -1) {
+				manager.appendTextToLog("now deleting dive id " + myDiveToDelete)
+				manager.deleteDive(myDiveToDelete)
+			} else {
+				manager.appendTextToLog("delayedDelete triggered with dive id -1");
+			}
+		}
+	}
+	property int diveToDelete: -1
 	property QtObject deleteAction: Kirigami.Action {
 		text: qsTr("Delete dive")
 		icon { name: ":/icons/trash-empty.svg" }
 		visible: currentItem && currentItem.myData && !currentItem.myData.isTrip
-		onTriggered: manager.deleteDive(currentItem.myData.id)
+		onTriggered: {
+			if (diveToDelete === -1) {
+				diveToDelete = currentItem.myData.id
+				manager.appendTextToLog("setting timer to delete dive " + diveToDelete)
+				delayedDelete.start()
+				contextDrawer.close()
+				manager.appendTextToLog("closed drawer, done with the action")
+			} else {
+				manager.appendTextToLog("deleteAction triggered with diveToDelete already set to " +
+							diveToDelete + " -- ignoring just to be safe");
+			}
+		}
 	}
 	property QtObject mapAction: Kirigami.Action {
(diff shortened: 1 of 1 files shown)


bstoeger:
> Apparently in some cases there is a race condition where the action to
delete a dive is still running while a corresponding QML object (but not
that action itself) is being deleted, likely because the action is
becoming invisible once the dive is actually deleted.

Huh? How is that possible? I thought all these things run in the UI thread. If they don't - there is a bigger problem.

This needs further investigation.

dirkhh:
I'd love to investigate further. If only I was able to reproduce it.
Tried at least 20 different ways. On five different devices, plus of course mobile-on-desktop
CANNOT reproduce it. Yet Theodor can reproduce the crash every single time. His log files (extracted from the Android device after the crash) clearly show that we are deleting a QML object before its signal handler has run. It's all in the 'Accidental file delete on mobile.' thread...

bstoeger:
I'll investigate tomorrow. Please don't commit yet.

bstoeger:
> Apparently in some cases there is a race condition where the action to
delete a dive is still running while a corresponding QML object (but not
that action itself) is being deleted, likely because the action is
becoming invisible once the dive is actually deleted. I'm speculating
that this may be somewhere inside the delegates for the repeater showing
the actions in the context drawer.

To test this hypothesis, I would suggest setting `visible` unconditionally to true.

The reason this all makes no sense to me is that the situation you describe is exactly what happens in my case - yet no crash.

Could this be a Qt-version problem?

What I do get after deletion and before the delete action completes is this:
```
INFO: Qt Quick Layouts: Detected recursive rearrange. Aborting after two iterations.
```

Does that have any meaning? In the logs posted to the mailing list, I see this instead:

```
qrc:/qml/DiveDetails.qml:406:5: QML DiveDetailsView: Binding loop detected for property "height"
```

Which indicates to me that there are different QML versions at play?

You seem to have good connections to the QML makers, perhaps it is a good idea to ask them?

dirkhh:
I assume that you aren't testing on Android but you are testing mobile-on-desktop?
That results in slightly different settings being used in Kirigami. And likely even in QML itself.
Because one is a platform with keyboard, mouse, and a window manager, where the app runs in a window, the other is a platform with a virtual keyboard, a touch screen, no window management, where the app runs full screen.
I have learned the hard way that there are quite a few subtle differences in how this affects our QML UI - most things you can test on desktop. Some things just behave differently on device.
Of course, getting a cheap Android device to test this on is always an option (there are 50€ tablets out there).
The 'binding loop for height' thing is a non-issue. It's just a sign that the way Kirigami and our code interoperate occasionally leaves QML guessing about the intended geometry of an element. In earlier versions that could send QML in a tail spin, but since Qt 5.12 it has become much smarter in detecting (and reporting) those issues and just making reasonable guesses (IIRC it picks the implicit size of the children now).

Setting this unconditionally to visible creates an odd user experience but I'm sure I can figure out a way to make this work... let me look how Kirigami implements disabled actions in the drawer - maybe that's enough...

bstoeger:
> Setting this unconditionally to visible creates an odd user experience but I'm sure I can figure out a way to make this work... let me look how Kirigami implements disabled actions in the drawer - maybe that's enough...

I do not suggest this to be a fix, but to test your theory. Once there is a good understanding of the situation, I recommend asking the QML developers how they suggest doing things. Or if we should just declare it a lost cause. After all, this can happen in other places. Think for example undo/redo. Adding random delays as means of thread synchronization is programming by fate.

bstoeger:
> So we agree that this is the wrong fix, but continue not to have any idea what the right fix would be.

To isolate the problem I still suggest to make the visibility unconditional and check with the user whether that still crashes.

Making the object invisible should not destroy it, so the whole issue is ominous.

PR #3177, opened 2021-01-20
statistics: in render thread resize first, delete items later

bstoeger:
In updatePaintNode() when a change of size is detected, the chart
items are rebuilt. For examples, the old grid and axis ticks are
deleted. Deleting works be adding chart items to a delete list.

However, the actual deleting of the items was done before, and
therefore the old items were kept for one cycle leading to weird
rendering artifacts.

These artifacts can be removed by exchanging the order of
operations: first recalculate the chart items, then delete
items marked for deletion.

However, this raises an important point: The function to resize
the chart is called in two contexts: the ui thread *and* the
rendering thread. This appears like a very bad idea and ultimately
the resizing should probably be done in the UI thread, not
the rendering thread.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a rendering bug, but raises the question whether the way resizing is handled is valid.

The resize function is called in two different thread contexts, which seems like a very bad idea. I will have to investigate. Sigh - this API is really nasty and badly documented.
 
### Changes made:

1) Change order of execution: on resize, recalculate chart first then delete objects marked for deletion.

### DIFF
--- a/stats/statsview.cpp
+++ b/stats/statsview.cpp
@@ -117,6 +117,20 @@ QSGNode *StatsView::updatePaintNode(QSGNode *oldNode, QQuickItem::UpdatePaintNod
 	if (!n)
 		n = rootNode = new RootNode(window());
 
+	// If the plot area changed, we might need to redraw the grid, axes, etc.
+	// This may do quite some extensive allocations and rendering, so it is
+	// questionable whether this should (or can) be done in the render thread.
+	// After all, each QObject has an owning thread and we might end up
+	// with a collection of objects belonging to different threads. Scary.
+	// Probably this should be treated differently and just signal the UI thread
+	// that the size changed?
+	QRectF rect = boundingRect();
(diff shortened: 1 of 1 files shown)


dirkhh (commented):
Oh jolly. Yeah. I think all of these changes should be in the UI thread and the render thread should just, err, render?

bstoeger:
The problem is that this is all very ill defined. Note for example that there is no locking. The way I read the documentation is that updateRenderNode() is only called when the UI thread is locked, but I am not 100% sure. It says QSG nodes should only be accessed on the render thread, but not what the render thread may or may not do.

Alternative PR that intercepts geometry change in the UI thread incoming.

dirkhh:
I believe this can be closed with #3178 merged

PR #3178, opened 2021-01-20
statistics: do resizing in UI thread, not render thread [Alternative to #3177]

bstoeger:
The updatePaintNode() function, which is run on the render
thread detected a geometry change and initiated recalculation
of the chart layout.

This means that plotAreaChanged() was called in two different
thread contexts, which is questionable. Instead, hook into
the geometryChanged() function and recalculate the chart items
there.

This fixes a rendering bug, because the old code would first
delete unneeded items and then rerender the chart. Thus, old
grid and tick items were still visible.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Alternative version of #3177. Just as contentious, because I don't know what intercepting the geometry change means and if it is guaranteed to run before updatePaintNode()?

### Changes made:

1) Intercept geometry changes and update the chart accordingly.

### Related issues:

Alternative to #3177.

### DIFF
--- a/stats/statsview.cpp
+++ b/stats/statsview.cpp
@@ -31,6 +31,7 @@ static const double sceneBorder = 5.0;			// Border between scene edges and stati
 static const double titleBorder = 2.0;			// Border between title and chart
 
 StatsView::StatsView(QQuickItem *parent) : QQuickItem(parent),
+	backgroundDirty(true),
 	highlightedSeries(nullptr),
 	xAxis(nullptr),
 	yAxis(nullptr),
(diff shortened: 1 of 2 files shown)


bstoeger:
This needs some testing: which (both?) version does crash on resize? Which still has artifacts?
Artifacts in current master can be provoked by switching to/from the statistics and windows resize.

dirkhh (commented):
So to me, with my limited understanding of how this should work, this seems like the more logical approach.
I will test this one to see if I can make it crash :-)

dirkhh:
After some "vigorous resizing" in dark mode (to more clearly see what's the statistics and what isn't):
Screen Shot 2021-01-20 at 3 12 30 PM

dirkhh:
take it back, doesn't require resizing to go sideways, this is right after the start:
Screen Shot 2021-01-20 at 3 14 43 PM

bstoeger:
Should be fixed [explanation: geometryChanged() on mobile gives the absolute rectangle with respect to the screen, to paint the background we need the rectangle with (0, 0) at the top left corner of the chart].

dirkhh:
After quite a bit of testing mobile-on-desktop, desktop, and mobile-on-device I'm happy to merge

PR #3179, opened 2021-01-21
Start rework of mainwindow quadrant code

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This completely reworks the quadrant code: instead of two orthogonal states (application state and view state), merge everything into one state.

Moreover, remove widgets from the splitters, because this is the only way to have a widget appear at different quadrants owing to Qt's tree-based ownership model.

This is unoptimized and I've observed flickering when changing state, because all widgets are removed from the splitters and then readded. The plan is to fix these issues in future commits, but for now the goal is to make it stable.

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -338,10 +338,21 @@ MainWindow::MainWindow() : QMainWindow(),
 #endif
 }
 
+static void clearSplitter(QSplitter &splitter)
+{
+	// Qt's ownership model is absolutely hare-brained.
+	// To remove a widget from a splitter, you reparent it, which
+	// informs the splitter via a signal. Wow.
+	while (splitter.count() > 0)
+		splitter.widget(0)->setParent(nullptr);
+}
+
 MainWindow::~MainWindow()
 {
 	// Remove widgets from the splitters so that they don't delete singletons.
-	clearSplitters();
+	clearSplitter(*topSplitter);
+	clearSplitter(*bottomSplitter);
+	clearSplitter(*ui.mainSplitter);
 	write_hashes();
 	m_Instance = nullptr;
 }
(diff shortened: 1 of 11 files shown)


dirkhh:
I thought there were ways to reparent widgets in Qt at runtime... but I'm not an expert (obviously)
I'll review this later this morning

dirkhh (commented):
I like the direction.
I think you have a couple commits from a different PR in here at the start.
I need to more carefully think through the code that removes and adds the widgets, but at first glance it seems to make sense.
Maybe @tcanabrava wants to comment? After all, he wrote most of the old code :-)

desktop-widgets/mainwindow.cpp
```diff
@@ -181,12 +183,21 @@ MainWindow::MainWindow() : QMainWindow(),
...
 	registerApplicationState(ApplicationState::Statistics, { { statistics, FLAG_NONE }, { profileContainer, FLAG_NONE },
 								 { diveList, FLAG_DISABLED },   { &filterWidget, FLAG_NONE } });
+	registerApplicationState(ApplicationState::MapMaximized, { { nullptr, FLAG_NONE }, { nullptr, FLAG_NONE },
+								   { nullptr, FLAG_NONE }, { mapWidget, FLAG_NONE } });
+	registerApplicationState(ApplicationState::ProfileMaximized, { { nullptr, FLAG_NONE }, { profileContainer, FLAG_NONE },
+								       { nullptr, FLAG_NONE }, { nullptr, FLAG_NONE } });
+	registerApplicationState(ApplicationState::ListMaximized, { { nullptr, FLAG_NONE },  { nullptr, FLAG_NONE },
+								    { diveList, FLAG_NONE }, { nullptr, FLAG_NONE } });
+	registerApplicationState(ApplicationState::InfoMaximized, { { mainTab.get(), FLAG_NONE }, { nullptr, FLAG_NONE },
+								    { nullptr, FLAG_NONE },       { nullptr, FLAG_NONE } });
```
dirkhh: since we are rethinking all of this - why does it make sense to implement the four full screen modes as states of the splitters. Maybe I'm missing the obvious problem, but couldn't we just simply hide the splitters and show those widgets as the only window content? Wouldn't that simplify things?

desktop-widgets/mainwindow.cpp
```diff
@@ -753,7 +753,7 @@ void MainWindow::planCreated()
 
 void MainWindow::on_actionReplanDive_triggered()
 {
-	if (!plannerStateClean() || !current_dive)
+	if (!plannerStateClean() || !current_dive || !userMayChangeAppState())
```
dirkhh, line 765: I like this. This is clever.

desktop-widgets/mainwindow.cpp
```diff
@@ -1571,29 +1573,29 @@ void MainWindow::registerApplicationState(ApplicationState state, Quadrants q)
 	applicationState[(int)state] = q;
 }
 
-void MainWindow::setQuadrantWidget(const Quadrant &q, QSplitter *splitter)
+void MainWindow::setQuadrantWidget(const Quadrant &q, QSplitter &splitter)
```
dirkhh: because this always confuses me... why is this better here...?

desktop-widgets/mainwindow.cpp
bstoeger: > Maybe I'm missing the obvious problem, but couldn't we just simply hide the splitters and show those widgets as the only window content? Wouldn't that simplify things?

That's essentially what the code does. Only the splitters with content are shown. The advantage is that any combination is possible - from one to four widgets. For example in the statistics mode three widgets are shown: one on top, two at the bottom. You could also have one right, one left or one top, one bottom. I don't think adding an extra mode for a single widget would make things more simple..?

desktop-widgets/mainwindow.cpp
bstoeger: A reference cannot be null and can't be reseated. The reader and the compiler know that there is no point in a null-check.

See: https://godbolt.org/z/1vd55o

bstoeger:
Clearly, nobody tested this, as there have been no complaints about flickering. I added a commit that tries to be smarter and solves the problem for me. YMMV.

dirkhh:
> Clearly, nobody tested this, as there have been no complaints about flickering. I added a commit that tries to be smarter and solves the problem for me. YMMV.

Not to be argumentative. But I did test this. And when run under KDE in a VM (that's my default test bed these days) I didn't seem to recall significant flickering.

bstoeger:
Interesting. Could it be that the VM adds additional buffering? For me (KDE, native) switching to the statistics screen and back flickered noticeably. Or perhaps I'm simply oversensitive concerning such visual artifacts.

dirkhh:
Still testing in a VM under Plasma (openSUSE Tumbleweed) and see no artifacts and wasn't able to confuse the code.
On first startup it gives a fairly odd window size with odd sizes for the four quadrants - I wonder if that default could be a little more balanced...
Screen Shot 2021-02-03 at 8 18 57 AM
Completely making this up... 60/40 top to bottom, top: 50/50 tab vs profile, bottom 60/40 dive list to map?
But other than that, this seems ready to merge (and even that change could very clearly be done post merge)

dirkhh:
oh, forgot: https://github.com/subsurface/subsurface/pull/3179/commits/3fbcb409f33b92c03d34ac853be556921b351ed9
`be more smarter` -- drop the `more`

bstoeger:
> Completely making this up... 60/40 top to bottom, top: 50/50 tab vs profile, bottom 60/40 dive list to map?
But other than that, this seems ready to merge (and even that change could very clearly be done post merge)

Added a commit with these values. Please test.

djunkins:
I had a chance to compile and test this branch. I haven’t seen any issues at this point and the annoying behavior of the quadrants reverting to the default sizes after toggling the filter selection is no longer there.

-Doug

dirkhh (approved):
Did more testing, like it all.

PR #3181, opened 2021-01-23
Small mobile updates

dirkhh:

- [ ] New feature

two small changes.
one allows users who experience crashes when applying GPS fixes to share their GPS fixes with the developers
the other is a small improvement to the user experience that tells the user the result of a manual sync with the cloud


dirkhh:
Neither of these are great, but they are moving us in the right direction. So I'll merge

PR #3183, opened 2021-01-23
Redo BT detection

dirkhh:

- [x] Code cleanup

I think this is more logical and more structured - but it's not much of an improvement
It does fix a few bugs (missing Pelagic patterns that we didn't have before - see if you can spot them in this mess)

### Mentions:

@torvalds

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -17,6 +17,26 @@ namespace {
 }
 BTDiscovery *BTDiscovery::m_instance = NULL;
 
+struct modelPattern {
+	uint16_t    model;
+	const char *vendor;
+	const char *product;
+};
+static struct modelPattern model[] = {
+	{ 0x4552, "Oceanic", "Pro Plus X" },
+	{ 0x455A, "Aqualung", "i750TC" },
+	{ 0x4647, "Sherwood", "Sage" },
+	{ 0x4648, "Aqualung", "i300C" },
+	{ 0x4649, "Aqualung", "i200C" },
+	{ 0x4651, "Aqualung", "i770R" },
+	{ 0x4652, "Aqualung", "i550C" },
+	{ 0x4653, "Oceanic", "Geo 4.0" },
+	{ 0x4654, "Oceanic", "Veo 4.0" },
+	{ 0x4655, "Sherwood", "Wisdom 4" },
+	{ 0x4656, "Oceanic", "Pro Plus 4" },
+	{ 0x4743, "Aqualung", "i470TC" }
+};
+
 static dc_descriptor_t *getDeviceType(QString btName)
 // central function to convert a BT name to a Subsurface known vendor/model pair
 {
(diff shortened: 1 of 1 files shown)


core/btdiscovery.cpp
```diff
@@ -133,6 +122,22 @@ static dc_descriptor_t *getDeviceType(QString btName)
...
+	} else { // finally try all the Pelagic/Aqualung names
+		// the source of truth for this data is in libdivecomputer/src/descriptor.c
+		// we'd prefer to use the filter functions there but current design makes that really challenging
+		// The Pelagic dive computers (generally branded as Oceanic, Aqualung, or Sherwood)
+		// show up with a two-byte model code followed by six bytes of serial
+		// number. The model code matches the hex model (so "FQ" is 0x4651,
+		// where 'F' is 46h and 'Q' is 51h in ASCII).
+		for (uint16_t i = 0; i < sizeof(model) / sizeof(struct modelPattern); i++) {
+			QString pattern = QString("^%1%2\\d{6}$").arg(QChar(model[i].model >> 8)).arg(QChar(model[i].model & 0xFF));
+			qDebug() << "trying pattern" << pattern;
```
dirkhh, line 122: eh, yeah, let's definitely drop this before merging 😇

dirkhh:
I started writing a PR for libdivecomputer - but the way this is structured really doesn't help us much - because we really want to set a specific type. And that's not how that code is organized. It really just wants to say "maybe this could be it" - it's not designed to identify a specific dive computer like we want.

So I think this is indeed the right way to do this.

torvalds:
Ack, this all looks much better than the old haphazard checks.

dirkhh:
I'll merge locally without the silly `qDebug()`.
Thanks

core/btdiscovery.cpp
```diff
@@ -17,6 +17,71 @@ namespace {
...
+	// so always pick the "Petrel 2" as product when seeing a Petrel
+	{ "Petrel", "Shearwater", "Petrel 2" },
+	{ "Perdix", "Shearwater", "Perdix" },
+	{ "Teric", "Shearwater", "Teric" },
+	{ "Peregrine", "Shearwater", "Peregrine" },
+	{ "NERD 2", "Shearwater", "NERD 2" },
+	{ "NERD", "Shearwater", "NERD" }, // order is important, test for the more specific one first
+	{ "Predator", "Shearwater", "Predator" },
+	{ "Predator", "Shearwater", "Predator" },
+	{ "Predator", "Shearwater", "Predator" },
```
henrik242, line 59: Just out of curiosity, why are there four identical Predators in this list?

dirkhh: That's to make extra sure that... err... you know what... yeah... cut'n'paste error on my part.
Oops

dirkhh: fixed

PR #3186, opened 2021-01-27
remove global state from DivePlannerPointsModel

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

More work in the "make profile reentrant" project. This removes `displayed_dive` from the DivePlannerPointsModel. The division of labour between model, profile and planner is incredibly complex. Therefore, currently only the global displayed_dive is supported, but this will change in the future.

To repeat the long-time goal of this work:
- Make it possible to have multiple profiles at the same time. On the one hand for printing, on the other hand for interactive profiles on mobile
- Make undo more fine grained (i.e. undo individual changes in the planner). This will makes less brittle, by removing modality.

Attention: this changes the UI behavior when moving planner points across each other. In the old code, the active planner point was switched, now it stays the same. For me this is more logical, though I'm sure that others will disagree. I can reinstate the old behavior - no problem.

This also fixes a bug when deleting multiple planner points. It's a wonder that this didn't crash.

And of course it will introduce a number of new bugs - only lightly tested.

### Release note:

Not yet.

### DIFF
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -50,12 +50,7 @@ DivePlannerWidget::DivePlannerWidget(QWidget *parent) : QWidget(parent, QFlag(0)
 	view->setColumnHidden(CylindersModel::DEPTH, false);
 	view->setItemDelegateForColumn(CylindersModel::TYPE, new TankInfoDelegate(this));
 	connect(ui.cylinderTableWidget, &TableView::addButtonClicked, plannerModel, &DivePlannerPointsModel::addCylinder_clicked);
-
-	// addStop actually accept a call with no parameters, due to default parameters, but the connect() syntax without SIGNAL/SLOT
-	// does not understand default parameters and causes errors to be thrown.
-	// Continue to use old syntax, to avoid problems.
-	connect(ui.tableWidget, SIGNAL(addButtonClicked()), plannerModel, SLOT(addStop()));
-
+	connect(ui.tableWidget, &TableView::addButtonClicked, plannerModel, &DivePlannerPointsModel::addDefaultStop);
 	connect(cylinders, &CylindersModel::dataChanged, GasSelectionModel::instance(), &GasSelectionModel::repopulate);
 	connect(cylinders, &CylindersModel::rowsInserted, GasSelectionModel::instance(), &GasSelectionModel::repopulate);
 	connect(cylinders, &CylindersModel::rowsRemoved, GasSelectionModel::instance(), &GasSelectionModel::repopulate);
@@ -63,7 +58,6 @@ DivePlannerWidget::DivePlannerWidget(QWidget *parent) : QWidget(parent, QFlag(0)
 	connect(cylinders, &CylindersModel::dataChanged, plannerModel, &DivePlannerPointsModel::cylinderModelEdited);
 	connect(cylinders, &CylindersModel::rowsInserted, plannerModel, &DivePlannerPointsModel::cylinderModelEdited);
 	connect(cylinders, &CylindersModel::rowsRemoved, plannerModel, &DivePlannerPointsModel::cylinderModelEdited);
-
 
 	ui.tableWidget->setBtnToolTip(tr("Add dive data point"));
 	connect(ui.startTime, &QDateEdit::timeChanged, plannerModel, &DivePlannerPointsModel::setStartTime);
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -540,8 +540,6 @@ PlannerWidgets::PlannerWidgets()
 void PlannerWidgets::planDive()
 {
 	DivePlannerPointsModel::instance()->setPlanMode(DivePlannerPointsModel::PLAN);
-
-	MainWindow::instance()->graphics->setPlanState(&displayed_dive, 0);
 	dc_number = 0;
 
 	// create a simple starting dive, using the first gas from the just copied cylinders
@@ -568,8 +566,6 @@ void PlannerWidgets::replanDive()
 	copy_dive(current_dive, &displayed_dive); // Planning works on a copy of the dive (for now).
 	DivePlannerPointsModel::instance()->setPlanMode(DivePlannerPointsModel::PLAN);
 	DivePlannerPointsModel::instance()->loadFromDive(&displayed_dive);
-
-	MainWindow::instance()->graphics->setPlanState(&displayed_dive, 0);
 
 	plannerWidget.setReplanButton(true);
 	plannerWidget.setupStartTime(timestampToDateTime(displayed_dive.when));
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -545,7 +545,7 @@ void PlannerWidgets::planDive()
 	dc_number = 0;
 
 	// create a simple starting dive, using the first gas from the just copied cylinders
-	DivePlannerPointsModel::instance()->createSimpleDive();
+	DivePlannerPointsModel::instance()->createSimpleDive(&displayed_dive);
 
 	// plan the dive in the same mode as the currently selected one
 	if (current_dive) {
@@ -563,17 +563,20 @@ void PlannerWidgets::planDive()
 
 void PlannerWidgets::replanDive()
 {
+	if (!current_dive)
+		return;
+	copy_dive(current_dive, &displayed_dive); // Planning works on a copy of the dive (for now).
 	DivePlannerPointsModel::instance()->setPlanMode(DivePlannerPointsModel::PLAN);
+	DivePlannerPointsModel::instance()->loadFromDive(&displayed_dive);
 
 	MainWindow::instance()->graphics->setPlanState();
 
 	plannerWidget.setReplanButton(true);
-	plannerWidget.setupStartTime(timestampToDateTime(current_dive->when));
-	if (current_dive->surface_pressure.mbar)
-		plannerWidget.setSurfacePressure(current_dive->surface_pressure.mbar);
-	if (current_dive->salinity)
-		plannerWidget.setSalinity(current_dive->salinity);
-	DivePlannerPointsModel::instance()->loadFromDive(current_dive);
+	plannerWidget.setupStartTime(timestampToDateTime(displayed_dive.when));
+	if (displayed_dive.surface_pressure.mbar)
+		plannerWidget.setSurfacePressure(displayed_dive.surface_pressure.mbar);
+	if (displayed_dive.salinity)
+		plannerWidget.setSalinity(displayed_dive.salinity);
 	reset_cylinders(&displayed_dive, true);
 	DivePlannerPointsModel::instance()->cylindersModel()->updateDive(&displayed_dive);
 }
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -563,7 +563,6 @@ void PlannerWidgets::planDive()
 
 void PlannerWidgets::replanDive()
 {
-	DivePlannerPointsModel::instance()->clear();
 	DivePlannerPointsModel::instance()->setPlanMode(DivePlannerPointsModel::PLAN);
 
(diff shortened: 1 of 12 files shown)


qt-models/diveplannermodel.h
```diff
@@ -73,7 +75,6 @@ public
 	void setVpmbConservatism(int level);
 	void setSurfacePressure(int pressure);
 	void setSalinity(int salinity);
-	int getSurfacePressure();
```
dirkhh, line 76: I'm don't think this is the case here (as git blame points to early commits from Tomaz) but one way to make things accessible from QML is to make it a slot (`Q_INVOKABLE` is the cleaner one - but at least early on I didn't realize that so I used slots for quite a few of these).

profile-widget/profilewidget2.cpp
```diff
@@ -1276,6 +1276,7 @@ void ProfileWidget2::setAddState()
 	DivePlannerPointsModel *plannerModel = DivePlannerPointsModel::instance();
 	connect(plannerModel, &DivePlannerPointsModel::dataChanged, this, &ProfileWidget2::replot);
 	connect(plannerModel, &DivePlannerPointsModel::cylinderModelEdited, this, &ProfileWidget2::replot);
+	connect(plannerModel, &DivePlannerPointsModel::modelReset, this, &ProfileWidget2::pointsReset);
```
dirkhh: comment on the commit message - there's one typo "The assumed single..." -- but overall I struggle a bit with understanding what it says. Would you mind rephrasing a bit?

qt-models/diveplannermodel.cpp
```diff
@@ -844,6 +844,17 @@ void DivePlannerPointsModel::editStop(int row, divedatapoint newData)
 	 */
 	int old_first_cylid = divepoints[0].cylinderid;
 
+	// Refuse creation of two points with the same time stamp.
+	// Note: we time in the positive direction to avoid moving below 0.
```
dirkhh: this is missing a word? like `move`?

desktop-widgets/diveplanner.cpp
```diff
@@ -563,17 +563,20 @@ void PlannerWidgets::planDive()
...
-		plannerWidget.setSurfacePressure(current_dive->surface_pressure.mbar);
-	if (current_dive->salinity)
-		plannerWidget.setSalinity(current_dive->salinity);
-	DivePlannerPointsModel::instance()->loadFromDive(current_dive);
+	plannerWidget.setupStartTime(timestampToDateTime(displayed_dive.when));
+	if (displayed_dive.surface_pressure.mbar)
+		plannerWidget.setSurfacePressure(displayed_dive.surface_pressure.mbar);
+	if (displayed_dive.salinity)
+		plannerWidget.setSalinity(displayed_dive.salinity);
+	DivePlannerPointsModel::instance()->loadFromDive(&displayed_dive);
```
dirkhh: This needs a closer review. `current_dive` is usually a dive in the dive list.
Whereas `displayed_dive` is a copy of such a dive. So without more searching (which I don't have time for) this seems like it changes behavior?

profile-widget/profilewidget2.cpp
```diff
@@ -1900,17 +1900,15 @@ void ProfileWidget2::keyDeleteAction()
 	if ((currentState != ADD && currentState != PLAN) || !plannerModel)
 		return;
 
-	int selCount = scene()->selectedItems().count();
```
dirkhh, line 1933: typo in commit message: "to noe delete"

dirkhh (commented):
As always, I am impressed with your willingness to get into gnarly code
A few comments on commit messages, and one serious question about a change

Overall looks right

desktop-widgets/diveplanner.cpp
bstoeger: Note that a few lines above `current_dive` is copied into `displayed_dive`, so there should be no behavioral change.

Once the profile can display arbitrary dives, the `displayed_dive` will be made a _local_ copy of the dive-to-be-edited. Ultimately, it should be removed completely with fine-grained undo.

bstoeger:
Fixed typos and added a commit that removes `displayed_dive` from the profile code. This is super-experimental!

dirkhh:
Did another round of code review and didn't find anything that caught my eye (and we all know that my ratio of catching errors in review approaches low double digits... urgs). I also did some light testing (I never use the planner - my interest in technical diving that caused me to start the planner quickly faded) - but I'm not quite sure WHAT I should be testing in the first place.

@bstoeger  - your comment that this is super-experimental doesn't instill a lot of confidence that I should be merging this code, on the other hand, it's been sitting here with no interaction for almost a week and that's not really helping, either.
Should we just merge it so that we get some more testing done? At least a small number of people tend to test when I push things to master...
@atdotde - you are the only other remaining _active_ developer... thoughts?

BTW: this is the last four releases:
```
$ git shortlog -s -n v4.9.7..
   342  Berthold Stoeger
   334  Dirk Hohndel
    25  Robert C. Helling
     4  Linus Torvalds
     4  Michał Sawicz
     4  Willem Ferguson
     4  Doug Junkins
     3  Jason Bramwell
     2  Lubomir I. Ivanov
     1  Miika Turkia
     1  Salvador Cuñat
     1  Damian Zaremba
     1  henrik242
```
that's... pretty bad...

bstoeger:
In my testing this broke mobile - do not merge. I got side-tracked with the statistics selection and other things, though.

The commit stats are disheartening, yes.

bstoeger:
Some more small updates. Note: this breaks the "edit profile" feature, whereas the planner still seems to work. I will investigate.

bstoeger:
> Note: this breaks the "edit profile" feature, whereas the planner still seems to work.

Fixed that particular problem. This would now need testing before I continue. Before addressing the printing madness, I would incorporate the profile-editing in the undo-system so that we can get rid of the weird modality.

bstoeger:
Gosh, the profile/planner code is killing me. See last two commits - this is all very experimental.

bstoeger:
Only one family of calls were removed and these were clearly bogus. They caused a double-replot of the profile, so this is a borderline bug fix.

bstoeger:
More small cleanups. OMG the control flow is crazy.

BTW: I think current master is broken: the deco steps are miscalculated and the profile is shown in warning mode. Clearly something broke with the profile-changes. It's somewhat discouraging that nobody even notices.

dirkhh:
I believe that very few people use dive planning.
I believe that very few people actively test latest master.
Very few times very few is really really very few, right?

atdotde:
I agree about the state of the control flow. That is what I was referring to when talking about preventing recalculations. 

How can I trigger "miscalculated deco steps" and "warning mode"? Over the weekend I might find some time to look into this. (Sorry, I was mostly absent the last weeks due to other commitments like watching over kids doing home schooling while trying to get some real work done etc).

bstoeger:
Happens for me every time:
image

Don't waste your time on this - I broke it, so I'll try to fix it.

atdotde:
I see it only in VPM-B mode (which is particularly bad to trace back thanks to its iterative nature). I have a few minutes, let's see if I can bisect it.

bstoeger:
I did my independent bisect (that works nice while writing boring stuff) and it seems that the culprit is
    9bfc6d252 profile: use a parameter to determine planner state
Interestingly, not the commit that I was expecting.

bstoeger:
Ouch - I think I see it - I inverted two(!) of the conditionals. Argh.

atdotde:
That looks pretty much like it. And it is the ceiling in the profile that is wrong not the dive plan and this is why the planner test did not spot this. I am a bit surprised that the test of the profile structure did not trigger...

bstoeger:
Attention: This is now ultra-experimental and I am sure that it breaks the planner in multiple ways. One obvious bug is that the color on the pressure-chart is wrong when entering the planner.

What it does it to remove a fundamental flaw of the code: the display function calculates the dive profile. It reverses this: calculate the profile first and then show it. However, I am sure that I didn't consider all the instances where the profile has to be recalculated, so this needs a close audit / testing.

dirkhh:
also punting on this (marked as highly experimental)

atdotde:
I haven't had time to look at this in detail, yet.

qt-models/diveplannermodel.cpp
```diff
@@ -821,36 +826,72 @@ int DivePlannerPointsModel::addStop(int milimeters, int seconds, int cylinderid_
...
-	if (row >= divepoints.count())
-		return;
+
+	// Refuse creation of two points with the same time stamp.
+	// Note: "time" is moved in the positive direction to avoid
+	// time becoming zero or, worse, negative.
+	while (std::any_of(divepoints.begin(), divepoints.begin() + row,
+			[t = newData.time] (const divedatapoint &data)
+			{ return data.time == t; }))
+		newData.time += 10;
```
atdotde, line 855: Why 10, shouldn't 1 be enough?

atdotde (commented):
I read through the commit messages and the diffs and it all seems to make sense. But haven't yet tested this.

bstoeger:
Added commit to fix the colors on the pressure-line. Feel free to merge by itself (should be pretty obvious).

atdotde:
Thumbs up

profile-widget/profilewidget2.cpp
```diff
@@ -1798,49 +1798,21 @@ void ProfileWidget2::repositionDiveHandlers()
 	}
 }
 
-int ProfileWidget2::fixHandlerIndex(DiveHandler *activeHandler)
```
dirkhh, line 1764: I think this is a reasonable change. I always hated this behavior - as I dislike how the 'grow time' behavior "feels"... but that's unrelated.
Please note that the commit message of 4d055d19 is missing a word... "with the mouse, the XXX tried to be smart"

qt-models/diveplannermodel.cpp
```diff
@@ -844,6 +844,18 @@ void DivePlannerPointsModel::editStop(int row, divedatapoint newData)
 	 */
 	int old_first_cylid = divepoints[0].cylinderid;
 
+	// Refuse creation of two points with the same time stamp.
+	// Note: "time" is moved in the positive direction to avoid
+	// time becoming zero or, worse, negative.
+	while (std::any_of(divepoints.begin(), divepoints.begin() + row,
+			[t = newData.time] (const divedatapoint &data)
+			{ return data.time == t; }))
+		newData.time += 10;
```
dirkhh, line 855: Agree with @atdotde - I'd add just 1 second

dirkhh (approved):
My head hurts. After three starts I made it all the way through. Small comments, but otherwise this looks good. I tested it on desktop and mobile and haven't managed to detect issues.

qt-models/diveplannermodel.cpp
bstoeger: I can't really remember, but I guess I used that since 10 s are used in other places in the same file:
```
                        if (duration <= 0)
                                secs += 10 - duration;
                        p.time = secs;
                        while (++i < divepoints.size())
                                if (divepoints[i].time < divepoints[i - 1].time + 10)
                                        divepoints[i].time = divepoints[i - 1].time + 10;
                        break;
```
and
```
                                int secs = value.toInt() * 60;
                                if (!secs)
                                        secs = 10;
```
and
```
                                if (newtime.seconds == lastrecordedtime.seconds)
                                        newtime.seconds += 10;
```

Of course, this code doesn't ensure that any segment is at least 10 sec. When moving we might fall just one second before or after a different handle and the code is not executed.

I really don't know what the right thing to do is in this case...

bstoeger: PS: Fixed the numerous typos in this particular commit message.

dirkhh: honestly, I don't know what's right, either... If we consistently used 10 seconds elsewhere... maybe there was a reason?
And of course I forgot to note down the SHA that I reviewed, so now I can't easily tell what you changed since I last looked at this. UGH. Wait, I have a local copy. that should work...
Ahh, no. You rebased (not wrong, please don't misunderstand).
Ok, staring at that diff, I think you changed nothing and just fixed commit messages?

PR #3188, opened 2021-02-06
statistics: don't place labels at half-integer values

bstoeger:
Placing labels at half-integer values gives horrible
rendering artifacts. Therefore, always round to integer
values. The easiest way to do this is right before setting
the position. Introduce a helper function to round QPointF
in such scenarios.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes nasty rendering artifacts in the bar-charts. Labels must never be placed at half-integer values, which could happen for bars with an odd-pixel height.


dirkhh (approved):
looks good

PR #3189, opened 2021-02-07
statistics: fix placement of horizontal labels in bar charts

bstoeger:
The subtraction of half the label width, needed for centered
labels, must have been lost somewhere.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes the label placement in bar charts. I didn't investigate when the bug was introduced or if it was there "forever".

### DIFF
--- a/stats/barseries.cpp
+++ b/stats/barseries.cpp
@@ -151,6 +151,7 @@ void BarSeries::BarLabel::updatePosition(bool horizontal, bool center, const QRe
 				setVisible(false);
 				return;
 			}
+			pos.rx() -= itemSize.width() / 2.0;
 		}
 		item->setPos(roundPos(pos)); // Round to integer to avoid ugly artifacts.
 	}


stats/barseries.cpp
```diff
@@ -151,6 +151,7 @@ void BarSeries::BarLabel::updatePosition(bool horizontal, bool center, const QRe
 				setVisible(false);
 				return;
 			}
+			pos.rx() -= itemSize.width() / 2.0;
```
dirkhh, line 154: elsewhere you point out that things look weird if they end up on non-integer coordinates. 
Is that a concern here as well? Should this be rounded or simply done as integer division?

dirkhh (commented):
don't know why I didn't see this yesterday... oh, I do... I ignored my computer yesterday and spent the day playing my bass and recording my daughter on her tuba... such COVID things to do...

stats/barseries.cpp
bstoeger: Not a concern - It's rounded two lines down with the `roundPos()` helper function.

dirkhh: ah, context matters. not enough context on Github

dirkhh (approved):
see earlier resolved conversation

PR #3190, opened 2021-02-07
Subsurface Desktop user manual: Statistics panel

willemferguson:
Update the user manual to take into account the statistics tool.
This is a first attempt. Comments, please!

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @bstoeger


dirkhh:
Thanks for all the work you did on that! Huge step forward.

Your default style is shining through - which is unsurprising... we are who we are. This works well for scientific paper and does a great job conveying the information the user needs. I think I will gently tweak this after merging to make it sound a little more personable (one of the key difference is the passive vs active verb use - 'one can' vs 'you can'). But what you have written from a raw content perspective makes perfect sense and the pictures illustrate your points well.

Thank you!

PR #3197, opened 2021-02-13
[CLEANUP] printing: "fix" progress indicator

bstoeger:
In TemplateLayout, there was a progress indication, which reported
the progress - not of the actual rendering - but of adding the
dives to the "to render" list. Which is of course done in less than
a ms, making the whole thing completely pointless.

Instead, emit progress when actually looping over the dives or
statistics.

Nobody ever noticed the problem because even rendering is done in
fractions of a second and indeed is accounted to only one fifth
of the total progress.

The real purpose of this "fix" is to get rid of the getTotalWork()
function, which was just insane. Instead of asking the TemplateLayout
how many dives it rendered, this number was extracted from
global state. Simply store the number of dives in the TemplateLayout
object instead.

Moreover, fix two coding style issues:
 - "Page" variable identifier starting with a capital
 - The Printer::render() being defined (as opposed to declared) with
   a default parameter. This is not how C++'s default parameters work,
   sorry.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a code cleanup commit that technically fixes a bug, which however is not a real issue. For the bug, see the commit message. The printing code really is some of the worst...
### Changes made:

1) remove silly `getTotalWork()` function.
2) coding style fixes.

### Release note:

No - small part of bigger cleanup work.

### DIFF
--- a/desktop-widgets/printer.cpp
+++ b/desktop-widgets/printer.cpp
@@ -122,7 +122,7 @@ void Printer::flowRender()
 	painter.end();
 }
 
-void Printer::render(int Pages = 0)
+void Printer::render(int pages)
 {
 	// keep original preferences
 	ProfileWidget2 *profile = MainWindow::instance()->graphics;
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
looks sane

PR #3198, opened 2021-02-14
Remove the global AppState and in_planner()

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a further step in removing global state, especially with respect to the profile.

This removes the global `in_planner()` function by passing down whether we are in the planner to the different classes. The ultimate goal is to make the profile reentrant for easier printing and for support of dynamic profile on mobile. This PR is far from perfect, since there still isn't any authoritative source for whether the app is in planner state or not. It's all still very convoluted, but it is getting clearer.

As a side-effect of this PR, the global AppState can be removed. This doesn't make sense in the context of the mobile UI and therefore should be desktop-only.

### Related issues:

See also the related PR #3186.

### Release note:

No.


dirkhh (approved):
Very systematic and structured PR. Quite honestly a joy to review.
Thanks!

PR #3200, opened 2021-02-16
Divelogs location

mturkia:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Streamlining our sync to divelogs.de according to Rainer's change in import of Subsurface XML. The only difference should be that this change is not using trip name as location. (This info is currently not available for the XSLT transform as we are processing individual dives.)

---8<---
if the divesite has something in the "Country" field, this is used for location

else if the divesite entry has slashes ("/"), then that he last one is the divesite and the rest is the location
    example: the divesite reads "Eygpt / Red Sea / Thistlegorm", then the divesite is "Thistlegorm" and the location is "Eygpt / Red Sea"

~~else if there is a trip name, then that is used for location~~

else the location stays empty
---8<---

### DIFF
--- a/xslt/divelogs-export.xslt
+++ b/xslt/divelogs-export.xslt
@@ -42,6 +42,17 @@
       <xsl:value-of select="//site[@uuid = $uuid]/@name"/>
     </xsl:variable>
     <xsl:choose>
+      <xsl:when test="/divelog/divesites/site[@uuid = $uuid]/geo/@value != ''">
+        <LOCATION>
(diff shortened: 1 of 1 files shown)


dirkhh:
Something's wrong with the openSUSE VM - nothing to do with this PR

dirkhh (approved):
looks reasonable

PR #3201, opened 2021-02-16
statistics: rudimentary theming of the color scheme

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Dirk wants to implement a dark-theme for mobile. This PR provides a basis for just that.

For testing purposes, this implements a way to change the theme on desktop, but that finally makes the desktop UI too big.

It also shows our problem with rendering bright fonts onto a dark background. If we find no fix for that, we might have to make the font themeable as well, i.e.~support different font weights for different themes.

@dirkhh: just let me know if you need any more logic. I'll gladly do it and expect that it isn't too hard to implement. However, the color-scheme itself is up to you (or someone else who requires a dark theme).

@willemferguson: this might also interesting to you, since it finally makes the desktop UI too cluttered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 - profile: include profile editing in undo system
+- mobile: Add a dark theme for statistics
 - core: avoid crash with corrupted cloud storage
 - mobile: fix profile scaling issue on high DPR devices
 - mobile: bring back profile icons
--- a/mobile-widgets/statsmanager.cpp
+++ b/mobile-widgets/statsmanager.cpp
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "statsmanager.h"
+#include "themeinterface.h"
 #include "stats/chartlistmodel.h"
 
-StatsManager::StatsManager() : view(nullptr), charts(nullptr)
+StatsManager::StatsManager() : view(nullptr), charts(nullptr), themeInitialized(false)
 {
 	updateUi();
 }
@@ -19,6 +20,9 @@ void StatsManager::init(StatsView *v, ChartListModel *m)
 		fprintf(stderr, "StatsManager::init(): no ChartListModel - statistics will not work.\n");
 	view = v;
 	charts = m;
+
+	connect(ThemeInterface::instance(), &ThemeInterface::currentThemeChanged,
+		this, &StatsManager::themeChanged);
 }
 
 void StatsManager::doit()
@@ -64,6 +68,8 @@ void StatsManager::updateUi()
 	operation2IndexChanged();
 	sortMode1IndexChanged();
 
+	if (view && !std::exchange(themeInitialized, true))
+		themeChanged();
 	if (charts)
 		charts->update(uiState.charts);
 	if (view)
@@ -122,4 +128,15 @@ void StatsManager::setChart(int idx)
 {
 	state.chartChanged(idx);
 	updateUi();
+}
+
+void StatsManager::themeChanged()
+{
+	if (!view)
+		return;
+
+	// We could just make currentTheme accessible instead of
+	// using Qt's inane propertySystem. Whatever.
+	QString theme = ThemeInterface::instance()->property("currentTheme").toString();
+	view->setTheme(theme == "Dark");
 }
--- a/mobile-widgets/statsmanager.h
+++ b/mobile-widgets/statsmanager.h
@@ -49,6 +49,8 @@ signals:
 	void binner2IndexChanged();
 	void operation2IndexChanged();
 	void sortMode1IndexChanged();
+private slots:
+	void themeChanged();
 private:
 	StatsView *view;
 	ChartListModel *charts;
@@ -66,8 +68,8 @@ private:
 	int operation2Index;
 	int sortMode1Index;
 	StatsState::UIState uiState;	// Remember UI state so that we can interpret indexes
+	bool themeInitialized;		// setTheme() crashes if called in init()
 	void updateUi();
-
 };
 
 #endif
--- a/mobile-widgets/themeinterface.cpp
+++ b/mobile-widgets/themeinterface.cpp
@@ -50,12 +50,14 @@ ThemeInterface *ThemeInterface::instance()
 	return self;
 }
 
-ThemeInterface::ThemeInterface()
+ThemeInterface::ThemeInterface() :
+	m_basePointSize(-1.0), // simply a placeholder to declare 'this isn't set, yet'
+	m_currentTheme(qPrefDisplay::theme()),
+	m_needSignals(true) // make sure the signals fire the first time
+
 {
 	// get current theme
-	m_currentTheme = qPrefDisplay::theme();
 	update_theme();
-	m_basePointSize = -1.0; // simply a placeholder to declare 'this isn't set, yet'
 }
 
 void ThemeInterface::set_currentTheme(const QString &theme)
@@ -79,14 +81,12 @@ double ThemeInterface::currentScale()
 
 void ThemeInterface::set_currentScale(double newScale)
 {
-	static bool needSignals = true; // make sure the signals fire the first time
-
 	if (!IS_FP_SAME(newScale, qPrefDisplay::mobile_scale())) {
 		qPrefDisplay::set_mobile_scale(newScale);
 		emit currentScaleChanged();
-		needSignals = true;
+		m_needSignals = true;
 	}
-	if (needSignals) {
+	if (m_needSignals) {
 		// adjust all used font sizes
 		m_regularPointSize = m_basePointSize * newScale;
 		defaultModelFont().setPointSizeF(m_regularPointSize);
@@ -101,7 +101,7 @@ void ThemeInterface::set_currentScale(double newScale)
 		m_titlePointSize = m_regularPointSize * 1.5;
 		emit titlePointSizeChanged();
 
-		needSignals = false;
+		m_needSignals = false;
 	}
 }
 
--- a/mobile-widgets/themeinterface.cpp
+++ b/mobile-widgets/themeinterface.cpp
@@ -4,7 +4,6 @@
 #include "qmlmanager.h"
 #include "core/metrics.h"
 #include "core/settings/qPrefDisplay.h"
-#include <QFontInfo>
 
 static const QColor BLUE_BACKGROUND_COLOR = "#eff0f1";
 static const QColor BLUE_CONTRAST_ACCENT_COLOR = "#FF5722";
--- a/mobile-widgets/themeinterface.h
+++ b/mobile-widgets/themeinterface.h
@@ -87,5 +87,7 @@ private:
 	double m_titlePointSize;
 
 	QString m_currentTheme;
+
+	bool m_needSignals;
 };
 #endif
--- a/mobile-widgets/themeinterface.h
+++ b/mobile-widgets/themeinterface.h
@@ -3,8 +3,6 @@
 #define THEMEINTERFACE_H
 #include <QObject>
 #include <QColor>
-#include <QSettings>
-#include <QQmlContext>
 
 class ThemeInterface : public QObject {
 	Q_OBJECT
--- a/stats/barseries.cpp
+++ b/stats/barseries.cpp
@@ -106,7 +106,7 @@ BarSeries::~BarSeries()
 BarSeries::BarLabel::BarLabel(StatsView &view, const std::vector<QString> &labels, int bin_nr, int binCount) :
 	isOutside(false)
 {
-	QFont f; // make configurable
+	const QFont &f = view.getCurrentTheme().labelFont;
 	item = view.createChartItem<ChartTextItem>(ChartZValue::SeriesLabels, f, labels, true);
 	//highlight(false, bin_nr, binCount);
 }
--- a/stats/barseries.cpp
+++ b/stats/barseries.cpp
@@ -116,16 +116,17 @@ void BarSeries::BarLabel::setVisible(bool visible)
 	item->setVisible(visible);
 }
 
-void BarSeries::BarLabel::highlight(bool highlight, int bin_nr, int binCount, const QColor &background)
+void BarSeries::BarLabel::highlight(bool highlight, int bin_nr, int binCount, const QColor &background, const StatsTheme &theme)
 {
 	// For labels that are on top of a bar, use the corresponding bar color
 	// as background. Rendering on a transparent background gives ugly artifacts.
-	item->setColor(highlight || isOutside ? darkLabelColor : labelColor(bin_nr, binCount),
+	item->setColor(highlight || isOutside ? theme.darkLabelColor : theme.labelColor(bin_nr, binCount),
 		       isOutside ? Qt::transparent : background);
 }
 
 void BarSeries::BarLabel::updatePosition(bool horizontal, bool center, const QRectF &rect,
-					 int bin_nr, int binCount, const QColor &background)
+					 int bin_nr, int binCount, const QColor &background,
+					 const StatsTheme &theme)
 {
 	QSizeF itemSize = item->getRect().size();
 	if (!horizontal) {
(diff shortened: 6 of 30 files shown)


dirkhh:
... trying to understand the comment 
> since it finally makes the desktop UI too cluttered

dirkhh (commented):
So in general, I don't want to enable users to change the colors - we do this on mobile and it's such a nightmare. So I'd remove that UI element. But of course we HAVE that on mobile, so the infrastructure that you build might be necessary anyway.
What we should be able to do on the desktop is to detect the current state (light or dark) and then automagically switch between two schemes.

I'm still catching up with things after the power outage, I'll try to find time for this soon.

bstoeger:
> So in general, I don't want to enable users to change the colors

Well, I need it for testing. Feel free to leave out that particular commit.

Now the fonts can be "themed" as well, since in dark-mode the text was barely readable for me. Please adjust as needed.

willemferguson:
I think in principle it is a good move to be able to have the graph colours fit in with different themes. I have never had much use for a dark theme, but I encounter it quite frequently in Internet documentation these days.
My main question is whether this is the right place to have that setting. If the statistics panel has theme options, so should the profile panel. Maybe this setting should be among the app preferences? Using the theming for the stats panel, my feeling is that the other panels and tabs of Subsurface should be dark as well. At the moment it is a single dark panel in a sea of light panels.
BUT, the Ubuntu dark theme is not really a dark theme at all, so I cannot say much on how is should look in a truly dark theme.
With respect to the dark theme, I do not particularly like the light brown bars. I would have kept the light blue bars and yellow highlighting for both light and dark themes.

dirkhh:
My apologies for continuing to ignore this. I have a few burning things that I need to take care of, first. But this is high on my list once those fires are handled.

dirkhh:
This clearly wasn't high enough on my list of things to handle.
punting until after the release

bstoeger:
I rebased on current master and removed the theme-switch on desktop.

This now switches the theme on mobile.

Of course, the actual dark-theme-commit has to be done (c.f. commit with "--- dummy commit ----" in the title).

@dirkhh: From my side, this is done. You wouldn't be happy with a dark them that I "designed". If you think this is not worth the hassle, feel free to close.

dirkhh:
may not be completely *done*...
```
/Users/runner/work/subsurface/subsurface/stats/statscolors.cpp:158:57: error: implicit instantiation of undefined template 'std::array<std::unique_ptr<const StatsTheme>, 2>'
static std::array<std::unique_ptr<const StatsTheme>, 2> statsThemes;
                                                        ^
/Applications/Xcode_13.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.0.sdk/usr/include/c++/v1/__tuple:219:64: note: template is declared here
template <class _Tp, size_t _Size> struct _LIBCPP_TEMPLATE_VIS array;
                                                               ^
1 error generated.
```

bstoeger:
ping :)

dirkhh:
> ping :)

fair.
I'll look at this within the next hour

dirkhh:
Definitely needs some edits (typos, the dummy commit, etc). I think I'll do that locally and push an update into this branch of that's ok with you?

bstoeger:
Sure.

dirkhh:
Hmmm. I don't think the dark bin colors get used. Phrased differently, when I switch to the dark theme, the bin colors appear to stay unchanged - the whole graph stays unchanged. So something isn't working as expected there (this is on a Mac).
Also, when I then quit the app, I get a crash:
```
Exception Type:        EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Illegal instruction: 4
Termination Reason:    Namespace SIGNAL, Code 0x4
Terminating Process:   exc handler [29846]

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   org.subsurface-divelog        	0x000000010b253e72 std::__1::array<std::__1::unique_ptr<StatsTheme const, std::__1::default_delete<StatsTheme const> >, 2ul>::~array() + 34 (array:127)
1   libsystem_c.dylib             	0x00007fff6fd4113c __cxa_finalize_ranges + 319
2   libsystem_c.dylib             	0x00007fff6fd41412 exit + 55
3   libdyld.dylib                 	0x00007fff6fc97cd0 start + 8

```

dirkhh:
So this is what I see in light and dark. What am I missing?

Screen Shot 2022-03-01 at 2 00 48 PM
Screen Shot 2022-03-01 at 1 59 44 PM

bstoeger:
> So this is what I see in light and dark. What am I missing?

That this is (for now) a mobile-only feature. 🤷

dirkhh:
> > So this is what I see in light and dark. What am I missing?
> 
> That this is (for now) a mobile-only feature. 🤷

snort. I clearly didn't read the code closely enough. Oops.

bstoeger:
> Also, when I then quit the app, I get a crash:

Can't reproduce, sorry. Neither the crash, nor any warnings with valgrind.

Perhaps freeing QFont objects crashes if QApp is gone? I added a commit that leaks the theme objects. Please check if that helps.

dirkhh:
These crashes at exit are a major PITA to debug. Running things in the debugger doesn't really give me any idea what is actually blowing up. And as usual, my testing was done on macOS - which likely is why you don't see this.
Luckily, leaking the `QFont` seems to be enough to prevent the crash.

I've played with this a bit with mobile on desktop - but not yet on device. I'm not convinced that changing the bin colors really is the right thing to do. The existing colors seem just fine, regardless of light or dark theme - the whole idea of these colors is that they are well balanced and far apart from each other. Simply inverting them doesn't seem an improvement.

And the default color for the bars (not quite sure what to name that... I could generously go with 'bronze', even though initially some other words came to mind)... so I certainly need to spend more time on figuring that part out.

bstoeger:
If you think it is not worth it, feel free to close.

> Simply inverting them doesn't seem an improvement.

Well, of course. That's why it's a dummy commit.

dirkhh:
I think this is now no longer awful and moves things in the right direction.
I'm ok with merging

dirkhh:
(assuming the builds pass, @bstoeger doesn't ate it, etc)

bstoeger:
LGTM.

cyberdiamond:
> 



> Also, when I then quit the app, I get a crash:

I also see a crash on exit, not on this commit as I have only been building against master but it has been happening for a while now.

Also see it randomly doing other things but not the same things (except on exit) so hard to reproduce.
I also do see it occasionally on resizing main window.

This is on Fedora / Wayland.

I know its not much help but just another data point.

bstoeger:
Since you build your own, could you perhaps run under valgrind or in gdb? Without backtraces I can't do anything. I certainly don't see these crashes, but I also don't have wayland.

bstoeger:
Ok - I see invalid accesses when exiting with unsaved changes.

The same pattern as usual:

The QUndoStack is a global object whose destructor is run after the QApplication object is destroyed. We really need a recipe against this. It happens too often. Perhaps no global objects at all and add everything to the MainWindow or QMLManager respectively?

dirkhh:
The memory management in Qt is just such a delight. It warms my heart how well designed all of this is.

Having the QUndoStack created in the MainWindow/QMLManager seems like a fairly logical and consistent fix.

cyberdiamond:
> Since you build your own, could you perhaps run under valgrind or in gdb? Without backtraces I can't do anything. I certainly don't see these crashes, but I also don't have wayland.


Fresh build of master then a very basic run just start open local xml log then closed

Let me know if you want anything else.


[steve@t490 src]$ valgrid ./subsurface/build/subsurface
bash: valgrid: command not found...
[steve@t490 src]$ valgrind ./subsurface/build/subsurface
==51751== Memcheck, a memory error detector
==51751== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==51751== Using Valgrind-3.18.1 and LibVEX; rerun with -h for copyright info
==51751== Command: ./subsurface/build/subsurface
==51751== 
QSocketNotifier: Can only be used with threads started with QThread
QObject::connect(QQuickWindow, QDeclarativeGeoMap_QML_6): invalid nullptr parameter
==51751== Invalid read of size 8
==51751==    at 0x2A7A4210: QGnomePlatformDecoration::updateButtonHoverState(Button) (in /usr/lib64/qt5/plugins/wayland-decoration-client/libqgnomeplatformdecoration.so)
==51751==    by 0x2A7A4573: QGnomePlatformDecoration::processMouseTop(QtWaylandClient::QWaylandInputDevice*, QPointF const&, QFlags<Qt::MouseButton>, QFlags<Qt::KeyboardModifier>) (in /usr/lib64/qt5/plugins/wayland-decoration-client/libqgnomeplatformdecoration.so)
==51751==    by 0x2A7A4814: QGnomePlatformDecoration::handleMouse(QtWaylandClient::QWaylandInputDevice*, QPointF const&, QPointF const&, QFlags<Qt::MouseButton>, QFlags<Qt::KeyboardModifier>) (in /usr/lib64/qt5/plugins/wayland-decoration-client/libqgnomeplatformdecoration.so)
==51751==    by 0x199ED008: QtWaylandClient::QWaylandWindow::handleMouseEventWithDecoration(QtWaylandClient::QWaylandInputDevice*, QtWaylandClient::QWaylandPointerEvent const&) (in /usr/lib64/libQt5WaylandClient.so.5.15.2)
==51751==    by 0x199ED423: QtWaylandClient::QWaylandWindow::handleMouse(QtWaylandClient::QWaylandInputDevice*, QtWaylandClient::QWaylandPointerEvent const&) (in /usr/lib64/libQt5WaylandClient.so.5.15.2)
==51751==    by 0x199DC51C: QtWaylandClient::QWaylandInputDevice::Pointer::flushFrameEvent() (in /usr/lib64/libQt5WaylandClient.so.5.15.2)
==51751==    by 0x19B6DC03: ffi_call_unix64 (in /usr/lib64/libffi.so.6.0.2)
==51751==    by 0x19B6D106: ffi_call (in /usr/lib64/libffi.so.6.0.2)
==51751==    by 0x19AC4D0F: ??? (in /usr/lib64/libwayland-client.so.0.3.0)
==51751==    by 0x19AC542A: ??? (in /usr/lib64/libwayland-client.so.0.3.0)
==51751==    by 0x19AC561B: wl_display_dispatch_queue_pending (in /usr/lib64/libwayland-client.so.0.3.0)
==51751==    by 0x199E0852: QtWaylandClient::QWaylandDisplay::flushRequests() (in /usr/lib64/libQt5WaylandClient.so.5.15.2)
==51751==  Address 0xf0000000a is not stack'd, malloc'd or (recently) free'd
==51751== 
==51751== 
==51751== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==51751==  Access not within mapped region at address 0xF0000000A
==51751==    at 0x2A7A4210: QGnomePlatformDecoration::updateButtonHoverState(Button) (in /usr/lib64/qt5/plugins/wayland-decoration-client/libqgnomeplatformdecoration.so)
==51751==    by 0x2A7A4573: QGnomePlatformDecoration::processMouseTop(QtWaylandClient::QWaylandInputDevice*, QPointF const&, QFlags<Qt::MouseButton>, QFlags<Qt::KeyboardModifier>) (in /usr/lib64/qt5/plugins/wayland-decoration-client/libqgnomeplatformdecoration.so)
==51751==    by 0x2A7A4814: QGnomePlatformDecoration::handleMouse(QtWaylandClient::QWaylandInputDevice*, QPointF const&, QPointF const&, QFlags<Qt::MouseButton>, QFlags<Qt::KeyboardModifier>) (in /usr/lib64/qt5/plugins/wayland-decoration-client/libqgnomeplatformdecoration.so)
==51751==    by 0x199ED008: QtWaylandClient::QWaylandWindow::handleMouseEventWithDecoration(QtWaylandClient::QWaylandInputDevice*, QtWaylandClient::QWaylandPointerEvent const&) (in /usr/lib64/libQt5WaylandClient.so.5.15.2)
==51751==    by 0x199ED423: QtWaylandClient::QWaylandWindow::handleMouse(QtWaylandClient::QWaylandInputDevice*, QtWaylandClient::QWaylandPointerEvent const&) (in /usr/lib64/libQt5WaylandClient.so.5.15.2)
==51751==    by 0x199DC51C: QtWaylandClient::QWaylandInputDevice::Pointer::flushFrameEvent() (in /usr/lib64/libQt5WaylandClient.so.5.15.2)
==51751==    by 0x19B6DC03: ffi_call_unix64 (in /usr/lib64/libffi.so.6.0.2)
==51751==    by 0x19B6D106: ffi_call (in /usr/lib64/libffi.so.6.0.2)
==51751==    by 0x19AC4D0F: ??? (in /usr/lib64/libwayland-client.so.0.3.0)
==51751==    by 0x19AC542A: ??? (in /usr/lib64/libwayland-client.so.0.3.0)
==51751==    by 0x19AC561B: wl_display_dispatch_queue_pending (in /usr/lib64/libwayland-client.so.0.3.0)
==51751==    by 0x199E0852: QtWaylandClient::QWaylandDisplay::flushRequests() (in /usr/lib64/libQt5WaylandClient.so.5.15.2)
==51751==  If you believe this happened as a result of a stack
==51751==  overflow in your program's main thread (unlikely but
==51751==  possible), you can try to increase the size of the
==51751==  main thread stack using the --main-stacksize= flag.
==51751==  The main thread stack size used in this run was 8388608.
==51751== 
==51751== HEAP SUMMARY:
==51751==     in use at exit: 169,434,755 bytes in 442,760 blocks
==51751==   total heap usage: 10,892,552 allocs, 10,449,792 frees, 1,496,399,262 bytes allocated
==51751== 
==51751== LEAK SUMMARY:
==51751==    definitely lost: 296 bytes in 24 blocks
==51751==    indirectly lost: 16,496 bytes in 2 blocks
==51751==      possibly lost: 2,610,398 bytes in 8,629 blocks
==51751==    still reachable: 166,729,957 bytes in 433,399 blocks
==51751==                       of which reachable via heuristic:
==51751==                         newarray           : 167,936 bytes in 216 blocks
==51751==                         multipleinheritance: 266,256 bytes in 419 blocks
==51751==         suppressed: 0 bytes in 0 blocks
==51751== Rerun with --leak-check=full to see details of leaked memory
==51751== 
==51751== For lists of detected and suppressed errors, rerun with: -s
==51751== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
Segmentation fault (core dumped)
[steve@t490 src]$

bstoeger:
Many thanks.

That looks like a Qt-on-wayland bug to me. At least I don't see any of our code in the stack traces. I can't exclude that this is due to memory smashing from our side, of course.

Maybe related: https://retrace.fedoraproject.org/faf/problems/376279

@dirkhh?

dirkhh:
I have seen a few comments in the Qt forums about Wayland bugs - and it seems that those won't be addressed in Qt 5 anymore (and of course Qt 6, which may or may not at some point fix these things, is missing several critical features that we need for Subsurface). So... as usual... Linux as a desktop is just a mess.
I

bstoeger:
> I have seen a few comments in the Qt forums about Wayland bugs - and it seems that those won't be addressed in Qt 5 anymore (and of course Qt 6, which may or may not at some point fix these things, is missing several critical features that we need for Subsurface).

Wow, that feels "unprofessional". At least that explains why Kubuntu hasn't switched to Wayland yet.

dirkhh:
so... my limited understanding of all this is that the Qt community feels that the distros broke the "platform promise".
Something that ran on the platform just fine yesterday, no longer runs on it today. And I find it hard to argue with that.
Of course, that happens all the time - but here it's annoying because it appears to cause crashes.
I assume you run Kubuntu and therefore aren't affected?

dirkhh:
having said all this - I see nothing in this PR that changes the behavior when it comes to Wayland.
And I no longer can reproduce crashes here. So I'm leaning towards merging this.

dirkhh:
do you want to rebase on master (since the conflicts are all with your own changes, anyway), or should I just do this locally?

bstoeger:
> I assume you run Kubuntu and therefore aren't affected?

Correct. Still on XOrg. I didn't follow all the XOrg/Wayland developments, so I have basically no idea what is going on there politically or technically and why porting from one to the other is so painful. X works just fine for me, but I understand that for some users it is not adequate.

I rebased on master, doing some detours. Please check.

dirkhh:
lovely -- thanks for doing that.

Issue #3182, opened 2021-01-23
Feature request: CCR setpoint changes (default & manual) during calculated ascents


justin-ashworth-research:
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [x] New feature request
- [ ] Discussion request

### Issue long description:

When editing a CCR dive in the planner, we would like for setpoint changes to be included (or a parameter) in the auto-magically generated ascent plan. 

For example, it is typical to adjust our ppO2 setpoint up to 1.4 during intermediate decompression stops, and up to 1.6 during the 6 metre decompression stop (i.e. oxygen rebreather mode). This common practice dramatically accelerates decompression and decreases runtime.

I have not yet examined the codebase to decide whether it would be appropriate for me to attempt a pull request.

### Subsurface version:

4.9.10

atdotde:
I understand this is a feature that could be useful. As you might have noticed, the user interface of the planner is already quite (some say far too) complex. How would you imagine to configure such set point change from a UI perspective? Would that happen at a specific depth/stop or how are the details determined. Please take into account that such a feature should not only scratch your personal itch but if implemented should help many users. Could you reasonably expect others to have different strategies here? How can we make most of the people happy without adding a million more configurable options that will confuse the rest of our users?

justin-ashworth-research:
I made a fork out of interest and added a function that seems to accomplish this in core/planner.c:
https://github.com/justin-ashworth-research/subsurface/commit/b5422eb7cf1941d8dba682a35f17b0fd7b8d926b

However I have not attempted to make this flexible within the interface.

justin-ashworth-research:
I put the setpoint updates too late in plan() in the above commit--too clean! (and not updating the deco)

justin-ashworth-research:
This seems to accelerate the deco... akin to what we aim for on some tech CCR dives:
https://github.com/justin-ashworth-research/subsurface/commit/4e3f31c99df3fe7394569bc83d6b9191d87d5d20

The plan() function is quite verbose and monolithic! Impressive work

atdotde:
As I said: the challenge is not to incorporate this in the planning but to come up with a decent user interface.

justin-ashworth-research:
Yes, perhaps a single checkbox. In the CCR tech deco courses it is typically taught as setpoint to 1.4 at 21 metres, and to 1.6 at 6 metres (O2 flush/O2 rebreather mode). 
[...]

PR #3202, opened 2021-02-21
Planner: Depth dependent setpoint changes in CCR mode

atdotde:
We had a user request to allow for setpoint changes
at certain depths for CCR deco.

You can now enter a cylinder with name like
"SP 1.4" ('S' and 'P' and ' ' and a float) with
a switch depth and that cylinder is interpreted as
a depth dependent setpoint switch.

This user interface is a hack. But I believe that such
setpoint changes are similar enough to gas switches during
deco and should thus be handled in a simiar manner.

I would be happy to hear ideas how this could be made
less easter eggish.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/planner.c
+++ b/core/planner.c
@@ -382,7 +382,23 @@ struct gaschanges {
 	int gasidx;
 };
 
-static struct gaschanges *analyze_gaslist(struct diveplan *diveplan, struct dive *dive, int *gaschangenr, int depth, int *asc_cylinder)
+// Return new setpoint if cylinderi is a setpoint change an 0 if not
+
+static int setpoint_change(struct dive *dive, int cylinderid)
+{
+	cylinder_t *cylinder = get_cylinder(dive, cylinderid);
+	if (!cylinder->type.description)
+		return 0;
+	if (!strncmp(cylinder->type.description, "SP ", 3)) {
+		float sp;
+		sscanf(cylinder->type.description + 3, "%f", &sp);
+		return (int) (sp * 1000);
+	} else {
+		return 0;
+	}
+}
+
+static struct gaschanges *analyze_gaslist(struct diveplan *diveplan, struct dive *dive, int *gaschangenr, int depth, int *asc_cylinder, bool ccr)
 {
 	int nr = 0;
 	struct gaschanges *gaschanges = NULL;
@@ -390,7 +406,7 @@ static struct gaschanges *analyze_gaslist(struct diveplan *diveplan, struct dive
 	int best_depth = get_cylinder(dive, *asc_cylinder)->depth.mm;
 	bool total_time_zero = true;
 	while (dp) {
-		if (dp->time == 0 && total_time_zero) {
+		if (dp->time == 0 && total_time_zero && (ccr == (bool) setpoint_change(dive, dp->cylinderid))) {
 			if (dp->depth.mm <= depth) {
 				int i = 0;
 				nr++;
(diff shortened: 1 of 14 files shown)


core/planner.c
```diff
@@ -382,15 +382,31 @@ struct gaschanges {
 	int gasidx;
 };
 
-static struct gaschanges *analyze_gaslist(struct diveplan *diveplan, struct dive *dive, int *gaschangenr, int depth, int *asc_cylinder)
+// Return new setpoint if cylinderi is a setpoint change an 0 if not
+
+static int setpoint_change(struct dive *dive, int cylinderid)
+{
+	cylinder_t *cylinder = get_cylinder(dive, cylinderid);
+	if (cylinder->type.description[0] == 'S' &&
```
bstoeger: Not a C-programmer, but I believe that `strncmp(cylinder->type.description, "SP ", 3) == 0` would be more idiomatic and also not invoke undefined behavior, should the length of `cylinder->type.description` be less than three.

Can `cylinder->type.description` ever be null?

atdotde:
Addressed @bstoeger 's good remarks

dirkhh (commented):
I think this makes sense.
The parsing of the cylinder name is indeed a hack (and not all that robust), but given that this is likely for a small audience, I'm fine with it.
What this does need is a CHANGELOG entry and most importantly documentation in the user manual, though.

bstoeger:
The UI is clearly very weird. Just for my understanding: how does that differ from the current "Add set point change" feature?

atdotde:
The difference is that you can add a set point change only for the manually entered part of the dive. The new feature makes the planner add a set point change during the generated ascent.

justin-ashworth-research:
I tested this, it does work. The user also has to realize that the depth-dependence will be set according to the %O2 given for the "SP" "gas."

Much quicker than a C++ refactor in which Diveplan becomes a class that privately manages all access to divedatapoint(s), as well as handling the accounting of duration vs. runtime--which would be one cleaner way to get "timeless" steps into the dive plan logic.

A remaining issue is that the CNS% and OTU are over-calculated for the steps shallower than 6 metres. The logic in the new function needs to automatically enforce that the maximum setpoint/ppO2 at 3 metres is actually 1.3 (100% * 1.3 ata)

atdotde:
Now with documentation (both manual and CHANGELOG). Turns out, two tests were affected by the change to OTU and CNS calculations (the part where ata was changed to bar, I don't know why we ever had ata there).

Fixes #3182

justin-ashworth-research:
That's great, I checked and this additional shallow depth correction is working in the planner (elegantly enough)

A remaining item in those functions calculate_otu() and calculate_cns_dive() is that the if/else and "trueo2" logics miss the correct OC gas po2 when the dive mode is bailout. I managed to get it counted right in these fork commits https://github.com/justin-ashworth-research/subsurface/commit/12837e16f66dbdcd9c2b74761688d649a4b32f6c and https://github.com/justin-ashworth-research/subsurface/commit/6640b01162f9e8c124e3448ee8a64937836d8545 (small fix for the first commit)

justin-ashworth-research:
I didn't even realize that I could edit gas switch depths directly until now. Thanks for the documentation 🙏

justin-ashworth-research (commented):
calculate_otu() has a similar issue, I think.

I was checking dc->divemode for each/all of the if() conditions that check setpoints--not just the first case (sample->o2sensor)

trueo2 seems an unnecessary variable, if the compound if/else statement(s) are simplified into a single level.

atdotde:
@dirkhh I think this openSUSE fail is cause by other things and this is ready to merge.

dirkhh (approved):
This looks good and should be ready to merge

PR #3206, opened 2021-02-26
Vpmbprofiletest

atdotde:
Add a test for the profile elements in VPMB mode.  

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -997,7 +997,7 @@ static void calculate_deco_information(struct deco_state *ds, const struct deco_
 	int prev_deco_time = 10000000, time_deep_ceiling = 0;
 	bool in_planner = planner_ds != NULL;
 
-	if (!planner_ds) {
+	if (!in_planner) {
 		ds->deco_time = 0;
 		ds->first_ceiling_pressure.mbar = 0;
 	} else {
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
nothing in here strikes me as wrong, but I haven't used the planner in at least 5 years, I have never used VPM for anything... I feel like I have no idea, so I'm not sure I'm the best person to review this.
Generally I view Robert as the maintainer of all of the planner / deco code

dirkhh:
What's going on with the GitHub actions? The openSUSE one was failing, but it looks like no actions ran on this PR???

atdotde:
openSUSE was broken for quite a while. What I am more worried about is that TestProfile did not catch previous problems. My suspicion is that it did not properly run recently.

dirkhh:
> openSUSE was broken for quite a while. What I am more worried about is that TestProfile did not catch previous problems. My suspicion is that it did not properly run recently.

my comment was about the fact that this PR has just one action, one check... not of the compile checks were triggered somehow.
And now I figured out why. This is a PR into @bstoeger 's branch, not into master!!!

dirkhh:
What would be a good ReleaseNotes entry for this fix? I'm on the process of pulling this all together (as you can see in master)

atdotde:
I don't think this needs an entry as this fixes a problem that was introduced after the last release.

dirkhh:
perfect. thanks

PR #3213, opened 2021-03-06
A few Coverity fixes

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A few coverity fixes. One might even be a bug (uninitialized GPS coordinates).

Overall, I'm not very motivated to fix all of these. For example, the `ChartItem`s are supposed to be low-level objects, akin to pure C structures. Therefore, I see it as bad style but not a defect if not all members are initialized in the constructor. Use it without properly initializing it and you'll get random behavior. On the other hand, it certainly doesn't make a measurable difference. *shrug*

It's all untested, BTW.

### DIFF
--- a/stats/informationbox.cpp
+++ b/stats/informationbox.cpp
@@ -12,7 +12,8 @@ static const int distanceFromPointer = 10; // Distance to place box from mouse p
 InformationBox::InformationBox(StatsView &v) :
 	ChartRectItem(v, ChartZValue::InformationBox,
 		      QPen(informationBorderColor, informationBorder),
-		      QBrush(informationColor), informationBorderRadius)
+		      QBrush(informationColor), informationBorderRadius),
+	width(0.0), height(0.0)
 {
 }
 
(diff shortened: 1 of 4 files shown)


stats/informationbox.cpp
```diff
@@ -12,7 +12,8 @@ static const int distanceFromPointer = 10; // Distance to place box from mouse p
 InformationBox::InformationBox(StatsView &v) :
 	ChartRectItem(v, ChartZValue::InformationBox,
 		      QPen(informationBorderColor, informationBorder),
-		      QBrush(informationColor), informationBorderRadius)
+		      QBrush(informationColor), informationBorderRadius),
+	width(0.0), height(0.0)
```
dirkhh, line 16: that seems to have a whitespace issue?

dirkhh (approved):
As usual, many of the coverity warnings may not point to actual bugs (at least the last one actually sounds like there was a real bug). But all of them seem to be good coding style, anyway.
There is the one small whitespace issue.
Do you want me to just clean that up when merging?

bstoeger:
> Do you want me to just clean that up when merging?

Yes, thanks.

dirkhh:
I did merge this manually and forgot to close it here

PR #3223, opened 2021-03-30
stats: fix visibility check of the statistics tab on desktop

bstoeger:
Apparently, the visibility flag of the view is not inherited
from the statistics widget. Therefore, the statistics is
redrawn on every action even if not visible.

Set the visibility explicitly in the show- and hide-events.

This is crazy.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I don't know whether this is a Qt bug or not, but at least on desktop the `isVisible()` function of the QQuickItem does not work and we have to set it ourselves. One would think that the visibility is passed down the object hierarchy!? Very annoying.

### Changes made:

1) Explicitly set the visibility of the 

### Additional information:

Will also have to check on mobile.

### DIFF
--- a/desktop-widgets/statswidget.cpp
+++ b/desktop-widgets/statswidget.cpp
@@ -93,6 +93,8 @@ StatsWidget::StatsWidget(QWidget *parent) : QWidget(parent)
 	view = qobject_cast<StatsView *>(root);
 	if (!view)
 		qWarning("Oops. The root of the StatsView is not a StatsView.");
+	if (view)
+		view->setVisible(isVisible()); // Synchronize visibility of widget and QtQuick-view.
 }
 
 // Initialize QComboBox with list of variables
@@ -209,6 +211,17 @@ void StatsWidget::showEvent(QShowEvent *e)
 	unrestrict();
 	updateUi();
 	QWidget::showEvent(e);
+	// Apparently, we have to manage the visibility of the view ourselves. That's mad.
+	if (view)
+		view->setVisible(true);
+}
+
+void StatsWidget::hideEvent(QHideEvent *e)
+{
+	QWidget::hideEvent(e);
+	// Apparently, we have to manage the visibility of the view ourselves. That's mad.
+	if (view)
+		view->setVisible(false);
 }
 
 void StatsWidget::restrict()
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -824,6 +824,9 @@ void process_loaded_dives()
 
 	/* Inform frontend of reset data. This should reset all the models. */
 	emit_reset_signal();
+
+	/* Now that everything is settled, select the newest dive. */
+	select_newest_visible_dive();
 }
 
 /*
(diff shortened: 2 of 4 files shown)


desktop-widgets/statswidget.cpp
```diff
@@ -93,6 +93,8 @@ StatsWidget::StatsWidget(QWidget *parent) : QWidget(parent)
 	view = qobject_cast<StatsView *>(root);
 	if (!view)
 		qWarning("Oops. The root of the StatsView is not a StatsView.");
+	if (view)
+		view->setVisible(isVisible()); // Start in non-visible state.
```
dirkhh: that `setVisible(isVisible())` call confuses me. Why does that make sense?
Especially given the comment? The comment would match a `setVisible(false)` call, right?

desktop-widgets/statswidget.cpp
```diff
@@ -209,6 +211,17 @@ void StatsWidget::showEvent(QShowEvent *e)
...
+	if (view)
+		view->setVisible(true);
+}
+
+void StatsWidget::hideEvent(QHideEvent *e)
+{
+	QWidget::hideEvent(e);
+	// Apparently, we have to manage the visibility of the view ourselves. That's mad.
+	if (view)
+		view->setVisible(false);
```
dirkhh, line 224: that's sad, but it shouldn't be harmful.

dirkhh (commented):
I'm confused by the first hunk... can you elaborate?

bstoeger:
Please test this - it should fix the post-free access in a more reliable manner. Of course, everything selection is inherently dangerous. But it now separates into well-defined steps:

1) Clear old selection (sends selection changed signal).
2) Clear dive data (sends reset signal).
3) Load dive data (sends reset signal).
4) Select first dive (sends selection changed signal).

Thus, no more selection-changed signal during reset signal.

It seems to work, but I haven't checked interactions with mobile.

desktop-widgets/statswidget.cpp
```diff
@@ -93,6 +93,8 @@ StatsWidget::StatsWidget(QWidget *parent) : QWidget(parent)
 	view = qobject_cast<StatsView *>(root);
 	if (!view)
 		qWarning("Oops. The root of the StatsView is not a StatsView.");
+	if (view)
+		view->setVisible(isVisible()); // Synchronize visibility of widget and QtQuick-view.
```
dirkhh, line 97: ahh, that comment makes more sense. Thanks - didn't realize you had pushed a new version

dirkhh (approved):
tested and seems to work well.
Any concerns if I create a 5.0.1 with just this fix?
I really want to get all those Windows crashes addressed

bstoeger:
You should at least include 378c1eb2cabffd0163fbff9f9cbec54a9b1f284d.

atdotde:
Please include #3186 as well. Even if it's not that well tested, it does fix a crash upon "Edit dive in planner".

dirkhh:
> Please include #3186 as well. Even if it's not that well tested, it does fix a crash upon "Edit dive in planner".

So here's the problem... reading through the comments this includes gems like "breaks mobile"... "largely untested"...
I'm happy to do a 5.0.2 with that one if we can spend a few days to get it ready for prime time. But for now I think the most reasonable step is to cherry pick 378c1eb - because yes, that one is obvious :)

PR #3226, opened 2021-04-02
printing: use sensible font-size in profiles

bstoeger:
The font-size in printed profiles is based on the size of the profile
in the main window. This makes no sense. Why should changing the
window size change the font-size on printouts?

Matter of fact, when making shrinking the height of the window to
its minimum, comical printouts are obtained (font way too big).

Therefore use an arbitrary rule: Say that profiles 600 pixels high
look reasonable and then scale up to the actual size on the printout.

This may need some tweaking for high-DPI mode. But that seems not
to be supported on desktop anyway?

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The printing code produces comical profiles if the window is made very "short", i.e. the height is made small. It is unclear why the printout should depend on window size? Therefore arbitrarily say that a 600-pixel high profile looks reasonable and use that. In the future, this may have to take high-dpi settings into account. But as far as I see, currently this is not supported on desktop anyway..?

PS: I need this to progress on #3186, but it is independent, so doing a separate PR.

### Changes made:

1) Use a fixed profile height instead of the current windows size.

### DIFF
--- a/desktop-widgets/printer.cpp
+++ b/desktop-widgets/printer.cpp
@@ -150,8 +150,12 @@ void Printer::render(int pages)
 
 	QSize originalSize = profile->size();
 	if (collection.count() > 0) {
-		printFontScale = (double)collection.at(0).geometry().size().height() / (double)profile->size().height();
-		profile->resize(collection.at(0).geometry().size());
+		// A "standard" profile has about 600 pixels in height.
+		// Scale the fonts in the printed profile accordingly.
+		// This is arbitrary, but it seems to work reasonably.
+		QSize size = collection[0].geometry().size();
+		printFontScale = size.height() / 600.0;
+		profile->resize(size);
 	}
 	profile->setFontPrintScale(printFontScale);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-
+printing: use sensible font size even for strange window size
 
 ---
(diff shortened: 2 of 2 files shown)


desktop-widgets/printer.cpp
```diff
@@ -150,8 +150,12 @@ void Printer::render(int pages)
...
 	QSize originalSize = profile->size();
 	if (collection.count() > 0) {
-		printFontScale = (double)collection.at(0).geometry().size().height() / (double)profile->size().height();
-		profile->resize(collection.at(0).geometry().size());
+		// A "standard" profile has about 600 pixels in height.
+		// Scale the fonts in the printed profile accordingly.
+		// This is arbitrary, but it seems to work reasonably.
+		QSize size = collection[0].geometry().size();
+		printFontScale = size.height() / 600.0;
+		profile->resize(size);
```
dirkhh, line 158: I can see the logic - but I think the whole size logic of 'everything' for printing is just a hack on top of a kludge on top of a bad design...
Happy to merge this as it doesn't make things worse AFAICT, but the icon sizes and font sizes really need to be fixed - because fonts aren't the only things that are using weird heuristics - we have so many reports that the tank change icons are way off at times.

dirkhh (approved):
I'll take it, but as I said - I think there's a lot more work to be done here

PR #3230, opened 2021-04-12
Cloud storage updates

dirkhh:

- [x] New feature

As I am moving the cloud server around, I came up with an idea to have a hot standby instead of just a backup.
I'm sure this needs more work on the infrastructure side, but first tests have been somewhat promising.

In order to use that, the clients need to be able to use different servers. This implements the basics of that. Pick a server based on geo location (because that seemed to make sense), and make sure that downloading and uploading to the different servers works and that the backend stays in sync.

Of course the next step then would be to detect if a server is down and use the / an other one.

But let's start with what I have so people can review the code and maybe even play with it.

The goal is to have different servers in different geos, with names with a `ssrf-cloud-XX.subsurface-divelog.org` pattern. So far we have US and EU (which is hilariously inconsistent, of course - country vs. continent). But you get the idea. Interestingly, neither of those two servers is actually the server that we are all currently using as `cloud.subsurface-divelog.org` - so indeed, right now there are actually three servers in use...

I'd love review and comments.

### Related issues:

the random cloud outages we had over the years...

### Release note:

missing

### Documentation change:

I'm trying really hard to make this transparent to the user so that there is nothing for us to document

### DIFF
--- a/core/checkcloudconnection.cpp
+++ b/core/checkcloudconnection.cpp
@@ -35,48 +35,51 @@ bool CheckCloudConnection::checkServer()
 	if (verbose)
 		fprintf(stderr, "Checking cloud connection...\n");
 
-	QTimer timer;
-	timer.setSingleShot(true);
 	QEventLoop loop;
-	QNetworkRequest request;
-	request.setRawHeader("Accept", "text/plain");
-	request.setRawHeader("User-Agent", getUserAgent().toUtf8());
-	request.setRawHeader("Client-Id", getUUID().toUtf8());
-	request.setUrl(QString(prefs.cloud_base_url) + TEAPOT);
 	QNetworkAccessManager *mgr = new QNetworkAccessManager();
-	reply = mgr->get(request);
-	connect(&timer, &QTimer::timeout, &loop, &QEventLoop::quit);
-	connect(reply, &QNetworkReply::finished, &loop, &QEventLoop::quit);
-	connect(reply, &QNetworkReply::sslErrors, this, &CheckCloudConnection::sslErrors);
-	for (int seconds = 1; seconds <= prefs.cloud_timeout; seconds++) {
-		timer.start(1000); // wait the given number of seconds (default 5)
-		loop.exec();
-		if (timer.isActive()) {
-			// didn't time out, did we get the right response?
-			timer.stop();
-			if (reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt() == HTTP_I_AM_A_TEAPOT &&
-			    reply->readAll() == QByteArray(MILK)) {
-				reply->deleteLater();
-				mgr->deleteLater();
-				if (verbose > 1)
-					qWarning() << "Cloud storage: successfully checked connection to cloud server";
-				return true;
+	do {
+		QNetworkRequest request;
+		request.setRawHeader("Accept", "text/plain");
+		request.setRawHeader("User-Agent", getUserAgent().toUtf8());
+		request.setRawHeader("Client-Id", getUUID().toUtf8());
+		request.setUrl(QString(prefs.cloud_base_url) + TEAPOT);
+		reply = mgr->get(request);
+		QTimer timer;
+		timer.setSingleShot(true);
+		connect(&timer, &QTimer::timeout, &loop, &QEventLoop::quit);
+		connect(reply, &QNetworkReply::finished, &loop, &QEventLoop::quit);
+		connect(reply, &QNetworkReply::sslErrors, this, &CheckCloudConnection::sslErrors);
+		for (int seconds = 1; seconds <= prefs.cloud_timeout; seconds++) {
+			timer.start(1000); // wait the given number of seconds (default 5)
+			loop.exec();
+			if (timer.isActive()) {
+				// didn't time out, did we get the right response?
+				timer.stop();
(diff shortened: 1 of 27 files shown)


core/qthelper.cpp
```diff
@@ -1455,6 +1456,15 @@ extern "C" char *cloud_url()
 	return copy_qstring(filename);
 }
 
+extern "C" const char *normalize_cloud_name(const char *remote_in)
+{
+	// replace ssrf-cloud-XX.subsurface... names with cloud.subsurface... names
+	// that trailing '/' is to match old code
+	QString ri(remote_in);
+	ri.replace(QRegularExpression(CLOUD_HOST_PATTERN), CLOUD_HOST_GENERIC "/");
```
atdotde, line 1464: where are those constants defined?

atdotde:
Looks good, but there's a typo in aa53543's commit message (to->too).

How do you plan to keep the servers in sync?

core/qthelper.cpp
dirkhh: https://github.com/subsurface/subsurface/blob/CloudStorageUpdates/core/git-access.h#L18-L21

dirkhh:
> Looks good, but there's a typo in [aa53543](https://github.com/subsurface/subsurface/commit/aa5354328b497302b5d0acdc535e61301a102f30)'s commit message (to->too).
> 
> How do you plan to keep the servers in sync?

Will fix the typo.

Every single repo shares a `post-update` hook that immediately pushes the changes in that one repo to the other servers.
You may remember the comment on the mailing list a few days ago how saves to the cloud were stalling. That was when one of the servers fell off the VPN and that push had to time out :-)

I can still come up with one scenarios where this could go wrong and cause a mess.

User has two device. The devices have gotten out of sync (so conflicting changes on each). For some reason the two devices point at two different servers (for example, one is a phone that the user had with them on an international trip and they haven't synced the phone, since). Now the user hits 'save to cloud' on both at the same time (and I mean literally within the same second or two). Likely both updates will succeed, but the pushes to update the other server will fail and the servers are now out of sync.

This is of course a rather artificial scenario, but not entirely impossible. Right now I get an email if a push fails and I could contact the user and offer to fix things. I'm also thinking about ways to automate this.

Things to note in this PR: 
- this hasn't been tested at all on mobile and is definitely missing code for that.
- this has to be extended to fail over to the other server(s) if a server is unreachable.

Also, the backend code needs a way to figure out when things have gotten out of sync and resync them automatically

Maybe I should mark this as WIP :)

dirkhh:
Digging into the failed tests was enlightening.
Some of our tests never initialized the preferences.
And since the code does away with checking if the cloud_base_url is NULL (as that should never be a valid state) those tests that try to parse files will crash now.
Ok, two solutions:
(a) do the right thing and initialize the preferences - because that way we test our code the same way that it is run in normal execution
(b) put those NULL checks back to hide the problem
Obviously I went with (a). The problem? A few of our tests are actually designed to work with no default preferences set. For example the "correct" output for the tests in TestProfile shows no `mod` for the dives (I haven't tracked down why that is... ) - and initializing the preferences therefore leads to different output.
For now I'll do a weird hybrid - I'll initialize the `cloud_base_url`
But I think it would be worth while to actually run the tests in the same state as the app is in, i.e. with reasonable preferences set.
@atdotde thoughts?

dirkhh:
With the latest commit this gets a bit more interesting - this now not only tries to pick the server that's "closer" to you, it also tries the other server(s) if the one that's currently set up for you fails to respond.

So in theory this should allow things to keep working if one of the cloud servers is down (of course, behind the scenes they still share an authentication database and when that is down nothing will work, but this should at least be resilient to little hiccups with the storage servers.

atdotde:
Thumbs up from here. Fixing TestProfile is easy: After it failed simply copy exportprofile.csv  and exportprofileVPMB.csv to ../dives and commit that (of course only if you are sure the change was intentional you you didn't break anything.

dirkhh:
This has now received a fair bit of testing with both desktop and mobile versions.

core/checkcloudconnection.cpp
```diff
@@ -91,6 +94,38 @@ void CheckCloudConnection::sslErrors(const QList<QSslError> &errorList)
...
+	struct serverTried {
+		const char *server;
+		bool tried;
+	};
+	static struct serverTried cloudServers[] = {
+		{ CLOUD_HOST_EU, false },
+		{ CLOUD_HOST_US, false }
+	};
+	const char *server = nullptr;
+	for (int i = 0; i < CLOUD_NUM_HOSTS; i++) {
```
torvalds: Ugh. This loop will start failing in odd ways (possible random memory corruption etc) if CLOUD_NUM_HOSTS is ever changed without changing the cloudServers[] array.

Which is really easy to do by mistake when adding a server.

So rather than have a separate #define for a number in a header file that can get out of sync, why not use ARRAY_SIZE() (currently only used by the git code in load-git.c).

torvalds (commented):
Looks sane. I'm not sure it made sense to remove the NULL pointer checks considering how they actually ended up being valid, but that's a design decision.

The only thing I really reacted to was that CLOUD_NUM_HOSTS issue that I think is bad practice, but not a bug.

core/checkcloudconnection.cpp
dirkhh: BRILLIANT. Good catch. Yes. Thanks. Will do.

PR #3231, opened 2021-04-20
xml parsing: add XML_PARSE_HUGE flag to xmlReadMemory()

torvalds:
It looks like libxml2 has some internal limitations by default that
causes parse failures in some situations.  Avoid them with
XML_PARSE_HUGE.

Without this, you get errors like

    test.xml:349250: parser error : internal error: Huge input lookup
    όμουν τουλάχιστον αλλά +2kg και ενδεχομένως +4
                                                                                   ^
when something in the xml file grows too large.

I don't know libxml2 internals, so I have no idea what exactly goes
wrong, but the docs say:

    XML_PARSE_HUGE = 524288 : relax any hardcoded limit from the parser

and that makes us successfully parse the Greek file from Kostas.

Reported-by: Kostas Katsioulis <kostaskatsioulis@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/parse-xml.c
+++ b/core/parse-xml.c
@@ -1748,9 +1748,9 @@ int parse_xml_buffer(const char *url, const char *buffer, int size,
 	state.sites = sites;
 	state.devices = devices;
 	state.filter_presets = filter_presets;
-	doc = xmlReadMemory(res, strlen(res), url, NULL, 0);
+	doc = xmlReadMemory(res, strlen(res), url, NULL, XML_PARSE_HUGE);
 	if (!doc)
-		doc = xmlReadMemory(res, strlen(res), url, "latin1", 0);
+		doc = xmlReadMemory(res, strlen(res), url, "latin1", XML_PARSE_HUGE);
 
 	if (res != buffer)
 		free((char *)res);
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
looks reasonable

atdotde:
Now, Coverty complains that with this option, users could try to parse files that blow up exponentially in space which would probably kill the process. Do we care? Do we ever parse xml files in situations where the supplier of such a crafted file would shoot into anybody's but her own foot?

Subsurface-divelog:
Linus and I discussed that and came to the conclusion that the type of structure that can cause the billion laughs issue aren't used in our file format. So in regular use this isn't an issue. And yes, a malicious user could send one of us a file that's created like that, but what's the worst that could happen? Subsurface crashes when you try to open that file. Not really something I'd lose sleep over.
We came to the conclusion that the benefits of this change far outweigh the potential downsides.

atdotde:
Good. I just wanted to make sure that there is no situation where a user somehow uploads such a hand crafted file and brings on of your servers (for example the one for the web based logs) down. But probably, one never uploads xml, just via git.

PR #3233, opened 2021-04-25
profile: ignore animation-speed setting when printing

bstoeger:
When printing, the animation speed was set to 0 by the
caller and later reset to the original value. Instead of
modifying global state, set it internally (in the profile-code)
to zero when in print mode.

This is another small step in making the printing independent
from the shown profile.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A (seemingly?) trivial disentangling of printing and profile: Don't use the global preferences flag to stop animations when printing. Simply set the speed to zero when in print mode.

Note that there are a few more callers of `qPrefDisplay::animation_speed()`. However, these are in hover-event functions or concern the "tooltip", which means that they are guaranteed to happen in UI context, not during printing.

Includes another random cleanup commit.

### Release note:

Nope.

### Documentation change:

Nope.

### DIFF
--- a/desktop-widgets/printer.cpp
+++ b/desktop-widgets/printer.cpp
@@ -128,7 +128,6 @@ void Printer::render(int pages)
 	// keep original preferences
 	ProfileWidget2 *profile = MainWindow::instance()->graphics;
 	int profileFrameStyle = profile->frameStyle();
-	int animationOriginal = qPrefDisplay::animation_speed();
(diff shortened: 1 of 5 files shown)


dirkhh (approved):
looks good. thanks

PR #3234, opened 2021-04-25
Minor code hygiene: profile code

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is just minor code hygiene, see commit messages for details.

On a side note: The positioning of the profile seems broken to me. You can't use relative positioning for the axes! This will lead to either free space or cut-off axes on resizing. Matter of fact, when exporting the planned dive, depths > 100 m/ft are cut off for me. I believe that the whole idea of a fixed (100, 100) size is broken. I am *not* at all looking forward to fix this.
### Release note:

Nope.

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -4,7 +4,6 @@
 #include "core/event.h"
 #include "core/subsurface-string.h"
 #include "core/qthelper.h"
-#include "core/picture.h"
 #include "core/profile.h"
 #include "core/settings/qPrefDisplay.h"
 #include "core/settings/qPrefTechnicalDetails.h"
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
I am 99% that the simple compare initially didn't work. I think there was an offset between the pointer values or something... if this works now, even better.
I like the DiveTextItem cleanup - and of course the header simplification

bstoeger:
There *is* an offset. But the compare still works, because the compiler understands that this is the only way it can upcast (sidecast?) the pointer to be compatible. Unless something completely  different changed, this construct should always have worked (before Qt even existed).

Maybe you were just confused by the `printf` output with different values? For historical interest it might be interesting to check out the original version of the code, but I don't think it's worth the hassle. :-P

bstoeger:
For illustration:
```
printf("%p %p\n", diveComputerText, (QGraphicsItem *)diveComputerText);
```
gives
```
0x55b4069a3d40 0x55b4069a3d50
```

This cast implicitly happens if you compare to `QGraphicsItem *`, so everything turns out fine.

dirkhh:
I will never understand C++

PR #3235, opened 2021-04-28
[RFC] Detach printing/exporting and UI profiles

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is my current state of the detach printing/export and display profile widgets work.

Since I have already spent some time on this, I'd like finish that before fixing the resizing issue of the profile (which will by a royal pain).

This has currently two major and annoying issues:
- Some strange artefactual gas line
- Still have to `show()` the profile used for printing/exporting, or the heat map will not show up!? Thus, this can't currently be reused on mobile.

One good point:
- It removes the profile UI code from `mainwindow.cpp`. Slowly the MainWindow code is becoming reasonably short.

### Release note:

Not yet, purely code hygiene.

### Documentation change:

No.


bstoeger:
No idea why the builds fail. Looks like something build-system related..?

dirkhh:
The build failure looks like a moc dependency issue. I'm a bit confused by this; it appears we aren't creating `ui_mainwindow.h` - which is not something that we do ourselves, but that is implicitly created.
I'm busy at work today (and then scheduled for family stuff in the evening), so I'm not quite sure when I'll be able to dig into this.

bstoeger:
No hurry - this is a long way from being finished.

bstoeger:
Added commit that fixes
> Still have to show() the profile used for printing/exporting, or the heat map will not show up!? Thus, this can't currently be reused on mobile.

`MainWindow` is now gone from the printing/export core code.

bstoeger:
Sigh. Very talented in stepping on my own foot.

@dirkhh: do we have a higher resolution version of `flag.png` (yes, `mapwidget-marker.png` but 32x32 is hardly better than 24x24), perhaps even svg?

The marker on printed pages is just unacceptable, imho:
profile53

dirkhh:
According to our git history this file was introduced in 42980e51d3f916433 and later darkened and renamed - but the original was apparently drawn by @neolit123 in CorelDraw / PhotoShop.
Lubo, do you still have the original file so you could create an SVG from it?
I mean, it wouldn't be crazy hard to recreate that, but if you have it, that would be nice...

neolit123:
i could not find the source files, so i created it again. it's close but not 100% the same.
here is a ZIP with SVG variants + CorelDraw source file.

[mapwidget-marker.zip](https://github.com/subsurface/subsurface/files/6437973/mapwidget-marker.zip)

EDIT: i'm not the author of flag.png. it comes form here:
https://github.com/subsurface/subsurface/commits/71c823441396d357b6f3bb329fa6bd43dfbab7e7/flag.png

bstoeger:
Very nice - thank you. It will take some time until I can incorporate it, because the profile scaling is very confusing.

bstoeger:
Concerning the spurious line at the bottom of the printed charts:

This is generated by `DiveGFLineItem`, which was introduced in 3fc9c1e0053a0ca1d4af7c1e5fb8e118afe0badf.

The thing is, I can't figure out how that line is hidden on desktop!? It cerainly isn't shown for me, even though the code is clearly executed.

bstoeger:
Oh - I get it now. When I call `profile->setEmptyState();` before printing, the GFLine is hidden.

There appears to be no code to unhide this line. Could it be that it is never actually shown? Can it be removed?

bstoeger:
Hmmm... locally `TestProfile` fails, but CI here completes. Strange.

bstoeger:
From my side, this is now ready for review / testing.

FYI: The next thing I'd like to try before trying to fix the resize code is to remove the `QGraphicsView` layer when printing / exporting. We have `QGraphicsScene` -> `QGraphicsView` -> export/prining. However, `QGraphicsView` is a `QWidget` and I see no reason why we would need an UI element to render to an image..? Just seems to make things more complicated with all that transformation rigmarole.

bstoeger:
BTW: TeX export can now be compiled on mobile with this trivial patch:
```
diff --git a/backend-shared/exportfuncs.cpp b/backend-shared/exportfuncs.cpp
index 35da4775a..79f980f6b 100644
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -30,8 +30,6 @@ bool ExportCallback::canceled() const
 	return false;
 }
 
-#if !defined(SUBSURFACE_MOBILE)
-
 // Let's say that 800x600 is a "reasonable" profile size. Use four times that for printing.
 static const constexpr int profileScale = 4;
 static const constexpr int profileWidth = 800 * profileScale;
@@ -321,7 +319,6 @@ void export_depths(const char *filename, bool selected_only)
 	}
 	free_buffer(&buf);
 }
-#endif /* ! SUBSURFACE_MOBILE */
 
 std::vector<const dive_site *> getDiveSitesToExport(bool selectedOnly)
 {
```
I am not sure whether this is still wanted though.

dirkhh:
I don't understand why someone would WANT TeX export on mobile... 
So my inclination would be not to expose this? But then, on the other hand... people are doing very weird stuff and if it is such an easy fix... yeah, I'm not sure what to tell you.

bstoeger:
> I'm not sure what to tell you

Good - now there's two of us. It obviously wasn't my idea to move this to `backend-shared`. I reckon we wait until someone volunteers to connect this to the mobile-ui...?

PS: I wonder why this is in `backend-shared` at all. Since it isn't UI related, the correct place seems to be `core`?
PPS: I wonder whether we should rename `backend-shared` to `ui-shared` as the way I understand this was supposed to be UI functions shared between desktop and mobile?

dirkhh:
These are all Jan-isms, IIRC.
I'm not sure I'm up for a rename orgy. But yeah, I think the idea of leaving sleeping dragons alone makes sense. No point in enabling this for mobile.

dirkhh (approved):
I looked through the code and it seems reasonable.
Also did a quick test and didn't notice anything broken (but not really exhaustive, TBH).

@atdotde - have you looked at this as well?

willemferguson:
Berthold,
When I place a bookmark along a dive profile, there are two problems:
1) The bookmark cannot be deleted.
2) I cannot edit the text (name) of the bookmark.
Does this have anything to do with your current initiative, or should I file a bug?
Kind regards, willem

bstoeger:
I'll look into it.

bstoeger:
I think I see the problem:
`DiveEventItem` uses an internal copy of the event and returns that in `getEvent()`. Trying to remove that from the dive obviously fails and even may lead to memory corruption. I'll have to think about this.

bstoeger:
Ha! We don't need any of that `QTransform` rigmarole if we directly print via the scene and bypass the view: 09a9a24fae1ec8714ad785dbad65956709ac4198

This makes things much easier to reason about. And it reinforces me in getting rid of the QGraphicsView for mobile/printing.

dirkhh:
btw, what's the state of this?
Or, asked differently - should we just call for a lot more testing of what's in master now to see if that happens to fix the "size of the gas change icon"? Or should we wait for this to have taken more shape?

bstoeger:
Currently, there is no point in this - let's postpone after release and try to fix the worst issues in current master.

bstoeger:
Superseded by #3300. I'm not forgetting about Lubomir's SVG flag. It will be used when redoing the scaling code.

PR #3238, opened 2021-04-30
Planner: Don't force ascent

atdotde:
The planner should not always ascent from the depth of
the last manually entered waypoint but consider the
possibility that it should first top where we are
before the next stop depth has cleared.

Reported-by: David Carron

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/planner.c
+++ b/core/planner.c
@@ -757,7 +757,7 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 
 	/* Find the first potential decostopdepth above current depth */
 	for (stopidx = 0; stopidx < decostoplevelcount; stopidx++)
-		if (*(decostoplevels + stopidx) >= depth)
+		if (decostoplevels[stopidx] > depth)
 			break;
 	if (stopidx > 0)
 		stopidx--;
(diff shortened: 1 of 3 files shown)


core/planner.c
```diff
@@ -757,7 +757,7 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 
 	/* Find the first potential decostopdepth above current depth */
 	for (stopidx = 0; stopidx < decostoplevelcount; stopidx++)
-		if (*(decostoplevels + stopidx) >= depth)
```
bstoeger, line 760: I can't comment on the patch itself. Just a coding-style comment:
Could we change the condition to `decostoplevels[stopidx] > depth`?
As it stands, that is super-cryptic.

atdotde: You are right. That code is ancient and was written before my time...

dirkhh:
The commit contains a libdivecomputer change.
NAK

atdotde:
Argh, this happens when I rebase...

dirkhh (approved):
thanks for fixing it

PR #3239, opened 2021-05-06
[TRIVIAL] cleanup: free print dialog in planner

bstoeger:
When printing the plan, a print-dialog was created with "new",
but not freed later. Strictly speaking, this is not a leak,
because the dialog is attached to the main-window in Qt's
object hierarchy. Thus it is freed on application exit. On
the other hand, it is a leak in the sense that resources are
pointlessly hogged until application exit.

Let's just turn it into a stack-allocated object.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A (seemingly?) trivial code cleanup: Automatically free the print-dialog after use by making it a stack-allocated object.

### Release note:

No - trivial change.

### DIFF
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -588,9 +588,9 @@ void PlannerWidgets::printDecoPlan()
 	free(disclaimer);
 
 	QPrinter printer;
-	QPrintDialog *dialog = new QPrintDialog(&printer, MainWindow::instance());
-	dialog->setWindowTitle(tr("Print runtime table"));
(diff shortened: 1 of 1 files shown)


dirkhh:
Sweet, now we have three tests failing... I need to get onto that, I guess.

dirkhh (approved):
yeah, that makes sense

dirkhh:
AppImage issue is addressed in master, planner isn't built on iOS (the old failure was a download timing out)... I'm running out of "time for Subsurface", so I'll merge this before the iOS build completes.

PR #3248, opened 2021-05-16
Adding the capability to copy / paste dive number and date / time between dives.

mikeller:
This is adding the capability to select 'Dive number' and 'Date / Time'
in the 'Copy dive components' dialog, and then copy them into the
clipboard.

image

When using 'Paste dive components, these values will then be pasted into
the selected dive(s).
This is intended to help with workflows that import dive information
from two different sources, like general information from another
logging program, and CCR ppO2 sensor readings from a unit log, and then
stitch them together into one cohesive entry with all data per dive.
Copied data is also output into formatted text when pasting the
clipboard outside of the application:

```
Dive number: 401
Date / time: Sun 2 May 2021 12:00 AM
```

No translations have been added as of now - I could not find any
information on how strings are translated for this project.

Reported-by: Michael Keller <mikeller@gmail.com>
Signed-off-by: Michael Keller <mikeller@gmail.com>

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -685,6 +685,10 @@ PasteState::PasteState(dive *dIn, const dive *data, dive_components what) : d(dI
 	}
 	if (what.weights)
 		copy_weights(&data->weightsystems, &weightsystems);
+	if (what.number)
+		number = data->number;
+	if (what.when)
+		when = data->when;
 }
 
 PasteState::~PasteState()
@@ -725,6 +729,10 @@ void PasteState::swap(dive_components what)
 		std::swap(cylinders, d->cylinders);
 	if (what.weights)
 		std::swap(weightsystems, d->weightsystems);
+	if (what.number)
+		std::swap(number, d->number);
+	if (what.when)
+		std::swap(when, d->when);
 }
 
 // ***** Paste *****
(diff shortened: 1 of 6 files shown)


atdotde:
Looks good. And you have added translations: By using the tr() function, you have marked the strings for translation.

mikeller:
@atdotde: That was easy for the translations then - I wasn't sure if the translated strings would have to be added as part of the pull request, or if an external translation platform is being used - maybe something worthwhile mentioning somewhere in `CONTRIBUTING.md`.
What do I have to do to make the 'DCO' test pass? There is a `Signed-off-by`line in my commit message.

mturkia:
CONTRIBUTING.md has the following mentioned:
Our translations are centrally handled at [Transifex][3] \-- please sign up for an account there and then request to join the [Subsurface Team][4].

mikeller:
@mturkia: Apologies, I must have overlooked this when reading the document.

dirkhh:
We usually don't have a "Reported-by:" line if that's the author... But I'm a bit confused about why the DCO test fails...

The commit title is a bit awkward, too long, and doesn't follow our pattern. But that's not causing that test failure.
May I suggest something like
desktop: add dive number and date to copy paste data

It may be that GitHub is trying to be smart here. The Author of the commit is `mikeller <github@ike.ch>` - which is different from the SOB. So maybe when you update the commit title you could do this with
```
git commit --amend --author "Michael Keller <mikeller@gmail.com>"
```
and see if that makes the DCO test happy?

PR #3252, opened 2021-05-17
Make the Mac SDK detection in build.sh more robust

ryangardner:
Update how the SDK is found when building on OS X to make it easier to build on the latest OS X version
with the default SDK versions provided by the command-line tools or XCode 12.5 (while it should still be
compatible with the existing build vms)

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [X] Build system change
- [ ] Documentation change
- [ ] Language translation

When trying to build on Big Sur, the xcode command-line tools
install are installed in /Library/Developer/CommandLineTools/SDKs,
and as of Xcode 12.5, it does not include a 10.x version
of SDK.

This changes it to search in the location of the command-line tools SDK
for a 10.x version, and if it can't find a 10.x version it will
find an explicit 11.x version of the SDK to use because it is
conceivable that in the near future Apple will stop installing any
10.x SDK's as part of the command-line tool installer.

If the SDK can't be found, the build script will exit now instead
of continuing with an unset BASESDK version that causes a later failure.
### Changes made:
* Search for a developer SDK in the default place that the xcode command-line-tools installs them 
* If a 10.x SDK is not found, try using a numbered 11.x SDK
* Set the `-mmacosx-version-min=` to the value specified by the `BASESDK`
* Exit with an error if an appropriate SDK can't be found

### Documentation change:
There are only developer-facing changes here that make the build process match existing documentation, not
require any new changes.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -159,13 +159,13 @@ if [ "$PLATFORM" = Darwin ] ; then
 	# find a 10.x base SDK to use, or if none can be found, find a numbered 11.x base SDK to use
 	BASESDK=$(ls $SDKROOT | grep "MacOSX10\.1.\.sdk" | head -1 | sed -e "s/MacOSX//;s/\.sdk//")
 	if [ -z "$BASESDK" ] ; then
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
Thanks - we could have combined those two commits, but this works.

PR #3257, opened 2021-05-30
Add numerus translation for dive context menu

### LINKED ISSUE #3256: Numerus form not applied for trip summary in en-GB
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
The summary line for dive trip shows "(# dive(s))" instead of using the numerus form from translation file (ie: "1 dive" vs "2 dives")

### Operating system:
Linux and Windows

### Subsurface version:
v5.0.1

### Steps to reproduce:
1) Select two or more dives from the dive list
2) Right-click, select "Create new trip above"

### Current behavior:
The trip summary line is shown as "<title>, <date> (x dive(s))"

### Expected behavior:
Trip summary line should show "<title>, <date> (x dives)" if multiple dives are in the trip
or 
"<title>, <date> (1 dive)" if only one dive is in the trip

### Mentions:

trip_numerus

schtibal:
Add numerus translation lookup for the right-click context menu in the dive list to show proper singular/plural text

Committer: Mark Stiebel <mark@stiebel.me>

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [X] Language translation

Updates to right-click context menu in the dive to enable numerus translations

### Changes made:

1) Remove dive(s) from trip
2) Add dive(s) ...
3) Renumber dive(s)
4) Also disable "Collapse Others" if the current dive is not in a trip.

### Related issues:

Fixes #3256

### Release note:
UI fixes to allow singular/plurals forms in dive list context menu

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -813,7 +813,7 @@ void DiveListView::contextMenuEvent(QContextMenuEvent *event)
 		}
 	}
 	if (d) {
-		popup.addAction(tr("Delete %n dive(s)","",amount_selected), this, &DiveListView::deleteDive);
+		popup.addAction(tr("Delete dive(s)","",amount_selected), this, &DiveListView::deleteDive);
 		if (d->invalid)
 			popup.addAction(tr("Mark dive(s) valid","",amount_selected), this, &DiveListView::markDiveValid);
 		else
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -779,12 +779,14 @@ void DiveListView::contextMenuEvent(QContextMenuEvent *event)
 			popup.addAction(tr("Collapse all"), this, &QTreeView::collapseAll);
 
 		// verify if there`s a need for collapse others
-		if (expanded_nodes > 1)
+		if (expanded_nodes > 1 && d->divetrip)
 			collapseAction = popup.addAction(tr("Collapse others"), this, &QTreeView::collapseAll);
 
 
 		if (d) {
-			popup.addAction(tr("Remove dive(s) from trip"), this, &DiveListView::removeFromTrip);
+			if (d->divetrip) {
+				popup.addAction(tr("Remove dive(s) from trip","",amount_selected), this, &DiveListView::removeFromTrip);
+			}
 			popup.addAction(tr("Create new trip above"), this, &DiveListView::newTripAbove);
 			if (!d->divetrip) {
 				struct dive *top = d;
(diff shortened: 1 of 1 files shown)


desktop-widgets/divelistview.cpp
```diff
@@ -811,17 +813,17 @@ void DiveListView::contextMenuEvent(QContextMenuEvent *event)
 		}
 	}
 	if (d) {
-		popup.addAction(tr("Delete dive(s)"), this, &DiveListView::deleteDive);
+		popup.addAction(tr("Delete %n dive(s)","",amount_selected), this, &DiveListView::deleteDive);
```
dirkhh: I'm curious how you decided where to add the `%n` and where not to add it.
This obviously could have been mentioned in the menu entries that add dive(s) to a trip, right?
Not saying that this is wrong, just trying to understand why here not not in the other cases.

dirkhh (commented):
Overall I like it, but I have a question...
(and sorry for the delay, it was a long weekend in the US and I decided to take a break from being at my computer all the time)

desktop-widgets/divelistview.cpp
schtibal: Good pickup.
Pretty sure that was just missed as I was getting my head around the translation system. The %n parameter isn't required in the source string as passed into tr(), as its added in the translation lookup. TBH, it seems strange to me that adding the %n to the source string even worked at all - or maybe my assumptions about how the translation magic works are wrong.

I've added another commit to bring the "Delete dive(s)" back in line with the rest.

I'll add here that my intent is to include the `amount_selected` value as a parameter, but not actually use the number itself in the translations (just use it to determine the singular/plural). 

It feels a bit weird from a UI perspective to fill the context menu with "Delete 4 dives", "Add 4 dives to trip", etc.
Regardless, the choice of whether to include the number or not can lie within the translation, not within the code.

dirkhh: You are correct, a translator who KNOWS that this is the case could add the %n - but I'd find that weird, TBH, and of course it isn't discoverable to the translator unless they are very familiar with the code as well.

dirkhh:
I think I'll merge this locally to avoid the back and forth - once the checks complete.
GitHub's new rule not to run these for first time contributors makes perfect sense - but also throws me since I look for completed tests before responding to PRs...

dirkhh:
merged - with small edits to make the commit message work for our format :)

PR #3262, opened 2021-06-05
Fix removing/renaming events on the profile

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Moving/renaming events was broken, because an internal copy of the event was passed to the undo-code. It is mysterious how this ever worked, but I'm quite sure that at one point it did...!?

Replace the internal copy with a pointer. By now (removal of displayed_dive, etc.), this should be safe.

### Changes made:

1) Pass event, model, axes at construction time of DiveEventItem, to avoid having to deal with partially constructed objects.
2) Remove internalEvent.

### Related issues:

All the profile changes.

### Release note:

Done.

@willemferguson: please check / comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: fix removal/editing of bookmarks on the profile
 export: show progress dialog when exporting to TeX
 printing: use sensible font size even for strange window size
 * Planner: Don't immediately ascent automatically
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
This seems right.
How did this ever work?

PR #3266, opened 2021-06-15
cleanup: factor out duplicate axis-initialization code

bstoeger:
The axes of the profile are setup when switching into
the "ProfileState" and also when the preferences are
changed. The same code existed twice for both cases.

Let's factor it out into a single function to avoid
future divergence and confusion.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a rather trivial cleanup of the profile code, but I'd prefer a second set of eyes.

### Changes made:

1) Factor out common code.

### Related issues:

#3235

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -794,7 +794,7 @@ void ProfileWidget2::actionRequestedReplot(bool)
 	settingsChanged();
 }
 
-void ProfileWidget2::settingsChanged()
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
I like it. @atdotde  ??

bstoeger:
Ok, since this is fairly simple, let's commit and see if someone complains.

PR #3272, opened 2021-07-04
profile: scale dive event items according to font print scale

bstoeger:
When printing with low DPI, the dive event items become comically
large, because they are not resized like the fonts. Therefore,
scale using the fontPrintScale.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

There are numerous reports of wrongly scaled event items. I cannot really reproduce them, with the one exception being very small DPI values. This quick-fix attempts to scale the event items according to the "fontPrintScale".

But I fear that there is something wrong with the size calculation. Selecting 100 DPI make the printout unreadable. I will try to dig deeper. If 100 DPI is really *that* bad, we should think about limiting the lower value to perhaps 300 DPI?

### Changes made:

1) Scale dive event items according to fontPrintScale.

### Release note:

First, we need confirmation that this really fixes something.

### DIFF
--- a/profile-widget/diveeventitem.cpp
+++ b/profile-widget/diveeventitem.cpp
@@ -16,7 +16,7 @@
 
 DiveEventItem::DiveEventItem(const struct dive *d, struct event *ev, struct gasmix lastgasmix,
 			     DivePlotDataModel *model, DiveCartesianAxis *hAxis, DiveCartesianAxis *vAxis,
-			     int speed, QGraphicsItem *parent) : DivePixmapItem(parent),
+			     int speed, double fontPrintScale, QGraphicsItem *parent) : DivePixmapItem(parent),
 	vAxis(vAxis),
 	hAxis(hAxis),
 	dataModel(model),
@@ -25,7 +25,7 @@ DiveEventItem::DiveEventItem(const struct dive *d, struct event *ev, struct gasm
 {
 	setFlag(ItemIgnoresTransformations);
 
-	setupPixmap(lastgasmix);
+	setupPixmap(lastgasmix, fontPrintScale);
 	setupToolTipString(lastgasmix);
 	recalculatePos(0);
 
@@ -48,7 +48,7 @@ struct event *DiveEventItem::getEventMutable()
 	return ev;
 }
 
-void DiveEventItem::setupPixmap(struct gasmix lastgasmix)
+void DiveEventItem::setupPixmap(struct gasmix lastgasmix, double fontPrintScale)
 {
 	const IconMetrics& metrics = defaultIconMetrics();
 #ifndef SUBSURFACE_MOBILE
@@ -63,6 +63,7 @@ void DiveEventItem::setupPixmap(struct gasmix lastgasmix)
 	int sz_bigger = metrics.sz_big + metrics.sz_med;
 #endif
 #endif
+	sz_bigger = lrint(sz_bigger * fontPrintScale);
 	int sz_pix = sz_bigger/2; // ex 20px
 
 #define EVENT_PIXMAP(PIX) QPixmap(QString(PIX)).scaled(sz_pix, sz_pix, Qt::KeepAspectRatio, Qt::SmoothTransformation)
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
there's no way this is really this easy...
I distinctly remember fighting with this in the past.
This is embarrassing.
Thanks, Berthold

bstoeger:
> there's no way this is really this easy...

Indeed, it probably isn't. That was just a shot in the dark, since I can't reliably reproduce the problem. :(

dirkhh:
This took me way too long, my apologies. Here's a test with this code on a Mac:
Screen Shot 2021-07-06 at 2 58 30 PM
Not quite as comically bad - but still not great. I think we should add some debugging code to understand why we get these unrealistic sizes in the first place...

dirkhh:
fascinatingly, if I remove the scaling by a factor of 4 introduced in 52aa7d83b6 the result looks... reasonable on this Mac
Screen Shot 2021-07-06 at 3 28 20 PM

Thoughts, @atdotde ?

dirkhh:
OMG... https://github.com/subsurface/subsurface/pull/2905
Screen Shot 2021-07-06 at 3 38 14 PM
I hate it when I almost get things right... but then don't follow up.
So, while this might be the right thing for **export**, I think it's the wrong thing to do for **printing**

Which brings us back to the earlier conversation with Berthold. The whole profile code is cesspool of ugly hacks. And the scaling of various elements is just one of them.

Right now I'm thinking... Robert - should we find a way to make this only work on exporting profiles (vs printing)? Or am I missing something that results in what you did in #2905 actually being the right thing to do?

Edited to add: I played with both the printing and the export of profile PNGs. Both on Linux and Mac. And in none of the cases that I tried (including different DPIs for printing) do I think that this 4* scale is the right thing to do. So I must be missing something.

bstoeger:
Perhaps that was just a workaround for the fact that the `fontPrintScale` was not factored in the scaling of the event icons, which is not necessary anymore. I would opt for removing it.

bstoeger:
PS: I wonder if we should set the lower DPI limit to 300 as such:

```
From 88778432026054552a063b917ee3b678d525355a Mon Sep 17 00:00:00 2001
From: Berthold Stoeger <bstoeger@mail.tuwien.ac.at>
Date: Wed, 7 Jul 2021 10:03:29 +0200
Subject: [PATCH] printing: set minimum DPI to 300

Below 300 DPI, the printouts of the profile become very blurry.
Let's set the minimum to this value.

---
 desktop-widgets/printoptions.ui | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/desktop-widgets/printoptions.ui b/desktop-widgets/printoptions.ui
index f12236552..ff1b08dbc 100644
--- a/desktop-widgets/printoptions.ui
+++ b/desktop-widgets/printoptions.ui
@@ -103,6 +103,9 @@
             <verstretch>0</verstretch>
            </sizepolicy>
           </property>
+          <property name="minimum">
+           <number>300</number>
+          </property>
           <property name="maximum">
            <number>4000</number>
           </property>
-- 
2.30.2
```

Nobody is using the print function to print posters, billboards or the like. And even if they do - they can reduce DPI in post-processing.

atdotde:
This is how I spent last year's summer holiday...

And yes, it was a hack that made things look right for me on my Mac. Let me quote from the original commit message

''Maybe somebody
with better Qt knowledge can take off from here. In my opinion
this is already a step in the right direction.''

I am happy to spend some more time on this to figure out how to do this properly, hopefully later today.

dirkhh:
This all comes down to one of the main challenges that we have. Too few active developers (umm... three?), too many platforms, too many features that few people ever use... it seems to be getting harder and harder to just keep things working.

I think that's a big part why I am so negative on almost all new feature requests. Because in the end I ask myself "how many things will this break". So... why do we want to be able to export profile PNGs. Sure, this is neat - but really, how many people actively use this? Who tests this for every release, on every platform?
Clearly we don't even test printing on all the platforms. :(

bstoeger:
ATTN: I think I found a bug in current master - the ceiling thing can't be turned on/off in the planner. I will investigate, but have only little time, since I have a guest for the whole upcoming week.

bstoeger:
Sorry - never mind. I'm just too stupid to understand how the icons work. :-P

PR #3279, opened 2021-07-14
cleanup: avoid potential crash accessing invalid cylinder

dirkhh:
This shouldn't happen if the plot_info and dive are in sync, yet we have
seen reports where a user could reproduce a crash in this call. The
added check shouldn't hurt and the warning might help us figure out
what's going wrong.

- [x] Bug fix
- [x] Code cleanup

This does seem strange, but... I also think that it wouldn't hurt. And a crash is a sucky experience for a user.

### Changes made:

Simply check the cylinder pointer we get back before dereferencing it.

### Related issues:

#3278 

### Release note:

not sure that's worth an entry

### Documentation change:

no

### Mentions:

@bstoeger @atdotde - am I going overboard here? Fixing the wrong thing?

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -1424,9 +1424,16 @@ static void plot_string(const struct dive *d, const struct plot_info *pi, int id
 		int mbar = get_plot_pressure(pi, idx, cyl);
 		if (!mbar)
 			continue;
-		struct gasmix mix = get_cylinder(d, cyl)->gasmix;
-		pressurevalue = get_pressure_units(mbar, &pressure_unit);
-		put_format_loc(b, translate("gettextFromC", "P: %d%s (%s)\n"), pressurevalue, pressure_unit, gasname(mix));
+		// one would assume that there always is such a cylinder, but users have seen
+		// crashes when this got dereferenced, so let's check there's actually a cylinder there
+		cylinder_t *cylinder = get_cylinder(d, cyl);
+		if (cylinder) {
+			struct gasmix mix = cylinder->gasmix;
+			pressurevalue = get_pressure_units(mbar, &pressure_unit);
+			put_format_loc(b, translate("gettextFromC", "P: %d%s (%s)\n"), pressurevalue, pressure_unit, gasname(mix));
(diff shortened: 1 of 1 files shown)


bstoeger:
Erm, according to the discussion in #3278, this is already fixed?

dirkhh:
Yes. What is fixed is the reason why `plot_info` and the dive appeared to be out of sync.
But the problem is that a similar bug elsewhere could once again cause a crash. So this is trying to avoid such a crash.

Or maybe that's a dumb thing to do and we should have things crash for our users to indicate a bug?

bstoeger:
Ok, but with that argument one could place an if around every single pointer access, no?

A crash might be inconvenient for the user, but at least we get a traceback.

In this particular case, calling `plot_string()` with inconsistent dive and plot_info is clearly invalid and I don't think that we should actively protect against it, but fix the caller(s).

If you really want to protect against this, I would suggest somewhere at the beginning of the function:
```
        if (pi->nr_cylinders != d->cylinders.nr) {
                put_string(b, translate("gettextFromC", "Something went wrong, please send report.\n"));
                return;
        }
```
Because under these circumstances the whole function makes no sense. And at least there is a chance of a user report. But even like this, I don't think it's a particularly good idea.

dirkhh:
Ok, makes sense

bstoeger:
After looking at 6b2e56e5131aa707216c4b0da5805199fcc0034d, I wonder what the root cause of this is.

Aren't *all* dives supposed to have samples? Aren't we supposed to add a "fake dc" in `fixup_dive` when loading the log or after importing? Something is odd and not behaving as it should.

Could you provide me with an example to reproduce the problem?

dirkhh:
I don't have one - that was only sent to @atdotde 
I simply worked my way back from offset in the stack trace to the line of code that caused the crash.

atdotde:
I ran into the problem by importing dives from csv (after about ten years with subsurface this pandemic made me sit down and type in all my old dives from paper log books). Those don't have samples.

PR #3280, opened 2021-07-15
core: always create a fake profile if there are no samples

bstoeger:
The profile-view can't handle dives without samples. This was
recently fixed in 6b2e56e5131, by showing the "empty state"
of the profile.

This is inconsistent, because in other parts of the code
(notably per_cylinder_mean_depth()) a fake profile is
created implicitly. The behaviour depends on whether the
user was on the equipment tab or not! That's quite obscure.

Be more consistent, by generating a fake profile with fake_dc()
for every dive in fixup_dc().

This will make a few fake_dc() calls obsolete, but so be it.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A fake profile was created under some circumstances, but not others. This ultimately led to crashes, fixed in 6b2e56e5131. Let's try to create a fake profile in every case.

I'm not sure if this is the right thing to do for multi-DC dives. Maybe in that case we only want to add a profile for the first dc and refuse to show the profile for DCs with no samples?

### Changes made:

1) Create fake profile in `fixup_dive()`.

### Mentions:

@atdotde: could you please check if that fixes your CSV import? Perhaps the problem is that there are crazy parameters and `fake_dc()` does not create a profile?

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1246,6 +1246,10 @@ static void fixup_dive_dc(struct dive *dive, struct divecomputer *dc)
 
 	/* Fixup CCR / PSCR dives with o2sensor values, but without no_o2sensors */
 	fixup_no_o2sensors(dc);
+
+	/* If there are no samples, generate a fake profile based on depth and time */
+	if (!dc->samples)
+		fake_dc(dc);
 }
 
 struct dive *fixup_dive(struct dive *dive)
--- a/core/dive.c
+++ b/core/dive.c
@@ -2614,7 +2614,6 @@ struct dive *merge_dives(const struct dive *a, const struct dive *b, int offset,
 	MERGE_MAX(res, a, b, rating);
 	MERGE_TXT(res, a, b, suit, ", ");
 	MERGE_MAX(res, a, b, number);
-	MERGE_NONZERO(res, a, b, cns);
 	MERGE_NONZERO(res, a, b, visibility);
 	copy_pictures(a->pictures.nr ? &a->pictures : &b->pictures, &res->pictures);
 	taglist_merge(&res->tag_list, a->tag_list, b->tag_list);
@@ -2630,6 +2629,9 @@ struct dive *merge_dives(const struct dive *a, const struct dive *b, int offset,
 		interleave_dive_computers(res, &res->dc, &a->dc, &b->dc, cylinders_map_a, cylinders_map_b, offset);
 	else
 		join_dive_computers(res, &res->dc, &a->dc, &b->dc, cylinders_map_a, cylinders_map_b, 0);
+
+	/* The CNS values will be recalculated from the sample in fixup_dive() */
+	res->cns = res->maxcns = 0;
 
 	/* we take the first dive site, unless it's empty */
 	*site = a->dive_site && !dive_site_is_empty(a->dive_site) ? a->dive_site : b->dive_site;
--- a/core/import-csv.c
+++ b/core/import-csv.c
@@ -245,6 +245,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
 			return -1;
 		}
 		mem_csv.size = ptr - (char*)mem_csv.buffer;
+		end_ptr += ptr - (char *)mem_csv.buffer;
 
 		iter = parse_dan_new_line(ptr + 1, NL);
 		if (iter && strncmp(iter, "ZDT", 3) == 0) {
@@ -268,7 +269,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
 			return -1;
 
 		ret |= parse_xml_buffer(filename, mem_csv.buffer, mem_csv.size, table, trips, sites, devices, filter_presets, params);
-		end_ptr += ptr - (char *)mem_csv.buffer;
+
 		free(mem_csv.buffer);
 	}
 
--- a/dives/test40-42.xml
+++ b/dives/test40-42.xml
@@ -6,32 +6,56 @@
 </site>
 </divesites>
 <dives>
-<dive number='1' date='2013-10-01' time='10:34:00' duration='45:00 min'>
+<dive number='1' otu='9' cns='5%' date='2013-10-01' time='10:34:00' duration='45:00 min'>
   <buddy>Dirk</buddy>
   <suit>wet, 5mm</suit>
-  <divecomputer model='csv' deviceid='ffffffff'>
+  <divecomputer model='csv' last-manual-time='45:00 min' deviceid='ffffffff'>
   <depth max='18.0 m' mean='16.0 m' />
+  <sample time='0:00 min' depth='0.0 m' />
+  <sample time='2:00 min' depth='18.0 m' />
+  <sample time='38:31 min' depth='18.0 m' />
+  <sample time='39:51 min' depth='5.94 m' />
+  <sample time='44:20 min' depth='5.94 m' />
+  <sample time='45:00 min' depth='0.0 m' />
   </divecomputer>
 </dive>
-<dive number='2' date='2013-10-01' time='12:13:00' duration='41:00 min'>
+<dive number='2' otu='5' cns='7%' date='2013-10-01' time='12:13:00' duration='41:00 min'>
   <buddy>Linus</buddy>
   <suit>wet, shorty, 3mm</suit>
-  <divecomputer model='csv' deviceid='ffffffff'>
+  <divecomputer model='csv' last-manual-time='41:00 min' deviceid='ffffffff'>
   <depth max='16.2 m' mean='13.5 m' />
+  <sample time='0:00 min' depth='0.0 m' />
+  <sample time='1:48 min' depth='16.2 m' />
+  <sample time='31:41 min' depth='16.2 m' />
+  <sample time='32:53 min' depth='5.346 m' />
+  <sample time='40:24 min' depth='5.346 m' />
+  <sample time='41:00 min' depth='0.0 m' />
   </divecomputer>
 </dive>
 <dive number='3' date='2014-10-01' time='10:02:00' duration='48:00 min'>
   <buddy>Tomaz</buddy>
   <suit>none</suit>
-  <divecomputer model='csv' deviceid='ffffffff'>
+  <divecomputer model='csv' last-manual-time='48:00 min' deviceid='ffffffff'>
   <depth max='13.3 m' mean='11.5 m' />
+  <sample time='0:00 min' depth='0.0 m' />
+  <sample time='1:29 min' depth='13.3 m' />
+  <sample time='39:02 min' depth='13.3 m' />
+  <sample time='40:02 min' depth='4.389 m' />
+  <sample time='47:31 min' depth='4.389 m' />
+  <sample time='48:00 min' depth='0.0 m' />
   </divecomputer>
 </dive>
-<dive number='4' date='2014-10-01' time='14:19:00' duration='34:00 min'>
+<dive number='4' otu='13' cns='5%' date='2014-10-01' time='14:19:00' duration='34:00 min'>
   <buddy>Don</buddy>
   <suit>dry, Whites Fusion</suit>
-  <divecomputer model='csv' deviceid='ffffffff'>
+  <divecomputer model='csv' last-manual-time='34:00 min' deviceid='ffffffff'>
   <depth max='24.9 m' mean='20.1 m' />
+  <sample time='0:00 min' depth='0.0 m' />
+  <sample time='2:46 min' depth='24.9 m' />
+  <sample time='25:35 min' depth='24.9 m' />
+  <sample time='27:26 min' depth='8.217 m' />
+  <sample time='33:05 min' depth='8.217 m' />
+  <sample time='34:00 min' depth='0.0 m' />
   </divecomputer>
 </dive>
 <dive number='417' otu='138' cns='54%' divesiteid='deadbeef' date='2015-05-23' time='13:23:00' duration='100:00 min'>
--- a/tests/testmerge.cpp
+++ b/tests/testmerge.cpp
@@ -12,7 +12,7 @@ void TestMerge::initTestCase()
 {
 	/* we need to manually tell that the resource exists, because we are using it as library. */
 	Q_INIT_RESOURCE(subsurface);
-	prefs.cloud_base_url = strdup(default_prefs.cloud_base_url);
+	copy_prefs(&default_prefs, &prefs);
 }
 
 void TestMerge::cleanup()
--- a/tests/testparse.cpp
+++ b/tests/testparse.cpp
@@ -35,7 +35,7 @@ void TestParse::initTestCase()
 {
 	/* we need to manually tell that the resource exists, because we are using it as library. */
 	Q_INIT_RESOURCE(subsurface);
-	prefs.cloud_base_url = strdup(default_prefs.cloud_base_url);
+	copy_prefs(&default_prefs, &prefs);
 }
 
 void TestParse::init()


bstoeger:
Ugh. This makes the merge-test fail, which makes sense. Will have to investigate if that is benign or a real error.

bstoeger:
Hmmm...

The problem is that the test merges two dives (47 and 48), one with and one without samples. In the new version, the dive without samples gets a fake dc on loading. For some reason this changes the cns value in the merged dive from 2% to 4%. I guess(!) this is because the fake profile has high cns (most time spent at max depth), and the higher cns is taken on merging. It appears that the samples are merged correctly, though.

I'm lost - help?

atdotde:
To compute the CNS values, you need a profile. I don't really know what we do/did lacking samples. But what is the correct™ thing to do in this case? Don't compute a value at all or make up a profile and compute it based on that? I tend to think the former is better.

bstoeger:
That means we would need a "fake_dc" flag that prevents calculation of CNS or at least give the CNS value a lower priority when merging.

atdotde:
My tendency is not to compute it at all (setting it to 0). It would then always lose the comparison.

atdotde:
OTOH, computing it with all the dive time spent at max depth would give a conservative estimate...

dirkhh:
if I have a 60 minute dive to 30m and you tell me my CNS is 0 afterwards, I'd have questions.
The fact that that's the case today when importing from CVS points to a bug in the current code, IMHO
So likely the test needs to be fixed

bstoeger:
Hmmm....

There is this comment:
```
/* this only gets called if dive->maxcns == 0 which means we know that
 * none of the divecomputers has tracked any CNS for us
 * so we calculated it "by hand" */
```

Which means that there is already a "calculated CNS" flag. Perhaps we can use this for merging...

Edit: Sorry - mixed comments.

bstoeger:
...and the first thing in `fixup_dive()`:

```
        dive->maxcns = dive->cns;
```
:(

bstoeger:
Please check the new commit. It fixes the test, but I am not at all sure if this is the right thing to do!

bstoeger:
OMG. The parse-test also fails.

dirkhh:
Tests are clearly still failing...
From a larger "program logic" perspective, I worry about simply resetting the CNS, but I haven't thought that through.
If we import from different dive software (or from CVS for that matter) - do we import CNS values? Should we keep them instead of just resetting them if that imported dive is merged with another dive computer? But then, the old logic didn't exactly do that either.
Strangely, CNS is a dive level value, not a dive computer value, yet we have a field for it in the samples.

UGH. Linus - any recollection what and why is happening here?

bstoeger:
I updated the second commit-message with a new interpretation:
```
Before making the cylinder-table dynamic, dives always
had at least one cylinger. When such a dive is displayed,
the TabDiveInformation class calls per_cylinder_mean_depth().
If there are no samples, this function generates a "fake
profile" with fake_dc(). Thus, effectively dives always
had samples once the user was displaying them.

When the cylinder-table was made dynamic, dives without
cylinders were supported. This can notably happen, when
importing from CSV (this could actually be a bug).
per_cylinder_mean_depth() exits early in that case and
doesn't create a fake profile. This lead to crashes
of the profile-widget, which were fixed in 6b2e56e.
Non-sample dives were now shown with the Subsurface-logo.

To restore the previous behavior, genarate a fake profile
for sample-less dives in fixup_dive(), which is called
anytime  a dive is loaded or imported. This seems to
have been the intention anyway and this worked only
"by chance". This will make a few fake_dc() calls obsolete,
but so be it.
```
I think this is the correct thing to do - it appears to have been the intention to always show a profile, even in the case of no dive-computer samples.

Whether the first commit is the right thing to do - I don't know.

I believe that the "no CNS" case never really worked. After looking at a dive, a fake profile was generated. When saving and reloading an CNS value was generated.

dirkhh:
So why is TestParse still failing on the Mac?

bstoeger:
I would guess a hickup? I don't have a Mac, so hard to say.

bstoeger:
Well, that wasn't it. My next guess would be a CR/LF issue.

torvalds (commented):
I don't hate this, and I think it's perfectly fine to merge this, but I do get the feeling that we should actually have gone the other way: never create fake profiles at all, and only do this fake profile thing on the *graph* side, not the actual samples side.

I think I did it on the samples side for bad reasons.

That said, the CNS issue does show that we have other crazy things going on. When merging dives, it's not actually sufficient to clear the CNS of the dive we merge - it could affect _subsequent_ dives too.

Anyway, I don't this change itself is in any way bad (except for the test failures), and I think the problems it hits are pre-existing, and unrelated.

bstoeger:
> I do get the feeling that we should actually have gone the other way: never create fake profiles at all

Yes, I was thinking the same. But mind you that the profile of manually entered dives can be edited in the UI. Only really create the profile once the user selects "edit dive"?

> I think I did it on the samples side for bad reasons.

Perhaps this behavior is an unintended side-effect of 920ff15f7160cfa6ed8d8d81e0c156dd1d7b27c6? Seems like I completely missed the point of this particular `fake_dc()` call!

> That said, the CNS issue does show that we have other crazy things going on. When merging dives, it's not actually sufficient to clear the CNS of the dive we merge - it could affect subsequent dives too.

I believe(!) that this is already done by `update_cylinder_related_info()` in `fixup_dive()`. I will have to investigate.

dirkhh:
@bstoeger are you still investigating?
Should I try to figure out why the test fails?
I am planning to do another release in a few days / maybe next weekend to bring in the printing fix and the libdivecomputer updates...

bstoeger:
Sorry, I haven't started investigating.

> Should I try to figure out why the test fails?

Yes, please.

dirkhh:
This is so fascinating...
It appears that in the Linux builds, we end up with 'drop like a stone' mode and our samples look like this
```
  <divecomputer model='csv' last-manual-time='45:00 min' deviceid='ffffffff'>
  <depth max='18.0 m' mean='16.0 m' />
  <sample time='0:00 min' depth='0.0 m' />
  <sample time='0:00 min' depth='18.0 m' />
  <sample time='37:32 min' depth='18.0 m' />
  <sample time='37:32 min' depth='5.94 m' />
  <sample time='45:00 min' depth='5.94 m' />
  <sample time='45:00 min' depth='0.0 m' />
```
(and those are the reference results that Berthold added).
But on the Mac, we take reasonable descent speed into account and the profile looks like this
```
  <divecomputer model='csv' last-manual-time='45:00 min' deviceid='ffffffff'>
  <depth max='18.0 m' mean='16.0 m' />
  <sample time='0:00 min' depth='0.0 m' />
  <sample time='1:48 min' depth='18.0 m' />
  <sample time='39:39 min' depth='18.0 m' />
  <sample time='41:16 min' depth='1.8 m' />
  <sample time='44:49 min' depth='1.8 m' />
  <sample time='45:00 min' depth='0.0 m' />
```
Doing some more debugging this suddenly becomes much more fun.
We use `prefs.ascratelast6m` as our `slope` in `fill_samples()` and then calculate the first below the surface sample to be at time `int t1 = lrint(max_d / slope);`
Can you spot the problem? If we don't initialize the prefs in this test (why???) and therefore `slope` is 0 we divide by 0 and interestingly (!!!) Linux and macOS do different things. On Linux the result of the `lrint()` is 0, but on macOS it's -1. So macOS fails to find a fake profile with this first attempt and goes to the second attempt - resulting in different profiles.

Now, there are different ways to address this.
One thing we definitely should do is not to allow the division by zero and to either hardcode a reasonably sane slope or just refuse the calculation - likely the better choice.
The other thing we should do is initialize our prefs to our default prefs. Running the test with the prefs at unreasonable values seems not really useful. And that of course means that we need to fix the results for the tests that we currently have.

I think I'll do that as a separate PR and dig through the changes this will take in the assumed test results.
And then we could rebase this one on that new PR once it is merged.

Does that sound reasonable, @bstoeger ?

bstoeger:
> The other thing we should do is initialize our prefs to our default prefs.

That sounds very reasonable to me. Thanks for debugging that. And yes, I will fix the tests.

dirkhh:
Oh wow. Setting the `default_prefs` in `testparse.cpp` results in just one failure.
The parsing of the DL7 file now only creates two dives instead of three 
I hate that file format and can't make sense of what it's trying to tell me in the first place...
@mturkia help!

bstoeger:
> Linux and macOS do different things.

Not unexpected according to https://en.cppreference.com/w/cpp/numeric/math/rint

- If arg is ±∞, FE_INVALID is raised and an *implementation-defined* value is returned
- If the result of the rounding is outside the range of the return type, FE_INVALID is raised and an *implementation-defined* value is returned
- If arg is NaN, FE_INVALID is raised and an *implementation-defined* value is returned

(emphasis added)

bstoeger:
> Oh wow. Setting the default_prefs in testparse.cpp results in just one failure.

With this:

```
diff --git a/tests/testmerge.cpp b/tests/testmerge.cpp
index 238044739..cb2f66f6a 100644
--- a/tests/testmerge.cpp
+++ b/tests/testmerge.cpp
@@ -12,6 +12,7 @@ void TestMerge::initTestCase()
 {
        /* we need to manually tell that the resource exists, because we are using it as library. */
        Q_INIT_RESOURCE(subsurface);
+       copy_prefs(&default_prefs, &prefs);
        prefs.cloud_base_url = strdup(default_prefs.cloud_base_url);
 }
 
diff --git a/tests/testparse.cpp b/tests/testparse.cpp
index 06eddaa8a..2c35e2439 100644
--- a/tests/testparse.cpp
+++ b/tests/testparse.cpp
@@ -35,6 +35,7 @@ void TestParse::initTestCase()
 {
        /* we need to manually tell that the resource exists, because we are using it as library. */
        Q_INIT_RESOURCE(subsurface);
+       copy_prefs(&default_prefs, &prefs);
        prefs.cloud_base_url = strdup(default_prefs.cloud_base_url);
 }
```
all tests still pass for me...? Could you give me your changes, please?

PS: On Friday I'm going to leave for the mountains for two weeks. In August I will be at a workshop and a conference. Don't expect too much from me at least until the last week of August.

dirkhh:
> all tests still pass for me...? Could you give me your changes, please?

Same as my change, except you don't need the cloud storage assignment anymore. That's include when you copy the preferences.

But as i said. Here the DL7 test fails.
Both in Linux and Mac.
I'll figure it out.

> PS: On Friday I'm going to leave for the mountains for two weeks. In August I will be at a workshop and a conference. Don't expect too much from me at least until the last week of August.

Enjoy. And thanks for the heads up

dirkhh:
OK, I can't seem to find a file format definition for DL7. Looking at the .zxu file I can't make any sense of this.
It appears that our tests right now expect that file to contain 3 dives. But if I import this in Subsurface, I see only two.
and with the change we also see only two.
So, err, no idea. what to do here

bstoeger:
Are you testing with current master (+your patch) or with this PR (+your patch)?

If the latter, it could be a problem introduced by this PR. With current master it works for me and import also gives 3 dives (Jan. 1, 2 and 3).

dirkhh:
Reading the spec my eyes are bleeding. This is... wow.
But I can now state that each `ZDH...ZDT` pair is one dive. So in theory that DL7 file does indeed contain three dives.
```
ZDH|1|1|I|QS|20180101101000|27|11|FO2|||
ZDT|1|1|10.0|20180101102000|25||
ZDH|2|2|I|V|20180102101000|27|11|FO2|||
ZDT|2|2|10.0|20180102110000|25||
ZDH|3|3|I|QS|20180103101000|28|11|FO2|||
ZDT|3|3|10.0|20180103102000|26||
```
One each on January 1,2,3. Yet, even current master seems to only import the first two dives. Which means for me, built locally, master fails the `make check`. Yet in GitHub Actions it passes.

I am seriously confused.

@mturkia @bstoeger - does master pass `make check` for you?

bstoeger:
Yes, master works for me (except TestProfile, for whatever reason):
```
97% tests passed, 1 tests failed out of 30

Total Test time (real) =  81.72 sec

The following tests FAILED:
          2 - TestProfile (Failed)
```

mturkia:
On Wed, Jul 21, 2021 at 9:06 PM Dirk Hohndel ***@***.***>
wrote:

> Reading the spec my eyes are bleeding. This is... wow.
> But I can now state that each ZDH...ZDT pair is one dive. So in theory
> that DL7 file does indeed contain three dives.
>
> ZDH|1|1|I|QS|20180101101000|27|11|FO2|||
> ZDT|1|1|10.0|20180101102000|25||
> ZDH|2|2|I|V|20180102101000|27|11|FO2|||
> ZDT|2|2|10.0|20180102110000|25||
> ZDH|3|3|I|QS|20180103101000|28|11|FO2|||
> ZDT|3|3|10.0|20180103102000|26||
>
> One each on January 1,2,3. Yet, even current master seems to only import
> the first two dives. Which means for me, built locally, master fails the make
> check. Yet in GitHub Actions it passes.
>
> I am seriously confused.
>
> @mturkia <https://github.com/mturkia> @bstoeger
> <https://github.com/bstoeger> - does master pass make check for you?
>
100% tests passed, 0 tests failed out of 30

>

bstoeger:
Perhaps some locate thing?
I currently use:
```
bs@laptop:~$ locale
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC=de_AT.UTF-8
LC_TIME=de_AT.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=de_AT.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER=de_AT.UTF-8
LC_NAME=de_AT.UTF-8
LC_ADDRESS=de_AT.UTF-8
LC_TELEPHONE=de_AT.UTF-8
LC_MEASUREMENT=de_AT.UTF-8
LC_IDENTIFICATION=de_AT.UTF-8
LC_ALL=
```

bstoeger:
Heh... Using valgrind, DL7 parsing also fails for me:
```
FAIL!  : TestParse::parseDL7() Compared values are not the same
   Actual   (dive_table.nr): 2
   Expected (3)            : 3
   Loc: [/home/bs/src/subsurface/tests/testparse.cpp(477)]
```
Looks like an uninitialzed variable thing...

dirkhh:
yes, my builds tend to be with ASAN enabled. So it's possible that this was triggered by uninitialized memory. I'll try to dig some more into this.

dirkhh:
and yes, creating a clean build without ASAN makes this indeed pass.
lovely, lovely, lovely...

dirkhh:
The best part, though? ASAN doesn't actually SHOW an uninitialized memory access. Oh how I love these Heisen-Bugs

bstoeger:
Uh, guys...?

In `parse_dan_format()`:
```
                if (try_to_xslt_open_csv(filename, &mem_csv, "csv"))
                        return -1;
                ...
                end_ptr += ptr - (char *)mem_csv.buffer;
```
but in `try_to_xslt_open_csv()`
```
        buf = realloc(mem->buffer, mem->size + 7 + strlen(tag) * 2 + amp * 4);
        ...
        mem->buffer = buf;
```

Why on earth does that function change the address of the buffer!?

And what is this supposed to mean:
```
                        if (iter && iter + 1) {
```
(iter being a pointer). Can this *ever* make sense?

dirkhh:
Yeah, we are both staring at the same code. The `if (iter && iter + 1)` could mean that `iter` is neither `NULL` nor `-1`... or it could be that this was meant to be `*iter && *(iter + 1)`
But it's the `realloc` that I worry about. I have instrumented the code some more to see if that's why it fails.

dirkhh:
And yes, the failure is indeed the `realloc`... my speculation would be that without a memory debugger it happens to work because there's enough space and it can expand the buffer in place? Or something crazy like that.
YUCK.
Are you going to try to fix it, or do you want me to deal with the problem?

bstoeger:
Of course that's why it fails. The value that comes out is essentially random. I guess in the non-valgrind case we were just lucky enough that the buffer was always big enough and there was no actual reallocation?

bstoeger:
If you could try to fix this, that would be great.

dirkhh:
> If you could try to fix this, that would be great.

will do

dirkhh:
Trivial fix
```
diff --git a/core/import-csv.c b/core/import-csv.c
index f24c5ec0b..34ac1348b 100644
--- a/core/import-csv.c
+++ b/core/import-csv.c
@@ -245,6 +245,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
                        return -1;
                }
                mem_csv.size = ptr - (char*)mem_csv.buffer;
+               end_ptr += ptr - (char *)mem_csv.buffer;
 
                iter = parse_dan_new_line(ptr + 1, NL);
                if (iter && strncmp(iter, "ZDT", 3) == 0) {
@@ -268,7 +269,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
                        return -1;
 
                ret |= parse_xml_buffer(filename, mem_csv.buffer, mem_csv.size, table, trips, sites, devices, filter_presets, params);
-               end_ptr += ptr - (char *)mem_csv.buffer;
+
                free(mem_csv.buffer);
        }
 
```
Simply not doing pointer math on the re-allocated buffer is all it takes.

dirkhh:
rudely pushed the fix into your branch

dirkhh:
Well, that was a clear improvement, we went from **one** failed check to **two**
🤷🏼‍♂️

dirkhh:
Ahh, I see, the `copy_prefs()` is missing in this branch.
-- err, no, that's not it. what the heck

bstoeger:
Of course the created profiles are now different and the tests have to be adapted. I'll look at it tomorrow.

dirkhh:
OK, last update from me for a little bit.

This branch still had the incorrect test reference info in it and is missing the change to add the `default_prefs` - so I'm guessing that @bstoeger simply hasn't pushed that part, yet.

I simply added two more commits to deal with this so that we could see if this passes muster now. But that last commit definitely should be squashed into your earlier one, @bstoeger - I can do that if you want, but I didn't just want to mess with your commits... rude enough that I'm pushing into your branch...

mturkia:
On Wed, Jul 21, 2021 at 10:18 PM bstoeger ***@***.***> wrote:when
querying the status over RPC I do see that the height is increasing and

> Uh, guys...?
>
> In parse_dan_format():
>
>                 if (try_to_xslt_open_csv(filename, &mem_csv, "csv"))
>                         return -1;
>                 ...
>                 end_ptr += ptr - (char *)mem_csv.buffer;
>
> but in try_to_xslt_open_csv()
>
>         buf = realloc(mem->buffer, mem->size + 7 + strlen(tag) * 2 + amp * 4);
>         ...
>         mem->buffer = buf;
>
> Why on earth does that function change the address of the buffer!?
>
realloc returns a pointer to the original buffer, or to address of new
buffer, if the buffer had to be relocated to increase its size. Naturally
we need to keep track of the current/new location of the memory buffer.

> And what is this supposed to mean:
>
>                         if (iter && iter + 1) {
>
> (iter being a pointer). Can this *ever* make sense?
>
if (iter) checks if the pointer is null.
if (iter + 1) checks if the next character in the string is null, i.e. if
the found '|' character is the last character in the string

>

Subsurface-divelog:
>>                 if (try_to_xslt_open_csv(filename, &mem_csv, "csv"))
>>                         return -1;
>>                 ...
>>                 end_ptr += ptr - (char *)mem_csv.buffer;
>>
>> but in try_to_xslt_open_csv()
>>
>>         buf = realloc(mem->buffer, mem->size + 7 + strlen(tag) * 2 +
>amp * 4);
>>         ...
>>         mem->buffer = buf;
>>
>> Why on earth does that function change the address of the buffer!?
>>
>realloc returns a pointer to the original buffer, or to address of new
>buffer, if the buffer had to be relocated to increase its size.
>Naturally
>we need to keep track of the current/new location of the memory buffer.

But the code then calculated that offset by subtracting a pointer relative to the old buffer from the new buffer address. Which create the garbage result.

See the second to last commit.

>
>> And what is this supposed to mean:
>>
>>                         if (iter && iter + 1) {
>>
>> (iter being a pointer). Can this *ever* make sense?
>>
>if (iter) checks if the pointer is null.
>if (iter + 1) checks if the next character in the string is null, i.e.
>if
>the found '|' character is the last character in the string

That would be *(iter+1), not just iter+1, though
>
>>

-- 
from my phone

mturkia:
On Thu, Jul 22, 2021 at 7:36 AM Subsurface ***@***.***> wrote:

>
>
>
> >> if (try_to_xslt_open_csv(filename, &mem_csv, "csv"))
> >> return -1;
> >> ...
> >> end_ptr += ptr - (char *)mem_csv.buffer;
> >>
> >> but in try_to_xslt_open_csv()
> >>
> >> buf = realloc(mem->buffer, mem->size + 7 + strlen(tag) * 2 +
> >amp * 4);
> >> ...
> >> mem->buffer = buf;
> >>
> >> Why on earth does that function change the address of the buffer!?
> >>
> >realloc returns a pointer to the original buffer, or to address of new
> >buffer, if the buffer had to be relocated to increase its size.
> >Naturally
> >we need to keep track of the current/new location of the memory buffer.
>
> But the code then calculated that offset by subtracting a pointer relative
> to the old buffer from the new buffer address. Which create the garbage
> result.
>
> See the second to last commit.
>

No idea what commit you are referring to. If I am looking at correct part
of the code, the mem->buffer has been updated to the new buffer before
these calculations.

>
> >
> >> And what is this supposed to mean:
> >>
> >> if (iter && iter + 1) {
> >>
> >> (iter being a pointer). Can this *ever* make sense?
> >>
> >if (iter) checks if the pointer is null.
> >if (iter + 1) checks if the next character in the string is null, i.e.
> >if
> >the found '|' character is the last character in the string
>
> That would be *(iter+1), not just iter+1, though


Nope. It is a char pointer. So iter+1 points to the character at index 1 of
the buffer. *(iter + 1) points to wherever in memory based on the content
of the buffer.

char buf[] = "asdf";
printf("%s\n", buf + 1);

This will print out: "sdf"

miika

>

dirkhh:
Umm. Ok. So.
If you take a look at https://github.com/subsurface/subsurface/blob/master/core/import-csv.c#L267-L271 in current master you'll see this
```
		if (try_to_xslt_open_csv(filename, &mem_csv, "csv"))
			return -1;

		ret |= parse_xml_buffer(filename, mem_csv.buffer, mem_csv.size, table, trips, sites, devices, filter_presets, params);
		end_ptr += ptr - (char *)mem_csv.buffer;
```
The pointer `ptr` is set earlier in the code - it's an offset into the `mem_csv.buffer`.
The call to `try_to_xslt_open_csv()` reallocates the `mem_csv.buffer`. If you are unlucky, that reallocation moves the base address of the buffer. Which means that the value of `end_prt` becomes meaningless garbage.
I don't know a different way to explain this.

That's fixed by moving the `end_ptr` calculation earlier in the function, before the call to `try_to_xslt_open_csv()` where it will actually create that offset that we can then reuse in the next iteration of the loop. That's 55ac2da1bb94d 

To the other point.
Yes, printing `buf + 1` as a string does indeed print what you describe. That's because printing a string dereferences the string. But that's not the code in question. The code in question is
```
if (iter && iter + 1)
```
Let's assume our trusty pointer `iter` points to `0x44200`.
Then the first half of this condition checks that `iter` is not the `NULL` pointer. And the second half of the condition checks that `iter` is not the `-1` pointer. Because `iter + 1` is `0x44201`.
What it absolutely, definitely, 100% NOT does is check the content of the string that `iter` points to.

Case in point.
```
char buf[2] = '\0\0';
char *iter = buf;
if (iter  && iter + 1)
   fprintf(stderr, "yeah, this string will always be printed, regardless of what's in the buffer");
```

mturkia:
> Umm. Ok. So.
> If you take a look at https://github.com/subsurface/subsurface/blob/master/core/import-csv.c#L267-L271 in current master you'll see this
> 
> ```
> 		if (try_to_xslt_open_csv(filename, &mem_csv, "csv"))
> 			return -1;
> 
> 		ret |= parse_xml_buffer(filename, mem_csv.buffer, mem_csv.size, table, trips, sites, devices, filter_presets, params);
> 		end_ptr += ptr - (char *)mem_csv.buffer;
> ```
> 
> The pointer `ptr` is set earlier in the code - it's an offset into the `mem_csv.buffer`.
> The call to `try_to_xslt_open_csv()` reallocates the `mem_csv.buffer`. If you are unlucky, that reallocation moves the base address of the buffer. Which means that the value of `end_prt` becomes meaningless garbage.
> I don't know a different way to explain this.
> 
> That's fixed by moving the `end_ptr` calculation earlier in the function, before the call to `try_to_xslt_open_csv()` where it will actually create that offset that we can then reuse in the next iteration of the loop. That's [55ac2da](https://github.com/subsurface/subsurface/commit/55ac2da1bb94d00770a120572cb07da8a28d9e60)

OK, I was looking at incorrect spot on the code. Will need to take a look into this one in the evening.

> To the other point.
> Yes, printing `buf + 1` as a string does indeed print what you describe. That's because printing a string dereferences the string. But that's not the code in question. The code in question is

And that is so obvious, that it should have been screaming to me even before sending the previous comment.

bstoeger:
I would suggest `if (iter && iter[1])` as the least cryptic variant.

bstoeger:
Not really related to this PR, but note that `TestProfile` fails for me owing to some locale issue. All numbers in `save_profiledata` are written with commas as decimal separator.

I believe that we set the locale to "C" on startup, but have't found the code that does this.

dirkhh (commented):
1f10c5d9dc49fe contains the word 'cylinger' in the commit message :)
also, did you want to add the `if (iter && iter + 1)` fix in this PR?

dirkhh:
added the `iter && iter + 1` fix and merged locally

dirkhh:
ugh, crap, pushed the wrong branch... so the 'cylinger' isn't fixed.
oh well, not messing with the master branch, so that stays in there.

PR #3283, opened 2021-07-18
sanitize cylinder-ids of pressure samples

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes long-standing out-of-bound accesses
### Changes made:

The two pressure-sensor values of samples were initialized to 0 and 1 by default. However, many dives don't have two cylinders. This produces out-of-bound accesses in the merging code. They were completely harmless as long as there was a fixed-size cylinder table. With the dynamic cylinder table, this can, at least theoretically, lead to crashes (access past a virtual memory page? I don't understand anything of that.).

This PR tries to fix that by sanitizing the cylinder ids in the parsers and in `fixup_dive()`. It is completely untested and will of course lead to massive breakage of the parsing tests.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1208,6 +1208,17 @@ static void fixup_no_o2sensors(struct divecomputer *dc)
 	}
 }
 
+static void fixup_dc_sample_sensors(struct divecomputer *dc, int nr_cylinders)
+{
+	for (int i = 0; i < dc->samples; i++) {
+		struct sample *s = dc->sample + i;
+		for (int j = 0; j < MAX_SENSORS; j++) {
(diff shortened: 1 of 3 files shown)


bstoeger:
> will of course lead to massive breakage of the parsing tests.

It doesn't. Weird.

bstoeger:
Ouch. Apparently parts of the code don't check for `sensor == -1`. This will need some sort of audit - closing for now.

bstoeger:
Fantastic - I didn't realize that sensor is a `uint8_t`, so this was never supposed to support "no sensor".

PR #3286, opened 2021-07-19
Fix coverity reported memory leaks

mturkia:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Coverity reported 2 memory leaks on recent divelogs.de export change. This fixes those.

### DIFF
--- a/core/uploadDiveLogsDE.cpp
+++ b/core/uploadDiveLogsDE.cpp
@@ -104,8 +104,10 @@ bool uploadDiveLogsDE::prepareDives(const QString &tempfile, bool selected)
 		 * Get the i'th dive in XML format so we can process it.
 		 * We need to save to a file before we can reload it back into memory...
 		 */
-		if (selected && !dive->selected)
+		if (selected && !dive->selected) {
+			free_xml_params(params);
 			continue;
+		}
 
 		/* make sure the buffer is empty and add the dive */
 		mb.len = 0;
@@ -151,6 +153,7 @@ bool uploadDiveLogsDE::prepareDives(const QString &tempfile, bool selected)
 			zip_close(zip);
 			QFile::remove(tempfile);
 			xsltFreeStylesheet(xslt);
+			free_xml_params(params);
 			return false;
 		}
 		free_buffer(&mb);


core/uploadDiveLogsDE.cpp
```diff
@@ -151,6 +153,7 @@ bool uploadDiveLogsDE::prepareDives(const QString &tempfile, bool selected)
 			zip_close(zip);
 			QFile::remove(tempfile);
 			xsltFreeStylesheet(xslt);
+			free_xml_params(params);
```
bstoeger, line 156: It seems to me that `membuf` still leaks. Also on normal termination of the loop body!?

C-style memory management is so unnecessarily painful. I would suggest using C++ primitives or C++ wrappers in source files that are C++ anyway.

dirkhh: Well - I understand the desire to go to things that are comfortable and familiar.
I don't think this is a moment where it makes sense to throw aways a very useful helper that is used all over the code.
I think Miika's fix is valid and addresses and issue, so let's merge it.

dirkhh (approved):
Seems reasonable

core/uploadDiveLogsDE.cpp
bstoeger: It's not about familiarity (or comfort), but the fact that nobody here (myself included) seems to be able to get this right.

How does this even work:

```
                membuf = mb_cstring(&mb);
```
get a pointer to the memory buffer string
```
                free_buffer(&mb);
```
free the buffer
```
                s = zip_source_buffer(zip, membuf, streamsize, 1);
```
use it again!? How?

In C code, the helper may be useful. In C++ code it is actively harmful, because it hides such problems from the static checkers.

mturkia: That seems to be a bug indeed. But it is unrelated to the changes I did now. Would be nice if Coverity would catch also that kind of bugs. Anyway, I am totally illiterate when it comes to C++ so cannot promise to switch to writing proper C++ code anytime soon :(

PR #3289, opened 2021-07-20
cleanup: remove function static variables

bstoeger:
There were two function-static variables in ToolTipItem::refresh(),
which is a very scary proposition. Curently, there is only
one ToolTipItem, but this may change on mobile, where there
are multiple profiles at the same time.

Remove this timebomb and make the two objects subobjects.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Removes a dangerous construct: two function-static objects. This might give hard to trackdown issues, when there are multiple ToolTipItems.

[Editorial note: static variables are _not_ allocated on the stack, as was stated in the commit message of the commit adding these variables]

### Changes made:

1) Turn static variables into member objects.


dirkhh (approved):
Yeah, when this was written I bet no one considered that we'd do multiple interactive profiles (and therefore multiple `ToolTipItem`s)
This makes sense.

PR #3291, opened 2021-07-20
core: use int16_t for sensor-id

bstoeger:
The sensor-id in the sample struct was a uint8_t, with all
the known problems of unsigned integers. In the rest of the
code cylinder ids are signed integers. To avoid confusion,
make it a signed int. int8_t should be enough (max. 127
cylinders). To allow for degenerate cases, use an int16_t.
16k cylinders should be enough for everyone.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Use `int16_t` for sensor cylinder-ids. Should fix a (in my opinion questionable) Coverity warning.

Feel free to change to `int8_t` if you feel that `int16_t` eats too much memory.

Completely untested!

### DIFF
--- a/core/sample.h
+++ b/core/sample.h
@@ -9,7 +9,7 @@ extern "C" {
 #endif
 
 #define MAX_SENSORS 2
-#define NO_SENSOR ((uint8_t)-1)
+#define NO_SENSOR -1
 
 struct sample                         // BASE TYPE BYTES  UNITS    RANGE               DESCRIPTION
 {                                     // --------- -----  -----    -----               -----------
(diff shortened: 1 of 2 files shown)


core/sample.h
```diff
@@ -21,11 +21,11 @@ struct sample                         // BASE TYPE BYTES  UNITS    RANGE
...
+	pressure_t pressure[MAX_SENSORS]; // int32_t  2x4    mbar   (0-2 Mbar)             cylinder pressures (main and CCR o2)
 	o2pressure_t setpoint;            // uint16_t   2    mbar   (0-65 bar)             O2 partial pressure (will be setpoint)
-	o2pressure_t o2sensor[3];         // uint16_t   6    mbar   (0-65 bar)             Up to 3 PO2 sensor values (rebreather)
+	o2pressure_t o2sensor[3];         // uint16_t 3x6    mbar   (0-65 bar)             Up to 3 PO2 sensor values (rebreather)
 	bearing_t bearing;                // int16_t    2  degrees  (-1 no val, 0-360 deg) compass bearing
-	uint8_t sensor[MAX_SENSORS];      // uint8_t    1  sensorID (0-254)                ID of cylinder pressure sensor
+	int16_t sensor[MAX_SENSORS];      // int16_t  2x2  sensorID (0-16k)                ID of cylinder pressure sensor
 	uint16_t cns;                     // uint16_t   1     %     (0-64k %)              cns% accumulated
 	uint8_t heartbeat;                // uint8_t    1  beats/m  (0-255)                heart rate measurement
 	volume_t sac;                     //            4  ml/min                          predefined SAC
```
dirkhh, line 31: would be nice to update the total size of the structure below. And man, it's huge. Given that we have a LOT of them in a large dive log with multiple dive computers, potentially short sample rate
Given that no one dives with more than 30 tanks (even if you look at the extreme dives to 300m), why aren't 127 sensors enough? Isn't this overkill?

dirkhh (commented):
This is of course a solution.
I wonder if it is a better solution than `int8_t` - and why

core/sample.h
```diff
@@ -21,18 +21,18 @@ struct sample                         // BASE TYPE BYTES  UNITS    RANGE
...
-	uint16_t cns;                     // uint16_t   1     %     (0-64k %)              cns% accumulated
+	int16_t sensor[MAX_SENSORS];      // int16_t  2x2  sensorID (0-16k)                ID of cylinder pressure sensor
+	uint16_t cns;                     // uint16_t   2     %     (0-64k %)              cns% accumulated
 	uint8_t heartbeat;                // uint8_t    1  beats/m  (0-255)                heart rate measurement
 	volume_t sac;                     //            4  ml/min                          predefined SAC
 	bool in_deco;                     // bool       1    y/n      y/n                  this sample is part of deco
 	bool manually_entered;            // bool       1    y/n      y/n                  this sample was entered by the user,
 					  //                                               not calculated when planning a dive
-};	                                  // Total size of structure: 57 bytes, excluding padding at end
+};	                                  // Total size of structure: 63 bytes, excluding padding at end
```
dirkhh, line 35: well, just below 64 bytes so... yeah... ok.
I'd still be curious why the signed 8 bits isn't sufficient.

core/sample.h
bstoeger: > would be nice to update the total size of the structure below. 

I didn't see that. Could you check my calculations - I don't think that they were correct before.

> Given that no one dives with more than 30 tanks (even if you look at the extreme dives to 300m), why aren't 127 sensors enough? Isn't this overkill?

They should of course be enough. Especially given that the highest tank that is entered here is the first O2 cylinder. I was just thinking, let's cover even pathological cases. This discussion is why I was reluctant with this PR. Feel free to replace it.

dirkhh: I think the better answer would be to go to `int8_t`.
If someone needs more than 127 sensors they are so outside of everything that we have ever tested... How would that even work on the profile?
No, let's just do `int_8` here

bstoeger: Nitpick: it would be two sensors, but >127 cylinders.

dirkhh: well, but don't we support an arbitrary number of cylinders? So should this be to 128bit integers for the people who have more than 4 billion cylinders with them?

dirkhh: clearly, this was too snarky of a response. my apologies

more seriously - where would we want to take this? 
I *think* that `int8_t` would be the right answer. And that there might be value in restricting ourselves to 63 cylinders because at that point so much of our infrastructure already breaks. Visualization is shite. The deco algorithms really aren't validated at all for the types of dive where you'd use more than half a dozen cylinders, and certainly once you are in the middle double digits all of this is way outside what we know how to do...
and by going to a limit of 63 for actual dives, the funky "twice the number of cylinders" requirement of the planner is still easily managed.

Thoughts?

Or should I just stop bike shedding and over thinking and merge this?

dirkhh:
going back and forth I ended up taking this as is. Yes, 127 is more than enough - but... yeah, I didn't want to mess with things again. Merged locally

PR #3294, opened 2021-07-22
desktop: explicitly enable shortcuts

### LINKED ISSUE #3293: Graphics when there are two computers
Con la versione 5.0.2 non riesco a passare da un grafico all'altro premendo i tasti freccia destro/sinistro e resto sempre sul computer 1/2

I cannot switch between graphs by pressing the left / right arrow keys and I can see only computer 1. The version is 5.0.2

dirkhh:
The changes in commit 4daf687876 ("profile: remove [disable|enable]Shortcuts()
signals") resulted in us no longer enabling the shortcuts on the desktop (at
least on macOS where I debugged this). This placement of the call feels like a
bit of overkill, but at least it shouldn't be wrong.

- [x] Bug fix

### Changes made:

rather brute force ensure that the shortcuts on the desktop get enabled

### Related issues:

Fixes #3293

### Release note:

done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- fix broken curser left/right shortcut for showing multiple dive computers
 
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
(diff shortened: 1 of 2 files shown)


dirkhh:
I believe Berthold is off on vacation, so let's keep moving forward here...

PR #3298, opened 2021-08-02
export: use unique temporary file for divelogs.de upload

rfuchs:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

On multi-user systems with a shared directory for temporary files, using
a static file name can lead to permissions problems and subsequent
errors due to collisions. Use a random unique file name for each
generated file to avoid these problems.

Note: the temporary file generated from the divelogs.de upload is still
left behind after the upload finishes.

### Changes made:

Change from a static temporary file name to a generated random unique file name using QTemporaryFile

### DIFF
--- a/core/uploadDiveLogsDE.cpp
+++ b/core/uploadDiveLogsDE.cpp
@@ -2,6 +2,7 @@
 #include "uploadDiveLogsDE.h"
 #include <QDir>
 #include <QDebug>
+#include <QTemporaryFile>
 #include <zip.h>
 #include <errno.h>
 #include "core/display.h"
@@ -34,21 +35,22 @@ uploadDiveLogsDE::uploadDiveLogsDE():
 }
 
 
+static QString makeTempFileName()
+{
+	QTemporaryFile tmpfile;
+	tmpfile.setFileTemplate(QDir::tempPath() + "/divelogsde-upload.XXXXXXXX.dld");
+	tmpfile.open();
+	QString filename(tmpfile.fileName());
+	tmpfile.close();
+	return filename;
+}
+
+
 void uploadDiveLogsDE::doUpload(bool selected, const QString &userid, const QString &password)
 {
 	QString err;
 
-
-	/* generate a temporary filename and create/open that file with zip_open */
-	QString filename(QDir::tempPath() + "/divelogsde-upload.dld");
-
-	// delete file if it exist
-
-	QFile f(filename);
-	if (f.open(QIODevice::ReadOnly)) {
-		f.close();
-		f.remove();
-	}
+	QString filename = makeTempFileName();
 
 	// Make zip file, with all dives, in divelogs.de format
 	if (!prepareDives(filename, selected)) {


core/uploadDiveLogsDE.cpp
```diff
@@ -40,15 +41,13 @@ void uploadDiveLogsDE::doUpload(bool selected, const QString &userid, const QStr
...
-	QFile f(filename);
-	if (f.open(QIODevice::ReadOnly)) {
-		f.close();
-		f.remove();
-	}
+	auto *tmpfile = new QTemporaryFile();
+	tmpfile->setFileTemplate(QDir::tempPath() + "/divelogsde-upload.XXXXXXXX.dld");
+	tmpfile->open();
+	QString filename = tmpfile->fileName();
+	tmpfile->close();
```
atdotde: This looks like almost the same race condition you are trying to resolve: What happens if the other process creates the file between you closing it here and it being reopened for writing? This only produces a filename that at some point in the past did not exist. Once you opened it above you should write the proper contents to it.

atdotde:
I like the idea but I think the execution needs improvement.

core/uploadDiveLogsDE.cpp
rfuchs: That could only happen if both users generate the same random file name at the same time, which is highly unlikely.

From what I see elsewhere in the code, the only way to directly use the temporary file name for zip_open is using an OS- and version-specific #ifdef case distinction and dup()ing the underlying fd, which I thought might be worth avoiding.

atdotde: Ah, now that you mention it, I remember there was a problem with libzip IIRC that it cannot be used with a file handle only with a filename. You could use the PID in the template which would at least make sure you do not step on the foot of another incarnation of Subsurface on the same machine. 

I have to admit I don't know how Qt fills the pattern, if it tries a random string or some deterministic way to find the 'first' non-existent file name.

rfuchs: On my system this comes out as /tmp/divelogsde-upload.mBToUtKs.dld so I assume it's random. I can add the PID if that helps, even though I'm not sure what to use because I assume getpid() is not portable (QCoreApplication::applicationPid() ?  - I'm not a Qt guy really). Another option would be to use a UUID, or even make sure each instance has its own private tmp dir (although I would consider that a more invasive change). Or perhaps leave the file name static and instead go for a dedicated temp dir via QTemporaryDir.

(Also in case this wasn't clear from my description: What this tries to solve is not a race condition. The invocations of the upload by two different users could be days apart. The second one will fail because the file (left over by the first one) would already exist but is owned by a different user and so is not deletable.)

mturkia: That function should give us a random file name based on the given template, and IMO that should be enough for us. Anything more complex sounds overkill :D

core/uploadDiveLogsDE.cpp
```diff
@@ -40,15 +41,13 @@ void uploadDiveLogsDE::doUpload(bool selected, const QString &userid, const QStr
...
-	QString filename(QDir::tempPath() + "/divelogsde-upload.dld");
-
-	// delete file if it exist
-
-	QFile f(filename);
-	if (f.open(QIODevice::ReadOnly)) {
-		f.close();
-		f.remove();
-	}
+	auto *tmpfile = new QTemporaryFile();
```
bstoeger: Please no plain `new`! We have so frustratingly many leaks owing to this archaic lifetime management. The simplest way would be to do this in a separate function, as such:
```
static QString make_unique_filename()
{
        QTemporaryFile tmpfile;
        tmpfile->setFileTemplate(QDir::tempPath() + "/divelogsde-upload.XXXXXXXX.dld");
        return tmpfile->fileName();
}
```

bstoeger (changes requested):
See comment above: No plain `new` in C++ code, please.

core/uploadDiveLogsDE.cpp
rfuchs: Done

atdotde (approved):
Thumbs up

dirkhh (approved):
We still don't delete the file after we're done, if I read this correctly, but this is of course a significant improvement!
Nice. And Thank You

PR #3299, opened 2021-08-03
export: clean up temp file after divelogs.de upload

rfuchs:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This adds a cleanup function to be called after a divelogs.de upload
finishes (successful or not) to make sure the temporary zip file is
closed and removed.

### Changes made:

1) New method uploadDiveLogsDE::cleanupTempFile()
2) Call this method upon completion of upload

### DIFF
--- a/core/uploadDiveLogsDE.cpp
+++ b/core/uploadDiveLogsDE.cpp
@@ -40,9 +40,7 @@ static QString makeTempFileName()
 	QTemporaryFile tmpfile;
 	tmpfile.setFileTemplate(QDir::tempPath() + "/divelogsde-upload.XXXXXXXX.dld");
 	tmpfile.open();
-	QString filename(tmpfile.fileName());
-	tmpfile.close();
-	return filename;
+	return tmpfile.fileName();
 }
 
 
@@ -196,6 +194,13 @@ bool uploadDiveLogsDE::prepareDives(const QString &tempfile, bool selected)
 }
 
 
+void uploadDiveLogsDE::cleanupTempFile()
+{
+	if (tempFile.isOpen())
+		tempFile.remove();
+}
+
+
 void uploadDiveLogsDE::uploadDives(const QString &filename, const QString &userid, const QString &password)
 {
 	QHttpPart part1, part2, part3;
@@ -219,12 +224,15 @@ void uploadDiveLogsDE::uploadDives(const QString &filename, const QString &useri
 	// prepare header with filename (of all dives) and pointer to file
 	args = "form-data; name=\"userfile\"; filename=\"" + filename + "\"";
 	part1.setRawHeader("Content-Disposition", args.toLatin1());
-	QFile *f = new QFile(filename);
-	if (!f->open(QIODevice::ReadOnly)) {
+
+	// open new file for reading
+	cleanupTempFile();
+	tempFile.setFileName(filename);
+	if (!tempFile.open(QIODevice::ReadOnly)) {
 		qDebug() << "ERROR opening zip file: " << filename;
 		return;
 	}
-	part1.setBodyDevice(f);
+	part1.setBodyDevice(&tempFile);
 	multipart->append(part1);
 
 	// Add userid
@@ -289,6 +297,7 @@ void uploadDiveLogsDE::uploadFinishedSlot()
 	QByteArray xmlData = reply->readAll();
 	reply->deleteLater();
 	reply = NULL;
+	cleanupTempFile();
 	char *resp = xmlData.data();
 	if (resp) {
 		char *parsed = strstr(resp, "<Login>");
@@ -324,6 +333,7 @@ void uploadDiveLogsDE::uploadTimeoutSlot()
 		reply->deleteLater();
 		reply = NULL;
 	}
+	cleanupTempFile();
 	QString err(tr("divelogs.de not responding"));
 	report_error(err.toUtf8());
 	emit uploadFinish(false, err);
@@ -337,6 +347,7 @@ void uploadDiveLogsDE::uploadErrorSlot(QNetworkReply::NetworkError error)
 		reply->deleteLater();
 		reply = NULL;
 	}
+	cleanupTempFile();
 	QString err(tr("network error %1").arg(error));
 	report_error(err.toUtf8());
 	emit uploadFinish(false, err);
--- a/core/uploadDiveLogsDE.h
+++ b/core/uploadDiveLogsDE.h
@@ -4,6 +4,7 @@
 #include <QNetworkReply>
 #include <QHttpMultiPart>
 #include <QTimer>
+#include <QFile>
 
 
 class uploadDiveLogsDE : public QObject {
@@ -28,10 +29,12 @@ private:
 	uploadDiveLogsDE();
 
 	void uploadDives(const QString &filename, const QString &userid, const QString &password);
+	void cleanupTempFile();
 
 	// only to be used in desktop-widgets::subsurfacewebservices
 	bool prepareDives(const QString &tempfile, bool selected);
 
+	QFile tempFile;
 	QNetworkReply *reply;
 	QHttpMultiPart *multipart;
 	QTimer timeout;


core/uploadDiveLogsDE.cpp
```diff
@@ -196,6 +196,15 @@ bool uploadDiveLogsDE::prepareDives(const QString &tempfile, bool selected)
 }
 
 
+void uploadDiveLogsDE::cleanupTempFile()
+{
+	if (tempFile.isOpen()) {
+		tempFile.close();
+		tempFile.remove();
```
dirkhh, line 200: so if the file is not open, we don't remove it? I'm confused.

dirkhh: also, `.remove()` closes the file if necessary, according to the Qt documentation, so the `.close()` call is redundant.

dirkhh (commented):
that logic of how you handle the `tempFile` confuses me a bit.
You don't initialize the `QFile`, but IIRC the default constructor creates simply an invalid `QFile`. Which means the `.isOpen()` will fail? But wouldn't it make more sense to check if the file exists? So something like 
```
    if (tempFile.exists())
        tempFile.remove();
```

bstoeger:
Intuitively, that looks bogus to me (though I think I understand what it is trying to achieve). But that could be due to the whole source file being ... not pretty, to put it mildly. I'll check it out in the upcoming days (still travelling / hiking).

PS: The `close()` invocation in the previous commit is definitely bogus. The file is closed automatically if the object goes out of scope. That's just pointless pessimization/bloat of the code.

rfuchs:
1) uploadDiveLogsDE is instantiated. QFile tempFile is invalid and not open.
2) doUpload() is called. QTemporaryFile is used to create a random temp file name. The temp file is removed but the file name is kept and passed to prepareDives().
3) prepareDives() passes the file name to zip_open() to create the actual temp file.
4) uploadDives uses QFile tempFile to open the just created temp file and passes it to QHttpPart as upload body.
5) Upload is performed.
6) Upload finishes and callbacks fire. QFile tempFile is open (was being read during the upload) and can now be closed and removed.
7) If another upload is done, repeat from step 2 with QFile tempFile now closed again. (At least that's my understanding that uploadDiveLogsDE is created only once, correct me if wrong.)

I agree that close/remove is redundant according to the docs but that's how I've seen it done in other parts of the code 🤷

dirkhh:
> 1. uploadDiveLogsDE is instantiated. QFile tempFile is invalid and not open.
> 2. doUpload() is called. QTemporaryFile is used to create a random temp file name. The temp file is removed but the file name is kept and passed to prepareDives().

But strangely it's only removed if it is open - that's the part that confuses me.

rfuchs:
> But strangely it's only removed if it is open - that's the part that confuses me.

It's open if it was used for an upload before which has finished and so it can now be removed (either in one of the callbacks, or if a second upload is started and none of the callbacks fired for some reason)

bstoeger (approved):
After a cursory glance: As I feared, many things are wrong with this source file. It starts with the class being a singleton (why?). Also the `connect()` calls seem to be redone multiple times, no? Same for `desktop-widgets/subsurfacewebservices.cpp`. What a mess.

This PR is a distinct improvement in that at least the `QFile` object does no longer leak. So I would say - if it is tested - go ahead after following Dirk's suggestion of removing the `close()` call.

But really, the whole thing should be redone. Moreover, I'd still suggest simplifying the temporary-filename function by losing the `close()` call and returning the string directly without all these string-assignment shenanigans. That's pointless bloat.

Ultimately however, this file should be redone.

> I agree that close/remove is redundant according to the docs but that's how I've seen it done in other parts of the code.

Please, don't copy questionable practices. Always strive for the simple solution.

rfuchs:
> This PR is a distinct improvement in that at least the `QFile` object does no longer leak. So I would say - if it is tested - go ahead after following Dirk's suggestion of removing the `close()` call.

Done.

> But really, the whole thing should be redone. Moreover, I'd still suggest simplifying the temporary-filename function by losing the `close()` call and returning the string directly without all these string-assignment shenanigans. That's pointless bloat.
> 
> Ultimately however, this file should be redone.

Happy to help rework it if you can point me to a sample that you consider acceptable (obv not `uploadDiveShare` as that seems to be doing the same thing).

> > I agree that close/remove is redundant according to the docs but that's how I've seen it done in other parts of the code.
> 
> Please, don't copy questionable practices. Always strive for the simple solution.

That seems to be in the eye of the beholder as I work(ed) with some developers who have a strict "everything that is opened must be closed explicitly" policy even if made redundant by the scope, which personally I never understood, but since it's not my code I have no problem with staying in line with what is done elsewhere. But anyway, happy to convert it to simpler code.

bstoeger:
> Done.

Thank you.

> Happy to help rework it if you can point me to a sample that you consider acceptable (obv not uploadDiveShare as that seems to be doing the same thing).

To be frank, your energy and time is probably better spent at different parts of the code. Of course, if you really want to improve the quality of this particular code, I'm not going to object. My first intuition would be to un-singletonify the whole thing. Also, the connect-statements have to be checked.

> That seems to be in the eye of the beholder as I work(ed) with some developers who have a strict "everything that is opened must be closed explicitly" policy even if made redundant by the scope, which personally I never understood, but since it's not my code I have no problem with staying in line with what is done elsewhere. 

Let me guess: these were (ex-)C programmers? I see where they're coming from, but ultimately this habit is bad at many levels. It clutters the code and makes it less readable. It is understood that when an object goes out of scope all its resources (memory, file handles, etc.) are freed. So if I see a `f.close()` I will wonder why it is there. One would use this construct to "reseat" the file object, but that is a code-smell by itself. Moreover, it's a maintenance nightmare. How can you know that you get all the control-flow cases right? RAII's main point is precisely to not have to think about it. But the argument that should convince most C-programmers is that it can lead to substantial code-pessimization. In your case, it is not bad owing to C++17's copy-elision guarantees, but in more complex cases it can get really bad.

rfuchs:
> Let me guess: these were (ex-)C programmers?

No, not even (and no Java background either, which at least would have made sense 😆). IIRC the language in question was Perl, which has similar scope/destruct semantics as C++. I'm actually the C guy here, but that doesn't mean I can't take advantage of destructors when available (and I like to use gcc's auto cleanup stuff in my own C code where appropriate), which is why this policy struck me as so strange. But anyway, now I know for the future.

bstoeger:
off-topic: Perl (like Python) is different, because all objects are reference-counted. If an object goes out-of-scope, there might still be a reference to it, which prevents it from being destructed. Thus, your colleagues might have a point. In C++, objects going out of scope will be destructed, period. If a different part of the code owns and uses a dangling reference, the application will crash or at least misbehave. Fundamentally different philosophies. Perl is like using `shared_ptr` for every variable - a nightmare for lifetime management.

on-topic: @atdotde: You're listed as reviewer - any opinion?

atdotde:
My opinion? In perl, I try to always explicitly close files. For one, checking the return code of close tells you if the write was successful but admittedly this is more important for pipes than for actual files. Also, closing it explicitly makes sure all buffers are flushed. But scopes in perl tend to be longer, so it's more important there.

atdotde (commented):
I think it's good. But I second @dirkhh 's comment: Can it happen that the file is not open but exits on disk? Is leaving it the right decision (as it is not "ours")?

rfuchs:
> Can it happen that the file is not open but exits on disk? Is leaving it the right decision (as it is not "ours")?

We first use QTemporaryFile to create a new file with a new random name that is guaranteed not to have existed before. That file is then removed and the same file name is passed to zip_open() to actually create and write the file. There's a tiny window in which someone else could have created a file with the exact same name in between, in which case zip_open() would fail and report an error. The file is then closed (zip_close) and eventually the QFile re-opens it for reading and uploading. Again there's a tiny window in which the file we've just created could have been manipulated (or removed) by another process, but it would have to be done by the same user (or admin). So to answer your question, AFAICS the only way for something to go wrong here is through deliberate interference by the same user or admin.

atdotde:
So, then why test if it is open rather than test for existence?

rfuchs:
> So, then why test if it is open rather than test for existence?

It's open if we were just using it for the upload. I wouldn't want to remove a file that we weren't using.

atdotde:
"AFAICS the only way for something to go wrong here is through deliberate interference by the same user or admin." seems to imply to me, it's fine to remove it anyways. If somebody did crazy stuff, they should expect their file to be removed. If it ended up unintentionally (we created it but somehow it's no longer open) it should go, don't you think?

rfuchs:
> "AFAICS the only way for something to go wrong here is through deliberate interference by the same user or admin." seems to imply to me, it's fine to remove it anyways. If somebody did crazy stuff, they should expect their file to be removed. If it ended up unintentionally (we created it but somehow it's no longer open) it should go, don't you think?

No not really. First of all it's a situation that "should never" occur. So if that case ever happens, something must have already gone wrong somewhere along the way, and IMO the safer way out is not to do more harm and just leave things alone. Also since file names are random, I don't think the application should take ownership of any file that just happens to match the same file name pattern. But if the consensus is that subsurface should remove files even if it's not known that they're really ours, then I'll change it.

atdotde:
Fair enough

atdotde (approved):

dirkhh:
It feels illogical to me to test for 'open', but I do accept your 'if it isn't ours, don't delete it' statement.

dirkhh (approved):
let's merge this instead of discussing it to death. It's definitely a clear improvement.

bstoeger:
> It feels illogical to me to test for 'open', but I do accept your 'if it isn't ours, don't delete it' statement.

FWIW, it is kind of logical to me, as it reasonably handles two "shouldn't happen" situations (did not delete or try to delete twice). What I'm apprehensive of is that all this is non-reentrant and hard to follow. But that's not the fault of this PR.

What I find extremely illogical though is the `QFile` API. It confuses/mixes the concepts "file" and "path". Again and again, the Qt-team cannot get non-UI APIs right.

PR #3300, opened 2021-08-08
[WIP][FYI] Split profile code in interactive and non-interactive parts

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This large PR is purely preparatory code reshuffling for two things:
1) Redo the scaling code of the profile
2) Implement an interactive profile on mobile

It splits out the non-interactive parts of the profile. This should remain a `QGraphicsScene` for printing. The plan is to convert the interactive parts from `QGraphicsView` to the (mad) QtQuick framework. The separation is not yet 100% clear. Notably, the "dive handles" stay in the interactive part, but they may be moved to the non-interactive part, if they should be printed.

The current state is a "psychological milestone", because printing and mobile do no longer use the `QGraphicsView` and therefore `QWidget`.

Attention: this removes the (wild) DPR code for mobile and therefore will break the placement of the profile. Therefore, this must not be merged. This will be fixed with the scaling code.

### Changes made:

Lots of code reshuffling.

### Related issues:

Contains and supersedes #3235.


bstoeger:
@dirkhh: If I understand things correctly, there are only two vertical axes with labels: depth and partial-pressure. Do you happen to know why the latter is shown on the right side of the graph? That's quite unusual, no?

The oldest commit I find with this is: 4882336315d3ddf01579b94473ffafc62c9807

dirkhh:
That was so long ago... I think we thought that this would make it less confusing and visually better separated...?

Oh, and the decision is older than that - I think this was on the right even in the Gtk version :)

tcanabrava:
that's a huge milestone.
I wonder how that the QQuick will be used as all the original code I
created is based on QGraphicsView for the positioning of the scene.
still, good to see improvements there.

bstoeger:
FYI: I started removing the relative scaling code and replace by absolute scaling.

This completely destroys the layout and getting things right will be a huge pain, but the old code was just unmaintainable. I see no way of making the old code scale properly. Relative positioning simply does not work with wildly different screen sizes.

That said, can someone explain the "Scale graph" (`prefs.zoomed_plot`) feature to me? Does anybody really use that? It makes no sense to me. Frankly, I find the whole code in `get_maxtime()` dubious. The free space it leaves to the right always struck me as very ugly. I reckon this is to show some deco-information? But usually my dives don't have any off-gassing shown, so I'd prefer to only show this if there *is* something to be shown.

dirkhh:
The `zoomed_plot` is something that was added in the fairly early days in 2012. The idea was that short dives would be immediately visible as such. I had long forgotten about this - but I remember that people asked for this quite a bit.
Does anybody actively use it today (given how cleverly it is hidden with the little buttons)? Heck if I know.

I don't hate the feature, but to me it provides no value.

And no, this isn't about deco - this is about making sure that really short dives really look, err, short.
7acd075bd5ad1c6d42ba77b484aebc2a0572f710

bstoeger:
I see, that makes sense. I still don't get the
>   We also need to add 180 seconds at the end so the min/max plots correctly

and
>  Rounded up to one minute, with at least 2.5 minutes to spare.

parts. I feel that the profile should decide whether it wants to add free space. After all, it knows how big the scene is.

Also, why is that different for free dives?

(And a petty coding-style comment: variables shouldn't be all-caps.)

bstoeger:
With 3404235dfd530bd9f33392c8651e3cb88c8d9b91, things look much less unsteady to me.

Yes, now there is not enough space for the average-depth text-item, but that should be solved in the profile-scene, not by adding some random time to the end of the plot. Only the scene knows how much space this label takes.

dirkhh:
please let me know how I can help, etc

bstoeger:
Ouch. I tried to use Lubomir's SVG and this is the ridiculous result:
image
It's so small that you need a magnifying glass to see it, but still ... that wasn't the point of the exercise. I'll see if I can turn this into an SVG that Qt can render, or will make a large PNG. That said, there are numerous other icons that need higher resolution for printing.

> please let me know how I can help, etc

I think it is (way?) too early for styling or comments on styling. However, any comments/review of the code so far are highly welcome.

Could you for example confirm that with the new percentage plot, the desktop UI feels "snappier", whereas printing is ... slower? That's my paradox impression without having benchmarked the code.

Also, could you review this (old) snippet:
```
#ifndef SUBSURFACE_MOBILE
        int sz_bigger = metrics.sz_med + metrics.sz_small; // ex 40px
#else
#if defined(Q_OS_IOS)
         // on iOS devices we need to adjust for Device Pixel Ratio
        int sz_bigger = metrics.sz_med  * metrics.dpr;
#else
        // SUBSURFACE_MOBILE, seems a little big from the code,
        // but looks fine on device
        int sz_bigger = metrics.sz_big + metrics.sz_med;
#endif
#endif
```
That seems completely wrong to me in the context of the new code. Note how on IOS we now scale twice by `dpr`!? Ideally I would prefer to not have any of these, but simply scale up DPR for mobile if necessary.

bstoeger:
PS: It has always bothered me that the flag is positioned via its top-left position, when the bottom side of the pole should indicate the position. Do others feel the same?

dirkhh:
> PS: It has always bothered me that the flag is positioned via its top-left position, when the bottom side of the pole should indicate the position. Do others feel the same?

agreed

dirkhh:
All the conditional code is a result of 
- this is so complex, I'm not touching the underlying code
- ergo, I need to find a way to hack things so they look right on the different platform (because generally I'm the only one looking at that)
- I forgot how I hacked this last year, let's add a new hack

Specifically, the hard coded dpr / offsets nonsense was created by trial and error because I fundamentally couldn't understand why I got the results that I got. I couldn't make sense of why the profile ended up where it did.
My choices were "give up", "become smart", "make hackish changes so it looks mostly ok on most platforms".
I went with the latter.

neolit123:
> It's so small that you need a magnifying glass to see it, but still ... that wasn't the point of the exercise. I'll see if I can turn this into an SVG that Qt can render, or will make a large PNG.

not sure what the problem is. perhaps Qt uses some weird rendering rules for SVG.
the icon looks fine at any scale in a browser (Chrome) and SVG editing software.

> That said, there are numerous other icons that need higher resolution for printing.

i can help with that, but vectorizing raster icons is a slow process.
"tracing" (raster -> vector) software doesn't work so great in most cases.

bstoeger:
> not sure what the problem is. perhaps Qt uses some weird rendering rules for SVG.

The Qt renderer is broken and can't handle your layers. I had to manually intersect the polygons with Inkscape and now it works:
image

> i can help with that, but vectorizing raster icons is a slow process.

That would of course be great. But no hurry, this is a long way from being ready.

neolit123:
Can you share the updated SVG so that I can have a look?

I haven't used inkscape but looks like it's capable of creating icons.

bstoeger:
Sure, it's in the PR: [flag.zip](https://github.com/subsurface/subsurface/files/7083593/flag.zip)

It's not about Inkscape or CorelDraw, but the fact that the Qt-renderer is not able to clip layers. I'm sure that you can do the same with CorelDraw.

neolit123:
> It's not about Inkscape or CorelDraw, but the fact that the Qt-renderer is not able to clip layers. I'm sure that you can do the same with CorelDraw.

ok, i see what you mean. CorelDraw calls this concept of putting objects inside a shape / masking them a "PowerClip".
...and clipping seems like a missing feature of the Qt renderer.

bstoeger:
Sorry - very little progress here, work is absorbing all my time.

@dirkhh: I note that the partial-pressure gas axis is not localized, i.e. it is always given in bar even for non-metric. Is that an oversight, or just the way it is done?

bstoeger:
OK - according to a quick Google search it appears that it is really always done in bar and rationalized as in "1.4 bar O2 is like breathing 140% oxygen at surface". We use metric units, but we don't admit it.

dirkhh:
Correct - one of the few moments where even the crazy imperial-loving countries realize that trying to do this in psi would be utter garbage :)

bstoeger:
Ok, found some time to work on the axes. We will have to rethink the whole animation thing. I have some ideas.

bstoeger:
Because I don't have 99 other problems, I thought let's redo the zooming as well.

@dirkhh: Before I waste time on this, please confirm that this is the right way to go. For me, the old zoom was basically unusable.

dirkhh:
> Because I don't have 99 other problems, I thought let's redo the zooming as well.
> 
> @dirkhh: Before I waste time on this, please confirm that this is the right way to go. For me, the old zoom was basically unusable.

The existing zoom is utter and complete garbage

bstoeger:
@dirkhh: would you kindly check if you are OK with the new peak-search (commits 62b9f97 and 7e89ba0), notably with the behavior on zooming? The old code had a rigid 9-min window, which made the labels sparse on zooming. Thanks.

I didn't benchmark the new code vs. the old one, but I suspect that the time spent for peak-search is trivial compared to rendering the labels, etc.

dirkhh:
So many Subsurface things to do 
So little time.
I'll try to look into this PR today

profile-widget/diveprofileitem.cpp
```diff
@@ -161,37 +166,89 @@ void DiveProfileItem::replot(const dive *d, int from, int to, bool in_planner)
...
+		int peak;
+	};
+	std::vector<Peak> peaks;
+	peaks.reserve(max_peaks);
+	int highest_peak = std::max_element(data + from, data + to, comp_depth) - data;
+	if (data[highest_peak].depth < min_depth)
+		return;
+	peaks.push_back(Peak{ from, to, highest_peak });
+	for (int act = 0; act < (int)peaks.size(); ++act) {
+		plot_depth_sample(data[peaks[act].peak], Qt::AlignHCenter | Qt::AlignTop, getColor(SAMPLE_DEEP));
```
dirkhh: I am used to the fact that I can barely parse the C++ and have to keep re-reading things... but I'm pretty sure this one makes no sense. in line 183/184 we create a new vector `peaks`. In 188 we add one element. And now we have a `for` loop over "all the elements" in `peaks`.
What am I missing?

profile-widget/diveprofileitem.cpp
```diff
@@ -161,37 +166,89 @@ void DiveProfileItem::replot(const dive *d, int from, int to, bool in_planner)
...
+
+		// Clear to the left and right of peak and add new peaks if there is place
+		const Peak &act_peak = peaks[act];
+		const plot_data &act_sample = data[act_peak.peak];
+		int valley = act_peak.peak;
+
```
dirkhh, line 204: is the comment in 192 related to the following three lines of code?
because... I don't understand how this 'clears' anything...

profile-widget/diveprofileitem.cpp
```diff
@@ -161,37 +166,89 @@ void DiveProfileItem::replot(const dive *d, int from, int to, bool in_planner)
...
+				if (data[new_peak].depth < min_depth)
+					break;
+				peaks.push_back(Peak{ act_peak.range_from, new_to, new_peak });
+
+				plot_depth_sample(data[valley], Qt::AlignHCenter | Qt::AlignBottom, getColor(SAMPLE_SHALLOW));
+				break;
+			}
+			if (data[new_to].depth < data[valley].depth)
+				valley = new_to;
+		}
```
dirkhh, line 253: this code is so temptingly similar to what's above... but making it direction independent would completely break my brain, I'm sure...

dirkhh (commented):
not a review of the PR, just the last few commits

profile-widget/diveprofileitem.cpp
bstoeger: There are items added to the vector in the loop and thus `peaks.size()` increases. The code is correct as is, but admittedly a bit confusing.

The thing is that this used to be a tree, which degenerated to a deque once I noticed that we never have to go "up".

Since this can also be realized (and more efficiently) as a stack, I turned it into a stack. This is hopefully more clear now.

profile-widget/diveprofileitem.cpp
bstoeger: It's related to the rest of the loop. Replaced `clear` by `skip`.

profile-widget/diveprofileitem.cpp
bstoeger: Indeed, I also thought that having the same code twice is probably less confusing.

One could of course do this _extremely_ elegantly as a template taking iterators where one passes forward- and backward-iterators, respectively. But I have a feeling that you wouldn't be happy about the resulting piece of art.

bstoeger:
Have you had time to look at the result? Note: Not generally, but specifically the depth labels. Also with respect to mobile - the whole point is to make this scale better.

dirkhh:
> Have you had time to look at the result? Note: Not generally, but specifically the depth labels. Also with respect to mobile - the whole point is to make this scale better.

I am hip deep in the disaster that is BLE on Android...
I can of course stop doing that and play with this...
choices, choices

What I remember about the old code was that it was a hack to have something that looks ok. What I see from your algorithms this seems to make a LOT more sense. But I'll try to find time to build it and try it.

bstoeger:
No worries - I was only wondering if you had looked at the result.

I was basically trying to emulate the old behavior with adaptive interval sizes. Thus, I was hoping that the result is at least not much worse.

bstoeger:
Added a small change that omits printing "valleys" if less than 2 m deep.

dirkhh:
Spend a fair bit of time looking at the profile... by and large I think the labels on valleys and peaks work well. There is the occasional profile where I wish that we'd do one or two more, but then when I try to come up with rules how we'd pick them I realize that "well, it would look good here" is not a great criterion, and anything more specific that I try to come up with is clearly not generic. So I think this deserves a "well done".
The benchmarking code for profile drawing seems to indicate that the performance is a non-issue

Of course running this in a VM and showing something like 50 profiles in a row and switching back and forth may have exposed a bug... I usually build with ASAN in that VM and that showed a buffer overflow access in 

https://github.com/subsurface/subsurface/blob/ebbe8808fb950153d4073678aa539840e78fd138/profile-widget/diveprofileitem.cpp#L177

bstoeger:
Thanks for trying it.

The whole point is making the peak-search (mildly) configurable. To increase the number of peaks you could for example try
'''
const int half_interval = (vAxis.getMinLabelDistance(hAxis) + 1) / 2;
'''
(or similar).

The line you quote is obviously wrong. Fixed.

dirkhh:
I don't see the crash anymore. I played with the plot itself some more and have a few additional comments:
- there seems to be no margin at all on the right... I don't think that looks good
- when resizing the window, the profile gets redrawn, but not the x axis scale
Screen Shot 2021-10-21 at 7 53 35 AM
- using segments that don't hit multiples of 10 or 15 for minutes seems... odd (I would have liked 5/10/15 much better here)
Screen Shot 2021-10-21 at 7 54 19 AM

bstoeger:
> there seems to be no margin at all on the right... I don't think that looks good

The code now accounts for the mean-depth label, which effectively gives a margin (one can't turn the thing off?).

> when resizing the window, the profile gets redrawn, but not the x axis scale

I can't remember actively fixing this, but it does not seem to happen for me anymore.

> using segments that don't hit multiples of 10 or 15 for minutes seems... odd (I would have liked 5/10/15 much better here)

This now uses fractions of an hour/minute to choose the interval.

This is long way from being finished (e.g. the animation feature has to be completely redone - it's simply not compatible with the new scaling of the axes). Yet I wonder if this should be merged soon to 1) get at least some exposure and 2) avoid extreme divergence of the branches. It appears that @atdotde is working on printing so that might end up in very complicated merges if we deviate too much.

This does already bring some user-visible improvements: zooming is (arguably?) better and printing seems to not clip the labels anymore.

dirkhh:
My thinking is that I need to get the fingerprint stuff out in a release because it's so frigging annoying right now for a small subset of our users.
Should I do a 5.0.5 really soon based on current master and we'd merge this next week? Assuming I can find the time to do a release this weekend which I'm not entirely positive about with the mess with the email server and the fact that I have a Taekwondo test this afternoon...

bstoeger:
Sounds good.

dirkhh:
> Sounds good.

not sure if this changes your assessment, but I am now certain that I won't manage a release today (too much going on in real life) and I'm going on yet another business trip on Monday. Which means it will likely be late next week before I can do a release. Sorry.

dirkhh:
OK - I'm merging all the other pending things that seem ready.
Next I'll look through the WebEngine PR in more detail, and then I'll go through this one again start to finish.
But my hope is to get this one merged in the next day or two.

dirkhh:
there's some super simplistic benchmarking code in place in `ProfileWidget2::plotDive()` that is used when you run in verbose mode. Those numbers aren't super stable and reproducible, but they give you an idea. Your code creates no visual issues that I was able to find and seems to be in almost all cases slightly to significantly faster than latest master. Using real dive profiles here's a comparison:
```
312 master took 14 ms  profile7 took 3 ms 
303 master took 74 ms  profile7 took 30 ms 
302 master took 45 ms  profile7 took 26 ms 
301 master took 53 ms  profile7 took 36 ms 
300 master took 47 ms  profile7 took 35 ms 
299 master took 33 ms  profile7 took 25 ms 
298 master took 31 ms  profile7 took 26 ms 
297 master took 39 ms  profile7 took 33 ms 
296 master took 38 ms  profile7 took 31 ms 
295 master took 26 ms  profile7 took 20 ms 
294 master took 25 ms  profile7 took 21 ms 
293 master took 35 ms  profile7 took 24 ms 
292 master took 4 ms  profile7 took 6 ms 
291 master took 34 ms  profile7 took 23 ms 
290 master took 33 ms  profile7 took 24 ms 
289 master took 34 ms  profile7 took 21 ms 
288 master took 37 ms  profile7 took 24 ms 
287 master took 39 ms  profile7 took 24 ms 
287 master took 39 ms  profile7 took 18 ms 
```
292 is interesting - it's basically an empty profile (0 minute dive)
The rest seems to be in the 30-50% faster range. So **well done**!!

profile-widget/profilewidget2.cpp
```diff
@@ -1095,6 +1095,38 @@ void ProfileWidget2::setProfileState(const dive *dIn, int dcIn)
...
+	bool setpointflag = currentdc && currentdc->divemode == CCR && prefs.pp_graphs.po2;
+	bool sensorflag = setpointflag && prefs.show_ccr_sensors;
+	o2SetpointGasItem->setVisible(setpointflag && prefs.show_ccr_setpoint);
+	ccrsensor1GasItem->setVisible(sensorflag);
+	ccrsensor2GasItem->setVisible(currentdc && sensorflag && currentdc->no_o2sensors > 1);
+	ccrsensor3GasItem->setVisible(currentdc && sensorflag && currentdc->no_o2sensors > 2);
+	ocpo2GasItem->setVisible(currentdc && currentdc->divemode == PSCR && prefs.show_scr_ocpo2);
+
+	heartBeatItem->setVisible(prefs.hrgraph);
+#endif
```
dirkhh: I realize that you are just moving code... but having just one `#ifdef` block would be easier to read, don't you think?

Also (and completely unrelated) there's a typo in the commit message of **profile: factor out updating of visibility from setProfileState()** "factor out" instead of "factor our"

profile-widget/profilewidget2.cpp
```diff
@@ -1119,7 +1119,7 @@ void ProfileWidget2::updateVisibility()
 	decoModelParameters->setVisible(prefs.calcceiling);
 #ifndef SUBSURFACE_MOBILE
 	for (DiveCalculatedTissue *tissue: allTissues)
-		tissue->setVisible(prefs.calcalltissues);
+		tissue->setVisible(prefs.calcalltissues && prefs.calcceiling);
```
dirkhh: the commit message in **profile: call DiveCalculatedTissue::setVisible() correctly.** confuses me when I look at this change. When you say "pass the correct parameter in the first place"  - is there "another place" where the call to `setVisible()` should be removed entirely because it's no longer needed?

Also, don't have a period at the end of a commit message title.

backend-shared/exportfuncs.cpp
```diff
@@ -30,6 +32,28 @@ bool ExportCallback::canceled() const
 }
 
 #if !defined(SUBSURFACE_MOBILE)
+
+#include "desktop-widgets/mainwindow.h" // Currently needed for profile printing. TODO: remove.
+
+static void exportProfile(const struct dive *dive, const QString filename)
+{
+	ProfileWidget2 *profile = MainWindow::instance()->graphics;
```
dirkhh: the last two commits created separate profiles for printing purposes. Why is this one reusing the `MainWindow::instance()->graphics` ?
Or am I just not reading far enough ahead and this is changed in a later commit?

backend-shared/exportfuncs.cpp
```diff
@@ -30,6 +32,28 @@ bool ExportCallback::canceled() const
...
+	profile->plotDive(dive, 0, false, true);
+	QImage image = QImage(profile->size() * 4, QImage::Format_RGB32);
+	QPainter paint;
+	paint.begin(&image);
+	profile->render(&paint);
+	image.save(filename);
+	profile->setToolTipVisibile(true);
+	profile->setFontPrintScale(scale);
+	profile->setPrintMode(false);
+	profile->plotDive(dive, 0, true); // TODO: Shouldn't this plot the current dive?
```
dirkhh: well, it plots what you pass in, right?

backend-shared/exportfuncs.cpp
```diff
@@ -35,23 +35,33 @@ bool ExportCallback::canceled() const
...
+
+static void resetProfile()
+{
+	ProfileWidget2 *profile = MainWindow::instance()->graphics;
 	profile->setToolTipVisibile(true);
-	profile->setFontPrintScale(scale);
+	profile->setFontPrintScale(1.0);
 	profile->setPrintMode(false);
-	profile->plotDive(dive, 0, true); // TODO: Shouldn't this plot the current dive?
+	profile->plotDive(current_dive, 0, true);
```
dirkhh: yeah, here that makes sense, I think

backend-shared/exportfuncs.cpp
```diff
@@ -35,23 +35,33 @@ bool ExportCallback::canceled() const
...
 	QImage image = QImage(profile->size() * 4, QImage::Format_RGB32);
 	QPainter paint;
 	paint.begin(&image);
 	profile->render(&paint);
 	image.save(filename);
+}
+
+static ProfileWidget2 *getPrintProfile()
+{
+	ProfileWidget2 *profile = MainWindow::instance()->graphics;
```
dirkhh: that's still just the one in the `MainWindow`. Here's a crazy thing. If you execute the main event loop during printing (to make sure that the progress bar actually moves forward), you can make see all the profiles rendered into the visible profile window as you print. I had assumed that you would also take a 'private' profile widget here...

Maybe I'm confused?

backend-shared/exportfuncs.cpp
```diff
@@ -37,33 +38,31 @@ bool ExportCallback::canceled() const
...
-	profile->setToolTipVisibile(false);
+	// Let's say that 800x600 is a "reasonable" profile size. Use four times that for printing.
+	const int scale = 4;
+	QSize size(800 * scale, 600 * scale);
+
+	// TODO: Annoyingly, this still needs a parent window? Otherwise,
+	// the profile is shown as its own window, when calling show() below.
+	auto profile = std::make_unique<ProfileWidget2>(nullptr, MainWindow::instance());
+	profile->resize(size);
+	profile->show();	// Ominous: if the scene isn't shown, parts of the plot are missing. Needs investigation.
```
dirkhh: ahhhhhh - ok, so the previous few comments were all just premature. I'll leave them because it's painful to go back when doing the reviews, but now this all makes sense. THANKS

backend-shared/exportfuncs.cpp
```diff
@@ -37,33 +38,31 @@ bool ExportCallback::canceled() const
...
 }
 
-static ProfileWidget2 *getPrintProfile()
+static std::unique_ptr<ProfileWidget2> getPrintProfile()
 {
-	ProfileWidget2 *profile = MainWindow::instance()->graphics;
-	profile->setToolTipVisibile(false);
+	// Let's say that 800x600 is a "reasonable" profile size. Use four times that for printing.
+	const int scale = 4;
+	QSize size(800 * scale, 600 * scale);
```
dirkhh: YES!!! that will solve one of my biggest issues with the current printing code.

dirkhh (commented):
reviewed the first dozen of these 116(!!!) commits.
I'll continue down the list whenever I have breaks, but this will take a bit.
Mostly these are small comments (and a few of them I resolve myself later, so read them all before responding).

profile-widget/profilewidget2.cpp
bstoeger: Yes the crazy `setVisible()`, which ignores the visible parameter, is removed in the next commit, a120b5a77e81be08b850ce85fa50fe9d3a2ea37c.

One could squash these commits.

backend-shared/exportfuncs.cpp
bstoeger: This is copy&paste of a comment introduced in 2789bb05b133a7cf54081d58d4f5c51c8977e951

I'm not sure what I was thinking at that time...?

In any case, as you note, the bogus comment will be removed later on.

profile-widget/profilewidget2.cpp
bstoeger: Yes, but note that these ifdefs will be removed later anyway.

Basically, I think the whole idea of mobile vs. desktop profiles is broken. This should rather depend on the size of the profile. For example on big tablets you might want to show all features. But that is for the future...

bstoeger:
Ok, here my attempt at the animations. This was easier than expected.

It makes more sense to me than the old code, because it doesn't change the value of existing ticks. Since the value of the ticks has to be saved, the code is somewhat more complex, though.

Disclaimer: To me animations are an unnecessary distracting feature. I prefer the chart to be plotted instantaneously without moving/wobbling/whatever. Therefore, I am the worst person to judge this. Please let me know if this is an improvement. To test, simply switch between dives that have different scales (waiting for the animation to finish). Thanks.

bstoeger:
*sigh*
When entering/exiting the planner, the profile is redrawn 5(sic!) times. This needs to be fixed.

bstoeger:
The last few commits are brutal - they remove the intermediate model-layer. There were two copies of plot_info floating around and this was just a pain to maintain.

I'm sure that this will break the planner, but I think ultimately it makes the significantly code easier to follow. The model thing makes no sense to me as an application-internal interface.

bstoeger:
> there's some super simplistic benchmarking code in place in ProfileWidget2::plotDive() that is used when you run in verbose mode. 

Thanks for testing this. However, these few tens of ms should not be perceptible by a human. So probably just imagination on my part.

dirkhh:
> The last few commits are brutal - they remove the intermediate model-layer. There were two copies of plot_info floating around and this was just a pain to maintain.
> 
> I'm sure that this will break the planner, but I think ultimately it makes the significantly code easier to follow. The model thing makes no sense to me as an application-internal interface.

I see you are having fun. I still need to continue the code review. ARG. I need more TIME
@atdotde  - will you test the planner with this branch? I haven't used the planner in so many years, I doubt I'd even notice if something broke...

desktop-widgets/mainwindow.cpp
```diff
@@ -53,6 +50,7 @@
 #include "desktop-widgets/statswidget.h"
 #include "commands/command.h"
 
+#include "profilewidget.h"
 #include "profile-widget/profilewidget2.h"
```
dirkhh, line 54: ok, that is a little odd `profilewidget.h` and `profile-widget/profilewidget2.h`
I understand the desire to go to `profileview` for the latter makes it less of an issue, but... meh - naming is hard

The changes in this commit are overall good. `graphics` always bugged me as name for the profile... and separating this code out makes sense. I tried to match the lines and this is one of those cases where I wish there was a better review UI - one that would clearly show you what changed in code that was moved...

backend-shared/exportfuncs.cpp
```diff
@@ -34,32 +34,27 @@ bool ExportCallback::canceled() const
 
 #if !defined(SUBSURFACE_MOBILE)
 
-#include "desktop-widgets/mainwindow.h" // Currently needed for profile printing. TODO: remove.
+// Let's say that 800x600 is a "reasonable" profile size. Use four times that for printing.
+static const constexpr int profileScale = 4;
+static const constexpr int profileWidth = 800 * profileScale;
+static const constexpr int profileHeight = 600 * profileScale;
```
dirkhh: I thought that `constexpr` implied `const`. Why do you need both here? Or asked another way, what's the difference between
`static constexpr int ...`
and
`static const constexpr int ...`

profile-widget/divetextitem.cpp
```diff
@@ -55,19 +49,6 @@ void DiveTextItem::setScale(double newscale)
...
 {
 	if (internalText != t) {
-		if (!connected) {
-			if (scene()) {
-				// by now we should be on a scene. grab the profile widget from it and setup our printScale
-				// and connect to the signal that makes sure we keep track if that changes
-				ProfileWidget2 *profile = qobject_cast<ProfileWidget2 *>(scene()->views().first());
-				connect(profile, SIGNAL(fontPrintScaleChanged(double)), this, SLOT(fontPrintScaleUpdate(double)), Qt::UniqueConnection);
-				fontPrintScaleUpdate(profile->getFontPrintScale());
-				connected = true;
```
dirkhh, line 65: so thrilled to see this confusing mess gone

desktop-widgets/profilewidget.cpp
```diff
@@ -8,6 +8,45 @@
 
 #include <QToolBar>
 #include <QHBoxLayout>
```
dirkhh, line 11: comment on the commit message of ** profile: implement "empty state" in desktop's ProfileWidget **
this is missing an empty line before the SOB

desktop-widgets/profilewidget.cpp
```diff
@@ -128,9 +174,33 @@ void ProfileWidget::setDive(const struct dive *d)
...
+		ui.profSAC->setDisabled(freeDiveMode);
+		ui.profTissues->setDisabled(freeDiveMode);
+
+		ui.profRuler->setDisabled(false);
+		ui.profScaled->setDisabled(false); // measuring and scaling
+		ui.profTogglePicture->setDisabled(false);
+		ui.profHR->setDisabled(false);
+		view->plotDive(current_dive, dc_number);
+	} else {
+		stack->setCurrentIndex(0);
```
dirkhh, line 206: that feels really clever!

profile-widget/profilescene.cpp
```diff
@@ -29,19 +42,63 @@ const static struct ProfileItemPos {
 	ProfileItemPos();
 } itemPos;
 
+template<typename T, class... Args>
```
dirkhh, line 49: another commit where I wish for a better review UI... I think this is the same code just moved around - but that's so painful to verify...

dirkhh (commented):
reviewed up to ** profile: set DiveCartesianAxis::fontLabelScale in constructor **

This removes the insane magic code that shifted things around depending on the `dpr` which I introduced to deal with 'modern' mobile displays. I don't understand why that code is no longer needed. What changes actually fixed the issue that I was trying to hack around with the magic shifting code?

I'm submitting this partial review because I'm not sure I'll manage to finish this review today and it's been already in process for more than a day and I worry that GitHub will lose my state / comments if @bstoeger force pushes updates into the branch...

bstoeger:
> This removes the insane magic code that shifted things around depending on the dpr which I introduced to deal with 'modern' mobile displays. I don't understand why that code is no longer needed. What changes actually fixed the issue that I was trying to hack around with the magic shifting code?

Replacing the relative scaling (percentage of the scene) by absolute scaling (so-and-so many pixels). The whole point is to remove that "insane magic code". Though I'm not sure if you already went over the corresponding commits or if they come later. I lost the overview.

backend-shared/exportfuncs.cpp
bstoeger: Yes `const` is weaker and redundant in this case. Removed.

dirkhh:
> > This removes the insane magic code that shifted things around depending on the dpr which I introduced to deal with 'modern' mobile displays. I don't understand why that code is no longer needed. What changes actually fixed the issue that I was trying to hack around with the magic shifting code?
> 
> Replacing the relative scaling (percentage of the scene) by absolute scaling (so-and-so many pixels). The whole point is to remove that "insane magic code". Though I'm not sure if you already went over the corresponding commits or if they come later. I lost the overview.

Yay!
I'm about 75 commits into the 127 commits in this PR...
yes, the absolute scaling / positioning code has been in that group
I will try my best to get through this whole PR today or tomorrow so we can get it merged - what a monster.

dirkhh (commented):
continued the review through ** profile: don't plot event-icons outside of range **
I think that means I have about 30 more commits to go

I should never have let this branch stay unreviewed for so long - this is making my head hurt. So many changes, so much code ripped out and replaced.

I didn't see anything that seemed worth complaining about - and that worries me... I'm not sure I'm still focused enough

dirkhh:
ok, while taking a break from reviewing code I did some more testing and found a small regression.
with the existing code, very short dives end when they end and then there's just nothing on the profile
Screen Shot 2021-12-03 at 3 27 26 PM

but with your branch, the profile extends on the surface
Screen Shot 2021-12-03 at 3 28 05 PM

The labels on the temperature line seem much better. The depth scale is one of those weird things about how people think about feet - where I'm not sure I'm representative. I think multiples of 30ft oddly useful - likely because I think of them as "almost multiples of 10m"... which might be a strong indication that you shouldn't listen to me. So while I first was going to write something about the 50/100 scale... I think you are actually right there.

bstoeger:
> I think multiples of 30ft oddly useful - likely because I think of them as "almost multiples of 10m"... which might be a strong indication that you shouldn't listen to me. So while I first was going to write something about the 50/100 scale... I think you are actually right there.

I feel that many people do that. Even in metric, 30 m is somewhat of a "magic" number. In any case, the depth axis you show is too sparsely populated and I think this was a bug introduced with the improvement of the time scale. Please check the force-pushed version.

We can easily allow for steps in 3s. But tell me:
- Only for >10 or also for single digits (3, 6, 9, ...)?
- Only for depth or also other axes?
- If only for depth: only for "imperial" (what a strange name) or also for metrx?

The code is in `divecartesianaxis.cpp/sensibleInterval()`. Feel free to experiment with it and force-push. I will not do any more development in this PR, lest we never finish.

Concerning the extension of the profile line for short dives - I will look at this later. My dive computer would not even register such a "dive". However, it should be trivial to generate such a dive for testing.

bstoeger:
Ok, manually adding dives broke. I don't think I will have time to look at this today.

bstoeger:
The line-extension should be fixed. That was a silly clipping bug. Please test.

dirkhh:
I wish GitHub didn't make it so hard comparing what you had yesterday with what's here today. Thankfully that's easier in the local tree :)

As for multiples of 3 on the axis - no, I don't think I want those in general. I believe metric divers sometimes end up thinking in multiples of 3m BECAUSE that's 10ft and that matters because so much of deco research was done by the US military (my understanding is that's why deco stops tend to be at 3/6/9/12m).

So I think my comment was mostly an observation, not a request to change direction there.

Yes, the clipping fix fixes short dive display.

I'm not sure what you mean by "manually adding dives is broken" - in my quick test that seemed to work, but maybe I'm not trying the thing that's broken?

bstoeger:
> I'm not sure what you mean by "manually adding dives is broken" - in my quick test that seemed to work, but maybe I'm not trying the thing that's broken?

More precisely, editing the depth/duration of a manually added dive will make the profile fail. The last commit, which I just pushed, fixes this. I'm currently compiling master to see if that problem already exists in master.

profile-widget/profilescene.cpp
```diff
@@ -138,40 +138,60 @@ void ProfileScene::clear()
...
-#endif
-	meanDepthItem->setVisible(prefs.show_average_depth);
-	reportedCeiling->setVisible(prefs.dcceiling);
-	tankItem->setVisible(prefs.tankbar);
+	const struct divecomputer *currentdc = get_dive_dc_const(d, dc);
+	if (!currentdc)
+		return;
+	bool ppGraphs = ppGraphsEnabled(currentdc, simplified);
+
+	if (simplified) {
```
dirkhh, line 217: I really like the idea of the 'simplified' profile - once again, a 12" iPad may want to show the whole thing, and this design makes this so much easier.

profile-widget/profilescene.cpp
```diff
@@ -350,8 +366,14 @@ void ProfileScene::plotDive(const struct dive *dIn, int dcIn, DivePlannerPointsM
 	create_plot_info_new(d, get_dive_dc_const(d, dc), &plotInfo, !calcMax, planner_ds);
 
 	bool hasHeartBeat = plotInfo.maxhr;
-	updateVisibility(hasHeartBeat);
-	updateAxes(hasHeartBeat);
+	// For mobile we might want to turn of some features that are normally shown.
+#ifdef SUBSURFACE_MOBILE
+	bool simplified = true;
```
dirkhh, line 447: ... except that this hardcodes it. Which is fine for now, of course, but should then turn into an option on mobile that takes screen space into account...

dirkhh:
> > I'm not sure what you mean by "manually adding dives is broken" - in my quick test that seemed to work, but maybe I'm not trying the thing that's broken?
> 
> More precisely, editing the depth/duration of a manually added dive will make the profile fail. The last commit, which I just pushed, fixes this. I'm currently compiling master to see if that problem already exists in master.

I have started a while ago to have two Subsurface directories on my system.
One called `subsurface-ref` that has whatever I'm comparing with - usually master, but sometimes the last release or something like that, and one called `subsurface` which is whatever I'm working on.
That makes it so much easier to go back and compare behaviors / visuals / etc

profile-widget/diveprofileitem.cpp
```diff
@@ -163,37 +168,93 @@ void DiveProfileItem::replot(const dive *d, int from, int to, bool in_planner)
...
+	// No point in searching peaks with less than three samples
+	if (to - from < 3)
+		return;
 
-		if ((entry == max_entry) && entry->depth / 100 != last) {
-			plot_depth_sample(entry, Qt::AlignHCenter | Qt::AlignBottom, getColor(SAMPLE_DEEP));
-			last = entry->depth / 100;
+	const int half_interval = vAxis.getMinLabelDistance(hAxis);
+	const int min_depth = 2000; // in mm
+	const int min_prominence = 2000; // in mm (should this adapt to depth range?)
```
dirkhh, line 181: hard to tell - and likely a matter of taste... for now I think this is a reasonable choice

bstoeger:
> As for multiples of 3 on the axis - no, I don't think I want those in general. I believe metric divers sometimes end up thinking in multiples of 3m BECAUSE that's 10ft and that matters because so much of deco research was done by the US military (my understanding is that's why deco stops tend to be at 3/6/9/12m).

Interesting - I didn't know that. For me, around 30 m is where the O2 alarm goes off, NDL dwindles alarmingly  and/or narcosis becomes precarious.

profile-widget/divecartesianaxis.cpp
```diff
@@ -106,8 +106,22 @@ int DiveCartesianAxis::getMinLabelDistance(const DiveCartesianAxis &timeAxis) co
 	return int(ceil(interval));
 }
 
-static double sensibleInterval(double inc, int decimals)
+static double sensibleInterval(double inc, int decimals, bool is_time_axis)
 {
+	if (is_time_axis && inc < 60.0) {
+		// for time axes and less than one hour increments, round to
+		// 1, 2, 3, 4, 5, 6  or 12 parts of an hour or a minute.
```
dirkhh: I'm intrigued by these choices... I would have expected 1,2,3,4,5,6,10,15,20 or something like that
oh, wait... those are how many parts we have? so this is actually 60, 30, 20, 15, 12, 10, 5?
If that's the case, then the explanation could be improved :)

dirkhh (commented):
you add commits faster than I can review :)
now up to ** profile: shorten DiveCartesianAxis::posAtValue(qreal value) **

Need to do some real life stuff but will be back, soon

bstoeger:
Removed last commit and posted as separate PR.

dirkhh:
> Removed last commit and posted as separate PR.

why? I mean, I expect to merge this PR "within hours" - or at least this weekend :)
But sure, I can merge that one separately, NBD

profile-widget/profilescene.cpp
```diff
@@ -16,9 +16,34 @@
...
 #include "qt-models/diveplannermodel.h"
+#include <QAbstractAnimation>
 
 static const double diveComputerTextBorder = 1.0;
 
+// Class for animations (if any). Might want to do our own.
+class ProfileAnimation : public QAbstractAnimation {
+	ProfileScene &scene;
+	// For historical reasons, speed is actually the duration
+	// (i.e. the reciprocal of speed). Ouch, that hurts.
```
dirkhh, line 26: so why aren't we renaming that?

profile-widget/divecartesianaxis.cpp
```diff
@@ -208,100 +201,142 @@ void DiveCartesianAxis::updateTicks(int animSpeed)
...
+			updateLabel(*actOld, 0.0, pos);
+			newLabels.push_back(std::move(*actOld));
+		}
 
-		if (position == Position::Bottom) {
-			// Fix size in case the scene changed
-			QLineF old = lines[i]->line();
-			lines[i]->setLine(old.x1(), old.y1(), old.x1(), old.y1() + rect.height());
-			Animations::moveTo(lines[i], animSpeed, childPos, rect.top());
+		double pos = ((position == Position::Bottom) != inverted) ?
```
dirkhh, line 298: right... right... that's totally... obvious...
my head hurts

profile-widget/divecartesianaxis.cpp
```diff
@@ -208,100 +201,142 @@ void DiveCartesianAxis::updateTicks(int animSpeed)
...
-		lines.push_back(line);
-		if (position == Position::Bottom) {
-			line->setLine(0.0, 0.0, 0.0, rect.height());
-			line->setPos(scene.sceneRect().width() + 10, rect.top()); // position it outside of the scene);
-			Animations::moveTo(line, animSpeed, childPos, rect.top());
-		} else {
-			line->setLine(0.0, 0.0, rect.width(), 0.0);
-			line->setPos(rect.left(), scene.sceneRect().height() + 10);
-			Animations::moveTo(line, animSpeed, rect.left(), childPos);
+	// If there are any labels left, mark them for deletion.
```
dirkhh, line 317: it took me a while to figure out... could you add "mark them for deletion by setting opacityEnd to zero" or something like that? Because I kept staring at this code, trying to figure out what was going on, having not read the comments below about how the Label class works...

profile-widget/divecartesianaxis.cpp
bstoeger: Added the corresponding reciprocals.

profile-widget/divecartesianaxis.cpp
```diff
@@ -209,100 +202,142 @@ void DiveCartesianAxis::updateTicks(int animSpeed)
...
+static QLineF operator*(double f, const QLineF &l)
+{
+	return QLineF(f*l.p1(), f*l.p2());
+}
+
+// Helper template: get point in interval (0.0: start, 1.0: end)
+template <typename T>
+T mid(const T &start, const T &end, double fraction)
+{
+	return start + fraction*(end - start);
```
dirkhh: coding style... space around `*`

dirkhh (commented):
i really like the new animation code...
since the diff changed while I was reviewing it, posting this review update now :/

dirkhh:
BTW: please don't take the last comment to mean that I want you to stop working in this branch
I just wanted to explain why I posted another review of just one additional commit
The GitHub UI sometime seems to lose things if I keep a review alive to long - so I wanted to err on the side of caution.

profile-widget/diveprofileitem.cpp
```diff
@@ -46,17 +46,17 @@ void AbstractProfilePolygonItem::clipStop(double &x, double &y, double prev_x, d
 
 std::pair<double, double> AbstractProfilePolygonItem::getPoint(int i) const
 {
-	double x = dataModel.index(i, hDataColumn).data().toReal();
```
dirkhh: I'd love to go back in time to figure out why we ever thought that was a good idea... were we thinking to have a different x axis? Plotting time over depth? Or depth of over partial pressures or...?
I have no idea

profile-widget/diveprofileitem.h
```diff
@@ -20,15 +18,10 @@
...
 class DiveCartesianAxis;
-class QAbstractTableModel;
 struct plot_data;
 struct dive;
 
-class AbstractProfilePolygonItem : public QObject, public QGraphicsPolygonItem {
-	Q_OBJECT
-	Q_PROPERTY(QPointF pos WRITE setPos READ pos)
-	Q_PROPERTY(qreal x WRITE setX READ x)
-	Q_PROPERTY(qreal y WRITE setY READ y)
```
dirkhh, line 36: I think we used this from QML at some point... but I can't remember for what - and I don't thing we still do? At least I couldn't find this via a few careful `git grep` commands... the sad thing is that we don't get build time warnings if we break this...

profile-widget/profilescene.cpp
```diff
@@ -45,23 +45,42 @@ class ProfileAnimation : public QAbstractAnimation {
...
 
+template <int IDX>
+double accessTissue(const plot_data &item)
+{
+	return item.ceilings[IDX];
+}
+
+// For now, the accessor functions for the profile data do not possess a payload.
+// To generate the 16 tissue (ceiling) accessor functions, use iterative templates.
+// Thanks to C++17's constexpr if, this is actually easy to read and follow.
```
dirkhh, line 75: "_pleasantly easy to follow..._"
Right.
I have absolutely not the faintest clue what this code does.
If you ever decide that you are done with Subsurface and move on, I have no idea who will maintain this code, because it certainly ain't me.
That's not to say "don't do that". It's to say "don't stop working on Subsurface, like, ever"...

dirkhh (commented):
another few commits

profile-widget/diveprofileitem.h
bstoeger: This was needed for the `QPropertyAnimation`. Since the animation is now routed via a central point, it is not needed anymore.

dirkhh (approved):
OK, made it all the way through.
I posted a few more comments and questions... but fundamentally I think we'll be much better off with this merged.
This is such a massive improvement over the existing code. Yes, there are parts in here that I struggle to follow syntacticly which bothers me. But let's consider those a learning opportunity for me if nothing else.
But the new code is significantly cleaner, more logical, better structured, and faster on top of all that.

profile-widget/divecartesianaxis.cpp
bstoeger: Hmmmm... It would probably be a good idea to acually delete these objects at the end of the animation. I'll think about that.

PS: I'll be travelling (to France) tomorrow (if they let me in), so I'll probably be less responsive the upcoming week.

dirkhh:
> PS: I'll be travelling (to France) tomorrow (if they let me in), so I'll probably be less responsive the upcoming week.

I think that makes an even stronger point for trying to get this merged today.
Are you going to rebase this on top of master or do you want me to do that?
Are you planning another force push to address the last couple of things that I have noted?

profile-widget/profilescene.cpp
bstoeger: > I have absolutely not the faintest clue what this code does.

Actually, I'm confident that you will figure it out in no time if you ever have to. `if constexpr (...)` is executed in the compiler and this implements a simple loop counting from 0 to 15 to generate the 16 accessor functions.

It's well possible that we'll replace this with a more dynamic approach (e.g. virtual functions) if it turns out that this is too restrictive.

dirkhh:
Oh hey - this rebases completely cleanly... 
apologies for assuming otherwise :)

bstoeger:
> Are you planning another force push to address the last couple of things that I have noted?

Yes, but only later today (a few hours from now). I can rebase, no problem.

dirkhh:
> > Are you planning another force push to address the last couple of things that I have noted?
> 
> Yes, but only later today (a few hours from now). I can rebase, no problem.

rebase is non-issue. Happy to wait until you have time to finalize this. I'll do more testing in the meantime.

dirkhh:
I think these are small optimizations and I don't believe these are required before merging, but since I was playing with the code...
Here's a profile with no deco information and a heat map.
Screen Shot 2021-12-04 at 11 14 31 AM
I'd say that there needs to be space above the profile (at least enough to see the '0' for surface) and between temperature and heat map (so that I can see the lowest temp as a line).
There's something weird happening with the very last grid line on the right.
That last one is something I'm less sure about... but is there a little too much empty space as far as the depth is concerned? It feels like it would be better if the depth scale was stretched to only go to 60ft? Or maybe even 50 as the last horizontal grid line / vertical tick mark?

profile-widget/divecartesianaxis.cpp
bstoeger: I thought you would like that gnarly C-style code. ;-)

Seriously: Yes, it' really ugly, but for some reason
1) C and therefore C++ doesn't have an XOR operator.
2) Some engineer decided to make the y-coordinate for screens in opposite direction to what humanity is using since .. Descartes(??).

profile-widget/divecartesianaxis.cpp
bstoeger: Added code to actually delete the labels at the end of the animation. And reworded comment. Please check.

bstoeger:
> I think these are small optimizations and I don't believe these are required before merging, but since I was playing with the code...

There is certainly still tons of fine-tuning needed!

> I'd say that there needs to be space above the profile (at least enough to see the '0' for surface)

The old code didn't display the '0'. I reimplemented that as a special case (yes, ugly).

> between temperature and heat map (so that I can see the lowest temp as a line).

Yes, that is a known problem. I will try to implement configurable space between the chart features, but so far didn't have time.

> There's something weird happening with the very last grid line on the right.

That is the line of the cylinder axis. An artifact from the old code - A further necessary cleanup. For now, the line is omitted for axes without labels.

> That last one is something I'm less sure about... but is there a little too much empty space as far as the depth is concerned? 

Yes, because the core code still was adding space, despite the profile doing so itself. I tried to remove that. Please test.

profile-widget/divecartesianaxis.cpp
bstoeger: Done.

profile-widget/divecartesianaxis.cpp
dirkhh: because CRTs scan top down and not bottom up because of... I actually don't know

but yes, the code is fine of course. It's just not immediately intuitive.
I need to stop commenting on things that I actually agree with :)

bstoeger:
> but is there a little too much empty space as far as the depth is concerned?

Do you have "zoomed plot" on? I thought for not-zoomed we show at least 30 m, but 70 ft is less than that. Now I am confused...

dirkhh:
Not a lot of very exhaustive testing since your last force push, but the diff looked reasonable and the changes to the vertical space below profiles look great to me.
I'd be willing to merge this and ask people for testing and feedback...

dirkhh:
> > but is there a little too much empty space as far as the depth is concerned?
> 
> Do you have "zoomed plot" on? I thought for not-zoomed we show at least 30 m, but 70 ft is less than that. Now I am confused...

yes, I try to test with a number of different options, so I flip through multiple dives and then toggle as many options as I have time for (one by one) and try again... so the screen shot was with zoom on.

profile-widget/divecartesianaxis.cpp
dirkhh: yes, the code and comments make more sense now. Thanks

atdotde:
OK, I had a few minutes, so I built this branch and had a look if I could break things in the planner. Turns out, that wasn't too hard: Add a few waypoints by double clicking in the profile and then start moving those around with the mouse. While dragging strange texts about gas usage (I guess) appear and at some point, the handles lose their relation with the depth profile. Furthermore, it is far too easy and fast to scroll down to depths like 500m
Screenshot 2021-12-04 at 22 42 41
.

dirkhh:
those look like labels of tank pressure lines?

and yes, I agree about the speed at which things scale in either direction. I think what's happening is that the new profile code is faster, so the pre-existing bug of things expanding to quickly in the graphical profile editor get exaggerated. But that isn't a _new bug_ per se. It's just making an old bug that has annoyed me for ever simply far easier to trigger (and more severe)

atdotde:
Yes, but those appear only while dragging the handle (wasn't even easy to screenshot those). What I consider much more problematic is that the depth line does no longe go through the handle markers.

bstoeger:
Oy vey. I'll look into it, but not today.

dirkhh:
the big question... leave the branch hanging - or fix this after merging?

dirkhh:
I guess we settled on the default action of "leave it hanging"... admittedly, this week is crazy at work, so that's ok by me...

bstoeger:
I'm sure that the planner UI is broken - I didn't pay any attention to it while working on this PR.

However, in quick tests I couldn't reproduce the inconsistent state where the profile doesn't follow the handles.

I did however note that very funky things happen if a handle with a gas change event is moved past other handles. Ouch. Teaches me to never touch anything planner related - whatever I do, I just can't not break it.

Note: I rebased on master so that we have a consistent state.

bstoeger:
Also no weird texts for me:

image

Not sure how to proceed..???

dirkhh:
I believe Robert said that he saw the weird texts *while moving* handles, not "at rest".
I think we should merge what we have and fix what is broken once it landed in master.
It's too much pain to keep such a huge branch separate

bstoeger:
Hmmm,... I can now reproduce a strange visual artifact: A line that follows the mouse pointer (even when not dragging a handle). Very ominous. But today, I will not be able to look at this.

bstoeger:
Oh ... I just realize what that is: the cross-hair. Completely forgot about that. Will try to fix tomorrow. :-P Perhaps still wait for that...

bstoeger:
Please check the last commit, which attempts to fix the cross hairs. Not sure if the vertical line should expand to the bottom of the chart or only the bottom of the depth profile...?

I think I can now reproduce the not-updating of the profile in the planner: It happens when one has (slightly) zoomed in. I see three ways to fix this:
- completely disable zoom in the planner.
- disable adding/moving handles when zoomed in.
- enable update when zoomed in.

I still don't see the spurious pressure readings when changing. Perhaps a chart feature that I haven't turned on?

bstoeger:
Ouch - found another bug: When zooming in, time-ticks of the form "15:00" are generated. When zooming out, these may be retained by the animation code. Then you have a mix of "10" and "15:00" style ticks.

atdotde:
Hi Berthold,

> On 10. Dec 2021, at 12:27, bstoeger ***@***.*** ***@***.***>> wrote:
> 
> Please check the last commit that attempts to fix the cross hairs. Not sure if the vertical line should expand to the bottom of the chart or only the bottom of the depth profile...?
> 
> I think I can now reproduce the not-updating of the profile in the planner: It happens when one has (slightly) zoomed in. I see three ways to fix this:
> 
> completely disable zoom in the planner.
> disable adding/moving handles when zoomed in.
> enable update when zoomed in.
> I still don't see the spurious pressure readings when changing. Perhaps a chart feature that I haven't turned on?
> 
> 


I managed to pin down what I need to do to make the handles unattached (and this might well be an old bug and not created by your code): Turn on „Drop to first depth“ and then move the first handle to a „too early“ time that cannot be reached from the surface with the prescribed descent velocity.

The problem with the pressure readings remains. This seems to come about from (while dragging) only the ascent parts having pressure information (and thus a pressure line) while the stops don’t. For whatever reason. The line is continuous once I release the handle.

Furthermore, when dragging the handles, I can make Subsurface crash. Stacktrace shows this is in diveeventitem.cpp line 207 when dc is NULL but it tries to evaluate dc->sample[0]. The function is called in profile scene.cpp  lien 447 so, it seems current_dc happens to be NULL there.

I uploaded a screencast to https://neu.atdotde.de/~robert/plannerscreen.mov <https://neu.atdotde.de/~robert/plannerscreen.mov>

Best
Robert

bstoeger:
> Furthermore, when dragging the handles, I can make Subsurface crash. Stacktrace shows this is in diveeventitem.cpp line 207 when dc is NULL but it tries to evaluate dc->sample[0]. The function is called in profile scene.cpp  lien 447 so, it seems current_dc happens to be NULL there.

Argh. A silly `currentdc` vs `current_dc` mixup. Yes, these names are not optimal. This should be fixed now, but it needed some major rebasing - I hope I didn't mess this up.

There are so few users of `current_dc` that I wonder whether we should remove this global symbol (or rather macro).

bstoeger:
Ok, I see the spurious pressure indications now - I had to set a starting pressure. Weird.

bstoeger:
This fixes the problem of the spurious pressure markings:
```
diff --git a/profile-widget/profilescene.cpp b/profile-widget/profilescene.cpp
index 4e6922e45..b87bbd03e 100644
--- a/profile-widget/profilescene.cpp
+++ b/profile-widget/profilescene.cpp
@@ -438,7 +438,7 @@ void ProfileScene::plotDive(const struct dive *dIn, int dcIn, DivePlannerPointsM
 	 * create_plot_info_new() automatically frees old plot data.
 	 */
 	if (!keepPlotInfo)
-		create_plot_info_new(d, get_dive_dc_const(d, dc), &plotInfo, !calcMax, planner_ds);
+		create_plot_info_new(d, get_dive_dc_const(d, dc), &plotInfo, false, planner_ds);
 
 	bool hasHeartBeat = plotInfo.maxhr;
 	// For mobile we might want to turn of some features that are normally shown.
```

However, I'm not sure that this is what we want to do. It does a full calculation of the plot information when dragging the handle.

The problem is that if we don't do that, the pressure data is garbage. That is also the case in the old code, but it isn't that obvious - there are only few artifacts if one looks closely.

Robert?

bstoeger:
@dirkhh: please check if the last commit fixes the missing space between temperature and the rest.

Feel free to play with the values and force push. I hope that it is self-explanatory.

dirkhh:
and... since we are merging things... I think this one is ready to be merged as well. The profile looked fine to me with the latest changes.

PR #3306, opened 2021-08-15
core: add downloaded GPS to existing dive site

dirkhh:
If we download a first dive computer and add a dive site to the dive (by
setting a location name for example), and then download from another
dive computer that provides us with GPS data, we should keep the
existing dive site information, but add the GPS data from the freshly
downloaded dive computer.

- [x] Functional change

This seems so obvious - if the Garmin gives us a GPS fix, keep an existing dive site, but if it doesn't have a GPS fix, add the one we just downloaded

### Release note:

done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- add GPS fix downloaded from a dive comuter to existing dive site
 - fix broken curser left/right shortcut for showing multiple dive computers
 - allow editing the profile for dives imported without samples (via CSV)
 - fix potential crash (and other errors) importing DL7 dives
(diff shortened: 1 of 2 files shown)


mturkia:
So simple and yet so useful. (Or would useful without travel restrictions.)

dirkhh:
Well, Linus and I are indeed in Hawaii on a dive trip. Which is why we are finding bugs everywhere. Nothing beats ACTUALLY using the software, I guess.
Then again, we are diving and after ten hours on a boat (or below one) I tend to be too tired in the evening to do anything useful...

PR #3309, opened 2021-08-16
Clean up divecomputer 'device' handling

torvalds:
We have this odd legacy notion of a divecomputer 'device', that was
originally just basically the libdivecomputer 'EVENT_DEVINFO' report
that was associated with each dive.  So it had firmware version,
deviceid, and serial number.

It had also gotten extended to do 'nickname' handling, and it was all
confusing, ugly and bad.  It was particularly bad because it wasn't
actually a 'per device' thing at all: due to the firmware field, a dive
computer that got a firmware update forced a new 'device'.

To make matters worse, the 'deviceid' was also almost random, because
we've calculated it a couple of different ways, and libdivecomputer
itself has changed how the legacy 32-bit 'serial number' is expressed.

Finally, because of all these issues, we didn't even try to make the
thing unique, so it really ended up being a random snapshot of the state
of the dive computer at the time of a dive, and sometimes we'd pick one,
and sometimes another, since they weren't really well-defined.

So get rid of all this confusion.

The new rules:

 - the actual random dive computer state at the time of a dive is kept
   in the dive data. So if you want to know the firmware version, it
   should be in the 'extra data'

 - the only serial number that matters is the string one in the extra
   data, because that's the one that actually matches what the dive
   computer reports, and isn't some random 32-bit integer with ambiguous
   formatting.

 - the 'device id' - the thing we match with (together with the model
   name, eg "Suunto EON Steel") is purely a hash of the real serial
   number.

   The device ID that libdivecomputer reports in EVENT_DEVINFO is
   ignored, as is the device ID we've saved in the XML or git files. If
   we have a serial number, the device ID will be uniquely associated
   with that serial number, and if we don't have one, the device ID will
   be zero (for 'match anything').

   So now 'deviceid' is literally just a shorthand for the serial number
   string, and the two are joined at the hip.

 - the 'device' managament is _only_ used to track devices that have
   serial numbers _and_ nicknames. So no more different device
   structures just because one had a nickname and the other didn't etc.

   Without a serial number, the device is 'anonymous' and fundamentally
   cannot be distinguished from other devices of the same model, so a
   nickname is meaningless. And without a nickname, there is no point in
   creating a device data structure, since all the data is in the dive
   itself and the device structure wouldn't add any value..

These rules mean that we no longer have ambiguous 'device' structures,
and we can never have duplicates that can confuse us.

This does mean that you can't give a nickname to a device that cannot be
uniquely identified with a serial number, but those are happily fairly
rare (and mostly older ones).  Dirk said he'd look at what it takes to
give more dive computers proper serial numbers, and I already did it for
the Garmin Descent family yesterday.

(Honesty in advertizing: right now you can't add a nickname to a dive
computer that doesn't already have one, because such a dive computer
will not have a device structure.  But that's a UI issue, and I'll sort
that out separately)

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


torvalds:
Updated branch with test changes so that maybe it even passes some of the tests that had random deviceid data in them

torvalds:
Updated branch again for Windows-only smtk import tool build fix

dirkhh:
This looks like fun.
- I don't think a lot of people use the nicknames
- I think the new concept based on serial number is the smarter solution - the old code that mixed the firmware into things was criminally stupid
- this creates a few problems, mainly that for people who used this code and the UI that came with it we took functionality away (even though I'm quite sure that this functionality has been badly broken a while ago)
- also, dive computers that don't have a Subsurface style textual serial number (and that's the majority of dive computers, but I hope I can reduce that number somewhat when I'm back at home and have access to all of my 'test' dive computers) can no longer get a nickname

What needs to get done:
- this needs to be hooked into the undo machinery - @bstoeger I am hoping that you'd lend a hand there
- this needs a UI to change the nickname - I'll work on that
- we need to push hard to get more dive computer backends to support the "Serial" text field
- it may need a tiny bit of testing - but seriously, there's no way we can really test this unless we find people who have used this in the past and are interested in the functionality... this was initially implemented for something that Linus and I used to do around 2012 or 13 - and that I haven't touched since. So I can kinda recreate part of the original idea. But, yeah, I really don't think I'm interested in this anymore as a feature. But in order to test it I will mess with it with some test data.

core/device.cpp
```diff
@@ -13,31 +13,33 @@ struct device_table device_table;
 bool device::operator==(const device &a) const
```
bstoeger, line 13: I think this comparison-operator is never used and can be removed.

dirkhh:
I have the undo/redo mostly working.
@bstoeger - when I edit the nickname, I simply call `replot()` in the profile code and that gets the new nickname shown.
But with undo/redo we don't update the profile - so if you undo such an edit, you have to switch dives / switch dive computers / do something else that triggers a replot.
Before I wildly hack this - is there an obvious way to hook that in that I'm missing?

Oh, and I'd love a review of those last few commits as I am always feeling out of my league when I touch your C++ code :)

bstoeger:
> Oh, and I'd love a review of those last few commits

From a very quick (<1 min) view, this looks reasonable.

> Before I wildly hack this - is there an obvious way to hook that in that I'm missing?

Something like
```
        connect(&diveListNotifier, &DiveListNotifier::deviceEdited, this, &ProfileWidget2::actionRequestedReplot);
```
in ProfileWidget2's constructor, next to the other connects? (Jeez, DiveListNotifier is such a silly name!)

dirkhh:
brilliant. yes, that's poorly named - but it works.

core/device.cpp
```diff
@@ -44,6 +44,21 @@ extern "C" const struct device *get_device_for_dc(const struct device_table *tab
 	return it != dcs.end() && same_device(*it, dev) ? &*it : NULL;
 }
 
+extern "C" const struct device *get_or_add_device_for_dc(struct device_table *table, const struct divecomputer *dc, int *idx)
```
bstoeger: I'm not strictly against this interface, but it would appear more reasonable to me to just return and index (-1 for not found) and then use `get_device(...)` by the caller if they need to.

Something like the completely untested:
```
extern "C" int get_or_add_device_for_dc(struct device_table *table, const struct divecomputer *dc)
{
        if (!dc->model || !dc->serial || !idx)
                return -1;

        device dev { dc->model, dc->serial };
        auto it = std::lower_bound(table->devices.begin(), table->devices.end(), dev);
        if (it != dcs.end() && same_device(*it, dev))
                return it - table->devices.begin();
        else
                return create_device_node(table, dc->model, dc->serial, "");
}
```

bstoeger:
Since there are no other users, you could also remove the `index` parameter of the signal.

core/device.cpp
dirkhh: yeah, I guess I could do that - it's only one caller so far, anyway.
I'll play with it.

dirkhh:
> Since there are no other users, you could also remove the `index` parameter of the signal.

you lost me - which signal?

dirkhh:
> > Since there are no other users, you could also remove the `index` parameter of the signal.
> 
> you lost me - which signal?

oh, you mean the `DeviceEdited(int)`? Yeah, we also should remove the other signals there as they aren't used anymore. I'll look into that as well.

bstoeger:
Suggestion (not compiled):
```
diff --git a/commands/command_device.cpp b/commands/command_device.cpp
index 4dd025347..90bd9376e 100644
--- a/commands/command_device.cpp
+++ b/commands/command_device.cpp
@@ -27,7 +27,7 @@ void EditDeviceNickname::redo()
 	if (!dev)
 		return;
 	std::swap(dev->nickName, nickname);
-	emit diveListNotifier.deviceEdited(index);
+	emit diveListNotifier.deviceEdited();
 }
 
 void EditDeviceNickname::undo()
diff --git a/commands/command_divelist.cpp b/commands/command_divelist.cpp
index baf81c1ac..f7e6b8f71 100644
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -553,10 +553,8 @@ void ImportDives::redoit()
 	divesAndSitesToRemove = std::move(divesAndSitesToRemoveNew);
 
 	// Add devices
-	for (const device &dev: devicesToAddAndRemove.devices) {
-		int idx = add_to_device_table(&device_table, &dev);
-		emit diveListNotifier.deviceAdded(idx);
-	}
+	for (const device &dev: devicesToAddAndRemove.devices)
+		add_to_device_table(&device_table, &dev);
 
 	// Add new filter presets
 	for (auto &it: filterPresetsToAdd) {
@@ -583,11 +581,8 @@ void ImportDives::undoit()
 	setSelection(selection, currentDive);
 
 	// Remove devices
-	for (const device &dev: devicesToAddAndRemove.devices) {
-		int idx = remove_device(&device_table, &dev);
-		if (idx >= 0)
-			emit diveListNotifier.deviceDeleted(idx);
-	}
+	for (const device &dev: devicesToAddAndRemove.devices)
+		remove_device(&device_table, &dev);
 
 	// Remove filter presets. Do this in reverse order.
 	for (auto it = filterPresetsToRemove.rbegin(); it != filterPresetsToRemove.rend(); ++it) {
diff --git a/core/subsurface-qt/divelistnotifier.h b/core/subsurface-qt/divelistnotifier.h
index 033458915..a778f7d91 100644
--- a/core/subsurface-qt/divelistnotifier.h
+++ b/core/subsurface-qt/divelistnotifier.h
@@ -132,9 +132,7 @@ signals:
 	void picturesAdded(dive *d, QVector<PictureObj> pics);
 
 	// Devices related signals
-	void deviceAdded(int index);
-	void deviceDeleted(int index);
-	void deviceEdited(int index);
+	void deviceEdited();
 
 	// Filter related signals
 	void filterPresetAdded(int index);
```

bstoeger (approved):
I only did a peephole review (and am rather tired), but this looks all reasonable to me.

I'm still somewhat confused by the non-deterministic behavior of the original code. I don't see how the binary search could have given unstable results.

Also, please consider the following patch, though I may just post it later:
```
diff --git a/core/device.cpp b/core/device.cpp
index fe3804501..be8d07efa 100644
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -10,12 +10,6 @@
 
 struct device_table device_table;
 
-bool device::operator==(const device &a) const
-{
-	return model == a.model &&
-	       serialNumber == a.serialNumber;
-}
-
 static bool same_device(const device &dev1, const device &dev2)
 {
 	return strcmp(dev1.model.c_str(), dev2.model.c_str()) == 0 &&
diff --git a/core/device.h b/core/device.h
index 04455a5a5..74fa83195 100644
--- a/core/device.h
+++ b/core/device.h
@@ -50,7 +50,6 @@ extern void free_device_table(struct device_table *devices);
 #include <string>
 #include <vector>
 struct device {
-	bool operator==(const device &a) const; // TODO: remove, once devices are integrated in the undo system
 	bool operator<(const device &a) const;
 	void showchanges(const std::string &n) const;
 	std::string model;
```

core/device.cpp
dirkhh: that was a GREAT suggestion. The resulting code is actually MUCH nicer. Well done.

dirkhh:
force pushed - I didn't include your commit to remove the equality operator, but I think I implemented all of the other suggested changes... I guess I didn't give you enough credit in the commit messages... :)

dirkhh:
now based on latest master - I think this is ready to merge. It's not perfect as it breaks a feature that I believe is extremely rarely used - but in so many ways that feature was already broken, so... meh

Thoughts?

dirkhh (approved):
seems ready to be merged

PR #3311, opened 2021-08-18
Planner: Correctly compute CNS and OTU for bailout segments

### LINKED ISSUE #3310: CNS and OTU calculations seems wrong in some cases (CCR + bailout exit)
### Describe the issue:

- [x] Bug

### Issue long description:
It seems that calculated values of CNS and OTU are too low in some cases, particularly in CCR dives with bailout deco on OC.

### Operating system:
macOS 10.14.6

### Subsurface version:
5.0.2

### Steps to reproduce:
1). Go to "Plan dive" mode
2). Set two gases: Air and EAN50
3). Set depth 40 meters with runtime 80 minutes
4). Set Dive mode: CCR, set checkbox "Bailout: Deco on OC"
5). Set GF 40/80

This is not a real plan, just an example.

### Current behavior:
CNS: 13%
OTU: 41

### Expected behavior:
Approximate values that seems to be correct:
CNS: 67%
OTU: 179
These are calculated using Android version of MultiDeco. I'm not telling MultiDeco is something should be trusted more than Subsurface, but CNS and OTU calculations are pretty close with them when planning pure OC or pure CCR dives.

### Additional information:

Also I've tried to calculate something on paper using this article https://www.borrett.id.au/downloads/cns_percentage_tracking_v1-1.pdf

```
Subsurface (5.0.2.0) dive plan created on 8/17/21
Runtime: 176min, Stop times: + 6:04 /m + 1:31 /min

  | depth | duration | runtime | gas
➘ | 40m | 2min | 2min | air (SP = 1.3bar CCR)
➙ | 40m | 78min | 80min |  
- | 40m | 1min | 81min | air Open circuit
- | 21m | 3min | 84min | EAN50
- | 18m | 3min | 87min |  
- | 15m | 6min | 93min |  
- | 12m | 10min | 103min |  
- | 9m | 14min | 117min |  
- | 6m | 19min | 136min |  
- | 3m | 37min | 173min |  
➚ | 0m | 3min | 176min |  
```

78 min on ppO2 1.3 is 0.56*78=43.68%. After 90 minutes it will be half of that, 21.84%. Of course in reality there is further increment in the ascent part that I haven't tried to calculate, but there is no way CNS could be 13% at the end of the dive at 176 minute.

Considering all above I believe current planner's calculations are not correct, probably they are not taking to the account the OC part of the dive. Please have a look.

atdotde:
For dives with mixed divemode, one needs to check sample.setpoint
to figure out if the segment is an OC segment and the po2 needs
to be computed from the gasmix and ambient pressure.

This fixes #3310

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -85,6 +85,35 @@ static int active_o2(const struct dive *dive, const struct divecomputer *dc, dur
 	return get_o2(gas);
 }
 
+// Do not call on first sample as it acccesses the previous sample
+static int get_sample_o2(const struct dive *dive, const struct divecomputer *dc, const struct sample *sample)
+{
+	int po2i, po2f, po2;
+	const struct sample *psample = sample - 1;
+	// Use sensor[0] if available
+	if ((dc->divemode == CCR || dc->divemode == PSCR) && sample->o2sensor[0].mbar) {
+		po2i = psample->o2sensor[0].mbar;
+		po2f = sample->o2sensor[0].mbar;	// then use data from the first o2 sensor
+		po2 = (po2f + po2i) / 2;
+	} else if (sample->setpoint.mbar > 0) {
+		po2 = MIN((int) sample->setpoint.mbar,
+			   depth_to_mbar(sample->depth.mm, dive));
+	} else {
+		double amb_presure = depth_to_bar(sample->depth.mm, dive);
+		double pamb_pressure = depth_to_bar(psample->depth.mm , dive);
+		if (dc->divemode == PSCR) {
+			po2i = pscr_o2(pamb_pressure, get_gasmix_at_time(dive, dc, psample->time));
+			po2f = pscr_o2(amb_presure, get_gasmix_at_time(dive, dc, sample->time));
+		} else {
+			int o2 = active_o2(dive, dc, psample->time);	// 	... calculate po2 from depth and FiO2.
+			po2i = lrint(o2 * pamb_pressure);	// (initial) po2 at start of segment
+			po2f = lrint(o2 * amb_presure);	// (final) po2 at end of segment
+		}
+		po2 = (po2i + po2f) / 2;
+	}
+	return po2;
+}
+
 /* Calculate OTU for a dive - this only takes the first divecomputer into account.
    Implement the protocol in Erik Baker's document "Oxygen Toxicity Calculations". This code
    implements a third-order continuous approximation of Baker's Eq. 2 and enables OTU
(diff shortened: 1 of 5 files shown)


core/divelist.c
```diff
@@ -85,6 +85,38 @@ static int active_o2(const struct dive *dive, const struct divecomputer *dc, dur
 	return get_o2(gas);
 }
 
+static int get_sample_o2(const struct dive *dive, const struct divecomputer *dc, const struct sample *sample)
+{
+	int po2i, po2f, po2;
+	bool trueo2 = false;
+	struct sample *psample = sample - 1;
```
dirkhh: could this ever be called with the very first sample? which would make this an illegal reference?

core/divelist.c
```diff
@@ -85,6 +85,38 @@ static int active_o2(const struct dive *dive, const struct divecomputer *dc, dur
...
+	if ((dc->divemode == CCR || dc->divemode == PSCR) && sample->o2sensor[0].mbar) {			// if dive computer has o2 sensor(s) (CCR & PSCR)
```
dirkhh: I know, stupid remark, but that comment is WAY OUT on the right... maybe better in the next line

core/divelist.c
```diff
@@ -85,6 +85,38 @@ static int active_o2(const struct dive *dive, const struct divecomputer *dc, dur
...
+		po2f = sample->o2sensor[0].mbar;	// then use data from the first o2 sensor
+		po2 = (po2f + po2i) / 2;
+		trueo2 = true;
+	}
+	if ((dc->divemode == CCR) && (sample->setpoint.mbar > 0) && (!trueo2)) {
+		po2 = MIN((int) sample->setpoint.mbar,
+			   depth_to_mbar(sample->depth.mm, dive));
+		trueo2 = true;
+	}
+	if (!trueo2) {
```
dirkhh: reading through these three if statements... why isn't this
```
if ((dc->divemode == CCR || dc->divemode == PSCR) && sample->o2sensor[0].mbar) {
...
} else if ((dc->divemode == CCR) && (sample->setpoint.mbar > 0) {
...
} else {
...
}
```
which means we wouldn't need the trueo2 variable and the whole flow would seem a bit more "normal"?

Ahh, I see, this is just existing code that you moved. The question remains :)

dirkhh (commented):
a few comments on the code - but I'll admit that the logic of calculating all this goes over my head...

core/divelist.c
atdotde: It could but it is not. Added comment to this effect.

core/divelist.c
atdotde: Moved to separate line

core/divelist.c
atdotde: Cleaned up the logic

dirkhh (approved):
thanks for the updates

Issue #3312, opened 2021-08-31
Corruption of Log File with Multiple Transmitters


jschripsema:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
When saving a log with a dive that is using multiple transmitters, in 5.0.3, the log becomes unreadable and generates a "failed to parse" error.

### Operating system:

macOS 11.5.2 English

Shearwater Perdix AI with Shearwater Transmitters

### Subsurface version:

5.0.3

Does NOT happen with 5.0.2

Official release

### Steps to reproduce:

1) Open log that involves multiple transmitters.
2) Save log.
3) Attempt to open log.
4) Exporting log and then importing does not resolve the issue.

### Current behavior:

Log file fails to parse.

### Expected behavior:

Opening log displays dives and not an error message.

### Additional information:

Example log file.
```xml
<divelog program='subsurface' version='3'>
<settings>
<divecomputerid model='Shearwater Perdix AI' deviceid='c5305f40' serial='300f79b1' firmware='71'/>
</settings>
<divesites>
<site uuid='55043254' name='Xxxxxxx Xxxxx' gps='47.348611 -122.324444'>
  <geo cat='2' origin='2' value='Xxxxxx Xxxxxx'/>
</site>
</divesites>
<dives>
<dive number='236' rating='4' visibility='5' sac='11.729 l/min' otu='6' cns='4%' tags='drysuit, salt, shore' divesiteid='55043254' date='2017-11-11' time='12:03:59' duration='51:30 min'>
  <buddy>Xxxxx Xxxxxxxxxx</buddy>
  <notes>Xxxx xxxx Xxxxx xx Xxxxxxx, x xxxxx xxxx xx xxx xxxxx xxxxx xxx XX, x xxxx xxxx-xxx, xxx x xxxxx xxx xxxxxxxx! Xxxxx xxxx xxxx xxx Xxxxxx XX. Xxxxx xxx xxxxxxxx xxxx xx xx, xxx xxxx'x xxxx xxxx. Xxxxxxxx xxxxxxx'x xxxx xxxx xxxx xx xxx xxx…</notes>
  <suit>XxxXxxx Xx 5000</suit>
  <cylinder size='12.087 l' workpressure='237.318 bar' description='HP100' />
  <weightsystem weight='2.3 kg' description='Dolphin Tech Steel Backplate' />
[...]

PR #3314, opened 2021-09-02
parser: XML_PARSE_RECOVER to xmlReadMemory()

bstoeger:
Due to changes in the handling of sensor-ids, invalid XMLs were
generated. In particular, these contained duplicate attributes
in the sample tags.

Even though these files shouldn't exist, let's try to parse
them anyway. Some data will be lost, but that's better than
not opening the file.

libxml2 can be told to try to recover from such petty(?) errors
by passing the XML_PARSE_RECOVER flag.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

At least try to parse broken XMLs. See discussion in #3312.
Makes me itch to replace the libxml2 by a custom xml parser. I have written one - it's not *that* bad and would give us much finer control.

### DIFF
--- a/core/parse-xml.c
+++ b/core/parse-xml.c
@@ -1745,9 +1745,9 @@ int parse_xml_buffer(const char *url, const char *buffer, int size,
 	state.sites = sites;
 	state.devices = devices;
 	state.filter_presets = filter_presets;
-	doc = xmlReadMemory(res, strlen(res), url, NULL, XML_PARSE_HUGE);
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
this seems to also be part of #3315
but it also seems obviously the right thing to do, so I'd merge it here why we wait for #3315

PR #3316, opened 2021-09-03
core: properly clear pressure data

bstoeger:
In 82f967ddb37022a69be2de3f38ac445093c0d6df sanitizing of sensor-ids
was introduced to prevent out-of-bound accesses. The sensor was
overwritten as -1 (NO_SENSOR).

However, this led to corrupted XML files, since the code sees this
as an "old format" indicator.

Other parts of the code simply take the pressure and sensor reading
from the previous sample. Do this here as well.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This hopefully prevents the XML file corruption described in #3312. Don't use -1 as no-sensor marker, as that's messing with the XML output. I don't understand the sensor-handling, sorry. This is simply copying a different part of the code.

Attention: completely untested!

### Release note:

Done.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1213,8 +1213,10 @@ static void fixup_dc_sample_sensors(struct divecomputer *dc, int nr_cylinders)
 	for (int i = 0; i < dc->samples; i++) {
 		struct sample *s = dc->sample + i;
 		for (int j = 0; j < MAX_SENSORS; j++) {
-			if (s->sensor[j] < 0 || s->sensor[j] >= nr_cylinders)
+			if (s->sensor[j] < 0 || s->sensor[j] >= nr_cylinders) {
 				s->sensor[j] = NO_SENSOR;
+				s->pressure[j].mbar = 0;
+			}
 		}
 	}
 }
--- a/core/save-git.c
+++ b/core/save-git.c
@@ -258,6 +258,9 @@ static void save_sample(struct membuffer *b, struct sample *sample, struct sampl
 		pressure_t p = sample->pressure[idx];
 		int sensor = sample->sensor[idx];
 
+		if (sensor == NO_SENSOR)
+			continue;
+
 		if (!p.mbar)
 			continue;
 
--- a/core/save-xml.c
+++ b/core/save-xml.c
@@ -256,6 +256,9 @@ static void save_sample(struct membuffer *b, struct sample *sample, struct sampl
 		pressure_t p = sample->pressure[idx];
 		int sensor = sample->sensor[idx];
 
+		if (sensor == NO_SENSOR)
+			continue;
+
 		if (!p.mbar)
 			continue;
 


dirkhh (approved):
that does seem like the right thing to do.

dirkhh:
Linus is in the middle of the merge window - I highly doubt that he'll look at this, but I'll ping him

dirkhh:
> Linus is in the middle of the merge window - I highly doubt that he'll look at this, but I'll ping him

his response:
> Yeah, not going to have time for that until later next week at the earliest. Remind me in a few days.

Would you like to wait, or move forward?

bstoeger:
I'd wait (you don't plan an emergency release, do you?). Your call.

dirkhh:
> I'd wait (you don't plan an emergency release, do you?). Your call.

Happy to wait 
This is such an unusual edge case..  no need for an emergency release, IMHO

core/dive.c
```diff
@@ -1208,13 +1208,26 @@ static void fixup_no_o2sensors(struct divecomputer *dc)
...
+/* Remove sensor and gas pressure info */
+static void clear_sensor(struct sample *samples, int sample_id, int sensor_id)
+{
+	struct sample *s = &samples[sample_id];
+	if (sample_id == 0) {
+		s->sensor[sensor_id] = 0;
+		s->pressure[sensor_id].mbar = 0;
+	} else {
+		s->sensor[sensor_id] = s[-1].sensor[sensor_id];
+		s->pressure[sensor_id].mbar = s[-1].pressure[sensor_id].mbar;
```
torvalds: I don't think this code is sensible.

Yeah, yeah, I realize it was copied from previous code, but that previous code wasn't really sensible either, and this change actually expands upon the bad behavior.

In particular, let's say that we have a sample with an invalid sensor.  The old code in fixup_dc_sample_sensors() would mark that as having NO_SENSOR, which admittedly didn't work so well, but at least it was somewhat sensible, and it didn't change the pressure.

The *new* code now overwrites both the invalid sensor data _and_ the pressure data. The latter might have been valid, just with the wrong cylinder index. Hmm?

core/dive.c
```diff
@@ -1208,13 +1208,26 @@ static void fixup_no_o2sensors(struct divecomputer *dc)
...
+}
+
 static void fixup_dc_sample_sensors(struct divecomputer *dc, int nr_cylinders)
 {
 	for (int i = 0; i < dc->samples; i++) {
 		struct sample *s = dc->sample + i;
 		for (int j = 0; j < MAX_SENSORS; j++) {
 			if (s->sensor[j] < 0 || s->sensor[j] >= nr_cylinders)
-				s->sensor[j] = NO_SENSOR;
+				clear_sensor(dc->sample, i, j);
```
torvalds: See, here the old code made sense. It fixes up the one thing it says it should fix up. It does so badly, but still - sensibly.

The new code says it clears sensor data, but that's not really what it does.

core/dive.c
```diff
@@ -1680,14 +1693,7 @@ static void sample_renumber(struct sample *s, int i, const int mapping[])
...
 		if (sensor == -1) {
-			// Remove sensor and gas pressure info
-			if (i == 0) {
-				s->sensor[j] = 0;
-				s->pressure[j].mbar = 0;
-			} else {
-				s->sensor[j] = s[-1].sensor[j];
-				s->pressure[j].mbar = s[-1].pressure[j].mbar;
-			}
+			clear_sensor(s, i, j);
```
torvalds: Ok, so here the new code did what the old code did. Which was always a bit odd, but at least it doesn't change behavior like above.

torvalds (commented):
I dunno. I don't like the change. It probably doesn't matter, but it smells like of the two places that were made to be consistent, the code picks the nonsensical one.

Why would we copy the previous sensor data if the new sensor data is bogus? I think it would be better to just clear it, and then we can interpolate when we show it, rather than use random previous state?

Alternative question: why would we throw out the pressure data if the sensor number is wrong. Maybe the pressure data is fine, we just got confused about where it came from. 

Do I know what the right solution is? No. And this may not matter. But the code doesn't smell sensible to me.

core/dive.c
bstoeger: > See, here the old code made sense. It fixes up the one thing it says it should fix up. It does so badly, but still - sensibly.

The first problem is - with this code, we produce broken XML files.

The second problem is - I don't understand this part of the code _at all_.

Personally, my preferred solution would be to remove writing "legacy format" sensor data. But I don't understand why it is written in the first place, so I don't dare to remove it.

Help?

bstoeger:
Perhaps 9209e6a2c04ea6305623e1b47d84be9da43c69a5 is the "real" fix...

dirkhh:
I think it makes no sense to write pressure data with NO_SENSOR - because what would we ever do with it later.
I haven't spent enough time to convince myself that this PR isn't throwing away data that we **could** reasonably keep.
So that will require a bit more careful reading on my part... but of course, this PR is gating a Subsurface 5.0.4 release :(

core/save-git.c
```diff
@@ -258,6 +258,9 @@ static void save_sample(struct membuffer *b, struct sample *sample, struct sampl
 		pressure_t p = sample->pressure[idx];
 		int sensor = sample->sensor[idx];
 
+		if (sensor == NO_SENSOR)
+			continue;
+
```
dirkhh, line 263: The commit message says "broken XML files"... but this is save-git ?

bstoeger:
Could you please check that the code:
```
        if (o2sensor >= 0) {
                dummy.sensor[0] = !o2sensor;
                dummy.sensor[1] = o2sensor;
        }
```
is reasonable?

Two things strike me as odd:
1) This will produce an invalid cylinder id for one-cylinder dives.
2) I don't see where `dummy.sensor[1]` is accessed.

I have to repeat myself: I don't understand the sensor-code _at all_. The data structure as-is makes no sense to me. Moreover, I don't have a dive computer with sensor support, much less any idea of CCR, etc. Also, my life is complicated enough that I don't want to get familiar with these things, sorry.

dirkhh:
I have **no bleeping idea** what that `!o2sensor` is doing there.
This is a coding style that drives me angry. I assume this was written by someone assuming that there could be only two sensors named 0 and 1 or something. And that if we have an o2 sensor, the "other one" (which I think that `not` operator is supposed to create) must be the regular pressure?
Heck if I know.

So much of the code that we have is effectively unmaintained, but has people that rely on it (as we find out every time we try to fix things that seem completely bogus).

So yes, this code makes no sense to me - but I bet if we cut it out, it will break something that works for CCR users today.

PR #3318, opened 2021-09-06
Fix off-by-one causing incorrect profile display

### LINKED ISSUE #3184: Profile shows unexpected bearing if cursor is after the end
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Just a very small thing I noticed. See attached image: when the cursor is placed somewhere after the end of the dive, the profile shows an extra 'bearing: 0' line.

### Operating system:
Windows 10 Home

### Subsurface version:
4.9.10 official, 4.9.10-421 from test

bearing

beckjoj:

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In commit 4724c88 get_plot_details_new was updated to pass an index instead of "entry[i]" into plot_string. This means we are passing "i" to plot_string after the final increment of the loop, instead of getting "entry[i]" within the loop before the final increment. This means if we are mousing over the far right of the graph, where the time based break is not hit, we will end up passing an index equal to nr-2, instead of nr-3, which is intended to shave off the final two rows containing data not useful to the display.

There are a handful of ways to fix this. This commit intends to be consistent with stylistic choices made elsewhere in the project.

### Related issues:
Fixes #3184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-
+core: fix off-by-one causing incorrect profile display #3184
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
 * Use this layout for new entries: `[Area]: [Details about the change] [reference thread / issue]`
--- a/core/profile.c
+++ b/core/profile.c
@@ -1566,18 +1566,17 @@ static void plot_string(const struct dive *d, const struct plot_info *pi, int id
 
 int get_plot_details_new(const struct dive *d, const struct plot_info *pi, int time, struct membuffer *mb)
 {
-	int idx, i;
+	int i;
 
 	/* The two first and the two last plot entries do not have useful data */
 	if (pi->nr <= 4)
 		return 0;
-	for (i = 2; i < pi->nr - 2; i++) {
-		idx = i;
+	for (i = 2; i < pi->nr - 3; i++) {
 		if (pi->entry[i].sec >= time)
 			break;
 	}
-	plot_string(d, pi, idx, mb);
-	return idx;
+	plot_string(d, pi, i, mb);
+	return i;
 }
 
 /* Compare two plot_data entries and writes the results into a string */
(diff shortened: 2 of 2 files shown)


dirkhh (changes requested):
Hey there - and welcome to the project!
I'm thrilled to see a first pull request... and while I started working on it, you already fixed the first issue (missing DCO).

I think that first commit could use some more work. I commented on the discrepancy between the commit message and your code (and believe the code is right). But I also think the code is needlessly convoluted.
What do you think?

beckjoj:
> Hey there - and welcome to the project!

Thanks! Excited to contribute.

> I'm thrilled to see a first pull request... and while I started working on it, you already fixed the first issue (missing DCO).
> 
> I think that first commit could use some more work. I commented on the discrepancy between the commit message and your code (and believe the code is right). But I also think the code is needlessly convoluted.
> What do you think?

I agree. That's how I fixed it locally originally (with `i < pi->nr - 3`). I was being overly cautious on changing execution as little as possible.

Technically there is a functional difference in that `i < pi->nr - 3` means we don't run the code block for the last "real" entry. I wasn't sure if that would be an issue and so wrote the extra lines to preserve original behavior as much as possible (even though there isn't a real difference given the check only breaks). Though, if someone wanted to add any other logic inside the loop, -3 means that logic won't be run for the final "real" element, which might not be obvious given the original bug's existence.  

All that said I'm over analyzing a pretty trivial change.


I'll update with the more straightforward `i < pi->nr - 3`, correct the commit message, and request a re-review

dirkhh:
Instead of playing "bring me a rock" with you, I'll fix things up locally and merge what I'd consider a "clean" version of this.

dirkhh:
Thanks for fixing this bug - and welcome as a Subsurface developer.

merged

PR #3323, opened 2021-09-10
Improve camera sync and reorder sync UI

tsegers:
### Improve camera sync and reorder sync UI

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

1) Add support for camera sync delta of more than 24h
2) Reorder ShiftImageTimes UI

### DIFF
--- a/desktop-widgets/simplewidgets.cpp
+++ b/desktop-widgets/simplewidgets.cpp
@@ -124,11 +124,6 @@ ShiftTimesDialog::ShiftTimesDialog(QWidget *parent) : QDialog(parent),
 
 void ShiftImageTimesDialog::buttonClicked(QAbstractButton *button)
 {
-	if (ui.buttonBox->buttonRole(button) == QDialogButtonBox::AcceptRole) {
-		m_amount = ui.timeEdit->time().hour() * 3600 + ui.timeEdit->time().minute() * 60;
-		if (ui.backwards->isChecked())
-			m_amount *= -1;
-	}
 }
 
 void ShiftImageTimesDialog::syncCameraClicked()
@@ -140,6 +135,10 @@ void ShiftImageTimesDialog::syncCameraClicked()
 							      tr("Image files") + " (*.jpg *.jpeg)");
 	if (fileNames.isEmpty())
 		return;
+
+	ui.timeEdit->setEnabled(false);
+	ui.backwards->setEnabled(false);
+	ui.forward->setEnabled(false);
 
 	picture.load(fileNames.at(0));
 	ui.displayDC->setEnabled(true);
@@ -160,7 +159,10 @@ void ShiftImageTimesDialog::dcDateTimeChanged(const QDateTime &newDateTime)
 	if (!dcImageEpoch)
 		return;
 	newtime.setTimeSpec(Qt::UTC);
-	setOffset(dateTimeToTimestamp(newtime) - dcImageEpoch);
+
+	m_amount = dateTimeToTimestamp(newtime) - dcImageEpoch;
+	if (m_amount)
+		updateInvalid();
 }
 
 void ShiftImageTimesDialog::matchAllImagesToggled(bool state)
--- a/desktop-widgets/simplewidgets.cpp
+++ b/desktop-widgets/simplewidgets.cpp
@@ -122,10 +122,6 @@ ShiftTimesDialog::ShiftTimesDialog(QWidget *parent) : QDialog(parent),
 	connect(quit, SIGNAL(activated()), parent, SLOT(close()));
 }
 
-void ShiftImageTimesDialog::buttonClicked(QAbstractButton *button)
-{
-}
-
 void ShiftImageTimesDialog::syncCameraClicked()
 {
 	QPixmap picture;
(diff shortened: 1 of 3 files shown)


tsegers:
One note on a limitation of commit 044017b: since there is no way to represent time deltas of more than 24h as a QTime, saving and restoring of a value larger than 24h via the preferences will of course not work. As long as QTime is used as a time delta container, this will remain a problem.

desktop-widgets/simplewidgets.cpp
```diff
@@ -124,11 +124,6 @@ ShiftTimesDialog::ShiftTimesDialog(QWidget *parent) : QDialog(parent),
 
 void ShiftImageTimesDialog::buttonClicked(QAbstractButton *button)
 {
-	if (ui.buttonBox->buttonRole(button) == QDialogButtonBox::AcceptRole) {
-		m_amount = ui.timeEdit->time().hour() * 3600 + ui.timeEdit->time().minute() * 60;
-		if (ui.backwards->isChecked())
-			m_amount *= -1;
-	}
```
dirkhh, line 131: so this leaves an empty `buttonClicked()` function? That seems... odd.

dirkhh (commented):
this isn't a feature that I've used in ages and I'm not really familiar with the code. At first glance this puzzles me a bit.
@atdotde - I really hoped that you'd comment since I think you're more familiar with this part of the code

desktop-widgets/simplewidgets.cpp
tsegers: At the risk of triggering Chesterton's Fence, I can only report that I've found no use for updating m_amount with the values from the UI when clicking the button, since any change to the values in the UI will trigger that update anyway (L254)

tsegers: I've just confirmed my suspicion. The only way to prevent `ShiftImageTimesDialog::timeEditChanged()` from triggering is by not changing the value in the `timeEdit` field _and_ having `LastImageTimeOffset=0` set in Subsurface.conf. In that case, the default value of `m_amount`, which is 0, is correct. In any other case, `timeEditChanged()` will be called after every change and `m_amount` updated accordingly. Performing this process when the Ok button is pressed serves no purpose as far as I can tell.

atdotde (commented):
I think this is ok. I am only worried that once the camera button is clicked, how can I get back to manually editing the shift time without closing the dialog?

dirkhh:
That's a weaker endorsement that I was hoping for :)
I can certainly merge this and we can fix it if it's broken.
But having the empty `buttonClicked()` function really puzzles me. Why do we need that function at all if it doesn't do anything?
I guess I need to familiarize myself with this code...

dirkhh:
I tried to play with this PR (finally found the time) and I can't get it to work at all.
Can you help me understand how you envision this to work?
I.e., what's a workflow / scenario that now works and that previously didn't work?

tsegers:
My specific issue was that I took photos on a dive with a camera where the internal clock had probably reset, it being set to somewhere in 2006. The dive happened in 2019. Technically I didn't use the camera to take a picture of my DC, but I did know the time the first picture in the series was taken, so I can just pretend it's a picture of my DC showing that time.

Steps to reproduce on mainline:
1. Select "Load media from file(s)"
2. Select a set of photos that have incorrect dates
3. The shift image time dialog pops up
4. Select "Select image of divecomputer showing time"
5. Select the first image in the initial set of photos selected at step 2
6. The "which date and time are displayed on the image?" field initializes to the photo's timestamp
7. Adjusting the time in that field by an hour updated the field at the very top to show "1:00" and the times of the photos in the mismatching media list adjust as well
8. Adjusting the date by one day (or more) in the "which date and time are displayed on the image?" field has no effect

The pull request addresses the incorrect behavior at step 8. Adjusting the date after the patch has been applied will have the expected behavior. Items in the mismatching media list will update their dates when the DC date is adjusted, and can be successfully aligned with the dive.

dirkhh:
Ah. Right. That's an interesting way to work around the limitation of the time widget.
I'm not in love with the implementation, TBH.
Clearly we don't need the calculations in the `buttonClicked()` method, so we should remove it.
The button is the accept button and simply closes the widget and the calling function then figures out what to do with the data... and the calculation of the `m_amount` happens every time the widget changes... that's why the content of that function isn't needed.

Also, the `QTimeEdit` widget appears to be simply the wrong widget to use here and that should be rethought as well.
Admittedly, maybe that's not what I should require as part of the PR.

So, my request for this PR: can we just drop the `ShiftImageTimesDialog::buttonClicked()` method and the `connect()` for it?

tsegers:
> Also, the QTimeEdit widget appears to be simply the wrong widget to use here and that should be rethought as well.

I'd agree. I think the main challenge is figuring out a user interface design that allows a user to input an arbitrary time offset in an intuitive way. My "solution" of simply disabling the manual input because it can't contain large values of `m_amount` should probably not be the the final fix for this problem.

dirkhh:
Parsing time information isn't that hard. I'll play with better options for this.
Thanks for this PR - I just triggered the checks (for first time contributors GitHub makes that a manual process - later it's automated).

dirkhh:
I have something that I like reasonably well, but I need to play with it some more. I ended up interleaving it with this PR which is why I haven't merged it, yet. But I think this is great and your commits will be part of what I'm doing. Once I like it all, I'll create a fresh PR that you can play with (hopefully later this evening)

dirkhh:
OK, #3327 is the one that I'm considering merging.
I hope you're ok with that @tsegers  - if yes, please close this PR - if no, please comment over there what I got wrong :)

Issue #3227, opened 2021-04-05
Location permission possibly impacting Background Location Service on Android


pabdakine:
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

After enabling the Location Service, I get a single fix but nothing later when the phone is asleep, despite having substantially changed location. When inspecting the Location permission in the OS/app settings, I have only "Deny" and "Allow only while using the app". An email exchange with Dirk indicated there should be a third, "Allow all the time". (I do see this on some of my other apps, so I don't think it's an OS issue.)

### Operating system:

Android 10 (English) / Oxygen OS 10.3.8
OnePlus 7 GM1900

### Subsurface version:

Subsurface-mobile 3.2.0 (5.0.0), installed from Google Play
Same behavior on the immediately prior version (tested 22 March, 2021).

### Steps to reproduce:

1. Enable Location Service
2. Set the device to sleep
3. Change location substantially
4. Wake the device and inspect the list of GPS fixes

### Current behavior:

Single location fix after enabling the Location Service, then no additional fixes.

### Expected behavior:

Various GPS fixes in the list as the device was moved.

dirkhh:
It appears that some Android flavors refuse to give the app background location permission.
I'm investigating the issue.

dirkhh:
with the latest version on a Google Pixel 5 I was able to run the service in the background.
But then that's just an example of "hey, here it works". This doesn't mean it will now work for your device.

pabdakine:
Thank you for looking into it further. For me, version 3.2.2 doesn't present the "Allow All The Time" location permission as an option. Perhaps it's something specific to my OS (now Oxygen OS 11.0.2.1 / Android 11). OTOH, I checked some weather apps, which do present that option. Puzzling.

Again, thank you for all your work on Subsurface. Cheers!

Paul

Aug 18, 2021 7:05:03 PM Dirk Hohndel ***@***.***>:

flindeberg:
I just skimmed the manifest and it seems to assemble some parts dynamically during the build process. So I don't actually know if these permissions are missing in the resulting manifest. Anyhow the following permission has to be in the manifest for Android 10+ for background location to work _at all_:

```
  <!-- Required only when requesting background location access on
       Android 10 (API level 29) and higher. -->
  <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
```
Could be of interest as well:
```
[...]

PR #3324, opened 2021-09-12
Remove gps

dirkhh:

- [x] Functional change

this may be going a bit too far... the ApplyGpsFixes interface might be useful in the future - including the related command stuff. So let's not merge this too aggressively :)

### Changes made:

- remove the GPS service and current position UI components
- remove the GPS fix list and UI components
- remove the qmlmanager interfaces used to implement all this
- remove the command to apply GPS fixes
- remove the GpsLocation class
- remove the related preferences

### Related issues:

#3227

### Additional information:

- This is pretty drastic. 
- We still need to remove the location related permissions and all the other Android and iOS build cruft around that.
- The location menu now is silly - needs to just become a map action
- And of course eventually we need a different way to import GPS data

### Release note:

TBD

### Documentation change:

TBD - lots of things to delete

### Mentions:

@willemferguson


bstoeger (commented):
I have never used the mobile version[*] nor am I involved with Google's or Apple's appstore process. Therefore, I cannot make an informed judgement. The code deletion looks reasonable to me, from what I can tell.

[*] nor do I plan to: I use my phone as a phone with Google maps and my tablet as a notepad with Google maps and always have my laptop with me for anything else.

dirkhh:
If we import GPS info from another app... Won't we need the apply gps fix stuff?

I wasn't sure if this was removing too much

bstoeger:
Yes, I reckon it depends on how you want to implement the alternative.

1) As an independent feature that imports multiple dive locations (like now). Then it would make sense to keep the undo-command.

2) As part of the edit-dive screen. Then probably the editDive undo-command should be adapted.

3) In a completely different way.

It's really impossible for me to judge without knowing your "vision". Since I only have a small dive log, I'm perfectly fine with fetching the GPS locations from various online databases / sources. And really, I don't care about precision down to the meter or so.

dirkhh:
My "vision". That made me snort.
As if my "vision" mattered. What matters is what people are willing to implement. And we have heard from a few people about importing gpx files - so that seems likely to be similar to what we have today.
So I think I convinced myself to drop the last few commits that deal with that...

PR #3326, opened 2021-09-19
Re-do the libdivecomputer fingerprint save/load code

torvalds:
This tries to make our fingerprinting code work better, by avoiding
using the "deviceid" field that has always been unreliable because we've
calculated it multiple different ways, and even for the same version of
subsurface, it ends up changing in the middle (ie we calculate one value
initially, then re-calculate it when we have a proper serial number
string).

So instead, the fingerprinting code will look up and save the
fingerprint file using purely "stable" information that is available
early during the download:

 - the device model name (which is a string with vendor and product name
   separated by a space)

 - the DC_EVENT_DEVINFO 32-bit 'serial' number (which is not necessarily
   a real serial number at all, but hopefully at least a unique number
   for the particular product)

but because the model name is not necessarily a good filename (think
slashes and other possibly invalid characters), we hash that model name
and use the resulting hex number in the fingerprint file name.

This way the fingerprint file is unambiguous at load and save time, and
depends purely on libdivecomputer data.

But because we also need to verify that we have the actual _dive_
associated with that fingerprint, we also need to save the final
deviceid and diveid when saving the fingerprint file, so that when we
load it again we can look up the dive and verify that we have it before
we use the fingerprint data.

To do that, the fingerprint file itself contains not just the
fingerprint data from libdivecomputer, but the last 8 bytes of the file
are the (subsurface) deviceid and the diveid of the dive that is
associated with the fingerprint.

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -600,7 +600,8 @@ static dc_status_t libdc_header_parser(dc_parser_t *parser, device_data_t *devda
 		return rc;
 	}
 
-	dive->dc.deviceid = devdata->deviceid;
+	// Our deviceid is the hash of the serial number
+	dive->dc.deviceid = 0;
 
 	if (rc == DC_STATUS_SUCCESS) {
 		tm.tm_year = dt.year;
@@ -771,16 +772,6 @@ static int dive_cb(const unsigned char *data, unsigned int size,
 	dive->dc.model = strdup(devdata->model);
 	dive->dc.diveid = calculate_diveid(fingerprint, fsize);
 
-	/* Should we add it to the cached fingerprint file? */
-	if (fingerprint && fsize && !devdata->fingerprint) {
-		devdata->fingerprint = calloc(fsize, 1);
-		if (devdata->fingerprint) {
-			devdata->fsize = fsize;
-			devdata->fdiveid = dive->dc.diveid;
-			memcpy(devdata->fingerprint, fingerprint, fsize);
-		}
-	}
-
 	// Parse the dive's header data
 	rc = libdc_header_parser (parser, devdata, dive);
 	if (rc != DC_STATUS_SUCCESS) {
@@ -796,6 +787,22 @@ static int dive_cb(const unsigned char *data, unsigned int size,
 	}
 
 	dc_parser_destroy(parser);
+
+	/*
+	 * Save off fingerprint data.
+	 *
+	 * NOTE! We do this after parsing the dive fully, so that
+	 * we have the final deviceid here.
+	 */
+	if (fingerprint && fsize && !devdata->fingerprint) {
+		devdata->fingerprint = calloc(fsize, 1);
+		if (devdata->fingerprint) {
+			devdata->fsize = fsize;
+			devdata->fdeviceid = dive->dc.deviceid;
+			devdata->fdiveid = dive->dc.diveid;
+			memcpy(devdata->fingerprint, fingerprint, fsize);
+		}
+	}
 
 	/* If we already saw this dive, abort. */
 	if (!devdata->force_download && find_dive(&dive->dc)) {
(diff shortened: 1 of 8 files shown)


torvalds:
Note: this is a first cut. Entirely untested. It built locally for me, so I uploaded it and made a pull request for comments.

It fails the builds, so I must have done something horribly wrong. 

The existing fingerprinting code definitely doesn't work, because it uses different 'deviceid' at save time from load time.  And even if the load were to work (it won't) the device id it used to use at load time is also used to then find and verify the dive, but because the deviceid in the actual dives we have is the new-style one, it won't verify.

torvalds:
Oh, and the build failures are becasue I removed the (now stale) DCDeviceData::deviceId() accessor functions, but apparently they are used on the mobile side in mobile-widgets/qmlmanager.cpp.

Fine, I'll redo this pull request tomorrow, but in the meantime the objective here is to see if somebody sees some complete braino in my thinking.

dirkhh:
The logic for the filename seems solid.
I'll review the code tomorrow.

torvalds:
I'll put this explanation for why the old code was broken, and why this new (currently also broken ;) change was done here too, just for posterity and for github users.

Basically the fingerprinting was broken, because of how we fixed the device management to have those stable deviceids based on the real serial number.  So we ended up with two different deviceids we used: (a) the old legacy one, and (b) the new stable one.

> At fingerprint *saving* time, we have the new deviceid, and we have in
> fact thrown away all the odd libdivecomputer stuff.
> 
> But at fingerprint *loading* time, we only have the DC_EVENT_DEVINFO
> event, and not anything else.
> 
> And the real problem is that the loading code actually wants to also
> verify that we *have* the dive described by the fingerprint, but in
> order to find that dive, it wants the proper device ID that we don't
> have yet.
> 
> Oh it's lovely.
> 
> Basically, the loading code happens before the real data has been
> downloaded and parsed - because the whole point is that we want to
> give the fingerprint to the downloading code.
> 
> But because the real data hasn't been downloaded and parsed yet, we
> don't have that real serial number yet, we only have the nasty
> 'serial' number from libdivecomputer that we don't rely on and don't
> actually save (and have never saved - we've saved the hashed
> 'deviceid' and now even that is no longer saved).
> 
> How very annoying.
>

torvalds:
Final note for Berthold: none of this fingerprint code matters for most dove computers. They just download dives newest first, and when we see a dive we recognize, we're done.

But some older models basically download all of the dive computer memory to even find that newest dive, so for them the fingerprint is basically a pointer to where in the dive computer memory we've already seen dive data, did that they can avoid downloading old and uninteresting stuff.

Which is also why neither I nor dirk tend to notice when it's broken. It only matters for some people, but then it can matter a *lot*

core/libdivecomputer.c
```diff
@@ -942,18 +858,57 @@ static void do_save_fingerprint(device_data_t *devdata, const char *tmp, const c
...
+ * up using the DC_EVENT_DEVINFO data.
+ *
+ * End result:
+ *
+ *   - fingerprint filename depends on the model name and 'devinfo.serial'
+ *     so that we can look it up at DC_EVENT_DEVINFO time before the full
+ *     info has been parsed.
+ *
+ *   - the fingerprint file contains the 'diveid' of the finegrprinted dive,
+ *     whish is just a hash of ther fingerprint itself.
```
bstoeger: typo: whish -> which

core/libdivecomputer.c
```diff
@@ -942,18 +858,57 @@ static void do_save_fingerprint(device_data_t *devdata, const char *tmp, const c
...
+ *   - we also save the final 'deviceid' in the fingerprint file, so that
+ *     looking up the dive associated with the fingerprint is possible.
  */
 static void save_fingerprint(device_data_t *devdata)
 {
+	uint32_t model, serial;
 	char *dir, *tmp, *final;
 
-	if (!devdata->fingerprint)
+	// Don't try to save nonexistend fingerprint data
```
bstoeger: typo: nonexistend -> nonexistent

core/libdivecomputer.c
```diff
@@ -942,18 +858,57 @@ static void do_save_fingerprint(device_data_t *devdata, const char *tmp, const c
...
+	if (!devdata->fingerprint || !devdata->fdiveid)
+		return;
+
+	// Don't save fingerprint for no dives downloaded
+	if (!devdata->download_table->nr)
```
bstoeger: Not sure, but if we return here, then `dev_data->fingerprint` is never freed..?

Also where are the other device_data_t members freed...?

core/libdivecomputer.c
```diff
@@ -992,14 +947,13 @@ static void verify_fingerprint(dc_device_t *device, device_data_t *devdata, cons
 {
```
bstoeger, line 957: For some reason I can't comment on the comment above, but note that now
```
 * The fingerprint cache files contain the actual libdivecomputer
 * fingerprint, followed by 4 bytes of diveid data.
```
is not the complete story anymore.

core/libdivecomputer.c
```diff
@@ -992,14 +947,13 @@ static void verify_fingerprint(dc_device_t *device, device_data_t *devdata, cons
...
 	unsigned int diveid, deviceid;
```
bstoeger: Nitpick: better use `uint32_t`?

core/libdivecomputer.c
```diff
@@ -1013,13 +967,19 @@ static void verify_fingerprint(dc_device_t *device, device_data_t *devdata, cons
...
 	if (devdata->force_download)
 		return;
-	cachename = format_string("%s/fingerprints/%04x",
-		system_default_directory(), devdata->deviceid);
+
+	// Model hash and libdivecomputer 32-bit 'serial number' for the file name
+	model = calculate_string_hash(devdata->model);
+	serial = devdata->devinfo.serial;
+
+	cachename = format_string("%s/fingerprints/%04x.%u",
```
bstoeger: Maybe factor this out into its own function, so that the fingerprint filenames can never run out of sync?

bstoeger (approved):
Wow, I was blissfully unaware of this topic.

I only did a peephole-review and it all looks reasonable to me. Only two typos in comments. I am somewhat overwhelmed by all this device-ids, string hashes, etc.

One thing that confuses me is that I don't see where the `device_data_t` members are freed..? Am I missing something?

core/libdivecomputer.c
bstoeger: Also ther -> the

torvalds:
Ok, after several false starts (oh how I love that windows-only smtk-import thing), I think the current state is ok. 

I also fixed up the code based on the commentary from Berthold.

Let's see that the current thing still works after the fixups, but fingers crossed.

torvalds:
Hold off. I got the directory handling wrong, and didn't use the "fingerprints" directory for reading.

Will update after actual testing.

torvalds:
Ok, I just pushed out something that I've actually tested and works right both on my Scubapro G2 (no proper serial number) and my Suunto Eon Steel Black (with proper serial number).

This adds some noise to the debug output too, so that we can see what is going on if this ever breaks.

Of course, maybe my final changes will have some build issue on other platforms, so don't merge until the checks all pass.

torvalds:
Ugh. One final error in the new debug message printing code.

I'm actually a bit surprised that we don't have the printf format checking for dev_info(), which is why I didn't notice it locally. We do have them for the membuffer functions (see __printf() macro). Oh well.

dirkhh:
Wait, you actually **TESTED** this? Has your account been taken over by some government actor?
I mean, had you claimed to have run this under a debugger I would have known that it's not you, but this is pretty close to being unconvincing...

That said, let's review this and test it here...

dirkhh (approved):
This looks comprehensive and reasonable.
I did a couple of quick tests and couldn't make it fail - and for the cases where it might fail at least we have better debug output. So overall a huge improvement.
Thanks

PR #3327, opened 2021-09-21
Image shift improvements

dirkhh:

- [x] Bug fix
- [x] Functional change

This is a slight redo of #3323 - it reuses those commits (with minor changes and some cleanup and updates to commit messages) and adds a couple of tweaks:
- consistent filter for image file names (it was weird that image files that you could use for the picture itself wouldn't show up for the "picture of the dive computer")
- redo the manual time delta input; `QTimeEdit` is too limited for this use case as @tsegers points out - so this implements a simple `QLineEdit` that does its own limited validation and acts reasonably well even for very large time deltas

### Related issues:

#3323 

### Release note:

done

### Documentation change:

I guess this might need small changes to the manual - but it's really not a big change, just making things more generic

### Mentions:

@tsegers @bstoeger

### DIFF
--- a/desktop-widgets/shiftimagetimes.ui
+++ b/desktop-widgets/shiftimagetimes.ui
@@ -55,61 +55,16 @@
         </property>
        </widget>
       </item>
-      <item>
-       <widget class="QLabel" name="warningLabel">
-        <property name="enabled">
-         <bool>true</bool>
-        </property>
-        <property name="styleSheet">
-         <string notr="true">color: red;</string>
-        </property>
-        <property name="text">
-         <string>Warning!
-Not all media files have timestamps in the range between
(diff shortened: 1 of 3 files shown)


tsegers:
Minute values restored from the preferences by `setOffset` are not 0-padded and therefore do not pass the field validator. Additionally, allowing only 5 digits in the hours field limits the maximum offset to 11 years. I suggest allowing at least 6 digits.

dirkhh:
I'll admit... the idea that the clock is 10+ years wrong seems silly (I know, I know).
And it seems even more silly that someone would translate this into hours.
I thought about allowing for years and days but it seemed like such a rare thing and like it would only cause additional problems.
But yeah, I'll adjust it.
Thanks for the review!

PR #3342, opened 2021-10-26
replace QRegExp with QRegularExpression

dirkhh:

- [x] Code cleanup

Qt 6 no longer supports the ancient `QRegExp` (that was basically the Qt 4 class that was supposed to be replaced by Qt 5's new `QRegularExpression`). This pull request replaces the dozen or so instances where we for some reason used `QRegExp`.
Using that class is slightly less verbose (which might be why we used it), but other than that there is no real advantage to it.

Of course, mechanical rip-and-replace efforts like this are fraught with risk. I'd really appreciate some solid review here.

@atdotde the first commit should be correct - but I have some serious questions about the use of the `site->notes` field here, vs using the taxonomy...

### Changes made:

rip and replace
`exactMatch()` doesn't exist anymore - so work around that with anchored patterns -- oh, one of those `exactMatch()` calls (that's the last commit) was completely bogus, given the pattern itself.

### Related issues:

well... this is required for any attempt to port to Qt 6.

### Mentions:

@bstoeger I'd love a critical eye
@tcanabrava if you have a moment :)
@atdotde

### DIFF
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -16,6 +16,7 @@
 #include "core/pref.h"
 #include "core/sample.h"
 #include "core/selection.h"
+#include "core/taxonomy.h"
 #include "exportfuncs.h"
 
 // Default implementation of the export callback: do nothing / never cancel
@@ -119,14 +120,10 @@ void export_TeX(const char *filename, bool selected_only, bool plain, ExportCall
 		struct tm tm;
 		utc_mkdate(dive->when, &tm);
 
+		const char *country = NULL;
 		dive_site *site = dive->dive_site;
-		QRegExp ct("countrytag: (\\w+)");
-		QString country;
-		if (site && ct.indexIn(site->notes) >= 0)
-			country = ct.cap(1);
-		else
-			country = "";
-
+		if (site)
+			country = taxonomy_get_country(&site->taxonomy);
 		pressure_t delta_p = {.mbar = 0};
 
 		QString star = "*";
@@ -158,7 +155,7 @@ void export_TeX(const char *filename, bool selected_only, bool plain, ExportCall
 		site ? put_format(&buf, "\\def\\%sgpslat{%f}\n", ssrf, site->location.lat.udeg / 1000000.0) : put_format(&buf, "\\def\\%sgpslat{}\n", ssrf);
 		site ? put_format(&buf, "\\def\\%sgpslon{%f}\n", ssrf, site->location.lon.udeg / 1000000.0) : put_format(&buf, "\\def\\gpslon{}\n");
 		put_format(&buf, "\\def\\%scomputer{%s}\n", ssrf, dive->dc.model);
-		put_format(&buf, "\\def\\%scountry{%s}\n", ssrf, qPrintable(country));
+		put_format(&buf, "\\def\\%scountry{%s}\n", ssrf, country ?: "");
 		put_format(&buf, "\\def\\%stime{%u:%02u}\n", ssrf, FRACTION(dive->duration.seconds, 60));
 
 		put_format(&buf, "\n%% Dive Profile Details:\n");
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -23,7 +23,7 @@
 #include "xmlparams.h"
 #include "core/git-access.h" // for CLOUD_HOST definitions
 #include <QFile>
-#include <QRegExp>
+#include <QRegularExpression>
 #include <QDir>
 #include <QDebug>
 #include <QStandardPaths>
@@ -139,13 +139,14 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
 		if (txt[pos].isDigit()) {
 			if (numberDefined)
 				return false;
-			QRegExp numberRe("(\\d+(?:[\\.,]\\d+)?).*");
-			if (!numberRe.exactMatch(txt.mid(pos)))
+			QRegularExpression numberRe("\\A(\\d+(?:[\\.,]\\d+)?).*");
+			QRegularExpressionMatch match = numberRe.match(txt.mid(pos));
+			if (!match.hasMatch())
 				return false;
-			number = numberRe.cap(1).toDouble();
+			number = match.captured(1).toDouble();
 			numberDefined = true;
 			posBeforeNumber = pos;
-			pos += numberRe.cap(1).size() - 1;
+			pos += match.captured(1).size() - 1;
 		} else if (positives.indexOf(txt[pos]) >= 0) {
 			if (sign != 0)
 				return false;
@@ -226,10 +227,11 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
 * Parse special coordinate formats that cannot be handled by parseCoord.
 */
 static bool parseSpecialCoords(const QString &txt, double &latitude, double &longitude) {
-	QRegExp xmlFormat("(-?\\d+(?:\\.\\d+)?),?\\s+(-?\\d+(?:\\.\\d+)?)");
-	if (xmlFormat.exactMatch(txt)) {
-		latitude = xmlFormat.cap(1).toDouble();
-		longitude = xmlFormat.cap(2).toDouble();
+	QRegularExpression xmlFormat("\\A(-?\\d+(?:\\.\\d+)?),?\\s+(-?\\d+(?:\\.\\d+)?)\\z");
+	QRegularExpressionMatch match = xmlFormat.match(txt);
+	if (match.hasMatch()) {
+		latitude = match.captured(1).toDouble();
+		longitude = match.captured(2).toDouble();
 		return true;
 	}
 	return false;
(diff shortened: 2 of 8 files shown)


backend-shared/exportfuncs.cpp
```diff
@@ -120,13 +121,13 @@ void export_TeX(const char *filename, bool selected_only, bool plain, ExportCall
...
 
 		dive_site *site = dive->dive_site;
-		QRegExp ct("countrytag: (\\w+)");
-		QString country;
-		if (site && ct.indexIn(site->notes) >= 0)
-			country = ct.cap(1);
-		else
-			country = "";
-
+		QString country = "";
```
bstoeger: I would suggest to simply do
```
QString country;
```
That's the most efficient and clear method to generate an empty `QString`.

core/qthelper.cpp
```diff
@@ -139,13 +139,14 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
...
 			if (numberDefined)
 				return false;
-			QRegExp numberRe("(\\d+(?:[\\.,]\\d+)?).*");
-			if (!numberRe.exactMatch(txt.mid(pos)))
+			QRegularExpression numberRe(QRegularExpression::anchoredPattern("(\\d+(?:[\\.,]\\d+)?).*"));
+			QRegularExpressionMatch match = numberRe.match(txt.mid(pos));
+			if (!match.hasMatch())
 				return false;
-			number = numberRe.cap(1).toDouble();
+			number = match.captured(1).toDouble();
```
bstoeger, line 146: Idle comment: The matches start at 1 ... that's very unidiomatic.

bstoeger (approved):
I'm currently at the airport and will be only very sparsely available for the next two weeks.

I couldn't review this in detail. Just one general comment: unless `QRegularExpression' does some `QStringLiteral` trickery, most of these regexp objects should be made `static`, so that they are compiled only once.

Outside of function scope (at source file scope) is generally preferred, because then they are initialized at application startup. At function scope it needs some thread-synchronization to check whether the object is already initialized. If the regexps depend on translation or other runtime things, that's of course a necessity.

backend-shared/exportfuncs.cpp
dirkhh: right. silly me

core/qthelper.cpp
dirkhh: so `captured(0)` is the substring that matched the overall regular expression. And then each of the "captures" are numbered starting at 1. That's not all that unusual, I thought...

dirkhh:
>  unless `QRegularExpression' does some `QStringLiteral`trickery, most of these regexp objects should be made`static`, so that they are compiled only once.

I need to look into this - we obviously didn't do this before for `QRegExp` - and we aren't (generally) doing it for the dozens of other uses of `QRegularExpression` that we already have in the code... so I don't think this fits the overall goal for this PR

atdotde:
I am currently looking into the country thing. Might take a bit longer since Xcode updated in the background and thus build failed in stupid ways, setting up again. IIRC for the TeX version of the logbook page, I preferred to have "D" for the country rather than "Germany" (which comes out fo the taxonomy)  and that might have been what the COUNTRYTAG did... As I said, I am investigating that.

backend-shared/exportfuncs.cpp
```diff
@@ -120,13 +121,13 @@ void export_TeX(const char *filename, bool selected_only, bool plain, ExportCall
 		utc_mkdate(dive->when, &tm);
 
 		dive_site *site = dive->dive_site;
-		QRegExp ct("countrytag: (\\w+)");
```
atdotde, line 123: Indeed, this should rather be replaced simply by
 ` QString country = QString(taxonomy_get_country(&site->taxonomy));`

plus an include of core/taxonomy.h

mobile-widgets/qmlmanager.cpp
```diff
@@ -1686,11 +1692,12 @@ QString QMLManager::getDate(const QString& diveId)
 
 QString QMLManager::getVersion() const
 {
-	QRegExp versionRe(".*:([()\\.,\\d]+).*");
```
atdotde, line 1689: shouldn't this also be anchored (as it was an exact match)?

mobile-widgets/qmlmanager.cpp
```diff
@@ -2207,15 +2214,15 @@ void QMLManager::importCacheRepo(QString repo)
 QStringList QMLManager::cloudCacheList() const
 {
 	QDir localCacheDir(QString("%1/cloudstorage/").arg(system_default_directory()));
-	QStringList dirs = localCacheDir.entryList().filter(QRegExp("...+"));
```
atdotde, line 2210: Is this a very unidiomatic way of saying "this string is at least 3 characters long"?!? . matches any character and the + means one or more. Or am I missing something?

atdotde (commented):
Most of this I think is clearly correct. Some parts in the QMLManager look scary, though. But I don't think this was affected in any way by the transition.

I agree with @bstoeger that at least those regexes where the pattern is a constant string and gets its own QRegularExpression variable should be moved to global scope.

backend-shared/exportfuncs.cpp
dirkhh: happy to replace that

mobile-widgets/qmlmanager.cpp
dirkhh: look at that regular expression.
look it deep in the eye.
...
...
wait for it
...
Now... do you still think it makes sense to anchor it?

mobile-widgets/qmlmanager.cpp
dirkhh: Yes - this uses the filter function to filter out filenames that are shorter than 2 characters. So it kicks out specifically `.` and `..`
I'm sure there are better ways to do this... but again... this is existing code and all I'm doing is remove the use of `QRegExp`

dirkhh:
> I agree with @bstoeger that at least those regexes where the pattern is a constant string and gets its own QRegularExpression variable should be moved to global scope.

But should that happen in this PR? Or should this review happen for ALL usage of regular expressions across the code? I think the latter.

backend-shared/exportfuncs.cpp
dirkhh:
```
@@ -119,14 +120,10 @@ void export_TeX(const char *filename, bool selected_only, bool plain, ExportCall
 		struct tm tm;
 		utc_mkdate(dive->when, &tm);
 
+		const char *country = NULL;
 		dive_site *site = dive->dive_site;
-		QRegExp ct("countrytag: (\\w+)");
-		QString country;
-		if (site && ct.indexIn(site->notes) >= 0)
-			country = ct.cap(1);
-		else
-			country = "";
-
+		if (site)
+			country = taxonomy_get_country(&site->taxonomy);
 		pressure_t delta_p = {.mbar = 0};
 
 		QString star = "*";
@@ -158,7 +155,7 @@ void export_TeX(const char *filename, bool selected_only, bool plain, ExportCall
 		site ? put_format(&buf, "\\def\\%sgpslat{%f}\n", ssrf, site->location.lat.udeg / 1000000.0) : put_format(&buf, "\\def\\%sgpslat{}\n", ssrf);
 		site ? put_format(&buf, "\\def\\%sgpslon{%f}\n", ssrf, site->location.lon.udeg / 1000000.0) : put_format(&buf, "\\def\\gpslon{}\n");
 		put_format(&buf, "\\def\\%scomputer{%s}\n", ssrf, dive->dc.model);
-		put_format(&buf, "\\def\\%scountry{%s}\n", ssrf, qPrintable(country));
+		put_format(&buf, "\\def\\%scountry{%s}\n", ssrf, country ?: "");
 		put_format(&buf, "\\def\\%stime{%u:%02u}\n", ssrf, FRACTION(dive->duration.seconds, 60));
 
 		put_format(&buf, "\n%% Dive Profile Details:\n");
```
and now without the detour via `QString`

dirkhh:
fun... `anchoredPattern()` is new in Qt 5.15 - so we can't simply use it (I only used it because that was the recommended conversion according to the Qt6 documents)... anyway, just manually adjusted the regular expressions instead.

This is such an incredibly constructive and useful thing to spend my time on. I am so thrilled about the opportunity to work on Qt 6.

Someone should buy me an m1 Mac so I at least have a reason to want this, instead of just a reason to hate this...

mobile-widgets/qmlmanager.cpp
atdotde: C'mon. Without the anchor, the leading and trailing .* makes no sense ;-)

dirkhh: and we are back to "fixing the code" vs. "replacing one data structure with another".
But your point is valid. I'll fix the code.

mobile-widgets/qmlmanager.cpp
atdotde: According to the docs, localCacheDir.entryList(QDir::NoDotAndDotDot) should just do the job.

backend-shared/exportfuncs.cpp
atdotde: thanks.

dirkhh:
I updated a few commit messages and merged locally.

PR #3344, opened 2021-10-31
build-system: correctly build against hidapi

dirkhh:
On some Linux system (like openSUSE) there are two versions of the
library and we need to explicitly pick one.

- [x] Build system change

this only makes sense if we merge https://github.com/subsurface/libdc/pull/30

### Changes made:

add the different library name as something to look for


dirkhh:
this is obviously bogus... I need to test this some more :)

dirkhh:
not only is it bogus, with some nudging from Jef I was able to see that this is plain wrong.

PR #3345, opened 2021-10-31
store fingerprints with the dive log

dirkhh:

- [x] Functional change
- [x] New feature

As discussed on the mailing list. Considering the fingerprint data "system local" is flawed - especially for divecomputers like the G2 that download all the data if we don't have a valid fingerprint, before they start to parse the data.

This creates an internal data structure to manage the fingerprints, including accessor functions that can be used from C code.
It then adds lookup and storage of fingerprints in that data structure
Finally it adds load/save support for both XML and git

### Additional information:

this deserves some solid review.
I don't think I am ready to contemplate how this might interact with the undo system. There seemed too many weird issues when I tried to think about undoing a download operation. We don't throw away the on disk fingerprint either (or go back to a previous version, which would be more idiomatic).

### Release note:

yeah, I should write something :)

### Documentation change:

I don't think that's something we want to really document. It fixes a weird cornercase that most people won't care about - but people with a G2 **really** care about.

### Mentions:

@bstoeger @torvalds

### DIFF
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -252,3 +252,54 @@ extern "C" void create_fingerprint_node(struct fingerprint_table *table, uint32_
 		table->fingerprints.insert(it, fpr);
 	}
 }
+
+extern "C" void create_fingerprint_node_from_hex(struct fingerprint_table *table, uint32_t model, uint32_t serial,
+						const char *hex_data, uint32_t fdeviceid, uint32_t fdiveid)
+{
+	QByteArray raw = QByteArray::fromHex(hex_data);
+	create_fingerprint_node(table, model, serial,
+				(const unsigned char *)raw.constData(), raw.size(), fdeviceid, fdiveid);
+}
+
+extern "C" int nr_fingerprints(struct fingerprint_table *table)
+{
+	return table->fingerprints.size();
+}
+
+extern "C" uint32_t fp_get_model(struct fingerprint_table *table, unsigned int i)
+{
+	if (!table || i >= table->fingerprints.size())
+		return 0;
+	return table->fingerprints[i].model;
+}
+
+extern "C" uint32_t fp_get_serial(struct fingerprint_table *table, unsigned int i)
+{
+	if (!table || i >= table->fingerprints.size())
+		return 0;
+	return table->fingerprints[i].serial;
+}
+
+extern "C" uint32_t fp_get_deviceid(struct fingerprint_table *table, unsigned int i)
+{
+	if (!table || i >= table->fingerprints.size())
+		return 0;
+	return table->fingerprints[i].fdeviceid;
+}
+
+extern "C" uint32_t fp_get_diveid(struct fingerprint_table *table, unsigned int i)
+{
+	if (!table || i >= table->fingerprints.size())
+		return 0;
+	return table->fingerprints[i].fdiveid;
+}
+
+extern "C" char *fp_get_data(struct fingerprint_table *table, unsigned int i)
+{
+	if (!table || i >= table->fingerprints.size())
+		return 0;
+	struct fingerprint_record *fpr = &table->fingerprints[i];
+	// fromRawData() avoids one copy of the raw_data
+	QByteArray hex = QByteArray::fromRawData((char *)fpr->raw_data, fpr->fsize).toHex();
+	return strdup(hex.constData());
+}
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "ssrf.h"
 #include "dive.h"
+#include "divelist.h"
 #include "subsurface-string.h"
 #include "device.h"
 #include "errorhelper.h" // for verbose flag
@@ -9,6 +10,7 @@
 #include <QString> // for QString::number
 
 struct device_table device_table;
+struct fingerprint_table fingerprint_table;
 
 bool device::operator==(const device &a) const
 {
@@ -196,4 +198,57 @@ extern "C" struct device_table *alloc_device_table()
 extern "C" void free_device_table(struct device_table *devices)
 {
 	delete devices;
+}
+
+// managing fingerprint data
+bool fingerprint_record::operator<(const fingerprint_record &a) const
+{
+	if (model == a.model)
+		return serial < a.serial;
+	return model < a.model;
+}
+
+// annoyingly, the Cressi Edy doesn't support a serial number (it's always 0), but still uses fingerprints
+// so we can't bail on the serial number being 0
+extern "C" unsigned int get_fingerprint_data(const struct fingerprint_table *table, uint32_t model, uint32_t serial, const unsigned char **fp_out)
+{
+	if (model == 0 || fp_out == nullptr)
+		return 0;
+	struct fingerprint_record fpr = { model, serial };
+	auto it = std::lower_bound(table->fingerprints.begin(), table->fingerprints.end(), fpr);
+	if (it != table->fingerprints.end() && it->model == model && it->serial == serial) {
+		// std::lower_bound gets us the first element that isn't smaller than what we are looking
+		// for - so if one is found, we still need to check for equality
+		if (has_dive(it->fdeviceid, it->fdiveid)) {
+			*fp_out = it->raw_data;
+			return it->fsize;
+		}
+	}
+	return 0;
+}
+
+extern "C" void create_fingerprint_node(struct fingerprint_table *table, uint32_t model, uint32_t serial,
+				       const unsigned char *raw_data_in, unsigned int fsize, uint32_t fdeviceid, uint32_t fdiveid)
+{
+	// since raw data can contain \0 we copy this manually, not as string
+	unsigned char *raw_data = (unsigned char *)malloc(fsize);
+	if (!raw_data)
+		return;
+	memcpy(raw_data, raw_data_in, fsize);
+
+	struct fingerprint_record fpr = { model, serial, raw_data, fsize, fdeviceid, fdiveid };
+	auto it = std::lower_bound(table->fingerprints.begin(), table->fingerprints.end(), fpr);
+	if (it != table->fingerprints.end() && it->model == model && it->serial == serial) {
+		// std::lower_bound gets us the first element that isn't smaller than what we are looking
+		// for - so if one is found, we still need to check for equality - and then we
+		// can update the existing entry; first we free the memory for the stored raw data
+		free(it->raw_data);
+		it->fdeviceid = fdeviceid;
+		it->fdiveid = fdiveid;
+		it->raw_data = raw_data;
+		it->fsize = fsize;
+	} else {
+		// insert a new one
+		table->fingerprints.insert(it, fpr);
+	}
 }
--- a/core/device.h
+++ b/core/device.h
@@ -44,8 +44,18 @@ extern void free_device_table(struct device_table *devices);
 // create fingerprint entry - raw data remains owned by caller
 extern void create_fingerprint_node(struct fingerprint_table *table, uint32_t model, uint32_t serial,
 				   const unsigned char *raw_data, unsigned int fsize, uint32_t fdeviceid, uint32_t fdiveid);
+extern void create_fingerprint_node_from_hex(struct fingerprint_table *table, uint32_t model, uint32_t serial,
+					    const char *hex_data, uint32_t fdeviceid, uint32_t fdiveid);
 // look up the fingerprint for model/serial - returns the number of bytes in the fingerprint; memory owned by the table
 extern unsigned int get_fingerprint_data(const struct fingerprint_table *table, uint32_t model, uint32_t serial, const unsigned char **fp_out);
+
+// access the fingerprint data from C
+extern int nr_fingerprints(struct fingerprint_table *table);
+extern uint32_t fp_get_model(struct fingerprint_table *table, unsigned int i);
+extern uint32_t fp_get_serial(struct fingerprint_table *table, unsigned int i);
+extern uint32_t fp_get_deviceid(struct fingerprint_table *table, unsigned int i);
+extern uint32_t fp_get_diveid(struct fingerprint_table *table, unsigned int i);
+extern char *fp_get_data(struct fingerprint_table *table, unsigned int i);
 
 #ifdef __cplusplus
 }
--- a/core/device.h
+++ b/core/device.h
@@ -15,6 +15,7 @@ struct dive_table;
 
 // global device table
 extern struct device_table device_table;
+extern struct fingerprint_table fingerprint_table;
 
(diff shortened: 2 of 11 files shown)


dirkhh:
did anyone have a chance to take a look?

bstoeger:
I'm away until 10th of November - don't expect anything substantial from until then

core/device.cpp
```diff
@@ -197,3 +199,102 @@ extern "C" void free_device_table(struct device_table *devices)
...
+
+// annoyingly, the Cressi Edy doesn't support a serial number (it's always 0), but still uses fingerprints
+// so we can't bail on the serial number being 0
+extern "C" unsigned int get_fingerprint_data(const struct fingerprint_table *table, uint32_t model, uint32_t serial, unsigned char **fp_out)
+{
+	if (model == 0 || fp_out == nullptr)
+		return 0;
+	struct fingerprint_record fpr = { model, serial };
+	auto it = std::lower_bound(table->fingerprints.begin(), table->fingerprints.end(), fpr);
+	if (it != table->fingerprints.end()) {
```
bstoeger: std::lower_bound finds the first element that is greater or equal. Therefore, you probably have to check for equality as well?

core/device.cpp
```diff
@@ -197,3 +199,102 @@ extern "C" void free_device_table(struct device_table *devices)
...
+{
+	// since raw data can contain \0 we copy this manually, not as string
+	unsigned char *raw_data = (unsigned char *)malloc(fsize);
+	if (!raw_data)
+		return;
+	memcpy(raw_data, raw_data_in, fsize);
+
+	struct fingerprint_record fpr = { model, serial, raw_data, fsize, fdeviceid, fdiveid };
+	auto it = std::lower_bound(table->fingerprints.begin(), table->fingerprints.end(), fpr);
+	if (it != table->fingerprints.end()) {
```
bstoeger: see above.

core/device.cpp
```diff
@@ -197,3 +199,102 @@ extern "C" void free_device_table(struct device_table *devices)
...
+		}
+	}
+	return 0;
+}
+
+extern "C" void create_fingerprint_node(struct fingerprint_table *table, uint32_t model, uint32_t serial,
+				       const unsigned char *raw_data_in, unsigned int fsize, uint32_t fdeviceid, uint32_t fdiveid)
+{
+	// since raw data can contain \0 we copy this manually, not as string
+	unsigned char *raw_data = (unsigned char *)malloc(fsize);
```
bstoeger, line 234: At a quick glance, I did not see a matching `free()`.

If you really want to bite the bullet and implement C++-style ownership semantics, you could use a `std::vector<unsigned char>` and instead of `malloc` do a `raw_data.resize(fsize)`? Get a pointer to the data with `&raw_data[0]`.

Or perhaps this is all unnecessary since these things are permanent during the application's life-time.

core/device.cpp
```diff
@@ -197,3 +199,102 @@ extern "C" void free_device_table(struct device_table *devices)
...
+	if (!table || i < 0 || i >= table->fingerprints.size())
+		return 0;
+	return table->fingerprints.at(i).fdeviceid;
+}
+
+extern "C" uint32_t fp_get_diveid(struct fingerprint_table *table, int i)
+{
+	if (!table || i < 0 || i >= table->fingerprints.size())
+		return 0;
+	return table->fingerprints.at(i).fdiveid;
```
bstoeger: For your interest: in std::vector, the difference between `at(i)` and `[i]` is that the former will throw an exception for out-of-bound accesses, whereas the latte will gladly do read/write beyond the buffer. This may or may not be what you want.

bstoeger:
Note that if you don't want to bother with low-level algorithms such as `std::lower_bound`, you could use a `std::map<fingerprint_key, fingerprint_data>`. Beware, it has some strange semantics: `table[key]` will create an entry if it doesn't exist!

On the other hand, I think the code is close to working correctly, just give me a day or two for a closer look.

One day, I should create a `sorted_vector` template that abstracts from this low-level stuff.

core/device.cpp
dirkhh: Yes - that's a clear bug in the code.
Teaches me not to use these fancy functions that I don't really understand, anyway.

core/device.cpp
dirkhh: the free needs to happen below where we update the value - if we ever update the value. Otherwise it's indeed app-life-time.
And of course the update is completely bogus as once again it misunderstands what `std::lower_bound` does

core/device.cpp
dirkhh: so you are saying I could simply use `table->fingerprints[i].fdiveid` here since I checked the bounds right above?

dirkhh:
> Note that if you don't want to bother with low-level algorithms such as `std::lower_bound`, you could use a `std::map<fingerprint_key, fingerprint_data>`. Beware, it has some strange semantics: `table[key]` will create an entry if it doesn't exist!

I clearly don't have a clue what I'm doing here. So using something that I know even less about doesn't seem an improvement. If anything, I think I should go back to writing this in C code that I actually understand, TBH.

> On the other hand, I think the code is close to working correctly, just give me a day or two for a closer look.

Not as it is right now. It will happily wipe out fingerprints that are ordered "larger" than the new one added. In my testing I just happened to have the fingerprint of the second dive computer that I played with be the larger one, so it behaved correctly, but the code I submitted is garbage.

> One day, I should create a `sorted_vector` template that abstracts from this low-level stuff.

Right, because that's what I really need. A template that I don't understand around a data structure that I clearly don't understand, either.

core/device.cpp
bstoeger: I think both version are perfectly fine - this was just a FYI. [] would probably be a bit more idiomatic, since it doesn't usually result in an application termination on out-of-bound access.

dirkhh: well, I switched to that in the last force push

bstoeger:
> I clearly don't have a clue what I'm doing here. So using something that I know even less about doesn't seem an improvement. If anything, I think I should go back to writing this in C code that I actually understand, TBH.

I think you're doing fine - don't give up so easily. I believe this is the most efficient data structure for smallish tables with few updates. Once we reach a million of fingerprints, we might change to a hash-map. ;-) 

Please give me another day or two for review - I'm "busy" vacationing. :-P

dirkhh:
I'm leaving on a business trip today - take your time.

core/device.h
```diff
@@ -40,6 +41,12 @@ const char *device_get_nickname(const struct device *dev);
 extern struct device_table *alloc_device_table();
 extern void free_device_table(struct device_table *devices);
 
+// create fingerprint entry - raw data remains owned by caller
+extern void create_fingerprint_node(struct fingerprint_table *table, uint32_t model, uint32_t serial,
+				   const unsigned char *raw_data, unsigned int fsize, uint32_t fdeviceid, uint32_t fdiveid);
+// look up the fingerprint for model/serial - returns the number of bytes in the fingerprint; memory owned by the table
```
bstoeger, line 49: If the table owns the memory, the last argument probably should be `const unsigned char **`, no?

core/device.cpp
```diff
@@ -197,3 +199,56 @@ extern "C" void free_device_table(struct device_table *devices)
...
+// annoyingly, the Cressi Edy doesn't support a serial number (it's always 0), but still uses fingerprints
+// so we can't bail on the serial number being 0
+extern "C" unsigned int get_fingerprint_data(const struct fingerprint_table *table, uint32_t model, uint32_t serial, unsigned char **fp_out)
+{
+	if (model == 0 || fp_out == nullptr)
+		return 0;
+	struct fingerprint_record fpr = { model, serial };
+	auto it = std::lower_bound(table->fingerprints.begin(), table->fingerprints.end(), fpr);
+	if (it != table->fingerprints.end() && it->model == model && it->serial == serial) {
+		// std_lower gets us the first element that isn't smaller than what we are looking
```
bstoeger: std::lower_bound

core/device.h
```diff
@@ -59,11 +66,26 @@ struct device {
 	uint32_t deviceId;	// Always the string hash of the serialNumber
 };
 
+struct fingerprint_record {
+	bool operator<(const fingerprint_record &a) const;
+	uint32_t       model;     // model and libdivecomputer serial number to
+	uint32_t       serial;    //    look up the fingerprint
+	unsigned char *raw_data;  // fingerprint data as provided by libdivecomputer
+	unsigned int   fsize;     // size of raw fingerprint data
```
bstoeger, line 84: I'm not sure that `unsigned int` is the best data type for expressing sizes. One of the most annoying features of C++ (to me) is `size_t` being unsigned. You are consistent in that choice however, so probably best to keep it.

core/libdivecomputer.c
```diff
@@ -967,10 +967,20 @@ static void lookup_fingerprint(dc_device_t *device, device_data_t *devdata)
 {
 	char *cachename;
 	struct memblock mem;
+	unsigned char *raw_data;
```
bstoeger: `const unsigned char *` - the fingerprint is not modified, is it?

core/device.cpp
```diff
@@ -252,3 +252,54 @@ extern "C" void create_fingerprint_node(struct fingerprint_table *table, uint32_
...
+						const char *hex_data, uint32_t fdeviceid, uint32_t fdiveid)
+{
+	QByteArray raw = QByteArray::fromHex(hex_data);
+	create_fingerprint_node(table, model, serial,
+				(const unsigned char *)raw.constData(), raw.size(), fdeviceid, fdiveid);
+}
+
+extern "C" int nr_fingerprints(struct fingerprint_table *table)
+{
+	return table->fingerprints.size();
```
bstoeger, line 266: Probably have to cast to int to avoid a conversion warning.
`(int)...` or `static_cast<int>(...)`

core/device.cpp
```diff
@@ -252,3 +252,54 @@ extern "C" void create_fingerprint_node(struct fingerprint_table *table, uint32_
...
+}
+
+extern "C" uint32_t fp_get_model(struct fingerprint_table *table, int i)
+{
+	if (!table || i < 0 || i >= table->fingerprints.size())
```
bstoeger: Likewise cast needed here and below...

core/device.cpp
```diff
@@ -252,3 +252,54 @@ extern "C" void create_fingerprint_node(struct fingerprint_table *table, uint32_
...
+		return 0;
+	return table->fingerprints[i].model;
+}
+
+extern "C" uint32_t fp_get_serial(struct fingerprint_table *table, int i)
```
bstoeger: I'm not a fan of these verbose accessors for each field. No big deal, though.

core/load-git.c
```diff
@@ -1016,6 +1016,59 @@ static void parse_settings_divecomputerid(char *line, struct membuffer *str, str
...
+		fph->hex_data = value;
+		return;
+	}
+	report_error("Unknown fingerprint key/value pair (%s/%s)", key, value);
+}
+
+
+static void parse_settings_fingerprint(char *line, struct membuffer *str, struct git_parser_state *state)
+{
+	struct fingerprint_helper fph = { 0, 0, 0, 0};
```
bstoeger: space before '}'

bstoeger (approved):
To me it makes definitely more sense to save the fingerprints into the log than onto the device.

Only a few minor comments. Beware though, I nearly fell asleep when going through the parser code.

core/device.h
dirkhh: > If the table owns the memory, the last argument probably should be `const unsigned char **`, no?

I'm sure this is a silly question, but... why?
The caller owns the memory, so the caller passes in a pointer to it, and the `vector` creates its own copy.
What am I missing

core/device.cpp
dirkhh: right

core/device.h
dirkhh: Well, sizes are kinda non-negative... so why do you dislike that?

core/libdivecomputer.c
dirkhh: good catch

core/device.cpp
dirkhh: I don't think I got a warning... but it would make more sense to make this an `unsigned int` function, don't you think?

core/device.cpp
dirkhh: ok

dirkhh:
> To me it makes definitely more sense to save the fingerprints into the log than onto the device.
> 
> Only a few minor comments. Beware though, I nearly fell asleep when going through the parser code.

thanks for the review... I'll update the PR tomorrow. Brain is fried after 8 hours of Board meeting today...

dirkhh:
ok, I think I have all of @bstoeger 's comments addressed

core/device.h
dirkhh: never mind - understood and addressed

core/device.cpp
dirkhh: yeah, me neither... but with the C/C++ interface here... I couldn't come up with a better way that wouldn't seriously break our layering...

core/device.cpp
bstoeger: > I don't think I got a warning...

Yes, it seems one has to explicitly turn on `-Wconversion` to get a warning here.

> but it would make more sense to make this an unsigned int function, don't you think?

If you want to make it "really right", then this should return `size_t`, which is the standard type for lengths and indexes. On standard 64-bit x86, `unsigned int` is 4 bytes, whereas `size_t` is 8 bytes.

core/device.h
bstoeger: I dislike it, because unsigned integer arithmetic is hard to get right. Things like
```
        if (len1 - 2 < len2)
```
fail if `len1` happens to be 0 or 1, etc. Also once one uses `size_t` for lengths one logically uses `size_t` for indexes and then loops like
```
        for (size_t index = len - 1; index > 0; --index)
```
fail to work.

Moreover, one has to use awkward values such as `std::string::npos` or `((size_t)-1)` to indicate "invalid" or "out of bound".

Of course I understand that this comes from a time where 32-bit was the norm and people wanted to use vectors with more than 2^31 elements. However, now that 64-bit are the norm, it is just painful.

In my own projects, I consistently use `size_t` for lengths and indexes, since this is what the standard library does, but it is kind of annoying...

core/device.cpp
dirkhh: Agreed. Yet... That's ridiculous overkill in this situation.

dirkhh:
Any further concerns, or is this good to be merged?

dirkhh:
OK, this now has received a reasonable amount of testing (for some value of '_reasonable_') and I found another bug... if we already had a fingerprint in memory (i.e. from cloud storage) and tried to download from a dive computer, the in memory fingerprint got overwritten by an empty one and on a _second_ attempt to download it wouldn't match and we'd got back to finding no valid fingerprint.
Adding an extra check to only update the fingerprint table with valid data solved that problem.

But with that fix in place, I think this is indeed ready to merge.

PR #3347, opened 2021-11-13
mobile: correctly store manually entered GPS

dirkhh:
If there was no pre-existing dive site for a dive, manually entered GPS
information didn't get saved.

- [x] Bug fix

### Changes made:

ensure that we have a valid name for the dive site - if there's no given, use the coordinates
save the location with a newly created dive site

### Related issues:

a fair number of bug reports from Subsurface-mobile useres

### Mentions:

@bstoeger - can you please double check my logic?

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -850,6 +850,7 @@ static void setupDivesite(DiveSiteChange &res, struct dive *d, struct dive_site 
 		res.location = location;
 	} else {
 		res.createdDs.reset(alloc_dive_site_with_name(locationtext));
+		res.createdDs.get()->location = location;
 		d->dive_site = res.createdDs.get();
 	}
 	res.changed = true;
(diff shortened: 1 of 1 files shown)


mobile-widgets/qmlmanager.cpp
```diff
@@ -850,6 +850,7 @@ static void setupDivesite(DiveSiteChange &res, struct dive *d, struct dive_site
 		res.location = location;
 	} else {
 		res.createdDs.reset(alloc_dive_site_with_name(locationtext));
+		res.createdDs.get()->location = location;
```
bstoeger, line 853: You can remove the `get()` here. You can directly dereference a `unique_ptr` with `->` or `*`.

`get()` is only needed if you explicitly want to transform the smart pointer into a plain pointer.

bstoeger (approved):
I wonder whether we should search for close (say 10m?) dive sites with that GPS location. On desktop we list all dive sites up to a certain distance, so this could use the same code...?

dirkhh:
> I wonder whether we should search for close (say 10m?) dive sites with that GPS location. On desktop we list all dive sites up to a certain distance, so this could use the same code...?

I wonder about the UI for that, though ... How would this be presented to the user?

bstoeger:
After filling out the GPS coordinates and no dive site is selected, auto-complete with the closest dive site up to 10 or maybe 100 m?

dirkhh:
Ha - I got so busy with other stuff that I forgot to finish this one. Oops.

I will skip @bstoeger 's suggestion to autocomplete for now, but at least get the basics merged (after I remove the unnecessary `.get()`)

dirkhh:
merged locally after removing the `.get()`

PR #3349, opened 2021-11-20
[WIP] Try to fix hidden-cylinder mess

bstoeger:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Request: don't hide unused cylinders if there are cylinders after them.

Also, fix a few obvious bugs while doing so.

Attention: Cylinders are currently added at the end, which means that adding a cylinder may make numerous hidden cylinders appear. This is work in progress.

Attention 2: Completely untested.

TODO: Perhaps add a "n cylinders hidden" clue to the UI?

### Changes made:

1) Remove spurious cylinder-columns in planner.
2) Show intermediate hidden cylinders.

### Release note:

TODO.

### Documentation change:

TODO.

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1050,6 +1050,7 @@ AddCylinder::AddCylinder(bool currentDiveOnly) :
 	else
 		setText(Command::Base::tr("Add cylinder (%n dive(s))", "", dives.size()));
 	cyl = create_new_cylinder(dives[0]);
+	indexes.reserve(dives.size());
 }
 
 AddCylinder::~AddCylinder()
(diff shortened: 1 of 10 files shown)


bstoeger:
Note that this is marked WIP, but I need input: Does that fix the problem?

dirkhh:
Note that I'm not sure what you are fixing 😁

bstoeger:
That intermediate cylinders should be shown even if unused.

This is actually a hard one. E.g.:

Cylinder 1: used
Cylinder 2: unused
Cylinder 3: used

All three should be shown. Delete Cylinder 3.

Now only one should be shown.

Add a cylinder. Now you have

Cylinder 1: used
Cylinder 2: used
Cylinder 3: unused

and only two should be shown.

Undo should reverse all that without problems.

Etc, etc..

I still think that in the case of hidden cylinders, there should be a button: "n cylinders hidden [Show]".

dirkhh:
Actually - no. When you add a cylinder and have two use, one unused, until you quit Subsurface, that 3rd cylinder is shown even before you add a gas switch to it (because otherwise when you add a cylinder you'll be confused that it doesn't show up) - but that state gets forgotten if you quit and restart Subsurface.
That behavior may seem weird and illogical when talking about in the abstract, but when you use the app I content that it is the "obviously correct" thing to do.

dirkhh:
I don't disagree with your last statement about having a way to "unhide" hidden cylinders right there from the equipment tab (instead of having to know that this is a thing and having to go to the preferences)

bstoeger:
None of the actions I described above set the `manually_added` flag of the unused cylinder. So it should not be shown unless there is a used cylinder after it, no?

In an email you said that a new cylinder should be added before unused cylinders.

Combining these two facts, the course of events should be as described above, isn't it?

dirkhh:
> None of the actions I described above set the `manually_added` flag of the unused cylinder. So it should not be shown unless there is a used cylinder after it, no?

Umm... "add a cylinder" doesn't add the `manually_added` flag? That seems like a bug to me. It should.

**Edited to add** never mind... that's not what's happening in your example. You are magically adding a used cylinder (which is not a thing I know how to do)
 
> In an email you said that a new cylinder should be added before unused cylinders.

That seems more logical once we implement this change where unused cylinders in the middle are shown.
Without this change, no I didn't think that was the right thing to do and that's not what existing code did.

bstoeger:
> Edited to add never mind... that's not what's happening in your example. You are magically adding a used cylinder (which is not a thing I know how to do)

OK, replace "add" by "add and set the pressure".

Or replace "used" by "used or manually added".

> That seems more logical once we implement this change where unused cylinders in the middle are shown.
Without this change, no I didn't think that was the right thing to do and that's not what existing code did.

Well, that's exactly what this PR attempts to do (show unused middle cylinder and add cylinders before hidden cylinders) and which then should end in the above scenario.

bstoeger:
ping?

dirkhh:
I'm "under water", but unfortunately only in the figurative sense of the expression.
I had hoped to find time to work on this last weekend, but err... yeah... didn't happen.

I spent a few more minutes looking at this and I think it is better than what we have - which usually I consider sufficient to merge things...

bstoeger:
Note that this is pretty gnarly code and it needs some testing. Ideally, by the original reporter?

PR #3352, opened 2021-11-24
Webengineagain

atdotde:
This replaces all references to WebKit with equivalent code using WebEngine, in particular in printing.

The replacement is conditional on USE_WEBENGINE. This should be ON by default on all platforms where WebEngine is available (i.e. all but Windows).

In particular this should preserve backwards compatibility for templates.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh:
A quick glance through the code seems to indicate that the flow of changes is a bit... rough.
Would you mind if I restructured your commits a bit so we don't have the remove WebKit / reintroduce WebKit pattern? Or is this something that you would prefer to have in the history for some reason?

dirkhh:
Ahhh and this includes libdivecomputer changes... need to make sure I don't merge those by mistake.
No biggie, easy to fix by merging locally :)

atdotde:
Sorry about the libdivecomputer. This keeps happening to me (but usually I notice before pushing). 

Reintroducing WebKit was a separate commit because I only learned later that we still need it for Windows, that could easily be avoided by squashing commits 2 and 3. I only did not do it because the commits are already bigger than what I would have liked. This is due to the fact that all this took quite a bit of experimenting to make things work and there were no natural steps in between to commit (in particular with the desire to only commit states where a build works). Feel free to reshuffle.

I originally had the plan to have a big git rebase -i session to split this up into smaller, more meaningful commits. But today, also thanks to your comment I wanted to get this out of the door in a functional state. But I should say I even expected a "please split this in more meaningful commits" maintainer comment...

dirkhh:
I will play with this locally and will merge 2 and 3 and then look at the resulting commit - I think this will be pretty reasonable.

dirkhh:
OK, this has now been pushed at #3353 (rebased on latest master as well). So far barely tested, pushing that one to get the compile tests started and not waste time while I test this.
I have a few other things to get done today, but ideally I'd love to get this merged so we can move forward and get this in the hands of more people.

dirkhh:
this has been replaced by #3554 - let's work over there

PR #3353, opened 2021-11-24
Webengine

dirkhh:

- [x] Functional change
- [x] Build system change

alternative to #3352 - still all @atdotde 's code, just merged the first three commits to make the whole thing more logical
This results in a large first commit - but I find this more readable this way.

### Mentions:

@atdotde please let me know what you think about that approach...


dirkhh:
Oh, wait, these build tests aren't useful because `option(USE_WEBENGINE "Use QWebEngine instead of QWebKit" OFF)` so this isn't tested anywhere... Need to fix that :)

neolit123:
should i review this or @atdotde 's original PR?

dirkhh:
> should i review this or @atdotde 's original PR?

This one, please - it's the same code, but I think I'll continue to work on this unless he protests :)

dirkhh:
@atdotde 
Testing this on a Mac (after installing QtWebEngine which I didn't have before) I get this when I try to print a dive:
```
Could not find QtWebEngineProcess
Abort trap: 6
```
and things crash. Is this something that you've seen?

dirkhh:
for the Mac build to test WebEngine I need to create a new "pre-packaged" set of bits... that will take me longer than I have (gotta run now to an appointment)

atdotde:
Haven't seen that error. Anything useful from the stack trace?

desktop-widgets/printdialog.cpp
```diff
@@ -110,16 +117,21 @@ PrintDialog::PrintDialog(bool inPlanner, QWidget *parent) :
 	QPushButton *printButton = new QPushButton(tr("P&rint"));
 	connect(printButton, SIGNAL(clicked(bool)), this, SLOT(printClicked()));
 
+	// To do: Port preview to WebEngine. For the time being: Use OS print preview
+#ifndef USE_WEBENGINE
 	QPushButton *previewButton = new QPushButton(tr("&Preview"));
 	connect(previewButton, SIGNAL(clicked(bool)), this, SLOT(previewClicked()));
+#endif
```
neolit123, line 124: i recall some issues around print preview in Qt not being well supported across OS, which meant we had to add this explicit preview button.

i think the problem was mainly due to Windows, which makes this branch fine given we are not building with webengine on windows yet. would be nice to complete the TODO eventually to have the consistent UI across OSes.

desktop-widgets/printdialog.cpp
```diff
@@ -178,7 +190,7 @@ void PrintDialog::createPrinterObj()
 		qprinter = new QPrinter;
 		qprinter->setResolution(printOptions.resolution);
 		qprinter->setOrientation((QPrinter::Orientation)printOptions.landscape);
-		printer = new Printer(qprinter, printOptions, templateOptions, Printer::PRINT, inPlanner);
+		printer = new Printer(qprinter, printOptions, templateOptions, Printer::PRINT, inPlanner, Q_NULLPTR);
```
neolit123, line 193: we can omit the last null arg here, because the "parent" arg already has a default in the arg list of the constructor definition (both the webengine and old header)

atdotde:
I built this branch here locally and it seems to be a classic case of "works for me".

desktop-widgets/printdialog.cpp
```diff
@@ -215,12 +227,25 @@ void PrintDialog::printClicked(void)
...
+	connect(printer, &Printer::jobDone, this, &PrintDialog::printingDone);
+	printer->print();
+}
+
+void PrintDialog::printingDone()
+{
+	close();
+}
+
+#else
```
neolit123: i think having the macro start in the middle of printClicked is not very readable but this is just a visual nit.
ideally it should wrap the affected functions even if a couple of lines have to be duplicated across the macro branches.

also is it necessary to auto-close this dialog once done, i think it is manual in the old code?
i.e. do we need the printingDone slot? [1]

desktop-widgets/printdialog.h
```diff
@@ -38,6 +38,13 @@ private
 	void printClicked();
 	void onPaintRequested(QPrinter *);
 	void createPrinterObj();
+
+#ifdef USE_WEBENGINE
+public
+slots:
+	void printingDone();
```
neolit123, line 45: [1]

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include "core/statistics.h"
+#include "core/qthelper.h"
+#include "core/settings/qPrefDisplay.h"
+#include <stdio.h>
+
+#include <algorithm>
+#include <QPainter>
+#include <QPrinter>
+#include <QtWebEngineWidgets>
+#include "profile-widget/profilewidget2.h"
```
neolit123: not sure how consistent we are about header imports, but i think these should be:
- c sdlib
- c++ stdlib
- subsurface
- Qt

separated by new line.

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+	templateOptions(templateOptions),
+	printMode(printMode),
+	inPlanner(inPlanner),
+	done(0)
+{
+	webView = new QWebEngineView(parent);
+	connect(webView, &QWebEngineView::loadFinished, this, &Printer::onLoadFinished);
+		if (printMode == PRINT) {
+		connect(this, &Printer::profiles_inserted, this, &Printer::printing);
+	}
```
neolit123:
```suggestion
	if (printMode == PRINT) {
		connect(this, &Printer::profiles_inserted, this, &Printer::printing);
	}
```

desktop-widgets/printerwebengine.h
```diff
@@ -0,0 +1,60 @@
...
+	void print();
+	void previewOnePage();
+	QString writeTmpTemplate(const QString templtext);
+	QString exportHtml();
+	void updateOptions(print_options &poptions, template_options &toptions);
+
+
+signals:
+	void progessUpdated(int value);
+	void profiles_inserted();
```
neolit123:
```suggestion
	void profilesInserted();
```

this would require a change in a few places.

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+}
+
+Printer::~Printer()
+{
+	delete webView;
+}
+
+void Printer::onLoadFinished()
+{
+	if (profiles_missing)
```
neolit123: should start with `{` given the block is multiple lines

atdotde:
Just realized: It does not work with NO_USERMANUAL=ON since then webenginewidgets don't get included....

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+		webView->page()->runJavaScript("   var profiles = document.getElementsByClassName(\"diveProfile\");\
+				       for (let profile of profiles) { \
+					  var id = profile.attributes.getNamedItem(\"Id\").value; \
+					  var img = document.createElement(\"img\"); \
+					  img.src = id + \".png\"; \
+					  img.style.height = \"100%\"; \
+					  img.style.width = \"100%\"; \
+					  profile.appendChild(img); \
+					} \
+				", [this](const QVariant &v) { emit profiles_inserted(); });
```
neolit123: style/readability nit: this can be something like:

```suggestion
		QString script = QStringLiteral(" \
			var profiles = document.getElementsByClassName(\"diveProfile\"); \
			for (let profile of profiles) { \
		  		var id = profile.attributes.getNamedItem(\"Id\").value; \
		  		var img = document.createElement(\"img\"); \
		  		img.src = id + \".png\"; \
			  	img.style.height = \"100%\"; \
			  	img.style.width = \"100%\"; \
			  	profile.appendChild(img); \
			}")
		webView->page()->runJavaScript(script, [this](const QVariant &v) { emit profiles_inserted(); });
```

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+	profiles_missing = false;
+	emit(progessUpdated(100));
+}
+
+void Printer::printing()
+{
+	QPrintDialog printDialog(&printer, (QWidget *) nullptr);
```
neolit123:
```suggestion
	QPrintDialog printDialog(&printer, (QWidget *) Q_NULLPTR);
```

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+	if (printDialog.exec() == QDialog::Accepted) {
+		webView->page()->print(&printer, [this](bool ok){ if (ok) emit jobDone(); });
+	}
```
neolit123:
```suggestion
	if (printDialog.exec() == QDialog::Accepted)
		webView->page()->print(&printer, [this](bool ok){ if (ok) emit jobDone(); });
```

still not a fan of the C++ lambdas, i must admit.

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+//value: ranges from 0 : 100 and shows the progress of the templating engine
+void Printer::templateProgessUpdated(int value)
+{
+	done = value / 5; //template progess if 1/5 of total work
+	emit progessUpdated(done);
+}
+
+QString Printer::exportHtml()
+{
+	// Does anybody actually use this? It will not contian profile images!!!
```
neolit123: there is a typo in `contain`

response: i don't remember the state of this. does the old code with webkit include images?

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+	templateOptions = toptions;
+	printOptions = poptions;
+	profiles_missing = true;
+}
+
+QString Printer::writeTmpTemplate(const QString templtext)
+{
+	QFile fd(printDir.filePath("template.html"));
+
```
neolit123, line 98:
```suggestion
```

desktop-widgets/printerwebengine.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+	fd.open(QIODevice::WriteOnly | QIODevice::Text);
+	QTextStream out(&fd);
+	out << templtext;
+	fd.close();
+	return fd.fileName();
+}
+
+void Printer::print()
```
neolit123, line 107: ideally until all OSes move to webengine, it might be a wise thing to make this function reusable between the two printer*.cpp files to avoid potential drift.

that of course is not that simple due to the emit() calls from the underlying QObject.

desktop-widgets/templateedit.cpp
```diff
@@ -43,7 +50,20 @@ TemplateEdit::TemplateEdit(QWidget *parent, const print_options &printOptions, t
...
+	preview = new Printer(nullptr, printOptions, newTemplateOptions, Printer::PREVIEW, false, this);
+	view = preview->webView;
+	int w = ui->label->width();
+	int h = ui->label->height();
+	view->resize(w, h);
+	preview->print();
+	view->setZoomFactor(0.25);
+	view->show();
+	ui->verticalLayout_3->replaceWidget(ui->label, view);
+#else
```
neolit123, line 64: why not have this in updatePreview?
it seems that webengine requires "preview" to persist as a pointer for some reason, while for the webkit case we can print the result to a pixmap for a Printer object on the stack. [3]

desktop-widgets/templateedit.h
```diff
@@ -29,13 +32,15 @@ private slots:
...
-	const print_options &printOptions;
+	print_options &printOptions;
 	template_options &templateOptions;
 	struct template_options newTemplateOptions;
 	QString grantlee_template;
 	void saveSettings();
 	void updatePreview();
-
+#ifdef USE_WEBENGINE
+	Printer *preview;
```
neolit123, line 42: [3]

desktop-widgets/templatelayout.cpp
```diff
@@ -152,7 +152,12 @@ QString TemplateLayout::generateStatistics()
 
 QString TemplateLayout::readTemplate(QString template_name)
 {
-	QFile qfile(getPrintingTemplatePathUser() + QDir::separator() + template_name);
+	QString filename;
+	if (template_name.contains("template.html"))
+		filename = template_name;
+	else
+		filename = getPrintingTemplatePathUser() + QDir::separator() + template_name;
+	QFile qfile(filename);
```
neolit123: could you explain why was this change needed?

desktop-widgets/usermanual.cpp
```diff
@@ -35,6 +35,27 @@ void SearchBar::enableButtons(const QString &s)
 	ui.findNext->setEnabled(s.length());
 }
 
+#ifdef USE_WEBENGINE
+MyQWebEnginePage::MyQWebEnginePage(QObject* parent) : QWebEnginePage(parent)
```
neolit123, line 39:
```suggestion
WebEnginePage::WebEnginePage(QObject* parent) : QWebEnginePage(parent)
```

naming is consistent with other inherited object cases.

desktop-widgets/usermanual.cpp
```diff
@@ -35,6 +35,27 @@ void SearchBar::enableButtons(const QString &s)
...
+{
+	if (type == QWebEnginePage::NavigationTypeLinkClicked)
+	{
+		QDesktopServices::openUrl(url);
+		return false;
+	}
+	return true;
+}
+
+
```
neolit123:
```suggestion
```

desktop-widgets/usermanual.cpp
```diff
@@ -97,6 +128,14 @@ UserManual::UserManual(QWidget *parent) : QDialog(parent)
 	setLayout(vboxLayout);
 }
 
+#ifdef USE_WEBENGINE
+void UserManual::search(QString text, QWebEnginePage::FindFlags flags = QFlag(0))
+{
+	userManual->findText(text, flags,
+			     [this, text](bool found) {searchBar->setStyleSheet(found || text.length() == 0 ? "" : "QLineEdit{background: red;}");});
```
neolit123, line 134: worth a comment explaining how this new search works.

neolit123 (commented):
thanks for the efforts @atdotde
this looks mostly OK to me.

my comments are mostly around style nits, which can be ignored for the initial PR.
a couple of more significant questions are around preview and related buttons.

also did you manage to resolve the empty page problem?

desktop-widgets/templatelayout.cpp
atdotde: I changed the semantics of the "Apply" button to: Update the preview to the current version of the template in the editor. (before Apply forced a "save"). So, unless "OK" is clicked, it should use a temporary template (the one form the editor window) which is not stored in the template directory but in the temporary directory of the printing job. That comes with an absolute file name.

desktop-widgets/printdialog.cpp
atdotde: The #ifdef formatting is clearly bad. About the connect: Will have to look into this: The thing is that rendering in WebEngine in not synchronous, so you have to listen to the completed signal before cleaning things up. But together with the preview, this might need more consideration.

desktop-widgets/templateedit.cpp
atdotde: The whole pain in this was coming form the fact that a QWebEngineView uses GL and thus cannot render to a QPixmap. The preview widget is now a full blown web browser that in Printer::print is told to render a local file (and upon Apply to render a new file).

neolit123: i think offscreen rendering should be still possible. according to this post it is a matter of syncing the loaded page (onViewLoaded):
https://stackoverflow.com/questions/53524652/is-it-possible-to-render-qwebenginepage-qwebengineview-offscreen

the QPainter target in that case is QImage but can also be a QPixmap.

desktop-widgets/templatelayout.cpp
neolit123: this seems fine, having a comment above might be easier for the readers to parse.
unless i'm mistaken it can also be `template_name == "template.html"`

scripts/build.sh
```diff
@@ -546,7 +546,7 @@ for (( i=0 ; i < ${#BUILDS[@]} ; i++ )) ; do
 	else
 		EXTRA_OPTS="-DNO_USERMANUAL=ON -DNO_PRINTING=ON"
 	fi
-	if [ "$SUBSURFACE_EXECUTABLE" = "DesktopExecutable" ] && [ "$BUILD_WITH_WEBENGINE" = "1" ]); then
+	if [ "$SUBSURFACE_EXECUTABLE" = "DesktopExecutable" ] && [ "$BUILD_WITH_WEBENGINE" = "1" ]; then
```
dirkhh, line 549: I hate it when I do stupid things like this.
I tested it, then did a "quick edit" and committed crud like this. Thanks for the fix... I'll rebase this of course

dirkhh:
> I built this branch here locally and it seems to be a classic case of "works for me".

Yeah, nothing from the stack trace. I'll play with it some more.
Which Qt version do you build against?

dirkhh:
> Just realized: It does not work with NO_USERMANUAL=ON since then webenginewidgets don't get included....

Ahhh, ok.
Of course if we do that, then that whole expression that you changed is pointless as we just overwrite what was set, anyway. So let's rewrite this.
More testing and thinking needed :)

desktop-widgets/printdialog.cpp
dirkhh: admittedly there is the question of "shouldn't print preview be an OS function in the first place?"
I find it really weird that every app should have to implement this, instead of having it in the OS print dialog. This seems to work very well on macOS, but simply isn't there on many Linux flavors (not sure - maybe it works on some but standard KDE/Plasma doesn't appear to have a print preview as part of the print dialog).
So yeah, unfortunately we might need this.

dirkhh:
Hmm. So it appears that deploying WebEngine needs some extra TLC.
As is so frustratingly common, the official documentation is less helpful than one would hope: https://doc.qt.io/qt-5/qtwebengine-deploying.html
So `macdeployqt` should do the right thing? Maybe?
Well for me (macOS 10.15 with Qt 5.15.1) it doesn't. And that required `QtWebEngineProcess` doesn't appear to get installed at all. Which makes me really puzzled how this works for you, Robert.
What do you get if you do a `find . -name QtWebEngineProcess` in your build folder?

I have things at least building "correctly" in different places (knock on wood). But when I run this on Linux, the print doesn't contain a profile... and has frighteningly low DPI. This is a full resolution screenshot of a 100% preview of a dive...
Screen Shot 2021-11-24 at 4 33 30 PM
..

atdotde:
Good morning!

So let’s see why it works for me but not for you.

My laptop happens to have Qt 5.14.0 at the moment. So that’s what I am building against.

> On 25. Nov 2021, at 01:34, Dirk Hohndel ***@***.*** ***@***.***>> wrote:
> 
> Hmm. So it appears that deploying WebEngine needs some extra TLC.
> As is so frustratingly common, the official documentation is less helpful than one would hope: https://doc.qt.io/qt-5/qtwebengine-deploying.html <https://doc.qt.io/qt-5/qtwebengine-deploying.html>
> So macdeployqt should do the right thing? Maybe?
> Well for me (macOS 10.15 with Qt 5.15.1) it doesn't. And that required QtWebEngineProcess doesn't appear to get installed at all. Which makes me really puzzled how this works for you, Robert.
> What do you get if you do a find . -name QtWebEngineProcess in your build folder?
> 
> 
 find . -name QtWebEngineProcess
./Subsurface.app/Contents/Frameworks/QtWebEngineCore.framework/Versions/5/Helpers/QtWebEngineProcess.app/Contents/MacOS/QtWebEngineProcess

It seems to come from

/Users/Helling/Qt/5.14.0/clang_64/lib/QtWebEngineCore.framework/Versions/5/Helpers/QtWebEngineProcess.app/Contents/MacOS/QtWebEngineProcess

I should also say, that the macdeployqt process has been broken for me for a while. So I build using the following script:

#!/bin/bash

make -j2 install
install_name_tool -add_rpath /Users/Helling/Qt/5.14.0/clang_64/lib/Subsurface.app/Contents/MacOS/Subsurface

cp ../../googlemaps/build/libqtgeoservices_googlemaps.dylib Subsurface.app/Contents/PlugIns/geoservices/


> I have things at least building "correctly" in different places (knock on wood). But when I run this on Linux, the print doesn't contain a profile... and has frighteningly low DPI. This is a full resolution screenshot of a 100% preview of a dive...
> 
> 
That’s a bummer. I can see two steps that fail adding the images: First writing the profiles as png’s to the temporary directory and then the javascript snippet running that injects those in the html. Maybe you can add some debugging info to see if those steps actually run or have a peek at the temporary directory to see if they are there?

Best
Robert

desktop-widgets/templatelayout.cpp
atdotde: I added a comment and used a less generic 'magic' filename. I had thought about detecting a directory separator as the actual templates are just simple filenames while the temporary one has an absolute path but wanted to avoid a possible / vs \ confusion (afaik Qt even on windows can use / but I don't know if it actively does the conversion, i.e. if the automatically generated file has a / or a \).

desktop-widgets/printdialog.cpp
atdotde: I would have thought so but I get a compiler error, so I left it in.

desktop-widgets/printdialog.h
atdotde: Don't understand the comment. Only the web engine version has a public slot.

desktop-widgets/templateedit.h
atdotde: Please elaborate

desktop-widgets/printdialog.cpp
neolit123: Odd, i guess it should be included in that case.

desktop-widgets/printdialog.h
neolit123: This is related to https://github.com/subsurface/subsurface/pull/3353#discussion_r756398837

It seems like we are auto closing the dialog now on done with this slot. I think it was manual in the old version.

desktop-widgets/printerwebengine.cpp
neolit123: To clarify this. We can leave it out of this PR but when making changes we have make sure both engine functions are updated.

desktop-widgets/templateedit.h
neolit123: This can be removed if we manage to resolve https://github.com/subsurface/subsurface/pull/3353#discussion_r756419031

desktop-widgets/templatelayout.cpp
neolit123: Widows should be fine with / but qt recommend using QDir::separator()

desktop-widgets/printdialog.h
atdotde: No, it was auto-closed before as well but this delays the close until the rendering is done.

desktop-widgets/printerwebengine.h
atdotde: Done

desktop-widgets/printerwebengine.cpp
atdotde: Done.

desktop-widgets/printerwebengine.cpp
atdotde: Done

desktop-widgets/printerwebengine.cpp
atdotde: Done

desktop-widgets/templateedit.cpp
atdotde: Thinking about this a bit more, I am now convinced that the whole QPixmap/QImage whatever thing is an unnecessary indirection and not having it but having the actual rendering thing is an advantage.

desktop-widgets/templateedit.h
atdotde: See above, I think this is better with a QWebEngineView directly.

desktop-widgets/templatelayout.cpp
atdotde: My understanding is that Qt accepts both / and \, but we would need the other direction: are we guaranteed that an absolute path returned by QFile::fileName() will contain QDir::separator() and not possibly another character instead? That was my worry.

neolit123: >  are we guaranteed that an absolute path returned by QFile::fileName() will contain QDir::separator() and not possibly another character instead?

it would return the right separator for the OS.

desktop-widgets/templateedit.cpp
neolit123: in the original, i think the idea is to render to a bitmap offscreen and to not show the QWebView widget in the dialog, because it would make the preview interactive in a weird way.
i think we should avoid that in the new version as well either use a bitmap again or lock the widget for clicks and key presses entirely.

performance wise, showing the full engine widget on screen might be slow, but rendering to bitmap offscreen is also slow, so not clear what is the winner in that regard.

my vote goes for using a bitmap, as that is the method i've seen elsewhere, but leaving the decision to your best judgement.

desktop-widgets/printer.cpp
```diff
@@ -14,7 +14,7 @@
 #include <QWebElement>
 #include "profile-widget/profilewidget2.h"
 
-Printer::Printer(QPaintDevice *paintDevice, const print_options &printOptions, const template_options &templateOptions, PrintMode printMode, bool inPlanner) :
+Printer::Printer(QPaintDevice *paintDevice, const print_options &printOptions, const template_options &templateOptions, PrintMode printMode, bool inPlanner, QWidget *parent = Q_NULLPTR) :
```
bstoeger, line 17: Please don't reintroduce `Q_NULLPTR`. It never made sense. Use `nullptr`, 0 or, if you insist on using a macro, `NULL`.

dirkhh:
so yesterday was the high holiday of gluttony here in the US (also sometimes called by its olden name "Thanksgiving"). And as a result of that I didn't spend any time at my computer.
Today is the high holiday of consumerism and buying things you don't need for allegedly cheap - even though everyone knows that they will be far cheaper at the end of December... Since I typically don't partake in THAT particular insanity, I think I might have time to spend on Subsurface today. Of course, if I look at all the things that I **should** work on with respect to that... uh oh.
I think goal #1 will be to get this PR to work on at least one of my systems. Right now it builds most places locally, but it works on none of my systems in a reasonable manner - that's not meant as criticism, that's an explanation of where I am right now. Your explanation how this works on your mac not using our existing install system wasn't really as helpful as one might hope because that doesn't translate into something that I can ship to users. Let's see what I can figure out.
Also, I saw that you did something cruel to the git history, so I'll need to clean that up as well.

Are you actively working on this as well, @atdotde ?

dirkhh:
Trying to rebase the changes that you tacked on at the end. Ugh. Git is such a powerful tool. And can be such a pain.
Since you didn't respond, @atdotde , I'll assume that you aren't currently working on this branch, so I'll do the cleanup now... hopefully I'll only have to do this once :)

dirkhh:
OK, created a new, once again clean branch and a new PR for that as #3554
Now I can start to actually work on what I had hoped to work on, i.e. making this thing, err, do something...

dirkhh:
closing this as it is replaced by #3554

PR #3354, opened 2021-11-26
Webengine clean

dirkhh:

- [x] Functional change
- [x] Code cleanup
- [x] Build system change

yet another redo of #3352 and #3353
It feels like I'm not really making progress on the issue at hand, just trying to give us clean PRs to work on top of.

PLEASE PLEASE PLEASE... @atdotde or anyone else working on this. Don't just merge this one into your current working directory and add on top of it. Check this out as a new clean branch and work on top of it. It is frustratingly painful to try to clean these up... this one should have all of @neolit123 and @bstoeger 's feedback implemented (I think)
If not, please holler.


dirkhh:
I think I need help here.
This works fine on macOS.
But on Linux only some of the profiles are printed.
I instrumented this code quite a bit to try to understand what's going wrong.
- we are indeed writing all the rendered profiles into the temporary folder that's created
- we are finished with the print() call on the page before that temporary folder is deleted
- writing out the post-processed HTML (so after we use JS to fill in the profile filenames) and opening this in Firefox shows all the profiles

yet in single dive mode there is no profile on Linux, and when printing multiple dives the first N (sometimes 3, sometimes a few more) have a profile, and after that no profiles are printed
I'm at a loss.
@bstoeger - would you be willing to poke at this?

atdotde:
Sounds like a race condition. Could it be that after the JS updates the html, the engine does not have enough time to render the images before the page is sent to the printer? Does the number of images rendered change when a somewhat long sleep is inserted before everything else in Printer::printing()?

atdotde:
Here's an idea: Maybe emitting profilesInserted() from the callback of runJavaScript is wrong. Maybe it should only be emitted when the page is actually loaded again (now with the images). Could you please try changing Printer::onLoadFinished() so that the callback is empty and the if(profilesMissing) gets an else clause where  profilesInserted() is emitted?

atdotde:
Please see https://github.com/subsurface/subsurface/pull/3355

dirkhh:
Oh my. So much NOT FUN

This seems to work on openSUSE. After four hours of beating my head against this I ran out of steam for today. This isn't tested AT ALL on other Linux or on macOS. I don't see why it would fail... but, yeah, right...

This fixes a couple of problems in @atdotde 's code that became reasonably obvious once I stared at that code for a while.
There were two identical `connect()` calls which caused `onLoadFinished()` to be called twice. And since `profilesMissing` was set to `false` not in the callback but at the end of the function invocation (remember, the JS runs asynchronously) I was sure that this would fix it. But no... it didn't.

So I came up with this insane idea to write the modified HTML out and load it again. And then print THAT. Which is stupid. But... it works.

So I hate the code - it clearly is way too noisy and needs to have a ton of debug stuff removed. But given that it works I wanted to push it into the branch so others could comment.

Yes, it's insane. I get that. I'll try again tomorrow with a fresh brain to avoid the ridiculous writing out of the intermediate step. There is no excuse for doing that. But in the meantime, I'd love to get some reviews on what's there :)

dirkhh:
OK, spent some more time on this. Regardless what I try - without writing the new HTML to disk and re-loading it, I cannot get openSUSE to render all of the profiles.
This branch has now been tested on Ubuntu as well and appears to work there.
It has also been tested on macOS 10.15 - strangely, the OS level preview doesn't work there for me (and I know that it works for @atdotde )
I still get the empty page at the end, but other than that this seems to be working as designed by Robert.
I know it's early evening in Europe - maybe a couple of people could look at the code or even test this

bstoeger:
Can we not simply(?) replace the `<div>`-tags by ``-tags in `desktop-widgets/templatelayout.cpp`? This already does some textual manipulations, no?

My other suggestion would be to kindly ask the people who are responsible for this completely insane[1] API to tell us how to wait for image loading in Javascript.

[1] I insist that having to inject JS to access/modify objects is crazy. Sorry for my negativity.

dirkhh:
Well, these are reasonable comments, @bstoeger 
Here are my thoughts. Users have modified templates. Their expectations will be that the templates continue to work.
There is no requirement that the templates are formatted exactly like ours. What the existing code does is that it injects the image into a container of a certain class and extracts the id element from that class to form the name of the image that we load. HTML gives you quite a few ways to structure such a container. The class and id can be in either order. There could be other things in that container. Etc.
Sure, I could write code in C++ that would hopefully work for all the sane cases. Or I could simply continue to run the same Javascript code which is designed to work on the object tree parsed from the HTML.
To me that seemed like the much, much, much smarter option.
Do I like it? no. Do I wish I could go back and talk to the GSOC student and the mentor and the maintainer who all thought this was a sane design? Sure, absolutely.

But I live in a reality where that isn't an option. And I have a strong preference for not breaking things I don't need to break.

I am certain that an experienced WebView developer would say "oh, all you need to do is call this API and do this other things in the call back and it's all good"... but I haven't found that. And while this workaround is stupid and wasteful, it's also straight forward and easy to debug. And because it uses the same JS code I am quite certain that it won't break user templates.

So to me the question was simple - how much time do I want to spend writing what is essentially a partial HTML parser in C++ in order to inject the stupid pictures. Or how much time do I want to spend on something ugly that at least does what people want.

And, reminder. I do not print logbooks. I wrote the original printing code (which was then completely changed by the GSOC student) because I was told that a dive base that Linus and I were going to go to required paper log books. When we arrived with our printouts they laughed at us and said "who told you that... we haven't done that in something like a decade!" (I pointed out that it was still on their website). That was in 2011. In the ten years since I have seen ZERO references to needing a printed log. And have ZERO times printed my log that weren't in order to debug this code for the few users who want to have paper logs... ironically every single user that I interact with who asks about the printing code is German-speaking...

bstoeger:
> Do I like it? no. Do I wish I could go back and talk to the GSOC student and the mentor and the maintainer who all thought this was a sane design? Sure, absolutely.

Actually, I think the design is very much sensible and they did a good job. A `<div>` block is a general HTML place-holder for whatever you want. Maybe we don't want to put in an image, but an image with a border or a title or a vector graphic or whatever. Hardcoding an `` tag would have shown little foresight.

What is not sensible is an HTML-render API where you cannot programmatically access objects in the object tree! Suppose you want to do something more complex than a simple search / replace. Now you have to write a Javascript generator or better yet: replace the whole application by JS.  Are you kidding me? What makes it worse is that this is a massive regression over what we had.

In this case "our" design is not the culprit. Here, clearly Qt is garbage.

That said, why can't we wait in Javascript for the image to load? I quickly tried with promises, but it didn't work. Therefore my suggestion that the geniuses who decided to abandon a sane API show us how to do this according to their new ways.

dirkhh:
Yeah, so I went through the docs with a fine tooth comb. And I still don't understand WHY only the first few images get loaded. The whole thing is broken IMHO. But heck if I can find an API that would do this. The callback of the JS invocation only says that that invocation has completed (so all the img's have been inserted). But why isn't there a way to wait for them to render? `QWebEnginePage::loadFinished()` explicitly says that it's not about the rendering being finished. But there is no other `...Finished()` signal that seems useful. `findTextFinished()` - yeah, not useful.
The `print()` method explanation says in the first sentence: **Renders the current content of the page into a temporary PDF document, then prints it using printer.**
Silly me for assuming that it would render the **current content** of the page, after JS modifications to the DOM.

And it was that realization that made me simply say "screw this" and write out the intermediate version - and hooray, that worked.

bstoeger:
> But why isn't there a way to wait for them to render?

In theory there is. You can listen to image.onload() with a promise:
https://stackoverflow.com/a/66180709

However, this didn't work for me. I'm not sure whether that was because I did something wrong (it didn't produce and error messages but still showed only the first few images) or because this whole thing is - as you note - broken.

You will find other solutions like counting the onload() events until it reaches the number of pictures, but that sounds like a recipe for disaster.

dirkhh:
What on earth... counting the `onload()` events? That's... genius.
I would have never come up with it. Also, that's definitely not what I think we should do instead. That makes my hack seem outright sane in comparison.

atdotde:
I am sure there must be a way to tell that the images have loaded. In the worst case, we can do that in JS as well (apparently). If you want to manipulate the html before sending it to the engine, that should probably be done using xslt as that should get the parsing under control. But I have the suspicion that an xml parser is more picky about incorrect syntax than your run of the mill html parser and remember, we are working on user supplied files.

And of course there is always the possibility that we write a GUI toolkit that has a memory model that holds up to modern standards and implement an html renderer in it that allows access to elements of the document tree from a sane language. 

But before we start that pleas give me a little more time to try things (it already took a substantial time of experimentation to get to this stage).

dirkhh:
> And of course there is always the possibility that we write a GUI toolkit that has a memory model that holds up to modern standards and implement an html renderer in it that allows access to elements of the document tree from a sane language.

Err - please, NO?
 
> But before we start that pleas give me a little more time to try things (it already took a substantial time of experimentation to get to this stage).

So far my plan is to continue down the road that this PR is on. It seems to work. I can understand it and argue about it and most importantly debug it.
I have no concern giving you more time to work on this PR if you prefer.
@bstoeger - should we merge your profile work first then? I can switch my review efforts to that.

atdotde:
To see if we can do it in JS, maybe it's possible after the profile.appendChild(img); to add a while(! img.complete){;}

That should wait in JS for the image to render. Of course, this makes things much slower and a better way would be to first add the images and then check they all loaded completely, but as a first step...

bstoeger:
> should we merge your profile work first then? I can switch my review efforts to that.

I leave that decision up to you. If you have the necessary time, I would of course appreciate it, because the growing gap between master and the profile branch starts to make me nervous. Note that I want to completely redo the animations and then start to QtQuick-ify it, so there is quite some work ahead.

I feel that profile feels "snappier", but that might be purely subjective and I didn't run any benchmark. Could you confirm this?

> To see if we can do it in JS, maybe it's possible after the profile.appendChild(img); to add a while(! img.complete){;}

For testing purposes - OK. As a final solution, busy-waiting is not acceptable (and I predict that it will end in infinite loops, so you would have to implement timeouts etc.). Frankly, I would prefer Dirk's funky write/reload to a busy wait. And that is a low bar. :-P

Note that I got the `onload()` JS events and yet the pictures where only half rendered. There seems to be weird things going on.

bstoeger:
PS: The "print preview" really should not be removed on Linux. At least on Kubuntu this doesn't work and it made testing significantly more painful to me.

Also I had weird UI issues: first I press print, then it takes ages until the OS dialog appears, then it again does something for quite a long time while the Qt dialog is responsive and then the Qt dialog suddenly disappears. This all feels very rough.

dirkhh:
Yeah, agreed on missing the print preview

I did not see the dialog confusion that you mention. Which Linux flavor are you running? What desktop environment?
I tried this with plain Ubuntu and Plasma on openSUSE...

bstoeger:
Sorry, I didn't express myself clearly. There is no confusion with dialogs, just with modality. Once I press print in the OS dialog, I can change things in the print dialog as it seems to print the HTML in the background. And when it has done that, the dialog disappears. It's not a big deal, just a strange experience. Either close the dialog when once printing is in progress or let it open even after printing. But don't let it disappear while the user is changing things.

PR #3355, opened 2021-11-27
Wait for profile images to render

atdotde:
Waiting for JS to finish is not enough, we need to wait for
the page to render the second time.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/desktop-widgets/printerwebengine.cpp
+++ b/desktop-widgets/printerwebengine.cpp
@@ -46,8 +46,10 @@ void Printer::onLoadFinished()
 					  img.style.width = \"100%\"; \
 					  profile.appendChild(img); \
 					} \
-				", [this](const QVariant &v) { emit profilesInserted(); });
+				", [](const QVariant &v) {;});
 
+	} else {
+		emit profilesInserted();
 	}
 	profilesMissing = false;
 	emit(progessUpdated(100));


desktop-widgets/printerwebengine.cpp
```diff
@@ -46,8 +46,10 @@ void Printer::onLoadFinished()
 					  img.style.width = \"100%\"; \
 					  profile.appendChild(img); \
 					} \
-				", [this](const QVariant &v) { emit profilesInserted(); });
+				", [](const QVariant &v) {;});
```
bstoeger, line 49: ? Why would you pass a callback that does nothing ?
There should be a non-callback version of that call - if the API forces you to use a callback I would think that it is crucial to do something in there.

desktop-widgets/printerwebengine.cpp
```diff
@@ -46,8 +46,10 @@ void Printer::onLoadFinished()
...
 
+	} else {
+		emit profilesInserted();
```
bstoeger, line 52: I have not looked at the code, but this function looks ominous.

You say that after the call above the profiles are not rendered. Yet, you set `profilesMissing` to false. Seems like a wrongly named flag?

Also, this seems to imply that `onLoadFinished()` is called multiple times - please add a code-comment detailing the control flow. The casual reader will be confused.

From a quick glance I only see that there are multiple(!) `connect()` calls, but no corresponding `disconnect()`s. What is going on there? Needs comments.

I don't think I will find time to look at this today, but for now, out of caution, I would NAK this. Needs more explanation of what is going on.

atdotde: Let's see first if this solves the problems @dirkhh encountered on Linux. This is a call back that is run whenever the web view finished loading a page. The first time (then profilesMissing is still true) it is called once the print.html file is read from disk. This time, we need to invoke the JS that adds the img tags to the page (as those are, for historical reasons, missing from the template). This modified page is then rendered once more, now with the profile images and after that finished, the call back is called once more, this time with profilesMissing==false. This is then time to ship out to the printer. If this is really the solution, I will add comments explaining this.

bstoeger: What about the two `connect()` statements? Seems like a recipe for disaster: wouldn't the callback be called twice, resulting in timing issues?

And why on earth do we need JavaScript to render images? I stopped wondering about Qt-APIs, but that is ... very strange.

Moreover, I wonder why does executing a JavaScript on a page result in an `onLoadFinished()` signal? Is that per image or at the end of the script?

atdotde:
The double connect is wrong and a left over of having to try too many things, will fix that.

Re why JS: One design goal was not to change the syntax of the templates so users that have their own templates don't have to change those. And as it happens, the old templates don't have image tags. Have a look at the old code: It parses the html, then wanders through the document tree to replace the div tags of type "diveProfile" by new image tags. This wandering through the document tree in C++ is no longer supported for WebEngine (check the transition document) and you are supposed to do that in JS. So that's what we do. An alternative might be to do this replacement while the html is still a QString that could be manipulated using RegExes. But I would think that that is much more fragile (remember: We are working on potentially user supplied template files) but also less popular. 

Would I have designed it this way? Hell no. But this is legacy that we have to live with.

OK, so now the JS injects the image tags into the document. The runJavaScript callback is called once that script finishes (i.e. after all image tags are inserted). But now, the web engine has to parse this new document, find the image tags and load the image files and insert those into the rendered page. Once that is done (the modified page is loaded) onLoadFinished() fires again and (I think at least this is the origin of the race condition Dirk sees) only now we are ready to ship out the page to the printer.

bstoeger:
> This wandering through the document tree in C++ is no longer supported for WebEngine (check the transition document) and you are supposed to do that in JS. So that's what we do.

These are the moments it's hard to not detest the Qt API. Keep the absurdly broken 1980s lifetime management and force us to use JS. Gee, thanks. Please hire competent developers.

In any case, I compiled this (even though currently I have 99 other problems - sigh). I tried to print - I felt that it was significantly slower than the old code, to the point of being unusable. I have a few more dives than last time I tried to print, but I'm not sure that can the explain (subjective) slow down. For some reason the printed document was only one page, even though I didn't select "print only selected dives". I tried to print again and the application crashed.

Edit: But FWIW, the single printed page had a profile...

bstoeger:
Second attempt: I removed the second `connect()` call - the OS print dialog never appears.

The `onLoadFinished()` *is* called though, but only once.

This makes me wonder: Could it be that `onLoadFinished()` was called twice for you, because you had two connects? And that gave these timing-issues?

dirkhh:
unfortunately today is once again a day filled with family fun. I am about to leave the house (just dealt with support emails) and am not sure when I'll have time today to sit down and play with this. I thought I had tried this one - but it's possible that there are more ways to check that the render is complete before printing. 
BTW, Robert - that was exactly my suspicion as well... 
and, Berthold - who the heck comes up with an API that allows the `print()` call to run before the underlying page is fully rendered???? these libraries baffle me so much...

atdotde:
Excuse me, but the bad design are the templates laking the profile image tags.

atdotde:
> On 27. Nov 2021, at 18:26, Dirk Hohndel ***@***.***> wrote:
> 
> unfortunately today is once again a day filled with family fun. I am about to leave the house (just dealt with support emails) and am not sure when I'll have time today to sit down and play with this. I thought I had tried this one - but it's possible that there are more ways to check that the render is complete before printing.
> 
Take your time. I also have other things to do at the moment.

> BTW, Robert - that was exactly my suspicion as well...
> and, Berthold - who the heck comes up with an API that allows the print() call to run before the underlying page is fully rendered???? these libraries baffle me so much…
> 
> 
Just remember, the WebEngine is not there in the first place to render stuff for printing but to have an embedded web browser (after all, it’s basically chrome). And I would guess, todays webpages in some sense never finish rendering for good as some JS magic constantly updates stuff. So, I think it’s fine to have a print that when invoked prints what you have at the moment. If you don’t want that, you have to wait till it’s done.

For what it’s worth: Besides Subsurface, I still maintain a paper logbook. But personally, I don’t use this printing mechanism at all, I go via the TeX export because I think the typesetting there is much nicer. But that might be my personal kink.

Best
Robert

bstoeger:
I believe we should close this PR - In my tests, I don't get a second "loaded" event if I remove the second `connect` invocation. Perhaps the waiting for the image loading has to be done in JS? This API is insane.

PR #3357, opened 2021-12-03
Round gas depth properly

atdotde:
The D in MOD, EAD, END, and EADD stands for "depth" and
as such these should be mm in int rather than double.

The intermediate fn2 and fhe2, however, as intermediate
value should not be rounded to an integer.

The upshot of this is a litle more numerical stability.
It should lead to more stable values in TestProfile
when run on architectures with different floating
point precision.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -41,7 +41,7 @@ const char *divemode_text_ui[] = {
 // For writing/reading files.
 const char *divemode_text[] = {"OC", "CCR", "PSCR", "Freedive"};
 
-static int calculate_depth_to_mbar(int depth, pressure_t surface_pressure, int salinity);
+static double calculate_depth_to_mbarf(int depth, pressure_t surface_pressure, int salinity);
 
 /*
  * The legacy format for sample pressures has a single pressure
@@ -657,7 +657,7 @@ void update_setpoint_events(const struct dive *dive, struct divecomputer *dc)
 				gasmix = get_gasmix_from_event(dive, ev);
 				next = get_next_event(ev, "gaschange");
 			}
-			fill_pressures(&pressures, calculate_depth_to_mbar(dc->sample[i].depth.mm, dc->surface_pressure, 0), gasmix ,0, dc->divemode);
+			fill_pressures(&pressures, lrint(calculate_depth_to_mbarf(dc->sample[i].depth.mm, dc->surface_pressure, 0)), gasmix ,0, dc->divemode);
 			if (abs(dc->sample[i].setpoint.mbar - (int)(1000 * pressures.o2)) <= 50)
 				dc->sample[i].setpoint.mbar = 0;
 		}
@@ -3209,7 +3209,7 @@ static double salinity_to_specific_weight(int salinity)
 /* Pa = N/m^2 - so we determine the weight (in N) of the mass of 10m
  * of water (and use standard salt water at 1.03kg per liter if we don't know salinity)
  * and add that to the surface pressure (or to 1013 if that's unknown) */
-static int calculate_depth_to_mbar(int depth, pressure_t surface_pressure, int salinity)
+static double calculate_depth_to_mbarf(int depth, pressure_t surface_pressure, int salinity)
 {
 	double specific_weight;
 	int mbar = surface_pressure.mbar;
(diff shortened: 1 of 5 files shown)


glance-:
:+1: 

Id love to see the units in struct plot_data also be changed to depth_t , duration_t and so on, but that might be another PR.

core/profile.c
```diff
@@ -1236,15 +1236,15 @@ static void calculate_gas_information_new(const struct dive *dive, const struct
...
 		struct plot_data *entry = pi->entry + i;
 
 		gasmix = get_gasmix(dive, dc, entry->sec, &evg, gasmix);
 		amb_pressure = depth_to_bar(entry->depth, dive);
 		current_divemode = get_current_divemode(dc, entry->sec, &evd, &current_divemode);
 		fill_pressures(&entry->pressures, amb_pressure, gasmix, (current_divemode == OC) ? 0.0 : entry->o2pressure.mbar / 1000.0, current_divemode);
-		fn2 = (int)(1000.0 * entry->pressures.n2 / amb_pressure);
-		fhe = (int)(1000.0 * entry->pressures.he / amb_pressure);
+		fn2 = (1000.0 * entry->pressures.n2 / amb_pressure);
+		fhe = (1000.0 * entry->pressures.he / amb_pressure);
```
bstoeger: Small style comment: You could remove the parentheses in the previous two lines.

No comment on the rest - this is above my pay-grade. Turning these two into floating points looks reasonable though.

core/profile.c
```diff
@@ -1255,14 +1255,14 @@ static void calculate_gas_information_new(const struct dive *dive, const struct
 		 * END takes O₂ + N₂ (air) into account ("Narcotic" for trimix dives)
 		 * EAD just uses N₂ ("Air" for nitrox dives) */
 		pressure_t modpO2 = { .mbar = (int)(prefs.modpO2 * 1000) };
-		entry->mod = (double)gas_mod(gasmix, modpO2, dive, 1).mm;
-		entry->end = (entry->depth + 10000) * (1000 - fhe) / 1000.0 - 10000;
-		entry->ead = (entry->depth + 10000) * fn2 / (double)N2_IN_AIR - 10000;
-		entry->eadd = (entry->depth + 10000) *
+		entry->mod = gas_mod(gasmix, modpO2, dive, 1).mm;
+		entry->end = mbar_to_depth((int)(depth_to_mbarf(entry->depth, dive) * (1000 - fhe) / 1000.0), dive);
```
bstoeger: This code is mathematically wrong.

`(int)(...)` will round 100.0-100.0*eps (99.9999999...) to 99.

You want `lrint(...)` or `(int)(round(...))`, as I have stated before.

Of course, that will still give the same rounding problem just for 100.5 +/- 100.5 * eps. But at least the result is not outright wrong.

To fully avoid the problem, I reckon `mbar_to_depth` should also exist as version taking a double.

core/dive.c
```diff
@@ -41,7 +41,7 @@ const char *divemode_text_ui[] = {
 // For writing/reading files.
 const char *divemode_text[] = {"OC", "CCR", "PSCR", "Freedive"};
 
-static int calculate_depth_to_mbar(int depth, pressure_t surface_pressure, int salinity);
+static double calculate_depth_to_mbarf(int depth, pressure_t surface_pressure, int salinity);
```
dirkhh, line 44: any function name ending on `barf` makes me happy

core/profile.c
dirkhh: you still could drop those parenthesis, right?

dirkhh (commented):
this is above my pay grade when it comes to the math, so just pointless comments

core/profile.c
atdotde: You are right, the (int) cast always rounds down not to nearest value. 

I don't think the problem at 100.5 is as big as we have many situations where we are at an int (e.g. END for nitrox dives) then do calculation that for real numbers is a NOOP but not for floating point and for (int) we were doing that at the edge of the discontinuity. But I would expect much fewer cases where we are sitting close the the .5 edge.

I disagree, though that a double version of mbar_to_depth would help: Then, in the test, we would be comparing doubles for equality which is bad, comparing the rounded ints has much better chance of not picking up noise.

atdotde:
Argh, I had removed the parentheses in the second commit that worked around the trivial cases and that I thew away after @bstoeger's comment.

Re "barf": I first encountered that term a long time ago in "The unix haters handbook" (https://en.wikipedia.org/wiki/The_UNIX-HATERS_Handbook) that came with a "barf bag" which was still there in pristine state in our library's copy.

core/profile.c
bstoeger: > I disagree, though that a double version of mbar_to_depth would help: Then, in the test, we would be comparing doubles for equality which is bad, comparing the rounded ints has much better chance of not picking up noise.

You misunderstood. I'm not suggesting to return a double, but to take a double argument. The fundamental problem in all this code is intermittent conversion to integral types. Fixed-point and floating-point are both fine, if done correctly. It's the combination that ends up in a disaster.

atdotde:
This seems to resolve the rounding problems encountered in TestProfile on different architectures. So let's merge.

bstoeger:
Just an idle comment, feel free to ignore:

> This seems to resolve the rounding problems encountered in TestProfile on different architectures. So let's merge.

I think this is the wrong way to think about it. The tests are there to find questionable code, it's not the code that is there to pass the tests. It's the same thing with compiler warnings (and other validation techniques) that many people get wrong.

In this case the tests alerted us to questionable code and I still think that the current version is wrong: It truncates mid-calculation to full mbar, which has less precision than the resulting mm.

However, since the imprecision is below the precision of the measured data I don't really care.

PR #3358, opened 2021-12-04
profile: fix editing depth / duration

bstoeger:
The undo commands for depth and duration editing cleared
the samples of the dive computer. They relied on the profile
automatically creating a fake profile. However, at some point
that code got removed. Therefore, do it explicitly in the undo
command.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a severe bug: Editing of the duration / depth of manually added dives made the profile not display anything. The reason being that the edit command clears the profile and relied on the profile-widget recalculating it. However, I removed that some time ago, not realizing that this was still needed in this case. Hrmpf.

### Release note:

Done.
### Documentation change:

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Profile: Fix editing of time / duration (profile of edited dive was not shown)
 Divelist: Don't attempt to compute SAC for CCR dives
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -306,6 +306,7 @@ void EditDuration::set(struct dive *d, int value) const
 	d->duration = d->dc.duration;
 	d->dc.meandepth.mm = 0;
 	d->dc.samples = 0;
+	fake_dc(&d->dc);
 }
 
 int EditDuration::data(struct dive *d) const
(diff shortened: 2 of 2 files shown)


dirkhh (approved):
thanks - this seems obviously correct

dirkhh:
annoyingly, one of the tests is still in "queued" state... 
You requested a review from Linus - he'll tell you that he has no idea what the Undo code does :)
Do you think that review is important - given the amount of change that we are about to merge in the monster branch?

bstoeger:
The crucial question is if that is the right way to change the depth/duration. Perhaps I'm missing something? If you are confident that this is correct, feel free to merge.

dirkhh:
> The crucial question is if that is the right way to change the depth/duration. Perhaps I'm missing something? If you are confident that this is correct, feel free to merge.

I don't understand your question, though... all you are doing is create a new fake dive computer structure.
So I'm 99% sure we are talking past each other.
Please elaborate your concern, and I'll ping Linus to take a look

bstoeger:
Maybe a fake_dc() is not enough, but a fixup_dive() is needed? I simply don't know. That is all.

torvalds (approved):

dirkhh:
No idea why this one run is 'queued' still - but I see no point in waiting for it. Merging this now.

PR #3362, opened 2021-12-17
Random cleanup

dirkhh:

- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

regular mobile version number updates
small fixes

main purpose of this work, really, is to get the number of commits up so that after the next merge into master the broken builds that were created after yesterday's mistaken 'force push' into master will disappear from Launchpad and OBS...

### Mentions:

@bstoeger this should all be trivial


bstoeger (approved):
All look good to me. Perhaps I would use `char *data` and avoid the cast later on, but doesn't matter.

PR #3364, opened 2021-12-31
Allow rudimentary ordering of the bar charts

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This was requested on the mailing list, when introducing the statistics module: allow sorting of bar charts by count. In the case of value-based charts, this sorts by value and by count.

The UI is inconsistent, since here the allowed options are based on the chart and not vice-versa. I found the other way round a bit confusing. YMMV.

The good thing about this is that it provides a rather obvious differentiation between histograms and categorical charts. That caused some confusion in the past.

For stacked/grouped bar charts, we might think about sorting on the second variable. However, the result might be extremely confusing.

The layout of the mobile version is funky in wide mode. Beware.

### Changes made:

1) Add rudimentary sorting of bar charts.
2) Add support on mobile.

### Release note:

Done.

### Documentation change:

Still has to be done. Could be a good way to clarify the dichotomy categorical charts / histograms.

### Mentions:

@willemferguson

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Statistics: Implement ordering of categorical bar charts
 Profile: Fix editing of time / duration (profile of edited dive was not shown)
 Divelist: Don't attempt to compute SAC for CCR dives
 ---
--- a/desktop-widgets/statswidget.cpp
+++ b/desktop-widgets/statswidget.cpp
@@ -84,6 +84,7 @@ StatsWidget::StatsWidget(QWidget *parent) : QWidget(parent)
 	connect(ui.var1Binner, QOverload<int>::of(&QComboBox::activated), this, &StatsWidget::var1BinnerChanged);
 	connect(ui.var2Binner, QOverload<int>::of(&QComboBox::activated), this, &StatsWidget::var2BinnerChanged);
 	connect(ui.var2Operation, QOverload<int>::of(&QComboBox::activated), this, &StatsWidget::var2OperationChanged);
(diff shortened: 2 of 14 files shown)


willemferguson:
It looks perfect to me, using my dive log as a test case. Thank you, Berthold. I will update the user manual.

willemferguson:
I need to mention that testing was using the desktop version of Subsurface

dirkhh:
So, err... I finally got around to build this and I can't figure out how to use it. Where exactly does that sort show up?
I tried both desktop and mobile and am feeling a bit... lame to ask this...
On mobile I at least see the sort field - except that I can't figure out how to create a graph where the sort is enabled.
On desktop I don't even see a sort UI...
What am I missing? Yes, I tried `make clean` in case something wasn't rebuilt... do I need to be more aggressive than that?

bstoeger:
You have to switch from a histogram to a categorical bar chart. A histogram can't be sorted: the axis is interpreted in a continuous (so far linear) way.

Yes, that will have to be explained in the documentation.

An alternative UI would let the user decide the sorting before selecting the chart and then not show charts with a continuous x-axis. However, I found that confusing. YMMV.

dirkhh:
Yeah, this does look reasonable to me. Documentation is key, as always.
But approved from my perspective.

dirkhh (approved):

PR #3369, opened 2022-01-04
tests: fix integer constants

dirkhh:
In TestUnitConversion we used casts instead of the more common suffix
designations to indicate the type of those integer constants.
Worse, in commit efab955d85 ("cleanup: make feet_to_mm signed") the
return type of feet_to_mm() changed, but the value it is compared to
wasn't adjusted in the test which caused some builds with more
aggressive compiler flags to fail.

- [x] Bug fix

This caused the Coverity build to fail

### Mentions:

@bstoeger

### DIFF
--- a/tests/testunitconversion.cpp
+++ b/tests/testunitconversion.cpp
@@ -10,14 +10,14 @@ void TestUnitConversion::testUnitConversions()
 	QCOMPARE(IS_FP_SAME(ml_to_cuft(1000), 0.0353147), true);
 	QCOMPARE(IS_FP_SAME(cuft_to_l(1), 28.316847), true);
 	QCOMPARE(IS_FP_SAME(mm_to_feet(1000), 3.280840), true);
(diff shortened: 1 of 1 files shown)


bstoeger (approved):
LGTM.I don't understand why that made the build fail, but so be it.

dirkhh:
> LGTM.I don't understand why that made the build fail, but so be it.

```
CMakeFiles/TestUnitConversion.dir/testunitconversion.cpp.o: In function `TestUnitConversion::testUnitConversions()':
/__w/subsurface/subsurface/tests/testunitconversion.cpp:13: undefined reference to `bool QTest::qCompare<long, unsigned long>(long const&, unsigned long const&, char const*, char const*, char const*, int)'
```

bstoeger:
OK, interesting. QTest seems to be one of those outdated things that nobody ever bothers to improve. I guess it does its job good enough. 🤷‍♂

PR #3379, opened 2022-01-13
[no merging, please] desktop: don't update notes field if unchanged

### LINKED ISSUE #3365: switching windows using alt-tab moves cursor in notes field of a dive
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
With the notes field of a dive focused and cursor positioned within field, switching to another window via Alt-Tab and then back to Subsurface with Alt-Tab moves the cursor to the start of the notes field. I expected the cursor to remain at the same position with the Notes field.

### Operating system:
Ubuntu Linux 20.04.3 LTS
KDE Plasma Version 5.18.5
KDE Frameworks Version 5.68.0
Qt Version 5.12.8

### Subsurface version:
5.0.5

### Steps to reproduce:
Click on a dive.
Click in the Notes field, for example at the end.
Press Alt-Tab to switch to another window.
Press Alt-Tab to switch back to Subsurface.
The cursor will have moved position.

### Current behavior:
The cursor moves position.

### Expected behavior:
The cursor does not move position.

bstoeger:
User report: when switching focus between windows, the
cursor position gets lost. This is due to a note-edit
command being fired, which then overwrites the notes tab.

This should probably fixed in a better way (recognize that
the field is being edited and not emit a signal). However,
this is a quick fix for now.

Should this be extended to other fields?

Fixes #3365

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a somewhat unsatisfying fix for #3365. It works in my simple testing though.

Of course, if we want to go this route, this should also be done for the other text fields! Let's wait for user input.
 
### Changes made:

Don't set notes tab when unchanged.

### Related issues:

Fixes #3365

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+Desktop: Don't lose cursor position in notes when switching between windows
 Statistics: Implement ordering of categorical bar charts
 Profile: Fix editing of time / duration (profile of edited dive was not shown)
 Divelist: Don't attempt to compute SAC for CCR dives
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -303,14 +303,27 @@ static bool isHtml(const QString &s)
 	return s.contains("<div", Qt::CaseInsensitive) || s.contains("<table", Qt::CaseInsensitive);
 }
 
+static QString htmlOrPlaintext(const QTextEdit &edit)
+{
+	QString html = edit.toHtml();
+	return isHtml(html) ? html : edit.toPlainText();
+}
+
 void MainTab::updateNotes(const struct dive *d)
 {
+	// Don't set notes field if the content is unchanged. That causes the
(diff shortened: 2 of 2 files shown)


bstoeger:
Please do not commit - I'll try a more general approach later.

dirkhh (commented):
I will admit that I don't understand how that would even fix the issue...

bstoeger:
1) Notes field loses focus
2) Notes editing command is executed
3) Notes of dive(s) are updated
4) New notes are written into notes field
5) Cursor is reset to beginning of field

This fixes the bug by preventing 4. and in consequence 5.

dirkhh:
Ahhhh - thanks for that explanation.
But you still don't want this merged, correct?

bstoeger:
Superseded by #3392

PR #3383, opened 2022-01-29
DM5 import: Add dive mode support (OC and CCR)

mturkia:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Import dive mode from DM5 database. Currently only OC and CCR are supported as those were the only modes available in database used to map the values.

### DIFF
--- a/core/import-suunto.c
+++ b/core/import-suunto.c
@@ -432,6 +432,21 @@ static int dm5_dive(void *param, int columns, char **data, char **column)
 	if (data[5])
 		utf8_string(data[5], &state->cur_dive->dc.model);
 
+	if (data[25]) {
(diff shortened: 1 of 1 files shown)


core/import-suunto.c
```diff
@@ -432,6 +432,21 @@ static int dm5_dive(void *param, int columns, char **data, char **column)
...
+		switch(atoi(data[25])) {
+			case 1:
+				state->cur_dive->dc.divemode = 0;
+				break;
+			case 5:
+				state->cur_dive->dc.divemode = 1;
+				break;
+			default:
+				state->cur_dive->dc.divemode = 0;
+				break;
```
dirkhh, line 446: one could argue that '1' and 'default' are kinda the same...

mturkia: I like to be explicit on the information I have been able to dig out from the DM5 (or other) format. Now it is documented that 1 is OC, unlike if I was using the default for it.

PR #3384, opened 2022-01-31
Add support for the new OSTC hardware

jefdriesen:
In the latest OSTC hardware, the Telit/Stollman bluetooth module has been replaced with a u-Blox Nina B2 bluetooth module. The BLE communication protocol remains roughly the same, except for a few minor differences:

 - New UUIDs for services and characteristics
 - Only one common characteristic for Rx and Tx
 - Credit based flow control is optional
 - Credit value of 255 corresponds to a disconnect

This also requires the following change in libdivecomputer: https://github.com/libdivecomputer/libdivecomputer/commit/d960a37e120757a7459c23a5bb3cc058e7269563


core/qt-ble.cpp
```diff
@@ -490,7 +492,7 @@ dc_status_t BLEObject::setupHwTerminalIo(const QList<QLowEnergyCharacteristic> &
 	 * 0x0000FEFB is a clear indication that the OSTC is equipped with this BT/BLE hardware.
 	 */
 
-	if (allC.length() != 4) {
+	if (allC.length() != 4 && allC.length() != 2) {
 		qDebug() << "This should not happen. HW/OSTC BT/BLE device without 4 Characteristics";
```
dirkhh, line 496: I guess we should update the debug message here as well... I can fix that when I merge it

dirkhh:
Thanks, this looks reasonable - I wonder a bit about the 255/254 change (and if that would potentially cause a concern on the old hardware), but from my naive reading of the code, I don't think so.

dirkhh:
merged locally

jefdriesen:
I don't think the actual values for the min/max credits are that important here. The minimum value should be high enough to make sure the device can keep sending data. But as long as the credits are updated in time, any value will work. I tried with zero, and even that worked fine, although I wouldn't use such a low value in production. The maximum value is the maximum number of packets you can queue in the receiver. Since we're only limited by available memory, we can use some large value here. The only value we can't use anymore is 255 (0xFF), because for the U-Blox that's a special value to terminate the connection. A large delta between min and max, means having to update the credits less frequently and thus less overhead.

Note that on the device side, the OSTC seems to be using 8/16 (U-Blox) and 2/23 (Telit).

PR #3387, opened 2022-02-10
[WIP] Apple m1

dirkhh:

- [x] Bug fix
- [x] Functional change
- [x] New feature
- [x] Code cleanup
- [x] Build system change

So this is early days... but at least it builds and I get a working (-ish) M1 build
Bluetooth still doesn't work.
Maps are gone.
The `install` step to create a self contained binary needs to be updated
I'm sure there are other things that are broken that I just haven't found yet.

### Changes made:

- small random cleanups
- remove dependency on QtLocation which is missing in Qt6
- deal with the changes in cmake (all these macros that contained 'qt5' in their name. Similar for certain paths
- deal with some gratuitous API breakage (renamed signals and slots, deprecated or removed implicit conversions)
- a couple of things that look like bugs that the Qt5 silently ignored (the Future type for example)
- adjustment to build our dependencies on M1 (especially OpenSSL, libmtp and libusb)

### Additional information:

I tried compiling this against Qt5 on an Intel Mac.
This is untested on anything else (that's part of why I'm creating the PR).
Qt5 on M1 Macs isn't supported by Qt (yeah, you can build that yourself from source but I'm not ready to go down this path). But of course it is reasonable to now also start looking at Qt6 on other platforms...

### Release note:

This will need some acknowledgment that Qt6 is missing the maps and that other things are still broken (especially BT)

### Mentions:

@thiagomacieira @tcanabrava I'd love comments from you if you have the time


dirkhh:
ok - oddly the one thing that broke ALL of these builds had worked locally with a Qt 5.15 build on macOS. Go figure.
This also cleans up a few commit messages, makes the order a bit more logical.

dirkhh:
OK, this fixes the breakage with Qt 5.9 - let's see how this goes.

CMakeLists.txt
```diff
@@ -284,39 +308,38 @@ if(ANDROID)
 	# our Qt installation. This is ugly, but it works.
 	set(CMAKE_FIND_ROOT_PATH "/;${CMAKE_FIND_ROOT_PATH}")
 endif()
-set(QT_FIND_COMPONENTS Core Concurrent Widgets Network Svg Positioning Quick Location ${QT_EXTRA_COMPONENTS})
+set(QT_FIND_COMPONENTS Core Concurrent Widgets Network Svg Positioning Quick ${QT_EXTRA_COMPONENTS})
 if (SUBSURFACE_TARGET_EXECUTABLE MATCHES "DesktopExecutable")
-	find_package(Qt5 5.9.1 REQUIRED COMPONENTS ${QT_FIND_COMPONENTS} LinguistTools Test QuickTest)
+	find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS ${QT_FIND_COMPONENTS} LinguistTools Test QuickTest)
```
thiagomacieira, line 313: I wonder how Qt Creator is doing it. I've just built it using CMake against Qt 6, but its CMakeLists.txt files still say "Qt5". 

See https://github.com/qt-creator/qt-creator/blob/master/CMakeLists.txt#L56-L68

core/qthelper.h
```diff
@@ -29,6 +30,18 @@ class QImage;
 #define SKIP_EMPTY QString::SkipEmptyParts
 #endif
 
+// this is annoying Qt5 / Qt6 incompatibility where we can't compare against string literals anymore
```
thiagomacieira, line 33: Yeah, we couldn't do much if we needed to remove QStringRef.

But you should be able to write:
```c++
    r.name().compare(QLatin1String(cs))
```

BTW, do note that one returns a memcmp-like result, with 0 being "equal", whereas the == operator returns a boolean, so 1 (true) is equal.

desktop-widgets/subsurfacewebservices.cpp
```diff
@@ -185,11 +185,15 @@ static DiveListResult parseDiveLogsDeDiveList(const QByteArray &xmlData)
 		// process <DiveDates>
 		seenDiveDates = true;
 		while (reader.readNextStartElement()) {
-			if (reader.name() != "date") {
+			if (nameCmp(reader, "date") != 0) {
 				// qDebug() << "Skipping" << reader.name();
 				continue;
 			}
+#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
+			QStringView id = reader.attributes().value("divelogsId");
```
thiagomacieira: Hint:
```c++
    auto id = reader.attributes().value("divelogsId");
```

tests/testparseperformance.cpp
```diff
@@ -10,6 +10,7 @@
 #include <QFile>
 #include <QDebug>
 #include <QNetworkProxy>
+#include "QTextCodec"
```
thiagomacieira, line 13: Qt 6.4 is removing a bunch of unused includes, so you may need to update this commit soon.

desktop-widgets/groupedlineedit.cpp
```diff
@@ -175,7 +175,11 @@ void GroupedLineEdit::paintEvent(QPaintEvent *e)
 	painter.setRenderHint(QPainter::Antialiasing, true);
 	painter.fillRect(0, 0, viewport()->width(), viewport()->height(), palette().base());
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
```
thiagomacieira: You shouldn't need this if you `#include <qvector.h>` in addition to `qlist.h`. See https://github.com/qt/qtbase/blob/dev/src/corelib/tools/qvector.h#L48-L53

thiagomacieira (changes requested):
The only wrong one is the XML name comparison.

CMakeLists.txt
dirkhh: so according to the docs there are three different ways you can do this. One works only with Qt 5.15 and newer. One works only up to Qt 5.15. And the way I did it is the one that's recommended for Qt 5 or Qt 6.
It's entirely possible that I simply followed bad instructions... but in my defense, what I have works :)

core/qthelper.h
dirkhh: I think I correctly adjusted the logic everywhere, but I'll go through that one more time very carefully just in case...

desktop-widgets/subsurfacewebservices.cpp
dirkhh: brilliant. I should have thought of that. 
YES.

tests/testparseperformance.cpp
dirkhh: Qt 6.3 still isn't out. The 6.3 alpha macOS Arm64 binaries the day I started were completely broken and qmake wouldn't find its own mkspec files, so all this above was done with 6.2.3

desktop-widgets/groupedlineedit.cpp
dirkhh: ok, let me try that again. I couldn't make it work and so I ended up with this solution.

tests/testparseperformance.cpp
dirkhh: oh, any insights into whether QtLocation is coming back at some point?
Not having the map widget really is a huge let-down :(
QtWebKit is now completely dead and gone and that's a pain - but at least that's only needed for printing and displaying the user manual within the app (that can be worked around using a browser). And who really prints their stuff (I know, I know).
But not having the map widget is really taking a major part of the value away :(

core/qthelper.h
dirkhh: brain not working... I now realize what you meant. will be fixed in the next force push

tests/testparseperformance.cpp
thiagomacieira: It's coming back some time, but I can't tell you when that is. It's still on "status = ignore" in the dev branch today.

CMakeLists.txt
thiagomacieira: Well, I'm not a CMake expert either. But it's possible there were some changes in 5.15 to allow for 5 & 6 simplification, but only if you can require 5.15 on the first place and it looks like you can't.

dirkhh:
OK, I think I implemented @thiagomacieira 's suggestions.
I'm still finding plenty of bugs. The new statistics crash. Some files are missing.
This will take quite a bit more time, still. But it's progress.

core/qthelper.h
bstoeger: Having wasted quite some time with disentangling the include hell, I'm not happy that now every user of `qthelper` slurps in the obscure `QXmlStreamReader` header.

Moreover, the helper function seems ridiculously specific.

Can't we do something like
```
#include <QString>
// Make QStringView and QStringRef behave the same with respect to
// comparison with C-strings.
// Needed, because the Qt v6 API changed from the latter to the former.
inline bool operator==(QStringView &v, const char *s)
{
        return v.compare(QString::fromUtf8(s)) == 0;
}
```
and just keep the rest?

Of course, that would be back to conditional compilation, because QStringView does not exist pre-Qt 5.10 or somesuch. 

To avoid conditional compilation, we could define a much more general template, such as:
```
#include <QString>
template <typename StringView>
bool isSameString(StringView v, const char *s)
{
        return v.compare(QString::fromUtf8(s)) == 0;
}
```
and then use it like so
```
        if (isSameString(gpxReader.name(), "time") {
                ...
        }
```
This should work equally for QStringView and QStringRef.

dirkhh: running way behind - had a lovely Sunday morning far away from computers.

I will simply add a separate header file for this that is only included in the couple of places where we do this. Putting this in `qthelper.h` was laziness on my part.

dirkhh:
replaced by #3398 and #3399

Issue #3385, opened 2022-02-05
UEMIS Zürich: Hitting "Retry" after reconnect message will start over at first dive


Schwaneberg:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Subsurface requests the user to reconnect the device and click "Retry", if the Uemis Zürich has almost used up all it's memory (answer files) during synchronization.
Unfortunately, this will start over at the first dive.

### Operating system:

Ubuntu 20.04

### Subsurface version:

5.0.6

### Steps to reproduce:

Sync UEMIS with ~50 or more unsynced dives.

### Current behavior:

Going into an infinit loop fs full -> reconnect -> retry -> fs full

### Expected behavior:

resume syncing at the last dive

dirkhh:
none of the developers have access to a working Uemis Zürich anymore. At this point I would consider the support for this device "likely broken, no longer actively supported" - unless another developer starts working on it.

dirkhh:
should be fixed now :)
Again, thanks

PR #3388, opened 2022-02-10
Fix/uemis retry resuming download without retry button

Schwaneberg:

- [X] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Is PR is a possible fix to issue #3385 .
There is a prompt to reconnect the Uemis from the status bar. Not a very elegant solution, but an easy solution for uemis downloader (which is written in C). I have an Uemis Zürich in excellent condition and tested the code with it.
I also added support for an environment variable to define an offset to start downloading from.

I know that this patch isn't very high quality. I'd rather suggest to re-implement the Uemis Zürich downloader from scratch in modern C++, if it wasn't for such an old legacy device. I just hope that you accept this small change to help the (few?) remaining Uemis users to elongate the useful lifespan of the remaining devices.

### Changes made:
1) Replaced "goto bail" with a prompt to reconnect the dive computer, if the synchronization memory is almost full.
2) Automatic termination, reset of filenr and re-initialization after a 60s timeout.
3) Reading an env variable UEMIS_DIVE_OFFSET, which can be used to modify the  start index for the (very slow) indexing process.

### Related issues:

#3385 

### Release note:
Support synchronization over multiple reconnects with Uemis Zürich.

### Documentation change:
As you can see below, I added some lines to the documentation to provide some explanation for the use of the optional env variable.

### Mentions:
@dirkhh

### DIFF
--- a/core/uemis-downloader.c
+++ b/core/uemis-downloader.c
@@ -32,7 +32,7 @@
 #include "subsurface-time.h"
 #include "core/subsurface-string.h"
 
-#define ERR_FS_ALMOST_FULL QT_TRANSLATE_NOOP("gettextFromC", "Uemis Zurich: the file system is almost full.\nDisconnect/reconnect the dive computer\nand click \'Retry\'")
+#define ERR_FS_ALMOST_FULL QT_TRANSLATE_NOOP("gettextFromC", "Disconnect/reconnect the SDA")
 #define ERR_FS_FULL QT_TRANSLATE_NOOP("gettextFromC", "Uemis Zurich: the file system is full.\nDisconnect/reconnect the dive computer\nand click Retry")
 #define ERR_FS_SHORT_WRITE QT_TRANSLATE_NOOP("gettextFromC", "Short write to req.txt file.\nIs the Uemis Zurich plugged in correctly?")
 #define ERR_NO_FILES QT_TRANSLATE_NOOP("gettextFromC", "No dives to download.")
@@ -1434,7 +1434,13 @@ const char *do_uemis_import(device_data_t *data)
 #if UEMIS_DEBUG & 4
 				fprintf(debugfile, "d_u_i out of memory, bailing\n");
 #endif
-				break;
+				char msg[strlen(translate("gettextFromC", ERR_FS_ALMOST_FULL)) + 4];
+				for (int wait=60; wait >=0; wait--){
+					sprintf(msg, "%s %ds", translate("gettextFromC", ERR_FS_ALMOST_FULL), wait);
+					uemis_info(msg);
+					filenr = 0;
+					usleep(1000000);
+				}
 			}
 			/* if the user clicked cancel, exit gracefully */
 			if (import_thread_cancelled) {
--- a/core/uemis-downloader.c
+++ b/core/uemis-downloader.c
@@ -33,6 +33,7 @@
 #include "core/subsurface-string.h"
 
 #define ACTION_RECONNECT QT_TRANSLATE_NOOP("gettextFromC", "Disconnect/reconnect the SDA")
+#define ERR_FS_ALMOST_FULL QT_TRANSLATE_NOOP("gettextFromC", "Uemis Zurich: the file system is almost full.\nDisconnect/reconnect the dive computer\nand click \'Retry\'")
 #define ERR_FS_FULL QT_TRANSLATE_NOOP("gettextFromC", "Uemis Zurich: the file system is full.\nDisconnect/reconnect the dive computer\nand click Retry")
 #define ERR_FS_SHORT_WRITE QT_TRANSLATE_NOOP("gettextFromC", "Short write to req.txt file.\nIs the Uemis Zurich plugged in correctly?")
 #define ERR_NO_FILES QT_TRANSLATE_NOOP("gettextFromC", "No dives to download.")
@@ -1374,7 +1375,8 @@ const char *do_uemis_import(device_data_t *data)
 
 	first = start = atoi(newmax);
 	dive_to_read = mindiveid < first ? first - mindiveid : first;
-	start += dive_offset;
+	if (dive_offset > 0)
+		start += dive_offset;
 	for (;;) {
 #if UEMIS_DEBUG & 2
 		debug_round++;
@@ -1446,7 +1448,6 @@ const char *do_uemis_import(device_data_t *data)
 				const char *errormsg = translate("gettextFromC", ACTION_RECONNECT);
 				for (int wait=60; wait >=0; wait--){
 					uemis_info("%s %ds", errormsg, wait);
-					filenr = 0;
 					usleep(1000000);
 				}
 				// Resetting to original state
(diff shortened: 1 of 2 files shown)


core/uemis-downloader.c
```diff
@@ -1434,7 +1434,13 @@ const char *do_uemis_import(device_data_t *data)
 #if UEMIS_DEBUG & 4
 				fprintf(debugfile, "d_u_i out of memory, bailing\n");
 #endif
-				break;
+				char msg[strlen(translate("gettextFromC", ERR_FS_ALMOST_FULL)) + 4];
```
dirkhh: so `translate(context, text)` returns a `const char *` into the `translationCache`.
Because of that this code is not quite as bad as it looks at first glance, but still... ugh. That's pretty darn wasteful.
Why not
```
const char *errormsg = translate("gettextFromC", ERR_FS_ALMOST_FULL);
char msg[strlen(errormsg) + 4];
```
and then use `errormsg` in the `sprintf()` call below.
I realize that I'm old... variable sized arrays have been around since C99 / C++14 and I still would have instinctively written this as a `malloc()` - even though this makes memory management easier on the code...

core/uemis-downloader.c
```diff
@@ -1434,7 +1434,13 @@ const char *do_uemis_import(device_data_t *data)
...
+				for (int wait=60; wait >=0; wait--){
+					sprintf(msg, "%s %ds", translate("gettextFromC", ERR_FS_ALMOST_FULL), wait);
+					uemis_info(msg);
+					filenr = 0;
+					usleep(1000000);
```
dirkhh, line 1451: wait, what? Why are you doing this busy loop? A 60s countdown in the status messages? With no way to interrupt?
Yeah, not a fan. Not at all a fan.
Also... did you take a look at the definition of `uemis_info()`? That already takes `printf()` style arguments, so that whole `sprintf()` and the whole thing with the `msg` buffer I commented on above is not needed in the first place.

But before we focus on that simplification... why the busy loop here? How does that help?

core/uemis-downloader.c
```diff
@@ -1434,9 +1434,9 @@ const char *do_uemis_import(device_data_t *data)
 #if UEMIS_DEBUG & 4
 				fprintf(debugfile, "d_u_i out of memory, bailing\n");
 #endif
-				char msg[strlen(translate("gettextFromC", ERR_FS_ALMOST_FULL)) + 4];
+				char msg[strlen(translate("gettextFromC", ACTION_RECONNECT)) + 4];
 				for (int wait=60; wait >=0; wait--){
-					sprintf(msg, "%s %ds", translate("gettextFromC", ERR_FS_ALMOST_FULL), wait);
+					sprintf(msg, "%s %ds", translate("gettextFromC", ACTION_RECONNECT), wait);
 					uemis_info(msg);
```
dirkhh: that part should have been squashed with the previous commit

core/uemis-downloader.c
```diff
@@ -1495,9 +1495,6 @@ const char *do_uemis_import(device_data_t *data)
 			next_table_index++;
 	}
 
-	if (uemis_mem_status != UEMIS_MEM_OK)
-		result = translate("gettextFromC", ERR_FS_ALMOST_FULL);
-
```
dirkhh, line 1494: and that should be explained in the commit message...

core/uemis-downloader.c
```diff
@@ -1367,6 +1374,7 @@ const char *do_uemis_import(device_data_t *data)
 
 	first = start = atoi(newmax);
 	dive_to_read = mindiveid < first ? first - mindiveid : first;
+	start += dive_offset;
```
dirkhh: what does the algorithm do with a negative offset? I mean, sure, garbage in, garbage out... but maybe the code should require this to be `>= 0` ?

core/uemis-downloader.c
```diff
@@ -1434,13 +1442,29 @@ const char *do_uemis_import(device_data_t *data)
...
+					printf(stderr, "Uemis: Device id has changed after reconnect!\n");
+					goto bail;
+				}
+				param_buff[0] = strdup(deviceid);
+				if (!uemis_get_answer(mountpath, "initSession", 1, 6, &result))
+					goto bail;
+				uemis_info(translate("gettextFromC", "Start download"));
+				if (!uemis_get_answer(mountpath, "processSync", 0, 2, &result))
+					goto bail;
+				param_buff[1] = "notempty";
```
dirkhh, line 1469: shouldn't the commit message (or even better, some comments here) explain what this is doing?

dirkhh (changes requested):
I have a few comments that are worth a conversation, and a few things that just really should be changed.

core/uemis-downloader.c
Schwaneberg: Ok. BTW: I doubt that the shortened message "Disconnect/reconnect the SDA" requires any translation at all.
It must be short to fit in the progress bar.

core/uemis-downloader.c
Schwaneberg: I'd really like to use a QMessageBox dialog instead with "resume" / "cancel" as options, but this is pure C and I don't see how without adding some ugly hooks.

core/uemis-downloader.c
Schwaneberg: This is just a partial initialization - just as above the for (;;) loop. I could have refactored the code and put this into a new function for reuser, but instead I implemented a check to compare the device id.

core/uemis-downloader.c
dirkhh: yeah - that would be through a callback. We use that mechanism in other places.
This really isn't a solution. This isn't even a decent hack 😂

core/uemis-downloader.c
```diff
@@ -1443,12 +1443,13 @@ const char *do_uemis_import(device_data_t *data)
 				fprintf(debugfile, "d_u_i out of memory, bailing\n");
 #endif
 				(void)uemis_get_answer(mountpath, "terminateSync", 0, 3, &result);
-				char msg[strlen(translate("gettextFromC", ACTION_RECONNECT)) + 4];
+				const char *errormsg = translate("gettextFromC", ACTION_RECONNECT);
 				for (int wait=60; wait >=0; wait--){
-					sprintf(msg, "%s %ds", translate("gettextFromC", ACTION_RECONNECT), wait);
-					uemis_info(msg);
+					uemis_info("%s %ds", errormsg, wait);
+					filenr = 0;
```
dirkhh: remind me - why are you setting `filenr` to 0 sixty-one times in a row?

core/uemis-downloader.c
```diff
@@ -1443,12 +1443,13 @@ const char *do_uemis_import(device_data_t *data)
...
 					usleep(1000000);
 				}
+				// Resetting to original state
 				filenr = 0;
```
dirkhh, line 1454: and then one more time, just to be sure? 😁

core/uemis-downloader.c
Schwaneberg: I can revert this, but it shouldn't even happen anymore.

dirkhh:
ok, I think this makes sense... I'll refactor the commits locally and create a new branch from that so we can run our CI/CD against the cleaned up version.

dirkhh:
closing this since #3393 was merged

Issue #3382, opened 2022-01-22
Dive List can't toggle the display of divemaster


ckuethe:
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Right-clicking on the header ribbon of the dive list to show or hide the divemaster column reveals that the divemaster is not a column cannot be toggled

### Operating system:

* Ubuntu 21.10 x64_64, en-us, linux 5.15.15-76051515-generic
* Windows 10 Home

* System76 Lemur 10 Pro
* HP Spectre 360

Yes, this happens both on Linux and Windows

### Subsurface version:

5.0.5

Untested with other subsurface versions

Official release from https://subsurface.github.io/download/ for windows, and ubuntu PPA

### Steps to reproduce:

1. open dive list
2. create two new dives; specify Bilbo Baggins as the divemaster on one dive, and Samwise Gamgee as the divemaster of the second dive. Set buddy to be Gandalf.
3. Right click on the header ribbon in the dive list to toggle display of divemaster and buddy.
4. Observe that the buddy display does toggle on and off as expected, while it it not possible to do the same for the divemaster.

Here's a li'l logbook that would demonstrate this.
[log.xml.gz](https://github.com/subsurface/subsurface/files/7918953/log.xml.gz)

### Current behavior:

![Screenshot_2022-01-22_08-54-26](https://user-images.githubusercontent.com/1131751/150648690-d7a17596-99d4-45cd-bf2e-4f5686a42e5e.png)

### Expected behavior:

I was expecting that divemaster would be one of the columns that could be toggled like Buddy, Country, or Location.

### Additional information:

See attached log.xml.gz

ckuethe:
Thanks for fixing this. The latest version works very well. :)

PR #3389, opened 2022-02-12
[EXPERIMENTAL] dive master -> dive guide

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Since a user requests listing the "dive master" in the dive list (see #3382), I guess it is the right moment to change from master to guide, as discussed in #2879.

This is of course very dangerous as divemaster is used as a tag in the XML and git logs as well as in foreign formats.

Nevertheless, I would suggest to start the transition now, before this becomes a never-ending story.

I shortly tested the desktop version and it seemed to load/save the dive guide and statistics still work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+General: rename dive master to dive guide
 Statistics: Implement ordering of categorical bar charts
 Profile: Fix editing of time / duration (profile of edited dive was not shown)
 Divelist: Don't attempt to compute SAC for CCR dives
--- a/Documentation/user-manual.txt
+++ b/Documentation/user-manual.txt
@@ -178,7 +178,7 @@ These dives can be imported from:
 This is usually the approach for dives without a dive computer. The basic record
 of information within _Subsurface_ is a dive. The most important information
 in a simple dive logbook usually includes dive type, date, time, duration,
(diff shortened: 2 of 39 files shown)


dirkhh:
I am wondering about separating the UI side from the file format side.
I completely agree with the UI changes. 
But I worry about changing especially the git format.
Because if we write the new format to cloud storage, no old app can parse the data that was written to the cloud...
So I think the approach would be to add the ability to read `diveguide`, but to still write `divemaster`. And then in maybe a year switch things to write `diveguide`...
thoughts?

bstoeger:
> So I think the approach would be to add the ability to read diveguide, but to still write divemaster. And then in maybe a year switch things to write diveguide...

That's what this PR attempts to do. It's very likely that I made a few errors here or there, of course.

dirkhh:
> > So I think the approach would be to add the ability to read diveguide, but to still write divemaster. And then in maybe a year switch things to write diveguide...
> 
> That's what this PR attempts to do. It's very likely that I made a few errors here or there, of course.

Doing PR reviews at 4am maybe isn't super smart.
I have a couple of errands to run, but will come back to this later today (I hope).

dirkhh:
merged via #3390

PR #3393, opened 2022-02-14
Uemis support: hack around the need to reconnect the Uemis Zurich

dirkhh:
When the file system of the Zurich gets full, the only way to continue to
download from it, is to disconnect and reconnect the dive computer (which
resets the FAT file system that it emulates to 'empty').

This solution is rather hacky and weird because it does a hard count down in a
busy loop, but given the narrow use case, this may be acceptable.

This also adds support for the UEMIS_DIVE_OFFSET environment variable that
allows the user to skip dives on the device.

[refactored by Dirk Hohndel]

Signed-off-by: Oliver Schwaneberg <oliver.schwaneberg@gmail.com>

- [x] Bug fix
- [x] Functional change
- [x] New feature
- [x] Documentation change

This is simply a combined version of the commits in #3388 - with a hopefully better commit message and some white space issues removed and a ChangeLog added

I still dislike the hard count down - but frankly, given how many people are using this particular feature, I'm not sure it's worth to create a cleaner solution

### Related issues:

#3385 

### Release note:

done

### Documentation change:

done

### Mentions:

@Schwaneberg

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Uemis support: fix the ability disconnect/reconnect the Zurich when its filesystem is full
 - libdivecomputer: add support for latest BLE hardware in OSTC dive computers
 
 ---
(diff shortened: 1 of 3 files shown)


dirkhh:
OK, no comments - it's an obscure part of the code. And it helps at least one user who went through the effort to submit the code. So I'll merge it.
Thank you @Schwaneberg

PR #3394, opened 2022-02-15
cleanup: Fix a Coverity warning

bstoeger:
Two pointers were checked against NULL and then both were
dereferenced if at least one was not NULL. Of course, this
should have been an and, not an or expression.

That said, this is a semi-false positive, since both pointers
are set in the constructor and therefore never can be NULL.
In principle, one could remove the whole check. Of course,
realizing that would require a global analysis by Coverity,
which I reckon it doesn't do.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a valid, but harmless, Coverity warning.

### DIFF
--- a/desktop-widgets/filterconstraintwidget.cpp
+++ b/desktop-widgets/filterconstraintwidget.cpp
@@ -328,7 +328,7 @@ void FilterConstraintWidget::update()
 	}
 
 	// Update the unit strings in case the locale was changed
-	if (unitFrom || unitTo) {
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
Coverity is 90% noise and 10% really really good stuff.
I appreciate that you are closing some of them. A LOT!!!

PR #3398, opened 2022-02-15
Cleanups

dirkhh:

- [x] Code cleanup
- [x] Build system change

second attempt as the first one included to commits that shouldn't have been in there

all this should work as I tested it locally (on Linux) against Q t5.9 and Qt 5.15 - but as the last attempt shows, that doesn't mean I got it right when pushing.


dirkhh:
@bstoeger - most of these changes are trivial and obvious (or just plain stupid).
I was hoping to get your critical eye on especially a couple of them.
The conversion warnings - simply because that's always an area where taste matters and I want a second opinion.
The 'unused argument and private member' one and the 'use already defined constant' one. Those are so obvious that I'm worried that I'm missing something :)

stats/statsvariables.cpp
```diff
@@ -113,15 +113,14 @@ template<> int invalid_value<int>()
 }
 template<> double invalid_value<double>()
 {
-	return std::numeric_limits<double>::quiet_NaN();
```
bstoeger, line 116: Yes, that makes sense.

desktop-widgets/divelogexportdialog.cpp
```diff
@@ -246,7 +246,7 @@ void DiveLogExportDialog::on_buttonBox_accepted()
 		qPrefDisplay::set_lastDir(fileInfo.dir().path());
 		// the non XSLT exports are called directly above, the XSLT based ons are called here
 		if (!stylesheet.isEmpty()) {
-			QFuture<void> future = exportUsingStyleSheet(filename, ui->exportSelected->isChecked(),
```
bstoeger, line 249: I reckon it worked because `int` can be cast to `void`. This is perfectly fine (reasonably modern) C++:
```
void test(int num)
{
    return (void)num;
}
```

The documentation of `QFuture` says:
> QFuture\<void\> is specialized to not contain any of the result fetching functions. Any QFuture\<T\> can be assigned or copied into a QFuture\<void\> as well. This is useful if only status or progress information is needed - not the actual result data.

So I wonder if this is a regression in Qt6?

core/color.h
```diff
@@ -7,65 +7,67 @@
 
 #include <QColor>
 
+#define MAKEQCOLOR(r,g,b,a) QColor::fromRgbF((float)(r), (float)(b), (float)(g), (float)(a))
```
bstoeger, line 10: I think you inverted blue and green here...? That must look funky?

Personally, I would find an inline function easier to read, such as
```
inline QColor makeColor(double r, double g, double b, double a = 1.0)
{
    return QColor::fromRgbF((float)r, (float)g, float(b), (float)a);
}
```
Thus, you could also remove the default `1` argument in most cases. But potato, patato.

profile-widget/divepercentageitem.cpp
```diff
@@ -52,29 +52,30 @@ static std::array<int, num_tissues> calcLinesPerTissue(int size)
...
 }
 
+#define SETHSVF(h, s, v) setHsvF((float)(h), (float)(s), (float)(v))
 static QRgb colorScale(double value, int inert)
 {
 	QColor color;
 	double scaledValue = value / (AMB_PERCENTAGE * inert) * 1000.0;
 	if (scaledValue < 0.8)	// grade from cyan to blue to purple
-		color.setHsvF(0.5 + 0.25 * scaledValue / 0.8, 1.0, 1.0);
+		color.SETHSVF(0.5 + 0.25 * scaledValue / 0.8, 1.0, 1.0);
```
bstoeger, line 61: Wow - that is brutal.

profile-widget/divepixmapcache.cpp
```diff
@@ -49,7 +49,7 @@ DivePixmaps::DivePixmaps(int dpr) : dpr(dpr)
 	// The transparen pixmap is a very obscure feature to enable tooltips without showing a pixmap.
 	// See code in diveeventitem.cpp. This should probably be replaced by a different mechanism.
 	QPixmap transparentPixmap(lrint(4 * dprf), lrint(20 * dprf));
-	transparentPixmap.fill(QColor::fromRgbF(1.0, 1.0, 1.0, 0.01));
+	transparentPixmap.fill(QColor::fromRgbF(1.0f, 1.0f, 1.0f, 0.01f));
```
bstoeger, line 52: Could reuse the inline function (or macro) in `color.h`.

bstoeger (commented):
I'm confused about the `QColor` changes. The function declarations in the documentation read as
```
static QColor fromHsvF(qreal h, qreal s, qreal v, qreal a = 1.0);
void setHsvF(qreal h, qreal s, qreal v, qreal a = 1.0);
```

I was under the impression that `qreal` is always a `double` and therefore I wonder how this could have given conversion warnings. The documentation says that there is a `-qreal float` flag. Is that enabled in some of your Qt builds? That would be most peculiar.

For me this is definitely not the case:
```
printf("double: %lu, float: %lu, qreal: %lu\n", sizeof(double), sizeof(float), sizeof(qreal));
```
gives
```
double: 8, float: 4, qreal: 8
```

And that means that the new code is a pessimization if it uses non-constant values: `double`s are pointlessly truncated even though they are then passed as double. Therefore, you shouldn't cast to `float`, but to `qreal`, which will be a noop in the (majority?) of cases.

Once that is cleared up, consider this PR as approved.

bstoeger:
In fact
```
double test(double num)
{
    return (float)num;
}
```
compiles to
```
        cvtsd2ss        xmm0, xmm0
        cvtss2sd        xmm0, xmm0
        ret
```
🤷

dirkhh:
> I'm confused about the `QColor` changes. The function declarations in the documentation read as
> 
> ```
> static QColor fromHsvF(qreal h, qreal s, qreal v, qreal a = 1.0);
> void setHsvF(qreal h, qreal s, qreal v, qreal a = 1.0);

in Qt6, however:

QColor	[fromHsvF](https://doc.qt.io/qt-6/qcolor.html#fromHsvF)(float h, float s, float v, float a = 1.0)

these are now floats :(

desktop-widgets/divelogexportdialog.cpp
dirkhh: seems like it. But on the flip side - it seems that my change isn't 'wrong'...

core/color.h
dirkhh: that was very clever of me to mess with the color components...
And I'm happy to do an inline function instead...
And to have it use the `float` cast only in Qt6

profile-widget/divepercentageitem.cpp
dirkhh: I was hoping you'd go with _subtle_ - but brutal isn't wrong...

profile-widget/divepixmapcache.cpp
dirkhh: true

bstoeger:
> in Qt6, however:
> 
> QColor [fromHsvF](https://doc.qt.io/qt-6/qcolor.html#fromHsvF)(float h, float s, float v, float a = 1.0)
> 
> these are now floats :(

Ok... Now they're trolling us.

Whatever.

core/color.h
dirkhh: fixed based on your recommendation

profile-widget/divepercentageitem.cpp
dirkhh: replaced with something less brutal

profile-widget/divepixmapcache.cpp
dirkhh: done

dirkhh:
I'll update the Qt6 branch, but not bother to update this (which was just to make it more obvious which part of the PR was doing what)

PR #3399, opened 2022-02-16
Qt6

dirkhh:

- [x] Code cleanup
- [x] Build system change

this is now rebased on the splitout cleanups branch #3398 

Qt6 support is still problematic
- no maps
- the statistics code breaks (somehow the integration via QML fails and things just crash)

### Changes made:

- make things work with Qt6 (and in the process, the M1 based Macs)
- this should still run fine on Qt5, but that of course requires more testing

### Release note:

missing

### Documentation change:

not sure if we even want to document this as right now this is a clear step backwards from using Qt5

### Mentions:

@bstoeger


profile-widget/divepercentageitem.cpp
```diff
@@ -52,29 +52,30 @@ static std::array<int, num_tissues> calcLinesPerTissue(int size)
...
 }
 
+#define SETHSVF(h, s, v) setHsvF((float)(h), (float)(s), (float)(v))
 static QRgb colorScale(double value, int inert)
 {
 	QColor color;
 	double scaledValue = value / (AMB_PERCENTAGE * inert) * 1000.0;
 	if (scaledValue < 0.8)	// grade from cyan to blue to purple
-		color.setHsvF(0.5 + 0.25 * scaledValue / 0.8, 1.0, 1.0);
+		color.SETHSVF(0.5 + 0.25 * scaledValue / 0.8, 1.0, 1.0);
```
bstoeger: I would suggest the following (not even compile tested, but you get the point):
```
static QRgb HSV2RGB(double r, double g, douible b, double a = 1.0)
{
        QColor c = QColor.fromHsvf((float)r, (float)g, (float)b, float(a));
        return c.rgba();
}
```
and then do the different if-branches as such:
```
        if (scaledValue < 0.8) // grade from cyan to blue to purple
                return HSV2RGB(0.75, 1.0, (1.0 - scaledValue) / 0.2);
        ...
```
Seems so much more clear.

dirkhh: the fact that your `hsv2rgb()` function has `r, g, b, a` arguments is a tag surprising... but I think I get your point.

dirkhh: modified and it seems to work now
force push coming up

dirkhh:
so while in my first quick test it seemed that the ownership change (last commit) was enough to prevent the crash, in a last second quick test before pushing I could once again produce a crash in Qt6.
Since this incorporates all the other suggested changes, I pushed anyway, but that last commit isn't sufficient, it seems.

dirkhh:
This has received some more testing with Qt5 and doesn't seem to break anything that I was able to find.
I'll do another run through it today, reviewing the patches myself and trying to make sure nothing stupid sneaks through.

@bstoeger I don't know how things are for you right now, if you have time, I'd appreciate another look from you as well
@atdotde I assume you are still mostly away?

dirkhh:
OK, I can show conclusively that even with `QQmlEngine::CppOwnership` the `StatsView` gets destroyed and reallocated whenever the parent of the `StatsWidget` changes.
So I need a different solution for that last commit.

dirkhh:
This latest version seems to work in my tests. It's not pretty, but what can I say. The thing that the documentation says _should_ work, doesn't. Or maybe I just don't understand what causes the `StatsView` object to be destroyed.

dirkhh:
I assume that either everyone loves this, or no one is looking at this. Either way, the obviously correct thing to do is to merge it...

PR #3401, opened 2022-02-19
profile: fix ESC in planner

bstoeger:
When in the planner, ESC should cancel the plan.

However, when the user manipulates the dive-handles in the
profile and presses ESC, first nothing happens, then an obscure
message appears.

The reason is that ESC "shortcuts" are introduced in two places.

To fix this, remove the ESC shortcut in the profile (the planner
widget cancels the plan anyway) and replace all the shortcuts in
the profile with a simple override of the keyPressEvent().

The latter is not strictly necessary, but hopefully avoids further
complications with multiple shortcuts. And the code is easier
to follow too.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a tiny UI issue: escape doesn't always work in planner. I wonder if we should replace the other ESC shortcuts as well or at least set their mode to "widget and children" instead of "window".

### Release note:

Done.

### Documentation change:

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- planner: make ESC (cancel plan) work when moving handles
 - dive list: make dive guide visible in dive list [#3382]
 - general: rename dive master to dive guide
 - desktop: Don't lose cursor position in notes when switching between windows [#3369]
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
this does seem reasonable

PR #3402, opened 2022-02-19
Allow editing sensors through equipment tab

michaelandreen:
My main use case is that I'm diving with both a Garmin and a Suunto. They each have one transmitter, connected to different cylinders. When I download the dives both transmitters gets assigned to the first cylinder, so need to change one of the sensors to the second cylinder.

I'm not sure how much of a hack this is, and how useful it is for a wider audience. I initially thought of adding a separate dialog, since I could then have one cylinder per sensor, but I realized that it was fairly easy to add a list of sensors to each cylinder like this. I think the editing turned out to be more intuitive than I initially expected and seems non-intrusive.
 
It seems to populate some undo buffer, but currently the undo doesn't do anything.

I guess something also has to be done about CCR and o2sensor, not sure if this breaks that use case.

Maybe sensors should be hidden by default?

There are some case blocks in the cylinder model that have deeper indentation than the coding style suggests. It confuses my clang format plugin, but I left those alone to keep the diff small. Same for the re-ordering of headers that clang-format wanted to do.

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -4,6 +4,7 @@
 #include "core/divelist.h"
 #include "core/fulltext.h"
 #include "core/qthelper.h" // for copy_qstring
+#include "core/sample.h"
 #include "core/selection.h"
 #include "core/subsurface-string.h"
 #include "core/tag.h"
@@ -1283,6 +1284,43 @@ void EditCylinder::redo()
 void EditCylinder::undo()
 {
 	redo();
+}
+
+EditSensors::EditSensors(int toCylinderIn, int fromCylinderIn)
+	: dc(current_dc), d(current_dive), toCylinder(toCylinderIn), fromCylinder(fromCylinderIn)
+{
+	if (!d || !dc)
+		return;
+
+	setText(Command::Base::tr("Edit sensors"));
+
+}
+
+void EditSensors::mapSensors(int toCyl, int fromCyl)
+{
+	for (int i = 0; i < dc->samples; ++i) {
+		for (int s = 0; s < MAX_SENSORS; ++s) {
+			if (dc->sample[i].pressure[s].mbar && dc->sample[i].sensor[s] == fromCyl)
+				dc->sample[i].sensor[s] = toCyl;
+		}
+	}
+	emit diveListNotifier.diveComputerEdited(dc);
+	invalidate_dive_cache(d); // Ensure that dive is written in git_save()
+}
+
+void EditSensors::undo()
+{
+	mapSensors(fromCylinder, toCylinder);
+}
+
+void EditSensors::redo()
+{
+	mapSensors(toCylinder, fromCylinder);
+}
+
+bool EditSensors::workToBeDone()
+{
+	return d && dc;
 }
 
 #ifdef SUBSURFACE_MOBILE
--- a/commands/command_edit.h
+++ b/commands/command_edit.h
@@ -424,6 +424,22 @@ private:
 	void redo() override;
 };
 
+class EditSensors : public Base
+{
+public:
+	EditSensors(int cylIndex, int fromCylinder);
+
+private:
+	struct divecomputer *dc;
+	struct dive *d;
+	int toCylinder;
+	int fromCylinder;
+	void mapSensors(int toCyl, int fromCyl);
+	void undo() override;
+	void redo() override;
+	bool workToBeDone() override;
+};
+
 #ifdef SUBSURFACE_MOBILE
 // Edit a full dive. This is used on mobile where we don't have per-field granularity.
 // It may add or edit a dive site.
--- a/desktop-widgets/tab-widgets/TabDiveEquipment.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveEquipment.cpp
@@ -20,6 +20,15 @@ static bool ignoreHiddenFlag(int i)
 	       i == CylindersModel::WORKINGPRESS_INT || i == CylindersModel::SIZE_INT;
 }
 
+static bool hiddenByDefault(int i)
+{
+	switch (i) {
+	case CylindersModel::SENSORS:
+		return true;
+	}
+	return false;
+}
+
 TabDiveEquipment::TabDiveEquipment(QWidget *parent) : TabBase(parent),
 	cylindersModel(new CylindersModel(false, true, this)),
 	weightModel(new WeightModel(this))
@@ -81,7 +90,8 @@ TabDiveEquipment::TabDiveEquipment(QWidget *parent) : TabBase(parent),
 	for (int i = 0; i < CylindersModel::COLUMNS; i++) {
 		if (ignoreHiddenFlag(i))
 			continue;
-		bool checked = s.value(QString("column%1_hidden").arg(i)).toBool();
+		auto setting = s.value(QString("column%1_hidden").arg(i));
+		bool checked = setting.isValid() ? setting.toBool() : hiddenByDefault(i) ;
 		QAction *action = new QAction(cylindersModel->headerData(i, Qt::Horizontal, Qt::DisplayRole).toString(), ui.cylinders->view());
 		action->setCheckable(true);
 		action->setData(i);
--- a/qt-models/cylindermodel.cpp
+++ b/qt-models/cylindermodel.cpp
@@ -7,6 +7,7 @@
 #include "core/color.h"
 #include "qt-models/diveplannermodel.h"
 #include "core/gettextfromc.h"
+#include "core/sample.h"
 #include "core/subsurface-qt/divelistnotifier.h"
 #include "core/subsurface-string.h"
 #include <string>
@@ -19,9 +20,10 @@ CylindersModel::CylindersModel(bool planner, bool hideUnused, QObject *parent) :
 	tempRow(-1),
 	tempCyl(empty_cylinder)
 {
-	//	enum {REMOVE, TYPE, SIZE, WORKINGPRESS, START, END, O2, HE, DEPTH, MOD, MND, USE, IS_USED, WORKINGPRESS_INT, SIZE_INT};
+	//	enum {REMOVE, TYPE, SIZE, WORKINGPRESS, START, END, O2, HE, DEPTH, MOD, MND, USE, WORKINGPRESS_INT, SIZE_INT, SENSORS};
 	setHeaderDataStrings(QStringList() << "" << tr("Type") << tr("Size") << tr("Work press.") << tr("Start press.") << tr("End press.") << tr("O₂%") << tr("He%")
-						 << tr("Deco switch at") <<tr("Bot. MOD") <<tr("MND") << tr("Use") << "" << "");
+					   << tr("Deco switch at") << tr("Bot. MOD") << tr("MND") << tr("Use") << ""
+					   << "" << tr("Sensors"));
 
 	connect(&diveListNotifier, &DiveListNotifier::cylindersReset, this, &CylindersModel::cylindersReset);
 	connect(&diveListNotifier, &DiveListNotifier::cylinderAdded, this, &CylindersModel::cylinderAdded);
@@ -237,6 +239,24 @@ QVariant CylindersModel::data(const QModelIndex &index, int role) const
 			return static_cast<int>(cyl->type.workingpressure.mbar);
 		case SIZE_INT:
 			return static_cast<int>(cyl->type.size.mliter);
+		case SENSORS: {
+			std::vector<int16_t> sensors;
+			for (int i = 0; i < current_dc->samples; ++i) {
+				auto &sample = current_dc->sample[i];
+				for (auto s = 0; s < MAX_SENSORS; ++s) {
+					if (sample.pressure[s].mbar) {
+						if (sample.sensor[s] == index.row())
+							return tr("Sensor attached, can't move another sensor here.");
+						else if (std::find(sensors.begin(), sensors.end(), sample.sensor[s]) == sensors.end())
+							sensors.push_back(sample.sensor[s]);
+					}
+				}
+			}
+			QStringList sensorStrings;
+			for (auto s : sensors)
+				sensorStrings << QString::number(s);
+			return tr("Select one of these cylinders: ") + sensorStrings.join(",");
+		}
 		}
 		break;
 	case Qt::DecorationRole:
@@ -272,6 +292,8 @@ QVariant CylindersModel::data(const QModelIndex &index, int role) const
 			return tr("Calculated using Bottom pO₂ preference. Setting MOD adjusts O₂%, set to '*' for best O₂% for max. depth.");
 		case MND:
 			return tr("Calculated using Best Mix END preference. Setting MND adjusts He%, set to '*' for best He% for max. depth.");
+		case SENSORS:
+			return tr("Index of cylinder that you want to move sensor data from.");
 		}
 		break;
 	}
@@ -453,6 +475,24 @@ bool CylindersModel::setData(const QModelIndex &index, const QVariant &value, in
 		}
 		type = Command::EditCylinderType::TYPE;
 		break;
+	case SENSORS: {
+		std::vector<int> sensors;
+		for (auto &sensor : vString.split(",")) {
+			bool ok = false;
+			int s = sensor.toInt(&ok);
+			if (ok && s < MAX_SENSORS)
+				sensors.push_back(s);
+		}
+
+		bool ok = false;
+		int s = vString.toInt(&ok);
+		if (ok) {
+			Command::editSensors(index.row(), s);
+			// We don't use the edit cylinder command and editing sensors is not relevant for planner
+			return true;
+		}
+		return false;
+	}
 	}
 
 	if (inPlanner) {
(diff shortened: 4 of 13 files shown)


packetninja:
+1 from me on this feature. :-)

michaelandreen:
Added undo/redo. Feels a bit less like a hack now. Added a separate EditSensors command, since it didn't feel natural to integrate it into EditCylinder.

I think it starts to get really convoluted to edit the sensors for multiple dives and multiple dive computers at the same time, so it only does it for the current dive computer. I'm not sure if this should be reflected in the UI somehow. Maybe make the sensors field hidden when multiple dives are selected or making it not editable?

dirkhh:
Interesting... not a use case I ever run into, frankly :)
I'll take a look soon (a bit overwhelmed with other items on my todo list right now... real life interfering with my priorities, so to speak)

commands/command_edit.cpp
```diff
@@ -1285,6 +1286,56 @@ void EditCylinder::undo()
...
+EditSensors::EditSensors(int cylIndex, divecomputer *dcIn, std::vector<int> sensors)
+	: EditDivesBase(true), dc(dcIn)
+{
+	if (dives.empty())
+		return;
+
+	if (dives.size() == 1)
+		setText(Command::Base::tr("Edit sensors"));
+	else
+		throw std::runtime_error("Can only edit sensors for current dive computer");
```
dirkhh: In general we don't throw runtime errors. Instead we try to fail gracefully (for some value of that term).
I'm not sure how you would get here with more than one dive selected... I'd prefer to disable THAT instead of trying to bail from here.

commands/command_edit.cpp
```diff
@@ -1285,6 +1286,56 @@ void EditCylinder::undo()
...
+
+	for (int s = 0; s < MAX_SENSORS; ++s) {
+		if (std::find(sensors.begin(), sensors.end(), s) != sensors.end()) {
+			undoMapping.emplace_back(s, dc->sample[0].sensor[s]);
+			redoMapping.emplace_back(s, cylIndex);
+		} else if (dc->sample[0].sensor[s] == cylIndex) {
+			undoMapping.emplace_back(s, cylIndex);
+			redoMapping.emplace_back(s, NO_SENSOR);
+		}
```
dirkhh: that code desperately needs some commenting. I've been staring at this for a couple of minutes and really don't know what you are doing here...

dirkhh (commented):
I really dislike the runtime error.
And I'd love to see some comments so that I can visualize what you are doing there.

commands/command_edit.cpp
bstoeger: Yes, unfortunately Qt is fundamentally not exception safe. And owing to its completely broken ownership model it can never be made exception safe.

I will try to look at the rest of the PR later this week.

michaelandreen: Yea. I basically just added it as an assert so I would notice if I triggered it. The command is currently being hard-coded to just be for one dive. It feels a bit strange when you have multiple dives selected and change the sensor and it just changes the current dive computer. It doesn't trigger the warning about multiple dives being edited and maybe there is something else with similar behavior. I've been looking into hiding or making the sensors field uneditable when multiple dives are selected but haven't figured that out yet.

commands/command_edit.cpp
michaelandreen: Maybe the new comments clears things up a bit.

commands/command_edit.cpp
```diff
@@ -1285,6 +1286,62 @@ void EditCylinder::undo()
...
+	: EditDivesBase(true), dc(dcIn)
+{
+	if (dives.empty())
+		return;
+
+	if (dives.size() == 1)
+		setText(Command::Base::tr("Edit sensors"));
+	else
+		// This shouldn't happen.
+		setText(Command::Base::tr("Edit sensors, only single dive computer"));
```
michaelandreen: Would be visible like this, but I guess it's a waste to translate it. Maybe just remove the whole if since it's always 1 dive now.

commands/command_edit.h
```diff
@@ -424,6 +424,21 @@ class EditCylinder : public EditCylinderBase {
 	void redo() override;
 };
 
+class EditSensors : public EditDivesBase
```
bstoeger: I think that `EditDivesBase` is the wrong base class here. I would model it after the `ReplanDive` command. Then these multiple- vs. one-dive issues should go away.

commands/command_edit.cpp
```diff
@@ -1285,6 +1286,58 @@ void EditCylinder::undo()
...
+}
+
+void EditSensors::mapSensors(std::vector<std::pair<int, int>> &mapping)
+{
+	for (int i = 0; i < dc->samples; ++i) {
+		for (auto &[s, cyl] : mapping) {
+			dc->sample[i].sensor[s] = cyl;
+		}
+	}
+	emit diveListNotifier.deviceEdited();
```
bstoeger: I wonder if `cylindersReset` would be more appropriate. Or a new signal?

qt-models/cylindermodel.cpp
```diff
@@ -453,6 +466,19 @@ bool CylindersModel::setData(const QModelIndex &index, const QVariant &value, in
...
+	case SENSORS: {
+		std::vector<int> sensors;
+		for (auto &sensor : vString.split(",")) {
+			bool ok = false;
+			int s = sensor.toInt(&ok);
+			if (ok && s < MAX_SENSORS)
+				sensors.push_back(s);
+		}
+		int count = Command::editSensors(index.row(), current_dc, std::move(sensors));
+		emit divesEdited(count);
```
bstoeger: Not necessary, if only one dive edited or is it?

commands/command_edit.cpp
bstoeger: I never understood the sensors code. I will just trust that this does the right thing.

commands/command_edit.cpp
```diff
@@ -1285,6 +1286,58 @@ void EditCylinder::undo()
...
+		} else if (dc->sample[0].sensor[s] == cylIndex) {
+			// Otherwise, if the sensor was previously attached to this cylinder then
+			// save that in the undoMapping and set it to NO_SENSOR in redoMapping so
+			// it is not attached to any cylinder.
+			undoMapping.emplace_back(s, cylIndex);
+			redoMapping.emplace_back(s, NO_SENSOR);
+		}
+	}
+
+	mapSensors(redoMapping);
```
bstoeger: Not here, please. Any action is performed in redo()/undo().

bstoeger (changes requested):

commands/command_edit.h
michaelandreen: It makes the code easier to understand. Main question with multiple- vs one-dive is the UI though, not sure how much of an issue it is though.

commands/command_edit.cpp
michaelandreen: `cylindersReset` doesn't replot the profile, which is what I needed to emit the signal for. Adding a diveComputerEdited signal, then it could be changed so the replotting isn't done if you undo the edit while viewing another dive computer, not sure if that change would be worth the added complexity though.

commands/command_edit.cpp
michaelandreen: Ah, so that's how the redo()/undo() system works.

desktop-widgets/tab-widgets/TabDiveEquipment.cpp
```diff
@@ -81,7 +81,8 @@ TabDiveEquipment::TabDiveEquipment(QWidget *parent) : TabBase(parent),
 	for (int i = 0; i < CylindersModel::COLUMNS; i++) {
 		if (ignoreHiddenFlag(i))
 			continue;
-		bool checked = s.value(QString("column%1_hidden").arg(i)).toBool();
+		auto setting = s.value(QString("column%1_hidden").arg(i));
+		bool checked = !setting.isValid() && i == CylindersModel::SENSORS ? true : setting.toBool();
```
michaelandreen: So played around with hiding the sensors by default. While doing this I noticed that the depth (deco switch at) column used to be hidden by default, but isn't anymore. Not sure what is best long term. It's easy to discover features if they're visible by default and you can hide them if you don't need them. At the same time, there is no point in showing things by default that isn't useful for most users and with every new column it starts getting crowded.

dirkhh:
This is still marked as WIP - @bstoeger, any more feedback?
I want to avoid that we let a new contributor 'hang' here for too long (but, @michaelandreen -- things sometimes take a few days since all of us have day jobs that compete for our time...)

desktop-widgets/tab-widgets/TabDiveEquipment.cpp
bstoeger: Careful: The CylindersModel code is a complete disaster, because it is used in two rather different places: in the planner and the equipment tab.

I could imagine that the column should not be shown in the equipment tab at all.

bstoeger (approved):
I am happy with the undo code now.

I haven't done any testing.

Only two comments:

- Please be very careful with the cylinders-model. I inadvertently messed up the planner times with changes there. I really wonder whether this should be split in two distinct models.

- I think this needs documentation.

desktop-widgets/tab-widgets/TabDiveEquipment.cpp
michaelandreen: This code didn't seem to be used in the planner and the column was always visible there, so I made it always hidden in the planner.

Still for the non-planner case, is this how we want to have it? Keep it hidden by default and people who might need it can discover it later? If so, should i make it more generic and possibly hide some other column by default?

michaelandreen:
@dirkhh No problem. I'm not in a hurry and things like profile scaling on the android app probably has more users being impacted =)

desktop-widgets/tab-widgets/TabDiveEquipment.cpp
michaelandreen: Decided to keep it hidden by default since I didn't get a reply to this. Also made it a bit more generic so it's easier to add new column as hidden by default if that need comes up.

qt-models/cylindermodel.cpp
```diff
@@ -237,6 +239,15 @@ QVariant CylindersModel::data(const QModelIndex &index, int role) const
 			return static_cast<int>(cyl->type.workingpressure.mbar);
 		case SIZE_INT:
 			return static_cast<int>(cyl->type.size.mliter);
+		case SENSORS: {
+			QStringList sensors;
+			auto *sample = current_dc->sample;
+			for (int s = 0; sample && s < MAX_SENSORS; ++s) {
```
michaelandreen: So i realized that just looking at the first sample doesn't work in all cases. It'll probably be correct for unedited dives, but when the dive is saved only the first sample with pressure data will save the sensor information, so when loading a dive like that the display will be wrong.

qt-models/cylindermodel.cpp
```diff
@@ -237,6 +239,28 @@ QVariant CylindersModel::data(const QModelIndex &index, int role) const
...
 			return static_cast<int>(cyl->type.size.mliter);
+		case SENSORS: {
+			QStringList sensorStrings;
+			std::vector<int> allSensors(MAX_SENSORS);
+			std::iota(allSensors.begin(), allSensors.end(), 0);
+			for (int i = 0; i < current_dc->samples; ++i) {
+				auto &sample = current_dc->sample[i];
+				for (auto it = allSensors.begin(); it != allSensors.end();) {
+					int s = *it;
+					if (sample.pressure[s].mbar) {
```
michaelandreen, line 247: So now it'll search for the first sample with pressure information and see which cylinder the sensor is attached to.

I wonder if maybe the sensor array should be moved from sample to divecomputer. Does it ever make sense to have the same sensor change cylinder during a dive? Technically someone could remove the regulator from one cylinder and attach it to another during the dive, but it would be very rare and the cases with a sensor attached to the regulator likely even rarer. The file format could still stay as it is, even though it would let you store information that isn't used.

michaelandreen: So looking a bit more into this it looks like I got the data model for the sensors wrong. I treated the index in the sensor array as a sensor id containing the cylinder this sensor was connected to. This works for my use case of having two dive computers with one transmitter each, since the transmitter is always in the first index, but with more transmitter per dive computer this doesn't work. The sensor index has no real meaning, it just shows that the pressure at index X is for cylinder at sensor index X. So for one sample index 0 could contain cylinder 0 and index 1 cylinder 1, for the next sample index 0 could be cylinder 1 and index 1 cylinder 2.

michaelandreen: So, it actually seems to work in most cases with up to two sensors for a dive computer, since add_sample_pressure() tries to maintain the index.

Two exceptions seems to be the liquidvision and uemis parsers, that always puts the active sensor in index 0.

commands/command_edit.cpp
```diff
@@ -1285,6 +1286,70 @@ void EditCylinder::undo()
...
 
+EditSensors::EditSensors(int cylIndex, const std::vector<int> &changedSensors)
+	: dc(current_dc), d(current_dive)
+{
+	if (!d || !dc)
+		return;
+
+	setText(Command::Base::tr("Edit sensors"));
+
+	std::vector<int> allSensors(MAX_SENSORS);
```
bstoeger: Just an extremely minor nit-pick: since `MAX_SENSORS` is a compile-time constant, one might prefer `std::array<int, MAX_SENSORS>` here.

bstoeger: Ok disregard previous comment - you then erase from the container. vector is the correct container here.

qt-models/cylindermodel.cpp
```diff
@@ -237,6 +239,28 @@ QVariant CylindersModel::data(const QModelIndex &index, int role) const
...
+			sensorsEditable = false;
+			std::vector<int16_t> sensors(MAX_SENSORS, NO_SENSOR);
+			for (int i = 0; i < current_dc->samples; ++i) {
+				auto &sample = current_dc->sample[i];
+				for (auto s = 0; s < MAX_SENSORS; ++s) {
+					if (sample.pressure[s].mbar) {
+						if (sensors[s] == NO_SENSOR)
+							sensors[s] = sample.sensor[s];
+						else if (sensors[s] != sample.sensor[s])
+							return tr("Sensor data too complex for this UI.");
```
bstoeger: If this return triggers, then `sensorEditable` is not reset to `true`. 

In #3410 I do this:
```
// Silly RAII-variable setter class: reset variable when going out of scope.
template <typename T>
struct Setter {
	T &var, old;
	Setter(T &var, T value) : var(var), old(var) {
		var = value;
	}
	~Setter() {
		var = old;
	}
};
```
and then something like
```
Setter s(sensorEditable, true);
```

Yes, I am kind of ashamed of it, but it should work.

michaelandreen: Actually, not setting sensorEditable to true is the intended behavior. When the data is too complex it doesn't allow editing, since it would mess things up and not be able to undo it.

I realized that I can make things work properly within the model though. So doing a rework of this that should work for all cases.

bstoeger: I see - that is super-subtle, as it assumes a certain call order [data() before flags()], no?

michaelandreen: Ah yea, that's true. Just seemed like the natural thing and it worked when I tried it out. I was wondering if data had to be const and just made the attribute mutable while I was playing around though. I guess I either have to put similar logic in flags() or calculate and cache it in some other place.

dirkhh:
I've played with it and it seems to work as expected. @bstoeger any more change requests from you?

bstoeger:
I don't see anything obviously wrong, but I also don't really know what this does. I don't have a DC with sensors.

Only one style comment: according to the coding style, no braces around single-line blocks (l. 256 in cylindermodel.cpp).

michaelandreen:
Fixed the braces and rebased on top of latest master.

dirkhh:
Thanks @michaelandreen -- I have spent some time with this today and I struggle with a good use case where this makes sense. Can you add a commit that adds a test dive in the `dives` directory (like `test52.xml` or `testsensors.xml`) and have the commit message give instructions how a user would interact with this?

With all my dives I get this additional column in the equipment tab that has these very long texts that even on 30" screen I can't really read. The tank with the attached sensor tells me that I can't move the sensor, the other tanks give me a long text  `Select one of these cylinders: 0` and I have no idea how I would interact with that... So I must be missing something (what else is new). And reading through your documents I think the core problem for me to understand how this is supposed to work is that I don't have a dive that matches the scenario that you describe... one attached sensor on each of the dive computers.

michaelandreen:
@dirkhh added a test dive called `testsensormove.xml` that data that need to be fixed and instructions in the commit message.

In your dives, you can use the functionality. When it says `Select one of these cylinders: 0` you basically change that to 0 and press enter to move the sensor to this tank. The text in this column is fairly long, but I tried to be as explicit as possible there and in the tooltip. I imagine that this column will only be visible when you need to fix the data and the rest of the time it's meaningless so people keep it hidden.

dirkhh:
Thank you - the UI is... unexpected... I think I would want shorter text in the field... but we are bike shedding now. The pull request does what you advertise, it solves a problem that you have... so I'll merge this.

dirkhh:
merged locally.
One thing I should have mentioned... it would be nice if you could document this in the user manual. If you can't find where / how to do that, I'll be happy to help.

Thanks

michaelandreen:
I'll see if I can find a good way to document it. It probably makes more sense to to do it with the sidemount dive that I got from Attila as base, since that's more likely a use case that someone else will bump into. Or a hypoxic dive where the sensor is used for bottom gas, but the travel gas ends up as cylinder 0 during the download, not sure if I have such a profile to look at though.

One potential improvement to the UI could be to have a combo box with the available sensors, similar to how the cylinder types are listed when you edit that field. I'm not really sure how to do that though, browsing through the code it looked much more complex than it would be worth for this fairly niche use case.

packetninja:
> On 13 Mar 2022, at 03:03, michaelandreen ***@***.***> wrote:
> 
> 
> I'll see if I can find a good way to document it. It probably makes more sense to to do it with the sidemount dive that I got from Attila as base, since that's more likely a use case that someone else will bump into. Or a hypoxic dive where the sensor is used for bottom gas, but the travel gas ends up as cylinder 0 during the download, not sure if I have such a profile to look at though.
> 
> One potential improvement to the UI could be to have a combo box with the available sensors, similar to how the cylinder types are listed when you edit that field. I'm not really sure how to do that though, browsing through the code it looked much more complex than it would be worth for this fairly niche use case.

I’d be happy to help, but I’m in Thailand now. I’ll am planning another Hypoxic dive though, so there will be data. :-)


— Attilla

PR #3406, opened 2022-02-26
Android: share log files as attachments

dirkhh:

- [x] Functional change
- [x] New feature
- [x] Code cleanup
- [x] Build system change

This is something that I wanted to do for a while, and @jefdriesen ended up prompting me to raise the priority and to work on this - because today our log files in support emails are truncated since they are passed via the Binder as message body (and some versions of the Gmail app make things worse because they strip all newlines from the logs)

This creates a Java helper to share files as attachments - and cleverly this can also share files to something like Dropbox - so a fairly logical and easy extension here would be to add the ability to share dive log XML files (e.g. via Dropbox) in order to back things up... allowing bidirectional sharing via Dropbox would be quite a bit harder and is left as an exercise for the reader.

### Changes made:

- add Java helper to share file(s)
- try using that, first, when creating support emails on Android

### Release note:

done

### Documentation change:

yeah, I think we should document that as the user experience no Android changed a little bit. Instead of simply opening the default email app, the user is now asked to pick an app to share the files with and if they do want to create a support email, the (obviously, I guess) need to pick an email app...

### DIFF
--- a/profile-widget/qmlprofile.cpp
+++ b/profile-widget/qmlprofile.cpp
@@ -51,7 +51,10 @@ void QMLProfile::paint(QPainter *painter)
 		timer.start();
 
 	// let's look at the intended size of the content and scale our scene accordingly
+	// for some odd reason the painter transformation is set up to scale by the dpr - which results
+	// in applying that dpr scaling twice. So we hard-code it here to be the identity matrix
 	QRect painterRect = painter->viewport();
+	painter->resetTransform();
 	if (m_diveId < 0)
 		return;
 	struct dive *d = get_dive_by_uniq_id(m_diveId);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- mobile: fix profile scaling issue on high DPR devices
 - mobile/Android: add logfiles as attachment to support emails
 - planner: make ESC (cancel plan) work when moving handles
 - dive list: make dive guide visible in dive list [#3382]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- mobile/Android: add logfiles as attachment to support emails
 - planner: make ESC (cancel plan) work when moving handles
 - dive list: make dive guide visible in dive list [#3382]
 - general: rename dive master to dive guide
--- a/android-mobile/AndroidManifest.xml
+++ b/android-mobile/AndroidManifest.xml
@@ -77,6 +77,15 @@
 	    <meta-data android:name="android.max_aspect" android:value="3" />
 
         </activity>
+	<provider
+            android:name="android.support.v4.content.FileProvider"
+            android:authorities="org.subsurfacedivelog.mobile.fileprovider"
+            android:grantUriPermissions="true"
+            android:exported="false">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/filepaths" />
+        </provider>
     </application>
 
     <uses-sdk android:minSdkVersion="21"
--- a/android-mobile/build.gradle
+++ b/android-mobile/build.gradle
@@ -27,8 +27,9 @@ allprojects {
 apply plugin: 'com.android.application'
 
 dependencies {
-    compile fileTree(dir: 'libs', include: ['*.jar'])
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'com.github.mik3y:usb-serial-for-android:v3.4.3'
+    implementation 'com.android.support:support-v4:25.3.1'
 }
 
 android {
--- a/android-mobile/res/xml/filepaths.xml
+++ b/android-mobile/res/xml/filepaths.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<paths>
+    <files-path path="./" name="logfiles" />
+</paths>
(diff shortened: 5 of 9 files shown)


dirkhh:
That last one has got nothing to do with the log file sharing... instead it tries to fix the scaling of the profile in the mobile apps. 
@bstoeger I desperately need your input on that one... it seems totally bogus... but it also appears to work...

profile-widget/qmlprofile.cpp
```diff
@@ -51,7 +52,12 @@ void QMLProfile::paint(QPainter *painter)
 		timer.start();
 
 	// let's look at the intended size of the content and scale our scene accordingly
+	// the ProfileScene code runs with a dpr of 1 and we shrink the target rectangle
+	// to make things fit again - this seems illogical, but appears to work
 	QRect painterRect = painter->viewport();
+	painterRect.setSize(QSize(lrint((double)painterRect.size().width() / devicePixelRatio()),
```
bstoeger: Yes, that is ugly.

You could try two things:

1) Remove all these changes and do a `painter->setViewTransformEnabled(false)`. That would mean that we do all the DPR handling.

2) Keep the changes, but remove this conversion and replace `painter->viewport()` by `painter->window()`.  This would mean that Qt does the DPR handling.

At least, that would be my interpretation of the documentation. Could be wrong though.

Is there a way to simulate high-DPI on mobile-on-desktop?

dirkhh: Sorta

`QT_DEVICE_PIXEL_RATIO=2.75 ./subsurface-mobile`

That just happens to be the ratio on my PIxel 5. Others show an even `2` or `2.5` or `3.1875`

The initial window size becomes weird, but you can resize and make it work.
Or, better, set it via `SUBSURFACE_MOBILE_WIDTH` and `SUBSURFACE_MOBILE_HEIGHT` to an appropriately scaled value - actually, that seems somewhat broken right now... need to figure out what's wrong there as well. But the `QT_DEVICE_PIXEL_RATIO` works to see the effect

bstoeger: Oh - nice. Good for visually impaired people. I used QT_SCALE_FACTOR=2.75.

Adding a `painter->setViewTransformEnabled(false);` before `QRect painterRect = painter->viewport();` seems to solve the issue for me. It is of course difficult to predict how this then looks on physical device.

dirkhh: You should get yourself one of those fancy "physical devices"... I hear they aren't very expensive... 🤣

dirkhh: Hmm - I just tried that on my device and that get's us right to where we were before my hack.
Did I understand you correctly that I was supposed to revert my last commit and simply add that one line?
That once again gets me just the top left 1/8th of the profile with HUGE fonts...

bstoeger: Never mind - ignore my previous comment. I will try to analyze this tomorrow.

dirkhh: Yeah, I have a few hours starting now, so I'll try some other approaches that don't take away the whole 'super-sampling' advantage of the high DPI screens... because that's why the results look so crappy with my hack...

dirkhh: OK, I think I got it.
The problem is that the painter has a transformation set that scales by the DPR. And that's simply wrong. I'll push my next attempt in a moment, but here's a quick comparison shot taken with mobile on desktop...
Screen Shot 2022-02-27 at 12 57 20 PM
Left is the new code, right is the hack you so rightfully criticized here...

bstoeger: I agree with the fix - that's what I hoped the `setViewTransformEnabled(false)` code would achieve. Obviously I didn't read the docs well enough.

Not sure I agree with the comment - I think that the transform should make the painting code DPR independent, so it is kind of logical (if convoluted). But since we do our own DPR scaling, things explode. That makes me wonder whether it wouldn't be preferred to remove our own DPR scaling.

dirkhh:
adding another completely unrelated commit to change the full version number (otherwise the app stores will reject the update 🤷🏼‍♂️)
That one deals with an oddity in the setup of my multi-qt / multi-target build system where I have different terminals / directories that each are set up to build desktop / mobile / different Qt versions... it's all a bit of a hack... but it makes it much easier to make sure that what I do works in all of our target version

dirkhh:
@bstoeger - any thoughts about https://github.com/subsurface/subsurface/pull/3406/commits/354fb15776a6b1f514a746a6e8d19fbea5353691
This seems to be much better - but people are still complaining that the fonts on the profile are too large

bstoeger:
Yes, agree with the change, as I already commented here:

https://github.com/subsurface/subsurface/pull/3406#discussion_r815493259

profile-widget/qmlprofile.cpp
dirkhh: I somehow missed these comments.
I'm merging this as is - but you are right... it almost looks like we should play with removing the scaling ...

dirkhh:
merged manually

bstoeger:
Concerning the complaint about fontsize, what do you think about this:
```
diff --git a/profile-widget/qmlprofile.cpp b/profile-widget/qmlprofile.cpp
index a47eb0e3a..35c720967 100644
--- a/profile-widget/qmlprofile.cpp
+++ b/profile-widget/qmlprofile.cpp
@@ -31,7 +31,7 @@ QMLProfile::~QMLProfile()
 
 void QMLProfile::createProfileView()
 {
-	m_profileWidget.reset(new ProfileScene(m_devicePixelRatio, false, false));
+	m_profileWidget.reset(new ProfileScene(m_devicePixelRatio * 0.8, false, false));
 }
 
 // we need this so we can connect update() to the scaleChanged() signal - which the connect above cannot do
```
Then at least the scale-factor has some use.
Of course, that scales all the plot, including line-widths, etc.

Another way would be to only change the fontsize on mobile in `divetextitem.cpp`.

bstoeger:
Note that the individual text items can also be given a different size if there is some imbalance.

PR #3410, opened 2022-03-04
[Testing needed] Include profile editing in undo system

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Removes one of those long-standing things: the modal profile editing, which only places an undo command when the explicitly saves the changes.

Replace by an edit mode that records every single change to the profile.

Still a bit rough around the edges.

### Release note:

Done.

### Documentation change:

The changed behavior should obviously be reflected in the documention.


dirkhh:
Just so you don't think I'm ignoring you.
I've looked through the code and I played with it.
I'm not entirely happy with the user experience... especially "discoverability" of how you stop editing and what that actually means... I would have naively assumed that ESC actually cancels the changes that I made... it's like "break off the current action".

I need to play more with this and ponder the options.

bstoeger:
> I'm not entirely happy with the user experience... especially "discoverability"

True, but to be fair, the whole feature is hard to discover (obscure context menu).

So let me try something more radical: *always* go into edit mode when displaying a manually added dive. I don't see the point of the separate mode. With this, the colors should probably be toned down a bit. Or not change at all - just add the handles (or handlers as the source code calls them).

bstoeger:
Note: for some reason with this PR, one can't save planned dives. Will investigate.

dirkhh:
This looks very promising. I don't use the editor enough to am confident that I'm testing all the corner cases, but after a few minutes with it I like it.
I'll play with it some more this morning - but unless I find something unexpected, I think this is ready to merge.

dirkhh:
So here's an odd thing. We used to be able to edit any profile. Which for most profiles that came from dive computers was INSANE... but if you had one broken data point (glitch in the pressure sensor that moved you to 300m for one sample - I had that happen both with an old OSTC and more than once with my Uemis), it was really easy to just fix this in the UI.
And that feature appears to no longer be there...

bstoeger:
> So here's an odd thing. We used to be able to edit any profile.

I think that's a different feature: Log > Edit Dive in Planner.

AFAIK, the "normal" editing was only ever enabled for manually added dives.

dirkhh:
and that, I think, shows conclusively how long it has been that I have taken that old OSTC (long broken) or that Uemis (almost as long broken) on a dive... I clearly haven't used this in ages.

Yes, with that comment in place... I think this is good to go.
I'll deal with the merge conflict locally, I think.
But first I need to take my poor dog on a walk before it starts to rain :)
Back in an hour

dirkhh:
Oh, thanks for rebasing. I'll merge this and #3402 locally anyway to avoid triggering the whole CI/CD thing twice in a row and burn up cpu time...

PR #3412, opened 2022-03-11
core: split proper divecomputer in split_divecomputer()

bstoeger:
split_divecomputer() is passed a dive and a divecomputer number.
However, it accesses the currently visible dc!

This would be a nasty bug if it werent for the fact that it is
called when placing an undo command and there it is passed the
current dive and divecomputer anyway.

Nevertheless, fix this.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a bug in a function, that should not be user visible.
Completely untested, but looks rather logical...?
### Changes made:

1) Access the correct divecomputer when splitting dives.


bstoeger:
Added a second commit that removes an ominous access to the global "dc_number" variable.

bstoeger:
And more commits to move handling of the global `dc_number` variable from core to desktop. It makes no sense on mobile, where multiple profiles are active at the same time.

dirkhh (approved):
the number of these stupid silly misunderstandings that we have everywhere... just bad

thanks for your diligence in fixing them over time.

dirkhh:
This looks very nice. The last commit has a typo in the commit message (plot/plit) - and I kinda agree with the question posed... I think it would be more logical to have a pointer back to the profile and use that to access the plot_info. Any reason you didn't do that in the end?

dirkhh:
and I'll take this one as well - and fix up the commit message

PR #3415, opened 2022-03-13
add list of global objects

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

We traditionally have problems with global (static) QObjects that get destructed after main exits. For some reason you mustn't call destructors of many Qt-objects after the application is gone.

The simplest way to handle this is to just leak them
Somewhat more satisfying is to delete them manually.

This PR creates a registry of global objects that will be deleted before the application object is destroyed.

Perhaps Qt has something to do the same (we surely aren't the only ones with that problem). But since I didn't find anything immediately, I just thought let's do our own.

No, I am not proud about the code. But it seems to work.

### DIFF
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -57,6 +57,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/gas-model.c \
 	core/gaspressures.c \
 	core/git-access.c \
+	core/globals.cpp \
 	core/liquivision.c \
 	core/load-git.c \
 	core/parse-xml.c \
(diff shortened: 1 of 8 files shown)


dirkhh (approved):
I actually quite like this.
Seems logical and clean.
Since it's an RFC I won't merge it.

bstoeger:
> Since it's an RFC I won't merge it.

Removed RFC, since you commented and I don't expect more comments.

Still bit ashamed of the implementation. 😛

dirkhh:
merged

PR #3421, opened 2022-03-17
translations, crash fix, allow entering only a date on mobile

dirkhh:

- [x] Bug fix
- [x] Functional change
- [x] Language translation

This mixes three things, sorry.
1) push out all the strings and pull in first round of translations
2) fix a crash with the undo stack in the global objects that only happened on mobile
3) accept the user entering only a date but not a time when editing a dive on mobile (use case is a user entering old dives on mobile where they only remember the date, but not the dive time... why they'd do that on mobile I'll never know, but whatever)


bstoeger (approved):
The crash fix looks obviously correct to me.

PR #3422, opened 2022-03-20
profile: Plot sensor data from other dive computers

michaelandreen:
If the current dive computer doesn't have a sensor for the cylinder then
check if another dive computer has sensor data available and use that
for the plot.

- [ ] Bug fix
- [x] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -705,9 +705,27 @@ static void calculate_sac(const struct dive *dive, const struct divecomputer *dc
 
 static void populate_secondary_sensor_data(const struct divecomputer *dc, struct plot_info *pi)
 {
-	UNUSED(dc);
-	UNUSED(pi);
+	int *seen = calloc(pi->nr_cylinders, sizeof(*seen));
+	for (int idx = 0; idx < pi->nr; ++idx)
+		for (int c = 0; c < pi->nr_cylinders; ++c)
+			if (get_plot_pressure_data(pi, idx, SENSOR_PR, c))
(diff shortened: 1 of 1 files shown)


core/profile.c
```diff
@@ -705,9 +705,25 @@ static void calculate_sac(const struct dive *dive, const struct divecomputer *dc
...
+		for (int c = 0; c < pi->nr_cylinders; ++c)
+			if (get_plot_pressure_data(pi, idx, SENSOR_PR, c))
+				++seen[c]; // Count instances so we can differentiate a real sensor from just start and end pressure
+	int idx = 0;
 	/* We should try to see if it has interesting pressure data here */
+	for (int i = 0; i < dc->samples && idx < pi->nr; i++) {
+		struct sample *sample = dc->sample + i;
+		for (; idx < pi->nr; ++idx) {
+			if (idx == pi->nr - 1 || pi->entry[idx].sec >= sample->time.seconds)
+				break;
```
dirkhh, line 721: I think this code deserves a comment. I keep staring at this and am trying to figure out what you are doing here... and it doesn't click...

michaelandreen: Done. Stole the code from the add_plot_pressure function just below.

dirkhh: ah. hmm. ok.
Yes, this should work.

PR #3423, opened 2022-03-23
Windows: fix missing DLL in installer

dirkhh:
The branch title is no longer correct. I made the code generic as a similar issue broke the macOS installers as well.

- [x] Bug fix
- [x] Build system change


dirkhh:
Merging this so we get test builds that people could... I don't know... what might people do with test builds... maybe test? Nah, that's silly....

PR #3425, opened 2022-03-26
translations: load parent language translations

dirkhh:
Many language have country specific differences. We recognize different
flavors of English (US, UK (and South Africa)), German (Germany and
Switzerland), and Portuguese (Brazil and Portugal). For many other
flavors of the languages that we have translations for we have no
support and the way we hard-coded the fallbacks in the past was odd and
meant that in the cases where we do have two flavors, missing strings in
one weren't taken from the other (English as the default language being
the exception).

This tries to do a better job of recognizing some of those parent
languages and loading translators for them, first. Which means if we
then find a translator for the specific language (i.e., de_CH), strings
missing in that translation are next searched in the parent language
(de_DE), before finally providing the source language string (en_US).

- [x] Functional change
- [x] Language translation

### Changes made:

Provide a list of parent languages and load those first, if the locale is a different flavor of the same language.

### Related issues:

All the missing strings in the Swiss translations that @charno pointed out

### Additional information:

This should do the right thing in many other circumstances and deserves some testing.

### Release note:

done

### Mentions:

@bstoeger (just for my sanity)

### DIFF
--- a/core/qt-init.cpp
+++ b/core/qt-init.cpp
@@ -11,7 +11,7 @@
 #include "core/settings/qPref.h"
 
 char *settings_suffix = NULL;
-static QTranslator qtTranslator, ssrfTranslator;
+static QTranslator qtTranslator, ssrfTranslator, parentLanguageTranslator;
 
 void init_qt_late()
 {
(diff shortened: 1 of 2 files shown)


core/qt-init.cpp
```diff
@@ -107,10 +97,40 @@ void init_qt_late()
...
 		}
 	}
+	// when creating our language names, we didn't define generic languages with additional
+	// country specific extensions, instead we included the "primary" country in the
+	// translation designations. This breaks the way Qt finds fall-back translations.
+	// Changing this now seems rather tedious, so instead we manually create a few
+	// obvious fallbacks
+	QPair<QLocale::Language, QLocale::Country> parents[] = {
+		{ QLocale::German, QLocale::Germany },
+		{ QLocale::Portuguese, QLocale::Brazil },
```
charno: Shouldn't this be `{ QLocale::Portuguese, QLocale::Portugal },` since the pt_PT translation is complete, and pt_BR is not?

dirkhh: Oops. Yes. Of course. What was I thinking... likely that we initially had a better Brazilian Portuguese translation, but that's long long past...

PR #3426, opened 2022-03-27
core: allow separators ,; after degree-style coord

qlyoung:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Tweak the Lat/Long coordinate parser to allow coordinates of the form:

`12.1049° N, 68.2296° W`

The coordinate parser works by tokenizing coordinates one at a time. Consequently it is invoked twice on user input to get latitude and then longitude. Normally, after parsing the first coordinate, intervening characters such as `,` or `;` and any whitespace are discarded from the input before parsing the second coordinate. But prior to this patch, if the coordinate format was in degrees followed by a sign (`N` is a sign in this example), the parser would skip the bit of code that fast forwards past any intervening separators and whitespace (`, ` in this example). This resulted in coordinates of the given form not being accepted, because the second parse would start with `, 68.2296° W` and reject this as an invalid coordinate.

To rectify this, the bit of code that fast forwards past separators and whitespace has been broken out from the tokenization loop and performed as a final step after a single coordinate has been completely parsed and validated. Doing it this way makes it independent of the state of the tokenizer, so that the fast-forward code will always execute once a coordinate has been successfully parsed.

I've also centralized the list of allowed separators into its own static string; this is necessary as part of the patch but should also make allowing additional separator characters between coordinates trivial in the future, if needed.

### Changes made:
1.  The bit of code that fast forwards past separators and whitespace has been broken out from the tokenization loop and performed as a final step after a single coordinate has been completely parsed and validated.

I saw no relevant test cases that cover this.

### Related issues:
No related issues I saw

### Additional information:
The coordinate format this adds support for happens to be the one that Google shows in quick-answer cards like this:
image

After this patch, those strings can now be pasted directly into Subsurface.

### Release note:
A release note would be nice and make me feel nice. I've updated `CHANGELOG.md` accordingly.

### Documentation change:
None

### Mentions:
@pvalsecc, I see your name on the blame for the coordinate parser.

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -129,6 +129,7 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
 		       const QString &negatives, const QString &others,
 		       double &value)
 {
+	static const QString separators = QString(",;");
 	bool numberDefined = false, degreesDefined = false,
 		minutesDefined = false, secondsDefined = false;
 	double number = 0.0;
@@ -200,11 +201,8 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
 			numberDefined = false;
 			secondsDefined = true;
 		} else if ((numberDefined || minutesDefined || secondsDefined) &&
-			   (txt[pos] == ',' || txt[pos] == ';')) {
-			// next coordinate coming up
-			// eat the ',' and any subsequent white space
-			while (++pos < txt.size() && txt[pos].isSpace())
-				/* nothing */ ;
+			   separators.indexOf(txt[pos]) >= 0) {
+			// separator; this coordinate is finished
 			break;
 		} else {
 			return false;
@@ -219,6 +217,12 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
 		value += number / 3600.0;
 	else if (numberDefined)
 		return false;
+
+	// We parsed a valid coordinate; eat any subsequent separators and/or
+	// whitespace
+	while (pos < txt.size() && (txt[pos].isSpace() || separators.indexOf(txt[pos]) >= 0))
+		pos++;
+
 	if (sign == -1) value *= -1.0;
 	return true;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- core: add support for separator characters when using degree-style coordinates
 - profile: include profile editing in undo system
 - mobile: Add a dark theme for statistics
 - core: avoid crash with corrupted cloud storage


core/qthelper.cpp
```diff
@@ -129,6 +129,7 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
 		       const QString &negatives, const QString &others,
 		       double &value)
 {
+	static const QString SEPARATORS = QString(",;");
```
atdotde: All caps is for macro names. Please use lower case (or camel case in C++ code) variable names.

core/qthelper.cpp
```diff
@@ -219,6 +217,14 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
 		value += number / 3600.0;
 	else if (numberDefined)
 		return false;
+
+	// We parsed a valid coordinate; eat up to one separator and any
+	// subsequent whitespace.
+	if (pos < txt.size() && (txt[pos].isSpace() || SEPARATORS.indexOf(txt[pos]) >= 0)) {
+		while (++pos < txt.size() && txt[pos].isSpace())
```
atdotde: This seems to not cover the case when there is a space before the separator. This might be ugly but I can see the case where people type this (In French for example, you type a space before the comma).

atdotde (changes requested):
Thank you for your contribution! It's always great to see new people coming in with patches. Two minor nitpicks though: One is coding style (for a list of our coding style rules the the file CODINGSTYLE.md) the other is a small  possible generalisation. 

I haven't yet looked at the overall logic, though.

core/qthelper.cpp
qlyoung: I read through CodingStyle.md and used clang-format before I submitted this patch. Normally I would indeed use lowercase/camelcase variables like you suggest, but the variable in this case is a `static const` - vaguely akin to a `#define` in terms of conceptual utility - and I didn't see anything explicit in CodingStyle.md regarding the correct style for static variables. Since in my experience these often have macro-like style (all caps) or other special style in C-like codebases, I looked for the closest adjacent code that uses them and, for consistency, copied the style from that code:

```cpp
bool parseGpsText(const QString &gps_text, double *latitude, double *longitude)
{
        static const QString POS_LAT = QString("+N") + gettextFromC::tr("N");
        static const QString NEG_LAT = QString("-S") + gettextFromC::tr("S");
        static const QString POS_LON = QString("+E") + gettextFromC::tr("E");
        static const QString NEG_LON = QString("-W") + gettextFromC::tr("W");
```
However, I also see that this code is in fact the only place in all of subsurface where this pattern is followed for `QString`:
```
qlyoung@corax ~/p/s/core (master)> rg "static const QString [A-Z]+" | wc -l
4
qlyoung@corax ~/p/s/core (master)> rg "static const QString [A-z]+" | wc -l
32
```

With this in mind:
1. if I've missed the part of CodingStyle.md that clarifies this, can you point it out so I can avoid this mistake in the future? If it's not there, I'm happy to submit a separate patch updating that file.
2. Should I fix the existing code as well in a separate patch?

Also, the word "macro" doesn't appear in CodingStyle.md - would you like me to submit a patch indicating macros should be capitalized as well?

core/qthelper.cpp
qlyoung: I considered writing this so that it would skip any number of spaces or separators - that would be the most general case, but I wasn't sure if this would be considered *too* permissive; i.e. it would allow things like any of the following:
```
12.1049° N,, 68.2296° W
12.1049° N ,, 68.2296° W
12.1049° N,,,, ,, ,,,;; 68.2296° W
```
Personally I'm leaning towards accepting this - I don't see any real drawbacks. I would prefer to do that, which would be a simple code change, rather than special case by keeping additional state regarding how many separators we've seen. Thoughts?

atdotde: I would tend to parse as much as possible. The only risk I see is that a user might unintentionally parse nonsense and not be warned by an error message about it. But I think it's better to be permissive.

qlyoung: Perfect, I'll make this change.

dirkhh:
Hi @qlyoung - thanks for sending this PR

The Codingstyle document is indeed not perfect and doesn't answer all questions.
You are correct that we aren't consistent here, but that the majority of those constants aren't all caps; looking at all types we get:
```
$ git grep "static const [A-z_]* [A-Z_]* " | wc -l
      40 
$ git grep "static const [A-z_]* [A-z_]* " | wc -l
     157 
```

And it turns out that the 40 all caps ones are only from two groups of code, both of which were indeed previously `#define` statements that got converted to typed variables later - but kept their all caps names.

Anyway, I think I'd like this better as `static const QString separators` - but clearly not enough so to have enforced this "rule" in the past :)

As for the second question - yes, more liberal / generous parsing seems to be better. We want to make it easy for people to copy values.

qlyoung:
Ok; I've amended to address both comments. I also opened #3427 to note the convention for macro names.

core/qthelper.cpp
qlyoung: I've changed to lowercase and submitted a separate PR to update CODINGSTYLE.md.

core/qthelper.cpp
```diff
@@ -219,6 +217,12 @@ static bool parseCoord(const QString &txt, int &pos, const QString &positives,
 		value += number / 3600.0;
 	else if (numberDefined)
 		return false;
+
+	// We parsed a valid coordinate; eat up to one separator and any
```
dirkhh: I know I am nit-picking (and I can easily fix this when merging), but that comment is no longer correct...

qlyoung: Ech, I fixed this and then un-fixed it by mistake

dirkhh (approved):
yep, this looks good

dirkhh:
merged locally in order to trigger the test builds only once

PR #3427, opened 2022-03-27
CODINGSTYLE.md: note that macros should be caps'd

qlyoung:
Skipping the PR template as the change is self-explanatory

## Reviewers
poke @atdotde, @dirkhh


dirkhh:
we actually have 25 exceptions for this 'rule' (out of 378 non-trivial macros). But I think this is an "intended" coding style, so this makes sense.

dirkhh:
merged locally in order to trigger the test builds only once

Issue #3304, opened 2021-08-13
Can't upload OSTC parameters


dadefay:
### Describe the issue:

- [x ] Bug
- [x ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

The "language" parameter is no more managed by hwOS (3.1.x and later), each language has its own hex file. When trying to upload parameters to the OSTC, language is not recognized, the computer exits Bluetooth mode and Subsurface locks itself : so it is impossible to change any OSTC parameter using Subsurface. By the way, computer model should be memorized by Subsurface, not only the manufacturer (I have to reselect "OSTC 2/3/Sport/cR" each time).
[subsurface.log](https://github.com/subsurface/subsurface/files/6981809/subsurface.log)

### Operating system:

Windows 10.

OSTC 2 hwOS 3.1.5.

### Subsurface version:

5.0.2.

Yes.

Official.

### Steps to reproduce:

1) Put the OSTC in Bluetooth Mode.
2) Select "Change settings on dive computer" in Subsurface.
3) Select "OSTC 2/3/Sport/cR" in the left panel.
4) Connect the OSTC (COM11 for my computer).
5) Once connected click on "Retrieve available details" : should be successful.
6) Don't change any parameter.
7) Click on "Save changes to device" : Subsurface displays "Error: Failed!", the OSTC displays "Unknown Parameter" and sometimes "Rx Data Timeout" then exits Bluetooth mode.

dadefay:
Same behaviour with Subsurface 5.0.3.

glance-:
I just tested this with a OSTC 3 running 3.16, Subsurface 5.0.5-dev-latest, and It works as expected.

dadefay:
Same test with Subsurface 5.0.7 and OSTC 2 running hwOS 3.17, the same fault as described on August 13, 2021 occurs. Idem with  OSTC Plus.

I wonder how **glance** made it working...

glance-:
Can you connect to the dive computer, read the settings, save a backup of your settings to a xml file and then try to write them back to the computer, and post the xml backup and the logfile?

I think the actual issue doesn't have anything to do with your initial guess about language, but rather about that you're running the computer in CCR-mode: Auto SP which didn't exist when this code was written, thus things becomes pear shaped.

dadefay:
Thanks Anton. I did exactly what you asked, backup and log files attached (OSTC configured for CCR with auto setpoint).

Let me know if you need files with different configuration.

[...]

PR #3428, opened 2022-03-28
configure-dc: Add support for ostc3 Auto SP mode

glance-:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Way back in time, in HwOS 1.86 a CCR mode was added which automatically
switched between setpoints based on depth.

This entry was never added in our system to configure the dc, and caused
the issue seen in #3304

This adds the Auto SP mode, to the dropdown, thus fixing #3304.

### Mentions:

@dadefay

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.ui
+++ b/desktop-widgets/configuredivecomputerdialog.ui
@@ -2413,6 +2413,11 @@
                 <string>Sensor</string>
                </property>
               </item>
+              <item>
+               <property name="text">
+                <string>Auto SP</string>
+               </property>
+              </item>
              </widget>
             </item>
             <item row="6" column="2">


dirkhh (commented):
I can't test this - but I'm a bit skeptical that adding the drop down to the `.ui` file is all it takes. I would have assumed that there is some corresponding change in the `.cpp` file to go along with this...

glance-:
Nope. The ID mapping happens from the index in that drop down. ui.ccrModeComboBox->currentIndex() -> deviceDetails->ccrMode -> WRITE_SETTING(OSTC3_CCR_MODE, ccrMode); , and the reverse for read.

dirkhh:
Ha.
Cool. And this isn't stored in the XML backup or anything?

glance-:
It's already stored there, for historical reasons, as a int.

PR #3429, opened 2022-03-31
Allow more than one media file when importing from the web

atdotde:
To import media files from the web, increas the size of the
dialog box and allow several URLs separated by newlines.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -905,46 +905,53 @@ void DiveListView::loadWebImages()
 	URLDialog urlDialog(this);
 	if (!urlDialog.exec())
 		return;
-	loadImageFromURL(QUrl::fromUserInput(urlDialog.url()));
+	loadImagesFromURLs(urlDialog.url());
 }
 
-void DiveListView::loadImageFromURL(QUrl url)
+void DiveListView::loadImagesFromURLs(QString urls)
 {
-	if (url.isValid()) {
-		QEventLoop loop;
-		QNetworkRequest request(url);
-		QNetworkReply *reply = manager.get(request);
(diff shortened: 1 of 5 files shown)


desktop-widgets/divelistview.cpp
```diff
@@ -905,46 +905,53 @@ void DiveListView::loadWebImages()
...
-			report_error(qPrintable(tr("%1 does not appear to be an image").arg(url.toString())));
-			return;
-		}
+	loadImagesFromURLs(urlDialog.url());
+}
+
+void DiveListView::loadImagesFromURLs(QString urls)
+{
+	QStringList validUrls;
+	QStringList urlList = urls.split('\n');
```
dirkhh, line 914: any reason why you wouldn't want to allow `,` and `;` as separators as well?

dirkhh (commented):
Seems like a nice feature.
I have a question - not a change request. Just curious.

desktop-widgets/divelistview.cpp
atdotde: Laziness, honestly. I went for the cheaper version of split which operates with a char rather than a regular expression. But I have to say: I actually first did this with comma but by accident, the two URLs I tested this with (two random images from the browser) already contained commas in the query string part of the URL. That in retrospect justified for me not splitting on comma (or semicolon for that matter) as well.

Of course I can change this to /[,;\n]/ but do you think it is worth that some URLs that contain these characters then break?

dirkhh:
I hadn't thought there could be a comma in a URL.
Yeah, not worth it

PR #3430, opened 2022-04-01
Add template variables for pictures

atdotde:
This introduces a new list the printing templates can loop
over: pictures contains the urls (local or remote) of media
associtated to the current dive. This allows constructs like

```
{% for pic in pictures %}
	<img src="{{ pic.url }}" width=20%>
{% endfor %}

```
in the printing template.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/desktop-widgets/templatelayout.cpp
+++ b/desktop-widgets/templatelayout.cpp
@@ -319,6 +319,16 @@ static std::vector<const cylinder_t *> cylinderList(const dive *d)
 	return res;
 }
 
+static QStringList pictureList(const dive *d)
+{
+	QStringList res;
+	res.reserve(d->pictures.nr);
+	FOR_EACH_PICTURE(d) {
+		res << QString(picture->filename);
+	}
+	return res;
+ }
+
 void TemplateLayout::parser(QList<token> tokenList, int from, int to, QTextStream &out, State &state)
 {
 	for (int pos = from; pos < to; ++pos) {
@@ -359,6 +369,11 @@ void TemplateLayout::parser(QList<token> tokenList, int from, int to, QTextStrea
 						parser_for(tokenList, pos, loop_end, capture, state, cylinderList(*state.currentDive), state.currentCylinderObject, false);
 					else
 						qWarning("cylinderObjects loop outside of dive");
+				} else if (listname == "pictures") {
+					if (state.currentDive)
+						parser_for(tokenList, pos, loop_end, capture, state, pictureList(*state.currentDive), state.currentPicture, false);
+					else
+						qWarning("pictures loop outside of dive");
 				} else {
 					qWarning("unknown loop: %s", qPrintable(listname));
 				}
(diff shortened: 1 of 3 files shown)


desktop-widgets/templatelayout.cpp
```diff
@@ -319,6 +319,16 @@ static std::vector<const cylinder_t *> cylinderList(const dive *d)
 	return res;
 }
 
+static std::vector<const QString> pictureList(const dive *d)
+{
+	std::vector<const QString> res;
+	res.reserve(d->pictures.nr); //FIXME
```
dirkhh: always love seeing FIXME comments... :(

dirkhh (commented):
the builds are failing with fascinating C++ error messages...

desktop-widgets/templatelayout.cpp
atdotde: According to April 1st, "FIXME" is there as a reminder to remove the "FIXME" comment. (The actual thing to fix I had already fixed but of course forgot the comment).

atdotde:
Yeah, these error messages were fascinating, WTF. Again, I learned I should stay away from attempting to use C++ standard library data types, they are beyond my pay scale (fixed the problem by replacing a std::vector<whatever> by a QStringList). 

Honestly, what is the point of abstraction when you cannot use it as a black box but have to worry about the intestines?

atdotde:
In other news, I found this to work only intermittently. Could others please test? There seems to be a race condition with downloading the actual images. This might be related to what stopped us in https://github.com/subsurface/subsurface/pull/3354

atdotde:
Here is the template I use for testing
[imagetemplate.txt](https://github.com/subsurface/subsurface/files/8402153/imagetemplate.txt).

dirkhh:
There is this very weird thing going on in our printing code where we apparently cannot wait for everything to fully "render" before things get printed.

In a way I'd love to completely rethink the way we do this. Get rid of WebView and WebKit and those gigantic libraries for something that really shouldn't be that hard. There's got to be a simpler way to lay out the template and create something that we can then send to the printer routines. These don't need to be web pages.

But of course that's a major project that will take time and patience and that's the thing that I am clearly lacking - as is most everyone else.

Still - what we do today is a dead end - and the "obvious" switch to WebView doesn't work on Windows, so it's also not the right path. 

There has got to be a better way to do this.

@thiagomacieira  how do other Qt apps print? Do they all use WebKit / WebView? Or is there another preferred mechanism?

dirkhh:
and after my, I don't know, fifth (?) trip down this rat hole... I still have no better idea how to do this.
Yes, you can totally use widgets to assemble a page and then print this. But you need to write the code to make this flow and page-inate nicely. And that feels like such a pain in the rear. And wherever you look people just say "oh just use a WebView", ignoring the tens of megabyte of crud and platform issues that this brings.
The fact that I cannot seem to find a simpler, easy to integrate solution for this problem (which I have got to think is fairly common) is just sad...

dirkhh:
To go back to your pull request - yes, I can make it fail fairly reliably. There must be a way to ensure that things have finished rendering before we print. We MUST be doing something wrong here.
Gah.

thiagomacieira:
> @thiagomacieira how do other Qt apps print? Do they all use WebKit / WebView? Or is there another preferred mechanism?

Poorly. QtPrintSupport has seen very little support in the last decade or so. It's been an issue for a lot of applications -- I remember KMail having the same problem.

atdotde:
Maybe we can talk to a car manufacturer to add a printer to the car. That might add encouragement to give art printing done TLC.

atdotde:
Next thing I will try is to ditch QWebWhatever and try if the limited html of QTextEdit might s suffice.

PR #3431, opened 2022-04-02
Clear undo stack on file close

michaelandreen:
The undo stack is only relevant to the dives that were loaded at the
time the command was executed. If a file is closed, by specifically
closing it or opening another file, then the undo commands will
reference dives that aren't available anymore. Clearing the undo stack
ensures that we don't crash or accidentally do some undefined
modifications to the currently open file.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 - Desktop: Allow more than one media file to be imported from web
+- undo: Clear undo stack when the current file is closed
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
 * Use this layout for new entries: `[Area]: [Details about the change] [reference thread / issue]`
--- a/commands/command_base.cpp
+++ b/commands/command_base.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0
 
+#include "command.h"
 #include "command_base.h"
 #include "core/globals.h"
 #include "core/qthelper.h" // for updateWindowTitle()
@@ -18,6 +19,7 @@ void init()
 {
 	undoStack = make_global<QUndoStack>();
 	QObject::connect(undoStack, &QUndoStack::cleanChanged, &updateWindowTitle);
+	QObject::connect(&diveListNotifier, &DiveListNotifier::dataReset, &Command::clear);
 	changesCallback = &changesMade;
 }
 
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -513,7 +513,6 @@ void MainWindow::closeCurrentFile()
 	if (!existing_filename)
 		setTitle();
 	disableShortcuts();
-	Command::setClean();
 }
 
 void MainWindow::updateCloudOnlineStatus()


dirkhh (approved):
That seems very reasonable.

@bstoeger - any thoughts?

bstoeger:
Yes, this is reasonable. The `setClean()` call should probably be removed as an empty undo-list should be clean by definition.

Note that the `clear()` call was removed in fb6210a99a1eda0f7f7197ed57f3c61a82132dff, so mobile is probably also affected.

Which makes me wonder whether this should be moved to `clear_dive_file_data`. On the other hand, calling C++ from C is so painful...

dirkhh:
> Yes, this is reasonable. The `setClean()` call should probably be removed as an empty undo-list should be clean by definition.
> 
> Note that the `clear()` call was removed in [fb6210a](https://github.com/subsurface/subsurface/commit/fb6210a99a1eda0f7f7197ed57f3c61a82132dff), so mobile is probably also affected.
> 
> Which makes me wonder whether this should be moved to `clear_dive_file_data`. On the other hand, calling C++ from C is so painful...

So we have several `setClean()` calls in that file. And I'm not entire sure what you are saying here...
That `setClean()` in general makes no sense and should be removed?
Or that it should be removed here (which it is in the patch)?

I'd just hate for me to try to follow the logic (or what I think I read in your comment) and then make things worse...

michaelandreen:
It would be fairly easy to add an extern "C" wrapper in qthelper.h/cpp or similar. There are currently no dependencies from core to commands and adding something like this would introduce circular dependencies between the libs. I don't know if that is something we care about.

michaelandreen:
> So we have several `setClean()` calls in that file. And I'm not entire sure what you are saying here...
> That `setClean()` in general makes no sense and should be removed?
> Or that it should be removed here (which it is in the patch)?

setClean() seems to be a way to say that the current state has been saved. The cases where it's used are save, save_as and save to cloud, so that seems reasonable.

bstoeger:
I'm saying that `setClean` and `clear` is redundant, so the new version should be correct.

> It would be fairly easy to add an extern "C" wrapper in qthelper.h/cpp or similar. There are currently no dependencies from core to commands and adding something like this would introduce circular dependencies between the libs. I don't know if that is something we care about.

There is precedence. For example, core calls into the C++ fulltext code.

I have no preference in this particular case. There are three calls to `clear_dive_file_data` in `qmlmanager.cpp`

> setClean() seems to be a way to say that the current state has been saved. The cases where it's used are save, save_as and save to cloud, so that seems reasonable.

Correct. This marks the "clean" state for which the asterisk next to the filename is removed.

dirkhh:
I posted an alternative PR that simply (Ha!) calls `Command::clear()` from the C routine `clear_dive_file_data()`

What it is missing is the `CHANGELOG` entry as well as the removal of `Command::setClean()` calls after we call `clear_dive_file_data()` -- I was simply trying to be quick to catch @bstoeger before he signs off for today :)

michaelandreen:
Since `clear_dive_file_data` emits `diveListNotifier.dataReset` we can just connect that to clear. Seems cleaner and avoids the circular dependency.

bstoeger:
LGTM. Ultimately, the `clean` function could be made local. I'm also fine with Dirk's version. Both have things going for and against them.

bstoeger (approved):

dirkhh:
I liked the logical clarity of my solution. This way you still have to remember to do this. Remind me, is this problem fixed now for mobile as well, or does it still need to be fixed there?

Anyway, I'll merge this.

michaelandreen:
> Remind me, is this problem fixed now for mobile as well, or does it still need to be fixed there?

I would assume that it is fixed for mobile as well, since it calls Command::init().

PR #3432, opened 2022-04-02
core: clearing dive data should clear undo stack

dirkhh:
If we clear out our dive data, we also need to clear out the undo stack
because it otherwise will refer to dives that no longer exist.

The cleanest way to do that would be to do it in the same function used
to clear the dive data. Which causes us to call C++ code from C code and
really is a bit of a mess with a circular dependency between our
libraries.

But this does seem better than relying on people to remember to call
into a second function after clearing the data.

Suggested-by: Michael Andreen <michael@andreen.dev>

This really is just an alternative to #3431 

@michaelandreen 
@bstoeger

### DIFF
--- a/commands/command_base.cpp
+++ b/commands/command_base.cpp
@@ -23,7 +23,10 @@ void init()
 
 void clear()
 {
-	undoStack->clear();
+	// this can get called from C code even if the Command code was never initialized
+	// (really only in the case of tests like TestParse)
+	if (undoStack)
+		undoStack->clear();
 }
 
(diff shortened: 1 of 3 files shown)


michaelandreen:
Seems less error prone. Just annoying with the circular dependencies. I was messing around with a similar thing, but got stuck trying to fix proper dependencies between the libraries, but some custom target (probably automoc or similar) made CMake complain that not all dependencies were static.

dirkhh:
we can close this one now

PR #3433, opened 2022-04-03
profile: remove last dc_number access from profile code

bstoeger:
dc_number is a global variable indicating the currently displayed
dive on desktop. It makes no sense on mobile, since multiple
profiles can be active at the same time. Therefore, the profile
code should not access this global, but use the dc number that
is passed in.

This removes the last access.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The commit message says it all: The profile shouldn't use global variables, since on mobile multiple profiles are active at the same time.

### Changes made:

1) Remove access to `dc_number`

### Related issues:

Nope. No user-visible change.


dirkhh (approved):
that seems straight forward...

PR #3436, opened 2022-04-11
More apple m1 qt6

dirkhh:

- [x] Code cleanup
- [x] Build system change

Continuing the slow work to get to fat binaries on Mac and as a requirement to that (but not limited to it), Qt6 support

### Additional information:

this is the basis of the latest DMG that I made today

### Release note:

missing

### Mentions:

@atdotde @bstoeger - I'm always grateful for an extra set of eyes... and Robert, you of course should be able to actually test the DMG I posted...


dirkhh:
@atdotde @bstoeger 
did either of you have a chance to take a look?
The binaries build based on that seem to work, but only the fat binary with Qt6 has received actual testing, and there the verdict was essentially "fewer things are broken than before", which isn't a ringing endorsement... but I don't think this should break anything for people on Qt 5 or on x86 Macs.
Famous last words, right?

core/uploadDiveShare.cpp
```diff
@@ -45,13 +45,19 @@ void uploadDiveShare::doUpload(bool selected, const QString &uid, bool noPublic)
 	reply = manager()->put(request, json_data);
 
 	// connect signals from upload process
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+	connect(reply, &QNetworkReply::finished, this, &uploadDiveShare::uploadFinishedSlot);
+	connect(reply, &QNetworkReply::errorOccurred, this, &uploadDiveShare::uploadErrorSlot);
```
bstoeger, line 50: Small rant: Jeez. I find that function renaming so tiresome. There are so many API problems that could/should be improved, why waste time on stupid function names?

bstoeger (approved):
The only code change looks reasonable to me.

Of course, one could simply merge the WebServices and DivelogsDeWebServices classes to avoid all that virtual function business. But then, maybe someone wants to add a different webservice later.

That said, couldn't downloadError et al. simply be moved up in the hierarchy, since there are no divelog-de specific things in them?

dirkhh:
> The only code change looks reasonable to me.
> 
> Of course, one could simply merge the WebServices and DivelogsDeWebServices classes to avoid all that virtual function business. But then, maybe someone wants to add a different webservice later.
> 
> That said, couldn't downloadError et al. simply be moved up in the hierarchy, since there are no divelog-de specific things in them?

I think that's totally reasonable. In this exercise I was focused on "get stuff to work", not so much on "make things better / more sane".

core/uploadDiveShare.cpp
dirkhh: yes... my jaw dropped with how many pointless renaming exercises they went through.
Sure the new names are "slightly more logical and consistent", but what the actual FART. Such a waste of time for the users.

dirkhh:
No further comments - the only way this will get tested is once merged... So let's do that and see what happens :)

dirkhh:
merged locally because reasons.

PR #3439, opened 2022-04-13
libdc integration: correctly parse DC_FIELD_SALINITY response

dirkhh:
libdivecomputer tries to be super careful in what it tells us. It only offers a
density value if that is something that the dive computer explicitly supports,
otherwise it just offers back a flag. We need to then update the density value
ourselves.

- [x] Bug fix
- [x] Functional change

This should fix the issue brought up @ martinmcwhorter

### Changes made:

if libdivecomputer sets the density to zero, calculate our own estimate based on the water type

### Release note:

done

### Documentation change:

not needed

### Mentions:

@torvalds


dirkhh:
those build failures on Ubuntu are just strange and pointless. I'll merge this so that @martinmcwhorter can test

PR #3440, opened 2022-04-13
build-system: work around latest GitHub Action issue

dirkhh:
This suddenly started. A couple of build would fail because the git submodule
checkout fails because of directory ownership issues. Hopefully this will fix
it.

- [x] Build system change

description says it all...


dirkhh:
if at first you don't succeed, try, try again...
UGH.

PR #3443, opened 2022-04-16
Collect and convert git repo data to 'struct git_info'

torvalds:
We have this nasty habit of randomly passing down all the different
things that we use to look up the local and remote git repository, and
the information associated with it.

Start collecting the data into a 'struct git_info' instead, so that it
is easier to manage, and easier and more logical to just look up
different parts of the puzzle.

This is a fairly mechanical conversion, but has moved all the basic
information collection to the 'is_git_repository()' function.  That
function no longer actually opens the repository (so the 'dry_run'
argument is gone, and instead a successful 'is_git_repository()' is
followed by 'opn_git_repository()' if you actually want the old
non-dry_run semantics.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/file.c
+++ b/core/file.c
@@ -273,55 +273,45 @@ static int parse_file_buffer(const char *filename, struct memblock *mem, struct 
 	return parse_xml_buffer(filename, mem->buffer, mem->size, table, trips, sites, devices, filter_presets, NULL);
 }
 
-int check_git_sha(const char *filename, struct git_repository **git_p, const char **branch_p)
+bool check_git_sha(const char *filename, struct git_info *info)
 {
-	struct git_repository *git;
-	const char *branch = NULL;
-
 	char *current_sha = copy_string(saved_git_id);
-	git = is_git_repository(filename, &branch, NULL, false);
-	if (git_p)
-		*git_p = git;
-	if (branch_p)
-		*branch_p = branch;
-	if (strstr(filename, prefs.cloud_base_url) && git == dummy_git_repository) {
-		/* opening the cloud storage repository failed for some reason,
-		 * so we don't know if there is additional data in the remote */
-		free(current_sha);
-		return 1;
-	}
-	/* if this is a git repository, do we already have this exact state loaded ?
-	 * get the SHA and compare with what we currently have */
-	if (git && git != dummy_git_repository) {
-		const char *sha = get_sha(git, branch);
-		if (!empty_string(sha) &&
-		    same_string(sha, current_sha)) {
+
+	if (is_git_repository(filename, info) && open_git_repository(info)) {
+		const char *sha = get_sha(info->repo, info->branch);
+		if (!empty_string(sha) && same_string(sha, current_sha)) {
 			fprintf(stderr, "already have loaded SHA %s - don't load again\n", sha);
 			free(current_sha);
-			return 0;
+			return false;
 		}
 	}
+
+	// Either the repository couldn't be opened, or the SHA couldn't
+	// be found.
 	free(current_sha);
-	return 1;
+	return true;
 }
 
 int parse_file(const char *filename, struct dive_table *table, struct trip_table *trips, struct dive_site_table *sites,
 	       struct device_table *devices, struct filter_preset_table *filter_presets)
 {
-	struct git_repository *git;
-	const char *branch = NULL;
+	struct git_info info;
 	struct memblock mem;
 	char *fmt;
 	int ret;
 
-	git = is_git_repository(filename, &branch, NULL, false);
-	if (strstr(filename, prefs.cloud_base_url) && git == dummy_git_repository)
-		/* opening the cloud storage repository failed for some reason
-		 * give up here and don't send errors about git repositories */
-		return -1;
+	if (is_git_repository(filename, &info)) {
+		if (!open_git_repository(&info)) {
+			/*
+			 * Opening the cloud storage repository failed for some reason
+			 * give up here and don't send errors about git repositories
+			 */
+			if (info.is_subsurface_cloud)
+				return -1;
+		}
 
-	if (git)
-		return git_load_dives(git, branch, table, trips, sites, devices, filter_presets);
+		return git_load_dives(&info, table, trips, sites, devices, filter_presets);
+	}
 
 	if ((ret = readfile(filename, &mem)) < 0) {
 		/* we don't want to display an error if this was the default file  */
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -28,8 +28,6 @@
 #include "gettext.h"
 #include "sha1.h"
 
-bool is_subsurface_cloud = false;
-
 // the mobile app assumes that it shouldn't talk to the cloud
 // the desktop app assumes that it should
 #if defined(SUBSURFACE_MOBILE)
@@ -168,12 +166,12 @@ static char *move_local_cache(struct git_info *info)
 
 static int check_clean(const char *path, unsigned int status, void *payload)
 {
-	UNUSED(payload);
+	struct git_info *info = (struct git_info *)payload;
 	status &= ~GIT_STATUS_CURRENT | GIT_STATUS_IGNORED;
 	if (!status)
 		return 0;
 	SSRF_INFO("git storage: local cache dir %s modified, git status 0x%04x", path, status);
-	if (is_subsurface_cloud)
+	if (info->is_subsurface_cloud)
 		report_error(translate("gettextFromC", "Local cache directory %s corrupted - can't sync with Subsurface cloud storage"), path);
 	else
 		report_error("WARNING: Git cache directory modified (path %s) status 0x%04x", path, status);
@@ -183,7 +181,7 @@ static int check_clean(const char *path, unsigned int status, void *payload)
 /*
  * The remote is strictly newer than the local branch.
  */
-static int reset_to_remote(git_repository *repo, git_reference *local, const git_oid *new_id)
+static int reset_to_remote(struct git_info *info, git_reference *local, const git_oid *new_id)
 {
 	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
 	opts.progress_cb = &progress_cb;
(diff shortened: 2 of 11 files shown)


torvalds:
Ok, I think this may be in a usable shape right now.

NOTE! The whole reason for this change (apart from just trying to make it cleaner) is in that big FIXME comment in core/save-git.c.

Basically, it used to be that "is_git_repository()" used to do all the repo open code and update to the upstream (except if the "dry_run" argument was true).  That was very wrong for saving, since you want to save locally first, before you start messing with upstream. 

This does *not* fix it, but now the infrastructure is in place for fixing it. In particular, "is_git_repository()" not just gathers all the information, and if that returns true, you are supposed to then do "open_git_repository()" to do that "sync with upstream".

The git_save_dives() function still does that, so you still have the old broken semantics here, but now - since is_git_repository() has prepared everything - the save code *could* instead just do

        error = git_repository_open(&info->repo, info->localdir);

and just do a local open instead of that "open_git_repository()" dance. But see the comment about how maybe we want to do at least some of that dance.

So basically this whole huge commit is supposed to be just prep-work with no semantic changes. Considering the size of it, that may be optimistic. I couldn't sanely make it multiple smaller and more targeted changes, because it was all about the calling conventions changing for all those git repo helper routines.

core/save-git.c
```diff
@@ -1328,43 +1331,66 @@ int do_git_save(git_repository *repo, const char *branch, const char *remote, bo
...
+	 *
+	 * NOTE! The "we already have a repo" case
+	 * case is fairly easy (we can just treat
+	 * it as a local repository and open it
+	 * as-is). But the "need to create it by
+	 * cloning" case is what we currently also
+	 * do in 'open_git_repository()', and that
+	 * all needs to be thought about a lot.
+	 *
+	 * Do we want to have the rule that you
```
torvalds, line 1382: This is the meat of the question. 

In many ways, it really would make sense to only save to a git repo that we opened earlier - if for no other reason than the fact that we need to have the parent commit. So the whole "create initial empty repo" is possibly only an issue at load time, and save-time could assume that a simple "git_repository_open(&info->repo, info->localdir)" should be sufficient.

Even if that's the case, we'd still want to try to push out those changes after writing, so it's not *just* about changing a "open_git_repository()" to "git_repository_open()".

dirkhh: I think the best approach to this is to think about the different scenarios that we need to deal with.
And that's surprisingly complicated for something that seems so simple...

(1) this system has previously connected to the cloud and has a local repo that's in sync with the remote
(2) this system has previously connected to the cloud and has a local repo that is ahead of the remote
(3) this system has previously connected to the cloud and has a local repo that's behind the remote
(4) this system has previously connected to the cloud and has a local repo that has diverged from the remote

So in cases 1-4 the local and remote repos have a common ancestor, and neither (1), the local repo (2), the remote repo (3) or both (4) have commits since then

(5) this system has previously been in no-cloud mode and has a local repo for that, but no common ancestor with the remote
(6) this system is trying to connect to the cloud for the first time, the remote repo is brand new (i.e., doesn't have the branch)
(7) this system is trying to connect to the cloud for the first time, the remote repo exists and has commits in it
(8) various weird error cases where we have issues with the local or remote repo

Additionally, there are a few different situations that we can be in locally
(A) first time use with these credentials - unknown if they are correct
(B) this system has known good credentials and wants to get new content from the cloud (i.e. the user started Subsurface with auto-sync or started Subsurface and did a manual sync)
(B) this system has changes that need to be stored to the local repo
(C) this system has changes in the local report that need to be pushed to the remote repo

And then you can build a little matrix with the flow that is needed. And in each case you need to account for the possibility that you can't connect to the cloud (what's the end state?) or that there is an issue merging or that there is some other error...

Are we having fun, yet?

Fundamentally this complexity is the reason for the utter nightmare of code that we have here. A few of these cases got added later - and I know that we are not doing the right thing for some of them today.
But I think the more we try to create a clear diagram of the different states and transitions and possible outcomes, the easier it will be to make sure that we get it right.

Oh, and I'm not even 100% sure that the list above is complete :)

core/file.c
```diff
@@ -273,55 +273,45 @@ static int parse_file_buffer(const char *filename, struct memblock *mem, struct
 	return parse_xml_buffer(filename, mem->buffer, mem->size, table, trips, sites, devices, filter_presets, NULL);
 }
 
-int check_git_sha(const char *filename, struct git_repository **git_p, const char **branch_p)
+bool check_git_sha(const char *filename, struct git_info *info)
```
dirkhh, line 276: since we are making this all better... should we change the name of this function?
`check_git_sha()` returns `true` if we can't access the remote of the HEAD of the branch in the remote has a different sha than our global variable `saved_git_id`.
That semantic seems to match better to a name `is_sha_head_of_remote_branch(const char *current_sha, const char *filename, git_info *info)` (which has the inverted return value, of course)... which also avoids the access to the global variable (at least in this function)

dirkhh: and even that name hides the fact that this function clears / overwrites the `git_info *info`... but maybe that is implied...

core/git-access.c
```diff
@@ -130,7 +130,7 @@ static int push_transfer_progress_cb(unsigned int current, unsigned int total, s
 	return git_storage_update_progress(buf);
 }
 
-char *get_local_dir(const char *remote_in, const char *branch)
+char *get_local_dir(const char *url, const char *branch)
```
dirkhh, line 131: all callers already have a `git_info` object. Let's just pass that down to this function.

core/git-access.c
```diff
@@ -160,9 +160,9 @@ char *get_local_dir(const char *remote_in, const char *branch)
 			hash[4], hash[5], hash[6], hash[7]);
 }
 
-static char *move_local_cache(const char *remote, const char *branch)
+static char *move_local_cache(struct git_info *info)
 {
-	char *old_path = get_local_dir(remote, branch);
+	char *old_path = get_local_dir(info->url, info->branch);
```
dirkhh, line 163: with the suggested change above, that just becomes `get_local_dir(info);`

core/git-access.c
```diff
@@ -320,9 +320,8 @@ int certificate_check_cb(git_cert *cert, int valid, const char *host, void *payl
 	return valid;
 }
 
-static int update_remote(git_repository *repo, git_remote *origin, git_reference *local, git_reference *remote, enum remote_transport rt)
+static int update_remote(struct git_info *info, git_remote *origin, git_reference *local, git_reference *remote)
 {
-	UNUSED(repo);
 	UNUSED(remote);
```
dirkhh, line 323: if that's unused, why pass it in?

core/git-access.c
```diff
@@ -743,63 +741,62 @@ int sync_with_remote(git_repository *repo, const char *remote, const char *branc
...
 	git_remote_free(origin);
 	git_storage_update_progress(translate("gettextFromC", "Done syncing with cloud storage"));
 	return error;
 }
 
-static git_repository *update_local_repo(const char *localdir, const char *remote, const char *branch, enum remote_transport rt)
+static bool update_local_repo(struct git_info *info)
 {
 	int error;
-	git_repository *repo = NULL;
```
dirkhh, line 763: I haven't walked back to the callers to understand this better... but here we used to initialize the `repo` as `NULL, but we don't initialize `info->repo` as `NULL`... is that intentional?

core/git-access.c
```diff
@@ -831,102 +828,102 @@ static int repository_create_cb(git_repository **out, const char *path, int bare
...
+		return false;
+
+	return true;
 }
 
-static git_repository *create_local_repo(const char *localdir, const char *remote, const char *branch, enum remote_transport rt)
+static bool create_local_repo(struct git_info *info)
 {
 	int error;
-	git_repository *cloned_repo = NULL;
```
dirkhh, line 876: again dropping a `NULL` initialization. Was this unnecessary in the old code?

dirkhh (commented):
ran out of time on this flight. need to continue the review at `extract_username()`

core/git-access.c
torvalds: Nope. See the tests/testgitstorage.cpp and tests/testparseperformance.cpp.

Neither of them have that 'git_info' thing for the cleanup phase.

core/git-access.c
torvalds: We initialize everything to zero with

         memset(info, 0, sizeof(*info));

in is_git_repository().

core/git-access.c
torvalds: Don't ask me. I didn't want to make changes outside of the actual git_info passing.

core/git-access.c
```diff
@@ -1095,14 +1063,14 @@ struct git_repository *is_git_repository(const char *filename, const char **bran
 	}
 
 	if (!flen)
-		return NULL;
+		return false;
 
 	/*
 	 * This is the "point of no return": the name matches
 	 * the git repository name rules, and we will no longer
 	 * return NULL.
```
dirkhh, line 1069: return false

core/git-access.c
```diff
@@ -1111,50 +1079,80 @@ struct git_repository *is_git_repository(const char *filename, const char **bran
...
-		free(loc);
-		free(branch);
-		return dummy_git_repository;
+	/*
+	 * Remember if the current git storage we are working on is our
+	 * cloud storage.
+	 *
+	 * This is used to create more user friendly error message and warnings.
+	 */
+	is_subsurface_cloud = strstr(info->url, prefs.cloud_base_url) != NULL;
```
dirkhh: Why do we keep the file global variable around. Currently there are two users of that variable that don't have access to the current `git_info`? And for both of them we could trivially pass in the `info` from their calling functions and avoid the function global variable. I'll add a commit to achieve that as further cleanup.

dirkhh (commented):
I think this is good. I think it's even better with the second commit I just pushed.

dirkhh:
OK, I think this is worth merging to keep moving things forward

PR #3444, opened 2022-04-18
Git cleanups

torvalds:
This does a number of cleanups to the git repo handling, with the final commit also making sure that we actually try to save to the local cloud cache repo before we do any remote sync - hopefully making saving more reliable.

This has only gotten some very cursory testing, but each step looks ObviouslyCorrect(tm) to me.

Let's see if the build check machinery agrees.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/file.c
+++ b/core/file.c
@@ -306,11 +306,15 @@ int parse_file(const char *filename, struct dive_table *table, struct trip_table
 			 * Opening the cloud storage repository failed for some reason
 			 * give up here and don't send errors about git repositories
 			 */
-			if (info.is_subsurface_cloud)
+			if (info.is_subsurface_cloud) {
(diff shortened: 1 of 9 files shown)


dirkhh (commented):
this looks sane, it seems to do the right thing with error handling and with handling the `git_info` object.

I am mildly worried about the lack of testing (and I'm on a business trip and have limited time to dig into this and create a few scenarios that I know are tricky).

I really want to stare at the code some more and at least trace that it does what I expect it to do with, for example, no network access.

Also, have you tested this at all with mobile (I'm guessing, mobileOnDesktop)?

PR #3446, opened 2022-04-23
build-system: don't build against Qt6 by default

dirkhh:
As some Linux distros start to ship both Qt5 and Qt6, it actually makes more
sense to build only against Qt6 when the user explicitly asks for it. Having it
preferred over Qt5 seems completely wrong in hind sight.

- [x] Functional change
- [x] Build system change

This deals with the problem people reported in the user group that on some distros we now build against Qt6 and things fails / miss featuers.


dirkhh:
tests pass - this should make no difference except for those building Qt6 from source

PR #3447, opened 2022-04-23
core: avoid crash when merging dive with no cylinders

dirkhh:
Arguably every dive should at least have one cylinder, but an imported
dive from divelogs.de might end up without one. Sadly, that breaks
assumptions that we make in the cylinder remapping.

To work around it, force at least on cylinder to be assumed in the merge
code.

- [x] Bug fix

### Release note:

done

### Mentions:

This might fix the bug Rainer Mohr reported.


dirkhh:
Everyone else is busy - this change seems very small and harmless, so let's merge it to enable Rainer to test this.

PR #3448, opened 2022-04-28
core: ignore secondary sensor on OC dives

dirkhh:
Only the primary sensor readings indicate that a cylinder is used
in open circuit dives. The secondary sensor is only used for the
O2 cylinder in closed circuit diving.

- [x] Bug fix
- [x] Functional change

the old behavior makes no sense to me.

### Release note:

done

### Mentions:

@bstoeger @torvalds


bstoeger (commented):
I never understood the sensors data model. All I know is that this change looks suspicious to me. @torvalds should be the authority on this.

atdotde:
Actually, I think @willemferguson wrote the CCR sensor part...

dirkhh:
Based on @michaelandreen 's comments on the mailing list this code is indeed completely bogus.

PR #3451, opened 2022-05-06
Use 10ft as deco step size in imperial units in profile

atdotde:
The calculation of the deco steps shown in the profile
infobox is somewhat independent of the planner. When
set to imperial units, the distance between deco stops
should be 10ft rather than 3m as 15m is only 49ft.

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -849,7 +849,7 @@ static void calculate_ndl_tts(struct deco_state *ds, const struct dive *dive, st
 	const int ascent_s_per_deco_step = 1;
 	/* how long time steps in deco calculations? */
 	const int time_stepsize = 60;
-	const int deco_stepsize = 3000;
+	const int deco_stepsize = M_OR_FT(3, 10);
 	/* at what depth is the current deco-step? */
 	int next_stop = ROUND_UP(deco_allowed_depth(
 					 tissue_tolerance_calc(ds, dive, depth_to_bar(entry->depth, dive), in_planner),


core/profile.c
```diff
@@ -849,7 +849,11 @@ static void calculate_ndl_tts(struct deco_state *ds, const struct dive *dive, st
 	const int ascent_s_per_deco_step = 1;
 	/* how long time steps in deco calculations? */
 	const int time_stepsize = 60;
-	const int deco_stepsize = 3000;
+	int deco_stepsize;
+	if (prefs.units.length == METERS)
+		deco_stepsize = 3000;
+	else
+		deco_stepsize = 3048;
```
dirkhh: No.
https://github.com/subsurface/subsurface/blob/master/core/units.h#L26-L30

We've had this for ages and I am 99.99% sure that early on the deco code used this.
I'm too lazy to look for this, but yeah... this was a thing that I introduced a very long time ago.

dirkhh (changes requested):
Please pay attention when your libdivecomputer submodule isn't current.
Please use the neat macro that makes all this so much easier to read...

core/profile.c
atdotde: Yes, forgot about M_OR_FT. And yes, also forgot to check for not committing a step backwards in libdivecomputer. See, this time, I made sure the tests pass (as I expected this would affect the TestProfile but that seems to run only in sane unit mode). Hope this is better now.

dirkhh (approved):
Thanks - that looks great

PR #3454, opened 2022-05-12
Handle Mulitple Seac Computers colliding during import

jwobser:
The Seac importer was getting samples based only on dive number,
which was causing samples from different computers but with the
same dive number to become interleaved.

To correct this, the SQL statement was updated to use the
dive_id to query for samples. The table schema uses dive_id
as a primary key, which will enforce uniqueness.

Additionally the device id is hashed and stored.

Reported-by: David Brebera <david.brebera@gmail.com>
Signed-off-by: James Wobser <james.wobser@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In the mailing list, David reported an import error when using multiple seac computers.
Samples from different computers but with the same dive number were becoming interleaved.

This pull request will correct the interleaving behavior, and ensure that samples are only pulled in for a single computer.

### Changes made:

1) Get dive_id from headers_dive query
2) Add dive_id to the WHERE clause of the SQL query for dive samples
3) Use `calculate_string_hash()` to provide dc.deviceid

### Additional information:

Corrected behavior has been verified using the sample DB provided by David.
Verified collisions no longer occur when using the database provided by David.

### DIFF
--- a/core/import-seac.c
+++ b/core/import-seac.c
@@ -17,6 +17,7 @@
 #include "tag.h"
 #include "errorhelper.h"
 #include <string.h>
+#include "divecomputer.h"
 
 /* Process gas change event for seac database.
  * Create gas change event at the time of the
@@ -51,7 +52,7 @@ static int seac_dive(void *param, int columns, char **data, char **column)
 	sqlite3 *handle = state->sql_handle;
 	sqlite3_stmt *sqlstmt;
 
-	const char *get_samples = "SELECT dive_number, runtime_s, depth_cm, temperature_mCx10, active_O2_fr, first_stop_depth_cm, first_stop_time_s, ndl_tts_s, cns, gf_l, gf_h FROM dive_data WHERE dive_number = ? ORDER BY runtime_s ASC";
+	const char *get_samples = "SELECT dive_number, runtime_s, depth_cm, temperature_mCx10, active_O2_fr, first_stop_depth_cm, first_stop_time_s, ndl_tts_s, cns, gf_l, gf_h FROM dive_data WHERE dive_number = ? AND dive_id = ? ORDER BY runtime_s ASC";
 		/*  0 = dive_number
 		 *  1 = runtime_s
 		 *  2 = depth_cm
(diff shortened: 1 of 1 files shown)


dirkhh:
neither of the two build failures are related to your work...

core/import-seac.c
```diff
@@ -303,3 +308,15 @@ int parse_seac_buffer(sqlite3 *handle, const char *url, const char *buffer, int
 
 	return 0;
 }
+
+
+// The Practice of Programming (HASH TABLES, pg. 57)
```
dirkhh: that doesn't follow our coding standard at all.
I understand the desire to cut and paste from a source - but if you do that, you also have to follow the LICENSE of that code - and those are typically not GPL compatible when it comes to books...
We already have a `calculate_string_hash()` - or is there a reason why this needs to be a specific hash function?

dirkhh (changes requested):
I think the string hash should use our existing function unless there is a strong reason not to.

core/import-seac.c
jwobser: No particular reason, I wasn't aware of the existing hash function.

I've already swapped it over to the hash function in `divecomputer.h`
You'll have to remind me, is the preference to rebase then force push so that the pull appears as a single commit?

dirkhh: That's an oversimplification :)
The preference is to have a "sane" git history.
We are happy to have multiple commits in a PR, but we prefer that they don't introduce something and then rip it back out (unless there is a REASON for that... e.g., you move from one infrastructure to a different one, you put in some helpers to make it possible to cleanly do this in multiple steps, and then remove the helpers).

Here, given that this really does seem to be one self-contained single step, I'd rebase and force push, TBH

jwobser: Thanks for the information; that makes a lot of sense.

I have force pushed the version using `calculate_string_hash()`

dirkhh (approved):
that looks great. thanks

core/import-seac.c
atdotde: Just one additional remark: Please make sure that with every commit the build succeeds and it is possible to run the program as otherwise a future git bisect can be painful.

PR #3455, opened 2022-05-13
build-system: consistently mark git directories safe

dirkhh:
Even on platforms that don't have the new git version, yet.
And using the official environment variable that should point to our checked
out tree in the GitHub Action.

- [x] Build system change

This should future proof the other GitHub actions and make things overall more consistent.


dirkhh:
oh how I love the GitHub Actions and how well things are documented and work exactly as documented.

PR #3457, opened 2022-05-20
Added option to choose between different depth grid quantization schema.

elf128:
This allows having 3m depth grid for metric users.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/pref.c
+++ b/core/pref.c
@@ -92,6 +92,7 @@ struct preferences default_prefs = {
 	.auto_recalculate_thumbnails = true,
 	.extract_video_thumbnails = true,
 	.extract_video_thumbnails_position = 20,		// The first fifth seems like a reasonable place
+	.standard_level_grid = false,
 };
 
 /* copy a preferences block, including making copies of all included strings */
--- a/core/pref.c
+++ b/core/pref.c
@@ -92,6 +92,7 @@ struct preferences default_prefs = {
 	.auto_recalculate_thumbnails = true,
 	.extract_video_thumbnails = true,
 	.extract_video_thumbnails_position = 20,		// The first fifth seems like a reasonable place
+	.three_m_based_grid = false,
 };
 
 /* copy a preferences block, including making copies of all included strings */
--- a/core/pref.h
+++ b/core/pref.h
@@ -103,6 +103,7 @@ struct preferences {
 	double      font_size;
 	double      mobile_scale;
 	bool        show_developer;
+	bool        three_m_based_grid;
 
 	// ********** Equipment tab *******
 	const char *default_cylinder;
--- a/core/pref.h
+++ b/core/pref.h
@@ -206,6 +206,7 @@ struct preferences {
 	bool                    coordinates_traditional;
 	enum unit_system_values unit_system;
 	struct units            units;
+	bool					standard_level_grid;
 
 	// ********** UpdateManager **********
 	update_manager_prefs_t update_manager;
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -67,6 +67,7 @@ void qPrefDisplay::loadSync(bool doSync)
 		load_lastState();
 		load_singleColumnPortrait();
 	}
+	disk_standard_level_grid(doSync);
 }
 
 void qPrefDisplay::set_divelist_font(const QString &value)
@@ -148,6 +149,8 @@ HANDLE_PREFERENCE_BOOL(Display, "displayinvalid", display_invalid_dives);
 
 HANDLE_PREFERENCE_BOOL(Display, "show_developer", show_developer);
 
+HANDLE_PREFERENCE_BOOL(Display, "standard_level_grid", standard_level_grid);
+
 void qPrefDisplay::setCorrectFont()
 {
 	// get the font from the settings or our defaults
@@ -197,3 +200,4 @@ HANDLE_PROP_QBYTEARRAY(Display, "MainWindow/windowState", windowState);
 HANDLE_PROP_INT(Display, "MainWindow/lastState", lastState);
 
 HANDLE_PROP_BOOL(Display, "singleColumnPortrait", singleColumnPortrait);
+
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -67,6 +67,7 @@ void qPrefDisplay::loadSync(bool doSync)
 		load_lastState();
 		load_singleColumnPortrait();
 	}
+	disk_three_m_based_grid(doSync);
 }
 
 void qPrefDisplay::set_divelist_font(const QString &value)
@@ -147,6 +148,8 @@ HANDLE_PREFERENCE_INT(Display, "animation_speed", animation_speed);
 HANDLE_PREFERENCE_BOOL(Display, "displayinvalid", display_invalid_dives);
 
 HANDLE_PREFERENCE_BOOL(Display, "show_developer", show_developer);
+
+HANDLE_PREFERENCE_BOOL(Display, "three_m_based_grid", three_m_based_grid);
 
 void qPrefDisplay::setCorrectFont()
 {
--- a/core/settings/qPrefDisplay.h
+++ b/core/settings/qPrefDisplay.h
@@ -25,6 +25,7 @@ class qPrefDisplay : public QObject {
 	Q_PROPERTY(QByteArray windowState READ windowState WRITE set_windowState NOTIFY windowStateChanged)
 	Q_PROPERTY(int lastState READ lastState WRITE set_lastState NOTIFY lastStateChanged)
 	Q_PROPERTY(bool singleColumnPortrait READ singleColumnPortrait WRITE set_singleColumnPortrait NOTIFY singleColumnPortraitChanged)
+	Q_PROPERTY(bool standard_level_grid READ standard_level_grid WRITE set_standard_level_grid NOTIFY standard_level_gridChanged)
 
 public:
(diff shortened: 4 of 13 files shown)


core/pref.c
```diff
@@ -92,6 +92,7 @@ struct preferences default_prefs = {
 	.auto_recalculate_thumbnails = true,
 	.extract_video_thumbnails = true,
 	.extract_video_thumbnails_position = 20,		// The first fifth seems like a reasonable place
+	.standard_level_grid = false,
```
dirkhh: not sure I love the 'standard' in the name... isn't this a `three_m_based_grid` or something like that?
Variables have to start with a letter or underscore... because `3m_based` would be nicer, but not a valid name...

core/pref.h
```diff
@@ -206,6 +206,7 @@ struct preferences {
 	bool                    coordinates_traditional;
 	enum unit_system_values unit_system;
 	struct units            units;
+	bool					standard_level_grid;
```
dirkhh: there looks to be a white space issue here.
I'm not sure which editor you used, but different editors have different ideas how to deal with tabs and indentation and alignment... the easiest way in my experience is to copy the line above and mimic its behavior (because, sadly, for the variable alignment we are not totally consistent between different files... mostly it's spaces, but sometimes there are tabs)
And as I said in the comment on `pref.c` - I'd prefer a different name for this member.

core/settings/qPrefDisplay.cpp
```diff
@@ -197,3 +200,4 @@ HANDLE_PROP_QBYTEARRAY(Display, "MainWindow/windowState", windowState);
 HANDLE_PROP_INT(Display, "MainWindow/lastState", lastState);
 
 HANDLE_PROP_BOOL(Display, "singleColumnPortrait", singleColumnPortrait);
+
```
dirkhh: new empty line?

core/settings/qPrefDisplay.h
```diff
@@ -25,6 +25,7 @@ class qPrefDisplay : public QObject {
 	Q_PROPERTY(QByteArray windowState READ windowState WRITE set_windowState NOTIFY windowStateChanged)
 	Q_PROPERTY(int lastState READ lastState WRITE set_lastState NOTIFY lastStateChanged)
 	Q_PROPERTY(bool singleColumnPortrait READ singleColumnPortrait WRITE set_singleColumnPortrait NOTIFY singleColumnPortraitChanged)
+	Q_PROPERTY(bool standard_level_grid READ standard_level_grid WRITE set_standard_level_grid NOTIFY standard_level_gridChanged)
```
dirkhh: of course the different name will affect almost everything you did. sorry.

desktop-widgets/preferences/preferences_defaults.cpp
```diff
@@ -28,11 +28,16 @@ void PreferencesDefaults::refreshSettings()
 	ui->fontsize->setValue(qPrefDisplay::font_size());
 	ui->velocitySlider->setValue(qPrefDisplay::animation_speed());
 
+	if (qPrefDisplay::standard_level_grid())
+		ui->gridLeveled->setChecked(true);
+	else
+		ui->gridNatural->setChecked(true);
```
dirkhh: again, think about the names... `generic` and `three_m_based` ?

desktop-widgets/preferences/preferences_defaults.ui
```diff
@@ -14,17 +14,24 @@
...
+      <property name="leftMargin">
+       <number>5</number>
+      </property>
+      <property name="topMargin">
+       <number>5</number>
+      </property>
+      <property name="rightMargin">
+       <number>5</number>
+      </property>
+      <property name="bottomMargin">
```
dirkhh: this is the annoying noise that different versions of the Qt tools add... frustrating, but not your fault.

desktop-widgets/preferences/preferences_defaults.ui
```diff
@@ -35,18 +42,16 @@
       <item>
        <widget class="QFontComboBox" name="font"/>
       </item>
-
       <item>
        <widget class="QLabel" name="label_default1">
+        <property name="text">
+         <string/>
```
dirkhh: I'm confused - what does this do? an empty label with a default name?

desktop-widgets/preferences/preferences_defaults.ui
```diff
@@ -114,11 +124,39 @@
...
+       <widget class="QRadioButton" name="gridNatural">
+        <property name="text">
+         <string>generic ( 1, 2, 4, 5, 10 )</string>
+        </property>
+       </widget>
+      </item>
+      <item>
+       <widget class="QRadioButton" name="gridLeveled">
+        <property name="text">
+         <string>standard levels ( 1, 3, 6, 15 )</string>
```
dirkhh: again... 'based on 3m intervals' or something like that...

profile-widget/divecartesianaxis.cpp
```diff
@@ -30,6 +30,7 @@ DiveCartesianAxis::DiveCartesianAxis(Position position, bool inverted, int integ
 	max(0),
 	textVisibility(textVisible),
 	lineVisibility(linesVisible),
+	metricIntervals(false),
```
dirkhh: now we have a new naming logic... what does `metricIntervals` mean?

profile-widget/divecartesianaxis.cpp
```diff
@@ -101,7 +102,7 @@ int DiveCartesianAxis::getMinLabelDistance(const DiveCartesianAxis &timeAxis) co
 	return int(ceil(interval));
 }
 
-static double sensibleInterval(double inc, int decimals, bool is_time_axis)
+static double sensibleInterval(double inc, int decimals, bool is_time_axis, bool is_metric_depth)
```
dirkhh: and now we have `is_metric_depth`
not sure you should pass this as an argument as you get the answer whether it's imperial or metric from the preferences... but I'm afraid, looking at the code, that this is actually about the `three_m_based_grid`

profile-widget/divecartesianaxis.cpp
```diff
@@ -130,11 +131,24 @@ static double sensibleInterval(double inc, int decimals, bool is_time_axis)
...
+		// Do increments quantized to 3. In general: 1, 3, 6, 15
+		if (inc_int > 6)
+			inc_int = 15;
+		else if (inc_int > 3)
+			inc_int = 6;
+		else if (inc_int == 2)
+			inc_int = 3;
+	}
+	else
+	{
```
dirkhh: can you take a look at the coding style document 
https://github.com/subsurface/subsurface/blob/master/CODINGSTYLE.md

this should be
```
        } else {
```

profile-widget/profilescene.cpp
```diff
@@ -164,6 +164,7 @@ ProfileScene::ProfileScene(double dpr, bool printMode, bool isGrayscale) :
 	addTissueItems<0,16>(dpr);
 
 	percentageItem->setZValue(1.0);
+	profileYAxis->setMetricIntervals( qPrefDisplay::standard_level_grid() );
```
dirkhh: yeah, this is not a great set of names...

dirkhh (changes requested):
First off, thank you for your contribution. It is really appreciated.
Now this review might look a bit negative - I tried to be detailed and to give lots of feedback. It's not uncommon that in a first PR there are a few things that I stumble over.
We have two documents that might be interesting to look at. CODINGSTYLE.md and CONTRIBUTING.md -- both contain a lot of good information.
There you'll find comments about the requirement to "sign off" on your commits. Some guidelines for your commit message, etc.
My biggest concern with the actual code is around the naming of properties and screen text - and the inconsistency in how this is done. If you read through this as a single PR I think you'll see what I mean.
I think it would be great to define one (well, two) terms to use for the two options and to use those consistently.
One thing that I haven't looked at at all is what your code does if the user is in imperial mode. Is the settings option shown?

Please don't be discouraged by all this feedback - I'm excited to see the change and happy to work with you to get this to the point where the change can go in.

core/settings/qPrefDisplay.h
elf128: That's fine. I was expecting a battle over variable naming. :)

profile-widget/divecartesianaxis.cpp
elf128: sensibleInterval can be called from any "Axis", and we've got quite a few.
Naming is a leftover. Initially I thought using units as an indicator to choose quantization. But then I realized that not all metric folks will be happy to see 12m instead of 10m. It has to be an option in preferences. This is why I end up having a zoo of names.

I'll fix it.

core/pref.c
elf128: Ok, "three_m_based_grid" it is.

core/pref.h
elf128: I can see, you use tabs for indentation, that's why I left tabs in settings. Apparently that also caused conversion of spaces into tabs. 
My bad. I usually try to mimic convention of the file I'm editing.

elf128: Also, it's in the wrong section. It used to be in Units initially, and then I moved it into Display.

desktop-widgets/preferences/preferences_defaults.ui
elf128: That's QT tools decided to change order of properties in ui file. I'll fix this file manually.

elf128:
Some unnecessary changes were commit and added to this PR by accident. 
Please disregard this one. 
New, clean request #3458 was created instead.

PR #3458, opened 2022-05-20
Added option to choose between different depth grid quantization schema.

elf128:
This allows having 3m depth grid for metric users.

* Added qml changes to make it work on mobile.
* All original properties ( named diferently ) were renamed to three_m_based_grid everywhere to be consistent.
* Plus other small changes requested during review.

Signed-off-by: Vlad A. <elf128@gmail.com>

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This request is a duplicate of #3457, but this one has all changes under one commit.

### DIFF
--- a/profile-widget/divecartesianaxis.cpp
+++ b/profile-widget/divecartesianaxis.cpp
@@ -30,6 +30,7 @@ DiveCartesianAxis::DiveCartesianAxis(Position position, bool inverted, int integ
 	max(0),
 	textVisibility(textVisible),
 	lineVisibility(linesVisible),
+	gridIsMultipleOfThree(false),
 	labelScale(labelScale),
 	dpr(dpr),
 	transform({1.0, 0.0})
@@ -101,7 +102,7 @@ int DiveCartesianAxis::getMinLabelDistance(const DiveCartesianAxis &timeAxis) co
 	return int(ceil(interval));
 }
 
-static double sensibleInterval(double inc, int decimals, bool is_time_axis)
+static double sensibleInterval(double inc, int decimals, bool is_time_axis, bool is_multiple_of_three)
 {
 	if (is_time_axis && inc < 60.0) {
 		// for time axes and less than one hour increments, round to
@@ -130,11 +131,22 @@ static double sensibleInterval(double inc, int decimals, bool is_time_axis)
 
 	double digits_factor = pow(10.0, digits);
 	int inc_int = std::max((int)ceil(inc / digits_factor), 1);
-	// Do "nice" increments of the leading digit. In general: 1, 2, 4, 5.
-	if (inc_int > 5)
-		inc_int = 10;
-	if (inc_int == 3)
-		inc_int = 4;
+	if (is_multiple_of_three)
+	{
+		// Do increments quantized to 3. In general: 1, 3, 6, 15
+		if (inc_int > 6)
+			inc_int = 15;
+		else if (inc_int > 3)
+			inc_int = 6;
+		else if (inc_int == 2)
+			inc_int = 3;
+	} else {
+		// Do "nice" increments of the leading digit. In general: 1, 2, 4, 5.
+		if (inc_int > 5)
+			inc_int = 10;
+		if (inc_int == 3)
+			inc_int = 4;
+	}
 	inc = inc_int * digits_factor;
 
 	return inc;
@@ -165,7 +177,7 @@ void DiveCartesianAxis::updateTicks(int animSpeed)
 
 	// Round the interval to a sensible size in display units
 	double intervalDisplay = stepValue * transform.a;
-	intervalDisplay = sensibleInterval(intervalDisplay, fractionalDigits, position == Position::Bottom);
+	intervalDisplay = sensibleInterval(intervalDisplay, fractionalDigits, position == Position::Bottom, gridIsMultipleOfThree);
 
 	// Choose full multiples of the interval as minumum and maximum values
 	double minDisplay = transform.to(dataMin);
@@ -476,6 +488,11 @@ double DiveCartesianAxis::maximum() const
 double DiveCartesianAxis::minimum() const
 {
 	return min;
+}
+
+void DiveCartesianAxis::setGridIsMultipleOfThree(bool arg1)
+{
+	gridIsMultipleOfThree = arg1;
 }
 
 std::pair<double, double> DiveCartesianAxis::screenMinMax() const
--- a/profile-widget/profilescene.cpp
+++ b/profile-widget/profilescene.cpp
@@ -164,6 +164,7 @@ ProfileScene::ProfileScene(double dpr, bool printMode, bool isGrayscale) :
 	addTissueItems<0,16>(dpr);
 
 	percentageItem->setZValue(1.0);
+	profileYAxis->setGridIsMultipleOfThree( qPrefDisplay::three_m_based_grid() );
 
 	// Add items to scene
 	addItem(diveComputerText);
@@ -303,6 +304,8 @@ void ProfileScene::updateAxes(bool diveHasHeartBeat, bool simplified)
 
 	if (width <= 10.0 * dpr)
 		return clear();
+
+	profileYAxis->setGridIsMultipleOfThree( qPrefDisplay::three_m_based_grid() );
 
 	// Place the fixed dive computer text at the bottom
 	double bottomBorder = sceneRect().height() - diveComputerText->height() - 2.0 * dpr * diveComputerTextBorder;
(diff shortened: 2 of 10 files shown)


dirkhh:
At first glance this looks good. Now I need to find time to try it both on desktop and mobile and in metric and imperial and make sure this doesn't fail in ways I cannot think of by just looking at the code - and I'm not sure I'll have time for that today (I'll try). I have authorized the test builds (for new contributors GitHub doesn't run those automatically) - let's see if that stumbles over anything...

bstoeger:
Do we really need another obscure setting for that? If triple steps are reasonable or even favorable, just always allow them?

Also it seems strange to me that this would be specific for metric users. I though 3 ft was some kind of "magic" increment, since it corresponds to 1 m.

🤷

elf128:
Well, that's the thing. 3ft is not 1m. it's 0.9144 m. 
10% is not a negligible error. Much closer approximation is when you do it in reverse.
10' is 3.048m ~3m. 

The irony is that usually metric people prefer multiple of 10 since entire unit system is build around it. For the same reason imperial folks also used to use multiple of 3 or 12, cause it's common denominator in their length units. This is why we have an upside-down case when both have to use an opposite logic in order to agree. 

So in general, non-metric folks would prefer having nice ×5 and ×10 grid cause they mostly count levels by tens, meanwhile metric folks would prefer multiples of 3. 

However, there's third group of people, metric people in recreational case, where people don't care about standardized levels, and doing safety stop at 5m. Those will be pissed off if you switch native ×10 grid, with some strange non-sense.

Anyway, since I can see an argument for and against grid as multiple of 3 in both metric and non-metric cases it is a perfect case for just allow user to decide. **No one ever complain about too many options, but people will fight tirelessly in case it will be hard coded.**  

As a background info. I live in Canada, the country which become metric more then half of a century ago, but still having really strange opinion on units. So unlike any other divers, who just use one units and sometimes face another one due to travel or internet, we are having "unit" conversation almost every time you go diving with new people.

dirkhh:
> Well, that's the thing. 3ft is not 1m. it's 0.9144 m. 10% is not a negligible error. Much closer approximation is when you do it in reverse. 10' is 3.048m ~3m.

The odd thing is that there are these two parallel worlds that "kinda" align, but not really, and that each think of themselves as the "correct" one.
The reason why deco stops for metric divers are at 3/6/9/12m are... because the first set of tables came from the US navy and was in ten feet increments. And while we constantly find people here who argue with us about 1% precision, the reality is that deco physics is all just random guesses, hiding behind fancy formulas and fudge factors.

That said, people get mad if you don't do exact calculations in their measurement system (see the recent issue with deco stops at 49ft)

> However, there's third group of people, metric people in recreational case, where people don't care about standardized levels, and doing safety stop at 5m. Those will be pissed off if you switch native ×10 grid, with some strange non-sense.

Yes, most recreational divers (and they were the initial target audience for Subsurface, BTW) do not think in 3m increments.
And because of the challenge that wave action brings to recreational divers at 10ft, most recreational instruction sets these days recommend doing a safety stop at 5m / 15ft. With no magically deco math to justify this...

> Anyway, since I can see an argument for and against grid as multiple of 3 in both metric and non-metric cases it is a perfect case for just allow user to decide. **No one ever complain about too many options, but people will fight tirelessly in case it will be hard coded.**

This is the first time I notice you say some silly and obviously wrong.
The worst thing you can do is to have too many options. Subsurface already comes close to (or more honestly, has long crossed) this threshold. The logs for the cloud storage backend show that the vast majority of our users go on two or fewer dive trips a year. They use the app with long, long breaks in between. Ironically, just as I looked (to make sure that this is still accurate) I see someone connect to the cloud with Subsurface 4.4.2 -- that was released a little over seven years ago... 
And they do not want to dig through pages and pages of options. We already have too many.

> As a background info. I live in Canada, the country which become metric more then half of a century ago, but still having really strange opinion on units. So unlike any other divers, who just use one units and sometimes face another one due to travel or internet, we are having "unit" conversation almost every time you go diving with new people.

I grew up in Germany, but learned to dive in the US. I get it. To me your change makes sense. I just need to actually test it.

bstoeger:
> No one ever complain about too many options, but people will fight tirelessly in case it will be hard coded.

People constantly whine about too many options in KDE.

If you really need this setting, from a UI point of view, it should probably go near the "Toggle calculated ceiling with 3m equivalents" button next to the chart. Which makes me wonder whether this existing feature isn't already enough for your usecase.

bstoeger:
> it should probably go near the "Toggle calculated ceiling with 3m equivalents" button next to the chart.

Alternatively, that option could be hidden in the preferences dialog. I always found it way to prominently placed. After all, it doesn't switch on/off a chart feature, but changes the behavior of a chart feature.

🤷

profile-widget/divecartesianaxis.cpp
```diff
@@ -130,11 +131,22 @@ static double sensibleInterval(double inc, int decimals, bool is_time_axis)
...
-	// Do "nice" increments of the leading digit. In general: 1, 2, 4, 5.
-	if (inc_int > 5)
-		inc_int = 10;
-	if (inc_int == 3)
-		inc_int = 4;
+	if (is_multiple_of_three)
+	{
+		// Do increments quantized to 3. In general: 1, 3, 6, 15
+		if (inc_int > 6)
+			inc_int = 15;
```
bstoeger, line 138: Another small comment:

The jump to '15' is illogical.

You do accept '1' as the leading digit, so '10' should be fine as well. After all an increment equivalent to '10' could come from a different rounding above.

elf128: This is intentional, cause otherwise you can get steps like 0, 10, 20, 30, 40 which is not multiple of 3. 
Having 15 will force steps 0, 15, 30, 45  and then 0, 30, 60, 120 for really deep dives.
And yes, I've tested that case.

bstoeger: > This is intentional, cause otherwise you can get steps like 0, 10, 20, 30, 40 which is not multiple of 3.

Neither is 1 a multiple of 3. It appears illogical to me that 1, 2, 3, 4 is fine, but 10, 20, 30, 40 is not.

But my real point is: in the admittedly unlikely case, but stranger things have happened, that inc calculates as 10.0 straight, then you get `inc_int = 1` and `digits = 1` and thus increments 10, 20, 30, 40. No?

That said, since I will not be using this option I really don't mind. It just strikes me as weird code.

bstoeger:
One more comment: This seems to also affect the partial pressure axis. I doubt that this is intentional?

dirkhh:
> One more comment: This seems to also affect the partial pressure axis. I doubt that this is intentional?

That seems like a mistake...

elf128:
That should not be the case. 
This is how it looks in the build.
image

profile-widget/profilescene.cpp
```diff
@@ -164,6 +164,7 @@ ProfileScene::ProfileScene(double dpr, bool printMode, bool isGrayscale) :
 	addTissueItems<0,16>(dpr);
 
 	percentageItem->setZValue(1.0);
+	profileYAxis->setGridIsMultipleOfThree( qPrefDisplay::three_m_based_grid() );
```
elf128, line 167: Notice, that only profileYAxis receives a call to set grid as multiple of three.

bstoeger:
Yes, you're right.

So from my side, this is OK - not too happy about another option but *shrug*.

profile-widget/divecartesianaxis.cpp
elf128: I do agree, that it would be cleaner to also change the power of digit factor. And use 15 instead of 10 everywhere. 
```
-double digits_factor = pow(10.0, digits);
+double digits_factor = pow(15.0, digits);
```
But that would lead to bunch of if's.

elf128: You know what. Maybe you right. Let me try it.

elf128: Oh, wait. no. Mornings doesn't do good to me.
This is not going to work. It will turn whole numbering to septadecimal. 
NM

dirkhh:
ok, now that the mailserver is happy again (my main gateway is offline because of a datacenter move) I have time to play with this...

dirkhh:
This seems to work and does what you intend. I share Berthold's concern about the number of options we keep adding, but I also understand where you are coming from... eventually we'll pay the price for all this, I'm sure :/

PR #3462, opened 2022-06-04
Deal with negative variation times

atdotde:
When computing plan variations, deco can get shorter when
staying longer if the last step is actually already at
off gasing depth. FRACTION forces unsiged, so this introduces
a sign aware version of FRACTION that returns a sign character
in addition.

Reported-by: Patrick Naujoks <p.naujoks@me.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -1258,9 +1258,9 @@ void DivePlannerPointsModel::computeVariations(struct diveplan *original_plan, c
 	restore_deco_state(save, &ds, false);
 
 	char buf[200];
-	sprintf(buf, ", %s: + %d:%02d /%s + %d:%02d /min", qPrintable(tr("Stop times")),
-		FRACTION(analyzeVariations(shallower, original, deeper, qPrintable(depth_units)), 60), qPrintable(depth_units),
-		FRACTION(analyzeVariations(shorter, original, longer, qPrintable(time_units)), 60));
+	sprintf(buf, ", %s: %c %d:%02d /%s %c %d:%02d /min", qPrintable(tr("Stop times")),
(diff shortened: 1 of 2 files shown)


qt-models/diveplannermodel.cpp
```diff
@@ -1258,9 +1258,9 @@ void DivePlannerPointsModel::computeVariations(struct diveplan *original_plan, c
 	restore_deco_state(save, &ds, false);
 
 	char buf[200];
-	sprintf(buf, ", %s: + %d:%02d /%s + %d:%02d /min", qPrintable(tr("Stop times")),
-		FRACTION(analyzeVariations(shallower, original, deeper, qPrintable(depth_units)), 60), qPrintable(depth_units),
-		FRACTION(analyzeVariations(shorter, original, longer, qPrintable(time_units)), 60));
+	sprintf(buf, ", %s: %c %d:%02d /%s %c %d:%02d /min", qPrintable(tr("Stop times")),
+		SIGNED_FRAC(analyzeVariations(shallower, original, deeper, qPrintable(depth_units)), 60), qPrintable(depth_units),
+		SIGNED_FRAC(analyzeVariations(shorter, original, longer, qPrintable(time_units)), 60));
```
dirkhh, line 1263: as discussed in email - does it make sense to have a negative time here?
I never use this - so if you tell me this makes sense to the user, I'll take the change....

dirkhh (approved):
after seeing your email response... ok... feels weird, but I understand your point that this is about a variation in duration, and that can indeed be negative.

PR #3464, opened 2022-06-08
Change semantic for editing cylinders for multiple dives

dirkhh:
Instead of trying to find matching cylinders, trigger on the cylinder
number first and then only edit that n-th cylinder if it matches the one
in the current dive.

- [x] Bug fix
- [x] Functional change

I'm not yet sure that this is a complete fix, but it significantly improves the situation and I can no longer reproduce the issue that started all this.

### Release note:

done

### Documentation change:

N/A

### Mentions:

@bstoeger 
@torvalds


dirkhh (commented):
Given that the code is obviously brilliant (err... I mean, that no one else commented)

I guess I'll merge it. In my testing it seems to work in all the cases I was able to come up with.

PR #3470, opened 2022-06-18
[SUGGESTION] filter: normalize text of fulltext search to base letters

bstoeger:
The liter symbol is written as 'ℓ'. To allow searching for
that, normalize unicode strings to their base letter. This
corresponds to the 'compatibility' mode.

We might also think about stripping diacritics.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Someone complained on the mailing list that script-l is not found in fulltext searches. Therefore, I suggest to normalize to the base letter.

This is not even compile tested, since I'm currently "on the road".

### Changes made:

1) Normalize unicode text with "compatibility mode".

### Release note:

Done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- filter: normalize unicode code points to base letter (for searching ℓ, etc.)
 - core: when modifying cylinders across multiple dives, match cylinder number before comparing type
 - core: merge all properties in a dive, including current, waveheight, etc
 - core: prevent crash when merging dives without cylinders (as we might get when importing from divelogs.de)
--- a/core/fulltext.cpp
+++ b/core/fulltext.cpp
@@ -85,8 +85,9 @@ bool fulltext_dive_matches(const struct dive *d, const FullTextQuery &q, StringF
 
 // Class implementation
 
-// Take a text and tokenize it into words. Normalize the words to upper case
-// and add to a given list, if not already in list.
+// Take a text and tokenize it into words. Normalize the words to the base
+// upper case base character (e.g. 'ℓ' to 'L') and add to a given list,
+// if not already in list.
 // We might think about limiting the lower size of words we store.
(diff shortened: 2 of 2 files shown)


bstoeger:
I compiled this now and it seems to work. I don't know if this solves the reported problem, but in any case it seems like the right thing to do.

dirkhh (commented):
I tried to read up on what the `NormalizationForm_KD` actually does but reading those standard documents just makes me sad... and I still have no idea if that's what I would want here. I mean, those are all words (maybe), but... what the heck? (https://www.unicode.org/reports/tr15/)

bstoeger:
To my (limited) understanding it does two things:

In Unicode "ö" can be represented in (at least) two ways: by the single code point  "Latin Small Letter O with Diaeresis" U+00F6 or by two code points, viz. "Latin Small Letter O" U+006F followed by "Combining Diaeresis" U+0308.

Normalization turns one into the other, so that strings can be compared codepoint wise. Moreover, we could convert the letters with diacritics to their base form by removing the "combining diacritical marks" so that "o" also finds "ö", "ò", "ó", etc.

But this is not the point here.

What the employed normalization function additionally does is turning letters into their base letters. E.g.
ℓ -> l
ℂ -> C

Thus, these letters can be found by fulltext search even without typing them in in their specific form.

Note that by upper-casing the words we already did a form of normalization. We do not want the search to be case sensitive.

I'm not sure if this solves the reported problem, but I still think it is the correct thing to do. After all, people may copy&paste weird Unicode letters from somewhere and then are not able to search them.

In my very limited test, this seemed to work.

dirkhh:
thanks for the explanation. That does seem to make sense to me.

dirkhh (approved):
finally played with it and this seems fine.

PR #3472, opened 2022-06-30
Add local time to dive info on profile

### LINKED ISSUE #3469: Feature Request: New dive graph toggle - Show 24hr/local time
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [x] New feature request
- [ ] Discussion request

### Issue long description:

A (hopefully) small feature request. 
If possible, it would be very useful for marine species recording if Subsurface had a feature where the user could view the 24hr time directly on the dive graph (as part of the pop out bubble when mousing over the profile). This would be in addition to (or an alternative to) the bottom timer time value.

The reason for this request is that I attempt to create accurate records of the species I see on my dives (for use in citizen science), with the aid of underwater camera. In order to achieve accurate depth readings of the animals I photograph, I sync the clocks on my camera and dive computer so I can note the exact depth and temperature that a particular animal is found in. 

Although I can add the instantaneous bottom time to the dive start time to determine the depth at which a photo was taken, it would be much easier if I could have a 24hr value directly on the dive graph (which would also allow for better synchronisation of dive and photos).

Thank you for all the hard work that has gone into Subsurface, it is a wonderful tool.

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Feature requested in #3469. Needs discussion.

### Changes made:

1) Minor code cleanups.
2) Add full datetime to dive info on profile.
### Related issues:

Fixes #3469.

### Release note:

Done.
### Documentation change:

No.

### DIFF
--- a/core/string-format.cpp
+++ b/core/string-format.cpp
@@ -2,7 +2,7 @@
 #include "deco.h"
 #include "dive.h"
 #include "divesite.h"
-#include "membuffer.h"
+#include "format.h"
 #include "profile.h"
 #include "qthelper.h"
 #include "subsurface-string.h"
@@ -309,7 +309,7 @@ QString formatTripTitleWithDives(const dive_trip *trip)
 
 static QString formatPlotInfoInternal(const dive *d, const plot_info *pi, int idx)
 {
-	membufferpp b;
+	QString res;
 	int pressurevalue, mod, ead, end, eadd;
 	const char *depth_unit, *pressure_unit, *temp_unit, *vertical_speed_unit;
 	double depthvalue, tempvalue, speedvalue, sacvalue;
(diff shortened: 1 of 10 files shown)


dirkhh (changes requested):
looks like a libdivecomputer change snuck into the last commit...

also, the tooltip is already HUGE. Do we really want to unconditionally add another entry to it? Even if most people have no use for it? No offense to the requestor, but I really don't care AT ALL about the exact time of a specific spot - I do care about how far into the dive that spot is...
so if we want to do that, it does need a preference option and it does need to be off by default...

bstoeger:
> looks like a libdivecomputer change snuck into the last commit...

Yes, for some reason I am stuck with old libdivecomputer. `git submodule update` or a direct `git pull` in the libdivecomputer directory just times out. :(

```
fatal: unable to connect to github.com:
github.com[0: 140.82.121.3]: errno=Connection timed out

Unable to fetch in submodule path 'libdivecomputer'; trying to directly fetch a17e466bd1d2e675666e20862182d618cf6d7190:
fatal: unable to connect to github.com:
github.com[0: 140.82.121.4]: errno=Connection timed out

fatal: Fetched in submodule path 'libdivecomputer', but it did not contain a17e466bd1d2e675666e20862182d618cf6d7190. Direct fetching of that commit failed.
fatal: 
```

dirkhh:
> Yes, for some reason I am stuck with old libdivecomputer. `git submodule update` or a direct `git pull` in the libdivecomputer directory just times out. :(

Yes, that's the fact that GitHub turned off the `git://` protocol.
Likely in your repo you still have a reference to that protocol. My guess would be it's 
`subsurface/.git/modules/libdivecomputer/config` , the `url` component of the remote `origin`

bstoeger:
Thanks - replacing 'git:' by 'https:' fixed it.

dirkhh:
> also, the tooltip is already HUGE. Do we really want to unconditionally add another entry to it? Even if most people have no use for it? No offense to the requestor, but I really don't care AT ALL about the exact time of a specific spot - I do care about how far into the dive that spot is... so if we want to do that, it does need a preference option and it does need to be off by default...

Did you respond to this comment and I missed it?

bstoeger:
Sorry, I'm currently working on such an absurd amount of projects, that I didn't find time contemplating on that.

PeterZaal:
I'm with Dirk, in 2000+ dives I have never needed to know the exact time into the dive. I do understand the particular situation of the requestor, but its quite a niche case... If at all, you could maybe add it after the time in the first line between parentheses, something like this "@: 24:50 (10:25:20)". That way it won't add another line.
Also, I think the requestor asked for the time only; the newly added line also adds the date?

bstoeger:
Now adds the local time in parentheses.

Since I had to convert this into C++ (I didn't want to format the localized time in C++ and convert it into a C string which then has to be manually free()d), I also did some refactoring of the formatting code. In retrospect probably completely pointless, but I feel better not having format strings with "%s" placeholders in the translations. As the commit message states, this prevents us from having right-to-left languages. If that is needed, I can convert to QString-style "%1" placeholders.

core/string-format.cpp
```diff
@@ -318,41 +318,43 @@ static QString formatPlotInfoInternal(const dive *d, const plot_info *pi, int id
 	const struct plot_data *entry = pi->entry + idx;
 
 	depthvalue = get_depth_units(entry->depth, NULL, &depth_unit);
-	put_format_loc(&b, translate("gettextFromC", "@: %d:%02d\nD: %.1f%s\n"), FRACTION(entry->sec, 60), depthvalue, depth_unit);
+	put_format_loc(&b, "%s: %d:%02d\n", translate("gettextFromC", "@"), FRACTION(entry->sec, 60));
+	put_format_loc(&b, "%s: %.1f%s\n", translate("gettextFromC", "D"), depthvalue, depth_unit);
```
dirkhh: I wonder if I agree with your intent here...
this creates new translation strings `"@"` and `"D"` which are completely context free and hard to understand for translators. While the existing translation string is weird, at least that's something they can easily find in the sources and have all already translated...

I hear you regarding potential crashes... but as it is I'd find this rather hostile for translators.

I don't think I've played with the translation comment feature to see if that gets surfaced in Transifex. The idea would be to follow the translate call with a magic comment like this
```
put_format_loc(&b, "%s: %d:%02d\n", translate("gettextFromC", "@"), FRACTION(entry->sec, 60));
//: this designates the time shown in the tooltip - should be a single character
```

I'll need to play with that before I'll agree with this change.

dirkhh: But one thing should be clear - this will create a noticeable amount of work for the translators. On strings that right now all work just fine. Even if I agreed with the underlying logic here - that's something I try to be very careful about.

dirkhh (commented):
I struggle with this one.
(a) I really don't know if I'm willing to add the data that the requestor wants to the time stamp in the info box for every user. My guess is that 0.1% or fewer of our users care about the absolute time during the dive the way the requestor does. Yet we changes this for everyone and make the info box even more busy and hard to visually 'parse'
(b) this creates a fairly noticeable number of new strings for the translators - and strings that are context free and make no sense. I mean, how do you translate the letter 'D' without knowing what it's used for? Or that at symbol '@'?
(c) I know you dislike C code. I get it. But this does feel a bit gratuitous.

PR #3475, opened 2022-07-07
build-system: allow adaptive building without WebKit

dirkhh:
This should avoid build failures on platforms where we don't know if
QtWebKit will be available or not.

The options for printing and user manual are awkward to work with. This
all needs to be cleaned up at some point. Right.

- [x] Build system change

with this we should be able to trigger OBS builds for openSUSE Leap 15.4 and Tumbleweed again

### Changes made:

this actually tries to make the logic of print and user manual enabling a bit cleaner.
We have options to turn those off. But if the options to turn those off aren't set, yet we still don't find QtWebKit, we don't fail the build, instead we simply automagically set those two options.


dirkhh:
This seems to not have broken anything, so let's merge and move on :)

Issue #3477, opened 2022-07-29
Automated merging during cloud sync after editing notes on multiple devices loses tank information


mwerle:
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I recently performed the following actions:

1. Imported dives (from multiple dive computers, one of which is AI) on my phone, then sync'd to cloud.
2. Opened Subsurface on my table and did some initial editing. Sync'd some of it to cloud; some not (lost internet connectivity on way home).
3. Opened Subsurface on my PC (Windows, v5.08 of Subsurface), sync'd, then proceeded to edit the dives, and sync'd to cloud.
4. Sync'd the tablet data as I had written some notes which I did not want to re-edit.
5. Lost tank, AI, exposure protection, and weight data for several dives.

ANALYSIS:
The "Notes" data was incorrectly merged, resulting in an additional quote character to be in the data.

EG: 

```
duration 49:00 min
tags "boat"
divesiteid c5299b50
suit "5mm long + 3mm hooded vest"
notes "First boat dive, (blah blah blah
    blah blah blah.
    blah blah blah "
 
    blah blah blah.
    blah blah blah "
 cylinder vol=10.0l description="10.0ℓ" o2=32.0% start=190.295bar end=52.676bar depth=39.845m
weightsystem weight=2.0kg description="weight"
```

Of course, this then stuffs up the parsing, which makes Subsurface think there is no tank.  But instead of reading the dive-computer data from the separate file and re-creating the tank, it _deletes_ the AI data from the dive computer file!

### Operating system:

Phone and tablet:
Created by Subsurface-mobile:3.4.3(5.0.6.83):Android  (9.0):arm64:en-GB

PC:
Created by Subsurface:5.0.8.0:Windows 10 Version 2009:x86_64:en-GB

Shearwater Perdix AI; Shearwater Peregrine

The issue arose due to a merge created by editing on multiple devices in parallel.

### Subsurface version:
Created by Subsurface-mobile:3.4.3(5.0.6.83):Android  (9.0):arm64:en-GB
Created by Subsurface:5.0.8.0:Windows 10 Version 2009:x86_64:en-GB

Official releases of Subsurface; did not try to reproduce the issue artificially as yet (sorry, would debug and attach a patch, but very limited time at present).

### Steps to reproduce:

Please see the "Issue Long Description".

I think it should be sufficient to:

1. Have a dive with AI data
[...]

PR #3479, opened 2022-07-31
git parser: handle left-over multi-line quoted strings better

torvalds:
The git save format is designed to be entirely line-based, where all the
dive data is on individual lines that are independent.

That is very much by design, so that you can merge these files
automatically, and not worry about what it does to the context (contrast
this to structured files like JSON or XML, where you have multiple
levels of indentation, and the context of a line matters).

So the parser can just ignore any conflict markers, and parse everything
one line at a time.

Well, almost.

We do have *one* special form of multi-line context, where flowed text
(think things like dive notes) will have one "header line" that starts
the note, and then it can continue for several lines until the final
line that ends the quote.

In such a situation, the dive merging can result in a partially merged
string note, which has the ending line from one dive, and then continues
with more string data from the other dive.

That will confuse our parser mightily, because it will have seen the end
of the string, and parsed the rest of those string comments as garbage lines.

That part in itself is fine - the garbage lines won't pass as any real
data (because they don't start with a proper keyword), but while parsing
that garbage the *next* end of the string will be seen as a start of a
new string.

And *that* then confuses the git parser to think that the line after
that is now part of the string, and so it won't correctly parse the
non-string line that follows.

To give a more concrete example, the git dive data (here indented and
abbreviated) might look like this:

	suit "5mm long + 3mm hooded vest"
	notes "First boat dive.
		Giant-stride entry."
		Saw a turtle."
	cylinder vol=10.0l description="10.0ℓ" depth=66.019m

where the two notes from the two dives were

	notes "First boat dive.
		Giant-stride entry"

and

	notes "First boat dive.
		Saw a turtle."

respectively, and the merged result contained parts of both.

When we parse this, we will parse the 'notes' line as having the string

	First boat dive.
	Giant-stride entry

which is fine. But then the next line will be that

	Saw a turtle."

and now the ending double quote character on that line will be seen as
the beginning of a new string, and the cylinder information on the next
line will then be mixed up.  The resulting mess will be ignored, but in
the process the data on the "cylinder" line will basically have been
lost.

There are several ways to deal with this, but this particular fix
depends on the fact that we can recognize stale string continuation
lines: they are either empty (for an empty line), or they start with a
TAB character.

So to solve the problem with the mis-identified end quote, this
recognizes that we're in such a "stale left-over comment line" context,
and will just skip such lines entirely.

That does mean that when you have conflicts in dive note sections due to
having edited the dive concurrently on different machines, you may just
lose some of the edits.

But this way at least you shouldn't lose any other data due to the merge
conflict.

NOTE! We could try to improve on this by instead noticing that a "end of
multi-line string has a continuation entry on the next line", and just
say "ok, that wasn't a real end after all".

But that would be an independent thing anyway - this "ignore stale text
comment lines" logic would be required anyway, in case those stale text
comments ended up somewhere *else* than right after another text line.

So do this more important fix first.

Reported-by: Michael Werle

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -1365,6 +1365,22 @@ static unsigned parse_one_line(const char *buf, unsigned size, line_fn_t *fn, st
 	char line[MAXLINE + 1];
 	int off = 0;
 
+	// Check the first character of a line: an empty line
+	// or a line starting with a TAB is invalid, and likely
+	// due to an early string end quote due to a merge
+	// conflict. Ignore such a line.
+	switch (*p) {
+	case '\n': case '\t':
+		do {
+			if (*p++ == '\n')
+				break;
+		} while (p < end);
+		SSRF_INFO("git storage: Ignoring line '%.*s'", (int)(p-buf-1), buf);
+		return p - buf;
+	default:
+		break;
+	}
+
 	while (p < end) {
 		char c = *p++;
 		if (c == '\n')


torvalds:
This is hopefully the fix for the issue @mwerle reported at

   https://github.com/subsurface/subsurface/issues/3477

and I tried to make the commit message somewhat exhaustive (and exhausting).

It seems to do the right thing.

As noted in the commit message, there is potential for improving things by actually noticing the contiguous comment case, and making the notes actually parse as a combination of the two merged notes. It will be messy to read, but you'd have more data that you could edit down to be better.

This does not do that, because I think it's independent, and the "stale string continuation" case could happen in other places depending on exactly what the two dive files had in common when merging.

I have not tested this extensively, but it seems to work, and makes sense to me.

I probably won't have time to look at the possible extension logic, but I can do at least this before the 5.20 kernel merge window opens for me.

dirkhh:
merged locally

PR #3480, opened 2022-07-31
[TRIVIAL] remove obscure division-assignment operator

bstoeger:
In utc_mkdate() we find the interesting statement
        val = timestamp /= 60;
which not only calculates timestamp / 60, but also overwrites
timestamp with the new value. However, timestamp is never used
in the remainder of the function, because the whole point is to
switch to 32-bit types. Thus, replace the division-assignment
by a simple division operator to avoid head-scratching.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is a completely trivial code-hygiene thing. This won't change anything, but make the code less confusing for future readers.

Please check that I didn't miss something subtle!?

### DIFF
--- a/core/time.c
+++ b/core/time.c
@@ -56,7 +56,7 @@ void utc_mkdate(timestamp_t timestamp, struct tm *tm)
 
 	/* minutes since 1900 */
 	tm->tm_sec = timestamp % 60;
-	val = timestamp /= 60;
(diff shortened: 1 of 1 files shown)


dirkhh:
merged locally

PR #3486, opened 2022-08-13
desktop: improve composition on TagWidgets

bstoeger:
The TagWidgets hook into the textChanged() signal to invoke
the word-completer. However, that signal is also emitted for
composition-keys, making composition impossible if the completer
decides that it should show some entries.

Instead, hook into the inputMethodEvent() function, where one
can test whether a real character was input.

Also, don't hook into cursorPositionChanged(), since that led
to an uncanny cascade of reparse() calls when editing text.

The UI experience is still rough as sometimes the completer
popup steals the focus and hinders further entry.

Also, this doesn't fix the location field, which is its own class.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Try to improve composition in the TagWidgets by not reacting to `textChanged` but to `inputMethodEvent`.

### Changes made:

1) Hook into `inputMethodEvent` instead of `textChanged`.
2) Don't listen to `cursorPositionChanged` signal.
3) Minot code cleanups.
### Related issues:

Fixes part of #1405.

Needs testing! The whole interaction with the completer is ominous.

### DIFF
--- a/desktop-widgets/tagwidget.cpp
+++ b/desktop-widgets/tagwidget.cpp
@@ -7,9 +7,6 @@
 
 TagWidget::TagWidget(QWidget *parent) : GroupedLineEdit(parent), m_completer(NULL), lastFinishedTag(false)
 {
-	connect(this, SIGNAL(cursorPositionChanged()), this, SLOT(reparse()));
-	connect(this, SIGNAL(textChanged()), this, SLOT(reparse()));
-
 	QColor textColor = palette().color(QPalette::Text);
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
 	float h, s, l, a;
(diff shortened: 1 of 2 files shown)


bstoeger:
OMG, this is so horrible. The popup opened by the completer by default does not have the `WA_InputMethodEnabled` set and therefore doesn't get these events. Thus, composition is broken as long as the popup is open.

Set the flag explicitly when starting completion.

@dirkhh: Could you perhaps check with your Qt contacts - I'm nearly certain that this is NOT how it should be done.

dirkhh:
@thiagomacieira who should we poke for this?

dirkhh (approved):
That is... creative.
I have no idea how you managed to figure this out, but lo and behold. I can reproduce the problem, and the fix "fixes" it...

I do, however, agree, that this is not the prettiest of solutions.

dirkhh:
I have tested this with two different Linux versions and it does seem to work reliably. So let's merge and see where this gets us.

bstoeger:
Ooops. That might have been a bit premature. I think we're missing something fundamental. See upcoming PR.

dirkhh:
most of the time I'm way too sluggish to respond. Here I happened to be at my computer, I tested it, so I went ahead...

There is no winning :)

dirkhh:
do you want me to revert it?

PR #3487, opened 2022-08-13
desktop: set composition flag in dive site list

bstoeger:
This is crazy: when view() is called, the dive-site-suggestion
popup (DiveLocationListView) clears its WA_InputMethodEnabled
flag. This makes key composition not work as long as the
popup is open.

Thus, when showing the popup, explicitly set the flag.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes composition in the dive-site field on the main tab. For some reason the dive site list clears its composition enabled flag, when show() is called. Extremely weird, but there must be some good reason. For now, just set it explicitly.

Contains a few coding-style fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: fix key composition in tag widgets and dive site widget
 mobile: allow cloud account deletion (Apple app store requirement)
 
 ---
--- a/desktop-widgets/locationinformation.cpp
+++ b/desktop-widgets/locationinformation.cpp
@@ -37,7 +37,7 @@ LocationInformationWidget::LocationInformationWidget(QWidget *parent) : QGroupBo
 	ui.diveSiteMessage->setText(tr("Dive site management"));
 	ui.diveSiteMessage->addAction(acceptAction);
 
-	connect(ui.geoCodeButton, SIGNAL(clicked()), this, SLOT(reverseGeocode()));
+	connect(ui.geoCodeButton, &QPushButton::clicked, this, &LocationInformationWidget::reverseGeocode);
 	ui.diveSiteCoordinates->installEventFilter(this);
 
 	connect(&diveListNotifier, &DiveListNotifier::diveSiteChanged, this, &LocationInformationWidget::diveSiteChanged);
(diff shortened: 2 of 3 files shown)


dirkhh (changes requested):
please strip the libdivecomputer change from this branch

dirkhh (approved):
This seems otherwise similar to what you did in the previous one. So should I merge this?
Or (given that you are likely asleep) - feel free to merge when you have time

bstoeger:
Ok, I think I found the culprit:
In https://github.com/qt/qtbase/blob/dev/src/widgets/itemviews/qabstractitemview.cpp when current is set to an item that is not "editable" composition is turned off:

```
setAttribute(Qt::WA_InputMethodEnabled, (current.isValid() && (current.flags() & Qt::ItemIsEditable)));
```

This means that all items must be marked as editable.

Usually I quite like QtWidgets (in stark constrast to QtCore or QtQuick!), but this just feels very messy!

PS: I'll be gone for two weeks now.

thiagomacieira:
Following up from #3486 - I have no clue what these flags do. I could guess who to ask for instruction, but can I ask for a concise description of what the problem is?

bstoeger:
Only widgets with the `Qt::WA_InputMethodEnabled` flag get `inputMethodEvent()` calls. These are necessary for composing characters from multiple key strokes, such as the East Asian languages do. However, apparently also some users of European languages use that input method.

The `TagWidget` and `DiveLocationLineEdit` widgets show auto-completers in the form of `QListView`s. To enable editing while the auto-completion popup is shown, the events received by the `QListView` are forwarded to the widget by event filters. However, if the `ItemIsEditable` flag is not set, the `QListView` clears the `Qt::WA_InputMethodEnabled` flag and thus breaks key composition. See the line I quoted above.

Thanks.

bstoeger:
The problem is: how do we prevent the loss of `inputMethodEvent()` calls while the auto-completion popup is opened?

bstoeger:
If we don't get an answer in due course, I suggest merging this. It is surely wrong and ugly, but it does (hopefully) fix a very significant UI problem.

dirkhh (approved):
I think this is good to merge, Berthold?

bstoeger:
Well - up to you. Do you want to wait for an authoritative answer, or just say "whatever, let's move on"?

dirkhh:
hehe... whatever... just move on.

PR #3490, opened 2022-08-17
Feature/map short names

### LINKED ISSUE #2936: Discussion: Add a divesite location string
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [ ] New feature request
- [X] Discussion request

### Issue long description:

Currently, dive sites have a "name", a "location" (which are GPS coordinates), and potentially some reverse-geocoded tags.

##### Issue 1 - choosing dive sites when entering dive information
The issue is that if a dive site is literally given a name (eg, "Manta Point"), then it becomes very difficult when entering dives to choose the correct dive site for a dive (eg, there may be many "Manta Point" divesites in a users' file, each in a different geographical location).

Having the pop-up include the reverse-geocoded tags (if they are available) would alleviate this issue but not remove it as the user has no control over the tags.  There could be several sites with the same name falling into the same set of tags, so the user would need a way to add further disambiguation information without reverting to including location information in the site name itself which is what I've been doing so far:

eg:
* Australia, QLD, Stradbroke  Island, Manta Point
* Phillipines, Malapascua, Manta Point
etc

This approach doesn't "play nicely" when exporting dive (sites) to other services. It also looks "messy" when viewing the dive sites on the map.

##### Issue 2 - search for all dives in a particular location
Currently when filtering, only the dive site name is used, not the geolocation tags.

Even if geolocation tags would be searched, they don't necessarily "play nicely".  An example is the Izu Peninsula south of Tokyo in Japan. I'd like to list all dives / dive sites in Izu, but the reverse geolocation data places the sites into various different prefectures.  My personal naming strategy for dive sites here is "Japan, Izu, <town>, <site>" - but see point above about this not playing nicely.

##### Issue 3 - playing nicely with external dive site / dive log managers
Most external dive site / dive log managers split site information into a name and a location (where location is a human-readable string, not a set of GPS coordinates). Some leave location as a free-form string, others split it into geographic regions (eg, country, state, city, etc).

Currently, Subsurface only exports the "name" part of a divesite; any reverse-geolookup data is not sent.  This means that the primary export (to divelogs.de) requires a lot of manual work to re-enter the dive site location for each dive. And even if geolocation data would be sent, see my comment above about how this is insufficient.

#### Suggested solution
1 - add a "dive site location" field which is a human-readable string, not GPS coordinates.

This could possibly be filled in automatically by the reverse geolocation lookup, but only if it is empty, and it must allow for editing afterwards.

With this:
- dive-site suggestion popup can show both fields to disambiguate between sites with the same/similar name (show coordinates and/or reverse-geocode lookup data if site-location is empty)
- filtering/searching can operate on both of these fields.
- the user can go to as much granularity, or renaming, of the location as they require.
- this approach "plays nicely" with most external dive log / dive site systems.

mwerle:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Adds a new preference option "Map Short Names". If enabled, only the last component of a dive-site name is displayed on the map. Dive site name components are separated using a slash (/) character.

For example:
Japan / Izu Peninsula / Atami / Chinsen-Aft

If "Map Short Names" is enabled, only "Chinsen-Aft" is displayed on the map.  This can help reduce display clutter when there are many dive sites in close proximity to each other.

### Changes made:

1) qt-models/maplocationmodel.cpp : add logic to shorten the dive site name for display purposes
2) (various files) : add a new preferences setting (map_short_names), as well as the various "plumbing" required to update the map when the setting is changed.
3) (various files) : minor code cleanup to remove compiler warnings in the files which were edited as part of this change.

### Related issues:

(Partially) Fixes #2936 

Instead of using the originally-discussed idea of having 2 major components of a dive site (location and dive-site) split by a separator, and using a different separator to split each of those into sub-components, this solution simply uses a single separator (slsh, "/") which is already used by our exporter.

This represents the minimal viable solution for this feature request and should not impact any other software modules or raise any compatibility concerns.

### Release note:

Add preference option to display a shortened Dive Site Name in the Map Widget.

### DIFF
--- a/qt-models/maplocationmodel.cpp
+++ b/qt-models/maplocationmodel.cpp
@@ -10,6 +10,24 @@
 
 #define MIN_DISTANCE_BETWEEN_DIVE_SITES_M 50.0
 
+// MKW If "Map Short Names" preference is set, only return the last component
+// of the full dive site name.
+// Example:
+// Japan/Izu Peninsula/Atami/Chinsen-Aft
+//    Short name: Chinsen-Aft
+static QString siteMapDisplayName(const char *sitename)
+{
+	const char Separator = '/';
+	QString fullname(sitename);
+	QString name = fullname.section(Separator, -1).trimmed();
+
+	if (name.isEmpty()) {
+		name = fullname;
+	}
+	return name;
+}
+
+
 MapLocation::MapLocation(struct dive_site *dsIn, QGeoCoordinate coordIn, QString nameIn, bool selectedIn) :
     divesite(dsIn), coordinate(coordIn), name(nameIn), selected(selectedIn)
 {
@@ -159,7 +177,7 @@ void MapLocationModel::reload(QObject *map)
 		}
 		if (!diveSiteMode && hasSelectedDive(ds) && !m_selectedDs.contains(ds))
 			m_selectedDs.append(ds);
-		QString name(ds->name);
+		QString name = siteMapDisplayName(ds->name);
 		if (!diveSiteMode) {
 			// don't add dive locations with the same name, unless they are
 			// at least MIN_DISTANCE_BETWEEN_DIVE_SITES_M apart
@@ -223,7 +241,7 @@ void MapLocationModel::diveSiteChanged(struct dive_site *ds, int field)
 		}
 		break;
 	case LocationInformationModel::NAME:
-		m_mapLocations[row]->name = ds->name;
+		m_mapLocations[row]->name = siteMapDisplayName(ds->name);
 		break;
 	default:
 		break;
--- a/qt-models/maplocationmodel.cpp
+++ b/qt-models/maplocationmodel.cpp
@@ -3,6 +3,7 @@
 #include "divelocationmodel.h"
 #include "core/divesite.h"
 #include "core/divefilter.h"
+#include "core/settings/qPrefDisplay.h"
 #if !defined(SUBSURFACE_MOBILE) && !defined(SUBSURFACE_DOWNLOADER)
 #include "qt-models/filtermodels.h"
 #include "desktop-widgets/mapwidget.h"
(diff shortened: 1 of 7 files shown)


qt-models/maplocationmodel.cpp
```diff
@@ -10,6 +10,24 @@
...
 #define MIN_DISTANCE_BETWEEN_DIVE_SITES_M 50.0
 
+// MKW If "Map Short Names" preference is set, only return the last component
+// of the full dive site name.
+// Example:
+// Japan/Izu Peninsula/Atami/Chinsen-Aft
+//    Short name: Chinsen-Aft
+static QString siteMapDisplayName(const char *sitename)
+{
+	const char Separator = '/';
```
dirkhh, line 21: we have tried to avoid imposing more syntax or a mandatory taxonomy on the location name.
This fixed separator isn't ideal, but it seems somewhat reasonable. The implied order also isn't ideal, but making this more flexible only makes things more complex, requires more options, and doesn't really significantly improve the user experience, so I think I'm ok with this.

dirkhh (changes requested):
I'm not in love with the change - but I agree that it is a reasonable minimal implementation and that for many people this can be an improvement.
My only request is that the default for the new option is FALSE - i.e. the default behavior isn't changed.

qt-models/maplocationmodel.cpp
mwerle: Hi Dirk, thank you for your comments.

Slash is already the baked-in separator used by the exporter (at least for divelog.de) As is the ordering (from most generic to most detailed). 

Prior to using Subsurface my personal preference was actually the other way around, and I used a different separator.

So these choices were made due to the existing Subsurface implementation.

As for it being optional, I'm sorry, I must have made a mistake, it was my intent that the default is "false". I'll make the relevant changes as soon as I'm back home (enjoying a short motorbike trip with friends for my birthday :)

mwerle:
> My only request is that the default for the new option is FALSE - i.e. the default behavior isn't changed.

So I'm sorry, but as far as I can tell it already defaults to FALSE.  If the setting doesn't exist in Subsurface.conf then the boolean is automatically FALSE. I couldn't find anywhere in the code where settings are explicitly defaulted to a value. It seems the defaults are implicit in the way the UI is parsed for the various settings.

dirkhh (approved):

dirkhh:
I have no idea why I this was turned on in my brief test. Because when I tried again, I got the expected result of it being off by default. Gremlins?

PR #3491, opened 2022-08-20
Some Coverity Fixes

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Some Coverity fixes that hopefully don't break things and don't create new warnings. Plus one leak fix(?). Untested.

### DIFF
--- a/core/import-csv.c
+++ b/core/import-csv.c
@@ -609,7 +609,7 @@ int parse_txt_file(const char *filename, const char *csv, struct dive_table *tab
 		 */
 
 		if (readfile(csv, &memcsv) < 0) {
-			free(dive);
+			free_dive(dive);
 			return report_error(translate("gettextFromC", "Poseidon import failed: unable to read '%s'"), csv);
 		}
 		lineptr = memcsv.buffer;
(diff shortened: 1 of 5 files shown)


core/picture.c
```diff
@@ -123,13 +123,13 @@ static bool dive_check_picture_time(const struct dive *d, timestamp_t timestamp)
  * The caller is responsible for actually adding the picture to the dive.
  * If no appropriate dive was found, no picture is created and NULL is returned.
  */
-struct picture *create_picture(const char *filename, int shift_time, bool match_all, struct dive **dive)
+struct picture *create_picture(const char *filename, offset_t shift_time, bool match_all, struct dive **dive)
```
dirkhh: I really wonder if this is a valid assumption.

One of the "typical" cases with incorrect dates appears to be an incorrect "Unix epoch" date - so January 1970.
So yeah, that's 52 years.

Now, I don't know if the code would be broken in different ways for that case - but I just a few weeks ago stumbled across pictures that claimed to be from Dec 31 1969 (thanks, time zones)...

dirkhh (commented):
I don't know if the code would work - I guess I should try it.
But at least I disagree with the commit message, if nothing else.

bstoeger:
Ok, dropped first commit. I have no time to look closer into that.

dirkhh:
> Ok, dropped first commit. I have no time to look closer into that.

works for me.

"I have no time" is my middle name right now...

(the tooling for the Ubuntu builds stumbles if I do two merges to quickly together... so I'll wait a few minutes and merge then)

PR #3495, opened 2022-08-24
Prefer real data over planned

atdotde:
When a dive has both real dive computers as well as at
least a planned version (which is just another dive
computer with a special name), only use the data from
real dive computers for aggregate values like maxdepth,
dive time, average depth etc in order not to have
imagined data on the dive list, statistics etc.

I think, the default for the statistics should be be also to ignore dives that are only planned (and including them should require the user to check some box) but It was beyond my powers to implement that due to lack of understanding of the statistics constraints. Maybe @bstoeger could help here.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -795,7 +795,7 @@ pressure_t calculate_surface_pressure(const struct dive *dive)
 	pressure_t res;
 	int sum = 0, nr = 0;
 
-	for_each_dc (dive, dc) {
+	for_each_relevant_dc(dive, dc) {
 		if (dc->surface_pressure.mbar) {
 			sum += dc->surface_pressure.mbar;
 			nr++;
@@ -826,7 +826,7 @@ static void fixup_water_salinity(struct dive *dive)
 	struct divecomputer *dc;
 	int sum = 0, nr = 0;
 
-	for_each_dc (dive, dc) {
+	for_each_relevant_dc (dive, dc) {
 		if (dc->salinity) {
 			if (dc->salinity < 500)
 				dc->salinity += FRESHWATER_SALINITY;
@@ -848,12 +848,13 @@ static void fixup_meandepth(struct dive *dive)
 	struct divecomputer *dc;
 	int sum = 0, nr = 0;
 
-	for_each_dc (dive, dc) {
+	for_each_relevant_dc (dive, dc) {
 		if (dc->meandepth.mm) {
 			sum += dc->meandepth.mm;
 			nr++;
 		}
 	}
+
 	if (nr)
 		dive->meandepth.mm = (sum + nr / 2) / nr;
 }
@@ -863,9 +864,9 @@ static void fixup_duration(struct dive *dive)
 	struct divecomputer *dc;
 	duration_t duration = { };
 
-	for_each_dc (dive, dc)
+	for_each_relevant_dc (dive, dc) {
 		duration.seconds = MAX(duration.seconds, dc->duration.seconds);
-
+	}
 	dive->duration.seconds = duration.seconds;
 }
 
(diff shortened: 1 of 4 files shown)


core/dive.c
```diff
@@ -795,10 +795,19 @@ pressure_t calculate_surface_pressure(const struct dive *dive)
 	pressure_t res;
 	int sum = 0, nr = 0;
 
-	for_each_dc (dive, dc) {
-		if (dc->surface_pressure.mbar) {
-			sum += dc->surface_pressure.mbar;
-			nr++;
+	if (has_planned(dive, false)) {
```
dirkhh: I'm sure I'm missing something obvious, so please bear with me... but...
why the if-clause... why not simply do a `for_each_real_dc()` ??

core/dive.c
```diff
@@ -3118,6 +3154,7 @@ struct dive *clone_delete_divecomputer(const struct dive *d, int dc_number)
 	res->id = dive_getUniqID();
 
 	delete_divecomputer(res, dc_number);
+	force_fixup_dive(res);
```
dirkhh, line 3123: I'm curious why this is needed here...

dirkhh (commented):
As I said in the comment below - I must be missing something... but I think this could be greatly simplified...

core/dive.c
atdotde: Otherwise, a dive that has only planned dive computers would end up without maxdepth, total time etc.

core/dive.c
atdotde: This is a bug I discovered on this way: If you delta a dive computer, the maxdepth etc is not recalculated for the divelist.

core/dive.c
atdotde: Or said differently: This PR should only influence dives that have both: real and planned dive computers.

dirkhh: but then... wouldn't it make more sense to have the macro do the right thing? i.e., move that logic into the macro.
If there are only plans, loop over the plans, otherwise, only loop over real dive computers? I just dislike the fact that we have the identical code twice, and that we do need this weird construct multiple times. Seems more error prone.
So I guess I'm suggesting something like
```for_each_relevant_dc()```

Comment (atdotde) [reply core/dive.c]: Yes, I though about this. But I could not figure out how to do this in a macro (include the if statement only in case some condition holds). My preprocessor experience is even smaller than my C experience (I was already quite proud of my new macro). Of course, one could include the first if with and in the condition of the inner if but then this condition would be tested for every dive computer (and the condition even loops over all dive computers as well). Could you suggest a macro that does this? If you can come up with something reasonable, I would be more than happy to use that instead.

Comment (dirkhh) [reply core/dive.c]: Well, the simplest way to do this would be something like this:
```
#define for_each_relevant_dc(_dive, _dc) \
	bool _all_planned = !has_planned(_dive, false); \
	for (_dc = &_dive->dc; _dc; _dc = _dc->next) if (_all_planned || !is_dc_planner(_dc))
```
that does of course leak the bool into the calling name space, which means that you CANNOT use it twice in the same block... but it seems straight forward enough.
Complexity wise, this is the same as your code - it's just easier on my eyes...

Now, if you would like a version that you can use more than once in same block, it gets a bit (ummm...) uglier:
```
// you need the secondary macro invocation in order for __LINE__ to be expanded so the concatenation works
// this way the for_each_relevant_dc macro can be used multiple times within the same name space
#define _UNIQ_NAME(_a) _CONCAT(_a, __LINE__)
#define _CONCAT(_a, _l) _CONCAT_HACK(_a, _l)
#define _CONCAT_HACK(_a, _l) _a ## _l
#define for_each_relevant_dc(_dive, _dc) \
	bool _UNIQ_NAME(_all_planned) = !has_planned(_dive, false); \
	for (_dc = &_dive->dc; _dc; _dc = _dc->next) if (_UNIQ_NAME(_all_planned) || !is_dc_planner(_dc))

```
(and for the people who are pain-curious... yes, the `__LINE__` macro expands to the line number in which `for_each_relevant_dc()` gets called, so this does work, even though it looks mighty unintuitive...)

atdotde: Thanks, I used the first version (if somebody has to use it twice in a row, throw in an extra pair of braces and it will still work. Indeed, this makes it much nicer.

dirkhh: I'm glad you like it!

dirkhh (approved):
let's merge

PR #3500, opened 2022-08-28
profile: implement panning of profile

bstoeger:
When zoomed in, the profile position was moved by hovering with
the mouse. What a horrible user experience. This is especially
useless if we want to implement an interactive profile on mobile.

Instead, let the user start the panning with a mouse click. The
code is somewhat nasty, because the position is given as a
real in the [0,1] range, which represents all possible positions
from completely to the left to completely to the right.

This commit also removes the restriction that the planner handles
can only be moved when fully zoomed out. It is not completely
clear what the implications are. Let's see.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Replace hovering by explicit click-panning when the chart is zoomed in.

Needs intense testing.

In particular this probably breaks the planner, therefore pinning @atdotde.

### Release note:

Done.

### Documentation change:

Don't know, haven't checked.

### Mentions:

@tggm complained about this.

### DIFF
--- a/profile-widget/profilescene.cpp
+++ b/profile-widget/profilescene.cpp
@@ -12,6 +12,7 @@
 #include "core/pref.h"
 #include "core/profile.h"
 #include "core/qthelper.h"	// for decoMode()
+#include "core/subsurface-float.h"
 #include "core/subsurface-string.h"
 #include "core/settings/qPrefDisplay.h"
 #include "qt-models/diveplannermodel.h"
@@ -491,11 +492,9 @@ void ProfileScene::plotDive(const struct dive *dIn, int dcIn, DivePlannerPointsM
 	percentageAxis->updateTicks(animSpeed);
 	animatedAxes.push_back(percentageAxis);
 
-	if (calcMax) {
-		double relStart = (1.0 - 1.0/zoom) * zoomedPosition;
-		double relEnd = relStart + 1.0/zoom;
-		timeAxis->setBounds(round(relStart * maxtime), round(relEnd * maxtime));
-	}
+	double relStart = (1.0 - 1.0/zoom) * zoomedPosition;
+	double relEnd = relStart + 1.0/zoom;
+	timeAxis->setBounds(round(relStart * maxtime), round(relEnd * maxtime));
 
 	// Find first and last plotInfo entry
 	int firstSecond = lrint(timeAxis->minimum());
@@ -626,4 +625,20 @@ void ProfileScene::draw(QPainter *painter, const QRect &pos,
 		}
 	}
 	painter->drawImage(pos, image);
+}
+
+// Calculate the new zoom position when the mouse is dragged by delta.
+// This is annoyingly complex, because the zoom position is given as
+// a real between 0 and 1.
+double ProfileScene::calcZoomPosition(double zoom, double originalPos, double delta)
+{
+	double factor = 1.0 - 1.0/zoom;
+	if (nearly_0(factor))
+		return 0.0;
+	double relStart = factor * originalPos;
+	double start = relStart * maxtime;
+	double newStart = start + timeAxis->deltaToValue(delta);
+	double newRelStart = newStart / maxtime;
+	double newPos = newRelStart / factor;
+	return std::clamp(newPos, 0.0, 1.0);
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+profile: implement panning of the profile
+planner: allow handle manipulation in zoomed in state
 divelist: do not include planned versions of a dive if there is real data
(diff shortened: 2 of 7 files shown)


profile-widget/profilescene.cpp
```diff
@@ -627,3 +625,19 @@ void ProfileScene::draw(QPainter *painter, const QRect &pos,
 	}
 	painter->drawImage(pos, image);
 }
+
+// Calculate the new zoom position when the mouse is dragged by delta.
+// This is annoyingly complex, because the zoom position is give as
```
dirkhh: given

profile-widget/profilescene.cpp
```diff
@@ -491,11 +491,9 @@ void ProfileScene::plotDive(const struct dive *dIn, int dcIn, DivePlannerPointsM
...
 	animatedAxes.push_back(percentageAxis);
 
-	if (calcMax) {
-		double relStart = (1.0 - 1.0/zoom) * zoomedPosition;
-		double relEnd = relStart + 1.0/zoom;
-		timeAxis->setBounds(round(relStart * maxtime), round(relEnd * maxtime));
-	}
+	double relStart = (1.0 - 1.0/zoom) * zoomedPosition;
+	double relEnd = relStart + 1.0/zoom;
+	timeAxis->setBounds(round(relStart * maxtime), round(relEnd * maxtime));
```
dirkhh, line 497: I'm curious why this was done only in the calcMax case before.

profile-widget/profilescene.cpp
```diff
@@ -627,3 +625,19 @@ void ProfileScene::draw(QPainter *painter, const QRect &pos,
...
 }
+
+// Calculate the new zoom position when the mouse is dragged by delta.
+// This is annoyingly complex, because the zoom position is give as
+// a real between 0 and 1.
+double ProfileScene::calcZoomPosition(double zoom, double originalPos, double delta)
+{
+	double factor = 1.0 - 1.0/zoom;
+	if (fabs(factor) < 1e-10)
+		return 0.0;
```
dirkhh, line 637: please use `IS_FP_SAME(factor, 0.0)`
For reasons I cannot comprehend that is now in `subsurface_string.h`

dirkhh (commented):
IIRC the planner / zoom interaction was absolutely unusable because trying to drag things when the moving of the mouse also panned the profile was completely useless.

profile-widget/profilescene.cpp
bstoeger: > please use `IS_FP_SAME(factor, 0.0)`

This macro invocation is silly. It expands to `fabs(factor) <= 0.000001 * fabs(factor)`, which is never `true` for `factor` unequal 0.0 unless you reach resolution limits. This seems like a very bad idea to me given the different precision of standard floating point registers and SIMD registers.

bstoeger: Actually - I don't see how this can ever be true for `factor` which is not precisely 0.0, which makes this whole thing pretty pointless.

dirkhh: We've had this code since 2014 - introduced in 2f822ec0dcd2d715bf4234023b510aed9fe90f0b by @atdotde 

Since this is the type of thing that I prefer to leave to the experts - can the two of you please figure out what a reasonable way is to compare floating point values?

bstoeger: The problem was introduced in b6a30dcdd3225e67e7eed7ead68c919c36243ffc. This fails when one of the arguments is 0. Note that there are other occurrences that have to be fixed.

profile-widget/profilescene.cpp
bstoeger: Whatever the reason - it is a bug, since it means that the full profile is shown when dragging dive handles.

profile-widget/profilescene.cpp
dirkhh: well - I don't want to change the macro without hearing Robert, so let's just leave your code as is and move on.
I still need to find time to test it...

dirkhh:
this looks good to me.
anything in particular you want us to test?

bstoeger:
> anything in particular you want us to test?

Well, I guess the scrolling itself and the interaction with the planner.

dirkhh:
I never use the planner... I'll play with it anyway...

dirkhh:
eine Sache, die mir im Code Review nicht klar wurde... das Zoom hier ist ausschliesslich in der Zeit Achse, aber nicht in der Tiefenachse. Ich weiss nicht, ob das _falsch_ ist, aber es ist zumindest _anders_...

dirkhh:
I merged this locally because of GitHub's challenges when trying to edit merge "conflicts"...

PR #3503, opened 2022-08-29
Ratio BLE detection fix

dirkhh:
For Ratio dive computers we can't tell by the Bluetooth name which model it is.
There are BT only models and BLE only models. The failure case here was a user
on iOS (BLE only) with a BLE only dive computer which we didn't recognize
because previously we returned a BT only device (which isn't supported on an
iPhone), and the lookup won't return a valid descriptor if the transport needed
isn't available.

These days BLE is far more common, so return a BLE enabled name by default, but
try a BT only name just in case.

- [x] Bug fix
- [x] Functional change

### Release note:

added

### Documentation change:

not needed

### Mentions:

@bstoeger just so someone else reads the code and I'm not doing something obviously stupid here...


bstoeger (commented):
Doesn't look wrong. 🤷

PR #3504, opened 2022-08-29
iOS email attachments

dirkhh:

- [x] Functional change

This includes a few cleanups for our iOS builds, but the main focus is the objC++ code to natively send email.
Now we can send support emails with the log files as attachments - just like we've been doing on Android for a while.

objC++ is yet another programming language that I DO NOT SPEAK but try to write code in. So this code is likely horrible (and it is to some extend borrowed from a discussion in the Qt Forums). But the good news is that it appears to work in my tests on an iPhone and an iPad.

### Changes made:

- create a helper class in C++ and instantiate when the QMLManager is created
- create an ObjC++ class that talks to the Controller View and deals with sending an email

### Release note:

done

### Mentions:

@bstoeger @atdotde  I don't think you know ObjC++ any more than I do... but just in case

### DIFF
--- a/ios/ios-share.h
+++ b/ios/ios-share.h
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef IOSSHARE_H
+#define IOSSHARE_H
+
+// onlt Qt headers and data structures allowed here
+#include <QString>
+
+class IosShare {
+public:
+	IosShare();
+	~IosShare();
+	void supportEmail(const QString &firstPath, const QString &secondPath);
+	void shareViaEmail(const QString &subject, const QString &recipient, const QString &body, const QString &firstPath, const QString &secondPath);
+private:
+	void *self;
+};
+
+#endif /* IOSSHARE_H */
--- a/ios/ios-share.mm
+++ b/ios/ios-share.mm
@@ -1,0 +1,114 @@
+// SPDX-License-Identifier: GPL-2.0
+//
+// this code was inspired by the discussions in
+// https://forum.qt.io/topic/88297/native-objective-c-calls-from-cpp-qt-ios-email-call
+
+// this include file only has the C++/Qt headers that can be used from C++
+#include "ios-share.h"
+
+// these are the required ObjC++ headers
+#import <Foundation/Foundation.h>
+#import <Foundation/NSString.h>
+#import <UIKit/UIKit.h>
+#import <MessageUI/MessageUI.h>
+
+// declare an ObjC++ class that will interact with the mail controller
+// that second member that is called when the mail app is finished is critical for this to work
+@interface IosShareObject : UIViewController  <MFMailComposeViewControllerDelegate>
+{
+}
+- (void)shareViaEmail:(const QString &) subject :(const QString &) recipient :(const QString &) body :(const QString &) firstPath :(const QString &) secondPath;
+- (void)mailComposeController:(MFMailComposeViewController *)controller didFinishWithResult:(MFMailComposeResult)result error:(nullable NSError *)error;
+@end
+
+@implementation IosShareObject
+// first, inside the implementation of the ObjC++ class, implement the Qt class
+IosShare::IosShare() : self(NULL) {
+	// call init to ensure that the ObjC++ object is instantiated, which in return
+	// apparently sets up the Controller
+	self = [ [IosShareObject alloc] init];
+}
+
+IosShare::~IosShare() {
+	[(id)self dealloc];
+}
+
+// simplified method that fills subject, recipient, and body for support emails
+void IosShare::supportEmail(const QString &firstPath, const QString &secondPath) {
+	QString subject("Subsurface-mobile support request");
+	QString recipient("in-app-support@subsurface-divelog.org");
+	QString body("Please describe your issue here and keep the attached logs.\n\n\n\n");
+	shareViaEmail(subject, recipient, body, firstPath, secondPath);
+}
+
+void IosShare::shareViaEmail(const QString &subject, const QString &recipient, const QString &body, const QString &firstPath, const QString &secondPath) {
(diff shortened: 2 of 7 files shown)


ios/ios-share.h
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef IOSSHARE_H
+#define IOSSHARE_H
+
+// onlt Qt headers and data structures allowed here
+#include <QString>
+#include <QObject>
+
+class IosShare : QObject {
```
bstoeger: Why must this be a QObject? It uses neither signals nor slots.

ios/ios-share.mm
```diff
@@ -0,0 +1,118 @@
...
+- (void)shareViaEmail:(const QString &) subject :(const QString &) recipient :(const QString &) body :(const QString &) firstPath :(const QString &) secondPath;
+- (void)mailComposeController:(MFMailComposeViewController *)controller didFinishWithResult:(MFMailComposeResult)result error:(nullable NSError *)error;
+@end
+
+@implementation IosShareObject
+// first, inside the implementation of the ObjC++ class, implement the Qt class
+IosShare::IosShare(QObject *parent) : QObject(parent), self(NULL) {
+	// call init to ensure that the ObjC++ object is instantiated, which in return
+	// apparently sets up the Controller
+	init();
```
bstoeger: What's the point of an separate init() function? The constructor is supposed to construct the object, so it is perfectly idiomatic to do initialization work in the constructor.

ios/ios-share.h
```diff
@@ -0,0 +1,20 @@
...
+public:
+	IosShare(QObject *parent);
+	~IosShare(void);
```
bstoeger: Declaring functions with `(void)` works for compatibility reasons, but is actually very weird in C++. This comes from the bad old times of K&R C where you didn't declare full function signatures.`int f();` meant `f` is a function returning int, but you don't say which arguments it has, because that would let the compiler help you and we don't want that (or rather we couldn't do that in 32KB of RAM). So later they had to introduce `int f(void)` to explicitly say that the function takes no arguments. However, in C++ you *must* declare functions, because the compiler must know whether arguments are passed by value or by reference [e.g. `int f(int i)` vs. `int f(int &i)`]. Thus, `int f()` became the idiomatic (and logical!) way of declaring a function without arguments.

mobile-widgets/qmlmanager.cpp
```diff
@@ -242,6 +241,8 @@ QMLManager::QMLManager() :
 	// on iOS we should save the data to the DocumentsLocation so it becomes accessible to the user
 	QStringList fileLocations =
 		QStandardPaths::standardLocations(QStandardPaths::DocumentsLocation);
+	// create the object that we use to send emails
+	iosshare = new IosShare(this);
```
bstoeger: What's the point of allocating this? Why not make it a simple sub-object of QMLManager? Can it only be allocated later?

If you really want to allocate it, make it a `unique_ptr<IosShare> iosshare` and allocate with `iosshare = std::make_unique<IosShare>();`. Thus, it will be automatically deleted when it goes out of scope (e.g. in the destructor).

mobile-widgets/qmlmanager.cpp
```diff
@@ -598,6 +608,9 @@ QMLManager::~QMLManager()
 #if defined(Q_OS_ANDROID) || defined(Q_OS_IOS)
 	if (appLogFileOpen)
 		appLogFile.close();
+#endif
+#if defined (Q_OS_IOS)
+	delete iosshare;
```
bstoeger: Not needed if using a subobject or a unique_ptr.

bstoeger (commented):
I had to use ObjC once ca. 25 years ago and even then I felt that it was a completely ridiculous Frankenstein's monster thing. Sorry, I cannot help you there. The C++ part of this PR is maybe a bit clumsy, so a few comments/questions/suggestions on that.

ios/ios-share.h
dirkhh: it doesn't. good catch. somewhere between attempt 15 and 31 I was trying to follow some advice about ensuring that there was a valid window associated with the object... that didn't work, but making this class derived from QObject was part of playing with this. Removed.

ios/ios-share.mm
dirkhh: Similar but different to the last response - this was part of trying to get the C++ object to correctly initialize and create the ObjC++ stuff that we need. Let me remove this and try again and see if this is just a leftover from something that I experimented with, or if this is part of the oddities that were actually needed to get this to work 🤷🏼‍♂️

Edited to add: ... and indeed, this, too, is just leftover nonsense.
Good Catch Number Two!

ios/ios-share.h
dirkhh: cut and paste from an example 😇
fixed.

mobile-widgets/qmlmanager.cpp
dirkhh: First I allocated it the support email creation function (and that's where it couldn't be a simple sub-object because it would be deleted before we were done with it) - but that caused us to have multiple instances of it if you tried to send email more than once and that caused Bad Things to happen. So I moved this in the constructor. But I guess here it makes more sense to just use a sub-object...
Simpler, and less code.
Good Catch Number Three

mobile-widgets/qmlmanager.cpp
dirkhh: nor when just using a sub-object

dirkhh:
Thanks for an excellent set of recommendations. This is why I love getting my code reviewed.
When I bang my head against something for a week or more and finally have something that works, it's hard to then go back and try which of the odd things I tried in the process were critical to get it working, and which were failed attempts and things that need to be re-evaluated.
Much appreciated.

dirkhh:
I think I'll merge as is - I want to use this as the basis for some additional work to finally get the people who hate the cloud off my back...

PR #3505, opened 2022-08-30
Floating point comparisons

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Some cleanups concerning the floating point comparisons. Notably, comparisons with 0 were broken. Generally, I think it is wrong to assume that generic absolute or relative thresholds always work and each case should be considered by itself.

### Changes made:

1) Create a new function for 0-comparisons.
2) Turn macro into inline function.
3) Move into new floating point header.

### Release note:

Not needed.
### Documentation change:

Not needed.


dirkhh (approved):
that seems great to me. I don't know if Robert is around - he hasn't responded to any of the pings, so I'd be willing to merge this once the builds pass.

bstoeger:
Note: completely untested!

dirkhh:
> Note: completely untested!

Well, some of this should be exercised by our tests here in the CI/CD, but I am well aware that this is not nearly enough.
Staring at the code I cannot see how that could cause issues, but that's not the same thing as rigorous testing...

atdotde:
Thumbs up. Sorry for not properly reacting: I am on a camp ground in southern France with extremely limited internet.

PR #3510, opened 2022-09-03
Use combobox for sensor move

michaelandreen:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

This changes the text field that was used for moving sensors between cylinders into a combobox with drop down of the cylinders with attached sensors.

It also lets you swap the sensor on two cylinders without the need for adding a third cylinder as staging area.


desktop-widgets/modeldelegates.cpp
```diff
@@ -281,6 +282,42 @@ void TankUseDelegate::setModelData(QWidget *editor, QAbstractItemModel *model, c
...
+}
+
+QWidget *SensorDelegate::createEditor(QWidget *parent, const QStyleOptionViewItem &, const QModelIndex &index) const
+{
+	QComboBox *comboBox = new QComboBox(parent);
+	std::vector<int16_t> sensors;
+	const struct divecomputer *currentdc = get_dive_dc(current_dive, dc_number);
+	for (int i = 0; i < currentdc->samples; ++i) {
+		auto &sample = currentdc->sample[i];
+		for (auto s = 0; s < MAX_SENSORS; ++s) {
```
dirkhh: consistency question  (I know, idle nit picking)... why an `int` in L295, but an `auto` here?

commands/command_edit.cpp
```diff
@@ -1361,6 +1361,9 @@ void EditSensors::mapSensors(int toCyl, int fromCyl)
 		for (int s = 0; s < MAX_SENSORS; ++s) {
 			if (dc->sample[i].pressure[s].mbar && dc->sample[i].sensor[s] == fromCyl)
 				dc->sample[i].sensor[s] = toCyl;
+			// In case the cylinder we are moving to has a sensor attached, move it to the other cylinder
+			else if (dc->sample[i].pressure[s].mbar && dc->sample[i].sensor[s] == toCyl)
+				dc->sample[i].sensor[s] = fromCyl;
```
dirkhh, line 1366: so if both to and from cylinders had a sensor, then they get swapped, but if no sensor was attached to to cylinder before, we still have the old behavior.
Very clever.

dirkhh (commented):
Reading the code I like it.
I need to compile it and play with it (and remind myself how to create a dive that actually can be used to test this... I never dive with more than one sensor)

qt-models/cylindermodel.cpp
```diff
@@ -240,23 +240,17 @@ QVariant CylindersModel::data(const QModelIndex &index, int role) const
...
-							sensors.push_back(sample.sensor[s]);
+							return QString::number(sample.sensor[s]);
 					}
 				}
 			}
-			QStringList sensorStrings;
-			for (auto s : sensors)
-				sensorStrings << QString::number(s);
-			return tr("Select one of these cylinders: ") + sensorStrings.join(",");
+			return "";
```
bstoeger: I would suggest `return QString();` here.

desktop-widgets/modeldelegates.cpp
michaelandreen: Since this is mostly taken from CylindersModel, I think the inner loop went through a few different iterations in my original sensor patch before I understood the sensor data model.

qt-models/cylindermodel.cpp
michaelandreen: Done

commands/command_edit.cpp
michaelandreen: Yea, I remember that I thought the swap case felt really messy and then it turned out to be simpler than making the cylinders with sensors non-editable. =)

dirkhh (approved):
Seems to work fine in my extremely limited testing...

Issue #3501, opened 2022-08-29
No details in information box about surface event ("yellow triangle sign") and event triangle not aligned properly


sdroege:
### Describe the issue:
- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I have a dive log from a Suunto Zoop Novo with the following content here

```xml
  <event time='0:00 min' type='25' flags='1' name='gaschange' cylinder='0' />
  <event time='0:00 min' type='9' name='surface' />
  <event time='41:05 min' type='9' name='surface' />
  <event time='42:33 min' type='9' name='surface' />
  [...]
  <sample time='41:00 min' depth='1.48 m' />
  <sample time='41:20 min' depth='1.72 m' />
  <sample time='41:40 min' depth='2.49 m' />
  <sample time='42:00 min' depth='2.34 m' />
  <sample time='42:20 min' depth='1.83 m' />
```

The second surface event is shown in the dive profile with a yellow triangle sign. According to the manual, hovering over the sign should give some details about the event in the information box but that doesn't seem to work in this case, see screenshot below:

![image](https://user-images.githubusercontent.com/301846/187229158-8891e5ae-ec7a-4e76-a1e0-a502b65d9071.png)

The cursor was centered on the yellow triangle at this point. Note also that the triangle is centered at 41:40 instead of 41:05. If going to 41:05 (cursor quite a bit left of the yellow triangle), still nothing is shown:

![image](https://user-images.githubusercontent.com/301846/187229499-c0c4c7e8-2bc4-4570-876c-866f2a57b7db.png)

### Operating system:

Debian sid using flatpak 1.12.7

### Subsurface version:

v5.0.9 installed via flatpak, but previous versions had the same behaviour

### Steps to reproduce:

Create a dive log as above with multiple surface events. I tried manually adding it to another dive log like this and it had the same effect.

### Current behavior:

The event yellow triangle sign is visible as seen above but at the wrong position and hovering over it does not show any additional information in the information box (at the triangle position as well as the actual time of the event).

### Expected behavior:

The information box would show that this is a surface event.

jefdriesen:
I'm not sure what kind of info you expect to see here. The info box clearly shows "surface" in your first screenshot (last line).

sdroege:
Ah, I guess I just want better glasses :) Somehow I completely missed that next to the other lines in the info box. Sorry

[...]

PR #3520, opened 2022-09-12
For warnings, show icon in infobox

atdotde:
Render a warning sign in front of the event string in the infobox. This is done in rich text.

Notes: 
1) This shows the warning sign for all events, not just warnings. Doing otherwise would require much more change in logic (as the profile icons are pixmaps that have a complicated selection logic for different types of events while here a file name is requested)
2) It adds some vertical space before the warning

Is this good enough to be an improvement over what we have so far?

This was suggested in #3501.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+infobox: show an icon for warnings
 import: allow import of divesites without UUID
 profile: implement panning of the profile
 planner: allow handle manipulation in zoomed in state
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
This seems like a reasonable implementation.

PR #3521, opened 2022-09-12
Sensor index fixup

torvalds:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes up the pressure sensor index code to not be quite so eager to get rid of pressures that it doesn't understand.

In particular, if for some reason a dive computer was saved with a pressure sensor index bigger than the number of cylinders, it now tries to fix up the situation by renumbering the indexes to a minimal set. And if even that doesn't then fit in the number of cylinders, we just accept the sensor data as-is rather than clear it.

The user can then add cylinders as needed, or whatever.

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1240,11 +1240,11 @@ static void fixup_dive_dc(struct dive *dive, struct divecomputer *dc)
 	/* Fix up gas switch events */
 	fixup_dc_gasswitch(dive, dc);
 
-	/* Fix up cylinder pressures based on DC info */
-	fixup_dive_pressures(dive, dc);
-
 	/* Fix up cylinder ids in pressure sensors */
 	fixup_dc_sample_sensors(dc, dive->cylinders.nr);
+
+	/* Fix up cylinder pressures based on DC info */
+	fixup_dive_pressures(dive, dc);
 
 	fixup_dc_events(dc);
 
--- a/core/dive.c
+++ b/core/dive.c
@@ -1210,16 +1210,40 @@ static void fixup_no_o2sensors(struct divecomputer *dc)
 	}
 }
 
-static void fixup_dc_sample_sensors(struct divecomputer *dc, int nr_cylinders)
+static void fixup_dc_sample_sensors(struct dive *dive, struct divecomputer *dc)
 {
+	unsigned long sensor_mask = 0;
+
 	for (int i = 0; i < dc->samples; i++) {
 		struct sample *s = dc->sample + i;
 		for (int j = 0; j < MAX_SENSORS; j++) {
-			if (s->sensor[j] < 0 || s->sensor[j] >= nr_cylinders) {
+			int sensor = s->sensor[j];
+
+			// No invalid sensor ID's, please
+			if (sensor < 0 || sensor > MAX_SENSORS) {
 				s->sensor[j] = NO_SENSOR;
 				s->pressure[j].mbar = 0;
+				continue;
 			}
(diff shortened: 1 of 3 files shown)


torvalds:
Hmm. I have no idea why that "TestRenumber" fails on Ubuntu but not on Fedora (and not on my local Fedora/arm64 build).

Strange.

torvalds:
Ahh. We have code that depends on the sensor index never being bigger than the number of cylinders.

Ugly, ugly.

3kh0 (approved):
Looking good!

dirkhh:
So, Linus, what's the status of this one?

torvalds:
Ok, looking at it. That test-case ends up with insane samples that have "s->sensor[x]" value of 255.

And the reason turns out to be that @bstoeger commit 361678dcbe introduced sanitize_sensor_id(), which does

        return d && nr >= 0 && nr < d->cylinders.nr ? nr : NO_SENSOR;

but the type of that function is 'uint8_t'.  So it doesn't actually return NO_SENSOR (-1) like the comment says, but 255.

We ended up hiding that by then later doing "if we have a sensor index that is larger than number of cylinders, clear it", so this bug was hidden by that whole thing. And when I removed the clearing, that just uncovered this odd issue.

I'm not sure why that test-case triggers this odd problem (and nothing else does), and I'll have to figure out what the proper solution is, but at least I see why it happens.

torvalds:
Oh, and that sanitize_sensor_id() was actually correct when introduced, but then @bstoeger commit 0c84f369c35b ("core: use int16_t for sensor-id") changed the type for the sensor index, but didn't fix this helper.

torvalds:
Ok, I have force-pushed an updated version of this branch.

It takes a different approach to the sensor index thing - instead of renumbering sensors to fit in the cylinders, it will just add empty cylinders as required to make the sensor data fit. 

I verified with your Palau 2016 data that this simply creates a new cylinder for those dives with the Atomic Aquatics dive computer, and it now looks like you have two different cylinders when you look at those dives. So now if you want to fix that up, you'd have to do the sensor remapping first, and then delete the second cylinder.

That was not only the simpler fix, but also the safer one. We don't do any "let's just automatically renumber sensors", and we don't throw any data away.

Let's see if it passes the test-suite now.

torvalds:
> Let's see if it passes the test-suite now.

So far so good, but it might be worth pointing out that this pull request has three commits, and two of them are basically just preparatory fixes that are only indirectly related. 

All the *meat* of the pull request is in that last commit, with the two first ones just fixing that sensor type confusion, and fixing up the order of the dive fixups.

It all looks "ObviouslyCorrect(tm)" to me, but hey, so did the previous version. It just happened to uncover preexisting issues that were hidden by the sensor clearing. This one feels much safer, but who knows what it might uncover.

dirkhh (approved):
I read the code. I ran it against my dive data (forcing saves so I see what it does).
It seems correct, and I like it MUCH better than the previous version. This seems logical and consistent.
I requested a review from @bstoeger as well as it touches his code :)

bstoeger (approved):
Not sure my comment is needed here.

The `uint` -> `int` change fixes an obvious oversight (and I find it slightly annoying that we didn't get a compiler warning here). As for the sensor thing I still don't understand the data structure and will never use this part of the code.

So fine with me.

PR #3523, opened 2022-09-17
profile: don't use displayed_dive to check whether dive changed

bstoeger:
The profile replots if the mode of the currently displayed dive changed. To do so, it compares the changed dive to the displayed_dive. However, that is only used for planned dives since quite some time.

Fix the check and make the replotting work again.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The profile was still using `displayed_dive` to check whether the currently displayed dive was changed. However, `displayed_dive` hasn't been updated for a long time, expect when in the planner. Fix this check.

### Release note:

It *is* a user-visible bug fix, but probably too minor to warrant a changelog entry.


dirkhh (approved):
that seems better :)

PR #3524, opened 2022-09-18
desktop: disable extradata tab if no site selected

bstoeger:
Owing to bit-rot, the extradata tab was not disabled if no dive is selected. The reason was that there was an additional tab (dive-computers) that should not be disabled. However that tab was removed and now the extradata tab was not disabled.

Fix this, but note that there is another of these 'parasitic' tabs, that should be removed, namely the dive-site tab.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A rather trivial cosmetic fix: disable the extradata tab when no dive is selected. This however points out the problem of the dive site tab, which should also be removed.

### Release note:

Not worth it.

### DIFF
--- a/desktop-widgets/tab-widgets/maintab.cpp
+++ b/desktop-widgets/tab-widgets/maintab.cpp
@@ -107,11 +107,11 @@ void MainTab::updateDiveInfo()
 	if (DivePlannerPointsModel::instance()->isPlanner())
 		return;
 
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
that seems obviously correct.

PR #3526, opened 2022-09-18
desktop: cache isGnome3Session result

bstoeger:
On gtk based distributions, isGnome3Session() tests for Gnome3 by executing a "pidof" command. This is called everytime a combobox is created in the equipment tab. This seems a bit excessive, so let's cache the result.

Use a lambda to calculate the flag to avoid unnecessary #ifdef-ery or complex control flow.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

On GTK based distributions a command is executed everytime a combobox in the equipment tab is opened. This seems excessive. Cache the result instead.

Totally untested, since I don't have a GTK based distribution.

### Release note:

Not necessary.

### DIFF
--- a/desktop-widgets/modeldelegates.cpp
+++ b/desktop-widgets/modeldelegates.cpp
@@ -127,15 +127,6 @@ QWidget *ComboBoxDelegate::createEditor(QWidget *parent, const QStyleOptionViewI
 	currCombo.model = const_cast<QAbstractItemModel *>(index.model());
 	currCombo.activeText = currCombo.model->data(index).toString();
 
-	// Current display of things on Gnome3 looks like shit, so
-	// let's fix that.
-	if (isGnome3Session()) {
-		QPalette p;
-		p.setColor(QPalette::Window, QColor(Qt::white));
-		p.setColor(QPalette::Base, QColor(Qt::white));
-		comboDelegate->lineEdit()->setPalette(p);
-		comboDelegate->setPalette(p);
-	}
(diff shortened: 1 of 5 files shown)


dirkhh (commented):
I wouldn't have implemented it like this in a hundred years - but of course once I read this, it's intriguingly elegant... 🤷🏼‍♂️
I have a VM with Gnome based distro and can try this... I just returned from two weeks of travel, so there are quite a few things in my backlog, this may not happen right away, but in concept this looks good

dirkhh:
Hmm... I wonder if this has worked in the last... I don't know... decade. And what it actually does...
So on a Fedora system running Gnome, your style (first check in the lambda) isn't gtk3
Therefore, at least on a Fedora system this doesn't appear to work... and I don't know where else you'd be using the gtk3+ style, TBH...
Looking back to the `git blame` of this we end up with e81bbc1dabc36 which makes me think that the correct fix would be to simply remove this check. It's extremely unlikely that this is still relevant today, nine years later...

bstoeger:
Ok - then let's remove it.

dirkhh (approved):
That's reasonable.
We should poke people who use this on Linux/Gnome to double check that this doesn't break some odd case that I didn't see - but in all my googling I didn't find an indication that anyone was using a style called `gtk+`, so that first check would always fail...

PR #3528, opened 2022-09-30
build-system/windows: don't build the libmtp examples

dirkhh:
We had that fail in the GitHub Action - but really, why would we build those in the first place.

- [x] Build system change

no idea if this will work


bstoeger:
> no idea if this will work

It seems to work, therefore commit to get rid of the annoying build failures.

dirkhh:
Too funny -- I was just about to merge this when the screen changed under my mouse :)

bstoeger:
It happens. 🤷

PR #3531, opened 2022-10-05
core: better commit message in git save

dirkhh:
This rarely gets seen / looked at, but it can help make it easier to understand what a user was doing when trying to restore dives that were inadvertantly deleted.

- [x] Functional change

This is a bit of a self serving change as most likely I'm the main person ever seeing this.
When we save to git storage, we put a summary of changes made into the commit message. And that summary was completely bogus in the past as it always showed the changes from the start of this session (not the last save) to the current point - which is completely misleading if the user made some changes, saved them, then undid them and saved again.
With this there is a better correlation between the commit message and what's in the commit.

### Mentions:

@bstoeger just a quick sanity check, maybe?

### DIFF
--- a/commands/command_base.cpp
+++ b/commands/command_base.cpp
@@ -86,11 +86,28 @@ QString getListOfDives(QVector<struct dive *> dives)
 
 // return a string that can be used for the commit message and should list the changes that
 // were made to the dive list
+// keep in mind that the changes could have been a number of undo commands, so we might have
+// to go backwards from the last one written; this isn't perfect as a user could undo a command
+// and then do something else instead of redoing that undo - the undo information is then lost
+// for the changelog -- but of course the git history will show what happened
 QString changesMade()
 {
+	static int nextToWrite = 0;
+	int curIdx = undoStack->index();
 	QString changeTexts;
-	for (int i = 0; i < undoStack->index(); i++)
-		changeTexts += undoStack->text(i) + "\n";
+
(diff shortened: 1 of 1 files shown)


commands/command_base.cpp
```diff
@@ -86,11 +86,28 @@ QString getListOfDives(QVector<struct dive *> dives)
...
+	if (curIdx > nextToWrite) {
+		for (int i = nextToWrite; i < curIdx; i++)
+			changeTexts += undoStack->text(i) + "\n";
+	} else if (curIdx < nextToWrite) { // we walked back undoing things
+		for (int i = nextToWrite - 1; i >= curIdx; i--)
+			changeTexts += "(undo) " + undoStack->text(i) + "\n";
+	} else { // so this means we undid something (or more than one thing) and then did something else
+		 // so we lost the information of what was undone - and how many things were changed;
+		 // just show the last change
+		changeTexts += undoStack->text(curIdx - 1) + "\n";
```
bstoeger, line 108: Perhaps check for `curIdx > 0` here? Not sure whether this can be called if `curIdx == nextToWrite == 0`?

bstoeger (approved):
I don't see anything fundamentally wrong apart from the point that you note yourself: when the user does undo/redo, a part of the history will be lost.

dirkhh:
merging manually after adjusting to the suggestion above.

PR #3536, opened 2022-10-17
desktop: make time shift dialog show correct times

### LINKED ISSUE #3535: Shift Dive Times show bad time data
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
The "before" and "shifted" time info shown when using "Shift Dive times" is displayed as time 0:0 (1-Jan-1970). The offset calculation is correct (at least relative to 1-Jan-1970). When applied, the correct time offset *is* applied to the dive. This looks slightly alarming but is actually cosmetic.

### Operating system:
MacOS Monterey 12.6 , MBP M1

### Subsurface version:
5.0.10 official build, (both standard and Qt6 builds)

### Steps to reproduce:
1) Select a dive in a dive log
2) right click, choose "Shift dive times"
3) popup window shows both current and shifted dive start time as "0:0"
4) enter an offset -> shifted time changes to offset based on 1-Jan-1970
5) save offset - new dive start time in log is correct (phew!)

### Current behavior:
Subsurface does not display correct dive start time in the popup window, always 0:0

### Expected behavior:
Actual dive start time should be the basis for the time offset.

### Additional information:
n/a
### Mentions:
n/a

dirkhh:
This got broken in commit 7417f865cd ("cleanup: un-singletonize ShiftTimesDialog") and no one ever noticed.

We need to intitialize the when variable and set up the initial texts in order for the time shift dialog to show the correct information. Doing this in the ButtonClicked member (right before the dialog is destroyed) makes absolutely no sense.

Fixes #3535

- [x] Bug fix

this has been broken for 2.5 years. Clearly a frequently used function?

### Changes made:

initialize both the time text and the `when` member variable to ensure that what is displayed makes sense
remove the pointless code updating the dialog text right before it gets destroyed

### Related issues:

Fixes #3535

### Release note:

done

### Documentation change:

not needed

### Mentions:

@bstoeger

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,4 @@
-mobile: implement paging through dive computers
-mobile: send log files as attachments for support emails on iOS
-mobile: allow cloud account deletion (Apple app store requirement)
-mobile: fix listing of local cloud cache directories
+desktop: fix dive time display in time shift dialog
 
 ---
 * Always add new entries at the very top of this file above other existing entries and this note.
--- a/desktop-widgets/simplewidgets.cpp
+++ b/desktop-widgets/simplewidgets.cpp
@@ -89,13 +89,6 @@ void ShiftTimesDialog::buttonClicked(QAbstractButton *button)
 		if (amount != 0)
 			Command::shiftTime(getDiveSelection(), amount);
 	}
-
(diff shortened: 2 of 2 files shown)


bstoeger (approved):
lgtm

PR #3542, opened 2022-10-20
planner: avoid out-of-bound access when initializing the planner

bstoeger:
For dives with many samples (i.e. logged dives), samples are merged.

I'm not exactly sure how this code works, but it does an out-of-bound access in some cases. Avoid that by a simple check.

That said, I wonder if this downsampling is a good idea. A user reports that they have logged dives marked as manually added dives. We now load them into edit mode, which means a significant loss of information.

Perhaps we should consider dives with more than 100 samples as non-manual dives?

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I have no idea if this is the right thing to do, but it should remove a reproducible out-of-bound access. I don't know if this is critical, but in theory (with ASAN?) it might be.

In any case, we seem to have a more serious problem with "manually added dives" that in reality are logged dives.

### Additional information:

Needs comment by @atdotde.

### DIFF
--- a/core/planner.c
+++ b/core/planner.c
@@ -1087,7 +1087,7 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 		 * otherwise odd things can happen, such as CVA causing the final ascent to start *later*
 		 * if the ascent rate is slower, which is completely nonsensical.
 		 * Assume final ascent takes 20s, which is the time taken to ascend at 9m/min from 3m */
-		ds->deco_time = clock - bottom_time - stoplevels[stopidx + 1] / last_ascend_rate + 20;
+		ds->deco_time = clock - bottom_time - (M_OR_FT(3,10) * ( prefs.last_stop ? 2 : 1)) / last_ascend_rate + 20;
 	} while (!is_final_plan);
 	decostoptable[decostopcounter].depth = 0;
 
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -147,10 +147,10 @@ void DivePlannerPointsModel::loadFromDive(dive *dIn)
 			break;
 		while (j * plansamples <= i * dc->samples) {
 			const sample &s = dc->sample[j];
-			const sample &prev = dc->sample[j-1];
 			if (s.time.seconds != 0 && (!hasMarkedSamples || s.manually_entered)) {
 				depthsum += s.depth.mm;
-				last_sp = prev.setpoint;
+				if (j > 0)
+					last_sp = dc->sample[j-1].setpoint;
 				++samplecount;
 				newtime = s.time;
 			}


dirkhh:
I completely agree that dives with more than 50 samples should not be considered manually added.

BTW: this confusion comes from people importing dives exported from other divelogs (or via CSV) and those ending up being considered manually imported...

bstoeger:
I found another out-of-bound access in the planner, when editing a downloaded dive in the planner:

In `planner.c`, `sort_stops` gets called with `dnr==1` and `gnr==0`. Thuis, it allocates one "stoplevel".

Later in this line:
```
                ds->deco_time = clock - bottom_time - stoplevels[stopidx + 1] / last_ascend_rate + 20;
```
`stopidx` is 0, which means that it accesses past that one element. That can't be right and `ds->deco_time` is random garbage.

I currently lack the mental capacity to follow this function - but clearly something is amiss.

dirkhh:
how do we continue here? Will you get back to it?

bstoeger:
I would prefer if Robert looks at this, because it would take me an unreasonable long time to understand what this does.

atdotde:
Sorry for the late comment.

What this line is supposed to do is to remove the actual time that it took to ascent from the last stop (note that stop levels is sorted in ascending order such that the further we ascent the smaller the index gets) and replace it with a fixed 20 seconds as explained in a comment before (but I don't remember writing that...). 

The situation you describe with only one element in stop levels would be a dive that ends so shallow that there are no decompressionstops anyway. So it's kind of a boundary case where planning does not really matter anyways.

But to fix this nevertheless, I think the best way forward would be: When this line is reached, stopidx should always be zero (otherwise we would not have left the inner ascent loop. So that stop would be the depth of the shallowest deco stop, which is determined by prefs.last_stop. So I would suggest replacing the access to the stop levels by 

`M_OR_FT(3,10) * ( prefs.last_stop ? 2 : 1)
`

atdotde:
Regarding the original question of what to do with dives with many samples: I think it is wrong to treat a dive that was imported via CSV as manually entered. Why don't we make that dive computer name "Imported dive" upon import? I think this would be better than arbitrarily setting a limit on numbers of samples.

bstoeger:
> Regarding the original question of what to do with dives with many samples: I think it is wrong to treat a dive that was imported via CSV as manually entered. Why don't we make that dive computer name "Imported dive" upon import? I think this would be better than arbitrarily setting a limit on numbers of samples.

It was a script by a user out of our control. The problem is that these dives are now presented with a loss of information, because the planner reduces the number of samples, since we automatically go into edit mode. So either we don't reduce the samples in the planner in edit-mode (as opposed to plan mode) or we recognize these dives as obviously(?) not manually added. The user can still open the dive in the planner if they really want to.

dirkhh:
What's the consensus on this one... as mentioned elsewhere - with the job change I've been a bit distracted...

bstoeger:
I tried to implement what @atdotde suggested above in a second commit, but I have no idea if this is correct. Frankly, I can't follow the code, not because the code is bad, but because I have no understanding whatsoever of all the decompression physics. And likewise I have no idea whether the first commit is reasonable. It's all Chinese to me!

So please, consider this more as a bug report as opposed to a PR.

bstoeger:
@atdotde: This still requires your input on both issues. They seem harmless, but in extremely rare circumstances they might lead to crashes. That is when the arrays end at a page boundary.

dirkhh:
Hey @atdotde  -- this is waiting on input from you, I think 😇

atdotde (approved):
Oops, totally forgot about this. 

The two corrections are obviously correct. Let's take them.

Regarding downsampling: This was introduced when we implemented "Edit dive in planner". The planner UI becomes unmanagable for the user with too many samples, so there it makes sense (when you want to investigate consequences of alterations of the dive for deco). The only thing that is wrong is doing that by default for dives with actual sensor data. Let's just not go into edit mode after importing if we have more than an handful waypoints.

bstoeger:
> The two corrections are obviously correct. Let's take them.

Ok, thanks. Will merge.
 
> Regarding downsampling: This was introduced when we implemented "Edit dive in planner". The planner UI becomes unmanagable for the user with too many samples, so there it makes sense (when you want to investigate consequences of alterations of the dive for deco). The only thing that is wrong is doing that by default for dives with actual sensor data. Let's just not go into edit mode after importing if we have more than an handful waypoints.

I think that is already the case. I will check when I find time.

PR #3545, opened 2022-10-28
Show correct gas density in CCR mode

atdotde:
When collecting the data for the infobox, we have
already computed the current partial pressures of the breathing gas taking into accoutn the divemode. Use those rather than fractions (which for CCR mode are those of diluent) to compute the gas density.

Reported-by: Pietro Tranquillini <p.tranquillini@gmail.com>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -1160,6 +1160,11 @@ static int calculate_ccr_po2(struct plot_data *entry, const struct divecomputer 
 	}
 }
 
+static double gas_density(const struct gas_pressures *pressures)
+{
+	return (pressures->o2 * O2_DENSITY + pressures->he * HE_DENSITY + pressures->n2 * N2_DENSITY) / 1000.0;
+}
+
 static void calculate_gas_information_new(const struct dive *dive, const struct divecomputer *dc, struct plot_info *pi)
 {
 	int i;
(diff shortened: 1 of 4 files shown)


core/profile.c
```diff
@@ -1160,6 +1160,11 @@ static int calculate_ccr_po2(struct plot_data *entry, const struct divecomputer
 	}
 }
 
+double gas_density(const struct gas_pressures *pressures)
```
bstoeger: If this function is used only in this source file, then it should be made of `static` linkage.

I think there is a typo in the commit message: 'diluent' with one 'l'.

atdotde: Addressed your concerns in an updated version.

atdotde:
@dirkhh do you understand why the Android build fails?

glance-:
> Could not GET 'https://jcenter.bintray.com/com/android/tools/build/builder/3.2.1/builder-3.2.1.pom'.
> Read timed out

Clearly external party issues which caused the build to fail.

glance-:
Triggered a re-run of the build, but if we keep seeing such issues we should look at something like:
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

dirkhh:
> Triggered a re-run of the build, but if we keep seeing such issues we should look at something like: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

See my comment elsewhere - I think we need to rev versions in the Android build dependencies again.
Just caching these files isn't really the answer - they usually get removed because there are real bugs in them.

dirkhh (commented):
I assume this is correct - this is your code :)

The massive diff in the CSV files is caused by the resulting change to the gas density, I assume?

atdotde:
Indeed, it's slightly different densities that are recorded. I think this is ready to merge.

PR #3546, opened 2022-10-30
Fix crash when switching from empty state to planner

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

See discussion on mailing list. I was never able to reproduce this.

### Changes made:

1) Recalculate plot info when switch from "too small" state.
2) First plot, then show profile
3) Clear dive pointer when clearing profile

This is all guessing, since I can't reproduce the crash.\


atdotde (approved):
This indeed fixes the crash without noticeable side effects. Thanks @bstoeger !

dirkhh (approved):
this looks good

PR #3547, opened 2022-10-30
Random code cleanups, mostly range-based for loops

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

There was a construct (looping over a vector and determining the index by searching the element) that rubbed me the wrong way so much that I quickly wrote an enumerating-iterator. And then use it in some places.

+minor other stuff, such as removing obsolete code.

### Release note:

No, nothing user visible.

### DIFF
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -206,6 +206,7 @@ HEADERS += \
 	core/pref.h \
 	core/profile.h \
 	core/qthelper.h \
+	core/range.h \
 	core/save-html.h \
 	core/statistics.h \
 	core/units.h \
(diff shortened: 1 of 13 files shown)


dirkhh:
we never keep up to date with the java tools... not sure if this is a temporary failure or something where I need to bump some versions again... The fact that I fundamentally have no idea what I'm doing here (I'm not a Java developer) isn't helping.
I wish we had a Java or Android developer who thought of themselves as maintainer of this build...

dirkhh (approved):
this is one of those cases where the new code makes much more sense to you, the old code made much more sense to me.
I believe this is all good (and I agree with the removal of convertThumbnails)

bstoeger:
Ok, but you have to admit that looping over an array and then for each element searching the element to get the index never makes sense.

PR #3552, opened 2022-11-09
git: don't access global dive site table

bstoeger:
When loading a git repository, dive sites where loaded into the global dive site table, not the local table. Apparently, nobody ever tried to import a git repository into an existing divelog (as opposed to opening it in the application). Because that would have probably given funky results.

Remove this access of a global variable.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes an embarrassing bug, where the git parser would load into global data structures. Thankfully, not many people know how to open git repositories via the file dialog and apparently nobody tried to import data from a git repository.

Since I don't use git, this needs independing testing.

### Release note:

Nobody seems to have notices, so not sure...?

### DIFF
--- a/core/load-git.c
+++ b/core/load-git.c
@@ -177,7 +177,7 @@ static void parse_dive_gps(char *line, struct membuffer *str, struct git_parser_
 	if (!ds) {
 		ds = get_dive_site_by_gps(&location, state->sites);
 		if (!ds)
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
that should be filed under ... uhhh... oops

PR #3553, opened 2022-11-12
[RFC] create divelog structure

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This tries to collect all data that are written into the divelog in a single struct. First of all, to avoid long lists of arguments to the parser functions. But secondly also to decrease the number of global objects. This now makes it very clear, where exactly the global divelog is accessed. And there are many places. Many of these should be eliminated. For example, the log-save routines should be passed an argument to the log, not access the global log, etc.

This fixes a tiny bug (and surely creates dozens of new bugs): when importing dives from a different log, the global autogroup flag was overwritten. This should not happen anymore.

The code is a bit nasty, because it mixes C and C++ paradigms. Since C-code only accesses, but doesn't generate divelog structs, allocation and copying is implemented via C++ constructs, which are #ifdefed out for C. Moreover, the thing has to use plain pointers even though they are owning, which overall gives a rather ugly mess.

Note: this is only compile tested and therefore marked as a draft. I haven't even started the app.

Note: this is large, but mostly semi-automatic code conversion, so complexity is actually quite limited. In particular the first commit could be split, but I hadn't time to come up with something "nice".

### Changes made:

1) Introduce divelog structure
2) Use the structure in various places
3) Minor code hygiene things

### Release note:

No - no user visible change.

### DIFF
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -55,6 +55,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/filterconstraint.cpp \
 	core/filterpreset.cpp \
 	core/divelist.c \
+	core/divelog.cpp \
 	core/gas-model.c \
 	core/gaspressures.c \
 	core/git-access.c \
@@ -227,6 +228,7 @@ HEADERS += \
 	core/filterconstraint.h \
 	core/filterpreset.h \
 	core/divelist.h \
+	core/divelog.h \
 	core/divelogexportlogic.h \
 	core/divesitehelpers.h \
 	core/exif.h \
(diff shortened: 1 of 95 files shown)


dirkhh (approved):
This is quite invasive and touches a lot of code - but once you look at it in context, it seems surprisingly clean and obvious.
What's your preference regarding merging this (and your other just recently rebased PR)?

As far as I'm concerned, this seems ready to merge.

bstoeger:
The other selection-PR is more important to me.

Perhaps ask @torvalds what he thinks of the general idea. I'm sure that he will be put off by the actual implementation, which is a crazy C/C++ mix.

bstoeger:
> As far as I'm concerned, this seems ready to merge.

OK, I rebased on master and will merge if this still compiles. This is invasive, yet somehow limited in scope.

bstoeger:
Well... Hail Mary, etcpp.

PR #3556, opened 2022-11-15
[RFC] statistics: don't import QtQuick in StatsView QML file

bstoeger:
This include prevented the statistics from loading for me on Qt6. And it appears to be unnecessary.

FWIW, with this I can at least start the statistics tab. It looks like garbage though:

```
Primitive topology 0x2 not supported
Primitive topology 0x6 not supported
```

Don't you love it, when APIs break?

### DIFF
--- a/desktop-widgets/qml/statsview2.qml
+++ b/desktop-widgets/qml/statsview2.qml
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: GPL-2.0
-import QtQuick 2.0
 import org.subsurfacedivelog.mobile 1.0
 
 StatsView {
--- a/stats/chartitem.cpp
+++ b/stats/chartitem.cpp
@@ -383,8 +383,8 @@ void ChartLineItem::render(const StatsTheme &)
 void ChartRectLineItem::render(const StatsTheme &)
 {
 	if (!node) {
-		geometry.reset(new QSGGeometry(QSGGeometry::defaultAttributes_Point2D(), 4));
-		geometry->setDrawingMode(QSGGeometry::DrawLineLoop);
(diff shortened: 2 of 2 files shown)


bstoeger:
OK, this is ominous:

The message seems to come from here:
https://code.qt.io/cgit/qt/qtdeclarative.git/tree/src/quick/scenegraph/coreapi/qsgbatchrenderer.cpp?h=v5.14.0
See line 235.

Note that we use `QSGGeometry::DrawTriangleFan` and `QSGGeometry::DrawLineLoop`, which according to this code aren't "supported" anymore.

But in the v5.13.0 I don't see such code. Thus, this misrendering doesn't even seem to be a Qt6 problem, but a Qt >=5.14 problem?

Help?

dirkhh:
IIRC this was one of those things where you HAVE to have openGL drivers for things to work?
I know I've seen this message as well and somehow got it to work, anyway...

dirkhh:
oh sweet, the macOS builds are broken because of some weird SDK dependency.
And I have zero mental bandwidth for any of this right now...

bstoeger:
> IIRC this was one of those things where you HAVE to have openGL drivers for things to work?

I don't think so. Rendering in general still seems to work, but somehow these particular two render modes got "lost" in the Qt code. This is very ominous.

bstoeger:
Ok, I think I know what you mean. This new code is only hit when Qt uses OpenGL? Maybe.

In any case I committed a fix that now makes the statistics tab work for me on Qt6.

Still, I want to file a note of protest that code, which worked perfectly before and followed the official documentation, suddenly breaks. Brrr.

PR #3557, opened 2022-11-16
statistics: fix zero-range on value axis

### LINKED ISSUE #3544: statistics: blank graph displayed if data range is zero
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
In statistics mode (%T), attempting to plot any quantity that has an constant value in all dives,  results in a blank graph.
Presumably this is because of some difficulty with scaling the axis with the constant data (there's no "range" in this case).

For example, a plot of Date (unbinned) (x-axis) vs Weight (y-axis), where the weight value is constant for all the dives.
x-axis and y-axis are both affected the same way.

I acknowledge that this isnt a very interesting graph, but it is more interesting than mysterious and unexpected blank graph.

### Operating system:
M1 MBP, MacOS Monterey 12.6

### Subsurface version:
5.0.10 official Qt5 build.

### Steps to reproduce:

1) load a logbook, and select some dives that have the same value for all dives, eg "weight"
2) view -> dive statistics, and, restrict to selection
3) set up a graph to plot Date (unbinned) vs Weight
4) graph is blank

### Current behavior:
graph is blank

### Expected behavior:
data should be plotted, and some arbitrary scaling of the afflicted axis should be done (eg +/- 1% ?)

### Additional information:
n/a

bstoeger:
The code that calculates the bounds of the value axis was broken when all items had the same value. In that case, increase the shown range explicitly. It doesn't really matter how much the range is increased, because all items will be at the center of the graph.

Also, don't overwrite the "decimal" value of the class. That was just weird.

Fixes #3544.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Create a reasonable value axis for degenerate (single value) value ranges.

I'm a bit tired, so not 100% sure about all the code, but it seemed to work reasonably in a cursory test.

### Related issues:

Fixes #3544.

### Release note:

Done.

### Documentation change:

Not needed.

### Mentions:

Reported by @KimDelmar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+statistics: fix value axis for degenerate value ranges
 profile: Show correct gas density when in CCR mode
 statistics: show correct color of selected scatter items when switching to unbinned mode
 statistics: fix display of month number in continuous date axis
--- a/stats/statsaxis.cpp
+++ b/stats/statsaxis.cpp
@@ -284,7 +284,7 @@ void ValueAxis::updateLabels()
 	// Use full decimal increments
 	double height = max - min;
 	double inc = height / numTicks;
-	double digits = floor(log10(inc));
+	double digits = std::max(floor(log10(inc)), static_cast<double>(-decimals));
 	int digits_int = lrint(digits);
 	double digits_factor = pow(10.0, digits);
 	int inc_int = std::max((int)ceil(inc / digits_factor), 1);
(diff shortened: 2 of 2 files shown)


dirkhh (approved):
I find this code much more logical and easier to understand

PR #3571, opened 2023-01-05
Fixes stray divemaster references in HTML export

AVGP:
In the HTML export, it's still `divemaster` in some places, but elsewhere the property is called `diveguide`. This broke search in the HTML export.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I noticed that the search in the HTML export was broken, so I looked into it.
Turns out that the previous change to this file renamed the `divemaster` property in the dives to `diveguide` but didn't catch all the references to it. This PR fixes the remaining references and, thus, the search feature.

### Changes made:
1. renamed remaining `divemaster` property references to `diveguide`.
2. fixed search feature

### Release note:
Fixes search functionality in the HTML export

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+htmlexport: fix search in HTML export
 statistics: fix value axis for degenerate value ranges
 profile: Show correct gas density when in CCR mode
 statistics: show correct color of selected scatter items when switching to unbinned mode
--- a/theme/dive_export.html
+++ b/theme/dive_export.html
@@ -101,7 +101,7 @@ window.onload=function(){
 
 	//initializing default searchers
 	searchingModules["location"] = new SearchModule(true);
-	searchingModules["divemaster"] = new SearchModule(true);
+	searchingModules["diveguide"] = new SearchModule(true);
 	searchingModules["buddy"] = new SearchModule(true);
 	searchingModules["notes"] = new SearchModule(true);
 	searchingModules["tags"] = new SearchModule(true);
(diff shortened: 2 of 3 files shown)


dirkhh (changes requested):
the goal is actually to go the other direction. So we need to change the export function to write this as `diveguide` instead of `divemaster`, not revert back to the old term that we want to move away from.

AVGP:
Whoops, sorry @dirkhh - fixed it into `diveguide` as needed.

dirkhh:
> Whoops, sorry @dirkhh - fixed it into `diveguide` as needed.

Now if I merge this locally and do a diff against the previous version, the only thing that changes is the CHANGELOG... so I think this may still be missing something 🤔
(you can replicate this here by looking at https://github.com/subsurface/subsurface/pull/3571/files - which I should have done before merging, I guess...)

AVGP:
Hm, I'll take a look soon when I'm not as tired anymore.

AVGP:
Alright, with a bit of sleep, I figured it was the `dive_export.html` not the js file where the references appeared. This includes a UI label. Fixed now :)

atdotde:
I know nothing about this part of the coffee, but why not search_item_diveguide for consistency?

AVGP:
@atdotde Good question. I dunno what the conventions are neither did I check. I just replaced `divemaster` with `diveguide` to fix the issue. But being a good scout, I'll change the name of the HTML element and the JS lookup code to follow the conventions.

atdotde:
:-)

BTW did you get my message on mastodon?

AVGP:
> :-)
> 
> BTW did you get my message on mastodon?

I did indeed and I'll reach out to you separately to figure out how to get you on the podcast :)

dirkhh:
merged #3574

Issue #3560, opened 2022-11-21
Shift dive time does not work


ghost:
### Describe the issue:

- [ x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Shift dive time does not work. The input of the time offset does not take keyboard input.

### Operating system:

macOS 13.0.1 Ventura

iMac

### Subsurface version:

5.0.10

Can't say.

official relase

### Steps to reproduce:

1) Select a dive with the right mouse button and choose "Shift dive times" from the context menue
2) A window (that annoyingly can't be moved) is shown stating Current time: 0:0, Shifted time: 0:0
<img width="351" alt="1_window_opened" src="https://user-images.githubusercontent.com/3070848/203150004-d7fbf0b0-1290-48f0-a8c8-8feeffc79aa7.png">
3) Try to enter a time difference

The input field does not take any input from the keyboard. The up/down arrows only influence the minutes. Changing the time to one minute earlier using the down arrow sets the shifted time to Thu, Jan 1, 1970 00:59.
<img width="347" alt="2_time_shifted_by_one_minute" src="https://user-images.githubusercontent.com/3070848/203150073-31fdb178-0fb7-4624-a346-3ce5e6ce748e.png">

### Current behavior:

Feature is unusable.

### Expected behavior:

Time of selected dive is shifted by a given value.

jwobser:
I was able to reproduce the Jan 1 1970 UI error on v5.0.10 / MacOS 13.0.1. 
However, upon clicking 'OK' the dive times are shifted appropriately in the dive list, so the error only seems to be in displaying the time in the dialog.

[I suspect the issue lies with updating the UI from the time string](https://github.com/subsurface/subsurface/blob/31277daa71b2ed7312672bbfa554a234c9ee5bdc/desktop-widgets/simplewidgets.cpp#L105) but I'm not familiar with Qt to know why it is not working correctly on Mac.

I'm also unable to compile from source on an M1 mac mini; Qt is trying to link using x86_64 framework libraries.

jwobser:
Looks like Windows is also impacted. #3359 

[...]

PR #3573, opened 2023-01-18
desktop: pass dive list to ShiftTimesDialog

bstoeger:
Users report that the ShiftTimesDialog does not work on Mac and Windows. Apparently, get_first_selected_dive returns NULL, which should not be possible because the dialog is only created when dives are selected. Very omninous.

Get the selected dives in the caller and pass them down. This bug at least should not happen anylonger. Perhaps now the dialog does not show at all, but that will narrow down the root cause of the problem.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Potential fix for the UI issue reported in #3560. At least it should give us a better idea of what is going on.

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -845,7 +845,10 @@ void DiveListView::contextMenuEvent(QContextMenuEvent *event)
 
 void DiveListView::shiftTimes()
 {
-	ShiftTimesDialog dialog(MainWindow::instance());
+	std::vector<dive *> dives = getDiveSelection();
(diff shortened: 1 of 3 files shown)


bstoeger:
N.B.: If this fixes things, we probably should also adapt ShiftImageTimesDialog.

dirkhh (approved):
this seems reasonable and as added benefit fixes the user's problem.
I keep staring at it, trying to figure out WHY, though...

PR #3585, opened 2023-02-03
NULL cylinders are never the same

### LINKED ISSUE #3578: macOS crash updating cylinder for multiple dives
### Describe the issue:

- [X ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
Attempting to enter cylinder type on multiple dives at once, app crashes. Updating cylinder on individual dives works. When selecting larger numbers (not always the same, but 100+ consistently crashes, but ~25 has both worked and crashed.

Background: New subsurface install, migrating data from MacDive. Exported all dives into UDDF, imported into Subsurface, then saved to cloud log. 

### Operating system:
macOS Monterey 12.5.1 (English)
iMac 27" 2017 (3.8GHz i5 40GB)

### Subsurface version:
5.0.10

### Steps to reproduce:
1. Select 100+ dives in dive log.
2. On Equipment tab, Cylinder type is empty, select AL80 from pulldown menu.
3. Crash

### Current behavior:
App Crashes

### Expected behavior:
Value would update

### Additional information:
Crash log attached
[subsurface macOS crash log.txt](https://github.com/subsurface/subsurface/files/10539259/subsurface.macOS.crash.log.txt)

atdotde:
When editing many dives it can happen that a cylinder index is invalid which results in  a NULL cylinter pointer. Treat such a cylinder as different from any other cylinder rather than dereferencing the NULL pointer and crashing.

Fixes #3578 (at least as far as I can tell)

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1165,16 +1165,20 @@ void AddCylinder::redo()
 	}
 }
 
-static bool same_cylinder_type(const cylinder_t &cyl1, const cylinder_t &cyl2)
+static bool same_cylinder_type(const cylinder_t *cyl1, const cylinder_t *cyl2)
 {
-	return same_string(cyl1.type.description, cyl2.type.description) &&
-	       cyl1.cylinder_use == cyl2.cylinder_use;
+	if (cyl1 == NULL || cyl2 == NULL)
+		return 0;
+	return same_string(cyl1->type.description, cyl2->type.description) &&
+	       cyl1->cylinder_use == cyl2->cylinder_use;
 }
 
-static bool same_cylinder_size(const cylinder_t &cyl1, const cylinder_t &cyl2)
+static bool same_cylinder_size(const cylinder_t *cyl1, const cylinder_t *cyl2)
 {
-	return cyl1.type.size.mliter == cyl2.type.size.mliter &&
-	       cyl1.type.workingpressure.mbar == cyl2.type.workingpressure.mbar;
+	if (cyl1 == NULL || cyl2 == NULL)
+		return 0;
+	return cyl1->type.size.mliter == cyl2->type.size.mliter &&
+	       cyl1->type.workingpressure.mbar == cyl2->type.workingpressure.mbar;
 }
 
 // Flags for comparing cylinders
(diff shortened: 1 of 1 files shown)


mikeller (approved):
Good find! I had run into this as well before, but hadn't yet gotten around to looking for a fix.

bstoeger:
I am not a really a fan of this and would prefer to fix the root cause.

bstoeger:
Well, I failed to properly review 3629a87fcc, which is broken.

I would prefer something like this:
```
diff --git a/commands/command_edit.cpp b/commands/command_edit.cpp
index d6de5f740..c375e1ab6 100644
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1200,10 +1200,13 @@ EditCylinderBase::EditCylinderBase(int index, bool currentDiveOnly, bool nonProt
 	cyl.reserve(dives.size());
 
 	for (dive *d: dives) {
+		if (index >= d->cylinders.nr)
+			continue;
 		if (nonProtectedOnly && is_cylinder_prot(d, index))
 			continue;
+		const cylinder_t &cylinder = *get_cylinder(d, index);
 		if (d != current &&
-		    (!same_cylinder_size(orig, *get_cylinder(d, index)) || !same_cylinder_type(orig, *get_cylinder(d, index))))
+		    (!same_cylinder_size(orig, cylinder) || !same_cylinder_type(orig, cylinder)))
 			// when editing cylinders, we assume that the user wanted to edit the 'n-th' cylinder
 			// and we only do edit that cylinder, if it was the same type as the one in the current dive
 			continue;
@@ -1212,7 +1215,7 @@ EditCylinderBase::EditCylinderBase(int index, bool currentDiveOnly, bool nonProt
 		// that's silly as it's always the same value - but we need this vector of indices in the case where we add
 		// a cylinder to several dives as the spot will potentially be different in different dives
 		indexes.push_back(index);
-		cyl.push_back(clone_cylinder(*get_cylinder(d, index)));
+		cyl.push_back(clone_cylinder(cylinder));
 	}
 	dives = std::move(divesNew);
 }
```
(Not even compile-tested, but you get the point).

dirkhh:
I like fixing this in the caller better. Because it's the caller where we have the error (dereferencing the NULL pointer).

bstoeger:
> I like fixing this in the caller better. Because it's the caller where we have the error (dereferencing the NULL pointer).

Indeed. This line in the original PR
```
cyl.push_back(clone_cylinder(*get_cylinder(d, index)));
```
looks like a potential disaster. But it can't happen for the very subtle reason that "NULL cylinders", whatever that should mean, always compare as unequal. That would need a big comment.

Instead, check that the cylinder exists and turn it into a reference, to signal the compiler and the fellow programmers that this is guaranteed to not be null.

bstoeger:
PS: I can post a PR later.

bstoeger:
Fixed by #3594.

PR #3586, opened 2023-02-05
parser: fix end-of-string check in import-csv.c

bstoeger:
For details, see 096de0efd01e12de6c4ce968eead90e95ff8157f.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A rather trivial bug fix. The old code looks actively dangerous, as the NULL check did not work and a smart compiler would throw it away.

### DIFF
--- a/core/import-csv.c
+++ b/core/import-csv.c
@@ -178,7 +178,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
 		memset(tmpbuf, 0, sizeof(tmpbuf));
 		iter = strchr(iter, '|');
 
-		if (iter && iter[1]) {
+		if (iter) {
 			iter = iter + 1;
 			iter_end = strchr(iter, '|');
 
@@ -203,7 +203,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
 			for (i = 0; i < 5 && iter; ++i)
 				iter = strchr(iter + 1, '|');
 
-			if (iter && iter + 1) {
+			if (iter) {
 				iter = iter + 1;
 				iter_end = strchr(iter, '|');
 
@@ -254,7 +254,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
 			for (i = 0; i < 5 && iter; ++i)
 				iter = strchr(iter + 1, '|');
 
-			if (iter && iter + 1) {
+			if (iter) {
 				iter = iter + 1;
 				iter_end = strchr(iter, '|');
 


bstoeger:
Alternatively, one could omit the second condition completely. The code will proceed correctly, if `iter[1]` is `\0`.

bstoeger:
Actually, the second condition should be removed. Will do tomorrow.

bstoeger:
Let's commit. If I overlook something, this can still be reverted.

core/import-csv.c
```diff
@@ -178,7 +178,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
 		memset(tmpbuf, 0, sizeof(tmpbuf));
 		iter = strchr(iter, '|');
 
-		if (iter && iter[1]) {
```
dirkhh, line 181: so this one makes sense. As it checks if `iter` is a non-Null pointer, and if the second character is not `'\0'`
Removing this seems potentially wrong, because now we could call `xml_params_add()` with an empty string which I don't think is what we would want here...

core/import-csv.c
```diff
@@ -203,7 +203,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
 			for (i = 0; i < 5 && iter; ++i)
 				iter = strchr(iter + 1, '|');
 
-			if (iter && iter + 1) {
```
dirkhh, line 206: and this one really should be
```
  if (iter && iter[1])
```

core/import-csv.c
```diff
@@ -254,7 +254,7 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
...
```
dirkhh, line 257: ditto

dirkhh:
yeah, I know - reviewed an hour after you merged it. 🤷🏼‍♂️

core/import-csv.c
bstoeger: Doesn't really make sense to me.

The code looks like this:
- Find '|'.
- If found, find another '|' after that
- If that is also found, do something with the string between both '|'s.

Why would it matter if the first '|' is the last character in the string?

And farther down, it just searches for the end of the line anyway.

core/import-csv.c
bstoeger: Again, I don't see why. What's the difference between an unmatched '|' being at the end of the string or not?

PR #3576, opened 2023-01-20
Improvement: Show gases as diluent by default for CCR dives.

mikeller:
### Describe the pull request:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Instead of adding all gases read from a dive computer as part of a dive log as 'OC-gas', add gases as 'diluent' if the dive has a dive mode of 'CCR'. This creates consistency with the ppO2 for CCR dives being tracked as sensor readings or a fixed setpoint, and not as the ppO2 of the current gas ad depth.
A follow up question from this is whether gas use in the cylinders list on the Equipment tab should be user editable. This seems to be inconsistent at the moment, with gas constituent percentages downloaded from the dive computer being editable, but gas use not.

### Changes made:

1) use 'diluent' as the default gas use if the dive mode is 'CCR'

### Additional information:

image

### Release note:

equipment: Use 'diluent' as default gas use type if the dive mode is 'CCR'

### Mentions:

Signed-off-by: Michael Keller <github@ike.ch>

PR #3512, opened 2022-09-03
Selection code reshuffling, step 1 of many.

bstoeger:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

The selection code has always been a thorn in my side. It is a single horrible layering violation. And it doesn't work well with mobile, where we can have multiple profiles at the same time.

This is a beginning of cleaning this up. It does fix a small bug and removes some truly horrible code in `DiveListView`. At some points the code becomes more verbose, though arguably less convoluted. Of course, it may break things.

Ultimately I think the root cause of the problem is that the selection- and filter-status is saved in the core, when it should not be. The filter status should be stored in the filter-model and the selection-status in the `DiveListView`. However, we are very far from that vision. I see two major stumbling blocks:
1) Getting the selection into C code. Handling collections in C is very cumbersome.
2) The model/view code can be appallingly slow.

### Release note:

Probably not needed.

### Documentation change:

Not needed.

PR #3587, opened 2023-02-05
Cylindermodel: Improve editing of tank use

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Use the drop down for editing the tank use in the gas list in both the equipment tab and the dive planner.
The tank use column is now available in the equipment tab for all dives and not just CCR dives, as 'not used' is a valid entry in both cases. However, if the current dive is an OC dive, only 'OC-gas' and 'not used' are shown.
There still seems to be a problem that in some cases, when opening the planner after selecting an existing CCR dive the drop down in the planner does not list CCR gas uses - for some reason `displayed_dive` does not seem to be updated correctly on opening of the planner. But I have not been able to reproduce this consistently, and changing 'Dive mode' fixes this.

### Changes made:

1) use the tank use drop down in the dive planner instead of the on / off toggle
2) always show the tank use column in the equipment tab
3) only populate the tank use combobox with 'OC-gas' and 'not used' if the current / planned dive is an OC dive

### Additional information:

image
image
image

### Release note:

desktop: use tank dynamic tank use drop down in equipment tab and planner

### Mentions:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: use dynamic tank use drop down in equipment tab and planner
 desktop: fix brightness configuration for OSTC4
 equipment: Use 'diluent' as default gas use type if the dive mode is 'CCR'
 htmlexport: fix search in HTML export
--- a/desktop-widgets/modeldelegates.cpp
+++ b/desktop-widgets/modeldelegates.cpp
@@ -12,6 +12,7 @@
 #include "qt-models/tankinfomodel.h"
 #include "qt-models/weightsysteminfomodel.h"
 #include "qt-models/weightmodel.h"
+#include "qt-models/diveplannermodel.h"
 #include "qt-models/divetripmodel.h"
 #include "qt-models/divelocationmodel.h"
 #include "core/qthelper.h"
@@ -254,9 +255,22 @@ TankUseDelegate::TankUseDelegate(QObject *parent) : QStyledItemDelegate(parent)
 
 QWidget *TankUseDelegate::createEditor(QWidget *parent, const QStyleOptionViewItem &, const QModelIndex &) const
 {
+	struct divecomputer *currentDc;
+	if (DivePlannerPointsModel::instance()->currentMode() != DivePlannerPointsModel::NOTHING) {
+		currentDc = &displayed_dive.dc;
+	} else {
+		currentDc = get_dive_dc(current_dive, dc_number);
+	}
 	QComboBox *comboBox = new QComboBox(parent);
-	for (int i = 0; i < NUM_GAS_USE; i++)
-		comboBox->addItem(gettextFromC::tr(cylinderuse_text[i]));
+	if (!currentDc) {
+		return comboBox;
+	}
+	bool isCcrDive = currentDc->divemode == CCR;
+	for (int i = 0; i < NUM_GAS_USE; i++) {
+		if (isCcrDive || (i != DILUENT && i != OXYGEN)) {
+			comboBox->addItem(gettextFromC::tr(cylinderuse_text[i]));
+		}
+	}
 	return comboBox;
 }
 
@@ -270,7 +284,7 @@ void TankUseDelegate::setEditorData(QWidget *editor, const QModelIndex &index) c
 void TankUseDelegate::setModelData(QWidget *editor, QAbstractItemModel *model, const QModelIndex &index) const
 {
 	QComboBox *comboBox = qobject_cast<QComboBox *>(editor);
-	model->setData(index, comboBox->currentIndex());
+	model->setData(index, cylinderuse_from_text(qPrintable(comboBox->currentText())));
 }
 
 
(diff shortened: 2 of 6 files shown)


bstoeger:
>  for some reason displayed_dive does not seem to be updated correctly on opening of the planner. But I have not been able to reproduce this consistently, and changing 'Dive mode' fixes this.

Note that the global `displiayed_dive` is (finally!) removed in #3512. However, owing to Dirk/s day job I'm not positive that this will be merged anytime soon :(.

dirkhh:
Berthold - you are more than welcome to merge things that you are confident in.
And I hope that things will ease up "soonishly" for me... unfortunately today I will spend mostly in the car, driving back from a campus tour with my kids, so I doubt I'll be reviewing / merging anything today :-/

bstoeger:
If you feel that things will ease up in a reasonable time frame, it's probably better to wait with the huge things.

desktop-widgets/modeldelegates.cpp
```diff
@@ -254,9 +255,16 @@ TankUseDelegate::TankUseDelegate(QObject *parent) : QStyledItemDelegate(parent)
...
+	struct divecomputer *currentDc;
+	if (DivePlannerPointsModel::instance()->currentMode() != DivePlannerPointsModel::NOTHING)
+		currentDc = &displayed_dive.dc;
+	else
+		currentDc = get_dive_dc(current_dive, dc_number);
+	bool isCcrDive = currentDc ? currentDc->divemode == CCR : true;
 	QComboBox *comboBox = new QComboBox(parent);
 	for (int i = 0; i < NUM_GAS_USE; i++)
-		comboBox->addItem(gettextFromC::tr(cylinderuse_text[i]));
+		if (isCcrDive || ((i != DILUENT) && (i != OXYGEN)))
```
bstoeger: Coding style: add braces to the `for`-loop since the body grew beyond a single line.

Also I am not sure if the parentheses around `i != DILUENT` and `i != OXYGEN` are idiomatic.

desktop-widgets/modeldelegates.cpp
```diff
@@ -254,9 +255,16 @@ TankUseDelegate::TankUseDelegate(QObject *parent) : QStyledItemDelegate(parent)
 
 QWidget *TankUseDelegate::createEditor(QWidget *parent, const QStyleOptionViewItem &, const QModelIndex &) const
 {
+	struct divecomputer *currentDc;
+	if (DivePlannerPointsModel::instance()->currentMode() != DivePlannerPointsModel::NOTHING)
+		currentDc = &displayed_dive.dc;
+	else
+		currentDc = get_dive_dc(current_dive, dc_number);
+	bool isCcrDive = currentDc ? currentDc->divemode == CCR : true;
```
bstoeger: If `currendDc` is null, this should probably exit early. The only time this can happen is if `current_dive` is null itself. In that case we don't want to display anything - do we?

So I would suggest a
```
        if (!currentDc)
               return new QComboBox(parent);
```
The line above then simplifies to
```
        bool isCcrDive = currentDc->divemode == CCR;
```

bstoeger (approved):
Since I am not familiar with gases and rebreathers, only stylistic comments.

bstoeger:
@atdotde: I think this is your call.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+desktop: use tank dynamic tank use drop down in equipment tab and planner
```
atdotde: This could be clearer...

atdotde (commented):
I see value in unification. But what was your original motivation to do this? I would say that an on/off switch is faster to use than a combo box with two entries. Why do you think this is an improvement? Just curious.

mikeller:
@atdotde: My original motivation for doing this comes out of the change I introduced in #3576: Now that we actually get gases added as 'diluent' to the dive log, we should also be able to edit their use, and plan with them.
In this respect the existing functionality in the planner is broken: If the user starts a plan based on an existing CCR dive with gases that are not 'OC-gas' or 'not used', and these gases get imported into the planner, the use of these gases can not be changed in the planner.
But consistency is another factor - when there is a mostly identical gas list used in multiple places, most users will find it easiest to use if the UX for the list is the same as much as possible across all the different places.
And last but not least, rebreather divers using Subsurface might wish to make gas plans that include their diluent and oxygen tanks as well, and have them clearly identified as what they are. Maybe somebody will even be keen to add the capability to track usage for O2 (time based) and diluent (based on depth increase) into the planner. This might be of value for divers planning long dives.

atdotde:
Ok, I am convinced. In the part, we have discussed tracking/planning of CCR gases a number of times, always with the result it can't be done in a reliable way.

atdotde (approved):
Will merge

PR #3574, opened 2023-01-19
HTML export divemaster / diveguide fix

dirkhh:
### Describe the pull request:

- [x] Bug fix
- [x] Code cleanup

### Pull request long description:

This is simply a rebase / cleanup / update of commit messages for #3571 

This will run the CI/CD (which will still fail for Snap) and make it easier to just merge.

### Mention
@AVGP I hope it's ok that I made a few minor cleanups (the code is unchanged) 😇

PR #3588, opened 2023-02-05
Htmlexport diveguide

KimDelmar:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

With very recent builds from git, dive guide(s) are shown as "undefined" when exported to HTML and the dive details are expanded.

### Changes made:

One change of divemaster -> diveguide in core/save-html.c so the correct field name ends up in the exported file.js

### Related issues:

Probably fallout from PR #3574

### Mentions:

Signed-off-by: Kim Delmar <62100831+KimDelmar@users.noreply.github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 desktop: fix brightness configuration for OSTC4
 equipment: Use 'diluent' as default gas use type if the dive mode is 'CCR'
 htmlexport: fix search in HTML export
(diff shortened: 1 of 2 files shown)


bstoeger:
I was reluctant to change the HTML export. But yeah, with #3574, that ship has sailed so we probably should merge this PR.

PR #3590, opened 2023-02-07
ci: bring Snap USNs back

Saviq:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Ref. https://bugs.launchpad.net/snapstore-server/+bug/2004008

While that gets resolved, ignore the extra packages when checking for security notices.

### Changes made:
Use the `--ignore-pockets` from review-tools `edge` channel to skip security fixes unavailable to the build system.

### Mentions:
@dirkhh this brings back the security checks / rebuilds. I will drop the `--edge` when that becomes available on stable. Note that i386 isn't built since 5.0.8, I wonder if we should unlist it from the store?


dirkhh (approved):
seems reasonable 
remind me why i386 is no longer built?

Saviq:
> remind me why i386 is no longer built?

(Full) i386 support was dropped before Ubuntu 20.04. There's only partial support left: https://wiki.ubuntu.com/i386.

dirkhh:
Ah, that makes sense. Then we should indeed just simply stop mentioning that.

PR #3593, opened 2023-02-09
Improve accuracy of the 'unlikely dive gas' warning.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Change the values supplied in the warning to be fractions. This is what is actually reported by libdivecomputer. The currently used thousandths are hard to interpret for users, as they are only used internally in Subsurface.

### Changes made:

1) changed 'unlikely dive gas' warnings to list the gases in the same format as libdivecomputer.

### Mentions:


atdotde (approved):
Makes sense since the float is the original value as obtained from libdivecomputer and in this case, the ints are derived values.

dirkhh (approved):
Agreed, this makes more sense from a user perspective.

PR #3595, opened 2023-02-12
cleanup: redo event-name code [+small bugfix]

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

While polishing the profile code, the event-name code always annoyed me. It was splattered all over the place, accessed global variables, leaked on exit, etc. This collects all the code in one place, rewrites it in C++ and provides a C-only interface.

To see why C++ is preferable here, compare
```
void remember_event(const char *eventname)		
{		
	int i = 0, len;		

	if (!eventname || (len = strlen(eventname)) == 0)		
		return;		
	while (i < evn_used) {		
		if (!strncmp(eventname, ev_namelist[i].ev_name, len))		
			return;		
		i++;		
	}		
	if (evn_used == evn_allocated) {		
		evn_allocated += 10;		
		ev_namelist = realloc(ev_namelist, evn_allocated * sizeof(struct ev_select));		
		if (!ev_namelist)		
			/* we are screwed, but let's just bail out */		
			return;		
	}		
	ev_namelist[evn_used].ev_name = strdup(eventname);		
	ev_namelist[evn_used].plot_ev = true;		
	evn_used++;		
}
```
to
```
extern "C" void remember_event_name(const char *eventname)
{
	if (empty_string(eventname))
		return;
	if (std::find(event_names.begin(), event_names.end(), eventname) != event_names.end())
		return;
	event_names.push_back({ eventname, true });
}
```

The second commit fixes a bug, which was introduced during undoification of the event system.

### Release note:

Probably not necessary, bug is too obscure.

### Mentions:

@dirkhh: this is a short one and should be easy to review.

### DIFF
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -51,6 +51,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/divecomputer.c \
 	core/divefilter.cpp \
 	core/event.c \
+	core/eventname.cpp \
 	core/filterconstraint.cpp \
 	core/filterpreset.cpp \
 	core/divelist.c \
@@ -200,6 +201,7 @@ HEADERS += \
 	core/dive.h \
 	core/divecomputer.h \
 	core/event.h \
+	core/eventname.h \
 	core/extradata.h \
 	core/git-access.h \
 	core/globals.h \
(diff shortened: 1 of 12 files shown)


dirkhh (approved):
I don't say this often - but yeah, the C++ code is much nicer. Well done.

That third commit does seem a bit unrelated, but whatever.

PR #3598, opened 2023-02-17
Desktop: Fix bug in diveplan for CCR dives with multiple segments.

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a bug reported in
https://groups.google.com/g/subsurface-divelog/c/8N3cTz2Zv5E: When planning a CCR dive with multiple segments, the textual dive plan was showing a non-existent gas change with bogus data.  The first part of the fix is uncluttering of the message printed: Since this change is _after_ the current diveplanpoint the data needs to come from `nextdp` and not `dp`.  The second part is that the message is not printed any more if the current and the following segments have been manually added: According to comments in the code the change should only be printed on the segment _before_ the change if this segment is an ascent segment that is followed by a manually entered segment.

### Changes made:

1) fixed contents of change text when printed on segment before change
2) fixed conditions for printing changes on segment before change

### Related issues:

https://groups.google.com/g/subsurface-divelog/c/8N3cTz2Zv5E

### Additional information:

image

In general I am not fully clear on why we are printing changes in the segment _before_ they happen - all I could find on this was a comment stating that this is so without a reason:

https://github.com/subsurface/subsurface/blob/ab7b9329c0ce00297faba256c22ddcdb1f1e8912/core/plannernotes.c#L325-L327

What I can see is that this may result in an unsafe dive plan if the diver is to execute the change at the start of the segment it is listed in, as is required for most changes: In the example below, the diver is asked to change to EAN32 at the start of an ascent segment when they are at 80 meters, resulting in a ppO2 of ~ 2.9!
So this begs the question if the special case of listing changes in the segment before they are required should be removed, or at least checks added to avoid doing this in situations where it results in dangerous conditions.

image

### Release note:

desktop: fix bug when printing a dive plan with multiple segments

### Mentions:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: fix bug when printing a dive plan with multiple segments
 desktop: fix bug in bailout gas selection for CCR dives
 desktop: fix crash on cylinder update of multiple dives
 desktop: use dynamic tank use drop down in equipment tab and planner
--- a/core/plannernotes.c
+++ b/core/plannernotes.c
@@ -325,15 +325,15 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
 				/* Normally a gas change is displayed on the stopping segment, so only display a gas change at the end of
 				 * an ascent segment if it is not followed by a stop
 				 */
-				if ((isascent || dp->entered) && gaschange_after && dp->next && nextdp && (dp->depth.mm != nextdp->depth.mm || nextdp->entered)) {
-					if (dp->setpoint) {
-						asprintf_loc(&temp, translate("gettextFromC", "(SP = %.1fbar CCR)"), dp->setpoint / 1000.0);
+				if (isascent && gaschange_after && dp->next && nextdp && nextdp->entered) {
+					if (nextdp->setpoint) {
+						asprintf_loc(&temp, translate("gettextFromC", "(SP = %.1fbar CCR)"), nextdp->setpoint / 1000.0);
 						put_format(&buf, "<td style='padding-left: 10px; color: red; float: left;'><b>%s %s</b></td>",
 							gasname(newgasmix), temp);
 						free(temp);
 					} else {
 						put_format(&buf, "<td style='padding-left: 10px; color: red; float: left;'><b>%s %s</b></td>",
-							gasname(newgasmix), lastdivemode == UNDEF_COMP_TYPE || lastdivemode == dp->divemode ? "" : translate("gettextFromC", divemode_text_ui[dp->divemode]));
+							gasname(newgasmix), dp->divemode == UNDEF_COMP_TYPE || dp->divemode == nextdp->divemode ? "" : translate("gettextFromC", divemode_text_ui[nextdp->divemode]));
 						if (isascent && (get_he(lastprintgasmix) > 0)) { // For a trimix gas change on ascent, save ICD info if previous cylinder had helium
 							if (isobaric_counterdiffusion(lastprintgasmix, newgasmix, &icdvalues)) // Do icd calulations
 								icdwarning = true;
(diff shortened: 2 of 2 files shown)


dirkhh (commented):
That code is not something I'm super comfortable reviewing... Robert, this one's for you :)

atdotde:
I will take care of this. The diveplan and where to print gas changes is super fragile (due to the waypoint, as in: point of time where something happens vs leg of the dive as in: the time interval between the events and where dp fits in between). I will look at this once I have some quiet time, hopefully early next week.

atdotde:
As I said, this is all very fragile. And yes, printing the gas change information on the ascent leg from 80m to 38m is clearly wrong, it should be shown at 38m.

IIRC this comes from the following situation: you ascent from  x to y, change gas without stopping then continue to z. Then I think one might argue that showing it on the first leg might be justified as you would interpret it as "ascent to y to change gas" but even in this case it is probably inconsistent. In the display, indicating a gas should probably better always mean "do this leg with the specified gas, i.e. understand that the gas change always happens in the beginning of the leg that it is indicated in. 

Would you like to fix this?

Just make sure when you change the very convoluted logic here that you don't break any for the many other cases (manual entered segments vs computed segments, explicit ascents vs implied ascents in deco and the verbatim plan is a totally different beast on its own.

As said before, many of these complications arise from the two different semantics: Internally we have dp's that indicate instances of time. If they have gas (or set point) information that is supposed to mean "from this point in time use the following gas" whereas on display we show segments (legs) of a dive and gas information means "use this gas for this segment, if necessary change gas at the start".

Maybe there is a simpler way to do this, but this has gone trough many many iterations in the past.

mikeller:
Thanks for your explanation @atdotde. I am happy to have look into potential improvements for the diveplan - I think whatever changes should be made will need to be discussed first.

But is it possible to get this pull request reviewed and merged before the next release? This is a fix for a user reported bug (https://groups.google.com/g/subsurface-divelog/c/8N3cTz2Zv5E), and apart from being surfaced in a confusing place, the information surfaced is also incorrect in the current implementation.

As for possible future improvements, I think the aim should be that Subsurface produces a dive plan that the diver can copy or print onto a slate, and then execute during the dive while potentially suffering from perceptual narrowing and / or narcosis. So it probably isn't ideal if a diver in these conditions has to figure out if a gaschange should be done at the beginning or at the end of the segment it appears on.

As to listing all gas changes to be executed at the beginning or end of the segment they appear on I am not sure what is better:
- 'execute at end' seems to be allowing a more natural way of reading the dive plan, e.g. 'ascend to 21m, then switch to EAN50'. But obviously it makes it harder to list the initial gas mix at the start of the dive - which could be done by adding a 0m / 0min descent start point;
- 'execute at start' is less natural to read, but is what a number of other deco planners that i checked are doing - so most divers will be familiar with this format (which doesn't mean we can't do it better ;-) ).

As a third option, if we decide that there is a good reason to stay with a combination of 'execute at end' and 'execute at start', then I'd at least want to add some form of annotation to indicate that a change is to be done different to the norm, like this:

```
  | depth | duration | runtime | gas
-- | -- | -- | -- | --
➘ | 55m | 2min | 2min | (20/20) (SP = 1.1bar CCR)
➙ | 55m | 18min | 20min |  
- | 55m | 4min | 24min | air Open circuit
➚ | 18m | 4min | 28min |(*) EAN50
- | 18m | 2min | 30min |
➚ | 15m | 0min | 30min |  
- | 15m | 2min | 32min |  
➚ | 12m | 0min | 32min |  
- | 12m | 4min | 36min |  
➚ | 9m | 0min | 36min |  
- | 9m | 6min | 42min |  
➚ | 6m | 0min | 42min | (*) oxygen
- | 6m | 7min | 49min |
➚ | 3m | 1min | 50min |  
- | 3m | 10min | 60min |  
➚ | 0m | 1min | 61min |  
(*): Execute gas change at the end of the segment
```

As for the verbatim dive plan, after looking at the code I think what we have now could be simplified considerably if it was refactored to use separate functions to generate the 'normal' and the verbatim dive plan. This will reduce the level of complexity considerably, and make it easier to maintain both variants.

PR #3842, opened 2023-03-08
Planner: create mirrored return profile

jwobser:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

2 commits: 1 to remove unnecessary method calls within planner, and another to implement a feature request from mailing list 2023 MAR 4

Add a shortcut (CTRL + M) to append a reversed dive profile in cases where a diver must exit the same way they entered. 

The reversing method does not check for decompression obligations. The user must manually extend dive segments where decompression is required. 

All properties of the point, except runtime, are copied when reversed. Runtime is updated as needed to make a mirror image. 

Sorting the vector of points on runtime is maintained, because the mirror points are appended in increasing runtime order.

### Changes made:

1) Add new shortcut (CTRL + M) to dive planner
2) Add method to get time deltas between planner points to planner model
3) Add method to create and append a reversed profile to planner model

### Documentation change:

Added comment about the shortcut in the technical dive planning section of the user manual.

### DIFF
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -78,6 +78,9 @@ DivePlannerWidget::DivePlannerWidget(QWidget *parent) : QWidget(parent, QFlag(0)
 	QShortcut *closeKey = new QShortcut(QKeySequence(Qt::Key_Escape), this);
 	connect(closeKey, &QShortcut::activated, plannerModel, &DivePlannerPointsModel::cancelPlan);
 
+	QShortcut *mirrorProfileShortcut = new QShortcut(QKeySequence(Qt::Key_M + Qt::CTRL), this);
+	connect(mirrorProfileShortcut, &QShortcut::activated, plannerModel, &DivePlannerPointsModel::addReverseProfile);
+
 	// This makes shure the spinbox gets a setMinimum(0) on it so we can't have negative time or depth.
 	// Limit segments to a depth of 1000 m/3300 ft and a duration of 100 h. Setting the limit for
 	// the depth will be done in settingChanged() since this depends on the chosen units.
--- a/Documentation/user-manual.txt
+++ b/Documentation/user-manual.txt
@@ -4078,6 +4078,10 @@ The waypoints listed in the _Dive Planner Points_ dialog can be edited by hand i
 order to get a precise presentation of the dive plan. In fact, it is sometimes more easy to create the
 whole dive profile by editing the _Dive Planner Points_ dialog.
 
+For planning a dive which has an exit profile which is a mirror image of the entry
+(e.g. cave diving), there is a shortcut to mirror the first half of the dive. Once the intial half of
+the dive is planned, press CTRL+M to create a mirrored exit profile.
+
 Show any changes in gas cylinder used by indicating gas changes as explained
 in the section <<S_CreateProfile,hand-creating a dive profile>>. These changes should
 reflect the cylinders and gas compositions defined in the table with _Available Gases_.
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -47,7 +47,6 @@ void DivePlannerPointsModel::removeSelectedPoints(const std::vector<int> &rows)
 {
 	removePoints(rows);
 
-	updateDiveProfile();
 	emitDataChanged();
 	cylinders.updateTrashIcon();
 }
@@ -179,7 +178,6 @@ void DivePlannerPointsModel::loadFromDive(dive *dIn)
 	DiveTypeSelectionModel::instance()->repopulate();
 	preserved_until = d->duration;
 
-	updateDiveProfile();
 	emitDataChanged();
 }
 
@@ -921,7 +919,6 @@ void DivePlannerPointsModel::removeControlPressed(const QModelIndex &index)
 	if (divepoints[0].cylinderid != old_first_cylid)
 		cylinders.moveAtFirst(divepoints[0].cylinderid);
 
-	updateDiveProfile();
 	emitDataChanged();
 }
 
@@ -958,7 +955,6 @@ void DivePlannerPointsModel::remove(const QModelIndex &index)
 	if (divepoints[0].cylinderid != old_first_cylid)
 		cylinders.moveAtFirst(divepoints[0].cylinderid);
 
-	updateDiveProfile();
 	emitDataChanged();
 }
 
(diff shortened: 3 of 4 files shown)


desktop-widgets/diveplanner.cpp
```diff
@@ -78,6 +78,9 @@ DivePlannerWidget::DivePlannerWidget(QWidget *parent) : QWidget(parent, QFlag(0)
 	QShortcut *closeKey = new QShortcut(QKeySequence(Qt::Key_Escape), this);
 	connect(closeKey, &QShortcut::activated, plannerModel, &DivePlannerPointsModel::cancelPlan);
 
+	QShortcut *mirrorProfileShortcut = new QShortcut(QKeySequence(Qt::Key_M + Qt::CTRL), this);
```
mikeller, line 81: Making this available only as a shortcut creates the problem that there will be no way for users to discover that this functionality exists, and consequently they will not be using it.

The fact that you have added a mention to the user manual is a good start, but probably does little to address the issue:
- not many users read 5000 lines of user manual for every release to discover new features hidden behind shortcuts;
- the user manual only exists in English - effectively we are shutting out any non English speakers from ever using this (and to make matters worse, for them `[Ctrl] - M` most likely isn't intuitive or easy to remember).

In order to create a User Experience that allows users to discover and use this, I think at least one of the following should be considered:
- adding a button for it (obviously clutters the UI, but provides a better UX);
- adding an entry for it to the menubar (a bit problematic as at the moment the menubar is not updated when the planner is opened, and therefore doesn't have much to provide to users while in the planner;
- adding a tooltip that is showing when the keyboard shortcut can be used (not great UX, but better than not allowing users to discover that the functionality exists.

In addition to this, an entry should be added to the [changelog](https://github.com/subsurface/subsurface/blob/master/CHANGELOG.md), giving the user the information required to find this shortcut (again only a limited mitigation, as not many users read the changelog, and even fewer remember it when they get to the point of wanting to use the functionality).

atdotde:
I like this. It was my suggestion to add this feature only with a key stroke binding as I would expect it to be used only by a tiny fraction of users. So blame me. But we would add a button next to the plus in the waypoint list.

I have made your second commit look a bit more Qt-ish and removed the helper function. Have a look at https://github.com/subsurface/subsurface/compare/master...atdotde:subsurface:plannerReverseProfile
What do you think?

jwobser:
@atdotde 
Oh that's a fun overload of '<<' for QVector, I assume it just does a push_back?

I like eliminating the helper function as you have done.

I'm not familiar with the preferences configuration, what is happening on with this line? 
	`divepoints.back().time = runtime + divepoints[0].time - divepoints[0].depth.mm / prefs.descrate;`

And 1 concern, I think you need to subtract 2 from the last row index here:
`	beginInsertRows(QModelIndex(), divepoints.size(), 2 * divepoints.size());`

[QAbstractItemModel uses zero based indexing](https://doc.qt.io/qt-5/qabstractitemmodel.html#beginInsertRows), and also we are actually inserting 1 fewer point than the current length of divepoints. The last point in the profile becomes the pivot, and then we add [n-1] additional points. See below, the 4 points add 3 to become 7.

[0, 1, 2, 3] and we are inserting [4, 5, 6]
So first inserted index would be divepoints.size() [4] and last would be [ 2 * 4 - 2].

When I was playing around, if I had the wrong last row index, there would be uninitialized rows in the tableView.

I haven't tried to compile your branch yet. 
Before
After

mikeller:
> But we would add a button next to the plus in the waypoint list.

Agreed.

jwobser:
@atdotde I figured out that was the descent rate, so you are adding one final point that should be the amount of time it takes  to drop to depth. This makes mirroring behave nicely with the "Drop to first depth" option. 

Just had to tweak 2 lines, but your terse solution works great!

Because we are adding one final point, only need to subtract 1 from the last row index.
`	beginInsertRows(QModelIndex(), divepoints.size(), 2 * divepoints.size() - 1);`
There was a subtract here for the descent time, but I believe it should have been an addition.
`	divepoints.back().time = runtime + divepoints[0].time + divepoints[0].depth.mm / prefs.descrate;`


I tried to find where the add row button lives in the code, but couldn't. 
I would be happy to implement if a Qt guru could point me in the right direction. I've been opening the .ui files individually, I assume there is a Qt Designer project hidden somewhere that would make this all easier.

atdotde:
You are of course right about the miscounting of the index of the final row (it took me several attempts to get the logic right end then in the end missed the "the pivot is not copied" thing). I pushed an updated version and also made it take into account if the the drop like a stone mode is actually used. 

Subtracting the descent time is correct IMHO: The stay at the first depth takes runtime minus the time to get there.

The "button like the add button" thing I did not do because it was getting too late yesterday. The add button is a property of our tableview (see tableview.cpp lines 48ff). Of course the mirror button should only be added in the planner and we need a reasonable icon (searching for the icon was the point where I actually gave up last night).

mikeller:
> The "button like the add button" thing I did not do because it was getting too late yesterday. The add button is a property of our tableview (see tableview.cpp lines 48ff). Of course the mirror button should only be added in the planner and we need a reasonable icon (searching for the icon was the point where I actually gave up last night).

I actually had an idea for adding a standardised way for adding non-mainstream functionality into a dynamic drop down menu, which may be preferable to adding a button - I just didn't get around to seeing if I can get it to work. If you give me a day or so I can put it into a pull request to discuss.

jwobser:
I mocked up these icons, inspired by the ruler icon, and using some subsurface colors.

planner_mirror_profile   planner_mirror_profile

Edit: Significantly thicker versions show up better at 23x23.
planner_mirror_profile   planner_mirror_profile

Alternate Colors
planner_mirror_profile  planner_mirror_profile   
 planner_mirror_profile planner_mirror_profile
planner_mirror_profile   planner_mirror_profile

atdotde:
I added the implementation of  the button to my branch (using your icon).

jwobser:
@atdotde Your branch works great! Would you like to make a pull from your branch? Or should I add a refactor commit here?

This worked for setting a tool tip (I didn't add a translation with the tr() function yet).

`ui.tableWidget->mirrorButton()->setToolTip("Mirror Dive Profile");`

And also I added a cap on the plan length (100 might be too low) but with exponential growth you can make subsurface very unhappy very quickly.
`if (divepoints.size() <= 1 || divepoints.size() > 100)
return;
`

atdotde:
I added the tooltip and your signed-off-by to my branch (if you agree with that) and would propose to turn that into a pull request to be merged. I did not take the restriction of the number of dive points and would rather allow the user to shoot themselves in the foot instead of introducing such arbitrary boundaries (which can easily be reached for example when editing a real dive in the planner). It's not that something bad happens, just Subsurface can become slow.

jwobser:
Refactor by @atdotde in new pull request. #3843

PR #3843, opened 2023-03-12
Planner reverse profile

atdotde:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In the planner, allow the user to mirror the profile, i.e. add copies of the waypoints in reverse order. The UI is a second button on top of the dive planner table and the Ctrl-M keyboard shortcut.

This PR supersedes #3842 and implements a suggestion by Alvaro A. on the google group.

### Documentation change:

Added explanation to user manual and CHANGELOG entry.

### Mentions:

Builds on contributions by @jwobser


subsurface.qrc
```diff
@@ -6,6 +6,7 @@
         <file alias="subsurface-icon">icons/subsurface-icon.png</file>
         <file alias="subsurface-mobile-icon">icons/subsurface-mobile-icon.png</file>
         <file alias="list-add-icon">icons/plus.png</file>
+        <file alias="mirror-icon">icons/mirror-icon.png</file>
```
mikeller, line 9: Is the `mirror-icon.svg` file actually needed, seeing that we are only referencing the `mirror-icon.png` file here?

atdotde: Yes, the svg as a vector format is the "source" for the image which is needed if you later want to change the icon or derive others from it.

mikeller: Cool, makes sense.
So if `mirror-icon.png` is just an artefact produced from `mirror-icon.svg`, wouldn't it make more sense to not store the artefact (like we don't store the `*.o` files produced as artefacts during compilation) and instead get the build process to produce the artefacts during the build?

atdotde: Dunno. I would do the conversation with inkscape which is a GUI tool. Ymmv but I would hate it if during build a GUI app would start up and I would have to click something like "export to png". We do similar things for the user manual where we keep the html around that as also an artifact.

bstoeger: You do realize that there is a `renderSVGIcon` function in `qthelper.cpp`?

mikeller: Inkscape actually does work on the command line as well: https://wiki.inkscape.org/wiki/Using_the_Command_Line
But I would probably rather use something like `convert` (based on [ImageMagick](https://imagemagick.org/index.php), which is intended for this kind of use and has a powerful set of command line options to control the properties of the output. And it's very common, so probably already installed on most developers' workstations.

desktop-widgets/tableview.h
```diff
@@ -45,7 +47,7 @@ class TableView : public QGroupBox {
 
 private:
 	Ui::TableView ui;
-	QPushButton *plusBtn;
+	QPushButton *plusBtn, *mirrorBtn;
```
bstoeger, line 50: Now that looks like a layering violation to me. Why would the general `TableView` class have a "mirror button"!?

subsurface.qrc
mikeller: I had a look into this, and it looks like SVG files can be used directly as the source for `QIcon` - in fact it this is already done in some places:

https://github.com/subsurface/subsurface/blob/d44787a0b33d96541c34b218f40ed69f58c02794/qt-models/divetripmodel.cpp#L198

I have opened a pull request to refactor the existing cases: #3844.

bstoeger: Hmmm... I think the quoted code should be changed to
```
static QPixmap &getGlobeIcon()
{
	static std::unique_ptr<QPixmap> icon = []() {
		const IconMetrics &im = defaultIconMetrics();
		return std::make_unique<QPixmap>(QIcon(":globe-icon").pixmap(im.sz_small, im.sz_small));
	}();
	return *icon;
}
```
That said, we had crashes when freeing Qt resources after main() exited. Because Qt memory management is the worst.

qt-models/diveplannermodel.cpp
dirkhh, line 1: oh I like seeing changes like this... there are so many redundant things that we keep doing where we call functions repeatedly instead of calling them once in the very end.

Documentation/user-manual.txt
dirkhh, line 1: extra credit for editing the user manual!!

desktop-widgets/tableview.h
dirkhh: I would prefer if this was addressed (or at least there was a response to the comment that explained why Berthold and I are wrong here)

atdotde: Hmm. Adding the button only in the planner is painful (you need to tell it where to go), in particular due to the non-standard positioning (top right) that needs to keep adjusting when the table is resized.

If I understand correctly, the proper way to handle this would be to derive a class PlannerTableView from TableView that has this additional button. Do you really want me to do this (rather than this hack of an invisible button that only becomes visible when the TableView is used in the planner)?

mikeller: I am currently looking into adding an 'Actions' menu that then can contain less often used functionality that is currently hard to find because it is behind keyboard shortcuts and context menus.
I think this will be a good candidate for this 'Actions' menu, which will remove the need for a button and subclassing of TableView.

bstoeger: > If I understand correctly, the proper way to handle this would be to derive a class PlannerTableView from TableView that has this additional button. Do you really want me to do this (rather than this hack of an invisible button that only becomes visible when the TableView is used in the planner)?

Either that or add a general `addButton` function that takes an icon and a callback. I can do this, if you want.

That said, I am not a fan of the whole `TableView` class at all.

PR #3844, opened 2023-03-14
Desktop: Use SVG directly for icons.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Use SVG files as source for icons where they exist, and remove the respective PNG artefacts generated from the SVG files from the repository.
This has only been done for the desktop version, as I don't know how good SVG support is on the mobile platforms.

### Changes made:

1. update references in resource files to point to SVG icon files where they exist;
2. remove PNG files generated from the SVG files above;
3. fold `warning.png` into `warning-icon.png` for consistency of UX and simplification;
4. replace use of `renderSVGIcon()` and `renderSVGIconWidth()` with direct loading of QPixmaps and scaling;
5. add workaround for video overlay scaling with a custom aspect ratio;
6. remove `renderSVGIcon()` and `renderSVGIconWidth()`.

### Mentions:


mikeller:
@bstoeger: In a way it looks to me like `renderSVGIcon()` could potentially be retired and the respective `QIcon` instances directly created from the SVG files - what are your thoughts on this?

https://github.com/subsurface/subsurface/blob/d44787a0b33d96541c34b218f40ed69f58c02794/core/qthelper.cpp#L1688-L1709

bstoeger:
> In a way it looks to me like renderSVGIcon() could potentially be retired and the respective QIcon instances directly created from the SVG files - what are your thoughts on this?

Less code is always better. If you can remove these, that's great.

mikeller:
> Less code is always better. If you can remove these, that's great.

Agreed. I'll see what I can do.

dirkhh:
This is definitely going in the right direction. And I agree that reducing code is overall a good thing.
Way way back when there were issues with SVGs, but this hasn't been true for a long time which is why I think some of the newer icons were only in svg in the first place.

mikeller:
@dirkhh: Still looking into this. One 'side effect' of `renderSVGIcon` is that the icon can be rendered in any given size. If the size that is needed is different from the size stored in the SVG then this will avoid artefacts from scaling. So I think `renderSVGIcon` will still be needed in some cases where the icon size is dynamic. I am still working out where this is, and will probably propose a rename to `renderScaledSvg` or similar to make it more apparent what the purpose of this is.

mikeller:
@dirkhh: This is ready now.

dirkhh (approved):
Thanks - looks reasonable

PR #3846, opened 2023-03-17
Desktop: Add option to force overwrite firmware for OSTC4.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In order to support development of the open source firmware of the OSTC4.
The checkbox is only enabled if the selected dive computer model is OSTC4.

Also reorder the buttons in the 'Configure dive computer' dialogue to group them into 'firmware update' and 'configuration management' functions.

Requires https://github.com/subsurface/libdc/pull/47 to be merged due to an API change.

### Changes made:

1. add `forceUpdate` to `startFirmwareUpdate()`;
2. reorder buttons in 'Configure dive computer' dialogue;
3. add 'Force overwrite firmware' button to the 'Configure dive computer' dialogue.

### Related issues:

https://github.com/subsurface/libdc/pull/47

### Additional information:

Screenshot from 2023-03-17 20-14-54

### Release note:

desktop: add option to force firmware update on OSTC4

### Mentions:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 core: fix bug when save sea water salinity given by DC
+desktop: add option to force firmware update on OSTC4
 desktop: add column for dive notes to the dive list table
 desktop: Add an option for printing in landscape mode
 desktop: fix bug when printing a dive plan with multiple segments
--- a/core/configuredivecomputer.cpp
+++ b/core/configuredivecomputer.cpp
@@ -491,13 +491,13 @@ bool ConfigureDiveComputer::restoreXMLBackup(QString fileName, DeviceDetails *de
 	return true;
 }
 
-void ConfigureDiveComputer::startFirmwareUpdate(QString fileName, device_data_t *data)
+void ConfigureDiveComputer::startFirmwareUpdate(QString fileName, device_data_t *data, bool forceUpdate)
 {
 	setState(FWUPDATE);
 	if (firmwareThread)
 		firmwareThread->deleteLater();
 
-	firmwareThread = new FirmwareUpdateThread(this, data, fileName);
+	firmwareThread = new FirmwareUpdateThread(this, data, fileName, forceUpdate);
 	connectThreadSignals(firmwareThread);
 
 	firmwareThread->start();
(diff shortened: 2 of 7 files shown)


atdotde:
I had to look at the source of the libdc update to understand what this is doing. Maybe call it "force reinstall|" rather than "force overwrite". Am I correct that this is supposed to treat the case where you want to replace the firmware with a new firmware that has the same version string but not when something went wrong with the upload (the other condition of the if)? Another approach would have been to open a modal dialog when you encounter the version strings agree rather than a checkbox beforehand.

mikeller:
@atdotde: Thanks! I think 'Force reinstall firmware' is better. Or should it be 'Ignore firmware version'?
Yes, forcing a clean state after a problematic update is one use case. Overwriting firmware with a new build with the same version number during firmware testing is the second use case - after all the OSTC4 firmware is open source: https://code.heinrichsweikamp.com/public/ostc4/file/tip
The confirmation dialogue is another possible approach, but this would require a more drastic change in the libdivecomputer API, as currently libdivecomputer reports 'ok' even if some or all parts of the firmware are skipped during an update. (The OSTC4 firmware consists of 3 different parts that can be updated independently and come with their own version.) Furthermore, some parts of the firmware (e.g. the fonts) are only changing infrequently, so popping up a dialogue about reinstalling them for all users for every firmware upgrade would probably create more annoyance than value.

dirkhh:
the builds all fail ... would you mind fixing this ?

Thanks

mikeller:
@dirkhh: The builds are failing because the change depends on the libdivecomputer API extension proposed in https://github.com/subsurface/libdc/pull/47. My proposed approach is to wait until this has been reviewed and merged, and then update the submodule in the pull request to the latest commit, which will make it build.
I _could_ make this build by changing the submodule to point to my branch, but then that's not what you'd want to merge. :-)

dirkhh:
Ahh - that's where my being gone is a problem... Linus never sees PRs without a gentle nudge... I'll do that right now

mikeller:
Merged in #3855, closing this.

PR #3854, opened 2023-03-29
Desktop: Fix salinity combo/icon when DC doesnt have salinity info

rafsalvioni:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Today salinity combo is editable if one of these rules matches: The dive was manually entered or if salinity edition is allowed in preferences. However we can have cases that dives were downloaded but its doesn't have salinity info.

This fix considers if there's a DC salinity info to decides combo edition and if salinity change indicator will be showed or not.
If DC doesn't have salinity, the UI behavior is the same of a manual dive. Its important because after PR #3852 was accepted, too many existant download dives will be inconsitents.

### Changes made:
Basically just change TabDiveInformation and fix whole salinity references.

### Additional information:
To exemplify, get a downloaded dive and open it with and without tag `//dive/divecomputer/water`, like this:
`<water salinity='1030 g/L' />`

### Mentions:
@mikeller

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveInformation.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveInformation.cpp
@@ -45,11 +45,10 @@ TabDiveInformation::TabDiveInformation(MainTab *parent) : TabBase(parent), ui(ne
 	connect(action, &QAction::triggered, this, &TabDiveInformation::closeWarning);
 	ui->multiDiveWarningMessage->addAction(action);
 	ui->multiDiveWarningMessage->hide();
-	manualDive = false;
 	updateWaterTypeWidget();
 	QPixmap warning (":salinity-warning-icon");
 	ui->salinityOverWrittenIcon->setPixmap(warning);
-	ui->salinityOverWrittenIcon->setToolTip("Water type differs from that of dc");
+	ui->salinityOverWrittenIcon->setToolTip("Water type differs from that of DC(s)");
 	ui->salinityOverWrittenIcon->setToolTipDuration(2500);
 	ui->salinityOverWrittenIcon->setVisible(false);
 }
@@ -94,9 +93,10 @@ void TabDiveInformation::closeWarning()
 }
 
 void TabDiveInformation::updateWaterTypeWidget()
-{    // Decide on whether to show the water type/salinity combobox or not
-	if (prefs.salinityEditDefault || manualDive)
-	{         // if the preference setting has been checked or this is a manually-entered dive
+{
+	// Decide on whether to show the water type/salinity combobox or not
+	bool hasDCSalinity = parent.currentDive->salinity != 0;
+	if (prefs.salinityEditDefault || !hasDCSalinity) {   // if the preference setting has been checked or DC doesnt have salinity info
 		ui->waterTypeText->setVisible(false);
 		ui->waterTypeCombo->setVisible(true); // show combobox
 	} else {  // if the preference setting has not been set
@@ -180,16 +180,14 @@ int TabDiveInformation::updateSalinityComboIndex(int salinity)
 		return SALTWATER;
 }
 
-// If dive->user_salinity != dive->salinity (i.e. dc value) then show the salinity-overwrite indicator
+// If dive->user_salinity != dive->salinity then show the salinity-overwrite indicator
 void TabDiveInformation::checkDcSalinityOverWritten()
 {
-	const struct divecomputer *currentdc = parent.getCurrentDC();
-	if (!currentdc)
-		return;
-	int dc_value = currentdc->salinity;
-	int user_value = parent.currentDive->user_salinity;
+	dive *currentDive = parent.currentDive;
+	int dc_value = currentDive->salinity;
+	int user_value = currentDive->user_salinity;
 	bool show_indicator = false;
-	if (!manualDive && user_value != 0 && user_value != dc_value)
+	if (dc_value != 0 && user_value != 0 && user_value != dc_value)
 		show_indicator = true;
 	ui->salinityOverWrittenIcon->setVisible(show_indicator);
 }
@@ -207,7 +205,6 @@ void TabDiveInformation::showCurrentWidget(bool show, int position)
 void TabDiveInformation::updateData(const std::vector<dive *> &, dive *currentDive, int currentDC)
 {
 	int salinity_value;
-	manualDive = is_manually_added_dc(&currentDive->dc);
 	updateWaterTypeWidget();
 	updateProfile();
 	updateWhen();
@@ -217,7 +214,7 @@ void TabDiveInformation::updateData(const std::vector<dive *> &, dive *currentDi
 	ui->atmPressType->setCurrentIndex(0);                // Set the atmospheric pressure combo box to mbar
 	salinity_value = get_dive_salinity(currentDive);
 	if (salinity_value) {			// Set water type indicator (EN13319 = 1.020 g/l)
-		if (prefs.salinityEditDefault) {   //If edit-salinity is enabled then set correct water type in combobox:
+		if (ui->waterTypeCombo->isVisible()) {   // If water salinity is editable then set correct water type in combobox:
 			ui->waterTypeCombo->setCurrentIndex(updateSalinityComboIndex(salinity_value));
 		} else {         // If water salinity is not editable: show water type as a text label
 			ui->waterTypeText->setText(get_water_type_string(salinity_value));
@@ -252,7 +249,7 @@ void TabDiveInformation::updateUi(QString titleColor)
 	ui->groupBox_wavesize->setStyleSheet(ui->groupBox_wavesize->styleSheet() + CSSSetSmallLabel);
 	ui->groupBox_surge->setStyleSheet(ui->groupBox_surge->styleSheet() + CSSSetSmallLabel);
 	ui->groupBox_chill->setStyleSheet(ui->groupBox_chill->styleSheet() + CSSSetSmallLabel);
-	ui->salinityOverWrittenIcon->setToolTip(ui->salinityOverWrittenIcon->styleSheet() + CSSSetSmallLabel);
+	ui->salinityOverWrittenIcon->setStyleSheet(ui->salinityOverWrittenIcon->styleSheet() + CSSSetSmallLabel);
 }
 
 // From the index of the water type combo box, set the dive->salinity to an appropriate value
@@ -260,10 +257,7 @@ void TabDiveInformation::on_waterTypeCombo_activated(int index)
 {
 	Q_UNUSED(index)
 	int combobox_salinity = 0;
-	const struct divecomputer *currentdc = parent.getCurrentDC();
-	if (!currentdc)
-		return;
-	int dc_salinity = currentdc->salinity;
+	int dc_salinity = parent.currentDive->salinity;
 	switch(ui->waterTypeCombo->currentIndex()) {
 	case FRESHWATER:
 		combobox_salinity = FRESHWATER_SALINITY;
@@ -292,8 +286,8 @@ void TabDiveInformation::on_waterTypeCombo_activated(int index)
 	else
 		ui->salinityText->clear();
 	divesEdited(Command::editWaterTypeUser(combobox_salinity, false));
-	// If salinity differs from that downloaded from dc, show warning
-	if (manualDive || dc_salinity == combobox_salinity)
+	// If choosed salinity differs from dive->salinity (if it has one), show warning
+	if (dc_salinity == 0 || dc_salinity == combobox_salinity)
 		ui->salinityOverWrittenIcon->setVisible(false);
 	else
 		ui->salinityOverWrittenIcon->setVisible(true);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ desktop: add support for multiple tanks to the profile ruler
 export: change format produced by 'CSV summary dive details' from TSV (tab separated) to CSV
 desktop: add function to merge dive site into site selected in list
 import: add option to synchronise dive computer time when downloading dives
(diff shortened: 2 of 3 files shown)


desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -220,7 +223,7 @@ void TabDiveInformation::updateData()
 	ui->atmPressType->setCurrentIndex(0);                // Set the atmospheric pressure combo box to mbar
 	salinity_value = get_dive_salinity(current_dive);
 	if (salinity_value) {			// Set water type indicator (EN13319 = 1.020 g/l)
-		if (prefs.salinityEditDefault) {   //If edit-salinity is enabled then set correct water type in combobox:
+		if (ui->waterTypeCombo->isVisible()) {   // If water salinity is editable then set correct water type in combobox:
```
mikeller, line 217: Looks like this is also fixing a bug here, where the salinity isn't set correctly for manual dives. Good find!

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -255,7 +258,7 @@ void TabDiveInformation::updateUi(QString titleColor)
 	ui->groupBox_wavesize->setStyleSheet(ui->groupBox_wavesize->styleSheet() + CSSSetSmallLabel);
 	ui->groupBox_surge->setStyleSheet(ui->groupBox_surge->styleSheet() + CSSSetSmallLabel);
 	ui->groupBox_chill->setStyleSheet(ui->groupBox_chill->styleSheet() + CSSSetSmallLabel);
-	ui->salinityOverWrittenIcon->setToolTip(ui->salinityOverWrittenIcon->styleSheet() + CSSSetSmallLabel);
+	ui->salinityOverWrittenIcon->setStyleSheet(ui->salinityOverWrittenIcon->styleSheet() + CSSSetSmallLabel);
```
mikeller, line 252: Another bug. Good find!

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -93,9 +93,12 @@ void TabDiveInformation::closeWarning()
...
 
 void TabDiveInformation::updateWaterTypeWidget()
-{    // Decide on whether to show the water type/salinity combobox or not
-	if (prefs.salinityEditDefault || manualDive)
-	{         // if the preference setting has been checked or this is a manually-entered dive
+{
+    struct divecomputer *currentdc = get_dive_dc(current_dive, dc_number);
+    bool hasDCSalinity = currentdc && currentdc->salinity != 0;
+    // Decide on whether to show the water type/salinity combobox or not
+	if (prefs.salinityEditDefault || manualDive || !hasDCSalinity)
```
mikeller: I think you can simplify this by getting rid of `manualDive` here (and replace it with `!hasDcSalinity` in the other places where it is used), as `hasDcSalinity` will evaluate to false for manual dives.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -93,9 +93,12 @@ void TabDiveInformation::closeWarning()
 }
 
 void TabDiveInformation::updateWaterTypeWidget()
-{    // Decide on whether to show the water type/salinity combobox or not
-	if (prefs.salinityEditDefault || manualDive)
-	{         // if the preference setting has been checked or this is a manually-entered dive
+{
+    struct divecomputer *currentdc = get_dive_dc(current_dive, dc_number);
+    bool hasDCSalinity = currentdc && currentdc->salinity != 0;
```
mikeller: This will lead to weird results for dives with multiple dive computers (which is common in technical diving), where only some of the computers report salinity:

image
(full disclosure: this import was fudged, the OSTC4 _does_ report salinity)

image

So a better solution will be to check all dive computers used on the dive, and only allow editing if none of them report salinity.

mikeller (approved):
To me it looks like there is no 'right' answer to the question if the preference for showing the salinity edit should be applied to the case when no salinity is provided by the dive computer.
But I guess the fact that we are already overriding the preference for manually edited dives can be taken as this being the right approach when no salinity is reported by the dive computer(s).

desktop-widgets/tab-widgets/TabDiveInformation.cpp
rafsalvioni: >as hasDcSalinity will evaluate to false for manual dives.

Why? Because manual dives doesnt have "dc" or beacuse its doesnt have salinity? Or both?

desktop-widgets/tab-widgets/TabDiveInformation.cpp
rafsalvioni: Im not sure this is better approach... Because the issue is not only enable combo or not... We have to work with salinity levels to show exclamation icon...

In case of more than 1 DC, what we should consider?

So, we have 3 solutions (I think):
1- Let the current way... Its considers the current DC selected;
2- Use always main DC (the first one). This indicates user's DC preference (this looks like better to me);
3- A mixed of all DCs registered in dive (looks like temperatures below...). Here, we have to choose what approach to calc this;

https://github.com/subsurface/subsurface/blob/eebb47ec22aafa34e29cd0385853fde10fe208ab/core/divecomputer.c#L346-L381

So I think this is a longer discussion and we can leave it the way it is while this idea is debated

desktop-widgets/tab-widgets/TabDiveInformation.cpp
rafsalvioni: image

desktop-widgets/tab-widgets/TabDiveInformation.cpp
rafsalvioni: About is:
https://github.com/subsurface/subsurface/blob/eebb47ec22aafa34e29cd0385853fde10fe208ab/desktop-widgets/tab-widgets/TabDiveInformation.cpp#L182-L193

What would be "dc_value" when more than 1 DC?

rafsalvioni: Correcting... Has already a merge between DCs salinity...
https://github.com/subsurface/subsurface/blob/eebb47ec22aafa34e29cd0385853fde10fe208ab/core/dive.c#L824-L839

So we can use dive->salinity instead dc->salinity. I did it in https://github.com/subsurface/subsurface/pull/3854/commits/526168e048ac5d6dd956374c4a251464edba4767

desktop-widgets/tab-widgets/TabDiveInformation.cpp
rafsalvioni: Done! https://github.com/subsurface/subsurface/pull/3854/commits/bad184cb0dd640e0d68f0c95aa021dfbb996efb9

desktop-widgets/tab-widgets/TabDiveInformation.cpp
mikeller: I am not sure that showing the user an average between the configured salinities on different dive computers is the right approach here. Using this average as the base for calculations may make sense, but showing it to the user is probably more confusing than helpful.
Salinity is different from measured dimensions like temperature or pressure - for these, differences between dive computers on the same dive are likely down to sensor tolerances, so it makes sense to use an average. For salinity, the user will always have set his dive computer(s) to a specific salinity, and if they differ between computers then there is a reason for it (potentially configuration error) - and not that they want to use an average.
So I think if we display a calculated average for salinity here we should at least use some form of indication (maybe a warning) to indicate to the user that this is happening, or we will be getting questions from users as to why we are showing a salinity that they haven't configured anywhere on any of their dive computers - or potentially show 'undefined' for salinity if this is the case.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -295,8 +291,8 @@ void TabDiveInformation::on_waterTypeCombo_activated(int index)
 	else
 		ui->salinityText->clear();
 	divesEdited(Command::editWaterTypeUser(combobox_salinity, false));
-	// If salinity differs from that downloaded from dc, show warning
-	if (manualDive || dc_salinity == combobox_salinity)
+	// If salinity differs from that downloaded from dc (if it has one), show warning
```
mikeller: This comment is now incorrect - we are not comparing the salinity to something that was downloaded from a dive computer, but to a calculated average.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
rafsalvioni: Well... Once I found function "fixup_water_salinity ", I thought that was a good solution for the issue...

> So I think if we display a calculated average for salinity

I did tests to understand it and, when we have more than 1 DC with salinity, the value showed is the result of fixup_water_salinity. So I conclude the salinity's average is used today and your concern is already real.
image
DC 1: `<water salinity='1030 g/l' />`
DC 2: `<water salinity='1040 g/l' />`

However, once it is used (and users know that), I believe this approach still the best at moment and we can open for discussion to find a better. Or we find a way to "mix" salinities ou we should to use just one of them, from current DC or from main one

desktop-widgets/tab-widgets/TabDiveInformation.cpp
rafsalvioni: Done https://github.com/subsurface/subsurface/pull/3854/commits/08329c43b25b3866bb03c117618e1918e3997395

desktop-widgets/tab-widgets/TabDiveInformation.cpp
mikeller: Yes, you are right. What you are doing is merely fixing the checks around the displayed salinity to be based on what is displayed and not around the current dive computer's reported salinity. The current behaviour seems to be wrong no matter what, and you are fixing it.

mikeller:
Looks like something went wrong with your latest rebase, and this is now including some changes that are already in `master` - can you please fix this?

rafsalvioni:
@mikeller I did something too bad in my branch when I tried a rebase... Im going to try fix it... My Git's knowledge is outdate...

mikeller (approved):
Looks much better now!

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -93,9 +92,11 @@ void TabDiveInformation::closeWarning()
 }
 
 void TabDiveInformation::updateWaterTypeWidget()
-{    // Decide on whether to show the water type/salinity combobox or not
-	if (prefs.salinityEditDefault || manualDive)
-	{         // if the preference setting has been checked or this is a manually-entered dive
+{
+    bool hasDCSalinity = current_dive && current_dive->salinity != 0;
+    // Decide on whether to show the water type/salinity combobox or not
+	if (prefs.salinityEditDefault || !hasDCSalinity)
```
bstoeger: Coding style: brace on same line as `if`.
I reckon that the code was like that before, but since you're changing it, might as well fix this.

bstoeger:
Is there consensus that this is the right thing to do? It looks reasonable to me.

desktop-widgets/tab-widgets/TabDiveInformation.cpp
rafsalvioni: Done! https://github.com/rafsalvioni/subsurface/commit/9334c8b1d45eaebd7706c1d7b605cd414906e8b7

mikeller:
> Is there consensus that this is the right thing to do? It looks reasonable to me.

@bstoeger: I think the absence of comments can be taken as consensus. ;-)

@rafsalvioni: This needs a rebase before it can be merged.

mikeller (approved):

bstoeger:
Something is really strange with the history here. I cannot follow it.

Could you rebase on current master, so that we can merge?

mikeller:
@rafsalvioni: I think there is a bit of a misunderstanding of what a 'rebase' is. What you are doing is mergeing the latest on `master` into your branch, which is _not_ a rebase.

To get a better idea i recommend reading https://docs.gitlab.com/ee/topics/git/git_rebase.html for a relatively succinct introduction to rebasing with git (obviously skip the bits about the gitlab UI). An important point to read is the bit about the force push at the very bottom which is crucial to get your local changes pushed to the remote.

mikeller:
Much better now. :100:

rafsalvioni:
@mikeller thanks for the clarification! my knowledge in Git is (and now "was") a little poor...

desktop-widgets/tab-widgets/TabDiveInformation.cpp
```diff
@@ -94,9 +93,10 @@ void TabDiveInformation::closeWarning()
 }
 
 void TabDiveInformation::updateWaterTypeWidget()
-{    // Decide on whether to show the water type/salinity combobox or not
-	if (prefs.salinityEditDefault || manualDive)
-	{         // if the preference setting has been checked or this is a manually-entered dive
+{
+	// Decide on whether to show the water type/salinity combobox or not
+	bool hasDCSalinity = parent.currentDive->salinity != 0;
```
mikeller, line 98: @rafsalvioni: This is causing SEGV crashes on startup in some circumstances - looks like `currentDive` needs to be null checked here.

rafsalvioni: @mikeller I fixed in https://github.com/rafsalvioni/subsurface/commit/865a37de90a06f0c6c4e8b83abfcfb849e2abc35

Are you going to reopen PR or I need to create a new one?

mikeller: I don't think there is a way to re-open a pull request once it's been merged, so please open a new one.

PR #3855, opened 2023-03-29
Ostc4

dirkhh:
combine #3846 with the libdc change in https://github.com/subsurface/libdc/pull/47


mikeller:
I have rebased #3846 to get rid of the merge conflicts (`CHANGELOG.md`), but I think you need to amend this to get the conflict-free version.

PR #3861, opened 2023-04-15
Build script: Qt has tagged versions currently

salvadorcunat:
It looks like Qt company has LGPLed versions tagged wich simplifies things a bit while building, e.g. 5.15.3 current workaround matches "v5.15.3-lts-lgpl" tag.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [X] Build system change
- [ ] Documentation change
- [ ] Language translation

I don't know if tags where in place when @dirkhh made the workaround and wrote the note on the script, but they are now, thus workaround for 5.15.3 is no more needed and the other 5.15.X versions are reachable via git tags.

Background: Debian Sid is currently at Qt 5.15.8 which is impossible to build from scratch with current script as only a few git versions are tagged in the format "v5.15.8" used in the script.

flameshot_2023-04-15_18-58-22

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -293,16 +293,8 @@ if [ "$PLATFORM" = Linux ] && [[ $QT_VERSION == 5* ]] ; then
 		rm -rf "$INSTALL_ROOT"/include/QtLocation > /dev/null 2>&1
 		rm -rf "$INSTALL_ROOT"/include/QtPositioning > /dev/null 2>&1
 
-		# thanks for nothing, Qt company...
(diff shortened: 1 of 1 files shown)


salvadorcunat:
OK. I see the problem with older versions used in some builds.  Hopefully force pushed changes will work.

dirkhh:
Sorry, I somehow completely missed this...
This seems like a very reasonable change to me. But of course I can't test it while on the road.
@salvadorcunat -- when I ignore you like this, feel free to ping me in email.

PR #3863, opened 2023-04-16
Break out qt-quick code from statistics widget

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

The long goal here is to reuse the glue layer between QtQuick and the statistics module for the profile. This glue layer tries to somewhat hide the horrible QtQuick API.

### Changes made:

1) Remove statistics-theme related code from the QtQuick code.
2) Make the number of layers (runtime) dynamic.
3) Split out QtQuick code (except for specific statistic-related ChartItems) into its own directory.

### Release note:

Nope. Hopefully not user visible.


dirkhh (commented):
I think this is a good direction. Clearly the Android build is getting messed up somehow - but it wasn't immediately obvious how... it seems to get confused linking against the chartitem...

bstoeger:
I think this has something to do with linkage-order. At least for desktop the order of linkage makes a difference. Apparently a "static library" using another one is not well defined!?

bstoeger:
@dirkhh: If you find time, I would really appreciate a comment on the pinching behavior on mobile. I can only test scroll-wheel with mobile-on-desktop (actually laptop).

glance-:
The general trick for libraries using each other are to put them in a group, allowing the linker to search them multiple times to find and fill in all the relevant symbols. This is done by --start-group archives --end-group , and some -Wl,<blah> to get the compiler to pass it to the linker. How to make qmake (news to me that we're using qmake again) do that is a different story.

You can grab the android-sdk and build tools, and run the app in a emulator, testing it in the right environment. mobile-on-desktop is way more convenient but running it on a emulator is more accurate where accuracy matters. It's a bit of a hassle setting it up if you've never done it, but it's there as a option.

dirkhh:
> @dirkhh: If you find time, I would really appreciate a comment on the pinching behavior on mobile. I can only test scroll-wheel with mobile-on-desktop (actually laptop).

I'm not quite sure what you want me to comment on... I haven't paid enough attention to this PR as I am now trying to get the Android build working with Qt6 (because, thanks, Google, in order to comply with their latest mandatory target SDK update we run into a series of dependency changes which in the end lead to something that isn't supported in the available Qt5 builds... it may be fixed in a later 5.15.x version for paying customers - but since Qt.io has decided that open source developers are not an audience they care about, we don't get working builds from them anymore - so the only option that I see is to move to Qt6).
I'm guessing that this means you can simply ignore the Android build issues for now.

Back to your question - what about the pinch zoom were you asking about?

bstoeger:
> How to make qmake (news to me that we're using qmake again) do that is a different story.

Thanks for pointing out that this is using qmake. No wonder that all my cmake changes didn't make a difference.

I still think this should be a simple link-order problem, since there should be no circular accesses. Rearranging didn't help, but maybe I can figure it out.

bstoeger:
> Back to your question - what about the pinch zoom were you asking about?

If it works, basically.. And if you find time: what you think about way the pinch events are transferred between QML and the profile widget.

dirkhh:
So I can't test any of this while traveling, really.
I love the direction we are going, but testing it without a device is a major pain.
Hey, @bstoeger - may I use any of the sponsorship money to buy you an Android phone so you can test this?
Or would this violate some oath that you have sworn to never touch a smartphone? 😎

bstoeger:
I do have an Android phone and also an Android tablet (needed it for lecturing during covid, now use it for note taking). My problem is that I don't have the mental capacity to compile for device and to move it there! Moreover, I still don't understand why the Android build fails.

dirkhh:
Creating a current Android build shouldn't be TOO hard. It won't pass the Google Play rules, but for testing on your own device it should 'just work'.
I have access to none of my usual build setup while traveling, but I'll be happy to try to set this up from scratch in order to come up with something that works for you so you can at least test and don't need to work blind.
It should be a two command sequence (that's what it is on my system at home that is safely protected behind a DMZ and a firewall). 
```
bash build.sh
adb install -r some-weird-path/something.apk
```
Then all you need to do is connect your device in 'USB debug' mode and magically it gets the test build...

dirkhh:
> I do have an Android phone and also an Android tablet (needed it for lecturing during covid, now use it for note taking). My problem is that I don't have the mental capacity to compile for device and to move it there! Moreover, I still don't understand why the Android build fails.

The Android build fails because of the missing backslash at the end of https://github.com/subsurface/subsurface/blob/309fa6a6bd646f6c6010a73aa21178b3b91aa662/Subsurface-mobile.pro#L343

dirkhh:
And now we are back to the link-ordering issues.
I won't have time to dig into this while traveling - but I'll add this to the top of my list the moment I have a block of time.
I vaguely remember that early on we had a similar problem and that the fix was simple... I tried to find it in the git log but apparently I didn't use any of the keywords that I thought of searching for in the commit message :(

mikeller:
@bstoeger: Happy to help out here, I've got a working android build system and a phone to test on. Let me know what I should be testing.
Getting a containerised build system for android to work should be relatively straightforward with only a dependency on a working docker environment, just copy the snippet from https://github.com/subsurface/subsurface/tree/master/packaging/android#in-a-container into a script and run it (also compiles the libraries only on the first run, so repeated runs are down to less than one third of the time of the first run).

dirkhh:
We now have working Android builds again and can test stuff.

Not sure where to pick up the pieces for things like this.

bstoeger:
First, I will have to rebase on master and include the recent changes to the chart code.

Then, let's see how we go from there...

bstoeger:
Rebased on master, but can't force-push:
```
~/src/subsurface$ git push -f
remote: Personal access tokens (classic) are forbidden from accessing this repository.
fatal: unable to access 'https://github.com/subsurface/subsurface.git/': The requested URL returned error: 403
```
Sigh.

dirkhh:
> Rebased on master, but can't force-push:
> 
> ```
> ~/src/subsurface$ git push -f
> remote: Personal access tokens (classic) are forbidden from accessing this repository.
> fatal: unable to access 'https://github.com/subsurface/subsurface.git/': The requested URL returned error: 403
> ```
> 
> Sigh.

Hmm. Why aren't you pushing via ssh? Add a public key to your account and then push to git@github.com:subsurface/subsurface -- I have never pushed to the https endpoint because authentication is so fragile.

bstoeger:
Implemented the "ruler item". That means that the features of the basic profile are all ported (but untested). The worst part is yet to come: edit and plan modes.

dirkhh:
It looks like some of the changes to the cmake file haven't be tracked in the qmake .pro file... that appears to be why the builds are failing.

bstoeger:
OMG, I completely messed up my build directory and the only solution was to re-clone the whole repository. What a pain.

bstoeger:
Also, I really wonder how one is supposed to implement the "context menu" with QtQuick!? There isn't a `contextMenuEvent` as there used to be for `QWidgets`: https://doc.qt.io/qt-6/qquickitem-members.html

dirkhh:
I of course no nothing about any of this... QtQuick seems to be primarily touch focused... is there a long press event or something?

mikeller:
> Also, I really wonder how one is supposed to implement the "context menu" with QtQuick!? There isn't a `contextMenuEvent` as there used to be for `QWidgets`: [doc.qt.io/qt-6/qquickitem-members.html](https://doc.qt.io/qt-6/qquickitem-members.html)

I think we are currently relying too much on context menus in some places. They are great for expert users because they allow for efficient workflows, but they are very hard to impossible (if undocumented) for new users to discover, and often lead to frustration and giving up.
So this might be an opportunity to add menu entries for functionality that is currently only accessible in a context menu. This will obviously also solve the problem of making these functions accessible to users on a touchscreen device.

bstoeger:
Ok, but things like "Add setpoint change", "Add bookmark" and "Split dive into two" depend on were the user clicked. How should that work with normal menu entries?

mikeller:
> Ok, but things like "Add setpoint change", "Add bookmark" and "Split dive into two" depend on were the user clicked. How should that work with normal menu entries?

These _could_ be done by having the function changing the mouse pointer and then capturing the user's next click into the profile as the location for the action.
This is probably a lot more awkward to implement than a context menu action, but it will ensure that users can find this functionality.

bstoeger:
Currently porting the show/hide/add event code.

Could someone please shed light/give opinions on this: Currently, one can hide events, but the hidden status is lost when a different dive is shown. Should the hidden status persist? Should this ultimately be integrated into the undo system and even written to disk?

Thanks.

bstoeger:
PS: Seems like a long time ago we lost the ability to globally hide events of a certain kind (see code in `core/eventname.h`) and nobody noticed!

mikeller:
> Could someone please shed light/give opinions on this: Currently, one can hide events, but the hidden status is lost when a different dive is shown. Should the hidden status persist? Should this ultimately be integrated into the undo system and even written to disk?

I can see at least three options to do this, with different implication:
- persisting the fact that the user has hidden events;
- persisting the change in the profile by removing the hidden events from the log;
- persisting the change in the profile by adding a 'hidden' flag to the events in the log.

dirkhh:
I would not remove the events from the log. I like a 'hidden' flag.

bstoeger:
This is now in a state where I would ask people to start testing it.

Note that there have been a few UI changes: showing/hiding events is via left-mouse button, not context menu, because I think that translates better to mobile.

I need help to get this to compile on Android/iOS. Mobile-on-desktop compiles for me, so I'm not sure what is wrong with those.

bstoeger:
Note: I absolutely detest this code. The memory management is completely broken. The reason is that while QtQuick appears to be "low-level" on first glance, in practice it is not. It may delete our objects at basically any time, which means that it can only work with a "heavy" object system such as QObject. Currently, all references have to be deleted manually, but in the long run, this appears to be an unmaintainable mess and we will have to implement some kind of heavier resource management, where we automatically keep track of references. Very frustrating.

dirkhh:
I am traveling and won't have time to look into this until next week.

The more I look at all of this mess, the more I wonder if Qt/QML and the way we do things really is long term maintainable.
But I don't know what to switch TO.

bstoeger:
I have looked at a number of GUI-toolkits for a different project and despite my frequent complaining, I think that Qt is still by far the best option on desktop. Yes, it is old-fashioned but at least it works and is multi-platform.

For mobile, I have no idea. The "proper" way to do QML is to code it in Javascript, but I refuse to do that. That's just no fun to me. All I can do for now is to try and make things as toolkit agnostic as possible.

bstoeger:
Ouch. Rebasing this one will be a horror show.

mikeller:
#4245 looks nightmarish too. 🙁

bstoeger:
I roughly rebased this on master. It wasn't as horrible an experience as expected, but still pretty annoying.

It would be great if we could give this some priority.

My absolute main problem: mobile-on-desktop compiles for me, but for some reason the Android  build fails presumably owing to link order issues. And I have NO idea why. I tried shuffling around the libraries to no avail :-(

bstoeger:
@mikeller: Please check if I properly rebased on #4316. Relevant commit: 2bde689b69.

bstoeger:
Jeeez! I think I found the core issue of the failing Android build: with qmake you can't have two files with the same name, because it doesn't do subdirectories. So it will link twice against the same file.

mikeller:
@bstoeger:

> @mikeller: Please check if I properly rebased on #4316. Relevant commit: [2bde689](https://github.com/subsurface/subsurface/commit/2bde689b6913c008e7c7d2437773e1d16b3dc21a).

Looks like I forgot to respond to this, but looks ok to me.

Is this ready for review and then merge?

bstoeger:
> Is this ready for review and then merge?

Definitely not ready for the user. But ready for testing / comments, yes. I have a few changes in my local branch, but the next two to three weeks no chance in working on that.

As mentioned above, I'm unhappy about the whole thing, but it is a start. Clearly, Qt wants us to do this in their JavaScript dialect with this deranged model/view thing, but I categorically refuse to do that. If someone else wants to, then be my guest!

PR #3864, opened 2023-04-18
divelog: fix erroneous use of std::move()

bstoeger:
This has to be applied to the object, not the pointer to the object. Fixes a double-free crash introduced in 8cd451f.

Alternatively, we could use std::swap() for C++98 charm and perhaps better readability for people unfamiliar with C++11. Nowadays, std::move() is more idiomatic though. Shrug.

Reported-by: Michael Keller <github@ike.ch>

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes bug introduced by #3512.

### Release note:

No, bug not in release.

### Mentions:

@mikeller

### DIFF
--- a/core/divelog.cpp
+++ b/core/divelog.cpp
@@ -40,8 +40,8 @@ divelog::divelog(divelog &&log) :
 	dives(new dive_table),
 	trips(new trip_table),
 	sites(new dive_site_table),
-	devices(std::move(log.devices)),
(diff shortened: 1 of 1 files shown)


mikeller:
Tested, confirmed working.

PR #3869, opened 2023-04-25
For discussion: Desktop: Fix build for Qt6

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This seems to be needed to make the build work with Qt 6.5 on MacOS (Qt 6 is needed to build for M1 based Macs).

Not sure if there is a better way to do this, but the implicit constructors for `QVariant` from `char *` and `const char *` have been marked as deleted in Qt 6.5.

### Changes made:

1. replace implicit `QVariant` constructors from `char *` and `const char *` with `QVariant::fromValue<>()`.

### Related issues:

Found when investigating #3868.

### DIFF
--- a/qt-models/divelocationmodel.cpp
+++ b/qt-models/divelocationmodel.cpp
@@ -91,11 +91,11 @@ QVariant LocationInformationModel::getDiveSiteData(const struct dive_site *ds, i
 	case Qt::DisplayRole:
 		switch(column) {
 		case DIVESITE: return QVariant::fromValue<dive_site *>((dive_site *)ds); // Not nice: casting away const
-		case NAME: return ds->name;
+		case NAME: return QString(ds->name);
 		case NUM_DIVES: return ds->dives.nr;
 		case LOCATION: return "TODO";
-		case DESCRIPTION: return ds->description;
-		case NOTES: return ds->name;
+		case DESCRIPTION: return QString(ds->description);
+		case NOTES: return QString(ds->name);
 		case TAXONOMY: return "TODO";
 		}
 	break;
--- a/qt-models/divesiteimportmodel.cpp
+++ b/qt-models/divesiteimportmodel.cpp
@@ -59,7 +59,7 @@ QVariant DivesiteImportedModel::data(const QModelIndex &index, int role) const
 	if (role == Qt::DisplayRole) {
 		switch (index.column()) {
 		case NAME:
-			return ds->name;
+			return QString(ds->name);
 		case LOCATION:
 			return printGPSCoords(&ds->location);
 		case COUNTRY:
@@ -70,7 +70,7 @@ QVariant DivesiteImportedModel::data(const QModelIndex &index, int role) const
 				get_dive_site_by_gps_proximity(&ds->location,
 				40075000, divelog.sites);
 			if (nearest_ds)
-				return nearest_ds->name;
+				return QString(nearest_ds->name);
 			else
 				return QString();
 		}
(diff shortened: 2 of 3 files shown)


qt-models/divelocationmodel.cpp
```diff
@@ -91,11 +91,11 @@ QVariant LocationInformationModel::getDiveSiteData(const struct dive_site *ds, i
 	case Qt::DisplayRole:
 		switch(column) {
 		case DIVESITE: return QVariant::fromValue<dive_site *>((dive_site *)ds); // Not nice: casting away const
-		case NAME: return ds->name;
+		case NAME: return QVariant::fromValue<const char *>(ds->name);
```
bstoeger: I would think `QString(ds->name)` is more readable.

bstoeger: But it doesn't work... OMG.

mikeller: > I would think `QString(ds->name)` is more readable.

I tried it that way first, but Qt 6.5.0 does not seem to like it very much:

```
[ 33%] Building CXX object desktop-widgets/CMakeFiles/subsurface_interface.dir/configuredivecomputerdialog.cpp.o
/Users/mikeller/git/subsurface/qt-models/divelocationmodel.cpp:94:21: error: functional-style cast from 'char *const' to 'QVariant' uses deleted function
                case NAME: return QVariant(ds->name);
                                  ^~~~~~~~~~~~~~~~~
/Users/mikeller/qt/6.5.0/macos/lib/QtCore.framework/Headers/qvariant.h:260:5: note: candidate constructor [with T = char *, $1 = false] has been explicitly deleted
    QVariant(T) = delete;
    ^
/Users/mikeller/qt/6.5.0/macos/lib/QtCore.framework/Headers/qvariant.h:250:24: note: candidate constructor
    QT_ASCII_CAST_WARN QVariant(const char *str) noexcept(false)
                       ^
/Users/mikeller/qt/6.5.0/macos/lib/QtCore.framework/Headers/qvariant.h:489:12: note: candidate constructor has been explicitly deleted
    inline QVariant(void *) = delete;
           ^
/Users/mikeller/qt/6.5.0/macos/lib/QtCore.framework/Headers/qvariant.h:206:5: note: candidate constructor
    QVariant(bool b) noexcept;
    ^
/Users/mikeller/qt/6.5.0/macos/lib/QtCore.framework/Headers/qvariant.h:199:5: note: candidate constructor
    QVariant(const QVariant &other);
    ^
/Users/mikeller/qt/6.5.0/macos/lib/QtCore.framework/Headers/qvariant.h:215:5: note: candidate constructor
    QVariant(const QByteArray &bytearray) noexcept;
    ^
/Users/mikeller/qt/6.5.0/macos/lib/QtCore.framework/Headers/qvariant.h:224:5: note: candidate constructor
    QVariant(const QString &string) noexcept;
    ^
/Users/mikeller/qt/6.5.0/macos/lib/QtCore.framework/Headers/qvariant.h:278:12: note: candidate constructor
    inline QVariant(QVariant &&other) noexcept : d(other.d)
           ^
```

Interestingly, building this branch with Qt 5.15.2 on MacOS is working fine for me - my next step is to investigate the difference between my build coniguration (using Qt installed with the official online installer as per `INSTALL`) and the build configuration used for the MacOS check differs, and why one likes the change and the other one doesn't...

bstoeger: Note that I meant `QString(ds->name)`, not `QVariant(ds->name)`.

mikeller: Haha, looks like I fell victim to confirmation bias. Using `QString()` does indeed work, and you are correct that it is a lot more concise about what we _want_ to do.

bstoeger:
LGTM, thanks.

PR #3873, opened 2023-04-28
Update comment on Qt6 destruction of StatsView

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Explain why Qt6 destroys our StatsView. I hope this is not too aggressive, but it really annoys me. In particular, it stops progress on the profile-to-QtQuick project. For the stats it seems to work, but I am not sure the profile is happy when the view is constantly destroyed. There, the connection between widget and view is much more convoluted.

@dirkhh: please check.

### DIFF
--- a/core/subsurfacestartup.c
+++ b/core/subsurfacestartup.c
@@ -91,6 +91,7 @@ static void print_help()
 	printf("\n --dc-product=product  Set the dive computer to download from");
 	printf("\n --device=device       Set the device to download from");
 #endif
+	printf("\n --no-opengl-sharing   Don't share OpenGL contexts between QQuickItems (for testing)");
 	printf("\n --cloud-timeout=<nr>  Set timeout for cloud connection (0 < timeout < 60)\n\n");
 }
 
@@ -147,6 +148,10 @@ void parse_argument(const char *arg)
 			}
 			if (strcmp(arg, "--allow_run_as_root") == 0) {
 				++force_root;
+				return;
+			}
+			if (strcmp(arg, "--no-opengl-sharing") == 0) {
+				// This was already checked in main()
 				return;
 			}
 #if SUBSURFACE_DOWNLOADER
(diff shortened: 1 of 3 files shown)


dirkhh (approved):
This seems to align with what we have observed.
Of course the obvious question remains... now what?

mikeller:
> This seems to align with what we have observed. Of course the obvious question remains... now what?

Introduce OpenGL as a hardware requirement for Subsurface? 🤣

bstoeger:
> Of course the obvious question remains... now what?

Well, either "fix" the Quadrant-widget thing. Removing/Adding widgets seemed like a good idea, but apparently it's not.

Or, alternatively, take care that destruction/creation of the view doesn't lead to unexpected results. With the profile, this will be "interesting". With this PR applied one will not be able to test it under "normal" circumstances, i.e. OpenGL enabled. Therefore, setting the flag should be disableable (is that a word?) by command line options. Sigh.

dirkhh:
> > Of course the obvious question remains... now what?
> 
> Well, either "fix" the Quadrant-widget thing. Removing/Adding widgets seemed like a good idea, but apparently it's not.
> 
> Or, alternatively, take care that destruction/creation of the view doesn't lead to unexpected results. With the profile, this will be "interesting". With this PR applied one will not be able to test it under "normal" circumstances, i.e. OpenGL enabled. Therefore, setting the flag should be disableable (is that a word?) by command line options. Sigh.

I believe that we need the adding / removing for the filter, correct?

bstoeger:
I hope that it can be done with some hide()/show()-ing. But I am not sure. 

The fundamental problem is that we use the same widgets in different places depending on the mode. In Qt's object-ownership model you can have a widget with a single parent only (no reference counting or the like).

There must be other projects with a similar interface - perhaps let's get some "inspirations" from there.

harshsingh32 (approved):
lgtm

bstoeger:
> Therefore, setting the flag should be disableable (is that a word?) by command line options.

Well - this is fantastic. The flag must be set before creating the `QApplication` object. But we can only access the arguments if the `QApplication` has been instantiated.

And no, we can't simply use the `argc`/`argv` interface, because Qt filters out its own flags and who knows what kind of weird encoding we get on Windows!?

dirkhh:
So... given that we pretty much have to get Qt6 to work completely in order to make progress on Android and macOS... I guess we need to revisit this?

PR #3874, opened 2023-04-29
[TEST] core: use C++ std::strings for default directory and filename

bstoeger:
The memory management and string concatenation was hard to follow.

Since the mobile files ios.cpp and android.cpp were already converted to C++, let's do the same for Unix, Windows and MacOS.

Simply store the default directory and filename in a function-level static string. Thus, it will be initialized on first call and freed on application exit. Since the std::string data is guaranteed to be contiguous and zero-terminated, it can be used from C code.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Since I am frustrated with the whole QtQuick topic, let's do something unimportant and simple. Convert some string-handling to C++.

This PR is for compile testing, since I can currently only compile Linux-Desktop. Mobile-on-desktop doesn't want to compile since I upgraded my system.


bstoeger:
Interesting - github reported a conflict, but locally it rebased without problems. It seems to have done the correct thing despite renaming of a source file...

dirkhh:
that's not uncommon. GitHub sucks at merging...

bstoeger:
This works for me. Could someone check if this does the right thing on Windows? This was completely blindfolded.

mikeller (approved):
I gave it a poke with a stick in Windows. Not sure what exactly to test, but at a quick glance:
- the fonts look sane;
- strings look sane;
- the proposed default file location / name looks sane:

image

bstoeger:
Thanks for testing. I guess the crux will be user directories with non ASCII characters. Perhaps we should just merge and wait for fallout. @dirkhh?

mikeller:
@bstoeger: You mean as in a user with a login name of '😀face'? I can probably test this. But in general I'd say just go for it and merge - after all the users using the nightly builds are exactly there to find and report problems with these edge cases.

atdotde:
You don't have to wait for user 💩-face. I think @dirkhh has a test account with a name heavily relying on röck döts like Çağdaş G̈r̈oẗus̈ Mötley Crüe or something like that.

bstoeger:
Ok, let's merge and look wait for fallout. I think this is a massive readability improvement (in an admittedly obscure part of the code base).

Note that even umlauts are non-ASCII, not to speak of Arabic, Hebrew, Cyrillic, etc. I reckon that non-ASCII user names are more common than one might imagine.

PR #3875, opened 2023-04-29
[TRIVIAL?] cleanup: remove unnecessary string duplication

bstoeger:
I don't understand why the functions must be called on a copy of the string.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Remove redundant `strdup()`.

@dirkhh: Am I missing something - that `strdup()` appears completely pointless to me...?

### DIFF
--- a/core/unix.c
+++ b/core/unix.c
@@ -90,7 +90,6 @@ int enumerate_devices(device_callback_t callback, void *userdata, unsigned int t
 	size_t i;
 	FILE *file;
 	char *line = NULL;
-	char *fname;
 	size_t len;
 	if (transport & DC_TRANSPORT_SERIAL) {
 		const char *dirname = "/dev";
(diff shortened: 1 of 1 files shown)


bstoeger:
This comes from 6259b0301bbdf10c8337be35d440f6f1521ee2b0 and doesn't make sense to me. Let's commit and see if a UEMIS user complains...

dirkhh:
> This comes from [6259b03](https://github.com/subsurface/subsurface/commit/6259b0301bbdf10c8337be35d440f6f1521ee2b0) and doesn't make sense to me. Let's commit and see if a UEMIS user complains...

There are only about 3 Uemis users left :)

mikeller:
> > This comes from [6259b03](https://github.com/subsurface/subsurface/commit/6259b0301bbdf10c8337be35d440f6f1521ee2b0) and doesn't make sense to me. Let's commit and see if a UEMIS user complains...
> 
> There are only about 3 Uemis users left :)

Good to know. Is there any way to get anonymised statistics of this kind (distribution of dive computer models, distribution of the use of features, fields, ...). I think this would be super valuable to make informed decisions around which parts of the application to put effort into, vs. which ones to potentially 'clean up' in order to manage complexity.

dirkhh:
no - we don't collect this data. The reason I have a good guess is that when a couple of years ago I completely broke Uemis by mistake and didn't notice for about a year, only one person complained. And then when I fixed it, a second person mentioned "oh, it works again". So let's assume that this is undercounting by 50% - that gets us to 3 🙈

PR #3906, opened 2023-05-21
Desktop: Rework the Summary CSV Export XSLT.

mikeller:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Rework of the XSLT used to generate the Summary CSV export.

### Changes made:

1. fixed a bug causing invalid CSV to be generated for double quotes (`""`);
2. changed quoting and escaping to be compliant with RFC 4180;
3. changed output to contain information for all cylinders for all dives (instead of limiting the number of cylinders to howevermany are used for the last dive);
4. added an index to the cylinder data headings;
5. changed unit designators to use `[]` instead of `()`.
6. some minor improvements to the XSLT.

### Additional information:

Before:

Screenshot from 2023-05-21 14-11-49

After:

Screenshot from 2023-05-21 23-54-58

### Release note:

export: fix bug resulting in invalid CSV for '""' in 'CSV summary dive details'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+export: fix bug resulting in invalid CSV for '""' in 'CSV summary dive details'
 desktop: add support for multiple tanks to the profile ruler
 export: change format produced by 'CSV summary dive details' from TSV (tab separated) to CSV
 desktop: add function to merge dive site into site selected in list
--- a/xslt/xml2manualcsv.xslt
+++ b/xslt/xml2manualcsv.xslt
@@ -6,197 +6,146 @@
   <xsl:param name="units" select="units"/>
   <xsl:output method="text" encoding="UTF-8"/>
 
-  <xsl:variable name="fs"><xsl:text>,</xsl:text></xsl:variable>
-  <xsl:variable name="lf"><xsl:text>
-</xsl:text></xsl:variable>
+  <xsl:variable name="fs" select="','"/>
+  <xsl:variable name="quote" select="'&quot;'"/>
+  <xsl:variable name="lf" select="'&#xa;'"/>
 
-  <xsl:template match="/divelog/dives">
-    <xsl:variable name="cylinders">
-      <xsl:value-of select="count(dive[position()=last()]/cylinder|trip[position()=last()]/dive[position()=last()]/cylinder)"/>
-    </xsl:variable>
+  <xsl:template name="EscapeQuotes">
+    <xsl:param name="value"/>
     <xsl:choose>
-      <xsl:when test="$units = 1">
-
-        <!-- Print beginning of the header -->
-        <xsl:value-of select="concat(
-          '&quot;dive number&quot;',
-          $fs,
-          '&quot;date&quot;',
-          $fs,
-          '&quot;time&quot;',
-          $fs,
-          '&quot;duration (min)&quot;',
-          $fs,
-          '&quot;sac (cuft/min)&quot;',
-          $fs,
-          '&quot;maxdepth (ft)&quot;',
-          $fs,
-          '&quot;avgdepth (ft)&quot;',
-          $fs,
-          '&quot;mode&quot;',
-          $fs,
-          '&quot;airtemp (F)&quot;',
-          $fs,
-          '&quot;watertemp (F)&quot;',
-          $fs)"/>
-
-        <!-- Print cylinder info according to the amount of cylinders in dive -->
-        <xsl:for-each select="dive[position()=last()]/cylinder|trip[position()=last()]/dive[position()=last()]/cylinder">
-          <xsl:value-of select="concat(
-            '&quot;cylinder size (cuft)&quot;',
-            $fs,
-            '&quot;startpressure (psi)&quot;',
-            $fs,
-            '&quot;endpressure (psi)&quot;',
-            $fs,
-            '&quot;o2 (%)&quot;',
-            $fs,
-            '&quot;he (%)&quot;',
-            $fs)"/>
-        </xsl:for-each>
-
-        <!-- Print rest of the header -->
-        <xsl:value-of select="concat(
-          '&quot;location&quot;',
-          $fs,
-          '&quot;gps&quot;',
-          $fs,
-          '&quot;divemaster&quot;',
-          $fs,
-          '&quot;buddy&quot;',
-          $fs,
-          '&quot;suit&quot;',
-          $fs,
-          '&quot;rating&quot;',
-          $fs,
-          '&quot;visibility&quot;',
-          $fs,
-          '&quot;notes&quot;',
-          $fs,
-          '&quot;weight (lbs)&quot;',
-          $fs,
-          '&quot;tags&quot;')"/>
+      <xsl:when test="contains($value, $quote)">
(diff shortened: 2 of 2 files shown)


atdotde:
I don't speak xslt, so I cannot really say. @mturkia could you please nod your head?

mturkia:
Skimming through the changes, this looks pretty good. However, I did a quick test with xsltproc to see how the output looks like and the set of few dives I included did reveal a bug. I had some cylinders without pressure information on one dive and this XSLT produces output with same start and end pressures on all cylinders. Not sure how easy / impossible this is to take into account in the XSLT. Anyway, here is a dive with few gas changes at the start to get extra cylinders there, so @mikeller can have a look.
[cylinders.zip](https://github.com/subsurface/subsurface/files/11601613/cylinders.zip)

mikeller:
Thanks @mturkia, that's a good finding.

As a preface I have to say that I do not know much about Air Integration as I am not personally using it.

From a quick check it looks like this is a problem that is already existing in the current CSV export functionality - this is to be expected as I have not changed anything in how pressure information is extracted:

image

A look into the subsurface XML format in `cylinders.xml` shows why this is happening:

```
  [...]
  <divecomputer model='Shearwater Perdix AI' deviceid='8f1f5914' diveid='832dd86c'>
  [...]
  <event time='0:10 min' type='25' flags='1' name='gaschange' cylinder='0' o2='34.0%' />
  <event time='2:00 min' type='25' flags='2' name='gaschange' cylinder='0' o2='34.0%' />
  <event time='2:20 min' type='25' flags='1' name='gaschange' cylinder='0' o2='34.0%' />
  <event time='3:30 min' type='25' flags='2' name='gaschange' cylinder='0' o2='34.0%' />
  <sample time='0:10 min' depth='4.1 m' temp='28.0 C' pressure='202.706 bar' ndl='99:00 min' cns='7%' />
  <sample time='0:20 min' depth='5.0 m' pressure='202.154 bar' />
  <sample time='0:30 min' depth='5.6 m' pressure='201.741 bar' />
  <sample time='0:40 min' depth='6.5 m' pressure='200.775 bar' />
  <sample time='0:50 min' depth='6.4 m' pressure='200.362 bar' />
  [...]
```

From this it looks like Subsurface is discarding pressure information for all but one tank when storing samples.
And to reconstruct which tank the pressure information in a sample is for it seems to be necessary to reconstruct the timeline of the dive from gas changes and samples in order to determine which gasmix / tank was used (and presumably reported the gas pressure change) when a given sample was logged.
Not sure how this is meant to deal with situations where more than one tank sensor reported a pressure change for a sample (using a tank sensor to monitor the gas reserve of a buddy / student), or where the only gas change reported was not for the tank with the currently used gasmix (shooting a DSMB on a CCR)?

For the purpose of generating the CSV should be possible to do this reconstruction in XSLT, but at the moment I can not see a way to do this without going over board in complexity, and creating something that will be brittle and likely to break in edge cases - especially in cases where multiple tanks with sensors were used during a dive so that all of them will have differing start / end pressures that can be retrieved from different samples before / after gas changes within the dive.

So I can see the following options:

### 1. leave it as it is

Pros:
- the (incorrect) data contents stay the same as they have been, so existing integrations that users have built will not break.

Cons:
- incorrect data is reported if tank sensors and multiple gasmixes are used.

### 2. change it to determine and report the correct start / end pressures for all dives

Pros:
- the data that is reported is correct;
- the maximum amount of data is reported.

Cons:
- high effort;
- XSLT is somewhat write-only - complex XSLT will be hard to maintain;
- potentially introduces brittleness and new bugs.

### 3. change the format to report one start / end pressure for the dive, independent of tanks

image

Pros:
- correct data is reported;
- simple to implement.

Cons:
- the existing CSV format is changed, potentially breaking users' integrations;
- the data can be confusing as end pressure can be higher than the start pressure if different tanks are used at the beginning and end of a dive (this is already the case for 1.).


My recommendation is to merge this as it is (unless there are other issues), as this is just keeping the status quo for start / end pressures, and then have a discussion about which option to take, and do any changes coming out of this in a separate pull request.

mturkia (approved):
My example does not really show what Subsurface does with multiple cylinders as I only had pressure sensor on the single cylinder. (And I actually had only one cylinder, just did the gas switches to get additional test data.) If you open up the XML in Subsurface, you only see single cylinder being used unless you have the option to show unused cylinders enabled. Then if you add start and end pressures to one of the "unused" cylinders, that info will be exported properly.

The case of pressure sensor on multiple cylinders, I have no idea how that is recorded. I.e. not sure if there will be cylinder index recorded, or if one just needs to keep track of which cylinder is switched to and assume the pressures to  be for that. But XSLT is really not suitable for complex cases, and most people do not dive with configuration that would hit this "feature".

Anyway, I do agree with your assessment. Let's merge as is for now.

mikeller:
@mturkia: It looks like https://github.com/subsurface/large-anonymous-sample-data doesn't contain any sample logs with air integration data. I'll pump some friends who are diving with multiple sensors for some logs so that we can have a peek at how subsurface is handling this.

mikeller:
@mturkia: I managed to get [a log](https://github.com/subsurface/subsurface/files/11622394/multiple_sensors.zip) from a dive with multiple (2 in this case) tank sensors, and this gives some more insight: Instead of a single `pressure` attribute in samples Subsurface is using `pressure0` and `pressure1` attributes:

```
  <divecomputer model='Shearwater Nerd 2' deviceid='64257c95' diveid='a746200e' dctype='CCR' no_o2sensors='3'>
  [...]
  <event time='0:10 min' type='25' flags='6' name='gaschange' cylinder='5' o2='16.0%' he='45.0%' />
  <sample time='0:10 min' depth='1.5 m' temp='12.0 C' pressure0='175.816 bar' pressure1='196.914 bar' sensor1='0.737 bar' sensor2='0.744 bar' sensor3='0.754 bar' po2='0.7 bar' />
  <sample time='0:20 min' depth='1.8 m' ndl='99:00 min' />
  <sample time='0:30 min' depth='2.1 m' pressure1='196.776 bar' sensor1='0.713 bar' sensor3='0.73 bar' />
```

It looks like the presence of a single `pressure` attribute  in the XML has to be read as 'pressure reported for a single tank', and the existence of `pressure0`, `pressure1`,... attributes defines pressure readings for multiple tanks.

Unfortunately at the moment Subsurface is lacking the facility to allow the user to define a mapping between gasmixes and tank pressures reported by the dive computer, and always assumes a 1:1 mapping between the gasmixes in the order they are reported by the dive computer and the tank sensor index - which is likely to be wrong for most multi-gas setups:

image

But this is a problem that is not limited to the exporting of information and will likely require a UI change and some way to persist a 'preferred mapping' of sensors.

What are your thoughts about a change for the XSLT that limits the pressure output for a single `pressure` attribute to the first tank, and adds support for the `pressure0`, `pressure1`, ... situation?

mturkia:
IMO the single `pressure` should be limited to the first tank, and support for multiple cylinder pressures should be added. And there is the case of manually entered start and end pressures, which should still be supported if the pressures are entered manually.

PR #3912, opened 2023-05-26
Mobile: Fix Editing of Gradient Factors

### LINKED ISSUE #3911: Can't set GF 100 on Android
### Describe the issue:
- [X ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
I can't set GF 100/100 on Android : when I set 100, the value is corrected to 10. So I had to use 99 instead, while the GF 100/100 is valid on the Linux desktop application.

Maybe related : I'm using the application in French language.

### Operating system:
Android 12 on Pixel 3 XL

### Subsurface version:
Version: 3.4.7 (5.0.9.35)

### Steps to reproduce:
In Settings / Advanced, set GF Low and/or GF High to 100

### Current behavior:
Value is reset to 10

### Expected behavior:
Value 100 is kept

### Additional information:
- 99 is valid
- This issue is not seen on Linux desktop application where I can set GF values to 100

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix the issue that Gradient Factors cannot be set to 100 in the mobile version. This is done by changing the edits from a text box to a spin edit, which seems to be a better match for numerical values. As a side effect this also solves the issue that the keyboard for the text edit is not properly displayed when settings are opened when dive details are already on the page stack.

### Changes made:

1. change `TextField` to `SpinBox` with appropriate range for gradiend factors in mobile settings;
2. fix the range for planner gradient factors in mobile;
3. fix indentation in mobile settings QML file.

### Related issues:

Fixes #3911.

### Additional information:

Look of the SpinBox based edits:

Screenshot_2023-05-26-13-54-05-930_org subsurfacedivelog mobile

Reproduction for the 'keyboard not showing' problem:

Screenshot_2023-05-26-10-02-44-754_org subsurfacedivelog mobile

### Mentions:

Reported-by: @gbetous

### DIFF
--- a/mobile-widgets/qml/DivePlannerSetup.qml
+++ b/mobile-widgets/qml/DivePlannerSetup.qml
@@ -228,7 +228,7 @@ TemplatePage {
 				}
 				TemplateSpinBox {
 					from: 1
-					to: 99
+					to: 100
 					stepSize: 1
 					value: Backend.gflow
 					textFromValue: function (value, locale) {
@@ -245,7 +245,7 @@ TemplatePage {
 				}
 				TemplateSpinBox {
 					from: 1
-					to: 99
+					to: 100
 					stepSize: 1
 					value: Backend.gfhigh
 					textFromValue: function (value, locale) {
--- a/mobile-widgets/qml/DivePlannerSetup.qml
+++ b/mobile-widgets/qml/DivePlannerSetup.qml
@@ -227,8 +227,8 @@ TemplatePage {
 					leftPadding: Kirigami.Units.smallSpacing * 2
 				}
 				TemplateSpinBox {
-					from: 1
-					to: 100
+					from: 10
+					to: 150
 					stepSize: 1
 					value: Backend.gflow
 					textFromValue: function (value, locale) {
@@ -244,8 +244,8 @@ TemplatePage {
 					leftPadding: Kirigami.Units.smallSpacing * 2
 				}
 				TemplateSpinBox {
-					from: 1
-					to: 100
+					from: 10
+					to: 150
 					stepSize: 1
 					value: Backend.gfhigh
 					textFromValue: function (value, locale) {
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -632,8 +632,8 @@ TemplatePage {
 				}
 				TemplateSpinBox {
 					id: gfLow
-					from: 1
-					to: 100
+					from: 10
+					to: 150
 					stepSize: 1
 					value: PrefTechnicalDetails.gflow
 					textFromValue: function (value, locale) {
@@ -649,8 +649,8 @@ TemplatePage {
 				}
 				TemplateSpinBox {
 					id: gfHigh
-					from: 1
-					to: 100
+					from: 10
+					to: 150
 					stepSize: 1
 					value: PrefTechnicalDetails.gfhigh
 					textFromValue: function (value, locale) {
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -585,16 +585,16 @@ TemplatePage {
 						PrefDisplay.singleColumnPortrait = checked
 					}
 				}
-                TemplateLabel {
-                    text: qsTr("Depth line based on ×3 intervals")
-                }
-                SsrfSwitch {
-                    checked: PrefDisplay.three_m_based_grid
-                    onClicked: {
-                        PrefDisplay.three_m_based_grid = checked
-                        rootItem.settingsChanged()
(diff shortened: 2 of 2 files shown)


mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -245,7 +245,7 @@ TemplatePage {
 				}
 				TemplateSpinBox {
 					from: 1
-					to: 99
```
atdotde, line 248: There can be situations where values above 100 are desirable. I would allow at least 120 if not 200.

mobile-widgets/qml/Settings.qml
```diff
@@ -585,16 +585,16 @@ TemplatePage {
 						PrefDisplay.singleColumnPortrait = checked
 					}
 				}
-                TemplateLabel {
```
atdotde, line 588: What is going on with white space here (I cannot distinguish spaces from tabs here in github)?

atdotde (changes requested):

mobile-widgets/qml/DivePlannerSetup.qml
mikeller: That's interesting - I have so far only ever come across Gradient Factors being used in the range of 1 - 100%. Do you have references for literature on the use of Gradient Factors above 100%?

mobile-widgets/qml/Settings.qml
mikeller: As stated in the pull request description, I have fixed incorrect indentation in this file as I was editing it anyway.

mobile-widgets/qml/DivePlannerSetup.qml
atdotde: Of course, that would be violating plain vanilla Bühlmann rules. But there are other models. For example, some VPM-B schedules, even with +2 conservatism can end up in the range of GF-high about 120. So, when you follow those you might want to compare to Bühlmann+GF and then you need such values. Not that I would recommmend diving those schedules... 

Literature? I could offer https://thetheoreticaldiver.org/wordpress/index.php/2017/11/02/vpm-b-gradients-as-gradient-factors/

mikeller: > Of course, that would be violating plain vanilla Bühlmann rules.

To my understanding the 'Bühlmann rules' really just denote that at GF 100/100 the risk of getting DCI was what Bühlmann considered to be acceptable. So it is entirely possible to think of situations where using a decompression schedule that has got a higher risk than this is the best possible course of action - like a contingency where there isn't enough gas available to complete decompression at GF 100/100. But then of course it is not likely that it will be possible to use Subsurface to create a decompression plan when such a situation eventuates...

> Literature? I could offer [thetheoreticaldiver.org/wordpress/index.php/2017/11/02/vpm-b-gradients-as-gradient-factors](https://thetheoreticaldiver.org/wordpress/index.php/2017/11/02/vpm-b-gradients-as-gradient-factors/)

Interesting article - I know next to nothing about how VPM-B works, so this was educational. By the way the link in the first paragraph seems to be broken.

gbetous-swir:
Side question on this topic... the default value seems to be 35/75 (or something like that). Is is really a so common value ? I would think about 90/90 for example which is more common talking to people around me.

Anyway, thanks a lot for those fixes !

atdotde (approved):

mikeller:
> Side question on this topic... the default value seems to be 35/75 (or something like that). Is is really a so common value ? I would think about 90/90 for example which is more common talking to people around me.

Without further qualification of the use case 35 / 75 is a common choice of gradient factors for technical diving, slightly on the conservative side - or at least it used to be some years back when more emphasis was put on deep stops. These days thinking seems to be going towards de-emphasising deep stops, which is done by moving GFlow and GFhigh closer together, so something like 45 / 65 will probably give you similar overall decompression times for most profiles as 35 / 75.

On the other hand there are lots of factors in play that can influence the choice of gradient factors, like the type / profile of the dive, distance to a recompression chamber, other risk factors like temperature or current,...

For technical decompression diving 90 / 90 seems to be a choice of gradient factors that is on the 'high risk' side of things. On the other hand the algorithms used in recreational dive computers seem to calculate the no decompression limits close to what 90 / 90 on Bühlmann tables will give, so for no-decompression dives with a safety stop this seems to be a safe choice.

atdotde:
Looks good. Let's merge. 

@gbetous-swir I agree with @mikeller that 90/90 is a bit too aggressive to be used as a default. Most divers I know use gradient factors significantly below that.

PR #3913, opened 2023-05-26
Desktop: Add Multi-Tank Support for Profile Ruler.

### LINKED ISSUE #3902: Ruler SAC incorrect with multiple transmitters
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

When sidemounting with two transmitters, the delta-P and SAC displayed in the Ruler detail is only using the data from one transmitter. The overall SAC is correctly computed. The "instantaneous" SAC shown in the Info window is also correct.

### Operating system:

OS: Linux/Manjaro 5.15.102-1

### Subsurface version:

v5.0.10-168-gcf95c3bd82c4 (complied myself)
Also happens on the version packaged in the AUR (v5.0.10-1)

### Steps to reproduce:

1) Import the attached dive
2) Activate the Ruler in the profile window
3) Drag the Ruler endpoint to 13:00, leaving the startpoint at 0:00

### Current behavior:

Displayed delta P and SAC are near 0

### Expected behavior:

Displayed delta P = 478 psi (24 psi from cyl 1, 454 psi from cyl 2), with the corresponding SAC calculation somewhere around 0.6 cuft/min.

### Mentions:

[Ruler SAC bug.ssrf.zip](https://github.com/subsurface/subsurface/files/11509366/Ruler.SAC.bug.ssrf.zip)

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add support for tracking the gas usage across multiple tanks to the 'bar used' and SAC values shown for the profile ruler.
The following rules are implemented:
- a tank is considered 'used' if at least one bar has been consumed;
- only used tanks are taken into account for calculations;
- 'bar used' is only shown if all tanks used have the same (or unknown) volume;
- SAC is only shown if all tanks used have a known volume.

### Changes made:

1. add support for tracking the gas usage across multiple tanks to the 'bar used' and SAC values shown for the profile ruler;
2. removed unneeded `memcpy()`.

### Related issues:

Fixes #3902.

### Additional information:

Screenshot from 2023-05-26 20-03-48

### Release note:

desktop: add support for multiple tanks to the profile ruler

### Mentions:

Reported-by: @pabdakine

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -1511,9 +1511,8 @@ void compare_samples(const struct dive *d, const struct plot_info *pi, int idx1,
 	char *buf2 = malloc(bufsize);
 	int avg_speed, max_asc_speed, max_desc_speed;
 	int delta_depth, avg_depth, max_depth, min_depth;
-	int bar_used, last_pressure, next_pressure, pressurevalue;
+	int pressurevalue;
 	int last_sec, delta_time;
-	bool crossed_tankchange = false;
 
 	double depthvalue, speedvalue;
 
@@ -1544,10 +1543,15 @@ void compare_samples(const struct dive *d, const struct plot_info *pi, int idx1,
 	avg_depth = 0;
 	max_depth = 0;
 	min_depth = INT_MAX;
-	bar_used = 0;
 
 	last_sec = start->sec;
-	last_pressure = get_plot_pressure(pi, idx1, 0);
+
+	volume_t cylinder_volume = { .mliter = 0, };
+	int *start_pressures = calloc((size_t)pi->nr_cylinders, sizeof(int));
+	int *last_pressures = calloc((size_t)pi->nr_cylinders, sizeof(int));
+	int *bar_used = calloc((size_t)pi->nr_cylinders, sizeof(int));
+	int *volumes_used = calloc((size_t)pi->nr_cylinders, sizeof(int));
+	bool *cylinder_is_used = calloc((size_t)pi->nr_cylinders, sizeof(bool));
 
 	data = start;
 	for (int i = idx1; i < idx2; ++i) {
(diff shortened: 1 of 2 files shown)


core/profile.c
```diff
@@ -1544,10 +1543,15 @@ void compare_samples(const struct dive *d, const struct plot_info *pi, int idx1,
 	avg_depth = 0;
 	max_depth = 0;
 	min_depth = INT_MAX;
-	bar_used = 0;
 
 	last_sec = start->sec;
-	last_pressure = get_plot_pressure(pi, idx1, 0);
+
+	volume_t cylinder_volume = { .mliter = 0, };
+	int *start_pressures = calloc((size_t)pi->nr_cylinders, sizeof(int));
```
bstoeger, line 1550: Not in this PR, but IMHO we really should switch that thing to C++ and start using `std::vector` erc.

core/profile.c
```diff
@@ -1567,14 +1571,36 @@ void compare_samples(const struct dive *d, const struct plot_info *pi, int idx1,
...
+		for (unsigned cylinder_index = 0; cylinder_index < pi->nr_cylinders; cylinder_index++) {
+			int next_pressure = get_plot_pressure(pi, i, cylinder_index);
+			if (next_pressure && !start_pressures[cylinder_index])
+				start_pressures[cylinder_index] = next_pressure;
+
+			if (start_pressures[cylinder_index]) {
+				if (last_pressures[cylinder_index]) {
+					bar_used[cylinder_index] += last_pressures[cylinder_index] - next_pressure;
+
+					cylinder_t *cyl = get_cylinder(d, cylinder_index);
```
bstoeger, line 1584: Is there a situation where `pi->nr_cylinders` and `d->cylinders.nr` could diverge?

(In other words: would it make sense to loop over the latter instead of the former?)

mikeller: For what I could find it is the same:

https://github.com/mikeller/subsurface/blob/d045c3235c237dbdc782bdbe010f9c170df557dc/core/profile.c#L410

I used `pi->nr_cylinders` because this is what is used in other places in `compare_samples()` - so it seems to be safe to use.

All in all it would probably make sense to eliminate `pi->nr_cylinders` entirely and replace all uses with `d->cylinders.nr` - the duplication seems to only create confusion.

Happy to do this in a separate pull request.

bstoeger:
LGTM.

core/profile.c
mikeller: @bstoeger:

 > All in all it would probably make sense to eliminate `pi->nr_cylinders` entirely and replace all uses with `d->cylinders.nr` - the duplication seems to only create confusion.
> 
> Happy to do this in a separate pull request.

I had a look into this, and unfortunately this will only _almost_ work in a clean way. The one case that will create awkward code if this is done is access to the contents of the `pi->pressures[]` array:

https://github.com/subsurface/subsurface/blob/a127c4ac63869b753f2912cc3903b0ef68f96ba8/core/profile.h#L121

To me this looks like the best way to eliminate `pi->nr_cylinders` will be to move this to C++ and replace `pi->pressures[]` with a multi-dimensional `std::vector` that can keep track of its size intrinsically.

PR #3917, opened 2023-06-01
Core: Fix bug using salinity and pressure values in mbar <-> depth conversion

rafsalvioni:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The conversion between mbar and depth sometimes uses DC's salinity, sometimes user's salinity. By other hand, it uses surface pressure given by user in calculation.

This fix try to standartize this values, using them from same source.

### Changes made:
dive.c: mbar <-> depth conversion functions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 desktop: added button to hide the infobox in the dive profile
 desktop: use persisted device information for the dive computer configuration
+core: fix bug using salinity and pressure values in mbar <-> depth conversion
 export: fix bug resulting in invalid CSV for '""' in 'CSV summary dive details'
 desktop: add support for multiple tanks to the profile ruler
 export: change format produced by 'CSV summary dive details' from TSV (tab separated) to CSV
--- a/core/dive.c
+++ b/core/dive.c
@@ -3254,12 +3254,20 @@ static double calculate_depth_to_mbarf(int depth, pressure_t surface_pressure, i
 
 int depth_to_mbar(int depth, const struct dive *dive)
 {
-	return lrint(calculate_depth_to_mbarf(depth, dive->surface_pressure, dive->salinity));
+	return lrint(depth_to_mbarf(depth, dive));
 }
 
 double depth_to_mbarf(int depth, const struct dive *dive)
 {
-	return calculate_depth_to_mbarf(depth, dive->surface_pressure, dive->salinity);
+	// To downloaded and planned dives, use DC's values
+	int salinity = dive->dc.salinity;
+	pressure_t surface_pressure = dive->dc.surface_pressure;
+
+	if (is_manually_added_dc(&dive->dc)) { // To manual dives, salinity and pressure in another place...
+		surface_pressure = dive->surface_pressure;
+		salinity = dive->user_salinity;
+	}
+	return calculate_depth_to_mbarf(depth, surface_pressure, salinity);
 }
 
 double depth_to_bar(int depth, const struct dive *dive)
@@ -3278,20 +3286,26 @@ double depth_to_atm(int depth, const struct dive *dive)
  * take care of this, but the Uemis we support natively */
 int rel_mbar_to_depth(int mbar, const struct dive *dive)
 {
-	double specific_weight = salinity_to_specific_weight(SEAWATER_SALINITY);
-	if (dive->dc.salinity)
-		specific_weight = salinity_to_specific_weight(dive->dc.salinity);
+	// To downloaded and planned dives, use DC's salinity. Manual dives, use user's salinity
+	int salinity = is_manually_added_dc(&dive->dc) ? dive->user_salinity : dive->dc.salinity;
+	if (!salinity)
+		salinity = SEAWATER_SALINITY;
+
 	/* whole mbar gives us cm precision */
+	double specific_weight = salinity_to_specific_weight(salinity);
 	return (int)lrint(mbar / specific_weight);
 }
 
 int mbar_to_depth(int mbar, const struct dive *dive)
 {
-	pressure_t surface_pressure;
-	if (dive->surface_pressure.mbar)
-		surface_pressure = dive->surface_pressure;
-	else
+	// To downloaded and planned dives, use DC's pressure. Manual dives, use user's pressure
+	pressure_t surface_pressure = is_manually_added_dc(&dive->dc)
+		? dive->surface_pressure
+		: dive->dc.surface_pressure;
+
+	if (!surface_pressure.mbar)
 		surface_pressure.mbar = SURFACE_PRESSURE;
+		
 	return rel_mbar_to_depth(mbar - surface_pressure.mbar, dive);
 }
 
--- a/core/planner.c
+++ b/core/planner.c
@@ -695,9 +695,20 @@ bool plan(struct deco_state *ds, struct diveplan *diveplan, struct dive *dive, i
 
 	set_gf(diveplan->gflow, diveplan->gfhigh);
 	set_vpmb_conservatism(diveplan->vpmb_conservatism);
-	if (!diveplan->surface_pressure)
-		diveplan->surface_pressure = SURFACE_PRESSURE;
-	dive->surface_pressure.mbar = diveplan->surface_pressure;
+
+	if (!diveplan->surface_pressure) {
+		// Lets use dive's surface pressure in planner, if have one...
+		if (dive->dc.surface_pressure.mbar) { // First from DC...
+			diveplan->surface_pressure = dive->dc.surface_pressure.mbar;
+		}
+		else if (dive->surface_pressure.mbar) { // After from user...
+			diveplan->surface_pressure = dive->surface_pressure.mbar;
+		}
+		else {
+			diveplan->surface_pressure = SURFACE_PRESSURE;
+		}
+	}
+	
 	clear_deco(ds, dive->surface_pressure.mbar / 1000.0, true);
 	ds->max_bottom_ceiling_pressure.mbar = ds->first_ceiling_pressure.mbar = 0;
 	create_dive_from_plan(diveplan, dive, is_planner);


rafsalvioni:
Checks with "TestProfile" fails.... Some MOD calculation could be different than reference profile data. If you agree PR, I can update tests reference

mikeller:
> Checks with "TestProfile" fails.... Some MOD calculation could be different than reference profile data. If you agree PR, I can update tests reference

Yes, please add the changes that are required to make the tests pass to the pull request - this will be helpful to discuss benefits and risks of the proposed changes.

I suspect that it is not as simple as what you are currently proposing to make Subsurface take the user selected salinity into account: This may be the right thing to do where the calculated depth is shown to the user, but may introduce an error into the data where it is used for further calculations that expect the salinity set in the dive computer to be used, like in this case:

https://github.com/subsurface/subsurface/blob/4c193d7bdfbcb1ca755250df1873131d9709a9df/core/uemis.c#L260-L266

(In my opinion showing the user a fractional number for the decompression stop depth if they manually adjust the salinity does not make sense here.)

atdotde:
Given that Uemis has gone out of business many many moons ago, I doubt that there are still a lot of users of their computers around and even fewer ones that decide to manually adjust salinity. So probably it does not matter much.

That said, I regret at this point that we went the way of allowing the user to adjust the salinity as I can really thing of very few cases where this is a good idea. We talked about this in the past but the quick reason is that no dive computers measure depth but only pressure and the assume some density to do the conversion. We store stuff in terms of depth but what matters for deco etc is really pressure again and so messing with this is very likely to introduce unecessary errors. 

In particular, when I have a computer that allows me to set the water type (fresh vs salt) and I forget it set it before the dive and then dive and download the dive, I should keep that setting. Changing the salinity after the fact ("I forgot to set the dive computer to fresh water after coming back from holiday") does not improve things but makes things worse. As long as you don't start measuring depths with a yard stick everything is nice and consistent no matter what your dive computer thinks the salinity is, it is only the displayed depth that changes. But if you use a different salinity in you log than what the computer used to record the dive, you are messing with the physics.

But since we have this setting, I think the PR does the right thing. We just need to make sure that dive->user_salinity is zero unless the user decided to shoot herself  in the knee so the correct value is used. If it's not zero, the user asked for it. Tough luck.

mikeller:
> That said, I regret at this point that we went the way of allowing the user to adjust the salinity as I can really thing of very few cases where this is a good idea. We talked about this in the past but the quick reason is that no dive computers measure depth but only pressure and the assume some density to do the conversion. We store stuff in terms of depth but what matters for deco etc is really pressure again and so messing with this is very likely to introduce unecessary errors.
> 
> In particular, when I have a computer that allows me to set the water type (fresh vs salt) and I forget it set it before the dive and then dive and download the dive, I should keep that setting. Changing the salinity after the fact ("I forgot to set the dive computer to fresh water after coming back from holiday") does not improve things but makes things worse. As long as you don't start measuring depths with a yard stick everything is nice and consistent no matter what your dive computer thinks the salinity is, it is only the displayed depth that changes. But if you use a different salinity in you log than what the computer used to record the dive, you are messing with the physics.

That's pretty much what I am worried about. In the end depth is 'just a number', and a 'display value' really, whereas the decompression information that we are calculating directly relates to the risk we are taking, and is often an important factor in planning and decision making. So to me it seems to be wrong to make changes that are likely to introduce errors into the decompression information all for the sake of correcting a display value.

> But since we have this setting, I think the PR does the right thing. We just need to make sure that dive->user_salinity is zero unless the user decided to shoot herself in the knee so the correct value is used. If it's not zero, the user asked for it. Tough luck.

For the above reason I think that we are giving up too easy if we accept that this will probably break decompression information if done in the way that is currently proposed.
I think a better (but more involved) way to get the desired result (i.e. getting correct depth information shown to the user if they adjust the salinity of imported dive logs) will be to keep the calculations and stored depth values based on the salinity used by the dive computer, but then adjust these values before they are shown to the user if an adjusted salinity is used.

>  If it's not zero, the user asked for it. Tough luck.

In my opinion this will need at least a warning that is shown to the user if they have adjusted the salinity information, letting them know that the decompression information is likey to be incorrect.

Or as an alternative we could introduce a preference for this, allowing the user to choose between adjusted depth information being displayed or correct decompression information being calculated.

rafsalvioni:
Before commenting, I would like to explain my motivation...

I started the PR due to the MOD calculation displayed in the cylinder data, because doing the calculation manually I noticed a difference in the values. So I tried to understand what was going on...

Analyzing the code, I noticed that the calculation considers the surface pressure and the salinity (or density) of the water. That was the difference compared to my manual calculation. Ok for me and clarified.

However, I noticed something that seemed strange to me: to convert mbar to depth, the surface pressure used can be changed by the user and, in this case, it is the value that will be used in the calculation. The salinity can also be changed by the user, but in this case, the value is ignored and the value given by the DC is used.
https://github.com/subsurface/subsurface/blob/7771b444a1014eb529c7bcdd975bd1d69c1409cf/core/dive.c#LL3275C1-L3296C2

Would that be correct? Why is information given by the user considered and not another? Following this line, I suggested using the salinity given by the user in the calculation.

Another fact that seemed strange to me: When converting mbar to depth, only a specific DC is used:
https://github.com/subsurface/subsurface/blob/7771b444a1014eb529c7bcdd975bd1d69c1409cf/core/dive.c#L3283

But in the inverse operation, depth to mbar, the merge of salinities given by all DCs is used:
https://github.com/subsurface/subsurface/blob/7771b444a1014eb529c7bcdd975bd1d69c1409cf/core/dive.c#L3262

I believe that in most cases, when only one DC is used, this is not a problem. But when there are more than 1 DC that eventually report different salinities, the calculation would be different, because `dive->salinity` is the [average salinity](https://github.com/subsurface/subsurface/blob/7771b444a1014eb529c7bcdd975bd1d69c1409cf/core/dive.c#L1296) of the DCs. And in those cases:
`dive->dc.salinity != dive->salinity`

It was for these reasons that I believed the best way to "standardize" the salinity information, in the order: salinity given by the user, salinity given by the DCs (average) or saltwater salinity.

That said, I hadn't addressed the point given by @atdotde. For data downloaded from the DC, mainly the profile samples, the reported depths uses the salinity configured in the DC. It doesn't matter if the DC was misconfigured. At this point we must use the salinity given by DC. Otherwise, the depths reported by it would be physically discrepant. For example: if the DC reported a depth of 20m in salt water, if we change the salinity to fresh water, the dive profile would be inconsistent as those 20m would be virtually wrong. On the other hand, this "discrepancy" can already occur today if the user changes the surface pressure.

Regarding @mikeller's point, I understand the question, but as I mentioned, today we already have a possible difference between the input data and the profile diving due to surface pressure which is editable.

Im going to examplify this:
image
Using 998 as mbar in surface pressure. MOD's cylinder and MOD's profile are equal

image
Using 700 as mbar in surface pressure. Manually edited... MOD's cylinder and MOD's profile are different

Same dive... Just 1 tank utilized

In this scenario, I see these possible solutions:
1- In the case of downloaded dives, we ALWAYS use the DC data (salinity and surface pressure). It doesn't matter what the user enters. And in the case of manual dives, we use the user's data... (since we don't have the DC ;-)). In the case of planned dives, I don't know how it would look. I guess that it would be the same approach as for manual dives;
2- We use the input data and recalculate the profile at each change (more hard to implement). Data reported by DCs would only be initial values

mikeller:
Hi @rafsalvioni, thanks for your detailed explanation.

> I started the PR due to the MOD calculation displayed in the cylinder data, because doing the calculation manually I noticed a difference in the values. So I tried to understand what was going on...
> 
> Analyzing the code, I noticed that the calculation considers the surface pressure and the salinity (or density) of the water. That was the difference compared to my manual calculation. Ok for me and clarified.

Ok. I guess at this point we can ask ourselves: What is the most useful value to be displayed to the user when they are looking at a dive log downloaded from a dive computer:
1. the MOD value for that was shown to the user on the dive computer during the dive, and used as the basis for gas switch prompts;
2. a MOD value that is calculated for user entered values.

To me it looks like 1. will be the winner in pretty much all cases.

> However, I noticed something that seemed strange to me: to convert mbar to depth, the surface pressure used can be changed by the user and, in this case, it is the value that will be used in the calculation. The salinity can also be changed by the user, but in this case, the value is ignored and the value given by the DC is used. [`7771b44`/core/dive.c#LL3275C1-L3296C2](https://github.com/subsurface/subsurface/blob/7771b444a1014eb529c7bcdd975bd1d69c1409cf/core/dive.c#LL3275C1-L3296C2) 
> 
> Would that be correct? Why is information given by the user considered and not another? Following this line, I suggested using the salinity given by the user in the calculation.

To me this pretty much looks like there is a bug in this code - [this line](https://github.com/subsurface/subsurface/blob/master/core/dive.c#L3291) should probably be 

```
if (dive->dc.surface_pressure.mbar)
```

Note that in the case of dives that are being planned there is a 'virtual dc' device that is used, so there already is an existing (and clean) way to inject user selected data for planned dives.

> Another fact that seemed strange to me: When converting mbar to depth, only a specific DC is used:
> 
> https://github.com/subsurface/subsurface/blob/7771b444a1014eb529c7bcdd975bd1d69c1409cf/core/dive.c#L3283
> 
> But in the inverse operation, depth to mbar, the merge of salinities given by all DCs is used:
> 
> https://github.com/subsurface/subsurface/blob/7771b444a1014eb529c7bcdd975bd1d69c1409cf/core/dive.c#L3262
> 
> I believe that in most cases, when only one DC is used, this is not a problem. But when there are more than 1 DC that eventually report different salinities, the calculation would be different, because `dive->salinity` is the [average salinity](https://github.com/subsurface/subsurface/blob/7771b444a1014eb529c7bcdd975bd1d69c1409cf/core/dive.c#L1296) of the DCs. And in those cases: `dive->dc.salinity != dive->salinity`

To me this looks like yet another bug of the same type - for dives that are based on data downloaded from a dive computer we should be using `dive->dc.salinity` for both ways of the calculation, as this is what is already the basis for any conversions between pressure and depth that is inherent in the rest of the data set - this is the only way to actually get consistent data.

Obviously in an ideal world we would be doing this independently for every dive computer used on a dive, and switch all displayed values every time the user changes the selected dive computer. But even doing this only for the first dive computer will result in better and more consistent data than allowing user edited values to introduce errors.

> It was for these reasons that I believed the best way to "standardize" the salinity information, in the order: salinity given by the user, salinity given by the DCs (average) or saltwater salinity.
> 
> That said, I hadn't addressed the point given by @atdotde. For data downloaded from the DC, mainly the profile samples, the reported depths uses the salinity configured in the DC. It doesn't matter if the DC was misconfigured. At this point we must use the salinity given by DC. Otherwise, the depths reported by it would be physically discrepant. For example: if the DC reported a depth of 20m in salt water, if we change the salinity to fresh water, the dive profile would be inconsistent as those 20m would be virtually wrong. On the other hand, this "discrepancy" can already occur today if the user changes the surface pressure.
>
> Regarding @mikeller's point, I understand the question, but as I mentioned, today we already have a possible difference between the input data and the profile diving due to surface pressure which is editable.
> 
> Im going to examplify this: image Using 998 as mbar in surface pressure. MOD's cylinder and MOD's profile are equal
> 
> image Using 700 as mbar in surface pressure. Manually edited... MOD's cylinder and MOD's profile are different
> 
> Same dive... Just 1 tank utilized

Yes, that is exactly what I mean - standardisation is good, but since some of the data has already been processed with the values set and measured by the dive computer, the only way we can standardise is to use these values, and treat the user entered values as 'informational display values'.
And in the current state it looks like we are already broken in this respect by using an user entered surface pressure.

> In this scenario, I see these possible solutions: 1- In the case of downloaded dives, we ALWAYS use the DC data (salinity and surface pressure). It doesn't matter what the user enters. And in the case of manual dives, we use the user's data... (since we don't have the DC ;-)). In the case of planned dives, I don't know how it would look. I guess that it would be the same approach as for manual dives; 2- We use the input data and recalculate the profile at each change (more hard to implement). Data reported by DCs would only be initial values

Agreed on this. I think 1. will be relatively low effort, as we can use the virtual dive computer to inject user entered values for planned dives. And the bug that the user entered surface pressure is used should be easy to fix as well.

On a side note, I am struggling to understand why we even give the user the option to manually enter a surface pressure if their dive computer is reporting one - what comes from the dive computer is a _measured_ value, so unlike salinity it is not likely to be impacted by misconfiguration. @atdotde do you know the context of why we are allowing this?

rafsalvioni:
> Agreed on this. I think 1. will be relatively low effort, as we can use the virtual dive computer to inject user entered values for planned dives. And the bug that the user entered surface pressure is used should be easy to fix as well.

@mikeller I did this in https://github.com/subsurface/subsurface/commit/b060c388f14495fcf1a37732ff8bf0b9aa2e4630. What do you think?

mikeller:
@rafsalvioni: Yes, this is looking a lot better in my opinion - now we do not get the discrepancy in MOD when entering a surface pressure of 700 mbar on a dive with a recorded surface pressure of 1010 mbar:

image

And for dive planned with the dive planner we still get a MOD that is adjusted to the selected surface pressure / altitude (700 mbar in this case):

image

However, for manually entered dives (without the planner) the MOD is not adjusted - so we probably need to have a look at where the data is taken from in this case.

mikeller:
Ah yes, got it. When saving a planned dive we get this:

```
  <divecomputer model='planned dive' last-manual-time='45:00 min'>
  <depth max='15.0 m' mean='11.636 m' />
  <surface pressure='0.7 bar' />
  <water salinity='1022 g/l' />
  <sample time='0:00 min' depth='0.0 m' pressure='206.843 bar' />
```

Whereas for a manually entered dive we get this:

```
  <divecomputer model='manually added dive' last-manual-time='40:00 min'>
  <depth max='15.0 m' mean='12.997 m' />
```

So in my opinion a good way to get the manually entered surface pressure / salininity properly taken into account for manually entered dives will be to have these values added to the dive computer data model for these dives.

rafsalvioni:
> However, for manually entered dives (without the planner) the MOD is not adjusted - so we probably need to have a look at where the data is taken from in this case.

I found a function in divecomputer.c that test a "manual" DC. I am using it to test manual dives to choose the right values in calculation.

To manual dives (i.e. user values), salinity and surface pressure are stored in `dive->user_salinity` and `dive->surface_pressure` respectively.

>So in my opinion a good way to get the manually entered surface pressure / salininity properly taken into account for manually entered dives will be to have these values added to the dive computer data model for these dives.

I thought about it, but I was worried about the impact it might have on saving the logbook, since on manual dives these values are stored in different places. I thought the way I did it caused less impact... Maybe it wasn't as elegant...

In SSRF, will be stored in "watersalinity" and "airpressure" attribs in `<dive>`element.

rafsalvioni:
> However, for manually entered dives (without the planner) the MOD is not adjusted - so we probably need to have a look at where the data is taken from in this case.

I found a function in divecomputer.c that test a "manual" DC. I am using it to test manual dives to choose the right values in calculation.

To manual dives (i.e. user values), salinity and surface pressure are stored in `dive->user_salinity` and `dive->surface_pressure` respectivally.

In SSRF, will be stored in "watersalinity" and "airpressure" attribs in `<dive>`element.

mikeller (commented):
This is looking good now.

I assume the CSV changes are to make the tests pass? Unfortunately it looks like tests are still failing for some of the platforms. :-(

rafsalvioni:
> I assume the CSV changes are to make the tests pass?

Yes, thats it. Errors were in TestProfile. So I did a new CSV profile references considering modifications, once I edited mod calculation

> Unfortunately it looks like tests are still failing for some of the platforms

So... I dont what happen... I did tests locally before send commit... Whole tests were passed..

```
On branch fix_mbar2depth
Your branch is up to date with 'origin/fix_mbar2depth'.

Test project /home/rafael/Dev/C/subsurface/build
      Start  1: TestUnitConversion
 1/29 Test  #1: TestUnitConversion ...............   Passed    0.01 sec
      Start  2: TestProfile
 2/29 Test  #2: TestProfile ......................   Passed    0.57 sec
      Start  3: TestGpsCoords
 3/29 Test  #3: TestGpsCoords ....................   Passed    0.05 sec
      Start  4: TestParse
 4/29 Test  #4: TestParse ........................   Passed    3.43 sec
      Start  5: TestAirPressure
 5/29 Test  #5: TestAirPressure ..................   Passed    0.04 sec
      Start  6: TestHelper
 6/29 Test  #6: TestHelper .......................   Passed    0.04 sec
      Start  7: TestParsePerformance
 7/29 Test  #7: TestParsePerformance .............   Passed   18.73 sec
      Start  8: TestPlan
 8/29 Test  #8: TestPlan .........................   Passed    0.29 sec
      Start  9: TestDiveSiteDuplication
 9/29 Test  #9: TestDiveSiteDuplication ..........   Passed    0.04 sec
      Start 10: TestRenumber
10/29 Test #10: TestRenumber .....................   Passed    0.03 sec
      Start 11: TestPicture
11/29 Test #11: TestPicture ......................   Passed    0.03 sec
      Start 12: TestMerge
12/29 Test #12: TestMerge ........................   Passed    0.03 sec
      Start 13: TestTagList
13/29 Test #13: TestTagList ......................   Passed    0.00 sec
      Start 14: TestQPrefCloudStorage
14/29 Test #14: TestQPrefCloudStorage ............   Passed    0.42 sec
      Start 15: TestQPrefDisplay
15/29 Test #15: TestQPrefDisplay .................   Passed    0.41 sec
      Start 16: TestQPrefDiveComputer
16/29 Test #16: TestQPrefDiveComputer ............   Passed    0.28 sec
      Start 17: TestQPrefDivePlanner
17/29 Test #17: TestQPrefDivePlanner .............   Passed    0.91 sec
      Start 18: TestQPrefGeneral
18/29 Test #18: TestQPrefGeneral .................   Passed    0.26 sec
      Start 19: TestQPrefLog
19/29 Test #19: TestQPrefLog .....................   Passed    0.34 sec
      Start 20: TestQPrefEquipment
20/29 Test #20: TestQPrefEquipment ...............   Passed    0.19 sec
      Start 21: TestQPrefMedia
21/29 Test #21: TestQPrefMedia ...................   Passed    0.22 sec
      Start 22: TestQPrefGeocoding
22/29 Test #22: TestQPrefGeocoding ...............   Passed    0.35 sec
      Start 23: TestQPrefLanguage
23/29 Test #23: TestQPrefLanguage ................   Passed    0.39 sec
      Start 24: TestQPrefPartialPressureGas
24/29 Test #24: TestQPrefPartialPressureGas ......   Passed    0.34 sec
      Start 25: TestQPrefProxy
25/29 Test #25: TestQPrefProxy ...................   Passed    0.35 sec
      Start 26: TestQPrefTechnicalDetails
26/29 Test #26: TestQPrefTechnicalDetails ........   Passed    1.49 sec
      Start 27: TestQPrefUnits
27/29 Test #27: TestQPrefUnits ...................   Passed    0.40 sec
      Start 28: TestQPrefUpdateManager
28/29 Test #28: TestQPrefUpdateManager ...........   Passed    0.30 sec
      Start 29: TestQML
29/29 Test #29: TestQML ..........................   Passed    0.04 sec

100% tests passed, 0 tests failed out of 29

Total Test time (real) =  30.01 sec

```

Would you know?

mikeller:
After having a look at the passing / failing checks, it looks like all of the checks that are running the tests are failing.

A quick check running the tests on your pull request, in the way they are run by the build runner show that the tests are failing on my machine (ubuntu 23.04) as well [test log](https://github.com/subsurface/subsurface/files/11649159/LastTest.log):

```
$cd subsurface/build/tests/
$make check
[...]
Test project /home/mikeller/git/subsurface/build/tests
      Start  1: TestUnitConversion
 1/29 Test  #1: TestUnitConversion ...............   Passed    0.02 sec
      Start  2: TestProfile
 2/29 Test  #2: TestProfile ......................***Failed    0.47 sec
      Start  3: TestGpsCoords
 3/29 Test  #3: TestGpsCoords ....................   Passed    0.06 sec
      Start  4: TestParse
[...]
```


Are you sure that you have pushed all the changes that you made to get the tests to pass?
How are you running the tests locally?

rafsalvioni:
> Are you sure that you have pushed all the changes that you made to get the tests to pass?

Yes... This is my branch status:
```
On branch fix_mbar2depth
Your branch is up to date with 'origin/fix_mbar2depth'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        ../.cproject
        ../.project
        ../Testing/
        ../googlemaps/
        ../install-root/
        ../mobile-widgets/3rdparty/ECM/
        ../mobile-widgets/3rdparty/breeze-icons/
        ../mobile-widgets/3rdparty/extra-cmake-modules/
        ../mobile-widgets/3rdparty/kirigami/

```

> How are you running the tests locally?

Below my environment:

```
$: cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.2 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.2 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
$: make --version
GNU Make 4.3
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
$: cmake --version
cmake version 3.22.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).

``` 
 Yesterday I ran the tests using `ctest`. Today I used `make test`. Same results... All tests passed...

```
$: cd ~/Dev/C/subsurface/build
$: make test
Running tests...
Test project /home/rafael/Dev/C/subsurface/build
      Start  1: TestUnitConversion
 1/29 Test  #1: TestUnitConversion ...............   Passed    0.01 sec
      Start  2: TestProfile
 2/29 Test  #2: TestProfile ......................   Passed    0.56 sec
      Start  3: TestGpsCoords
 3/29 Test  #3: TestGpsCoords ....................   Passed    0.04 sec
[......]
      Start 29: TestQML
29/29 Test #29: TestQML ..........................   Passed    0.03 sec

100% tests passed, 0 tests failed out of 29

Total Test time (real) =  29.68 sec

```

mikeller:
This has got me stumped a bit - your versions are probably closer to the ones used in the build runners than mine, yet the tests are failing on the build runners and on my machine.

I think as a reasonable next step, can you try checking out your branch into a new directory and building and running the tests from there?

mikeller:
:-D Good find - the fact that GitHub refuses to show a diff for the CSV files does not help.

rafsalvioni:
@mikeller It was a locale issue.... My machine is in pt-BR... here we use "," as decimal separator. But TestProfile output uses "." as separator (for example, in your machine). Thats why in my computer the test passed and in your didnt...

I just replace "\d,\d" to "\d\.\d" in files

mikeller (approved):

rafsalvioni:
@mikeller There is another bug in profile...

In a manual dive, when I change salinity, cylinder's MOD and profile's MOD (information window) are equal. However, if I change surface pressure, profile's MOD dont changed. Just cylinder's MOD.

I analyzed profile.c, did some tests but I didnt find the issue...

mikeller:
From a quick look I suspect that this might be caused by `dc->surface_pressure` being used in places in the profile, and us not updating this for manually added dives:

https://github.com/subsurface/subsurface/blob/3939cc8da28e7ab39e0bed22c9ce2298f1a72cbd/core/profile.c#L928

rafsalvioni:
I started the search in:
https://github.com/subsurface/subsurface/blob/3939cc8da28e7ab39e0bed22c9ce2298f1a72cbd/core/profile.c#L1379-L1380

After, the point where `entry->mod` is filled:
https://github.com/subsurface/subsurface/blob/3939cc8da28e7ab39e0bed22c9ce2298f1a72cbd/core/profile.c#L1196

So, I got the point where MOD is calculated... And that was modified in this PR:
https://github.com/rafsalvioni/subsurface/blob/fd10d1d072b9447a664a04c14cd391efc8e41931/core/dive.c#L3313-L3320

Well, looks fine to me... So I thought that the issue is in `dive->surface_pressure` input:
https://github.com/subsurface/subsurface/blob/3939cc8da28e7ab39e0bed22c9ce2298f1a72cbd/desktop-widgets/tab-widgets/TabDiveInformation.cpp#L468
and
https://github.com/subsurface/subsurface/blob/3939cc8da28e7ab39e0bed22c9ce2298f1a72cbd/commands/command_edit.cpp#L290-L293

Looks fine to me again... And issue persists

mikeller:
Yeah, this looks all sane indeed. The only explanation that I can come up with would be that somehow the data in `dive` in `gas_mod(gasmix, modpO2, dive, 1).mm` must be wrong.
One way to investigate will be to output some of the data in `dive` to `stderr` in order to debug what is going on.

rafsalvioni:
The problem was in the planner...

Manual dives are using the profile in edit mode. To enter this mode, he made a kind of dive plan and this plan changed the surface pressure of the dive to the standard pressure. Therefore, the MOD calculation would consider the standard pressure... A chain of problems...

dirkhh:
@atdotde - any comments?

atdotde:
I am currently traveling without a computer. Please let me look at this once I am back next week.

dirkhh:
Just a quick ping to @atdotde 
Of course, now I am traveling - at least this morning I found time to go through some of the open PRs :)

mikeller:
@atdotde: Can you please have another look at this - to me the changes look ok.

atdotde:
Sorry for the long delay. I agree it looks fine.

mikeller:
@rafsalvioni: This now needs a rebase to resolve the conflicts.

PR #3923, opened 2023-06-12
Desktop: Use Persisted Device Information for Dive Computer Configuration.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Use the dive computer / device information that was persisted when previously downloading dives or configuring the dive computer in the dive computer configuration dialog.

### Changes made:

1. change the list of supported devices to an `std::array`;
2. use the persisted device information to try and find the device when opening the configuration dialog;
3. disable bluetooth device selection for devices that do not support bluetooth;
4. preserve the entered device name when repopulating the device list;
5. made the use of bluetooth dependent on the format of the device being a bluetooth address;
6. move persisting of device information to after a successful connection attempt;
7. rename 'Connect with bluetooth' and 'Cancel' buttons in the dialog to make them more consistent with what they do.

### Release note:

desktop: use persisted device information for the dive computer configuration

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.cpp
+++ b/desktop-widgets/configuredivecomputerdialog.cpp
@@ -8,6 +8,7 @@
 // For fill_computer_list, descriptorLookup
 #include "core/downloadfromdcthread.h"
 
+#include <array>
 #include <QFileDialog>
 #include <QMessageBox>
 #include <QNetworkReply>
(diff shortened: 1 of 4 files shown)


desktop-widgets/configuredivecomputerdialog.cpp
```diff
@@ -98,6 +99,32 @@ void GasTypeComboBoxItemDelegate::setModelData(QWidget *editor, QAbstractItemMod
...
+	DiveComputerEntry(QString vendor, QString product, unsigned int transport, bool fwUpgradePossible) :
+		vendorInt(vendor), productInt(product), transportInt(transport), fwUpgradePossibleInt(fwUpgradePossible) {}
+
+	QString vendor() const { return vendorInt; }
+	QString product() const { return productInt; }
+
+	unsigned int transport() const { return transportInt; }
+	bool fwUpgradePossible() const { return fwUpgradePossibleInt; }
+private:
+	QString vendorInt;
```
dirkhh, line 114: while I admit this is silly - my brain immediately assumes that `vendorInt` and `productInt` are... integers. Which at first confused me when I read this. Not a big deal, but I'm curious... does this stand for "internal" or what does the "Int" designate?
I think elsewhere we tried using `m_vendor` or something like this to indicate member names...

dirkhh (approved):
I think this is great.
@glance-  -- any comments from you?

glance-:
LGTM.

desktop-widgets/configuredivecomputerdialog.cpp
mikeller: Ha, the naming was a 'quick fix' that I put in after adding the getters caused build errors due to naming clashes with the private members, with the intent to figure out the naming convention for private members later on. But then I forgot. 😅

Anyhow, a cleanup is in #3925.

PR #3925, opened 2023-06-13
Cleanup: Use Correct Naming for Private Members.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Rename inconsistently named private members introduced in #3923.

### Changes made:

1. rename inconsistently named private members introduced in #3923.

### Related issues:

https://github.com/subsurface/subsurface/pull/3923#discussion_r1226824278

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.cpp
+++ b/desktop-widgets/configuredivecomputerdialog.cpp
@@ -99,31 +99,19 @@ void GasTypeComboBoxItemDelegate::setModelData(QWidget *editor, QAbstractItemMod
 		QStyledItemDelegate::setModelData(editor, model, index);
 }
 
-class DiveComputerEntry
-{
-public:
-	DiveComputerEntry(QString vendor, QString product, unsigned int transport, bool fwUpgradePossible) :
-		vendorInt(vendor), productInt(product), transportInt(transport), fwUpgradePossibleInt(fwUpgradePossible) {}
-
-	QString vendor() const { return vendorInt; }
-	QString product() const { return productInt; }
-
-	unsigned int transport() const { return transportInt; }
-	bool fwUpgradePossible() const { return fwUpgradePossibleInt; }
-private:
-	QString vendorInt;
-	QString productInt;
-
-	unsigned int transportInt;
-	bool fwUpgradePossibleInt;
+struct DiveComputerEntry {
+	QString vendor;
+	QString product;
+	unsigned int transport;
+	bool fwUpgradePossible;
 };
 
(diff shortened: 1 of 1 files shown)


bstoeger:
*shrug*
We seem to use `m_` only in a minority of the cases anyway.

What bothers me more: Is there any reason this class couldn't be a simple struct without getters? At first sight, this appears like complexity for no gain to me...

mikeller:
> _shrug_ We seem to use `m_` only in a minority of the cases anyway.

So what is the preferred naming pattern for private members that are exposed by getters / setters then?
 
> What bothers me more: Is there any reason this class couldn't be a simple struct without getters? At first sight, this appears like complexity for no gain to me...

I guess I tried too hard to use C++ instead of C - happy to change to a simple `struct`.

bstoeger:
> I guess I tried too hard to use C++ instead of C - happy to change to a simple `struct`.

I would much prefer that. There are situations were setters/getters are called for. For example in shared resources, where the user must not change the value. Or when implementing the PIMPL idiom. Here, we have a classical value type, which is only used locally, so there is no upside to the added complexity. Also, this is more of a Java-ism than a C++-ism.

mikeller:
> Also, this is more of a Java-ism than a C++-ism.

Figures. My experience with Java is a lot deeper and more recent than with C++.

bstoeger:
I like that much better. The commit description is now somewhat off, but fine with me.

Notes on the original PR:
- Probably the combobox should be filled dynamically.
- Any reason that the array is not const?
- Using `DiveComputerEntry supportedDiveComputers[] = {...}` would allow omission of the fixed size. It can by accessed (at compile time) by `std::size(supportedDiveComputers)`.
- More idiomatic, but admittedly also more cryptic would have been
```
        auto it = std::find_if(supportedDiveComputers.begin(), supportedDiveComputers.end(), [](const DiveComputerEntry &e) { ... });
        if (it != supportedDiveComputers.end())
                selectedDiveComputerIndex = it - supportedDiveComputers.begin();
```
I reckon, we should write small helpers for these recurring idioms, but so far I didn't deem it necessary.

bstoeger:
Also note: The `struct` in `std::array<struct DiveComputerEntry, 4>` is not necessary.

The only technical difference between `struct` and `class` is that the former starts with public members, the latter with private members. Apart from that, they are interchangeable.  One could muse forever on the semantic difference, but let's not do that...

PR #3926, opened 2023-06-14
Desktop: Add Meaningful Error Messages for libdivecomputer Dump.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add meaningful error messages when creating a libdivecomputer dump. In particular show if creating a dump is not supported on the dive computer that is used.

### Changes made:

1. add meaningful error messages when creating a libdivecomputer dump.

### Related issues:

https://github.com/subsurface/libdc/pull/51

### Additional information:

image

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -1138,13 +1138,23 @@ static const char *do_device_import(device_data_t *data)
 		}
 
 		dc_buffer_free(buffer);
+
+		if (rc != DC_STATUS_SUCCESS) {
+			progress_bar_fraction = 0.0;
+
+			if (rc == DC_STATUS_UNSUPPORTED)
+				return translate("gettextFromC", "Dumping not supported on this device");
+
+			return translate("gettextFromC", "Dive data dumping error");
+		}
 	} else {
 		rc = dc_device_foreach(device, dive_cb, data);
-	}
 
-	if (rc != DC_STATUS_SUCCESS) {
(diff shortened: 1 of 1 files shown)


core/libdivecomputer.c
```diff
@@ -1125,6 +1125,7 @@ static const char *do_device_import(device_data_t *data)
 	if (rc != DC_STATUS_SUCCESS)
 		return translate("gettextFromC", "Error registering the cancellation handler.");
 
+	const char *result = NULL;
```
bstoeger: What's the point of this more complex control flow? Because it is a C-File without RAII?

mikeller: I did this more 'out of habit' than anything else - in general I find that code without early returns throughout the functionality is easier to maintain once the functions become large and complex. But in this case it's really just an 'error out' pattern and could be changed back to early returns.

bstoeger: I thought the original was much easier to follow.

Yes, in C this technique may be needed to clean up resources, but here thankfully that is not the case.

If early returns become too complex,  it's probably time to split into multiple functions. In fact, I have seen people use local lambdas to emulate early return-like behavior.

mikeller: I'll change it back.

bstoeger:
Thanks.

PR #3927, opened 2023-06-16
Mobile: Make Cloud Auto Synch Status Easier to Understand.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] UX Improvement

Add an explicit checkbox to the 'cloud auto synch' status toggle in the mobile 'Dive Management' menu in order to make it more intuitive to understand.
This isn't super pretty, but I think it will improve the usability. A prettier way to achieve this will be to redesign the `ic_cloud_off.svg` / `ic_cloud_done.svg` in order to make them intuitively recognisable as unchecked / checked checkboxes.

### Changes made:

1. add an explicit checkbox to the 'cloud auto synch' status toggle in the mobile 'Dive Management' menu in order to make it more intuitive to understand.

### Related issues:

Example of a support request caused by confusion from the current implementation: https://groups.google.com/g/subsurface-divelog/c/9X-hTt9NFlE/m/ZcqtdOOhBQAJ

### Additional information:

Screenshot_2023-06-16-15-06-06-678_org subsurfacedivelog mobile

Screenshot_2023-06-16-15-06-45-940_org subsurfacedivelog mobile

### DIFF
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -347,7 +347,7 @@ Kirigami.ApplicationWindow {
 				icon {
 					name: PrefCloudStorage.cloud_auto_sync ?  ":/icons/ic_cloud_off.svg" : ":/icons/ic_cloud_done.svg"
 				}
-				text: PrefCloudStorage.cloud_auto_sync ? qsTr("Disable auto cloud sync") : qsTr("Enable auto cloud sync")
+text: (PrefCloudStorage.cloud_auto_sync ? "\u2610 " : "\u2611 ") + qsTr("Auto cloud sync")
 					visible: Backend.cloud_verification_status !== Enums.CS_NOCLOUD
 					onTriggered: {
 						PrefCloudStorage.cloud_auto_sync = !PrefCloudStorage.cloud_auto_sync


bstoeger:
I don't use mobile, so I don't have any opinion on this. I assigned to potential reviewers.

dirkhh:
So I think this is the right idea. The bigger problem is that this needs to be re-tested on iOS and Android. At one point I had to completely disable it on one of the two platforms (I don't even recall which one) as it caused hung connections and could lead to damaged git repos...  I think it was iOS, but it's been a while ago.
And on the platform where we don't support it, we shouldn't offer the option at all...

mikeller:
@dirkhh: This is just a change of how the element is displayed in the UI, in a bid to make it less confusing to avoid more support tickets like https://groups.google.com/g/subsurface-divelog/c/9X-hTt9NFlE/m/ZcqtdOOhBQAJ.
I have tested it on android (see attached screenshots), but I don't have an iPhone and don't know if we have a way to run this in an emulator, so if somebody else can test the look of it on iPhone it would be great.

dirkhh:
That's fine. I was thinking bigger picture... Does it actually sync in the background as we claim it does 
I'll merge it

mikeller:
> Does it actually sync in the background as we claim it does I'll merge it

At least for android it definitely does.

PR #3930, opened 2023-06-18
Desktop: Load the Dive Computer List in 'Configure Dive Computer' Dynamically.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Load the dive computer list in the 'Change Settings on Dive Computer' dialog dynamically.
Also incorporate suggestions from
https://github.com/subsurface/subsurface/pull/3925#issuecomment-1595784076.

### Changes made:

1. use constant array for dive computer list;
2. load the dive computer list in the 'Change Settings on Dive Computer' dialog dynamically.

### Mentions:

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.cpp
+++ b/desktop-widgets/configuredivecomputerdialog.cpp
@@ -104,16 +104,17 @@ struct DiveComputerEntry {
 	QString product;
 	unsigned int transport;
 	bool fwUpgradePossible;
+	bool forcedFirmwareUpgradeSupported;
 };
 
 //WARNING: Do not edit this list or even just change its order without
 // making corresponding changes to the `DiveComputerList` element
 // in `configuredivecomputerdialog.ui` or this functionality will stop working.
 static const DiveComputerEntry supportedDiveComputers[] = {
-	{ "Heinrichs Weikamp", "OSTC 2N", DC_TRANSPORT_SERIAL, true },
-	{ "Heinrichs Weikamp", "OSTC Plus", DC_TRANSPORT_BLUETOOTH, true },
-	{ "Heinrichs Weikamp", "OSTC 4", DC_TRANSPORT_BLUETOOTH, true },
-	{ "Suunto", "Vyper", DC_TRANSPORT_SERIAL, false },
+	{ "Heinrichs Weikamp", "OSTC 2N", DC_TRANSPORT_SERIAL, true, false },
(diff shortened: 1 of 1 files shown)


glance-:
I don't see the upside with this. I've been working with the ui in designer, and not having the selection panel there will just be a pita.

So, no, I don't like moving this out of the ui files.

bstoeger:
> I don't see the upside with this. I've been working with the ui in designer, and not having the selection panel there will just be a pita.
> 
> So, no, I don't like moving this out of the ui files.

The point is - the data is now found in two places. If you edit the ui file, you must edit the C++ file. This change makes it so that you only have to edit the C++ file. Seems reasonable to me.

glance-:
For changes in all ui components you need to change both the .ui files, and plumb up the values to the c++ code to make anything happen. This only changes this for a small fraction of the ui logic, with the downside of making testing of this directly in the qt-designer harder because the computer switcher/chooser thingie isn't there any longer.


There are 4 computers supported, and that haven't changed in ages, since I added the Ostc4 like... 5? years ago. It won't change that much in the future either. Shure, if it where code that was changed every week or so, fine. then it would get out of sync often enough but this is really low traffic code who now is changed the first time in 5 years.

So, slight inconvenience in keeping ui and cpp in sync, which they need to be for _everything else_ already is a tiny price for making the ui still testable standalone in qt-designer.

bstoeger:
Ok, to me that's a very foreign way of doing things, but fair enough. In fact, I stopped using QtCreator and edit these XML-files by hand. Since I'll never use this part of the code, it's not my call.

dirkhh:
I'm leaning towards @glance- 's argument here. There really are only four supported dive computers and it's not likely to change. This is such an odd, rarely used feature... I don't think it's worth making his workflow harder - and he's the main person working on this...

mikeller:
Ok, tried to salvage the improvements from this, and added a warning about the reliance on matching indices across multiple files.

glance-:
I'd suggest you break out the forcedFirmwareUpgradeSupported changes to a separate commit, but other than that, it LGTM.

mikeller:
@glance-: Done.

glance- (approved):

dirkhh (approved):
looks good

PR #3937, opened 2023-07-09
Equipment: Include Unused Tanks in Merge if Preference is Enabled.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Include unused tanks in merges of multiple logs into a single dive if the 'Show unused cylinders' preference is enabled. Also rename the preference (in code) to `include_unused_tanks` to reflect the fact that it is already used in more places than just the display (exporting, cloning dives).
Simplified the cylinder model to make forced inclusion of unused tanks dependent on use of the model in planner.
Leaving the persisted name of the preference as `display_unused_tanks` to avoid resetting this for all users - is there a good way to migrate preference names?

### Changes made:

1. include unused tanks in merges of multiple logs into a single dive if the 'Show unused cylinders' preference is enabled.
2. rename the preference (in code) to `include_unused_tanks` to reflect the fact that it is already used in more places than just the display (exporting, cloning dives);
3. simplify the cylinder model to make forced inclusion of unused tanks dependent on use of the model in planner;

### Additional information:

image

### Release note:

equipment: include unused cylinders in merged dive if the preference is enabled

### Documentation change:

Updated.


mikeller:
Does anybody have any objections to this being merged? Or has got any comments for a review?

PR #3939, opened 2023-07-09
Equipment: Fix 'used' Gas Selection for CCR Dives.

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix how gases are marked as 'used' and kept from being deleted in the equipment tab for CCR dives.
It does not make sense to treat the (arbitrary) first gas in the list with a usage type of 'diluent' or 'oxygen' as 'used' and prevent the user from deleting it. Dive computers report the initial diluent and any other diluents used through a 'gaschange' event, so the actually used diluents are already picked up as part of gaschange event based logic.
Also clarify the selection of the first diluent used as a default if no gaschange events exist.
Also fixed the test data - gases that have a pressure change should be 
included in the profile if they do not have a gas change recorded 
against them by other dive computers, even if they are oxygen. 
A secondary problem shown by this is that the pressure change is not 
applied to the profile - the pressure is currently shown as constant on
the start pressure. But this is for another pull request.

### Changes made:

1. fix how gases are marked as 'used' and kept from being deleted in the equipment tab for CCR dives;
2. clarify the selection of the first diluent used as a default if no gaschange events exist;
3. fixed the test data.

### Additional information:

Before:
image

After:
image

### Release note:

equipment: fix bug showing the first diluent in the gas list as 'used' for CCR dives

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 equipment: mark gas mixes reported by the dive computer as 'inactive' as 'not used'
 equipment: include unused cylinders in merged dive if the preference is enabled
+equipment: fix bug showing the first diluent in the gas list as 'used' for CCR dives
 desktop: added button to hide the infobox in the dive profile
 desktop: use persisted device information for the dive computer configuration
 core: fix bug using salinity and pressure values in mbar <-> depth conversion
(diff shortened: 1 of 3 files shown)


mikeller:
Does anybody have any objections to this being merged? Or has got any comments for a review?

bstoeger:
No objection from me, but I don't do multi-gas dives, so 🤷.

Just an idle comment the placement of `is_gas_used` etc. in `statistics.c` is weird...

Let's page @dirkhh.

PR #3948, opened 2023-07-27
Desktop: Consider Severity when Hiding Events.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When events are hidden in the profile, only hide events with the same name and the same severity (flags).

### Changes made:

1. when events are hidden in the profile, only hide events with the same name and the same severity (flags);
2. split out hiding of single events into its own menu item 'Hide event';
3. remove confirmation dialogue that is now inappropriate.

### Related issues:

From discussion in https://github.com/subsurface/libdc/pull/54.

### Release note:

desktop: hide only events with the same severity when 'Hide similar events' is used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: hide only events with the same severity when 'Hide similar events' is used
 equipment: mark gas mixes reported by the dive computer as 'inactive' as 'not used'
 equipment: include unused cylinders in merged dive if the preference is enabled
 equipment: fix bug showing the first diluent in the gas list as 'used' for CCR dives
--- a/core/eventname.cpp
+++ b/core/eventname.cpp
@@ -39,13 +39,6 @@ extern "C" bool is_event_hidden(const char *eventname)
 	return it != event_names.end() && !it->plot;
 }
 
-extern "C" void hide_event(const char *eventname)
-{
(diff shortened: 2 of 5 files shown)


dirkhh (commented):
I think the title is a bit misleading... you are also removing a dialog asking the user what they want to do. Which is a behavioral change as they no longer can hide a single event of a certain type.

I'm not saying that's wrong, I'm just pointing out that this isn't captured in the commit message and surprised me when I looked at the code.

mikeller:
Ah, my bad. I was goaded by the UI into thinking that the functionality of this menu action was to hide similar events, and that the dialogue was just a (in my opinion unnecessary) confirmation for something that isn't a permanent change to the data:

image

I honestly have doubts that anybody will realise that clicking 'Cancel' in a dialogue asking them 'Hide all ascent events?' will hide only a single event - this is as unintuitive as it can be (plus the code is obfuscating this by the duality of `hide_event(event->name)` and `item->hide()`).

image

Also, 'Hide all ascent events of severity 8?' isn't the most helpful dialogue.

Taking this into consideration, my proposal is to leave the confusing dialogue out, and instead of add a 'Hide event' menu item.

mikeller:
Like so:
image

PR #3950, opened 2023-08-07
fix: merge_pressure does not calculate starting pressures correctly

mwerle:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The existing logic correctly calculates the minimum (ie, ending) pressure, but not the maximum (ie starting) pressure.

For example, 2 tanks A and B with manual pressures (same tank on subsequent dives, which were then merged):
A: 205 - 84
B: 83 - 55

When merging the starting pressures, the call is : merge_pressure(205, 0, 83, 0, false)

The final comparison is:
  if(false && 205 < 83) return 205;
  else return 83;

-> So 83 is returned even though 205 should have been.
### Changes made:

1) fix the logic of the final check to see which value should be returned based on the value of the `take_min` flag.

### Additional information:

Issue can be replicated by merging dives #2 and #3 using the attached sample log.

Q: We should really add unit testing to the project; not just the current test scenarios. What would be the best way to do so?

### Release note:

n/a; minor fix.

### Documentation change:

n/a; internal change.

### Mentions:

[fix_merge_pressure.ssrf.zip](https://github.com/subsurface/subsurface/files/12274911/fix_merge_pressure.ssrf.zip)

### DIFF
--- a/core/dive.c
+++ b/core/dive.c
@@ -1852,10 +1852,9 @@ static pressure_t merge_pressures(pressure_t a, pressure_t sample_a, pressure_t 
 		a = b;
 	if (!b.mbar)
 		b = a;
-	if (take_min && a.mbar < b.mbar)
-		return a;
-	else
-		return b;
+	if(take_min)
+		return a.mbar < b.mbar? a : b;
+	return a.mbar > b.mbar? a : b;
 }
 
 /*


dirkhh:
We have a bunch of tests already, adding things like this would be easy.
Load a dive file, call the undo function to merge them, compare to the desired output...
Or test the function directly, from within our test harness.

dirkhh (commented):
So the old logic was to simply take b if take_min wasn't set.
It might be worth at least a quick look if there was ever a reason for that logic

mwerle:
I know there's a bunch of `scenario` tests.. I was meaning actual real unit tests able to test individual functions for their expected output given a bunch of test input, especially for edge-cases.  I didn't see an (easy) way to add them but then again I have limited experience with QT Creator/QT.

>  It might be worth at least a quick look if there was ever a reason for that logic

I noticed it was an issue when I merged two dives and the resulting SAC was completely wrong.  How can we check the historical reasoning behind the incorrect implementation given no comments justifying it? I guess can always ask the person who implemented it..?

dirkhh:
> I know there's a bunch of `scenario` tests.. I was meaning actual real unit tests able to test individual functions for their expected output given a bunch of test input, especially for edge-cases. I didn't see an (easy) way to add them but then again I have limited experience with QT Creator/QT.

I have never tried to do tests with QtCreator... I've always done that from the command line.
And I am anything but an expert... yet the couple of people with deep Qt knowledge who worked on the project in the past have all moved on.
So most of the changes I made to that code in the last few years have been of the GTER approach - Google, Try, Error, Repeat.

> > It might be worth at least a quick look if there was ever a reason for that logic
> 
> I noticed it was an issue when I merged two dives and the resulting SAC was completely wrong. How can we check the historical reasoning behind the incorrect implementation given no comments justifying it? I guess can always ask the person who implemented it..?

We TRY to have decent commit messages. And yes, worst case it should at least tell you who wrote it.
Frankly, since it's obviously broken, I don't think it would be wrong to merge - but I often found that understanding what was the original intent lead to better fixes.

mwerle:
@bstoeger : Could you please explain your reasoning when implementing `merge_pressures` in commit 6729428b6a942f83892129d7b8d1e5ef04930d05 ?  The function doesn't do what the comment says it should do.

bstoeger:
Sorry, I can't remember writing that. Probably just a stupid mistake.

mwerle:
> Sorry, I can't remember writing that. Probably just a stupid mistake.

Thanks for the confirmation :)

I've tried to make a set of unit tests for this function, but running into "fun" issues between C and C++.. it would be nice if the project was one or the other (and yes, I realise it's historical..).  But for now I've given up. If someone has a good suggestion on how to do it I'd appreciate it.

My approach was to write a QTest test application, and the .cpp did a `#include "/core/dive.c"` (in order to access the `static` functions for testing) - and in hindsight it's rather obvious that the compilation blows up in all sorts of interesting ways because we're now trying to compile that file as C++ instead of C.

bstoeger:
Frankly, I don't understand what is wrong with how the tests currently work. They are automatically run by "continuous integration" and you can run them explicitly from the command line. I suggest extending the current tests with whatever test case you can come up with. I see no reason why one couldn't implement more advanced techniques such as fuzzing in the current framework...

mwerle:
> Frankly, I don't understand what is wrong with how the tests currently work. They are automatically run by "continuous integration" and you can run them explicitly from the command line. I suggest extending the current tests with whatever test case you can come up with. I see no reason why one couldn't implement more advanced techniques such as fuzzing in the current framework...

There is nothing "wrong" with the current tests - but I'm unable to write unit-tests for `merge_pressures` as it's a `static` (ie private) function inside `dive.c`.  There's an argument to be made that tests should only test the public interface, not the private implementation..    but conversely there's an argument that we should validate that the various parts of the implementation do what they are supposed to do as well to avoid obscure errors which may be difficult to track down when only testing at a high level. I guess I could write a small C test wrapper for the internal functions which the actual tests then call.

Anyway, this should be a separate discussion; sorry for derailing the actual PR.

core/dive.c
```diff
@@ -1852,10 +1852,9 @@ static pressure_t merge_pressures(pressure_t a, pressure_t sample_a, pressure_t
 		a = b;
 	if (!b.mbar)
 		b = a;
-	if (take_min && a.mbar < b.mbar)
-		return a;
-	else
-		return b;
+	if(take_min)
```
mikeller, line 1855: The Subsurface [coding style](https://github.com/subsurface/subsurface/blob/master/CODINGSTYLE.md#basic-rules) is ` if (<condition>)` with a space before the brace.

core/dive.c
```diff
@@ -1852,10 +1852,9 @@ static pressure_t merge_pressures(pressure_t a, pressure_t sample_a, pressure_t
...
+		return a.mbar < b.mbar? a : b;
```
mikeller, line 1856: The commonly used form throughout Subsurface is `<condition> ? <true case> : <false case>` with a space before the `?`.

mikeller (changes requested):

core/dive.c
mwerle: Thank you; fixed.  Too many coding styles :)

I'll see if I can create a pre-commit hook to check the style (it's what my new employer does), although it might generate an extremely large number of changes depending on how unified the coding style is at present.

core/dive.c
mwerle: Thank you; fixed.

core/dive.c
dirkhh: > Thank you; fixed. Too many coding styles :)
> 
> I'll see if I can create a pre-commit hook to check the style (it's what my new employer does), although it might generate an extremely large number of changes depending on how unified the coding style is at present.

The coding style is different for different parts of the code. In part for historical reasons, in part to make merging with outside code easier, in part because we weren't always consistently enforcing it.

I appreciate it every time someone pays attention with what we intended to do.

mwerle: So there's already a `.clang-format` file in the repository; it would be trivial to run it on commit on the files which were changed. 

As far as current mis-match is concerned, we could always have a single pass over the entire codebase and fix up the style everywhere.  Although I must say I'm personally not a 100% fan of automated restyling tools - there are always situations in which hand-formatted code and comments are MUCH more readable even if it breaks the general coding guidelines/style.

Not sure if it's (easily) possible to have a different style for different parts of the codebase if maintaining merge-compatibility remains a requirement.

mikeller (approved):

mwerle:
FWIW, I played around with it and managed to create a set of unit-tests for `merge_pressures`. Not sure if it's the best approach (in fact, I'm positive there's better ways), but if people are interested in taking a look and commenting, I've created a new personal branch: https://github.com/subsurface/subsurface/commit/35ec11040d1215ed4927a90b5adf13b3d8ef242d

bstoeger:
To be honest, that looks significantly over-engineered to me. If testing this particular function is such a concern to you, just make it non-static. We have countless C functions in the global name space. One more doesn't matter.

However, we have probably thousands of functions of this complexity level - writing tests for all of them appears counterproductive to me. A better approach would be to extend the already existing `TestMerge` with test cases that test the various code-paths of this function. That would appear much more useful to me, as it will also catch bugs on other levels. If merging breaks, finding the cause is trivial and therefore I seriously doubt the usefulness of such fine-grained tests.

mwerle:
> To be honest, that looks significantly over-engineered to me. If testing this particular function is such a concern to you, just make it non-static. We have countless C functions in the global name space. One more doesn't matter.

It's more a case of trying to come up with a good way of unit-testing functionality in general instead of being obsessed about this one particular function. Unit tests are there to firstly ensure the function behaves as intended during development (avoiding mistakes as were made during the previous refactor) as well as testing against future regressions if/when related functionality is modified.

As I already mentioned, I agree that this particular function may be too low-level to concentrate on, on the other hand, the existing `TestMerge` is, IMHO, far too high-level. It relies on a significant chunk of functionality and is more what I would call an integration test or a scenario test rather than a unit test. It' seems to me to also be a lot of work to craft a large variety of test files in order to test all the underlying functionality, 99% of which is not required for an individual test case.  There's also no easy way to see what it is that's actually being tested as it's split between the test logic and the data in the files and the rest of the code.

In fact, having taken a quick look at the `TestMerge` data, I'm not even sure the test is testing what it says it is since the second data file doesn't even specify a cylinder at all!

----

Anyway; is there anything else which needs doing for this PR? I fixed up signing the commits (should be a pre-commit hook; I'll look into it); the failed CI was due to a network timeout and I don't have the ability to re-run it.

bstoeger:
PR is fine with me. I would just squash the three commits into one.

mwerle:
I dont (think I) have write-access to the repo for a squash-and-merge; I guess you need to do that? Or what is the procedure?

mikeller:
> I dont (think I) have write-access to the repo for a squash-and-merge; I guess you need to do that? Or what is the procedure?

You can squash it in your branch, and then it can be merged by a contributor.

bstoeger:
Yes, just squash it in your branch and force-push.

What I do in the console is `git rebase -i master`, mark the commits to be squashed and then `git push -f`.

But I'm sure that there are better ways of doing this...

mikeller:
> But I'm sure that there are better ways of doing this...

There is an option for a maintainer to do this as part of a merge, but it's disabled in this repository. In my experience, mixing squash and no-squash merges is problematic in git, as it can make it impossible to do cherry-picks, so having only the non-squash option enabled is probably better in the long run.

mwerle:
Done; right, I'm currently mostly working in an environment which uses squash-commits (personally not a fan), although there is a point to be made for squashing trivial commits which only add noise to the commit history.

bstoeger:
By some weird git magic, I fixed the reappeared coding style issue and merged this locally. I have no idea what I did, but it seemed to work.

Closing.

mwerle:
> By some weird git magic, I fixed the reappeared coding style issue and merged this locally. I have no idea what I did, but it seemed to work.
> 
> Closing.

Ok, so my rebase doesn't appear to have worked; it looked correct locally but now I look at the remote repos it looks like it just reset the branch to my first commit.   Apologies.

PR #3952, opened 2023-08-18
Fix building against upstream libdivecomputer and the v0.9.0 api changes

jefdriesen:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

With these changes, subsurface can be build against:
- Current subsurface fork of libdivecomputer
- Upstream libdivecomputer v0.8.0
- Upstream libdivecomputer v0.9.0

I use these changes for testing libdivecomputer fixes, but they might be useful for other developers too. When merging the v0.9.0 changes into the subsurface fork, some more changes might be necessary. Feel free to modify the changes in this PR as necessary.

### DIFF
--- a/core/configuredivecomputerthreads.cpp
+++ b/core/configuredivecomputerthreads.cpp
@@ -2218,7 +2218,11 @@ void FirmwareUpdateThread::run()
 	}
 	switch (dc_device_get_type(m_data->device)) {
 	case DC_FAMILY_HW_OSTC3:
+#ifdef DC_FIELD_STRING
 		rc = hw_ostc3_device_fwupdate(m_data->device, qPrintable(m_fileName), m_forceUpdate);
+#else
+		rc = hw_ostc3_device_fwupdate(m_data->device, qPrintable(m_fileName));
+#endif
 		break;
 	case DC_FAMILY_HW_OSTC:
 		rc = hw_ostc_device_fwupdate(m_data->device, qPrintable(m_fileName));
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -167,13 +167,19 @@ static int parse_gasmixes(device_data_t *devdata, struct dive *dive, dc_parser_t
 				cyl.type.workingpressure.mbar = lrint(tank.workpressure * 1000);
 
 				cyl.cylinder_use = OC_GAS;
+#ifdef DC_TANKINFO_CC_O2
 				// libdivecomputer treats these as independent, but a tank cannot be used for diluent and O2 at the same time
 				if (tank.type & DC_TANKINFO_CC_DILUENT)
 					cyl.cylinder_use = DILUENT;
 				else if (tank.type & DC_TANKINFO_CC_O2)
 					cyl.cylinder_use = OXYGEN;
+#endif
 
+#ifdef DC_TANKINFO_IMPERIAL
 				if (tank.type & DC_TANKINFO_IMPERIAL) {
+#else
+				if (tank.type == DC_TANKVOLUME_IMPERIAL) {
+#endif
 					if (same_string(devdata->model, "Suunto EON Steel")) {
 						/* Suunto EON Steele gets this wrong. Badly.
 						 * but on the plus side it only supports a few imperial sizes,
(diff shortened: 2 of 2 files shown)


core/configuredivecomputerthreads.cpp
```diff
@@ -2218,7 +2218,11 @@ void FirmwareUpdateThread::run()
 	}
 	switch (dc_device_get_type(m_data->device)) {
 	case DC_FAMILY_HW_OSTC3:
+#ifdef DC_FIELD_STRING
```
mikeller, line 2221: What does this define do, and how does this relate to the support for forced updates on OSTC4?

core/libdivecomputer.c
```diff
@@ -167,13 +162,26 @@ static int parse_gasmixes(device_data_t *devdata, struct dive *dive, dc_parser_t
 				cyl.type.workingpressure.mbar = lrint(tank.workpressure * 1000);
 
 				cyl.cylinder_use = OC_GAS;
+#if DC_VERSION_CHECK(0,9,0)
```
mikeller, line 165: These version checks add a lot of complexity to the code. Can't we just build against a fixed version (which is what we are doing anyway), and avoid the added complexity?

core/configuredivecomputerthreads.cpp
jefdriesen: It's not related to the OSTC4 firmware upgrade at all. It's just a random macro that happens to be present in the subsurface fork of libdivecomputer, but not in the upstream version, and which is used here to distinguish between the two.

core/libdivecomputer.c
jefdriesen: The api changes in libdivecomputer v0.9.0 are not backwards compatible. So without the version checks you can either support the v0.8.0 or the v0.9.0 api, but not both. The subsurface fork is currently at v0.8.0, so if someone already wants to try some of the v0.9.0 changes now, the build will fail without these patches. Of course, once the v0.9.0 api gets merged in the subsurface fork, the version checks won't be needed anymore, because at that point the v0.8.0 compatibility can be dropped.

Anyway, even if this PR is not accepted as-is, the code can still serve as an example for what is needed to move to v0.9.0 api. I already had the code, so why not share it?

core/configuredivecomputerthreads.cpp
glance-: I looked at this and it would have been good if there where a macro either identifying the subsurface fork or the force update feature but until then I'd say this is enough.

Maybe add a comment about it to clear any future confusion.

core/libdivecomputer.c
glance-: I think they're small enough for now, and it makes it easy to yank them out in the future when the 0.8.0 api is irrelevant.

glance-:
Maybe add a comment about using DC_FIELD_STRING to identify the subsurface fork for firmware forced update, but otherwise it LGTM.

dirkhh:
To be honest, I see no value in this. 
It makes a mess of our code, it removes features for the users.

We have been trying for years to get you to merge the string values and other improvements in our branch. I think that's the much more reasonable approach.

NAK

core/libdivecomputer.c
dirkhh: No. We will not add these version checks back.
Subsurface very intentionally builds against the bundled version of libdc in order to avoid this mess.

mikeller: > The subsurface fork is currently at v0.8.0, so if someone already wants to try some of the v0.9.0 changes now, the build will fail without these patches.

I think the changes in Subsurface that are needed to actually make use of the API / data format changes in libdivecomputer 0.9 will have to go deeper than this - right down to a change of the data model, to treat and store gasmixes and tank sensor information as separate groups of entities, and the addition of a UI to allow the user to define and maintain mappings between the two that match their gear configuration(s).

I can probably find some time to start on this, and once this is ready this can be merged together with an update of libdivecomputer to 0.9. But in my opinion this should be after the next release of Subsurface has been completed.

As a side note, in [SemVer](https://semver.org/#summary), breaking API changes are typically denoted in a _major_ version number change - following this pattern will make it easier for users of the library to understand what is going on.

PR #3954, opened 2023-09-04
[HELP] media: load metadata and thumbnails of raw pictures using libraw

bstoeger:
The distinguished photographer shoots raw images. There is a comprehensive library that can extract metadata and thumbnails from these images. Let's use it if available.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Attempt to extract thumbnails and metadata from raw images.

Unfortunately I was not able to find out how to test for the existence of the libraw library.

Moreover, I couldn't test two aspects:
- Thumbnails extracted as BITMAP (the files I have access to store their thumbnail as JPEGs).
- GPS metadata.

### Release note:

Done.

### Documentation change:

Not sure, if necessary...?

### Mentions:

@dirkhh: please help with `cmake` check for the existence of the library.


mikeller:
This looks interesting, good idea!

Re checking for the presence of the library: For my understanding of the build and packaging process if this should be done it should be on the system that subsurface is deployed to, instead of on the build system. On the build system we can simply make this a prerequisite for building, and potentially a config option to force building without it.

Also, making the assumption that most of our users are on Windows, I think the recommended way to use libraw (on Windows) will be to package the .dll with Subsurface (from https://www.libraw.org/docs/Install-LibRaw-eng.html):
> No installation under Windows is supported. It is assumed that all DLLs will be supplied together with the software using them (and this software will perform the installation). Accordingly, in building of programs using LibRaw, the paths to libraries, DLLs, and include-files should be specified manually.

I suspect that this will also be the simplest way to deal with this dependency on MacOS - not sure if there is a reasonable way to require packages from distribution systems like homebrew as part of the installation of a .dmg application image in MacOS. For linux the preferrable way will probably be to declare a dependency on libraw in the package formats we are supporting.

bstoeger:
I think I found out how to test for the existence of the library. I guess this should now work for people compiling on Linux with the libraw-dev package installed.

Now getting it into the various build systems, that is waaaay above my head...

dirkhh (commented):
Looks interesting.
Of course now someone needs to get this included in the Windows and macOS build.
And my motivation for working on Subsurface these days hovers around -100

bstoeger:
Sorry, I cannot do that. I tried adding a build-dependency for the Ubuntu/Debian package, but I have no idea if this helps.

In any case, before this becomes relevant I need to test this with raw pictures that have GPS data and that export thumbnails as bitmap instead of JPEG.

bstoeger:
Jeez - I can't find any example Kodak DCS200 (.KC2), Rollei or SINAR (.ia) files on the web (apparently these are the ones were we get the thumbnails as bitmap instead of JPEG). For some reason all the classical search for files by type search engines don't work anymore.

mikeller:
@bstoeger: Did you manage to find a RAW image with embedded GPS? I could probably ask a friend to provide one. I can also reach out to Sinar to see if I can get a .IA file.

bstoeger:
That would be great - I am unable to find RAW test images on the net. There are instructions how to set the GPS coordinates, but I think an original file would be preferred.

mikeller:
@bstoeger: I sent you a RAW test image with GPS coordinates, taken with a Sony A77. No success with the Sinar file (it is only their legacy format), but seeing that all of the cameras that ever supported bitmap thumbnails were either middle / large format or the size of a shoe box (and required an external power source), I think we can just leave this untested and trust on the fact that it's never going to be used.

bstoeger:
We didn't find raw images with GPS data, so I guess we have to proceed without testing this particular case. In the unlikely event that anybody ever has such a use case they will complain.

I'm un-drafting this. Last time I tried, it worked for me on Linux, but I don't know how to integrate it in the build system on either Linux, much less Windows or Mac.

mikeller:
> I'm un-drafting this. Last time I tried, it worked for me on Linux, but I don't know how to integrate it in the build system on either Linux, much less Windows or Mac.

The CICD runs seem to be fine on this so this should be working I think.

bstoeger:
Rebased on master.

Even if this isn't on Mac & Windows, I think it doesn't hurt. I have a number of RAW images and really would like to include them.

mikeller:
> Even if this isn't on Mac & Windows, I think it doesn't hurt. I have a number of RAW images and really would like to include them.

Agreed - feature parity is a nice to have, not a need to have.
And let's face it, if you're not on linux you're always going to be subar. 😉

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ desktop: fix gas switches in UDDF exports
 core: allow of up to 6 O2 sensors (and corresponding voting logic)
 desktop: add divemode as a possible dive list column
 profile-widget: Now zomed in profiles can be panned with horizontal scroll.
+media: support raw files if libraw is installed
```
mikeller, line 11: I think with the move to publishing CICD releases this file is getting to the end of its usefulness...

mikeller (approved):
Looks good to me - we can always have another stab at adding this for Windows / MacOS as a next step.

mikeller (changes requested):
Actually, thinking about this some more, I think we need to declare a dependency on `libraw` in https://github.com/subsurface/subsurface/pull/3954/files#diff-ca1085ec4b075c04ab5e4d1a8f15a4f16c1ac2483f5a19d2d26afc1cc567d903R50 - we are currently building the distribution package in a way that expects libraw to be installed on the user's system, so if it is missing this will probably result in an application crash when they are trying to add RAW images.

mikeller:
We should probably add similar dev / runtime requirements pairs to other distributions like copr, snap, and AppImage. @dirkhh , @Saviq do you have bandwidth / motivation to look into this once this has been merged?

@notrege: Do you have an idea how we could add an external library to the distribution on MacOS?

dirkhh:
Yeah, I think I know how to do this for all of them, actually

Saviq:
- #4314 should be good for the snap

dirkhh:
I rudely pushed one more commit into your branch @bstoeger 
I had forgotten that the test builds here use their own dependency setup.
Snap and macOS already used libraw in the previous round.
Windows support is most obviously lacking as that needs to be added to the container in order for the build to find it.

dirkhh:
Oh  my...
```
/__w/subsurface/subsurface/core/metadata.cpp:555:74: error: 'struct libraw_gps_info_t' has no member named 'longitude'; did you mean 'longtitude'?
2024-09-10T20:33:16.8012285Z   555 |  metadata->location.lon = degminsec_to_udeg(raw.imgdata.other.parsed_gps.longitude);
2024-09-10T20:33:16.8013255Z       |                                                                          ^~~~~~~~~
2024-09-10T20:33:16.8016524Z       |                                                                          longtitude
2024-09-10T20:33:17.0258252Z make[2]: *** [core/CMakeFiles/subsurface_corelib.dir/build.make:700: core/CMakeFiles/subsurface_corelib.dir/metadata.cpp.o] Error 1
```
Seriously, `longtitude` ???

dirkhh:
> Oh my...
> 
> ```
> /__w/subsurface/subsurface/core/metadata.cpp:555:74: error: 'struct libraw_gps_info_t' has no member named 'longitude'; did you mean 'longtitude'?
> 2024-09-10T20:33:16.8012285Z   555 |  metadata->location.lon = degminsec_to_udeg(raw.imgdata.other.parsed_gps.longitude);
> 2024-09-10T20:33:16.8013255Z       |                                                                          ^~~~~~~~~
> 2024-09-10T20:33:16.8016524Z       |                                                                          longtitude
> 2024-09-10T20:33:17.0258252Z make[2]: *** [core/CMakeFiles/subsurface_corelib.dir/build.make:700: core/CMakeFiles/subsurface_corelib.dir/metadata.cpp.o] Error 1
> ```
> 
> Seriously, `longtitude` ???

https://github.com/LibRaw/LibRaw/commit/8a9d1b6f48fd89ffbdcc32c274dacea52e6fba8f

So we need at least version 0.20 - which is 4 years old... shouldn't be an issue on current systems, but will break on older distros.

dirkhh:
the other issue is that `smtk2ssrf` hasn't had an update to its `CMakeLists.txt` in order to pull the same dependencies.

dirkhh:
ok, since I am being rude anyway, I figured I might as well continue. Let's see if this builds on all platforms, now

Again, this is **still** missing Windows.

smtk-import/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ pkg_config_library(LIBGIT2 libgit2 REQUIRED)
 pkg_config_library(LIBMDB libmdb REQUIRED)
 pkg_config_library(LIBDC libdivecomputer REQUIRED)
 pkg_config_library(LIBFTDI libftdi1 QUIET)
+pkg_config_library(LIBRAW libraw)
```
mikeller, line 38: Is this actually needed here - not sure if smtk-import is handling images anywhere?

packaging/copr/subsurface.spec
```diff
@@ -35,6 +35,7 @@ BuildRequires:  libssh2-devel
 BuildRequires:  libcurl-devel
 BuildRequires:  libgit2-devel
 BuildRequires:  libmtp-devel
+BuildRequires:  LibRaw-devel
```
mikeller, line 38: Does this need to declare an installation dependency on libraw?

packaging/OBS/subsurfacedaily.spec
```diff
@@ -38,6 +38,7 @@ BuildRequires:	libssh2-devel
 BuildRequires:	libcurl-devel
 BuildRequires:  libgit2-devel
 BuildRequires:  libmtp-devel
+BuildRequires:  libraw-devel
```
mikeller, line 41: Does this need to declare an installation dependency on libraw?

packaging/ubuntu/debian/control
```diff
@@ -20,6 +21,7 @@ Build-Depends: asciidoc,
                libusb-1.0-0-dev,
                libbluetooth-dev,
                libmtp-dev,
+               libraw-dev,
```
mikeller, line 23: I think we've doubled up here. 😆

bstoeger:
I think the DCO bot is confused...?

smtk-import/CMakeLists.txt
dirkhh: the link step fails without it. So a component that is linked into smtk2ssrf requires that library.
We can argue that that shouldn't be the case - but my solution above at least solves the failing builds.

packaging/copr/subsurface.spec
dirkhh: IIRC the way this works is that this then automagically creates the runtime dependency.
You see that none of the other libraries like `libgit2` or `libxml2` have runtime dependencies, either

packaging/ubuntu/debian/control
dirkhh: I didn't notice.
And you are adding the runtime dependency I notice, which again, I believe, isn't needed.
Obviously we should drop one of them (although the tools are smart enough not to care)

dirkhh:
> I think the DCO bot is confused...?

I've seen this with merges in a PR in the past. We can always manually merge it.

mikeller:
Once #4321 has been merged, and the docker image version in `.github/workflows/windows.yml` has been updated the Windows build will be ready to include libraw.

dirkhh:
I think I found one online: https://github.com/ianare/exif-samples/blob/master/heic/mobile/HMD_Nokia_8.3_5G.heif

mikeller:
> I think I found one online: [ianare/exif-samples@`master`/heic/mobile/HMD_Nokia_8.3_5G.heif](https://github.com/ianare/exif-samples/blob/master/heic/mobile/HMD_Nokia_8.3_5G.heif?rgh-link-date=2024-09-13T13%3A50%3A33Z)

I don't think HEIF falls into the 'raw' image category - it is rather the opposite, as it is highly processed and compressed.

smtk-import/CMakeLists.txt
mikeller: @dirkhh: I figured out the 'why' for this - our CICD builds for smtk2ssrf are re-using the compiler output from the subsurface build that they do beforehand - and this build has got LIBRAW configured.

Do you know what the purpose is for this re-use?

dirkhh: I don't. I only enabled those extra builds at some point when we ended up breaking the smtk2ssrf build without noticing...

mikeller: I found something in the way of an explanation:

https://github.com/subsurface/subsurface/blob/a6a15f9d3a75262220c1d3e6f8f8e20759b9061a/packaging/windows/smtk2ssrf-mxe-build.sh#L8-L12

I am not sure if this re-use of the library that was built for Subsurface makes sense in the way we are delivering the apps at the moment:
- we are dragging the dependencies that Subsurface requires into smtk2ssrf, if they are needed or not;
- we seem to be only delivering smtk2ssrf with its own installer, and its own packaged `libsubsurface_corelib`.

When I looked around, I could only find the standalone Windows installer as a user facing distribution format for smtk2ssrf - do you know if smtk2ssrf is included in any of the packages we are building for linux distributions?

mikeller (approved):
No point holding this up over the `smtk2ssrf` build issue - but I am still keen to get this build sorted, or `smtk2ssrf` removed from the builds if nobody is using this any more @dirkhh.

By now we also have builds that include libraw in most places, including Windows.

bstoeger:
> No point holding this up over the `smtk2ssrf` build issue - but I am still keen to get this build sorted, or `smtk2ssrf` removed from the builds if nobody is using this any more @dirkhh.

I think at least @salvadorcunat still uses and actively maintains this.

atdotde:
I run this as a web service on https://thetheoreticaldiver.org/rch-cgi-bin/smtk2ssrf.pl which get accessed about once a day on average.

salvadorcunat:
As far as future changes don't break compatibility with current XML logs (version 3, I think), whichever current build of smtk2ssrf should work. Thus @atdotde can keep a recent build running on his page while it is dropped from CI/CD. Windows versions are built statically so current binaries will run in the future.
The only drawback can come from future improvements in libdc for the Scubapro/Uwatec families supported by SmatTrak, but I don't think there is a lot of improvement to do here.

I'm assuming libraw can't be a optionally supported dependency, just as we are currently doing with BT or FTDI support. Smtk2ssrf doesn't need it so setting off the support in the build script would do the job.

smtk-import/CMakeLists.txt
salvadorcunat: This was wrote thinking in a fresh build from source in a local machine, as build.sh ensures everything is in place. Then smtk2ssrf-build.sh rebuilds libsubsurface_corelib.a dropping some unused dependencies like USB or BT.  If libraw is not built into corelib it should be possible to drop it too from smtk2ssrf builds.

salvadorcunat: > When I looked around, I could only find the standalone Windows installer as a user facing distribution format for smtk2ssrf - do you know if smtk2ssrf is included in any of the packages we are building for linux distributions?

No AFAIK. It's an standalone app and is not supposed to be shipped with Subsurface. BTW is possible to build an CLI version which is used by @atdotde in his web page.

mikeller:
@atdotde:
> I run this as a web service on [thetheoreticaldiver.org/rch-cgi-bin/smtk2ssrf.pl](https://thetheoreticaldiver.org/rch-cgi-bin/smtk2ssrf.pl) which get accessed about once a day on average.

Does 'accessed' in the above mean that somebody actually converts a file with it, or is this just an access to the website?


@salvadorcunat:

> As far as future changes don't break compatibility with current XML logs (version 3, I think), whichever current build of smtk2ssrf should work. Thus @atdotde can keep a recent build running on his page while it is dropped from CI/CD. Windows versions are built statically so current binaries will run in the future. The only drawback can come from future improvements in libdc for the Scubapro/Uwatec families supported by SmatTrak, but I don't think there is a lot of improvement to do here.

The Windows build of `smtk2ssrf` is actually also 100% dependent on the Subsurface build:

https://github.com/subsurface/subsurface/blob/716b350af2012c981db5ba9e682d7bf8e655a450/packaging/windows/smtk2ssrf-mxe-build.sh#L8-L12

> I'm assuming libraw can't be a optionally supported dependency, just as we are currently doing with BT or FTDI support. Smtk2ssrf doesn't need it so setting off the support in the build script would do the job.

Not sure what you mean by 'optionally supported dependency'?
From what I can tell, because of how our `smtk2ssrf` build scripts are done, there is no optionality in these libraries, `smtk2ssrf` can not be built without having these in place - even though it does not seem to be getting any value from any of them.

salvadorcunat:
> The Windows build of `smtk2ssrf` is actually also 100% dependent on the Subsurface build:
>
Sure, it is. I meant it doesn't need CI/CD, a build from current master will work in two years too.  So smtk2ssrf can be removed from CI/CD builds.
 

> Not sure what you mean by 'optionally supported dependency'? From what I can tell, because of how our `smtk2ssrf` build scripts are done, there is no optionality in these libraries, `smtk2ssrf` can not be built without having these in place - even though it does not seem to be getting any value from any of them.

Thinking on CMake here. BT and USB are just examples of features that you can turn on/off Subsurface build by setting it in CMakeFile. Smtk2ssrf build script turns these off before rebuilding subsurface.

A third option would be reworking smtk2ssrf without corelib dependencies, sure.

mikeller:
> Thinking on CMake here. BT and USB are just examples of features that you can turn on/off Subsurface build by setting it in CMakeFile. Smtk2ssrf build script turns these off before rebuilding subsurface.

This sounds like an interesting option - do you have a pointer for where this is done? All the CICD scripts I looked at re-use the artifacts built by the 'normal' Subsurface build with all the options and unwanted libraries enabled.
If we can replicate this for libraw then we will be fine.

> A third option would be reworking smtk2ssrf without corelib dependencies, sure.

I don't think it needs to be reworked to remove the corelib dependencies - it will be enough if we move to a build of corelib that only builds the parts needed for `smtk2ssrf`, like what you describe above. I am just not versed enough in cmake to figure out how this can be done.

atdotde:
> Does 'accessed' in the above mean that somebody actually converts a file with it, or is this just an access to the website?

There are some bots but there are also actual conversions.

PR #3957, opened 2023-09-10
Properly handle the planner state

### LINKED ISSUE #3955: Cancelling planner and restarting
Start Subsurface, enter the planner (you see the default dive), press cancel (so you return to the log), start the planner again: The profile is in a weird state with no profile line but the heatmap shows that something is there. Clicking in the profile reproduciably leads to a crash.

### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Operating system:

Tested here on MacOS on ARM Mac.

### Subsurface version:

5.0.10

atdotde:
This does two independent things:

It sets the planner state early enough so the appropriate default profile for the planner is created (without safety stop).

Upon cancelling the planner, it resets the profile widget to profile more (rather than planner) as otherwise upon the next change into the planner the planner model is not properly initialized.

Fixes #3955 

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@bstoeger could you please check this is done correctly? I am not 100% sure I do these mode changes in the appropriate places. But at least, this fixes the crash ;-)

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -581,7 +581,7 @@ void MainWindow::enableShortcuts()
 void MainWindow::showProfile()
 {
 	enableShortcuts();
-	profile->plotCurrentDive();
+	profile->plotDive(current_dive, profile->dc);
 	setApplicationState(ApplicationState::Default);
 }
 
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -568,6 +568,7 @@ divemode_t PlannerWidgets::getRebreatherMode() const
 
 void PlannerWidgets::preparePlanDive(const dive *currentDive)
 {
+	DivePlannerPointsModel::instance()->setPlanMode(DivePlannerPointsModel::PLAN);
 	// create a simple starting dive, using the first gas from the just copied cylinders
 	DivePlannerPointsModel::instance()->createSimpleDive(planned_dive.get());
 
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -393,6 +393,7 @@ static void hideAll(const T &container)
 
 void ProfileWidget2::clear()
 {
+	currentState = INIT;
 #ifndef SUBSURFACE_MOBILE
 	clearPictures();
 #endif


mikeller (approved):
Looks good to me.

dirkhh (commented):
Not code I'm sufficiently familiar with that I should say "looks great".
I can't find anything that looks wrong, but that doesn't say much.

bstoeger:
The second change feels a bit sketchy to me. Why redraw the planned dive, when we actually exit plan-mode? I'll have a look.

atdotde:
I am not sure what you are talking about @bstoeger: This commit does not change anything regarding redrawing, it only sets  ProfileWidget2's state to PROFILE rather than PLANNER what it was before.

bstoeger:
Yes, but with a dive that makes no sense anymore.

I would prefer something like this, but I am not sure about the ramifications:

```
diff --git a/desktop-widgets/mainwindow.cpp b/desktop-widgets/mainwindow.cpp
index 19f3e46f2..9c03f2e5b 100644
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -581,7 +581,7 @@ void MainWindow::enableShortcuts()
 void MainWindow::showProfile()
 {
 	enableShortcuts();
-	profile->plotCurrentDive();
+	profile->plotDive(current_dive, profile->dc);
 	setApplicationState(ApplicationState::Default);
 }
 
diff --git a/profile-widget/profilewidget2.cpp b/profile-widget/profilewidget2.cpp
index d362c462d..22ac062ca 100644
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -393,6 +393,7 @@ static void hideAll(const T &container)
 
 void ProfileWidget2::clear()
 {
+	currentState = INIT;
 #ifndef SUBSURFACE_MOBILE
 	clearPictures();
 #endif
```

That feels right to me, because the empty state is not the profile state, after all. But I am not 100% sure.

desktop-widgets/mainwindow.cpp
```diff
@@ -581,7 +581,8 @@ void MainWindow::enableShortcuts()
 void MainWindow::showProfile()
 {
 	enableShortcuts();
-	profile->plotCurrentDive();
+//	profile->plotCurrentDive();
```
bstoeger: Any reason to keep this as a comment? I think this is wrong, as it will plot the currently shown dive. However, that is the planned dive when exiting the planner and thus is, in a sense, not valid anymore.

desktop-widgets/mainwindow.cpp
```diff
@@ -636,6 +637,7 @@ bool MainWindow::plannerStateClean()
 
 void MainWindow::planCanceled()
 {
+	profile->view->setProfileState(plannerWidgets->getDive(), profile->dc);
```
bstoeger: Is that still necessary? The `showProfile()` call in the next line will either show the empty profile or the current dive, and thus reset the profile's status.

bstoeger (approved):
lgtm

PR #3965, opened 2023-09-26
remove unsupported versions and add newer ones

dirkhh:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation


packaging/ubuntu/make-package.sh
```diff
@@ -116,8 +116,8 @@ debuild -S -d
 # and as of hirsute this package no longer exists. So simply remove it from the control
 # file when building the newer ones
 sed -i.bak "/qt5-default/d" debian/control
-rel=bionic
-others="focal jammy kinetic"
+rel=focal
```
mikeller, line 113: Good idea to update this - is it possible to get rid of the bionic-specific code above as well?

dirkhh: UGH... I hate it when I'm being stupid. Thanks for catching this, pushed another commit that should address this. The annoying thing is that the change will only be truly exercised once it is merged into master (due to the way that the "forwarding to Launchpad" works).

mikeller (approved):

PR #3967, opened 2023-10-04
Fix certificate_check_cb() return value

rfuchs:
libgit documents the return value of callbacks.certificate_check[0] as 0 for success and <0 as failure. Returning 1 for success (kind of) works because some parts of libgit[1] check for <0 return values and only treat those as errors, but the function actually calling the callback (check_certificate) treats anything non-zero as error[2] and ends up setting a spurious error message with a return value of 1.

[0] https://libgit2.org/libgit2/#HEAD/group/callback/git_transport_certificate_check_cb
[1] https://github.com/libgit2/libgit2/blob/maint/v1.5/src/libgit2/transports/httpclient.c#L1054
[2] https://github.com/libgit2/libgit2/blob/maint/v1.5/src/libgit2/transports/httpclient.c#L785

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -313,9 +313,9 @@ int certificate_check_cb(git_cert *cert, int valid, const char *host, void *payl
 		// if we are connecting to the cloud server we alrady called 'canReachCloudServer()'
 		// which will fail if the SSL certificate isn't valid, so let's simply always
 		// tell the caller that this certificate is valid
-		return 1;
+		return 0;
 	}
-	return valid;
+	return valid ? 0 : -1;
 }
 
 static int update_remote(struct git_info *info, git_remote *origin, git_reference *local, git_reference *remote)


mikeller (commented):
Hi Richard.

The change looks good, but can you please add a DCO / signed-off-by line to your commit message?

More information here: https://github.com/subsurface/subsurface/pull/3967/checks?check_run_id=17408105475

rfuchs:
> Hi Richard.
> 
> The change looks good, but can you please add a DCO / signed-off-by line to your commit message?
> 
> More information here: https://github.com/subsurface/subsurface/pull/3967/checks?check_run_id=17408105475

Should be better now. (I had it included already but the author was set to the wrong account, whoopsie)

mikeller (approved):

PR #3972, opened 2023-10-11
Use horizontal mouse wheel for zoomed panning

glance-:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Since a long time you have had the possibility to do vertical scroll on the profile to zome. This adds the possibility to do horizontal scroll in this zoomed in mode to pan the profile.

### Changes made:

Added a case in the wheelEvent listener in the profile-widget for horizontal scroll.

### Release note:

Now zomed in profiles can be panned with horizontal scroll.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 desktop: add divemode as a possible dive list column
+profile-widget: Now zomed in profiles can be panned with horizontal scroll.
 desktop: hide only events with the same severity when 'Hide similar events' is used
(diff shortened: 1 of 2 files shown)


mikeller (approved):
Seems to be a good idea.
I don't have a way of testing it because I don't have a pointing device that can do horizontal scrolling.

PR #3974, opened 2023-10-11
Get dive gas fixes

glance-:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a couple of corner cases in the get_dive_gas function, like always having a ..100% on CCR dives, including gases flagged as not used and not always finding the min-o2 gas used.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -47,6 +47,8 @@ void get_dive_gas(const struct dive *dive, int *o2_p, int *he_p, int *o2max_p)
 			continue;
 		if (o2 > maxo2)
 			maxo2 = o2;
+		if (o2 < mino2)
+			mino2 = o2;
 		if (he > maxhe)
 			goto newmax;
 		if (he < maxhe)
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -43,6 +43,8 @@ void get_dive_gas(const struct dive *dive, int *o2_p, int *he_p, int *o2max_p)
 			continue;
 		if (cyl->cylinder_use == OXYGEN)
 			continue;
+		if (cyl->cylinder_use == NOT_USED)
+			continue;
 		if (o2 > maxo2)
 			maxo2 = o2;
 		if (he > maxhe)
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -47,7 +47,7 @@ void get_dive_gas(const struct dive *dive, int *o2_p, int *he_p, int *o2max_p)
 			continue;
 		if (o2 > maxo2)
 			maxo2 = o2;
-		if (o2 < mino2)
+		if (o2 < mino2 && maxhe <= 0)
 			mino2 = o2;
 		if (he > maxhe)
 			goto newmax;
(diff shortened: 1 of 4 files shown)


mikeller (commented):
Excluding not used gases and oxygen seems to make sense.

But I am not 100% sure if the handling of `mino2` is correct (but it seems to be also incorrect in the current version according to what the comment on the function says): To my understanding `mino2` is used to track 'the oxygen content of the highest helium trimix' (and has a misleading name), so it should probably only be adjusted for gases that contain helium. But then it should also be updated for any subsequent 'equally high helium' gases in the list if they have higher (or lower, the comment is inconclusive) oxygen.

Independent of this, the `goto` and label in the existing function seem to be unnecessary, as `if (o2 <= maxo2) continue` will always be taken because of the earlier `if (o2 > maxo2) maxo2 = o2`, so the code after the label can just be pulled up into the `if (he > maxhe)`.

glance-:
mino2 isn't just for helium content, and no, I don't see a simpler way to rewrite those statements without the goto.

mino2 is used for tracking the gas with the lowest o2 , even if it's higher than air. Example is when doing a dive on 32% nitrox and deco on 100%, the gas string should be 32..100% to show you used gases with 32% as the lowest and 100% as the highest.

I added even more test cases showing this.

If you do find any way to rewrite it in any cleaner way, feel free to.

mikeller:
Ah ok in this case the comment on the function (and the function name to some extent) is misleading - this isn't returning any gas mixes that are configured for the dive, but rather the minimum / maximum values for O2 and the maximum value for He out of all gases configured.

glance-:
The comment says nothing about gasmix just "Get "maximal" dive gas for a dive" which is used in formatDiveGasString for the gas column in the divelist.

mikeller:
Not sure what you understand as 'dive gas for a dive', but to my understanding this seems to imply that this is an actual gas that was carried and configured for the dive.

To give you an example try running a test for a CCR dive with TMx 22/20 DIL and air as bailout:

```
void TestformatDiveGasString::test_ccr_bailout() {
        struct dive dive = {0};
        cylinder_t *cylinder = get_or_create_cylinder(&dive, 0);

        cylinder->gasmix.o2.permille = 1000;
        cylinder->cylinder_use = OXYGEN;
        cylinder->start.mbar = 230000;
        cylinder->end.mbar = 100000;

        cylinder = get_or_create_cylinder(&dive, 1);

        cylinder->gasmix.o2.permille = 220;
        cylinder->gasmix.he.permille = 200;
        cylinder->cylinder_use = DILUENT;
        cylinder->start.mbar = 230000;    
        cylinder->end.mbar = 100000;
        
        cylinder = get_or_create_cylinder(&dive, 2);
        
        cylinder->gasmix.o2.permille = 210;
        cylinder->gasmix.he.permille = 0;
        cylinder->start.mbar = 230000;
        cylinder->end.mbar = 100000;
        
        QCOMPARE(formatDiveGasString(&dive), "22/20");
}
```

glance-:
Yes, it's a description of the actual gases carried during the dive.

Yes, it looked like you found another case where the previous code had a bug. I fixed that and included the test case.

mikeller (approved):
This looks good now, thanks for fixing.

PR #3982, opened 2023-11-03
build-system: use gcc 10 for MXE builds

atdotde:
It's unclear if this will be enough to use gcc 10 by default when building Subsurface using this container.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/scripts/docker/mxe-build-container/settings.mk
+++ b/scripts/docker/mxe-build-container/settings.mk
@@ -8,6 +8,8 @@ JOBS := 8
 # This variable controls the targets that will build.
 MXE_TARGETS :=  x86_64-w64-mingw32.shared i686-w64-mingw32.shared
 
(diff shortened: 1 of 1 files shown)


mikeller (approved):
According to https://github.com/subsurface/subsurface/commit/551f10591ab1c78e44d9d6f5a37aa7e64acdfab7#commitcomment-131633380 this is sufficient to make the build in this container work with gcc 10.

mikeller:
Still trying to build the docker image of the rebased version of this branch, not having much luck so far...

mikeller:
Turns out this needs a fairly grunty machine, or it will run into errors. I managed to build it on a Xeon based machine in 5 hours. So the build does work, but I doubt we'll be able to get this built on GitHub infrastructure.

PR #3983, opened 2023-11-03
Add BLE detection for the Aqualung i330R and Apeks DSX

jefdriesen:
Add the Aqualung i330R and Apeks DSX model numbers to the Pelagic pattern table. These two models also use a new BLE service UUID.

Note that this only adds the BLE detection. Actual support for the devices in libdivecomputer is still being worked on.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


mikeller (approved):
Looks good to me.

dirkhh (approved):
seems reasonable

PR #3984, opened 2023-11-03
Use git rev-parse to get the SHA1

jefdriesen:
The 'git describe' command finds the most recent tag instead of the SHA1. When trying to build from a repository without any tags, for example in a fork, this causes the build to fail.

Use the 'git rev-parse' commands instead, which also returns the full SHA1.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/packaging/windows/mxe-based-build.sh
+++ b/packaging/windows/mxe-based-build.sh
@@ -178,7 +178,7 @@ if [ "$MXEBUILDTYPE" = "x86_64-w64-mingw32.shared" ] ; then
 fi
 
 cd "$BUILDDIR"
-CURRENT_SHA=$(cd "$BASEDIR"/subsurface/libdivecomputer ; git describe)
+CURRENT_SHA=$(cd "$BASEDIR"/subsurface/libdivecomputer ; git describe --always --long)
 PREVIOUS_SHA=$(cat "libdivecomputer.SHA" 2>/dev/null || echo)
 if [ ! "$CURRENT_SHA" = "$PREVIOUS_SHA" ] || [ ! -d libdivecomputer ] || [ -f build.libdivecomputer ] ; then
 	rm -f build.libdivecomputer
--- a/packaging/windows/mxe-based-build.sh
+++ b/packaging/windows/mxe-based-build.sh
@@ -178,9 +178,9 @@ if [ "$MXEBUILDTYPE" = "x86_64-w64-mingw32.shared" ] ; then
 fi
 
 cd "$BUILDDIR"
-CURRENT_SHA=$(cd "$BASEDIR"/subsurface/libdivecomputer ; git describe --always --long)
+CURRENT_SHA=$(cd "$BASEDIR"/subsurface/libdivecomputer ; git describe --always --long --dirty)
 PREVIOUS_SHA=$(cat "libdivecomputer.SHA" 2>/dev/null || echo)
-if [ ! "$CURRENT_SHA" = "$PREVIOUS_SHA" ] || [ ! -d libdivecomputer ] || [ -f build.libdivecomputer ] ; then
+if [ ! "$CURRENT_SHA" = "$PREVIOUS_SHA" ] || [[ "$CURRENT_SHA" == *-dirty ]] || [ ! -d libdivecomputer ] || [ -f build.libdivecomputer ] ; then
 	rm -f build.libdivecomputer
 	mkdir -p libdivecomputer
 	cd libdivecomputer


packaging/windows/mxe-based-build.sh
```diff
@@ -178,7 +178,7 @@ if [ "$MXEBUILDTYPE" = "x86_64-w64-mingw32.shared" ] ; then
 fi
 
 cd "$BUILDDIR"
-CURRENT_SHA=$(cd "$BASEDIR"/subsurface/libdivecomputer ; git describe)
+CURRENT_SHA=$(cd "$BASEDIR"/subsurface/libdivecomputer ; git rev-parse --verify HEAD)
```
mikeller: Not sure this is the right fix - it will always return the changeset hash for the HEAD, even if there are uncommitted local changes.
I think a better option will be to use `git describe --always --long --dirty`, and then also check if `[[ $CURRENT_SHA == *-dirty ]]`. The `--always --long` will take care of the case where no tags exist and make sure that an abbreviated changeset hash is always added.

jefdriesen: Using the `--always` option should work too. If we want to support uncommitted local changes, then we'll indeed need some extra checks too. Not sure if that is really needed here, because I assume this script is mainly used for building in CI? On the other hand, it might be useful for people running the scripts manually.

mikeller: In my opinion components of the CI build should work when run locally as well - otherwise it's just gonna be hard to maintain them.
We already have some pitfalls in the android build that require changes to libdivecomputer to be in a changeset that is being pointed to by the main repository, and this makes testing these changes on mobile even more tedious than mobile testing already is...

jefdriesen: I have updated the PR with your suggestion.

However, I suspect this SHA1 testing has never worked because the `libdivecomputer.SHA` file is never written anywhere. Thus the `PREVIOUS_SHA` variable is always empty, and the result is that libdivecomputer will always get rebuild, even without these changes.

mikeller (approved):

packaging/windows/mxe-based-build.sh
mikeller: Yes, you are right. :-) I had assumed that the file is written somewhere else in the script, but never bothered to check.
All in all I don't think always building libdivecomputer is an issue as it builds relatively quickly compared to the rest of the application.
Should we just remove the check to prevent more confusion in the future?

jefdriesen: The libdivecomputer build time is indeed very small compared to the rest of the build. Rebuilding is even faster, unless you always rebuild from scratch (as the script does).

Anyway, it looks like the script is really designed to NOT rebuild the dependencies by default. As the comment in the top of the file explains, you need to manually run `build.<component>` to force a rebuild of a dependency. Thus removing the SHA1 testing (both `$PREVIOUS_SHA` and `*-dirty`) would mean no automatic libdivecomputer rebuilds anymore. That would be consistent with the other dependencies, but I'm not sure if that's the intended result?

mikeller: The way libdivecomputer is used in subsurface through a submodule is reasonably different from how other dependencies are used, where the user has to download the sources separately, and into a directory outside of the repository.
So for these resources the user probably doesn't have the expectation that making local modifications, or switching to a different version will be picked up automatically. I think for libdivecomputer this is different, and users making modifications to either subsurface or the libdivecomputer branch used by subsurface will expect these changes to be picked up - this is why I am proposing to remove the entire check - for the changed SHA, but also for `build.libdivecomputer` and the existence of the `libdivecomputer` directory.
In effect this will retain the current behaviour, as `[ ! "$CURRENT_SHA" = "$PREVIOUS_SHA" ]` currently always evaluates to true, causing the other parts of the or clause to be ignored.

dirkhh: All these scripts used to do dual duty... once in the CI, and once in my home lab where the 'actual' release builds are done. And there those files were written and there not rebuilding the dependencies when nothing changed did make a lot of sense as I was trying to get ready for a release.

Now that we are moving away from that setup, I'll instead clean up the scripts and remove all that overhead since it makes a lot more sense to avoid carrying all this around.

mikeller: Thanks @dirkhh - I'll merge this for now to give @jefdriesen the environment he is after.

PR #3985, opened 2023-11-03
Return the actual libdivecomputer error code

jefdriesen:
The divecomputer_device_open() function tries all supported transports one by one, and exits as soon as one is opened successfully. When the end of the function is reached, the DC_STATUS_UNSUPPORTED error code is returned.

The annoying side effect is that the actual error code returned by the transport is ignored and changed into DC_STATUS_UNSUPPORTED. This is very confusing while troubleshooting download problems.

Fixed by initializing the error code to DC_STATUS_UNSUPPORTED, in case no transport is available for trying, and returning the last reported error to caller.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -1343,7 +1343,7 @@ static dc_status_t bluetooth_device_open(dc_context_t *context, device_data_t *d
 
 dc_status_t divecomputer_device_open(device_data_t *data)
 {
-	dc_status_t rc;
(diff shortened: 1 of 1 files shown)


mikeller (approved):
Looks good to me.

PR #3986, opened 2023-11-05
libdivecomputer.c: rework divecomputer_device_open()

salvadorcunat:
After 1abf400c6 pre-setting "rc" value to DC_STATUS_UNSUPPORTED, perhaps the "if ... return" conditionals would be clearer with a switch/case structure.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [X] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -1353,74 +1353,59 @@ dc_status_t divecomputer_device_open(device_data_t *data)
 	transports &= supported;
 	if (!transports) {
 		report_error("Dive computer transport not supported");
-		return DC_STATUS_UNSUPPORTED;
-	}
-
+	} else {
+		switch (transports) {
 #ifdef BT_SUPPORT
-	if (transports & DC_TRANSPORT_BLUETOOTH) {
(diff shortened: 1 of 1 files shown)


mikeller (changes requested):
Your proposed change is not the equivalent of the existing code - since `transports` is a  bitset, more than one of the `TRANSPORT_<x>` flags can be set, i.e.:
1. the check for equality that `switch` may not match any of the `TRANSPORT_<x>` values;
2. more than one transport may be tried before one returns a success (hence the `if (rc == DC_STATUS_SUCCESS)` tests on the returns, which is a bit of a giveaway.

I think because of 1. there is no straightforward to improve this with a `switch` statement.

salvadorcunat:
Hi Mike.
You are right. I forgot some DCs have more than one way to connect (e.g. ble or serial USB), as mine have just one. For those DCs the switch will fail miserably. Nevertheless the PR doesn't add anything and current code works fine so I'm closing this PR.
Thanks for your review.

PR #3990, opened 2023-11-14
Core: extend last gas mix instead of defaulting to air.

mwerle:
Instead of defaulting to air when we run out of gas mixes to assign to cylinders, use the bottom gas mix provided by the dive computer. The bottom gas mix is the one with the deepest MOD.  Note that no actual MOD calculations are performed, the relative MOD between 2 gas mixes is a simple comparison between the amount of O2 in the mixes or, if equal, helium.

If no gas mixes are provided at all, then default to air.

This prevents Subsurface from "inventing" gas mixes which are not reported by the dive computer. It also works very nicely with a sidemount configuration where the dive computer typically reports two cylinders but only a single gas mix.

- [ ] Bug fix
- [X] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Added a new variable, `bottom_gas`, which tracks the bottom gas mix read from the dive computer. The contents of this variable are assigned to cylinders if there are no more gas mixes.

In the case where the dive computer does not report any gas mixes, the behaviour is identical to the original code - all cylinders are assigned 'air'. 

In the case there are the same (or more) gas mixes than cylinders, the behaviour is also identical to the original code.

In the case where there are less gas mixes than cylinders, the behaviour now differs. Whereas previously all additional cylinders were assigned 'air', these additional cylinders are now assigned the bottom gas mix read from the dive computer.

For example, 

1)
Mixes = [ 99%, 51%, 32%, 'air' ];
Cylinders = [ 1, 2 ];

1 - 99%
2 - 51%

2)
Mixes = [32%];
Cylinders = [ 1, 2, 3 ];

1 - 32%
2 - 32% (previously air)
3 - 32% (previously air)

### Changes made:

1) core/libdivecomputer.c : parse_gasmixes
Calculate the relative MOD between 2 gas mixes and save the one with the deeper MOD as the bottom gas. If there are no more gas mixes, but more tanks, assign this gas to the additional cylinders.

### Additional information:

[subsurface_perdixai_2.log](https://github.com/subsurface/subsurface/files/13345042/subsurface_perdixai_2.log)

### Mentions:

@jefdriesen

### DIFF
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -117,11 +117,11 @@ static int parse_gasmixes(device_data_t *devdata, struct dive *dive, dc_parser_t
 		}
 	}
 	bool no_volume = true;
-	cylinder_t cyl = empty_cylinder;
 	struct gasmix last_mix = gasmix_air; /* default to air */
 
 	clear_cylinder_table(&dive->cylinders);
 	for (i = 0; i < MAX(ngases, ntanks); i++) {
+		cylinder_t cyl = empty_cylinder;
 		cyl = empty_cylinder;
 		if (i < ngases) {
 			dc_gasmix_t gasmix = { 0 };
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -113,14 +113,16 @@ static int parse_gasmixes(device_data_t *devdata, struct dive *dive, dc_parser_t
 			report_error("Warning: different number of gases (%d) and cylinders (%d)", ngases, ntanks);
 		} else if (ntanks > ngases) {
 			shown_warning = true;
-			report_error("Warning: smaller number of gases (%d) than cylinders (%d). Assuming air.", ngases, ntanks);
+			report_error("Warning: smaller number of gases (%d) than cylinders (%d).", ngases, ntanks);
 		}
 	}
 	bool no_volume = true;
+	cylinder_t cyl = empty_cylinder;
+	struct gasmix last_mix = gasmix_air; /* default to air */
 
 	clear_cylinder_table(&dive->cylinders);
-	for (i = 0; i < ngases || i < ntanks; i++) {
-		cylinder_t cyl = empty_cylinder;
+	for (i = 0; i < MAX(ngases, ntanks); i++) {
+		cyl = empty_cylinder;
 		if (i < ngases) {
 			dc_gasmix_t gasmix = { 0 };
 			int o2, he;
@@ -147,9 +149,10 @@ static int parse_gasmixes(device_data_t *devdata, struct dive *dive, dc_parser_t
 				}
 				he = 0;
 			}
-			cyl.gasmix.o2.permille = o2;
-			cyl.gasmix.he.permille = he;
+			last_mix.o2.permille = o2;
+			last_mix.he.permille = he;
 		}
+		cyl.gasmix = last_mix;
 
 		if (rc == DC_STATUS_UNSUPPORTED)
 			// Gasmix is inactive
--- a/core/libdivecomputer.c
+++ b/core/libdivecomputer.c
@@ -123,7 +123,7 @@ static struct gasmix get_deeper_gasmix(struct gasmix a, struct gasmix b)
 	if (get_o2(a) < get_o2(b)) {
 		return a;
 	}
-	if (get_o2(a) > get_o2(b))
+	if (get_o2(a) > get_o2(b)) {
 		return b;
 	}
 	return get_he(a) < get_he(b) ? b : a;
@@ -161,7 +161,10 @@ static int parse_gasmixes(device_data_t *devdata, struct dive *dive, dc_parser_t
 		}
 	}
 	bool no_volume = true;
-	struct gasmix bottom_gas = { {1000}, {0} }; /* Default to pure O2 */
+	struct gasmix bottom_gas = { {1000}, {0} }; /* Default to pure O2, or air if there are no mixes defined */
+	if (ngases == 0) {
+		bottom_gas = gasmix_air;
+	}
 
 	clear_cylinder_table(&dive->cylinders);
 	for (i = 0; i < MAX(ngases, ntanks); i++) {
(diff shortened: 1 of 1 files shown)


core/libdivecomputer.c
```diff
@@ -113,14 +113,16 @@ static int parse_gasmixes(device_data_t *devdata, struct dive *dive, dc_parser_t
 			report_error("Warning: different number of gases (%d) and cylinders (%d)", ngases, ntanks);
 		} else if (ntanks > ngases) {
 			shown_warning = true;
-			report_error("Warning: smaller number of gases (%d) than cylinders (%d). Assuming air.", ngases, ntanks);
+			report_error("Warning: smaller number of gases (%d) than cylinders (%d).", ngases, ntanks);
 		}
 	}
 	bool no_volume = true;
+	cylinder_t cyl = empty_cylinder;
```
mikeller: Not sure why this is moved out of the loop - this seems to have an incorrect scope now, making the code harder to read / understand.

mikeller (commented):
I agree that using a gasmix that is not configured on the dive computer as the gasmix for additional sensors is not the best solution - but not sure if using the _last_ mix is the best solution - on most dive computers this is likely to be the last deco mix that the diver intends to use, which is unlikely to be carried in more than one tank. I think it will be better to use the mix with the highest MOD - this is likely to be the bottom mix, and carried in the largest quantity.

core/libdivecomputer.c
mwerle: Simply a minor optimization - no need to constantly wind/unwind stack frames.  Can revert. I like all variables to be defined (and initialized) at the top of a function but understand that most people prefer to place them throughout the code just prior to first use.

The loop contents should probably be a separate function anyway, overall the `parse_gasmixes` functions is -way- too long/complicated to easily understand. Parts of it perform data acquisition/parsing, parts of it perform the assignment of mixes to cylinders, etc.

mwerle:
> I agree that using a gasmix that is not configured on the dive computer as the gasmix for additional sensors is not the best solution - but not sure if using the _last_ mix is the best solution - on most dive computers this is likely to be the last deco mix that the diver intends to use, which is unlikely to be carried in more than one tank. I think it will be better to use the mix with the highest MOD - this is likely to be the bottom mix, and carried in the largest quantity.

Ok, so my background is with Shearwater computers, which sort mixes by O2 content, meaning deco gases are first, not last.  It also made for a trivial implementation.

Happy to change it.

mikeller:
Yes, your proposal will probably work well for Shearwater computers because of the way they automatically order the configured gasmixes in a 'lowest MOD to highest MOD' order.
Coming from non-Shearwater computers I always found that Shearwater's ordering is done in a 'weird' way and in the opposite direction of what feels 'natural' to me to set up my gases, starting with the first to be used, and going in the order that I use them.
Using the gas with the highest MOD for extra sensors should yield the same results on Shearwater computers as what your proposal does.

mwerle:
Absolutely..  but the code will become more complicated as it will need to firstly figure out the highest MOD gasmix (although I guess it's just the one with the least O2 content so quite trivial anyway), but  secondly, it now needs to first check that we've run out of mixes before assigning the mix to cylinders. Again, not difficult, just not as trivial.

I was initially going to implement this only for situations where there are only 2 cylinders and 1 mix (implying sidemount) but decided that it was too specific and would have added a lot of checks/code inside an already convoluted loop.

For anything but the trivial case the user will most likely need to reassign mixes to cylinders anyway unless the DC has them ordered correctly; not sure if other DC's allow you to assign mixes to transmitters, but Shearwater ones don't (yet).

mwerle:
> (although I guess it's just the one with the least O2 content so quite trivial anyway)
... actually it's not quite as trivial since we use the special value {0,0} to denote air... gotta love shortcuts and special values in code bases...

mikeller:
Hmm, from a quick look at the code I it looked to me like this could be implemented relatively simply with the existing single pass over the list of gases and tanks - keep it as it is, but track something like `extra_gas` in place of `last_gas` and update it if a new gas with higher MOD is found, then use this when you've run out of gasmixes.
But tell you what - probably a better approach will be to pull in the (API breaking) changes in libdivecomputer - I think the latest version supports decoding of the 'gas integrated sidemount' setting on Shearwater computers, and using this to properly identify which sensors are mounted on the redundant bottom gas tanks - for any gas integrated sidemount setup that properly reports remaining gas time on Shearwater computers this will make it possible to identify which tank sensors share the same gas.
I can look into this when I get a bit of time - currently my work situation does not leave me much time or energy to work on this.

mwerle:
All the MOD calculations take in a dive - because it takes into account the actual situation (eg, altitude or freshwater).

We don't have dives yet at this stage in the processing so we can't use those functions verbatim. We're only interested in the relative MOD between mixes, but unless we mostly duplicate the code (eww) from the various  functions in `dive.c` using defaults instead of actual dive data I don't see a trivial way of achieving this.

I may have overlooked something.

mikeller:
Well the thing is that we do not need to calculate the _absolute_ MOD for each gasmix - we just need to rank them in terms of MOD relative to each other. And for this, ranking by lowest O2 percentage, and lowest N2 percentage in case of equal O2 percentage should be sufficient - similar to what Shearwater does when ordering their gas list.

core/libdivecomputer.c
bstoeger: Not really an optimization and makes the code harder to reason about. The compiler should not have to rebuild the stackframe for every pass:
https://godbolt.org/z/q8PWzcE3Y
And if it does, I'd think that it is completely irrelevant performance-wise.

mwerle: Reverted.

mwerle:
Changed the calculation to use the `bottom gas` mix instead of the last mix.

I don't have a suitable dataset to test this. Still "works for me" with the data I have on my Shearwater DC.

mikeller (approved):

PR #3992, opened 2023-11-20
UI: hide first gas change

### LINKED ISSUE #3935: Incorrect gas mix on Subsurface mobile when downloading from Shearwater
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Downloading dives from Shearwater Petrel AI gets incorrect gas information. Dive computer is configured to EAN32 and download to mobile results in EAN33.

### Operating system:

iOS 16.5.1

### Subsurface version:

Beta 3.4.8(5.0.10.177)

### Steps to reproduce:

- Download EAN32 dives to Subsurface mobile.

### Current behavior:

Used cylinder (cylinder 1) is showing EAN33 instead of the used EAN32.

### Expected behavior:

Correct gas mix is shown

mwerle:
Adds an option to hide the first gaschange event, defaulting to `true` to keep the current behaviour.

- [ ] Bug fix
- [X] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The first gaschange event is hidden if it occurs at the beginning of the dive log as it is considered unimportant.

However in certain scenarios Subsurface imports "phantom cylinders" which did not actually exist on a dive. If such a "phantom cylinder" is attached to the first gaschange event, there is no way for the user to delete such a cylinder, since there is no way to edit or delete the first gaschange event using the UI.

This change adds an option to show the first gaschange event. Or rather, to hide it, and it defaults to true. This means that by default, the program behaves as it always did, but now users have an option to toggle hiding this event. If it s shown, the user can edit or delete the event, and thus edit or delete the cylinder to which the event is attached.

### Changes made:

1) Add a new preference option to hide the first gaschange event (defaulting to true). This was added in the `Display -> Dive profile depth grid` section, which has been renamed to `Dive profile options`. Both the existing `Depth line intervals` and the new `Hide first Gas Change` options affect the same part of the Subsurface UI, so it was natural to group them together.

2) Check the new preference option and only hide the first gas change events if this option is set to `true` (the default).

3) Unrelated, but during development of this feature it was discovered that the `map_short_names` preference setting was not being explicitly initialized to `false`. Fixed.

### Related issues:

While this change does not fix #3935, it does allow the user to mitigate it by editing the dive data of affected dives using the UI.

### Release note:

Yes.

"desktop: add `hide first gas change` preference option."

### Documentation change:

New option, `Hide first Gas Change` in the `Preferences -> Display` tab.
This option is checked by default, which means that the first gas change event is hidden from view. This reflects the behaviour of Subsurface prior to release XXX. For most users and most of the time this event is just clutter, however in certain circumstances it may be useful for the user to enable viewing these events in order to be able to edit them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+desktop: add `hide first gas change` preference option.
 desktop: add divemode as a possible dive list column
 profile-widget: Now zomed in profiles can be panned with horizontal scroll.
 desktop: hide only events with the same severity when 'Hide similar events' is used
--- a/core/pref.c
+++ b/core/pref.c
@@ -96,6 +96,7 @@ struct preferences default_prefs = {
 	.extract_video_thumbnails = true,
 	.extract_video_thumbnails_position = 20,		// The first fifth seems like a reasonable place
 	.three_m_based_grid = false,
+	.map_short_names = false,
 	.graph_hide_first_gaschange = true,
 	.sync_dc_time = false,
 };
--- a/core/pref.c
+++ b/core/pref.c
@@ -1,7 +1,10 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "pref.h"
 #include "subsurface-string.h"
+#include "units.h"
 #include "git-access.h" // for CLOUD_HOST
+
+#include <stdbool.h>
 
 struct preferences prefs, git_prefs;
 struct preferences default_prefs = {
@@ -93,6 +96,7 @@ struct preferences default_prefs = {
 	.extract_video_thumbnails = true,
 	.extract_video_thumbnails_position = 20,		// The first fifth seems like a reasonable place
 	.three_m_based_grid = false,
+	.graph_hide_first_gaschange = true,
 	.sync_dc_time = false,
 };
 
--- a/core/pref.h
+++ b/core/pref.h
@@ -106,6 +106,7 @@ struct preferences {
 	bool        show_developer;
 	bool        three_m_based_grid;
 	bool        map_short_names;
+	bool        graph_hide_first_gaschange;
 
 	// ********** Equipment tab *******
 	const char *default_cylinder;
--- a/core/settings/qPrefDisplay.cpp
+++ b/core/settings/qPrefDisplay.cpp
@@ -69,6 +69,7 @@ void qPrefDisplay::loadSync(bool doSync)
 	}
 	disk_three_m_based_grid(doSync);
 	disk_map_short_names(doSync);
+	disk_graph_hide_first_gaschange(doSync);
 }
 
 void qPrefDisplay::set_divelist_font(const QString &value)
@@ -153,6 +154,8 @@ HANDLE_PREFERENCE_BOOL(Display, "show_developer", show_developer);
 HANDLE_PREFERENCE_BOOL(Display, "three_m_based_grid", three_m_based_grid);
 
 HANDLE_PREFERENCE_BOOL(Display, "map_short_names", map_short_names);
+
+HANDLE_PREFERENCE_BOOL(Display, "graph_hide_first_gaschange", graph_hide_first_gaschange);
 
 void qPrefDisplay::setCorrectFont()
 {
--- a/core/settings/qPrefDisplay.h
+++ b/core/settings/qPrefDisplay.h
@@ -27,6 +27,7 @@ class qPrefDisplay : public QObject {
 	Q_PROPERTY(bool singleColumnPortrait READ singleColumnPortrait WRITE set_singleColumnPortrait NOTIFY singleColumnPortraitChanged)
 	Q_PROPERTY(bool three_m_based_grid READ three_m_based_grid WRITE set_three_m_based_grid NOTIFY three_m_based_gridChanged)
 	Q_PROPERTY(bool map_short_names READ map_short_names WRITE set_map_short_names NOTIFY map_short_namesChanged)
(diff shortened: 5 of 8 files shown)


mikeller:
I think this needs some more analyisis (or documentation) of the problem this is trying to solve with second order thinking:
What are the reasons are that this first gas change event is added for some dive computers, and what possible benefits or consequences does this have in what configurations and / or dive profiles?

mwerle:
I have no idea what causes the first gaschange event to be added; it's just a fact of life that it is.

Subsurface then decides that it doesn't need to display it, and, in most situations, that decision is correct. For 99% of all dives there is only one tank, one mix, and the initial gaschange event simply changes the mix to that tank. No need to display it.

But in situations where there are multiple tanks and/or mixes, sometimes Subsurface creates a gaschange event for a tank/mix which either did not exist and is simply an artifact of the import process, or assigns it to a deco tank which is either not used or not used until the end of the dive.  In these situations the user currently has no simple way to edit/modify this initial gas change event.  Combine it with sensor data for some tanks, and there's no way to fudge it either by reordering tanks/mixes/pressures.

In those situations, exposing the inital gaschange event comes in useful. The user can now edit the event to point to an actual tank/mix used at the beginning of the dive, or delete the event entirely.

My own use case:
2x Alu80 sidemount with AI integration to my dive computer (Shearwater)
Optionally additional deco tanks, without AI integration
Shearwater configured with multiple mixes, with inactive mixes disabled.

Currently Subsurface/libdivecomputer 080 imports ALL mixes from Shearwater (fixed in libdivecomputer 090). This results in 5 tanks being imported, in order of oxygen content. The first two mixes are assigned to my pressure data, the rest are assigned to phantom cylinders. The last one is then used for the first gaschange event.  This means I can't delete that cylinder and the only approach has been to manually edit the data using a text editor and git client.

With this change, I can use the UI to assign the first gaschange event to one of my sidemount cylinders with pressure data attached, and that automatically hides the "phantom" cylinder as it is no longer considered to be in-use by Subsurface.

I have a similar issue when diving my CCR.

mwerle:
[perdix_ai_import-20231122-sidemount.zip](https://github.com/subsurface/subsurface/files/13432014/perdix_ai_import-20231122-sidemount.zip)

Example file containing several dives imported from Perdix AI.

Perdix AI was configured with gases 99% (off), 51% (off), 32% (ON), 21% (off), <unconfigured>, and AI with 2x transmitters in sidemount mode.  Dive with 2x Alu80 (EAN32).

Each dive is imported with 1x 99% and 1x 51% tanks, which have the AI transmitter data attached. These tanks can be deleted.  There is an additional tank at 32%, which is attached to the initial gaschange event, and hence cannot be deleted.  Basically the transmitters are assigned 1:1 to the mixes in-order, and then the initial gaschange event is assigned to the next mix in the list.

I can, of course, edit the dive to make the transmitter mixes EAN32, but unless I edit the file manually, there is no way to delete the additional tank.  Also, since this additional tank is attached to the first gaschange event, but has no pressure data, it breaks the SAC calculations.

mwerle:
As for why (some) divecomputers add the first gaschange event, I guess some dive computers might allow the user to  to assign mixes to particular transmitters and/or the user has an interface to manually specify which gas is currently being breathed from.  

During tech-dives I certainly tell my Shearwater when I switch gases but without performing a new import of several hundred dives I don't have a current example as an unedited log for checking exactly what information is present at the time of import.

jefdriesen:
Reporting the initial gas mix on the first sample is needed to inform the application which gas mix is used on the first segment of the dive. Of course, if there is only one gas mix configured, that's not really needed, but once there are multiple gas mixes configured that becomes much more complex (or even impossible) to get right. Thus if the information is available, libdivecomputer always reports the active gas mix, regardless of how many gas mixes are configured. If that's not the case, it's a bug.

vldgeorgiev:
Is there a plan to merge this change?
I have the exact same situation as @mwerle below. There is a hidden gas change event at the beginning of the dive, resulting in wrong display of the gasses or inability to delete the wrong gas. I had to resort to manual editing of the file in the cloud cache folder and pushing to the git repo too.
 
> My own use case: 2x Alu80 sidemount with AI integration to my dive computer (Shearwater) Optionally additional deco tanks, without AI integration Shearwater configured with multiple mixes, with inactive mixes disabled.
> 
> Currently Subsurface/libdivecomputer 080 imports ALL mixes from Shearwater (fixed in libdivecomputer 090). This results in 5 tanks being imported, in order of oxygen content. The first two mixes are assigned to my pressure data, the rest are assigned to phantom cylinders. The last one is then used for the first gaschange event. This means I can't delete that cylinder and the only approach has been to manually edit the data using a text editor and git client.
>

mikeller:
@vldgeorgiev: Please try again with the latest version of Subsurface from https://subsurface-divelog.org/latest-release/?lang=en - this isn't guaranteed to fix it (based on your setup), but it at least hides disabled gases on Shearwater computers.
A more complete solution will require the addition of functionality to Subsurface that allows the user to configure a mapping between gasmixes and tank sensor information.
See https://github.com/subsurface/subsurface/pull/3992#issuecomment-1828640060 for an explanation of why the proposed change is not a viable solution to the problem.

vldgeorgiev:
@mikeller Thanks, I will try the latest build for the next dive download.

I understand the reasoning behind the "initial gas change" event, which is hidden and identifies the starting gas. What I was referring to was the extra setting that @mwerle has added and which allows unhiding that initial gas event. It is only an option to see/hide the event and edit it in the UI, it doesn't seem to change the business logic. 

If the latest build doesn't fix my problem, I would still have to manually map the gasses to the cylinders. This is fine and I am ok doing it, but since the initial gas change is hidden, my only option is to manually edit files with a text editor to correct it.

PR #3994, opened 2023-12-04
Copr and Launchpad builds from GitHub

dirkhh:

- [x] Build system change

The main part is a new GitHub action (combined with new secrets in the repo) that allows us to build Fedora RPMs directly from GitHub. The other two commits are a lame INSTALL file change that I stumbled across when setting up a new VM - and a few fixes caused by new compile flags enforced in Fedora Rawhide that made builds on copr fail

Additional commits to add Ubuntu / Launchpad to the GitHub Actions. There is an issue with missing dependencies (private headers) for the newer of these builds - but that's a Launchpad issue, not a GitHub issue - so the Action to trigger Launchpad builds seems to work already.

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
     - master
-    - github-action-test
   pull_request:
     branches:
     - master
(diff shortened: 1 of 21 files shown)


mikeller (approved):

dirkhh:
and of course because this was a 'rebase and merge' the Launchpad build got rejected as it has the same version number... so the next merge will be the first actual test (and I still need to find the time to create the missing 'private Qt headers' package for Launchpad).
But this does seem like progress.

Given how few openSUSE downloads we have I'll focus on Windows next, I think

PR #4000, opened 2023-12-12
CICD updates

dirkhh:

- [x] Build system change

This is a fairly significant overhaul of the CICD system that creates working (but unsigned) packages for Windows, macOS, and Android, These are now posted in a parallel repo https://github.com/subsurface/nightly-builds as releases - each with a tag that includes the date and a built number.

### Changes made:

1) get the macOS builds to work with the current versions of the dependencies
2) create a convoluted and stupid system to atomically create monotonic build numbers that are identical across the different workflows
3) set things up so that the build artifacts are posted as releases in the nightly-builds repo

### Additional information:

We'll need to test these to make sure there are no surprises - and then we need to update our documentation to reflect how things are set up

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -53,6 +53,19 @@ jobs:
         repository: subsurface/nightly-builds
         token: ${{ secrets.NIGHTLY_BUILDS }}
         prerelease: false
-        body: CICD release artifact
         fail_on_unmatched_files: true
         files: ${{ steps.build.outputs.dmg }}
+        body: |
+          CICD release artifact
+
+          These builds are created on every merge or push into the [Subsurface repo](http://github.com/subsurface/subsurface).
+          This build is based on http://github.com/subsurface/subsurface/commit/${{ github.sha }}
+
+          None of these artifacts are signed.
+          The Android APK can be side-loaded on most Android devices. If you had a previous Subsurface-mobile version installed from the Google Play store, you'll have to uninstall that first.
+          The Windows installer will ask you to confirm installation of an app from an unknown developer.
+          The macOS DMG makes it even harder with a multi-step dance that requires opening the Privacy & Security settings in the System Preferences and explicitly confirming that you are willing to install this app.
+
+          You can find similar Subsurface-Daily builds for [Ubuntu](https://ppa.launchpadcontent.net/subsurface) and Subsurface-test for [Fedora](https://copr.fedorainfracloud.org/coprs/dirkhh/Subsurface-test).
+
+          Please report any issues with these builds in the [Subsurface user forum](https://groups.google.com/g/subsurface-divelog).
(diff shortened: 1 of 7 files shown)


dirkhh:
Hmm - I did the merge locally and force pushed it to the branch so I could merge it here, but the UI still claims that it can't merge. I guess I'll go with my local merge then 🤷🏼‍♂️

Since this only changes CICD / GitHub Action stuff and no actual code in Subsurface, I think I'm ok to not wait for a review from anyone else :)
After all, I'm the main victim when the CICD breaks...

scripts/get-dep-lib.sh
```diff
@@ -12,7 +12,7 @@ CURRENT_LIBSSH2="libssh2-1.8.0"
 CURRENT_XSLT="v1.1.34"
 CURRENT_SQLITE="3190200"
 CURRENT_LIBXML2="v2.9.4"
-CURRENT_LIBFTDI="1.3"
+CURRENT_LIBFTDI="abd19b721f7e9b4d514ed319ece173ebc7b1ea72"
```
mikeller, line 15: Do we _need_ a git hash here? (I.e. is there no released version that will work?)

.github/workflows/mac.yml
```diff
@@ -9,68 +9,81 @@ on:
...
-        DIR=$(pwd)
+        latest=$(cat ../nightly-builds/latest-subsurface-buildnumber)
+        today=$(date '+%Y-%m-%d')
+        version="$today.$latest"
+        echo "version=$version" >> $GITHUB_OUTPUT
 
-        # now Subsurface with WebKit
-        bash -e -x ./subsurface/scripts/build.sh -desktop -build-with-webkit -release
+    # only publish a 'release' on push events (those include merging a PR)
+    - name: upload binaries
```
mikeller, line 66: There _might_ be a way to avoid duplicating the code for the publishing step in every build by centralising the publishing in an action that grabs the artifacts produced by the other builds - what is the rationale for having each action publish on their own?

mikeller (commented):
Some questions / input for possible future improvement, but looks good overall.

scripts/get-dep-lib.sh
dirkhh: Sorry - given time zones I thought you had gone to sleep already...
I couldn't find a released one that (a) built and (b) worked in my macOS test :(

.github/workflows/mac.yml
dirkhh: This is higher level GitHub Action magic that I have never managed to get to work. I was happy to get to this point where I had the "atomic monotonic counter" working and was able to publish the results of one build.
I do not know how to write a yml file that would wait for a different workflow to complete and then access the results of that workflow.

neufeind:
Great that we have updated Android-builds (APK) available again now as well. For a long time there were no newer apks although libdivecomputer already included some fixes. This is resolved now with "nightly" builds.
For me the newer APK makes the Scubapro G2 with firmware 2.x work. So far it only worked with 1.x or with newer libdivecomputer which was not available with subsurface-mobile. And once upgraded you couldn't downgrade the firmware *sigh*

PR #4001, opened 2023-12-15
[smtk-import] Move to current mdbtools github repo

salvadorcunat:
Original mdbtools developer gave up the project some time ago, and it's continued with his permission in a forked repo:
	https://github.com/mdbtools/mdbtools.git

There was a nasty bug in libmdb, triggered under some rare circustances, that is solved in the new repo which is, BTW, under current development.

Move our scripts to the new repo and set our working version to the latest release tag, currently, "v1.0.0"

### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [X] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@dirkhh @atdotde

PR #4004, opened 2023-12-16
[smtk-import] Do not try to import SLG file format 1000

salvadorcunat:

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

@atdotde has just found out in his importer web page some files with SmartTrak DB version 10000.

This is the oldest format I know for SmartTrak databases. Probably the first one.  It just supports one tank, only air/nitrox and the format of the database is dramatically different from the other two formats known to me.

It has different tables, and the "Dives" table differs a lot from newer versions.

I don't think it's worth to give support for this format, as newer versions of SmartTrak software automatically convert the oldest format to newer one.  Thus, finding a lot of this format files is not expected except from some corner cases of users who had not updated their SmartTrak software for years now.

The import, as is right now, produces a disastrous .ssrf file and, yet worse, a crash in the importer, due to the bug in libmdb solved in #4001 .

So that, we just refuse to try to import the file and give a tip to the user on how to solve the import problem.

### Mentions:

@mikeller @atdotde

### DIFF
--- a/smtk-import/smartrak.c
+++ b/smtk-import/smartrak.c
@@ -52,12 +52,15 @@ static void smtk_free(char **array, int count)
 }
 
 /*
- * There are AFAIK two versions of Smarttrak. The newer one supports trimix and up
- * to 10 tanks. The other one just 3 tanks and no trimix but only nitrox. This is a
+ * There are AFAIK three versions of Smarttrak DB format. The newer one supports trimix and up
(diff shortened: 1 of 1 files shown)


mikeller (approved):

dirkhh (approved):
that seems like the right call.

PR #4005, opened 2023-12-18
Build system: Fix Local Build Script for android.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Change the name of the `GITHUB_WORKSPACE` environment variable in the android build script to `OUTPUT_DIR`, which is more intuitive when the script is used for local builds.
Also test if the variable is defined before attempting to use it as the target of the build output.

### Changes made:

1. change the name of the `GITHUB_WORKSPACE` environment variable in the android build script to `OUTPUT_DIR`, which is more intuitive when the script is used for local builds;
2. test if the variable is defined before attempting to use it as the target of the build output.


dirkhh (approved):
Yep, that makes sense

PR #4006, opened 2023-12-28
More cloud servers

dirkhh:

- [x] Functional change

This allows us to use four distinct servers instead of only two.
Right now they are still fairly close to each other (two in Germany, two in California), but they are in different data centers and from different hosting providers, hopefully making all of this a bit more resillient.

### Release note:

This should be completely transparent to the user. People using older versions will only access the -us and -eu server, while people using this code will also try the -u2 and -e2 servers

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,2 @@
+github: subsurface
+ko_fi: dirkhh
--- a/core/checkcloudconnection.cpp
+++ b/core/checkcloudconnection.cpp
@@ -103,7 +103,9 @@ bool CheckCloudConnection::nextServer()
 	};
 	static struct serverTried cloudServers[] = {
 		{ CLOUD_HOST_EU, false },
-		{ CLOUD_HOST_US, false }
+		{ CLOUD_HOST_US, false },
+		{ CLOUD_HOST_E2, false },
+		{ CLOUD_HOST_U2, false }
 	};
 	const char *server = nullptr;
 	for (serverTried &item: cloudServers) {
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -306,7 +306,11 @@ int certificate_check_cb(git_cert *cert, int valid, const char *host, void *payl
 	UNUSED(payload);
 	if (verbose)
 		SSRF_INFO("git storage: certificate callback for host %s with validity %d\n", host, valid);
-	if ((same_string(host, CLOUD_HOST_GENERIC) || same_string(host, CLOUD_HOST_US) || same_string(host, CLOUD_HOST_EU)) &&
+	if ((same_string(host, CLOUD_HOST_GENERIC) ||
+	     same_string(host, CLOUD_HOST_US) ||
+	     same_string(host, CLOUD_HOST_U2) ||
+	     same_string(host, CLOUD_HOST_EU) ||
+	     same_string(host, CLOUD_HOST_E2)) &&
 			cert->cert_type == GIT_CERT_X509) {
 		// for some reason the LetsEncrypt certificate makes libgit2 throw up on some
 		// platforms but not on others
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -887,6 +887,8 @@ static bool create_local_repo(struct git_info *info)
 		if (giterr_last()) {
 			 msg = giterr_last()->message;
 			 SSRF_INFO("git storage: error message was %s\n", msg);
+		} else {
+			 SSRF_INFO("git storage: giterr_last() is null\n");
 		}
 		char *pattern = format_string("reference 'refs/remotes/origin/%s' not found", info->branch);
 		// it seems that we sometimes get 'Reference' and sometimes 'reference'
--- a/core/git-access.h
+++ b/core/git-access.h
@@ -14,7 +14,9 @@ extern "C" {
 #endif
 
 #define CLOUD_HOST_US "ssrf-cloud-us.subsurface-divelog.org"
+#define CLOUD_HOST_U2 "ssrf-cloud-u2.subsurface-divelog.org"
 #define CLOUD_HOST_EU "ssrf-cloud-eu.subsurface-divelog.org"
+#define CLOUD_HOST_E2 "ssrf-cloud-e2.subsurface-divelog.org"
 #define CLOUD_HOST_PATTERN "ssrf-cloud-..\\.subsurface-divelog\\.org"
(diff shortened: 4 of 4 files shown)


mikeller (approved):
The changes look good.

Tested from New Zealand with Subsurface desktop:
- 'normal' with all servers active: `ssrf-cloud-eu.subsurface-divelog.org` is used;
- with the existing servers 'grounded' in the `hosts` file: `ssrf-cloud-u2.subsurface-divelog.org` is used.

I could not test with the mobile version as there seems to be an authentication issue at the moment (see separate email).

I am wondering about some improvements to more evenly distribute the load across the servers:
- should we modify `gotContinent()` to prefer the new (and more grunty) servers precedence (or round-robin between old and new per continent)?
- should we implement a more elaborate scheme to distribute requests in a way that keeps the load / response time across all servers even?

dirkhh:
> The changes look good.
> 
> I could not test with the mobile version as there seems to be an authentication issue at the moment (see separate email).

There are a couple of odd issues with authentication popping up right now which I cannot reproduce. I created a new account to test. I changed my password on an existing account to test. I'm really not sure what is happening here, TBH

I don't think I've seen email from you about the problems you encountered...
 
> I am wondering about some improvements to more evenly distribute the load across the servers:
> 
>     * should we modify `gotContinent()` to prefer the new (and more grunty) servers precedence (or round-robin between old and new per continent)?

So the servers have been renamed... -us and -eu are the new servers, -u2 and -e2 are the previous servers
That seemed like the best approach, because this way people using the old apps still get the new servers

>     * should we implement a more elaborate scheme to distribute requests in a way that keeps the load / response time across all servers even?

Yes, I would love a better system to distribute requests. I just haven't come up with a good plan.
One important caveat is that the client shouldn't change servers once it started to talk to one (you don't want to create git inconsistencies by switching servers in mid-transaction)

mikeller:
> There are a couple of odd issues with authentication popping up right now which I cannot reproduce. I created a new account to test. I changed my password on an existing account to test. I'm really not sure what is happening here, TBH

No problem, let me know when I should test again.

> I don't think I've seen email from you about the problems you encountered...

Concurrency issue... ;-)

> 
> > I am wondering about some improvements to more evenly distribute the load across the servers:
> > ```
> > * should we modify `gotContinent()` to prefer the new (and more grunty) servers precedence (or round-robin between old and new per continent)?
> > ```
> 
> So the servers have been renamed... -us and -eu are the new servers, -u2 and -e2 are the previous servers That seemed like the best approach, because this way people using the old apps still get the new servers

Ah that makes sense - maybe adding comments to the defines for the server hostnames could clarify this.

> > ```
> > * should we implement a more elaborate scheme to distribute requests in a way that keeps the load / response time across all servers even?
> > ```
> 
> Yes, I would love a better system to distribute requests. I just haven't come up with a good plan. One important caveat is that the client shouldn't change servers once it started to talk to one (you don't want to create git inconsistencies by switching servers in mid-transaction)

Good point - the selection should definitely only happen for the first connection attempt after Subsurface is started. Maybe it could even be done on the server side completely, by returning a 'go try another server' answer to `checkServers()` if there are servers with significantly lower load available? Load information to enable servers to make this determination might be baked into the  backend synchronisation protocol.

dirkhh:
> > The changes look good.
> > I could not test with the mobile version as there seems to be an authentication issue at the moment (see separate email).
> 
> There are a couple of odd issues with authentication popping up right now which I cannot reproduce. I created a new account to test. I changed my password on an existing account to test. I'm really not sure what is happening here, TBH

And after writing this here I went back to the reports that I did get and suddenly noted the commonality. Everyone who had problems was using the new European (well: non-Americas) server. And indeed, that one had an issue with the authentication database.

This should be fixed (and has nothing to do with this pull request, really)

dirkhh:
> > So the servers have been renamed... -us and -eu are the new servers, -u2 and -e2 are the previous servers That seemed like the best approach, because this way people using the old apps still get the new servers
> 
> Ah that makes sense - maybe adding comments to the defines for the server hostnames could clarify this.

I can do that... but of course the host names are managed in DNS, so there's a good chance that the DNS reality and what is in the code could get out of sync. But yeah, I'll add that.

> > Yes, I would love a better system to distribute requests. I just haven't come up with a good plan. One important caveat is that the client shouldn't change servers once it started to talk to one (you don't want to create git inconsistencies by switching servers in mid-transaction)
> 
> Good point - the selection should definitely only happen for the first connection attempt after Subsurface is started. Maybe it could even be done on the server side completely, by returning a 'go try another server' answer to `checkServers()` if there are servers with significantly lower load available? Load information to enable servers to make this determination might be baked into the backend synchronisation protocol.

So ironically, this isn't really so much about "load balancing" of the servers, but optimizing access from around the world. This all started when Linus and I were on a dive trip (I think somewhere in Australia?) and had really bad connections to the back then only one backend server, and I noticed that we had MUCH better connectivity to servers in Europe. So I added a second server in Europe and made it pick that one if you weren't in the Americas.

Yeah, it's super simplistic, but it seemed to work. I guess I should add a third set of servers in APAC next? 😇

mikeller:
> > Ah that makes sense - maybe adding comments to the defines for the server hostnames could clarify this.
> 
> I can do that... but of course the host names are managed in DNS, so there's a good chance that the DNS reality and what is in the code could get out of sync. But yeah, I'll add that.

Agreed that having information about the cloud topology embedded in the app isn't great. But we are already doing this when only using the beefier / primary servers as the only candidates when doing geo-selection, so having some context will make it possible to understand the 'why' behind this - maybe even rename the defines to `CLOUD_HOST_US_PRIMARY` / `CLOUD_HOST_US_SECONDARY` or similar.

> Yeah, it's super simplistic, but it seemed to work. I guess I should add a third set of servers in APAC next? 😇

Ha, interesting bit of history. I think these days we have fairly good connectivity to almost everywhere from APAC, thanks to some new-ish fibre-optical links across the Pacific, so we should be fine. :-)

As for the load balancing, this could potentially even be done in a somewhat backwards compatible way by extending the 'teapot protocol':
- have it return the existing 418 answer if there are no preferable servers;
- have it return a different code (possibly 303) with the hostname of the preferred server if there is one.

Legacy clients will just iterate through the client list until they have found an acceptable (not necessarily the preferable) server, whereas new clients will be able to extract the information and jump straight to the preferable server.
On the server side the protocol should be such that at most 2 of the 4 servers will return a 'preferable server' answer at any given time, to reduce the risk of clients getting into a 'no servers available' state if one server becomes unavailable.

mikeller (approved):

PR #4011, opened 2024-01-02
recreate a working iOS device build

dirkhh:
The current XCode and Qt 5.15.2 (the newest version that we can use due to Kirigami and the lack of binaries for the later open source releases of Qt 5.15) have some issues. Work around those.

Also, don't create fat armv7/arm64 binaries anymore for iOS - there are no supported armv7 devices anymore.

- [x] Build system change

This should all be transparent. Famous last words.
I submitted a build based on this to TestFlight - let's see if it gets accepted.

### DIFF
--- a/packaging/ios/build.sh
+++ b/packaging/ios/build.sh
@@ -1,13 +1,15 @@
 #!/bin/bash
+# shellcheck disable=SC2155
 #
 # show what you are doing and stop when things break
 set -x
 set -e
 
 DEBUGRELEASE="Release"
-ARCHS="armv7 arm64 x86_64"
+ARCH="arm64"
 TARGET="iphoneos"
 TARGET2="Device"
+
 # deal with all the command line arguments
 while [[ $# -gt 0 ]] ; do
 	arg="$1"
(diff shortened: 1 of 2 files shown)


dirkhh:
it's so ironic that the iOS build would fail on the PR.
The problem is that the simulator build should be "x86_64-only". Yet `qmake` suddenly decides to build both arm64 and x86_64.
Not sure what's up with that... I need to compare this to a working build to see if there's a thinko in my approach to building either for device or for simulator.

packaging/ios/build.sh
```diff
@@ -72,16 +91,14 @@ fi
...
+# shellcheck disable=SC2002
 cat Info.plist.in | sed "s/@MOBILE_VERSION@/$MOBILEVERSION/;s/@CANONICAL_VERSION@/$CANONICALVERSION/;s/@PRODUCT_BUNDLE_IDENTIFIER@/$BUNDLE/" > Info.plist
 
 popd
 if [ "$versionOnly" = "1" ] ; then
 	exit 0
 fi
 
-# Build Subsurface-mobile by default
-SUBSURFACE_MOBILE=1
```
mikeller, line 83: Could this be screwing with qmake's architecture selection?

dirkhh: Entirely possible.
I was so focused on getting the "real" build working that I may have removed things we need for the simulator build.

mikeller: Shall I have a look at what this does to the android build?

dirkhh: This is just the shell script doing the iOS build - so it shouldn't affect the Android build at all.

I am making good progress on fixing this.

I now have a Simulator build completing again -- only the Device build fails. So I guess it's incremental improvements :)

Seriously, I gotta go to bed as my work day tomorrow starts fairly early, but I think I'll get this figured out fairly quickly.

dirkhh:
OK, with that I think there's no reason not to merge... after all, all this does is mess with the iOS build that no one but me appears to touch...

PR #4012, opened 2024-01-03
Version changes

dirkhh:

- [x] Build system change

this removes the mobile version and consistently uses the 'main' version everywhere
it also changes the release numbering of the nightly builds to something that's based on the main version plus monotonic build number


dirkhh:
pushed rather dramatic and pretty much completely untested code for the draft versioning approach. Let's see how badly this goes (also, of course, many of the use cases don't get used until we push this to master)

dirkhh:
Wow this was painful. But at least for the pull request case it appears to do something marginally useful.
For obvious reasons the `push` case is so far untested... and I would bet money that I broke some things there as well.
@mikeller - I'd appreciate a review for what I have here so far, and then maybe over the weekend we can merge this and see what nonsense it produces with releases. Maybe we could initially change the release announcement text and warn people off from using them unless they know what they are doing.

But at least I think that implements more or less what you and I discussed over the past week or so.

scripts/get-version
```diff
@@ -1,76 +1,80 @@
...
+	# this assumes that (a) we are in a checked out git tree and
+	# (b) we have the ability to check out another git repo
+	# in situations where either of these isn't true, it's the caller's
+	# responsibility to ensure that the latest-subsurface-buildnumber file exists
+	if [ ! -d "$SUBSURFACE_SOURCE/nightly-builds" ] ; then
+		git clone https://github.com/subsurface/nightly-builds &> /dev/null || croak "failed to clone nightly-builds repo"
+	fi
+	cd nightly-builds
+	git fetch &> /dev/null
+	LAST_BUILD_BRANCHS=$(git branch -a --sort=-committerdate --list | grep remotes/origin/branch-for | head -50 | cut -d/ -f3)
```
mikeller: \<nitpick>
`LAST_BUILD_BRANCHES`?
\</nitpick>

.github/workflows/android.yml
```diff
@@ -31,6 +31,16 @@ jobs:
         git push origin main
         cd ..
         bash -x subsurface/scripts/get-or-create-build-nr.sh ${{ github.sha }}
+        cp nightly-builds/latest-subsurface-buildnumber subsurface
```
mikeller, line 34: What is the reason that we are storing this information in a file inside the action, and not loading it into an environment variable or supplying it as a command line parameter to the scripts that consume it?

mikeller (approved):
Wow, great work, this looks good to me.

mikeller:
>  Maybe we could initially change the release announcement text and warn people off from using them unless they know what they are doing.

I don't think that there is much risk of Subsurface eating homework (or divelogs) caused by the versioning / release changes. But we could add a note saying that the versions showing in various places might still be buggy / nonsensical.

.github/workflows/android.yml
dirkhh: Because this is read during build time by `scripts/get-version`
That allows us to use the same mechanism for builds in GitHub or from that files (Launchpad and Copr)

scripts/get-version
dirkhh: oops, forgot to fix - will force push

dirkhh:
Given that the only relevant changes aren't tested in a pull request (but require a push, i.e. merge)... let's merge and see what happens :)

dirkhh:
I pushed a couple more commits that all cannot be tested in a pull request straight to master, avoiding the noise of another PR and pointless GH Action runs... not the best approach... I know... so sue me...

dirkhh:
Gotta love it when shellcheck causes bugs like this 
Well, it's my foolish use of shellcheck I guess.

Will try again

scripts/get-version
```diff
@@ -1,76 +1,80 @@
...
-#      (see below)
-
-set -eu
-#set -x
+# consistently name all builds, regardless of OS or desktop/mobile
+#
+# we do need to be able to create three digit (M.m.p) and four digit (M.m.p.c) version strings
+# default is VERSION_EXTENSION version - an argument of '4' or '3' gets you a digits only version string
+#
+# we hardcode a base version - this will rarely change
```
mikeller, line 9: I think this is just what we need for CICD builds - it makes sure that the version number is defined when the automatic build starts, and doesn't change after this.
One minor improvement could be to have the base version stored in a checked in file of its own - this way it will be trivial to get a version history by looking at the `git log` of this file.

dirkhh: I'm not sure I understand that comment. Can you elaborate?

mikeller: I was referring to hardcoding the base version, which I think is the right thing to do for CICD releases.
But as a small improvement we could store the version base in a file like `subsurface-base-version` and then read it in the script as `$(cat subsurface-base-version)` - I can pull request this if you want.

dirkhh: Given that it's stored in one place in one file... isn't that the better way to do it? That's the one file that you mess with when you want to make changes with the version...
I mean I'm fine either way - I'm just not sure I see the benefit.

PR #4015, opened 2024-01-05
CICD: Update the android Build Docker Image.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Update the android build docker image:
- rebase on ubuntu 22.04;
- add tooling required to sign APKs;
- changes to make the container re-usable;
- change to a multi-stage build to keep the image size smaller;
- generic improvements to the Dockerfile

Also update the example script for how to use the container.

### Additional information:

Since there is no offline installer available for Qt 5.15, the new docker container re-uses the Qt install from the existing 'subsurface/android-build-container:5.15.1' image, by using it as the first stage of the multi-stage build. In order to avoid overwriting this source image the new image is named 'android-build'.

### DIFF
--- a/.github/workflows/android-dockerimage.yml
+++ b/.github/workflows/android-dockerimage.yml
@@ -1,24 +1,18 @@
 name: Android Docker Image CI
 
-# -- disabled for now, as the resulting image is HUGE and causes our
-#    Android builds to fail
-#on:
-#  push:
-#    paths:
-#    - scripts/docker/android-build-container/Dockerfile
-#    - .github/workflows/android-docker*
-
-# this is here to prevent errors about not having an on: clause
 on:
-  repository_dispatch:
-    types:
-    - unused
+  push:
+    paths:
+    - scripts/docker/android-build-container
+    - .github/workflows/android-docker*
+    - packaging/android/android-build-setup.sh
+    - packaging/android/variables.sh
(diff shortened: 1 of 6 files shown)


dirkhh (approved):
this looks very clever. well done.
I'll approve but not merge so you can do this according to your needs.

mikeller:
@dirkhh: If you are ok with the proposed approach I'll do up the action that builds and publishes the docker image, and we can use this to publish the new 'production' image.

dirkhh:
Yes, that sounds great

mikeller:
@dirkhh: This is ready now - once the CICD on it has run it might be good to add a description to https://hub.docker.com/repository/docker/subsurface/android-build/general, and update https://hub.docker.com/repository/docker/subsurface/android-build-container/general to tell users to use https://hub.docker.com/repository/docker/subsurface/android-build/general instead.

dirkhh:
I wonder if the CICD has the necessary credentials to push to docker... 
But yeah, let's merge this and improve from there...

dirkhh:
That's what I was worried about. I created a new access token and re-ran the workflow. Fingers-crossed.

PR #4022, opened 2024-01-08
CICD: Add APK Signing for the Android CICD Pipeline.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Add signing of the android APK to the scripts used by the CICD pipeline.
Also update the example for running these scripts locally, and add the artefacts generated by doing so to .gitignore.

### Changes made:

1. add signing of the android APK to the scripts used by the CICD pipeline;
2. update the pipeline .yml;
3. update the example for running these scripts locally;
4. add the artefacts generated by doing so to .gitignore.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -7,49 +7,56 @@ on:
     branches:
     - master
 
+env:
+  BUILD_ROOT: ${{ github.workspace }}/..
+  KEYSTORE_FILE: ${{ github.workspace }}/../subsurface.keystore
+
 jobs:
   buildAndroid:
     runs-on: ubuntu-latest
     container:
-      image: docker://subsurface/android-build-container:5.15.1
+      image: docker://subsurface/android-build:5.15.2
 
     steps:
     - name: checkout sources
-      uses: actions/checkout@v1
+      uses: actions/checkout@v4
 
     - name: atomically create or retrieve the build number
       id: version_number
       if: github.event_name == 'push'
       run: |
-        cd .. # check out parallel to subsurface sources
+        cd $BUILD_ROOT # check out parallel to subsurface sources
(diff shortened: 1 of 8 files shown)


.github/workflows/android.yml
```diff
@@ -93,10 +96,14 @@ jobs:
           This build is based on http://github.com/subsurface/subsurface/commit/${{ github.sha }}
 
           None of these artifacts are signed.
-          The Android APK can be side-loaded on most Android devices. If you had a previous Subsurface-mobile version installed from the Google Play store, you'll have to uninstall that first.
+          The Android APK can be side-loaded on most Android devices.
```
dirkhh: So that assumes that I add the existing keystore / password as secrets, correct?

dirkhh (commented):
This looks like it's going in the right direction...

.github/workflows/android.yml
mikeller: I sent you an email about that. I'd probably generate a new key and keystore dedicated to CICD releases - there doesn't seem to be a downside to using self signed keys for this.

glance-: The android verification process is basically that on upgrade, the new package is signed by the same key that the old package is signed by. That way you can't switch from one signing lineage of package to another signing lineage without uninstalling the app and removing the data.

dirkhh: Correct - and so our CICD binaries aren't using the same key as the Google Play store binaries (I was convinced that that would be a violation of the terms of my developer agreement), which forces this first reinstall. But after that, every CICD build uses the same key and therefore upgrades work.

PR #4023, opened 2024-01-08
CICD: Improve the Text for CICD Releases.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

1. add a note about the artifacts being added asynchronously
2. add the commit message

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -89,7 +89,9 @@ jobs:
           CICD release artifact
 
           These builds are created on every merge or push into the [Subsurface repo](http://github.com/subsurface/subsurface).
-          This build is based on http://github.com/subsurface/subsurface/commit/${{ github.sha }}
+
+          This build is based on changeset ${{ github.event.head_commit.id }} ([link](${{ github.event.head_commit.url }})):
+          ${{ github.event.head_commit.message }}
 
           None of these artifacts are signed with an official key.
           The Android APK can be side-loaded on most Android devices. If you had a previous Subsurface-mobile version installed from the Google Play store, you'll have to uninstall that first.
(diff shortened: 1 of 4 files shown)


dirkhh (approved):
looks great - I do want to move these texts into a file so that we don't have to keep editing it four times...

mikeller:
@dirkhh:

> looks great - I do want to move these texts into a file so that we don't have to keep editing it four times...

I've had a think about that, and one way to do this will be to create our own custom action(s) that we then can call from within the other actions.

dirkhh:
So IIRC the release action that I use (https://github.com/softprops/action-gh-release) allows you to set a `body_path` instead of defining the body in the workflow. I use that in a different project of mine :)

dirkhh:
So that could be a fixed part of the script that takes the variable components (commit, link to the PR if that's available (I don't think it is because we build at push... but I haven't looked)) and combines them with a standard text, writes it to the final text and then passes a link to that to the `action-gh-release`

mikeller:
> So IIRC the release action that I use ([softprops/action-gh-release](https://github.com/softprops/action-gh-release)) allows you to set a `body_path` instead of defining the body in the workflow. I use that in a different project of mine :)

That will work for the release. But I'd be keen to centralise the 'get version' step as well.

dirkhh:
So on the push to master we get get the details from the pull request that caused this push: https://github.com/marketplace/actions/get-merged-pull-request-info
This means my bad habit of just pushing to master needs to stop :)
Centralizing the `get-version` step shouldn't be too hard... I already use a script for the hard part of it, I'm sure more of what's in the workflow files could get moved into the script.
And really, that stuff I'm less worried about as it doesn't seem to change that often. But I can take a look at making it more concise in the workflow

PR #4024, opened 2024-01-08
[scripts - docker] Fix build instructions

salvadorcunat:
Current instruction to get the installer out the container will fail with bash complaining about cp being a binary. Then would fail too because subsurface-installer.exe doesn't exist any more.

Proposed instruction should work.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh (approved):
LGTM

mikeller (approved):

PR #4025, opened 2024-01-08
iOS: crash fix: don't try to deallocate IosShare

dirkhh:
We have a strange crash on exit on iOS and this looks like the likely culprit. And since it happens at app exit, I'm not too worried about leaking memory...

- [x] Bug fix


mikeller (approved):

dirkhh:
one reason why in the past I have pushed to master was if there were multiple pull requests that I would combine in order to trip the build actions (and Launchpad, and Copr) only once. I guess we can't really do this anymore...

PR #4028, opened 2024-01-09
build-system: avoid output that confuses the script

dirkhh:
The `get-or-create-buildnr.sh` script writes a nice message to `stdout` which is useful when using it interactively - but it broke the scripting; so redirect that output.

- [x] Bug fix
- [x] Build system change

### DIFF
--- a/scripts/get-atomic-buildnr.sh
+++ b/scripts/get-atomic-buildnr.sh
@@ -15,7 +15,7 @@ cd nightly-builds
 git remote set-url origin "$url"
 git push origin main
 cd ..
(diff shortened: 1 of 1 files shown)


dirkhh:
Reading through all the errors, I think this should fix all of them.
But I noticed that we'll likely have to fork and build our own version of the new action as it uses the old, soon to be disabled version of creating output variables.
But really that's not such a bad thing (famous last words).

Given that this either works or doesn't, and that I still don't know how else to test it... I'll simply merge it :(

dirkhh:
Oh well, still not right. `pr_num` is missing. So more work to be done here.
Tomorrow.

PR #4030, opened 2024-01-09
CICD: Fix the release note generation.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Escape special characters in the input to the string substitution.

### DIFF
--- a/scripts/create-releasenotes.sh
+++ b/scripts/create-releasenotes.sh
@@ -4,5 +4,5 @@
 #
 # Usage: create-releasenotes.sh pr_num pr_url pr_title commit_id commit_url
 
-sed "s/PRNUM/$1/;s/PRURL/$2/;s/PRTITLE/$3/;s/COMMITID/$4/;s/COMMITURL/$5/" < gh_release_notes.in > gh_release_notes
+awk "{sub(\"PRNUM\", \"$1\"); sub(\"PRURL\", \"$2\"); sub(\"PRTITLE\", \"$3\"); sub(\"COMMITID\", \"$4\"); sub(\"COMMITURL\", \"$5\"); print}" < gh_release_notes.in > gh_release_notes
 
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ packaging/ios/Qt
 packaging/ios/asset_catalog_compiler.Info.plist
 appdata/subsurface.appdata.xml
 android-mobile/Roboto-Regular.ttf
+gh_release_notes


dirkhh (commented):
oh. my.
I'm sure there is a better way to do this. Mine clearly was stupid.

scripts/create-releasenotes.sh
```diff
@@ -4,5 +4,9 @@
 #
 # Usage: create-releasenotes.sh pr_num pr_url pr_title commit_id commit_url
 
-sed "s/PRNUM/$1/;s/PRURL/$2/;s/PRTITLE/$3/;s/COMMITID/$4/;s/COMMITURL/$5/" < gh_release_notes.in > gh_release_notes
+prurl=$(sed -e 's/[&\\/]/\\&/g; s/$/\\/' -e '$s/\\$//' <<< "$2")
+prtitle=$(sed -e 's/[&\\/]/\\&/g; s/$/\\/' -e '$s/\\$//' <<< "$3")
+commiturl=$(sed -e 's/[&\\/]/\\&/g; s/$/\\/' -e '$s/\\$//' <<< "$5")
```
dirkhh: I've been doing some googling, and there are many other things that we'd have to quote. sed is a bit of a mess.
awk would be much easier... using `sub(pattern, input)` you only need to escape `\` and`&`...

mikeller:
> oh. my. I'm sure there is a better way to do this. Mine clearly was stupid.

XSLT? /ducks

Back to seriousness, I think the 'keep it as simple as possible' approach will be fine for this.

dirkhh:
> > oh. my. I'm sure there is a better way to do this. Mine clearly was stupid.
> 
> XSLT? /ducks
> 
> Back to seriousness, I think the 'keep it as simple as possible' approach will be fine for this.

Oh, I agree. I just realize that the response could contain any URL characters, so a few more that sed would stumble over.
But I'm happy to see us try what you have, TBH.

dirkhh (commented):
don't hate me... your commit message no longer matches...

dirkhh:
OK, I think I have a better way to do this. Much better, actually.

Can you take a look at this action run: https://github.com/dirkhh/testrepo/actions/runs/7467660626/job/20321592864

Click on the "now from a script" step

Here I just print the created release notes to stdout.

As I was playing with this, I actually thought that for more complex merge requests pointing to the last commit wasn't that useful and to drop that part and just link to the pull request used.

What do you think, should I redo the releasenote creation here based on this?

dirkhh:
Actually, @mikeller -- this one is better: https://github.com/dirkhh/testrepo/actions/runs/7467726058/job/20321766525
The other one still had odd line breaks included.

dirkhh:
ditto, also should be closed, right?

PR #4031, opened 2024-01-09
build-system: assemble release notes using gh CLI

dirkhh:
Instead of using a thirdparty action and painfully passing things around, simply use the GitHub CLI (gh) and assemble the release notes on the fly.

This makes for much simpler and much easier to maintain code.

- [x] Bug fix
- [x] Code cleanup
- [x] Build system change

### Related issues:

See #4030 

### Additional information:

This avoids the problems with the thirdparty action and the quoting issues for the arguments and a few other challenges. And it's much easier to read and maintain, IMHO.

### Mentions:

@mikeller -- I know that code reviews for this stuff are hard, but I'd appreciate some careful eyeballs...

### DIFF
--- a/scripts/create-releasenotes.sh
+++ b/scripts/create-releasenotes.sh
@@ -2,7 +2,19 @@
 
 # use to assemble the details for our release notes and write them to a file
 #
-# Usage: create-releasenotes.sh pr_num pr_url pr_title commit_id commit_url
+# Usage: create-releasenotes.sh merge_sha
 
-sed "s/PRNUM/$1/;s/PRURL/$2/;s/PRTITLE/$3/;s/COMMITID/$4/;s/COMMITURL/$5/" < gh_release_notes.in > gh_release_notes
+json=$(gh pr list -s merged -S "$1" --json title,number,url)
 
+cp gh_release_notes_top gh_release_notes
+(
+	echo -n 'This build was created by [merging PR'
+	echo -n $json | jq -j '.[0]|{number}|join(" ")'
+	echo -n '('
+	echo -n $json | jq -j '.[0]|{title}|join(" ")'
+	echo -n ')]('
+	echo -n $json | jq -j '.[0]|{url}|join(" ")'
+	echo ' )'
+) >> gh_release_notes
+
+cat gh_release_notes_bottom >> gh_release_notes
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -7,10 +7,10 @@ on:
     branches:
     - master
 
-
 env:
   BUILD_ROOT: ${{ github.workspace }}/..
   KEYSTORE_FILE: ${{ github.workspace }}/../subsurface.keystore
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
   buildAndroid:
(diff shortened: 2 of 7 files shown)


mikeller (commented):
This looks good except for the copy / paste issue.
Getting rid of the only marginally supported 3rd party action is definitely a move in the right direction.

For posteirority: GitHub action expressions (`${{ ... }}`) has a [`format()`](https://docs.github.com/en/actions/learn-github-actions/expressions#format) function that should be a good tool for string interpolation like this.

scripts/create-releasenotes.sh
```diff
@@ -2,7 +2,19 @@
 
 # use to assemble the details for our release notes and write them to a file
 #
-# Usage: create-releasenotes.sh pr_num pr_url pr_title commit_id commit_url
+# Usage: create-releasenotes.sh merge_sha
 
-sed "s/PRNUM/$1/;s/PRURL/$2/;s/PRTITLE/$3/;s/COMMITID/$4/;s/COMMITURL/$5/" < gh_release_notes.in > gh_release_notes
+json=$(gh pr list -s merged -S "$1" --json title,number,url)
 
+cp gh_release_notes_top gh_release_notes
```
mikeller, line 9: `gh_release_notes_top` now contains the entire message template - I think this needs to be shortened.

dirkhh: eh.

dirkhh:
> This looks good except for the copy / paste issue. Getting rid of the only marginally supported 3rd party action is definitely a move in the right direction.

Fixed. Thanks for catching that. Unsaved editor window :(

> For posteirority: GitHub action expressions (`${{ ... }}`) has a [`format()`](https://docs.github.com/en/actions/learn-github-actions/expressions#format) function that should be a good tool for string interpolation like this.

Yes. I looked at all kinds of options - and there's always /something/ that you have to escape. While the two part manual assembly isn't as elegant, I was hoping it would be more robust.

Of course, once you I merge this it will all blow up in my face and prove my optimism misguided - but that's just Murphy proving his law.

mikeller (approved):
Looks good, let's merge and check the output!

dirkhh:
watching the action status with trepidation... 🤷🏼‍♂️

dirkhh:
While all the build VMs include `gh` and `jq`, our containers do not. Darn.
I had thought "excellent - these are all dependencies that are always installed"... but I clearly didn't think that all the way through.

mikeller:
lol, it's not like the GitHub CLI would be a useful thing to have access to inside a GitHub action... ;-P

Want me to have another go at my idea?

dirkhh:
I think you misunderstand my point.
Our Android and Windows build containers (i.e. the containers that you and I maintain) are missing those tools.
I believe a far more appropriate response would be "would you please add those two apps the next time you rebuild those containers?"

dirkhh:
Personally (and I may be biased) I think that the release note with the link to the correct PR (and it's number/title) is a bit of an improvement...
Screenshot 2024-01-09 at 18 44 28

mikeller:
Yes, I did misunderstand - still wrapping my head around the fact that in these cases the actions controlled by GitHub with their weird pathing and all are actually executing inside a container that is controlled by us. But it seems that your change in #4032 solves this problem intrinsically.
The layout is definitely starting to look good - my idea is more around generating the release notes by interpolating a single file instead of stitching them together from multiple sources.

PR #4032, opened 2024-01-10
build-system: create release notes only once

dirkhh:
Some experimentation showed what should have been obvious. The release information is additive. So it's enough if ONE of the actions creates release notes, all the others can simply add additional release artifacts.

To make this more obvious, this commit creates a new action that does nothing but create the release notes and publish the release. Since it really doesn't do anything else, it's likely to be the quickest to complete, but that doesn't matter - the last action that has a body or body_path in the gh-release action determines the release notes. And we now have exactly one action that does so.

- [x] Bug fix
- [x] Code cleanup
- [x] Build system change

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -10,7 +10,6 @@ on:
 env:
   BUILD_ROOT: ${{ github.workspace }}/..
   KEYSTORE_FILE: ${{ github.workspace }}/../subsurface.keystore
-  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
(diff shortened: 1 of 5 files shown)


dirkhh:
only one way to find out...

dirkhh:
close. so very close.

mikeller:
Good finding, much better than having the release notes generation run by all actions.

dirkhh:
Of course what worked brilliantly in my test repo (with otherwise empty actions to allow rapid testing) is now still stuttering. But I think this is mainly noise. I'm confident that I can get this working in the next ... few... attempts :)

PR #4033, opened 2024-01-10
build-system: small changes for better debugging

dirkhh:
In order to make it easier to see what's happening inside get-atomic-buildnr.sh write the result to a file that can be read by the caller. Not quite as elegant, but hopefully more practical to see what's going wrong when no new build number is created.

Make sure that post-releasenotes is successfull by actually posting a release artifact (apparently the gh release action otherwise quietly fails).

Try to ensure we find the Android APK when uploading to the release.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -25,8 +25,7 @@ jobs:
       id: version_number
       if: github.event_name == 'push'
       run: |
-        version="$(bash scripts/get-atomic-buildnr.sh ${{ github.sha }} ${{ secrets.NIGHTLY_BUILDS }} "CICD-release")"
-        echo "version=$version" >> $GITHUB_OUTPUT
+        bash scripts/get-atomic-buildnr.sh ${{ github.sha }} ${{ secrets.NIGHTLY_BUILDS }} "CICD-release"
 
     - name: store dummy version and build number for non-push build runs
       if: github.event_name != 'push'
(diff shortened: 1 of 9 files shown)


dirkhh:
There are moments when the GitHub actions really are making me reconsider our choice to do these things on GitHub.
`${DEITY}` this is painful.

mikeller:
On the bright side it looks like it's only a few minor kinks left to be ironed out...

dirkhh:
So the Android failure is weird.
It appears to not sign the apk in `qmake-build.sh` because the `KEYSTORE_PASSWORD` variable evaluates as empty (line 414) - even though that seems to be set up correctly from the secret in the `android.yml` file.
Can you take a look at this because to me this looks all correct (and it's unchanged from your code from before, I believe)

dirkhh:
never mind. I think I found it. my editor was clever and re-formatted the .yml :headdesk:

dirkhh:
no take it back... you did that in 867d31bcbc552eda1823426f6edf786260a2152b

dirkhh:
do GitHub Actions automatically export all variables? Oh well, so I'm back to not understanding why the `KEYSTORE_PASSWORD` isn't set

mikeller:
I'll have a look.

mikeller:
> do GitHub Actions automatically export all variables? Oh well, so I'm back to not understanding why the `KEYSTORE_PASSWORD` isn't set

No, turns out they don't, and https://github.com/subsurface/subsurface/blob/30efede730461a5fa51c2bbe38c767e9cce062d4/.github/workflows/mac.yml#L43 is just never used and can probably be removed.

Also, it's not so much the missing variables for signing, as in the absence of these `qmake-build.sh` just copies the APK without re-signing it, but the bug was triggered by `	OUTPUT_DIR` not being set.

PR #4038, opened 2024-01-11
CICD: attempt to read file content in GH yaml

dirkhh:
It is very strange that in some yaml files the $(<release-version) construct works just fine, but in others it evaluates to an empty string, even though the file is there an has the correct content.

Attempting to get more debugging info and also use a different expression to extract the information.

I'm not claiming that I'm confident this will fix the issue - but the extended logging in the last attempt appears to imply that this might be the issue here.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -26,7 +26,7 @@ jobs:
       if: github.event_name == 'push'
       run: |
         bash scripts/get-atomic-buildnr.sh ${{ github.sha }} ${{ secrets.NIGHTLY_BUILDS }} "CICD-release"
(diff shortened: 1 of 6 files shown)


dirkhh:
Well, at least that build failure isn't on me
```
fatal: unable to access 'https://github.com/nih-at/libzip.git/': Could not resolve host: github.com
```
And here I was wondering how I managed to break the iOS test build... 🤦🏻‍♂️

dirkhh:
Now tickle me pink... this appears to have worked...
Now let's clean up all the noise and hope that I don't break it by cleaning up.

PR #4041, opened 2024-01-13
cleanup: fix enum in profile.c

bstoeger:
There are two enums related to the type of dive.
There is the global

enum divemode_t {OC, CCR, PSCR, FREEDIVE, NUM_DIVEMODE,
                 UNDEF_COMP_TYPE};

and the anonymous

enum {AIR, NITROX, TRIMIX, FREEDIVING} dive_type;

in struct plot_info.

In profile.c FREEDIVE (of divemode_t) is assigned to dive_type. This only works because by chance(?) FREEDIVE and FREEDIVING are the fourth element of each enum.

Fix this. C truly is a bad language when it comes to types (very weak) and namespaces (non existing).

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Not really a bug, since the code worked by chance, but still clearly wrong.

Note: detected when trying to compile profile.c with C++, because C string handling is so painful.

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -1302,9 +1302,9 @@ void create_plot_info_new(const struct dive *dive, const struct divecomputer *dc
 	free_plot_info_data(pi);
 	calculate_max_limits_new(dive, dc, pi, in_planner);
 	get_dive_gas(dive, &o2, &he, &o2max);
-	if (dc->divemode == FREEDIVE){
-		pi->dive_type = FREEDIVE;
-	} else 	if (he > 0) {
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
Good catch.

Would it make sense to simply combine the enums instead of having two?

bstoeger:
> Would it make sense to simply combine the enums instead of having two?

I don't know. They seem to represent two orthogonal things, viz. the type of the gas and the breathing apparatus.

PR #4042, opened 2024-01-14
Update AppStream metainfo file

Alexander-Wilms:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

Changes the legacy `/usr/appdata/subsurface.appdata.xml` path to `/usr/metainfo/subsurface.metainfo.xml` (Cf. https://freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location) and updates the file so it passes `appstreamcli validate`

### Changes made:

1) Update file name and installation path from `/usr/appdata/subsurface.appdata.xml` to `/usr/metainfo/subsurface.metainfo.xml`
2) Add more information to metainfo file and update it to the AppStream 1.0 standard


dirkhh (approved):
I'm not super familiar with the spec, but a cursory look at the references seems to indicate that this is indeed the right thing to do.

Thanks!

PR #4046, opened 2024-01-16
Don't crash when we see too many O2 sensors

dirkhh:
I can think of very, very, VERY few circumstances where it makes sense to crash an end user application. This isn't one of them.

- [x] Bug fix
- [x] Functional change

I don't understand why anyone ever thought this made sense. And apparently we now see CCR dive logs with more than 3 PO2 sensors

### Mentions:

@atdotde

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -6,7 +6,6 @@
 #include "gettext.h"
 #include <limits.h>
 #include <string.h>
-#include <assert.h>
 #include <stdlib.h>
 
 #include "dive.h"
@@ -1156,7 +1155,8 @@ static int calculate_ccr_po2(struct plot_data *entry, const struct divecomputer 
 				return sump / 3;
 		}
 	default: // This should not happen
-		assert(np <= 3);
+		// assert(np <= 3);
+		printf("calculate_ccr_po2: np > 3 -- that's a problem, but maybe not enough to crash the app\n");
 		return 0;
 	}
 }


core/profile.c
```diff
@@ -1156,7 +1155,8 @@ static int calculate_ccr_po2(struct plot_data *entry, const struct divecomputer
 				return sump / 3;
 		}
 	default: // This should not happen
-		assert(np <= 3);
+		// assert(np <= 3);
```
mikeller, line 1158: It seems reasonable to _not_ crash here. :-)
The question is should we just change the voting logic to work with an arbitrary number of sensors?

Alternatively we should probably return `entry->o2pressure.mbar` here and not 0, which is a bad value for ppO2.

dirkhh: Robert is working with Jef to figure out the right behavior

mikeller: I have got an idea:
Keep voting out sensors that deviate more than a set percentage from the average of the rest until none of these are left (or only two sensors are left), then average the remaining sensors.
This seems to be aligned with what Shearwater computers (used as controllers in a number of CCRs) do, and very similar to the behaviour of other CCRs that document their inner workings, like the APD Inspiration.

core/profile.c
```diff
@@ -1156,7 +1155,8 @@ static int calculate_ccr_po2(struct plot_data *entry, const struct divecomputer
...
+		printf("calculate_ccr_po2: np > 3 -- that's a problem, but maybe not enough to crash the app\n");
```
bstoeger, line 1159: This message is pretty confusing if you don't know the code. I would suggest something more helpful such as "more than 3 pO2 sensors currently not supported".

PR #4050, opened 2024-01-17
More Coverity fixes / silencing

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Most of this is pointless, because Coverity isn't aware of the copy-on-write character of Qt primitives. But there are at least one real bug fix in there.

### DIFF
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -350,8 +350,8 @@ std::vector<const dive_site *> getDiveSitesToExport(bool selectedOnly)
 	return res;
 }
 
-QFuture<int> exportUsingStyleSheet(QString filename, bool doExport, int units,
-	QString stylesheet, bool anonymize)
+QFuture<int> exportUsingStyleSheet(const QString &filename, bool doExport, int units,
+	const QString &stylesheet, bool anonymize)
 {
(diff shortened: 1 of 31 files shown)


mikeller (approved):
Good work!

I think doing these 'pointless changes' to make Coverity happy has got value, as this will make it easier to use Coverity to stop new problems from being introduced, as there will be less background noise.

dirkhh (approved):
I find reviewing these always challenging... they all seem so obvious, and I always worry about unintentionally introducing some weird error where we actually relied on the (seemingly incorrect) previous behavior.
I certainly have done that myself in the past, trying to address all the Coverity false positives.

But from what I was able to read... these seem fine.

Again, my sincerest thanks to @bstoeger  for doing this thankless work!

bstoeger:
Coverity should only emit these warnings if it is sure that the moved-from variable is not used anymore and has no reference to it, i.e. is created locally. The most likely problem is that a put the `std::move` to the wrong variable owing to tiredness. But I would suppose that Coverity detects a potential use-after-free. The only problem would be reference types, where Coverity can't see whether the variable is used outside of the function.

PR #4054, opened 2024-01-20
 Removed section about GPS recording from user-manual

exxus:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

The GPS tracker functionality was removed from Subsurface-Mobile a while ago. This changes the documentation to reflect the current state of the Subsurface-Mobile application.

### Changes made:

removed outdated information from user-manual.txt 

### Documentation change:

This is a documentation change.

### DIFF
--- a/desktop-widgets/CMakeLists.txt
+++ b/desktop-widgets/CMakeLists.txt
@@ -16,6 +16,8 @@ include_directories(.
 	${CMAKE_BINARY_DIR}
 )
 
+set(CMAKE_AUTOMOC ON)
+
 add_subdirectory(preferences)
 
 set (SUBSURFACE_UI
--- a/Documentation/user-manual.txt
+++ b/Documentation/user-manual.txt
@@ -742,8 +742,7 @@ New dive locations can be recorded in one of four ways:
 
 1. xref:S_existing_locations[Typing in known coordinates by hand]
 2. xref:S_dive_map_location[Positioning the dive site flag on the dive map]
-3. xref:S_Subsurface_mobile[Dive site coordinates from the Subsurface-Mobile phone app]
-4. xref:S_GPS[Dive site coordinates from a GPS track]
+3. xref:S_GPS[Dive site coordinates from a GPS track]
 
 [[S_existing_locations]]
 *(1): Enter coordinates by hand* if they are known, using one of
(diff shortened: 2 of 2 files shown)


desktop-widgets/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ include_directories(.
 	${CMAKE_BINARY_DIR}
 )
 
+set(CMAKE_AUTOMOC ON)
```
mikeller, line 19: Is this meant to be part of this pull request? It is not mentioned in the description.

exxus: Thank you for reviewing the pull request.

The change is mentioned in the corresponding commit message. I needed this change to be able to include the manual in the desktop application.  The built-in documentation needs qt5-webkit which is outdated and not recommended due to security reasons (https://blogs.gnome.org/mcatanzaro/2022/11/04/stop-using-qtwebkit/).
Since it is only used to show the manual and not to browse the web, it may still be ok to use it.
That is why I fixed the CMake file. The change is not directly related to the user manual, why I put it in a separate commit.

I can mention the change in the pull request description or remove the commit that changes the cmake file. Maybe do a separate pull request?
What do you prefer?

mikeller (approved):

desktop-widgets/CMakeLists.txt
mikeller: No, all good, thank you for your explanation.
But it looks like the DCO checker is not happy because the signed-off line does not match your commit author settings: https://github.com/subsurface/subsurface/pull/4054/checks?check_run_id=20682604486

exxus: Thank you for the approval.
I changed my primary email address to match the Signed-off-by section in the commit message (instead of changing the commit message to the obscure email given by github). How to rerun the checks?

dirkhh: Umm, no. QtWebkit is required for printing.
Sadly, and I wish we had found a different way to do this, but that's it's main purpose.
Showing the user manual with is i6s just easier because it's already there. That one could easily be replaced with something else.

I haven't done a full review, please give me time to actually wake up, deal with real life, and then make sure this doesn't break things

dirkhh: Ok, managed to look at the change in more detail on my phone.

This should be ok. I got worried by your explanation 😏

dirkhh:
DCO still fails. Hrmpf.
The simple answer is "never try to run git commands on GitHub" - pretty much everything about how the GitHub UI does things is broken.
Given the discussions here and given the fact that these aren't "key IP changes to the code", I'll just ignore this. I'll just suggest that next time you please do the git commit on an actual computer :)

PR #4057, opened 2024-01-21
Allow for more O2 sensors

atdotde:
The Divesoft Liberty has four O2 sensors. So far, we had a hard coded limit of three sensors and crashed with a failed assert when we encoutered more than three. This allows for up to
MAX_O2_SENSORS which is currently 6. The voting logic is adapted accordingly.

In some of the import/export functions the value is still hard coded to 6 thanks to explicit field names.

In particular, this removes an assert that triggered when more than four sensors were detected.

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -476,9 +476,9 @@ static void populate_plot_entries(const struct dive *dive, const struct divecomp
 		entry->cns = sample->cns;
 		if (dc->divemode == CCR || (dc->divemode == PSCR && dc->no_o2sensors)) {
 			entry->o2pressure.mbar = entry->o2setpoint.mbar = sample->setpoint.mbar;     // for rebreathers
-			entry->o2sensor[0].mbar = sample->o2sensor[0].mbar; // for up to three rebreather O2 sensors
-			entry->o2sensor[1].mbar = sample->o2sensor[1].mbar;
-			entry->o2sensor[2].mbar = sample->o2sensor[2].mbar;
+			int i;
+			for (i = 0; i < MAX_O2_SENSORS; i++)
+				entry->o2sensor[i].mbar = sample->o2sensor[i].mbar;
 		} else {
 			entry->pressures.o2 = sample->setpoint.mbar / 1000.0;
 		}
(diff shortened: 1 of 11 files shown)


core/profile.c
```diff
@@ -1117,48 +1117,47 @@ static void calculate_deco_information(struct deco_state *ds, const struct deco_
...
-		assert(np <= 3);
-		return 0;
-	}
+	else if (np == 1)
+		return entry->o2sensor[sensorn[0]].mbar;
+
+	maxp = np - 1;
+	qsort_r(sensorn, np, sizeof(int), compare_o2_pressures, entry);
+
+	while (maxp - minp > 1 && entry->o2sensor[sensorn[maxp]].mbar - entry->o2sensor[sensorn[minp]].mbar > diff_limit)
```
mikeller: I like the idea of sorting the readings, and then iteratively eliminating outliers.
Most dive computers with support for ppO2 sensors seem to be doing this based on the difference of a reading from the average, and in most cases the limit for discarding a reading seems to be set as a fraction of the average.
What do you think of making Subsurface work in this way as well?

core/sample.h
```diff
@@ -23,7 +24,7 @@ struct sample                         // BASE TYPE BYTES  UNITS    RANGE
 	temperature_t temperature;        // uint32_t   4    mK     (0-4 MK)               ambient temperature
 	pressure_t pressure[MAX_SENSORS]; // int32_t  2x4    mbar   (0-2 Mbar)             cylinder pressures (main and CCR o2)
 	o2pressure_t setpoint;            // uint16_t   2    mbar   (0-65 bar)             O2 partial pressure (will be setpoint)
-	o2pressure_t o2sensor[3];         // uint16_t 3x2    mbar   (0-65 bar)             Up to 3 PO2 sensor values (rebreather)
+	o2pressure_t o2sensor[MAX_O2_SENSORS];// uint16_t 6x2    mbar   (0-65 bar)             Up to 6 PO2 sensor values (rebreather)
```
bstoeger, line 27: I wonder at which point this should become a dynamic table dependent on `dc->no_o2sensors`.

atdotde: I would propose I do that the next time the number of sensors has to be increased. Already now, this is the special case (Divesoft Liberty) of a special case (CCR dives). Already now, I think four (two redundant units that are in themselves already redundant) is a bit high...

core/profile.c
atdotde: I just submitted an updated commit that uses distance from average as you suggest. I still use total spread as the condition if some sensor reading has to be removed as this to me seems to be the better indicator that there is an outlier.

atdotde:
qsort_r is not available on all platforms and for at most 6 values it is probably a bit of overkill, anyway. So I replaced it with an explicit bubble sort.

core/profile.c
mikeller: Would it make sense to add a comment explaining why we are using logic here that is different than what dive computers are using, and why this is better?
I think that even if we are using total spread as condition for eliminating readings then the spread limit should be relative to the average. Since cell voltage and its distortion by the environment are linear to ppO2, the threshold should also be relative to the cell output. In practical terms, discarding readings with a spread of > 0.1 bar at 1.3 bar seems to be overly selective and will get a lot of readings eliminated that are considered acceptable by dive computers, whereas at 0.7 bar 0.1 bar seems to be ballpark reasonable.
For example, if we wanted to keep in line with what is used in Shearwater computers (which are likely to be the most common type of technical dive computers used for CCR diving), we should be discarding readings only if they have a spread of more than 0.4 times the average (identical to discarding individual readings if they diverge for more than 20% from the average).

atdotde: Do you have a source for this? I did some searching and found that Shearwater uses distance to the closest other value and no mention of the average. I am hesitant, because if you have a faulty sensor, the average is not really meaningful.

What is the basis for the claim that relative error is more relevant than absolute? It might be but it's not obvious to me.

mikeller: > Do you have a source for this?

https://www.shearwater.com/wp-content/uploads/2022/05/Petrel-3_Technical_Manual_RevB.pdf, page 57, 'Dual Sensor Mode' - their manual only elaborates on the 2 sensor case, but from personal observation and reports from other CCR divers in forums the same approach seems to be used when 3 sensors are used.

> I did some searching and found that Shearwater uses distance to the closest other value and no mention of the average.

That's interesting - where did you find this?

> I am hesitant, because if you have a faulty sensor, the average is not really meaningful.

Yes, for the 'average' approach to work the average will have to be taken from the other sensors, excluding the one that is being evaluated.
But 'distance to the closest' seems to be reasonable as well, as long as the limit for exclusion is relative to the absolute reading.

> What is the basis for the claim that relative error is more relevant than absolute? It might be but it's not obvious to me.

Factors that impact have a relative impact on the sensor output:
- cell age (i.e. anode to electrolyte surface area) ;
- moisture on membrane;
- temperature.

Factors that have an absolute impact on the sensor output:
- short circuit (reading will be discarded no matter what);
- open circuit (no pun, reading will be discarded no matter what).

Not in this list, as not easily detectable without taking the history of readings into consideration:
- stuck cell;
- current limited cell.

atdotde:
OK, I changed the logic to that of Shearwater (described in https://www.shearwater.com/wp-content/uploads/2018/08/Shearwater-NERD-UserManual-DocRevB.pdf page 55): As long as we have more than two sensors we remove those that differ by more than 20% from the closest other sensor value.

Still no idea why some builds keep failing the TestProfile.

dirkhh:
So that's running the tests in a mobile-on-desktop build. You should be able to reproduce this locally...

atdotde:
I should. But here the test passes (both in desktop and mobile-on-desktop build and I don't really see where that would make a difference for that test).

atdotde:
Is it somehow possible that those builds don't see the updated reference files? Is it somehow possible to see the files these test runs produced?

mikeller:
It looks like the problem is that in your updated CSV reference files some of the floating point numbers have been converted to a format using comma as the decimal separator, probably due to having been generated on a system with a presumably German locale setting:

```
mikeller@L052KV:~/git/subsurface$ git diff --word-diff --patience master dives/exportprofilereference.csv
diff --git a/dives/exportprofilereference.csv b/dives/exportprofilereference.csv
index 72635b2a5..c0672dc55 100644
--- a/dives/exportprofilereference.csv
+++ b/dives/exportprofilereference.csv
@@ -1,10439 +1,10439 @@
"in_deco", "sec", "pressure_0_cylinder", "pressure_0_interpolated", "temperature", "depth", "ceiling", "ceiling_0", "ceiling_1", "ceiling_2", "ceiling_3", "ceiling_4", "ceiling_5", "ceiling_6", "ceiling_7", "ceiling_8", "ceiling_9", "ceiling_10", "ceiling_11", "ceiling_12", "ceiling_13", "ceiling_14", "ceiling_15", "percentage_0", "percentage_1", "percentage_2", "percentage_3", "percentage_4", "percentage_5", "percentage_6", "percentage_7", "percentage_8", "percentage_9", "percentage_10", "percentage_11", "percentage_12", "percentage_13", "percentage_14", "percentage_15", "ndl", "tts", "rbt", "stoptime", "stopdepth", "cns", "smoothed", "sac", "running_sum", "pressureo2", "pressuren2", "pressurehe", "o2pressure", "o2sensor0", "o2sensor1", "o2sensor2", "o2setpoint", "scr_oc_po2", "mod", "ead", "end", "eadd", "velocity", "speed", "in_deco_calc", "ndl_calc", "tts_calc", "stoptime_calc", "stopdepth_calc", "pressure_time", "heartbeat", "bearing", "ambpressure", "gfline", "surface_gf", "density", "icd_warning"
"0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", [-"0.000000", "0.000000", "0.000000",-]{+"0,000000", "0,000000", "0,000000", "0", "0", "0",+} "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "STABLE", "0", "0", "0", "0", "0", "0", "0", "0", "0", [-"0.000000", "0.000000", "0.000000", "0.000000",-]{+"0,000000", "0,000000", "0,000000", "0,000000",+} "0"
```

Due to the fact that we are just string-comparing the CSV files this is resulting in test failures - probably a shortcoming of  how our tests are written, as we've had this before: https://github.com/subsurface/subsurface/pull/3917#issuecomment-1577989569

mikeller (approved):
Thanks for changing this - having this aligned with how dive computers are doing this will be valuable.

Good find on the Shearwater documentation of the algorithm used - the same paragraph is in the Petrel manual as well, but it is in a hard to find location (page 67 in https://www.shearwater.com/wp-content/uploads/2022/05/Petrel-3_Technical_Manual_RevB.pdf).

atdotde:
Now with English locale (I was on a laptop with newly installed Ubuntu. That had English as system language set but still used a German locale for numbers...)

atdotde:
OK, it was indeed the locale issue. Which brings the question: Why did only some of the builds notice this?

mikeller:
> OK, it was indeed the locale issue. Which brings the question: Why did only some of the builds notice this?

Some of our GitHub actions do not include test runs:
- actions that do builds for mobile devices (the respective tests are run in the actions that contain the mobile-on-desktop builds);
- actions that only push data and or trigger builds in third party packaging systems.

The test failures look a bit weird as in most cases the mobile-on-desktop builds and tests are run in the same action as the desktop builds and tests, so if the mobile tests fail the desktop tests aren't even attempted. But in terms of being economical with build resources (of which we get a limited free allocation), I think the way the actions are currently set up is reasonable.

dirkhh:
So I think this is ready to be merged, correct?

PR #4060, opened 2024-01-21
Ubuntu: remove obsolete dependency

dirkhh:
It appears that this dependency is no longer provided (as of Mantis), but also no longer needed (as a build without it appears to completed). Let's see if that fixes our Mantis build issue.

- [x] Bug fix
- [x] Build system change

### DIFF
--- a/packaging/ubuntu/debian/control
+++ b/packaging/ubuntu/debian/control
@@ -4,7 +4,6 @@ Priority: optional
 Maintainer: Subsurface CI Bot <dirk@subsurface-divelog.org>
 Build-Depends: asciidoc,
                debhelper (>= 9),
-               libgconf2-dev,
                libtool,
                libxml2-dev,
                libxslt-dev,


mikeller (approved):

dirkhh:
Of course the PR checks are pointless since this only affects a Launchpad build and that only happens on the push/merge 🤷

mikeller:
> Of course the PR checks are pointless since this only affects a Launchpad build and that only happens on the push/merge 🤷

I guess we just need to get used to remembering to add `[skip ci]` to the last commit message in these pull requests: https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

Update: I guess we _do_ want the build for launchpad to happen in this case. But what we could do is add files used only in `push` builds to the `paths-ignore` directive for `pull_request` triggers.

dirkhh:
the challenge with this, if I understand the docs correctly, is that this would also prevent the push workflows from running when merged - and those are the ones that I needed here.

mikeller:
> the challenge with this, if I understand the docs correctly, is that this would also prevent the push workflows from running when merged - and those are the ones that I needed here.

In the `[skip ci]` case yes.
But doing this with `paths-ignore` in the workflow files should work.
As much as working out and maintaining `paths` and `paths-ignore` rules for all the different workflow files will add complexity, I think the benefits of avoiding noise and wastage of our runtime allocation outweigh this.

PR #4064, opened 2024-01-25
desktop: add password reset to cloud preferences

dirkhh:
This must be the number one support request we get. I can't believe we never thought of adding a button to do this.

- [x] Functional change
- [x] New feature

### Documentation change:

I guess this needs to be documented in the user manual, the FAQ, etc.


mikeller (approved):
:-)

PR #4070, opened 2024-01-28
enable creating current releases for Ubuntu and Fedora

dirkhh:
While for the other platforms we can simply copy our binaries (maybe after signing them), for Fedora and Ubuntu we have to trigger fresh builds.

The most logical way that I could think of to do this was to push the same commit corresponding with the intended current release into a branch named 'current' and have that trigger Copr and Launchpad builds that post into our release repos.

So 'master' keeps moving forward, keeps creating new build numbers. At some point we pick a build number that we want to be the next 'current' release. We then update the current branch to the commit that corresponds to that build number and push the current branch which triggers new builds in the correct repos on Copr and Launchpad.

This commit removes the silly 'push' argument from the make-package scripts (after all, they are used to push those packages to the respective build services) and instead use the branch name as argument to those scripts - allowing us to pick which repo to push into.

- [x] Build system change

unsurprisingly this is untested... and we can't really test it until we pick the next 'current' build that comes AFTER this gets merged. Oh well.

### DIFF
--- a/.github/workflows/fedora-copr-build.yml
+++ b/.github/workflows/fedora-copr-build.yml
@@ -5,6 +5,7 @@ on:
     - scripts/docker/**
     branches:
     - master
+    - current
 
 jobs:
   setup-build:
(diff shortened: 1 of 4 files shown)


mikeller (approved):
> unsurprisingly this is untested... and we can't really test it until we pick the next 'current' build that comes AFTER this gets merged. Oh well.

But luckily we aren't building life support equipment here. :-)

Looks good, I like how the scripts work in a branch-dependent way.

PR #4071, opened 2024-01-28
Mobile: Fix Gradient Factor Preference Setting.

### LINKED ISSUE #4052: Use imported GF (instead of the Tech menu entry) to calculate TTS
### Describe the issue:

- [ ] Bug
- [X] Change request
- [X] New feature request
- [ ] Discussion request

### Issue long description:

Would it be possible to use the imported GF instead of the entered in the tech diving menu entry ?
Actually I have to switch the GF every time I change it for my dive.
Use of the imported from the computer would be better.

### Operating system:

Linux Kubuntu
Windows 10
French

Shearwater Perdix AI + Petrel 3

Yes ; on every OS

### Subsurface version:

6.0.5051-CICD

Yes

Official release

### Steps to reproduce:

Only open a dive

### Current behavior:

Subsurface uses the GF in the menu

### Expected behavior:

Should use the imported GF from the computer to calculate the TTS

### Mentions:

subsurface

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix the persisting and use of gradient factor preferences for dive profiles in the mobile version.
Also rename the mobile backend gradient factor settings to make it obvious that they are used by the (not currently enabled) planner.

### Changes made:

1. fix the persisting and use of gradient factor preferences for dive profiles in the mobile version;
2. rename the mobile backend gradient factor settings to make it obvious that they are used by the (not currently enabled) planner.

### Related issues:

Fixes #4052.

### DIFF
--- a/mobile-widgets/qml/DivePlannerSetup.qml
+++ b/mobile-widgets/qml/DivePlannerSetup.qml
@@ -230,12 +230,12 @@ TemplatePage {
 					from: 10
 					to: 150
 					stepSize: 1
-					value: Backend.gflow
+					value: Backend.planner_gflow
 					textFromValue: function (value, locale) {
 						return value + volumeUnit
 					}
 					onValueModified: {
-						Backend.gflow = value
+						Backend.planner_gflow = value
 					}
 				}
 
@@ -247,12 +247,12 @@ TemplatePage {
 					from: 10
 					to: 150
 					stepSize: 1
-					value: Backend.gfhigh
+					value: Backend.planner_gfhigh
 					textFromValue: function (value, locale) {
 						return value + volumeUnit
 					}
 					onValueModified: {
-						Backend.gfhigh = value
+						Backend.planner_gfhigh = value
 					}
 				}
 
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -639,8 +639,8 @@ TemplatePage {
 					textFromValue: function (value, locale) {
 						return value + "%"
 					}
-					onValueModified: {
-						PrefTechnicalDetails.gflow = gfLow.text
+					onValueChanged: {
+						PrefTechnicalDetails.gflow = value
 						rootItem.settingsChanged()
 					}
 				}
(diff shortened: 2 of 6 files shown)


dirkhh (commented):
I'm a bit confused here... which part of the change actually fixes the bug?

mikeller:
> I'm a bit confused here... which part of the change actually fixes the bug?

The change from `onValueModified` to `onValueChanged` (alongside with assigning the value, and not the text in the event handler) in the QML.
The seemingly 'anything goes' nature of QML, together with less than helpful documentation, makes me love Qt just a little less...

And I only figured this out after a long wild goose chase around the mobile backend properties `gflow` / `gfhigh`, only to figure out that they seem to be intended for the planner.

dirkhh:
It would be hard for me to love QML/Qt any less than I already do.
It was the right choice when we made it. Today I dearly wish we had made different choices.

atdotde:
Wait, this does not sound right to me. There is no planner on mobile. Why would the mobile version care about the planner GF settings? It needs the GF only to display the ceiling. And when not in planner, that should be controlled by the GF settings from the prefs, not the ones in the PlannerPointsModel, shouldn't it?

mikeller:
> Wait, this does not sound right to me. There is no planner on mobile.

There is no planner that is currently exposed to the user, but there definitely are some fragments intended for a planner, like in https://github.com/subsurface/subsurface/blob/fb41999c73d8e4a3933e40a5d835d9ff627699c4/mobile-widgets/qml/DivePlannerSetup.qml#L229-L257 where the `planner_gflow` and `planner_gfhigh` backend properties are referenced.

And personally I think that a planner will be the most valuable addition to be done for Subsurface-mobile, and I am keen to put some time into having a go at making what we have work. For this reason I am reluctant to just rip out bits of the unused fragments that already exist.

> And when not in planner, that should be controlled by the GF settings from the prefs, not the ones in the PlannerPointsModel, shouldn't it?

That's exactly what we are doing - we use the gradient factors directly from the preferences, without an intermediate property in the mobile backend:

https://github.com/subsurface/subsurface/pull/4071/files#diff-59b03f793fa3af1afe01d29d78c68ef1d7cdef99f9ce8f33204a31f92174ed63L638-R660

PR #4078, opened 2024-02-03
 snap: use the rolling release version

Saviq:
- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

This makes the snap build process retrieve the buildnumber from nightly-builds, if not already populated (e.g. by the GitHub CI setup). I've disabled the `push` builds on `master` as this will build on the Launchpad / Snapcraft infrastructure, triggered by a push.

### Changes made:
1) pick up the rolling version number
2) skip building `master` on GitHub
3) check security notices on the `candidate` channel as well.

### Mentions:
@dirkhh hey, this should work for the new version numbers, I've set up the git `current` -> snap `candidate` builds [here](https://launchpad.net/~subsurface/+snap/subsurface-candidate). I'll work on refreshing the snap after we're happy with all this.

### DIFF
--- a/.github/workflows/linux-snap.yml
+++ b/.github/workflows/linux-snap.yml
@@ -1,11 +1,6 @@
 name: Linux Snap
 
 on:
-  push:
-    paths-ignore:
-    - scripts/docker/**
-    branches:
-    - master
   pull_request:
     paths-ignore:
     - scripts/docker/**
--- a/.github/workflows/linux-snap.patch
+++ b/.github/workflows/linux-snap.patch
@@ -1,3 +1,5 @@
+diff --git a/snap/snapcraft.yaml b/snap/snapcraft.yaml
+index 6b9542555..7f5ccabf2 100644
 --- a/snap/snapcraft.yaml
 +++ b/snap/snapcraft.yaml
 @@ -35,6 +35,7 @@ parts:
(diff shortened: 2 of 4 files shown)


mikeller (approved):
Looks good, thank you for updating this!

.github/workflows/linux-snap.yml
```diff
@@ -1,11 +1,6 @@
 name: Linux Snap
 
 on:
-  push:
-    paths-ignore:
-    - scripts/docker/**
-    branches:
-    - master
```
dirkhh, line 8: Why would this NOT run on push to master - that's how we usually trigger the builds...
So I would have thought the other way around... don't run on pull!

dirkhh (commented):
thanks for working on this... see my one comment... my assumption was that the 'edge' snap would rebuild on a push to `master`, and the 'candidate' snap would build on a push to `current`

.github/workflows/linux-snap.yml
Saviq: It _will_ build on `master`, but in https://snapcraft.io/subsurface/builds, for all the architectures together - and publish to the `edge` channel. And for `current` and `snap-stable`, it will build [on Launchpad](https://launchpad.net/subsurface/+snaps). That's how it was until now (modulo `current` -> `candidate`, which is new).

I mean, we _could_ build here, but only for amd64 - for any other architecture we'd need to send it to Launchpad anyway (or qemu... slooow), so I thought to focus on that - it's a better infrastructure for everything snap :)

So the actions workflow I've left for PR validation, but published builds come from LP.

dirkhh: Sorry about my confusion. Makes sense

dirkhh:
Nice... We already have the correct version for edge!
Next time we rev current that should propagate to candidate as well 

Thank you so much, Michal

PR #4079, opened 2024-02-04
Mobile: Fix Configuration of Ceiling Display.

### LINKED ISSUE #4052: Use imported GF (instead of the Tech menu entry) to calculate TTS

mikeller:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix the configuration of the deco ceilings in the mobile version.

### Changes made:

1. make the settings work;
2. remove reading of the dive computer ceiling from git;
3. hide the gradient factor in the profile when the calculated ceiling is not shown;
4. when the calculated ceiling is disabled in the settings, disable editing of the gradient factor.

### Related issues:

Fixes #4052 (baby one more time)

### Additional information:

image
image

@dirkhh: There are a number of other settings in the mobile application (units, display of tank bar, display of ccr information) that we attempt to load from git:

https://github.com/subsurface/subsurface/blob/1cc7c05170b4d7a188f88feabb4dcd800ebef85a/mobile-widgets/qmlmanager.cpp#L279-L289

Presumably these are meant to be persisted by the desktop app? I tried to get them to work but could not. What is the intent here, should these be fixed, or should these be removed from git and added as settings in the mobile app?

### Release note:

mobile: fix configuration of decompression ceiling and gradient factors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 desktop: add a button linking to the 'Contribute' page
+mobile: fix configuration of decompression ceiling and gradient factors
 desktop: fix gas switches in UDDF exports
 core: allow of up to 6 O2 sensors (and corresponding voting logic)
 desktop: add divemode as a possible dive list column
--- a/core/dive.c
+++ b/core/dive.c
@@ -3048,8 +3048,6 @@ void set_git_prefs(const char *prefs)
 {
 	if (strstr(prefs, "TANKBAR"))
 		git_prefs.tankbar = 1;
-	if (strstr(prefs, "DCCEILING"))
-		git_prefs.dcceiling = 1;
 	if (strstr(prefs, "SHOW_SETPOINT"))
 		git_prefs.show_ccr_setpoint = 1;
 	if (strstr(prefs, "SHOW_SENSORS"))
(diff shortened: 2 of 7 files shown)


dirkhh:
The idea was to not deal with the settings on mobile and have them set on desktop.

I think today most users would prefer to just deal with the mobile UI and make the changes there. 

So removing them from git seems reasonable

Issue #4086, opened 2024-02-06
Crashes in Dive Planner


BlackFire01:
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
When using the dive planner, there are two quite annoying crashes I encountered. This is bad because if not saved, all the planned deco gases etc. are cleared and you start from scratch. In both of these cases, when the action leading up to the crash is made, the app freezes and then crashes completely.
The first one sometimes appears when deleting a tank. See below for how to reproduce it, in my case 100% of the time.
The second one is a bit more troubling. It happens when you have the tissue heatmap enabled with multiple gasses. If you then try to set a surface interval after the dive to judge off-gasing effectiveness, the app just crashes. You cannot set the interval at all. Also occurs if you accelerate the decompression with more than one gas already set.

### Operating system:
My system is running Windows 11, Version 10.0.23615 Build 23615.
I have tested both crashes with English and German language setting, they occur regardless.

### Subsurface version:
I am running Subsurface 6.0.5079-CICD-release.
The issues have been occurring since I started working with the dive planner.
I did not compile the code myself.

### Steps to reproduce:

Both crashes are completely independent of algorithm choice.

Crash 1 - Tank Removal
1) Go to Dive Planner.
2) Set the first depth marker to something deeper (e. g. 60m - 200ft).
3) Set the second default depth marker at the same depth for some time (I chose the default 19 minutes).
4) Add a tank not containing air. I chose EAN50. Tank properties should not matter, but I selected 12l steels and 200bar initial pressure.
5) Add another tank not containing any of the currently available gases. I chose EAN99.
6) Now simply delete the first tank you added (should be in the middle) and see the window vanish.

Crash 2 - Surface Interval
1) Go to Dive Planner.
2) Add another tank (even air works).
4) Enable tissue heatmap.
5) Now, add a surface interval on the bottom of the middle column of the dive parameters. Crash ensues.

### Current behavior:
Well, the app crashes.

### Expected behavior:
In the first case, the gas should simply be removed and the decompression schedule re-adjusted if necessary.
For the second part, one should expect a surface interval to show effectiveness of off-gassing in heatmap, typically used for accelerated deco planning.

dirkhh:
Well, I tried to reproduce this - but of course I avoid doing anything on Windows as much as I can. And on Linux I cannot reproduce the crash. So I'll keep looking at this - just so you know this isn't being ignored.

[...]

PR #4088, opened 2024-02-07
planner: fix crash when planning with surface interval

bstoeger:
The planner uses a one-past-end pseudo cylinder for marking the surface interval outside of water. This overflowed arrays in setup_gas_sensor_pressure().

See #4086. Note: contains a second unrelated crash report.

As a band-aid allocate bigger arrays. But obviously, the proper fix is to not generate invalid gas-change events.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Hopefully fixes one of the crashes reported in #4086.

### DIFF
--- a/core/profile.c
+++ b/core/profile.c
@@ -751,12 +751,14 @@ static void setup_gas_sensor_pressure(const struct dive *dive, const struct dive
 	if (pi->nr_cylinders == 0)
 		return;
 
-	int *seen = malloc(pi->nr_cylinders * sizeof(*seen));
-	int *first = malloc(pi->nr_cylinders * sizeof(*first));
-	int *last = malloc(pi->nr_cylinders * sizeof(*last));
+	/* FIXME: The planner uses a dummy one-past-end cylinder for surface air! */
+	int num_cyl = pi->nr_cylinders + 1;
+	int *seen = malloc(num_cyl * sizeof(*seen));
+	int *first = malloc(num_cyl * sizeof(*first));
+	int *last = malloc(num_cyl * sizeof(*last));
 	const struct divecomputer *secondary;
 
-	for (i = 0; i < pi->nr_cylinders; i++) {
+	for (i = 0; i < num_cyl; i++) {
 		seen[i] = 0;
 		first[i] = 0;
 		last[i] = INT_MAX;
@@ -769,7 +771,11 @@ static void setup_gas_sensor_pressure(const struct dive *dive, const struct dive
 		int sec = ev->time.seconds;
 
 		if (cyl < 0)
+			continue; // unknown cylinder
+		if (cyl >= num_cyl) {
+			fprintf(stderr, "setup_gas_sensor_pressure(): invalid cylinder idx %d\n", cyl);
 			continue;
+		}
 
 		last[prev] = sec;
 		prev = cyl;


dirkhh (approved):
Thanks for submitting that. It certainly seems correct to me.

the Snap failure appears to be missing secrets (which makes sense on a pull request... you don't want to make those available to un-tested code).

mikeller (approved):
Looks good to me.
I think the current approach to calculate surface intervals as dive segments with depth 0 while breathing from a tank of air of unlimited size is not half bad - it allows re-use of the existing logic for this.
And having this segment included in the SAC rate calculation should not have a negative effect, as without having the start / end volume of the 'big tank in the sky' set the segment will be ignored in this calculation as a required input value is missing.

atdotde:
I think the proper way to do this would be to make this special infinite cylinder be number 0 and have loops start at 1 when it should not be included.

dirkhh:
For now I think we'll go with "merge and keep iterating"...

bstoeger:
> I think the proper way to do this would be to make this special infinite cylinder be number 0 and have loops start at 1 when it should not be included.

I fear that would be *many* places. For example, I reckon that the whole models code would look pretty awkward.

A further problem: what about saved dives? Would we have to juggle around existing dives? Will this introduce incompatibilities with older versions?

I tried a different approach: introduce a special cylinder id for "surface air" and at least it doesn't seem to crash with a few fixes. However, the problem of saved dives remains: what if such a "new" dive is loaded into an old version?

bstoeger:
Some more experimentation with master (before application of the patch) shows that:

- The surface interval is *not* reloaded, when editing a planned dive in the planner. Instead, the `prefs.surface_segment` value is taken.
- The `prefs.surface_segment` value seems not to persist across application exit.
- The switch to the "surface air" cylinder is not saved into the XML file.
- However, the surface interval is properly recognized if one edits a saved plan.

This is all a bit obscure...

atdotde:
In my understanding, the surface segment is not really part of a dive but rather more like a display option: It should be viewed in context of the heat map: For many dives, a large part of the over saturation of the tissues actually happens after surfacing and this is when a substantial fraction of bubbles is produced (as you can verify with Doppler measurements at intervals after the dive). So to analyse/visualise the amount of this over saturation you need to show the heat map for the time after the end of the dive. This is what this setting is for. If you like, we should not save this part of the dive (since it's not part of the dive) but have a display option to show the saturation after starting to breath air at the surface pressure (which can be computed rather than read from disk).

bstoeger:
Ok, that raises two questions: Should the surface interval be shown only for planned dives or for all dives? Should it always be shown or only in the planner?

If the gas change is not saved, I'm not sure save/reload currently does the correct thing.

bstoeger:
> If the gas change is not saved, I'm not sure save/reload currently does the correct thing.

Actually, it does recognize the surface interval and switches to non-existing "air". Interesting.

mikeller:
@bstoeger:

> > I think the proper way to do this would be to make this special infinite cylinder be number 0 and have loops start at 1 when it should not be included.
> 
> I fear that would be _many_ places. For example, I reckon that the whole models code would look pretty awkward.

Agreed on the fear about inconsistencies introduced by using index 0 for the 'surface air' gasmix - this will also make our indices different from the ones reported by libdivecomputer.

I guess we could use `-1` or `INT_MAX` for this mix if we wanted to use a constant id. But 'last + 1' seems to be working as it is.






@atdotde:
> In my understanding, the surface segment is not really part of a dive but rather more like a display option: It should be viewed in context of the heat map: For many dives, a large part of the over saturation of the tissues actually happens after surfacing and this is when a substantial fraction of bubbles is produced (as you can verify with Doppler measurements at intervals after the dive). So to analyse/visualise the amount of this over saturation you need to show the heat map for the time after the end of the dive. This is what this setting is for. If you like, we should not save this part of the dive (since it's not part of the dive) but have a display option to show the saturation after starting to breath air at the surface pressure (which can be computed rather than read from disk).

Fully agreed - I think the UI for the surface segment that is currently implemented in the planner makes sense. The only thing that might be worth considering is to allow the surface segment to be displayed for logged dives as well - as you say, the heat map often shows a lot of supersaturation for the time after surfacing.

bstoeger:
> I guess we could use -1 or INT_MAX for this mix if we wanted to use a constant id.

-1 is already reserved for "unknown" as per `core/event.h`. So the logical extension would be -2. In my experiment, it seems to work with some minor tweaking. Note that in the long run we probably want to C++-ify all the "table" code and then it would be nice if the cylinder tables work consistently as all the other tables.

> But 'last + 1' seems to be working as it is.

Not really. Case in point: this patch and numerous other kludges.

I like Robert's idea of not saving the surface interval at all and adding it on the fly in the planner and on the profile. This would make it possible to also show it for logged dives, if the users prefers. But since that is deeply interwoven with the profile I prefer not doing it as long as I'm working on the accursed QtQuickification of the profile.

mikeller:
> > I guess we could use -1 or INT_MAX for this mix if we wanted to use a constant id.
> 
> -1 is already reserved for "unknown" as per `core/event.h`. So the logical extension would be -2. In my experiment, it seems to work with some minor tweaking. Note that in the long run we probably want to C++-ify all the "table" code and then it would be nice if the cylinder tables work consistently as all the other tables.

Makes sense.

> > But 'last + 1' seems to be working as it is.
> 
> Not really. Case in point: this patch and numerous other kludges.

Well it's working _now_. 😉

> I like Robert's idea of not saving the surface interval at all and adding it on the fly in the planner and on the profile. 

To my understanding this is how it already is done in the planner, and is probably not far from working for logged dives as a lot of the profile code is shared.

> But since that is deeply interwoven with the profile I prefer not doing it as long as I'm working on the accursed QtQuickification of the profile.

That makes sense.

There are other improvements to cylinder management that I would like to work on, most notably getting away from the fixed 1 : 1 mapping of gasmixes to tank sensor data sets as they are read from libdivecomputer. One prerequisite for this is to update to the latest API implemented by libdivecomputer, which I am doing in https://github.com/subsurface/libdc/pull/56.  But I can hold off of any work on this until you are done.

PR #4092, opened 2024-02-11
core: remove eventitem.*

bstoeger:
The sole purpose of this source file was to remember whether some events should be hidden globally. However, that feature was removed in 26f20b805d3c95445f0cbc2c68a6d59dd6c7d915.

So we might just remove the whole thing.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Removes dead code. Note: I don't no whether the removal of the feature (globally hide events of a certain kind) was intentional.

### Release note:

No - code cleanup.


bstoeger:
Actually, this is also a bit of a bug fix. I don't see how the "show all events" menu item could have ever been shown...

bstoeger:
Hmmmm.... looking closer it appears that the removal of `hide_event()` was not intentional. Let me know.

dirkhh:
I don't recall intentionally removing that. It seems like a feature that we used to have - but I'd have to dig through git history to understand what / how this all happened... Didn't we have a UI where you click on an event and can hide that event or all of that type? I just tried and couldn't figure it out - but I'm pretty sure we had this at some point...

mikeller:
https://github.com/subsurface/subsurface/commit/26f20b805d3c95445f0cbc2c68a6d59dd6c7d915 was mine, and it was not intentional that it removed the capability to persist hiding of events - I simply overlooked that `hide_event()` worked globally across all dives currently loaded, whereas `event->hide()` only affected the currently showing dive.
So the right way to go is probably to restore this functionality.

bstoeger:
Ok, then let's restore it.

PR #4093, opened 2024-02-12
Profile: Reinstate Hiding of Events by Type across Dives.

mikeller:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Reinstate the hiding of events by event type across all dives in the log. This was unintentionally removed in #3948. Also change the event type to be specific to name and severity, and fix bug causing 'Unhide all events' to not show when only individual events were hidden.

This still leaves the inconsistency that hiding of similar events is persisted across the switch between dives, but hiding of individual events is lost when switching dives, which is mildly confusing.

### Changes made:

1. reinstate the hiding of events by event type across all dives in the log;
2. change the event type to be specific to name and severity
3. fix bug causing 'Unhide all events' to not show when only individual events were hidden.

### Related issues:

Follow-up to #4092.

### DIFF
--- a/core/event.c
+++ b/core/event.c
@@ -80,7 +80,7 @@ struct event *create_event(unsigned int time, int type, int flags, int value, co
 		break;
 	}
 
-	remember_event_name(name);
+	remember_event_name(name, flags);
 
 	return ev;
 }
--- a/core/eventname.cpp
+++ b/core/eventname.cpp
@@ -8,15 +8,16 @@
 
 struct event_name {
 	std::string name;
+	int flags;
 	bool plot;
 };
 
 static std::vector<event_name> event_names;
 
 // Small helper so that we can compare events to C-strings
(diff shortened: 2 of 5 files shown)


bstoeger:
What a coincidence - I just made my own version, which is a bit more radical...

bstoeger (approved):
Yes, fine with me.

I had a somewhat more involved version making the severity thing explicit: https://github.com/subsurface/subsurface/commit/8b923e0e222af13876b0c83b62c511b821d223d4

But we can do these things later with the QtQuickification...

mikeller:
Oh no, apologies for the double up. But at least we are both going in the same direction.
I don't mind which version we use.

bstoeger:
Simple is good for now.

PR #4094, opened 2024-02-12
Mobile: Fix Cloud Synch Status Management.

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix how the cloud synch status is handled - currently the preference is used incorrectly, switching cloud synch off when it should be on. Also hide the 'Manual cloud synch' button when no cloud credentials are stored, and don't show the 'manual synch' popup when auto synch is enabled.

### Changes made:

1. fix how the cloud synch status is handled ;
2. hide the 'Manual cloud synch' button when no cloud credentials are stored;
3. don't show the 'manual synch' popup when auto synch is enabled.

### Related issues:

From https://groups.google.com/g/subsurface-divelog/c/ccHX2Zse3Qc/m/aNi3OHtwAQAJ

### DIFF
--- a/mobile-widgets/qml/CloudCredentials.qml
+++ b/mobile-widgets/qml/CloudCredentials.qml
@@ -125,6 +125,7 @@ Item {
 				id: signin_register_normal
 				text: qsTr("Sign-in or Register")
 				onClicked: {
+					PrefCloudStorage.cloud_auto_sync = true
 					manager.saveCloudCredentials(login.text, password.text, "")
 				}
 			}
--- a/mobile-widgets/qml/main.qml
+++ b/mobile-widgets/qml/main.qml
@@ -334,7 +334,7 @@ Kirigami.ApplicationWindow {
 						name: ":/icons/cloud_sync.svg"
 					}
 					text: qsTr("Manual sync with cloud")
-					enabled: Backend.cloud_verification_status === Enums.CS_VERIFIED
+					visible: Backend.cloud_verification_status !== Enums.CS_NOCLOUD
 					onTriggered: {
 						globalDrawer.close()
 						detailsWindow.endEditMode()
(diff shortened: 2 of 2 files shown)


mobile-widgets/qml/main.qml
```diff
@@ -334,7 +334,7 @@ Kirigami.ApplicationWindow {
 						name: ":/icons/cloud_sync.svg"
 					}
 					text: qsTr("Manual sync with cloud")
-					enabled: Backend.cloud_verification_status === Enums.CS_VERIFIED
+					visible: Backend.cloud_verification_status !== Enums.CS_NOCLOUD
```
dirkhh, line 337: well - that's a behavioral change that might create confusion.
If the user has requested to create an account, but not verified their PIN, the entry will now be available, but the sync won't work.

mobile-widgets/qml/CloudCredentials.qml
```diff
@@ -125,6 +125,7 @@ Item {
 				id: signin_register_normal
 				text: qsTr("Sign-in or Register")
 				onClicked: {
+					PrefCloudStorage.cloud_auto_sync = true
```
dirkhh, line 128: why do you want to enable auto_sync by default?
An unsuspecting user could create quite hefty data bills while roaming with this setting - that's why it's off by default and you need to manually turn it on.

dirkhh (commented):
I'm trying to understand what you mean by the last item in your commit message. Which 'manual synch' popup are you blocking?

mobile-widgets/qml/CloudCredentials.qml
mikeller: Apologies, the curse of knowledge has hit me when writing the commit message.
The actual relevant bit of change is this here:

https://github.com/subsurface/subsurface/pull/4094/files#diff-40007045ffbbfbaeff5d836fd0b0974f15574ea83cfb11783da2837aed84f63dR354

I.e. in the existing version `cloud_auto_sync == true`  means 'cloud auto synch is disabled', which is likely the root cause for the warning message (https://github.com/subsurface/subsurface/pull/4094/files#diff-40007045ffbbfbaeff5d836fd0b0974f15574ea83cfb11783da2837aed84f63dL356) being shown at the wrong time (when auto synch is enabled), which is what Martin had reported and what triggered my investigation.

This change actually does not change the behaviour with respect to auto synch being enabled after the user has successfully set the cloud credentials - this change is just accommodating the fact that `cloud_auto_sync == true` now means 'cloud auto synch is enabled'.
I get your point about data usage, but:
- after entering cloud credentials, an intitial synch is always done anyway, and if anything this will be the one that downloads megabytes of log - subsequent updates should be smaller;
- I am not sure that we need / want to give the user a way to specifically moderate data usage inside Subsurface. If this is the problem they want to solve they will probably want to do this in the mobile OS configuration by only allowing selected apps to use mobile data. Stopping subsurface from synchronising logs while keeping mobile data enabled for all apps is probably not going to give them the outcome they want;
- the default has now been for years that cloud auto synch was on after successfully entering credentials, and this does not seem to have resulted in negative user feedback.

mobile-widgets/qml/main.qml
mikeller: For what I can tell it's the other way round - in the old version the 'Manual synch' button is showing but cannot be clicked, making it appear 'dead' to the user. Now it is hidden, like the 'auto synch switch' button.

PR #4095, opened 2024-02-13
Mobile: improve GFLow/High settings UI

dirkhh:
- move the spinbox closer to the senter
- improve sizing and spacing of the spinbox
- hide it when we aren't showing the calculated ceiling

Also address an odd whitespace issue.

- [x] Functional change

it's been quite a while since I've played with the QML code. Kinda fun to try my hand at this again.
This is just a tiny improvement triggered by the discussions toward the end of #4052
With this the spinboxes should be much easier to see and use.

### DIFF
--- a/mobile-widgets/qml/Settings.qml
+++ b/mobile-widgets/qml/Settings.qml
@@ -536,11 +536,11 @@ TemplatePage {
 			GridLayout {
 				visible: sectionAdvanced.isExpanded
 				width: parent.width
-				columns: 2
+				columns: 3
 
 				TemplateLine {
 					visible: sectionAdvanced.isExpanded
-					Layout.columnSpan: 2
+					Layout.columnSpan: 3
 				}
 				TemplateLabel {
 					text: qsTr("Bluetooth")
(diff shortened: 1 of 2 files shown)


mobile-widgets/qml/Settings.qml
```diff
@@ -647,12 +654,14 @@ TemplatePage {
 					}
 				}
 				TemplateLabel {
-					enabled: PrefTechnicalDetails.calcceiling
+					visible: PrefTechnicalDetails.calcceiling
```
mikeller, line 657: Good move, from what I've seen in #4094, there is no obvious visual difference between enabled and disabled elements (at least in our styling), so it's probably more user friendly to go with hiding inactive elements.

mikeller (approved):
Ugh, how the arrangement of elements is done looks quite awkward - is there a good primer on how this works?

mobile-widgets/qml/Settings.qml
dirkhh: The grid model isn't great.

There are a few decent introductions to QML, but all are way too simple and never cover the stuff you really want to see 😞

I have done a lot of trial and error

PR #4099, opened 2024-02-19
Better handling of event hiding

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Some tweaking of the event code. User "visible" changes:
- Differentiate between unhiding of event types and individual events.
- Remember hiding of individual events across dive switch.

Also some code cleanup: rename `event_name` to `event_type`.

User interface changes and more fine grained unhiding, I will probably do in #3863.

### Release note:

No - wait for #3863.

### Documentation change:

No - wait for #3863.


profile-widget/profilewidget2.cpp
```diff
@@ -610,7 +610,7 @@ void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
 	if (DiveEventItem *item = dynamic_cast<DiveEventItem *>(sceneItem)) {
 		m.addAction(tr("Remove event"), [this,item] { removeEvent(item); });
 		m.addAction(tr("Hide event"), [this, item] { hideEvent(item); });
-		m.addAction(tr("Hide similar events"), [this, item] { hideSimilarEvents(item); });
+		m.addAction(tr("Hide events of this type"), [this, item] { hideEventType(item); });
```
mikeller: Not sure what the user expectation is for  'events of this type' means when they are looking at events with identical names but different severity - to me this implies all events with the same name.

dirkhh (approved):
This does seem reasonable.
Certainly having the hidden state of events persist across dive changes is a huge plus.
And I assume that this will increase the desire to actually store this state as well.

profile-widget/profilewidget2.cpp
```diff
@@ -656,8 +656,11 @@ void ProfileWidget2::contextMenuEvent(QContextMenuEvent *event)
 		}
 #endif
 	}
-	if (any_events_hidden() || std::any_of(profileScene->eventItems.begin(), profileScene->eventItems.end(), [] (const DiveEventItem *item) { return !item->isVisible(); }))
-		m.addAction(tr("Unhide all events"), this, &ProfileWidget2::unhideEvents);
+	if (any_event_types_hidden())
+		m.addAction(tr("Show all event types"), this, &ProfileWidget2::unhideEventTypes);
+	if (std::any_of(profileScene->eventItems.begin(), profileScene->eventItems.end(),
+	    [] (const DiveEventItem *item) { return item->getEvent()->hidden; }))
+		m.addAction(tr("Show hidden events of this dive"), this, &ProfileWidget2::unhideEvents);
```
mikeller: While I can see the value of the fine grained control that you are implementing here, I am not sure if users will understand what the options do. 'Show hidden events of this dive' to me implies that _all_ hidden events for this dive are shown, so maybe -Show individually hidden events of this dive'?
Also, to me 'Show hidden events' implies that these events will be shown (temporarily), but stay hidden, as opposed to 'unhide'.

mikeller (approved):
The functionality looks good, but I have some comments about how it is advertised to the user.

mikeller (approved):

bstoeger:
Make unhiding even more fine grained: let the user choose which event-type to show.

mikeller (approved):
This is very comprehensive functionality now, I like it!

PR #4101, opened 2024-02-24
mobile: show incorrect cloud password state

dirkhh:
While the startup flow should make it obvious when a user is not correctly logged into the cloud, we still do see fairly frequent situations where a user has an incorrect password on a mobile device and is confused about why their data isn't syncing with their PC. Now this is clearly shown in the main menu.

- [x] Functional change
- [x] New feature


dirkhh:
my absolute favorite build failures are when the GitHub action fails because the build process can't connect to... "GitHub"...
 `¯\_(ツ)_/¯`

mikeller (approved):

PR #4110, opened 2024-02-28
cleanup: fix memory leak (probably / untested)

bstoeger:
get_local_dir() returns the copy of a c-string. It therefore has to be free()d.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix a memory leak.

Note: C-memory management is just unmaintainable. If nobody objects I will start converting the C sourcefiles to C++ and then replace C strings by std::string or QString. My plan is the use the former (UTF-8 stored as C string) in the "core". However, the separation will not be completely clear.

### DIFF
--- a/core/git-access.c
+++ b/core/git-access.c
@@ -161,7 +161,9 @@ char *get_local_dir(const char *url, const char *branch)
 static char *move_local_cache(struct git_info *info)
 {
 	char *old_path = get_local_dir(info->url, info->branch);
-	return move_away(old_path);
+	char *new_path = move_away(old_path);
+	free(old_path);
+	return new_path;
 }
 
 static int check_clean(const char *path, unsigned int status, void *payload)


mikeller (approved):
Looks good to me, and YES to moving to C++.

dirkhh (approved):
Given that the fierce opponent to this approach basically doesn't contribute anymore, this seems fine

bstoeger:
Is it me or is the code in `load_git.c` completely intractable?

Note how `pop_cstring()` returns a copy of a string.

In `parse_kayvalue_entry` it may or may not be called, which is already ominous. How do we know if `val` now points to the newly copied string?

Then look at one of the users of the value: `parse_cylinder_keyvalue`. The value may be stored in `cylinder->type.description` or not stored. Seems a bit ... inconsistent?

The second call is in `parse_settings_divecomputerid`. Where is this free()d?

Can someone shed light on this, because I certainly can't make any sense of it.

dirkhh:
That is mainly Linus' code

bstoeger:
Could someone of you Unix experts explain this, please:

In the function `readfile` in `file.c` don't we have to check the return code of `read` for `EINTR` or incomplete reads? I remember faintly that reading from files under Unix was always a bit annoying. This code was written by Linus, so I reckon it is correct. But still would be interesting what the guarantees are...

bstoeger:
Also sometimes it sets `errno` and returns -1, sometimes it returns the negative error code...?

torvalds:
> Is it me or is the code in `load_git.c` completely intractable?

The word you are looking for is "efficient". This was written because our XML parser was way too slow when you have thousands of dives.

torvalds:
> 
> In the function `readfile` in `file.c` don't we have to check the return code of `read` for `EINTR` or incomplete reads?

For regular files, not really.

Yeah, yeah, some network filesystems aren't really POSIX-compatible, so it's not necessarily a bad idea to deal with EINTR and partial reads, but mounting a NFS filesystem with "intr" is legacy and a bad idea. On Linux, for example, you can still *kill* a process in the middle of a read, but returning EINTR or partial results does violate POSIX and will break a number of apps.

(On Linux, if you do a read over 2GB in one go, the kernel will also say "screw you, you're trying to DDoS the system, I will truncate your insane read". Our readfile() basically doesn't care about either of those situations)

torvalds:
Also, please DO  NOT try to convert load-git.c.

That code doesn't work on strings. It works on the whole file read into one large buffer, and then has two separate buffers: the fixed-size "line" buffer on stack, and the dynamically allocated sized - and *REUSED* - "struct membuffer str". So that 'str' thing will create one single allocation that ends up being re-sized and re-used potentially over and over for the whole parsing of the file.

Allocating individual strings and copyiong them is part of what made the XML parser horrendously slow. 

(The other part of the XML parser overhead is of course "XML". Complete garbage data format, basically designed to be hard to parse for both humans and computers)

bstoeger:
Cool, thanks for the info.

Yes, the whole XML parser library is not optimal, as it does lots and lots of copying. For example for tag and attribute names it is not necessary to create a new strings. Ultimately, the library should be replaced by something with better performance, but that is very far down on the list of priorities.

No worries - I was never planning to create new strings. Just replace `char *` by `std::string` and `memblock` by `std::vector<char>`, which automates memory management and simplifies resizing, etc. This should give negligible performance penalties.

Dirk has a huge divelog with thousands of dives - I will not propose anything that makes reading it (user-)noticeably slower.

Still I think we are better off by (slowly) replacing `char *` with `std::string`, if only because I am too stupid to work with C-style memory management.

bstoeger:
One further example: in `parse_txt_file()`, where is the data of `memtxt` or `memcsv` free()d? Is it?

bstoeger:
Merge this, or directly go with #4117, which also fixes it?

dirkhh:
I prefer to merge this now instead of waiting for #4117

dirkhh:
(hehe, given that we show the PR title in the changelog, I should have edited the PR title... but I thought about this two tenths of a second after I clicked confirm...)

bstoeger:
> (hehe, given that we show the PR title in the changelog, I should have edited the PR title... but I thought about this two tenths of a second after I clicked confirm...)

Huh? Since when?

mikeller:
> Huh? Since when?

Since we have moved to our new website: https://www.subsurface-divelog.org/latest-release/ :-)

PR #4115, opened 2024-03-01
[TRIVIAL(?)] cleanup: remove ominous pointer calculation

bstoeger:
Firstly, why calculate something when the next statement is a return anyway.

Secondly, the calculation subtracts two completely unrelated pointers.

This must be some code reshuffling artifact.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This code left me very confused until I realized that it has no effect, so can be removed.

### DIFF
--- a/core/import-csv.c
+++ b/core/import-csv.c
@@ -222,10 +222,8 @@ static int parse_dan_format(const char *filename, struct xml_params *params, str
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
that seems to make sense.

PR #4117, opened 2024-03-02
Port some parts of string handling to C++ (part 1 of many)

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

I am convinced that we should port core structures to C++, notably strings and collections. Experience shows that again and again, we are unable to produce proper C code that doesn't leak left and right and can be reasoned about. Yes, some galaxy brains with infinite time can do this, but that's just not a luxury we have.

The end goal is to convert the core structures such as `struct dive`, etc. But since these are accessed everywhere, we first have to convert everything to at least compile in C++. This PR is a humble start, concerning mostly string handling of the parser. It is a very long way to go.

I want to do this in "smallish" packages and let them stabilize, otherwise things will get unmanageable for me.

Just by this mostly trivial replacement of C-strings by C++-strings, the memory leaked by `TestParser` got cut nearly in half:
```
==704562== LEAK SUMMARY:
==704562==    definitely lost: 4,010 bytes in 21 blocks
==704562==    indirectly lost: 13,657 bytes in 35 blocks
==704562==      possibly lost: 480 bytes in 1 blocks
==704562==    still reachable: 54,603 bytes in 371 blocks
==704562==         suppressed: 0 bytes in 0 blocks
```
vs
```
==705012== LEAK SUMMARY:
==705012==    definitely lost: 2,221 bytes in 18 blocks
==705012==    indirectly lost: 13,657 bytes in 35 blocks
==705012==      possibly lost: 0 bytes in 0 blocks
==705012==    still reachable: 20,581 bytes in 96 blocks
```

Comments:

- This is a pretty much automatic conversion and yes, there will be new bugs introduced.
- Yes, this introduces an intermediate state, which is quite ugly.
- This uses std::string, which has a performance regression: It clears the data when allocating. So when reading a file, the memory is cleared just to be overwritten. If someone shows me that this has a noticeable effect, I will solve this problem with a custom class, but not before. The reason for using std::string instead of vector<char>: std::string guarantees to have a trailing 0-byte *after* the actual data. The parser actually expects this behavior.

More to come...

### DIFF
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -78,7 +78,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/worldmap-save.c \
 	core/libdivecomputer.c \
 	core/version.c \
-	core/save-git.c \
+	core/save-git.cpp \
 	core/datatrak.c \
 	core/ostctools.c \
 	core/planner.c \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -82,7 +82,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/datatrak.c \
 	core/ostctools.c \
 	core/planner.c \
-	core/save-xml.c \
+	core/save-xml.cpp \
 	core/cochran.c \
 	core/deco.c \
 	core/divesite.c \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -95,7 +95,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/strtod.c \
 	core/tag.c \
 	core/taxonomy.c \
-	core/time.c \
+	core/time.cpp \
 	core/trip.c \
 	core/units.c \
 	core/uemis.c \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -43,7 +43,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/windowtitleupdate.cpp \
 	core/file.c \
 	core/fulltext.cpp \
-	core/subsurfacestartup.c \
+	core/subsurfacestartup.cpp \
 	core/pref.c \
 	core/profile.c \
 	core/device.cpp \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -26,7 +26,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/downloadfromdcthread.cpp \
 	core/qtserialbluetooth.cpp \
 	core/plannernotes.c \
-	core/uemis-downloader.c \
+	core/uemis-downloader.cpp \
 	core/qthelper.cpp \
 	core/checkcloudconnection.cpp \
 	core/color.cpp \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -61,7 +61,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/git-access.cpp \
 	core/globals.cpp \
 	core/liquivision.c \
-	core/load-git.c \
+	core/load-git.cpp \
 	core/parse-xml.c \
 	core/parse.c \
 	core/picture.c \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -41,7 +41,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/qt-init.cpp \
 	core/subsurfacesysinfo.cpp \
 	core/windowtitleupdate.cpp \
-	core/file.c \
+	core/file.cpp \
 	core/fulltext.cpp \
 	core/subsurfacestartup.cpp \
 	core/pref.c \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -67,12 +67,12 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/picture.c \
 	core/pictureobj.cpp \
 	core/sample.cpp \
-	core/import-suunto.c \
-	core/import-shearwater.c \
-	core/import-seac.c \
-	core/import-cobalt.c \
-	core/import-divinglog.c \
-	core/import-csv.c \
+	core/import-suunto.cpp \
+	core/import-shearwater.cpp \
+	core/import-seac.cpp \
+	core/import-cobalt.cpp \
+	core/import-divinglog.cpp \
+	core/import-csv.cpp \
 	core/save-html.c \
 	core/statistics.c \
 	core/worldmap-save.c \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -47,7 +47,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/pref.c \
 	core/profile.c \
 	core/device.cpp \
-	core/dive.c \
+	core/dive.cpp \
 	core/divecomputer.c \
 	core/divefilter.cpp \
 	core/event.c \
--- a/Subsurface-mobile.pro
+++ b/Subsurface-mobile.pro
@@ -60,7 +60,7 @@ SOURCES += subsurface-mobile-main.cpp \
 	core/gaspressures.c \
 	core/git-access.cpp \
 	core/globals.cpp \
-	core/liquivision.c \
+	core/liquivision.cpp \
 	core/load-git.cpp \
 	core/parse-xml.cpp \
 	core/parse.cpp \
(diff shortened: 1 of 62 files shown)


bstoeger:
Ok, just a small update: avoid string copying in the git parser. Once the core structures are `std::string`s, there will be no copying at all (unless of course there is additional formatting or small string optimization).

This is the last new "feature" of this PR, from now on only bug fixing.

bstoeger:
Well, I'm glad that the Windows compiler agrees with me that that casting pointers to long in datatrak.c/h is insane. But it is curious that this wasn't a problem as long as the source code was in C.

bstoeger:
Comment on performance. Even though this now does *lots* of useless copies from std::string to c-strings, for me `git_load_dives()` is now *faster* than before. And there are still two obvious ways to improve this:
1) turn core data structures to std::string, which will remove one allocate+copy per data item.
2) Only  generate std::strings when the value is stored (we have a number of unused values): Store `std::string_view`s instead of `std::string` in `parse_one_line()` and only allocate+convert when needed.

For me, this is absolutely the right way to go - the code is shorter, simpler, faster and not even optimized. Though I would appreciate if someone could confirm this on their setup.

I use the following "benchmark":
```
diff --git a/subsurface-desktop-main.cpp b/subsurface-desktop-main.cpp
index 57a4c3eda..1164cb62a 100644
--- a/subsurface-desktop-main.cpp
+++ b/subsurface-desktop-main.cpp
@@ -17,6 +17,7 @@
 #include "core/tag.h"
 #include "desktop-widgets/mainwindow.h"
 #include "core/checkcloudconnection.h"
+#include "core/file.h"
 
 #include <QApplication>
 #include <QLoggingCategory>
@@ -83,6 +84,11 @@ int main(int argc, char **argv)
 	reset_tank_info_table(&tank_info_table);
 	parse_xml_init();
 	taglist_init_global();
+printf("LOAD\n");
+parse_file("/home/bs/.subsurface/cloudstorage/342eb002b5ae09d9[git]", &divelog);
+printf("LOAD DONE\n");
+exit(0);
+
 	init_ui();
 	if (no_filenames) {
 		if (prefs.default_file_behavior == LOCAL_DEFAULT_FILE) {
```
where `/home/bs/.subsurface/cloudstorage/342eb002b5ae09d9` is the famous 3000+dives log.

mikeller:
All in all this looks good to me and seems to be a step in the right direction.

One thing I noted is that we probably should investigate the possibilities for the device-specific file parsers (like `import-shearwater.cpp`) to be changed to leverage the parser in libdivecomputer where the file format is similar enough to the device format.

I didn't see much of a difference with the test log from https://github.com/subsurface/large-anonymous-sample-data - most likely because I did something wrong:

```
master:
mikeller@wsl:~/git$ ./install-root/bin/subsurface | ts -i %.T
QStandardPaths: wrong permissions on runtime directory /run/user/1000/, 0755 instead of 0700
00:00:01.780989 LOAD
00:00:00.000140 LOAD DONE
mikeller@wsl:~/git$ ./install-root/bin/subsurface | ts -i %.T
QStandardPaths: wrong permissions on runtime directory /run/user/1000/, 0755 instead of 0700
00:00:01.841901 LOAD
00:00:00.000128 LOAD DONE
mikeller@wsl:~/git$ ./install-root/bin/subsurface | ts -i %.T
QStandardPaths: wrong permissions on runtime directory /run/user/1000/, 0755 instead of 0700
00:00:01.793641 LOAD
00:00:00.000103 LOAD DONE
```

vs.

```
bstoeger-cpp:
mikeller@wsl:~/git$ ./install-root/bin/subsurface | ts -i %.T
QStandardPaths: wrong permissions on runtime directory /run/user/1000/, 0755 instead of 0700
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
00:00:01.722357 LOAD
00:00:00.000170 LOAD DONE
mikeller@wsl:~/git$ ./install-root/bin/subsurface | ts -i %.T
QStandardPaths: wrong permissions on runtime directory /run/user/1000/, 0755 instead of 0700
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
00:00:01.705856 LOAD
00:00:00.000102 LOAD DONE
mikeller@wsl:~/git$ ./install-root/bin/subsurface | ts -i %.T
QStandardPaths: wrong permissions on runtime directory /run/user/1000/, 0755 instead of 0700
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
ERROR: Unknown filter preset constraint key/value pair (negate data="/)
00:00:01.736821 LOAD
00:00:00.000108 LOAD DONE
```

mikeller (approved):

bstoeger:
> most likely because I did something wrong:

No - your computer is too fast. :) Point is - it's not getting slower even though there are now lots of unneeded std::string to C-string copies, which will be gone in the future.

bstoeger:
Another data point: loading the large git repository.
master:
```
==930444== LEAK SUMMARY:
==930444==    definitely lost: 4,489,759 bytes in 4,287 blocks
==930444==    indirectly lost: 418 bytes in 9 blocks
==930444==      possibly lost: 80,198 bytes in 121 blocks
==930444==    still reachable: 3,141,894 bytes in 40,529 blocks
==930444==         suppressed: 0 bytes in 0 blocks
```
This PR:
```
==925682== LEAK SUMMARY:
==925682==    definitely lost: 73,960 bytes in 5 blocks
==925682==    indirectly lost: 418 bytes in 9 blocks
==925682==      possibly lost: 71,856 bytes in 110 blocks
==925682==    still reachable: 1,831,382 bytes in 20,458 blocks
==925682==                       of which reachable via heuristic:
==925682==                         newarray           : 2,432 bytes in 16 blocks
==925682==         suppressed: 0 bytes in 0 blocks
```

That's a massive reduction in memory leaks, just by a stupid manual replacement of C strings by C++ strings. And the resulting code is 5% faster.

There seems to be no technical arguments for staying with C data structures.

mikeller:
@bstoeger: I did a quick review of your changes and tested Subsurface with them with no problems. Let me know when you are done with the changes and I can give them another review.

bstoeger:
Thanks for looking into this. I don't plan any more changes for this PR.

It is of course dangerous, because a bug in the parsers/writers can cause data loss. Not sure how we can test this beyond the already existing tests.

dirkhh:
yeah, these are the type of changes where our "new" release setup makes me slightly nervous as something that gets merged into master will potentially go to a fair number of users fairly quickly. I need to run statistics on the cloud servers again, but when I looked a few weeks after the change (and admittedly, at that point the "current" release idea wasn't as fully formed and we were pointing everyone at "latest"), and was surprised just how many people were already using the bleeding edge versions.

If you have concerns about data loss then we should certainly mark that fairly clearly in the title of the Pull request (and the title of this one isn't all that useful for end users, so editing that before merging might be a good idea). That is basically the one tool we have to communicate concerns with the stability / data loss risk to the users who might update to 'latest'.

bstoeger:
I'm not sure this new automatic system is a good idea. I feel that PRs don't have the proper granularity for changelog entries. Some are huge projects (e.g. #3863) others are trivial one-liner code cleanups.

Not sure how you want to proceed here. Feel free to change the title to something that you deem appropriate.

mikeller (approved):

mikeller:
I think we are over-interpreting the 'changelog' in https://www.subsurface-divelog.org/latest-release/ here.
https://www.subsurface-divelog.org/latest-release/ is nothing more than the previous 'nightly builds' directory that was available on the old web server. The only differences are:
- we have shifted building of the artifacts into GitHub, in order to make it easier to have maintenance for this done as a collaborative effort;
- it is a bit easier to find and use these artifacts now - this is in the hope of getting a reasonable number of more technical minded / adventurous users to test them, and provide us feedback if there are problems.
When I suggested that we add a 'changelog' to https://www.subsurface-divelog.org/latest-release/ I did not intend this to be a comprehensive listing of all changes - technical minded users will still have to go and look at the merged pull requests to get this. It was more an effort to make it easier for less technical minded users who have reported a bug to see when a fix for 'their' bug has been merged, so that they can go and test and confirm the fix without a developer having to point them to the nightly build.

@dirkhh: Not sure how much effect more explicit communication on a per-merge basis will have. We are already clear in https://www.subsurface-divelog.org/latest-release/ that this is code that has not had much testing, and in my experience people are either in for a drink from the firehose (and will accept more risky changes), or they are not. I think as an improvement we could add links to the pull requests to the changelog entries, in order to make it easier for users to have a look at the changes.

We could probably look into a UI improvement for the in-app release notifications (once we have implemented the selection of release track to follow) that allows users to choose to 'skip this one' when they get a notification for a new release - but what we don't want (in my opinion) is for users to move from latest to weekly track for these high risk changes, as this will simply mean that we move the risk from latest to weekly as tester coverage in latest will drop.

But anyhow, I think it will be good to promote the release before this one (6.0.5115) as the new weekly release?

PR #3840, opened 2023-02-28
Desktop: Fix use of bluetooth address for remembered dive computer.

mikeller:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Fix a bug causing the bluetooth address not being used when downloading from a 'remembered' dive computer if the device selection is populated. This specifically excludes MacOS, as 'remembering' bluetooth connections does not seem to be working there as per
https://github.com/subsurface/subsurface/pull/2158#issuecomment-508933672. I am not super sure why we are _not_ trying to use the 'remembered' device if the device selection is populated (`ui.device->currentIndex() == -1`) - maybe this should be clarified in a comment?

### Changes made:

1) add condition to check about using the remembered device to do this if it is a bluetooth device and we're not on MacOS

### Related issues:

#2158
Fixes #2139

### Additional information:

Tested on linux.

### Documentation change:

desktop: fix remembering of bluetooth address for remembered dive computers (not MacOS)

### Mentions:

Signed-off-by: Michael Keller <github@ike.ch>

PR #2158, opened 2019-07-04
Desktop: update BT selection when picking remembered dive computer

dirkhh:
When switching to a BT dive computer, the device selection dialog is opened,
when switching away from BT, the device address is set.

Fixes #2139

Signed-off-by: Dirk Hohndel <dirk@hohndel.org>

### Describe the pull request:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Mentions:

@mturkia

PR #4120, opened 2024-03-05
Avoid unnecessary bluetooth device address scanning. 

notrege:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When switching from a non-bluetooth computer to a bluetooh computer an unnecessary bluetooth scan may be forced.   This change will avoid the scan if the bluetooth device address is known.

I discovered this issue when sharing a laptop to upload dives. If I upload dives using a bluetooth computer I have to rescan the device address after a dive is uploaded with a different computer. This change uses the existing device address instead of scanning. A scan can be forces by clearing/rechecking the "Choose Bluetooth download mode" checkbox.

### Changes made:
There are two changes. First, set the device address using the prefs value when a dive computer is selected. Second, avoid scanning if the device address is already known.

### DIFF
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -642,7 +642,7 @@ void DownloadFromDCWidget::enableBluetoothMode(int state)
 
 	if (state == Qt::Checked) {
 		if (ui.device->currentText().isEmpty()) 
-         		selectRemoteBluetoothDevice();
+			selectRemoteBluetoothDevice();
 	} else {
 		ui.device->setCurrentIndex(-1);
 	}
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -152,12 +152,10 @@ void DownloadFromDCWidget::DC##num##Clicked() \
 	productModel.setStringList(productList[qPrefDiveComputer::vendor##num()]); \
 	ui.product->setCurrentIndex(ui.product->findText(qPrefDiveComputer::product##num())); \
 	bool isBluetoothDevice = isBluetoothAddress(qPrefDiveComputer::device##num()); \
-
 	/* If we have a Bluetooth device, set the ui.device->CurrentText to avoid starting \
    	a Bluetooth scan in enableBluetoothMode(). \
    	Note: enableBlueToothMode() will set ui.device.CurrentIndex to -1 unless \
     	ui.bluetoothMode is checked. */ \
-
 	if (isBluetoothDevice) \
 		ui.device->setCurrentText(qPrefDiveComputer::device##num()); \
 	ui.bluetoothMode->setChecked(isBluetoothDevice); \
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -641,7 +641,7 @@ void DownloadFromDCWidget::enableBluetoothMode(int state)
 	ui.chooseBluetoothDevice->setEnabled(state == Qt::Checked);
 
 	if (state == Qt::Checked) {
-		if (ui.device->currentText().isEmpty()) 
+		if (!isBluetoothAddress(ui.device->currentText()))
 			selectRemoteBluetoothDevice();
 	} else {
 		ui.device->setCurrentIndex(-1);
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -152,24 +152,17 @@ void DownloadFromDCWidget::DC##num##Clicked() \
 	productModel.setStringList(productList[qPrefDiveComputer::vendor##num()]); \
 	ui.product->setCurrentIndex(ui.product->findText(qPrefDiveComputer::product##num())); \
 	bool isBluetoothDevice = isBluetoothAddress(qPrefDiveComputer::device##num()); \
-	bool isMacOs = QSysInfo::kernelType() == "darwin"; \
-	/* If we have a Bluetooth address, use it instead of rescanning. For \
-	MacOs, save the mount point.  This may be the wrong thing to do, but for \
-	the Catalina 10.15.7 laptop and Oceanic Atom 3.1 I use, the mountpoint \
-	is /dev/tty.usbserial-20030001 and it doesn't change regardless of the \
-	USB port being used.   If it isn't saved, it must be reentered. */ \
-	if (isBluetoothDevice || isMacOs) \
+
+	/* If we have a Bluetooth device, set the ui.device->CurrentText to avoid starting \
+   	a Bluetooth scan in enableBluetoothMode(). \
+   	Note: enableBlueToothMode() will set ui.device.CurrentIndex to -1 unless \
+    	ui.bluetoothMode is checked. */ \
+
+	if (isBluetoothDevice) \
 		ui.device->setCurrentText(qPrefDiveComputer::device##num()); \
 	ui.bluetoothMode->setChecked(isBluetoothDevice); \
 	if (ui.device->currentIndex() == -1 || isBluetoothDevice) \
 		ui.device->setCurrentIndex(deviceIndex(qPrefDiveComputer::device##num())); \
-	if (isMacOs) { \
-		/* it makes no sense that this would be needed on macOS but not Linux */ \
-		QCoreApplication::processEvents(); \
-		ui.vendor->update(); \
-		ui.product->update(); \
-		ui.device->update(); \
-	} \
 }
 #else
 #define DCBUTTON(num) \
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -641,11 +641,10 @@ void DownloadFromDCWidget::enableBluetoothMode(int state)
 	ui.chooseBluetoothDevice->setEnabled(state == Qt::Checked);
 
 	if (state == Qt::Checked) {
-      		if (ui.device->currentText().isEmpty()) {
+		if (ui.device->currentText().isEmpty()) 
          		selectRemoteBluetoothDevice();
-   		} else {
-      			ui.device->setCurrentIndex(-1);
-   		}
+	} else {
+		ui.device->setCurrentIndex(-1);
 	}
 }
 #endif
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -629,15 +629,14 @@ void DownloadFromDCWidget::enableBluetoothMode(int state)
 	ui.chooseBluetoothDevice->setEnabled(state == Qt::Checked);
 
 	/*	This is convoluted enough to warrant explanation:
- 		1. If we have a Bluetooth device, but no Bluetooth address then scan.
-		2. If we have a Bluetooth device and address, skip scan.
-		3. If it's not Bluetooth, clear the device address unless it's a Mac.
-		Mac USB addresses tend to be persistent and are better saved. */
+ 		1. If Bluetooth is enabled, but no Bluetooth address then scan.
+		2. If Bluetooth is enabled and we have a Bluetooth address, skip scan.
+		3. If Bluetooth is not enabled, but it's a Bluetooth address, clear it. */
 	if (state == Qt::Checked) {
 		if (!isBluetoothAddress(ui.device->currentText()))
 			selectRemoteBluetoothDevice();
 	} else
-		if (isBluetoothAddress(ui.device->currentText()) || QSysInfo::kernelType() != "darwin")
+		if (isBluetoothAddress(ui.device->currentText()))
 			ui.device->setCurrentIndex(-1);
 }
 #endif
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -151,16 +151,8 @@ void DownloadFromDCWidget::DC##num##Clicked() \
 	ui.vendor->setCurrentIndex(ui.vendor->findText(qPrefDiveComputer::vendor##num())); \
 	productModel.setStringList(productList[qPrefDiveComputer::vendor##num()]); \
 	ui.product->setCurrentIndex(ui.product->findText(qPrefDiveComputer::product##num())); \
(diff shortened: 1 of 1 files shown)


desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -149,6 +149,7 @@ int DownloadFromDCWidget::deviceIndex(QString deviceText)
 void DownloadFromDCWidget::DC##num##Clicked() \
 { \
 	ui.vendor->setCurrentIndex(ui.vendor->findText(qPrefDiveComputer::vendor##num())); \
+	ui.device->setCurrentText(qPrefDiveComputer::device##num()); \
```
mikeller: This seems to be doing the same as what we are already doing in https://github.com/subsurface/subsurface/pull/4120/files#diff-a0bde534c60cabf020b85f0079e0a8d567f5465514e8a59629b3a50a228d17f9R160 - but with slightly different connotations, like not overriding the selection for serial devices (which probably makes sense), and not doing it for MacOS where it does not seem to be working.

The core of the problem seems to be that https://github.com/subsurface/subsurface/pull/4120/files#diff-a0bde534c60cabf020b85f0079e0a8d567f5465514e8a59629b3a50a228d17f9L156 is triggering `enableBluetoothMode()`, which is springing a bluetooth selection dialogue - so I think it will be better if we just move this line down below `ui.device->setCurrentIndex()`.

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -639,10 +640,13 @@ void DownloadFromDCWidget::enableBluetoothMode(int state)
...
 	ui.chooseBluetoothDevice->setEnabled(state == Qt::Checked);
 
-	if (state == Qt::Checked)
-		selectRemoteBluetoothDevice();
-	else
-		ui.device->setCurrentIndex(-1);
+	if (state == Qt::Checked) {
+      		if (ui.device->currentText().isEmpty()) {
+         		selectRemoteBluetoothDevice();
+   		} else {
```
mikeller: You've moved the `else` clause from the outer `if` to the (new) inner `if` - which is not the way this should be working.

Also note that [coding style in Subsurface](https://github.com/subsurface/subsurface/blob/master/CODINGSTYLE.md) calls for tab as tabs (and not spaces), and for `if` / `else` with single statements to be without curly braces.

mikeller (commented):
Congratulations on your first pull request!

A couple of remarks that need attention.

desktop-widgets/downloadfromdivecomputer.cpp
notrege: Thanks very much for the help and for looking at this.   I'm not sure why it should be different for serial vs Bluetooth devices, but I must be missing something.    In my test case, I have an Oceanic atom using /dev/tty.usbserial-20030001 and a G2 using a Bluetooth address.   With the changes above, I seem to be able to switch back and forth between the two without a Bluetooth scan or losing the serial mount point.    I'm doing on this on a sadly old Macbook running Catalina (10.15.7).    Won't your suggestion of moving this line down below ui.device-setCurrentIndex() be skipped on MacOs?

desktop-widgets/downloadfromdivecomputer.cpp
notrege: Sorry about that.    I'm afraid that happened when retyping the code in git.   

FWIW, the coding style does not state that single statements go without curly braces - the only relevant point I saw was:

o both sides of an if / else clause either use or do not use curly braces:

desktop-widgets/downloadfromdivecomputer.cpp
mikeller: The thinking behind not using persisted serial port data in #3840 is that on today's computers practically all serial devices are connected through USB, and on most operating systems have dynamic identifiers (based on the order they were detected in). So in most cases the serial port identifier may or may not have changed. Not sure how this is for MacOS, but the device identifier that you supplied looks a lot like it is dependent on the USB bus ID, which will change if a different port is used.
Re not using remembered bluetooth ids on MacOS, this is based on https://github.com/subsurface/subsurface/pull/2158#issuecomment-508933672, so you'll have to work out with @dirkhh if this is still relevant.

desktop-widgets/downloadfromdivecomputer.cpp
mikeller: All good.
In essence, the purpose of having a defined coding style in a project and adhering to it is to avoid unneeded 'noise' from people reworking the style of sections of code when they are reworking or extending it. As such the 'coding style' is much more 'the style used for a majority of the code in a project' than it is a document that tries to codify this style. And in subsurface the predominant style is to not have curly braces for single `if` / `else` statements.

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -639,10 +640,12 @@ void DownloadFromDCWidget::enableBluetoothMode(int state)
 {
 	ui.chooseBluetoothDevice->setEnabled(state == Qt::Checked);
 
-	if (state == Qt::Checked)
-		selectRemoteBluetoothDevice();
-	else
+	if (state == Qt::Checked) {
+		if (ui.device->currentText().isEmpty()) 
+         		selectRemoteBluetoothDevice();
```
mikeller: One more case of spaces - it often pays to enable 'show whitespace' in your editor before submitting code changes.

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -639,10 +640,12 @@ void DownloadFromDCWidget::enableBluetoothMode(int state)
 {
 	ui.chooseBluetoothDevice->setEnabled(state == Qt::Checked);
 
-	if (state == Qt::Checked)
-		selectRemoteBluetoothDevice();
-	else
+	if (state == Qt::Checked) {
+		if (ui.device->currentText().isEmpty())
```
mikeller: I think this will work in an undesirable way if a serial port is selected for the device, and the user then checks the bluetooth box - in this case the serial port device selection will be retained. I think a better way to do this will be to use `isBluetoothAddress()` here to check if the device is a bluetooth device.

notrege: Thanks - that's better than just checking for a device id.

desktop-widgets/downloadfromdivecomputer.cpp
mikeller: @dirkhh: Do you have any insights about https://github.com/subsurface/subsurface/pull/2158#issuecomment-508933672 still being valid, or should this exclusion be dropped?

dirkhh: From comments that others have made, and from a quick experiment, it seems like Apple may have fixed that.

notrege:
I can only speak anecdotally - the change works fine on my relatively old
laptop running Catalina.

The comment struck me as odd given that I can upload without scanning as
long as I don't change the computer.  It wasn't an issue until my wife
started sharing my laptop on dive trips.   If I change to a non-bluetooth
and then back to my G2 I was forced to rescan.   I can only guess that I
misunderstood the comment.

thank you both for looking at this,
Jim

desktop-widgets/downloadfromdivecomputer.cpp
mikeller: Sounds good @dirkhh, thanks for checking.
@notrege we still have the problem that this line essentially doubles up what https://github.com/subsurface/subsurface/pull/4120/files#diff-a0bde534c60cabf020b85f0079e0a8d567f5465514e8a59629b3a50a228d17f9R160 already does, and that it overwrites the serial port if it has already been selected, even if the 'remembered' serial port is not currently available.
So I'd probably remove this line, and instead drop the `!isMacOs` (and the comment about it) in the code below, and then move `ui.bluetoothMode->setChecked(isBluetoothDevice)` to below `ui.device->setCurrentIndex(deviceIndex(qPrefDiveComputer::device##num()))`.

notrege:
Thanks - am diving this week (and enjoying not rescanning) so I'll resume
with this when I'm back.

thanks for the help with this,
Jim

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -153,9 +153,15 @@ void DownloadFromDCWidget::DC##num##Clicked() \
 	ui.product->setCurrentIndex(ui.product->findText(qPrefDiveComputer::product##num())); \
 	bool isBluetoothDevice = isBluetoothAddress(qPrefDiveComputer::device##num()); \
 	bool isMacOs = QSysInfo::kernelType() == "darwin"; \
+	/* If we have a Bluetooth address, use it instead of rescanning. For \
+	MacOs, save the mount point.  This may be the wrong thing to do, but for \
+	the Catalina 10.15.7 laptop and Oceanic Atom 3.1 I use, the mountpoint \
+	is /dev/tty.usbserial-20030001 and it doesn't change regardless of the \
+	USB port being used.   If it isn't saved, it must be reentered. */ \
+	if (isBluetoothDevice || isMacOs) \
+		ui.device->setCurrentText(qPrefDiveComputer::device##num()); \
```
mikeller: I think it will be better to set the device by index, like how it is done in https://github.com/subsurface/subsurface/pull/4120/files#diff-a0bde534c60cabf020b85f0079e0a8d567f5465514e8a59629b3a50a228d17f9L159 - this way, if the remembered device does not currently exist (i.e. the dive computer is not present / detected) the device will be cleared.

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -153,9 +153,15 @@ void DownloadFromDCWidget::DC##num##Clicked() \
...
 	ui.bluetoothMode->setChecked(isBluetoothDevice); \
-	if (ui.device->currentIndex() == -1 || (isBluetoothDevice && !isMacOs)) \
-		/* macOS seems to have a problem connecting to remembered bluetooth devices  if it hasn't already had a connection in the current session */ \
+	if (ui.device->currentIndex() == -1 || isBluetoothDevice) \
```
mikeller: Not sure we still need this - the device is already set before https://github.com/subsurface/subsurface/pull/4120/files#diff-a0bde534c60cabf020b85f0079e0a8d567f5465514e8a59629b3a50a228d17f9L159.

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -153,9 +153,15 @@ void DownloadFromDCWidget::DC##num##Clicked() \
...
 		ui.device->setCurrentIndex(deviceIndex(qPrefDiveComputer::device##num())); \
 	if (isMacOs) { \
```
mikeller: Not really required for this, but since you're already testing this with MacOS: Do you want to test if this extra code is still needed for MacOS, or if we can get a simplification by removing this - as stated, this is working fine without the extra code on other platforms.

mikeller (commented):
This is starting to look good now!

desktop-widgets/downloadfromdivecomputer.cpp
notrege: The change I made in enableBluetoothMode() depends on the value of ui.device->currentText().    Consequently, it must be set before enableBluetoothMode is called (via ui.bluetoothMode->setChecked().    If bluetoothMode is not checked, enableBluetoothMode sets the device index to -1.   Consequently, either the device index must be set both before and after enableBluetoothMode(), or we need a different flag, such as setting CurrentText, for enableBluetoothMode to skip bluetooth scanning.

desktop-widgets/downloadfromdivecomputer.cpp
notrege: I removed this and don't see any adverse effects.    Given that I don't have a Windows or Linux platform to test on, it would be very helpful if someone could make sure I didn't break anything on Linux or Windows.

mikeller: Happy to test this, but since this code is not run (and potentially even optimised away at build time) on non-MacOS I am fairly confident that removing it won't break anything there.

desktop-widgets/downloadfromdivecomputer.cpp
mikeller: > The change I made in enableBluetoothMode() depends on the value of ui.device->currentText(). Consequently, it must be set before enableBluetoothMode is called (via ui.bluetoothMode->setChecked().

Fully agreed on this part - my point is just that it is probably a better way to set this indirectly, through `ui.device->setCurrentIndex(deviceIndex(qPrefDiveComputer::device##num()))` (might also need a `QCoreApplication::processEvents()` to get `ui.device->currentText()` to update).

> If bluetoothMode is not checked, enableBluetoothMode sets the device index to -1.

Good point, I had not considered the case of remembering the non-bluetooth devices on MacOS.
I suspect that the right thing there is to add another refinement to `enableBluetoothMode()`, to only set the index to -1 on disabling bluetooth if the selected device is a bluetooth device - equivalent to what you are doing when enabling bluetooth.

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -639,10 +628,17 @@ void DownloadFromDCWidget::enableBluetoothMode(int state)
...
+	/*	This is convoluted enough to warrant explanation:
+ 		1. If we have a Bluetooth device, but no Bluetooth address then scan.
+		2. If we have a Bluetooth device and address, skip scan.
+		3. If it's not Bluetooth, clear the device address unless it's a Mac.
+		Mac USB addresses tend to be persistent and are better saved. */
+	if (state == Qt::Checked) {
+		if (!isBluetoothAddress(ui.device->currentText()))
+			selectRemoteBluetoothDevice();
+	} else
+		if (isBluetoothAddress(ui.device->currentText()) || QSysInfo::kernelType() != "darwin")
```
mikeller: I did some testing on linux with this, and this is now removing the remembered filesystem path for Garmin devices - so I think it will be better if we remove the ` || QSysInfo::kernelType() != "darwin"` - the fact that we are loading the device from a list of the detected devices with `ui.device->setCurrentIndex(deviceIndex(qPrefDiveComputer::device##num()))` will make sure that we are not setting a non-existent device.

notrege: Thanks for testing this - removed Mac specific behavior and any non-Bluetooth mount points will be preserved.

mikeller (approved):

notrege:
Woo hoo!   Thanks for all the help with this Michael.

mikeller:
Thanks for sticking with this @notrege - the result looks very clean and is a reduction in complexity, which is always welcome.
Tested in linux.

PR #4121, opened 2024-03-06
core: fix memory leak in tables code

bstoeger:
The function clear_*_table frees all elements of the table. However, persumably as a performance feature, it kept the memory of the table itselt (i.e. it only reset the number of elements but kept the capacity).

That is fine if the table is reused later. However, this function was also used when freeing the table and this would leak the table memory.

This commit frees the table memory. An alternative would be to have separate clear_*_table and free_*_table functions. But let's wait with that until we port the table code to C++. Then this will be "automatically" fixed.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a leak in the table code.

This again shows that we can't get C memory management right. In this case I am the culprit.

### DIFF
--- a/core/table.h
+++ b/core/table.h
@@ -96,6 +96,9 @@
 	{									\
 		for (int i = 0; i < table->nr; i++)				\
 			free_##item_name(table->array_name[i]);			\
+		free(table->array_name);					\
+		table->array_name = NULL;					\
+		table->allocated = 0;						\
 		table->nr = 0;							\
 	}
 


mikeller (approved):
Looks good to me - this ties in with how `MAKE_GROW_TABLE` will be reallocating space on the first insertion.

dirkhh (approved):
seems reasonable

PR #4122, opened 2024-03-09
cleanup: fix slot/signal code in preferences code

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Replace SIGNAL/SLOT macros by function pointers, which does compile-time checking. SIGNAL/SLOT should be removed/banned.

This fixes one signal, but it is unclear whether it also fixes a bug, since I didn't notice any change in the UI. At least it doesn't spam an "unknown signal" message to the console.

### Changes made:

1) Fix a signal
2) Replace macro by local function

### DIFF
--- a/desktop-widgets/preferences/preferences_cloud.cpp
+++ b/desktop-widgets/preferences/preferences_cloud.cpp
@@ -96,7 +96,7 @@ void PreferencesCloud::syncSettings()
 				return;
 			}
 			CloudStorageAuthenticate *cloudAuth = new CloudStorageAuthenticate(this);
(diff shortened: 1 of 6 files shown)


mikeller (approved):

dirkhh (approved):
Both look great.
And I agree with replacing the SIGNAL/SLOT garbage.
I know that a couple of years ago I made a pretty good go at this, but clearly I gave up before I was done.

PR #4123, opened 2024-03-09
preferences: fix language selector on recent Qt

bstoeger:
On initialization, the old code searched for the first language code containing a '-'. However, my Qt version gives de-Latn-DE as the first entry. That messed up the preferences code: it didn't recognize that entry. Thus, simply opening and closing the preferences switched the language to Bulgarian.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

See commit description: The language selection was messed up for me. Saving the preferences would switch the language to Bulgarian. Not a nice behavior.

The language selection code is very obscure to me. Not sure if this is the "right" fix, but it works for me.
### Changes made:

1) Select the first language code with a single '-' from the list provided by Qt.

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -462,16 +462,12 @@ void initUiLanguage()
 		loc = QLocale(QLocale().uiLanguages().first());
 	}
 
+	// Find language code with one '-', or use the first entry.
 	QStringList languages = loc.uiLanguages();
 	QString uiLang;
-	if (languages[0].contains('-'))
-		uiLang = languages[0];
-	else if (languages.count() > 1 && languages[1].contains('-'))
-		uiLang = languages[1];
-	else if (languages.count() > 2 && languages[2].contains('-'))
-		uiLang = languages[2];
-	else
-		uiLang = languages[0];
+	auto it = std::find_if(languages.begin(), languages.end(), [](const QString &s)
+				{ return s.count('-') == 1; });
+	uiLang = it == languages.end() ? languages[0] : *it;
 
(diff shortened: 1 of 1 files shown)


mikeller (approved):

dirkhh (approved):
Is this with Qt6?
Or with Qt5?
We have the one weird report that Subsurface-mobile opened in Korean for someone... might this be related???
You are using the beautiful stdC++ constructs that to me always are... challenging - but I think I can guess what this one does.
What I don't understand is how this fixes your issue... I assume you get more than three languages and there's one further down the list that has just one '-'? Otherwise I think that if the old code failed, the new one might as well.

Long story short, I think this makes sense.

bstoeger:
> Is this with Qt6? Or with Qt5? We have the one weird report that Subsurface-mobile opened in Korean for someone... might this be related???

I think with Qt6. And it is probably unrelated to the report. It chooses Bulgarian for me, because that is the first in the dropdown.

>You are using the beautiful stdC++ constructs that to me always are... challenging - but I think I can guess what this one does.

I keep it light on "modern" constructs. But these are quite readable, I think: Search for the first entry that has exactly one '-'. If found choose that, otherwise the first in the list.

> What I don't understand is how this fixes your issue... I assume you get more than three languages and there's one further down the list that has just one '-'? Otherwise I think that if the old code failed, the new one might as well.

I get 'de-Latn-DE' as first entry, 'de-DE' as second and 'de' as third. The old code chose the first, which was not recognized by the preferences dialog and therefore left the dropdown uninitialized.

PR #4126, opened 2024-03-10
More conversion of C to C++, mostly in the planner

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Compile more files with C++ to avoid memory management troubles.

Note that one of the files, `tag.c` is wild! In `taglist_cleanup` it simply drops list member on the floor. Ooops. In a comment it says that the `divetag`s are only kept in the global `g_tag_list`. But that is apparently not true, see `copy_tl`. So you never know if the pointer to the `divetag` is owning or not. And therefore it doesn't even attempt to do any memory management. This PR does _not_ fix this file. I plan to do this when the users are all C++.

@atdotde: This PR mostly concerns planner source files. Please check.


dirkhh:
I am so tempted to do a `git blame`... or maybe not.

bstoeger:
The only thing to blame is a language with unmanageable string-handling and memory-management. Even Linus got it wrong in `load-git.h`.

core/planner.cpp
```diff
@@ -450,20 +450,23 @@ static struct gaschanges *analyze_gaslist(struct diveplan *diveplan, struct dive
...
-	if (gnr == 0) {
-		memcpy(stoplevels, dstops, dnr * sizeof(int));
+	if (gstops.empty()) {
+		std::copy(dstops, dstops + dnr, stoplevels.begin());
 		return stoplevels;
 	}
-	i = total - 1;
-	gi = gnr - 1;
-	di = dnr - 1;
+	int i = static_cast<int>(total) - 1;
```
atdotde, line 467: Why the cast and not making i size_t as well?

atdotde (approved):
Looks ok to me (as far as I can see).

core/planner.cpp
bstoeger: Because `size_t` is unsigned and that would make the loop `while (i >= 0)` an infinite loop.

Unsigned integers are evil and `size_t` was, in hindsight, a terrible design decision. Though understandable in the context of 32 bit systems.

bstoeger:
@mikeller: This now contains cloud stuff - would be great if you could test it (also #4137).

bstoeger:
I have no way of testing the conversion of `smartrak.c`. :(

bstoeger:
I'm quite sure that I butchered the SmarTrak import with the last commit. Need Salva's input on this.

dirkhh:
@salvadorcunat can you take a look, please?

salvadorcunat:
I can't build your bstoeger-planner2 branch.  I'm getting a lot of "template with C linkage" error coming from mdbtools.h.  Is this the issue you have @bstoeger ?

bstoeger:
@salvadorcunat: My problem is that I have no idea how to get this built on Linux.

>  I'm getting a lot of "template with C linkage" error coming from mdbtools.h. 

Ouch! That seems to be an mdbtools bug:

It first does
```
#ifdef __cplusplus
  extern "C" {
#endif
```
then 
```
#include <glib.h>
```

But `glib` realizes that this is a C++ source and then includes stuff incompatible with `extern "C"`.

One workaround might be to include `glib.h` first and hope that the "include guards" prevent the inclusion of `glib.h` from `mdbtools.h`.

But we should file a bug report / pull request to mdbtools!

bstoeger:
FYI: opened a PR: https://github.com/mdbtools/mdbtools/pull/435

salvadorcunat:
Sorry, on the phone right now.

El lun., 18 mar. 2024 7:39, bstoeger ***@***.***> escribió:

>  My problem is that I have no idea how to get this built on Linux.
>
subsurface/scripts/smtk2ssrf-build.sh -y   running from src folder should
do the work.

I'll take a look at the glib thing and the rest this afternoon.

Best regards.

>

bstoeger:
> subsurface/scripts/smtk2ssrf-build.sh -y running from src folder should do the work.

Cool - I didn't know that. Probably should be integrated in CI given how many times I managed to break this. :-P

Do we have any test-files?

I tried to apply the workaround of including `glib.h` first. Let's see if that fixes the compilation issues.

mikeller:
> Cool - I didn't know that. Probably should be integrated in CI given how many times I managed to break this. :-P

See #4139.

smtk-import/smartrak.cpp
```diff
@@ -698,49 +670,44 @@ static std::string smtk_locate_buddy(MdbHandle *mdb, char *dive_idx, char *buddi
...
 /*
  * Add data from tables related in Dives table which are not directly supported
  * in Subsurface. Write them as tags or dive notes by setting true or false the
  * boolean parameter "tag".
  */
-static void smtk_parse_other(struct dive *dive, char *list[], const char *data_name, char *idx, bool tag)
+static void smtk_parse_other(struct dive *dive, const std::vector<std::string> &list, const char *data_name, char *idx, bool tag)
 {
        int i = atoi(idx) - 1;
-       char *str = NULL;
```
salvadorcunat, line 718: I would add a check here for i value >= 0.  If a table is read wrongly we can get idx=0 and thus i < 0.

Just 
`
        if (i < 0)
              return;  // idx should be 1 at least
`
Works fine.

salvadorcunat:
As @bstoeger said just moving #include \<mdbtools\> after #include \<vector\> avoids the linkage issues.

BTW, the build crashes for different reasons.  I've seen 2 of them, commented on the patch,  and am a bit stuck with a crash in the libdc_buffer_parser() function which will take longer to debug.

I'll get on it tomorrow.

bstoeger:
Could you post abacktrace? Sadly I still don't get an executable. It fails during linking. Perhaps a Gt6 issue?

bstoeger:
It seems like `./subsurface/scripts/smtk2ssrf-build.sh` doesn't work with Qt6:

```
[...]
CMake Warning at CMakeLists.txt:207 (find_package):
  Found package configuration file:

    /usr/lib/x86_64-linux-gnu/cmake/Qt6/Qt6Config.cmake

  but it set Qt6_FOUND to FALSE so package "Qt6" is considered to be NOT
  FOUND.  Reason given by package:

  Failed to find required Qt component "Location".

  Expected Config file at
  "/usr/lib/x86_64-linux-gnu/cmake/Qt6Location/Qt6LocationConfig.cmake" does
  NOT exist



  Configuring with --debug-find-pkg=Qt6Location might reveal details why the
  package was not found.

  Configuring with -DQT_DEBUG_FIND_PACKAGE=ON will print the values of some
  of the path variables that find_package uses to try and find the package.



CMake Deprecation Warning at CMakeLists.txt:3 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


/usr/bin/ld: /home/bs/src-qt6/subsurface/build/core/libsubsurface_corelib.a(eventtype.cpp.o): in function `event_type_name(QString, event_severity)':
eventtype.cpp:(.text+0xcb5): undefined reference to `QArrayData::deallocate(QArrayData*, long long, long long)'
/usr/bin/ld: eventtype.cpp:(.text+0xdf3): undefined reference to `QString::_empty'
/usr/bin/ld: eventtype.cpp:(.text+0xe03): undefined reference to `QString::_empty'
/usr/bin/ld: /home/bs/src-qt6/subsurface/build/core/libsubsurface_corelib.a(eventtype.cpp.o): in function `event_type_name(event const*)':
eventtype.cpp:(.text+0xe97): undefined reference to `QByteArray::QByteArray(char const*, long long)'
/usr/bin/ld: eventtype.cpp:(.text+0xea9): undefined reference to `QString::fromUtf8(QByteArrayView)'
/usr/bin/ld: eventtype.cpp:(.text+0xecd): undefined reference to `QArrayData::deallocate(QArrayData*, long long, long long)'
/usr/bin/ld: eventtype.cpp:(.text+0xf2b): undefined reference to `QArrayData::deallocate(QArrayData*, long long, long long)'
[...]
```

salvadorcunat:
> Could you post abacktrace? Sadly I still don't get an executable. It fails during linking. Perhaps a Gt6 issue?

Hi @bstoeger I've been KO with a flue for some days.

Here is the back trace:
#0  __strncasecmp_l_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:283
#1  0x00005555555e06d5 in sample_cb (type=<optimized out>, pvalue=<optimized out>, userdata=0x555555b22d88) at /home/boret/src/subsurface/core/libdivecomputer.cpp:437
#2  0x00007ffff7ee74a9 in uwatec_smart_parse () from /home/boret/src/install-root/lib/libdivecomputer.so.0
#3  0x00007ffff7ee76ec in uwatec_smart_parser_samples_foreach () from /home/boret/src/install-root/lib/libdivecomputer.so.0
#4  0x00007ffff7ecea14 in dc_parser_samples_foreach () from /home/boret/src/install-root/lib/libdivecomputer.so.0
#5  0x00005555555e2f4b in libdc_buffer_parser (dive=0x555555b22cd0, data=0x555555b92f70, buffer=<optimized out>, size=3804) at /home/boret/src/subsurface/core/libdivecomputer.cpp:1619
#6  0x00005555555bd944 in smartrak_import (file=0x7fffffffe9e9 "/home/boret/src/subsurface/dives/Demo_SmartTrak.slg", log=0x555555798640 <divelog>)
    at /home/boret/src/subsurface/smtk-import/smartrak.cpp:980
#7  0x00005555555b9c93 in main (argc=3, argv=0x7fffffffe698) at /home/boret/src/subsurface/smtk-import/smtk_standalone.cpp:50

This looks like some kind of memory corruption to me.  I've dumped the buffers passed to libdc_buffer_parser() and look correct.

The only weird thing I can see is 
`
  834                  concat((char *)dev_data->model, "", format_string_std("%s %s", dev_data->vendor, dev_data->product));
`
This concat() is not working, dev_data->model is NULL after the call, while format_string_std() seems to work fine. As dev_data is passed to libdc in libdc_buffer_parser() call, this could be triggering the segfault in libdc.

salvadorcunat:
> It seems like `./subsurface/scripts/smtk2ssrf-build.sh` doesn't work with Qt6:
> 

Well, I've been unable to build Subsurface with Qt6 til now :-). 
Never thought about this, and this wasn't supposed to run on Apple devices as SmartTrak is Windows based software. If I'm not wrong, Subsurface only builds on Qt6 for Macs.
But seems quite simple, will try this afternoon.

salvadorcunat:
> The only weird thing I can see is `834 concat((char *)dev_data->model, "", format_string_std("%s %s", dev_data->vendor, dev_data->product));` This concat() is not working, dev_data->model is NULL after the call, while format_string_std() seems to work fine. As dev_data is passed to libdc in libdc_buffer_parser() call, this could be triggering the segfault in libdc.

In fact, if dev_data->model is set to a previous value (like a "-"), then we get this crash:

#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6,
    no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007ffff5ca81cf in __pthread_kill_internal (signo=6, threadid=<optimized out>)
    at ./nptl/pthread_kill.c:78
#2  0x00007ffff5c5a472 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff5c444b2 in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007ffff5c451ed in __libc_message (fmt=fmt@entry=0x7ffff5db778c "%s\n")
    at ../sysdeps/posix/libc_fatal.c:150
#5  0x00007ffff5cb1ae5 in malloc_printerr (
    str=str@entry=0x7ffff5dba0b8 "munmap_chunk(): invalid pointer") at ./malloc/malloc.c:5658
#6  0x00007ffff5cb1cac in munmap_chunk (p=<optimized out>) at ./malloc/malloc.c:3036
#7  0x00007ffff5cb6228 in __GI___libc_free (mem=<optimized out>) at ./malloc/malloc.c:3357
#8  0x00005555555baacb in concat (orig_in=@0x7fffffffd788: 0x555555af64b0 "-Uwatec Smart Z",
    sep=0x555555733e40 "", s="Uwatec Smart Z")
    at /home/boret/src/subsurface/smtk-import/smartrak.cpp:282
--Type <RET> for more, q to quit, c to continue without paging--
#9  0x00005555555bcff8 in prepare_data (data_model=28, serial=0x555555b858b0 "24100029", dc_fam=DC_FAMILY_NULL, dev_data=0x555555b4fcc0)
    at /home/boret/src/subsurface/smtk-import/smartrak.cpp:835
#10 0x00005555555bd77d in smartrak_import (file=0x7fffffffe9ea "/home/boret/src/subsurface/dives/Demo_SmartTrak.slg", log=0x555555798640 <divelog>)
    at /home/boret/src/subsurface/smtk-import/smartrak.cpp:968

The crash may be expected, as we are trying to free the "-". But line #8 in the back trace shows the joined string "-Uwatec Smart Z" which sould be dev_data->model, so concat() seems to work fine here, as it does everywhere else, but the resulting model is still "NULL" when we pass the resulting dev_data structure.

bstoeger:
Sorry to hear about the flu and thanks for debugging this! Can you try the updated version - if this shows the same problem, I will have to roll this part back until we switch to `std::string` for core structures.

salvadorcunat:
Thanks, @bstoeger .
Crash is gone with latest force push.
Still the issue with 
 663 ▍        return idx >- 0 && static_cast<size_t>(idx) < v.size() ? v[idx] : T();

which goes away with ">=" instead.

There is something going wrong on in the smtk_locate_buddy() thing. It's building a string with the last nickname and a full list of complete names, where it should be "nickname(full name), nicknameII(full nameII)...".

I'll debug it later.

bstoeger:
> 663 ▍ return idx >- 0 && static_cast<size_t>(idx) < v.size() ? v[idx] : T();

How embarrassing...

> I'll debug it later.

Thanks. If I find time I will try to compile under Qt5. Is there a test file? Probably we should add it to the testing infrastructure, so that I don't have to bother you...

salvadorcunat:
Hi @bstoeger, on the phone now.

El vie., 22 mar. 2024 15:44, bstoeger ***@***.***> escribió:

>  Is there a test file?
>
>
There is no test AFAIK but there's a demo file useful for test purposes in
subsurface/dives folder. Not sure about file's name,  but something like
Demo_SmartTrak.slg.

salvadorcunat:
> There is something going wrong on in the smtk_locate_buddy() thing. It's building a string with the last nickname and a full list of complete names, where it should be "nickname(full name), nicknameII(full nameII)...".
> 
> I'll debug it later.

It's just variable "fullname" non being reseted inside the loop in smtk_build_buddies().  Simply add
`
fullname = "";`

At the end of the while loop and it'll be done.

With this fixed, I can't see any other issue here and would be a thumbs up for me @bstoeger

bstoeger:
How embarrassing. Many thanks for debugging this!

bstoeger:
@dirkhh: I was so fed up with the inconsistency of `qDebug()`, `fprintf(stderr,...)` and `report_info()` that I changed mostly everything to the latter. I'm not  fan of the C-style interface, but once we convert everything to compile under C++, we can work on that. At least it is somewhat consistent now. Of course, this may mess with formatting of log-messages. Please let me know if that is a reasonable way to proceed.

Turning into a "draft", because `subsurface-mobile` now crashes with
```
Cannot add multiple registrations for QtQml.Models
```
Will investigate...

I will add no more changes to this, so that I can finalize the QtQuick profilechart.

bstoeger:
> Cannot add multiple registrations for QtQml.Models

Well, the same thing happens with current master. It seems that I messed up all my build environments. Sigh.

dirkhh:
I will admit that massive changes like "oh, let's change the way all output works across all build types and all platforms" scare the crap out of me... So I generally shy away from them. There are so many weird subtle corner cases...

I'm not opposed to it. It's just... A challenge that you've given yourself here

bstoeger:
I'm not so much worried about the debug(!) output. I'm rather worried about the inherently unsafe C-interface, where things like
```
report_info("msg: %s\n", i);
```
will crash if `i` is say an integer. And we only get compiler warnings, which I might not have caught. I wonder if we can turn such formatting warnings into hard compile errors?

Still, I think unifying this whole mess to our own function is the way to go. And we can replace it by a safe variant later.

glance-:
-Wformat along with -Wformat-signedness for getting warnings, and  -Werror=format and -Werror=format-signedness for getting them as errors.

To ensure such errors are found:
void report_info(const char *fmt, ...) __attribute__ ((format (printf, 1, 2)))

bstoeger:
> -Wformat along with -Wformat-signedness for getting warnings

Thanks. Let's do that.

Why on earth doesn't the Windows build recognize `%td` as format specifier for pointer differences. :(

bstoeger:
Any objections to this? I'm losing track of all my different branches...

dirkhh:
You and me, both
I haven't done more than a very cursory test, and I tend not to use the planner, ever...

mikeller:
No objections at all from me.
Happy to give this a review, and do some testing on it, including the planner.

mikeller (approved):
I gave the planner a spin for OC and CCR, and did some tests with cloud synchronisation, and everything seems to be working fine.
Should we get this merged @bstoeger?

bstoeger:
Thanks!

Let's merge and handle the fallout so that we can move on...

mikeller:
@bstoeger: Looks like this is causing crashes in Windows (but not linux) when opening dives imported from a Garmin dive computer: https://groups.google.com/g/subsurface-divelog/c/7Uo7A9YTGhg/m/39JxckLkBQAJ

I was able to replicated, but the application debug output did not have any indication as to the location - are there instructions for attaching a debugger to this in Windows?

bstoeger:
> when opening dives imported from a Garmin dive computer

So not when downloading, but when opening dives? Is there an affected logbook?

> I was able to replicated, but the application debug output did not have any indication as to the location - are there instructions for attaching a debugger to this in Windows?

I neither have windows nor a Garmin computer. What sometimes helps is running it under valgrind on Linux. I will try if I find something with my dive computer tonight.

mikeller:
> > when opening dives imported from a Garmin dive computer
> 
> So not when downloading, but when opening dives? Is there an affected logbook?

Correct, and good point (opening dive 1 on Windows results in a crash):

[garmin_crash_test.ssrf.zip](https://github.com/subsurface/subsurface/files/15172489/garmin_crash_test.ssrf.zip)

> > I was able to replicated, but the application debug output did not have any indication as to the location - are there instructions for attaching a debugger to this in Windows?
> 
> I neither have windows nor a Garmin computer. What sometimes helps is running it under valgrind on Linux. I will try if I find something with my dive computer tonight.

Valgrind does not seem to be currently supporting Windows. :-(

When opening the above log file on Windows I get:

`QObject::connect(QQuickWindow, QDeclarativeGeoMap_QML_6): invalid nullptr parameter`

And poking a bit more, this happens irrespective of any dives in the log having coordinates attached to their location or not.

bstoeger:
> [garmin_crash_test.ssrf.zip](https://github.com/subsurface/subsurface/files/15172489/garmin_crash_test.ssrf.zip)

Thanks.
 
> Valgrind does not seem to be currently supporting Windows. :-(

I meant running valgrind under Linux. Sometimes it reports problems that do not crash on Linux but on other platforms.

In fact, it finds something that I would expect to crash (even on Linux).

Can you try this patch:

```
diff --git a/core/profile.cpp b/core/profile.cpp
index a6e71839c..bfe7ff5d1 100644
--- a/core/profile.cpp
+++ b/core/profile.cpp
@@ -1367,8 +1367,8 @@ static std::vector<std::string> plot_string(const struct dive *d, const struct p
 	std::vector<std::string> res;
 
-	res.push_back(casprintf_loc(translate("gettextFromC", "@: %d:%02d"), FRACTION(entry->sec, 60), depthvalue));
-	res.push_back(casprintf_loc(translate("gettextFromC", "D: %.1f%s"), depth_unit));
+	res.push_back(casprintf_loc(translate("gettextFromC", "@: %d:%02d"), FRACTION(entry->sec, 60)));
+	res.push_back(casprintf_loc(translate("gettextFromC", "D: %.1f%s"), depthvalue, depth_unit));
 	for (cyl = 0; cyl < pi->nr_cylinders; cyl++) {
 		int mbar = get_plot_pressure(pi, idx, cyl);
 		if (!mbar)
```

The compiler doesn't warn in this case, because it doesn't understand that translate() is (should be) a noop in this case. :(

> `QObject::connect(QQuickWindow, QDeclarativeGeoMap_QML_6): invalid nullptr parameter`

I get that on Linux too.

> And poking a bit more, this happens irrespective of any dives in the log having coordinates attached to their location or not.

Weird.

PR #4128, opened 2024-03-10
Improve automated language handling

dirkhh:
This mimics the code added in commit cf990b0f39 ("preferences: choose language code with one '-'") and adds some debugging for the mobile case - some people are being presented with Subsurface-mobile in Korean for some reason.

- [x] Bug fix (?)

Initially I just wanted to add some debug code to figure out while some users on iOS suddenly see Korean as default language - but in the process I noticed that there was essentially the same code block that Berthold had already replaced a second time just a few lines below. So I tried to update that as well - and added the debug output that I was looking for.

### Mentions:

@bstoeger - I think I adapted this correctly, but always appreciate your keen eye on this.

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -466,18 +466,22 @@ void initUiLanguage()
 	auto it = std::find_if(languages.begin(), languages.end(), [](const QString &s)
 				{ return s.count('-') == 1; });
 	uiLang = it == languages.end() ? languages[0] : *it;
+#ifdef SUBSURFACE_MOBILE
+	qDebug() << "uiLanguages was" << languages << ", picked" << uiLang;
+#endif
 
 	// there's a stupid Qt bug on MacOS where uiLanguages doesn't give us the country info
 	if (!uiLang.contains('-') && uiLang != loc.bcp47Name()) {
 		QLocale loc2(loc.bcp47Name());
 		loc = loc2;
 		QStringList languages = loc2.uiLanguages();
-		if (languages[0].contains('-'))
-			uiLang = languages[0];
-		else if (languages.count() > 1 && languages[1].contains('-'))
-			uiLang = languages[1];
-		else if (languages.count() > 2 && languages[2].contains('-'))
-			uiLang = languages[2];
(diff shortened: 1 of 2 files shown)


mikeller (commented):
Looks correct, but I am wondering if there is a way to refactor this so that the code to select the language is not duplicated, to avoid having to do fixes in two almost identical places in the future?

dirkhh:
Well, the two places are somewhat different. The first one looks for exactly one dash, the second one looks for at least one dash.

That said, the whole code was written by someone who didn't really (and still doesn't) understand how this is supposed to work (me), so... any refactoring is of course welcome.

For now I want to understand why a couple of iOS users suddenly see Subsurface-mobile in Korean. Which isn't perceived as an improvement, given that neither of them knows the language...

mikeller:
> Well, the two places are somewhat different. The first one looks for exactly one dash, the second one looks for at least one dash.

Is this actually needed though, i.e. are there cases on iOS where the first result has got more than one dashes, and is still recognised by iOS as a valid language string?
If not then we can check for exactly one dash in the second case as well.

> For now I want to understand why a couple of iOS users suddenly see Subsurface-mobile in Korean. Which isn't perceived as an improvement, given that neither of them knows the language...

Ha, reminds me of the good old days of VCRs. 😂 Just need to learn the magic key sequence.

dirkhh:
So I wrote this like a hundred years ago - but the genesis was that we would get vastly different language lists between desktop and mobile, and between different OSs. And I was having a really hard time sorting them in a way that the user would indeed end up with their preferred language. I'm hoping that the debug messages added will show us if this is indeed still something that we need.

dirkhh:
Ugh - and while the debug output works great with mobile on desktop, I don't see the output on iOS in the AppLog. So I need to figure out how to show this info later when the log data collection is already set up.
There's always something...

bstoeger (approved):
Yes, I missed that there is a second version of that selection code.

At some point we will have to rethink the language selection code, but it surpasses my mental capacity right now.

PR #4131, opened 2024-03-12
[cleanup] make error reporting more robust

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add compiler warnings for `report_error()` and `report_info()`.

@dirkhh: Is there any point to these `SSRF_INFO` macros? They are just visual clutter to me. If you feel these are necessary, please drop the last commit.

### DIFF
--- a/core/errorhelper.h
+++ b/core/errorhelper.h
@@ -10,7 +10,10 @@ extern "C" {
 
 extern int verbose;
 extern int report_error(const char *fmt, ...);
+extern void report_info(const char *fmt, ...);
 extern void set_error_cb(void(*cb)(char *));	// Callback takes ownership of passed string
+#define SSRF_INFO(fmt, ...) report_info(fmt, ##__VA_ARGS__)
+
 
 #ifdef __cplusplus
 }
--- a/core/errorhelper.h
+++ b/core/errorhelper.h
@@ -8,9 +8,15 @@
 extern "C" {
 #endif
 
+#ifdef __GNUC__
+#define __printf(x, y) __attribute__((__format__(__printf__, x, y)))
+#else
+#define __printf(x, y)
+#endif
+
 extern int verbose;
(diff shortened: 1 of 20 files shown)


dirkhh:
As so often, `SSRF_INFO` has historical reasons. Initially this replaced `fprintf()` statements so that we could also write to the Android logger when building on Android. Of course this evolved over time and at this point you are correct, it's just a silly wrapper, so I'm good with simply dropping it.

core/errorhelper.h
```diff
@@ -8,10 +8,18 @@
...
+#define __printf(x, y) __attribute__((__format__(__printf__, x, y)))
+#else
+#define __printf(x, y)
+#endif
+
 extern int verbose;
-extern int report_error(const char *fmt, ...);
+extern int __printf(1, 2) report_error(const char *fmt, ...);
+extern void __printf(1, 2) report_info(const char *fmt, ...);
 extern void set_error_cb(void(*cb)(char *));	// Callback takes ownership of passed string
```
dirkhh, line 20: so now we have no report_error declaration anymore? My quick search seems to indicate that errorhelper.h is the only place where it was declared in a header.
What am I missing?

dirkhh (commented):
This all looks great and I was almost ready to merge it - and then I stumbled across the removal of the declaration which confused me...

core/errorhelper.h
bstoeger: It's not removed. The `__printf(1,2)` attribute is added so that we get compiler checks.

dirkhh: and this is why you shouldn't do code reviews at 6 in the morning... duh

dirkhh (approved):
-ENOTENOUGHCOFFEE

PR #4133, opened 2024-03-13
Import / Export: Remove Unsafe XML Handling Options.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] Security improvement

Remove the options to expand entities and so continue when encountering invalid / malformed XML, as both of these can be exploited by supplying maliciously crafted XML.

### DIFF
--- a/core/parse-xml.cpp
+++ b/core/parse-xml.cpp
@@ -1765,9 +1765,9 @@ extern "C" int parse_xml_buffer(const char *url, const char *buffer, int, struct
 
 	state.log = log;
 	state.fingerprints = &fingerprint_table; // simply use the global table for now
-	doc = xmlReadMemory(res, strlen(res), url, NULL, XML_PARSE_HUGE | XML_PARSE_RECOVER);
+	doc = xmlReadMemory(res, strlen(res), url, NULL, XML_PARSE_HUGE);
 	if (!doc)
-		doc = xmlReadMemory(res, strlen(res), url, "latin1", XML_PARSE_HUGE | XML_PARSE_RECOVER);
+		doc = xmlReadMemory(res, strlen(res), url, "latin1", XML_PARSE_HUGE);
 
 	if (res != buffer)
 		free((char *)res);
@@ -2346,7 +2346,6 @@ static xmlDoc *test_xslt_transforms(xmlDoc *doc, const struct xml_params *params
 			}
 			xmlFree(attribute);
 		}
-		xmlSubstituteEntitiesDefault(1);
 		xslt = get_stylesheet(info->file);
 		if (xslt == NULL) {
 			report_error(translate("gettextFromC", "Can't open stylesheet %s"), info->file);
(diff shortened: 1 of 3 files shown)


dirkhh (commented):
I wonder why we had these options there in the first place. I never worked on that part of the code, so... 🤷🏼‍♂️

mikeller:
> I wonder why we had these options there in the first place. I never worked on that part of the code, so... 🤷🏼‍♂️

Event parsing was enabled ~ 12 years ago, and back then there probably was a lot less focus on security, and this was considered a desirable feature:

https://github.com/subsurface/subsurface/commit/350462949d2dc205355e5c94ccaacf83a0775257#diff-347aeaa7212923fda1acc2b518bd37d73f34e795717a0a59a8da8df24f4bcc0aR1493

Parsing of invalid / malformed files was added much more recently, and apparently it was intended to provide a workaround for buggy XML generated by Subsurface:

https://github.com/subsurface/subsurface/commit/5eda1c0e39b7c5e4e68ec68596cb023e37503ba7

If there still are files with this bug out there I'd be keen to add a workaround that caters for only this bug, instead of enabling parsing of invalid / malformed (and potentially malicious files).

bstoeger:
I'm fine with this, though malicious XMLs can probably cause problems in other parts of the code as well. So I'm not sure that this is all that important.

That said, ultimately we should probably use a better (hand-written?) XML parser that can handle things such as multiple arguments without having to enable unsafe features.

bstoeger:
Wanted to merge this, but was in the wrong tab and merged #4137 instead. :(

Now I don't dare to merge this, because it might make reverting more difficult.

dirkhh:
Let's figure out what we want to do about 4137 - I took a look and it seemed fine, @glance- said it worked for him, so...
I'd lean towards not making this too complicated and fixing things if 4137 breaks something after all.
I consider the 'latest' builds 'high risk' - and I can certainly take either an older build or wait a few days before making a new 'current' release. Doing those weekly seems way too aggressive, anyway.

bstoeger:
Ok, let's merge this then and sweep the whole episode under the rug.

mikeller:
> That said, ultimately we should probably use a better (hand-written?) XML parser that can handle things such as multiple arguments without having to enable unsafe features.

Not sure I understand that - what are the 'multiple arguments' you have in mind?

bstoeger:
Sorry, I meant multiple XML "attributes", the reason I added the "unsafe" option.

mikeller:
> Sorry, I meant multiple XML "attributes", the reason I added the "unsafe" option.

Ok. But that will just mean that the format produced is not XML any more  but something proprietary - which may be fine for ssrf, but likely not for any of the other XML based formats we are supporting.
And because we are handling a lot of these other formats through XSL transformations from ssrf, which expects XML as input a lot of these will have to be redone in this case as well...

bstoeger:
I'm not suggesting writing a different format. I'm suggesting making the parser more robust with respect to non-standard XML without compromising "safety". But not anytime soon...

Yes, the XSLT thing is a problem. I'll think about it, when the time comes. So not in the near future. :)

PR #4137, opened 2024-03-16
Fix memory management in configure dive computer code

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The configure dive computer code is very ominous. This is a drop in the ocean, but as a start: use value semantics for `DeviceDetails` (though pass it as references through functions).

Contains more cleanups:
- Conversion of some variables to std::string
- Less global variable accesses
- Removal of archaic `Q_FOREACH` and foreach.

Note: I don't have a divecomputer that can be configured, so this is 100% untested!
Note: The `device_data_t` is still shared among threads, which is pretty sketchy. Since libdivecomputer is a C-API, it is unclear what the semantics of this struct are with respect to copying. Will have to investigate at one point.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -10,7 +10,7 @@ ColumnLimit: 0
 ConstructorInitializerAllOnOneLineOrOnePerLine: true
 ConstructorInitializerIndentWidth: 8
 ContinuationIndentWidth: 8
-ForEachMacros: [ 'foreach', 'for_each_dc', 'for_each_relevant_dc', 'for_each_dive', 'for_each_line', 'Q_FOREACH' ]
+ForEachMacros: [ 'for_each_dc', 'for_each_relevant_dc', 'for_each_dive', 'for_each_line' ]
 IndentFunctionDeclarationAfterType: false #personal taste, good for long methods
 IndentWidth: 8
 MaxEmptyLinesToKeep: 2
--- a/core/configuredivecomputer.cpp
+++ b/core/configuredivecomputer.cpp
@@ -574,8 +574,8 @@ QString ConfigureDiveComputer::dc_open(device_data_t *data)
 	FILE *fp = NULL;
 	dc_status_t rc;
 
-	if (data->libdc_log)
-		fp = subsurface_fopen(logfile_name, "w");
+	if (data->libdc_log && !logfile_name.empty())
+		fp = subsurface_fopen(logfile_name.c_str(), "w");
 
 	data->libdc_logfile = fp;
 
--- a/core/configuredivecomputer.cpp
+++ b/core/configuredivecomputer.cpp
@@ -41,7 +41,7 @@ void ConfigureDiveComputer::readSettings(device_data_t *data)
 	readThread->start();
 }
 
-void ConfigureDiveComputer::saveDeviceDetails(DeviceDetails *details, device_data_t *data)
+void ConfigureDiveComputer::saveDeviceDetails(const DeviceDetails &details, device_data_t *data)
 {
 	setState(WRITING);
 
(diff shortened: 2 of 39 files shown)


dirkhh:
the dive computer configuration is mostly @glance- 's code...

mikeller:
I did a quick test with an OSTC4, and I get this error (and no data displayed) when reading the dive computer configuration:

```
QObject::connect: Cannot queue arguments of type 'DeviceDetails'
(Make sure 'DeviceDetails' is registered using qRegisterMetaType().)
```

I also get an error when trying to write the configuration, but I suspect that this is triggered by the configuration data being invalid:

```
[53.980213] ERROR: Invalid parameter specified. [in src/hw_ostc3.c:1048 (hw_ostc3_device_config_write)]
```

bstoeger:
Thanks - I will look into it.

bstoeger:
OMG, I totally forgot about the crazy behavior of the signal/slot system, when it comes to threading:
https://doc.qt.io/qt-6/threads-qobject.html#signals-and-slots-across-threads

I will have to rethink that.

dirkhh:
> I will have to rethink that.

A very common response when I try to reason about threads and Qt...

glance-:
It's really not "mine" but rather Joseph who did it as a Google summer of code project, with Thiago as mentor, so I've just assumed that all the Qt-thingies are as they should be. It might be dated, it's 10 years old code.

I've just added more features and computers.

I took a quick glance, and most of it just looked to be pointer vs. reference. I need to look closer to evaluate the other things. 

I think the OSTC2 memory bank debug code is still there, which can be used to test everything without a supported computer. Not sure what's needed to use it on top of my head.

If wanted I can take a closer look and test things later this week. All my diving stuff is tucked away due to everything being still being frozen here for a couple of months.

bstoeger:
@mikeller: Does the latest commit fix this?

@glance-: I think someone gave up and just let the memory leak instead of fighting against thread synchronization. Arguably, a leak is better than a hard to reproduce crash. But I hope that using value semantics should make this reasonably robust.

mikeller:
@bstoeger: Tested the latest version of this branch, reading / writing the configuration and firmware update all work as expected on OSTC4. I'll review the changes later on.

bstoeger:
Nice, thanks! I squashed the commit into the original one.

mikeller (approved):
LGTM

bstoeger:
Note that #4137, #4126 and #3863 start to seriously step on each other's toes. I can try to merge #4137 and #4126 if that is preferred.

bstoeger:
Oh fuck - merged the wrong PR. Wanted to merge #4133.

@dirkhh: revert? Last time I did that it completely messed up the CI thing.

dirkhh:
So this PR was tested and reviewed - why would you revert the merge?
Just trying to understand where we are.

bstoeger:
It was @mikeller who said it worked - no feedback from @glance- so far.

Well, it's a major embarrassment, but yes - (hard) reverting probably only makes things worse. :(

dirkhh:
Yes, but @mikeller has one of the dive computers that are supported.
I'm really leaning towards leaving this in and fixing it up if we find breakage.

glance-:
I was about to get around to this, but all my computers where out of battery so it took a while.

It works with the OSTC4 but something's broken with the OSTC3, but that looks like a libdivecomputer / hwos incompatibility, or just my device being weird. It reports a hardware id of 0xE, which would say it reports having "OSTC has analog inputs and S8 digital" which it don't.

Anyway, that's not a subsurface issue, and not a issue related to this PR.

Firmware updates for OSTC3 works to.

My good old OSTC 2N has a dying screen, so I didn't bother to test that one. My Suunto Vyper is out of battery and I don't have such a battery at hand, and If the OSTC3's and OSTC4's works, the rest probably works to.

glance- (commented):
LGTM

PR #4142, opened 2024-03-24
[TRIVIAL?] cleanup: remove picturedir_string()

bstoeger:
This seems to be dead code.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Unless I'm missing something, this doesn't seem to be used anywhere!?

In any case, I messed up my build environment (again!), so I need this PR for CI.

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -1194,16 +1194,6 @@ extern "C" const char *local_file_path(struct picture *picture)
 	return copy_qstring(localFilePath(picture->filename));
 }
 
-const QString picturedir()
-{
-	return QString(system_default_directory()).append("/picturedata/");
-}
-
-extern "C" char *picturedir_string()
-{
-	return copy_qstring(picturedir());
-}
-
 QString get_gas_string(struct gasmix gas)
 {
 	uint o2 = (get_o2(gas) + 5) / 10, he = (get_he(gas) + 5) / 10;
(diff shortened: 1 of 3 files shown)


bstoeger:
Note: last user seems to have been removed in 21ee440e5f96a2f7ddc810ced89f9ea1e7174bae.

In general, I would think it is a good idea to save pictures in the cloud, but git is probably not the right tool for this.

glance-:
Git lfs could be a interesting way...

dirkhh:
We went down this path before. git is DEFINITELY not the way to go.
And offering people unlimited storage for images is not a sustainable proposition.
I remember a certain member of the early developer team (back when I foolishly thought "hey, let's put them in git") reach a 30GB git repo in no time...
So a hard NO on that.

I just don't know how this would work

bstoeger:
Yes, it probably would have to be a user-owned cloud.

In any case, I will merge this, since it seems "obviously" correct.

atdotde:
There were ideas to maintain links in the log to online photo libraries like flickr, but that never took off. The problem was that manually inserting URLs is not very convenient and scraping them from web pages (the idea would be: You give the URL of a web page showing all those pictures and Subsurface finds all reasonable img links) is quite complicated given the use of JS to construct those pages.

PR #4143, opened 2024-03-24
desktop: fix crash in mainwindow.cpp

bstoeger:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

And the first crash caused by merging #4137.


bstoeger:
Merging, because this fixes a crash in current master and at least shouldn't make things worse.

PR #4148, opened 2024-03-30
desktop: unglobalize ComboBox-models

bstoeger:
The combo-boxes (cylinder type, weightsystem, etc.) were controlled by global models. Keeping these models up-to-date was very combersome and buggy.

Create a new model everytime a combobox is opened. Ultimately it might even be better to create a copy of the strings and switch to simple QStringListModel. Set data in the core directly and don't do this via the models.

The result is much simpler and easier to handle.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The models are some of the most convoluted parts of the code base and have always been a thorn in my side. The fact that Qt's model/view API is comically broken doesn't help.

Let's try to simplify things by cutting out the model-craziness as far as possible. In particular I don't understand why we should keep in sync global models for the combo-boxes. That only would make sense if the combo-boxes are updated while they are active. I don't think so. So let's allocate the models on each invocation of a combo-box.

### Additional information:

This is only very lightly tested! Would need some good test run.

Please also check the second commit - probably I'm missing something. But the cylinder names are not in QT_TRANSLATE_NOOP or similar, so back-translation makes no sense...?

### DIFF
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1090,11 +1090,6 @@ EditWeight::EditWeight(int index, weightsystem_t wsIn, bool currentDiveOnly) :
 		dives.clear();
 		return;
 	}
-
-	WSInfoModel *wsim = WSInfoModel::instance();
-	QModelIndexList matches = wsim->match(wsim->index(0, 0), Qt::DisplayRole, gettextFromC::tr(new_ws.description));
-	if (!matches.isEmpty())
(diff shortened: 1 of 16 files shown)


mikeller (approved):
I gave this a quick spin, and it seems to be working.

Issue #4119, opened 2024-03-04
Crash when creating a journey or adding a dive to a journey (in mobile app)


OlivierV78:
### Describe the issue:

- [X ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

The mobile app crash when creating a journey or adding a dive to a journey

### Operating system:

Android

Galaxy S21

No

### Subsurface version:

6.0.5112-CICD

Nightly build

### Steps to reproduce:

Open subsurface
Create a new journey with long hold on a dive
OR add a dive to a existing journey

### Current behavior:

App crash

### Expected behavior:

App not crash

### Additional information:

Log in attachment but as I have to launch subsurface again after crash I don't know if the crash can be seen in the log

[subsurface.log](https://github.com/subsurface/subsurface/files/14479928/subsurface.log)

mikeller:
Confirmed the behaviour in the android app. Doesn't happen on the desktop. Still need to figure out how to generate / access a log of the crash.

bstoeger:
Does it happen on mobile-on-desktop? There you can simply run it in a debugger.

mikeller:
I've never actually done a mobile-on-desktop build - looking into this now.

`adb logcat` when running on Android gives me the following:

```
03-11 21:47:12.222  4665  4785 I Subsurface: INFO: "17.230: showProgress: [Undo]Changes saved:'add 1 dives to trip [#1]'.\nUndo possible via context menu"
03-11 21:47:12.897  4665  4785 I Subsurface: INFO: Object 0x7a9ce06fc0 destroyed while one of its QML signal handlers is in progress.
03-11 21:47:12.897  4665  4785 I Subsurface: Most likely the object was deleted synchronously (use QObject::deleteLater() instead), or the application is running a nested event loop.
03-11 21:47:12.897  4665  4785 I Subsurface: This behavior is NOT supported!
03-11 21:47:12.897  4665  4785 I Subsurface: qrc:/org/kde/kirigami.2/private/ContextDrawerActionItem.qml:84: function() { [native code] }
```

mikeller:
No dice with mobile-on-desktop - adding dives to a trip or creating a trip works just fine.

mikeller:
Second time's the charm:

```
INFO: Object 0x55555856cdd0 destroyed while one of its QML signal handlers is in progress.
[...]

PR #4150, opened 2024-03-30
mobile: save changes from main event loop, not signal handler

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is my attempt at fixing #4119, see discussion there. I have no idea whether this is a good idea and what, if any, negative consequences this has. In light testing, it didn't crash anymore. I never understood the whole cloud thing, so beware!

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -602,6 +602,14 @@ void QMLManager::finishSetup()
 	// this could have brought in new cache directories, so make sure QML
 	// calls our getter function again and doesn't show us outdated information
 	emit cloudCacheListChanged();
+
+	// This is used to instruct the main-thread to sync the current state to disk/cloud.
+	// We must not sync to cloud from signal-handlers in the main thread, because cloud
+	// access runs the main event loop and that might delete the object that caused the
+	// signal. By using a connection of the "QueuedConnection" type, the signal will be
+	// queued and only executed once the signal handler finishes and the main event
+	// loop retakes control.
+	connect(this, &QMLManager::changesNeedSavingSignal, this, &QMLManager::saveUnsaved, Qt::QueuedConnection);
 }
 
 QMLManager::~QMLManager()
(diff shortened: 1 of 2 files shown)


mikeller (approved):
I am by no means a mobile Qt expert, but these changes look good to me - do you want me to test if this fixes the crashes, or have you been able to reproduce the problem reported in #4119?

bstoeger:
For me this fixes the crashes. Confirmation would obviously be appreciated. Moreover, I cannot tell whether this opens up a different can of worms. I think only @dirkhh is fluent with the code in `qmlmanager.cpp`.

dirkhh:
And I am traveling this weekend - will be back on the road in a few minutes...
I think moving this into the main loop is the right way to do it. The problem that I had with this in the past was that it froze the UI in some circumstances. but even that is better than simply crashing, so yeah, go for it.

dirkhh:
minor oh-shit moment.
on a slow connection this could make the app completely freeze if this is in the main thread. But I guess the simple remedy is 'turn off autosave'.

bstoeger:
But it currently *is* in the main thread. That's the whole reason for the crash: it runs the main loop while hanging in a signal handler of the main thread.

That said, I think this should be solved differently: Instead of a timer, send an signal with an `QueuedConnection`. Then the code will be executed once the signal handler returns to the main loop.

bstoeger:
Ok - I changed the thing to use a signal instead of a timer. This should execute the saving from the main event loop and not cause any delay, because it is executed immediately after the signal handler returns. In other words, this should not give any user-visible changes (except hopefully the lack of crashing, obviously).

Please check.

bstoeger:
@mikeller: Can you confirm that the new version fixes the problem(s)? I'm a bit apprehensive of this change.

mikeller:
Apologies, still on my cave trip, so my fingers are a bit worn. 
I gave it a quick spin on android, and creating and adding to trips from the dive list menu does not seem to trigger crashes any more.

bstoeger:
Thanks for checking. Let's merge and revert if it breaks more things than it fixes. 🤷

PR #4158, opened 2024-04-19
Add delete dive computer button

notrege:

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Added a 'X' button beside dive computer on imports window to delete the dive computer.   If this change is accepted, a follow on would be to remove the delete all computers from the preferences page.    For what it's worth,  I wouldn't have necessarily chosen macros for implementation, but I figure it's better to be consistent with the DC buttons than go down a different path.

### Documentation change:
This change adds an X (delete) button for each dive computer button on the import window.
### Mentions:

@mikeller


notrege:
something went terribly wrong with unsigned commits - have to try to start over.

PR #4160, opened 2024-04-20
Add delete dive computer buttons.

notrege:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Added a button to delete individual dive computers from the import window.    If this change is accepted, a potential follow on would be removing the delete all dive computers from the preferences window.    

### Changes made:
1. Added 4 buttons to the UI to accompany the dive computer buttons.
2. Added code to downloaddivecomputer to delete dive computers.  The code isn't necessarily the way I would choose to implement, but I tried to be consistent with the existing code for the dive computer buttons.

### Documentation change:
This change adds an X button beside each dive computer on the import window.

### Mentions:
@mikeller

### DIFF
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -117,13 +117,15 @@ DownloadFromDCWidget::DownloadFromDCWidget(const QString &filename, QWidget *par
 #define SETUPDC(num) \
 	if (!qPrefDiveComputer::vendor##num().isEmpty()) { \
 		ui.DC##num->setVisible(true); \
-		ui.DDC##num->setVisible(true); \
+		ui.DCFrame##num->setVisible(true); \
+		ui.DeleteDC##num->setVisible(true); \
 		ui.DC##num->setText(qPrefDiveComputer::vendor##num() + " - " + qPrefDiveComputer::product##num()); \
 		connect(ui.DC##num, &QPushButton::clicked, this, &DownloadFromDCWidget::DC##num##Clicked, Qt::UniqueConnection); \
-		connect(ui.DDC##num, &QPushButton::clicked, this, &DownloadFromDCWidget::DDC##num##Clicked, Qt::UniqueConnection); \
+		connect(ui.DeleteDC##num, &QPushButton::clicked, this, &DownloadFromDCWidget::DeleteDC##num##Clicked, Qt::UniqueConnection); \
 	} else { \
 		ui.DC##num->setVisible(false); \
-		ui.DDC##num->setVisible(false); \
+		ui.DCFrame##num->setVisible(false); \
+		ui.DeleteDC##num->setVisible(false); \
 	}
 
 void DownloadFromDCWidget::showRememberedDCs()
@@ -176,11 +178,11 @@ DCBUTTON(3)
 DCBUTTON(4)
 
 // Delete DC button slots
-#define DDCBUTTON(num) \
-void DownloadFromDCWidget::DDC##num##Clicked() \
+#define DELETEDCBUTTON(num) \
+void DownloadFromDCWidget::DeleteDC##num##Clicked() \
 { \
 	ui.DC##num->setVisible(false); \
-	ui.DDC##num->setVisible(false); \
+	ui.DeleteDC##num->setVisible(false); \
 	int dc = num; \
 	switch (dc) { \
 		case 1: \
@@ -203,13 +205,13 @@ void DownloadFromDCWidget::DDC##num##Clicked() \
 	qPrefDiveComputer::set_vendor(qPrefDiveComputer::vendor1()); \
 	qPrefDiveComputer::set_product(qPrefDiveComputer::product1()); \
 	qPrefDiveComputer::set_device(qPrefDiveComputer::device1()); \
-   DownloadFromDCWidget::showRememberedDCs(); \
+	DownloadFromDCWidget::showRememberedDCs(); \
 }
 
-DDCBUTTON(1)
-DDCBUTTON(2)
-DDCBUTTON(3)
-DDCBUTTON(4)
+DELETEDCBUTTON(1)
+DELETEDCBUTTON(2)
+DELETEDCBUTTON(3)
+DELETEDCBUTTON(4)
 
 void DownloadFromDCWidget::updateProgressBar()
 {
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -202,9 +202,6 @@ void DownloadFromDCWidget::DeleteDC##num##Clicked() \
 			qPrefDiveComputer::set_product4(QString()); \
 			qPrefDiveComputer::set_device4(QString()); \
 	} \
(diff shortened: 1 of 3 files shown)


desktop-widgets/downloadfromdivecomputer.ui
```diff
@@ -155,33 +155,104 @@
          </item>
          <item row="6" column="0">
           <layout class="QGridLayout" name="DCGrid">
+           <property name="verticalSpacing">
```
mikeller: Visually this does not look too great to be honest:

Screenshot from 2024-04-21 10-53-23

Two ideas for how this can be improved:
- use the 'delete' icon that is already used in the application;
- combine the dive computer name and the delete button in the same visual element.

Quick example for how this can be done (spacing not quite right yet):

image


```suggestion
               <item row="0" column="0">
+            <widget class="QFrame" name="frameConfigure">
+             <property name="frameShape">
+              <enum>QFrame::StyledPanel</enum>
+             </property>
+             <property name="frameShadow">
+              <enum>QFrame::Raised</enum>
+             </property>
              <layout class="QGridLayout" name="DC1Grid">
               <property name="horizontalSpacing">
                <number>0</number>
@@ -168,6 +175,9 @@
                 <property name="autoDefault">
                  <bool>false</bool>
                 </property>
+                <property name="flat">
+                 <bool>true</bool>
+                </property>
                </widget>
               </item>
               <item row="0" column="1">
@@ -175,12 +185,17 @@
                 <property name="autoDefault">
                  <bool>false</bool>
                 </property>
-               <property name="text">
-                <string>X</string>
+                <property name="flat">
+                 <bool>true</bool>
+                </property>
+                <property name="icon">
+                 <iconset>
+                  <normaloff>:list-remove-icon</normaloff>:list-remove-icon</iconset>
                 </property>
                </widget>
               </item>
              </layout>
+            </widget>
            </item>
        
```

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -117,10 +117,13 @@ DownloadFromDCWidget::DownloadFromDCWidget(const QString &filename, QWidget *par
 #define SETUPDC(num) \
 	if (!qPrefDiveComputer::vendor##num().isEmpty()) { \
 		ui.DC##num->setVisible(true); \
+		ui.DDC##num->setVisible(true); \
```
mikeller: `DDC` seems to be hard to understand, and introduces the potential for confusion with `DC` - why not be clear and use `DeleteDC`?

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -178,6 +175,42 @@ DCBUTTON(2)
...
+			qPrefDiveComputer::set_device3(qPrefDiveComputer::device4()); \
+		case 4: \
+			qPrefDiveComputer::set_vendor4(QString()); \
+			qPrefDiveComputer::set_product4(QString()); \
+			qPrefDiveComputer::set_device4(QString()); \
+	} \
+	qPrefDiveComputer::set_vendor(qPrefDiveComputer::vendor1()); \
+	qPrefDiveComputer::set_product(qPrefDiveComputer::product1()); \
+	qPrefDiveComputer::set_device(qPrefDiveComputer::device1()); \
+   DownloadFromDCWidget::showRememberedDCs(); \
```
mikeller:
```suggestion
	DownloadFromDCWidget::showRememberedDCs(); \
```

desktop-widgets/downloadfromdivecomputer.cpp
```diff
@@ -178,6 +175,42 @@ DCBUTTON(2)
...
+		case 3: \
+			qPrefDiveComputer::set_vendor3(qPrefDiveComputer::vendor4()); \
+			qPrefDiveComputer::set_product3(qPrefDiveComputer::product4()); \
+			qPrefDiveComputer::set_device3(qPrefDiveComputer::device4()); \
+		case 4: \
+			qPrefDiveComputer::set_vendor4(QString()); \
+			qPrefDiveComputer::set_product4(QString()); \
+			qPrefDiveComputer::set_device4(QString()); \
+	} \
+	qPrefDiveComputer::set_vendor(qPrefDiveComputer::vendor1()); \
```
mikeller: Not sure why this we are overwriting these preferences here?

mikeller (commented):
Good idea!

When this is done, we could also consider remvoving the 'Delete all dive computer connections' functionality from the Preferences - this will not really provide an advantage over deleting the presets individually directly where they are used:

image

Also n.b. that it is better form to mention people by their `@<github name>` handle (if this is known) than using their email address - I suspect that emails in here will be harvested by spammers.

desktop-widgets/downloadfromdivecomputer.ui
notrege: Apologies about the email - hope that doesn't create any problems for you.

If this is accepted, I will remove the Delete All from the preference page.

I agree that your layout looks better.   I don't really like the sizing, but am not sure how to fix that.

desktop-widgets/downloadfromdivecomputer.cpp
notrege: I may have this wrong, but my intention was to leave the prefs file as though the deleted computer never existed so they are renumbered.   In other words, if we have computers 1..4 and we delete #2, we'll end up with 1..3 instead of 1,3,4.

mikeller: The shifting around of the 'remembered' dive computers 1...4 makes sense - but I am not sure if setting the one dive computer that is shown when opening the dialog to the remembered dive computer 1 in every case is the right thing to do. I would probably just leave this, and assume that the user will be selecting a different model if this was the deleted dive computer.

notrege: I was trying to avoid setting the remembered dive computer to one that had been deleted, but I was on the fence about this behavior and I think it may have had a bad side effect if you deleted all but 1 computer.

mikeller: Can you please try to reproduce and document what the side effect was?
I think it is not ideal to change the pre-selected dive computer in cases when it is not the preset that is being deleted - so we should either:
- not change the pre-selected dive computer - if the user deletes this preset and then selects another dive computer to download from this will take care of updating the pre-select;
- only change the pre-selected dive computer if this is the remembered dive computer that was deleted.

mikeller (approved):
This looks great now, thank you very much!

mikeller:
We can do the removal of 'Delete all remembered dive computers' from the preferences in a separate pull request.

notrege:
Thanks Michael.  I will do the removal in the near future.

desktop-widgets/downloadfromdivecomputer.cpp
notrege: I didn't look very closely at causation as I agreed with your suggestion to leave the selected dive computer unchanged.   With the code in to change it to dive computer #1, if you deleted all but 1 dive computer the selected dive computer manufacturer, model, and device wasn't changed and clicking on the remaining dive computer did nothing.

PR #4167, opened 2024-04-25
stop silly warnings

dirkhh:
These both created warnings about potential truncation that in an of themselves were already bogus, but were turned into errors by zealous distro compiler defaults.
The new code is annoying and marginally pointless, but... alas.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation
- [x] none of the above... more like 'warning avoidance engineering'

### DIFF
--- a/core/uemis-downloader.cpp
+++ b/core/uemis-downloader.cpp
@@ -285,10 +285,12 @@ static char *build_filename(const char *path, const char *name)
 	int len = strlen(path) + strlen(name) + 2;
 	char *buf = (char *)malloc(len);
 #if WIN32
-	snprintf(buf, len, "%s\\%s", path, name);
+	l = snprintf(buf, len, "%s\\%s", path, name);
 #else
-	snprintf(buf, len, "%s/%s", path, name);
+	l = snprintf(buf, len, "%s/%s", path, name);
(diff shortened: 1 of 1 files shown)


mikeller (commented):
I think we should add a comment explaining the fact that this is intentionally wacky code to work around compiler shortcomings, or else some helpful soul might attempt to 'improve' it in the future.

Also, I think the compiler now thinks that `l` could do with a declaration. ;-)

dirkhh:
Grr.
I broke things when editing the patch. 
Will fix tomorrow

bstoeger:
> were turned into errors by zealous distro compiler defaults

We turned that on, because the printf() interface is super-dangerous (see #4171) and we have so many warnings that they are easily overlooked. Sadly the compiler can't catch errors when the format string is passed through `translate(...)`. :(

Also, the fundamental problem here seems to be the use of fixed-size buffers in the first place. The whole `uemis-downlosder.cpp` source file is completely bonkers.

bstoeger:
I think that become irrelevant with #4182. Closing.

PR #4171, opened 2024-05-01
profile: fix string formating in profile.cpp

bstoeger:
ae299d5e663cd672d1114c3fe90cf026b9ab463e introduced a format- string bug by splitting a format-string in two and splitting the arguments at the wrong place.

The compiler doesn't warn in this case, because the format- string is passed through translate(...).

This should have crashed, but for some reason didn't, at least on Linux.

Fix the arguments.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix potential crash. See commit description and discussion in #4126.

### DIFF
--- a/core/profile.cpp
+++ b/core/profile.cpp
@@ -1368,8 +1368,8 @@ static std::vector<std::string> plot_string(const struct dive *d, const struct p
 	std::vector<std::string> res;
 
 	depthvalue = get_depth_units(entry->depth, NULL, &depth_unit);
-	res.push_back(casprintf_loc(translate("gettextFromC", "@: %d:%02d"), FRACTION(entry->sec, 60), depthvalue));
-	res.push_back(casprintf_loc(translate("gettextFromC", "D: %.1f%s"), depth_unit));
+	res.push_back(casprintf_loc(translate("gettextFromC", "@: %d:%02d"), FRACTION(entry->sec, 60)));
+	res.push_back(casprintf_loc(translate("gettextFromC", "D: %.1f%s"), depthvalue, depth_unit));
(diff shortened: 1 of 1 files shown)


mikeller (approved):
Yes, this fixes the crashing problem in Windows that was reported in https://groups.google.com/g/subsurface-divelog/c/7Uo7A9YTGhg/m/39JxckLkBQAJ. 
Thank you for the quick turnaround!

core/profile.cpp
```diff
@@ -1368,8 +1368,8 @@ static std::vector<std::string> plot_string(const struct dive *d, const struct p
 	std::vector<std::string> res;
 
 	depthvalue = get_depth_units(entry->depth, NULL, &depth_unit);
-	res.push_back(casprintf_loc(translate("gettextFromC", "@: %d:%02d"), FRACTION(entry->sec, 60), depthvalue));
-	res.push_back(casprintf_loc(translate("gettextFromC", "D: %.1f%s"), depth_unit));
+	res.push_back(casprintf_loc(translate("gettextFromC", "@: %d:%02d"), FRACTION(entry->sec, 60)));
```
mikeller, line 1371: I can see how a macro (having function syntax) returning a tuple is basically inviting this kind of bug - might be helpful to make it a bit more explicit by calling it `FRACTION_TUPLE()` or similar.

bstoeger: Well, I tripped over that macro a few times already, so I should have known better. 🤷

PR #4172, opened 2024-05-01
Refactoring: Improve Naming of FRACTION and SIGNED_FRAC defines.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Make it more obvious that the FRACTION and SIGNED_FRAC defines return a tuple / triplet of values.

### Related issues:

Fixes https://github.com/subsurface/subsurface/pull/4171#discussion_r1585941133

### Mentions:

Complained-about-by: @bstoeger


bstoeger (approved):
lgtm.

Ultimately, we might think about something like `std::format` and a special fraction formatter. But first, let's get rid of the C-string mess.

PR #4181, opened 2024-05-06
Planner: Fix Editing of Plans in Multi-Divecomputer Dives.

### LINKED ISSUE #1913: Plan dive: Should be a Main Menu item
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [ ] New feature request
- [x ] Discussion request

### Issue long description:

The Plan dive menu item is actually launching a very comprehensive module. So much so that it could be argued that this menu item should not be under the Log menu but should should be a Main Menu item of its own named Plan.

Since dive planning and dive logging are different tasks, separated in both space and time, it may also be argued that Dive Plans are not a part of the Dive List. They may therefore be either:
- included in the Dive List
- included in a Dive Plan List

If the decision is to include Dive Plans into the Dive List as it currently is implemented, one could suggest that there should exist an easy visual way to distinguish actual dive logs and dive plans from each other. This naturally may also have further feature requests such as the ability to filter these logs by:
- All logs
- Dive logs
- Dive plans

If the decision is to provide separate lists then filtering is such a requirement. However, the ability to associate and compare a plan against a dive (vis-a-versa) would be a good feature that will enable comparison and analysis of how good dive planning was and how closely the dive was to planning (Plan the Dive, Dive the Plan).

### Subsurface version:

v4.8.4

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Currently editing of planned dives that have been merged with actual (logged) dives only works if the 'Planned dive' divecomputer is the first divecomputer, and this divecomputer is selected when clicking 'Edit planned dive'. In other cases the profile of the first divecomputer is overlaid with the profile of the planned dive, and the first divecomputer's profile is overwritten when saving the dive plan. Fix this problem.

### Related issues:

Fixes #1913 

### Mentions:

Triggered by @SeppoTakalo's comment (https://github.com/subsurface/subsurface/issues/1913#issuecomment-2075562119): Users don't like that planned dives show up as their own entries in the dive list, so being able to merge them with the actual dive after it has been executed is a good feature - but this wasn't working well until now.

### DIFF
--- a/core/dive.cpp
+++ b/core/dive.cpp
@@ -339,28 +339,27 @@ extern "C" void selective_copy_dive(const struct dive *s, struct dive *d, struct
 }
 #undef CONDITIONAL_COPY_STRING
 
-/* copies all events from all dive computers before a given time
+/* copies all events from the given dive computer before a given time
    this is used when editing a dive in the planner to preserve the events
    of the old dive */
-extern "C" void copy_events_until(const struct dive *sd, struct dive *dd, int time)
+extern "C" void copy_events_until(const struct dive *sd, struct dive *dd, int dcNr, int time)
 {
 	if (!sd || !dd)
 		return;
 
 	const struct divecomputer *s = &sd->dc;
-	struct divecomputer *d = &dd->dc;
+	struct divecomputer *d = get_dive_dc(dd, dcNr);
 
-	while (s && d) {
-		const struct event *ev;
-		ev = s->events;
-		while (ev != NULL) {
-			// Don't add events the planner knows about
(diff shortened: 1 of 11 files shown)


SeppoTakalo:
Just out of curiosity:
If you merge a recorded dive and planned dive together. Are both or just one "computer" used when calculating dive statistics?

Sometimes it is useful or even required to know total hours of dives (for example total hours of CCR) before going to next course. So these numbers should be based on record, not a plan.

atdotde:
IIRC only the first divecomputer for a dive is used for these calculations. So it depends on how you sort your computers.

PR #4183, opened 2024-05-07
CICD: Improve the GitHub Actions for Linux.

mikeller:
Do a few things:
- add a build for Debian trixie (as discussed in #4182);
- add a build for Ubuntu 24.04;
- rename the build definitions to match the build names;
- update the artifact uploads to use a non-deprecated version of the action, and name the artifact appropriately.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Improve the GitHub Actions for Linux.

### Changes made:

Do a few things:
1. add a build for Debian trixie;
2. add a build for Ubuntu 24.04;
3. rename the build definitions to match the build names;
4. update the artifact uploads to use a non-deprecated version of the action, and name the artifact appropriately;
5. remove a stale workflow file.

### Related issues:

As discussed in #4182.


.github/workflows/linux-appimage.yml
```diff
@@ -45,18 +48,12 @@ jobs:
...
 
-    - name: prepare PR artifacts
-      if: github.event_name == 'pull_request'
-      run: |
-        mkdir -p Linux-artifacts
-        mv Subsurface.AppImage Linux-artifacts
-
     - name: PR artifacts
       if: github.event_name == 'pull_request'
       uses: actions/upload-artifact@v3
```
mikeller: Now this is an interesting one @dirkhh:
`upload-artifact@v3` will stop working on 30 Nov 2024, but `upload-artifact@v4` does not work on our `subsurface/trusty-qt512:1.1` docker image, because the glibc version is too old for the node version used in the action.
Do you think it will make sense to move the AppImage generation to a newer base image?

dirkhh: Yes
I'm surprised anything still builds on trusty 
That's the years old

mikeller: It's the AppImage - not sure if there's any constraints around what versions the resulting AppImage then can run on?

dirkhh: I understand. The idea with the AppImage is to build it on "as old as possible" a base so its dependency are likely to be met by whatever distro it is installed on. So I'd try 16.04 and if that also fails 18.04, etc

dirkhh: actually @probonopd might have good ideas which base distros work best - this is all based on his work

probonopd: Unless you bundle _all_ libraries (including libc), the AppImage will only run on target systems that are not older than the build system. So for practical purposes, the oldest still-supported LTS release sounds reasonable to me. For Ubuntu, this would currently be 20.04 LTS (Focal Fossa).

mikeller: Thanks @probonopd, this makes sense. 

> For Ubuntu, this would currently be 20.04 LTS (Focal Fossa).

What is this based on?

I was about to have a go at setting up a build on 16.04, and failing that, 18.04, in order to see if we can build an AppImage that can run on a wider range of legacy distributions.

glance-: I'd rather see that we build the appimage on something which still receives security patches, than something ancient to allow a very few users who should upgrade to something which still receives security patches to run it.

mikeller: @glance-: Ubuntu provides security updates for 12 years past the release date for their LTS (Long Term Support) releases. So even 14.04 will still be supported for 2 more years from a security perspective: https://canonical.com/blog/canonical-expands-long-term-support-to-12-years-starting-with-ubuntu-14-04-lts

glance-: With "Ubuntu Pro" yes, but not without it.

dirkhh: I appreciate the thought of "only use security supported things".
Yet, for this use case, I consider this completely irrelevant.
Anyone running thirdparty binaries that are delivered as AppImages on their system clearly isn't concerned with security in the first place.
So as far as trade-offs go, let's err on the side of supporting more people vs assume that there is a serious security risk in people using an older library in the first place.

mikeller: > With "Ubuntu Pro" yes, but not without it.

Ubuntu Pro is completely free for non-enterprise use.

probonopd:
>> For Ubuntu, this would currently be 20.04 LTS (Focal Fossa).

> What is this based on?

I was looking at https://ubuntu.com/about/release-cycle "end of standard support".

PR #4187, opened 2024-05-09
CICD: Move MacOS / iOS Build Qt Resources into GitHub.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Move the Qt resources required for the build for MacOS and iOS into
GitHub, into their own repositories. This removes the need to publish
them on an external file server and download them from there for every
build.
It will also make it easier for contributors to update these resources
if needed.

### Related issues:

https://github.com/subsurface/qt-mac
https://github.com/subsurface/qt-ios

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -11,13 +11,13 @@ on:
     branches:
     - master
 
-env:
-  BUILD_ROOT: ${{ github.workspace }}/..
-  KEYSTORE_FILE: ${{ github.workspace }}/../subsurface.keystore
 
 jobs:
   buildAndroid:
     runs-on: ubuntu-latest
+    env:
+      BUILD_ROOT: ${{ github.workspace }}/..
+      KEYSTORE_FILE: ${{ github.workspace }}/../subsurface.keystore
     container:
       image: docker://subsurface/android-build:5.15.2
 
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -12,7 +12,7 @@ on:
     - master
 
 jobs:
-  iOSBuild:
+  build:
     runs-on: macOS-11
     steps:
     - name: switch to Xcode 11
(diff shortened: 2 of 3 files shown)


mikeller:
image

dirkhh:
nice

mikeller:
We could probably get more speed improvements by caching things like the google maps plugin or the 3rd party components, but they are downloaded and installed inside the build scripts, which will make it hard to make this depend on the outcome of the cache load.

mikeller:
@dirkhh: I have thought about this a bit more. We could probably get completely off Backblaze, and avoid having to download this data from external at all if we put the prebuilt Qt components for the MacOS / iOS builds into a repository of their own, and consumed them in the same way that we are already consuming the Google maps plugin - there is no point in caching stuff that is already stored in GitHub.

dirkhh:
IIRC the problem with that was that a build of these components isn't possible inside a GitHub Action (requires too much memory / too much time).
But of course we should be able to figure out how to simply upload the binary artifacts that are built elsewhere.

mikeller:
My thoughts too - having these resources available in pre-built form in a local repository will be an effective way to keep them at hand for the MacOS based builds that need them.
If you create two repositories `qt-mac` and `qt-ios` or similar, and give me access I can get this set up.

dirkhh:
> If you create two repositories `qt-mac` and `qt-ios` or similar, and give me access I can get this set up.

You should be able to that without my help. At least I'm 95% sure that I gave you the necessary permissions...

mikeller:
@dirkhh: I have re-engineered this one to keep everything inside GitHub.

mikeller:
Note that it now also publishes artifacts:
- MacOS .dmg for pull request builds;
- iOS simulator app for all builds.

dirkhh (approved):
Nice. And thank you for doing that!

PR #4191, opened 2024-05-12
core: fix recently leak when clearing dive log

bstoeger:
46cf2fc0867 fixed a bug where clearing of a divelog, such as the one used for import, would erase dives in the global(!) divelog.

However, the new code used the function clear_dive_table(), which only cleared the table without unregistering the dives. In particular, the dives were not removed from the trips, which means that the trips were not free()d.

This reinstates the old code, but now passes a divelog paremeter to delete_single_dive() instead of accessing the global divelog. Moreover, delete dives from the back to avoid unnecessary copying.

An alternative and definitely simpler solution might be to just add a "clear_trip_table()" after "clear_dive_table()".

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This fixes a memory leak that was recently introduced to fix another bug.

The much more simpler and pragmatic way to fix this might be a `clear_trip_table`. However, I think it is a good idea to reduce the number of functions that implicitly access the global dive table and I think this is the right fix. It also detects when the dive<->trip relation gets messed up.

### Changes made:

1) reinstate old dive-deletion code, but access the supplied divelog, not the global log.

### Related issues:

Fix of fix in #4179.

### Release note:

Not necessary - only mem leak.

### DIFF
--- a/core/divelist.c
+++ b/core/divelist.c
@@ -767,18 +767,6 @@ struct dive *unregister_dive(int idx)
 	return dive;
 }
 
-/* this implements the mechanics of removing the dive from the global
- * dive table and the trip, but doesn't deal with updating dive trips, etc */
-void delete_single_dive(int idx)
-{
-	struct dive *dive = get_dive(idx);
-	if (!dive)
-		return; /* this should never happen */
-	remove_dive_from_trip(dive, divelog.trips);
-	unregister_dive_from_dive_site(dive);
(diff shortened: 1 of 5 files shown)


mikeller (approved):
Well spotted!
Tested, cancelling out of an import from a dive computer works just as it should.

PR #4198, opened 2024-05-14
[RFC] More conversions to C++ (including uemis-download refactoring)

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This is more conversion to C++. It is painful and puts us in a weird transition state, but at some points pieces start to fall together and that is very satisfying. I really have to restrain myself and want to stabilize what I have here, before continuing.

This contains two milestones:
1) No more C-files, that means no more `#ifdef __cplusplus`, no more `extern "C"`. 
2) First conversion away from that crazy `TABLE` macro stuff: `divesite.cpp`.

To get an impression of what I have in mind, please look at the final `divesite.cpp` and `divesite.h` files. Not perfect, but SO MUCH nicer than the original. Finally, no more guessing who owns what and what will be freed where by whom, or whether it will be free()d at all.

Also contains the rewrite of `uemis-downloader.cpp`. I know that Dirk is very apprehensive of this give that there isn't even a handful of users left. And yes, the chance that I broke this at multiple points is 100%. However:
1) This is one of the worst files and a major embarrassment. It shows that most of our contributors (myself included) can't write proper C. Which makes it a bad language for this project.
2) I still hope that one of the last users could help in porting this to proper C++, without resource leaks and other extremely sketchy code. Sometimes I wonder how this things does not crash. Some of the code-paths seem to be never executed.

I'm marking as a Draft, as this is not ready.

### Changes made:

1) Convert numerous files to C++.
2) Convert C-strings to std::string.
3) Use std::vector<> for divesites.

### Release note:

No.
### Documentation change:

No.


bstoeger:
Ok, so an update: Since no one commented, I continued conversion of C to C++. Rebasing on master was the expected disaster.

This is very very far from perfect, but still I think it is a huge step forward. 

Some of the problems I identified, which should however be fixable by followups:
- This introduces lots of vexing casts between `size_t` and `int`, because we tend to use signed indexes, but the standard containers use unsigned indexes. I plan to solve this for loops with something like `for (int idx: index_range(v)` and for index checks with `check_index(idx, v)`. That should eliminate the vast majority of casts.
- There is now a total mess between pointer- and reference-arguments. The rule should be: use a reference if the called function does not handle null-pointers.
- There are numerous pointless UTF16 (QString) <->UTF8 (std::string) conversions. I'd like to have a clear interface between core (std::string) and UI (QString). That one is a bit problematic, because QString has RegExp functionality and the libstd regexp is horrible. I'll have to take a closer look at that.
- The `owning_table` code is quite idiosyncratic. It tries to formalize what we already do in the undo-code. Suggestions how to improve this are welcome.
- There is now more entanglement of include headers. I hope that this can be solved by splitting up the headers of objects and table-of-objects.

Concerning Uemis: Would it be a good idea if I contact the known(?) users, asking whether there even is interest in new Subsurface versions supporting this computer? And whether they would be willing to help debugging it?

mikeller:
Any idea when this will be ready?
I have got another pull request in the works to simplify and improve gas selection in the planner, and that one will probably be another doozie to rebase (either way).

bstoeger:
Depends on what you mean with "ready". Ready for intense testing? I hope rather soon (next week-ish). Conversion of the dive_table would be a satisfying target for that. Ready for merging? That strongly depends on the help I get for testing and how we want to/can resolve the "Uemis issue".

So I would think just go ahead. I fear it won't be the last rebase.

dirkhh:
Uemis is dead. One way or another. I think it would be just as reasonable to simply remove support for it and point the three people (?) who still use it to the last release that supported it.
I give it a ZERO PERCENT likelihood that this continues to work after the C++ conversion, anyway, so removing seems like the much more honest approach.

As for the rest - let's identify a solid candidate for the next "current" release... the one that I foolishly called weekly at some point when I should have said "once or twice a quarter". And once we have that, let's merge this and have people using latest test it for us.

mikeller:
> Depends on what you mean with "ready". Ready for intense testing? I hope rather soon (next week-ish). Conversion of the dive_table would be a satisfying target for that. Ready for merging? That strongly depends on the help I get for testing and how we want to/can resolve the "Uemis issue".

I am happy to do some functional testing as part of a review when you indicate that this is ready.
I think that with the 'agile' approach of releasing all changes in `master` to customers on 'latest' it will be easier if we break down these refactorings into smaller chunks (like 'weekly achievements') - this will allow us to leverage the user-feedback loop that we've created, and will allow us to keep delivering value to customers by implementing and pushing bugfixes and small improvements based on customer feedback without incurring a big cost for rebasing from them.

> Uemis is dead. One way or another. I think it would be just as reasonable to simply remove support for it and point the three people (?) who still use it to the last release that supported it. I give it a ZERO PERCENT likelihood that this continues to work after the C++ conversion, anyway, so removing seems like the much more honest approach.

Let's just treat this as a SCREAM test. If Uemis users pipe up after we remove Uemis support we know it is still wanted, and we have contacts to work with on getting it reinstated and fixed up. 😁

> As for the rest - let's identify a solid candidate for the next "current" release... the one that I foolishly called weekly at some point when I should have said "once or twice a quarter". And once we have that, let's merge this and have people using latest test it for us.

I think [6.0.5204](https://github.com/subsurface/nightly-builds/releases/tag/v6.0.5204-CICD-release) is a good candidate for this - there are some improvements in there like the move to a MacOS build that runs on Apple silicon, and we have done some bug fixes coming out of these improvements.

On a side note, gosh, 5204 means that we have delivered benefit to users already > 200 times since we moved to an 'agile' way of releasing around the beginning of the year. Go us!

dirkhh:
There are several things worth celebrating.
- more than 200 builds per PR merge... that's HUGE
- the project is thriving and doing better than in quite a while, even though I still feel fairly burned out... the fact that you have taken on so much of the load is amazing and wonderful
- we have landed on a model that really seems to work well - fast moving "latest", and slow moving, but not completely outdated "current". I really like that.

As to using 5204 for the next current. Let's earmark that thought and give it a few days. If we still think so by the end of the week, I'll be happy to push that along. At that point it will have been almost exactly a month since I pushed 5144 as current which feels like a very nice rhythm.

mikeller:
> As to using 5204 for the next current. Let's earmark that thought and give it a few days. If we still think so by the end of the week, I'll be happy to push that along.

Sounds like a plan!

> At that point it will have been almost exactly a month since I pushed 5144 as current which feels like a very nice rhythm.

Once a month sounds like a good frequency. Not too sure that I'd want to generalise this though, as I think that we should be flexible and try to base this on making good (educated) choices around which releases to bless for inclusion in the 'weekly' (or however we want to call it) release channel is where the human element comes in on top of all of the automation and 'business as usual' process, and should not be governed by sticking to a pattern. The frequency of 'weekly' releases is always going to be a trade-off between giving users access to bugfixes and new features sooner, and overloading with the need to update all of the time.

So in the end the decision to 'bless' a release should be based on:
1. the risk coming from change contained in releases following the to be 'blessed' one;
2. value added in new changes since the last 'weekly' release;
3. the release frequency we would like to adhere to.

dirkhh:
Just to be very clear. I agree that this should not be schedule driven.

bstoeger:
> I give it a ZERO PERCENT likelihood that this continues to work after the C++ conversion, anyway, so removing seems like the much more honest approach.

Sure. Just for statistical reasons, there will be a dozen or so typos / "brain-farts". But these should all be easily fixable (see e.g. smartrak import). If we remove the code, resurrecting it will be much harder than if we drag it along. We could keep the code for now, but disable it and wait whether someone complains...

bstoeger:
Update: I converted the dive-table. That is very hairy, because there are weird assumptions  onwhen this table is sorted and when not. As commented elsewhere, I would prefer having it _always_ sorted, but that breaks the tests for now.

Good news: finally no more `MAKE_TABLE_*` macros.
Bad news: When working on this, I got quite frustrated realizing how many core functions implicitly access the global dive table. And you never know which do.

That said, this is holding things up, so I will start concentrating on testing. @mikeller: if you could look at the OSTC configuration thing, that would be great, because that is one of the things I can't test.

bstoeger:
@mikeller: Could you please check if I properly rebased on the cylinder changes? Excuses in advance for butchering it - I'm a bit overwhelmed at the moment with all the loose ends.

bstoeger:
Dive import crashes - let me fix that first.

commands/command_divelist.cpp
```diff
@@ -404,34 +391,33 @@ AddDive::AddDive(dive *d, bool autogroup, bool newNumber)
 	setText(Command::Base::tr("add dive"));
 	// By convention, d is a pointer to "displayed dive" or a temporary variable and can be overwritten.
 	d->maxdepth.mm = 0;
-	d->dc.maxdepth.mm = 0;
+	d->dcs[0].maxdepth.mm = 0;
```
mikeller, line 393: Should this rather reset `maxdepth` for _all_ dive computers in the dive? Unfortunately this seems to be quite a common 'antipattern' in our codebase that in a lot of places we only handle the (arbitrary) first dive computer, resulting in hard to track down breakages.
(This is not a new problem here though.)

commands/command_divelist.h
```diff
@@ -15,16 +15,16 @@ namespace Command {
 
 // This helper structure describes a dive that we want to add.
 struct DiveToAdd {
-	OwningDivePtr	 dive;		// Dive to add
-	dive_trip	*trip;		// Trip the dive belongs to, may be null
-	dive_site	*site;		// Site the dive is associated with, may be null
+	std::unique_ptr<struct dive>	 dive;		// Dive to add
+	dive_trip			*trip;		// Trip the dive belongs to, may be null
```
mikeller, line 19: The introduction of this kind of noise is why I see aligned lists in source code as a net negative,

commands/command_edit.cpp
```diff
@@ -305,11 +304,11 @@ QString EditAtmPress::fieldName() const
 // ***** Duration *****
 void EditDuration::set(struct dive *d, int value) const
 {
-	d->dc.duration.seconds = value;
-	d->duration = d->dc.duration;
-	d->dc.meandepth.mm = 0;
-	d->dc.samples = 0;
-	fake_dc(&d->dc);
+	d->dcs[0].duration.seconds = value;
```
mikeller, line 307: Will this actually work for dives with multiple dive computers?

core/dive.cpp
```diff
@@ -17,10 +17,13 @@
 #include "errorhelper.h"
 #include "event.h"
 #include "extradata.h"
+#include "format.h"
 #include "interpolate.h"
 #include "qthelper.h"
+#include "range.h"
 #include "membuffer.h"
 #include "picture.h"
+#include "range.h"
```
mikeller, line 26: Double include.

```suggestion
```

core/dive.cpp
```diff
@@ -2352,123 +2048,92 @@ static int likely_same_dive(const struct dive *a, const struct dive *b)
...
-			if (same_dc(dc, check) || (prefer_downloaded && might_be_same_device(dc, check))) {
-				*p = check->next;
-				check->next = NULL;
-				free_dc(check);
-				continue;
-			}
-			p = &check->next;
-		}
+	// Note: since the vector doesn't grow and we only erase
+	// elements after the iteratore, this is fine.
```
mikeller:
```suggestion
	// elements after the iterator, this is fine.
```

mikeller:
Quick status update: I have tested my recent changes to cylinder management with this, and I have tested the OSTC4 configuration manager and both are fine. I am about a third through with reviewing, and hopefully I'll get around to completing it over the weekend.

bstoeger:
> I have tested my recent changes to cylinder management with this, and I have tested the OSTC4 configuration manager and both are fine.

Wow, that's miraculous! Thanks for testing!

core/owning_table.h
```diff
@@ -0,0 +1,171 @@
...
+ * are deleted (freed) when removed from the table.
+ * This header defines an std::vector<> of unique_ptr<>s to make this
+ * explicit.
+ *
+ * The state I want to reach across the code base: whenever a part of the
+ * code owns a heap allocated object, it *always* possesses a unique_ptr<>
+ * to that object. All naked pointers are invariably considered as
+ * "non owning".
+ *
+ * There are two ways to end ownershup:
```
mikeller:
```suggestion
 * There are two ways to end ownership:
```

core/datatrak.cpp
```diff
@@ -213,11 +209,13 @@ static char *dt_dive_parser(unsigned char *runner, struct dive *dt_dive, struct
 	 * Subsurface only have a location variable, so we have to merge DTrak's
 	 * Locality and Dive points.
 	 */
-	snprintf(buffer, sizeof(buffer), "%s, %s", locality, dive_point);
-	ds = get_dive_site_by_name(buffer, log->sites);
-	if (!ds)
-		ds = create_dive_site(buffer, log->sites);
-	add_dive_to_dive_site(dt_dive, ds);
+	{
```
mikeller, line 212: For my understanding: What is the reason to use a block here?

core/units.h
```diff
@@ -327,15 +319,15 @@ struct units {
  * actually use. Similarly, C instead of Kelvin.
  * And kg instead of g.
  */
-#define SI_UNITS                                                                                           \
-        {                                                                                                  \
-	        .length = METERS, .volume = LITER, .pressure = BAR, .temperature = CELSIUS, .weight = KG,  \
-		.vertical_speed_time = MINUTES, .duration_units = MIXED, .show_units_table = false         \
+#define SI_UNITS 																\
```
mikeller, line 322: Tabs are better than spaces, but not sure we want that many of them.

mikeller:
@bstoeger: What is your plan for this pull request? I was through with reviewing for about 240 out of the 330 files, and now with the changes that you pushed I am back to 160, and with GitHub's UI I'll have to do another review of everything in the previously reviewed files that were changed - so trying to get everything reviewed while you keep pushing changes seems to be a Sisyphus task with no end in sight.

bstoeger:
> @bstoeger: What is your plan for this pull request?

Get this merged ASAP. It is holding up numerous other projects. For example I found that copy/paste of the divesite is buggy, but I neither want to fix it on master, nor here.

> I was through with reviewing for about 240 out of the 330 files, and now with the changes that you pushed I am back to 160, and with GitHub's UI I'll have to do another review of everything in the previously reviewed files that were changed - so trying to get everything reviewed while you keep pushing changes seems to be a Sisyphus task with no end in sight.

Sorry, when I find errors, I have to rebase and that gives a mess. I stopped doing any more conversions, because things are becoming too cumbersome (and I am currently drowning in "real-life" work).

Maybe a detailed review is not necessary and we should just weed out bugs after merging? I'm convinced that it is a big step forward (even though _very_ far from perfect).

mikeller (approved):
I am good with getting this merged without a complete review - what I've seen of it looks very good, I've had only a few cosmetic remarks, and some questions.

commands/command_divelist.cpp
bstoeger: This doesn't seem to be an issue here, because `addDive()` is only called on dives with a single DC. For consistency, this perhaps should be changed. Or perhaps we should use `force_fixup_dive()`?

In any case, I tried to change behavior as little as possible, so let's do that kind of changes afterward.

commands/command_divelist.h
bstoeger: I have no hard opinion on that topic.

commands/command_edit.cpp
bstoeger: We can only set the duration on "manually added" dives. Not sure what happens if a manually added dive is merged with a dive computer dive. That should be checked. But as noted above: I tried to avoid changing behavior.

core/dive.cpp
bstoeger: Oops. Fixed locally, will push later.

core/datatrak.cpp
bstoeger: To free `buffer2` as soon as it is not needed anymore.
To not pollute the outer namespace with `buffer2`.

Not really necessary...

core/units.h
bstoeger: I believe this is just copy&paste. Feel free to change.

Ultimately, I want to remove this macro anyway.

core/owning_table.h
bstoeger: Fixed.

core/dive.cpp
bstoeger: Fixed.

core/units.h
```diff
@@ -327,15 +319,15 @@ struct units {
...
+        {																	\
```
mikeller:
```suggestion
        { \
```

core/units.h
```diff
@@ -327,15 +319,15 @@ struct units {
...
+	        .length = units::METERS, .volume = units::LITER, .pressure = units::BAR, .temperature = units::CELSIUS, .weight = units::KG,	\
```
mikeller:
```suggestion
	        .length = units::METERS, .volume = units::LITER, .pressure = units::BAR, .temperature = units::CELSIUS, .weight = units::KG, \
```

core/units.h
```diff
@@ -327,15 +319,15 @@ struct units {
...
+		.vertical_speed_time = units::MINUTES, .duration_units = units::MIXED, .show_units_table = false				\
```
mikeller:
```suggestion
		.vertical_speed_time = units::MINUTES, .duration_units = units::MIXED, .show_units_table = false \
```

mikeller (approved):

mikeller (approved):

commands/command_edit.cpp
mikeller: We are definitely not doing anything to stop users from combining manually added dives with other dives. And there might be a use case for it when people use manually added dives as a 'poor man's dive planner' for recreational dives.
And as I've found out in #4181, #4205, and #4210, there are A LOT of places where our code blindly uses or modifies the first dive computer in a dive, and ends up creating a mess in multi dive computer dives.
So this is probably an area where a lot of buggy behaviour can by fixed by addressing this one thing.

bstoeger:
@salvadorcunat: A small heads-up: I probably broke smtk import again. Sorry in advance for that. It really should be the last time that I do major data-structure reshuffling. Promised.

@dirkhh: Any objections? I would like merge in the near future and then fix the inevitable fallout. Concerning Uemis, let's see if someone complains. If nobody does in a year or so, we might think about removing the code.

dirkhh:
> @dirkhh: Any objections? I would like merge in the near future and then fix the inevitable fallout. Concerning Uemis, let's see if someone complains. If nobody does in a year or so, we might think about removing the code.

I have not done a full code review, but I think this needs to go in sometime soon before the merge becomes too painfull.

I found out that there is apparently at least one currently active Uemis user, so we'll learn if this breaks things once he updates :)

bstoeger:
Ok, let's bite the bullet. Waiting seems to just pointlessly hold up things.

mikeller:
@dirkhh: Do you want to bless 6.0.5231 - this will unlock the latest updates before this went in (including libdivecomputer) for users of 'current'.

dirkhh:
Sure, I'll add that to my list

PR #4199, opened 2024-05-14
core: fix INFO() and ERROR() macros in core/serial_ftdi.cpp

bstoeger:
Error introduced in da7ea17b66: the INDO() and ERROR() macros pass stdout instead of the format string as first parameter to report_error(). Ooooops. How did this ever pass the compile tests!?

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Stupid bug - how did this ever compile on Windows!?

### DIFF
--- a/core/serial_ftdi.c
+++ b/core/serial_ftdi.c
@@ -38,8 +38,8 @@
 #endif
 
 #include "errorhelper.h"
-#define INFO(context, fmt, ...)	report_info(stderr, "INFO: " fmt, ##__VA_ARGS__)
-#define ERROR(context, fmt, ...)	report_info(stderr, "ERROR: " fmt, ##__VA_ARGS__)
+#define INFO(context, fmt, ...)	report_info("INFO: " fmt, ##__VA_ARGS__)
+#define ERROR(context, fmt, ...)	report_info("ERROR: " fmt, ##__VA_ARGS__)
 //#define SYSERROR(context, errcode)	ERROR(__FILE__ ":" __LINE__ ": %s", strerror(errcode))
 #define SYSERROR(context, errcode)	;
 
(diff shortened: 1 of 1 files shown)


core/serial_ftdi.c
```diff
@@ -38,8 +38,8 @@
 #endif
 
 #include "errorhelper.h"
-#define INFO(context, fmt, ...)	report_info(stderr, "INFO: " fmt, ##__VA_ARGS__)
-#define ERROR(context, fmt, ...)	report_info(stderr, "ERROR: " fmt, ##__VA_ARGS__)
+#define INFO(context, fmt, ...)	report_info("INFO: " fmt, ##__VA_ARGS__)
+#define ERROR(context, fmt, ...)	report_info("ERROR: " fmt, ##__VA_ARGS__)
```
mikeller, line 42: While the bugfix looks good, should we as well get rid of the `context` parameter where it still exists, seeing that this is not used any more? This should probably be done in a separate pull request though.

mikeller (approved):
Looks good to me, well spotted.

bstoeger:
> well spotted.

Not be me though, the compiler refused once I turned the file to C++.  🤷

core/serial_ftdi.c
bstoeger: I wonder if this parameter ever was used?

mikeller: I think that's a pattern from libdivecomputer, where the context is providing the sink for the log data.

glance-: This code grew up in libdivecomputer but ended up in subsurface.

I'm not even sure if this code is in use in any of our builds any more, since we moved to the jni-wrapped java serial drivers on android.

PR #4203, opened 2024-05-17
Desktop: Fix Finding Reported by Coverity.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


bstoeger:
The Coverity "defect" is kind of pointless, because QString is reference counted, so passing it by value doesn't do any actual copy. But OK.

PR #4205, opened 2024-05-24
Cleanup: Improve the Use of 'Planned dive' and 'Manually added dive'.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

1. standardise the naming;
2. use it consistently;
3. apply the 'samples < 50' only when putting manually added dives into edit mode - everywhere else manually added dives should be treated as such.

### DIFF
--- a/core/divecomputer.c
+++ b/core/divecomputer.c
@@ -492,11 +492,6 @@ void add_extra_data(struct divecomputer *dc, const char *key, const char *value)
 	}
 }
 
-bool is_dc_planner(const struct divecomputer *dc)
-{
-	return same_string(dc->model, "planned dive");
-}
-
 /*
  * Match two dive computer entries against each other, and
  * tell if it's the same dive. Return 0 if "don't know",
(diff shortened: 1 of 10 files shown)


core/divecomputer.c
```diff
@@ -548,15 +543,30 @@ void free_dc(struct divecomputer *dc)
...
-	return dc && dc->samples <= 50 &&
-	       same_string(dc->model, manual_dc_name);
+	return dc && same_string(dc->model, planner_dc_name);
 }
 
-void make_manually_added_dc(struct divecomputer *dc)
+void make_planner_dc(struct divecomputer *dc)
 {
-	free((void *)dc->model);
+	if (dc->model)
```
bstoeger: I think in general we don't check for null when using `free()`, because the latter handles this gracefully.

That said, I'm currently converting all this to `std::string`, so it really doesn't matter.

bstoeger:
Yes, this is more consistent.

PR #4210, opened 2024-05-26
Desktop: Fix Gas Editing for Manually Added Dives.

### LINKED ISSUE #3966: BUG: gas switch event not serialized to XML on manually added dives
Guys, thanks for this awesome tool!

### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

When adding a dive manually that uses more than 1 cylinder, the gas change event is not saved to the XML. It is shown in the editor for the time of editing, but when de-selecting and re-selecting the dive, they dissappear from the profile. Having a look at the XML reveals, that they are not saved to file.

### Issue long description:

- When adding log entries via "Add dive", the gas change events are not applied in the profile viewer and not saved to XML
- When adding the same dive via "Plan dive" and planning it for the past, the gas change events are shown in the profile viewer and saved to XML
- A dive that was created manually via "Add dive" (not planned) can be edited via "Edit dive in planner" and if the gas change event is added there, it is saved to the XML

--> Therefore, the bug only persists when adding multi-gas dives via "Add dive"

### Operating system:

- Windows 10 Pro (10.0.19045 Build 19045)
- AMD Ryzen 7 7700X

### Subsurface version:

- Official Release v5.0.10
- Downloaded from homepage @ 27.09.23

### Steps to reproduce:

1. Create an empty log book
2. Add a dive **via the "Log > Add dive" option**
3. Configure the following cylinders:
   - 12L 200bar: 200bar-50bar, AIR
   - 6L 200bar: 200bar-60bar, EAN40
4. Create an arbitrary dive profle/curve in the editor
5. Try to add a gas switch for any given point
   - When I right-click a point in the profile, it does not even show me the EAN40, just AIR, (but that's a different bug...)
7. Save via CTRL+S
8. have a look at the generated XML file and see, that the gas switch has not been saved
9. de-select the dive by clicking somewhere else in the dive log list
10. now select the dive again and see that the gas switch is missing and the shown calculations are therefore incorrect

### Mentions:

I'm an expert on C++, but I'm quite OK with it. If you can give me a hint on where to find the methods related to event saving etc, I can see if I can fix it myself and open a PR... ;)

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

1. show the correct gasmix in the profile;
2. make gases available for gas switches in the profile after they have been added;
3. persist gas changes;
4. add air as a default gas when adding a dive.

### Related issues:

Fixes #3966.

### Additional information:

This still has problems when undoing a gas switch - instead of completely removing the gas switch it is just moved to the next point in the profile.

### DIFF
--- a/desktop-widgets/profilewidget.cpp
+++ b/desktop-widgets/profilewidget.cpp
@@ -52,7 +52,7 @@ void EmptyView::resizeEvent(QResizeEvent *)
 	update();
 }
 
-ProfileWidget::ProfileWidget() : d(nullptr), dc(0), originalDive(nullptr), placingCommand(false)
+ProfileWidget::ProfileWidget() : d(nullptr), dc(0), placingCommand(false)
 {
 	ui.setupUi(this);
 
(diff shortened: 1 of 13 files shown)


desktop-widgets/profilewidget.cpp
```diff
@@ -338,25 +364,36 @@ void ProfileWidget::stopAdded()
...
+	Setter s(placingCommand, true);
+	Command::editProfile(editedDive.get(), dc, Command::EditProfileType::MOVE, count);
+}
+
+void ProfileWidget::stopEdited()
+{
+	if (!editedDive)
+		return;
+
+	copy_events(get_dive_dc(editedDive.get(), dc), get_dive_dc(d, dc));
```
bstoeger, line 395: Revsisiting this, since I'm rebasing my C++ branch.

Does this work properly with undo? After all it seems that this overwrites the current dive outside of the undo system. Do you want me to look into that?

mikeller: Thanks for pointing this out. I had wondered why undo did not work, but it did not occur to me that I simply had to remove the `copy_events()` that I'd added before I switched to using the undo commands - doing this now in #4230, which works as it should, but feel free to include it in your changes if this makes it easier for you.

PR #4222, opened 2024-05-31
Profile: Add Gas Description to Disambiguate.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

1. add the gas description to the label on pressure graphs to disambiguate if multiple identical gasmixes are shown;
2. move the label to the right, where the end pressures will typically be more spread out than the starting pressures.

### Additional information:

image

### DIFF
--- a/profile-widget/diveprofileitem.cpp
+++ b/profile-widget/diveprofileitem.cpp
@@ -600,7 +600,9 @@ void DiveGasPressureItem::replot(const dive *d, int fromIn, int toIn, bool in_pl
 		}
 	}
 
+	bool showDescriptions = false;
 	for (int cyl = 0; cyl < pInfo.nr_cylinders; cyl++) {
+		showDescriptions = showDescriptions || same_gasmix_cylinder(get_cylinder(d, cyl), cyl, d, true) != -1;
 		if (act_segments[cyl].polygon.empty())
 			continue;
 		act_segments[cyl].cyl = cyl;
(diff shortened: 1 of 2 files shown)


profile-widget/diveprofileitem.cpp
```diff
@@ -618,45 +620,47 @@ void DiveGasPressureItem::replot(const dive *d, int fromIn, int toIn, bool in_pl
...
+		startAlignVar ^= Qt::AlignTop | Qt::AlignBottom;
 	}
 }
 
-void DiveGasPressureItem::plotPressureValue(double mbar, double sec, QFlags<Qt::AlignmentFlag> align, double pressure_offset)
+double DiveGasPressureItem::plotPressureValue(double mbar, double sec, QFlags<Qt::AlignmentFlag> align, double y_offset)
 {
 	const char *unit;
-	int pressure = get_pressure_units(lrint(mbar), &unit);
+	auto label = QString("%1%2").arg(get_pressure_units(lrint(mbar), &unit)).arg(unit);
```
bstoeger, line 644: I know that you only moved code here and not that it matters at all, but the 'pure' thing to do is `QStringLiteral()`, as the format string will then not be constructed at runtime.

That said, the more you work on the profile, the more it becomes a pressing issue to get a resolution for #3863. Rebasing is becoming very painful and error-prone. :(

profile-widget/diveprofileitem.cpp
```diff
@@ -618,45 +620,47 @@ void DiveGasPressureItem::replot(const dive *d, int fromIn, int toIn, bool in_pl
...
 }
 
-void DiveGasPressureItem::plotGasValue(double mbar, double sec, struct gasmix gasmix, QFlags<Qt::AlignmentFlag> align, double gasname_offset)
+void DiveGasPressureItem::plotGasValue(double mbar, double sec, const cylinder_t *cylinder, QFlags<Qt::AlignmentFlag> align, double x_offset, double y_offset, bool showDescription)
 {
-	QString gas = get_gas_string(gasmix);
+	QString gas = get_gas_string(cylinder->gasmix);
+	QString label;
+	if (showDescription)
+		label = QString("(%1) %2").arg(cylinder->type.description).arg(gas);
```
bstoeger, line 658: Not sure, but I think if both arguments are `QString`s, `.arg(s1, s2)` is preferred. `.arg(s1).arg(s2)` does some funky rewriting of the format string (replace %2 by %1, etc).

bstoeger (approved):

profile-widget/diveprofileitem.cpp
bstoeger: PS: of course the biggest pain is self-inflicted with my porting to C++.

mikeller: @bstoeger: Yes, I suspect that #3863 will be a rebase from hell, mostly due to changes that I made. But a lot of these changes were motivated by bug reports or shortcomings reported by users, and I don't think holding off this kind of thing for months on end is helpful if we want to keep our users engaged with helping us by providing feedback.
Is there anything that can be done to unblock you on #3863?

bstoeger: > Yes, I suspect that https://github.com/subsurface/subsurface/pull/3863 will be a rebase from hell, mostly due to changes that I made.

No worries, the worst problem will be my own conversions to C++.

I'm stuck with the link-order issues on mobile. The only fix I can think about is moving all the chart/QtQuick related stuff in a single "librabry". :( Then it needs of course testing on mobile, with pinch-zooming, etc.

Currently, conversion to C++ has priority for me, though.

PR #4223, opened 2024-05-31
core: remove put_format_loc()

bstoeger:
This was replaced by C++ functions in ae299d5e663c.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Another (trivial?) dead code removal.

### DIFF
--- a/core/format.cpp
+++ b/core/format.cpp
@@ -343,17 +343,6 @@ QString vqasprintf_loc(const char *fmt, va_list ap_in)
 	return ret;
 }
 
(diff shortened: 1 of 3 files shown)


mikeller:
It looks like ubuntu 16.04 that we are currently using to build the AppImage has got a consistency problem in its package repository, causing our AppImage builds to fail. I think we should give this a couple of days to get sorted before we start poking at it from our end.

PR #4226, opened 2024-05-31
test: fix subtle bug in testplan.cpp

bstoeger:
testplan.cpp had a subtle bug since converting from a fixed-size cylinder table to a dynamic cylinder table.

As noted in equipment.h, pointers to cylinders are *not* stable when the cylinder table grows. Therefore, a construct such as
        cylinder_t *cyl0 = get_or_create_cylinder(&dive, 0);
        cylinder_t *cyl1 = get_or_create_cylinder(&dive, 1);
        cylinder_t *cyl2 = get_or_create_cylinder(&dive, 2);
can give dangling cyl0 and cyl1 pointers. This was not an issue with the old table code, since it had a rather liberal allocation pattern. However, when switching to std::vector<>, the problem becomes active.

To "fix" this, simply access the highest index first. Of course, this should never be done in real code! Therefore, add a comment at each instance.

Quickly checked all other get_or_create_cylinder() calls and they seemed to be safe.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A subtle bug in the plan-test.

### DIFF
--- a/tests/testplan.cpp
+++ b/tests/testplan.cpp
@@ -52,9 +52,12 @@ void setupPlan(struct diveplan *dp)
 	struct gasmix ean36 = {{360}, {0}};
 	struct gasmix oxygen = {{1000}, {0}};
 	pressure_t po2 = {1600};
+	// Note: we add the highest-index cylinder first, because
+	// pointers to cylinders are not stable when reallocating.
+	// For testing OK - don't do this in actual code!
+	cylinder_t *cyl2 = get_or_create_cylinder(&dive, 2);
 	cylinder_t *cyl0 = get_or_create_cylinder(&dive, 0);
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
this is yet another sign that the way "modern" C++ code works is extremely unintuitive to the C programmer. And most of our C++ code was written (at least initially) by C programmers.

Looks good

bstoeger:
I don't see how this is related to C vs. C++.

In fact, in master the dynamic cylinder table is implemented in pure C. It is only when I ported the thing to C++ (in a local branch) that this problem became obvious.

bstoeger:
That said, the interface is indeed counter-intuitive and in hindsight `get_or_create_cylinder()` was an error.

PR #4228, opened 2024-06-01
Cleanup: Optimise String Handling.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Cleanup of the sub-optimal string handling in #4222.

### Related issues:

From https://github.com/subsurface/subsurface/pull/4222#discussion_r1621882092
### Mentions:

@bstoeger: There seems to be plenty of runtime construction of `QString` literals  already going on - this could possibly be solved by a `:g/QString("/s//QStringLiteral("/g` across all files - or what do you think?

### DIFF
--- a/profile-widget/diveprofileitem.cpp
+++ b/profile-widget/diveprofileitem.cpp
@@ -333,7 +333,7 @@ void DiveHeartrateItem::createTextItem(int sec, int hr, bool last)
 	int flags = last ? Qt::AlignLeft | Qt::AlignBottom :
 			   Qt::AlignRight | Qt::AlignBottom;
 	auto text = std::make_unique<DiveTextItem>(dpr, 0.7, flags, this);
-	text->set(QString("%1").arg(hr), getColor(HR_TEXT));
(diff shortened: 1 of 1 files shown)


bstoeger:
> There seems to be plenty of runtime construction of QString literals already going on - this could possibly be solved by a :g/QString("/s//QStringLiteral("/g across all files - or what do you think?

I doubt that this would be worth it. It's a few ns gained in UI code, where we do tons of UTF8<->UTF16 conversions and other much more costly operations. On the other hand, I'm also not opposing it.

mikeller:
> I doubt that this would be worth it. It's a few ns gained in UI code, where we do tons of UTF8<->UTF16 conversions and other much more costly operations. On the other hand, I'm also not opposing it.

Meh, 5 minutes grepping work, to get it done 'properly', even if it isn't a massive performance gain.
But I'll wait until we've got #3863 and #4198 merged.

PR #4231, opened 2024-06-02
[RFC] tests: make profile test work with non-C locales

bstoeger:
For reasons unknown to me, the profile test is executed with a weird locale, resulting in wrong formatting.

By setting the locale manually to "C", the tests start to work.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This makes the profile-test work for me. I feel that this is the wrong way to do it, because we don't have a `setlocale()` call in the code base, yet it seems to work for the rest of the code.

My locale at the start of the test is:
` LC_CTYPE=en_US.UTF-8;LC_NUMERIC=de_AT.UTF-8;LC_TIME=de_AT.UTF-8;LC_COLLATE=en_US.UTF-8;LC_MONETARY=de_AT.UTF-8;LC_MESSAGES=en_US.UTF-8;LC_PAPER=de_AT.UTF-8;LC_NAME=de_AT.UTF-8;LC_ADDRESS=de_AT.UTF-8;LC_TELEPHONE=de_AT.UTF-8;LC_MEASUREMENT=de_AT.UTF-8;LC_IDENTIFICATION=de_AT.UTF-8`
and not `C` as it should be!?

### Changes made:

1) Set locale to "C" in profile test.

### DIFF
--- a/tests/testprofile.cpp
+++ b/tests/testprofile.cpp
@@ -18,6 +18,8 @@
 
 void TestProfile::init()
 {
+	setlocale(LC_ALL, "C");
+
 	// Set UTF8 text codec as in real applications
 	QTextCodec::setCodecForLocale(QTextCodec::codecForMib(106));
 	// first, setup the preferences


mikeller (approved):

mikeller:
From past problems that myself and other contributors have had on getting tests to work locally vs. in CICD there are likely more pitfalls in locale settings when it comes to the tests for importing / exporting.

bstoeger:
I just don't understand why we don't have the same problem in the main application. What is the magic call that solves this?

mikeller:
> I just don't understand why we don't have the same problem in the main application. What is the magic call that solves this?

From what I have seen this is because some of our tests check for equality instead of equivalence. We start with a test file, which we import, and then export to a file. Then we check if the result is identical to the test file, without allowing for things like localised number formatting or a change of non-siginificant precision of identical numbers to be recognised as valid.

bstoeger:
Oh, I just checked and indeed the calculated profile data export uses the user's locale. Not sure if this is a good idea...

bstoeger:
Let's commit this so that I can run this test locally.

PR #4237, opened 2024-06-05
Add basic support to handle sub-second values

sjaeckel:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

libdivecomputer reports sub-second values, where the sub-second accuracy is lost when imported into Subsurface.

Fix this by saving those values on import and storing them in the xml.

Also read those values back when parsing the xml.

### DIFF
--- a/core/parse-xml.cpp
+++ b/core/parse-xml.cpp
@@ -340,9 +340,15 @@ static void temperature(const char *buffer, temperature_t *temperature, struct p
 static void sampletime(const char *buffer, duration_t *time)
 {
 	int i;
-	int hr, min, sec;
+	int hr, min, sec, ms;
 
-	i = sscanf(buffer, "%d:%d:%d", &hr, &min, &sec);
+	i = sscanf(buffer, "%d:%d.%d", &min, &sec, &ms);
+	if (i != 3) {
+		i = sscanf(buffer, "%d:%d:%d", &hr, &min, &sec);
+	} else {
+		hr = 0;
+		i = 4;
+	}
 	switch (i) {
 	case 1:
 		min = hr;
@@ -354,7 +360,11 @@ static void sampletime(const char *buffer, duration_t *time)
 		hr = 0;
 	/* fallthrough */
 	case 3:
+		ms = 0;
+	/* fallthrough */
+	case 4:
 		time->seconds = (hr * 60 + min) * 60 + sec;
+		time->ms = ms;
 		break;
 	default:
 		time->seconds = 0;
--- a/core/parse-xml.cpp
+++ b/core/parse-xml.cpp
@@ -339,37 +339,7 @@ static void temperature(const char *buffer, temperature_t *temperature, struct p
 
 static void sampletime(const char *buffer, duration_t *time)
 {
-	int i;
-	int hr, min, sec, ms;
-
-	i = sscanf(buffer, "%d:%d.%d", &min, &sec, &ms);
-	if (i != 3) {
-		i = sscanf(buffer, "%d:%d:%d", &hr, &min, &sec);
-	} else {
-		hr = 0;
-		i = 4;
-	}
-	switch (i) {
-	case 1:
-		min = hr;
-		hr = 0;
-	/* fallthrough */
-	case 2:
-		sec = min;
-		min = hr;
-		hr = 0;
-	/* fallthrough */
-	case 3:
-		ms = 0;
-	/* fallthrough */
-	case 4:
-		time->seconds = (hr * 60 + min) * 60 + sec;
-		time->ms = ms;
-		break;
-	default:
-		time->seconds = 0;
-		report_info("Strange sample time reading %s", buffer);
-	}
+	parse_duration(buffer, time, false);
 }
 
 static void offsettime(const char *buffer, offset_t *time)
@@ -388,17 +358,7 @@ static void offsettime(const char *buffer, offset_t *time)
 
 static void duration(const char *buffer, duration_t *time)
 {
-	/* DivingLog 5.08 (and maybe other versions) appear to sometimes
-	 * store the dive time as 44.00 instead of 44:00;
-	 * This attempts to parse this in a fairly robust way */
-	if (!strchr(buffer, ':') && strchr(buffer, '.')) {
-		std::string mybuffer(buffer);
-		char *dot = strchr(mybuffer.data(), '.');
-		*dot = ':';
-		sampletime(mybuffer.data(), time);
-	} else {
-		sampletime(buffer, time);
-	}
+	parse_duration(buffer, time, true);
 }
 
 static void percent(const char *buffer, fraction_t *fraction)
--- a/core/save-xml.cpp
+++ b/core/save-xml.cpp
@@ -236,7 +236,10 @@ static void save_sample(struct membuffer *b, struct sample *sample, struct sampl
 {
 	int idx;
 
-	put_format(b, "  <sample time='%u:%02u min'", FRACTION_TUPLE(sample->time.seconds, 60));
+	if (sample->time.ms == 0)
+		put_format(b, "  <sample time='%u:%02u min'", FRACTION_TUPLE(sample->time.seconds, 60));
+	else
+		put_format(b, "  <sample time='%u:%02u.%03u min'", FRACTION_TUPLE(sample->time.seconds, 60), sample->time.ms);
 	put_milli(b, " depth='", sample->depth.mm, " m'");
 	if (sample->temperature.mkelvin && sample->temperature.mkelvin != old->temperature.mkelvin) {
 		put_temperature(b, sample->temperature, " temp='", " C'");
--- a/core/units.c
+++ b/core/units.c
@@ -1,7 +1,10 @@
 // SPDX-License-Identifier: GPL-2.0
 #include "units.h"
+#include "core/errorhelper.h"
 #include "gettext.h"
 #include "pref.h"
+#include <stddef.h>
+#include <stdlib.h>
 
 #define IMPERIAL_UNITS                                                                                     \
         {                                                                                                  \
@@ -182,4 +185,50 @@ double get_weight_units(unsigned int grams, int *frac, const char **units)
 const struct units *get_units()
 {
 	return &prefs.units;
+}
+
+int parse_duration(const char *buffer, duration_t *time, bool dot_as_seperator)
+{
+	char *line = (char*)buffer;
+	size_t n = 0;
+	long values[3];
+	char separator;
+	while (n < sizeof(values)/sizeof(values[0])) {
+		values[n++] = strtol(line, &line, 10);
+		separator = *line;
+		line++;
+		/* special case of `mm.ss`
+		 * DivingLog 5.08 (and maybe other versions) appear to sometimes
+		 * store the dive time as 44.00 instead of 44:00 */
+		if (dot_as_seperator && n == 1 && separator == '.')
+			continue;
+		if (separator != ':')
+			break;
+	}
+	switch (n) {
+	case 1:
+		/* mm */
+		time->seconds = values[0] * 60;
+		time->ms = 0;
+		goto exit_out;
+	case 2:
+		/* mm:ss */
+		time->seconds = values[0] * 60 + values[1];
+		break;
+	case 3:
+		/* hh:mm:ss */
+		time->seconds = values[0] * 360 + values[1] * 60 + values[2];
+		break;
+	default:
+		time->seconds = 0;
+		time->ms = 0;
+		report_info("Strange sample time reading %s", buffer);
+		return 0;
+	}
+	long ms = 0;
+	if (separator == '.')
+		ms = strtol(line, &line, 10);
+	time->ms = ms;
+exit_out:
+	return line - buffer;
 }
--- a/core/libdivecomputer.cpp
+++ b/core/libdivecomputer.cpp
@@ -396,6 +396,7 @@ sample_cb(dc_sample_type_t type, const dc_sample_value_t *pvalue, void *userdata
 		// Mark depth as negative
 		sample = prepare_sample(dc);
 		sample->time.seconds = value.time / 1000;
+		sample->time.ms = value.time % 1000;
 		sample->depth.mm = -1;
 		// The current sample gets some sticky values
 		// that may have been around from before, these
--- a/core/load-git.cpp
+++ b/core/load-git.cpp
@@ -693,13 +693,16 @@ static struct sample *new_sample(struct git_parser_state *state)
 
 static void sample_parser(char *line, struct git_parser_state *state)
 {
-	int m, s = 0;
+	int m, s = 0, ms = 0;
 	struct sample *sample = new_sample(state);
 
 	m = strtol(line, &line, 10);
 	if (*line == ':')
 		s = strtol(line + 1, &line, 10);
 	sample->time.seconds = m * 60 + s;
+	if (*line == '.')
+		ms = strtol(line + 1, &line, 10);
+	sample->time.ms = ms;
 
 	for (;;) {
 		char c;
--- a/core/load-git.cpp
+++ b/core/load-git.cpp
@@ -693,16 +693,9 @@ static struct sample *new_sample(struct git_parser_state *state)
 
 static void sample_parser(char *line, struct git_parser_state *state)
 {
-	int m, s = 0, ms = 0;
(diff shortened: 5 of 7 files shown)


core/save-xml.cpp
```diff
@@ -236,7 +236,10 @@ static void save_sample(struct membuffer *b, struct sample *sample, struct sampl
 {
 	int idx;
 
-	put_format(b, "  <sample time='%u:%02u min'", FRACTION_TUPLE(sample->time.seconds, 60));
+	if (sample->time.ms == 0)
+		put_format(b, "  <sample time='%u:%02u min'", FRACTION_TUPLE(sample->time.seconds, 60));
+	else
+		put_format(b, "  <sample time='%u:%02u.%03u min'", FRACTION_TUPLE(sample->time.seconds, 60), sample->time.ms);
```
mikeller, line 242: This breaks backwards compatibility with all existing versions of Subsurface. This will likely lead to a flurry of bug reports from users of dive computers reporting sub-second timestamps when they find out that they are unable to open their log files with Subsurface on some of their devices.

It also drops feature parity between file based and cloud based log files, likely leading to another flurry of bug reports.

mikeller (changes requested):
I don't think this should be considered for a merge until we are at are ready to provide solutions for the bug reports that are to be expected after this has been merged.

core/parse-xml.cpp
```diff
@@ -340,9 +340,15 @@ static void temperature(const char *buffer, temperature_t *temperature, struct p
 static void sampletime(const char *buffer, duration_t *time)
 {
 	int i;
-	int hr, min, sec;
+	int hr, min, sec, ms;
 
-	i = sscanf(buffer, "%d:%d:%d", &hr, &min, &sec);
+	i = sscanf(buffer, "%d:%d.%d", &min, &sec, &ms);
```
mikeller: What is your reason for not leaving the support for (optional) hours in the version with millisecond support?

core/save-xml.cpp
sjaeckel: > This breaks backwards compatibility with all existing versions of Subsurface. This will likely lead to a flurry of bug reports from users of dive computers reporting sub-second timestamps when they find out that they are unable to open their log files with Subsurface on some of their devices.
> 
> It also drops feature parity between file based and cloud based log files, likely leading to another flurry of bug reports.

I've created my logbook with this version, synced it to the cloud, edited it with the mobile version, synced it back, and opened it with the latest stable version. I couldn't see any problem.

You're right that while storing this to the cloud, the precision is lost again. I forgot to check that and will have a look.

core/parse-xml.cpp
sjaeckel: Because I prefer to be strict and ignore what others maybe do, I only want to ingest what I created and leave the rest up to the previous path.

core/save-xml.cpp
sjaeckel: After the latest commit I added we also export the sub-second part when storing to Git and we're able to read it again.

As expected the precision is still lost when storing it with an older version in between, but "it works"^TM.

core/parse-xml.cpp
mikeller: And what benefit do you see in this approach?

sjaeckel: I try not to assume something I'm not sure of whether it even exists.
Do you have a real use case in mind? Some software that already uses this format?

mikeller: And what disadvantages do you see in implementing a parser that is accomodating to sensible future formats?

core/save-xml.cpp
mikeller: This does not work for me - after importing a dive with ms resolution timestamps, saving it to the cloud, closing the log, opening the cloud storage again, and saving the previously imported dive into a file the ms resolution is gone.

core/parse-xml.cpp
sjaeckel: YAGNI.

We can still do that once this future format pops up.

core/save-xml.cpp
sjaeckel: Indeed, I missed that. Should be fixed.

sjaeckel: Thanks for the review and thorough testing :)

I still have to get used to the code base and all the use cases and flows that I'm not aware of yet.

core/parse-xml.cpp
mikeller: I don't really see how 'YAGNI' should be a justification for adding unneeded complexity by supporting a format of 'hh:mm:ss OR mm:ss.xxx BUT NOT hh:mm:ss.xxx' instead of just 'hh:mm:ss.xxx'...

mikeller:
I think one big (but not immediately obvious) issue of the proposed approach is that it only adds support for carrying milliseconds as a 'payload', but it does not add support for the use of milliseconds in Subsurface's logic. All of the program logic will treat a log with samples with millisecond resolution and a sample frequency > 1 Hz as a log of a sequence of second resolution samples with lots of duplicate timestamps.
While this isn't explicitly prevented from working in the logic, it also isn't a use case that the logic has been designed for, and only works because we are over-specific in having timestamps and the ordering of the sample list to define the ordering of the sequence of samples. But all of the logic uses sample timestamps for all calculations around samples, and it's just coincidence that this does not (yet) break in obvious ways.
To get an idea of Subsurface breaking on unsupported input in more obvious ways, try importing a log file (can be a manually edited XML log) that has <= 1 Hz samples with swapped ordering of the timeline. Opening this will work (just as it does for a log with 'millisecond timestamps' and sample frequency > 1 Hz, but in this case the breakage is much more obvious.
So better ways to implement this will be:
- change the internal timestamps to be in milliseconds, and adapt all importers / exporters and logic using timestamps in calculations to match;
- change all logic to re-create the millisecond timestamp from seconds and ms whenever ms is not 0.

bstoeger:
Fully agree with @mikeller's comments.

> change the internal timestamps to be in milliseconds, and adapt all importers / exporters and logic using timestamps in calculations to match;

Since we're currently touching the whole code base anyway, I think this is the way to go. Having a split `duration_t` is inconsistent with the rest of the units.

Of course, it could also be done in two steps: first with a split `duration_t` and once all the importers/exporters work with a unified version.

BTW: There are a number of unit libraries. Could we take some inspiration from that?

bstoeger:
Of course, what bites us in the ass here, is that we are very inconsistent about using the unit-types. Which is certainly due to the fact that their usage (notably comparison) is quite cumbersome.

bstoeger:
BTW: This really might be a use case for going to C++20:
```
#include <cstdint>
#include <compare>

struct duration_t {
    int32_t milliseconds;
    auto operator<=>(const duration_t &) const = default;
};

int test(duration_t a, duration_t b)
{
    return a < b;
}


int test2(duration_t a, duration_t b)
{
    return a > b;
}


int test3(duration_t a, duration_t b)
{
    return a == b;
}
```

sjaeckel:
> So better ways to implement this will be:
> 
> * change the internal timestamps to be in milliseconds, and adapt all importers / exporters and logic using timestamps in calculations to match;
> * change all logic to re-create the millisecond timestamp from seconds and ms whenever ms is not 0.

I'm completely with you in that regard, but this is not the target of this PR.

What I want to solve here is to simply not loose the precision. The real endeavor of reworking everything to be sub-second based will likely take multiple iterations and steps over weeks/months and I don't want to loose all those details of my dives in the meantime, and it doesn't matter to me if the calculations are correct (for now).

I'm not sure if that's really the case, but I think that the fact that nobody started this discussion before shows that it's no real problem for the existing user base. So there's no pressure in getting the entire core/calculations to sub-second accuracy.

> Of course, it could also be done in two steps: first with a split `duration_t` and once all the importers/exporters work with a unified version.

Exactly. Iterate slowly but constantly, instead of trying to fix the world in one big PR.

> Of course, what bites us in the ass here, is that we are very inconsistent about using the unit-types.

> BTW: There are a number of unit libraries. Could we take some inspiration from that?

> BTW: This really might be a use case for going to C++20:

I like the direction we're heading here :)

core/parse-xml.cpp
sjaeckel: > 'hh:mm:ss OR mm:ss.xxx BUT NOT hh:mm:ss.xxx' instead of just 'hh:mm:ss.xxx'...

Currently this enables `mm:ss.xxx` (as that's the one Subsurface uses when it writes the values) OR all the previous supported variations.

If I'm not mistaken that's all previous supported variations of a duration:

```
(-)mm
mm.ss -- that one's patched to mm:ss before invocation of `sampletime()`
(-)mm:ss
(-)hh:mm:ss
```

So you want that all those formats also get sub second support?

What should we do with `mm.ss`? Keep on patching it before or add support built-in?

dirkhh:
> Fully agree with @mikeller's comments.
> 
> > change the internal timestamps to be in milliseconds, and adapt all importers / exporters and logic using timestamps in calculations to match;
> 
> Since we're currently touching the whole code base anyway, I think this is the way to go. Having a split `duration_t` is inconsistent with the rest of the units.

I still fundamentally think that ms resolution is nonsense, just as mm depth is nonsense.

bstoeger:
Yes, I'm only commenting on technical feasibility, not whether it makes sense.

atdotde:
Just a quick question: What happens to a dive with ms information if you modify the profile with "edit dive in planner"?

bstoeger:
I would say it's going to be downsampled to 10s or so. Don't we do that already?

core/parse-xml.cpp
mikeller: I'd treat `mm.ss` as a special case - it looks like this was added in specifically to fix inconsistent use of the format in one specific version of a third party app.
And for adding milliseconds to `mm` a reasonable approach to write this duration will be to use `mm:00.xxx`, yielding the `mm:ss.xxx` case - so all we are left with is `mm`, `mm:ss[.xxx]`, and `hh:mm:ss[.xxx]`, plus the special case.

mikeller:
> What I want to solve here is to simply not loose the precision. The real endeavor of reworking everything to be sub-second based will likely take multiple iterations and steps over weeks/months and I don't want to loose all those details of my dives in the meantime, and it doesn't matter to me if the calculations are correct (for now).

The problem is that Subsurface was not designed to support logs with multiple samples for the same timestamp. While we are not explicitly preventing logs with an incorrect sequence of sample timestamps from being opened (maybe we should), opening such logs will break subsurface in subtle or not so subtle ways.

I am aware that this isn't a problem that is introduced with this pull request, but before it the simple fix if users report problems when importing from devices with sub-second sampling rates will be to eliminate all but one sample per second when importing from such devices.

So if what you want is to preserve logs with sub-second samples then a better way will be to so in a branched version of subsurface until all the changes are ready to properly support this without the risk of breaking things because Subsurface wasn't designed with this in mind.

core/parse-xml.cpp
sjaeckel: Fine like that?

mikeller: Yes, the format support looks good now, thank you. Not sure if there is a better way to do the parsing in C++, so this is potentially another case of waiting for #4198 to drop and then do it the C++ way - @bstoeger should know more.

sjaeckel:
> I still fundamentally think that ms resolution is nonsense, just as mm depth is nonsense.

I still fundamentally think that a factor 1000 doesn't matter if one has 64bit integers available.

> The problem is that Subsurface was not designed to support logs with multiple samples for the same timestamp. While we are not explicitly preventing logs with an incorrect sequence of sample timestamps from being opened (maybe we should), opening such logs will break subsurface in subtle or not so subtle ways.
> 
> I am aware that this isn't a problem that is introduced with this pull request, but before it the simple fix if users report problems when importing from devices with sub-second sampling rates will be to eliminate all but one sample per second when importing from such devices.
> 
> So if what you want is to preserve logs with sub-second samples then a better way will be to so in a branched version of subsurface until all the changes are ready to properly support this without the risk of breaking things because Subsurface wasn't designed with this in mind.

I don't see the point in that argumentation. Subsurface already "supports" (as in "doesn't break on") sub-second values now. Which I find btw. pretty nice and shows the robustness of the architecture. Nobody complained too much about the wrong calculations or if someone did, the solution was by telling them that they should fix the values themselves. I don't understand why this PR can't simply be merged. The only difference that it makes to the current situation is that it will improve the future situation. Yes the calculations will be wrong, but they are already.

Btw. with this PR merged we'd even have a "cheap"^TM way to make the calculations with sub-second samples right again, with a work-around: only take the values where `ms==0` into account for the calculations.

core/parse-xml.cpp
bstoeger: I'm not aware of a better way than `scanf` in the standard library.

I'm not following closely, but there appears to be a proposal for an `std::format` counterpart: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1729r1.html
However, that will not be available for many years, if ever.

sjaeckel: > I'm not aware of a better way than `scanf` in the standard library.

Could you please have a look at the latest version I implemented in `core/units.c`? That one doesn't use `scanf` anymore.

bstoeger: Where do I see that?

sjaeckel: In this PR, the latest commit.

core/units.c
```diff
@@ -183,3 +186,49 @@ const struct units *get_units()
...
 	return &prefs.units;
 }
+
+int parse_duration(const char *buffer, duration_t *time, bool dot_as_seperator)
+{
+	char *line = (char*)buffer;
+	size_t n = 0;
+	long values[3];
+	char separator;
+	while (n < sizeof(values)/sizeof(values[0])) {
```
bstoeger, line 196: please use `std::size(values)` - more succinct.

core/units.c
```diff
@@ -183,3 +186,49 @@ const struct units *get_units()
 {
 	return &prefs.units;
 }
+
+int parse_duration(const char *buffer, duration_t *time, bool dot_as_seperator)
+{
+	char *line = (char*)buffer;
```
bstoeger, line 192: I don't like that const cast, but understand its raison d'être: the `strtol` interface.

If you want to get rid of it, you could use a helper function such as
```
#include <charconv>

template <typename T>
static const char *parse_int(const char *buf, size_t size, T &i)
{
    return std::from_chars(buf, buf + size, i).ptr;
}
```
or directly write below
```
line = std::from_chars(line, line + 10, values[n++]).ptr;
```
But from my side, you can also leave it as as.

core/units.c
```diff
@@ -183,3 +186,49 @@ const struct units *get_units()
...
+	size_t n = 0;
+	long values[3];
+	char separator;
+	while (n < sizeof(values)/sizeof(values[0])) {
+		values[n++] = strtol(line, &line, 10);
+		separator = *line;
+		line++;
```
bstoeger, line 199: What if we hit the terminal `\0' here? Then line is now pointing past the buffer?

core/units.c
```diff
@@ -183,3 +186,49 @@ const struct units *get_units()
...
+			continue;
+		if (separator != ':')
+			break;
+	}
+	switch (n) {
+	case 1:
+		/* mm */
+		time->seconds = values[0] * 60;
+		time->ms = 0;
+		goto exit_out;
```
bstoeger, line 213: I don't understand the control-flow here.

This seems to be considering the case `11.111` when `dot_as_separator` is `false`. Ok, but the code above skipped over the `.`. Is this what we want (i.e. consume the `.`)?

core/units.c
```diff
@@ -183,3 +186,49 @@ const struct units *get_units()
...
+	default:
+		time->seconds = 0;
+		time->ms = 0;
+		report_info("Strange sample time reading %s", buffer);
+		return 0;
+	}
+	long ms = 0;
+	if (separator == '.')
+		ms = strtol(line, &line, 10);
+	time->ms = ms;
```
bstoeger, line 231: So the last separator was not a `.`, we skipped it. However, if the last separator was a `.` we parse milliseconds, but don't skip the next character.

This feels inconsistent to me.

core/parse-xml.cpp
bstoeger: I'm not against this code, but it seems a bit inconsistent to me with respect to whether or not the separator is consumed. It would be great if you could write some tests (e.g. in `tests/testhelper.cpp`?) that makes the expected outcome clear.

PR #4245, opened 2024-06-09
Planner: Improve Gas Handling in CCR Mode.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This has become a bit of a catch-all overhaul of a large portion of the planner - I started out wanting to improve the CCR mode, but then as I started pulling all the other threads that needed addressing started to come with it.

Improve how the gas selection is handled when planning dives in CCR mode, by making the type (OC / CCR) of segments dependent on the gas use type that was set for the selected gas.

One point that is still open is to hide gas usage graphs in the planner profile if the gas isn't used for OC, as there is no way to meaningfully interpolate such usage.
### Changes made:

1. improve how the gas selection is handled when planning dives in CCR mode, by making the type (OC / CCR) of segments dependent on the gas use type that was set for the selected gas;
2. add a preference to allow the user to chose to use OC gases as diluent, in a similar fashion to the original implementation;
3. hide gases that cannot be used in the currently selected dive mode in all drop downs;
4. include usage type in gas names if this is needed;
5. hide columns and disable elements in the 'Dive planner points' table if they can they can not be edited in the curently selected dive mode;
6. visually identify gases and usage types that are not appropriate for the currently selected dive mode;
7. move the 'Dive mode' selection to the top of the planner view, to accommodate the fact that this is a property of the dive and not a planner setting;
8. show a warning instead of the dive plan if the plan contains gases that are not usable in the selected dive mode;
9. fix the data entry for the setpoint in the 'Dive planner points' table;
10. fix problems with enabling / disabling planner settings when switching between dive modes;
11. refactor some names to make them more appropriate for their current usage.

### Additional information:

image
--
image
--
image
--
image
--
image
--
image

### Documentation change:

Included in pull request.


atdotde:
I think this is much better than what we had before. I am not a CCR diver myself but my impression is that some people would use the diluent as bail out gas. Am I missing something or is this no longer possible since a cylinder is either diluent or OC?

mikeller:
@atdotde: Yes, it is indeed common for CCR divers to use the same offboard tank for diluent and bailout gas. However, practically every current dive computer with CCR support requires the user to configure this gas twice, once as a diluent, and once as a bailout gas. So setting them up separately in the planner will result in a gas list that is more similar to what they will be getting from the dive computer when downloading the actual dive.
And if a user wants to go the other way and configure only one gas for the combined diluent / bailout gas, they can check 'allow open circuit gas to be used as bailout', and they will be able to do this, and choose the gas use of open circuit cases in each segment separately. Gases configured as diluent will still only be available in closed circuit mode, as on-board diluent gases are often not configured with a second stage to breathe them from.

mikeller:
@bstoeger I'd appreciate if you could have a look at this. It seems to have rebased relatively well, and functional testing is ok, but I am a bit worried about having to comment out [this line](https://github.com/subsurface/subsurface/pull/4245/files#diff-dc0c576c47a7fad7e426860bfd438c309f836fbf35c5e99336821e064e00fec0R559) in order to avoid a SEGFAULT on program start:

```
dive::get_dc (this=0x0, nr=0)
    at /home/mikeller/git/subsurface/core/dive.cpp:2556
2556		if (dcs.empty()) // Can't happen!
(gdb) bt
#0  dive::get_dc (this=0x0, nr=0)
    at /home/mikeller/git/subsurface/core/dive.cpp:2556
#1  0x000055555585ab01 in DivePlannerPointsModel::setDiveMode
    (this=0x555555ce07a0 <DivePlannerPointsModel::instance()::self>, mode=mode@entry=0) at /home/mikeller/git/subsurface/qt-models/diveplannermodel.cpp:593
#2  0x0000555555771b2e in PlannerWidgets::setDiveMode
    (this=0x555556266430, mode=0)
    at /home/mikeller/git/subsurface/desktop-widgets/diveplanner.cpp:684
#3  0x0000555555773b6c in PlannerWidgets::PlannerWidgets
    (this=this@entry=0x555556266430)
    at /home/mikeller/git/subsurface/desktop-widgets/diveplanner.cpp:559
#4  0x00005555557b9086 in MainWindow::MainWindow
    (this=this@entry=0x555555f7fde0)
    at /home/mikeller/git/subsurface/desktop-widgets/mainwindow.cpp:141
#5  0x000055555570018d in GlobalObject<MainWindow>::GlobalObject
    (this=0x555555f7fde0) at /home/mikeller/git/subsurface/core/globals.h:28
#6  make_global<MainWindow> ()
    at /home/mikeller/git/subsurface/core/globals.h:38
#7  init_ui () at /home/mikeller/git/subsurface/subsurface-helper.cpp:62
#8  0x00005555556f8502 in main (argc=<optimized out>, argv=<optimized out>)
    at /home/mikeller/git/subsurface/subsurface-desktop-main.cpp:84
```

This makes it look like the initialisation of `dcs` in `core/dive.cpp` isn't working the way it should.
(I suspect that the initialisation that I had to comment out is redundant as we are setting the divemode when entering the planner, but I couldn't exclude there being need for it in edge cases.)

bstoeger:
I'm currently away and will look closer on weekend.

> This makes it look like the initialisation of dcs in core/dive.cpp isn't working the way it should.

From your backtrace it looks like `get_dc()` is called on a null dive.

You must only call functions of the `DivePlannerPointsModel` once the dive was set with `createSimpleDive()` or `loadFromDive()`.

Note: I didn't come up with this crazy model/view thing.

bstoeger:
One suggestion would be to move this line:
```
DivePlannerPointsModel::instance()->setDiveMode(mode);
```
to after `createSimpleDive()` and/or `loadFromDive()`.

Allternatively to make dive-mode a parameter of these functions.

Though I'm not sure if it makes sense for `loadFromDive()`? Perhaps there the CCR mode should be extracted from the dive? I'm not really familiar with the planner and rebreathers.

qt-models/diveplannermodel.cpp
```diff
@@ -537,14 +588,16 @@ int DivePlannerPointsModel::gfLow() const
 	return diveplan.gflow;
 }
 
-void DivePlannerPointsModel::setRebreatherMode(int mode)
+void DivePlannerPointsModel::setDiveMode(int mode)
```
bstoeger: This function looks like a layering violation. The model shouldn't modify the dive. I will have to untangle this.

mikeller: > This function looks like a layering violation. The model shouldn't modify the dive. I will have to untangle this.

I think the whole layering is somewhat weird around the DivePlannerPointsModel, as this 'owns' a Dive (that isn't part of the dive list either). Besides the layering violation this is also creating a lot of headaches coming from the 'dual use' of the Dive structure.
But if we want to refactor this, can I propose that we do this outside of this pull request that is intended to bring a quite substantial functional change?

mikeller:
@bstoeger: I have stripped out the initialisation of the dive mode in the constructor - it sounds like this just worked by coincidence. The correct dive mode is set when the dive used by the planner is loaded / initialised anyway.
This should be good for a review and merge now.

qt-models/diveplannermodel.cpp
bstoeger: > I think the whole layering is somewhat weird around the DivePlannerPointsModel, as this 'owns' a Dive (that isn't part of the dive list either). 

It doesn't own the dive - it only holds a reference to it. The owners are `PlannerWidgets` (for planned dives) and `ProfileWidget` (for edited dives) respectively. The model shouldn't modify the dive (yes, I am painfully aware of `createSimpleDive()`).

I would suggest something like this, i.e. let the owner manipulate the dive and then inform the model of the change (totally untested):
```
diff --git a/desktop-widgets/diveplanner.cpp b/desktop-widgets/diveplanner.cpp
index 64ce4758a..774b5cecc 100644
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -555,7 +555,6 @@ PlannerWidgets::PlannerWidgets() :
 	plannerSettingsWidget(this)
 {
 	plannerWidget.setDiveMode(OC);
-	//setDiveMode(OC);
 
 	connect(plannerDetails.printPlan(), &QPushButton::pressed, this, &PlannerWidgets::printDecoPlan);
 	connect(DivePlannerPointsModel::instance(), &DivePlannerPointsModel::calculatedPlanNotes,
@@ -680,7 +679,11 @@ void PlannerWidgets::printDecoPlan()
 
 void PlannerWidgets::setDiveMode(int mode)
 {
-	DivePlannerPointsModel::instance()->setDiveMode(mode);
+	struct divecomputer *dc = planned_dive->get_dc(dcNr);
+	if (dc)
+		dc->divemode = (divemode_t) mode;
+
+	DivePlannerPointsModel::instance()->cylindersChanged();
 	plannerWidget.setColumnVisibility(mode);
 	plannerSettingsWidget.setBailoutVisibility(mode);
 }
diff --git a/qt-models/diveplannermodel.cpp b/qt-models/diveplannermodel.cpp
index f203d772f..a84c0e587 100644
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -588,12 +588,10 @@ int DivePlannerPointsModel::gfLow() const
 	return diveplan.gflow;
 }
 
-void DivePlannerPointsModel::setDiveMode(int mode)
+void DivePlannerPointsModel::cylindersChanged()
 {
-	struct divecomputer *dc = d->get_dc(dcNr);
-	if (dc)
-		dc->divemode = (divemode_t) mode;
-
+	if (!d)
+		return;
 	cylinders.updateDive(d, dcNr);
 
 	emitDataChanged();
diff --git a/qt-models/diveplannermodel.h b/qt-models/diveplannermodel.h
index a3bc25bd5..9acdc7660 100644
--- a/qt-models/diveplannermodel.h
+++ b/qt-models/diveplannermodel.h
@@ -66,7 +66,7 @@ public:
 
 	void loadFromDive(dive *d, int dcNr);
 	void addStop(int millimeters, int seconds);
-	void setDiveMode(int mode);
+	void cylindersChanged();
 public
 slots:
 	void addDefaultStop();
```

I think I tried to remove the `DivePlannerPointsModel` global a number of times, but gave up every time. Let's wait for a rainy afternoon, when I find more time...

bstoeger: PS: note that the `if (dc)` is unnecessary, because every dive has at least one "dive computer" and `get_dc()` never returns zero. Manually added dives get a special dive computer.

Not sure if we should ever support computer-less dives. If not, we might change the return value of `get_dc()` to `divecomputer &`.

mikeller:
@bstoeger: I have fixed the layering violations that you pointed out - unfortunately this was not possible in the mobile version, so the (unused) planner code in this is now further broken.

PR #4247, opened 2024-06-14
Filter: Fix Loading of Negated Conditions from git.

### LINKED ISSUE #4246: unknown filter preset constrain, when loading/importing from cloud
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I use a saved filter preset, which looks in the xml file like:

<filterpreset name='name'>
  <constraint type='tags' string_mode='starts_with'>ccr</constraint>
  <constraint type='tags' string_mode='starts_with' negate='1'>training</constraint>
</filterpreset>

When I save it to the cloud storage, and later reimport, the error
  unknown filter preset constraint key/value pair (negate data="/)
is shown and the negate part is lost...

### Operating system:
 
running on a linux pc 

### Subsurface version:

build from recent git (b126fcc)

### Steps to reproduce:

1) create a filter preset using a "is not" like shown above.
2) save to cloud storage
3) re-open/load from cloud storage

### Current behavior:

xml file is changed (negate from filter preset is erased)

### Expected behavior:

xml should stay unchanged

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix loading of the negation of filter conditions. Unlike other conditions that are persisted as `<key>="<value>"`, this is persisted to git as `negate`.
This fix remediates this for all cases where the condition has already been saved to the cloud storage.

### Related issues:

Fixes #4246.

### Additional information:

Saving to XML takes a different approach and indicates negated conditions with `negate="1"`, making it identical to all other attributes. The question is if this approach should be implemented in addition to the above fix, in order to unify the storage format.

### DIFF
--- a/core/load-git.cpp
+++ b/core/load-git.cpp
@@ -357,22 +357,26 @@ static char *parse_keyvalue_entry(void (*fn)(void *, const char *, const std::st
 		line++;
 	}
 
-	if (c == '=')
+	if (c != 0)
 		*line++ = 0;
 
-	char *start_val = line;
-	while ((c = *line) != 0) {
-		if (isspace(c))
-			break;
-		line++;
+	std::string val;
+	if (c == '=') {
+		const char *start_val = line;
+		while ((c = *line) != 0) {
+			if (isspace(c))
+				break;
+			line++;
+		}
+
+		/* Did we get a string? Take it from the list of strings */
+		val = start_val[0] == '"' ? pop_cstring(state, key)
+					  : std::string(start_val, line - start_val);
+
+		if (c)
+			line++;
+
 	}
-
-	/* Did we get a string? Take it from the list of strings */
-	std::string val = start_val[0] == '"' ? pop_cstring(state, key)
-					      : std::string(start_val, line - start_val);
-
-	if (c)
-		line++;
 
 	fn(fndata, key, val);
 	return line;


core/load-git.cpp
```diff
@@ -357,22 +357,28 @@ static char *parse_keyvalue_entry(void (*fn)(void *, const char *, const std::st
...
+		/* Did we get a string? Take it from the list of strings */
+		val = start_val[0] == '"' ? pop_cstring(state, key)
+					  : std::string(start_val, line - start_val);
 
-	if (c)
-		line++;
+		if (c)
+			line++;
+
+	} else if (isspace(c)) {
```
bstoeger: I don't understand the point of this `else if` branch. Whether this branch is taken or not, val is the empty string, no?

core/load-git.cpp
```diff
@@ -357,22 +357,28 @@ static char *parse_keyvalue_entry(void (*fn)(void *, const char *, const std::st
...
 
-	char *start_val = line;
-	while ((c = *line) != 0) {
-		if (isspace(c))
-			break;
-		line++;
-	}
+	std::string val;
+	if (c == '=') {
+		char *start_val = line;
```
bstoeger: I understand that you are just using old code here. But is there any reason that this isn't `const char *`?

core/load-git.cpp
```diff
@@ -357,22 +357,28 @@ static char *parse_keyvalue_entry(void (*fn)(void *, const char *, const std::st
 		line++;
 	}
 
-	if (c == '=')
+	if (c != 0)
```
bstoeger, line 360: PS: Could it be that this is actually the only line needed to fix the problem?

bstoeger: Answering to myself: no, because the next key would be interpreted as a value. :(

core/load-git.cpp
mikeller: Correct, the `else` can be avoided.

core/load-git.cpp
mikeller: Copy / pasta.

bstoeger (approved):
lgtm.

Final question: Have you checked that TestGitStorage is unchanged (it currently fails one test for me on master)?

mikeller:
> Final question: Have you checked that TestGitStorage is unchanged (it currently fails one test for me on master)?

Yes, the tests do run fine for me on this branch and on `master` (and they are also run for qute a few of our CICD runs).
But some of the tests seem to be affected by locale settings in the environment - could this be affecting you?

bstoeger:
Oh, I thought this one was disabled because it failed so often.

The loading/saving depending on the locale is a scary proposition, so I will have to investigate...

bstoeger:
This is the diff of the failed test:
```
--- ./SampleDivesV3plus10local.ssrf     2024-06-15 23:07:07.993540062 +0200
+++ ./SampleDivesV3plus10viacloud.ssrf  2024-06-15 23:07:09.414452523 +0200
@@ -260,13 +260,13 @@
 </dive>
 </trip>
 <trip date='2011-01-01' time='18:00:00' location='10th test dive, 3 tanks, 1 tank change'>
-<dive number='10' sac='38.319 l/min' otu='32' cns='11%' divesiteid=' 2478ae1' date='2011-01-01' time='18:00:00' duration='30:00 min'>
+<dive number='10' sac='38.334 l/min' otu='32' cns='11%' divesiteid=' 2478ae1' date='2011-01-01' time='18:00:00' duration='30:00 min'>
   <notes>Shows two pressure plots, each from 200 to 100bar and gaschange event at 10 min mark.</notes>
   <cylinder size='9.987 l' workpressure='206.843 bar' description='AL72' o2='33.0%' start='200.0 bar' end='100.0 bar' />
   <cylinder size='9.987 l' workpressure='206.843 bar' description='AL72' o2='50.0%' start='200.0 bar' end='100.0 bar' />
   <cylinder size='9.987 l' workpressure='206.843 bar' description='AL72' o2='75.0%' start='200.0 bar' end='100.0 bar' />
   <divecomputer last-manual-time='30:00 min'>
-  <depth max='30.0 m' mean='15.0 m' />
+  <depth max='30.0 m' mean='14.998 m' />
   <temperature air='27.0 C' water='26.0 C' />
   <event time='10:00 min' type='11' value='50' name='gaschange' cylinder='1' o2='50.0%' />
   <sample time='0:00 min' depth='0.0 m' />
```
Not sure what to make of that...

mikeller:
> Not sure what to make of that...

Rounding errors - not that uncommon when dealing with floats. What I've seen done in other projects is adding a fudge factor to the tests to make them accept results within a range instead of expecting exact results. Unfortunately this then makes the tests ignorant to drift.

bstoeger:
A rounding error in the 4th significant digit would point to a severely unstable algorithm.

A mean value calculation being wrong in the 5th digit is also very sketchy.

Something is amiss...

PR #4248, opened 2024-06-14
Filtering: Usability Improvements to Filter Constraints.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Some improvements.

### Changes made:

1. for constraints where the condition is applied to any one of a list of items, show the condition as 'any X';
2. when filtering for tags, do not include divemode as a 'tag' - this is hard to impossible for users to understand, and we do have a dedicated condition for the dive mode.

### Additional information:

image

### DIFF
--- a/core/filterconstraint.cpp
+++ b/core/filterconstraint.cpp
@@ -64,11 +64,11 @@ static struct type_description {
 
 	{ FILTER_CONSTRAINT_DIVE_MODE, "dive_mode", QT_TRANSLATE_NOOP("gettextFromC", "dive mode"), false, false, false, FILTER_CONSTRAINT_NO_UNIT, 0, false, false },
 
-	{ FILTER_CONSTRAINT_TAGS, "tags", QT_TRANSLATE_NOOP("gettextFromC", "tags"), true, false, false, FILTER_CONSTRAINT_NO_UNIT, 0, false, false },
-	{ FILTER_CONSTRAINT_PEOPLE, "people", QT_TRANSLATE_NOOP("gettextFromC", "people"), true, false, false, FILTER_CONSTRAINT_NO_UNIT, 0, false, false },
-	{ FILTER_CONSTRAINT_LOCATION, "location", QT_TRANSLATE_NOOP("gettextFromC", "location"), true, false, false, FILTER_CONSTRAINT_NO_UNIT, 0, false, false },
-	{ FILTER_CONSTRAINT_WEIGHT_TYPE, "weight_type", QT_TRANSLATE_NOOP("gettextFromC", "weight type"), true, false, false, FILTER_CONSTRAINT_NO_UNIT, 0, false, false },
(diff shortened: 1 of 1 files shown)


bstoeger:
fine with me.

@dirkhh: care to comment on this UI change?

bstoeger (approved):

dirkhh:
I'm not sure I understand why calling this "any ... " is easier to understand. 
I would read it as "this triggers with any value"

It's a translation text change and I'd be concerned if the translators will know how to reflect the intention without more information.

So, I'm not against making improvements, but I worry about how this will play out for the user, especially the non-English users

bstoeger:
We could do "person (any of)". Optionally, have "(any of)" as separate translation string and have it automatically added if an `any_of` flag is set in the type description?

mikeller:
> I'm not sure I understand why calling this "any ... " is easier to understand. I would read it as "this triggers with any value"

This is just based on the observation that, when taken as natural language, the filter condition 'tags is starting with X' is clearly not correct in cases where it is not the first tag in the list that matches 'X':

image

> It's a translation text change and I'd be concerned if the translators will know how to reflect the intention without more information.

Yes, this will require the translators to get some context, either by checking in-app to see where this change is applying, or by being given cues (not sure if transifex supports this).

> We could do "person (any of)".

Yes, the '(any of)' is an alternative - but not sure if this wil make it easier for translators to grasp how it will be used without being provided with context.

> Optionally, have "(any of)" as separate translation string and have it automatically added if an `any_of` flag is set in the type description?

Unfortunately this will definitely break translations, or at least make them even more awkward, as in at least some of the languages where nouns have a gender the 'any' will have to follow the gender of the noun. (cf. 'ein' / 'eine' in German).

bstoeger:
> Yes, this will require the translators to get some context, either by checking in-app to see where this change is applying, or by being given cues (not sure if transifex supports this).

Qt has //: comments for translators: https://doc.qt.io/qt-5/i18n-source-translation.html Not sure if this works for the QT_TRANSLATE_NOOP macros, though.

> Unfortunately this will definitely break translations, or at least make them even more awkward, as in at least some of the languages where nouns have a gender the 'any' will have to follow the gender of the noun. (cf. 'ein' / 'eine' in German).

Yes, the translators might have to be creative in that case. I think @atdotde is responsible for the German translation (I don't use it) - best to ask him about proper gendering.

PS: Perhaps we should expand the code to allow for any of / all of / none of searches...?

mikeller:
> Yes, the translators might have to be creative in that case. I think @atdotde is responsible for the German translation (I don't use it) - best to ask him about proper gendering.

Or just leave it as the fixed expression for the handful of cases that we have. 😉

> PS: Perhaps we should expand the code to allow for any of / all of / none of searches...?

We already have 'is not' that can be used to compose a 'none', and users can add multiple conditions for the same 'any' element to get 'all of' semantics for multiple search terms.

PR #4250, opened 2024-06-16
tests: run profile test using default preferences

bstoeger:
As noted in a comment introduced in fe074ccad1, the profile test should probably best be run using the default preferences. This wasn't done back then, because the reference data assumes a (bogus) setting of modpO2 of 0.

This commit runs the test using the default preferences and updates the reference data accordingly.

This is in preparation of changes to the preference system, where the preference structure initializes itself to the default values.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

See commit message: run profile test using the default preferences. This should probably have been done a long time ago.

### DIFF
--- a/tests/testprofile.cpp
+++ b/tests/testprofile.cpp
@@ -22,13 +22,9 @@ void TestProfile::init()
 
 	// Set UTF8 text codec as in real applications
 	QTextCodec::setCodecForLocale(QTextCodec::codecForMib(106));
-	// first, setup the preferences
 
-	// normally we should be able to do this - but it makes this test fail because the reference data
-	// assume that the prefs are all 0 / false
-	// copy_prefs(&default_prefs, &prefs);
-	// instead we just set up the cloud_base_url to prevent parse_file() from crashing
-	prefs.cloud_base_url = strdup(default_prefs.cloud_base_url);
+	// first, setup the preferences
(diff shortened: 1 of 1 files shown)


bstoeger:
It will probably be best to rebase this on #4241.

@dirkhh: Or would it be best to combine these into a single commit? After all, changing the 31k lines of the profile data twice in a short time seems excessive.

mikeller:
> @dirkhh: Or would it be best to combine these into a single commit? After all, changing the 31k lines of the profile data twice in a short time seems excessive.

I'd keep them separate - this will be useful if we ever need to bisect against it.

And there's worse ways we are wasting GitHub resources 😉: https://github.com/subsurface/qt-ios

mikeller (approved):

dirkhh:
> It will probably be best to rebase this on #4241.
> 
> @dirkhh: Or would it be best to combine these into a single commit? After all, changing the 31k lines of the profile data twice in a short time seems excessive.

I don't really care about excessive changes, 
I dislike squashing commits for many reasons.

bstoeger:
Ok, I believe in this case nothing would be lost by squashing, but let's do it separately.

PR #4260, opened 2024-06-25
Fix crash on MacOS during download from new BLE computer

### LINKED ISSUE #4259: [macOS] crashes when trying first download from a new BLE dive computer.
### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

Ran into a crash (segmentation violation fault) when trying to do the first download from a brand new BLE dive computer.
Appears to happen on a new installation or on the first download from a BLE device (i.e. program does not have any remembered bluetooth devices.)

Running this from the command line produces the following output, which seems to be related to the issue:
`INFO: Starting download from BLE`
`INFO: downloading only new dives`
`INFO: ---> stopping the discovery agent`
`QObject::killTimer: Timers cannot be stopped from another thread`
`QObject::~QObject: Timers cannot be stopped from another thread`
`INFO: BT/BLE finished discovery`

### Operating system:

MacBook Pro (intel core i7 - 2020) Sonoma 14.5

Connecting to Shearwater Perdix 2 (firmware V95/BT11)

From code, appears to be a macOS problem only.

### Subsurface version:

Custom build from from master branch (master - PR #4253 )
compiled with QT 5.15.13 (subsurface/qt-mac)
built on MacBook Pro (intel core i7 - 2020) Sonoma 14.5

### Steps to reproduce:

1. remove all subsurface user-related files to remove saved dive computers
   - ~/Library/Caches/Subsurface
   - ~/Library/Application Support/Subsurface
   -  ~/Library/Preferences/org.subsurface-divelog.plist
   -  ~/Library/Preferences/org.hohndel.subsurface.Subsurface.plist
   -  ~/Library/Saved Application State/org.subsurface-divelog.savedState

3. Start subsurface
4. Import -> Import from dive computer (cmd-D)
5. The "Device or mount point" text box should be empty and "Choose Bluetooth download mode" should be unchecked.
6. Select Vendor and Dive computer (in my case Shearwater/Perdix 2)
7. Click on checkbox for "Choose Bluetooth download mode", which should immediately bring up the "Remote Bluetooth device selection" dialog
8. Press Scan button to start scan, select device from list when it is discovered, and then press Save to close the dialog
9. Press the Download button to start the download (looks like it needs to be pressed within 17 seconds of closing the previous dialog)

### Current behavior:

The download starts, but after the first or second dive, the program crashes with a segmentation violation fault.

### Expected behavior:

Should download the dives and not crash.

### Additional information:

From looking at the code, the following appears to be happening:
After user selects the device from the "Remote Bluetooth device selection" dialog and hits save, DownloadFromDCWidget::bluetoothSelectionDialogIsFinished() is called with the result.
It then sets the ui.device text with the selected device.
Which triggers the on_device_currentTextChanged() method.
If this is the first time this field has been set to a bluetooth address, it creates a new BTDiscover instance.
Which in turn creates a new QBluetoothDeviceDiscoveryAgent and starts a new scan.
Internally, this creates a QTimer, and sets it for 17 seconds.
When the user presses the Download button, it fires up the download process in a new thread.
One of the first things the download code does is call getBtDeviceInfo() - which finds the device in the saved devices and then attempts to stop the discovery agent - which will attempt to dispose of the QTimer.
Because this is being called from a different thread from where the timers were created, QT does not allow the timer to be killed and deregistered. So when QT attempts to handle the timeout event for this timer, it is dealing with freed up memory and causes the seg fault.

Some notes - only macOS starts up the scan when the device text is changed.
The QTimer seems to have been added some time between 5.15.2 and 5.15.13.

sbcwsgit:

- [x ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Ensure that any bluetooth scan started (by changing the device name) in the download-from-dive-computer dialog (desktop version) is stopped before the download process is started up.
Because the QT bluetooth discovery agent uses a QTimer internally, it must be stopped from the same thread as it was started from. Otherwise the download process (using a different thread), calls getBtDeviceInfo and attempts to shut down any running scan - it will dispose of the QTimer but is not able to kill or deregister the timer. So when the timer goes off it results in a memory violation.

### Changes made:

1. In DownloadFromDCWidget::on_downloadCancelRetryButton_clicked(), added a call to getBtDeviceInfo() if BTDiscovery instance has been created. This ensures that a) the device was discovered and b) any outstanding scan is stopped (from the same thread that started it).
2. Added BT_SUPPORT conditional compile to assignment of btd member in DownloadFromDCWidget, since the btd member is only present if BT_SUPPORT is defined.

### Related issues:

Fixes #4259 

### Additional information:

Only happens on MacOS.
QT5.15.2 does not use the QTimer internally, 5.15.13 does (not sure what version they added it).
From the code, looks like any BLE device would be affected - classical bluetooth devices may be affected (depends how fast the scan discovers it).
The device name needs to be empty or a non bluetooth address when to import dialog first appears - so that the BTDiscovery instance is first created when the remote device selection dialog returns (and download button is pressed while internal QTimer is still active).

### Release note:

have added the following to the change log:
`bluetooth: fix crash on MacOS when doing first download from a new BLE device`

### Documentation change:

none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+bluetooth: fix crash on MacOS when doing first download from a new BLE device
 statistics: show proper dates in January
 desktop: add country to the fields indexed for full text search
 import: update libdivecomputer version, add support for the Scubapro G3 / Luna and Shearwater Tern
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -365,7 +365,7 @@ void DownloadFromDCWidget::on_product_currentTextChanged(const QString &)
 
 void DownloadFromDCWidget::on_device_currentTextChanged(const QString &device)
 {
-#if defined(Q_OS_MACOS)
+#if defined(BT_SUPPORT) && defined(Q_OS_MACOS)
 	if (isBluetoothAddress(device)) {
 		// ensure we have a discovery running
 		if (btd == nullptr)
@@ -429,6 +429,16 @@ void DownloadFromDCWidget::on_downloadCancelRetryButton_clicked()
 			address = extractBluetoothNameAddress(ui.device->currentText(), name);
 			data->setDevName(address);
 			data->setDevBluetoothName(name);
+		}
+		if (btd) {
+			// btd should only be active for mac os.
+			// Need to ensure that any scan is shut down BEFORE starting the download
+			// because internally (as of 5.15.13) the QT discovery agent uses a QTimer
+			// that can only be shut off from the same thread it was started from.
+			QString devAddr = data->devName().startsWith("LE:")
+							 ? data->devName().right(data->devName().size()-3)
+							 : data->devName();
+			getBtDeviceInfo(devAddr);
 		}
 	} else
 		// this breaks an "else if" across lines... not happy...


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+bluetooth: fix crash on MacOS when doing first download from a new BLE device
```
mikeller, line 1: @dirkhh: Do we want to retire `CHANGELOG.md`, or at least remove it from the instructions for pull requests? With our new approach of doing CICD releases it does not seem to serve any particular function any more.

mikeller (approved):
Good find, thank you!

It looks like the email used in your DCO does not match the email configured in your git client:

https://github.com/subsurface/subsurface/pull/4260/checks?check_run_id=26629172835

sbcwsgit:
> Good find, thank you!
> 
> It looks like the email used in your DCO does not match the email configured in your git client:
> 
> https://github.com/subsurface/subsurface/pull/4260/checks?check_run_id=26629172835

Sorry about the email confusion - should be fixed up now.

PR #4268, opened 2024-07-24
From libdivecomputer: Add support for Mares Sirius.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

Changes from https://github.com/subsurface/libdc/pull/65.


mikeller:
@rmultan: Can you please us one of the above artifacts to test that your proposed changes in https://github.com/subsurface/libdc#65 are working?

rmultan:
@mikeller It works on MacOS 14.5 (M1)
image

Checking for update gives the below (I hope it's caused by the nightly build ID)
image


On a separate note, build with Qt5.15.13 didn't work on my M1 Mac after following these instructions: https://github.com/subsurface/subsurface/blob/master/INSTALL.md. I had to make few changes to build script.  After building I had to sign Subsurface.app with disabled library validation in order to run it (it's not mentioned in the guide). I should mention that I have 0 experience with MacOS development. I wonder if it's my local env setup or just outdated build script for MacOS (lines I've changed are 4-6 years old).

mikeller:
@rmultan: I am not a MacOS expert either, and I do not have access to an M1 Mac to test. Maybe @notrege can help?

@dirkhh: I think the version parsing in the update check on the server is a bit too strict - can this be changed to just count any release-candidate part as 'newer than version X' irrespective of what the release-candidate information is? (I know this is the opposite of what SemVer wants us to do, but this way makes more sense for our use case.)

rmultan:
I can open a separate issue & PR to showcase my findings, I made it work to test libdc changes.

notrege:
I have not tried to build with 5.15.13 on an M1.   My builds were done on an old (really old) Intel MacBook Pro and then tested on the M1.   My assumption is that the production build machine is Intel based.   I did get a build to complete on a M1, but that was with a newer version of QT and I didn't get it fully working before I found that an Intel build with 5.15.13 got Bluetooth uploads working with my M2.

rmultan:
bluetooth import works for me on M1 Mac with Subsurface built on the same Mac with Qt5.15.13 (as per INSTALL.md)

dirkhh:
> @rmultan: I am not a MacOS expert either, and I do not have access to an M1 Mac to test. Maybe @notrege can help?
> 
> @dirkhh: I think the version parsing in the update check on the server is a bit too strict - can this be changed to just count any release-candidate part as 'newer than version X' irrespective of what the release-candidate information is? (I know this is the opposite of what SemVer wants us to do, but this way makes more sense for our use case.)

Yeah... I'm not going to hand code parsing for any silly version. The update notifications are designed for people who use either latest or current builds. If you're using something else then I assume you know if and when you want to update

dirkhh:
> bluetooth import works for me on M1 Mac with Subsurface built on the same Mac with Qt5.15.13 (as per INSTALL.md)

Bluetooth import works for some people on some M1 Mac (or M2 Mac, for that matter) under some random circumstances. 
Whether they are built locally or not seems to make no difference. I actually have one M1 Mac where I cannot get BLE to work with the Qt5.15.x build, but it works in the (otherwise limited) Qt6 build.

mikeller:
> > @rmultan: I am not a MacOS expert either, and I do not have access to an M1 Mac to test. Maybe @notrege can help?
> > @dirkhh: I think the version parsing in the update check on the server is a bit too strict - can this be changed to just count any release-candidate part as 'newer than version X' irrespective of what the release-candidate information is? (I know this is the opposite of what SemVer wants us to do, but this way makes more sense for our use case.)
> 
> Yeah... I'm not going to hand code parsing for any silly version. The update notifications are designed for people who use either latest or current builds. If you're using something else then I assume you know if and when you want to update

I wouldn't consider [SemVer](https://semver.org/) to be silly. But you are right, hand-parsing is not the best approach here - it will likely be better to use a library for this, they should be available for most common languages / environments.
I just think it is less than ideal user experience if a user is running an artifact that they downloaded from our GitHub organisation and they are greeted with a 'System error' message.

dirkhh:
We are using a semver library...

mikeller:
> We are using a semver library...

Cool. Which one? I am keen to check what trips it up in our current naming convention for the artifact builds.

dirkhh:
semver 3.0.2 from PyPi

PR #4272, opened 2024-07-27
Reduce attack vector in artifact-links.yml

probonopd:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

`artifact-links.yml` uses a third-party GitHub Action. This pull request reduces the attack vector introduced by it. This should probably be done in a similar way for all other third-party GitHub Actions being used.

### Changes made:

- Pin action to git hash, https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash
- Restrict permissions for the GITHUB_TOKEN, https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

### Mentions:
Thanks @TheAssassin

### DIFF
--- a/.github/workflows/artifact-links.yml
+++ b/.github/workflows/artifact-links.yml
@@ -10,10 +10,15 @@ jobs:
     name: Add artifact links to PR and issues
     runs-on: ubuntu-22.04
 
+    permissions:
+      issues: write
+      pull-requests: write
+      actions: read
+
     steps:
     - name: Add artifact links to PR and issues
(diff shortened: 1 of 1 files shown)


mikeller (approved):
Fully on board for the permissions clamp-down, not so sure about using commit hashes in dependencies (at the cost of implicity version linkage). But alas, security.

PR #4276, opened 2024-08-08
FIX: Bug in "negate" filter constraint attribute when loaded from Git

rafsalvioni:

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

To filter presets using constraints with "negate" attribute, when loaded from Git (i.e cloud storage), throws a error. This is because "negate" attribute was saved without a value. Filter preset parser expects key/value pair. This fix add a default value "1" to "negate" attribute. This is the way used in SSRF, for example

### Changes made:
When save on Git, added a default value "1" to "negate" attribute

### Additional information:
Error after load cloud storage:

image

### DIFF
--- a/core/save-git.cpp
+++ b/core/save-git.cpp
@@ -968,7 +968,7 @@ static void format_one_filter_constraint(int preset_id, int constraint_id, struc
 		show_utf8(b, " rangemode=", mode, "");
 	}
 	if (constraint->negate)
-		put_format(b, " negate");
+		put_format(b, " negate=\"1\"");
 	std::string data = filter_constraint_data_to_string(constraint);
 	show_utf8(b, " data=", data.c_str(), "\n");
 }


mikeller:
What version did you reproduce this bug with?

This should actually be fixed by https://github.com/subsurface/subsurface/pull/4247, which adds support for reading of the filter constraints without a `=<value>` - from git.

rafsalvioni:
6.0.5214

Em qui., 8 de ago. de 2024, 20:30, Michael Keller ***@***.***>
escreveu:

mikeller:
> 6.0.5214

Figures. 6.0.215 has the fix for this problem: https://github.com/subsurface/nightly-builds/releases/tag/v6.0.5215-CICD-release

Can I recommend that next time, before you investigate a problem and implement a fix, you check with the latest version if the problem still reproduces?

rafsalvioni:
@mikeller I think official Subsurface website deceived me... I update my Subsurface this week... So I believed that I have last version. But, in official website, the published version is 6.0.5214:

image

You can close PR

mikeller:
@rafsalvioni: That's our current 'stable' release - we also do have a 'latest' release train that includes all the latest changes:

image

Issue #4275, opened 2024-08-07
UDDF Export: when select "all dives", exported file is empty


rafsalvioni:
### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:
When export dives to UDDF format and "all files" selection checked, result file is empty

### Operating system:
Ubuntu 22.04.4 LTS

### Subsurface version:
6.0.5230

### Steps to reproduce:
In a opened logbook:
1) Go to File -> Export (Ctrl+E);
2) Select "UDDF" in export format;
3) Select "all dives" in selection box;
4) Press "OK" and choose file location

### Current behavior:
Exported file is empty

### Expected behavior:
Exported file contents should be whole logbook in UDDF format

rafsalvioni:
I try to analyze the code, but I didnt found the issue... I think that is in XSLT transformation

mikeller:
I had a quick check, and with my test logs the UDDF exported just fine when 'all dives' is checked.
So I suspect that some data that you have in at least one of the dives in your log is causing an error and abort during the XSLT transformation.
Did you try running Subsurface with -v -v -v command line arguments and checking if there is anything in the log hinting at where the transformation is failing?

rafsalvioni:
@mikeller Log shows a error in XSLT:

![image](https://github.com/user-attachments/assets/589d2cb8-ee3f-4fd5-a82d-0933c7a42730)

I run Subsurface 6.0.5230 on Linux

mikeller:
@rafsalvioni: Can you please test if the artifacts in #4278 fix this?

rafsalvioni:
> @rafsalvioni: Can you please test if the artifacts in #4278 fix this?

Tested. Fix OK

PR #4279, opened 2024-08-10
FIX bug in UDDF export "all dives"

rafsalvioni:

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix bug https://github.com/subsurface/subsurface/issues/4275

### Changes made:
Added "test" variable on uddf-export.xslt

### Related issues:
https://github.com/subsurface/subsurface/issues/4275

### Release note:
FIX bug in UDDF export "all dives"

### Mentions:
@mikeller

### DIFF
--- a/xslt/uddf-export.xslt
+++ b/xslt/uddf-export.xslt
@@ -19,6 +19,8 @@
     </xsl:for-each>
   </xsl:variable>
 
+  <xsl:variable name="test" select="''" />
+
   <xsl:variable name="eventmap">
     <!--entry key="safety stop (mandatory)"></entry-->
     <!--entry key="deco"></entry-->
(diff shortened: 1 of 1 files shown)


mikeller:
Good job for finding the problem, and good try at fixing it.
I suspect that this `xsl:choose` (which should really be an `xsl:if` like all the other cases with a single `xsl:when`) was added to facilitate 'testing' by generating repeatable output when run in a test.
In #4278 I am copying existing behaviour from other places where `testid<x>` is used, and 'detect' if this is run in a test by looking at the `$units` input variable.
Apparently there are no tests that actually verify the export of trips, as this would otherwise break in the tests as well...

PR #4282, opened 2024-08-13
fix copy/paste of dive-site

bstoeger:
The copy/pasting of dive-sites was fundamentally broken in at least two ways:

1) The dive-site pointer in struct dive was simply overwritten, which
   breaks internal consistency. Also, no dive-site changed signals where
   sent.

2) The copied dive-site was stored as a pointer in a struct dive. Thus,
   the user could copy a dive, then delete the dive-site and paste.
   This would lead to a dangling pointer and ultimately crash the
   application.

Fix this by storing the UUID of the dive-site, not a pointer. To do that, don't store a copy of the dive, but collect all the data in a `dive_paste_data` structure.
If the dive site has been deleted on paste, do nothing. Send the appropriate signals on pasting.

The mobile version had an additional bug: It kept a pointer to the dive to be copied, which might become stale by undo.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The copy & paste of dive-site data was horribly buggy. See commit message: Not only would it not properly unregister/register the dives from the dive site, it was also keeping potentially stale pointers (either to dive sites or to dives). A crash is not completely easy to pull of but can be done.

Try to fix this by implementing value semantics instead of using references (pointers). The end result is not particularly nice, but hopefully better than outright wrong.

I didn't test the mobile version, because I neither found the settings screen where one can configure copy/paste, nor the actual buttons to copy/paste. Weird.

### Changes made:

1) Properly unregister/register dives on paste.
2) Use value semantics: store UUID of dive site, not pointer.
3) Copy dive data, don't keep a pointer to the copied dive on mobile.

### Release note:

Done.

### Documentation change:

Not needed.

### DIFF
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -267,9 +267,9 @@ int editDiveGuide(const QStringList &newList, bool currentDiveOnly)
 	return execute_edit(new EditDiveGuide(newList, currentDiveOnly));
 }
 
-void pasteDives(const dive *d, dive_components what)
+void pasteDives(const dive_paste_data &data)
 {
-	execute(new PasteDives(d, what));
+	execute(new PasteDives(data));
 }
 
 void replanDive(dive *d)
--- a/commands/command.h
+++ b/commands/command.h
@@ -13,6 +13,7 @@
 struct divecomputer;
 struct divelog;
 struct dive_components;
+struct dive_paste_data;
 struct dive_site;
 struct dive_trip;
 struct event;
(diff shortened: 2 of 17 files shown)


mikeller (approved):
Looks good to me, and seems to be working fine, including copy / pasting of tanks that never worked the way I thought it should.

PR #4295, opened 2024-08-22
correct re-plan menu label in documentation

tdhammer:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [x] Documentation change
- [ ] Language translation

The (presumably) older label "Re-plan dive" in the Log menu was replaced with "Edit dive in planner", but the documentation section "Modifying an existing dive plan" was not updated.

### Changes made:

1) replace "Re-plan dive" with "Edit dive in planner"

### Related issues:

None found _(should I create one?)_

### Documentation change:

_see above_

### DIFF
--- a/Documentation/user-manual.txt
+++ b/Documentation/user-manual.txt
@@ -4347,7 +4347,7 @@ The volumes of gases required for bailout can be found at the bottom of the *Div
 Normally, when a dive plan has been saved, it is accessible from the *Dive List*, like any
 other dive log. Within the *Dive List* there is no way to change a saved dive plan.
 To change a dive plan, select it on the *Dive List*. Then, in the main menu,
-select _Log ->  Re-plan dive_. This will open the selected dive plan within the dive planner,
+select _Log ->  Edit dive in planner_. This will open the selected dive plan within the dive planner,
 allowing changes to be made and saved as usual.
 
 In addition, there is the option "Save new". This keeps the original


Documentation/user-manual.html.git
```diff
@@ -6107,7 +6107,7 @@ overlapping icons.</td>
 <div class="paragraph"><p>Normally, when a dive plan has been saved, it is accessible from the <strong>Dive List</strong>, like any
 other dive log. Within the <strong>Dive List</strong> there is no way to change a saved dive plan.
 To change a dive plan, select it on the <strong>Dive List</strong>. Then, in the main menu,
-select <em>Log &#8594;  Re-plan dive</em>. This will open the selected dive plan within the dive planner,
+select <em>Log &#8594;  Edit dive in planner</em>. This will open the selected dive plan within the dive planner,
```
mikeller: The .git files are actually generated from the .txt files, so they do not need to be updated in a pull request.

tdhammer: Ah, I wondered about that extension...
I have just started looking into the implementation and hadn't gotten to that point. I see now that my build last night regenerated the files.
In my experience, generated files are usually not under source control.

I will add a commit to remove the change to that file.
Do you all prefer keeping fixups to PRs in separate commits and squashing just before merge?
Or should I just squash it before pushing the update?

mikeller: > Ah, I wondered about that extension... I have just started looking into the implementation and hadn't gotten to that point. I see now that my build last night regenerated the files. In my experience, generated files are usually not under source control.

I think this was originally done so that developers could do local builds that include the manual without having the toolchain installed that is required to build the manual.
Not sure how relevant this is today, as the toolchain is simple enough to install and run in a docker image.

> I will add a commit to remove the change to that file. Do you all prefer keeping fixups to PRs in separate commits and squashing just before merge? Or should I just squash it before pushing the update?

I don't mind either way.

mikeller (approved):

PR #4298, opened 2024-08-26
Fix Suunto FIT file import dive duration mis-calculation

torvalds:
The Suunto FIT files end up creating a sample every second, but a lot of those samples with no depth information, marked as "depth.mm" being negative.

That doesn't end up being a problem for subsurface, _except_ that it really confuses our "dc_fixup_duration()" logic, and the dive duration ends up being completely nonsensical (generally roughly by a factor of five: every tenth sample has a depth, and we only count samples that "begin or end under water" as being relevant for the dive duration, so two out of the ten samples will count towards the dive time).

Saving the dive will then not save these invalid depths, so saving and reloading the dive ends up fixing the dive duration calculation.

The fix is trivial - we just ignore samples with negative depth in dc_fixup_duration().

The FIT file parser should probably be taught to not even bother sending empty samples to subsurface, but that's a separate cleanup.  This fixes the actual bad behavior.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/divecomputer.cpp
+++ b/core/divecomputer.cpp
@@ -272,6 +272,10 @@ void fixup_dc_duration(struct divecomputer &dc)
 		int time = sample.time.seconds;
 		int depth = sample.depth.mm;
 
+		/* Do we *have* a depth? */
+		if (depth < 0)
+			continue;
+
 		/* We ignore segments at the surface */
 		if (depth > SURFACE_THRESHOLD || lastdepth > SURFACE_THRESHOLD) {
 			duration += time - lasttime;


dirkhh (approved):
well that looks obvious.
and yes, that seems like a lot of "noise" in the .FIT file.

torvalds:
Oh, and it's not even that libdivecomputer would be doing that negative depth. It's just that libdivecomputer will send DC_SAMPLE_TIME samples every second, because the FIT file has temperature and "altitude" data in it (although it seems to be just repeats - I wonder if the original Suunto data is fine, and it's an artefact of how they then turn it into a FIT file).

Silly, but mostly harmless, except for this odd effect.

mikeller (approved):
Makes sense!

PR #4299, opened 2024-08-26
Profile: Fix the Initial Gasmix.

mikeller:
Fix the initial gasmix that is shown in the tank bar of the profile.

Also add a meaningful gas name for gases with negative values for percentages.

@bstoeger: This is a side effect of the `event_loop` functionality introduced as part of #4198. In the case of an `event_loop("gasmix")` this does not take into account the edge case where there is no gaschange event at the very beginning of the dive, and the first gasmix is implicitly used as the starting gasmix. This happens for planned and manually added dives, but also for some dive computers. We are using the
same kind of loop in a number of other places in `core/profile.cpp`, `core/dive.cpp`, `core/gaspressures.cpp`, and `profile-widget/tankitem.cpp`, and I am wondering if we should be converting these to use `gasmix_loop` instead to avoid being bit by this special case?

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/event.h
+++ b/core/event.h
@@ -68,12 +68,26 @@ public:
 class gasmix_loop {
 	const struct dive &dive;
 	const struct divecomputer &dc;
-	struct gasmix last;
+	bool first_run;
 	event_loop loop;
-	const struct event *ev;
+	const struct event *next_event;
+	int last_cylinder_index;
+	int last_time;
 public:
 	gasmix_loop(const struct dive &dive, const struct divecomputer &dc);
-	gasmix next(int time);
+	// Return the next cylinder index / gasmix from the list of gas switches
+	// and the time in seconds when this gas switch happened
+	// (including the potentially imaginary first gas switch to cylinder 0 / air)
+	std::pair<int, int> next_cylinder_index(); // -1 -> end
+	std::pair<gasmix, int> next(); // gasmix_invalid -> end
+
+	// Return the cylinder index / gasmix at a given time during the dive
+	// and the time in seconds when this switch to this gas happened
+	// (including the potentially imaginary first gas switch to cylinder 0 / air)
+	std::pair<int, int> cylinder_index_at(int time); // -1 -> end
+	std::pair<gasmix, int> at(int time); // gasmix_invalid -> end
+
+	bool has_next() const;
 };
 
 /* Get divemodes at increasing timestamps. */
--- a/core/profile.cpp
+++ b/core/profile.cpp
@@ -673,7 +673,7 @@ static void setup_gas_sensor_pressure(const struct dive *dive, const struct dive
 
 	/* FIXME: The planner uses a dummy one-past-end cylinder for surface air! */
 	int num_cyl = pi.nr_cylinders + 1;
-	std::vector<int> seen(num_cyl, 0);
+	std::vector<int> seen(num_cyl, false);
 	std::vector<int> first(num_cyl, 0);
 	std::vector<int> last(num_cyl, INT_MAX);
 
@@ -696,7 +696,7 @@ static void setup_gas_sensor_pressure(const struct dive *dive, const struct dive
 				first[cylinder_index] = time;
 		}
 
-		seen[cylinder_index] = 1;
+		seen[cylinder_index] = true;
 
 		prev = cylinder_index;
 	}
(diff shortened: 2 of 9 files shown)


bstoeger (approved):
lgtm.

bstoeger:
> We are using the same kind of loop in a number of other places in core/profile.cpp, core/dive.cpp, core/gaspressures.cpp, and profile-widget/tankitem.cpp, and I am wondering if we should be converting these to use gasmix_loop instead to avoid being bit by this special case?

If you feel inclined to do so - why not?

dirkhh (approved):
this looks reasonable to me

mikeller:
I am looking into implementing a more generic solution that solves the 'first cylinder' special case in the entire application.

mikeller:
@bstoeger, @dirkhh: This went way deeper down the rabbithole than I'd intended, but at least now we have consistency in how the 'initial gas switch' is handled.

core/profile.cpp
```diff
@@ -677,31 +677,28 @@ static void setup_gas_sensor_pressure(const struct dive *dive, const struct dive
...
-	seen[prev] = 1;
-
-	event_loop loop("gaschange");
-	while (auto ev = loop.next(*dc)) {
-		int cyl = ev->gas.index;
-		int sec = ev->time.seconds;
+	int prev = -1;
+	gasmix_loop loop(*dive, *dc);
+	while (loop.has_next()) {
+		std::pair<int, int> gasmix_event = loop.next_cylinder_index();
```
bstoeger: Here, and elsewhere: It could make sense to use structured bindings to make the meaning of this code more clear.

core/profile.cpp
```diff
@@ -677,31 +677,28 @@ static void setup_gas_sensor_pressure(const struct dive *dive, const struct dive
...
-		last[cyl] = sec;
-		if (!seen[cyl]) {
-			// The end time may be updated by a subsequent cylinder change
-			first[cyl] = sec;
-			seen[cyl] = 1;
+			if (!seen[gasmix_event.first])
+				first[gasmix_event.first] = gasmix_event.second;
 		}
+
+		seen[gasmix_event.first] = 1;
```
bstoeger: Since you are touching this line, you could use the more sensible `true` instead of `1`.

core/event.h
```diff
@@ -68,12 +68,18 @@ class event_loop
...
+	bool first_run;
 	event_loop loop;
-	const struct event *ev;
+	const struct event *next_event;
+	int last_cylinder_index;
+	int last_time;
 public:
 	gasmix_loop(const struct dive &dive, const struct divecomputer &dc);
-	gasmix next(int time);
+	std::pair<int, int> next_cylinder_index(); // -1 -> end
```
bstoeger, line 81: Would be great if you could add a comment, what these functions do and what their return values are.

core/profile.cpp
mikeller: Ha, I didn't know that they exist (yet) in C++ - done, looks neat.

core/event.h
mikeller: Done.

core/profile.cpp
mikeller: This was used in some form of tri-state-ish. Refactored in [`b2e679d` (#4299)](https://github.com/subsurface/subsurface/pull/4299/commits/b2e679d34b1d133484f05d8fb425fcc6ea0aab37)

core/profile.cpp
bstoeger: Yes this makes code more readable. Just be careful: these aren't real variables, but some obscure notion of "bindings", which can have weird behavior.

core/profile.cpp
```diff
@@ -673,7 +673,7 @@ static void setup_gas_sensor_pressure(const struct dive *dive, const struct dive
 
 	/* FIXME: The planner uses a dummy one-past-end cylinder for surface air! */
 	int num_cyl = pi.nr_cylinders + 1;
-	std::vector<int> seen(num_cyl, 0);
+	std::vector<bool> seen(num_cyl, false);
```
bstoeger: Please don't use `std::vector<bool>`. This has some insane optimization, which usually is a pessimization. A huge error in the stdlib.

mikeller: Ha, I didn't know. Yes, using bitfields for this seems to be an odd choice for a high level language. Fixed.

mikeller: Interesting that they are using bitfields here - seems to be an odd choice of implementation for a high-level language. Fixed.

PR #4301, opened 2024-08-28
Fix sample times in dive merging

torvalds:
Commit c27314d60 ("core: replace add_sample() by append_sample()") broke the dive computer interleaving when merging two dives: the sample merging (done by "merge_samples()") no longer took the offset between the two merged dives into account, and instead just blindly copied the samples from the second dive computer with no time offset.

The end result was a completely broken profile.

This adds back the sample offset.  It also takes the offset not from the difference in time of the two dives, but the difference in time of the dive computers.  That way we're not mixing up different times from different sources that aren't necessarily in sync (the time *difference* is hopefully the same, but still..).

The dive merging still messes up the dive location. That's some other bug.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.cpp
+++ b/core/dive.cpp
@@ -1210,7 +1210,9 @@ static void merge_samples(struct divecomputer &res,
 		/* Only samples from b? */
 		if (at < 0) {
 		add_sample_b:
-			merge_one_sample(*bs, res);
+			struct sample sample = *bs;
+			sample.time.seconds += offset;
+			merge_one_sample(sample, res);
 			renumber_last_sample(res, cylinders_map_b);
 			bs++;
 			continue;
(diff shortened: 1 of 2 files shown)


bstoeger (approved):
Looks good to me.

torvalds:
Added another commit to also expose the new location data in the merged dive.

Berthold (in email) said:
 "From a quick glance, the code in dive_table::merge_dives looks
  fundamentally broken, because it may overwrite site->location outside
  of the undo system. I.e. this will not be undone.


  For now, I think your one-liner is OK, though it makes the whole 
  storage of the dive site in merge_result moot"

so there's a bigger issue with the lack of undo for the 'site->location" data, but that seems entirely independent of then exposing this site in the new merged dive.

bstoeger:
Thanks. Now one could remove
```
		d->dive_site = site;
```
In the caller `MergeDives::MergeDives()`. But let's just keep it that way for now.

PR #4300, opened 2024-08-27
Qt6:Build with QT6.x.x on Macos with map support

Egbertdepauw:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

Building on macos with the Qt6 framework.

made changes to several files to make it possible to build subsurface desktop with the Qt 6.5.x and higher framework.

Tested versions: 6.5.2, 6.6.3, 6.7.2

code builds and googlemaps works, made some adjustments to get panning and zooming with mouse or trackpad working.

See issue #3577 "Build Fails, macOS" for build details.

### Changes made:

1) MapWidget.qml 
- Removed import versions.
- Added PinchHandler WheelHandler and DragHandler for mouse zooming and panning support
- Changed Parameter passing from anonymous to lambda as per default in Qt6
2) qmlmapwidgethelper.cpp
- Changed the return string in function QString MapWidgetHelper::pluginObject() to remove the versioning of the import statements.
3) scripts/get-dep-lib.sh
- use googlemaps master from vladest instead of the forked one.
### Related issues:

#3577 "Build Fails, macOS" for build details.
### Mentions:

@dirkhh @arrmo

PR #4302, opened 2024-08-28
Egbertdepauw qt6devel

dirkhh:
See #4300

### DIFF
--- a/map-widget/qmlmapwidgethelper.cpp
+++ b/map-widget/qmlmapwidgethelper.cpp
@@ -4,20 +4,20 @@
 #include <QDebug>
 #include <QVector>
 
-#include "qmlmapwidgethelper.h"
 #include "core/divefilter.h"
 #include "core/divelist.h"
 #include "core/divelog.h"
 #include "core/divesite.h"
 #include "core/qthelper.h"
 #include "core/range.h"
-#include "qt-models/maplocationmodel.h"
+#include "qmlmapwidgethelper.h"
 #include "qt-models/divelocationmodel.h"
+#include "qt-models/maplocationmodel.h"
 #ifndef SUBSURFACE_MOBILE
 #include "desktop-widgets/mapwidget.h"
 #endif
 
-#define SMALL_CIRCLE_RADIUS_PX            26.0
+#define SMALL_CIRCLE_RADIUS_PX 26.0
 
 MapWidgetHelper::MapWidgetHelper(QObject *parent) : QObject(parent)
 {
@@ -44,7 +44,7 @@ void MapWidgetHelper::centerOnDiveSite(struct dive_site *ds)
 	} else {
 		// dive site with GPS
 		m_mapLocationModel->setSelected(ds);
-		QGeoCoordinate dsCoord (ds->location.lat.udeg * 0.000001, ds->location.lon.udeg * 0.000001);
+		QGeoCoordinate dsCoord(ds->location.lat.udeg * 0.000001, ds->location.lon.udeg * 0.000001);
 		QMetaObject::invokeMethod(m_map, "centerOnCoordinate", Q_ARG(QVariant, QVariant::fromValue(dsCoord)));
 	}
 }
@@ -69,7 +69,7 @@ void MapWidgetHelper::centerOnSelectedDiveSite()
 	// find the most top-left and bottom-right dive sites on the map coordinate system.
 	qreal minLat = 0.0, minLon = 0.0, maxLat = 0.0, maxLon = 0.0;
 	int count = 0;
-	for(struct dive_site *dss: selDS) {
+	for (struct dive_site *dss : selDS) {
 		if (!has_location(&dss->location))
 			continue;
 		qreal lat = dss->location.lat.udeg * 0.000001;
@@ -92,7 +92,7 @@ void MapWidgetHelper::centerOnSelectedDiveSite()
 	// Pass coordinates to QML, either as a point or as a rectangle.
 	// If we didn't find any coordinates, do nothing.
 	if (count == 1) {
-		QGeoCoordinate dsCoord (selDS[0]->location.lat.udeg * 0.000001, selDS[0]->location.lon.udeg * 0.000001);
+		QGeoCoordinate dsCoord(selDS[0]->location.lat.udeg * 0.000001, selDS[0]->location.lon.udeg * 0.000001);
 		QMetaObject::invokeMethod(m_map, "centerOnCoordinate", Q_ARG(QVariant, QVariant::fromValue(dsCoord)));
 	} else if (count > 1) {
 		QGeoCoordinate coordTopLeft(minLat, minLon);
@@ -134,7 +134,7 @@ void MapWidgetHelper::selectedLocationChanged(struct dive_site *ds_in)
 		return;
 	QGeoCoordinate locationCoord = location->coordinate;
 
-	for (auto [idx, dive]: enumerated_range(divelog.dives)) {
+	for (auto [idx, dive] : enumerated_range(divelog.dives)) {
 		struct dive_site *ds = dive->dive_site;
 		if (!ds || !ds->has_gps_location())
 			continue;
@@ -151,9 +151,9 @@ void MapWidgetHelper::selectedLocationChanged(struct dive_site *ds_in)
 	}
 	int last; // get latest dive chronologically
 	if (!selectedDiveIds.isEmpty()) {
-		 last = selectedDiveIds.last();
-		 selectedDiveIds.clear();
-		 selectedDiveIds.append(last);
+		last = selectedDiveIds.last();
+		selectedDiveIds.clear();
+		selectedDiveIds.append(last);
 	}
 #endif
 	emit selectedDivesChanged(selectedDiveIds);
@@ -162,7 +162,7 @@ void MapWidgetHelper::selectedLocationChanged(struct dive_site *ds_in)
 void MapWidgetHelper::selectVisibleLocations()
 {
 	QList<int> selectedDiveIds;
-	for (auto [idx, dive]: enumerated_range(divelog.dives)) {
+	for (auto [idx, dive] : enumerated_range(divelog.dives)) {
 		struct dive_site *ds = dive->dive_site;
 		if (!ds || ds->has_gps_location())
 			continue;
@@ -171,7 +171,7 @@ void MapWidgetHelper::selectVisibleLocations()
 		QGeoCoordinate dsCoord(latitude, longitude);
 		QPointF point;
 		QMetaObject::invokeMethod(m_map, "fromCoordinate", Q_RETURN_ARG(QPointF, point),
-		                          Q_ARG(QGeoCoordinate, dsCoord));
+					  Q_ARG(QGeoCoordinate, dsCoord));
 		if (!qIsNaN(point.x()))
 #ifndef SUBSURFACE_MOBILE // indices on desktop
 			selectedDiveIds.append(idx);
@@ -181,9 +181,9 @@ void MapWidgetHelper::selectVisibleLocations()
 	}
 	int last; // get latest dive chronologically
 	if (!selectedDiveIds.isEmpty()) {
-		 last = selectedDiveIds.last();
-		 selectedDiveIds.clear();
-		 selectedDiveIds.append(last);
+		last = selectedDiveIds.last();
+		selectedDiveIds.clear();
+		selectedDiveIds.append(last);
 	}
 #endif
 	emit selectedDivesChanged(selectedDiveIds);
@@ -205,11 +205,11 @@ void MapWidgetHelper::calculateSmallCircleRadius(QGeoCoordinate coord)
 {
 	QPointF point;
 	QMetaObject::invokeMethod(m_map, "fromCoordinate", Q_RETURN_ARG(QPointF, point),
-	                          Q_ARG(QGeoCoordinate, coord));
+				  Q_ARG(QGeoCoordinate, coord));
 	QPointF point2(point.x() + SMALL_CIRCLE_RADIUS_PX, point.y());
 	QGeoCoordinate coord2;
 	QMetaObject::invokeMethod(m_map, "toCoordinate", Q_RETURN_ARG(QGeoCoordinate, coord2),
-	                          Q_ARG(QPointF, point2));
+				  Q_ARG(QPointF, point2));
 	m_smallCircleRadius = coord2.distanceTo(coord);
 }
 
@@ -251,8 +251,8 @@ QString MapWidgetHelper::pluginObject()
 {
 	QString lang = getUiLanguage().replace('_', '-');
 	QString cacheFolder = QString::fromStdString(system_default_directory() + "/googlemaps").replace("\\", "/");
-	return QStringLiteral("import QtQuick 2.0;"
-			      "import QtLocation 5.3;"
+	return QStringLiteral("import QtQuick;"
+			      "import QtLocation;"
 			      "Plugin {"
 			      "    id: mapPlugin;"
 			      "    name: 'googlemaps';"
@@ -263,5 +263,6 @@ QString MapWidgetHelper::pluginObject()
 			      "            console.warn('MapWidget.qml: cannot find a plugin named: ' + name);"
 			      "        }"
 			      "    }"
-			      "}").arg(lang, cacheFolder);
+			      "}")
+		.arg(lang, cacheFolder);
 }
--- a/map-widget/qml/MapWidget.qml
+++ b/map-widget/qml/MapWidget.qml
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0
-import QtQuick 2.5
-import QtLocation 5.3
-import QtPositioning 5.3
+import QtQuick 
+import QtLocation 
+import QtPositioning 
 import org.subsurfacedivelog.mobile 1.0
 
 Item {
@@ -15,7 +15,7 @@ Item {
 		id: mapHelper
 		map: map
 		editMode: false
-		onSelectedDivesChanged: rootItem.selectedDivesChanged(list)
+		onSelectedDivesChanged: (list) => { rootItem.selectedDivesChanged(list) }
 		onEditModeChanged: editMessage.isVisible = editMode === true ? 1 : 0
 		onCoordinatesChanged: {}
 		Component.onCompleted: {
@@ -29,7 +29,6 @@ Item {
 		id: map
 		anchors.fill: parent
 		zoomLevel: defaultZoomIn
-
 		property var mapType
 		readonly property var defaultCenter: QtPositioning.coordinate(0, 0)
 		readonly property real defaultZoomIn: 12.0
@@ -41,12 +40,46 @@ Item {
 		property real newZoomOut: 1.0
 		property var clickCoord: QtPositioning.coordinate(0, 0)
 		property bool isReady: false
-
 		Component.onCompleted: isReady = true
 		onZoomLevelChanged: {
 			if (isReady)
 				mapHelper.calculateSmallCircleRadius(map.center)
 		}
+        property geoCoordinate startCentroid
+        startCentroid: newCenter
+
+        PinchHandler {
+            id: pinch
+            target: null
+            onActiveChanged: if (active) {
+                map.startCentroid = map.toCoordinate(pinch.centroid.position, false)
+            }
+            onScaleChanged: (delta) => {
+                map.zoomLevel += Math.log2(delta)
+                map.alignCoordinateToPoint(map.startCentroid, pinch.centroid.position)
+            }
+            onRotationChanged: (delta) => {
+                map.bearing -= delta
+                map.alignCoordinateToPoint(map.startCentroid, pinch.centroid.position)
+            }
+            grabPermissions: PointerHandler.TakeOverForbidden
+        }
+        WheelHandler {
+            id: wheel
+            // workaround for QTBUG-87646 / QTBUG-112394 / QTBUG-112432:
+            // Magic Mouse pretends to be a trackpad but doesn't work with PinchHandler
+            // and we don't yet distinguish mice and trackpads on Wayland either
+            acceptedDevices: Qt.platform.pluginName === "cocoa" || Qt.platform.pluginName === "wayland"
+                             ? PointerDevice.Mouse | PointerDevice.TouchPad
+                             : PointerDevice.Mouse
+            rotationScale: 1/120
+            property: "zoomLevel"
+        }
+        DragHandler {
+            id: drag
+            target: null
+            onTranslationChanged: (delta) => map.pan(-delta.x, -delta.y)
+        }
 
 		MapItemView {
 			id: mapItemView
@@ -67,7 +100,9 @@ Item {
 					}
 					MouseArea {
 						drag.target: (mapHelper.editMode && model.isSelected) ? mapItem : undefined
-						anchors.fill: parent
+            anchors.fill: parent
+            hoverEnabled: true
+
 						onClicked: {
 							if (!mapHelper.editMode && model.divesite)
 								mapHelper.selectedLocationChanged(model.divesite)
@@ -122,8 +157,8 @@ Item {
 
 		MouseArea {
 			anchors.fill: parent
-			onPressed: { map.stopZoomAnimations(); mouse.accepted = false }
-			onWheel: { map.stopZoomAnimations(); wheel.accepted = false }
+			onPressed: (mouse) => { map.stopZoomAnimations(); mouse.accepted = false }
+			onWheel: (wheel) => { map.stopZoomAnimations(); wheel.accepted = false }
 			onDoubleClicked: map.doubleClickHandler(map.toCoordinate(Qt.point(mouseX, mouseY)))
 		}
 
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -571,7 +571,11 @@ if [ "$QUICK" != "1" ] && [ "$BUILD_DESKTOP$BUILD_MOBILE" != "" ] && ( [[ $QT_VE
 	# build the googlemaps map plugin
 
 	cd "$SRC"
-	./${SRC_DIR}/scripts/get-dep-lib.sh single . googlemaps
+	if [ "$BUILD_WITH_QT6" = "1" ] ; then
+		./${SRC_DIR}/scripts/get-dep-lib.sh -qt6 single . googlemaps
+	else
+		./${SRC_DIR}/scripts/get-dep-lib.sh single . googlemaps
+	fi
 	pushd googlemaps
 	mkdir -p build
 	mkdir -p J10build
--- a/scripts/get-dep-lib.sh
+++ b/scripts/get-dep-lib.sh
@@ -169,8 +169,9 @@ for package in "${PACKAGES[@]}" ; do
 			git_checkout_library breeze-icons $CURRENT_BREEZE_ICONS https://github.com/kde/breeze-icons.git
 			;;
 		googlemaps)
-			git_checkout_library googlemaps master https://github.com/Subsurface/googlemaps.git
-			;;
+			#git_checkout_library googlemaps master https://github.com/Subsurface/googlemaps.git
+			git_checkout_library googlemaps master https://github.com/vladest/googlemaps.git
+      ;;
 		hidapi)
 			git_checkout_library hidapi master https://github.com/libusb/hidapi.git
 			;;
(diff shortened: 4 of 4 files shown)


dirkhh:
Dropping the versions appears to cause issues. 
But at least now googlemaps builds again.
Not sure how much more I'll get done this evening (I'm in GMT+12 right now).
We should most likely just have a little helper script that removes the qml import versions when we are building against Qt6 and otherwise keep them (for the Qt5 builds).

I'm sure there's more that's needed.

Egbertdepauw:
I will have a look at the googlemaps part separately. I have made a little Qt project just to test this part and I will try to build it against Qt5 and Qt6. I hope to have some better results today.

dirkhh:
> I will have a look at the googlemaps part separately. I have made a little Qt project just to test this part and I will try to build it against Qt5 and Qt6. I hope to have some better results today.

So the issue is that vladest's port simply switches the interfaces used from the Qt5 ones to the Qt6 ones.
So my way of addressing this is to take my old branch for Qt5, and the new version from vladest for Qt6.

But what we need to do is undo the 'remove versions from import' in your patch and instead strip those versions if we notice that the system is building against Qt6.

dirkhh:
The Android build is using the qml compiler and that notices the missing versions. I'm pretty sure that you'd see the same error when running any of the artifacts from this PR.

dirkhh:
Yep, here's a local Linux build:
```
./subsurface
qrc:/qml/MapWidget.qml: Library import requires a version
qrc:/qml/MapWidget.qml: Library import requires a version
qrc:/qml/MapWidget.qml: Library import requires a version
qrc:/qml/MapWidget.qml:48:18: Expected property type 
             property geoCoordinate startCentroid 
                      ^
INFO: qrc:/qml/MapWidget.qml failed to load with status: 3
```

Egbertdepauw:
@dirkhh Do you know what the highest version numbers are that the Qt5 will build with? We could try that, I am reading the Qt docs and they state that there are plans to drop the versioning completely, but for now you can still use versioning.

So if we can try with the highest version number, maybe it will build with Qt5 and Qt6

dirkhh:
I do not. 
And that seems like one of those "goose chase"  non-solutions. 
I think I prefer the well tested versions we have right now and to simply strip them via script (I mean, that's a simple `sed` command) for Qt6 builds

mikeller:
> I think I prefer the well tested versions we have right now and to simply strip them via script (I mean, that's a simple `sed` command) for Qt6 builds

What's the long-term plan here? For how long do we expect to have to do 'the splits' and keep Qt5 and Qt6 versions buildable? Even if we can address issues like the version numbers in build script I suspect that this will add more overhead to development, and will result in regression bugs going undetected in one or the other version, especially as a lot of the differences will be UI related, which means they are unlikely to be covered by CICD.

Egbertdepauw (approved):

dirkhh:
> > I think I prefer the well tested versions we have right now and to simply strip them via script (I mean, that's a simple `sed` command) for Qt6 builds
> 
> What's the long-term plan here? For how long do we expect to have to do 'the splits' and keep Qt5 and Qt6 versions buildable? 

That's a great question. My goal is of course to move to Qt6 fairly swiftly. Right now I see three major milestones
- printing and displaying the user manual -- @atdotde said he was going to attempt to look at printing. To me that one is a show stopper - we cannot lose printing, and in particular printing that still works with templates that people created.
- mobile app -- as I mentioned, I have started working on that, still a long ways to go
- successful builds on iOS and Android -- that's hard to test before the mobile app at least starts (which it doesn't right now). That said, I think I have worked through most of the **other** Qt6 Android build issues (fingers crossed) and haven't even started to look at iOS

So, honestly, I can't give you an end time, but I can tell you that it certainly isn't supposed to be a permanent solution.

> Even if we can address issues like the version numbers in build script I suspect that this will add more overhead to development, and will result in regression bugs going undetected in one or the other version, especially as a lot of the differences will be UI related, which means they are unlikely to be covered by CICD.

Yes, absolutely agreed. On the flip side, staying on 5.15 forever isn't a solution, either. A couple of other projects have found out that Android becomes almost impossible these days with 5.15, and support for modern BLE on macOS in 5.15 is severely lacking. Etc.

Egbertdepauw:
> * printing and displaying the user manual -- @atdotde said he was going to attempt to look at printing. To me that one is a show stopper - we cannot lose printing, and in particular printing that still works with templates that people created.

I think the problem is that subsurface uses Qt5 Qtwebkit and Qt6 has moved to QtWebEngine, which is included again in the official Qt downloads and maintenance tools. If I do a quick search through subsurface code I only see 1 include of QtWebKiWidgets and that is in the file desktop-widgets/printer.cpp

We should probably port the use of QtWebKit to QtWebEngine as stated in [Qt's porting guide](https://doc.qt.io/qt-6/qtwebenginewidgets-qtwebkitportingguide.html)

Egbertdepauw:
And QWebview is used in the usermanual.

dirkhh:
> We should probably port the use of QtWebKit to QtWebEngine as stated in [Qt's porting guide](https://doc.qt.io/qt-6/qtwebenginewidgets-qtwebkitportingguide.html)

Yes, and the last time we looked into this a key feature of QtWebKit was missing in QtWebEngine. That is what, I believe, @atdotde was going to look into.
Basically the pagination of the printouts use a QtWebKit function to determine the vertical space used by a block - and that ability was missing. Without that our pagination doesn't work.

Admittedly, it's been a couple of years since I've looked into THAT rats nest.

Egbertdepauw:
> Yes, and the last time we looked into this a key feature of QtWebKit was missing in QtWebEngine. That is what, I believe, @atdotde was going to look into.

QWebElement & QWebElementCollection are missing in QWebEngine. These are used as in printer.cpp for pagination and styling.

> Admittedly, it's been a couple of years since I've looked into THAT rats nest.

Qt now proposes that you use [javascript](https://doc.qt.io/qt-5/qwebenginepage.html#runJavaScript-1) for these kind of things....

Egbertdepauw:
Qt has an example of how you should change styles etc. with Javascript.
The example where they use this functionality is called [recipe browser](https://doc.qt.io/qt-6/qtwebengine-webenginewidgets-recipebrowser-example.html)

In this example they use javascript to change styles of an existing web page.

**Inserting a stylesheet**
```
void MainWindow::insertStyleSheet(const QString &name, const QString &source, bool immediately)
{
    QWebEngineScript script;
    QString s = QString::fromLatin1("(function() {"
                                    "    css = document.createElement('style');"
                                    "    css.type = 'text/css';"
                                    "    css.id = '%1';"
                                    "    document.head.appendChild(css);"
                                    "    css.innerText = '%2';"
                                    "})()")
                        .arg(name, source.simplified());
    if (immediately)
        ui->webEngineView->page()->runJavaScript(s, QWebEngineScript::ApplicationWorld);

    script.setName(name);
    script.setSourceCode(s);
    script.setInjectionPoint(QWebEngineScript::DocumentReady);
    script.setRunsOnSubFrames(true);
    script.setWorldId(QWebEngineScript::ApplicationWorld);
    ui->webEngineView->page()->scripts().insert(script);
}
```

**Removing a stylesheet**
```
void MainWindow::removeStyleSheet(const QString &name, bool immediately)
{
    QString s = QString::fromLatin1("(function() {"
                                    "    var element = document.getElementById('%1');"
                                    "    element.outerHTML = '';"
                                    "    delete element;"
                                    "})()")
                        .arg(name);
    if (immediately)
        ui->webEngineView->page()->runJavaScript(s, QWebEngineScript::ApplicationWorld);

    const QList<QWebEngineScript> scripts = ui->webEngineView->page()->scripts().find(name);
    if (!scripts.isEmpty())
        ui->webEngineView->page()->scripts().remove(scripts.first());
}
```

glance-:
We started a transition to QWebEngine back in https://github.com/subsurface/subsurface/commit/529a4d499b8dad35de6cd43e3004017ccfd288aa
But that was later dropped in 
https://github.com/subsurface/subsurface/commit/643f4a5726926b86693b442e318f40cf28fee04a

So, bringing back https://github.com/subsurface/subsurface/commit/529a4d499b8dad35de6cd43e3004017ccfd288aa is probably a good start.

Egbertdepauw:
Got some of it working, but still a lot to do:

- Usermanual works, but searching doesn't (I think this is an easy fix)
- print dialog works
- printing itself doesn't but this is expected because I had to comment out a lot of code.
- print dialog -> export to HTML works

Because exporting to html works like it's supposed to, maybe that should be the starting point to rewrite the printing?

mikeller:
> So, honestly, I can't give you an end time, but I can tell you that it certainly isn't supposed to be a permanent solution.

Thanks @dirkhh, makes sense. So it might make sense to declare a 'sprint' on Qt6 and ask users to help with converting / testing. Once we got this somewhat going, and have given some time for feedback on #4198 to come in, we could cut a 'current', and then pull off the bandaid and move latest to Qt6 - keeping this experimental and only exposed to a small group of developers isn't going to do us any favours.

> Yes, absolutely agreed. On the flip side, staying on 5.15 forever isn't a solution, either.

I never saw that as an option. 😉

> In this example they use javascript to change styles of an existing web page.

Man that's fugly. If they want this done in JS, then I'd much rather go for having all of the style management done in JS in a self-contained, and only inject the parameters for the styling.

dirkhh:
I think it will take quite a while before everything works with Qt6
We've been trying on and off for 3+ years

I didn't think it's realistic to think we'll be able to switch with a hard break. 
I much rather be able to support both for a while

mikeller:
> I didn't think it's realistic to think we'll be able to switch with a hard break. I much rather be able to support both for a while

Not sure how you envision the details of 'supporting both for a while' - will this be to release some of the platforms in Qt6, while others are still Qt5?

dirkhh:
we had a very rudimentary Fedora Qt6 build for a while.
What I envision is adding additional Qt6 builds for Windows / macOS / Android / iOS -- as they get closer to being useful.
And maybe I'm wrong and we should just force the change.
I'm playing with build system magic to have different qml files depending on which Qt version is used, but cmake is making this extremely painful (actually, my lack of skill with cmake is likely to blame here)

mikeller:
> we had a very rudimentary Fedora Qt6 build for a while. What I envision is adding additional Qt6 builds for Windows / macOS / Android / iOS -- as they get closer to being useful. 

I guess the value of the Qt6 builds (before they are complete enough for end users to use them) will be to get CICD coverage?
If so, I am wondering if we should do this work in a long lived feature branch (on which we can set up additional CICD if we want to), and avoid the additional work from 'doing the splits' for CICD. This will also avoid the work of building back the splits after we've switched.

> And maybe I'm wrong and we should just force the change.

I think we can probably leverage the 'two tier' release process that we now have: Do the Qt6 work in a branch until we have an MVP, then put out the last Qt5 'current' release, merge the Qt6, branch, and from there on in get Qt6 exposed through 'latest' releases, enabling us to get feedback from users.

And in my opinion, while we definitely want printing and an integrated manual, these are not needed for an MVP, as long as we are transparent to users about it:
- the manual can be viewed online for the time being;
- I suspect not many people bother to print their logs these days (and the ones who are will be able to stay on current until printing is done in Qt6, or use current on the machine they are printing from, and latest everywhere else).

dirkhh:
Yeah - as I think more about this... maybe having a qt6 branch to work in for a bit is the way to go, and then just do a hard switch when all the features are ready.
@Egbertdepauw -- I saw that you looked into printing and other areas... is this something you plan to **work** on or was that just a quick exploration? Just trying to understand who is doing what.
@atdotde -- any success with looking at printing?
Next week I'll clean up the work I did for the mobile app with Qt6 and hopefully we can combine all this into one branch that we collaborate on.

Egbertdepauw:
> I saw that you looked into printing and other areas... is this something you plan to **work** on or was that just a quick exploration? Just trying to understand who is doing what.

@dirkhh Yes I am looking into to and trying to get a version with the **usermanual** and **printing support**. For now I have some rudimentary printing with templates and the **usermanual** works as expected but without searching, which I expect is a small thing to fix.

I have another issue with the printing and that is that even if I download the **latest official mac binaries**, I have no print options. So I cannot test (without setting up a VM, which I will do) what it should look like. 

I also have another problem and that is that I have to bypass the "official" QPrintDialog because it crashes, the strange thing is that the QPrintPreviewDialog doesn't ?

Cmake is also a thing, I got everything building by making changes to the cmakelists.txt, but probably not in the most elegant way or according to subsurface standards. So for now, just me testing things, this is fine, but eventually when I have working CPP code, someone should have a look at the cmake file.

dirkhh:
So printing on Mac should work (works in my own builds), but possibly was broken in the cicd build. Can you check if that binary contains the QtWebKit dylib?

Our cmake "standards"... well that's a big word. 
There's a lot of hacking around from some people with more and some people with less cmake expertise 😁

dirkhh:
So printing on Mac should work (works in my own builds), but possibly was broken in the cicd build. Can you check if that binary contains the QtWebKit dylib?

Our cmake "standards"... well that's a big word. 
There's a lot of hacking around from some people with more and some people with less cmake expertise 😁

Forget to mention... We had that print dialog crash once before. But I thought that was fixed

Egbertdepauw:
> So printing on Mac should work (works in my own builds), but possibly was broken in the cicd build. Can you check if that binary contains the QtWebKit dylib?

No it doesn't and print menu and usermanual are also missing from the file / help menu. So I think it's a cmake thing.

image

I am using Subsurface 6.0.5231-CICD-release

dirkhh:
Yeah, just got back from dinner and checked on my computer - the build kit is missing QtWebKit. Back when I provided our macOS binaries built locally on a machine at my home, I had a hand-built QtWebKit for that (but that was against a much older version of Qt 5) and embarrassingly I never noticed that that's missing from `subsurface/Qt5.15.13`   🤦🏻‍♂️

@mikeller -- remind me how we created the binaries in that repo... are they just from the Qt distribution or are they build somewhere? There hasn't been an official QtWebKit package for Qt 5.x for a **long** time...

I turned off the machine at home where those builds used to run and I do recall that building QtWebKit against Qt 5.15.x is a bit painful. And the internet speed here on Fiji is definitely not conducive to me downloading the sources and kicking off builds. So I'll add that to my "to do when I'm back home next week" list 🤷‍♂️

dirkhh:
Oh wow. Look what some kagi searching found me... https://github.com/movableink/webkit
this even appears to include work on Qt6 support 👀 
Here's my attempt to kick of a conversation with the folks working on that: https://github.com/movableink/webkit/discussions/42

Egbertdepauw:
> Oh wow. Look what some kagi searching found me... https://github.com/movableink/webkit
> this even appears to include work on Qt6 support 👀

I am afraid this isn't going to be "the solution" Windows isn't really supported at the moment.
[windows support](https://github.com/movableink/webkit/pull/41#issuecomment-2313496784)

dirkhh:
I was mostly hoping for an easier way to get printing with Qt5.15 to work in the short term. 

And my understanding is that Windows will be a passion no matter what path we take with Qt6

atdotde:
Turns out my family vacation did not leave a lot of time for subsurface work...

Regarding the transition to JavaScript: the problem was the handling of the profile image: the (unfortunate) way we did this in the past was to have a<div> placeholder that we later replaced by the image tag. That replacement had to be done in JS, and that worked as well (to some degree). The problem was to inform when this is done and the actual images loaded so the page OS ready for shipping to the printer. The signal that is supposed to do that somehow did not quite do the right thing for Linux while it worked for me on Mac. I don't know of this works better now.

mikeller:
> @mikeller -- remind me how we created the binaries in that repo... are they just from the Qt distribution or are they build somewhere? There hasn't been an official QtWebKit package for Qt 5.x for a **long** time...

@dirkhh: https://github.com/subsurface/qt-mac was built by @notrege from the Qt source distribution, https://github.com/subsurface/qt-ios is just a copy of what we had hosted on Backblaze.

notrege:
Dirk - I think you mentioned trying to upgrade to QT6.   For Macs, would that be built on an X86 or Mn architecture?

dirkhh:
> Dirk - I think you mentioned trying to upgrade to QT6. For Macs, would that be built on an X86 or Mn architecture?

for now, I think I'll create fat binaries. the infrastructure to do so is already in the build scripts -- but is only lightly tested, and only for Qt6.

dirkhh:
> > @mikeller -- remind me how we created the binaries in that repo... are they just from the Qt distribution or are they build somewhere? There hasn't been an official QtWebKit package for Qt 5.x for a **long** time...
> 
> @dirkhh: https://github.com/subsurface/qt-mac was built by @notrege from the Qt source distribution

@notrege -- if you still have that build around somewhere, any chance you'd be able to add QtWebKit to it? 
Assuming things haven't changed too much in the last couple of years, from within an already built source tree, adding the `webkit-2.12` branch of https://github.com/movableink/webkit and then following their Mac build instructions **should** work.

notrege:
I can give that a try - probably not until next week.   

I'd like to try and help with QT6, but my Intel Mac is so old I can't install QT6 on it so I'm hesitant to try and work on a different architect than your build machine.  In theory it should work, but ...

Egbertdepauw:
> Regarding the transition to JavaScript: the problem was the handling of the profile image: the (unfortunate) way we did this in the past was to have a placeholder that we later replaced by the image tag. That replacement had to be done in JS, and that worked as well (to some degree)

@atdotde  Just curious but can't we place the image in "templatelayout" where we generate the html? and also the pagination? Maybe a dumb question, but for me it seems we generate the html and then parse it once again to place the images and apply some styles/pagination.

bstoeger:
IIRC, the problem is that the images load in a background thread. The main rendering function returns before the images are loaded. And there is no 'load synchronously' flag, which is pretty insane.

I can try to look into it if Robert has no time.

map-widget/qmlmapwidgethelper.cpp
```diff
@@ -162,7 +162,7 @@ void MapWidgetHelper::selectedLocationChanged(struct dive_site *ds_in)
 void MapWidgetHelper::selectVisibleLocations()
 {
 	QList<int> selectedDiveIds;
-	for (auto [idx, dive]: enumerated_range(divelog.dives)) {
+	for (auto [idx, dive] : enumerated_range(divelog.dives)) {
```
bstoeger, line 165: Please don't do gratuitous white space changes. In particular when they are inconsistent with the rest of the code base.

map-widget/qmlmapwidgethelper.cpp
```diff
@@ -4,20 +4,20 @@
...
-#include "qt-models/maplocationmodel.h"
+#include "qmlmapwidgethelper.h"
 #include "qt-models/divelocationmodel.h"
+#include "qt-models/maplocationmodel.h"
 #ifndef SUBSURFACE_MOBILE
 #include "desktop-widgets/mapwidget.h"
 #endif
 
-#define SMALL_CIRCLE_RADIUS_PX            26.0
+#define SMALL_CIRCLE_RADIUS_PX 26.0
```
bstoeger, line 20: I don't get the point of these whitespace changes.

If you want to touch this line, at least transform it into a proper constant, such as
```
static constexpr double small_circle_radius_px = 20.0;
```

Egbertdepauw: > I don't get the point of these whitespace changes.

The point is, that there is no point, I just made some quick changes and didn't notice I had the "prettier" plugin enabled, which autosaves files and changes whitespaces/indentation etc.

So that's all. And yes you are right I should have changed it back, but I didn't

dirkhh: > I don't get the point of these whitespace changes.
> 
> If you want to touch this line, at least transform it into a proper constant, such as
> 
> ```
> static constexpr double small_circle_radius_px = 20.0;
> ```

You know, I liked Egbert's change better. He at least didn't change the value :)

But joking aside - I had pointed that out in my first review and we established it was unintentional. All good.

Of course, I am the first to admit that I can be really sensitive on whitespace - and I'm super annoyed that new integrated IDEs all want to change the whitespace in my files... so one of the reasons I'm having posted a quick snapshot of what I have is that it's a complete mess, whitespace wise, because Qt Creator somehow decided to change things even though I have our settings loaded... grrr...

mikeller: Since we already have a discussion about whitespace going here this seems to be a good place to chip in. 😉

And I think Egbert has got a point here - probably somewhat unintentional, but a point nonetheless:
If we want to enforce whitespace standards, can we make them something that is structured enough that it can be tested (or even enforced) by a linter like `prettier`?

In the past I've always found that formatting rules that avoid dependencies on context (in the context of whitespace: vertical alignment) result in less 'noise' coming from code changes, and more compact pull requests - and this is typically what can be achieved relatively easy with linters.

dirkhh: We have a Qt Creator config in the coding standard document, iirc.
I'm happy to consider changes, but those need to be easily implementable for people using different tools on different platforms. 
We have developers on Linux and Mac (I doubt anyone is on Windows), and people use a wife array of tools. uemacs, vim, Qt Creator, VSC, etc. 
So yes, by all means, let's make things consistent, but don't break things for the people who have worked on this project for more than a dozen years...

mikeller: > We have a Qt Creator config in the coding standard document, iirc.

Not sure if it makes sense to ask everybody contributing to Subsurface to use Qt Creator?

> I'm happy to consider changes, but those need to be easily implementable for people using different tools on different platforms. We have developers on Linux and Mac (I doubt anyone is on Windows), and people use a wife array of tools. uemacs, vim, Qt Creator, VSC, etc.

In general I think it will make sense to go for a standalone tool like Prettier or `clang-tidy`, that can be run by developers independently of the IDE they are using, and can also be run in a GitHub Action, to automate pull request feedback about coding style.

> So yes, by all means, let's make things consistent, but don't break things for the people who have worked on this project for more than a dozen years...

I am a bit on the fence on this one - while I see the value of retaining _'we've always done it this way'_ workflows, I also see appeal in moving to a coding style that can be automated with tooling, and minimises the amount of effort and friction for our workflows around pull requests. And I suspect that there won't be an approach that fully satisfies both of these.

dirkhh: And in all honesty, the people who started this project (those still active) are not C++ programmers and the second generation of people engaged has already taken the project in a direction none of us are very familiar with. 

Half the code I can barely make sense of, most of the time. Which is a big part of why I have written less and less code in the last few years and mostly focused on the backend and infrastructure.

And after battling the very disruptive changes that Qt6 brings to QML projects, I'm not sure how much more I'll be able to constructively contribute. It seems that the fundamental way in which we integrate C++ code in QML has all been replaced by mechanisms that are designed around the idea that you have a QML app that gets a few additional features with dedicated Qt code. Not a C/C++ app that has a QML UI.

mikeller: > And in all honesty, the people who started this project (those still active) are not C++ programmers and the second generation of people engaged has already taken the project in a direction none of us are very familiar with.

I am not a C++ programmer either - but then I've always seen myself as a coder in <insert language that can get me what I want in the most effective way>, and (mostly) enjoyed the learning that came from it.

> And after battling the very disruptive changes that Qt6 brings to QML projects, I'm not sure how much more I'll be able to constructively contribute. It seems that the fundamental way in which we integrate C++ code in QML has all been replaced by mechanisms that are designed around the idea that you have a QML app that gets a few additional features with dedicated Qt code. Not a C/C++ app that has a QML UI.

I think this is more a 'sign of the times' than a direction that is particular to Qt / QML. html / CSS / JavaScript has become the 'lingua franca' that drives pretty much all multi - platform UI development, and QML is just the glue logic that goes in between Qt / C++ and the UI framework. And while using this isn't always a pleasurable experience, I suspect that this is something that is inherent to any UI design of sufficient complexity, and even more so if it strives to be multi-platform.

But to get back on topic, I think a sensible first step in the direction of automating our style guide will be to set up a GitHub action that checks changed lines in pull requests against a guide, and returns the result as a recommendation (i.e. doesn't fail) - we can then start with a template set of rules, and refine it to where we want it.
Thoughts?

bstoeger: > Half the code I can barely make sense of, most of the time.

If something is unclear, you are always welcome to ask.

Concerning white space, different people have different styles and I think that is OK. The current rough guidelines seem adequate to me.

When changing a source file, just use the overall style in there.

I have no problem with the odd whitespace fix in a commit (if it is consistent with the guidelines). However, when half of it are whitespace fixes, that makes reviewing just unnecessarily painful. Don't do it, please.

> It seems that the fundamental way in which we integrate C++ code in QML has all been replaced by mechanisms that are designed around the idea that you have a QML app that gets a few additional features with dedicated Qt code. Not a C/C++ app that has a QML UI.

If one can't have a thin display layer of QML on top of a C++ app, that's the point where I would start thinking about alternatives. Though I doubt it.

Egbertdepauw: > If one can't have a thin display layer of QML on top of a C++ app, that's the point where I would start thinking about alternatives. Though I doubt it.

I think you are right. There is more separation of GUI and backend code. Also everything is becoming more web based/web language oriented. QML is in fact just a QT Javascript front end language. A lot of apps nowadays are written in a web framework and brought to the desktop as an electron app.  

Even Vscode itself is an Electron app, which in fact is nothing more than a web app packaged together with a dedicated stripped browser.

Back on topic, the Usermanual seems no problem to get it working, searching etc. is working, the only problem I have at the moment is that we need to overload the class QWebEngineView function acceptNavigationRequest  because setLinkDelegationPolicy function isn't supported anymore. 

Concerning the printing a have been testing some code to use the  runJavaScript() function because the findAllElements() isn't supported anymore, I have it somewhat working, but it isn't pretty. Maybe it's better and more readable to parse the HTML as HTML text and do some styling and insert the profile image instead of trying to use the Webview to do it.

notrege:
Sorry, but no luck with building webkit-2.12.   I built QT-5.15.13 with txcode developer tools 12.4.  Webkit seems to require xcode to build so I installed 12.4 and ran build-webkit.  It throws an error about a missing HaveInternalSDK.xconfig.   Haven't had any luck resolving that - I suspect it's too new of an xcode version, but I don't want to play mix and match with the SDK components.

dirkhh:
Thanks for trying
I'm back home and will take a look as well

dirkhh:
OK, it looks like I have a working build that includes QtWebKit, supports printing, user manual, HID, MTP, FTDI, maps, etc.
**AND** runs on anything macOS 10.15 or newer. **AND** is signed.

Right now I have no idea how I could replicate **THAT** build on GitHub, but I'm hoping to at least get the Qt framework with QtWebKit in place so our automated binaries include printing again.

For those who'd like to test: https://subsurface-divelog.org/downloads/Subsurface-6.0.5252-10-15.dmg

notrege:
That sounds fantastic, but it doesn't seem to run on 10.15.7.

dirkhh:
That's... disappointing. I tried it here on a macOS 11 VM and it seemed to run there.
Silly question - you are sure you ran the Subsurface.app on that dmg? I'm sure you are. Hmm. Back to the drawing board, I guess.

notrege:
Unfortunately I'm pretty sure I installed the right .dmg.  Out for a dive,
but if there is something I can do to help, back in a few hours.

dirkhh:
I found the issue. And have a working dmg... but in validating THAT dmg I found a different issue... and I'm rebuilding the dmg again to address that. So please stay tuned. My x86 Mac is elderly and takes a while to run these builds (and they have to be run locally in order to be able to support 10.15)

dirkhh:
OK, here we are. I'm reasonably optimistic that this will in fact work.
(mostly because this time I downloaded and tested the signed DMG from the website instead of testing the local one :) )
https://subsurface-divelog.org/downloads/Subsurface-6.0.5253-for-10.15.dmg

notrege:
Installed and connected to my G2!

dirkhh:
so you are saying it works? I am really happy. thanks for testing.

notrege:
Very limited testing (it's close to 17:00 here), but it installed and connected so a good start - well done.

thuebner0427:
I've discovered your test version by accident while trying to install a new rolling release and use it since 2 month +/- without any issues on an old MacBook Pro with MacOS 10.15.7. 
I'm mainly using the mobile Android App and doing some minor editing on the Mac version. So if probably not tested all functions.
Anyhow, it would be great to have the pull request accepted allowing me to follow the usual rolling release on my old Mac.

mikeller:
@thuebner0427:

Thank you for your testing and feedback.

> I'm mainly using the mobile Android App and doing some minor editing on the Mac version.

Therein lies the problem: The currently proposed change is breaking the build for android, and from the loooks of it it will take some (potentially considerable) additional effort to extend this change to be buildable on all currently supported platforms:

image

PR #4307, opened 2024-09-01
planner: use std::move() to store planner notes

bstoeger:
Avoids a copy and makes coverity happy (rightfully so).

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Coverity fixes, including a memory leak.

@mikeller: Could you please check CID 442380, 442379. Especially the latter is too complex for me.

### DIFF
--- a/core/plannernotes.cpp
+++ b/core/plannernotes.cpp
@@ -143,9 +143,7 @@ void add_plan_to_notes(struct diveplan *diveplan, struct dive *dive, bool show_d
 		buf += format_string_std("<span style='color: red;'>%s </span> %s<br/>",
 			translate("gettextFromC", "Warning:"), message);
 
-		// TODO: avoid copy
-		dive->notes = buf;
-		dive->notes = strdup(buf.c_str());
+		dive->notes = std::move(buf);
 
 		return;
 	}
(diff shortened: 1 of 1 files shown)


mikeller (approved):

mikeller:
> Could you please check CID 442380, 442379. Especially the latter is too complex for me.

I was planning to have a look at the coverity problem reports once the bug in the tank display in the profile is fixed.

As for 442380, I was thinking about ripping out the weird custom number parsing and replace it with something that is drawing on C++ / Qt - do you have any recommendations?

I'll have to have a closer look at 442379 - I pretty much copy / pasted existing code, without trying to understand how it is working.

bstoeger:
> As for 442380, I was thinking about ripping out the weird custom number parsing and replace it with something that is drawing on C++ / Qt - do you have any recommendations?

Something like:
```
#include <charconv>
#include <cmath>

static int get_decimals(const char *begin, const char **endp, int decimals)
{
	double f;
	auto [ptr, ec] = std::from_chars(begin, begin + strlen(begin), f);
	if (ec != std::errc())
		return -1;
	*endp = ptr;

	f *= pow(10.0, decimals);
	return static_cast<int>(round(f));
}
```
??

Though attention: this changes behavior: `get_decimals("1.3599", &end, 2)` now (correctly?) rounds to 136, whereas the old code would give the (sketchy?) 135.  If you want old behavior, replace `round()` by `floor()`.

> I'll have to have a closer look at 442379 - I pretty much copy / pasted existing code, without trying to understand how it is working.

Yes, the planner code is not easy to read...

mikeller:
@bstoeger : 442380 is addressed in #4308
As for 442379, I can see where the warning is coming from, as we are decrementing `stopidx` and then accessing `stoplevels[stopidx]` repeatedly in a loop of 180 lines.
I could not figure out what the invariant is that we are using to prevent an underrun, but since the planner doesn't crash there must be one.

PR #4309, opened 2024-09-02
Playing around: make unit types smarter

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The unit types weren't the worst idea, as they make their base type explicit to the user. However, they were used very inconsistently, presumably because they are a pain to use.

This PR adds convenience operators so that the user needn't know the base type when adding  / subtracting.

The biggest gain will probably be comparison function, but for those I want to wait until we up the compiler version to C++20. Then one has to only implement a single `<=>` operator (a.k.a the space ship operator) instead of (`=`, `!=`, `<`, `>`, `<=`, `>=`). Then one should be able to use other convenience functions, such as `std::min`, `std::max` and `std:::clamp`.

Multiplication / division by scalars should obviously also be supported.

Note: related to #4237.

### DIFF
--- a/core/dive.cpp
+++ b/core/dive.cpp
@@ -469,7 +469,7 @@ static void update_min_max_temperatures(struct dive &dive, temperature_t tempera
  */
 static int same_rounded_pressure(pressure_t a, pressure_t b)
 {
-	return abs(a.mbar - b.mbar) <= 500;
+	return abs((a - b).mbar) <= 500;
 }
 
 static double calculate_depth_to_mbarf(int depth, pressure_t surface_pressure, int salinity);
@@ -678,7 +678,7 @@ static void fixup_duration(struct dive &dive)
 		if (logged || !is_dc_planner(&dc))
 			duration.seconds = std::max(duration.seconds, dc.duration.seconds);
 	}
-	dive.duration.seconds = duration.seconds;
+	dive.duration = duration;
 }
 
 static void fixup_watertemp(struct dive &dive)
@@ -724,7 +724,7 @@ static void fixup_dc_events(struct divecomputer &dc)
 			continue;
 		for (int idx2 = idx - 1; idx2 > 0; --idx2) {
 			const auto &prev = dc.events[idx2];
-			if (event.time.seconds - prev.time.seconds > 60)
+			if ((event.time - prev.time).seconds > 60)
 				break;
 			if (range_contains(to_delete, idx2))
 				continue;
@@ -1115,12 +1115,12 @@ static void merge_one_sample(const struct sample &sample, struct divecomputer &d
 
 			/* Init a few values from prev sample to avoid useless info in XML */
 			surface.bearing.degrees = prev.bearing.degrees;
-			surface.ndl.seconds = prev.ndl.seconds;
+			surface.ndl = prev.ndl;
 			surface.time.seconds = last_time + 20;
 
 			append_sample(surface, &dc);
 
-			surface.time.seconds = sample.time.seconds - 20;
+			surface.time = sample.time - duration_t { .seconds = 20 };
 			append_sample(surface, &dc);
 		}
 	}
@@ -1402,7 +1402,7 @@ static void sample_renumber(struct sample &s, const struct sample *prev, const i
 				s.pressure[j].mbar = 0;
 			} else {
 				s.sensor[j] = prev->sensor[j];
-				s.pressure[j].mbar = prev->pressure[j].mbar;
+				s.pressure[j] = prev->pressure[j];
 			}
 		} else {
 			s.sensor[j] = sensor;
@@ -1547,9 +1547,9 @@ static pressure_t merge_pressures(pressure_t a, pressure_t sample_a, pressure_t 
 static void merge_one_cylinder(cylinder_t *a, const cylinder_t *b)
 {
 	if (!a->type.size.mliter)
-		a->type.size.mliter = b->type.size.mliter;
+		a->type.size = b->type.size;
 	if (!a->type.workingpressure.mbar)
-		a->type.workingpressure.mbar = b->type.workingpressure.mbar;
+		a->type.workingpressure = b->type.workingpressure;
 	if (a->type.description.empty())
 		a->type.description = b->type.description;
 
@@ -1562,8 +1562,8 @@ static void merge_one_cylinder(cylinder_t *a, const cylinder_t *b)
 	a->end = merge_pressures(a->end, a->sample_end, b->end, b->sample_end, true);
 
 	/* Really? */
-	a->gas_used.mliter += b->gas_used.mliter;
-	a->deco_gas_used.mliter += b->deco_gas_used.mliter;
+	a->gas_used += b->gas_used;
+	a->deco_gas_used += b->deco_gas_used;
 	a->bestmix_o2 = a->bestmix_o2 && b->bestmix_o2;
 	a->bestmix_he = a->bestmix_he && b->bestmix_he;
 }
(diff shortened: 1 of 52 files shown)


core/dive.cpp
```diff
@@ -1148,12 +1148,12 @@ static void merge_one_sample(const struct sample &sample, struct divecomputer &d
...
 			/* Init a few values from prev sample to avoid useless info in XML */
 			surface.bearing.degrees = prev.bearing.degrees;
-			surface.ndl.seconds = prev.ndl.seconds;
+			surface.ndl = prev.ndl;
 			surface.time.seconds = last_time + 20;
 
 			append_sample(surface, &dc);
 
-			surface.time.seconds = sample.time.seconds - 20;
+			surface.time -= duration_t { .seconds = 20 };
```
mikeller: This seems to be incorrect as it removes `sample.time` from the assignment.

core/load-git.cpp
```diff
@@ -105,6 +105,11 @@ static pressure_t get_pressure(const char *line)
 	return p;
 }
 
+static o2pressure_t get_o2pressure(const char *line)
+{
+	return o2pressure_t { .mbar = static_cast<uint16_t>(lrint(1000 * ascii_strtod(line, NULL))) };
```
mikeller: Should `get_pressure()` use the same pattern for consistency?

core/dive.cpp
mikeller: I think this is wrong, as it removes `sample.time` from the assignment.

core/load-git.cpp
```diff
@@ -105,6 +105,11 @@ static pressure_t get_pressure(const char *line)
 	return p;
 }
 
+static o2pressure_t get_o2pressure(const char *line)
```
mikeller, line 104: Should `get_pressure()` be using the same pattern as this?

core/sample.h
```diff
@@ -21,7 +21,7 @@ struct sample                         // BASE TYPE BYTES  UNITS    RANGE
 	pressure_t pressure[MAX_SENSORS]; // int32_t  2x4    mbar   (0-2 Mbar)             cylinder pressures (main and CCR o2)
 	o2pressure_t setpoint;            // uint16_t   2    mbar   (0-65 bar)             O2 partial pressure (will be setpoint)
 	o2pressure_t o2sensor[MAX_O2_SENSORS];// uint16_t 6x2    mbar   (0-65 bar)             Up to 6 PO2 sensor values (rebreather)
-	bearing_t bearing = { -1 };       // int16_t    2  degrees  (-1 no val, 0-360 deg) compass bearing
+	bearing_t bearing = { .degrees = -1 };// int16_t    2  degrees  (-1 no val, 0-360 deg) compass bearing
```
mikeller, line 24: This is an example for where the 'vertical alignment' whitespace pattern comes apart for me - if we'd wanted to stick to the pattern we'd need to add 20+ lines with whitespace only changes to the pull request.

mikeller (commented):
See my comments for some potential bugs and possible improvements.

All in all I am not the biggest fan of the structs in `unit.h` - to me they are providing little benefit over being explicit with units when naming variables (like `time_s`), and have the potential to lead to errors when they are used inconsistently like before this pull request, like when values in mbar are implicitly initialised with a value in bar.

So this is definitely going into the right direction.

mikeller (approved):

core/dive.cpp
bstoeger: Yes, I misread sample as surface.

bstoeger:
> See my comments for some potential bugs and possible improvements.
> 
> All in all I am not the biggest fan of the structs in `unit.h` - to me they are providing little benefit over being explicit with units when naming variables (like `time_s`), and have the potential to lead to errors when they are used inconsistently like before this pull request, like when values in mbar are implicitly initialised with a value in bar.
> 
> So this is definitely going into the right direction.

I like the idea - forcing a base type and making it explicit when setting/reading seems reasonable.

However, the current implementation and usage is horrid. It completely defeats the purpose of a unit type if one has to state the base type for comparing and doing arithmetic.

glance-:
I really liked the declared types, and it made it way easier to ensure you actually passed the right types to the right functions. Sure, it made most sense in C code, where you'd might have had int get_usable_volume(int he, int o2, int cylinder_size, int pressure) and you easily could have mixed up the order. When doing it with volume_t get_usable_volume(struct gasmix mix, volume_t cylinder_size, presure_t pressure) you'd get a compiler error if you mixed them up.
There have been quite a few such bugs in this projects history.

Also, the fact that it's easy to see in which unit this depth, pressure and volume are have for me been a great helper so one doesn't mix them up. 

Now when we're moving to c++, it also makes sense to implement sane operations like these with types and between types where they make sense.

mikeller (approved):
Looks good to me.

bstoeger:
Opinions on the user-defined literals? @dirkhh? @atdotde?

dirkhh:
I'm not sure what you mean by user defined literals.
Things like 0_m and 0_bar?

bstoeger:
Yes, precisely: https://en.cppreference.com/w/cpp/language/user_literal

dirkhh:
Let's add this to the long list of "this is weird" and go on with it. I can never see me wrapping my mind around this, but I understand that this makes sense to people, so go ahead.

bstoeger:
I think it's a neat convenience feature and makes the code more readable. But it isn't important to me.

dirkhh:
Maybe my point is hidden behind my... writing patterns.
I agree that this is useful. I understand that for people who are not me this creates code that is easier to read.
So I was encouraging you to include this.

bstoeger:
> Maybe my point is hidden behind my... writing patterns.

No no, I was just giving my personal opinion pretty much independently of yours.

mikeller:
> I think it's a neat convenience feature and makes the code more readable. But it isn't important to me.

Agreed re readability.
I think the one thing that we need to keep in mind with operator overloads like these ones is that this can be confusing and hard to follow for casual contributors:
If they come across these constructs in the file they want to modify, it will not be obvious how it is done, and finding the location where the overload is defined will be hard for them, without knowledge of the code base.
Not sure if there is much we can do to help with this.

bstoeger:
Yes, at the very least there should be comments in units.h... Perhaps there also should be a chapter on the unit-types in the coding-style document.

mikeller:
> Yes, at the very least there should be comments in units.h... Perhaps there also should be a chapter on the unit-types in the coding-style document.

Good idea to put it into `CODINGSTYLE.md`. Having an explanation in `units.h` is good as well, but at least for new users it will be hard to figure out that this is where to look for this.

core/dive.cpp
```diff
@@ -2399,7 +2401,7 @@ bool dive::cache_is_valid() const
 pressure_t dive::get_surface_pressure() const
 {
 	return surface_pressure.mbar > 0 ? surface_pressure
-					 : pressure_t { SURFACE_PRESSURE };
+					 : pressure_t { .mbar = SURFACE_PRESSURE };
```
mikeller: While we are playing with unit types, should we replace / augment `SURFACE_PRESSURE` with something that allows us to flex our new unit system when using it?

bstoeger:
Let's commit so that I can rebase the other stuff without things diverging too much.

I'm still on the fence concerning the comparison operators. Switch to C++20 and implement a single `<=>` or do all the annoying `==`, `!=`, `<`, `>`, `<=`, `>=`?

dirkhh:
Just make sure you don't needless break older builds (like the ones for macOS 10.15). I don't think that compiler that I need to use for that supports C++20

PR #4311, opened 2024-09-04
planner: don't pass reference to asynchronous lambda

bstoeger:
A reference to a unique_ptr<> was captured by a lambda used to calculate variations in the background.

This is of course disastrous, because if the caller runs first it will delete the object. It's a wonder that this didn't crash regularly!?

The problem is that capturing unique_ptr<>s in lambdas works, but makes the lambda non-copyable. Sadly, the QtConcurrent::run() function wants to copy the lambda.

For now, do this by a release/reaquire pair. This is not exception safe. However, sine Qt doesn't support exceptions, we can live with that.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fixes a concurrency bug - see commit description. I really wonder why this never crashed - is it guaranteed that the called thread runs first?

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -1142,8 +1142,20 @@ void DivePlannerPointsModel::updateDiveProfile()
 		// Since we're calling computeVariations asynchronously and plan_deco_state is allocated
 		// on the stack, it must be copied and freed by the worker-thread.
 		auto plan_deco_state_copy = std::make_unique<deco_state>(plan_deco_state);
-		QtConcurrent::run([this, plan_copy, &plan_deco_state_copy] ()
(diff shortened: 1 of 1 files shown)


mikeller:
> This is of course disastrous, because if the caller runs first it will delete the object. It's a wonder that this didn't crash regularly!?

It does so, reproducibly:

`double free or corruption (!prev)`

It looks like nobody is actively using the 'Display plan variations' option at the moment.

mikeller (approved):
Tested, fixes crash when using 'Display plan variations'.

bstoeger:
Thanks.

PR #4313, opened 2024-09-06
C++-ification of diveplan

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

One of the few remaining C-style linked lists. Replace by an std::vector<>. The second commit might be wrong, I still haven't understood how the planner works.

Please check.

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -881,13 +881,7 @@ int DivePlannerPointsModel::addStop(int milimeters, int seconds, int cylinderid_
 
 	// add the new stop
 	beginInsertRows(QModelIndex(), row, row);
-	divedatapoint point;
-	point.depth.mm = milimeters;
-	point.time = seconds;
-	point.cylinderid = cylinderid;
-	point.setpoint = ccpoint;
-	point.minimum_gas.mbar = 0;
-	point.entered = entered;
+	divedatapoint point(seconds, milimeters, cylinderid, ccpoint, entered);
 	point.divemode = divemode;
 	divepoints.insert(divepoints.begin() + row, point);
 	endInsertRows();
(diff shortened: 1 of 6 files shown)


qt-models/diveplannermodel.cpp
```diff
@@ -1185,26 +1184,10 @@ void DivePlannerPointsModel::saveDuplicatePlan()
 	createPlan(true);
 }
 
-struct divedatapoint *DivePlannerPointsModel::cloneDiveplan(struct diveplan *plan_src, struct diveplan *plan_copy)
+struct divedatapoint *DivePlannerPointsModel::cloneDiveplan(struct diveplan &plan_src, struct diveplan &plan_copy)
```
mikeller: I think now that finding the last segment has become trivial we should be refactoring this to not return the unintuitive `divedatapoint *`, and instead leave fetching of the last segment to the caller.

bstoeger: You are right - but it just occurred to me that I made a big mistake:

last_segment should be the second to last, not the last element. Ooops!

mikeller: Yeah, the naming is slightly less than intuitive. I guess in a plan the second to last segment is the _last segment under water_.
But I stand by my point, let's make this SOLID, and remove the returning of a specific segment from cloning of the plan. Looking at where this is called from, this could probably do with some refactoring as well, and possibly avoid most of the cloning calls.

bstoeger: > Looking at where this is called from, this could probably do with some refactoring as well, and possibly avoid most of the cloning calls.

I'm not going to touch that, since I don't understand any of it.

mikeller (approved):
I've given the planner a quick spin with this, and apart from finding an existing bug that I introduced with #4245 the planner worked as expected.
There is a change in how the (invalid) profile is displayed when unavailable gases are selected (e.g. 'diluent' on an OC dive), but this isn't correct either way, and we should probably just blank the profile window in these cases.
In this case the textual dive plan is displaying an error message to make it clear that this is not a diveable profile.

PR #4315, opened 2024-09-08
media: load metadata and thumbnails of raw pictures using libraw

mikeller:
This will prevent the debian package from crashing if the user tries to add RAW images without having libraw installed.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/packaging/ubuntu/debian/control
+++ b/packaging/ubuntu/debian/control
@@ -47,7 +47,7 @@ Homepage: http://subsurface-divelog.org
 
 Package: subsurface
 Architecture: any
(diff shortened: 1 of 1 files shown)


bstoeger:
Thanks - I didn't know one could do PRs on PRs.

> This will prevent the debian package from crashing if the user tries to add RAW images without having libraw installed.

If a library is missing, won't the application simply not start?

PR #4318, opened 2024-09-10
Dirkhh for bstoeger raw

dirkhh:

- [x] Code cleanup
- [x] Build system change

- small `cmake` change to allow output of `pkg_config_library()` macro to be shown in CICD build output
- small fix to `get-version.sh` so it doesn't clutter the `cmake` output with strange errors about invalid commit names
- add libraw to various CICD builds (the macOS one is tested, the others I need to double check - hence this PR)

### Mentions:

@bstoeger @mikeller


bstoeger (approved):
Very nice, thanks!
(Not that I understand any of that, but it looks reasonable.)

dirkhh:
> Very nice, thanks! (Not that I understand any of that, but it looks reasonable.)

now you know how I feel with many of the C++ PRs 🤣
This is a PR against your branch as that's where it makes sense - so I'll leave it to you if / how / when you want to merge this.

dirkhh:
oh - the logic of when the builds run surprised me here. I had hoped that this PR would still create test builds so we can see if at least some of the changes actually work (the one for Copr won't be triggered until this gets merged, sadly)

.github/workflows/mac.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         submodules: recursive
 
     - name: setup Homebrew
-      run: brew install hidapi libxslt libjpg libmtp create-dmg confuse automake
+      run: brew install hidapi libxslt libjpg libmtp libraw create-dmg confuse automake
```
mikeller, line 27: Does this actually result in the entire libraw library being included in the .dmg that is built, or does this just add the library stubs so that Subsurface is built against libraw?
i.e. what happens when a version of Subsurface built in this way is run on a Mac that does not have the homebrew `libraw` installed when the user tries to add a RAW image?
The same applies for other distribution channels like COPR as well, where I think we need to declare a dependency on libraw in the manifest.

dirkhh: macOS binaries are very much like AppImages or Snaps. They contain all of the libraries that they depend on (with the exception of what is considered 'core system libraries'). So this treats `libraw` just like we treat `libxslt` or `libjpg` or `libpng`. They all get bundled with the `Subsurface.app`

mikeller: TIL

PR #4321, opened 2024-09-12
CICD: Add libraw to the MXE build container.

mikeller:
In support of #3954.

Also update MXE to the latest version.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh (approved):
this looks good.
remind me - how does the CICD-built Docker image actually get pushed to docker hub? That seems to still only show 3.1.0

mikeller:
> remind me - how does the CICD-built Docker image actually get pushed to docker hub? That seems to still only show 3.1.0

The build and push to the docker registry are triggered on pull request merge only. The build takes about 4 hours, so I feel we don't want to waste our CICD compute allocation by doing dry runs.

PR #4328, opened 2024-09-17
CICD: Add build for Debian Buster.

mikeller:
To avoid problems with the package published to the repository for Debian Buster, like the one fixed in #4326.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/.github/workflows/linux-debian-buster-5.15.yml
+++ b/.github/workflows/linux-debian-buster-5.15.yml
@@ -1,0 +1,19 @@
+name: Debian buster / Qt 5
+
+on:
+  push:
+    paths-ignore:
+    - scripts/docker/**
+    branches:
+    - master
+  pull_request:
+    paths-ignore:
+    - scripts/docker/**
+    branches:
+    - master
+
+jobs:
+  do-build-test:
+    uses: ./.github/workflows/linux-debian-generic.yml
+    with:
+      container-image: debian:buster


mikeller:
Not quite sure what packages need to be installed on top of what we are installing in the base image (https://github.com/mikeller/subsurface/blob/03e03e12fbf0d6d58e6d781c7ae449e35518bf2d/.github/workflows/linux-debian-generic.yml#L24-L34):

```
++ qmake -query QT_VERSION
qmake: could not find a Qt installation of ''
+ QT_VERSION=
Error: Process completed with exit code 1.
```

@dfxdj: Can you help here?

bstoeger:
Do we really want to continue supporting this? According to https://www.debian.org/releases/buster/:

> Security updates have been discontinued as of June 30th, 2022.

and

> Buster also has benefited from Long Term Support (LTS) until 30th June, 2024

If someone insists on using an abandoned distribution, they probably can compile from source (if they require the latest version of a divelog software of all things)...

mikeller:
> Do we really want to continue supporting this? According to [debian.org/releases/buster](https://www.debian.org/releases/buster/):
> 
> > Security updates have been discontinued as of June 30th, 2022.
> 
> and
> 
> > Buster also has benefited from Long Term Support (LTS) until 30th June, 2024

That's a good point actually - enabling people to use insecure software doesn't feel right in the 'big picture'.

dfxdj:
I concur that there's not much point in carrying support for Buster. I'm only still building it because I have it already set up and it still works. Once there are fatally breaking changes I will also drop it.

dfxdj:
> I concur that there's not much point in carrying support for Buster. I'm only still building it because I have it already set up and it still works. Once there are fatally breaking changes I will also drop it.

.... which may be now because there are more instances of the same error:

```
/build/subsurface-beta-202409170411/core/datatrak.cpp: In function 'void wlog_compl_parser(std::__cxx11::string&, dive*, int)':
/build/subsurface-beta-202409170411/core/datatrak.cpp:621:90: error: could not convert '{{(tmp * 10)}, translate(((const char*)"gettextFromC"), ((const char*)"unknown")), false}' from '<brace-enclosed initializer list>' to 'weightsystem_t'
   weightsystem_t ws = { {.grams = tmp * 10}, translate("gettextFromC", "unknown"), false };
                                                                                          ^
```

bstoeger:
To be fair, the designated initializers are a non-standard GNU extension. So it is no wonder that some compilers refuse.
Probably we should change this to somthing like
```
        return duration_t::from_seconds(xxx);
```
etc.

mikeller:
@dfxdj: 

> .... which may be now because there are more instances of the same error:

I suspect that this will probably happen every time somebody changes something to do with our implementation of units. So unless we manage to get a CICD build for Buster to work, this will result in us playing whack-a-mole with these to keep the Buster build alive.


@bstoeger:

> To be fair, the designated initializers are a non-standard GNU extension. So it is no wonder that some compilers refuse. Probably we should change this to somthing like
> 
> ```
>         return duration_t::from_seconds(xxx);
> ```

Always good to use the most commonly used form.

dfxdj:
So what's the consensus here? Should we continue to amend the unit patches to make older gcc happy, or abandon Buster?

bstoeger:
> So what's the consensus here?

Since you are doing the work - I would say that it is up to you.

mikeller:
Happy to leave this up to you @dfxdj , but if we keep going we should probably put in the work to get a CICD build for Buster going so that we can shif the onus of making proposed changes work with Buster onto the contributors proposing the changes.

dfxdj:
> Happy to leave this up to you @dfxdj , but if we keep going we should probably put in the work to get a CICD build for Buster going so that we can shif the onus of making proposed changes work with Buster onto the contributors proposing the changes.

Ok, try with adding `qt5-default` to the installation list.

```
$ qmake -query QT_VERSION
qmake: could not find a Qt installation of ''
$ sudo apt install qt5-default
...
Unpacking qt5-default:amd64 (5.11.3+dfsg1-1+deb10u6) ...
Setting up qt5-default:amd64 (5.11.3+dfsg1-1+deb10u6) ...
$ qmake -query QT_VERSION
5.11.3
```

FTR this is my control file for Buster: https://github.com/dfxdj/subsurface-debian/blob/master/beta/buster/debian/control

dirkhh:
The desire to support older versions is always a good one.
As many of you know, I keep trying to provide binaries for (relatively) ancient macOS versions (where the need is more pronounced with hardware that actively gets abandoned by Apple in newer versions).
That said, Buster is officially EOL (https://wiki.debian.org/LTS/Buster) and I find it a lot harder to identify why someone would be prevented from upgrading to a newer version.
So my take here would be "it would be nice to get this to work, but the amount of extra work needed should be proportional to the benefit (which I expect to be at best a handful of people)".
All that said, the current build fails because it can't find Qt, so this is a completely unrelated issue. @dfxdj if you need help with setting up a CICD build, let me know and I can try to poke at his...

dfxdj:
Alright then. Based on my access logs, the number of Buster users is approximately zero. Time to let it die.

dirkhh:
> Alright then. Based on my access logs, the number of Buster users is approximately zero. Time to let it die.

now THERE's a convincing argument.
My logs show about 50 people on macOS 10.13 (EOL in 2020) and about half a dozen on 10.11 (EOL 2018 - using a 4y/o version of Subsurface to do so) accessing the cloud storage servers... sadly my tooling isn't clever enough to tell the version of Linux that people are using :(

mikeller:
> sadly my tooling isn't clever enough to tell the version of Linux that people are using :(

\<BOFH>
`nmap -O -v ${SOURCE_IP}` is your friend 😉
\</BOFH>

dirkhh:
I think a LOT of users (me included) would find that rather offensive.

PR #4343, opened 2024-09-29
Make libraw support optional

salvadorcunat:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [X] Build system change
- [ ] Documentation change
- [ ] Language translation

Build support by default if libraw is found, but make it optional for builds were it's not needed, like smtk2ssrf or subsurface-downloader.

### Changes made:
1) Add an option to turn on/off support (on by default)
2) Set libraw support to off for smtk2ssrf and subsurface-downloader
3) Remove libraw pkgconfig search in smtk2ssrf (as a bonus remove search for libftdi too).

### Related issues:

As commented in #3954 
### Mentions:

@bstoeger @mikeller @atdotde


mikeller (approved):
Nice work!
And this is building just fine in CICD, so it looks like the smtk2ssrf build does not care in the end that `libsubsurface_corelib` that is used comes with support for libraw and libftdi.

salvadorcunat:
It doesn't. smtk2ssrf-build.sh rebuilds subsurface stripping as much as possible as there are a lot of things in corelib it doesn't need.

mikeller:
> It doesn't. smtk2ssrf-build.sh rebuilds subsurface stripping as much as possible as there are a lot of things in corelib it doesn't need.

:laughing: 

In this case our CICD scripts contain outdated warnings:

https://github.com/subsurface/subsurface/blob/821f3fc551bb574e1f2290ff21910b80bafb537d/.github/workflows/linux-debian-generic.yml#L98

PR #4354, opened 2024-10-23
Make the "Save dive data as subtitles" feature more configurable.

vigge93:
Previously, the subtitle generation was hardcoded, making it unsuitable if you didn't want all of the displayed values. This has been replaced by a format string that is configurable in the settings, using predefined tags that are replaced with the values. The default value for this has been set to (mostly) match the currently generated subtitle string. This also provides a good starting point for users that want to modify the string.

- [ ] Bug fix
- [X] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Previously, the subtitle generation was hardcoded, making it unsuitable if you didn't want all of the displayed values. This has been replaced by a format string configurable in the settings, using predefined tags replaced with the corresponding values. The default value for this has been set to (mostly) match the currently generated subtitle string. This also provides a good starting point for users who want to modify the string. Further, a small fix has been added where the subtitles now start at the beginning of the video, and not at the first datapoint after the video starts. A table of the available tags has also been added directly below the input field.

### Changes made:

1) Added "Subtitle format string" and "Available tags" under a new "Dive data as subtitles" section in the Media preferences.
2) Added `subtitles_format_string` to `preferences` struct.
3) Updated `format_st_event` to use the `subtitles_format_string` and replace the tags.
4) Added translation messages to translation source.

### Release note:

Release note has been added at top of `CHANGELOG.md`

### Documentation change:

New setting in media preferences to specify the subtitle format string.

### DIFF
--- a/core/save-profiledata.cpp
+++ b/core/save-profiledata.cpp
@@ -47,10 +47,10 @@ static std::string video_time(int secs)
 	return format_string_std("%d:%02d:%02d.000,", hours, mins, secs);
 }
 
-void replace_all(std::string& str, const std::string& old_value, const std::string& new_value) {
+static void replace_all(std::string &str, const std::string &old_value, const std::string &new_value) {
     if (old_value.empty())
         return;
-    size_t start_pos = std::string::npos;
+    size_t start_pos = 0;
     while ((start_pos = str.find(old_value, start_pos)) != std::string::npos) {
         str.replace(start_pos, old_value.length(), new_value);
         start_pos += new_value.length(); // In case 'new_value' contains 'old_value', like replacing 'x' with 'yx'
--- a/core/save-profiledata.cpp
+++ b/core/save-profiledata.cpp
@@ -6,7 +6,9 @@
 #include "core/errorhelper.h"
 #include "core/file.h"
 #include "core/format.h"
+#include "core/gettext.h"
 #include "core/membuffer.h"
+#include "core/pref.h"
 #include "core/subsurface-string.h"
 #include "core/version.h"
 #include <errno.h>
@@ -43,6 +45,16 @@ static std::string video_time(int secs)
 	int mins = secs / 60;
 	secs -= mins * 60;
 	return format_string_std("%d:%02d:%02d.000,", hours, mins, secs);
+}
+
+void replace_all(std::string& str, const std::string& old_value, const std::string& new_value) {
+    if (old_value.empty())
+        return;
+    size_t start_pos = std::string::npos;
+    while ((start_pos = str.find(old_value, start_pos)) != std::string::npos) {
+        str.replace(start_pos, old_value.length(), new_value);
+        start_pos += new_value.length(); // In case 'new_value' contains 'old_value', like replacing 'x' with 'yx'
+    }
 }
 
 static void put_pd(struct membuffer *b, const struct plot_info &pi, int idx)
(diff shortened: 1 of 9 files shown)


translations/subsurface_source.ts
```diff
@@ -7258,6 +7258,186 @@ See https://doc.qt.io/archives/qt-4.8/qdatetime.html#fromString</source>
         <source>One can specify te relative position within the video from where the thumbnail is retrieved. Use the slider below do do that.</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
```
mikeller: The translations are extracted automatically, so this part is not needed (and might interfere with adding the generated translations).

vigge93: Alright, I have reverted the changes.

atdotde:
From what I see this is a great improvement (if in fact there are people that use this feature)! Thanks a lot.

core/save-profiledata.cpp
```diff
@@ -175,32 +187,214 @@ static std::string format_st_event(const plot_data &entry, const plot_data &next
 	double value;
 	int decimals;
 	const char *unit;
-
-	if (entry.sec < offset || entry.sec > offset + length)
+	if (next_entry.sec < offset || entry.sec > offset + length)
```
mikeller, line 190: Why are we relaxing this condition here? Presumably if `entry.sec` is smaller than `offset` we are outside of the video timeline, even if `next_entry.sec` is not smaller than `offset`.

core/save-profiledata.cpp
```diff
@@ -175,32 +187,214 @@ static std::string format_st_event(const plot_data &entry, const plot_data &next
...
+	}
+
+	if (entry.density > 0) {
+		replace_all(format_string, "[density]", format_string_std("%.1fg/ℓ", entry.density));
+	} else {
+		replace_all(format_string, "[density]", "");	
+	}
+
+	if (entry.icd_warning) {
+		replace_all(format_string, "[icd_warning]", translate("gettextFromC", "ICD in leading tissue"));
```
mikeller, line 392: Seems to be a bit wordy, I am worried that the subtitles will be displayed in a very small font if this is ever hit during a dive.

mikeller (approved):
I like the idea.

However, it looks like there's still a bug of some sort in this, a in a quick test the placeholders did not get updated for me:

Screenshot 2024-10-24 161906

Testing the same dive / video with current master was working fine:

Screenshot 2024-10-24 162413

I also think that it would make sense to add a mention of this to the user manual (or even move the full list of supported placeholders there): https://github.com/subsurface/subsurface/blob/1a9b09682ad4bb99a4526734c5b8041f884d47af/Documentation/user-manual.txt#L1817

mikeller (changes requested):
See comments.

core/save-profiledata.cpp
vigge93: It's the fencepost problem. If we have one datapoint at 35:20 and one at 35:40, but the video starts at 35:30, we need to use the data from the 35:20 datapoint for the first 10 seconds of the clip.

core/save-profiledata.cpp
vigge93: I agree that it may be a bit verbose. Something like "ICD warning" might be better. However, "ICD in leading tissue" is used in the plot, so I don't know if we want to keep the same formulation for consistency?

vigge93:
> I like the idea.
> 
> However, it looks like there's still a bug of some sort in this, a in a quick test the placeholders did not get updated for me:
> 
> Screenshot 2024-10-24 161906
> 
> Testing the same dive / video with current master was working fine:
> 
> Screenshot 2024-10-24 162413
> 
> I also think that it would make sense to add a mention of this to the user manual (or even move the full list of supported placeholders there):
> 
> https://github.com/subsurface/subsurface/blob/1a9b09682ad4bb99a4526734c5b8041f884d47af/Documentation/user-manual.txt#L1817

Yes, a small bug managed to sneak its way into the `replace_all` function while I was cleaning up the code. It has been fixed now. I have also updated the user manual and added the list of tags. I still think there is value in keeping it in the settings as well as a quick reference. Is there any way to make it collapsible so it doesn't have to be present all of the time?

core/save-profiledata.cpp
```diff
@@ -45,6 +47,16 @@ static std::string video_time(int secs)
 	return format_string_std("%d:%02d:%02d.000,", hours, mins, secs);
 }
 
+void replace_all(std::string& str, const std::string& old_value, const std::string& new_value) {
```
bstoeger: I know it's a "religious" issue, but as per coding-style, we put the`&` with the variable, not the type. Rationale: That's how the C++ grammar works, like it or not.

Also I suggest making this a global helper function.

If not a global helper, it should be of static linkage.

core/save-profiledata.cpp
```diff
@@ -45,6 +47,16 @@ static std::string video_time(int secs)
...
+    if (old_value.empty())
```
bstoeger, line 51: This check seems unnecessary to me. If `old_value` is empty, the loop body will simply not execute. The common case will be a non-empty string, so no point in wasting code on this.

mikeller:
@vigge93:

> I still think there is value in keeping it in the settings as well as a quick reference. Is there any way to make it collapsible so it doesn't have to be present all of the time?

Makes sense. For now there is plenty of space available in the 'Media preferences' tab. If we ever add more stuff there we can always remove or shorten the list, and put in a reference to the manual.
I am not a Qt expert, so I don't know how to do collapsible content in Qt.

mikeller (approved):
Thanks for fixing, this is working well now.

One thought I had when testing: It would be an improvement if it was possible to have an optional prefix for individual datapoints, at least for the ones that are not always displayed, so that something like 'ceiling=[ceiling]' would be hidden in its entirety if there was no ceiling. Maybe encoded as `{ceiling=[ceiling]}`. But this can be a future extension.

mikeller (approved):
Thanks for fixing, this is working well now.

One thought I had when testing: It would be an improvement if it was possible to have an optional prefix for individual datapoints, at least for the ones that are not always displayed, so that something like 'ceiling=[ceiling]' would be hidden in its entirety if there was no ceiling. Maybe encoded as `{ceiling=[ceiling]}`. But this can be a future extension.

core/save-profiledata.cpp
vigge93: If `old_value` is empty, it will always match at `start_pos`, resulting in it getting stuck in an infinite loop.

core/save-profiledata.cpp
vigge93: Fixed. Made the function static.

core/save-profiledata.cpp
bstoeger: Yes correct - I thought this was testing the modified string.

vigge93:
I think this should be ready to merge once the workflows have been run @mikeller.

PR #4362, opened 2024-11-03
export-html: prevent invalid json

dirkhh:
A dive without events could lead to json with extra closing brackets that prevents browsers from rendering the dive log.

- [x] Bug fix

we had three different ways with how we dealt with writing the `key` part of the json key/value when writing out HTML. One of them was flat out wrong in the case that the structure was empty. A dive with no events would have an extra `],` at the end, which caused the browser to fail to parse the json and no dive list to be shown.

With this change we always write `"events": [],` (and add events within the brackets if there are any), which should ensure correct json.

### DIFF
--- a/core/save-html.cpp
+++ b/core/save-html.cpp
@@ -92,7 +92,8 @@ static void write_dive_status(struct membuffer *b, const struct dive &dive)
 
 static void put_HTML_bookmarks(struct membuffer *b, const struct dive &dive)
 {
-	const char *separator = "\"events\":[";
+	const char *separator = "";
+	put_string(b, "\"events\":[");
 	for (const auto &ev: dive.dcs[0].events) {
 		put_string(b, separator);
 		separator = ", ";


mikeller (approved):

core/save-html.cpp
```diff
@@ -92,7 +92,8 @@ static void write_dive_status(struct membuffer *b, const struct dive &dive)
 
 static void put_HTML_bookmarks(struct membuffer *b, const struct dive &dive)
 {
-	const char *separator = "\"events\":[";
```
bstoeger, line 95: Actually I think this is the wrong fix.

The bug was introduced in 27dbdd35c6d5d9d8c2d3c607405376ef4db6e60f when removing the check for "no events".

I would just add
```
        if (dive.dcs[0].events.empty())
                return;
```
to restore the old behavior.

Adding the new version on top might be a good idea anyway - the old code was very confusing.

bstoeger: That said - I just quickly skimmed over the source file and it is very inconsistent on how to treat empty lists. :( This should be unified.

dirkhh: I'd agree that this isn't the only possible fix - I'm just not sure if it's the _wrong_ fix.
I see no strong argument either way... do we want an empty array for the events (ensuring that the events array always exists), or is it better not to have the array. I see other spots in the code where we are making the wrong choice in this matter. For example, the html dive list table doesn't render correctly because elements are missing instead of being empty.

bstoeger: Yes, with "wrong" I meant not restoring old behavior before the breaking commit.

As I said, this should all be unified.  I would go as far as writing a template that outputs a list. Something like `void output_list(const char *name, const T &list, FUNC output_function);`
Optionally a `bool output_empty_list = false` parameter.

Issue #4364, opened 2024-11-07
Divesoft Freedom+ log support (header "DiVE")


jakubhecko:
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [x] New feature request
- [ ] Discussion request

### Issue long description:

Add support for Divesoft logs (*.DLF) with new header format starting with signature "DiVE".
This is a log header format only, the rest of the dive profile data format is identical.
Main differences from a v1 header "DivE":

-  Length of the header - 64B instead of 32B
- Time is handled as a separate UTC time and a timezone
- Added average depth, surface interval and device S/N
- Some values are expanded to achieve higher max value and/or better precision.

This log header format is already supported (more than a year) as a part of libdivecomputer, unfortunately Subsurface uses a custom parser for "Import / Import log files".

Some users tried already adjusting the dive log signature from "DiVE" to "DivE", with some success, unfortunately the extra 32B of header makes for erroneous data shown in the dive profile.
https://groups.google.com/g/subsurface-divelog/c/FWh8gwfE444

### Operating system:

Not relevant, this is a parser request.

### Subsurface version:

(CICD) Release: 6.0.5285

Official release

### Steps to reproduce:

Open new logbook, try importing a log file "Import / Import log files" from any Freedom+.

### Current behavior:

The dive is not imported.

### Expected behavior:

Dive being imported and properly parsed.

### Additional information:

See the log file and "DiVE" header format attached [00000031_DiVE.zip](https://github.com/user-attachments/files/17657890/00000031_DiVE.zip)

### Mentions:

@jefdriesen

glance-:
The subsurface custom parser of the Divesoft files is way older than the support in libdivecomputer, and was a reverse engineered parser, and not based on any information from Divesoft.

It has outlived its usefulness and probably the best way would be to throw it away and replace it with the libdivecomputer parser.

mikeller:
> It has outlived its usefulness and probably the best way would be to throw it away and replace it with the libdivecomputer parser.

Agreed on this - and it should probably be relatively straightforward to add support for the file system transport to this dive computer model in libdivecomputer.

glance-:
Shure, file system transport would be nice for "downloading" , but I think the shorter path to success would be to replace the content of parse_dlf_buffer() with a call to libdc_buffer_parser, and just replace the content parser with the one from libdivecomputer, just like it's done for ostctools_import(), datatrak_import() and smartrak_import().

[...]

PR #4367, opened 2024-11-13
Import: Add Parsing for Divesoft Freedom+ Format Files.

mikeller:
Switch the Divesoft log file importer to use the parser in
libdivecomputer. This adds support for the newer Divesoft Freedom+ log
file format.
Also refactor the OSTCTools log file importer to share common
functionality.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


mikeller:
Fixes #4364.

mikeller:
@salvadorcunat: As the original author of the OSTCTools file import, are you able to supply some sample files to test this importer - as it turns out this tool does not seem to support my Heinrichs Weikamp dive computer (OSTC4).

salvadorcunat:
> @salvadorcunat: As the original author of the OSTCTools file import, are you able to supply some sample files to test this importer - as it turns out this tool does not seem to support my Heinrichs Weikamp dive computer (OSTC4).

Hello @mikeller. Yes, I've plenty of them. But my DC is an OSTC 2N which uses older model for dive logging. 
In regular directory subsurface/dives should be 2 sample dives.
I think the problem must be coming from the change to HWOs in OSTC3 and newer models. This must have forced some changes in OSTC Tools too, as in [his web page](https://angermayr.eu/index.php/en/software-en/ostctools-win/ostctools-download) developer has two different tools for both OSs. 

I could take a look at it this weekend as we are having pretty rough weather here in Valencia. but I don't have a single sample of a dive with the HWOs.

mikeller:
Thanks @salvadorcunat - the files in `/dives` were sufficient for me to test the OSTCTools import functionality - and it turns out I broke it. I'll fix this.

core/divesoft.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "divelist.h"
+#include "divelog.h"
+#include "extradata.h"
+#include "format.h"
+#include "libdivecomputer.h"
+
+int divesoft_import(const std::unique_ptr<std::vector<unsigned char>> &buffer, struct divelog *log)
+{
+	int model = 0;
+	if (strncmp((char *)(buffer->data() + 52), "7026", 4) == 0)
```
glance-: Please add some comments explaining the magic constants.

I who've reversed engineered this format once don't know what you're checking here.

core/divesoft.cpp
```diff
@@ -0,0 +1,39 @@
...
+		model = 10;
```
glance-: Please add a comment saying that its is Liberty, and the same for the 19 bellow for Freedom.

core/file.cpp
```diff
@@ -267,6 +267,44 @@ bool remote_repo_uptodate(const char *filename, struct git_info *info)
...
+	const char *failed_to_read_msg = translate("gettextFromC", "Failed to read '%s'");
+
+	struct stat file_status;
+	if (stat(file, &file_status) < 0) {
+		report_error(failed_to_read_msg, file);
+
+		return NULL;
+	}
+
+	// Open the archive
```
glance-: archive feels just wrong here. "log" or why not just file, fd, f or something less confusing?

core/divesoft.cpp
glance-: I actually think this whole model bit here is unnecessary, because all the divesoft models use the same parser, and the log will get the right model anyway from the DEVINFO event.

mikeller:
@salvadorcunat: I have found and fixed the problem with importing OSTCTools files.

core/divesoft.cpp
mikeller: > I actually think this whole model bit here is unnecessary, because all the divesoft models use the same parser, and the log will get the right model anyway from the DEVINFO event.

libdivecomputer does not allow us to _not_ specify a model here - and supplying what is potentially the incorrect model seems to be a questionable approach.
Even if the Divesoft parser does not depend on the model information _at the moment_, there are no guarantees that this will not change in a future version, and introduce incorrect parsing when libdivecomputer is updated.
Also, note that the `DEVINFO` event is emitted by the Divesoft device handler, and not by the parser, so this doesn't happen when we are reading log files.

glance-: You just need to pass a valid value, no reason to pass a specific value. I'd prefer if we'd skip the guessing game. The only relevant thing is to pass the correct dc_family_t.

Also, this variant of guessing game based on serial is only valid for the new log file format, and not for the old one.

That whole thing is kinda why we'd drop the old parser and accept the loss of functionality to user the libdivecomputer one.

The actual model is available in the cfg-sample, but that would involve more parsing of the file, which was exactly what we didn't want to do.

Also, just looking at the amount of information that's missing in the libdivecomputer parser makes me question the chose to drop our current parser.

mikeller: > Also, just looking at the amount of information that's missing in the libdivecomputer parser makes me question the chose to drop our current parser.

I don't mind either way, if somebody wants to put in the effort to add support for the models currently sold by Divesoft to our standalone logfile parser then that's ok with me.

[This problem](https://groups.google.com/g/subsurface-divelog/c/FWh8gwfE444) has been reported to us two months ago, and nothing has happened on our side, so it does not look like there is motivation to get our own implementation of this parser updated, and likely also not to put in the effort to incorporate any future changes in the format. During the same two months, libdivecomputer has updated their parser to support the current models, so we can get this essentially 'for free', together with any future updates.

In general, seeing that we are struggling to find the people willing to do the work to keep this project maintained, it seems to me that keeping two parsers for the same format maintained is a wasteful approach that doesn't fit with where we are at. If we are not happy with the scope of functionality in the parser included in libdivecomputer then it might be a better use of time to add the functionality that we would like to have to it, instead of maintaining an additional implementation of the same parser.

dirkhh: I couldn't agree more.
Duplicative work is the last thing we want to do.

core/file.cpp
```diff
@@ -304,8 +341,13 @@ int parse_file(const char *filename, struct divelog *log)
 	}
 
 	/* Divesoft Freedom */
-	if (fmt && (!strcasecmp(fmt + 1, "DLF")))
-		return parse_dlf_buffer((unsigned char *)mem.data(), mem.size(), log);
+	if (fmt && (!strcasecmp(fmt + 1, "DLF"))) {
+		auto buffer = read_into_buffer(filename);
```
glance-, line 345: We've already read the file once, just above:
auto [mem, err] = readfile(filename);

Why read it here again with another function into another format?

mikeller: The behaviour is copied from the pattern used when importing OSTCTools logfiles through libdivecomputer here: https://github.com/subsurface/subsurface/pull/4367/files#diff-a83ea946c1779a430769c854cf7ebf7df3ee91e20c057271eef4354d50e861f8L324

I assumed that there was a valid reason that this was not re-using the buffer above.
If you think that this can be improved then feel free to open a pull request to implement the improvement.

PR #4369, opened 2024-11-14
planner: fix calculations of variations [potential fix for #4368]

### LINKED ISSUE #4368: DIve Planner: Stop times per additional metre and minute always zero
stoptimes

### Describe the issue:

- [X] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

I have been using Subsurface to plan a lot of dives. A useful feature has always been the deco penalty per additional metre of depth and minute stayed at maximum depth. However, upon upgrading from an earlier version to the latest relase and/or nightly build, I discovered that these values are now always zero, no matter the dive plan.

### Operating system:

Windows 11, English.

### Subsurface version:

Subsurface 6.0.5286-CICD-release

### Steps to reproduce:

1) Go to Dive Planner
2) Create any deco dive plan
3) Check out the top of the dive plan window; additional deco per meter depth and minute will read 0:00

### Current behavior:

These values do not show.

### Expected behavior:

The values are calculated and displayed, as was the case a few version prior.

### Additional information:

See attached example dive plan.

bstoeger:
In 8704a8b6f97920090d144919b4546344e7df8bd8 the code in cloneDiveplan() was replaced by a simple assignement statement.

Alas, the original code was more complex: it copied only up to a certain point (it stopped at automatically generated steps).

The new behavior made the calculations of variations fail.

Therefore, readd a function that copies a dive plan, but "forgets" the automatically added stops.

Fixes #4368

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Potential fix for #4368. See commit description and discussion in #4368. Please check / comment.

### Changes made:

1) Remove automatically added stops when copying dive plan.

### Related issues:

Hopefully fixes #4368.

### DIFF
--- a/core/planner.cpp
+++ b/core/planner.cpp
@@ -74,6 +74,17 @@ diveplan::~diveplan()
 {
 }
 
+diveplan diveplan::copy_planned_segments() const
+{
+	diveplan res = *this;
+
+	// Remove non manually added entries
+	auto it = std::find_if(res.dp.begin(), res.dp.end(),
+			       [] (const divedatapoint &dp) { return dp.time && !dp.entered; });
+	res.dp.erase(it, res.dp.end());
+	return res;
+}
+
 bool diveplan::is_empty() const
 {
 	return std::none_of(dp.begin(), dp.end(), [](const divedatapoint &dp) { return dp.time != 0; });
--- a/core/planner.h
+++ b/core/planner.h
@@ -53,6 +53,7 @@ struct diveplan {
 	bool is_empty() const;
 	void add_plan_to_notes(struct dive &dive, bool show_disclaimer, planner_error_t error);
 	int duration() const;
+	diveplan copy_planned_segments() const; // copy diveplan, but ignore automatically added stops
 };
 
 struct deco_state_cache;
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -1127,7 +1127,7 @@ void DivePlannerPointsModel::updateDiveProfile()
 	if (isPlanner() && shouldComputeVariations()) {
 		auto plan_copy = std::make_unique<struct diveplan>();
 		lock_planner();
-		*plan_copy = diveplan;
+		*plan_copy = diveplan.copy_planned_segments();
 		unlock_planner();
 #ifdef VARIATIONS_IN_BACKGROUND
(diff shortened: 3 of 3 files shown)


atdotde (approved):
You beat me to it by 20 minutes. https://github.com/subsurface/subsurface/commit/bcda8ac12a5d6808d89c37e6a419b28763eb0bd0

core/planner.cpp
```diff
@@ -74,6 +74,17 @@ diveplan::~diveplan()
 {
 }
 
+diveplan diveplan::copy() const
```
mikeller: `copy()` seems to be creating the impression that this does actually a copy of the data structure (from a data point of view), and not a copy of the planned segments only. Should this rather be `copyPlannedSegments()`?

mikeller (commented):
It definitely brought back the 'penalties', but I suspect something is wrong with the calculations now, as the penalties are larger now than they were before:

image

6c8f158569ede6712cb3c30a89af874467e884a0 for comparison:

image

(a way to assess the accuracy of the predicted penalties without having to build multiple versions is to just add 1m / 1minute to the last planned segment, and check that the new runtime is congruent with adding the penalty to the original runtime, which isn't the case with the proposed changes)

atdotde:
In https://github.com/subsurface/subsurface/commit/bcda8ac12a5d6808d89c37e6a419b28763eb0bd0 I had to delete the automatic datapoints in all five iterations. But the numbers are closer to the old ones of @mikeller .

mikeller:
> You beat me to it by 20 minutes. [bcda8ac](https://github.com/subsurface/subsurface/commit/bcda8ac12a5d6808d89c37e6a419b28763eb0bd0)

https://github.com/subsurface/subsurface/commit/bcda8ac12a5d6808d89c37e6a419b28763eb0bd0 returns the correct 'penalty' values - so it looks like the plan segments need to be copied (or reset) multiple times throughout the calculations.

bstoeger:
Yes, #4313 also lost a few of the copies.

core/planner.cpp
bstoeger: You are right.

But the whole thing looks sketchy to me. Let's try something different.

mikeller (approved):
This gives the expected results now, thanks!

PR #4370, opened 2024-11-15
Round MOD to lower depths

### LINKED ISSUE #4365: MOD calculation calculated incorrectly in planner.
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

In the dive planner, under the Available Gases The calculation of MOD (maximum operating depth) for _Deco Switch at_  is rounding up, creating in some instances warnings about high ppO2 on the planner result.

### Operating system:

Arch Linux 
### Subsurface version:

Subsurface 6.0.5214-CICD-release
### Steps to reproduce:

Go to the dive planner, under the available gases pane, insert a Tank and assign a mix of 67% Oxigen 0% Helium
Set the Deco P02 to 1.6 under the Gas Options pane.

### Current behavior:

 The planner automatically assigns a Deco Switch depth of 15m.
Calculating the depth manually,  [(1.6 / 0.67 )-1]*10 = 13.88m you would expect a deco switch depth of 14m  or even 13m if we round towards safety, but definitely  not 15 this throws a high ppo2 warning on the Dive plan Details. 

### Expected behavior:

Correct calculation of MOD and rounding down towards safety. 

### Additional information:

ExampleSubsurface

atdotde:
When computing the suggested switch depth for a gas, we should take the next stop depth above the MOD, i.e. round down. Otherwise we can produce MOD violation warnings.

We need, however, a bit of fudge as otherwise we do not suggest to switch to o2 at 6m.

Fixes #4365

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/dive.cpp
+++ b/core/dive.cpp
@@ -2457,7 +2457,9 @@ int dive::mbar_to_depth(int mbar) const
 depth_t dive::gas_mod(struct gasmix mix, pressure_t po2_limit, int roundto) const
 {
 	double depth = (double) mbar_to_depth(po2_limit.mbar * 1000 / get_o2(mix));
-	return depth_t { .mm = int_cast<int>(depth / roundto) * roundto };
+	// Rounding should be towards lower=safer depths but we give a bit
+	// of fudge to all to switch to o2 at 6m. So from .9 we round up.
+	return depth_t { .mm = (int)(depth / roundto + 0.1) * roundto };
 }
 
 /* Maximum narcotic depth rounded to multiples of roundto mm */
--- a/tests/testplan.cpp
+++ b/tests/testplan.cpp
@@ -865,17 +865,17 @@ void TestPlan::testVpmbMetricRepeat()
 
 	// check minimum gas result
 	dp = std::find_if(testPlan.dp.begin(), testPlan.dp.end(), [](auto &dp) { return dp.minimum_gas.mbar != 0; });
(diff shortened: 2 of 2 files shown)


mikeller (commented):
Definitely fixes the bug in #4365. But it looks like the planner tests need a change to their test data now.

Out of interest, what is the reason that we are not generating plans that switch to oxygen at 6m? This seems to be the recommended best practice in most technical diving instruction.

atdotde:
Without the fudge factor, the MOD for o2 ends up being 5,809m. The main factor here is that the surface pressure is actually more than 1000 mbar. So we need the fudge factor to consider safe what everybody is actually doing.

Will look into the tests.

mikeller (approved):
Nice, this is giving switch depths that are identical to what dive computers and other dive planners are giving, without the incorrect warnings (6m for 100%, 21m for 50%, for a decompression ppO2 of 1.6 bar).

hhassey (commented):
@atdotde please beware that there is a switch in the planning pane that changes the Deco switch depth to happen where the  selected O2 partial pressure dictates (if not checked) or otherwise at the next shallower Deco Stop. (if checked)

image

PR #4371, opened 2024-11-15
planner: fix calculations of variations [alternative to #4369]

### LINKED ISSUE #4368: DIve Planner: Stop times per additional metre and minute always zero

bstoeger:
In 8704a8b the code in cloneDiveplan() was replaced by a simple assignement statement.

Alas, the original code was more complex: it copied only up to a certain point (it stopped at automatically generated steps).

The new behavior made the calculations of variations fail, because a call to plan() adds deco stops.

Therefore, copy the plan _before_ calling plan().

Fixes #4368

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Alternative version of #4369. I really don't understand what that `lock_planner()` stuff is about. This all looks very fishy to me.

Please comment.

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -1118,17 +1118,20 @@ void DivePlannerPointsModel::updateDiveProfile()
 	if (diveplan.is_empty())
 		return;
 
+	// For calculating variations, we need a copy of the plan. We have to copy _before_
+	// calling plan(), because that adds deco stops.
+	bool computeVariations = isPlanner() && shouldComputeVariations();
+	std::unique_ptr<struct diveplan> plan_copy;
+	if (computeVariations)
+		plan_copy = std::make_unique<struct diveplan>(diveplan);
+
 	deco_state_cache cache;
 	struct deco_state plan_deco_state;
 
 	plan(&plan_deco_state, diveplan, d, dcNr, decotimestep, cache, isPlanner(), false);
 	updateMaxDepth();
 
-	if (isPlanner() && shouldComputeVariations()) {
-		auto plan_copy = std::make_unique<struct diveplan>();
-		lock_planner();
-		*plan_copy = diveplan;
-		unlock_planner();
+	if (computeVariations) {
 #ifdef VARIATIONS_IN_BACKGROUND
 		// Since we're calling computeVariations asynchronously and plan_deco_state is allocated
 		// on the stack, it must be copied and freed by the worker-thread.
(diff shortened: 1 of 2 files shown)


qt-models/diveplannermodel.cpp
```diff
@@ -1118,17 +1118,20 @@ void DivePlannerPointsModel::updateDiveProfile()
 	if (diveplan.is_empty())
 		return;
 
+	// For calculating variations, we need a copy of the plan. We have to copy _before_
+	// calling plan(), because that adds deco stops.
+	bool computeVariations = isPlanner() && shouldComputeVariations();
+	auto plan_copy = computeVariations ?
+		std::make_unique<struct diveplan>(diveplan) :
+		std::make_unique<struct diveplan>();
```
mikeller: Should this just be `nullptr` if `computeVariations` is `false` - we are not using it in this case.
And the same thing below.

mikeller (approved):
The calculations seem to be correct with this one.

qt-models/diveplannermodel.cpp
bstoeger: Yes, the default constructor is the same as `nullptr`. But I think then one couldn't use the `? :` operator. I'll make this more readable.

mikeller:
I prefer this to #4369.

mikeller (approved):
Nice!

bstoeger:
Added a commit that replaces `std::unique_ptr` by value semantics.

mikeller (approved):
Nice!

bstoeger:
The last copy in `computeDiveVariations` should probably either use `std:move` or the `plan()` function should work on `original_plan`. But I can't be bothered. Copying the plan is certainly not time critical compared to the rest of the planner code. 🤷

PR #4379, opened 2024-11-20
Bugfix: Fix planner debug code.

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix the debug code in `core/planner.cpp`.

### DIFF
--- a/core/planner.cpp
+++ b/core/planner.cpp
@@ -27,6 +27,8 @@
 #include "qthelper.h"
 #include "version.h"
 
+#define DEBUG_PLAN 255
+
 static constexpr int base_timestep = 2; // seconds
 
 static std::vector<depth_t> decostoplevels_metric = { 0_m, 3_m, 6_m, 9_m, 12_m, 15_m, 18_m, 21_m, 24_m, 27_m,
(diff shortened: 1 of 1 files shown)


bstoeger:
Thanks.

I wonder if it would be a good idea to replace
```
#if DEBUG_PLAN & 16
```
By
```
static constexpr int debug_plan = 0;
...
        if constexpr (debug_plan & 16 != 0) {
```

That way it would always be compiled but no code would be emitted.

Alternatively, just remove the conditional debug code - personally, I always found it pointless. If I want to debug it, I just add debugging output where I need it. This stuff hasn't helped me once.

mikeller:
Good point, and a good suggestion for how to mitigate the risk of rot.

I don't know (yet) about the usefulness of the debug code, I was stuck trying to fix a gas consumption graph issue and tried enabling the debug code for this, and this is when I found out that it had rotted, and decided I didn't want this cleanup in what already is a complex change.

bstoeger (approved):
Just for records: From my side, feel free to commit.

mikeller:
@bstoeger: Still mulling over your feedback - it will definitely be good to protect this from rot, but at the same time the `#ifdef DEBUG...` pattern is well established and recognised by developers (and IDEs), so I think there's value in sticking with it.
Then I started thinking about enabling planner debugging for debug builds, only to discover that we are currently releasing some debug builds. Which then led me down rabbit hole #4380 ...

core/planner.cpp
```diff
@@ -45,23 +47,19 @@ static int decostoplevels_imperial[] = { 0, 3048, 6096, 9144, 12192, 15240, 1828
 #if DEBUG_PLAN
 void dump_plan(struct diveplan *diveplan)
```
bstoeger: I suggest changing the argument to `const struct diveplan &diveplan)` because
1) I don't see why it would make sense to allow a NULL ptr.
2) The `dump()` function hopefully doesn't modify the plan, does it?

PR #4385, opened 2024-11-26
Redirect logs to the application log file again

jefdriesen:
In commit 4de109bbf10ceb6858ba613b386e972ae871db91, qDebug() was replaced with report_info(). As a side effect, the subsurface log messages are no longer redirected to the application log file. Only the internal Qt logging ends up in the log file.

For bug reporting by end-users, this application log file is important because it's the easiest way to collect all the necessary information to investigate the problem.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


bstoeger:
Cool, thanks.

PR #4387, opened 2024-11-30
Avoid UTF8->UTF16->UTF8 roundtrip when logging on mobile

bstoeger:
Make writeToAppLogFile() take an std::string parameter to avoid an unnecessary UTF8->UTF16->UTF8 (std::string->QString->char *) roundtrip.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Tiny code cleanup that avoids unnecessary UTF8->UTF16->UTF8 conversions.

### Related issues:

Follow up to #4385.

### DIFF
--- a/mobile-widgets/qmlmanager.cpp
+++ b/mobile-widgets/qmlmanager.cpp
@@ -1849,20 +1849,20 @@ void QMLManager::setBtEnabled(bool value)
 
 #if defined(Q_OS_ANDROID) || defined(Q_OS_IOS)
 
-void writeToAppLogFile(QString logText)
+void writeToAppLogFile(const std::string &logText)
 {
 	// write to storage and flush so that the data doesn't get lost
-	logText.append("\n");
 	QMLManager *self = QMLManager::instance();
 	if (self) {
 		self->writeToAppLogFile(logText);
 	}
 }
 
-void QMLManager::writeToAppLogFile(QString logText)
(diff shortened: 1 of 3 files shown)


bstoeger:
@jefdriesen: I would appreciate your opinion, since I don't use mobile. Thanks.

mikeller (approved):

mobile-widgets/qmlmanager.cpp
```diff
@@ -1849,20 +1849,20 @@ void QMLManager::setBtEnabled(bool value)
...
 	QMLManager *self = QMLManager::instance();
 	if (self) {
 		self->writeToAppLogFile(logText);
 	}
 }
 
-void QMLManager::writeToAppLogFile(QString logText)
+void QMLManager::writeToAppLogFile(const std::string &logText)
 {
+	std::string line = logText + "\n";
```
jefdriesen: The appending of the new line could be moved inside the if statement, such that you only have to copy the string when it's actually written to the log.

jefdriesen:
Looks good to me. The fewer conversions (and copies) the better.

mobile-widgets/qmlmanager.cpp
bstoeger: Good point.

Issue #3979, opened 2023-10-29
Support for Android 14


D-Vanditmars:
### Describe the issue:

- [ ] Bug
- [ ] Change request
- [x] New feature request
- [ ] Discussion request

### Issue long description:

Support for Android 14

### Operating system:

Android 14

Pixel 5

Not sure if Android 13 was supported

### Subsurface version:

Latest one as of Oct 20, 2013 from Google play store, (app does not work with this version of OS)

Do not know

Official release

N/A

### Current behavior:

Android 14 Google Play store does not app to be installed

### Expected behavior:

The would install

matasbbb:
Request to bump target version to API 35

dirkhh:
Please check https://subsurface-divelog.org/current-release and install the APK provided there. 

The Google Play store has become an impossible target for an all volunteer open source project, so we are providing our app for sideloading instead.

matasbbb:
Not sure if it is due to specific accounts added, APK cannot be sideloaded due target SDK being 30. For Android 15, at least on my device it complains that target SDK should be updated.

dirkhh:
Hmm. I'm running the apk on a Pixel Fold with Android 15. So... not sure what to tell you.
We can certainly look into bumping the API target version.
As with everything else in the Android ecosystem, all of this is designed to create maximum pain and frustration for the developers who aren't using the Google provided tools and Google approved libraries...

D-Vanditmars:
OK, thanks for the explainations, I was using the old Google Play app and not side loading the APK...

PR #4391, opened 2024-12-05
Android: bump target SDK Version to 35

dirkhh:
The treadmill never stops.

- [x] Build system change

I doubt it will be this easy. It never is. But let's give it a try.

### DIFF
--- a/android-mobile/AndroidManifest.xml
+++ b/android-mobile/AndroidManifest.xml
@@ -18,6 +18,7 @@
             android:label="@string/app_name"
             android:theme="@style/AppTheme"
             android:launchMode="singleTop"
+            android:exported="true"
             android:windowSoftInputMode="adjustResize"
             android:screenOrientation="unspecified" >
             <intent-filter>
@@ -88,8 +89,8 @@
         </provider>
     </application>
 
-    <uses-sdk android:minSdkVersion="21"
-	      android:targetSdkVersion="30" />
+    <uses-sdk android:minSdkVersion="26"
+	      android:targetSdkVersion="34" />
 
     <supports-screens
         android:anyDensity="true"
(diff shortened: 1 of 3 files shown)


mikeller:
I saw https://github.com/subsurface/subsurface/issues/3979#issuecomment-2520555641 - but I am not 100% convinced that the SDK warning isn't just a red herring, and this doesn't work because some form of device lockdown.

@matasbbb can you please test with the android artifact in this pull request to confirm that this can be installed on your phone.

dirkhh:
> I saw [#3979 (comment)](https://github.com/subsurface/subsurface/issues/3979#issuecomment-2520555641) - but I am not 100% convinced that the SDK warning isn't just a red herring, and this doesn't work because some form of device lockdown.
> 
> @matasbbb can you please test with the android artifact in this pull request to confirm that this can be installed on your phone.

@mikeller I can reproduce it here. You need to uninstall the existing app and try to install it fresh in order to see it. Depending on "some stuff" (don't ask me) it appears that updates still work, it's just new installs that fail.

I'm now on round four of trying to track down what it takes to get this sorted. I absolutely love how well Google treats its developers. 

Oh, btw - if you are on Android 15 you now get a super obnoxious popup when side-loading an app that requires you to find a hidden option in order to actually install the app... and here I thought abandoning the store would protect us from most of Google's shenanigans. Not so much, it seems.

dirkhh:
Hmm. now I'm even more confused. On a different phone on Android 15, I CAN side load the "current" build of Subsurface. I get a different warning about lacking privacy, etc, but after finding the hidden option it does appear to install successfully and work.
This is all not great, as far as I'm concerned.

mikeller:
@dirkhh: I think this will become more and more valuable / urgent over time - do you have motivation / capacity to have another stab at it?

dirkhh:
Not for several weeks.

dirkhh:
this will be replaced by the Qt 6 port

PR #4393, opened 2024-12-06
macOS: build against Qt 5.15.16

dirkhh:
With this we should be able to switch to a newer mac build VM.

- [x] Build system change

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   build:
-    runs-on: macOS-12
+    runs-on: macOS-13
     steps:
     - name: checkout sources
       uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


dirkhh:
Ahh - looks like QtLinquist is missing from my Qt 5.15.16 build.
So I'll need to update that, again.

dirkhh:
Woot - macOS builds -- I tested things and the binary seems to work. Since this seems like such a clear step forward (and obviously cannot impact anything else - see the diff) I'll merge this myself :)

mikeller:
Thanks @dirkhh. 
The solution looks nice and clean.

mikeller:
Oh wait - only now did I realise that you had to build and push Qt 5.15.16 into our `qt-mac` repository.
As a thought, when we are upgrading Qt for Mac / iOS, should we delete the old version? (Or delete the -2 version, or delete with a grace period, so that work in progress that is building againt a previous version isn't broken)?
The speedup for our CICD runs will likely be minimal, but in terms of GitHub resources it seems to be the right thing to do.

dirkhh:
Yes, we should delete both of the old ones and force push to actually remove them from the repo as well. 
Since builds against the old one will fail, that seems fine. 

Otherwise (without the hard break thanks to GitHub) the "keep the last two" approach is of course better

PR #4394, opened 2024-12-08
Create UUID when adding a dive site to core

bstoeger:
3f8b4604be introduced a bug: it replaced alloc_dive_site() by the divesite constructor. However, the latter does not generate a UUID (for that it would have to access the dive site table).

Thus newly added dive sites had no UUID and could not be saved properly.

Use the register_site() instead of the put() function when adding the dive site to the core, so that the UUID is created if needed.

This needs an audit.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Bug fix for creation of dive sites.

### Changes made:

1) Create UUID when adding UUID-less dive site to core

### DIFF
--- a/commands/command_divesite.cpp
+++ b/commands/command_divesite.cpp
@@ -30,7 +30,7 @@ static std::vector<dive_site *> addDiveSites(std::vector<std::unique_ptr<dive_si
 		}
 
 		// Add dive site to core, but remember a non-owning pointer first.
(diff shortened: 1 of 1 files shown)


dirkhh (approved):
reading through the code in `register_site()`, that does look like it would be a safe replacement for simply calling `put()`.

PR #4392, opened 2024-12-06
Refactored and enhanced cylinder depth and duration handling

FMain89:
- Implemented modular functions for per-cylinder depth and duration calculations.
- Added validation checks for gas changes and cylinder mappings.
- Improved handling of dives with missing cylinders or undefined sensor data.
- Optimized gasmix loop logic for dives with multiple gas switches.
- Updated comments for better clarity and documentation.

Reported-by: Freddie Main Issue #4375
Signed-off-by: Freddie Main mainf@oregonstate.edu

### Describe the pull request:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [X] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

This pull request adds comments and improves the readability of the dive.cpp file. Additionally, it includes code refactoring for better maintainability while adhering to the project's coding standards.

### Changes made:

1. Added detailed comments throughout the dive.cpp file to improve readability and code clarity.
2. Modularized repetitive logic into helper functions (e.g., initialize_cylinder_data, handle_unknown_cylinders, calculate_depth_and_duration).
3. Enhanced adherence to the coding style by breaking down lengthy functions into smaller, focused units.
4. Improved inline documentation to make function purposes and parameters clearer.

### Related issues:

Fixes #4375 

### Release note:

No functional changes are introduced; hence, a release note is not required.

### Documentation change:

No changes to user-facing documentation are required. This is an internal code improvement.

PR #4397, opened 2024-12-13
Let depth_to_...() take a depth_t arguments

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Inspired by #4392, make the `depth_to_...()` functions accept a `depht_t`. The commit then became more invasive as I hoped for. Once we define comparison operators and multiplication of scalars, this should become much more palatable. But at least for the comparisons I'd wait for C++20's spaceship operator.
 
### Changes made:

1) Use more `depht_t`.
2) Make the user-defined unit literals constexpr.

### Related issues:

Inspired by #4392.


mikeller (approved):

bstoeger:
@FMain89: In the last commit I refactor the function of your PR #4392.

Sorry for crossing work, but in this case it was easier to just do it myself instead of explaining how this interface could be simplified. And yes, this kind of refactoring requires some intimate knowledge of the (very inconsistent) code base.

bstoeger:
Last update to this PR, since I'm starting to royally shoot myself into the foot (with respect to the Profile-to-QtQuick-PR).

core/planner.cpp
```diff
@@ -29,18 +29,19 @@
...
-					182880, 193040, 203200, 223520, 243840, 264160, 284480, 304800,
-					325120, 345440, 365760, 386080 };
+static std::vector<depth_t> decostoplevels_metric = { 0_m, 3_m, 6_m, 9_m, 12_m, 15_m, 18_m, 21_m, 24_m, 27_m,
+					30_m, 33_m, 36_m, 39_m, 42_m, 45_m, 48_m, 51_m, 54_m, 57_m,
+					60_m, 63_m, 66_m, 69_m, 72_m, 75_m, 78_m, 81_m, 84_m, 87_m,
+					90_m, 100_m, 110_m, 120_m, 130_m, 140_m, 150_m, 160_m, 170_m,
+					180_m, 190_m, 200_m, 220_m, 240_m, 260_m, 280_m, 300_m,
+					320_m, 340_m, 360_m, 380_m };
+// Note: for fractional feet (e.g. 333.33 ft), we use mm, since currently our custom depth literals only support integers.
+static std::vector<depth_t> decostoplevels_imperial = { 0_ft, 10_ft, 20_ft, 30_ft, 40_ft, 50_ft, 60_ft, 70_ft, 80_ft, 90_ft,
```
mikeller, line 39: Frankly, not sure if we want to keep these in fractions, or if we should just truncate to 333 ft, 666 ft and so on - these are mostly theoretical depths anyway as there's only a handful of people on this planet diving to depths that will require stops at these depths.

core/planner.cpp
```diff
@@ -656,22 +649,17 @@ std::vector<decostop> plan(struct deco_state *ds, struct diveplan &diveplan, str
...
 	/* If the user has selected last stop to be at 6m/20', we need to get rid of the 3m/10' stop.
 	 * Otherwise reinstate the last stop 3m/10' stop.
 	 * Remark: not reentrant, but the user probably won't change preferences while this is running.
 	 */
 	if (prefs.last_stop)
-		*(decostoplevels + 1) = 0;
+		decostoplevels[1] = 0_m;
 	else
-		*(decostoplevels + 1) = M_OR_FT(3,10);
+		decostoplevels[1] = m_or_ft(3,10);
```
mikeller: Should this have a space after the comma?

core/profile.cpp
```diff
@@ -123,7 +122,7 @@ static int get_local_sac(struct plot_info &pi, int idx1, int idx2, struct dive *
 		return 0;
 
 	/* Mean pressure in ATM */
-	depth = (entry1.depth + entry2.depth) / 2;
+	depth_t depth = entry1.depth + entry2.depth / 2;
```
mikeller: This may work out to be correct based on how operator overloading impacts operator precedence, but I think retaining the braces will make this easier to understand.

core/units.h
```diff
@@ -86,6 +86,20 @@
  *	_permille	-> fraction_t in ‰
  *	_percent	-> fraction_t in %
  *
+ * For now, addition and subtraction of values of the same type
+ * are supported.
+ *
+ * Moreover, multiplication and division with an integral are
+ * suppoerted. Attention: the latter uses standard C++ integer
```
mikeller:
```suggestion
 * supported. Attention: the latter uses standard C++ integer
```

core/planner.cpp
bstoeger: Yes, I wanted to avoid any change in the calculations. But the tests only test the metric table and I hope nobody depends on 10 cm depth difference. Added an additional commit that rounds to integral feet.

core/planner.cpp
bstoeger: Fixed here and elsewhere.

core/profile.cpp
bstoeger: Ouch! Operator overloading can thankfully never change operator precedence. Fixed.

core/units.h
bstoeger: Fixed.

mikeller (approved):

PR #4415, opened 2025-01-10
fix std:clamp usage

dirkhh:
We appear to consistently assume that clamp can be called with 0, size as interval, but that actually results in a possible out of bounds access at the upper end.

- [x] Bug fix


bstoeger (approved):
LGTM

mikeller (approved):

PR #4419, opened 2025-01-15
Import: Fix the Handling of ppO2 Sensor IDs.

mikeller:

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Instead of assigning sensor ids in the order that values are reported,
actually use the sensor ids reported by libdivecomputer. This will fix
the problem that for some dive computers (e.g. Shearwater) the dive
computer calculated ppO2 is currently reported first, thus pushing out
all actual sensor values.
A new fixed id (7) outside of the range of currently supported sensor
IDs is used for these dive computer calculated values.

### Changes made:

1. use the actual sensor ID when reading ppO2 sensor values from the dive computer;
2. use the new position 7 in the sensor array when storing a `DC_SENSOR_NONE` value;
3. persist and read the `DC_SENSOR_NONE` value as 'sensor7' when persisting and reading from XML and git.

### Related issues:

https://groups.google.com/g/subsurface-divelog/c/wTcSp4gvWJY/m/mr19MVk9CgAJ

### Additional information:

Limitiations:
- 'sensor7' is not handled in XSLT transformations
- `DC_SENSOR_NONE` values are not graphed in the profile

For discussion:
I am not sure if exporting / importing this as 'sensor7' is the correct approach, as this:
- makes it appear as if this were a sensor value;
- will create problems if we ever wanted to increase the number of supported sensors.

Possible alternatives (just for import / export, retaining the in-memory representation as it is):
- 'sensor99'
- 'dc_ppo2_value'

@dirkhh thoughts?

### DIFF
--- a/core/libdivecomputer.cpp
+++ b/core/libdivecomputer.cpp
@@ -379,7 +379,6 @@ static void handle_gasmix(struct divecomputer *dc, const struct sample &sample, 
 void
 sample_cb(dc_sample_type_t type, const dc_sample_value_t *pvalue, void *userdata)
 {
-	static unsigned int nsensor = 0;
 	struct divecomputer *dc = (divecomputer *)userdata;
 	dc_sample_value_t value = *pvalue;
 
@@ -388,8 +387,6 @@ sample_cb(dc_sample_type_t type, const dc_sample_value_t *pvalue, void *userdata
 	 * Other types fill in an existing sample.
 	 */
 	if (type == DC_SAMPLE_TIME) {
-		nsensor = 0;
-
 		// Create a new sample.
 		// Mark depth as negative
 		struct sample *sample = prepare_sample(dc);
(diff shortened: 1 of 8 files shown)


dirkhh:
I think we should special case this and not have it as "a fake sensor". So `dc_sensor_none` or something along those lines

mikeller:
> I think we should special case this and not have it as "a fake sensor". So `dc_sensor_none` or something along those lines

`dc_supplied_ppo2`? `dc_sensor_none` seems to be the antipattern of naming things for what they _aren't_.

dirkhh:
> `dc_supplied_ppo2`? `dc_sensor_none` seems to be the antipattern of naming things for what they _aren't_.

That works for me. @atdotde ?

atdotde:
Well, since you asked (I don't think I can add much of value to this discussion): Special values tend to bite back later. Think of our convention of having po2 == 0 signalling AIR rather than "you will die immediately when you breathe this" which we only mitigated by using helper functions get_o2 and never use the bare value. And similar the case of the extra cylinder with special number that I introduced to handle the breathing of ambient air during surface intervals (which to me seemed to be an easy solution for extending the heat map into after the dive but for which I received a lot of scolding from Berthold later on when he introduced the gas index and the corresponding table).

mikeller:
Tested persisting / loading in XML and git, both seem to be working fine:

image

PR #4420, opened 2025-01-16
Desktop: Close Warning when Closing Logbook.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Close the warning / error notification when the logbook is closed.

These notifications lingering around long after the condition that
caused them has been addressed or has become irrelevant is annoying.
This could probably be done in a better way by 'remembering' when a
notification is posted, and closing it when the context changes and it
has not been superseded.
But at least when closing the logbook we know that whatever the
notification has been about has become stale.

@bstoeger: What would be a good way in C++ to create a unique 'handle' for the notification to be able to close it if and only if it is still showing?

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -509,6 +509,8 @@ bool MainWindow::okToClose(QString message)
 
 void MainWindow::closeCurrentFile()
 {
+	ui.mainErrorMessage->hideNotification();
+
 	/* free the dives and trips */
 	clear_git_id();
 	clear_dive_file_data(); // this clears all the core data structures and resets the models


bstoeger:
> What would be a good way in C++ to create a unique 'handle' for the notification to be able to close it if and only if it is still showing?

I'm not exactly sure what you're asking. What you did seems fine to me.

And yes, the whole notification UI is somewhat idiosyncratic.

mikeller:
@bstoeger: I think the case that I added in this pull request is possibly the only case where we want to close the notification irrespective of what the currently showing contents are.
In other cases, like when closing the import dialogue, we only want to close the notification if the currently showing message was generated by something that happened in the dialogue, but not for example if the warning comes from the dive list update. So it would be nice if we could get back something like a 'handle' when showing a notification, and then use this to close the notification _only if this is the one currently showing_.
Or maybe I am overthinking this...

PR #4423, opened 2025-01-22
Desktop: Show Warning for Dives with No Viable Cylinders.

### LINKED ISSUE #4413: commit 07898f2 breaks correct display of ceilings for CCR dives
After my last upgrade I was surprised that all my recent CCR dives show significant violations of the decompression ceilings.
I used "git bisect" to track the issue down. The first bad commit is 07898f2.
Seems to affect CCR dives only.
Compare the displayed decompression ceilings (GF40/80) for the attached example dive with subsurface before and after the bad commit.

[example.ssrf.gz](https://github.com/user-attachments/files/18307605/example.ssrf.gz)

### Describe the issue:

- [x ] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Show a warning when a dive is opened / edited and doesn't have any
usable cylinders for the selected dive mode (i.e. no open circuit gas
for open circuit / PSCR, or no diluent for CCR).

### Related issues:

Fixes #4413.

### Release note:

image

### Mentions:

Reported-by: @kruegerha

### DIFF
--- a/core/dive.cpp
+++ b/core/dive.cpp
@@ -1042,6 +1042,27 @@ static void fixup_dc_sample_sensors(struct dive &dive, struct divecomputer &dc)
 	}
 }
 
+static void fixup_dc_cylinder_use(struct dive &dive, struct divecomputer &dc)
+{
+
+	if (dc.divemode != OC && dc.divemode != CCR && dc.divemode != PSCR)
+		return;
+
+	// Check that we have at least one cylinder that is suitable for the dive
+
+	// For OC we default to air if we don't have any cylinders
+
+	if (dc.divemode == OC && dive.cylinders.empty())
+		return;
+
+	for (auto &cylinder: dive.cylinders)
+		if ((dc.divemode == CCR && cylinder.cylinder_use == DILUENT) || ((dc.divemode == PSCR || dc.divemode == OC) && cylinder.cylinder_use == OC_GAS))
+			return;
+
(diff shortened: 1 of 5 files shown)


mikeller:
After finding that several older dives in my log revealed this problem, I crafted some XSLT to automatically fix this for all trivial cases of a CCR dive with only a single cylinder, which in this case has to be the diluent:

```
<xsl:stylesheet version="3.1" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
  <xsl:output method="xml" indent="yes" encoding="UTF-8" omit-xml-declaration="no"/>
  <xsl:mode on-no-match="shallow-copy"/>

  <xsl:template match="/divelog/dives/dive/cylinder[../divecomputer[@dctype = 'CCR'] and count(../cylinder) = 1 and not(@use = 'diluent')]">
    <xsl:copy>
      <xsl:apply-templates select="@*"/>
      <xsl:attribute name="use" select="'diluent'"/>
      <xsl:apply-templates select="node()"/>
    </xsl:copy>
  </xsl:template>

</xsl:stylesheet>
```

I developed and used this with https://www.npmjs.com/package/xslt3

PR #4424, opened 2025-01-22
Desktop: Fix Crash when Adding JPG Image on MacOS.

### LINKED ISSUE #4377: MacOS App crashes when trying to import media files (.jpg images)
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

App crashes when trying to import media files into a dive

### Operating system:

macOS, Ventura 13.6.7

iMac 2017

### Subsurface version:

6.0.5276-CICD-release

Didn't test.

Official release

### Steps to reproduce:

1) open cloud storage
2) select dive
3) select "media" tab
4) right click in media tab and load media from file(s)
5) select 1 or more .jpg file
6) manually shift time of image, so it fits within the dive time
7) hit "OK" to start import

### Current behavior:

app crashes immediately after hitting "OK"

### Expected behavior:

app will not crash, but import the media files instead

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix a crash on MacOS when trying to open a JPG image with libraw by filtering files by valid extensions for RAW files.

### Related issues:

Fixes #4377.

### Mentions:

@atdotde, @chaehni can you please test this?

### DIFF
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -74,9 +74,9 @@ void ImageDownloader::saveImage(QNetworkReply *reply)
 	reply->deleteLater();
 }
 
-static bool hasVideoFileExtension(const QString &filename)
+static bool hasFileExtension(const QString &filename, const QStringList extensionsList)
 {
-	for (const QString &ext: videoExtensionsList)
+	for (const QString &ext: extensionsList)
 		if (filename.endsWith(ext, Qt::CaseInsensitive))
 			return true;
 	return false;
@@ -136,7 +136,7 @@ Thumbnailer::Thumbnail Thumbnailer::fetchImage(const QString &urlfilename, const
 
 #ifdef LIBRAW_SUPPORT
 		// If note, perhaps a raw image?
-		if (thumb.isNull())
+		if (thumb.isNull() && hasFileExtension(filename, rawExtensionsList))
 			thumb = fetchRawThumbnail(filename);
 #endif
 		if (!thumb.isNull()) {
@@ -148,7 +148,7 @@ Thumbnailer::Thumbnail Thumbnailer::fetchImage(const QString &urlfilename, const
 		// Neither our code, nor Qt could determine the type of this object from looking at the data.
 		// Try to check for a video-file extension. Since we couldn't parse the video file,
 		// we pass 0 as the duration.
-		if (hasVideoFileExtension(filename))
+		if (hasFileExtension(filename, videoExtensionsList))
 			return fetchVideoThumbnail(filename, originalFilename, duration_t());
 
 		// Give up: we simply couldn't determine what this thing is.
--- a/core/imagedownloader.cpp
+++ b/core/imagedownloader.cpp
@@ -74,14 +74,6 @@ void ImageDownloader::saveImage(QNetworkReply *reply)
 	reply->deleteLater();
 }
 
-static bool hasFileExtension(const QString &filename, const QStringList extensionsList)
-{
-	for (const QString &ext: extensionsList)
-		if (filename.endsWith(ext, Qt::CaseInsensitive))
-			return true;
-	return false;
-}
-
 #ifdef LIBRAW_SUPPORT
 QImage fetchRawThumbnail(const QString &filename)
 {
--- a/core/metadata.cpp
+++ b/core/metadata.cpp
@@ -571,7 +571,7 @@ mediatype_t get_metadata(const char *filename_in, metadata *data)
 	data->location.lon.udeg = 0;
 
 #ifdef LIBRAW_SUPPORT
-	if (parseRaw(filename_in, data))
+	if (hasFileExtension(filename_in, rawExtensionsList) && parseRaw(filename_in, data))
 		return MEDIATYPE_PICTURE;
 #endif
 
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -1094,13 +1094,21 @@ const QStringList videoExtensionsList = {
 // Raw extensions according to https://en.wikipedia.org/wiki/Raw_image_format
 const QStringList rawExtensionsList = {
 #ifdef LIBRAW_SUPPORT
-	"*.3fr", "*.ari", "*.arw", "*.bay", "*.braw", "*.crw", "*.cr2", "*.cr3", "*.cap",
-	"*.data", "*.dcs", "*.dcr", "*.dng", "*.drf", "*.eip", "*.erf", "*.fff", "*.gpr",
-	"*.iiq", "*.k25", "*.kdc", "*.mdc", "*.mef", "*.mos", "*.mrw", "*.nef", "*.nrw",
-	"*.obm", "*.orf", "*.pef", "*.ptx", "*.pxn", "*.r3d", "*.raf", "*.raw", "*.rwl",
-	"*.rw2", "*.rwz", "*.sr2", "*.srf", "*.srw", "*.x3f"
+	".3fr", ".ari", ".arw", ".bay", ".braw", ".crw", ".cr2", ".cr3", ".cap",
+	".data", ".dcs", ".dcr", ".dng", ".drf", ".eip", ".erf", ".fff", ".gpr",
+	".iiq", ".k25", ".kdc", ".mdc", ".mef", ".mos", ".mrw", ".nef", ".nrw",
+	".obm", ".orf", ".pef", ".ptx", ".pxn", ".r3d", ".raf", ".raw", ".rwl",
+	".rw2", ".rwz", ".sr2", ".srf", ".srw", ".x3f"
 #endif
 };
+
+bool hasFileExtension(const QString &filename, const QStringList extensionsList)
+{
+	for (const QString &ext: extensionsList)
+		if (filename.endsWith(ext, Qt::CaseInsensitive))
+			return true;
+	return false;
+}
 
 QStringList mediaExtensionFilters()
 {
@@ -1112,7 +1120,9 @@ QStringList imageExtensionFilters()
 	QStringList filters;
 	for (QString format: QImageReader::supportedImageFormats())
 		filters.append("*." + format);
-	return filters + rawExtensionsList;
+	for (const QString &extension: rawExtensionsList)
+		filters.append("*" + extension);
+	return filters;
 }
 
 QStringList videoExtensionFilters()
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -1092,7 +1092,7 @@ const QStringList videoExtensionsList = {
 };
 
 // Raw extensions according to https://en.wikipedia.org/wiki/Raw_image_format
-static const QStringList rawExtensionsList = {
+const QStringList rawExtensionsList = {
 #ifdef LIBRAW_SUPPORT
 	"*.3fr", "*.ari", "*.arw", "*.bay", "*.braw", "*.crw", "*.cr2", "*.cr3", "*.cap",
 	"*.data", "*.dcs", "*.dcr", "*.dng", "*.drf", "*.eip", "*.erf", "*.fff", "*.gpr",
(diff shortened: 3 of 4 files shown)


atdotde:
I am not sure that this is actually a fix and not just hiding the problem: I would expect file extensions to be unreliable with respect to determining file type. Furthermore, I would expect (assuming the absence of a bug in libra) that the error handling  in l. 91 of imagedownloader.cpp would detect problems in parsing the file as raw and return an empty QImage.

At least, we should make sure that creating a thumbnail for actual raw files works. But I don't have such a file available.

mikeller:
@chaehni: Are you able to test with the [MacOS artifact](https://github.com/subsurface/subsurface/suites/33387957544/artifacts/2469907802) to see if this fixes the problem in #4377?

@atdotde:
Yes, I think you are correct, and this is caused by a bug in libraw.

chaehni:
@mikeller Thanks for working on this. Unfortunately, this build still crashes when loading JPEGs.

mikeller:
@chaehni Can you please try again with https://github.com/subsurface/subsurface/suites/33538227327/artifacts/2487749208?

chaehni:
@mikeller Patch 2 does not fix the problem, either.

mikeller:
> @mikeller Patch 2 does not fix the problem, either.

:frowning_face:  - but not much surprised. Thing is, now I am struggling to see how #3954 could have broken this.

Thanks for testing.

chaehni:
@mikeller If it helps, I could try to set up a dev environment such that I can run the debugger on the code and see where exactly it fails.

mikeller:
> @mikeller If it helps, I could try to set up a dev environment such that I can run the debugger on the code and see where exactly it fails.

@chaehni: Yes, that would help. Thank you.

In any case, can you please also try https://github.com/subsurface/subsurface/suites/33583311313/artifacts/2493326177?
This is essentially a last ditch attempt, building the MacOS artifact without libraw support, to test if it is caused by libraw, or some coincidence like a version update in the GitHub MacOS runners that happened exactly when #3954 went in.

chaehni:
@mikeller Patch 3 does not crash. So then it seems to be a problem with libraw or how it is called from subsurface.

mikeller:
> @mikeller Patch 3 does not crash. So then it seems to be a problem with libraw or how it is called from subsurface.

Now I am completely puzzled - for what I can see, patch.2 already results in no functions in libraw being called from our code for any files unless they have a RAW specific extension. So the only thing that I can see that would explain what we are seeing is if Qt detects the presence of libraw on its own, and uses a different code path if it is detected. :face_with_diagonal_mouth:

chaehni:
@mikeller That puzzled me, too. Just the absence of libraw with no code change fixes this. Definitely strange. I'll try and set up a debug env tomorrow.

mikeller:
@chaehni: In the mean time I've got another change to test: https://github.com/subsurface/subsurface/suites/33589189708/artifacts/2494066657 This 'only' fixes a bug in my extension filtering, so it _should_ not change the outcome, but at this stage I want to exclude as many factors as possible.

chaehni:
@mikeller Well, it does. This build works.

I'm really not sure how the code works exactly. Before you added the extension filter, did Subsurface infer from the header whether to use libraw or not? In that case I could give you a sample file. On the other hand, I tried jpegs from different source cameras and they all failed.

mikeller:
> @mikeller Well, it does. This build works.

Gah, I fixed it, but I don't know how. :laughing: 

> I'm really not sure how the code works exactly. Before you added the extension filter, did Subsurface infer from the header whether to use libraw or not? In that case I could give you a sample file. On the other hand, I tried jpegs from different source cameras and they all failed.

Yes, pretty much - we'd throw every image file at libraw, and expect libraw to figure out if this was a valid RAW image or not. The change that I am making in this pull request is to filter down the files based on extension, and only try to open files with known RAW file extensions in libraw.

The weird thing is that before my latest fix, filtering should have been broken in a way that it would not send any file to libraw, as it was looking for a literal `*` in the file name...

In general, I think that the extension based filtering is needed in this case, as some of the formats supported by libraw are pretty much a memory dump from a digital camera (hence the name RAW) - so it's possible that the MacOS / ARM version of the library ends up mis-identifying JPG images as one of these formats, and then goes off the rails when trying to unpack the data in it.

chaehni:
@mikeller That's indeed very odd. Neither of the two filters should have matched my .jpeg file.
Edit: Never mind. I just tried with a different image file and this build still crashes. Sorry for the wrong info before. I think I might have used an image that I had imported successfully before. In that case the metadata seems to be cached and re-used.

chaehni:
In the meantime I tried to get the build running on my system so I can debug. Following the install instructions, running `bash subsurface/scripts/build.sh -build-deps`  I get this error: `configure: error: one or more libs available at link-time are not available run-time. Libs used at link-time: -lnghttp2 -lidn2 -lssl -lcrypto -lssl -lcrypto -lzstd  -lbrotlidec  -lz`

mikeller:
@chaehni:

> In the meantime I tried to get the build running on my system so I can debug. Following the install instructions, running `bash subsurface/scripts/build.sh -build-deps` I get this error: `configure: error: one or more libs available at link-time are not available run-time. Libs used at link-time: -lnghttp2 -lidn2 -lssl -lcrypto -lssl -lcrypto -lzstd -lbrotlidec -lz`

I have only limited experience with building on MacOS - all I have is one Apple machine, and I set up the build environment on it some time ago and it's always worked since then. But to make sure, did you use the instructions in https://github.com/subsurface/subsurface/blob/master/INSTALL.md#building-subsurface-under-macosx?

chaehni:
@mikeller Yes, I followed those instructions. Maybe they're outdated or don't work for M1 Systems. Do you know anyone who could help?

dirkhh:
builds on current macOS versions with Qt5 work on Apple M systems, but of course run emulated.
builds with Qt6 support arm64 builds, but the build integration (i.e., identifying and copying missing libraries) is still buggy.
Go through the libraries that it reports as missing and manually copy them into the output directory.

In a quick test build right now I actually get these warnings, but the resulting build appears to work

atdotde:
I have now a build working on my silicone Mac that crashes even when running with -g compile option. I have started investigating this with a debugger but so for inconclusive. A bit short on time but I hope to make some progress the upcoming weekend.

chaehni:
@dirkhh I used QT5 for my build attempt. I had various libs that where not found, QT paths not detected, qmake not being found and some other problems. I install QT via Homebrew. Maybe that's not intended.

dirkhh:
You did run make install, right?
Actually, reading again ...
You may want to try our hand crafted Qt build instead of homebrew.
You can see the steps to get that and how to set it up in the GitHub actions ..

chaehni:
I ran `build.sh`. From what I can tell, it runs `make install`.

atdotde:
What I learned: This is a pain to debug. The crash goes away, if in imagedownloader.cpp line 485 we force the thumbnail fetching to be in the main thread rather than delegating it with QtConcurrent::run. This suggests that it comes from libraw getting confused when first calling it on a non-raw file and then, possibly from a  different thread calling it again (it crashes only the second time it comes around). I am a bit lost now what to do with this. Anybody any suggestions how to proceed?

atdotde:
Ha! Thread safety was the keyword! Turns out, normal libraw is not thread safe, but there is libraw_r which is. Using that instead fixes the crash. At least for me. See #4437

mikeller:
> Ha! Thread safety was the keyword! Turns out, normal libraw is not thread safe, but there is libraw_r which is. Using that instead fixes the crash. At least for me. See #4437

Nice, good finding!

LibRaw:
Dear Colleagues:

We was contacted via our site contact form on Jan 29 with links to this pull request and related issue. No additional information (LibRaw version, file sample) was provided on 1st contact. 
Today I've received the reply, in short, 'now we know about libraw_r'.

Most likely, libraw_r is NOT the thing you really need, possible problem cause is described below:

According to the stack trace found in issue comment: https://github.com/subsurface/subsurface/issues/4377#issuecomment-2604344634  the problem is 'stack overflow' (the conclusion is based on '___chkstk_darwin' at top of stack trace.

I've read your code, you use LibRaw as on stack object:  `LibRaw raw`;

As mentioned in LibRaw docs, LibRaw object is BIG (800kb or even more):  https://www.libraw.org/docs/API-notes.html#memory

Your case is specificaly mentioned in LibRaw docs (I know that no one reads the documentation, but if something goes wrong, then the documentation is the first place to look):

> Thus, the work of one LibRaw instance may require about 1MB of stack memory. This is not a problem for most contemporary architectures. However, **when working in a multithreaded environment**, one should not forget to allocate a sufficient amount of memory for the thread stack....

If I guessed the essence of the problem correctly, then the solution is dynamic (heap) allocation of LibRaw object via 'new LibRaw' or std::make_unique if one prefers std::unique_ptr to ease memory cleanup on return from your function(s).

Also: initial problem report is incorrect: There was a complaint about 'non RAW' files  (e.g. any JPEG), while the problem will raise on RAW files too (If I've guessed the problem source right)

Alex Tutubalin, LibRaw team

LibRaw:
Followup:  Increasing the thread stack size should help too without changing your code (+2MB will be enough), but it may not be very easy to do in the QtConcurrent environment you are using.  Switching to Qt's QThread (manual thread creation with setStackSize() call) will require more code changes than simply change from `LibRaw raw` to `std::unique_ptr<LibRaw>=...`

LibRaw:
And Followup #2:  yes, libraw_r is more correct in your case, but since you are using the library only to extract previews - there is no difference: Sony decrypted metadata may be incorrect but you don't need them (another non-thread-safety exists in several RAW decoders you do not use for now).

mikeller:
@LibRaw: Hi Alex, thank you for getting back to us!

> As mentioned in LibRaw docs, LibRaw object is BIG (800kb or even more): [libraw.org/docs/API-notes.html#memory](https://www.libraw.org/docs/API-notes.html#memory)
> 
> Your case is specificaly mentioned in LibRaw docs (I know that no one reads the documentation, but if something goes wrong, then the documentation is the first place to look):
> 
> > Thus, the work of one LibRaw instance may require about 1MB of stack memory. This is not a problem for most contemporary architectures. However, **when working in a multithreaded environment**, one should not forget to allocate a sufficient amount of memory for the thread stack....
> 
> If I guessed the essence of the problem correctly, then the solution is dynamic (heap) allocation of LibRaw object via 'new LibRaw' or std::make_unique if one prefers std::unique_ptr to ease memory cleanup on return from your function(s).

Thank you, this is very insightful.

> Also: initial problem report is incorrect: There was a complaint about 'non RAW' files (e.g. any JPEG), while the problem will raise on RAW files too (If I've guessed the problem source right)

In retrospect I think this was a bit of a red herring that was caused by the fact that there probably aren't many users actually adding RAW images directly - if the shots were taken under water then they will need to have the colour shift corrected, and most users will want to add the processed images.

> And Followup #2: yes, libraw_r is more correct in your case, but since you are using the library only to extract previews - there is no difference: Sony decrypted metadata may be incorrect but you don't need them (another non-thread-safety exists in several RAW decoders you do not use for now).

We do read _some_ metadata (coordinates), so `libraw_r` might be needed:: https://github.com/subsurface/subsurface/blob/2a1b44e8ec740ce309ad5be121dc83a6e6da54fc/core/metadata.cpp#L545-L562

LibRaw:
> so `libraw_r` might be needed::

libraw_r is definitely better: Maybe in the future you will start unpacking/rendering RAW and it will be surprising to see that it does not work for all formats.

But I'm almost completely sure that the cause of the original problem is a stack overflow, and not something else (and it doesn't even get to the execution of the library code, so the regular and thread-safe versions do not work the same way)

atdotde:
Thanks @LibRaw, I have updated PR #4437 accordingly.

PR #4427, opened 2025-01-23
profile-widget2: enable vertical line with infobox

kruegerha:
Enables the vertical mouseFollower in currentState==PROFILE, while prefs.infobox is active.

The values shown in the infobox correspond to a specific data sample. This sample is exactly pointed out by the vertical line of the MouseFollower, enabling a more intuitive analysis

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

When analyzing the dive profile in detail, the Infobox may provide the data of each logged sample.
However, to locate the exact position, the mouse pointer has to be positioned very close to the profile plot.
This change enables the mouseFollowerVertical, to help highlighting the exact position of the cursor for analysis.

### Changes made:

changes in profilewidget2.cpp ensure, that 
mouseFollowerVertical->setVisible(true);
is set, when prefs.infobox == true, and the currentState==PROFILE.

### DIFF
--- a/profile-widget/profilewidget2.cpp
+++ b/profile-widget/profilewidget2.cpp
@@ -255,6 +255,17 @@ void ProfileWidget2::divesChanged(const QVector<dive *> &dives, DiveField field)
 
 void ProfileWidget2::actionRequestedReplot(bool)
 {
+	/* this is called vai infoboxChanged, therefore in currentState==PROFILE
+	   we have to set the mouseFollowerVertical to visible (if prefs.infobox)    hk  */
+#ifndef SUBSURFACE_MOBILE
+	if (currentState == PROFILE) {
+		mouseFollowerHorizontal->setVisible(false);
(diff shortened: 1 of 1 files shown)


profile-widget/profilewidget2.cpp
```diff
@@ -255,6 +255,17 @@ void ProfileWidget2::divesChanged(const QVector<dive *> &dives, DiveField field)
 
 void ProfileWidget2::actionRequestedReplot(bool)
 {
+	/* this is called vai infoboxChanged, therefore in currentState==PROFILE
```
mikeller:
```suggestion
	/* this is called via infoboxChanged, therefore in currentState==PROFILE
```

Also, the 'hk' doesn't seem to be super useful

mikeller (commented):
Good idea!

PR #4428, opened 2025-01-23
profile: make the partial pressure plot resizeable

kruegerha:
This adds an option to change the size of the partial pressure plot, in relation to the (depth)profile plot above.
prefs.gasplot_frac was added, with an option to change it on the technical settings tab. Allowed values are 0.1 - 0.7.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

the plot of the partial pressures was way too small for any useful analysis of sensor differences/performance, which is of much interest in CCR diving. This patch adds a way to change the fractional height of the pO2 (and pN2 an pHe) plot.
The tec-setup page as an Gasplot Fraction box, to select a value between 0.1 and 0.7, according to which the size of the gas-plots is scaled.

### Changes made:
prefs.gasplot_fraction
minor changed to the ui
main changes are in ProfileScene::updateAxes

### Release note:

not sure if needed

### Documentation change:

if this is considered, the documentation should mention the new value in tec-setup


profile-widget/profilescene.cpp
```diff
@@ -339,30 +339,51 @@ void ProfileScene::updateAxes(bool diveHasHeartBeat, bool simplified)
...
-		{ gasYAxis, 75.0, 0.0, ppGraphsEnabled(currentdc, simplified) },
-		{ temperatureAxis, 50.0, 2.0, true },
+		// hk: using absolute numbers here causes troubles with small windows! all changed to fractions of height
+		{ heartBeatAxis, heartBeatFraction * height, 0.0, prefs.hrgraph && diveHasHeartBeat },
+		{ percentageAxis, percentagegraphFraction * height, 0.0, prefs.percentagegraph },
+		{ gasYAxis, prefs.gasplot_frac * height, 0.0, ppGraphsEnabled(currentdc, simplified) },
+		{ temperatureAxis, temperatureFraction * height, 2.0, true },
         };
 
+	/* hk: I think this is not needed anymore:
```
mikeller: If we don't need it we should be removing it.

profile-widget/profilescene.cpp
```diff
@@ -339,30 +339,51 @@ void ProfileScene::updateAxes(bool diveHasHeartBeat, bool simplified)
...
+	while ( (1 - (_sumOfFractions + prefs.gasplot_frac)) < minProfileFraction )
+		prefs.gasplot_frac -= 0.05;
+
+	VerticalAxisLayout secondaryAxes[] = {
 		// Note: axes are listed from bottom to top, since they are added that way.
-		{ heartBeatAxis, 75.0, 0.0, prefs.hrgraph && diveHasHeartBeat },
-		{ percentageAxis, 50.0, 0.0, prefs.percentagegraph },
-		{ gasYAxis, 75.0, 0.0, ppGraphsEnabled(currentdc, simplified) },
-		{ temperatureAxis, 50.0, 2.0, true },
+		// hk: using absolute numbers here causes troubles with small windows! all changed to fractions of height
```
mikeller: The 'hk' isn't going to help much in an open source project with dozens of contributors - `git blame`  is a simple way to see who added the comment.

mikeller (commented):
Thank you for this - having a way to zoom into the O2 sensor graph is very useful for CCR dives:

image

I am not sure if zooming in / out through  preferences is the right way to go here - I would assume that most CCR divers will want to be able to flip between zooming into the depth graph and zooming into the ppO2 graph. So this could be solved with a button on the left of the profile.

kruegerha:
You can achieve the desired effect by switching on or off the O2 plot (and other N2 He)
Will remove the deactivated code and initials

profile-widget/profilescene.cpp
kruegerha: removed

profile-widget/profilescene.cpp
kruegerha: removed

mikeller:
When testing a bit more I get a `SIGSEGV` when opening the preferences dialogue when a CCR dive is showing:

```
0x00007ffff76ca2c0 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
(gdb) bt
#0  0x00007ffff76ca2c0 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#1  0x00007ffff76ca46b in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#2  0x00007ffff76be0fc in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#3  0x00007ffff76be20c in QHeaderView::initializeSections(int, int) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#4  0x00007ffff76c2395 in QHeaderView::initializeSections() () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#5  0x00007ffff76c714e in QHeaderView::reset() () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#6  0x00007ffff5cdf7c8 in ?? () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#7  0x00007ffff5c41071 in QAbstractItemModel::modelReset(QAbstractItemModel::QPrivateSignal) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#8  0x0000555555848c3f in DivePlannerPointsModel::cylindersChanged (this=0x555555cd6a00 <DivePlannerPointsModel::instance()::self>)
    at /home/mikeller/git/subsurface/qt-models/diveplannermodel.cpp:601
#9  0x000055555576673b in PlannerWidgets::setDiveMode (this=0x55555609bfb0, mode=0) at /home/mikeller/git/subsurface/desktop-widgets/diveplanner.cpp:682
#10 0x00007ffff5cdf793 in ?? () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#11 0x00007ffff5cdf793 in ?? () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#12 0x0000555555ace1c8 in PreferencesDialog::refreshPages (this=<optimized out>) at /home/mikeller/git/subsurface/desktop-widgets/preferences/preferencesdialog.cpp:107
#13 0x00005555557a73f5 in MainWindow::on_actionPreferences_triggered (this=<optimized out>) at /home/mikeller/git/subsurface/desktop-widgets/mainwindow.cpp:603
#14 0x00005555556fbe73 in MainWindow::qt_metacall (this=0x55555608b8a0, _c=QMetaObject::InvokeMetaMethod, _id=11, _a=0x7fffffffcf70)
    at /home/mikeller/git/subsurface/build/desktop-widgets/subsurface_interface_autogen/EWIEGA46WW/moc_mainwindow.cpp:371
#15 0x00007ffff5cdf4e5 in ?? () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#16 0x00007ffff742abe6 in QAction::triggered(bool) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#17 0x00007ffff742d8fc in QAction::activate(QAction::ActionEvent) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#18 0x00007ffff75b8142 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#19 0x00007ffff75bfd39 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#20 0x00007ffff74744ee in QWidget::event(QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#21 0x00007ffff7431713 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#22 0x00007ffff7439364 in QApplication::notify(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#23 0x00007ffff5ca7e3a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#24 0x00007ffff7437e47 in QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) ()
   from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#25 0x00007ffff748e843 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#26 0x00007ffff7490fd5 in ?? () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#27 0x00007ffff7431713 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#28 0x00007ffff5ca7e3a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#29 0x00007ffff67ac307 in QGuiApplicationPrivate::processMouseEvent(QWindowSystemInterfacePrivate::MouseEvent*) () from /lib/x86_64-linux-gnu/libQt5Gui.so.5
#30 0x00007ffff6781a2c in QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Gui.so.5
#31 0x00007ffff0d59d6e in ?? () from /lib/x86_64-linux-gnu/libQt5XcbQpa.so.5
#32 0x00007ffff4221d3b in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#33 0x00007ffff42772b8 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#34 0x00007ffff421f3e3 in g_main_context_iteration () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#35 0x00007ffff5d010b8 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#36 0x00007ffff5ca675b in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#37 0x00007ffff5caecf4 in QCoreApplication::exec() () from /lib/x86_64-linux-gnu/libQt5Core.so.5
#38 0x00005555556e95dd in main (argc=<optimized out>, argv=<optimized out>) at /home/mikeller/git/subsurface/subsurface-desktop-main.cpp:106
```

> You can achieve the desired effect by switching on or off the O2 plot (and other N2 He) 

Not sure if it will be a desirable user experience for CCR dives to have to choose between viewing the dive profile at a decent size but not seeing any gas information, or seeing the gas information but only a minimised profile.

Before I forget it, this will also need testing for the mobile version - we are using a simplified profile there and a limited set of preferences. The CICD builds are running now, this will give us an artifact for android.

kruegerha:
Segmentation fault:  strange, I tested quite a bit here and did not observe any. Possible related to the ui? I made the changes to preferences_graph.ui with qtdesigner6... I'll later try with qtdesigner5, maybe some of the produced attributes?

To me it appeared as a good solution to have the fraction/size of the gasplot customizable. 
One possible alternative would be a button (on the panel left to the profile) to switch between two states (one like before) and a fixed zoom (like gasplot_frac = 0.5)?   Any other ideas?

Didn't think about the mobile version yet. I guess the prefs.gasplot_frac will stay there with the default setting, which should look almost as before. If we want to have that changeable, we'll have to add it to the settings (as there are no buttons and panels), which would support to have it on the desktop in the settings too.

kruegerha:
is there anything special about installing the android artifacts? I get a very unspecific error, like "you cannot install this app on your device"

mikeller:
> is there anything special about installing the android artifacts? I get a very unspecific error, like "you cannot install this app on your device"

You'll probably have to uninstall the already installed build of Subsurface first before you can sideload this one. When building pull request artifacts a temporary signing key is generated for each build, and android refuses to update an application with a version that was signed with a different key.
Or if this is a company owned device, a policy might block you from installing applications signed with a temporary key.

mikeller:
@kruegerha:

Apologies, I didn't see this until now.

> Segmentation fault: strange, I tested quite a bit here and did not observe any. Possible related to the ui? I made the changes to preferences_graph.ui with qtdesigner6... I'll later try with qtdesigner5, maybe some of the produced attributes?

Looks like it was the artefacts added by QtDesigner - it is working fine now.
In general I find that for small UI changes like this it's almost not worthwhile setting up a proprietary IDE - editing the .ui file by hand is faster and avoids the noise from the IDE 'decorating' the code.

> To me it appeared as a good solution to have the fraction/size of the gasplot customizable. One possible alternative would be a button (on the panel left to the profile) to switch between two states (one like before) and a fixed zoom (like gasplot_frac = 0.5)? Any other ideas?

Yes, I think having flexibility in the sizing of the plot components is good - depending on screen size users might want to use different arrangements. It's just that when looking at profiles from different dive types (CCR/sensors, CCR/fixed SP, OC,...), one might want to be able to switch between different zoom configurations quickly. And having to hide the gas plot entirely isn't a great option for this, in particular for CCR/fixed SP where automatic setpoint changes are shown in the gas plot.

So I think that keeping the preference, but adding a button to switch between the default (0.1) and the custom setting will cover all use cases and yield good usability.

There is a possibility that the 'Scale graph' button doesn't do anything meaningful at the moment - I'll investigate this, and if confirmed we might be able to re-use or trade in this one:

image

> Didn't think about the mobile version yet. I guess the prefs.gasplot_frac will stay there with the default setting, which should look almost as before. If we want to have that changeable, we'll have to add it to the settings (as there are no buttons and panels), which would support to have it on the desktop in the settings too.

The mobile app uses a simplified version of the profile with much less customisation. So I think leaving it as it is will be fine. We will just have to test to make sure the change does not break the simplified profile.

kruegerha:
I just empirically found out, that the scale graph seems to change something for shallow dives (like setting the max depth at the y-axis to 32?).  For most of my dives this doesn't change a thing, so I was wondering what that is supposed to do.

For the mobile version, I just tested a bit, seems to look as before. However, to me it seems as there is some space reserved for the temperature axis, which is actually not shown (if pO2 graph is on). Note the empty gap between profile and pO2 plot.  Not sure, if I can abuse the PR as an remote cross-compiler for testing?!

btw:  in an unrelated PR (4432) , I messed up the DCO, because of inconsistency of my mail address. Will someone have a look at that anyway? It seems I cannot fix it, without removing and committing another PR...   - sorry for bothering you with this

mikeller:
> I just empirically found out, that the scale graph seems to change something for shallow dives (like setting the max depth at the y-axis to 32?). For most of my dives this doesn't change a thing, so I was wondering what that is supposed to do.

I found this in the documentation:

image

(an outdated icon and caption didn't make it easy to locate)

So it looks like the default behaviour of the profile is to use a fixed depth scaling for all dives up to 27m depth. The depth seems to be fairly arbitrary, but I guess this is good for recreational divers (probably 90% of our users) wanting to be able to flick through their log and easily see shallow vs. deeper dives. And the button then allows them to zoom in.

I am wondering if we can overload this:
- when not selected, keep the default behaviour (small gas graph, no scaling for shallow dives);
- when selected, scale all dives, and extend the gas graph to the fraction selected in the preferences (if any gas is enabled).

Since the gases are not enabled by default, this will actually not change the behaviour for recreational divers.

But it should probably be added to the documentation to make it understandable.

> For the mobile version, I just tested a bit, seems to look as before. However, to me it seems as there is some space reserved for the temperature axis, which is actually not shown (if pO2 graph is on). Note the empty gap between profile and pO2 plot. Not sure, if I can abuse the PR as an remote cross-compiler for testing?!

Yes, space usage there isn't ideal, and there might be room for improvement. The caveat is that I don't know how well we are adapting to different screen sizes, so what shows as empty space on one screen might be 'fully used' on another.

> btw: in an unrelated PR (4432) , I messed up the DCO, because of inconsistency of my mail address. Will someone have a look at that anyway? It seems I cannot fix it, without removing and committing another PR... - sorry for bothering you with this

You should be able to fix this by editing the `Signed-off-by:` line in the commit message:

```
git switch <your branch>
git commit --amend
<edit message, fix DCO>
git push --force-with-lease
```

Happens to me as well...

kruegerha:
ok, I'll couple the use of gasplot_frac to prefs.zoomed_plot in a way, that gasplot_frac is used only, when zoomed_plot is active otherwise an fixed value of 0.1 is internally used (which looks like before).
I'll limit the range in which gasplot_frac may be changed from 0.2 – 0.7. (A value of 0.1, does not make sense anymore, as thats the same as zoomed_plot switched off).
A default value of gasplot_frac = 0.3 is probably a suitable choice.

I assume zoomed_plot is alway false on mobile... consequently this will have no impact on mobile.

Documentation... where to edit? 

Thanks for help with git :-)

kruegerha:
The hard limits of the viewport envelope is 30minutes/30metres, without the zoomed_plot.
However, the max depth of the resulting plot also depends on the depth raster intervals (generic vs. 3m intervals).
Also I think 30 minutes is very short, even for OWD-level rec dives. Anyhow, thats a different story.

mikeller:
> ok, I'll couple the use of gasplot_frac to prefs.zoomed_plot in a way, that gasplot_frac is used only, when zoomed_plot is active otherwise an fixed value of 0.1 is internally used (which looks like before). I'll limit the range in which gasplot_frac may be changed from 0.2 – 0.7. (A value of 0.1, does not make sense anymore, as thats the same as zoomed_plot switched off). A default value of gasplot_frac = 0.3 is probably a suitable choice.

Sounds good - I'd keep 0.1 in the acceptable range, for backwards compatibility as this will restore the behaviour prior to this change in the case of gas plots being enabled.
I am also wondering if we want to change the scaling of the preference from a fraction to a percentage value, and make it an `unsigned int`? 

> I assume zoomed_plot is alway false on mobile... consequently this will have no impact on mobile.

I think so.

> Documentation... where to edit?

The (english) master version is in https://github.com/subsurface/subsurface/blob/master/Documentation/user-manual.txt The other formats are generated from it, and the translations are maintained manually by translators.

mikeller:
> The hard limits of the viewport envelope is 30minutes/30metres, without the zoomed_plot. However, the max depth of the resulting plot also depends on the depth raster intervals (generic vs. 3m intervals). Also I think 30 minutes is very short, even for OWD-level rec dives. Anyhow, thats a different story.

I do not know any of the background that led to the implementation that we currently have, but it has enough complexity / effort put into it that I am wary of running into Chesterton's fence if it was changed wantonly.

kruegerha:
not sure what the benefit of changing gasplot_frac to an unsigned int would be.
It would cause quite some extra work without changing anything except from displaying percentages instead of fraction.
So I would like to keep it as is.

Good point on the lower limit of 0.1 to have an option to keep traditional behaviour.

I'll add it to the documentation. The images in the documentation need to be jpegs for some reason? Otherwise png's would look a lot better (Pref_tech.jpg)

kruegerha:
I added this feature to the user manual... also replaced two images there.
Maybe someone could check if that looks ok in the compiled manual. 
The manual doesn't build properly on my system and I'm not sure whether that is related to missing tools or due to the changes I inflicted.

kruegerha:
> > The hard limits of the viewport envelope is 30minutes/30metres, without the zoomed_plot. However, the max depth of the resulting plot also depends on the depth raster intervals (generic vs. 3m intervals). Also I think 30 minutes is very short, even for OWD-level rec dives. Anyhow, thats a different story.
> 
> I do not know any of the background that led to the implementation that we currently have, but it has enough complexity / effort put into it that I am wary of running into Chesterton's fence if it was changed wantonly.

git blames Linus ~2011 for the comments on this 30-30 scaling in profile.cpp ....  Probably not touched since then

dirkhh:
> git blames Linus ~2011 for the comments on this 30-30 scaling in profile.cpp .... Probably not touched since then

This was actively discussed early on and there was strong user feedback that they want to see at a glance if a dive is very short / very shallow. And I still will tell you that that's true. I wouldn't be supportive of a change there.

kruegerha:
no plans to change this, I was just curious.

mikeller:
> Maybe someone could check if that looks ok in the compiled manual. The manual doesn't build properly on my system and I'm not sure whether that is related to missing tools or due to the changes I inflicted.

https://github.com/subsurface/subsurface/actions/runs/13070900228/artifacts/2516113039 contains the generated documents. And have a look at https://github.com/subsurface/subsurface/blob/master/.github/workflows/documentation.yml for the toolchain that we are using to generate the documentation in CICD.

And don't worry about the failing check - #4435 will get rid of that.

kruegerha:
After checking the generated documention files, I observe that many of the icons are missing in the created pdf.
(html is good)
This affect .jpg as well as .png's, however not all of them.

Same behavior when compiling locally.  Maybe files used as icons need to have specific size/properties?

kruegerha:
fixed up some wasted space in profiles on the mobile platforms, which occurred when no temperature plot was shown (e.g. with CCR dives). See pics, comparing before and after the fix:
d1_master
d1_mobiprof2

mikeller:
@kruegerha: This needs a rebase and conflict resolution - are you able to do this?

PR #4429, opened 2025-01-24
profile: fix for temperature items

### LINKED ISSUE #4407: truncated temp label in dive profile panel
### Describe the issue:

- [x] Bug
- [ ] Change request
- [ ] New feature request
- [ ] Discussion request

### Issue long description:

In the dive profile panel, under several specific circumstances the water temperature label is truncated - it is displayed too far left and the left-most digit is obscured.  Resizing the window or panels does not make it visible.

This always occurs if the dive has a completely constant temperature (eg due to limited resolution of the dive computer temp sensor). 

It also happens infrequently with non-constant temp profiles, but requires at least the temperature to change between the first and second (?) time samples. However it doesn't always happen under these conditions.

### Operating system:

Affects Linux Ubuntu 22.04 and MacOS Sequoia at least. Bug is present in all subsurface versions that I tried, 5.0.10-XX to 6.0.XXXX

Linux Ubuntu 22.04, Subsurface 6.0.5306 (latest github checkin). locally compiled.
MacOS Sequoia M1, Subsurface 6.0.5241-CICD Official build
MacOS Sequoia M1, Subsurface 5.0.10-98 Official build

### Steps to reproduce:

Load a dive profile with a completely constant temperature profile (eg, due to the dive computer only reporting integral degree temperatures)

### Current behavior:

Temperature label is offset too far left, first digit is truncated.

### Expected behavior:

Temperature label should remain within the panel area and not be truncated.

### Additional information:
Temp label should be 29.0°C, rather than the "9.0°C" in the screenshots
sub-const
sub-delta

kruegerha:
The last of the textItems in DiveTemperatureItem::replot is aligned to the right. This causes a problem, if there is only one item. Make the first item alway aligned to the left.

Fixes #4407

Reported-by: @KimDelmar

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/profile-widget/diveprofileitem.cpp
+++ b/profile-widget/diveprofileitem.cpp
@@ -404,7 +404,12 @@ void DiveTemperatureItem::replot(const dive *, int fromIn, int toIn, bool)
 
 	for (size_t i = 0; i < textItems.size(); ++i) {
 		auto [sec, mkelvin] = textItems[i];
-		createTextItem(sec, mkelvin, i == textItems.size() - 1);
+		// never treat the first item as the last one (different text alignmet)
+		// even if there is only one. // Fixes bug #4407
+		if ( i == 0 )
+			createTextItem(sec, mkelvin, false);
+		else
+			createTextItem(sec, mkelvin, i == textItems.size() - 1);
 	}
 }
 
@@ -413,6 +418,7 @@ void DiveTemperatureItem::createTextItem(int sec, int mkelvin, bool last)
 	temperature_t temp;
 	temp.mkelvin = mkelvin;
 
+	// the last item should aligned to the right to fit in the plot
 	int flags = last ? Qt::AlignLeft | Qt::AlignBottom :
 			   Qt::AlignRight | Qt::AlignBottom;
 	auto text = std::make_unique<DiveTextItem>(dpr, 0.8, flags, this);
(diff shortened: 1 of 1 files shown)


profile-widget/diveprofileitem.cpp
```diff
@@ -404,7 +404,12 @@ void DiveTemperatureItem::replot(const dive *, int fromIn, int toIn, bool)
 
 	for (size_t i = 0; i < textItems.size(); ++i) {
 		auto [sec, mkelvin] = textItems[i];
-		createTextItem(sec, mkelvin, i == textItems.size() - 1);
+		// never treat the first item as the last one (different text alignmet)
+		// even if there is only one. // Fixes bug #4407
+		if ( i == 0 )
```
mikeller: The conditions could probably be combined to `i != 0 && i == textItems.size() - 1` to make it easier to understand what is done here, and remove the need for a comment.

mikeller (approved):
Apart from the cosmetic remark this is looking good:

image

mikeller (approved):

profile-widget/diveprofileitem.cpp
kruegerha: you're correct, it's probably easier to understand like this

mikeller (approved):

PR #4431, opened 2025-01-27
get_depth_string: decimals in metric max depth

kruegerha:
A hard limit of 20m was coded, above which one decimal, and below no decimal was included in the string.
This somehow arbitrarily chosen limit may now be changed by the knowing user by adding the "maxDepth_metric_decimal_limit" keyword to the [GeneralSettings] section of Subsurface.conf.
The default is still 20, consequently there is no change for the untaught user.

I don't think that this is of interest to many users, therefore it may stay as a 'secret feature'.
However, I know someone to whom this random limit is a major inconvenience ;-)

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/pref.cpp
+++ b/core/pref.cpp
@@ -96,7 +96,8 @@ preferences::preferences() :
 	allowOcGasAsDiluent(false),
 	coordinates_traditional(true),
 	unit_system(METRIC),
-	units(SI_UNITS)
+	units(SI_UNITS),
+	maxDepth_metric_decimal_limit(20.0)
 {
 }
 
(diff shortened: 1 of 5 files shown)


mikeller:
Not sure that 'secret features' are a good way to extend the existing functionality - doing it this way will mean that it will minimise the value (as nobody will use it unless they come across it by accident), while still contributing to complexity and maintenance effort.
My hunch would be that 20 was probably chosen as the limit to minimise the space taken up by the numbers and make them fit on small displays - so if we want to adopt this feature then there shouldn't be a reason to not make it configurable in the preferences, and make it discoverable for all users.

kruegerha:
you're right. I'll send a more general approach later

PR #4435, opened 2025-01-31
Documentation: Remove Checked-in Artifacts.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Remove checked-in artifacts, build documentation
artifacts for every build instead.

### Changes made:

1. remove the checked in documentation artifacts (\*.git);
2. add all of the formerly checked in documentation artifacts to the documentation build;
3. remove the check for changes in the documentation from the documentation workflow;
3. run the documentation workflow on every pull request and push;
4. add links to the documentation artifacts as pull request comments

### Additional information:

@dirkhh: After the rigamarole in #4434 to add checks to detect updated documentation and then ask people to update the checked-in artifacts, I think that this is the wrong way round, and we should rather get rid of the checked-in artifacts, and instead move to building the documentation for every pull request / push.
What are your thoughts on this?
What I did not manage is to find a way to build `mobile-manual-v2.html`, `mobile-manual_es.html`, and `mobile-manual_fr.html` to be identical to the checked in artifacts - the formatting seems to be fundamentally different. Do you have any instructions on how to build these correctly?


atdotde:
The original idea of having checked in artifacts was that many people don't have the tools installed to build these.

dirkhh:
The checked in docs make it much easier to build 
But then again, fewer and fewer people build anymore, clearly

glance-:
I think the best solution to that would be a helper who just grabs the latest package of artifacts from github.

mikeller:
> The checked in docs make it much easier to build But then again, fewer and fewer people build anymore, clearly

But what is the reason that we need the actual documentation files for local builds in the first place?

> I think the best solution to that would be a helper who just grabs the latest package of artifacts from github.

I am going to look into doing this for the CICD builds. But since there is no canonical URL to download GitHub artifacts from, I suspect that this might not work easily for local builds, unless we do something to make them accessible, like adding them to the releases.

dirkhh:
> > The checked in docs make it much easier to build But then again, fewer and fewer people build anymore, clearly
> 
> But what is the reason that we need the actual documentation files for local builds in the first place?

Because they are bundled into some of the builds

mikeller:
@dirkhh:

> Because they are bundled into some of the builds

But do we have to use the actual documentation for this on local builds?

Right now we are not doing a great job of manually updating the checked in documentation (which is why `master`is failing), so it's not like users are getting what they are meant to be getting, just something that is deceptively close. And that applies to our releases as well 😢 

Could we just change our builds so that by default they don't try to build / include documentation, and use a parameter to trigger this on CICD builds? (I am not sure if we'd want to consume pre-built documentation there, or if we'd want to build it as part of every CICD build that uses it...)

mikeller:
I finally got the inspiration (and stamina) to make this into something.

So now the changes are:
1. remove the checked in documentation artifacts (*.git);
2. add all of the formerly checked in documentation artifacts to the documentation build;
3. move the target directory for the documentation build into `Documentation/output` (to facilitate caching / hash generation);
4. flip the parameters for the build script to default to not build / use documentation (e.g. remove `NO_DOCS`, add `-build-docs` (built the documentation and use it) and `-install-docs` (install previously built / cached documentation)). This means that going forward, local builds on developer machines will not attempt to build / install the documentation, and consequently have no dependency on `asciidoc` or checked-in artifacts any more;
5. have the build script only try to install the documentation if either `-build-docs` or `-install-docs` are set. In other cases the developer will just get a documentation browser stating 'Cannot find the Subsurface manual' - which  should be just fine for development;
6. create a reusable workflow job to build an then cache the documentation;
7. add the job in 6. to all workflows generating desktop artifacts, and set `-install-docs` on these. All other builds will not try to build or include documentation;
8. add a dedicated workflow to build the documentation and publish it as an artifact;
9. update the configuration for externally run builds (for ubuntu / fedora /snapCraft packages) to keep building the documentation as part of the build.

Screenshot 2025-03-03 110159

@dirkhh, @saviq: Can I please get your help to review the changes for 9., and then verify that they were successful once this has been merged?

dirkhh (commented):
I can't see anything obviously wrong, but I'll admit that a lot of the yaml stuff I can never figure out reading code (and never get right writing, either), so as far as I'm concerned the proof is in the doing.

One comment on the mobile-manual-v2 thingy. That's an old old leftover from a transition years ago. No point in keeping that around.

Issue #4442, opened 2025-02-05
unexpected behaviour with 2 gas on TERIC


Julaye:
Desktop version : 6.0.5299 
Mobile version : 6.0.5276

Have some dives with 2 gas : 
- 21% air on the principal, with a probe
- 50% O2 on the stage

For some reason, the TERIC is sorting the gas by % of O2, so stage is on position 1 and principal on position 2 (even if active first) in the gas list on the TERIC list of gas.

After the dive (without gas switch), seems the software is lost with a mixture. See the snapshots ! 

First snapshot : principal is now the 50% O2 ... but the bar values (start at 219 bar and finish at 62 bar with datas from the probe) are right !

![Image](https://github.com/user-attachments/assets/bea7c797-b1cb-49f8-a360-8f14af238bd8)

Second snapshot, same issue in the editing window, principal is 50% O2 ... :

![Image](https://github.com/user-attachments/assets/e406f46c-8fc3-4b23-a37d-b4bd94666399)

The same behaviour on mobile and desktop version.

PS: the same mix if I do a gas switch, the 02 information is not with the good tank ...

What do you need to investigate ?

mikeller:
> For some reason, the TERIC is sorting the gas by % of O2, so stage is on position 1 and principal on position 2 (even if active first) in the gas list on the TERIC list of gas.

Yes, this is the way Shearwater dive computers work, and there is no way to change it.

> After the dive (without gas switch), seems the software is lost with a mixture. See the snapshots !
> 
> First snapshot : principal is now the 50% O2 ... but the bar values (start at 219 bar and finish at 62 bar with datas from the probe) are right !
> 
> ![Image](https://github.com/user-attachments/assets/bea7c797-b1cb-49f8-a360-8f14af238bd8)
> 
> Second snapshot, same issue in the editing window, principal is 50% O2 ... :
> 
> ![Image](https://github.com/user-attachments/assets/e406f46c-8fc3-4b23-a37d-b4bd94666399)

Yes, this is normal as well - the dive computer does have no concept of which tank sensor belongs wo which gas, so the default it to associate the first tank sensor with the first gas in the list, and so on.

But if you go into the gas list on the Equipment tab, and right-click into a list heading, you can then check the 'Sensors' column to make it visible:

![Image](https://github.com/user-attachments/assets/39258d26-7399-4858-8952-33f7c132750d)

[...]

PR #4443, opened 2025-02-07
Desktop: Show Cylinder Number in the List of Cylinders.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add the cylinder number as a column in the list of cylinders. This will
make it easier for users to interpret gas changes in the dive profile.

### Additional information:

image

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveEquipment.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveEquipment.cpp
@@ -15,7 +15,7 @@
 
 static bool ignoreHiddenFlag(int i)
 {
-	return i == CylindersModel::REMOVE || i == CylindersModel::TYPE ||
+	return i == CylindersModel::NUMBER || i == CylindersModel::REMOVE || i == CylindersModel::TYPE ||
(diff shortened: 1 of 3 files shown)


bstoeger (approved):
LGTM.

I have an irrational distaste for the `setHeaderDataStrings`, but that is not a problem of this PR.

At one point, the code in `ignoreHiddenFlag` should probably be replaced by a `switch`, but OK.

mikeller:
> I have an irrational distaste for the `setHeaderDataStrings`, but that is not a problem of this PR.

Yeah, the code looks like a stuffup waiting to happen. What is a better approach for setting these header strings?

> At one point, the code in `ignoreHiddenFlag` should probably be replaced by a `switch`, but OK.

Can do. I am working on an improvement of how tank sensors are handled (following up on #4442), and will make the 'Sensors' column show by default, but hide if there aren't any sensors showing, so I'll have to touch this code anyway.

PR #4449, opened 2025-02-12
Desktop: Add Consistent Loading of Media with Invalid Extensions.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

We are allowing users to add RAW images with ivalid file name
extensions. Make this behaviour consistent by allowing users to add
videos with invalid file name extensions.

### Related issues:

Alternative to #4424.

### Additional information:

image

### DIFF
--- a/desktop-widgets/divelistview.cpp
+++ b/desktop-widgets/divelistview.cpp
@@ -68,8 +68,8 @@ DiveListView::DiveListView(QWidget *parent) : QTreeView(parent),
 	QSettings s;
 	s.beginGroup("DiveListColumnState");
 	for (int i = 0; i < model()->columnCount(); i++) {
-		QString title = QString("%1").arg(model()->headerData(i, Qt::Horizontal).toString());
-		QString settingName = QString("showColumn%1").arg(i);
+		QString title = QStringLiteral("%1").arg(model()->headerData(i, Qt::Horizontal).toString());
+		QString settingName = QStringLiteral("showColumn%1").arg(i);
 		QAction *a = new QAction(title, header());
 		bool showHeaderFirstRun = i == DiveTripModelBase::NR ||
 					  i == DiveTripModelBase::DATE ||
@@ -103,11 +103,11 @@ DiveListView::~DiveListView()
 			continue;
 		// we used to hardcode them all to 100 - so that might still be in the settings
 		if (columnWidth(i) == 100 || columnWidth(i) == initialColumnWidths[i])
-			settings.remove(QString("colwidth%1").arg(i));
+			settings.remove(QStringLiteral("colwidth%1").arg(i));
 		else
-			settings.setValue(QString("colwidth%1").arg(i), columnWidth(i));
+			settings.setValue(QStringLiteral("colwidth%1").arg(i), columnWidth(i));
 	}
-	settings.remove(QString("colwidth%1").arg(DiveTripModelBase::COLUMNS - 1));
+	settings.remove(QStringLiteral("colwidth%1").arg(DiveTripModelBase::COLUMNS - 1));
 	settings.endGroup();
 }
 
@@ -176,7 +176,7 @@ void DiveListView::setColumnWidths()
 	for (int i = DiveTripModelBase::NR; i < DiveTripModelBase::COLUMNS; i++) {
 		if (isColumnHidden(i))
 			continue;
-		QVariant width = settings.value(QString("colwidth%1").arg(i));
+		QVariant width = settings.value(QStringLiteral("colwidth%1").arg(i));
 		if (width.isValid())
 			setColumnWidth(i, width.toInt());
 		else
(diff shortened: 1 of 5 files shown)


bstoeger:
Not sure I understand: The way I read this is that if either video-extraction is turned off by the user or ffmpeg refuses to start, then every non-image file is registered as a video file. Can you verify that?

Also: Does this call ffmpeg on every non-image file, or do we do some sort of analysis to check if this even makes sens?

Sorry - too long that I touched this code.

mikeller:
> Not sure I understand: The way I read this is that if either video-extraction is turned off by the user or ffmpeg refuses to start, then every non-image file is registered as a video file. Can you verify that?

Pretty much - every image that can not be loaded as either an image file or a RAW file will result in an attempt to load as a video file.

> Also: Does this call ffmpeg on every non-image file, or do we do some sort of analysis to check if this even makes sens?

No, we just do a 'hail mary' attempt to load as a video file - which is consistent with the approach we are taking for loading RAW files.

Alternatively, if we want to do a simple (extension based) analysis of the plausibility of files being of a certain type, and do this consistently for RAW and video files, this is implemented in #4424.

I am neither here nor there on which way this is done, but I think it should be done consistently.

bstoeger:
> Pretty much - every image that can not be loaded as either an image file or a RAW file will result in an attempt to load as a video file.

Perhaps I'm missing something, but the way I read the code is that if  `prefs.extract_video_thumbnails` is `false`, then `Thumbnailer::fetchVideoThumbnail` will add a video thumbnail without image. So any random file which is not a picture would be shown as a video file.

If, on the other hand, `prefs.extract_video_thumbnails` is `true`, then `Thumbnailer::fetchVideoThumbnail` will call the video-frame extractor, which I believe starts ffmpeg. This is heavy in itself, but it comes even worse: even if it fails it registers the file as a video.

Therefore I think the analogy you draw is not correct: If the RAW-parser fails, then the file is not registered as an image. However, with your new code it will be registered as a video, even if the video parsing fails.

The point is: in `Thumbnailer::fetchImage` we only check for the file extension if our metadata-reader was not able to identify the file as a video. The check for the extension is a last-ditch heuristics, since we simply don't know what that file is. Note that under Linux file extensions have no inherent meaning (in contrast to Windows).

So perhaps the correct fix is to enhance our metadata-parser. Could you send me a video where it fails?

bstoeger:
> Alternatively, if we want to do a simple (extension based) analysis of the plausibility of files being of a certain type, and do this consistently for RAW and video files, this is implemented in https://github.com/subsurface/subsurface/pull/4424.

Actually, I don't think this is the same at all. It only starts the RAW parser if the file extension fits.

Making the behavior consistent would be checking for an image file extension (normal or raw) and then register the file as image, even if we couldn't parse it.

Though I am not sure that this is a reasonable behavior.

mikeller:
@bstoeger:

> Perhaps I'm missing something, but the way I read the code is that if `prefs.extract_video_thumbnails` is `false`, then `Thumbnailer::fetchVideoThumbnail` will add a video thumbnail without image. So any random file which is not a picture would be shown as a video file.

Good point, the behaviour for `prefs.extract_video_thumbnails == false` was indeed not desirable - I have reverted this case to the existing behaviour.

> If, on the other hand, `prefs.extract_video_thumbnails` is `true`, then `Thumbnailer::fetchVideoThumbnail` will call the video-frame extractor, which I believe starts ffmpeg. This is heavy in itself, but it comes even worse: even if it fails it registers the file as a video.

Not to my understanding - files for which no media type can be established and which fail to load in ffmpeg will be shown as unknown:

https://github.com/subsurface/subsurface/blob/21c3058652e29b736c7d0843701a538031f7eaec/core/imagedownloader.cpp#L420-L429

But yes, we waste some resources on having ffmpeg do a 'last ditch' attempt at files with unknonwn media type even if they don't have a 'known video' extension.

> Therefore I think the analogy you draw is not correct: If the RAW-parser fails, then the file is not registered as an image. However, with your new code it will be registered as a video, even if the video parsing fails.

See above.

> So perhaps the correct fix is to enhance our metadata-parser. Could you send me a video where it fails?

I tested this behaviour with an ogg file, like https://lachy.id.au/dev/markup/examples/video/bus.ogg

desktop-widgets/divelistview.cpp
```diff
@@ -831,15 +831,11 @@ void DiveListView::shiftTimes()
 void DiveListView::loadImages()
 {
 	QStringList m_filters = mediaExtensionFilters();
-	QStringList i_filters = imageExtensionFilters();
-	QStringList v_filters = videoExtensionFilters();
 	QStringList fileNames = QFileDialog::getOpenFileNames(this,
 							      tr("Open media files"),
 							      lastUsedImageDir(),
-							      QString("%1 (%2);;%3 (%4);;%5 (%6);;%7 (*.*)")
+							      QString("%1 (%2);;%3 (*.*)")
```
bstoeger: Since you're touching this line - this should probably be `QStringLiteral(...)`. Not that it matters, but out of principle...

bstoeger:
You seem to know the code better than I do. :)

This appears reasonable to me now.

I'll see if an OGG parser is a simple thing (unfortunately very busy at the moment).

mikeller:
> I'll see if an OGG parser is a simple thing (unfortunately very busy at the moment).

OGG was really just an example that I used for a not-so-common video format.
I think what we are doing at the moment is a reasonable approach: We natively recognise the most common video formats, but if users want to use uncommon ones they'll have to have ffmpeg installed and configured. It's just that until now this was a bit hard because we excluded uncommon formats based on filename extension.

mikeller:
@bstoeger:

Looks like the problem was that we were doing it wrong due to an off-by-one callback:

[`5f3a3af` (#4449)](https://github.com/subsurface/subsurface/pull/4449/commits/5f3a3afe2ed0b3bd8422d032545c7dc1c3dc6b9b)

mikeller:
@bstoeger: What do you think of this version now?

bstoeger:
Looks good to me.

PR #4450, opened 2025-02-16
Export: Fix 'rating' in UDDF Export.

### LINKED ISSUE #4444: UDDF export: rating in informationafterdive should  not get exported if ratingvalue is empty
In Mac version 6.0.5229-CICD-release the UDDF export exports empty ratingvalue inside of structures in informationafterdive. If a ratingvalue is empty it should not get exported and probably also rating itself should not get exported in this case.

It would be great if you could fix this. It let's UDDF parsers fail.

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Make the export of UDDF files compliant with the UDDF 3.2.3
specification for the `rating` element.

### Related issues:

Fixes #4444.

### DIFF
--- a/xslt/uddf-export.xslt
+++ b/xslt/uddf-export.xslt
@@ -679,18 +679,14 @@
             <link ref="{concat(../@number, ':', @filename, '@', @offset)}"/>
           </xsl:for-each>
         </notes>
-        <rating>
-          <ratingvalue>
-            <xsl:choose>
-              <xsl:when test="./@rating = 0">
-                <xsl:value-of select="'1'"/>
-              </xsl:when>
-              <xsl:when test="./@rating != ''">
-                <xsl:value-of select="./@rating * 2"/>
-              </xsl:when>
-            </xsl:choose>
-          </ratingvalue>
-        </rating>
+        <xsl:if test="./@rating">
+          <rating>
+            <ratingvalue>
+              <!-- The UDDF rating scale is 1-10, while Subsurface uses 1-5. -->
+              <xsl:value-of select="round(1 + ((./@rating - 1) div 4 * 9))"/>
+            </ratingvalue>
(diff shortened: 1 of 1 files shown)


bstoeger:
XSLT is indecipherable to me. Paging @mturkia.

dirkhh:
> XSLT is indecipherable to me. Paging @mturkia.

it always feels like "I SHOULD be able to make sense of it, but... AAAAHHHHHHHHHHHHHHH"
🤣

mikeller:
@bstoeger , @dirkhh: Well, this one is simple really, and not functional at all - instead of always emitting a rating element, and then adding in an actual value only if one exists in the input, we now make the rating element conditional to a rating existing in the input, so that it is only emitted when it has a value.
Additionally I fixed the scaling of the emitted value from our 1 to 5 to UDDF's 1 to 10, as our XML never actually contains a'0' rating value.

But I am keen to see if @cheinelt can test this with whatever he tries to import the UDDF into - it might be that this is giving up early and there are other places that need a fix.

bstoeger:
Good, fine with me. Perhaps add a comment that says that the 1->5 range is mapped onto 1->10.

mikeller:
@bstoeger:

> Good, fine with me. Perhaps add a comment that says that the 1->5 range is mapped onto 1->10.

I let 'pilot do that for me. 😆

mturkia (approved):
LGTM

bstoeger:
> But I am keen to see if @cheinelt can test this with whatever he tries to import the UDDF into - it might be that this is giving up early and there are other places that need a fix.

It improves the export, so we can merge and add other fixes later.

PR #4451, opened 2025-02-17
Desktop: Fix Automatic Firmware Upgrade for Heinrichs Weikamp.

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix the automatic firmware upgrade after downloading dives by not
closing the download dialog until the firmware download is complete.

### Changes made:

1. fix the automatic firmware upgrade after downloading dives by not
   closing the download dialog until the firmware download is complete;
2. change Heinrichs Weikamp download links to https;
3. amend the firmware upgrade dialog to point out that Bluetooth has
   to be re-enabled for the upgrade.

### Additional information:

Screenshot From 2025-02-18 00-14-30

image

### DIFF
--- a/desktop-widgets/configuredivecomputerdialog.cpp
+++ b/desktop-widgets/configuredivecomputerdialog.cpp
@@ -268,14 +268,14 @@ OstcFirmwareCheck::OstcFirmwareCheck(const QString &product) : parent(0)
 	QUrl url;
 	devData = device_data_t();
 	if (product == "OSTC 3" || product == "OSTC 3+" || product == "OSTC cR" || product == "OSTC Plus") {
-		url = QUrl("http://www.heinrichsweikamp.net/autofirmware/ostc3_changelog.txt");
-		latestFirmwareHexFile = QString("http://www.heinrichsweikamp.net/autofirmware/ostc3_firmware.hex");
+		url = QUrl("https://www.heinrichsweikamp.net/autofirmware/ostc3_changelog.txt");
+		latestFirmwareHexFile = QUrl("https://www.heinrichsweikamp.net/autofirmware/ostc3_firmware.hex");
 	} else if (product == "OSTC Sport") {
-		url = QUrl("http://www.heinrichsweikamp.net/autofirmware/ostc_sport_changelog.txt");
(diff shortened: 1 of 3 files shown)


desktop-widgets/configuredivecomputerdialog.cpp
```diff
@@ -290,6 +290,7 @@ void OstcFirmwareCheck::parseOstcFwVersion(QNetworkReply *reply)
 	int firstOpenBracket = parse.indexOf('[');
 	int firstCloseBracket = parse.indexOf(']');
 	latestFirmwareAvailable = parse.mid(firstOpenBracket + 1, firstCloseBracket - firstOpenBracket - 1);
+	reply->close();
```
bstoeger, line 293: Is this necessary?

One would think that the reply gets destructed anyway?

mikeller: Not super sure it is strictly needed, but I think it can't hurt if we close the network connection before we move to the next step in the sequence that will do more http requests. I added this while I was trying to find why the download wasn't working - in the end it turned out to be early destruction of the `QNetworkManager` due to the closing of the dialog.

mikeller: @bstoeger: Are you ok with the way this is now, would you prefer the `close()` to be removed?

bstoeger: I don't object to it - it's up to you.

Frankly, the whole resource management in this file is crazy, which is probably due to the `QNetworkAccessManager` interface being a mess. Who owns `reply`? Do you have to delete it with `deleteLater()`? I skimmed the documentation and found it frustratingly unclear. This is a horrible pre-C++11 API.

mikeller: Agreed re. mess. But also not keen to spend eons cleaning up something that will affect only a fraction of users, and will only ever be used very infrequently - this is probably why it has been broken for a long time.
Re cleaning up of `reply`, I got it on good authority that this is indeed deleted (and the connection closed) when the owning object is destroyed. This is what caused the bug in the first place, as the second connection was closed from our side before the download of the firmware file was completed.

PR #4461, opened 2025-03-06
Desktop: Improve Management of Tank Sensors.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Improve the UI for the assignment of tank pressure sensors to cylinders, in particular in the case of multiple tank sensors.

@dirkhh: I know we have talked about this before, but I still think the current implementation of this functionality is not ideal:
When it is used, tank pressure sensor ids (`pressure<X>=Y`) are modified in the actual dive data samples. This irreversibly modifies the data that was downloaded from the dive computer, making it impossible for the user to ever go back to the original dive data. Furthermore, the user will not be able at a later stage to figure out which sensor was which, and map them to metadata like the sensor's serial number or battery level that is often supplied by the dive computer as extra data. In addition to this, this approach also doesn't allow the user to have an 'unused' sensor that isn't assigned to a cylinder.
For this reason I think it will better to add extra syntax to our formats to encode this mapping without having to modify the data downloaded from the dive computer. What I propose are zero to many occurrences of an extra child element for `<divecomputer>`:

```
<tank_sensor id='<id from samples>' cylinder='<cylinder id>'>
```

The absence of any of these mappings will be interpreted as a 1:1 mapping (i.e. the existing behaviour).
### Changes made:

1. fix bugs preventing sensors for tanks other than the first three from being shown;
2. remove the hiding of the 'Sensor' column by default;
3. only show the 'Sensor' column if the dive has tank sensors;
4. add start / end pressure to the tank sensor selection in the drop-down;
5. replace the misleading dynamic sensor index with 'X' in the table;
6. update the cylinder table after a sensor has been reassigned.

### Related issues:

https://groups.google.com/g/subsurface-divelog/c/K9u2SBEyUlw

### Additional information:

Screenshot From 2025-03-06 17-14-42

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ translations/*.qm
 *.exe
 *.dmg
 *.patch
-*.xml
+*.bak
 ssrf-version.h
 !dives/*.xml
 *~
@@ -46,3 +46,4 @@ gh_release_notes.md
 release_content_title.txt
 /output/
 /Documentation/output/
+/metainfo/subsurface.metainfo.xml
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -1311,7 +1311,7 @@ void EditSensors::mapSensors(int toCyl, int fromCyl)
 				sample.sensor[s] = fromCyl;
 		}
 	}
-	emit diveListNotifier.diveComputerEdited(dc);
+	emit diveListNotifier.diveComputerEdited(*d, *dc);
 	d->invalidate_cache(); // Ensure that dive is written in git_save()
 }
 
--- a/commands/command_edit.cpp
+++ b/commands/command_edit.cpp
@@ -11,6 +11,7 @@
 #include "core/selection.h"
 #include "core/subsurface-string.h"
 #include "core/tag.h"
+#include "core/tanksensormapping.h"
 #include "qt-models/weightsysteminfomodel.h"
 #include "qt-models/tankinfomodel.h"
 #ifdef SUBSURFACE_MOBILE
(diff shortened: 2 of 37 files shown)


dirkhh:
I think this is reasonable on the first glance 
I of course worry about unintended consequences and about changes to the user experience. How is this exposed to a user, especially in subsurface-mobile. 
That said, I'm supportive

mikeller:
> I of course worry about unintended consequences and about changes to the user experience. How is this exposed to a user, especially in subsurface-mobile.

The UI will not change, except that the now dynamic 'sensor index' in the drop-down will be fixed, and 1 based, even after sensors have been reassigned:

Screenshot From 2025-03-06 17-14-42

And there will be a new 'No sensor' entry in the drop-down, allowing the user to un-assign a given sensor.
And we could probably go back to showing the sensor index in the table instead of the 'X', as it will convey useful information now.

On the mobile side we are already not supporting editing of tank sensor mapping, and I am not intending to change this. So the only thing that needs to be done there is to correctly use the tank sensor mapping when plotting the tank graphs in the dive profile.

mikeller:
@dirkhh this is ready for review now - not sure if this needs testing or a format version update for the added `tanksensormapping` element in [git](https://github.com/subsurface/subsurface/pull/4461/files#diff-8a3f3f9df4a9050074cf22a06f228678efaaa7ef868b54213d420de599068d2fR131-R137) and [XML](https://github.com/subsurface/subsurface/pull/4461/files#diff-c45d99faabf4e3a15501196250efce6480cf4804275b8351217713c396592a2fR403-R413).

dirkhh:
Did you check that `export_html` is happy?

mikeller:
> Did you check that `export_html` is happy?

Yes I did. The current version outputs an error when a post-#4461 repository clone with tank sensor mappings is processed, but the output seems to be ok (except that the mappings are not applied obviously):

```
ERROR: Unmatched action 'tanksensormapping'
```

The post-#4461 binary doesn't output the error.

dirkhh:
Once merged I'll update the binaries on the four cloud servers

mikeller:
@dirkhh:

> Once merged I'll update the binaries on the four cloud servers

This has been merged now. You should be able to pull the binary from https://github.com/subsurface/subsurface/actions/runs/14118972753/artifacts/2835696853 (provided you want to run it on linux / x86).

Issue #4460, opened 2025-03-05
Cannot sync Cressi Donatello through Bluetooth Interface


2opremio:
I just purchased a [Cressi BT Interface for big screen computers](https://www.cressiusa.com/catalogue/cressitech.asp?id=1277):

<img src="https://github.com/user-attachments/assets/3143c0a6-5676-413a-b459-71396564357e" width="500"> 

It's a BT-only interface, without the option to use a USB cable. It works great when syncing my Cressi Donatello (FW 1.02) with the (very limited) Cressi app (both on iOS and on my mac).

However, Subsurface doesn't find the BT interface (I tried with both stable and nightly and in the same computer in which the Cressi app works fine):

<img src="https://github.com/user-attachments/assets/96c0f8c8-7851-4022-86ea-cc1ed909e1c2" width="500"> 

 Here's some info about the BT interface device, as obtained with the nrfConnect app:

<img src="https://github.com/user-attachments/assets/1ab9cf21-fca2-4e89-ae85-f63cf8ee0ab0" width="500">
<img src="https://github.com/user-attachments/assets/f1653f54-5fd6-4b56-9ad1-3d79c6800e82" width="500">
<img src="https://github.com/user-attachments/assets/404a5458-d976-4f25-b495-b5fb025ce042" width="500">
<img src="https://github.com/user-attachments/assets/889cdc21-e7ff-4e00-922e-6f52c43b04f5" width="500">
<img src="https://github.com/user-attachments/assets/51d31c8c-100e-493e-b559-5dd5eb746c26" width="500">

mikeller:
Have you tried checking 'Show all BT devices' and then selecting the device?

2opremio:
I've tried now, but when manually choosing the device name (`4_1e52`) I get an error:

<img width="747" alt="Image" src="https://github.com/user-attachments/assets/58e9c49a-a6e0-4ab7-bd80-17028e025d0a" />


Here is the log file:

```
Subsurface: v6.0.5347.0, built with libdivecomputer v0.9.0-devel-Subsurface-NG (6f19abf8d260dfb85257e33c08411dd08f9f4dc0)
[223.056111] INFO: Open: transport=32
[223.056120] INFO: Configure: baudrate=115200, databits=8, parity=0, stopbits=0, flowcontrol=0
[223.056122] INFO: Timeout: value=5000
[223.056123] INFO: RTS: value=0
[223.056124] INFO: DTR: value=0
[...]

PR #4463, opened 2025-03-06
Add support for libdivecomputer's DC_IOCTL_BLE_CHARACTERISTIC_READ

2opremio:
Most BLE enabled dive computers use a communication protocol which simply emulates UART communication over the BLE link. There is a BLE service containing two characteristics: one for sending and one for receiving. The high level communication is done by sending commands and receiving the response over this virtual UART, just like with plain serial (or usb-serial) communication.

The Cressi dive computers are a bit different in this regard because they do have a UART service and characteristics, but they also have a few extra characteristics for information like the model number, firmware version and serial number. In their serial communication protocol, this information could be read by sending a command, but for the BLE communication they removed this command and we now have to read the info from those extra characteristics instead. Unfortunately this information is crucial, so we can't proceed without it.

The new ioctl's were added to libdivecomputer to support reading/writing such secondary characteristics.

- [x] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

Implements support for libdivecomputer's `DC_IOCTL_BLE_CHARACTERISTIC_READ`.

In order to do so it keeps a list with all the BLE services of the device.

### Related issues:

Addresses part of #4460

### Documentation change:

None

### Mentions:

@jefdriesen

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -205,9 +205,6 @@ void BLEObject::addService(const QBluetoothUuid &newService)
 		return;
 	}
 
-	auto service = controller->createServiceObject(newService, this);
-	allSafeServices.append(service);
-
 	//
 	// If it's a known serial service, clear any other previous
 	// services we've found - we'll use this one.
@@ -227,7 +224,7 @@ void BLEObject::addService(const QBluetoothUuid &newService)
 	details = is_known_serial_service(newService);
 	if (details) {
 		report_info(" .. recognized service %s", details);
-		serialCandidateServices.clear();
+		services.clear();
 	} else {
 		bool isStandardUuid = false;
 
@@ -238,7 +235,7 @@ void BLEObject::addService(const QBluetoothUuid &newService)
 		}
 	}
 
-
+	auto service = controller->createServiceObject(newService, this);
 	if (service) {
 		// provide some visibility into what's happening in the log
 		service->connect(service, &QLowEnergyService::stateChanged,[=](QLowEnergyService::ServiceState newState) {
@@ -252,7 +249,7 @@ void BLEObject::addService(const QBluetoothUuid &newService)
 				 [=](QLowEnergyService::ServiceError newError) {
 			report_info("error discovering service details %d", static_cast<int>(newError));
 		});
-		serialCandidateServices.append(service);
+		services.append(service);
 		report_info("starting service characteristics discovery");
 		service->discoverDetails();
 	}
@@ -271,7 +268,7 @@ BLEObject::~BLEObject()
 {
 	report_info("Deleting BLE object");
 
-	qDeleteAll(allSafeServices);
+	qDeleteAll(services);
 
 	delete controller;
 }
@@ -420,7 +417,7 @@ dc_status_t BLEObject::read(void *data, size_t size, size_t *actual)
 dc_status_t BLEObject::select_preferred_service()
 {
 	// Wait for each service to finish discovering
-	for (const QLowEnergyService *s: serialCandidateServices) {
+	for (const QLowEnergyService *s: services) {
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
 		WAITFOR(s->state() != QLowEnergyService::RemoteServiceDiscovering, BLE_TIMEOUT);
 		if (s->state() == QLowEnergyService::RemoteServiceDiscovering)
@@ -432,7 +429,7 @@ dc_status_t BLEObject::select_preferred_service()
 	}
 
 	// Print out the services for debugging
-	for (const QLowEnergyService *s: serialCandidateServices) {
+	for (const QLowEnergyService *s: services) {
 		report_info("Found service %s %s", to_str(s->serviceUuid()).c_str(), qPrintable(s->serviceName()));
 
 		for (const QLowEnergyCharacteristic &c: s->characteristics()) {
@@ -444,7 +441,7 @@ dc_status_t BLEObject::select_preferred_service()
 	}
 
 	// Pick the preferred one
-	for (QLowEnergyService *s: serialCandidateServices) {
+	for (QLowEnergyService *s: services) {
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
 		if (s->state() != QLowEnergyService::RemoteServiceDiscovered)
 #else
@@ -769,27 +766,30 @@ dc_status_t BLEObject::get_name(char *data, size_t size)
 
 dc_status_t BLEObject::read_characteristic(const QBluetoothUuid &uuid, char *res, size_t size)
 {
-	for (QLowEnergyService *s: allSafeServices) {
-		const QList<QLowEnergyCharacteristic> chars = s->characteristics();
-		for (const QLowEnergyCharacteristic& ch : chars) {
-			if (ch.uuid() == uuid) {
-				report_info("Reading BLE characteristic %s from service %s",
-					to_str(uuid).c_str(),
-					to_str(s->serviceUuid()).c_str());
-				s->readCharacteristic(ch);
-				QByteArray val = ch.value();
-				if ((size_t)val.size() != size) {
-					report_error("unexpected read characteristic size (%lld): expected %lld", (long long)val.size(), (long long)size);
-					return DC_STATUS_DATAFORMAT;
-				}
-				report_info("Obtained characteristic value: %s", val.toHex().toStdString().c_str());
-				memcpy(res, val.data(), size);
-				return DC_STATUS_SUCCESS;
-			}
-		}
+	auto s = preferredService();
+	if (!s) {
+		return DC_STATUS_NODEVICE;
 	}
+
+	const QList<QLowEnergyCharacteristic> chars = s->characteristics();
+	for (const QLowEnergyCharacteristic& ch : chars) {
+		if (ch.uuid() != uuid) {
+			continue;
+		}
+		report_info("Reading BLE characteristic %s", to_str(uuid).c_str());
+		s->readCharacteristic(ch);
+		QByteArray val = ch.value();
+		if ((size_t)val.size() != size) {
+			report_error("Unexpected read characteristic size (%lld): expected %lld", (long long)val.size(), (long long)size);
+			return DC_STATUS_DATAFORMAT;
+		}
+		report_info("Obtained characteristic value: %s", val.toHex().toStdString().c_str());
+		memcpy(res, val.data(), size);
+		return DC_STATUS_SUCCESS;
+	}
+
 	// Not found
-	report_error("characteristic %s not found in any of the services", to_str(uuid).c_str());
+	report_error("characteristic %s not found", to_str(uuid).c_str());
 	return DC_STATUS_NOACCESS;
 }
 
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -796,16 +796,15 @@ dc_status_t BLEObject::read_characteristic(const QBluetoothUuid &uuid, char *res
 dc_status_t qt_ble_ioctl(void *io, unsigned int request, void *data, size_t size)
 {
 	BLEObject *ble = (BLEObject *) io;
-	struct quint128 uuid;
-	size_t readsize;
-	char *p;
-
 	switch (request) {
 	case DC_IOCTL_BLE_GET_NAME:
 		return ble->get_name((char *) data, size);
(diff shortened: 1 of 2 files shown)


core/qt-ble.cpp
```diff
@@ -195,6 +195,9 @@ void BLEObject::addService(const QBluetoothUuid &newService)
 
 	report_info("Found service %s", to_str(newService).c_str());
 
+	auto service = controller->createServiceObject(newService, this);
+	all_services.append(service);
```
mikeller: I think it is preferrable that we exclude bluetooth chip firmware upgrade services that are filtered below with `is_known_bad_service()` - we don't want to accidentally mess with those.

core/qt-ble.cpp
```diff
@@ -268,7 +271,7 @@ BLEObject::~BLEObject()
 {
 	report_info("Deleting BLE object");
 
-	qDeleteAll(services);
+	qDeleteAll(good_services);
```
mikeller: We probably want to clean up `all_services` as well here.

core/qt-ble.h
```diff
@@ -43,7 +44,8 @@ public slots:
 	dc_status_t setupHwTerminalIo(const QList<QLowEnergyCharacteristic> &allC);
 	dc_status_t setHwCredit(unsigned int c);
 private:
-	QVector<QLowEnergyService *> services;
+	QVector<QLowEnergyService *> good_services;
```
mikeller: Naming in our C++ code base is camelCase.

mikeller (commented):
Looks good, with a few suggestions. Thanks for looking into this.

core/qt-ble.cpp
```diff
@@ -764,13 +767,47 @@ dc_status_t BLEObject::get_name(char *data, size_t size)
...
+		const QList<QLowEnergyCharacteristic> chars = s->characteristics();
+		for (const QLowEnergyCharacteristic& ch : chars) {
+			if (ch.uuid() == uuid) {
+				report_info("Reading BLE characteristic %s from service %s",
+					to_str(uuid).c_str(),
+					to_str(s->serviceUuid()).c_str());
+				s->readCharacteristic(ch);
+				QByteArray val = ch.value();
+				if ((size_t)val.size() != size) {
+					report_error("unexpected read characteristic size (%d): expected %lu", val.size(), size);
```
mikeller: I just noted that some of the CICD builds failed - you will need to use `%lld` to print `size`, as some of the compilers used are less tolerant to incorrect format strings.

2opremio: Yeah, I noticed too. I am reading that the platform independent way is to use `"%zu"`

2opremio:
@mikeller PTAL . I created a separate commit with the fixes to make it more clear. Please squash on merge.

core/qt-ble.cpp
2opremio: But it turns out that the Windows compiler used in CI doesn't like `%zu` so I resorted to use `%lld` as you suggested.

mikeller (approved):

core/qt-ble.cpp
2opremio: Actually, I think that's wrong, since that will lead to multiple deletions (I am experiencing a segmentation fault).

Only `all_services` should be deleted since it contains any potential services in `good_services`.

I pushed a fix. Please double-check it since I am new to QT, and its object management in particular.

jefdriesen:
Just a minor remark. In the `read_characteristic` function you loop over all available services. That's not really wrong here, but the `DC_IOCTL_BLE_CHARACTERISTIC_{READ,WRITE}` ioctl's were defined with the assumption that these characteristics are all part of a single service, and that service is also the main uart service. Otherwise I would have added the service uuid to ioctl data.

I'm not 100% sure whether characteristics are supposed to be globally unique (across all services provided by the device) or only within the same service. If it's the latter, then you could in theory have two different services with the same characteristics.

2opremio:
>  ioctl's were defined with the assumption that these characteristics are all part of a single service, and that service is also the main uart service.

That wasn't clear to me. Maybe the documentation for `DC_IOCTL_BLE_CHARACTERISTIC_{READ,WRITE}` could be updated to reflect it.

It didn't help that Subsurface [was picking the wrong serial service](https://github.com/subsurface/subsurface/issues/4460#issuecomment-2705339951), distinct from the one expected from `libdivecomputer`. That's why this PR iterates through all the services and book-keeps them separately. I will now revert those changes.

@jefdriesen One suggestion, if I may. Wouldn't it be better if service discovery (or at least service selection) was handled by `libdivecomputer`? Having each `libdivecomputer` consumer hardcode the expected service UUIDs and implement service election independently doesn't seem to be great.

2opremio:
@jefdriesen @mikeller PTAL. I have squashed all the prior changes and created a new commit with the new one. Please squash them both together before merging.

core/qt-ble.cpp
```diff
@@ -764,13 +764,50 @@ dc_status_t BLEObject::get_name(char *data, size_t size)
...
+
+	// Not found
+	report_error("characteristic %s not found", to_str(uuid).c_str());
+	return DC_STATUS_NOACCESS;
+}
+
 dc_status_t qt_ble_ioctl(void *io, unsigned int request, void *data, size_t size)
 {
 	BLEObject *ble = (BLEObject *) io;
+	struct quint128 uuid;
```
mikeller: While variable hoisting exists in older parts of the Subsurface codebase, the current approach is to not do this, and instead declare the variable right before its use, and use the minimum scope for it: https://github.com/subsurface/subsurface/blob/master/CODINGSTYLE.md#coding-conventions

For more complex functions this often greatly improves readability.

core/qt-ble.h
```diff
@@ -28,6 +28,7 @@ class BLEObject : public QObject
 	dc_status_t write(const void* data, size_t size, size_t *actual);
 	dc_status_t read(void* data, size_t size, size_t *actual);
 	dc_status_t get_name(char *res, size_t size);
+	dc_status_t read_characteristic(const QBluetoothUuid &uuid, char *res, size_t size);
```
mikeller, line 31: This should be private - again this will make it easier to read the code.

2opremio: It can't be private, because it's used by `qt_ble_ioctl()`

core/qt-ble.cpp
2opremio: I agree (and that's even how I coded it first) but I was getting errors due to the switch statement. It turns out you can declare the variables within the switch if you don't initialize them at the same time https://stackoverflow.com/questions/92396/why-cant-variables-be-declared-in-a-switch-statement

I've done that now

mikeller: Yes, that works. Probably the more common way to get around this is to create a block for the case statement that starts with a variable initialisation:

```
case DC_IOCTL_BLE_CHARACTERISTIC_READ: {
[...]
}
case [...]
```

2opremio: Would you prefer the block? I think it’s fine without it.

mikeller: Nah, all good the way it is.

jefdriesen:
> @jefdriesen One suggestion, if I may. Wouldn't it be better if service discovery (or at least service selection) was handled by `libdivecomputer`? Having each `libdivecomputer` consumer hardcode the expected service UUIDs and implement service election independently doesn't seem to be great.

That's difficult because the bluetooth service discovery code is very platform specific. That's why I documented the services and characteristics on the wiki:

https://github.com/libdivecomputer/libdivecomputer/wiki/BLE

Note that for the built-in transports, libdivecomputer already handles this all internally. For example for USB devices we have an internal list of in/out endpoints to use:

https://github.com/libdivecomputer/libdivecomputer/blob/master/src/usb.c#L134-L137

At the moment, there is only one native USB dive computer, the Atomic Aquatics Cobalt, but adding another one would be a matter of adding a new entry.

For the prototype BLE implementation I'm working on, I'm using a similar table:

```
typedef struct dc_ble_uart_t {
	const char *service;
	struct {
		const char *rx;
		const char *tx;
		const char *rx_credits;
		const char *tx_credits;
	} characteristics;
} dc_ble_uart_t;

static const dc_ble_uart_t g_uarts[] = {
	// Telit/Stollmann (Heinrichs Weikamp)
	{"0000fefb-0000-1000-8000-00805f9b34fb", {
		"00000001-0000-1000-8000-008025000000",
		"00000002-0000-1000-8000-008025000000",
		"00000003-0000-1000-8000-008025000000",
		"00000004-0000-1000-8000-008025000000"}},
	// U-Blox (Heinrichs Weikamp)
	{"2456e1b9-26e2-8f83-e744-f34f01e9d701", {
		"2456e1b9-26e2-8f83-e744-f34f01e9d703",
		"2456e1b9-26e2-8f83-e744-f34f01e9d703",
		"2456e1b9-26e2-8f83-e744-f34f01e9d704",
		"2456e1b9-26e2-8f83-e744-f34f01e9d704"}},
	// Nordic Semiconductor (Deepblu, Oceans, Divesoft)
	{"6e400001-b5a3-f393-e0a9-e50e24dcca9e", {
		"6e400002-b5a3-f393-e0a9-e50e24dcca9e",
		"6e400003-b5a3-f393-e0a9-e50e24dcca9e",
		NULL, NULL}},
	// Microchip (Ratio, McLean)
	{"49535343-fe7d-4ae5-8fa9-9fafd205e455", {
		"49535343-8841-43f4-a8d4-ecbe34729bb3",
		"49535343-1e4d-4bd9-ba61-23c647249616",
		NULL, NULL}},
	// Shearwater
	{"fe25c237-0ece-443c-b0aa-e02033e7029d", {
		"27b7570b-359e-45a3-91bb-cf7e70049bd2",
		"27b7570b-359e-45a3-91bb-cf7e70049bd2",
		NULL, NULL}},
	// Mares
	{"544e326b-5b72-c6b0-1c46-41c1bc448118", {
		"99a91ebd-b21f-1689-bb43-681f1f55e966",
		"1d1aae28-d2a8-91a1-1242-9d2973fbe571",
		NULL, NULL}},
	// Suunto
	{"98ae7120-e62e-11e3-badd-0002a5d5c51b", {
		"c6339440-e62e-11e3-a5b3-0002a5d5c51b",
		"d0fd6b80-e62e-11e3-a2e9-0002a5d5c51b",
		NULL, NULL}},
	// ScubaPro
	{"fdcdeaaa-295d-470e-bf15-04217b7aa0a0", {
		"a188b7dd-debb-449a-852d-c243d46b4b1a",
		"aa0c68f0-ea9c-493d-8112-62879e72af68",
		NULL, NULL}},
	// Pelagic
	{"cb3c4555-d670-4670-bc20-b61dbc851e9a", {
		"6606ab42-89d5-4a00-a8ce-4eb5e1414ee0",
		"a60b8e5c-b267-44d7-9764-837caf96489e",
		NULL, NULL}},
	// Pelagic
	{"ca7b0001-f785-4c38-b599-c7c5fbadb034", {
		"ca7b0003-f785-4c38-b599-c7c5fbadb034",
		"ca7b0002-f785-4c38-b599-c7c5fbadb034",
		NULL, NULL}},
	// Deep Six
	{"f000ffe0-ab12-45ec-84c8-46483f4626e9", {
		"f000ffe1-ab12-45ec-84c8-46483f4626e9",
		"f000ffe1-ab12-45ec-84c8-46483f4626e9",
		NULL, NULL}},
	// Divesoft
	{"0000fcef-0000-1000-8000-00805f9b34fb", {
		"6e400002-b5a3-f393-e0a9-e50e24dcca9e",
		"6e400003-b5a3-f393-e0a9-e50e24dcca9e",
		NULL, NULL}},
	// Divesoft (16bit transitional)
	{"0000fcef-0000-1000-8000-00805f9b34fb", {
		"00000002-0000-1000-8000-00805f9b34fb",
		"00000003-0000-1000-8000-00805f9b34fb",
		NULL, NULL}},
	// Halcyon Symbios
	{"00000001-8c3b-4f2c-a59e-8c08224f3253", {
		"00000101-8c3b-4f2c-a59e-8c08224f3253",
		"00000201-8c3b-4f2c-a59e-8c08224f3253",
		NULL, NULL}},
};
```
Maybe this is what you're looking for? It's just a straightforward translation of the info in the BLE wiki page.

This is much simpler than the logic used by subsurface. It only uses known services and characteristics instead of trying to perform autodetection. Disadvantage of this approach is that adding a devices requires a change in the code, but the advantage is that it won't try to use the wrong service (except for the Cressi scenario where a device has another service that's also on the list).

PR #4464, opened 2025-03-07
Add support for Cressi BLE serial service

2opremio:
Add known Bluetooth LE serial service for the Cressi Bluetooth interfaces.

Unfortunately, the Cressi BT offers two known BLE serial services: the service I just added and a Nordic Semi UART (which I am assuming is somehow used by the BT interface itself).

The pre-existing code chose the service based on the discovery ordering (`BLEObject::addService()` cleared any prior services when finding a known one), which led to prioritizing the wrong service (the Nordic UART).

Now instead, all known services are kept until service discovery is completed and only then we choose a single known service based on its priority (determined by it's ordering in the `serial_service_uuids` array).

- [x] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

Fixes part of #4460 

### Release note:

Yes, added:

> core: add support for Cressi Bluetooth interfaces #4460

### Mentions:

@mikeller @jefdriesen

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -138,7 +138,7 @@ static const char *match_uuid_list(const QBluetoothUuid &match, const struct uui
 
 //
 // Known BLE GATT service UUID's that we should prefer for the serial
-// emulation.
+// emulation. Services listed first have higher precendence.
 //
 // The Bluetooth SIG is a disgrace, and never standardized any serial
 // communication over BLE. They should just have specified a standard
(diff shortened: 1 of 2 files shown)


core/qt-ble.cpp
```diff
@@ -154,6 +154,7 @@ static const char *match_uuid_list(const QBluetoothUuid &match, const struct uui
 // Oh. It did, didn't it?
 //
 static const struct uuid_match serial_service_uuids[] = {
+	{ "6e400001-b5a3-f393-e0a9-e50e24dc10b8", "Cressi"},
```
mikeller: I think we need a comment here to explain the situation, to allow future developers to understand that this has to be higher priority than Nordic Semi UART.

2opremio: Done. I also moved the Nordic UART to the end. PTAL

mikeller (approved):

kruegerha:
just to enlarge your sample:
Donatello:  4_04a1
Leonardo 2.0:  3_2f35

PR #4468, opened 2025-03-10
Add support for the Halcyon BLE service

jefdriesen:
The two Rx and Tx characteristics support the same operations (read, write and indicate). This confuses the logic to detect which characteristic should be used for Tx and Rx.

Fixed by adding an extra read/write exclude list.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -120,16 +120,22 @@ void BLEObject::writeCompleted(const QLowEnergyDescriptor&, const QByteArray&)
 	desc_written++;
 }
 
+#define ACCESS_NONE  0x00
+#define ACCESS_READ  0x01
+#define ACCESS_WRITE 0x02
+
 struct uuid_match {
 	const char *uuid, *details;
+	unsigned int access;
 };
 
-static const char *match_uuid_list(const QBluetoothUuid &match, const struct uuid_match *array)
+static const char *match_uuid_list(const QBluetoothUuid &match, const struct uuid_match *array, unsigned int access)
 {
 	const char *uuid;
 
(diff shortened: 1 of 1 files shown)


core/qt-ble.cpp
```diff
@@ -294,10 +295,21 @@ static const struct uuid_match skip_characteristics[] = {
 	{ NULL, }
 };
 
+static const struct uuid_match skip_characteristics_read[] = {
```
mikeller: This can probably be made a bit easier to read and maintain by combining the three `skip_...` lists into one and using a bitmask to encode which characteristic(s) it should be masked for.

jefdriesen: That could result in more complex code instead of a simplification. The `uuid_match` data structure and the `match_uuid_list` function are not only used for characteristics, but also used for services and those don't have read/write properties.

mikeller (changes requested):
Doing 'reuse by copy/paste' is an antipattern, so this should be cleaned up as per comment.

mikeller (approved):
This is looking better now, thank you.

The changes needed to support Halcyon devices do not seem to be in https://github.com/subsurface/libdc/tree/Subsurface-DS9 - so I guess this should go in when the next update of the `Subsurface-DS9` branch happens?

jefdriesen:
Yes, this will obviously also need the Halcyon changes from upstream libdivecomputer.

mikeller:
@jefdriesen: Makes sense. I'll merge this then when I am doing the next update of libdivecomputer from upstream.

PR #4469, opened 2025-03-10
Add discovery support for Cressi BT interfaces

### LINKED ISSUE #4460: Cannot sync Cressi Donatello through Bluetooth Interface

2opremio:
Based on a very limited sample count [1,2], the Cressi Bluetooth LE interfaces seem to advertise their computers with the following pattern: `<model number>_<4 digit lowercase hex serial number>`. We can use that pattern to identify Cressi computers.

[1] I own two Donatellos which advertise as 4_254d and 4_1e52.
[2] There is a Goa adverising as 2_466a https://github.com/deepsealabs/libdc-swift/issues/8#issuecomment-2571343354

It's worth noting that that the Goa and Donatello require different BT interface devices, so they seem to implement a consistent naming scheme.

A more reliable way would be to check for service UUID 6e400001-b5a3-f393-e0a9-e50e24dc10b8 on discovery and then read the device identification values. But this is a big deviation from the name-based matching btdiscovery.cpp is doing today.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Related issues:

Fixes #4460 

### Release note:

> core: add support for Cressi Bluetooth interfaces

### Mentions:

@mikeller @jefdriesen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-core: add support for Cressi Bluetooth interfaces #4460
+core: add support for Cressi Bluetooth interfaces #4460 #4469
 desktop: add support to import .asd and .script files from Scubapro's LogTrak and SmartTrak
 desktop: update "Save dive data as subtitles" feature to make it more configurable.
 bluetooth: fix crash on MacOS when doing first download from a new BLE device
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -40,6 +40,20 @@ static struct modelPattern model[] = {
 	{ 0x4744, "Aqualung", "i330R" },
 };
 
+// The Cressi BT interfaces for Goa-style computers advertise names with pattern:
+//  <model number>_<4 digit lowercase hex serial number>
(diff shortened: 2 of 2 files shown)


mikeller (approved):

Tobi1394:
I don’t know if this is important, but the Leonardo 2.0 has the model number 3. In my case, it’s 3_2f35. With the same interface that you showed in #4460.
I hope this helps in some way.

And by the way, thanks for your work here!

PR #4470, opened 2025-03-11
Use the libdivecomputer filter function

jefdriesen:
Replace the custom bluetooth device name detection code with the libdivecomputer filter function.

This PR is a request for comments because the libdivecomputer filter function isn't a 1:1 replacement for the existing code: 

The current subsurface code tries to identify the exact model where possible, while the libdivecomputer filter is much more broad because the same filter is shared by all models within the same device family. That shouldn't be a problem if the goal is to detect whether a bluetooth name matches a known dive computer. For example as done in the `matchesKnownDiveComputerNames()` function.

The proposed change also doesn't distinguish between BT Classic and LE. Since the BT device names are identical for both, that doesn't make a difference for the underlying matching. But not all devices are dual stack. Some are Classic only or BLE only. The result is that the `getDeviceType()` may select a Classic only or BLE only descriptor. If later on the descriptor is used to open the transport, and it happens to be of the wrong type (e.g. a BT Classic descriptor when trying to open a BLE connection or vice versa), the `divecomputer_device_open()` function will fail with `DC_STATUS_UNSUPPORTED`. The current subsurface code uses some heuristics to work around this problem, but the correct solution would be to pass the desired transport type such that only descriptors supporting the correct transport are selected. That will require some more refactoring.

The main advantage is that subsurface no longer needs to maintain its own logic. Whenever libdivecomputer is updated with support for new devices, the bluetooth discovery will also be updated to recognize those new devices.

Note: this change requires some of the recent commits on libdivecomputer master.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -17,7 +17,7 @@ namespace {
 }
 BTDiscovery *BTDiscovery::m_instance = NULL;
 
-static dc_descriptor_t *getDeviceType(QString btName)
+static dc_descriptor_t *getDeviceType(QString btName, int transports)
 // central function to convert a BT name to a Subsurface known vendor/model pair
 {
 	dc_status_t status = DC_STATUS_SUCCESS;
@@ -34,10 +34,17 @@ static dc_descriptor_t *getDeviceType(QString btName)
 
 	dc_descriptor_t *descriptor = NULL;
 	while ((status = dc_iterator_next (iterator, &descriptor)) == DC_STATUS_SUCCESS) {
-		if (dc_descriptor_filter(descriptor, DC_TRANSPORT_BLE, namestr) ||
-			dc_descriptor_filter(descriptor, DC_TRANSPORT_BLUETOOTH, namestr)) {
-			result = descriptor;
-			break;
+		if (transports & DC_TRANSPORT_BLE) {
+			if (dc_descriptor_filter(descriptor, DC_TRANSPORT_BLE, namestr)) {
+				result = descriptor;
+				break;
+			}
+		}
+		if (transports & DC_TRANSPORT_BLUETOOTH) {
+			if (dc_descriptor_filter(descriptor, DC_TRANSPORT_BLUETOOTH, namestr)) {
+				result = descriptor;
+				break;
+			}
 		}
 
 		dc_descriptor_free (descriptor);
@@ -55,7 +62,7 @@ static dc_descriptor_t *getDeviceType(QString btName)
 
 bool matchesKnownDiveComputerNames(QString btName)
 {
-	return getDeviceType(btName) != nullptr;
+	return getDeviceType(btName, DC_TRANSPORT_BLUETOOTH | DC_TRANSPORT_BLE) != nullptr;
 }
 
 BTDiscovery::BTDiscovery(QObject*) : m_btValid(false),
@@ -212,8 +219,11 @@ void BTDiscovery::btDeviceDiscoveredMain(const btPairedDevice &device, bool from
 {
 	btVendorProduct btVP;
 
+	dc_transport_t transport = device.address.startsWith("LE:") ?
+		DC_TRANSPORT_BLE : DC_TRANSPORT_BLUETOOTH;
+
 	QString newDevice;
-	dc_descriptor_t *newDC = getDeviceType(device.name);
+	dc_descriptor_t *newDC = getDeviceType(device.name, transport);
 	if (newDC)
 		newDevice = dc_descriptor_get_product(newDC);
 	else
--- a/core/btdiscovery.cpp
+++ b/core/btdiscovery.cpp
@@ -17,175 +17,40 @@ namespace {
 }
 BTDiscovery *BTDiscovery::m_instance = NULL;
 
-struct modelPattern {
-	uint16_t    model;
-	const char *vendor;
-	const char *product;
-};
-static struct modelPattern model[] = {
-	{ 0x4552, "Oceanic", "Pro Plus X" },
-	{ 0x455A, "Aqualung", "i750TC" },
-	{ 0x4647, "Sherwood", "Sage" },
-	{ 0x4648, "Aqualung", "i300C" },
-	{ 0x4649, "Aqualung", "i200C" },
-	{ 0x4749, "Aqualung", "i200Cv2" },
-	{ 0x4651, "Aqualung", "i770R" },
-	{ 0x4652, "Aqualung", "i550C" },
-	{ 0x4653, "Oceanic", "Geo 4.0" },
-	{ 0x4654, "Oceanic", "Veo 4.0" },
-	{ 0x4655, "Sherwood", "Wisdom 4" },
-	{ 0x4656, "Oceanic", "Pro Plus 4" },
-	{ 0x4741, "Apeks", "DSX" },
-	{ 0x4743, "Aqualung", "i470TC" },
-	{ 0x4744, "Aqualung", "i330R" },
-};
-
-// The Cressi BT interfaces for Goa-style computers advertise names with pattern:
-//  <model number>_<4 digit lowercase hex serial number>
-// The source of truth for model numbers is in libdivecomputer/src/descriptor.c
-static QRegularExpression cressiBTNamePattern("^([1-9][0-9]?)_[0-9a-f]{4}$");
-static QMap<int, QString> cressiModelNumToProduct{
-	{ 1, "Cartesio"},
-	{ 2, "Goa"},
-	{ 3, "Leonardo 2.0"},
-	{ 4, "Donatello"},
-        { 5, "Michelangelo"},
-	{ 9, "Neon"},
-	{10, "Nepto"}
-};
-
-struct namePattern {
-	const char *prefix;
-	const char *vendor;
-	const char *product;
-};
-// search is in order of this array, and as a prefix search, so more specific names
-// should be added before less specific names (i.e. "Perdix 2" before "Perdix")
-static struct namePattern name[] = {
-	// Shearwater dive computers
-	{ "Predator", "Shearwater", "Predator" },
-	{ "Perdix 2", "Shearwater", "Perdix 2"},
-	{ "Petrel 3", "Shearwater", "Petrel 3"},
-	// both the Petrel and Petrel 2 identify as "Petrel" as BT/BLE device
-	// but only the Petrel 2 is listed as available dive computer on iOS (which requires BLE support)
-	// so always pick the "Petrel 2" as product when seeing a Petrel
-	{ "Petrel", "Shearwater", "Petrel 2" },
-	{ "Perdix", "Shearwater", "Perdix" },
-	{ "Teric", "Shearwater", "Teric" },
-	{ "Peregrine", "Shearwater", "Peregrine" },
-	{ "NERD 2", "Shearwater", "NERD 2" },
-	{ "NERD", "Shearwater", "NERD" }, // order is important, test for the more specific one first
-	{ "Predator", "Shearwater", "Predator" },
-	{ "Tern", "Shearwater", "Tern" },
-	// Suunto dive computers
-	{ "EON Steel", "Suunto", "EON Steel" },
-	{ "EON Core", "Suunto", "EON Core" },
-	{ "Suunto D5", "Suunto", "D5" },
-	// Scubapro dive computers
-	{ "G2", "Scubapro", "G2" },
-	{ "HUD", "Scubapro", "G2 HUD" },
-	{ "G3", "Scubapro", "G3" },
-	{ "Aladin", "Scubapro", "Aladin Sport Matrix" },
-	{ "A1", "Scubapro", "Aladin A1" },
-	{ "A2", "Scubapro", "Aladin A2" },
-	{ "Luna 2.0 AI", "Scubapro", "Luna 2.0 AI" },
-	{ "Luna 2.0", "Scubapro", "Luna 2.0" },
-	// Mares dive computers
-	{ "Mares Genius", "Mares", "Genius" },
-	{ "Sirius", "Mares", "Sirius" },
-	{ "Mares", "Mares", "Quad" }, // we actually don't know and just pick a common one - user needs to fix in UI
-	// Cress dive computers
-	{ "CARESIO_", "Cressi", "Cartesio" },
-	{ "GOA_", "Cressi", "Goa" },
-	// Deepblu dive computesr
-	{ "COSMIQ", "Deepblu", "Cosmiq+" },
-	// Oceans dive computers
-	{ "S1", "Oceans", "S1" },
-	// McLean dive computers
-	{ "McLean Extreme", "McLean", "Extreme" },
-	// Tecdiving dive computers
-	{ "DiveComputer", "Tecdiving", "DiveComputer.eu" }
-};
-
 static dc_descriptor_t *getDeviceType(QString btName)
 // central function to convert a BT name to a Subsurface known vendor/model pair
 {
-	QString vendor, product;
+	dc_status_t status = DC_STATUS_SUCCESS;
+	dc_descriptor_t *result = NULL;
+	const QByteArray tmp = btName.toLocal8Bit();
+	const char *namestr = tmp.constData();
 
-	if (btName.startsWith("OSTC")) {
-		vendor = "Heinrichs Weikamp";
-		if (btName.mid(4,1) == "3") product = "OSTC Plus";
-		else if (btName.mid(4,2) == "s#") product = "OSTC Sport";
-		else if (btName.mid(4,2) == "s ") product = "OSTC Sport";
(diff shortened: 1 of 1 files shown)


mikeller:
I'm aways keen to reduce duplication of functionality - this will allow developers to focus their efforts on less code.

At the same time I think we need to retain the ability to specifically select Bluetooth classic or BLE for devices that support both - unfortunately our Bluetooth stack does not work reliably for both types on all the platforms that we support.

jefdriesen:
There are currently two places where the `getDeviceType() `function is used:

The first one is the `matchesKnownDiveComputerNames()` function. There, we are only interested in whether the bluetooth device name matches a known dive computer. Since we don't really care which dive computer (the returned descriptor isn't used at all), there is no change in functionality.

The second one is the `btDeviceDiscoveredMain()` function. Here things become more complex, because the returned descriptor is actually used.

At the moment, the code doesn't distinguish between classic and BLE. Take for example the Ratio dive computers. There are some classic only models (e.g. iX3M GPS series) and BLE only models (e.g. iX3M 2021 series and newer), but they all share the same bluetooth device name. Hence, the proposed code will always select the iX3M GPS Pro model because that happens to be the first bluetooth enabled model. To make a better choice we need to take into account the bluetooth type (classic vs BLE). This could be improved by looking at the bluetooth address. If the "LE" prefix is present, we can restrict the search to BLE enabled models, and if it's not present restrict the search to the BT classic enabled models. This isn't too difficult to implement.

The libdivecomputer filter is also much more broad compared to the current subsurface implementation. Take for example the Heinrichs Weikamp dive computers, with a bluetooth name "OSTC+ XXXX". The same filter function is used for all models (ostc, frog and ostc3 backends). So searching for a bluetooth classic model will find the Frog descriptor as the first match, and for the BLE model it's the OSTC 2. In the desktop version, the bluetooth name is shown with the MAC (e.g. "OSTC+ XXXX (XX:XX:XX:XX:XX:XX)"), but the mobile version shows the name from the descriptor plus the MAC (e.g. "Frog XX:XX:XX:XX:XX:XX)". Internally it wouldn't make much difference. Usually if you stay within the same family, the exact model doesn't really matter for setting up the connection (although there are also a few exceptions). In this particular case, the frog has its own backend which is different from the one used by the other hwos models, so depending on how the descriptor is being used, it could still matter here, but that's something that could be fixed. But for the user, displaying "Frog" might be more confusing here. Maybe it's already sufficient to also show the bluetooth name here, like in the desktop version, instead of showing the underlying descriptor name?

In theory we could make the libdivecomputer filter more strict, but that would break an important other use-case. In the scenario where the user first selects the dive computer model, and the filter function is used to only show bluetooth devices matching that descriptor. This is how it's used internally in libdivecomputer's built-in transports. Here you really want the less strict filter, such that when you select the Shearwater Perdix descriptor, you don't filter out the Perdix 2. Many dive computers have very similar names, which is often confusing for users, and the relaxed filter helps to avoid mistakes here.

mikeller:
> This could be improved by looking at the bluetooth address. If the "LE" prefix is present, we can restrict the search to BLE enabled models, and if it's not present restrict the search to the BT classic enabled models. This isn't too difficult to implement.

Yes, that probably makes sense. But at the same time, the only place where this is used for more than prefilling drop down boxes is on mobile, and I suspect that on mobile our Bluetooth support is currently limited to older devices, as we have to update from Qt5 to Qt6 to get a Bluetooth stack that works with the latest mobile OS.

> But for the user, displaying "Frog" might be more confusing here. Maybe it's already sufficient to also show the bluetooth name here, like in the desktop version, instead of showing the underlying descriptor name?

Not sure about that. Agree that showing the descriptor name will be better for cases like Heinrichs Weikamp, but then as far as I know there are also cases (in particular Bluetooth enabled adaptors for older dive computers with a serial interface) where the descriptor name is something unrecognisable. Not sure how the user experience would be for those.

> In theory we could make the libdivecomputer filter more strict, but that would break an important other use-case. In the scenario where the user first selects the dive computer model, and the filter function is used to only show bluetooth devices matching that descriptor. This is how it's used internally in libdivecomputer's built-in transports. Here you really want the less strict filter, such that when you select the Shearwater Perdix descriptor, you don't filter out the Perdix 2. Many dive computers have very similar names, which is often confusing for users, and the relaxed filter helps to avoid mistakes here.

I don't think that adding filtering of Bluetooth devices shown based on selected dive computer model would be an improvement in user experience - this should be the other way round, like it already is in Subsurface. In reality, in almost all cases there will only be 1 dive computer type Bluetooth device within range when a user wants to download dives - so it's a reasonable assumption that this is the user's dive computer.

jefdriesen:
> Yes, that probably makes sense. But at the same time, the only place where this is used for more than prefilling drop down boxes is on mobile, and I suspect that on mobile our Bluetooth support is currently limited to older devices, as we have to update from Qt5 to Qt6 to get a Bluetooth stack that works with the latest mobile OS.

I'll push an extra commit to address this.

Note that in the desktop version, it's also used to populate the device or mount point drop down. The presence of the LE prefix in the address decides whether to use BLE or BT classic. So it's a bit more than just filling in drop down boxes.

> Not sure about that. Agree that showing the descriptor name will be better for cases like Heinrichs Weikamp, but then as far as I know there are also cases (in particular Bluetooth enabled adaptors for older dive computers with a serial interface) where the descriptor name is something unrecognisable. Not sure how the user experience would be for those.

I'm not sure what you mean here. First of all, I doubt using a generic (usb)serial to bluetooth adapters is something that will work. Yes, there are dive computers that provide such a dongle (e.g. Mares Bluelink Pro), but this is a manufacturer specific dongle and it works the same as their native BLE enabled dive computers. That's not going to be the case for a generic dongle. For BT classic, it might work because there we have a standard protocol for serial communication over bluetooth, but not for BLE.  A generic dongle will likely also have a generic bluetooth name, and thus also won't be recognized as a dive computer. So in that case we can't match it to a dive computer at all. (I guess that's what your comment is about and with "descriptor name" you actually mean the "bluetooth name" and not the name from the libdivecomputer `dc_descriptor_t`.)

Note that the issue with the Heinrichs Weikamp (Frog vs OSTC) could be addressed by using a different filter function for the Frog and the other OSTC's in libdivecomputer. Same for the Shearwater dive computers (older Predator vs newer Petrel PNF models). Or maybe even try to merge those backends into just one because they are very similar anyway.

> I don't think that adding filtering of Bluetooth devices shown based on selected dive computer model would be an improvement in user experience - this should be the other way round, like it already is in Subsurface. In reality, in almost all cases there will only be 1 dive computer type Bluetooth device within range when a user wants to download dives - so it's a reasonable assumption that this is the user's dive computer.

Subsurface isn't the only application using libdivecomputer. Most applications let the user select the dive computer vendor/model first, and then offer the available serial/usb/bluetooth/etc connectivity options based on that choice.

Subsurface also uses this model. Subsurface indeed tries to detect the dive computer model based on the bluetooth name, but for the actual connection it uses the choice of the user. For example if the bluetooth scan discovers a Heinrichs Weikamp dive computer and the users selects a Shearwater dive computer, then subsurface will try to download using the shearwater protocol (which fails of course). The detected model is ignored at that point.

mikeller:
> Note that in the desktop version, it's also used to populate the device or mount point drop down. The presence of the LE prefix in the address decides whether to use BLE or BT classic. So it's a bit more than just filling in drop down boxes.

Frankly, the selection of Bluetooth 'classic' or BLE is one of the things that I find to be not very user friendly in the current Subsurface. What I am finding is that on a lot of operating systems, only one or the other is working reliably - but the one that is working well seems to be doing so consistently across different dive computers.
So, instead of forcing the user to pre-select the option that works separately for each dive computer, it might be better to give the user a "prefer 'classic'" / "prefer BLE" preference setting, and then respect this across all dive computers that they are downloading from.

> I'm not sure what you mean here. 

See https://github.com/subsurface/subsurface/issues/4460#issuecomment-2701856729 - '4_1e52' doesn't sound like something a user would be able to identify as a dive computer.

> Note that the issue with the Heinrichs Weikamp (Frog vs OSTC) could be addressed by using a different filter function for the Frog and the other OSTC's in libdivecomputer. Same for the Shearwater dive computers (older Predator vs newer Petrel PNF models). Or maybe even try to merge those backends into just one because they are very similar anyway.

What exactly would we achieve with this?

> Subsurface also uses this model. Subsurface indeed tries to detect the dive computer model based on the bluetooth name, but for the actual connection it uses the choice of the user. For example if the bluetooth scan discovers a Heinrichs Weikamp dive computer and the users selects a Shearwater dive computer, then subsurface will try to download using the shearwater protocol (which fails of course). The detected model is ignored at that point.

I am fine with that - if the user thinks they know better, or they know something we don't, why would we not allow them to proceed? Possibly the current behaviour could be improved by showing them a warning / confirmation box telling them about the difference between detected / selected dive computer model.

jefdriesen:
> Frankly, the selection of Bluetooth 'classic' or BLE is one of the things that I find to be not very user friendly in the current Subsurface. What I am finding is that on a lot of operating systems, only one or the other is working reliably - but the one that is working well seems to be doing so consistently across different dive computers. So, instead of forcing the user to pre-select the option that works separately for each dive computer, it might be better to give the user a "prefer 'classic'" / "prefer BLE" preference setting, and then respect this across all dive computers that they are downloading from.

Which BT variant works best also depends on the dive computer. I have some (dual stack) dive computers that work fine with BT classic and where BLE is less reliable. But for others, BLE works just fine on the same system. Of course most people only have one or maybe two dive computers, while I test with many different ones.

Another thing I noticed that is not always very intuitive is the fact that different transports have slightly different workflows. For (usb)serial, users need to select the serial port in the drop-down widget. For the USB and USB HID transports, the dropdown is disabled and the device is auto-detected (in theory there can be multiple connected, although in practice that will be rare). For the filesystem based transports, you need to use the small button next to the dropdown widget to select the directory. For bluetooth devices, you need to use "Choose bluetooth download mode" option at the bottom to bring up the bluetooth dialog. I've seen several people struggle with how to download from a bluetooth dive computer due to this.

There is also no indication which transports are available for a particular dive computer. Especially when some dive computers can supports multiple transports. Maybe another dropdown to select the transport would be useful here? The contents on the Device dropdown can then be populated depending on the chosen transport. The small button next to the dropdown widget can be used to shown the filesystem or bluetooth dialog.

Certainly an interesting discussion, but also not really on-topic here...

> See [#4460 (comment)](https://github.com/subsurface/subsurface/issues/4460#issuecomment-2701856729) - '4_1e52' doesn't sound like something a user would be able to identify as a dive computer.

Okay, now I see what you meant. There are indeed a few dive computers with weird bluetooth device names.

> > Note that the issue with the Heinrichs Weikamp (Frog vs OSTC) could be addressed by using a different filter function for the Frog and the other OSTC's in libdivecomputer. Same for the Shearwater dive computers (older Predator vs newer Petrel PNF models). Or maybe even try to merge those backends into just one because they are very similar anyway.
> 
> What exactly would we achieve with this?

Right now, the descriptors for both the frog and ostc family share the same `dc_filter_hw` function, which checks for both the "Frog" and "OSTC" prefixes. Thus when a bluetooth device named "OSTC+ XXXX" is discovered, it will match for both the frog and ostc descriptor. If you only consider the first match (as subsurface does), then it will find the frog descriptor because that happens to be the first one in the list. If you use the found descriptor only for the name, then the name might be a bit confusing, but if you would also use that descriptor to open the device, it will fail because you end up with the wrong backend.

Using two separate filter functions, one for the frog backend (checking only the "Frog" prefix) and another one for the ostc backend (checking only the "OSTC" prefix) does fix this problem.

> > Subsurface also uses this model. Subsurface indeed tries to detect the dive computer model based on the bluetooth name, but for the actual connection it uses the choice of the user. For example if the bluetooth scan discovers a Heinrichs Weikamp dive computer and the users selects a Shearwater dive computer, then subsurface will try to download using the shearwater protocol (which fails of course). The detected model is ignored at that point.
> 
> I am fine with that - if the user thinks they know better, or they know something we don't, why would we not allow them to proceed? Possibly the current behaviour could be improved by showing them a warning / confirmation box telling them about the difference between detected / selected dive computer model.

Isn't it a bit strange to first autodetect the dive computer model, and then not use this information? For example you first choose a Shearwater Perdix with the vendor/product dropdowns. Next, you open the bluetooth dialog and select an OSTC from the list of paired/discovered bluetooth devices. At that point subsurface knows this is an OSTC, but if you try to connect right away, subsurface will try the shearwater protocol. Wouldn't it make more sense to update the vendor/product dropdowns and select the detected OSTC? Of course the user can still select a Shearwater Perdix if he/she really wants to try that.

mikeller:
> Another thing I noticed that is not always very intuitive is the fact that different transports have slightly different workflows. For (usb)serial, users need to select the serial port in the drop-down widget. For the USB and USB HID transports, the dropdown is disabled and the device is auto-detected (in theory there can be multiple connected, although in practice that will be rare). For the filesystem based transports, you need to use the small button next to the dropdown widget to select the directory. For bluetooth devices, you need to use "Choose bluetooth download mode" option at the bottom to bring up the bluetooth dialog.

Yes, the 'Import from dive computer' dialog is still messy in the way it is now.

> I've seen several people struggle with how to download from a bluetooth dive computer due to this.

#4511 should have improved this - but then of course there are the dive computers that support Bluetooth AND another transport, for which this isn't helping.

> For example you first choose a Shearwater Perdix with the vendor/product dropdowns. Next, you open the bluetooth dialog and select an OSTC from the list of paired/discovered bluetooth devices. At that point subsurface knows this is an OSTC, but if you try to connect right away, subsurface will try the shearwater protocol. Wouldn't it make more sense to update the vendor/product dropdowns and select the detected OSTC? Of course the user can still select a Shearwater Perdix if he/she really wants to try that.

Can you describe a use case when any reasonable user would want to use these steps?

dirkhh:
also, please note that the BLE workflow will always be different depending on OS because the different OSs have opposite requirements for an app interacting with them.
macOS requires that you scan, Windows requires that you don't scan but instead pretend that reading its pre-populated list is the same thing as actually scanning. And Linux allows you to scan or use a remembered value which will be valid if you simply try using again

mikeller (commented):
Quick testing shows that this seems to be working on linux, Android, macOS.

But there are a lot of cases (like https://github.com/subsurface/subsurface/pull/4470/changes#diff-f713938353f214f3c7d263ba383295457395d00ea88a59eb0d0452e59f144e80R29) where the coding style and established pattern are violated by having space in between a function name and the parameter brackets - can this please be changed to get a consistent coding style?

jefdriesen:
> > I've seen several people struggle with how to download from a bluetooth dive computer due to this.
> 
> #4511 should have improved this - but then of course there are the dive computers that support Bluetooth AND another transport, for which this isn't helping.

Adding another dropdown box to choose the transport type might help here. Only list the options that are available for the selected dive computers and disable the widget if there is only one option. For the user it will be more obvious there are multiple options available and they can be selected in a central place.

> > For example you first choose a Shearwater Perdix with the vendor/product dropdowns. Next, you open the bluetooth dialog and select an OSTC from the list of paired/discovered bluetooth devices. At that point subsurface knows this is an OSTC, but if you try to connect right away, subsurface will try the shearwater protocol. Wouldn't it make more sense to update the vendor/product dropdowns and select the detected OSTC? Of course the user can still select a Shearwater Perdix if he/she really wants to try that.
> 
> Can you describe a use case when any reasonable user would want to use these steps?

Of course most users won't do that on purpose, but mistakes do happen. Some dive computers have very confusing names and users sometimes don't know exactly which one they have. For example there are 7 variants of the Oceanic Veo and they are not all using the same protocol. Same for the Aqualung models with a very similar name (i330R, i300, i300C, etc). For some dive computers the bluetooth name is also not exactly human recognizable as a dive computer (e.g. oceanic, cressi, halcyon) and users try whatever they find.

> But there are a lot of cases (like https://github.com/subsurface/subsurface/pull/4470/changes#diff-f713938353f214f3c7d263ba383295457395d00ea88a59eb0d0452e59f144e80R29) where the coding style and established pattern are violated by having space in between a function name and the parameter brackets - can this please be changed to get a consistent coding style?

Sure, no problem.

jefdriesen:
> Windows requires that you don't scan but instead pretend that reading its pre-populated list is the same thing as actually scanning.

That's not entirely correct because Windows has multiple bluetooth apis.

With the win32 api you can indeed not scan for bluetooth devices. You can only enumerate previously paired devices. That's why users must first pair the device using the windows control panel. This enumeration will also return devices that are not in range and thus trying to connect to them will of course fail. But there is also a more modern winrt api available that is capable of scanning, pairing, etc.

Qt5 has backends for both, but in the builds subsurface is using, only the win32 one is enabled. I think that's because the winrt one is difficult/impossible to build with mingw. In Qt6 the win32 backend has been removed:

https://doc.qt.io/qt-6/qtbluetooth-changes-qt6.html

PR #4472, opened 2025-03-15
Export a json file in addition to the existing JS file

### LINKED ISSUE #4471: export the trips in proper json instead of a JS file
Hi,
I'd like to be able to download a json file instead of a JS file. 
Currently I'm using the file `file.js` that I can download from https://cloud.subsurface-divelog.org/user/felash@gmail.com/dives.html_files/file.js, but it's not super great because I have to insert it using a `<script>` tag, and I don't always get the latest version after a cloud push (or maybe it takes some time to update it?).

Instead I'd like to generate a json file, with proper CORS headers too, so that it could be `fetch`-ed.

My current work is at https://julienw.github.io/subsurface-webviewer/ (source: https://github.com/julienw/subsurface-webviewer/), and the logic to download the trips is: https://github.com/julienw/subsurface-webviewer/blob/f94f6eb96d2f4d6ce94a1ea5bcb084ebb9916132/src/store/dataSlice.ts#L33-L47

I found the executable export-html.cpp with all its pieces, and could run it locally too, but I'm not sure how this is used for the server-side.

I couldn't find where the code handling the cloud server is stored, do you have pointers? I'd love to be able to contribute to it if that's possible.

Thanks!

julienw:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

To make it easier to reuse the data stored in the cloud storage, a JSON file is now generated in addition to all the other files. On purpose all existing behavior has been kept so that this change isn't a breaking change.

We'd still need CORS headers to be added on the cloud server so that web pages could download it. But I'm not sure how to do that.

### Changes made:

1) Minor refactor
2) Export trips as a JSON file called trips.json

### Related issues:

Fixes #4471

### Additional information:

After building the export-html binary with:
```
make -C subsurface/build export-html
```
I'm using this command to generate the export using my own data:
```
./subsurface/build/export-html --source file:///home/julien/.subsurface/cloudstorage/45290a30c2e8dd22/[felash@gmail.com] --output /home/julien/perso/git/subsurface-root/output/dives.html
```

Of course you need to tweak the command with your own data.

### Release note:

A changelog entry has been added, it's up to you to decide about adding this as a release note.

### Documentation change:

Probably not, but maybe it could be added in https://subsurface-divelog.org/subsurface-user-manual/#S_Cloud_storage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 statistics: add month-of-year binning
+htmlexport: export dives as a json file as well
 core: add support for Cressi Bluetooth interfaces #4460 #4469
 desktop: add support to import .asd and .script files from Scubapro's LogTrak and SmartTrak
 desktop: update "Save dive data as subtitles" feature to make it more configurable.
--- a/core/divelogexportlogic.cpp
+++ b/core/divelogexportlogic.cpp
@@ -129,6 +129,7 @@ void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes)
 	mainDir.mkdir(exportFiles);
 
 	QString js_dive_data = exportFiles + "file.js";
+	QString json_dive_data = exportFiles + "trips.json";
 	QString json_settings = exportFiles + "settings.js";
 	QString translation = exportFiles + "translation.js";
 	QString stat_file = exportFiles + "stat.js";
@@ -143,6 +144,7 @@ void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes)
 	export_translation(qPrintable(translation));
 
 	export_JS(qPrintable(js_dive_data), qPrintable(photosDirectory), hes.selectedOnly, hes.listOnly);
+	export_JSON(qPrintable(json_dive_data), hes.selectedOnly, hes.listOnly);
 
 	QString searchPath = getSubsurfaceDataPath("theme");
 	if (searchPath.isEmpty()) {
--- a/core/divelogexportlogic.cpp
+++ b/core/divelogexportlogic.cpp
@@ -128,7 +128,7 @@ void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes)
 	QString exportFiles = file.fileName() + "_files" + QDir::separator();
 	mainDir.mkdir(exportFiles);
 
-	QString json_dive_data = exportFiles + "file.js";
+	QString js_dive_data = exportFiles + "file.js";
 	QString json_settings = exportFiles + "settings.js";
 	QString translation = exportFiles + "translation.js";
 	QString stat_file = exportFiles + "stat.js";
@@ -142,7 +142,7 @@ void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes)
 	exportHTMLstatistics(stat_file, hes);
 	export_translation(qPrintable(translation));
 
-	export_HTML(qPrintable(json_dive_data), qPrintable(photosDirectory), hes.selectedOnly, hes.listOnly);
+	export_JS(qPrintable(js_dive_data), qPrintable(photosDirectory), hes.selectedOnly, hes.listOnly);
 
 	QString searchPath = getSubsurfaceDataPath("theme");
 	if (searchPath.isEmpty()) {
--- a/core/save-html.cpp
+++ b/core/save-html.cpp
@@ -480,6 +480,27 @@ void export_JS(const char *file_name, const char *photos_dir, const bool selecte
 	}
 }
 
+void export_list_as_JSON(struct membuffer *b, const bool selected_only, const bool list_only)
+{
+	write_trips(b, /* photos_dir */ nullptr, selected_only, list_only);
+}
+
+void export_JSON(const char *file_name, const bool selected_only, const bool list_only)
+{
+	FILE *f;
+
+	membuffer buf;
+	export_list_as_JSON(&buf, selected_only, list_only);
+
+	f = subsurface_fopen(file_name, "w+");
+	if (!f) {
+		report_error(translate("gettextFromC", "Can't open file %s"), file_name);
+	} else {
+		flush_buffer(&buf, f); /*check for writing errors? */
+		fclose(f);
+	}
+}
+
 void export_translation(const char *file_name)
 {
 	FILE *f;
--- a/core/save-html.cpp
+++ b/core/save-html.cpp
@@ -437,6 +437,7 @@ static void write_trips(struct membuffer *b, const char *photos_dir, bool select
 	char sep_ = ' ';
 	char *sep = &sep_;
 
+	put_string(b, "[");
 	for (auto &trip: divelog.trips)
 		trip->saved = 0;
 
@@ -454,21 +455,21 @@ static void write_trips(struct membuffer *b, const char *photos_dir, bool select
 
 	/*Save all remaining trips into Others*/
 	write_no_trip(b, &dive_no, selected_only, photos_dir, list_only, sep);
-}
-
-void export_list(struct membuffer *b, const char *photos_dir, bool selected_only, const bool list_only)
-{
-	put_string(b, "trips=[");
-	write_trips(b, photos_dir, selected_only, list_only);
 	put_string(b, "]");
 }
 
-void export_HTML(const char *file_name, const char *photos_dir, const bool selected_only, const bool list_only)
+void export_list_as_JS(struct membuffer *b, const char *photos_dir, bool selected_only, const bool list_only)
+{
+	put_string(b, "trips=");
+	write_trips(b, photos_dir, selected_only, list_only);
+}
+
+void export_JS(const char *file_name, const char *photos_dir, const bool selected_only, const bool list_only)
 {
 	FILE *f;
 
 	membuffer buf;
-	export_list(&buf, photos_dir, selected_only, list_only);
+	export_list_as_JS(&buf, photos_dir, selected_only, list_only);
 
 	f = subsurface_fopen(file_name, "w+");
 	if (!f) {
--- a/core/save-html.h
+++ b/core/save-html.h
@@ -19,6 +19,8 @@ void put_HTML_volume_units(struct membuffer *b, unsigned int ml, const char *pre
 
 void export_JS(const char *file_name, const char *photos_dir, const bool selected_only, const bool list_only);
 void export_list_as_JS(struct membuffer *b, const char *photos_dir, bool selected_only, const bool list_only);
+void export_JSON(const char *file_name, const bool selected_only, const bool list_only);
+void export_list_as_JSON(struct membuffer *b, const bool selected_only, const bool list_only);
 void export_translation(const char *file_name);
 
 #endif
--- a/core/save-html.h
+++ b/core/save-html.h
@@ -17,8 +17,8 @@ void put_HTML_pressure_units(struct membuffer *b, pressure_t pressure, const cha
 void put_HTML_weight_units(struct membuffer *b, unsigned int grams, const char *pre, const char *post);
 void put_HTML_volume_units(struct membuffer *b, unsigned int ml, const char *pre, const char *post);
 
-void export_HTML(const char *file_name, const char *photos_dir, const bool selected_only, const bool list_only);
-void export_list(struct membuffer *b, const char *photos_dir, bool selected_only, const bool list_only);
+void export_JS(const char *file_name, const char *photos_dir, const bool selected_only, const bool list_only);
+void export_list_as_JS(struct membuffer *b, const char *photos_dir, bool selected_only, const bool list_only);
 void export_translation(const char *file_name);
 
 #endif
--- a/core/uploadDiveShare.cpp
+++ b/core/uploadDiveShare.cpp
@@ -28,7 +28,7 @@ void uploadDiveShare::doUpload(bool selected, const QString &uid, bool noPublic)
 {
 	//generate json
 	membuffer buf;
-	export_list(&buf, NULL, selected, false);
+	export_list_as_JS(&buf, NULL, selected, false);
 	QByteArray json_data(buf.buffer, buf.len);
 
 	//Request to server
--- a/theme/dive_export.html
+++ b/theme/dive_export.html
@@ -5,6 +5,11 @@
 <script>
 //////////////////////////////////
 //advance settings window//
+
+// Because fetching the JSON file is async, we need a way to wait that it's
+// fetched and parsed. This promise will be assigned in load_script below, and
+// awaited in the onload function.
+let waitForTripsLoadPromise;
 load_scripts();
 
 function load_script_sync(name)
@@ -16,8 +21,15 @@ function load_script_sync(name)
 	document.getElementsByTagName("head")[0].appendChild(fileref);
 }
 
+async function loadTrips() {
+	const fetchResponse = await fetch(location.pathname + "_files/trips.json");
+	const trips = await fetchResponse.json();
+	window.trips = trips;
+}
+
 function load_scripts()
 {
+	waitForTripsLoadPromise = loadTrips();
 	var fileref=document.createElement("link");
 	fileref.setAttribute("rel", "stylesheet");
 	fileref.setAttribute("type", "text/css");
@@ -94,7 +106,9 @@ function loadSettings(){
 
 var searchingModules = new Array();
 
-window.onload=function(){
+window.onload = async function(){
+	// Wait that the trips JSON file is loaded before going on.
+	await waitForTripsLoadPromise;
 
 	//initialize settings
 	loadSettings();


mikeller:
Can you please provide some more information on what your intended use for this change is?

julienw:
Briefly, I'm working on https://github.com/julienw/subsurface-webviewer/, mostly for my own usage (showing a graph of ascent and descent speeds, things like that).

https://github.com/subsurface/subsurface/issues/4471 has some more context about that.

mikeller:
Ok, in this case it is preferred that you make the change in https://github.com/julienw/subsurface/ and not propose it as a pull request for the main repository.

julienw:
I don't understand: I'd like to have this on the official cloud server. Moreover I think it would be useful for more people than just me.

mikeller:
Ah, now I get it - what you want is for the Subsurface cloud service to provide a public API that you can consume from your own project.

The intended purpose of the Subsurface cloud service is to provide users with a way to synchronise data between multiple instances of Subsurface running on different devices. The web UI is really just leveraging what we already have in place, and a convenient way for users to self-debug synchronisation issues.

I will leave it up to @dirkhh who is taking care of operations and funding for the cloud infrastructure to decide if we want to open this up by providing APIs to extract data, and how we want to control access to it (e.g CORS).

Alternatively, since this is for personal use, it should be relatively easy to apply this patch to your own fork, and then have your cloud server do a git clone of your dive data repository and run the modified code against it to get the desired output.

julienw:
> Ah, now I get it - what you want is for the Subsurface cloud service to provide a public API that you can consume from your own project.

Yes that's it !

> Alternatively, since this is for personal use, it should be relatively easy to apply this patch to your own fork, and then have your cloud server do a git clone of your dive data repository and run the modified code against it to get the desired output.

I've at least one friend using this with his own data.

Longer term I'm seeing this as a possible alternative UI for consuming the data.

But I understand that there are scaling questions to be discussed.

julienw:
Also I can see other users consuming their dive data using other languages than JavaScript, to do whatever they want with it, and the current setup doesn't allow this as well. Although this might be possible with the local XML file, I haven't researched this much.

dirkhh:
Running the cloud backend is quite a thankless and frustrating task, I'm not sure that I want to add to my workload with providing an API. 
The local XML contains absolutely everything and should allow you to do whatever you want with your own data.

julienw:
I totally understand that this isn't a fun task.
I believe that the patch here is the only API we need, provided that the cors headers are present.
The local XML doesn't allow to develop a Web UI unfortunately. 
The use case is that I can already show the data to my students after a dive. I use the mobile version of subsurface to import the data from my computer, then I can see the analyzed data from the web UI.

julienw:
I guess only having the CORS headers without this patch could work as well. This would make it possible to use the fetch API instead of inserting a script.

mikeller:
@julienw:

I think a basic difference in understanding here is that Subsurface is not a web service that happens to have a client app. It is a multi-platform client app that happens to have its own cloud synchronisation service. And as such, efforts to improve Subsurface and extend its functionality should focus on the client, where all of its users will be able to benefit, and not on the server.

> Longer term I'm seeing this as a possible alternative UI for consuming the data.

To be honest this is the only way that I can see this going forward, as it will not have a negative impact on maintenance effort and cost for the Subsurface web infrastructure.

> Also I can see other users consuming their dive data using other languages than JavaScript

See above, the primary and supported way of exporting data from Subsurface is through the XML generated by the Subsurface application. This has the advantage that the XML can be transformed into practically anything with XSLT (e.g. https://github.com/bramstein/xsltjson).

> I believe that the patch here is the only API we need, provided that the cors headers are present.

If this was merged, deployed to the Subsurface web infrastructure, and then used to generate and store additional .json files every time a Subsurface user pushed an updated log file, would you be willing to pay on an ongoing basis for the additional compute and storage cost that this generated?

> The local XML doesn't allow to develop a Web UI unfortunately. 

I disagree - there are plenty of web apps consuming XML data (by using libraries like https://www.npmjs.com/package/xml2js).

> The use case is that I can already show the data to my students after a dive. I use the mobile version of subsurface to import the data from my computer, then I can see the analyzed data from the web UI.

I don't really see why this has to pull data from the cloud - this could probably just run on your local machine and analyse the data after synchronising with the cloud.

I am aware that there currently isn't an option to run Subsurface from the command line, get it to pull (or open) the cloud storage, and then export the data into an XML file - maybe this is an option that would help you with achieving what you want that we could look into adding?

mikeller:
> I guess only having the CORS headers without this patch could work as well. This would make it possible to use the fetch API instead of inserting a script.

I can't see any reason not to allow third party web apps to consume Subsurface dive logs, and a low risk of malicious websites 'impersonanting' Subsurface. In the end the user will still need to be authenticated with us with Basic Auth.

On the flip side I am not sure this is even possible in a reasonable way: My reading of https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin#directives suggests that the use of `*` (allow all) together with Basic Auth will result in an error thrown by the user's browser.

julienw:
> @julienw:
> 
> I think a basic difference in understanding here is that Subsurface is not a web service that happens to have a client app. It is a multi-platform client app that happens to have its own cloud synchronisation service. And as such, efforts to improve Subsurface and extend its functionality should focus on the client, where all of its users will be able to benefit, and not on the server.
> 
> > Longer term I'm seeing this as a possible alternative UI for consuming the data.
> 
> To be honest this is the only way that I can see this going forward, as it will not have a negative impact on maintenance effort and cost for the Subsurface web infrastructure.
> 
> > Also I can see other users consuming their dive data using other languages than JavaScript
> 
> See above, the primary and supported way of exporting data from Subsurface is through the XML generated by the Subsurface application. This has the advantage that the XML can be transformed into practically anything with XSLT (e.g. https://github.com/bramstein/xsltjson).
> 
> > I believe that the patch here is the only API we need, provided that the cors headers are present.
> 
> If this was merged, deployed to the Subsurface web infrastructure, and then used to generate and store additional .json files every time a Subsurface user pushed an updated log file, would you be willing to pay on an ongoing basis for the additional compute and storage cost that this generated?

What additional cost are we talking about here?

If we can accept that `file.js` isn't present anymore, I believe I can have a patch that would replace `file.js` with `trips.json`. Then no more cost.

> 
> > The local XML doesn't allow to develop a Web UI unfortunately.
> 
> I disagree - there are plenty of web apps consuming XML data (by using libraries like https://www.npmjs.com/package/xml2js).

That's not my point. My point is that the local XML is... local. If it was generated on the cloud server this would be fine enough by me (provided it could be fetch-ed).

> 
> > The use case is that I can already show the data to my students after a dive. I use the mobile version of subsurface to import the data from my computer, then I can see the analyzed data from the web UI.
> 
> I don't really see why this has to pull data from the cloud - this could probably just run on your local machine and analyse the data after synchronising with the cloud.

That's the point: when doing an exercise dive there's no local machine besides the mobile phone. I find the mobile app not good enough at this point (hopefully this might change, but I can't help about that, as I don't know much about mobile programming).

> 
> I am aware that there currently isn't an option to run Subsurface from the command line, get it to pull (or open) the cloud storage, and then export the data into an XML file - maybe this is an option that would help you with achieving what you want that we could look into adding?

My ideal way of working would be to provide that as an API.

> > I guess only having the CORS headers without this patch could work as well. This would make it possible to use the fetch API instead of inserting a script.
> 
> I can't see any reason not to allow third party web apps to consume Subsurface dive logs, and a low risk of malicious websites 'impersonanting' Subsurface. In the end the user will still need to be authenticated with us with Basic Auth.
> 
> On the flip side I am not sure this is even possible in a reasonable way: My reading of https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin#directives suggests that the use of `*` (allow all) together with Basic Auth will result in an error thrown by the user's browser.

Ah that's right, I think we'd need to mirror the request's `Origin` header instead of using `*`. But that's not all.
From my basic testing, we also need that the server can answer the preflight "OPTIONS" request... without authentication.
This answer needs to have `Access-Control-Request-Headers` having `authorization` in it (because we'd have to specify it manually in the request).

I understand that this is more work than what we'd like to do at the moment on the server side. It's not "just adding CORS headers".


In the mean time I added some cache busting parameters in my current way of working. I'm hopeful that this might be enough for my use case. Longer term I might need as server side component (esp if I want to be able to do modifications), but then I could just use the git protocol directly -- that's not for today though.

Feel free to close the PR if you think that's not a useful addition, I won't be sad :-)

mikeller:
> What additional cost are we talking about here?

@dirkhh would have the details of this - but from my understanding this would essentially be a doubling of the cost for the web interface / API because we'd double up on generation / storage of data.

> If we can accept that `file.js` isn't present anymore, I believe I can have a patch that would replace `file.js` with `trips.json`. Then no more cost.

Agreed. And the effort to convert the current web UI from consuming a .js to a .json would likely be relatively minimal.

> > I disagree - there are plenty of web apps consuming XML data (by using libraries like [npmjs.com/package/xml2js](https://www.npmjs.com/package/xml2js)).
> 
> That's not my point. My point is that the local XML is... local. If it was generated on the cloud server this would be fine enough by me (provided it could be fetch-ed).

Goes back to my first point - the focus of Subsurface is local. And I honestly can't see this changing anytime in the near or mid-term future, for a number of reasons.

> That's the point: when doing an exercise dive there's no local machine besides the mobile phone. I find the mobile app not good enough at this point (hopefully this might change, but I can't help about that, as I don't know much about mobile programming).

Ok, I get that. Qt on mobile isn't a great environment to do number crunching (or anything). But building something that uses a locally built / accessible .xml file might be an alternative to taking the detour of getting a .json from a servcer that is likely to be only a subset of the data in the .xml.

> > I can't see any reason not to allow third party web apps to consume Subsurface dive logs, and a low risk of malicious websites 'impersonanting' Subsurface. In the end the user will still need to be authenticated with us with Basic Auth.
> > On the flip side I am not sure this is even possible in a reasonable way: My reading of [developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin#directives](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin#directives) suggests that the use of `*` (allow all) together with Basic Auth will result in an error thrown by the user's browser.
> 
> Ah that's right, I think we'd need to mirror the request's `Origin` header instead of using `*`. But that's not all. From my basic testing, we also need that the server can answer the preflight "OPTIONS" request... without authentication. This answer needs to have `Access-Control-Request-Headers` having `authorization` in it (because we'd have to specify it manually in the request).

Yeah. From my experience with CORS this will end up in a rather ugly 'bending out of shape to make the browser happy' exercise, as CORS is the anathema of 'data exchange'. The simplest / easiest way for you to get what you want will probably be to set up your own server on a free cloud tier, forward authentication, and add all the 'garnishes' that the client's browser wants in order to allow consumption of the data.

> Feel free to close the PR if you think that's not a useful addition, I won't be sad :-)

Happy to leave it open for now, it's a good discussion. :-)

dirkhh:
The local vs. webservice argument is interesting to me.
I developed the backend storage (and a good chunk, but not all of the HTML rendering code) as a means to a very specific end. In the early 2010s people were starting to think of mobile phones as their main device. Which to me appeared insane, but here we are.
So the first step was the idea to have a way to just show a web site with the dives on your phone. Then to have a very simple "display only" mobile app that used cloud storage to get to your data. And only then did the mobile app turn into something that actually allowed editing and downloading and all that.

So with this in mind, and with past attacks on the cloud servers that caused me significant cost (because a person who was pissed off at the project created several TB of traffic that I ended up paying for - and not at today's cloud cost), I tend to always be very hesitant about uses of the cloud backend for anything but the primary purpose of connecting Subsurface and Subsurface-mobile.

But reading through this thread - the added cost of providing a `trips.json` file (again, at today's cost structure) really won't make a big difference. And today (different from ten years ago) there is a very generous community of users making small donations to fund the cloud infrastructure, so I'm ok with adding this.

The issues with the CORS headers is more concerning as those can be a major pain to get working. I'm not sure I fully understand what the requirement is to get this implemented here, so I'd love some more detail. Responding to an unauthenticated OPTIONS request doesn't seem too crazy (I do have some reasonably functional flood-protection in place on these servers). But the details are a bit unclear to me right now, so please help me understand what needs to be done here.

julienw:
wow, time flies, sorry for the delay in answering this.

> Goes back to my first point - the focus of Subsurface is local. And I honestly can't see this changing anytime in the near or mid-term future, for a number of reasons.

This is my vision: you keep the focus where you want it to be, but let the community build on top of it.
Open data in addition to open source, if you will.



> The local vs. webservice argument is interesting to me. I developed the backend storage (and a good chunk, but not all of the HTML rendering code) as a means to a very specific end. In the early 2010s people were starting to think of mobile phones as their main device. Which to me appeared insane, but here we are. So the first step was the idea to have a way to just show a web site with the dives on your phone. Then to have a very simple "display only" mobile app that used cloud storage to get to your data. And only then did the mobile app turn into something that actually allowed editing and downloading and all that.
> 
> So with this in mind, and with past attacks on the cloud servers that caused me significant cost (because a person who was pissed off at the project created several TB of traffic that I ended up paying for - and not at today's cloud cost), I tend to always be very hesitant about uses of the cloud backend for anything but the primary purpose of connecting Subsurface and Subsurface-mobile.

Thanks for the background!

> 
> But reading through this thread - the added cost of providing a `trips.json` file (again, at today's cost structure) really won't make a big difference. And today (different from ten years ago) there is a very generous community of users making small donations to fund the cloud infrastructure, so I'm ok with adding this.
> 
> The issues with the CORS headers is more concerning as those can be a major pain to get working. I'm not sure I fully understand what the requirement is to get this implemented here, so I'd love some more detail. Responding to an unauthenticated OPTIONS request doesn't seem too crazy (I do have some reasonably functional flood-protection in place on these servers). But the details are a bit unclear to me right now, so please help me understand what needs to be done here.

When doing a `fetch` with this command:
```js
fetch("http://localhost:8080/", { // replace the URL with the real URL
  headers: {
    Authorization: "Basic XXX", // To be computed in JS
  },
});
```

This means we need to send the header `Authorization`, then the `OPTIONS` request that's sent first contains these headers:
```
Access-Control-Request-Method: GET
Access-Control-Request-Headers: authorization
Origin: http://<server name>
```
Of course the `Origin` depends on the website.

Then the answer could be a 204 with these headers:
```
Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: origin,x-requested-with,content-type,accept,range,authorization
Access-Control-Allow-Methods: GET
```

Then the actual `GET` request is sent with just this header:
```
Origin: http://<server name>
```

The answer needs to have these headers:
```
Access-Control-Allow-Origin: *
access-control-allow-headers: Origin, X-Requested-With, Content-Type, Accept, Range, Authorization
```

I initially thought that `Access-Control-Allow-Origin` needs to mirror the request's `Origin` but from my testing it doesn't seem necessary here, because the credentials is sent explicitely without the `credentials` flag (eg a Cookie). This makes things easier because then it can all be static configuration.

I couldn't find where the code for the backend is, do you have it somewhere? The various scripts, the server configuration, etc.

Hope this helps!

julienw:
Note I tested locally using a local server run with `npx http-server --cors=Authorization` and a local HTML file containing just the code above. I took care to access the server with "127.0.0.1:8080" so that the origin would be different than the one in the fetch.

mikeller:
> This is my vision: you keep the focus where you want it to be, but let the community build on top of it. Open data in addition to open source, if you will.

Not sure if I would call it 'on top of it' if we enable the community to build on something that is by intent a second class citizen, and likely not to get updates and improvements that are made to the standalone app, or only get them with a lot of delay.
We'll probably have to make a clear statement to users of this API that there is no expectation that this will ever reach feature parity with the standalone app.

But I'll leave @dirkhh to provide feedback on his motivation and availability to work on adding custom headers.

dirkhh:
Hmm. I tried to enable this on the backend. Let me know if this is sufficient.

julienw:
> Hmm. I tried to enable this on the backend. Let me know if this is sufficient.

The OPTIONS request seem to be still password-protected, I get a 401 on it.

Here is the full request sent by Firefox copied as a Curl command, so that you can try it locally:
```
curl -i 'https://cloud.subsurface-divelog.org/user/USE_YOUR_LOGIN/dives.html_files/file.js?_uncache=1745058198344' -X OPTIONS -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:137.0) Gecko/20100101 Firefox/137.0' -H 'Accept: */*' -H 'Accept-Language: fr-FR,fr;q=0.8,en-US;q=0.5,en;q=0.3' -H 'Accept-Encoding: gzip, deflate, br, zstd' -H 'Access-Control-Request-Method: GET' -H 'Access-Control-Request-Headers: authorization' -H 'Referer: http://localhost:5173/' -H 'Origin: http://localhost:5173' -H 'DNT: 1' -H 'Sec-GPC: 1' -H 'Connection: keep-alive' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: cross-site' -H 'Priority: u=4'
```
(replace `USE_YOUR_LOGIN` with your subsurface login, but that's probably not needed to try it out?)

This command needs to be available without authorization.

julienw:
I wanted to mention one more thing: with the CORS headers configured, I don't really need this PR landed. I can work with the existing `file.js` by simply removing the prefix.
Getting the CORS headers right is the most important piece, and will make it much easier to get the data without inserting the script as a script, which wasn't great from a security standpoint.

I also want to highlight more thing: the CORS headers specified here still disallow implicit credentials, which is a good thing! This means that a user logged in on the subsurface cloud still can't unknowingly get their data stolen.

dirkhh:
please check again

julienw:
Looks like it works well!

mikeller:
> I wanted to mention one more thing: with the CORS headers configured, I don't really need this PR landed. I can work with the existing `file.js` by simply removing the prefix.

Rather than having this API endpoint serve the data as JSON _in addition_ to the existing .js file endpoint that is used for the Subsurface web interface (and have it 'dangling' with few eyes on it), I think it would make more sense to make it replace the .js endpoint and switch the web frontend to consume JSON.
Would you be ok to help us with this @julienw?

julienw:
I had a quick look at the code, I think it should be doable with some minimal changes. I'll have a look tomorrow.

julienw:
Hey,
I did something :-) It's not the best Javascript code but it seems to work properly (see last commit).

There's one drawback compared to before though: this doesn't work when opening the file locally anymore, `fetch` doesn't work in that context. And I think this is a goal for the export feature, so this change might not be good enough.
[A newer feature in browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) will make this possible in the future, but it's not implemented in Firefox yet (it's due for this year). It works fine in chrome though (I tested it).

So my recommendation would be to not implement this yet.

mikeller (approved):
This looks good, thank you!

@dirkhh do you have a way / want to test this in some way? Once it is merged we should be able to get this deployed to the Subsurface online log.

mikeller:
@dirkhh: I've merged this now, but obviously it will also need to be deployed to the web interface servers to go live.

PR #4506, opened 2025-04-19
Configuration: Update Settings for OSTC 4/5.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Update the dive computer settings that can be read, edited, and
persisted for the OSTC 4/5.

### Changes made:

1. update the dive computer settings that can be read, edited, and persisted for the OSTC 4/5;
2. reorganise the arrangement on the settings pages to be a bit more intuitive by separating read-only and read-write parts;
3. pull the 'Sync dive computer time with PC' option out of the settings page to be less confusing, and made it work when writing _and reading_ the settings.

### Related issues:

Requires https://github.com/mikeller/ostc4/pull/31 to fully work.

### Additional information:

image

image

### DIFF
--- a/core/configuredivecomputer.cpp
+++ b/core/configuredivecomputer.cpp
@@ -117,12 +117,9 @@ bool ConfigureDiveComputer::saveXMLBackup(const QString &fileName, const DeviceD
 
 	//Other Settings
 	writer.writeTextElement("DiveMode", QString::number(details.diveMode));
-	writer.writeTextElement("Saturation", QString::number(details.saturation));
-	writer.writeTextElement("Desaturation", QString::number(details.desaturation));
 	writer.writeTextElement("LastDeco", QString::number(details.lastDeco));
 	writer.writeTextElement("Brightness", QString::number(details.brightness));
 	writer.writeTextElement("Units", QString::number(details.units));
-	writer.writeTextElement("SamplingRate", QString::number(details.samplingRate));
 	writer.writeTextElement("Salinity", QString::number(details.salinity));
 	writer.writeTextElement("DiveModeColor", QString::number(details.diveModeColor));
 	writer.writeTextElement("Language", QString::number(details.language));
@@ -137,22 +134,35 @@ bool ConfigureDiveComputer::saveXMLBackup(const QString &fileName, const DeviceD
 	writer.writeTextElement("FutureTTS", QString::number(details.futureTTS));
 	writer.writeTextElement("CcrMode", QString::number(details.ccrMode));
 	writer.writeTextElement("DecoType", QString::number(details.decoType));
-	writer.writeTextElement("AGFSelectable", QString::number(details.aGFSelectable));
 	writer.writeTextElement("AGFHigh", QString::number(details.aGFHigh));
 	writer.writeTextElement("AGFLow", QString::number(details.aGFLow));
-	writer.writeTextElement("CalibrationGas", QString::number(details.calibrationGas));
 	writer.writeTextElement("FlipScreen", QString::number(details.flipScreen));
 	writer.writeTextElement("SetPointFallback", QString::number(details.setPointFallback));
-	writer.writeTextElement("LeftButtonSensitivity", QString::number(details.leftButtonSensitivity));
-	writer.writeTextElement("RightButtonSensitivity", QString::number(details.rightButtonSensitivity));
-	writer.writeTextElement("BottomGasConsumption", QString::number(details.bottomGasConsumption));
+	writer.writeTextElement("AlwaysShowppO2", QString::number(details.alwaysShowppO2));
 	writer.writeTextElement("DecoGasConsumption", QString::number(details.decoGasConsumption));
+
+	writer.writeComment("Heinrichs Weikamp OSTC4 only:");
+	writer.writeTextElement("TravelGasConsumption", QString::number(details.travelGasConsumption));
+	writer.writeTextElement("VPMConservatism", QString::number(details.vpmConservatism));
+	writer.writeTextElement("ButtonSensitivity", QString::number(details.buttonSensitivity));
+	writer.writeTextElement("ButtonBalance", QString::number(details.buttonBalance));
+
+	writer.writeComment("Not used on Heinrichs Weikamp OSTC4:");
+	writer.writeTextElement("BottomGasConsumption", QString::number(details.bottomGasConsumption));
+	writer.writeTextElement("Saturation", QString::number(details.saturation));
+	writer.writeTextElement("Desaturation", QString::number(details.desaturation));
+	writer.writeTextElement("SamplingRate", QString::number(details.samplingRate));
+	writer.writeTextElement("AGFSelectable", QString::number(details.aGFSelectable));
+	writer.writeTextElement("CalibrationGas", QString::number(details.calibrationGas));
+	writer.writeTextElement("RightButtonSensitivity", QString::number(details.rightButtonSensitivity));
+	writer.writeTextElement("LeftButtonSensitivity", QString::number(details.leftButtonSensitivity));
 	writer.writeTextElement("ModWarning", QString::number(details.modWarning));
 	writer.writeTextElement("DynamicAscendRate", QString::number(details.dynamicAscendRate));
 	writer.writeTextElement("GraphicalSpeedIndicator", QString::number(details.graphicalSpeedIndicator));
-	writer.writeTextElement("AlwaysShowppO2", QString::number(details.alwaysShowppO2));
+	writer.writeTextElement("SafetyStopEndDepth", QString::number(details.safetyStopResetDepth));
+	writer.writeTextElement("SafetyStopEndDepth", QString::number(details.safetyStopResetDepth));
 
-	// Suunto vyper settings.
+	writer.writeComment("Suunto vyper specific settings:");
 	writer.writeTextElement("Altitude", QString::number(details.altitude));
 	writer.writeTextElement("PersonalSafety", QString::number(details.personalSafety));
 	writer.writeTextElement("TimeFormat", QString::number(details.timeFormat));
@@ -414,6 +424,9 @@ bool ConfigureDiveComputer::restoreXMLBackup(const QString &fileName, DeviceDeta
 			if (settingName == "DecoType")
 				details.decoType = keyString.toInt();
 
+			if (settingName == "VPMConservatism")
+				details.vpmConservatism = keyString.toInt();
+
 			if (settingName == "AGFSelectable")
 				details.aGFSelectable = keyString.toInt();
 
@@ -438,17 +451,32 @@ bool ConfigureDiveComputer::restoreXMLBackup(const QString &fileName, DeviceDeta
 			if (settingName == "RightButtonSensitivity")
 				details.rightButtonSensitivity = keyString.toInt();
 
+			if (settingName == "ButtonSensitivity")
+				details.buttonSensitivity = keyString.toInt();
+
+			if (settingName == "ButtonBalance")
+				details.buttonBalance = keyString.toInt();
+
 			if (settingName == "BottomGasConsumption")
 				details.bottomGasConsumption = keyString.toInt();
 
 			if (settingName == "DecoGasConsumption")
 				details.decoGasConsumption = keyString.toInt();
 
+			if (settingName == "TravelGasConsumption")
+				details.travelGasConsumption = keyString.toInt();
+
 			if (settingName == "ModWarning")
 				details.modWarning = keyString.toInt();
 
 			if (settingName == "DynamicAscendRate")
 				details.dynamicAscendRate = keyString.toInt();
+
+			if (settingName == "SafetyStopEndDepth")
+				details.safetyStopEndDepth = keyString.toInt();
+
+			if (settingName == "SafetyStopResetDepth")
+				details.safetyStopResetDepth = keyString.toInt();
 
 			if (settingName == "GraphicalSpeedIndicator")
 				details.graphicalSpeedIndicator = keyString.toInt();
--- a/core/configuredivecomputerthreads.cpp
+++ b/core/configuredivecomputerthreads.cpp
@@ -368,20 +368,6 @@ static dc_status_t write_suunto_vyper_settings(dc_device_t *device, DeviceDetail
 		progress_cb(device_data.device, DC_EVENT_PROGRESS, &progress, userdata); \
 	} while (0)
 
-static dc_status_t ostc3_sync_time(device_data_t &data)
-{
-	dc_status_t rc = DC_STATUS_SUCCESS;
-
-	if (data.sync_time) {
-		dc_datetime_t now;
-		dc_datetime_localtime(&now, dc_datetime_now());
-
-		rc = dc_device_timesync(data.device, &now);
-	}
-
-	return rc;
-}
-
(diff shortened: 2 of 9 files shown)


core/configuredivecomputer.cpp
```diff
@@ -137,22 +134,35 @@ bool ConfigureDiveComputer::saveXMLBackup(const QString &fileName, const DeviceD
...
+	writer.writeTextElement("AlwaysShowppO2", QString::number(details.alwaysShowppO2));
 	writer.writeTextElement("DecoGasConsumption", QString::number(details.decoGasConsumption));
+
+	writer.writeComment("Heinrichs Weikamp OSTC4 only:");
+	writer.writeTextElement("TravelGasConsumption", QString::number(details.travelGasConsumption));
+	writer.writeTextElement("VPMConservatism", QString::number(details.vpmConservatism));
+	writer.writeTextElement("ButtonSensitivity", QString::number(details.buttonSensitivity));
+	writer.writeTextElement("ButtonBalance", QString::number(details.buttonBalance));
+
+	writer.writeComment("Not used on Heinrichs Weikamp OSTC4:");
```
glance-, line 150: Why stating what it's not used for here, rather than what it's used for, like in the other comments?

core/configuredivecomputerthreads.cpp
```diff
@@ -360,16 +361,36 @@ static dc_status_t write_suunto_vyper_settings(dc_device_t *device, DeviceDetail
...
+		dc_datetime_localtime(&now, dc_datetime_now());
+
+		rc = dc_device_timesync(data.device, &now);
+	}
+
+	return rc;
+}
 
-static dc_status_t read_ostc4_settings(dc_device_t *device, DeviceDetails &deviceDetails, dc_event_callback_t progress_cb, void *userdata)
+static dc_status_t read_ostc4_settings(device_data_t &device_data, DeviceDetails &deviceDetails, dc_event_callback_t progress_cb, void *userdata)
```
glance-, line 371: Why the type change here?

glance-: If you have a good reason for this type change and you're going to change the type the read_*_settings and write_*_settings functions take, you should change them all.

desktop-widgets/configuredivecomputerdialog.ui
```diff
@@ -2741,14 +2744,21 @@
               </property>
              </widget>
             </item>
+            <item row="2" column="0">
+             <widget class="QLabel">
```
glance-: Why a empty label?

glance- (changes requested):

core/configuredivecomputer.cpp
mikeller: I frankly don't know what is not used / used in any of the other dive computer models, and neither do I have access to any of them, so I am not in a position to figure this out.
But if you know more about what is used where, it is probably easiest if you add a clarification in a follow up pull request.

core/configuredivecomputerthreads.cpp
mikeller: > Why the type change here?

Because of this: https://github.com/subsurface/subsurface/pull/4506/files#diff-3507c651d293ef05d1c035ad0ff570252b85615743fa0bedae6e576eda2a5d48R375

> If you have a good reason for this type change and you're going to change the type the read___settings and write___settings functions take, you should change them all.

I disagree. Parameters supplied to a function should be defined by the minimal amount of context that is needed for the function to work, and not by what parameters are supplied to functions following the same naming pattern.

desktop-widgets/configuredivecomputerdialog.ui
mikeller: To prevent the Qt UI generator from collapsing row 2, and instead create the visual separation shown in [this image](https://private-user-images.githubusercontent.com/4742747/435338914-a6b98144-eac7-40fc-a05d-b118be975282.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDUyODQzMjQsIm5iZiI6MTc0NTI4NDAyNCwicGF0aCI6Ii80NzQyNzQ3LzQzNTMzODkxNC1hNmI5ODE0NC1lYWM3LTQwZmMtYTA1ZC1iMTE4YmU5NzUyODIucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDQyMiUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA0MjJUMDEwNzA0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MTExZDJmYjU1ZDQ2MTlmYWNjNzA2OWUxMzcwNGU3MzE0NmRiMjU0NDAwZjhlZGNhNmUwYjU5NmQ5MTNmMTdjNyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.oN9aT36ikYA6sqWeZwzsTZfBdZEfXRfILl55V9mjPrs).

It is probably not the best way to do this, but I am not versed enough in Qt's UI definition language, and a quick search didn't bring up something useful, so I went for this pragmatic approach that seems to be working just fine.
Feel free to add a suggestion for how to do this better.

core/configuredivecomputerthreads.cpp
glance-: read_ostc_settings , which doesn't consume anything else but device_data.device, also got  device_data, so either way that's a bit odd. 

I'd probably just used the existing deviceDetails.syncTime , or maybee even extract it to the run() functions, and call it from there.

core/configuredivecomputerthreads.cpp
```diff
@@ -360,16 +361,36 @@ static dc_status_t write_suunto_vyper_settings(dc_device_t *device, DeviceDetail
...
 	return rc;
 }
+#undef EMIT_PROGRESS
+
+#define EMIT_PROGRESS()	do { \
+		progress.current++; \
+		progress_cb(device_data.device, DC_EVENT_PROGRESS, &progress, userdata); \
+	} while (0)
+
+static dc_status_t ostc3_sync_time(device_data_t &data)
```
glance-: The ostc3_sync_time name is a bit confusing. It's neither ostc or ostc3 specific. I'd suggest just something like dc_sync_time or libdivecomputer_sync_time .

It's generic and can be used on any dc that supports dc_device_timesync, which is quite a few more than just OSTC's.

core/configuredivecomputerthreads.cpp
```diff
@@ -625,10 +656,14 @@ static dc_status_t read_ostc4_settings(dc_device_t *device, DeviceDetails &devic
 	deviceDetails.customText = ar.trimmed();
 	EMIT_PROGRESS();
 
+	if (rc == DC_STATUS_SUCCESS)
```
glance-: This check is redundant. If rc is anything but DC_STATUS_SUCCESS , we would already have returned.,

core/configuredivecomputerthreads.cpp
```diff
@@ -360,16 +361,36 @@ static dc_status_t write_suunto_vyper_settings(dc_device_t *device, DeviceDetail
...
+static dc_status_t read_ostc4_settings(device_data_t &device_data, DeviceDetails &deviceDetails, dc_event_callback_t progress_cb, void *userdata)
 {
 	// This code is really similar to the OSTC3 code, but there are minor
 	// differences in what the data means, and how to communicate with the
 	// device. If anyone can find a good way to harmonize the two, be my guest.
 	dc_status_t rc = DC_STATUS_SUCCESS;
 	dc_event_progress_t progress;
 	progress.current = 0;
-	progress.maximum = 47;
+	progress.maximum = 48;
```
glance-: I think this value isn't correct. Either it was wrong before, or it's wrong after this.

There's two READ_SETTING calls added, which calls EMIT_PROGRESS , and two manual calls to EMIT_PROGRESS, so I'd guess it should be 51.

I had some prototype code to generate these counters, but that was ugly and it was nested in the old qmake build system, so it died in the cmake migration.

desktop-widgets/configuredivecomputerdialog.ui
glance-: Ok.

One trick one can use is 
```
 <widget class="QLabel" name="label">
    <property name="whatsThis">
        <string extracomment="foobar"/>
    </property>
...
```
That way one can document this for the next guy.

Also probably a notr="true"  so it won't show up in the translations.

core/configuredivecomputerthreads.cpp
mikeller: Yes, I wasn't aware of `progress.maximum` when I added in the extra settings, and only came across it when changing the time synchronisation later on. Updating now.

core/configuredivecomputerthreads.cpp
mikeller: > read_ostc_settings , which doesn't consume anything else but device_data.device, also got device_data, so either way that's a bit odd.

That's an oversight on my part, I meant to copy the code to set the dive computer time into `read_ostc_settings()` as well.

> I'd probably just used the existing deviceDetails.syncTime , or maybee even extract it to the run() functions, and call it from there.

The problem is that `deviceDetails` isn't passed into the `read_...()` functions. So reusing `deviceData.sync_time` seemed to be a good option.

core/configuredivecomputerthreads.cpp
mikeller: > The ostc3_sync_time name is a bit confusing. It's neither ostc or ostc3 specific. I'd suggest just something like dc_sync_time or libdivecomputer_sync_time .
> 
> It's generic and can be used on any dc that supports dc_device_timesync, which is quite a few more than just OSTC's.

In the context of the dive computer configuration, this is actually specific to the OSTC 3 (and the OSTC 4/5, which uses the 'OSTC 3' protocol - this becomes evident as even the OSTC uses different code to generate the correct time format in `write_ostc_settings()`.

core/configuredivecomputer.cpp
glance-: Some simple grep'ing in the code quite clearly shows their usage.

BottomGasConsumption is still both read and written to a ostc4 / 5.  It still exists in the ostc4 source code as gasConsumption_bottom_l_min, but with this PR, it's just removed from the UI.

Saturation and Desaturation are used in the OSTC's and the OSTC3.

SamplingRate is used in the vyper settings, the  OSTC settings, and the OSTC3 settings.

AGFSelectable is used in the OSTC's and the OSTC3.

CalibrationGas is only used in the OSTC3's.

And so on.

I suggest either not re-sorting the file at all, sort the list just alphabetically, extract the device specific into sections, for all the specific ones, or do some complete documentation about what in the xml means what and is used by whom, with the last being way to much work for me to ever concider.

The fact is also that the values in the xml isn't really portable between different dive computers, and mostly just the internal representation that the dc uses.

core/configuredivecomputerthreads.cpp
glance-: No, its not. The OSTC uses different code because it never got ported to the new libdivecomputer api's when those were introduced. The ostc3 code was ported back in 128fc39274791525113bee5ac7882352beb8cb70 but the ostc code weren't.

The timezone bits there is just a red herring introduced in 6554e4f21e4deaa6b140ff09e07f629474257308 as a equality check , which probably produced a warning which got fixed in abc6c7d2b3ff630a0c7cbc5e1aeda74919e7a078.
The backing code in hw_ostc_device_timesync doesn't care about any timezone.

My old OSTC 2N screen died a couple of years back, so I can't really test this any longer.

glance-: We could even call sync_divecomputer_time from core/libdivecomputer.cpp here, rather than having it's own copy of a equivalent function.

core/configuredivecomputerthreads.cpp
glance-: All the different functions in core/configuredivecomputerthreads.cpp were designed to look and feel the same for less confusion when working on them. That's why I think we should keep them the same.

I think the thing that a read operation changes the state on the device is questionable.

I'd completely forgotten about the existence of deviceData.sync_time.

core/configuredivecomputerthreads.cpp
```diff
@@ -1169,10 +1203,14 @@ static dc_status_t read_ostc3_settings(dc_device_t *device, DeviceDetails &devic
 	deviceDetails.customText = ar.trimmed();
 	EMIT_PROGRESS();
 
+	if (rc == DC_STATUS_SUCCESS)
```
glance-: This check is redundant. If rc is anything but DC_STATUS_SUCCESS , we would already have returned.,

core/configuredivecomputerthreads.cpp
```diff
@@ -1423,30 +1462,27 @@ static dc_status_t write_ostc3_settings(dc_device_t *device, const DeviceDetails
...
 	EMIT_PROGRESS();
 
-	//sync date and time
-	if (deviceDetails.syncTime) {
-		dc_datetime_t now;
-		dc_datetime_localtime(&now, dc_datetime_now());
-
-		rc = dc_device_timesync(device, &now);
-	}
+	if (rc == DC_STATUS_SUCCESS)
```
glance-: This check is redundant. If rc is anything but DC_STATUS_SUCCESS , we would already have returned.,

core/configuredivecomputerthreads.cpp
```diff
@@ -883,28 +921,19 @@ static dc_status_t write_ostc4_settings(dc_device_t *device, const DeviceDetails
...
-
-	//sync date and time
-	if (deviceDetails.syncTime) {
-		dc_datetime_t now;
-		dc_datetime_localtime(&now, dc_datetime_now());
-
-		rc = dc_device_timesync(device, &now);
-	}
-
+	if (rc == DC_STATUS_SUCCESS)
```
glance-: This check is redundant. If rc is anything but DC_STATUS_SUCCESS , we would already have returned.,

core/configuredivecomputerthreads.cpp
mikeller: I disagree. Having excess scope in function signatures makes it more confusing to understand the code, and harder to maintain.

core/configuredivecomputerthreads.cpp
mikeller: Thanks, I wasn't aware of the context.

core/configuredivecomputer.cpp
mikeller: Frankly I don't get what you want here.
My proposed change is adding _some_ context to make it easier for users to figure out what is relevant for their dive computer when they are using this functionality to back up the settings on their dive computer. It does not solve _everything_, but it is improving things.
And as you say, sorting this out in all detail for the remaining (mostly legacy) dive computers seems to be a big task, for little gain.

mikeller:
@glance- do you have any more actual suggestions for improvements, or is this ready to be merged?

PR #3568, opened 2022-12-25
Desktop: Fix brightness configuration for OSTC4

mikeller:
### Describe the pull request:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Pull request long description:

The Heinrichs Weikamp OSTC4 introduced an additional
'Cave' brightness level that is dimmer than all existing ones.

### Changes made:

1) 'Cave' level 'for the brightness combobox contents for OSTC4

### Release note:

Not sure of the desired level of detail covered in the changelog, but this is my proposal:
```desktop: fix brightness configuration for OSTC4```

### Mentions:

Signed-off-by: Michael Keller <github@ike.ch>

PR #4510, opened 2025-04-23
Remove the broken Divesoft model detection

jefdriesen:
Only the V2 header contains a serial number, and there are also cases where the serial number hasn't been filled in correctly.

Libdivecomputer doesn't use the actual model number for parsing the Divesoft dives, so hardcoding the Freedom works fine. The main disadvantage is that all imported dives will show the Divesoft Freedom as their dive computer model.

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/divesoft.cpp
+++ b/core/divesoft.cpp
@@ -26,7 +26,7 @@ int divesoft_import(const std::string &buffer, struct divelog *log)
 		return report_error("%s", translate("gettextFromC", "Unknown DC"));
 
 	auto d = std::make_unique<dive>();
-	d->dcs[0].model = devdata.vendor + " " + devdata.model + " (Imported from file)";
+	d->dcs[0].model = devdata.vendor + " (Imported from file)";
 
 	// Parse the dive data
 	dc_status_t rc = libdc_buffer_parser(d.get(), &devdata, (const unsigned char *)buffer.data(), buffer.size());
--- a/core/divesoft.cpp
+++ b/core/divesoft.cpp
@@ -13,27 +13,15 @@
 #include "format.h"
 #include "libdivecomputer.h"
 
-// As supplied by Divesoft
-static constexpr std::string_view divesoft_liberty_serial_prefix = "7026";
-static constexpr std::string_view divesoft_freedom_serial_prefix = "7044";
-static constexpr std::string_view divesoft_freedom_plus_serial_prefix = "7273";
-
 // From libdivecomputer
 static constexpr int divesoft_liberty_model = 10;
 static constexpr int divesoft_freedom_model = 19;
 
 int divesoft_import(const std::string &buffer, struct divelog *log)
 {
-	std::string model_identifier = buffer.substr(52, 4);
-	int model = 0;
-	if (model_identifier == divesoft_liberty_serial_prefix)
-		model = divesoft_liberty_model;
-	else if (model_identifier == divesoft_freedom_serial_prefix || model_identifier == divesoft_freedom_plus_serial_prefix)
-		model = divesoft_freedom_model;
-
 	device_data_t devdata;
 	devdata.log = log;
-	int ret = prepare_device_descriptor(model, DC_FAMILY_DIVESOFT_FREEDOM, devdata);
+	int ret = prepare_device_descriptor(divesoft_freedom_model, DC_FAMILY_DIVESOFT_FREEDOM, devdata);
 	if (ret == 0)
 		return report_error("%s", translate("gettextFromC", "Unknown DC"));
 


mikeller:
> Only the V2 header contains a serial number, and there are also cases where the serial number hasn't been filled in correctly.

Did you get an answer from Divesoft as to why the serial number should be missing on logs generated by devices with current firmware?

> Libdivecomputer doesn't use the actual model number for parsing the Divesoft dives, so hardcoding the Freedom works fine. The main disadvantage is that all imported dives will show the Divesoft Freedom as their dive computer model.

Not sure I would call it a 'bugfix' to remove functionality that is working exactly as it should for most users, and instead return information that is verifiably incorrect.

core/divesoft.cpp
```diff
@@ -13,27 +13,15 @@
...
-	int model = 0;
-	if (model_identifier == divesoft_liberty_serial_prefix)
-		model = divesoft_liberty_model;
-	else if (model_identifier == divesoft_freedom_serial_prefix || model_identifier == divesoft_freedom_plus_serial_prefix)
-		model = divesoft_freedom_model;
-
 	device_data_t devdata;
 	devdata.log = log;
-	int ret = prepare_device_descriptor(model, DC_FAMILY_DIVESOFT_FREEDOM, devdata);
+	int ret = prepare_device_descriptor(divesoft_freedom_model, DC_FAMILY_DIVESOFT_FREEDOM, devdata);
```
mikeller, line 24: This is incorrect for all devices except for the Divesoft Freedom. Please fix.

mikeller (changes requested):

glance-:
LGTM. This was one of the things I suggested back in https://github.com/subsurface/subsurface/pull/4367#discussion_r1842404054 , 

Future improvements would be that libdivecomputer would figure out the model value itself, if it could and return that in a DC_EVENT_DEVINFO event.

jefdriesen:
> Did you get an answer from Divesoft as to why the serial number should be missing on logs generated by devices with current firmware?

Yes. The factory test dive was done before the serial number was filled in. So this shouldn't affect real dives.

core/divesoft.cpp
jefdriesen: I can't do that without copying significant parts of the parser code. Even if we do that, I don't have a complete list of the product numbers and their corresponding model number.

mikeller:
> LGTM. This was one of the things I suggested back in [#4367 (comment)](https://github.com/subsurface/subsurface/pull/4367#discussion_r1842404054) ,
> 
> Future improvements would be that libdivecomputer would figure out the model value itself, if it could and return that in a DC_EVENT_DEVINFO event.

I am not sure about the approach of 'let's break this now for all Divesoft users except the ones using a Freedom, and then wait for it to be fixed at a future point in time'. How about this alternative: You go and implement the change in libdivecomputer now, and we merge this when your change is available?



> > Did you get an answer from Divesoft as to why the serial number should be missing on logs generated by devices with current firmware?
> 
> Yes. The factory test dive was done before the serial number was filled in. So this shouldn't affect real dives.

Ok, that makes sense. Thanks for checking.

core/divesoft.cpp
mikeller: > I can't do that without copying significant parts of the parser code.

Couldn't the parser just report. back the correct dive computer with `DC_EVENT_DEVINFO`?
Alternatively, the current implementation seems to be what is needed to make this work correctly - it is just incomplete. And it could probably have a fall through case if the serial number does not match a known pattern.

> Even if we do that, I don't have a complete list of the product numbers and their corresponding model number.

We could just ask Divesoft for a list - when I originally implemented this that worked like a charm. It just looks like they have added more different models in the meantime.

glance-:
Nothing is going to get broken, rather more cases will get supported.

Could you state a case where it would break?

 I have some old V1 DLF files which the current code can't import, and this fixes that.
[dlf.zip](https://github.com/user-attachments/files/19876239/dlf.zip)

mikeller:
> Nothing is going to get broken, rather more cases will get supported.
> 
> Could you state a case where it would break?

See https://github.com/subsurface/subsurface/pull/4510#discussion_r2056050144.

> I have some old V1 DLF files which the current code can't import, and this fixes that. [dlf.zip](https://github.com/user-attachments/files/19876239/dlf.zip)

See https://github.com/subsurface/subsurface/pull/3568#pullrequestreview-1274995287. You need to update the firmware on your Divesoft dive computer to get it supported in Subsurface.
Alternatively, if you want to import historical log files, you can always use an [older version of Subsurface](https://github.com/subsurface/nightly-builds/releases/tag/v6.0.5289-CICD-release) to do so - they are still available.

glance-:
> > Nothing is going to get broken, rather more cases will get supported.
> > Could you state a case where it would break?
> 
> See [#4510 (comment)](https://github.com/subsurface/subsurface/pull/4510#discussion_r2056050144).

Can you show me a case where it will break? Aka a DLF file wich can't be imported after this change?

> > I have some old V1 DLF files which the current code can't import, and this fixes that. [dlf.zip](https://github.com/user-attachments/files/19876239/dlf.zip)
> 
> See [#3568 (review)](https://github.com/subsurface/subsurface/pull/3568#pullrequestreview-1274995287). You need to update the firmware on your Divesoft dive computer to get it supported in Subsurface. Alternatively, if you want to import historical log files, you can always use an [older version of Subsurface](https://github.com/subsurface/nightly-builds/releases/tag/v6.0.5289-CICD-release) to do so - they are still available.

Will old DLF files get rewritten on firmware upgrade?

Settings are different from importing log files. Settings affect the future usage of the computer, and if you're there changing your settings, you should take the time to upgrade the firmware to.

Log data are a different thing. These log files might be from a device which no longer has them on it, thus it can't be upgraded. Why not support those log files if it just involves removing unnecessary code?

glance- (approved):

PR #4511, opened 2025-04-24
Import: Improve the Selection of bluetooth Transport.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In the 'Import from dive computer' dialog, only enable selection of
bluetooth as the transport if bluetooth is a supported transport for the
selected dive computer model.
If bluetooth is _the only_ option, pre-select bluetooth and make it
impossible to deselect.
This should reduce user confusion around the correct transport mode to
choose.

### Changes made:

1. in the 'Import from dive computer' dialog, only enable selection of bluetooth as the transport if bluetooth is a supported transport for the selected dive computer model;
2. if bluetooth is _the only_ option, pre-select bluetooth and make it impossible to deselect.

### Additional information:

bluetooth not supported:
(@jefdriesen not quite sure about this one - is it possible for users to use a 'serial to bluetooth' adaptor with a dive computer with a built in serial connection, and then connect to the dive computer directly in Subsurface, without setting up a serial port device in the operating system first?)

image

bluetooth and others supported:

image

Only bluetooth supported:

image

### DIFF
--- a/desktop-widgets/downloadfromdivecomputer.cpp
+++ b/desktop-widgets/downloadfromdivecomputer.cpp
@@ -368,7 +368,7 @@ void DownloadFromDCWidget::on_vendor_currentTextChanged(const QString &vendor)
 
 void DownloadFromDCWidget::on_product_currentTextChanged(const QString &)
 {
-	updateTransportSelection();
+	updateTransportSelection(true);
 }
 
 void DownloadFromDCWidget::on_device_currentTextChanged(const QString &device)
@@ -592,7 +592,7 @@ void DownloadFromDCWidget::on_ok_clicked()
 	}
 }
 
-void DownloadFromDCWidget::updateTransportSelection()
+void DownloadFromDCWidget::updateTransportSelection(bool changeSelection)
 {
 	unsigned int transports = dc_descriptor_get_transports(descriptorLookup.value(ui.vendor->currentText().toLower() + ui.product->currentText().toLower()));
 
@@ -607,17 +607,20 @@ void DownloadFromDCWidget::updateTransportSelection()
 
 #if defined(BT_SUPPORT)
 	if (transports & (DC_TRANSPORT_BLUETOOTH | DC_TRANSPORT_BLE)) {
-		if (!(transports & ~(DC_TRANSPORT_BLUETOOTH | DC_TRANSPORT_BLE))) {
-			// only bluetooth is supported by the dive computer, pre-select it
+		unsigned int nonBluetoothTransports = transports & ~(DC_TRANSPORT_BLUETOOTH | DC_TRANSPORT_BLE);
+		if (!nonBluetoothTransports) {
 			ui.bluetoothMode->setChecked(true);
 			ui.bluetoothMode->setEnabled(false);
 			ui.chooseBluetoothDevice->setEnabled(true);
 		} else {
+			if (changeSelection && nonBluetoothTransports == DC_TRANSPORT_SERIAL)
+				// only bluetooth and (virtual bluetooth) serial port are supported - prefer bluetooth
+				ui.bluetoothMode->setChecked(true);
+
 			ui.bluetoothMode->setEnabled(true);
 			ui.chooseBluetoothDevice->setEnabled(true);
 		}
 	} else {
-
 		ui.bluetoothMode->setChecked(false);
 		ui.bluetoothMode->setEnabled(false);
 		ui.chooseBluetoothDevice->setEnabled(false);
@@ -648,7 +651,7 @@ void DownloadFromDCWidget::markChildrenAsDisabled()
 
 void DownloadFromDCWidget::markChildrenAsEnabled()
 {
-	updateTransportSelection();
+	updateTransportSelection(false);
 	ui.vendor->setEnabled(true);
 	ui.product->setEnabled(true);
 	ui.forceDownload->setEnabled(true);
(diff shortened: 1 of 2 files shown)


jefdriesen:
> bluetooth not supported: (@jefdriesen not quite sure about this one - is it possible for users to use a 'serial to bluetooth' adaptor with a dive computer with a built in serial connection, and then connect to the dive computer directly in Subsurface, without setting up a serial port device in the operating system first?)

I never tried anything like that before, so I don't know for sure. For a bluetooth classic adapter, I suspect this may work because rfcomm behaves very similar to a normal serial port (both are stream oriented transports) and libdivecomputer doesn't treat `DC_TRANSPORT_SERIAL` and `DC_TRANSPORT_BLUETOOTH` different. For a BLE adapter, I don't think it will work without an extra custom I/O to handle the packets and disguise the BLE transport as a serial one (e.g. as done in libdivecomputer's src/packet.c).

mikeller:
> > bluetooth not supported: (@jefdriesen not quite sure about this one - is it possible for users to use a 'serial to bluetooth' adaptor with a dive computer with a built in serial connection, and then connect to the dive computer directly in Subsurface, without setting up a serial port device in the operating system first?)
> 
> I never tried anything like that before, so I don't know for sure. For a bluetooth classic adapter, I suspect this may work because rfcomm behaves very similar to a normal serial port (both are stream oriented transports) and libdivecomputer doesn't treat `DC_TRANSPORT_SERIAL` and `DC_TRANSPORT_BLUETOOTH` different. For a BLE adapter, I don't think it will work without an extra custom I/O to handle the packets and disguise the BLE transport as a serial one (e.g. as done in libdivecomputer's src/packet.c).

Thanks @jefdriesen - based on this I think the benefits of the proposed change outweigh the potential downsides.

One more thing I noticed is that some of the newer, bluetooth based technical dive computers like the (Shearwater Petrel 2)[https://github.com/subsurface/libdc/blob/644b7d79ad6ba1b1442c3ac9c0cfdb853a66eccf/src/descriptor.c#L370] or the [Heinrichs Weikamp OSTC 4/5](https://github.com/subsurface/libdc/blob/644b7d79ad6ba1b1442c3ac9c0cfdb853a66eccf/src/descriptor.c#L336) are listing `DC_TRANSPORT_SERIAL`, even though there is no obvious way to attach a serial device to them. What is the practical use of their 'serial communication' capability?

jefdriesen:
> One more thing I noticed is that some of the newer, bluetooth based technical dive computers like the (Shearwater Petrel 2)[https://github.com/subsurface/libdc/blob/644b7d79ad6ba1b1442c3ac9c0cfdb853a66eccf/src/descriptor.c#L370] or the [Heinrichs Weikamp OSTC 4/5](https://github.com/subsurface/libdc/blob/644b7d79ad6ba1b1442c3ac9c0cfdb853a66eccf/src/descriptor.c#L336) are listing `DC_TRANSPORT_SERIAL`, even though there is no obvious way to attach a serial device to them. What is the practical use of their 'serial communication' capability?

With bluetooth classic devices it's possible to create a virtual serial port. That's the reason why `DC_TRANSPORT_BLUETOOTH` is always combined with `DC_TRANSPORT_SERIAL`.

mikeller:
> With bluetooth classic devices it's possible to create a virtual serial port. That's the reason why `DC_TRANSPORT_BLUETOOTH` is always combined with `DC_TRANSPORT_SERIAL`.

I see. This seems to be working reasonably well, at least under linux where the bluetooth connection is only established once Subsurface is trying to connect. But not sure if any / many people are using it, or if there are cases where this is the only possible way to download from a dive computer.
Do you know if there are any devices that support bluetooth, but also support serial communication over a physical connection?

jefdriesen:
> > With bluetooth classic devices it's possible to create a virtual serial port. That's the reason why `DC_TRANSPORT_BLUETOOTH` is always combined with `DC_TRANSPORT_SERIAL`.
> 
> I see. This seems to be working reasonably well, at least under linux where the bluetooth connection is only established once Subsurface is trying to connect. But not sure if any / many people are using it, or if there are cases where this is the only possible way to download from a dive computer.

If an application doesn't have a bluetooth classic transport available, the virtual serial port emulation is the only option. For example the libdivecomputer built-in implementation is only available on Windows and Linux, but not on Mac, BSD, etc.

>Do you know if there are any devices that support bluetooth, but also support serial communication over a physical connection?

I'm not 100% sure, but I think some of the Ratio computers support both. They have the same USB connector as the non-bluetooth models, but I don't know if it can be used for charging only. I can't check because the one I have for testing is a non-bluetooth model. I can ask Ratio.

mikeller:
> If an application doesn't have a bluetooth classic transport available, the virtual serial port emulation is the only option. For example the libdivecomputer built-in implementation is only available on Windows and Linux, but not on Mac, BSD, etc.

Makes sense - we probably want to keep this in Subsurface as an alternative way to connect.

> I'm not 100% sure, but I think some of the Ratio computers support both. They have the same USB connector as the non-bluetooth models, but I don't know if it can be used for charging only. I can't check because the one I have for testing is a non-bluetooth model. I can ask Ratio.

Yes please - in case we ever opt for a solution where the the serial transport isn't enabled if it's only a stand-in for bluetooth.

jefdriesen:
Ratio confirmed their bluetooth enabled dive computers also support USB communication. So that means there are indeed dive computers where `DC_TRANSPORT_SERIAL` is enabled for both usb-serial communication and the bluetooth virtual serial port emulation.

mikeller:
> Ratio confirmed their bluetooth enabled dive computers also support USB communication. So that means there are indeed dive computers where `DC_TRANSPORT_SERIAL` is enabled for both usb-serial communication and the bluetooth virtual serial port emulation.

Thanks for getting this clarified. I think the current proposal should still work fine for these models, as bluetooth is only pre-selected, but not enforced if bluetooth _and_ serial are supported.

PR #4531, opened 2025-06-05
[Pro forma]Garmin: Add MTP Support for More Models.

mikeller:
For artifact testing only!

N.b.: This will not work on MacOS, as MTP support on MacOS has been disabled.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Add MTP support for more of the newer Garmin Descent models.


siw1973:
Michael, 

I'm not familiar with how to install the patches, which of the files do I need to directly install from the Windows ZIP ?

image

Also, I see with interest that there is a new Android APK, does this have an ability to download into Subsurface Android App from a Garmin that's connected ?? I thought that you'd agreed not to mess with a directly Bluetooth connected computer, but if the Garmin is connected to Android and is exposed in MTP that could be a way to download from the watch direct using standard already exposed protocols for file access rather than a bespoke probing.

Or it could just be a generic updated APK that get's spat out as part of your build process :-(

mikeller:
> I'm not familiar with how to install the patches, which of the files do I need to directly install from the Windows ZIP ?

`subsurface-<version and name>-installer.exe` is the Windows installer.

> Also, I see with interest that there is a new Android APK, does this have an ability to download into Subsurface Android App from a Garmin that's connected ?? I thought that you'd agreed not to mess with a directly Bluetooth connected computer, but if the Garmin is connected to Android and is exposed in MTP that could be a way to download from the watch direct using standard already exposed protocols for file access rather than a bespoke probing.

No, unfortunately file based import is not supported in the mobile versions.
 
> Or it could just be a generic updated APK that get's spat out as part of your build process :-(

Yes, this is correct, these are all built automatically, for all pull requests and releases, even if there are no changes for a particular version.

siw1973:
> > I'm not familiar with how to install the patches, which of the files do I need to directly install from the Windows ZIP ?
> 
> `subsurface-<version and name>-installer.exe` is the Windows installer.

Thanks for this !

> 
> > Also, I see with interest that there is a new Android APK, does this have an ability to download into Subsurface Android App from a Garmin that's connected ?? I thought that you'd agreed not to mess with a directly Bluetooth connected computer, but if the Garmin is connected to Android and is exposed in MTP that could be a way to download from the watch direct using standard already exposed protocols for file access rather than a bespoke probing.
> 
> No, unfortunately file based import is not supported in the mobile versions.

Is this a structural constraint, or just that the need is so limited that it's not worth investing any time in ?

> 
> > Or it could just be a generic updated APK that get's spat out as part of your build process :-(
> 
> Yes, this is correct, these are all built automatically, for all pull requests and releases, even if there are no changes for a particular version.

Cool, thought this would actually be the case unfortunately.....

mikeller:
@siw1973:

> Is this a structural constraint, or just that the need is so limited that it's not worth investing any time in ?

'bit of both' I think - Android and in particular iOS make it hard for applications to access files that were not created by the application, and are not of a type registered by the application. But it's also that nobody has yet put in the time to try and overcome these limitations.

PR #4538, opened 2025-06-15
Divesoft: Fix the Log File Import and Add New Liberty Model.

### LINKED ISSUE #4536: [Bug]: |Issue importing log from divesoft liberty
### Description

When attempting to import a dive log from a divesoft liberty, currently getting a `Unknown DC` error. I have attached a log file I am having issues witth

### Subsurface Version

6.0.5365

### Platform(s)

Windows

### Platform Version

Windows 10

### Divecomputer Model

Divesoft Liberty 2021

### Steps to Reproduce

1. Go to top menu
2. Click on `Import > Import log file`
3. Select a log exported from liberty

Attached a zip with a dive log in it

[L_Dive-01478656.zip](https://github.com/user-attachments/files/20747358/L_Dive-01478656.zip)

### Logs

```shell
INFO: validateGL(): created OpenGLContext.
INFO: validateGL(): obtained QOpenGLFunctions.
INFO: validateGL(): detected OpenGL version 4.6.
INFO: loading en-GB translations
QObject::connect(QQuickWindow, QDeclarativeGeoMap_QML_6): invalid nullptr parameter
INFO: DivePixmaps DPR: 1.000000 metrics: 24 16 sz_bigger: 40
INFO: returning cloud URL https://ssrf-cloud-eu.subsurface-divelog.org/git/michael.fisher86@gmail.com[michael.fisher86@gmail.com]
INFO: loading dive data from: https://ssrf-cloud-eu.subsurface-divelog.org/git/michael.fisher86@gmail.com[michael.fisher86@gmail.com]
Cloud storage: checking connection to cloud server https://ssrf-cloud-eu.subsurface-divelog.org/git/michael.fisher86@gmail.com
INFO: Checking cloud connection...
INFO: IP used for cloud server access 88.97.224.138
Cloud storage: successfully checked connection to cloud server
INFO: git storage: Synchronising data file
INFO: git storage: update local repo

INFO: git storage: sync with remote https://ssrf-cloud-eu.subsurface-divelog.org/git/michael.fisher86@gmail.com[michael.fisher86@gmail.com]

INFO: git storage: Sync with cloud storage
INFO: git storage: delete proxy setting

Cloud storage: checking connection to cloud server https://ssrf-cloud-eu.subsurface-divelog.org/git/michael.fisher86@gmail.com
INFO: Checking cloud connection...
Cloud storage: successfully checked connection to cloud server
INFO: git storage: fetch remote https://ssrf-cloud-eu.subsurface-divelog.org/git/michael.fisher86@gmail.com

INFO: git storage: Successful cloud connection, fetch remote
INFO: git storage: certificate callback for host ssrf-cloud-eu.subsurface-divelog.org with validity 1

INFO: git storage: certificate callback for host ssrf-cloud-eu.subsurface-divelog.org with validity 1

INFO: git storage: check remote status

INFO: git storage: try to update

INFO: git storage: Done syncing with cloud storage
INFO: git storage: Load dives from local cache
ERROR: Dive: 262, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 263, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 264, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 265, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 266, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 267, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 268, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 269, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 270, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 271, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 272, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 273, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 280, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 281, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 285, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 286, dive computer: DLF import: CCR dive, but no diluent cylinder found. Please add or select the correct cylinder use.
ERROR: Dive: 287, dive computer
```

### Mentions

_No response_

mikeller:

- [x] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix the import of Divesoft log files to default to the Freedom model if
an unknown model identifier is encountered.
Also add support for the 'Liberty 2021' model identifier.

### Changes made:

1. fix the import of Divesoft log files to default to the Freedom model if an unknown model identifier is encountered;
2. add support for the 'Liberty 2021' model identifier.

### Related issues:

Fixes #4536.


glance-:
I'd still prefer #4510 which solves this issue in a more future proof and less fragile way.

mikeller:
> I'd still prefer #4510 which solves this issue in a more future proof and less fragile way.

If you are able to supply explanations for what you consider to be 'not future proof' and 'fragile' in the proposed approach I am happy to consider amendments to address them.

glance-:
This still assumes that the DLF-files are in V2 format, without verifying that. It would be better to just leave the whole file parsing to libdivecomputer, which already supports both V1 and V2 files.

Also, I think it's unnecessary to emit a error if the bytes checked doesn't match  one of the known prefixes. Just leave that parsing to libdivecomputer, which already handles it.

mikeller:
@glance-: I can not see how any of your statements relate to 'not being future proof' or to being 'fragile'.  In the future please avoid making allegations that you can not back up, or your comments will be ignored.

> Also, I think it's unnecessary to emit a error if the bytes checked doesn't match one of the known prefixes. Just leave that parsing to libdivecomputer, which already handles it.

If you read the source of libdivecomputer, you will notice that the parser in libdivecomputer does not currently check or fix the model of Divesoft dive computer that is set, it just accepts it even if it is incorrect. So setting an incorrect model to start with will result in a bug and an incorrect dive computer being displayed.

glance-:
> @glance-: I can not see how any of your statements relate to 'not being future proof' or to being 'fragile'. In the future please avoid making allegations that you can not back up, or your comments will be ignored.

This type of threats is not productive. If you don't understand my arguments, fine, but making assumptions and threats are really harmful for this community.

> > Also, I think it's unnecessary to emit a error if the bytes checked doesn't match one of the known prefixes. Just leave that parsing to libdivecomputer, which already handles it.
> 
> If you read the source of libdivecomputer, you will notice that the parser in libdivecomputer does not currently check or fix the model of Divesoft dive computer that is set, it just accepts it even if it is incorrect. So setting an incorrect model to start with will result in a bug and an incorrect dive computer being displayed.

Which I'd argue is a way lesser bug than dropping support for all V1 DLF files and popping unnecessary error messages.

If you think the importance of showing the right model is so high, please put that parsing code after the libdc_buffer_parser call, where you have access to libdivecomputer's parsed information, that this is actually a parsable DLF file and so on.

mikeller:
@glance-:

> This type of threats is not productive.

This is not a treat - I am just highlighting a potential consequence of your behaviour.

> If you don't understand my arguments, fine, but making assumptions and threats are really harmful for this community.

The problem is that you are not making reasoned arguments, you are just making unfounded allegations and then not backing them when asked to.
And yes, I think that your toxic behaviour is harmful.

glance-:
There are two cases reported, https://groups.google.com/g/subsurface-divelog/c/dsr0mPdfH3U/m/kDPpnNl3AAAJ and https://groups.google.com/g/subsurface-divelog/c/WOP00QlYOWM/m/4Nc6XatlDgAJ where this code that assumes all files are DLF v2 and trying to match serial numbers have failed for users. This clearly shows the fragility of this approach.

The other argument is that this code is unnecessary, because the actual parsing is done by libdivecomputer. Just removing these assumptions based on serial number will solve these and and future changes and added ranges of serial numbers.

If we would like to do any additional parsing of the DLF content to augment the libdivecomputer parsing, either add it to libdivecomputer or add it after the libdivecomputer parsing of the dive file, and don't reject the file just because the data doesn't match subsurface assumptions of how the data looks.

I'd would have just excluded any model name from the model string, and just write something like:
> d->dcs[0].model = devdata.vendor + " (Imported from file)";

mikeller:
> There are two cases reported, [groups.google.com/g/subsurface-divelog/c/dsr0mPdfH3U/m/kDPpnNl3AAAJ](https://groups.google.com/g/subsurface-divelog/c/dsr0mPdfH3U/m/kDPpnNl3AAAJ) and [groups.google.com/g/subsurface-divelog/c/WOP00QlYOWM/m/4Nc6XatlDgAJ](https://groups.google.com/g/subsurface-divelog/c/WOP00QlYOWM/m/4Nc6XatlDgAJ) where this code that assumes all files are DLF v2 and trying to match serial numbers have failed for users. This clearly shows the fragility of this approach.

You are incorrect. None of the cases that you reference applies to the code in this pull request. Please stop spreading lies.

> The other argument is that this code is unnecessary, because the actual parsing is done by libdivecomputer. Just removing these assumptions based on serial number will solve these and and future changes and added ranges of serial numbers.

You are incorrect again - currently the parser in libdivecomputer does not try to detect the correct type of Divesoft dive computer: https://github.com/subsurface/subsurface/pull/4510#issuecomment-2824491678

glance-:
Working code doesn't lie: #4542

mikeller:
> Working code doesn't lie: #4542

Thanks for adding tests. They work just fine on this pull request.

glance-:
> > Working code doesn't lie: #4542
> 
> Thanks for adding tests. They work just fine on this pull request.

I expect a big apology for your toxic accusatory behavior. Your behavior here and in other places, are just not acceptable and a violation of the CoC.

mikeller:
> I expect a big apology for your toxic accusatory behavior.

No, definitely not.
I stand by the comments I made about your problematic behaviour. In my opinion, the fact that you made a contribution to the project does not excuse or redeem any of your behaviour.

If you have the expectation that contributions in this project work in this kind of transactional manner then I'd rather not use your contribution, and instead add my own tests.

glance-:
> > I expect a big apology for your toxic accusatory behavior.
> 
> No, definitely not. I stand by the comments I made about your problematic behaviour. In my opinion, the fact that you made a contribution to the project does not excuse or redeem any of your behaviour.

Wait a minute...

> > There are two cases reported, [groups.google.com/g/subsurface-divelog/c/dsr0mPdfH3U/m/kDPpnNl3AAAJ](https://groups.google.com/g/subsurface-divelog/c/dsr0mPdfH3U/m/kDPpnNl3AAAJ) and [groups.google.com/g/subsurface-divelog/c/WOP00QlYOWM/m/4Nc6XatlDgAJ](https://groups.google.com/g/subsurface-divelog/c/WOP00QlYOWM/m/4Nc6XatlDgAJ) where this code that assumes all files are DLF v2 and trying to match serial numbers have failed for users. This clearly shows the fragility of this approach.
> 
> You are incorrect. None of the cases that you reference applies to the code in this pull request. Please stop spreading lies.

https://groups.google.com/g/subsurface-divelog/c/dsr0mPdfH3U/m/kDPpnNl3AAAJ?pli=1 is clearly about a new serial number range, which the broken Divesoft model detection code couldn't handle. So calling it a lie is just abusive.

In https://groups.google.com/g/subsurface-divelog/c/WOP00QlYOWM/m/4Nc6XatlDgAJ the following is stated:
> For the last few years, I have been able to successfully import .dlf files into Subsurface (Divesoft Liberty and Freedom logs). As of these two versions, the import returns a 'Unknown DC' banner at the bottom of the screen.
Which only can imply that it's DLF v1 format on the files the user tries to import, because that's the only format the old log file importer supported, which the broken Divesoft model detection code couldn't handle. So calling it a lie is just abusive.

> If you have the expectation that contributions in this project work in this kind of transactional manner then I'd rather not use your contribution, and instead add my own tests.

A quick check showed that I've contributed 643 commits that's already in master. Nothing transactional about that. The test code was to provide proof as code that the Divesoft model detection code was fragile and broken, and removing it solves the issues reported and is a less fragile way forward. I also really dislike when someone throws around false accusations, which I can disprove with a couple of simple lines of code.

So, I can't see how you can provide a single coherent argument that anything I've said is a lie. 

Also comments like (from https://github.com/subsurface/subsurface/pull/4510#issuecomment-2825319584)
> I am not sure about the approach of 'let's break this now for all Divesoft users except the ones using a Freedom, and then wait for it to be fixed at a future point in time'.

Is also false. Nothing got broken, just more things work, and they work better with less code.

A quite common definition of a "lie" is:
"Lies" refers to statements that are known to be untrue, intentionally made to deceive.

I've clearly showed that nothing I've said (or written) is untrue, but I've shown that multiple things you have written is. I won't be toxic and call them "lies" for that, just that they're untrue. I don't know your intentions with your statements, so I won't try to guess them.

jefdriesen:
@mikeller: @glance- isn't lying! There is plenty of data available to support our claims.

The current code and your proposed fix assumes a V2 header without checking. For a V1 DLF file, that will result in interpreting some random data as the serial number. That code is simply wrong.

The list of serial number prefixes in your proposed fix is still incomplete. There are at least two Freedom models missing. While we could easily add those, the list will become out of date again in the future when Divesoft adds some more new models. That's the future proof problem @glance- is referring to: your code will need an update whenever a new serial number becomes available, while the parsing itself doesn't need any changes.

Your proposed fix is an improvement in the sense that it will no longer fail to import V1 files, but the model detection will still not always be accurate. It's still completely wrong for V1 files (and will even crash the application when trying to import a V1 file that only contains the header) and also doesn't work for V2 files without a serial number. 

@glance- and I consider it a major bug when a DLF file, which can be parsed by libdivecomputer, fails to import in subsurface. Being able to display the correct dive computer model is less critical for us and is at most a small cosmetic bug. May still be annoying, but it doesn't prevent users from importing their dives. You appear to have different priorities here and consider that last issue as the most important bug and/or don't care about V1 files?

I certainly understand why you're trying to guess and display the correct model. As long as that doesn't result in rejecting DLF files which can be parsed without any problems, I'm even okay with that. But don't blame us if it's still not 100% reliable or even correct, because that's what we're trying to tell you.

> If you read the source of libdivecomputer, you will notice that the parser in libdivecomputer does not currently check or fix the model of Divesoft dive computer that is set, it just accepts it even if it is incorrect. So setting an incorrect model to start with will result in a bug and an incorrect dive computer being displayed.

As I explained before, the model number isn't needed in the parser and thus doesn't get passed at all. That's on purpose because it's a feature, not a bug. Divesoft did a good job here and used a data format that's completely independent of the model. That's exactly how it should be!

Why should libdivecomputer try to guess the model, especially when it doesn't need this information? Note that libdivecomputer isn't even able to always reliably identify the correct model, because in some cases the information simply isn't present.

mikeller:
> The list of serial number prefixes in your proposed fix is still incomplete. There are at least two Freedom models missing. While we could easily add those, the list will become out of date again in the future when Divesoft adds some more new models.

Can you please point me to the user reports about these, I am keen to add them.

> That's the future proof problem @glance- is referring to: your code will need an update whenever a new serial number becomes available, while the parsing itself doesn't need any changes.

I really can not follow your argument here - the code proposed in #4510 is just as much 'not future proof' - when a new model ID becomes available it will NOT give the user the correct model name - just like the code in this pull request. The only difference is that this code is giving the user the information in the cases where we have it, instead of just ignoring it.

> Your proposed fix is an improvement in the sense that it will no longer fail to import V1 files, but the model detection will still not always be accurate. It's still completely wrong for V1 files (and will even crash the application when trying to import a V1 file that only contains the header) and also doesn't work for V2 files without a serial number.

You seem to be very worried about supporting V1 files - I am happy to investigate support for properly detecting them. Can you please point me to where I can obtain the software that is generating them, or how a current diver would obtain them?
And the same goes for files without a serial number - how can a diver generate one of these?

> @glance- and I consider it a major bug when a DLF file, which can be parsed by libdivecomputer, fails to import in subsurface. Being able to display the correct dive computer model is less critical for us and is at most a small cosmetic bug.

I think this is a difference in perspective - as a technical diver most of my dives will have 2 or even 3 profiles from different dive computers. So being able to easily distinguish which dive computer's profile I am looking at is way more than a 'cosmetic issue' for me as a technical diver.
And I suspect that, if anything, this will be exacerbated by the fact that in all likelihood, a lot of divers using the Divesoft Liberty CCR will be using Divesoft Freedom+ dive computers, in order to take advantage of the unified UI between CCR and dive computer.

jefdriesen:
> Can you please point me to the user reports about these, I am keen to add them.

There are no user reports. It's based on info I received from Divesoft. The missing ones are 7011 and 7044.

> I really can not follow your argument here - the code proposed in #4510 is just as much 'not future proof' - when a new model ID becomes available it will NOT give the user the correct model name - just like the code in this pull request. The only difference is that this code is giving the user the information in the cases where we have it, instead of just ignoring it.

In the current code, the model number is used for two different purposes: for passing to the parser and for displaying the model name. Since it's not needed for the first case, I proposed to remove that code. Code that isn't there, doesn't need updates. @glance- made an extra suggestion to take care of the display name.

I don't mind trying to do the detection anyway, but at the moment that code doesn't handle the V1 format.

> You seem to be very worried about supporting V1 files - I am happy to investigate support for properly detecting them.

I think it's important to keep supporting V1 files. Remember not so long ago, the V1 format was the only format subsurface supported. The V2 format is still relative new and there will still be plenty of users with V1 files. See the bug reports that were already mentioned earlier.

> Can you please point me to where I can obtain the software that is generating them, or how a current diver would obtain them? And the same goes for files without a serial number - how can a diver generate one of these?

They are generated by all Divesoft dive computers which haven't upgraded to the new format. I don't even know if all models can upgrade to newer firmware with the V2 format, so it's possible that some users with an older model are stuck with V1 forever.

The missing serial number should only happen for the factory test dives, so that's less critical.

> > @glance- and I consider it a major bug when a DLF file, which can be parsed by libdivecomputer, fails to import in subsurface. Being able to display the correct dive computer model is less critical for us and is at most a small cosmetic bug.
> 
> I think this is a difference in perspective - as a technical diver most of my dives will have 2 or even 3 profiles from different dive computers. So being able to easily distinguish which dive computer's profile I am looking at is way more than a 'cosmetic issue' for me as a technical diver. And I suspect that, if anything, this will be exacerbated by the fact that in all likelihood, a lot of divers using the Divesoft Liberty CCR will be using Divesoft Freedom+ dive computers, in order to take advantage of the unified UI between CCR and dive computer.

Sure, I certainly understand your argument. But that doesn't make the display name more important than being able to import the dive in the first place! Right now users simply can't import their V1 files at all. I'm pretty sure that's more important to them than a correct display name.

mikeller:
> > Can you please point me to the user reports about these, I am keen to add them.
> 
> There are no user reports. It's based on info I received from Divesoft. The missing ones are 7011 and 7044.

Ok, I'll contact them to get a complete list.

> > You seem to be very worried about supporting V1 files - I am happy to investigate support for properly detecting them.
> 
> I think it's important to keep supporting V1 files. Remember not so long ago, the V1 format was the only format subsurface supported. The V2 format is still relative new and there will still be plenty of users with V1 files. See the bug reports that were already mentioned earlier.

Do you have an approximate idea for when Divesoft switched from V1 to V2?
I think I tested this code against all user bug reports mentioned in this pull request, and I could not find a single one that mentioned a V1 log file, or that contained a log file that could not be processed by this change - in fact all but one of them could be processed by the changes made in https://github.com/subsurface/subsurface/pull/4367.

Can you point out the specific user bug report that references a V1 file, so that I can contact the user?

> > Can you please point me to where I can obtain the software that is generating them, or how a current diver would obtain them? And the same goes for files without a serial number - how can a diver generate one of these?
> 
> They are generated by all Divesoft dive computers which haven't upgraded to the new format. I don't even know if all models can upgrade to newer firmware with the V2 format, so it's possible that some users with an older model are stuck with V1 forever.

So the V1 files can be obtained by attaching the Divesoft dive computer as a storage device and downloading the logs directly from the dive computer?

> Sure, I certainly understand your argument. But that doesn't make the display name more important than being able to import the dive in the first place! Right now users simply can't import their V1 files at all. I'm pretty sure that's more important to them than a correct display name.

Agreed. If you can point me to a user who has this problem with their current hardware I am keen to work with them to get them supported.

jefdriesen:
> Ok, I'll contact them to get a complete list.

There is no need to contact them again and ask for info we already have. I already listed the missing prefixes in my previous reply. For completeness, the complete list is:

7011 Freedom "old" [HW 2.x]
7044 Freedom "classic" [HW 3.x]
7273 Freedom+ [HW 4.x]
7185 Freedom+ [HW 5.x]

7026 eCCR Liberty [HW 2.x]
7203 eCCR Liberty [HW 2.x]

> Do you have an approximate idea for when Divesoft switched from V1 to V2?

No, I don't know.

> I think I tested this code against all user bug reports mentioned in this pull request, and I could not find a single one that mentioned a V1 log file, or that contained a log file that could not be processed by this change - in fact all but one of them could be processed by the changes made in #4367.
> 
> Can you point out the specific user bug report that references a V1 file, so that I can contact the user?

I have many V1 files available, but most of them are not from public subsurface bug reports. If you want one that's publicly available, you can use the one that @glance- already provided: https://github.com/subsurface/subsurface/pull/4510#issuecomment-2825618177

BTW, if you discard the profile with
```
dd if=00000023.DLF of=header-only.DLF bs=32 count=1
```
and then try to import the resulting DLF file in subsurface, it will crash the application in your model detection code. The libdivecomputer parser doesn't have any problems with this input:
```
./dctool -f freedom parse -o /dev/stdout header-only.DLF 2> /dev/null

<device>
<dive>
<number>1</number>
<size>32</size>
<datetime>2018-09-08 11:16:52</datetime>
<divetime>02:12</divetime>
<maxdepth>1.80</maxdepth>
<temperature type="minimum">17.4</temperature>
<gasmix>
   <he>28.0</he>
   <o2>20.0</o2>
   <n2>52.0</n2>
   <usage>diluent</usage>
</gasmix>
<divemode>ccr</divemode>
<decomodel>buhlmann</decomodel>
<salinity density="1000.0">fresh</salinity>
<atmospheric>1.00740</atmospheric>
</dive>
</device>
```

> So the V1 files can be obtained by attaching the Divesoft dive computer as a storage device and downloading the logs directly from the dive computer?

Yes. Only the Freedom+ and Liberty have bluetooth support.

> Agreed. If you can point me to a user who has this problem with their current hardware I am keen to work with them to get them supported.

As @glance- already [pointed out](https://github.com/subsurface/subsurface/pull/4538#issuecomment-3001200136), the following report is most likely from a subsurface user with V1 files:

https://groups.google.com/g/subsurface-divelog/c/WOP00QlYOWM/m/4Nc6XatlDgAJ

mikeller:
> > Ok, I'll contact them to get a complete list.
> 
> There is no need to contact them again and ask for info we already have. I already listed the missing prefixes in my previous reply. For completeness, the complete list is:

It's ok, don't worry, I am already talking to @jakubhecko to get this resolved.

glance-:
This has been stalled for quite some time, and #4510 is ready to be merged, so my suggestion is to pick than one instead of having this stalled for any longer.

We can always revisit other alternatives in new PR's.

jakubhecko:
After a brief testing of the latest build I am quite happy with the results. The Subsurface parsed everything I was able to throw at it (Import / Import log files), including logs with invalid or missing serial numbers. Thanks for the fix !
> **Artifacts:** [Subsurface-Windows-6.0.5425-patch.1.pull-request.fix-divesoft-file-import-2](https://github.com/subsurface/subsurface/suites/43736391554/artifacts/3792890294) _**WARNING:** Use at your own risk._

mikeller:
@jakubhecko Thanks for checking. I have been able to get back to working on this, and I am preparing a change that addresses the concerns that have been raised about properly recognising V1 log files and files generated by factory test dives. This should be ready in the next week or so.

mikeller:
@jakubhecko: This is ready for review now - I have added some of the sample logs provided by you for testing.

PR #4543, opened 2025-06-23
Fix the purging of the received packets queue

jefdriesen:
These changes should help fixing the BLE problems with the Mares dive computers.

- [x] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/qt-ble.cpp
+++ b/core/qt-ble.cpp
@@ -325,13 +325,6 @@ dc_status_t BLEObject::write(const void *data, size_t size, size_t *actual)
 {
 	if (actual) *actual = 0;
 
-	if (!receivedPackets.isEmpty()) {
-		report_info(".. write HIT with still incoming packets in queue");
-		do {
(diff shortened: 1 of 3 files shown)


core/qt-ble.cpp
```diff
@@ -371,6 +364,14 @@ dc_status_t BLEObject::poll(int timeout)
 	return DC_STATUS_SUCCESS;
 }
 
+dc_status_t BLEObject::purge(dc_direction_t direction)
+{
+	if (direction & DC_DIRECTION_INPUT) {
```
mikeller: The [coding style](https://github.com/subsurface/subsurface/blob/master/CODINGSTYLE.md) in Subsurface is that `if` with a single expression does not use curly braces.

mikeller (approved):

PR #4559, opened 2025-07-18
Desktop: Planner Reverse Profile with Menu Action.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

In the planner, allow the user to mirror the profile, i.e. add copies of the waypoints in reverse order. The UI is a second button on top of the dive planner table and the Ctrl-M keyboard shortcut.

This PR supersedes #3842 and implements a suggestion by Alvaro A. on the google group.

This PR supersedes #3843 and makes the functionality available in a menu action.

### Changes made:

1. Planner: Feature to add reverse profile to the end of a plan;
2. Use a custom menu to make the 'reverse profile' action accessible;
3. Remove unnecessary calls to updateDiveProfile in planner.

### Additional information:

image


tcanabrava:
mmmm, I don't really like it the way the menu is attached. There's a way to do this without changing mainwindow at all:

on DivePlan activate:

```
     auto *mainWindow = qobject_cast<QMainWindow*>(parent());
     auto *menuBar = mainWindow->menu();
     // add the new action here.
```

on DivePlan deactivate:

```
     auto *mainWindow = qobject_cast<QMainWindow*>(parent());
     auto *menuBar = mainWindow->menu();
     // remove the new action here.
```
but I also think that *maybe* the `Add` button could have a little toggle for reverse mode.

qt-models/diveplannermodel.cpp
```diff
@@ -58,7 +58,6 @@ void DivePlannerPointsModel::removeSelectedPoints(const std::vector<int> &rows)
 {
 	removePoints(rows);
 
-	updateDiveProfile();
```
tcanabrava, line 61: Those were added so you could see the profile changed in real time after moving /adding or removing points, did you test this before removing this line?

mikeller:
> mmmm, I don't really like it the way the menu is attached. There's a way to do this without changing mainwindow at all:
> 
> on DivePlan activate:
> 
> ```
>      auto *mainWindow = qobject_cast<QMainWindow*>(parent());
>      auto *menuBar = mainWindow->menu();
>      // add the new action here.
> ```
> 
> on DivePlan deactivate:
> 
> ```
>      auto *mainWindow = qobject_cast<QMainWindow*>(parent());
>      auto *menuBar = mainWindow->menu();
>      // remove the new action here.
> ```

This does look a lot better. I am not very familiar with Qt, hence my solution is a bit pedestrian. Do you have any hints for how to make the custom menu show up to the left of the 'Help' menu?

qt-models/diveplannermodel.cpp
mikeller: This part was originally implemented by @jwobser - not sure if he tested the 'live updates'?

mikeller: These are indeed not needed because of this: https://github.com/subsurface/subsurface/pull/4559/files#diff-e31fe99db95b3a34a37cb0700a72be8e9f588f8403cdf8b82f34dc48bec5772bR514

PR #4568, opened 2025-08-03
Implement the backend dive planner functionality for mobile

fnightfox:

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Enable dive plan setup, dive plan edit, and dive plan view to do the following: 

Edit: The primary  body of the UI work. Contains all of the UI code to add cylinders and segments to the plan, links to the plan settings and preview and save functionality. 

Setup: Contains plan specific settings

View: Used to generate a simplified view of the calculated plan. Because of how the UI works, a true depth profile cannot be inserted here easily, but a simplified version is. The true profile is added once the plan is saved

Backend functions for plan calculation are added. These additions only function to create the backend objects and pass them to existing planning functions.

### Changes made:
1) Add mobile UI and implementation of dive planning functionality on mobile
2) Add a simple gas calculator to calculate MOD and EAD at MOD

### Additional information:
Some simple plans have been run to test in OC mode, and output seems similar to desktop. CCR modes are mostly untested and should be looked at by someone with more experience with CCR planning.

### Documentation change:
**TODO**
Developer wants feedback before user manual is updated to include planning functionality

### DIFF
--- a/mobile-widgets/qml/DivePlannerEdit.qml
+++ b/mobile-widgets/qml/DivePlannerEdit.qml
@@ -12,18 +12,81 @@ Kirigami.ScrollablePage {
 	property string pressureUnit: (Backend.pressure === Enums.BAR) ? qsTr("bar") : qsTr("psi")
 	property string depthUnit: (Backend.length === Enums.METERS) ? qsTr("m") : qsTr("ft")
 
+	property string planNotes: ""
+	property string maxDepth: ""
+	property string duration: ""
+	property var profileData: []	
+
 	// --- Data Models ---
 	ListModel { id: cylinderListModel }
 	ListModel { id: segmentListModel }
 	property var gasNumberModel: []
 
 	// --- Functions ---
-		function updateGasNumberList() {
+	function updateGasNumberList() {
 		var newList = [];
 		for (var i = 0; i < cylinderListModel.count; i++) {
 			newList.push(qsTr("Gas %1").arg(i + 1));
 		}
 		gasNumberModel = newList;
+	}
+
+	function updateLivePlanInfo() {
+		if (visible && segmentListModel.count > 0 && cylinderListModel.count > 0) {
+			var cylinderData = []
+			for (var i = 0; i < cylinderListModel.count; i++) {
+				var item = cylinderListModel.get(i)
+				cylinderData.push({
+										"type": item.type, "mix": item.mix,
+										"pressure": item.pressure, "use": item.use
+									})
+			}
+
+			var segmentData = []
+			if (Backend.drop_stone_mode && segmentListModel.count > 0) {
+				var descentRate = Backend.descrate;
+				var firstSegment = segmentListModel.get(0)
+				var descentDuration = 1
+				if (descentRate > 0) {
+                    descentDuration = Math.ceil(firstSegment.depth / descentRate);
+				}
+				segmentData.push({
+										"depth": firstSegment.depth,
+										"duration": descentDuration,
+										"gas": firstSegment.gas
+									})
+			}
+			for (var j = 0; j < segmentListModel.count; j++) {
+				var item = segmentListModel.get(j)
+				segmentData.push({
+									"depth": item.depth,
+									"duration": item.duration,
+									"gas": item.gas
+								})
+			}
+			var planResult = Backend.divePlannerPointsModel.calculatePlan(
+				cylinderData, segmentData, planDate.text, planTime.text,
+				diveModeBox.currentIndex, false
+			)
+			// Handle planResult
+			planNotes = planResult.notes
+			maxDepth = planResult.maxDepth
+			duration = planResult.duration
+			profileData = planResult.profile
+		}
+
+
+	}
+
+    onVisibleChanged: {
+        // This code runs every time the page becomes visible
+        if (visible) {
+            updateLivePlanInfo()
+        }
+	}
+
+	onProfileDataChanged: {
+		profileCanvas.requestPaint();
 	}
 
 	Component.onCompleted: {
@@ -35,18 +98,25 @@ Kirigami.ScrollablePage {
 		});
 
 		segmentListModel.append({
-			"depth": (Backend.length === Enums.METERS) ? 20 : 60,
-			"duration": 1,
+			"depth": (Backend.length === Enums.METERS) ? 14 : 45,
+			"duration": 20,
 			"gas": 0 // Default to the first cylinder (index 0)
 		});
 
 		updateGasNumberList();
+		updateLivePlanInfo();
 	}
 
 	Connections {
 		target: cylinderListModel
-		onRowsInserted: updateGasNumberList()
-		onRowsRemoved: updateGasNumberList()
+		onRowsInserted: {
+			updateGasNumberList();
+			updateLivePlanInfo();
+		}
+		onRowsRemoved: {
+			updateGasNumberList();
+			updateLivePlanInfo();
+		}
 	}
 
 	ColumnLayout {
@@ -54,7 +124,6 @@ Kirigami.ScrollablePage {
 		spacing: Kirigami.Units.gridUnit
 		Layout.margins: Kirigami.Units.gridUnit
 
-// --- NEW: Plan Details Section ---
 		Controls.Label {
 			text: qsTr("Plan Details")
 			font.bold: true
@@ -91,6 +160,85 @@ Kirigami.ScrollablePage {
 			}
 		}
 
+		Canvas {
+			id: profileCanvas
+			Layout.fillWidth: true
+			Layout.preferredHeight: Kirigami.Units.gridUnit * 20
+
+			onPaint: {
+				if (profileData.length < 2) return;
+
+				var ctx = getContext("2d");
+				ctx.reset();
+
+				// --- Find max values and define geometry ---
+				var maxTime = 0;
+				var maxDepthMm = 0;
+				for (var i = 0; i < profileData.length; i++) {
+					if (profileData[i].time > maxTime) maxTime = profileData[i].time;
+					if (profileData[i].depth > maxDepthMm) maxDepthMm = profileData[i].depth * 1.1;
+				}
+				if (maxTime === 0 || maxDepthMm === 0) return;
+
+				var padding = 30; // Increased padding for labels
+				var drawWidth = width - (padding * 1.5);
+				var drawHeight = height - (padding * 1.5);
+				var xScale = drawWidth / maxTime;
+				var yScale = drawHeight / maxDepthMm;
+
+				ctx.strokeStyle = "lightgray";
+				ctx.lineWidth = 0.5;
+				ctx.fillStyle = "gray";
+				ctx.font = "10px sans-serif";
+
+				var numDepthLines = 4;
+				ctx.textAlign = "right";
+				ctx.textBaseline = "middle";
+				for (var i = 1; i <= numDepthLines; i++) {
+					var y = padding + (i * drawHeight / numDepthLines);
+					ctx.beginPath();
+					ctx.moveTo(padding, y);
+					ctx.lineTo(width - padding / 2, y);
+					ctx.stroke();
+
+					var depthValMm = (i / numDepthLines) * maxDepthMm;
+					var depthInUserUnits = (Backend.length === Enums.METERS) ? depthValMm / 1000 : depthValMm / 304.8;
+					ctx.fillText(depthInUserUnits.toFixed(0), padding - 5, y);
+				}
+
+				var numTimeLines = Math.floor(maxTime / 60 / 5); // A line every 5 minutes
+				ctx.textAlign = "center";
+				ctx.textBaseline = "top";
+				for (var j = 1; j <= numTimeLines; j++) {
+					var timeVal = j * 5 * 60; // Every 5 minutes
+					if (timeVal > maxTime) continue;
+					var x = padding + (timeVal * xScale);
+					ctx.beginPath();
+					ctx.moveTo(x, padding);
+					ctx.lineTo(x, height - padding);
+					ctx.stroke();
+					ctx.fillText((timeVal / 60).toFixed(0), x, height - padding + 5);
+				}
+
+				ctx.strokeStyle = Kirigami.Theme.highlightColor;
+				ctx.lineWidth = 2;
+				ctx.beginPath();
+				var firstPoint = profileData[0];
+				var startX = padding + (firstPoint.time * xScale);
+				var startY = padding + (firstPoint.depth * yScale);
+				ctx.moveTo(startX, startY);
+
+				for (var j = 1; j < profileData.length; j++) {
+					var point = profileData[j];
+					var x = padding + (point.time * xScale);
+					var y = padding + (point.depth * yScale);
+					ctx.lineTo(x, y);
+				}
+				ctx.stroke();
+			}
+		}
+
+
 		// --- 1. Cylinders Section ---
 		RowLayout {
 			Controls.Label {
@@ -109,6 +257,7 @@ Kirigami.ScrollablePage {
 							"pressure": (Backend.pressure === Enums.BAR) ? 200 : 3000,
 							"use": 0 // Default to OC_GAS
 						});
+						updateLivePlanInfo();
 					}
 			}
 		}
@@ -153,7 +302,10 @@ Kirigami.ScrollablePage {
 					id: mixField
 					Layout.preferredWidth: Kirigami.Units.gridUnit * 4
 					text: mix
-					onEditingFinished: cylinderListModel.setProperty(index, "mix", text)
+					onEditingFinished: {
+						cylinderListModel.setProperty(index, "mix", text);
+						updateLivePlanInfo();
+					}
 					validator: RegExpValidator { regExp: /(EAN100|EAN\d\d|AIR|100|\d{1,2}|\d{1,2}\/\d{1,2})/i }
 					onActiveFocusChanged: cylinderListView.interactive = !activeFocus
 				}
@@ -168,7 +320,10 @@ Kirigami.ScrollablePage {
 					Layout.fillWidth: true
 					text: pressure.toString()
 					validator: IntValidator { bottom: 0; top: 10000 }
-					onEditingFinished: cylinderListModel.setProperty(index, "pressure", Number(text))
+					onEditingFinished: {
+						cylinderListModel.setProperty(index, "pressure", Number(text));
+						updateLivePlanInfo();
+					}
 					onActiveFocusChanged: cylinderListView.interactive = !activeFocus
 				}
 				TemplateButton {
@@ -176,7 +331,10 @@ Kirigami.ScrollablePage {
 					font.bold: true
 					Layout.preferredWidth: Kirigami.Units.gridUnit * 2
 					enabled: cylinderListModel.count > 1
-					onClicked: cylinderListModel.remove(index)
+					onClicked: {
+						cylinderListModel.remove(index);
+						updateLivePlanInfo();
+					}
 				}
 			}
 		}
@@ -200,6 +358,7 @@ Kirigami.ScrollablePage {
 							"duration": 10,
 							"gas": lastSegment.gas
 						});
+						updateLivePlanInfo();
 					}
 				}
 			}
@@ -236,7 +395,10 @@ Kirigami.ScrollablePage {
 					Layout.fillWidth: true
 					text: depth.toString()
 					validator: IntValidator { bottom: 0; top: 300 }
-					onEditingFinished: segmentListModel.setProperty(index, "depth", Number(text))
+					onEditingFinished: {
+						segmentListModel.setProperty(index, "depth", Number(text));
+						updateLivePlanInfo();
+					}
 					onActiveFocusChanged: segmentListView.interactive = !activeFocus
 				}
 
@@ -244,7 +406,10 @@ Kirigami.ScrollablePage {
 					Layout.fillWidth: true
 					text: duration.toString()
 					validator: IntValidator { bottom: 1; top: 999 }
-					onEditingFinished: segmentListModel.setProperty(index, "duration", Number(text))
+					onEditingFinished: {
+						segmentListModel.setProperty(index, "duration", Number(text));
+						updateLivePlanInfo();
+					}
 					onActiveFocusChanged: segmentListView.interactive = !activeFocus
 				}
 
@@ -257,6 +422,7 @@ Kirigami.ScrollablePage {
 					onCurrentIndexChanged: {
 						if (currentIndex !== -1)
 						   segmentListModel.setProperty(index, "gas", currentValue)
+						updateLivePlanInfo();
 					}
 				}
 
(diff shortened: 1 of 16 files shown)


mikeller:
This is great, thank you so much for this work!
I have been keen to get a mobile planner going for a long time, and keen to help you with seeing this through.

atdotde:
Thank you so much for this! A mobile implementation of the dive planner has been requested by many for years. But so far, nobody dared to take up the challenge to code a decent UI for it in QML.

Now that the start is done, there are of course a million ideas of how it could be improved. My personal goal would be to get closer to the desktop version in one in my mind crucial aspect: What sets the Subsurface dive planner apart from all others on the market is that any changes in data (depth, duration, gas choices, deco model parameters) have an instantaneous effect on the plan. I understand, that in QML it is much harder to have a profile plot that updates immediately, but maybe it could at least be done with a text representation of the dive? Can we have the dive plan on the same page as the plan data (waypoints)?

Another, much simpler improvement would be to have more sensible default values for the settings as well as having the "drop to first depth" option from the desktop.

But let me emphasise again: This as it is is already a great! Thanks so much!

fnightfox:
> Thank you so much for this! A mobile implementation of the dive planner has been requested by many for years. But so far, nobody dared to take up the challenge to code a decent UI for it in QML.
> 
> Now that the start is done, there are of course a million ideas of how it could be improved. My personal goal would be to get closer to the desktop version in one in my mind crucial aspect: What sets the Subsurface dive planner apart from all others on the market is that any changes in data (depth, duration, gas choices, deco model parameters) have an instantaneous effect on the plan. I understand, that in QML it is much harder to have a profile plot that updates immediately, but maybe it could at least be done with a text representation of the dive? Can we have the dive plan on the same page as the plan data (waypoints)?
> 
> Another, much simpler improvement would be to have more sensible default values for the settings as well as having the "drop to first depth" option from the desktop.
> 
> But let me emphasise again: This as it is is already a great! Thanks so much!

At first I was having trouble with the drop to first depth mode, but I just figured out a way to do it (since the desktop version can directly manipulate the segment list from the settings and I cannot, it was causing problems. I added that back in, and I will work on a view that contains everything in the preview screen on the main planning screen. I'm worried it will be too busy for a mobile app, but I would like feedback on that. I will try to get that pushed within a few hours.

fnightfox:
Here is what the updates screens look like, they are interactive/update once the text fields are no longer focused. Drop to first depth is again enabled. I would like peoples thoughts on if that screen is too crowded. @mikeller , since this is my first time contributing to this repo, what process/items would you like to see in order to get this merged?
Screenshot_20250804-210010

Screenshot_20250804-210019

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -51,65 +53,61 @@ TemplatePage {
 					font.bold: true
 				}
 				TemplateLabel {
-					text: qsTr("below 75% avg. depth")
+					text: qsTr("below 75% avg. depth (%1)").arg(speedUnit)
```
mikeller: This seems to be a much better way to integrate units than processing them as a part of the edit's content. But `[<units>]` seems to be more aligned with  common use in literature than `(<units>)`.

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -221,41 +217,7 @@ TemplatePage {
...
-					value: Backend.planner_gfhigh
-					textFromValue: function (value, locale) {
-						return value + volumeUnit
-					}
-					onValueModified: {
-						Backend.planner_gfhigh = value
-					}
-				}
-
+				//Gradient Factors pulled from main application settings
```
mikeller: In my view it makes sense to have separate settings for the gradient factors used in planner, and the ones applied to the profiles of all dives in the log - this allows users to play 'what ifs' in planner without having to remember to reset the gradient factors when they are done, in order to have their logged dives show in the way they want.

In addition to this, changing of gradient factors for the dive log seems to be borked at the moment:

Image

mobile-widgets/qml/DivePlannerEdit.qml
```diff
@@ -1,22 +1,576 @@
...
+				Layout.fillWidth: true
+				placeholderText: qsTr("Time")
+				// Use a standard, unambiguous format
+				text: Qt.formatTime(new Date(), "hh:mm:ss")
+			}
+			Controls.Label {
+				text: qsTr("Dive Mode")
+				verticalAlignment: Text.AlignVCenter
+			}
+			TemplateComboBox {
```
mikeller, line 177: This is doubled up at the moment - there is another edit for this in the planner settings.

But I think having this here makes more sense, as most CCR users will also be doing and planning OC dives.

From what I can see CCR mode does not currently work for the mobile planner - likely because the gas type (OC gas / diluent) is not hooked up correctly, and the planner expects to be told for each segment of a CCR dive if it was done on CCR or OC (bailout), through the selection of the appropriate gas.

Another aspect of this is that for OC dives, 'diluent' should not be available as gas type. And 'oxygen' doesn't really do anything in the planner, it's just there to allow CCR divers to enter and track their complete gas plan in the desktop version. For deco on 100% O2 the correct gas is '100% O2 OC Gas'.

My idea to deal with this for now would be to limit the mobile planner to OC only at the moment, and get it in front of users and get feedback. We can then start working on the additional constraints that need to be built in for planning CCR dives.

mobile-widgets/qml/DivePlannerEdit.qml
```diff
@@ -1,22 +1,576 @@
...
+		onTriggered: {
+			showPage(divePlannerSetupWindow)			
+		}
+	}
+	actions.main: Kirigami.Action {
+		icon {
+			name: state = ":/icons/document-save.svg"
+			color: subsurfaceTheme.primaryColor
+		}
+		text: "Save"
```
mikeller: The current behaviour of the gas / segment list is that the profile preview does not update when a value is changed, until another edit is selected, making this a little bit unintuitive.

Should we add a 'refresh' button in here to give users a 'go-to' to get the profile to update?

mobile-widgets/qml/main.qml
```diff
@@ -309,6 +309,22 @@ Kirigami.ApplicationWindow {
 					globalDrawer.close()
 				}
 			},
+			Kirigami.Action {
```
mikeller: Considering most of the users of Subsurface are probably recreational divers who are not going to be interested in using the planner or the gas calculator, I am wondering if there is a better way to add this to the menu that has less users wander into it and not understand what it does, cutting down on support requests. Ideas:
- add this into a submenu 'Technical Diving' or similar;
- add a configuration setting that allows the user to select between 'recreational mode' and 'technical mode', and only show this in 'technical mode'.

mikeller (commented):
In general I think this looks very good - thank you for all your effort!

See my comment about initially restricting this to OC planning.

In general I am wondering if it would make sense to change the ordering of the planner page to make it follow the flow of planning a dive:

- have the gas / segment list up top, as this is where planning starts;
- then the profile preview to allow users to validate the plan;
- then the actual text form plan;
- have 'Save to logbook' as a button on the page, either above or below the text form of the plan.

Some minor bugs that I found:
- 'calculate variants' does not seem to be working;
- on an older android test device, the gas calculator seems to be failing:

Image

atdotde:
What @mikeller says.

I see it as it's pretty close to a state that can be shown to users that things are moving towards a planner. Before merging you should weed out a few minor things like removing commented out regions and fix some white space issues (you will see them when you git show your commits).

On the other hand, I would say, the mobile implementation of the planner is far from finished in particular from a usability perspective. We should really aim to make the typical use case of a dive planner on a mobile device (I would guess not at home but at a dive site checking a few last things), in particular taking into account the very limited screen estate and the touch interface.

Maybe you want to have a look at what I did on [mobileImplementDivePlan](https://github.com/atdotde/subsurface/tree/mobileImplementDivePlan). I picked some low hanging fruits:

1) Centered the "+" signs on the buttons (on the long run those should probably be icons).

2) Replaced some text inputs with tumber wheels. I find those easier to use on a touch interface. But more experimentation is needed. For example one could also have one tumbler per digit. I don't know if that makes things more confusing or simpler.

3) I made the profile smaller to save some screen estate. Maybe what really should be shown on top would be the total runtime as I think this is what may try to optimize when planning deco.

4) I moved the planner and gas calculator to a less prominent place in the main menu as I would at the moment more to be in a proof of concept stage than to be actually useful.

5) I have not really looked at the gas calculator in any detail. On desktop, we have the MOD and END calculations integrated in the gas selection (including the option to give "*" for the O2 and He fractions meaning to automatically pick best mix for the dive. Maybe there is a clever way to do similar things on mobile? For a gas calculator, I would rather add a gas blending calculator (my own go at this is https://github.com/atdotde/realblender but I never got around to implement that in QML).

What do you think?

Maybe we should move this discussion to the Subsurface mailing list https://mailman.subsurface-divelog.org/hyperkitty/list/subsurface@subsurface-divelog.org/

fnightfox:
Thank you for the reviews!

I will address comments all at once as I see them in this post.

Edit page (main planner page):
-About OC only vs CCR, I am perfectly fine removing the CCR setting, as again, I do not currently have a CCR unit and have not trained on how to plan those dives, so it is hard for me to test properly.
-I can see how changing the ordering on the page would make sense (from profile -> gas -> segment -> notes to gas -> segment ->profile -> notes)
-The save button is on the page at the bottom already, it should be the primary action button.
-I will check the calculate variants button in the code to see if I can find what's going on with it
-Do you think I should leave the dropdowns/columns in place for gas use and dive mode, so we can get feedback on the locations, just disable them? Or should they be removed?
-The profile changes if focus is lost on the segment or gas entries (either by selecting something else, or by pressing a button) or if you hit the checkmark button on the keyboard (I assume this is an enter key). I have been trying to get it to update as values are entered, but so far that has not worked. I will think about the best solution to this.

Gas Calculator:
-I have no idea what's up with older versions showing that weird, but I will change the layout to be a different type (not a TableView object), so maybe that will fix it

Settings:
-I will change units to be [unit] vs (unit)
-I removed the gradient factors from the settings so that when the user sees their plan on the log they wont immediately see it go past their factors if the main app is different from the planner, but I can see the value in having it as a possibility. I will add them back.
-I will remove the Dive Mode select from the settings. (I will be removing it entirely for now, but it will return in the Edit page)

Main Menu:
-I will look into adding a technical diving submenu and gating this behind a setting under advanced, like the Developer options are.

General Comments:
-I will take a look at the linked mobileImplementDivePlan. From a quick glance, it does a good job cleaning up some areas, so I will probably grab some ideas from that. Thank you!
-My ultimate goal is to have a gas blending page, I put gas calculator there just for simplicity and to create a placeholder for added functionality. I might take a look at the blending items and see if they can be tossed in there as well.

I will take another crack at this over the weekend and hope to have some items polished by end of the 10th. I would be fine moving general discussion to the mailing list, but I would like comments directly relating to the code (or referencing lines of code, etc) to be here as well, at least.

Again, thank you everyone for looking at this. I hope we can get at least an initial implementation in before too long!

mikeller:
> -About OC only vs CCR, I am perfectly fine removing the CCR setting, as again, I do not currently have a CCR unit and have not trained on how to plan those dives, so it is hard for me to test properly.


I can help with adding the CCR functionality - I have recently done a rework of the CCR functionality in the desktop planner, to align it a bit more with how gases are managed for CCR in most current dive computers, and I think it will make sense to do it the same way in the mobile version.

> -The save button is on the page at the bottom already, it should be the primary action button.


Apologies, I was not very clear on that one. Based on the assumption that users will want to review the proposed profile for the planned dive before hitting 'Save', so I think it might make sense to have 'Save to logbook' as a button that is part of the page flow below the profile, and not an action button that is always showing. One concern I have with using an action button for this is that users will accidentally save incomplete planned dives to their logbook when they click on 'Save'.

> -I will check the calculate variants button in the code to see if I can find what's going on with it


Variants are calculated in a background thread - it might be that this is just not built in the mobile version for simplicity. Might be another one to disable to start with, and then add at a later stage.

> -Do you think I should leave the dropdowns/columns in place for gas use and dive mode, so we can get feedback on the locations, just disable them? Or should they be removed?


I'd remove them, or at least disable them - in the desktop version users can only choose OC gas when planning OC dives.

> General Comments:
> -My ultimate goal is to have a gas blending page, I put gas calculator there just for simplicity and to create a placeholder for added functionality. I might take a look at the blending items and see if they can be tossed in there as well.

Happy to support an 'MVP' approach that adds a basic gas / ppO2 table first, and then gradually builds it up to a full gas blender - as long as intermediate steps are done in a way that will not confuse users and lead to a lot of support requests.

mikeller (commented):
This starts to look really good, great work!

The gas planner is now working as intended even on my old Android testing phone.

I would probably move the 'Technical diving' menu down, below the location - this way we get all of the (more commonly used) recreational functions first, and then the more advanced stuff.

One thing that I noticed that could probably be improved is that right now it is possible to enter impossible gas mixes (like '16/99') - the planner already handles this by defaulting to air, but the impossible value keeps showing in the frontend. Maybe we can validate this and change it to the 'nearest' possible value (like '16/84'). The same goes for the gas planner. And seeing that we are already using the 'O2/He' notation in the planner, we could just make it into a component and re-use it in the gas planner.

The variants are still not populated.

I have not done anything to valideate the calculations, but since it's reusing the backend from the desktop version I am relatively confident that they will be accurate. Maybe @atdotde can help here?

fnightfox:
For the variations, I can see three relatively simple ways to handle it. 
1. Disable variations
2. Have it not displayed during the preview, but when it saves I can run a new synchronous version of the variations calculation (that will return a string).
3. This is similar to 2, but would be me changing the existing function to either emit the value or return it as a string (basically allow it to switch between sync and async operation). I am not sure exactly how difficult this would be, but I think it would be doable and it would reduce the duplicated code.

The menu location will be changed, and I will add a setting to enable Technical features (beta).

I will look into input validation

mikeller:
> For the variations, I can see three relatively simple ways to handle it.
> 
> 1. Disable variations
> 2. Have it not displayed during the preview, but when it saves I can run a new synchronous version of the variations calculation (that will return a string).
> 3. This is similar to 2, but would be me changing the existing function to either emit the value or return it as a string (basically allow it to switch between sync and async operation). I am not sure exactly how difficult this would be, but I think it would be doable and it would reduce the duplicated code.

I am not too fussed about the variations, I think 2. will be nice if calculating the variations can be made to work on mobile easily, and if not the 1. will be entirely sufficient.
I don't think there is much value in 3. - when users want a quick 'what if' plan they are probably not overly interested in the variations.

qt-models/diveplannermodel.cpp
```diff
@@ -1295,6 +1297,70 @@ void DivePlannerPointsModel::computeVariationsDone(QString variations)
 	emit calculatedPlanNotes(notes);
 }
 
+QString DivePlannerPointsModel::calculateVariationsSync(const struct diveplan &original_plan, const struct deco_state &ds_in)
+{
+    // nothing to do unless there's an original plan
```
mikeller: Coding style: Spaces instead of tabs.

mikeller (commented):
Thanks for adding the gas mix validation - this is much less confusing.

In retrospect, I am not so sure if making the 'Technical Diving' menu is such a good idea:
- where the menu is now it is 'out of the way' for recreational divers just wanting to log their dives;
- if we disable the menu by default then it is unlikely that technical divers will find it on their own, hurting uptake of the new functionality;
- if it is enabled by default then I think recreational divers will just ignore the menu, but not go and search for an option to disable it.

Apart from that, the only thing I can see that should be addressed is that the code changes use a mix of tabs and spaces for indentation - this is probably something that can easily be addressed in your IDE / linter settings. See https://github.com/subsurface/subsurface/blob/master/CODINGSTYLE.md#basic-style-rules for the conventions and an example configuration.

Another thing that I noticed is that currently, the gas calculator does not seem to take the 'oxygen is narcotic' planner preference into account - this could probably be addressed in a future change, for consistency.

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -186,7 +172,7 @@ TemplatePage {
 				}
 
 				TemplateLabel {
-					text: qsTr("Reserve gas")
+					text: qsTr("Reserve gas [%1]").arg(volumeUnit)
```
mikeller:
```suggestion
					text: qsTr("Reserve gas [%1]").arg(pressureUnit)
```

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -221,19 +206,16 @@ TemplatePage {
 						Backend.planner_deco_mode = Enums.BUEHLMANN
 					}
 				}
-
+				
 				TemplateLabel {
 					text: qsTr("GFLow")
```
mikeller:
```suggestion
					text: qsTr("GFLow [%%]")

Comment (mikeller) [reply mobile-widgets/qml/DivePlannerSetup.qml]: Same for GFHigh below.

File: mobile-widgets/qml/DivePlannerSetup.qml [hunk 9]
```diff
@@ -266,33 +245,35 @@ TemplatePage {
 				}
 
 				TemplateLabel {
-					text: qsTr("Conservatism level")
+					text: qsTr("Conservatism level (+)")
```
mikeller:
```suggestion
					text: qsTr("Conservatism level [+]")
```

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -266,33 +245,35 @@ TemplatePage {
...
 				TemplateCheckBox {
 					text: qsTr("Plan backgas breaks")
 					Layout.columnSpan: 2
-					checked: Backend.last_stop6m
+					checked: Backend.doo2breaks
 					onClicked: {
-						Backend.last_stop6m = checked
+						Backend.doo2breaks = checked
 					}
 				}
```
mikeller, line 279: 'Min switch time' and 'Surface segment' below should have '[min]' as units, and not '+'.

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -9,10 +9,12 @@ import org.kde.kirigami 2.4 as Kirigami
 
 TemplatePage {
 	title: qsTr("Dive planner setup")
+	id: divePlannerSetupWindow
 
 	property string speedUnit: (Backend.length === Enums.METERS) ? qsTr(" m/min") : qsTr(" ft/min")
 	property string volumeUnit: (Backend.volume === Enums.LITER) ? qsTr(" l/min") : qsTr(" cuft/min")
```
mikeller:
```suggestion
	property string volumeUnit: (Backend.volume === Enums.LITER) ? qsTr("L/min") : qsTr("cuft/min")
```

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -392,70 +377,77 @@ TemplatePage {
...
+					valueFromText: function(text) {
+						return Math.round(parseFloat(text) * 10);
+					}
 					onValueModified: {
 						Backend.sacfactor = value
 					}
 				}
 				TemplateLabel {
-					text: qsTr("Problem solving time")
+					text: qsTr("Problem solving time (min)")
```
mikeller:
```suggestion
					text: qsTr("Problem solving time [min]")
```

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -9,10 +9,12 @@ import org.kde.kirigami 2.4 as Kirigami
 
 TemplatePage {
 	title: qsTr("Dive planner setup")
+	id: divePlannerSetupWindow
 
 	property string speedUnit: (Backend.length === Enums.METERS) ? qsTr(" m/min") : qsTr(" ft/min")
```
mikeller, line 14:
```suggestion
	property string distanceUnit: (Backend.length === Enums.METERS) ? qsTr("m") : qsTr("ft")
	property string speedUnit: (Backend.length === Enums.METERS) ? qsTr("m/min") : qsTr("ft/min")
```

mobile-widgets/qml/DivePlannerSetup.qml
```diff
@@ -392,70 +377,77 @@ TemplatePage {
...
+						return Math.round(parseFloat(text) * 100)
 					}
 					onValueModified: {
-						Backend.decopo2 = value
+						Backend.decopo2 = value * .01
 					}
 				}
 				TemplateLabel {
-					text: qsTr("Best mix END")
+					text: qsTr("Best mix END [%1]").arg(speedUnit)
```
mikeller, line 441:
```suggestion
					text: qsTr("Best mix END [%1]").arg(distanceUnit)
```

mikeller (commented):
I did a comparison of plans calculated on the mobile version and on the desktop version, and they are pretty close but not 100% identical - not sure if the discrepancies are due to rounding or what: 
image

And while comparing the settings to the settings in the desktop version I found a bunch of settings that have incorrect units in the mobile version (from before you started improving them).

mobile-widgets/qml/DivePlannerSetup.qml
fnightfox: I'm not sure what is causing the discrepancy between the desktop and mobile versions, honestly. At first I thought maybe salinity, but when I added that to mobile it did not make a difference. Might just be rounding?

mikeller (approved):
Good find about the salinity - with the added setting of salinity the resulting run times actually match the desktop version, and the only discrepancy I can find is an 0.5% or so difference in calculated gas consumption. And I am happy to put this down to rounding discrepancies.

mikeller:
@fnightfox: We got this bit of feedback on our developer mailing list:
>  subsurface-mobile 6.0.5436 - Dark Theme:
>
> I'm trying out the dive planner.  My default is to use a dark theme and that made using the dive planner very difficult because gas mix & bar entries and dive segments are using black text on a black background.  The headers for the fields and the pulldown menus worked ok for a dark theme.
>
> Switched to a light theme, restarted subsurface, and now I'm going to start testing more....

fnightfox:
> @fnightfox: We got this bit of feedback on our developer mailing list:
> 
> > subsurface-mobile 6.0.5436 - Dark Theme:
> > I'm trying out the dive planner.  My default is to use a dark theme and that made using the dive planner very difficult because gas mix & bar entries and dive segments are using black text on a black background.  The headers for the fields and the pulldown menus worked ok for a dark theme.
> > Switched to a light theme, restarted subsurface, and now I'm going to start testing more....

I received the email and I have coded a solution. The required changes are in DivePlannerEdit.qml:

Kirigami.ScrollablePage -> TemplatePage
Each instance of Controls.Label -> TemplateLabel
each instance of Controls.TextField -> TemplateTextField
and the Controls.TextArea around line 512 needs the following line:
color: subsurfaceTheme.textColor

I can either create this patch with the feature I am working on now (the gas blending part of the calculator, which I am hoping I can get into PR this week, though I will talk more about this below), I can create a separate patch myself, or you can feel free to make these changes on the repo itself. I am fine with any option.

The gas blender is split into two modes: simple and full. Simple will assume unlimited pressure and volume for all source gasses, where full will allow you to specify each one as well as a cost per volume unit. The system attempts to minimize cost, if possible. I think I can have Simple done tonight or tomorrow, full will take a bit longer.

mikeller:
> I received the email and I have coded a solution. The required changes are in DivePlannerEdit.qml:

Good to know - I wasn't sureif you are on the developer mailing list, as this is not widely advertised any more.


> Kirigami.ScrollablePage -> TemplatePage Each instance of Controls.Label -> TemplateLabel each instance of Controls.TextField -> TemplateTextField and the Controls.TextArea around line 512 needs the following line: color: subsurfaceTheme.textColor
> 
> I can either create this patch with the feature I am working on now (the gas blending part of the calculator, which I am hoping I can get into PR this week, though I will talk more about this below), I can create a separate patch myself, or you can feel free to make these changes on the repo itself. I am fine with any option.

Thanks for the guidance, I can have a go at this.

> The gas blender is split into two modes: simple and full. Simple will assume unlimited pressure and volume for all source gasses, where full will allow you to specify each one as well as a cost per volume unit. The system attempts to minimize cost, if possible. I think I can have Simple done tonight or tomorrow, full will take a bit longer.

Both of these sound good - and I am happy to leave it up to you if you want to get the simple version done first to give users a chance for feedback, or if you want to drop both at once. And by all means, don't feel rushed. This is open source, and things happen when they happen.

For your information, we are currently in the process of doing a 'weekly' release, the first that includes the planner. This will also make the planner available to more than a handful of iOS users for the first time.

PR #4573, opened 2025-08-10
Match any horizontal whitespace in the time format test

### LINKED ISSUE #4572: [Bug]: Preferences chokes on UI default time format
### Description

Hi!

I tried to reopen https://github.com/subsurface/subsurface/issues/3849 for this, but I don't have permission for that.

With current git head, in a current Debian Trixie Plasma VM, trying to apply any change in the preferences dialog was triggering the "Non-standard character(s) in time format." warning - which was curious because I hadn't done anything to change the format in plasma, and Subsurface was set to use the UI default.

The UI default ostensibly looks like this: "h:mm Ap" - which it seemed like the check regex should match - except the UI is being Too Clever And Modern for us, and that isn't an ordinary space, it's U+202F (Narrow No-Break Space).

This change fixes that:

```
--- a/desktop-widgets/preferences/preferences_language.cpp
+++ b/desktop-widgets/preferences/preferences_language.cpp
@@ -98,7 +98,8 @@ void PreferencesLanguage::syncSettings()
        refreshSettings();
 
        QString qDateTimeWeb = tr("These will be used as is. This might not be what you intended. To avoid this warning wrap the literal parts in quotes (').\nSee https://doc.qt.io/archives/qt-4.8/qdatetime.html#fromString");
-       QRegularExpression timeStandardFormat("^([hHmszaApP\\s:\\.]|'[^']*')*$");
+       QRegularExpression timeStandardFormat("^([hHmszaApP\\h:\\.]|'[^']*')*$");
        if (!timeStandardFormat.match(ui->timeFormatEntry->currentText()).hasMatch())
                QMessageBox::warning(this, tr("Literal characters"),
                        tr("Non-standard character(s) in time format.\n") + qDateTimeWeb);
```
I can send that as a formal MR if you like, but it seemed a bit overkill for a one character change to replace \s with \h and match any horizontal whitespace.  It seems safe to assume qDateTime is ok with that (because if it's not this probably turns into a KDE/Qt bug too ;)

  Cheers,
  Ron

### Subsurface Version

git head

### Platform(s)

Linux

### Platform Version

Debian Trixie, KDE plasma

### Divecomputer Model

_No response_

### Steps to Reproduce

Click Apply in the preferences dialog with the offending default time format.

### Logs

```shell

```

### Mentions

_No response_

RonOnGithub:
The default time format for KDE Plasma on Debian Trixie is "h:mm Ap" where the space character is U+202F (Narrow No-Break Space), but what \s matches in pcre is locale dependent, so use \h instead.

  $ echo "h:mm Ap" | grep -P "^([hHmszaApP\\s:\\.]|'[^']*')*$"
  $ echo "h:mm Ap" | grep -P "^([hHmszaApP\\h:\\.]|'[^']*')*$"
  h:mm Ap

Fixes #4572

- [ x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/desktop-widgets/preferences/preferences_language.cpp
+++ b/desktop-widgets/preferences/preferences_language.cpp
@@ -98,7 +98,7 @@ void PreferencesLanguage::syncSettings()
 	refreshSettings();
 
 	QString qDateTimeWeb = tr("These will be used as is. This might not be what you intended. To avoid this warning wrap the literal parts in quotes (').\nSee https://doc.qt.io/archives/qt-4.8/qdatetime.html#fromString");
-	QRegularExpression timeStandardFormat("^([hHmszaApP\\s:\\.]|'[^']*')*$");
+	QRegularExpression timeStandardFormat("^([hHmszaApP\\h:\\.]|'[^']*')*$");
 	if (!timeStandardFormat.match(ui->timeFormatEntry->currentText()).hasMatch())
 		QMessageBox::warning(this, tr("Literal characters"),
 			tr("Non-standard character(s) in time format.\n") + qDateTimeWeb);


mikeller (approved):
Thanks for your contribution!

RonOnGithub:
@mikeller  Just a followup note to this issue.

I'm currently working through cleaning up the locale handling and it turns out that quite a few of the locale-specific default formats for time and dates (now?) contain all manner of characters that trigger these warnings (and often confusingly both of them, one for the time format and then another immediately after dismissing the first for the date format).

So I think the only sane option is that we simply get rid of it, on the assumption that the QLocale supplied defaults are all already correctly quoted if they need to be - and that if the user then explicitly overrides it, and explicitly enters a garbage format string, then they'll get exactly what they asked for.

I don't think there's anywhere we use this where making that mistake will cause permanent lasting damage, we should only use these for displaying date/time to the user - so if they later go back and fix their typo or mistake then All Will Be Fine again.  In the meantime any problem should be obvious all over the dive list, so it's not like it would go unnoticed for long.

If we still want some sort of sanity check, then maybe we should just display the current time and date using the current format somewhere in that preference window?

Unless someone points out something terrible I've missed, I'll remove this as part of the locale fixes in my stack now.

PR #4580, opened 2025-08-25
Don't limit make to -j4

mankoff:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [X] Build system change
- [ ] Documentation change
- [ ] Language translation

Current build system hard-codes "make -j4" which may be problematic if you have fewer than 4 cores, and is inefficient if you have a lot more. I suggest users should `export MAKEFLAGS="-j$(nproc)"` or similar to have make build on as many cores as they want.

### Changes made:
1) Removed all " -j4" strings from Makefile.

### Documentation change:
It may be worth adding `export MAKEFLAGS="-j$(nproc)"` to the documentation and I can do that if others feel it should be done.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -57,6 +57,12 @@ jobs:
         qtpositioning5-dev qtscript5-dev qttools5-dev qttools5-dev-tools \
         qtquickcontrols2-5-dev xvfb libbluetooth-dev libmtp-dev libraw-dev
 
+    - name: set environment variables
+      id: set-variables
+      run: |
+        NUM_CORES=$(nproc)
+        echo "number-cores=$NUM_CORES" >> $GITHUB_OUTPUT
+
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v3
@@ -68,6 +74,8 @@ jobs:
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
 
     - name: Build
+      env:
+        MAKEFLAGS: "j${{ steps.set-variables.outputs.number-cores }}"
       run: |
         cd ..
         git config --global --add safe.directory $GITHUB_WORKSPACE
--- a/scripts/build-libdivecomputer.sh
+++ b/scripts/build-libdivecomputer.sh
@@ -20,6 +20,16 @@ fi
 
 ../configure --disable-examples --prefix=$INSTALL_ROOT
 
+# Use all cores, unless user set their own MAKEFLAGS
+if [[ -z "${MAKEFLAGS+x}" ]]; then
+	if [[ ${PLATFORM} == "Linux" ]]; then
+		MAKEFLAGS="-j$(nproc)"
+	elif [[ ${PLATFORM} == "Darwin" ]]; then
+		MAKEFLAGS="-j$(sysctl -n hw.logicalcpu)"
+	else
+		MAKEFLAGS="-j4"
+	fi
+fi
 
 make -j4
 make install
--- a/packaging/android/qmake-build.sh
+++ b/packaging/android/qmake-build.sh
@@ -34,6 +34,7 @@ QUICK=""
 ARCHITECTURES=""
 BUILD_ABIS=""
 versionOnly=""
+PLATFORM=$(uname)
 
 while [ "$#" -gt 0 ] ; do
 	case "$1" in
@@ -120,6 +121,17 @@ else
 	exit 1
 fi
 
+# Use all cores, unless user set their own MAKEFLAGS
+if [[ -z "${MAKEFLAGS+x}" ]]; then
+	if [[ ${PLATFORM} == "Linux" ]]; then
+		MAKEFLAGS="-j$(nproc)"
+	elif [[ ${PLATFORM} == "Darwin" ]]; then
+		MAKEFLAGS="-j$(sysctl -n hw.logicalcpu)"
+	else
+		MAKEFLAGS="-j4"
+	fi
+fi
+
 QMAKE=$QT5_ANDROID/android/bin/qmake
 echo $QMAKE
 $QMAKE -query
@@ -182,7 +194,7 @@ if [ "$QUICK" = "" ] ; then
 	    mkdir -p googlemaps-build
 	    pushd googlemaps-build
 	    $QMAKE ANDROID_ABIS="$BUILD_ABIS" ../googlemaps/googlemaps.pro
-	    make -j4
+	    make
             make install
 	    popd
 	fi
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -175,6 +175,17 @@ while [[ $# -gt 0 ]] ; do
 	shift
 done
 
+# Use all cores, unless user set their own MAKEFLAGS
+if [[ -z "${MAKEFLAGS+x}" ]]; then
+	if [[ ${PLATFORM} == "Linux" ]]; then
+		MAKEFLAGS="-j$(nproc)"
+	elif [[ ${PLATFORM} == "Darwin" ]]; then
+		MAKEFLAGS="-j$(sysctl -n hw.logicalcpu)"
+	else
+		MAKEFLAGS="-j4"
+	fi
+fi
+
 # recreate the old default behavior - no flag set implies build desktop
 if [ "$BUILD_MOBILE$BUILD_DOWNLOADER" = "" ] ; then
 	BUILD_DESKTOP="1"
@@ -382,7 +393,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	mkdir -p build
 	cd build
 	cmake $MAC_CMAKE ..
-	make -j4
+	make
 	make install
 	popd
 
@@ -399,7 +410,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 		../Configure --prefix="$INSTALL_ROOT" --openssldir="$INSTALL_ROOT" "$MAC_OPTS_OPENSSL" $OS_ARCH
 		make depend
 		# all the tests fail because the assume that openssl is already installed. Odd? Still things work
-		make -j4 -k
+		make -k
 		make -k install
 		cd ..
 	done
@@ -422,7 +433,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	cd build
 	CFLAGS="$MAC_OPTS" ../configure --prefix="$INSTALL_ROOT" --with-openssl \
 		--disable-tftp --disable-ftp --disable-ldap --disable-ldaps --disable-imap --disable-pop3 --disable-smtp --disable-gopher --disable-smb --disable-rtsp
-	make -j4
+	make
 	make install
 	popd
 
@@ -431,7 +442,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	mkdir -p build
 	cd build
 	cmake $MAC_CMAKE -DBUILD_SHARED_LIBS=ON -DBUILD_TESTING=OFF -DBUILD_EXAMPLES=OFF ..
-	make -j4
+	make
 	make install
 	popd
 	if [ "$PLATFORM" = Darwin ] ; then
@@ -454,7 +465,7 @@ if [[ "$LIBGITMAJ" -lt "1" && "$LIBGIT" -lt "26" ]] ; then
 	mkdir -p build
 	cd build
 	cmake $MAC_CMAKE -DBUILD_CLAR=OFF ..
-	make -j4
+	make
 	make install
 	popd
 
@@ -478,7 +489,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	mkdir -p build
 	cd build
 	cmake $MAC_CMAKE ..
-	make -j4
+	make
 	make install
 	popd
 
@@ -489,7 +500,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	mkdir -p build
 	cd build
 	CFLAGS="$MAC_OPTS" ../configure --prefix="$INSTALL_ROOT"
-	make -j4
+	make
 	make install
 	popd
 
@@ -499,7 +510,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	mkdir -p build
 	cd build
 	CFLAGS="$MAC_OPTS" ../configure --prefix="$INSTALL_ROOT" --disable-examples
-	make -j4
+	make
 	make install
 	popd
 
@@ -508,7 +519,7 @@ if [[ $PLATFORM = Darwin && "$BUILD_DEPS" == "1" ]] ; then
 	mkdir -p build
 	cd build
 	cmake $MAC_CMAKE ..
-	make -j4
+	make
 	make install
 	popd
 fi
@@ -559,7 +570,7 @@ if [ "$PLATFORM" = Darwin ] ; then
 	grep CLOCK_GETTIME config.h
 	sed -i .bak 's/^#define HAVE_CLOCK_GETTIME 1/#undef HAVE_CLOCK_GETTIME /' config.h
 fi
-make -j4
+make
 make install
 # make sure we know where the libdivecomputer.a was installed - sometimes it ends up in lib64, sometimes in lib
 STATIC_LIBDC="$INSTALL_ROOT/$(grep ^libdir Makefile | cut -d/ -f2)/libdivecomputer.a"
@@ -583,7 +594,7 @@ if [ "$QUICK" != "1" ] && [ "$BUILD_DESKTOP$BUILD_MOBILE" != "" ] && ( [[ $QT_VE
 	else
 		$QMAKE "INCLUDEPATH=$INSTALL_ROOT/include" "CONFIG+=release" ../googlemaps.pro
 	fi
-	make -j4
+	make
 	if [ "$PLATFORM" = Darwin ]  && [[ $QT_VERSION == 6* ]] && [[ $ARCHS == *" "* ]] ; then
 		# we can't build fat binaries directly here, so let's do it in two steps
 		# above we build the 'native' binary, now build the other one
@@ -592,7 +603,7 @@ if [ "$QUICK" != "1" ] && [ "$BUILD_DESKTOP$BUILD_MOBILE" != "" ] && ( [[ $QT_VE
 		mkdir -p ../build-$OTHERARCH
 		cd ../build-$OTHERARCH
 		$QMAKE "INCLUDEPATH=$INSTALL_ROOT/../qtlocation/build/include/QtLocation/6.3.0" QMAKE_APPLE_DEVICE_ARCHS=$OTHERARCH ../googlemaps.pro
-		make -j4
+		make
 		# now combine them into one .dylib
 		mkdir -p "$INSTALL_ROOT"/plugins/geoservices
 		lipo -create ./libqtgeoservices_googlemaps.dylib ../build/libqtgeoservices_googlemaps.dylib -output "$INSTALL_ROOT"/plugins/geoservices/libqtgeoservices_googlemaps.dylib
@@ -666,7 +677,7 @@ for (( i=0 ; i < ${#BUILDS[@]} ; i++ )) ; do
 	fi
 
 	if [ ! "$PREP_ONLY" = "1" ] ; then
-		LIBRARY_PATH=$INSTALL_ROOT/lib make -j4
+		LIBRARY_PATH=$INSTALL_ROOT/lib make
 		LIBRARY_PATH=$INSTALL_ROOT/lib make install
 
 		if [ "$CREATE_APPDIR" = "1" ] ; then


mikeller (changes requested):
I get the point of your idea, and I think it makes sense.
However, one area where the current settings for paralellism are used are our CICD builds. Removing the `-j4`  results in at least a doubling of the run time for most of our automated builds, which is not desirable at all.
In order for this to be accepted, it will also need to substitute the current fixed settings for parallelism for all of our CICD builds with something that enables (either fixed or dynamic) parellelism.

mankoff:
Thanks for the feedback. Are you suggesting I also edit all of the calls to `build.sh` in `.github/workflows`? Add a `MAKEFLAGS=-j4` to each call? Is there some top-level GitHub workflow environment setup where I would enter this 1x?

glance-:
I'd suggest just adding a line in the workflows that sets `MAKEFLAGS=-j$(nproc)` or equivalent if that doesn't work on the different github actions images.
That way we will use the amount of cores that each runner exposes.

After we see the times those runs produces we might decide to tune that number up by 1.5x or something to saturate the cpu cores while others are waiting on IO.

mikeller:
There's also https://github.com/marketplace/actions/cpu-core-count that works in a platform-independent way and eliminates the need to install nproc on each platform.

glance-:
nproc is a part of coreutils , from which we already use a bunch of tools, and it's already installed in all our environments that I could find, including containers, except the mac builds, for which the equivalent is `sysctl -n hw.logicalcpu`.

So nothing extra needs to be installed.

mankoff:
I missed some additional four `make -j4` locations:

+ https://github.com/subsurface/subsurface/blob/445cd3ac0441f177a096ce3d19fde20d8af6ffc8/packaging/android/qmake-build.sh#L185
+ https://github.com/subsurface/subsurface/blob/445cd3ac0441f177a096ce3d19fde20d8af6ffc8/scripts/build-libdivecomputer.sh#L24
+ https://github.com/subsurface/subsurface/blob/445cd3ac0441f177a096ce3d19fde20d8af6ffc8/packaging/OBS/subsurfacedaily.spec#L112
+ https://github.com/subsurface/subsurface/blob/445cd3ac0441f177a096ce3d19fde20d8af6ffc8/packaging/copr/subsurface.spec#L84

Question: should I continue on the current plan: Remove all `-j4`, set it in an environment variable for the GitHub build system, and adding documentation for users (who compile the code). Or alternatively, replacing all `-j4`, including those already remove in the first commit, with `-j$(nproc)`?

Some pros/cons:
+ Benefits of coding `-j$(nproc)`: Builds as fast as possible for everyone by default.
+ Problems of hard-coding: Users may not want to use all cores.
+ Benefits of leaving in environment var: Give users control
+ Problem of env var: Builds slow by default.

glance-:
I'd vote for not having any `make -j<whatever>` in the code, and let the calling user set those.

I have a `export MAKEFLAGS=-j$(nproc)` in my rc-file so I always get one that matches my current machine. If I for some case (like buggy makefiles, big build while doing other things etc) would like to tune it some other way, that's my exception.

We could have a helping hand that checks if you have a -j flag in your MAKEFLAGS and warns you that you should set one.
Something like:
```bash
if getopts "j:" _ "$MAKEFLAGS" ; then
    warn 'You should run "export MAKEFLAGS=-j$(nproc)" or your build will be slow'
fi
```
The other alternative would be to set MAKEFLAGS=-j$(nproc) if there's no -j in MAKEFLAGS. I'd go with just a warning.

mikeller:
Agreed on not hardcoding the `-j` option in scripts - this isn't ideal.

I think relying on `MAKEFLAGS` is a good approach, and we should set this in the appropriate way for all of our supported platforms in the CICD workflows. This will be a bit of manual work as there is no good / centralised way to do this, but I should be able to help with this. For testing we should just be able to compare workflow run times before / after the change.

I'm not a fan of specifically referring to `nproc` in scripts or documentation, as this is specific to platform and installation, and will likely result in bug reports from users for whom it does not work. 'include a `-j<number of cores>` statement in your `MAKEFLAGS` to optimise the build for your hardware' puts the ball into the users' court for this.

mikeller:
> I missed some additional four `make -j4` locations:
> 
> * https://github.com/subsurface/subsurface/blob/445cd3ac0441f177a096ce3d19fde20d8af6ffc8/packaging/android/qmake-build.sh#L185
> * https://github.com/subsurface/subsurface/blob/445cd3ac0441f177a096ce3d19fde20d8af6ffc8/scripts/build-libdivecomputer.sh#L24

I think these should be removed.

> * https://github.com/subsurface/subsurface/blob/445cd3ac0441f177a096ce3d19fde20d8af6ffc8/packaging/OBS/subsurfacedaily.spec#L112
> * https://github.com/subsurface/subsurface/blob/445cd3ac0441f177a096ce3d19fde20d8af6ffc8/packaging/copr/subsurface.spec#L84

These are actually build definitions that are run on the distributions' build servers, so the environment is mostly outside of our control. So I'd just leave them as they are.

mankoff:
In addition to leaving it in the `spec` files as suggested above, it remains in one other place, the Dockerfile: 

https://github.com/mankoff/subsurface/blob/2c2616b96d700fe67914d6aab85c147bbc2571fb/scripts/docker/mxe-build-container/Dockerfile#L82-L84

I've tried to set the variable correctly per Workflow, and added a variation of the user warning to the build.sh (wasn't clear where in that script to place it).

.github/workflows/codeql-analysis.yml
```diff
@@ -72,6 +72,7 @@ jobs:
         cd ..
         git config --global --add safe.directory $GITHUB_WORKSPACE
         git config --global --add safe.directory $GITHUB_WORKSPACE/libdivecomputer
+	export MAKEFLAGS=-j$(nproc)
```
mikeller: ~I am surprised that this is even working,~ as YAML requires the use of spaces instead of tabs for indentation : https://yaml.org/spec/1.2.2/#61-indentation-spaces

Turns out it's not working, GitHub is just not alerting about not running malformed YAML files:

image

mankoff: Yup, not working. My tab-mode solution to the earlier tabs v. spaces issue was too aggressive. Is there a `make test` or some command I could have run to catch this since I'm not very familiar with YAML format? Or is diagnosing via GitHub Workflow a reasonable approach. Seems computationally expensive.

mikeller: > Is there a `make test` or some command I could have run to catch this since I'm not very familiar with YAML format? Or is diagnosing via GitHub Workflow a reasonable approach. Seems computationally expensive.

Na, one of the peculiarities of working on CICD is that testing involves a lot of CICD compute. You _could_ probably run a YAML validator over the files, but correct YAML is still a long way away from working workflows / actions.

mankoff:
Well... the tests passed, but it looks like things are slower. I'm not sure if the Workflow YML files are seeing the changes? Another approach could be to have build.sh detect if it's running as a GitHub workflow, and if so, set MAKEFLAGS internally then?

I find the timing of the workflows under the "Checks" tab at the top of this (and other) PRs:

Here's a link to the Fedora build for this PR at 8m33s: https://github.com/subsurface/subsurface/actions/runs/17256938507

Here's a link to the Fedora build for the recent LIBZIP PR at 7m44s: https://github.com/subsurface/subsurface/actions/runs/17236399963

.github/workflows/coverity-scan.yml
```diff
@@ -37,6 +37,7 @@ jobs:
       run: |
         git config --global --add safe.directory $GITHUB_WORKSPACE
         git config --global --add safe.directory $GITHUB_WORKSPACE/libdivecomputer
+        export MAKEFLAGS=-j$(nproc)
```
mikeller: I suspect that this will not work, as the individual steps will be executed in different shells. We _might_ be able to make this work by leveraging `env:` in the actual build step, but we need to first sort out why this doesn't seem to work at all.

mikeller (commented):
Giving it a go with a different format.

.github/workflows/coverity-scan.yml
mankoff: Per ChatGPT, `env:` cannot execute commands (e.g., `$(nproc)`). But we can go back to the original `-j4` in the `env:` section. If we do this, workflows should run exactly as before.

atdotde:
Why can't we set MAKEFLAGS to a good value in build.sh if it is unset? Then users have control but you don't have to start setting environment variables when first building subsurface. Even the old -j4 is better than nothing or something testing the number of cores available.

mankoff:
I think this is a fine approach, and suggested it above (https://github.com/subsurface/subsurface/pull/4580#issuecomment-3225538412) but perhaps not as clearly as you did here. 

It would just be a few lines to check if `nproc` exists and set to that, else check if `sysctl -n hw.logicalcpu` returns something, else set to either `-j4` or set to nothing.

And all this can be done only if MAKEFLAGS is unset, respecting power users preference for not using all cores.

.github/workflows/coverity-scan.yml
mikeller: > Per ChatGPT, `env:` cannot execute commands (e.g., `$(nproc)`). But we can go back to the original `-j4` in the `env:` section. If we do this, workflows should run exactly as before.

There are ways to get the dynamic behaviour that we are looking for, see [`d2c5e70` (#4580)](https://github.com/subsurface/subsurface/pull/4580/commits/d2c5e70611822aa342d111aed626c70b8acc0489) for one way of doing this.

mikeller:
> Why can't we set MAKEFLAGS to a good value in build.sh if it is unset? Then users have control but you don't have to start setting environment variables when first building subsurface. Even the old -j4 is better than nothing or something testing the number of cores available.

I like this idea.

Apart from this, the CICD run times in the current version seem to be reasonable - they are close to what we had before the change, which likely means that the `-j4` setting was a reasonable assumption for the infrastructure that the workflows are running on.

mankoff:
Current status. I reverted most of the previous exploratory commits. Now it should only be 1) removal of all `-jX` flags on calls to make, and 2) setting MAKEFLAGS in build.sh with

```bash
if ! [[ ${MAKEFLAGS:-} == *"-j"* ]]; then # not found in var which may or may not exist
  [[ ${PLATFORM} == "Darwin" ]] && MAKEFLAGS="${MAKEFLAGS:-} -j$(sysctl -n hw.logicalcpu)"
  [[ ${PLATFORM} == "Linux" ]] && MAKEFLAGS="${MAKEFLAGS:-} -j$(nproc)"
  [[ ! (${MAKEFLAGS:-} == *"-j"*) ]] && MAKEFLAGS="${MAKEFLAGS:-} -j4" # default to -j4
fi
```

I do worry that either I've been a bit too aggressive, or need to set MAKEFLAGS elsewhere. The following build(ish).sh scripts also had -jX removed, and may need the above code, or something simpler, set:

- packaging/android/qmake-build.sh
- packaging/ios/build.sh
- packaging/macosx/make-package.sh
- packaging/ubuntu/debian/rules
- scripts/build-libdivecomputer.sh

I'm not very familiar with the build system yet. Are some of these called from build.sh and therefore inherit MAKEFLAGS (if so, do we need to `export` it)?. Are some of these a known platform/OS (and possibly known that MAKEFLAGS isn't set because we're in a Docker container) and so MAKEFLAGS can be set with simpler code?

mankoff:
> ```shell
> if ! [[ ${MAKEFLAGS:-} == *"-j"* ]]; then # not found in var which may or may not exist
>   [[ ${PLATFORM} == "Darwin" ]] && MAKEFLAGS="${MAKEFLAGS:-} -j$(sysctl -n hw.logicalcpu)"
>   [[ ${PLATFORM} == "Linux" ]] && MAKEFLAGS="${MAKEFLAGS:-} -j$(nproc)"
>   [[ ! (${MAKEFLAGS:-} == *"-j"*) ]] && MAKEFLAGS="${MAKEFLAGS:-} -j4" # default to -j4
> fi
> ```


Whitespace error noticed and fixed. Waiting to push to combine with other changes to save resources.

glance-:
I prefer using optparse rather than globbing for -j in MAKEFLAGS, like:
`if getopts "j:" _ "$MAKEFLAGS" ; then`

glance-:
Later when you start to feel that the end result shapes up, rebase your branch and drop the commits and corresponding reverts, so you get one clean sane history.

mankoff:
> Later when you start to feel that the end result shapes up, rebase
> your branch and drop the commits and corresponding reverts, so you get
> one clean sane history.

My approach would have been a squash merge. But I can clean it up from my end too. I think what you suggest implies a force push? I know mikeller has contributed to this branch, and I think force pushes to public repositories, especially when others have pulled your branch, are frowned upon, but maybe it's ok in this case?

  -k.

glance-:
In a feature branch anything goes.

If you just squash everything or do a rebase / fixup / squash is up to you.

glance-:
-j [jobs], --jobs[=jobs]
            Specifies  the number of jobs (commands) to run simultaneously.  If there is more than one -j option, the last one is effective.  If the -j option is given without an argument, make will not limit the number of jobs that can run simultaneously. When make invokes a sub-make, all instances of make will coordinate to run the specified number of jobs at a time; see the section PARALLEL MAKE AND THE JOBSERVER for details.

make -j can easily over saturate the cpu and io, and the scheduler just swaps between all the things that would like to run simultaneously, and it takes more time, so I don't think that's a good approach.

mankoff:
Current status:

I've saved a branch with all the experimental work, but have pushed a basic squashed commit with only these changes:

1. Remove `-jN` from all shell scripts and replace with `MAKEFLAGS="-j ${MAKEFLAGS:-}"`. There is no OS-specific logic or calls to determine the number of cores, because Make will use all available cores if flag `-j` is set. There is no check if `-j` is already set, because Make uses the last `-j`, so if a user already has `MAKEFLAGS=-jN`, setting a `-j` in front of that should not override their preferences.

2. In one place where I was unsure about setting this variable (packaging/ubuntu/debian/rules which might be a Makefile?) I changed `-j8` to just `-j`.

mankoff:
> make -j can easily over saturate the cpu and io

Oh geeze. Our comments crossed in the ether. Well this would also suggest that `make -j$(nproc)` is also not a good solution, right?

Perhaps I should abandon this PR? Too much work for not enough payback? Other suggestions?

glance-:
make -j$(nproc) is fine, because it's limited to the number of cores, meanwhile make -j is unlimited in the amount of jobs make runs.

That's why you need to inform make about a sane amount of job-slots to have.

glance-:
Also, make's job-server can be passed on to subsequent calls to make, so one make calling a script that calls another make can still talk to the "outer" make to limit the amount of concurrent tasks that will be running.
Just prefix the outer makes recepie lines with + to pass the jobserver arguments and fd's on.

glance-:
So, don't hard code make -j<anything> (or worse make -j without a job limit) anywhere, just call make, and have the outer most logic discover the amount of cores with nproc, or for mac sysctl -n hw.logicalcpu, and set MAKEFLAGS=-j<number of cores> and that's it.

make flags on the cmdline will have higher priority that what it finds in the env-var.

mankoff:
> So, don't hard code make -j (or worse make -j without a job limit) anywhere, just call make, and have the outer most logic discover the amount of cores with nproc, or for mac sysctl -n hw.logicalcpu, and set MAKEFLAGS=-j and that's it.

Agreed. I thought this is what `-j` was doing (I thought `-j` meant use all cores, not use infinite). I think the current version may be correct or at least getting much closer to what you describe here. 

I'm not sure what 'outer most logic' means. I don't understand (yet) the entire build system. When GitHub fires off Workflows, is there something outer/inner, or is each system building independently?

Also, from your earlier comment, I do not know Make well enough to understand 

> Just prefix the outer makes recepie lines with + to pass the jobserver arguments and fd's on.

scripts/build-libdivecomputer.sh
```diff
@@ -20,6 +20,14 @@ fi
 
 ../configure --disable-examples --prefix=$INSTALL_ROOT
 
+# Use all cores, unless user set -j<n>
+if [[ ${PLATFORM} == "Linux" ]]; then
+	MAKEFLAGS=" -j$(nproc) ${MAKEFLAGS:-}"
```
mikeller: This will break the use of user defined `MAKEFLAGS` with single letter options according to https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html

Also, if we are using `nproc` in a generic build script we should really test if it is available first.

mikeller (commented):
I think we are overcooking this now, and adding more complexity than is needed.
I think we can simplify this to:
- in the build scripts, check if `MAKEFLAGS` is set, and if so assume that the user is knowledgeable enough to manage the parallel execution on their infrastructure;
- if it is not set, revert to the old behaviour of hardcoding `-j 4` in `MAKEFLAGS` - this has worked without user complaints until now;
- explicitly set the correct parallelism per core in build scripts, as we know and control the environment in these cases.

mankoff:
I'm going to stop pushing code and just paste it here for discussion, if that's OK with y'all.

> - (1) in the build scripts, check if `MAKEFLAGS` is set, and if so assume that the user is knowledgeable enough to manage the parallel execution on their infrastructure;

> - (2) if it is not set, revert to the old behaviour of hardcoding `-j 4` in `MAKEFLAGS` - this has worked without user complaints until now;

> - (3) explicitly set the correct parallelism per core in build scripts, as we know and control the environment in these cases.

My interpretation of the above is:

```bash
[ -z "${MAKEFLAGS+x}" ] && MAKEFLAGS="-j4"
```

This handles (1) and (2): If var undefined, set to -j4. I'm unclear what is meant by (3) or where to explicitly set this and which build scripts you are referring to that are different than the scripts in (1). The latest changes only impact

packaging/android/qmake-build.sh
scripts/build-libdivecomputer.sh
scripts/build.sh

mikeller:
> My interpretation of the above is: ```bash [ -z "${MAKEFLAGS+x}" ] && MAKEFLAGS="-j4" ``` This handles (1) and (2): If var undefined, set to -j4.

Yes, pretty much.

> I'm unclear what is meant by (3) or where to explicitly set this and which build scripts you are referring to that are different than the scripts in (1). The latest changes only impact packaging/android/qmake-build.sh scripts/build-libdivecomputer.sh scripts/build.sh

It's essentially what was added to the workflow definitions in `.github` in 2c2616b96d700fe67914d6aab85c147bbc2571fb and d2c5e70611822aa342d111aed626c70b8acc0489.

mankoff:
> It's essentially what was added to the workflow definitions in .github in https://github.com/subsurface/subsurface/commit/2c2616b96d700fe67914d6aab85c147bbc2571fb and https://github.com/subsurface/subsurface/commit/d2c5e70611822aa342d111aed626c70b8acc0489.

I can make that change, but can you give specific guidance on the implementation. I'm not sure why you added the 2nd commit (https://github.com/subsurface/subsurface/commit/d2c5e70611822aa342d111aed626c70b8acc0489). Did my initial version of `export MAKEFLAGS="-j$(nproc)"`not work? Should I change everything to match the https://github.com/subsurface/subsurface/commit/d2c5e70611822aa342d111aed626c70b8acc0489 version? And finally, per your earlier comments, should this be hard-coded `-j4` or `-j$(nproc)`? I assume you want `-j$(nproc)` because the build.sh script sets `-j4` in the current version we're discussing.

Alternative suggestion that may be easier to maintain: `if [ "$GITHUB_ACTIONS" = "true" ]; then...` for code specific to GitHub Workflow runs in the three places we currently set MAKEFLAGS. The YAML solution touches many more files and many more lines of code.

scripts/build-libdivecomputer.sh
glance-: We don't test for the existence of ls, mv or other coreutils parts so why do something specific about nproc ?

mikeller:
> > It's essentially what was added to the workflow definitions in .github in [2c2616b](https://github.com/subsurface/subsurface/commit/2c2616b96d700fe67914d6aab85c147bbc2571fb) and [d2c5e70](https://github.com/subsurface/subsurface/commit/d2c5e70611822aa342d111aed626c70b8acc0489).
> 
> I can make that change, but can you give specific guidance on the implementation. I'm not sure why you added the 2nd commit ([d2c5e70](https://github.com/subsurface/subsurface/commit/d2c5e70611822aa342d111aed626c70b8acc0489)). 

I added ([d2c5e70](https://github.com/subsurface/subsurface/commit/d2c5e70611822aa342d111aed626c70b8acc0489)) to enable parallelism for the codeQL CICD build.

> Did my initial version of `export MAKEFLAGS="-j$(nproc)"`not work? 

It did work - but you had noted in https://github.com/subsurface/subsurface/pull/4580#issuecomment-3226699810 that it did not seem to give us a speed improvement, so I wanted to test if setting `MAKEFLAGS` in the way that is [documented](https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html) would change this.
But as it turns out your earlier change was already effective - just that the hardcoded `-j 4` seems to already be optimal when running on CICD.

> Should I change everything to match the [d2c5e70](https://github.com/subsurface/subsurface/commit/d2c5e70611822aa342d111aed626c70b8acc0489) version?

It doesn't seem to matter much with the current version of make - it seems to accept both versions.

> And finally, per your earlier comments, should this be hard-coded `-j4` or `-j$(nproc)`? I assume you want `-j$(nproc)` because the build.sh script sets `-j4` in the current version we're discussing.

Yes, I think `j$(nproc)` is preferable, in case GitHub changes the hardware specs of their build runners.

> Alternative suggestion that may be easier to maintain: `if [ "$GITHUB_ACTIONS" = "true" ]; then...` for code specific to GitHub Workflow runs in the three places we currently set MAKEFLAGS. The YAML solution touches many more files and many more lines of code.

I don't like the approach of adding in more different code paths - this will result in more maintenance and testing work.

mankoff:
Sorry I was distracted for a few weeks. I think this might be what you're looking for? If yes, there are now 7 misc messy commits, I suggest a squash-merge here, or tell me and I can squash and force-push.

mikeller (commented):
Yes, this looks good - except that we should probably remove the hardcoded `-j4` stanzas as well. ;-)

mankoff:
> Yes, this looks good - except that we should probably remove the hardcoded `-j4` stanzas as well. ;-)

Done for `qmake-build.sh`. But there remains:

+ `make -j` in `scripts/docker/mxe-build-container/Dockerfile`
+ `make -j` in `packaging/macosx/make-package.sh`
+ `make -j4` in `packaging/OBS/subsurfacedaily.spec`
+ `make -j8` in `packaging/ubuntu/debian/rules` (3x)
+ `make -j4` in `packaging/copr/subsurface.spc`

I'm not sure how to fix these, or if it is necessary. I don't even know what language some of these are in.

mikeller:
> > Yes, this looks good - except that we should probably remove the hardcoded `-j4` stanzas as well. ;-)
> 
> Done for `qmake-build.sh`. But there remains:
> 
> * `make -j` in `scripts/docker/mxe-build-container/Dockerfile`
> * `make -j` in `packaging/macosx/make-package.sh`
> * `make -j4` in `packaging/OBS/subsurfacedaily.spec`
> * `make -j8` in `packaging/ubuntu/debian/rules` (3x)
> * `make -j4` in `packaging/copr/subsurface.spc`
> 
> I'm not sure how to fix these, or if it is necessary. I don't even know what language some of these are in.

I think we should remove `-j4` from `scripts/build.sh` as well - this is the main build script.

We can deal with the rest of them in a later pull request, most of them will need some testing.

mankoff:
> I think we should remove `-j4` from `scripts/build.sh` as well - this is the main build script.

Agreed. That was the whole point of this PR and I'm not sure how I missed them on this remake of the branch. Removed and pushed.

mikeller (approved):

mikeller:
Unfortunately it looks like this has turned out poorly for our CICD setup - quite a few of our automated builds have started taking around twice as long after this has been merged:

image

The affected builds are:

```
coverity-scan.yml
mac.yml
linux-debian-bookworm-5.15.yml
linux-ubuntu-20.04-qt5.yml
linux-ubuntu-20.04-qt5-appimage.yml
linux-ubuntu-22.04-qt5.yml
linux-ubuntu-24.04-qt5.yml
linux-ubuntu-24.04-qt5-export-html.yml
```

I have started investigating why this isn't working as intended.

mankoff:
Sorry this patch wasn't good. If your fix isn't working, I suggest again that we can detect when we're in the CI/CD space via a GITHUB env var. But I agree with your earlier comment that this is not clean, the code shouldn't know or care where it is being executed.

mikeller:
All good, I fixed it in #4606. In the end it was adding the `export` that fixed it,  but since I had already done the other improvements I left them in.

Issue #4587, opened 2025-09-06
[Feature Request]: Social Media Sharing for Dive Stories


Nesslax:
### Problem Statement

Since the increase of sharing activities on social media, especially within the running and fitness communities, I’d like to suggest adding a new feature to the mobile app: social media sharing for dives.

### Description

When viewing a dive in the mobile app, add a “Share” button that generates a shareable image with transparent background including the main dive data.

Example Output (image):
🌊 Dive Story — [Location Name] 🌊

📍 [Dive site name, city/country]  
🗓️ [Date]  
🤿 Dive #XX  

Stats:  
  • ⏱️ Bottom Time: 45 min  
  • 📏 Max Depth: 32m  
  • 🌡️ Water Temp: 27°C  
  • 🏝️ Visibility: 20m  
  • 🐠 Gas Mix: Air / Nitrox 32  

•	Include a time-series plot of the dive profile.
•	Add the Subsurface logo to promote the app’s identity.

Benefits:
	•	Encourages divers to share their dives on social media, similar to how runners share activity summaries.
	•	Promotes Subsurface brand visibility.
	•	Provides a cleaner and more professional alternative to screenshots of dive computers.

Possible Enhancements:
	•	Option to customize background color (transparent / gradient / photo overlay).
	•	Ability to export as PNG/JPEG for sharing on Instagram, Facebook, Twitter, etc.

cetra3:
You can use [DiveDB](https://github.com/cetra3/divedb) to do this by:

* Syncing your dives from SubSurface cloud
* Marking a dive as public.

Here's an example: https://divedb.net/dives/00000273-292f-5789-9186-c44a5fa279e1

It also integrates into the fediverse

Nesslax:
I was talking about something more like this : 

<img width="1080" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0c476aae-5db2-4467-972b-7d17fe3ae574" />

but with a better version

dirkhh:
ironically we used to have something similar years ago. It was essentially unused (as the stats showed), yet created a constant workload on the maintainers as the APIs from social media keep shifting and increasingly required legal agreements to be signed, etc.

I'd consider this out of scope for Subsurface.

cetra3:
[...]

PR #4593, opened 2025-09-15
Show dive information in profile export

atdotde:
This was requested for social media posting.

This adds an option to overlay the profile picture with the most important dive information.

It addresses #4587 in a minimalistic way.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -18,7 +18,9 @@
 #include "core/selection.h"
 #include "core/taxonomy.h"
 #include "core/sample.h"
+#include "core/string-format.h"
 #include "profile-widget/profilescene.h"
+#include "core/qthelper.h"
 #include <QDir>
 #include <QFileInfo>
 #include <QtConcurrent>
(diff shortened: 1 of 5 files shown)


mikeller (commented):
Another thing that I noticed is that the profile seems to be zooming the tissue heat-map if it is enabled in the profile, which makes the dive data hard to read in this case:
dive

Nitpick: Your changes are using spaces instead of tabs.

atdotde:
Yes, the scaling is totally screwed up. I will investigate this. In addition, some of the symbols don't show up properly on my Mac. And I will fix the spacing.

mikeller:
I suspect that the scaling problems in the profile export might have been introduced in #4428 - at the time I wasn't aware that the profile code is also used for this export, and did not test this.

atdotde:
Indeed, yesterday I did a bisect which located it to f27850e421d7ba22f99512ae50f5f2caa4849cb0.  Hope to find some more time to look into this later today.

atdotde:
Found the problem. Will send a patch once I fixed the other problems.

mikeller:
I've given the latest version another spin, but now Subsurface crashes on clicking 'Save' for the dive profile export for some but not all dives - my suspicion is that it happens for CCR dives with a ppO2 graph.
I can get a stack trace for it tonight.

backend-shared/exportfuncs.cpp
```diff
@@ -41,12 +43,42 @@ static constexpr int profileScale = 4;
 static constexpr int profileWidth = 800 * profileScale;
 static constexpr int profileHeight = 600 * profileScale;
 
-static void exportProfile(ProfileScene &profile, const struct dive &dive, const QString &filename)
+static QString profileText(const struct dive &dive)
+{
+	QString text;
+
+	if (dive.dive_site->name.length() > 0)
+		text += "🗺 " + QString::fromStdString(dive.dive_site->name) + " \n";
```
mikeller, line 51: Found the problem causing the crash - `dive.dive_site->name` needs to be null checked:

```
0x00005555557b1b05 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::length (this=0x8)
    at /usr/include/c++/13/bits/basic_string.h:1079
1079          { return _M_string_length; }
(gdb) bt
#0  0x00005555557b1b05 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::length (this=0x8)
    at /usr/include/c++/13/bits/basic_string.h:1079
#1  profileText (dive=...) at /home/mikeller/git/subsurface_build/subsurface/backend-shared/exportfuncs.cpp:50
#2  exportProfile (profile=<optimized out>, dive=..., filename=..., diveinfo=diveinfo@entry=true)
    at /home/mikeller/git/subsurface_build/subsurface/backend-shared/exportfuncs.cpp:75
#3  0x00005555557b2bf1 in exportProfile (filename=..., selected_only=selected_only@entry=true, cb=..., diveinfo=diveinfo@entry=true)
    at /usr/include/c++/13/bits/unique_ptr.h:199
#4  0x00005555556ad6bb in DiveLogExportDialog::on_buttonBox_accepted (this=this@entry=0x7fffffffcbf0)
    at /home/mikeller/git/subsurface_build/subsurface/desktop-widgets/divelogexportdialog.cpp:227
#5  0x0000555555655ed8 in DiveLogExportDialog::qt_static_metacall
    (_a=0x7fffffffbe10, _id=0, _c=QMetaObject::InvokeMetaMethod, _o=0x7fffffffcbf0)
    at /home/mikeller/git/subsurface_build/build/desktop-widgets/subsurface_interface_autogen/EWIEGA46WW/moc_divelogexportdialog.cpp:78
#6  DiveLogExportDialog::qt_metacall (this=0x7fffffffcbf0, _c=QMetaObject::InvokeMetaMethod, _id=0, _a=0x7fffffffbe10)
    at /home/mikeller/git/subsurface_build/build/desktop-widgets/subsurface_interface_autogen/EWIEGA46WW/moc_divelogexportdialog.cpp:115
```

mikeller:
The profile export looks like it should now (if a location is set) - the only niggle is that the text overlay is obscured by the tissue heat-map and the gas bar if they are enabled - maybe we want to shift the text up above them?

And some of the icons are still not loading on linux.

dive5

atdotde:
Fixed the nulptr defer and the overlap. But for the icons, I am at loss. Maybe @thiagomacieira can help? How do I get an emoj (for example 📅)i that I for example look up on https://symbl.cc/en/1F4C5-calendar-emoji/ into a QString? I tried copy and paste on Ubuntu but this seems to get lost when running on another Linux (as @mikeller does) or on Mac. I failed to get it form some hex endcoding (like we do it for the arrows in pannernotes) but that only produces the literal string. I am really a unicode noob.

atdotde:
Just checked on my Mac: For unclear reasons, now all icons are there: 
image

atdotde:
OK, I did not receive any further feedback and it Works For Me. So will just merge and fix problems later if they occur.

PR #4597, opened 2025-09-18
Modify the mobile planner to fix three bugs relating to cylinders

fnightfox:
…ction

Fixes bugs related to selecting the gas in the segment, having custom gasses present in the dropdown, and using the start pressure in the calculations.

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Additional information:
The changes are related to the mobile planner. In the initial implementation there were three quickly identified minor bugs:
1. The cylinder list, for the first cylinder, would have been created before reading in all of the stored dives, thus causing the loss of all custom cylinders.
2. The gas dropdown in the segment list was implemented incorrectly, causing issues.
3. The cylinder type and pressure were incorrectly being passed on to the backend.

Please note, I do not like how I handle start pressures in calculatePlan in the c++ code. I am open to suggestions on how to handle this better.

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -1429,18 +1429,17 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 		cylinder_t newCyl;
 		std::pair<volume_t, pressure_t> type_info = get_tank_info_data(tank_info_table, map["type"].toString().toStdString());
 		volume_t size = type_info.first;
-		pressure_t pressure = type_info.second;
 		newCyl.type.size = size;
-		newCyl.type.workingpressure = pressure;
+		if (prefs.units.pressure == units::BAR)
+			newCyl.type.workingpressure.mbar = map["pressure"].toInt() * 1000;
+		else
+			newCyl.type.workingpressure.mbar = psi_to_mbar(map["pressure"].toInt());
 		newCyl.type.description = map["type"].toString().toStdString();
 		QString mix = map["mix"].toString();
 		newCyl.gasmix.o2.permille = parseGasMixO2(mix);
 		newCyl.gasmix.he.permille = parseGasMixHE(mix);
 		sanitize_gasmix(newCyl.gasmix);
-		if (prefs.units.pressure == units::BAR)
-			newCyl.start.mbar = map["pressure"].toInt() * 1000;
-		else
-			newCyl.start.mbar = psi_to_mbar(map["pressure"].toInt());
+
 		int useIndex = map["use"].toInt();
 		newCyl.cylinder_use = (enum cylinderuse)useIndex;
 		d->cylinders.add(d->cylinders.size(), newCyl);
@@ -1488,8 +1487,7 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 	if (!diveplan.is_empty()) {
 		deco_state_cache cache;
 		struct deco_state plan_deco_state;
-		plan(&plan_deco_state, diveplan, d, dcNr, 60, cache, true, true, nullptr);
-
+		plan(&plan_deco_state, diveplan, d, dcNr, 60, cache, true, true, nullptr);			
 		if (shouldComputeVariations()) {
 			QString variations = computeVariations(plan_copy, plan_deco_state, nullptr);
 			if (!variations.isEmpty()) {
--- a/mobile-widgets/qml/DivePlannerEdit.qml
+++ b/mobile-widgets/qml/DivePlannerEdit.qml
@@ -22,6 +22,8 @@ TemplatePage {
 	ListModel { id: segmentListModel }
 	property var gasNumberModel: []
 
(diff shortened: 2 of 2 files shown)


qt-models/diveplannermodel.cpp
```diff
@@ -1543,7 +1545,21 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 		deco_state_cache cache;
 		struct deco_state plan_deco_state;
 		plan(&plan_deco_state, diveplan, d, dcNr, 60, cache, true, true);
-
+		//for each cylinder in dive, reset the start pressure
+		for (size_t i = 0; i < d->cylinders.size(); ++i) {
```
mikeller: This is indeed not ideal - have you been able to figure out why the start pressure of the actual cylinder is lost when you are adding it to the model?

qt-models/diveplannermodel.cpp
```diff
@@ -1543,7 +1545,21 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
...
+			d->cylinders[i].end.mbar = d->cylinders[i].end.mbar - modified_amount;
+			d->cylinders[i].start.mbar = d->cylinders[i].start.mbar - modified_amount;
+		}
+		//need a way to determine errors..
+		std::string old_notes = d->notes;
+		planner_error_t error;
+		error = PLAN_OK;
+		diveplan.add_plan_to_notes(*d, true, error);
+		//If d->notes has a different beginning ~30 characters, then this is a planner error
+		if (d->notes.substr(0, 30) != old_notes.substr(0, 30))
```
mikeller: Not sure that this does what you want it to do - the error message is actually swapped in instead of the plan based on the `error` that is passed in.

qt-models/diveplannermodel.cpp
fnightfox: In the dive planner the cylinders are reset whenever "reset_cylinders" is called. The main question I have is what the desktop planner does to make this function reset to the set values. I am still investigating that, but this was the quickest option to get something that worked.

qt-models/diveplannermodel.cpp
fnightfox: Yes. I have no way to know if an error happened during plan generation, and if so I want that error message to be set as the result instead of whatever is set with the second generation (there is an error flag that propagates through the system, but it is lost on return from plan).

mikeller: #4603 should help with this.

fnightfox:
The change such that the error message is returned would have solved the issue with the previous version of this PR, but I realized I was doing something wrong with setting up the cylinders. I thought the size, as held in the type, was the filled size (such as 100cuft for an HP100), but not I believe this to be the "wet" size (the actual internal capacity of the tank). Now that I understand that, I believe I have fixed the issue by just changing the working pressure. Compared to the desktop version, it is accurate to ~1-3%, which I can write off to rounding errors or differences in settings. 
I feel silly for making that mistake and not realizing it until now, but you live and you learn. This version should be correct.

qt-models/diveplannermodel.cpp
```diff
@@ -1424,6 +1424,7 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 	d->dcs[0].divemode = (enum divemode_t)diveMode;
 
 	// Populate cylinders from QML data
+	std::vector<int> cylinder_start_pressures;
```
mikeller: Not quite sure why you are still keeping this - I don't think it is needed.

qt-models/diveplannermodel.cpp
```diff
@@ -1432,15 +1433,18 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 		pressure_t pressure = type_info.second;
 		newCyl.type.size = size;
 		newCyl.type.workingpressure = pressure;
+		if (prefs.units.pressure == units::BAR)
+			newCyl.start.mbar = map["pressure"].toInt() * 1000;
+		else
+			newCyl.start.mbar = psi_to_mbar(map["pressure"].toInt());
```
mikeller: I don't think setting the start pressure is needed here - this will be reset to the working pressure at the beginning of `plan()`. And setting it here will be misleading for people reading the code.

mikeller:
> I thought the size, as held in the type, was the filled size (such as 100cuft for an HP100), but not I believe this to be the "wet" size (the actual internal capacity of the tank).

Yes, that is correct - Subsurface always uses the water capacity to track cylinder volume, even when the units are set to imperial.

qt-models/diveplannermodel.cpp
fnightfox: You are correct. I have removed this.

qt-models/diveplannermodel.cpp
fnightfox: You are correct. I have removed this.

mikeller (approved):

PR #4610, opened 2025-09-29
Convert tankvolume when importing/exporting UDDF

schmittc:
UDDF uses SI units which means cubic meters for volumes. As subsurface uses liters internally we need to convert the unit accordingly during import/export

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Subsurfaces uses liters for tank volumes while UDDF uses SI units.
This PR therefore converts cubic meters to liters on import (and vice vcerso for export)

### Additional information:
A uddf file for testing can be found here: https://groups.google.com/group/subsurface-divelog/attach/9568577a7b3d/test.uddf?part=0.1


mikeller:
Thanks for supplying the test data. Have a look at `exportUDDF()` and `importApdInspirationUddf()` in `tests/testparse.cpp`, it should be relatively easy to use this to add tests for this functionality (and any other changes that might unintentionally change the behaviour when importing / exporting this particular format.

schmittc:
`exportUDDF()` actually already contains a test as the file `test40.xml` contains a tank which is first exported to UDDF and then the UDDF is imported back in again (and there the conversion seems to work). Not sure which additional test you have in mind?

mikeller:
> `exportUDDF()` actually already contains a test as the file `test40.xml` contains a tank which is first exported to UDDF and then the UDDF is imported back in again (and there the conversion seems to work). Not sure which additional test you have in mind?

Good point. But looking at `exportUDDF()`, it seems that in its current state this really only tests that importing and then exporting an UDDF will yield an identical file:

https://github.com/subsurface/subsurface/blob/6db3c5e0c646b2988fce48d21674adc114336f1f/tests/testparse.cpp#L422-L435

Unfortunately this does nothing to catch problems like the one that you are fixing with this pull request, because both the importer and the exporter are wrong. In fact, the 2 liter tanks in `test40.xml` are translated into 2000 liter tanks in the exported UDDF:

https://github.com/subsurface/subsurface/blob/6db3c5e0c646b2988fce48d21674adc114336f1f/dives/test40.xml#L7

```
<tankdata>
          <link ref="mix(100/0)"/>
          <tankvolume>2.0</tankvolume>
          <tankpressurebegin>19000000</tankpressurebegin>
          <tankpressureend>13000000</tankpressureend>
        </tankdata>
```

So an improvement will be to add one of the generated UDDF files to the repository, and have the test compare the generated UDDF to the reference in the repository. And in a second step, the Subsurface XML resulting from the import should be stored as well, and the result of the import during the test compared to it.

mikeller (approved):

PR #4611, opened 2025-09-29
Desktop: Improve the export UI.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Disable the export options if the currently selected format does not
support them.


atdotde:
Thank you for implementing this. I had thought of it as well but did not realise it could be done in the .ui and was too lazy to do it in the C++ code.

mikeller:
I found that when I was searching for how it was done for the units drop down. 😆

PR #4616, opened 2025-10-11
Modify the mobile planner to fix a bug when drop to first depth is en…

fnightfox:
…abled

The planner has an indexing error during the save function on mobile. When drop to first depth is enabled, the first segment gets duplicated. This commit consolidates the functions called for save and preview, which also resolves this bug.

- [X] Bug fix
- [ ] Functional change
- [ ] New feature
- [X] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This pull request modifies the UI of the mobile dive planer to address the bug where the first segment is duplicated on save if drop-stone-mode is enabled. This is caused by an indexing error in the save function. The resolution to this was to clean the code up and consolidate the preview and save functions, as they were mostly identical.

### Changes made:
1) DivePlannerEdit.qml updated to consolidate preview and save functionality, ensuring that the indexing issue present in the save function is NOT present.

### DIFF
--- a/mobile-widgets/qml/DivePlannerEdit.qml
+++ b/mobile-widgets/qml/DivePlannerEdit.qml
@@ -33,7 +33,7 @@ TemplatePage {
 		gasNumberModel = newList;
 	}
 
-	function updateLivePlanInfo() {
+	function generatePlan(savePlan = false) {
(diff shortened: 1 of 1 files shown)


mikeller (approved):
Nice, good work on consolidating!

PR #4630, opened 2025-11-05
Export dive sites as KML

atdotde:
Offer to export the dive sites as a KML file that can be opened in Google Earth and similar.

This is user request from the google group.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/Documentation/user-manual.txt
+++ b/Documentation/user-manual.txt
@@ -3242,8 +3242,8 @@ A dive log or part of it can be saved in several formats:
 
 * _Subsurface XML_ format. This is the native format used by _Subsurface_.
 
-* _Subsurface dive sites XML_. Export the dive sites in XML format. This allows the
-  sharing of dive sites with other divers or software.
+* _Dive sites_. Export the dive sites in Subsurface XML format. This allows the
+  sharing of dive sites with other divers or software. By selecting the "as KML" option, you can also export the dive sites to a file that can be imported to Google Earth and similar programs.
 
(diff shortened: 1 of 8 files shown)


mikeller (approved):

mikeller:
This definitely provides user value.
But the approach to add custom code for each new XML format seems to be not very sustainable - it might be better to implement this with an XSLT transformation.

And this applies even more to the existing  'Subsurface dive sites XML' - this pretty much consists of code copy / pasted from `save_dives_buffer()`, making the transformation trivial.

mikeller:
And it could probably do with some tests - import / export is very easy to test.

atdotde:
While I agree in principle with what you say I would merge this now and postpone the changes to another day. 
For the tests that might be in the near future when I sit down and produce a number of sample exports and wrap them with corresponding test code. For turning all this into xslt, the day would be in the farther future as my understanding of xslt is indistinguishable from 0 short googling of course took me to StackOverflow where I found https://stackoverflow.com/questions/6112874/how-to-replace-a-tag-with-another-tag-in-xml-using-xsl
And yes, I could probably twist this example to transform our sites export to kml but my guess would be that the resulting code ends up to me longer and even less maintainable than what this PR proposes (assuming that reading comprehension of  xslt is not exactly wide spread amongst active Subsurface developers). Thinking about it, a sed script would probably be even easier... or some regular expressions.

PR #4633, opened 2025-11-10
CICD: Switch from code.qt.io to github.com.

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

code.qt.io has been down intermittently but more and more often for the
last month or so. Try switching to github.com as the source of
qtlocation, to try and get back to reliable CICD builds.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -329,8 +329,8 @@ if [ "$PLATFORM" = Linux ] && [[ $QT_VERSION == 5* ]] ; then
 		rm -rf "$INSTALL_ROOT"/include/QtLocation > /dev/null 2>&1
 		rm -rf "$INSTALL_ROOT"/include/QtPositioning > /dev/null 2>&1
 
-		git clone --branch "v$QT_VERSION" https://code.qt.io/qt/qtlocation.git --depth=1 $QTLOC_GIT ||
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
yep, that looks like the right thing to do.

dirkhh (commented):
oops, I take it back... the snapcraft URL was transformed incorrectly -- let me fix that

mikeller:
@dirkhh: Interesting that this is now working. I had created this a couple of weeks ago when code.qt.io was down for days on end, but at the time the builds did not seem result in working map support, as it looked like the https://github.com/qt/qtlocation repository did not contain all of the required headers.

So I think that this change might need to some monitoring, to make sure we consistently get what we need from https://github.com/qt/qtlocation

dirkhh:
Your conversion of the path was wrong.
GitHub isn't using gitolite for it's web UI, so the path that you used to get to the headers simply got you a 404 error.
Hence no private headers, hence no working maps.

mikeller:
> Your conversion of the path was wrong. GitHub isn't using gitolite for it's web UI, so the path that you used to get to the headers simply got you a 404 error. Hence no private headers, hence no working maps.

Was that the snapcraft build that you fixed? I never got around to looking into that one.
The bog standard build for linux managed to clone the repository and switch to the correct branch fine, but the contents were not recognised as the QtLocation headers during the build, so no maps were built. 😑

PR #4634, opened 2025-11-13
Mobile: Add CCR Support for the Dive Planner.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### Changes made:

- Add Rebreather Support to the Dive Planner;
- Planner add CCR default ppO2;
- Improve planner settings grouping;
- Show dive mode for first segment (except for OC).

### Mentions:

@fnightfox: I finally got around to doing this, please review.
It actually turned out to be less involved than expected to make this work, probably because the changes in #4616 already fixed some of the issues. But then I ended up doing quite a few minor improvements to make this work well.

One problem with the mobile planner that I discovered when working on this is concerns the planner settings: All of the settings values are stored as metric values - this means that when subsurface is set to imperial units, the values need to be converted between the UI and the backend. I suspect that this will be quite gnarly and require a lot of boilerplate to fix in the frontend, and we could probably reuse this between mobile and desktop if we pushed this into the backend, but I can't quite see how to best do this - maybe you have an idea?

### DIFF
--- a/Documentation/user-manual_de.txt
+++ b/Documentation/user-manual_de.txt
@@ -2267,7 +2267,7 @@ Zeigen Sie Änderungen an der verwendeten Gasflasche an, indem Sie die Gaswechse
 
 Die für den Plan verwendeten Flaschen müssen in der Tabelle „Verfügbare Gase“ eingetragen werden. Wählen Sie in der Spalte „Typ“ die passende Flaschengröße aus, indem Sie auf eine Zelle in dieser Spalte doppelklicken. Standardmäßig sind zahlreiche Größen verfügbar. Sie können eine neue Flaschengröße erstellen, indem Sie diese in das Textfeld eingeben. Flaschengröße, Startdruck und Standard-Umschalttiefen werden automatisch initialisiert. Geben Sie die Gaszusammensetzung (z. B. Helium- und Sauerstoffgehalt) und – bei der Planung eines CCR-Tauchgangs – die beabsichtigte Verwendung als Verdünnungsgas oder OC-Bailout-Gas an. Bei der Planung von CCR-Tauchgängen wird ein Segment als geschlossenes Kreislaufsegment berechnet, wenn das verwendete Gas ein Verdünnungsgas ist, und als offenes Kreislaufsegment, wenn es sich um ein OC-Gas handelt. Wenn der Benutzer in den Einstellungen des „Tech-Setups“ die Option „Verwendung von offenem Kreislaufgas als Bailout zulassen“ aktiviert, wird eine zusätzliche Spalte angezeigt, in der der Benutzer den „Tauchmodus“ für jedes mit einem „OC-Gas“ getauchte Segment auswählen kann.
 
-image::images/Planner_OC_gas_for_CCR.png["Abbildung: Tauchplaner: OC Gas als Bailout",align="center"]
+image::images/Planner_OC_gas_for_CCR.png["Abbildung: Tauchplaner: OC-Gas als Bailout",align="center"]
 
 Für CCR-Tauchgänge wird in der Tabelle „Punkte des Tauchplaners“ eine zusätzliche Spalte angezeigt, in der der Benutzer den Sollwert für jedes Segment mit geschlossenem Kreislauf festlegen kann.
 
--- a/core/planner.cpp
+++ b/core/planner.cpp
@@ -347,7 +347,7 @@ struct gaschanges {
 	int gasidx;
 };
 
-// Return new setpoint if cylinderi is a setpoint change an 0 if not
+// Return new setpoint if cylinderid is a setpoint change and 0 if not
 
 static int setpoint_change(struct dive *dive, int cylinderid)
 {
--- a/core/plannernotes.cpp
+++ b/core/plannernotes.cpp
@@ -346,7 +346,7 @@ void diveplan::add_plan_to_notes(struct dive &dive, bool show_disclaimer, planne
 							newgasmix.name().c_str(), temp.c_str());
 					} else {
 						buf += format_string_std("<td style='padding-left: 10px; color: red; float: left;'><b>%s %s</b></td>",
-							newgasmix.name().c_str(), dp->divemode == UNDEF_COMP_TYPE || dp->divemode == nextdp->divemode ? "" : translate("gettextFromC", divemode_text_ui[nextdp->divemode]));
+							newgasmix.name().c_str(), dive.dcs[0].divemode == OC || dp->divemode == UNDEF_COMP_TYPE || dp->divemode == nextdp->divemode ? "" : translate("gettextFromC", divemode_text_ui[nextdp->divemode]));
 						if (isascent && (get_he(lastprintgasmix) > 0)) { // For a trimix gas change on ascent, save ICD info if previous cylinder had helium
 							if (isobaric_counterdiffusion(lastprintgasmix, newgasmix, &icdvalues)) // Do icd calulations
 								icdwarning = true;
@@ -367,7 +367,7 @@ void diveplan::add_plan_to_notes(struct dive &dive, bool show_disclaimer, planne
 						buf += format_string_std("<td style='padding-left: 10px; color: red; float: left;'><b>%s %s</b></td>", gasmix.name().c_str(), temp.c_str());
 					} else {
 						buf += format_string_std("<td style='padding-left: 10px; color: red; float: left;'><b>%s %s</b></td>", gasmix.name().c_str(),
-							   lastdivemode == UNDEF_COMP_TYPE || lastdivemode == dp->divemode ? "" : translate("gettextFromC", divemode_text_ui[dp->divemode]));
+							   dive.dcs[0].divemode == OC || dp->divemode == UNDEF_COMP_TYPE || lastdivemode == dp->divemode ? "" : translate("gettextFromC", divemode_text_ui[dp->divemode]));
 						if (get_he(lastprintgasmix) > 0) {  // For a trimix gas change, save ICD info if previous cylinder had helium
 							if (isobaric_counterdiffusion(lastprintgasmix, gasmix, &icdvalues))  // Do icd calculations
 								icdwarning = true;
--- a/core/plannernotes.cpp
+++ b/core/plannernotes.cpp
@@ -130,7 +130,7 @@ void diveplan::add_plan_to_notes(struct dive &dive, bool show_disclaimer, planne
 			break;
 		case PLAN_ERROR_NO_SUITABLE_BAILOUT_GAS:
 			message = translate("gettextFromC", "No suitable gas for OC bailout at the planned final depth found in the gaslist. "
-				"Please add an OC gas with an MOD suitable for the planned final depth to enable the generation of a dive plan.");
+				"Please add an OC-gas with an MOD suitable for the planned final depth to enable the generation of a dive plan.");
 
 			break;
 		default:
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -126,12 +126,9 @@ DivePlannerWidget::~DivePlannerWidget()
 void DivePlannerWidget::disableDecoElements(int mode, divemode_t divemode)
 {
 	if (mode == RECREATIONAL) {
-		ui.label_bailout->setVisible(false);
 		ui.bailout->setVisible(false);
 	} else if (mode == VPMB || mode == BUEHLMANN) {
-		bool isRebreatherDive = IS_REBREATHER_MODE(divemode);
-		ui.label_bailout->setVisible(isRebreatherDive);
-		ui.bailout->setVisible(isRebreatherDive);
+		ui.bailout->setVisible(IS_REBREATHER_MODE(divemode));
 	}
 }
 
--- a/desktop-widgets/diveplanner.ui
+++ b/desktop-widgets/diveplanner.ui
@@ -148,13 +148,6 @@
          </property>
         </widget>
        </item>
-       <item row="2" column="2">
-        <widget class="QLabel" name="label_bailout">
-         <property name="text">
-          <string>Rebreather:</string>
-         </property>
-        </widget>
-       </item>
        <item row="3" column="2">
         <widget class="QCheckBox" name="bailout">
          <property name="text">
--- a/desktop-widgets/plannerSettings.ui
+++ b/desktop-widgets/plannerSettings.ui
@@ -213,6 +213,13 @@
                </property>
               </widget>
              </item>
+             <item row="1" column="0" colspan="2">
+              <widget class="QCheckBox" name="drop_stone_mode">
+               <property name="text">
+                <string>Drop to first depth</string>
+               </property>
+              </widget>
+             </item>
             </layout>
             <zorder>descRate</zorder>
             <zorder>descent</zorder>
@@ -287,7 +294,7 @@
             </property>
            </spacer>
           </item>
-          <item row="20" column="1" colspan="2">
+          <item row="19" column="1" colspan="2">
            <widget class="QCheckBox" name="switch_at_req_stop">
             <property name="toolTip">
              <string>Postpone gas change if a stop is not required</string>
@@ -297,7 +304,7 @@
             </property>
            </widget>
           </item>
-          <item row="16" column="1" colspan="2">
+          <item row="15" column="1" colspan="2">
            <widget class="QCheckBox" name="lastStop">
             <property name="text">
              <string>Last stop at 6m</string>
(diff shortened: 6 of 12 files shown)


fnightfox:
My initial thoughts are to keep things metric on the backend, otherwise we would need to save which unit type and do conversions on all units when the user changes from one to another. Not sure what the best way to handle it on the frontend is, though. I'll think on it and review it more closely in a week (if it's still open, initial quick review looks good) once I return home from travel. 

--Liz

fnightfox:
I am still unsure how the settings should be changed, I think you would be the best person to decide that as you have much more familiarity with the backend structures. I would say for the frontend that unless we notice performance hits keep the conversions there, as they are only needed for a few screens. 

For the CCR implementation, I have a couple of notes on the UI. The settings page looks good to me, but I noticed that if you go from CCR back to Open Circuit, there are settings that you can set in the CCR mode that cannot be changed in open circuit mode and can require you to switch back to CCR to fix. I am not sure what the best way to handle this would be (it could be that the gas use is ignored in OC mode, or the user just has to realize that they need to go back and fix things). 

The second UI item is just that the columns do not line up correctly with the headers on my screen. Mainly the setpoint [bar] label is partially over the gas selection dropdown, and the mix, use, and pressure labels are shifted slightly to the left.

Overall, it seems to work and the code looks good.

mikeller:
> I am still unsure how the settings should be changed, I think you would be the best person to decide that as you have much more familiarity with the backend structures. I would say for the frontend that unless we notice performance hits keep the conversions there, as they are only needed for a few screens.

'Frontend' is definitely what the desktop version is doing, having the conversions in `desktop-widgets/`, where we can't make use of them for mobile. I am less worried about performance, as these are not things that users will be fiddling with often, more about having to re-implement a parallel set of conversions for meter-to-feet, bar-to-psi,... and vice versa, and then maintain them. Being able to move the ones in `desktop-widgets/` to a place where they can be used by desktop and mobile seems like the easy way out. :-)

But I think getting this sorted can be done in a separate pull request, so we can get the CCR support in front of some divers.

> For the CCR implementation, I have a couple of notes on the UI. The settings page looks good to me, but I noticed that if you go from CCR back to Open Circuit, there are settings that you can set in the CCR mode that cannot be changed in open circuit mode and can require you to switch back to CCR to fix. I am not sure what the best way to handle this would be (it could be that the gas use is ignored in OC mode, or the user just has to realize that they need to go back and fix things).

This is actually intentional - my aim here is to make it so that settings that settings that are not available to the user in a certain mode are hidden, and their 'default' value for this mode is used where needed. So it should never be the case that changing a setting in one mode will change the outcome of the calculation in another mode where it is hidden - if this is the case then I missed something, please let me know.
The rationale for this is that most divers on a rebreather will also want to plan open circuit dives from time to time, and resetting all of their rebreather-specific settings when they do this will be bad user experience.
The rationale for hiding (as opposed to disabling the inputs like it's done on the desktop) is that the UI is already quite cramped on mobile, so showing less seems to be more.

> The second UI item is just that the columns do not line up correctly with the headers on my screen. Mainly the setpoint [bar] label is partially over the gas selection dropdown, and the mix, use, and pressure labels are shifted slightly to the left.

Yes. I have not even tried to make this look aligned. I do not know enough about QML and how it controls the UI, and I don't want to use trial and error as I am sure anything that works for the screen sizes on the two test devices that I have will likely look all wrong on most other screen sizes. Is this something that you can help with? If not I am happy to accept that what we have conveys the information it needs to, and use it as is.

fnightfox:
I will try to work on the centering of columns tomorrow. 

For the settings having an impact across modes item, I noticed the following: 
Deco on oc gas selected, switch back to open circuit, but have no gas with an acceptable MOD in the table. Expected behavior would be too generate the plan still, but have the warning present (as it exists now), actual behavior is a refusal to generate a plan. I realize this is a bit of an edge case, but it might be worth changing. 

I also was wondering if it would be possible to make it such that, if no OC gas is specified but the deco on oc gas is selected, could we interpret all dilluent gases as OC (maybe with an additional warning indicating such). I think I heard that it is common to plan to bail out with dilluent gas bottles on simpler dives and this would allow an easy way to view that plan.

fnightfox:
I have made some changes to the tables to address the layout. I had to remove any width filling because of the dynamic visibility of some columns (I am open to someone else who knows how to get that to work correctly adding that back, but I do not know how). In order to shrink the width further, I changed the "Use" column on the cylinder to "Dil" and a checkbox, with the intention that this makes a cylinder as a diluent vs open circuit. Feel free to change that back.

My main concern is that on screens with small resolution the fields will go off the screen, but I have no way to test this.

```diff
--- a/mobile-widgets/qml/DivePlannerEdit.qml
+++ b/mobile-widgets/qml/DivePlannerEdit.qml
@@ -259,20 +259,20 @@ TemplatePage {
 			}
 		}
 		RowLayout {
-			Layout.fillWidth: true
+			//Layout.fillWidth: true
 			spacing: Kirigami.Units.smallSpacing // Reduced spacing
 
 			// Header labels updated for new layout
 			TemplateLabel { text: qsTr("#"); Layout.preferredWidth: Kirigami.Units.gridUnit * 2; font.bold: true }
-			TemplateLabel { text: qsTr("Type"); Layout.preferredWidth: Kirigami.Units.gridUnit * 6; font.bold: true }
-			TemplateLabel { text: qsTr("Mix"); Layout.preferredWidth: Kirigami.Units.gridUnit * 4; font.bold: true }
+			TemplateLabel { text: qsTr("Type"); Layout.minimumWidth: Kirigami.Units.gridUnit * 8; Layout.maximumWidth: Kirigami.Units.gridUnit * 8; font.bold: true }
+			TemplateLabel { text: qsTr("Mix"); Layout.preferredWidth: Kirigami.Units.gridUnit * 3; font.bold: true }
 			TemplateLabel {
-				text: qsTr("Use");
-				Layout.preferredWidth: Kirigami.Units.gridUnit * 5;
+				text: qsTr("Dil");
+				Layout.preferredWidth: Kirigami.Units.gridUnit * 3;
 				visible: overallDivemode.currentIndex == 1
 				font.bold: true
 			}
-			TemplateLabel { text: qsTr("[%1]").arg(pressureUnit); Layout.fillWidth: true; font.bold: true }
+			TemplateLabel { text: qsTr("[%1]").arg(pressureUnit); Layout.preferredWidth: Kirigami.Units.gridUnit * 4; font.bold: true }
 			Item { Layout.preferredWidth: Kirigami.Units.iconSizes.medium }
 		}
 
@@ -284,7 +284,7 @@ TemplatePage {
 			model: cylinderListModel
 
 			delegate: RowLayout {
-				width: cylinderListView.width
+				//width: cylinderListView.width
 				spacing: Kirigami.Units.smallSpacing // Reduced spacing
 
 				TemplateLabel {
@@ -295,7 +295,8 @@ TemplatePage {
 				}
 				TemplateComboBox {
 					id: typeBox
-					Layout.preferredWidth: Kirigami.Units.gridUnit * 10
+					Layout.minimumWidth: Kirigami.Units.gridUnit * 8
+					Layout.maximumWidth: Kirigami.Units.gridUnit * 8
 					model: cylinderTypesModel
 					currentIndex: model.indexOf(type)
 					onActivated: {
@@ -310,7 +311,7 @@ TemplatePage {
 				}
 				TemplateTextField {
 					id: mixField
-					Layout.preferredWidth: Kirigami.Units.gridUnit * 4
+					Layout.preferredWidth: Kirigami.Units.gridUnit * 3
 					text: mix
 					onTextChanged: {
 						cylinderListModel.setProperty(index, "mix", text);
@@ -337,19 +338,20 @@ TemplatePage {
 					validator: RegExpValidator { regExp: /(EAN100|EAN\d\d|AIR|100|\d{0,2}|\d{0,2}\/\d{0,2})/i }
 					onActiveFocusChanged: cylinderListView.interactive = !activeFocus
 				}
-				TemplateComboBox {
-					Layout.preferredWidth: Kirigami.Units.gridUnit * 5
-					model: [ qsTr("OC-gas"), qsTr("diluent") ]
-					currentIndex: use
-					visible: overallDivemode.currentIndex == 1
-					onCurrentIndexChanged: {
-						cylinderListModel.setProperty(index, "use", currentIndex);
-						generatePlan();
-					}
-				}
+				TemplateCheckBox {
+                    Layout.preferredWidth: Kirigami.Units.gridUnit * 3
+                    // Map the model's 'use' property (0 or 1) to the checkbox state (false or true)
+                    checked: use === 1
+                    visible: overallDivemode.currentIndex == 1
+                    onClicked: {
+                        // Update 'use': if checked is true, set 'use' to 1 (Diluent); otherwise, set to 0 (OC-gas)
+                        cylinderListModel.setProperty(index, "use", checked ? 1 : 0);
+                        generatePlan();
+                    }
+                }
 				TemplateTextField {
 					id: pressureField
-					Layout.fillWidth: true
+					Layout.preferredWidth: Kirigami.Units.gridUnit * 4
 					text: pressure.toString()
 					validator: IntValidator { bottom: 0; top: 10000 }
 					onTextChanged: {
@@ -403,33 +405,33 @@ TemplatePage {
 		}
 
 		RowLayout {
-			Layout.fillWidth: true
+			//Layout.fillWidth: true
 			spacing: Kirigami.Units.gridUnit
 
 			TemplateLabel {
 				text: qsTr("Depth [%1]").arg(depthUnit);
-				Layout.fillWidth: true;
+				Layout.preferredWidth: Kirigami.Units.gridUnit * 4
 				font.bold: true
 			}
 			TemplateLabel {
-				text: qsTr("Duration [min]");
-				Layout.fillWidth: true;
+				text: qsTr("Time [min]");
+				Layout.preferredWidth: Kirigami.Units.gridUnit * 4
 				font.bold: true
 			}
 			TemplateLabel {
 				text: qsTr("Gas");
-				Layout.fillWidth: true;
+				Layout.preferredWidth: Kirigami.Units.gridUnit * 5
 				font.bold: true;
 			}
 			TemplateLabel {
 				text: qsTr("Setpoint [bar]");
-				Layout.fillWidth: true;
+				Layout.preferredWidth: Kirigami.Units.gridUnit * 5
 				font.bold: true;
 				visible: overallDivemode.currentIndex == 1;
 			}
 			TemplateLabel {
 				text: qsTr("Dive Mode");
-				Layout.fillWidth: true;
+				Layout.preferredWidth: Kirigami.Units.gridUnit * 6
 				font.bold: true;
 				visible: overallDivemode.currentIndex == 2;
 			}
@@ -444,11 +446,11 @@ TemplatePage {
 
 			model: segmentListModel
 			delegate: RowLayout {
-				width: segmentListView.width
+				//width: segmentListView.width
 				spacing: Kirigami.Units.gridUnit
 
 				TemplateTextField {
-					Layout.fillWidth: true
+					Layout.preferredWidth: Kirigami.Units.gridUnit * 4
 					text: depth.toString()
 					validator: IntValidator { bottom: 0; top: 900 }
 					onTextChanged: {
@@ -459,7 +461,7 @@ TemplatePage {
 				}
 
 				TemplateTextField {
-					Layout.fillWidth: true
+					Layout.preferredWidth: Kirigami.Units.gridUnit * 4
 					text: duration.toString()
 					validator: IntValidator { bottom: 1; top: 999 }
 					onTextChanged: {
@@ -470,8 +472,7 @@ TemplatePage {
 				}
 
 				TemplateComboBox {
-					Layout.fillWidth: true
-					Layout.minimumWidth: Kirigami.Units.gridUnit * 6
+					Layout.preferredWidth: Kirigami.Units.gridUnit * 5
 					model: gasNumberModel
 					currentIndex: gas
 					onCurrentIndexChanged: {
@@ -482,7 +483,7 @@ TemplatePage {
 				}
 
 				TemplateTextField {
-					Layout.fillWidth: true
+					Layout.preferredWidth: Kirigami.Units.gridUnit * 5
 					text: cylinderListModel.get(gas) && cylinderListModel.get(gas).use === 1 ? (setpoint / 1000.0).toFixed(2) : ""
 					validator: DoubleValidator {
 						bottom: 0.16;
@@ -502,8 +503,7 @@ TemplatePage {
 				}
 
 				TemplateComboBox {
-					Layout.fillWidth: true
-					Layout.minimumWidth: Kirigami.Units.gridUnit * 6
+					Layout.preferredWidth: Kirigami.Units.gridUnit * 6
 					model: [ qsTr("OC"), qsTr("pSCR") ]
 					// Skip CCR (value === 1) as it's not applicable for pSCR mode
 					currentIndex: divemode === 2 ? 1 : divemode
```

mikeller:
> For the settings having an impact across modes item, I noticed the following: Deco on oc gas selected, switch back to open circuit, but have no gas with an acceptable MOD in the table. Expected behavior would be too generate the plan still, but have the warning present (as it exists now), actual behavior is a refusal to generate a plan. I realize this is a bit of an edge case, but it might be worth changing.

Yes, this is a bug that I introduced in #4625. Fixed in #4639. Thanks for pointing it out.

> I also was wondering if it would be possible to make it such that, if no OC gas is specified but the deco on oc gas is selected, could we interpret all dilluent gases as OC (maybe with an additional warning indicating such). I think I heard that it is common to plan to bail out with dilluent gas bottles on simpler dives and this would allow an easy way to view that plan.

I don't think using this simplification is appropriate for the planner, as this would need a way for the user to configure which diluent gases they will not be able to use, or are not willing to use during OC bailout - or else the planner will come up with an incorrect plan for OC bailout.
Moreover, on practically all modern tec dive computers, the user will have to enter any diluent gasmix that they want to use for bailout as well twice - once in the diluent list, and once more in the OC gas list. So the current way results in gas lists for plans that are more closely aligned to what they will be getting when downloading the dive from their dive computer.

In the desktop version users have a similar option in the preferences ('Allow open circuit gas to be used as diluent for CCR') - this is essentially a reverse of this, where OC gases can be used as diluent, but a 'diluent' gas is still only usable as a diluent. This avoids the pitfall of diluents being selected for bailout even if the tank cannot be breathed from directly. But it requires the user to select the OC / diluent use for each segment of a CCR dive. I tried this for mobile, but could not make it work in a useful way with the screen width available.

> I have made some changes to the tables to address the layout. I had to remove any width filling because of the dynamic visibility of some columns (I am open to someone else who knows how to get that to work correctly adding that back, but I do not know how). In order to shrink the width further, I changed the "Use" column on the cylinder to "Dil" and a checkbox, with the intention that this makes a cylinder as a diluent vs open circuit. Feel free to change that back.

I like it, it looks good, thanks.

Using 'Dil' and a checkbox feels way more natural than the drop down.

fnightfox:
I did a quick look at it, and I think it looks good now. No further comments, thank you for this!

mikeller:
No problem, thanks for the review.

Do you want to have a stab at the unit conversion for the mobile planner settings, or shall I?

fnightfox:
> No problem, thanks for the review.
> 
> Do you want to have a stab at the unit conversion for the mobile planner settings, or shall I?

If you could take an initial stab at it I will be happy to assist. I am still trying to get a working gas blender that has all the features I want in there.

mikeller:
> > No problem, thanks for the review.
> > Do you want to have a stab at the unit conversion for the mobile planner settings, or shall I?
> 
> If you could take an initial stab at it I will be happy to assist. I am still trying to get a working gas blender that has all the features I want in there.

👍
I like this plan - been looking for a decent free gas blending calculator for ages!

mikeller:
@fnightfox: I found an interesting bug last night when trying to plan a more complex dive:

image

(note the 4 10 minute segments making up the 58 minute bottom time that are not showing)

This also seems to affect the gasmix list, which also gets cut off (on my device) after the first 3 gases:

image

Since you have already worked on the horizontal spacing, you might have an idea what is needed to make these lists expand as needed when more elements are added?

fnightfox:
> @fnightfox: I found an interesting bug last night when trying to plan a more complex dive:
> image
> 
> (note the 4 10 minute segments making up the 58 minute bottom time that are not showing)
> 
> This also seems to affect the gasmix list, which also gets cut off (on my device) after the first 3 gases:
> image
> 
> Since you have already worked on the horizontal spacing, you might have an idea what is needed to make these lists expand as needed when more elements are added?

Those lists should be scrollable. I chose to limit the size to try to keep the graph on the screen when possible.

mikeller:
@fnightfox 

> Those lists should be scrollable. I chose to limit the size to try to keep the graph on the screen when possible.

Yes, they are indeed. But since there is no visual scroll bar except for the one for scrolling the entire page, this is hard to discover.

Not sure I would put this much emphasis on having the graph visible without scrolling down on the page. To me, planning a dive is normally a sequential process, and the page layout allows me to enter information, and then scroll down and enter subsequent information without having to go back to the previous information. So I will naturally get to the graph by the time I have entered all of the information.

On the flip side, not being able to see the whole list of segments as entered while looking at the graph seems to be a downside to me.

So from my point of view, we should either make a scroll bar for the scrollable sections visible, or make them expand instead of being scrollable.

We could probably also improve the information display if we replaced the 'Gas X' labels for the selected gas in the segment list with a '21/0', '16/45 (d)',... for the selected gas in the segment list, allowing the user to compare the entered segment list with the dive plan without having to consult the gasmix list.

PR #4636, opened 2025-11-16
Use new divelogs API for export to Divelogs.de

vigge93:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The old API using /DivelogsDirectImport.php is deprecated. Switch to new REST API using /api/dives. The new API uses JSON instead of XML, which required parsing from XML to JSON to be implemented.

Since we no longer upload a ZIP file to the server, the temporary storage of the dive data is no longer necessary.

This also fixes a problem where a dive having multiple tanks would result in incorrect pressures for the additional tanks, due to the old divelogs API interpreting the pressures for the additional tanks as mbar instead of bar.

### Changes made:
1) Update uploadDiveLogsDE to use new REST API.


core/uploadDiveLogsDE.cpp
```diff
@@ -172,45 +156,58 @@ bool uploadDiveLogsDE::prepareDives(const QString &tempfile, bool selected)
...
+		}
+		xmlDocDumpMemory(transformed, (xmlChar **)&membuf, &streamsize);
+		xmlFreeDoc(doc);
+		xmlFreeDoc(transformed);
+
 		/*
-		 * Save the XML document into a zip file.
+		 * Fix the JSON to match divelogs.de expectations:
+		 *   - remove the two null tanks that divelogs-export.xslt always creates
+		 *   - remove the xml declaration and {"dive": ... wrapping
```
vigge93: For some reason it seems that libxslt ignores the `omit-xml-declaration="yes"` directive in the xslt files. I don't have any experience with this, so maybe someone else has a better idea of what is going on?

vigge93: Found the answer in this email chain from 2009:
https://mail.gnome.org/archives/xslt/2009-July/msg00007.html

Apparently, it's a problem with `xmlDocDumpMemory`, and we should instead use `xsltSaveResultToString`

core/uploadDiveLogsDE.cpp
```diff
@@ -172,45 +156,60 @@ bool uploadDiveLogsDE::prepareDives(const QString &tempfile, bool selected)
...
-			     qPrintable(QDir::toNativeSeparators(tempfile)), ze, se, zip_strerror(zip));
-		return false;
-	}
-	return true;
-}
+		std::string json = std::string(membuf, streamsize);
+		free(membuf);
+		std::string tank_placeholder = "\"tanks\":[null,null,";
+		size_t start_pos = json.find(tank_placeholder);
+		json.replace(start_pos, tank_placeholder.length(), "\"tanks\":[");
```
mikeller: **suggestion:** I think it will be better if we fix these problems in the output in the XSLT that does the conversion in the first place - using string manipulation to modify is not very robust - what if some platforms started inserting space between entities into the JSON? And having this in here will also make it hard to test the actual 'divelogs JSON' that is produced by Subsurface.

mikeller (commented):
Thanks for this - always good to move off legacy APIs.

I think it will be good to add some tests - testing XSLT transformations is easy enough, see `tests/testparse.cpp`. This will also ensure that this keeps working across improvements in the XSLT.

core/uploadDiveLogsDE.cpp
vigge93: Fixed it. The JSON parsing now supports two attributes: ignore-root="yes" and type="array". Also updated the number parsing since it didn't handle numbers starting with 0.xxx.

vigge93:
> Thanks for this - always good to move off legacy APIs.
> 
> I think it will be good to add some tests - testing XSLT transformations is easy enough, see `tests/testparse.cpp`. This will also ensure that this keeps working across improvements in the XSLT.

I added a test for the Subsurface XML -> DivelogsDE XML transformation. However, I didn't find any method for transforming non-Subsurface XML files for testing the DivelogsDE XML -> JSON transformation. Do you know if that exists or if we have to implement it?

mikeller:
> I added a test for the Subsurface XML -> DivelogsDE XML transformation. However, I didn't find any method for transforming non-Subsurface XML files for testing the DivelogsDE XML -> JSON transformation. Do you know if that exists or if we have to implement it?

Thank you for adding the tests.
I think testing against the intermediate XML gets us almost there - the transformation to JSON is done with a well-known and stable library, so it's unlikely that we'll be changing this part any time soon.
If we wanted to test against the actual JSON we'd probably have to copy the code that does the second transformation from `core/uploadDiveLogsDE.cpp` - which does not seem to make sense as this still involves quite a bit of manual assembly of the array.

vigge93:
> > I added a test for the Subsurface XML -> DivelogsDE XML transformation. However, I didn't find any method for transforming non-Subsurface XML files for testing the DivelogsDE XML -> JSON transformation. Do you know if that exists or if we have to implement it?
> 
> Thank you for adding the tests. I think testing against the intermediate XML gets us almost there - the transformation to JSON is done with a well-known and stable library, so it's unlikely that we'll be changing this part any time soon. If we wanted to test against the actual JSON we'd probably have to copy the code that does the second transformation from `core/uploadDiveLogsDE.cpp` - which does not seem to make sense as this still involves quite a bit of manual assembly of the array.

Alright, I agree, let's leave it as is. Anything else needed before we can get this merged?

mikeller (approved):

PR #4640, opened 2025-11-26
fix MTP and RAW support on macOS

dirkhh:
This is a bit of an awkward combo of things, sorry about that.

- breaks out the building of the dependencies for macOS into its own script
- in the process, fixes the libmtp build so that the Garmin dive computers supporting MTP work again
- notice that libraw_r hadn't been working with the macOS binaries, so added that as a dependency as well
- remove more Homebrew contamination in the process and update the versions of several dependencies (libssh, libcurl, libz) and switch more of the dependencies to cmake builds
- add a Qt 6 action for macOS and build arm64 and x86_64 binaries
- completely redo the handling of BASE_SDK in the macOS builds, as what we did in the past was the exact opposite of what Apple tells us to do
- macOS Qt 5 builds require macOS 12.3 or newer, the Qt 6 builds require macOS 13

In the process I ran into a hilarious set of oddities and corner cases of macOS runners at GitHub.
I tried to clean up many of the things that I hacked into place years ago and didn't fully understand and no one dared to touch, since.

In that same direction, this tries to add some caching for the dependencies and the packages downloaded from Homebrew (we do need a couple of the tools that aren't pre-installed). This doesn't appear to work as intended, yet, but I left it in the code as it doesn't hurt the way it is and maybe after a few builds happen I get a better idea what is failing.


packaging/macosx/make-package.sh
```diff
@@ -39,7 +33,11 @@ cmake $OLDER_MAC_CMAKE \
 	-DFTDISUPPORT=ON \
 	.
 
-LIBRARY_PATH=${DIR}/install-root/lib make -j
+if [ "$RUNNER_OS" = "macOS" ]; then
+	LIBRARY_PATH=${DIR}/install-root/lib make -j 2
```
mikeller: **question:**  Why are we using a fixed number of threads here? Wouldn't it be more effective to allow the system to determine how many cores we're running on?

See https://github.com/subsurface/subsurface/pull/4606/files#diff-fe04c5ddd34745c53a203fb037da13aeaea8515b2a7aede99c26c5cba943a783R129

.github/workflows/mac-qt6.yml
```diff
@@ -0,0 +1,162 @@
...
+
+jobs:
+  build-documentation:
+    uses: ./.github/workflows/documentation-build.yml
+
+  build:
+    needs: build-documentation
+    strategy:
+      matrix:
+        include:
```
mikeller: **suggestion:** Instead of creating a more-or-less duplicate of the `mac.yml`, I think we should add the Qt5 build as another entry into the matrix - this will make it so much more maintainable.

.github/workflows/mac-qt6.yml
```diff
@@ -0,0 +1,162 @@
...
+      id: version_number
+      uses: ./.github/actions/manage-version
+      with:
+        nightly-builds-secret: ${{ secrets.NIGHTLY_BUILDS }}
+
+    - name: cache deps
+      id: cache-deps
+      uses: actions/cache@v4
+      with:
+        path: ${{ github.workspace }}/../install-root
```
mikeller: **question:** Is there a way to be more specific with what we are caching? This seems to include the built binary, so if something in the build goes wrong, but we don't raise it through the multiple layers of scripts that we are using, we'll be publishing builds of cached artifacts, creating a debugging nightmare.
Alternatively we can probably go and explicitly delete the artifacts after publishing them, to remove this risk.

.github/workflows/mac-qt6.yml
```diff
@@ -0,0 +1,162 @@
...
+    - name: publish pull request artifacts
+      if: github.event_name == 'pull_request'
+      uses: actions/upload-artifact@v5
+      with:
+        name: Subsurface-MacOS-${{ matrix.arch }}-qt6-${{ steps.version_number.outputs.version }}
+        path: ${{ steps.build.outputs.dmg }}
+        compression-level: 0
+
+    # only publish a 'release' on push events (those include merging a PR)
+    - name: upload binaries
```
mikeller: **question:** Since we are adding these artifacts to releases, should we amend the help text for macOS to instruct users which .dmg to use: https://github.com/subsurface/subsurface/blob/master/gh_release_notes_bottom.md#macos

.github/workflows/mac-qt6.yml
```diff
@@ -0,0 +1,162 @@
...
+
+        REG_IMG=$(grep ^created: mp.log | tail -1 | cut -b10-)
+        [ "${{ runner.arch }}" = "X64" ] && CUR_ARCH="x86_64"
+        [ "${{ runner.arch }}" = "ARM64" ] && CUR_ARCH="arm64"
+        IMG="${REG_IMG/Subsurface-/Subsurface-${CUR_ARCH}-qt6-}"
+        mv "$REG_IMG" "$IMG"
+        echo "Created $IMG"
+        echo "dmg=$IMG" >> $GITHUB_OUTPUT
+
+    - name: publish pull request artifacts
```
mikeller: **suggestion:** We should add this to the list of builds to have artifact links added to pull requests - finding artifacts in the actions is a pain in GitHub: https://github.com/subsurface/subsurface/blob/master/.github/workflows/artifact-links.yml

packaging/macosx/make-package.sh
dirkhh: because the GitHub runner has 3 under-powered cores and if you allow it to run with -j it will hang itself and stall until it times out
Took me HOURS to figure out that that was why my builds got stuck.

.github/workflows/mac-qt6.yml
dirkhh: totally agreed. I was trying to keep the Qt 5 one working while figuring out the Qt 6 one. Once this all works I'll figure out if I can combine them. I'm not sure this will be easier to maintain, though, because this is not a clean matrix... and I worry about the amount of special casing. But I'll try

packaging/macosx/make-package.sh
mikeller: Good to know - we'll potentially want to change this for the actual macos build as well then.

.github/workflows/mac-qt6.yml
dirkhh: this should ONLY cache the dependencies, not the build results for Subsurface.
Right now it doesn't cache a thing - so it's just a distraction.

.github/workflows/mac-qt6.yml
dirkhh: approximately 0.1% of our users (who aren't developers themselves - or involved in the development) get their binaries like this from GitHub. The rest download from the website, which won't mention these builds until they are WAY further down the road.

.github/workflows/mac-qt6.yml
dirkhh: this is something that you invented :)
I haven't even looked at how that works. Actually, I didn't even know we had that. 🤣
I'm done for today - just kicked of a desperate last attempt to get this thing to build again, but I'll be hanging out with my family again before these builds complete

mikeller: > Actually, I didn't even know we had that. 🤣

That's what spams pull requests with all of these comments:

image

But on the plus side, if you tell users to go test a pull request, they can actually figure out how to do it.

> I'm done for today - just kicked of a desperate last attempt to get this thing to build again, but I'll be hanging out with my family again before these builds complete

Sounds like the right thing to do on thanksgiving weekend!

packaging/macosx/make-package.sh
glance-: Just -j is almost never right, because it will just overload.

Just don't provide a -j<anything> to make and it will pick it up from MAKEFLAGS instead, if that's already set in this case, or look at referenced diff above on how to compute a good value to pass to -j.

.github/workflows/mac-qt6.yml
dirkhh: ok, I gave this a go, let's see how that works out 😂

.github/workflows/mac-qt6.yml
dirkhh: turns out that this only works once that change has been merged into master... I'll see if I can hack around that "security" mechanism

dirkhh: yeah, great security people, so that now works.

.github/workflows/mac-qt6.yml
dirkhh: seems to work. of course, I haven't tested any of the resulting binaries. pesky details...

dirkhh:
I think this should be good for another review and then hopefully could be merged...

.github/workflows/mac.yml
```diff
@@ -53,10 +85,50 @@ jobs:
...
+        export PATH=$QT_ROOT/bin:$PATH
+        export CMAKE_PREFIX_PATH=$QT_ROOT/lib/cmake
+        export ACLOCAL_FLAGS="-I $(brew --prefix gettext)/share/aclocal"
+        # ensure that only our packages are used, not anything from Homebrew
+        export PKG_CONFIG_LIBDIR=$(pwd)/install-root/lib/pkgconfig
+
+        # now setup Subsurface and build the dependencies, using the generic build script
+        bash -e -x ./subsurface/scripts/build.sh -build-deps-only -ftdi ${BUILD_EXTRA_ARGS}
+
+    - name: force cache save
```
mikeller: **question:** What is this actually doing? I suspect that with this, we run the risk of storing artifacts even if the build of said artifacts has failed, and then trying to use non-working artifacts in subsequent builds.

mikeller (commented):
Looks all good, works fine on MacOS 12.7 - even bluetooth seems to be working now, which is not the case with the Qt5 based builds.

One comment about a forced cache write that I don't understand the reasoning behind.

And I've updated `make-package.sh` to remove the `-j` - `MAKEOPTS` work just fine to give us the parallelism that we want for each platform.

.github/workflows/mac.yml
dirkhh: The cashing stuff is still... Confusing to me

Let's remove that stanza

.github/workflows/mac.yml
```diff
@@ -11,27 +11,59 @@ on:
...
       uses: actions/checkout@v6
       with:
         fetch-depth: 0
         submodules: recursive
 
+    - name: Cache Homebrew packages
+      uses: actions/cache@v4
+      with:
+        path: ~/Library/Caches/Homebrew
+        key: brew-${{ runner.os }}-${{ hashFiles('.github/workflows/build.yml') }}
```
mikeller: **issue:** This doesn't key properly on the workflow definition, so it will retain the cache even when the list of homebrew packages in it changes.

```suggestion
        key: brew-${{ runner.os }}-${{ hashFiles('.github/workflows/mac.yml') }}
```

.github/workflows/mac.yml
```diff
@@ -53,10 +85,43 @@ jobs:
       with:
         nightly-builds-secret: ${{ secrets.NIGHTLY_BUILDS }}
 
+    - name: cache deps
+      id: cache-deps
+      uses: actions/cache@v4
+      with:
+        path: ${{ github.workspace }}/../install-root
+        key: macos-deps-${{ runner.arch }}-${{ hashFiles('scripts/build-deps.sh', 'scripts/get-dep-lib.sh') }}
```
mikeller: **suggestion:** I'd probably throw `subsurface/scripts/build.sh` and the workflow file itself into the key as well, to make sure we won't get fooled by cache hits after we make changes to these that affect how the dependencies are built.
```suggestion
        key: macos-deps-${{ runner.arch }}-${{ hashFiles('.github/workflows/mac.yml', 'subsurface/scripts/build.sh', 'scripts/build-deps.sh', 'scripts/get-dep-lib.sh') }}
```

.github/workflows/mac.yml
```diff
@@ -53,10 +85,43 @@ jobs:
...
+        restore-keys: macos-deps-${{ runner.arch }}-
```
mikeller: **question:**  Not sure if this is actually worth the effort to optimise the build time by partially reusing the cache whenever a dependency version is changed - these typically don't change that often, and with the way this is done now we are probably risking building with a polluted cache if something else in the build chain is changed and we then either skip the dependency rebuild, or rebuild on top of a polluted cache.
If we want to do this we should separate the hash for `.github/workflows/mac.yml`, `subsurface/scripts/build.sh`, and `scripts/build-deps.sh', and put it into the `restore-keys` slug. Or even separate the list of dependency versions into a dedicated file and then only glob across these.

dirkhh: I'm happy to overindex on "when in doubt, don't cache" and to add a lot more files to the key.
So far it never seems to work, even if none of the key'ed files change, so I'm not convinced the basics are right.

The reason why I want to cache them is that building the dependencies usually takes longer than building Subsurface - and that, for technical reasons, includes the building of more dependencies like googlemaps and soon our HTML rendering component, as anything that requires Qt will NOT be build in the build-deps step...

But at the end of the day we can just drop this and say "whatever, we don't care that the GitHub action takes 40 minutes to run"... I personally find this insanely frustrating when I actually have time to work on something, but for someone who doesn't care about macOS builds, I guess this is just noise and risk.

mikeller: > So far it never seems to work, even if none of the key'ed files change, so I'm not convinced the basics are right.

Can you point me to a workflow run where you think the cache should have worked but it didn't? Happy to try and figure out what's going wrong.

> The reason why I want to cache them is that building the dependencies usually takes longer than building Subsurface - and that, for technical reasons, includes the building of more dependencies like googlemaps and soon our HTML rendering component, as anything that requires Qt will NOT be build in the build-deps step...
> 
> But at the end of the day we can just drop this and say "whatever, we don't care that the GitHub action takes 40 minutes to run"... I personally find this insanely frustrating when I actually have time to work on something, but for someone who doesn't care about macOS builds, I guess this is just noise and risk.

Agreed, the caching is just convenience, but not absolutely needed.

And yes, working in CICD is like trying to service the engine while the plane is airborne...

dirkhh: I pushed a commit that touched nothing but the README. Yet every macOS build rebuilt the dependencies

mikeller: > Warning: Invalid pattern '/Users/runner/work/subsurface/subsurface/../install-root'. Relative pathing '.' and '..' is not allowed.

You're going to have to move the to-be-cached directory into the workspace, or this will not work.

.github/workflows/mac.yml
```diff
@@ -11,27 +11,59 @@ on:
...
         submodules: recursive
 
+    - name: Cache Homebrew packages
+      uses: actions/cache@v4
+      with:
+        path: ~/Library/Caches/Homebrew
+        key: brew-${{ runner.os }}-${{ hashFiles('.github/workflows/mac.yml') }}
+
     - name: install additional packages with Homebrew
       run: |
```
mikeller, line 53:
```suggestion
      if: steps.cache-homebrew.outputs.cache-hit != 'true'
      run: |
```

.github/workflows/mac.yml
```diff
@@ -11,27 +11,59 @@ on:
...
+            brewpackages: 'libjpg'
+    runs-on: ${{ matrix.runner }}
     steps:
     - name: checkout sources
       uses: actions/checkout@v6
       with:
         fetch-depth: 0
         submodules: recursive
 
+    - name: Cache Homebrew packages
```
mikeller, line 45:
```suggestion
    - name: Cache Homebrew packages
      id: cache-homebrew
```

dirkhh:
Yeah - I have no idea about the cache thing. Feel free to just add a commit that fixes it
otherwise I'll just rip it out and not worry about it. I'm not paying for the wasted cycles.

README.md
```diff
@@ -1,3 +1,5 @@
+Dummy change to see if caches get used.
```
mikeller: **todo:**  This probably should come out, but otherwise it looks good to go.

dirkhh:
This bundles a bunch of nonsense under `Subsurface.app/Contents/Resources/qml/QtQuick/Controls` which needs to be removed before the app is signed and packaged.
```
% du -sh Resources/qml/QtQuick/Controls/* | sort -h
4.0K    Resources/qml/QtQuick/Controls/plugins.qmltypes
4.0K    Resources/qml/QtQuick/Controls/qmldir
204K    Resources/qml/QtQuick/Controls/libqtquickcontrols2plugin.dylib
572K    Resources/qml/QtQuick/Controls/designer
1.4M    Resources/qml/QtQuick/Controls/impl
2.0M    Resources/qml/QtQuick/Controls/libqtquickcontrols2plugin.dylib.dSYM
2.9M    Resources/qml/QtQuick/Controls/Basic
2.9M    Resources/qml/QtQuick/Controls/Fusion
2.9M    Resources/qml/QtQuick/Controls/Imagine
2.9M    Resources/qml/QtQuick/Controls/Universal
3.0M    Resources/qml/QtQuick/Controls/Material
 11M    Resources/qml/QtQuick/Controls/macOS
 26M    Resources/qml/QtQuick/Controls/iOS
 34M    Resources/qml/QtQuick/Controls/FluentWinUI3
```
and it appears that our version of `qlitehtml` is responsible for that. Which seems... really odd an needs more investigation. We obviously shouldn't need iOS or FluentWinUI symbols...

dirkhh:
Ok, that was a LOT of pain. But I knew that going in.

@mikeller -- I think this may be ready for another review.

packaging/macosx/homebrew-local-build.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
mikeller, line 1: **question:** Should this rather go into `packaging/macosx`?

And it would probably help if this had a header with some explanation of what it does.

packaging/macosx/build-deps.sh
```diff
@@ -0,0 +1,269 @@
+#!/bin/bash
```
mikeller, line 1: **question:** This seems to be only used for macOS (at least at the moment) - should we make this explicit in the file location / name / header comments?

And are we planning on using this for other architectures as well?

gh_release_notes_bottom.md
```diff
@@ -7,10 +7,10 @@ If you would like to get automatic updates, you can use [Obtainium](https://obta
...
 
 # macOS
-The macOS DMG (named ** Subsurface-\<version\>-CICD-release.dmg) makes it even harder with a multi-step dance that requires opening the Privacy & Security settings in the System Preferences and explicitly confirming that you are willing to install this app.
+The macOS DMG (named **Subsurface-\<version\>-CICD-release.dmg**) makes it even harder with a multi-step dance that requires opening the Privacy & Security settings in the System Preferences and explicitly confirming that you are willing to install this app. Currently there are three different macOS DMGs. The default one is an x86 binary that runs on all current Macs and supports printing, but appears to have occasional issues with some Bluetooth dive computers. Alternatively there are two Qt 6 based versions (one for Apple Silicon, one for older x86_64 Macs) that overall seems to work better, but doesn't have printing support.
```
mikeller: **nitpick:** 
```suggestion
The macOS DMG (named **Subsurface-\<version\>-CICD-release.dmg**) makes it even harder with a multi-step dance that requires opening the Privacy & Security settings in the System Preferences and explicitly confirming that you are willing to install this app. Currently there are three different macOS DMGs. The default one is an x86 binary that runs on all current Macs and supports printing, but appears to have occasional issues with some Bluetooth dive computers. Alternatively there are two Qt 6 based versions (one for Apple Silicon, one for older x86_64 Macs) that overall seem to work better, but don't have printing support.
```

mikeller (approved):
Looks good, a couple of remarks about naming / comments.

packaging/macosx/homebrew-local-build.sh
dirkhh: done

packaging/macosx/build-deps.sh
dirkhh: done

gh_release_notes_bottom.md
dirkhh: done

mikeller (approved):

dirkhh:
Oh, oops - WIP commit snuck in there. Will push the right version, run the CI one more time to make sure everything compiles, then merge.

mikeller:
@dirkhh how's this for timing?

image

Is there any chance to move the macOS / Qt5 build to macOS-14 or newer?

dirkhh:
The move to macos-15 should be no problem with the changes we just merged

packaging/copr/subsurface.spec
```diff
@@ -23,7 +23,8 @@ BuildRequires:  desktop-file-utils
 BuildRequires:  fdupes
 BuildRequires:  gcc-c++
 BuildRequires:  make
-BuildRequires:  asciidoc
+BuildRequires:  asciidoctor
+BuildRequires:  rubygem-asciidoctor-pdf.noarch
```
mikeller, line 27: **question:** @dirkhh are you able to check the outcome of the builds of the nightlies for Fedora? The latest that I can install with `dnf` at the moment is 6.0.5494, which happens to be the one before this was merged.

PR #4642, opened 2025-11-28
CICD: Make the Build Number Generation more Robust.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Create a new build number only if the workflow is triggered by a push
AND is running on the `master` branch.
This removes the dependency on the workflow trigger condition being
correct.


dirkhh (approved):
That seems reasonable

PR #4643, opened 2025-12-02
Qlitehtml for the user manual

dirkhh:

- [x] Functional change

allow building against qlitehtml and use that for the user manual

### Changes made:

- build qlitehtml as a dependency on macOS (no other OS has been touched for this)
- switch to asciidoctor to avoid the need for JavaScript for the table of contents

This isn't really working - the PR is here to get started on testing builds...


dirkhh:
oh, and since I am doing this on my Mac (as that's where I did the Qt 6 work which got me down this path int he first place), this PR is of course based on #4640

atdotde:
This is really nice, I particularly like the font rendering which is very clear.

One oddity I found is that when you search using Command-F, enter a search term and press return, it does an upwards search rather than a downwards search as I would have expected.

.github/workflows/documentation-build.yml
```diff
@@ -42,10 +42,10 @@ jobs:
...
         fi
-        DEBIAN_FRONTEND=noninteractive ${SUDO} apt-get -y -q install make asciidoc docbook-xml w3m
+        DEBIAN_FRONTEND=noninteractive ${SUDO} apt-get -y -q install make asciidoctor docbook-xml w3m
 
     - name: build the documentation
       if: steps.cache-documentation.outputs.cache-hit != 'true'
       run: |
         git config --global --add safe.directory $GITHUB_WORKSPACE
-        make -C Documentation
+        make -C Documentation doc
```
mikeller, line 51: **suggestion:** The point of this workflow is to test the documentation build (all of it) - so I think the intended target should be `all` (or the default target).

ReleaseNotes/Makefile
mikeller, line 22: **note:** For all that I know this `Makefile` isn't used in any of our processes any more - all we are consuming are the `.txt` files themselves. So I think we may as well remove it, lest more work is spent updating something that is no longer used.

And if not, we should fix this: 😉


```suggestion
all: $(HTMLDOCS) $(TEXTDOCS) $(PDFDOCS)
```

.github/workflows/documentation-build.yml
dirkhh: I thought this caused a build failure... let's see what happens if I change it back

ReleaseNotes/Makefile
dirkhh: yeah, this was for me back when I published the release notes as html on the old website.

.github/workflows/documentation-build.yml
dirkhh: well, what do you know...
https://github.com/subsurface/subsurface/actions/runs/19970197525/job/57272710805#step:7:47
🤷

dirkhh:
now that this seems functional, I merged it into the overall macos-qt6 thingy as that's what it depends on, anyway

.github/workflows/documentation-build.yml
```diff
@@ -42,10 +42,10 @@ jobs:
         if type sudo; then
           SUDO=sudo
         fi
-        DEBIAN_FRONTEND=noninteractive ${SUDO} apt-get -y -q install make asciidoc docbook-xml w3m
```
mikeller, line 45: **issue:**  This will break rules in the `Makefile`, as `a2x` is part of `asciidoc-base`. See https://github.com/subsurface/subsurface/pull/4643#discussion_r2593375818

.github/workflows/documentation-build.yml
mikeller: So the proposed changes break some of the builds in the Makefile (possibly with a simple fix, see my other comment). Should we:
- try to fix them
- decide we don't care about their output, and just remove them

Personally I think that generating the manual in PDF form has got some value.

.github/workflows/documentation-build.yml
dirkhh: ah, gotcha. well, we need asciidoctor because that is capable of rendering html that we can display with the new HTML rendering module qlitehtml - not sure if there's an equivalent to a2x in that stack - I'll investigate.

mikeller: See https://github.com/subsurface/subsurface/pull/4640/commits/fd91b5c4859aa481d73aa8ca2a00311b9cb17aba - seems to be working, downloading the artifacts right now to see how the PDF looks.

And no issues at all with moving from asciidoc to asciidoctor - it seems to be better documented and less temperamental. But we also need the `asciidoc` package to get `a2x` for the PDF build.

dirkhh: Going down a slightly different path. 
Give me a few hours

dirkhh: Going down a slightly different path. 
Give me a few hours

PR #4647, opened 2025-12-08
Qt6 warning cleanup

dirkhh:

- [x] Code cleanup

This should be fairly mechanical, simply aiming to get rid of many of the warnings that are cluttering the build (and avoid future problems), but of course I always worry that a change like this introduces a subtle change that I don't see and that comes back to bite us in the future.

### Changes made:
See commits, simply did a Qt 6 build and tried to address the warnings (and do Qt 5 builds to make sure I didn't break things) until I ran out of patience...

### Mentions:

Code reviews would be nice, but of course this just as much requires actual testing of both Qt5 and Qt6 builds...


mikeller (approved):
Looks good to me. All of the `QT_VERSION_CHECK()` instances ain't pretty, but then it will be easy to find and remove them once we are done with Qt5.

PR #4649, opened 2025-12-10
Implement gas blender - Mobile

fnightfox:

- [ ] Bug fix
- [ ] Functional change
- [X] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This pull request implements a gas blender and ties this functionality into the mobile app, technical diving->gas calculator

The algorithm I implemented is not one that I found online or read anywhere. I am looking to write it up in a formal way and (ideally) publish it somewhere for further review, but so far it seems accurate and complete. The basics involve the following:

1. Convert EVERYTHING into volumes. We should deal with pressures as little as possible, and when we do try to include the real gas law vs the ideal gas law. To this end, you will see a copy of the tables used elsewhere in subsurface in the core/gasblender cpp file. Liner approximation is used for values between those presented, and should be accurate enough for this application.

2. Take all gas sources and split them into two types: unlimited (called "GasSources") and limited gasses
3. The limited gas sources are used to create "Blend" objects. These objects are order dependent, and need to have every combination possible, including draining the target cylinder to ATM. This causes a factorial explosion such that only 3 limited volume sources cause ~16 distinct blends. 
4. These are grouped by order used in the steps section of the struct, and then every combination of 3 blends is used to create a triangle where the x,y coordinate is the neededO2, neededHe for the REMAINING gas to fill the target cylinder to full with our desired mix.

5. All of the GasSource objects are used to form as many distinct triangles as possible. These represent the POSSIBLE blends of unlimited gas we have access to.

6. We check for any overlap between the two sets of triangles. This is the computationally expensive step.

7. Each overlapping "point of interest" (where the triangles intersect, any point of one triangle contained in another, etc..)  is filled out as a blend. These are saved.

8. All of those blends are checked for cost. The lowest cost blend is returned.

I did a decent amount of checking and validation to ensure this algorithm is sound, but if there were any questions please raise them in the comments.

**Some items of note:**
1. You must have at least one unlimited source
2. Simple Blending makes all sources unlimited
3. Source order for unlimited sources leans towards the order they are on your source list
4. PRESSURE calculations are only done when needed, and the target pressure is only used at the beginning. The pressure calculated at the end is not checked against the target and should be used to assure yourself of the accuracy of the calculations. I noted a ~9PSI difference when targeting 3000PSI. This is acceptable and represents a .3% calculation error throughout the calculations. 
5. The final mix should be within .1% of the target.
6. I am interested in blending by weight, as it does not depend on the temperature of the gas or accuracy of the pressure gauge, so that is why I put that value first (and it is more accurate than pressure is anyway).

**PLEASE DO NOT MERGE UNTIL THIS HAS BEEN TESTED, DEBUG CODE REMOVED, AND (IDEALLY) SOME DOCUMENTATION HAS BEEN CREATED**

Items I know need to be done:
1. Convert spaces to tabs
2. Remove the commented out code using the convexHull. This method greatly reduces the number of triangles, but does not work as a minimum cost method. I have left it in because I do not want to delete it until others comment.

### Changes made:
1. Add GasBlender cpp core code
2. Add unit tests for the gas blender cpp code
3. Add Gas Calculator mobile front end codeub that this PR fixes an issue. -->

### Additional information:
See long description for notes and algorithm

### Documentation change:
None, yet

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -27,6 +27,7 @@
 #include <QTextDocument>
 #include <QtConcurrent>
 #include <QVariantMap>
+#include "core/gasblending.h"
 
 #define VARIATIONS_IN_BACKGROUND 1
 
@@ -1443,6 +1444,7 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 	d->dcs[dcNr].divemode = static_cast<enum divemode_t>(diveMode);
 
 	// Populate cylinders from QML data
+	std::vector<int> cylinder_start_pressures;
 	for (const QVariant &cylData : cylindersData) {
 		QVariantMap map = cylData.toMap();
 		cylinder_t newCyl;
@@ -1464,7 +1466,7 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 		d->cylinders.add(d->cylinders.size(), newCyl);
 	}
 	this->cylinders.updateDive(d, dcNr);
-	reset_cylinders(d, true);
+	//reset_cylinders(d, true);
 
 	// Add available OC-gases as "time=0" waypoints for the planner engine
 	pressure_t deco_po2_limit = { .mbar = qPrefDivePlanner::decopo2() };
@@ -1474,6 +1476,7 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 			depth_t mod = d->gas_mod(cyl.gasmix, deco_po2_limit, 1_m);
 			divedatapoint point(0, mod, i, 0, false); // time=0, depth=MOD, cylinderid=i
 			diveplan.dp.push_back(point);
+			//log cylinder pressure
 		}
 	}
 	diveplan.salinity = waterType;
@@ -1549,21 +1552,15 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 	return results;
 }
 
-QVariantList DivePlannerPointsModel::calculateGasInfo(const QString &cylinderType, int o2_permille, int he_permille)
+QString DivePlannerPointsModel::calculateGasInfo(const QString &cylinderType, int o2_permille, int he_permille)
 {
-	// Create a temporary dive context for calculations
 	struct dive temp_dive;
 	make_planner_dc(&temp_dive.dcs[0]);
-
-	// Create a temporary cylinder with the specified gas mix
 	cylinder_t temp_cyl;
 
-	// Populate cylinder type info
 	std::pair<volume_t, pressure_t> type_info = get_tank_info_data(tank_info_table, cylinderType.toStdString());
-	volume_t size = type_info.first;
-	pressure_t pressure = type_info.second;
-	temp_cyl.type.size = size;
-	temp_cyl.type.workingpressure = pressure;
+	temp_cyl.type.size = type_info.first;
+	temp_cyl.type.workingpressure = type_info.second;
 	temp_cyl.type.description = cylinderType.toStdString();
 
 	// Populate gas mix
@@ -1571,19 +1568,13 @@ QVariantList DivePlannerPointsModel::calculateGasInfo(const QString &cylinderTyp
 	temp_cyl.gasmix.he.permille = he_permille;
 	sanitize_gasmix(temp_cyl.gasmix);
 
-	// Get narcotic preference
 	bool o2_is_narcotic = qPrefDivePlanner::o2narcotic();
-	int fNarcotic_permille;
-	if (o2_is_narcotic) {
-		// If O2 is narcotic, the narcotic fraction is N2 + O2.
-		fNarcotic_permille = get_n2(temp_cyl.gasmix) + o2_permille;
-	} else {
-		// Otherwise, it's just N2.
-		fNarcotic_permille = get_n2(temp_cyl.gasmix);
-	}
+	int fNarcotic_permille = o2_is_narcotic ? (get_n2(temp_cyl.gasmix) + o2_permille) : get_n2(temp_cyl.gasmix);
 
-	QVariantList results;
-	// Calculate for a standard range of pO₂ values
+	QString results = "<table width='100%'>";
+	QString ead_header = o2_is_narcotic ? tr("END @ MOD") : tr("EAD @ MOD");
+	results += QString("<tr><th align='center' style='padding-bottom: 5px;'><b>pO₂</b></th><th align='center' style='padding-bottom: 5px;'><b>MOD</b></th><th align='center' style='padding-bottom: 5px;'><b>%1</b></th></tr>").arg(ead_header);
+
 	double po2_values[] = { 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2.0 };
 
 	for (double po2 : po2_values) {
@@ -1591,8 +1582,6 @@ QVariantList DivePlannerPointsModel::calculateGasInfo(const QString &cylinderTyp
 
 		// Calculate MOD
 		depth_t mod = temp_dive.gas_mod(temp_cyl.gasmix, po2_limit, 1_m);
-
-		// Calculate EAD/END at the MOD
 		double p_amb_at_mod = temp_dive.depth_to_atm(mod);
 
 		double p_narcotic = p_amb_at_mod * fNarcotic_permille / 1000.0;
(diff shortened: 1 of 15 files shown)


mikeller:
Wow, this is certainly a **very** comprehensive blender! Good work.

I think @atdotde is best positioned to comment on the mathematical approach taken.

From a user perspective this is already quite good to use, just a couple of rough edges:
- use a better default than 3000 if the pressure is set to bar (or persist the user's preferences for these settings);
- since the ordering of gases in the list dictates the order of filling for the simple blender, allow users to reorder rows without having to delete / re-create.

Some ideas for possible improvements:
- in the full blender, allow a limited gas to be boosted - still only a limited volume available, but not limited by blending only to equilibrium;
- in the full blender, allow the user to create 'blend plans' - when a desired mix is accepted by the user, store it for later retrieval, and update the gas list with the remaining gas after the blend was done, so they can plan the next blend with what is left in the supply tanks.

atdotde:
So far I only had a brief look, but this looks far too complicated to me. And you are reinventing the whee in far too many places. Subsurface already has all required helper functions to deal with gas calculations we we took great care to get those right (in particular for dealing with real gases). Please use those! 

Also I think the approach to express everything in volumes is wrong. The calculation does not depend on intensive variables like pressure, then also the formulas get simpler and only in the very end, you multiply by the actual volume of the cylinder.

I also don't see where you need any iterations over filling steps. The amounts of gas are determined by a system of linear equations:

You start from a given mixture at a given pressure and have a target mixture and pressure. You can convert both gases to say 1bar to find the amount (if you like in linters) of each of the partial gases that are required to be added. Since you need to add a specific number of litres of O2, N2 and He, you need three linearly independent sources of gas to add (if you are blending nitrox, everything reduces to 2). But those sources in general are not pure but are certain pictures. Finding the amount of each source is done by solving a system of three linear equations and three unknown, You can use closed formulas for that, even implementing a matrix inverse is overkill. Once you know how much of each source you need at 1bar you can (using again the compressibility factor) convert that to the different intermediate pressures. And at that point you could check that your limited sources actually contain enough gas to allow for that plan.

All this  is implemented in lines 123--214 of https://github.com/atdotde/realblender/blob/main/blender.pl , It doesn't have to be more complicated than that.

Or am I missing something?

fnightfox:
The complications mostly come from the need to limit the amount of gas. In the solution you stated the viability of a mixture is only checked at the end, and even then there is no way to validate if it is the cheapest possible blend.

I will look through the core code again for helper functions to see what's been implemented and what I can pull into this, but I do believe the method you detailed is not comprehensive enough for the use case I was imagining, where a limited gas can be used up to the limit and then the blend is continued from there, finding the cheapest option.

atdotde:
So, can you describe the problem you are actually trying to solve here? You have more than three sources of gas (as in He, O2 and air), say some additional trimix of some sort, so there is more than one "solution" to the filling problem. And then you assign a cost per volume to all these gases and try to find the cheapest option? And how do you decide what is the limited amount, do you assume you have a booster pump so you can bring any gas to any pressure is the limit achieved when you equilibrate two cylinders?

And how realistic is this scenario (as in: is it worth considering the finiteness of the available gases or do people pretty much always more gas in their storage cylinders than what is needed for this fill)?

Can you give me a concrete, real world example with numbers that you would like the app to provide the solution for?

I think even with more than three gas sources, the problem has an analytic solution: You have linear equations and linear inequalities and a linear price function. Then you know the minimum is always in one of the corners of the allowed solution space.

fnightfox:
> So, can you describe the problem you are actually trying to solve here? You have more than three sources of gas (as in He, O2 and air), say some additional trimix of some sort, so there is more than one "solution" to the filling problem. And then you assign a cost per volume to all these gases and try to find the cheapest option? And how do you decide what is the limited amount, do you assume you have a booster pump so you can bring any gas to any pressure is the limit achieved when you equilibrate two cylinders?
>

I am assuming I have at least an air compressor (unlimited, 21/0) and boosted O2 and He(in my case I treated these as unlimited, but it is a fair point to add BOOSTED as an option for limited blends, but this is besides the point). This means that I know that a blend is possible for any realistic blend (there is a triangle from pure nitrogen to pure helium to air that is not possible, but that is not a blend I have ever heard of anyone desiring). I also am thinking about a case where I have a tank of trimix with some gas left over, or a rich O2 mix that has leftover gas in it. Reusing these gases would be cheaper than filling the tank using the pure He/O2 mixes because I have already paid for these mixes. If you would like numbers, assume the following scenario:
 1. Unlimited Air
 2. (Functionally) unlimited O2 and He, from boosted bank bottles, with costs of $.15/cuft for O2 and $1.75/cuft Helium
 3. An HP100 at 1500 PSI of 18/45
 4. Another HP100 at 1200 PSI of 30/30

I am targeting an AL80 currently empty to 3000PSI and 21/18.
The simple solution is to blend from the three boosted bottles, giving a cost of 24.74 and removing the ability to reuse anything from the other bottles. Using the blender proposed here, you can use 14.55cuft of 30/30 (dropping that source to ~700PSI), 10.54cuft of 18/45 (dropping that source to ~1100 PSI), and only have a cost of 8.25.

> And how realistic is this scenario (as in: is it worth considering the finiteness of the available gases or do people pretty much always more gas in their storage cylinders than what is needed for this fill)?
>

I believe that if you are always using bank bottles, you probably have enough. The use of this is to allow the reuse of already blended tanks to reduce wasted gas.
 
> Can you give me a concrete, real world example with numbers that you would like the app to provide the solution for?
> 

See above for a scenario with numbers. I came up with this one just off the top of my head, but it seems like a reasonable thing to want to reuse as much gas as you can. 

> I think even with more than three gas sources, the problem has an analytic solution: You have linear equations and linear inequalities and a linear price function. Then you know the minimum is always in one of the corners of the allowed solution space.

I agree with this. The triangle solution is used to construct that solution space, with the corners being the blends calculated and compared for cost. The programming to have a generalized solver for the linear equations seemed more complicated than finding the barycentric weighting of the three points of the triangle, so I elected to implement a barycentric weighting approach to solving the problem, since the result for this application appeared to be the same.

atdotde:
OK, fair enough. But how often do you really do that rather than just topping up a cylinder that still contains some gas? And is it really fair to consider the gas in those extra two cylinders to be free of charge? At some point you will have to fill those again and then you will have to spend more on topping up those. Since the price is dominated by the price for pure He, you will have to spend exactly the same amount of money more that you save this time.

In any case, if you consider that spare gas free, the cheapest option will always be to first use that gas completely  (with the only exception that that gas already contains too much He). But then you can first do that and then analyse what you got (you will have to do that anyways to make sure not only the pressure equilibrated but also the composition) and use that as your starting point to top up from the infinite reservoirs.

fnightfox:
I would say that the costs of that gas can be variable. In this case I picked free, but it is possible that it's valued just at a different rate than pure He. The case here is what proportion of each of these gases do you use to maximize the reuse while not going above the max O2 or He for your top up gas (or the other unlimited sources). This is exactly how my algorithm works, by examining the limited gas space (with reference to what will be needed after you add it) and seeing where it overlaps with the unlimited sources available. 

I believe, in order to accomplish the case I presented, it is required to essentially examine all possible combinations of gases presented, first by looking at limited sources then unlimited. I believe the only functional difference between what you presented and what my algorithm does is that I use triangles and the weighting of each point to solve those equations vs implementing an equation solver, as I found this method easier to visualize and implement.

dirkhh:
I would say that this is a cool niche application for the nine people on earth who want to bother with it to save 6 bucks. And I think that this complexity (especially the UI side of it) is not useful or appreciated (at least from my perspective) for Subsurface as that is far, far, FAR outside the target audience.

fnightfox:
> I would say that this is a cool niche application for the nine people on earth who want to bother with it to save 6 bucks. And I think that this complexity (especially the UI side of it) is not useful or appreciated (at least from my perspective) for Subsurface as that is far, far, FAR outside the target audience.

As one of "the nine people on earth" who does actually want the features I described, I don't actually think it adds much complexity from the user side. Which precise piece of the UI do you feel can be improved to reduce complexity?

dirkhh:
a gas blending feature that allows (based on your description) unlimited number of input gases with additional input values (the available pressure, the availability of a booster, and, I assume, given your desire to match a realistic use case, the performance characteristics of said booster) is a very specific, very narrow, and by necessity very complex UI.

I encourage you to build an Android or iOS app with that exact feature set.

I do not agree that this is a feature that should be added to Subsurface.

fnightfox:
> a gas blending feature that allows (based on your description) unlimited number of input gases with additional input values (the available pressure, the availability of a booster, and, I assume, given your desire to match a realistic use case, the performance characteristics of said booster) is a very specific, very narrow, and by necessity very complex UI.
> 
> I encourage you to build an Android or iOS app with that exact feature set.
> 
> I do not agree that this is a feature that should be added to Subsurface.

This is why there is a difference between simple and full blending screens. I do agree that the use case is fairly niche for people blending gasses, but the simple blending screen won't have anything more than it already does, and that handles almost all use cases. It is only the rare case where someone would need the added complexity that they would switch over to the other screen and appreciate that it exists. 

The only reason I proposed adding it here is that there is already a gas calculator screen that only calculated MOD and END, so there was a spot for it and it has been requested by multiple people (one to me directly, and at least one by others on here).

mikeller:
> I would say that this is a cool niche application for the nine people on earth who want to bother with it to save 6 bucks. And I think that this complexity (especially the UI side of it) is not useful or appreciated (at least from my perspective) for Subsurface as that is far, far, FAR outside the target audience.

Full disclosure: I am another one of the nine. 😉

But joke aside, I think most technical divers doing their own trimix blending fall into the 'limited gases' category. And in many parts of the world this is the only option to do trimix diving, as there simply aren't any dive shops that are set up and willing to sell trimix fills.

And I think having the option to select between the 'simple blender' UI which will be good for most Nitrox applications, and the 'full blender', which will be preferable for most Trimix blending is a good way to solve the UI complexity. Possibly it could be turned into a preference, assuming that most users will want to use either one or the other.

fnightfox:
I have made some changes, including adding boosting to the full blender (It uses hardcoded limits right now, to a bit over 100PSI), and making the available cylinders drag and drop.

I still want to go through and ensure I use the existing gas management code. I also want to write a whitepaper detailing the methodology for review and documentation on the UI for users. 

Sorry this took so long, things are swamped close to the holidays for me.

fnightfox:
Sorry for the delay, had some medical things to take care of (and a dive trip!).

I found a bug after the drag and drop was added, so this fixes that and adds a top up calculator if you only include a single gas available. I'm still planning on getting a write up on the blending math done and on doing a check for reusing other gas functions in subsurface to keep that logic consolidated, so for now keep this flagged as don't merge.

RonOnGithub:
Hi!

Just a heads up that I'm right in the middle of porting the mobile UI to Qt6
and there's some unavoidable significant changes to the QML and how we
get C++ data to and from it which will come with that.  So you might want to
not sink too much time into polishing that side of it just yet.

But adding a gas blender was on my list of things to do after that too, so
this is really cool to see, and if all-y'all can come to consensus on the
algorithm to use and options to present, then I'll be happy to fix up all
the bits of the mobile side that I'm on the pointy end of busting in the
meantime.

fnightfox:
Sure thing, I've found a bug or two with my implementation so far, but I'm going to also be working on a quick algorithm white paper for why I think this method works well (and better than partial pressure blending since this is temperature independent, blending based on volume and weight, then giving pressures as a last step).

I'm hoping to have most of that done over the next month, but I've had some personal medical challenges come up that have made it impossible for me to spend much time on this at the moment.

fnightfox:
Attached to this commend is a quick write up I did on this methodology. I am still tracking down a bug in handling the hyperbolic problem of limited gas sources (this needs to be handled in two places, I have solved one but need to figure out the other. It really only comes up when limited sources that are very large are used, ones that could be handled as unlimited in the current implementation and give workable results, but I want to get it nailed down completely before merging, if this is the route we decide to go down).

I am posting this as a way to get discussion on this method vs others. I am happy to recode the entire thing if another method makes more sense, but I don't want to just because this is a new way of approaching the problem, as I feel this solves some of the tricky aspects of blending (like limited sources, ordering the sources, and finding minimum cost).

Still TODO:
1. fix UI issues from the move to QT6
2. Check existing gas functions for possibility of reuse
 - dealing with absolute vs gauge pressure
 - dealing with gas compressibility
3. Track down method of handling the hyperbolic issue for trimix.  
 - UPDATE: I think I solved this by fixing the volumes of limited gasses to clamp, in order they are added, to max at the target value for the cylinder being filled. Not pushing this yet, until I get the UI fixes in.

[blending-paper-1.pdf](https://github.com/user-attachments/files/26576244/blending-paper-1.pdf)

fnightfox:
Ok, major changes in the most recent version. I reworked the entire system to use linear programming with a simplex matrix and solver., which is more traditional for solving this problem So please ignore the document I sent earlier, as it no longer represents the code in this PR. I have updated the blender to use the compression factor/real gas constants generated by the rest of the app, but I do not think there is a real opportunity for reuse for the sections dealing with the absolute pressure/gauge pressure conversions, since that existing code is mainly used with dive planning or dive objects and one does not exist in the blending context.

I have added a set of unit tests as well, and updated the UI for QT6. 

At this point, I am content with this merging if people want/like this implementation, and if the UI is accepted I can work on the documentation (at lease as much as I can do, I have never worked on the documentation in this project before).

Thank you all,
--Liz

fnightfox:
@mikeller @RonOnGithub I know you were both somewhat interested in this feature. I think it is in a mostly stable state, but I would like some additional people to give input (both on functionality and, ultimately, if this fits into subsurface). I know this PR was open for a bit, but if this is something desired, I think it's in a good beta state.

mikeller:
I've been having a busy week - will be looking into this on the weekend.

qt-models/diveplannermodel.cpp
```diff
@@ -1462,7 +1464,7 @@ QVariantMap DivePlannerPointsModel::calculatePlan(const QVariantList &cylindersD
 		d->cylinders.add(d->cylinders.size(), newCyl);
 	}
 	this->cylinders.updateDive(d, dcNr);
-	reset_cylinders(d, true);
+	//reset_cylinders(d, true);
```
mikeller: **question:** Should this be permanently removed?

qt-models/diveplannermodel.cpp
```diff
@@ -1600,14 +1589,240 @@ QVariantList DivePlannerPointsModel::calculateGasInfo(const QString &cylinderTyp
...
+				output += "Remove " + weightString + volumeString + " to " + pressureString + "<br>";
+			}
+			
+		}
+		if (cheapest.isBestEffort) {
+			//Loop again, using the actual found amounts as the target mix. This should help get the right volumes for pressure to equalize
+			target.targetO2_permille = cheapest.finalO2;
+			target.targetHe_permille = cheapest.finalHe;
+			target.fillVolume.mliter = 0.0;
+            Blender::calculate_cylinder_volumes(target);
```
mikeller: **nitpick:** Mix of tabs and spaces.

qt-models/diveplannermodel.cpp
```diff
@@ -1600,14 +1589,240 @@ QVariantList DivePlannerPointsModel::calculateGasInfo(const QString &cylinderTyp
...
+		}
+	}
+	//Add the final mix, as calculated
+	if (cheapest.isBestEffort || loops > 1) {
+		output += "<br><b><span style='color: red;'>THIS MIX FAILED TO CONVERGE, IT IS THE CLOSEST MIX FOUND TO THE TARGET</span></b><br>";
+	}
+	output += "<br>Final Mix: " + QString::number(cheapest.finalO2 / 10.0, 'f', 2) + "/" + QString::number(cheapest.finalHe / 10.0, 'f', 2) + "<br>";
+	output += tr("<br>Total Cost: %1").arg(QString::number(real_cost, 'f', 2)) + "<br><br>";
+
+    return output;
```
mikeller: **thought:** I am wondering if `diveplannermodel` is the right place for the code relating to the blender, or if it would make more sense to add a new model for this?

mikeller (commented):
This looks by and large ok to me - I haven't tested it yet.

@atdotde, @bstoeger can you please have a look?

qt-models/diveplannermodel.cpp
fnightfox: It would probably make sense to pull this into a new model. I will also correct tabs and spaces (my IDE uses spaces, so sometimes I forget to convert), and I will remove the commented line. I will try to have these changes in place this week.

mikeller:
@fnightfox  are you still interested in this?

fnightfox:
> @fnightfox are you still interested in this?

I definitely am, I was hoping to get a chance to actually blend a mix using it at my dive shop as a quick test before getting it finalized, but ended up not being able to. I'll get back to work on it to close it out.

mikeller:
I gave this a quick spin, and the 'simple mode' seems to be pretty much hitting the spot for home blending. The 'advanced mode' looks like something for people doing commercial blending, or blending large volumes.

fnightfox:
> I gave this a quick spin, and the 'simple mode' seems to be pretty much hitting the spot for home blending. The 'advanced mode' looks like something for people doing commercial blending, or blending large volumes.

Advanced (or full )is also made to allow someone to use tanks that they already have filled (like of you still have a half filled tank from a previous dive), or if you don't have access to large quantities of source gas. I made it with something like "a trans fill whip in the back of a car" in mind.

fnightfox:
I just noticed weird behavior in other parts of the app. I want a chance to test if it's my fault/this PR or unrelated. The SAC rate on logged dives is off now.

fnightfox:
@mikeller I have checked and the bug IS present in 5660 nightly release. It appears that the SAC on mobile that is displayed is either only for one tank or is 1/[# of tanks]. I dive sidemount and my SAC listed is almost exactly 1/2 of what I expect it to be. This appears to only happen when there are more than one tank present. I will look into it tomorrow if I have time, otherwise I will submit a bug report. I have no issues with this PR continuing to be evaluated/merged or declined.

--Liz

fnightfox:
ok. I have tracked down the issue I was seeing, and it probably comes down to a setting on my perdix. My second tank was being imported as "not used" vs OC Gas, and the recent merge made it such that the tank was not being counted. As a result, this would probably be fixed with a dropdown on the edit dive screen for "use" for the cylinder, but it is not in fact a bug.

fnightfox:
Hey, @mikeller , just wanted to see if this is a feature you'd like to have me keep open or changes still to be made?

mikeller:
I definitely still think that a gas blender will be a valuable addition. @atdotde can you please have another look at this?

PR #4659, opened 2025-12-17
fix copr build

dirkhh:
Well, maybe - we won't know until it is merged since we don't do copr builds on pull requests.

This also updates the OBS spec file - but when looking into this I noticed that we have stopped triggering OBS builds more than 2 years ago and apparently no one ever noticed or cared. Oops. Maybe it's time to stop pretending we build for opensuse?

- [x] Build system change

Not sure where I found that incorrect package name. Embarrassing...

### Mentions:

@mikeller -- the automation of builds for this PR won't help, it doesn't test the COPR build. Maybe we need to create a 'simulated' COPR build on GitHub? Not sure. Definitely not something that I'll try on this crawling internet connection.

Which means for now, short of merging this, there's no way to know if this really fixes things.

### DIFF
--- a/packaging/OBS/subsurfacedaily.spec
+++ b/packaging/OBS/subsurfacedaily.spec
@@ -23,7 +23,7 @@ BuildRequires:	fdupes
 BuildRequires:	gcc-c++
 BuildRequires:	make
 BuildRequires:	asciidoctor
-BuildRequires:  rubygem-asciidoctor-pdf.noarch
+BuildRequires:  rubygem-asciidoctor.noarch
 BuildRequires:	autoconf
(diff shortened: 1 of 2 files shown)


mikeller (approved):

mikeller:
Looks good to me.

Not sure that setting up a simulated 'COPR' build would give us much, unless there is an official mechanism that includes updates whenever the actual process is changed.
It's probably easier to just test with the real thing. And as with any CICD changes, if you're not sure if something is going to work you can always opt to push the exploratory changes straight into `master` to cut down on the process overhead / delay / noise.

PR #4668, opened 2025-12-26
Add support for the Shearwater Swift GPS

### LINKED ISSUE #97: Add support for Swift GPS
Hello!

libdivecomputer just added support for GPS data from Swift GPS transmitters.  Here’s the commit:  https://github.com/libdivecomputer/libdivecomputer/commit/4f5abbddc66843c83fabc05959301ec956fd1b77

I’m sure you will gather up these changes at some point.  I was just hoping to create some awareness and see if that point might be sooner!

Thank you very much from a happy Subsurface user who has a new Swift GPS… :)

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Pro forma pull request to generate test artifacts for the changes in the
libdivecomputer submodule.

### Related issues:

Implements libdivecomputer/libdivecomputer#53.
Fixes subsurface/libdc#97.

### DIFF
--- a/core/libdivecomputer.cpp
+++ b/core/libdivecomputer.cpp
@@ -608,7 +608,7 @@ uint32_t calculate_string_hash(const char *str)
 	return calculate_diveid((const unsigned char *)str, strlen(str));
 }
 
-static void parse_string_field(device_data_t *devdata, struct dive *dive, dc_field_string_t *str)
+static void parse_string_field(device_data_t *devdata, struct dive *dive, dc_field_string_t *str, bool got_location)
 {
 	// Our dive ID is the string hash of the "Dive ID" string
 	if (!strcmp(str->desc, "Dive ID")) {
@@ -621,8 +621,8 @@ static void parse_string_field(device_data_t *devdata, struct dive *dive, dc_fie
 	add_extra_data(&dive->dcs[0], str->desc, str->value);
 
 	/* GPS data? */
-	if (!strncmp(str->desc, "GPS", 3)) {
-		char *line = (char *) str->value;
+	if (!got_location && !strncmp(str->desc, "GPS", 3)) {
+		char *line = (char *)str->value;
 		location_t location;
 
 		/* Do we already have a divesite? */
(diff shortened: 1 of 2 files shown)


dirkhh (approved):
why a sane person would want this to be a float and not an integer is beyond me, but whatever. Appoved

chaehni:
Thanks for opening the PR. I installed the Subsurface version created from this commit and imported a dive from my Perdix 2. However, i cannot see the coordinates anywhere.  They are not in the location textbox nor under "Extra Info." This is my first computer with GPS capabilities, so it might be that I just don't know where to look.

mikeller:
> Thanks for opening the PR. I installed the Subsurface version created from this commit and imported a dive from my Perdix 2. However, i cannot see the coordinates anywhere. They are not in the location textbox nor under "Extra Info." This is my first computer with GPS capabilities, so it might be that I just don't know where to look.

Thanks for the feedback - as it turns out Subsurface uses a format for location reporting that predates location support in libdivecomputer. This will need some changes in Subsurface to make location support work for dive computers other than Garmin.

dirkhh:
I would rather fix the libdc code and match the way we get gps days from the Garmin
Random floats are not a reasonable way to do this. 

This has been an ongoing debate and fundamental disagreement for a decade with many other examples.

mikeller:
> why a sane person would want this to be a float and not an integer is beyond me, but whatever. Appoved

If I got a dollar each time I saw floats used for lat / lon in 'commercial grade' software I'd be filthy rich by now...

dirkhh:
If I got a cent (US$ or NZ$, doesn't matter), for every time I saw a stupid choice of data type and data transfer protocol or data encapsulation protocol, I could buy myself quite a few nice toys...

mikeller:
> I would rather fix the libdc code and match the way we get gps days from the Garmin Random floats are not a reasonable way to do this.
> 
> This has been an ongoing debate and fundamental disagreement for a decade with many other examples.

Piggybacking on an 'extra info' filed with a specific name seems to be somewhat of a workaround - to me using a dedicated field for location seems to a more robust approach.

And doing our own thing here also comes with the risk that whenever libdivecomputer adds support for a dive computer, this will not push through into Subsurface unless somebody picks up on it and adds our own way - as has already happened for the DiveSystem, Divesoft, and Halcyon dive computers...

mikeller:
@dirkhh: Not sure where we want to take this from here?

I can see the following options:
1. extend existing approach in Subsurface to use an Extra Info field with a special name to convey location information, and retrofit this to all dive computers that report location information (for now);
2. leverage the the `DC_FIELD_LOCATION` approach that libdivecomputer has implemented for the dive computers supporting location (except for Garmin), and accept that using floats for lat / lon is a [flawed approach](https://xkcd.com/2170/);
3. leverage `DC_FIELD_LOCATION`, but change it to use a fixed decimal notation for lat / lon.

The way I see it, 1. is not particularly sustainable as this will get us to play catch up whenever a new dive computer has got location support. And I really don't think that 'leverage special names for Extra Info to transmit data' is an approach that should be turned into best practice and used more widely. 2. gets us to swallow the frog of float lat / lons, but at least new dive computers will work out of the box. 3. is most work, but then it will fail fast and get us to update the backend for new dive computers if they support location, in order to make them build in Subsurface.

dirkhh:
I am frustrated with the choices that Jef makes. They are inconsistent and weird - but that hasn't changed in 15 years and is unlikely to change in the future. Let's go with the lowest effort approach of (2) and just move on.

dirkhh:
> I would rather fix the libdc code and match the way we get gps days from the Garmin Random floats are not a reasonable way to do this.
> 
> This has been an ongoing debate and fundamental disagreement for a decade with many other examples.

in other words... this response is technically correct and something I still believe, but the effort isn't worth it

mikeller:
@chaehni Can you please try importing again, with the latest artifacts?

chaehni:
Hi @mikeller 
Thank you for the update. I downloaded some of the newest dives to a fresh logbook. Unfortunately, I still can't see any location information.
One small thing I noticed, the linked artifact here in Github is called patch.2 whereas the file contained in the zip is patch.3.
Not sure if that is intended and makes a difference here.

mikeller:
Thanks @chaehni.

> Hi @mikeller Thank you for the update. I downloaded some of the newest dives to a fresh logbook. Unfortunately, I still can't see any location information.

My bad, I removed the code that populates the location when I did the final cleanup of my test code before pushing. Fixed now.

> One small thing I noticed, the linked artifact here in Github is called patch.2 whereas the file contained in the zip is patch.3. Not sure if that is intended and makes a difference here.

That's expected. It's a consequence of GitHub running the CICD build on a speculative merge of the pull request content into the target branch - so there's an extra commit in the version that is built.

chaehni:
Now it works!
Really cool. Thanks everyone. :)

PR #4669, opened 2025-12-27
More iOS fixes

dirkhh:

- [x] Build system change

I think I'm effectively the only one building iOS binaries.
This gets all the dependencies back to working from a clean build and reduces some of the pointless noise when building from the command line -- you still need to build from within Xcode in order to get a binary that you can install on device or submit to the AppStore.
But it's at least an incremental improvement.

I'm reasonably certain that this should have no impact on anything else since all changes are in packaging/ios so I expect the CICD to be pointless - but maybe @mikeller is willing to take a quick look if anything looks obviously stupid...

(and maybe at some point there'll be a second person who uses their macOS machine to do local iOS builds 😇 )


mikeller (approved):
Looks reasonable to me. And I am definitely keen to have a go at building this locally, and check if there is a way to get the iOS mobile-on-desktop builds for iOS going in CICD again.

PR #4670, opened 2025-12-30
remove coverity

dirkhh:
Their analysis tool is broken and crashes with C++17 code. When I tried to submit a bug report they made it clear that they no longer are interested in supporting open source projects using Coverity and only will respond to requests from paying customers.

So let's drop this.

Obviously this doesn't need a PR and can just be merged, but I wanted to give @mikeller and @glance- (who initially added Coverity support a hundred years ago) a chance to comment.


mikeller (approved):
Makese sense to me - this seems to have been broken for a long time now.

And having a pull request is always a good idea, as it makes it easy to discuss and later find things.
It's only when CICD changes are somewhat explorative and may require further changes to make them work as intended that pull requests get noisy and create a lot of overhead.

glance-:
Some ideas in compensating for removing coverity would be setting up CI jobs running clang-scan , ASan, UBSan and possibly MSan. the llvm suite contains quite powerful tools for finding potential bugs nowadays.

clang-scan is pretty straight forward, just wrap the build step with clang-scan and it will analyze all the code for potential bug. As usual with this kind of tools about 90% will be false positives, but it usually finds a couple of real ones.

ASan and UBsan with friends are also quite straight forward, but they depend on having good test coverage actually exercising the relevant code paths in tests. To my memory, subsurface isn't that well overaged in tests, but at least it will help with what's covered by tests.

MSan is a bit trickier, because it requires all dependencies to be instrumented, which would involve building all of them without asm optimizations (looking at you ffmpeg...), QT and everything else. It have helped me find a couple of bugs in other projects but it's not that low hanging fruit so I'd suggest putting it at the end off the list of nice-to-have.


Of course, one initial step would be for a motivated developer to hand build subsurface with these tools and play around with them. My recommendation would be to start with clang-scan.

PR #4677, opened 2026-01-06
Add average depth input field for manual dive entry on iOS

Michael-Tanzer:
Currently, the iOS version of Subsurface does not allow users to input an average depth when manually adding a dive. This prevents proper SAC (Surface Air Consumption) calculation, which depends on the average depth.

This change adds:
- Average depth input field in the dive edit UI
- Model support to expose average depth data
- Display of average depth in the dive view
- Backend validation and storage of average depth
- Preservation of user-entered average depth when generating dive profiles

The implementation follows the same pattern as the max depth field and allows users to clear the average depth by leaving the field empty.

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR adds the ability for iOS users to input an average depth when manually adding or editing a dive. Previously, this field was missing from the mobile UI, which prevented proper SAC rate calculations for manually entered dives.

**Problem:**
- SAC calculation requires average depth to compute gas consumption rates
- Without average depth input, manually added dives could not have accurate SAC values

**Solution:**
The implementation adds average depth support throughout the mobile stack:
1. **UI Layer**: Added "Average depth" input field in the dive edit form, positioned after the max depth field
2. **Model Layer**: Added `AverageDepthRole` to expose mean depth data from the dive structure
3. **View Layer**: Added display of average depth in the dive view (shown as "Avg: [depth]" when available)
4. **Backend Layer**: Added `checkMeanDepth()` function to validate and store average depth, following the same pattern as `checkDepth()`
5. **Data Preservation**: User-entered average depth is preserved when `fake_dc()` generates dive profiles

The field is optional.

### Changes made:

1) **UI Components** (`mobile-widgets/qml/DiveDetailsEdit.qml`):
   - Added "Average depth:" input field with validation
   - Added property alias `averageDepthText` for data binding
   - Updated `clearDetailsEdit()` to reset average depth field
   - Updated `saveData()` to pass average depth to backend

2) **View Components** (`mobile-widgets/qml/DiveDetailsView.qml`):
   - Added conditional display of average depth in dive view
   - Shows "Avg: [depth]" when average depth data is available

3) **Data Binding** (`mobile-widgets/qml/DiveDetails.qml`):
   - Added property alias for average depth
   - Populates average depth field when editing existing dives

4) **Model Layer** (`qt-models/mobilelistmodel.h`, `qt-models/mobilelistmodel.cpp`, `qt-models/divetripmodel.cpp`):
   - Added `AverageDepthRole` enum value
   - Added role name mapping "averageDepth"
   - Implemented data retrieval from dive structure's `meandepth` field

5) **Backend Validation** (`mobile-widgets/qmlmanager.h`, `mobile-widgets/qmlmanager.cpp`):
   - Added `checkMeanDepth()` function declaration and implementation
   - Validates average depth input (0-500m range, same as max depth)
   - Updates both `dive.meandepth` and `dive.dcs[0].meandepth` for consistency
   - Updated `commitChanges()` signature to accept `averageDepth` parameter
   - Added call to `checkMeanDepth()` in commit flow
   - Preserves user-entered average depth when `fake_dc()` generates profiles

### Documentation change:

**User Experience Changes:**
- iOS users can now input average depth when manually adding or editing dives
- Average depth is displayed in the dive view when available
- This enables proper SAC calculation for manually entered dives
- The field appears in the edit form between "Depth" and "Duration" fields

**Documentation Updates Needed:**
- Mobile user manual should mention the average depth field in the manual dive entry section
- Consider adding a note about how average depth affects SAC calculations

### DIFF
--- a/mobile-widgets/qml/DiveDetails.qml
+++ b/mobile-widgets/qml/DiveDetails.qml
@@ -24,6 +24,7 @@ Kirigami.Page {
 	property alias diveguideModel: detailsEdit.diveguideModel
 	property alias tagText: detailsEdit.tagText
 	property alias depth: detailsEdit.depthText
+	property alias averageDepth: detailsEdit.averageDepthText
 	property alias duration: detailsEdit.durationText
 	property alias location: detailsEdit.locationText
 	property alias gps: detailsEdit.gpsText
@@ -347,6 +348,7 @@ Kirigami.Page {
 		gps = modelData.gps
 		duration = modelData.duration
 		depth = modelData.depth
+		averageDepth = modelData.averageDepth !== undefined ? modelData.averageDepth : ""
 		airtemp = modelData.airTemp
 		watertemp = modelData.waterTemp
 		suitIndex = manager.suitList.indexOf(modelData.suit)
--- a/mobile-widgets/qml/DiveDetailsEdit.qml
+++ b/mobile-widgets/qml/DiveDetailsEdit.qml
@@ -31,6 +31,7 @@ Item {
 	property alias notesText: txtNotes.text
 	property alias durationText: txtDuration.text
 	property alias depthText: txtDepth.text
+	property alias averageDepthText: txtAverageDepth.text
 	property alias weightText: txtWeight.text
 	property var usedGas: []
 	property var endpressure: []
@@ -62,6 +63,7 @@ Item {
 		detailsEdit.locationText = ""
 		detailsEdit.durationText = ""
 		detailsEdit.depthText = ""
+		detailsEdit.averageDepthText = ""
 		detailsEdit.airtempText = ""
 		detailsEdit.watertempText = ""
 		detailsEdit.diveguideText = ""
@@ -117,7 +119,7 @@ Item {
 
 		// apply the changes to the dive_table
 		manager.commitChanges(dive_id, detailsEdit.number, detailsEdit.dateText, locationBox.editText, detailsEdit.gpsText, detailsEdit.durationText,
-				      detailsEdit.depthText, detailsEdit.airtempText, detailsEdit.watertempText,
+				      detailsEdit.depthText, detailsEdit.averageDepthText, detailsEdit.airtempText, detailsEdit.watertempText,
 				      suitBox.editText, buddyBox.editText, diveguideBox.editText, detailsEdit.tagText,
 				      detailsEdit.weightText, detailsEdit.notesText, startpressure,
 				      endpressure, usedGas, usedCyl,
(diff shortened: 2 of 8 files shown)


dirkhh:
How would a user know their accurate average depth, unless they have a dive computer. 
And then just download from that?
This makes no sense to me. I see no realistic use case

Michael-Tanzer:
> How would a user know their accurate average depth, unless they have a dive computer. And then just download from that? This makes no sense to me. I see no realistic use case

Hi @dirkhh. Thank you for the feedback.

This is a problem I am personally facing and other people I dived with recently. As someone that doesn't own a dive computer I have 2 options:
1. Rent one from a dive shop: in this case the computer is often very cheap and doesn't support syncing at all (it does, however, show the average depth).
2. Don't use a dive computer and read the data off of someone else's computer: in this case it's much faster to copy the relevant information instead of pairing and syncing. Obviously in this case it won't be perfect, but enough for a reasonable measurement over the course of a dive (assuming sometimes you float above and sometimes below fellow divers, which is a reasonable assumption in my opinion).

Perhaps it's not an extremely common issue, but there is a realistic use case.

I would also like to point out that other major apps for logging dives (e.g., SSI) support this.

mikeller (changes requested):
This fails the most basic 'does it work' test even in the mobile version, which is its main use case according to the description:

- manually add a dive, explicitly set an 'average depth'

Image

- open the dive from the dive list - the correct average depth matching the dive profile is shown

Image

This is hardly surprising, as Subsurface is built around the invariant that the average depth is calculated on the fly, based on the profile of the dive. This holds true even for manually added dives, and results in any explicitly set 'average depth' being overwritten every time a dive is validated before it is stored / after it has been loaded.

The only way I can see this work cleanly, without any confusing workarounds to allow incorrect 'average depths' to be persisted is to add a new dive type for 'dive without a profile'. But this will likely break all kinds of things in all kinds of places - which is likely why this does not exist in Subsurface, in contrast to other dive log applications that are designed to be more tolerant of incorrect data.

@Michael-Tanzer: This looks like something that was generated by an LLM. In the future, if you decide to take this approach, please make sure that you **run some basic test on the code to make sure it actually works**.

In order to make this acceptable the following is needed:
- a robust implementation of support for dives with an assumed average depth;
- support across all platforms
- unit tests

As a workaround for the use case that you describe, it will be relatively easy to track the average depth by adding something like 'Average depth: 15.2m' to the end of the notes field, or directly add the SAC based on what you get from a calculator like https://www.divebuddy.com/mobile/calculator/sac.aspx

atdotde:
Long ago, in the desktop version we had a way to manually enter a dive just giving dive time, max depth and average depth and then it would create a fake profile that matches those data.

mikeller:
> Long ago, in the desktop version we had a way to manually enter a dive just giving dive time, max depth and average depth and then it would create a fake profile that matches those data.

That would be a good way to handle the use cases described by @Michael-Tanzer.

Michael-Tanzer:
I apologise, it wasn't done by an LLM, but it was definitely done a bit hastily. I definitely should have tested it, sorry.

If there is interest in the feature, I can try implementing it as @atdotde suggested. Would that be ok?

mikeller:
> I apologise, it wasn't done by an LLM, but it was definitely done a bit hastily. I definitely should have tested it, sorry.

Yes please. Unfortunately building for iOS is a bit involved, so you might have to make arrangements to build and test with Android, or with the 'mobile-on-desktop' build. And if you need build artifacts from GitHub for testing, please mark the pull request to generate the artifacts as 'Draft' so maintainers know not to review it yet.

> If there is interest in the feature, I can try implementing it as @atdotde suggested. Would that be ok?

Yes, I think this would be a good way of getting this functionality - it will require relatively local changes (to figure out a profile that fits the duration / max depth / average depth parameters), but then all of the assumptions about dive profiles made across the codebase will be satisfied.
But I think it should be implemented in a way that makes it available in the mobile and desktop version, i.e. with the calculation in the backend.

PR #4679, opened 2026-01-11
WIP: Liteprinter

atdotde:
These are the first steps using QLiteHtml for printing logbook pages in Qt6.

State of affairs: It prints logbook pages but all sizes are wrong. But I think it looks promising.
It does not try to really parse the html (which was a show stopper in the past) but simply uses a regex to do the 
div class="diveProfile" to img replacement. This could also be done with xslt but not by me.

It requires the latest version of litehtml which contains a fix for a bug that would overlap tables. Maybe @dirkhh could update his QLiteHtml branch.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation


dirkhh:
I'm traveling (actually diving, so not complaining), it may take be a couple of days before I get to this.

dirkhh:
OK, I didn't review the code, and I didn't update your PR, but if you apply this:
```
diff --git a/scripts/get-dep-lib.sh b/scripts/get-dep-lib.sh
index 731a875ae..f5ed96326 100755
--- a/scripts/get-dep-lib.sh
+++ b/scripts/get-dep-lib.sh
@@ -19,7 +19,7 @@ CURRENT_MDBTOOLS="v1.0.0"
 CURRENT_QT_ANDROID_CMAKE="master"
 CURRENT_LIBMTP="v1.1.22"
 CURRENT_LIBRAW="0.21.4"
-CURRENT_QLITEHTML="b8f9096eae730ec9464b24874e4e6312144aa9ce"
+CURRENT_QLITEHTML="865c9bb2eb9dcef6a0fed5bc5c4a9ab4bd777f74"
 
 croak() {
        echo $@
```
you should get a qlitehtml library with the latest litehtml code.

desktop-widgets/printer.cpp
```diff
@@ -12,15 +12,20 @@
...
 #include <QWebElementCollection>
 #include <QWebElement>
+#endif
 
 Printer::Printer(QPaintDevice *paintDevice, const print_options &printOptions, const template_options &templateOptions, PrintMode printMode, dive *singleDive) :
 	paintDevice(paintDevice),
+#if defined(USE_WEBKIT)
 	webView(new QWebView),
-	printOptions(printOptions),
-	templateOptions(templateOptions),
```
dirkhh, line 23: just doing a quick read-through before heading to the harbor...
since you `#ifdef` that code, why are you removing those two lines? Or is that an unrelated change?

desktop-widgets/printer.cpp
```diff
@@ -15,21 +15,24 @@
 #if defined(USE_QLITEHTML)
 # include <QUrl>
 # include <QFile>
+# include <qlitehtmlwidget.h>
 #elif defined(USE_WEBKIT)
-# include <QtWebKitWidgetst>
+#include <QtWebKitWidgetst>
```
dirkhh: odd whitespace change? either we do the space indent, or we don't?

desktop-widgets/printer.cpp
```diff
@@ -15,21 +15,24 @@
...
 
 Printer::Printer(QPaintDevice *paintDevice, const print_options &printOptions, const template_options &templateOptions, PrintMode printMode, dive *singleDive) :
 	paintDevice(paintDevice),
-#if defined(USE_WEBKIT)
-	webView(new QWebView),
-#endif
+	templateOptions(templateOptions),
 	printMode(printMode),
 	singleDive(singleDive),
+	printOptions(printOptions),
```
dirkhh: ah, I think this explains what my first comment was about... but then it should happen at the same time, in the same commit, no?

desktop-widgets/printer.cpp
```diff
@@ -241,15 +251,31 @@ void Printer::previewOnePage()
...
 			webView->setHtml(t.generate(getDives()));
+#else
+		;
+#endif
 		else if (printOptions.type == print_options::STATISTICS )
+#ifdef USE_WEBKIT
 			webView->setHtml(t.generateStatistics());
+#else
+			;
+#endif
```
dirkhh, line 273: not my favorite way to express this -- why not just `#ifdef` the whole `if/else-if`?
What am I missing?

desktop-widgets/printer.cpp
```diff
@@ -312,7 +311,7 @@ void Printer::Preview(QString content)
 		}
 		return QByteArray();
 	});
-	previewWidget.setGeometry(QRect(0,0,1600,1000));
+	previewWidget.setGeometry(QRect(0,0,1200,1000));
```
dirkhh, line 317: frankly, neither of these hard coded values is the right thing to do. Something-something based on the current window size? But that's me just being annoying, this is good for now

desktop-widgets/printer.cpp
```diff
@@ -327,8 +326,15 @@ void Printer::Preview(QString content)
 					  " { selection-background-color: %1; selection-color: %2; }")
 					  .arg(colorBack).arg(colorText));
 
-	previewWidget.setHtml(content);
+	// QFile file("/Users/Helling_1/logbuch.html");
+	// if (!file.open(QIODevice::ReadOnly | QIODevice::Text))
+	// 	return;
```
dirkhh, line 327: 😁
maybe not in the final version?

desktop-widgets/printer.cpp
```diff
@@ -333,6 +340,10 @@ void Printer::Preview(QString content, QPrinter *printer)
 	// QTextStream in(&file);
 	// previewWidget.setHtml(in.readAll());
 
+	QString repl = QString("<img height=\"30%\" width=\"30%\" src=\"file:///%1/dive_\\1.png\">").arg(printDir.path());
+	content.replace(QRegularExpression("<div\\s+class=\"diveProfile\"\\s+id=\"dive_([^\"]*)\"\\s*>\\s*</div>"), repl);
```
dirkhh, line 333: brute force, but not wrong. And better than making it crazy complex and bringing in more dependencies, maybe.

dirkhh (commented):
A few comments, mostly pointless.
This looks like a really promising start, Robert.
Thanks for working on this!

dirkhh:
clearly missing a widget in order to build...

desktop-widgets/printer.cpp
mikeller: I can probably provide some XSLT to do this in a more robust way, adding support for things like self-closing elements, or additional properties on the element itself.

mikeller (commented):
It looks like this is pulling in some likely unwanted libdivecomputer changes - you probably need to make sure that libdivecomputer is on the latest version of `Subsurface-DS9` in your branch.

desktop-widgets/printer.cpp
mikeller: See https://github.com/atdotde/subsurface/pull/5 for a way to do the html transformation with XSLT.

bstoeger:
If QLiteHtml is supported on all platforms, I would suggest replacing the old code instead of adding ifdef-ery complexity.

desktop-widgets/printer.cpp
```diff
@@ -254,3 +290,53 @@ void Printer::previewOnePage()
...
+		}
+		return QByteArray();
+	});
+	previewWidget.setGeometry(QRect(0,0,1200,1000));
+	QString colorBack = previewer.palette().highlight().color().name(QColor::HexRgb);
+	QString colorText = previewer.palette().highlightedText().color().name(QColor::HexRgb);
+	previewWidget.setStyleSheet(QString(
+					  "QLiteHtmlWidget"
+					  " { selection-background-color: %1; selection-color: %2; }")
+					  .arg(colorBack).arg(colorText));
```
bstoeger, line 323: Does `.arg(colorBack, colorText)` not work?

Yeah, this is not performance sensitive, but still...

PR #4682, opened 2026-01-14
Fix rendering of dive data with special characters and switch geo-lookups to HTTPS

RinZ27:
The HTML export currently fails to correctly render dive locations or notes if they contain characters like '<' or '&', as they are injected directly into the DOM via 'innerHTML'. This refactor adds a basic escaping helper to ensure the output remains consistent and safe regardless of the input content.

Additionally, updated the Geonames API URLs to use HTTPS. Querying location data over cleartext is unnecessary and should be avoided.

### DIFF
--- a/core/divesitehelpers.cpp
+++ b/core/divesitehelpers.cpp
@@ -78,9 +78,9 @@ static QJsonObject doAsyncRESTGetRequest(const QString& url, int msTimeout)
 /// Performs a reverse-geo-lookup of the coordinates and returns the taxonomy data.
 taxonomy_data reverseGeoLookup(degrees_t latitude, degrees_t longitude)
 {
-	const QString geonamesNearbyURL = QStringLiteral("http://api.geonames.org/findNearbyJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
-	const QString geonamesNearbyPlaceNameURL = QStringLiteral("http://api.geonames.org/findNearbyPlaceNameJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
-	const QString geonamesOceanURL = QStringLiteral("http://api.geonames.org/oceanJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
+	const QString geonamesNearbyURL = QStringLiteral("https://api.geonames.org/findNearbyJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
+	const QString geonamesNearbyPlaceNameURL = QStringLiteral("https://api.geonames.org/findNearbyPlaceNameJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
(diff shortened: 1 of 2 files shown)


mikeller (approved):
Looks good to me - would be great to have some tests as well for the HTML / JS escaping.

core/divesitehelpers.cpp
```diff
@@ -78,9 +78,9 @@ static QJsonObject doAsyncRESTGetRequest(const QString& url, int msTimeout)
 /// Performs a reverse-geo-lookup of the coordinates and returns the taxonomy data.
 taxonomy_data reverseGeoLookup(degrees_t latitude, degrees_t longitude)
 {
-	const QString geonamesNearbyURL = QStringLiteral("http://api.geonames.org/findNearbyJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
-	const QString geonamesNearbyPlaceNameURL = QStringLiteral("http://api.geonames.org/findNearbyPlaceNameJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
-	const QString geonamesOceanURL = QStringLiteral("http://api.geonames.org/oceanJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
+	const QString geonamesNearbyURL = QStringLiteral("https://api.geonames.org/findNearbyJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
+	const QString geonamesNearbyPlaceNameURL = QStringLiteral("https://api.geonames.org/findNearbyPlaceNameJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
+	const QString geonamesOceanURL = QStringLiteral("https://api.geonames.org/oceanJSON?lang=%1&lat=%2&lng=%3&radius=50&username=dirkhh");
```
dirkhh, line 83: fascinating that these all run with my username 🤯

dirkhh (approved):
looks fine

core/divesitehelpers.cpp
RinZ27: @dirkhh Haha, I noticed that too when checking the diff! I just updated the protocol to HTTPS and left the params exactly as they were to be safe. Let me know if you want me to address that in a separate PR.

jcnventura:
Wanted to comment here that the HTTPS solution here was incomplete, as it is also necessary to change the hostname. I've created PR #4776 to address that.

mikeller:
> Wanted to comment here that the HTTPS solution here was incomplete, as it is also necessary to change the hostname. I've created PR #4776 to address that.

'Incomplete' is a big word to use here - the approach worked at the time.

jcnventura:
The change here to simply add "s" to the protocol failed to take into account some remarks dating from 2017 where the HTTPS API endpoint is supposed to be secure.geonames.org.

The geonames.org documentation at https://www.geonames.org/export/web-services.html also clearly states "Secure endpoint available at secure.geonames.org"

Maybe incomplete is too strong a word, let's settle for "missing some information"?

RinZ27:
@jcnventura Thanks for the catch on the Geonames secure hostname. I missed that detail, but it's great to see #4776 handling the hostname change correctly.

PR #4689, opened 2026-01-22
Add the Timezone Offset to Extra Info.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

**Update:** This now shows the dive computer reported time zone offset in the read only 'Date' field on the 'Dive Computer Info' tab.

Essentially a 'cut out the middleman' for
https://github.com/subsurface/libdc/pull/100.

### Additional information:

image

### DIFF
--- a/core/divecomputer.cpp
+++ b/core/divecomputer.cpp
@@ -375,6 +375,12 @@ void add_extra_data(struct divecomputer *dc, const std::string &key, const std::
 	// Include support for dives imported with the legacy Garmin / Uwatec backends
 	if (key == STRING_KEY_FIRMWARE_VERSION || key == "Firmware" || key == "DC Firmware Version")
 		dc->fw_version = value;
+	if (key == STRING_KEY_TIMEZONE_OFFSET) {
+		char *end = nullptr;
+		long offset = strtol(value.c_str(), &end, 10);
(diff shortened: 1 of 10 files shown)


bstoeger:
Frankly, as I noted on the mailinglist - this is precisely the data, which I would store explicitly in `struct dive` and `struct divecomputer`.

In extra_data, I would store things like "Central European daylight saving time".

bstoeger:
> Frankly, as I noted on the mailinglist - this is precisely the data that I would store explicitly in `struct dive` and `struct divecomputer`.
> 
> In extra_data, I would store things like "Central European daylight saving time".

mikeller:
> Frankly, as I noted on the mailinglist - this is precisely the data, which I would store explicitly in `struct dive` and `struct divecomputer`.

Yes, we can definitely consider storing it there as well - but this will not give us much unless we surface it to the user in some form. And when it comes to surfacing it, 'Extra Info' seems to be a good place at the moment.

> In extra_data, I would store things like "Central European daylight saving time".

After looking through libdivecomputer, I don't think any of the current dive computer models support this granularity - 'offset from UTC' seems to be the best they can do.

PR #4690, opened 2026-01-24
Make dive::when private as first step for "timezone" support

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The goal here is to make it clear that time-accesses are with respect to local time. Once we support "offset to utc" it should be always made explicit, whether we're treating local time (as in local, where/when the dive took place) or time relative to UTC (for example when sorting dives or calculating N2 loading between consecutive dives).

Therefore, all time accesses should be via accessor function. This PR does that for `struct dive`.  Timestamps are also stored in `struct diveplan` and `struct divecomputer`. However, these are mostly accessed by local code and therefore I suggest tacking those, when introducing a `timeoffset` field.

Completely untested (apart from starting and loading a log)!

### DIFF
--- a/backend-shared/exportfuncs.cpp
+++ b/backend-shared/exportfuncs.cpp
@@ -49,7 +49,7 @@ static QString profileText(const struct dive &dive)
 
 	if (dive.dive_site && dive.dive_site->name.length() > 0)
 		text += "🗺 " + QString::fromStdString(dive.dive_site->name) + " \n";
-	if (dive.when) {
+	if (dive.get_time_local()) {
 		text += "🗓 " + formatDiveDateTime(&dive) + "\n";
 	}
 	text += "⏱: " + formatDiveDuration(&dive) + "\n";
@@ -172,7 +172,7 @@ void export_TeX(const char *filename, bool selected_only, bool plain, ExportCall
 		cb.setProgress(done++ * 1000 / todo);
 		exportProfile(*profile, *dive, texdir.filePath(QString("profile%1.png").arg(dive->number)), false);
 		struct tm tm;
-		utc_mkdate(dive->when, &tm);
+		utc_mkdate(dive->get_time_local(), &tm);
 
 		std::string country;
 		dive_site *site = dive->dive_site;
--- a/commands/command.cpp
+++ b/commands/command.cpp
@@ -33,6 +33,11 @@ void shiftTime(const std::vector<dive *> &changedDives, int amount)
 	execute(new ShiftTime(changedDives, amount));
 }
 
+void setUtcOffset(const std::vector<dive *> &changedDives, std::optional<int32_t> offset)
+{
+	execute(new SetUtcOffset(changedDives, offset));
+}
+
 void renumberDives(const QVector<QPair<dive *, int>> &divesToRenumber)
 {
 	execute(new RenumberDives(divesToRenumber));
--- a/commands/command.h
+++ b/commands/command.h
@@ -45,6 +45,7 @@ void addDive(std::unique_ptr<dive> d, bool autogroup, bool newNumber);
 void importDives(struct divelog *log, int flags, const QString &source); // The tables are consumed!
 void deleteDive(const QVector<struct dive*> &divesToDelete);
 void shiftTime(const std::vector<dive *> &changedDives, int amount);
+void setUtcOffset(const std::vector<dive *> &changedDives, std::optional<int32_t> offset);
 void renumberDives(const QVector<QPair<dive *, int>> &divesToRenumber);
 void removeDivesFromTrip(const QVector<dive *> &divesToRemove);
 void removeAutogenTrips();
--- a/commands/command_base.cpp
+++ b/commands/command_base.cpp
@@ -56,7 +56,7 @@ QString diveNumberOrDate(struct dive *d)
 	if (d->number != 0)
 		return QStringLiteral("#%1").arg(d->number);
 	else
-		return QStringLiteral("@%1").arg(get_short_dive_date_string(d->when));
+		return QStringLiteral("@%1").arg(get_short_dive_date_string(d->get_time_local()));
 }
 
 QString getListOfDives(const std::vector<struct dive*> &dives)
--- a/commands/command_divelist.cpp
+++ b/commands/command_divelist.cpp
@@ -630,7 +630,7 @@ void ShiftTime::redoit()
 
 	// Send signals
 	QVector<dive *> dives = stdToQt<dive *>(diveList);
-	emit diveListNotifier.divesTimeChanged(timeChanged, dives);
+	emit diveListNotifier.divesTimeChanged(dives);
 	emit diveListNotifier.divesChanged(dives, DiveField::DATETIME);
 
 	// Select the changed dives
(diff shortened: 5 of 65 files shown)


RonOnGithub:
I *really* think we need to pump the brakes a bit on this one.

You can't solve every problem with "just wrap it in an accessor", and I think  it's very premature to be building a public "API" when we haven't got a design for a solution, a common understanding of what such a design needs to be able to do or even what units and data types it needs to properly represent the problem.  Let alone consensus on just about anything else about it.  And I very strongly think that All Of Those Things need to be built first.

The implementation is the easy bit if we get the design right.  The design bit just gets harder if we're neck deep in a premature implementation that wasn't built around a solid understanding of the problem.

If things like ```dive->set_time_local_dc(utc_mktime(&tm));``` aren't a massive red-flag, they really ought to be.  And that's before we start on how extant "local time" APIs all work relative to *actual* local time of where you are right now, which isn't at all the time we are mostly interested in here.  Or how a time_t count in seconds is not well specified to count in anything except UTC, or how set_time_local has no way to return an error if an invalid "local time" is passed to it.

A huge part of the backend work that I'm currently doing for the Qt/QML update is *removing* duplicated code and 'failed' APIs that were added prematurely and then either never used how they were imagined, or their use was reverted and they were never entirely removed again.  And I'd love it if we don't add more of them before that work is done.

So please, let's first come up with a well thought out design, that's fit for purpose, which we can then review any potential implementation against.  Because it's really hard to review something with a straight face that's basically "this is an API, we can figure out how it works later".  And I'm sorry if that sounds a bit brutal, but I (innocently, I swear!) threw the match that lit this fire, so I also feel a bit responsible for fighting it where it has broken the original containment lines.

There's more things I could pick on specifically about this patch, but that's not what I want to waste time on, or the kind of conversation I'd like to see about what the future wrt to this really might or ought to look like.

bstoeger:
> I _really_ think we need to pump the brakes a bit on this one.
> 
> You can't solve every problem with "just wrap it in an accessor"

I never said that I want to solve "every problem". I clearly said "first step".

> The implementation is the easy bit if we get the design right. The design bit just gets harder if we're neck deep in a premature implementation that wasn't built around a solid understanding of the problem.

No offense, but for me that's meaningless management yackety-yak. I'm not a software engineer and have zero intentions of becoming one. But, out of necessity I've seen enough large code bases, so that I don't require lectures on software design.
 
The point of this PR is to provide a "solid understanding of the problem".

> If things like `dive->set_time_local_dc(utc_mktime(&tm));` aren't a massive red-flag, they really ought to be.

Looks funky yes. But this is what we were doing all the time and the objective was to make these things explicit. Goal achieved.

> A huge part of the backend work that I'm currently doing for the Qt/QML update is _removing_ duplicated code and 'failed' APIs that were added prematurely and then either never used how they were imagined, or their use was reverted and they were never entirely removed again. And I'd love it if we don't add more of them before that work is done.

Sorry, I haven't followed the discussions recently. Where can I see this "backend work"? I hope this comes in small digestible packages? I'd hate to have to be presented with a huge fait accompli.

RonOnGithub:
> > I _really_ think we need to pump the brakes a bit on this one.
> > You can't solve every problem with "just wrap it in an accessor"
> 
> I never said that I want to solve "every problem". I clearly said "first step".

On the mailing list you said. "How it's stored is a minor implementation detail once everything is accessed by accessor functions." - which might be true for some problems, but it very much isn't for this one.  How and what we store is the *entire crux* of getting this right, and it's the accessor methods in this case which are the minor implementation detail.

> > The implementation is the easy bit if we get the design right. The design bit just gets harder if we're neck deep in a premature implementation that wasn't built around a solid understanding of the problem.
> 
> No offense, but for me that's meaningless management yackety-yak. I'm not a software engineer and have zero intentions of becoming one. But, out of necessity I've seen enough large code bases, so that I don't require lectures on software design.

None taken, so I hope you'll see my comments in the same light when I say the reason that seems meaningless to you is because you don't actually understand the problem yet.  This isn't a software design problem.  This is a Political Problem.  Because "local time" is a political problem.  And you can't just pretend that "local time" and "UTC" are simply interchangeable, or even the same kind of problem.  Or that an API which works for one will also work for the other.  There is no 1:1 mapping between them, there is no simple mathematical function that connects them.  There is no sane interpretation of local time as a time_t count of seconds, because local time is arbitrarily discontinuous.  And our problem here isn't just "local time", it's an arbitrary collection of an arbitrary number of mazes of twisty little local times, all different.  And all shown together in the same divelist.  Sometimes even as part of the same dive, if you have a locally created dive plan and multiple different devices recording the live dive data.

The people who didn't account for those things in their software design are responsible for everything that is horrible wrt handling time in software and device firmware.  A lot of very experienced software designers have got this wrong before you.

This is why Dirk and Linus are so hesitant to open this can of worms in the first place.  The problems are subtle, and solving them portably is even worse.  And I'm all for letting people learn from their own mistakes, but we do have enough people here with enough prior experience with this, that *this* codebase is not the right place to watch those mistakes be repeated.

> The point of this PR is to provide a "solid understanding of the problem".

If you have absolutely no intention of merging this, then sure, kick it around your own garden and see what talking points might shake out and what you learn from it - but if you want the spoiler alert from someone who has dealt with this multiple times before - your first attempt is so far heading in almost the polar opposite direction to a "solid solution to the problem".  And that's no slight on your "software design" skill, it's just that kind of problem.

> > If things like `dive->set_time_local_dc(utc_mktime(&tm));` aren't a massive red-flag, they really ought to be.
> 
> Looks funky yes. But this is what we were doing all the time and the objective was to make these things explicit. Goal achieved.

Yes, I get you that you simply replaced the existing code with your new accessor, with no actual functional change, but assigning a UTC time_t to a "local time" should ring the same kind of alarm bells that ```#define CAT DOG``` would as to the applicability and logical correctness of the proposed interface.

The only thing that is really explicit about that is that there is no sense where that makes sense.  And I don't think that was really your goal?

> > A huge part of the backend work that I'm currently doing for the Qt/QML update is _removing_ duplicated code and 'failed' APIs that were added prematurely and then either never used how they were imagined, or their use was reverted and they were never entirely removed again. And I'd love it if we don't add more of them before that work is done.
> 
> Sorry, I haven't followed the discussions recently. Where can I see this "backend work"? I hope this comes in small digestible packages? I'd hate to have to be presented with a huge fait accompli.

You could come and look over my shoulder :)  Otherwise you'll have to wait until I've finished iterating on it and have something ready to land, but we've been discussing aspects of it for a while now and it's exactly what the "QML / Qt 6 in Subsurface-mobile" in the subject line of the mailing list thread is referring to, and how this particular subject blew up in the first place, with me raising the question of how to handle it across that barrier.

I may be new around here, but we're a very long way from my first rodeo.  And I feel kind of responsible for (inadvertently) sparking the firestorm this has turned into, so it would be irresponsible to not take some ownership of that and the problem of trying to get it back into a more appropriately sized and shaped fire pit again.

So I really do hope we can either put this to rest for a while, like Dirk would *really* love us to do.  Or that we can at least all take a deep breath and slow down a little, and try to form a calm consensus on what problem we actually want to solve, what problems there are with trying to solve that, what the best solution might be, what the collateral damage from that is, and whether the cost/benefit makes it worth proceeding.  Because that might sound like "management speak" to you, but there really are people here whose starting point of understanding this is much closer to what is actually required than what this patch series demonstrates, and we don't have to follow Robert's Rules Of Order to have that discussion and learn from them and avoid repeating the same mistakes We've Already Seen Before.

That would be a nice goal to achieve :D   But really, we have a pretty long list of More Important things we should probably get to before this bubbles to the top of them.

PR #4691, opened 2026-01-24
Desktop: show depth and time for non-manual dives

bstoeger:
In the UI-definition of the notes-tab there was code suggesting that we wanted to show the depth and time of the dive even in non-manual dives.

Either we lost this, or it was never properly implemented.

Implement it, by showing QLabels with the information for non-manually added dives.what the intention was.

I tried reusing the QLineEdit and setting it to read-only for non-manual dives. However, the UI doesn't grey out the QLineEdit, making the UI confusing. Bad Qt.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

See commit description: show depth and time even for non-manual dives in a read-only QLabel.

If we do not want this, we should at least remove the spurious UI elements.

I lack time to dig in the history to find out whether we list this feature at one point or whether it was never implemented.

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveNotes.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveNotes.cpp
@@ -235,8 +235,10 @@ void TabDiveNotes::updateData(const std::vector<dive *> &, dive *currentDive, in
 		ui.NotesLabel->setText(tr("Trip notes"));
 		ui.notes->setText(QString::fromStdString(currentTrip->notes));
 		ui.depth->setVisible(false);
+		ui.depthFixed->setVisible(false);
 		ui.depthLabel->setVisible(false);
 		ui.duration->setVisible(false);
+		ui.durationFixed->setVisible(false);
 		ui.durationLabel->setVisible(false);
 	} else {
 		// make all the fields visible writeable
@@ -255,6 +257,8 @@ void TabDiveNotes::updateData(const std::vector<dive *> &, dive *currentDive, in
 		ui.dateEdit->setReadOnly(false);
 		ui.timeLabel->setVisible(true);
 		ui.timeEdit->setVisible(true);
+		ui.depthLabel->setVisible(true);
+		ui.durationLabel->setVisible(true);
 		/* and fill them from the dive */
 		ui.rating->setCurrentStars(currentDive->rating);
 		// reset labels in case we last displayed trip notes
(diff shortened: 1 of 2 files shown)


dirkhh (commented):
I have no idea what the thought process was or should be.
And I have no opinion on whether I want this here or not.
I have never missed it, so I guess... not?

mikeller (commented):
This definitely makes sense. But note that the units indicated in the label do not match what is shown if the dive duration isn't in full minutes:

Image

bstoeger:
> And I have no opinion on whether I want this here or not.

Me neither. It does make the UI somewhat more consistent ... I think?

bstoeger:
> This definitely makes sense. But note that the units indicated in the label do not match what is shown if the dive duration isn't in full minutes

Iteresting. But observe  that it is also possible to edit manually added dives to 1:13:30. So I guess this is an independent quirk?

bstoeger:
A bit of digging reveals: The ominous field that let me think this was intentional was added in commit 08db77333b8a6fe7be71f30bf0d5ee15a6bf01f1.

After reading said commit, I'm not sure if it was intentional after all.

mikeller:
> > And I have no opinion on whether I want this here or not.
> 
> Me neither. It does make the UI somewhat more consistent ... I think?

In my opinion the consistency is a plus. From personal experience I think this was not missed until now because depth / duration is also visible in the dive list. But a user might have hidden these columns, or scrolled them out of view.

> > This definitely makes sense. But note that the units indicated in the label do not match what is shown if the dive duration isn't in full minutes
> 
> Iteresting. But observe that it is also possible to edit manually added dives to 1:13:30. So I guess this is an independent quirk?

Yes, you are right - I had never tried to manually add seconds there. It's a weird approach in how we are formatting durations with / without seconds, which makes it unintuitive to read ambiguos times like '01:15'.

I also note that we are ignoring the 'Show hours in duration' preference here:

image

Looks like this is a can of worms. 😑

bstoeger:
> Yes, you are right - I had never tried to manually add seconds there. It's a weird approach in how we are formatting durations with / without seconds, which makes it unintuitive to read ambiguos times like '01:15'.

We could change the label to 'hh:mm(:ss)" or even dynamically change the label?

> I also note that we are ignoring the 'Show hours in duration' preference here:

True. However, as this is usually an input field, I think this is an acceptable inconsistency.

> Looks like this is a can of worms. 😑

Yeah, but mostly orthogonal to this PR.

I'm also fine with just removing the superfluous fields from the UI definition. After all, the information is shown on the "Information" and "Summary" tabs.

mikeller:
> > Yes, you are right - I had never tried to manually add seconds there. It's a weird approach in how we are formatting durations with / without seconds, which makes it unintuitive to read ambiguos times like '01:15'.
> 
> We could change the label to 'hh:mm(:ss)" or even dynamically change the label?
> 
> > I also note that we are ignoring the 'Show hours in duration' preference here:
> 
> True. However, as this is usually an input field, I think this is an acceptable inconsistency.

Makes sense. I'd probably show the 'hh:mm(:ss)' on the input field, and show not units and use the same formatting as in the dive list for the display field.

> I'm also fine with just removing the superfluous fields from the UI definition. After all, the information is shown on the "Information" and "Summary" tabs.

I can see the value of having the input fields as an easy way for divers to log dive depths / durations if they can't download from their dive computer, instead of requiring them to drag points in the profile.

But this _really_ is a can of worms, as the current implementation has got severe pitfalls:

https://photos.app.goo.gl/8cE93i38uk8aazEf7

bstoeger:
> I can see the value of having the input fields as an easy way for divers to log dive depths / durations if they can't download from their dive computer, instead of requiring them to drag points in the profile.

I just recently needed this input field. My new divecomputer annoyingly has much less memory than my old one and I lost profile data of two dives on my last trip owing to not downloading frequently enough. So I have to manually enter them.

As you note, dragging points on the profile to set the dive time, depth is not a reasonable UI.
 
> But this _really_ is a can of worms, as the current implementation has got severe pitfalls:

Fair. But these issues are completely unrelated to this PR, or am I mistaken?

So what are we going to do - add these fields or rsimply emove the spurious UI definitions? I have no preference.

bstoeger:
Ok, I now understand what these ominous fields are there for - they make sure that the date/duration fields do not change in size. Close.

PR #4693, opened 2026-01-25
Add seconds to duration label on dive notes tab

bstoeger:
The label says "(h:mm)", but factually, one also can enter seconds resulting values such as "1:23:45". Let the label reflect that.

The whole thing is a bit finicky.

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

See commit description: The label of the duration field was factually wrong, as it is also possible to enter second values (why not?).

See discussion in #4691.


mikeller (approved):

bstoeger:
Thanks.

PR #4694, opened 2026-01-26
Cloud staging

dirkhh:
thanks to @RonOnGithub a quick hack to test a staging server from a Subsurface binary.
This is mainly intended to create artifacts that can be used for that purpose


dirkhh:
no longer needed or wanted

PR #4696, opened 2026-01-28
Web UI

dirkhh:

- [x] New feature

add the CLI, the webUI, and some tests for the CLI.
purpose of the PR is to see the builds fail so I know what to clean up

there's obvious disagreement whether this should be merged at all and I really don't care either way. I can just as easily keep it in my branch and simply deploy on the cloud server from that branch. No reason to have that in master


RonOnGithub:
Well I for one am quite sad about this.
I was actually really looking forward to having this in the mainline.

I hope you're going to keep it in a fork or branch somewhere that I can play along, because I'm happy to ensure there's a patch for it to fix the main() sequence changes I have pending and anything else that I'm doing in the backend which might effect it.

PR #4697, opened 2026-01-30
Cli

dirkhh:

- [x] New feature
- [x] Code cleanup

As requested, here is just the CLI.
This is of course useless for most people and makes no sense to have around in its current state as it is custom-tailored for the webUI which will live in a separate repository.

While getting this to the point that the builds complete (famous last words) I noticed a bunch of bugs in the code where we assign NULL to std::string variables, which is a violation of C++ standards and which some versions of glibc/libstd-c++ take offense in.

As mentioned in the commits - part of the code was first developed with the help of Claude Code. I reviewed and tested all of the code, which led to some significant rewrites, so I'm comfortable signing off on it.


dirkhh:
well, that's not as bad as feared...

dirkhh:
> well, that's not as bad as feared...

HAHAHAHAHA

```
 git clone https://github.com/libusb/libusb.git libusb
Cloning into 'libusb'...
error: RPC failed; curl 56 LibreSSL SSL_read: LibreSSL/3.3.6: error:06FFF064:digital envelope routines:CRYPTO_internal:bad decrypt, errno 0
error: 1416 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
+ croak 'git clone https://github.com/libusb/libusb.git failed'
+ echo git clone https://github.com/libusb/libusb.git failed
```

I love myself a rock solid CICD...

cli/config.cpp
```diff
@@ -52,16 +131,98 @@ CliConfig loadConfig(const QString &configPath)
...
+	QString userid = obj.value("userid").toString();
+	if (userid.length() > 254) {  // RFC 5321 max email length
+		fprintf(stderr, "userid too long\n");
+		return config;
+	}
+	// Email pattern allowing common characters:
+	// Local part: alphanumeric, dots, plus (for sub-addressing), hyphen, underscore
+	// Domain: alphanumeric, dots, hyphens
+	// This covers user+subsurface@sub.domain.com style addresses
+	static QRegularExpression useridRe(
```
RonOnGithub: It might be nice to make a validateEmail() function that's a single source of truth for this.
We have a similar test in desktop-widgets/preferences/preferences_cloud.cpp where
we declare it valid if it just matches "^[a-zA-Z0-9@.+_-]+$".

I don't want to go down the rathole of what valid email and domains we exclude with this,
but it would be good to be consistent everywhere, so that we don't allow in the desktop UI
something that the server will then reject, or that the server will accept but the web UI will
then reject etc.

cli/config.cpp
```diff
@@ -52,16 +131,98 @@ CliConfig loadConfig(const QString &configPath)
...
 	return config;
 }
+
+// Create a secure temp directory with a random name under the configured temp_dir.
+// This prevents symlink attacks and race conditions by:
+// 1. Using a cryptographically random UUID in the directory name
+// 2. Creating with restrictive permissions (owner only)
+// 3. Verifying the created directory is actually a directory we own
+// Returns empty string on failure.
+QString createSecureTempDir(const CliConfig &config)
```
RonOnGithub, line 208: Does QTemporaryDir not do what you need here (given we're already in Q space).
[I won't be surprised if it doesn't exactly, just checking you know it exists].

cli/main.cpp
```diff
@@ -40,10 +40,11 @@ static void printUsage()
 
 int main(int argc, char *argv[])
 {
-	// QGuiApplication is needed for QGraphicsScene (used in profile rendering)
-	// We use offscreen platform to avoid needing a display
+	// QApplication (not QGuiApplication) is needed because ProfileScene uses
+	// QGraphicsScene which requires the full QtWidgets infrastructure.
+	// We use offscreen platform to avoid needing a display.
```
RonOnGithub, line 52: We have the same issue with the mobile app.  I'm not sure if we can fix it,
or that if we can that it will gain us anything worth the effort, but it went
on my "somewhere way down the line" list of nits to look at more.

RonOnGithub:
The std::string thing is a bit of mystery.  Not that it's invalid, but that it didn't blow up before.

The ball of flames is usually a runtime exception thrown though, so maybe some combination of Qt never growing up to handle them, it being built without exception support, really aggressive compile time dead code removal, and just never actually hitting the cases where NULL was returned?

Curiosity got the better of me, so I ran a quick test in some Qt code, and adding a ```std::string foo = NULL;``` in just the right place ends up with some really wild compilation errors.  No diagnostic about the invalid string assignment - but lots of errors related to the other code around it seeming to have completely ceased to exist (as if the result of the uncaught exception was being handled at compile time).

dirkhh:
We can argue whether that behavior is right or wrong under the C++ standards, but frankly, who the heck cares?
```
echo '#include <string>
int main() { std::string s(nullptr); }' > /tmp/test_null.cpp && g++ -o /tmp/test_null /tmp/test_null.cpp && /tmp/test_null
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string: construction from null is not valid
Aborted (core dumped)
```
I find this pretty compelling evidence that this is something we should avoid doing in our code.

you can reproduce this for yourself quite easily:

```
docker run -it --rm -v $(pwd):/src debian:bookworm bash

# apt update
# apt install g++ gdb
```
and then run the command above, or run the little binary under `gdb`, or...

RonOnGithub:
> I find this pretty compelling evidence that this is something we should avoid doing in our code.

Yes, sorry, I guess I wasn't clear.  No argument there at all, it's absolutely invalid and something we should never do.
It's a pretty common way code at a C/C++ boundary can fail that I've seen lots of times before.

I just got curious about why we weren't already seeing it blow up exactly like your test code above.

But things got even weirder when I tossed a random ```std::string f(NULL);``` into the qPrefLanguage constructor (which just happened to be the code I was in when I saw this) to try and deliberately provoke a failure.  Because that (with the current code I have), results in a massive sequence of compile time failures, with *lots* of errors like:

```
subsurface/core/settings/qPrefLanguage.cpp: In static member function ‘static void qPrefLanguage::set_system_locale_override(bool)’:
subsurface/core/settings/qPrefPrivate.h:172:56: error: ‘this’ pointer is null [-Werror=nonnull]
  172 |                         emit instance()->field##Changed(value);
```

Which seems entirely unrelated, and I have no idea how you'd ever connect that to the real cause - which appears to be that the compiler knows string(NULL) will throw, so the ctor will never complete, so the pointer in instance() will never be initialised, resulting in those errors.

There is still no compile time diagnostic about string(NULL), which isn't very surprising, but the compiler already knows that will happen and doesn't compile the 'impossible' branch.  And still didn't really answer why we haven't seen it failing before, but my first guess is something to do with the mixed environment for (lack of) exception handling Qt gives us.

So yes, your fix for this was perfectly correct.  I just got sucked into wondering why we hadn't seen it before.

.github/workflows/mac.yml
```diff
@@ -167,6 +167,27 @@ jobs:
...
+
+        echo "--------------------------------------------------------------"
+        echo "building CLI"
+        bash -e -x ./subsurface/scripts/build.sh -cli -build-with-qt6
+
+        echo "--------------------------------------------------------------"
+        echo "running CLI tests"
+        cd subsurface/build-cli
+        python3 ../cli/tests/test_cli.py --cli-path ./cli/subsurface-cli
+
```
mikeller, line 221: **suggestion:** I wonder if it would make sense to include the CLI as a build artifact on the platforms we build it for - and potentially even include it in the CICD releases as an artifact, or bundle it in the .dmg?

core/configuredivecomputerthreads.cpp
```diff
@@ -164,7 +164,7 @@ static dc_status_t read_suunto_vyper_settings(dc_device_t *device, DeviceDetails
 		if (desc) {
 			// We found a supported device
 			// we can safely proceed with reading/writing to this device.
-			deviceDetails.model = dc_descriptor_get_product(desc);
+			deviceDetails.model = dc_descriptor_get_product(desc) ?: "";
```
mikeller: **question:** Is this actually related to the CLI, and needed?

core/subsurfacestartup.cpp
```diff
@@ -19,6 +19,9 @@ int quit, force_root, ignore_bt;
 #ifdef SUBSURFACE_MOBILE_DESKTOP
 std::string testqml;
 #endif
+#ifdef SUBSURFACE_CLI
+std::string settings_suffix;
```
mikeller, line 25: **question:** Does it make sense to have this defined in different places depending on the build target, or should we move the definition to here for everything?

mikeller (approved):

core/configuredivecomputerthreads.cpp
RonOnGithub: See the discussion above about assigning (char*)NULL to std::string.  Or are you saying that cannot ever return NULL?

core/subsurfacestartup.cpp
RonOnGithub: This one is on the chopping block in my branch too, as part of the comand line options / default prefs / prefs / exporting them to QML changes, so I probably wouldn't sweat it too much as part of this changeset - we can come back to it once my changes merge with the new early-init code in this one.

.github/workflows/mac.yml
dirkhh: No one uses the CLI for anything. 
Until that changes, until it's actually useful for normal users, I don't see the point in that

dirkhh: No one uses the CLI for anything. 
Until that changes, until it's actually useful for normal users, I don't see the point in that

core/configuredivecomputerthreads.cpp
dirkhh: That's the crash fix we discussed in so much detail else where

Had nothing to do with implementing the CLI

dirkhh: That's the crash fix we discussed in so much detail else where

Had nothing to do with implementing the CLI

core/subsurfacestartup.cpp
dirkhh: I don't understand this comment

dirkhh: I don't understand this comment

RonOnGithub: I assumed Mike meant we have a setting_suffix global in qt-init.cpp and was talking about replacing that with a single instance that both the CLI and and other builds could use.

RonOnGithub:
Apologies for all the belated comments that only just appeared.
I only just realised this hell platform has a secret hidden button I also needed to click before all'yall could see them.

cli/config.cpp
dirkhh: This was very much coded with a harsh view of the on average ten thousand attempts a day to hack the cloud server. So it's fairly aggressive in restricting what arguments it accepts. Full disclosure - Claude Code helped with that particular code (since it has actually proven pretty good in spotting potential code weaknesses), so this is not entirely written by me - and it may indeed be overkill (or underkill?)

cli/config.cpp
dirkhh: same response as above. This was implemented by Claude since I didn't trust myself to be aware of all the ways people could abuse that code. Remember that this was all written with the thought that it was tightly coupled to the webUI - which of course is no longer the case. So I have no personal ambitions attached to all of this. Happy to have someone streamline it and follow some other code that may or may not exist. That's simply not how I created the code.

dirkhh:
Clearly there are old and new comments interspersed, and we managed to have discussions here and on the mailing list and who knows where else and this all feels a bit disconnected.

We could factor out the whole "don't allow code paths that could result in returning NULL from a function that creates a value that is assigned to a std:string because bad things can happen" changes into their own PR. 

I created this separate PR because some of you felt that half my code should be here, the other half somewhere else. Now some people appear to want this split into more PRs because those things aren't logically connected.

Y'all have write access to this repo - change at will. My work on this PR is complete. It does what I need - I have the code running on all the servers and it runs from my private repo so I know what's actually running on my servers. Whether or not this code goes into the official repo at this point makes no difference to me.

RonOnGithub:
I'm happy to see this pushed as is.

I'd like us to unify the cloud credentials validation at some point, but that can be done later.

cli/config.cpp
RonOnGithub: I don't actually mind what we restrict it to, so long as everywhere that checks it uses the same rules.

Obviously the server is definitive, but it would be nice if the clients use the same rules it does to catch anything it will reject before doing a round trip failure - so they can show a nice "pick another ID" warning instead of "it failed".  And leaving you the option to honeypot or ignore abusers without catching people using the legitimate clients.

Email is one of those things where what is "valid" according to the IETF cannot be determined by regex.  But we're not a mail server, so it's fine for our version of valid to be a subset of theirs for this use.

We can chat about this more later though, it doesn't have to be part of this changeset.

core/subsurfacestartup.cpp
mikeller: > I don't understand this comment

@dirkhh: Like so https://github.com/subsurface/subsurface/pull/4714/changes

dirkhh: sure, that works

dirkhh:
apparently rebasing this after merging the 'path fix' broke a bunch of builds. Something to look at later.

PR #4699, opened 2026-02-01
Turn "extra info" tab into "dive computer info" tab

bstoeger:

- [ ] Bug fix
- [ ] Functional change
- [x] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

The "Extra Info" tab is about additional info for each dive computer. Make that more clear.

Moreover, show the recorded start-time of each dive computer, which was stored, but to my knowledge not shown anywhere.

The UI is at the proof-of-concept stage. No idea how to do this properly (and little motivation to find out).

In my tests, this seems to work.

### DIFF
--- a/desktop-widgets/tab-widgets/TabDiveExtraInfo.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveExtraInfo.cpp
@@ -28,9 +28,9 @@ void TabDiveExtraInfo::updateData(const std::vector<dive *> &, dive *currentDive
 	if (currentDive) {
 		const divecomputer *dc = currentDive->get_dc(currentDC);
 
-		ui->model->setText(QString::fromStdString(dc->model));
-		ui->serial->setText(QString::fromStdString(dc->serial));
-		ui->firmware->setText(QString::fromStdString(dc->fw_version));
+		ui->model->setText(QString::fromStdString(dc->model).trimmed());
+		ui->serial->setText(QString::fromStdString(dc->serial).trimmed());
+		ui->firmware->setText(QString::fromStdString(dc->fw_version).trimmed());
 		ui->date->setText(get_dive_date_string(dc->when));
 		ui->duration->setText(get_dive_duration_string(dc->duration.seconds, tr("h"), tr("min"), tr("sec"),
 				" ", dc->divemode == FREEDIVE));
--- a/desktop-widgets/tab-widgets/TabDiveExtraInfo.cpp
+++ b/desktop-widgets/tab-widgets/TabDiveExtraInfo.cpp
@@ -2,7 +2,9 @@
 #include "TabDiveExtraInfo.h"
 #include "ui_TabDiveExtraInfo.h"
 #include "core/dive.h"
+#include "core/gettextfromc.h"
 #include "core/selection.h"
+#include "core/string-format.h"
 #include "qt-models/divecomputerextradatamodel.h"
 
 TabDiveExtraInfo::TabDiveExtraInfo(MainTab *parent) :
@@ -23,15 +25,36 @@ TabDiveExtraInfo::~TabDiveExtraInfo()
 
 void TabDiveExtraInfo::updateData(const std::vector<dive *> &, dive *currentDive, int currentDC)
 {
-	if (currentDive)
-		extraDataModel->updateDiveComputer(currentDive->get_dc(currentDC));
+	if (currentDive) {
+		const divecomputer *dc = currentDive->get_dc(currentDC);
 
-	ui->extraData->setVisible(false); // This will cause the resize to include rows outside the current viewport
-	ui->extraData->resizeColumnsToContents();
-	ui->extraData->setVisible(true);
+		ui->model->setText(QString::fromStdString(dc->model));
+		ui->serial->setText(QString::fromStdString(dc->serial));
+		ui->firmware->setText(QString::fromStdString(dc->fw_version));
+		ui->date->setText(get_dive_date_string(dc->when));
+		ui->duration->setText(get_dive_duration_string(dc->duration.seconds, tr("h"), tr("min"), tr("sec"),
+				" ", dc->divemode == FREEDIVE));
+		if (dc->divemode >= 0 && dc->divemode < NUM_DIVEMODE)
+			ui->diveMode->setText(gettextFromC::tr(divemode_text_ui[dc->divemode]));
+		else
+			ui->diveMode->clear();
+
+		extraDataModel->updateDiveComputer(dc);
+		ui->extraData->setVisible(false); // This will cause the resize to include rows outside the current viewport
+		ui->extraData->resizeColumnsToContents();
+		ui->extraData->setVisible(true);
+	} else {
+		clear();
+	}
 }
 
 void TabDiveExtraInfo::clear()
 {
+	ui->model->clear();
+	ui->serial->clear();
+	ui->firmware->clear();
+	ui->date->clear();
+	ui->duration->clear();
+	ui->diveMode->clear();
 	extraDataModel->clear();
 }
--- a/desktop-widgets/tab-widgets/TabDiveExtraInfo.ui
+++ b/desktop-widgets/tab-widgets/TabDiveExtraInfo.ui
@@ -10,9 +10,181 @@
     <height>300</height>
    </rect>
   </property>
-  <layout class="QVBoxLayout" name="verticalLayout">
-   <item>
-    <widget class="QTableView" name="extraData"/>
+  <layout class="QGridLayout" name="gridLayout">
+   <item row="0" column="0">
+    <widget class="QGroupBox">
+     <property name="title">
+      <string>Model</string>
+     </property>
+     <property name="alignment">
+      <set>Qt::AlignHCenter</set>
+     </property>
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Preferred" vsizetype="Maximum">
+      </sizepolicy>
+     </property>
+     <layout class="QHBoxLayout">
+      <property name="sizeConstraint">
+       <enum>QLayout::SetMinimumSize</enum>
+      </property>
+      <property name="alignment">
+       <set>Qt::AlignHCenter</set>
+      </property>
+      <item>
+       <widget class="QLabel" name="model" />
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="0" column="1">
+    <widget class="QGroupBox">
+     <property name="title">
+      <string>Serial</string>
+     </property>
+     <property name="alignment">
+      <set>Qt::AlignHCenter</set>
+     </property>
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Preferred" vsizetype="Maximum">
+      </sizepolicy>
+     </property>
+     <layout class="QHBoxLayout">
+      <property name="sizeConstraint">
+       <enum>QLayout::SetMinimumSize</enum>
+      </property>
+      <property name="alignment">
+       <set>Qt::AlignHCenter</set>
+      </property>
+      <item>
+       <widget class="QLabel" name="serial" native="true"/>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="0" column="2">
+    <widget class="QGroupBox">
+     <property name="title">
+      <string>Firmware</string>
+     </property>
+     <property name="alignment">
+      <set>Qt::AlignHCenter</set>
+     </property>
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Preferred" vsizetype="Maximum">
+      </sizepolicy>
+     </property>
+     <layout class="QHBoxLayout">
+      <property name="sizeConstraint">
+       <enum>QLayout::SetMinimumSize</enum>
+      </property>
+      <property name="alignment">
+       <set>Qt::AlignHCenter</set>
+      </property>
+      <item>
+       <widget class="QLabel" name="firmware" native="true"/>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="1" column="0">
+    <widget class="QGroupBox">
+     <property name="title">
+      <string>Date</string>
+     </property>
+     <property name="alignment">
+      <set>Qt::AlignHCenter</set>
+     </property>
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Preferred" vsizetype="Maximum">
+      </sizepolicy>
+     </property>
+     <layout class="QHBoxLayout">
+      <property name="sizeConstraint">
+       <enum>QLayout::SetMinimumSize</enum>
+      </property>
+      <property name="alignment">
+       <set>Qt::AlignHCenter</set>
+      </property>
+      <item>
+       <widget class="QLabel" name="date" />
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="1" column="1">
+    <widget class="QGroupBox">
+     <property name="title">
+      <string>Duration</string>
+     </property>
+     <property name="alignment">
+      <set>Qt::AlignHCenter</set>
+     </property>
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Preferred" vsizetype="Maximum">
+      </sizepolicy>
+     </property>
+     <layout class="QHBoxLayout">
+      <property name="sizeConstraint">
+       <enum>QLayout::SetMinimumSize</enum>
+      </property>
+      <property name="alignment">
+       <set>Qt::AlignHCenter</set>
+      </property>
+      <item>
+       <widget class="QLabel" name="duration" native="true"/>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="1" column="2">
+    <widget class="QGroupBox">
+     <property name="title">
+      <string>Dive Mode</string>
+     </property>
+     <property name="alignment">
+      <set>Qt::AlignHCenter</set>
+     </property>
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Preferred" vsizetype="Maximum">
+      </sizepolicy>
+     </property>
+     <layout class="QHBoxLayout">
+      <property name="sizeConstraint">
+       <enum>QLayout::SetMinimumSize</enum>
+      </property>
+      <property name="alignment">
+       <set>Qt::AlignHCenter</set>
+      </property>
+      <item>
+       <widget class="QLabel" name="diveMode" native="true"/>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="2" column="0" colspan="3">
+    <widget class="QGroupBox">
+     <property name="title">
+      <string>Extra Data</string>
+     </property>
+     <property name="alignment">
+      <set>Qt::AlignHCenter</set>
+     </property>
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Preferred" vsizetype="Maximum">
+      </sizepolicy>
+     </property>
+     <layout class="QHBoxLayout">
+      <property name="sizeConstraint">
+       <enum>QLayout::SetMaximumSize</enum>
+      </property>
+      <property name="alignment">
+       <set>Qt::AlignHCenter</set>
+      </property>
+      <item>
+       <widget class="QTableView" name="extraData"/>
+      </item>
+     </layout>
+    </widget>
    </item>
   </layout>
  </widget>
--- a/desktop-widgets/tab-widgets/TabDiveExtraInfo.ui
+++ b/desktop-widgets/tab-widgets/TabDiveExtraInfo.ui
@@ -10,9 +10,6 @@
     <height>300</height>
    </rect>
   </property>
-  <property name="windowTitle">
-   <string>Dive Computer Info</string>
-  </property>
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
     <widget class="QTableView" name="extraData"/>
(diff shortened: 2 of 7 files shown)


bstoeger:
Alternatively, one could add the date to the `ExtraDataModel` to stay with the tabular format.

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -42,7 +42,7 @@ MainTab::MainTab(QWidget *parent) : QTabWidget(parent),
 	extraWidgets << new TabDivePhotos(this);
 	addTab(extraWidgets.last(), tr("Media"));
 	extraWidgets << new TabDiveExtraInfo(this);
-	addTab(extraWidgets.last(), tr("Extra Info"));
+	addTab(extraWidgets.last(), tr("Dive Compputer Info"));
```
mikeller: **chore:** 
```suggestion
	addTab(extraWidgets.last(), tr("Dive Computer Info"));
```

mikeller (commented):
Looks good to me - I was thinking of doing a similar thing.

I'll rework https://github.com/subsurface/subsurface/pull/4689 to be displayed in the same way once this has been merged.

We'll just need to work out where to store the timezone offset per dive computer.

And then a long time bugbear of mine should be looked into as well - the 'dive mode'. We really should be persisting the dive mode reported by each dive computer as is and NOT allow the user to change this, as doing so will turn the profile information to chaos in almost all cases.
But then we probably need to allow the user to select a dive mode __for the dive__ - and persist this somewhere.

But I am getting ahead of myself, one step at a time.

bstoeger:
> We'll just need to work out where to store the timezone offset per dive computer.

Yeah, since I'm currently experimenting with this, let's find a common solution for the XML and git files.

My suggestion would be an optional "utc_offset" or "offset_to_utc" field with second precision (i.e. the same precision as our regular time stamps).

This is precisely the information we get from libdivecomputer and it absolutely makes sense to me.

If dive computers start providing more detailed timezone information, I would suggest adding an additional free-format string field. Possibly simply in extra_data.

mikeller:
@bstoeger: This change now highlights a problem that we have in our pre-existing code: The handling of dive timestamp vs. divecomputer timestamp is not consistent between local storage and cloud storage:

Reproduction:

- open log
- change dive timestamp
- save log
- open log again

Local storage:

=> dive shows new timestamp, divecomputer shows original timestamp

Cloud storage:

=> dive shows new timestamp, divecomputer ([0] if multiple) shows new timestamp


My hunch is that what local storage does is probably what we want - allow the user to change the timestamp for a dive, but preserve the data that came from the divecomputer.
And we probably want to change the function that persists the divecomputer timestamp to git to store it even for divecomputer[0] if it's not identical to the dive timestamp.

mikeller:
> @bstoeger: This change now highlights a problem that we have in our pre-existing code: The handling of dive timestamp vs. divecomputer timestamp is not consistent between local storage and cloud storage:

Answer to self: Found and fixed the bug in https://github.com/subsurface/subsurface/pull/4699/changes#diff-e7beb11c2f7b4a6b88ea6efd7c016ecd3cc1455ad51369482bf172055cf7d5f4R621

bstoeger:
> @bstoeger: This change now highlights a problem that we have in our pre-existing code: The handling of dive timestamp vs. divecomputer timestamp is not consistent between local storage and cloud storage:

Good catch! I noticed that for local storage we don't write the dive computer time if it is equal to dive tome. But on reading, the dive computer time seems to be filled with dive time if it is not set, so I though that was no problem. I didn't test cloud storage (since I have my own cloud space, where I simply put the XML file).

> My hunch is that what local storage does is probably what we want - allow the user to change the timestamp for a dive, but preserve the data that came from the divecomputer.

Agree.

> Answer to self: Found and fixed the bug in https://github.com/subsurface/subsurface/pull/4699/changes#diff-e7beb11c2f7b4a6b88ea6efd7c016ecd3cc1455ad51369482bf172055cf7d5f4R621

Oh. Ouch. Feel free to merge tihis into master, as it seems to be independent from this PR and a major bug.

bstoeger:
Ok - I tried to merge into master. Usually, when I touch master directly, things go down in flames. Fingers crossed not this time.

bstoeger:
Since there are quite a lot of things in `struct divecomputer` that we might want to display, I think I will copy the UI from the "Information" or "Summary" tabs.

atdotde:
> And then a long time bugbear of mine should be looked into as well - the 'dive mode'. We really should be persisting the dive mode reported by each dive computer as is and NOT allow the user to change this, as doing so will turn the profile information to chaos in almost all cases. But then we probably need to allow the user to select a dive mode **for the dive** - and persist this somewhere.

The problem with the dive mode is that it can change during the dive, typically when bailing out from CCR to OC. So the "dive mode of the the dive" is only the dive mode of the beginning of the dive.

bstoeger:
I tried to copy the UI from the DiveInformation tab.

What do you think? Information density seems a bit low to me.

bstoeger:
> The problem with the dive mode is that it can change during the dive, typically when bailing out from CCR to OC. So the "dive mode of the the dive" is only the dive mode of the beginning of the dive.

Fair point, but I think somewhat orthogonal to @mikeller's remark: We treat most divecomputer provided data as immutable, but not for dive mode. So I agree that the divemode should be stored in the dive for consistency.

Another weird thing is the "manually added" divecomputer. Here we want to store changes in the "computer", don't we.

mikeller:
> The problem with the dive mode is that it can change during the dive, typically when bailing out from CCR to OC. So the "dive mode of the the dive" is only the dive mode of the beginning of the dive.

There are two 'dive modes':
1. the 'global' dive mode for the entire dive. This is normally returned as a header field when downloading dives;
2. the effective dive mode at a specific time in the dive. This is calculated from the global dive mode and the cylinder use for the currently active cylinder: 
https://github.com/subsurface/subsurface/blob/c44222eabb3a67d08fd2b7c221452934660ccaa0/core/dive.cpp#L1733C1-L1740C2

In the current implementation the user can change 1. for any given dive computer, which then in turn will screw over 2. in cases where CCR segments are turned into OC segments, throwing off ppO2 and decompression calculations.

> Fair point, but I think somewhat orthogonal to @mikeller's remark: We treat most divecomputer provided data as immutable, but not for dive mode. So I agree that the divemode should be stored in the dive for consistency.

We allow the user to change quite a few aspects of downloaded dives, like gasmixes, gas changes, or the dive mode. In most cases this will result in a dive log that is inconsistent when it comes to things like the dive computer reported NDL or decompression obligation, and results in regular support requests from users who have done this and now can not make sense of their dive log. So in my opinion it is desirable if we move towards treating the data downloaded from a dive computer as immutable and only 'enrich' it with data attributed to the dive.

> Another weird thing is the "manually added" divecomputer. Here we want to store changes in the "computer", don't we.

Yes, this is another 'interesting' case. Right now we support planned dives and manually added dives.
- planned dives are created in the planner, which is a fully fledged editor for all aspects of the dive, so we are good if we treat them as immutable as downloaded dives in the dive list view;
- manually added dives are really a much simplified version of planned dives that do not need the planner. Not sure what best to do with those, but I wonder if we want to just not store any attributes on the 'manually added' divecomputer, and have it inherit everything from its dive, except for its profile?


> Since there are quite a lot of things in `struct divecomputer` that we might want to display, I think I will copy the UI from the "Information" or "Summary" tabs.

I was thinking the same - I think this is going in the right direction, just needs some more attention to spacing and label contents. And we probably want to filter out the firmware version and serial in 'Extra Info' - no point showing this twice.

image

As for 'dive mode' in the 'Info' tab - I think we should keep this editable by the user, but persist it as an attribute on the dive. The default value can be derived from the dive computer's dive mode, but then we also should allow the user to change it for a dive - for cases like when a bottom timer and tables are used.

desktop-widgets/tab-widgets/maintab.cpp
```diff
@@ -42,7 +42,7 @@ MainTab::MainTab(QWidget *parent) : QTabWidget(parent),
 	extraWidgets << new TabDivePhotos(this);
 	addTab(extraWidgets.last(), tr("Media"));
 	extraWidgets << new TabDiveExtraInfo(this);
-	addTab(extraWidgets.last(), tr("Extra Info"));
+	addTab(extraWidgets.last(), tr("Dive Copputer Info"));
```
mikeller: **suggestion:** 
```suggestion
	addTab(extraWidgets.last(), tr("Dive Computer Info"));
```

desktop-widgets/tab-widgets/TabDiveExtraInfo.ui
```diff
@@ -10,12 +10,182 @@
...
+      <item>
+       <widget class="QLabel" name="duration" native="true"/>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="1" column="2">
+    <widget class="QGroupBox" name="groupBox6">
+     <property name="title">
+      <string>Duration</string>
```
mikeller: **suggestion:** 
```suggestion
      <string>Dive Mode</string>
```

desktop-widgets/tab-widgets/maintab.cpp
bstoeger: I was a few seconds faster. :)

bstoeger:
> I was thinking the same - I think this is going in the right direction, just needs some more attention to spacing and label contents.

I fixed the labels - not sure what is wrong with the alignment of Firmware. I don't see this.

> And we probably want to filter out the firmware version and serial in 'Extra Info' - no point showing this twice.

See added commit.

Probably should add fields for salinity, temperature, etc., right?

bstoeger:
> See added commit.

Never mind - that was a stupid idea - now the firmware and serial isn't saved anymore.

atdotde:
> We allow the user to change quite a few aspects of downloaded dives, like gasmixes, gas changes, or the dive mode. In most cases this will result in a dive log that is inconsistent when it comes to things like the dive computer reported NDL or decompression obligation, and results in regular support requests from users who have done this and now can not make sense of their dive log.

I would argue that at least as often we have the case that the dive computer information is incorrect or incomplete: I still might take an additional  dive computer that cannot handle gas switches on a dive or not bother to tell all my computers that I did a gas switch or changed to OC. During the dive I am aware that the deco information on that computer is not accurate but I don't care. Still, in my log, I want to download the dive from that computer and then manually correct that information by adding gas switches or dive mode changes. I don't think that it is a good idea to treat all dive computer data as god given.

mikeller:
> I fixed the labels - not sure what is wrong with the alignment of Firmware. I don't see this.

Found it - the Heinrichs Weikamp OSTC 3/4/5 parser is adding a newline to the end of the firmware version string. 😬

https://github.com/subsurface/libdc/pull/101

> Probably should add fields for salinity, temperature, etc., right?

We don't need to solve for world peace here - this is a good start and we can add more to it over time.
I'll also want to add the timezone offset that started all of this. 😉

bstoeger:
> Found it - the Heinrichs Weikamp OSTC 3/4/5 parser is adding a newline to the end of the firmware version string. 😬

Ok, good. Still an UI bug.

> We don't need to solve for world peace here - this is a good start and we can add more to it over time.

Ok, up to you. :)

Please check the latest commit, which tries to filter out serial and FW Version.

> I'll also want to add the timezone offset that started all of this.

I would suggest just adding it to the date string: " (+5:00)".

mikeller:
> I would argue that at least as often we have the case that the dive computer information is incorrect or incomplete: I still might take an additional dive computer that cannot handle gas switches on a dive or not bother to tell all my computers that I did a gas switch or changed to OC. During the dive I am aware that the deco information on that computer is not accurate but I don't care. Still, in my log, I want to download the dive from that computer and then manually correct that information by adding gas switches or dive mode changes. I don't think that it is a good idea to treat all dive computer data as god given.

I carry extra computers like this as well. But I think these are rather exotic cases for testing dive computer firmware changes, or getting test dive logs. And in every case these are computers that I carry in addition to the primary and backup computers on which I rely to give me the information that need to survive the dive and not get bent.
So even in these cases I will have at least 2 dive logs that reliably show the actual representation of how the dive was done, and I am not sure there is much point in trying to manipulate any additional logs into something that is closer to the actual dive, but in all likelihood will be inconsistent with respect to things like decostops and TTS.

bstoeger:
> I want to download the dive from that computer and then manually correct that information by adding gas switches or dive mode changes.

Indeed, the profile widget allows lots of manipulation, such as adding and modifying events.

Perhaps profile data should be differently compared to other dive computer data?

desktop-widgets/tab-widgets/TabDiveExtraInfo.cpp
```diff
@@ -23,15 +25,36 @@ TabDiveExtraInfo::~TabDiveExtraInfo()
...
 {
-	if (currentDive)
-		extraDataModel->updateDiveComputer(currentDive->get_dc(currentDC));
+	if (currentDive) {
+		const divecomputer *dc = currentDive->get_dc(currentDC);
 
-	ui->extraData->setVisible(false); // This will cause the resize to include rows outside the current viewport
-	ui->extraData->resizeColumnsToContents();
-	ui->extraData->setVisible(true);
+		ui->model->setText(QString::fromStdString(dc->model));
```
mikeller: **suggestion:** 
```suggestion
		ui->model->setText(QString::fromStdString(dc->model).replace("\n", ""));
```

desktop-widgets/tab-widgets/TabDiveExtraInfo.cpp
```diff
@@ -23,15 +25,36 @@ TabDiveExtraInfo::~TabDiveExtraInfo()
...
+		ui->serial->setText(QString::fromStdString(dc->serial));
```
mikeller: **suggestion:** 
```suggestion
		ui->serial->setText(QString::fromStdString(dc->serial).replace("\n", ""));
```

desktop-widgets/tab-widgets/TabDiveExtraInfo.cpp
```diff
@@ -23,15 +25,36 @@ TabDiveExtraInfo::~TabDiveExtraInfo()
...
+		ui->firmware->setText(QString::fromStdString(dc->fw_version));
```
mikeller: **suggestion:** Prevent these libdivecomputer-provided values from breaking the UI if they contain newlines.
```suggestion
		ui->firmware->setText(QString::fromStdString(dc->fw_version).replace("\n", ""));
```

bstoeger: What about a general "strip()". And how about cleaning this up when filling out "struct divecomputer"?

RonOnGithub:
> > I would argue that at least as often we have the case that the dive computer information is incorrect or incomplete: I still might take an additional dive computer that cannot handle gas switches on a dive or not bother to tell all my computers that I did a gas switch or changed to OC. During the dive I am aware that the deco information on that computer is not accurate but I don't care. Still, in my log, I want to download the dive from that computer and then manually correct that information by adding gas switches or dive mode changes. I don't think that it is a good idea to treat all dive computer data as god given.
> 
> I carry extra computers like this as well. But I think these are rather exotic cases for testing dive computer firmware changes, or getting test dive logs.

I've definitely done this for much more mundane cases, such as fat finger accepting (and then immediately reverting) a prompted gas change that I didn't do, instead of dismissing it - because it was a shallow rec dive, with no chance of ever hitting NDL and I wasn't actually carrying deco gas, I just hadn't told the computer about that.

And it is nice to be able to fix that up in the dive record without hand-hacking the xml.

I definitely think we should preserve dc provided data as accurately as possibly by default, so as not to lose any information that we might later be able to use - but not to the point of making it immutable to a user who knows they get to keep all the pieces if they edit it incorrectly or unwisely.

mikeller:
> And it is nice to be able to fix that up in the dive record without hand-hacking the xml.

You can always do that (and actually change all of it including the profile, and not just some selected aspects) by opening the dive in the dive planner.

> I definitely think we should preserve dc provided data as accurately as possibly by default, so as not to lose any information that we might later be able to use - but not to the point of making it immutable to a user who knows they get to keep all the pieces if they edit it incorrectly or unwisely.

I think 'the user knows' is the key point here. While in your example, an incorrect gas for a minute or so will not have made a significant difference to the NDL / TTS logged by the dive computer, many users don't seem to be aware of the fact that this will render the dive computer supplied information shown in Subsurface completely inconsistent if they do it for more than tiny deviations from the information that was given to the dive computer during the dive.

RonOnGithub:
> > I definitely think we should preserve dc provided data as accurately as possibly by default, so as not to lose any information that we might later be able to use - but not to the point of making it immutable to a user who knows they get to keep all the pieces if they edit it incorrectly or unwisely.
> 
> I think 'the user knows' is the key point here.

Ok, I missed the point where you'd had plenty of people footgun themselves then complain.
I'm ok with the raw data being something immutable that you can always 'undo' to, so long as there is an overriding overlay that you can edit to make the normally visible profile and log look as sensible as possible for whatever differences there are between Reality and the computer's version of events.

bstoeger:
So any objections to this PR?

Fixing the newline is probably best done in a separate PR, possibly in conjunction with fixing libdc.

mikeller:
> So any objections to this PR?
> 
> Fixing the newline is probably best done in a separate PR, possibly in conjunction with fixing libdc.

None from me - but I am keen to get #4708 merged first to test if this already fixes the UI alignment problems. And that one is giving me problems at the moment.

bstoeger:
> None from me - but I am keen to get #4708 merged first to test if this already fixes the UI alignment problems. And that one is giving me problems at the moment.

Ok, thanks. I will try to rewrite #4708  without usage of `std::string_view`. That's the only explanation I have.

mikeller:
> > None from me - but I am keen to get #4708 merged first to test if this already fixes the UI alignment problems. And that one is giving me problems at the moment.
> 
> Ok, thanks. I will try to rewrite #4708 without usage of `std::string_view`. That's the only explanation I have.

Hold off for a couple of minutes - I had one last idea (corrupted cache). And I'd like to put this to @dirkhh as well - to me it does not make sense that the use of `std::string_view` should cause a build failure only in one particular build.

dirkhh:
We have seen issues where builds failed on only one particular compiler version 
I have no idea what's going on here

qt-models/divecomputerextradatamodel.cpp
```diff
@@ -46,10 +46,18 @@ int ExtraDataModel::rowCount(const QModelIndex&) const
...
+				new_items.push_back(data);
+		}
+	}
+
 	beginResetModel();
-	if (dc)
-		items = dc->extra_data;
-	else
-		items.clear();
+	items = std::move(new_items);
```
dirkhh, line 61: and again we replace perfectly readable code with gobbledigook.

dirkhh (commented):
I'm not convinced that renaming this tab serves any purpose beyond confusing people and requiring changes to the user manual.

I'm equally not convinced that showing the dive start time as envisioned by the dive computer is particularly enlightening to anyone, but hey, that's at least something we can argue about.

qt-models/divecomputerextradatamodel.cpp
RonOnGithub: > and again we replace perfectly readable code with gobbledigook.

Actually this one is just a blistering symptom of a much more chronic disease. and the real red flag here is not the quoted 'gobbledigook', but the trainwreck horror that you didn't quote above it.

The smallpox blanket that is really responsible for this is commit bc761344d4699a0ee33255d49440760196266e18
which took a perfectly functional linked list and 'simplified' it into an entirely inappropriate container for the job as this change so aptly demonstrates.

There's a lot of this sort of stuff that's making getting my patchset ready to land a lot harder than it ought to be.

So I agree this is another of those things that should be better planned *before* being implemented rather than being bandaged afterward.

bstoeger:
> and again we replace perfectly readable code with gobbledigook.

Show me how to write this in a "prefectly readable" way, please.

bstoeger:
> The smallpox blanket that is really responsible for this is commit https://github.com/subsurface/subsurface/commit/bc761344d4699a0ee33255d49440760196266e18
which took a perfectly functional linked list and 'simplified' it into an entirely inappropriate container for the job as this change so aptly demonstrates.

WTF? In a linked list it would have been easier to filter out certain values? Are you drunk?

RonOnGithub:
> Are you drunk?

No.  But I'm pretty sure I still would have known a vector was the wrong tool for this job even if I was.

RonOnGithub:
> Show me how to write this in a "prefectly readable" way, please.

Ok.  But let me get the other more important fixes finished first.

mikeller:
> I'm not convinced that renaming this tab serves any purpose beyond confusing people and requiring changes to the user manual.
> 
> I'm equally not convinced that showing the dive start time as envisioned by the dive computer is particularly enlightening to anyone, but hey, that's at least something we can argue about.

I think the value of this change for users depends on how they are using Subsurface:
- for recreational users diving and logging with one dive computer it will make little to no difference
- for technical users diving and logging with multiple dive computers, being able to see which dive computer reported what exactly is valuable

Re the dive computer reported time of day, just speaking from personal experience: Having dives logged with an incorrect time of day, caused by travel across time zones without reconfiguring the dive computer before the first dive seems to be by far the most common type of dive computer misconfiguration that I get, often leaving me with a game of 'guess the hour' after a dive trip. Having all of the original information even after attempts at reconciliation seems to be a good thing to me.

dirkhh:
I defer to you on this @mikeller

PR #4702, opened 2026-02-02
cleanup: move (un)lock_planner() to profile.cpp

bstoeger:
The only callers (un)lock_planner() are in profile.cpp, so we might just move it there and shrink qthelper.cpp.

This helper function was needed, when planner.c was pure C. Not any longer.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

A rather trivial cleanup and part of the long-term project "shrink qthelper.cpp".

### DIFF
--- a/core/profile.cpp
+++ b/core/profile.cpp
@@ -3,10 +3,6 @@
 /* creates all the necessary data for drawing the dive profile
  */
 #include "gettext.h"
-#include <limits.h>
-#include <string.h>
-#include <assert.h>
-#include <stdlib.h>
 
 #include "dive.h"
 #include "divelist.h"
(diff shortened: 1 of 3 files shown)


mikeller (approved):

bstoeger:
Given that this is the only caller (last one was removed in 9f55f167b2eca69a30a6e9c3db0f627ce3553fc2), I wonder if this lock is necessary at all. What does it protect?

atdotde:
IIRC it is there to prevent the planner to be called several times while the UI updates.

bstoeger:
Ok, thanks. Let's merge this now and when time permits do a closer analysis.

PR #4703, opened 2026-02-02
cleanup: move print_qt_version() into subsurfacestartup.cpp

bstoeger:
This function was in qthelper.cpp from a time when the startup code was pure C.

Just fold it into the only caller. Also, there seems to be no reason to use convoluted Qt constructs, since the version information comes as C-style strings.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Another trivial "make qthelper.cpp smaller" commit. Since the startup code is now C++, there is no reason to access the version information in an external file.

### DIFF
--- a/core/qthelper.cpp
+++ b/core/qthelper.cpp
@@ -1546,11 +1546,6 @@ void parse_seabear_header(const char *filename, struct xml_params *params)
 	f.close();
 }
 
-void print_qt_versions()
-{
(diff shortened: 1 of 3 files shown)


mikeller (approved):

bstoeger:
Thanks. I think we can commit this. If @dirkhh objects, he can revert.

PR #4705, opened 2026-02-04
Fix Firmware Version Detection for Garmin Dive Computers.

mikeller:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Properly recognise the Extra Info field for the firmware version that is
returned by Garmin dive computers.

### Additional information:

Doing it this way instead of changing the libdivecomputer backend means
it will work for already imported dive logs as well.

### DIFF
--- a/core/divecomputer.cpp
+++ b/core/divecomputer.cpp
@@ -368,12 +368,12 @@ struct event *get_event(struct divecomputer *dc, int idx)
 
 void add_extra_data(struct divecomputer *dc, const std::string &key, const std::string &value)
 {
-	if (key == "Serial") {
+	if (key == STRING_KEY_SERIAL_NUMBER) {
 		dc->deviceid = calculate_string_hash(value.c_str());
 		dc->serial = value;
 	}
-	// Support dives imported with the legacy Garmin backend
-	if (key == STRING_KEY_FIRMWARE_VERSION || key == "Firmware")
+	// Include support for dives imported with the legacy Garmin / Uwatec backends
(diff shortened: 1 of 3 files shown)


bstoeger (approved):

bstoeger:
Is there a reason libdivecomputer exports some firmware information as "FW Version" and some as "Firmware" extra data?

@jefdriesen

dirkhh:
anything that's a string is a Subsurface extension

anything that's a random encoding of what is a text (the firmware version that the user sees) as a number, is an incarnation of the libdivecomputer way of doing things in very interesting ways because `${REASON}`

We added the string to some, but IIRC not all backends in our fork.

bstoeger:
So would it make sense to unify this to either "FW Version" or "Firmware"?

mikeller:
> So would it make sense to unify this to either "FW Version" or "Firmware"?

This will be hard to do for dives that the user has already downloaded - which is why I went for this approach to fixing it.

But we could standardise the libdivecomputer backends as well. And move constants for the names of extra info fields that have a special meaning, and use them in Subsurface to read these fields as well.

bstoeger:
> But we could standardise the libdivecomputer backends as well.

Yes, that sounds like a good idea.

mikeller:
> Yes, that sounds like a good idea.

https://github.com/subsurface/libdc/pull/102

RonOnGithub:
As I noted in the other thread, there's also (at least) "DC Firmware Version" used in core/import-logtrak.cpp
and I haven't looked hard to see if there are other permutations used elsewhere too.

PR #4708, opened 2026-02-05
Remove leading/trailing whitespace from extra_data values

bstoeger:

- [ ] Bug fix
- [x] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Since some libdivecomputer devices seem to add spurious whitespace to the firmware version, remove it.

See discussion in #4699.

@mikeller: What do you think of this "brutal" solution to the whitespace problem? Does one ever want to keep the whitespece at the beginning or end of the extra data value?

### DIFF
--- a/core/device.cpp
+++ b/core/device.cpp
@@ -73,7 +73,7 @@ int create_device_node(device_table &dcs, const std::string &m, const std::strin
 		it->nickName = n;
 		return it - dcs.begin();
 	} else {
-		dev.deviceId = calculate_string_hash(s.c_str());
+		dev.deviceId = calculate_string_hash(s);
 		it = dcs.insert(it, dev);
 		return it - dcs.begin();
 	}
--- a/core/divecomputer.cpp
+++ b/core/divecomputer.cpp
@@ -369,7 +369,7 @@ struct event *get_event(struct divecomputer *dc, int idx)
 void add_extra_data(struct divecomputer *dc, const std::string &key, const std::string &value)
 {
 	if (key == "Serial") {
-		dc->deviceid = calculate_string_hash(value.c_str());
+		dc->deviceid = calculate_string_hash(value);
 		dc->serial = value;
 	}
 	if (key == "FW Version")
--- a/core/divecomputer.cpp
+++ b/core/divecomputer.cpp
@@ -366,16 +366,20 @@ struct event *get_event(struct divecomputer *dc, int idx)
 	return &dc->events[idx];
 }
 
-void add_extra_data(struct divecomputer *dc, const std::string &key, const std::string &value)
+void add_extra_data(struct divecomputer *dc, const std::string &key, std::string_view value)
 {
 	if (key == "Serial") {
 		dc->deviceid = calculate_string_hash(value);
 		dc->serial = value;
 	}
-	if (key == "FW Version")
-		dc->fw_version = value;
 
-	dc->extra_data.push_back(extra_data { key, value });
+	// Some libdivecomputer backends add a spurious '\n' character at the
+	// end of FW Version. Remove this for now. Note that, for now,
+	// the sanitized string will not be written to the log.
+	if (key == "FW Version")
+		dc->fw_version = trimmed(value);
+
+	dc->extra_data.push_back(extra_data { key, std::string(value) });
 }
 
 /*
--- a/core/divecomputer.h
+++ b/core/divecomputer.h
@@ -67,7 +67,7 @@ extern int add_event_to_dc(struct divecomputer *dc, struct event ev); // event s
 extern struct event *add_event(struct divecomputer *dc, unsigned int time, int type, int flags, int value, const std::string &name);
 extern struct event remove_event_from_dc(struct divecomputer *dc, int idx);
 struct event *get_event(struct divecomputer *dc, int idx);
-extern void add_extra_data(struct divecomputer *dc, const std::string &key, const std::string &value);
+extern void add_extra_data(struct divecomputer *dc, const std::string &key, std::string_view value);
 extern uint32_t calculate_string_hash(std::string_view sv);
 extern bool is_dc_planner(const struct divecomputer *dc);
 extern void make_planner_dc(struct divecomputer *dc);
--- a/core/divecomputer.h
+++ b/core/divecomputer.h
@@ -6,6 +6,7 @@
 #include "units.h"
 #include <set>
 #include <string>
+#include <string_view>
 #include <vector>
 
 struct extra_data;
@@ -67,7 +68,7 @@ extern struct event *add_event(struct divecomputer *dc, unsigned int time, int t
 extern struct event remove_event_from_dc(struct divecomputer *dc, int idx);
 struct event *get_event(struct divecomputer *dc, int idx);
 extern void add_extra_data(struct divecomputer *dc, const std::string &key, const std::string &value);
-extern uint32_t calculate_string_hash(const char *str);
+extern uint32_t calculate_string_hash(std::string_view sv);
 extern bool is_dc_planner(const struct divecomputer *dc);
 extern void make_planner_dc(struct divecomputer *dc);
 extern const char *manual_dc_name;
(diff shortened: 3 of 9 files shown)


mikeller:
> @mikeller: What do you think of this "brutal" solution to the whitespace problem? Does one ever want to keep the whitespece at the beginning or end of the extra data value?

I can think of one situation where one wants to keep it - if the value in question is used in the calculation of the fingerprint used to check if a given dive has been downloaded already. If we accidentally modify this then affected users will be forced to download all of the dives from their dive computer again, which is undesirable.

Also note that this will go and modify potentially 100s of dives that a user has stored in the cloud or a local file - not sure if this is desirable either, or if it will be better to deal with this at the time that the data is displayed - like the QTableView currently showing the Extra Data already does.

bstoeger:
> I can think of one situation where one wants to keep it - if the value in question is used in the calculation of the fingerprint used to check if a given dive has been downloaded already. If we accidentally modify this then affected users will be forced to download all of the dives from their dive computer again, which is undesirable.

OMG. Please tell me that we don't use trailing white space in the calculation of fingerprints/hashes!

> Also note that this will go and modify potentially 100s of dives that a user has stored in the cloud or a local file - not sure if this is desirable either, or if it will be better to deal with this at the time that the data is displayed - like the QTableView currently showing the Extra Data already does.

If it's not supposed to be there - I personally have no problem with removing it. Seems preferable over having different versions of the firmware string in the same file.

bstoeger:
Now this leaves the extra_data values untouched, but removes it from the values that are displayed.

mikeller:
> Now this leaves the extra_data values untouched, but removes it from the values that are displayed.

Not actually sure of this - from what I can tell, `add_extra_data()` is called when loading from git or XML, so the values will be modified then, and on the next 'Save' the modified values will be written back.

bstoeger:
> Not actually sure of this - from what I can tell, add_extra_data() is called when loading from git or XML, so the values will be modified then, and on the next 'Save' the modified values will be written back.

Yeah, but observe that `extra_data` is not changed, only `divecomputer::serial` and `divecomputer::fw_version`. I don't think these are written to the dive log. They seem to be used only for display. The `deviceid` calculated from the serial number is not changed.

One potential problem might be the device list. My gut feeling tells me to also sanitize the serial number there. It's insane to potentially have distinct devices that differ by trailing whitespace in the serial number.

From a quick glance, the code seems to use `divecomputer::deviceid` and `device::deviceId` for matching devices. Unfortunately the comments concerning this are  contradictory:

We find:

> uint32_t deviceId;      // Always the string hash of the serialNumber

and

> * NOTE! The deviceid and diveid are model-specific *hashes* of
> * whatever device identification that model may have. Different
> * dive computers will have different identifying data, it could
> * be a firmware number or a serial ID (in either string or in
> * numeric format), and we do not care.

This is a royal mess!

If `deviceid` is really only a hash of `serial`, we probably should get rid of deviceid altogether and systematically remove trailing whitespace from the serial numbers, in struct device and in struct divecomputer. Yes, the logs will change, but so what?

bstoeger:
All that said, we only ever observed trailing whitespace for `fw_version`, which is not used for anything, correct? So the whole discussion may be moot.

Still, that `deviceid`/`serial` code will have to be straightened at one point.

bstoeger:
OK,  `get_or_add_device_for_dc()` uses `dc->serial()` to calculate `deviceId` in the same way that `dc->deviceid` was calculated.

I'm more and more convinced, that `divecomputer::deviceid` and `device::deviceId` should just be removed.

In any case, for now, let's not mess with `serial`.

mikeller (approved):

bstoeger:
Heh... What is going on with this PR? I got a bazillion of mails. Could it be that the Mac build doesn't like my trim() function? I don't get it.

bstoeger:
Ok, I will try to do both commits as separate PRs to see which one breaks.

mikeller:
@bstoeger: I think it's not you, it's our build setup for macOS - are you ok if we un-draft this pull request and I try to get this fixed?

bstoeger:
> are you ok if we un-draft this pull request and I try to get this fixed?

Yes, sure.

dirkhh (changes requested):
Can we stop inflicting more std::c++ -isms on the code base for a while.
There is absolutely no need to convert strings from C string to QString to std::string and back and sideways. The code is overflowing with this to the point that it has become a maintenance nightmare. Especially since a trim() function really doesn't require std::anything -- and I'm not even sure why we need this, in the first place

RonOnGithub:
> Can we stop inflicting more std::c++ -isms on the code base for a while. There is absolutely no need to convert strings from C string to QString to std::string and back and sideways. The code is overflowing with this to the point that it has become a maintenance nightmare. Especially since a trim() function really doesn't require std::anything -- and I'm not even sure why we need this, in the first place

I'm ok with converting things to a C++ construct where it takes a dozen lines of code and turns it into a clear one-liner.

But when a 'conversion' takes a clear one-liner and turns it into dozens of lines of extra and more verbose code, and a new function that just reimplements already existing tools that we already have and use, and a new header include in all the places that use it - then someone really needs to ask for a sane explanation of exactly what is being 'improved' here.

Don't get me wrong, I hate QString (and much of the rest of Qt) as much as anyone possibly could - but when you wildly complicate the code, for no other reason than to take something that started as a QString, and is only used by things which require it to be a QString, and convert it back and forth for no other reason than to store it in a local variable that isn't a QString, which nothing other than the to and from QString conversions touch.  Then it's probably not unfair to say that sight of the big picture has pretty seriously been lost.

And I've been having to wade through far too much of that sort of thing too.  And it's doubly unfortunate because in some cases real issues were being addressed - but instead of just fixing that issue, the changes overstepped and introduced entirely new issues all of their own.

bstoeger:
> There is absolutely no need to convert strings from C string to QString to std::string and back and sideways.

Have you actually looked at the commits? The first one _simplifies_ things by using a `std::string_view`, so one can pass either a std::string _or_ a C-style string. So for std::string, there is one less strlen() and less c_str() calls. For C-style strings, _nothing_ changes.

Show me where the first commit makes anything more complex, please.

The second one adds funcitonality.

> The code is overflowing with this to the point that it has become a maintenance nightmare.

The maintenance nightmare started when you realized that C-style strings are completely unsuitable for any kind of string manipulation and added random stuff to various places ending in a complete mess of UTF8 <-> UTF16 conversions with fixed buffer lengths and memory leaks.

> Especially since a trim() function really doesn't require std::anything 

Of course it doesn't. It can also be written it in assembly. What kind of argument is that? The trimmed() function I wrote has the advantage of taking C-style strings _and_  std::string and returns a std::string, where the user doesn't have to do manual memory management.

The whole point of such a function is hiding implementation details. If it is a problem for you, show me hoe to write a better one, please. Observe that I used C-style code flow on purpose, instead of iterators.

I'm deeply convinced that QString (and other Qt primitives) should stay out of core code (for portability reasons, but also because the Qt developers are clearly incompetent). If we already have a trim() function, operating on std::string, please show it to me. Otherwise I consider it a useful addition to the code base.

> and I'm not even sure why we need this, in the first place

It fixes a display bug on the redone "extra infos" tab, where libdc added spurious newlines to the firmware version.

RonOnGithub:
> I'm deeply convinced that QString (and other Qt primitives) should stay out of core code (for portability reasons, but also because the Qt developers are clearly incompetent).

And the evidence is that you are very terribly wrong about at least the former.  This code is up to its neck in Qt, even the non-gui parts of it, and pointlessly converting things to and from its space helps nothing and nobody.  It just makes the code more convoluted, less readable and leaves more places for subtle portability bugs to hide.

What non-Qt thing are you actually planning on using the 'core code' with which makes this a problem?
What 'portability problem' did any of those changes fix?

This is not the code to wage a doomed holy war in.

If you have some realistic replacement for Qt, that would let us *completely* eliminate it, then I'm all ears.  But if you don't, then reimplementing it badly with 'subsurface-specific' replacements that we now have to maintain and that new contributors have to learn the quirks of is not an Improvement.  In any sense of that word.

mikeller:
Not needed any longer - trimming with QString.trimmed() was added directly to #4699.

PR #4709, opened 2026-02-07
cleanup: use std::string's sorting function in divepicturemodel.cpp

bstoeger:
This one is ironic: In the code of DivePictureModel, strcmp() was used to compare two std::strings with a comment that this is due to using the same sorting as in core. However, core has long been changed to use the std::string's sorting function.

Use the idiomatic version also in the model.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Trivial (hopefully) cleanup. See commit description.

### DIFF
--- a/core/picture.cpp
+++ b/core/picture.cpp
@@ -25,7 +25,6 @@ void add_picture(picture_table &t, struct picture newpic)
 
 int get_picture_idx(const picture_table &t, const std::string &filename)
 {
-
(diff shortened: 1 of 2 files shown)


mikeller (approved):
:-) Good find!

PR #4712, opened 2026-02-08
TEST - please ignore!

bstoeger:
A pointless wihtespace change to test the build system.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

### DIFF
--- a/core/divelogexportlogic.h
+++ b/core/divelogexportlogic.h
@@ -17,4 +17,3 @@ struct htmlExportSetting {
 void exportHtmlInitLogic(const QString &filename, struct htmlExportSetting &hes);
 
 #endif // DIVELOGEXPORTLOGIC_H
-
--- a/core/exif.cpp
+++ b/core/exif.cpp
@@ -899,4 +899,3 @@ time_t easyexif::EXIFInfo::epoch()
 	tm.tm_sec = sec;
 	return utc_mktime(&tm);
 }
-
--- a/core/settings/qPrefEquipment.h
+++ b/core/settings/qPrefEquipment.h
@@ -42,4 +42,3 @@ private:
 };
 
 #endif
-
--- a/core/triptable.h
+++ b/core/triptable.h
@@ -22,4 +22,3 @@ extern void dump_trip_list();
 Q_DECLARE_METATYPE(trip_table *);
 
 #endif
-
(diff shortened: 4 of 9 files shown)


bstoeger:
@dirkhh @mikeller 

This is a pure whitespace change and still the Mac builds fail. Ominously, #4714 does not fail.

So either the build system has a problem with me personally or something is broken in master.

RonOnGithub:
> @dirkhh @mikeller
> 
> This is a pure whitespace change and still the Mac builds fail. Ominously, #4714 does not fail.
> 
> So either the build system has a problem with me personally or something is broken in master.

Unless I'm reading this wrong, the log says it's choking on busted deps:

  Found libraries with problematic absolute path dependencies:
    QtWebKit -> /usr/local/opt/jpeg/lib/libjpeg.9.dylib

et al.

What's responsible for providing Qt &co. to this?

dirkhh:
> Unless I'm reading this wrong, the log says it's choking on busted deps:
> 
> Found libraries with problematic absolute path dependencies: QtWebKit -> /usr/local/opt/jpeg/lib/libjpeg.9.dylib
> 
> et al.
> 
> What's responsible for providing Qt &co. to this?

We have our own private Qt build. And that jpeg library should pick up one that's already installed on the system, so potentially a rebuild of the Qt build would fix this?

RonOnGithub:
> > What's responsible for providing Qt &co. to this?
> 
> We have our own private Qt build. And that jpeg library should pick up one that's already installed on the system, so potentially a rebuild of the Qt build would fix this?

My last Apple was an Apple][, so I'm only vicariously familiar with Mac stuff - but I think I'd be looking for what in that chain of deps *didn't* find the system jpeg library, and did that thing all the cool kids seem to think is cool these days and downloaded its own local copy from some random place on the internet.

I have a fuzzy memory of spotting an error like that in the kirigami or KF CI build a couple of months back when someone reported it being Unexplainably Busted too.  Details of that are probably in scrollback of a Modern "secure" and unsearchable chatroom though. [ed: ah, no that one was freetype doing this.  but same smell - no .pc found so a dep assumed it was uninstalled, downloaded its own copy, and that poisoned the rest of the build].

Should be evident somewhere in the build logs for that though.

PR #4713, opened 2026-02-08
cleanup: remove unnecessary 'this' capture from lambdas

bstoeger:
Fixes a valid warning.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Trivial cleanup to test the Mac build.

### DIFF
--- a/desktop-widgets/diveplanner.cpp
+++ b/desktop-widgets/diveplanner.cpp
@@ -407,9 +407,9 @@ PlannerSettingsWidget::PlannerSettingsWidget(PlannerWidgets *parent)
 	connect(ui.min_switch_duration, QOverload<int>::of(&QSpinBox::valueChanged), &PlannerShared::set_min_switch_duration);
 	connect(ui.surface_segment, QOverload<int>::of(&QSpinBox::valueChanged), &PlannerShared::set_surface_segment);
 
-	connect(ui.recreational_deco, &QAbstractButton::clicked, [this, parent] { parent->disableDecoElements(RECREATIONAL, parent->getDiveMode()); });
-	connect(ui.buehlmann_deco, &QAbstractButton::clicked, [this, parent] { parent->disableDecoElements(BUEHLMANN, parent->getDiveMode()); });
-	connect(ui.vpmb_deco, &QAbstractButton::clicked, [this, parent] { parent->disableDecoElements(VPMB, parent->getDiveMode()); });
+	connect(ui.recreational_deco, &QAbstractButton::clicked, [parent] { parent->disableDecoElements(RECREATIONAL, parent->getDiveMode()); });
+	connect(ui.buehlmann_deco, &QAbstractButton::clicked, [parent] { parent->disableDecoElements(BUEHLMANN, parent->getDiveMode()); });
+	connect(ui.vpmb_deco, &QAbstractButton::clicked, [parent] { parent->disableDecoElements(VPMB, parent->getDiveMode()); });
 
 	connect(ui.sacfactor, QOverload<double>::of(&QDoubleSpinBox::valueChanged), &PlannerShared::set_sacfactor);
 	connect(ui.problemsolvingtime, QOverload<int>::of(&QSpinBox::valueChanged), plannerModel, &DivePlannerPointsModel::setProblemSolvingTime);


bstoeger:
@dirkhh: Seems like Mac doesn't like me. This commit should be pretty innocuous.

mikeller (approved):

dirkhh (approved):
Yes, that's reasonable. I mean - I don't understand the C++ syntax, but it doesn't seem to break things. And the Mac build failure is well documented / understood (but frustrating, as those are our 'main' macOS binaries, still)

bstoeger:
>  I don't understand the C++ syntax

The capture lists the outside variables the lambda can access. In C you would pass an argument "extra_data" to the function that contains the captured variables. For each variable you can state whether it is passed as copy or as reference. A "lambda" is a pointer to a function and (optionally) a struct of captured variables. Obviously, if you pass a lambda to a different thread, you usually don't want to capture by reference, unless you can guarantee that the life-time of the referenced variables is longer than the lifetime of the thread, etcpp.

If you just write [=], all outside variables that are accessed from the lambda are automatically captured by copy. Same with [&], but by reference.

RonOnGithub:
> I mean - I don't understand the C++ syntax, but it doesn't seem to break things.

I'll sign off on this one as being correct for the code as it currently stands.
*Assuming* that it really wants to call the parent disableDecoElements() and not
the implementation this class provides.

But I assume it's *that* change (doing it indirectly via the parent) which led to the
capture of `this` becoming vestigial.

mikeller:
Now I am really surprised and puzzled that this merge did NOT break the macOS build for `master`.

dirkhh:
That is. Umm. What? Is the version 9 of libjpeg back? Maybe they updated the image they use for x86 macos?

That makes no sense. I had that on my list of things to look at this weekend -- not that I'm complaining...

RonOnGithub:
Did someone look at the logs for the Qt build to see how QtWebKit was ending up with a local build?
(I'm not sure where to find them or where all that happens here).

There could be a race where an implicit dependency does or does not get pulled in depending on the order things finish in if that's not strictly and tightly defined?

mikeller:
> That is. Umm. What? Is the version 9 of libjpeg back? Maybe they updated the image they use for x86 macos?
> 
> That makes no sense. I had that on my list of things to look at this weekend -- not that I'm complaining...

Unfortunately the other builds still seem to be failing:

https://github.com/subsurface/subsurface/actions/runs/21964657519/job/63451227244?pr=4708

Can it be that this is this flakey?

RonOnGithub:
> Can it be that this is this flakey?

Unless I'm missing something (and I probably am, because just finding build logs on this thing is currently a nightmare task to me) - it's just the actual error *capture* that failed for this one which made it "succeed".

Looking at https://github.com/subsurface/subsurface/actions/runs/21947742499/job/63389819673

Exactly the same error is reported as before in the build log:
```
Found libraries with problematic absolute path dependencies:
  QtWebKit -> /usr/local/opt/jpeg/lib/libjpeg.9.dylib
```

The only difference is the CI reported the job succeeded instead of saying it failed.
Is something different about building for master, or is catching the error code successfully itself racy?

RonOnGithub:
Oh!  I see what's happening.  I think.

It's not actually failing on the failed *build* like it should.
In the cases where it's failing, it's failing in the *pull request* check which notices required artifacts are missing.

But since pushing to master isn't a pull request, that check never happens, and everything pretends to be fine.

I guess we need CI to test the CI :D

ed: which would seem to be confirmed by the lack of the qt5 dmg here:
https://github.com/subsurface/nightly-builds/releases/tag/v6.0.5540-CICD-release

mikeller:
> It's not actually failing on the failed _build_ like it should. In the cases where it's failing, it's failing in the _pull request_ check which notices required artifacts are missing.
> 
> But since pushing to master isn't a pull request, that check never happens, and everything pretends to be fine.

Ah yes, good find. It IS failing, but since our MacOS build fails to fail properly, and the step to publish the DMG does not care if the file does not exist, master releases are incomplete and we don't know about it.

RonOnGithub:
Yeah, so we're back to 'just' figuring out the cause for the errors in the build log, (and making the build script actually fail when it catches them).

dirkhh:
Ok, so I'll add that back to my list for this weekend. 
Fun

RonOnGithub:
I can't test stuff directly on Mac, but feel free to point at the relevant build logs here if you want an extra set of eyes to look them over.

dirkhh:
I think I fixed this in our qt-mac package. At least the build appears to have worked, and when I download the artifact, it appears to work on my Mac. I will merge #4718 to trigger another build of master to see if that also creates working artifacts.

mikeller:
@dirkhh This seems to have worked, thank you, the Intel / Qt5 .dmg is attached to the latest release.

In my attempts to fix this I was also getting to the realisation that a good fix for it will be to fix the 'base image' in `qt-mac` - as much as this isn't a great option.

PR #4715, opened 2026-02-09
Desktop: Fix Crash when Opening Preferences.

mikeller:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix a crash when opening the Preferences dialog after scrolling to the
end of the dive list with the Down key and scrolling one past the end.

### DIFF
--- a/qt-models/diveplannermodel.cpp
+++ b/qt-models/diveplannermodel.cpp
@@ -38,6 +38,12 @@ static double unit_factor()
 
 static constexpr int decotimestep = 60; // seconds
 
+static bool cylinder_index_valid(const struct dive *d, int cylinderId)
+{
+	// Allow for a cylinder one beyond the limit of the array to allow for the implied 'surface air' cylinder
+	return d && cylinderId >= 0 && static_cast<size_t>(cylinderId) <= d->cylinders.size();
+}
+
 CylindersModel *DivePlannerPointsModel::cylindersModel()
 {
 	return &cylinders;
@@ -246,6 +252,12 @@ void DivePlannerPointsModel::setPlanMode(Mode m)
 	mode = m;
 	// the planner may reset our GF settings that are used to show deco
 	// reset them to what's in the preferences
+	if (m == NOTHING) {
+		clear();
+		diveplan.dp.clear();
+		d = nullptr;
+		dcNr = 0;
+	}
 	if (m != PLAN) {
 		set_gf(prefs.gflow, prefs.gfhigh);
 		set_vpmb_conservatism(prefs.vpmb_conservatism);
@@ -272,7 +284,12 @@ static divemode_t get_local_divemode(struct dive *d, int dcNr, int cylinderId, d
 
 		break;
 	case CCR:
-		divemode = d->get_cylinder(cylinderId)->cylinder_use == DILUENT ? CCR : OC;
+		if (cylinder_index_valid(d, cylinderId)) {
+			cylinder_t *cyl = d->get_cylinder(cylinderId);
+			divemode = (cyl && cyl->cylinder_use == DILUENT) ? CCR : OC;
+		} else {
+			divemode = OC;
+		}
 		if (divemode == OC && prefs.allowOcGasAsDiluent && selectedDivemode == CCR)
 			divemode = CCR;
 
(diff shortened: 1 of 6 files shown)


dirkhh:
What a fascinating PR.
It mixes things which makes it harder to reason about it. And it almost seems like the thing in the title has become a side show.
I like the Claude(? -- I am very familiar with that particular style and emoji use...) generated code that tries to untangle what's going on in `make-package.sh` ...

I think it would be preferable to have those two things be separate - but since I resisted that same request with my PR, I'm clearly being a hypocrite 😁  -- so please ignore this.

The missing libjpeg.9.dylib ... I assume that is somehow linked into the Qt binaries that we import? I haven't had the time to unpack those (as I don't use them anymore here locally, now that they are no longer the binaries that I used to manually create for that purpose). I wonder if all it took to fix this would be to kick off a new build of those binaries on the same base image as we use to then build the artifacts?

mikeller:
> What a fascinating PR. It mixes things which makes it harder to reason about it. And it almost seems like the thing in the title has become a side show. I like the Claude(? -- I am very familiar with that particular style and emoji use...) generated code that tries to untangle what's going on in `make-package.sh` ...

Yeah, I couldn't make sense of the Claude - generated status messages in our build log, so i figured maybe Claude can. 😉

> I think it would be preferable to have those two things be separate - but since I resisted that same request with my PR, I'm clearly being a hypocrite 😁 -- so please ignore this.

Definitely - the plan is to rebase the commits fixing the macOS build (and making it fail if it can't produce a DMG). But I needed to start somewhere that is broken, and I was not going to break `master` doing this.

> The missing libjpeg.9.dylib ... I assume that is somehow linked into the Qt binaries that we import? I haven't had the time to unpack those (as I don't use them anymore here locally, now that they are no longer the binaries that I used to manually create for that purpose).

I think so - I suspect the trigger might have been homebrew updating libjpeg from 8 to 9

> I wonder if all it took to fix this would be to kick off a new build of those binaries on the same base image as we use to then build the artifacts?

Not sure how you mean this - from my understanding of the process, the macOS / Qt5 / Intel build is the only one still using the 'base image' in https://github.com/subsurface/qt-mac, and the Qt6 builds have moved to a cached on-the-fly install of Qt through a GitHub action?

dirkhh:
And the Qt5 build is the one that is falling. 
The others are just cancelled as they are in the same y'all
At least that's what it looks like to me

qt-models/diveplannermodel.cpp
```diff
@@ -288,8 +305,15 @@ static divemode_t get_local_divemode(struct dive *d, int dcNr, int cylinderId, d
...
 {
+	if (!d || mode == NOTHING)
+		return QVariant();
+
+	if (!index.isValid() || index.row() < 0 || index.row() >= divepoints.count())
+		return QVariant();
+
 	const divedatapoint p = divepoints.at(index.row());
-	bool isInappropriateCylinder = !is_cylinder_use_appropriate(*d->get_dc(dcNr), *d->get_cylinder(p.cylinderid), false);
+	cylinder_t *cyl_for_check = cylinder_index_valid(d, p.cylinderid) ? d->get_cylinder(p.cylinderid) : nullptr;
```
bstoeger: Would it make sense to encapsulate this in a function: Return a pointer to a cylinder or nullptr if the index is invalid?

bstoeger (approved):
On a cursory check, the changes look reasonable, though it is not clear to me why precisely the code crashed.

It seems that we somehow got into inconsistent state?

I would suggest adding a comment to the commit-message detailing how the DivePlannersPointModel relates to the preferences and how the inconsistent state comes to be.

qt-models/diveplannermodel.cpp
```diff
@@ -38,6 +38,12 @@ static double unit_factor()
 
 static constexpr int decotimestep = 60; // seconds
 
+static bool cylinder_index_valid(const struct dive *d, int cylinderId)
+{
+	// Allow for a cylinder one bejond the limit of the array to allow for the implied 'surface air' cylinder
```
RonOnGithub: s/bejond/beyond/

PR #4717, opened 2026-02-12
Fix packaging for ubuntu questing

hyperair:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Fix build failures of the `subsurface` package in the [releases](https://launchpad.net/~subsurface/+archive/ubuntu/subsurface/+packages) and [daily](https://launchpad.net/~subsurface/+archive/ubuntu/subsurface-daily/+packages) PPAs in ubuntu `questing`, caused by unsatisfiable build-dependency `libqt5webkit5-dev`.

There is also a separate issue of `subsurface-qt-private-headers` not having been properly copied to `questing` in the PPA, but I can't fix that in a PR.

### Changes made:

- Change `packaging/ubuntu/make-package.sh` to swap out the `libqt5webkit5-dev` build-dep for `qtwebengine5-dev` on Ubuntu Questing
- Disable LTO optimization (enabled by default in Ubuntu questing) to prevent segfaulting during application startup.

### Mentions:

@Saviq @mikeller @dirkhh

### DIFF
--- a/desktop-widgets/mainwindow.cpp
+++ b/desktop-widgets/mainwindow.cpp
@@ -16,7 +16,6 @@
 #include <QStatusBar>
 #include <QNetworkProxy>
 #include <QUndoStack>
-
 #include "core/color.h"
 #include "core/device.h"
 #include "core/divelog.h"
@@ -73,7 +72,10 @@
 #include "qt-models/yearlystatisticsmodel.h"
 #include "preferences/preferencesdialog.h"
 
-#ifndef NO_USERMANUAL
+#ifdef NO_USERMANUAL
+#include <QDesktopServices>
+#include <QUrl>
+#else
 #include "usermanual.h"
 #endif
 
@@ -205,9 +207,6 @@ MainWindow::MainWindow() :
 	diveList->setFocus();
 	diveList->expand(diveList->model()->index(0, 0));
 	diveList->scrollTo(diveList->model()->index(0, 0), QAbstractItemView::PositionAtCenter);
-#ifdef NO_USERMANUAL
-	ui.menuHelp->removeAction(ui.actionUserManual);
-#endif
 
 	updateManager = new UpdateManager(this);
 	undoAction = Command::undoAction(this);
@@ -866,7 +865,9 @@ void MainWindow::on_action_Check_for_Updates_triggered()
 
 void MainWindow::on_actionUserManual_triggered()
 {
-#ifndef NO_USERMANUAL
+#ifdef NO_USERMANUAL
+	QDesktopServices::openUrl(QUrl("https://subsurface-divelog.org/subsurface-user-manual/"));
+#else
 	if (!helpView)
 		helpView = new UserManual(this);
 	helpView->show();
--- a/packaging/ubuntu/make-package.sh
+++ b/packaging/ubuntu/make-package.sh
@@ -99,19 +99,39 @@ else
 fi
 
 cp debian/changelog ../changelog$SUFFIX
+cp debian/control ../control$SUFFIX
 
 debuild -S -d
 
 # create builds for the newer Ubuntu releases that Launchpad supports
 #
 rel=focal
-others="jammy noble questing"
-for next in $others
-do
+others="jammy noble"
+others_nowebkit="questing"
+
+build() {
+	local next="$1"
+
 	sed -i "s/${rel}/${next}/g" debian/changelog
 	debuild -S -d
 	rel=$next
+}
+
+# older ubuntu releases have libqt5webkit5-dev, so use that
+for next in $others
+do
+	build "$next"
 done
+
+# newer ubuntu releases no longer have libqt5webkit5-dev, so drop the dependency
+sed -i '/libqt5webkit5-dev/d' debian/control
+sed -i -re 's/NO_(PRINTING|USERMANUAL)=OFF/NO_\1=ON/g' debian/rules
+
+for next in $others_nowebkit
+do
+	build "$next"
+done
+
 if [ -f debian/control.bak ] ; then
 	mv debian/control.bak debian/control
 fi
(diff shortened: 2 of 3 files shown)


dirkhh:
Webengine support will be removed. This is a 150MB monster library that is not the right tool for the job.
We are working on a different toolkit to use for user manual and printing and will make this switch together with the overall switch to Qt 6 for building Subsurface. For now I think Ubuntu 25.10 will remain unsupported - we could consider just dropping print/user manual in the meantime. I might look into that if I have time.

hyperair:
Would you consider merging a PR that disables print/user manual support for Ubuntu 25.10 so that Ubuntu 25.10 users at least have a usable Subsurface version? Because The 25.04 version isn't installable in 25.10 right now and it would suck to need to run subsurface in a VM.

mikeller:
> Would you consider merging a PR that disables print/user manual support for Ubuntu 25.10 so that Ubuntu 25.10 users at least have a usable Subsurface version? Because The 25.04 version isn't installable in 25.10 right now and it would suck to need to run subsurface in a VM.

Go for it - help is appreciated and it's great if we can avoid leaving 25.10 users high and dry until we get Qt6 support done.

Bonus points if the user manual action opens https://subsurface-divelog.org/subsurface-user-manual/ in a browser until we get native manual support back.

hyperair:
@mikeller @dirkhh Done, please have a look:
 - Ubuntu questing has `libqt5webkit5-dev` build-dep dropped, and is built with `NO_USERMANUAL=ON` and `NO_PRINTING=ON`
 - When `NO_USERMANUAL` is set, the `User manual` menuitem will no longer be removed from the `Help` menu; instead it will launch https://subsurface-divelog.org/subsurface-user-manual/ in a browser using `QDesktopServices::OpenUrl`.

mikeller (approved):
Seems to do the trick. Rebasing once #4732 has been merged should fix the Fedora build problem.

@dirkhh you are the expert on the ubuntu build system - what do you think?

In a future pull request we could probably also add the selected language to the URL as `?lang=en_GB` and similar.

packaging/ubuntu/make-package.sh
```diff
@@ -99,19 +99,39 @@ else
...
+
+# older ubuntu releases have libqt5webkit5-dev, so use that
+for next in $others
+do
+	build "$next"
+done
+
+# newer ubuntu releases no longer have libqt5webkit5-dev, so use
+# qtwebengine5-dev instead
+sed -i 's/libqt5webkit5-dev/qtwebengine5-dev/' debian/control
```
dirkhh: We are going a bit back and forth here between this commit and the rest of the series... not a huge problem, but maybe we should rebase / combine these commits for them to make more sense...

desktop-widgets/mainwindow.cpp
```diff
@@ -205,9 +207,6 @@ MainWindow::MainWindow() :
 	diveList->setFocus();
 	diveList->expand(diveList->model()->index(0, 0));
 	diveList->scrollTo(diveList->model()->index(0, 0), QAbstractItemView::PositionAtCenter);
-#ifdef NO_USERMANUAL
-	ui.menuHelp->removeAction(ui.actionUserManual);
-#endif
```
dirkhh, line 210: this makes the name of the #define a bit illogical, but it's temporary in many ways, so I'm not too worried.

mikeller:
Oh and @hyperair you will have to fix the sign-off on your commit messages, see https://github.com/subsurface/subsurface/pull/4717/checks?check_run_id=64790567736.

dirkhh (changes requested):
Would you mind adding a Signed-off-by to the last two commits?
I think I'd prefer if you rebased the commits to avoid the bit of back-and-forth with QWebEngine, but that's not required.
Other than that this looks great.
But we try not to merge things without SOB, so I'll submit this as a "Request changes" even though I approve the changes themselves...

hyperair:
@dirkhh Done.

Could you also fix the package copies for `subsurface-qt-private-headers` in the PPAs please?

image

image


I think you'll need to redo the copy, selecting the "Copy existing binaries" option as opposed to "Rebuild the copied sources" when doing the copy to have it work.

dirkhh (approved):
thanks - I once again tried to get Launchpad to copy those private header packages

dirkhh:
Thank you for working on this. On my third attempt the copying of the private headers worked and I verified that we now have a Subsurface-daily package for 25.10

desktop-widgets/mainwindow.cpp
mikeller: > this makes the name of the #define a bit illogical, but it's temporary in many ways, so I'm not too worried.

@dirkhh Maybe `NO_HARD_TO_BUILD_USERMANUAL` 😁
Jokes aside, I've been toying with the thought of moving the desktop manual completely to the web version for some time. This would cut some scope from the list of things that are currently hard-to-do for us in Qt6, and after all this has been working fine for the mobile version for a long time, with no user complaints for what I can tell.

dirkhh: Well ... User manual works fine with Qt6 and QLiteHtml 😁

mikeller: It works - not sure I find the current look so fine:

image

So all in all I am wondering if we shouldn't see the effort that went into making this work with QLiteHtml as sunk cost, and avoid future effort to keep this working and possibly make it look better by moving to the browser-based approach...

dirkhh: Dive site without Internet. Liveaboard.
Not needing the Internet to look at the manual is useful. 
That's why we did all this in the first place

mikeller: > Dive site without Internet. Liveaboard. Not needing the Internet to look at the manual is useful. That's why we did all this in the first place

My hunch would be that more people who get into this situation will be bringing a mobile phone rather than a laptop, and we seem to be happy to accept this shortcoming there...

But point taken. If this is a concern then we can simply open the packaged HTML user manual in the system default browser, like what #4735 does. Or we could even package the PDF manual for a smaller, more compact binary, and open it in the system viewer.

PR #4719, opened 2026-02-13
Desktop: Trim Whitespace from the Firmware Version.

mikeller:

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

Trim extra whitespace from the firmware version reported by
libdivecomputer.

### Additional information:

This is to fix display problems in https://github.com/subsurface/subsurface/pull/4699.

It is an alternative to
https://github.com/subsurface/subsurface/pull/4708 that minimises the
amount of new C++ constructs that are introduced.

### DIFF
--- a/core/subsurface-string.cpp
+++ b/core/subsurface-string.cpp
@@ -1,4 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0
+#include <string>
+#include <algorithm>
+#include <cctype>
+
 #include "subsurface-string.h"
 
 std::string join(const std::vector<std::string> &l, const std::string &separator, bool skip_empty)
@@ -14,4 +18,19 @@ std::string join(const std::vector<std::string> &l, const std::string &separator
 		first = false;
 	}
 	return res;
+}
+
+void trim(std::string &s)
+{
+	const auto wsfront = std::find_if_not(s.begin(), s.end(), [](unsigned char c){ return std::isspace(c); });
+	s.erase(s.begin(), wsfront);
+
+	const auto wsback = std::find_if_not(s.rbegin(), s.rend(), [](unsigned char c){ return std::isspace(c); }).base();
+	s.erase(wsback, s.end());
+}
+
+std::string trimmed(std::string s)
+{
+	trim(s);
+	return s;
 }
--- a/core/subsurface-string.h
+++ b/core/subsurface-string.h
@@ -54,4 +54,7 @@ inline bool contains(std::string_view haystack, const std::string &needle)
 
 std::string join(const std::vector<std::string> &l, const std::string &separator, bool skip_empty = false);
 
+void trim(std::string &s);
+std::string trimmed(std::string s);
+
 #endif // SUBSURFACE_STRING_H


RonOnGithub:
This still kind of feels like putting a smiley face bandaid on a wart to me.
It doesn't actually fix the wart, and it's one more piece of single use plastic that will eventually end up in the ocean for something else to choke on.

There's at least one caller also using the key "DC Firmware Version".  And there's no promise that other arbitrary string fields won't have leading or trailing whitespace, or that it won't be some combination of the unicode multibyte whitespace characters which this won't catch, or all sorts of things like that.

And that's before we start on the problems related to irreversibly mangling the original string, if anything ever actually needs to use it to compare with or search for the original data again.

Why is this stray whitespace here in the first place?  And if we can't fix that at the source, why should we be lying about its existence here?  If the authoritative source says there's whitespace in the arbitrary string which is the "firmware version" (or anything else), then there *is* whitespace in that string, and it *is* significant.

It's not up to us to decide that is not valid or shouldn't be shown, any more than it might be to say "I don't like the letter Z.  We should remove that character from all the strings".

I get the ick factor that 'stray' whitespace gives some people, but this is data, not a typo in a page title.  And we shouldn't be lying about the content of data.

mikeller-nz:
> This still kind of feels like putting a smiley face bandaid on a wart to me. It doesn't actually fix the wart, and it's one more piece of single use plastic that will eventually end up in the ocean for something else to choke on.

Frankly, to me building a GUI app for end users feels like it's 80% putting bandaids onto warts, and 20% coding.

> Why is this stray whitespace here in the first place? And if we can't fix that at the source, why should we be lying about its existence here? If the authoritative source says there's whitespace in the arbitrary string which is the "firmware version" (or anything else), then there _is_ whitespace in that string, and it _is_ significant.

I think you are missing context here: https://github.com/subsurface/libdc/pull/101/
But this obviously does nothing to fix the instances of this firmware version string that have already been imported and are in people's logs now.

> I get the ick factor that 'stray' whitespace gives some people, but this is data, not a typo in a page title. And we shouldn't be lying about the content of data.

Yes, ick indeed:

image

But I think you are right, we should fix this where it is getting ick, that is when we display the firmware version.

RonOnGithub:
> Frankly, to me building a GUI app for end users feels like it's 80% putting bandaids onto warts, and 20% coding.

Indeed, but the trick is to do it with the fewest number of bandaids possible :D

> I think you are missing context here: [subsurface/libdc#101](https://github.com/subsurface/libdc/pull/101) But this obviously does nothing to fix the instances of this firmware version string that have already been imported and are in people's logs now.

Yeah, I saw the earlier discussion about being (rightfully) nervous about changing this in existing logs and possibly messing up hashes, but wasn't sure what the source of them was.  I haven't really (needed to) put my head into libdc yet - I'm trying hard to not let my current set scope creep outside what is needed to make the QML UI play nice, but there's a lot of technical debt which is the root cause of that being Harder Than It Should Be - so I'm really keen to get that down and keep it down.

But that context definitely fleshes out the "all sorts of things like that" set of issues here.  Because we're 'fixing' one very specific problem with one very specific character included for one very specific computer by very generically and fairly arbitrarily mangling a whole class of characters from the data from *all* computers.  Regardless of whether it's a problem we created that we probably should fix, or one created by the device firmware - or by its Deliberately Evil author who decided firmware version 1a should look to users like version 1, but be distinguished by the number of trailing whitespace characters - or more likely his slightly less competent nephew who screwed that up while on work experience and nobody caught it and now they can't change it either.  Or whatever.  The reason doesn't matter.  Only that it's not *strictly impossible* and is Out Of Our Control does.

Because dealing with hardware is 20% coding and 80% finding some elegant way to not lie about some horrible flaw in its design that makes everything you wanted to do with it harder than it should be - but everything that you do do with it crash horribly if you just pretend that broken reality doesn't exist.

> > I get the ick factor that 'stray' whitespace gives some people, but this is data, not a typo in a page title. And we shouldn't be lying about the content of data.
> 
> Yes, ick indeed:

Ok, so what I'm seeing from that image, is also that this was probably less of a problem, or at least less glaring, with the current code that doesn't interpret extra data at all and just displays two columns of arbitrary key/value data (which I assume is how it went unnoticed before now) - and only really becomes Max Ick if we try to interpret it and present it in a grid.

Which in turn has other problems, given that we evidently have more than one arbitrary key value being used to indicate a firmware version value.

So if we do want to assign *explicit* meanings to some keys and lay them out specially, *that* probably needs some thought and investigation and assessment too ...  At least if we want it to not be ick.

> But I think you are right, we should fix this where it is getting ick, that is when we display the firmware version.

Yes, I'd be less worried about a future full of Unintended Consequences if this was only done right at the point of display, along with a nice detailed comment about the *exact* reason for doing that.  But it too would fall over if we ever decided to allow the user to edit those fields to 'clean them up', since we'd have mangled the starting base for that too.

And at that point it will want to be a QString again, so we can just *use* QString::trimmed - instead of reinventing it and speculatively creating our own SSRFString library of functions full of things that *we don't actually use yet*, or wouldn't have needed to use if we didn't convert something away from being a QString just so we could use our own function on it instead of theirs (and the latter problem really has been making a *lot* of work for me in other code here so I'm very keen to avoid more of it wherever possible).

But honestly, what and how bad is the real ick here?  It's only *ever* going to be this evident when viewing old logs, downloaded before the libdc fix, in this new interface.  It's not ongoing ick, because all your new logs are going to be fine.  So I am kind of inclined to still prefer the So Help Me God, truth and nothing but path here - where the viewer shows the data exactly as it *is*.  Because that way you can never be mistaken about exactly what is recorded in your logs, and if *that* still really gives someone ick, they can weigh up the pros and cons of *actually* changing their logs retrospectively.

Which is possibly the Actually Best option here.  The bug is fixed in libdc.  It *only* effects users of OSTC (who I'd argue is probably biased toward the more Technically Capable end of the spectrum).  So maybe it's worth assessing *exactly* what could or will break if someone modifies their own logs to strip this out, and provide a very small shell script and some precise instructions on how to do that.

Because that way, this would be a lot more like just having a wart cut off - some brief pain, but then it's all over - rather than scope creeping its way into a case of herpes that we now need to live with forever, even after the source of that infection is very long gone.

And I really like code that we don't have to keep forever because of some past silly mistake.  It's what keeps that 80% proportion from growing.

mikeller:
Not needed any longer - trimming with `QString.trimmed()` was added directly to #4699.

PR #4722, opened 2026-02-14
Fix dive-site duplication by resetting pending-site state only on actual text edits

chaehni:

- [x] Bug fix
- [ ] Functional change
- [ ] New feature
- [x] Code cleanup
- [ ] Build system change
- [ ] Documentation change
- [ ] Language translation

This PR fixes a dive-site duplication path caused by location input state handling.

When editing the dive-site field, the widget used to reset its internal pending-site pointer (`currDs`) for many non-text key presses. If the user had selected an existing dive site and then pressed such a key before saving, the save path could incorrectly treat the value as “create new dive site” and create a duplicate entry.

The fix changes key handling so the pending-site pointer is reset only when the text actually changes. This preserves selected existing dive sites across non-text key events.

Additionally, this PR fixes an unrelated but valid loader bug in `load-git`: the comparison for appending “additional name for site” notes was inverted and is corrected from `==` to `!=`.

### Changes made:
1) `desktop-widgets/locationinformation.cpp`:
- Refactored `DiveLocationLineEdit::keyPressEvent()` to use intent based key handling logic.
- Reset `currDs` to `RECENTLY_ADDED_DIVESITE` only when the line edit text actually changed.
- Preserve behavior for popup interaction (`Escape`, `Up/Down`).

2) `core/load-git.cpp`:
- Fixed condition in `parse_dive_location()`:
  - `if (ds->name == name)` -> `if (ds->name != name)`.

3) Tests:
- No permanent test changes included in this PR.
- Existing local test suite was run and passed after the fix.

### Additional information:
- Reproduced on unpatched build with a non-text key press (Ctrl  + S) between selecting an existing dive site and saving. (all platforms)
- Verified patched behavior no longer creates duplicate dive sites in that flow.
- Issue discussion board:  [https://groups.google.com/g/subsurface-divelog/c/IOJ6Ifu2Dqo](https://groups.google.com/g/subsurface-divelog/c/IOJ6Ifu2Dqo)

### Documentation change:
No user-facing workflow change intended; this is a correctness fix for existing behavior.
No documentation update required.

### Mentions:
@mikeller

### DIFF
--- a/core/load-git.cpp
+++ b/core/load-git.cpp
@@ -226,7 +226,7 @@ static void parse_dive_location(char *, struct git_parser_state *state)
 			ds->name = name.c_str();
 		} else {
 			// and that dive site had a name. that's weird - if our name is different, add it to the notes
-			if (ds->name == name) {
+			if (ds->name != name) {
 				ds->notes += '\n';
 				ds->notes += format_string_std(translate("gettextFromC", "additional name for site: %s\n"), name.c_str());
 			}
--- a/desktop-widgets/locationinformation.cpp
+++ b/desktop-widgets/locationinformation.cpp
@@ -607,18 +607,16 @@ void DiveLocationLineEdit::setTemporaryDiveSiteName(const QString &name)
 
 void DiveLocationLineEdit::keyPressEvent(QKeyEvent *ev)
 {
+	const QString before = text();
+	const int key = ev->key();
 	QLineEdit::keyPressEvent(ev);
-	if (ev->key() != Qt::Key_Left &&
-	    ev->key() != Qt::Key_Right &&
-	    ev->key() != Qt::Key_Escape &&
-	    ev->key() != Qt::Key_Return) {
-
-		if (ev->key() != Qt::Key_Up && ev->key() != Qt::Key_Down)
-			currDs = RECENTLY_ADDED_DIVESITE;
-		else
-			showPopup();
-	} else if (ev->key() == Qt::Key_Escape) {
+	const bool textChanged = (text() != before);
+	if (key == Qt::Key_Escape) {
 		view->hide();
+	} else if (key == Qt::Key_Up || key == Qt::Key_Down) {
+		showPopup();
+	} else if (textChanged) {
+		currDs = RECENTLY_ADDED_DIVESITE;
 	}
 }
 


RonOnGithub:
I haven't poked around in this code too much yet, so I might be missing something.  But I looked at this patch because the core issue very much caught my attention - accidentally duplicated sites in my mobile device logs that I couldn't clean up has been a pet peeve of mine for a long time ...

Anyhow, my first question from what I'm seeing, is why on earth are we trying to decide this at the time the user is still interacting with the input control?

Shouldn't we just totally ignore anything they do there until they actually *submit* their final answer, and only then check to see if they selected an existing site or entered an entirely new one?  That seems both simpler and more robust.

Am I missing something?

chaehni:
> I haven't poked around in this code too much yet, so I might be missing something. But I looked at this patch because the core issue very much caught my attention - accidentally duplicated sites in my mobile device logs that I couldn't clean up has been a pet peeve of mine for a long time ...
> 
> Anyhow, my first question from what I'm seeing, is why on earth are we trying to decide this at the time the user is still interacting with the input control?
> 
> Shouldn't we just totally ignore anything they do there until they actually _submit_ their final answer, and only then check to see if they selected an existing site or entered an entirely new one? That seems both simpler and more robust.
> 
> Am I missing something?

Totally fair question, and I agree with your direction.

From debugging this, the core issue is indeed that we currently decide “new vs existing site” *during* key interaction (including non-text keys) using a sentinel value, instead of deciding once at final commit. That made the flow fragile and is what allowed accidental duplicate creation.

For this PR, I went with the least invasive fix to stop the bug with minimal behavioral change and risk in surrounding code.

If maintainers agree, I’d be in favor of a commit-time decision model:
- user finishes editing
- if text is empty: clear site
- if an existing site was explicitly selected: reuse that site
- otherwise: create a new site from entered text

I haven’t audited every related path yet, so I can’t promise there are no side effects elsewhere (map/mobile/etc.), but conceptually this is simpler and more robust.

If it were up to me, I’d also remove:
1. The `RECENTLY_ADDED_DIVESITE` sentinel state entirely.  
- It couples transient UI input state to backend behavior and is easy to desync.
2. The injected dropdown entry (second item) that effectively creates a new site with the same name as  the existing site that starts with that prefix.  
- That option is confusing in practice and can steer users into duplicate creation by mistake.

mikeller (approved):
I like 'least invasive'.

Happy to go with this for now as it fixes the bug and stops further log pollution with duplicate dive sites.

But feel free if you want to poke around and see if you can get a deeper refactor done without running into side effects.

RonOnGithub:
> For this PR, I went with the least invasive fix to stop the bug with minimal behavioral change and risk in surrounding code.

Yeah, none of what I said was critique (or even analysis) of the patch itself.  It definitely helped make the problem case much clearer, so was a good first step.  And I'm also a fan of minimal fixes.  I'd just much rather see the *problem* fixed than one of the symptoms masked - so I went straight to wondering why any of this code exists at all.

Fixing a problem and adding functionality simply by *deleting* unnecessary code is always the gold standard of minimal fixes :)

> Happy to go with this for now as it fixes the bug

Does it actually?  Doesn't it just change what you need to do to hit it?  It prevents the arrow keys from doing that, but if you typo a change which actually changes the text (and then undo that), don't we still have exactly the same problem?

@dirkhh  is there something else about this we're missing?

chaehni:
> I like 'least invasive'.
> 
> Happy to go with this for now as it fixes the bug and stops further log pollution with duplicate dive sites.
> 
> But feel free if you want to poke around and see if you can get a deeper refactor done without running into side effects.

Thank you for the approval. I will try to find some time to do a deeper refactor.


> Does it actually? Doesn't it just change what you need to do to hit it? It prevents the arrow keys from doing that, but if you typo a change which actually changes the text (and then undo that), don't we still have exactly the same problem?

Good question. In the current interaction model, the behavior of your example (type a change and undo) is intentional:

- Reusing an existing site requires explicit activation of that site from the popup (click / Enter / Tab).
- Any subsequent text edit means “you are editing free text again”, so an existing-site selection must be re-activated if that is still the intent.

So for the type/undo scenario: if the user edits after activation and does not re-activate an existing row, the flow is treated as “create new” by design.

That behavior is important if we want to allow distinct dive sites that share the same name (distinguished by other fields like coordinates/notes). If we auto-resolve by name alone at commit time, we lose that possibility and can silently bind to the wrong site.

So the real policy question is: do we still want to allow multiple sites with the same name?  
- If yes, explicit activation is the only disambiguation signal.  
- If no, we can collapse to name-based reuse on commit.

mikeller:
> So the real policy question is: do we want to allow multiple sites with the same name?

I think the answer here is 'yes' - the existing code clearly wants to allow this, and specifically contains code to support this.
So unless we go through the process of finding the original reasons for taking this approach, and find that they are outdated or were incorrect, we should not try to change it. See https://fs.blog/chestertons-fence/

dirkhh:
oh absolutely. I have sites with the same name that are thousands of miles apart. Like "Paradise". Or "Great Wall"

RonOnGithub:
> That behavior is important if we want to allow distinct dive sites that share the same name

If we accept that line of thought though, then *this* patch is probably wrong, because now the only way to invoke this "important behaviour" is to deliberately typo the name you want to duplicate and then correct that.  Which seems a long way from line of least surprise?

> So the real policy question is: do we still want to allow multiple sites with the same name?
>
>    If yes, explicit activation is the only disambiguation signal.
>    If no, we can collapse to name-based reuse on commit.

I think that's a correct assessment.

And *personally* I think allowing multiple sites with the same name is madness.  The interface we present for handling that is *awful* (do you know the GPS coordinates that distinguish all the sites in the world that are locally named Coral Garden? - and on mobile we offer *no way at all* to distinguish between them in the drop down) - so it's hard to believe people would prefer to do that deliberately than to name one "Coral Garden (here)" and the other "Coral Garden (there)".  At least if they ever plan on going back to either of them ever again.  It's a footgun with few if any positive benefits.

But if we do choose to continue allowing that (as opposed to just supporting legacy logs where they already exist, which we can do independently to fixing this) - then it should either be *absolutely clear and explicit*, in the tick a box sense, not some secret easter egg that you invoke by making a typo then fixing it - or the current behaviour where *any interaction* with the text entry box puts you in "this is a new entry" mode.

> unless we go through the process of finding the original reasons for taking this approach, and find that they are outdated or were incorrect, we should not try to change it.

Which I think applies to this change too, because if anyone really wants and expects this behaviour, I'm sure this will break their current muscle memory and their spacebar heating (https://xkcd.com/1172/).

And I'm not proposing we change anything until we understand the history, which is why I tagged in Dirk, and think that if he doesn't have some Definite Answers to clarify things, that a trip through git history is an essential prerequisite to changing anything about this.

RonOnGithub:
> oh absolutely. I have sites with the same name that are thousands of miles apart. Like "Paradise". Or "Great Wall"

And how do you know which is which to choose between them in the mobile app?

dirkhh:
I don't say that YOU need to have sites with the same name.
I said that in my workflow I have them and I use them.
How and where I do this really isn't relevant to the conversation. But since you asked: I always first download dives from my Garmin Descent - which only works on the desktop. And then set the name for the new site that it creates. Which shows me the distance to all previous dive sites of the name.

RonOnGithub:
> I don't say that YOU need to have sites with the same name.

Right, but the problem that opened this is "People are making mistakes with the UI".

And this patch just changes the way they can make those mistakes, which makes it harder for some people who do want to do this to do it the way they have previously, but doesn't prevent the people who didn't from still making them.

So the core questions are:
 - what do we want to allow?
 - what UI do we need, to allow doing that unambiguously without having to have read the source code?

And if you unambiguously want to continue the ability to have ambiguous names (and always have GPS data that makes that easier for *you* to do), then I'm ok with that, even if it's not *my* first choice - but it means how do we *properly* fix the UI, for both desktop and mobile, is still an open question.

It seems that the problem here is less about *how* the "add a new site" mode is triggered than that it's not clear to users *when* it is, or how to undo that (and if they successfully have) if they did it accidentally.

chaehni:
Thanks everyone for the input and feedback.

> because now the only way to invoke this "important behaviour" is to deliberately typo the name you want to duplicate and then correct that

Depends on what we mean by “important behavior.”

If we mean “create a new dive site with this name,” that is still explicit and available:
- the user can select the **new dive site with this name** entry, or
- select the **existing dive site with this name** entry (if one exists).

If we mean “create a new site with the same name *and cloned state* from the currently assigned site,” then yes, that remains a hidden behavior. From reading deeper, that cloning appears intentional in the command layer:

```cpp
static struct dive_site *createDiveSite(const std::string &name)
{
	struct dive_site *ds = new dive_site;
	if (current_dive && current_dive->dive_site)
		*ds = *current_dive->dive_site;

	// If the current dive has a location, use that as location for the new dive site
	if (current_dive) {
		location_t loc = current_dive->get_gps_location();
		if (has_location(&loc))
			ds->location = loc;
	}

	ds->name = name;
	return ds;
}
```

So editDiveSiteNew() does not start from a blank site. It clones the current site first, then overrides the name.
This copy even includes associated dives of the original site causing another bug. The newly cloned site has a wrong number of dives assigned since the reverse index `dive_site -> dives` carried over from the original. The forward index `dive -> dive site` remains correct. This causes the dive site table to show a completely wrong number of dives for cloned sites.

To summarize:
The observed duplication comes from that cloning behavior combined with UI-state flipping during text editing: if flow accidentally re-enters the new site branch after an existing site was already assigned, it creates a near-identical copy.

I’m not fully sure what the right next step is.

If cloning is intentional behavior, then I agree with @RonOnGithub  that it should be explicit in the UI. Right now it is too implicit and easy to trigger accidentally.

At minimum, the choices should be clearly separated, for example:
- **Create new site with this name**
- **Use existing site with this name**
- **Clone existing site with this name**

If cloning is a supported workflow, it should be a deliberate, clearly labeled action, not a side effect of transient input-state changes. In addition, we then also need to fix the issue with wrongly associated dives to this cloned site.

mikeller:
Could it be that the cloning is not intended to be used in the drop-down, but is instead used in a different place? Cloning from the drop-down seems to be an odd user interaction.

chaehni:
> Could it be that the cloning is not intended to be used in the drop-down, but is instead used in a different place? Cloning from the drop-down seems to be an odd user interaction.

Sorry, should have mentioned this. The clone behavior is only reachable from the location textbox commit flow as far as I can tell.

chaehni:
The commit where cloning was introduced is this one: https://github.com/subsurface/subsurface/commit/6f574c53a38ad848bfd364acb0ae16680c3f6c8a

Furthermore, I just found that cloning via drop down is very much intentional.

Screenshot 2026-02-15 120304

The relevant code specifically checks that a dive site is already assigned: [https://github.com/subsurface/subsurface/blob/master/desktop-widgets/locationinformation.cpp#L424-L427](https://github.com/subsurface/subsurface/blob/master/desktop-widgets/locationinformation.cpp#L425-L427)

mikeller:
Ah that's that functionality - it for dive computers with GPS it adds the location to a new dive site - but to my knowledge if this is a newly added dive then no 'phantom site' with the same coordinates will be stored, so this isn't really 'cloning'.

@bstoeger and @dirkhh might know more what other uses it is intended for.

chaehni:
> Ah that's that functionality - it for dive computers with GPS it adds the location to a new dive site - but to my knowledge if this is a newly added dive then no 'phantom site' with the same coordinates will be stored, so this isn't really 'cloning'.
> 
> @bstoeger and @dirkhh might know more what other uses it is intended for.

That I'd understand and makes sense imo. But it explicitly does both. See again the createDiveSite code.
It first tries to do a clone and then overwrites the location and name.

```
static struct dive_site *createDiveSite(const std::string &name)
{
	struct dive_site *ds = new dive_site;
	if (current_dive && current_dive->dive_site)
		*ds = *current_dive->dive_site;

	// If the current dive has a location, use that as location for the new dive site
	if (current_dive) {
		location_t loc = current_dive->get_gps_location();
		if (has_location(&loc))
			ds->location = loc;
	}

	ds->name = name;
	return ds;
}
```

dirkhh:
> Ah that's that functionality - it for dive computers with GPS it adds the location to a new dive site - but to my knowledge if this is a newly added dive then no 'phantom site' with the same coordinates will be stored, so this isn't really 'cloning'.
> 
> @bstoeger and @dirkhh might know more what other uses it is intended for.

The way I remember this... part of the problem was that in order to allow undo of adding a dive site via download from GPS enabled dive computer... we need to add some code that at the time felt unintuitive to me because one needs to think backwards to allow for the undo to do its thing. And I believe that's what makes the code a little harder to follow?

It's been a long time since I've really tried to crawl through that part. @bstoeger - any thoughts from you?

mikeller:
@chaehni Have you had any further insights into this, or should we go ahead and merge this as a fix for the issue at hand?

chaehni:
Hi @mikeller

I have a more robust fix ready.

It is based on the assumption that cloning dive sites in this flow is not intended behavior. With that assumption, the cloning logic is removed entirely, while still copying the location information from the dive itself.

From my testing, undo and redo continue to work as expected.

I was waiting for additional background on why the cloning logic exists in the first place, in case there is a specific use case I might be overlooking.

Should I push the changes?

mikeller:
> I have a more robust fix ready.
> 
> It is based on the assumption that cloning dive sites in this flow is not intended behavior. With that assumption, the cloning logic is removed entirely, while still copying the location information from the dive itself.
> 
> From my testing, undo and redo continue to work as expected.
> 
> I was waiting for additional background on why the cloning logic exists in the first place, in case there is a specific use case I might be overlooking.
> 
> Should I push the changes?

Do you want to push the changes in a separate pull request, to allow people to test both?

As per the prior discussion, the core use cases for this functionality are to enable the creation of dive sites with GPS coordinates downloaded from a dive computer (until now only Garmin Descent models support this), and the 'Undo' functionality for this step.
I can do some testing for this, and @dirkhh should be able to as well.

chaehni:
Sounds good. I'll create another PR. Thank you for the help.

mikeller:
Superseded by #4731, closing.

PR #4724, opened 2026-02-15
Install qlitehtml in correct directory on non-mac

bstoeger:
The CMAKE_INSTALL_PREFIX was only set for mac.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

The qt6 version wouldn't compile for me, because it tried to install qlitehtml to "/usr/local".

I'm certain that this is the wrong fix, since it now passes the correct path twice on Mac, but for now this kludge does it for me.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -469,7 +469,7 @@ if [[ "$QUICK" != "1" && "$BUILD_DESKTOP" == "1" && "$BUILD_WITH_QT6" == "1" ]] 
 	git submodule update
 
 	# qlitehtml currently (2025-12-01) only allows in source tree builds
-	cmake $MAC_CMAKE .
+	cmake $MAC_CMAKE -DCMAKE_INSTALL_PREFIX=${INSTALL_ROOT} .
 	make
 	make install
 	if [ "$PLATFORM" = Darwin ] ; then


mikeller:
> I'm certain that this is the wrong fix, since it now passes the correct path twice on Mac, but for now this kludge does it for me.

If it's stupid but it works, it ain't stupid. 😉

mikeller (approved):

bstoeger:
Ok, that didn't work. :-1: 

> $ ./subsurface -v
> [...]
> built with Qt Version 5.15.17, runtime from Qt Version 5.15.17

dirkhh:
We have only a couple of Qt 6 builds in GitHub actions...
We need to add all the various current Linux ones...

I have Windows/ MSVC about 80% working, but that one doesn't do qlitehtml yet. I hope to get to that today

bstoeger:
For the record (on Kubuntu):

I had to install  `qt6-5compat-dev`.

Then I had to remove a bunch of `/6.9.2` from include paths in a cmake file.

Finally I had to apply:
```
diff --git a/desktop-widgets/mapwidget.h b/desktop-widgets/mapwidget.h
index 09ce7f006..1eed82a82 100644
--- a/desktop-widgets/mapwidget.h
+++ b/desktop-widgets/mapwidget.h
@@ -5,7 +5,7 @@
 #include "core/units.h"
 #include "core/subsurface-qt/divelistnotifier.h"
 #include <vector>
-#include <QQuickWidget>
+#include <QtQuickWidgets/QQuickWidget>
 #include <QList>
 
 #undef IGNORE
```
which I'm sure breaks other builds.

All that to get a Qt6 build, which crashes my laptop when showing a profile.

Clearly, the Kubuntu packages have regressed, as I am sure that this worked a year or so ago.

dirkhh:
qlitehtml is now part of the Windows MSVC build :)

I'm also starting to work on Qt 6 for Ubuntu distros, so I'm sure I'll run into all that nastiness "very soon".
If you want, feel free to look at other stuff while I sort this out? I think I've done it enough now that hopefully I'll be able to make sense of all this stuff more easily.

bstoeger:
Thanks - let's close this.

Interestingly, the map-widget crashed deep in the guts of Qt6 when invoking `centerOnRectangle`, but not when invoking `centerOnCoordinate`. I think my Qt6 installation is messed up.

dirkhh:
which Ubuntu / Kubuntu are you on?
24.04 has Qt 6.4 which is too old for the map stuff I realized a moment ago.
Oh bother.

bstoeger:
> which Ubuntu / Kubuntu are you on?

Kubuntu 25.10, Qt is 6.9.2.

But clearly, this whole combination of cmake / Qt is completely broken. It keeps adding ominous version strings to include paths.

And it is too old for current kirigami:
```
CMake Error at CMakeLists.txt:23 (find_package):
  Could not find a configuration file for package "ECM" that is compatible
  with requested version "6.23.0".

  The following configuration files were considered but not accepted:

    /usr/share/ECM/cmake/ECMConfig.cmake, version: 6.17.0
```

This is _extremely_ frustrating!

dirkhh:
I'm not sure if this was public or private exchanges... but I mentioned several times in the last few weeks that I thought we'd be better off switching away from Kirigami and at least implementing all parts of our UI ourselves. But that of course is its own can of worms, since QML is incredibly limited in the features it provides.

The `qt6_debian` branch _should_ have enough to get Desktop to build on your Kubuntu 25.10 - please let me know if that isn't true. But of course that's Desktop, so no Kirigami. And btw, the reason we are on such an ancient version of Kirigami is that they keep breaking our UI design with small changes that they make between minor versions.

The tight coupling of the various parts of KDE's software stack of course makes sense for them. But for an outside project trying to use some of this as if it was an independent library intended for independent use - not a great experience.

PR #4725, opened 2026-02-16
Windows MSVC build

dirkhh:

- [x] Build system change

This is quite an extensive change - and invasive. I had some help from Claude, which means extra scrutiny is needed. I did review all of these (my workflow is to not sign off on a commit until I have read and understood it and am comfortable with, so on the powershell stuff and some of the mechanical replacements for compiler feature differences where Claude wrote the initial code, I spend a fair bit of time digging and prodding to feel comfortable. And... the powershell stuff... still feels weird...

### Changes made:

- added new powershell scripts and a new GitHub workflow
- changes to cmake files to accomodate the native Windows build
- changes to C++ files to deal with compiler incompatibilities (some of these changes are really sad)
- I needed to also make one change to libdivecomputer - so I need a PR for that at some point as well; those are the same type of compiler incompatibilites... the rest of libdc is MSVC clean, but our extensions were not.

### Mentions:

@bstoeger there are a couple small C++ changes where I ended up trusting Claude because I wasn't quite sure how to do this right


.github/workflows/windows-msvc-qt6.yml
```diff
@@ -0,0 +1,371 @@
...
+  # Qt version - use 6.8.x which is an LTS release (promised until 2029) and
+  # includes Positioning and Location modules
+  QT_VERSION: "6.8"
+  QT_ARCH: "win64_msvc2022_64"
+  # vcpkg configuration
+  VCPKG_DEFAULT_TRIPLET: x64-windows
+  # Build configuration
+  BUILD_TYPE: Release
+
+concurrency:
```
mikeller: **question:** What is the reason we are introducing concurrency here?
While this may be considered best practice, we should probably adopt it in a separate pull request across all workflows (and add some commentary about it) if we want to follow it.

.github/workflows/documentation-build.yml
```diff
@@ -6,12 +6,18 @@ on:
...
+      cache-key:
+        description: "The cache key used for the documentation"
+        value: ${{ jobs.documentation.outputs.cache-key }}
 
 jobs:
   documentation:
     runs-on: ubuntu-24.04
     container:
       image: ${{ inputs.container-image }}
+    outputs:
```
mikeller, line 19: **praise:** Good idea, this is much better than the 'hit the right hash key' golf that we are playing in other workflows.

.github/workflows/windows-msvc-qt6.yml
```diff
@@ -0,0 +1,371 @@
...
+        shell: pwsh
+        run: |
+          New-Item -ItemType Directory -Force -Path "${{ github.workspace }}\vcpkg-cache"
+
+      - name: Restore vcpkg cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ github.workspace }}/vcpkg-cache
+          # key: vcpkg-x64-windows-${{ hashFiles('.github/workflows/windows-msvc-qt6.yml') }}
+          key: vcpkg-x64-windows-v1
```
mikeller: **suggestion:** This should probably cache on someting that changes when the build changes, or we'll get stuck on builds using outdated caches before too long.

Also note that I've tried to name caches (where we can) `<architecture>-<type>-<id>`, making it easier to use `gh cache list` and `gh cache delete` to manage caches if needed.

.github/workflows/windows-msvc-qt6.yml
```diff
@@ -0,0 +1,371 @@
...
+
+          # Generate revision.h with git commit hash
+          $revision = git rev-parse --verify HEAD
+          "#define DC_VERSION_REVISION `"$revision`"" | Out-File -FilePath "src/revision.h" -Encoding utf8
+          Write-Host "Generated revision.h with revision: $revision"
+
+      - name: Set up MSBuild
+        uses: microsoft/setup-msbuild@v2
+
+      - name: Build libdivecomputer with MSVC
```
mikeller, line 164: **thought:** In this case we should probably also re-enable the MSVC tests in our not-fork of libdivecomputer: https://github.com/subsurface/libdc/blob/51cfeb7d581842bb4ebfd83eceb17ed3a5d29a9e/.github/workflows/build.yml#L130-L159

core/membuffer.cpp
```diff
@@ -220,10 +220,12 @@ void put_quoted(struct membuffer *b, const char *text, int is_attribute, int is_
 		case 0:
 			escape = NULL;
 			break;
-		case 1 ... 8:
+		case 1: case 2: case 3: case 4: case 5: case 6: case 7: case 8:
 		case 11:
 		case 12:
-		case 14 ... 31:
+		case 14: case 15: case 16: case 17: case 18: case 19: case 20:
```
mikeller: **thought:** Ok, this makes me wonder if `switch` is still the right fit for this, or if an `if` / `else` will be better.

core/metadata.cpp
```diff
@@ -78,11 +78,15 @@ static bool parseExif(QFile &f, struct metadata *metadata)
...
 		case 0xffc2:
 		case 0xffc4:
-		case 0xffd0 ... 0xffd7:
+		case 0xffd0: case 0xffd1: case 0xffd2: case 0xffd3:
+		case 0xffd4: case 0xffd5: case 0xffd6: case 0xffd7:
 		case 0xffdb:
 		case 0xffdd:
 		case 0xffe0:
-		case 0xffe2 ... 0xffef:
+		case 0xffe2: case 0xffe3: case 0xffe4: case 0xffe5:
```
mikeller: **thought:** See above, might want to switch to `if` / `else`.

core/parse-xml.cpp
```diff
@@ -1148,7 +1157,7 @@ static degrees_t parse_degrees(const char *buf, const char **end)
 
 	/* Rounding */
 	switch (*buf) {
-	case '5' ... '9':
+	case '5': case '6': case '7': case '8': case '9':
```
mikeller: No, just no, Claude. 😬

core/save-git.cpp
```diff
@@ -65,10 +64,12 @@ static void quote(struct membuffer *b, const char *text)
 		case 0:
 			escape = NULL;
 			break;
-		case 1 ... 8:
+		case 1: case 2: case 3: case 4: case 5: case 6: case 7: case 8:
 		case 11:
 		case 12:
-		case 14 ... 31:
+		case 14: case 15: case 16: case 17: case 18: case 19: case 20:
```
mikeller: **suggestion:** `if` / `else`

core/save-git.cpp
```diff
@@ -701,20 +702,10 @@ static void create_trip_name(dive_trip *trip, struct membuffer *name, struct tm
 
 		for (i = 0; i < MAXTRIPNAME; ) {
 			char c = *p++;
-			switch (c) {
-			case 0:
-			case ',':
-			case '.':
+			if (c == 0 || c == ',' || c == '.')
```
mikeller, line 693: **praise:** And here this is done in the sensible way.

core/windows.cpp
```diff
@@ -187,7 +228,7 @@ int enumerate_devices(device_callback_t callback, void *userdata, unsigned int t
 		RegCloseKey(hKey);
 	}
 	if (transport & DC_TRANSPORT_USBSTORAGE) {
-		int i;
+		int j;
```
mikeller, line 231: **question:** Any idea what was wrong with `i` here? 😆

qt-models/divesummarymodel.cpp
```diff
@@ -67,10 +67,14 @@ QVariant DiveSummaryModel::data(const QModelIndex &index, int role) const
 		return dataDisplay(row, 1);
 	case SECTION_ROLE:
 		switch (row) {
-		case DIVES ... PLANS: return tr("Number of dives");
-		case TIME ... TIME_AVG: return tr("Time");
-		case DEPTH_MAX ... DEPTH_AVG: return tr("Depth");
-		case SAC_MIN ... SAC_AVG: return tr("SAC");
+		case DIVES: case DIVES_EAN: case DIVES_DEEP: case PLANS:
```
mikeller: **question:** `if` / `else`?

scripts/get-version.ps1
```diff
@@ -0,0 +1,153 @@
+#Requires -Version 5.1
```
mikeller, line 1: **thought:** The thought of having to maintain our already fragile auto-versioning system is a bit frightening to me - and in particular in something that is as non-shell-like as PowerShell.

I think we should try to find a way to run the version 'magic' in linux-land even for these builds.

smtk-import/smartrak.cpp
```diff
@@ -312,7 +315,7 @@ static void smtk_wreck_site(MdbHandle *mdb, char *site_idx, struct dive_site *ds
 					}
 					break;
 				}
-				case 6 ... 9:
+				case 6: case 7: case 8: case 9:
```
mikeller: **question:** `if` / `else`?

stats/statsvariables.cpp
```diff
@@ -901,10 +901,10 @@ struct DateQuarterBinner : public SimpleContinuousBinner<DateQuarterBinner, Date
 
 		int year = tm.tm_year;
 		switch (tm.tm_mon) {
-		case 0 ... 2: return { year, 1 };
-		case 3 ... 5: return { year, 2 };
-		case 6 ... 8: return { year, 3 };
-		default:      return { year, 4 };
+		case 0: case 1: case 2: return { year, 1 };
```
mikeller: **question:** `if` / `else`?

mikeller (commented):
All in all this looks good to me, a few minor things that I think can be improved, in particular how the versioning is done.

Once merged, do we want this to add its artifact to the releases in`nightly-builds`, just like we do for macOS / Qt6?

.github/workflows/windows-msvc-qt6.yml
dirkhh: Pretty sure that's a Claude thing that I missed in my review. 
I don't remember doing this intentionally

.github/workflows/windows-msvc-qt6.yml
dirkhh: This needs more thought, honestly. 
vcpkg is useful, not I'd like more control to ensure that we get the versions we want. So yeah, this needs better granularity. 

What you see here is my attempt to make the builds go faster as I iterated through all the necessary other changes

.github/workflows/windows-msvc-qt6.yml
dirkhh: Those need to be updated to include all the files we added... IIRC 
I can look at that

core/membuffer.cpp
dirkhh: Yup. This was all done mechanically with the help of Claude 

A couple were already converted

core/save-git.cpp
dirkhh: Yeah, a couple of those I overruled and then they got better

Should have gone back and fixed the rest. 

But hey, it compiled 😇😁

core/windows.cpp
dirkhh: Yes, the msvc compiler complained because we reused a variable name inside the loop or something

scripts/get-version.ps1
dirkhh: I looked at that and it made things really complicated in the builds. But maybe we can pass this between steps like what I ended up doing with some of the other stuff

dirkhh (commented):
Great feedback

Happy to look at implementing these changes this afternoon

dirkhh:
OK, pushed a few commits to deal with the review feedback.

dirkhh:
pushed a small cleanup / refactoring plus hopefully a fix for the NSIS install failure. Also this should stop the double execution of the Windows action... I thought I pushed those last two last night, but apparently I didn't (or they got lost). I double checked that I undid nothing of what you did @mikeller when I force pushed this one :)

dirkhh:
That nsis install disaster is ridiculous. 
WTF Microsoft

mikeller:
All good. Not sure why the Fedora build could not find `qt6-qtbase-private-devel` when I tried installing this first time round, before I went on the tangent of mocking it...

mikeller (approved):

PR #4727, opened 2026-02-16
build: increase C++ version to C++20

bstoeger:
New kirigami compiles with C++20. We could try to only build kirigami with C++20, but let's see how this goes.

The only necessary adaptation was a named initializer list.

This used to be a gcc extension and with becoming standard, it became more struct. Oh well.

- [ ] Bug fix
- [ ] Functional change
- [ ] New feature
- [ ] Code cleanup
- [x] Build system change
- [ ] Documentation change
- [ ] Language translation

Since new Kirigami needs C++20, try to increase the version. I fear that we still have old builds that do not fully support C++20, so we might have to selectively set the standards version.

Let's see.


bstoeger:
With this commit, kirigami v6.17.0 compiles  for me in-tree.

However the mobile build now fails with 
```
make[2]: *** No rule to make target 'qrc_shaders.cpp', needed by 'CMakeFiles/subsurface-mobile.dir/qrc_shaders.cpp.o'.  Stop.
```

I'm so out of the loop that I don't even really know what a shader is. :-/

dirkhh:
As I said elsewhere - MSVC also require C++20 in order to compile. So if we switch everywhere, that means we can drop one commit from that branch 😁

The shaders are needed for QML to talk to the graphics backend (which of course is different on different OSs - the days of everyone wanting OpenGL are long gone). I vaguely remember just that issue and hacking around it in my attempts to get QML/Qt6 to work, but it's been months since the last time I looked at that code. So I don't remember off hand how I hacked around that.

dirkhh:
BTW: I think you'll be happier with the MSVC branch merged as that contains a bunch of weird fixes for initializers and stuff to make them work (just commenting on the one initializer fix that I saw in this commit)

bstoeger:
> BTW: I think you'll be happier with the MSVC branch merged as that contains a bunch of weird fixes for initializers and stuff to make them work (just commenting on the one initializer fix that I saw in this commit)

Good plan. I'll continue on the MSVC branch (tomorrow).

